Issue #13343, opened 2022-05-03
Limit the Vulnerability Detector task message queue size.


MiguelazoDS:
# Description

As part of the epic #9220 a task message queue was implemented in https://github.com/wazuh/wazuh/issues/13108. It is required to fix the queue to a maximum size instead of letting it grows indefinitely. 

# Requirements

- [x] Define a maximum default size for the task message queue.
- [x] Implement a setting that rules the maximum size of the task message queue.
- [x] Modify the behavior of the working thread to wait when attempting to insert a new message in the queue if is full.

MiguelazoDS:
# Description

The task message struct

```c
typedef struct vu_task_msg_ctx_t {
    int task_id;
    task_status task_status;
    char *status_message;
} vu_task_msg_ctx_t;
```
Occupies 16 Bytes per node plus the string allocated for the status_message pointer (maximum 128)

## Maximum task message queue for: 

- 1000 items = 144000Bytes = 140,65KiBytes
- 10000 Items = 1440000Bytes = 1,37MiBytes
- 100000 Items = 14400000Bytes = 13.73MiBytes

## Realistic

In real cases there are no messages 128 characters long, they are around 20 characters long.

- 1000 Items = 36000Bytes = 35.15KiBytes
- 10000 Items = 360000Bytes = 351.5KiBytes
- 100000 Items = 3600000Bytes = 3.43MiBytes

MiguelazoDS:
# Sequence diagram 

<details>
<summary>Click to expand!</summary>

```plantuml
@startuml

' Participants
control Scheduler_Thread as ST #orange
queue Message_Queue as MQ #lightgreen
control Working_Thread as WT #orange

' Sequence
ST -> ST 
note left #lightblue : Reads message queue size "n"\nfrom internal settings
ST -> MQ 
note right #lightblue : Initializes empty queue\nwith size=n
loop Pending tasks
  ST -> ST 
  note left #lightblue : Schedules task \n(tasks queue is \nnot shown in this diagram)
  WT -> WT 
  note right #lightblue : Executes task \n(tasks queue is \nnot shown in this diagram)
  alt message queue not full
    WT -> MQ 
    note right #lightblue : Enqueues response in \nmessages queue
  else 
    WT -> WT 
    note right #lightblue : Working thread waits
  end
  MQ -> ST 
  note left #lightblue : Reads message from queue
  MQ --> WT 
[...]

Issue #9220, opened 2021-07-06
Improve the Vulnerability Detector and Syscollector modules to allow a manual scan executions


pereyra-m:
# Description

Currently, the Vulnerability Detector and Syscollector modules work scanning at intervals of time, and it isn't possible to perform a scan request to these modules.

The aim of this epic is to design and implement a mechanism that allows the users to manually trigger scans of both modules. This will enable the possibility of having quick knowledge of the vulnerabilities that affect an agent without waiting for the next scan.

The analysis of this issue began in #9219.

# High-Level Requirements

- [ ] The Wazuh UI should expose buttons in the individual agent views to trigger the scans.
- [ ] The scans should be capable of being triggered also using the Wazuh API.
- [ ] For the Vulnerability Detector module, we should also allow the user to trigger a feeds update.
- [ ] The Wazuh UI should expose information about the scans executed.
- [ ] In case of triggering both, Syscollector and Vulnerability Detector scans at the same time, the Vulnerability Detector scan should start after the Syscollector scan finalized the data synchronization.
- [ ] The Task Manager should be used to register and track the progress of all the scanning tasks.

# Design

## Syscollector scans

For the case of the Syscollector scans, we leverage the capabilities of the **AR forwarder**. This component already implements the capability of sending actions to be taken to the agents. With this, the implementation involves:

- The Wazuh API will register the tasks in the Task Manager and then, it will send a message to the agents by using the **AR forwarder** to trigger the scans.
- Modification of the **receiver** component to handle the new message that instructs the agent to start a Syscollector scan.
- Implement the trigger of the Syscollector scan when the message is handled by the **receiver** component.

**TO BE DEFINED:** After the Syscollector scan, when the synchronization of the data finalizes, the task registered in the task manager should be set as **DONE**. For doing this, there are two options that should be evaluated:

1. When a global integrity checksum is received and it matches with the one calculated by the manager, the Syscollector tasks in the Task Manager for this particular agent should be set as DONE.
2. When analysisd determines that he shouldn't trigger a data synchronization with the agent, the Syscollector tasks in the Task Manager for this particular agent should be set as DONE.

## Vulnerability Detector scans

For the Vulnerability Detector module, we decided to implement two threads:

- **Scheduler thread**: This thread is actually the main thread of the vulnerability detector module. It will be responsible of monitoring the intervals of feed updates and scans, and will implement the interaction with the task manager as well. For details on the design of this thread, check the next issue:
[...]

PR #13395, opened 2022-05-06
Task message queue limited size

MiguelazoDS:
|Related issue|
|---|
|#13343|

## Description

This PR limits the task message queue to a maximum size. Once the queue reaches its maximum capacity, the working thread won't be able to keep pushing messages.

## Tests

  - [x] Linux
- [x] Source installation
- [x] Source upgrade

- Memory tests for Linux
  - [x] Scan-build report

### DIFF
--- a/etc/internal_options.conf
+++ b/etc/internal_options.conf
@@ -359,6 +359,9 @@ wazuh_database.interval=60
 # 0. Use system default
 wazuh_database.max_queued_events=0
 
+# Vulnerability Detector task message queue size
+wazuh_vuln_detector.task_message_queue_size=10000
+
 # Enable download module
 # 0. Disabled
 # 1. Enabled (default)
--- a/etc/internal_options.conf
+++ b/etc/internal_options.conf
@@ -360,7 +360,7 @@ wazuh_database.interval=60
 wazuh_database.max_queued_events=0
 
 # Vulnerability Detector task message queue size
-wazuh_vuln_detector.task_message_queue_size=10000
+wazuh_vuln_detector.message_queue_size=10000
 
 # Enable download module
 # 0. Disabled
--- a/src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
+++ b/src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
@@ -415,7 +415,7 @@ int *vu_queue;
 // Define time to sleep between messages sent
 int usec;
 
-w_linked_queue_t *vu_messages_queue = NULL;
+w_queue_t *vu_messages_queue = NULL;
 w_linked_queue_t* vu_tasks_queue = NULL;
 
 const wm_context WM_VULNDETECTOR_CONTEXT = {
@@ -5276,7 +5276,7 @@ void wm_vuldet_create_and_send_message(vu_task_ctx_t* task, task_status TASK_STA
                 break;
         }
         if (message) {
-            linked_queue_push_ex(vu_messages_queue, (void*) message);
+            queue_push_ex_block(vu_messages_queue, (void*) message);
         }
     }
 }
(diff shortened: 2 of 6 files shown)


MiguelazoDS:
# Test

For a message queue size 1, the working thread will wait indefinitely trying to enqueue a message. The scheduler thread won't be able to register new tasks in the working queue.

image

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.h
```diff
@@ -121,6 +121,10 @@
 #define VU_BUILD_REF_BUGZ "https://bugzilla.redhat.com/show_bug.cgi?id=%s"
 #define VU_BUILD_REF_RHSA "https://access.redhat.com/errata/%s"
 
+// Task message queue size
+#define VU_TASK_MSG_MIN_QUEUE_SIZE 1000
```
pereyra-m, line 125: We should check these limits, because they can't be modified by the user without re-compiling.
Specially the max size, considering that surpassing it doesn't generate a failure or an instability, only an increase in the memory.

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
```diff
@@ -8885,12 +8885,13 @@ int wm_vuldet_get_software(int agent_id, bool not_triaged, cJSON** requested_ite
 
 void wm_vuldet_init_queues() {
     vu_tasks_queue = linked_queue_init(vu_tasks_queue_push_callback, vu_tasks_queue_pop_callback);
-    vu_messages_queue = linked_queue_init(NULL, NULL);
+    vu_messages_queue = queue_init(getDefine_Int("wazuh_vuln_detector", "task_message_queue_size",
```
pereyra-m, line 8888: I think the read of the internal options file should occur outside this method. For example, at the beginning of the vulnerability detector main.

etc/internal_options.conf
```diff
@@ -359,6 +359,9 @@ wazuh_database.interval=60
 # 0. Use system default
 wazuh_database.max_queued_events=0
 
+# Vulnerability Detector task message queue size
+wazuh_vuln_detector.task_message_queue_size=10000
```
pereyra-m, line 363: Considering that the corresponding queue is called `vu_messages_queue`, I'd remove the word `task` from the setting name.

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
```diff
@@ -5276,7 +5276,7 @@ void wm_vuldet_create_and_send_message(vu_task_ctx_t* task, task_status TASK_STA
                 break;
         }
         if (message) {
-            linked_queue_push_ex(vu_messages_queue, (void*) message);
+            queue_push_ex_block(vu_messages_queue, (void*) message);
```
pereyra-m, line 5279: If the queue is full then vulnerability detector will stop working. I think this situation should be properly warned to the user.

pereyra-m (changes requested):
GJ!
Some comments to discuss with the team.

etc/internal_options.conf
Dwordcito: In what issue will we attack the documentation of this internal configuration?

MiguelazoDS: @Dwordcito We have this tracking issue with all the required changes in the documentation related to the epic #9220 
https://github.com/wazuh/wazuh-documentation/issues/5186

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
MiguelazoDS: Added a warning message

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.h
MiguelazoDS: I considered that 100k would be hard to reach if for some reason the task message queue starts to grow giving enough time to address the issue. 
I did a quick test of the heap for different queue sizes here https://github.com/wazuh/wazuh/issues/13343#issuecomment-1125096130

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
pereyra-m: Great! 
I'd just modify the `Working thread will stop executing tasks` part for something more clear for the user.
We should say that `vulnerability detector` will be blocked, because the `working thread` is an internal name.

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
```diff
@@ -5517,7 +5520,10 @@ void *wm_vuldet_main(wm_vuldet_t* vuldet) {
 
     wm_vuldet_check_db();
 
-    wm_vuldet_init_queues();
+    int message_queue_size = getDefine_Int("wazuh_vuln_detector", "message_queue_size",
+                                           VU_TASK_MSG_MIN_QUEUE_SIZE, VU_TASK_MSG_MAX_QUEUE_SIZE);
+
+    wm_vuldet_init_queues(message_queue_size);
```
pereyra-m, line 5526: Great!

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.h
pereyra-m: We need to consider the amount of messages, not only their size. Also, the max EPS could be limited and a full queue prevents vulnerability detector from working.

Remember that all the scan by interval tasks generate one message for agent and their interval is around 5 minutes by default.

I'd increase `VU_TASK_MSG_MAX_QUEUE_SIZE` and leave the default value as you defined.

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
MiguelazoDS: Moved!

pereyra-m (approved):
LGTM!

Dwordcito (approved):

PR #13418, opened 2022-05-10
Added PoC with rxcpp builder abstraction and other suggestions

JavierBejMen:
|Related issue|
|---|
|#13564|

This Draft PR implements abstraction of rxcpp in Engine builder module. This PR it's only for visibility and easy of discuss purpose. 

The main goal is to show how we can abstract rxcpp of our builders, where builders builds a graph describing the whole environment, assets sub-graphs with each asset operations. Once we have this description a factory of the chosen back end should be able to implement the pipeline with the given description. 

In this approach a builder is a function that returns a connectable, `std::function<std::shared_ptr<Connectable>(std::any)>`, Combining the builders results in a graph of connectables, where Connectable specializations indicate the semantics of the Connectable, mainly Group and Operation connectable to represent internal and leaf nodes respectively.

Other minor changes are implemented in the PoC to be discussed too, such as Json abstraction, Registtry without Register class, operation as function `Result(Event)`, Event wrapper class to force value move semantics, Optimization of the Connectable graph, an rxcpp factory example and more.

Target _builder_test is available with an end-to-end example of a decoder graph, along some tests for the added classes.
All code has been implemented under `source/_builder` and `test/source/_builder` folders.


src/engine/source/_builder/asset.hpp
```diff
@@ -0,0 +1,128 @@
...
+#include <fmt/format.h>
+
+#include "_builder/expression.hpp"
+#include "_builder/json.hpp"
+#include "_builder/registry.hpp"
+
+class Asset
+{
+public:
+    enum class Type
```
LucioDonda, line 16: is there a way in which we don't duplicate this enum? there's another with env in Builder.hpp

src/engine/source/_builder/environment.hpp
```diff
@@ -0,0 +1,331 @@
...
+        {
+            auto assetNames = json.getArray();
+
+            m_graphs.insert(
+                std::make_pair<std::string,
+                               Graph<std::string, std::shared_ptr<Asset>>>(
+                    std::string {name},
+                    {std::string(name + "Input"),
+                     std::make_shared<Asset>(name + "Input",
+                                             getAssetType(name))}));
```
LucioDonda, line 142: could we get the asset type first and store it in a variable inside the for loop in order to avoid calling it 3 times in the same block ?

src/engine/source/_builder/asset.hpp
```diff
@@ -0,0 +1,128 @@
...
+        if (namePos != objectDefinition.end())
+        {
+            m_name = std::get<1>(*namePos).getString();
+            objectDefinition.erase(namePos);
+        }
+        else
+        {
+            throw std::runtime_error("Asset definition must have a name");
+        }
+
```
LucioDonda, line 56: This block gets repeated several times, can we assign an enum on each stage (with it's weight) and put it inside a loop.
First Stage::name = 0, Stage::parent and so on.
This way we also avoid those raw strings and in the future, we could add stages without modifying all the methods.
This could also imply modifying how we are treating m_chack and m_stages Expressions.

LucioDonda (commented):
Review still in progress

src/engine/source/_builder/expression.hpp
```diff
@@ -0,0 +1,447 @@
...
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+#include <fmt/format.h>
+
+// namespace builder::internals
+// {
+
+// The Formula hierarchy is a description of the graph expression to be built,
```
LucioDonda, line 17: By separating the description from the expression you are able to use filters, map, and subscription, (the way it's done inside rxcppFactory) right?

Issue #13389, opened 2022-05-06
Review and update Gcloud unit tests


CarlosRS9:
| Affected integration |
|---|
| GCloud |

## Description
As part of #13387 we need to review the existing unit tests for the Gcloud module, expanding them to ensure we are close to the 100% coverage and adding new test if required. This is our current coverage:

```
Name                               Stmts   Miss  Cover
------------------------------------------------------
gcloud/buckets/access_logs.py         16      5    69%
gcloud/buckets/bucket.py             130     82    37%
gcloud/exceptions.py                  23      0   100%
gcloud/gcloud.py                      71     29    59%
gcloud/integration.py                 36      3    92%
gcloud/pubsub/subscriber.py           60     32    47%
gcloud/tests/conftest.py              25      0   100%
gcloud/tests/test_access_logs.py      20      0   100%
gcloud/tests/test_bucket.py           19      0   100%
gcloud/tests/test_gcloud.py           14      0   100%
gcloud/tests/test_integration.py      38      0   100%
gcloud/tests/test_subscriber.py       18      0   100%
gcloud/tools.py                       48     24    50%
utils.py                              57     27    53%
------------------------------------------------------
TOTAL                                575    202    65%
```

## Tasks
- [x] Expand the existing tests to raise our coverage.
- [x] Add new tests, if necessary.
- [x] Add entry to the changelog if necessary.

CarlosRS9:
# Status update

We have finished reviewing and improving the unit tests for the Gcloud module. Here are some details about the actions taken:

- In addition to the creation of new unit tests to address functions and code snippets that were not being taken into account, it has also been necessary to replace the existing tests. 

- A new set of tests have been developed, as well as some auxiliary functions for their correct performance.

- Some modules and functions have been deprecated or updated.

- Several improvements and modifications have been made to the GCP module code. Some unnecessary code fragments have also been removed.


On the other hand, it has not been possible to create a `tools.py` auxiliary module for the unit tests due to the problem of the wodle folder not being an installable python package. Once https://github.com/wazuh/wazuh/issues/12874 is addressed, it will be possible to do so, further optimising the design of these unit tests.

Because we have changed the code, we are currenlty performing manual tests to ensure the correct functioning of the module.

CarlosRS9:
# Pub/sub manual testing

No data must be modified or removed unless it was specified for the test. 

The tests assume there is a valid authentication file located at `/var/ossec/wodles/gcloud/credentials.json`.

[...]

PR #13420, opened 2022-05-11
Improve external integrations SQLite queries

### LINKED ISSUE #10881: Make external integrations SQLite queries use the DB API's parameter substitution
|Component|
|---|
| External integration modules |

## Description  
The SQLite queries in the AWS and Google Cloud modules use the Python string operations to create the queries, as can be seen in the following code snippet:
https://github.com/wazuh/wazuh/blob/285a5df64d92e0b52e5ddf67935e33e2bf306263/wodles/aws/aws_s3.py#L551-L557  

The recommended parameter substitution method for SQLite queries is to use the DB API, as stated in the [documentation](https://docs.python.org/3/library/sqlite3.html); the use of Python's string operations is totally discouraged and should be fixed.

CarlosRS9:
|Related issue|
|---|
| #10881 |

## Description

This PR closes #10881. The queries available in the `AWS` and `GCP` external integration modules has been enhanced by using parameter substitution method as stated in the [documentation](https://docs.python.org/3/library/sqlite3.html) instead of using Python's string operations is totally discouraged and should be fixed. Azure module is not affected as it uses its own orm module.

## Tests

Due to the major impact of these changes, especially on the `AWS` module, a thorough manual testing has been carried out. The results and the details can be found in #10881.


wodles/aws/aws_s3.py
```diff
@@ -574,44 +568,21 @@ def _get_last_key_processed(self, aws_account_id: str) -> str or None:
...
-
     def already_processed(self, downloaded_file, aws_account_id, aws_region):
-        cursor = self.db_connector.execute(self.sql_already_processed.format(
-            bucket_path=self.bucket_path,
-            table_name=self.db_table_name,
-            aws_account_id=aws_account_id,
-            aws_region=aws_region,
-            log_name=downloaded_file
-        ))
+        cursor = self.db_connector.execute(self.sql_already_processed.format(table_name=self.db_table_name), {
```
Selutario, line 581: This `self.db_table_name` attribute is defined in the child classes but not in this one, which causes warnings. Since some methods use it, it might be useful to define it in `__init__` of this parent class as well, as `None`:
```
self.db_table_name = None
```

wodles/aws/aws_s3.py
```diff
@@ -2159,18 +2098,16 @@ def list_paths_from_dict(d, discard_levels=None, glue=".", path=None):
 
     def iter_regions_and_accounts(self, account_id, regions):
         # Only <self.retain_db_records> logs for each region are stored in DB. Using self.bucket as region name
-        # would prevent to lose lots of logs from different buckets.
+        # would prevent to loose lots of logs from different buckets.
```
Selutario, line 2101:
```suggestion
        # would prevent to lose lots of logs from different buckets.
```

wodles/aws/aws_s3.py
CarlosRS9: This is already being fixed in #13389 which is where we are upgrading, improving and fixing gcp's Unit tests as well as applying any required change related to those tests, such as this one

Selutario (approved):

davidjiglesias (approved):

Issue #13768, opened 2022-06-07
Modulesd prints that agents have no OS information available


vikman90:
|Wazuh version|Component|Install type|
|---|---|---|
| 4.3.3 | Vulnerability Detector | Manager |

Vulnerability detector sometimes prints this log:

```
wazuh-modulesd INFO: Agents DB (8): No OS information available.
```

## Rationale

Vulnerability Detector needs to collect each agent's OS data in order to select its appropriate feed. Such data may not be available, due to:

- Syscollector (OS data at less) is disabled on that agent.
- The agent OS data table has not been synced to the manager yet.

This is normally a timing issue, that's why the message is an information log instead of a warning or an error. However, the log comes from a helper function, we expect that Vuln Detector prints it. In addition, the message should be more descriptive.

## Proposed changes

- Move the log from the helper function directly to Vulnerability Detector.
- Improve the log description.

PR #13915, opened 2022-06-20
Improve OS info log unavailable for Vulnerability Detector

MarcelKemp:
|Related issue|
|---|
|#13768|

## Description

In this PR, the changes proposed in the issue description #13768 have been applied to make it easier to recognize the problem and to provide a better context.

## Configuration options

In the configuration, to cause the issue and show the log, it is only necessary to **enable** _Vulnerability Detector_ in the manager and **disable** _Syscollector_ in the agent.

This way, the manager will not have the necessary information from the agent and will display the log message.

- Agent:
```xml
  <!-- System inventory -->
  <wodle name="syscollector">
    <disabled>yes</disabled>
    <interval>1h</interval>
    <hardware>yes</hardware>
    <os>yes</os>
    <network>yes</network>
    <packages>yes</packages>
    <ports all="no">yes</ports>
    <processes>yes</processes>

    <!-- Database synchronization settings -->
    <synchronization>
      <max_eps>10</max_eps>
    </synchronization>
  </wodle>
```

## Logs/Alerts example

Previously, the log was displayed as follows:
```
wazuh-modulesd INFO: Agents DB (1): No OS information available.
```

Whereas now, the new log would appear with exactly the same problem, but showing the following message:
```
wazuh-modulesd:vulnerability-detector: INFO: (5700): Unable to get the OS information for agent '001'. It may not have the OS inventory enabled.
```

## Tests

  - [x] Linux
- [x] Source installation
- [x] Package installation
- [x] Source upgrade
- [x] Package upgrade

- Memory tests for Linux
  - [x] Scan-build report
  - [ ] Coverity
  - [ ] AddressSanitizer

- [x] Retrocompatibility with older Wazuh versions

### DIFF
--- a/src/error_messages/information_messages.h
+++ b/src/error_messages/information_messages.h
@@ -12,7 +12,7 @@
 #define INFO_MESSAGES_H
 
 /* Vulnerability-Detector info messages*/
-#define VU_NO_OSINFO_AG       "(5700): Unable to get the OS information for agent '%.3d'. It may not have the OS inventory enabled."
+#define VU_NO_OSINFO_AG       "(5700): Unable to get the OS information for agent '%.3d'. Inventory data may not yet be synchronized."
 
 /* File integrity monitoring info messages*/
 #define FIM_DAEMON_STARTED                  "(6000): Starting daemon..."
--- a/src/error_messages/information_messages.h
+++ b/src/error_messages/information_messages.h
@@ -11,6 +11,9 @@
 #ifndef INFO_MESSAGES_H
 #define INFO_MESSAGES_H
 
+/* Vulnerability-Detector info messages*/
+#define VU_NO_OSINFO_AG       "(5700): Unable to get the OS information for agent '%.3d'. It may not have the OS inventory enabled."
+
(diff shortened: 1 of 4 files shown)


src/error_messages/information_messages.h
```diff
@@ -11,6 +11,9 @@
 #ifndef INFO_MESSAGES_H
 #define INFO_MESSAGES_H
 
+/* Vulnerability-Detector info messages*/
+#define VU_NO_OSINFO_AG       "(5700): Unable to get the OS information for agent '%.3d'. It may not have the OS inventory enabled."
```
chemamartinez, line 15: Regarding the issue comment, the two causes that may cause this log to be printed are:

- Syscollector (OS data at less) is disabled on that agent.
- The agent OS data table has not been synced to the manager yet.

The second one will be the cause in most cases so I think it would be better to mention it instead of the possibility of having the inventory disabled (which is enabled by default).

I suggest something like:
```
#define VU_NO_OSINFO_AG "(5700): Unable to get the OS information for agent '%.3d'. Inventory data may not yet be synchronized."
```

PR #14066, opened 2022-06-28
Algorithm in charge of checking the conditions prior to agent reconnection

### LINKED ISSUE #13794: Define and develop the algorithm in charge of verifying if the cluster is unbalanced
In this issue we will develop the algorithm in charge of verifying if the cluster is unbalanced. For this we have to define the following:
- Criteria to determine the unbalance.
- Tolerance to the unbalance.
- Capacity to be able to specify a black list of agents so that these are not taken into account and are not balanced in any case (It will be taken into account in the global computation because the least conneciton algorithm has no knowledge of these agents).
- Define a stop condition in case the balancing fail.
- Capacity to check if the agents remain in the same worker iteration after iteration, in case they remain in the same worker or at least 10% of the agents do not reconnect, the process will be stopped.
- In case of verifying an unbalance several times in a row, it will indicate that there is a problem and the process will stop.

Initially we have proposed that the algorithm should be somewhat simple, take the number of worker nodes and the number of agents, divide the latter by the number of workers and the result will be multiplied by a tolerance. Any node that has more of these agents connected will need to be rebalanced.
```
n_agents // n_workers * (1 + tolerance)
```

image

Therefore, the initial design can be modified as long as it is justified and improves on the design initially proposed.

AdriiiPRodri:
|Related issue|
|---|
|#13794|

## Description

This PR closes #13794. In this PR we have added a phase which contains two procedures. The first one checks that the previous agent reconnection was successful, the second one checks the current state of the cluster.

### DIFF
--- a/framework/wazuh/core/cluster/agents_reconnect.py
+++ b/framework/wazuh/core/cluster/agents_reconnect.py
@@ -1,14 +1,19 @@
 import contextlib
+
+from collections import defaultdict
+from datetime import timedelta
 from enum import Enum
+from typing import overload
 from xmlrpc.client import Boolean
 
 from wazuh.core import utils
+from wazuh.core.cluster import control, local_client
 from wazuh.core.common import DECIMALS_DATE_FORMAT
 
 
 class AgentsReconnectionPhases(str, Enum):
     NOT_STARTED = "Not started"
-    CHECK_WORKERS_STABILITY = "Check workers stability"
+    CHECK_NODES_STABILITY = "Check nodes stability"
     CHECK_PREVIOUS_RECONNECTIONS = "Check previous reconnections"
     CHECK_AGENTS_BALANCE = "Check agents balance"
     RECONNECT_AGENTS = "Reconnect agents"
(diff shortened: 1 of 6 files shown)


framework/wazuh/core/cluster/agents_reconnect.py
```diff
@@ -30,107 +36,284 @@ def __init__(self, logger, nodes, blacklisted_nodes, workers_stability_threshold
...
+        """Get all agents in the system flagged as active.
+
+        Returns
+        -------
+        dict
+            Dictionary with agents information.
+        """
+        lc = local_client.LocalClient()
+        return await control.get_agents(lc, filter_status=["active"],
+                                        select_fields={'id', 'node_name', 'lastKeepAlive'})
```
vicferpoy, line 133: I suggest using the `forward_function` function in combination with the `get_agents` function from the agent SDK.

vicferpoy (changes requested):

vicferpoy (approved):
LGTM

Issue #14108, opened 2022-07-01
AES decryption doesn't put termination character on decrypted buffer


sebasfalcone:
## Description
During the UT development for the `/src/os_crypto/aes/aes_os.c` module, I've found that the function `decrypt_aes` could lead to unexpected results.

The problem is that the output buffer is not properly trimmed before exiting the function. So, if it is not done afterwards, you could read garbage from the buffer. This wasn't an issue because the only time the `aes_encrypt` function is called is here:
https://github.com/wazuh/wazuh/blob/2d9d7d9f47bb0e634729f088b22abe22e14f4306/src/os_crypto/shared/msgs.c#L334-L335

And after that, the output buffer is properly finished:
https://github.com/wazuh/wazuh/blob/2d9d7d9f47bb0e634729f088b22abe22e14f4306/src/os_crypto/shared/msgs.c#L347

## How to reproduce
In the following UT we can reproduce an unexpected situation:
```
int test_AES()
{
    const char *initialString = "teststring";

    uchar *iv  = (uchar *)"FEDCBA0987654321";
    uchar *key = (uchar *)"1234567890ABCDEF";
    
    const int buffersize = 1024;
    char encryptedBufffer[buffersize];
    char decryptedBufffer[buffersize];

    int encryptedSize = encrypt_AES((uchar*)initialString, (int)strlen(initialString), key, iv, (uchar*)encryptedBufffer);
    
    // AES output string is non readable, we need to decrypt for the assert
    int decryptedSize = decrypt_AES((uchar*)encryptedBufffer, encryptedSize, key, iv, (uchar*)decryptedBufffer);
    
    assert_string_equal(decryptedBufffer, initialString);
    
     // AES decryption fails to put escape charater on the output string
    decryptedBufffer[decryptedSize] = '\0';
    assert_string_equal(decryptedBufffer, initialString);
}

[==========] tests: Running 1 test(s).
[ RUN      ] test_AES
[  ERROR   ] --- "teststring" != "teststring"
[   LINE   ] --- /home/sebas/Documents/wazuh/src/unit_tests/os_crypto/aes/test_aes_op.c:37: error: Failure!
[  FAILED  ] test_AES
[==========] tests: 1 test(s) run.
[  PASSED  ] 0 test(s).
[  FAILED  ] tests: 1 test(s), listed below:
[  FAILED  ] test_AES
```

## Solution

Modify the `aes_decrypt` function to the following:
```
[...]

PR #14124, opened 2022-07-04
Fix: add termination character to AES decrypted strings

sebasfalcone:
|Related issue|
|---|
|[Closing #14108](https://github.com/wazuh/wazuh/issues/14108)|

## Description

- This PR originally indented to add an unit test for the AES module. Now it also fixes bad trimming on AES decrypted strings.

## Tests
  - [x] Linux
  - [ ]  Windows
  - [ ]  MAC OS X


src/os_crypto/aes/aes_op.c
```diff
@@ -95,7 +95,7 @@ int decrypt_AES(const unsigned char *ciphertext, int ciphertext_len, unsigned ch
     }
 
 	plaintext_len += len;
-
+    plaintext[plaintext_len] = '\0';
```
pereyra-m, line 98: This is already managed in `ReadSecMSG()` (` cleartext[buffer_size] = '\0';`).

Instead of modifying the method just to test it, we could check the test result in another way (**strncmp()** for example).

src/unit_tests/os_crypto/aes/test_aes_op.c
```diff
@@ -0,0 +1,46 @@
...
+    // AES output string is non readable, we need to decrypt for the assert
+    int decryptedSize = decrypt_AES((uchar*)encryptedBufffer, encryptedSize, key, iv, (uchar*)decryptedBufffer);
+    
+    assert_string_equal(decryptedBufffer, initialString);
+}
+
+int main(void) 
+{
+    const struct CMUnitTest tests[] = {
+        cmocka_unit_test(test_AES),
```
pereyra-m, line 43: Could we add more tests to increase coverage?

For example: NULL arguments, non-happy cases, isolate the encrypt method for the decrypt method, etc.

sebasfalcone: Good idea, I will cover this in more depth

src/os_crypto/aes/aes_op.c
HanesSciarrone: This function never validates whether the plaintext variable is null or not. We should check this at the beginning. If we don't check a null pointer could produce a `SEGV`

PR #14131, opened 2022-07-05
Add the new configuration to the ossec.conf file

### LINKED ISSUE #13802: Add the new configuration to the ossec.conf file
In this issue we will modify the code created in the previous issues so that the values are extracted from the `ossec.conf` file. We have pending to extend the requirements of this development to modify the `ossec.conf` file but initially we have several parameters secured:

- Option to enable or disable the load balancing system.
- Maximum wait value to determine that the environment is ready (counter).
- Maximum value of failed rebalancing attempts before stopping the process.
- Maximum value of agents that have not reconnected before stopping the process.
- Tolerance when evaluating the agents in each node.

We should decide if this will go in the `<cluster>...</cluster>` block or in a new block.

AdriiiPRodri:
|Related issue|
|---|
|#13802|

## Description

This PR closes #13802. In this PR we have implemented the necessary changes to add the new configuration for the agents reconnect functionality. This new configuration is as follows:

```
<agent_reconnection>
    <disabled>yes</disabled>
    <node_blacklist></node_blacklist>
  </agent_reconnection>
```

### DIFF
--- a/src/config/cluster-config.c
+++ b/src/config/cluster-config.c
@@ -64,12 +64,6 @@ int Read_Cluster(XML_NODE node, void *d1, __attribute__((unused)) void *d2) {
                 return OS_INVALID;
             }
             os_strdup(node[i]->content, Config->node_name);
-        } else if (!strcmp(node[i]->element, node_blacklist)) {
-            if (strspn(node[i]->content, N_BLACKLIST) < strlen(node[i]->content)) {
-                merror("Detected a not allowed character in node blacklist: \"%s\". Characters allowed: \"%s\".", node[i]->content, N_BLACKLIST);
-                return OS_INVALID;
-            }
-            // os_strdup(node[i]->content, Config->node_blacklist);
         } else if (!strcmp(node[i]->element, node_type)) {
             if (!strlen(node[i]->content)) {
                 merror("Node type is empty in configuration");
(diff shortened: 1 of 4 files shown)


src/config/cluster-config.c
```diff
@@ -29,7 +31,7 @@ int Read_Cluster(XML_NODE node, void *d1, __attribute__((unused)) void *d2) {
     static const char *port = "port";
     static const char *bind_addr = "bind_addr";
     static const char *C_VALID = "!\"#$%&'-.0123456789:<=>?ABCDEFGHIJKLMNOPQRESTUVWXYZ[\\]^_abcdefghijklmnopqrstuvwxyz{|}~";
-
+    static const char *N_BLACKLIST = "!\"#$%&'-.,0123456789:<=>?ABCDEFGHIJKLMNOPQRESTUVWXYZ[\\]^_abcdefghijklmnopqrstuvwxyz{|}~";
```
Selutario, line 34: This file is actually used to limit which characters are used in options read by other services (for example, `node_name` being inserted into `global.db`). 

Since no service other than `wazuh-clusterd` uses the Agents reconnection info, I think there is no need to create a new list of characters. In fact, the validation should be done (if it is necessary) in the `wazuh-clusterd` process, since it can be started manually and the validation of this file would not be applied.

AdriiiPRodri: We have created a list of characters since concatenating the string `C_VALID` with the `,` character was more expensive, also, we do this so that analysisd raises an exception in case it finds an error in the configuration:
```
2022/07/05 15:07:53 wazuh-db[13264] debug_op.c:70 at _log(): DEBUG: Logging module auto-initialized
2022/07/05 15:07:53 wazuh-db[13264] main.c:111 at main(): DEBUG: Wazuh home directory: /var/ossec
2022/07/05 15:07:53 wazuh-analysisd: ERROR: Detected a not allowed character in node name: "�master-node". Characters allowed: "!"#$%&'-.0123456789:<=>?ABCDEFGHIJKLMNOPQRESTUVWXYZ[\]^_abcdefghijklmnopqrstuvwxyz{|}~".
2022/07/05 15:07:53 wazuh-analysisd: ERROR: (1202): Configuration error at 'etc/ossec.conf'.
2022/07/05 15:07:53 wazuh-analysisd: CRITICAL: (1202): Configuration error at 'etc/ossec.conf'.
wazuh-analysisd: Configuration error. Exiting
```

PR #14186, opened 2022-07-08
Update external integrations github template

CarlosRS9:
## Description
This PR updates the External modules integrations github template to indicate that unit tests should be run specifically with python 3.6 to ensure compatibility.


fdalmaup:
The PR is outdated since the minimum accepted Python version for the latest 4.x versions is 3.8. Also, the external integrations will be reworked for 5.0 and no modification is expected in the Wazuh repository.

PR #14243, opened 2022-07-14
Third-party vendor vulnerabilities content config file

### LINKED ISSUE #14154: Migration from third-party vulnerability content to Wazuh Content

jnasselle:
|Related issue|
|---|
|Close wazuh/wazuh-content#4|

## Description

This PR aims to create third-party vendor feeds description files, that will be used as configuration files and ingested by the Migration tool

- [x]  Ubuntu
- [x] SUSE Linux Enterprise
- [x] Debian
- [x] RedHat
- [x] RedHat JSON
- [x] NVD
- [x] Amazon Linux
- [ ] Microsoft


src/content/third_party_migration/vendors/redhat-json.json
```diff
@@ -0,0 +1,30 @@
...
+    "remotes": [
+        {
+            "url": "https://access.redhat.com/labs/securitydataapi/cve.json",
+            "request": "api",
+            "type": "json",
+            "parameters": {
+                "after": {
+                    "type": "fixed",
+                    "description": "Year",
+                    "value": "1999"
```
sebasfalcone, line 13: - after parameter expects ISO 8601 format -> YYYY-MM-DD or YYYYMMDD

src/content/third_party_migration/vendors/redhat-json.json
```diff
@@ -0,0 +1,30 @@
...
+                },
+                "per_page": {
+                    "type": "fixed",
+                    "description": "Vulns per page",
+                    "value": "1000"
+                },
+                "page": {
+                    "type": "variable-incremental",
+                    "value": [
+                        "0",
```
sebasfalcone, line 23: - page parameter starts from 1 (this should be contemplated on the schema?)

src/content/third_party_migration/vendors/redhat-json.json
sebasfalcone: Set just year on configurations and do the work on the module? I think this will bring lots of if-else if-else statements on the fetcher code

src/content/third_party_migration/vendors/suse.json
```diff
@@ -0,0 +1,30 @@
...
+    "remotes": [
+        {
+            "url": "https://ftp.suse.com/pub/projects/security/oval/suse.linux.enterprise.{type}.{version}.xml",
+            "request": "file",
+            "type": "xml",
+            "parameters": {
+                "version": {
+                    "type": "fixed",
+                    "description": "SLE version number",
+                    "value": [
```
jftuduri, line 13: The values for version and type are mixed up.

src/content/third_party_migration/vendors/suse.json
```diff
@@ -0,0 +1,30 @@
...
+                        "desktop",
+                        "enterprise"
```
jftuduri, line 15: should be "server" instead of "enterprise".

src/content/third_party_migration/vendors/redhat-json.json
jnasselle: Nice catch! working on it

src/content/third_party_migration/vendors/redhat-json.json
```diff
@@ -0,0 +1,30 @@
...
+    "remotes": [
+        {
+            "url": "https://access.redhat.com/labs/securitydataapi/cve.json?after={after}&per_page={per_page}&page={page}",
+            "request": "api",
+            "type": "json",
+            "parameters": {
+                "after": {
+                    "type": "fixed",
+                    "description": "Date in ISO 8601 format (YYYY-MM-DD or YYYYMMDD)",
+                    "value": "1999-01-01"
```
sebasfalcone, line 13: - To make the schema simpler change `value` from string to array:
    - `"value":["1999-01-01"]`
- Change the type name from `fixed` to `list` as is more descriptive of what it is.

src/content/third_party_migration/vendors/redhat-json.json
```diff
@@ -0,0 +1,30 @@
...
+                "per_page": {
+                    "type": "fixed",
+                    "description": "Vulnerabilities per page",
+                    "value": "1000"
+                },
+                "page": {
+                    "type": "variable-incremental",
+                    "value": [
+                        "1",
+                        "inf"
```
sebasfalcone, line 24: - To make the schema  and the logic on the fetchers simpler:
    - Change `value` from array to object.
    - Add the following fields:
        - Start: String that contains solely numbers
        - End: String that contains solely numbers or keyword inf
        - Step (optional): String that contains solely numbers

src/content/third_party_migration/vendors/nvd.json
```diff
@@ -0,0 +1,21 @@
...
+            "url": "https://nvd.nist.gov/feeds/json/cve/1.1/nvdcve-1.1-{year}.json.gz",
+            "request": "file",
+            "type": "json",
+            "compressed": "gzip",
+            "parameters": {
+                "year": {
+                    "type": "variable-incremental",
+                    "value": [
+                        "2010",
+                        "currentyear"
```
sebasfalcone, line 15: - With the objective to make the schema more solid and make the fetcher logic simpler:
    - The same considerations as [here](https://github.com/wazuh/wazuh/pull/14243/files#r928006398), but keyword can be `currentYear`.
    - Rename the type to `variable-incremental-year`

src/content/third_party_migration/vendors/debian.json
```diff
@@ -0,0 +1,22 @@
+{
+    "vendor": "debian",
+    "description": "Debian",
+    "remotes": [
+        {
+            "url": "https://www.debian.org/security/oval/oval-definitions-{version}.xml",
+            "request": "file",
+            "type": "xml",
```
sebasfalcone, line 8: - JSON schema uses the keyword type (a lot). To make the schema easier to develop, debug and read; we should use this keyword the as little as possible.
   - We suggest to change this to `dataFormat` (it also makes this field more descriptive).

sebasfalcone (changes requested):
Questions to discuss:
- Should we add the field compressed to every configuration? 
    - Yes: When no compression is present specify it with `none`.
    -  No: Assume default case to be `none`.

Issue #14228, opened 2022-07-13
Update framework functions that communicate with `queue/sockets/request` to use `queue/sockets/remote` instead


mcarmona99:
This issue is part of https://github.com/wazuh/wazuh/issues/13099.

The `queue/sockets/request` socket has been unified with `queue/sockets/remote`. From now on, `queue/sockets/remote` is the only socket used for communications with remote.

We should modify all the framework functions using the `request` socket.

We also need to take into account that the remote socket receive messages following the **JSON communication protocol** we defined.

### wazuh/wazuh
- Unit tests without failures. Updated and/or expanded if there are new functions/methods/outputs:
  - [x] Cluster (`framework/wazuh/core/cluster/tests/` & `framework/wazuh/core/cluster/dapi/tests/`)
  - [x] Core (`framework/wazuh/core/tests/`)
  - [x] SDK (`framework/wazuh/tests/`)
  - [x] RBAC (`framework/wazuh/rbac/tests/`)
  - [x] API (`api/api/tests/`)
- API tavern integration tests without failures. Updated and/or expanded if needed (`api/test/integration/`):
  - [x] Affected tests 
  - [x] Affected RBAC (black and white) tests
- [x] Review integration test mapping using the script (`api/test/integration/mapping/integration_test_api_endpoints.json`)
- [x] Review of spec.yaml examples and schemas (`api/api/spec/spec.yaml`)
- [x] Review exceptions remediation when any endpoint path changes or is removed (`framework/wazuh/core/exception.py`)
- [x] Changelog (`CHANGELOG.md`)

### wazuh/wazuh-documentation
- [x] Migration from 3.X for changed endpoints (`source/user-manual/api/equivalence.rst`)
- [x] Update RBAC reference with new/modified actions/resources/relationships (`source/user-manual/api/rbac/reference.rst`)

Issue #13929, opened 2022-06-21
Update framework function to get active configuration when component is `analysis`


mcarmona99:
## Description

This issue is part of the epic https://github.com/wazuh/wazuh/issues/13099.

We need to update the `get_active_configuration` function of `wazuh/core/configuration.py` as from now on, the communication with the `analysis` socket will be done following the JSON standard defined in https://github.com/wazuh/wazuh/issues/5934.

https://github.com/wazuh/wazuh/blob/b5ff55eaf450f69b8e4fe9b0c426e19fc7056207/framework/wazuh/core/configuration.py#L768-L857

As we can see, when the component is `analysis`, the message built and sent to the socket is `command = f"getconfig {configuration}"`. As we said this is not correct due to the new development done in the related epic issue.

New message example: `{'version': 1, 'origin': {'module': 'api'}, 'command': 'getconfig', 'parameters': {'section': 'global'}}`

Response example: `{'error': 0, 'data': {…}, 'message': 'ok'}`

We should update this function and consider dividing it into different functions, as `get_active_configuration` is used for both agents and managers, by using an if/else statement, which can be improved.

The endpoints affected are, at least:

`/agents/{agent_id}/config/{component}/{configuration}`
`/cluster/{node_id}/configuration/{component}/{configuration}`
`/manager/configuration/{component}/{configuration}`

### wazuh/wazuh
- Unit tests without failures. Updated and/or expanded if there are new functions/methods/outputs:
  - [x] Cluster (`framework/wazuh/core/cluster/tests/` & `framework/wazuh/core/cluster/dapi/tests/`)
  - [x] Core (`framework/wazuh/core/tests/`)
  - [x] SDK (`framework/wazuh/tests/`)
  - [x] RBAC (`framework/wazuh/rbac/tests/`)
  - [x] API (`api/api/tests/`)
- API tavern integration tests without failures. Updated and/or expanded if needed (`api/test/integration/`):
  - [x] Affected tests 
  - [x] Affected RBAC (black and white) tests
- [x] Review integration test mapping using the script (`api/test/integration/mapping/integration_test_api_endpoints.json`)
- [x] Review of spec.yaml examples and schemas (`api/api/spec/spec.yaml`)
- [x] Review exceptions remediation when any endpoint path changes or is removed (`framework/wazuh/core/exception.py`)
- [x] Changelog (`CHANGELOG.md`)

### wazuh/wazuh-documentation
[...]

PR #14259, opened 2022-07-15
Use the `request` socket instead of `remote` in framework functions

mcarmona99:
|Related issue|
|---|
| #14228  |

This PR closes https://github.com/wazuh/wazuh/issues/14228.

In this pull request, I have updated the framework functions that use the socket queue/sockets/request to use queue/sockets/remote instead.

The most affected function is get_active_configuration of wazuh/core/configuration.py. In this function, changes were needed to use remote when getting the agents' active configuration and to use the remote socket when requesting the request configuration of 000.

The get_daemons_stats_from_socket function of wazuh/core/stats.py was also updated.

I have updated all the related unit tests too.

The endpoints affected are working properly but the API integration tests will still fail due to https://github.com/wazuh/wazuh/issues/13929.


framework/wazuh/core/configuration.py
```diff
@@ -32,7 +32,7 @@
 #   entry.
 #   * Last -> there can be multiple sections in the configuration but only the last one will be returned.
 #   The rest are ignored.
-conf_sections = {
+CONF_SECTIONS = {
```
vicferpoy, line 35: Upper case words are used for "constants". However, python dictionaries are not. If you really want to define this dictionary as constant, consider using `MappingProxyType`. If it needs to be mutable, undo this change.

framework/wazuh/core/configuration.py
```diff
@@ -798,62 +795,129 @@ def get_active_configuration(agent_id, component, configuration):
...
-                  "monitor", "request", "syscheck", "wmodules"}
-
-    # checks if the component is correct
+    # Check if the component is correct
+    components = component_socket_mapping.keys()
     if component not in components:
         raise WazuhError(1101, f'Valid components: {", ".join(components)}')
 
-    sockets_path = os_path.join(common.WAZUH_PATH, "queue", "sockets")
+    def get_active_configuration_000():
```
vicferpoy, line 818: Is the `_000` necessary? Look for name alternatives.

framework/wazuh/core/configuration.py
```diff
@@ -798,62 +795,129 @@ def get_active_configuration(agent_id, component, configuration):
...
-        command = f"getconfig {configuration}"
         # Verify component configuration
         if not os.path.exists(dest_socket):
-            raise WazuhError(1121, extra_message=f"please verify that the component '{component}' \
-                is properly configured")
-    else:
-        dest_socket = os_path.join(sockets_path, "request")
-        command = f"{str(agent_id).zfill(3)} {component} getconfig {configuration}"
+            raise WazuhError(1121,
+                             extra_message=f"please verify that the component '{component}' is properly configured")
```
vicferpoy, line 826:
```suggestion
                             extra_message=f"Please verify that the component '{component}' is properly configured")
```

framework/wazuh/core/configuration.py
```diff
@@ -798,62 +795,129 @@ def get_active_configuration(agent_id, component, configuration):
...
+            s.send(msg.encode())
+
+            # Receive response
+            try:
+                # Receive data length
+                rec_msg_ok, rec_msg = s.receive().decode().split(" ", 1)
+            except ValueError:
+                raise WazuhInternalError(1118, extra_message="Data could not be received")
+
+            s.close()
```
vicferpoy, line 848: Either call the `close()` method in a `finally` block or use the context manager to guarantee that the socket will close.

framework/wazuh/core/configuration.py
```diff
@@ -798,62 +795,129 @@ def get_active_configuration(agent_id, component, configuration):
...
+            # Send message
+            s.send(msg)
+
+            # Receive response
+            try:
+                response = s.receive(raw=True)
+            except ValueError:
+                raise WazuhInternalError(1118, extra_message="Data could not be received")
+
+            s.close()
```
vicferpoy, line 873: Same as above.

framework/wazuh/core/configuration.py
```diff
@@ -798,62 +795,129 @@ def get_active_configuration(agent_id, component, configuration):
...
+        s.send(msg.encode())
+
+        # Receive response
+        try:
+            # Receive data length
+            rec_msg_ok, rec_msg = s.receive().decode().split(" ", 1)
+        except ValueError:
+            raise WazuhInternalError(1118, extra_message="Data could not be received")
+
+        s.close()
```
vicferpoy, line 901: Same as above.

framework/wazuh/core/configuration.py
```diff
@@ -124,7 +126,7 @@ def _insert(json_dst, section_name, option, value):
         else:
             json_dst[option] = value  # Update values
     else:
-        if section_name in conf_sections and option in conf_sections[section_name]['list_options']:
+        if section_name in CONF_SECTIONS and option in CONF_SECTIONS[section_name]['list_options']:
```
vicferpoy, line 129: This could be reduced using a double `dict.get()`.

framework/wazuh/core/stats.py
```diff
@@ -178,7 +178,7 @@ def get_daemons_stats_from_socket(agent_id, daemon):
         dest_socket = os.path.join(sockets_path, daemon)
         command = "getstate"
     else:
-        dest_socket = os.path.join(sockets_path, "request")
+        dest_socket = os.path.join(sockets_path, "remote")
```
vicferpoy, line 181: Don't we have a constant with this exact socket path?

vicferpoy (changes requested):

framework/wazuh/core/configuration.py
mcarmona99: It does not need to be mutable, therefore, I will use MappingProxyType.

framework/wazuh/core/configuration.py
mcarmona99: Changed to `get_active_configuration_manager`

framework/wazuh/core/configuration.py
mcarmona99: `finally` added

framework/wazuh/core/stats.py
mcarmona99: We do, including this change too.

davidjiglesias (approved):

Issue #13524, opened 2022-05-17
Engine - Decode FIM events


juliancnn:
| Wazuh version | Component | Install type | Install method   | Platform   |
| ------------- | --------- | ------------ | ---------------- | ---------- |
| 5.0.0         | Engine    | Manager      | any | Linux |

# Description

As part of epic #13038, it is necessary to decode the FIM events.

### Syscheck/FIM
**Queue: `'8'` / `0x38` / `SYSCHECK_MQ`**

In this case, it is identified by the id `'8'` (SYSCHECK_MQ) and will push to the rules queue after being decoded. 
If the agent has 3.1.1 version or older, the event received will be in .log format and handled by "DecodeSyscheck". If not it will be taken by "decode_fim_event".
The format is used to determine the event version, being the later FIM messages on .json.

#### New F.I.M. module:
This is the only one that has a well-documented event format.
Functions used by the wdb module:

- `Fim_db_search`: interactive query mode, it can use previous query results data on new ones, e.g. by using fim_update_date.
- `Fim_get_scantime`: Searchs for scan-time from db.
- `Fim_control_msg`: Query without using the returned value (e.g. delete). 
- `Fim_update_date`: DITTO
- `Fim_database_clean`: DITTO
- `Wdbc_query_ex`: DITTO

<details>
<summary>Message format CLICK ME</summary>

```json
{
   "type":                  "event"
   "data": {
     "path":                string
     "hard_links":          array
     "mode":                "scheduled"|"realtime"|"whodata"
     "type":                "added"|"deleted"|"modified"
     "timestamp":           number
     "changed_attributes": [
       "size"
       "permission"
       "uid"
       "user_name"
       "gid"
       "group_name"
       "mtime"
       "inode"
       "md5"
       "sha1"
       "sha256"
     ]
     "tags":                string
     "content_changes":     string
     "old_attributes": {
       "type":              "file"|"registry"
       "size":              number
       "perm":              string
       "user_name":         string
       "group_name":        string
       "uid":               string
       "gid":               string
       "inode":             number
       "mtime":             number
       "hash_md5":          string
       "hash_sha1":         string
       "hash_sha256":       string
       "win_attributes":    string
       "symlink_path":      string
       "checksum":          string
     }
     "attributes": {
       "type":              "file"|"registry"
       "size":              number
       "perm":              string
       "user_name":         string
       "group_name":        string
       "uid":               string
       "gid":               string
       "inode":             number
       "mtime":             number
[...]

Issue #13523, opened 2022-05-17
Engine - Decode CIS-CAT events


nmkoremblum:
| Wazuh version | Component | Install type | Install method   | Platform   |
| ------------- | --------- | ------------ | ---------------- | ---------- |
| 5.0.0         | Engine    | Manager      | any | Linux |

# Description

As part of epic #13038, it is necessary to decode the CIS-CAT events.

### CIS-CAT

**Queue: `'e'` / `0x65` / `CISCAT_MQ`**

At the moment, the method DecodeCiscat() is in charge of decoding CIS-CAT events, identified by the decoder ID 0x65 (CISCAT_MQ), it shares the same thread as Syslog.

CIS-CAT json logs are first parsed, if that results ok and the event contains a non-null `/type` field then it will be delivered to the rules matching stage without previously passing through decoders. When field `/type` matches `scan_info` it will query that scan to wazuhDB.

The DB query format has the following format:

```
agent <agent_id> ciscat save PARAM1 | PARAM2 | PARAM3 ….
```

Issue #13522, opened 2022-05-17
Engine - Decode SCA events


juliancnn:
| Wazuh version | Component | Install type | Install method   | Platform   |
| ------------- | --------- | ------------ | ---------------- | ---------- |
| 5.0.0         | Engine    | Manager      | any | Linux |

# Description

As part of epic #13038, it is necessary to decode the SCA events.

### SCA
**Queue: `'p'` / `0x70` / `SCA_MQ`**

The [DecodeSCA()](https://github.com/wazuh/wazuh/blob/dacf07da9930c4445ea55b1e5cdf6c7baea9a8d6/src/analysisd/decoders/security_configuration_assessment.c#L168) method is in charge of handling the SCA events, which are identified by the queue id 0x70 (SCA_MQ). Each new incoming message triggers the wazuhdb synchronization and retrieves its previous state to enrich the event.

It seems that only the `FillScanInfo()` and `FillCheckEventInfo()` modify the original event.
When enriching the event, the previous scan is stored on a new field called "sca.check.previous_result". This previous state is then matched with the rules to look for changes.

Issues found while starting to resolve this issue:
- [x]  Handling nested code: causes a complex, long and hard-to-debug piece of code, in order to tackle this a complete analysis has been done
- [x] how to tackle message pushed to CFGA queue (should be applied something like AR in wdbSync)

Discarded:

- How to engage with iterations inside the decoder?
- we need a helper or an engine functionality that allows us to take different field members (basically from an array) applying a helper, map, checking, or any other operation on it each time. This could be applied on a specific range or in all the array.  
- Helper function for cutting string X chars 

Analysis on DecodeSCA function:

![SCA decoder diagram](https://user-images.githubusercontent.com/95248059/176958576-dd9a142e-0205-43cb-b02f-ef0f392d6a70.jpg)

After discussing with the whole team the option of leaving this Decoder as a (not user-reachable function) was contemplated as possible. 
Another option is dividing it into two parts, in the first, the message gets decoded, all necessary fields are taken and separated, and once that it's achieved, all DB queries are made sitting the results that the user needs.
The option of modifying how the db talked it's arguments (not a single key - value but allowing it to take an array or a json directly) was also analyzed. 

The defined final approach was to unify all process inside a single helper function

LucioDonda:
Complete Raw logs obtained from a centos 8 agent for testing purposes:
[...]

Issue #13521, opened 2022-05-17
Engine - Decode dbsync events


juliancnn:
| Wazuh version | Component | Install type | Install method   | Platform   |
| ------------- | --------- | ------------ | ---------------- | ---------- |
| 5.0.0         | Engine    | Manager      | any | Linux |

# Description

As part of epic #13038, it is necessary to decode the dbsync events.

### DBSync (Manager)
**Queue: `'5'` / `0x35` / `DBSYNC_MQ`**

![image](https://user-images.githubusercontent.com/1891958/168679974-820b4476-1fc7-4e10-81e7-7305806871fa.png)

At the moment, the agents' and managers' databases synchronization is performed inside wazuh-analysisd. DBSync messages are received through the same queues as the unprocessed log messages, but they don't go through the decoding and rule matching stages. DBSync events are identified with the queue id 5 ([DBSYNC_MQ](https://github.com/wazuh/wazuh/blob/dacf07da9930c4445ea55b1e5cdf6c7baea9a8d6/src/analysisd/analysisd.c#L1337)), which allows analysisd to identify such messages to be treated in a specific way by using the [DispatchDBSync()](https://github.com/wazuh/wazuh/blob/dacf07da9930c4445ea55b1e5cdf6c7baea9a8d6/src/analysisd/decoders/dbsync.c#L257) method.

When synchronizing the [databases of the manager and its embedded agent, ](https://github.com/wazuh/wazuh/blob/dacf07da9930c4445ea55b1e5cdf6c7baea9a8d6/src/analysisd/decoders/dbsync.c#L104)as they don't use the same databases, wazuh-analysisd establishes direct communication with the syscollector and syscheck modules through their queues. On the other hand, when synchronizing a remote agent, the queue to be used is the remoted one ([queue/alerts/ar](https://github.com/wazuh/wazuh/blob/dacf07da9930c4445ea55b1e5cdf6c7baea9a8d6/src/analysisd/decoders/dbsync.c#L64)).

- Queries to the wazuhdb have the format `<component> <query>`, and the messages protocol uses the same header as the secureTCP communication, where the message size is specified.
- Messages can be 64K long at most.

PR #14367, opened 2022-07-20
Engine - WazuhDB

JcabreraC:
|Related issue|
|---|
|#13038|

This PR is to support WazuhDB in the new engine, as this is currently part of analysisd.

We need to create a wazudb client, and then use it to build an output of the engine, which will store events into the wazudb as defined in that output.

- [x] Implement the wazudb client:
    - #13519
    - #13520
    - #13521 
    - #13522 
    - #13523 
    - #13524
    - Implement tests and benchmarks


nmkoremblum (approved):

JavierBejMen (approved):
LGTM

LucioDonda (approved):
LGTM!

PR #14414, opened 2022-07-25
Add version control for agents reconnect procedure

### LINKED ISSUE #14370: Reconnect agents based on version in the TCP balancing system
## Description

The balancing system developed at https://github.com/wazuh/wazuh/issues/13195 is based on sending reconnection requests to the agents. However, this feature (forcing reconnection) was introduced in v4.3.0, so it is not possible to balance those agents with lower versions using this method.

Therefore, we should apply `version>"Wazuh v4.3.0"` filter when obtaining the ID of the agents to reconnect:
https://github.com/wazuh/wazuh/blob/1ade8b4d0593637c00e3185cb25bfdf485469765/framework/wazuh/core/cluster/agents_reconnect.py#L269-L274

In case fewer agents than expected are obtained (when there are too many old agents), a warning should be logged saying that the balancing will not work correctly for the reason described above. Other additional measures may also need to be taken in such cases (for example stopping the system). 

The goal is to avoid the system from stopping as if the balancer were misconfigured (agents not being rebalanced) since the requests would only be sent to those that can reconnect.

### wazuh/wazuh
- Unit tests without failures. Updated and/or expanded if there are new functions/methods/outputs:
  - [ ] Cluster (`framework/wazuh/core/cluster/tests/` & `framework/wazuh/core/cluster/dapi/tests/`)
  - [ ] Core (`framework/wazuh/core/tests/`)
  - [ ] SDK (`framework/wazuh/tests/`)
  - [ ] RBAC (`framework/wazuh/rbac/tests/`)
  - [ ] API (`api/api/tests/`)
- API tavern integration tests without failures. Updated and/or expanded if needed (`api/test/integration/`):
  - [ ] Affected tests 
  - [ ] Affected RBAC (black and white) tests
- [ ] Review integration test mapping using the script (`api/test/integration/mapping/integration_test_api_endpoints.json`)
- [ ] Review of spec.yaml examples and schemas (`api/api/spec/spec.yaml`)
- [ ] Review exceptions remediation when any endpoint path changes or is removed (`framework/wazuh/core/exception.py`)
- [ ] Changelog (`CHANGELOG.md`)

### wazuh/wazuh-documentation
- [ ] Migration from 3.X for changed endpoints (`source/user-manual/api/equivalence.rst`)
- [ ] Update RBAC reference with new/modified actions/resources/relationships (`source/user-manual/api/rbac/reference.rst`)

AdriiiPRodri:
|Related issue|
|---|
|#14370|

## Description

This PR closes #14370. In this PR we have modified the system of obtaining agents in the cluster, now only the agents with version equal or superior to 4.3.0 will be candidates to be balanced.

### DIFF
--- a/framework/wazuh/core/cluster/agents_reconnect.py
+++ b/framework/wazuh/core/cluster/agents_reconnect.py
@@ -1,5 +1,4 @@
 import contextlib
-
 from datetime import timedelta
 from enum import Enum
 from math import ceil
@@ -270,7 +269,8 @@ class AgentsReconnect:
                 if info['agents'] > 0:
                     agent_query = WazuhDBQueryAgents(
                         count=False, filters={"status": "active", "node_name": node},
-                        limit=info['agents'], sort={"fields": ["id"], "order": "desc"}, select=["id"])
+                        limit=info['agents'], sort={"fields": ["id"], "order": "desc"}, select=["id"],
+                        query="(version>Wazuh v4.3.0,version=Wazuh v4.3.0);id!=000")
                     current_balance[node]['agents'] = [info["id"] for info in agent_query.run()["items"]]
                 else:
                     current_balance[node]['agents'] = []
--- a/framework/wazuh/core/cluster/master.py
+++ b/framework/wazuh/core/cluster/master.py
@@ -1210,7 +1210,6 @@ class Master(server.AbstractServer):
             except agents_reconnect.SkippingException:
                 # Skip current iteration
                 logger.info("Skipping current iteration.")
-                pass
 
             # Check if the current phase is Halt
             if self.agents_reconnect.get_current_phase() == agents_reconnect.AgentsReconnectionPhases.HALT:


Selutario (approved):

framework/wazuh/core/cluster/agents_reconnect.py
```diff
@@ -270,7 +270,8 @@ async def get_agents(current_balance) -> dict:
                 if info['agents'] > 0:
                     agent_query = WazuhDBQueryAgents(
                         count=False, filters={"status": "active", "node_name": node},
-                        limit=info['agents'], sort={"fields": ["id"], "order": "desc"}, select=["id"])
+                        limit=info['agents'], sort={"fields": ["id"], "order": "desc"}, select=["id"],
+                        query="version>Wazuh v4.3.0,version=Wazuh v4.3.0")
```
Selutario, line 274: In my test this is returning agent 000 of the master node, but it can't be reconnected so it should be added to the query:
```suggestion
                        query="(version>Wazuh v4.3.0,version=Wazuh v4.3.0);id!=000")
```

Selutario (changes requested):
If all agents requested are `version < Wazuh v4.3.0`, this condition will match:
```PYTHON
if all(info['agents'] == [] for info in current_unbalanced_agents.values()):
            self.logger.info('The agents connected to the cluster are balanced.')
            current_unbalanced_agents.clear()
```

It will log that agents are balanced (which isn't true) and Balance algorithm won't be called (it is where a log is printed if there are not enough agents). 

Something extra will be needed to differentiate when the cluster is balanced from when all agent's version are lower than required.

framework/wazuh/core/cluster/agents_reconnect.py
AdriiiPRodri: Done

Selutario (approved):

davidjiglesias (approved):

Issue #12140, opened 2022-02-09
Improve gcp error management and error messages


CarlosRS9:
| Affected integration |
|---|
| GCloud |

The gcloud module does not handle exceptions properly. For example, when we provide a path to a credentials file it does not check if the file exists (Core checks this instead) nor if it has valid contents, which can generate a generic exception with a message that is not easy to interpret.

<details><summary>Real logs from the module simulating this described example</summary>

    2022/02/08 14:37:01 wazuh-modulesd:gcp-bucket[4571] wm_gcp.c:151 at wm_gcp_bucket_main(): DEBUG: Starting fetching of logs.
    2022/02/08 14:37:01 wazuh-modulesd:gcp-bucket[4571] wm_gcp.c:183 at wm_gcp_bucket_main(): INFO: Executing Bucket Analysis: (Bucket: framework-test-bucket, Type: access_logs, Credentials file: /var/ossec/wodles/gcloud/credentials.json)
    2022/02/08 14:37:01 wazuh-modulesd:gcp-bucket[4571] wm_gcp.c:352 at wm_gcp_bucket_run(): DEBUG: Create argument list
    2022/02/08 14:37:01 wazuh-modulesd:gcp-bucket[4571] wm_gcp.c:394 at wm_gcp_bucket_run(): DEBUG: Launching command: wodles/gcloud/gcloud --integration_type access_logs --bucket_name framework-test-bucket --credentials_file /var/ossec/wodles/gcloud/credentials.json --only_logs_after 2021-Sep-01 --log_level 1
    2022/02/08 14:37:02 wazuh-modulesd:gcp-bucket[4571] wm_gcp.c:407 at wm_gcp_bucket_run(): WARNING: Command returned exit code 1
    2022/02/08 14:37:02 wazuh-modulesd:gcp-bucket[4571] wm_gcp.c:411 at wm_gcp_bucket_run(): WARNING: Unknown error.
    2022/02/08 14:37:02 wazuh-modulesd:gcp-bucket[4571] wm_gcp.c:431 at wm_gcp_bucket_run(): DEBUG: OUTPUT: gcloud_wodle - CRITICAL - An exception happened while running the wodle: Expecting value: line 1 column 1 (char 0)
    2022/02/08 14:37:02 wazuh-modulesd:gcp-bucket[4571] wm_gcp.c:188 at wm_gcp_bucket_main(): DEBUG: Fetching logs finished.
[...]

PR #14500, opened 2022-08-05
Show correct client error messages for AWS module

### LINKED ISSUE #13436: AWS Integration - "Invalid credentials" message displayed instead of one related to the current error (request time different from AWS)
|---|---|---|---|---|
| 4.3.0-40310 | aws-s3 | Manager | Package (Quickstart) | CentOS8 |

## Configuration and step by step

* Stop the manager: `systemctl stop wazuh-manager`
* Configure modulesd in debug mode (`/var/ossec/etc/local_internal_options.conf`):
```
wazuh_modules.debug=2
```
* Enable the AWS wodle in the manager using the access key and the secret key:
```
<wodle name="aws-s3">
  <disabled>no</disabled>
  <remove_from_bucket>no</remove_from_bucket>
  <interval>1m</interval>
  <skip_on_error>no</skip_on_error>
  <bucket type="cloudtrail">
      <name>${replace_by_your_cloudtrail_bucket_name}</name>
      <access_key>${replace_by_your_AwsAccessKey}</access_key>
      <secret_key>${replace_by_your_AwsSecretKey}</secret_key>
      <only_logs_after>2022-MAY-12</only_logs_after>
  </bucket>
</wodle>
```
* Set the date of the Manager before the current UTC time.
```
timedatectl set-ntp false
timedatectl set-time '2022-05-12 17:50:23'
timedatectl set-ntp true
```
* Check that both times are different:
```
AWSdate=$(curl -v --silent -k https://s3.amazonaws.com/ 2>&1 | grep Date) ; echo $AWSdate && date -u
```
* Edit the `/var/ossec/wodles/aws/aws-s3.py` file (to see the error, otherwise you will see only the "invalid credentials" error showed below)
```python
        except botocore.exceptions.ClientError as e: # <-- MODIFY THIS LINE
            print("ERROR: Invalid credentials to access S3 Bucket")
            print(e) # <-- ADD THIS LINE
            exit(3)
```
* Start the manager `systemctl start wazuh-manager`

## Error in the `ossec.log` file without changing the `aws-s3` wodle source code

```
2022/05/12 15:35:21 wazuh-modulesd:aws-s3[16613] wm_aws.c:447 at wm_aws_run_s3(): WARNING: Bucket:  -  Returned exit code 3
2022/05/12 15:35:21 wazuh-modulesd:aws-s3[16613] wm_aws.c:466 at wm_aws_run_s3(): WARNING: Bucket:  -  Invalid credentials to access S3 Bucket
```

## Error with the change added to the source code

```
2022/05/12 15:35:21 wazuh-modulesd:aws-s3[16613] wm_aws.c:447 at wm_aws_run_s3(): WARNING: Bucket:  -  Returned exit code 3
2022/05/12 15:35:21 wazuh-modulesd:aws-s3[16613] wm_aws.c:466 at wm_aws_run_s3(): WARNING: Bucket:  -  Invalid credentials to access S3 Bucket
An error occurred (RequestTimeTooSkewed) when calling the ListObjectsV2 operation: The difference between the request time and the current time is too large.
```

## Screenshots
image

nico-stefani:
|Related issue|
|---|
|#13436|

## Description

This PR closes #13436. It fixes a bug caused when  `botocore.exceptions.ClientError` is catch. Now displays different messages for  `SignatureDoesNotMatch` and `RequestTimeTooSkewed` errors.

### Displaying an error for invalid credentials
```bash
root@1ba6185a394b:/var/ossec# /var/ossec/wodles/aws/aws-s3 -b wazuh-aws-wodle-cloudtrail -t cloudtrail -s 2022-Feb-01 -p dev -d2
DEBUG: +++ Debug mode on - Level: 2
ERROR: Invalid credentials to access S3 Bucket
```

### Displaying an error for time difference
```bash
root@1ba6185a394b:/var/ossec# FAKETIME="2022-07-01 16:00:00" /var/ossec/wodles/aws/aws-s3 -b wazuh-aws-wodle-cloudtrail -t cloudtrail -s 2022-Feb-01 -p dev -d2
DEBUG: +++ Debug mode on - Level: 2
ERROR: The server datetime and datetime of the AWS environment differ
```

### DIFF
--- a/wodles/aws/aws_s3.py
+++ b/wodles/aws/aws_s3.py
@@ -25,6 +25,7 @@
 #   16 - Throttling error
 #   17 - Invalid key format
 #   18 - Invalid prefix
+#   19 - The server datetime and datetime of the AWS environment differ
 
 import argparse
 import signal
@@ -69,6 +70,11 @@ DEPRECATED_MESSAGE = 'The {name} authentication parameter was deprecated in {rel
 
 # Enable/disable debug mode
 debug_level = 0
+INVALID_CREDENTIALS_ERROR_CODE = "SignatureDoesNotMatch"
+INVALID_REQUEST_TIME_ERROR = "RequestTimeTooSkewed"
+
+INVALID_CREDENTIALS_ERROR_MESSAGE = "Invalid credentials to access S3 Bucket"
+INVALID_REQUEST_TIME_ERROR_MESSAGE = "The server datetime and datetime of the AWS environment differ"
 
 ################################################################################
 # Classes
@@ -1083,14 +1089,24 @@ class AWSBucket(WazuhIntegration):
             else:
                 print("ERROR: No files were found in '{0}'. No logs will be processed.".format(self.bucket_path))
                 exit(14)
-        except botocore.exceptions.ClientError as err:
-            if err.response['Error']['Code'] == 'ThrottlingException':
+
+        except botocore.exceptions.ClientError as error:
+            error_message = "Unknown"
+            exit_number = 1
+            error_code = error.response.get("Error", {}).get("Code")
+
+            if error_code == 'ThrottlingException':
                 debug(f'Error: The "check_bucket" request was denied due to request throttling. ', 2)
                 sys.exit(16)
-            else:
-                print("ERROR: Invalid credentials to access S3 Bucket")
-                exit(3)
+            elif error_code == INVALID_CREDENTIALS_ERROR_CODE:
+                error_message = INVALID_CREDENTIALS_ERROR_MESSAGE
+                exit_number = 3
+            elif error_code == INVALID_REQUEST_TIME_ERROR:
+                error_message = INVALID_REQUEST_TIME_ERROR_MESSAGE
+                exit_number = 19
 
+            print(f"ERROR: {error_message}")
+            exit(exit_number)
         except botocore.exceptions.EndpointConnectionError as e:
             print(f"ERROR: {str(e)}")
             exit(15)
@@ -1598,7 +1614,7 @@ class AWSVPCFlowBucket(AWSLogsBucket):
                 flow_log_id,
                 log_key,
                 processed_date,
-                created_date) 
+                created_date)
             VALUES (
                 :bucket_path,
                 :aws_account_id,
@@ -2005,7 +2021,7 @@ class AWSCustomBucket(AWSBucket):
                 aws_account_id,
                 log_key,
                 processed_date,
-                created_date) 
+                created_date)
             VALUES (
                 :bucket_path,
                 :aws_account_id,
@@ -2562,9 +2578,23 @@ class AWSServerAccess(AWSCustomBucket):
             if not 'CommonPrefixes' in self.client.list_objects_v2(Bucket=self.bucket, Delimiter='/'):
                 print("ERROR: No files were found in '{0}'. No logs will be processed.".format(self.bucket_path))
                 exit(14)
-        except botocore.exceptions.ClientError as err:
-            debug(f'ERROR: The "check_bucket" request failed: {err}', 1)
-            sys.exit(16)
+        except botocore.exceptions.ClientError as error:
+            error_message = "Unknown"
+            exit_number = 1
+            error_code = error.response.get("Error", {}).get("Code")
+
+            if error_code == 'ThrottlingException':
+                debug(f'ERROR: The "check_bucket" request failed: {error}', 1)
+                sys.exit(16)
+            elif error_code == INVALID_CREDENTIALS_ERROR_CODE:
+                error_message = INVALID_CREDENTIALS_ERROR_MESSAGE
+                exit_number = 3
+            elif error_code == INVALID_REQUEST_TIME_ERROR:
+                error_message = INVALID_REQUEST_TIME_ERROR_MESSAGE
+                exit_number = 19
+
+            print(f"ERROR: {error_message}")
+            exit(exit_number)
 
     def load_information_from_file(self, log_key):
         """Load data from a S3 access log file."""
--- a/wodles/aws/aws_s3.py
+++ b/wodles/aws/aws_s3.py
@@ -71,11 +71,12 @@ DEPRECATED_MESSAGE = 'The {name} authentication parameter was deprecated in {rel
 # Enable/disable debug mode
 debug_level = 0
 INVALID_CREDENTIALS_ERROR_CODE = "SignatureDoesNotMatch"
-INVALID_REQUEST_TIME_ERROR = "RequestTimeTooSkewed"
-THROTTLING_EXCEPTION_ERROR = "ThrottlingException"
+INVALID_REQUEST_TIME_ERROR_CODE = "RequestTimeTooSkewed"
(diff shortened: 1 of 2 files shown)


CarlosRS9 (changes requested):
After the changes it is still not clear enough for the user. 

The message `The difference between the request time and the current time is too large` is too confusing to understand what it refers to without the necessary context.

It would be advisable to try to capture the `RequestTimeTooSkewed` itself and display a more readable message to the user that their problem is that the datetime of their host and the datetime of their AWS environment differ.

On the other hand, this PR introduces interesting changes in exception handling, creating an auxiliary function and a structure to determine the corresponding message according to the exception. This is interesting and necessary, but we should not apply this change in this PR. A change like this should be done in a separate PR, which includes the complete refactoring of all the exceptions of the module. Moreover, this new exception handling should go hand in hand with what we did in the [Improve gcp error management and error messages](https://github.com/wazuh/wazuh/issues/12140) issue, to unify as much as possible the implementations of both modules.

Therefore, it would be advisable instead to open a new issue to address this proposal separately.

nico-stefani:
> After the changes it is still not clear enough for the user.
> 
> The message `The difference between the request time and the current time is too large` is too confusing to understand what it refers to without the necessary context.
> 
> It would be advisable to try to capture the `RequestTimeTooSkewed` itself and display a more readable message to the user that their problem is that the datetime of their host and the datetime of their AWS environment differ.

Ok, i will make the changes to catch this error too.

> 
> On the other hand, this PR introduces interesting changes in exception handling, creating an auxiliary function and a structure to determine the corresponding message according to the exception. This is interesting and necessary, but we should not apply this change in this PR. A change like this should be done in a separate PR, which includes the complete refactoring of all the exceptions of the module. Moreover, this new exception handling should go hand in hand with what we did in the [Improve gcp error management and error messages](https://github.com/wazuh/wazuh/issues/12140) issue, to unify as much as possible the implementations of both modules.
> 
> Therefore, it would be advisable instead to open a new issue to address this proposal separately.

Yes, understand that you are saying. But i did this in order to avoid repeat the code, because the same problem is present in two different places. I think this solution is an intermediate step between what we have now and the entire refactor of all the module. 
Saying this, do yo still consider that i should make the changes?

CarlosRS9:
> Yes, understand that you are saying. But i did this in order to avoid repeat the code, because the same problem is present in two different places. I think this solution is an intermediate step between what we have now and the entire refactor of all the module.
Saying this, do yo still consider that i should make the changes?

I agree with you that your implementation avoids redundant code. This is something we need to improve in these modules.

As you say, your proposal is an intermediate step. However, we are not clear if this intermediate step goes in the direction we want to go for the final implementation, because as I said the idea would be to use a model similar to the one we apply in https://github.com/wazuh/wazuh/issues/12140, with an `exception.py` file, among other things. It would not make much sense to apply this intermediate step and then discard it completely in a future revision. 

Before applying this change we must meet with the rest of the team and define what will be the final proposal, so I think it is better not to block this PR with changes that go beyond its original scope and instead implement the proposed changes, which are less intrusive, to unlock this development.

You should create a new issue to address this question so we can discuss it with the team.

wodles/aws/aws_s3.py
```diff
@@ -1084,14 +1090,24 @@ def check_bucket(self):
...
                 exit(14)
-        except botocore.exceptions.ClientError as err:
-            if err.response['Error']['Code'] == 'ThrottlingException':
+
+        except botocore.exceptions.ClientError as error:
+            error_message = "Unknown"
+            exit_number = 1
+            error_code = error.response.get("Error", {}).get("Code")
+
+            if error_code == 'ThrottlingException':
```
CarlosRS9, line 1099: Adapt this code fragment to behave in the same way as the other ones you updated. Use `error_message` and `exit_number` for consistency.

wodles/aws/aws_s3.py
```diff
@@ -2543,9 +2559,23 @@ def check_bucket(self):
...
                 exit(14)
-        except botocore.exceptions.ClientError as err:
-            debug(f'ERROR: The "check_bucket" request failed: {err}', 1)
-            sys.exit(16)
+        except botocore.exceptions.ClientError as error:
+            error_message = "Unknown"
+            exit_number = 1
+            error_code = error.response.get("Error", {}).get("Code")
+
+            if error_code == 'ThrottlingException':
```
CarlosRS9, line 2567: Same as above

wodles/aws/aws_s3.py
```diff
@@ -69,6 +70,12 @@
 
 # Enable/disable debug mode
 debug_level = 0
+INVALID_CREDENTIALS_ERROR_CODE = "SignatureDoesNotMatch"
+INVALID_REQUEST_TIME_ERROR = "RequestTimeTooSkewed"
```
CarlosRS9, line 74: Naming conventions must be consistent. Use the `_ERROR` or `ERROR_CODE` suffix but stick to one of them.

wodles/aws/aws_s3.py
```diff
@@ -1084,14 +1091,24 @@ def check_bucket(self):
...
-                print("ERROR: Invalid credentials to access S3 Bucket")
-                exit(3)
 
+        except botocore.exceptions.ClientError as error:
+            error_message = "Unknown"
+            exit_number = 1
+            error_code = error.response.get("Error", {}).get("Code")
+
+            if error_code == THROTTLING_EXCEPTION_ERROR:
+                error_message = "The 'check_bucket' request was denied due to request throttling."
```
CarlosRS9, line 1101: To be consistent with the rest of your changes this message should be placed in a constant, too. You should create something like `THROTTLING_EXCEPTION_ERROR_MESSAGE` and set this to:
```suggestion
                error_message = f"{THROTTLING_EXCEPTION_ERROR_MESSAGE}: {error}"
```
Apply this change accordingly every time a throttling exception is being handled in the rest of the module.

CarlosRS9 (changes requested):

wodles/aws/aws_s3.py
```diff
@@ -69,6 +70,13 @@
 
 # Enable/disable debug mode
 debug_level = 0
+INVALID_CREDENTIALS_ERROR_CODE = "SignatureDoesNotMatch"
+INVALID_REQUEST_TIME_ERROR_CODE = "RequestTimeTooSkewed"
+THROTTLING_EXCEPTION_ERROR_CODE = "ThrottlingException"
+
+INVALID_CREDENTIALS_ERROR_MESSAGE = "Invalid credentials to access S3 Bucket"
+INVALID_REQUEST_TIME_ERROR_MESSAGE = "The server datetime and datetime of the AWS environment differ"
+THROTTLING_EXCEPTION_ERROR_MESSAGE = "The 'check_bucket' request was denied due to request throttling"
```
CarlosRS9, line 79: This message is not only used in the `check_bucket` function but also in `find_account_ids` and `find_regions`. Update it so it can be used in these functions, too.

```suggestion
THROTTLING_EXCEPTION_ERROR_MESSAGE = "The '{name}' request was denied due to request throttling"
```

CarlosRS9 (changes requested):

davidjiglesias (approved):

PR #14519, opened 2022-08-09
Check mtime in cluster files before overwriting them

### LINKED ISSUE #14505: Check mtime in cluster files before overwriting them
| Epic issue |
|--|
| #14492 |

## Description
Now, the decoders/rules/cdb lists files can be modified in the workers by the Integrity sync task and DAPI. In addition, these tasks are run in parallel by child processes within the cluster.

Possibly this leads to various race conditions that we will have to study and test in the coming days. One of them could be the following:
1. A rule is modified in the master node.
2. Worker receives zip with files to update coming from the master.
3. Worker unzip it and start processing those files.
4. A rule is modified in the cluster (including this worker) through an API distributed request. The worker did not finish processing unzipped files yet.
5. The rule file coming from master in step `2`. is processed and overwrites the file modified in step `4.`

The worker should check if each file's modification date is later than when the zip was received from the master.

## Checks

The following elements have been updated or reviewed (should also be checked if no modification is required):
- [x] Tests (unit tests, API integration tests).
- [ ] Changelog.
- [x] Documentation.
- [x] Integration test mapping (using `api/test/integration/mapping/_test_mapping.py`).

Selutario:
|Related issue|
|---|
| Closes #14505 |

## Description

This PR changes the worker's Integrity sync task behavior so the use case described at #14505 is prevented. Now, while replacing worker files with those coming from the master, any file whose modification time is newer than the timestamp when the zip started to be created in the master, won't be replaced. 

This way, if the DAPI changed a custom rule in all nodes, the master won't replace the file with an old copy. However, if the file is still different in the next iteration, it will be changed (since the zip timestamp will be newer than the file's modification time).

## Logs/Alerts example
This message will be logged when a file isn't replaced:
```
2022/08/08 12:50:03 INFO: [Worker worker1] [Integrity sync] Starting.
2022/08/08 12:50:03 INFO: [Worker worker1] [Integrity sync] Files to create: 0 | Files to update: 1 | Files to delete: 0 | Files to send: 0
2022/08/08 12:50:14 WARNING: [Worker worker1] [Integrity sync] The compression date of this master's file is older than its modification date on the current node, so it will not be replaced: /var/ossec/etc/rules/local_rules.xml
2022/08/08 12:50:28 INFO: [Worker worker1] [Integrity sync] Finished in 25.375s.
```

### DIFF
--- a/framework/wazuh/core/cluster/cluster.py
+++ b/framework/wazuh/core/cluster/cluster.py
@@ -312,7 +312,7 @@ def compress_files(name, list_path, cluster_control_json=None):
         Path where the zip file has been saved.
     """
     failed_files = list()
-    zip_file_path = path.join(common.wazuh_path, 'queue', 'cluster', name, f'{name}-{time()}-{str(random())[2:]}.zip')
+    zip_file_path = path.join(common.wazuh_path, 'queue', 'cluster', name, f'{name}-{str(random())[2:]}-{time()}.zip')
     if not path.exists(path.dirname(zip_file_path)):
         mkdir_with_mode(path.dirname(zip_file_path))
     with zipfile.ZipFile(zip_file_path, 'x') as zf:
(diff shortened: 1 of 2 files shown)


Selutario:
Finally, we implemented a different approach in https://github.com/wazuh/wazuh/issues/14520

Issue #14508, opened 2022-08-08
Deprecate the plain text authentication method from AWS and Azure


CarlosRS9:
| Affected integration |
|---|
| AWS and Azure |

## Description

In https://github.com/wazuh/wazuh/issues/13660 we have reviewed the different authentication options and we have concluded that we should deprecate and remove the option that allows the user to store their credentials as plain text in the ossec.conf. We will indicate them that it is necessary to use credential files instead.

We will follow the following steps for its removal:

1. In `4.4` we will deprecate the user and password parameters, updating the ossec.conf parser to show a warning message if one of them are being used. The idea is to warn the user that for the moment they are still accepted, but in the next major they will be completely removed.
2. In `5.0` we will remove these options. A configuration containing one of those will result in an error and the module will not run.

Having this in mind, we need to apply the following changes for `4.4`:

1. Update the ossec.conf parser to detect these parameters and show a warning message.
2. Update `-h` option of the python module to show that these parameters are now deprecated.
3. Update the python module to show a warning if the user try to manually run it using these parameters.
4. Update the Wazuh documentation to acomodate these changes.

## Tasks
- [x] Test in a manager.
- [x] Test in an agent.
- [x] Unit tests without failures. Updated if there are any relevant changes.
- [x] Integration tests without failures. Updated if there are any relevant changes.
- [x] Update the documentation if necessary.
- [x] Add entry to the changelog if necessary.

CarlosRS9:
## Ossec.conf parser changes
We have already implemented the necessary changes in our `ossec.conf` parser to detect the deprecated tags and warn the user with a message.

When a deprecated tag, such as `access_key` is detected the following messages are shown:
```
2022/08/08 12:51:36 wazuh-modulesd: WARNING: Deprecated tag <access_key> found at module 'aws-s3'. This tag was deprecated in 4.4. Please, use a different authentication method.
2022/08/08 12:51:36 wazuh-modulesd: WARNING: Check https://documentation.wazuh.com/current/amazon/services/prerequisites/credentials.html for more information.
```
These messages are shown as part of the Wazuh service output instead of the ossec.log, which means they will be visible when the user attempts to start or restart the Wazuh service.

Here is an example of a real configuration using AWS buckets, AWS services, Azure Graph, Azure Log Analytics y Azure Storage:

<details><summary>Configuration used</summary>

	<wodle name="aws-s3">
	    <disabled>no</disabled>
	    <remove_from_bucket>no</remove_from_bucket>
[...]

Issue #13660, opened 2022-06-01
Review the AWS authentication method that allows users to set up their credentials in the Wazuh configuration file


CarlosRS9:
| Affected integration |
|---|
| AWS |

Currently our AWS module offers different [authentication options](https://documentation.wazuh.com/current/amazon/services/prerequisites/credentials.html#configuring-aws-credentials) to allow the user to provide the module with the necessary credentials for its operation. One of these available options, although it is the least recommended, is to store directly in the ossec.conf the values for `access_key` and `secret_key`, like in this configuration example:

```
<bucket type="cloudtrail">
  <name>my-bucket</name>
  <access_key>xxxxxx</access_key>
  <secret_key>yyyyyy</secret_key>
</bucket>
```

If a user applies a configuration like the previous one this results in his `access_key` and `secret_key` to appear as plain text in our `ossec.log` and in the Wazuh Service Status output.

We need to study the impact of replacing or removing this authentication option completely, as well as the alternatives available to reduce the impact on users currently using this method.

Issue #14535, opened 2022-08-10
Review and improve the way the cloud modules send their output to the ossec.log


CarlosRS9:
| Affected integration |
|---|
| AWS, GCloud, Azure and DockerListener|

## Description

We should review and update the way we are capturing the output of our cloud modules (Azure, AWS y GCP) as well as the Docker-Listener module.

To better understand why this is necessary, first it is important to know how these modules are implemented. They are made of two different component:

- The core part, written in C, which launches the modules, catches the output and prints it to the ossec.log.
- The module itself, written in python, which executes the integration and generates some output to be shown in the stdout.

The issue here is that the way the output is written and sent to the ossec.log is not consistent between modules. In addition to that, the output of the different modules will be ignored, even if there are warning or error messages, if the debug mode is not enabled, which is disabled by default.

Here is a detailed explanation on how every module works currently:

### AWS

The python module runs and its output, which is written using `print` statements, is caught. However, for the output to be dumped to the ossec.log at least one of the following conditions must be met:

- If there is an internal error in the C part the output will be written to ossec.log if debug mode is 1 or higher.
- If the execution ends with an error code `1` the output will be written to ossec.log if the `Unknown error` string was present.
- If the execution ends with an error code `2` the output will be written to ossec.log if the `aws.py: error:` string was present.
- If the execution ends with an error code `3` or higher the output will be written to ossec.log as a WARNING, regardless of if it was an ERROR.
- If the execution ends successfully with an return value `0` or fails and returns a negative value the output will be written to the ossec.log only if debug mode is set to `2`.

### GCP

The python module runs and its output, which is written using `logging` statements, is caught. However, for the output to be dumped to the ossec.log at least one of the following conditions must be met:

- If debug level is 0 only CRITICAL, ERROR and WARNING messages will be written
- If debug level is 1 it will show INFO messages in addition to the previously listed.
- If debug level is 2 it will show DEBUG messages in addition to the previously listed.

This is done by looking for the `WARNING`, `ERROR`, `INFO` strings present in the output. If not present, the output is skipped.

### Azure

The azure module uses logging in a similar way to how the GCP was implemented. However, the C implementation is not expecting output from this module and it is not processing it, hence nothing will be written into the ossec.log at all. This is because the older implementation wrote its output directly to a custom file and now it is set to send it to the stdout but the core part must be updated to expect and catch this output.

### Docker Listener

[...]

PR #14543, opened 2022-08-11
Deprecate Azure and AWS authentication method

CarlosRS9:
|Related issue|
|---|
| #14508 |

## Description

In #13660 we have reviewed the different authentication options and we have concluded that we should deprecate and remove the option that allows the user to store their credentials as plain text in the ossec.conf. 

This PR applies the necessary changes to deprecate this option. The idea is to deprecate this in `4.4` and completely remove the option in `5.0`.

Here is some of the major changes we have applied:
- Updated the `ossec.conf` parser to detect these parameters and show a warning message.
- Updated `-h` option of the python module to show that these parameters are now deprecated.
- Updated the python module to show a warning if the user try to manually run it using these parameters.

More details about the introduced changes and the testing performed can be found in #14508.

### DIFF
--- a/src/config/wmodules-aws.c
+++ b/src/config/wmodules-aws.c
@@ -56,6 +56,7 @@ static const char *CLOUDWATCHLOGS_SERVICE_TYPE = "cloudwatchlogs";
 static const char *CISCO_UMBRELLA_BUCKET_TYPE = "cisco_umbrella";
 
 static const char *AUTHENTICATION_OPTIONS_URL = "https://documentation.wazuh.com/current/amazon/services/prerequisites/credentials.html";
+static const char *DEPRECATED_MESSAGE = "Deprecated tag <%s> found at module '%s'. This tag was deprecated in %s; please use a different authentication method. Check %s for more information.";
 
 // Parse XML
 
@@ -238,7 +239,7 @@ int wm_aws_read(const OS_XML *xml, xml_node **nodes, wmodule *module)
                         }
                     } else if (!strcmp(children[j]->element, XML_ACCESS_KEY)) {
                         if (strlen(children[j]->content) != 0) {
-                            mwarn("Deprecated tag <%s> found at module '%s'. This tag was deprecated in 4.4. Please, use a different authentication method. Check %s for more information.", children[j]->element, WM_AWS_CONTEXT.name, AUTHENTICATION_OPTIONS_URL);
+                            mwarn(DEPRECATED_MESSAGE, children[j]->element, WM_AWS_CONTEXT.name, "4.4", AUTHENTICATION_OPTIONS_URL);
                             free(cur_bucket->access_key);
                             os_strdup(children[j]->content, cur_bucket->access_key);
                         }
@@ -249,7 +250,7 @@ int wm_aws_read(const OS_XML *xml, xml_node **nodes, wmodule *module)
                         }
                     } else if (!strcmp(children[j]->element, XML_SECRET_KEY)) {
                         if (strlen(children[j]->content) != 0) {
-                            mwarn("Deprecated tag <%s> found at module '%s'. This tag was deprecated in 4.4. Please, use a different authentication method. Check %s for more information.", children[j]->element, WM_AWS_CONTEXT.name, AUTHENTICATION_OPTIONS_URL);
+                            mwarn(DEPRECATED_MESSAGE, children[j]->element, WM_AWS_CONTEXT.name, "4.4", AUTHENTICATION_OPTIONS_URL);
                             free(cur_bucket->secret_key);
                             os_strdup(children[j]->content, cur_bucket->secret_key);
                         }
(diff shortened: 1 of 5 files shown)


src/config/wmodules-aws.c
```diff
@@ -236,6 +238,7 @@ int wm_aws_read(const OS_XML *xml, xml_node **nodes, wmodule *module)
                         }
                     } else if (!strcmp(children[j]->element, XML_ACCESS_KEY)) {
                         if (strlen(children[j]->content) != 0) {
+                            mwarn("Deprecated tag <%s> found at module '%s'. This tag was deprecated in 4.4. Please, use a different authentication method. Check %s for more information.", children[j]->element, WM_AWS_CONTEXT.name, AUTHENTICATION_OPTIONS_URL);
```
nico-stefani, line 241: I think that the message string should be placed in a constant too, something like that  
```c
static const char *DEPRECATED_MESSAGE_TEMPLATE = "Deprecated tag <%s> found at module '%s'. This tag was deprecated in 4.4. Please, use a different authentication method. Check %s for more information."
```
And later be used in the four cases

```c
mwarn(DEPRECATED_MESSAGE_TEMPLATE, children[j]->element, WM_AWS_CONTEXT.name, AUTHENTICATION_OPTIONS_URL);
```

src/config/wmodules-azure.c
```diff
@@ -237,9 +239,11 @@ int wm_azure_api_read(const OS_XML *xml, XML_NODE nodes, wm_azure_api_t * api_co
 
         } else if (!strcmp(nodes[i]->element, XML_APP_ID)) {
             if (*nodes[i]->content != '\0')
+                mwarn("Deprecated tag <%s> found at module '%s'. This tag was deprecated in 4.4. Please, use a different authentication method. Check %s for more information.", nodes[i]->element, WM_AZURE_CONTEXT.name, AUTHENTICATION_OPTIONS_URL);
```
nico-stefani, line 242: Same suggestion as in AWS module.

wodles/azure/azure-logs.py
```diff
@@ -77,9 +78,11 @@ def get_script_arguments():
 
     # Log Analytics arguments #
     parser.add_argument("--la_id", metavar='ID', type=str, required=False,
-                        help="Application ID for Log Analytics authentication.")
+                        help="Application ID for Log Analytics authentication. This option was deprecated in 4.4. "
```
nico-stefani, line 81: Same suggestion as in AWS module. Some parts of the help messages are common and can be improved using a constant in order to maintain the readability of code.

nico-stefani (changes requested):

src/config/wmodules-aws.c
CarlosRS9: Done

wodles/aws/aws_s3.py
```diff
@@ -3438,8 +3444,16 @@ def get_script_arguments():
...
+    parser.add_argument('-a', '--access_key', dest='access_key', default=None,
+                        help='S3 Access key credential. This option was deprecated in 4.4. Please use another '
+                             'authentication method instead. Check '
+                             'https://documentation.wazuh.com/current/amazon/services/prerequisites/credentials.html '
+                             'for more information.')
+    parser.add_argument('-k', '--secret_key', dest='secret_key', default=None,
+                        help='S3 Access key credential. This option was deprecated in 4.4. Please use another '
+                             'authentication method instead. Check '
+                             'https://documentation.wazuh.com/current/amazon/services/prerequisites/credentials.html '
+                             'for more information.')
```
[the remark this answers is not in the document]
CarlosRS9: Done

src/config/wmodules-azure.c
CarlosRS9: Done

wodles/azure/azure-logs.py
CarlosRS9: Done

nico-stefani (approved):

davidjiglesias (changes requested):
Should we add information about when is the feature to be removed? Not only deprecated.

Also, we should update v4.4.0 documentation to include the deprecation, remove the future releases message.

The WARNING should appear in ossec.log without debug enabled.

CarlosRS9:
> Should we add information about when is the feature to be removed? Not only deprecated.

We cannot add this information yet. Because of the impact of these breaking changes we must wait until the removal of the agent `000` from the managers is completed. Therefore, at the moment we cannot specify with certainty the exact version in which it will be finally removed. Once it is defined it will be updated accordingly.

> Also, we should update v4.4.0 documentation to include the deprecation, remove the future releases message.

Agreed. I have created https://github.com/wazuh/wazuh-documentation/pull/5563 to update it accordingly, stating `4.4` as the version when it was deprecated.

> The WARNING should appear in ossec.log without debug enabled.

As I indicated in the issue [here](https://github.com/wazuh/wazuh/issues/14508#issuecomment-1209361557), due to the way our core implementation works we can only add this message as a debug one. Once https://github.com/wazuh/wazuh/issues/14535 is done this will appear as a warning message, no additional changes needed.

davidjiglesias (approved):

davidjiglesias (approved):

PR #14726, opened 2022-08-24
Wazuh agent build on alpine linux

### LINKED ISSUE #4592: Fix the Wazuh agent compilation for Alpine Linux
| 3.11 | Agent | Packages/Sources | Alpine Linux |

Is there a Wazuh agent package for Alpine Linux? Can I build it from source code?
It would be nice to work on it.

Thanks!

FrancoRivero:
|Related issue|
|---|
|Closes #4592|

## Description
Hello team, this pull-request is to be able to build on Alpine Linux and to be able to install an agent inside Alpine.
Regards

## Tests
  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [x] Source installation
- [x] Package installation
- [ ] Source upgrade
- [ ] Package upgrade

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -75,6 +75,7 @@ USE_MSGPACK_OPT=yes
 DISABLE_JEMALLOC?=no
 DISABLE_SYSC?=no
 DISABLE_CISCAT?=no
+ALPINE?=no
 
 ifeq (${TARGET},winagent)
 CMAKE_OPTS=-DCMAKE_SYSTEM_NAME=Windows -DCMAKE_C_COMPILER=${MING_BASE}${CC} -DCMAKE_CXX_COMPILER=${MING_BASE}${CXX}
@@ -136,6 +137,9 @@ else
 endif
 		OSSEC_LDFLAGS+='-Wl,-rpath,$$ORIGIN/../lib'
 		AR_LDFLAGS+='-Wl,-rpath,$$ORIGIN/../../lib'
+ifneq (,$(filter ${ALPINE},YES yes y Y 1))
+		OSSEC_LIBS+=-lintl
+endif
 		OSSEC_LIBS+=-lrt -ldl -lm
 		OSSEC_LDFLAGS+=-pthread -lrt -ldl
 		AR_LDFLAGS+=-pthread -lrt -ldl
@@ -593,6 +597,7 @@ help: failtarget
 	@echo "   make DISABLE_CISCAT=yes      Not to build the CIS-CAT module (for unsupported systems). Allowed values are 1, yes, YES, y and Y, otherwise, the flag is ignored"
 	@echo "   make OPTIMIZE_CPYTHON=yes    Enable this flag to optimize the python interpreter build, which is performed when used PYTHON_SOURCE. Allowed values are 1, yes, YES, y and Y, otherwise, the flag is ignored"
 	@echo "   make PYTHON_SOURCE=yes       Used along the deps target. Downloads the sources needed to build the python interpreter. Allowed values are 1, yes, YES, y and Y, otherwise, the flag is ignored"
+	@echo "   make ALPINE=yes 	       Build by Alpine Linux systems"
 	@echo
 	@echo "Database options: "
 	@echo "   make DATABASE=mysql          Build with MYSQL Support"
@@ -637,6 +642,7 @@ settings:
 	@echo "    USE_AUDIT:          ${USE_AUDIT}"
 	@echo "    DISABLE_SYSC:       ${DISABLE_SYSC}"
 	@echo "    DISABLE_CISCAT:     ${DISABLE_CISCAT}"
+	@echo "    ALPINE:	        ${ALPINE}"
 	@echo "Mysql settings:"
 	@echo "    includes:           ${MI}"
 	@echo "    libs:               ${ML}"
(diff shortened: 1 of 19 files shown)


src/Makefile
```diff
@@ -1008,9 +1013,12 @@ PROCPS_INCLUDE=-I./${EXTERNAL_PROCPS}
 
 procps_c := $(wildcard ${EXTERNAL_PROCPS}*.c)
 procps_o := $(procps_c:.c=.o)
+ifeq (${CHECK_ALPINE},YES)
+alpine_debug := "-DPATH_MAX=4096"
```
chemamartinez, line 1017: Why is needed to include the `PATH_MAX` definition when compiling the `procps` lib. What error is throwing?

src/Readme.md
```diff
@@ -187,4 +187,24 @@ Output Example executing the SCANBUILD analysis with `winagent` target:
 [CleanInternals: PASSED]
 [ScanBuild: PASSED]
 <winagent>[SCANBUILD: PASSED]<winagent>
-```
\ No newline at end of file
+```
+
+## Compile Wazuh on Alpine Linux
```
chemamartinez, line 192: Regarding the whole content of this readme, I think it has no sense to include here the guide to installing the agent in Alpine Linux. Let's remove this change and let the documentation explain that.

chemamartinez (commented):
I also miss the fix for the `wazuh-control` script. In previous conversations about the Alpine Linux agent, there was found a problem when starting the agent since `ps -p` is not available in Alpine, so the script doesn't detect that the binaries are up, deleting the PID files.

src/Makefile
FrancoRivero: Hi Chema, when you build Wazuh in debug mode un Alpine Linux if you don't set PATH_MAX to procps in Alpine this library cannot be built for this distribution with a message similar to this:
```
external/procps/readproc.c:1038:26: error: 'PATH_MAX' undeclared (first use in this function)
 1038 |         static char path[PATH_MAX], sbuf[1024];
      |                          ^~~~~~~~
external/procps/readproc.c:1038:26: note: each undeclared identifier is reported only once for each function it appears in
external/procps/readproc.c:1038:21: warning: unused variable 'path' [-Wunused-variable]
 1038 |         static char path[PATH_MAX], sbuf[1024];
      |                     ^~~~
external/procps/readproc.c: In function 'simple_nexttid':
external/procps/readproc.c:739:46: warning: '%s' directive output may be truncated writing up to 255 bytes into a region of size between 41 and 51 [-Wformat-truncation=]
  739 |   snprintf(path, PROCPATHLEN, "/proc/%d/task/%s", p->tgid, ent->d_name);
      |                                              ^~
external/procps/readproc.c:739:3: note: 'snprintf' output between 14 and 279 bytes into a destination of size 64
  739 |   snprintf(path, PROCPATHLEN, "/proc/%d/task/%s", p->tgid, ent->d_name);
```

src/Readme.md
FrancoRivero: OK, I delete this section !

src/Makefile
chemamartinez: Thank you! Could you also please uppercase the variable to `ALPINE_DEBUG` just to be consistent with the other variables in the Makefile?

chemamartinez:
> Hi @chemamartinez, the version I am working on `ps -p` is available and works fine. NAME="Alpine Linux" ID=alpine VERSION_ID=3.16.2 PRETTY_NAME="Alpine Linux v3.16" HOME_URL="https://alpinelinux.org/"

Thanks, Franco! I am using the same version, maybe missing some needed package but it is not available in my host.
```
localhost:~# cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.16.2
PRETTY_NAME="Alpine Linux v3.16"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
localhost:~# ps -p
ps: unrecognized option: p
```

Let's include the fix to ensure it works for all Alpine versions.

chemamartinez:
I found one more error during the installation process. The script [adduser.sh](https://github.com/wazuh/wazuh/blob/4592-wazuh-agent-package-for-alpine-linux/src/init/adduser.sh) is not able to create the `wazuh` user and group.

```
./init/adduser.sh: line 61: /usr/sbin/groupadd: not found
install: unknown user wazuh
install: unknown group wazuh
```

Instead of including this as a previous manual step, it is better to fix the script just detecting when installing on Alpine Linux by doing the following assignment:
```
        GROUPADD="/usr/sbin/addgroup"
        USERADD="/usr/sbin/adduser"
```
which are the corresponding binaries in Alpine.

src/Makefile
FrancoRivero: Done

src/init/adduser.sh
```diff
@@ -79,6 +83,8 @@ else
             GID=$(cat /etc/group | grep wazuh| cut -d':' -f 3)
             uid=$(( $GID + 1 ))
             echo "${USER}:x:$uid:$GID::${DIR}:/bin/false" >> /etc/passwd
+        elif [ $(grep "Alpine Linux" /etc/os-release > /dev/null  && echo 1) ]; then
+            ${USERADD} "${USER}" -H "${DIR}" -s ${OSMYSHELL} -G "${GROUP}"
```
chemamartinez, line 87: Regarding the help message of `adduser` in Alpine Linux:
```
BusyBox v1.35.0 (2022-08-01 15:14:44 UTC) multi-call binary.

Usage: adduser [OPTIONS] USER [GROUP]

Create new user, or add USER to GROUP

	-h DIR		Home directory
	-g GECOS	GECOS field
	-s SHELL	Login shell
	-G GRP		Group
	-S		Create a system user
	-D		Don't assign a password
	-H		Don't create home directory
	-u UID		User id
	-k SKEL		Skeleton directory (/etc/skel)
```

You have to use `-h` instead of `-H` for the home directory. Otherwise, it will create a default one:
```
wazuh:x:100:101:Linux User,,,:/home/wazuh:/sbin/nologin
```

Apart from that, it works as expected.

chemamartinez (changes requested):

src/init/adduser.sh
FrancoRivero: I have changed that option to have the same behavior as other OS!

chemamartinez (approved):

chemamartinez (approved):

jmv74211 (approved):
## QA review

- **Type**: Manual testing.
- **Status**: Approved  🟢 
- **Testing issue**: https://github.com/wazuh/wazuh-qa/issues/3261
- **Comments**:  https://github.com/wazuh/wazuh-qa/issues/3261#issuecomment-1271266866

Issue #13803, opened 2022-06-09
Engine - Feature refinement through testing - Assets


gdiazlo:
| Wazuh version | Component | Install type | Install method   | Platform   |
| ------------- | --------- | ------------ | ---------------- | ---------- |
| 5.0.0         | Engine    | Manager      | any | Linux |

### Description 
We want to test all features we have developed so far to further polish its cohesion within the engine.
To do this we are going to write the following decoders:

- [x] #13864 (syslog / pam / ssh / su / sudo)
- [x] #13870
- [x] #13871
- [x] #13864 (nginx)
- [x] #14667
- [x] #14147
- [x] #14668
- [x] #14669
- [x] #14670
- [x] #14671
- [x] #14151
- [x] #14803 
- [x] #14829
- [x] #14840
- [x] #14974

For each of these devices we need to create a decoder which generates an ECS compliant event, and rules to alert about possible security threats.

We need to use:

* high level parser
* logical expressions
* function helpers
* mapping new fields
* kvdb

All decoders will become part of a single enviroment which will receive logs and will generate the corresponding events.

PR #14790, opened 2022-09-02
Engine update json parser

JavierBejMen:
|Related issue|
|---|
|Epic #13803 |

## Description
JSON parser parses all valid JSONs, including primitives types. This means that the JSON decoder will accept and decode every log starting with a `"string"` or plain number.

Normally, a log in JSON format or containing some JSON string will be an object, so we are proposing that the JSON parser will extract objects by default and allows other types with parametrization. 

Regardless if we use parametrization or multiple parsers we need the ability to parse specific types of JSON.

Additional fixes:

- Escaped issue: If in the literal matching there was a '\\' the parser was remaining block in an infinite loop

### DIFF
--- a/src/engine/source/hlp/src/specificParsers.cpp
+++ b/src/engine/source/hlp/src/specificParsers.cpp
@@ -5,6 +5,7 @@
 #include <string>
 #include <string_view>
 #include <unordered_map>
+#include <unordered_set>
 
 #include "date/date.h"
 #include "hlpDetails.hpp"
@@ -19,6 +20,32 @@
 #include <profile/profile.hpp>
 #include <rapidjson/stringbuffer.h>
 #include <rapidjson/writer.h>
+
+// Complementary functions for JSON parser //
+enum class JSON_TYPE
+{
+    J_ANY,
+    J_STRING,
+    J_BOOL,
+    J_NUMBER,
+    J_OBJECT,
+    J_ARRAY,
+    J_NULL
+};
+
+static const std::unordered_map<std::string, JSON_TYPE> jsonTypes = {
+    {"string", JSON_TYPE::J_STRING},
+    {"bool", JSON_TYPE::J_BOOL},
+    {"number", JSON_TYPE::J_NUMBER},
+    {"object", JSON_TYPE::J_OBJECT},
+    {"array", JSON_TYPE::J_ARRAY},
+    {"null", JSON_TYPE::J_NULL},
+    {"any", JSON_TYPE::J_ANY}};
+
+bool validJsonType(const std::string& type)
+{
+    return jsonTypes.find(type) != jsonTypes.end();
+}
 
 // TODO For all the rfc timestamps there are variations that we don't parse
 // still, this will need a rework on how this list work
@@ -39,7 +66,7 @@ static const std::unordered_map<std::string_view, std::tuple<const char*, const 
         {"RFC3154_ID_TIMEZONE", {"%b %d %R:%6S %Z", "Mar  1 18:48:50.483 UTC"}},
         {"RFC3154_TIMEZONE", {"%b %d %T %Z", "Mar  1 18:48:50 UTC"}},
         {"24H_LOCALTIME", {"%T", "15:16:01"}},
-        {"SYSLOG", {"%b %d %T", "Jun 14 15:16:01"}}, //TODO specify the RCF, rfc3164?
+        {"SYSLOG", {"%b %d %T", "Jun 14 15:16:01"}}, // TODO specify the RCF, rfc3164?
         {"ISO8601", {"%Y-%d-%mT%T%z", "2018-08-14T14:30:02.203151+02:00"}},
         {"ISO8601_Z", {"%Y-%d-%mT%TZ", "2018-08-14T14:30:02.203151Z"}},
         {"HTTPDATE", {"%d/%b/%Y:%T %z", "26/Dec/2016:16:22:14 +0000"}},
@@ -183,6 +210,43 @@ bool configureIgnoreParser(Parser& parser, std::vector<std::string_view> const& 
     return true;
 }
 
+bool configureJsonParser(Parser& parser, std::vector<std::string_view> const& args)
+{
+    // If no arguments parse object
+    if (args.empty())
+    {
+        parser.options.push_back("object");
+    }
+    else
+    {
+        // If one argument, it must be a valid json type or "any"
+        if (args.size() == 1)
+        {
+            if (validJsonType(std::string {args[0]}))
+            {
+                parser.options.push_back(std::string {args[0]});
+            }
+            else
+            {
+                const auto msg = fmt::format("[HLP] Invalid arguments for json Parser. "
+                                             "Expected one of [string, bool, number, "
+                                             "object, array, null, any], got [{}]",
+                                             args[0]);
+                throw std::runtime_error(msg);
+            }
+        }
+        else
+        {
+            // If more than one argument then is an error
+            const auto msg = fmt::format("[HLP] Invalid quantity of arguments for json "
+                                         "Parser. Expected one got [{}]",
+                                         args.size());
+            throw std::runtime_error(msg);
+        }
+    }
+    return true;
+}
+
 bool parseIgnore(const char** it, Parser const& parser, ParseResult& result)
 {
     auto start = *it;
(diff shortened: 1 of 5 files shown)


src/engine/source/base/utils/getExceptionStack.hpp
```diff
@@ -0,0 +1,37 @@
...
+    {
+        std::rethrow_if_nested(e);
+    }
+    catch (const std::exception& nestedException)
+    {
+        ss << getExceptionStack(nestedException, level + 1);
+    }
+    catch (...)
+    {
+    }
```
juliancnn, line 31: not catch all exceptions and should not be ignored

src/engine/source/cmds/include/cmds/cmdRun.hpp
```diff
@@ -0,0 +1,28 @@
...
+ * @param environment Name of the environment to be loaded.
+ */
+void run(const std::string& kvdbPath,
+         const std::string& endpoint,
+         const int queueSize,
+         const int threads,
+         const std::string& fileStorage,
+         const std::string& environment,
+         const int logLevel);
+} // namespace cmd
```
juliancnn, line 26: Using a structure to group configurations.

src/engine/source/cmds/src/cmdRun.cpp
```diff
@@ -0,0 +1,163 @@
+#include "cmds/cmdRun.hpp"
```
juliancnn, line 1: Why does the CMD module include the other modules and initialize them, I think it would be necessary to add this module to the design document and rename it. cmd is not just a parser, it is a new module in charge of initializing and configuring the  Engine.

juliancnn (changes requested):

src/engine/source/cmds/include/cmds/cmdRun.hpp
JavierBejMen: why?

src/engine/source/cmds/src/cmdRun.cpp
JavierBejMen: I don't understand, as you said cmds are not parsers, are commands. When the API is implemented this structure will probably change. They only build the Engine or the required functionality of it to perform the different tasks (run engine as in production, run engine in test mode and generate the graphs).

LucioDonda (approved):
LGTM

src/engine/source/hlp/src/specificParsers.cpp
```diff
@@ -20,6 +21,32 @@
...
+    J_ANY,
+    J_STRING,
+    J_BOOL,
+    J_NUMBER,
+    J_OBJECT,
+    J_ARRAY,
+    J_NULL
+};
+
+static const std::unordered_map<std::string, JSON_TYPE> jsonTypes = {
```
juliancnn, line 36: Why not?
```
static const std::unordered_map<std::string_view, JSON_TYPE> jsonTypes = {
```

src/engine/source/hlp/src/specificParsers.cpp
```diff
@@ -20,6 +21,32 @@
...
+    {"string", JSON_TYPE::J_STRING},
+    {"bool", JSON_TYPE::J_BOOL},
+    {"number", JSON_TYPE::J_NUMBER},
+    {"object", JSON_TYPE::J_OBJECT},
+    {"array", JSON_TYPE::J_ARRAY},
+    {"null", JSON_TYPE::J_NULL},
+    {"any", JSON_TYPE::J_ANY}};
+
+bool validJsonType(const std::string& type)
```
juliancnn, line 45: If the map is string_view, this prototype should change to string_view.

src/engine/source/hlp/src/specificParsers.cpp
```diff
@@ -183,6 +210,43 @@ bool configureIgnoreParser(Parser& parser, std::vector<std::string_view> const&
...
+    if (args.empty())
+    {
+        parser.options.push_back("object");
+    }
+    else
+    {
+        // If one argument, it must be a valid json type or "any"
+        if (args.size() == 1)
+        {
+            if (validJsonType(std::string {args[0]}))
```
juliancnn, line 225: If the map is a string_view it is not necessary to copy the memory and create a temporary value (which will be downloaded after its single use) from a string_view.

src/engine/source/hlp/src/specificParsers.cpp
```diff
@@ -309,18 +370,68 @@ bool parseJson(const char** it, Parser const& parser, ParseResult& result)
...
+                valid = true;
+            }
+            break;
+        case JSON_TYPE::J_NULL:
+            if (doc.IsNull())
+            {
+                valid = true;
+            }
+            break;
+    }
```
juliancnn, line 424: If the if condition is true, then `valid` is true, so then: 
```suggestion

switch (jsonTypes.at(parser.options[0]))
    {
        case JSON_TYPE::J_ANY: valid = true; break;
        case JSON_TYPE::J_STRING: valid = doc.IsString(); break;
        case JSON_TYPE::J_BOOL: valid = doc.IsBool(); break;
        case JSON_TYPE::J_NUMBER: valid = doc.IsNumber(); break;
        case JSON_TYPE::J_OBJECT: valid = doc.IsObject(); break;
        case JSON_TYPE::J_ARRAY: valid = doc.IsArray(); break;
        case JSON_TYPE::J_NULL: valid = doc.IsNull(); break;
    }
```

src/engine/test/source/hlp/hlp_test.cpp
```diff
@@ -241,8 +233,56 @@ TEST(hlpTests_IPaddress, IPV6_failed)
...
+{
+    const char* logQl = "<_json/json/object>";
+
+    const char* event = "{\"key1\":\"value1\",\"key2\":\"value2\"}";
+
+    auto parseOp = getParserOp(logQl);
+    ParseResult result;
+    bool ret = parseOp(event, result);
+
+    ASSERT_EQ(true, static_cast<bool>(parseOp));
```
juliancnn, line 258: Use `ASSERT_TRUE`

src/engine/test/source/hlp/hlp_test.cpp
```diff
@@ -241,8 +233,56 @@ TEST(hlpTests_IPaddress, IPV6_failed)
...
+    ASSERT_EQ("{\"key1\":\"value1\",\"key2\":\"value2\"}",
+              std::any_cast<JsonString>(result["_json"]).jsonString);
```
juliancnn, line 260: Why not ASSERT_SRTEQ to compare string?
A code like the following would be much easier to read and write:
```CPP
    ASSERT_TRUE(static_cast<bool>(parseOp));
    const auto expectedResult {R"({"key1":"value1","key2":"value2"})"};
    ASSERT_STREQ(expectedResult,
                 std::any_cast<JsonString>(result["_json"]).jsonString.data());
```

juliancnn (changes requested):

src/engine/source/hlp/src/specificParsers.cpp
```diff
@@ -20,6 +21,32 @@
...
+// Complementary functions for JSON parser //
+enum class JSON_TYPE
+{
+    J_ANY,
+    J_STRING,
+    J_BOOL,
+    J_NUMBER,
+    J_OBJECT,
+    J_ARRAY,
+    J_NULL
```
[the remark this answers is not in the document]
LucioDonda: I choose that because I couldn't use NULL as a member because it's a reserved keyword

src/engine/source/hlp/src/specificParsers.cpp
LucioDonda: ok, it could be an option

src/engine/source/hlp/src/specificParsers.cpp
LucioDonda: ok

src/engine/source/hlp/src/specificParsers.cpp
LucioDonda: got it

src/engine/source/hlp/src/specificParsers.cpp
LucioDonda: Nice!

src/engine/test/source/hlp/hlp_test.cpp
LucioDonda: got it

src/engine/test/source/hlp/hlp_test.cpp
LucioDonda: got it

JcabreraC (approved):
LGTM !

Issue #14754, opened 2022-08-29
Improve the ThrottlingException handling in the AWS module


CarlosRS9:
| Affected integration |
|---|
| AWS |

## Description
We need to review the AWS module to determine which parts of its code could be affected by a unhandled `ThrottlingException` in a similar way to what we need in https://github.com/wazuh/wazuh/issues/10754. It seems this affects to the CloudWath Logs implementation only, but we need to confirm it.

Regarding CloudWatchLogs, during https://github.com/wazuh/wazuh/issues/14718 we discovered that the `DescribeLogStreams` can raise `ThrottlingException` too and because of that we should update our module to expect and catch these exceptions.

## Tasks
- [x] Investigate whether this affects only cloudwatch logs or the whole AWS module
- [x] Add the Throttling exception handling to the `DescribeLogStreams` and any other affected function.
- [x] Unit tests without failures. Updated if there are any relevant changes.
- [x] Integration tests without failures. Updated if there are any relevant changes.
- [x] Update the documentation if necessary.
- [x] Add entry to the changelog if necessary.

fdalmaup:
# Issue Update

I have made some changes to the code, specifically in the `get_log_streams` method which is the one that uses `DescribeLogStreams` and raises the `ThrottlingException`. Manual tests have been carried out to verify it works as expected:
```
/var/ossec/wodles/aws/aws-s3 -sr cloudwatchlogs -g 4_4_test -d 2 -p dev -r us-east-1
DEBUG: +++ Debug mode on - Level: 2
DEBUG: +++ Getting alerts from "us-east-1" region.
DEBUG: only logs: None
DEBUG: Getting log streams for "4_4_test" log group
DEBUG: WARNING: The "describe_log_streams" request was denied due to request throttling. Attempting again.
DEBUG: Found "test_stream" log stream in 4_4_test
DEBUG: Getting data from DB for log stream "test_stream" in log group "4_4_test"
DEBUG: Token: "f/37060011712092978178900541551456153760331542018780004351/s", start_time: "1661828400000", end_time: "1661828400000"
DEBUG: Getting CloudWatch logs from log stream "test_stream" in log group "4_4_test" using token "f/37060011712092978178900541551456153760331542018780004351/s", start_time "1661828400001" and end_time "None"
DEBUG: Saving data for log group "4_4_test" and log stream "test_stream".
[...]

PR #14837, opened 2022-09-09
Improve AWS throttling exception error messages

CarlosRS9:
## Description

In https://github.com/wazuh/wazuh/issues/14754 we updated and improved the way we are handling the `Throttling` exceptions in our AWS module. 

In addition to that, we expanded our official documentation in https://github.com/wazuh/wazuh-documentation/pull/5559 to show how to configure the Retry option that AWS provides to avoid these exceptions.

Now, this PR expands upon that by improving the error messages we show to our users when this exception happens, adding a link to our official documentation where they will find how to configure the retry option, which as explained before is the feature used to avoid these type of issues.

### DIFF
--- a/wodles/aws/aws_s3.py
+++ b/wodles/aws/aws_s3.py
@@ -63,6 +63,7 @@ if sys.version_info[0] == 3:
 ################################################################################
 
 CREDENTIALS_URL = 'https://documentation.wazuh.com/current/amazon/services/prerequisites/credentials.html'
+RETRY_CONFIGURATION_URL = 'https://documentation.wazuh.com/current/amazon/services/prerequisites/considerations.html#Connection-configuration-for-retries'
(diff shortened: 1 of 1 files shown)


CarlosRS9:
> Please consider putting the message in a constant and reuse in each debug call. Something like that
> 
> ```python
> THROTTLING_MESSAGE = ('Error: The "{method_name}" request was denied due to request throttling. If the problem '
>                       'persists check the following link to learn how to use the Retry configuration to avoid it: {url}')
> 
> debug(THROTTLING_MESSAGE.format("method_name", RETRY_CONFIGURATION_URL), 2)
> ```

The change you suggest has been applied to the master branch. To avoid conflicts with master we have decided to leave this as it is. Once 4.4 is merged onto master both changes will be combined.

nico-stefani (approved):

davidjiglesias (approved):

Issue #14840, opened 2022-09-09
Engine - benchmark research and refinement 


JcabreraC:
| Wazuh version | Component | Install type | Install method   | Platform   |
| ------------- | --------- | ------------ | ---------------- | ---------- |
| 5.0.0         | Engine    | Manager      | Packages/Sources | OS version |

After the creation of a set of assets, it is necessary to test the engine results. The functionality of this issue is twofold: on the one hand, to check the amount of events per second that we are able to process; on the other hand, to refine and fix small bugs that may appear when testing the full functionality.

In this same issue we will be adding and fixing the problems derived from the benchmarks.

- [x] Benchmarks
- [x] Unprocessed logs: when sending sequences of logs, some of them are not processed and remain as:
```
{"wazuh":{"queue":49,"origin":"
```

#### Ignore parser:

- [x] Change default behavior, the parser ignore will be a new parser, and will not be the default parser in the '_' field.
- [x]  Create ignore custom parser and fix parser The pointer points one character back when the first match fails.
- [ ]  Create test 
- [x]  Adapat Actual assets
    
#### CSV parser:

- [x]  Create a CSV parser
- [x]  Create test a CSV parser
- [x]  Apply in postgrest decoders.

#### Wazuh-DB:
- [x] Testing
- [x] Fix issues
- [x] Update syscollector (4.4)

### Other fixes:

- [x] Fixed memory problems related to the protocol handler, which solved the problem of cut logs in the output.
- [X] Add ASAN to cmake
- [X] Fix memori issues related to ASAN (HLP + TEST)
- [X] Create script to run ASAN after compiling
- [X] Create script to run Valgrind (meme check) after compiling

nmkoremblum:
As part of the issue, Wazuh's SCA decoder has been tested, and some of the logs collected at the end of the operations chain (after being processed) are the following:

- Dump End:
```json
{"wazuh":{"queue":112,"origin":"sca"},"agent":{"id":"001","name":"10-111-amazon2-agent","registeredIP":"any"},"event":{"original":"{"type":"dump_end","policy_id":"cis_amazon_linux_2","elements_sent":183,"scan_id":1990336635}"},"_sca_event_json":{"type":"dump_end","policy_id":"cis_amazon_linux_2","elements_sent":183,"scan_id":1990336635},"decode_result":true}
```

- Summary:
```json
[...]

PR #14973, opened 2022-09-19
Engine refinement

juliancnn:
|Related issue|
|---|
| #14840 |

### Ignore parser:

- [x] Change default behavior, the parser ignore will be a new parser, and will not be the default parser in the '_' field.
- [x]  Create ignore custom parser and fix parser The pointer points one character back when the first match fails.
- [x]  Create test 
- [x]  Adapat Actual assets
    
Parser ignore, now changed. It will be a parser like any other and not a specific parser for the field '_'.
The parser can receive 0 or 1 argument. If it receives no arguments, it will ignore everything up to the endtoken, that is, up to the first character of the next parser.
When it receives an argument, it will ignore that string, consecutively, for example: <_/ignore/wazuh>123, it will match the following log: `wazuhwazuhwa123`.

### CSV parser:

- [x]  Create a CSV parser
- [x]  Create test a CSV parser
- [x]  Apply in postgrest decoders.

The CSV parser receives a variable number of arguments (minimum 1), each argument is the destination key of the extracted field. If the log has less fields than the specified number of arguments, the parser will fail. If it has the same amount or more, the parser will succeed, and will be positioned on the last extracted field, i.e. the next parser will succeed with the next log:

```
logql: <_custom/csv/field_1/field_2>,<_dummy>
log: h1,hi2,hi3,hi4 bye
```
Event result:

```json
{
    "_custom": {
        "field_1": "hi1",
        "field_2": "hi2"
    },
    "_dummy": "hi3,hi4 bye"
}
```

### Other fixes:

- [x] Fixed memory problems related to the protocol handler, which solved the problem of cut logs in the output.
- [X] Add ASAN to cmake
- [X] Fix memori issues related to ASAN (HLP + TEST)
- [X] Create script to run ASAN after compiling
- [X] Create script to run Valgrind (meme check) after compiling


LucioDonda (approved):
LGTM!

nmkoremblum (approved):
LGTM :)

JcabreraC (approved):
LGTM !

JavierBejMen (approved):
LGTM!!

Issue #14769, opened 2022-08-30
Agent is not raising `integrity_clear` logs


QU3B1M:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.4 | RSync | Agent | Packages | Linux/WindowsXP |

## Description
Working on the QA issue [3008](https://github.com/wazuh/wazuh-qa/issues/3008) when validating logs I found that the debug logs for the method `integrity_clear` is not appearing even though the method is working properly. This bug did not affect he functionality of the component, only debug logs.\
It was noticed that also the logs of `start sync` are not being raised.

## Evidences

https://user-images.githubusercontent.com/44633633/187469277-30322aab-3aed-4528-a486-739871333b0b.mp4

> Windows agent [ossec.log](https://github.com/wazuh/wazuh/files/9453889/ossec.log)

## Environment details
  | Module | VM Image | Extra Packages | Notes |
  | --------- | -------------- | -------- | -------- |
  | `wazuh-agent` | [dvgamerr/win-xp-sp3](https://app.vagrantup.com/dvgamerr/boxes/win-xp-sp3) | [Hotfix kb4024323](https://catalog.s.download.windowsupdate.com/d/csa/csa/secu/2017/06/windowsxp-kb4024323-x86-custom-enu_edb5a850bcbcc640c2b8c8aad3ab4d36a3b6fb26.exe) | |
  | `wazuh-manager`/`wazuh-agent` | [bento/centos-8](https://app.vagrantup.com/bento/boxes/centos-8) | [sqlite3](https://linuxhint.com/install-sqlite-centos-8/) - [python3](https://linuxize.com/post/how-to-install-python-on-centos-8/) |  |

## Steps to reproduce

<details><summary>Windows XP Hotfix case</summary>

On a Windows XP machine with a wazuh-agent and no hotfixes installed
   1. Edit the file `C:\Program Files\ossec-agent\local_internal_options.conf` to set the debug mode
       ```
      wazuh_modules.debug=2
      windows.debug=2
       ```
   2. Using the wazuh-ui `(C:\Program Files\ossec-agent\win32ui)` stop the wazuh-agent clicking _**manage -> stop**_
[...]

PR #14994, opened 2022-09-22
Agent is not raising `integrity_clear` logs - Implementation

pereyra-m:
|Related issue|
|---|
|#14769|

## Description

This PR enables the logs for the RSync module on runtime if the library is loaded with a new logger helper class (the UT were included.
Also, in the `startRSync()` method, both `integrity_check_global` and `integrity_clear` messages are logged.

It was required a new "full log" function to properly print the line, function and file of the logs in RSync. Otherwise, all the logs would have the same source.

## Logs/Alerts example

These are the logs generated by the implementation

```syslog
2022/09/22 21:47:48 loggerHelper[34732] rsyncImplementation.cpp:114 at startRSync(): DEBUG: Remote sync started: integrity_check_global
2022/09/22 21:47:48 loggerHelper[34732] rsyncImplementation.cpp:114 at startRSync(): DEBUG: Remote sync started: integrity_clear
```

These logs were generated in different environments:

- Manager

- Agent

- Win agent
- 

## Tests

  - [x] Linux
  - [x] Windows
- [x] Source installation


src/shared_modules/common/commonDefs.h
```diff
@@ -120,6 +120,18 @@ typedef struct
  */
 typedef void((*log_fnc_t)(const char* msg));
 
+/**
+ * @brief Callback function for user defined logging but with extra details.
```
MiguelazoDS, line 124: What are those details?

src/wazuh_modules/wm_syscollector.c
```diff
@@ -151,6 +152,15 @@ void* wm_sys_main(wm_sys_t *sys) {
...
         syscollector_sync_message_ptr = so_get_function_sym(syscollector_module, "syscollector_sync_message");
+
+        void* rsync_module = NULL;
+        if(rsync_module = so_check_module_loaded("rsync"), rsync_module) {
+            void (*rsync_initialize_ptr)(full_log_fnc_t) = so_get_function_sym(rsync_module, "rsync_initialize_log_function");
+            rsync_initialize_ptr(mt_log_wrapper);
+            #ifndef WIN32
+            so_free_library(rsync_module);
+            #endif
+        }
```
MiguelazoDS, line 163: So the logging only works if the rsync_lib is loaded. What would be the case that wouldn't happen?

src/wazuh_modules/wm_syscollector.c
```diff
@@ -151,6 +152,15 @@ void* wm_sys_main(wm_sys_t *sys) {
         syscollector_start_ptr = so_get_function_sym(syscollector_module, "syscollector_start");
         syscollector_stop_ptr = so_get_function_sym(syscollector_module, "syscollector_stop");
         syscollector_sync_message_ptr = so_get_function_sym(syscollector_module, "syscollector_sync_message");
+
+        void* rsync_module = NULL;
+        if(rsync_module = so_check_module_loaded("rsync"), rsync_module) {
+            void (*rsync_initialize_ptr)(full_log_fnc_t) = so_get_function_sym(rsync_module, "rsync_initialize_log_function");
+            rsync_initialize_ptr(mt_log_wrapper);
+            #ifndef WIN32
+            so_free_library(rsync_module);
```
MiguelazoDS, line 161: Why only for non-win32 systems?

src/wazuh_modules/wm_syscollector.c
```diff
@@ -151,6 +152,15 @@ void* wm_sys_main(wm_sys_t *sys) {
         syscollector_start_ptr = so_get_function_sym(syscollector_module, "syscollector_start");
         syscollector_stop_ptr = so_get_function_sym(syscollector_module, "syscollector_stop");
         syscollector_sync_message_ptr = so_get_function_sym(syscollector_module, "syscollector_sync_message");
+
+        void* rsync_module = NULL;
+        if(rsync_module = so_check_module_loaded("rsync"), rsync_module) {
+            void (*rsync_initialize_ptr)(full_log_fnc_t) = so_get_function_sym(rsync_module, "rsync_initialize_log_function");
```
MiguelazoDS, line 158: I would rename this as `void (*rsync_init_logging_ptr)(full_log_fnc_t) = so_get_function_sym(rsync_module, "rsync_init_logging");` to match the definition convention followed above for syscollector ptrs.

MiguelazoDS, line 158: What about using typedef? I imagine the assignation would end like this: 
`rsync_initialize_ptr = so_get_function_sym(rsync_module, "rsync_initialize_log_function");`

src/shared_modules/common/commonDefs.h
```diff
@@ -120,6 +120,18 @@ typedef struct
...
+ *
+ * @param log_type Log level (info, debug, etc.)
```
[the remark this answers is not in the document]
pereyra-m: Sure.
Fixed!

src/shared_modules/common/commonDefs.h
pereyra-m: Comment updated.

src/wazuh_modules/wm_syscollector.c
pereyra-m: Maybe I am not understanding your question. 
The `rsync_initialize_log_function` will only be called if the `rsync` module is found. 

On the other side, the `rsync_initialize_log_function` might not be available. I'm adding an extra check just in case.

src/wazuh_modules/wm_syscollector.c
pereyra-m: This is related to the previous `so_check_module_loaded()` call.

In Linux, the flag `RTLD_NOLOAD` for `dlopen()` increments the internal counter anyway. We need a matching `dlclose()`. See https://stackoverflow.com/questions/50570223/dlopenrtld-noload-still-returns-not-null-after-dlclose

The documentation doesn't say anything about it
https://linux.die.net/man/3/dlopen

But here, we have a mention about it
https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dlopen.3.html

```
     RTLD_NOLOAD     The specified image is not loaded.  However, a valid
                     handle is returned if the image already exists in the
                     process. This provides a way to query if an image is
                     already loaded.  The handle returned is ref-counted, so
                     you eventually need a corresponding call to dlclose()
```

src/wazuh_modules/wm_syscollector.c
pereyra-m: Like we already have the `rsync_initialize()` method, I selected `rsync_initialize_log_function()`.

To follow the convention, I'm writing

```c++
void (*rsync_initialize_log_function_ptr)(full_log_fnc_t) = so_get_function_sym(rsync_module, "rsync_initialize_log_function");
```

What do you think?

pereyra-m: I've used a **typedef** now, please check.

src/wazuh_modules/wm_syscollector.c
MiguelazoDS: Nevermind, it was strange at first for me this dynamic loading mechanism to implement a log function, I was just not aware of this prior development.

MiguelazoDS (dismissed):

tdrauncieal (approved):

MiguelazoDS (dismissed):

src/shared/sym_load.c
```diff
@@ -46,6 +46,22 @@ void* so_get_module_handle(const char *so){
 #endif
 }
 
+void* so_check_module_loaded(const char *so){
+#ifdef WIN32
+    char file_name[MAX_PATH] = { 0 };
+    snprintf(file_name, MAX_PATH-1, "%s.dll", so);
+    return GetModuleHandle(file_name);
```
MiguelazoDS, line 53: What about this? 
https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulehandlea

> The GetModuleHandle function returns a handle to a mapped module without incrementing its reference count. However, if this handle is passed to the [FreeLibrary](https://learn.microsoft.com/en-us/windows/desktop/api/libloaderapi/nf-libloaderapi-freelibrary) function, the reference count of the mapped module will be decremented. Therefore, do not pass a handle returned by GetModuleHandle to the FreeLibrary function. Doing so can cause a DLL module to be unmapped prematurely.

src/shared/sym_load.c
```diff
@@ -46,6 +46,22 @@ void* so_get_module_handle(const char *so){
...
+#else
+    char file_name[4096] = { 0 };
```
MiguelazoDS, line 55: Can we define this value outside this conditional?

src/shared_modules/utils/tests/loggerHelper_test.h
```diff
@@ -0,0 +1,45 @@
...
+
+        virtual void SetUp()
+        {
+            ssOutput.str("");
+        }
+
+        virtual void TearDown()
+        {
+            ssOutput.str("");
+        }
```
MiguelazoDS, line 43: Is this really needed?

src/shared/sym_load.c
pereyra-m: I added a comment in the header, no one should call `FreeLibrary` with this handle
https://github.com/wazuh/wazuh/blob/f265a7a969ab4eaa1422e71620669f605df57a8a/src/headers/sym_load.h#L23-L29

Do you think we should be more explicit?

src/shared/sym_load.c
pereyra-m: I followed the same convention than the rest of the functions defined here.
The thing is that in Windows, `MAX_PATH` is 260 by default, not 4096
https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

src/shared_modules/utils/tests/loggerHelper_test.h
pereyra-m: This extra step was added just to be more explicit, it isn't required.

MiguelazoDS (dismissed):

MiguelazoDS (approved):

matias-braida (dismissed):
Great job

src/shared/debug_op.c
```diff
@@ -575,3 +575,35 @@ char * win_strerror(unsigned long error) {
     return messageBuffer;
 }
 #endif
+
+void mt_log_wrapper(const char* log_type, const char *tag, const char * file, int line, const char * func, const char *msg, ...){
```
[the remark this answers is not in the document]
pereyra-m: I created a wrapper so `rsync` can call any type of log function (info, error, etc.) but passing only one.
To identify the type of log, I used a string because using an enumeration could:
- Create a dependency in the compilation of `rsync` by having to include that definition
- Or duplicate the code definition in both sides

I assumed the code penalty of the string comparison to avoid this situations and to make the code more clear.

But I'll make the modification. 
Thank you.

src/shared/debug_op.c
```diff
@@ -575,3 +575,35 @@ char * win_strerror(unsigned long error) {
...
+        _mtdebug1(tag, file, line, func, msg, args);
+    }
+    else if(0 == strncmp(log_type, "debug_verbose", strlen(log_type)))
+    {
+        _mtdebug2(tag, file, line, func, msg, args);
+    }
+    else if(0 == strncmp(log_type, "info", strlen(log_type)))
+    {
+        _mtinfo(tag, file, line, func, msg, args);
+    }
```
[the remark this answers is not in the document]
pereyra-m: I assumed that the logger class in `rsync` knows the right type of parameter. 
I can set a debug log by default when I apply the other suggestion.

src/headers/debug_op.h
```diff
@@ -38,6 +38,19 @@
...
+ * @brief Enumeration of the different log levels.
+ *
+ */
+typedef enum {
+    DEBUG_LOG = 0,
+    DEBUG_VERBOSE_LOG,
+    INFO_LOG,
+    WARNING_LOG,
+    ERROR_LOG
+} log_level_t;
```
matias-braida, line 51: The numeric constant could be ordered in the same order of importance.
I think DEBUG_VERBOSE_LOG is less important than DEBUG_LOG. So the enum order could be:
    DEBUG_VERBOSE_LOG = 0,
    DEBUG_LOG,
    INFO_LOG,
    WARNING_LOG,
    ERROR_LOG

src/shared_modules/utils/loggerHelper.h
```diff
@@ -0,0 +1,200 @@
...
+#include <unordered_map>
+
+enum class LogLevel : int
+{
+    DEBUG_LOG = 0,
+    DEBUG_VERBOSE_LOG,
+    INFO_LOG,
+    WARNING_LOG,
+    ERROR_LOG
+};
```
matias-braida, line 30: Just to force that constants values in C++ are the same as in C language, in this enum, I would force the value of each constant to the constant in C language enum

matias-braida (commented):
Good job

src/headers/debug_op.h
pereyra-m: This has been changed in the last commit.

src/shared_modules/utils/loggerHelper.h
pereyra-m: Could you provide an example? Remember that `rsync` is a separate library and the shared code of the C code isn't included.

tdrauncieal (approved):

MiguelazoDS (dismissed):

src/shared_modules/utils/loggerHelper.h
pereyra-m: Please, check the last changes.

matias-braida (commented):
GJ

src/shared_modules/rsync/src/rsync.cpp
```diff
@@ -216,6 +224,16 @@ void RemoteSync::initialize(std::function<void(const std::string&)> logFunction)
     }
 }
 
+void RemoteSync::initializeFullLogFunction(full_log_fnc_t debugVerboseFunction, full_log_fnc_t debugFunction,
+                                           full_log_fnc_t infoFunction, full_log_fnc_t warningFunction, full_log_fnc_t errorFunction )
+{
+    Log::debugVerbose.assignLogFunction(debugVerboseFunction, "rsync");
```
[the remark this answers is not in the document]
pereyra-m: Updated!

tdrauncieal (approved):

matias-braida (approved):

Dwordcito (approved):

Issue #14974, opened 2022-09-20
Engine - Replace yml library


JcabreraC:
| Wazuh version | Component | Install type | Install method   | Platform   |
| ------------- | --------- | ------------ | ---------------- | ---------- |
| 5.0.0         | Engine    | Manager      | Packages/Sources | OS version |

It is necessary to change the yml library we are using. When converting Yml types to Json it checks if the value can be transformed to the specific type, not if it is of that type. 

If you have a string with a number, as it is first checked if it can be transformed to integer it does it automatically:
https://github.com/wazuh/wazuh/blob/fb6950114987efcd266631933de517cfd22f9af4/src/engine/source/catalog/yml2Json.hpp#L20-L48

This makes it impossible to create a list of strings.

## Example
The string:
```yml
 _tmpSubjUserCodes: ["4657", "4670", "4672", "4673", "4674", "4688", "4689", "4697", "4698", "4699"]
``` 
Becomes:
```json
"_tmpSubjUserCodes": [
        4657,
        4670,
        4672,
        4673,
        4674,
        4688,
        4689,
        4697,
        4698,
        4699
]
```

JavierBejMen:
## Update

After further research, quoted scalar values are tagged with an exclamation symbol. This allows us to handle the ambiguity between numbers and string numbers.

No need to change the library.

Fix:
```c++
constexpr auto QUOTED_TAG = "!";
...
    if (QUOTED_TAG == node.Tag())
    {
        v.SetString(node.as<std::string>().c_str(), allocator);
    }
...
```

Issue #14829, opened 2022-09-08
Engine - Update json parser


JcabreraC:
| Wazuh version | Component | Install type | Install method   | Platform   |
| ------------- | --------- | ------------ | ---------------- | ---------- |
| 5.0.0         | Engine    | Manager      | any | Linux |

### Description 
JSON parser parses all valid JSONs, including primitives types. This means that the JSON decoder will accept and decode every log starting with a `"string"` or plain number.

Normally, a log in JSON format or containing some JSON string will be an object, so we are proposing that the JSON parser will extract objects by default and allows other types with parametrization. 

Regardless if we use parametrization or multiple parsers we need the ability to parse specific types of JSON.

Additional fixes:

- Escaped issue: If in the literal matching there was a '\\' the parser was remaining block in an infinite loop

Issue #14803, opened 2022-09-05
Engine - XML parser and windows assets adaption


JcabreraC:
| Wazuh version | Component | Install type | Install method   | Platform   |
| ------------- | --------- | ------------ | ---------------- | ---------- |
| 5.0.0         | Engine    | Manager      | Packages/Sources | OS version |

### XML parser:
Some events might have embedded XML strings in its
contents.
We need a parser that will parse that information into an auxiliary
field in JSON format.

The parser:
- Must have tests for valid and invalid cases, proving the desired behavior. 
  Any behavior not tested will be considered unsupported.
- Must have at least one synthetic benchmark. Ideally, we should have 
  benchmarks for success and error cases. These will be used in regression tests.
- Must document the usage of the supported context within all the assets.
- Might be used as function helpers.

Once the parser has been created, it is necessary to adapt the assets previously created in windows for its use.

Issue #14151, opened 2022-07-06
Engine - Windows events decoder


JavierBejMen:
| Wazuh version | Component | Install type | Install method   | Platform   |
| ------------- | --------- | ------------ | ---------------- | ---------- |
| 5.0.0         | Engine    | Manager      | any | Linux |

As part of #13803 we want to develop a Windows events decoder.

hitman28594:
Hi,

If the windows decoders are being updated, can we update the field name structure in the eventchannel.c decoder to be more inline with ecs?

Current: win.eventdata.{FIELDNAME}
Update: winlog.event_data.{FIELDNAME}

This would allow Wazuh Opensearch setup to integrate with Sigma rules out of the box!! This would be a huge improvement which would significantly help progress the development of the Wazuh ruleset.

But if that .c decoder isn’t possible to change, nxlog has already mapped windows fields to ecs.

There are many windows fields which are not mapped to ecs fields as part of this commit, many of the missing ones can be scraped from the following documentation:

https://docs.nxlog.co/userguide/integrate/elastic-common-schema.html

JavierBejMen:
Moved conversation to #14262

PR #15028, opened 2022-09-28
Engine features integration

JcabreraC:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/13803|

We want to test all features we have developed so far to further polish its cohesion within the engine.
To do this we are going to write the following decoders:

- [x] #13864 (syslog / pam / ssh / su / sudo)
- [x] #13870
- [x] #13871
- [x] #13864 (nginx)
- [x] #14667
- [x] #14147
- [x] #14668
- [x] #14669
- [x] #14670
- [x] #14671
- [x] #14151
- [x] #14803 
- [x] #14829
- [x] #14840
- [x] #14974

For each of these devices we need to create a decoder which generates an ECS compliant event, and rules to alert about possible security threats.

We need to use:

* high level parser
* logical expressions
* function helpers
* mapping new fields
* kvdb

All decoders will become part of a single enviroment which will receive logs and will generate the corresponding events.


JavierBejMen (approved):
LGTM!!

JcabreraC:
LGTM !

Issue #15036, opened 2022-09-28
Engine - API - Callback registration and protocol handler


juliancnn:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0| Engine| Manager| Packages/Sources | OS version |

It is necessary to create the API base, this first iteration will not be related to the connection protocols, nor with socket handlers, nor threads, but it will define the way of interconnection between the modules and the API:

- [x] A callback registry should be provided, where modules can register the commands they accept and a function callback to process them.
- [x] Validate a string or a json (to be obtained by a request to the API) against the protocol specified in the issue https://github.com/wazuh/wazuh/issues/5934
- [x]  Provide an interface so that modules can create their callback function and always be compliant with the response protocol specified in the issue #5934

In addition, this contemplates:
- [ ] Tests
- [x] Thread-Safe implementation

This issue will allow to separate the implementation of the server with the execution of the API and its relationship with the other modules.
![Untitled drawing](https://user-images.githubusercontent.com/1891958/192895922-0780def3-589e-4383-b06d-0d80d62d0b5e.jpg)

Issue #5934, opened 2020-09-07
Define communication protocol specifications for wazuh daemons


davidjiglesias:
Hello team,

This issue is part of https://github.com/wazuh/wazuh/issues/5927.

It aims to study, identify and define the design for a communication protocol to be used for Wazuh daemons. We should ensure new designs follow the same structure when communicating.

The new protocol should define the structure and fields of the response, as well as the error when it occurs.

Best regards,

David J. Iglesias

vicferpoy:
# Socket response proposal

We are aiming to standarize every Wazuh socket response. The purpose of it is to be able to parse every response in the same way, enabling a future microservices architecture which would improve scalability and ease future developments.

## Structure

The response must be a JSON type object with the following **required** fields:

- `error`: an **integer** equal or greater than 0.
    This field must be **0** when the response was a __successs__. Otherwise, it must be an actual and unique code error.

- `data`: either a **JSON object** or a **list**. Every module must decide which type of __data__ will return. It must be consistent, ignoring the number of elements.
    This field must contain any useful information that could be parsed. This is why it is expected to be a **JSON object** or a list of these. Every module will have its own __data__ structure. `data` structures which are a list must have its own **error** field since any element could have a different error/exception. The external field `error` must be a general code.

- `message`: an __optional__ **string**.
    It will contain a message with human readable information about the operation status. Although **its value is optional**, this field must be present in the response.


## Examples

### Successful response (JSON data)

```json
{
    "error": 0,
    "data": {
        "id": "016",
        "name": "agent-test-authd",
        "ip": "any",
        "key": "5f806e70652ca72b2e65f679f159d813a312c5f8b057128ef9ce212f679309a7"
    },
    "message": "Successful"
}
```

### Error response

```json
{
    "error": 1045,
    "data": {
        "id": "016"
    },
    "message": "Error trying to register an agent"
}
```

### Successful response (data list)

```json
{
    "error": 0,
    "data": [
        {
            "agent": 1,
            "task_id": 3,
            "module": "upgrade_module",
            "command": "upgrade",
            "status": "Updated",
            "create_time": "2020/09/03 13:56:04",
            "update_time": "2020/09/03 13:57:14"
        },
        {
            "agent": 3,
            "task_id": 5,
            "module": "upgrade_module",
            "command": "upgrade",
            "status": "Updated",
            "create_time": "2020/09/03 13:58:04",
            "update_time": "2020/09/03 13:59:14"
        }
    ],
[...]

PR #15043, opened 2022-09-29
Engine - API - Callback registration and protocol handler

juliancnn:
|Related issue|
|---|
| #15036  |

According to isue #15036, this PR implements:

- [x] A callback registry should be provided, where modules can register the commands they accept and a function callback to process them.
- [x] Validate a string or a json (to be obtained by a request to the API) against the protocol specified in the issue https://github.com/wazuh/wazuh/issues/5934
- [x]  Provide an interface so that modules can create their callback function and always be compliant with the response protocol specified in the issue #5934

In addition, this contemplates:
- [x] Tests
- [x] Thread-Safe implementation

### DIFF
--- a/src/engine/CMakeLists.txt
+++ b/src/engine/CMakeLists.txt
@@ -137,6 +137,7 @@ target_include_directories(main
     ${CLI11_SOURCE_DIR}/include
 )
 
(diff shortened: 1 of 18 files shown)


JavierBejMen (approved):
LGTM

JavierBejMen (approved):
LGTM

Issue #14982, opened 2022-09-21
Error while upgrading an agent in Amazon Linux 2022 (through WPK)


mauromalara:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.3.8 | WPK Upgrade | Manager | Packages | Amazon Linux 2022 |

## Description

An error was thrown while upgrading a Wazuh agent from `4.3.7` to `4.3.8` (using a Wazuh manager 4.3.8). Also, it was thrown during the manual testing performed in: https://github.com/wazuh/wazuh-qa/issues/3266 (upgrading from `4.3.7`, which was the latest version at that time, to `4.4.0`)

The OS architecture of the Wazuh agent is: `aarch64`, and the Wazuh manager architecture is: `x86_64`.

Also, this does not happen when using a `x86_64` agent package (the upgrade is made correctly).

## Steps to reproduce the error

<details><summary>Follow these steps to get the error message</summary>

1. Install the Wazuh manager (`v4.3.8` - current): 
```
rpm -ivh https://packages.wazuh.com/4.x/yum/wazuh-manager-4.3.8-1.x86_64.rpm
systemctl daemon-reload
systemctl enable wazuh-manager
systemctl start wazuh-manager
```
2. Install the Wazuh agent (`v4.3.7`):
```
rpm -ivh https://packages.wazuh.com/4.x/yum/wazuh-agent-4.3.7-2.aarch64.rpm
systemctl daemon-reload
systemctl enable wazuh-agent
sed -i 's/MANAGER_IP/172.31.83.218/g' /var/ossec/etc/ossec.conf                                                                      systemctl start wazuh-agent
```

<br>

> **The following steps should be followed in the agent (because this VM has the aarch64 architecture)**

3. Install the Python package:
```
python3 -m pip install cryptography
```
4. Install the dependencies:
```
yum install make gcc gcc-c++ python3 python3-policycoreutils automake autoconf libtool openssl-devel cmake -y; yum-config-manager --enable powertools; yum install libstdc++-static -y
```
5. Download and extract the latest version (commit 88bf15d at this moment):
```
curl -sL https://github.com/wazuh/wazuh/tarball/88bf15d2cbb2040e197e34a94dda0f71f607afad | tar zx
```
6. Modify the `./wazuh-wazuh-88bf15d/etc/preloaded-vars.conf` file uncommenting the following lines:
```
USER_LANGUAGE="en"
USER_NO_STOP="y"
[...]

PR #15151, opened 2022-10-11
Obtaining the 'aarch64' architecture in the manager and thus fixing the WPK upgrade. 

MarcelKemp:
|Related issue|
|---|
|#14982|

## Description

The problem, found in issue #14982, was due to the fact that the manager could not correctly obtain the `aarch64` architecture that the agent had, so the upgrade failed because it appeared with a `NULL` value.

This bug can be reproduced using a manager with any version that does not have this fix, and also the architecture of the agent that you want to upgrade is `aarch64`. 

With the change made in this PR, the manager now correctly detects the architecture (_aarch64_) of the agent, and this allows the WPK upgrade to be performed correctly.

> The change simply adds the “new” architecture to the array so that it matches and sets the variable correctly.

## Configuration options

The steps to set up the correct configuration for testing can be found in the description of issue #14982.

## Logs/Alerts example

With the fix, we can see how the manager saves the value of _`aarch64`_ in the variable `arch`, inside the `global.db`:
```
# sqlite3 /var/ossec/queue/db/global.db "SELECT OS_NAME, OS_MAJOR, OS_MINOR, OS_ARCH FROM AGENT WHERE ID = 2" -table
+--------------+----------+----------+---------+
|   os_name    | os_major | os_minor | os_arch |
+--------------+----------+----------+---------+
| Amazon Linux | 2022     |          | aarch64 |
+--------------+----------+----------+---------+
```

Then we see how after getting the value of `arch` and applying the same steps as in the issue to create the WPK for _arm64_, the agent is upgraded correctly:
```
# /var/ossec/bin/agent_upgrade -a 002 -f /tmp/myagent.wpk -x ./upgrade.sh

Upgrading...

Upgraded agents:
	Agent 002 upgraded: Wazuh v4.4.0 -> Wazuh v4.4.0
```

In addition, I attach the _upgrade.log_ file after the successful upgrade:
[upgrade.log](https://github.com/wazuh/wazuh/files/9754639/upgrade.log)

## Tests

  - [x] Linux
- [x] Source installation
- [x] Package installation
- [x] Source upgrade
- [x] Package upgrade

- Memory tests for Linux
  - [x] Scan-build report
  - [ ] Coverity
  - [ ] AddressSanitizer

### DIFF
--- a/src/shared/remoted_op.c
+++ b/src/shared/remoted_op.c
@@ -22,7 +22,7 @@
  * @retval A string pointer to the architecture. NULL if not found.
  */
 char * get_os_arch(char * os_header) {
-    const char * ARCHS[] = { "x86_64", "i386", "i686", "sparc", "amd64", "i86pc", "ia64", "AIX", "armv6", "armv7", NULL };
+    const char * ARCHS[] = { "x86_64", "i386", "i686", "sparc", "amd64", "i86pc", "ia64", "AIX", "armv6", "armv7", "aarch64", NULL };
     char * os_arch = NULL;
     int i;
 


chemamartinez (approved):
LGTM

chemamartinez (approved):

jmv74211 (approved):
## QA review

- **Type**: Manual testing.
- **Status**: Approved  🟢 
- **Testing issue**: https://github.com/wazuh/wazuh-qa/issues/3415
- **Comments**:  https://github.com/wazuh/wazuh-qa/issues/3415#issuecomment-1293556211

Issue #10696, opened 2021-10-29
Add SCA Policy for HP-UX B.11.31


dariommr:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.2.3-rev | SCA | Agent | Packages | HP-UX B.11.31 |

Hello Team,
I am reaching you with this request for Security Assessment Policies for HP-UX B.11.31.

The only document I found is this, and it is from 2009 and for version 11i: https://www.cisecurity.org/wp-content/uploads/2017/04/CIS_HP-UX_11i_Benchmark_v1.5.0.pdf

I hope this could be helpful.

There will be two different policies for the same OS, one will use the bastille tool from HP, the other one will use standard UNIX commands

jcruzlp:
Branch for tracking the changes and progress: https://github.com/wazuh/wazuh/tree/10696-hpux-sca

maumrsms:
Hello team,
I've tried to add the policies worked here but they don't seem to work:
![image](https://user-images.githubusercontent.com/29696136/152064399-7e2a24f8-3482-4346-a61e-138329848833.png)
![image](https://user-images.githubusercontent.com/29696136/152064407-67dee933-7190-424b-972c-1e5d507f8349.png)
![image](https://user-images.githubusercontent.com/29696136/152064429-68c83408-e346-478a-a23a-396307b6bd7d.png)
![image](https://user-images.githubusercontent.com/29696136/152064444-dbc397fe-4d21-4b54-bb97-31355146d9de.png)

Now checking the rules I noticed we're using some Bastille related files. For example:
![image](https://user-images.githubusercontent.com/29696136/152065889-75efc2e1-1e5c-4a07-888d-fad46a82a994.png)

I haven't seen any mention about this nor its configuration (I'm not even sure if Bastille comes out-of-the-box in HP-UX).
I imagine that some specific configuration would be needed on it too (files ownership in `/var/opt/sec_mgmt/bastille/log/Assessment` at least)
Could you please provide more information about this?

maumrsms:
Hello team,

I've just confirmed that this solution requires Bastille to be installed in `/opt/sec_mgmt/bastille/` for the Agent to be able to create it source files in `/var/opt/sec_mgmt/bastille/log/Assessment`.

I managed to replicate the situation in a different box after renaming the Bastille binary (`/opt/sec_mgmt/bastille/bin/bastille`):
[...]

PR #15157, opened 2022-10-11
Added HPUX 11i SCA policies for bastilles and without bastille

72nomada:
|Related issue|
|---|
|#10696|

Create HP UX 11i SCA policies for a scenario with Bastille and a scenario without Bastille.

There is no current CIS Benchmark document. The older found is from 2009 and for version 11i: https://www.cisecurity.org/wp-content/uploads/2017/04/CIS_HP-UX_11i_Benchmark_v1.5.0.pdf

QA issue:
- https://github.com/wazuh/wazuh-qa/issues/3387

Packages issue:
- https://github.com/wazuh/wazuh-packages/issues/1881


fabamatic (approved):
LGTM

jmv74211 (approved):
## QA review

- **Type**: Manual testing.
- **Status**: Approved  🟢 
- **Testing issue**: https://github.com/wazuh/wazuh-qa/issues/3387
- **Comments**:  https://github.com/wazuh/wazuh-qa/issues/3387#issuecomment-1310178022

jmv74211:
### Observations

There are some checks in which the remediation fields and the rules themselves are inconsistent, indicating a different remediation than the rule commands themselves. For example:

**Check**: Activate service.
**Remediation**: Activate service.
**Rule**: Check that the service is deactivated.

This is so, because the check itself recommends that the service is only activated if it is really necessary, but nevertheless it must be given as failed in the case of being enabled, assuming that the user is aware of the use of such service. In addition, there are other checks in the policy itself that would conflict, because they expect a set of services to be disabled in order to give `PASSED` checks.

By this is meant, that the logical and user-safe implementation has been carried out, but that the check itself is inconsistent.

Issue #14738, opened 2022-08-25
Engine - Filebeat, mappings and ECS


gdiazlo:
| Wazuh version | Component | Install type | Install method   | Platform   |
| ------------- | --------- | ------------ | ---------------- | ---------- |
| 5.0.0         | Engine    | Manager      | any | Linux |

Today, Wazuh delivers analyzed events to wazuh-indexer using filebeat and a wazuh module which maps custom wazuh fields to elastic types.

We need to modify how we use filebeat to deliver events to wazuh-indexer. 

- Research how elastic do their mappings from the ECS to elastic search types. 
- Do not use default conversions, we must use always explicit mapping.
- We need to support opendistro, opensearch and elasticsearch versions, which constrain the filebeat version we're able to use. This means we might be unable to use the [libbeats](https://github.com/elastic/beats/tree/main/libbeat/ecs) ECS support.

We need to automate the schema and mapping generation from the original ECS schema. So the maintenance of this module is minimal.

The engine will implement this by itself in the future, eliminating the need for filebeat, but in this phase, we need to be able to interoperate with Wazuh as it is today.

At the end of this issue, we will have a filebeat configuration and a set of mappings that are able to deliver events from the new engine to wazuh-indexer, and all other variants.

Please take a look at the [ECS repository ](https://github.com/elastic/ecs/tree/main/generated). ECS mappings, etc are already generated. We need to update them with our own fields and extensions.

PR #15211, opened 2022-10-21
Engine - filebeat mappings ecs engine

JcabreraC:
|Related issue|
|---|
|#14738|

## Description
Today, Wazuh delivers analyzed events to wazuh-indexer using filebeat and a wazuh module which maps custom wazuh fields to elastic types.

We need to modify how we use filebeat to deliver events to wazuh-indexer. 

- Research how elastic do their mappings from the ECS to elastic search types. 
- Do not use default conversions, we must use always explicit mapping.
- We need to support opendistro, opensearch and elasticsearch versions, which constrain the filebeat version we're able to use. This means we might be unable to use the [libbeats](https://github.com/elastic/beats/tree/main/libbeat/ecs) ECS support.

We need to automate the schema and mapping generation from the original ECS schema. So the maintenance of this module is minimal.

The engine will implement this by itself in the future, eliminating the need for filebeat, but in this phase, we need to be able to interoperate with Wazuh as it is today.

At the end of this issue, we will have a filebeat configuration and a set of mappings that are able to deliver events from the new engine to wazuh-indexer, and all other variants.

Please take a look at the [ECS repository ](https://github.com/elastic/ecs/tree/main/generated). ECS mappings, etc are already generated. We need to update them with our own fields and extensions.


JavierBejMen (approved):
LGTM!

nmkoremblum (approved):
LGTM

LucioDonda (approved):
LGTM

JcabreraC (commented):
LGTM !

Issue #14792, opened 2022-09-02
Engine - API module


gdiazlo:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0| Engine| Manager| Packages/Sources | OS version |

We want to expose functionality from the engine to other pieces of Wazuh and users through an API.

This API will work as a drop-in replacement for the current analysis implementation. The initial approach for its creation comprehend: 

- [x] Create a new endpoint in the server
- [x] Create a new protocol handler to handle the internal Wazuh API protocol
- [x] Create an API manager class inside the engine that will implement all external API calls.

The API class should work independently of the transport used to send the messages, so any protocol information must be handled in the protocol handler layer. We will use the Wazuh internal API today, but we will want to support other protocols such as HTTP in the future.

The next diagram tries to represent the idea:

```mermaid
graph TD;
    WC[Wazuh Client]-->WP[Wazuh protocol];
    WP-->S[Server];
    S-->PH[Protocol Handler 1 ];
    S-->PH2[Protocol Handler N]
    PH-->API[API manager];
   PH2-->API[API manager];
    HC[HTTP Client]-->HTTP[HTTP];
    HTTP-->S;
```

To implement the first interaction of the API, we also need to:

- [x] Analyze the current analisysd API and document it
- [x] Design the API manager to support what we want to achieve in the future as detailed in the design document, but implement just the required methods to impersonate analysis

We will update this issue as we work on the solution.

| Dev branch |
| --- | 
| [14792-Engine-api](https://github.com/wazuh/wazuh/tree/14792-Engine-api) |

PR #15213, opened 2022-10-21
Engine API

JcabreraC:
|Related issue|
|---|
|#14792|

We want to expose functionality from the engine to other pieces of Wazuh and users through an API.

This API will work as a drop-in replacement for the current analysis implementation. The initial approach for its creation comprehend: 

- Create a new endpoint in the server
- Create a new protocol handler to handle the internal Wazuh API protocol
- Create an API manager class inside the engine that will implement all external API calls.

The API class should work independently of the transport used to send the messages, so any protocol information must be handled in the protocol handler layer. We will use the Wazuh internal API today, but we will want to support other protocols such as HTTP in the future.

The next diagram tries to represent the idea:

```mermaid
graph TD;
    WC[Wazuh Client]-->WP[Wazuh protocol];
    WP-->S[Server];
    S-->PH[Protocol Handler 1 ];
    S-->PH2[Protocol Handler N]
    PH-->API[API manager];
   PH2-->API[API manager];
    HC[HTTP Client]-->HTTP[HTTP];
    HTTP-->S;
```


juliancnn (approved):
LGTM!

PR #15284, opened 2022-11-02
Add agentlessd UT

tdrauncieal:
### Related Issue
N / A
### Description
As part of the goals of my income add UT for the agentlessd module (updated)
### New coverage
        CMakeLists.txt
        agentlessd/CMakeLists.txt
        agentlessd/test_agentlessd.c

### Output
Screenshot from 2022-11-08 15-52-36


src/unit_tests/agentlessd/test_agentlessd.c
```diff
@@ -0,0 +1,157 @@
...
+    lessdc.entries[0]->type = NULL;
+    lessdc.entries[0]->frequency = 86400;
+    lessdc.entries[0]->state = 0;
+    lessdc.entries[0]->current_state = 0;
+    lessdc.entries[0]->port = 0;
+    lessdc.entries[0]->error_flag = 0;
+
+    expect_StartMQ_call(DEFAULTQUEUE, WRITE, 0);
+
+    Agentlessd();
```
pereyra-m, line 141: Just a question: can we confirm that the line `494,17:                 run_periodic_cmd(lessdc.entries[i], test_it);` isn't executed during the tests?

src/unit_tests/agentlessd/test_agentlessd.c
```diff
@@ -0,0 +1,157 @@
...
+    expect_any(__wrap_fprintf, __stream);
+    expect_string(__wrap_fprintf, formatted_msg, "type: diff\n");
+    will_return(__wrap_fprintf, 0);
+
+    expect_fclose((FILE *)2, 0);
+    // save_agentless_entry() call
+
+    expect_fclose((FILE *)1, 0);
+
+    gen_diff_alert(HOST, SCRIPT, ADT);
```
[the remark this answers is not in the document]
tdrauncieal: this function always return 0, even out with error or not.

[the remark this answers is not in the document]
tdrauncieal: hwo to proced in these cases?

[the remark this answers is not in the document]
pereyra-m: Well, that seems to be a problem of the method and we won't fix that in this PR. I didn't noticed that.

But we should check the returned value anyway, because if the tested function changes in the future, we'll force the developer to check and update the UT too.

src/unit_tests/agentlessd/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+target_link_libraries(OS_AGENTLESSD_O ${WAZUHLIB} ${WAZUHEXT} -lpthread)
+
+#include wrappers
+include(${SRC_FOLDER}/unit_tests/wrappers/wazuh/shared/shared.cmake)
+
+list(APPEND agentlessd_names "test_agentlessd")
+#list(APPEND agentlessd_flags "-Wl,--wrap,Privsep_GetUser -Wl,--wrap,Privsep_GetGroup ${DEBUG_OP_WRAPPERS} \
+#                                -Wl,--wrap,StartMQ -Wl,--wrap,CreateThread ${STDIO_OP_WRAPPERS}")
+list(APPEND agentlessd_flags "${DEBUG_OP_WRAPPERS} ${STDIO_OP_WRAPPERS} \
+                            -Wl,--wrap,StartMQ -Wl,--wrap,CreateThread -Wl,--wrap,fprintf \
```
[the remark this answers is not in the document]
tdrauncieal: You say me that do not use a  'global' CreateThread mock, instead should be use a local create_thread(), like even see in analysisd UT?

src/unit_tests/agentlessd/test_agentlessd.c
```diff
@@ -0,0 +1,157 @@
...
+
+    expect_StartMQ_call(DEFAULTQUEUE, WRITE, 0);
+
+    Agentlessd();
+
+    os_free(lessdc.entries[0]);
+    os_free(lessdc.entries);
+}
+
+int main(void) {
```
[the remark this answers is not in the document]
tdrauncieal: Ok add new wrong cases.

src/unit_tests/agentlessd/CMakeLists.txt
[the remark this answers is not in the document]
pereyra-m: I mean that we have added the wrappers for threads, but we aren't using them. 

> instead should be use a local create_thread()

If there is a global wrapper, we shouldn't create a local one. Even when we see it as an example.

Now I see that all this is because the global wrapper (in `wazuh/src/unit_tests/wrappers/wazuh/shared/pthreads_op_wrappers.c`) isn't really well defined

```
int __wrap_CreateThread(__attribute__((unused)) void * (*function_pointer)(void *),
                        __attribute__((unused))  void *data) {
    return 1;
}
```
So you can ignore this comment.
Thanks.

src/unit_tests/agentlessd/test_agentlessd.c
tdrauncieal: I don't think so, but i'll control it

MiguelazoDS (dismissed):

MiguelazoDS:
I'd just create an issue for this PR

src/agentlessd/agentlessd.h
```diff
@@ -20,7 +20,11 @@
 /** Prototypes **/
 
 /* Main monitord */
+#ifdef WAZUH_UNIT_TESTING
+void Agentlessd(void);
+#else
 void Agentlessd(void) __attribute__((noreturn));
+#endif
```
jnasselle, line 27: Design question: why?

src/agentlessd/agentlessd.c
```diff
@@ -497,10 +501,13 @@ void Agentlessd()
...
             sleep(i);
         }
-
+#ifdef WAZUH_UNIT_TESTING
+        break;
+#else
         /* We only check every minute */
         test_it = 0;
         sleep(60);
+#endif
```
jnasselle, line 510: Design nitpick: check `FOREVER()` implementation to avoid real infinite loops handling on UT

src/agentlessd/agentlessd.c
```diff
@@ -8,6 +8,10 @@
  * Foundation
  */
 
+#ifdef WAZUH_UNIT_TESTING
+#define static
+#endif
+
```
jnasselle, line 14: Desing severe: replacing a C keyword with nothing is not recommended. Check `STATIC` usage on other already implemented UT

src/unit_tests/agentlessd/test_agentlessd.c
```diff
@@ -0,0 +1,205 @@
...
+    return OS_SUCCESS;
+}
+
+#define HOST "host"
+#define SCRIPT "script"
+#define AGTTYPE "agttype"
+#define MSG "msg"
+#define STR_HELPER(s) #s
+#define STR(s) STR_HELPER(s)
+#define ADT 1667392718
```
jnasselle, line 51: Complexity nitpick: defines/includes/global variables in an ordered way will improve readability!

jnasselle (changes requested):

src/agentlessd/agentlessd.c
tdrauncieal: ok I didn't saw, and it is a better implementation

src/agentlessd/agentlessd.c
tdrauncieal: maybe wouldn't be the optimal solution, but is implemented in every source code that needs a UT.

src/agentlessd/agentlessd.h
tdrauncieal: the `noreturn` attribute used in a real environment makes the function out an unpredictable result because don't expect to continue running, anyway this trick is used in another UT.

src/agentlessd/agentlessd.c
jnasselle: Only 5 UT use this approach. The other ones use the next one
```c
#ifdef WAZUH_UNIT_TESTING
// Remove STATIC qualifier from tests
#define STATIC
#else
#define STATIC static
#endif
```

src/unit_tests/agentlessd/test_agentlessd.c
```diff
@@ -0,0 +1,208 @@
...
+    int rc = save_agentless_entry(HOST, SCRIPT, AGTTYPE);
+    assert_int_equal(rc, 0);
+}
+
+void test_Agentlessd_ok(void **state) {
+    (void) state;
+
+    os_calloc(2, sizeof(agentlessd_entries *), lessdc.entries);
+    os_calloc(1, sizeof(agentlessd_entries), lessdc.entries[0]);
+    lessdc.entries[1] = NULL;
```
jnasselle, line 166: Design nitpick: this is no longer needed with new calloc change :+1:

jnasselle (approved):
Nice work @tdrauncieal ! you are ready to go :rocket: !

Damian-Mangold (approved):

pereyra-m (approved):

Issue #15312, opened 2022-11-07
Engine - Reorganization of the ruleset


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0 | Engine | - | - | - |

It is necessary to reorganize the ruleset of the engine, because as the number of rules and decoders grows, we need a better division for a quick search of them. In addition, metadata will be added to existing decoders.

We will take advantage of this issue also to make a recursive call to the subfolders when we make a call to load assets from a directory. We will use the `-r` option for this. The API call would look like this:
```
./main catalog load -r decoder|rules|filter|output PATH
```

## Tasks
- [x] Implement functionality
- [x] Update documentation
- [x] QA validate

PR #15314, opened 2022-11-07
Ruleset reorganization and recursive decoder loading

JcabreraC:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/15312|

In this PR, the ruleset will be reorganized for better reading. In addition, metadata will be added to the current decoders.
Finally, the `-r` option has been created for the recursive call when loading a directory. The syntax is:
```
./main catalog load -r decoder|rules|filter|output PATH
```

## Tasks
- [x] Implement functionality
- [x] Update documentation
- [x] QA validate


src/engine/source/main.cpp
```diff
@@ -343,6 +344,9 @@ void configureSubCommandCatalog(std::shared_ptr<CLI::App> app)
             "path", args::catalogPath, "Path to the directory containing the item files.")
         ->required()
         ->check(CLI::ExistingDirectory);
+    load_subcommand
+        ->add_flag(
+            "-r, --recursive", args::catalogRecursive, "Recursive loading of the directory.");
```
LucioDonda, line 349: if not required should it have a default value or is false taken as default?

LucioDonda (approved):
LGTM

JavierBejMen (approved):

src/engine/source/main.cpp
JcabreraC: The `-r` option is a flag. It is used to load the decoders of a folder recursively. If this option is not set, by default it loads only the decoders of that folder in the first level.

Issue #15439, opened 2022-11-18
Engine - Support multiple KVDB instances


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0 | Engine | - | - | - |

Currently, when running the engine and executing the test, it indicates that the KVDBs are blocked since it uses the same ones as the engine.

It is necessary to support multiple KVDB instances to be able to test on the current environment.

## Tasks
- [ ] Implement functionality
- [ ] Update Documentation
- [ ] QA validate

PR #15444, opened 2022-11-18
Support multiple instances of test mode (KVDB temporary)

juliancnn:
|Related issue|
|---|
|#15439|

## Description

With the intention of supporting multiple instances of test mode, and furthermore, not affecting the production environment, this PR takes the approach of cloning the RocksDB databases from the production environment to a temporary environment, allowing to modify these KVDBs in the future without altering the production environment. It also allows multiple instances of test mode running in parallel, even if the production environment is also running.


JcabreraC:
This PR is closed in favor of https://github.com/wazuh/wazuh/pull/15450

PR #15450, opened 2022-11-21
Changed default engine logtest path

JcabreraC:
|Related issue|
|---|
|#15439|

## Description

After several tests, it has been decided to change the default path of the engine test environment, so that it does not affect the KVDBs in the production environment.

Note that, by default, the directory of the test KVDBs is `/var/ossec/etc/kvdb_test/`, while the production one is `/var/ossec/etc/kvdb/`. This ensures that the production KVDBs are not used for testing. If you need to test the production KVDBs, it is necessary to back up and copy them to the test directory. Later on, this process will not be required as different environments will be managed simultaneously.

### DIFF
--- a/src/engine/source/main.cpp
+++ b/src/engine/source/main.cpp
@@ -59,6 +59,7 @@ std::string file_storage;
 unsigned int queue_size;
 unsigned int threads;
 std::string kvdb_path;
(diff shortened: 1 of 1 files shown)


JavierBejMen (approved):
LGTM

Issue #14673, opened 2022-08-18
Auto causes copy - Coverity issue


Dwordcito:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.2.x | Syscollector | Agent | Packages/Sources | All |

## Description
Based on the execution delivered in https://github.com/wazuh/wazuh/issues/14657, some issues appear on this execution.
Basically the RC of these defects are the unnecessary copy of some values.

CID1519862 proposed fix.
```
1464    auto data { m_spInfo->ports() };
1465
1466    if (!data.is_null())
1467    {
1468        for (auto& item : data)
1469        {
1470            .....
1515        }
1516    }
```

CID1519855 proposed fix.
```
365        auto rows { Utils::split(fileContent, '\n') };
366        auto fileBody { false };
367
368        for (auto& row : rows)
369        {
370            ....
382        }

```

Dwordcito:
Issue blocked because of Jenkins maintenance.

PR #15608, opened 2022-12-05
Auto causes copy

tdrauncieal:
|Related issue|
|---|
|#14673|

## Description
Partially made the issue proposed changes, due code missing
## Coverity test
image

## Update evidence at 02/17/2023
image

### DIFF
--- a/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
+++ b/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
@@ -1476,11 +1476,11 @@ nlohmann::json Syscollector::getPortsData()
     constexpr auto PORT_LISTENING_STATE { "listening" };
     constexpr auto TCP_PROTOCOL { "tcp" };
     constexpr auto UDP_PROTOCOL { "udp" };
-    const auto& data { m_spInfo->ports() };
+    auto data(m_spInfo->ports());
 
     if (!data.is_null())
     {
-        for (auto item : data)
+        for (auto& item : data)
         {
             const auto protocol { item.at("protocol").get_ref<const std::string&>() };
 


src/wazuh_modules/syscollector/src/syscollectorImp.cpp
```diff
@@ -1476,11 +1476,11 @@ nlohmann::json Syscollector::getPortsData()
     constexpr auto PORT_LISTENING_STATE { "listening" };
     constexpr auto TCP_PROTOCOL { "tcp" };
     constexpr auto UDP_PROTOCOL { "udp" };
-    const auto& data { m_spInfo->ports() };
+    auto data { m_spInfo->ports() };
```
pereyra-m, line 1479: This change has generated the errors in the Checks, you can see it as `[json.exception.type_error.304] cannot use at() with array`.

The thing is that the expression `auto data { m_spInfo->ports() };` will generate a new array and inside it will store the result from the function that already is an array.

Previously, the expression `const auto& data { m_spInfo->ports() };`  was a constant reference to the array returned by the method and no exception was generated.

See the following examples.

### Code 1

```c++
nlohmann::json function() {
  auto array = nlohmann::json::array();
  array.push_back({{"name", "test"}});
  return array;
}

int main() {
  const auto &data{function()};

  if (data.is_array()) {
    std::cout << "Data is an array" << std::endl;
  } else {
    std::cout << "Data is not an array" << std::endl;
  }
  std::cout << "Data: " << data << std::endl;

  for (auto &item : data) {
    if (item.is_array()) {
      std::cout << "Item is an array" << std::endl;
    } else {
      std::cout << "Item is not an array" << std::endl;
    }
    std::cout << "Item: " << item << std::endl;

    item.at("name");
  }

  return 0;
}
```

**Output**

```
Data is an array
Data: [{"name":"test"}]
Item is not an array
Item: {"name":"test"}
``` 

### Code 2

```c++
nlohmann::json function() {
  auto array = nlohmann::json::array();
  array.push_back({{"name", "test"}});
  return array;
}

int main() {
  auto data{function()}; // Only changed this expression 

  if (data.is_array()) {
    std::cout << "Data is an array" << std::endl;
  } else {
    std::cout << "Data is not an array" << std::endl;
  }
  std::cout << "Data: " << data << std::endl;

  for (auto &item : data) {
    if (item.is_array()) {
      std::cout << "Item is an array" << std::endl;
    } else {
      std::cout << "Item is not an array" << std::endl;
    }
    std::cout << "Item: " << item << std::endl;

    item.at("name");
  }

  return 0;
}
```

**Output**

```
Data is an array
Data: [[{"name":"test"}]]
Item is an array
Item: [{"name":"test"}]
terminate called after throwing an instance of 'nlohmann::detail::type_error'
  what():  [json.exception.type_error.304] cannot use at() with array
[1]    201573 IOT instruction (core dumped)  ./test
```

pereyra-m: But another question: why did you change the reference for a local variable?

pereyra-m: The report from Coverity only makes a reference to the **auto** in the loop as far as I can see.
So the change in the line 1479 might not be required (although it's suggested in the issue)

coverity

tdrauncieal: yes sure, but `item` is related to `data`, internally in the loop `item` is being modified so the original object would be modified

pereyra-m (dismissed):

Damian-Mangold (dismissed):

pereyra-m (approved):

jmv74211 (approved):
## QA review

- **Type**: Manual testing.
- **Status**: Approved  🟢 
- **Testing issue**: https://github.com/wazuh/wazuh-qa/issues/3740
- **Comments**:  https://github.com/wazuh/wazuh-qa/issues/3740#issuecomment-1400667507

pereyra-m (approved):

pereyra-m (dismissed):

Issue #15613, opened 2022-12-05
Warning when chkconfig is not installed during WPK rollback 


chemamartinez:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.4.0 | WPK Upgrade | Agent | Packages | All |

## Description

During the WPK testing for v4.4.0-alpha1 https://github.com/wazuh/wazuh/issues/15529, it appears a warning during the rollback of the upgrade failure tests for Ubuntu 22.04.

```
2022/12/02 12:23:37 - Installation result = 1
2022/12/02 12:23:38 - Waiting connection... Status = connected. Remaining attempts: 29.
2022/12/02 12:23:38 - Upgrade failed. Restoring...
Killing wazuh-modulesd... 
Killing wazuh-logcollector... 
Killing wazuh-syscheckd... 
Killing wazuh-agentd... 
Killing wazuh-execd... 
Wazuh v4.4.0 Stopped
2022/12/02 12:23:38 - Deleting upgrade files...
2022/12/02 12:23:38 - Restoring backup....
2022/12/02 12:23:39 - Wazuh SELinux module installation skipped (SELinux is disabled).
./var/upgrade/src/init/pkg_installer.sh: line 269: /sbin/chkconfig: No such file or directory
Starting Wazuh v4.3.10...
Started wazuh-execd...
Started wazuh-agentd...
Started wazuh-syscheckd...
Started wazuh-logcollector...
Started wazuh-modulesd...
Completed.
```

As we can see, this log appears during the rollback procedure at [`pkg_installer.sh#L267`](https://github.com/wazuh/wazuh/blob/803f3a3160fd5e59d3e956496b65637f9fb606a2/src/init/pkg_installer.sh#L267). The cause of the error is that `chkconfig` is not installed in the agent. However, this is supposed to be controlled by the script without success:
```
    # Restore service
    if [ -n "${INIT_PATH}" ]; then
        if [ $CHK_CONFIG -eq 1 ]; then
            /sbin/chkconfig --add ${SERVICE} >> ./logs/upgrade.log 2>&1
        fi
    fi
```

This issue didn't appear in previous tests like [the v4.3.10 testing](https://github.com/wazuh/wazuh/issues/15388) because `chkconfig` was installed. Therefore, this is not a new issue of v4.4.0.

Despite the warning message, the rollback is working as expected.

## Steps to reproduce

[...]

PR #15829, opened 2023-01-05
Fix chkconfig detection in pkg_installer.sh 

mjcr99:
|Related issue|
|---|
|#15613|

## Description
This PR aims to solve a warning in the upgrade.log file after running a WPK upgrade in a Linux agent where `chkconfig` is not installed. The new implementation checks straightly if chkconfig is available running `chk=$(which chkconfig)` if this bash operation returns a text output which will be stored in the `chk` variable it means that chkconfig is installed in the agent. 

In the previous version this check was made by running the following line: `if [ -f "/etc/init.d/${SERVICE}" ] && [ ! -h /etc/init.d ]`, having `SERVICE = wazuh-agent` but it has been proved that the checked file can exist without having `chkconfig` in the agent where the upgrade is being run. Specifically, when the agent is installed in a ubuntu distro and the installation is made via Wazuh packages, the `/etc/init.d/wazuh-agent` is created without `chkconfig`, resulting in unexpected behavior. 

Before the modification, we were getting the following lines in the `upgrade.log`:
```
2023/01/05 11:12:34 - Deleting upgrade files...
2023/01/05 11:12:34 - Restoring backup....
./var/upgrade/src/init/pkg_installer.sh: line 269: /sbin/chkconfig: No such file or directory
Starting Wazuh v4.3.10...
```

After the performed modification we get:
```
2023/01/05 13:18:48 - Deleting upgrade files...
2023/01/05 13:18:48 - Restoring backup....
Synchronizing state of wazuh-agent.service with SysV service script with /lib/systemd/systemd-sysv-install.
Executing: /lib/systemd/systemd-sysv-install enable wazuh-agent
Created symlink /etc/systemd/system/multi-user.target.wants/wazuh-agent.service → /lib/systemd/system/wazuh-agent.service.
Starting Wazuh v4.3.10...
```

Now there are no unexpected calls to `chkconfig`. Worth mentioning that the new lines in the upgrade.log file refer to the process of re-enabling the wazuh-agent service. This is not completely necessary because running the WPK upgrade does not modify the enabled disabled state of the wazuh-agent service even if the upgrade fails, as it is in this case. This command will take real effect in situations where the wazuh-agent is not enabled, where, after upgrading, it will be.

It's worth mentioning to that in yum distros, chkconfig should be added as a dependency to cover in the documentation since it's not ensured it's natively installed. 

## Test

- Logs without unexpected warnings.
  - [x] Ubuntu 22 agent packages installation - full upgrade
  - [x] Ubuntu 22 agent sources installation - triggering rollback
  - [x] Centos-7 agent packages installation - triggering rollback

### DIFF
--- a/src/init/pkg_installer.sh
+++ b/src/init/pkg_installer.sh
@@ -13,7 +13,6 @@ OSSEC_LIST_FILES=""
 RESTORE_OSSEC_OWN=0
 SYSTEMD_SERVICE_UNIT_PATH=""
 INIT_PATH=""
-CHK_CONFIG=0
 
 # Create the ossec user and group if they don't exist
 function create_ossec_ug {
@@ -155,14 +154,12 @@ fi
 # Init backup
 # REHL <= 6 / Amazon linux
 if [ -f "/etc/rc.d/init.d/${SERVICE}" ] && [ ! -h /etc/rc.d/init.d ]; then
-    CHK_CONFIG=1
     INIT_PATH="/etc/rc.d/init.d/${SERVICE}"
     mkdir -p "${TMP_DIR_BACKUP}/etc/rc.d/init.d/"
     cp -a "${INIT_PATH}" "${TMP_DIR_BACKUP}${INIT_PATH}"
 fi
 
 if [ -f "/etc/init.d/${SERVICE}" ] && [ ! -h /etc/init.d ]; then
-    CHK_CONFIG=1
     INIT_PATH="/etc/init.d/${SERVICE}"
     mkdir -p "${TMP_DIR_BACKUP}/etc/init.d/"
     cp -a "${INIT_PATH}" "${TMP_DIR_BACKUP}${INIT_PATH}"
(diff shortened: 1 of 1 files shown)


src/init/pkg_installer.sh
```diff
@@ -263,9 +260,11 @@ else
 
     # Restore service
     if [ -n "${INIT_PATH}" ]; then
-        if [ $CHK_CONFIG -eq 1 ]; then
+        chk=$(which chkconfig &> /dev/null)
```
chemamartinez, line 263: Forwarding the output to `/dev/null` makes `$chk` empty even when `chkconfig` exists so the `if` will be always false.
```
root@wazuh:~# which cmake &> /dev/null
root@wazuh:~# which cmake
/usr/local/bin/cmake
```
```suggestion
        chk=$(which chkconfig)
```

chemamartinez (dismissed):
LGTM!

chemamartinez (approved):

jmv74211 (approved):
## QA review

- **Type**: Manual testing.
- **Status**: Approved  🟢 
- **Testing issue**: https://github.com/wazuh/wazuh-qa/issues/3726
- **Comments**:  https://github.com/wazuh/wazuh-qa/issues/3726#issuecomment-1419405815

PR #12385, opened 2022-02-21
Optional all-columns retrieving in row deletion

jnasselle:
|Related issue|
|---|
|#12329|

## Description

This PR aims to bring two flavors to get deleted rows during a dbsync transaction: only getting the PKs of deleted rows or get all columns of it. While syscollector deltas only needs PKs, currently FIM design needs all of the data because does not support delta messages yet. 

## Tests

- Compilation without warnings in every supported platform
  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade
- [ ] Review logs syntax and correct language
- [ ] QA templates contemplate the added capabilities

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Valgrind (memcheck and descriptor leaks check)
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

- [ ] Retrocompatibility with older Wazuh versions
- [ ] Working on cluster environments
- [ ] Configuration on demand reports new parameters
- [ ] The data flow works as expected (agent-manager-api-app)
- [ ] Added unit tests (for new features)
- [ ] Stress test for affected components

- Decoder/Rule tests
  - [ ] Added unit testing files ".ini"
  - [ ] runtests.py executed without errors

PR #12094, opened 2022-02-03
Implement new DBSync options for syncRowData operation

RamiroRD:
|Related issues|
|---|
|#11434 and #11435|

## Description
This PR combines two previous PRs (https://github.com/wazuh/wazuh/pull/11872 and https://github.com/wazuh/wazuh/pull/12064) and implements a mechanism for passing options to DBSync's syncRowData method. We use this mechanism to implement the ability to ignore a set of columns when comparing the old and new row values and also to optionally return the old values of the row. 

## Tests

- Compilation without warnings in every supported platform
  - [x] Linux
  - [x] Windows
  - [x] MAC OS X
- [x] Source installation
- [x] Package installation
- [ ] Source upgrade
- [ ] Package upgrade
- [ ] Review logs syntax and correct language
- [ ] QA templates contemplate the added capabilities

- Memory tests for Linux
  - [x] Scan-build report
  - [ ] Coverity
  - [x] Valgrind (memcheck and descriptor leaks check)
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

- [ ] Retrocompatibility with older Wazuh versions
- [ ] Working on cluster environments
- [ ] Configuration on demand reports new parameters
- [ ] The data flow works as expected (agent-manager-api-app)
- [ ] Added unit tests (for new features)
- [ ] Stress test for affected components

- Decoder/Rule tests
  - [ ] Added unit testing files ".ini"
  - [ ] runtests.py executed without errors

Issue #9103, opened 2021-06-24
Replace FIM DB implementation with dbsync and rsync


Molter73:
|Wazuh version|Component|
|---|---|
| | Wazuh FIM |

### Description

As a step towards unifying DB handling between different processes, we need to integrate dbsync and rsync into FIM and replace the current custom handling of the DB with these more generic libraries. This epic is meant to group together all of the required steps we need to take to accomplish this goal, as well as track our progress.

###  Class diagram

![class_diagram](https://user-images.githubusercontent.com/29475387/182315214-2c16598e-0936-4782-8432-f0900fcc74b9.svg)

###  Sequence diagram

![image](https://user-images.githubusercontent.com/29475387/182315490-96590ac8-f0b0-4705-aa24-7e24e0528441.png)

### E-R diagram

![image](https://user-images.githubusercontent.com/29475387/182315726-25a42bc2-d77a-40e6-915c-69993aadb27e.png)

chemamartinez:
Everything closed! GJ @wazuh/watchdogs! 🚀

FrancoRivero:
Felicitaciones team !!! Por fin, tenemos el nuevo FIMDB.
Saludos

PR #15944, opened 2023-01-16
FIM DB merge from master after dbsync module modifications

jotacarma90:
|Related issue|
|---|
|#9103|

## Description
This PR contains the last merge from master that we had to do to our branch. There are several changes related to dbsync that have been a bit tricky.

Fixed also the RTR dbsync tests.

## Summary of the changes contained in the PR to facilitate the review

PR merged into master that affects dbsync:
https://github.com/wazuh/wazuh/pull/15675

Info, old PRs merged in our `9103-replace-fim-db-dbsync` branch that affected this part:
https://github.com/wazuh/wazuh/pull/12385
https://github.com/wazuh/wazuh/pull/12094

Extra changes that were needed contained in this merge:
- `Makefile`
  - Removed double quotes: `DEFINES+=-DCA_NAME='${CA_NAME}'` for Linux, on Windows they are missing. Commit: https://github.com/wazuh/wazuh/pull/15944/commits/52a48f3ef84b28252c1e6b83c3dff23e046ce83c
- `sqlite_dbengine.cpp`
  - Function `returnRowsMarkedForDelete`, removed `allColumns` functionality.
    - It has been necessary to remove also one of the parameters of the function, and all the trace that followed that parameter, this is located in this commit: https://github.com/wazuh/wazuh/pull/15944/commits/5f6f95842291ca0937ace998e619391d88a36c52
  - Function `getRowDiff`, moved the output jsons out of the // Diff found condition.
- The tests have conflicts in general because arrays are no longer used inside the jsons and because of the changes in the functions mentioned above.
  - `dbsync_test.cpp`
  - `dbsyncPipelineFactory_test.cpp`


Dwordcito (approved):

jotacarma90:
## Some errors in FIM

Added some more commits after finding a bug in FIM scan. Some false positives were being generated in the scheduled scan, for modifications where there were not really any modified fields. 
The cause was that after the delta changes, as dbsync was returning all the fields in the `json_result`, it was processing to call the FIM callback in all cases. This was causing false alerts. 

I needed to modify the `transaction_callback` that we passed to dbsync during the fim_scan, so that it checks if it is really a modification event. We should only process the modification event if the `old` field of the `json_result` contains a field that is not one of the primary keys (or the `last_event` field, which is always modified but does not imply anything in the file).

chemamartinez (approved):

Issue #11088, opened 2021-11-29
Errors found in log related to SecurityInfo


okynos:
As reported in https://github.com/wazuh/wazuh/issues/11019#issuecomment-980453570 we have detected some SecurityInfo errors inside the Windows Wazuh Agent 

Errors found in:
- [All-except-AR](https://github.com/wazuh/wazuh/issues/10990)
- [All-except-sca](https://github.com/wazuh/wazuh/issues/11045)
- [All-except-rootcheck](https://github.com/wazuh/wazuh/issues/11044)
- [ALL](https://github.com/wazuh/wazuh/issues/11039)
```
wazuh-agent: ERROR: GetSecurityInfo error = 0
wazuh-agent: ERROR: GetSecurityInfo error = 183
wazuh-agent: ERROR: GetSecurityInfo error = 183
```
All the information could be found into referred issues.

alromeros:
**#11088 update:**

Hi team,

As described in https://github.com/wazuh/wazuh/issues/11019#issuecomment-981848619, this error message is also found in **4.2.5** logs.

This leads us to believe that said error should not be considered stopper for **4.3**.

Best regards.

alromeros:
**#11088 update:**

## Error description:
The error happens due to an unexpected return in the [GetSecurityInfo](https://docs.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-getsecurityinfo) function,  which fails after trying to obtain the security descriptor of the registry key that's currently being scanned by **FIM**. An error code is then extracted by [GetLastError](https://docs.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-getlasterror) that's printed as a log message.

Three different [system error codes](https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-) have been observed within the logs:

```
wazuh-agent: ERROR: GetSecurityInfo error = 0

wazuh-agent: ERROR: GetSecurityInfo error = 183

wazuh-agent: ERROR: GetSecurityInfo error = 122
```

These error codes are specified as follows:

- **ERROR_SUCCESS** (0)
>ERROR_SUCCESS
>0 (0x0)
>The operation completed successfully.

Meaning that, though `GetSecurityInfo` fails with an error code different from 0, `GetLastError` then returns a successful output that does not match the expected return. This one can be considered the most suspicious error of the three.

- **ERROR_ALREADY_EXISTS** (183)
[...]

PR #15986, opened 2023-01-20
Fix an error code in GetSecurityInfo function in FIM code

jotacarma90:
|Related issue|
|---|
|#11088|

## Description
Hello team, 
This PR is to modify the error output of the GetSecurityInfo function. 
We had encountered some problems with the error code we were displaying using GetLastError as shown in the examples in the Windows documentation:
https://learn.microsoft.com/en-us/windows/win32/secauthz/finding-the-owner-of-a-file-object-in-c--

After analyzing the problem in depth, and seeing the errors that appear with "meaningless" error codes, they must be race conditions in which an error is first given during the execution of the GetSecurityInfo function, but nevertheless, the GetLastError function returns a value that does not seem to correspond to the function we are analyzing.

However, in this case for the GetSecurityInfo function it is not necessary, since the error code is returned in the function itself.
We will use the return to display the code, and also add the formatted error message.

### DIFF
--- a/src/shared/syscheck_op.c
+++ b/src/shared/syscheck_op.c
@@ -694,7 +694,6 @@ char *get_file_user(const char *path, char **sid) {
 
 char *get_user(const char *path, char **sid, HANDLE hndl, SE_OBJECT_TYPE object_type) {
     DWORD dwRtnCode = 0;
-    DWORD dwSecurityInfoErrorCode = 0;
     PSID pSidOwner = NULL;
     BOOL bRtnBool = TRUE;
     char AcctName[BUFFER_LEN];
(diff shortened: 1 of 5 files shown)


chemamartinez (approved):
LGTM!

chemamartinez (approved):

chemamartinez (dismissed):

damarisg:
## QA review
- **Type**: Manual testing.
- **Testing issue**: https://github.com/wazuh/wazuh-qa/issues/3903
- **Comments**: https://github.com/wazuh/wazuh-qa/issues/3903#issuecomment-1428197158

vikman90 (approved):

Issue #15979, opened 2023-01-19
AlmaLinux support: Create kernel and OS packages to match NVD vulnerabilities


MarcelKemp:
### Description

For every Linux agent scanned, we are currently indexing two extra packages manually when analysing vulnerabilities against the NVD data. One for the kernel version and another one for the OS itself. 

The reason for this is the NVD naming standard, which doesn't match with the information the agents' inventory provides us, making impossible the direct matching. Without the correct matching of the kernel version and the OS, we would lose much interesting information during the scan.

This addition happens at https://github.com/wazuh/wazuh/blob/36c9cc3db56e37d51350c09908b7b708ce525b38/src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c#L5619

For the kernel package, we will call it, as for the rest of the Linux platforms, `linux_kernel`. 

For the OS, after a little research, it has been observed that there is no specific CPE for `AlmaLinux`, so being a `RHEL` clone we will also rely on its CPEs.

PR #16081, opened 2023-02-02
Generate OS and Kernel packages for AlmaLinux to match NVD vulnerabilities

mateocervilla:
|Related issue|
|---|
|#15979|

## Description

This PR aims to generate the OS and kernel packages for AlmaLinux, so that they match with the NVD.

It has been observed that there is no specific CPE for AlmaLinux, so being a RHEL clone we will also rely on its CPEs.

### DIFF
--- a/src/error_messages/error_messages.h
+++ b/src/error_messages/error_messages.h
@@ -446,6 +446,7 @@
 #define VU_WIN_CPE_GEN_ERROR        "(5594): Could not generate the OS CPE for the agent '%.3d'"
 #define VU_ALMA_VERSION_ERROR       "(5595): Invalid AlmaLinux OS version."
 
+
 /* File integrity monitoring error messages*/
 #define FIM_ERROR_WHODATA_SUM_MAX                   "(6603): The whodata sum for '%s' file could not be included in the alert as it is too large."
 #define FIM_ERROR_CHECK_THREAD                      "(6605): Could not create the Whodata check thread."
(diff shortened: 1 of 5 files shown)


MarcelKemp (changes requested):
The vendor list for AlmaLinux (`vu_vendor_list_almalinux`) needs to be added, including it in the **`wm_vuldet_linux_rm_nvd_not_affected_packages`** function in order to avoid false positives of official AlmaLinux repository packages that have been detected by OVAL as unaffected.

pereyra-m:
### Exploratory

This branch was installed in a Ubuntu 22.04 VM and vulnerability detector was enabled.
These are the agents connected (Wazuh v4.4.5):

- Linux alma8.localdomain 4.18.0-425.13.1.el8_7.x86_64 (AlmaLinux 8.7 (Stone Smilodon) x86_64)
- Linux alma9.localdomain 5.14.0-162.22.2.el9_1.x86_64 (AlmaLinux 9.1 (Lime Lynx) x86_64 )


In both cases, the alerts were generated successfully 

```
root@ubuntu-jammy:/var/ossec# sqlite3 queue/db/001.db "SELECT count(*) from vuln_cves"
64
root@ubuntu-jammy:/var/ossec# sqlite3 queue/db/002.db "SELECT count(*) from vuln_cves"
62

```

pereyra-m (approved):
LGTM!

PR #16089, opened 2023-02-03
Vulnerability Detection doesn't analyze brew packages from macOS agents - Implementation

### LINKED ISSUE #15805: Vulnerability Detection doesn't analyze brew packages from macOS agents
|---|---|---|---|---|
| 4.3.9-4309 | Wazuh component | Manager/Agent | Packages/Sources | OS version |

This issue is for development tasks.

Thanks @maumrsms for the issue.

Hello team!

After checking and verifying, apparently during the work of this [issue](https://github.com/wazuh/wazuh/issues/14074), it was discovered that `homebrew`/`brew` vendors could not be retrieved by the `syscollector` module.
For this reason, all brew packages show an space `" "` in the `vendor` field in the packages report from Wazuh API:
```
      {
        "scan":{
          "id":0,
          "time":"2022-10-10T07:19:17Z"
        },
        "version":"3.9.4",
        "name":"python@3.9",
        "format":"pkg",
        "description":"Interpreted, interactive, object-oriented programming language",
        "location":"/usr/local/Cellar",
        "size":0,
        "vendor":" ",
        "source":"homebrew",
        "architecture":" ",
        "agent_id":"XXX"
      },
```
And unfortunately, this seems to be affecting the Vulnerability Detection report.

Apparently this happens in all versions of Wazuh, as Wazuh seems to be getting the package name and version from the corresponding folders in `/usr/local/Cellar`.

Looking around, it seems to be possible that homebrew, when installing a package in these paths, always keeps the same structure within the installed applications, so that within the path `/usr/local/Cellar/<package>/<version>/` there is a file from which Wazuh could parse the information to get the `vendor` (although it is difficult to confirm that the structure is maintained in all programs).

The best option would be to find another path from which it is possible to get this information.

Once we get the vendor, the respective Vulnerabilities should appear in Vulnerability Detector.

## DoD
- [ ] Modify vulnerability detector code to be compliant with empty vendors(only for brew packages).
- [ ] Modify syscollector to split the version from the package name. ex putty@1.4.73
- [ ] Add unit tests.
- [ ] Full green CI Checks.

pereyra-m:
|Related issue|
|---|
|Closes #15805|

## Description

This PR performs the following changes:
- The brew packages in macOS won't be reported with a `@` in their name
- Vulnerability detector will always trim the name section of the package with `@` for macOS
- A new string helper was added with its UT

## Logs/Alerts example

Some packages with specific versions were added with `homebrew` in a macOS Monterrey.
They look like this in the manager's DB

package_with_at_1

package_with_at_2

package_with_at_3

After these changes, the name of the packages no longer have the version inside

package_without_at_1

package_without_at_2

package_without_at_3

It was confirmed too that now these packages are properly analyzed by vulnerability detector.
These are the vulnerabilities before the change

cves_not_detected

After the changes, 5 new vulnerabilities were found. All these packages were installed with `homebrew`, some of them had a version in the name and some of them not

cves_detected

## Tests

  - [x] Linux
  - [x] MAC OS X
- [x] Source installation

### DIFF
--- a/src/data_provider/src/packages/brewWrapper.h
+++ b/src/data_provider/src/packages/brewWrapper.h
@@ -52,8 +52,11 @@ class BrewWrapper final : public IPackageWrapper
               but we'll only remove the last occurrence if it matches with a version
               in case there is a '@' in the package name */
             const auto pos { m_name.rfind('@') };
-            if(pos != std::string::npos) {
-                if(std::isdigit(m_name[pos + 1])){
+
+            if (pos != std::string::npos)
+            {
+                if (std::isdigit(m_name[pos + 1]))
+                {
                     m_name.resize(pos);
                 }
             }
--- a/src/data_provider/src/packages/brewWrapper.h
+++ b/src/data_provider/src/packages/brewWrapper.h
@@ -48,8 +48,15 @@ class BrewWrapper final : public IPackageWrapper
                 }
             }
 
-            // Some brew packages have the version in the name separated by a '@'
-            m_name = Utils::substrOnLastOccurrence(m_name, "@");
+            /* Some brew packages have the version in the name separated by a '@'
+              but we'll only remove the last occurrence if it matches with a version
+              in case there is a '@' in the package name */
+            const auto pos { m_name.rfind('@') };
+            if(pos != std::string::npos) {
+                if(std::isdigit(m_name[pos + 1])){
+                    m_name.resize(pos);
+                }
+            }
         }
 
         ~BrewWrapper() = default;
--- a/src/data_provider/src/packages/brewWrapper.h
+++ b/src/data_provider/src/packages/brewWrapper.h
@@ -47,6 +47,9 @@ class BrewWrapper final : public IPackageWrapper
                     break;
                 }
             }
+
+            // Some brew packages have the version in the name separated by a '@'
+            m_name = Utils::substrOnLastOccurrence(m_name, "@");
         }
 
         ~BrewWrapper() = default;
--- a/src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
+++ b/src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
@@ -5563,7 +5563,16 @@ int wm_vuldet_collect_agent_software(scan_agent *agent, sqlite3 *db, scan_ctx_t*
 
                 if (agent->dist == FEED_MAC) {
                     wchr_replace(name, ' ', '_');
-                    if (vendor) wchr_replace(vendor, ' ', '_');
+                    // Some packages may come with a '@' followed up by the product's version.
+                    // Since we already have said version, we can discard it.
+                    // We are using the reverse variant just in case the package name contains a '@'.
+                    char* tmp = NULL;
+                    if (tmp = strrchr(name, '@'), tmp != NULL) {
+                        *tmp = '\0';
+                    }
+                    if (vendor) {
+                        wchr_replace(vendor, ' ', '_');
+                    }
                 }
 
                 char* nvd_architecture = NULL;
--- a/src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
+++ b/src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
@@ -5563,16 +5563,7 @@ int wm_vuldet_collect_agent_software(scan_agent *agent, sqlite3 *db, scan_ctx_t*
 
                 if (agent->dist == FEED_MAC) {
                     wchr_replace(name, ' ', '_');
-                    // Some packages may come with a '@' followed up by the product's version.
-                    // Since we already have said version, we can discard it.
-                    // We are using the reverse variant just in case the package name contains a '@'.
-                    char* tmp = NULL;
-                    if (tmp = strrchr(name, '@'), tmp != NULL) {
-                        *tmp = '\0';
-                    }
-                    if (vendor) {
-                        wchr_replace(vendor, ' ', '_');
-                    }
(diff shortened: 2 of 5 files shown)


src/data_provider/src/packages/brewWrapper.h
```diff
@@ -47,6 +47,9 @@ class BrewWrapper final : public IPackageWrapper
                     break;
                 }
             }
+
+            // Some brew packages have the version in the name separated by a '@'
```
MiguelazoDS, line 51: Can we make it clear that the version will appear always at the end of the string?

src/wazuh_modules/vulnerability_detector/wm_vuln_detector_nvd.c
```diff
@@ -2524,8 +2524,8 @@ int wm_vuldet_check_generic_package(sqlite3 *dbCVE,
 
     sqlite3_bind_text(stmt, 1, temp_name, -1, NULL);
     if (app_package_mac) {
-        /* Selecting all vulnerabilities for a mac app package if vendor is null */
-        if (!pkg_vendor) {
+        /* Selecting all vulnerabilities for a mac app package if vendor is null, empty, space or '_' */
+        if (!pkg_vendor || pkg_vendor && (*pkg_vendor == '\0' || !strncmp(pkg_vendor, " ", 1) || !strncmp(pkg_vendor, "_", 1))) {
```
MiguelazoDS, line 2528: What are the cases where the pkg_vendor will be null, "", " " or "_"? because in the function `wm_vuldet_collect_agent_software` we are replacing ' ' by '_' so if I'm not wrong it would never be ' ' (space)
It just a detail, more checks are better than less checks

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
```diff
@@ -5563,7 +5563,16 @@ int wm_vuldet_collect_agent_software(scan_agent *agent, sqlite3 *db, scan_ctx_t*
...
+                    // Some packages may come with a '@' followed up by the product's version.
+                    // Since we already have said version, we can discard it.
+                    // We are using the reverse variant just in case the package name contains a '@'.
+                    char* tmp = NULL;
+                    if (tmp = strrchr(name, '@'), tmp != NULL) {
+                        *tmp = '\0';
+                    }
+                    if (vendor) {
+                        wchr_replace(vendor, ' ', '_');
+                    }
```
MiguelazoDS, line 5575: I understand the use of strrchr to avoid splitting a name that may have @ in its name, but given the case we have a name like this random@pkg@1.0, we are removing the version in the data provider functionality, and then here we are removing "@pkg" which interfere with the purpose of the function.

MiguelazoDS (commented):
Great Job!

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
matias-braida: If the name contains a @ character (I dont know if it is possible...). For example "random@pkg@1.0" the name captured would be "random".
Maybe a helper string function could be created that uses regex to detect this kind of "{package_name}@{version_number}" form. 
Maybe something like this one:
^(.+)@((?:[0-9]+\.)?(?:[0-9]+\.)?(?:[0-9]+))
The first capture group is the package name and the second capture group is version (supports major, major.minor and major.minor.patch forms)

src/data_provider/src/packages/brewWrapper.h
matias-braida: If the name contains a @ character and dont have version number (I dont know if it is possible...). For example "random@pkg" the name captured would be "random" and the version number would be "pkg".
Maybe a helper string function could be created that uses regex to detect this kind of "{package_name}@{version_number}" form. 
Maybe something like this one:
^(.+)@((?:[0-9]+\.)?(?:[0-9]+\.)?(?:[0-9]+))
The first capture group is the package name and the second capture group is version (supports major, major.minor and major.minor.patch forms)

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
pereyra-m: This section of the code was included just in case another provider (not `homebrew`) could have a version inserted in the name. 
But as you say, if the package name contains a `@` this will fail.

I'll remove these changes and leave only the fix in **syscollector.**

src/wazuh_modules/vulnerability_detector/wm_vuln_detector_nvd.c
pereyra-m: I left the pre-existent checks and added some more.

For example, the space `" "` is the default unknown value of **syscollector**. But it ends up being an underscore `_` because vulnerability detector tries to convert all the spaces to match the NVD. So the vendor being a `_` it's almost an accident.

I tried to be generic to avoid a little change breaking the functionality again.

src/data_provider/src/packages/brewWrapper.h
pereyra-m: The version is properly extracted in another section. 
I only need to remove everything after `@` if it's a version.

I make the assumption that in the remote case we have a `@` and a version, it will at the end. 
I'll correct this section.

MiguelazoDS (approved):

matias-braida (approved):

tdrauncieal (approved):

damarisg:
## QA review
- **Type**: Manual testing.
- **Testing issue**: https://github.com/wazuh/wazuh-qa/issues/4017
- **Comments**:  https://github.com/wazuh/wazuh-qa/issues/4017#issuecomment-1489054940

Issue #16155, opened 2023-02-13
Engine - Review CSV and KVmap


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
|5.0.0 | Engine | Manager | Packages/Sources | OS version |

## Description
It is necessary to check the kvmap and csv in the engine, as there are cases where it does not work correctly.

Once the functionality is checked, we need to make new test cases for these and correct the not expected behaviors. To do this, we will review feedback from other users who have been using the engine in recent months.

## Tasks
- [x] Check kvmap performance
- [x] Create new test corner cases for kvmap
- [x] Fix issues in kvmap
- [x] Check csv performance
- [x] Create new test corner cases for csv
- [x] Fix issues in csv
- [x] Update Documentation
- [ ] QA validate

Issue #15881, opened 2023-01-12
The kv helper breaks when receiving not well-balanced strings


roronoasins:
| Target version | Related issue | Related PR/dev branch |
|--------------------|--------------------|-----------------|
|           5.0             |   https://github.com/wazuh/wazuh-qa/issues/3735   |     https://github.com/wazuh/wazuh/issues/15421 |

# Description

During the function helpers' testing, was detected that using non-well balanced strings breaks the `parse_kv` function_helper

<details><summary>Decoder's signature</summary>

```
map:
    - ~outMap: +parse_kv/$~json.testing/=/,/'/'
```

</details>

<details><summary>Log input</summary>

```
{"type":"test","testing":"key1='value=1',key2=value''2,key3='value,3',key4='value=,''4'"}
```

</details>

```
OUTPUT:

{
    "wazuh": {
        "queue": 49,
        "origin": "/dev/stdin"
    },
    "event": {
        "original": "{\"type\":\"test\",\"testing\":\"key1='value=1',key2=value''2,key3='value,3',key4='value=,''4'\"}"
    },
    "~json": {
        "type": "test",
        "testing": "key1='value=1',key2=value''2,key3='value,3',key4='value=,''4'"
    },
    "~outMap": {
        "key1": "value=1",
        "key2": "alue''2,key3='value,3",
        "key4": "value=,'4"
    }
}
```

> The key2's value `key2=value''2,key3=` has three special chars and the first `v` is not there anymore.

## Considerations

More details in the comment within the testing issue https://github.com/wazuh/wazuh-qa/issues/3735#issuecomment-1380754376 within the kv cases.

| OS | OS version | Deployment | Image/AMI |
|-- |-- |-- |--|
| Ubuntu | 20.04 | Vagrant | `generic/ubuntu2004` | 

> The dev team provided a vagrantfile with the engine's mvp

JcabreraC:
This bug is fixed:


</details>

<details><summary>Log input</summary>

```
{"type":"test","testing":"key1='value=1',key2=value''2,key3='value,3',key4='value=,''4'"}
```

</details>

```
OUTPUT:

{
    "wazuh": {
        "queue": 49,
        "origin": "/dev/stdin"
    },
    "event": {
        "original": "{\"type\":\"test\",\"testing\":\"key1='value=1',key2=value''2,key3='value,3',key4='value=,''4'\"}"
    },
    "~json": {
        "type": "test",
[...]

Issue #15854, opened 2023-01-10
The `dsv` parser unescaping is not working


roronoasins:
| Target version | Related issue | Related PR/dev branch |
|--------------------|--------------------|-----------------|
|           5.0             |   https://github.com/wazuh/wazuh-qa/issues/3703   |     https://github.com/wazuh/wazuh/issues/15421 |

# Description

During the HLP testing, the `dsv` parser is not unescaping the quoting characters correctly.

<details><summary>decoder that was used</summary>

```
name: decoder/custom-json/0

sources:
- decoder/json/0

check:
- event.original: +s_starts/{
- ~json.type: +r_match/^test

parse:
logpar:
    - ~json.testing: <~outField/dsv/|/"/\\/out1/out2/out3/outN>
```

</details>

<details><summary>Log input</summary>

```
{"type":"test","testing":"\"\\\"value1\\\"\"|value2|value3|valueN"}
```

</details>

```
OUTPUT:

{
    "wazuh": {
        "queue": 49,
        "origin": "/dev/stdin"
    },
    "event": {
        "original": "{\"type\":\"test\",\"testing\":\"\\\"\\\\\\\"value1\\\\\\\"\\\"|value2|value3|valueN\"}"
    },
    "~json": {
        "type": "test",
        "testing": "\"\\\"value1\\\"\"|value2|value3|valueN"
    },
    "~outField": {
        "out1": "\\value1\\",
        "out2": "value2",
        "out3": "value3",
        "outN": "valueN"
    }
}
```

> The unescape value for `out1` should be `\"value1\"`. Looks like the unescaping is not correct.

## Considerations

More details in the comment within the testing issue https://github.com/wazuh/wazuh-qa/issues/3703#issuecomment-1367506757.

| OS | OS version | Deployment | Image/AMI |
|-- |-- |-- |--|
| Ubuntu | 20.04 | Vagrant | `generic/ubuntu2004` | 

> The dev team provided a vagrantfile with the engine's mvp

JcabreraC:
Solved: https://github.com/wazuh/wazuh/pull/22688

<details><summary>decoder that was used</summary>

```
name: decoder/custom-json/0

sources:
- decoder/json/0

check:
- event.original: +s_starts/{
- ~json.type: +r_match/^test

parse:
logpar:
    - ~json.testing: <~outField/dsv/|/"/\\/out1/out2/out3/outN>
```

</details>

<details><summary>Log input</summary>

```
[...]

PR #16217, opened 2023-02-15
Engine - Review CSV and KVmap

nbertoldo:
|Related issue|
|---|
|#16155|

## Description
This PR is intended to fix different corner cases detected in kvmap, csv and dsv.

### Fixed issues
After the QA testing phase, the following issues were found and opened, which have been fixed in this PR:
- #15854
- #15881

We have also received feedback from other users who have been using the engine in recent months. The following cases have been fixed in this PR:

- `kv` parser fails when a non-zero offset is used:
```
Enter a log in single line (Crtl+C to exit):

Jan  3 13:45:36 192.168.5.1 id=firewall sn=000SERIAL time="2007-01-03 14:48:06" fw=89.160.20.156


DECODERS:
  decoder/test/0  ->  failure
    [decoder/test/0] [/event/original: <event.start/%b %d %T> <host.hostname> <~tmp/kv/=/ /"/'>] -> Failure: Parse operation failed: 
    Main error: <~tmp/kv/=/ /"/'>: Invalid key-value string at 28
    Jan  3 13:45:36 192.168.5.1 id=firewall sn=000SERIAL time="2007-01-03 14:48:06" fw=89.160.20.156 
    ----------------------------^
    
    List of errors:
    <~tmp/kv/=/ /"/'>: Invalid key-value string at 28
    
    [decoder/test/0] [condition]:failure
```
- `kv` parser fails when the event contains an empty value `key=""`:
```
Enter a log in single line (Crtl+C to exit):

id=firewall sn=000SERIAL fw=89.160.20.156 pri=6 c=262144 m=98 msg="Connection Opened" n="" proto=tcp/50000


DECODERS:
  decoder/test/0  ->  failure
    [decoder/test/0] [/event/original: <~tmp/kv/=/ /"/'>] -> Failure: Parse operation failed: 
    Main error: <~tmp/kv/=/ /"/'>: Invalid number of key-value fields at 0
    id=firewall sn=000SERIAL fw=89.160.20.156 pri=6 c=262144 m=98 msg="Connection Opened" n="" proto=tcp/50000
    ^
    
    List of errors:
    <~tmp/kv/=/ /"/'>: Invalid number of key-value fields at 0
    
    [decoder/test/0] [condition]:failure
```
- `kv` parser converts a string `"key"="0123"` to a numeric value `"key"=123`:
```
Enter a log in single line (Crtl+C to exit):

key="0123"


DECODERS:
  decoder/test/0  ->  success
    [decoder/test/0] [/event/original: <~tmp/kv/=/ /"/'>] -> Success
    [decoder/test/0] [condition]:success

RULES:


OUTPUT:

{
    "wazuh": {
        "queue": 49,
        "origin": "/dev/stdin"
    },
    "event": {
        "original": "key=\"0123\""
    },
    "~tmp": {
        "key": 123
    }
}
```

## Tasks
- [x] Check kvmap performance
- [x] Create new test corner cases for kvmap
- [x] Fix issues in kvmap
- [x] Check csv performance
- [x] Create new test corner cases for csv
- [x] Fix issues in csv
- [x] Update Documentation
- [ ] QA validate

### DIFF
--- a/src/engine/source/hlp2/src/alphanumeric.cpp
+++ b/src/engine/source/hlp2/src/alphanumeric.cpp
@@ -28,19 +28,16 @@ parsec::Parser<json::Json> getAlphanumericParser(const std::string& name, Stop, 
             return res.value();
         }
 
-        auto begin {text.begin()};
-        std::advance(begin, index);
+        auto end = std::find_if(text.begin() + index, text.end(), [](char const& c) { return !std::isalnum(c); });
 
-        auto it = std::find_if(begin, text.end(), [](char const& c) { return !std::isalnum(c); });
+        auto endPos = end - text.begin();
 
-        if (it == text.end())
+        if (end == text.end())
         {
             json::Json alphaNumeric;
+            alphaNumeric.setString(std::string {text.substr(index, endPos - index)});
 
-            alphaNumeric.setString(std::string {text});
-            index += text.size();
-
-            return parsec::makeSuccess<json::Json>(std::move(alphaNumeric), index);
+            return parsec::makeSuccess<json::Json>(std::move(alphaNumeric), endPos);
         }
         else
         {
--- a/src/engine/source/hlp2/src/alphanumeric.cpp
+++ b/src/engine/source/hlp2/src/alphanumeric.cpp
@@ -28,21 +28,19 @@ parsec::Parser<json::Json> getAlphanumericParser(const std::string& name, Stop, 
             return res.value();
         }
 
-        auto end = std::find_if(text.begin() + index, text.end(), [](char const& c) { return !std::isalnum(c); });
+        const auto end = std::find_if(text.begin() + index, text.end(), [](char const& c) { return !std::isalnum(c); });
 
-        auto endPos = end - text.begin();
+        const auto endPos = end - text.begin();
 
-        if (end == text.end())
+        if (endPos == index)
         {
-            json::Json alphaNumeric;
-            alphaNumeric.setString(std::string {text.substr(index, endPos - index)});
+            return parsec::makeError<json::Json>(fmt::format("{}: Nothing to parse", name), endPos);
+        }
 
-            return parsec::makeSuccess<json::Json>(std::move(alphaNumeric), endPos);
-        }
-        else
-        {
-            return parsec::makeError<json::Json>(fmt::format("{}: Expected alphanumeric input", name), index);
-        }
+        json::Json alphaNumeric;
+        alphaNumeric.setString(std::string {text.substr(index, endPos - index)});
+
+        return parsec::makeSuccess<json::Json>(std::move(alphaNumeric), endPos);
     };
 }
 } // namespace hlp
--- a/src/engine/source/hlp2/src/dsv_csv.cpp
+++ b/src/engine/source/hlp2/src/dsv_csv.cpp
@@ -72,7 +72,7 @@ inline auto dsvParserFunction(std::string name,
             auto fValue = field.value();
 
             auto v = remaining.substr(fValue.start(), fValue.len());
-            updateDoc(doc, headers[i], v, fValue.isEscaped(), R"(")");
+            updateDoc(doc, headers[i], v, fValue.isEscaped(), R"(")", fValue.isQuoted());
 
             start += fValue.end() + 1;
             i++;
--- a/src/engine/source/hlp2/src/dsv_csv.cpp
+++ b/src/engine/source/hlp2/src/dsv_csv.cpp
@@ -72,7 +72,7 @@ inline auto dsvParserFunction(std::string name,
             auto fValue = field.value();
 
             auto v = remaining.substr(fValue.start(), fValue.len());
-            updateDoc(doc, headers[i], v, fValue.isEscaped(), R"(")", fValue.isQuoted());
+            updateDoc(doc, headers[i], v, fValue.isEscaped(), std::string {escapeChar}, fValue.isQuoted());
 
             start += fValue.end() + 1;
             i++;
--- a/src/engine/source/hlp2/src/dsv_csv.cpp
+++ b/src/engine/source/hlp2/src/dsv_csv.cpp
@@ -72,7 +72,7 @@ inline auto dsvParserFunction(std::string name,
             auto fValue = field.value();
 
             auto v = remaining.substr(fValue.start(), fValue.len());
-            updateDoc(doc, headers[i], v, fValue.isEscaped(), std::string {escapeChar}, fValue.isQuoted());
+            updateDoc(doc, headers[i], v, fValue.isEscaped(), std::string_view {&escapeChar, 1}, fValue.isQuoted());
 
             start += fValue.end() + 1;
             i++;
--- a/src/engine/source/hlp2/src/parse_field.cpp
+++ b/src/engine/source/hlp2/src/parse_field.cpp
@@ -56,7 +56,7 @@ std::optional<Field> getField(std::string_view input,
                 }
                 else
                 {
-                    bool escaped = (last_escape_location + 1 == i);
+                    bool escaped = (last_escape_location + 1 == i) && (i > 1);
                     isEscaped = isEscaped || escaped;
                     last_escape_location += (i - last_escape_location) * size_t(!escaped);
                     quote_opened = escaped || (input[i + 1] != delimiter);
--- a/src/engine/source/hlp2/src/parse_field.cpp
+++ b/src/engine/source/hlp2/src/parse_field.cpp
@@ -90,7 +90,8 @@ void updateDoc(json::Json& doc,
                std::string_view hdr,
                std::string_view val,
                bool is_escaped,
-               std::string_view escape)
+               std::string_view escape,
+               bool is_quoted)
 {
     if (val.empty())
     {
@@ -98,21 +99,24 @@ void updateDoc(json::Json& doc,
         return;
     }
 
-    int64_t i;
-    auto [ptr, ec] {utils::from_chars(val.data(), val.data() + val.size(), i)};
-    if (std::errc() == ec && (val.data() + val.size()) == ptr)
+    if(!is_quoted)
     {
-        doc.setInt64(i, hdr);
-        return;
-    }
-    else
-    {
-        double_t d;
-        auto [ptr, ec] {utils::from_chars(val.data(), val.data() + val.size(), d)};
+        int64_t i;
+        auto [ptr, ec] {utils::from_chars(val.data(), val.data() + val.size(), i)};
         if (std::errc() == ec && (val.data() + val.size()) == ptr)
         {
-            doc.setDouble(d, hdr);
+            doc.setInt64(i, hdr);
             return;
+        }
+        else
+        {
+            double_t d;
+            auto [ptr, ec] {utils::from_chars(val.data(), val.data() + val.size(), d)};
+            if (std::errc() == ec && (val.data() + val.size()) == ptr)
+            {
+                doc.setDouble(d, hdr);
+                return;
+            }
         }
     }
 
--- a/src/engine/source/hlp2/src/parse_field.cpp
+++ b/src/engine/source/hlp2/src/parse_field.cpp
@@ -45,7 +45,7 @@ std::optional<Field> getField(std::string_view input,
                 {
                     // Handle opening quote
                     quote_opened = true;
-                    if (strict && (i > 1) && input[i - 1] != delimiter)
+                    if (strict && (1 < i) && input[i - 1] != delimiter)
                     {
                         // Invalid field if strict parsing is enabled and there is no delimiter before opening quote
                         return {};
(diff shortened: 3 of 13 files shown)


src/engine/source/hlp2/src/parse_field.cpp
```diff
@@ -56,7 +56,7 @@ std::optional<Field> getField(std::string_view input,
                 }
                 else
                 {
-                    bool escaped = (last_escape_location + 1 == i);
```
juliancnn, line 59: The doxygen documentation is missing, you can base it on the following to add it to the header file
```hpp
/**
 * @brief Parses a field from a string using the given delimiter, quote, and escape characters.
 * 
 * @param input The string to parse the field from.
 * @param delimiter The delimiter character used to separate fields.
 * @param quote The quote character used to enclose fields that contain delimiter characters.
 * @param escape The escape character used to escape quote or escape characters inside a field.
 * @param strict Whether strict parsing should be used. If true, fields not enclosed in quotes cannot contain quotes.
 * @return An optional Field object containing the parsed field, or std::nullopt if the input is invalid.
 */
```

Some additional comments might be useful to explain some design decisions or to clarify some implementation details.
Please analyze and consider the following code, What do you think?
```cpp
std::optional<Field> getField(std::string_view input,
                              const char delimiter,
                              const char quote,
                              const char escape,
                              bool strict)
{
    size_t last_escape_location = 0;

    if (input.empty())
    {
        // Empty field
        return Field {0, 0, false, false};
    }

    bool quote_opened {false};
    bool isEscaped = false;
    bool isQuoted = false;

    for (auto i = 0ul; i < input.size(); i++)
    {
        if (input[i] == delimiter && !quote_opened)
        {
            // Found delimiter outside quotes, return field
            return Field {0, i, isEscaped, isQuoted};
        }
        else
        {
            if (quote_opened && quote != escape && input[i] == escape)
            {
                // Handle escape character
                last_escape_location = i;
            }
            if (input[i] == quote)
            {
                if (!quote_opened)
                {
                    // Handle opening quote
                    quote_opened = true;
                    if (strict && (i > 1) && input[i - 1] != delimiter)
                    {
                        // Invalid field if strict parsing is enabled and there is no delimiter before opening quote
                        return {};
                    }
                    if (quote == escape)
                    {
                        last_escape_location = i;
                    }
                }
                else
                {
                    // Handle closing quote
                    bool escaped = (last_escape_location + 1 == i) && (i > 1);
                    isEscaped = isEscaped || escaped;
                    last_escape_location += (i - last_escape_location) * size_t(!escaped);
                    quote_opened = escaped || (input[i + 1] != delimiter);
                    isQuoted = true;
                }
            }
        }
    }
    // Unclosed quote
    if (quote_opened && input.back() != quote)
    {
        return {};
    }

    // Return field
    return Field {0, input.size(), isEscaped, isQuoted};
};

```

src/engine/source/hlp2/src/dsv_csv.cpp
```diff
@@ -72,7 +72,7 @@ inline auto dsvParserFunction(std::string name,
             auto fValue = field.value();
 
             auto v = remaining.substr(fValue.start(), fValue.len());
-            updateDoc(doc, headers[i], v, fValue.isEscaped(), R"(")");
+            updateDoc(doc, headers[i], v, fValue.isEscaped(), std::string {escapeChar}, fValue.isQuoted());
```
juliancnn, line 75: Why do you create from a char, a std::string when the function is of type std::string_view?

src/engine/source/hlp2/src/parse_field.cpp
```diff
@@ -90,30 +90,34 @@ void updateDoc(json::Json& doc,
...
+        int64_t i;
+        auto [ptr, ec] {utils::from_chars(val.data(), val.data() + val.size(), i)};
         if (std::errc() == ec && (val.data() + val.size()) == ptr)
         {
-            doc.setDouble(d, hdr);
+            doc.setInt64(i, hdr);
             return;
         }
+        else
+        {
```
juliancnn, line 112: This else is not necessary, remove it to make it easier to read.

src/engine/source/hlp2/src/parse_field.cpp
```diff
@@ -90,30 +90,34 @@ void updateDoc(json::Json& doc,
...
+            double_t d;
+            auto [ptr, ec] {utils::from_chars(val.data(), val.data() + val.size(), d)};
```
juliancnn, line 114: Do not re-declare variables in a smaller scope if the name was used in the higher scope, this makes it difficult to read the scope and is very error prone.

src/engine/source/hlp2/src/parse_field.cpp
```diff
@@ -90,30 +90,34 @@ void updateDoc(json::Json& doc,
...
+            if (std::errc() == ec && (val.data() + val.size()) == ptr)
+            {
+                doc.setDouble(d, hdr);
+                return;
+            }
+        }
     }
 
     auto vs = std::string {val.data(), val.size()};
```
juliancnn, line 123: A version with these bugs fixed would be something like:
```cpp

void updateDoc(json::Json& doc,
               std::string_view header,
               std::string_view value,
               bool is_escaped,
               std::string_view escape,
               bool is_quoted)
{
    if (value.empty())
    {
        doc.setNull(header);
        return;
    }

    // If the value is not quoted, try parsing it as an int or double
    if (!is_quoted)
    {
        // Try parsing as int
        int64_t i;
        auto [ptr, ec] {utils::from_chars(value.data(), value.data() + value.size(), i)};
        if (std::errc() == ec && (value.data() + value.size()) == ptr)
        {
            doc.setInt64(i, header);
            return;
        }

        // Try parsing as double
        double_t d;
        auto [ptr2, ec2] {utils::from_chars(value.data(), value.data() + value.size(), d)};
        if (std::errc() == ec2 && (value.data() + value.size()) == ptr2)
        {
            doc.setDouble(d, header);
            return;
        }
    }

    // If the value is a string, unescape it if necessary and add it to the JSON document
    auto vs = std::string {val.data(), val.size()};
    unescape(is_escaped, vs, escape);
    doc.setString(vs, header);
}
```
I have not tested it (nor compiled it :rofl: ), but I recommend to take it into account for changes.

src/engine/source/hlp2/src/parse_field.cpp
```diff
@@ -90,30 +90,34 @@ void updateDoc(json::Json& doc,
                std::string_view hdr,
                std::string_view val,
                bool is_escaped,
-               std::string_view escape)
+               std::string_view escape,
```
juliancnn, line 93: The doxygen documentation is missing, and you could change the names from hdr to header? or Key?, val to value. I think it would be more readable

juliancnn (changes requested):

src/engine/source/hlp2/src/parse_field.cpp
nbertoldo: I agree, I will replace it

src/engine/test/source/hlp2/run_test.hpp
```diff
@@ -58,31 +58,43 @@ std::string to_string(const TestCase& testCase)
 }
 } // namespace
 
-static void runTest(
-    TestCase t,
-    std::function<parsec::Parser<json::Json>(std::string, Stop, Options)> parserBuilder)
+static void runTest(TestCase t,
+                    std::function<parsec::Parser<json::Json>(std::string, Stop, Options)> parserBuilder,
+                    std::string header = "",
+                    std::string tail = "")
```
LucioDonda, line 64: Update documentation with te new parameters added to runTest(

src/engine/test/source/hlp2/kvmap_test.cpp
```diff
@@ -167,11 +167,56 @@ TEST(KVParser, parser)
...
+                  Options {"=", ",", "'", "'"},
+                  fn(R"({"key1":"value=1","key2":"value'2","key3":"value,3","key4":"value=,'4"})"),
+                  61}
+        };
     for (auto t : testCases)
     {
         runTest(t, hlp::getKVParser);
+        runTest(t, hlp::getKVParser, "header", "");
+        runTest(t, hlp::getKVParser, "header", "tail");
+        runTest(t, hlp::getKVParser, "", "tail");
```
LucioDonda, line 219: This could be moved inside runTest(), right? calling it with both new parameters and testing this 3 scenarios inside that function:
1. without header and tail
2. with both
3. with header
4. with tail

LucioDonda (changes requested):

src/engine/source/hlp2/src/parse_field.cpp
nbertoldo: Ok, I made the changes and tested it, and it works

src/engine/test/source/hlp2/run_test.hpp
```diff
@@ -58,35 +58,63 @@ std::string to_string(const TestCase& testCase)
...
+    for (int i = 0; i < 4; i++)
     {
-        ASSERT_FALSE(r.error().empty());
+        try
+        {
+            auto stopString = printStop(std::get<2>(t));
+            if(stopString == emptyString)
+            {
+                std::string endString {stopString + matrix[i][1]};
+                std::list<std::string> stopPrueba = {endString};
```
LucioDonda, line 90: try to avoid non-english names

src/engine/test/source/hlp2/run_test.hpp
```diff
@@ -58,35 +58,63 @@ std::string to_string(const TestCase& testCase)
...
+                parser = parserBuilder({}, stopPrueba, std::get<3>(t));
+            }
+            else
+            {
+                parser = parserBuilder({}, std::get<2>(t), std::get<3>(t));
+            }
```
LucioDonda, line 96: can this be merged into a single block?
```
auto stopString = printStop(std::get<2>(t));
std::list<std::string> stopPrueba = (stopString == emptyString ? {matrix[i][1]} : {stopString});
parser = parserBuilder({}, stopString, std::get<3>(t));
```
not checked but it can be use as a base idea

src/engine/test/source/hlp2/run_test.hpp
```diff
@@ -58,35 +58,63 @@ std::string to_string(const TestCase& testCase)
...
+        }
+        catch (std::runtime_error& e)
+        {
+            ASSERT_FALSE(expectedSuccess)
+                << fmt::format("Error building parser: {}", e.what());
+            return;
+        }
```
LucioDonda, line 103: if the method that throws the exception is the parse builder, then I think it could be better to not include the rest.

LucioDonda (changes requested):

src/engine/source/hlp2/src/alphanumeric.cpp
```diff
@@ -28,19 +28,16 @@ parsec::Parser<json::Json> getAlphanumericParser(const std::string& name, Stop,
...
+        auto end = std::find_if(text.begin() + index, text.end(), [](char const& c) { return !std::isalnum(c); });
 
-        auto it = std::find_if(begin, text.end(), [](char const& c) { return !std::isalnum(c); });
+        auto endPos = end - text.begin();
 
-        if (it == text.end())
+        if (end == text.end())
         {
             json::Json alphaNumeric;
+            alphaNumeric.setString(std::string {text.substr(index, endPos - index)});
```
juliancnn, line 39: Please check this decoder:

```yml
---
name: decoder/zz_test/0

metadata:
  description: Accept all

check: +ef_exists/wazuh.queue AND (+i_eq/wazuh.queue/49 OR +i_eq/wazuh.queue/50)

parse:
  logpar:
    - event.original: "<~test_tail/alphanumeric> %%%"
    - event.original: "### <~test_offset/alphanumeric>"
    - event.original: "### <~test_offset_and_tail/alphanumeric> %%%"

```
there is an issue with cases where the % % % characters are included. Specifically, the decoder is failing when it encounters the following two cases:

`### 123 %%%`
`123 %%%`

I will fix the parser to properly handle these cases. Once the parser is fixed, I will update the tests and the runtest if necessary.

To clarify the order in which I will make the fixes:

Fix the parser to handle cases with theoffset and tail.
Update the tests to ensure that all cases are handled correctly
If necessary, update the runtest to verify that the decoder works as expected in a real-world scenario

juliancnn (changes requested):

src/engine/source/hlp2/src/parse_field.cpp
```diff
@@ -56,7 +57,8 @@ std::optional<Field> getField(std::string_view input,
                 }
                 else
                 {
-                    bool escaped = (last_escape_location + 1 == i);
+                    // Handle closing quote
+                    bool escaped = (last_escape_location + 1 == i) && (i > 1);
```
nmkoremblum, line 61: remember to keep the variables at the right of the comparison

nmkoremblum (changes requested):

LucioDonda (approved):
LGTM!

juliancnn (approved):
LGTM!

nmkoremblum (approved):
LGTM

JcabreraC (approved):
LGTM !

Issue #16156, opened 2023-02-13
Engine - Reorganize API & API event ingest


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
|5.0.0 | Engine | Manager | Packages/Sources | OS version |

## Description
It is necessary to standardize the engine API. Currently, each API module uses different messages, so we want them to be consistent.

Also, we need to be able to send events to the engine through the API.

All this must be properly documented for future endpoints to use good practices.

## Tasks
- [ ] Standardize API
- [ ] Implement event sending through the API
- [ ] Update Documentation
- [ ] QA validate

PR #16235, opened 2023-02-16
Engine reorganize api

juliancnn:
|Related issue|
|---|
|#16156|

## Description

The Wazuh Engine API is in charge of providing a communication mechanism between different Wazuh modules, as well as with other applications that wish to integrate with Wazuh. The API is designed to allow obtaining and updating data and configurations of the different components of the Engine, such as the router, catalog, metrics and configuration.

The Engine API is based on GitHub issue #5934, which defines a communication protocol between the different Wazuh components. This protocol is used by the API to send and receive structured messages, in which you specify the action you want to perform and the parameters required to do so. The messages are transmitted in JSON format.
To allow independence from the underlying communication protocol, the Wazuh Engine API uses the Proto3 format to define the parameters of each command and the corresponding responses. Proto3 is a protocol definition language developed by Google, which allows defining the structure of messages sent between different applications or components.

The use of Proto3 allows to define in a clear and structured way the data sent and received through the API, which facilitates its understanding and its integration with other applications. In addition, by using a structured message format, the probability of errors is reduced. In the Engine API, proto3 data is serialized in JSON format. The proto3 messages are obtained from the /parameters field in the requests and serialized in the request as a JSON object. In the responses, the serialization is stored in the /data field also in JSON object format. Each command is defined by a Proto3 message, which specifies the parameters needed to perform the requested action, and a corresponding response, which indicates the status of the operation and the data obtained.

## Wazuh-Engine: request and response

_Check #5934 first_

In the Wazuh-Engine context, the "command" refers to the action to be performed, such as deleting a route, query the KVDB, change some engine configuration or insert an event in the router queue. The "parameters" are the additional data needed to perform that action, such as the route name or database name and the key to query the KVDB.
Currently, there is no limitation on the value of "command" (beyond it being a JSON string), but there is a naming convention:

**\<module.resource>/\<method>**

"Module" is the module involved in the request, that is, the module that manages the resource, so that it performs the action (<method>) on the resource managed by that module. For example, "router.route/post" indicates that a resource called "route" is being added to the router.
In the Wazuh Engine API, proto3 data is serialized in JSON format. The proto3 messages are obtained from the /parameters field in the requests and serialized in the JSON request. In the responses, the serialization is stored in the /data field in JSON format. This allows the data to be sent and received in a structured way, making it easy to understand and integrate with other applications.

## Proto3 file structure

Each Wazuh Engine module that makes use of the API has an associated .proto file in the source/proto/src folder, which is used to generate the API code. To maintain the file structure, names and messages in the API, the following rules must be followed:

- The proto3 syntax is used.
- Each module has a namespace (pkg) that will be under the "com.wazuh.api.engine" package. For example, - "com.wazuh.api.engine.catalog".
- Each command uses two proto3 messages: one for the query and one for the response.
- The messages must be defined one below the other, i.e., first the query and then the response.
- The messages must have documentation, not only for the message but also for each field.
- The enums values must be lowercase (except for ReturnStatus) so that, when queries are serialized in JSON, values must not be written in uppercase. For example, {"format": "YAML"} or {"type": "ENVIRONEMNT"}.

### For the query message:
- All query messages have the suffix "_Request".
- The message name uses camel case and has the resource name followed by the method with no periods or special characters. - For example, RouterPost_Request for the router.route/post command.
- Messages must have a header describing the use of the command.

### For the response message:
- All reply messages have the suffix "_Response".
- All response messages have a status field of type ReturnStatus, which takes the values OK or ERROR.
- All response messages have an optional error field of type string containing an error message that can be read by a human in case status contains the value ERROR. If status contains the value OK, this field is not added to the response.
- If the response message only needs the status field and the error, then the GenericStatus_Response message defined in the engine proto package must be used. For example, a command that performs an action such as uploading a route only needs to know if it went well and, if it went wrong, a description with the error.
- If the response message is a GenericStatus_Response, there must be a comment below the query message indicating this.

The following is an example of how the query and response messages should be defined:

```proto
syntax = "proto3";

import "engine.proto"; <<<< Contains the ReturnStatus type

package com.wazuh.api.engine.config; <<<< PKG Name inside engine

/***************************************************
 * Get the runtime configuration of the manager <<<< cmd brief
 *
 * The argument name is optional. If it is not provided, the whole 
 * configuration is returned. <<<< Details
 * command: config.runtime/get <<<< Command related to msg
 **************************************************/
message RuntimeGet_Request
{
	optional string name = 1; // Name of the configuration option (its optional)
}

message RuntimeGet_Response
{
	ReturnStatus status = 1; 	// Status of the query
	optional string error = 2;   // Error message if status is ERROR
	optional string content = 3; // Content of the configuration option if status is OK
}

/***************************************************
 * Update the runtime configuration of the manager
 *
 * command: config.runtime/put (<resource>/<action>)
 **************************************************/
message RuntimePut_Request
{
	optional string name = 1;	// Name of the configuration option
	optional string content = 2; // Content of the configuration option
}
// message RuntimePut_Response -> Return a GenericStatus_Response <<<< Generic response
```

## API ERROR responses:
The command (/command) of the request in the Wazuh-Engine context is used to route the request to the corresponding handler. If the handler exists, then the result of its execution will be stored in the /data field of the response and the /error field will be 0.
The /data field of the response and /parameters field of the request will be defined by Protobuf to allow independence from the Wazuh communication protocol.
Communication and protocol related errors, i.e. all those errors that are not from the handler or, in other words, from the execution of the command, will have the /error field non-zero and a message in the /message field accompanying them. These errors are well defined and are as follows:
- Error 1: Unknown error. This error should never occur and reports a logic problem (Exception not catched).
- Error 2: Invalid JSON request. The request does not comply with the JSON standard.
- Error 3: Invalid message size. Indicates a protocol problem, the request size is larger than allowed (<2MB).
- Error 4: Protocol error. The request is a valid JSON but does not comply with protocol version 1.
- Error 5: Command not found.

## Serialization
In a Wazuh-Engine API request, proto3 messages are obtained from the /parameters field and the message type is identified by the /command field. To serialize the JSON object and obtain the message from Protobuf, the JSON mapping rules defined in the JSON Mapping section of the Protobuf documentation (https://protobuf.dev/programming-guides/proto3/#json) must be followed. It is important to note that, as a general rule, enums should always be treated as strings and not as numbers.
There are several libraries and tools that facilitate serialization and deserialization of Protobuf messages in different programming languages. In Python you can use the Google protobuf library to work with Protobuf messages, while in Java you can use the protobuf-java library also provided by Google.

## Data flow and API Components
The API implementation is in its simplest state. When a client makes a request to the Wazuh-Engine API, the request is received by the server and the request is checked for validity and format. Then, the server searches the API store for the handler that corresponds to the received request.
The handler is searched with the /command parameter of the request in the map of handlers defined in the API implementation. Each handler is responsible for checking the /parameters field of the request and transforming it into the corresponding protobuf message that is used as input for the functions.
Once the handler has been identified, it is executed with the client request to produce a response. The handler is a specific function that takes the client request and produces a response for the client. The response is sent back to the client in Wazuh Response format (JSON that completes with protocol #5934).

image

As for the API data flow, it can be described as follows:

1. The server receives the request from the client and verifies its validity and format.
2. The server searches the API store for the handler that corresponds to the received request.
3. The handler is executed with the client request to produce a response.
4. The response is sent to the client.

## Code generation

The source/proto/generateCode.sh script formats and generates the code for all the *.proto files in the src folder and saves them in the include folder. It is important to note that every time a change is made to any *.proto file in the src folder, this script must be run to update the generated code in the include folder.
CMake takes care of finding all the *.hpp/cpp files in the src folder and generating the corresponding static library. However, if you delete or rename *.proto files, you need to manually delete the corresponding auto-generated files in the include folder. This is important to keep in mind to avoid adding deprecated files to the library. Finally, the installation of proto can be seen in the source/proto/readme.md file.


src/engine/source/proto/src/catalog.proto
```diff
@@ -0,0 +1,106 @@
...
+    output = 4;      // Output
+    environment = 5; // Environment
+    schema = 6;      // Schema
+    collection = 7;  // Collection
+}
+
+/***************************************************
+ * Post a resource in the catalog
+ *
+ * command: catalog.resource/post (<resource>/<accion>)
```
LucioDonda, line 39: replace `accion` for _action_, same in the documentation

LucioDonda (changes requested):
pending manual test and part of files review

src/engine/source/api/include/api/registry.hpp
```diff
@@ -55,24 +53,24 @@ class Registry
...
+     * @brief Register a route in the registry
      *
      * @param command The command name, can't be empty
      * @param callback The callback function
      * @return true If the command was registered
      * @return false If the command was not registered (already exists, the command is
      * empty or the callback is null)
      */
-    bool registerCommand(const std::string& command, const CommandFn callback);
+    bool registerHandler(const std::string& command, const Handler callback);
```
JavierBejMen, line 64: Handler should be passed by reference?

src/engine/source/builder/builders/opBuilderHelperMap.cpp
```diff
@@ -641,7 +641,8 @@ base::Expression opBuilderHelperStringFromArray(const std::any& definition)
             }
 
             // accumulated concation without trailing indexes
-            const std::string composedValueString {utils::string::join(stringArray, separator)};
+            const std::string composedValueString {
```
JavierBejMen, line 644: This format seems off

JavierBejMen (changes requested):

src/engine/test/source/api/CMakeLists.txt
```diff
@@ -1,12 +1,14 @@
...
-    wazuhResponse_test.cpp
+    adapter_test.cpp
 )
 
 target_link_libraries(api_test gtest_main base api)
 gtest_discover_tests(api_test)
 
-add_subdirectory(catalog)
+# Print warning to Cmake to not build these tests
+message(WARNING "TODO These tests are not built, fix them and remove this warning")
```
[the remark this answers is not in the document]
juliancnn: Fixed

src/engine/source/proto/src/catalog.proto
juliancnn: Fixed

src/engine/source/api/include/api/registry.hpp
juliancnn: done

src/engine/source/builder/builders/opBuilderHelperMap.cpp
juliancnn: done (merge issue)

JavierBejMen (approved):
LGTM

JcabreraC (approved):
LGTM !

Issue #16226, opened 2023-02-16
FIM realtime support for macOS and BSD


chemamartinez:
### Description

The goal of this issue is to support the real-time mode of FIM for the following agents:
- macOS
- BSD family (FreeBSD, OpenBSD, NetBSD...)

Before this issue for development, we performed research and PoC to determine the viability of the available technologies. More information can be found at https://github.com/wazuh/support/issues/101.
It also replaces https://github.com/wazuh/wazuh/issues/296.

This feature will be powered by the [libinotify-kqueue](https://github.com/libinotify-kqueue/libinotify-kqueue) library. It provides an Inotify API based in kqueue with the same features that the Linux inotify library.

chemamartinez:
For the who-data feature check this issue: https://github.com/wazuh/wazuh/issues/8602

annarober:
Hello I'm new to Wazuh but found out that the FIM is not working for MacOS.   Wondering if there are any plans to make that available in future releases.  I posted on the community slack channel today when I was looking for a potential answer.  Currently I'm at 4.14.6.   Thanks!

PR #16316, opened 2023-03-02
16226 fim realtime support for macos and bsd

mjcr99:
|Related issue|
|---|
|#16226|

## Description

This pull request aims to integrate the libinotify-kqueue library into the Makefile of the Wazuh project, in order to enable this library functionalities to provide realtime FIM in BSD based systems and Mac OS systems.

The new dependency is covered if the `uname -s` shows `Darwin OpenBSD FreeBSD NetBSD`, in the rest of cases, the dependency is not installed since it's just used in these systems.

During this development, we have found a problem regarding the Darwin cases. When the libinotify-kqueue is compiled and built, it generated the static and dynamic versions of this library. In Mac Os distros, this has been a problematic behavior, because the library builds a `libinotify.dylib` file and a `libinotify.0.dylib`. Due to some reason, we have not been capable of totally understanding, the presence of both dynamic libraries was a source of problems, and this is why we added this modification in order to delete this problematic file.

It's still necessary to apply some changes in order to download this dependency from the wazuh's packages warehouse, so to test this PR is mandatory tu download and place manually the library in src/external/ folder. Just decompress the library and copy the folder into the mentioned one.

## Library dependency 
[libinotify-kqueue.zip](https://github.com/wazuh/wazuh/files/10874217/libinotify-kqueue.zip)


chemamartinez (approved):
LGTM

Issue #3766, opened 2019-07-29
Increase the maximum EPS for the agent bucket


chemamartinez:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 3.9.3 | Agentd | Agent | All | All |

Hi team,

Since the agent queue appeared, the agent EPS by default is set to 500 (https://documentation.wazuh.com/current/user-manual/reference/ossec-conf/client_buffer.html?highlight=events_per_second#events-per-second) and the maximum EPS available is 1000.

That limit is too short, it has to be extended to 100000.

The current implementation should be changed, currently, we use the `Sleep()` function of the Windows API (https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleep). https://github.com/wazuh/wazuh/blob/91de7ca0118f96027805be613937b37a3bb41a83/src/client-agent/buffer.c#L224

Which has a granularity of milliseconds. So, that solution doesn't work for us. A possible solution could be to use the `select()` function (https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-select) but its documentation says is not supported on Windows older than Vista, in addition, it seems that cannot be used without creating a socket like the Unix version (https://blogs.msdn.microsoft.com/cellfish/2008/09/17/sleep-less-than-one-millisecond/).

Further investigation is needed to find the optimal solution.

Regards.

## Subtasks

- [ ] ~https://github.com/wazuh/wazuh-qa/issues/4016~
- [ ] https://github.com/wazuh/wazuh-documentation/issues/5967
- [ ] https://github.com/wazuh/qa-integration-framework/pull/26

myanaros:
I'm curious, why was the EPS maximum set to something so low? 

1000 EPS is barely enough to handle average production web server loads.

jotacarma90:
## Update
Hello team, 
We have decided to implement a new EPS control mechanism for the client buffer. We are going to use the existing library in analysisd, `limits.c`, which makes use of semaphores to put the thread on hold when the maximum number of events per second is reached.
For this it is necessary to generalize this library to a common directory. In addition to the configuration setting for events per second (currently in the client buffer we have `events_per_second`), we are going to add a `timeframe`, which indicates how often the remaining event count will be updated before blocking the thread (credits).

## List of issue tasks

- [ ] Move the library to a common directory.
- [ ] Generalize the internal limits_t structure to contain all the necessary variables.
[...]

PR #16346, opened 2023-03-08
Adapt `limits.c` to use with the agent bucket

jotacarma90:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/3766|

## Description

Hello team,
In this PR we are going to generalize the limits.c library, which is used to implement the functionality of limiting the events per second of the analysisd module.
We want to adapt it to use it also in the agent buffer, which right now makes use of a nanosleep function, which is not accurate enough to increase the limit currently imposed (1000).

## Configuration options

```
<client_buffer>
  <!-- Agent buffer options -->
  <disabled>no</disabled>
  <queue_size>5000</queue_size>
  <events_per_second>100000</events_per_second>
  <eps_timeframe>10</eps_timeframe>
</client_buffer>
```

## Logs/Alerts example

`2023/03/14 14:08:20 wazuh-agentd: INFO: EPS limit enabled, EPS: '5', timeframe: '1'`

## Tests

  - [x] Linux
  - [x] Windows
- [x] Source installation

- Memory tests for Linux
  - [x] Coverity
  - [x] Scan-build report
- Memory tests for Windows
  - [x] Scan-build report

- [x] Configuration on demand reports new parameters


src/analysisd/analysisd.c
```diff
@@ -984,7 +986,10 @@ void OS_ReadMSG_analysisd(int m_queue)
     }
 
     /* Initialize EPS limits */
-    load_limits(Config.eps.maximum, Config.eps.timeframe, Config.eps.maximum_found);
+    analysisd_limits = init_limits(Config.eps.maximum, Config.eps.timeframe);
+    if (!analysisd_limits->enabled && !Config.eps.maximum_found && analysisd_limits->timeframe > 0) {
+            mwarn("The eps maximum value is missing in the configuration block.");
```
chemamartinez, line 991: Several doubts here:

- This is a warning? Because I don't see the limits section in the config template, so I think this warning would appear always by default.

- If the limits have values by default we should apply these values if we don't find it configured so this warning should not appear. If we are not applying limits by default, we cannot continue the execution with null values.

chemamartinez (dismissed):
LGTM!

src/analysisd/analysisd.c
jotacarma90: Done!

jotacarma90:
Hello team, the PR checks that are failing are not related to this PR or expected:

- Install agent on Solaris 11 (i386): Failure to generate packages for those specific builds, they are not related to this PR.
- Install manager on CentOS 7 (amd64) (solaris-11): Same, tested on local, work correctly

- Test: Analysisd IT - Tier 0,1 -manager: The analysisd integration tests fail because they have been fixed in the new framework (you can see in the corresponding workflow).

- Logtest IT - Tier 0,1 -manager: The logtest failures are general from 4.6.0, you can see more in this issue:
https://github.com/wazuh/wazuh-qa/issues/4354

vikman90 (approved):

Issue #16385, opened 2023-03-15
Engine - Improve server performance


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
|5.0.0 | Engine | Manager | Packages/Sources | OS version |

## Description
It is necessary to make a deeper study of the libuv library to improve the performance of the server.

This library was incorporated in the engine. Now, it is required to use it so that the API events are asynchronous.

Also, the necessary code cleanup will be done and corrections in the module detected by the RTR tools will be applied.

## Tasks
- [ ] Study Libuv library
- [ ] Implement asynchrony
- [ ] Apply RTR corrections
- [ ] Update Documentation
- [ ] QA validate

PR #16432, opened 2023-03-19
Engine improve server performance

juliancnn:
|Related issue|
|---|
|#16385|

This PR Implement:

- The new queue abstraction for event (With flooded file) for isolate moodycamel:
  - [x] Implementation
  - [x] Test
  - [x] Replace actual implementation on server and router 
- The new API object, to remove dependency on the server API. It should be in charge of accepting a message (String type) and Returning a String type response. The api will be in charge of transforming the buffer of the received message into the Wazuh protocol, then routing it to the corresponding handler, receiving the response and finally returning the response as a string so that the server can send it to the client.
  - [x] Implementation
  - [ ] Test
  - [x] Replace actual implementation on server and router 
- New implementation of the server, where endpoints can be registered with a protocol (Stream socket, Datagram socket, but expandable to more). Endpoints can be registered with protocol and a callback function for handling each message. The server must never process events of any kind. If the messages need to be answered, as in the case of the API, they should be processed in thread workers. In the event that messages do not require a response, they can be pushed to a queue so that other threads process them without returning a response.
  - [x] Base Implementation
  - [x] Event Endpoint implementation (unix socket datagram)
  - [x] API Endpoint
  - [x] Test
  - [x] Replace actual implementation on server and router 
  
 ## Other tasks
- [x] Study Libuv library
- [x] Implement asynchrony
- [x] Apply RTR corrections
- [x] Update Documentation
- [ ] QA validate


src/engine/source/api/include/api/api.hpp
```diff
@@ -1 +1,93 @@
...
+     * #TODO: add the correct description to the errors
+     * #TODO: Test
+     * Errors:
+     * - If the request is invalid, the response will be an error message.
+     * - If the handler throws an exception, the response will be an error message.
+     * - If the handler is not registered, the response will be an error message.
+     * @param message Request message
+     * @return std::string Response message
+     */
+    std::string processRequest(const std::string& message)
```
LucioDonda, line 56: Shouldn't we add some test for this method?

src/engine/source/api/include/api/config/config.hpp
```diff
@@ -135,11 +135,11 @@ api::Handler runtimeSave(ConfHandler<ConfDriver> confHandler)
 }
 
 template<typename ConfDriver>
-bool registerHandlers(std::shared_ptr<api::Registry> registry, ConfHandler<ConfDriver> confHandler)
+bool registerHandlers(std::shared_ptr<api::Api> api, ConfHandler<ConfDriver> confHandler)
```
LucioDonda, line 138: Can this registerHandlers be change to void or to unify the others catalog, kvdb, config) to bool return type?

LucioDonda (changes requested):

src/engine/source/api/src/api.cpp
```diff
@@ -0,0 +1,5 @@
+#include <api/api.hpp>
+
+namespace api
+{
+}
```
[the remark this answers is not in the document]
juliancnn: Yes, this is necessary since it is distributed as packages but it is header-only.

src/engine/source/cmds/src/defaultSettings.hpp
```diff
@@ -3,38 +3,66 @@
...
+
+constexpr auto ENGINE_SRV_EVENT_SOCK = "/var/ossec/queue/sockets/queue";
+constexpr auto ENGINE_SRV_EVENT_SOCK_ENV = "WZE_EVENT_SOCK";
+
+constexpr auto ENGINE_SRV_EVENT_QUEUE_TASK = 100000;
+constexpr auto ENGINE_SRV_EVENT_QUEUE_TASK_ENV = "WZE_EVENT_QUEUE_TASK";
 
-constexpr auto ENGINE_API_SOCK = "/var/ossec/queue/sockets/engine-api";
-constexpr auto ENGINE_API_SOCK_ENV = "WZE_API_SOCK";
+constexpr auto ENGINE_SRV_API_SOCK = "/var/ossec/queue/sockets/engine-api";
```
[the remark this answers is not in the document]
juliancnn: This PR adds 6 new configurations to the existing ones, which are too many configurations not to have a related name. Now the configurations are associated to the modules, you can also see this separated in the main of the server command.

src/engine/source/api/include/api/config/config.hpp
juliancnn: done

src/engine/source/api/include/api/api.hpp
juliancnn: done

src/engine/source/api/src/catalog/handlers.cpp
```diff
@@ -298,19 +298,17 @@ api::Handler resourceValidate(std::shared_ptr<Catalog> catalog)
...
-        registry->registerHandler("catalog.resource/validate", resourceValidate(catalog));
-    }
-    catch (const std::exception& e)
+    bool ok = api->registerHandler("catalog.resource/post", resourcePost(catalog))
+              && api->registerHandler("catalog.resource/get", resourceGet(catalog))
+              && api->registerHandler("catalog.resource/put", resourcePut(catalog))
+              && api->registerHandler("catalog.resource/delete", resourceDelete(catalog))
+              && api->registerHandler("catalog.resource/validate", resourceValidate(catalog));
+
+    if (!ok)
```
nmkoremblum, line 309: Can we make it const?

src/engine/source/api/src/kvdb/handlers.cpp
```diff
@@ -290,24 +289,22 @@ api::Handler dbPut(std::shared_ptr<kvdb_manager::KVDBManager> kvdbManager)
...
+    bool ok = api->registerHandler("kvdb.manager/post", managerPost(kvdbManager))
+              && api->registerHandler("kvdb.manager/delete", managerDelete(kvdbManager))
+              && api->registerHandler("kvdb.manager/get", managerGet(kvdbManager))
+              && api->registerHandler("kvdb.manager/dump", managerDump(kvdbManager)) &&
+              // Specific KVDB (Works on a specific KVDB instance, not on the manager, create/delete/modify keys)
+              api->registerHandler("kvdb.db/put", dbPut(kvdbManager))
+              && api->registerHandler("kvdb.db/delete", dbDelete(kvdbManager))
+              && api->registerHandler("kvdb.db/get", dbGet(kvdbManager));
+
+    if (!ok)
```
nmkoremblum, line 305: Can we make it const?

nmkoremblum: Same for every similar case

src/engine/source/server/src/endpoint.cpp
```diff
@@ -0,0 +1,29 @@
...
+            if (0 != unlink(m_address.c_str()))
+            {
+                auto msg = fmt::format("Cannot remove the socket '{}': {} ({})", m_address, strerror(errno), errno);
+                throw std::runtime_error(std::move(msg));
+            }
+        }
+        else
+        {
+            auto msg = fmt::format("Path '{}' already exists and it is not a socket", m_address);
+            throw std::runtime_error(std::move(msg));
```
nmkoremblum, line 25: can we directly use fmt::format?

src/engine/source/server/src/endpoints/unixDatagram.cpp
```diff
@@ -0,0 +1,264 @@
...
+{
+    if (address.empty())
+    {
+        throw std::runtime_error("Address must not be empty");
+    }
+
+    if (address.length() >= sizeof(sockaddr_un::sun_path))
+    {
+        auto msg = fmt::format("Path '{}' too long, maximum length is {} ", address, sizeof(sockaddr_un::sun_path));
+        throw std::runtime_error(msg);
```
nmkoremblum, line 35: Can we directly use the fmt::format? The same for every occurrence

src/engine/source/server/src/endpoints/unixDatagram.cpp
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    if (m_address[0] != '/')
```
nmkoremblum, line 38: Please invert the condition :smile:

src/engine/source/server/src/endpoints/unixDatagram.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+    m_loop = loop;
+    m_handle = m_loop->resource<uvw::UDPHandle>();
+
+    // Listen for incoming data
+    m_handle->on<uvw::UDPDataEvent>(
+        [this](const uvw::UDPDataEvent& event, uvw::UDPHandle& handle)
+        {
+            // Get the data
+            auto data = std::string {event.data.get(), event.length};
```
nmkoremblum, line 76: can we make it const?

src/engine/source/server/src/endpoints/unixStream.cpp
```diff
@@ -0,0 +1,354 @@
...
+UnixStream::UnixStream(const std::string& address,
+                       std::shared_ptr<ProtocolHandlerFactory> factory,
+                       const std::size_t taskQueueSize,
+                       std::size_t timeout)
+    : Endpoint(address, taskQueueSize)
+    , m_handle(nullptr)
+    , m_timeout(timeout)
+    , m_factory(std::move(factory))
+{
+    if (m_timeout == 0)
```
nmkoremblum, line 19: Please invert the condition, same for every ocurrence

nmkoremblum (changes requested):

src/engine/source/server/src/endpoints/unixStream.cpp
juliancnn: done

src/engine/source/server/src/endpoints/unixDatagram.cpp
juliancnn: No, since it would not be possible to apply movement semantics.

src/engine/source/server/src/endpoints/unixDatagram.cpp
juliancnn: it is not necessary, m_address is a const.

src/engine/source/server/src/endpoints/unixDatagram.cpp
juliancnn: No, it is not necessary.

src/engine/source/server/src/endpoint.cpp
juliancnn: No, it is not necessary.

src/engine/source/api/src/kvdb/handlers.cpp
juliancnn: done

src/engine/source/api/src/catalog/handlers.cpp
juliancnn: done

JavierBejMen (approved):
LGTM

Issue #16265, opened 2023-02-22
Memory leaks in Vulnerability Detector after disk errors


MarcelKemp:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.4 | Vulnerability Detector | Manager | All | All |

## Description

After investigating the following issue https://github.com/wazuh/wazuh/issues/16189, we have found the memory leak impacting the _wazuh-modules_ daemon, where you can see how the RSS memory starts to increase when the RSS memory of _analysisd_ stops and crashes, just as you can see that _modulesd_ also reaches the limit and causes the daemon to crash.
![image](https://user-images.githubusercontent.com/14913942/218501434-e0e29388-ecde-44d4-b528-be2179a5b0f4.png)

## Memory report

```
==31800== LEAK SUMMARY:
==31800==    definitely lost: 28,928 bytes in 826 blocks
==31800==    indirectly lost: 399,594 bytes in 10,123 blocks
==31800==      possibly lost: 4,912 bytes in 16 blocks
==31800==    still reachable: 967,627 bytes in 13,979 blocks
==31800==         suppressed: 0 bytes in 0 blocks
==31800== 
==31800== For lists of detected and suppressed errors, rerun with: -s
==31800== ERROR SUMMARY: 30 errors from 30 contexts (suppressed: 0 from 0)
```

Full report: [memleak_vdt_sql_fail_all_feeds.log](https://github.com/wazuh/wazuh/files/10806954/memleak_vdt_sql_fail_all_feeds.log)

## Cause
The problem is caused by disk malfunctions affecting the SQL database, which throws disk errors and returns error codes. This results in some functions not terminating as expected and not freeing the corresponding structure, causing the memory leak.
Example of SQL error causing memory leaks:
```
ERROR: (5503): SQL error: 'database is locked'
```

The problem is in the `wm_vuldet_index_feed` function, where the affected _struct_ would be `wm_vuldet_db`:
https://github.com/wazuh/wazuh/blob/36d17b36a732ecc4b4bffda7ddd047d60a3c79ac/src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c#L4632

PR #16478, opened 2023-03-22
Fix memory leaks in Vulnerability Detector after disk failures

MarcelKemp:
|Related issue|
|---|
|#16265|

## Description

This PR aims to free the memory of the structures that store the feeds when there is a problem, and it does not complete correctly.

In this case, the cause of the issue that has been investigated and found was due to disk errors, which caused the memory of the parsed feeds not to be freed, as it did not finish the flow and did not reach the end of the insert function, where these data structures are freed.

The way to replicate the problem, has been moving the path where the `cve.db` is stored, to a partition with limited storage, that way whenever I tried to insert data to this DB, it showed an error like the following one:
```
wazuh-modulesd:vulnerability-detector: ERROR: (5503): SQL error: 'database or disk is full'
```

## Configuration options

To replicate the error, it is necessary to move the DB to a path with limited storage or cause disk errors. For these tests, the following line has been modified, replacing the path with `/run/lock/`:
https://github.com/wazuh/wazuh/blob/36d17b36a732ecc4b4bffda7ddd047d60a3c79ac/src/wazuh_modules/vulnerability_detector/wm_vuln_detector_db.h#L15

Also, to limit that storage, a file has been created that would take up much of that space in the partition, so that it would only allow the DB to be built and the module to be started, without allowing much more information to be stored:

```
dd if=/dev/urandom of=/run/lock/test bs=1024 count=4400
```

And finally, use has been made of our [custom feeds](https://github.com/wazuh/wazuh-tools/tree/master/utils/vulnerability-detector/custom-feeds), with the configuration found in the README (which is available for _v4.4_), to allow the use of the `valgrind` tool to observe memory leaks.

## Memory report

```
==13447== LEAK SUMMARY:
==13447==    definitely lost: 0 bytes in 0 blocks
==13447==    indirectly lost: 0 bytes in 0 blocks
==13447==      possibly lost: 4,320 bytes in 14 blocks
==13447==    still reachable: 231,565 bytes in 178 blocks
==13447==         suppressed: 0 bytes in 0 blocks
==13447== 
==13447== For lists of detected and suppressed errors, rerun with: -s
==13447== ERROR SUMMARY: 40 errors from 12 contexts (suppressed: 0 from 0)
```

## Tests

  - [x] Linux
- [x] Source installation
- [ ] Package installation
- [x] Source upgrade
- [ ] Package upgrade

- Memory tests for Linux
  - [x] Scan-build report
  - [x] Coverity
  - [ ] AddressSanitizer

### DIFF
--- a/src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
+++ b/src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
@@ -95,6 +95,13 @@ STATIC void wm_vuldet_add_vulnerability_info(wm_vuldet_db *ctrl_block);
  */
 STATIC vulnerability *wm_vuldet_clean_oval_vulnerability_node(vulnerability *vuln_node);
 
+STATIC void wm_vuldet_db_clean(wm_vuldet_db *ctrl_block);
+STATIC void wm_vuldet_clean_vulnerabilities(wm_vuldet_db *ctrl_block);
+STATIC void wm_vuldet_clean_variables(wm_vuldet_db *ctrl_block);
+STATIC void wm_vuldet_clean_info_state(wm_vuldet_db *ctrl_block);
+STATIC void wm_vuldet_clean_info_obj(wm_vuldet_db *ctrl_block);
+STATIC void wm_vuldet_clean_info_test(wm_vuldet_db *ctrl_block);
+
 /**
  * @brief Clean the CVEs linked list data structure.
  * @param ctrl_block Parsed oval data structure.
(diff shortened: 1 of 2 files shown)


damarisg:
## QA review
- **Type**: Manual testing.
- **Testing issue**: https://github.com/wazuh/wazuh-qa/issues/3977
- **Comments**:  https://github.com/wazuh/wazuh-qa/issues/3977#issuecomment-1494657752

vikman90 (approved):

Issue #12906, opened 2022-03-29
`Run agent configuration Interface` option does not work correctly


Rebits:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.3.0 | Agent | Agent | Packages | Windows Server 2019 |

### Description 

During [Windows event manual testing](https://github.com/wazuh/wazuh/issues/12896#issuecomment-1081885456), it has been detected that the `Run agent configuration Interface` option does not work correctly in Windows agent installation. 

https://user-images.githubusercontent.com/11089305/160567860-a6707308-49f3-49cf-98f2-a1dca276b70a.mp4

ncvicchi:
Several attempts were made to achieved correct functionality, all attemps failed, either silently or with a newer error.

- Tried executing it from a command line executed as administrator -> succes. It is a permissions issue.

- Adding:
`<Property Id="MSIUSEREALADMINDETECTION" Value="1" />`

Result: silent failure

- Adding:
```
<Package Id="*"
             Description="Desc"
             Keywords="Installer"
             Manufacturer="$(var.Manufacturer)"
             SummaryCodepage="1252"
             InstallerVersion="200"
             Compressed="yes"
             InstallScope="perMachine"
             AdminImage="yes"
             InstallPrivileges="elevated" />
```
Result: silent failure

- try changing folder permissions:

```
<DirectoryRef Id="APPLICATIONFOLDER">
        <Component Id="INSTALLDIR_perm" Guid="EC4352C1-4240-4E6A-9A5E-E31F22702705" Permanent="yes">
            <CreateFolder>
                <!--Everyone gets full permissions to the folder-->
                <util:PermissionEx User="Everyone" GenericAll="yes" />
            </CreateFolder>
        </Component>  
    </DirectoryRef> 

```
Result: silent failure

- Try removing the following lines (a warning was being raised at compile time):

```

 BinaryKey="WixCA"
 DllEntry="WixShellExec"

```
Result: silent failure

- Try requesting elevated permissions:

```
<Property Id="ALLUSERS" Value="1" />    <!--equals to install="permachine" at package element but this element depricated -->
<Property Id="MSIUSEREALADMINDETECTION" Value="1" /> 

<Condition Message="Please Run as Administrator.">
      Privileged
</Condition>

```

Result: silent failure

- Tried changing execute to deferred:

```

        <CustomAction 	Id="LaunchUI"
        				Directory="APPLICATIONFOLDER"
[...]

PR #16542, opened 2023-03-30
Fix UI launch after successfull installation

ncvicchi:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/12906|

This PR aims to fix the situation where a user requests the agent UI to be launched after installation but the UI won't be launched, failing silently.

## Description

The problem was caused by 2 different situations
- Access to the agent installation folder is completely closed, so the installer had no access to it to execute it
- The agent UI requires elevated privileges to be executed, even with access to the folder, the execution would not occur.

The first situation was solved from the VB script, allowing the currently logged user to access the folder.

The second situation was solved from the Wix script, by allowing impersonation, thus allowing the user to execute the UI (after elevated privileges are granted).

## Tests

  - [X] Linux
  - [X] Windows

### DIFF
--- a/src/win32/InstallerScripts.vbs
+++ b/src/win32/InstallerScripts.vbs
@@ -351,6 +351,10 @@ public function config()
         remUserPerm = "icacls """ & install_dir & """ /remove *S-1-5-32-545 /q"
         WshShell.run remUserPerm, 0, True
 
+        userSID = GetUserSID()
+        grantUserPerm = "icacls """ & install_dir & """ /grant *" & userSID & ":(RX) /t"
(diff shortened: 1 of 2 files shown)


jotacarma90 (approved):
LGTM

juliamagan (approved):
## QA review

- **Type**: Manual testing.
- **Status**: Approved :green_circle:
- **Testing issue**: https://github.com/wazuh/wazuh-qa/issues/4000
- **Comments**: https://github.com/wazuh/wazuh-qa/issues/4000#issuecomment-1511373960

PR #16605, opened 2023-04-03
Metrics module findings to fix

### LINKED ISSUE #16412: Metrics module findings to fix
Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0 | Engine | Manager| - | - |

## Description.
The goal of this task is fix several code review findings in Metrics module. 

## Findings to fix
 - [x] KVDB segfault fix.
 - [x] Replace nlohmann/json by rapidjson library.
 - [x] Use namespace by modules.
 - [x] Use aliases for long namespaces.
 - [x] Review AbstractHandler name, it should have a more specific, it depens on the difference in Chain of Responsibility is the search algorithm or how the values are set.
 - [x] Set an order for headers included in source code files:
	 - HPP of the CPP we have if we are in a CPP
	 - System libraries.
	 - GLIBC / STL
	 - Externals libraries.
	 - Internal libraries.
	 - Others module headers.
 - [x] Do not include in the HPP headers that are only used in the CPP.	 
 - [x] Write the defines with a single _ and ending in HPP, so that it is more descriptive, e.g. _METRICS_METRICS_HPP.
 - [x] Adjust the constant names in the header files to have the format PATH_DIR_FILE_NAME.
 - [x] Document private members.
 - [x] Use doxygen only with /**/ mode, not //.
 - [x] Preferably use variant when returning value, error, otherwise use optional.
 - [x] Avoid copy strings, the arguments of the functions should be const string& if they are going to work with the map or string_view if the map type changes, but not string alone.
 - [x] Use the iterator in map searches.
 - [x] Avoid look for an instrument in a map, an option is register and deliver objects or funtions.
 
 ## DoD
 - [x] All tasks completed. 
 - [x] Unit test run and tested.
 - [x] Clang format checked.

JorgeMarinoDev:
|---|---|---|
| #16412| [User manual](https://docs.google.com/document/d/1N0p7pd7XszQwFU9HBjpB7txIw2DCcAsV-BKfEZSLixw/edit#heading=h.c7k5f8w2i2dv) |  |

## Description.
The goal of this task is fix several code review findings in Metrics module. 

## Findings to fix
 - [x] KVDB segfault fix.
 - [x] Replace nlohmann/json by rapidjson library.
 - [x] Use namespace by modules.
 - [x] Use aliases for long namespaces.
 - [x] Review AbstractHandler name, it should have a more specific, it depens on the difference in Chain of Responsibility is the search algorithm or how the values are set.
 - [x] Set an order for headers included in source code files:
	 - HPP of the CPP we have if we are in a CPP
	 - System libraries.
	 - GLIBC / STL
	 - Externals libraries.
	 - Internal libraries.
	 - Others module headers.
 - [x] Do not include in the HPP headers that are only used in the CPP.	 
 - [x] Write the defines with a single _ and ending in HPP, so that it is more descriptive, e.g. _METRICS_METRICS_HPP.
 - [x] Adjust the constant names in the header files to have the format PATH_DIR_FILE_NAME.
 - [x] Document private members.
 - [x] Use doxygen only with /**/ mode, not //.
 - [x] Preferably use variant when returning value, error, otherwise use optional.
 - [x] Avoid copy strings, the arguments of the functions should be const string& if they are going to work with the map or string_view if the map type changes, but not string alone.
 - [x] Use the iterator in map searches.
 - [x] Avoid look for an instrument in a map, an option is register and deliver objects or funtions.
 
 ## DoD
 - [x] All tasks completed. 
 - [x] Unit test run and tested.
 - [x] Clang format checked.


cborla (approved):
LGTM!

PR #16607, opened 2023-04-03
Incorporate delay load mechanism and explicit preload verification for libraries in the Windows Agent - Implementation

### LINKED ISSUE #16319: Incorporate delay load mechanism and explicit preload verification for libraries in the Windows Agent
|---|---|---|---|---|
| 4.x rev | Wazuh component | Manager/Agent | Packages | Windows |

## Description

This issue aims to incorporate a "delay load" mechanism for libraries to avoid the implicit loading of libraries that Wazuh distributes. In addition, signature verification before explicit library load.

Implicitly loading a library can be problematic because it can result in unintended behavior and potential security risks, especially if the library being loaded is a third-party library.

To avoid these risks, a "delay load" mechanism will be incorporated to ensure that libraries are only loaded explicitly when needed. Additionally, signature verification will be added to ensure that the libraries being loaded are trusted and have not been tampered with.

The main goal of this issue is to prevent third-party libraries from being loaded when signed binaries are copied to another directory to use custom libraries with potentially malicious uses. By using a custom implementation of `LoadLibrary` in the Windows agent's main, we can ensure that libraries are only loaded explicitly when needed and that signature verification is performed to ensure that they are trusted.

## Expected

After this issue is completed, we should expect the following:
- A "delay load" mechanism should be implemented to avoid the implicit loading of libraries that Wazuh distributes.
- Signature verification should be added to ensure that the libraries are explicitly loaded.
- In the Windows agent's main, custom implementation of `LoadLibrary` should be used to delay the loading of libraries.

## DoD

- [x] Incorporate a "delay load" mechanism to avoid the implicit loading of libraries that Wazuh distributes.
- [x] Add signature verification to ensure that the libraries are explicitly loaded.
- [x] Use a custom implementation of `LoadLibrary` in the Windows agent's main to delay the loading of libraries.
- [x] Test thoroughly to ensure that the changes do not affect the behavior of the code.
- [ ] Update any relevant documentation to reflect the changes made.

pereyra-m:
|Related issue|
|---|
|Closes #16319 |

## Description

The objective of this PR is to avoid a malicious actor to use a signed Wazuh binary to run arbitrary code due to DLL hijacking. When an executable is linked dynamically against a library, it is searched by the OS **before** the program begins. If the search order of these libraries isn't properly defined, the current folder will be used and a fake version of the required DLL will be found. Then, any code in `DllMain` for example will be executed without letting the signed verification of the binary being executed.

Three different approaches were used considering each case:

- Delay load: this characteristic consists in dynamically linking a library and don't load it before the program start but only when/if required. This technique was used with `libwazuhshared.dll` and `libwazuhext.dll`. These libraries are generated by Wazuh and always will be located in the same folder as the executable, we only need to check their authenticity before loading them.
- Static linking: if the code resides inside the binary, then no library will be required. The `libwinpthread-1.dll` was statically linked with all the Wazuh .exe but it still is deployed by the installer because some libraries like `libgcc_s_dw2-1.dll` and `libstdc++-6` require it and we can't change those libraries.
- Explicit search location: this technique was used for all the system libraries because we know that they'll be always located in the same place. An XML [manifest](https://learn.microsoft.com/en-us/windows/win32/sbscs/manifests) was embedded in every executable to force the system to look for the listed DLLs in `%windir%\System32\` with the undocumented option **loadFrom.** This manifest also avoids a [DotLocal](https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-redirection) redirection (a simple file/folder alongside the .exe can force it to look for in the current folder).

To make sure any future change creates a DLL dependency that could be exploited, a test was created to run every Wazuh executable with an instance of Process Monitor in the background. If there is a failed attempt to load a DLL outside the Windows folders, the test fails.

Finally, now that we can get to the executable's main function loading only the minimum and trusted libraries, we call `enable_dll_verification()` to make sure that all present and future libraries will be signed. 

## Logs/Alerts example

See the issue for some captures of the work.

## Tests

  - [x] Windows
- [x] Source installation


src/headers/cryptography.h
```diff
@@ -14,6 +14,7 @@
 
 #ifdef WIN32
 #include <windows.h>
+#include "os_err.h"
```
pereyra-m, line 17: This include should be removed

src/win32/win_service.c
```diff
@@ -13,6 +13,7 @@
 #include "shared.h"
 #include "os_win.h"
 #include <winsvc.h>
+#include "dll_load_notify.h"
```
pereyra-m, line 16: This include should be removed

sempervictus:
Thank you

src/win32/win_agent.c
```diff
@@ -20,6 +20,7 @@
 #include "os_execd/execd.h"
 #include "os_crypto/md5/md5_op.h"
 #include "external/cJSON/cJSON.h"
+#include "dll_load_notify.h"
```
MiguelazoDS, line 23: Why is this added?

MiguelazoDS (commented):
GJ! very important feature for the project!

src/win32/win_agent.c
matias-braida: No falta el llamado a "enable_dll_verification()" en el main?

src/win32/setup-syscheck.c
```diff
@@ -13,11 +13,16 @@
 #include "../error_messages/error_messages.h"
 #include <errno.h>
 #define OSSEC_CONFIG_TMP  ".tmp.ossec.conf"
-
+#include "dll_load_notify.h"
 
 /* Enable Syscheck */
 int main(int argc, char **argv)
 {
+#ifdef WIN32
```
matias-braida, line 21: Why the compilation conditional is added here (same as in other .c files inside src/win32)?
That a file is inside the directory src/win32 doesn't guarantee that is for windows compilation?

matias-braida (commented):
Impressive work

src/win32/win_agent.c
pereyra-m: In some previous commits, I had added the `enable_dll_verification()` here. But this executable is used to enable the service also and the verification method caused a delay that prevented Windows to properly start the Wazuh agent service.

The `enable_dll_verification()` is placed in `local_start()` for the `wazuh-agent`.

.github/workflows/windows-dll-hijack.yml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Upload Artifact shared_libraries
+        uses: actions/upload-artifact@v3
+        with:
+          name: shared_libraries
+          path: src/shared_libraries
+      - name: Copy executables
+        run: |
+          mkdir -p src/exe_files
+          # We exclude the installer from the test
+          find src/win32/ -regex '.*wazuh-agent-[0-9].[0-9].[0-9]\.exe$' -delete
```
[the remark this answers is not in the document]
pereyra-m: This regex is incomplete as you point out, thank you.

src/win32/setup-syscheck.c
pereyra-m: I'll check this, it might there unnecessarily.
Thank you.

src/win32/qa/test_dll_hijack.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # Read .csv file
+    with open(PROCMON_CSV_OUTPUT) as csv_file:
+        csv_reader = csv.reader(csv_file, delimiter=',')
+        first_line = True
+        for row in csv_reader:
+            if first_line:
+                first_line = False
+                continue
+            else:
```
tdrauncieal, line 71: the else clause can be avoided, how to other cases below

src/win32/qa/test_dll_hijack.py
```diff
@@ -0,0 +1,79 @@
...
+
+@pytest.fixture(name='current_exe', scope='module', params=list(map(str, list(pathlib.Path(EXE_PATH).glob('*.exe')))))
+def get_exe_path(request):
+    return request.param
+
+def test_dll_not_found(current_exe):
+    # Generate .csv output
+    generate_csv_logs(current_exe)
+
+    # Read .csv file
```
tdrauncieal, line 63: Consider a more simplistic way to first element skip:
image

matias-braida (dismissed):
Great job

src/win32/qa/test_dll_hijack.py
pereyra-m: Fixed.
Thank you

tdrauncieal (dismissed):

MiguelazoDS (dismissed):

tdrauncieal (dismissed):

matias-braida (dismissed):

src/headers/cryptography.h
pereyra-m: No, it shouldn't.

MiguelazoDS (dismissed):

matias-braida (approved):

tdrauncieal (approved):

MiguelazoDS:
# Tests 

During the exploratory (https://github.com/wazuh/wazuh/issues/17138), some issues were found and already addressed [here](https://github.com/wazuh/wazuh/pull/16607/commits/5f6eef2725a9ad256988d2c9a3c22b7883e004ec). 

Results after the tests re-run 

## Overview 
|OS|Result|
|-|-|
|Windows Server 2003 | :green_circle: |
|Windows Server 2016 | :green_circle: |
|Windows 11 22H2 | :green_circle: 
## Windows Server 2003 
## Windows Server 2016 

## Windows 11 22H2

MiguelazoDS (approved):

tdrauncieal (approved):

matias-braida (approved):

Dwordcito (approved):

Issue #16682, opened 2023-04-11
Syscollector JSON exception invalid UTF-8 character


MiguelazoDS:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.3.10| Syscollector | Agent | Packages/Sources | Windows |

# Description

Reported by a user https://github.com/wazuh/wazuh/issues/16652 the issue is related to https://json.nlohmann.me/home/exceptions/#jsonexceptiontype_error316 where an exception is thrown in the function `getItemChecksum()` due to dump().

https://github.com/wazuh/wazuh/blob/538d72232c20f760903e33c9512070984c8f2dcd/src/wazuh_modules/syscollector/src/syscollectorImp.cpp#L911-L917

The exception triggers when a Chinese package is installed. For testing purposes, the package installed was this (https://www.kdocs.cn/)

But during the Syscollector scan the exception is not related directly to this package

![image](https://user-images.githubusercontent.com/13010397/231497256-bf74cad8-40be-44e5-8222-53e8bae96521.png)

# Steps to reproduce

- Windows simplified Chinese (zh-cn) installation. 

<details>
<summary>Expand image</summary>

![image](https://user-images.githubusercontent.com/13010397/232506093-54f7cb2c-3ab7-40f8-a1ff-40662c3a3ced.png)

</details>

- Install a Chinese package. (i.e. https://www.kdocs.cn/)

- Restart the agent to force a syscollector scan 
- An error is shown in the ossec.log file
![image](https://user-images.githubusercontent.com/13010397/232510323-5244dc7d-fb83-47c5-8ff5-30cb86fd1754.png)

**Note: The issue was not reproducible using the same Chinese package on Windows en-us**

MiguelazoDS:
# New findings

Going further into this issue, the cause is not the package itself, but some processes it starts. 

The `sys_package` inventory is properly filled with the proper information. 

![image](https://user-images.githubusercontent.com/13010397/231505852-d05a06b3-8fd3-456a-89d3-ab352c5090db.png)

pereyra-m:
Blocked by https://github.com/wazuh/wazuh/issues/16901

MiguelazoDS:
Blocked by https://github.com/wazuh/wazuh/issues/17186 (created after #16901)

MiguelazoDS:
Cmocka for Windows agent failed. Issue related

https://github.com/wazuh/wazuh-jenkins/issues/5081

PR #16701, opened 2023-04-12
Non UTF8 process information

MiguelazoDS:
|Related issue|
|---|
|#16682|

## Description

Convertion to UTF8 was not being considered for processes, only for packages. 

## Logs/Alerts example

Processes

image

Ports

image

## Tests

  - [x] Windows
- [x] Source installation
- [x] Package installation

### DIFF
--- a/src/data_provider/src/sysInfoWin.cpp
+++ b/src/data_provider/src/sysInfoWin.cpp
@@ -291,8 +291,8 @@ static nlohmann::json getProcessInfo(const PROCESSENTRY32& processEntry)
         SysInfoProcess process(pId, processHandle);
 
         // Current process information
-        jsProcessInfo["name"]       = processName(processEntry);
-        jsProcessInfo["cmd"]        = (isSystemProcess(pId)) ? "none" : process.cmd();
+        jsProcessInfo["name"]       = Utils::EncodingWindowsHelper::stringAnsiToStringUTF8(processName(processEntry));
+        jsProcessInfo["cmd"]        = Utils::EncodingWindowsHelper::stringAnsiToStringUTF8((isSystemProcess(pId)) ? "none" : process.cmd());
         jsProcessInfo["stime"]      = process.kernelModeTime();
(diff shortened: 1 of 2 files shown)


src/data_provider/src/sysInfoWin.cpp
pereyra-m, line 1: This issue has been fixed.
But there could be many hidden hazards in the data provider, because **nlohmann JSON** library will always throws an [exception](https://json.nlohmann.me/api/basic_json/dump/#exceptions) if the string stored isn't UTF-8.

It might not be reasonable to set the proper encoding function to absolutely all the inputs for all OS, but we could discuss with the team the possibility of adding an `error_handler` in the **dump()** method to replace/ignore any invalid data and prevent future exceptions

https://json.nlohmann.me/api/basic_json/error_handler_t/

```c++

enum class error_handler_t {
    strict,
    replace,
    ignore
};

```

pereyra-m (commented):
GJ!

src/data_provider/src/sysInfoWin.cpp
MiguelazoDS: I don't like exceptions, you have to track through all the code to determine where they occur, and this lib relies on them a lot. We could discuss this for sure, but it will require several changes if this is the way to go. Although the exception is thrown at dump() the issue is not there, so would this help us in this case?.

matias-braida (approved):

pereyra-m (approved):
LGTM!

tdrauncieal (approved):

pereyra-m (approved):

damarisg:
## QA review
- **Type**: Manual testing.
- **Testing issue**: https://github.com/wazuh/wazuh-qa/issues/4100
- **Comments**:  https://github.com/wazuh/wazuh-qa/issues/4100#issuecomment-1518054182

matias-braida (approved):

tdrauncieal (approved):

Dwordcito (approved):

PR #16839, opened 2023-04-21
Syscollector running on Windows XP does not retrieve the path of processes correctly - Implementation

### LINKED ISSUE #10698: Syscollector running on Windows XP does not retrieve the path of processes correctly
|---|---|---|---|---|
| master | sysinfo - syscollector | Manager and Agent | Any | Windows XP |

Hi team!

During #7279 exploratory session (#8811), we found that #10011 don't fix processes paths for Windows XP.
Example:

- Expected cmd:
C:\WINDOWS\system32\wbem\wmiprvse.exe
- Actual cmd:
\Device\HarddiskVolume1\WINDOWS\system32\wbem\wmiprvse.exe

Regards,
Nico

pereyra-m:
|Related issue|
|---|
|Closes #10698 |

## Description

This PR adds an alternative method to map the NT path (\Device\HarddiskVolume1, etc.) in processes' cmd field to their corresponding logical drives (C:, D:, etc.)  for Windows XP. 

It was found that the call to `QueryDosDevice()` failed inside `getNtWin32DrivesMap()` with an error like `234: More data is available`. But the buffers already had a size of 65kB.

So the fallback method loads the map with a different approach. The method `QueryDosDevice()` is called again with a null argument to get the full list of drives and then each of them is compared against the `logicalDrives()` list. Only when there is a match, we call `QueryDosDevice()` with that parameter to get the equivalent. All the buffers used have a 32kB lenght.

## Logs/Alerts example

There isn't a Windows XP image available in Github actions. Manual tests were executed with the `sysinfo_test_tool`.
We can see that the path in the **cmd** field contains the logical drive instead of the NT form

## Tests

A simple script was created to test this function in an isolated way

<details>

<summary>
Test program
</summary>

```c++

#define WIN32_LEAN_AND_MEAN
#include <windows.h>
#include <stdio.h>
#include <ntstatus.h>
#include <winternl.h>
#include <ntdef.h>



void* so_get_module_handle(const char *so){
#ifdef WIN32
    char file_name[MAX_PATH] = { 0 };
    snprintf(file_name, MAX_PATH-1, "%s.dll", so);

    HMODULE handle = NULL;
    char full_path[MAX_PATH] = { 0 };

    // Get full path to the module's file.
    // If the function succeeds, the return value is the length of the string that is copied to the buffer,
    // in characters, not including the terminating null character.
    // If the function fails, the return value is NULL.

    if (GetFullPathName(file_name, MAX_PATH, full_path, NULL)) {
        handle = LoadLibrary(full_path);
    }

    return handle;
#else
    char file_name[4096] = { 0 };
#if defined(__MACH__)
    snprintf(file_name, 4096-1, "lib%s.dylib", so);
#else
    snprintf(file_name, 4096-1, "lib%s.so", so);
#endif
    return dlopen(file_name, RTLD_LAZY);
#endif
}

void* so_get_function_sym(void *handle, const char *function_name){
#ifndef WIN32
    return dlsym(handle, function_name);
#else
    return (void *)(intptr_t)GetProcAddress((HINSTANCE)handle, function_name);
#endif
}

void* so_get_module_handle_on_path(const char *path, const char *so){
#ifdef WIN32
    char file_name[MAX_PATH] = { 0 };
    snprintf(file_name, MAX_PATH-1, "%s%s.dll", path, so);
    return LoadLibrary(file_name);
#else
    char file_name[4096] = { 0 };
#if defined(__MACH__)
    snprintf(file_name, 4096-1, "%slib%s.dylib", path, so);
#else
    snprintf(file_name, 4096-1, "%slib%s.so", path, so);
#endif
    return dlopen(file_name, RTLD_LAZY);
#endif
}

#include <iostream>

char * win_strerror(unsigned long error) {
    static TCHAR messageBuffer[4096];
    LPSTR end;

    FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, NULL, error, 0, messageBuffer, sizeof(messageBuffer) / sizeof(TCHAR), NULL);

    if (end = strchr(messageBuffer, '\r'), end) {
        *end = '\0';
    }

    return messageBuffer;
}


int main(int argc, char **argv) {

    char physical[65536];
    char logical[65536];
    char cDrive[65536];
    char dDrive[65536];

    std::cout << "**** " << std::endl;

    void* handle = so_get_module_handle_on_path("C:\\Windows\\System32\\","fltlib");
    if(handle) {
        typedef HRESULT (*filter_get_dos_name_func)(LPCWSTR lpVolumeName, LPWSTR  lpDosName, DWORD dwDosNameBufferSize);
        filter_get_dos_name_func filter_get_dos_name_ptr = (filter_get_dos_name_func)so_get_function_sym(handle, "FilterGetDosName");

        if(filter_get_dos_name_ptr) {
            LPCWSTR lpVolumeName = L"\\Device\\HarddiskVolume1\\";
            LPWSTR lpDosName = new WCHAR[MAX_PATH];
            DWORD dwDosNameBufferSize = MAX_PATH;

            HRESULT hr = filter_get_dos_name_ptr(lpVolumeName, lpDosName, dwDosNameBufferSize);
            if (hr != S_OK) {
                std::cout << "FilterGetDosName failed with error: " << hr << std::endl;
            } else {

                std::cout << "FilterGetDosName: " ;
                std::wcout << lpDosName << std::endl;
            }
            free(lpDosName);
        } else {
            std::cout << "FilterGetDosName not found" << std::endl;
        }
    }
    std::cout << "**** " << std::endl;

    if ( argc > 1) {
        for (int i=1; i<argc; i++) {
            QueryDosDevice(argv[i],logical, sizeof(logical));
            std::cout << argv[i] << " : \t" << logical << std::endl << std::endl;
        }
        return 0;
    }
    int ret;
    ret = QueryDosDevice("C:", cDrive, sizeof(cDrive));
    if (ret==0)
    {
        unsigned long err = GetLastError();
        std::cout << "QueryDosDevice failed with error: " << err<< " . Err:  "<<  win_strerror(err) << std::endl;
    }

    ret = QueryDosDevice("D:", dDrive, sizeof(dDrive));
    if (ret==0)
    {
        unsigned long err = GetLastError();
        std::cout << "QueryDosDevice failed with error: " << err<< " . Err:  "<<  win_strerror(err) << std::endl;
    }
    std::cout << "C: : \t" << cDrive << std::endl << std::endl;
    std::cout << "D: : \t" << dDrive << std::endl << std::endl;

    ret = QueryDosDevice(NULL, physical, sizeof(physical));
    if (ret==0)
    {
        unsigned long err = GetLastError();
        std::cout << "QueryDosDevice failed with error: " << err<< " . Err:  "<<  win_strerror(err) << std::endl;
    }

    std::cout << "devices: " << std::endl;
    std::cout << "physical: \t"<< physical << std::endl << std::endl;

    for (char *pos = physical; *pos; pos+=strlen(pos)+1) {
        if(QueryDosDevice(pos, logical, sizeof(logical))) {
            std::cout << pos << " : \t" << logical << std::endl << std::endl;
        } else {
            std::cout << "FAILED!" << std::endl;
        }
    }

    std::cout << MAX_PATH << std::endl;

    return 0;
}


```

</details>

  - [x] Windows
- [x] Source installation


MiguelazoDS (dismissed):

src/data_provider/src/sysInfoWin.cpp
```diff
@@ -207,6 +207,43 @@ class SysInfoProcess final
...
+                {
+                    const auto tokens = Utils::splitNullTerminatedStrings(spPhysicalDevices.get());
+                    const auto spDosDevice { std::make_unique<char[]>(OS_SIZE_32768) };
+
+                    for (const auto& token : tokens)
+                    {
+                        // Checking if token is found in logicalDrives to avoid a large map with unnecessary volumes.
+                        // logicalDrives contains a slash at the end but the result of QueryDosDevice doesn't.
+                        if (std::find_if(logicalDrives.begin(), logicalDrives.end(), [&](const auto & logicalDrive)
+                    {
```
tdrauncieal, line 226: Could be arrange better the lambda definition here

pereyra-m: It is indeed hard to read. The Astyle tool forced this strange writing.
I made a commit to fix it.

MiguelazoDS (dismissed):

matias-braida (dismissed):

GabrielEValenzuela (dismissed):
GJ!

Dwordcito (dismissed):

Dwordcito (approved):

PR #16854, opened 2023-04-24
Remove std::any from Engine's function helper builders

### LINKED ISSUE #16853: Engine - Remove std::any from helper builders
|---|---|---|---|---|
|5.0.0 | Engine | Manager | Packages/Sources | OS version |

As part of https://github.com/wazuh/wazuh/issues/16688
To simplify and get compile errors when using helper function builders the signature will no longer use std::any. Instead a new registry only of helpers will be added, being the new signature the parameters it needs directly.

JavierBejMen:
|Related issue|
|---|
|Closes #16853|

## Changes
- The registry class now has the type of builder as a template parameter.
- A new registry only for function helpers has been added.
- Helper builders now have the signature `std::function<base::Expression(const std::string&, const std::string&, const std::vector<std::string>&)>`


juliancnn (approved):
LGTM

LucioDonda (approved):
LGTM

Issue #16677, opened 2023-04-11
Improve logical expression


cborla:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0 | Engine | Manager| - | - |

## Description

The new engine needs expand the logical expression capabilities, supporting varios data types, and comparators.

- [x]  Include all json data types.
        For `==` and ` !=` support all types of JSON data, numerical, string, array, object, boolean and null.
- [x]  For Operators `>` `<` `>=` `<=`
        Only support for numbers.
- [x]  Change in the build to make it support all operators. 
        `stageBuilderCheck` function.
- [x]  Identifies whether an operator is inside a helper function and takes it as a string.
- [x]  Improve the UT for stageBuilderCheck_test.cpp.
- [x]  Delete an extra file that is empty.

## DoD
 - [x] All tasks completed. 
 - [x] Unit test run and tested.
 - [x] Clang format checked.
 - [x] Documentation updated.

NahuFigueroa97:
##   :green_circle: Logical expression functionality

#### :green_circle: < <= > >== != == for numbers and strings

- Int Helper Example

```
parse:
  logpar:
    - event.original: <message>

normalize:
  - map:
      - ~json: +parse_json/$message
  - check: ~json<3
    map:
      - event.action: "is less than 3"
  - check: ~json<=3
    map:
      - event.action: "is less equal than 3"
  - check: ~json>3
    map:
      - event.action: "is greate than 3"
  - check: ~json>=3
    map:
      - event.action: "is greate equal than 3"  
  - check: ~json==3
    map:
      - event.action: "is equal than 3"
  - check: ~json!=3
    map:
      - event.action: "is not equal than 3"
```

```
    decoder/test/0  ->  success
        [decoder/test/0] [condition.value[/wazuh/queue==103]] -> Success
        [decoder/test/0] [/event/original: <message>] -> Success
        [decoder/test/0] [condition]:success
        [decoder/test/0] [helper.parse_json[/~0json]] -> Success
        [decoder/test/0] [check: ~json<3] -> Success
        [decoder/test/0] [map.value[/event/action="is less than 3"]] -> Success
        [decoder/test/0] [check: ~json<=3] -> Success
        [decoder/test/0] [map.value[/event/action="is less equal than 3"]] -> Success
        [decoder/test/0] [check: ~json>3] -> Failure
        [decoder/test/0] [check: ~json>=3] -> Failure
[...]

PR #16863, opened 2023-04-25
Improve logical expression

NahuFigueroa97:
|---|---|---|
| #16677 |[Link](https://docs.google.com/document/d/1N0p7pd7XszQwFU9HBjpB7txIw2DCcAsV-BKfEZSLixw/edit#bookmark=kix.6d9amw6d92gn) | Testing issue |

## Description

The new engine needs expand the logical expression capabilities, supporting varios data types, and comparators.

- [x]  Include all json data types.
        For `==` and ` !=` support all types of JSON data, numerical, string, array, object, boolean and null.
- [x]  For Operators `>` `<` `>=` `<=`
        Only support for numbers.
- [x]  Change in the build to make it support all operators. 
        `stageBuilderCheck` function.
- [x]  Identifies whether an operator is inside a helper function and takes it as a string.
- [x]  Improve the UT for stageBuilderCheck_test.cpp.
- [x]  Delete an extra file that is empty.

### DIFF
--- a/src/engine/source/builder/builders/stageBuilderCheck.cpp
+++ b/src/engine/source/builder/builders/stageBuilderCheck.cpp
@@ -33,7 +33,7 @@ base::Expression stageBuilderCheckList(const std::any& definition,
     if (!jsonDefinition.isArray())
     {
         throw std::runtime_error(fmt::format("Check stage: Invalid json definition type: "
-                                             "expected \"array\" but got \"{}\"",
+                                             "expected \"array\" but got \'{}\'",
                                              jsonDefinition.typeName()));
     }
 
@@ -47,14 +47,14 @@ base::Expression stageBuilderCheckList(const std::any& definition,
                        if (!condition.isObject())
                        {
                            throw std::runtime_error(fmt::format("Check stage: Invalid array item type, expected "
-                                                                "\"object\" but got \"{}\"",
+                                                                "\"object\" but got \'{}\'",
                                                                 condition.typeName()));
                        }
                        if (condition.size() != 1)
                        {
                            throw std::runtime_error(
                                fmt::format("Check stage: Invalid object item size, expected exactly "
-                                           "one key/value pair but got \"{}\"",
+                                           "one key/value pair but got \'{}\'",
                                            condition.size()));
                        }
                        return registry->getBuilder("operation.condition")(condition.getObject().value()[0]);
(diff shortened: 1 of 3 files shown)


src/engine/test/source/builder/builders/stageBuilderCheck_test.cpp
```diff
@@ -100,24 +100,151 @@ TEST(StageBuilderCheckTest, ListBuildsCorrectExpression)
...
+                    std::make_tuple(Json {R"("field<=null")"},
+                                    "Check stage: The \"<=\" operator only allows operate with numbers or string"),
+                    std::make_tuple(Json {R"("field>={\"key\":\"value\"}")"},
+                                    "Check stage: The \">=\" operator only allows operate with numbers or string"),
+                    std::make_tuple(Json {R"("field>=[\"value1\",\"value2\"]")"},
+                                    "Check stage: The \">=\" operator only allows operate with numbers or string"),
+                    std::make_tuple(Json {R"("field>=false")"},
+                                    "Check stage: The \">=\" operator only allows operate with numbers or string"),
+                    std::make_tuple(Json {R"("field>=null")"},
+                                    "Check stage: The \">=\" operator only allows operate with numbers or string")));
```
JorgeMarinoDev, line 222: Are we sure Check stage all cases in Testing::values map to the same fmt string "`Check stage: The \">=\" operator only allows operate with numbers or string""`

src/engine/source/builder/builders/stageBuilderCheck.cpp
```diff
@@ -103,39 +95,134 @@ base::Expression stageBuilderCheckExpression(const std::any& definition,
             field = term.substr(pos1 + 1, pos2 - pos1 - 1);
             value = term.substr(0, pos1) + term.substr(pos2);
```
JavierBejMen, line 96: We don't set valueJson when is a helper function, this results in the operation builder generating a function to check if the field is null and not the desired function. Add tests for this use case and fix the implementation.

JavierBejMen (changes requested):

src/engine/test/source/builder/builders/stageBuilderCheck_test.cpp
JorgeMarinoDev: Disregard. Clear now.

JorgeMarinoDev (approved):
LGTM

cborla (approved):
LGTM

juliancnn (approved):
LGTM!

PR #16865, opened 2023-04-25
fix(integration test): use docker compose v2

### LINKED ISSUE #16812: Update integration test python pytest scripts to work with docker compose V2
|Wazuh version|Component|
|---|---|
| X.Y.Z-rev | Wazuh component |

## Description
The integration test README states that they depend on docker compose v1. However, upon inspection of its usage, it seems to be possible to rework them to use the v2. This is important because [Docker will remove the v1 from Docker Desktop at the end of June](https://docs.docker.com/compose/release-notes/).

## Checks

The following elements have been updated or reviewed (should also be checked if no modification is required):
- [x] Tests (unit tests, API integration tests).
- [x] Changelog.
- [x] Documentation.
- [x] Integration test mapping (using `api/test/integration/mapping/_test_mapping.py`).

aradkdj:
|Related issue|
|---|
|#16812|

### Context
[As announced in January](https://www.docker.com/blog/new-docker-compose-v2-and-v1-deprecation/), from the end of June (two months after originally planned) Docker will drop support for Compose V1 and as a result Docker Desktop will no longer include Compose V1 (`docker-compose`). As part of our integration testing runs, a cluster is brought up. This cluster is managed via Compose V1. 

### Solution
This PR fixes #16812 by introducing the necessary changes to manage said cluster via Compose V2. This update should also somewhat speed the cluster creation, since
Compose V2 uses Buildkit by default.
- Replace `docker-compose` with `docker compose`.
- Replace _ with - as word separator in container names, following the new default.

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade

### DIFF
--- a/api/test/integration/conftest.py
+++ b/api/test/integration/conftest.py
@@ -34,7 +34,7 @@ cluster_env_mode = 'cluster'
 
 
 def pytest_addoption(parser):
-    parser.addoption('--nobuild', action='store_false', help='Do not run docker-compose build.')
+    parser.addoption('--nobuild', action='store_false', help='Do not run docker compose build.')
 
 
 def pytest_collection_modifyitems(items: list):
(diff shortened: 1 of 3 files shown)


nico-stefani (changes requested):
After executing the tests all the containers still are up and running.
image
Please review this because the normal behavior is to remove all the environment.

aradkdj:
This should be fixed now.

nico-stefani (changes requested):
Why are you deleting the section `profiles` from the compose?

I'm trying to run with the marker `standalone`, and it isn't working.

image

image

aradkdj:
Having profiles for `wazuh-master`, `wazuh-agent{1-8}` and `nginx-lb` was redundant as the compose file spec dictates that services without a profile are [always enabled](https://docs.docker.com/compose/compose-file/05-services/#profiles) (and they were part of every profile).

I can't reproduce your issue.
image

nico-stefani:
> Having profiles for `wazuh-master`, `wazuh-agent{1-8}` and `nginx-lb` was redundant as the compose file spec dictates that services without a profile are [always enabled](https://docs.docker.com/compose/compose-file/05-services/#profiles) (and they were part of every profile).
> 
> I can't reproduce your issue. image

Sorry, my bad I try again and run without problems. It was a matter of time.

nico-stefani (approved):
LGMT

PR #16878, opened 2023-04-26
Removing failing check (1.1.5) from Win 10 SCA file

### LINKED ISSUE #15554: Resolve a SCA Windows 10 check 15004 false positive
|---| --- | --- |
| 4.3.x | SCA | Error|

As reported by Colin Edwards there is a potential false negative when checking Windows 10 (maybe others) SCA check 15004 Ensure 'Password must meet complexity requirements' is set to 'Enabled'

Colin post: 

The SCA for Win 10, cis_win10_enterprise.yml, contains ID 15004 to validate that password complexity is enabled, but the detection doesn't work correctly.  The same may be true for SCA for other versions of Windows...I haven't checked.

The rule is defined as 'c:powershell Get-ADDefaultDomainPasswordPolicy -Current LoggedOnUser -> r:ComplexityEnabled\s+: True'

In most environments, this is going to be a false failure (false-negative) for 2 reasons that I know of:
1 - The system executing the PowerShell command needs to have the ActiveDirectory PowerShell module installed, and that usually only exists on sysadmin systems.  If you run it on a system that doesn't have the module, it will return an error...

2 - If the AD ps module is installed, the command still needs to be executed by a user on the domain for it to actually retrieve the policy.  Forgive me if this isn't 100% accurate because I'm new to using Wazuh, but I believe everything executed for SCA is done by the Agent Service, so it runs as System, and System can't retrieve the domain policy.

This Issue aims to investigate and resolve in the next SCA policy releases as needed.

fabamatic:
|Related issue|
|---|
|Closes #15554 |

## Description

This PR removes check 1.1.5 from Windows 10 SCA file. The command being used was not effective at verifying this step and caused false positives.

The rule was defined as 'c:powershell Get-ADDefaultDomainPasswordPolicy -Current LoggedOnUser -> r:ComplexityEnabled\s+: True'

In most environments, this is going to be a false failure (false-negative) for 2 reasons:
1 - The system executing the PowerShell command needs to have the ActiveDirectory PowerShell module installed, and that usually only exists on sysadmin systems. If it's on a system that doesn't have the module, it will return an error

2 - If the AD ps module is installed, the command still needs to be executed by a user on the domain for it to actually retrieve the policy. As the SCA is executed by the Agent Service, it's runs as System, and System can't retrieve the domain policy.

No alternatives provided satisfactory results for this check

### DIFF
--- a/ruleset/sca/windows/cis_win10_enterprise.yml
+++ b/ruleset/sca/windows/cis_win10_enterprise.yml
@@ -77,18 +77,6 @@ checks:
       - 'c:net.exe accounts -> n:Minimum password length:\s+(\d+) compare >= 14'
 
   - id: 15504
-    title: "Ensure 'Password must meet complexity requirements' is set to 'Enabled'."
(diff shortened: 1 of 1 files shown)


72nomada (approved):
LGTM

Issue #16981, opened 2023-05-02
String object has not attribute 'timestamp' in cluster.log


Selutario:
|Wazuh version|Component|
|---|---|
| 4.4.2-rc1 | Wazuh cluster |

## Description

@juliamagan reported this error is being raised in the master node of clusters with version 4.4.2:
```
2023/05/02 14:16:27 ERROR: [Worker wazuh-worker2] [Main] Unhandled error processing request 'b'syn_w_g_e'': 'str' object has no attribute 'timestamp'
Traceback (most recent call last):
  File "/var/ossec/framework/python/lib/python3.9/site-packages/wazuh-4.4.2-py3.9.egg/wazuh/core/cluster/common.py", line 799, in dispatch
    command, payload = self.process_request(command, payload)
  File "/var/ossec/framework/python/lib/python3.9/site-packages/wazuh-4.4.2-py3.9.egg/wazuh/core/cluster/master.py", line 264, in process_request
    start_time = self.send_agent_groups_status['date_start'].timestamp()
AttributeError: 'str' object has no attribute 'timestamp'
```

The reason is very possibly related to the changes performed here:
- https://github.com/wazuh/wazuh/pull/16394

## Checks

The following elements have been updated or reviewed (should also be checked if no modification is required):
- [ ] Tests (unit tests, API integration tests).
- [ ] Changelog.
- [ ] Documentation.
- [ ] Integration test mapping (using `api/test/integration/mapping/_test_mapping.py`).

RamosFe:
## Update

All the status properties inside the `MasterHandler` are saved as a `String` containing the date of the event.
Example:
https://github.com/wazuh/wazuh/blob/13228ade7786495fd6d75f34775efcfc30eeee64/framework/wazuh/core/cluster/master.py#L693-L696

The data from `send_full_agent_groups_status` and `sync_agent_info_status` is used in the `process_request` method as a `start_date`, so it needs to be converted to a `date` type. The following code was added to use a type guard in case the value is an `String` and convert it to the desired `date` type:

https://github.com/wazuh/wazuh/blob/13228ade7786495fd6d75f34775efcfc30eeee64/framework/wazuh/core/cluster/master.py#L262-L277

PR #16986, opened 2023-05-02
Handle case where the "date_start" is a str

RamosFe:
|Related issue|
|---|
| [#16981](https://github.com/wazuh/wazuh/issues/16981) |

## Description

All the status properties inside the `MasterHandler` are saved as a `String` containing the date of the event.
Example:
https://github.com/wazuh/wazuh/blob/13228ade7786495fd6d75f34775efcfc30eeee64/framework/wazuh/core/cluster/master.py#L693-L696

The data from `send_full_agent_groups_status` and `sync_agent_info_status` is used in the `process_request` method as a `start_date`, so it needs to be converted to a `date` type. The following code was added to use a type guard in case the value is an `String` and convert it to the desired `date` type:

https://github.com/wazuh/wazuh/blob/13228ade7786495fd6d75f34775efcfc30eeee64/framework/wazuh/core/cluster/master.py#L262-L277

### DIFF
--- a/framework/wazuh/core/cluster/master.py
+++ b/framework/wazuh/core/cluster/master.py
@@ -261,11 +261,19 @@ class MasterHandler(server.AbstractServerHandler, c_common.WazuhCommon):
             return self.process_sync_error_from_worker(data)
         elif command == b'syn_w_g_e':
             logger = self.task_loggers['Agent-groups send']
-            start_time = self.send_agent_groups_status['date_start'].timestamp()
+            start_time = self.send_agent_groups_status['date_start']
+            if isinstance(start_time, str):
+                start_time = datetime.strptime(start_time, DECIMALS_DATE_FORMAT)
+
+            start_time = start_time.timestamp()
             return c_common.end_sending_agent_information(logger, start_time, data.decode())
         elif command == b'syn_wgc_e':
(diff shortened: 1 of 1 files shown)


framework/wazuh/core/cluster/master.py
```diff
@@ -261,11 +261,19 @@ def process_request(self, command: bytes, data: bytes) -> Tuple[bytes, bytes]:
...
+            start_time = self.send_agent_groups_status['date_start']
+            if isinstance(start_time, str):
+                start_time = datetime.strptime(start_time, DECIMALS_DATE_FORMAT)
+
+            start_time = start_time.timestamp()
             return c_common.end_sending_agent_information(logger, start_time, data.decode())
         elif command == b'syn_wgc_e':
             logger = self.task_loggers['Agent-groups send full']
-            start_time = self.send_agent_groups_status['date_start'].timestamp()
+            start_time = self.send_agent_groups_status['date_start']
```
Selutario, line 272:
```suggestion
            start_time = self.send_full_agent_groups_status['date_start']
```

Selutario (changes requested):

Selutario (approved):

davidjiglesias (approved):

Issue #16855, opened 2023-04-24
Engine - Develop and integrate in-memory variable functionality for components


JavierBejMen:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
|5.0.0 | Engine | Manager | Packages/Sources | OS version |

Develop a system for in-memory variables to be used in the components. It will replace/upgrade the definition module. For now, it will be constant, these variables won't be modified on runtime. This system should be modular and require the least work when developing helper functions.

Study the viability of maintaining/changing the current syntax to refer to these variables in the YAML.

PR #17016, opened 2023-05-05
Engine in memory vars

JavierBejMen:
|Related issue|
|---|
|#16855|

Added module defs, to handle definitions. Builders have been modified to include the definition handler as parameter, this is due to definitions being defined for each asset.

The definition module allows to check if a reference points to a definition and to get its value. It also exposes functionality to replace definitions in a string with the value of definitions. If the value is a string is replaced with it, if the value is another type it is replaced with the string representation.

Each builder or actor using the definition handlers is responsible for managing or processing the definitions.

Helpers, logical expressions and logpars now supports definitions.


nbertoldo (approved):
LGTM!

juliancnn (approved):
LGTM!

Issue #17032, opened 2023-05-08
Segfault in analysisd due to multiple threads accesing FTS `OSList`


jotacarma90:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.3 | Analysisd | Manager | All | Linux |

# Description

Hello team, found multiple segfaults related to analysisd's FTS function:
```
(gdb) bt
#0  0x0000000000416c48 in FTS (lf=0x7f1e700d7ff0, fts_list=0x852390 <os_analysisd_fts_list>, 
    fts_store=0x852388 <os_analysisd_fts_store>) at analysisd/fts.c:321
#1  0x000000000042a264 in OS_CheckIfRuleMatch (lf=0x7f1e700d7ff0, last_events=0x27c7190, 
    cdblists=0x8523a0 <os_analysisd_cdblists>, curr_node=0x3010700, rule_match=0x7f1e1e793e10, 
    fts_list=0x852390 <os_analysisd_fts_list>, fts_store=0x852388 <os_analysisd_fts_store>, 
    save_fts_value=true, rules_debug_list=0x0) at analysisd/rules.c:2908
#2  0x000000000042abd2 in OS_CheckIfRuleMatch (lf=0x7f1e700d7ff0, last_events=0x27c7190, 
    cdblists=0x8523a0 <os_analysisd_cdblists>, curr_node=0x2987490, rule_match=0x7f1e1e793e10, 
    fts_list=0x852390 <os_analysisd_fts_list>, fts_store=0x852388 <os_analysisd_fts_store>, 
    save_fts_value=true, rules_debug_list=0x0) at analysisd/rules.c:3110
#3  0x000000000043bf88 in w_process_event_thread (id=0x1) at analysisd/analysisd.c:2038
#4  0x00007f1e9e955ea7 in __pthread_create_2_1 (newthread=<optimized out>, attr=<optimized out>, 
    start_routine=<optimized out>, arg=<optimized out>) at pthread_create.c:871
```

The bug seems to be in the `OSList` functionality that is used to store the fts events.

Analyzing the coredumps with gdb, I have been able to observe that the list has the following status:
```
(gdb) p **fts_list
[...]

PR #17126, opened 2023-05-15
FTS list race condition crash

ncvicchi:
|Related issue|
|---|
|[#17032](https://github.com/wazuh/wazuh/issues/17032)|

## Description

When on high RPMs values, eventually a race condition might occur that generates 
crash when a thread frees a list being used by another thread.

On systems with this situation, the race condition was avoided by configuring a
single thread, with positive results.

To prevent the race condition from happening when multiple threads are allowed,
a mutex was placed to synchronize threads and prevent this from happening.

### DIFF
--- a/src/analysisd/fts.c
+++ b/src/analysisd/fts.c
@@ -306,6 +306,8 @@ char * FTS(Eventinfo *lf, OSList **fts_list, OSHash **fts_store)
         }
     }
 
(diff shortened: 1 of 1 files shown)


vikman90:
## Coverity scan

Snapshot ID: 68153.

No defects were introduced or fixed.

jotacarma90 (approved):
LGTM

PR #17140, opened 2023-05-16
Fix windows integration

fabamatic:
|Related issue|
|---|
||

## Description

This PR fixes an issue were  Windows integration wasn't ingesting logs correctly
This change identifies windows logs using event.module field so that they can correctly be parsed by Windows integration decoders

Correct decoding present in UI
image

## Configuration options

A windows Wazuh agent with event channel collection is required

## Logs/Alerts example

image


JcabreraC (approved):
LGTM !

PR #17154, opened 2023-05-18
Fix Infinite loop of test command

LucioDonda:
|Related issue|
|---|
|#17147|

## Description

With this change the test command is able to be used with redirection (from a single line like using echo command, or with several lines like cat to a several-lines file) and with the user keyboard input as it's mainly used.

`╰─# cat file.txt | wazuh-engine test` 
Will print an output and assets result for each line in file.txt

`╰─# echo 'some event' | wazuh-engine test`
Prints the output and assets involved in event containing "some event" as log line.

`╰─# wazuh-engine test`
Will start a test session, waiting for the user to enter a log line.

As described on issue #17147, it is required to improve the `test` command to be able to process large example logs.
To achieve this goal it's necesary disable temporaly the canonical mode. [documentation](https://man7.org/linux/man-pages/man3/termios.3.html)

### DIFF
--- a/src/engine/source/cmds/src/test.cpp
+++ b/src/engine/source/cmds/src/test.cpp
@@ -350,7 +350,10 @@ void run(const Options& options)
 
     }
 
-    clear_icanon(false);
+    if (isatty(fileno(stdin)))
+    {
+        clear_icanon(false);
+    }
     g_exitHanlder.execute();
 }
 
--- a/src/engine/source/cmds/src/test.cpp
+++ b/src/engine/source/cmds/src/test.cpp
@@ -252,14 +252,17 @@ void run(const Options& options)
 
     std::cout << std::endl << std::endl << "Enter a log in single line (Crtl+C to exit):" << std::endl << std::endl;
 
-    // Stdin loop
-    std::string line;
-
-    if(!clear_icanon(true))
+    // Only set non-canonical mode when connected to terminal
+    if (isatty(fileno(stdin)))
     {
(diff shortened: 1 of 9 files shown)


src/engine/source/cmds/src/test.cpp
```diff
@@ -38,6 +39,33 @@ void sigint_handler(const int signum)
...
+int clear_icanon(bool flag)
+{
+    struct termios settings;
+
+    if (tcgetattr(STDIN_FILENO, &settings) < 0)
+    {
+        return 0;
+    }
+
+    if (true == flag)
```
cborla, line 52: `if(flag)` :)

LucioDonda: Done!

src/engine/source/cmds/src/test.cpp
```diff
@@ -222,93 +250,107 @@ void run(const Options& options)
...
-        catch (const std::exception& e)
+        else
         {
-            LOG_ERROR("Engine 'test' command: An error occurred while parsing a message: {}.", e.what());
+            std::cout << std::endl << std::endl << "Enter a log in single line (Crtl+C to exit):" << std::endl << std::endl;
         }
+
     }
 
+    clear_icanon(false);
```
cborla, line 353: For consistency, this section should also include the following condition:

`if (isatty(fileno(stdin)))`

cborla (approved):
LGTM

JavierBejMen (approved):
LGTM

JcabreraC (approved):
LGTM !

Issue #17233, opened 2023-05-24
Engine - Test - Create session commands


cborla:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
|5.0.0 | Engine | Manager | Packages/Sources | OS version |

## Description

The idea of this improvement is to give the possibility to select with which policy the test is executed, making it as efficient as possible. Therefore, the concept of session is born, which contains a name, a route/policy to test on, and a lifetime, which will determine how many minutes the session will remain created.
At the same time, by introducing the session concept, the creation of policies and controllers for each test query is avoided.  

### Requirements
- Create a set of commands to manage sessions, through the console and the API.

Subcommand
- [ ] New [SessionName] [PolicyName] [SessionLifeSpan=optional] [SessionLifeSpan=optional] 
    - SessionName: name of the session to create. In case it exists, it returns an error.
    - PolicyName: name of the policy to be routed for that session.
    - SessionLifeSpan: session lifetime (define time unit).
- [ ] List 
    - List all the sessions created.
- [ ] Get [SessionName].
    - Lists a session with its attributes.
Returns the list of names, the policy, the creation date and the set lifeSpan (if lifeTime is not set, "Unlimited" will be printed).
- [ ] Remove [SessionName | removeAll=optional].
    - SessionName: name of the session to be removed.
    - RemoveAll: removes all sessions.
- [ ] Run [SessionName=optional] [event].
    - SessionName: The label added to the event to identify it in the router. If the session name is not included, no label is added to the event.

## DoD
 - [ ] All tasks completed. 
 - [ ] Unit test run and tested.
 - [ ] Clang format checked.
 - [ ] Documentation updated.

PR #17363, opened 2023-05-30
Engine - Handle Test sessions

nmkoremblum:
|Related issue|
|---|
|#17233|

## Description

This pull request addresses the requirements stated in issue #17233 by introducing a new object designed explicitly for managing test sessions. Alongside the object, a set of API commands is implemented to facilitate the utilization of the event test functionality.

Changes Made:

- Implemented a new object responsible for handling test sessions, enabling efficient selection of testing policies while ensuring optimal execution.
- Developed a console-based and API-based command set for managing sessions, allowing seamless control over session creation, listing, retrieval, and removal.
- Introduced the concept of sessions with customizable attributes, including a unique name, the associated policy for routing, and an optional session lifespan.
- Enhanced the session management system to avoid the need for separate policies and controllers for each test query, streamlining the overall workflow.


src/engine/source/api/src/test/sessionManager.cpp
```diff
@@ -0,0 +1,124 @@
...
+            fmt::format("Policy '{}' is already assigned to a route ('')", policyName, m_policyMap[policyName])};
+    }
+
+    // TODO: interact with router to get a valid route
+    // routeName = getRoute(policyName);
+    auto routeName = "routeName";
+
+    Session session(sessionName, policyName, routeName, lifespan);
+    m_activeSessions.emplace(sessionName, session);
+    m_routeMap[routeName] = sessionName;
```
NahuFigueroa97, line 41: improve-desing: Why isn't the emplace() method still used? with this you avoid creating explicit temporary objects

src/engine/source/api/include/api/test/sessionManager.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+    static SessionManager& getInstance(void);
+
+    std::optional<base::Error>
+    createSession(const string& sessionName, const string& policyName, uint32_t lifespan = 0);
+
+    std::vector<string> getSessionsList(void);
+    std::optional<Session> getSession(const string& sessionName);
+
+    bool removeSessions(const bool removeAll, const string sessionName = "");
```
NahuFigueroa97, line 73: question - desing: Why removeSessions() has a parameter to remove all sessions and then there is removeAllSessions() which already does it? Wouldn't a single method with default paramatters be better?

NahuFigueroa97 (commented):
Great job! just a few small observations

src/engine/source/api/include/api/test/sessionManager.hpp
```diff
@@ -0,0 +1,91 @@
...
+struct Session
+{
+public:
+    const std::string m_filterName;
+    const std::string m_policyName;
+    const std::string m_routeName;
+    const std::string m_sessionID;
+    const std::string m_sessionName;
+    const std::time_t m_creationDate;
+    const uint32_t m_lifespan; ///< Session m_lifespan in seconds. 0 means no expiration.
```
JorgeMarinoDev, line 34: **thouhgt:** why public, if we have getters afterward?

src/engine/source/api/src/test/handlers.cpp
```diff
@@ -0,0 +1,412 @@
...
+
+    int32_t minAvailablePriority {MINIMUM_PRIORITY};
+
+    // The condition may be confusing as the actual priority increases while the value decreases
+    while (takenPriorities.count(minAvailablePriority) > 0 && MAXIMUM_PRIORITY > minAvailablePriority)
+    {
+        // If priority is taken, decrease the value (so, increase the priority)
+        minAvailablePriority--;
+    }
+
```
JorgeMarinoDev, line 56: **thought:** In `/src/engine/source/router/src/router.cpp` we can see at line 209 that `std::vector<Router::Entry> Router::getRouteTable()` returns a sorted table by Priority. Wouldn't this mean that despite the confusing order of priorities we can get the max or min priority to get the minAvailablePriority avoiding this loop?

src/engine/source/api/src/test/handlers.cpp
```diff
@@ -0,0 +1,412 @@
...
+    if (error.empty())
+    {
+        const auto postResourceError = catalog->postResource(targetResource, filterContent);
+        if (postResourceError)
+        {
+            error = postResourceError.value().message;
+        }
+    }
+
+    return (error.empty() ? std::nullopt : std::make_optional(base::Error {error}));
```
JorgeMarinoDev, line 94: **thought:** Why calling `std::make_optional` here instead of just returning `base::Error(error)` ? If we had a more complex process to create the the base::Error object, it would definitely make more sense to use make_optional, but I don't think it's needed here because error is not constexpr and there are no narrowing conversions avoided either. std::optional is "smart" enough and it might improve readability. Thinking Out Loud. Please disregard is there is another valid reason to use make_optional.

src/engine/source/api/src/test/handlers.cpp
```diff
@@ -0,0 +1,412 @@
...
+    catalog::Resource targetResource;
+    try
+    {
+        base::Name name {filterName};
+        targetResource = catalog::Resource {name, catalog::Resource::Format::json};
+    }
+    catch (const std::exception& e)
+    {
+        error = e.what();
+    }
```
JorgeMarinoDev, line 119: **thought:** catalog::Resource constructor only throws std::runtime_error and here we are catching std::exception. Should we add a default catch(...) block here or add the **noexcept** modifier to `deleteFilterFromStore` . It seems that the idea is to wrap any exception to the base::Error message. Please disregard if any other code in the stack deals with any other exception, in which case no re-trhown is here to bubble them up.

src/engine/source/api/src/test/handlers.cpp
```diff
@@ -0,0 +1,412 @@
...
+        // If the request is not valid, return the error
+        if (std::holds_alternative<api::wpResponse>(res))
+        {
+            return std::move(std::get<api::wpResponse>(res));
+        }
+
+        const auto& eRequest = std::get<RequestType>(res);
+
+        const auto errorMsg =
+            !eRequest.has_name() && !eRequest.has_removeall()
```
JorgeMarinoDev, line 356: **thought:** Wouldn't it be safer to enclose the logic expression ` !eRequest.has_name() && !eRequest.has_removeall()` in parentheses before applying the conditional operator `?` ?

src/engine/source/api/src/test/handlers.cpp
```diff
@@ -0,0 +1,412 @@
...
+            for (auto& sessionName : sessionManager.getSessionsList())
+            {
+                const auto deleteSessionError = deleteSession(sessionName, router, catalog);
+                if (deleteSessionError.has_value())
+                {
+                    return genericError<ResponseType>(deleteSessionError.value().message);
+                }
+            }
+        }
+        else if (eRequest.has_name())
```
JorgeMarinoDev, line 377: **thought:** are we here sure that at this point that both conditions `eRequest.has_removeall() && eRequest.removeall()` are false? What if just `eRequest.has_removeall()` is false, does the else clause make sense? Please disregard if both conditions are semantically linked.

src/engine/source/api/src/test/sessionManager.cpp
```diff
@@ -0,0 +1,129 @@
...
+        sessionRemoved = true;
+    }
+    else
+    {
+        // Remove a specific session by sessionName
+        auto sessionIt = m_activeSessions.find(sessionName);
+        if (sessionIt != m_activeSessions.end())
+        {
+            const auto policyName = sessionIt->second.getPolicyName();
+            const auto routeName = sessionIt->second.getRouteName();
```
JorgeMarinoDev, line 106: **thought:** We could use & references here to avoid copy. Maybe already optimized by code generation flags. Not sure here.

src/engine/source/api/include/api/test/sessionManager.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+    std::optional<base::Error> createSession(const std::string& sessionName,
+                                             const std::string& routeName,
+                                             const std::string& policyName,
+                                             uint32_t lifespan = 0);
+
+    std::vector<std::string> getSessionsList(void);
+    std::optional<Session> getSession(const std::string& sessionName);
+
+    bool deleteSessions(const bool removeAll, const std::string sessionName = "");
```
JorgeMarinoDev, line 84: **thought:** this might be made private. Since it's called by `deleteAllSessions` and `deleteSession`. Might be redundant?

JorgeMarinoDev (approved):
LGTM. No blocking or critical issues found. Just some thoughts regarding exception handling and possibly unnecessary collection traversal may be crucial, and other general comments regarding code readability. Great work!

src/engine/source/api/include/api/test/sessionManager.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+#include <fmt/format.h>
+
+#include <error.hpp>
+
+namespace api::sessionManager
+{
+
+constexpr auto FILTER_CONTENT_FORMAT =
+    R"({{"name": "{}", "check":[{{"~TestSessionName":"{}"}}]}})"; ///< Filter content format, where '{}' is the session
```
NahuFigueroa97, line 20: improve - desing: isn't valid json. Use https://jsonformatter.curiousconcept.com/# for verification

nmkoremblum: It is formatted like that so it is processed by `fmt::format` afterward

src/engine/source/api/src/test/handlers.cpp
nmkoremblum: If `eRequest.has_removeall()` is false, then `eRequest.removeall()` does not exist

src/engine/source/api/src/test/handlers.cpp
nmkoremblum: Priorities are not assigned in a continuous way, so it might be tricky actually to find the minimum available priority in that way.

src/engine/source/api/include/api/test/sessionManager.hpp
```diff
@@ -0,0 +1,86 @@
...
+{
+private:
+    std::unordered_map<std::string, Session> m_activeSessions;
+    std::unordered_map<std::string, std::string> m_policyMap;
+    std::unordered_map<std::string, std::string> m_routeMap;
+
+    std::shared_mutex m_sessionMutex;
+
+public:
+    SessionManager() = default;
```
NahuFigueroa97, line 22: improve-desing: These declarations are already made in the interface, please remove them

src/engine/source/api/include/api/test/sessionManager.hpp
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * @brief Create a new session.
+     *
+     * @param sessionName Name of the session.
+     * @param routeName Name of the route.
+     * @param policyName Name of the policy.
+     * @param lifespan Lifespan of the session in seconds. 0 means no expiration.
+     * @return std::optional<base::Error>
+     */
+    std::optional<base::Error> createSession(const std::string& sessionName,
```
NahuFigueroa97, line 42: improve-desing: inherited methods end with modify "override"

src/engine/source/api/include/api/test/sessionManager.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include "iSessionManager.hpp"
+
+#include <shared_mutex>
+#include <unordered_map>
+
+namespace api::sessionManager
+{
+
+class SessionManager : public ISessionManager
```
NahuFigueroa97, line 12: improve-desing: the "final" operator can be placed to ensure that no other class will inherit from "SessionManager"

src/engine/source/api/include/api/test/sessionManager.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * @brief Get a Session Manager instance.
+     *
+     * @return SessionManager&
+     */
+    static SessionManager& getInstance(void);
+
+    /**
+     * @brief Create a new session.
```
NahuFigueroa97, line 34: suggestion: In Doxygen, you can use the @copydoc tag to indicate that the documentation for a class or function is copied from somewhere else, such as the base class or interface. This avoids redundant documentation in the inheriting class.
example:
`/**
 * @class DerivedClass
 * @brief Clase derivada
 *
 * @copydoc BaseClass
 */`

NahuFigueroa97 (approved):
LGTM!

Issue #17370, opened 2023-05-31
Engine - Update Vagrantfile and default policy


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0 | Engine | Manager | Packages/Sources | OS version |

We need to update the default policy of the Wazuh engine so that, as soon as our vagrant environment is up, we have the following integrations available:
- System
- Windows
- Suricata
- Apache

To do this, we must not only update the policy, but also the vagrant to incorporate these integrations.

PR #17376, opened 2023-05-31
Update Vagrantfile and default policy

JcabreraC:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/17370|

## Description
We need to update the default policy of the Wazuh engine so that, as soon as our vagrant environment is up, we have the following integrations available:
- System
- Windows
- Suricata
- Apache

To do this, we must not only update the policy, but also the vagrant to incorporate these integrations.


juliancnn (approved):
LGTM!

JavierBejMen (approved):
LGTM

Issue #16130, opened 2023-02-09
 Race condition in wazuh-analysisd


thejbte:
|Wazuh version|Component|Install type|Regresion|
|---|---|---|---|
| 4.5 | wazuh-analysisd | Manager |#1874 #14099|

Thread Sanitizer detected the following race condition when investigating https://github.com/wazuh/wazuh/issues/14099:

## Rationale
When a rule with ignore options is activated, the following race conditions are generated
```yml
<rule id="100222" level="5" frequency="3" timeframe="120" ignore="10">
    <if_matched_sid>5716</if_matched_sid>
    <match>^Failed|^error: PAM: Authentication</match>
    <description>sshd: authentication failed.</description>
    <mitre>
      <id>T1110</id>
    </mitre>
    <group>authentication_failed,gdpr_IV_35.7.d,gdpr_IV_32.2,gpg13_7.1,hipaa_164.312.b,nist_800_53_AU.14,nist_800_53_AC.7,pci_dss_10.2.4,pci_dss_10.2.5,tsc_CC6.1,tsc_CC6.8,tsc_CC7.2,tsc_CC7.3,</group>
</rule>
 ```

Full log for launch the rule.
`Dec 10 01:02:02 host sshd[1234]: Failed none for root from 192.168.56.66 port 1066 ssh2`

### Race conditions
<details><summary>triggered rule many times</summary>

```vim
when the rule is fired multiple times in a row
========================
Note: child_rule->prev_rule sigid:[x] ifsid:[(x)] ifmatchsid[x] is is a message placed while debugging
this case is difficult to reproduce


==================
WARNING: ThreadSanitizer: data race (pid=194448)
  Read of size 4 at 0x55d1bd8d5684 by thread T109 (mutexes: write M518571266417710816):
    #0 OS_CheckIfRuleMatch analysisd/rules.c:3133 (wazuh-analysisd+0x4f3eb)
    #1 OS_CheckIfRuleMatch analysisd/rules.c:3111 (wazuh-analysisd+0x4f208)
    #2 w_process_event_thread analysisd/analysisd.c:2021 (wazuh-analysisd+0x61ba7)

  Previous write of size 4 at 0x55d1bd8d5684 by thread T11 (mutexes: write M517163891534161760):
    #0 DumpLogstats analysisd/analysisd.c:1138 (wazuh-analysisd+0x5dcca)
[...]

PR #17377, opened 2023-05-31
Fix hourly_events variable race condition

ncvicchi:
|Related issue|
|---|
|[#16130](https://github.com/wazuh/wazuh/pull/16130)|

## Description
This PR avoids a race condition that might occurr when reading hourly_events in analysisd module.
No protection was present for a situation where several threads might access the variable,
when it was used as a parameter to a printf.

Although the referred issue contains several race conditions detected, some have been
corrected before this PR, none was reproduced during testings and this is the only
one that, by inspection, was detected as probable to ocurr.

Several other race conditions were on heap allocated variables, and the original source code
was not identify (it was incorrectly marked a 3.7.1). Because of this, there is no correlation between
the logs in the issue and the source codes that were inspected.

### DIFF
--- a/src/analysisd/analysisd.c
+++ b/src/analysisd/analysisd.c
@@ -1151,10 +1151,14 @@ static void DumpLogstats()
     } while ((rulenode_pt = rulenode_pt->next) != NULL);
 
     /* Print total for the hour */
+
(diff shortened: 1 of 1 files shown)


jotacarma90 (approved):

vikman90:
We're discarding this PR.

Issue #16614, opened 2023-04-04
Engine - KVDB module improvements


cborla:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0 | Engine | Manager| - | - |

## Description

The primary objective of KVDB is to function as a high-throughput interface for storing key-value pairs permanently.

While the existing implementation satisfies the requirements of the original design, there is room for improvement and the need to address technical debt.

Proposed improvements to the current system include the following:

Improve the DB Handlers to:

1. **Improve the DB Handlers** to:
   1. Count instances.
   2. Identify the Owners.
   3. Automatically unregister the unused handler to facilitate DB closure.
2. **Add resource tuning** at the Manager level, including memory and thread caps.  [Depends on #16702 results]
3. **Allow for raw-level DB operations** such as merging, appending, and replacing entire or partial contents of the DB 
   with raw representations of key-value pairs. Consider implementing patch application.

**Nice to have**:
  1. **Add optional triggers** to subscribe to specific keys and dispatch async handlers when values change. This would allow for triggering specific operations in one module after another module changes relevant information.
  2. ZSH breaks KVDB API when writing a value containing a null character. Might be related to CLI library.

Implementation considerations include:

1. Prioritize the use of `string_view` where possible.
2. Explore the usage of a common transport data type such as an array of bytes.
3. To improve logging in basic or primitive operations, minimize writing to the Wazuh log and return messages to the user for greater control. 
   Include a severity switch to adjust the level of detail in the log output as needed.
   
## Tasks
 - [x] #16702
 - [x] #16704
 - [x] #17056
 - [x] #17128

## DoD
 - [x] Complete all tasks. 
 - [x] Unit test.
 - [x] Update the design document.
 - [x] Update the user manual.
 - [x] Memory Leak analysis.

soynof:
## Decoder to test helper functions

We have created a decoder to be able to test all the helper functions that use the KVDB module.

In order to perform the test, the following must be done:

1. Create an integration and upload the decoder to it:

```
---
name: decoder/test2/0

metadata:
  module: test2
  title: Test logs decoder
  description: Decoder for test
  author:
    name: Wazuh, Inc.
    url: https://wazuh.com
    date: 2023/01/30

sources:
  - decoder/integrations/0

parse:
  logpar:
    # KVDB 30 May 10:04:45 clients connected (0 slaves), 618932 bytes in use, 0 shared objects
[...]

PR #17446, opened 2023-06-06
Engine KVDB Improvements

cborla:
|---|---|---|
| #16614|   [User manual](https://docs.google.com/document/d/1N0p7pd7XszQwFU9HBjpB7txIw2DCcAsV-BKfEZSLixw/edit) - [Design document](https://docs.google.com/document/d/125d8tpj192MNdggLcNjk_R0tQWEuhjBEmtjONHbUkHo/edit)|  |

## Description

The primary objective of KVDB is to function as a high-throughput interface for storing key-value pairs permanently.

While the existing implementation satisfies the requirements of the original design, there is room for improvement and the need to address technical debt.

## Details of the new KVDB Module

* RocksDB remains as underlying technology.
* Versatile architecture to balance fine tuning. 
  * A 'KVDB' is now a logical key-value space. Implemented as a Column Family. 
  * A KVDB Manager is now a single DataBase (in a filesystem path) holding many "KVDBs"
  * Many instances of Managers can have different optimization or behavior settings.
* KVDB Handlers, used across the Engine belong to a logical Scope, similar to a "user" of a DB Engine.
   * It means, we can identify what entity or submodule of the Engine is using a KVDB and how many handlers are held.
   * It gives the freedom to set the desired design granularity to direct optimizations towards the different requirements of every piece of the Engine that uses a KVDB.
   * No longer depending on memory reference counters of spread across handlers but belonging to logical spaces.
   
## Future Improvements:

New architecture was thought and implemented having in mind these features (yet to be implemented). 

* **Pagination** Retrieving data in smaller chunks or pages. Instead of returning raw results we can return results as a payload with a custom semantic and json metadata that contains pagination data. Also adding extra parameters to the Get function. Challenge here is on the backend, in terms of caching previous queries or storing temporal iterators' descriptors. Providing state-based functionality through a state-less API would at least demand some sort of cache, or a temporal ReadOnly copy of the DB being queried. Lots to discuss here.
* **Prefixes** Enhancing search performance with RocksDB 'prefixes'. The main advantage of prefix searches in RocksDB is the capability and high performance to iterate through a set of keys starting with a 'prefix'. This would only be noticeable when reading previously stored data under very specific circumstances . Nice to have before taking this feature into consideration: a specific use case, read only, and a hierarchically arranged set of keys with a valid semantic meaning. On the other hand, and under a theoretical POV, adding a prefix to a key can be used to implement all of the features of a relational DB in a non-relational DB like RocksDB. Manual and automatic prefixes can be used. tldr; Prefix search is worth it in read only scenarios where keys keep a hierarchical relationship in between, or where more complex filtering can be done at insert time to generate additional indexes to emulate complex queries later on. Lots to discuss here also. Keep in mind that fully functional relational database engines were implemented on top of RocksDB. Prefix search is the bridge that allows this.
* **Usage Information** The new KVDB Manager has 2 internal methods to retrieve the ownership status of the handlers. One to obtain who is using a KVDB, and another one, as a reverse lookup method to obtain what KVDBs are used by the different modules. i.e. useful to know what DBs are locked by the Builder. Currently it's mainly used in Unit Testing, but planning to expose this information through the API and CLI. Potential use case, we want to delete a KVDB but it fails. After asking this information, we know who is using the KVDB. Currently, the Builder registers itself as the owner of the KVDBs used by the decoders no matter what policy they belong to. The new KVDB architecture allows to define as many scopes as needed, even though the current architecture of the Builder locks the DBs when HF handlers are built, not run. Execution state is not available.
* **Metrics** RocksDB provides a comprehensive set of tools to obtain metrics about Column Families, DB and overall performance. The new architecture already provides structures and collections to later bind this information to. It's also very straightforward to expose this through the API and CLI.
* **TTL Support**: RocksDB allows certain DBs to keep KVs for a certain amount of time and to be automatically deleted afterwards. The current architecture could be easily adapted to support this, since all the logical key spaces are independent of the unique internal DB object of RocksDB. Could be implemented through a setup option switching the underlying DB Object as needed.
* **API raw import/export**. This is the capability to insert or retrieve data to/from DBs in raw or json format from the command line or through the API. Could be used to import or export data with external scripts. It could also be used in upgrade stages to perform backups or hosts migrations. This feature heavily depends on the solution after **pagination**. This new KVDB Manager interface is completely stateless. We would need to add new methods into the KVDBManager using some kind of iterator hashes, and the notion of "transaction" or "interactive mode", both for reading or writing data. This could certainly be the most "complex" feature to add, since we need to define an arbitrary procedure that will finally impact to the end user who wants to import or export raw data. There will certainly be limitations in the API MTU and tty for CLI access. On the CLI's end, it's possible to add an interactive mode that internally deals with the API responses, iterators or ranges.

## Technical Debt besides future improvements

* Propagation and handling of errors or exceptions in a hierarchical manner.
  * Filesystem exceptions, OS cross-platform checks.
  * Value content transport. Inherent exceptions related to MTUs or CLI tty limitations.
* Missing intensive and concurrent I/O operations after a stable scenario in terms of Scopes and Handlers.
* Checking Write Ahead Log features.

## Unit Testing

Even though tested coverage is yet known to be improved, this initial integration aims to at least cope with the existing testing criteria. This means that after some adaptations, the new code passes the existing UT. Results Here,

image
image
image
image
image


src/engine/source/kvdb/include/kvdb/kvdbManager.hpp
```diff
@@ -1,226 +1,231 @@
...
+#include <kvdb/iKVDBHandlerManager.hpp>
+#include <kvdb/iKVDBManager.hpp>
+#include <kvdb/kvdbHandlerCollection.hpp>
+#include <kvdb/kvdbScope.hpp>
+#include <kvdb/kvdbSpace.hpp>
+
+#include <utils/baseMacros.hpp>
+
+#include <rocksdb/db.h>
+#include <rocksdb/options.h>
```
juliancnn, line 18:
```suggestion
#include <rocksdb/db.h>
#include <rocksdb/options.h>

#include <utils/baseMacros.hpp>

#include <kvdb/iKVDBHandlerManager.hpp>
#include <kvdb/iKVDBManager.hpp>
#include <kvdb/kvdbHandlerCollection.hpp>
#include <kvdb/kvdbScope.hpp>
#include <kvdb/kvdbSpace.hpp>
```

src/engine/source/kvdb/include/kvdb/kvdbSpace.hpp
```diff
@@ -0,0 +1,106 @@
+#ifndef _KVDB_SPACE_H
+#define _KVDB_SPACE_H
+
+#include <kvdb/iKVDBHandler.hpp>
+#include <kvdb/kvdbManagedHandler.hpp>
+
+#include "rocksdb/db.h"
```
juliancnn, line 7: Check the order.

src/engine/source/kvdb/src/kvdbHandlerCollection.cpp
```diff
@@ -0,0 +1,94 @@
+#include <fmt/format.h>
+#include <kvdb/kvdbHandlerCollection.hpp>
+#include <logging/logging.hpp>
```
juliancnn, line 3:
```suggestion
#include <fmt/format.h>

#include <logging/logging.hpp>

#include <kvdb/kvdbHandlerCollection.hpp>
```

src/engine/source/kvdb/include/kvdb/kvdbHandlerCollection.hpp
```diff
@@ -0,0 +1,155 @@
...
+     * @brief Pointer to the Manager that deals with handlers.
+     *
+     */
+    IKVDBHandlerManager* m_handleManager {nullptr};
+
+    /**
+     * @brief Mutex to protect the internal map.
+     *
+     */
+    std::mutex m_mutex;
```
juliancnn, line 150: Is it necessary to do mutual exclusion? perhaps, if we mostly read and not write the map, it is convenient to use std::shared_mutex. 
If the mutex is only needed at compile time of the execution pipeline (environment construction) and not at run time of the expression chain, it would not be necessary. What do you think?

juliancnn (changes requested):

src/engine/source/kvdb/include/kvdb/refCounter.hpp
```diff
@@ -0,0 +1,92 @@
...
+     */
+    ~RefCounter() { m_refMap.clear(); }
+
+    /**
+     * @brief Add n reference(s) to the counter.
+     *
+     * @param name Reference name to increase
+     * @param times How many times to increase the reference. Default is 1.
+     */
+    void addRef(const std::string& name, const unsigned int& times = 1);
```
JavierBejMen, line 44: Always pass trivial types by value.

Also using uint to increase an int.

src/engine/source/kvdb/src/kvdbHandlerCollection.cpp
```diff
@@ -0,0 +1,94 @@
...
+        instance->addScope(scopeName);
+    }
+    else
+    {
+        auto spInstance = std::make_shared<KVDBHandlerInstance>();
+        spInstance->addScope(scopeName);
+        m_mapInstances.insert(std::make_pair(dbName, std::move(spInstance)));
+    }
+
+    return std::make_unique<KVDBSpace>(m_handleManager, db, cfHandle, dbName, scopeName);
```
JavierBejMen, line 27: Use make_shared

src/engine/source/kvdb/include/kvdb/kvdbExcept.hpp
```diff
@@ -0,0 +1,80 @@
...
+#include <exception>
+#include <string>
+
+namespace kvdbManager
+{
+
+/**
+ * @brief Exception thrown when critical KVDB Functions Fails
+ */
+class KVDBException : public std::exception
```
JavierBejMen, line 13: Is this exception thrown? Also we try to use only standard exceptions

src/engine/source/builder/builders/opBuilderKVDB.cpp
```diff
@@ -154,23 +160,24 @@ base::Expression existanceCheck(const std::string& targetField,
...
         name,
-        [=, kvdb = std::get<kvdb_manager::KVDBHandle>(result), targetField = std::move(targetField)](base::Event event)
+        [=, 
+        targetField = std::move(targetField),
+        kvdbHandler = std::move(std::get<std::shared_ptr<kvdbManager::IKVDBHandler>>(resultHandler))
+        ]
+        (base::Event event)
         {
             bool found = false;
             try // TODO We are only using try for JSON::get. Is correct to
```
JavierBejMen, line 183: JSON::get only throws if the JPath is invalid, which should be checked on buildtime

src/engine/source/kvdb/include/kvdb/refCounter.hpp
JavierBejMen: I'm sorry, but there seems to be a misunderstanding. What I meant about the unsigned int is that the map stores an int and we use an unsigned type to increase said int. Therefore performing a narrowing conversion.

This kind of error should be warned by clang-tidy, which is integrated into vs code:
image

JorgeMarinoDev: Fair enough observation Javier. Fixed. Thank you.

src/engine/source/kvdb/include/kvdb/kvdbExcept.hpp
JorgeMarinoDev: Custom exceptions were already removed. Nice catch. No need to use them here at the moment and will rely on std ones. Thank you.

LucioDonda (changes requested):

LucioDonda (approved):
LGTM!

juliancnn (approved):

src/engine/source/kvdb/src/refCounter.cpp
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+}
+
+unsigned int RefCounter::count(const std::string& name) const
+{
+    if (m_refMap.count(name) > 0)
+    {
+        return m_refMap.at(name);
+    }
```
juliancnn, line 33: Should not do the search twice, nor should a previous check be done so as not to launch an execption:

```suggestion
    auto it = m_refMap.find(name);
    if (it != m_refMap.end())
    {
        return it->second;
    }
```

src/engine/source/kvdb/src/refCounter.cpp
```diff
@@ -0,0 +1,58 @@
...
+void RefCounter::removeRef(const std::string& name)
+{
+    if (m_refMap.count(name) > 0)
+    {
+        m_refMap[name]--;
+        if (0 == m_refMap[name])
+        {
+            m_refMap.erase(name);
+        }
+    }
```
juliancnn, line 25: Avoid mutiple search:
```suggestion
    auto it = m_refMap.find(name);
    if (it != m_refMap.end())
    {
        it->second--;
        if (it->second == 0)
        {
            m_refMap.erase(it);
        }
    }
```

src/engine/source/kvdb/src/refCounter.cpp
```diff
@@ -0,0 +1,58 @@
+#include <kvdb/refCounter.hpp>
+
+namespace kvdbManager
+{
+
+RefCounter::RefCounter(const RefCounter& other)
```
juliancnn, line 6: Why is the default copy constructor replaced?

juliancnn (changes requested):
1) Recheck the file inclusion order.

2) Check and apply The rule of three/five/zero

3) Avoid the use of raw pointers whenever possible, unless there is a specific case where they are necessary. Raw pointers can lead to memory management problems, such as memory leaks or invalid accesses, and make it difficult to track and understand the code.

Instead of using raw pointers, it is recommended to use smart pointers from the C++ standard library. Smart pointers, such as std::unique_ptr or std::shared_ptr, provide automatic memory management and take care of properly freeing memory when they are no longer needed.

If the use of raw pointers is unavoidable, the creation of an object that manages the raw pointer using smart pointers can be considered. This allows for safer and more robust memory management, avoiding "volatile" raw pointers that can cause problems in the code.

4) No code suggestions were compiled or tested. They must be analyzed, checked and tested.

src/engine/source/kvdb/include/kvdb/iKVDBHandlerManager.hpp
```diff
@@ -0,0 +1,51 @@
+#ifndef _I_KVDB_HANDLER_MANAGER_H
+#define _I_KVDB_HANDLER_MANAGER_H
+
+#include <kvdb/iKVDBHandler.hpp>
```
cborla, line 4: check the include order

nmkoremblum (changes requested):

src/engine/source/api/src/kvdb/handlers.cpp
```diff
@@ -71,19 +72,42 @@ api::Handler managerPost(std::shared_ptr<kvdb_manager::KVDBManager> kvdbManager)
...
+            return ::api::adapter::genericError<ResponseType>(resultCreate.value().message);
+        }
+
+        if (eRequest.has_path())
+        {
+            auto resultLoad = kvdbManager->loadDBFromFile(eRequest.name(), eRequest.path());
+
+            if (resultLoad)
+            {
+                std::string message = fmt::format("The DB was created but loading data returned: {}", resultLoad.value().message);
```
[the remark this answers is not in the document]
soynof: I think that the proposed message does not make it clear if the DB was created or not, what do you think?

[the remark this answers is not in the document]
JorgeMarinoDev: Please disregard this. It's a temporal message because it's part of a pending feature. This message just states what happened. But to keep consistency we used that message.

src/engine/source/builder/builders/opBuilderKVDB.cpp
```diff
@@ -349,88 +364,97 @@ base::Expression KVDBSet(const std::string& targetField,
...
+         kvdbHandler =
+             std::move(std::get<std::shared_ptr<kvdbManager::IKVDBHandler>>(resultHandler))](base::Event event)
+        {
+            event->setBool(false, targetField);
+
+            std::string resolvedKey;
+            if (Parameter::Type::REFERENCE == key.m_type)
+            {
+                const auto value = event->getString(key.m_value);
+                if (value)
```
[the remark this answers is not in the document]
JorgeMarinoDev: It is the same, since std::optional provides a [bool operator](https://en.cppreference.com/w/cpp/utility/optional/operator_bool) to achieve this. But to keep clarity and consistency with the rest of the Engine's code, we will change this to has_value(). 
Thank you!

src/engine/source/kvdb/src/refCounter.cpp
```diff
@@ -0,0 +1,56 @@
...
+    {
+        refNames.push_back(ref.first);
+    }
+    return refNames;
+}
+
+std::map<std::string, uint32_t> RefCounter::getRefMap() const
+{
+    return m_refMap;
+}
```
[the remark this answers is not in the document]
JorgeMarinoDev: It's a good observation. This helper class RefCounter is used internally by **KVDBHandlerInstance**, and every method called is already mutexed by the caller on it's own scope. 
If we need to export this helper class to a more general context, such as a base and commonly used utility , it would make sense to mutex the collections as you mention. But as for now, it's not necessary, again, unless we extract and expose this helper to a more generic context used anywhere else.

src/engine/source/kvdb/src/kvdbManager.cpp
```diff
@@ -1,412 +1,335 @@
...
+        assert(opStatus.ok());
     }
 
-    return nullptr;
+    m_mapCFHandles.clear();
+
+    opStatus = m_pRocksDB->Close();
+    assert(opStatus.ok());
+
+    delete m_pRocksDB;
```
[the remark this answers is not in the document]
soynof: As far as the official documentation refers to, that is the intended way to close a DB.

src/engine/source/api/src/kvdb/handlers.cpp
```diff
@@ -71,19 +72,42 @@ api::Handler managerPost(std::shared_ptr<kvdb_manager::KVDBManager> kvdbManager)
...
+            return ::api::adapter::genericError<ResponseType>(resultCreate.value().message);
+        }
+
+        if (eRequest.has_path())
+        {
+            const auto resultLoad = kvdbManager->loadDBFromFile(eRequest.name(), eRequest.path());
+
+            if (resultLoad)
+            {
+                std::string message = fmt::format("The DB was created but loading data returned: {}", resultLoad.value().message);
```
nmkoremblum, line 100: can we make it const? the same recommendation applies to other cases

src/engine/source/api/src/kvdb/handlers.cpp
```diff
@@ -134,8 +172,28 @@ api::Handler managerDump(std::shared_ptr<kvdb_manager::KVDBManager> kvdbManager)
...
+
+        if (std::holds_alternative<base::Error>(resultHandler))
+        {
+            return ::api::adapter::genericError<ResponseType>(std::get<base::Error>(resultHandler).message);
+        }
 
-        auto dumpRes = kvdbManager->rDumpDB(eRequest.name());
+        auto handler = std::get<std::shared_ptr<kvdbManager::IKVDBHandler>>(resultHandler);
+
+        auto dumpRes = handler->dump();
```
nmkoremblum, line 196: can we make it const? the same recommendation applies to other cases

nmkoremblum (changes requested):

src/engine/source/kvdb/src/kvdbManager.cpp
```diff
@@ -1,412 +1,346 @@
...
+    std::vector<rocksdb::ColumnFamilyHandle*> cfHandles;
+
+    bool hasDefaultCF = false;
+    auto listStatus = rocksdb::DB::ListColumnFamilies(rocksdb::DBOptions(), dbNameFullPath, &columnNames);
+    if (listStatus.ok())
     {
-        m_dbs.erase(name);
+        for (const auto& cfName : columnNames)
+        {
+            if (cfName == rocksdb::kDefaultColumnFamilyName)
```
nmkoremblum, line 78:
```suggestion
            if (rocksdb::kDefaultColumnFamilyName == cfName)
```

src/engine/source/kvdb/src/kvdbManager.cpp
```diff
@@ -1,412 +1,346 @@
...
-}
-
-std::optional<base::Error> KVDBManager::writeRaw(const std::string& name,
-                                                 const std::string& key,
-                                                 const std::string value)
-{
-    auto handle = getHandler(name);
-    if (std::holds_alternative<base::Error>(handle))
+    // TODO: to improve
+    if (std::empty(path))
```
nmkoremblum, line 216:
```suggestion
    if (path.empty())
```
Same recommendation applies to other occurrencies

src/engine/source/kvdb/src/refCounter.cpp
```diff
@@ -0,0 +1,56 @@
...
+    m_refMap[name] += times;
+}
+
+void RefCounter::removeRef(const std::string& name)
+{
+    auto it = m_refMap.find(name);
+    if (it != m_refMap.end())
+    {
+        it->second--;
+        if (it->second == 0)
```
nmkoremblum, line 17:
```suggestion
        if (0 == it->second)
```

nmkoremblum: Same recommendation applies to other cases

nmkoremblum (changes requested):

src/engine/source/builder/builders/opBuilderKVDB.cpp
```diff
@@ -144,8 +150,9 @@ base::Expression existanceCheck(const std::string& targetField,
                                 const std::string& rawName,
                                 const std::vector<std::string>& rawParameters,
                                 std::shared_ptr<defs::IDefinitions> definitions,
-                                bool checkExist,
-                                std::shared_ptr<kvdb_manager::KVDBManager> kvdbManager)
+                                bool shouldMatch,
+                                std::shared_ptr<IKVDBManager> kvdbManager,
+                                const std::string& kvdbScopeName)
```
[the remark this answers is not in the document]
JorgeMarinoDev: We definitely have to sort these parameters out. But the scope of change is IMHO, broader than the expected as a PR review change. Regarding shouldMatch, sounds good.

src/engine/source/kvdb/include/kvdb/refCounter.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <vector>
+
+namespace kvdbManager
+{
+
+/**
+ * @brief Basic Helper wrapping std::map<std::string, uint32_t> to count references.
+ *
+ */
+class RefCounter
```
LucioDonda, line 15: does it make sense to add UT to this class?

LucioDonda (changes requested):

src/engine/source/builder/builders/opBuilderKVDB.cpp
```diff
@@ -116,36 +121,37 @@ base::Expression KVDBGet(const std::string& targetField,
...
+                                        const std::string& rawName,
+                                        const std::vector<std::string>& rawParameters,
+                                        std::shared_ptr<defs::IDefinitions> definitions)
     {
-        return KVDBGet(targetField, rawName, rawParameters, definitions, true, kvdbManager);
+        return KVDBGet(kvdbManager, kvdbScopeName, targetField, rawName, rawParameters, definitions, true);
     };
 }
 
 // TODO: documentation and tests of this method are missing
```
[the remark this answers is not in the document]
soynof: Yes, still pending.

src/engine/source/builder/builders/opBuilderKVDB.cpp
```diff
@@ -154,88 +160,88 @@ base::Expression existanceCheck(const std::string& targetField,
...
+                {
+                    return base::result::makeFailure(event,
+                                                     failureTrace + ": " + std::get<base::Error>(result).message);
+                }
+
+                found = std::get<bool>(result);
             }
-            else
+
+            if ((shouldMatch && found) || (!shouldMatch && !found))
```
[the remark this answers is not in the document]
cborla: This method is used for 2 helpers, the one that matches and the one that does not match, the first condition is for case 1, the second for case 2, in case it is not found and the helper that was used is not match, it must return ok.

src/engine/source/kvdb/include/kvdb/refCounter.hpp
soynof: This class is covered with the UTs that were put together in KVDBTest.

src/engine/source/builder/builders/opBuilderKVDB.cpp
```diff
@@ -154,88 +160,88 @@ base::Expression existanceCheck(const std::string& targetField,
...
             {
-                return base::result::makeFailure(event, failureTrace);
+                return base::result::makeSuccess(event, successTrace);
             }
+
+            return base::result::makeFailure(event, failureTrace);
         });
 }
 
 // TODO: tests for this method are missing
```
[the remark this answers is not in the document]
soynof: Yes, still pending.

src/engine/source/api/src/kvdb/handlers.cpp
```diff
@@ -192,16 +252,43 @@ api::Handler dbGet(std::shared_ptr<kvdb_manager::KVDBManager> kvdbManager)
...
+
+        const auto resultHandler = kvdbManager->getKVDBHandler(eRequest.name(), kvdbScopeName);
+
+        if (std::holds_alternative<base::Error>(resultHandler))
+        {
+            return ::api::adapter::genericError<ResponseType>(std::get<base::Error>(resultHandler).message);
+        }
+
+        auto handler = std::get<std::shared_ptr<kvdbManager::IKVDBHandler>>(resultHandler);
+        auto result = handler->get(eRequest.key());
```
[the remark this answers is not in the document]
nmkoremblum: Please, check if the "const" approach is valid

src/engine/source/kvdb/src/kvdbManager.cpp
```diff
@@ -1,412 +1,366 @@
...
-
-            return base::Error {
-                fmt::format("Database '{}' is already in use '{}' times", name, handler.use_count() - MAX_USE_COUNT)};
-        }
-    }
-    // Mark for deletion
-    handler->cleanupOnClose();
-
-    return std::nullopt;
+            auto opStatus = pRocksDB->DestroyColumnFamilyHandle(ptr);
```
nmkoremblum, line 357:
```suggestion
            const auto opStatus = pRocksDB->DestroyColumnFamilyHandle(ptr);
```

src/engine/source/kvdb/src/kvdbManager.cpp
```diff
@@ -1,412 +1,366 @@
...
-                                                                const std::string& key)
-{
-    auto handle = getHandler(name);
-    if (std::holds_alternative<base::Error>(handle))
+    entries = fileContentsJson.getObject().value();
+
+    for (const auto& [key, value] : entries)
     {
-        return std::get<base::Error>(handle);
+        auto status = m_pRocksDB->Put(rocksdb::WriteOptions(), cfHandle.get(), key, value.str());
```
nmkoremblum, line 259:
```suggestion
        const auto status = m_pRocksDB->Put(rocksdb::WriteOptions(), cfHandle.get(), key, value.str());
```

src/engine/source/kvdb/src/kvdbManager.cpp
```diff
@@ -1,412 +1,366 @@
...
-            return base::Error {fmt::format("An error occurred while parsing the JSON file '{}'", path.c_str())};
+            auto opStatus = m_pRocksDB->DropColumnFamily(cfHandle.get());
+            if (opStatus.ok())
+            {
+                m_mapCFHandles.erase(it);
+            }
+            else
+            {
+                return base::Error {
+                    fmt::format("Could not remove the DB {}: {}", name, opStatus.ToString())};
```
nmkoremblum, line 184:
```suggestion
                    fmt::format("Database '{}' could not be removed: {}", name, opStatus.ToString())};
```

src/engine/source/kvdb/src/kvdbManager.cpp
```diff
@@ -1,412 +1,366 @@
...
+            }
         }
-
-        if (!jKv.isObject())
+        catch (const std::runtime_error& e)
         {
-            return base::Error {
-                fmt::format("An error occurred while parsing the JSON file '{}': JSON is not an object", path.c_str())};
+            return base::Error {fmt::format("Could not remove the DB {}: {}", name, e.what())};
```
nmkoremblum, line 189:
```suggestion
            return base::Error {fmt::format("Database '{}' could not be removed: {}", name, e.what())};
```

src/engine/source/kvdb/src/kvdbManager.cpp
```diff
@@ -1,412 +1,366 @@
...
 
-        json::Json jKv;
         try
         {
-            jKv = json::Json {contents.c_str()};
-        }
-        catch (const std::exception& e)
-        {
-            return base::Error {fmt::format("An error occurred while parsing the JSON file '{}'", path.c_str())};
+            auto opStatus = m_pRocksDB->DropColumnFamily(cfHandle.get());
```
nmkoremblum, line 176:
```suggestion
            const auto opStatus = m_pRocksDB->DropColumnFamily(cfHandle.get());
```

src/engine/source/kvdb/src/kvdbManager.cpp
```diff
@@ -1,412 +1,366 @@
...
-std::optional<base::Error> KVDBManager::createFromJFile(const std::string& dbName,
-                                                        const std::filesystem::path& path)
+std::optional<base::Error> KVDBManager::deleteDB(const std::string& name)
 {
-    std::vector<std::tuple<std::string, json::Json>> entries {};
+    auto handlersInfo = getKVDBHandlersInfo();
 
-    // Open file and read content
-    if (!path.empty())
+    auto refCount = handlersInfo.count(name);
```
nmkoremblum, line 163:
```suggestion
    const auto refCount = handlersInfo.count(name);
```

src/engine/source/kvdb/src/kvdbManager.cpp
```diff
@@ -1,412 +1,366 @@
...
-            contents.resize(in.tellg());
-            in.seekg(0, std::ios::beg);
-            in.read(&contents[0], contents.size());
-            in.close();
-        }
-        else
-        {
-            return base::Error {fmt::format("An error occurred while opening the file '{}'", path.c_str())};
-        }
+        return base::Error {fmt::format("Could not remove the DB {}. Usage Reference Count: {}.", name, refCount)};
```
nmkoremblum, line 166:
```suggestion
        return base::Error {fmt::format(""Database '{}' could not be removed. Usage Reference Count: {}.", name, refCount)};
```

src/engine/source/kvdb/src/kvdbManager.cpp
```diff
@@ -1,412 +1,366 @@
...
+
+    std::vector<std::string> columnNames;
 
-    auto databasesInUse = m_spMetricsScope->getUpDownCounterInteger("DatabasesInUse");
+    std::vector<rocksdb::ColumnFamilyDescriptor> cfDescriptors;
+    std::vector<rocksdb::ColumnFamilyHandle*> cfHandles;
 
-    if (isLoaded)
+    bool hasDefaultCF = false;
+    auto listStatus = rocksdb::DB::ListColumnFamilies(rocksdb::DBOptions(), dbNameFullPath, &columnNames);
```
nmkoremblum, line 65:
```suggestion
    const auto listStatus = rocksdb::DB::ListColumnFamilies(rocksdb::DBOptions(), dbNameFullPath, &columnNames);
```

src/engine/source/kvdb/src/kvdbManager.cpp
```diff
@@ -1,412 +1,366 @@
...
+    m_rocksDBOptions.OptimizeLevelStyleCompaction();
+    m_rocksDBOptions.create_if_missing = true;
 }
 
-void KVDBManager::unloadDB(const std::string& name)
+void KVDBManager::initializeMainDB()
 {
-    std::unique_lock lkW(m_mtx);
-    const bool isLoaded = m_dbs.find(name) != m_dbs.end();
+    auto dbStoragePath = m_ManagerOptions.dbStoragePath.string();
```
nmkoremblum, line 53:
```suggestion
    cosnt auto dbStoragePath = m_ManagerOptions.dbStoragePath.string();
```

nmkoremblum (changes requested):

nmkoremblum (approved):
LGTM

NahuFigueroa97 (approved):
LGTM!

src/engine/source/kvdb/src/kvdbManager.cpp
```diff
@@ -1,412 +1,382 @@
...
 
-KVDBHandle KVDBManager::getDB(const std::string& name)
-{
-    std::shared_lock lkReadDBs(m_mtx);
+    if (!hasDefaultCF)
+    {
+        auto newDescriptor =
+            rocksdb::ColumnFamilyDescriptor(rocksdb::kDefaultColumnFamilyName, rocksdb::ColumnFamilyOptions());
+        cfDescriptors.push_back(newDescriptor);
+    }
```
LucioDonda, line 85: Can this case happen ? if so, then I guess that we could force it to create kvdbs without default cf, right? if not, can we change **kDefaultColumnFamilyName** to a not-so-common name ?

LucioDonda (approved):
LGTM!

JavierBejMen (approved):
LGTM

src/engine/source/kvdb/src/kvdbManager.cpp
soynof: It is not possible not to load the "default" CF, since it is mandatory for backward compatibility.

If the "default" CF is not specified RocksDB throws the exception "An error occurred while trying to open the database: Invalid argument: Default column family not specified".

For more information: https://github.com/EighteenZi/rocksdb_wiki/blob/master/Column-Families.md

soynof (approved):
LGTM!

JorgeMarinoDev (approved):
LGTM.

JavierBejMen (approved):
LGTM

juliancnn (approved):
LGTM!

PR #17447, opened 2023-06-07
Set database permissions for dbsync file databases

### LINKED ISSUE #17304: Unify agent's database file permissions
|Version|Components|
|---|---|
|4.5 (dev)|FIM & Syscollector|

@jnasselle has reported that the FIM database file permissions have changed:

|File|4.4|4.5|
|---|---|---|
|**queue/fim/db/fim.db**|`-rw-rw----`|`-rw-r--r--`|
|**queue/fim/db/fim.db-journal**|`-rw-rw----`|`-rw-r--r--`|

The behavior in version 4.4 was due to this code in [fim_db.c](https://github.com/wazuh/wazuh/blob/c35659999ea342ff9254b83ba5f3d3f2515b182b/src/syscheckd/db/fim_db.c#L242-L245).

However, FIM is now using the DBsync component together with Syscollector, whose database has the same permissions:

|File|≥ 4.2|
|---|---|
|**queue/syscollector/local.db**|`-rw-r--r--`|
|**queue/syscollector/local.db-journal**|`-rw-r--r--`|

## Proposal

It is interesting to maintain the permissions of the files between versions (unless justified). On the other hand, we also want all databases to have the same permissions.

Thus, I propose to unify the permissions of the databases that DBsync manages. In this case, IMO the most appropriate permissions would be `rw-r-----` (640).

Dwordcito:
|Related issue|
|---|
|Closes #17304 |

## Description
This PR aims to set the defined permissions for dbsync databases under the SQLite implementation.

Basically, open the database set file permissions and after this, reopen the file with the new permissions applied.

The component test covers this change.


src/shared_modules/dbsync/src/sqlite/sqlite_wrapper.cpp
pereyra-m, line 1: If I read the code correctly, for every connection to the database in non-Windows agents we:
- We open the DB with the flags `SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE`
- We make sure the DB isn't of type **memory**
- We set the proper permissions 
- We close the DB and open it again with the flag `SQLITE_OPEN_READWRITE`

But a DB managed by `dbsync` is only created once, and accessed many times.
So it might not be efficient to set the permissions and opening the DB twice for every connection.

I'd have expected something more similar to:
- Only if the DB doesn't exists, create it and set the proper permissions (if the DB isn't **memory** type)
- If the DB exists, we open it just once. We don't check if it's a **memory** DB

pereyra-m (commented):
Gj!
I haven't too much experience in this section of the code, so the suggestion might not be possible or relevant.

src/shared_modules/dbsync/src/sqlite/sqlite_wrapper.cpp
Dwordcito: @pereyra-m The database(File) is created in each initialization by design.

pereyra-m: Thank you @Dwordcito !
I didn't know the DB was created every time.
Now, if this is the case, why we need to open it once, close it and open it again?

MiguelazoDS (dismissed):

src/shared_modules/dbsync/src/sqlite/sqlite_wrapper.cpp
Dwordcito: it is to correctly validate that sqlite3 will be able to have the required access in case it has to open or close the associated database file internally.

pereyra-m: Ok, thank you.

pereyra-m (approved):

PR #17463, opened 2023-06-07
Prevent invalid free of stack memory at osquery integration module

### LINKED ISSUE #17440: Legacy bad free Scan-build report.
|---|---|---|---|---|
| 4.6 | Wazuh-modules | Manager/Agent | Packages/Sources | Any |

# Description 

It was found that some PRs with the master branch as the target are failing due to the Scan-build report. 

image

The error appears after some changes (https://github.com/wazuh/wazuh/commit/dd01ecae89f792eab3966ab5a0f0d0211d2c4846) in the `wm_osquery_monitor.c`

- Full scan-build report from RTR for manager. 

`python3 build.py --scanbuild server`

[scan-build.zip](https://github.com/wazuh/wazuh/files/11667362/scan-build.zip)

vikman90:
|Related issue|
|---|
|Closes #17440|

This PR aims to fix a bug that mishandles logging data from osquery.

There is a 4 KB buffer allocated in the thread's stack that receives each log from the osquery output. The log parser may call `wm_osquery_already_running()` to match a message like:

> osqueryd (1000) is already running

This function should return a heap-allocated string containing either the process PID, `"unknown"`, or a null pointer (if the message did not match the pattern). However, a recent change let the function return the source string itself. As the caller assumes that the returned string is independent and heap-allocated, it tries to free it after, causing an invalid free hazard.

## Proposed fix

Let `wm_osquery_already_running()` return `NULL` if the input string matches no pattern.

### DIFF
--- a/src/wazuh_modules/wm_osquery_monitor.c
+++ b/src/wazuh_modules/wm_osquery_monitor.c
@@ -425,6 +425,8 @@ char * wm_osquery_already_running(char * text) {
             // Find "Pidfile::Error::Busy"
         } else if (strstr(text, PATTERNS[2]) != NULL) {
             os_strdup("unknown", text);
+        } else {
+            text = NULL;
         }
     }
     return text;
--- a/src/wazuh_modules/wm_osquery_monitor.c
+++ b/src/wazuh_modules/wm_osquery_monitor.c
@@ -37,6 +37,13 @@
 #define mdebug1(msg, ...) _mtdebug1(WM_OSQUERYMONITOR_LOGTAG, __FILE__, __LINE__, __func__, msg, ##__VA_ARGS__)
 #define mdebug2(msg, ...) _mtdebug2(WM_OSQUERYMONITOR_LOGTAG, __FILE__, __LINE__, __func__, msg, ##__VA_ARGS__)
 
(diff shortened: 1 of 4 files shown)


src/wazuh_modules/wm_osquery_monitor.c
```diff
@@ -427,7 +427,7 @@ char * wm_osquery_already_running(char * text) {
             os_strdup("unknown", text);
         }
```
thejbte, line 428: This will work.
```suggestion
        } else {
            text = NULL;
        }
```

src/wazuh_modules/wm_osquery_monitor.c
```diff
@@ -427,7 +427,7 @@ char * wm_osquery_already_running(char * text) {
...
     }
-    return text;
+    return NULL;
```
thejbte, line 430:
```suggestion
    return text;
```

thejbte (changes requested):

thejbte (dismissed):

Dwordcito:
@vikman90 is this PR blocked by this another one https://github.com/wazuh/wazuh/pull/17636 ? we may be deadlocked.

Issue #5232, opened 2020-06-15
Syscollector is reading a bad value for the process RSS memory


chemamartinez:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 3.12.3 | Syscollector | Agent | All | All |

Syscollector embeds the library `libproc` to gather information about the running processes in Linux agents. One of the reporting data for each process is the resident memory of the process.

To get that value, the field `resident` of the `proc_t` struct is being read, which is the incorrect one according to its documentation:

```
resident:          number of resident set (non-swapped) pages (4k)
```

To get the `RSS` memory occupied by the process, the value that has to be read is `rss` whose definition is:

```
rss:           resident set size from /proc/#/stat (pages)
```

or `vm_rss` instead, which is the same size in KB.

MiguelazoDS:
# Analysis 

In the actual release (4.4.4) the information according to the documentation is displayed in Bytes. 

![image](https://github.com/wazuh/wazuh/assets/13010397/689c3a6d-1535-4970-a9e3-738b6c02e5ba)

**Notice that there's a typo in the description**

PR #17600, opened 2023-06-16
Syscollector is reading a bad value for the process RSS memory

tdrauncieal:
|Related issue|
|---|
| #5232 |

## Description

Fix the origin of  RSS memory data of processes reported by `syscollector`.

the scope stated in the related issue is all OS, however, this information is currently only available in `Linux` environments.

## Tests

image
two runs of `sysinfo_test_tool` focused on similar processes, after & before of change were made,  showing the data difference

  - [x] Linux


pereyra-m (approved):
LGTM!

But please, create an issue for the UI, because this field isn't reflected in the agent's inventory. We are seeing `size` instead, that seems less useful

```
	size,		// statm           total # of pages of memory
```

matias-braida (approved):
LGTM

MiguelazoDS (approved):

Issue #17402, opened 2023-06-05
Invalid OS Name for Arch linux in the agent list


Dwordcito:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.4.0 | Wazuh component | Agent | Packages/Sources | Arch linux |

### Description

The reporting mechanism concatenates the `VERSION_ID` field, which contains the value "TEMPLATE_VERSION_ID", to the OS name. This results in an incorrect representation of the OS name for Arch Linux.

![image](https://github.com/wazuh/wazuh/assets/14300208/4be7ad7f-c827-4269-94c5-dd00655da422)
![image](https://github.com/wazuh/wazuh/assets/14300208/1a841651-ccac-4a0d-9f2d-d1d60140a6cf)

This is also wrong in openSUSE tumbleweed which is also a rolling release.
We should get the VERSION_ID from the OS report that are rolling releases.
![image](https://github.com/wazuh/wazuh/assets/14300208/3c393dae-0596-44d9-a1d1-8e4c689f6083)

### Proposed Solution

The proposed solution is to modify the reporting mechanism to correctly retrieve and report the OS name for Arch Linux. This can be achieved by updating the code to exclude the `VERSION_ID` field if it contains the value "TEMPLATE_VERSION_ID".

### Tasks

- Identify the code section responsible for retrieving and reporting the OS name.
- Update the code to exclude the `VERSION_ID` field if it contains the value "TEMPLATE_VERSION_ID".
- Test the changes to ensure that the OS name is correctly reported for Arch Linux.

### Expected Outcome

After implementing the proposed solution, the OS name reporting for Arch Linux should no longer include the `VERSION_ID` field with the value "TEMPLATE_VERSION_ID". The reported OS name should accurately represent the Arch Linux distribution.

PR #17631, opened 2023-06-22
Invalid os_version Arch Linux and openSUSE Tumbleweed

MiguelazoDS:
|Related issue|
|---|
|#17402|

## Description

This PR fixes the wrong description in `os_version` attribute in `agent` table from the `global.db`. When the file /etc/os-release has a tag `VERSION_ID`, the empty string was not being set in the `os_version` variable.

### Before 

Was not reproducible on Arch. 

The reason why it was only reproduced with openSUSE is that the os-release file for Arch Linux there was not a `VERSION_ID` tag. But as mentioned in the issue, the behavior also may occur in Arch Linux. 

<details><summary> Expand </summary>

</details>

### After 

<details><summary> Expand </summary>

</details>

  - [x] Linux
- [x] Source installation
- [x] Source upgrade

### DIFF
--- a/src/shared/version_op.c
+++ b/src/shared/version_op.c
@@ -424,7 +424,7 @@ os_info *get_unix_version()
                         info->os_name = strdup(name);
                     }
                 } else if (strcmp (tag,"VERSION") == 0) {
-                    if (!version) {    
+                    if (!version) {
                         if (version_id) {
                             os_free(info->os_version);
                         }
@@ -456,6 +456,7 @@ os_info *get_unix_version()
         fclose(os_release);
 
         // If the OS is CentOS, try to get the version from the 'centos-release' file.
+        // If the OS is Arch Linux, openSUSE Tumbleweed set os_version as empty string.
         if (info->os_platform) {
             if (strcmp(info->os_platform, "centos") == 0) {
                 regex_t regexCompiled;
@@ -479,8 +480,10 @@ os_info *get_unix_version()
                     fclose(version_release);
                 }
             }
-            else if (strcmp(info->os_platform, "opensuse-tumbleweed") == 0) {
-                os_strdup("rolling", info->os_build);
+            else if (strcmp(info->os_platform, "opensuse-tumbleweed") == 0 ||
+                        strcmp(info->os_platform, "arch") == 0) {
+                os_free(info->os_version);
+                os_strdup("", info->os_version);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/shared/version_op.c
```diff
@@ -577,8 +579,9 @@ os_info *get_unix_version()
                 }
             }
             if (info->os_version == NULL) {
-                os_strdup("rolling", info->os_build);
+            os_strdup("", info->os_version);
```
pereyra-m, line 582: Please, fix this indentation.

src/unit_tests/shared/test_version_op.c
```diff
@@ -160,7 +308,6 @@ void test_get_unix_version_opensuse_tumbleweed(void **state)
     assert_string_equal(ret->os_version, "");
     assert_string_equal(ret->os_platform, "opensuse-tumbleweed");
     assert_string_equal(ret->sysname, "Linux");
-    assert_string_equal(ret->os_build, "rolling");
```
pereyra-m, line 163: We've removed the `rolling` word but this doesn't seem to affect the final result (at least in my Manjaro).
Do you know why?

pereyra-m (commented):
Good job!
Just a detail and a question.

I've tested this in Manjaron Linux and seems Ok

src/unit_tests/shared/test_version_op.c
MiguelazoDS: After the discussion we have, I only see that assignment for internal use and then the variable is freed, nothing about "rolling" is present in the global db, that's why I thought it was not strictly necessary to use that step in between.

src/shared/version_op.c
```diff
@@ -851,52 +854,51 @@ os_info *get_unix_version()
...
-            if (info->os_codename) {
-                char * tmp_os_version;
-                size_t len = 4;
-                len += strlen(info->os_version);
-                len += strlen(info->os_codename);
-                os_malloc(len, tmp_os_version);
-                snprintf(tmp_os_version, len, "%s (%s)", info->os_version, info->os_codename);
-                free(info->os_version);
-                info->os_version = tmp_os_version;
+	    if (strcmp(info->os_version, "") != 0) {
```
matias-braida, line 857: Indentation issue

matias-braida (commented):
GJ

src/unit_tests/shared/test_version_op.c
pereyra-m: Ok, thank you!

src/shared/version_op.c
MiguelazoDS: Fixed

matias-braida (approved):

pereyra-m (approved):
LGTM!

GabrielEValenzuela (approved):
Great job ! 😃

Dwordcito (approved):

Issue #17055, opened 2023-05-11
Fetch NVD API feeds (CVEs & CPEs)


MarcelKemp:
## Description

To change the feeds that are going to be deprecated by the new ones of the API 2.0 in the NVD, it is necessary to generate a new feed that allows us not to be limited by the API rate limit, and for that we will generate a feed for the CVEs and another one for the CPEs, which contain all the available information and that we simply have to download from the Wazuh's manager to insert it in the vulnerability databases.  

The feeds offered by the NVD are distributed from the API 2.0 in JSON format:
- CVE: `https://services.nvd.nist.gov/rest/json/cves/2.0` ([documentation](https://nvd.nist.gov/developers/vulnerabilities))
- CPE: `https://services.nvd.nist.gov/rest/json/cpes/2.0` ([documentation](https://nvd.nist.gov/developers/products))

However, the results of the number of objects returned by the API is limited to 2000 for CVEs and 10000 for CPEs, so it is necessary to make several requests and merge the results into a single feed for each case. So we will get feeds that will have a list of JSON objects with the following schemas:

<details><summary>CVE API Schema</summary>

```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "title": "JSON Schema for NVD Vulnerability Data API version 2.0",
	"definitions": {
		"def_cve_item": {
			"properties": {
				"cve": {"$ref": "#/definitions/cve_item"}
			},
			"required": ["cve"],
			"additionalProperties": false
		},
		"cve_item": {
			"type": "object",
			"properties": {
				"id": { "$ref": "#/definitions/cve_id" },
				"sourceIdentifier": { "type": "string" },
				"vulnStatus": { "type": "string" },
				"published": {"type": "string", "format": "date-time"},
				"lastModified": {"type": "string", "format": "date-time"},
				"evaluatorComment": {"type": "string"},
				"evaluatorSolution": {"type": "string"},
				"evaluatorImpact": {"type": "string"},
				"cisaExploitAdd": {"type": "string", "format": "date"},
				"cisaActionDue": {"type": "string", "format": "date"},
				"cisaRequiredAction": {"type": "string"},
				"cisaVulnerabilityName": {"type": "string"},
				"descriptions": {
				  "type": "array",
				  "minItems": 1,
				  "items": {"$ref": "#/definitions/lang_string"}
				},
				"references": {
				  "type": "array",
				  "items": { "$ref": "#/definitions/reference" },
				  "maxItems": 500,
[...]

PR #17672, opened 2023-06-27
NVD 2 CVE feed fetch refactor

mateocervilla:
|Related issue|
|---|
|#17055|

## Description

This PR aims to make the new feeds of the NVD API 2.0 fetch compatible with the code.

## Configuration options

The `<update_from_year>` tag will no longer be supported for the NVD feed.

The `<path>` and `<url>` tags will continue to work for the NVD feed, even with multiple files.

## Changes

The macro `JSON_MAX_FSIZE` was increased from 512 MB to 2 GB.

To make this possible, it was also necessary to change the data type of the `w_get_file_content()` function to `long`.


MarcelKemp (approved):
LGTM.

Issue #17615, opened 2023-06-21
Engine - Review the fields inserted by Filebeat in the processed events


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
|5.0.0 | Engine | Manager | Packages/Sources | OS version |

## Description
Currently, in the engine, we are experiencing the insertion of several fields in the events processed by Filebeat. It is crucial that we conduct an investigation into these fields and remove any that do not provide essential information.

JavierBejMen:
Host fields are added by default when using the type log in filebeat configuration, to disable them add the option:
```
publisher_pipeline.disable_host: true
```

Ref: https://www.elastic.co/guide/en/beats/filebeat/7.10/filebeat-reference-yml.html

PR #17681, opened 2023-06-28
Removed unwanted filebeat fields and added mappings for agent field

JavierBejMen:
|Related issue|
|---|
|#17615|

The following fields added by default by Filebeat has ben removed:
- host.hostname
- host.name
- agent.ephemeral_id
- input.type
- log.file.path
- log.offset

The log and host fields are only deleted if added by the Filebeat log ingestion type.

Added mappings for agent fields, as we want these fields and are not overwritten by Filebeat if already present in the event.


juliancnn (approved):

JcabreraC (approved):
LGTM !

Issue #17690, opened 2023-06-29
Fix nginx-lb image entrypoint.sh file used by the integration test docker environment


mhamra:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.4.4 and above | Integration Test | Manager/Agent | N/A | N/A |

After the investigation done in issue [17588](https://github.com/wazuh/wazuh/issues/17588), it was found that the entrypoint.sh file needs to be fixed to make nginx start (see [comment](https://github.com/wazuh/wazuh/issues/17588#issuecomment-1612333225))

mhamra:
### UPDATE

- entrypoint.sh and health check implemented in PR [17703](https://github.com/wazuh/wazuh/pull/17703)

mhamra:
### UPDATE

Test run with AWS t3a.xlarge VMs. Test succeded and documented [here](https://github.com/wazuh/wazuh/issues/17522#issuecomment-1618945766)

Issue #17588, opened 2023-06-14
API Integration tests failures during 4.4 merge


GGP1:
## Description

During the execution of the API integration tests for the merge of [4.4.4 to 4.4](https://github.com/wazuh/wazuh/pull/17578), many tests failed for unexpected reasons. Here is a detailed summary.

Artifacts: [artifacts.zip](https://github.com/wazuh/wazuh/files/11745842/artifacts.zip)

### test_agent_DELETE_endpoints

| Test | Stage | Error | Reason |
| -- | -- | -- | -- |
| DELETE /agents | Delete agent with a proper status | List item(s) not present in response | Agent inactive |

### test_agent_GET_endpoints

| Test | Stage | Error | Reason |
| -- | -- | -- | -- |
| GET /agents | Pagination with offset 0 and limit 2 | List item(s) not present in response | Agent inactive |
| GET /agents/{agent_id}/config/{component}/{configuration} | Get client configuration from agent component | Status code was 400, expected 200. Structure of returned data was different than expected | Agent inactive |
| GET /agents/{agent_id}/daemons/stats | Get all daemons' stats from agent 001 | Key mismatch: expected["error"] = '0', actual["error"] = '1' | Agent inactive |
| GET /agents/{agent_id}/stats/{component} | Get logcollector stats from agent 002 | Key mismatch: expected["error"] = '0', actual["error"] = '1' | Agent not connected |
| GET /groups/{group_id}/agents | Filter select one field | List item(s) not present in response | OS information missing |
| GET /agents?name=agent_name | Select three fields | List item(s) not present in response | Agent inactive |
| GET /agents/outdated | Filter agents by query | List item(s) not present in response | Agent disconnected |
| GET /agents/stats/distinct | Get the status and os.name | List item(s) not present in response | Information mismatch |
| GET /agents/summary/status | Get the agents status summary | Key mismatch | Agents never connected/synced |
| GET /agents/upgrade_result | Upgrade all agents (only old agents will appear in affected_items). We use q to leave agent 008 with no tasks in DB | List item(s) not present in response | No upgrades were performed |

### test_agent_PUT_endpoints

| Test | Stage | Error | Reason |
| -- | -- | -- | -- |
| PUT /agents/group | Assign agents 003, 004 and 005 to group1 | Error calling validate function '\<function healthcheck_agent_restart\>' | Healthcheck failure |
[...]

Issue #17522, opened 2023-06-12
Investigate about the API integration tests stability


GGP1:
## Description

This issue aims to initiate an investigation into the API integration tests' stability, possible improvements, and refactors. 

We noticed that they are unstable and sporadic. There is often no apparent failure in the logs, and they must be re-run to pass. This happens in `4.5`, `4.6` and in `master`.

The test typically marks many endpoints as failed, sometimes just one that is not repeated across executions.

## Related issues

- https://github.com/wazuh/wazuh/issues/17200
- https://github.com/wazuh/wazuh/issues/17588
- https://github.com/wazuh/wazuh/issues/17654
- https://github.com/wazuh/wazuh/issues/17690
- https://github.com/wazuh/wazuh-jenkins/issues/5323
- https://github.com/wazuh/wazuh-jenkins/issues/5435

#### Merge 4.4.4 into 4.4

Run: https://github.com/wazuh/wazuh/pull/17578/checks?check_run_id=14251322424

Artifacts: [artifacts.zip](https://github.com/wazuh/wazuh/files/11745842/artifacts.zip)

#### Add distinct to multiple endpoints

https://github.com/wazuh/wazuh/pull/16867/checks?check_run_id=14195098445

#### Add query, select and distinct parameters to API endpoints

https://ci.wazuh.info/job/Test_integration_endpoints/3588/

Artifacts: [artifacts.zip](https://github.com/wazuh/wazuh/files/12111531/artifacts.zip)

GGP1:
## Update

We have been investigating the AIT failures in https://github.com/wazuh/wazuh/issues/17588 and we have found two potential issues with them:

- The nginx Docker container entrypoint was using the `service` command to start the server and in the latest version of the official Docker image the command is no longer pre-installed, causing nginx to fail repeteadly at the start of the setup. A fix is being developed in https://github.com/wazuh/wazuh/issues/17690.
- The CPU utilization of the tests reached 100% and crashed the host when running them in EC2 t3a medium instances (the ones Jenkins use) as explained [here](https://github.com/wazuh/wazuh/issues/17200#issuecomment-1607841169). We are analyzing the possibility of increasing the specs of the instances to avoid this in https://github.com/wazuh/wazuh-jenkins/issues/5323

mhamra:
### UPDATE

Results after pushing commits in [PR 17703](https://github.com/wazuh/wazuh/pull/17703):

[...]

PR #17703, opened 2023-06-29
Fix Integration Test nginx-lb entrypoint and add container log to Artifacts.zip

mhamra:
|Related issue|
|---|
|[17690](https://github.com/wazuh/wazuh/issues/17690)|

## Description
This PR fixes errors detected during integration test execution. See issue [17588](https://github.com/wazuh/wazuh/issues/17588) and comments

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade


api/test/integration/env/base/nginx-lb/Dockerfile
```diff
@@ -3,3 +3,5 @@ FROM nginx
 COPY nginx.conf /etc/nginx/nginx.conf
 
 ADD entrypoint.sh /scripts/entrypoint.sh
+
+HEALTHCHECK --interval=30s --timeout=3s CMD curl --fail http://localhost/ || exit 1
```
GGP1, line 7: I don't mark it as a requested change because I'm not sure if there is a policy for this, but perhaps we should add an additional line below

GGP1, line 7: The interval seems a little bit high to me, if nginx fails at the start because the server isn't up yet, it will wait 30s to try again. We could reduce that number to 5/10 to avoid waiting too much.

The manager and agent base images healthchecks use an interval of `1s` and a timeout of `30s`, we could use similar values. Moreover, they specify a number of retries (900), we may want to set one to avoid unlimited tries. 

Lastly, I'm not sure if there is a policy for this, but perhaps we should add the EOF line

GGP1 (changes requested):

api/test/integration/env/base/nginx-lb/Dockerfile
mhamra: I agree with the suggestion to reduce the interval value, a 1-second interval seems too often, but it shouldn't affect the environment.

The timeout is ok because nginx solves this call in localhost, it just returns an HTML standard page, so the response is immediate.

I don't know if adding EOF line obeys a standard. We can delete it to maintain the format of the prior version.

api/test/integration/conftest.py
```diff
@@ -170,14 +170,19 @@ def check_health(interval: int = 10, node_type: str = 'manager', agents: list =
...
                 f"docker inspect env_wazuh-agent{agent}_1 -f '{{{{json .State.Health.Status}}}}'", shell=True)
             if not health.startswith(b'"healthy"'):
                 return False
-        return True
+    elif node_type == 'nginx-lb':
+        health = subprocess.check_output(
+            f"docker inspect env_nginx-lb_1 -f '{{{{json .State.Health.Status}}}}'", shell=True)
+        if not health.startswith(b'"healthy"'):
+            return False
+
```
mhamra, line 184: I suggest adding an else: statement to return False or Raise an exception because the node_type is invalid.

GGP1 (dismissed):
LGTM!

The cluster test failed but it's not related to nginx or the agents' connections. We suspect it might be related to the resource limits of the EC2 instances where the tests run. More about this [here](https://github.com/wazuh/wazuh-jenkins/issues/5323)

api/test/integration/env/base/nginx-lb/Dockerfile
```diff
@@ -3,3 +3,5 @@ FROM nginx
 COPY nginx.conf /etc/nginx/nginx.conf
 
 ADD entrypoint.sh /scripts/entrypoint.sh
+
+HEALTHCHECK --interval=1s --timeout=3s --retries=10 CMD curl --fail http://localhost/ || exit 1
```
Selutario, line 7: Missing empty line at the end of the file

Selutario (changes requested):

mhamra:
### UPDATE

Test run with AWS t3a.xlarge VMs. Test results documented [hear](https://github.com/wazuh/wazuh/issues/17522#issuecomment-1618945766)

Selutario (approved):

Issue #13174, opened 2022-04-19
FPs 92655 due to substring false positive


branchnetconsulting:
In Wazuh ruleset (master branch), rule 92655 matches for a single Windows Event ID like this
`<field name="win.system.eventID" type="pcre2">808</field>`
which will match any Windows event ID containing the number 808, including unintended ones like Event 33808.
Also it doesn't appear helpful to invoke pcre2 with this match when plain OS_Match or OS_Regex already works here. 

Proposed fix in pull request: https://github.com/wazuh/wazuh/pull/13173

Kevin

PR #17779, opened 2023-07-07
FPs 92655 due to substring false positive

jk-olaoluwa:
|Related issue|
|---|
|#13174|

Matching Windows eventID just for 808 will catch other eventIDs like event 33808 that have nothing to do this rule.  Always match specific event ID numbers surrounded by ^ and $ symbols so you are not just searching for a substring.  
Also this rule has no need for PCRE2.  Plain OS_Match or OS_Regex can handle what we're doing here.   I assume from a CPU performance standpoint we only want to use PCRE2 when it is actually giving us benefit beyond what we can already do easily with the older/faster OS_Match or OS_Regex options.

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests

### DIFF
--- a/ruleset/rules/0840-win_event_channel.xml
+++ b/ruleset/rules/0840-win_event_channel.xml
@@ -2,11 +2,11 @@
   Copyright (C) 2015, Wazuh Inc.
 -->
 
-<!--
+<!-- 
   General Windows Event channel detection rules: 92650 - 92700
 -->
 
-<group name="win_evt_channel,windows,">
+<group name="win_evt_channel,">
 
   <rule id="92650" level="12">
     <if_sid>61138</if_sid>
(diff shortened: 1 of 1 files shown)


72nomada (approved):
LGTM

PR #17994, opened 2023-07-21
Should be undefined for local installations.

tongson:
Building local installations fails with:

```
/usr/bin/ld: libwazuh.a(wm_database.o): in function `wm_database_main':
wm_database.c:(.text+0x8f5): undefined reference to `wm_sync_agents'
/usr/bin/ld: wm_database.c:(.text+0x90f): undefined reference to `wm_sync_legacy_groups_files'
/usr/bin/ld: wm_database.c:(.text+0x916): undefined reference to `wm_clean_dangling_wdb_dbs'
```

`wm_sync_agents`, `wm_sync_legacy_groups_files`, and `wm_clean_dangling_wdb_dbs` should be undefined for local installations.

### DIFF
--- a/src/wazuh_modules/wm_database.c
+++ b/src/wazuh_modules/wm_database.c
@@ -115,21 +115,27 @@ void* wm_database_main(wm_database *data) {
     // Manager name synchronization
     wm_sync_manager();
 
+    #ifndef LOCAL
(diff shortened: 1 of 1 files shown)


tongson:
Looks like already fixed in the tree.

Closing.

Issue #17845, opened 2023-07-10
Migrate API and Framework unit tests to Github Actions


fdalmaup:
## Description

This issue aims to migrate the API and Framework unit tests to GitHub Actions. Currently, these tests are run locally and in every pull request involving the mentioned components but without taking advantage of the functionality provided by GitHub.

### Proposed Solution

The proposed solution is to create a dedicated GitHub Actions workflow that runs the API and Framework unit tests. This workflow should be triggered on every pull request to ensure continuous code testing. The test results should be displayed as annotations in the GitHub UI, allowing contributors to quickly identify and address any potential issues.

### Tasks

- [x] Create new GitHub Actions workflow files specifically for API and Framework Unit Tests.
- [x] Configure the workflow to trigger when opening pull requests related to API and Framework modifications.
- [x] Set up the necessary steps to run the unit tests.
- [x] Ensure the test results are displayed in the GitHub UI.
- [x] Test the workflow on different branches to ensure its functionality and reliability.

GGP1:
## Update

Created the API and Framework unit tests workflow in the branch [enhancement/17845-unit-tests-migration](https://github.com/wazuh/wazuh/tree/enhancement/17845-unit-tests-migration). 

They run on every pull request that has modified the `api` or `framework` folders. It uses `setup-python` to cache dependencies and avoid installing them on every run when there are no changes.

GGP1:
## Update

Added the ability to run the actions manually and modified the list of files that trigger it.

PR #18008, opened 2023-07-21
Add API and Framework unit tests workflows to Github Actions

GGP1:
|Related issue|
|---|
| Closes https://github.com/wazuh/wazuh/issues/17845 |

## Description

Add two workflow actions to run API and Framework unit tests on pull requests that modify  `api` or `framework` directories.

They `ubuntu-latest` and python `3.9` versions. Dependencies are cached using `setup-python` to avoid installing the same dependencies multiple times.

### Executions

The workflow actions added were run in the pull request https://github.com/wazuh/wazuh/pull/18162 to verify that they work as expected.

### DIFF
--- a/.github/workflows/api-unit-tests.yml
+++ b/.github/workflows/api-unit-tests.yml
@@ -12,6 +12,8 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ['3.9']
+    env:
+      PYTHONPATH: /home/runner/work/wazuh/wazuh/api:/home/runner/work/wazuh/wazuh/framework
     steps:
       - uses: actions/checkout@v3
 
@@ -28,4 +30,4 @@ jobs:
           pip install -r framework/requirements-dev.txt
 
       - name: Run API tests
-        run: python3 -m pytest api/api
+        run: python -m pytest api/api
--- a/.github/workflows/api-unit-tests.yml
+++ b/.github/workflows/api-unit-tests.yml
@@ -2,6 +2,11 @@ name: API unit tests
 
 on:
   workflow_dispatch:
+    inputs:
+      base_branch:
+        description: 'Base branch'
+        required: true
+        default: 'main'
   pull_request:
     paths:
       - '.github/workflows/api-unit-tests.yml'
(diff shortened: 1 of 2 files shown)


EduLeon12 (dismissed):
The proposed solution works, is well executed and the scope of the issue has been fulfilled.

Would only mention that maybe is an opportunity to include post-failure steps like an event trigger to redirect logs, alerts, etc

Selutario (approved):
LGTM

.github/workflows/api-unit-tests.yml
```diff
@@ -0,0 +1,33 @@
+name: API unit tests
+
+on:
```
TomasTurina, line 3: Don't you need also the `workflow_dispatch` case? This will allow to run them manually, which can be useful to run nightly executions.

.github/workflows/api-unit-tests.yml
```diff
@@ -0,0 +1,33 @@
...
+  pull_request:
+    paths: 
+    - 'api/**'
```
TomasTurina, line 6: What about changes in this file? And also in the Makefile?

.github/workflows/framework-unit-tests.yml
```diff
@@ -0,0 +1,33 @@
+name: Framework unit tests
+
+on:
```
TomasTurina, line 3: Same here.

.github/workflows/framework-unit-tests.yml
```diff
@@ -0,0 +1,33 @@
...
+  pull_request:
+    paths: 
+    - 'framework/**'
```
TomasTurina, line 6: Same here.

TomasTurina:
Also, I don't see these tests running in the PR checks.

GGP1:
> Also, I don't see these tests running in the PR checks.

They were run in a separate [PR](https://github.com/wazuh/wazuh/pull/18162) since the file itself was not included in the paths that trigger the action. Now, it will run in this PR since both files were added to the list.

TomasTurina (approved):

PR #18188, opened 2023-08-01
Remove FIM fields uid and gid if empty

jotacarma90:
## Description
This PR is to ignore the gid and uid fields when they come with an empty string. This way the alert is cleaner, and we can take advantage of the current integration tests.

This happens because in Windows the gid field is not extracted for files.

  - [x] Linux
  - [x] Windows
- [x] Source installation

### DIFF
--- a/src/syscheckd/src/create_db.c
+++ b/src/syscheckd/src/create_db.c
@@ -1465,11 +1465,11 @@ cJSON * fim_attributes_json(const fim_file_data * data) {
 #endif
     }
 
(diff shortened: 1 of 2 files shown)


vikman90:
The failing check "RTR test folder agent" is surely related to a _sigkill_ caused by an OOM condition.

vikman90 (changes requested):

vikman90 (approved):

Issue #18312, opened 2023-08-09
Updating a rule or decoder file causes other file is removed from the ruleset


Desvelao:
|Wazuh version|Component|
|---|---|
| 4.6.0 | Wazuh component |

## Description

Wazuh server: built from sources branch `4.6.0`

Context:

I added custom directories for decoders and rules:
```
<ruleset>
...
  <!-- User-defined custom directory ruleset -->
  <rule_dir>etc/rules/custom_dir</rule_dir>
  <decoder_dir>etc/decoders/custom_dir</decoder_dir>
</ruleset>
```

Then I added rule and decoders files to the custom directories:
```
etc/rules/local_rules.xml
etc/rules/local_rules_custom.xml
etc/rules/custom_dir/local_rules.xml
etc/decoders/local_decoders.xml
etc/decoders/local_decoders_custom.xml
etc/decoders/custom_dir/local_decoders.xml
```

```console
root@wazuh-manager-460-7102:/var/ossec/etc# tree rules decoders
rules
|-- custom_dir
|   `-- local_rules.xml
|-- local_rules.xml
`-- local_rules_custom.xml
decoders
|-- custom_dir
|   `-- local_decoders.xml
|-- local_decoders.xml
`-- local_decoders_custom.xml
```

When I edit from the UI the file: `etc/decoders/custom_dir/local_decoders.xml` and save, the following API request is done: ``
![image](https://github.com/wazuh/wazuh-kibana-app/assets/34042064/09ab5da3-5a5f-4933-869b-41455bd7038f)

API log:
```
2023/08/08 13:22:44 INFO: wazuh-wui 172.26.0.7 "GET /decoders/files/local_decoders.xml" with parameters {"raw": "true", "relative_dirname": "etc/decoders/custom_dir"} and body {} done in 0.034s: 200
2023/08/08 13:23:03 INFO: wazuh-wui 172.26.0.7 "PUT /decoders/files/local_decoders.xml" with parameters {"overwrite": "true", "relative_dirname": "etc/decoders/custom_dir"} and body {} done in 0.031s: 200
2023/08/08 13:23:03 INFO: wazuh-wui 172.26.0.7 "GET /cluster/status" with parameters {} and body {} done in 0.021s: 200
2023/08/08 13:23:04 INFO: wazuh-wui 172.26.0.7 "GET /cluster/configuration/validation" with parameters {} and body {} done in 1.250s: 200
[...]

PR #18315, opened 2023-08-09
Fix  upload_decoder_file and upload_rule_file function when two different files have the same name

mhamra:
|Related issue|
|---|
|#18312|

## Description
Fix upload_decoder_file and upload_rule_file functions. 

Updating a file in a subdirectory with the same name as a file in the base user ruleset directory wrongly deletes the file in the user base directory.

### DIFF
--- a/framework/wazuh/decoder.py
+++ b/framework/wazuh/decoder.py
@@ -17,7 +17,7 @@ from wazuh.core.exception import WazuhInternalError, WazuhError
 from wazuh.core.results import AffectedItemsWazuhResult
 from wazuh.core.rule import format_rule_decoder_file
 from wazuh.core.utils import process_array, safe_move, validate_wazuh_xml, \
-    delete_file_with_backup, upload_file, to_relative_path
+    upload_file, to_relative_path, full_copy
 from wazuh.rbac.decorators import expose_resources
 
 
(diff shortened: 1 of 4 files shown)


framework/wazuh/decoder.py
```diff
@@ -354,7 +354,13 @@ def upload_decoder_file(filename: str, content: str, relative_dirname: str = Non
...
         elif overwrite and exists(full_path):
             backup_file = f'{full_path}.backup'
-            delete_file_with_backup(backup_file, full_path, delete_decoder_file)
+            try:
+                full_copy(full_path, backup_file)
+            except IOError as exc:
+                raise WazuhError(1019) from exc
+
+            delete_decoder_file(filename=filename,
+                                relative_dirname=relative_dirname)
```
Selutario, line 363: I just realized our RBAC reference documentation is wrong. `PUT /decoders/files/{filename}` requires `decoders:update` and `decoders:delete` RBAC actions, but only `decoders:update` is listed:
- https://documentation.wazuh.com/current/user-manual/api/rbac/reference.html#decoders-update

The same happens for rules too. Could you open an issue in wazuh-documentation to fix this in the future?

Selutario (approved):

Issue #16699, opened 2023-04-12
Add a version control mechanism in cve.db for Vulnerability Detector


pereyra-m:
## Description

The Vulnerability Detector module stores all the external feeds data in the `cve.db`. This DB also has some tables that are temporarily used during the scans.

It lacks any schema upgrade mechanism or version control because this DB is supposed to be deleted during the upgrade.
But this assumption might not be valid in a Docker deployment for example. So, if the tables in the DB have a breaking change in their structure, Vulnerability Detector will fail to read it and the `cve.db` will have to be manually deleted to make the module work again.

See for example
- https://github.com/wazuh/wazuh/issues/4401
- https://wazuh.slack.com/archives/C0A933R8E/p1680543595265869
- https://wazuh.slack.com/archives/C0A933R8E/p1680511376992489

## DoD

- [x] Add a new table in `cve.db` to store the schema version
- [x] Add a mechanism to check the DB version in Vulnerability Detector and recreate the DB in case it isn't compatible
- [x] Remove all the references to the removal of `cve.db` during an upgrade (in all repositories) because this step won't be needed
- [x] Make sure to properly document this change, so every future modification in the schema will be done together with a version change 
- [x] Test the upgrade in different ways: sources, packages, docker, etc.
- [x] All CI tests must pass.

Dwordcito:
Close as not planned because 4.8.0 is coming.

PR #18473, opened 2023-08-17
Add a version control mechanism in cve.db for Vulnerability Detector

tdrauncieal:
* New table to save metadata, currently only used for versioning
* version control when opening existing cve.db
* Unit Test cases for new code added.

|Related issue|
|---|
|#16699 

## Description
These changes can maintain the schema of the `vulnerability detector` DB (`cve.db`) versioned.

To do this look and compare the `version ID` stored on `cve.db` at first check of database schema, when `analisd module` is started, then a case of version mismatch, the current DB is removed and regenerated.
 

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests


src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
```diff
@@ -834,18 +834,59 @@ int wm_vuldet_create_file(const char *path, const char *source) {
...
     const char *tail;
-    sqlite3 *db = NULL;
-    sqlite3_stmt *stmt = NULL;
     int result;
     uid_t uid;
     gid_t gid;
+    bool already_removed = false;
+    bool db_existed = (access(path, F_OK) == OS_SUCCESS);
+
+opendb: ;
```
mateocervilla, line 843:
```suggestion
opendb:
```

tdrauncieal: Yes, this seems a mistake, but it's necessary  because the next initialization makes a compiler error.
https://en.cppreference.com/w/c/language/goto

mateocervilla (approved):
LGTM!

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
```diff
@@ -839,16 +839,60 @@ int wm_vuldet_create_file(const char *path, const char *source) {
...
+            int len = strlen(path) + strlen(db_suffix) + 1;
+
+            os_malloc(len, db_journal);
+            snprintf(db_journal, len, "%s%s", path, db_suffix);
+            (void)remove(db_journal);
+            os_free(db_journal);
+
+            mtdebug1(WM_VULNDETECTOR_LOGTAG, VU_DB_CHANGE_SCHEMA, path, CVE_DB_VERSION);
+            already_removed = true;
+            goto opendb;
```
MarcelKemp, line 894: To improve the readability of the code, we will try to avoid the use of `goto` in this function. 
To do so, we have several alternatives, such as `if-else` structure, use of `while` or `do-while`, or _recursive function_.

Consider the different alternatives and try to choose the most efficient and readable one possible.

MarcelKemp (changes requested):

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
```diff
@@ -839,17 +839,68 @@ int wm_vuldet_create_file(const char *path, const char *source) {
...
+        if (sqlite3_open_v2(path, &db, SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE, NULL)) {
+            mterror(WM_VULNDETECTOR_LOGTAG, VU_CREATE_DB_ERROR);
+            return wm_vuldet_sql_error(db, stmt);
+        }
+    }
+
+    // Always try to create the metadata table
+    mtdebug2(WM_VULNDETECTOR_LOGTAG, "Force create VU_METADATA if not exists");
+    sqlite3_exec(db, vu_queries[CREATE_VU_METADATA], NULL, NULL, NULL);
+
```
MarcelKemp, line 862: To avoid having to modify both the schema and the global variable, wouldn't it be better to update the value after removing the DB?

It could be something like the following:
```suggestion

    if (already_removed) {
        if (wm_vuldet_prepare(db, vu_queries[UPDATE_VU_METADATA], -1, &stmt, NULL) != SQLITE_OK) {
            return wm_vuldet_sql_error(db, stmt);
        }
        sqlite3_bind_text(stmt, 1, CVE_DB_VERSION, -1, NULL);

        if (wm_vuldet_step(stmt) != SQLITE_DONE) {
            return wm_vuldet_sql_error(db, stmt);
        }
        wdb_finalize(stmt);
    }

```

src/wazuh_db/schema_vuln_detector.sql
```diff
@@ -8,6 +8,9 @@
 
 BEGIN;
 
+/* db_version value must be the same of CVE_DB_VERSION constant, it must be updated on every schema change */
```
MarcelKemp, line 11: If changes were to be applied, modify or delete code comments
```suggestion
```

src/wazuh_modules/vulnerability_detector/wm_vuln_detector_db.h
```diff
@@ -15,6 +15,10 @@
 #define CVE_DBS_PATH            "queue/vulnerabilities/"
 #define CVE_DB CVE_DBS_PATH     "cve.db"
 
+// every version change should be synced with metadata table (VU_METADATA), 
+// see schema_vuln_detector.sql
```
MarcelKemp, line 19: Same
```suggestion
// This value must be updated each time a change is made to schema_vuln_detector.sql
```

MarcelKemp (changes requested):

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
MarcelKemp: Well, I'd rather simplify the upgrade than not do it at all because the code is already tested.
Because I don't see it as unnecessary, but in my opinion it would be a more sustainable code.

MarcelKemp (changes requested):

mateocervilla (approved):
LGTM!

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
tdrauncieal: Reviewing this suggestion again, I understand that this code added as suggested here would imply the elimination of part of the current code, if so, what would be the code to eliminate?

MarcelKemp: It does not imply the removal of any code, it is simply an addition to update the schema versioning to the same version set in the `CVE_DB_VERSION` variable in case the DB has been updated.

In any case, if you add this change, you must also add the corresponding query: `UPDATE_VU_METADATA`.

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
```diff
@@ -839,17 +839,67 @@ int wm_vuldet_create_file(const char *path, const char *source) {
...
+        // ...delete and create a new one if schema version don't match
+        result = wm_vuldet_step(stmt);
+        if (result != SQLITE_ROW) {
+
+            if (already_removed) {
+                already_removed = false;
+                return wm_vuldet_sql_error(db, stmt);
+            }
+
+            wdb_finalize(stmt);
```
tdrauncieal, line 879: Knowing this, then this entire section of code would be duplicated.

src/wazuh_modules/vulnerability_detector/wm_vuln_detector.c
tdrauncieal: Knowing this, then the code section that goes from L864-L879 would be duplicated.

tdrauncieal:
## Update
* finding the best solution to make changes in one place

tdrauncieal:
## Update
* Working on a solution similar to the DB schema
* correct some UT (still working)
### ETA update
* full testing will made, due to the changes
* fix UT issue pending
* needs to be reviewed

tdrauncieal:
## Update
* UTs fixed
* test possible DB configuration scenarios

tdrauncieal:
## Update
* working on failed checks, still failing it yet.

Issue #17970, opened 2023-07-19
Unable to extract data from syscheck.path


juliamagan:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.4.5-40413 | FIM | Manager/Agent | Packages | CentOS 8/Windows Server 2022|

## Description

When we try to click on the `syscheck.path` file we get:
![not fetch](https://github.com/wazuh/wazuh/assets/80041853/8f28b93c-906a-4cd1-8f10-382c8abc5f48)
![full error](https://github.com/wazuh/wazuh/assets/80041853/e2f7e0de-b3d4-4324-8626-f47479be9b04)

This happens both from the event, as well as from the agent initial page, but it does not happen in the `Inventory` tab of FIM of the agent.

If we enable `wazuh_db.debug=2`, we can see that the error appears:

```
2023/07/19 15:53:18 wazuh-db[66986] wdb_parser.c:558 at wdb_parse(): DEBUG: DB(001) Cannot execute SQL query.
```

The queries that it tries to execute are:

```
2023/07/19 15:53:18 wazuh-db[66986] wdb_parser.c:263 at wdb_parse(): DEBUG: Agent 001 query: sql select count(*) from (select mtime as 'mtime',attributes as 'attributes',perm as 'perm',inode as 'inode',gname as 'gname',md5 as 'md5',sha1 as 'sha1',file as 'file',type as 'type',sha256 as 'sha256',arch as 'arch',gid as 'gid',changes as 'changes',date as 'date',uname as 'uname',value_type as 'value.type',value_name as 'value.name',uid as 'uid',size as 'size' from fim_entry where (file = 'c:\0-test-doc' collate nocase)  ((type = 'file' collate nocase) or (type = 'registry_key' collate nocase))  )
[...]

PR #18509, opened 2023-08-18
Adds $ to the regex pattern in WazuhDBQuery.

RamosFe:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/17970|

## Description

Adds new character for the regex pattern due to errors matching file names with a `$` character.


mhamra (dismissed):
It is a small change to a REGEX expression to add '$' as a valid character in queries.

The new test cases added to the unit test are working.

LGTM!

Selutario (changes requested):
Changes look good to me, but the PR should be merged in the next patch (4.5.3)

Selutario (approved):

Issue #18436, opened 2023-08-15
Show a correct message to the user when upgrade incorrectly through WPK


mauromalara:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.5.1-rc1 | WPK Upgrade | Manager | Packages | Ubuntu 22.04 LTS |

## Description

While executing a [manual test in RC1](https://github.com/wazuh/wazuh/issues/18409) we found a lack of clarity in the error displayed when upgrading a macOS (**arm64**) agent through WPK using an **intel64** package.

### Steps to reproduce

1. Try to upgrade a macOS (arm64) agent using an intel64 WPK package (check the related issue for additional details)

### Expected results

- A clear message about what is going wrong (the user is making a mistake)

### Current results

- The following message is shown when upgrading:
```
Agent information not found in database
concurrent.futures.process._RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/var/ossec/framework/python/lib/python3.9/concurrent/futures/process.py", line 246, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/var/ossec/framework/python/lib/python3.9/site-packages/wazuh-4.5.1-py3.9.egg/wazuh/core/cluster/dapi/dapi.py", line 247, in run_local
    data = f(**f_kwargs)
  File "/var/ossec/framework/python/lib/python3.9/site-packages/wazuh-4.5.1-py3.9.egg/wazuh/rbac/decorators.py", line 423, in wrapper
    result = func(*args, **kwargs) if not skip_execution else None
  File "/var/ossec/framework/python/lib/python3.9/site-packages/wazuh-4.5.1-py3.9.egg/wazuh/agent.py", line 989, in upgrade_agents
    raise WazuhInternalError(error_code, cmd_error=True, extra_message=agent_result['message'])
wazuh.core.exception.WazuhInternalError: Error 1816 - Agent information not found in database
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/var/ossec/framework/python/lib/python3.9/site-packages/wazuh-4.5.1-py3.9.egg/wazuh/core/cluster/dapi/dapi.py", line 292, in execute_local_request
    data = await asyncio.wait_for(task, timeout=timeout)
  File "/var/ossec/framework/python/lib/python3.9/asyncio/tasks.py", line 442, in wait_for
    return await fut
[...]

PR #18510, opened 2023-08-18
Catch Wazuh internal errors

GGP1:
|Related issue|
|---|
| https://github.com/wazuh/wazuh/issues/18436 |

## Description

Catches the `WazuhInternalError` exceptions that are thrown from the framework side.

Used the generic `1000` error code since using the one from Core's (`1836`) is reserved. Also, raising the `WazuhInternalError` directly still showed the traceback for some reason that we couldn't identify.

## Logs/Alerts example

```console
root@ip-172-31-88-29:/var/ossec/framework/python/lib/python3.9/site-packages/wazuh-4.5.1-py3.9.egg/scripts# /var/ossec/framework/python/bin/python3 /var/ossec/framework/python/lib/python3.9/site-packages/wazuh-4.5.1-py3.9.egg/scripts/agent_upgrade.py -a 001 -f /home/ubuntu/wazuh_agent_v4.5.1_macos_x86_64.wpk -F
Error 400: Wazuh Internal Error: Error 1836 - The WPK package used does not fit with the agent's architecture
```

### DIFF
--- a/framework/scripts/agent_upgrade.py
+++ b/framework/scripts/agent_upgrade.py
@@ -25,7 +25,7 @@ try:
     from wazuh.core import common
     from wazuh.core.agent import Agent
     from wazuh.core.cluster.dapi.dapi import DistributedAPI
-    from wazuh.core.exception import WazuhError
+    from wazuh.core.exception import WazuhException
 except Exception as e:
     print("Error importing 'Wazuh' package.\n\n{0}\n".format(e))
     exit()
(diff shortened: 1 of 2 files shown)


Selutario:
Fixed in 
- https://github.com/wazuh/wazuh/pull/18521

The `dapi` is used by all API endpoints. Modifying which exceptions are caught and raised could greatly impact the framework, so any change should be carefully studied. For this reason, I close this PR in favor of the one mentioned above.

PR #18680, opened 2023-08-28
Remove deprecated plain text auth from AWS and Azure wodles

### LINKED ISSUE #17975: Remove deprecated plain text authentication method from AWS and Azure
| Affected integration |
|---|
| AWS, Azure |

## Description

As part of the deprecation process initiated in #14508, we must remove the authentication option that allows the user to store their credentials as plain text in the `ossec.conf`. 

The following changes need to be done in the appropriate version (at first it was version `5.0`): 

1. Update the ossec.conf parser removing these parameters as part of the module configuration.
2. Update -h option of the python modules removing the parameters.
3. Update the corresponding python modules to show an error if the user tries to manually run it using these parameters.
4. Update the Wazuh documentation to accommodate these changes, removing the parameters and deprecation messages from the reference. The documentation changes will be issued in wazuh/wazuh-documentation#6363.

## Tasks
- [x] Test in a manager.
- [x] Test in an agent.
- [x] Unit tests without failures. Updated if there are any relevant changes.
- [x] Integration tests without failures. Updated if there are any relevant changes.
- [x] Update the documentation if necessary.
- [x] Add entry to the changelog if necessary.

nico-stefani:
|Related issue|
|---|
|#17975|

## Description

This PR closes #17975. Removes the deprecated plain text methods for AWS and Azure integrations.

### DIFF
--- a/src/config/wmodules-aws.c
+++ b/src/config/wmodules-aws.c
@@ -657,7 +657,8 @@ int wm_aws_read(const OS_XML *xml, xml_node **nodes, wmodule *module)
                 os_strdup(aws_config->bucket, cur_bucket->bucket);
             }
         }
-        else if (aws_config->remove_from_bucket) {
+
+        if (aws_config->remove_from_bucket) {
             cur_bucket->remove_from_bucket = aws_config->remove_from_bucket;
         }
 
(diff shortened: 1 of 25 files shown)


GGP1 (changes requested):

Selutario (approved):
LGTM

Issue #18306, opened 2023-08-08
Provide default SCA policies for operating systems (Tier 1)


ooniagbi:
This issue will track the development of default SCA policies adapting CIS benchmarks for the OSs listed below.

| Status              | Description|
| ---------------- | --------------------|
| 🟢        | The OS has an SCA policy that uses the latest available CIS benchmark.                |
| 🟠         | The OS has an SCA policy that uses an older CIS benchmark.                 |
| 🔴          | The OS does not have an SCA policy.                 |

## OS and CIS Benchmark actions 
| Operating system              | Status           | Currently supported CIS benchmark| Latest CIS Benchmark available                       | Life Cycle Status | Issue    | Issue  status  |
| -- | -------------------------------------------|-------------------------------------------------- | --------------------------- | ------------------ | --------| --------|
| Red Hat Enterprise Linux 7       | 🟠                  | v3.0.1                                               | v3.1.1                                               | Rework               | #17355 | Done ✅ (Merged to 4.9.0) |
| Red Hat Enterprise Linux 8       | 🟠                 | v1.0.1                                               | v2.0.0                                               | Rework               | #17356      | Done ✅ (Merged to 4.9.0) |
| Red Hat Enterprise Linux 9      | 🟢                   | v1.0.0  | No new version                                    | Rework | #17221      | Done ✅ (Merged to 4.10.0) |                                              
| CentOS 7       | 🟠                  | v3.0.0                                               | v3.1.2                                               | Rework | #14359      | Done ✅ (Merged to 4.9.0) |
| CentOS 8       | 🟠                  | v1.0.0          | v2.0.0                                               | Rework               | #14704      | Done ✅ (Merged to 4.9.0) |
| Debian 10       | 🟠                  | v1.0.0     | v2.0.0                                        | Rework               | #14702      | Done ✅ (Merged to 4.10.0) |      
| Debian 11       |🟢                 | v1.0.0     | No new version                                       | Maintenance          |         N/A    | N/A |
| Debian 12       | 🔴                | N/A     | v1.0.0 (In draft)                                                          | Create (using Debian Family benchmark)               | #17325      | Done ✅ (Merged to 4.7.2) |
| Ubuntu 18       | 🟠                  | v2.0.1     | v2.1.0                                               | Rework               | #17361      | Done ✅ (Merged to 4.10.0) |
[...]

Issue #15463, opened 2022-11-22
Create SCA Policy for Alma Linux OS 8


72nomada:
| Component | Action type           | Main Issue 
| --- | --- | ---
| SCA       | Create |  #14358 

## Main tasks

- [x] Use the latest CIS benchmark PDF Draft version - CIS Alma Linux OS 8 Benchmark v2.0.0 - v3.0.0 now available, will be updated after 4.8.1
- [x] Verify IDs numbers.
- [x] Verify texts are correct: Title, Description, Rationale and Remediation.
- [x] Verify Compliance: CIS, CIS_CSC.
- [x] Verify condition and rules:
  - [x] To Pass.
  - [x] To Fail.

  
### Issue QA:
- TBD

### Syntax and semantic

- [x] a) ID of each policy must be contiguous.
- [x] b) The order and format set in [Documentation](https://documentation.wazuh.com/current/user-manual/capabilities/sec-config-assessment/creating-custom-policies.html) must be respected.
- [x] c) YML must be valid to avoid errors.

### Content

- [x] a) Compare each check with its analog from CIS Benchmark.
- [x] b) Try maintaining each rule as similar as possible to the `Audit` section from the CIS check.
- [x] c) Check that the commands provide the expected output.
- [x] d) When a failure is discovered, check similar policies to avoid repetition of the issue.

### Unit testing

- [x] a) Output from `ossec.log` after the SCA scan and a raw output of the result of the checks.
<details><summary>Tests results</summary>
<p>

#### Analysisd (server or local)
analysisd.debug=2

#### Auth daemon debug (server)
authd.debug=0

#### Exec daemon debug (server, local, or Unix agent)
execd.debug=0

#### Monitor daemon debug (server, local, or Unix agent)
monitord.debug=0

#### Log collector (server, local or Unix agent)
logcollector.debug=0

#### Integrator daemon debug (server, local, or Unix agent)
integrator.debug=0

#### Unix agentd
agent.debug=2

```
2024/01/03 13:09:54 sca[5652] wm_sca.c:2598 at wm_sca_send_summary(): DEBUG: Sending summary event for file: 'cis_alma_linux_8.yml'
[...]

PR #18741, opened 2023-08-31
Create SCA Policy for Oracle Linux 9

jk-olaoluwa:
|Wazuh version| Component | 
|---| --- |
| 4.8.1 | SCA | 

| Component | Action type           | Main Issue 
| --- | --- | ---
| SCA       | Create |  #18306

## Main tasks

- [x] Use the latest CIS benchmark PDF - CIS Oracle Linux 9 Benchmark v1.0.0
- [x] Verify IDs numbers.
- [x] Verify texts are correct: Title, Description, Rationale and Remediation.
- [x] Verify Compliance: CIS, CIS_CSC.
- [x] Verify condition and rules:
  - [x] To Pass.
  - [x] To Fail.

### Syntax and semantic

- [x] a) ID of each policy must be contiguous.
- [x] b) The order and format set in [Documentation](https://documentation.wazuh.com/current/user-manual/capabilities/sec-config-assessment/creating-custom-policies.html) must be respected.
- [x] c) YML must be valid to avoid errors.

### Content

- [x] a) Compare each check with its analog from CIS Benchmark.
- [x] b) Try maintaining each rule as similar as possible with the `Audit` section from the CIS check.
- [x] c) Check that the commands provide the expected output.
- [x] d) When a failure is discovered, check similar policies to avoid repetition of the issue.

### Unit testing

- [x] a) Output from `ossec.log` after the SCA scan and a raw output of the result of the checks.
<details><summary>Tests results</summary>
<p>

```
2024/01/09 12:02:05 sca[6619] wm_sca.c:2598 at wm_sca_send_summary(): DEBUG: Sending summary event for file: 'cis_oracle_linux_9.yml'
2024/01/09 12:02:05 sca[6619] wm_sca.c:271 at wm_sca_send_alert(): DEBUG: Sending event: {"type":"summary","scan_id":851002259,"name":"Center for Internet Security Oracle Linux 9 Benchmark v1.0.0.","policy_id":"cis_oracle_linux_9","file":"cis_oracle_linux_9.yml","description":"This document provides prescriptive guidance for establishing a secure configuration posture for Oracle Linux 9 systems running on x86 and x64 platforms. This document was tested against Oracle Linux 9.","references":"https://www.cisecurity.org/cis-benchmarks/","passed":83,"failed":95,"invalid":2,"total_checks":180,"score":46.629215240478516,"start_time":1704801715,"end_time":1704801722,"hash":"0252e3c1962d038f2f556cc6953c077f4e2967690202592f4bf0a1c2a16fe8b7","hash_file":"3508a6529a4aba49cae96311e9ab43b7d2263389640dc3b3b30815954765ef83","first_scan":1}

```

#### Analysisd (server or local)
analysisd.debug=2

#### Auth daemon debug (server)
authd.debug=0

#### Exec daemon debug (server, local, or Unix agent)
execd.debug=0

#### Monitor daemon debug (server, local, or Unix agent)
monitord.debug=0

#### Log collector (server, local or Unix agent)
logcollector.debug=0

#### Integrator daemon debug (server, local, or Unix agent)
integrator.debug=0

#### Unix agentd
agent.debug=2

</p>
</details>

### Deployment

- [x] a) If the policy it's new, it must be added to the `sca.files` templates.
- [x] b) If the OS has many supported SCA policies, a policy must be set as the default policy. ([as example](https://github.com/wazuh/wazuh/blob/master/etc/templates/config/centos/sca.files))

### Documentation

- [x] a) Ensure documentation SCA list includes the created or updated SCA.


sej7278:
is this based off of vNext as there currently is no v2.0.0 benchmark (we're working on it now).

also oracle linux support is included in the rhel9 SCA policy:

https://github.com/wazuh/wazuh/blob/master/ruleset/sca/rhel/9/cis_rhel9_linux.yml#L28

its probably better to update that - we also need to add AlmaLinux and Rocky Linux to the requirement check.

jk-olaoluwa:
Hello @sej7278  https://github.com/wazuh/wazuh/pull/18741#issuecomment-1704302791

This issue is based off of v1 as that is the current version available for download from CIS.

Like you rightly mentioned, OL9 support is included in the RHEL9 SCA Policy but we are creating its own policy from the CIS document provided by CIS since it contains minor textual and command differences.

Also, we regularly update all versions of the supported OSes; including OL, RHEL and every other one.

Thank you for your observation and recommendation!

sej7278:
Hmm ok, so if we want an AlmaLinux SCA, should we copy and modify the RHEL one, is that what you've done here? 

Again AlmaLinux will have minor differences such as no subscription-manager and repogpgcheck=1 can be enabled.

When writing the benchmarks we have a common fedora doc which we copy and modify to create individual benchmarks for RHEL, AlmaLinux, oracle etc; so makes sense to do the same here.

jk-olaoluwa:
@sej7278 - https://github.com/wazuh/wazuh/pull/18741#issuecomment-1705663301
_so if we want an AlmaLinux SCA, should we copy and modify the RHEL one, is that what you've done here?_ - Yes, since majority of the rules are not entirely different, we do not have to write the rules from scratch, we focus on creating rules for the new/different commands/audit checks in the SCA.

_Again AlmaLinux will have minor differences such as no subscription-manager and repogpgcheck=1 can be enabled._ - this is correct, this is why we try to create SCA for each OS with the pdfs provided by CIS

_When writing the benchmarks we have a common fedora doc which we copy and modify to create individual benchmarks for RHEL, AlmaLinux, oracle etc; so makes sense to do the same here_. - we adopt this method too

In this case, since we have both Alma 8 and 9 in the work, you can copy and modify the RHEL one since you already know the differences so that you are sure that the checks cover everything mentioned in the documents.
I put the issues for Alma below should you want to track the progress
https://github.com/wazuh/wazuh/issues/15463 - Alma 8
https://github.com/wazuh/wazuh/issues/16616 - Alma 9

ooniagbi (approved):
LGTM!

Issue #18896, opened 2023-09-08
Unchecked return value at file_op.c


vikman90:
|Wazuh version|Component|CID|Source|
|---|---|---|---|
| 4.6.0 r40600 (dev) | Shared library|1567823|#8309|

Calling `rename_ex` without checking return value:

https://github.com/wazuh/wazuh/blob/f05d708c7f319fc774c6854c9c6b08760a502014/src/shared/file_op.c#L757

Issue #8309, opened 2021-04-21
Error when opening Rootcheck files on Windows


chemamartinez:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.2.0 | Rootcheck | Agent | All | Windows Server 2019 |

### Description

This is a bug reported by the @wazuh/qa team in the unstable version 4.2.

Sometimes, when the Windows agent is installed and started for the first time, its first Rootcheck scan is not able to find the benchmarks located at `$INSTALLDIR\ossec-agent\shared`, showing the following errors:

![image](https://user-images.githubusercontent.com/29475387/115577130-0cf70a00-a2bc-11eb-827e-a97af21da182.png)

The steps to reproduce it are:

- Install Wazuh 4.2.0 in a Windows Server 2019
- Connect it to a manager
- Wait until the first Rootcheck scan

alromeros:
**#8309 update:**

Hello team,

After a thorough investigation, we have come to the following conclusions regarding the error described in this issue:

- Said error is happening due to a race condition present in **Windows** agents. This race condition occurs when the `fopen()` function in `run_rk_check.c` attempts to open a file that has been previously deleted by `cldir_ex_ignore()`. These deleted files are supposed to be restored before any opening attempt.

```c
// cldir_ex_ignore() is called in receiver-win.c, which attempts to erase all the content inside ossec-agent\shared
int cldir_ex_ignore(const char * name, const char ** ignore) {    
        ...
        // Erase content    
        ...
        if (rmdir_ex(path) < 0) {
            closedir(dir);
            return -1;
        }
        ...
````

```c
// In run_rk_check.c, fopen() attempts to open the file after it's been deleted
            ...
            fp = fopen(rootcheck.winmalware, "r");
            if (!fp) {
                mterror(ARGV0, "No winmalware file: '%s'", rootcheck.winmalware);
            }
            ...
````

This can be verified by looking at this next log, which has been obtained with a modified version of the code with additional warning messages:


![timestamp](https://user-images.githubusercontent.com/52510206/115902374-ab24d480-a462-11eb-9947-d8761b676be9.PNG)



- As this [commit](https://github.com/wazuh/wazuh/commit/e813188601d4089f495086ac66a75717cde4347e) shows, this behavior has been present in **Wazuh** since **v3.0.0**. However, it has not been replicated until **v4.2.0** due to the randomness and rarity of the race condition.

[...]

PR #18964, opened 2023-09-11
Fix to check function output in case of error

MarcelKemp:
|Related issue|
|---|
|#18896|

## Description

With this PR, we avoid the bug reported by Coverity, because it was not checking the return value of the `rename_ex` function, which tells you if the action has been performed correctly or if there is an error when executing the `rename`.

|Snapshot ID|Coverity version|Platform|Total detected|Newly detected|Newly eliminated|
|---|---|---|---|---|---|
|🔴|1567823|Unchecked return value|Medium|September 7, 2023|Shared|#8309|**#18896**|

-> `Fixed`

## Coverity report

- https://scan9.scan.coverity.com/reports.htm#v55589/p12779
 - Snapshot ID: `70941`
image

## Tests

  - [x] Linux
- [x] Source installation

- Memory tests for Linux
  - [x] Scan-build report
  - [x] Coverity
  - [ ] AddressSanitizer

### DIFF
--- a/src/shared/file_op.c
+++ b/src/shared/file_op.c
@@ -754,7 +754,9 @@ int UnmergeFiles(const char *finalpath, const char *optdir, int mode, const char
         }
 
         /* Mv to original name */
-        rename_ex(tmp_file, final_name);
+        if (rename_ex(tmp_file, final_name) != 0) {
+            ret = 0;
+        }
 
         if (unmerged_files != NULL) {
             /* Removes path from file name */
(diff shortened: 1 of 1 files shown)


src/shared/file_op.c
```diff
@@ -754,7 +754,9 @@ int UnmergeFiles(const char *finalpath, const char *optdir, int mode, const char
         }
 
         /* Mv to original name */
-        rename_ex(tmp_file, final_name);
+        if (rename_ex(tmp_file, final_name) != 0) {
+            ret = 0;
```
vikman90, line 758: Wouldn't it seem appropriate to clean up and go out? I mean that this code would let the code fill `unmerged_files` and also leave the temporary file missing.

```suggestion
            unlink(tmp_file);
            ret = 0;
            break;
```

vikman90 (changes requested):

vikman90 (approved):

Issue #18935, opened 2023-09-11
Engine - Update windows decoders


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
|5.0.0 | Engine | Manager | Packages/Sources | OS version |

### Description

The objective of this issue is to enhance the `windows-security.yml` decoder in the Wazuh engine to align it more closely with the established syntax and improve overall decoding consistency. This update will not only make the decoder more complete but also serve as a reference for creating other decoders.

### Tasks

- [x] **Remove Parent Decoder References**
   - In `windows-event.yml`, remove the following lines:
     ```
     parents:
       - decoder/integrations/0
     ```

- [x] **Change Variable Name**
   - In `windows-event.yml`, change the variable name `_MapWindowsLevelToLogLevel` to `_win_log_level`.

- [x] **Unify Formatting for Multi-line Checks**
   - Ensure that all multi-line checks in decoders use the `>-` format for consistency.

- [x] **Include Empty Line**
   - In `windows-security.yml`, include an empty line before `# Generic fields comment` to improve readability.

- [x] **Remove Unnecessary Single Quotes**
   - Review and remove unnecessary single quotes from the `windows-security.yml` decoder.

- [x] **Create Helper Function for User Account Control Attributes**
   - Create a helper function in the `windows-security.yml` decoder to set User Account Control Attributes.

- [x] **Create Helper Function for Kerberos Ticket Options**
   - Create a helper function in the `windows-security.yml` decoder to set Kerberos Ticket Options.

- [x] **Update Variable Name `kerberos_encryption`**
   - Replace `kerberos_encryption` with `_kerberos_encryption` in `windows-security.yml` (two occurrences).

- [x] **Fix Variables in Set Audit Information**
   - Ensure that `subcategory_guid` and `audit_information_array` are treated as variables in `windows-security.yml`.

- [x] **Fix Variable `failure_code`**
    - Treat `failure_code` as a variable in `windows-security.yml`.

- [x] **Fix Variable `policy_changes`**
    - Treat `policy_changes` as a variable in `windows-security.yml`.

- [x] **Fix Variable `access_list`**
    - Treat `access_list` as a variable in `windows-security.yml`.

- [x] **Implement Helper Function for Array of Values**
    - Implement a helper function in `windows-security.yml` to return an array of values.

- [x] **Create Helper Function for Access Mask Descriptions**
[...]

PR #18971, opened 2023-09-12
Update of the windows decoders in the engine

JcabreraC:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/18935|

The following tasks will be performed on the windows decoders:

- [x] Remove Parent Decoder References
- [x] Change Variable Name
- [x] Unify Formatting for Multi-line Checks
- [x] Include Empty Line
- [x] Remove Unnecessary Single Quotes
- [x] Create Helper Function for User Account Control Attributes
- [x] Create Helper Function for Kerberos Ticket Options
- [x] Update Variable Name `kerberos_encryption`
- [x] Fix Variables in Set Audit Information
- [x] Fix Variable `failure_code`
- [x] Fix Variable `policy_changes`
- [x] Fix Variable `access_list`
- [x] Implement Helper Function for Array of Values
- [x] Create Helper Function for Access Mask Descriptions
- [x] Avoid Duplicating Elements in Related Arrays
- [x] Define Variable `member_name_parts`
- [x] Define Variable `group_domain`
- [x] Define Variable `process_parts`
- [x] Define Variables `process_parts` and `parent_process_parts`
- [x] Define Variables `file_parts` and `file_parts.ext`
- [x] Create Helper Functions for Object Policy Change and SidListDesc
- [x] Eliminate Extra Spaces


soynof:
Please, check clang-tidy format and trailing spaces.

src/engine/source/kvdb/src/kvdbHandler.cpp
```diff
@@ -32,8 +32,9 @@ std::optional<base::Error> KVDBHandler::set(const std::string& key, const std::s
             }
             else
             {
-                return base::Error {
-                    fmt::format("Can not save value '{}' in key '{}'. Error: {}", value, key, status.getState())};
+                std::string_view error = status.getState() != nullptr ? status.getState() : "Unknown";
+                pRocksDB->Put(rocksdb::WriteOptions(), pCFhandle.get(), rocksdb::Slice(key), rocksdb::Slice(value));
```
soynof, line 36: Why is put called again in case of error?

src/engine/test/source/builder/builders/opBuilderKVDBGetArray_test.cpp
```diff
@@ -0,0 +1,235 @@
...
+#include <metrics/metricsManager.hpp>
+
+using namespace metricsManager;
+
+namespace
+{
+using namespace base;
+using namespace builder::internals::builders;
+
+std::filesystem::path uniquePath()
```
soynof, line 25: Could we relocate this functionality to a common place to avoid duplicating the code and be able to reuse it when needed?

soynof (changes requested):

src/engine/source/builder/builders/opBuilderHelperMap.hpp
```diff
@@ -384,6 +375,10 @@ base::Expression opBuilderHelperHashSHA1(const std::string& targetField,
                                          const std::vector<std::string>& rawParameters,
                                          std::shared_ptr<defs::IDefinitions> definitions);
 
+//*************************************************
```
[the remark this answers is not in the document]
juliancnn: no.

src/engine/test/source/builder/builders/opBuilderKVDBGetArray_test.cpp
juliancnn: Not really, these tests are poorly located and it is necessary to move them to the next iteration of KVDB so they will not share files.

src/engine/source/builder/builders/opBuilderHelperMap.cpp
```diff
@@ -986,62 +986,184 @@ base::Expression opBuilderHelperRegexExtract(const std::string& targetField,
...
+        {
+            throw std::runtime_error(
+                fmt::format("\"{}\" function: Target field \"{}\" is not an array", name, targetField));
+        }
 
-    // Return result
-    return base::Term<base::EngineOp>::create(
-        name,
-        [=, targetField = std::move(targetField)](base::Event event) -> base::result::Result<base::Event>
+        bool first = true;
```
NahuFigueroa97, line 1011: improve - desing: to follow this part is a few complex.
The logic could be reorganized so that reference parameters and regular parameters are treated in separate blocks to avoid repetitions.
You could use just one variable like type = json::Json::Type type; but initialized in json::Json::Type::Null for example. If type is still json::Json::Type::Null, we set it to the type of the first parameter found. If type has already been set, we compare the current type with type and throw an exception if they do not match.

These are suggestions that I think would make the code more readable, please let me know if any of this is not possible.

src/engine/source/builder/builders/opBuilderKVDB.cpp
```diff
@@ -446,4 +446,363 @@ HelperBuilder getOpBuilderKVDBDelete(std::shared_ptr<IKVDBManager> kvdbManager,
...
+                {
+                    return base::result::makeFailure(event, referenceNotFound);
+                }
+                if (keys.value().empty())
+                {
+                    return base::result::makeFailure(event, emptyArray);
+                }
+
+                // Get values from KVDB
+                bool first = true;
```
NahuFigueroa97, line 524: improve-desing: ditto. I think we can use just one variable

src/engine/source/builder/builders/windowsHelper.cpp
```diff
@@ -0,0 +1,224 @@
...
+ * @return std::vector<std::string>
+ */
+std::vector<std::string> parserListSID(const std::string& listSrt)
+{
+
+    std::vector<std::string> result = base::utils::string::split(listSrt, ' '); // TODO Check format
+
+    for (auto& sid : result)
+    {
+        if (sid.size() > 3)
```
NahuFigueroa97, line 31: magic number

src/engine/source/builder/builders/windowsHelper.cpp
```diff
@@ -0,0 +1,224 @@
...
+        {
+            // Remove header '%{' and tail '}'
+            sid = sid.substr(2, sid.size() - 3);
```
NahuFigueroa97, line 34: ditto

src/engine/source/builder/builders/windowsHelper.cpp
```diff
@@ -0,0 +1,224 @@
...
+               std::shared_ptr<defs::IDefinitions> definitions) -> base::Expression
+    {
+        // Process parameters
+        auto parameters = processParameters(rawName, rawParameters, definitions);
+
+        // Format name
+        const auto name = formatHelperName(targetField, rawName, parameters);
+
+        // Check parameters
+        checkParametersSize(rawName, parameters, 2);
```
NahuFigueroa97, line 63: magic number

src/engine/source/builder/builders/windowsHelper.cpp
```diff
@@ -0,0 +1,224 @@
...
+        {
+            throw std::runtime_error(
+                fmt::format("Error getting the kvdb handler: {}", base::getError(kbdbRes).message));
+        }
+
+        auto kvdbHandler = base::getResponse<std::shared_ptr<kvdbManager::IKVDBHandler>>(kbdbRes);
+
+        // Get the lists
+        // Account SID Description
+        std::map<std::string, std::string> asdMap;
```
NahuFigueroa97, line 91: improve-desing: Both adsMap and dssMap have the same logic, could repeated code be eliminated?

src/engine/source/builder/builders/opBuilderHelperMap.cpp
juliancnn: The problem here is that null is a valid type, that's why we don't mix up the use of the variable. The type variable is used to identify the type (and all types are valid) and the first variable is used to know if the type is set.

src/engine/source/builder/builders/opBuilderKVDB.cpp
juliancnn: same

src/engine/source/builder/builders/windowsHelper.cpp
juliancnn: How do you propose to eliminate this magic number? This number is defined here because it is the definition of the size of the arguments

src/engine/source/builder/builders/windowsHelper.cpp
juliancnn: Yes!, done

soynof (approved):
LGTM!

Issue #18882, opened 2023-09-07
Compilation error in HP-UX 4.6.0 agent


rauldpm:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| At least 4.6.0 | -- | Agent | Packages | HP-UX |

## Description

During the test installation of the HP-UX 4.6.0 package, an error was encountered when starting the agent as the `wazuh-execd` binary was not found, apparently, it's not the only missing binary

```
# /var/ossec/bin/wazuh-control start
/var/ossec/bin/wazuh-control[4]: /var/ossec/bin/wazuh-execd:  not found.
wazuh-execd: Configuration error. Exiting
```

```
# ls -l /var/ossec/bin
total 16
-rwxr-x---   1 root       root          7144 Sep  7 05:21 wazuh-control
# ls -l /var/ossec/bin | wc -l
2
```

- This seems to be due to a compiler failure during package generation (thanks to @jnasselle for the hint):

```
13:19:19  In file included from /home/jfzem/hp-ux/wazuh-sources/src/data_provider/src/osinfo/sysOsParsers.h:16,
13:19:19                   from /home/jfzem/hp-ux/wazuh-sources/src/data_provider/src/osinfo/sysOsParsers.cpp:12:
13:19:19  /home/jfzem/hp-ux/wazuh-sources/src/external/nlohmann/json.hpp: In member function ‘void nlohmann::json_abi_v3_11_2::detail::binary_writer<BasicJsonType, CharType>::write_cbor(const BasicJsonType&)’:
13:19:19  /home/jfzem/hp-ux/wazuh-sources/src/external/nlohmann/json.hpp:15118:26: error: ‘isnan’ is not a member of ‘std’; did you mean ‘isnan’?
13:19:19  15118 |                 if (std::isnan(j.m_value.number_float))
13:19:19        |                          ^~~~~
13:19:19  In file included from /home/okkam/bootstrap-gcc/gcc94_prefix/include/c++/cmath:45,
13:19:19                   from /home/okkam/bootstrap-gcc/gcc94_prefix/include/c++/valarray:37,
13:19:19                   from /home/jfzem/hp-ux/wazuh-sources/src/external/nlohmann/json.hpp:171,
13:19:19                   from /home/jfzem/hp-ux/wazuh-sources/src/data_provider/src/osinfo/sysOsParsers.h:16,
[...]

PR #19027, opened 2023-09-15
nlohmann cmath library incompatibility fixed on HP-UX

matias-braida:
|Related issue|
|---|
|#18882|

## Description

This PR fixes compilation on HP-UX systems.
The implementation modifies the version of nlohmann dep if the requested version is greater than the max version number (21).

## Tests

The version on nlohmann dep is limited on deps downloading:
image

The compilation on HP-UX now ends correctly:
image

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -1214,7 +1214,8 @@ GUNZIP := gunzip
 GZIP := gzip
 CURL := curl -so
 DEPS_VERSION = 22
-RESOURCES_URL := https://packages.wazuh.com/deps/$(DEPS_VERSION)
+RESOURCES_URL_BASE := https://packages.wazuh.com/deps/
+RESOURCES_URL := $(RESOURCES_URL_BASE)$(DEPS_VERSION)
 CPYTHON := cpython
 PYTHON_SOURCE := no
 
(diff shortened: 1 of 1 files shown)


GabrielEValenzuela (approved):
LGTM!

src/Makefile
```diff
@@ -1325,6 +1326,39 @@ external/$(CPYTHON).tar.gz:
 	rm $(patsubst %.gz,%,$@)
 endif
 
+ifeq (${uname_S},HP-UX)
+# nlohmann library
+RESOURCES_URL_BASE_NLOHMANN=$(subst https://,http://,$(RESOURCES_URL_BASE))
+ifeq ($(findstring $(RESOURCES_URL_BASE_NLOHMANN),$(RESOURCES_URL)),$(RESOURCES_URL_BASE_NLOHMANN))
```
pereyra-m, line 1332: Just a detail, no changes are requested: if the `RESOURCES_URL` variable is externally set with `https` instead of `http`, the mechanism of setting the **nlohmann** library to version 21 won't be applied.

But this doesn't seem a problem, because in HP-UX, we always should define it with `http`.

pereyra-m (approved):
LGTM!

Dwordcito (approved):

Issue #19021, opened 2023-09-15
Engine - Update Syscollector Decoder to Enhance Host Information Processing


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0 | Wazuh Engine | Manager | Packages/Sources | any |

**Objective**:
The syscollector decoder in the Wazuh engine requires an update to address issues stemming from recent changes in the syscollector component. These changes have disrupted the correct processing of messages within the engine.

**Problem Statement**:
Currently, due to the recent modifications in syscollector, the engine is unable to process certain messages effectively. As a result, the engine fails to detect and capture vital host information when it's present in an event, preventing it from being saved in the agent information Key-Value Database (kvdb). The absence of data in the kvdb leads to a deficiency in enriching events with crucial host-related information, rendering our decoded events incomplete.

**Proposed Solution**:
To rectify this issue, we need to update both the syscollector and dbsync decoders. These updates should enable the decoders to once again capture host information and enrich events with the requisite data. By doing so, we can ensure that our decoded events are comprehensive and provide valuable insights into the security and operational aspects of the system.

**Scope of Work**:
- Review and analyze the recent changes in syscollector to identify the specific areas affecting message processing.
- Modify the syscollector decoder codebase to accommodate the changes and enable the correct processing of messages.
- Update the dbsync decoder to ensure synchronization with the syscollector updates.
- Verify that host information is being captured and saved in the kvdb as expected.
- Test the updated decoder thoroughly to ensure it functions correctly under various scenarios.
- Document the changes made to the syscollector and dbsync decoders for reference and transparency.

**Additional Information**:
- This issue is critical as it impacts the completeness and accuracy of event data processed by the Wazuh engine.
- The update should maintain backward compatibility with existing configurations and deployments.

PR #19038, opened 2023-09-18
Updated dbsync decoder to correctly treat spaces

JavierBejMen:
|Related issue|
|---|
|#19021|

Decoder dbsync now correctly parses spaces in helper concat when creating WDB query, effectively mapping host ECS fields on events coming from agents.

Example from fresh Vagrant with an agent:
image


JcabreraC (approved):
LGTM !

Issue #17912, opened 2023-07-13
Implement a read interface to get vulnerability candidates


Dwordcito:
## Description

This issue focuses on implementing a read interface to retrieve data from the vulnerability manager feed. The vulnerability manager feed contains the stored vulnerability data that needs to be accessed by other components or systems. To enable seamless access to this data, a read interface should be implemented to query and retrieve relevant vulnerability information.

## Expected

After completing this issue, the following outcomes are expected:

- Implement a read interface in the vulnerability feed manager to get vulnerability information.
- A well-defined set of methods or functions that allow users to query and retrieve vulnerability data.
- Proper handling of data retrieval requests, including error handling and response formatting.
- Comprehensive testing of the read interface to ensure its functionality and reliability.

Interface:
void getVulnerabilitiesCandidates(const std::string& cnaName, const ScanContext& context, const std::function<bool(const ScanVulnerabilityCandidate& data)>& fnc)

Schema flat buffer:
```
namespace CVE;

enum Status: byte {
 affected = 0,  
 unaffected = 1,
 unknown = 2
}

table Version {
  status: Status;
  version: string;
  lessThan: string;
  versionType: string;
}

table ScanInfo {
  cveId: string;
  defaultStatus: Status;
  platforms: [string];
  versions: [Version];
}

root_type ScanVulnerabilityCandidate;
```

#### Database information
Key: PackageName_CVEID
Value: FlatBuffer
Search/select-> PackageName_*

## DoD (Definition of Done)

To consider this issue completed, the following tasks should be accomplished:

- [x] Implement the necessary methods or functions to retrieve vulnerability data.
- [x] Perform thorough testing to ensure the correctness and reliability of the read interface.
- [x] Handle any potential errors or exceptions during data retrieval.

PR #19059, opened 2023-09-19
Dev 17912 vulnerabilities read interface

MiguelazoDS:
|Related issue|
|---|
|#17912|

## Description

Read interface for vulnerability candidate in database feed manager.


.gitignore
```diff
@@ -159,3 +159,4 @@ src/wazuh_modules/vulnerability_detector/msu.json.gz
 # Flatbuffers generated headers.
 src/shared_modules/utils/flatbuffers/include/
 src/wazuh_modules/syscollector/include/
+src/wazuh_modules/vulnerability_scanner/include/*_generated.h
```
GabrielEValenzuela, line 162: I think regex is not supported fully in `.gitignore`, probably have to add each file to skip them from the VCS.

src/wazuh_modules/vulnerability_scanner/CMakeLists.txt
```diff
@@ -68,3 +71,5 @@ else()
   #add_subdirectory(testtool)
 endif(UNIT_TEST)
 
+#add_subdirectory(testtool)
+#add_subdirectory(tests)
```
GabrielEValenzuela, line 75: Please, check these lines.

src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
```diff
@@ -30,6 +31,7 @@
 
 constexpr auto DESCRIPTION_DATABASE_PATH {"queue/vd/descriptions"};
 constexpr auto REMEDIATIONS_DATABASE_PATH {"queue/vd/remediations"};
+constexpr auto ROCKS_DB_PATH {"./test.db"};
```
GabrielEValenzuela, line 34: Will this database used in production?

src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
```diff
@@ -72,16 +74,61 @@ class TDatabaseFeedManager final : public Observer<nlohmann::json&>
...
             });
     }
+    /**
+     * @brief Get the Vulnerabilities Candidates information.
+     *
+     * @param cnaName RocksDB table identifier.
+     * @param context Scan context structure.
+     * @param callback Store vulnerability data.
+     */
+    void getVulnerabilitiesCandidates(const std::string& cnaName,
```
GabrielEValenzuela, line 93: What is the error flow of this function? If an exception is thrown, how notify which error causes the caller?

src/wazuh_modules/vulnerability_scanner/tests/databaseFeedManager/databaseFeedManager_tests.cpp
```diff
@@ -0,0 +1,197 @@
+/*
+ * Wazuh Vulnerability scanner
```
GabrielEValenzuela, line 2: Please, correct this

src/wazuh_modules/vulnerability_scanner/tests/databaseFeedManager/databaseFeedManager_tests.hpp
```diff
@@ -0,0 +1,110 @@
+/*
+ * Wazuh Vulnerability scanner
```
GabrielEValenzuela, line 2: Please, correct this.

src/wazuh_modules/vulnerability_scanner/tests/databaseFeedManager/databaseFeedManager_traps.hpp
```diff
@@ -0,0 +1,392 @@
+/*
+ * Wazuh Vulnerability scanner
```
GabrielEValenzuela, line 2: Please, correct this.

src/wazuh_modules/vulnerability_scanner/tests/databaseFeedManager/main.cpp
```diff
@@ -0,0 +1,18 @@
+/*
+ * Wazuh Vulnerability scanner
```
GabrielEValenzuela, line 2: Please correct this

src/wazuh_modules/vulnerability_scanner/schemas/scanVulnerabilityCandidate.fbs
```diff
@@ -0,0 +1,23 @@
+namespace NSVulnerabilityScanner;
```
GabrielEValenzuela, line 1: Probably, we can use all the same namespace to avoid multiple `using...`

GabrielEValenzuela (changes requested):
Great job! I leave some comments and questions 🙂

.gitignore
MiguelazoDS: It does work, it also used some lines above. This is the outcome 
image

src/wazuh_modules/vulnerability_scanner/tests/databaseFeedManager/main.cpp
```diff
@@ -0,0 +1,18 @@
...
+ * Foundation.
+ */
+
+#include "gtest/gtest.h"
+
+int main(int argc, char** argv)
+{
+    ::testing::InitGoogleTest(&argc, argv);
+    return RUN_ALL_TESTS();
+}
```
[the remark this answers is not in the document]
MiguelazoDS: Added

src/wazuh_modules/vulnerability_scanner/tests/databaseFeedManager/main.cpp
MiguelazoDS: Done

src/wazuh_modules/vulnerability_scanner/tests/databaseFeedManager/databaseFeedManager_traps.hpp
MiguelazoDS: Done

src/wazuh_modules/vulnerability_scanner/tests/databaseFeedManager/databaseFeedManager_tests.hpp
MiguelazoDS: Done

src/wazuh_modules/vulnerability_scanner/tests/databaseFeedManager/databaseFeedManager_tests.cpp
MiguelazoDS: Done

src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
MiguelazoDS: I tried to respect as much as possible the definition mentioned in the issue. I think the best way to do this is by logging the errors, warnings, and info messages. Throwing exceptions in that function will break the loop.

src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
MiguelazoDS: I'm not sure it was not specified in this issue, but I can change it to use the same path used above.

src/wazuh_modules/vulnerability_scanner/CMakeLists.txt
MiguelazoDS: Good catch, probably some rebase caused that. Removed those lines.

src/wazuh_modules/vulnerability_scanner/CMakeLists.txt
```diff
@@ -37,6 +39,7 @@ link_directories(${SHARED_MODULES}/indexer_connector/build)
 
 add_subdirectory("src/scanOrchestrator")
 add_subdirectory("src/databaseFeedManager")
+add_subdirectory(tests)
```
matias-braida, line 42: I think tests subdirectory should only be added when compilation define "UNIT_TEST" is set.
In my similar PR use this piece of code:
```
if(UNIT_TEST)
  target_link_libraries(vulnerability_scanner scan_orchestrator database_feed gcc_s)
  if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
    target_link_libraries(vulnerability_scanner -fprofile-arcs)
  else()
    target_link_libraries(vulnerability_scanner gcov)
  endif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")

  add_subdirectory(tests)
else()
  target_link_libraries(vulnerability_scanner scan_orchestrator database_feed router content_manager indexer_connector gcc_s)
  if(FSANITIZE)
    set(CMAKE_CXX_FLAGS_DEBUG "-g -fsanitize=address,leak,undefined")
  endif(FSANITIZE)
  #add_subdirectory(testtool)
endif(UNIT_TEST)
```

src/wazuh_modules/vulnerability_scanner/schemas/scanVulnerabilityCandidate.fbs
matias-braida: The issue I found when I tried to compile my fbs was that the namespace had the same name that the table. So, later in the code compilation they had names miss understanding.
At this point I just added "NS" at the start of the namespace just to solve that compilation naming problem.
But, if we all use the same namespace for me is a right choice.

GabrielEValenzuela: We agree to use all the same namespace.

src/wazuh_modules/vulnerability_scanner/CMakeLists.txt
MiguelazoDS: You're right as we discussed earlier.

src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
GabrielEValenzuela: What do think about refactoring this function in this way:

```cpp
        try
        {
            Utils::RocksDBWrapper rocksDBWrapper(ROCKS_DB_PATH);
            if (!context.contextData.empty())
            {
                try
                {
                    packageName = context.contextData.at("data").at("name");
                    for (const auto& [key, value] : rocksDBWrapper.seek(packageName))
                    {
                        if (callback(reinterpret_cast<const char*>(value.data()), value.size()))
                        {
                            // Logging
                        }
                    }
                }
                catch (...)
                {
                        //Loggin
                        std::cerr<<"Invalid package\n";
                }
            }
        }
        catch (...)
        {
                        //Loggin
                        std::cerr<<"Unable to open database\n";
        }
    }
```

src/wazuh_modules/vulnerability_scanner/tests/databaseFeedManager/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+
+project(database_feed_manager_tests)
+
+include_directories(${SRC_FOLDER}/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/)
+include_directories(${SRC_FOLDER}/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/)
+include_directories(${SRC_FOLDER}/shared_modules/)
+include_directories(${SRC_FOLDER}/shared_modules/content_manager/src/)
+include_directories(${SRC_FOLDER}/wazuh_modules/vulnerability_scanner/tests/mocks/)
+
+file(COPY ${SRC_FOLDER}/wazuh_modules/vulnerability_scanner/schemas/scanVulnerabilityCandidate.fbs DESTINATION
```
GabrielEValenzuela, line 11: You can avoid the `COPY` operation using a `target_compile_definitions` which specifies compile definitions to use when compiling a given <target>. The named <target> must have been created by a command such as [add_executable()](https://cmake.org/cmake/help/latest/command/add_executable.html#command:add_executable) or [add_library()](https://cmake.org/cmake/help/latest/command/add_library.html#command:add_library) and must not be an [ALIAS target](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#alias-targets).

```cmake
target_compile_definitions(${PROJECT_NAME} PUBLIC SCHEMA_ROOT_PATH="${SRC_FOLDER}/wazuh_modules/vulnerability_scanner/schemas/")
```

And then in your `databaseFeddManager_test.hpp`

https://github.com/wazuh/wazuh/blob/460943386c69d6f6da80e790e82c3c5a299088c9/src/wazuh_modules/vulnerability_scanner/tests/databaseFeedManager/databaseFeedManager_tests.hpp#L35

Put 

```cpp
const std::string flatbufferSchema {SCHEMA_ROOT_PATH "scanVulnerabilityCandidate.fbs"};
```

matias-braida:
GJ! Tested locally with successful results
image

matias-braida (approved):

GabrielEValenzuela (approved):
LGTM!

Issue #15994, opened 2023-01-24
Relative path support in GET /rules/files/{filename} and GET /decoders/files/{filename} endpoints


davidjiglesias:
|Wazuh version|Component|
|---|---|
| >4.1.0 | Wazuh API |

## Description
We currently do not have the option to the path of the file when reading the content of a decoder file or rule file.

We need to add a new parameter to the API endpoint `GET /rules/files/{filename}` and `GET /decoders/files/{filename}` to allow it. 

The parameter name is `relative_dirname`, and in this case, its default value would be `None`.

## Checks

The following elements have been updated or reviewed (should also be checked if no modification is required):
- [x] Tests (unit tests, API integration tests).
- [x] Changelog.
- [x] Documentation.
- [x] Integration test mapping (using `api/test/integration/mapping/_test_mapping.py`).

vicferpoy:
## Issue update

This implementation must be done to `PUT /decoders/files/{filename}` as well.

davidjiglesias:
After discussion with frontend team: https://github.com/wazuh/wazuh-kibana-app/issues/5176#issuecomment-1441481894 we have to change the approach for this issue.

davidjiglesias:
We need to remove the changes from https://github.com/wazuh/wazuh/pull/16033 also, and documentation.

mhamra:
### UPDATE
Created git branch [fix/15994-revert-pr-16033](https://github.com/wazuh/wazuh/tree/fix/15994-revert-pr-16033). I've closed prior PRs.

mhamra:
### UPDATE

Started the required implementation. The APIs in spec.yaml will have a new optional string parameter called custom_ruleset_path:
- /rules/files/{filename}:
- /decoders/files/{filename}:

The following api/api/controllers will add this new parameter:
- decoder_controller.py
- rule_controller.py

The following framework/wazuh functions will add this new optional parameter and will implement the logic to get the files:
- decoder.get_decoder_file
- rule.get_rule_file

mhamra:
### UPDATE
- Increase the number of unit tests by  parameterizing the test_get_decoder_file and test_get_rule_file test functions
- It is needed to define if the functions get_rule_file and get_decoder_file could return more than one result as in the current implementation
- It is needed to review if it is possible to use the file_name parameter to pass the relative path instead of changing the parameter custom_ruleset_path

mhamra:
### UPDATE
[...]

PR #19210, opened 2023-09-25
Fixes overwrite when modifying CDB List

RamosFe:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/19419|

## Description
Fixes error when using `overwrite` for the CDB List and adds the `relative_dirname` to the `cdb_list` endpoints.

### DIFF
--- a/api/api/controllers/cdb_list_controller.py
+++ b/api/api/controllers/cdb_list_controller.py
@@ -126,7 +126,7 @@ async def get_file(request, pretty: bool = False, wait_for_complete: bool = Fals
 
 
 async def put_file(request, body: dict, overwrite: bool = False, pretty: bool = False, wait_for_complete: bool = False,
-                   filename: str = None) -> web.Response:
+                   filename: str = None, relative_dirname: str = None) -> web.Response:
     """Upload content of CDB list file.
 
     Parameters
@@ -142,6 +142,8 @@ async def put_file(request, body: dict, overwrite: bool = False, pretty: bool = 
         If set to false, an exception will be raised when updating contents of an already existing filename.
     filename : str
         Name of the new CDB list file.
+    relative_dirname : str
+        Relative directory where the file is located.
 
     Returns
     -------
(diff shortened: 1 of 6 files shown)


EduLeon12 (dismissed):
LGTM !

Selutario (changes requested):
## More changes needed

Currently, the [GET /lists/files/{filename}](https://documentation.wazuh.com/current/user-manual/api/reference.html#operation/api.controllers.cdb_list_controller.get_file) endpoint only supports the file name as a parameter and not a path (absolute or relative) to the file.

Since there can (and are) subfolders with lists, the API searches all subfolders for that filename and returns it. This is the reason why it currently prevents uploading files if another file with the same name already exists: doing so would not allow you to access the original file in the subfolder.

For example, running `GET /lists/files/aws-eventnames` would clearly return this marked file:
```
/var/ossec/etc/lists/
|-- amazon
|   |-- aws-eventnames <---------- THIS WOULD BE RETURNED
|   |-- aws-eventnames.cdb
|   `-- aws-sources
|-- audit-keys
|-- audit-keys.cdb
|-- security-eventchannel
`-- security-eventchannel.cdb
```

However, after changes in this PR, it would be possible to upload a new file with the same name (`GET /lists/files/aws-eventnames`), always in the root lists folder:
```
/var/ossec/etc/lists/
|-- amazon
|   |-- aws-eventnames <---------- ORIGINAL FILE
|   |-- aws-eventnames.cdb
|   `-- aws-sources
|-- audit-keys
|-- audit-keys.cdb
|-- aws-eventnames <-------------- NEW UPLOADED FILE
|-- security-eventchannel
`-- security-eventchannel.cdb
```

So now, when running `GET /lists/files/aws-eventnames`, the file in the subfolder would never be returned:
```
/var/ossec/etc/lists/
|-- amazon
|   |-- aws-eventnames <---------- NO ACCESSIBLE THROUGH THE API
|   |-- aws-eventnames.cdb
|   `-- aws-sources
|-- audit-keys
|-- audit-keys.cdb
|-- aws-eventnames <-------------- THIS WOULD BE RETURNED
|-- security-eventchannel
`-- security-eventchannel.cdb
```

### Conclusion

A new `relative_dirname` parameter should be added to required methods and endpoints, same as here:
- https://github.com/wazuh/wazuh/issues/15994

## Base branch
It will involve breaking changes, so it must be merged into 4.8.0.

Issue #18527, opened 2023-08-21
Engine - Use the agent simulator in test CLI command


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0 | Engine | Manager| - | - |

## Description

Currently, the engine has an agent simulator for testing on it. Reference: [#6695](https://github.com/wazuh/wazuh/issues/16695)

It is necessary to integrate the agent simulator with the test command, so that the end user can send an event, just as it takes it from the source (windows event, for example), without having to know all the necessary parameters of the test command of the engine.

### Considerations

The user who uses the agent simulator will be able to:

- Use of events in the original format of the source (example, eventlog, eventchannel)

- Being able to enter a multiline event

## Tasks

- [x] Research of events
- [x] Adapt the agent simulator to the requirements
- [x] Try different types of events
- [ ] Update documentation

soynof:
## List of sources

<b style="font-weight:normal;" id="docs-internal-guid-9d93291b-7fff-89f1-ec9f-391ccf4fa931"><div dir="ltr" style="margin-left:0pt;" align="left">

Module | sources | arguments | queue | example
-- | -- | -- | -- | --
logcollector | audit | Format, label | 49 |  
| | command  |   | 49 |  
| | djb-multilog | | 49 |
| | eventchannel | | 102 |
| | eventlog | | 49 |
| | full_command | | 49 |
| | iis | | 49 |
| | json | | 49 |
| | macos | | 49 |
| | multi-line-regex | | 49 |
| | multi-line | | 49 |
| | mysql_log | | 49 |
| | nmapg | | 49 |
| | postgresql_log | | 49 |
| | snort-full | | 49 |
| | squid | | 49 |
| | syslog | | 49 |
wodles | Github |   | WIP |  
| | Office 365 |   | WIP |  
| | AWS-s3 |   | 49 |  
| | OpenSCAP |   | WIP |  
| | Cis-cat |   | WIP |  
| | Osquery |   | WIP |  
| | Virustotal |   | WIP |  
| | Docker-listener |   | WIP |  
| | Azure |   | WIP |  
Internal modules | DB sync |   | 53 |  
| | hostinfo |   | 51 |  
| | rootcheck |   | 57 |  
| | sca |   | 112 |  
| | syscheck |   | 56 |  
| | syscollector |   | 100 |  
| | upgrade |   | 117 |  
Remote syslog |   |   |   |  

</div></b>

[...]

PR #19269, opened 2023-09-27
Engine improve indexed fields

JcabreraC:
There is a need to enhance and fix the Windows decoder in the Wazuh project. This set of changes aims to address various identified issues in the Windows decoder. The proposed fixes and improvements include:

- [x] **Removal of Unnecessary Single Quotes**: Unnecessary single quotes will be removed in the `windows-security.yml` file at lines 38-39, 46, and 104.

- [x] **Review/Removal of TODO Comments**: TODO comments at lines 37, 45, 333, and 334 will be reviewed and removed as appropriate.

- [x] **Empty Fields in Indexing**: The issue of empty fields being indexed in Windows security events will be addressed. While some of these fields may be introduced by Filebeat and are beyond the scope of this fix, we will investigate and document the source of these empty fields. Measures will be taken to prevent them from appearing when empty if feasible.

- [x] **Removal of the "fields" Field**: The field fields is added by filebeat, but cannot be removed with a drop_fields processor as it is added at a later stage. More research is needed to see if it can be dropped or prevented by a filebeat configuration setting. Currently, in the engine, we do not use any filebeat module, as opposed to the current Wazuh implementation where it is dropped.

- [x] **Missing "message" Field in Windows Events**: A solution will be sought to add the "message" field to Windows events, similar to the event description in ECS. This may involve translating Windows event IDs into human-readable descriptions using KVDBs.

- [x] **Stop Using Numbers for Queues**: An associated improvement is already in progress, and the related issue can be found here: #18527

### DIFF
--- a/src/engine/extension/filebeat/7.x/filebeat.yml
+++ b/src/engine/extension/filebeat/7.x/filebeat.yml
@@ -61,9 +61,10 @@ processors:
       when:
         equals:
           log.file.path: "/var/ossec/logs/alerts/alerts-ECS.json"
-      fields: ["log.offset", "log.file.path"]
+      fields: ["log.offset", "log.file"]
+        
   - drop_fields:
-      fields: ["agent.ephemeral_id", "agent.hostname", "agent.version", "input.type"]
+      fields: ["agent.ephemeral_id", "agent.hostname", "agent.version", "input"]
   - drop_fields:
       when:
         not:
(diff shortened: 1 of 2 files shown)


JcabreraC:
LGTM !

NahuFigueroa97 (approved):
LGTM

PR #19452, opened 2023-10-06
Merge 4.5.3 into 4.6.0

JcabreraC:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/19447|

This PR aims to merge branch 4.5.3 into 4.6.0.


Selutario (dismissed):

Selutario (approved):

src/win32/qa/test_win_upgrade.py
```diff
@@ -0,0 +1,77 @@
...
+    installed_files_dict = {}
+    populate_dict(installed_files_dict, installed_files)
+
+    assert len(installed_files_dict) == len(files_to_install_dict), f"Installed files: '{installed_files_dict}'\nFiles to install: '{files_to_install_dict}'"
+
+    success = True
+    failed_keys = []
+    for key in files_to_install_dict:
+        # Compare hashes
+        if installed_files_dict[key] != files_to_install_dict[key]:
```
MarcelKemp, line 72: The following check is failing because `files_to_install_dict` has the new `key` (in this case _`fimdb.dll`_), but `installed_files_dict` does not have that `key`, so it shows the error:
- [check_upgrade_result (4.5.2-1)](https://github.com/wazuh/wazuh/actions/runs/6429974259/job/17460167109?pr=19452#logs)


To fix this, it is necessary to just check the **`key`** that exist in both `installed_files_dict` and `files_to_install_dict` for this condition.

All other cases, such as a newly installed or removed _`.dll`_, should be considered in additional conditions.

Possible workaround to avoid the condition failure:
- Check before the **`if`**, if the `key` exists in `installed_files_dict`.

MarcelKemp (changes requested):

MarcelKemp (approved):
The above problem should be fixed in the next issue:
- https://github.com/wazuh/wazuh/issues/19453

JcabreraC:
Failure of the known fim test: https://github.com/wazuh/wazuh-qa/issues/4514

PR #19461, opened 2023-10-06
Updating Windows tests with new libfimdb.dll

### LINKED ISSUE #19453: Fix windows update test
|---|---|---|---|---|
| 4.6.0-rev | Wazuh component | Agent | Packages/Sources | windows |

## Description
In the Windows upgrade test ([test_win_upgrade.py](https://github.com/wazuh/wazuh/blob/4.5.3/src/win32/qa/test_win_upgrade.py)), we have identified an issue with the code containing the error:

```python
for key in files_to_install_dict:
    # Compare hashes
    if installed_files_dict[key] != files_to_install_dict[key]:
```

The problem arises when `files_to_install_dict` contains a new key (e.g., fimdb.dll), but `installed_files_dict` does not have that key, resulting in an error condition. This issue is causing the following check to fail:

[Check Upgrade Result (4.5.2-1)](https://github.com/wazuh/wazuh/actions/runs/6429974259/job/17460167109?pr=19452#logs)

This test has provoked an exception, and we should account for such cases. Specifically, the test has uncovered a situation where `files_to_install_dict` contains keys that are not present in `installed_files_dict`.

## Proposed Solution
To address this, we need to modify the code to only check keys that exist in both `installed_files_dict` and `files_to_install_dict` for the given condition. In other words, we should avoid comparing keys that are present in `files_to_install_dict` but not in `installed_files_dict`.

Additionally, we should consider handling other cases, such as newly installed or removed .dll files, with additional conditions.

pereyra-m:
|Related issue|
|---|
|Closes #19453|

## Description

This PR updates all required tests with the new incorporation of `libfimdb.dll` in the Windows agent.
It also adds the corresponding file details to the library.

  - [x] Windows

### DIFF
--- a/.github/workflows/windows-file-details.yml
+++ b/.github/workflows/windows-file-details.yml
@@ -25,7 +25,11 @@ jobs:
           # We remove the executables that aren't distributed in the signed installer
           rm -f src/win32/setup-*.exe
           cp src/win32/*.exe src/bin_files/
-          cp src/*.dll src/bin_files/
+          # We exclude the libraries that aren't built by Wazuh
+          rm -f src/win32/libwinpthread-1.dll
+          rm -f src/win32/libgcc_s_dw2-1.dll
+          rm -f src/win32/libstdc++-6.dll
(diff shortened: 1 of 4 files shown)


Dwordcito (approved):
LGTM.

PR #19593, opened 2023-10-11
Add support for the Server Name Indication TLS extension during automatic enroll

dmaasland:
|Related issue|
|---|
|None|

## Description
This PR adds support for sending the SNI extension when doing automatic enroll. This allows the server side to use (for example) reverse proxies to route the request to the correct backend location

## Configuration options
None

## Logs/Alerts example
None

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

### DIFF
--- a/src/shared/enrollment_op.c
+++ b/src/shared/enrollment_op.c
@@ -244,7 +244,6 @@ static int w_enrollment_connect(w_enrollment_ctx *cfg, const char * server_addre
     if (1 != SSL_set_tlsext_host_name(cfg->ssl, server_address)) {
         mwarn("Unable to set SNI hostname: %s", server_address);
     }
-
     
     ERR_clear_error();
     int ret = SSL_connect(cfg->ssl);
--- a/src/shared/enrollment_op.c
+++ b/src/shared/enrollment_op.c
@@ -241,13 +241,10 @@ static int w_enrollment_connect(w_enrollment_ctx *cfg, const char * server_addre
     SSL_set_bio(cfg->ssl, sbio, sbio);
 
     /* Do SNI */
-    if (!strcmp(ip_address, server_address)) {
-        if (1 != SSL_set_tlsext_host_name(cfg->ssl, server_address)) {
-            merror("Unable to set SNI hostname: %s", server_address);
-        } else {
-            minfo("SNI Hostname set to: %s", server_address);
-        }
+    if (1 != SSL_set_tlsext_host_name(cfg->ssl, server_address)) {
+        merror("Unable to set SNI hostname: %s", server_address);
     }
+
     
     ERR_clear_error();
     int ret = SSL_connect(cfg->ssl);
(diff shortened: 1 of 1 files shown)


vikman90 (changes requested):
@dmaasland Thank you for proposing this change. We need some changes before merging it:

## An issue is required

We need an issue explaining the context and the requirement, for us to understand why it's required and if this is a bug or an enhancement. Additionally, this allows us to manage it in a project and prioritize it correctly.

**Every PR must be related to an existing issue.**

## Unsafe call to `strtok()`

The function `strtok()` is **thread-unsafe** as it stores the remaining string pointer in global memory. If the current thread calls `strtok()` just after another thread, we would be overwriting its memory which could cause undefined behavior.

On the other hand, `strtok()` accepts `char *` (no `const`) as it may modify the input string (replace `/` with `\0`).

### Compiler warning

```
shared/enrollment_op.c: In function ‘w_enrollment_connect’:
shared/enrollment_op.c:205:33: warning: passing argument 1 of ‘strtok’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  205 |         server_address = strtok(server_address, "/");
      |                                 ^~~~~~~~~~~~~~
In file included from ./headers/shared.h:67,
                 from ./os_auth/../addagent/manage_agents.h:14,
                 from ./os_auth/auth.h:33,
                 from shared/enrollment_op.c:11:
/usr/include/string.h:356:39: note: expected ‘char * restrict’ but argument is of type ‘const char *’
  356 | extern char *strtok (char *__restrict __s, const char *__restrict __delim)
      |                      ~~~~~~~~~~~~~~~~~^~~
```

### Segmentation fault

In normal cases, the bug won't appear as we call `w_enrollment_request_key()` with a dynamic string stored in the heap (`const` is not enforced). But, indeed, if we call `w_enrollment_request_key()` with a hardcoded string:

```
2023/11/15 10:34:21 wazuh-agentd: INFO: Started (pid: 9709).
2023/11/15 10:34:21 wazuh-agentd: INFO: Requesting a key from server: 172.28.54.78
2023/11/15 10:34:21 wazuh-agentd: INFO: EPS limit enabled, EPS: '500', timeframe: '1'
Segmentation fault (core dumped)
```

## Testing is required

- Every modification of the source code requires **unit test coverage** (_src/unit_tests/shared/test_enrollment_op.c_).
- As this is a use case extension, we must update the **integration tests**, located at https://github.com/wazuh/wazuh-qa/tree/master/tests/integration/test_agentd.

Thank you.
Best regards.

Issue #14153, opened 2022-07-06
Vulnerability Detector refactor


jnasselle:
## Description

Vulnerability Detector module is one of the key components of Wazuh, providing crucial information about exploitable threats on endpoints, based on information (CVEs feeds) retrieved from a variety of vendors

#### Vulnerabilities feeds
- The information retrieved from different vendors (Linux distribution feed, NVD, Microsoft via Wazuh MSU) may take a long time to be fetched, ingested, and inserted into a database
- There are so many configurations to tweak feeds that end up confusing the user
- CVEs information is not 100% trustable. Format inconsistency, missing fields, duplicates, and other reasons lead to over-hardening the vulnerability scanner and hardcoding hundred of business rules, making the product so rigid.
- Sometimes there are problems with feed availability
- There's no place to add customizations from Wazuh side
#### Scanning performance
- Current design throughput in environments with hundreds/thousands of agents
#### Results visualization
- Results are scattered throughout the entire Wazuh Cluster and is expensive to retrieve them all at the same time by Wazuh API to provide global dashboards

### Proposal
- Sanitize third-party vulnerabilities information and deliver it through the Wazuh Content model
- Reduce the Vulnerability Detector time-to-value by 
  - Fetching Wazuh Content deltas if possible
  - Provide OOTB Wazuh Content baseline at installation time
- Improve UX by reducing Vulnerability Detector configuration options.
- Simplify vulnerability scan by removing current complexity handling from code to Wazuh Content generation step.
- Enhance the vulnerability scanner to avoid unnecessary scans by changing to an event-driven approach.
- Push Vulnerability detector results to Wazuh Indexer

### Software development tools
- GCC 9.4
- CMake 3.12.4
- GTests 1.10 framework for component and unit tests.
- python3 for integration.

### Software verification tools
- Valgrind 3.18.1
- lcov 1.14
- Address Sanitizer Included in gcc 9.4
- CPP Check 2.8.2
- clang tools 14 (scan-build)
- clang-format 14 (style compliance)

## Derived epics  
- wazuh/wazuh#14155
- wazuh/wazuh#14156 

### Shared components
- wazuh/wazuh#17894
- wazuh/wazuh#17895

## Framework
- https://github.com/wazuh/wazuh/issues/18284

## UI
- https://github.com/wazuh/wazuh-kibana-app/issues/5763

## QA
- https://github.com/wazuh/wazuh-qa/issues/4369

## Border
[...]

PR #19708, opened 2023-10-17
Add `index.query.default_field`

AlexRuiz7:
|Related issue|
|---|
| #14153, https://github.com/wazuh/wazuh-indexer/issues/6|

## Description

This PR adds default query fields to the vulnerability detector index (`wazuh-states-vulnerability`).

Use this command to upload the template (change the name of the template as desired):

```bash
curl -u admin:admin -k -X PUT "https://localhost:9200/_index_template/wazuh-vulnerability-detector" -H "Content-Type: application/json" -d @legacy-template.json
```

image
image
image

### DIFF
--- a/src/wazuh_modules/vulnerability_scanner/indexer/template/legacy-template.json
+++ b/src/wazuh_modules/vulnerability_scanner/indexer/template/legacy-template.json
@@ -1,9 +1,38 @@
 {
   "index_patterns": [
-    "vulnerability-state-*"
+    "wazuh-states-vulnerabilities"
   ],
   "priority": 1,
   "template": {
+    "settings": {
+      "index": {
+        "codec": "best_compression",
+        "mapping": {
+          "total_fields": {
+            "limit": 1000
+          }
+        },
+        "number_of_replicas": "0",
+        "number_of_shards": "1",
+        "query.default_field": [
+          "base.tags",
+          "agent.id",
+          "ecs.version",
+          "event.id",
+          "event.module",
+          "event.severity",
+          "host.os.family",
+          "host.os.full.text",
+          "host.os.version",
+          "package.name",
+          "package.version",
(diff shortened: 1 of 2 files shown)


src/wazuh_modules/vulnerability_scanner/indexer/template/legacy-template.json
Dwordcito, line 1: I understand that we should do a check with github actions similar to this:
https://github.com/wazuh/wazuh/blob/master/.github/workflows/wazuh-template-elasticsearch.yml

Apart from this, I see that some fields like event.severity have longs, which causes problems due to the type in the searches, since yes or yes they must be strings (according to my experience in Wazuh, you will show me xD)

base.tags exists? does it have properties?

Dwordcito (changes requested):

src/wazuh_modules/vulnerability_scanner/indexer/template/legacy-template.json
AlexRuiz7: Good point on creating a test. I'll give that a try.

I did some manual tests and having long as default query field (like `event.severity`) didn't raise any problem.

As you can see in the image below, no field was specified for the search.

image


However, if this has been problematic before, I'd avoid having these types in as default query fields. Please, feel free to suggest any [field](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html) you see fit.

Finally, yes, [base.tags](https://www.elastic.co/guide/en/ecs/current/ecs-base.html#field-tags) does exist. It's a list of keywords. Maybe [base.message](https://www.elastic.co/guide/en/ecs/current/ecs-base.html#field-message) could be used instead, as it is of type text*, which is analyzed and inserted into the inverted index.

*It's actual type is `match_only_text`, but as OpenSearch does not support this data type, we transform it to `text`.

Issue #19739, opened 2023-10-19
Utils - Add `gz`, decompress helper


Damian-Mangold:
|Component|
|---|
| [shared_modules/utils](https://github.com/wazuh/wazuh/tree/dev-14153-vulndet-refactor/src/shared_modules/utils) |

## Description

This issue aims to work on the incorporation of a helper that allows decompressing files with `gz` format.

## Expected

After this issue is completed, we should expect the following:
- A helper function to simply decompress `gz` files.

## DoD

- [x] Issue has been fixed.
- [ ] Add new tests to verify that the described problem no longer occurs.
- [x] Add or Update Unit and Component.
- [ ] Compliance with:
  - [x] [RTR](https://github.com/wazuh/wazuh/blob/dev-14153-vulndet-refactor/src/wazuh_modules/vulnerability_scanner/doc/rtr.md)
  - [ ] [GitHub Actions](https://github.com/wazuh/wazuh/actions/workflows/vulnerability-scanner-tests.yml)
- [ ] Add or Update any relevant documentation to reflect the changes made.

PR #19747, opened 2023-10-19
Utils - Add  `ZLibHelper` decompressor class

tsarquis88:
|Related issue|
|---|
|#19739 |

# Description

This PR introduces a new class called `ZLibHelper` that is in charge of decompressing files. It is a wrapper around zlib.

Change log:
- New class added.
- New UT added.

# Tests

Apart from the unit tests implemented, no module is using this class so the RTR is not affected by the decompressor. Anyway, I checked that Wazuh's compilation is not affected and I've run the workflow locally and everything went well.


Damian-Mangold (approved):
LGTM !

Issue #19793, opened 2023-10-23
Key longer than 65 characters fails in HMAC


diegodfrf:
On Branch master

If the `OS_HMAC_SHA1_Str(const char *key, const char *text, os_sha1 output)` or `OS_HMAC_SHA1_File(const char *key, const char *file_path, os_sha1 output, int mode)` function is called with a `key` longer than 65 characters, it fails, and displays the following error when executing the unit tests:

``` 
[==========] tests: Running 5 test(s).
[ RUN      ] test_hmac_string
[       OK ] test_hmac_string
[ RUN      ] test_hmac_string_length_key
=================================================================
==7581==ERROR: AddressSanitizer: stack-use-after-scope on address 0x7fff29faf430 at pc 0x7f3f2a65f397 bp 0x7fff29faf2f0 sp 0x7fff29faea98
READ of size 20 at 0x7fff29faf430 thread T0
    #0 0x7f3f2a65f396 in __interceptor_memcpy ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:827
    #1 0x564344df0a02 in OS_HMAC_SHA1_Str os_crypto/hmac/hmac.c:44
    #2 0x564344def80a in test_hmac_string_length_key /home/user/wazuh2/src/unit_tests/os_crypto/hmac/test_hmac_op.c:74
    #3 0x7f3f29d2ba21 in cmocka_run_one_test_or_fixture (/usr/local/lib/libcmocka.so.0+0x6a21)
    #4 0x7f3f29d2c884 in _cmocka_run_group_tests (/usr/local/lib/libcmocka.so.0+0x7884)
    #5 0x564344df05f9 in main /home/user/wazuh2/src/unit_tests/os_crypto/hmac/test_hmac_op.c:126
    #6 0x7f3f294f1d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #7 0x7f3f294f1e3f in __libc_start_main_impl ../csu/libc-start.c:392
    #8 0x564344dee9b4 in _start (/home/user/wazuh2/src/unit_tests/cmake-build-debug-docker/os_crypto/hmac/test_hmac_op+0x849b4)

Address 0x7fff29faf430 is located in stack of thread T0 at offset 240 in frame
    #0 0x564344df073c in OS_HMAC_SHA1_Str os_crypto/hmac/hmac.c:22

  This frame has 5 object(s):
    [48, 144) 'context' (line 30)
[...]

PR #19794, opened 2023-10-23
Fix key longer than 65 characters fails in HMAC

diegodfrf:
os_sha1 was accessed out of scope.

|Related issue|
|---|
|#19793|

## Description

The variable `os_sha1` of the functions `OS_HMAC_SHA1_File(const char *key, const char *file_path, os_sha1 output, int mode)` and `OS_HMAC_SHA1_Str(const char *key, const char *text, os_sha1 output)` was accessed out of its scope, causing the error.

```
    if (key_length > HMAC_SHA1_BLOCKSIZE){
        os_sha1 sha_key;  <<<---
        OS_SHA1_Str(key, key_length, sha_key);
        key = sha_key;    <<<---
        key_length = SHA_DIGEST_LENGTH;
    }

    memset(o_key_pad, 0, sizeof(o_key_pad));
    memset(i_key_pad, 0, sizeof(i_key_pad));
    memcpy(o_key_pad, key, key_length);
    memcpy(i_key_pad, key, key_length);
```

## Tests

  - [x] Linux
  - [x] Windows
  - [x] MAC OS X
- [x] Source installation

### DIFF
--- a/src/os_crypto/hmac/hmac.c
+++ b/src/os_crypto/hmac/hmac.c
@@ -23,6 +23,7 @@ int OS_HMAC_SHA1_Str(const char *key, const char *text, os_sha1 output)
     unsigned char result[SHA_DIGEST_LENGTH + 1];
     unsigned char o_key_pad[HMAC_SHA1_BLOCKSIZE + 1];
     unsigned char i_key_pad[HMAC_SHA1_BLOCKSIZE + 1];
+    unsigned char key_temp[HMAC_SHA1_BLOCKSIZE + 1];
 
     int i;
     size_t key_length;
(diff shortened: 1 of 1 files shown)


matias-braida (dismissed):
LGTM

src/os_crypto/hmac/hmac.c
```diff
@@ -32,8 +32,8 @@ int OS_HMAC_SHA1_Str(const char *key, const char *text, os_sha1 output)
     key_length = strlen(key);
     text_length = strlen(text);
 
+    os_sha1 sha_key;
     if (key_length > HMAC_SHA1_BLOCKSIZE){
```
matias-braida, line 36: When this condition is true, the parameter "key" returns an invalid pointer that points to a non existing local variable "sha_key". This could be a problem if the calling function tries to use it.

src/os_crypto/hmac/hmac.c
```diff
@@ -84,8 +84,8 @@ int OS_HMAC_SHA1_File(const char *key, const char *file_path, os_sha1 output, in
 
     key_length = strlen(key);
 
+    os_sha1 sha_key;
     if (key_length > HMAC_SHA1_BLOCKSIZE){
```
matias-braida, line 88: When this condition is true, the parameter "key" returns an invalid pointer that points to a non existing local variable "sha_key". This could be a problem if the calling function tries to use it.

diegodfrf: done

matias-braida (approved):
LGTM

Issue #19943, opened 2023-10-30
Scan context std::variant usage for different inputs in the VD scanner


Dwordcito:
### Description

This issue focuses on enhancing the Vulnerability Detector (VD) Scanner by adding the usage of `std::variant` for various input types for handling heterogeneous data.

### User Stories

- As a developer, I want to fill the scanning context based on the type of message using std::variant.
- As a system administrator, I want the VD Scanner to ensure safe and correct usage of `std::variant` to prevent potential runtime issues.

### Tasks

1. Design Context Scanner Class:
   - Create a class within the VD Scanner responsible for scanning `std::variant` usage.
   - Define methods and data structures for analyzing and tracking `std::variant` usage.

2. Identify Input Types:
   - Implement functionality to identify the input types used with `std::variant` within the VD Scanner.
   - Recognize all potential data types that can be encapsulated by `std::variant`.

3. Integration:
   - Integrate the Context Scanner class into the VD Scanner pipeline.
   - Ensure that it is invoked when necessary to scan `std::variant` usage.

6. Testing:
   - Create comprehensive test cases to validate the Context Scanner's ability to identify and analyze `std::variant` usage.
   - Verify that feedback and guidance are provided correctly.

### Definition of Done

- The Context Scanner class is designed and implemented within the VD Scanner.
- It can identify and analyze `std::variant` usage for different input types.
- Feedback mechanisms are in place to notify developers of potential issues.
- Integration with the VD Scanner pipeline is successful.
- Comprehensive testing validates the functionality and feedback mechanisms.

PR #19984, opened 2023-11-02
Change to use fbs from memory (remove includes), changes in default policy to use CTI dev environment, add std variant for scan orchestration.

Dwordcito:
|Related issue|
|---|
|#19943|

This PR adds some important changes to the scan orchestration flow.


pereyra-m (commented):
LGTM!

Issue #19946, opened 2023-10-30
Store ECS VD state with dbsync.


Dwordcito:
## Description

This issue is focused on the instantiation of DBSync with persistent storage. The goal is to create a schema for the database with specific primary key fields, including the CVE ID(string), agent name(string), node name(string), and package name(string). Additionally, the schema will include a field for the payload in JSON format following the Elastic Common Schema and another field for the checksum, which is a hash of the record.

### User Stories

- As a developer, I want to instance the DBSync module that enables persistent storage with a defined schema.
- As a system administrator, I want to store and manage vulnerability data efficiently with specified primary key fields.
- As a user, I want the DBSync module to handle the storage and retrieval of vulnerability data seamlessly.

### Tasks

1. Schema Design:
   - Design a schema for the database that includes primary key fields: CVE ID, agent name, node name, and package name.
   - Define additional fields for storing payload data in Elastic Common Schema format (one string) and a checksum field (string).

2. Database Configuration:
   - Ensure that the DBSync class can be configured to use persistent storage with the specified schema.
   - Allow for configuration of database connection settings.

3. Testing:
   - Create comprehensive test cases to validate the functionality of the DBSync class with persistent storage.
   - Verify data insertion, retrieval, and integrity checks.

### Definition of Done

- The DBSync class with persistent storage is instantiated.
- The schema includes primary key fields for CVE ID, agent name, node name, and package name, and data values (payload and checksum)
- Comprehensive testing validates the functionality of the DBSync class.

### Implementation details
Instantiate in vulnerabilityScannerFacade.cpp and pass as a reference during the instantiation of InventorySync class.

PR #19997, opened 2023-11-02
Store ecs vd state with dbsync

diegodfrf:
|Related issue|
|---|
|#19946|

## Description

The DBSync class with persistent storage is instantiated.
The schema includes primary key fields for CVE ID, agent name, node name, and package name, and data values (payload and checksum)
Instantiate in vulnerabilityScannerFacade.cpp


matias-braida (approved):
LGTM

PR #20397, opened 2023-11-23
DatabaseFeedManager graceful shutdown improved

### LINKED ISSUE #20274: Improve the graceful shutdown of the databaseFeedManager when there are still messages to process
## Description

The `databaseFeedManager` module receives the content published by the `contentManager` and updates the corresponding local DBs.
But it was found that if we have still pending messages to process and we request a shutdown, the rocksDB wrapper is closed too early.

**How to reproduce**

The testtool was used to reproduce the issue with a timeout

```
./database_feed_manager_testtool -c ../../../../../wazuh_modules/vulnerability_scanner/testtool/databaseFeedManager/config.json -s 5 -i ./58000-api_file.json
```
[out.txt](https://github.com/wazuh/wazuh/files/13379064/out.txt)

## DoD

- [x] Find the root cause of the issue
- [x] Make the necessary changes to finish the current message processing before the shutdown
- [x] UT added
- [x] CI Pass

matias-braida:
|Related issue|
|---|
|Closes #20274 |

## Description

This PR modified the class TDatabaseFeedManager to allow a graceful shutdown when it already has some messages to process.

## Tests

New UT added "DatabaseFeedManagerTest.GracefulShutdown".

image

  - [ ] Linux
- [ ] UT running ok

### DIFF
--- a/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
@@ -369,6 +369,10 @@ public:
     }
 
 private:
+    /**
+     * Do not change the order of definition of these variables.
+     * Since it is important at the object destruction time.
+     */
     std::shared_ptr<TIndexerConnector> m_indexerConnector;
     std::unique_ptr<TContentRegister> m_contentRegistration;
     std::unique_ptr<Utils::RocksDBWrapper> m_descriptionsDatabase;
--- a/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
@@ -369,15 +369,15 @@ public:
     }
 
 private:
-    std::unique_ptr<TRouterSubscriber> m_contentUpdateSubscription;
-    std::unique_ptr<TContentRegister> m_contentRegistration;
     std::shared_ptr<TIndexerConnector> m_indexerConnector;
+    std::unique_ptr<TContentRegister> m_contentRegistration;
     std::unique_ptr<Utils::RocksDBWrapper> m_descriptionsDatabase;
     std::unique_ptr<Utils::RocksDBWrapper> m_remediationsDatabase;
     std::unique_ptr<Utils::RocksDBWrapper> m_translationsDatabase;
     std::unique_ptr<LRUCache<std::string, std::string>> m_translationsCache;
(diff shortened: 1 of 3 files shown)


pereyra-m:
@matias-braida , just a comment: the branch doesn't follow the required convention. It also might be too long.

src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
```diff
@@ -369,15 +369,15 @@ class TDatabaseFeedManager final : public Observer<nlohmann::json&>
     }
 
 private:
-    std::unique_ptr<TRouterSubscriber> m_contentUpdateSubscription;
-    std::unique_ptr<TContentRegister> m_contentRegistration;
     std::shared_ptr<TIndexerConnector> m_indexerConnector;
+    std::unique_ptr<TContentRegister> m_contentRegistration;
```
pereyra-m, line 373: If the order of the variables is important for the correct behavior of the class, I'd add a comment to prevent this problem from happening again.

src/wazuh_modules/vulnerability_scanner/tests/databaseFeedManager/databaseFeedManager_test.cpp
```diff
@@ -134,6 +136,58 @@ void DatabaseFeedManagerTest::SetUp()
...
+                    "dateUpdated": "2017-05-11T14:29:20Z",
+                    "state": "REJECTED"
+                },
+                "dataType": "CVE_RECORD",
+                "dataVersion": "5.0"
+            }
+        }
+    )"_json;
+    int cveNumber = 1;
+    for (auto itemNumber = 0; itemNumber < 1000; itemNumber++)
```
pereyra-m, line 178: We are creating this file many times, once per test, but using it only once. This makes the tests take longer.
We could create it only for the test that requires it.

pereyra-m (commented):
Great job with this bug!

I was able to verify that the test tool doesn't fail any more with the fix.

src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
matias-braida: fixed

src/wazuh_modules/vulnerability_scanner/tests/databaseFeedManager/databaseFeedManager_test.cpp
matias-braida: fixed

pereyra-m (approved):
LGTM!

Issue #20189, opened 2023-11-13
Update the flatbuffers' schema target to build them only if the file has changed


pereyra-m:
## Description

It was found that the compilation of the schemas for flatbuffer use the CMake `add_custom_target()` and it always considers that the target is outdated, causing the re-compilation of this target and all that depend on it even if the output is present and the source hasn't changed (for example `src/wazuh_modules/vulnerability_scanner/CMakeLists.txt`).

This behavior causes delays in the development because the project re-compiles code unnecessarily.

Another option could be the use of `add_custom_command()` with the `OUTPUT` and `DEPENDS` options.
https://cmake.org/cmake/help/latest/command/add_custom_command.html#examples-generating-files

## DoD

- [x] All the flatbuffer schemas used in the project are only built when required
- [x] CI Pass

PR #20599, opened 2023-12-01
Fix build flatbuffer schema target

diegodfrf:
|Related issue|
|---|
|#20189|

## Description
Now when compiling flatbuffer schemes, it validates whether the files exist or have been modified, to avoid compiling unnecessarily.

Changes were applied to SysCollector and Vulnerability_scanner

  - [x] Linux
- [x] Source installation


MiguelazoDS (approved):
GJ! 

First compilation after full clean up successfully

image

Re-compilation, just one second.

image

Issue #20641, opened 2023-12-05
index-status setting shouldn't be evaluated while vulnerability scanner is disabled


MiguelazoDS:
# Description 

The error following error shouldn't be present if the vulnerability scanner module is disabled. 

`2023/12/05 14:44:21 wazuh-modulesd:vulnerability-scanner: ERROR: VulnerabilityScannerFacade::start: Indexer cannot be disabled while vulnerability detection index is enabled.`

![image](https://github.com/wazuh/wazuh/assets/13010397/cc4f8636-d4be-4863-a389-671faca4576c)

# DoD 

- [x] Make the changes required.
- [x] Implement UTs if needed.

PR #20648, opened 2023-12-05
Ignore index-status if vulnerability scanner is disabled

MiguelazoDS:
|Related issue|
|---|
|#20641|

## Description
Do not consider index-status state if the vulnerabiilty scanner is disabled. 

## Logs/Alerts example

Before 

After

image

  - [x] Linux


GabrielEValenzuela (approved):
LGTM! The overall structure is clear, and the code is easy to read and understand.
The variable names are descriptive, and the comments provide helpful insights into the logic, wich now is correct according with the logic table of possible outcomes.

Great job Migue!

Issue #21102, opened 2023-12-26
Engine - Modify the api module to be able to register handlers that can respond to clients themselves.


NahuFigueroa97:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
|5.0.0 | Engine | Manager | Packages/Sources | OS version |

### Description
Currently, the API HANDLER expects a request and returns a response synchronously. The proposal is to modify this behavior to now expect both a request and a callback, eliminating synchronous return. The callback will be responsible for handling the response to a specific component each time a request is made to the API, thus enabling asynchronous responses.

### Objective
The main objective of this modification is to enable asynchronous responses instead of synchronous responses from the API HANDLER.

### Proposed Tasks

- Modify the API HANDLER to expect both a request and a callback instead of returning a response.
- Create a function that transforms existing synchronous handlers into asynchronous handlers without modifying already created modules that expect a request and return a response.
- Update the tester's handler to adapt to the new asynchronous approach, removing the use of futures and promises if they are no longer necessary.
- Overload existing functions that use futures and promises, allowing them to accept a callback as a parameter without removing the previous functionality.
- Additional Considerations
- Ensure that the transformation from synchronous to asynchronous handlers does not affect the functionality of existing modules.
- Thoroughly test to ensure that the proposed changes do not introduce errors in existing functionality.
- Document the changes made appropriately and provide usage examples for the new methods.

### Acceptance Criteria

- The API HANDLER now expects a request and a callback and does not return anything.
- The function for transforming synchronous handlers to asynchronous ones works correctly.
- The tester's handler has been updated to work asynchronously without relying on futures and promises.
- Functions using futures and promises have been overloaded to accept a callback as a parameter.
- API tests have been corrected and adapted to the new asynchronous approach.

PR #21108, opened 2023-12-26
Engine - Modify the api module to be able to register asyncronous handlers

NahuFigueroa97:
|Related issue|
|---|
|#21102|

### Description

Currently, the API HANDLER expects a request and returns a response synchronously. The proposal is to modify this behavior to now expect both a request and a callback, eliminating synchronous return. The callback will be responsible for handling the response to a specific component each time a request is made to the API, thus enabling asynchronous responses.


JavierBejMen (commented):
Why do we have an ingest method that receives a string? I don't see the use case for this

soynof (approved):
LGTM!

JavierBejMen (approved):
LGTM

PR #21117, opened 2023-12-26
Update 0810-sysmon_id_3.xml

branchnetconsulting:
|Related issue|
|---|
|#21321|

## Description
Fixing false-positive prone 92108 that matches destination port with substring 3389 rather than absolute string 3389.
We are seeing hits on the original 92108 rule where the destination port is something like 23389 or 33891 which is clearly not the intent of the rule.

## Tests

- Decoder/Rule tests


ooniagbi:
Closing this and opening a new PR to solve this issue.

PR #21146, opened 2023-12-28
Flatbuffers verifiers added

### LINKED ISSUE #21017: Add flatbuffer verifiers before accessing data from DBs
## Description

The internals DBs used by the new vulnerability detector use flatbuffers. The access to the fields of these structures is fast but if the binary data is malformed, the offset would point to anywhere and there are high chances of a segmentation fault.

Even when it adds a small overhead, it's much safer to add verifiers after reading data from the DBs in case the data is corrupt.
Some examples of these lack of sanity checks are:
- `getVulnerabilitiesCandidates()` (src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp)
- `processEvent()` (src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/eventDecoder.hpp)

There could be some other places where the sanity check is required.

## DoD

- [x] Find all the places where a potentially corrupt flatbuffer could be read and generate a crash
- [x] Implement the verifications
- [x] Manual tests
- [x] Add UT
- [x] CI Pass

matias-braida:
|Related issue|
|---|
|Closes #21017 |

## Description

This PR adds flatbuffers verifiers to data read from rocksdb previous to be used to avoid errors related to data corruption.

## Tests

Tests "DatabaseFeedManagerTest.GetVulnerabilityCandidatesCorrupted", "EventDecoderTest.TestUpdatedResourceCorrupted" and "EventDecoderTest.TestUpdatedTranslationResourceCorrupted" were added.

  - [ ] Linux
- [ ] Manual tests running ok
- [ ] UT running ok

### DIFF
--- a/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
@@ -432,6 +432,13 @@ public:
 
         for (const auto& [key, value] : m_feedsDatabases.at(cnaName)->seek(packageNameWithSeparator))
         {
+            flatbuffers::Verifier verifier(reinterpret_cast<const uint8_t*>(value.data()), value.size());
+            if (!NSVulnerabilityScanner::VerifyScanVulnerabilityCandidateArrayBuffer(verifier))
+            {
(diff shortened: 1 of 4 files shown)


pereyra-m (approved):
LGTM!

Issue #20963, opened 2023-12-20
Disabling the packages scan in syscollector doesn't remove the vulnerabilities in the inventory


pereyra-m:
## Description

During the E2E test it was found that the `integrity_clear` message isn't parsed by the new vulnerability detector.
When the packages scan is disabled in the agent, the agent's inventory is cleared but the vulnerabilities are still present.

## DoD

- [x] Handle the described situation properly
- [x] Add UT
- [x] CI Pass

GabrielEValenzuela:
Re-open, comment in PR

GabrielEValenzuela:
Segfault found, re open to fix it

PR #21174, opened 2024-01-02
Fix segfault in `integrity_clear` orchestration

GabrielEValenzuela:
|Related issue|
|---|
|#20963|

This PR aims to fix a segmentation fault error when an `integrity_clear` message is received and the instruction flow tries to check if a `syscollector_packages`


pereyra-m (approved):
LGTM!

After testing the development branch, it was confirmed that now **modulesd** doesn't crash.

Issue #21038, opened 2023-12-22
Include `wazuh.cluster.name` in events reported by Vulnerability Detection (VD)


GabrielEValenzuela:
### Problem Description
The field `wazuh.cluster.name` is currently not present in the events reported by the Vulnerability Detection (VD) system. Including this field is crucial for accurate cluster identification and analysis. This absence may result in a lack of context and hinder effective incident response and monitoring.

### Expected Behavior
The field `wazuh.cluster.name` should be included in the events reported by the Vulnerability Detection (VD) system. This will enable better correlation and analysis of events related to a specific Wazuh cluster.

#### Story 1: As a Security Analyst
- **I want the `wazuh.cluster.name` field to be present in VD events for accurate identification of the originating Wazuh cluster.**
- **I want to leverage this field to enhance incident response and analysis.**

#### Story 2: As a System Administrator
- **I want the VD system to report events with complete information, including the `wazuh.cluster.name` field.**
- **I want to have the necessary context to distinguish events originating from different Wazuh clusters.**

### Definition of Done Checklist

- [ ] Update the configuration of the Vulnerability Detection (VD) system to include the `wazuh.cluster.name` field in reported events.
- [ ] Confirm that the updated events contain the `wazuh.cluster.name` field.
- [ ] Verify the impact of the change on incident response and analysis.

PR #21179, opened 2024-01-03
Include cluster_name in vulnerability-detection config

diegodfrf:
|Related issue|
|---|
|#21038|

## Description

Added cluster_name to the vulnerability-detection configuration, from cluster-config.c in wm_vulnerability_scanner_main.

### DIFF
--- a/src/wazuh_modules/wm_vulnerability_scanner.c
+++ b/src/wazuh_modules/wm_vulnerability_scanner.c
@@ -68,6 +68,10 @@ void* wm_vulnerability_scanner_main(wm_vulnerability_scanner_t * data)
         {
             cJSON_AddStringToObject(data->vulnerability_detection, "feed-update-interval", "60m");
         }
+        if (!cJSON_GetObjectItem(data->vulnerability_detection, "cluster-name"))
+        {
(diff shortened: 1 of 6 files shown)


src/wazuh_modules/wm_vulnerability_scanner.c
```diff
@@ -68,6 +68,10 @@ void* wm_vulnerability_scanner_main(wm_vulnerability_scanner_t * data)
         {
             cJSON_AddStringToObject(data->vulnerability_detection, "feed-update-interval", "60m");
         }
+        if (!cJSON_GetObjectItem(data->vulnerability_detection, "clusterName"))
+        {
+            cJSON_AddStringToObject(data->vulnerability_detection, "clusterName", get_cluster_name());
```
pereyra-m, line 73: The call to `get_cluster_name()` here is leaking memory because cJSON creates another copy of the string.

pereyra-m (changes requested):
GJ!

The event to index contains the expected field

src/wazuh_modules/wm_vulnerability_scanner.c
diegodfrf: Done

pereyra-m (approved):
LGTM!

PR #21184, opened 2024-01-04
Refactor to use column families and set offset after update one file during the offset update.

### LINKED ISSUE #21010: Update the local offset only when the database feed manager has processed all the elements
## Description

Currently, the `content_manager` module updates the local DB with the new offset when it publishes the messages. But it could happen that if the manager is stopped before the `databaseFeedManager` module has processed all the messages, the manager will miss the information.

In this issue, the `databaseFeedManager` should be in charge of updating the local DB only when it has finished processing all the data. For this task, is also mandatory to implement a transaction mechanism for RocksDB , to ensure that either all changes or none have been applied

image (25)

## DoD

- [x] Implement all the bulk/transaction methods required in the RocksDB wrapper 
- [x] Move the offset update mechanism to the  `databaseFeedManager`
- [x] Confirm with manual tests that the process can be stopped and resumed at any time
- [ ] Add UT
- [ ] CI Pass

Dwordcito:
|Related issue|
|---|
|Closes #21010 |

## Description
This PR aims to add two things.

The first one is the management of column families, to avoid having multiple database instances in the code. The principal benefit of this is the maintenance and clean code.

The second one is the management of transactions to set the current offset only when the transactions are committed.

### DIFF
--- a/src/shared_modules/utils/rocksDBWrapper.hpp
+++ b/src/shared_modules/utils/rocksDBWrapper.hpp
@@ -17,24 +17,46 @@
 #include <filesystem>
 #include <memory>
 #include <rocksdb/db.h>
+#include <rocksdb/utilities/transaction.h>
+#include <rocksdb/utilities/transaction_db.h>
 #include <stdexcept>
 #include <string>
 #include <vector>
 
 namespace Utils
 {
+    class RocksDBTransaction;
+    class IRocksDBWrapper
+    {
+    public:
+        virtual void put(const std::string& key, const rocksdb::Slice& value, const std::string& columnName) = 0;
+        virtual void put(const std::string& key, const rocksdb::Slice& value) = 0;
+        virtual void delete_(const std::string& key, const std::string& columnName) = 0; // NOLINT
+        virtual void delete_(const std::string& key) = 0;                                // NOLINT
+        virtual void commit() = 0;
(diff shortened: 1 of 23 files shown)


src/shared_modules/utils/rocksDBWrapper.hpp
```diff
@@ -392,9 +369,96 @@ namespace Utils
...
+        /**
+         * @brief Initialize transaction.
+         * @return RocksDBTransaction Transaction object.
+         */
+        std::unique_ptr<RocksDBTransaction> createTransaction()
+        {
+            return std::make_unique<RocksDBTransaction>(this);
+        }
+
+        void commit() override
```
GabrielEValenzuela, line 381: This method appears to be called in `dabaseFeedManager.hpp:157`, but the behavior is undefined. Is the expected behavior?

Dwordcito: yes, because I use an abstract base class, and this method is needed when the IRocksDBWrapper points to RockDBTransacion

GabrielEValenzuela (approved):
LGTM! The overall structure is clear, the code is easy to read and understand, and the behavior is expected during the manager installation and test.

No errors were found and the functionality is correct.

Great job Octa!

PR #21197, opened 2024-01-04
Improve logging helper to avoid multiple vsnprintf calls

### LINKED ISSUE #21020: Vulnerability detector modulesd crashed
# Description 

It was found during the E2E that the vulnerabilities found (matches), the indexed vulnerabilities, and the alerts do not match. 
Later on, we find out that it was that `modulesd` crashed during the scan.

## Steps to reproduce 

- Connect several agents (this may increase the change of this happening but we are not sure yet).
- Wait until feeds are downloaded, processed, and stored in databases.
- Stop manager and remove agent databases.
- Start manager and wait until scan finishes.

**It was reproduced in a cluster environment** 

Many difference between matches, alerts and indexed vulnerabilities occurred in both, worker and master 

### Worker 

We can see the alerts logged are not the same in the alerts.json file and do not match the indexer logs number either.

<details><summary>Expand</summary>

image
image

</details>

Vulnerabilities found also different

<details><summary>Expand</summary>

image

</details>

No duplicates found

<details><summary>Expand</summary>

image

</details>

Indexed vulnerabilities do not match its log messages.

<details><summary>Expand</summary>

image
image

</details>

### Master 

We can see that we have different values in the master too and even different results than those previously found in the worker node 

<details><summary>Expand</summary>

image
image
image

</details>

### Multiple errors 

<details><summary>Expand</summary>

image
image

</details>

It was not reproducible in a local environment using a single node.

<details><summary>Expand</summary>

image
image

</details>

pereyra-m:
|Related issue|
|---|
|Closes #21020 |

## Description

This PR updates the `loggingHelper` to use a `va_list` as argument when calling the real log function to void an extra call to `vsnprintf`.

## Tests

Manual tests were performed:
- The environment that crashed now prints the test string succesfully
- The local environment shows all the normal logs as expected
- The testtool of vulnerability detector keeps writing the output to the console as expected

  - [x] Linux
- [x] Source installation

### DIFF
--- a/src/headers/debug_op.h
+++ b/src/headers/debug_op.h
@@ -124,7 +124,7 @@ int isChroot(void);
  * @param line Line number.
  * @param func Function name.
  * @param msg Message to send into the log.
- * @para args Variable arguments list.
+ * @param args Variable arguments list.
  */
 void mtLoggingFunctionsWrapper(int level, const char* tag, const char* file, int line, const char* func, const char* msg, va_list args);
 
--- a/src/headers/debug_op.h
+++ b/src/headers/debug_op.h
@@ -124,7 +124,8 @@ int isChroot(void);
  * @param line Line number.
  * @param func Function name.
  * @param msg Message to send into the log.
+ * @para args Variable arguments list.
  */
-void mtLoggingFunctionsWrapper(int level, const char* tag, const char* file, int line, const char* func, const char* msg);
+void mtLoggingFunctionsWrapper(int level, const char* tag, const char* file, int line, const char* func, const char* msg, va_list args);
 
 #endif /* DEBUG_H */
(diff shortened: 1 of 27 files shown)


src/headers/debug_op.h
```diff
@@ -124,7 +124,8 @@ int isChroot(void);
  * @param line Line number.
  * @param func Function name.
  * @param msg Message to send into the log.
+ * @para args Variable arguments list.
```
Dwordcito, line 127:
```suggestion
 * @param args Variable arguments list.
```

Dwordcito, line 127:
```suggestion
 * @param args Variable arguments list.
```

Dwordcito (approved):

Issue #20727, opened 2023-12-12
Windows Agent 4.8.0 permission errors on Windows 11 after uprade


Deblintrake09:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.8.0-alpha1 | Wazuh component |Agent | Packages | Windows 11 |

## Description

During testing of Issue https://github.com/wazuh/wazuh/issues/20675, it was found that the Windows Agent on Windows 11 does not open the UI after upgrade from `4.7.0`. When trying to open the UI manually a permission error is shown.

## Evidence

![image](https://github.com/wazuh/wazuh/assets/14501079/c0789572-221c-41f6-9239-35b89f8a9c91)
![image](https://github.com/wazuh/wazuh/assets/14501079/95431613-7274-4dbd-a0b2-5168c058ef7b)
![image](https://github.com/wazuh/wazuh/assets/14501079/83aeaeaa-ee6a-4e4e-811d-32adf7566cdd)
![image](https://github.com/wazuh/wazuh/assets/14501079/3b811449-9d82-4b62-82c4-536217e4c3d4)

vikman90:
Possibly related to #12906.

vikman90:
I think the problem is that the permissions are applying for "**this folder only**":

![image](https://github.com/wazuh/wazuh/assets/10536251/acce5eb8-78ef-43c8-8c39-cc3b79b6bf93)


This is preventing the contained files from inheriting the permissions.

Deblintrake09:
- The permission Issue is present also when updating through WPK from a lower version to `4.8.0`
- When Force Updating `4.8.0 -> 4.8.0` the permission bug is not present.

ncvicchi:
UPDATE:

A series of tests were carried out to narrow down the possibilities.

Progressive updates were made from version 4.2.3 to 4.8.0, without ever uninstalling the previous version.

In these cases, the behavior described in the issue was not observed.

However, when installing one version, upgrading to the next, and then uninstalling it before performing the next test, some different behaviors were observed:

#### Changing the test: Install, upgrade, uninstall:

<details><summary>Tests</summary>

|Test|Install|Permissions|Upgrade|Permissions|
|--|--|--|--|--|
|4.2.3 -> 4.2.4|OK|OK|OK|OK|
|4.2.4 -> 4.2.5|OK|OK|OK|OK|
|4.2.5 -> 4.2.6|OK|OK|OK|OK|
[...]

PR #21198, opened 2024-01-04
Fixed permissions inheritance on Windows install script

ncvicchi:
|Related issue|
|---|
|#20727|

## Description

A problem with permissions inheritance on Windows installer was detected which prevented access to Wazuh's folder when a upgrade was made, although the error was in the installer (the installer left a state that was functional but would be troublesome when an update was executed).

Permissions inheritance was enabled for the Wazuh's 'folder in order to mitigate this.

### DIFF
--- a/src/win32/InstallerScripts.vbs
+++ b/src/win32/InstallerScripts.vbs
@@ -345,7 +345,7 @@ public function config()
         ' Remove last backslash from home_dir
         install_dir = Left(home_dir, Len(home_dir) - 1)
 
-        setPermsInherit = "icacls """ & install_dir & """ /inheritancelevel:d /q"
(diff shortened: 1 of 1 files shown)


lchico (approved):
Thanks, @ncvicchi. LGTM!

Deblintrake09 (changes requested):
When updating previous installs with Custom 4.7.2 package, the reported issue is not fixed. It only works when doing a clean install of 4.7.2 and updating.

The fix should allow users to update from lower versions to 4.7.2, not only clean installs.

ncvicchi:
The issue is that since older versions of the installer were faulty, they left the system with the permissions problem. This PR solves that the installer won't break the folder permissions.

I am not sure that is should be the scope of this PR to fix broken systems.

PR #21242, opened 2024-01-07
Fix invalid get when recover the cve list from the inventory list.

### LINKED ISSUE #21241: Fix Exception Caused by Malformed JSON Deleting Indexer Elements
#### Description

This issue addresses an exception that occurs due to improperly formatted JSON when attempting to delete elements from the indexer. The issue manifests as an exception being thrown, leading to unexpected behavior. The goal is to identify the root cause of this problem and implement a solution.

#### User Stories

- As a developer, I want to ensure that the indexer can handle malformed JSON without causing exceptions.
- As a system administrator, I want to prevent unexpected failures when interacting with the indexer due to malformed data.

#### Tasks

1. **Identify Root Cause:**
   - Investigate the source of the malformed JSON that triggers the exception.

2. **Fix**
   - Implement a fix for this issue.

#### Definition of Done

- The root cause of the malformed JSON triggering the exception is identified
- Fix

image

Dwordcito:
|Related issue|
|---|
|Closes #21241 |

## Description
This PR aims to solve a problem during the integrity clear event the previous code uses the data() method, which retrieves the value from the slice, but the returned information from data() not are null-terminated, For this specific case the correct function to call is ToString.

### DIFF
--- a/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/inventorySync.hpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/inventorySync.hpp
@@ -57,7 +57,7 @@ public:
             data->m_isInventoryEmpty = true;
             for (const auto& dbQuery : m_inventoryDatabase.seek(agentPackageKey))
             {
-                auto listCve = Utils::split(dbQuery.second.data(), ',');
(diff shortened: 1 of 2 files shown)


GabrielEValenzuela (approved):
LGTM!

Good catch Octa!

Issue #21194, opened 2024-01-04
Vulnerability detector refactor flatbuffer unable to parse invalid value


MiguelazoDS:
# Description 

It was found an invalid value parsing message during a non-related implementation of a JSON message from the manager.

The value in data is already invalid 

https://github.com/wazuh/wazuh/blob/ae2b8ef88b2a4bf8cbfb5a0a214bf0db22ce4eb3/src/wazuh_modules/wm_syscollector.c#L102-L120

```console
2024/01/04 16:01:55 wazuh-modulesd[166035] wm_syscollector.c:119 at wm_sys_send_dbsync_message(): INFO: MESSAGE DATA SERVER: {"component":"syscollector_network_iface","data":{"attributes":{"adapter":null,"checksum":"038761887bf3f7441d3461655124f23f151516af","item_id":"7a60750dd3c25c53f21ff7f44b4743664ddbb66a","mac":"02:d0:f1:1a:d8:67","mtu":1500,"name":"enp0s3","rx_bytes":-953270236,"rx_dropped":0,"rx_errors":0,"rx_packets":2380655,"scan_time":"2024/01/04 19:01:55","state":"up","tx_bytes":39978629,"tx_dropped":0,"tx_errors":0,"tx_packets":506088,"type":"ethernet"},"index":"7a60750dd3c25c53f21ff7f44b4743664ddbb66a","timestamp":""},"type":"state"}
```

![2024-01-04_16-51](https://github.com/wazuh/wazuh/assets/13010397/5973d931-6019-4c42-bcf5-87d31f3f6808)

# DoD 

- [x] Find the root cause of the issue.
- [x] Propose a fix.
- [x] Add UTs.

PR #21292, opened 2024-01-08
Enhancement/21194 network small data type

MiguelazoDS:
|Related issue|
|---|
|#21194|

## Description

This simple PR is intended to avoid flatbuffer conversion errors by not making the flatbuffer schema too strict. 
The information came as a char*, it would add overhead to sanitize that string, without mentioning that converting to JSON to later come back to char * is not practical.

## Logs/Alerts example

Please check the analysis here https://github.com/wazuh/wazuh/issues/21194#issuecomment-1881407603

  - [x] Linux
- [x] Source installation


pereyra-m (approved):
LGTM!

PR #21323, opened 2024-01-10
Fix regex in rule 92108 causing false positives

ooniagbi:
|Related issue|
|---|
|#21321|

## Description
Fixing false-positive prone 92108 that matches destination port with substring 3389 rather than absolute string 3389.
We are seeing hits on the original 92108 rule where the destination port is something like 23389 or 33891 which is clearly not the intent of the rule.

## Tests

- Decoder/Rule tests
  - [x] Added unit testing files ".ini"
  - [x] runtests.py executed without errors


IsExec (approved):
LGTM

PR #21429, opened 2024-01-13
correct some Debian 12 SCA checks (wrong files, typo, etc)

alaunay:
## Description

Modify SCA yml code for Debian 12

## Changes
- one space too much for the freevxfs check
- use the sudoers.d directory (wrongly tried the sudoers file for a directory) for two checks
- check gdm installation (or not), copied from debian 11
- allow the yescrypt algorithm used by Debian since deb11

## Checks

- Decoder/Rule tests
  - [X] runtests.py executed without errors


Trooper-Ish:
The extra space on line 75 (freevxfs check) was tripping me up. came here to report, and found Alaunay already reported that, and a bunch of other nugggets. thanks sir/ma'am/whatever

PR #21445, opened 2024-01-15
Method TDatabaseFeedManager::getCnaName was modified to detect vendor…

### LINKED ISSUE #21232: The Ubuntu agent packages aren't scanned against the vendor feed because of the vendor package field
## Description

The alerts generated and not generated by both the legacy vulnerability detector and the refactored one were compared. See https://github.com/wazuh/wazuh/issues/21012 for more details.

Currently, the vulnerability scanner expects that the vendor field of the package starts with the word Ubuntu to scan it against the vendor feed. If it doesn't, the NVD is used as fallback. But there are sometimes when the vendor is from a person with a mail @ubuntu. 

- CVE-2022-2806: **sosreport**, `4.0-1ubuntu2` (https://ubuntu.com/security/CVE-2022-2806 https://nvd.nist.gov/vuln/detail/CVE-2022-2806). The legacy vuln-det detects this as vulnerable because version is lower than `4.3-1ubuntu0.20.04.2` (`focal`) but the refactored module fails to find this vulnerability because its vendor is Eric Desrochers <slashd@ubuntu.com> and the scan against the NVD doesn't find a match (it even has a different version type). 

The issue should fixed in two stages:
- Here, we'll add more methods to the `databaseFeedManager` to look for an specific key in more ways, not only at the beginning
- The `databaseFeedManager`will in the future get this information from the downloaded content #21233

## DoD

- [x] Implement the changes and make sure that these type of situations is now covered
- [x] Add UT
- [x] CI pass

matias-braida:
|Related issue|
|---|
|Closes #21232  |

## Description

This PR modified the method TDatabaseFeedManager::getCnaName to allow the search of the vendor substring in the whole string of the package vendor.
UTs were also added.

## Tests

  - [ ] Linux
- [ ] Run UTs successfully
- [ ] Run Manual tests

### DIFF
--- a/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
@@ -566,9 +566,10 @@ public:
 
         const auto vendorLowerCase = Utils::toLowerCase(vendor.data());
         // LCOV_EXCL_START
-        const auto it = std::find_if(vendorToCnaName.begin(),
-                                     vendorToCnaName.end(),
-                                     [&](const auto& pair) { return Utils::startsWith(vendorLowerCase, pair.first); });
+        const auto it =
+            std::find_if(vendorToCnaName.begin(),
+                         vendorToCnaName.end(),
+                         [&](const auto& pair) { return vendorLowerCase.find(pair.first) != std::string::npos; });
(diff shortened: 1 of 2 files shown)


pereyra-m (approved):
LGTM!

@matias-braida , please upload the evidence of manual tests (logs, etc.)

pereyra-m:
@matias-braida , please update the PR name.

PR #21580, opened 2024-01-23
Create a Github Actions matrix to allow content versioning

### LINKED ISSUE #21540: Update the vulnerability scanner DB generator workflow to allow versioning
## Description

The `.github/workflows/vulnerability-scanner-generate-database.yml` workflow currently allows only a fixed output file name.
In the future, the content that will be distributed with the installer might change from one version to another, so it's required to update the workflow and allow a customization.

Also, it was found that the workflow runs the UT unnecessary and we could avoid this to speed up the process. 

## DoD

- [x] Create a version [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) that will hold for now the v4.8.0 value only. Update the name in all the required places.
- [x] Create another action that compiles the project without running the tests.
- [x] Manual tests to verify the changes.

pereyra-m:
|Related issue|
|---|
|Closes #21540|

## Description

This PR creates a matrix to allow generating the vulnerability detector DBs with different version.
It also speeds up the workflow:
- Only the **syscollector** schemas are compiled
- The tests are skipped when not required

## Tests

It was confirmed that now the vulnerability detector workflow takes less time after removing the unnecessary compilation steps 

The vulnerability detector DB generation now has a matrix and the value is the expected

Also, it takes less time now that it doesn't run the UT

- [x] CI Pass

### DIFF
--- a/.github/workflows/vulnerability-scanner-generate-database.yml
+++ b/.github/workflows/vulnerability-scanner-generate-database.yml
@@ -4,9 +4,18 @@ on:
   workflow_dispatch:
   schedule:
     - cron: '0 0 * * *'
+  pull_request:
+    paths:
+      - ".github/workflows/vulnerability-scanner-generate-database.yml"
+      - ".github/actions/compile_and_test/action.yml"
+      - ".github/actions/vulnerability_scanner_deps/action.yml"
 
 jobs:
   vulnerability-scanner-upload-database:
+    strategy:
+          fail-fast: false
+          matrix:
+              wazuh_version: ["4.8.0"]
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/vulnerability-scanner-generate-database.yml
```diff
@@ -4,16 +4,38 @@ on:
...
       - name: Checkout repository
         uses: actions/checkout@v3
         with:
           submodules: recursive
+          fetch-depth: 0
+
+      - name: Checkout correct branch
+        run: |
+          git checkout tags/v${{ matrix.wazuh_version }} || echo "Failed to find tag v${{ matrix.wazuh_version }}"
+          git checkout ${{ matrix.wazuh_version }} || echo "Failed to find branch ${{ matrix.wazuh_version }}"
```
sebasfalcone, line 31: +1: Great solution

.github/workflows/vulnerability-scanner-generate-database.yml
```diff
@@ -73,11 +99,16 @@ jobs:
 
       - name: Vulnerability scanner - Uploading file to S3
         run: |
```
sebasfalcone, line 101: Nitpick: The condition can be moved here, [see example](https://docs.github.com/en/actions/learn-github-actions/contexts#example-usage-of-the-inputs-context-in-a-reusable-workflow)

sebasfalcone (approved):
LGTM! 

The comment I've done is just a nitpick so if you consider that is not applicable mark it as resolved

GabrielEValenzuela (approved):
LGTM! The overall structure is clear, and the code is easy to read and understand.

Great job Mati!

.github/workflows/vulnerability-scanner-generate-database.yml
pereyra-m: I've tried to skip the steps but the **if** condition seems to work only for jobs.

Issue #21575, opened 2024-01-23
Add interface and static libkeystore.a library in the CMake project build process


Dwordcito:
### Description

This issue is focused on enhancing the project by adding methods that enable the insertion or update of key-value pairs within a specified column family. It also involves integrating the project into the root CMake configuration and adding a dependency on the indexer connector module. Below are the key aspects of this issue:

#### Add interfaces
The project requires the implementation of methods that facilitate the insertion or updating of key-value pairs in a specific column family. For instance, a sample method might be:

```cpp
/**
 * Insert or update a key-value pair in the specified column family.
 *
 * @param columnFamily The target column family.
 * @param key The key to be inserted or updated.
 * @param value The corresponding value.
 */
void put(const std::string& columnFamily, const std::string& key, const rocksdb::Slice& value);
```

```cpp
/**
 * Get the key value in the specified column family.
 *
 * @param columnFamily The target column family.
 * @param key The key to be inserted or updated.
 * @param value The corresponding value to be returned.
 */
void get(const std::string& columnFamily, const std::string& key, rocksdb::PinnableSlice& value);
```
#### Project Integration with CMake

To ensure seamless integration with the project build system, it is necessary to create CMakelists files and include the project using add_subdirectory. This ensures that the project can be easily built and managed as part of the larger codebase.

#### Dependency Indexer Connector

The indexer connector module depends of this library. Therefore, the necessary dependencies and linkage should be added to enable seamless interaction with this module.

### Definition of done
- Add interfaces.
- Successful project integration with CMake using add_subdirectory.
- Appropriate dependencies and linkage added for interaction with the indexer connector module.

## Project structure

```tree
|-- src
    |-- shared_modules
        |-- keystore
            |-- include
                 |-- keyStore.hpp
            |-- src
                 |-- keyStore.cpp
            |-- tests
                 |-- unit
                     |-- unitTestFile_test.hpp
                     |-- unitTestFile_test.cpp
                 |-- component
                     |-- componentTestFile_test.hpp
                     |-- componentTestFile_test.cpp
```

mateocervilla:
## Update

- Create branch [`21575-libkeystore-lib-interface`](https://github.com/wazuh/wazuh/tree/21575-libkeystore-lib-interface)
- Create project tree
- Add headers and definitions
- Create `CMakelists.txt`
- Test if library works
- Fix CMake for github actions builds
- Fix CMake for github actions Ubuntu 20 build
[...]

PR #21609, opened 2024-01-24
Add interface and static libkeystore.a library

mateocervilla:
|Related issue|
|---|
|#21575|

## Description

As it is explained in the related issue, this PR adds methods that enable the insertion or update of key-value pairs within a specified column family. It also integrates the project into the root CMake configuration and adds a dependency on the indexer connector module. 

#### Interfaces
These are the implementation of methods that facilitate the insertion or updating of key-value pairs in a specific column family:

```cpp
void put(const std::string& columnFamily, const std::string& key, const rocksdb::Slice& value);

void get(const std::string& columnFamily, const std::string& key, rocksdb::PinnableSlice& value);
```

## Project structure

```tree
|-- src
    |-- shared_modules
        |-- keystore
            |-- CMakeLists.txt
            |-- include
                 |-- keyStore.hpp
            |-- src
                 |-- keyStore.cpp
            |-- tests
                 |-- unit
                     |-- unitTestFile_test.hpp
                     |-- unitTestFile_test.cpp
                 |-- component
                     |-- componentTestFile_test.hpp
                     |-- componentTestFile_test.cpp
```

## Tests
  - [x] Linux
- [x] Source installation


src/shared_modules/keystore/src/keyStore.cpp
```diff
@@ -0,0 +1,10 @@
+/*
+ * Wazuh keystore
+ * Copyright (C) 2015, Wazuh Inc.
+ * January 24, 2024.
+ *
+ * This program is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU General Public
+ * License (version 2) as published by the FSF - Free Software
+ * Foundation.
+ */
```
ncvicchi, line 10: put() and get() bodies should be present and callable in the cpp files

ncvicchi:
LGTM!

ncvicchi (approved):
LGTM!

Leoquicenoz (approved):
LGTM!!

Issue #21574, opened 2024-01-23
Remove user and password tags from indexer configuration


Dwordcito:
### Description

This issue aims to remove the parsing processes currently used to load the user and password configuration of the 'indexer' tag from the ossec.conf file. This change is part of the larger effort to transition to using the Wazuh keystore for secure configuration management.

### User Story

As a developer, I want to remove the load of the user and password fields from the  'indexer' tag in ossec.conf. This simplification is a preliminary step towards implementing the use of the Wazuh keystore for more secure and efficient configuration management.

### Proposed Solution

The proposed solution involves the following key steps:

1. **Remove Parsing Code:**
   - Identify and remove the parsing code responsible for extracting the user and password from the 'indexer' tag configuration if necessary, and modify unit tests.

2. **Default configuration:**
   - Identifiy and remove fields from the default configuration. -> https://github.com/wazuh/wazuh/blob/master/etc/templates/config/generic/wodle-indexer.manager.template

### Definition of Done

- Parsing code for user and password in the 'indexer' tag configuration has been removed if necessary.
- Remove from the default configuration.

Leoquicenoz:
## Update

1. Remove the username and password fields referring to the indexer in all files containing it, including default configuration (templates) and Unit tests.
2. It compiles without any problem.
3. It remains to be seen if the UTs pass

Leoquicenoz:
## Update

1. Changes are made to some unit tests that had not been modified before and then it is verified that they all pass, the coverage is not modified.
2. PR created and changes uploaded.

ncvicchi:
Reopen to address several situations that we realized that should be taken into account

ncvicchi:
Changed ETA to give time to resolve new situations

Leoquicenoz:
We closed the issue again since it was reopened thinking that the modification of the `wazuh-install.sh` file was included here and that we had overlooked it, but in issue [#2794](https://github.com/wazuh/wazuh-packages/issues/2794) this modification is being carried out.

PR #21643, opened 2024-01-26
Keystore get method implementation

### LINKED ISSUE #21577: Adapt the indexer connector to use the keystore database.
### Description

This issue aims to integrate the `get` method within the project to retrieve the required authentication credentials (user and password) from the indexer. These credentials will be used for authentication within the indexer connector module.

#### Method Functionality

The `get` method should be designed to retrieve the following authentication credentials:

- **User**: The username required for authenticating with the indexer.
- **Password**: The password associated with the specified user.

#### Integration Process

The integration process should include the following steps:

1. Implement the `get` method within the project's codebase. (Data is encripted with RSA algorithm using the sslmanager.cert and sslmanager.key)
2. Ensure that the method can successfully retrieve the user and password information from the indexer configuration.
3. Store the retrieved credentials in secure memory or variables to prevent exposure.

### Definition of Done

- The `get` method is successfully implemented and functional.
- It retrieves the user and password from the indexer configuration.
- The retrieved credentials are securely stored or handled.
- Appropriate error handling is in place for any potential issues during retrieval.

### Notes.
- Rocksdb is used as database. Path: queue/keystore
- Open rocksdb instance to queue/keystore as readonly.

mateocervilla:
|Related issue|
|---|
|#21577|

## Description

This PR adds the Get method implementation for retrieving key-value pairs within a specified column family from the database and updates the `indexer_connector` to implement this method. Closes #21577

```cpp
void Keystore::get(const std::string& columnFamily, const std::string& key, std::string& value)
```

## Tests

  - [x] Linux
- Memory tests for Linux


src/shared_modules/keystore/src/keyStore.cpp
```diff
@@ -40,6 +40,38 @@ void Keystore::put(const std::string& columnFamily, const std::string& key, cons
...
 
-void Keystore::get(const std::string& columnFamily, const std::string& key, rocksdb::PinnableSlice& value)
+void Keystore::get(const std::string& columnFamily, const std::string& key, std::string& value)
 {
+    std::string encryptedValue;
+
+    try
+    {        
+        // Get from DB
+        Utils::RocksDBWrapper keystoreDB = Utils::RocksDBWrapper(DATABASE_PATH, false);
```
ncvicchi, line 50: Is it possible that keystorDB not to be valid? Can that be validated before use?

src/shared_modules/keystore/src/keyStore.cpp
```diff
@@ -40,6 +40,38 @@ void Keystore::put(const std::string& columnFamily, const std::string& key, cons
...
+
+        if(!keystoreDB.columnExists(columnFamily)) {
+            char msg[100];
+            sprintf(msg, "Column %d does not exists in the database.", columnFamily.c_str());
```
ncvicchi, line 54: I think is nittier `Column %d does not exists at the database` instead of `Column %d does not exists in the database`

src/shared_modules/keystore/src/keyStore.cpp
```diff
@@ -40,6 +40,38 @@ void Keystore::put(const std::string& columnFamily, const std::string& key, cons
...
+            logError(KS_NAME, msg);
+            throw std::runtime_error(msg);
+        }
+
+        if (!keystoreDB.get(key, encryptedValue, columnFamily)) {
+            char msg[100];
+            sprintf(msg, "Could not find key '%s' in column '%s'.", key, columnFamily);
```
ncvicchi, line 61: NIT: I think is nittier `Could not find key '%s' at column '%s'.` instead of `Could not find key '%s' in column '%s'.`

ncvicchi: What would happen if key and olumnFamily together sums more than 70 bytes? I think asprintf would be a better option, of snprintf & some dynamic allocation method if not available in all environments

src/shared_modules/indexer_connector/src/indexerConnector.cpp
```diff
@@ -75,7 +81,18 @@ IndexerConnector::IndexerConnector(
         }
     }
 
-    secureCommunication.sslCertificate(sslCertificate)
+    try
+    {
+        Keystore::get(INDEXER_COLUMN, USER_KEY, username);
```
ncvicchi, line 86: Should we check that username and password are not empty?

src/shared_modules/indexer_connector/src/indexerConnector.cpp
```diff
@@ -75,7 +81,18 @@ IndexerConnector::IndexerConnector(
...
+        Keystore::get(INDEXER_COLUMN, PASSWORD_KEY, password);
+    }
+    catch (const std::exception& e)
+    {
+        logError(IC_NAME, "%s", e.what());
+    }
```
tdrauncieal, line 92: Include all function code inside of the first `try-catch` block:

```c++
    try
    {
        Keystore::get(INDEXER_COLUMN, USER_KEY, username);
        Keystore::get(INDEXER_COLUMN, PASSWORD_KEY, password);

        secureCommunication.basicAuth(username + ":" + password)
            .sslCertificate(sslCertificate)
            .sslKey(sslKey)
            .caRootCertificate(caRootCertificate);
.
.
.
                    }
                } while (!m_initialized && !m_cv.wait_for(lock, sleepTime, [&]() { return m_stopping.load(); }));
            });
    }
    catch (const std::exception& e)
    {
        logError(IC_NAME, "%s", e.what());
    }
}
```

src/shared_modules/keystore/src/keyStore.cpp
mateocervilla: The object cannot be invalid, if any error occurred during the constructor execution, the try-catch will handle it.

tdrauncieal (changes requested):

src/shared_modules/indexer_connector/src/indexerConnector.cpp
mateocervilla: I think it is beyond the scope of this issue. I should only replace the code (deleted in the issue #21574):
```cpp
    if (config.contains("username") && config.contains("password"))
    {
        username = config.at("username").get_ref<const std::string&>();
        password = config.at("password").get_ref<const std::string&>();
    }
```
for:
```cpp
    try
    {
        Keystore::get(INDEXER_COLUMN, USER_KEY, username);
        Keystore::get(INDEXER_COLUMN, PASSWORD_KEY, password);
    }
    catch (const std::exception& e)
    {
        logError(IC_NAME, "%s", e.what());
    }
```

tdrauncieal: An alternative solution is, if `Keystore::get()` would generate a potential exception, 
do not catch, to avoid continuing with the flow.

tdrauncieal (approved):
LGTM!

ncvicchi (approved):

ncvicchi:
LGTM!

Leoquicenoz (approved):

src/shared_modules/indexer_connector/src/indexerConnector.cpp
```diff
@@ -14,9 +14,13 @@
 #include "loggerHelper.h"
 #include "secureCommunication.hpp"
 #include "serverSelector.hpp"
+#include "keyStore.hpp"
 #include <fstream>
 
 constexpr auto NOT_USED {-1};
+constexpr auto INDEXER_COLUMN {"indexer"};
+constexpr auto USER_KEY {"username"};
```
ncvicchi, line 22: If the key is "username" the key name should be "USERNAME_KEY" for coherence

Issue #20851, opened 2023-12-14
default group merged.mg error


IOTech17:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.7.0 | Group management | Manager | officiel docker image | n/a |

I have upgrade from Wazuh docker 4.5.2 to 4.60 then from 4.6.0 to 4.7.0

When checking the merged.mg file of the default group I get an error that the file contains an error

error on line 1 at column 847: xmlParseEntityRef: no name

![image](https://github.com/wazuh/wazuh/assets/68947790/4fde4d85-dd26-4d7e-8d71-b26d65712820)

Is this a bug or normal?

Another question is there a way to make sure any news group created get all the files the default has?

![image](https://github.com/wazuh/wazuh/assets/68947790/b48c9d66-8820-41c7-af48-119d0bb38eb3)

Thank you in advance,

rauldpm:
Investigated in https://github.com/wazuh/wazuh-docker/issues/1154

Reproduced through a 4.7.1 Wazuh installation assistant deployment, not related to Wazuh docker, added to @wazuh/core initially to triage it

Selutario:
## Update

After taking a look at the code and with the help of @Desvelao, these are the conclusions:

The dashboard is sending a `GET /groups/default/files/merged.mg/xml` API request to fetch the XML content of the `merged.mg` file. However, only `agent.conf` files are returned as XML. All others are parsed to JSON:
https://github.com/wazuh/wazuh/blob/037ede2424c1c4402068bc0a1651836f732ccc70/framework/wazuh/core/configuration.py#L819-L829

There is an XML validator in the dashboard, which is the one that shows the error message since the content obtained is JSON format.

## Proposal

As I mentioned, except for the `agent.conf` files, all others are transformed into JSON even if the `GET /groups/{group_id}/files/{file_name}/xml` endpoint is used. Also, `merged.mg` is a compressed file and is being treated as an `rcl`, so what it returns is wrong. 

We need to rethink and refactor these endpoints:
- [GET /groups/{group_id}/files/{file_name}/xml](https://documentation.wazuh.com/current/user-manual/api/reference.html#operation/api.controllers.agent_controller.get_group_file_xml)
- [GET /groups/{group_id}/files/{file_name}/json](https://documentation.wazuh.com/current/user-manual/api/reference.html#operation/api.controllers.agent_controller.get_group_file_json)

[...]

PR #21653, opened 2024-01-29
Merge group files endpoints

GGP1:
|Related issue|
|---|
| Closes https://github.com/wazuh/wazuh/issues/20851 |

## Description

Merges the `/groups/:group_id/files/:file_name/json` and `/groups/:group_id/files/:file_name/xml` endpoints into a single one that takes the `raw` parameter which defines whether the response content will be formatted in json or plain text. The response content-type is defined based on the file extension retrieved.

## Logs/Alerts example

<details><summary>Get merged.mg in plain text</summary>

`GET /groups/default/files/merged.mg?raw=true`

```
#default
!228 ar.conf
restart-ossec0 - restart-ossec.sh - 0
restart-ossec0 - restart-ossec.cmd - 0
restart-wazuh0 - restart-ossec.sh - 0
restart-wazuh0 - restart-ossec.cmd - 0
restart-wazuh0 - restart-wazuh - 0
restart-wazuh0 - restart-wazuh.exe - 0
!76 agent.conf
<agent_config>

  <!-- Shared agent configuration here -->

</agent_config>
!28411 cis_apache2224_rcl.txt
...
```

</details>

<details><summary>Get merged.mg in JSON</summary>

`GET /groups/default/files/merged.mg`

```json
{
    "data": [
        {
            "file_name": "ar.conf",
            "file_size": 228,
            "file_content": "restart-ossec0 - restart-ossec.sh - 0\nrestart-ossec0 - restart-ossec.cmd - 0\nrestart-wazuh0 - restart-ossec.sh - 0\nrestart-wazuh0 - restart-ossec.cmd - 0\nrestart-wazuh0 - restart-wazuh - 0\nrestart-wazuh0 - restart-wazuh.exe - 0\n"
        },
        {
            "file_name": "agent.conf",
            "file_size": 76,
            "file_content": "<agent_config>\n\n  <!-- Shared agent configuration here -->\n\n</agent_config>\n"
        },
        ...
 	]
}
```

</details>

<details><summary>Get file content using a configuration type</summary>

`GET /groups/default/files/agent.conf?type=conf`

```json
{
    "data": {
        "total_affected_items": 1,
        "affected_items": [
            {
                "filters": {},
                "config": {}
            }
        ]
    },
    "error": 0
}
```

</details>

> Failed checks are not related to the changes introduced.

### DIFF
--- a/framework/wazuh/core/configuration.py
+++ b/framework/wazuh/core/configuration.py
@@ -678,7 +678,7 @@ def get_agent_conf(group_id: str = None, offset: int = 0, limit: int = common.DA
     filename : str
         Name of the file to get. Default: 'agent.conf'
     return_format : str
-        Return format.
+        Response content format.
 
     Raises
     ------
@@ -703,9 +703,9 @@ def get_agent_conf(group_id: str = None, offset: int = 0, limit: int = common.DA
 
     try:
         # Read RAW file
-        if filename == 'agent.conf' and return_format and 'xml' == return_format.lower():
-            with open(agent_conf, 'r') as xml_data:
-                data = xml_data.read()
+        if filename == 'agent.conf' and return_format and return_format.lower() == 'plain':
+            with open(agent_conf, 'r') as raw_data:
+                data = raw_data.read()
                 return data
         # Parse XML to JSON
         else:
@@ -771,7 +771,7 @@ def get_agent_conf_multigroup(multigroup_id: str = None, offset: int = 0, limit:
 
 
 def get_file_conf(filename: str, group_id: str = None, type_conf: str = None, return_format: str = None) -> dict:
-    """Return the configuration file as dictionary.
+    """Return the configuration file content.
 
     Parameters
     ----------
@@ -779,10 +779,10 @@ def get_file_conf(filename: str, group_id: str = None, type_conf: str = None, re
         ID of the group with the file we want to get.
     filename : str
         Name of the file to get.
-    return_format : str
-        Return format.
     type_conf : str
         Type of the configuration we want to get.
+    return_format : str
+        Response content format.
 
     Raises
     ------
@@ -796,7 +796,7 @@ def get_file_conf(filename: str, group_id: str = None, type_conf: str = None, re
     Returns
     -------
     dict
-        Configuration file as dictionary.
+        File content as plain text or dictionary.
     """
     if not os_path.exists(os_path.join(common.SHARED_PATH, group_id)):
         raise WazuhResourceNotFound(1710, group_id)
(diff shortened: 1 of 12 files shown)


EduLeon12 (approved):
Great work @GGP1

 LGTM!,

My only question is if it would be better for readibility purposes to change the raw default value to a bool `False` instead of `None.

GGP1:
> My only question is if it would be better for readibility purposes to change the raw default value to a bool `False` instead of `None.

I've seen both in the code, I don't think there's a standard but I agree that using `False` is slightly better so I modified it. Thanks!

framework/wazuh/core/configuration.py
```diff
@@ -770,33 +820,33 @@ def get_agent_conf_multigroup(multigroup_id: str = None, offset: int = 0, limit:
...
         Group was not found.
     WazuhError(1006)
-        agent.conf does not exist or there is a problem with the permissions.
+        The file does not exist or there is a problem with the permissions.
     WazuhError(1104)
         Invalid file type.
 
     Returns
     -------
     dict
```
Selutario, line 848:
```suggestion
    dict or str
```

Selutario (changes requested):
Looks good! Just some minor changes were requested. Also, I found a strange behaviour. I guess this is expected, but I would like to confirm.

I tried creating a `test.json` file. When the content-type returned is `application/json`, it is seen this way:

> `curl -k -X GET "https://localhost:55000/groups/default/files/test.json?raw=true" -H  "Authorization: Bearer $TOKEN" -v`

image

However, this does not happen when xml content is returned as `application/xml`:
image

Nor when the json file is returned as plain :thinking::
image

GGP1:
> Looks good! Just some minor changes were requested. Also, I found a strange behaviour. I guess this is expected, but I would like to confirm.
> 
> I tried creating a `test.json` file. When the content-type returned is `application/json`, it is seen this way:
> 
> > `curl -k -X GET "https://localhost:55000/groups/default/files/test.json?raw=true" -H  "Authorization: Bearer $TOKEN" -v`
> 
> image

This is expected, the content of the file is returned as a string, and if the Content-Type is `json` then it will be interpreted as a json string instead of an object. This is why new lines are present and the double quotes are escaped. The other content types do not behave the same because they don't need to escape or wrap anything (and thus the output makes it seem like the actual file content).

GGP1:
> Looks good! Just some minor changes were requested. Also, I found a strange behaviour. I guess this is expected, but I would like to confirm.
> 
> I tried creating a `test.json` file. When the content-type returned is `application/json`, it is seen this way:
> 
> > `curl -k -X GET "https://localhost:55000/groups/default/files/test.json?raw=true" -H  "Authorization: Bearer $TOKEN" -v`
> 
> image

@Selutario This is expected, the content of the file is returned as a string, and if the Content-Type is `json` then it will be interpreted as a json string instead of an object.

In order to include all the content inside the string, new lines have to be literal and double quotes have to be escaped. The other content types do not behave the same because they don't need to escape or wrap anything (and thus the output is the actual file content).

Selutario (approved):
LGTM. I think we'll need to update the documentation too, at least the [RBAC reference](https://documentation.wazuh.com/current/user-manual/api/rbac/reference.html). 

Please, open a PR for this @GGP1

Issue #21686, opened 2024-01-31
Vulnerability Scanner - Amazon version matcher has bad logic


tsarquis88:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.8.0 | Vuln. Scanner version matcher | Manager | - | - |

## Description

During the #21677 implementation, it has been discovered that the method `VersionObjectAmzn::match()` contains an incorrect logic that makes the matcher to always return true.

The error seems to be here
https://github.com/wazuh/wazuh/blob/ff9570750a675a3ff245ae98e7ce9e68b8d2a044/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectAmzn.hpp#L59
where, if `std::regex_search()` return false (no match), the method ends up returning true, as if a match happened.

## DoD

- [x] Correct method logic
- [x] Add UTs

tsarquis88:
# Update

PR raised with the fix: #21697

PR #21697, opened 2024-02-01
Vulnerability Scanner - Fix bad version match logic for Amazon

tsarquis88:
|Related issue|
|---|
| #21686 |

## Description

This issue fixes a bad logic in the match method for the Amazon versions. In addition, the regex is also improved.

## Results

The previous UT keep passing OK, and new tests for invalid versions were added.

I've also checked that all the normalized versions of the ALAS feeds pass this validation.

### DIFF
--- a/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectAmzn.cpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectAmzn.cpp
@@ -11,4 +11,4 @@
 
 #include "versionObjectAmzn.hpp"
 
-std::regex VersionObjectAmzn::m_parserRegex(R"((\d+:)?(\d+)\.?(\d+)?\.?(\d+)?-(\d+)?)");
+std::regex VersionObjectAmzn::m_parserRegex(R"((\d+:)?(\d+)\.?(\d+)?\.?(\d+)?-?(\d+)?)");
--- a/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectAmzn.hpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectAmzn.hpp
@@ -56,19 +56,19 @@ public:
     static bool match(const std::string& version, Amzn& output)
     {
         std::smatch parserMatches;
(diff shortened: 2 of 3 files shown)


src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectAmzn.cpp
```diff
@@ -11,4 +11,4 @@
 
 #include "versionObjectAmzn.hpp"
 
-std::regex VersionObjectAmzn::m_parserRegex(R"((\d+:)?(\d+)\.?(\d+)?\.?(\d+)?-(\d+)?)");
+std::regex VersionObjectAmzn::m_parserRegex(R"((\d+:)?(\d+)\.?(\d+)?\.?(\d+)?-?(\d+)?)");
```
tsarquis88, line 14: Note: Since the group after the dash (`-`) is optional, the dash should be optional as well.

src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectAmzn.hpp
```diff
@@ -56,19 +56,19 @@ class VersionObjectAmzn final : public IVersionObject
     static bool match(const std::string& version, Amzn& output)
     {
         std::smatch parserMatches;
-        if ((std::regex_search(version, parserMatches, m_parserRegex) == false) || (parserMatches.size() != 5))
```
tsarquis88, line 59: Note: Demanding the presence of the 5 regex groups is not necessary in my opinion:
- The regex is already demanding the presence of one group. If the five groups are necessary, why not modifying the regex to demand all of them?
- A lot of versions of ALAS with which we are working won't match the 5 groups. We'd discard a lot of cases.

I've personally checked that all the ALAS versions pass this match method. Maybe in the future we could make this more restrictive.

src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectAmzn.cpp
tsarquis88: The same happens with the previous dots (`.`): `\.?`.

Damian-Mangold (approved):
LGTM !

PR #21885, opened 2024-02-12
Fixed hexa decode helper and added router's callback execution safeguard

JavierBejMen:
This PR fixes:
- The decode hexa helper to ensure only valid ASCII characters are decoded, failing otherwise.
- Adds safeguard exception handling to the API tester callback execution inside the worker threads.


JcabreraC (approved):
LGTM !

PR #21932, opened 2024-02-15
Remove depecrated linker option on macOS 13 Ventura 

### LINKED ISSUE #21830: MacOS 13 linker parameter fail
|---|---|---|---|---|
| 4.8.0 | Wazuh component | Agent | Sources | Mac OS 13 |

## Description
In the context of the tests carried out when creating/modifying a PR, this bad behavior arises in the `build-ventura` step, which at the time of compilation presents a problem with a linker parameter 
```shell
ld: unknown options: -noall_load
``` 

I share part of the output log
```shell
Makefile:2602: warning: overriding commands for target `win32/ui_resource.o'
Makefile:2542: warning: ignoring old commands for target `win32/ui_resource.o'
Makefile:2605: warning: overriding commands for target `win32/auth_resource.o'
Makefile:2545: warning: ignoring old commands for target `win32/auth_resource.o'
/Applications/Xcode_15.0.1.app/Contents/Developer/usr/bin/make build_sysinfo build_shared_modules build_syscollector
grep: /etc/os-release: No such file or directory
Makefile:2602: warning: overriding commands for target `win32/ui_resource.o'
Makefile:2542: warning: ignoring old commands for target `win32/ui_resource.o'
Makefile:2605: warning: overriding commands for target `win32/auth_resource.o'
Makefile:2545: warning: ignoring old commands for target `win32/auth_resource.o'
    CC libwazuhext.dylib
ld: unknown options: -noall_load 
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[1]: *** [libwazuhext.dylib] Error 1
make: *** [agent] Error 2
Error: Process completed with exit code 2.
```

tdrauncieal:
|Related issue|
|---|
|Closes #21830 |

## Description
The unknown parameter `-noall_load` reported in the MacOS 13 build (ventura) is a deprecated option even in previous versions (bigsur and moterrey).
While in previous versions quotes were reported as a warning.
image

In this case it has been completely eliminated and is treated as an error.
image

This option has been removed in the creation of `Darwin` systems since it was not being considered anyway.
 
## Analisys
The removed option `-noall_load` is intended to prevent the selective incorporation of references to external objects.

But even in old versions, when this facility was incorporated, the strategy is not correct, since the `-all_load` option includes all objects and does not take into account those declared immediately after.

The correct option to selectively incorporate external objects is using the `-force_load` options next to each object that you want to add.

But unfortunately this option is not available in the repertoire available for `Darwin` systems in [gcc](https://gcc.gnu.org/onlinedocs/gcc-9.5.0/gcc/Darwin-Options.html#Darwin-Options).

I leave [here](https://pewpewthespells.com/blog/objc_linker_flags.html) a reference that explains the behavior of `-all_load` and `-force_load`

  - [ ] Linux
  - [ ] Windows
  - [X] MAC OS X

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -1870,7 +1870,7 @@ ifeq (${uname_S},Darwin)
 WAZUH_SHFLAGS=-install_name @rpath/libwazuhext.$(SHARED)
 
 $(WAZUHEXT_LIB): $(EXTERNAL_LIBS)
-	$(OSSEC_SHARED) $(OSSEC_CFLAGS) $(WAZUH_SHFLAGS) -o $@ -Wl,-all_load $^ -Wl,-noall_load $(OSSEC_LIBS)
+	$(OSSEC_SHARED) $(OSSEC_CFLAGS) $(WAZUH_SHFLAGS) -o $@ -Wl,-all_load $^ $(OSSEC_LIBS)
 else
 ifeq (${TARGET}, winagent)
 $(WAZUHEXT_LIB): $(WAZUHEXT_DLL) $(WAZUHEXT_LIB_DEF)
@@ -2102,7 +2102,7 @@ ifeq (${uname_S},Darwin)
 WAZUH_SHARED_SHFLAGS=-install_name @rpath/libwazuhshared.$(SHARED)
 
 $(WAZUH_LIB): $(WAZUHEXT_LIB) $(AR_PROGRAMS_DEPS)
-	$(OSSEC_SHARED) $(OSSEC_CFLAGS) $(WAZUH_SHARED_SHFLAGS) -o $@ -Wl,-all_load $^ -Wl,-noall_load $(OSSEC_LIBS)
+	$(OSSEC_SHARED) $(OSSEC_CFLAGS) $(WAZUH_SHARED_SHFLAGS) -o $@ -Wl,-all_load $^ $(OSSEC_LIBS)
 else
 ifeq (${TARGET}, winagent)
 $(WAZUH_DLL) $(WAZUH_DEF) : $(WAZUHEXT_DLL) $(AR_PROGRAMS_DEPS) win32/version-dll.o


ncvicchi:
I don't understand your decision to replace "noall_load" with "all_load" instead, since it seems to do exactly the opposite, and it is even discouraged to use:

```code
The -all_load flag tells the linker that it should link every object file (member)
of every static library that is passed to the linker. This is a rather drastic
option and can cause your executable binary to dramatically increase in size.
This flag should be avoided if at all possible when using many static libraries.
```

According to man ld:

```console
-noall_load
             This is the default.  This option is obsolete.
```

Then as I understand the fix approach should be just to delete the flag, not to change it to all_load since this would change the expected linker behaviour.

src/Makefile
```diff
@@ -1870,7 +1870,7 @@ ifeq (${uname_S},Darwin)
 WAZUH_SHFLAGS=-install_name @rpath/libwazuhext.$(SHARED)
 
 $(WAZUHEXT_LIB): $(EXTERNAL_LIBS)
-	$(OSSEC_SHARED) $(OSSEC_CFLAGS) $(WAZUH_SHFLAGS) -o $@ -Wl,-all_load $^ -Wl,-noall_load $(OSSEC_LIBS)
+	$(OSSEC_SHARED) $(OSSEC_CFLAGS) $(WAZUH_SHFLAGS) -o $@ -Wl,-all_load $^ $(OSSEC_LIBS)
```
ncvicchi, line 1873:
```suggestion
	$(OSSEC_SHARED) $(OSSEC_CFLAGS) $(WAZUH_SHFLAGS) -o $@ -Wl, $^ $(OSSEC_LIBS)
```

src/Makefile
```diff
@@ -2102,7 +2102,7 @@ ifeq (${uname_S},Darwin)
 WAZUH_SHARED_SHFLAGS=-install_name @rpath/libwazuhshared.$(SHARED)
 
 $(WAZUH_LIB): $(WAZUHEXT_LIB) $(AR_PROGRAMS_DEPS)
-	$(OSSEC_SHARED) $(OSSEC_CFLAGS) $(WAZUH_SHARED_SHFLAGS) -o $@ -Wl,-all_load $^ -Wl,-noall_load $(OSSEC_LIBS)
+	$(OSSEC_SHARED) $(OSSEC_CFLAGS) $(WAZUH_SHARED_SHFLAGS) -o $@ -Wl,-all_load $^ $(OSSEC_LIBS)
```
ncvicchi, line 2105:
```suggestion
	$(OSSEC_SHARED) $(OSSEC_CFLAGS) $(WAZUH_SHARED_SHFLAGS) -o $@ -Wl, $^ $(OSSEC_LIBS)
```

src/Makefile
tdrauncieal: You didn't understand the change, please take a look again and you will see that the `-all_load` option already existed.

src/Makefile
tdrauncieal: idem.

ncvicchi (approved):
LGTM!

Issue #21907, opened 2024-02-14
Engine - Enhancement of Helper Function Parser in Wazuh-Engine


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0 | Engine | Manager | Packages/Sources | OS version |

### Description

The parser for helper functions within the Wazuh-Engine plays a crucial role in interpreting and executing helper functions written in the assets. This issue aims to align the parser's code with the newly reworked asset builder, focusing on improving code readability, maintainability, and ensuring compliance with new coding standards. Additionally, this task involves adapting existing tests to the new builder code and extensively documenting the parser's operation for both internal developers and end users.

### Objectives

- **Code Quality Improvement**: Enhance the readability and maintainability of the parser code, aligning it with the latest coding standards and practices.
- **Test Adaptation**: Review and adapt existing unit tests to ensure compatibility with the updated parser code and the new builder framework.
- **Documentation**: Create comprehensive documentation detailing how to write helper functions and the supported functionalities, targeting both end users and developers.

### Tasks

- [ ] Refactor the parser code for helper functions to improve readability and maintainability, ensuring alignment with new coding standards.
- [ ] Review existing unit tests for the parser, updating and adapting them as necessary to match the updated parser and builder code.
- [ ] Develop user documentation to be included in the Wazuh-Engine manual, explaining how to write helper functions and outlining the supported functionalities.
- [ ] Document the internal workings of the parser within the code using Doxygen, providing clear insights for developers and contributors.

### Additional Considerations

- **Code Review and Quality Assurance**: Implement a thorough code review process to ensure the refactored parser meets the project's quality standards.
- **Testing Strategy**: Ensure comprehensive test coverage for the parser, including edge cases and error handling scenarios.
- **Documentation Clarity**: Aim for clear, concise, and accessible documentation, catering to both technical and non-technical audiences.

### Acceptance Criteria

- The parser code for helper functions is successfully refactored, demonstrating improved readability, maintainability, and alignment with current coding standards.
- All relevant unit tests are updated and validated to ensure they accurately test the parser's functionality within the context of the new builder framework.
- Comprehensive documentation is available, guiding users on writing helper functions and detailing the parser's supported functionalities.
- Internal code documentation using Doxygen is complete, offering clear guidance for developers working on or with the parser.

JavierBejMen:
## Update
- Exposed each individual parser and added tests, concisely for the following parsers:
[...]

PR #21964, opened 2024-02-19
Engine builder helper parser refactor

JavierBejMen:
|Related issue|
|---|
|#21907|

This PR refactors the Builder helper parsers to expose each individual parser for maintainability and testability.

Also addresses the technical debt of the parsers, now the final HelperToken is returned by all parsers without the previous redundant transformations.

Parsing rules:
When using a helper function in a map or check stage:
```yml
target.field: helper_name(args...)
```

Each argument is tried to be parsed in the following order:
- Quoted argument: `'value'` if it starts with a single quote it parses until the closing single quote, allowed scaped characters are: `\` and `'`
- Reference: `$reference` if it starts with the dollar symbol a reference is parsed, which is any alphanumeric plus the extended symbols `#`, `_`, `@`, and `-` separated by dots.
- JSON value: parses a JSON string, any JSON string with JSON escaping rules.
- Raw string: if the other parsers fail a string value will be assigned, here we allow the escaping of the symbols: `$`, `'`, `,`, ` `, `)`, and `\`.

Invalid escape sequences will always fail.

When parsing a helper function inside a logical check expression the same rules apply adding that at least one argument is expected for the helper, specifying the target field:
```yml
check: helper_name($target.field, args...)
```
Added we can specify comparison helpers as operators:
```yml
check: $target.field<op><value>
```
Where `value` is parsed as a single helper argument following the same escaping rules and order. 
Where `op` is any of the following:
```
==
!=
<
<=
>
>=
```
When using any operator that is not the equality operator only string or integer values are allowed.

When using the default map or filter functions for strings, i.e. direct string mapping or filtering:
```yml
check:
  - target.field: string value
map:
  - target.field: string value
```
If the value starts with the reference symbol ´$` a valid reference is parsed, failing to build otherwise.
If not, a raw string value is assigned, allowing the escaping of the reference symbol only if it appears at the beginning of the string.

### DIFF
--- a/src/engine/ruleset/integrations/apache-http/decoders/apache-access.yml
+++ b/src/engine/ruleset/integrations/apache-http/decoders/apache-access.yml
@@ -48,7 +48,7 @@ normalize:
 
       - source.ip: $source.address
       - _tls: split($network.protocol, 'v')
-      - _tls-1: $_tls.1
+      - _tls_1: $_tls.1
       - _client_ip: split($_forwarded_for, ',')
       - client.ip: $_client_ip.0
       - network.forwarded_ip: $_client_ip.0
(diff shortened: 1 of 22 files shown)


juliancnn (approved):
LGTM!

JcabreraC (approved):
LGTM !

PR #21997, opened 2024-02-21
Resolve transitive mutex locking issue in wazuh-db

### LINKED ISSUE #17737: Agents getting frequently pending and disconnecting and API problems
Hello, We notice an issue with on few Wazuh servers  with versions 4.3.10---4.4.3
with many agents that are pending or getting disconnected. Also there are an issue with api checks that failing a lot.

image

Lately we had to find a solution for a customer who used duplicate images of PC's with wazuh agent installed that he change the pc name later – what made us a problem with duplicate names and refuse from the manager to accept them. 
I've found a way to refresh it by cleaning the client.keys by using the following method

Systemctl stop wazuh-manager
And 
/var/ossec/bin/wazuh-control stop
And 
Rm /var/ossec/etc/client.keys
And 
/var/ossec/bin/wazuh-control start
And 
Systemctl start wazuh-manager

After following this fix on another server we noticed it also fixed issues with agents pending
and we had way more connected clients and it connected lot of the disconnected agents
,and the server work faster without API issues.

By retracing our action I've noticed that on untouched servers the file client.keys have permissions of
rw-r- , while after the delete it have rw-r-r

Also by checking the internal file client.keys on untouched server
we noticed that the file that contains info about all the agents -  have lot of mess
while the list of the agents are not going by order and lot of the agents id's
are not listed in their position. 
For example it goes by id 001,002,140,003,050,004 etc…

We are seeing that from time to time this lists going out of order again even after the fix,
and when it happens it giving us the same problems with pending and disconnected clients and api issues.
After doing the same commands procedure it get temporary fixed.  

Hope you are will able to replicate some of it and fix it.
If more Info is needed - I will try to provide it.

vikman90:
|Related issue|
|---|
|Closes #17737|

This pull request addresses a performance degradation issue in our concurrent database management system caused by transitive mutex locking within the database pool. We have identified that the individual mutex locks were leading to unnecessary blocking, impacting the overall service performance.

To mitigate this issue, we have encapsulated the database pool structure and introduced some key invariants:

- Each node in the pool can represent either an open or closed database connection.
- The pool is designed to never be blocked due to the possession of a mutex by any individual node.

By implementing these changes, we ensure that the database pool operates efficiently even under concurrent access, without encountering unnecessary locking conflicts. This enhancement significantly improves the overall performance and reliability of our database management system.

## Tests

- [X] Unit tests for new pool encapsulation.
- [X] Coverity @ manager.
- [X] Valgrind @ wazuh-db.
- [X] ThreadSanitizer @ wazuh-db.

### DIFF
--- a/src/wazuh_db/wdb.c
+++ b/src/wazuh_db/wdb.c
@@ -337,7 +337,7 @@ wdb_config wconfig;
 rwlock_t pool_mutex;
 wdb_t * db_pool_begin;
 wdb_t * db_pool_last;
-int db_pool_size;
+int wdb_open_count;
 OSHash * open_dbs;
 
 STATIC wdb_t * wdb_get_db_from_pool(const char* db_name) {
@@ -1031,7 +1031,7 @@ void wdb_pool_append(wdb_t * wdb) {
         db_pool_begin = db_pool_last = wdb;
     }
 
-    db_pool_size++;
+    wdb_open_count++;
 
     if (r = OSHash_Add(open_dbs, wdb->id, wdb), r != 2) {
         merror_exit("OSHash_Add(%s) returned %d.", wdb->id, r);
@@ -1052,7 +1052,7 @@ void wdb_pool_remove(wdb_t * wdb) {
             db_pool_last = NULL;
         }
 
-        db_pool_size--;
+        wdb_open_count--;
     } else if (prev = wdb_pool_find_prev(wdb), prev) {
         prev->next = wdb->next;
 
@@ -1060,7 +1060,7 @@ void wdb_pool_remove(wdb_t * wdb) {
             db_pool_last = prev;
         }
 
-        db_pool_size--;
+        wdb_open_count--;
     } else {
         merror("Database for agent '%s' not found in the pool.", wdb->id);
     }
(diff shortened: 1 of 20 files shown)


src/wazuh_db/wdb.c
```diff
@@ -325,269 +325,149 @@ STATIC int wdb_execute_single_int_select_query(wdb_t * wdb, const char *query, i
...
-    rwlock_lock_read(&pool_mutex);
-    if (wdb = (wdb_t *)OSHash_Get(open_dbs, db_name), wdb) {
-        // The corresponding w_mutex_unlock(&wdb->mutex) is called in wdb_leave(wdb_t * wdb)
-        w_mutex_lock(&wdb->mutex);
-        wdb->refcount++;
-    }
-    rwlock_unlock(&pool_mutex);
-    return wdb;
-}
+_Atomic(int) wdb_open_count;
```
cborla, line 329: 👍🏻

cborla (approved):
LGTM

src/wazuh_db/wdb.c
```diff
@@ -1311,36 +1110,25 @@ int wdb_update_last_vacuum_data(wdb_t* wdb, const char *last_vacuum_time, const
...
-        if (node == NULL || db_pool_size <= wconfig.open_db_limit) {
-            rwlock_unlock(&pool_mutex);
+        if (node == NULL) {
             continue;
         }
 
-        w_mutex_lock(&node->mutex);
-
-        if (node->refcount == 0 && !node->transaction) {
-            w_mutex_unlock(&node->mutex);
```
LucioDonda, line 1335: Just a simple doubt, it's ok that bedfore we checked for a refcount == 0 and now ==1 ?

LucioDonda (approved):
LGTM!

TomasTurina (approved):

Issue #22130, opened 2024-02-26
 DB Error Logs in Footprints


rafabailon:
| Wazuh Version | Component | Install Type | Install Method | Platform |
| --- | --- | --- | --- | --- |
| 4.7.3-rc1 | Wazuh DB | Manager | Packages | CentOS |

### Description

During the execution of the footprints for 4.7.3-rc1 the following error logs have appeared

```console
2024/02/25 01:35:40 wazuh-analysisd ERROR: Unable to connect to socket 'queue/db/wdb'.
2024/02/25 01:35:48 wazuh-remoted ERROR: Error querying Wazuh DB to get agent's groups.
2024/02/25 01:35:42 wazuh-analysisd ERROR: dbsync: Cannot communicate with database.
2024/02/25 01:35:48 wazuh-remoted ERROR: Cannot send message: (32) 'Broken pipe'.
2024/02/25 01:35:48 wazuh-remoted ERROR: Connection with wazuh-db lost. Reconnecting.
2024/02/25 01:38:42 wazuh-modulesd ERROR: Agents DB (0) Error querying Wazuh DB to insert vuln_cves
2024/02/25 01:50:27 wazuh-modulesd:task-manager ERROR: (8209): Tasks DB Cannot execute SQL query: err database 'queue/tasks/tasks.db'
2024/02/25 02:25:22 wazuh-authd ERROR: Error querying Wazuh DB to get the agent's 5293 information.
2024/02/25 01:35:21 wazuh-modulesd ERROR: Cannot receive message: Connection reset by peer (104)
2024/02/25 01:36:14 wazuh-remoted WARNING: Unable to save last keepalive and set connection status as pending for agent: 5292
```

### Evidences

The evidences can be found at https://github.com/wazuh/wazuh/issues/22098

TomasTurina:
We have proposed this PR to solve this problem: https://github.com/wazuh/wazuh/pull/22155

The way Wazuh DB cleans open databases has been improved.

PR #22155, opened 2024-02-26
Improve the way Wazuh DB cleans open databases

TomasTurina:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/22130|

## Description

It was discovered that the `wdb_pool_size` variable could contain an invalid value, which could cause an excessive number of file descriptors to be opened and the `wazuh-db` daemon to use excessive memory due to an incorrect filter in the database cleaning. This was because the `wdb_close` function (where `wdb_pool_size` was decremented) is not always called with a valid database pointer, which could lead to decrementing the number of `wdb_pool_size` when it was not necessary.

This PR replaces the `wdb_open_count` variable with `wdb_pool_size`. This new variable corresponds to the number of nodes that are stored in `wdb_pool` where all open objects are stored, whether the database is already open or not.

This new variable is used as a filter in the `wdb_close_old` function instead of `wdb_pool_size`. Now, every time a new node is created/deleted, this new variable is modified, so it will always correspond to the current number of databases stored in memory.

## Tests

  - [x] Linux
  - [x] Windows
  - [x] MAC OS X
- [x] Source installation
- [x] Package installation
- [x] Source upgrade
- [x] Package upgrade

### DIFF
--- a/src/unit_tests/wazuh_db/test_wdb_pool.c
+++ b/src/unit_tests/wazuh_db/test_wdb_pool.c
@@ -30,6 +30,7 @@ static int setup_test(void **state) {
         snprintf(node_name, 10, "node%d", i);
         wdb_t * node = wdb_init(node_name);
         rbtree_insert(wdb_pool.nodes, node_name, node);
+        wdb_pool.size++;
     }
 
     test_mode = 1;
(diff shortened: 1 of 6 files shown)


vikman90 (changes requested):
@TomasTurina I've understood where the issue was. Thanks.

The aspect I'd like to tweak is regarding the `wdb_pool_size` counter. It's equivalent to `rbtree_size(wdb_pool.nodes)` but less costly. However, it jeopardizes encapsulation by allowing someone to write directly to that variable.

My proposal is to incorporate `pool_size` within `wdb_pool`, and have `wdb_pool_size()` as a function,

JcabreraC (approved):
LGTM !

vikman90 (approved):

Issue #22143, opened 2024-02-26
Engine - Enhancement of Date Parsing Functionality in Wazuh-Engine for Predefined Formats and Locale Support


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0 | Engine | Manager | Packages/Sources | OS version |

### Description

The Wazuh-Engine's `parse_date` function currently exhibits limitations in handling predefined date formats and effectively applying locale settings. While this functionality is not actively in use, enhancing it will significantly improve the engine's capability to accurately parse and normalize date strings across various formats and locales. This enhancement aims to extend the `parse_date` function's versatility and reliability, ensuring comprehensive support for predefined formats and locale-specific parsing.

### Enhancement Goals

- **Extend Predefined Format Support**: Ensure comprehensive and accurate support for all predefined date formats as specified in the documentation, leveraging Howard Hinnant's date library.
- **Refine Locale Parameter Handling**: Enhance the application of the `locale` parameter to ensure it accurately influences the parsing process, reflecting the specified locale's conventions.
- **Strengthen Parsing Functionality**: Elevate the `parse_date` function to a more robust and versatile tool capable of handling a wide array of date formats and locale settings.

### Tasks

- [x] Conduct a thorough review of the `parse_date` function to identify current limitations with predefined formats and locale handling.
- [x] Implement enhancements to accurately support all listed predefined formats, ensuring the function adheres closely to Howard Hinnant's date library specifications.
- [x] Refine the handling of the `locale` parameter to ensure its effective application in the parsing process, accommodating various international date formats.
- [x] Create a comprehensive suite of unit tests covering all predefined formats and a range of locale settings to validate the enhanced functionality.
- [x] Update the Wazuh-Engine documentation to reflect the enhanced capabilities of the `parse_date` function, including detailed examples and guidance for users.

### Acceptance Criteria

- The `parse_date` function demonstrates flawless parsing across the full spectrum of predefined formats and effectively utilizes the locale parameter for locale-specific parsing.
- A robust set of unit tests confirms the accuracy and versatility of the enhanced date parsing functionality.
- Documentation accurately describes the enhanced capabilities, providing clear examples and operational guidance for users.

### Additional Information

For custom format specifications, refer to: [Howard Hinnant's Date Library](https://howardhinnant.github.io/date/date.html#from_stream_formatting)

### Example Enhancements

[...]

PR #22203, opened 2024-02-28
Enhancement Parse Date for support format name

NahuFigueroa97:
|Related issue|
|---|
|#22143|

## Summary

This pull request addresses three issues:

1. **Interpretation of Date Format Names:**
   - Previously, date format names were not being interpreted correctly in decoders.
   - The solution involves verifying that the format name is in the list of supported formats and selecting the corresponding format. This required changes in all decoders using the format instead of the name.

2. **Installation Path of tzdb:**
   - The tzdb was being installed in /root/Downloads.
   - The fix involves utilizing the functionality provided by the API to set the installation path. The path is no longer set at the time of download with CPM; instead, the installation path is specified at runtime.

3. **Improved Error Handling for Undefined Locale:**
   - Previously, when the locale was not defined, the error message was challenging to follow.
   - The error message for an undefined locale has been improved to provide better clarity.

## Changes

- Modified decoders to correctly interpret date format names.
- Updated tzdb installation path to use API functionality, allowing runtime specification of the installation path.
- Improved error message for cases where the locale is not defined.

## Implementation Details

- For the interpretation of date format names, changes were made in the logic of each affected decoder. The verification of the format name against the supported list was added.

- The tzdb installation path was updated to use the API's functionality for specifying the installation path at runtime. The changes were applied consistently across the relevant codebase.

- The error handling for undefined locales was improved by adjusting the error message to provide more helpful information.


src/engine/source/cmds/src/start.cpp
```diff
@@ -138,6 +142,22 @@ void runStart(ConfHandler confManager)
...
+            auto& tzdb  = date::reload_tzdb();
+        }
+        catch (const std::exception& e)
+        {
+            LOG_ERROR("An error occurred while reload the tzdb: {}.", utils::getExceptionStack(e));
+            exitHandler.execute();
+            return;
+        }
+    }
+
```
juliancnn, line 160: This block is part of the engine initialization, it should be inside the try, just before or inside the HLP initialization, and it should throw a runtime_exception describing the error without the return.

src/engine/source/hlp/src/parsers/date.cpp
```diff
@@ -89,8 +89,8 @@ SemParser getSemParser(std::string_view targetField,
  * Supported formats, this will be injected by the config module in due time
  */
 const std::vector<std::tuple<std::string, std::string>> TIME_FORMAT {
-    {"ANSIC", "%a %b %d %T %Y"},        // Mon Jan _2 15:04:05 2006
-    {"UnixDate", "%a %b %d %T %Z %Y"},  // Mon Jan _2 15:04:05 MST 2006
+    {"ANSIC", "%a %b %d %T %Y"},        // Mon Jan 2 15:04:05 2006
+    {"UnixDate", "%a %b %d %T %Z %Y"},  // Mon Jan 2 15:04:05 MST 2006
```
juliancnn, line 93: Do not delete the `_`, this marks the optional., %d is the day of the month as a decimal number. The modified command %Nd where N is a positive decimal integer specifies the maximum number of characters to read. If not specified, the default is 2. Leading zeroes are permitted but not required.

juliancnn (changes requested):

juliancnn (approved):
LGTM!

JcabreraC (approved):
LGTM !

Issue #21988, opened 2024-02-20
Some Red Hat / CentOS vulnerabilities contain a release that can't be found in the packages 


pereyra-m:
## Description 

During the analysis of issue #21669, it was found that even after fixing the RPM version matcher and using the updated content (that removes the architecture from the versions), there are still some situations when some false positives could be reported.

We take the same example from the issue where the Firefox package `91.13.0-1.el7.centos` is considered less than `91.13.0-1.el7_9` (https://access.redhat.com/security/cve/CVE-2022-38472, https://access.redhat.com/errata/RHSA-2022:6179)

```
wazuh-modulesd:vulnerability-scanner:packageScanner.hpp:76 operator() : The platform is in the list based on CPE comparison for Package: firefox, Version: 91.13.0-1.el7.centos, CVE: CVE-2022-38472, Content platform CPE: cpe:/o:redhat:enterprise_linux:7 OS CPE: cpe:/o:redhat:enterprise_linux:7
wazuh-modulesd:vulnerability-scanner:packageScanner.hpp:122 operator() : Scanning package - 'firefox' (Installed Version: 91.13.0-1.el7.centos, Security Vulnerability: CVE-2022-38472). Identified vulnerability: Version: 0. Required Version Threshold: 91.13.0-1.el7_9. Required Version Threshold (or Equal): .
wazuh-modulesd:vulnerability-scanner:packageScanner.hpp:174 operator() : Match found, the package 'firefox', is vulnerable to 'CVE-2022-38472'. Current version: '91.13.0-1.el7.centos' (less than '91.13.0-1.el7_9' or equal to ''). - Agent '' (ID: '002', Version: '').
```

This comparison result is the same that we get using `rpmdev-vercmp`

```
[root@localhost vagrant]# rpmdev-vercmp 91.13.0-1.el7.centos 91.13.0-1.el7_9
91.13.0-1.el7.centos < 91.13.0-1.el7_9
```
Because even when the epoch and the version are the same, the release `el7_9` (Red Hat 7.9) is newer than `el_7` (Red Hat 7).

The problem is that the package Firefox `91.13.0-1.el7_9.centos` doesn't exist, even if we have a CentOS 7.9 OS

```
[root@6f6b763e4bc4 /]# cat /etc/centos-release
CentOS Linux release 7.9.2009 (Core)
[root@6f6b763e4bc4 /]# rpm --eval "%dist"
.el7
[...]

PR #22415, opened 2024-03-07
Enhance RPM matcher by excluding `dist` tag from release part

GabrielEValenzuela:
|Related issue|
|---|
|#21988|

## Objective
The aim of this PR is to refine the functionality of the RPM matcher. Following the identification of an issue where the inclusion of the distribution tag in the release part of RPM package versions could lead to inaccurate matching, this update seeks to improve matching accuracy by excluding the distribution tag from the release part during comparisons.

## Description
- An analysis of the RPM matcher revealed that the presence of distribution tags in the release part of package versions was causing mismatches during package comparison operations. These mismatches could potentially lead to incorrect package updates or vulnerability assessments.
- To address this issue, we have implemented changes to the RPM matcher logic, specifically by modifying the algorithm to ignore the distribution tag when creating the object compared to the release part of RPM package versions. 

#### Unit Testing
- Tests were designed to cover a variety of version formats and comparison scenarios, confirming the matcher's improved accuracy and reliability.

#### Integration Testing
N/A

### DIFF
--- a/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectRpm.hpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectRpm.hpp
@@ -330,8 +330,25 @@ public:
         output.version = version.substr(versionStart, versionEnd - versionStart);
         output.release = version.substr(releaseStart);
 
+        // List of known distribution tags
+        // Keep .module always in first place
+        std::vector<std::string> distTags = {
+            ".module", ".el", ".fc", ".amzn", ".centos", ".rhel", ".sles", ".alma", ".mga"};
+
+        // Search and truncate the release string at the first occurrence of any distribution tag
+        for (const auto& tag : distTags)
+        {
+            size_t tagPos = output.release.find(tag);
+            if (tagPos != std::string::npos)
+            {
+                output.release = output.release.substr(0, tagPos);
+                break; // Exit the loop after finding the first tag
+            }
+        }
+
         return true;
     }
+
     /**
      * @brief Constructor.
      *
--- a/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectRpm.hpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectRpm.hpp
@@ -330,20 +330,19 @@ public:
         output.version = version.substr(versionStart, versionEnd - versionStart);
         output.release = version.substr(releaseStart);
 
-        // List of known distribution tags
-        // Keep .module always in first place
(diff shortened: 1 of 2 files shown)


src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectRpm.hpp
sebasfalcone, line 1: Given that we only care about numeric values, can we iterate the string until we find a non-numeric value? (ignoring the dots)

This approach would be something like this:

```C++
    for (auto it = output.release.begin(); it != output.release.end(); ++it)
    {
        if (!std::isdigit(*it) && *it != '.')
        {
            // We found a non-numeric character. We remove from the dot to the end
            output.release.erase(it - 1, str.end());
            break;
        }
    }
```

sebasfalcone: @GabrielEValenzuela Do you think this could present some problem?

src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectRpm.hpp
```diff
@@ -330,8 +330,25 @@ class VersionObjectRpm final : public IVersionObject
         output.version = version.substr(versionStart, versionEnd - versionStart);
         output.release = version.substr(releaseStart);
 
+        // List of known distribution tags
+        // Keep .module always in first place
+        std::vector<std::string> distTags = {
+            ".module", ".el", ".fc", ".amzn", ".centos", ".rhel", ".sles", ".alma", ".mga"};
```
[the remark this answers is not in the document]
GabrielEValenzuela: Removed in the new approach!

src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/versionMatcher/versionObjectRpm.hpp
GabrielEValenzuela: Good catch! I take your idea and make it work with a little change

```cpp

// Search and truncate the release string at the first occurrence of any distribution tag
        auto it = output.release.begin();

        // Iterate through the release string until a non-numeric or non-tag character is found
        while (it != output.release.end() && (std::isdigit(*it) || *it == '.' || *it == '+' || *it == '^'))
        {
            ++it;
        }

        // Truncate the string at the position of the first non-numeric or non-tag character found
        if (it != output.release.end())
        {
            output.release.erase(it - 1, output.release.end());
        }

```

GabrielEValenzuela: Done!

Issue #22367, opened 2024-03-05
Migrate Filebeat templates and adapt the Filebeat module


jotacarma90:
| Related Issue|
|---|
| [Wazuh packages redesign tier 1 - Stage 2 - Migrate and adapt Wazuh subsystem repositories](https://github.com/wazuh/wazuh/issues/21755) |

# Description

Hello team,

This issue is part of the project to migrate all the package tools from the wazuh-packages repository to the wazuh/wazuh repository.
In this issue, we are going to make all the necessary modifications to move the Filebeat files needed to creation of Filebeat package.

## Tasks

- [x] Research and set ETA for the parent issue (notify @rauldpm)

## Tasks (templates)

- [x] Research changes related to Filebeat templates
- [x] Migrate Filebeat templates from wazuh-packages to wazuh/wazuh
- [x] Adapt changes related to Filebeat templates

## Tasks (module)

- [x] Manual package generation
  - Check the [README](https://github.com/wazuh/wazuh/blob/master/extensions/filebeat/7.x/wazuh-module/README.md) file 
- [x] Package creation through CI
  - Packages that require a sign during the build phase must implement this functionality (for example Windows and macOS)
- [x] Add CI testing
- [x] Upload packages to S3 through CI. This must be done after the testing phase

rauldpm:
After a meeting with the teams involved, it has been decided the following:

- Wazuh Filebeat module will be versioned in a tier 2

jotacarma90:
## Update 15/03/2024
- Research usefulness of filebeat module and all its files.
- Working on the workflow related to the creation of the Filebeat package, steps in this readme: https://github.com/wazuh/wazuh/blob/master/extensions/filebeat/7.x/wazuh-module/README.md
- Thread on some related comments: https://wazuh-team.slack.com/archives/C038AR25C8N/p1710428187030819

rauldpm:
Missing CI test, please update project fields and complete the missing tasks

PR #22558, opened 2024-03-15
Migrate Filebeat module to wazuh/wazuh repository

jotacarma90:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/22367|

## Description

In this PR we are migrating the Filebeat template and creating a workflow that will generate the corresponding tar file for the Filebeat module.

### DIFF
--- a/.github/workflows/packages-filebeat.yml
+++ b/.github/workflows/packages-filebeat.yml
@@ -1,0 +1,58 @@
+name: Packages - Build filebeat package
+on:
+  workflow_dispatch:
+    inputs:
+      revision:
+        description: 'Revision used to naming Filebeat package wazuh-filebeat-X.X.tar.gz'
+        required: true
+
+jobs:
+  Filebeat-module-generation:
+    runs-on: ubuntu-latest
+
+    steps:      
+      - name: Cancel previous runs
+        uses: fkirc/skip-duplicate-actions@master
+        with:
+          cancel_others: 'true'
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          skip_after_successful_duplicate: 'false'
+
+      - uses: actions/checkout@v4
+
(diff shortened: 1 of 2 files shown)


.github/workflows/packages-filebeat.yml
```diff
@@ -0,0 +1,58 @@
+name: Packages - Build filebeat package
+on:
+  workflow_dispatch:
+    inputs:
+      revision:
+        description: 'Revision used to naming Filebeat package wazuh-filebeat-X.X.tar.gz'
```
MarcelKemp, line 6: I think it would be more correct to use `version`, rather than `revision`, as _revision_ would be in the case after _version_. Example: `wazuh-filebeat-X.X`**`-X`**
```suggestion
      version:
        description: 'Version used to naming Filebeat package wazuh-filebeat-X.X.tar.gz'
```

.github/workflows/packages-filebeat.yml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          cancel_others: 'true'
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          skip_after_successful_duplicate: 'false'
+
+      - uses: actions/checkout@v4
+
+      - name: Set package name
+        run: |
+          echo "FILEBEAT_TAR_NAME=wazuh-filebeat-${{ github.event.inputs.revision }}.tar.gz" >> $GITHUB_ENV
```
MarcelKemp, line 25:
```suggestion
          echo "FILEBEAT_TAR_NAME=wazuh-filebeat-${{ github.event.inputs.version }}.tar.gz" >> $GITHUB_ENV
```

MarcelKemp (changes requested):

MarcelKemp (approved):

Issue #22567, opened 2024-03-15
API cache option deprecation warning always appears even if it is not explicitly enabled


Rebits:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
|4.8.0-beta4 | API | Manager | Packages | centOS7 |

### Description

During https://github.com/wazuh/wazuh/issues/22547 it has been detected that warning messages concerning the deprecation of the [cache option](https://github.com/wazuh/wazuh/pull/22387) always appear even if the cache option is not enabled by the user

```
2024/03/15 14:00:01 WARNING: The `cache` API configuration option was deprecated in 4.8.0 and will be removed in the next minor release.
```

We should consider avoiding showing this message if the user does not enable it explicitly

RamosFe:
## Update

Added warning for only the cases where the user added the `cache` API option.

https://github.com/wazuh/wazuh/blob/be2f1aafd28465c8581dcddfeca75cd8c54a8085/api/api/configuration.py#L314-L316

And created a new unit test for testing this specific case:
https://github.com/wazuh/wazuh/blob/be2f1aafd28465c8581dcddfeca75cd8c54a8085/api/api/test/test_configuration.py#L159-L174

Results:
```
python3 -m pytest api/api/test/test_configuration.py 
========================================================================================================== test session starts ==========================================================================================================
platform linux -- Python 3.10.12, pytest-7.3.1, pluggy-1.4.0
rootdir: /home/federamos/Documents/Wazuh/Repositories/wazuh/api/api
configfile: pytest.ini
plugins: asyncio-0.18.1, html-2.1.1, metadata-3.1.1, tavern-1.23.5, anyio-4.3.0, trio-0.8.0
asyncio: mode=auto
collected 49 items                                                                                                                                                                                                                      

api/api/test/test_configuration.py .................................................                                                                                                                                              [100%]

=========================================================================================================== warnings summary ============================================================================================================
api/api/validator.py:11
api/api/validator.py:11
  /home/federamos/Documents/Wazuh/Repositories/wazuh/api/api/validator.py:11: DeprecationWarning: Accessing jsonschema.draft4_format_checker is deprecated and will be removed in a future release. Instead, use the FORMAT_CHECKER attribute on the corresponding Validator.
    from jsonschema import draft4_format_checker

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
[...]

PR #22587, opened 2024-03-18
Changed cache API option warning logic.

RamosFe:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/22567|

## Description

Changed the logic of the warning related to the `cache` API option. Now the warning only appears when the option is added to the configuration by the user-

### DIFF
--- a/api/api/configuration.py
+++ b/api/api/configuration.py
@@ -6,6 +6,7 @@ import copy
 import datetime
 import os
 from typing import Dict, Tuple, Any, List
+import logging
 
 import yaml
 from cryptography import x509
@@ -20,6 +21,9 @@ import wazuh.core.utils as core_utils
 from api.api_exception import APIError
 from api.constants import CONFIG_FILE_PATH, SECURITY_CONFIG_PATH, API_SSL_PATH
 from api.validator import api_config_schema, security_config_schema
+
+CACHE_DEPRECATED_MESSAGE = 'The `cache` API configuration option was deprecated in {release} and will be removed ' \
+                           'in the next minor release.'
 
 default_security_configuration = {
     "auth_token_exp_timeout": 900,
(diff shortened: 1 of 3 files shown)


fdalmaup (approved):
The failed check passed in Github Actions.

LGTM!

api/api/configuration.py
```diff
@@ -305,6 +310,11 @@ def replace_bools(conf: dict):
     else:
         # If any value is missing from user's configuration, add the default one:
         dict_to_lowercase(configuration)
+
+        # Check if cache is enabled
+        if configuration.get('cache', {}).get('enabled', {}):
+            logger.warning(CACHE_DEPRECATED_MESSAGE.format(release="4.8.0"))
```
Selutario, line 316: The logger is only used under a specific condition, I would instantiate it here instead:
```suggestion
        # Check if cache is enabled
        if configuration.get('cache', {}).get('enabled', {}):
            logger = logging.getLogger('wazuh-api')
            logger.warning(CACHE_DEPRECATED_MESSAGE.format(release="4.8.0"))
```

Selutario (changes requested):
Just a minor change. 

There is a problem though: the logger is not yet configured by the time the warning is printed. Therefore, the warning is printed in the stderr, but not in the `api.log` file:
image



I don't think we can fix this since the `api.yaml` options must be read in order to configure the logger.

api/api/configuration.py
```diff
@@ -285,6 +289,7 @@ def replace_bools(conf: dict):
                         conf[k] = True
                     elif conf[k].lower() == 'no':
                         conf[k] = False
+    logger = logging.getLogger('wazuh-api')
```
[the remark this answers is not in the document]
RamosFe: Changed in commit bf7a24717e541a34035d81c8a1124eba363c2611.

api/api/configuration.py
RamosFe: Changed in commit bf7a24717e541a34035d81c8a1124eba363c2611.

Selutario (approved):
LGTM

Issue #22570, opened 2024-03-15
Unit tests for vulnerability scanner don't pass 


cborla:
|Wazuh version|Component|Install type|Install method|Platform|Source|Previous|
|---|---|---|---|---|--|--|
| 4.8.0-beta4 | Unit tests | Manager | Sources | Ubuntu 22 | #22527 | #21819|

### Commands

```shell
make deps
make TARGET=server DEBUG=1 TEST=1
./wazuh_modules/vulnerability_detection/test_wm_vulnerability_detection 
```

### Result

Case 1:
```
[==========] tests_configuration: Running 12 test(s).
[ RUN      ] test_read_custom_cti_url_full_configuration
2024/03/15 12:01:07 unset: ERROR: (1226): Error reading XML file 'etc/ossec.conf':  (line 0).
```

Case 2:
```
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
```

[result-ctest.txt](https://github.com/wazuh/wazuh/files/14617275/result-ctest.txt)

MiguelazoDS:
Unable to reproduce this error with ctest 

![image](https://github.com/wazuh/wazuh/assets/13010397/13bdd96c-57dc-4329-bc67-90263441828b)
![image](https://github.com/wazuh/wazuh/assets/13010397/186d8fd1-5d62-4f14-9272-885aca65733c)

The issue https://github.com/wazuh/wazuh/issues/22527 reports that there are no errors for test_wm_vulnerability_detection what mean that running with ctest passed. 

![image](https://github.com/wazuh/wazuh/assets/13010397/6993a78a-f9cb-47f5-ac8f-db264ab5838e)

From a random PR the test are executed with ctest 
https://github.com/wazuh/wazuh/pull/22621/checks?check_run_id=22906056859

There are errors for vulnerability_scanner_unit_tests (but they are not the same as test_wm_vulnerability_detection)

![image](https://github.com/wazuh/wazuh/assets/13010397/02fa23a2-024a-4f4e-8a8e-7876db2027e2)

MiguelazoDS:
# Update 

Unable to reproduce vulnerability_scanner_unit_tests failing test 

[...]

PR #22644, opened 2024-03-23
Added UNIX socket path boundary check

MiguelazoDS:
|Related issue|
|---|
|#22570|

## Description

This PR adds a check for socket path length, the test is a dummy check that does not prove any functionality. The tested function should use the data context and then update the test case.

### DIFF
--- a/src/shared_modules/utils/socketWrapper.hpp
+++ b/src/shared_modules/utils/socketWrapper.hpp
@@ -84,7 +84,7 @@ public:
         m_unixAddr.sun_family = AF_UNIX;
         if (path.size() >= sizeof(m_unixAddr.sun_path))
         {
-            throw std::runtime_error {"Error setting socket path too long"};
+            throw std::runtime_error {"Error setting socket path (too long)"};
         }
 
         std::copy(path.begin(), path.end(), m_unixAddr.sun_path);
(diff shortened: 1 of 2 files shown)


GabrielEValenzuela (approved):
LGTM!

sebasfalcone (approved):

sebasfalcone (approved):
After these changes, the tests that failed before gave us this error:
```
[----------] 2 tests from QueryAllPackagesTest
[ RUN      ] QueryAllPackagesTest.SingleDeleteAndInsertTest
unknown file: Failure
C++ exception with description "Error setting socket path (too long)" thrown in the test body.
Error removing FD from interface.
[  FAILED  ] QueryAllPackagesTest.SingleDeleteAndInsertTest (94 ms)
[ RUN      ] QueryAllPackagesTest.InsertAllTest
unknown file: Failure
C++ exception with description "Error setting socket path (too long)" thrown in the test body.
Error removing FD from interface.
[  FAILED  ] QueryAllPackagesTest.InsertAllTest (84 ms)
[----------] 2 tests from QueryAllPackagesTest (179 ms total)
```

There is a restriction in the [UNIX socket configuration structure](https://www.man7.org/linux/man-pages/man7/unix.7.html), the path to it cannot have more than 108 characters. We weren't checking for boundaries before, that's what ASAN detected.

But. given that this is an OS restriction, it's considered as an expected behaviour (this won't be a problem in the normal working of the product because the socket lives in a relatively short path (`/var/ossec/queue`)

Issue #22679, opened 2024-03-27
Engine - Enhance Schema Validator Issue with Array Elements in Assets


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0 | Engine | Manager | Packages/Sources | OS version |

**Description:**

We've identified a critical issue within the Wazuh-Engine's asset schema validator. When accessing an element of an array that is part of a schema, for example, using `host.related.ip.0`, the validator incorrectly processes the request. This is because, in the schema, `host.related.ip` is defined as an object, but it is actually an array of objects. This discrepancy leads to validation errors, impacting the correct functioning of asset management and potentially leading to data inconsistencies.

**Objective:**

To resolve this issue, we need to enhance the schema validator to accurately recognize and validate when an array element, as part of a larger schema, is accessed. This will ensure the integrity of asset data and allow for precise data manipulation and retrieval.

**Expected Outcomes:**

- Correct validation of schema elements that are arrays of objects, eliminating the current discrepancy and avoiding validation errors.
- Improved accuracy in asset management functionalities that depend on accessing specific elements within arrays in the schema.

**Tasks:**

- [x] Review and identify the limitations in the current schema validator that lead to improper validation of array elements.
- [x] Develop a solution to enable the schema validator to accurately identify and validate access to elements within arrays, considering them as objects.
- [x] Implement comprehensive tests to ensure that accessing array elements within assets is correctly validated against the schema.
- [x] Update documentation to reflect the improved schema validation capabilities and provide examples for correctly accessing elements within arrays.

By addressing this validation issue, we will significantly improve the reliability and usability of asset management in Wazuh-Engine, ensuring that users can accurately and efficiently work with complex data structures within assets.

PR #22733, opened 2024-04-04
Updated Engine's schemf module to support array paths

JavierBejMen:
|Related issue|
|---|
|#22679|

This PR adds to the schemf module the ability to query array items types of schema fields that are arrays.


src/engine/source/schemf/src/schema.cpp
```diff
@@ -118,16 +137,48 @@ const Field& Schema::get(const DotPath& name) const
...
-            return false;
+            if (!isParentSchema)
+            {
+                return false;
+            }
+
+            throw std::runtime_error(fmt::format("Field '{}' does not exist in '{}'", it->data(), name.str()));
         }
 
+        isParentSchema = true;
```
NahuFigueroa97, line 154: question: Why do you set this variable and then not use it?

src/engine/source/schemf/src/schema.cpp
```diff
@@ -98,7 +98,26 @@ const Field& Schema::get(const DotPath& name) const
 
         if (it != name.cend() - 1)
         {
-
+            // Handle arrays e.g. "field/0"
+            // If the field is an array and the next part is the last one and a number, return a new field with the
+            // array type
+            if (entry->second.isArray() && (it + 1) == (name.cend() - 1))
```
NahuFigueroa97, line 104: improve-desing: I notice that the first part of this block is repeated for both get() and has_field(), could you modulate it a little more?

src/engine/source/schemf/test/src/component/schema_test.cpp
```diff
@@ -141,3 +149,18 @@ INSTANTIATE_TEST_SUITE_P(SchemaTest,
                                            LoadTuple(R"({"a": {}})", false),
                                            LoadTuple(R"({"a": [{"type": "text"}]})", false),
                                            LoadTuple(R"([{"a": {"type": "keyword"}}])", false)));
+
+TEST(SchemaTest, ArrayItem)
```
NahuFigueroa97, line 153: Is this a particular case? If it is just a case, could you document the test?

src/engine/source/schemf/src/schema.cpp
JavierBejMen: It is used

src/engine/source/schemf/test/src/component/schema_test.cpp
JavierBejMen: I think the test is self documented. It just tests array items get and exists.

src/engine/source/schemf/src/schema.cpp
JavierBejMen: The little details makes it not so simple to be the same for both, and it its only used on those two functions.

NahuFigueroa97 (approved):
LGTM!

juliancnn (approved):
LGTM!

JcabreraC (approved):
LGTM !

Issue #22864, opened 2024-04-10
Add Integration test for SQLite synchronization policy


cborla:
|Wazuh version|Component|Install type|
|---|---|---|
| 4.9.0 (dev) | Wazuh-DB | Manager |

## Description
The purpose of this issue is to ensure that the wazuhdb global database synchronization is set to NORMAL synchronization mode at the time of use.

## Task
The task involves implementing `PRAGMA synchronous` on the global database upon each opening.

```
  input: global sql PRAGMA synchronous
  output: ok [{"synchronous":2}]
```
Related to:
- [Improve Wazuh-DB performance by adjusting SQLite synchronization policy](https://github.com/wazuh/wazuh/issues/22753#top)

cborla:
Since we intend to avoid adding the test in version 4.8.0, because all integration tests are being migrated to 4.9.0, we chose to include the integration test directly in version 4.9.0 which does not yet contain the change of the synchronization mode.

Warning: when performing the branch maintenance, in the merge process from 4.8.x versions to 4.9.0, the expected value in the ouput must be updated to 1, or NORMAL.

From:
`output: ok [{"synchronous":2}]`

To
`output: ok [{"synchronous":1}]`

PR #22865, opened 2024-04-10
feat: SQLite synchronization mode validate

cborla:
|Related issue|
|---|
|#22864|

## Description

The synchronization mode in which the global db is at the moment of being opened must be validated.

  - [X] Linux
- [X] QA templates contemplate the added capabilities


cborla:
Since we intend to avoid adding the test in version 4.8.0, because all integration tests are being migrated to 4.9.0, we chose to include the integration test directly in version 4.9.0 which does not yet contain the change of the synchronization mode.

Warning: when performing the branch maintenance, in the merge process from 4.8.x versions to 4.9.0, the expected value in the ouput must be updated to 1, or NORMAL.

From:
`output: ok [{"synchronous":2}]`

To
`output: ok [{"synchronous":1}]`

vikman90 (approved):

Issue #22957, opened 2024-04-16
Wazuh manager workflow should specify what is being created instead of using a generic title


rauldpm:
## Description

Testing the Wazuh manager workflow build package, I discovered that the workflow title does not specify what is being created, I executed the following GitHub API calls:

```
gh workflow run packages-build-manager.yml --repo=wazuh/wazuh --ref 'migration/21755-wazuh-packages-redesign' -f revision=999 -f architecture=amd64 -f system=deb -f revision=2 -f is_stage=false -f checksum=true
gh workflow run packages-build-manager.yml --repo=wazuh/wazuh --ref 'migration/21755-wazuh-packages-redesign' -f revision=999 -f architecture=amd64 -f system=deb -f revision=3 -f is_stage=true -f checksum=true
gh workflow run packages-build-manager.yml --repo=wazuh/wazuh --ref 'migration/21755-wazuh-packages-redesign' -f revision=999 -f architecture=x86_64 -f system=rpm -f revision=3 -f is_stage=false -f checksum=false
gh workflow run packages-build-manager.yml --repo=wazuh/wazuh --ref 'migration/21755-wazuh-packages-redesign' -f revision=999 -f architecture=x86_64 -f system=rpm -f revision=3 -f is_stage=true -f checksum=true
```

However I could not see which workflow corresponds to each build, we need to address this to allow the correct usage of the workflow

![image](https://github.com/wazuh/wazuh/assets/14913942/e83367ff-92b1-4a6b-905b-ef7d8c727a6e)

Issue #22955, opened 2024-04-16
The filebeat build module workflow is missing the  workflow_call item


rauldpm:
## Description

The GitHub Action workflow used to build the Filebeat module does not define a `workflow_call` item, this item is needed to be able to use it as a reusable workflow

https://github.com/wazuh/wazuh/blob/migration/21755-wazuh-packages-redesign/.github/workflows/packages-build-manager.yml

The current configuration allows the workflow to be called using the GitHub CLI and the GitHub UI (not yet, we need to merge it)

mjcr99:
## Update
**(17/04/2024)** Opened [PR](https://github.com/wazuh/wazuh/pull/22967) to fix requested changes.

PR #22967, opened 2024-04-17
Apply minor changes in manager and filebeat packages workflows

mjcr99:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/22955 & https://github.com/wazuh/wazuh/issues/22957|

## Description
This PR adds:
- `workflow_call` tag in the `.github/workflows/packages-filebeat.yml` file, letting this workflow be called from another one.
- Dynamic workflow name for the `.github/workflows/packages-build-manager.yml` workflow calls, using `run-name` instead of `name` to set the job execution name. [Reference](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#run-name).

## Test
The Job name changes in every execution depending on the arguments. For example, for `.deb` and `amd64` the job name is set as follows:
image

Manager package
- performed execution with `debug` and `checksum`: https://github.com/wazuh/wazuh/actions/runs/8739953751
- performed execution without `debug` and `checksum`: https://github.com/wazuh/wazuh/actions/runs/8740058785

Filebeat package performed execution: https://github.com/wazuh/wazuh/actions/runs/8737877001

### DIFF
--- a/.github/workflows/packages-filebeat.yml
+++ b/.github/workflows/packages-filebeat.yml
@@ -3,12 +3,17 @@ on:
   workflow_dispatch:
     inputs:
       revision:
-        description: 'Revision used to naming Filebeat package wazuh-filebeat-X.X.tar.gz'
-        required: true
+        description: | 
+          Revision used to naming Filebeat package wazuh-filebeat-X.X.tar.gz.
+          Default 0.0.
+        required: false
+        default: "0.0"
+        type: string
+
   workflow_call:
     inputs:
       revision:
-        required: true
+        required: false
         default: "0.0"
         type: string
     
--- a/.github/workflows/packages-filebeat.yml
+++ b/.github/workflows/packages-filebeat.yml
@@ -1,11 +1,17 @@
-name: Packages - Build filebeat package
+name: Packages - Build Filebeat Package
 on:
   workflow_dispatch:
     inputs:
       revision:
         description: 'Revision used to naming Filebeat package wazuh-filebeat-X.X.tar.gz'
         required: true
-
+  workflow_call:
+    inputs:
+      revision:
+        required: true
+        default: "0.0"
+        type: string
+    
 jobs:
   Filebeat-module-generation:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


MarcelKemp (approved):

.github/workflows/packages-filebeat.yml
```diff
@@ -1,11 +1,17 @@
...
     inputs:
       revision:
         description: 'Revision used to naming Filebeat package wazuh-filebeat-X.X.tar.gz'
         required: true
-
+  workflow_call:
+    inputs:
+      revision:
+        required: true
+        default: "0.0"
```
rauldpm, line 12: If there is a default value why it is required?

rauldpm (changes requested):
Please add a link to each action in the PR to prove that the changes are functional, also, issues do not have any work update

.github/workflows/packages-filebeat.yml
rauldpm: Note: All development packages should use revision 0 (in this case it should be 0.0

mjcr99: This workflow only manages an input value. 

Should another parameter be added to manage the `development/stage` package name or is it still intended to be handled just by the revision field? In this case, I think the set `default` value is correct.

mjcr99: Added b3121120804317e81bbb3eddd49656feb9519540 to fix `required` field.

rauldpm: Ah yes, sorry, the default will set a value in the input field, and if empty, the required will show an error, it is ok

rauldpm: If you set the input as required, the input will ask for a value if the field is empty, if you set the default value, the field will have a value in the UI field, but if you remove it, the input will show an error

If you set required as false and set a default value, then the input field will show the default value, and if you remove it and do not specify a value, then the input will set the default value internally, that is what I am referring, the default value has two functions:

1. Set the input value in the UI
2. Set the input value if the UI field is not set

That is why I am saying that it makes no sense to have a default value with a required input in this case

rauldpm (approved):

PR #23341, opened 2024-05-08
Fix full scan, to exclude the manager scan if the setting is disabled

### LINKED ISSUE #23340: Scan of manager elements occur if the policy is disabled
## Description

This issue aims to solve a problem detected in the last demo testing, basically when a full scan occurs (if we toggle the vd enable/disable setting to force all agent scans) and the manager scan is disabled, the scan for agent 000 occurs ignoring the disable scan manager setting value.

## DoD
- [x] Fix
- [x] Manual testing.
- [x] Add automated test.

Dwordcito:
|Related issue|
|---|
|Closes #23340 |

## Description

This PR aims to fix the full scan to exclude the manager if the "scan manager" setting is disabled.

It checks the current value of the policy setting to decide if we need to scan or not the agent 000.

### DIFF
--- a/etc/internal_options.conf
+++ b/etc/internal_options.conf
@@ -430,7 +430,7 @@ vulnerability-detection.osdata_lru_size=1000
 # Vulnerability detector - Enable or disable the scan manager
 # 0. Enabled
 # 1. Disabled
-vulnerability-detection.disable_scan_manager=0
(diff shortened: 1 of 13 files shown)


MiguelazoDS:
# Revision

image

- Scan by events 

image

- Re-scan 

image

MiguelazoDS (approved):

PR #23384, opened 2024-05-13
Fixes helper builder in wrong stages

JavierBejMen:
This PR enforces helper builders to be called only on the allowed stages, e.g. filter builders can only be defined in stage check or stage blocks. Failing to build the asset otherwise.

Also fixes suricata decoder.


juliancnn (approved):
LGTM

JcabreraC (approved):
LGTM !

PR #23503, opened 2024-05-17
Change the trigger size to run a bulk query, to insert, delete or modify elements

### LINKED ISSUE #23481: Increase the bulk query elements max size for the indexer connector
## Description
This issue aims to increase the maximum size of operations in a bulk query triggered by the indexer connector component.

Currently, this size is 50, and after setting the wait_for policy for each HTTP call, the processing is slower, therefore we will increase this number to avoid making so many calls.

Desired number: 1000 elements per query.

Dwordcito:
|Related issue|
|---|
|Closes #23481 |

# Description
This PR aims to change the value to be used as a queue pop limit we wait 10 seconds or 1000 elements to dispatch a single HTTP call with a bulk query payload.

Fewer HTTP calls are equal to better performance because we wait for a refresh to the next call.


sebasfalcone (approved):
Awesome work!

Issue #23558, opened 2024-05-22
Add python unit tests coverage to github actions


GGP1:
## Description

As part of every release testing, we are calculating the python unit tests coverage manually by executing one of our internal scripts. This has caused us some [issues](https://github.com/wazuh/wazuh/issues/23539) as the results rely on the environment where they are executed, resulting in different coverages depending on who did the tests.

To avoid this and to automate the process, we could edit our existing unit tests (adding `--cov` to the tests command) and have this information in our workflow.

## Checks

The following elements have been updated or reviewed (should also be checked if no modification is required):
- [x] Tests (unit tests, API integration tests).
- [x] Changelog.
- [x] Documentation.
- [x] Integration test mapping (using `api/test/integration/mapping/_test_mapping.py`).

Selutario:
Reopening since it was automatically closed after merging https://github.com/wazuh/qa-integration-framework/pull/168

GGP1:
## Update

Added coverage reporting in the existing Python unit tests.

PR #23585, opened 2024-05-23
Add python unit tests coverage action

GGP1:
|Related issue|
|---|
| Closes https://github.com/wazuh/wazuh/issues/23558 |

## Description

Adds a github action to get the Python unit tests coverage.


nico-stefani (dismissed):

Selutario (changes requested):
GJ @GGP1. I would also add `--cov` to our existing framework, wodle and api unit tests workflows so it always shows the coverage. Even if it is not formatted like a markdown table, it will help us to detect PRs where the coverage decreases.

I guess we'll need to add pytest-cov to the requirements-dev file.

Selutario (approved):

Issue #9913, opened 2021-08-26
Generate Wazuh symbols for macOS, linux and windows separate from the binary file


Dwordcito:
Branch: [enhancement/9913-generate-debug-symbols-epic](https://github.com/wazuh/wazuh/tree/enhancement/9913-generate-debug-symbols-epic)

## Description
This epic aims to generate the debugging symbols files for macOS, Linux, and Windows, having enough information to solve the issues quickly, without having to reproduce issues with a different version than the one that goes out on the production builds.
Apart from this, the epic also aims to be able to carry out an active search for problems in the component that cause crashes and memory problems.
This epic tries to help improve the product at a faster rate!

# Goals
- Always generate debug information.
- Upload debug symbols on each CI compilation.
- Enable dump generation on installer process by default -> with references on documentation (how to disable the dump generation) quick reference.

# Functional requirements
1. **Debug Symbol Generation:** 
   - Debug symbols must be generated for all project components during the build.
   - Debug symbols should include information about functions, variables, and source code lines.
   - Debug symbols must be generated in a format compatible with the target platform (e.g., DWARF for macOS and Linux, PDB for Windows).
   
2. **Automatic Upload:**
   - Debug symbols generated during the CI/CD pipeline must be automatically uploaded to a centralized location.
   - Uploaded symbols should be organized by build version and platform for easy retrieval.
   
3. **Installer Dump Generation:**
   - The installer process must include a feature to generate crash dumps by default.
   - Documentation should clearly explain how to disable dump generation for advanced users.

4. **Crashpad Integration (Second Iteration):**
   - Integrate Crashpad to capture crash reports and generate minidump files.
   - Configure Crashpad to report crashes to external services.

5. **Integrate Crashpad for Advanced Crash Reporting:** In the second phase, integrate Crashpad to enable advanced crash reporting and integration with services.

# Non-functional requirements
1. **Performance:**
   - Debug symbol generation and uploading should not significantly impact execution, build, or deployment times.
   
2. **Reliability:**
   - Debug symbols must be generated consistently and accurately across different platforms.
   - Uploaded symbols should be securely stored and easily accessible for debugging purposes.

# Implementatin restrictions
1. **Platform Compatibility:**
   - Ensure that the chosen debugging symbol format is compatible with target platforms (macOS, Linux, Windows).
   
2. **Documentation:**
   - Properly document enabling and disabling dump generation for users and contributors.

# Plan

### Phase 1

[...]

PR #23609, opened 2024-05-24
Destination path accepts backslash

ncvicchi:
|Related issue|
|---|
|#9913|

The macOS script was not accepting "/tmp" as a destionation parameter as stated in the documentation.
This PR fix this behaviour by sanitizing the parameter, removing any backslash that might be used and using the parameter accordingly.


aritosteles (approved):
LGTM

Issue #23646, opened 2024-05-24
Force no-index when a full rescan is executed


Dwordcito:
## Description

This issue aims to force a no-index feature when a cluster is enabled.

Because we cannot get current node agents, we need to re-scan locally and index this information only when an agent launches a synchronization.

PR #23647, opened 2024-05-24
Force no index on re-scan in cluster environment

sebasfalcone:
|Related issue|
|---|
| #23646 |

## Description
This PR updates the re-scan logic used on the vulnerability scanner to prevent data loss in a cluster environment

If we are running in a cluster environment we shouldn't index the re-scanned data, but relly on the synchronization mechanism between indices


sebasfalcone (commented):
## Testing
- Started with no vulnerabilities

### Cluster disabled 🟢 
- Vulnerabilities indexed
- Used hostname for the index
image
image

### Cluster enabled

- Vulnerabilities not indexed
image

- After sync event
  ```
  2024/05/24 17:32:38 indexer-connector[269310] 
  indexerConnector.cpp:437 at operator()(): DEBUG: Syncing agent 
  '001' with the indexer.
  ```
  image

Dwordcito (approved):

Issue #23648, opened 2024-05-24
Exposure of sensitive information when integratod is configured in debug mode 2


Damian-Mangold:
|Component|Install type|
|---|---|
| integratord | Manager | 

# Description 

When integratord module is set to `debug=2`, all the parameters used are printed to `ossec.log`. The problem is that `integration` configurations could include API keys, which should be private and protected:

```XML
<integration>
  <name>pagerduty</name>
  <api_key>API_KEY</api_key> <!-- Replace with your PagerDuty API key -->
</integration>
```

They are protected as long as contained in the `ossec.conf`, but once the integratord logs the parameter to the `ossec.log` or `ìntegrator.log` file, anyone could see it.

[Ref](https://github.com/wazuh/internal-devel-requests/issues/74)

# Steps to reproduce using Virustotal integration
 - Manager: OVAL version 4.7.4
 - Agent: Ubuntu 22.04 with agent version 4.7.4
 - [VirusTotal integration](https://documentation.wazuh.com/current/user-manual/capabilities/malware-detection/virus-total-integration.html#virustotal-integration)
 - Debug Mode: Enable debug level 2 in the integratord module (`integrator.debug=2` on `/var/ossec/etc/internal_options.conf`)
 - Malicious File: Download a malicious file as specified in the [guide](https://documentation.wazuh.com/current/user-manual/capabilities/malware-detection/virus-total-integration.html#test-the-configuration)
 - Inspect `ossec.log` and `ìntegrator.log` file.

Damian-Mangold:
Update

- Sensitive information will no longer be printed in the `ossec.log` file by removing a debug log statement from the `integrator.c` file.
- Sensitive information will no longer be printed in the `integrations.log` file by initializing the `msg` variable as empty in each integration script.

PR #23833, opened 2024-05-31
integratord - Remove sensitive information from log files

Damian-Mangold:
|Related issue|
|---|
|#23648|

## Description

The objective of this PR is to prevent the publication of sensitive information in log files.

- Sensitive information will no longer be printed in the `ossec.log` file by removing a debug log statement from the `integrator.c` file.
- Sensitive information will no longer be printed in the `integrations.log` file by initializing the `msg` variable as empty in each integration script.

#### After
image

#### Before
image

#### After
image

#### Before
image

### DIFF
--- a/integrations/maltiverse.py
+++ b/integrations/maltiverse.py
@@ -234,12 +234,12 @@ def main(args: list):
         if len(args) >= 4:
             debug_enabled = len(args) > 4 and args[4] == 'debug'
         else:
-            msg = '# ERROR: Wrong arguments'
+            msg = '# ERROR: Wrong arguments\n'
             bad_arguments = True
 
         # Logging the call
         with open(LOG_FILE, 'a') as f:
-            f.write(msg + '\n')
+            f.write(msg)
 
         if bad_arguments:
             debug('# ERROR: Exiting, bad arguments. Inputted: %s' % args)
--- a/integrations/maltiverse.py
+++ b/integrations/maltiverse.py
@@ -230,10 +230,8 @@ def main(args: list):
     try:
         # Read arguments
         bad_arguments = False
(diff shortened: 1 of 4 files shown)


integrations/maltiverse.py
Dwordcito, line 1: We need to add a conditional to avoid writing an empty line if len(args) >= 4

integrations/pagerduty.py
Dwordcito, line 1: We need to add a conditional to avoid writing an empty line if len(args) >= 4

integrations/virustotal.py
Dwordcito, line 1: We need to add a conditional to avoid writing an empty line if len(args) >= 4

Dwordcito (changes requested):

integrations/maltiverse.py
Damian-Mangold: Instead of adding a conditional I included the line break to the only possible message that can be printed in case len(args) < 4.
This is how the log looks now:
image

integrations/pagerduty.py
Damian-Mangold: the same as [here](https://github.com/wazuh/wazuh/pull/23833#discussion_r1623204796)

integrations/virustotal.py
Damian-Mangold: the same as [here](https://github.com/wazuh/wazuh/pull/23833#discussion_r1623204796)

integrations/maltiverse.py
Damian-Mangold: Here I have simulated an error between two runs, you can see that the log is generated correctly, respecting the line breaks.
image

Dwordcito (approved):

Issue #18560, opened 2023-08-23
Add option to hide traceback from `WazuhInternalError`


Selutario:
## Description

There are multiple types of errors in the framework. Some of them are `WazuhError`, which is returned as a `400 Bad Request` in the API:
https://github.com/wazuh/wazuh/blob/2960662dcf23c88ea73eefdd2c330028426957aa/framework/wazuh/core/exception.py#L784-L790

and `WazuhInternalError`, which is returned as a `500 Internal Server Error`:
https://github.com/wazuh/wazuh/blob/2960662dcf23c88ea73eefdd2c330028426957aa/framework/wazuh/core/exception.py#L733-L739

When a `WazuhInternalError` is raised, the `dapi` module logs the error trace:
https://github.com/wazuh/wazuh/blob/2960662dcf23c88ea73eefdd2c330028426957aa/framework/wazuh/core/cluster/dapi/dapi.py#L188-L193

Printing the error trace is useful in the `api.log`, but it looks bad when printed as the output of a CLI:
```
# /var/ossec/bin/agent_upgrade -a 001 -f /home/ubuntu/wazuh_agent_v4.5.1_macos_x86_64.wpk -F
The WPK package used does not fit with the agent's architecture
concurrent.futures.process._RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/var/ossec/framework/python/lib/python3.9/concurrent/futures/process.py", line 246, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/var/ossec/framework/python/lib/python3.9/site-packages/wazuh-4.5.1-py3.9.egg/wazuh/core/cluster/dapi/dapi.py", line 247, in run_local
    data = f(**f_kwargs)
  File "/var/ossec/framework/python/lib/python3.9/site-packages/wazuh-4.5.1-py3.9.egg/wazuh/rbac/decorators.py", line 423, in wrapper
    result = func(*args, **kwargs) if not skip_execution else None
  File "/var/ossec/framework/python/lib/python3.9/site-packages/wazuh-4.5.1-py3.9.egg/wazuh/agent.py", line 989, in upgrade_agents
    raise WazuhInternalError(error_code, cmd_error=True, extra_message=agent_result['message'])
wazuh.core.exception.WazuhInternalError: Error 1836 - The WPK package used does not fit with the agent's architecture
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #23852, opened 2024-06-03
Added basic filter for CLI logging.

RamosFe:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/18560|

## Description

Adds Filter to loggers that are executed in the CLI.

### DIFF
--- a/framework/wazuh/core/tests/test_wlogging.py
+++ b/framework/wazuh/core/tests/test_wlogging.py
@@ -221,12 +221,13 @@ def test_customfilter():
     assert not cf.filter(MockedRecord('testB'))
 
 
-@pytest.mark.parametrize('value, expected',
-                         [
-                             ('Example log', True),
-                             ('Wazuh Internal Error', False),
-                             ('WazuhInternalError', False)
-                         ])
+@pytest.mark.parametrize('value, expected', [
+    ('Example log', True),
+    ('Wazuh Internal Error', False),
+    ('WazuhInternalError', False),
+    ('WazuhError', True),
+    ('InternalError', True)
+])
 def test_cli_custom_filter(value, expected):
     """
     Test if CLIFilter class works properly.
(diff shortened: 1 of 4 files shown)


framework/wazuh/core/tests/test_wlogging.py
```diff
@@ -219,3 +219,24 @@ def __init__(self, log_type):
...
     cf = wlogging.CustomFilter('testA')
     assert not cf.filter(MockedRecord('testB'))
+
+
+@pytest.mark.parametrize('value, expected',
+                         [
+                             ('Example log', True),
+                             ('Wazuh Internal Error', False),
+                             ('WazuhInternalError', False)
+                         ])
```
GGP1, line 229:
```suggestion
@pytest.mark.parametrize('value, expected', [
    ('Example log', True),
    ('Wazuh Internal Error', False),
    ('WazuhInternalError', False),
    ('WazuhError', True),
    ('InternalError', True)
])
```

framework/wazuh/core/wlogging.py
```diff
@@ -237,3 +237,29 @@ def __getattr__(self, item: str) -> object:
...
+        Parameters
+        ----------
+        record : logging.LogRecord
+            Contains all the information to the event being logged.
+
+        Returns
+        -------
+        bool
+            Boolean used to determine if the log entry should be logged.
+        """
```
GGP1, line 261:
```suggestion
    def filter(self, record: logging.LogRecord) -> bool:
        """Filter the log entry depending on its message contents.

        Parameters
        ----------
        record : logging.LogRecord
            Contains the information of the event being logged.

        Returns
        -------
        bool
            Whether the log entry should be logged or not.
        """
```

GGP1 (changes requested):

framework/wazuh/core/tests/test_wlogging.py
RamosFe: Fixed in 08f6d1a3a8a5db68c5dddd11bc1b88a047614105.

GGP1 (approved):
LGTM

Selutario (approved):
LGTM. According to the example shown [here](https://github.com/wazuh/wazuh/issues/18560#issuecomment-2145606882), less accurate information (`Agent 002 status: Send upgrade command error`) would be shown now if compared to the previous message (`Error 500: The WPK package used does not fit with the agent's architecture`) though.

PR #19993, opened 2023-11-02
Fix shuffle request warning

GGP1:
|Related issue|
|---|
|Closes https://github.com/wazuh/wazuh/issues/19604 |

## Description

Enables SSL certificate verification for the requests performed in the shuffle integration.

> I also added the `timeout` parameter to avoid indefinitely hang ups.

## Configuration options

```xml
<integration>
   <name>shuffle</name>
   <hook_url>https://shuffler.io/api/v1/hooks/WEBHOOK_ID</hook_url>
   <level>3</level>
   <alert_format>json</alert_format>
</integration>
```

Also, the value `integrator.debug` from the configuration in `internal_options.conf` was set to 2.

## Logs/Alerts example

<details><summary>ossec.log</summary>

```console
root@wazuh-master:/# cat /var/ossec/logs/ossec.log | grep wazuh-integratord
...
2023/11/02 15:02:42 wazuh-integratord[9334] integrator.c:143 at OS_IntegratorD(): INFO: Enabling integration for: 'shuffle'.
2023/11/02 15:02:42 wazuh-integratord[9334] integrator.c:154 at OS_IntegratorD(): DEBUG: jqueue_next()
2023/11/02 15:02:43 wazuh-integratord[9334] integrator.c:154 at OS_IntegratorD(): DEBUG: jqueue_next()
2023/11/02 15:02:44 wazuh-integratord[9334] integrator.c:154 at OS_IntegratorD(): DEBUG: jqueue_next()
2023/11/02 15:02:45 wazuh-integratord[9334] integrator.c:154 at OS_IntegratorD(): DEBUG: jqueue_next()
2023/11/02 15:02:46 wazuh-integratord[9334] integrator.c:154 at OS_IntegratorD(): DEBUG: jqueue_next()
2023/11/02 15:02:47 wazuh-integratord[9334] integrator.c:154 at OS_IntegratorD(): DEBUG: jqueue_next()
2023/11/02 15:02:47 wazuh-integratord[9334] integrator.c:161 at OS_IntegratorD(): DEBUG: Sending new alert.
2023/11/02 15:02:47 wazuh-integratord[9334] integrator.c:293 at OS_IntegratorD(): DEBUG: File /tmp/shuffle-1698937367-1525090471.alert was written.
[...]

Issue #19604, opened 2023-10-11
Fix shuffle insecure request warning


GGP1:
| Affected integration |
|---|
| Shuffle |

## Description

During the [4.6.0 RC1 release testing](https://github.com/wazuh/wazuh/issues/19602) of the Shuffle module, the following warning was found in the logs

```console
/var/ossec/framework/python/lib/python3.9/site-packages/urllib3/connectionpool.py:1013: 
InsecureRequestWarning: Unverified HTTPS request is being made to host 'shuffler.io'. 
Adding certificate verification is strongly advised. 
See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
```

We should make sure that the integration runs without any warnings.

## Tasks
- [x] Integration executes without warnings. Updated if there are any relevant changes.

GGP1:
## Update

Fixed a warning in the shuffle integration script and added the `timeout` parameter to avoid infinite hang ups.

PR #23924, opened 2024-06-05
Shuffle: SSL cert error fix in default configuration

0x0elliot:
## Description

Hey! I'm Aditya, an SRE at Shuffle. We often see users trying to integrate Wazuh with Shuffle, but they run into an SSL error due to the self-signed certificate that comes with the default Shuffle setup. This causes issues when users try to use the default HTTPS shuffle instance link in Wazuh, leading to errors. To fix this, users need to generate a new certificate with something like certbot, which can be frustrating and cause them to give up. It would be much easier if we could update the script to skip SSL certificate verification.

## Configuration options

N/A

## Logs/Alerts example

N/A

## Tests
  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests

### DIFF
--- a/integrations/shuffle.py
+++ b/integrations/shuffle.py
@@ -29,6 +29,11 @@ except ModuleNotFoundError:
     print("No module 'requests' found. Install: pip install requests")
     sys.exit(ERR_NO_REQUEST_MODULE)
 
+try:
+    requests.packages.urllib3.disable_warnings(requests.packages.urllib3.exceptions.InsecureRequestWarning)
+except Exception as e:
+    print(f"Failed to disable SSL warnings: {e}")
+
 # ossec.conf configuration structure
 # <integration>
 #  <name>shuffle</name>
--- a/integrations/shuffle.py
+++ b/integrations/shuffle.py
@@ -226,7 +226,7 @@ def send_msg(msg: str, url: str) -> None:
         URL of the integration.
     """
     headers = {'content-type': 'application/json', 'Accept-Charset': 'UTF-8'}
-    res = requests.post(url, data=msg, headers=headers, timeout=10)
+    res = requests.post(url, data=msg, headers=headers, timeout=10, verify=False)
     debug('# Response received: %s' % res.json)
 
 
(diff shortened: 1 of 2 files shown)


0x0elliot:
Hey! This is a very small fix. Wouldn't want to keep this waiting unnecessarily for too long. @MarcelKemp Anyone who would care to look? :)

MarcelKemp:
Hi @0x0elliot,

Could you update the unit tests? Because without them passing, I can't consider the PR.
It is only necessary to update the expected output of the test:
- https://github.com/wazuh/wazuh/actions/runs/9384040195/job/26193315019?pr=23924

Thanks in advance!

0x0elliot:
Done @MarcelKemp

0x0elliot:
PS: I can't see the current workflow runs. I apologise in advance if I am missing something.

MarcelKemp:
Hi again @0x0elliot, 

Sorry for the delay.
I'm communicating with the team responsible for coding the Shuffle integration, and apparently, they have deliberately removed this parameter in the following PR, to avoid the _`Warning`_ that was appearing:
- Issue: https://github.com/wazuh/wazuh/issues/19604
- PR: https://github.com/wazuh/wazuh/pull/19993

Therefore, we are deciding internally which is the best option.
As soon as we decide, I will inform you of the decision taken.

0x0elliot:
It has a simple fix here: https://stackoverflow.com/questions/27981545/suppress-insecurerequestwarning-unverified-https-request-is-being-made-in-pytho

0x0elliot:
There you go. Wasn't so hard. Let me know if this works! @MarcelKemp

MarcelKemp:
@0x0elliot Sorry, the problem as such has to do with the _`warning`_ but also with the security part.

In `4.7.1`, the change was added to avoid the _`warning`_ and to enforce security through certificates, so we have been discussing the best option, and we have come to the conclusion to keep the change made in `4.7.1` for the sake of security:
- https://github.com/wazuh/wazuh/pull/19993

So to fix this problem, it will be necessary to fix it in the system itself by making the system trust the certificates.

Sorry for the inconvenience!

PR #23951, opened 2024-06-06
Fixed wrong engine vs schema helper pattern and QoL changes

JavierBejMen:
|Related issue|
|---|
|#23952|

- This PR fixes the vs schema helper pattern to recognize latest format.
- Changed module to integration in asset metadata.
- QoL in engine-schema py script.

### DIFF
--- a/src/engine/ruleset/schemas/fields.json
+++ b/src/engine/ruleset/schemas/fields.json
@@ -5,8 +5,12 @@
   "type": "object",
   "additionalProperties": false,
   "patternProperties": {
-    "^~.+": {
-      "description": "Custom field.",
+    "^_.+": {
+      "description": "Custom variable.",
+      "ecs_type": null
+    },
+    ".+": {
+      "description": "Custom field not present in the schema.",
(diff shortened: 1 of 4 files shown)


juliancnn (changes requested):
I have noticed some details. Is it possible to fix them?

1.  `parse|field`: There are errors in the stage parse: 

image

2. End of string is not checked in the regex of the helpers functions

image

3. Only the 3er case should be valid:

image
4. Only 1 and 2 case should be valid:
image

5. Should Byte support floats if it is long?

image

JavierBejMen:
## Update
Fixed end of string in helper and ref pattern
Fixed parse item definition in schema
Ip not included as it will be overkill to create a regex to match all possible IP formats
Float distinction not included as in json only number is available

juliancnn (approved):
LGTM!

NahuFigueroa97 (approved):
LGTM!

JcabreraC (approved):
LGTM !

Issue #24182, opened 2024-06-20
Packages building Docker images auto-retag fails


mjcr99:
## Description

Hi team, 
The automatic workflow that retags the Docker images when a new release version branch is created has failed as it can be seen in this job:

* https://github.com/wazuh/wazuh/actions/runs/9585525991

The problem is related to the last step, where the `new_version` and `old_version` variables are set. As this workflow was triggered by a push event, the `{{ inputs.old_version }}` and `{{ inputs.new_version }}` are not set with their default value, making the `if` part run while the one that should be run is the `else` and not setting any value to the `new_version` and `old_version` variables.

PR #24187, opened 2024-06-20
Change variable value setting condition in Docker retag workflow

mjcr99:
|Related issue|
|---|
|#24182|

## Description

This PR causes the last step of the workflow file `.github/workflows/packages-retag-images.yml` to correctly set the values of the variables used in the script call, basing the variable values on the event that triggered the workflow instead of the default values of the input variables that may not be set if the workflow is triggered by an event that is not `workflow_dispatch`.

- https://github.com/wazuh/wazuh/actions/runs/9596107657 - :green_circle:

### DIFF
--- a/.github/workflows/packages-retag-images.yml
+++ b/.github/workflows/packages-retag-images.yml
@@ -63,7 +63,7 @@ jobs:
 
       - name: Run retag script
         run: |
-          if [ "${{ inputs.old_version }}" != "none" ] && [ "${{ inputs.new_version }}" != "none" ]; then
+          if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
             new_version=${{ inputs.new_version }}
             old_version=${{ inputs.old_version }}
           else


.github/workflows/packages-retag-images.yml
```diff
@@ -63,7 +63,7 @@ jobs:
 
       - name: Run retag script
         run: |
-          if [ "${{ inputs.old_version }}" != "none" ] && [ "${{ inputs.new_version }}" != "none" ]; then
+          if [ "${{ github.event_name }}" == "workflow_dispatch" }} ]; then
```
MarcelKemp, line 66:
```suggestion
          if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
```

MarcelKemp (changes requested):

.github/workflows/packages-retag-images.yml
mjcr99: Fixed https://github.com/wazuh/wazuh/pull/24187/commits/4e2a77721f4fa5100fe1bad49af3d2841871ddbb

MarcelKemp (approved):

Issue #24236, opened 2024-06-20
Revert permissions added in the generation of the Filebeat package


jotacarma90:
# Description

Hello team,
During the migration of the package tools, one of the steps was the `Filebeat` package generation, a modification was made both in the README and in the workflow that generates the package with respect to the old wazuh-packages code, adding the 755 permissions to an extra directory, `_meta`, which previously was not being done.

We want to revert this commit and leave it as it was, we are going to open a separate issue to investigate the issue and verify the impact this could have.
Issue to investigate:  https://github.com/wazuh/wazuh/issues/24237

## To do

- [x] Modify the README so that the need for this change is not expressed.
- [x] Delete the line where these new permissions are introduced in the workflow.
- [x] Modify the smoke test of the package permissions.

## Testing
- [x] Workflow successful: https://github.com/wazuh/wazuh/actions/runs/9599281841/job/26472642689

PR #24238, opened 2024-06-20
Revert permissions added in the generation of the Filebeat package

jotacarma90:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/24236|

## Description
In this PR we are reverting the new permissions added in both the Filebeat README and the Filebeat package generation workflow.

## Testing
- [x] Workflow successful: https://github.com/wazuh/wazuh/actions/runs/9599281841/job/26472642689


MarcelKemp (changes requested):
~If we remove the permissions setting from the `_meta` folder, I understand that the files inside the folder would not need to be modified either.~
It seems that the files previously did have their permissions set.

MarcelKemp (approved):

Issue #24111, opened 2024-06-14
wazuh-keystore doesn't support keys other than RSA


artur-borys:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.8.0 | wazuh-keystore | Manager | Packages | all |

When using ECDSA P-384 for ``/var/ossec/etc/sslmanager.key``, wazuh-keystore command doesn't work:
```
[root@706ce80d6e72 /]# /var/ossec/bin/wazuh-keystore -f indexer -k user -v admin
Unsupported key type
```
At least I suspect it doesn't support keys with algorithm other than RSA, based on https://github.com/wazuh/wazuh/blob/master/src/shared_modules/keystore/include/keyStore.hpp#L50

MiguelazoDS:
# Overview 

Indeed, there's no support for any other than RSA, I guess that the implementation was addressed following the current implementation of the `wazuh-authd` tool that only works with RSA. 

https://github.com/wazuh/wazuh/blob/1dc944432b9d3ab62050bc33ea953539a418b8a9/src/os_auth/generate_cert.c#L15-L50

Without considering, that we can generate a pair with a different encryption algorithm manually, which seems to be this case. 

https://documentation.wazuh.com/current/user-manual/agent/agent-enrollment/security-options/manager-identity-verification.html#manager-identity-validation

Dwordcito:
@MiguelazoDS thanks for the analysis. 

What we must do this time is use the highest-level API to avoid the decision of selecting the appropriate algorithm for asymmetric encryption. To do this we have to use the openssl envelope AKA EVP API

MiguelazoDS:
Studying openssl library, most of the currently used primitives are discouraged in favor of EVP API as Dworcito suggested.

MiguelazoDS:
I'm not seeing an easy way to support this type, as far as I know, this is more suitable for signature rather than encryption. 
Updating the primitives used in the keystore tool for EVP functions like 

```
EVP_PKEY_encrypt_init
EVP_PKEY_encrypt
```

`40A71EAC2A7A0000:error:03000096:digital envelope routines:evp_pkey_asym_cipher_init:operation not supported for this keytype:crypto/evp/asymcipher.c:189:`

My proposal is to generate temporary RSA pair in case of failure while using the keystore command line tool.

artur-borys:
With elliptic-curve, there's something called ECIES - elliptic-curve integrated encryption scheme:
[...]

PR #24357, opened 2024-06-28
Use dedicated RSA keys for keystore encryption

MiguelazoDS:
|Related issue|
|---|
|#24111|

## Description 

Using dedicated RSA keys for Keystore

## DoD (Update)

v4.8.0
```console
root@jammy:/home/vagrant/INDEXER# manage_agents -V

Wazuh v4.8.0 - Wazuh Inc.

This program is free software; you can redistribute it and/or modify
it under the terms of the GNU General Public License (version 2) as
published by the Free Software Foundation. For more details, go to
https://www.gnu.org/licenses/gpl.html
```

Non-RSA sslmanager keys
```console
root@jammy:/home/vagrant/INDEXER# openssl x509 -in /var/ossec/etc/sslmanager.cert -text | grep Public
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (384 bit)
```

wazuh-keystore tool fails
```console
root@jammy:/home/vagrant/INDEXER# /var/ossec/bin/wazuh-keystore -f default -k username -v test_user
Unsupported key type
```

Upgrade
```console
 Wazuh v4.8.1 (Rev. 40813) Installation Script - https://www.wazuh.com

 You are about to start the installation process of Wazuh.
 You must have a C compiler pre-installed in your system.

  - System: Linux jammy 5.15.0-113-generic (ubuntu 22.04)
  - User: root
  - Host: jammy


  -- Press ENTER to continue or Ctrl-C to abort. --


 - You already have Wazuh installed. Do you want to update it? (y/n):
```

```console
Unsupported key type
Generating RSA keys for Keystore.
```

Valid keys, indexer-connector successfully connected
```console
root@jammy:/home/vagrant/wazuh# grep "indexer-connector" /var/ossec/logs/ossec.log
2024/07/01 16:33:22 indexer-connector[378247] indexerConnector.cpp:88 at initConfiguration(): WARNING: No username found in the keystore, using default value.
2024/07/01 16:33:22 indexer-connector[378247] indexerConnector.cpp:319 at initialize(): INFO: IndexerConnector initialized successfully for index: wazuh-states-vulnerabilities-jammy.
root@jammy:/home/vagrant/wazuh# diff /var/ossec/etc/{sslmanager.key,keystore.key} -q
Files /var/ossec/etc/sslmanager.key and /var/ossec/etc/keystore.key differ
```

### DIFF
--- a/src/init/inst-functions.sh
+++ b/src/init/inst-functions.sh
@@ -1306,7 +1306,12 @@ InstallServer()
     ${INSTALL} -d -m 0750 -o ${WAZUH_USER} -g ${WAZUH_GROUP} ${INSTALLDIR}/queue/keystore
     ${INSTALL} -m 0750 -o root -g ${WAZUH_GROUP} wazuh-keystore ${INSTALLDIR}/bin/
 
-    GenerateKeystoreCert
+    if [ "X${update_only}" = "Xyes" ]; then
+      cp -pn ${INSTALLDIR}/etc/sslmanager.cert ${INSTALLDIR}/etc/keystore.cert
+      cp -pn ${INSTALLDIR}/etc/sslmanager.key ${INSTALLDIR}/etc/keystore.key
+    else
+      GenerateKeystoreCert
+    fi
 }
 
 InstallAgent()
--- a/src/init/inst-functions.sh
+++ b/src/init/inst-functions.sh
@@ -234,6 +234,21 @@ GenerateAuthCert()
     fi
 }
 
+#########################
+#GenerateKeystoreCert()
+#########################
+GenerateKeystoreCert()
+{
+  keystore_key=/etc/keystore.key
+  keystore_cert=/etc/keystore.cert
+  if [ ! -f "X${INSTALLDIR}${keystore_key}" ] && [ ! -f "${INSTALLDIR}${keystore_cert}" ]; then
+      echo "Generating RSA keys for Keystore."
+      ${INSTALLDIR}/bin/wazuh-authd -C 365 -B 2048 -K ${INSTALLDIR}${keystore_key} -X ${INSTALLDIR}${keystore_cert} -S "/C=US/ST=California/CN=wazuh/"
+      chmod 600 ${INSTALLDIR}${keystore_key}
+      chmod 600 ${INSTALLDIR}${keystore_cert}
+  fi
+}
+
 ##########
 # WriteLogs()
 ##########
@@ -1290,6 +1305,8 @@ InstallServer()
     # Keystore
     ${INSTALL} -d -m 0750 -o ${WAZUH_USER} -g ${WAZUH_GROUP} ${INSTALLDIR}/queue/keystore
     ${INSTALL} -m 0750 -o root -g ${WAZUH_GROUP} wazuh-keystore ${INSTALLDIR}/bin/
+
+    GenerateKeystoreCert
 }
 
 InstallAgent()
--- a/src/init/inst-functions.sh
+++ b/src/init/inst-functions.sh
@@ -239,17 +239,13 @@ GenerateAuthCert()
 #########################
 GenerateKeystoreCert()
 {
+  # Regenerate keys if they are not valid.
   keystore_key=/etc/keystore.key
   keystore_cert=/etc/keystore.cert
-  # Testing keys
-  ${INSTALLDIR}/bin/wazuh-keystore -f default -k username -v test_user
-  # Regenerate keys if they are not valid.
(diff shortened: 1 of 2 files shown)


artur-borys:
What will happen with existing keystores? Will there be a step in upgrade procedure to re-add the username and password to the keystore?

sebasfalcone:
> What will happen with existing keystores? Will there be a step in upgrade procedure to re-add the username and password to the keystone?

About to ask the same, @MiguelazoDS we have to test this

MiguelazoDS:
@artur-borys @sebasfalcone 

If we now do the same with sslmanager keys we will lose the information in the keystore anyway without even noticing, after this if we perform an upgrade we can copy the sslmanager keys used in the 4.8.0 version and rename the keys so the module can find them. I agree is not intuitive to copy those old keys after the failure, so I added the step during upgrade only if the keys in the new location do not exist to avoid replace them in further upgrades.

src/init/inst-functions.sh
```diff
@@ -1290,6 +1305,13 @@ InstallServer()
     # Keystore
     ${INSTALL} -d -m 0750 -o ${WAZUH_USER} -g ${WAZUH_GROUP} ${INSTALLDIR}/queue/keystore
     ${INSTALL} -m 0750 -o root -g ${WAZUH_GROUP} wazuh-keystore ${INSTALLDIR}/bin/
+
+    if [ "X${update_only}" = "Xyes" ]; then
+      cp -pn ${INSTALLDIR}/etc/sslmanager.cert ${INSTALLDIR}/etc/keystore.cert
+      cp -pn ${INSTALLDIR}/etc/sslmanager.key ${INSTALLDIR}/etc/keystore.key
+    else
+      GenerateKeystoreCert
+    fi
```
sebasfalcone, line 1314: Good solution! We should add a comment explaining this for future generations

sebasfalcone (approved):
LGMT!

---

The proposal makes the update compatible with already 4.8.0 installed versions of the manager. Modifying the certificates used by the Keystore will render the values unreadable but this "problem" was present before this changes

src/init/inst-functions.sh
```diff
@@ -1290,6 +1304,19 @@ InstallServer()
...
+    keystore_key=/etc/keystore.key
+    keystore_cert=/etc/keystore.cert
+    if [ ! -f "${INSTALLDIR}${keystore_key}" ] || [ ! -f "${INSTALLDIR}${keystore_cert}" ]; then
+      cp -p ${INSTALLDIR}/etc/sslmanager.cert ${INSTALLDIR}${keystore_cert}
+      cp -p ${INSTALLDIR}/etc/sslmanager.key ${INSTALLDIR}${keystore_key}
+      ${INSTALLDIR}/bin/wazuh-keystore -f default -k username -v test_user
+      if [ $? -eq 1 ]; then
+            GenerateKeystoreCert
+      fi
+    fi
```
sebasfalcone, line 1319: This logic has some flaws, what happens if the certificates are there but are not valid?

```suggestion
    keystore_key=/etc/keystore.key
    keystore_cert=/etc/keystore.cert

    # If we come from 4.8.0, no certificates will be found.
    # Since the Keystore tool previously used sslmanager keys for encryption,
    # We copy them to the new location to be able to recover the information.
    if [ ! -f "${INSTALLDIR}${keystore_key}" ] && [ ! -f "${INSTALLDIR}${keystore_cert}" ]; then
      cp -p ${INSTALLDIR}/etc/sslmanager.cert ${INSTALLDIR}${keystore_cert}
      cp -p ${INSTALLDIR}/etc/sslmanager.key ${INSTALLDIR}${keystore_key}
    fi
    
    # Test if the certificates are valid. If don't, re-generate them
    ${INSTALLDIR}/bin/wazuh-keystore -f default -k certificate_test -v test
    if [ $? -eq 1 ]; then
          GenerateKeystoreCert
    fi
```

src/init/inst-functions.sh
```diff
@@ -1290,6 +1304,19 @@ InstallServer()
     # Keystore
     ${INSTALL} -d -m 0750 -o ${WAZUH_USER} -g ${WAZUH_GROUP} ${INSTALLDIR}/queue/keystore
     ${INSTALL} -m 0750 -o root -g ${WAZUH_GROUP} wazuh-keystore ${INSTALLDIR}/bin/
+
+    # Since the Keystore tool previously used sslmanager keys for encryption
```
artur-borys, line 1308: And what if the existing keypair isn't RSA?

Dwordcito (approved):

PR #24398, opened 2024-07-02
Update Wazuh Keystore to allow setting password from file and stdin

### LINKED ISSUE #24110: Unsafe method of provisioning wazuh-keystore
|---|---|---|---|---|
| 4.8.0 | Keystore | Wazuh Manager | Packages | all |

Hello, I've noticed the addition of wazuh-keystore in release 4.8.0

I think it has a major issue, because the only way to provide a password to be stored in the keystore, is to pass it via an argument in clear text. This is a bad practice, as the command invocation can be persisted in shell history. Additionally, there might be a short period of time, during which the password can be seen in plain-text in the output of command like ``ps`` or any tool which is monitoring processes, so there's a risk of it being leaked to unauthorized users.

The history can be disabled for the time of command execution, but it doesn't help with hiding the plain-text password on the process list.

Ideally, there should be possibility to provide the password via stdin, similarily to filebeat keystore, i.e:
``` bash
echo 'SecretPassword' | /var/ossec/bin/wazuh-keystore -f indexer -k password
```

In most, if not all, shell implementations, the commands like ``echo`` and ``printf`` are built-ins, which means that they are not visible on the process list. The only thing visible on the process list would be ``/var/ossec/bin/wazuh-keystore -f indexer -k password``, so there's no risk of leaking the password in plain-text.

pereyra-m:
|Related issue|
|---|
|Closes #24110|

## Description

This PR adds more secure ways to provide the password to the `wazuh-keystore`.
A small test tool was added to make the testing easier.

The Github Actions workflow verifies the new feature with different passwords and insertion methods.

## Tests

Manual tests:

As expected, the `-v` parameter makes the password visible

```
./wazuh-keystore -f indexer -k password -v 'SecretPassword'
```

But using the `echo` doesn't

```
echo 'SecretPassword' | ./wazuh-keystore -f indexer -k password
```

  - [x] Linux
- [x] Source installation

### DIFF
--- a/src/shared_modules/keystore/qa/test_keystore.py
+++ b/src/shared_modules/keystore/qa/test_keystore.py
@@ -1,0 +1,168 @@
+# Copyright (C) 2015, Wazuh Inc.
+# Created by Wazuh, Inc. <info@wazuh.com>.
+# This program is free software; you can redistribute it and/or modify it under the terms of GPLv2
+
+from subprocess import Popen, PIPE
+import logging
+import os
+import pytest
+import trustme
+from pathlib import Path
+
+LOGGER = logging.getLogger(__name__)
+KEYSTORE_BINARY = "./wazuh-keystore"
+KEYSTORE_TESTTOOL_BINARY = "./wazuh-keystore-testtool"
+CERTS_PATH = 'etc/'
+PRIVATE_KEY_FILE = "etc/sslmanager.key"
+CERTIFICATE_FILE ="etc/sslmanager.cert"
+KEYSTORE_DB_PATH = "queue/keystore"
+
+# Helper methods
+
+def run_command(command):
+    """ Runs a command using subprocess.Popen and returns the stdout. In case of failure, it logs the error and fails the test.
+
+    Args:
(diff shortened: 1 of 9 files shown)


src/shared_modules/keystore/qa/test_keystore.py
```diff
@@ -0,0 +1,168 @@
...
+import os
+import pytest
+import trustme
+from pathlib import Path
+
+LOGGER = logging.getLogger(__name__)
+KEYSTORE_BINARY = "./wazuh-keystore"
+KEYSTORE_TESTTOOL_BINARY = "./wazuh-keystore-testtool"
+CERTS_PATH = 'etc/'
+PRIVATE_KEY_FILE = "etc/sslmanager.key"
```
artur-borys, line 16: The keypair used for keystore was changed in 4.8.1:
https://github.com/wazuh/wazuh/pull/24357
https://github.com/wazuh/wazuh/issues/24111

sebasfalcone (approved):
Good work Mati!

LGTM!

PR #24446, opened 2024-07-07
Discord Integration

elwali10:
## Description
Hello team, 

I am adding a native discord integration.

## Configuration options

```
 <integration>
 <name>discord</name>
 <hook_url>https://discord.com/api/webhooks/hookID</hook_url> <!-- Replace with your discord hook URL -->
 <level>3</level>
 <alert_format>json</alert_format>
 </integration>

```

## Logs/Alerts example

image


straysheep-dev:
Hi @elwali10, I posted this to the Wazuh Slack channel and they suggested I echo my question here in the comments. I've also been testing a Discord integration for just over a month in my lab environment: https://github.com/straysheep-dev/alert-service/tree/main/wazuh-integrations

I tried to use the slack.py integration as a base and just port the changes over, adding fields for windows events, Sysmon(+forlinux), Sysmon alerts with src and dst networking information, and wrapping any logs or possible URL strings in backticks so they aren't clickable. 

[This is the section that has the most changes + comments in mine for reference](https://github.com/straysheep-dev/alert-service/blob/6835de3c6fc58d384dd2a82a66c1a53b3675ecec/wazuh-integrations/custom-discord.py#L195).

Something else I realized you've already figured out -- some of my `full_log` auditd logs were getting dropped for some reason. Once I saw them go through with an empty string, then a small string, I realized the 1024 char limitation is on `fields.value`, and was causing the 400 response from Discord. I'd need to rewrite some of my conditional blocks to use the `description` field (4096 char limit) instead of using this temporary fix: 

```python
                'name': 'Full Log',
                'value': '```{0}```'.format(full_log[:1024]),
                'inline': False
```

...so they match the format you have here. 👍

If it's something you'd like me to help with and you think those additions will work here, what's the best way for me contribute?

PR #24524, opened 2024-07-10
Fixed engine's logpar choices

### LINKED ISSUE #24511: Engine - Bug in parser expressions using 'Choice' construct in HLP
|---|---|---|---|---|
| 5.0.0 | Engine| Manager | Packages/Sources | OS version |

### Description

In our parser expressions, we have the option to use a `choice` construct, which allows choosing between two parsers. For example:

```xml
<client.ip>?<client.address>
```

The parser will first evaluate `client.ip`, and if it fails, it will evaluate `client.address`. If both parsers fail, the expression will also fail. However, a problem has been identified with the `choice` construct after the implementation of parsers where each parser is divided into syntactic and semantic components.

### Issue Detail
If the syntactic parser for `client.ip` succeeds, but later, the entire expression fails when the semantic parser for `client.ip` fails, then even if the syntactic parser for `client.address` might have succeeded, the whole expression will still fail.

This issue often occurs when the syntactic parser of the first option is too flexible, especially when it has a stop token. To address this, two actions are required:

1. **Restriction on Stop Tokens**: Parsers on the left side of a `choice` must not have stop tokens.
2. **Mandatory Stop Tokens**: Some parsers have an optional stop token, which needs to be eliminated; these parsers should not require a stop token.

### Summary of Parsers and Stop Token Requirements

| Parser         | Need Stop Token          |
|----------------|--------------------------|
| Alphanumeric   | No                       |
| between        | No                       |
| bool           | No                       |
| date           | No                       |
| DSV            | Yes                      |
| CSV            | Yes                      |
| base64         | No                       |
| File           | Yes                      |
| Ignore         | No                       |
| IP             | Optionally (to be removed)|
| JSON           | No                       |
| KV             | Optionally (to be removed)|
| Literal        | No                       |
| Number         | No                       |
| Quoted         | No                       |
| Text           | Yes                      |
| FQDN           | Optionally (to be removed)|
| UA             | Yes                      |
| URI            | Yes                      |
| XML            | Yes                      |

### Proposed Solution
The proposed changes aim to make the parser behavior more predictable and robust, reducing the chance of incorrect parsing due to overly flexible syntactic rules. This adjustment ensures that parser expressions behave more consistently and according to expectations.

**Action Required:**
- Update the parser configurations according to the new requirements.
- Remove optional stop tokens from parsers where applicable.

JavierBejMen:
|Related issue|
|---|
|closes #24511 |

The following parsers no longer use the stop token:
- IP
- KV
- FQDN

When building logpar choices stop tokens are no longer passed to the first choice.


juliancnn (approved):
LGTM!

JcabreraC (approved):
LGTM !

PR #24362, opened 2024-06-29
Keystore refactor

zbalkan:
|Related issue|
|---|
| https://github.com/wazuh/wazuh/issues/24544 |

## Description

After upgrading to Wazuh 4.8.0, I had issues with keystore. Unfortunately all I got was `wazuh-modulesd:vulnerability-scanner: ERROR: VulnerabilityScannerFacade::start: RSA decryption failed`, and it is not helpful in troubleshooting.

In order to improve internal visibility, I only added some `logDebug2` options. Also did some refactoring around keystore application.

## Logs/Alerts example

Before:
```
wazuh-modulesd:vulnerability-scanner: ERROR: VulnerabilityScannerFacade::start: RSA decryption failed
```
After:

```
[DEBUG] keystore: Encryption successful for key: myKey
[DEBUG] keystore: Inserted encrypted value for key: myKey into column family: myFamily
```

```
[DEBUG] keystore: Decryption successful for key: myKey
```

```
[ERROR] keystore: Exception during encryption for key: myKey, Error: RSA encryption failed
```

```
[DEBUG] keystore: Encryption successful for key: myKey
[ERROR] keystore: Exception during database insertion for key: myKey, Error: Failed to put key-value pair into RocksDB
```

```
[DEBUG] keystore: Encryption successful for key: myKey
[DEBUG] keystore: Inserted encrypted value for key: myKey into column family: myFamily
[ERROR] keystore: Exception during decryption for key: myKey, Error: RSA decryption failed
```

```
[ERROR] keystore: Exception during database retrieval for key: myKey, Error: Failed to get value from RocksDB
```

## Tests

- Compilation without warnings in every supported platform
  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade
- [ ] Review logs syntax and correct language
- [ ] QA templates contemplate the added capabilities

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Valgrind (memcheck and descriptor leaks check)
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

- [ ] Retrocompatibility with older Wazuh versions
[...]

PR #24529, opened 2024-07-10
Remove hardcoded cipher text size

MiguelazoDS:
|Related issue|
|---|
|https://github.com/wazuh/external-devel-requests/issues/2552|

## Description

A hardcoded cipher text size was causing the decryption to fail for certain keys. 

## Logs/Alerts example

```console
root@jammy:/home/vagrant/wazuh# wazuh-keystore -f indexer -k username -v admin
root@jammy:/home/vagrant/wazuh# wazuh-keystore -f indexer -k password -v omRrPW5kfjrq+2d8J+LfGRDPV5?ags6.
root@jammy:/home/vagrant/wazuh# wazuh-control restart
root@jammy:/home/vagrant/wazuh# grep "RSA" /var/ossec/logs/ossec.log
root@jammy:/home/vagrant/wazuh#
```

image

### Update 

Log improvement 

> [!NOTE]
> We are already capturing all exceptions for method used by keystore, so there's no need to try catch everything in keystore.

image
image
image
For this last case the code was slightly modified to throw an exception even though the execution was successful

```console
2024/07/18 12:11:02 keystore[98293] keyStore.hpp:93 at get(): DEBUG: Decryption successful for key: 'username'
```
```console
2024/07/15 11:54:29 wazuh-modulesd:vulnerability-scanner[65921] vulnerabilityScannerFacade.cpp:484 at start(): ERROR: VulnerabilityScannerFacade::start: RSA decryption failed: oaep decoding error
```

### DIFF
--- a/src/shared_modules/keystore/include/keyStore.hpp
+++ b/src/shared_modules/keystore/include/keyStore.hpp
@@ -58,6 +58,10 @@ public:
         }
 
         keystoreDB.put(key, rocksdb::Slice(encryptedValue), columnFamily);
+        logDebug2(KS_NAME,
+                  "Encryption successful: value for key '%s', inserted into column family: '%s'",
+                  key.c_str(),
+                  columnFamily.c_str());
     }
 
     /**
@@ -88,6 +92,7 @@ public:
             }
             // Decrypt value
             TRSAPrimitive().rsaDecrypt(PRIVATE_KEY_FILE, encryptedValue, value);
+            logDebug2(KS_NAME, "Decryption successful for key: '%s'", key.c_str());
         }
     }
 };
--- a/src/shared_modules/keystore/include/keyStore.hpp
+++ b/src/shared_modules/keystore/include/keyStore.hpp
@@ -58,10 +58,6 @@ public:
         }
 
         keystoreDB.put(key, rocksdb::Slice(encryptedValue), columnFamily);
-        logDebug2(KS_NAME,
-                  "Encryption successful: value for key '%s', inserted into column family: '%s'",
-                  key.c_str(),
-                  columnFamily.c_str());
     }
 
     /**
--- a/src/shared_modules/keystore/include/keyStore.hpp
+++ b/src/shared_modules/keystore/include/keyStore.hpp
@@ -23,7 +23,7 @@ constexpr auto PRIVATE_KEY_FILE {"etc/sslmanager.key"};
 constexpr auto CERTIFICATE_FILE {"etc/sslmanager.cert"};
 
 constexpr auto KS_NAME {"keystore"};
-template<typename TRSAPrimitive = RSAHelper<>>
+template<typename TRSAPrimitive = RSAHelper>
 class TKeystore final
 {
 
--- a/src/shared_modules/utils/rsaHelper.hpp
+++ b/src/shared_modules/utils/rsaHelper.hpp
@@ -24,7 +24,6 @@
 constexpr int RSA_PRIVATE {0};
 constexpr int RSA_PUBLIC {1};
 constexpr int RSA_CERT {2};
-constexpr int RSA_SIZE {256};
 
 template<typename T = OpenSSLPrimitives, typename U = OSPrimitives>
 class RSAHelper final
@@ -93,7 +92,7 @@ public:
         DEFER([&]() { T::RSA_free(rsa); });
 
         // Decrypt the ciphertext using RSA private key
-        const auto decryptedLen = T::RSA_private_decrypt(RSA_SIZE,
+        const auto decryptedLen = T::RSA_private_decrypt(input.size(),
                                                          reinterpret_cast<const unsigned char*>(input.data()),
                                                          reinterpret_cast<unsigned char*>(decryptedText.data()),
                                                          rsa,
--- a/src/shared_modules/utils/rsaHelper.hpp
+++ b/src/shared_modules/utils/rsaHelper.hpp
@@ -64,7 +64,8 @@ public:
                                                         RSA_PKCS1_OAEP_PADDING);
         if (encryptedLen < 0)
         {
-            throw std::runtime_error("RSA encryption failed");
+            throw std::runtime_error("RSA encryption failed: " +
+                                     std::string(T::ERR_reason_error_string(T::ERR_get_error())));
         }
 
         output = std::string(encryptedValue.begin(), encryptedValue.end());
@@ -101,7 +102,8 @@ public:
 
         if (decryptedLen < 0)
         {
-            throw std::runtime_error("RSA decryption failed");
+            throw std::runtime_error("RSA decryption failed: " +
+                                     std::string(T::ERR_reason_error_string(T::ERR_get_error())));
         }
 
         // Display the decrypted plaintext
--- a/src/shared_modules/utils/rsaHelper.hpp
+++ b/src/shared_modules/utils/rsaHelper.hpp
@@ -55,12 +55,13 @@ public:
         // Defered free
         DEFER([&]() { T::RSA_free(rsa); });
 
+        // PKCS  #1  v1.5  padding.  This currently is the most widely used mode.  However, it is highly recommended to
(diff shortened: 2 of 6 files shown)


src/shared_modules/utils/rsaHelper.hpp
```diff
@@ -93,7 +92,7 @@ class RSAHelper final
         DEFER([&]() { T::RSA_free(rsa); });
 
         // Decrypt the ciphertext using RSA private key
-        const auto decryptedLen = T::RSA_private_decrypt(RSA_SIZE,
+        const auto decryptedLen = T::RSA_private_decrypt(input.size(),
```
sebasfalcone, line 95: From the [openssl documnetation](https://www.openssl.org/docs/man1.1.1/man3/RSA_private_decrypt.html):

> flen must not be more than RSA_size(rsa) - 11 for the PKCS #1 v1.5 based padding modes, not more than RSA_size(rsa) - 42 for RSA_PKCS1_OAEP_PADDING and exactly RSA_size(rsa) for RSA_NO_PADDING. When a padding mode other than RSA_NO_PADDING is in use, then RSA_public_encrypt() will include some random bytes into the ciphertext and therefore the ciphertext will be different each time, even if the plaintext and the public key are exactly identical. The returned ciphertext in to will always be zero padded to exactly RSA_size(rsa) bytes. to and from may overlap.

We have to be careful with the size we provide to this method based on the padding

sebasfalcone (changes requested):
Good finding! This was a serious bug

I have a question about the implementation

src/shared_modules/utils/rsaHelper.hpp
pereyra-m: I'd add tests for this change.

src/shared_modules/utils/rsaHelper.hpp
```diff
@@ -56,12 +55,13 @@ class RSAHelper final
         // Defered free
         DEFER([&]() { T::RSA_free(rsa); });
 
+        // PKCS  #1  v1.5  padding.  This currently is the most widely used mode.  However, it is highly recommended to
+        // use RSA_PKCS1_OAEP_PADDING in new applications.
         const auto encryptedLen = T::RSA_public_encrypt(input.length(),
```
sebasfalcone, line 60: This issue is still pending:
- https://github.com/wazuh/wazuh/pull/24529#discussion_r1672326487

We may want to have a table that represents this PADDING to max length relationships

MiguelazoDS: I don't understand the issue there.

sebasfalcone: I will expand my comment

---

```C++
 int RSA_public_encrypt(int flen, const unsigned char *from,
                        unsigned char *to, RSA *rsa, int padding);

 int RSA_private_decrypt(int flen, const unsigned char *from,
                         unsigned char *to, RSA *rsa, int padding);
```

The documentation states that `flen` must not be bigger than `RSA_size(rsa) - k` (being `k` a variable depending on the padding type)

---

This line may catch the problem, but its logging is poor
```C++
        if (decryptedLen < 0)
        {
            throw std::runtime_error("RSA decryption failed");
        }
```

We could improve if we use the error functions from the library:
- [ERR_get_error](https://www.openssl.org/docs/man1.1.1/man3/ERR_get_error.html)
- [ERR_error_string](https://www.openssl.org/docs/man1.1.1/man3/ERR_error_string.html)

MiguelazoDS: The first message [here](https://github.com/wazuh/wazuh/pull/24529#discussion_r1672326487) mentioned some checks to the size provided to the OpenSSL primitives that I didn't consider necessary since everything is handled internally by the library, that's why I didn't understand the request. Now that everything is clear I added a minor change to clarify the error message, please take a look.  

This would've been the error prior to this fix

`2024/07/15 11:54:29 wazuh-modulesd:vulnerability-scanner[65921] vulnerabilityScannerFacade.cpp:484 at start(): ERROR: VulnerabilityScannerFacade::start: RSA decryption failed: oaep decoding error`

src/shared_modules/utils/rsaHelper.hpp
```diff
@@ -92,16 +92,18 @@ class RSAHelper final
...
-        // Decrypt the ciphertext using RSA private key
-        const auto decryptedLen = T::RSA_private_decrypt(RSA_SIZE,
+        // PKCS  #1  v1.5  padding.  This currently is the most widely used mode.  However, it is highly recommended to
+        // use RSA_PKCS1_OAEP_PADDING in new applications.
+        const auto decryptedLen = T::RSA_private_decrypt(input.length(),
                                                          reinterpret_cast<const unsigned char*>(input.data()),
                                                          reinterpret_cast<unsigned char*>(decryptedText.data()),
                                                          rsa,
-                                                         RSA_PKCS1_PADDING);
+                                                         RSA_PKCS1_OAEP_PADDING);
```
sebasfalcone, line 101: This change breaks the compatibility with values already saved on the Keystore

The padding change is a recommendation from the library devs ([see documentation](https://www.openssl.org/docs/man1.1.1/man3/RSA_public_encrypt.html)):
>[!IMPORTANT]
> Decryption failures in the RSA_PKCS1_PADDING mode leak information which can potentially be used to mount a Bleichenbacher padding oracle attack. This is an inherent weakness in the PKCS #1 v1.5 padding design. Prefer RSA_PKCS1_OAEP_PADDING.

@Dwordcito I've performed some tests. If we maintain the old padding, the changes in this PR will be compatible with previous versions. Changing the padding breaks that compatibility but improves the security. How should we proceed?

sebasfalcone (approved):
Good Work @MiguelazoDS!

Changes LGTM and tests land the expected result, I have one doubt that I will discuss it with @Dwordcito

src/shared_modules/utils/rsaHelper.hpp
Dwordcito: Maintain the old padding so as not to break upgrades.

sebasfalcone: @MiguelazoDS Lets revert the padding changes, we are going to address this padding problem in another release (to not break retro-compatibility)

Great finding btw!

sebasfalcone (changes requested):

sebasfalcone (approved):

sebasfalcone (changes requested):
@MiguelazoDS Lets add the debug logs proposed here:
- https://github.com/wazuh/wazuh/pull/24362

sebasfalcone (approved):
LGTM!

src/shared_modules/keystore/include/keyStore.hpp
```diff
@@ -58,6 +58,10 @@ class TKeystore final
         }
 
         keystoreDB.put(key, rocksdb::Slice(encryptedValue), columnFamily);
+        logDebug2(KS_NAME,
```
Dwordcito, line 61: Let's remove this log.
1- Out of scope.
2- Terms like Encryption, column family are implementation details.
3- Printed in stdout of the tool.

src/shared_modules/keystore/include/keyStore.hpp
```diff
@@ -88,6 +92,7 @@ class TKeystore final
             }
             // Decrypt value
             TRSAPrimitive().rsaDecrypt(PRIVATE_KEY_FILE, encryptedValue, value);
+            logDebug2(KS_NAME, "Decryption successful for key: '%s'", key.c_str());
```
Dwordcito, line 95: We can keep this log because it is debug and by default it does not appear in the ossec.log

Dwordcito (changes requested):

src/shared_modules/keystore/include/keyStore.hpp
MiguelazoDS: I wasn't sure to include that log either. Removed.

Dwordcito (approved):

Issue #21899, opened 2024-02-13
The macOS agent displays an error message on startup stating `No such file or directory`


rauldpm:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| At least 4.7.2 | wazuh-control | Agent | Packages | macOS 14.2.1 M1 |

While testing a community report https://github.com/wazuh/community/issues/19542, the following error message was encountered when starting the macOS agent

```
sh-3.2# /Library/Ossec/bin/wazuh-control start
cat: /Library/Ossec/var/start-script-lock/pid: No such file or directory
Starting Wazuh v4.7.2...
```

Searching, only two references to this message have been found:

- E2E 4.8.0 alpha 2: https://github.com/wazuh/wazuh/issues/21358#issuecomment-1892699856
- Pull request: https://github.com/wazuh/wazuh/pull/21637#issuecomment-1912558925

We need to determine the origin of the message and if it can cause any problems in the operation of the agent

<details><summary>Output</summary>

```
sh-3.2# installer -pkg wazuh-agent-4.7.2-1.arm64.pkg -target /
installer: Package name is Wazuh Agent
installer: Installing at base path /
installer: The install was successful.
sh-3.2# ls -l /Library/Ossec/etc/ossec.conf 
-rw-r-----  1 root  wazuh  4933 Feb 13 14:58 /Library/Ossec/etc/ossec.conf
sh-3.2# grep "server" /Library/Ossec/etc/ossec.conf 
    <server>
    </server>
sh-3.2# nano /Library/Ossec/etc/ossec.conf 
sh-3.2# /Library/Ossec/bin/wazuh-control start
Starting Wazuh v4.7.2...
Started wazuh-execd...
Started wazuh-agentd...
Started wazuh-syscheckd...
Started wazuh-logcollector...
Started wazuh-modulesd...
Completed.
sh-3.2# tail -f /Library/Ossec/logs/ossec.log 
2024/02/13 15:07:13 sca: INFO: Starting Security Configuration Assessment scan.
2024/02/13 15:07:13 wazuh-modulesd:syscollector: INFO: Module started.
2024/02/13 15:07:13 wazuh-modulesd:syscollector: INFO: Starting evaluation.
2024/02/13 15:07:13 sca: INFO: Starting evaluation of policy: '/Library/Ossec/ruleset/sca/cis_apple_macOS_14.0.yml'
[...]

PR #24531, opened 2024-07-10
Redirect cat errors correctly in wazuh-control

mjcr99:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/21899|

## Description

Hi team,

this PR fixes the cat error output redirection to avoid showing the message when the file is not created. 

The line `pid='cat ${LOCK_PID}' 2>/dev/null` does not correctly suppress the errors because the redirect `2>/dev/null`only applies to the external part of the command (the pid assignment), not to the cat command.

### DIFF
--- a/src/init/wazuh-client.sh
+++ b/src/init/wazuh-client.sh
@@ -63,7 +63,7 @@ lock()
         # Waiting 1 second before trying again
         sleep 1;
         i=`expr $i + 1`;
-        pid=`cat ${LOCK_PID}` 2>/dev/null
+        pid=$(cat ${LOCK_PID} 2>/dev/null)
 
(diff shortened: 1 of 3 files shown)


lchico (approved):
Tested and works as expected! GJ @mjcr99

MarcelKemp (approved):

MarcelKemp:
## Failing checks

### Build Linux x86_64/arm64v8 manager rpm package
- Both are expected, because _wazuh-packages_ from `4.9.0` onwards are not being updated, causing these checks to fail due to lack of SPECS updates.
  - https://github.com/wazuh/wazuh-packages/issues/3030

Issue #24369, opened 2024-07-01
SPIKE - Define the API schema for the validation request for uninstallation process


jotacarma90:
Parent issue:
- https://github.com/wazuh/wazuh/issues/23466

# Description

As described in the main issue, we need to develop a mechanism to prevent the uninstallation of the Wazuh agent from Linux endpoints. This will enhance security by ensuring that agents are not removed without proper authorization. To achieve this, we will implement a validation process involving the Wazuh manager API.

We will need the agent to have access to the manager's CA certificate in order to use the API. In addition, it is necessary to add a new API endpoint:
```
GET /agents/{agent_id}/uninstall_permission
```

Which will require a new RBAC action:
```
agent:uninstall
```

The successful response should look similar to [POST /groups](https://documentation.wazuh.com/current/user-manual/api/reference.html#operation/api.controllers.agent_controller.post_group) for consistency reasons.
- **HTTP Status Code:** 200 OK
- **JSON response:**
  ```JSON
  {
    "message": "Agent has permission to be uninstalled",
    "error": 0
  }
  ``` 

The error response should be like this:
- **HTTP Status Code:** 403 Forbidden
- **JSON response:**
  ```JSON
  {
    "title": "Permission Denied",
    "detail": "Permission denied: Resource type: *:*",
    "remediation": "Please, make sure you have permissions to execute the current request. For more information on how to set up permissions, please visit https://documentation.wazuh.com/4.8/user-manual/api/rbac/configuration.html",
    "error": 4000,
    "dapi_errors": {
        "unknown-node": {
            "error": "Permission denied: Resource type: *:*"
        }
    }
  }
  ``` 

## Plan
This is a spike, we only need to create the endpoint prototype to confirm whether this can be achieved, but tests are not needed yet.
- [x] Create new RBAC action.
- [x] Create new API endpoint.
- [x] Verify the response is as expected both when using (or not using) the expected permissions.

RamosFe:
## Update

I continue with the development of the endpoint. The endpoint is already developed with hardcoded values, its logic needs to be finished

RamosFe:
## Update 

Couldn't continue with the development due to other Issues having closer ETA and Community support.

RamosFe:
## Design - RBAC - Resources and Actions

[...]

PR #24621, opened 2024-07-15
Added validation request for uninstallation process

RamosFe:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/24369|

## Description

Adds endpoint `/agents/uninstall` for uninstallation process validation.

### DIFF
--- a/api/api/spec/spec.yaml
+++ b/api/api/spec/spec.yaml
@@ -167,6 +167,14 @@ x-rbac-catalog:
         actions: ['agent:restart']
         resources: ['agent:id:050', 'agent:id:049']
         effect: "deny"
+    'agent:uninstall':
+      description: "Uninstall agents"
+      resources:
+        - $ref: '#/x-rbac-catalog/resources/*:*'
+      example:
+        actions: ['agent:uninstall']
+        resources: [ '*:*:*' ]
+        effect: "allow"
     'group:create':
       description: "Create new agent groups"
       resources:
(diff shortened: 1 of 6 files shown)


nico-stefani (changes requested):
LGTM. Just remove the changelog entry.

nico-stefani (approved):
LGTM

api/api/spec/spec.yaml
```diff
@@ -167,6 +167,14 @@ x-rbac-catalog:
...
         resources: ['agent:id:050', 'agent:id:049']
         effect: "deny"
+    'agent:uninstall':
+      description: "Check user's permission to uninstall agents"
+      resources:
+        - $ref: '#/x-rbac-catalog/resources/*:*'
+      example:
+        actions: ['agent:uninstall']
+        resources: [ '*:*:*' ]
+        effect: "allow"
```
fdalmaup, line 177: Why is the action over every resource?

framework/wazuh/rbac/default/policies.yaml
```diff
@@ -8,6 +8,7 @@ default_policies:
         actions:
           - agent:create
           - group:create
+          - agent:uninstall
```
fdalmaup, line 11: Similar doubt related to the one above.

fdalmaup (changes requested):

api/api/spec/spec.yaml
RamosFe: @fdalmaup As described in https://github.com/wazuh/wazuh/issues/24369#issuecomment-2228769413, if the user has the action `agents:uninstall`, it can uninstall any agents. Due to this decision, the action is `resourceless`, so there is no associated resource to the action. All actions that are `resourceless` are associated with the resource `*:*:*`, for example, the `agent:create` and `agent:group`:

https://github.com/wazuh/wazuh/blob/ee9f897c02fee8c8469abefc1e792785deace598/framework/wazuh/rbac/default/policies.yaml#L4-L14

In this case, `agent:create` can create new agents or can't, there is no in-between, equal to `agent:uninstall` and doesn't make any changes/modifications to the already created resources in the system, so it doesn't need to have a specific resource.

If we add. the resource `agent:id:*` to `agent:uninstall` we end with the second approach described in https://github.com/wazuh/wazuh/issues/24369#issuecomment-2223365815.

framework/wazuh/rbac/default/policies.yaml
RamosFe: Same explanation as https://github.com/wazuh/wazuh/pull/24621#discussion_r1684427336.

fdalmaup (approved):
LGTM! GJ

PR #22973, opened 2024-04-18
Fix: For developer tag, if the branch name has "/" change change by "…

lchico:
|Related issue|
|---|
https://github.com/wazuh/wazuh/issues/22969
||

## Description

Resolved issue with image tag naming during Docker uploads and package generation workflows by replacing '/' with '--' as a solution to ensure compliance with valid ASCII characters.

## Tests
change_tag
test_tag_2

PR #24759, opened 2024-07-22
Support PPC64 architecture in workflows to upload images used for generate agent packages

jotacarma90:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent-packages/issues/55|

## Description

In this PR we are going to modify the names of some of the workflows that have been modified in the wazuh-agent-packages repo to support PPC64. Some of those names are triggered from this other repo, so after having modified their names, we must make those same modifications in this repo.

In this PR it has been necessary to adapt multiple files involved in the generation of ppc64le packages, both to improve the performance and to introduce the new image names:
- `retag_image.sh`
  - Added new image names
  - `set -ex` to avoid false successes.
- `install_component.sh`
  - To check that the necessary parameters arrive.
- `packages-retag-images.yml`
  - Edit tag names to prevent them from containing the invalid char `/`.
- `packages-upload-images.yml`
  - Introducing the new files and image names to the automation.
- `wazuh-agent.spec`
  - Adapting the SPECs to prevent an error during ppc64le specific package generation.
- `Dockerfile (ppc64le)`
  - Error detected with some non-existing mirrorlist in this version `FROM ppc64le/centos:7`.

### Retag workflows testing for new images 🟢 
- `pkg_deb_agent_builder_ppc64le`: https://github.com/wazuh/wazuh/actions/runs/10738767137
- `pkg_rpm_agent_builder_ppc64le`: https://github.com/wazuh/wazuh/actions/runs/10738762423


.github/workflows/packages-upload-images.yml
```diff
@@ -130,6 +140,13 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.CI_WAZUH_AGENT_PACKAGES }}
 
+      - name: Request pkg_deb_agent_builder_ppc64le update
+        if: steps.changes.outputs.pkg_deb_agent_builder_ppc64le == 'true'
+        run: |
+          gh workflow run packages-upload-agent-images-ppc64le.yml --repo wazuh/wazuh-agent-packages -r ${{ github.ref_name }} -f docker_image_tag=${{ env.TAG }} -f system=deb -f architecture=ppc64le -f source_reference=${{ github.ref_name }}
```
lchico, line 146: Review the **yml** name: [packages-upload-agent-images-ppc.yml](https://github.com/wazuh/wazuh-agent-packages/pull/64/files#diff-150f4e669f4c57ec69baa1092150c68a2cee5c213a4fd3eecf1d2f6927fdf623R1)

```suggestion
          gh workflow run packages-upload-agent-images-ppc.yml --repo wazuh/wazuh-agent-packages -r ${{ github.ref_name }} -f docker_image_tag=${{ env.TAG }} -f system=deb -f architecture=ppc64le -f source_reference=${{ github.ref_name }}
```

.github/workflows/packages-upload-images.yml
```diff
@@ -144,6 +161,13 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.CI_WAZUH_AGENT_PACKAGES }}
 
+      - name: Request pkg_rpm_agent_builder_ppc64le update
+        if: steps.changes.outputs.pkg_rpm_agent_builder_ppc64le == 'true'
+        run: |
+          gh workflow run packages-upload-agent-images-ppc64le.yml --repo wazuh/wazuh-agent-packages -r ${{ github.ref_name }} -f docker_image_tag=${{ env.TAG }} -f system=rpm -f architecture=ppc64le -f source_reference=${{ github.ref_name }}
```
lchico, line 167: Review the same.

lchico (changes requested):
GJ! Some minor changes.

.github/workflows/packages-upload-images.yml
jotacarma90: done

.github/workflows/packages-upload-images.yml
jotacarma90: done

lchico (dismissed):
LGTM!

MarcelKemp (changes requested):
Checking the following execution of `retag_image`, I have observed that it is a false positive, as it is actually failing. So, let's correct the following two points:
- If an error appears, let it appear in :red_circle: (error checking).
- In the case of using `/` in the tag, let's do the corresponding replacement (by `--`), as we have done in other workflows: https://github.com/wazuh/wazuh/pull/22973

jotacarma90:
> Checking the following execution of `retag_image`, I have observed that it is a false positive, as it is actually failing. So, let's correct the following two points:
> 
> * If an error appears, let it appear in 🔴 (error checking).
> * In the case of using `/` in the tag, let's do the corresponding replacement (by `--`), as we have done in other workflows: [Fix: For developer tag, if the branch name has "/" change change by "… #22973](https://github.com/wazuh/wazuh/pull/22973)

Fix introduced: https://github.com/wazuh/wazuh/pull/24759/commits/ff61e12a562652245d76e68c845835311f73d6e7
To solve it I have preferred to modify inside the workflow the variables that contain the char `/`, editing it by `--` as we had already done in other places:

I also added set -ex to the script to make the workflow fail correctly in cases like this.

MarcelKemp (approved):

PR #24924, opened 2024-07-24
Enhancement multithreading dispatching

GabrielEValenzuela:
|Related issue|
|---|
|#24352|

# Objective

This PR aims to enhance the `ThreadEventDispatcher` logic within the Indexer Connector, enabling robust multithreading capabilities for dispatching alerts. By adopting an approach similar to that used in the `threadDispatcher` class, this update ensures efficient and scalable handling of alerts in multi-threaded environments, while maintaining backward compatibility with existing systems.

## Description
- Recognizing the need to improve alert processing performance as system demand scales, we have revised the `ThreadEventDispatcher` to support multithreading. This modification allows the Indexer Connector to handle a larger volume of alerts concurrently, significantly improving throughput and response time.

### Static Analysis
N/A

#### Unit Testing
- Tests were specifically designed to check for backward compatibility, ensuring that existing functionalities are not affected by the introduction of multithreading capabilities.

> [!NOTE]
> In a sandbox environment, run the test 100 times and no errors were found.

-----------------------

### DIFF
--- a/src/shared_modules/utils/tests/threadEventDispatcher_test.cpp
+++ b/src/shared_modules/utils/tests/threadEventDispatcher_test.cpp
@@ -21,7 +21,8 @@ void ThreadEventDispatcherTest::TearDown() {
 };
 
 constexpr auto BULK_SIZE {50};
-TEST_F(ThreadEventDispatcherTest, ConstructorTestSingleThread) {
+TEST_F(ThreadEventDispatcherTest, ConstructorTestSingleThread)
+{
     static const std::vector<int> MESSAGES_TO_SEND_LIST {120, 100};
 
     for (auto MESSAGES_TO_SEND : MESSAGES_TO_SEND_LIST)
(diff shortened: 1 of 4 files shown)


src/shared_modules/utils/tests/threadEventDispatcher_test.cpp
sebasfalcone, line 1: We need to add a test for the multi-thread alternative. Behaviour should be the same as in a singleThread scenario

sebasfalcone (changes requested):

src/shared_modules/utils/tests/threadEventDispatcher_test.cpp
GabrielEValenzuela: We added!

sebasfalcone (changes requested):

src/shared_modules/utils/threadEventDispatcher.hpp
```diff
@@ -25,6 +25,7 @@
 template<typename T,
          typename U,
          typename Functor,
+         int TNumberOfThreads = 1,
```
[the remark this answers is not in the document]
GabrielEValenzuela: Done! Thanks!

src/shared_modules/utils/threadEventDispatcher.hpp
sebasfalcone, line 1: We can add a simplification to the static asserts (which is an awesome idea BTW!)

```C++
private:
    /**
     * @brief Check if the queue type is a `TSafeMultiQueue`.
     *
     */
    static constexpr bool isTSafeMultiQueue =
        std::is_same_v<Utils::TSafeMultiQueue<T, U, RocksDBQueueCF<T, U>>, TSafeQueueType>;

    /**
     * @brief Check if the queue type is a `TSafeQueue`.
     *
     */
    static constexpr bool isTSafeQueue = std::is_same_v<Utils::TSafeQueue<T, U, RocksDBQueue<T, U>>, TSafeQueueType>;
```

Then, we can replace this
```C++
                static_assert(std::is_same_v<Utils::TSafeQueue<T, U, RocksDBQueue<T, U>>, TSafeQueueType> ||
                                  std::is_same_v<Utils::TSafeMultiQueue<T, U, RocksDBQueueCF<T, U>>, TSafeQueueType>,
                              "This method is not supported for this queue type");
```

With this:
```C++
                static_assert(isTSafeQueue || isTSafeMultiQueue, "This method is not supported for this queue type");
```

sebasfalcone, line 1: We can add a simplification to the static asserts (which is an awesome idea BTW!)

```C++
private:
    /**
     * @brief Check if the queue type is a `TSafeMultiQueue`.
     *
     */
    static constexpr bool isTSafeMultiQueue =
        std::is_same_v<Utils::TSafeMultiQueue<T, U, RocksDBQueueCF<T, U>>, TSafeQueueType>;

    /**
     * @brief Check if the queue type is a `TSafeQueue`.
     *
     */
    static constexpr bool isTSafeQueue = std::is_same_v<Utils::TSafeQueue<T, U, RocksDBQueue<T, U>>, TSafeQueueType>;
```

Then, we can replace this
```C++
                static_assert(std::is_same_v<Utils::TSafeQueue<T, U, RocksDBQueue<T, U>>, TSafeQueueType> ||
                                  std::is_same_v<Utils::TSafeMultiQueue<T, U, RocksDBQueueCF<T, U>>, TSafeQueueType>,
                              "This method is not supported for this queue type");
```

With this:
```C++
                static_assert(isTSafeQueue || isTSafeMultiQueue, "This method is not supported for this queue type");
```

src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/scanOrchestrator.hpp
```diff
@@ -27,15 +27,18 @@ constexpr auto INVENTORY_DB_PATH = "queue/vd/inventory";
 constexpr auto DELAYED_EVENTS_BULK_SIZE {1};
 constexpr auto DELAYED_QUEUE_PATH = "queue/vd/delayed";
 constexpr auto DELAYED_POSTPONE_SECONDS {60};
+constexpr auto MAX_THREADS {1};
```
sebasfalcone, line 30: Compilation is breaking if I provide more than one thread

Its breaking in [this push](https://github.com/wazuh/wazuh/blob/98ee756f4b160b298734936aac708bfa1a755df3/src/shared_modules/utils/threadEventDispatcher.hpp#L315-L316), because it expects two parameters

sebasfalcone (changes requested):

Issue #25062, opened 2024-08-05
Improve logging for indexer connector monitoring class


pereyra-m:
## Description

Working on https://github.com/wazuh/external-devel-requests/issues/2691, it was found that it would be really useful to improve the logging in the indexer connector module, more precisely in the `Monitoring` class.

https://github.com/wazuh/wazuh/blob/038fe11d988fa54cca8a6c3101fc7906fad5da3b/src/shared_modules/indexer_connector/src/monitoring.hpp#L157

When the request fails to obtain the cluster status, no logs are printed making the troubleshooting task harder.

## Update

- [x] Improve the logging of the suggested class and any related
- [x] Add UT
- [x] CI Pass

PR #25189, opened 2024-08-12
Improve log messages monitoring class

MiguelazoDS:
|Related issue|
|---|
|#25062|

## Description

Some improvements over the monitoring class were added. 

## Logs/Alerts example

Take a look at this comment https://github.com/wazuh/wazuh/issues/25062#issuecomment-2284779664

## Tests

  - [x] Linux
- [x] Source installation

### DIFF
--- a/src/shared_modules/indexer_connector/src/monitoring.hpp
+++ b/src/shared_modules/indexer_connector/src/monitoring.hpp
@@ -153,7 +153,7 @@ public:
                                         }
                                         logDebug2(IC_NAME,
                                                   "Cluster health status '%s' for server '%s'",
-                                                  Utils::toUpperCase(fields.at(HealthCheckColumns::STATUS)).c_str(),
+                                                  fields.at(HealthCheckColumns::STATUS).c_str(),
                                                   value.first.c_str());
                                     }
                                     else
@@ -166,9 +166,8 @@ public:
                                 },
                                 [&](const std::string& error, const long statusCode)
                                 {
-                                    std::string errorMessage = error;
-
-                                    errorMessage += " (Status code: " + std::to_string(statusCode) + ")";
+                                    const std::string errorMessage =
+                                        error + ". Status code: " + std::to_string(statusCode);
                                     logDebug2(IC_NAME, errorMessage.c_str());
 
                                     value.second = false;
--- a/src/shared_modules/indexer_connector/src/monitoring.hpp
+++ b/src/shared_modules/indexer_connector/src/monitoring.hpp
@@ -28,8 +28,7 @@
 #include <vector>
 
 constexpr auto INTERVAL = 60u;
-constexpr auto IC_MONITORING {"indexer-connector"};
-
+constexpr auto IC_NAME {"indexer-connector"};
 namespace HealthCheckRows
 {
     enum Type : std::size_t
@@ -150,16 +149,17 @@ public:
                                         }
                                         else
                                         {
-                                            logWarn(IC_MONITORING,
-                                                    "Cluster health status: '%s'",
-                                                    Utils::toUpperCase(fields.at(HealthCheckColumns::STATUS)).c_str());
                                             value.second = false;
                                         }
+                                        logDebug2(IC_NAME,
+                                                  "Cluster health status '%s' for server '%s'",
+                                                  Utils::toUpperCase(fields.at(HealthCheckColumns::STATUS)).c_str(),
+                                                  value.first.c_str());
                                     }
                                     else
                                     {
-                                        logDebug2(IC_MONITORING,
-                                                  "Bad response from _cluster/health: %s",
+                                        logDebug2(IC_NAME,
+                                                  "Invalid response from /_cat/health?v: %s",
                                                   originalResponse.c_str());
                                         value.second = false; // LCOV_EXCL_LINE
                                     }
@@ -167,13 +167,10 @@ public:
                                 [&](const std::string& error, const long statusCode)
                                 {
                                     std::string errorMessage = error;
-                                    constexpr auto NOT_USED {-1};
 
-                                    if (statusCode != NOT_USED)
-                                    {
-                                        errorMessage += " (Status code: " + std::to_string(statusCode) + ")";
-                                        logDebug2(IC_MONITORING, errorMessage.c_str());
-                                    }
+                                    errorMessage += " (Status code: " + std::to_string(statusCode) + ")";
+                                    logDebug2(IC_NAME, errorMessage.c_str());
+
                                     value.second = false;
                                 },
                                 "",
--- a/src/shared_modules/indexer_connector/src/monitoring.hpp
+++ b/src/shared_modules/indexer_connector/src/monitoring.hpp
@@ -19,6 +19,7 @@
 #include <atomic>
 #include <chrono>
 #include <condition_variable>
+#include <loggerHelper.h>
 #include <map>
 #include <mutex>
 #include <string>
@@ -27,6 +28,7 @@
 #include <vector>
 
 constexpr auto INTERVAL = 60u;
+constexpr auto IC_MONITORING {"indexer-connector"};
 
 namespace HealthCheckRows
 {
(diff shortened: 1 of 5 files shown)


src/shared_modules/indexer_connector/src/monitoring.hpp
```diff
@@ -146,15 +150,32 @@ class Monitoring final
                                         }
                                         else
                                         {
+                                            logWarn(IC_MONITORING,
+                                                    "Cluster health status: '%s'",
```
pereyra-m, line 154: Improve: I understand that the pre-existent code queries all the nodes of the cluster with the same query, and that all of them will return the same result because all belong to the same cluster.

But considering this, in the case a cluster isn't 'green', we'd see the same message many times. I'd add the name of the server to the message to mitigate this situation.

src/shared_modules/indexer_connector/src/monitoring.hpp
```diff
@@ -27,6 +28,7 @@
 #include <vector>
 
 constexpr auto INTERVAL = 60u;
+constexpr auto IC_MONITORING {"indexer-connector"};
```
pereyra-m, line 31: Nitpick: maybe if we move the `IC_NAME` constant, we can avoid declaring two expressions with the same value.

src/shared_modules/indexer_connector/src/monitoring.hpp
```diff
@@ -146,15 +150,32 @@ class Monitoring final
                                         }
                                         else
                                         {
+                                            logWarn(IC_MONITORING,
```
pereyra-m, line 153: Question: is it Ok to print this failure as a WARNING message but the other ones as DEBUG?
We should unify the criteria.

If we consider that the requests are run every 60 seconds, I vote to use DEBUG for all of them-

src/shared_modules/indexer_connector/src/monitoring.hpp
```diff
@@ -146,15 +150,32 @@ class Monitoring final
...
+                                                    "Cluster health status: '%s'",
+                                                    Utils::toUpperCase(fields.at(HealthCheckColumns::STATUS)).c_str());
                                             value.second = false;
                                         }
                                     }
                                     else
                                     {
+                                        logDebug2(IC_MONITORING,
```
pereyra-m, line 161: Improve: For the same reasons detailed above, I'd add the name of the server that failed.

src/shared_modules/indexer_connector/src/monitoring.hpp
```diff
@@ -146,15 +150,32 @@ class Monitoring final
...
+                                                  "Bad response from _cluster/health: %s",
```
pereyra-m, line 162: Nitpick: in fact, the endpoint is `_cat/health`

src/shared_modules/indexer_connector/src/monitoring.hpp
```diff
@@ -146,15 +150,32 @@ class Monitoring final
...
                                         value.second = false; // LCOV_EXCL_LINE
                                     }
                                 },
-                                [&](const std::string& error, const long /*statusCode*/) { value.second = false; },
+                                [&](const std::string& error, const long statusCode)
+                                {
+                                    std::string errorMessage = error;
+                                    constexpr auto NOT_USED {-1};
+
+                                    if (statusCode != NOT_USED)
```
pereyra-m, line 172: Question: why we avoid logging the error message in case `statusCode` is `NOT_USED` ?

src/shared_modules/indexer_connector/tests/unit/monitoring_test.cpp
pereyra-m, line 1: Improve: in this type of tests, I'd a flag to the logging callback to make sure it was executed.
Because if we check the content of the message but don't confirm it was called, it might fail silently.

pereyra-m (changes requested):
Gj!

I've left some comments.

src/shared_modules/indexer_connector/src/monitoring.hpp
MiguelazoDS: I left a comment on the issue about that. For it was enough with the log in ServerSelector.

src/shared_modules/indexer_connector/src/monitoring.hpp
```diff
@@ -146,15 +150,32 @@ class Monitoring final
                                         }
```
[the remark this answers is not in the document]
MiguelazoDS: done

src/shared_modules/indexer_connector/src/monitoring.hpp
MiguelazoDS: done

src/shared_modules/indexer_connector/src/monitoring.hpp
MiguelazoDS: I didn't like the idea of moving there, but I can do that, the NOT_USED flag should be fixed in http-request but I'd do the same if you agree.

src/shared_modules/indexer_connector/src/monitoring.hpp
MiguelazoDS: done

src/shared_modules/indexer_connector/src/monitoring.hpp
MiguelazoDS: done

src/shared_modules/indexer_connector/src/monitoring.hpp
MiguelazoDS: thanks

src/shared_modules/indexer_connector/tests/unit/monitoring_test.cpp
MiguelazoDS: Thanks. done

src/shared_modules/indexer_connector/src/monitoring.hpp
```diff
@@ -148,13 +151,28 @@ class Monitoring final
                                         {
                                             value.second = false;
                                         }
+                                        logDebug2(IC_NAME,
+                                                  "Cluster health status '%s' for server '%s'",
+                                                  Utils::toUpperCase(fields.at(HealthCheckColumns::STATUS)).c_str(),
```
sebasfalcone, line 156: Reduce overhead
```suggestion
                                                  fields.at(HealthCheckColumns::STATUS).c_str(),
```

src/shared_modules/indexer_connector/src/monitoring.hpp
```diff
@@ -148,13 +151,28 @@ class Monitoring final
...
+                                                  originalResponse.c_str());
                                         value.second = false; // LCOV_EXCL_LINE
                                     }
                                 },
-                                [&](const std::string& error, const long /*statusCode*/) { value.second = false; },
+                                [&](const std::string& error, const long statusCode)
+                                {
+                                    std::string errorMessage = error;
+
+                                    errorMessage += " (Status code: " + std::to_string(statusCode) + ")";
```
sebasfalcone, line 171: Improve (design): This could be a const
```suggestion
                                    const std::string errorMessage = error + ". Status code: " + std::to_string(statusCode);
```

sebasfalcone (approved):
## Tests

### Cluster green status
image
```
2024/08/13 18:33:57 indexer-connector[307816] monitoring.hpp:154 at operator()(): DEBUG: Cluster health status 'green' for server 'https://127.0.0.1:9200'
```

### Cluster yellow status
image

```
2024/08/13 18:32:57 indexer-connector[307816] monitoring.hpp:154 at operator()(): DEBUG: Cluster health status 'yellow' for server 'https://127.0.0.1:9200'
```

### Emulate a 500 error
I generated a python server which responds 500 on the healthcheck endpoint

```
2024/08/13 18:42:04 indexer-connector[323512] monitoring.hpp:171 at operator()(): DEBUG: HTTP response code said error. Status code: 500
```

src/shared_modules/indexer_connector/src/monitoring.hpp
pereyra-m: It was just a nitpick.
What you decide is Ok.

src/shared_modules/indexer_connector/src/monitoring.hpp
pereyra-m: I understand.
Thank you.

PR #25247, opened 2024-08-17
Changed the way the windows agent state file is updated

### LINKED ISSUE #25092: Windows agent does not indicate pending status during restart in stress tests
|---|---|---|---|---|
| 4.9.0-beta1 | Wazuh component | Agent | Packages | Windows |

## Description

After investigation in issue https://github.com/wazuh/wazuh-jenkins/issues/6842, due to a significant difference between the stress tests of [4.9.0-beta1](https://github.com/wazuh/wazuh/issues/25037) and [4.8.1-rc2](https://github.com/wazuh/wazuh/issues/24471) on the Windows agent, it has been possible to verify a rare behavior in the agent related to its state during a restart.
This difference in the graphs was already reported in alpha1 in this [issue](https://github.com/wazuh/wazuh/issues/24297), and it was closed because the problem was in the CSV file. That CSV file collects the agent state during the whole test from the `wazuh-agent.state` file. In versions prior to 4.9.0, the file recorded the `pending` state frequently, but since 4.9.0 this happens rarely, producing the difference in the graphs.

The difference is significant and can be observed with the following:

```shell
➜  egrep "pending" Test_stress_B5394_windows_agentd_state.csv | wc -l  # 4.8.0-rc2
1756
➜  egrep "pending" Test_stress_B5477_windows_agentd_state.csv | wc -l # 4.9.0-beta1
11
```

**4.9.0**

image

**4.8.1**

image

We suspect that there may be a problem with the state of the Windows agent during restart that may be causing these differences, as no changes have been made in the stress tests related to this problem, so this needs further research.

For more information see issue https://github.com/wazuh/wazuh-jenkins/issues/6842

cborla:
|Related issue|
|---|
|Close #25092|

## Tests

  - [x] Windows
- [x] Source installation
- [x] Source upgrade

- Memory tests for Linux

### DIFF
--- a/src/client-agent/state.c
+++ b/src/client-agent/state.c
@@ -66,6 +66,9 @@ int write_state() {
     struct tm tm = { .tm_sec = 0 };
     const char * status;
     char path[PATH_MAX - 8];
+
+    char path_temp[PATH_MAX + 1];
+    char path_backup[PATH_MAX + 9];  // Backup path
     char last_keepalive[1024] = "";
     char last_ack[1024] = "";
     int buffered_event;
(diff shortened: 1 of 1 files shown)


vikman90 (changes requested):
@cborla Good job with this PR. 

I have sincere doubts about whether it is worth putting this code in 4.9.0. First, because the error is in a test rather than in the product; second, because the rename operation actually has the same problem.

The agent already uses `wfopen()`: it allows other programs to open the file while the agent keeps it open. However, other programs —such as the test— could open the file without opening permissions for third parties, preventing the agent from updating the status.

Putting a backup file is generating a new file: it must be included in the upgrade list and in WPK. I think it's too strong a proposal.
Yes, I agree with atomic writing, because it behaves the same in UNIX and Windows. Now, being in 4.x, I would consider requesting that the test be improved.

Thanks for the proposal!

PR #25314, opened 2024-08-21
Use date from vcpkg and remove CMP

### LINKED ISSUE #25177: Use date from VCPKG
## Description

based on research, carried out in: https://github.com/wazuh/wazuh/issues/25073

When we wanted to incorporate the date dependency, it gave certain compilation errors and it was also found that tz.cpp, something that is compiled in the engine, is not distributed by vcpkg.

We should investigate if it is possible to remove the compilation of tz.cpp and use the library directly so that we can directly use date from vcpkg.

## Expected
- Use date library from vcpkg

juliancnn:
Closes #25177

This PR makes use of vcpkg for the date library, overlaying the package. It also removes CPM completely from the project as it is no longer needed.


NahuFigueroa97 (approved):
LGTM!

PR #25634, opened 2024-09-07
Modified regexp string related to diff binary in order to avoid matching of /dev/null and /dev/full character device files

### LINKED ISSUE #11808: Rootkit false positive on diffutils' `diff`
|---| --- | --- |
| 4.2.5 (also) | Rules/Decoders | New/Error/Improve |

## Description

There is a false positive when checking for rootkits on GNU diffutil's `diff`. Building `diff` from [upstream's sources](https://git.savannah.gnu.org/git/diffutils.git) yields a binary that matches the pattern, too:

```
 ~/git/diffutils $ grep -E 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' src/diff
grep: src/diff: binary file matches
```

The exact matching string is `/dev/full`, which comes from bundled gnulib and seems to have been in diffutils for quite a while:

```
 ~/git/diffutils $ strings src/diff | grep -E 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh'
/dev/full
 ~/git/diffutils $ grep -r '/dev/full'
grep: lib/stdopen.o: binary file matches
grep: lib/libdiffutils.a: binary file matches
tests/help-version:    if test -w /dev/full && test -c /dev/full; then
tests/help-version:      env $i --help    >/dev/full 2>/dev/null && fail=1
tests/help-version:      env $i --version >/dev/full 2>/dev/null && fail=1
tests/help-version:        echo "  with --help or --version output redirected to /dev/full" 1>&2
grep: src/sdiff: binary file matches
grep: src/cmp: binary file matches
grep: src/diff: binary file matches
grep: src/diff3: binary file matches
gnulib/lib/fwriteerror.c:/* Name of a file on which writing fails.  On systems without /dev/full,
gnulib/lib/fwriteerror.c:#define UNWRITABLE_FILE "/dev/full"
gnulib/lib/stdopen.c:             have /dev/full, by opening that write-only instead of
gnulib/lib/stdopen.c:          int full_fd = fd == STDIN_FILENO ? open ("/dev/full", mode) : -1;
 ~/git/diffutils $ git log -S'/dev/full'
commit e454c1471aaf8b3e6bead2313838aefebfe97bca
Author: Jim Meyering <meyering@redhat.com>
Date:   Sat Nov 14 10:31:41 2009 +0100

    tests: add the first script; hook up gnulib-tests

    * configure.ac (AC_CONFIG_FILES): Add tests/Makefile and
    gnulib-tests/Makefile.
    * tests/help-version: New file, from coreutils.
    * tests/test-lib.sh: Likewise.
    * tests/Makefile.am: New file, from gzip.
    * Makefile.am (SUBDIRS): Add tests and gnulib-tests.
    * tests/t-local.sh: New, empty(for now) file.
 ~/git/diffutils $ git tag --contains e454c1471aaf8b3e6bead2313838aefebfe97bca
v2.9
v3.0
v3.1
v3.2
v3.3
v3.4
v3.5
v3.6
v3.7
v3.8
```

### Service/Product/Module

https://github.com/wazuh/wazuh/blob/master/ruleset/rootcheck/db/rootkit_trojans.txt#L43

### Current results

In Kibana, I see a rootcheck event: `Trojaned version of file '/usr/bin/diff' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).`

### Expected results

There should be no event, becaues the file is not trojaned.

### LINKED ISSUE #13278: False positive on "Trojaned version of file detected."
|---| --- | --- |
| 4.2 | Rule: 510 | False-Positive |

## Description
Wazuh rule 510 detects the /bin/diff of Ubuntu 22.04 as trojaned version.

### Service/Product/Module
Rule 510

### Current results
`Trojaned version of file '/bin/diff' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).`

### Expected results
No detection at all.

### Log source / integration
`Trojaned version of file '/bin/diff' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).`

### Log reference
_none_

### Log examples
`Trojaned version of file '/bin/diff' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).`

### Threats and compliance
_none_

### LINKED ISSUE #19000: /bin/diff file falsely recognized as malicious
On working with Wazuh, I found out that /bin/diff is repeatedly recognized as infected by rootkit.
Analysing the regex on the rule below, i found out what is the problem.

https://github.com/wazuh/wazuh/blob/17469f88206a90e54e0d5f6ee81a0a04996ada1f/ruleset/rootcheck/db/rootkit_trojans.txt#L43C34-L43C34

The problem is that new version of diff contain a reference to the `/dev/full` pseudo-device. And the rule triggers if any `/dev` file are on the executable, except by those starting by **n**, probably by referencing `/dev/null`.

My suggestion is to modify the rule from:
`bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh`
to:
`bash|^/bin/sh|file\.h|proc\.h|/dev/[^nf]|^/bin/.*sh`

### LINKED ISSUE #19346: Trojaned version of file '/bin/diff' detected
Hello guys!

This is the issue was opened long time ago here https://github.com/ossec/ossec-hids/issues/2020 so I believe no one from Wazuh is watching/working on it, so I'm posting this again here.

Received From: archvbox->rootcheck
Rule: 510 fired (level 7) -> "Host-based anomaly detection event (rootcheck)."
Portion of the log(s):

Trojaned version of file '/bin/diff' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).

--END OF NOTIFICATION

Received From: archvbox->rootcheck
Rule: 510 fired (level 7) -> "Host-based anomaly detection event (rootcheck)."
Portion of the log(s):

Trojaned version of file '/sbin/diff' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).

--END OF NOTIFICATION

Received From: archvbox->rootcheck
Rule: 510 fired (level 7) -> "Host-based anomaly detection event (rootcheck)."
Portion of the log(s):

Trojaned version of file '/usr/bin/diff' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).

--END OF NOTIFICATION

Received From: archvbox->rootcheck
Rule: 510 fired (level 7) -> "Host-based anomaly detection event (rootcheck)."
Portion of the log(s):

Trojaned version of file '/usr/sbin/diff' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).

djosip:
Newer diff binaries on various operating systems now contain string /dev/full, not just /dev/null. This change will fix false positives that report diff binary as a trojaned version of /bin/diff.

Related issues:
Closes #13278
Closes #19000
Closes #19346
Closes #11808

contribution

## Description
The root of the problem is that newer diff binaries contain not just /dev/null string but also /dev/full string.

The command `strings /usr/bin/diff | grep /dev/` can be used to check this.

In order to fix the false postives, regexp for the diff binary has been changed in order to exclude both, /dev/null and /dev/full character device files.

### DIFF
--- a/ruleset/rootcheck/db/rootkit_trojans.txt
+++ b/ruleset/rootcheck/db/rootkit_trojans.txt
@@ -40,7 +40,7 @@ sudo        !satori|vejeta|conf\.inv!
 crond       !/dev/[^nt]|bash!
 gpm         !bash|mingetty!
 ifconfig    !bash|^/bin/sh|/dev/tux|session.null|/dev/[^cludisopt]!
-diff        !bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh!
+diff        !bash|^/bin/sh|file\.h|proc\.h|/dev/[^nf]ull|^/bin/.*sh!
 md5sum      !bash|^/bin/sh|file\.h|proc\.h|/dev/|^/bin/.*sh!
 hdparm      !bash|/dev/ida!
 ldd         !/dev/[^n]|proc\.h|libshow.so|libproc.a!


iasdeoupxe:
> Related issues:
> Closes #13278
> Closes #19000
> Closes #19346
> Closes #11808

If you use it like this GitHub should automatically close the issues once this is merged: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

djosip:
Thanks.
Fixed the comment.

iasdeoupxe:
@wazuh team: Any chance to get this reviewed and merged? Seems like minor change to review and would solve / close four issues, two open since 2+ years.

brownowski:
We're hitting this problem as well. Why did you add in the 'ull' to the rule?

My reading of it is that the rule should try to catch anything referencing files in /dev/, but it should override where there is a reference to /dev/null?

The original rule looks allow a reference to any devices that begin with an 'n'. Your rule will only match references to devices that are named '?ull', where ? is not an n or f. Neither strike me as complete expressions for matching /dev/* but not /dev/null or /dev/full.

Even just changing it to /dev/[^nf], without the 'ull' at the end would probably be better.

Looking back at the history of the rootkit_trojans.txt file, it's barely had a meaningful update over 10 years. It's hard to know how these rules were determined or how effective they even are. Seems to be mostly generating false positives.

djosip:
> We're hitting this problem as well. Why did you add in the 'ull' to the rule?

My intention is to exclude precisely the string "/dev/null" and the string "/dev/full" while retaining the logic of the original rule (matching everything else not referenced by the diff binary).

> My reading of it is that the rule should try to catch anything referencing files in /dev/, but it should override where there is a reference to /dev/null?

Almost.
It should match anything referencing files in /dev/ directory while ignoring references to /dev/null and /dev/full.
Older versions of diff used to reference only /dev/null but newer versions in addition to /dev/null also reference the file /dev/full which causes the false positive we all suffer from.

> The original rule looks allow a reference to any devices that begin with an 'n'. Your rule will only match references to devices that are named '?ull', where ? is not an n or f. Neither strike me as complete expressions for matching /dev/* but not /dev/null or /dev/full.

Exactly.
If you use the `strings` command to check strings contained in the diff binary, you will find that only files in /dev it references, are /dev/null and /dev/full, thus, these are the only files you would want to exclude.

The whole point of using this rule is to get alerted in case the diff binary starts to reference files in /dev/ directory that are not normally referenced by the uncompromised diff binary.

> Even just changing it to /dev/[^nf], without the 'ull' at the end would probably be better.

I don't agree.
Such rule would ignore not just /dev/null and /dev/full but also anything that starts with /dev/n or /dev/f.
For example, this would ignore references to:
- /dev/null
- /dev/full
- /dev/nvram
- /dev/net
- /dev/net/tun
- /dev/fb0
- /dev/fd0
- /dev/fuse
... just to name few files I can see on my system.

You most probably wouldn't want to ignore all of these files as the current diff binary only references /dev/null and /dev/full. Anything else should trigger an alarm.

> Looking back at the history of the rootkit_trojans.txt file, it's barely had a meaningful update over 10 years. It's hard to know how these rules were determined or how effective they even are. Seems to be mostly generating false positives.

The lack of maintenance in the past cannot be used as an excuse for not fixing or enhance the rules and leaving the existing false positives.
After all, this an OpenSource project and the community is entitled to ask for changes or to make contributions to the project.

vikman90:
@djosip Thank you very much for your proposal to address this long-standing issue.

I'm going to accept your PR on an intermediate branch, so that it becomes part of issue:
- #26137

This way, we'll merge it into the 4.10 branch along with the fix for `apachectl`.

Thanks again and best regards!

brownowski:
> I don't agree.
> Such rule would ignore not just /dev/null and /dev/full but also anything that starts with /dev/n or /dev/f.
> For example, this would ignore references to:
> 
>     /dev/null
>     /dev/full
>     /dev/nvram
>     /dev/net
>     /dev/net/tun
>     /dev/fb0
>     /dev/fd0
>     /dev/fuse
>     ... just to name few files I can see on my system.

You missed the issue. Not only does your rule ignore those too, but pretty much everything else in dev. Your regex only searches for files matching the pattern /dev/?ull, excluding null or full, but also excluding /dev/block/..., /dev/bus/..., /dev/loop..., /dev/sd.... Anything that doesn't have a 'ull' after the first character.

> The lack of maintenance in the past cannot be used as an excuse for not fixing or enhance the rules and leaving the existing false positives.
> After all, this an OpenSource project and the community is entitled to ask for changes or to make contributions to the project.

This wasn't a comment about not making any improvement to it and reducing a false positive. However, I was questioning if this gives a false sense of security that this rootkit detection is doing something, when it mainly just looks to be generating false positives (which are the only edits that have been made to the file, apart from changing licensing or copyright information, in over 10 years).

iasdeoupxe:
Always misinterpreted the syntax of the file here and thought this is a "not included" one (due to the `!`) but rootkit_trojans.txt actually includes on top:

```
# Each line must be in the following format:
# file_name !string_to_search!Description
```

This means what @brownowski is describing seems to be a valid concern:

The original `/dev/[^n]` regex had not reported for the valid `/dev/null` but also caused false negatives for things like e.g. `/dev/net`.

But while fixing the false positive for `/dev/full` the new regex `/dev/[^nf]full` seems to has increased the false negative rate because it now only reports anything in `/dev` which:
1. Doesn't start with a `n`
2. Doesn't start with a `f`
3. Includes `ull`

https://regexper.com/#%5C%2Fdev%5C%2F%5B%5Enf%5Dull

iasdeoupxe:
Due to the limitation in regex concerning "not" matching the only change is probably to use `/dev/[^nf]` like suggested in https://github.com/ossec/ossec-hids/issues/2020#issuecomment-1418331744:

```
diff        !bash|^/bin/sh|file\.h|proc\.h|/dev/[^nf]|^/bin/.*sh!
```

It still cause false negatives for e.g. `/dev/fb0` or `/dev/fuse`  but don't know how to solve that false positive otherwise without introducing additional false negatives.

iasdeoupxe:
Opened another PR: #26160

djosip:
> > I don't agree.
> > Such rule would ignore not just /dev/null and /dev/full but also anything that starts with /dev/n or /dev/f.
> > For example, this would ignore references to:
> > ```
> > /dev/null
> > /dev/full
> > /dev/nvram
> > /dev/net
> > /dev/net/tun
> > /dev/fb0
> > /dev/fd0
> > /dev/fuse
> > ... just to name few files I can see on my system.
> > ```
> 
> You missed the issue. Not only does your rule ignore those too, but pretty much everything else in dev. Your regex only searches for files matching the pattern /dev/?ull, excluding null or full, but also excluding /dev/block/..., /dev/bus/..., /dev/loop..., /dev/sd.... Anything that doesn't have a 'ull' after the first character.

@brownowski
You are completely right.
I'll look into it as soon as possible.

brownowski:
I believe something like the below would minimise corner cases.

/dev/[^nf]|/dev/[nf][^u]|/dev/[nf]u[^l]|/dev/[nf]ul[^l]

djosip:
Here is the proposal for a new solution.

Original string:
`diff        !bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh!`
New line:
`diff        !bash|^/bin/sh|file\.h|proc\.h|/+dev/+[nf]ull|^/bin/.*sh!`

In this case, the part of the rule `/+dev/+[nf]ull` would match /dev/null or /dev/full, covering the case where multiple slashes (e.g. //dev/null) are used by a trojan to avoid detection.

It might be a good idea to replace all slashes with `/+`.

djosip:
> Um, that is not a good suggestion. I feel like you still haven't understood the file correctly.

Yes, it's a bit confusing.

> The file is for matching strings in the binary file. The idea is to search for access to /dev/ but exclude /dev/null or /dev/full. [nf]ull as a search string is the exact opposite of what is needed.
> 
> As I stated above, if you want to avoid corner cases, something like the below:
> 
> New line: `diff !bash|^/bin/sh|file\.h|proc\.h|/dev/[^nf]|/dev/[nf][^u]|/dev/[nf]u[^l]|/dev/[nf]ul[^l]|^/bin/.*sh!`

What about something like this?
`^/dev/(?![nf]ull).*$`

This is a valid PCRE regexp which uses negative lookahead but I will have to test it with the Wazuh and check the results.

With the input:
/dev/full
/dev/null
/dev/tuse
/dev/fuse
/usr/share
/dev/blah
/ev/tmp

It would match:
/dev/tuse
/dev/fuse
/ev/tmp

It would match everything in /dev/ except null and full which is what we want.

brownowski:
> This is a valid PCRE regexp which uses negative lookahead but I will have to test it with the Wazuh and check the results.

According to the docs, there is only a small subset of regex Wazuh supports with these rules, and that doesn't include negative lookahead. Which then leads to the solution which I've already stated.

`/dev/[^nf]|/dev/[nf][^u]|/dev/[nf]u[^l]|/dev/[nf]ul[^l]`

My solution would match:
Anything in /dev/, except for devices starting with n or f,
Then all devices starting with n or f, but not followed by u,
Then all devices starting with n or f, then u, but not followed by l,
Then all devices starting with n or f, then u, then l, but not followed by l

So collectively, anything in /dev, except devices with names beginning with null or full.

Also, I don't think having ^ and $ (start and end delimiters) would fit with that context either.

Issue #25547, opened 2024-09-02
Error in stringHelper function


sebasfalcone:
## Description
It was discovered that the **[replaceAll](https://github.com/wazuh/wazuh/blob/bde6280610f6571baad79db0e4f344cd41f76af3/src/shared_modules/utils/stringHelper.h#L70-L83)** function from **stringHelper** has a bug (see steps to reproduce)

The proposed solution is as follows:
```C++
static bool replaceAll(std::string& data, const std::string& toSearch, const std::string& toReplace)
{
    bool found = false;
    auto pos = data.find(toSearch);

    while (pos != std::string::npos)
    {
        data.replace(pos, toSearch.size(), toReplace);
        found = true;
        pos = data.find(toSearch, pos + toReplace.size());
    }

    if (found && toSearch != toReplace)
    {
        while (replaceAll(data, toSearch, toReplace)) {}
    }

    return found;
}
```

## Steps to reproduce
[This code example](https://godbolt.org/z/dx66fe5e8) shows how to reproduce this issue (The **toReplace** string is a subset of the **toSearch** string)

## DoD
- [x] Implement the fix
- [x] Validate the fix
- [x] Update UTs

sebasfalcone:
## Updated ETA
Delayed due to issues related to 4.9.1 and 4.10.0 releases being pushed to completion

GabrielEValenzuela:
#  Update

- Co-implementation with @sebasfalcone complete

Issue #25717, opened 2024-09-16
Improve indexer healthcheck response parsing


sebasfalcone:
# Description
While performing this task:
- https://github.com/wazuh/wazuh/pull/25417

It was found that the current response from the indexer could be json formatted, instead of plain text. This was happening because we were overwriting the default headers for the http-request library ([see comment](https://github.com/wazuh/wazuh/pull/25417#issuecomment-2350731611)), which by default sets the response to json format

To address this issue we should align with the changes present on the monitor class to [this commit](https://github.com/wazuh/wazuh/pull/25417/commits/ed8c1b88a32c415499e2f203a1b1e80ea3abda36)

> [!IMPORTANT]
> [This tests](https://github.com/wazuh/wazuh/issues/25415) should be aligned to comply with the new response format

# DoD
- [x] New response implemented
- [x] Updated UTs

PR #25739, opened 2024-09-17
Fix replaceAll util

sebasfalcone:
|Related issue|
|---|
| #25547 |

## Description
This PR fixes the implementation of the replaceAll helper function when the `toReplace` string is a subset of the `toSearch` string

### DIFF
--- a/src/engine/source/base/src/utils/stringUtils.cpp
+++ b/src/engine/source/base/src/utils/stringUtils.cpp
@@ -204,7 +204,7 @@ bool isNumber(const std::string& str)
 
 bool replaceAll(std::string& data, const std::string_view toSearch, const std::string_view toReplace)
 {
-    if (toSearch.empty() || toSearch == toReplace)
+    if (toSearch.empty() || toSearch == toReplace || toReplace.find(toSearch) != std::string_view::npos)
     {
         return false; // Nothing to search for if toSearch is empty
     }
--- a/src/engine/source/base/src/utils/stringUtils.cpp
+++ b/src/engine/source/base/src/utils/stringUtils.cpp
@@ -210,9 +210,8 @@ bool replaceAll(std::string& data, const std::string_view toSearch, const std::s
     }
 
     bool found = false;
-    size_t pos = data.find(toSearch);
 
-    if (pos != std::string::npos)
+    if (size_t pos = data.find(toSearch); pos != std::string::npos)
     {
         found = true;
         data.replace(pos, toSearch.length(), toReplace);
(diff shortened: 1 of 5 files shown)


sebasfalcone (commented):
LGTM!

## Testing
Indexer status response is being parsed as expected ([see testing](https://godbolt.org/z/xd83sjKos))

```
rows size 2
field size 15
```

> [!NOTE]
> The parsing will be changed to json [here](https://github.com/wazuh/wazuh/issues/25717)

sebasfalcone:
## Low coverage
Modules that have low coverage will be deprecated (those are string helper functions that are duplicated)

src/engine/source/base/src/utils/stringUtils.cpp
```diff
@@ -204,16 +204,23 @@ bool isNumber(const std::string& str)
...
+    if (toSearch.empty() || toSearch == toReplace)
+    {
+        return false; // Nothing to search for if toSearch is empty
+    }
+
+    bool found = false;
+    size_t pos = data.find(toSearch);
 
-    while (std::string::npos != pos)
+    if (pos != std::string::npos)
```
Dwordcito, line 215:
```suggestion
    if (size_t pos = data.find(toSearch); pos != std::string::npos)
```

GabrielEValenzuela: Done. Thanks!

src/engine/source/base/src/utils/stringUtils.cpp
```diff
@@ -204,16 +204,22 @@ bool isNumber(const std::string& str)
 
 bool replaceAll(std::string& data, const std::string_view toSearch, const std::string_view toReplace)
 {
-    auto pos {data.find(toSearch)};
-    const auto ret {std::string::npos != pos};
+    if (toSearch.empty() || toSearch == toReplace)
```
Dwordcito, line 207:
```suggestion
    if (toSearch.empty() || toSearch == toReplace || toReplace.find(toSearch) != std::string_view::npos)
```

Dwordcito: to avoid this issue:

add test with 

```
    std::string input = "a";
    replaceAll(input, "a", "aa");
```

sebasfalcone: Indeed it does
image

GabrielEValenzuela: Thank you both! Good catch!

Dwordcito (approved):

PR #25741, opened 2024-09-17
Make the ID field optional for the indexerConnector queries

### LINKED ISSUE #25541: ID as optional field in the indexer connector
## Description
The new uses that the `indexerConnector` will endure, in the engine, forced us to make the `_id` field for queries optional (so OpenSearch can assign it on its own)

The following methods should take into account this new requirement:
https://github.com/wazuh/wazuh/blob/bde6280610f6571baad79db0e4f344cd41f76af3/src/shared_modules/indexer_connector/src/indexerConnector.cpp#L165-L186

> [!TIP]
> We could use the [builder pattern](https://refactoring.guru/design-patterns/builder) to make the inferace to this queries ([see example](https://github.com/wazuh/wazuh/blob/8f93ae377aa120ed58d16b17a4d9af50c1cd6f5e/src/shared_modules/utils/wazuhDBQueryBuilder.hpp))

## DoD
- [x] Feature implemented
- [x] UTs updated

pereyra-m:
|Related issue|
|---|
|Closes #25541|

## Description

This PR allows to index elements that don't contain an ID field letting the indexer to create one.

## Tests

  - [x] Linux

The changes were validated manually using the test tool and a Wazuh docker deployment.
With these configuration files and command

<details><summary>config.json</summary>
<p>

```json
{
    "name": "wazuh-states-vulnerabilities-wazuh.manager",
    "enabled": "yes",
    "hosts": [
        "https://wazuh.indexer:9200"
    ],
    "ssl" : {
        "certificate_authorities" : ["/etc/filebeat/certs/root-ca.pem"],
        "certificate" : "/etc/filebeat/certs/filebeat.pem",
        "key" : "/etc/filebeat/certs/filebeat-key.pem"
    },
    "username" : "admin",
    "password" : "SecretPassword"
}
```

</p>
</details> 

<details><summary>events.json</summary>
<p>

```json
{
    "operation": "INSERTED",
    "data": {
        "package": {
            "name": "testPackage"
        }
    }
}
```

</p>
</details> 

```
./indexer_connector_tool -c config.json -e events.json -t /workspaces/wazuh.worktrees/master/src/engine/source/indexerconnector/qa/test_data/template.json
```

The test package was indexed with an ID assigned automatically

> [!Note]
> Using the indexer tool, I found that the logging isn't working properly for the stdout. This could be solved at another issue
```
root@fb659c773250:/workspaces/wazuh.worktrees/master/src/engine/build/source/indexerconnector/tool# ./indexer_connector_tool -c config.json -e events.json -t /workspaces/wazuh.worktrees/master/src/engine/source/indexerconnector/qa/test_data/template.json
Press enter to stop the indexer connector tool... 
2024-09-19 14:44:12.854 32232:32252 indexerConnector.cpp:182 at operator()::<lambda>handleSuccessfulPostResponse(): debug: Response: %s
```

### DIFF
--- a/src/engine/source/indexerconnector/src/indexerConnector.cpp
+++ b/src/engine/source/indexerconnector/src/indexerConnector.cpp
@@ -93,14 +93,26 @@ static void builderBulkDelete(std::string& bulkData, std::string_view id, std::s
 
 static void builderBulkIndex(std::string& bulkData, std::string_view id, std::string_view index, std::string_view data)
 {
-    bulkData.append(R"({"index":{"_index":")");
-    bulkData.append(index);
-    bulkData.append(R"(","_id":")");
-    bulkData.append(id);
-    bulkData.append(R"("}})");
-    bulkData.append("\n");
-    bulkData.append(data);
-    bulkData.append("\n");
+    if (!id.empty())
+    {
+        bulkData.append(R"({"index":{"_index":")");
+        bulkData.append(index);
+        bulkData.append(R"(","_id":")");
+        bulkData.append(id);
+        bulkData.append(R"("}})");
+        bulkData.append("\n");
+        bulkData.append(data);
+        bulkData.append("\n");
+    }
+    else
+    {
+        bulkData.append(R"({"index":{"_index":")");
+        bulkData.append(index);
+        bulkData.append(R"("}})");
+        bulkData.append("\n");
+        bulkData.append(data);
+        bulkData.append("\n");
+    }
 }
 
 IndexerConnector::IndexerConnector(const nlohmann::json& config, const uint32_t& timeout, const uint8_t workingThreads)
@@ -146,14 +158,15 @@ IndexerConnector::IndexerConnector(const nlohmann::json& config, const uint32_t&
                 auto data = dataQueue.front();
                 dataQueue.pop();
                 auto parsedData = nlohmann::json::parse(data);
-                const auto& id = parsedData.at("id").get_ref<const std::string&>();
 
                 if (parsedData.at("operation").get_ref<const std::string&>().compare("DELETED") == 0)
                 {
+                    const auto& id = parsedData.at("id").get_ref<const std::string&>();
                     builderBulkDelete(bulkData, id, m_indexName);
                 }
                 else
                 {
+                    const auto& id = parsedData.contains("id") ? parsedData.at("id").get_ref<const std::string&>() : "";
                     const auto dataString = parsedData.at("data").dump();
                     builderBulkIndex(bulkData, id, m_indexName, dataString);
                 }
--- a/src/engine/source/indexerconnector/src/indexerConnector.cpp
+++ b/src/engine/source/indexerconnector/src/indexerConnector.cpp
@@ -93,26 +93,19 @@ static void builderBulkDelete(std::string& bulkData, std::string_view id, std::s
 
 static void builderBulkIndex(std::string& bulkData, std::string_view id, std::string_view index, std::string_view data)
 {
+    bulkData.append(R"({"index":{"_index":")");
+    bulkData.append(index);
+
     if (!id.empty())
     {
-        bulkData.append(R"({"index":{"_index":")");
-        bulkData.append(index);
         bulkData.append(R"(","_id":")");
         bulkData.append(id);
-        bulkData.append(R"("}})");
-        bulkData.append("\n");
-        bulkData.append(data);
-        bulkData.append("\n");
     }
-    else
-    {
-        bulkData.append(R"({"index":{"_index":")");
-        bulkData.append(index);
-        bulkData.append(R"("}})");
-        bulkData.append("\n");
-        bulkData.append(data);
-        bulkData.append("\n");
-    }
+
+    bulkData.append(R"("}})");
+    bulkData.append("\n");
+    bulkData.append(data);
+    bulkData.append("\n");
 }
 
 IndexerConnector::IndexerConnector(const nlohmann::json& config, const uint32_t& timeout, const uint8_t workingThreads)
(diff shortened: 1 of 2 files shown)


sebasfalcone (approved):
LGTM!

src/engine/source/indexerconnector/include/indexerConnector/indexerConnector.hpp
```diff
@@ -98,6 +101,78 @@ class EXPORTED IndexerConnector final : public IIndexerConnector
      * @copydoc IIndexerConnector::publish
      */
     void publish(const std::string& message) override;
+
+    /**
+     * @brief Auxiliary class to create the queries that will be sent to the indexer.
+     *
+     */
+    class QueryBuilder final : public base::utils::patterns::Builder<QueryBuilder>
```
Dwordcito, line 109: 1- Let's remove this from the include, this is not part of what we want to expose with the library. Create a new class.

2- I think there is too much abstraction for this, it seems better to implement two classes.


1. To delete
2. For bulkindex

The delete one has as input the index and the id, therefore you could build the string in the build method using string_views that are attributes of the class.

For bulkindex, it has 3 input data, the index, the id and data. Since id is optional, you could take the if to the build() method.

It would look like this:
bulkData.append(BulkIndexBuilder::builder()
                            .index(index)
                            .id(id)
                            .rawData(data)
                            .build());
                            
bulkData.append(DeleteQueryBuilder::builder()
                        .index("_index", index)
                        .id("_id", id)
                        .build());

src/engine/source/indexerconnector/src/indexerConnector.cpp
```diff
@@ -81,26 +81,54 @@ static void initConfiguration(SecureCommunication& secureCommunication, const nl
...
-    bulkData.append(R"({"index":{"_index":")");
-    bulkData.append(index);
-    bulkData.append(R"(","_id":")");
-    bulkData.append(id);
-    bulkData.append(R"("}})");
-    bulkData.append("\n");
-    bulkData.append(data);
-    bulkData.append("\n");
+
+    if (id.empty())
```
Dwordcito, line 104: This conditional could be a part of the build() method.

Dwordcito (changes requested):

src/engine/source/indexerconnector/include/indexerConnector/indexerConnector.hpp
pereyra-m: Hi @Dwordcito 

1- I created a nested class because the query builder didn't seem generic enough to be an util. I can declare the classes at `indexerConnector.cpp`

2- If I create a class that is specific to build an insert/delete query, the builder pattern might not be the most convenient choice then in my opinion. A simple class with a constructor and default parameters would do the trick. 

This is because I need to allow the user to change the order of the methods or omit some, from the outside all of them should be valid

```c++

bulkData.append(BulkIndexBuilder::builder().index(index).id(id).rawData(data).build());
bulkData.append(BulkIndexBuilder::builder().id(id).index(index).rawData(data).build());
bulkData.append(BulkIndexBuilder::builder().rawData(data).index(index).id(id).build());
bulkData.append(BulkIndexBuilder::builder().rawData(data).index(index).build());
```

To cover this situation, I can't construct the string appending the result of each method but saving each value until the end (rawData for example is out of the JSON and adds new line characters).

The resulting "builder" would be really similar to a simple `bulkData.append(BulkIndexBuilder(index,id,rawData)` that already has the two types of possible resulting queries stored and replaces the placeholders.  

Another option is to simply rollback all the approach and make a tiny change at `builderBulkIndex` using the same handcrafted mechanism that is implemented right now. This would be straightforward.

Dwordcito: 1- No, you must give it enough entity and create a new header if necessary.

2- Talk to your TL if you want to change the approach, my suggestions are, but as it stands I won't approve it.

sebasfalcone: After discussing this in the Daily we concluded to aim for the minimalistic approach

Even though we think a proper query-generating class will be for the best, we don't have the appropriate time to implement such a feature

pereyra-m: 1 - I'll consider this in the future.

src/engine/source/indexerconnector/src/indexerConnector.cpp
pereyra-m: According to the discussion above, the approach will be changed.

src/engine/source/indexerconnector/src/indexerConnector.cpp
```diff
@@ -93,14 +93,26 @@ static void builderBulkDelete(std::string& bulkData, std::string_view id, std::s
...
+    }
+    else
+    {
+        bulkData.append(R"({"index":{"_index":")");
+        bulkData.append(index);
+        bulkData.append(R"("}})");
+        bulkData.append("\n");
+        bulkData.append(data);
+        bulkData.append("\n");
+    }
```
sebasfalcone, line 115: This can be simplified
```suggestion
    bulkData.append(R"({"index":{"_index":")");
    bulkData.append(index);
    
    if (!id.empty()) {
        bulkData.append(R"(","_id":")");
        bulkData.append(id);
    }
    
    bulkData.append(R"("}})");
    bulkData.append("\n");
    bulkData.append(data);
    bulkData.append("\n");
```

pereyra-m: You are right. I'll commit the suggestion.

sebasfalcone (approved):

Dwordcito (approved):

Issue #25579, opened 2024-09-04
Engine - Develop the Metrics Module for Wazuh-Engine


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0 | Engine | Manager | Packages/Sources | OS version |

## Description
Develop the foundational metrics module that will be responsible for collecting and aggregating key metrics within the Wazuh-Engine. This module will serve as the backbone for telemetry data collection, processing, and reporting. It should be designed to be highly configurable, allowing users to enable or disable specific metrics depending on their observability needs.

The metrics module should be integrated into the main Wazuh-Engine workflow, ensuring that telemetry data is collected from key components like the Server, KVDB, API, and Router. Additionally, the module should support modular enable/disable functionality for performance testing, ensuring that telemetry can be turned off entirely or partially to reduce resource usage during specific operations.

## Tasks:
- [x] Design and implement a foundational metrics module for Wazuh-Engine.
- [ ] Integrate the metrics module into key Wazuh-Engine components (Server, KVDB, API, Router).
- [x] Implement configuration options to allow users to enable/disable individual metrics.
- [ ] Conduct performance testing with telemetry enabled/disabled to assess the impact on system resources.

JavierBejMen:
## Update
- Deleted Old metrics implementation
- Working on Manager instantiation and configuration

JavierBejMen:
## Update
- Working on SingletonLocator pattern for testing and managing singletons in production
- Added opentelemetry tearup and teardown to manager class

JavierBejMen:
## Update
- Finished SingletonLocator 
- Added hot configuration reload
- Added tests for enable, disable, reload and configure
- Working on metrics creation/getters

JavierBejMen:
## Update
- Implemented Metrics registration
- Metrics can be enabled/disabled
- Working on tests

JavierBejMen:
## Update
- Reimplemented Metric interface and Metric implementation for simplicity
- Working on tests

JavierBejMen:
## Update
- Deleted old metrics module (deleted cmd and API) deleted/commented metrics usage in server, router and queue
- Deleted Old metrics fakes
- Added unit tests for all new metrics manager
- Segregated metrics manager interfaces
- Added multithreaded manager safety
- Working on multithreaded tests and component tests

JcabreraC:
ETA expanded. The development of the module had more complexity than expected and has been underestimated.

[...]

PR #25758, opened 2024-09-18
Refactors Engine Metrics module

JavierBejMen:
|Related issue|
|---|
|#25579|

This PR refactors the Engine Metrics module to use the new Indexer exporter and deals with performance, testing issues and readability of the old implementation.

## Changes
- Added singleton locator service to allow different singleton instance management in testing and production.
- Added helper classes for parallel testing in base module
- Added metrics module 
- Deleted old metrics module 
- Deleted metrics command and API
- Updated modules to not use metrics


juliancnn (changes requested):
LGTM!, but the code style needs to be fixed. On the other hand, it is possible to use the imetri and imanager interfaces in the UT, I think that in this way the coverage of the virtual functions will be marked?

JcabreraC (approved):
LGTM !

juliancnn (approved):
LGTM

Issue #22880, opened 2024-04-10
AWS Security Lake - Source Version 2


kclinden:
| Affected integration |
|---|
| AWS|

## Description
When integrating with AWS Security Lake with [Source Version 2](https://docs.aws.amazon.com/security-lake/latest/userguide/subscriber-query-examples2.html) I am getting the following error:
```
DEBUG: Processing file aws/SH_FINDINGS/2.0/region=us-west-2/accountId=123456789000/eventDay=20240410/6cc7641e31d4d10209bed021dd7ebe35.gz.parquet in aws-security-data-lake-us-east-1-txlvnteculqdhzkmgk89pghsxqrlmu
DEBUG: +++ Error: Object of type datetime is not JSON serializable
Unknown error: Object of type datetime is not JSON serializable
Traceback (most recent call last):
  File "/var/ossec/wodles/aws/aws-s3.py", line 4250, in <module>
    main(sys.argv[1:])
  File "/var/ossec/wodles/aws/aws-s3.py", line 4237, in main
    asl_queue.sync_events()
  File "/var/ossec/wodles/aws/aws-s3.py", line 3797, in sync_events
    self.bucket_handler.process_file(message["route"])
  File "/var/ossec/wodles/aws/aws-s3.py", line 3667, in process_file
    events_in_file = self.obtain_logs(bucket=message_body['bucket_path'],
  File "/var/ossec/wodles/aws/aws-s3.py", line 3655, in obtain_logs
    events.append(json.dumps(j))
  File "/var/ossec/framework/python/lib/python3.9/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/var/ossec/framework/python/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/var/ossec/framework/python/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/var/ossec/framework/python/lib/python3.9/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type datetime is not JSON serializable
[root@wazuh-manager-0 logs]# vim /var/ossec/wodles/aws/aws-s3.py[root@wazuh-manager-0 logs]# vim /var/ossec/wodles/aws/aws-s3.py
```

## Tasks
- [x] Test in a manager.
- [x] Test in an agent.
[...]

PR #25761, opened 2024-09-18
Integrate AWS Security Lake with v2.0

javiersanchz:
|Related issue|
|---|
| #22880  |

## Description

This PR closes https://github.com/wazuh/wazuh/issues/22880, where work and testing are done to ensure that `Security Lake` is compatible with both `v1.0` and `v2.0`.


GGP1 (changes requested):
We should update the `test_aws_sl_subscriber_bucket_obtain_logs` unit test with a case that parses events with a datetime type (which would have raised an error before this change) and validates no exception is thrown.

fdalmaup (changes requested):
We should check the content of the `events` before these changes to verify that we are not breaking the current behavior of the module and the events sent to Analysisd.

fdalmaup (approved):
LGTM!

Issue #26015, opened 2024-09-27
Indexer Connector - Drop invalid messages


sebasfalcone:
## Description
As a new requirement, the indexer-connector module needs to reject non-json data that is received

To achieve this, we need to set the `allow_exeptions` flag to false while parsing the incoming data;
https://github.com/wazuh/wazuh/blob/f79cdb0e72ed19a2159e32c6e162e0dd15b8d365/src/shared_modules/indexer_connector/src/indexerConnector.cpp#L401

And discard the message

## DoD
- [x] Non-JSON data is now discarded on the indexer
- [x] Updated UTs

GabrielEValenzuela:
# Update

- Task complete and in review process.

PR #26087, opened 2024-10-01
Drop invalid mesages of `indexerConnector`

GabrielEValenzuela:
|Related issue|
|---|
|[26015](https://github.com/wazuh/wazuh/issues/26015)|

# Objective

This PR aims to enable `json::parser` function to accept a bool parameter allow_exceptions which controls whether an exception is thrown when a parse error occurs (true, default) or whether a discarded value should be returned (false).
This behavior is similar to the function `accept()`, which can be used which does not return a JSON value, but a bool indicating whether the input is valid JSON.

In any case, no diagnostic information available in this scenario.

### DIFF
--- a/src/engine/source/indexerconnector/src/indexerConnector.cpp
+++ b/src/engine/source/indexerconnector/src/indexerConnector.cpp
@@ -17,7 +17,6 @@
 #include "secureCommunication.hpp"
 #include "serverSelector.hpp"
 #include <fstream>
-#include <ostream>
 
 constexpr auto INDEXER_COLUMN {"indexer"};
 constexpr auto USER_KEY {"username"};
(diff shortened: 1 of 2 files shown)


src/engine/source/indexerconnector/src/indexerConnector.cpp
```diff
@@ -17,6 +17,8 @@
 #include "secureCommunication.hpp"
 #include "serverSelector.hpp"
 #include <fstream>
+#include <iostream>
```
sebasfalcone, line 20:
```suggestion
```

src/engine/source/indexerconnector/src/indexerConnector.cpp
```diff
@@ -17,6 +17,7 @@
 #include "secureCommunication.hpp"
 #include "serverSelector.hpp"
 #include <fstream>
+#include <ostream>
```
sebasfalcone, line 20:
```suggestion
```

sebasfalcone (approved):

Dwordcito (approved):

Issue #26108, opened 2024-10-02
macOS syslog and ULS not configured out of box 


thecotilking:
# Description

During this test https://github.com/wazuh/wazuh/issues/26042, it was observed that we only have this configuration present out of the box for log analysis 

### Configuration present out of the box for log analysis 

```console

  <!-- Log analysis -->
  <localfile>
    <log_format>full_command</log_format>
    <command>netstat -an | awk '{if ((/^(tcp|udp)/) && ($4 != "*.*") && ($5 == "*.*")) {print $1" "$4" "$5}}' | sort -u</command>
    <alias>netstat listening ports</alias>
    <frequency>360</frequency>
  </localfile>
```

Only Agent restart is being captured by the Wazuh server as seen in the test https://github.com/wazuh/wazuh/issues/26042#issuecomment-2389423420 until an additional local file block is added before other logs are captured.

In the linked issue, the following test has been done: 

- Test that macOS log data collection works out of the box for syslog and ULS. Please, attach screenshots with alerts for these logs.

This test is needed to prove that the data collection works for `syslog` and for `ULS` out of the box (without adding anything), that is, using different log formats, check https://github.com/wazuh/wazuh/issues/26042#issuecomment-2389682586

But apparently, based on what @thecotilking observed there is no `localfile` with `macos` as `log_format`, and based on the blog post https://wazuh.com/blog/enhancing-macos-protection, there should be one.

```
The Wazuh macOS agents contain a default configuration to collect the appropriate logs.
...
The command uses filters to ensure that the log tool shows only the events of interest. Using the same technique, the Wazuh agent uses the following configuration included in macOS agents to collect only relevant logs:

<localfile>
   <location>macos</location>
   <log_format>macos</log_format>
   <query type="trace,log,activity" level="info">(process == "sudo") or (process == "sessionlogoutd" and message contains "logout is complete.") or (process == "sshd") or (process == "tccd" and message contains "Update Access Record") or (message contains "SessionAgentNotificationCenter") or (process == "screensharingd" and message contains "Authentication") or (process == "securityd" and eventMessage contains "Session" and subsystem == "com.apple.securityd")</query>
</localfile>
```

We need to check if this is correct and if ULS should work out of the box on macOS agents

vikman90:
## Issue reproduced

@thecotilking I've reproduced the issue, and I confirm this is a bug.

### Default configuration

This is a default ossec.conf on a macOS 15 @ ARM agent:

[...]

PR #26112, opened 2024-10-03
Add new check to verify the default configuration of new OS versions

MarcelKemp:
|Related issue|
|---|
|#26108|

## Description

With this PR, we modify the template used to generate the issues of the new OSs that appear, so that the agent test section will have the new check to verify the default configurations of the new OS. 
- This allows us to adapt the new OSs, so that they act similarly to the old versions.


jotacarma90 (approved):
LGTM

mjcr99 (approved):

vikman90 (approved):

PR #26141, opened 2024-10-04
Support updated versions of diff and apachectl at rootkit trojans file

vikman90:
|Related issue|
|---|
|Closes #26137 and #9337|

## Description

This PR aims to fix the following false positives:

### `diff` on Linux

```
Rule: 510 (level 7) -> 'Host-based anomaly detection event (rootcheck).'
Trojaned version of file '/bin/diff' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).
title: Trojaned version of file detected.
file: /bin/diff

Rule: 510 (level 7) -> 'Host-based anomaly detection event (rootcheck).'
Trojaned version of file '/usr/bin/diff' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).
title: Trojaned version of file detected.
file: /usr/bin/diff
```

#### Rationale

The `diff` tool now references `/dev/full` in addition to `/dev/null`.

### `apachectl` on macOS

```
Rule: 510 (level 7) -> 'Host-based anomaly detection event (rootcheck).'
Trojaned version of file '/usr/sbin/apachectl' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).
title: Trojaned version of file detected.
file: /usr/sbin/apachectl
```

#### Rationale

Now, _/usr/sbin/apachectl_ is implemented as a Bash script. Therefore, it includes the "bash" word in the shebang:
```bash
#!/bin/bash
```

## Proposed changes

- Update the `diff` entry to exclude both "/dev/null" and "/dev/full", by @djosip.
- Exclude the "bash" from the `apachectl` entry.

## Tests

- [x] The agent produces no false positives from Rootcheck on Ubuntu 24.04.
- [x] The agent produces no false positives from Rootcheck on macOS 15 Sequoia.

Issue #26137, opened 2024-10-04
Update rootcheck old signatures


gdiazlo:
# Description

Some rootcheck checks are outdated and generate false positives, we want to modify those checks so it won't produce them.

We believe the rootcheck usage on old systems is marginal now, so we can update these.

# Related issues

- https://github.com/wazuh/wazuh/issues/9337
- https://github.com/wazuh/wazuh/issues/19346
- https://github.com/wazuh/wazuh/issues/19000
- https://github.com/wazuh/wazuh/issues/11808

iasdeoupxe:
While #25634 fixed a false positive it has introduced additional false negatives, see the current discussion there.

iasdeoupxe:
Opened another PR: #26160

PR #26160, opened 2024-10-06
Fix newly introduced "diff" false negatives in rootkit_trojans.txt

iasdeoupxe:
|Related issue|
|---|
|#25634, #26137, #26141, #13278, #19000, #19346, #11808|

## Description

#25634 changed the related regex from:

```
/dev/[^n]
```

to:

```
/dev/[^nf]ull
```

with goal to exclude `/dev/null` and `/dev/full`. But actually introduced a lot of false negatives because it now only reports anything in `/dev` which:
1. Doesn't start with a `n`
2. Doesn't start with a `f`
3. Includes `ull`

The newly updated regex (as suggested in https://github.com/ossec/ossec-hids/issues/2020#issuecomment-1418331744) now only reports anything in `/dev` which:
1. Doesn't start with a `n`
2. Doesn't start with a `f`

Unfortunately it still causes false negatives for e.g. `/dev/fd0` or `/dev/fuse` but this is now a middle ground (due to the limitations of the search pattern functionality).

Another possibility would be to drop the `/dev` check completely as done in https://github.com/ossec/ossec-hids/pull/2062 but this is IMHO causing even more false negatives.

Yet another alternative is currently discussed in https://github.com/wazuh/wazuh/pull/26160#discussion_r1789004744

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests

### DIFF
--- a/ruleset/rootcheck/db/rootkit_trojans.txt
+++ b/ruleset/rootcheck/db/rootkit_trojans.txt
@@ -40,7 +40,7 @@ sudo        !satori|vejeta|conf\.inv!
 crond       !/dev/[^nt]|bash!
 gpm         !bash|mingetty!
 ifconfig    !bash|^/bin/sh|/dev/tux|session.null|/dev/[^cludisopt]!
-diff        !bash|^/bin/sh|file\.h|proc\.h|/dev/[^nf]ull|^/bin/.*sh!
+diff        !bash|^/bin/sh|file\.h|proc\.h|/dev/[^nf]|^/bin/.*sh!
 md5sum      !bash|^/bin/sh|file\.h|proc\.h|/dev/|^/bin/.*sh!
 hdparm      !bash|/dev/ida!
 ldd         !/dev/[^n]|proc\.h|libshow.so|libproc.a!


ruleset/rootcheck/db/rootkit_trojans.txt
```diff
@@ -40,7 +40,7 @@ sudo        !satori|vejeta|conf\.inv!
 crond       !/dev/[^nt]|bash!
 gpm         !bash|mingetty!
 ifconfig    !bash|^/bin/sh|/dev/tux|session.null|/dev/[^cludisopt]!
-diff        !bash|^/bin/sh|file\.h|proc\.h|/dev/[^nf]ull|^/bin/.*sh!
+diff        !bash|^/bin/sh|file\.h|proc\.h|/dev/[^nf]|^/bin/.*sh!
```
iasdeoupxe, line 43:
```suggestion
diff        !bash|^/bin/sh|file\.h|proc\.h|/dev/[^nf][^u]|^/bin/.*sh!
```

@brownowski as you left some notes in #25634 any opinion on this alternative?

It would still not catch `/dev/fuse` but would at least prevent an additional false negative against e.g. `/dev/fd` and `/dev/net` AFAICT.

brownowski: I believe something like the below would minimise corner cases.

/dev/[^nf]|/dev/[nf][^u]|/dev/[nf]u[^l]|/dev/[nf]ul[^l]

This catches:
All devices but devices starting with n or f
Then all devices starting with n or f, but not followed by u
Then all devices starting with n or f, then u, but not followed by l
Then all devices starting with n or f, then u, then l, but not followed by l

iasdeoupxe:
Any remarks/comments from Wazu team side?

vikman90:
Hello @iasdeoupxe,

Thank you for your contribution and for taking the time to improve the regex logic.  

Implementing this regex properly requires extensive testing to ensure it works correctly without unintended side effects. Unfortunately, the current list of trojans in Rootcheck is not actively maintained, and we lack the proper test infrastructure to validate the regex thoroughly at this time. For this reason, we cannot proceed with merging this PR into the `4.10.0` branch.  

Thank you for understanding.

Best regards,

iasdeoupxe:
@vikman90 This is bad and i really wonder how / why #25634 has been merged then if the Rootcheck is not actively maintained and there is a lack of proper test infrastructure.

Nevertheless i have created #27604 then, false negatives should be avoided at any costs (anyone who had the previous false positives could have created an overwriting rule, with the false negatives we can't do that).

iasdeoupxe:
For reference, solved now in #28982

Issue #25973, opened 2024-09-26
Review and fix Python security scan alerts


DamianNud:
## Description
New alerts were found in the last weekly security scan in the Static Code Analysis where `httpx` is called without a timeout. These changes were introduced in https://github.com/wazuh/wazuh/pull/25768. These alerts, which are identified as warnings, should be reviewed and fixed if appropriate, or marked as false positive leaving a comment that explains the reason.

## Edit
The changes were not introduced in the mentioned PR. Analyzing the different alerts, all the places where the scan detected a call to `httpx` without a timeout and their respective commit where found:

https://github.com/wazuh/wazuh/blob/cede9224dda880221d488e64b4201242152be991/framework/wazuh/core/manager.py#L338
As it can be seen, a timeout has been defined here, however, it was detected as an issue.
https://github.com/wazuh/wazuh/blob/cede9224dda880221d488e64b4201242152be991/framework/wazuh/core/engine/__init__.py#L28
https://github.com/wazuh/wazuh/blob/cede9224dda880221d488e64b4201242152be991/framework/wazuh/core/cluster/hap_helper/proxy.py#L167-L168
https://github.com/wazuh/wazuh/blob/cede9224dda880221d488e64b4201242152be991/framework/wazuh/core/cluster/hap_helper/proxy.py#L114

### Commits:
- https://github.com/wazuh/wazuh/commit/a5f011ab1f0a632855b61adfe304bd37807697fd
- https://github.com/wazuh/wazuh/commit/503ddc3380009abe1332e8a959b6a28bcd47d774
- https://github.com/wazuh/wazuh/commit/7962ce2e16f7f77c50b617fcb361b12a7593c83d

## Checks
The following elements have been updated or reviewed (should also be checked if no modification is required):
- [x] Review alerts.
- [x] Apply fixes and/or dismiss the alerts.

DamianNud:
## Update

Bandit checks for `httpx` calls without a timeout ([B113: request_without_timeout](https://bandit.readthedocs.io/en/latest/plugins/b113_request_without_timeout.html)) because, according to their documentation, failure to doing so can cause a program to hang indefinitely.

[...]

PR #26222, opened 2024-10-08
Add timeout to HTTPX calls

DamianNud:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/25973|

## Description

As commented [here](https://github.com/wazuh/wazuh/issues/25973#issuecomment-2397986246), it was decided to specify a timeout of 10 seconds to each `httpx` call, which sets 5 seconds by default, to be consistent across all calls and with the Management API.

No new definitions were added to the exception catching of each call because `httpx.TimeoutException` inherits from `httpx.RequestError`, which means that the specific timeout error will be caught with its parent class.

There were no changes in the behavior of the calls besides increasing the default timeout value.

### DIFF
--- a/framework/wazuh/core/manager.py
+++ b/framework/wazuh/core/manager.py
@@ -37,6 +37,7 @@ WAZUH_TAG_KEY = 'wazuh-tag'
 USER_AGENT_KEY = 'user-agent'
 DEFAULT_TIMEOUT = 10.0
 
+
 class LoggingFormat(Enum):
     plain = "plain"
     json = "json"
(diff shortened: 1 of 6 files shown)


framework/wazuh/core/manager.py
```diff
@@ -35,6 +35,7 @@
 WAZUH_UID_KEY = 'wazuh-uid'
 WAZUH_TAG_KEY = 'wazuh-tag'
 USER_AGENT_KEY = 'user-agent'
+DEFAULT_TIMEOUT = 10.0
```
GGP1, line 38: Add a new line to conform with the PEP8 style guide (classes should be separed by two empty lines)

```suggestion
DEFAULT_TIMEOUT = 10.0

```

GGP1 (changes requested):
We could update the unit tests to assert that the `httpx.AsyncClient` mocked instances are being called with the timeouts specified.

This way, we ensure that if they are removed in the future, the tests will fail and we will notice it.

DamianNud:
@GGP1 That is a good idea. I just added test cases checking that `httpx` is called with a timeout.

GGP1 (approved):
LGTM

fdalmaup (approved):
LGTM, although the ones related to the HAPHelper could have been omitted and the Bandit alerts dismissed since the component will be removed in 5.0.

PR #26287, opened 2024-10-11
Fix curl precompiled build

### LINKED ISSUE #26286: Error building wazuh[curl] on macOS x86
## Description
After the upgrade to DEPS 35, we receive this error during the build process in macOS x64.

```
configure: error: /Users/***/Documents/wazuh/src/external/openssl/ is a bad --with***penssl prefix!
make[2]: *** [config.status] Error 1
make[1]: *** [../config.status] Error 2
make: *** [external/curl/lib/.libs/libcurl.a] Error 2
```
After that we decide to rebuild the curl 8.10.1 and during the compilation on macOS Sierra we receive an other error:

```
checking for psl_builtin in -lpsl... no
configure: error: libpsl libs and/or directories were not found where specified!
make: *** [external/curl/Makefile] Error 1
```

## DoD
- [x] Fix libpsl error in the compilation machine.
- [x] Upload dependencies to s3

Dwordcito:
|Related issue|
|---|
|Closes #26286|

This pr aims to solve a compilation/configuration error during the build of curl.


MarcelKemp (approved):
LGTM!

Issue #26343, opened 2024-10-16
Define used Python version for GitHub Actions workflows 


fdalmaup:
## Description

It was found that some of the workflows executed to run code-related tests are failing due to running in `ubuntu-latest` virtual machines without defining the exact Python version to use, thus leaving using the system default one. Since `Ubuntu 24.04`, the Python version is `3.11` which requires to setup a virtual environment or specifying `--break-system-packages` when running `pip` as root. The other alternative is to define the Python version to `3.10` like [this case](https://github.com/wazuh/wazuh/blob/918f693384cb80e465ea5e024985a3a1d6ddac9f/.github/workflows/api-unit-tests.yml#L14-L17) or using `ubuntu-22.04` as the host machine. If non of this is carried out, the following error is found when executing the workflows:
```
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
Error: Process completed with exit code 1.
``` 

The found workflows currently failing due to this are:
- [API Integration tests Tier 0/1](https://github.com/wazuh/wazuh/blob/master/.github/workflows/integration-tests-api-tier-0-1.yml)
- [API Integration tests Tier 2](https://github.com/wazuh/wazuh/blob/master/.github/workflows/integration-tests-api-tier-2.yml)
- [AWS Integration tests](https://github.com/wazuh/wazuh/blob/master/.github/workflows/integration-tests-aws-tier-0-1.yml)
- [Integration tests for RBAC (Framework) - Tier 0 and 1](https://github.com/wazuh/wazuh/blob/master/.github/workflows/integration-tests-rbac-tier-0-1.yml)

## Checks

The following elements have been updated or reviewed (should also be checked if no modification is required):
[...]

Issue #23431, opened 2024-05-15
Add mechanism to check the regions parameter for AWS buckets


fdalmaup:
| Affected integration |
|---|
| AWS |

## Description
Since #22514, the AWS integration tests are generating their resources to test the module's behavior. Each launched test involving an S3 bucket creates the bucket, populates it with the corresponding logs, and at the end deletes the resources. For the `--regions` parameter, a test case involves requesting the logs for an unexistent region (`us-fake-1`) which in a normal module execution treats the parameter as a part of the log path and would return the following:
```
/var/ossec/wodles/aws/aws-s3 -b test-bucket -t config -s 2022-JAN-31 -p dev -d2 --regions us-fake-1
DEBUG: +++ Debug mode on - Level: 2
DEBUG: Generating default configuration for retries: mode standard - max_attempts 10
DEBUG: +++ Table does not exist; create
DEBUG: +++ Working on 123456789123 - us-fake-1
DEBUG: +++ Marker: AWSLogs/123456789123/Config/us-fake-1/2022/1/31
DEBUG: +++ No logs to process in bucket: 123456789123/us-fake-1
DEBUG: +++ DB Maintenance
```
Nevertheless, since the test builds up the bucket entirely and the region is not part of the available AWS regions, it fails and does not find the mentioned fake region.

The objective of the present issue is to modify the bucket's behavior to check the input `--regions` parameter and in case a nonexistent region is passed, to display a warning message that the region does not exist.

## Tasks
- [x] Test in a manager.
- [x] Test in an agent.
- [x] Unit tests without failures. Updated if there are any relevant changes.
- [x] Integration tests without failures. Updated if there are any relevant changes.
- [x] Update the documentation if necessary.
- [x] Add entry to the changelog if necessary.

javiersanchz:
## Update

The idea for this issue is to throw an error message as `cloudwatchlogs` does, for example, when the `-r `(`region`) is nonexistent, as in the following example (`-r us-fake-1`): 

```console
root@wazuh-master:/# /var/ossec/wodles/aws/aws-s3 -sr cloudwatchlogs -g 4_5_test -d 2 -s 2022-JUL-29 -p default -r us-fake-1
DEBUG: +++ Debug mode on - Level: 2
DEBUG: +++ ERROR: Invalid region 'us-fake-1'
```

javiersanchz:
## Update 

- This validation was also added for the `buckets`:

```console
[...]

PR #26346, opened 2024-10-16
Use Python 3.10 for workflow tests execution

fdalmaup:
|Related issue|
|---|
| #26343 |

## Description

Defines the Python version to run the team's ITs (`3.10`) to avoid using the default one from the OS that might break the execution.


nico-stefani (approved):
LGTM, the test failure is because of a [fix](https://github.com/wazuh/wazuh/issues/23431) included in 4.10.

RamosFe (approved):
## Update - Review

The errors found in the AWS ITs aren't related to the changes made in this PR and were solved in https://github.com/wazuh/wazuh/issues/23431.

> The changes made in https://github.com/wazuh/wazuh/issues/23431 will be reflected in `4.10.0`

Issue #26466, opened 2024-10-22
Python UT Coverage workflow doesn't exist in `4.10.0`


RamosFe:
|Wazuh version|
|---|
| 4.10.0 |

## Description

The following message appears when trying to run the GitHub Action for the Python UT Coverage.

![image](https://github.com/user-attachments/assets/f82201f9-ffd7-47e4-b1bd-8357f33859e2)

This happens because the workflow only exists in `master`. We should replicate the changes made at https://github.com/wazuh/wazuh/pull/23585/commits/5c8f59f7a3e987f09372b19b15142c2c0a4cb80d for the next RC of `4.10.0`

RamosFe:
## Update 

Added the changes made to https://github.com/wazuh/wazuh/issues/23558 to `4.10.0`. Also, it was necessary to change some dependencies of `qa-integration-framework` to solve some conflicts between the dependencies of `wazuh` and `qa-integration-framework` repositories.

Execution results: https://github.com/wazuh/wazuh/actions/runs/11462726706/job/31896441578?pr=26470

PR #26470, opened 2024-10-22
Add python unit tests coverage action to `4.10.0`

RamosFe:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/26466|

## Description

Adds a GitHub Action to get the Python unit tests coverage.

### DIFF
--- a/.github/workflows/api-unit-tests.yml
+++ b/.github/workflows/api-unit-tests.yml
@@ -18,10 +18,10 @@ jobs:
     env:
       PYTHONPATH: /home/runner/work/wazuh/wazuh/api:/home/runner/work/wazuh/wazuh/framework
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Setup Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
           cache: 'pip'
@@ -33,4 +33,4 @@ jobs:
           pip install -r framework/requirements-dev.txt --no-build-isolation
 
       - name: Run API tests
-        run: python -m pytest api/api
+        run: python -m pytest api/api --cov
--- a/.github/workflows/api-unit-tests.yml
+++ b/.github/workflows/api-unit-tests.yml
@@ -33,4 +33,4 @@ jobs:
           pip install -r framework/requirements-dev.txt --no-build-isolation
 
       - name: Run API tests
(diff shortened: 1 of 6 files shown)


.github/workflows/python-unit-tests-coverage.yml
```diff
@@ -0,0 +1,49 @@
...
+  pull_request:
+    paths:
+      - '.github/workflows/python-unit-tests-coverage.yml'
+jobs:
+  build:
+    env:
+      BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
+      BRANCH_BASE: ${{ github.base_ref || inputs.base_branch }}
+      PYTHONPATH: /home/runner/work/wazuh/wazuh/api:/home/runner/work/wazuh/wazuh/framework
+    runs-on: ubuntu-latest
```
GGP1, line 18: Use `ubuntu-22.04` to avoid a potential issue with `ubuntu-24.04` and the dependencies installation (in master it is ok to keep it in latest becasue the installation is not yet defined)

GGP1 (changes requested):

.github/workflows/python-unit-tests-coverage.yml
RamosFe: After the changes made in 5705cb63559f969b415102ea8eff605bb287e552 all test passed as expected. Test results: https://github.com/wazuh/wazuh/actions/runs/11480518503/job/31949103196?pr=26470

GGP1 (dismissed):
LGTM

.github/workflows/api-unit-tests.yml
fdalmaup, line 1: For some reason, the workflow is taking into account the code inside the `framework` folder for the coverage (checked in [this build](https://github.com/wazuh/wazuh/actions/runs/11480518471/job/31957192389)). We need to review that.

RamosFe: The `framework` folder was executed in the coverage due to the dependency of the `api` module with `framework`. After adding `--cov=api` in the Github Action `yml`, the coverage only executes for the `api` module, as shown in the following [build](https://github.com/wazuh/wazuh/actions/runs/11709870101/job/32614840744?pr=26470).

RamosFe: When `--cov` is used on UTs, the files in the UTs themselves are taken into account. Because this result is only an estimate and there is already a workflow in charge of coverage, we can conclude that the UT workflows work as expected.

fdalmaup (approved):
LGTM

PR #26484, opened 2024-10-23
Remove published_date field from YARA template

asteriscos:
|Related issue|
|---|
| https://github.com/wazuh/wazuh-dashboard/issues/377 |

## Description

This pull request removes the definition of the field `data.YARA.published_date` from the `wazuh-template.json` file because no visualizations use this field and it causes unnecessary conflicts with previous installations that already have a YARA integration, as it was interpreted as a keyword.


yenienserrano (approved):
🟢 LGTM

PR #26679, opened 2024-11-01
Improve the signal handling to avoid crashes during stop

### LINKED ISSUE #26536: wazuh-modulesd crashes at Vulnerability Scanner module
|Wazuh version|Component|Install type|Platform|
|---|---|---|---|
| ≤ 4.8.2 – 4.9.1 | Vulneraiblity Scanner | Manager | CentOS 7, Ubuntu 22.04 |

We've found that _wazuh-modulesd_ sometimes crashes with segmentation fault, in the following conditions:
- Stop the process a few seconds after starting.
- Default manager's configuration.

## Backtrace

```
#0  store (__m=std::memory_order_seq_cst, __i=true, this=0x48)
    at /usr/local/gcc-9.4.0/include/c++/9.4.0/bits/atomic_base.h:397
#1  std::__atomic_base<bool>::operator= (this=0x48, __i=true)
    at /usr/local/gcc-9.4.0/include/c++/9.4.0/bits/atomic_base.h:290
#2  0x00007fffe551db59 in std::atomic<bool>::operator= (this=0x48, __i=true)
    at /usr/local/gcc-9.4.0/include/c++/9.4.0/atomic:81
#3  0x00007fffe55572ff in SocketClient<Socket<OSPrimitives, SizeHeaderProtocol>, EpollWrapper>::stop (this=0x0)
    at /build_wazuh/rpmbuild/BUILD/wazuh-manager-4.8.2/src/shared_modules/utils/socketClient.hpp:92
#4  0x00007fffe553e74b in SocketDBWrapper::teardown (
    this=0x7fffe5990080 <Singleton<SocketDBWrapper>::instance()::s_instance>)
    at /build_wazuh/rpmbuild/BUILD/wazuh-manager-4.8.2/src/shared_modules/utils/socketDBWrapper.hpp:210
#5  0x00007fffe5517ea9 in VulnerabilityScannerFacade::stop (
    this=0x7fffe598fd20 <Singleton<VulnerabilityScannerFacade>::instance()::s_instance>)
    at /build_wazuh/rpmbuild/BUILD/wazuh-manager-4.8.2/src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScannerFacade.cpp:497
#6  0x00007fffe54c50d9 in VulnerabilityScanner::stop (
    this=0x7fffe598f6a8 <Singleton<VulnerabilityScanner>::instance()::s_instance>)
    at /build_wazuh/rpmbuild/BUILD/wazuh-manager-4.8.2/src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScanner.cpp:36
#7  0x00007fffe54c530c in vulnerability_scanner_stop ()
    at /build_wazuh/rpmbuild/BUILD/wazuh-manager-4.8.2/src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScanner.cpp:66
#8  0x0000000000484cfb in wm_vulnerability_scanner_stop (data=0x7ffff5a08970)
    at wazuh_modules/wm_vulnerability_scanner.c:148
#9  0x0000000000407f7e in wm_handler (signum=2) at wazuh_modules/main.c:206
#10 <signal handler called>
#11 0x00007ffff7de9e29 in _dl_fixup () from /lib64/ld-linux-x86-64.so.2
#12 0x00007ffff7df1923 in _dl_runtime_resolve_fxsave () from /lib64/ld-linux-x86-64.so.2
#13 0x00007fffe3c94b18 in rocksdb::ColumnFamilyData::ValidateOptions(rocksdb::DBOptions const&, rocksdb::ColumnFamilyOptions const&) ()
   from /var/ossec/bin/../lib/../lib/librocksdb.so.8
#14 0x00007fffe3d83c80 in rocksdb::DBImpl::ValidateOptions(rocksdb::DBOptions const&, std::vector<rocksdb::ColumnFamilyDescriptor, std::allocator<rocksdb::ColumnFamilyDescriptor> > const&) () from /var/ossec/bin/../lib/../lib/librocksdb.so.8
#15 0x00007fffe3d88a49 in rocksdb::DBImpl::Open(rocksdb::DBOptions const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<rocksdb::ColumnFamilyDescriptor, std::allocator<rocksdb::ColumnFamilyDescriptor> > const&, std::vector<rocksdb::ColumnFamilyHandle*, std::allocator<rocksdb::ColumnFamilyHandle*> >*, rocksdb::DB**, bool, bool) () from /var/ossec/bin/../lib/../lib/librocksdb.so.8
#16 0x00007fffe3d8af56 in rocksdb::DB::Open(rocksdb::DBOptions const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<rocksdb::ColumnFamilyDescriptor, std::allocator<rocksdb::ColumnFamilyDescriptor> > const&, std::vector<rocksdb::ColumnFamilyHandle*, std::allocator<rocksdb::ColumnFamilyHandle*> >*, rocksdb::DB**) () from /var/ossec/bin/../lib/../lib/librocksdb.so.8
#17 0x00007fffe557041d in Utils::TRocksDBWrapper<rocksdb::DB>::TRocksDBWrapper (
    this=0x7fffeda54180, dbPath=..., enableWal=true)
    at /build_wazuh/rpmbuild/BUILD/wazuh-manager-4.8.2/src/shared_modules/utils/rocksDBWrapper.hpp:118
#18 0x00007fffe555aaf8 in std::make_unique<Utils::TRocksDBWrapper<rocksdb::DB>, char const* const&> () at /usr/local/gcc-9.4.0/include/c++/9.4.0/bits/unique_ptr.h:857
#19 0x00007fffe551738e in VulnerabilityScannerFacade::start(std::function<void (int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, __va_list_tag*)> const&, nlohmann::json_abi_v3_11_2::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::json_abi_v3_11_2::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char> > > const&, bool, bool, bool) (
    this=0x7fffe598fd20 <Singleton<VulnerabilityScannerFacade>::instance()::s_instance>,
    logFunction=..., configuration=..., noWaitToStop=true, reloadGlobalMapsStartup=true,
    initContentUpdater=true)
    at /build_wazuh/rpmbuild/BUILD/wazuh-manager-4.8.2/src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScannerFacade.cpp:351
#20 0x00007fffe54c50bc in VulnerabilityScanner::start(std::function<void (int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::b---Type <return> to continue, or q <return> to quit---
asic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, __va_list_tag*)> const&, nlohmann::json_abi_v3_11_2::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::json_abi_v3_11_2::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char> > > const&, bool, bool, bool) (
    this=0x7fffe598f6a8 <Singleton<VulnerabilityScanner>::instance()::s_instance>,
    logFunction=..., configuration=..., noWaitToStop=true, reloadGlobalMapsStartup=true,
    initContentUpdater=true)
    at /build_wazuh/rpmbuild/BUILD/wazuh-manager-4.8.2/src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScanner.cpp:30
#21 0x00007fffe54c528c in vulnerability_scanner_start (
    callbackLog=0x425259 <mtLoggingFunctionsWrapper>, configuration=0x7fffeda78c00)
    at /build_wazuh/rpmbuild/BUILD/wazuh-manager-4.8.2/src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScanner.cpp:52
#22 0x0000000000484c2d in wm_vulnerability_scanner_main (data=0x7ffff5a08970)
    at wazuh_modules/wm_vulnerability_scanner.c:125
#23 0x00007ffff6292ea5 in start_thread () from /lib64/libpthread.so.0
#24 0x00007ffff5fbbb0d in clone () from /lib64/libc.so.6
```

## Core dump

- [wazuh-modulesd-4.9.1-segfault-vd.core.tar.gz](https://github.com/user-attachments/files/17517164/wazuh-modulesd-4.9.1-segfault-vd.core.tar.gz)

pereyra-m:
|Related issue|
|---|
|Closes #26536|

## Description

To avoid the crash of the `wazuh-modulesd` during the shutdown, some improvements were added:
- The signals are ignored at startup and until the modules have started. This avoids the termination of the module during the `sleep()` calls.
- There is a wait time of 1 second after the modules have started to allow them settle.
- The signal handling setup was moved after the added wait time to avoid calling a stop before the start has finished.
- The signal handling now has a mutex, it will only process one signal and will ignore the rest

**Update**

The original proposal described above might have many changes. A simplified one is under analysis, where the sleep occurs at the stop method and no sig-ignore is configured at startup. 

## Tests

The reproduction steps now don't cause crashes.

  - [x] Linux

### DIFF
--- a/src/wazuh_modules/main.c
+++ b/src/wazuh_modules/main.c
@@ -16,13 +16,21 @@ static void wm_help();                  // Print help.
 static void wm_setup();                 // Setup function. Exits on error.
 static void wm_cleanup();               // Cleanup function, called on exiting.
 static void wm_handler(int signum);     // Action on signal.
+static void wm_signals_ignore();        // Ignore signals until the modules have started.
+static void wm_signals_configure();     // Configure signal handling.
 
 static int flag_foreground = 0;         // Running in foreground.
+
+static bool wm_sig_received = false;   // Flag to indicate if a signal is being handled.
+static pthread_mutex_t sig_lock = PTHREAD_MUTEX_INITIALIZER;
 
 // Main function
 
 int main(int argc, char **argv)
 {
+    // Ignore signals until the modules have started
+    wm_signals_ignore();
+
     int c;
     int wm_debug = 0;
     int test_config = 0;
@@ -98,8 +106,13 @@ int main(int argc, char **argv)
     // Start com request thread
     w_create_thread(wmcom_main, NULL);
 
-    // Wait for threads
+    mdebug2("Waiting for modules to start.");
+    sleep(1);
 
+    // Signal management
+    wm_signals_configure();
+
+    // Wait for threads
     for (cur_module = wmodules; cur_module; cur_module = cur_module->next) {
         pthread_join(cur_module->thread, NULL);
     }
@@ -127,8 +140,6 @@ void wm_help()
 
 void wm_setup()
 {
-    struct sigaction action = { .sa_handler = wm_handler };
-
     // Read XML settings and internal options
 
     if (wm_config() < 0) {
@@ -153,19 +164,6 @@ void wm_setup()
         minfo("No configuration defined. Exiting...");
         exit(EXIT_SUCCESS);
     }
-
-    // Signal management
-
-    atexit(wm_cleanup);
-    sigaction(SIGTERM, &action, NULL);
-
-    if (flag_foreground) {
-        sigaction(SIGHUP, &action, NULL);
-        sigaction(SIGINT, &action, NULL);
-    }
-
-    action.sa_handler = SIG_IGN;
-    sigaction(SIGPIPE, &action, NULL);
 
     // Create PID file
 
@@ -192,8 +190,40 @@ void wm_cleanup()
 
 // Action on signal
 
+void wm_signals_ignore()
+{
+    struct sigaction action = { .sa_handler = SIG_IGN };
+
+    sigaction(SIGTERM, &action, NULL);
+    sigaction(SIGPIPE, &action, NULL);
+}
+
+void wm_signals_configure()
+{
+    struct sigaction action = { .sa_handler = wm_handler };
+
+    atexit(wm_cleanup);
+    sigaction(SIGTERM, &action, NULL);
+
+    if (flag_foreground) {
+        sigaction(SIGHUP, &action, NULL);
+        sigaction(SIGINT, &action, NULL);
+    }
+}
+
 void wm_handler(int signum)
 {
+    w_mutex_lock(&sig_lock);
+    if (wm_sig_received) {
+        mdebug2("Signal received: %d, ignoring.", signum);
+        w_mutex_unlock(&sig_lock);
+        return;
+    } else {
+        wm_sig_received = true;
+        mdebug2("Signal received: %d, handling.", signum);
+    }
+    w_mutex_unlock(&sig_lock);
+
     wmodule * cur_module;
     switch (signum) {
     case SIGHUP:
@@ -201,6 +231,7 @@ void wm_handler(int signum)
     case SIGTERM:
         // For the moment only gracefull shutdown will be for syscollector, in the future
         // it will be modified for all wmodules, modifying the mainloop of each thread.
+        minfo("Shutting down Wazuh modules.");
         for (cur_module = wmodules; cur_module && cur_module->context && cur_module->context->name; cur_module = cur_module->next) {
             if (cur_module->context->stop) {
                 cur_module->context->stop(cur_module->data);
--- a/src/wazuh_modules/main.c
+++ b/src/wazuh_modules/main.c
@@ -16,21 +16,16 @@ static void wm_help();                  // Print help.
 static void wm_setup();                 // Setup function. Exits on error.
 static void wm_cleanup();               // Cleanup function, called on exiting.
 static void wm_handler(int signum);     // Action on signal.
-static void wm_signals_ignore();        // Ignore signals until the modules have started.
 static void wm_signals_configure();     // Configure signal handling.
 
 static int flag_foreground = 0;         // Running in foreground.
 
-static bool wm_sig_received = false;   // Flag to indicate if a signal is being handled.
 static pthread_mutex_t sig_lock = PTHREAD_MUTEX_INITIALIZER;
 
 // Main function
 
 int main(int argc, char **argv)
 {
-    // Ignore signals until the modules have started
-    wm_signals_ignore();
-
     int c;
     int wm_debug = 0;
     int test_config = 0;
@@ -106,13 +101,11 @@ int main(int argc, char **argv)
     // Start com request thread
     w_create_thread(wmcom_main, NULL);
 
-    mdebug2("Waiting for modules to start.");
-    sleep(1);
-
     // Signal management
     wm_signals_configure();
 
     // Wait for threads
+
     for (cur_module = wmodules; cur_module; cur_module = cur_module->next) {
         pthread_join(cur_module->thread, NULL);
     }
@@ -190,14 +183,6 @@ void wm_cleanup()
 
 // Action on signal
 
-void wm_signals_ignore()
-{
-    struct sigaction action = { .sa_handler = SIG_IGN };
-
-    sigaction(SIGTERM, &action, NULL);
-    sigaction(SIGPIPE, &action, NULL);
-}
-
 void wm_signals_configure()
 {
     struct sigaction action = { .sa_handler = wm_handler };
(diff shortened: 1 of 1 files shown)


src/wazuh_modules/main.c
```diff
@@ -16,13 +16,21 @@ static void wm_help();                  // Print help.
...
 
+static bool wm_sig_received = false;   // Flag to indicate if a signal is being handled.
+static pthread_mutex_t sig_lock = PTHREAD_MUTEX_INITIALIZER;
+
 // Main function
 
 int main(int argc, char **argv)
 {
+    // Ignore signals until the modules have started
+    wm_signals_ignore();
```
sebasfalcone, line 32: Note: Performed some testing over this

If the module initialization takes too long or it hungs we are still able to kill the process, I don't know if this desirable or not but its working as intended

sebasfalcone (commented):
Amazing work @pereyra-m

vikman90 (commented):
The proposed solution is on the right track! However, after reviewing the trace, I would suggest considering some alternative approaches.

One concern from the trace is that the `wm_handler` exit routine is executed by the Vuln Detector thread, while ideally, it should be handled by the main thread.

Additionally, the use of `sleep(1)` raises some concerns. This approach could lead to inconsistent behavior across different systems—on some, it might be too long, while on others, it could be insufficient. It appears that a sleep function is being used to mitigate a race condition, which is not an ideal solution.

Given that the modules seem to have a flag for handling graceful shutdowns, I would recommend exploring the replacement of `wm_signals_ignore` with a per-thread signal mask, ensuring proper signal handling across different threads.

Cheers!

src/wazuh_modules/main.c
```diff
@@ -16,13 +16,21 @@ static void wm_help();                  // Print help.
 static void wm_setup();                 // Setup function. Exits on error.
 static void wm_cleanup();               // Cleanup function, called on exiting.
 static void wm_handler(int signum);     // Action on signal.
+static void wm_signals_ignore();        // Ignore signals until the modules have started.
+static void wm_signals_configure();     // Configure signal handling.
 
 static int flag_foreground = 0;         // Running in foreground.
 
+static bool wm_sig_received = false;   // Flag to indicate if a signal is being handled.
```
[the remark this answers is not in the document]
pereyra-m: You are right.

src/wazuh_modules/main.c
pereyra-m: I'm open to comments here. Ignoring the signals also avoids the `Termination` of the process when it receives a signal during the starting phase, but it might change the current behavior too much.

The real fix is in the mutex, I can revert this.

pereyra-m:
Hi @vikman90 

Thank you for the review.

> One concern from the trace is that the wm_handler exit routine is executed by the Vuln Detector thread, while ideally, it should be handled by the main thread.

I don't understand why this happens. Can you share the backtrace? I think that setting the signal handling in the main thread shouldn't have this behavior. 

> Additionally, the use of sleep(1) raises some concerns. This approach could lead to inconsistent behavior across different systems—on some, it might be too long, while on others, it could be insufficient. It appears that a sleep function is being used to mitigate a race condition, which is not an ideal solution.

I don't like it either, but I thought that if we had some sleeps at the start of other modules, it wouldn't be that bad. I agree that a real coordination of all the modules is much better, because we could know the exact moment when all of them are ready. But this issue isn't reproduced naturally in the product, so I preferred to make fewer changes and not to have a major rework. 
I proposed it here to allow all the modules to settle, but it isn't the most important part of the fix and I can remove it.

> I would recommend exploring the replacement of wm_signals_ignore with a per-thread signal mask, ensuring proper signal handling across different threads.

I proposed the `wm_signals_ignore()` method to avoid a premature shutdown of the module. I meant to ignore all signals until the daemon is ready to process them. I can remove it also, because in my tests, the most important part of the fix is the mutex at `wm_handler()`

pereyra-m:
@sebasfalcone @vikman90 

I've uploaded another proposal for your consideration.

Given that making each module start routine notify the main thread that is ready is a big change, I moved the sleep to the stop section. I try to cause the less impact possible here, but without this, the SIGTERM signal happens too fast and the module crashes. Now I don't affect the time it takes the module to start.

Also, I've removed the signal ignore method at startup to have fewer changes in the PR. Consider that sometimes, the process is terminated because the signal is received during a pre-existent sleep and it isn't ignored anymore; but this isn't a crash.

src/wazuh_modules/main.c
```diff
@@ -192,15 +183,48 @@ void wm_cleanup()
...
+        w_mutex_unlock(&sig_lock);
+        return;
+    } else {
+        wm_sig_processing = true;
+        mdebug2("Signal received: %d, handling.", signum);
+    }
+    w_mutex_unlock(&sig_lock);
+
+    // We wait in case the modules haven't completed the initialization
+    sleep(1);
```
sebasfalcone, line 218: We know this is not the best approach, but is the best we can do without inquiring about a big refactor regarding the initialization and stopping process of the modules

The stop-and-start interfaces we currently have are not prepared for this kind of behaviour

sebasfalcone (approved):
LGTM

The segfault prevented with this approach, in a normal situation it shouldn't happen anymore

Issue #26678, opened 2024-11-01
Engine - fix uptime metadata in Router API Responses


juliancnn:
### Description:

There is a bug in the router API of Wazuh-Engine where the `uptime` for disabled routes is incorrectly returned. Instead of showing the expected `uptime: 0` for disabled routes, the API currently returns the actual uptime since the route was last created. This issue may cause confusion for users monitoring the status of routes through the API.

### Objectives:

- Correct the uptime display in the router API to ensure that it accurately reflects the status of routes, particularly when they are disabled.

### Tasks:
1. **Identify the Cause**: Review the router API code to understand why the uptime is being incorrectly calculated or displayed for disabled routes.
2. **Implement the Fix**: Modify the API to return `uptime: 0` for all disabled routes, aligning the output with expected behavior.
3. **Update Unit Tests**: Ensure that existing unit tests reflect this change and add new tests if necessary to cover this scenario.

### Expected Outcome:

An updated router API that accurately returns `uptime: 0` for disabled routes, enhancing the reliability and usability of API responses.

### DoD (Definition of Done):
- [x] The router API correctly return `uptime: 0` for disabled routes.
- [x] Unit tests confirm the correct behavior and are passing.

This fix will improve the accuracy of the router API's responses, ensuring that users receive correct data regarding the status of routes in Wazuh-Engine, which is crucial for effective monitoring and management of the system.

NahuFigueroa97:
## Update

- Identify the Cause: Review the router API code to understand why the uptime is being incorrectly calculated or displayed for disabled routes.
- Implement the Fix: Modify the API to return uptime: 0 for all disabled routes, aligning the output with expected behavior.
- Create integration test

PR #26682, opened 2024-11-01
Fix uptime for route with disable state

NahuFigueroa97:
|Related issue|
|---|
|#26678|

### Description:

There is a bug in the router API of Wazuh-Engine where the `uptime` for disabled routes is incorrectly returned. Instead of showing the expected `uptime: 0` for disabled routes, the API currently returns the actual uptime since the route was last created. This issue may cause confusion for users monitoring the status of routes through the API.


juliancnn (approved):
LGTM!

JcabreraC (approved):
LGTM !

Issue #26582, opened 2024-10-28
Wazuh Manager crashes when modifying configuration, and a maild process fails to terminate


n0tl3an:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.9.1| Wazuh Manager | Manager | Sources | Oracle Linux 9 |

To enhance the security of our company's various IT environments, we have deployed and configured Wazuh. This instance is running on a standalone server, without any clusters, in a single-node configuration for all components.

### Issue Description

Until recently, Wazuh has been running smoothly. However, I noticed that it was not sending emails for alerts, as it was initially configured to do. Checking the logs on the web interface, I confirmed connection errors with the mail relay. While attempting to edit the manager configuration to address this issue, I encountered a more critical problem: each time I save any configuration changes to the manager, it crashes and returns an API access error.

### Observations and Symptoms

When attempting a manual restart of the manager via CLI, I receive an error indicating that the maild PID failed to stop. At this point, I observe that Wazuh Manager’s memory usage spikes dramatically, consuming up to 20GB of RAM, which leads to a crash from either an OOM (Out of Memory) error on my system or a timeout from Wazuh itself.

I’m at a loss as to the cause of this issue. I’ve been using Wazuh for several years and have never encountered anything like this.
Logs from the Wazuh Manager GUI during the crash

```ruby
AxiosError@https://My_Manager_IP/49104/bundles/plugin/wazuh/wazuh.plugin.js:1:70632
settle@https://My_Manager_IP/49104/bundles/plugin/wazuh/wazuh.plugin.js:15:27877
onloadend@https://My_Manager_IP/49104/bundles/plugin/wazuh/wazuh.plugin.js:15:37305
```

### Logs from Service Status Check

```bash
Oct 28 15:35:02 My_FQDN systemd[1]: Starting Wazuh manager...
Oct 28 15:35:03 My_FQDN env[9947]: 2024/10/28 15:35:03 wazuh-db[9947] debug_op.c:116 at _log_function(): DEBUG: Logging module auto-initialized
Oct 28 15:35:03 My_FQDN env[9947]: 2024/10/28 15:35:03 wazuh-db[9947] main.c:125 at main(): DEBUG: Wazuh home directory: /var/ossec
Oct 28 15:35:40 My_FQDN systemd[1]: wazuh-manager.service: A process of this unit has been killed by the OOM killer.
[...]

PR #26704, opened 2024-11-04
Fix overflow when reading email alerts without email_to attribute

NahuFigueroa97:
|Related issue|
|---|
|#26582|

## Description

Fixes overflow when reading email alerts without email_to attribute. The variables were defined as unsigned integers and a subtraction was performed to the value zero, not supported by that type of data when no recipient for the email was indicated in the configuration. In addition, the problem of redefining email_to if it was previously defined in the global configuration is solved.


juliancnn (approved):
LGTM!

JcabreraC (approved):
LGTM !

PR #26716, opened 2024-11-05
Allow unknown syslog identifier at Logcollector's journald reader

### LINKED ISSUE #25534: The logcollector module reports journal error in AL2 (4.9.0)
## Description

While working on https://github.com/wazuh/wazuh-installation-assistant/issues/20, we observed that an error was reported in an AL2 machine after performing an AIO installation:
```console
[root@ip-172-31-81-120 ec2-user]# grep -i -E "error" /var/ossec/logs/ossec.log 
2024/09/02 09:12:44 wazuh-logcollector: ERROR: (1611): Failed to get the message from the journal
```

We need to investigate this error and try to fix it.

## Steps to reproduce
1. Provision an AL2 machine
2. Download and perform an AIO installation with the `curl -sO https://packages-dev.wazuh.com/4.9/wazuh-install.sh`
3. See the `ossec.logs` error.

> [!NOTE]
> This behavior was not observed in other OSs.

vikman90:
|Related issue|
|---|
|Closes #25534|

## Description

This PR aims to fix an issue in Logcollector that sometimes made it produce this log:
> 2024/11/05 11:19:01 wazuh-logcollector: ERROR: (1611): Failed to get the message from the journal

**This happens when logs have no identifier (tag).**

### Rationale

The Logcollector's journald format reader gathers logs from _journald_ and tries to format them into the _syslog_ format, like this:

```
Nov 05 11:46:14 localhost run-parts(/etc/cron.daily)[30856]: starting mlocate
--------------- --------- -------------------------- -----   ----------------
Timestamp       Hostname  Identifier (tag)           PID     Message
```

For that, the agent needs to catch the following data:

- **Timestamp**:
   This field belongs to the log's metadata.
- **`_HOSTNAME`**:
   This is a **trusted field**, meaning that it's implicitly added by the journal.
- **`SYSLOG_IDENTIFIER`**:
   This is the log's tag, usually the program name.
- **`SYSLOG_PID` or `_PID`**:
   This is the PID of the log source.
- **`MESSAGE`**:
   Body of the log.

**Reference:** [systemd.journal-fields(3)](https://man7.org/linux/man-pages/man7/systemd.journal-fields.7.html)

In other words, we can expect the timestamp and the hostname to be included in every log. On the other hand:
- `SYSLOG_IDENTIFIER` may be missing. In this case, `journalctl` inserts "unknown" in its place.
- If `SYSLOG_PID` is missing, Logcollector gets `_PID` instead (which should always exist).
- If `MESSAGE` is missing, the log has no sense.

## Change proposal

We propose a behavior similar to what we have noticed in the `journalctl` tool:

1. If the identifier (or tag) is missing, silently replace it with `"unknown"` and send the log.
2. If the message is missing, print a debug log and discard the log.

## Tests

- [X] Push a log with a missing tag → Logcollector inserts `"unknown"` in its place.
- [X] Push a log with a missing message → Logcollector skips it.
- [X] No error logs in any case.
- [X] Unit tests.

### Testing tool

<details><summary>logger.py</summary>

```python
#! /usr/bin/env python3

import socket
from sys import argv

JOURNAL_SOCKET = '/run/systemd/journal/socket'

def format_log(message: str, tag: str):
    if tag is None:
        return f'MESSAGE={message}\n'
    else:
        return f"SYSLOG_IDENTIFIER={tag}\nMESSAGE={message}\n"

def log_to_journald(message: str, tag: str = None):
    log_message = format_log(message, tag)

    with socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM) as sock:
        sock.connect(JOURNAL_SOCKET)
        sock.sendall(log_message.encode('utf-8'))

if __name__ == '__main__':
    if len(argv) == 1:
        print('Syntax: ./logger.py [TAG] <MESSAGE>')
    elif len(argv) == 2:
        log_to_journald(argv[1])
    else:
        log_to_journald(argv[2], argv[1])
```

</details>

### Results (archives)

- Push a log with a tag:
```bash
python3 logger.py 'Test' 'Hello World'
```
> 2024 Nov 05 13:10:27 (amazon2) any->journald Nov 05 12:10:25 localhost Test[31016]: Hello World
- Push a log without a tag:
```bash
python3 logger.py 'Hello World'
```
> 2024 Nov 05 13:09:50 (amazon2) any->journald Nov 05 12:09:49 localhost unknown[31009]: Hello World

### DIFF
--- a/src/logcollector/journal_log.c
+++ b/src/logcollector/journal_log.c
@@ -148,11 +148,11 @@ STATIC INLINE char * find_library_path(const char * library_name) {
     while (getline(&line, &len, maps_file) != -1) {
         if (strstr(line, library_name) != NULL) {
             char * path_start = strchr(line, '/');
-            if (path_start == NULL) {                
+            if (path_start == NULL) {
                 break; // Never happens
             }
             char * path_end = strchr(path_start, '\n');
-            if (path_end == NULL) {                
+            if (path_end == NULL) {
                 break; // Never happens
             }
             *path_end = '\0';
@@ -310,8 +310,8 @@ void w_journal_context_update_timestamp(w_journal_context_t * ctx) {
     }
 }
 
-int w_journal_context_seek_most_recent(w_journal_context_t * ctx) {  
-    
+int w_journal_context_seek_most_recent(w_journal_context_t * ctx) {
+
     if (ctx == NULL) {
         return -1;
     }
@@ -330,7 +330,7 @@ int w_journal_context_seek_most_recent(w_journal_context_t * ctx) {
 }
 
 int w_journal_context_seek_timestamp(w_journal_context_t * ctx, uint64_t timestamp) {
-    
+
     if (ctx == NULL) {
         return -1;
     }
@@ -366,7 +366,7 @@ int w_journal_context_seek_timestamp(w_journal_context_t * ctx, uint64_t timesta
 }
 
 int w_journal_context_next_newest(w_journal_context_t * ctx) {
-    
+
     if (ctx == NULL) {
         return -1;
     }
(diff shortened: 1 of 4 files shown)


JcabreraC (approved):
LGTM !

PR #26769, opened 2024-11-08
Fix PID parsing method in Wazuh-DB processes

pereyra-m:
## Description

This PR aims to fix a parsing issue in Wazuh-DB processes method.

### DIFF
--- a/src/wazuh_db/wdb_syscollector.c
+++ b/src/wazuh_db/wdb_syscollector.c
@@ -1101,7 +1101,7 @@ int wdb_syscollector_processes_save2(wdb_t * wdb, const cJSON * attributes)
 {
     const char * scan_id = "0";
     const char * scan_time = cJSON_GetStringValue(cJSON_GetObjectItem(attributes, "scan_time"));
-    const int pid = strtol(cJSON_GetStringValue(cJSON_GetObjectItem(attributes, "pid")),NULL,10);
+    const int pid = cJSON_GetStringValue(cJSON_GetObjectItem(attributes, "pid")) ? strtol(cJSON_GetStringValue(cJSON_GetObjectItem(attributes, "pid")),NULL,10) : 0;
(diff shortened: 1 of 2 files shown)


sebasfalcone (changes requested):
Good Catch!

Lets add some UTs and we are ready to go

src/wazuh_db/wdb_syscollector.c
```diff
@@ -1101,7 +1101,7 @@ int wdb_syscollector_processes_save2(wdb_t * wdb, const cJSON * attributes)
 {
     const char * scan_id = "0";
     const char * scan_time = cJSON_GetStringValue(cJSON_GetObjectItem(attributes, "scan_time"));
-    const int pid = strtol(cJSON_GetStringValue(cJSON_GetObjectItem(attributes, "pid")),NULL,10);
+    const int pid = cJSON_GetStringValue(cJSON_GetObjectItem(attributes, "pid")) ? strtol(cJSON_GetStringValue(cJSON_GetObjectItem(attributes, "pid")),NULL,10) : 0;
```
sebasfalcone, line 1104: Question: Is it right to set the PID to 0?

pereyra-m: Hi @sebasfalcone , interesting question.

If the value isn't present at the JSON, we could set here `-1`, this will reject the insertion

https://github.com/wazuh/wazuh/blob/f41b71a7212612a9d74f34d56c0f96987127500b/src/wazuh_db/wdb_syscollector.c#L972-L973

This is because PID is a primary key
https://github.com/wazuh/wazuh/blob/f41b71a7212612a9d74f34d56c0f96987127500b/src/wazuh_db/schema_agents.sql#L232

If we let this as 0, that invalid package would still be inserted. But considering the primary key restriction, only one package like this could be inserted.

sebasfalcone (approved):
LGTM!

If the PID cannot be obtained, the operation fails

Dwordcito (approved):

PR #26832, opened 2024-11-12
Define mandatory fields in flatbuffer schema to discard invalid messages

pereyra-m:
## Description

This PR adds defines some `required` fields in the flarbuffer schemas to discard invalid message.
With this addition, they are discarded during the parsing phase and aren't forwarded to the vulnerability detector module.
The related UTs were added.


sebasfalcone (dismissed):
LGTM!

The invalid messages will be caught here:
https://github.com/wazuh/wazuh/blob/0f3ac1bba4cc7fbedb9565a1e502080aaad945e8/src/shared_modules/router/src/router.cpp#L261-L265

Dwordcito (approved):

Issue #26869, opened 2024-11-15
Wazuh server configuration refinement


fdalmaup:
# Description

After the complete refactoring of the Wazuh server configuration in #26356, we performed some tests that showed some adjustments that needed to be developed:

- When installing the `wazuh-server.deb` package, the `wazuh-server.yml` with the minimum default configurations is not generated:
    ```
    root@7e8591388f09:/# wazuh-server start -rf
    2024/11/15 08:15:26 ERROR: [Cluster] [Main] Configuration file not found: /etc/wazuh-server/wazuh-server.yml
    ```
     This is not the expected behavior.
    - [ ] It needs to be manually included for the server to work.

- Reviewing the location where the `wazuh-server.yml` needs to be found, the following structure was found:
    ```
    root@57b64eb801c4:/# ll /etc/wazuh-server/
    total 60
    drwxr-x--- 1 root  999 4096 Nov 15 08:44 ./
    drwxr-xr-x 1 root root 4096 Nov 15 08:44 ../
    drwxr-x--- 1 root  999 4096 Nov 14 15:46 api/
    drwxr-x--- 1 root  999 4096 Nov 15 08:44 cluster/
    -rw-r----- 1 root  999 7138 Nov 13 17:18 ossec.conf
    -rw-r--r-- 1 root root 1919 Nov 15 08:44 server.ca
    -rw------- 1 root root 1424 Nov 15 08:44 server.crt
    -rw-r--r-- 1 root root  920 Nov 15 08:44 server.csr
    -rw------- 1 root root 1700 Nov 15 08:44 server.key
    -rw-r--r-- 1 root root   41 Nov 15 08:44 server.srl
    drwxr-x--- 2 root  999 4096 Nov 13 17:18 shared/
    ```
    Also, inside the `api` and `cluster` directories, the `configuration/api.yaml` and `cluster.json` files are found. 
   - [ ] Any reference to `ossec` and unsupported configuration files must be removed.

RamosFe:
## Update

When starting a new environment using Docker, after the changes to the code made in the PR, the following results were obtained:

- Path `/etc/wazuh-server/`:
```console
root@wazuh-master:/#  ll /etc/wazuh-server/
total 44
drwxr-x--- 1 root  wazuh 4096 Nov 15 12:22 ./
drwxr-xr-x 1 root  root  4096 Nov 15 11:43 ../
drwxr-x--- 1 root  wazuh 4096 Nov 15 11:42 api/
[...]

PR #26873, opened 2024-11-15
Server configuration refinement

RamosFe:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/26869|

## Description

- It adds a minimum default configuration when installing Wazuh.
- Deletes references to `cluster.json` and `api.yml` caused by merge conflicts.


api/Makefile
```diff
@@ -27,7 +27,6 @@ install:
     # Copy files and create folders
 	# ETC
 	$(INSTALL_DIR) $(ETC_INSTALLDIR)/api
-	$(INSTALL_RW_DIR) $(ETC_INSTALLDIR)/api/configuration
```
GGP1, line 30: Restore this line

api/scripts/wazuh_apid.py
```diff
@@ -72,10 +72,10 @@ def configure_ssl(params: dict, config: APISSLConfig):
                         'Attempting to generate them')
             private_key = generate_private_key(config.key)
             logger.info(
-                f"Generated private key file in WAZUH_PATH/{to_relative_path(config.key)}")
+                f"Generated private key file in {config.key}")
```
GGP1, line 75: If the user sets a relative path in the `config.key` the message will be misleading. I would prefer to get the absolute path and show that instead.

If this doesn't involve a single line change then I would leave it as a technical debt though.

etc/wazuh-server.yml
```diff
@@ -1,112 +1,15 @@
...
-#    key: "/etc/wazuh-server/api/configuration/ssl/server.key"
-#    cert: "/etc/wazuh-server/api/configuration/ssl/server.crt"
-#    use_ca: false
-#    ca: ""
-#    ssl_protocol: "auto"
-#    ssl_ciphers: ""
+  host: wazuh-indexer
+  port: 9200
+  user: admin
+  password: SecretPassword1%
```
GGP1, line 15: These values should be empty. The username and password will depend on the user environment

GGP1 (changes requested):

api/Makefile
RamosFe: Fixed in cf533d856c1db1e32d99244bb707857b05e4f521.

etc/wazuh-server.yml
RamosFe: Fixed in cf533d856c1db1e32d99244bb707857b05e4f521.

framework/wazuh/core/exception.py
```diff
@@ -26,7 +26,7 @@ class WazuhException(Exception):
         # Wazuh: 0999 - 1099
         999: 'Incompatible version of Python',
         1000: {'message': 'Wazuh Internal Error',
-               'remediation': 'Please, check `WAZUH_HOME/logs/ossec.log`, `WAZUH_HOME/logs/cluster.log` and '
+               'remediation': 'Please, check `WAZUH_HOME/logs/cluster.log` and '
```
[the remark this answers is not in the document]
RamosFe: Fixed in cf533d856c1db1e32d99244bb707857b05e4f521.

api/scripts/wazuh_apid.py
RamosFe: Because it is a passed string and not a structure like the ones we are using from pathlib, I think the change would be more than 1 line of code. I add to the technical debt to review the paths in the error messages as well as make this format consistent.

GGP1 (approved):
LGTM

fdalmaup (approved):
LGTM!

Issue #26867, opened 2024-11-14
Validate engine logging function calls


pereyra-m:
## Description

It was found that some calls to the logging function don't work because they use an invalid structure.
For example

https://github.com/wazuh/wazuh/blob/2be90523bbfe89dbd8b2c443eec5cc5832a39f21/src/engine/source/api/include/api/api.hpp#L79

The `%s` should be `{}`.

It happened in other modules, and it was corrected 
https://github.com/wazuh/wazuh/pull/26216/commits/802327bbacdbf75530180b5d7eb33811cae25730

Also, it seems that the logging isn't unified in across the files. Some functions still use the `wazuh/src/shared_modules/utils/loggerHelper.h` class.

## DoD

- [x] Find and fix all places where the logging isn't working
- [ ] Consider creating another issue for the logging unification

GabrielEValenzuela:
# Update

- Task completed
- Awaited review

PR #26920, opened 2024-11-19
Fix logging arguments

GabrielEValenzuela:
|Related issue|
|---|
|#26867|

# Objective

This PR aims to fix the argument passing to the logging function

## Testing

> [!NOTE]
> `sca` could not be tested, it requires connecting an agent


pereyra-m (approved):

sebasfalcone (approved):
I wasn't able to find any other instance of bad arguments in the logs

PR #27322, opened 2024-12-17
VD Description performance improvements.

### LINKED ISSUE #27314: Vulnerability Scanner - Performance improvement - DescriptionsHelper::vulnerabilityDescription optimization
## Description
After the investigations performed here:
- https://github.com/wazuh/wazuh/issues/27260

It was found that we are calling **DescriptionsHelper::vulnerabilityDescription** more than it needed:
https://github.com/wazuh/wazuh/blob/9b0f5b17058fddfba05e19bb67808fcd2f0db584/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/eventDetailsBuilder.hpp#L204-L208

Because the **context->m_vulnerabilitySource** doesn't change for the same event, this call:
https://github.com/wazuh/wazuh/blob/9b0f5b17058fddfba05e19bb67808fcd2f0db584/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/descriptionsHelper.hpp#L181-L183

Is redundant when using the helper in a loop, like in this example:
https://github.com/wazuh/wazuh/blob/9b0f5b17058fddfba05e19bb67808fcd2f0db584/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/eventDetailsBuilder.hpp#L200-L209

What we can do is calculate these sources before calling the helper and passing that pair instead.

> [!NOTE]
> The proposed function is just a POC, it should be analyzed for and tested

> [!TIP]
> [Here](https://github.com/wazuh/wazuh/issues/27260#issuecomment-2542754071) @pereyra-m implemented tracy to profile the module

## DoD
- [x] Improvement implemented
- [x] Improvement tested

Dwordcito:
|Related issue|
|---|
|Closes #27314|

This change avoid to call multiple times the same function returning the same value for all iterations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@ All notable changes to this project will be documented in this file.
 - Fixed concurrency issues in LRU caches ([#26073](https://github.com/wazuh/wazuh/pull/26073))
 - Removed all CVEs related to a deleted agent from the indexer. ([#26232](https://github.com/wazuh/wazuh/pull/26232))
 - Prevented an infinite loop when indexing events in the Vulnerability Detector. ([#26922](https://github.com/wazuh/wazuh/pull/26922))
-- Fixed segmentation fault in `DescriptionsHelper::vulnerabilityDescription`. ([#26842](https://github.com/wazuh/wazuh/pull/26842))
 - Fixed vulnerability scanner re-scan triggers in cluster environment. ([#24034](https://github.com/wazuh/wazuh/pull/24034))
 - Fixed an issue where elements in the delayed list were not purged when changing nodes. ([#27145](https://github.com/wazuh/wazuh/pull/27145))
 - Added logic to avoid re-scanning disconnected agents. ([#27145](https://github.com/wazuh/wazuh/pull/27145))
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -14,7 +14,6 @@ All notable changes to this project will be documented in this file.
 - Fixed concurrency issues in LRU caches ([#26073](https://github.com/wazuh/wazuh/pull/26073))
 - Removed all CVEs related to a deleted agent from the indexer. ([#26232](https://github.com/wazuh/wazuh/pull/26232))
 - Prevented an infinite loop when indexing events in the Vulnerability Detector. ([#26922](https://github.com/wazuh/wazuh/pull/26922))
-- Fixed segmentation fault in `DescriptionsHelper::vulnerabilityDescription`. ([#26842](https://github.com/wazuh/wazuh/pull/26842))
```
Dwordcito, line 17: This issue not is a part of versions less than 4.10

GabrielEValenzuela:
# Testing 🟢 

We conducted a baseline stress test on the `dev/4.10.0-stress-test` branch and compared it against the test results with the proposed changes, using Tracy for profiling.

The comparison of both profiling outputs shows a significant performance improvement

🍋 It's the baseline output (🟨 graphs)
♦️  It's the change output (🟥 graphs)

#### vulnerabilityDescription zone

image


The information used in this comparison, can be found [here](https://drive.google.com/file/d/1jryE6aQuYOypUN_X8DpDtGLv2gpbdD0Z/view?usp=sharing)

GabrielEValenzuela (approved):
Testing 🟢

Issue #26391, opened 2024-10-18
Engine - Comprehensive update of python internal tools


JcabreraC:
|Epic|
|---|
|#26520|

### Description:
With the shift towards a new server interface using cpp-httplib and Protobuf, this issue aims to update all Python internal tools to align with the updated architecture. This involves ensuring that all tools can interact with the new API and server setup without relying on outdated methods or configurations.

### Objectives:
- Update Python internal tools to be compatible with the new server and API interfaces.
- Enhance tool functionality to support development and operational processes effectively.

### Tasks:
1. **Assess Current Tools**: Evaluate all existing Python tools for their current state and compatibility with the new server setup.
2. **Refactor Tools**: Refactor tools to interact with the new API and server interfaces, removing any dependencies on deprecated methods.
3. **Enhance Tool Capabilities**: Enhance the functionalities of tools where necessary to support new features or improved workflows.
4. **Integration Testing**: Perform thorough testing to ensure tools function correctly with the updated server and API configurations.
5. **Documentation Update**: Revise tool documentation to reflect changes and provide guidance on their use with the new system architecture.

### Expected Outcome:
- Python internal tools that are fully compatible with the updated Wazuh-Engine architecture, offering robust support for development and operational tasks.
- Detailed and updated documentation that helps developers and system administrators utilize the tools effectively.

### DoD (Definition of Done):
- [x] All internal tools are assessed and updated as necessary.
- [x] Tools are fully integrated with the new API and server interfaces.
- [x] Enhanced functionalities are tested and confirmed to be working as expected.
- [ ] Documentation is comprehensive and up to date.

JcabreraC:
ETA extended by priority in other issues.

JcabreraC:
ETA extended by priority in other issues. Need merge #26390

JavierBejMen:
## Update
- Updated api-communication to use httpx (easy http client over unix socket)
- Updated endpoints in api-communication
- Added better error handling and messages
- Working on adding timeout parametrization to api-communication
- Working on resource handler update

JavierBejMen:
## Update
- Updated engine-catalog commands
- Updated and improved engine-clear command
- Working on engine-integration commands

JcabreraC:
ETA extended by priority in other issues.

PR #27394, opened 2024-12-20
Updates Engine tools to use new http API protocol

JavierBejMen:
|Related issue|
|---|
|#26391|

Updates tools and tests to use new http protocol:
- engine-suite
- api-communication
- integration tests
- health tests

Also changes engine test endpoint to use synchronous call when processing test events.


NahuFigueroa97 (approved):
LGTM!

Issue #21034, opened 2023-12-22
Create lint github action


GGP1:
## Description

In https://github.com/wazuh/wazuh/issues/20753, we are introducing a pre-commit hook to validate some conventions before opening a PR. 

Since the pre-commit hook is local and could be misconfigured or not present at all in the developer's development environment, this issue aims to create a github action that validates that the changes submitted in pull requests are correctly linted and formatted.

We could use something like

```yml
# framework-lint.yml
name: Lint
on: 
  pull_request:
    paths:
      - '.github/workflows/framework-lint.yml'
      - 'api/**'
      - 'framework/**'
      - 'integrations/**'
      - 'wodles/**'

jobs:
  lint:
    runs-on: ubuntu-latest 
    steps:
      - uses: actions/checkout@v4
      - uses: chartboost/ruff-action@v1
        with:
          src: '.ruff.toml'
```

## Checks

The following elements have been updated or reviewed (should also be checked if no modification is required):
- [x] Tests (unit tests, API integration tests).
- [x] Changelog.
- [x] Documentation.
- [x] Integration test mapping (using `api/test/integration/mapping/_test_mapping.py`).

GGP1:
## Update

Added a lint workflow for PyServer team Python code.

GGP1:
## Update

Applied suggested changes.

PR #27437, opened 2024-12-23
Add Python lint GitHub action

GGP1:
|Related issue|
|---|
| Closes https://github.com/wazuh/wazuh/issues/21034 |

## Description

Adds a Python lint workflow to prevent uncompliant code to be uploaded to PyServer-owned directories.

> [!Note]
> The workflow is expected to fail for this pull request, as the amount of changes required is enormous.
> 
> We should decide whether to fix all the errors in a specific PR or loose the rules to avoid being too strict.

### Tests

Some tests were performed in this pull request: https://github.com/wazuh/wazuh/pull/27447

### DIFF
--- a/.github/workflows/python-lint.yml
+++ b/.github/workflows/python-lint.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   lint:
-    runs-on: ubuntu-latest 
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
       - uses: astral-sh/ruff-action@v2
--- a/.github/workflows/python-lint.yml
+++ b/.github/workflows/python-lint.yml
@@ -1,0 +1,17 @@
+name: Python lint
+on: 
+  pull_request:
+    paths:
+      - '.github/workflows/python-lint.yml'
+      - 'api/**'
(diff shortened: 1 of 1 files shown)


nico-stefani (approved):
LGTM

.github/workflows/python-lint.yml
```diff
@@ -0,0 +1,17 @@
...
+      - '.github/workflows/python-lint.yml'
+      - 'api/**'
+      - 'apis/**'
+      - 'framework/**'
+      - 'integrations/**'
+      - 'wodles/**'
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
```
fdalmaup, line 14: We should define a specific version instead of using the latest to avoid future compatibility issues.
```suggestion
    runs-on: ubuntu-24.04
```

fdalmaup (changes requested):
Regarding the comment on the number of errors, since 5.0 is currently being developed and more changes are yet to be carried out, we should analyze if the workflow can be modified to just validate the PR modified/added files and have a manual trigger to analyze the complete Python codebase.

fdalmaup (approved):
LGTM!

PR #27458, opened 2024-12-27
Enforce Maild startup verification to ignore XML comments

elwali10:
## Description

Hello Team,

When a user add an xml comment to keep track of the configuration changes similar to `<!-- OLD (changed by elwali.karkoub) <email_notification>no</email_notification> -->` it does not allow the maild to startup while the email_notification is set to `yes`. It is caused by how the Wazuh control verify the configuration: 

Current Behavior:

notstartedmaild

The fix: consists of a simple exclusion of the XML comment.

FIXmaildstartup


juliancnn:
Hi @elwali10,

The fix contemplates a specific case when the configuration is:

```xml
    <!-- <email_notification>no</email_notification>  -->
    <email_notification>yes</email_notification>
```

But it does not contemplate the other common cases such as:

```xml
    <!--  old config
    <email_notification>no</email_notification> 
     -->
    <email_notification>yes</email_notification>

```

or 

```xml
<!-- Backup config
  <global>
    ...
    <email_notification>no</email_notification>
    ...
  </global>
-->

<--! test config -->
  <global>
    ...
    <email_notification>no</email_notification>
    ...
  </global>
```

I think the solution could be a bit more complex than it seems, as it would require a more integrated solution. 

I will check with the team

elwali10:
Hello @juliancnn, 

That is definitely a valid point. I have made the following tests and I think it might be doable using `awk`, I am sharing details below and let me know your opinion if should include it as function or not: 

- I have the logs file with below content: 

```
    <!-- <email_notification>no</email_notification>  -->
    <email_notification>yes</email_notification>


<!-- Backup config
  <global>
    ...
    <email_notification>no</email_notification>
    ...
  </global>
-->

<--! test config -->
  <global>
    ...
    <email_notification>yes</email_notification>
    ...
  </global>



    <!--  old config
    <email_notification>no</email_notification> 
     -->
    <email_notification>yes</email_notification>


```


- I parse it and to skip commented sections/lines (Printing out the result) using `awk 'BEGIN { inside_comment = 0 } /<!--/ { inside_comment = 1 } /-->/ { inside_comment = 0; next } !inside_comment && !/<!--/ { print }' /tmp/testlog` 

- Output result: 

```
    <email_notification>yes</email_notification>



  <global>
    ...
    <email_notification>yes</email_notification>
    ...
  </global>



    <email_notification>yes</email_notification>


```


Regards,
Wali

juliancnn:
Yes, I think using AWK is a good idea. In the case you share (I haven't tested it), but I would discard the line completely if it has an inline comment right? For example:

```xml
<email_notification>no</email_notification> <!-- Temporary disabled-->
```

elwali10:
Hello @juliancnn, 

I have found a different approach which would be global using `sed` to remove `inline comments` (replacing with spaces) and delete `multi comments` then the result can be used for the search. The implementation can be as follows: 

- The command `sed -e 's/<!--.*-->/ /g'  -e '/<!--/,/-->/d' file.conf`

- `file.conf` content: 

```
<!--  old config
<email_notification>no</email_notification> 
-->
<email_notification>yes</email_notification>

<!-- Backup config
  <global>
    ...
    <email_notification>no</email_notification>
    ...
  </global>
-->

<!-- test config -->
  <global>
    ...
    <email_notification>yes</email_notification>
    ...
  </global> 

<email_notification>no</email_notification> <!-- Temporary disabled-->


```

- Testing  : 
```
sed -e 's/<!--.*-->/ /g'  -e '/<!--/,/-->/d' file.conf
<email_notification>yes</email_notification>


 
  <global>
    ...
    <email_notification>yes</email_notification>
    ...
  </global> 

<email_notification>no</email_notification> 
```

- If you agree with the solution, I plan to include it to the code as follows: 
```
if [ X"$i" = "Xwazuh-maild" ]; then
    sed -e 's/<!--.*-->/ /g' \
        -e '/<!--/,/-->/d' \
        ${DIR}/etc/ossec.conf | grep "<email_notification>no<" >/dev/null 2>&1
    if [ $? = 0 ]; then
        continue
    fi
fi
```

Let me know what you think @juliancnn. 

Thanks.

fdalmaup:
It has been decided to close the PR since it is not providing an integral solution to the script.

Issue #27520, opened 2025-01-08
Add match condition condition is in the vulnerability states


Dwordcito:
## Description
We need to add the match condition to [vulnerability events](https://github.com/wazuh/wazuh/blob/4.10.0/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/eventDetailsBuilder.hpp), at the field defined here:
- https://github.com/wazuh/wazuh/issues/27518

The logic used for the alerts should be followed:
https://github.com/wazuh/wazuh/blob/3a7a2fb95231df860ee15b0398613f76f55f68eb/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/scanOsAlertDetailsBuilder.hpp#L180-L207

https://github.com/wazuh/wazuh/blob/3a7a2fb95231df860ee15b0398613f76f55f68eb/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/eventPackageAlertDetailsBuilder.hpp#L182-L209

## DoD
- [x] Changes Implemented
- [x] Tests updated

sebasfalcone:
## Issue blocked
We are awaiting the spike to define the field in which the data will be appended:
- https://github.com/wazuh/wazuh/issues/27520

PR #27603, opened 2025-01-14
Added condition field in indexed vulnerabilities

MiguelazoDS:
|Related issue|
|---|
|#27520|

## Description

- Vulnerability condition was added to indexed vulnerabilities report considering the same approach as in alerts. 
- Some UTs were modified to test the changes with mocked conditions 
- A default value was proposed to avoid having null condition. 

## Logs/Alerts example

image

- Indexed vulnerabilities.

[vulnerabilities.tar.gz](https://github.com/user-attachments/files/18416042/vulnerabilities.tar.gz)

### DIFF
--- a/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/eventDetailsBuilder.hpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/eventDetailsBuilder.hpp
@@ -267,6 +267,34 @@ public:
                                 .at("adp");
                         ecsData["vulnerability"]["under_evaluation"] = isUnderEvaluation();
 
+                        if (const auto it = context->m_matchConditions.find(cve);
+                            it != context->m_matchConditions.end())
+                        {
+                            if (it->second.condition == MatchRuleCondition::LessThanOrEqual)
+                            {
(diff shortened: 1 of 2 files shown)


src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/eventDetailsBuilder.hpp
```diff
@@ -267,6 +267,35 @@ class TEventDetailsBuilder final : public AbstractHandler<std::shared_ptr<TScanC
...
+                                ecsData["vulnerability"]["scanner"]["condition"] = "Package default status";
+                            }
+                            else if (it->second.condition == MatchRuleCondition::Equal)
+                            {
+                                ecsData["vulnerability"]["scanner"]["condition"] =
+                                    "Package equal to " + it->second.version;
+                            }
+                            else
+                            {
+                                ecsData["vulnerability"]["scanner"]["condition"] = "Unknown";
```
pereyra-m, line 294: After evaluating the code, this situation doesn't seem possible.
Writing unknown doesn't harm, and can give us more information in the future in case we add more conditions without modify this.

pereyra-m (approved):
LGTM!

Issue #27569, opened 2025-01-10
Failed test on `vulnerability-scanner-modules-asan (ubuntu-22.04)`


JcabreraC:
### **Description:**  
While merging branch `4.11.0` into `4.12.0`, we encountered a failure in the `vulnerability-scanner-modules-asan` (Ubuntu 22.04) during the `IndexerConnectorTest.QueueCorruptionTest`.  

The test terminates with the following error message:  
```bash
[ RUN ] IndexerConnectorTest.QueueCorruptionTest
terminate called after throwing an instance of 'std::length_error'
what(): basic_string::_M_create
```

### **Relevant workflow and job details:**  
Workflow: [GitHub Actions Workflow](https://github.com/wazuh/wazuh/actions/runs/12707727643/job/35423556683?pr=27567)

MiguelazoDS:
# Update 

The issue is easily reproduced using the constructor `string (const char* s, size_t n);` and passing a negative number (or a really huge size). 

![Image](https://github.com/user-attachments/assets/3a59a5d3-be92-4f11-90c7-cc7747551cfd)

Although I don't see exactly that in the `IndexerConnectorTest.QueueCorruptionTest` test, but I suspect about this code section here where there's a string manipulation and paths comes into discussion. 

The code here removes a MANIFEST file when is found. 

```cpp
    bool corrupted {false};
    std::string prefix {"queue/indexer/indexer_connector_test/MANIFEST"};
    for (const auto& entry : std::filesystem::directory_iterator("queue/indexer/indexer_connector_test"))
    {
        if (entry.path().string().substr(0, prefix.size()).compare(prefix) == 0)
        {
            std::filesystem::remove(entry.path());
            corrupted = true;
            break;
        }
    }
```

Although I could not reproduce anything by simulating changes in the filesystem while the loop is being ran, changing permissions, ownership, etc. I still think that this may be the cause of this random error.

pereyra-m:
@MiguelazoDS Great RCA!

Now that I see all this information, I wonder if it could be related to this
https://github.com/wazuh/wazuh/pull/26684/commits/d02482cd596304f8be4d4b16127c07ffd72d5588

MiguelazoDS:
Thanks @pereyra-m for this information, effectively there are so many issues that arise once we compile with sanitize thread flag. 

![Image](https://github.com/user-attachments/assets/d502c413-5261-4d23-8b47-a59522547573)

[...]

PR #27685, opened 2025-01-16
Fix race condition error indexer connector test

MiguelazoDS:
|Related issue|
|---|
|#27569|

## Description

This PR back ports a race condition fix that solves the error reported in the issue. 

## DoD

- With the fix, it is not reproduced in 10k executions
image

- Without the fix it is reproducible way before than 10k 
image
image
image


GabrielEValenzuela (approved):
Great job!

pereyra-m (approved):
LGTM!

PR #27750, opened 2025-01-21
Add cleanup logic for content folder

### LINKED ISSUE #27691: Implementation - Vulnerability detection doesn't clean content temporal folder
| Wazuh Version | Component        | Install Type | Install Method   | Platform |
|---|-----------------|--------------|----------------|---------|
| 4.9.x/4.10.x        | Content Updater | Manager       | Packages/Sources | All      |

### Description  
During tests realized by MaP [here](https://groups.google.com/g/wazuh/c/Pd7n4LyCjS0/m/EH7aeen1BgAJ), with offline content in the vulnerability detection module, it was observed that temporary files in the `vd_updater/tmp/content` directory are not automatically cleaned up after the setup process is completed. Currently, these files are only removed when the module is restarted.  

Image

Image

Image

This behavior can lead to unnecessary disk space usage over time. To address this, a cleanup mechanism should be implemented to ensure that temporary files are properly removed after the setup process concludes.

> [!TIP]
> We should check the type of offline content and remove its corresponding folder here.

https://github.com/wazuh/wazuh/blob/c3afcedb8f246958840c382c58f9b891cdb773c2/src/shared_modules/content_manager/src/components/cleanUpContent.hpp#L38-L48

### Definition of Done (DoD)
- [x] Implement a mechanism to automatically clean up the `vd_updater/tmp/content` folder after the setup process.
- [x] Verify that temporary files are not present after a successful setup.

GabrielEValenzuela:
|Related issue|
|---|
|Closes #27691 |

# Objective

This PR aims to add the removing logic for both, the download and content folder inside of the `vd_updater/tmp` directory.

# PoC

gif_issue

### DIFF
--- a/src/shared_modules/content_manager/tests/unit/cleanUpContent_test.cpp
+++ b/src/shared_modules/content_manager/tests/unit/cleanUpContent_test.cpp
@@ -40,5 +40,5 @@ TEST_F(CleanUpContentTest, OutputPathSet)
 
     // Check the folder content
     EXPECT_TRUE(std::filesystem::is_empty(DOWNLOAD_DIR));
-    EXPECT_FALSE(std::filesystem::is_empty(CONTENTS_DIR));
+    EXPECT_TRUE(std::filesystem::is_empty(CONTENTS_DIR));
 }
--- a/src/shared_modules/content_manager/src/components/cleanUpContent.hpp
+++ b/src/shared_modules/content_manager/src/components/cleanUpContent.hpp
@@ -34,21 +34,44 @@ private:
      */
     void cleanUp(const UpdaterContext& context) const
     {
-        // Get the path to the folder.
-        const auto& path = context.spUpdaterBaseContext->downloadsFolder;
+
+        logDebug1(WM_CONTENTUPDATER,
+                  "Cleaning up the folder: %s.",
+                  context.spUpdaterBaseContext->downloadsFolder.string().c_str());
 
         // Check if the path exists.
-        if (!std::filesystem::exists(path))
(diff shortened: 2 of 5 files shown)


MiguelazoDS:
# Testing :green_circle: 

- Configuration 
```xml
    <offline-url>file:///home/vagrant/CONTENT/1251692_1737570906.zip</offline-url>
```

## 4.12.0 

While processing I see this 
image
image

After finishing the offline processing
image

> [!NOTE]
> The json file is not deleted

## Dev branch 

While processing I see this 
image
image

After finishing the offline processing
image
image

src/shared_modules/content_manager/tests/unit/cleanUpContent_test.cpp
```diff
@@ -40,5 +40,5 @@ TEST_F(CleanUpContentTest, OutputPathSet)
 
     // Check the folder content
     EXPECT_TRUE(std::filesystem::is_empty(DOWNLOAD_DIR));
-    EXPECT_FALSE(std::filesystem::is_empty(CONTENTS_DIR));
+    EXPECT_TRUE(std::filesystem::is_empty(CONTENTS_DIR));
```
pereyra-m, line 43: Comment: the module didn't delete the folder before because the router published the files. Now, the callback reads the data and after, it's safe to remove them.

pereyra-m (approved):
LGTM!

We have verified that the even when the `policyManager` sets the offline content as `raw`, the `contentManager` module auto-deduces the compression.

We won't update the configuration. The current approach is working.

PR #27771, opened 2025-01-22
Improve delimiters on xml

juliancnn:
This PR improves detection of delimiters in xml strings


Dwordcito (approved):

branchnetconsulting:
Does this by any chance fix the case where the dashboard rule editor can't handle a rule line containing a string ending with `//` like this for example?

`<field name="office365.ObjectId" type="pcre2">(?i)^C:\\\$Recycle.Bin\\</field>`

PR #27913, opened 2025-01-28
Modify check 3.5.3.1.2 to ensure nftables and iptables don't co-exist in Ubuntu SCAs

IsExec:
## Description
User reported that their host fail the check even though it had only nftables without iptables.

## Change to be made

```
- condition: all  # iptables or nftables installed AND iptables or nftables not installed
        rules:
          - "c:dpkg-query -W -f='${binary:Package}\\t${Status}\\t${db:Status-Status}\\n' nftables iptables -> r:install ok installed" 
          - "c:dpkg-query -W -f='${binary:Package}\\t${Status}\\t${db:Status-Status}\\n' nftables iptables -> r:unknown ok not-installed|no packages found matching"
```

### DIFF
--- a/ruleset/sca/ubuntu/cis_ubuntu18-04.yml
+++ b/ruleset/sca/ubuntu/cis_ubuntu18-04.yml
@@ -2007,8 +2007,8 @@ checks:
       - pci_dss_v3.2.1: ["1.4"]
     condition: all
     rules:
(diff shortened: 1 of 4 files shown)


ooniagbi:
Please link to issue.

ooniagbi (approved):

PR #27935, opened 2025-01-29
Rename generic SCA policy in check file templates

MarcelKemp:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-packages/issues/3173|

## Description

This PR has the following modifications:
- Renaming of the generic SCA file in the check file templates + Change of file size.
- Addition of another `*` pattern to accept a larger number of characters in the revision field.

According to the SPECS, it has not been necessary to modify them, because the name of the generic configuration (`sca.files`), which is the one checked by the package, has not been changed.


.github/actions/check_files/rpm5_linux_agent_amd64.csv
```diff
@@ -1,58 +1,58 @@
 full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_error
-/var/ossec/tmp/sca-4.*.el5-tmp/rocky/cis_rocky_linux_9.yml,root,wazuh,640,file,-rw-r-----,300524,0.1
```
sebasfalcone, line 2: This changes the behaviour, its really expected?
image
image

sebasfalcone (commented):
- Manager changes 🟢 
- Agent changes 🟡

.github/actions/check_files/rpm5_linux_agent_amd64.csv
MarcelKemp: Yes, since the **version** always comes with a **_revision_**. 
So it will always come `4.<minor>.<patch>-<revision>`.

sebasfalcone: Perfect, then its useful to prevent non expected file names

sebasfalcone (approved):
LGTM!

lchico (approved):
LGTM!

Issue #26056, opened 2024-09-30
Engine - Escaping not working Correctly in parse logpar


JcabreraC:
## Description

There is an issue with the escaping mechanism in the `parse logpar` function. Specifically, the escape sequence `\<field>\>` does not function as expected. This leads to incorrect parsing of log fields that require escaping.

## Steps to Reproduce

1. Use the `parse logpar` function with an escaped field format, such as `\<field>\>`.
2. Observe that the escaping does not work, and the log parsing fails or behaves incorrectly.

## Expected Behavior

The escape sequence `\<field>\>` should be recognized properly, allowing for correct parsing of fields that contain special characters.

## Actual Behavior

The engine fails to process the escape sequence as intended, leading to errors in parsing the logs that contain these fields.

## Impact

This bug prevents the proper parsing of logs that require escaping for specific fields, which can lead to data loss or incorrect log analysis.

JavierBejMen:
Not a bug, we only need to espace the opening angle char `<`.

PR #28036, opened 2025-02-05
Added more cases to engine logpar utest

JavierBejMen:
|Related issue|
|---|
|#26056|

Not a bug, when parsing a literal we only need to escape opening angles `<`.


NahuFigueroa97 (approved):
LGTM! unit test here https://github.com/wazuh/wazuh/actions/runs/13156962714/job/36718537718?pr=28036

PR #28077, opened 2025-02-07
Ignore unexpected fields in CVE 5.0 schema

### LINKED ISSUE #27976: Spike - Analyze the addition of new fields in CTI events
## Description
We need to investigate the possibility of adding new [x_* fields](https://github.com/CVEProject/cve-schema/blob/a29f28e5d48383cc5e179f9c6655ac49e8ffe1f9/schema/CVE_Record_Format.json#L662) into the events received from CTI and its implication. The idea is to generate new consumers that contain these new fields and prevent the modules from failing in case they encounter 

> [!NOTE]
> Check the **skip_unexpected_fields_in_json** option to see if we can ignore these unused fields

## DoD
- [x] Impact determined (if any)
- [x] Find all places where the option should be applied

pereyra-m:
|Related issue|
|---|
|Closes #27976|

## Description

This PR adds the `skip_unexpected_fields_in_json` flatbuffer option to the CVE 5.0 JSON parsing to avoid error messages in cas `x_` fields are received. 

## Tests

  - [x] Linux
 
### DIFF
--- a/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/eventDecoder.hpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/eventDecoder.hpp
@@ -160,6 +160,7 @@ private:
                     }
                     flatbuffers::IDLOptions options;
                     options.output_default_scalars_in_json = true;
+                    options.skip_unexpected_fields_in_json = true;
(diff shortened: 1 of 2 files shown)


pereyra-m:
The changes won't be applied in a Spike, but we have confirmed what we needed.

Issue #28287, opened 2025-02-18
The generated RPM package does not include the KVDBs


NahuFigueroa97:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| v5.x | Wazuh component | Manager | RPM | Ubuntu 24.04 |

## Description:

A problem was detected when installing the Wazuh RPM package, where the key-value databases (KVDBs) are not being included in the package. This causes the router to be disabled and prevents events from being ingested into the indexer, leading to their discard.

## Expected Behavior:

- After installing the package on Debian, executing engine-kvdb --api-socket /run/wazuh-server/engine-api.socket list returns a list of properly loaded KVDBs.

- The router is enabled, and events can be ingested.

## Observed Behavior in RPM:

- Running engine-kvdb --api-socket /run/wazuh-server/engine-api.socket list returns an empty list [].

- Running engine-router get default indicates that the router is disabled.

- Executing engine-clear -f removes multiple routes, policies, and integrations, suggesting that assets are loaded but not KVDBs.

- After installing the RPM package, the directory /var/lib/wazuh-server/engine/kvdb/ is empty, whereas in Debian, it contains files such as 000004.log, CURRENT, IDENTITY, LOCK, LOG, MANIFEST-000005, OPTIONS-000049, OPTIONS-000051.

## Evidence:

- In Debian, KVDBs load correctly and are listed with engine-kvdb list.

- In RPM, KVDBs are not listed, and /var/lib/wazuh-server/engine/kvdb/ is empty after installation.

- In the GitHub Actions workflow where the build is generated ([job link](https://github.com/wazuh/wazuh/actions/runs/13334551204/job/37246670520)), it is shown that the router is enabled, contradicting the behavior in the RPM installation.

- It was verified that this issue did not exist in previous versions of the RPM package.

## Impact:

- Events cannot be ingested as the router is disabled.

- This may affect production environments that rely on event ingestion for detection and response.

## Possible Causes:

- Recent changes in the RPM package generation affecting KVDB inclusion.

- Differences in the installation process between Debian and RPM.

- Issues with the /var/lib/wazuh-server/engine/kvdb/ path during RPM installation.

## Request:

- Verify if there have been recent changes in the RPM packaging process that may have affected KVDB inclusion.

- Compare Debian and RPM installations to identify differences in KVDB generation and location.

- Evaluate a possible solution to ensure KVDBs are correctly included in the RPM package.

Issue #27954, opened 2025-01-30
Error in creating the RPM package locally


javiersanchz:
## Description 

During the creation of the RPM package locally for master, the following was obtained:

```console
symlinked /usr/lib/debug/usr/share/wazuh-server/framework/python/bin/python3.10.debug to /usr/lib/debug/usr/share/wazuh-server/framework/python/bin/python3.debug
symlinked /usr/lib/debug/usr/share/wazuh-server/framework/python/lib/libpython3.10.so.1.0.debug to /usr/lib/debug/usr/share/wazuh-server/framework/python/lib/libpython3.10.so.debug
+ /usr/local/lib/rpm/brp-compress /usr/local
+ /usr/local/lib/rpm/brp-strip /usr/bin/strip
+ /usr/local/lib/rpm/brp-strip-static-archive /usr/bin/strip
+ /usr/local/lib/rpm/brp-strip-comment-note /usr/bin/strip /usr/bin/objdump
Processing files: wazuh-server-5.0.0-test.x86_64
error: Directory not found: /build_wazuh/rpmbuild/BUILDROOT/wazuh-server-5.0.0-test.x86_64/var/lib/wazuh-server/engine/kvdb
error: File not found: /build_wazuh/rpmbuild/BUILDROOT/wazuh-server-5.0.0-test.x86_64/var/lib/wazuh-server/engine/kvdb/*


RPM build errors:
    Directory not found: /build_wazuh/rpmbuild/BUILDROOT/wazuh-server-5.0.0-test.x86_64/var/lib/wazuh-server/engine/kvdb
    File not found: /build_wazuh/rpmbuild/BUILDROOT/wazuh-server-5.0.0-test.x86_64/var/lib/wazuh-server/engine/kvdb/*
+ return 1
+ return 1
+ clean 1
+ exit_code=1
+ find /home/wazuh/Git/wazuh/packages/rpms/amd64 '(' -name '*.sh' -o -name '*.tar.gz' -o -name 'wazuh-*' ')' '!' -name docker_builder.sh -exec rm -rf '{}' +
+ exit 1
```
The error indicates that the RPM package creation process is looking for the directory `/build_wazuh/rpmbuild/BUILDROOT/wazuh-server-5.0.0-test.x86_64/var/lib/wazuh-server/engine/kvdb` but cannot find it.

From GitHub Actions, this error does not occur, as can be seen in: https://github.com/wazuh/wazuh/actions/runs/13055219002

[...]

PR #28291, opened 2025-02-18
Remove wrong RPM file spec for kvdb

jnasselle:
|Related issue|
|---|
|#28287|

## Description

This PR aims to fix a wrong RPM file spec for kvdb folder and files when the package is created using the GitHub workflow https://github.com/wazuh/wazuh/blob/main/.github/workflows/packages-build-server.yml, leading to errors during wazuh-server runtime

## Fix proposal

According to [RPM Virtual File Attribute(s)](https://rpm-software-management.github.io/rpm/manual/spec.html), ghost directive will cause the exclusion of those files from the package.

## Fix cons

Including files wildcards on packages is a not recommended situation at `%files` directive, since we are not getting noticed if those files were, in fact, part of our package. This could lead to silent bugs due to undeployed files.

That's why RPM is stringent on check-files step.

We should move the S3_PRECOMPILED_STORE fetching from workflows to a more deeper mechanism, such as SPECS or source code installation/compilation. With this approach, we will include the storage for packages (created by workflow or locally) and source code installation

**TL;DR:
:warning: 
After merging this issue, https://github.com/wazuh/wazuh/issues/27954 should be opened again**

## Testing

- With engine storage (GHA):  https://github.com/wazuh/wazuh/actions/runs/13435081427


GGP1 (approved):
LGTM

fdalmaup (approved):

PR #28292, opened 2025-02-18
Add logic of remove of delete by query when indexer is disabled

### LINKED ISSUE #28286: Properly handle all types of elements at vulnerability detector when indexing is disabled
## Description

This development implemented a type of element related to the delete by query:
- https://github.com/wazuh/wazuh/pull/26232#top

And this other one keeps the local DB synced in case VD is enabled but indexing not:
- https://github.com/wazuh/wazuh/pull/26132#top

But these changes aren't aligned.
This check
https://github.com/wazuh/wazuh/blob/b3de9acf06147d614f053e87f959a75a0a132e86/src/shared_modules/indexer_connector/src/indexerConnector.cpp#L590

Isn't found here

https://github.com/wazuh/wazuh/blob/b3de9acf06147d614f053e87f959a75a0a132e86/src/shared_modules/indexer_connector/src/indexerConnector.cpp#L817

Even when this situation isn't frequent, we should fix this.

## DoD

- [ ] Update the corresponding class with the proper event handling
- [ ] Add tests

GabrielEValenzuela:
|Related issue|
|---|
|Closes #28286 |

# Objective

This PR aims to add the missing logic when the indexer is disabled and delete by query operation is received.

## Testing

> [!NOTE]
> Disable index status on vulnerability detection configuration

### Steps
- Start manager and enable debug logging
- Connect an agent
- Remove the agent

image

> [!NOTE]
> Log entries were added only for testing purposes

### DIFF
--- a/src/shared_modules/indexer_connector/src/indexerConnector.cpp
+++ b/src/shared_modules/indexer_connector/src/indexerConnector.cpp
@@ -818,6 +818,14 @@ IndexerConnector::IndexerConnector(
                 {
                     m_db->delete_(id);
                 }
+                // We made the same operation for DELETED_BY_QUERY as for DELETED
+                else if (parsedData.at("operation").get_ref<const std::string&>().compare("DELETED_BY_QUERY") == 0)
+                {
+                    for (const auto& [key, _] : m_db->seek(id))
+                    {
+                        m_db->delete_(key);
+                    }
+                }
                 else
                 {
                     const auto dataString = parsedData.at("data").dump();
--- a/src/shared_modules/indexer_connector/src/indexerConnector.cpp
+++ b/src/shared_modules/indexer_connector/src/indexerConnector.cpp
@@ -828,6 +828,12 @@ IndexerConnector::IndexerConnector(
                 }
                 else
                 {
+                    // If the data does not contain the required fields, log a warning and continue.
+                    if (!parsedData.contains("data"))
+                    {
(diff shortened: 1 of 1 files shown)


src/shared_modules/indexer_connector/src/indexerConnector.cpp
```diff
@@ -818,6 +818,14 @@ IndexerConnector::IndexerConnector(
...
                     m_db->delete_(id);
                 }
+                // We made the same operation for DELETED_BY_QUERY as for DELETED
+                else if (parsedData.at("operation").get_ref<const std::string&>().compare("DELETED_BY_QUERY") == 0)
+                {
+                    for (const auto& [key, _] : m_db->seek(id))
+                    {
+                        m_db->delete_(key);
+                    }
+                }
```
pereyra-m, line 828: The changes are Ok.

I'd just add the check for `data` field below, now that we are aligning both methods.
Moving the duplicate code to a common function might require more changes, so we need to replicate all cases.

pereyra-m (commented):
Gj!

Just a comment.
We can take the opportunity to make another fix now that we are correcting this.

src/shared_modules/indexer_connector/src/indexerConnector.cpp
GabrielEValenzuela: Done! Thanks

pereyra-m (approved):
LGTM!

Issue #28261, opened 2025-02-17
Error in `vulnerability-scanner-modules-qa (4.11.0)`. Missing Log entries for CVE processing


JcabreraC:
## Description  

During the execution of the `vulnerability-scanner-modules-qa (4.11.0)` test, the process failed due to missing log entries for CVE processing. The failure was observed in workflow run [#13367125451](https://github.com/wazuh/wazuh/actions/runs/13367125451/job/37327252594).  

## Steps to Reproduce  

1. Trigger the `vulnerability-scanner-modules-qa (4.11.0)` test workflow.  
2. Observe the test execution.  
3. The test fails due to missing log entries related to CVE processing.  

## Error
```
------------------------------ Captured log call ------------------------------- 
ERROR test_efficacy_log:test_efficacy_log.py:267 Log entry not found: Processing and publish key: CVE-2024-23206
ERROR test_efficacy_log:test_efficacy_log.py:267 Log entry not found: Processing and publish key: CVE-2024-23213
```

## Expected Behavior  

The test should locate and validate the expected log entries for CVE processing, ensuring the vulnerability scanner is functioning correctly.

PR #28300, opened 2025-02-19
Removed discarded CVEs for macOS and change the affected status for rhel CVEs

MiguelazoDS:
|Related issue|
|---|
|#28261|

## Description

Changes applied: 

- Some CVEs that affect macOS software do not have the lower bound version. They were discarded.
- CVEs that affect the kernel changed the default status to affected.


pereyra-m (approved):
LGTM!

fdalmaup (approved):
LGTM!

Issue #28348, opened 2025-02-21
Global queries - Technical debt - UT failing


pereyra-m:
## Description

After creating the new module workflow, it was found that the UT for vulnerability scanner and the legacy server tests are failing.

## DoD

- [x] Find the cause and fix the tests
- [x] CI Pass

MiguelazoDS:
# Update 

Since we do not have 4.13.0 branch, I'll use 4.12.0

MiguelazoDS:
# Update 

Fixed legacy tests, working on vulnerability detector compilation after flatbuffer schema location change

MiguelazoDS:
# Update 

Fixed vulnerability scanner compilation (due to flatbuffer schema location change)

pereyra-m:
I will move the ETA so the PR can be reviewed and any suggested changes can be applied.

MiguelazoDS:
# Update 

Rebased, so many things changed from my first PR. I removed all conflicts by keeping the already existent ones in the base branch.

MiguelazoDS:
@Dwordcito after taking a look at the changes in legacy tests, most of the check deletion are because now a conditional does not accept invalid event messages, and the other one due to the split of router handles. 

That was discussed in the PR here https://github.com/wazuh/wazuh/pull/28360#discussion_r1968369430 and here https://github.com/wazuh/wazuh/pull/28360#discussion_r1968373708

Dwordcito:
Delayed by final reviewer -> @Dwordcito

PR #28345, opened 2025-02-20
Add inventory harvester test workflows

pereyra-m:
|Related issue|
|---|
|Closes #27897|

## Description

This PR fixes the current workflows compilation and adds a new one for inventory harvester module.
The failing tests won't be fixed here.

Also, an integration test was created to verify the test tool properly indexes the events.

PR #28360, opened 2025-02-21
Fixed vulnerability scanner and legacy server test cases

MiguelazoDS:
|Related issue|
|---|
|#28348|

## Description

This PR includes the following changes: 

- Fix legacy server tests: 
  - remoted 
  - wazuh_db/integrity
- Vulnerability/flatbuffer schema compilation 
- Fixing vulnerability scanner tests
  - Wrong variable type
  - Some changes in the fbs now admit missing fields (required() flag was removed)


src/shared_modules/utils/flatbuffers/CMakeLists.txt
```diff
@@ -7,6 +7,10 @@ enable_testing()
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
+if (NOT SRC_FOLDER)
+    get_filename_component(SRC_FOLDER     ${CMAKE_SOURCE_DIR}/../../../ ABSOLUTE)
```
GabrielEValenzuela, line 11: ❓ Can we avoid relatives' paths?

GabrielEValenzuela (commented):
Great job Migue!

One little question

src/shared_modules/utils/flatbuffers/CMakeLists.txt
pereyra-m: I've done the same here

https://github.com/wazuh/wazuh/pull/28345/files

The thing is that the `SRC_FOLDER` variable is defined at the root CMakeLists
https://github.com/wazuh/wazuh/blob/enhancement/27897-add-inventory-workflows-automation/src/CMakeLists.txt#L18

When it isn't, then it means that only this module is being built.
In that case, we either define the variable in a workflow or we get the path to `src` this way.

src/unit_tests/remoted/test_secure.c
```diff
@@ -2225,20 +2235,7 @@ void test_router_message_forward_invalid_delta_json_message(void** state)
...
-                  formatted_msg,
-                  "Unable to forward message "
-                  "'{\"agent_info\":{\"agent_id\":\"001\",\"agent_ip\":\"192.168.33.20\",\"agent_name\":\"focal\"}}' "
-                  "for agent 001");
-
-    will_return(__wrap_OSHash_Get_ex_dup, NULL);
-    expect_value(__wrap_OSHash_Get_ex_dup, self, (OSHash*)1);
-    expect_string(__wrap_OSHash_Get_ex_dup, key, data->agent_id);
-
+    // This type of message must be discarded
```
pereyra-m, line 2238: Just a question, why do we remove the message about the discarded message?

src/wazuh_db/wdb_integrity.c
```diff
@@ -57,7 +57,7 @@ extern void mock_assert(const int result, const char* const expression,
 #endif
 
 void wdbi_report_removed(const char* agent_id, wdb_component_t component, sqlite3_stmt* stmt) {
-    if (!router_fim_events_handle || !router_inventory_events_handle) {
+    if (!router_fim_events_handle && !router_inventory_events_handle) {
```
pereyra-m, line 60: Question: why do we change "any of the handles missing is an error" to "only is an error to have both missing"?

src/wazuh_modules/vulnerability_scanner/tests/unit/eventSendReport_test.cpp
```diff
@@ -57,7 +57,6 @@ const std::string SYNC_STATE_MSG {
                          "version": "0.1"
                     },
                     "index": "e051a99ad3950084b538c6538c0fcad0f1c4c713",
-                    "timestamp": ""
```
pereyra-m, line 60: Question: just to understand, why do we remove this?
It has to do with the schemas, but I don't see it.

src/wazuh_modules/vulnerability_scanner/tests/unit/scanContext_test.cpp
```diff
@@ -763,7 +760,8 @@ TEST_F(ScanContextTest, TestSyscollectorDeltasNoOperation)
 {
     flatbuffers::Parser parser;
     ASSERT_TRUE(parser.Parse(syscollector_deltas_SCHEMA));
-    ASSERT_FALSE(parser.Parse(DELTA_NO_OPERATION_MSG.c_str()));
+    // Parser does not fail when operation is missing.
```
pereyra-m, line 763: We can ask about this.

src/wazuh_modules/vulnerability_scanner/CMakeLists.txt
```diff
@@ -104,7 +104,10 @@ add_library(vulnerability_scanner SHARED
     )
 target_compile_options(vulnerability_scanner PRIVATE -Wall -Wextra -Wshadow -Wnon-virtual-dtor -Woverloaded-virtual -Wunused -Wcast-align -Wformat=2)
 
-add_dependencies(database_feed compile_schemas_vd compile_schemas_input) #Add a dependency between top-level targets.
+if (NOT TARGET compile_schemas_input)
+    include(${SRC_FOLDER}/shared_modules/utils/flatbuffers/schemas/CMakeLists.txt)
```
pereyra-m, line 108: +1 This is another way to solve the problem. The dependency between the CMakeLists file exists, and here you are simply defining the missing target.

pereyra-m (changes requested):
GJ!

src/wazuh_modules/vulnerability_scanner/tests/unit/scanContext_test.cpp
MiguelazoDS: Yeap, I fixed the tests, but I'm not sure why that fbs changed.

src/wazuh_modules/vulnerability_scanner/tests/unit/eventSendReport_test.cpp
MiguelazoDS: The timestamp was removed from fbs

src/wazuh_db/wdb_integrity.c
MiguelazoDS: These two handles were added besides the existent one (router_agent_event_handle (or something like that). I found that a mistake, because the handle is always fim or inventory, I don't think we should get the error message "Router handle not available" when we certainly have the router handle.

src/unit_tests/remoted/test_secure.c
MiguelazoDS: Due to the recently changes implemented, now that is invalid, so those messages no longer happen.

src/shared_modules/utils/flatbuffers/CMakeLists.txt
GabrielEValenzuela: Excellent, solved :)

src/wazuh_db/wdb_integrity.c
pereyra-m: I've reviewed the code and we have 3 handles:
- WDB_AGENT_EVENTS_TOPIC: is used to report the agent deletion and/or upgrade
- WDB_FIM_EVENTS_TOPIC: reports RSync deletion events for FIM
- WDB_INVENTORY_EVENTS_TOPIC: reports RSync deletion events for syscollector

This method at `wdb_integrity` is in charge of reporting RSync deletion events, and the function `wdbi_report_removed()` will use the corresponding handle according to the component. The WDB_AGENT_EVENTS_TOPIC handle isn't involved here.

Checking if any of the handles isn't available is a way to fail early.
We can improve the logging and say that the one of the handles for deletion events isn't available, but we shouldn't change this behavior in this issue.

GabrielEValenzuela (approved):
- Code 🟢 
   - I think that all issues has been resolved and testing is passing on GA

Great work Migue!

pereyra-m (approved):
Great job.
LGTM!

src/wazuh_modules/vulnerability_scanner/tests/unit/scanContext_test.cpp
pereyra-m: @Dwordcito 

We think that the `required` flags in the .fbs help to discard invalid messages. But this has changed.

Issue #27879, opened 2025-01-28
FIM whodata eBPF: Development


jotacarma90:
## Description

As part of the effort to enhance the **File Integrity Monitoring (FIM)** module in Wazuh, this issue focuses on the development of the **eBPF driver** for who-data monitoring. The goal is to implement a performant, modular, and maintainable eBPF-based solution for real-time file monitoring, capturing user and process context (who-data) while ensuring compatibility with Linux Kernel 5.8+.

The eBPF driver must be implemented as a **dynamic library**, rather than being integrated monolithically into FIM, to maintain modularity and facilitate future enhancements.

## Requirements

1. The eBPF driver must support **Linux Kernel 5.8+**.
2. The implementation must be modular, with the eBPF logic encapsulated in a **dynamic library**.
3. The eBPF-based monitoring must provide the same core functionality as the Audit-based who-data implementation:
   - Detect file operations (creation, modification, deletion, renaming).
   - Capture actor details (UID, process ID, process name).
4. The implementation should prioritize **performance efficiency**:
   - Minimize kernel overhead.
   - Avoid excessive event generation by filtering unneeded paths.
   - Ensure minimal EPS (events per second) overhead compared to the Audit driver.
5. The fallback logic must:
   - Prefer eBPF when available.
   - Fall back to Audit if eBPF is unavailable.
   - Fall back to inotify (without who-data) if both are unavailable.
6. The new implementation must be tested across all **supported Linux distributions**.

## BPF driver architecture

```mermaid
flowchart LR
 subgraph s1["eBPF [Kernel]"]
        kopen("<code>open</code> hook")
        ksetattr("<code>setattr</code> hook")
        kunlink("<code>unlink</code> hook")
        ktable(["FD Table"])
        kbuffer(["Buffer"])
  end
 subgraph s2["FIM [User]"]
        udispatch("Dispatch")
    subgraph s3["BPF Provider"]
        ufilter["Filter Path"]
        ubuffer(["Buffer"])
    end
  end
    kopen --> kbuffer
    kopen -.-> ktable
    kunlink -.-> ktable
    ktable <-.-> ksetattr
    ksetattr --> kbuffer
    kunlink --> kbuffer
    kbuffer --> ufilter
    ufilter --> ubuffer
    ubuffer --> udispatch
    ufilter@{ shape: diam}
```

## Challenges

1. **Accessing eBPF**:
   - Various approaches exist for eBPF integration, each with different levels of portability and complexity.
[...]

PR #28476, opened 2025-03-05
Whodata ebpf libsinsp fix integration test

lchico:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/27879|

## Description

Avoiding code builds in Windows that should only be required on Linux.
- The new configuration is being updated due to changes in Whodata. The eBPF driver will now be the default instead of Audit.
- Adding null pointer checks for the following:
  - logFn
  - abspathFn
- To create an abstract layer between ebpf_whodata and syscheck, I had to pass some variables as arguments:
  - is_fim_shutdown
  - syscheck
- CMakeLists can be fixed by adding the required libfimebpf dynamic libraries and target library.


jotacarma90 (approved):
LGTM

PR #28698, opened 2025-03-18
Fix TSAN warning messages

MiguelazoDS:
|Related issue|
|---|
|#28072|

## Description (WIP)

This PR adds the following changes: 

- Fix data race condition on threadEventDispatcher tests 
- Fix heap use after free on threadEventDispatcher tests
No TSAN errors 
image


GabrielEValenzuela (approved):
Good catch — using atomics here will definitely save us from future headaches and the UT is a great backup to validate the approach. Great work, Migue!

pereyra-m (approved):
LGTM!

PR #17982, opened 2023-07-20
Support to NPM and PYPI packages in the packages table.

Dwordcito:
|Related issue|
|---|
|Closes #5443 |

## Description

This PR supports getting, displaying, and synchronizing NPM and PYPI installed packages.

The support is only for the default paths of those packages' installation.

PR #28699, opened 2025-03-19
Add reflective JSON unit tests

### LINKED ISSUE #28421: Global queries - Technical debt - reflective JSON util
## Description

During the development of 
- https://github.com/wazuh/wazuh/issues/28347

It was found that `src/shared_modules/utils/reflectiveJson.hpp` requires UT.
This will increase the feature testing coverage and will help to find any hidden bug.

## DoD

- [x] Add UT for the class, considering many different cases: from real examples, corner cases, etc.
- [x] CI Pass

GabrielEValenzuela:
|Related issue|
|---|
|Closes #28421 |

# Summary
This PR enhances test coverage for the `serializeToJSON` and `jsonFieldToString` functions of `reflectiveJson` implementation, ensuring correctness in various edge cases. The tests validate serialization of primitive types, containers, nested structures, and escaping logic. Additionally, a specific test (`NumericBufferZeroing`) verifies that `std::fill(buffer, buffer + sizeof(buffer), '\0');` correctly resets the buffer during numeric conversions.

## Evidence

```console
└─$ ./utils_unit_test --gtest_filter="ReflectiveJsonTest*"
Note: Google Test filter = ReflectiveJsonTest*
[==========] Running 13 tests from 1 test suite.
[----------] 13 tests from ReflectiveJsonTest
[ RUN      ] ReflectiveJsonTest.EmptyMapReturnsEmptyBraces
[       OK ] ReflectiveJsonTest.EmptyMapReturnsEmptyBraces (0 ms)
[ RUN      ] ReflectiveJsonTest.SingleStringEntry
[       OK ] ReflectiveJsonTest.SingleStringEntry (0 ms)
[ RUN      ] ReflectiveJsonTest.MultipleStringEntries
[       OK ] ReflectiveJsonTest.MultipleStringEntries (0 ms)
[ RUN      ] ReflectiveJsonTest.NumericValues
[       OK ] ReflectiveJsonTest.NumericValues (0 ms)
[ RUN      ] ReflectiveJsonTest.NestedMap
[       OK ] ReflectiveJsonTest.NestedMap (0 ms)
[ RUN      ] ReflectiveJsonTest.EscapedStringValue
[       OK ] ReflectiveJsonTest.EscapedStringValue (0 ms)
[ RUN      ] ReflectiveJsonTest.BasicStructSerialization
[       OK ] ReflectiveJsonTest.BasicStructSerialization (0 ms)
[ RUN      ] ReflectiveJsonTest.NestedStructSerialization
[       OK ] ReflectiveJsonTest.NestedStructSerialization (0 ms)
[ RUN      ] ReflectiveJsonTest.BasicStructSerializationWithEmptyFields
[       OK ] ReflectiveJsonTest.BasicStructSerializationWithEmptyFields (0 ms)
[ RUN      ] ReflectiveJsonTest.BasicStructSerializationWithQuotes
[       OK ] ReflectiveJsonTest.BasicStructSerializationWithQuotes (0 ms)
[ RUN      ] ReflectiveJsonTest.BooleanValues
[       OK ] ReflectiveJsonTest.BooleanValues (0 ms)
[ RUN      ] ReflectiveJsonTest.SpecialCharactersInKeysAndValues
[       OK ] ReflectiveJsonTest.SpecialCharactersInKeysAndValues (0 ms)
[ RUN      ] ReflectiveJsonTest.NumericBufferZeroing
[       OK ] ReflectiveJsonTest.NumericBufferZeroing (0 ms)
[----------] 13 tests from ReflectiveJsonTest (0 ms total)

[==========] 13 tests from 1 test suite ran. (0 ms total)
[  PASSED  ] 13 tests.
```

### DIFF
--- a/src/shared_modules/utils/reflectiveJson.hpp
+++ b/src/shared_modules/utils/reflectiveJson.hpp
@@ -12,6 +12,7 @@
 #ifndef _REFLECTIVE_JSON_HPP
 #define _REFLECTIVE_JSON_HPP
 
+#include <array>
 #include <charconv>
 #include <map>
 #include <string>
@@ -20,6 +21,9 @@
 #include <type_traits>
 #include <unordered_map>
 #include <vector>
+
+constexpr auto CHAR_SIZE {256};
+constexpr auto BUFFER_SIZE {32};
 
 #define REFLECTABLE(...)                                                                                               \
     static constexpr auto fields()                                                                                     \
@@ -30,15 +34,15 @@
 #define MAKE_FIELD(keyLiteral, memberPtr)                                                                              \
     std::make_tuple(std::string_view {keyLiteral}, std::string_view {"\"" keyLiteral "\":"}, memberPtr)
 
-static std::array<const char*, 256> ESCAPE_TABLE = []
+static std::array<const char*, CHAR_SIZE> ESCAPE_TABLE = []
 {
-    std::array<const char*, 256> table {};
-    for (int i = 0; i < 256; ++i)
+    std::array<const char*, CHAR_SIZE> table {};
+    for (int i = 0; i < CHAR_SIZE; ++i)
     {
         table[i] = nullptr;
     }
 
-    table['"'] = R"("\"")";
+    table['"'] = R"(\")";
     table['\\'] = R"(\\)";
     table['\b'] = "\\b";
     table['\f'] = "\\f";
@@ -50,7 +54,7 @@ static std::array<const char*, 256> ESCAPE_TABLE = []
     {
         if (!table[i])
         {
-            static char buffer[256][7];
+            static char buffer[CHAR_SIZE][7];
             snprintf(buffer[i], 7, "\\u%04x", i);
             table[i] = buffer[i];
         }
@@ -174,7 +178,7 @@ std::string jsonFieldToString(const std::unordered_map<K, V>& map)
 {
     std::string json = "{";
     size_t count = 0;
-    char buffer[32] = {};
+    char buffer[BUFFER_SIZE] = {};
     for (const auto& [key, value] : map)
     {
         if (count++ > 0)
@@ -212,6 +216,7 @@ std::string jsonFieldToString(const std::unordered_map<K, V>& map)
             {
                 json.append("0");
             }
+            std::fill(buffer, buffer + sizeof(buffer), '\0');
         }
         else
         {
@@ -232,7 +237,7 @@ std::enable_if_t<IsReflectable<T>::value, void> serializeToJSON(const T& obj, st
 {
     json.push_back('{');
     constexpr auto fields = T::fields();
-    char buffer[32] = {};
+    char buffer[BUFFER_SIZE] = {};
 
     size_t count = 0;
     std::apply(
@@ -266,8 +271,8 @@ std::enable_if_t<IsReflectable<T>::value, void> serializeToJSON(const T& obj, st
                              json.push_back('"');
                          }
                          else if constexpr ((std::is_arithmetic_v<std::decay_t<decltype(data)>> ||
-                                             std::is_same_v<double, std::decay_t<decltype(data)>>)&&!std::
-                                                is_same_v<bool, std::decay_t<decltype(data)>>)
+                                             std::is_same_v<double, std::decay_t<decltype(data)>>) &&
+                                            !std::is_same_v<bool, std::decay_t<decltype(data)>>)
                          {
 
                              auto [ptr, ec] = std::to_chars(buffer, buffer + sizeof(buffer), data);
@@ -279,6 +284,7 @@ std::enable_if_t<IsReflectable<T>::value, void> serializeToJSON(const T& obj, st
                              {
                                  json.append("0");
                              }
+                             std::fill(buffer, buffer + sizeof(buffer), '\0');
                          }
                          else if constexpr (std::is_same_v<bool, std::decay_t<decltype(data)>>)
                          {
@@ -313,8 +319,8 @@ std::enable_if_t<IsReflectable<T>::value, void> serializeToJSON(const T& obj, st
                                      json.push_back('\"');
                                  }
                                  else if constexpr ((std::is_arithmetic_v<std::decay_t<decltype(value)>> ||
-                                                     std::is_same_v<double, std::decay_t<decltype(value)>>)&&!std::
-                                                        is_same_v<bool, std::decay_t<decltype(value)>>)
+                                                     std::is_same_v<double, std::decay_t<decltype(value)>>) &&
+                                                    !std::is_same_v<bool, std::decay_t<decltype(value)>>)
                                  {
                                      auto [ptr, ec] = std::to_chars(buffer, buffer + sizeof(buffer), value);
                                      if (ec == std::errc())
@@ -362,8 +368,8 @@ std::enable_if_t<IsReflectable<T>::value, void> serializeToJSON(const T& obj, st
                                      json.push_back('\"');
                                  }
                                  else if constexpr ((std::is_arithmetic_v<decltype(v)> ||
-                                                     std::is_same_v<const double&, decltype(v)>)&&!std::
-                                                        is_same_v<bool, std::decay_t<decltype(v)>>)
+                                                     std::is_same_v<const double&, decltype(v)>) &&
+                                                    !std::is_same_v<bool, std::decay_t<decltype(v)>>)
                                  {
                                      auto [ptr, ec] = std::to_chars(buffer, buffer + sizeof(buffer), v);
                                      if (ec == std::errc())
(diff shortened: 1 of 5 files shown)


src/shared_modules/utils/tests/CMakeLists.txt
```diff
@@ -2,6 +2,28 @@ cmake_minimum_required(VERSION 3.12.4)
...
+        char buffer[32];
+        float f = 3.14f;
+        auto [ptr, ec] = std::to_chars(buffer, buffer + sizeof(buffer), f);
+        return ec == std::errc() ? 0 : 1;
+    }
+" HAS_TO_CHARS_FLOAT)
+
+if(HAS_TO_CHARS_FLOAT)
+  add_definitions(-DHAS_TO_CHARS_FLOAT=true)
+  set(CMAKE_CXX_STANDARD 17)
```
pereyra-m, line 21: I think the standard is already defined at `src/shared_modules/utils/CMakeLists.txt`.
Also, isn't enough with this ?
```
if (${TARGET} STREQUAL "server")
```

src/shared_modules/utils/tests/reflectiveJson_test.cpp
pereyra-m, line 1: This file isn't present in the utils coverage report.

src/shared_modules/utils/tests/reflectiveJson_test.cpp
```diff
@@ -0,0 +1,218 @@
...
+    obj.fieldTwo = 0;
+    obj.fieldThree = "";
+
+    std::string json;
+    serializeToJSON(obj, json);
+
+    EXPECT_EQ(json.find(R"("fieldOne":)"), std::string::npos);
+    EXPECT_NE(json.find(R"("fieldTwo":0)"), std::string::npos);
+    EXPECT_EQ(json.find(R"("fieldThree":)"), std::string::npos);
+}
```
pereyra-m, line 143: Could we check the empty value here also?

src/shared_modules/utils/tests/reflectiveJson_test.cpp
```diff
@@ -0,0 +1,218 @@
...
+    serializeToJSON(obj, json);
+
+    EXPECT_NE(json.find(R"("fieldOne":"\"001\"")"), std::string::npos);
+    EXPECT_NE(json.find(R"("fieldTwo":30)"), std::string::npos);
+    EXPECT_NE(json.find(R"("fieldThree":true)"), std::string::npos);
+}
+
+TEST_F(ReflectiveJsonTest, SpecialCharactersInKeysAndValues)
+{
+    std::unordered_map<std::string, std::string> map {{"newline", "Line1\nLine2"},
```
pereyra-m, line 177: I would add more special characters to test the full ESCAPE_TABLE.
We are missing the non-printable characters that require UNICODE.

src/shared_modules/utils/tests/reflectiveJson_test.cpp
```diff
@@ -0,0 +1,218 @@
...
+                                                      {"tabbed", "A\tB\tC"},
+                                                      {"backslash", "Path\\To\\File"},
+                                                      {"quote", "\"Quoted text\""}};
+    std::string result = jsonFieldToString(map);
+
+    EXPECT_NE(result.find(R"("newline":"Line1\nLine2")"), std::string::npos);
```
pereyra-m, line 183: This approach of using find is due to the unpredictable order of the keys.
But we must validate that the output is a valid JSON and this is done only for simple cases.

For these complex cases, I suggest to add a validation with nlohmann for example.
This way we verify the final output can be parsed.

pereyra-m (changes requested):
Gj!

src/shared_modules/utils/tests/CMakeLists.txt
GabrielEValenzuela: When compiling with ASAN on Ubuntu 20.04 with the server target, the compilation fails. After researching the issue, I followed the approach defined [here](https://github.com/wazuh/wazuh/pull/17982/commits/4c84b17b87539a43537f78408de23792aa7e7437#r1287909903). Since we share utilities with the agent, perhaps we should conditionally compile this test only if the C++17 standard is available.

Please let me know if this approach aligns with your expectations or if you'd prefer we consider an alternative solution :)

src/shared_modules/utils/tests/reflectiveJson_test.cpp
GabrielEValenzuela: We tested using `EXPECT_EQ`, and according to the current implementation behavior, the value is not set if it is empty. However, we can discuss this during today's daily meeting and determine if a refactor is necessary.

src/shared_modules/utils/tests/reflectiveJson_test.cpp
GabrielEValenzuela: Good catch, I adding now

src/shared_modules/utils/tests/CMakeLists.txt
GabrielEValenzuela: This feature requires a C++17 standard

src/shared_modules/utils/tests/reflectiveJson_test.cpp
GabrielEValenzuela: I think we should check it on today's daily. I see the file being compiled and test executed, but, I don't know why coverage is not taking on consideration.

src/shared_modules/utils/tests/reflectiveJson_test.cpp
GabrielEValenzuela: Currently this is the output

```cpp
{"fieldTwo":0}
```

If the value is empty, is not added to JSON

src/shared_modules/utils/tests/reflectiveJson_test.cpp
GabrielEValenzuela: I think I found it

src/shared_modules/utils/tests/reflectiveJson_test.cpp
GabrielEValenzuela: I confirmed that this is the expected behavior, due indexer doesn't allow empty values :)

pereyra-m: > I confirmed that this is the expected behavior, due indexer doesn't allow empty values :)

This is interesting.
But @Dwordcito , should this JSON util behave like this?

src/shared_modules/utils/tests/reflectiveJson_test.cpp
pereyra-m: What was the issue?

GabrielEValenzuela: Layer 8 error 😅

src/shared_modules/utils/tests/reflectiveJson_test.cpp
Dwordcito: This utility is intended to have the highest possible performance, writing empty spaces implies writing keys and quotes, saving us from those two operations is an improvement. For our use case we don't mind the printing of voids yet, there could be a variant in the future when the need arises.

src/shared_modules/utils/tests/CMakeLists.txt
pereyra-m: I've tested a package from this PR, and I didn't find any problem to process the elements to index.

src/shared_modules/utils/tests/reflectiveJson_test.cpp
pereyra-m: I see this suggestion has been applied in some cases.

src/shared_modules/utils/tests/reflectiveJson_test.cpp
pereyra-m: Now the file appears.

src/shared_modules/utils/reflectiveJson.hpp
```diff
@@ -30,15 +34,15 @@
...
-    std::array<const char*, 256> table {};
-    for (int i = 0; i < 256; ++i)
+    std::array<const char*, CHAR_SIZE> table {};
+    for (int i = 0; i < CHAR_SIZE; ++i)
     {
         table[i] = nullptr;
     }
 
-    table['"'] = R"("\"")";
+    table['"'] = R"(\")";
```
pereyra-m, line 45: Good catch!

src/shared_modules/utils/reflectiveJson.hpp
```diff
@@ -21,6 +22,9 @@
 #include <unordered_map>
 #include <vector>
 
+constexpr auto CHAR_SIZE {256};
```
pereyra-m, line 25: +1 This avoids magic numbers.

pereyra-m (approved):

Issue #28690, opened 2025-03-05
Wazuh agent version sorting is not working correctly.


bonyjohn05:
|Wazuh version|Component|Install type|
|---|---|---|
| 4.11.0 | Manager | OVA 4.11.0 |

## Description
Wazuh agent sorting based on the agent version is not working correctly.

In the Wazuh Dashboard agent Summary tab, sorting agents by version does not work as expected. The sorting order is incorrect in both ascending and descending modes. I have attached a screenshot of the issue. 

![Image](https://github.com/user-attachments/assets/f27654b6-8211-408a-9a4e-6ac008d53606)

### Expected results
When sorting agents by version, the order should be:
Ascending : 4.9.2, 4.10.1, 4.11.0
Descending : 4.11.0, 4.10.1, 4.9.2

### Current results
Ascending : 4.9.2, 4.11.0, 4.10.1
Descending : 4.10.1, 4.11.0, 4.9.2

RamosFe:
## Update

After some research, the following code segment is in charge of returning:

https://github.com/wazuh/wazuh/blob/6ec412cf144795dbf4c013420c4658e6e9fb04ba/framework/wazuh/agent.py#L300-L354

I am testing in a new environment to determine what the database returns and whether we need to sort it before sending it to the client.

RamosFe:
## Update

After raising an environment with several agents, the following results were obtained:

<details>
<summary>Requests</summary>

```console
ramosfe@pop-os:~/Documents/Repositories/wazuh-tools/framework/environments/dev$ curl -k -X GET "https://localhost:55050/agents?select=version,id&pretty=true&sort=-version" -H "Authorization: Bearer $TOKEN"
{
   "data": {
      "affected_items": [
         {
            "id": "001",
            "version": "Wazuh v4.9.2"
         },
         {
            "id": "000",
            "version": "Wazuh v4.11.1"
         },
         {
            "id": "003",
            "version": "Wazuh v4.11.0"
         },
         {
            "id": "004",
            "version": "Wazuh v4.11.0"
         },
         {
            "id": "002",
            "version": "Wazuh v4.10.1"
         }
      ],
      "total_affected_items": 5,
      "total_failed_items": 0,
      "failed_items": []
   },
   "message": "All selected agents information was returned",
   "error": 0
}

---
[...]

PR #28723, opened 2025-03-20
Fix sort for agent versions

RamosFe:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/28690|

## Description

Fixes a bug where the agent versions aren't sorted correctly.

### DIFF
--- a/framework/wazuh/core/agent.py
+++ b/framework/wazuh/core/agent.py
@@ -108,6 +108,11 @@ class WazuhDBQueryAgents(WazuhDBQuery):
         if field == 'os.version':
             # Order by os major version and os minor version
             return "CAST(os_major AS INTEGER) {0}, CAST(os_minor AS INTEGER) {0}".format(self.sort['order'])
+        elif field == 'version':
+            return  ("CAST(substr(trim(version, 'Wazuh v'), 1, instr(trim(version, 'Wazuh v'), '.') - 1) as INTEGER) {0}, "
+                     "CAST(rtrim(substr(trim(version, 'Wazuh v'), instr(trim(version, 'Wazuh v'), '.') + 1, 2 ), '.') as INTEGER) {0}, "
+                     "CAST(REPLACE(version, 'Wazuh v'|| substr(trim(version, 'Wazuh v'), 1, instr(trim(version, 'Wazuh v'), '.') - 1) || '.' || rtrim(substr(trim(version, 'Wazuh v'), instr(trim(version, 'Wazuh v'), '.') + 1, 2 ), '.') || '.' , '') as INTEGER) {0}"
+                     ).format(self.sort['order'])
         return WazuhDBQuery._sort_query(self, field)
 
     def _add_search_to_query(self):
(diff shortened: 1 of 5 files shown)


framework/wazuh/core/agent.py
```diff
@@ -108,6 +108,11 @@ def _sort_query(self, field: str) -> str:
         if field == 'os.version':
             # Order by os major version and os minor version
             return "CAST(os_major AS INTEGER) {0}, CAST(os_minor AS INTEGER) {0}".format(self.sort['order'])
+        elif field == 'version':
```
GGP1, line 111: Rather than doing the string formatting operations in SQL, I would prefer we do them in Python, which allows for easier and simpler processing.

I think we should explore the possibility of fixing the version sorting in the `sort_array` function which is the one in charge of ordering the information after retrieving it from the database. Given that we might keep using it in 5.0, this would also allow us to fix any issue with it.

GGP1 (changes requested):
The requests shown in [this](https://github.com/wazuh/wazuh/issues/28690#issuecomment-2741271431) update sort by ascending and descending order, but both responses show the same result. Let's please review this.

framework/wazuh/core/agent.py
RamosFe: The changes were added to the class responsible for executing the query because I believe it is more efficient. By adding it to this class, we divided responsibilities, as it specializes in agent-related queries. Additionally, we delegated the sorting to SQL. While it doesn't make the code any clearer, I think it's more efficient for the database to handle the sorting rather than for us to do it after retrieving the information from the database.

I would keep the query as it is and maintain `sort_array` as a generic function, without adding specific functionality for the agent model.

GGP1: I would prioritize maintainability over efficiency in this case, as the performance gain is negligible.

Given that this code won't be part of 5.0, I will let the final reviewer decide on what to do.

GGP1 (changes requested):
We should add an AIT test case to validate that the sorting works as expected on the `version` field for both ascending and descending orders, and update the method's [unit test](https://github.com/wazuh/wazuh/blob/db7f0197fe995173bc75454ea0c7e0b582235f5c/framework/wazuh/core/tests/test_agent.py#L167).

GGP1 (dismissed):
LGTM

> Given that the API integration tests added are not executed automatically, please run them manually and upload the results.

framework/wazuh/core/agent.py
fdalmaup: Changes in how we query the DB are not trivial. We should take into account how the addition of multiple `CAST` operations would affect large environments where the global DB is constantly being handled with thousands of agents. We need to avoid adding more workload to the DB.
Check [EXPLAIN QUERY PLAN](https://sqlite.org/eqp.html) with the complete query made to the DB, determine how it could affect performance, and also explore the option mentioned by @GGP1.

fdalmaup (changes requested):
Also:
- Manually launch AITs once the changes are applied.

PR #28797, opened 2025-03-25
Remove JSON building in favor to use built-in types

Dwordcito:
|Related issue|
|---|
|Closes https://github.com/wazuh/devel-major-incidents/issues/15|

## Description

This PR aims to use the built-in types in favour the get better performance results in the usage.
In other side, if we get the current hash, we can check if after the calculation the number change, to decide if the update query will be processed.

### Results previous the change
2079149 / 246 -> 8451 ms
### Results after the change
427208 / 2197 -> 194 ms

### DIFF
--- a/src/wazuh_db/wdb.h
+++ b/src/wazuh_db/wdb.h
@@ -257,6 +257,7 @@ typedef enum wdb_stmt {
     WDB_STMT_GLOBAL_UPDATE_AGENT_INFO,
     WDB_STMT_GLOBAL_GET_GROUPS,
     WDB_STMT_GLOBAL_GET_AGENTS,
+    WDB_STMT_GLOBAL_GET_AGENTS_AND_GROUP,
     WDB_STMT_GLOBAL_GET_AGENTS_CONTEXT,
     WDB_STMT_GLOBAL_GET_AGENTS_BY_CONNECTION_STATUS,
     WDB_STMT_GLOBAL_GET_AGENTS_BY_CONNECTION_STATUS_AND_NODE,
@@ -1352,7 +1353,7 @@ int wdb_global_recalculate_agent_groups_hash(wdb_t* wdb, int agent_id, char* syn
  * @return WDBC_OK Success.
  *         WDBC_ERROR On error.
  */
-int wdb_global_recalculate_agent_groups_hash_without_sync_status(wdb_t* wdb, int agent_id);
+int wdb_global_recalculate_agent_groups_hash_without_sync_status(wdb_t* wdb, int agent_id, char * group);
 
 /**
  * @brief Function to recalculate the agent group hash for all agents.
--- a/src/wazuh_db/wdb_global.c
+++ b/src/wazuh_db/wdb_global.c
@@ -1058,17 +1058,45 @@ wdbc_result wdb_global_sync_agent_info_get(wdb_t *wdb, int* last_agent_id, char 
 }
 
 char* wdb_global_calculate_agent_group_csv(wdb_t *wdb, int id) {
-    cJSON* j_agent_groups = wdb_global_select_group_belong(wdb, id);
     char* result = NULL;
-    if (j_agent_groups) {
-        cJSON* j_group_name = NULL;
-        cJSON_ArrayForEach(j_group_name, j_agent_groups) {
-            wm_strcat(&result, cJSON_GetStringValue(j_group_name), MULTIGROUP_SEPARATOR);
-        }
-        cJSON_Delete(j_agent_groups);
+    sqlite3_stmt *stmt = NULL;
+
+    if (!wdb->transaction && wdb_begin2(wdb) < 0) {
+        mdebug1("Cannot begin transaction");
+        return NULL;
     }
-    else {
-        mdebug1("Unable to get groups of agent '%03d'", id);
+
+    if (wdb_stmt_cache(wdb, WDB_STMT_GLOBAL_SELECT_GROUP_BELONG) < 0) {
+        mdebug1("Cannot cache statement");
+        return NULL;
+    }
+
+    stmt = wdb->stmt[WDB_STMT_GLOBAL_SELECT_GROUP_BELONG];
+
+    if (sqlite3_bind_int(stmt, 1, id) != SQLITE_OK) {
+        merror("DB(%s) sqlite3_bind_int(): %s", wdb->id, sqlite3_errmsg(wdb->db));
+        return NULL;
+    }
+
+    int _status = SQLITE_ROW;
+
+    while ((_status = wdb_step(stmt)) == SQLITE_ROW) {
+        char * group_hash = (char *) sqlite3_column_text(stmt, 0);
+
+        if (group_hash == NULL) {
+            mdebug1("Group hash is NULL");
+            continue;
+        }
+
+        if (result != NULL && WDB_MAX_RESPONSE_SIZE < strlen(result) + strlen(group_hash) + 1) {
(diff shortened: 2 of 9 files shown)


src/wazuh_db/wdb_global.c
```diff
@@ -1058,17 +1058,45 @@ wdbc_result wdb_global_sync_agent_info_get(wdb_t *wdb, int* last_agent_id, char
...
+
+    if (sqlite3_bind_int(stmt, 1, id) != SQLITE_OK) {
+        merror("DB(%s) sqlite3_bind_int(): %s", wdb->id, sqlite3_errmsg(wdb->db));
+        return NULL;
+    }
+
+    int _status = SQLITE_ROW;
+
+    while ((_status = wdb_step(stmt)) == SQLITE_ROW) {
+        char * group_hash = (char *) sqlite3_column_text(stmt, 0);
```
MiguelazoDS, line 1084: Isn't more accurate group_name?

src/wazuh_db/wdb_global.c
```diff
@@ -1058,17 +1058,45 @@ wdbc_result wdb_global_sync_agent_info_get(wdb_t *wdb, int* last_agent_id, char
...
+
+        if (group_hash == NULL) {
+            mdebug1("Group hash is NULL");
```
pereyra-m, line 1087: Having NULL in a group name is an unrecoverable error, I wouldn't continue.

src/wazuh_db/wdb_global.c
```diff
@@ -1058,17 +1058,45 @@ wdbc_result wdb_global_sync_agent_info_get(wdb_t *wdb, int* last_agent_id, char
...
+            continue;
         }
-        cJSON_Delete(j_agent_groups);
+
+        if (result != NULL && WDB_MAX_RESPONSE_SIZE < strlen(result) + strlen(group_hash) + 1) {
+            mdebug1("The agent's groups exceed the socket maximum response size.");
```
pereyra-m, line 1092: In this case, I wouldn't return an incomplete group CSV string but fail completely.

src/wazuh_db/wdb_global.c
```diff
@@ -1058,17 +1058,45 @@ wdbc_result wdb_global_sync_agent_info_get(wdb_t *wdb, int* last_agent_id, char
...
+            break;
+        }
+        wm_strcat(&result, group_hash, MULTIGROUP_SEPARATOR);
     }
-    else {
-        mdebug1("Unable to get groups of agent '%03d'", id);
+
+    if (SQLITE_DONE != _status) {
+        mdebug1("SQL statement execution failed");
```
pereyra-m, line 1099: Same here, in case of error we must fail instead of return an incomplete string.

src/wazuh_db/wdb.h
```diff
@@ -1352,7 +1353,7 @@ int wdb_global_recalculate_agent_groups_hash(wdb_t* wdb, int agent_id, char* syn
  * @return WDBC_OK Success.
  *         WDBC_ERROR On error.
  */
-int wdb_global_recalculate_agent_groups_hash_without_sync_status(wdb_t* wdb, int agent_id);
+int wdb_global_recalculate_agent_groups_hash_without_sync_status(wdb_t* wdb, int agent_id, char * group);
```
pereyra-m, line 1356: We haven't updated this doxygen.

src/wazuh_db/wdb_global.c
```diff
@@ -1058,17 +1058,45 @@ wdbc_result wdb_global_sync_agent_info_get(wdb_t *wdb, int* last_agent_id, char
 }
 
 char* wdb_global_calculate_agent_group_csv(wdb_t *wdb, int id) {
-    cJSON* j_agent_groups = wdb_global_select_group_belong(wdb, id);
     char* result = NULL;
-    if (j_agent_groups) {
-        cJSON* j_group_name = NULL;
-        cJSON_ArrayForEach(j_group_name, j_agent_groups) {
```
pereyra-m, line 1065: +1: The JSON isn't required and this improves perfomance.

src/wazuh_db/wdb_global.c
```diff
@@ -1418,9 +1446,12 @@ int wdb_global_recalculate_agent_groups_hash_without_sync_status(wdb_t* wdb, int
         mdebug1("No groups in belongs table for agent '%03d'", agent_id);
     }
 
-    if (WDBC_ERROR == wdb_global_set_agent_group_hash(wdb, agent_id, agent_groups_csv, agent_groups_csv ? groups_hash : NULL)) {
-        result = WDBC_ERROR;
-        merror("There was an error assigning the groups hash to agent '%03d'", agent_id);
+    // if the previous group is different from the new one, we update the agent group context
```
pereyra-m, line 1449: Another improvement that can be considered is avoiding the hash calculation in case the groups haven't changed.

pereyra-m (commented):
Gj!
Just some comments to be discussed.

src/wazuh_db/wdb_global.c
MiguelazoDS: When I started the review I saw that but It does not seems important, then I focused just on UTs

src/wazuh_db/wdb_global.c
MiguelazoDS: I do not how often that would happen, I'm not sure if the logging message without exiting is or not on purpose

src/wazuh_db/wdb_global.c
MiguelazoDS: The same as above, I think the log messages is what we want for now, adding a pagination mechanism probably is out of scope.

src/wazuh_db/wdb_global.c
MiguelazoDS: Yeap, I think several checks are missing even before this implementation, this wdb code needs a deeper revision.

src/wazuh_db/wdb_global.c
MiguelazoDS: Yeap, you mean not only check if the wdb_global_calculate_agent_group_csv return a non-null value.

MiguelazoDS:
@Dwordcito if we do not want too much changes in this implementation since it was already tested in the failing environment, we can just dismiss the discussion above if you agree @pereyra-m

src/wazuh_db/wdb_global.c
pereyra-m: We don't have time, we can leave it.

src/wazuh_db/wdb_global.c
pereyra-m: Me neither. But it might not happen.
We haven't much time.

src/wazuh_db/wdb_global.c
pereyra-m: I agree about the error message, and the pagination isn't an option.
Just suggested to return.
But we don't have time.

src/wazuh_db/wdb_global.c
pereyra-m: I agree.
We'll leave like it is because we don't have time.

src/wazuh_db/wdb.h
pereyra-m: It isn't critical.
We don't have time.

src/wazuh_db/wdb_global.c
pereyra-m: We can leave it for another moment.
We don't have time.

pereyra-m (approved):

Issue #28804, opened 2025-03-26
Update DELETE /agent purge option description


GGP1:
## Description

The `DELETE /agents` endpoint has a parameter called `purge` that indicates whether the agent ID should be removed from local storage (true) or simply marked as deleted (false). 

However, this parameter is no-op when the wazuh-authd [purge configuration option](https://documentation.wazuh.com/current/user-manual/reference/ossec-conf/auth.html#purge) is set to `yes` because it has precedence over the parameter.

We should add a comment in the parameter description to make this point clear.

## Checks

The following elements have been updated or reviewed (should also be checked if no modification is required):
- [x] Tests (unit tests, API integration tests).
- [x] Changelog.
- [x] Documentation.
- [x] Integration test mapping (using `api/test/integration/mapping/_test_mapping.py`).

GGP1:
## Update

Updated the `purge` field description to clarify that the wazuh-auth configuration takes precedence over it.

GGP1:
## Update

Answered a suggestion to add a documentation link to the API specification.

PR #28805, opened 2025-03-26
Update purge param description

GGP1:
|Related issue|
|---|
| Closes https://github.com/wazuh/wazuh/issues/28804 |

## Description

Updates the `DELETE /agents` `purge` parameter description to clarify that the option from wazuh-authd has precedence over its value.

### DIFF
--- a/api/api/spec/spec.yaml
+++ b/api/api/spec/spec.yaml
@@ -6084,7 +6084,9 @@ components:
     purge:
       in: query
       name: purge
-      description: "Permanently delete an agent from the key store"
+      description: |
+        Permanently delete an agent from the key store. The wazuh-authd `purge` configuration option has 
+        precedence over this parameter, when enabled, agents will be permanently deleted regardless of this value.
       schema:
         type: boolean
         default: false


api/api/spec/spec.yaml
```diff
@@ -6084,7 +6084,9 @@ components:
     purge:
       in: query
       name: purge
-      description: "Permanently delete an agent from the key store"
+      description: |
+        Permanently delete an agent from the key store. The wazuh-authd `purge` configuration option has
```
jnasselle, line 6088: How about linking to https://documentation.wazuh.com/4.13/user-manual/reference/ossec-conf/auth.html#purge ?

If not, my suggestion is  making reference of `wazuh-authd `purge` configuration` as `The `purge` option, part of `auth` configuration block`

GGP1: Adding a documentation link would require us to maintain it on every realese, if the URL or the page content changes, we would have to update the spec. That's why I didn't reference it directly.

Regarding the other suggestion, I didn't understand it. Is it mentioning the specific place where that configuration is?

jnasselle (approved):

fdalmaup (approved):
LGTM

Issue #28932, opened 2025-04-01
RPM5 agent fails with a compilation error


rauldpm:
## Description

The RPM5 i386 agent is failing due to a eBPF error in the code compilation stage

```
 /build_wazuh/rpmbuild/BUILD/wazuh-agent-4.12.0/src/syscheckd/src/ebpf/src/ebpf_whodata.cpp:14:10: fatal error: modern.skel.h: No such file or directory
   14 | #include "modern.skel.h"
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```

https://github.com/wazuh/wazuh-agent-packages/actions/runs/14197728759/job/39776835344

PR #28938, opened 2025-04-01
Prevent compilation of eBPF for CentOS 5

MarcelKemp:
|Related issue|
|---|
|#28932|

## Description

Because we do not support eBPF for CentOS 5 (as the kernel version is <5.8), with this change we prevent eBPF from being compiled, so it does not break the FIM compilation and the package can be generated correctly.

## Tests

- [x] https://github.com/wazuh/wazuh-agent-packages/actions/runs/14199085076 - :green_circle: 
- [x] https://github.com/wazuh/wazuh-agent-packages/actions/runs/14199117080 - :green_circle:


jotacarma90 (approved):
LGTM

Issue #28911, opened 2025-03-31
Engine - Automate API documentation generation in mdBook


JcabreraC:
## Description
We need to automate the generation of Wazuh-Engine API documentation in mdBook, aligning it with the same format used by the @wazuh/devel-xdrsiem-server-div2  team. Currently, the API documentation is included in mdBook, but lacks a standardized process for automatic creation and maintenance.

## Tasks
1. **Standardize format**  
   - Adopt the same mdBook layout and style guidelines that the Framework team uses for API docs (developed at #28620).
2. **Automation**  
   - Implement a script or workflow to generate API documentation automatically, ensuring updates are reflected in mdBook whenever the API changes.
3. **Verification**  
   - Confirm that the final output matches the  @wazuh/devel-xdrsiem-server-div2  team's style, with consistent formatting, navigation, and indexing.

## Expected Outcome
A unified, automated documentation process that keeps the Wazuh-Engine API docs consistent with the rest of the project's mdBook-based documentation, improving maintainability and readability.

JavierBejMen:
## Update
- Added event stateless, catalog and policy to spec.yaml format (pending to add detailed examples and improve descriptions)
- Researching and learning spec.yaml schema

JavierBejMen:
## Update
- Made added endpoints documentation to be compliant with OpenApi 3 spec schema.
- Researching api_client automatic generation with OpenApi, tldr: could not get around the client to comunicate with http over unix sockets, but generator worked correctly.

JavierBejMen:
## Update
- Added kvdb, router and tester spec docs

JavierBejMen:
## Update
- Added geo spec

JcabreraC:
ETA extended by priority in urgent E2E testing tasks.

JcabreraC:
We need the API examples we had before to be included in this new documentation.
Also, we must ensure they are compliant with OpenAPI 3.

@JavierBejMen  currently handling urgent E2E test issues.

@NahuFigueroa97 will take this issue to finish it.

NahuFigueroa97:
## Update

- Add all errors examples in endpoints

PR #28948, opened 2025-04-02
Added Engine API documentation as OpenApi spec file

JavierBejMen:
|Related issue|
|---|
|#28911|


NahuFigueroa97 (changes requested):
Really great work! Just one thing about the names. I think it should be clear which ones have names with parts and which don't. For example, the catalog's validate endpoint makes it clear, but the policy endpoints don't. Review this.

juliancnn (approved):
LGTM!

image

JcabreraC (approved):
LGTM !

PR #29181, opened 2025-04-11
Handle case when version is missing

### LINKED ISSUE #29102: Vulnerability detector: handle missing agent version from Wazuh-DB
## Description 

During some local tests, it was found that are some cases when the VD module can generate the following error message
```
2025/04/04 23:02:14 wazuh-modulesd:vulnerability-scanner[3748866] buildAllAgentListContext.hpp:98 at handleRequest(): DEBUG: Error reading global agent data: [json.exception.out_of_range.403] key 'version' not found
```
In this section of the code

https://github.com/wazuh/wazuh/blob/87af8e2ee6ec8baefaeaca7a92f0bcffb17268a2/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/buildAllAgentListContext.hpp#L87

This happens when the `global.db` doesn't contain the agent version (the first keep alive hasn't arrived yet for example) because if the value isn't found, it isn't included in the reply

https://github.com/wazuh/wazuh/blob/87af8e2ee6ec8baefaeaca7a92f0bcffb17268a2/src/wazuh_db/wdb.c#L1157-L1159

## DoD

- [x] Implement the changes in VD to handle this situation
- [x] UT
- [x] CI Pass

pereyra-m:
|Related issue|
|---|
|Closes #29102|

## Description

This PR handles the case when Wazuh-DB doesn't send the agent version.
This can happen if the agent is registered but didn't send the first keepalive.

### DIFF
--- a/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/buildAllAgentListContext.hpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/buildAllAgentListContext.hpp
@@ -82,10 +82,11 @@ public:
                 // If the agent is the manager and the manager scan is disabled, skip it
                 if (!(isManagerScanDisabled && agent.at("id").get<int>() == 0))
                 {
-                    data->m_agents.push_back({Utils::padString(std::to_string(agent.at("id").get<int>()), '0', 3),
-                                              agent.at("name"),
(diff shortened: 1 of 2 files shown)


GabrielEValenzuela (approved):
The fix looks solid — I rechecked with the unit test case, and it confirms the root cause is now properly addressed. Great job, Mati!

Issue #29095, opened 2025-04-07
Error with dispatch for IT - API Tier 2


RamosFe:
## Description

The following error appears when trying to run the API Tier 2 Github Action:

```console
 gh workflow run .github/workflows/4_testintegration_api-tier-2.yml --ref v4.12.0-alpha1 -f base_branch=v4.12.0-alpha1
HTTP 404: workflow .github/workflows/4_testintegration_api-tier-2.yml not found on the default branch (https://api.github.com/repos/wazuh/wazuh/actions/workflows/.github%2Fworkflows%2F4_testintegration_api-tier-2.yml)

```

We should investigate why this happens and propose the necessary changes to fix it.

jnasselle:
# RCA

This is caused by not having those workflow files at the default branch (`main`). We must ensure that 4_x workflows exist (just a facade, not a functional one because them will not work in other branch rather than 4.x) there

GGP1:
## Update

Added the missing 4.x workflows to the main branch except for `.github/workflows/4_builderpackage_manager-multiples-2.yml`, which was removed in https://github.com/wazuh/wazuh/issues/28972 because it is triggered in 5.0 PRs unnecessarily.

Before submitting the PR, I'm checking if there is an alternative way of executing the tests in the UI without having the files repeated in the main branch.

GGP1:
## Update

The only workaround that could be implemented to have 4.x workflows in the UI without having the files in the branch would be:

1. Add the files to the main branch
2. Execute all workflows at least once
3. Remove them from the branch
4. At this point you can already run them using the `gh workflow run` command, but if you want to them in a UI, that would require creating a custom one based on the `gh workflows list` options.

In summary, it is possible but the amount of work required is probably not worth it. So we will just leave all files in the main branch.

GGP1:
## Update

Applied suggested changes.

fdalmaup:
The workflow can now be correctly run:
```
$ gh workflow run 4_testintegration_api-tier-2.yml --repo=wazuh/wazuh --ref v4.12.0-alpha1 base_branch=v4.12.0-alpha1
✓ Created workflow_dispatch event for 4_testintegration_api-tier-2.yml at v4.12.0-alpha1

To see runs for this workflow, try: gh run list --workflow=4_testintegration_api-tier-2.yml
```

Issue #28972, opened 2025-04-03
Disable `4.x RPM/DEB build` workflow for the main branch.


juliancnn:
## Description 

Currently the 5.x pull requests that point to the `main` branch are attempting to build the 4.x package as can be seen in the following code snippet:

https://github.com/wazuh/wazuh/blob/3828faf0d966e4bc9324cc8e27b73b0768a293c2/.github/workflows/4_builderpackage_manager-multiples-2.yml#L5-L11

For example https://github.com/wazuh/wazuh/actions/runs/14248411503/attempts/1?pr=28866

The workflow was added as part of the workflows remaining carried out in #28225 to let it be manually triggered. It should be removed to avoid unnecessary runs.

PR #29200, opened 2025-04-11
Add missing 4.x workflows

GGP1:
|Related issue|
|---|
| Closes https://github.com/wazuh/wazuh/issues/29095 |

## Description

Adds missing 4.x workflows to be able to run them from the actions UI. 

They are just a facade, and are not expected to pass in the main branch, since they were developed to work in 4.x.

> [!Note]
> `.github/workflows/4_builderpackage_manager-multiples-2.yml` wasn't added, the team removed it in https://github.com/wazuh/wazuh/issues/28972 because it is triggered in 5.0 PRs unnecessarily.


.github/workflows/4_codeanalysis_python.yml
nico-stefani, line 1: We should keep this one as it is. The **Code scanning alerts ** depend on that.

nico-stefani (changes requested):

.github/workflows/4_codeanalysis_python.yml
GGP1: Oh, okay. It could even be removed from the 4.x branches because it is pointing to main anyway.

Just restored its name 👍🏼

nico-stefani (approved):
LGTM

jnasselle:
Hi @GGP1 ,

Facade 4_x workflows on main or 5.x branches should not include on: pull_request trigger because it will cause its execution on branches that are not compatible with them.

My suggestion is to evaluate just adding them, describing the workflow_dispatch and a fixed step failing always. This way will fail if any 4_x workflow is executed at the current branch

PR #29221, opened 2025-04-14
Fix Microsoft windows 11 enterprise benchmark v3.0.0 SCA check ID 26009 issue

IsExec:
|Related issue|
|---|
|https://github.com/wazuh/external-devel-requests/issues/5648|

# Description

A customer reported  customer having issues with SCA ID 46009. The check fails due to apparent contradictory conditions that should be met.

- The correct check is meant to be:

```console
    condition: any
    rules:
      - 'c:net user guest -> r:Account active\s+No'
      - 'c:net user guest -> r:The user name could not be found.'
```

- Other affected OS - Windows server 2012 R@ and windows 10


ooniagbi (approved):
LGTM!

Issue #29169, opened 2025-04-10
Review the Global queries related documentation


fdalmaup:
# Description

Throughout the development of the new Global queries feature, technical documentation has been developed to describe the libraries, components, and modules that compose the changes to be added, expanding on the Inventory Harvester, the main module designed for this new functionality to be carried out. This issue aims to review the documentation merged into the objective's development branch (`dev/fim-syscollector-global-queries`) and make the necessary improvements to it.

## Tasks

- [x] Review the Global queries documentation.
- [x] Determine key points to be enhanced/expanded and missing descriptions.
- [x] Carry out the necessary enhancements/changes.

PR #29231, opened 2025-04-14
Updated Inventory Harvester documentation

MiguelazoDS:
|Related issue|
|---|
|#29169|

## Description

Documentation changes related to global queries implementation


docs/ref/modules/inventory-harvester/api-reference.md
pereyra-m, line 1: +1 Really useful

pereyra-m (commented):
Great job!

It takes time to gather all the examples.

I think the current documentation covers the basic module behavior and general details. And it could be improved ad infinitum if we want.

Just to mention some extra stuff that could or not be included according to the issue ETA:
- Upgrade details: briefly mention that the module requires a full re-sync after the upgrade
- Configuration: mention that the module doesn't have an specific block, and that it depends on the indexer connector module for working
- Agent 000: by design, the manager isn't included and we should document that

docs/ref/modules/inventory-harvester/api-reference.md
MiguelazoDS: > Upgrade details: briefly mention that the module requires a full re-sync after the upgrade 

I modified the comment in the architecture.md file about the `upgradeAgentDb` component

> Configuration: mention that the module doesn't have an specific block, and that it depends on the indexer connector module for working

Added a new file configuration.md 

> Agent 000: by design, the manager isn't included and we should document that

Added a new bullet in the main readme

MiguelazoDS: I added a high level diagram, please take a look

pereyra-m (approved):
Gj!

The diagram represents the current architecture, and the suggestions were added (the mention to the configuration is clear and detailed).

Issue #29601, opened 2025-05-12
Fix agents status synchronization race condition


GGP1:
## Description

During the testing of the performance enhancements that were introduced recently, we noticed a race condition causing an agent's status to remain disconnected even though it is actually connected.

This is because the sync of the worker where the agent connected to is performed before the sync of the worker where the agent disconnected from, causing the final status to be `disconnected` and incorrectly representing the current agent status.

This can be grasped with the following example:

1. An agent connects to worker 2
2. Worker 2 sends the sync information to the master (wazuh-db master marks it as active)
3. The agent disconnects from worker 2 and connects to worker 1
4. In this step, there is a race to see who sends the sync to the master first. Since there is no guarantee, it may happen that worker 1 sends its sync first (the agent is marked as active on the master) and then worker 2 does so (the agent is marked as disconnected on the master). This is where the errors we see in the tests occur.
5. The master remains with the agent disconnected even though it is marked as `active` on worker 1. Worker 1 does not send the sync again because it already did so when the agent connected. The status remains `disconnected` until the agent is restarted and a new synchronization is performed.

This condition isn't new; it has existed before, but it was masked by the fact that each keep-alive would update the agent's status, leaving it in a state of disconnection for a few seconds (until the next keep-alive).

We need to figure out and apply a solution to this issue to be able to continue testing the improvements we made.

PR #29624, opened 2025-05-12
Fix worker-master race condition.

Dwordcito:
#### Root cause (recap)
| Step | Event | Effect on `wazuh-db` (master) |
|------|-------|--------------------------------|
| 1 | Agent connects to **worker-2** | status ➜ **active** |
| 2 | Agent disconnects from **worker-2** and immediately connects to **worker-1** | — |
| 3 | **worker-1** syncs first | status ➜ **active** |
| 4 | **worker-2** syncs afterwards | status ➜ **disconnected** (stale) |

Because we accept the **last** sync packet as authoritative, an “older” *disconnect* can overwrite a newer *connect* when they arrive out of order.  [oai_citation:0‡GitHub](https://github.com/wazuh/wazuh/issues/29601?utm_source=chatgpt.com)

## Description
These changes cause the value to be adjusted to the actual data in the subsequent update.

### DIFF
--- a/src/shared_modules/router/src/wazuh-db/endpointPostV1AgentsSync.hpp
+++ b/src/shared_modules/router/src/wazuh-db/endpointPostV1AgentsSync.hpp
@@ -103,9 +103,10 @@ public:
         {
             if (jsonBody.contains("syncreq_keepalive"))
             {
-                DBStatement stmt(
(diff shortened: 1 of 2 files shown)


juliancnn (approved):
LGTM!

Issue #29660, opened 2025-05-14
Integration tests Remoted tier 2 failed - Failed


JcabreraC:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| v4.10.2 | Integration test | Manager | Packages/Sources | Windows |

## Description

During test [Week 20 - Integration/Component test manager](https://github.com/wazuh/wazuh/issues/25537)](https://github.com/wazuh/wazuh/issues/29658), it was found that the test case for remoted Tier 2 forfails constantly.

- Test [link](https://github.com/wazuh/wazuh/actions/workflows/integration-tests-remoted-tier-2.yml)
- Result [link](https://github.com/wazuh/wazuh/actions/runs/15016769364/attempts/1)

<details><summary>Error</summary>

```
------------------------------ Captured log setup ------------------------------
DEBUG    wazuh_testing:conftest.py:329 Wazuh control set to True
DEBUG    wazuh_testing:conftest.py:341 Restarting wazuh using wazuh-control
___ ERROR at setup of test_shared_configuration[test shared configuration2] ____

test_metadata = {'description': 'test shared configuration', 'name': 'test shared configuration', 'patterns': ['.*up file.*merged.mg.*', '.*default.*', '.*close file.*'], 'protocol': 'TCP'}

    @pytest.fixture()
    def simulate_agents(test_metadata):
    
        agents_amount = test_metadata.get("agents_number", 1)
>       agents = create_agents(agents_amount , 'localhost')

conftest.py:755: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/local/lib/python3.10/dist-packages/wazuh_testing/tools/simulators/agent_simulator.py:1780: in create_agents
    agents.append(Agent(manager_address, cypher, fim_eps=fim_eps, authd_password=authd_password,
/usr/local/lib/python3.10/dist-packages/wazuh_testing/tools/simulators/agent_simulator.py:326: in __init__
    self.setup(disable_all_modules=disable_all_modules)
/usr/local/lib/python3.10/dist-packages/wazuh_testing/tools/simulators/agent_simulator.py:339: in setup
    self.register()
/usr/local/lib/python3.10/dist-packages/wazuh_testing/tools/simulators/agent_simulator.py:419: in register
    self._register_helper()
[...]

PR #29661, opened 2025-05-14
Fix integration test for remoted tier 2

JcabreraC:
|Related issue|
|---|
|#29660|

## Description

The objective of this pull request is to fix the bug in the remoted integration test with tear 2. The problem was that when setting the environment variables for the manager installation, the auhtd was disabled, which made it impossible for the manager to connect to the agent.

Test: https://github.com/wazuh/wazuh/actions/runs/15019998572


fdalmaup (approved):
LGTM

Issue #29608, opened 2025-05-12
Coverity execution in `4.13.0`


GabrielEValenzuela:
# Description

This issue aims to execute Coverity Scan, a static analysis tool, on the `4.13.0` implementation to identify potential defects and vulnerabilities in the codebase. The goal is to ensure the overall code quality and security of the new GQ module. Additionally, any issues detected during the Coverity Scan execution will be addressed and resolved by the development team.

## Addressing Identified Issues

After the Coverity Scan execution, carefully review the generated reports to identify the specific issues found in the new minor version. 

#### DoD - Tasks
- [x] Fixing reported problems
- [ ] Code Review / CI Pass.

MiguelazoDS:
# Update 

I'm having issues launching the workflow. Asking for information.

MiguelazoDS:
# Update 

Execution completed [analysis](https://scan.coverity.com/projects/wazuh-ossec-wazuh?tab=overview)

pereyra-m:
HI @MiguelazoDS 

> From my point of view we can move these variables since their scope is inside the startWorker method. But they are suppressed for a reason that I should discuss with the team.

I'm not sure why the Coverity comment was added.

MiguelazoDS:
yeah @pereyra-m,  we are also suppressing the same error in the vulnerabilityScannerFacade, so there must be a particular case that I'm not considering and moving may not be correct.

PR #29671, opened 2025-05-14
Fix use of npos in substr

MiguelazoDS:
|Related issue|
|---|
|#29608|

## Description
This prevent this case but it will never happen since the file always contains "/" 

```console
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::substr: __pos (which is 18446744073709551615) > this->size() (which is 13)
[1]    13953 IOT instruction (core dumped)  ./a.out
```


pereyra-m (approved):
LGTM!

GabrielEValenzuela (approved):
Good catch!

Issue #29546, opened 2025-05-07
Wazuh-Analisysd - Event queue size rc


juliancnn:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.x | Wazuh-Analisysd | Manager | Any | Any |

When the decoder threads receive the messages, they check if there is space in the queue to push the events. The calculation is done on the fly and your internal variables are not protected in this method. I think the solution may be related to issue https://github.com/wazuh/wazuh/issues/29544.

```

==================
WARNING: ThreadSanitizer: data race (pid=97969)
  Write of size 8 at 0x7b2c00019500 by thread T173 (mutexes: write M300187754082440464):
    #0 queue_pop shared/queue_op.c:100 (wazuh-analysisd+0x1892a1)
    #1 queue_pop_ex shared/queue_op.c:111 (wazuh-analysisd+0x1893d4)
    #2 w_decode_event_thread analysisd/analysisd.c:1814 (wazuh-analysisd+0x690ff)
    #3 <null> <null> (libtsan.so.0+0x2e5ff)

  Previous read of size 8 at 0x7b2c00019500 by thread T5 (mutexes: read M301272050075310240):
    #0 queue_full shared/queue_op.c:37 (wazuh-analysisd+0x188bd6)
    #1 ad_input_main analysisd/analysisd.c:1438 (wazuh-analysisd+0x67917)
    #2 <null> <null> (libtsan.so.0+0x2e5ff)

  Location is heap block of size 176 at 0x7b2c000194f0 allocated by main thread:
    #0 <null> <null> (libtsan.so.0+0x31edc)
    #1 queue_init shared/queue_op.c:16 (wazuh-analysisd+0x1887b4)
    #2 w_init_queues analysisd/analysisd.c:2458 (wazuh-analysisd+0x6c5bf)
    #3 main analysisd/analysisd.c:879 (wazuh-analysisd+0x65035)

  Mutex M300187754082440464 is already destroyed.

  Mutex M301272050075310240 is already destroyed.

  Thread T173 (tid=98248, running) created by main thread at:
    #0 <null> <null> (libtsan.so.0+0x605b8)
    #1 CreateThreadJoinable shared/pthreads_op.c:47 (wazuh-analysisd+0x187732)
    #2 CreateThread shared/pthreads_op.c:62 (wazuh-analysisd+0x1877fb)
[...]

PR #29672, opened 2025-05-14
Race condition solved in the event queue

NahuFigueroa97:
|Related issue|
|---|
|#29546|

## Description

This PR addresses a data race condition detected by ThreadSanitizer when checking if there is space in the event queue used by decoder threads in analysisd.

The root cause of the issue was a concurrent read-write access to internal queue variables (begin, end, size) without proper synchronization. The queue_full() function was being called in ad_input_main() without holding the queue mutex, while queue_pop() (which modifies the same internal state) was being called in w_decode_event_thread() under a mutex lock.

##  Changes introduced

- Added a new function queue_full_ex() which wraps queue_full() in a critical section (mutex-protected).

- Replaced direct calls to queue_full() in ad_input_main() with the thread-safe queue_full_ex() version.

- Added unit test to ensure the correctness of queue_is_full_safe() and its synchronization behavior.

### DIFF
--- a/src/analysisd/analysisd.c
+++ b/src/analysisd/analysisd.c
@@ -1229,7 +1229,7 @@ void * ad_input_main(void * args) {
             w_rwlock_rdlock(&g_hotreload_ruleset_mutex);
 
             if (msg[0] == SYSCHECK_MQ) {
-                if (!queue_full(decode_queue_syscheck_input)) {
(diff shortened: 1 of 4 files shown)


juliancnn (approved):
LGTM!

juliancnn (approved):
LGTM!

Dwordcito (approved):

PR #29796, opened 2025-05-19
Fix VERSION file read

### LINKED ISSUE #29654: Update version and revision file paths in Coverity workflow
# Description 

It was found that the Coverity workflow is looking for files that were deprecated. 

Image

Currently that information is provided by the `VERSION.json` file. 

## DoD 

- [ ] Fix the failing step.

pereyra-m:
|Related issue|
|---|
|Closes #29654 |

## Description

This PR fixes the read of the VERSION file for the Coverity script.

### DIFF
--- a/.github/workflows/4_codeanalysis_coverity.yml
+++ b/.github/workflows/4_codeanalysis_coverity.yml
@@ -84,20 +84,23 @@ jobs:
         working-directory: wazuh/src
         id: build
         run: |
-          VERSION="$(cat VERSION)-r$(cat REVISION)"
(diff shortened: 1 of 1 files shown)


MiguelazoDS (commented):
Just a question, 

This is a test with the changes


This is how previously worked


That version is used then to upload the result to coverity and set the version in the UI. 

is it OK to not include the revision when we export the version?

pereyra-m:
Hi @MiguelazoDS 

I think you are right, I'm changing the original approach.
I'll try to keep the old behavior.

Issue #29790, opened 2025-05-19
`wazuh-modulesd` doesn't initialize in Solaris 11


hossam1522:
### Description

While doing manual tests in https://github.com/wazuh/wazuh/issues/29731, it was discovered that `wazuh-modulesd` doesn't initialize in the Solaris 11 agent. 

> [!IMPORTANT]
> This bug has been detected in 4.10.0, 4.10.1 and 4.10.2. In previous versions like 4.9.2, it seems to be working without issues.

```console
root@sossp585:~# /var/ossec/bin/wazuh-control start
Starting Wazuh v4.10.2...
wazuh-execd already running...
wazuh-agentd already running...
wazuh-syscheckd already running...
wazuh-logcollector already running...
Started wazuh-modulesd...
Completed.
root@sossp585:~# /var/ossec/bin/wazuh-control status
wazuh-modulesd: Process 22539 not used by Wazuh, removing ..
wazuh-modulesd not running...
wazuh-logcollector is running...
wazuh-syscheckd is running...
wazuh-agentd is running...
wazuh-execd is running...
```

### Environment

A Wazuh AMI and a Solaris 11 agent.

### Steps to reproduce

<details><summary>1. Download and install the 4.10.2 Solaris 11 package</summary>

```console
root@sossp585:~# uname -a
SunOS sossp585 5.11 11.3 sun4v sparc sun4v
root@sossp585:~# curl -sO https://packages-dev.wazuh.com/pre-release/solaris/sparc/11/wazuh-agent_v4.10.2-sol11-sparc.p5p
root@sossp585:~# pkg install -g wazuh-agent_v4.10.2-sol11-sparc.p5p wazuh-agent
           Packages to install:  1
            Services to change:  1
       Create boot environment: No
Create backup boot environment: No

DOWNLOAD                                PKGS         FILES    XFER (MB)   SPEED
Completed                                1/1       126/126      6.6/6.6 26.1M/s

PHASE                                          ITEMS
Installing new actions                       184/184
Updating package state database                 Done 
Updating package cache                           0/0 
Updating image state                            Done 
Creating fast lookup database                   Done 
Updating package cache                           2/2
``` 

</details>

<details><summary>2. Connect the instance to the AMI manager, start the agent, and check the status</summary>

```console
root@sossp585:~# nano /var/ossec/etc/ossec.conf
root@sossp585:~# grep address /var/ossec/etc/ossec.conf
      <address>xx.xx.xx.xx</address>
[...]

PR #29863, opened 2025-05-22
Fix environment variables to prevent freezing

MarcelKemp:
|Related issue|
|---|
|#29790|

## Description

This PR fixes the environment variable issue, which affected the first launch of a Solaris 11 SPARC environment (on provisioning): 
- Always at the first launch of a Solaris 11 SPARC environment (in provisioning), because it would freeze when using the system GCC.
- And when installing a Wazuh agent package, because of the same problem (using the system GCC libraries).

## Tests

- [x] [Packages - Build Wazuh agent special package - System solaris11 - check](https://github.com/wazuh/wazuh-agent-packages/actions/runs/15189433415)
- [x] [Packages - Build Wazuh agent special package - System solaris11 - checks](https://github.com/wazuh/wazuh-agent-packages/actions/runs/15190179758)

### DIFF
--- a/packages/solaris/solaris11/generate_wazuh_packages.sh
+++ b/packages/solaris/solaris11/generate_wazuh_packages.sh
@@ -4,6 +4,7 @@
 # This program is a free software; you can redistribute it and/or modify it under the terms of GPLv2
 # Wazuh Solaris 11 Package builder.
 
+set -x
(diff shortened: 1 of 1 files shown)


vikman90 (approved):
- Changes make sense
- Tests pass

LGTM!

mjcr99 (approved):

PR #29953, opened 2025-05-29
Index inode for files

### LINKED ISSUE #29951: Global queries: index inode for FIM files entries
## Description

Currently, the inventory harvester isn't indexing the inode of the files.
We should add this field.

## DoD

- [x] Index requested field
- [x] Add tests
- [x] CI Pass

pereyra-m:
|Related issue|
|---|
|Closes #29951|

## Description

This PR adds the `inode` field for files.
It can't be added as `std::string_view` because the origin uses int64 and the indexed value is a string.

## Tests

WIP

### DIFF
--- a/src/wazuh_modules/inventory_harvester/qa/test_data/fim/file/000_insert_delta/result.json
+++ b/src/wazuh_modules/inventory_harvester/qa/test_data/fim/file/000_insert_delta/result.json
@@ -20,7 +20,8 @@
                     "owner": "root",
                     "path": "/home/folderB/AFileInB.txt",
                     "size": 0,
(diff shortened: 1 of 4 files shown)


GabrielEValenzuela (commented):
## Testing

`inode` field now is indexed correctly. We should check QA tests

image

GabrielEValenzuela (approved):
The QA test fails due this PR https://github.com/wazuh/wazuh/pull/29910

GJ Mati!

pereyra-m:
@MiguelazoDS 

I tried to use the template as the source of truth.
But we can check if this is what we want.

Issue #29618, opened 2025-05-12
Rocky 9 SCA issue false positives


fuckingdarkknight:
Hello,

based on regexp, some rules generate false positives.

Eg. "c:modprobe -n -v squashfs -> r:install /bin/false|Module squashfs not found"
=>
"c:modprobe -n -v squashfs -> r:install /bin/false|install /bin/true|Module squashfs not found"

Other eg. : sshd -T generate only key/values on lowercase => fixed
Some conditions should be condition: all to condition: any.

fuckingdarkknight:
How to push a branch for PR please ?

fuckingdarkknight:
I've forked the main branch to my own repository : https://github.com/fuckingdarkknight/wazuh/tree/fix/rocky9_sca_rulesets

In addition, I've joined the fixed file.

[cis_rocky_linux_9.txt](https://github.com/user-attachments/files/20182597/cis_rocky_linux_9.txt)

Damian-Mangold:
For visibility @wazuh/devel-cti-threatintel

fuckingdarkknight:
Many thanks, I've made more fixes, file joined.

[cis_rocky_linux_9.txt](https://github.com/user-attachments/files/20189835/cis_rocky_linux_9.txt)

jk-olaoluwa:
Hello @fuckingdarkknight 

I'm taking a look at this and will provide updates accordingly

jk-olaoluwa:
> Many thanks, I've made more fixes, file joined.
> 
> [cis_rocky_linux_9.txt](https://github.com/user-attachments/files/20189835/cis_rocky_linux_9.txt)

Hello @fuckingdarkknight 

I noticed you have the full modified sca document in the txt file you attached, can you let me know the IDs you found the false positives with so I can take a look and test them for applicability.

Thank you!

fuckingdarkknight:
Hello,

IDs list :
31500
31501
31528
31578
31580
31583
31598
31621
31622
31623
31624
31625
31626
31627
31628
31629
31630
31632
31633
31635
31636
31637

Please note that the issue may be the same for Rocky 8, Alma 8, Alma 9 ... EL8 and EL9 in others words.

Thanks

jk-olaoluwa:
> Hello,
> 
> IDs list : 31500 31501 31528 31578 31580 31583 31598 31621 31622 31623 31624 31625 31626 31627 31628 31629 31630 31632 31633 31635 31636 31637
> 
[...]

PR #29976, opened 2025-05-30
Fix multiple Rocky Linux SCA checks generating incorrect results

IsExec:
## Description

A user reported multiple SCA checks with false results on Rocky Linux 9 [here](https://github.com/wazuh/wazuh/issues/29618).

## Observation

Check ID | Change Required
-- | --
31500 | Condition   should be ANY and `install /bin/true`    should be included in the regex
31501 | Condition   should be ANY and `install /bin/true`    and `blacklist` should be included in the regex
31528 | Condition   should be ANY and `install /bin/false`   and `blacklist` should be included in the regex
31578 | Condition   should be ANY and `install /bin/true`    and `blacklist` should be included in the regex
31580 | Logic   needs to be redefined to verify only one of NFT or Firewalld is active while   the other is disabled.
31583 | Duplicate   rule should be removed.
31598 | The   current rule works as intended as the proposed rule introduces a false   positive that was reported in   https://github.com/wazuh/external-devel-requests/issues/5040
31621 | Having   the Condition as ANY is more suitable.
31622 | Having   the Condition as ANY is more suitable.
31623 | Having   the Condition as ANY is more suitable.
31624 | Having   the Condition as ANY is more suitable.
31625 | Having   the Condition as ANY is more suitable.
31626 | Having   the Condition as ANY is more suitable.
31627 | Having   the Condition as ANY is more suitable.
31628 | Having   the Condition as ANY is more suitable.
31629 | Having   the Condition as ANY is more suitable.
31630 | Having   the Condition as ANY is more suitable.
31632 | Having   the Condition as ANY is more suitable.
31633 | Having   the Condition as ANY is more suitable.
31635 | The   current rule works as expected.
31636 | Having   the Condition as ANY is more suitable.
31637 | Having   the Condition as ANY is more suitable.

### DIFF
--- a/ruleset/sca/rocky/cis_rocky_linux_9.yml
+++ b/ruleset/sca/rocky/cis_rocky_linux_9.yml
@@ -48,9 +48,9 @@ checks:
       - pci_dss_3.2.1: ["1.1.6", "1.2.1", "2.2.2", "2.2.5"]
       - pci_dss_4.0: ["1.2.5", "2.2.4", "6.4.1"]
       - soc_2: ["CC6.3", "CC6.6"]
-    condition: all
+    condition: any
     rules:
-      - "c:modprobe -n -v squashfs -> r:install /bin/false|Module squashfs not found"
+      - "c:modprobe -n -v squashfs -> r:install /bin/false|install /bin/true|Module squashfs not found"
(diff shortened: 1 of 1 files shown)


AwwalQuan (approved):
LGTM!

PR #30259, opened 2025-06-10
Exclude null/empty groups from agents summary

### LINKED ISSUE #30234: Fix wazuh-db HTTP server summary endpoint
## Description

We executed the API integration tests for https://github.com/wazuh/wazuh/issues/29563 and there seems to be some issues with the results returned by the wazuh-db HTTP endpoint `/agents/summary`  after the default groups rollback.

It's returning empty named groups, which should be omitted in the response since those agents do not belong to any group.

```
{'': 1, 'ubuntu': 5}
```

And groups with too many agents assigned to them

```
{'group1': 50} != {'group1': 7}
{'group2': 50} != {'group2': 8}
{'group3': 40} != {'group3': 4}
```

Which might be related to the fact that the query we were using before may be outdated and requires to be updated after the changes in the agent groups behavior.

We should review the current behavior of the endpoint and fix it to return the expected output.

Dwordcito:
|Related issue|
|---|
|Closes #30234 |

### Description

This Pull Request enhances the agents summary endpoint (`POST /agents/summary`) to exclude groups that have a null or empty name (`NULL` or `''`), ensuring that only meaningful data is displayed in the API response.

#### Problem Solved

Currently, the agents summary endpoint may include groups in its count that do not have an assigned name in the database. This can lead to the API returning an "empty" group category, which is not useful for users and can cause confusion in frontend visualizations.

#### Description of Changes

To address this, the following modifications have been made:

1.  **Updated SQL Queries:**
    *   The `queryGroupsNoFilter` query has been modified to include the condition `g.name IS NOT NULL AND g.name <> ''`. This ensures that the general group summary only counts agents belonging to groups with valid names.
    *   The same filter has been applied to the `queryGroup` query, which is used when a summary is requested for a specific set of agents.

2.  **Updated Unit Tests:**
    *   Two new test cases (`NoFilter_ShouldIgnoreNullAndEmptyGroups` and `Filtered_ShouldIgnoreNullAndEmptyGroups`) have been added to explicitly verify that null or empty groups are ignored by the queries.
    *   Existing tests (`NoFilterHappyCase` and `FilteredHappyCase`) have been updated so their assertions match the new SQL queries, ensuring the entire test suite continues to pass.

### DIFF
--- a/src/shared_modules/router/src/wazuh-db/endpointPostV1AgentsSummary.hpp
+++ b/src/shared_modules/router/src/wazuh-db/endpointPostV1AgentsSummary.hpp
@@ -54,8 +54,9 @@ public:
         // Queries for connections
         constexpr std::string_view queryConnection = "SELECT id, connection_status AS status FROM agent WHERE id > 0;";
         constexpr std::string_view queryOs = "SELECT id, os_platform AS platform FROM agent WHERE id > 0;";
-        constexpr std::string_view queryGroup = "SELECT b.id_agent, g.name AS group_name FROM belongs b JOIN 'group' g "
-                                                "ON b.id_group=g.id WHERE b.id_agent > 0;";
+        constexpr std::string_view queryGroup =
+            "SELECT b.id_agent, g.name AS group_name FROM belongs b JOIN 'group' g "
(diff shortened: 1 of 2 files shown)


juliancnn (approved):
LGTM!

Issue #30304, opened 2025-06-11
Vulnerability Scanner - Database Generation test fail: make target `compile_schemas` not found


JcabreraC:
### Description

During the execution of the `Vulnerability Scanner - Database Generation` test, the build fails due to a missing make target:

```
-- Build files have been written to: /home/runner/work/wazuh/wazuh/src/wazuh_modules/syscollector/build
make: *** No rule to make target 'compile_schemas'.  Stop.
make: Entering directory '/home/runner/work/wazuh/wazuh/src/wazuh_modules/syscollector/build'
make: Leaving directory '/home/runner/work/wazuh/wazuh/src/wazuh_modules/syscollector/build'
Error: Process completed with exit code 2.
```

### Analysis

It seems the make system is expecting a target named `compile_schemas`, but it isn't defined. This may be related to recent changes in the schema generation or flag handling logic (possibly related to `OS` flag configuration or conditional targets).

### Steps to Reproduce

1. Run the `Vulnerability Scanner - Database Generation` test (CI pipeline or manual build).
2. Observe failure during schema compilation stage.

### Expected Behavior

Make should find and execute the `compile_schemas` target successfully during the build process.

### CI Workflow Link

- [Vulnerability Scanner Database Workflow](https://github.com/wazuh/wazuh/actions/workflows/4_builderprecompiled_vulnerability-scanner-database.yml?query=event%3Aschedule)

JcabreraC:
## Update
- Fix implemented and tested on a temporary branch
- Local tests confirm that the DB is built correctly
- Compilation warnings were found and will also be addressed in this PR

GabrielEValenzuela:
# Update

I made a research and I found that actions uses the `action.yml` in `main` branch, more specifically the last commit merged on this PR https://github.com/wazuh/wazuh/pull/30258 

On https://github.com/wazuh/wazuh/pull/29704, the `action.yml` of compilation doesn't fully align with https://github.com/wazuh/wazuh/pull/28345, as the `CMakeLists.txt` does. 

I propose the same change and leave this pending to be reviewed by team, due if I make a manual trigger of the Database Generation, the workflow is different that scheduled.

GabrielEValenzuela:
After a meet with @fdalmaup , we decide apply a fix modifying the workflow and not the action

PR #25798, opened 2024-09-19
Vulnerability Detector - Add 4.10.0  pipeline to generate content

sebasfalcone:
|Related issue|
|---|
| #25782 |

## Description
This PR introduces a new pipeline for the VD-generated content. It now uses 4.10.0 branch to generate the DBs and runs the efficacy tests using this new content

To do so it:
- Aligns the scanner testtool to work with the changes in the components
- Updates the workflows to also run on 4.10.0
- Updates the logic in code to retrieve the new generated compressed DB

## New content
New content was generated here:
- https://github.com/wazuh/wazuh/actions/runs/10950973244/job/30407211577

PR #30383, opened 2025-06-13
Fix compile_schemas build failure

JcabreraC:
|Related issue|
|---|
|#30304|

## Description

This PR resolves a critical build failure in the Vulnerability Scanner - Database Generation workflow, which was caused by a wrong action execution.

## Changes

- Apply a workaround for version `4.8.0`, changing name of the input
- Avoid use main workflow, and use branch workflow

### DIFF
--- a/.github/workflows/4_builderprecompiled_vulnerability-scanner-database.yml
+++ b/.github/workflows/4_builderprecompiled_vulnerability-scanner-database.yml
@@ -3,6 +3,13 @@ name: Vulnerability Scanner - Database Generation
 on:
   schedule:
     - cron: '20 0 * * *'
+  pull_request:
+    paths:
+      - ".github/workflows/4_builderprecompiled_vulnerability-scanner-database.yml"
+      - ".github/actions/compile_and_test/action.yml"
+      - ".github/actions/vulnerability_scanner_deps/action.yml"
+      - ".github/actions/vulnerability_scanner/content_generation/action.yml"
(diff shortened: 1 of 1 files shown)


src/wazuh_modules/syscollector/CMakeLists.txt
```diff
@@ -109,6 +109,8 @@ else()
   target_link_libraries(syscollector dbsync rsync sysinfo wazuhext -Wl,-blibpath:${INSTALL_PREFIX}/lib:${CXX_IMPLICIT_LINK_DIRECTORIES_STR}:${PLATFORM_REQUIRED_RUNTIME_PATH_STR})
 endif(NOT CMAKE_SYSTEM_NAME STREQUAL "AIX")
 
+add_dependencies(syscollector compile_schemas_input)
```
GabrielEValenzuela: I think we should add a conditional to only add this if target equals server

fdalmaup (approved):
- Adding conditions to run the workflow on pull request modifications will allow us to identify if any breaking changes are being introduced in the `main` branch.
- The modification of `wazuh_version` to put `content_version` with the `sed` command is due to the change introduced at https://github.com/wazuh/wazuh/pull/25798. The `v4.8.0` workflow is run with the `wazuh_version` field, so when running the workflow from main, it will use a different field ( `content_version` ) and break. Therefore, when checking out, we are changing the actions to receive `content_version` as output. 

LGTM

Issue #30474, opened 2025-06-18
Propagate repository bumper functionality from 4.13.0 through 6.0.0


rauldpm:
We need to propagate the repository bumper development from the `4.13.0` branch to all subsequent branches.

> **Requirements and Restrictions**  
> Detailed information is available in the following issue:  
> https://github.com/wazuh/internal-devel-requests/issues/2290

### Tasks

- [ ] Migrate and adapt changes from the `4.13.0` branch to:
  - [ ] `4.13.1`
  - [ ] `4.14.0`
  - [ ] `main` (`5.0.0`)
  - [ ] `6.0.0`

 
> [!NOTE]
> This action should be performed for every branch from version 4.13.0 onward.
> If the repository is intended to be used with a version later than 4.13.0, the functionality should be implemented in accordance with the specified requirements and restrictions (Script + CI).

JcabreraC:
All branches up to `main` (`5.0.0`) already include the script and workflow.

Should we rename the workflow file from `4_bumper_repository.yml` to `5_bumper_repository.yml` in the `main` branch?

Also, should we add a new workflow named `6_bumper_repository.yml` in the `6.0.0` branch?

PR #30499, opened 2025-06-23
Added bumper workflow to 6.0.0

JcabreraC:
|Related issue|
|---|
|#30474|

This PR add the workflow file `6_bumper_repository.yml`.

### DIFF
--- a/.github/workflows/6_bumper_repository.yml
+++ b/.github/workflows/6_bumper_repository.yml
@@ -1,0 +1,140 @@
+name: Repository bumper
+run-name: Bump ${{ github.ref_name }} (${{ inputs.id }})
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Target version (e.g. 1.2.3)'
+        default: ''
+        required: false
+        type: string
+      stage:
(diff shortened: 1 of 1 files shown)


.github/workflows/6_bumper_repository.yml
```diff
@@ -0,0 +1,140 @@
...
+        required: false
+        type: string
+      stage:
+        description: 'Version stage (e.g. alpha0)'
+        default: ''
+        required: false
+        type: string
+      date:
+        description: 'Release date in format YYYY-MM-DD (e.g., 2025-04-14)'
+        default: ''
```
GabrielEValenzuela, line 19: I think we have an error on this line

`error parsing called workflow ".github/workflows/6_builderpackage_server_pr.yml" -> "./.github/workflows/5_builderpackage_server.yml" : failed to fetch workflow: workflow was not found.`

fdalmaup: The problem is with the `6_builderpackage_server_pr.yml` workflow, which has not been modified accordingly when the `main` and `6.0.0` branches were generated.

PR #30608, opened 2025-06-27
Adds GeoIP enrichment for Azure AD user and caller IPs

yenienserrano:
## Description

Geolocation for Azure logs is added to the alerts pipeline taking into account the `data.properties.initiatedBy.user.ipAddress` and `data.callerIpAddress` fields.


asteriscos (approved):
LGTM :green_circle:

Issue #30615, opened 2025-06-30
Update RBAC black manager AIT restart response status code


GGP1:
## Description

During the execution of the tests for https://github.com/wazuh/wazuh/pull/30598, I noticed that the test `test_rbac_black_manager_endpoints.tavern.yaml` was failing and although there is a known issue with the `GET /manager/configuration/validation` endpoint which returns a timeout, there was another failure during the manager restart caused by a reused expected response which has a 200 status code instead of 202.

We should update the response to avoid the error.

## Checks

The following elements have been updated or reviewed (should also be checked if no modification is required):
- [ ] Tests (unit tests, API integration tests).
- [ ] Changelog.
- [ ] Documentation.
- [ ] Integration test mapping (using `api/test/integration/mapping/_test_mapping.py`).

GGP1:
## Update

Applied the change and executed the integration test.

PR #30616, opened 2025-06-30
Update RBAC black manager test case response

GGP1:
|Related issue|
|---|
| Closes https://github.com/wazuh/wazuh/issues/30615 |

## Description

Updates the expected test response status code to 202 instead of sharing it with the other endpoints and expecting 200.

### DIFF
--- a/api/test/integration/test_rbac_black_manager_endpoints.tavern.yaml
+++ b/api/test/integration/test_rbac_black_manager_endpoints.tavern.yaml
@@ -247,4 +247,9 @@ stages:
       headers:
         Authorization: "Bearer {test_login_token}"
     response:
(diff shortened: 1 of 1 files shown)


GGP1:
> ## Review
> 
> This test was failing when run through the pipeline but passed locally. We should probably try running it from Jenkins with the changes to make sure everything works as expected.

That error is thrown in the `GET /manager/configuration/validation`, it is not related to the changes in this PR and I'm investigating it separately. I'm about to open an issue related to it, because it seems the timeout is caused by the core module in charge of handling the command.

Edit: issue https://github.com/wazuh/wazuh/issues/30623 created for this

PR #30627, opened 2025-06-30
Fix DFM graceful shutdown

Dwordcito:
|Related issue|
|---|
|Closes #37|

## Description

The key issue is that `ScanOrchestrator` captures `m_databaseFeedManager` by value in the lambda passed to `EventDispatcher::startWorker()`:

```cpp
m_eventDispatcher->startWorker(
    [scanOrchestrator](std::queue<rocksdb::PinnableSlice>& dataQueue) {
        // scanOrchestrator holds a reference to DatabaseFeedManager
        // This prevents DatabaseFeedManager from being destroyed
    });
```

Since the scanner threads are still running and holding this reference, the `DatabaseFeedManager` reference count remains > 0, preventing its destruction and the subsequent cleanup of the `Action` object that would set `shouldStop = false`.

## Solution
1. Removing the `shouldStop` parameter** from callback signatures
2. Using a member variable `m_shouldStop`** in `DatabaseFeedManager` 
3. Properly managing the shutdown flag** without relying on destructor order

PR #30652, opened 2025-07-01
Fix ThreadEventDispatcher shutdown delay

### LINKED ISSUE #30651: ThreadEventDispatcher shutdown delay due to unnecessary sleep
# ThreadEventDispatcher shutdown delay due to unnecessary sleep

## Summary

The `ThreadEventDispatcher` class has a performance issue where it executes a 1-second sleep during graceful shutdown, causing unnecessary delays in the shutdown process.

## Problem Description

When the `ThreadEventDispatcher` is shut down gracefully (via the `cancel()` method), the dispatch loop may encounter exceptions during the shutdown process. Currently, these exceptions trigger a 1-second sleep even when the dispatcher is stopping, which:

1. **Delays shutdown**: The process takes longer to close than necessary
2. **Poor user experience**: Users have to wait unnecessarily for the application to close
3. **Resource consumption**: The thread remains active longer than needed
4. **Inconsistent behavior**: Shutdown timing varies depending on exception timing

## Current Behavior

```cpp
catch (const std::exception& ex)
{
    // Sleep for a second to avoid busy loop
    std::this_thread::sleep_for(std::chrono::seconds(1));
    std::cerr << "Dispatch handler error, " << ex.what() << "\n";
}
```

The sleep is executed regardless of whether the dispatcher is running or shutting down.

## Expected Behavior

The exception handler should:
- Only execute the sleep when the dispatcher is actively running
- Complete shutdown immediately when the dispatcher is stopping
- Provide responsive shutdown behavior

## Proposed Solution

Modify the exception handling to include a check for the `m_running` state:

```cpp
catch (const std::exception& ex)
{
    // Sleep for a second to avoid busy loop
    if (m_running)
    {
        std::this_thread::sleep_for(std::chrono::seconds(1));
        std::cerr << "Dispatch handler error, " << ex.what() << "\n";
    }
    else
    {
        std::cout << "ThreadEventDispatcher dispatch end.\n";
    }
}
```

Dwordcito:
|Related issue|
|---|
|Closes #30651|

# Fix ThreadEventDispatcher shutdown delay

## Description

This PR fixes a performance issue in the `ThreadEventDispatcher` class where a 1-second sleep was being executed during graceful shutdown, causing unnecessary delays in the shutdown process.

## Changes

- **Modified**: `src/shared_modules/utils/threadEventDispatcher.hpp`
  - Removed the 1-second sleep during graceful shutdown in the exception handling block
  - Added proper shutdown message when the dispatcher is stopping
  - Improved shutdown responsiveness by avoiding unnecessary delays

## Technical Details

The change modifies the exception handling block in the `dispatch()` method to avoid the sleep during shutdown:

```cpp
catch (const std::exception& ex)
{
    // Sleep for a second to avoid busy loop
    if (m_running)
    {
        std::this_thread::sleep_for(std::chrono::seconds(1));
        std::cerr << "Dispatch handler error, " << ex.what() << "\n";
    }
    else
    {
        std::cout << "ThreadEventDispatcher dispatch end.\n";
    }
}
```

This ensures that:
- The 1-second sleep only occurs when the dispatcher is actively running
- Shutdown operations complete immediately without unnecessary delays
- The dispatcher responds quickly to shutdown requests

## Testing

- [x] Verified that error handling still works correctly during normal operation
- [x] Confirmed that shutdown completes immediately without delays

### DIFF
--- a/src/shared_modules/utils/threadEventDispatcher.hpp
+++ b/src/shared_modules/utils/threadEventDispatcher.hpp
@@ -214,8 +214,15 @@ private:
             catch (const std::exception& ex)
             {
                 // Sleep for a second to avoid busy loop
-                std::this_thread::sleep_for(std::chrono::seconds(1));
-                std::cerr << "Dispatch handler error, " << ex.what() << "\n";
+                if (m_running)
+                {
+                    std::this_thread::sleep_for(std::chrono::seconds(1));
+                    std::cerr << "Dispatch handler error, " << ex.what() << "\n";
+                }
+                else
+                {
+                    std::cout << "ThreadEventDispatcher dispatch end.\n";
+                }
             }
         }
     }


src/shared_modules/utils/threadEventDispatcher.hpp
```diff
@@ -214,8 +214,15 @@ class TThreadEventDispatcher
             catch (const std::exception& ex)
             {
                 // Sleep for a second to avoid busy loop
-                std::this_thread::sleep_for(std::chrono::seconds(1));
-                std::cerr << "Dispatch handler error, " << ex.what() << "\n";
+                if (m_running)
```
GabrielEValenzuela, line 217: Excellent approach on this fix!

GabrielEValenzuela (approved):
LGTM!

I don't have the fix of https://github.com/wazuh/wazuh/pull/30627 but, compilation and stop command works as expected

```console
root@deploy:/home/deploy/wazuh# /var/ossec/bin/wazuh-control stop
Killing wazuh-clusterd...
Killing wazuh-modulesd...
Killing wazuh-monitord...
Killing wazuh-logcollector...
Killing wazuh-remoted...
Killing wazuh-syscheckd...
Killing wazuh-analysisd...
wazuh-maild not running...
Killing wazuh-execd...
Killing wazuh-db...
Killing wazuh-authd...
wazuh-agentlessd not running...
wazuh-integratord not running...
wazuh-dbd not running...
wazuh-csyslogd not running...
Killing wazuh-apid...
Wazuh v4.13.0 Stopped
```

PR #30928, opened 2025-07-16
Fix libyaml build in 5.x

aritosteles:
## Description

Fix build issues related to libyaml

### DIFF
--- a/src/wazuh_modules/sca/sca_impl/src/yaml_node.cpp
+++ b/src/wazuh_modules/sca/sca_impl/src/yaml_node.cpp
@@ -270,101 +270,113 @@ int YamlNode::GetId() const
 
 YamlNode YamlNode::CreateEmptySequence(const std::string& key)
 {
-    // if (!IsMap())
-    // {
-    //     throw std::runtime_error("Not a map node");
-    // }
+    if (!IsMap())
+    {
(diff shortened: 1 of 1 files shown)


src/wazuh_modules/sca/CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ if(NOT CMAKE_BUILD_TYPE)
   endif()
 endif()
 
+# find_package(pcre2 CONFIG REQUIRED)
```
jr0me: Please revert this file's changes.

jr0me (commented):
I noticed that there are some new calls to free with the pointer to `tag`. Should these be added to 6.x as well?

aritosteles:
> I noticed that there are some new calls to free with the pointer to `tag`. Should these be added to 6.x as well?

I'm not sure. These result from a build error that doesn´t come up in 6.x. But probably safer to add them to 6.x too.

src/wazuh_modules/sca/CMakeLists.txt
aritosteles: Done

jr0me (approved):
LGTM

(no worry about the free to tag, I realized that this yaml implementation is not used in 6.x)

PR #31095, opened 2025-07-25
Enhancement in remoted inventory data transformation

Dwordcito:
Description

During performance testing of the 4.13 release, we observed a significant degradation in the performance of the wazuh-remoted module.

The number of processed events has dropped notably, while the number of discarded events has increased. This issue seems to correlate with the processing queue filling up, which indicates that events are not being dispatched quickly enough, leading to backpressure and eventual drops.


juliancnn (approved):
LGTM!

PR #31100, opened 2025-07-25
Set cmocka version in UT macOS workflows

jotacarma90:
# Description
In this PR we are forcing the installation of cmocka in macOS unit test workflows to have version 1.1.7 to avoid compilation problems.

# Test


lchico (dismissed):
LGTM!

mjcr99 (dismissed):

vikman90 (approved):
LGTM.

This command won't work:

```
brew install cmocka@1.1.7
```

That's why we need to install from sources.

PR #31114, opened 2025-07-28
Add single writer buffer manager instance for each indexer connector instances.

Dwordcito:
## Summary

This PR introduces a single writer buffer manager instance for each indexer connector instance, improving buffer management and concurrency safety.

### Key Changes

- **Buffer Management:**  
  - Added a dedicated writer buffer manager for each indexer connector instance.
- **Configuration:**  
  - Updated `etc/internal_options.conf` to support the new buffer management approach.
- **Code Cleanup:**  
  - Updated and improved buffer usage in `rocksDBQueue.hpp` and `rocksDBQueueCF.hpp`.
  - Minor improvements in `threadEventDispatcher.hpp` and `main.c`.
- **Testing:**  
  - Added comprehensive tests for the new buffer management logic in `rocksDBSafeQueue_test.cpp`.


GabrielEValenzuela (dismissed):
Although I wasn't able to run a large-scale test like the ones done in QA, I installed this version on an AIO setup and connected agents from different OSes, keeping all modules active and letting the system reach a steady state after a few hours.
Meanwhile, I monitored the memory status via, every 10 seconds:
**Monitoring /proc/self/statm**

```console
Time       Size     Resident Shared   Text     Lib      Data     Dirty  
18:12:04   1457     448      448      5        0        123      0  
```

These values remained constant with no variation.

No errors or warnings were found in `ossec.log`.

The code looks safe and well structured, so to fully validate this approach, it would be necessary to run the same QA test to compare results using the same parameters.
That said, **LGTM**.

PR #31251, opened 2025-08-06
State persistance: Timeout logic implementation

GabrielEValenzuela:
_WIP_

### DIFF
--- a/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
+++ b/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
@@ -457,9 +457,8 @@ public:
     }
     void bulkDelete(std::string_view id, std::string_view index)
     {
-        constexpr auto FORMATTED_SIZE {21 + 9 + 2 + 1};
-        std::lock_guard lock(m_mutex);
-        if (m_bulkData.length() + FORMATTED_SIZE + index.size() + id.size() > MaxBulkSize)
+        if (constexpr auto FORMATTED_SIZE {21 + 9 + 2 + 1};
+            m_bulkData.length() + FORMATTED_SIZE + index.size() + id.size() > MaxBulkSize)
         {
             processBulk();
         }
@@ -473,10 +472,8 @@ public:
     }
     void bulkIndex(std::string_view id, std::string_view index, std::string_view data)
     {
-        constexpr auto FORMATTED_SIZE {20 + 8 + 2 + 2 + 1};
-        std::lock_guard lock(m_mutex);
-
-        if (m_bulkData.length() + FORMATTED_SIZE + index.size() + id.size() + data.size() > MaxBulkSize)
+        if (constexpr auto FORMATTED_SIZE {20 + 8 + 2 + 2 + 1};
+            m_bulkData.length() + FORMATTED_SIZE + index.size() + id.size() + data.size() > MaxBulkSize)
         {
             processBulk();
         }
@@ -498,5 +495,15 @@ public:
         {
             processBulk();
         }
+    }
+
+    std::mutex& scopeLock()
+    {
+        return m_mutex;
+    }
+
+    void registerNotify(std::function<void()> callback)
+    {
+        m_notify.push_back(std::move(callback));
     }
 };
--- a/src/wazuh_modules/inventory_sync/src/agentSession.hpp
+++ b/src/wazuh_modules/inventory_sync/src/agentSession.hpp
@@ -1,3 +1,15 @@
+
+/*
+ * Wazuh inventory sync
+ * Copyright (C) 2015, Wazuh Inc.
+ * August 6, 2025.
+ *
+ * This program is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU General Public
+ * License (version 2) as published by the FSF - Free Software
+ * Foundation.
+ */
+
 #ifndef _AGENT_SESSION_HPP
 #define _AGENT_SESSION_HPP
 
@@ -44,15 +56,27 @@ private:
     std::string m_message;
 };
 
+/**
+ * @class AgentSessionImpl
+ * @brief Manages the lifecycle of a data ingestion session from a specific agent.
+ *
+ * Handles sequential chunk reception, stores data into RocksDB, tracks missing pieces using GapSet,
+ * and sends appropriate acknowledgments or error notifications via dispatcher queues.
+ *
+ * @tparam TStore Store interface (e.g., RocksDB wrapper).
+ * @tparam TIndexerQueue Queue used to notify indexing system upon session completion.
+ * @tparam TResponseDispatcher Dispatcher used to send ACK/NACK messages during session flow.
+ */
+
 template<typename TStore, typename TIndexerQueue, typename TResponseDispatcher>
 class AgentSessionImpl final
 {
-    std::unique_ptr<GapSet> m_gapSet;
-    std::shared_ptr<Context> m_context;
-    TStore& m_store;
-    TIndexerQueue& m_indexerQueue;
-    bool m_endReceived = false;
-    std::mutex m_mutex;
+    std::unique_ptr<GapSet> m_gapSet;   ///< Tracker for received/missing data chunks
+    std::shared_ptr<Context> m_context; ///< Shared metadata for the current session
+    TStore& m_store;                    ///< Reference to RocksDB store
+    TIndexerQueue& m_indexerQueue;      ///< Response queue for indexing subsystem
+    bool m_endReceived = false;         ///< Whether the END message has been received
+    std::mutex m_mutex;                 ///< Mutex to guard shared state
 
 public:
     explicit AgentSessionImpl(const uint64_t sessionId,
@@ -68,12 +92,6 @@ public:
         {
             throw AgentSessionException("Invalid data");
         }
-        // Create new session.
-        if (data->size() == 0)
-        {
-            throw AgentSessionException("Invalid size");
-        }
-        m_gapSet = std::make_unique<GapSet>(data->size());
 
         if (data->module_() == nullptr)
         {
@@ -84,6 +102,13 @@ public:
         {
             throw AgentSessionException("Invalid id");
         }
+
+        // Create new session.
+        if (data->size() == 0)
+        {
+            throw AgentSessionException("Invalid size");
+        }
+        m_gapSet = std::make_unique<GapSet>(data->size());
 
         m_context = std::make_shared<Context>(Context {.mode = data->mode(),
                                                        .sessionId = sessionId,
@@ -96,6 +121,25 @@ public:
         responseDispatcher.sendStartAck(Wazuh::SyncSchema::Status_Ok, m_context);
     }
 
+    /// Deleted copy constructor and assignment operator (C.12 compliant).
+    AgentSessionImpl(const AgentSessionImpl&) = delete;
+    AgentSessionImpl& operator=(const AgentSessionImpl&) = delete;
+
+    /// Deleted move constructor and assignment operator.
+    AgentSessionImpl(AgentSessionImpl&&) = delete;
+    AgentSessionImpl& operator=(AgentSessionImpl&&) = delete;
+
+    ~AgentSessionImpl() = default;
+
+    /**
+     * @brief Handles an incoming data chunk.
+     *
+     * Stores the raw payload and marks the chunk as observed in the GapSet.
+     * Triggers indexing if `handleEnd()` was already called and the session is now complete.
+     *
+     * @param data Parsed flatbuffer metadata (e.g., sequence number).
+     * @param dataRaw Raw binary payload of the chunk.
+     */
     void handleData(Wazuh::SyncSchema::Data const* data, const std::vector<char>& dataRaw)
     {
         const auto seq = data->seq();
@@ -107,16 +151,22 @@ public:
         std::lock_guard lock(m_mutex);
         m_gapSet->observe(data->seq());
 
-        // std::cout << "Data received: " << std::to_string(session) + "_" + std::to_string(seq) << "\n";
-
         if (m_endReceived)
         {
             if (m_gapSet->empty())
             {
-                m_indexerQueue.push(Response {.status = ResponseStatus::Ok, .context = m_context});
+                m_indexerQueue.push(Response {ResponseStatus::Ok, m_context});
             }
         }
     }
+
+    /**
+     * @brief Handles the end-of-transmission signal from the agent.
+     *
+     * If all chunks were received, pushes the final acknowledgment. Otherwise, triggers missing range dispatch.
+     *
+     * @param responseDispatcher Dispatcher used to report missing sequences (if any).
+     */
 
     void handleEnd(const TResponseDispatcher& responseDispatcher)
     {
@@ -125,7 +175,7 @@ public:
         if (m_gapSet->empty())
         {
             std::cout << "End received and gap set is empty\n";
-            m_indexerQueue.push(Response {.status = ResponseStatus::Ok, .context = m_context});
+            m_indexerQueue.push(Response {ResponseStatus::Ok, m_context});
         }
         else
         {
@@ -133,13 +183,14 @@ public:
         }
     }
 
+    /**
+     * @brief Checks whether the session has timed out based on last activity.
+     * @param timeout The allowed inactivity duration.
+     * @return true if the session has been idle for longer than the timeout.
+     */
     bool isAlive(const std::chrono::seconds timeout) const
     {
-        if (m_gapSet->lastUpdate() + timeout < std::chrono::steady_clock::now())
-        {
-            return false;
-        }
-        return true;
+        return m_gapSet->lastUpdate() + timeout < std::chrono::steady_clock::now();
     }
 };
 
--- a/src/wazuh_modules/inventory_sync/src/agentSession.hpp
+++ b/src/wazuh_modules/inventory_sync/src/agentSession.hpp
@@ -1,12 +1,12 @@
 #ifndef _AGENT_SESSION_HPP
 #define _AGENT_SESSION_HPP
 
+#include "asyncValueDispatcher.hpp"
 #include "context.hpp"
 #include "flatbuffers/include/inventorySync_generated.h"
 #include "gapSet.hpp"
 #include "responseDispatcher.hpp"
 #include "rocksDBWrapper.hpp"
-#include "threadDispatcher.h"
 #include <functional>
 #include <memory>
 #include <string>
@@ -24,8 +24,8 @@ struct Response
     std::shared_ptr<Context> context;
 };
 
-using WorkersQueue = Utils::AsyncDispatcher<std::vector<char>, std::function<void(const std::vector<char>&)>>;
-using IndexerQueue = Utils::AsyncDispatcher<Response, std::function<void(const Response&)>>;
+using WorkersQueue = Utils::AsyncValueDispatcher<std::vector<char>, std::function<void(const std::vector<char>&)>>;
+using IndexerQueue = Utils::AsyncValueDispatcher<Response, std::function<void(const Response&)>>;
 
 class AgentSessionException : public std::exception
 {
@@ -85,8 +85,10 @@ public:
             throw AgentSessionException("Invalid id");
         }
 
-        m_context =
-            std::make_shared<Context>(Context {data->mode(), sessionId, data->agent_id(), data->module_()->str()});
+        m_context = std::make_shared<Context>(Context {.mode = data->mode(),
+                                                       .sessionId = sessionId,
+                                                       .agentId = data->agent_id(),
+                                                       .moduleName = data->module_()->str()});
 
         std::cout << "AgentSessionImpl: " << m_context->sessionId << " " << m_context->agentId << " "
                   << m_context->moduleName << std::endl;
@@ -102,7 +104,7 @@ public:
         m_store.put(std::to_string(session) + "_" + std::to_string(seq),
                     rocksdb::Slice(dataRaw.data(), dataRaw.size()));
 
-        std::lock_guard<std::mutex> lock(m_mutex);
+        std::lock_guard lock(m_mutex);
         m_gapSet->observe(data->seq());
 
         // std::cout << "Data received: " << std::to_string(session) + "_" + std::to_string(seq) << "\n";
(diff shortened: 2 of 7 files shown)


src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -457,9 +457,8 @@ class IndexerConnectorSyncImpl final
     }
     void bulkDelete(std::string_view id, std::string_view index)
     {
-        constexpr auto FORMATTED_SIZE {21 + 9 + 2 + 1};
-        std::lock_guard lock(m_mutex);
-        if (m_bulkData.length() + FORMATTED_SIZE + index.size() + id.size() > MaxBulkSize)
+        if (constexpr auto FORMATTED_SIZE {21 + 9 + 2 + 1};
```
MiguelazoDS: Couldn't this be more clear? what do represent those numbers?

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -473,10 +472,8 @@ class IndexerConnectorSyncImpl final
     }
     void bulkIndex(std::string_view id, std::string_view index, std::string_view data)
     {
-        constexpr auto FORMATTED_SIZE {20 + 8 + 2 + 2 + 1};
-        std::lock_guard lock(m_mutex);
-
-        if (m_bulkData.length() + FORMATTED_SIZE + index.size() + id.size() + data.size() > MaxBulkSize)
+        if (constexpr auto FORMATTED_SIZE {20 + 8 + 2 + 2 + 1};
```
MiguelazoDS: Couldn't this be more clear? what do represent those numbers?

src/wazuh_modules/inventory_sync/src/agentSession.hpp
```diff
@@ -107,39 +151,46 @@ class AgentSessionImpl final
...
         m_gapSet->observe(data->seq());
 
-        // std::cout << "Data received: " << std::to_string(session) + "_" + std::to_string(seq) << "\n";
-
         if (m_endReceived)
         {
             if (m_gapSet->empty())
             {
-                m_indexerQueue.push(Response {.status = ResponseStatus::Ok, .context = m_context});
+                m_indexerQueue.push(Response {ResponseStatus::Ok, m_context});
```
MiguelazoDS, line 158: Why removing the designated initializers here?

src/wazuh_modules/inventory_sync/src/gapSet.hpp
```diff
@@ -57,34 +104,39 @@ class GapSet final
...
+    std::vector<bool> m_observed; ///< Observed entries (bitmap)
+    uint64_t m_observedCount {0}; ///< Count of observed entries
+    std::chrono::time_point<std::chrono::steady_clock> m_lastUpdate {
+        std::chrono::steady_clock::now()};                     ///< Last update time
+    bool m_gapCacheDirty {true};                               ///< Whether the gap cache is dirty
+    std::vector<std::pair<uint64_t, uint64_t>> m_cachedRanges; ///< Cached result of gap ranges
 };
 
-#endif // _GAP_SET_HPP
+#endif // _GAP_SET_HPP
```
MiguelazoDS: Add new line here

src/wazuh_modules/inventory_sync/src/agentSession.hpp
```diff
@@ -107,39 +151,46 @@ class AgentSessionImpl final
...
+
     void handleEnd(const TResponseDispatcher& responseDispatcher)
     {
         std::lock_guard lock(m_mutex);
         m_endReceived = true;
         if (m_gapSet->empty())
         {
             std::cout << "End received and gap set is empty\n";
-            m_indexerQueue.push(Response {.status = ResponseStatus::Ok, .context = m_context});
+            m_indexerQueue.push(Response {ResponseStatus::Ok, m_context});
```
MiguelazoDS, line 178: Why removing the designated initializers here?

src/wazuh_modules/inventory_sync/src/agentSession.hpp
```diff
@@ -107,39 +151,46 @@ class AgentSessionImpl final
...
+     * @return true if the session has been idle for longer than the timeout.
+     */
     bool isAlive(const std::chrono::seconds timeout) const
     {
-        if (m_gapSet->lastUpdate() + timeout < std::chrono::steady_clock::now())
-        {
-            return false;
-        }
-        return true;
+        return m_gapSet->lastUpdate() + timeout < std::chrono::steady_clock::now();
```
MiguelazoDS: Is this ok? Isn't expected to return false if the current time is greater than lastUpdate plus timeout? 

I think this definition 

> true if the session has been idle for longer than the timeout.

is the opposite to the function name description.

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
```diff
@@ -256,21 +262,23 @@ class InventorySyncFacadeImpl final
...
+                        [this, ctx = res.context]()
+                        {
+                            // Send ACK to agent.
+                            m_responseDispatcher->sendEndAck(Wazuh::SyncSchema::Status_Ok, ctx);
+                            // Delete data from database.
+                            m_dataStore->deleteByPrefix(std::to_string(ctx->sessionId));
+                            // Delete Session.
+                            if (m_agentSessions.erase(ctx->sessionId) == 0)
+                            {
+                                std::cerr
```
MiguelazoDS, line 284: Do we want this console output directives here? Shouldn't we use loggerHelper class

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
```diff
@@ -129,16 +129,15 @@ class InventorySyncFacadeImpl final
...
             }
-            else
-            {
-                // Handle end.
-                std::cout << "Handling end for session: " << end->session() << std::endl;
-                it->second.handleEnd(*m_responseDispatcher);
-            }
+
+            // Handle end.
+            std::cout << "Handling end for session: " << end->session() << std::endl;
```
MiguelazoDS, line 140: Do we want this console output directives here? Shouldn't we use loggerHelper class?

MiguelazoDS (commented):
Is it expected to have a failing check for Windows agent compilation?

src/wazuh_modules/inventory_sync/src/agentSession.hpp
```diff
@@ -102,34 +148,50 @@ class AgentSessionImpl final
...
     }
+
+    /**
+     * @brief Checks whether the session has timed out based on last activity.
+     * @param timeout The allowed inactivity duration.
+     * @return true if the session has been idle for longer than the timeout.
+     */
+    bool isAlive(const std::chrono::seconds timeout) const
+    {
+        return m_gapSet->lastUpdate() + timeout < std::chrono::steady_clock::now();
```
JcabreraC:
```suggestion
bool isExpired(const std::chrono::seconds timeout) const
{
    return m_gapSet->lastUpdate() + timeout < std::chrono::steady_clock::now();
}

// Or:
bool hasTimedOut(const std::chrono::seconds timeout) const
{
    return m_gapSet->lastUpdate() + timeout < std::chrono::steady_clock::now();
}
```
The logic is inverted. Let’s analyze it step by step:

- `m_gapSet->lastUpdate()` = time of the last activity  
- `m_gapSet->lastUpdate() + timeout` = deadline until it is considered alive  
- `std::chrono::steady_clock::now()` = current time  

The condition `lastUpdate + timeout < now` means:

> "The deadline has already passed" → The session has ALREADY EXPIRED  

So it returns `true` when the session is DEAD/EXPIRED.

Option 2: Fix the logic
```
bool isAlive(const std::chrono::seconds timeout) const
{
    return m_gapSet->lastUpdate() + timeout >= std::chrono::steady_clock::now();
}
```

src/wazuh_modules/inventory_sync/src/agentSession.hpp
JcabreraC: +1

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
JcabreraC: Don`t use magic number, define with `constexpr`

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
```diff
@@ -304,6 +312,29 @@ class InventorySyncFacadeImpl final
...
+
+                    for (auto& [sessionId, session] : m_agentSessions)
+                    {
+                        if (!session.isAlive(std::chrono::seconds(10)))
+                        {
+                            m_agentSessions.erase(sessionId);
+                        }
+                    }
+                }
+            });
```
JcabreraC: The timeout thread implementation has two critical concurrency bugs:
- Condition variable never wakes up - only wakes when stopping, never executes timeout logic
- Iterator invalidation - erasing from map while iterating causes undefined behavior

Suggestion (analyze and test)
```suggestion
m_sessionTimeoutThread = std::thread(
    [this]()
    {
        while (!m_stopping.load())
        {
            std::unique_lock lk(m_sessionTimeoutMutex);
            m_sessionTimeoutCv.wait_for(lk, std::chrono::seconds(5), [this] { return m_stopping.load(); });
            lk.unlock();
            if (m_stopping.load()) break;

            // Two-phase approach: collect expired sessions first
            std::vector<uint64_t> expiredSessions;
            {
                std::shared_lock sessionsLock(m_agentSessionsMutex);
                for (const auto& [sessionId, session] : m_agentSessions)
                {
                    if (!session.isAlive(std::chrono::seconds(10)))
                    {
                        expiredSessions.push_back(sessionId);
                    }
                }
            }

            // Then remove them safely
            if (!expiredSessions.empty())
            {
                std::unique_lock sessionsLock(m_agentSessionsMutex);
                for (auto id : expiredSessions)
                {
                    m_agentSessions.erase(id);
                }
            }
        }
    });
```

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -499,4 +496,14 @@ class IndexerConnectorSyncImpl final
             processBulk();
         }
     }
+
+    std::mutex& scopeLock()
+    {
+        return m_mutex;
+    }
```
JcabreraC: Exposing internal mutex by reference violates encapsulation and can lead to:
- Deadlocks from inconsistent lock ordering
- Race conditions from uncontrolled external locking
- Maintenance issues when internal locking strategy changes

```suggestion
class ScopeGuard
{
public:
    explicit ScopeGuard(std::mutex& m) : m_lock(m) {}
    ScopeGuard(const ScopeGuard&) = delete;
    ScopeGuard& operator=(const ScopeGuard&) = delete;
private:
    std::unique_lock<std::mutex> m_lock;
};

ScopeGuard scopeLock()
{
    return ScopeGuard{m_mutex};
}
```
or template function

JcabreraC (changes requested):

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
GabrielEValenzuela: Done!

src/wazuh_modules/inventory_sync/src/agentSession.hpp
GabrielEValenzuela: Done!

src/wazuh_modules/inventory_sync/src/gapSet.hpp
GabrielEValenzuela: I think already has a new line

src/wazuh_modules/inventory_sync/src/agentSession.hpp
GabrielEValenzuela: According to clangd, the designated initializers don’t add semantic value here—they’re just verbose. We can use aggregate initialization instead because the braced initializer list matches the member declaration order in Response, both members have the correct types, and there’s no narrowing or implicit conversions. This is shorter to write, consistent with normal aggregate initialization, and avoids any ambiguity that designated initializers would resolve.

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
GabrielEValenzuela: We implemented the logging in #31313

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
GabrielEValenzuela: +1

src/wazuh_modules/inventory_sync/src/agentSession.hpp
GabrielEValenzuela: Thanks!

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
GabrielEValenzuela: Fixed!

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
GabrielEValenzuela: Good catch ! Instead of exposing the raw internal `std::mutex` to external callers, the method now:

* Returns a `std::unique_lock<std::mutex>` that already holds the lock.
* Allows the caller to manage the lock's lifetime using RAII — the lock is automatically released when the returned object goes out of scope.
* Prevents external code from directly touching the internal mutex, preserving **encapsulation**.

The reason is based on:

1. **Encapsulation** — The internal synchronization mechanism stays private.
2. **RAII Safety** — Lock acquisition and release are tied to object lifetime, reducing the risk of forgetting to unlock.
3. **Flexibility** — Unlike `std::lock_guard`, `std::unique_lock` allows deferred locking, unlocking/relocking, and moving.
4. **Clear Usage Pattern** — Callers just do:

   ```cpp
   auto lock = obj.scopeLock();
   // perform thread-safe operations
   ```

   making the intent explicit and readable.

* The method gives the caller control over the lock duration, which is powerful but also requires discipline — holding the lock too long can hurt concurrency.
* Returning a `std::unique_lock` means the caller must store it in a variable to keep the lock active (not as terse as `std::scoped_lock` inline usage).

src/shared_modules/indexer_connector/src/indexerConnectorAsyncImpl.hpp
```diff
@@ -39,6 +39,11 @@ constexpr auto HTTP_VERSION_CONFLICT {409};
 constexpr auto HTTP_TOO_MANY_REQUESTS {429};
 constexpr auto MINIMAL_ELEMENTS_PER_BULK {1};
 
+// Overhead is the fixed JSON scaffolding for one bulk item:
+// {"index":{"_index":"<index>","id":"<id>"}}
+constexpr auto FORMATTED_LENGTH {20 + 8 + 2 + 2 + 1};
```
JcabreraC, line 44: You shouldn’t use magic numbers, but instead define those numbers with constexpr.

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
JcabreraC: You shouldn’t use magic numbers, but instead define those numbers with constexpr.

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -38,6 +38,10 @@ constexpr auto HTTP_CONTENT_LENGTH {413};
 constexpr auto HTTP_VERSION_CONFLICT {409};
 constexpr auto HTTP_TOO_MANY_REQUESTS {429};
 
+// Overhead is the fixed JSON scaffolding for one bulk item:
+// {"index":{"_index":"<index>","id":"<id>"}}
+constexpr auto FORMATTED_LENGTH {20 + 8 + 2 + 2 + 1};
```
JcabreraC: You shouldn’t use magic numbers, but instead define those numbers with constexpr.

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
```diff
@@ -160,7 +160,7 @@ class InventorySyncFacadeImpl final
     {
         std::cout << "Starting InventorySyncFacade..." << std::endl;
 
-        // TODO This database should be ephemeral and not persistent.
+        std::filesystem::remove_all("inventory_sync");
```
JcabreraC: `std::filesystem::remove_all` may fail

```suggestion
std::error_code ec;
std::filesystem::remove_all("inventory_sync", ec);
if (ec) {
    std::cerr << "Warning: Could not remove inventory_sync directory: " 
              << ec.message() << std::endl;
    // Continue execution, not critical
}```

File: src/wazuh_modules/inventory_sync/src/gapSet.hpp [hunk 15]
```diff
@@ -1,85 +1,264 @@
...
 
 private:
-    uint64_t m_size {0};
-    std::vector<bool> m_observed;
+    uint64_t m_size;
     uint64_t m_observedCount {0};
-    std::chrono::time_point<std::chrono::steady_clock> m_lastUpdate {};
+    bool m_allObserved {false};
+    std::set<Interval> m_intervals;
+    std::chrono::time_point<std::chrono::steady_clock> m_lastUpdate {std::chrono::steady_clock::now()};
```
JcabreraC: It is initialized at the point of declaration, not in the constructor. This may cause synchronization issues.
```
// At the declaration
std::chrono::time_point<std::chrono::steady_clock> m_lastUpdate {};

// In the constructor
explicit GapSet(uint64_t size)
    : m_size(size)
{
    if (size == 0)
    {
        m_allObserved = true;
    }
    m_lastUpdate = std::chrono::steady_clock::now();
}
```

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
```diff
@@ -304,6 +313,34 @@ class InventorySyncFacadeImpl final
...
+                        {
+                            it = m_agentSessions.erase(it); // erase returns next iterator
+                        }
+                        else
+                        {
+                            ++it;
+                        }
+                    }
+                }
+            });
```
JcabreraC, line 348: The timeout thread accesses `m_agentSessions` with proper locking ?
Check carefully, something like this
```
m_sessionTimeoutThread = std::thread([this]() {
    std::unique_lock lock(m_sessionTimeoutMutex);
    while (!m_stopping.load())
    {
        m_sessionTimeoutCv.wait(lock, [this]() { return m_stopping.load(); });
        
        if (m_stopping.load()) break;
        
        // Use shared lock for reading
        {
            std::shared_lock sessionsLock(m_agentSessionsMutex);
            for (auto& [sessionId, session] : m_agentSessions)
            {
                if (!session.isAlive(std::chrono::seconds(DEFAULT_TIME * 2)))
                {
                    // Mark for removal
                    sessionsToRemove.push_back(sessionId);
                }
            }
        }
        
        // Remove with exclusive lock
        if (!sessionsToRemove.empty())
        {
            std::unique_lock sessionsLock(m_agentSessionsMutex);
            for (auto sessionId : sessionsToRemove)
            {
                m_agentSessions.erase(sessionId);
            }
            sessionsToRemove.clear();
        }
    }
});
```

src/wazuh_modules/inventory_sync/src/gapSet.hpp
```diff
@@ -1,85 +1,264 @@
...
+     * @brief Marks a sequence number as observed.
+     * O(log n) complexity where n is number of intervals.
+     */
+    void observe(uint64_t seq)
     {
-        if (seq >= m_size || m_observed[seq])
+        if (seq >= m_size || m_allObserved)
         {
             return;
         }
```
JcabreraC: Do we need to validate the parameters?

```
void observe(uint64_t seq)
{
    if (seq >= m_size)
    {
        throw std::out_of_range("Sequence number out of range");
    }
    
    if (m_allObserved)
    {
        return;
    }
```

src/wazuh_modules/inventory_sync/src/agentSession.hpp
JcabreraC: Fix !

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
GabrielEValenzuela: I think introducing a second mutex creates more problems than it solves. The single mutex approach is simpler, safer, and follows standard patterns. If we need reader/writer optimization later, we should use `std::shared_mutex` instead of two separate mutexes

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
GabrielEValenzuela: Fixed!

src/wazuh_modules/inventory_sync/src/gapSet.hpp
GabrielEValenzuela: Added!

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
GabrielEValenzuela: Done!

src/wazuh_modules/inventory_sync/src/gapSet.hpp
GabrielEValenzuela: Done!

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
GabrielEValenzuela: I think introducing a second mutex creates more problems than it solves. The single mutex approach is simpler, safer, and follows standard patterns. If we need reader/writer optimization later, we should use `std::shared_mutex` instead of two separate mutexes.

src/wazuh_modules/inventory_sync/src/agentSession.hpp
MiguelazoDS: So we are not using designated initializers. It will stay without them.

src/wazuh_modules/inventory_sync/src/agentSession.hpp
MiguelazoDS: I see the change

Issue #31257, opened 2025-08-07
Solaris 11 i386 `4.10.3` packages fail to be built because cJSON external dependency


mjcr99:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.10.3 | Agent | Agent | Packages and Sources | Solaris11 |

## Description

Hi team,

while working on the following issue:
* https://github.com/wazuh/internal-devel-requests/issues/2771

It has been observed that workflows to build the Solaris 11 i386 packages are failing with the following output:

```console
+ gmake deps TARGET=agent
grep: can't open /etc/os-release
Makefile:2677: warning: overriding recipe for target `win32/ui_resource.o'
Makefile:2617: warning: ignoring old recipe for target `win32/ui_resource.o'
Makefile:2680: warning: overriding recipe for target `win32/auth_resource.o'
Makefile:2620: warning: ignoring old recipe for target `win32/auth_resource.o'
curl -so external/cJSON.tar.gz https://packages.wazuh.com/deps/4.10.3/libraries/solaris/i386/cJSON.tar.gz || true
cd external && test -e cJSON.tar.gz && gunzip cJSON.tar.gz || true
cd external && test -e cJSON.tar && tar -xf cJSON.tar || true
test -e external/cJSON.tar && rm external/cJSON.tar || true
test -d external/cJSON ||\
(curl -so external/cJSON.tar.gz https://packages.wazuh.com/deps/4.10.3/libraries/sources/cJSON.tar.gz &&\
cd external && gunzip cJSON.tar.gz && tar -xf cJSON.tar && rm cJSON.tar)
gmake: *** [external/cJSON.tar.gz] Error 60
++ uname -p
+ arch=i386
+ '[' i386 = sparc ']'
+ gmake -j 4 TARGET=agent USE_SELINUX=no
grep: can't open /etc/os-release
Makefile:2677: warning: overriding recipe for target `win32/ui_resource.o'
Makefile:2617: warning: ignoring old recipe for target `win32/ui_resource.o'
Makefile:2680: warning: overriding recipe for target `win32/auth_resource.o'
Makefile:2620: warning: ignoring old recipe for target `win32/auth_resource.o'
Makefile:1015: *** No external directory found. Run "gmake deps" before compiling external libraries.  Stop.
+ exit 1
```

[...]

PR #31225, opened 2025-08-05
Backport: add package Dockerfile dependencies to `4.10.3` version

mjcr99:
|**Related Issue**|
|---|
|https://github.com/wazuh/internal-devel-requests/issues/2771|

## Description

Hi team,

This PR adds some dependencies to the DEB packages Dockerfiles used in the packages generation procedures to allow the upgrade tests in the package building.

PR #31266, opened 2025-08-07
Backport: fix Solaris 11 SSL certs error

mjcr99:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/31257|

## Description

This PR fixes the SSL certificate problem on both _Solaris 11 i386_ and _Solaris 11 SPARC_, both of which had the same error.
The solution has been to download and pass the `curl` certificate bundle to the machine, to be used by default via the `CURL_CA_BUNDLE` environment variable.


MarcelKemp:
- Change applied in the following PR: https://github.com/wazuh/wazuh/pull/31225

PR #31460, opened 2025-08-19
Enhancement/30846 add reason field to sca scan messages

jr0me:
## Description

This PR adds a _reason_ field to the check results reports when a result turns out to be invalid.

### DIFF
--- a/src/wazuh_modules/sca/sca_impl/include/isca_policy.hpp
+++ b/src/wazuh_modules/sca/sca_impl/include/isca_policy.hpp
@@ -14,12 +14,12 @@ public:
     /// @brief Runs the policy check
     /// @param scanInterval Scan interval in milliseconds
     /// @param scanOnStart Scan on start
-    /// @param reportCheckResult Function to report check result
+    /// @param reportCheckResult Function to report check result (policyId, checkId, result, reason)
     /// @param wait Function to wait for the next scan
     virtual void
     Run(std::time_t scanInterval,
         bool scanOnStart,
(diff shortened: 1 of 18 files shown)


src/wazuh_modules/sca/sca_impl/include/isca_policy.hpp
```diff
@@ -5,6 +5,14 @@
 #include <functional>
 #include <string>
 
+struct CheckResult
```
TomasTurina, line 8: 😄

src/wazuh_modules/sca/sca_impl/src/sca_policy_parser.hpp
```diff
@@ -48,12 +53,15 @@ class PolicyParser
...
+    /// @param jsonNode The JSON object to process.
+    void ReplaceVariablesInJson(nlohmann::json& jsonNode);
 
-    /// @brief Document loaded from the YAML file.
-    std::unique_ptr<IYamlDocument> m_yamlDocument;
+    /// @brief Document loaded and converted from the YAML file.
+    nlohmann::json m_jsonDocument;
+
+    /// @brief Path to the original YAML file.
+    std::filesystem::path m_filename;
```
TomasTurina: Why is this for?

jr0me: I discarded the attempts to remove the second yaml parser. There were some problems with the JSON parsing, the code is convoluted and hard to understand unfortunately, the refactor was promising but it will require some extra time. I'll come back to it afterwards so we don't fall behind with our objectives.

PR #31629, opened 2025-08-28
Prevent upgrading agents with obsolete configurations through WPK

jpcerrone:
## Description

Add functionality in the `upgrade_agents.py` script to look for unsupported configurations (UDP protocol and blowfish encryption method) before upgrading an agent with WPK. If the agent contains one of these  configurations, the upgrade process is aborted and an error message is shown. Closes https://github.com/wazuh/wazuh-agent-packages/issues/436

Update the `upgrade_agents` tests to take into consideration the added logic inside the function. Also adding a new test that validates invalid configurations raise the appropriate exception.

## Tests:
For the following tests I added the unsupported configurations inside a 4.X agent's `ossec.conf` and restarted it. I then launched the `upgrade_agents.py` script in a 5.X manager: `python3 agent_upgrade.py -a 001`
### udp
```
Agents that cannot be upgraded:
        Agent 001 upgrade failed. Status: Error 1761 - The next set of configuration values are no longer supported: [protocol: udp] 
```

### blowfish

```
Agents that cannot be upgraded:
        Agent 001 upgrade failed. Status: Error 1761 - The next set of configuration values are no longer supported: [crypto_method: blowfish]
```

### Both settings
```
Agents that cannot be upgraded:
        Agent 001 upgrade failed. Status: Error 1761 - The next set of configuration values are no longer supported: [protocol: udp], [crypto_method: blowfish]
```

### Since I changed the inactive agent detection logic I also testsed with inactive agents to see that flow still works as expected: 
```
Agents that cannot be upgraded:
        Agent 001 upgrade failed. Status: Error 1707 - Cannot send request, agent is not active
```

### DIFF
--- a/framework/wazuh/agent.py
+++ b/framework/wazuh/agent.py
@@ -1173,7 +1173,7 @@ def get_outdated_agents(agent_list: list = None, offset: int = 0, limit: int = c
     return result
 
 @expose_resources(actions=["agent:upgrade"], resources=["agent:id:{agent_list}"],
-                  post_proc_kwargs={'exclude_codes': [1701, 1703, 1707, 1731] + ERROR_CODES_UPGRADE_SOCKET})
+                  post_proc_kwargs={'exclude_codes': [1701, 1703, 1707, 1731, 1761] + ERROR_CODES_UPGRADE_SOCKET})
 def upgrade_agents(agent_list: list = None, wpk_repo: str = None, version: str = None, force: bool = False,
                    use_http: bool = False, package_type: str = None, file_path: str = None, installer: str = None,
                    filters: dict = None, q: str = None) -> AffectedItemsWazuhResult:


framework/wazuh/agent.py
```diff
@@ -1219,13 +1219,34 @@ def upgrade_agents(agent_list: list = None, wpk_repo: str = None, version: str =
         not_found_agents = agent_list - system_agents
         [result.add_failed_item(id_=agent, error=WazuhResourceNotFound(1701)) for agent in not_found_agents]
 
-        # Add non active agents to failed_items
```
jpcerrone, line 1222: Note: unified the `agent['status'] != 'active'` logic here with what i'm adding to check the config values, so that we don't iterate through `data['items']` twice.

framework/wazuh/core/tests/data/test_agent/schema_global_test.sql
```diff
@@ -80,7 +80,7 @@ INSERT INTO agent (id, name, register_ip, internal_key, os_name, os_version, os_
                    'b3650e11eba2f27er4d160c69de533ee7eed6016fga85ba2455d53a90927747f', 'Ubuntu','16.04.1 LTS','16','04',
                    'Xenial','ubuntu',
                    'Linux |agent-1 |4.15.0-43-generic |#46-Ubuntu SMP Thu Dec 6 14:45:28 UTC 2018 |x86_64','x86_64',
-                   'Wazuh v3.6.2','ab73af41699f13fgt81903b5f23d8d00','f8d49771911ed9d5c45bdfa40babd065','master',
```
jpcerrone, line 83: Note: bumping the version in these files since `get_agent_conf` requires `>v3.7.0`. The other tests that use these values don't test anything in particular about the old version and continue to work correctly after the bump.

framework/wazuh/agent.py
```diff
@@ -1219,13 +1219,36 @@ def upgrade_agents(agent_list: list = None, wpk_repo: str = None, version: str =
...
+                        if extra_message != '':
+                            extra_message += ', '
+                        extra_message += "[crypto_method: blowfish]"
+                    if found_invalid_config:
+                        invalid_config_agents.add(agent['id'])
+                        result.add_failed_item(id_=agent['id'], error=WazuhError(1761, extra_message = extra_message))
 
         # Add non eligible agents to failed_items
-        non_eligible_agents = agent_list - not_found_agents - non_active_agents - filtered_agents
+        non_eligible_agents = agent_list - not_found_agents - non_active_agents - filtered_agents - invalid_config_agents
```
jomafer97: GJ @jpcerrone! I've left a pair comments for review.

Could this line be replaced with the following?

If filtering agents means exclusion, then the expression could be:

```python
non_eligible_agents = not_found_agents | non_active_agents | invalid_config_agents | filtered_agents
```

Or if it means inclusion, it could be:

```python
non_eligible_agents = (not_found_agents | non_active_agents | invalid_config_agents) - filtered_agents
```

Also, when performing the set difference, the elements in the filtered_agents set probably are of a different type than the others; would that work correctly?

```
agent_list = set(agent_list)
filtered_agents = set([agent['id'] for agent in data['items']])
```

If it doesn’t work, it would be necessary to create a set with the filtered agents in order to operate with it later.

framework/wazuh/agent.py
```diff
@@ -1234,7 +1257,7 @@ def upgrade_agents(agent_list: list = None, wpk_repo: str = None, version: str =
             )
         )) for ag in non_eligible_agents]
 
-        eligible_agents = agent_list - not_found_agents - non_active_agents - non_eligible_agents
+        eligible_agents = agent_list - not_found_agents - non_active_agents - non_eligible_agents - invalid_config_agents
```
jomafer97: Since the non-eligible agents were calculated earlier, this expression could perhaps be updated to the following:

```
eligible_agents = agent_list - non_eligible_agents
```

jotacarma90 (dismissed):
LGTM

framework/wazuh/agent.py
jpcerrone: Hi @jomafer97 , thanks for reviewing! Basing myself on your observations, I changed the name of `non_eligible_agents` to `excluded_agents` which better reflects its purpose and simplified some of the set operations.
You mention that the elements in `filtered_agents` (now renamed to `included_agents`)  are of a different type than the others, but this shouldn't be the case since the line:
`filtered_agents = set([agent['id'] for agent in data['items']])` produces a set of strings just like all the other filters. Let me know if this is correct or I'm missing something.

framework/wazuh/agent.py
jpcerrone: I ended up renaming `non_eligible_agents` to `excluded_agents` since it better reflects what it is doing. I think this line should remain as it is since `non_eligible_agents` didn't really include everything that's not eligible.

lchico (approved):
GJ! LGTM!

framework/wazuh/agent.py
jomafer97: Hi @jpcerrone , I've been reviewing the changes and I appreciate that you applied them. 

Regarding the variable type issue, I thought that `agent_list` wasn't a set of strings, my apologies for that.

MarcelKemp (changes requested):
Could we try to improve the readability of the code by simplifying part of it using functions?

For example:
```
  def _validate_agent_configuration(agent_conf: dict) -> list:
      """Validate agent configuration for upgrade compatibility.
      
      Args:
          agent_conf (dict): Agent configuration dictionary
          
      Returns:
          list: List of invalid configuration messages
      """
      invalid_configs = []

      # Check for UDP protocol in any server configuration
      for server in agent_conf.get('client', {}).get('server', []):
          if server.get('protocol') == 'udp':
              invalid_configs.append('[protocol: udp]')
              break

      # Check for blowfish encryption method
      if agent_conf.get('client', {}).get('crypto_method') == 'blowfish':
          invalid_configs.append('[crypto_method: blowfish]')

      return invalid_configs
```

jpcerrone:
> Could we try to improve the readability of the code by simplifying part of it using functions?
> 
> For example:
> 
> ```
>   def _validate_agent_configuration(agent_conf: dict) -> list:
>       """Validate agent configuration for upgrade compatibility.
>       
>       Args:
>           agent_conf (dict): Agent configuration dictionary
>           
>       Returns:
>           list: List of invalid configuration messages
>       """
>       invalid_configs = []
> 
>       # Check for UDP protocol in any server configuration
>       for server in agent_conf.get('client', {}).get('server', []):
>           if server.get('protocol') == 'udp':
>               invalid_configs.append('[protocol: udp]')
>               break
> 
>       # Check for blowfish encryption method
>       if agent_conf.get('client', {}).get('crypto_method') == 'blowfish':
>           invalid_configs.append('[crypto_method: blowfish]')
> 
>       return invalid_configs
> ```

Makes sense! I've made the suggested changes in the last commit and re-run the tests:
```
Agents that cannot be upgraded:
        Agent 001 upgrade failed. Status: Error 1761 - The next set of configuration values are no longer supported: [protocol: udp],[crypto_method: blowfish]
```

lchico (approved):
LGTM!

MarcelKemp:
> You need to review this API integration test
> https://github.com/wazuh/wazuh/blob/09e0a34ff98ed269ebfa0716af7679570c927c99/api/test/integration/test_agent_PUT_endpoints.tavern.yaml#L886
> 
> test_agent_PUT_endpoints.tavern.yaml::PUT /agents/upgrade

Due to various issues in the `main` branch for setting up the testing environment, it has not been possible to fix this problem. Therefore, we have created the following issue, so that it can be fixed in the future when the testing environments have been fixed:
- https://github.com/wazuh/wazuh/issues/31823

@nico-stefani, if you can, please review the PR again.

MarcelKemp (approved):

nico-stefani (approved):
LGTM

Issue #25587, opened 2024-09-05
Add CephFS type to skip_nfs flag


audriusb:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| v4.7.2-40711 | rootcheck(and others)| Agent | APT repo Package | Ubuntu 22.04.3 LTS |

I completely understand why nfs and cifs have an option to be skipped. That is because it's a remote filesystem and multiple or agants scanning the same files is not a good idea. However NFS and CIFS are not the only ones. 

For example: 
I have CephFS volumes mounted on multiple (hundreds) pods in multiple kubernetes clusters with "ReadWriteMany" mode. Meaning they all have that mounted at the same time. So agent being deployed on all nodes just traverses everything under /var/lib with rootcheck.

`/var/lib/kubelet/pods/<UID>/volumes/kubernetes.io~csi/<volume_name>/mount` is mounted multiple times even on the same node.

`lsof` shows a tone of open operations typed `DIR`  I don't want to get into cephfs internals, but this in ends up in slow operations or oversized cache between client and ceph metadata service, since the same operations are coming from a lot of clients and the same time.

Mounting is done from kernel, not fuse. 

Since "yet-another-hardcoded-value" does not look like the best possible solution, my suggestion is to make this configurable somehow. Like there's the skipFS function:
https://github.com/wazuh/wazuh-agent/blob/master/src/common/fs_op/src/fs_op.c#L96
I think this could be configurable. Though the values of FS type are predefined, I don't believe there isn't another way to detect FS type.

vikman90:
Dear @audriusb, 

Thanks for the report and for the PR — and sorry for the delay.

I’m not very familiar with CephFS internals, but I understand the performance impact you’re describing in Kubernetes/ReadWriteMany scenarios, especially when the same volume is mounted across many pods/nodes and Rootcheck ends up traversing it repeatedly.

For cases like NFS and AUFS, we excluded those filesystems because Rootcheck checks are not compatible with them and they could lead to a large number of false positives (and potentially unreliable results).

In your case, I understand the performance concern, but I’d like to confirm whether CephFS scanning is also causing false positives/negatives in Rootcheck results. If it does, then I agree that making this behavior configurable (and/or accepting your approach) makes sense. If it doesn’t, then we may prefer to keep the current behavior and recommend using `<exclude>` to omit paths like:

- `/var/lib/kubelet/pods`

[...]

PR #31846, opened 2025-09-09
feat(#25587): add CephFS type to skip_nfs flag

stevenc99:
[ceph-csi](https://github.com/ceph/ceph-csi) for Kubernetes mounts CephFS filesystems under /var/lib/kubelet/plugins/kubernetes.io/csi/cephfs.csi.ceph.com/*/globalmount
 
Wazuh rootcheck scans include /var/lib by default
https://documentation.wazuh.com/current/user-manual/reference/ossec-conf/rootcheck.html?utm_source=chatgpt.com#readall
 
As a result, Wazuh on Kubernetes hosts will recurse into those network filesystems, which is typically not desired, as they may be mounted by multiple hosts at the same time, as well as potentially causing performance issues.
 
I suggest that skip_nfs include CephFS (but not necessarily Ceph RBD) in addition to CIFS and NFS.

I can confirm that this patch (backported to 4.x) worked as intended for the Linux systems that I'm administrator for. This was confirmed by manually triggering a rootcheck via
`/var/ossec/bin/agent_control -r`
and observing the scanned paths on the monitored hosts with
`strace -f -e trace=lstat -p $(pidof wazuh-syscheckd)`
 
Closes https://github.com/wazuh/wazuh/issues/25587

### DIFF
--- a/src/shared/fs_op.c
+++ b/src/shared/fs_op.c
@@ -27,6 +27,7 @@
 #define SYSFS       0x62656572
 #define OVERLAYFS   0x794c7630
 #define BTRFS       0x9123683E
+#define CEPHFS      0x00C36400
 #define CIFS        0xFF534D42
 #define V9FS        0x01021997
 #define ST_NODEV    4
@@ -43,6 +44,7 @@
 const struct file_system_type network_file_systems[] = {
 #ifdef __linux__
     {.name="NFS",  .f_type=NFS, .flag=1},
+    {.name="CEPHFS", .f_type=CEPHFS, .flag=1},
     {.name="CIFS", .f_type=CIFS, .flag=1},
 #endif
     /*  The last entry must be name=NULL */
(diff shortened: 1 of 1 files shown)


vikman90:
Dear @stevenc99,

Thank you for the proposal and for taking the time to work on this change.

As explaioned at https://github.com/wazuh/wazuh/issues/25587#issuecomment-3943477661, for the CephFS/Kubernetes ReadWriteMany use case, we think the most appropriate approach is to **exclude the affected paths explicitly** (e.g., via Rootcheck `<exclude>` for directories like `/var/lib/kubelet/pods` and related volume mount paths), rather than adding or expanding a **filesystem-type-based exclusion**.

Since this can be handled through configuration and avoids introducing additional hardcoded filesystem logic, we’re going to **close this pull request**.

Thanks again for your contribution.

PR #32073, opened 2025-09-22
Merge 4.14.1 into main

TomasTurina:
This issue aims to merge the branch `4.14.1` into the branch `main`


MiguelazoDS:
Error

image

It works in Ubuntu 24 due to higher version, we need 24.04 for main, so this may work without changes for 5.0.0

lcov: LCOV version 2.0-1

Conditionally add the option for version 2.0-1 or greater 

This should fix all except inventory sync error that I think is waiting for a job that didn't finished, maybe adding a dependency between jobs?

## Update 

I just noticed that the dependency is already added, this may be related to a regex in the jq command 

image

PR #32377, opened 2025-09-30
Preliminary support new OSs - Mac OS 26 Tahoe - Add SCA content

Johnng007:
# Description

| Related issue | 
|---|
| #30436 | 

| Agent tier | Central components support | OS type |
|-|-|-|
| 1 | No | Major |

# Plan

**ThreatIntel**
- [ ] Define a plan to support the new OS, particularly with regard to SCA policies. If there's no official CIS policy for that OS version, we either use an existing draft or we adapt a previous version.


AwwalQuan:
There are CIS benchmarks specifically for macOS 26 Tahoe: https://www.cisecurity.org/benchmark/apple_os
 We need to create new SCA content based on this.

PR #32383, opened 2025-09-30
Avoid indefinitive wait at whodata health_check

### LINKED ISSUE #21310: Indefinitive wait detected at FIM whodata health_check
The following defect related to FIM's run_check have been reported by Coverity:

- **CID 1575403**: Indefinite wait (BAD_CHECK_OF_WAIT_COND)
wait_cond_improperly_checked: The wait condition prompting the wait upon audit_hc_mutex is not checked correctly. This code can wait for a condition that has already been satisfied, which can cause a never-ending wait.
https://github.com/wazuh/wazuh/blob/b28ebb9172d16e417ddc78ba9e73e2fc2a901418/src/syscheckd/src/whodata/audit_healthcheck.c#L89

mjcr99:
|**Related Issue**|
|---|
|https://github.com/wazuh/wazuh/issues/21310|

## Description

Hi team,

This PR wraps `pthread_cond_timedwait()` in a loop to deal with the following situations:

* Spurious wakeups: `pthread_cond_wait()`/`pthread_cond_timedwait()` may return even if no thread actually signaled the condition variable.
* Race conditions: even if a signal was sent, another thread might have consumed the state change before this thread resumes.

For these reasons, the standard requires re-checking the condition in a loop after every wakeup.

### DIFF
--- a/src/syscheckd/src/whodata/audit_healthcheck.c
+++ b/src/syscheckd/src/whodata/audit_healthcheck.c
@@ -12,6 +12,7 @@
 
 atomic_int_t audit_health_check_creation = ATOMIC_INT_INITIALIZER(0);
 atomic_int_t hc_thread_active = ATOMIC_INT_INITIALIZER(0);
+atomic_int_t hc_thread_finished = ATOMIC_INT_INITIALIZER(0);
 
 pthread_mutex_t audit_hc_mutex;
 pthread_cond_t audit_hc_cond;
@@ -27,6 +28,9 @@ int audit_health_check(int audit_socket) {
     struct timespec wait_time = {0, 0};
 
     w_mutex_init(&audit_hc_mutex, NULL);
+    
+    // Reset the finished flag for this health check
+    atomic_int_set(&hc_thread_finished, 0);
 
     // Audit needs an absolute path to add rules
     abspath(AUDIT_HEALTHCHECK_DIR, abs_path_healthcheck, PATH_MAX);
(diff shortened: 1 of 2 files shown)


src/syscheckd/src/whodata/audit_healthcheck.c
jotacarma90, line 89: `hc_thread_active` is set to 0 before your while conditional, and this is equivalent to having a while (true)
So I understand that we would need another flag like `hc_thread_active`, but to notify that the secondary thread has finished, something like `hc_thread_finished` (because we are trying to use the same flag both for the main thread to notify the secondary thread to finish, and for the secondary thread to notify the main thread that it has finished, and that is not possible).

Proposal, new flag:
```
static atomic_int_t hc_thread_finished = ATOMIC_INT_INITIALIZER(0);
```
- Use it in the while loop you have marked:
```
while (atomic_int_get(&hc_thread_finished) == 0) {
```
- And that the secondary thread sets it when it's finished:
```
void *audit_healthcheck_thread(int *audit_sock) {
    w_mutex_lock(&audit_hc_mutex);
    atomic_int_set(&hc_thread_active, 1);
    w_cond_signal(&audit_hc_cond);
    w_mutex_unlock(&audit_hc_mutex);

    mdebug2(FIM_HEALTHCHECK_THREAD_ACTIVE);

    audit_read_events(audit_sock, &hc_thread_active);

    mdebug2(FIM_HEALTHCHECK_THREAD_FINISHED);

    w_mutex_lock(&audit_hc_mutex);
    atomic_int_set(&hc_thread_finished, 1);
    w_cond_signal(&audit_hc_cond);
    w_mutex_unlock(&audit_hc_mutex);

    return NULL;
}
```



Extra:
- We should analyze whether setting the aforementioned variable to 0 should also be included within the corresponding locks:
```
w_mutex_lock(&audit_hc_mutex);
atomic_int_set(&hc_thread_active, 0);
w_mutex_unlock(&audit_hc_mutex);
```

jotacarma90 (changes requested):

src/syscheckd/src/whodata/audit_healthcheck.c
mjcr99: Hi @jotacarma90, I think this new implementation properly solves the case you describe. Please review it when possible:
* https://github.com/wazuh/wazuh/pull/32383/commits/1b056445ca7c009fa4efde542531c684cc855f7c

jotacarma90 (approved):
LGTM

src/syscheckd/src/whodata/audit_healthcheck.c
```diff
@@ -80,13 +84,21 @@ int audit_health_check(int audit_socket) {
...
     w_mutex_lock(&audit_hc_mutex);
     gettime(&wait_time);
     wait_time.tv_sec += 5;
-    pthread_cond_timedwait(&audit_hc_cond, &audit_hc_mutex, &wait_time);
+    while (atomic_int_get(&hc_thread_finished) == 0) {
+        if (pthread_cond_timedwait(&audit_hc_cond, &audit_hc_mutex, &wait_time) != 0) {
+            // Timeout or error occurred
+            break;
+        }
+    }
```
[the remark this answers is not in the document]
mjcr99: I think this change is not needed.

Antoniogm03 (approved):
LGTM

lchico (approved):
LGTM!

MarcelKemp (approved):

PR #32389, opened 2025-10-01
Fix `RSA` decryption validation for `OpenSSL 3.5.1` compatibility

### LINKED ISSUE #32380: Component test failure: KeyStoreComponentTest.TestUpgradeFail in keystore_component
## Summary

The unit test `KeyStoreComponentTest.TestUpgradeFail` in the `keystore_component` is failing.

## Failure details
```
[ RUN      ] KeyStoreComponentTest.TestUpgradeFail
Upgrading 'key1' key pair.
Decryption successful for key: 'key1'
Key pair 'key1' upgraded.
/src/src/shared_modules/keystore/tests/component/keyStoreComponent_test.cpp:146: Failure
  out
    Which is: "\x9F\x8A\x92\x86\xC0?JU7\x8E\xCEH*_ \x9E\xF8" "5Q>/n\xC9\xA0\xA2w01\xC1K\xB3W\x1+\x87Y\x93t\x97Q\xD4\x86\xD8\xCA\x12" "e\xE9\x17\xE8u\x83" "4_-\x5\xBD\x81\x84\x14\xAD" "B\xE4\x1E" "2\x93\x84\xF1" "B\xFC\xC3\x80p\xEE=\xB3\xC1" "4,\x82" "f\xD2\xD1nW\x3\xDF\xE8\xF8\xF7" "9/M-\xA4\x9F*\x16\xFD\x16\xFC\xCC\xEE%\xD9\xCA\xC3\xE7u\xBF\t\"\xAD-\xF2\x88" "Ds\xE1\xB2\x1E\xDD\xF7\xEE" "B]\x98lxwk\x93\xE3\xE8N\xB9\xD1\xFn\xB6;\xB2\x86\xA3\x14\x5" "f_M4\xCB\xBC\xA0\xD1\xE4\x98\x8A.I\xC4\xB8\xCA!\xDE" "Afno\x10LM\x91\x13*)J\xBA\x96\xCBM\x89\xE1\x16\xA8\x94\xB0" "fU\xA3\xC3,P\x9Bp\xD3\xFB\x99\x13\xFA\xD3\x87\x94\xE7\x9A\xF2\xA2"
  ""
[  FAILED  ] KeyStoreComponentTest.TestUpgradeFail (116 ms)
```

## Steps to reproduce

1. Navigate to the `src` directory.
2. Run `make deps`.
3. Run `make TARGET=server DEBUG=1 TEST=1`.
4. Navigate to `build`.
5. Run `shared_modules/keystore/tests/component/keystore_component_tests`.

## Environment

|Wazuh version|Component|Platform|
|---|---|---|
| 4.14.0-alpha1 | keystore_component | Ubuntu 22.04 @ Docker |

GabrielEValenzuela:
Closes #32380 

# Summary

This change fixes keystore upgrade failure detection after upgrading from OpenSSL 3.0.12 to 3.5.1, invalid RSA ciphertext was no longer consistently rejected, which caused corrupt data to be stored instead of triggering the expected upgrade failure path.

## Problem statement

The `TestUpgradeFail` unit test failed after upgrading to OpenSSL 3.5.1. This test verifies that the keystore upgrade process correctly handles invalid (non-RSA-encrypted) data by:

1. Detecting the decryption failure.
2. Deleting all corrupted keys.
3. Allowing new keys to be stored with the updated AES-256 scheme.

**Observed behavior (OpenSSL 3.5.1):**

* `RSA_private_decrypt()` returned a positive value even for invalid input.
* No exception was thrown during the upgrade.
* Garbage data was AES-encrypted and stored.
* The test assertion failed: expected empty string, received encrypted garbage.

**Expected behavior (OpenSSL 3.0.12):**

* `RSA_private_decrypt()` returned `-1` with “pkcs decoding error” for invalid input.
* An exception was thrown.
* The keystore was cleared and version updated.
* The test passed.

## Possible RCA

According to the OpenSSL documentation, `RSA_private_decrypt` may accept ciphertext shorter than `RSA_size(key)` by interpreting it as zero-padded. In OpenSSL 3.0.12, invalid inputs in practice triggered an immediate “pkcs decoding error.” In OpenSSL 3.5.1, the same invalid inputs can be tolerated long enough to return a non-negative result with meaningless plaintext, rather than producing a hard error.

This behavioral difference caused the keystore upgrade logic to incorrectly treat invalid data as valid and continue with AES re-encryption of garbage values.

## Solution

The RSA decryption logic in `rsaHelper.hpp` has been hardened with explicit validation:

### 1. Ciphertext Size Validation

Ensure input size matches the expected RSA ciphertext size before attempting decryption:

```cpp
if (input.length() != static_cast<size_t>(rsaSize)) {
    throw std::runtime_error("Invalid RSA ciphertext size. Expected: " +
                             std::to_string(rsaSize) +
                             ", Got: " + std::to_string(input.length()));
}
```

### 2. Error Queue Inspection

Clear the OpenSSL error queue before decryption and check it afterwards to detect silent failures:

```cpp
T::ERR_clear_error();
const auto decryptedLen = T::RSA_private_decrypt(...);

unsigned long err = T::ERR_get_error();
if (decryptedLen < 0 || err != 0) {
    throw std::runtime_error("RSA decryption failed");
}
```

### 3. Output Length Validation

Reject decryption results that produce zero-length plaintext:

```cpp
if (decryptedLen == 0) {
    throw std::runtime_error("RSA decryption produced zero-length output");
}
```

## Changes Made

* **`rsaHelper.hpp`**: Enhanced `rsaDecrypt()` with ciphertext size validation, error queue inspection, and output length checks.
* **`opensslPrimitives.hpp`**: Added wrappers for `ERR_clear_error()` and `ERR_peek_error()` to support better diagnostics.

### Unit Tests

* `KeyStoreComponentTest.TestUpgradeFail` now passes with OpenSSL 3.5.1:

  * Invalid input is rejected immediately.
  * Keystore is cleared on upgrade failure.
  * New keys can be added after the failed upgrade.

### DIFF
--- a/src/shared_modules/utils/opensslPrimitives.hpp
+++ b/src/shared_modules/utils/opensslPrimitives.hpp
@@ -150,6 +150,16 @@ protected:
     {
         return ::ERR_reason_error_string(e);
     }
+
+    inline void ERR_clear_error(void)
+    {
+        return ::ERR_clear_error();
+    }
+
+    inline unsigned long ERR_peek_error(void)
+    {
+        return ::ERR_peek_error();
+    }
 #pragma GCC diagnostic pop
 };
 
--- a/src/shared_modules/utils/opensslPrimitives.hpp
+++ b/src/shared_modules/utils/opensslPrimitives.hpp
@@ -151,9 +151,12 @@ protected:
         return ::ERR_reason_error_string(e);
     }
 
+    // From OpenSSL 3.9.1 we need to use ERR_clear_last_error and ERR_peek_last_error on this
+    // wrapper
+
     inline void ERR_clear_error(void)
     {
-        return ::ERR_clear_error();
+        ::ERR_clear_error();
     }
 
     inline unsigned long ERR_peek_error(void)
--- a/src/shared_modules/utils/rsaHelper.hpp
+++ b/src/shared_modules/utils/rsaHelper.hpp
@@ -86,10 +86,23 @@ public:
 
         createRSA(rsa, filePath, RSA_PRIVATE);
 
+        const int rsaSize = T::RSA_size(rsa);
+
+        // Input validation: RSA ciphertext must be exactly RSA_size bytes
+        if (input.length() != static_cast<size_t>(rsaSize))
+        {
+            T::RSA_free(rsa);
+            throw std::runtime_error("Invalid RSA ciphertext size. Expected: " + std::to_string(rsaSize) +
+                                     ", Got: " + std::to_string(input.length()));
+        }
+
         std::string decryptedText(T::RSA_size(rsa), 0); // Initialize with zeros
 
         // Defered free
         DEFER([&]() { T::RSA_free(rsa); });
+
+        // Clear error queue
+        T::ERR_clear_error();
 
         // Decrypt the ciphertext using RSA private key
         const auto decryptedLen = T::RSA_private_decrypt(input.length(),
@@ -98,13 +111,23 @@ public:
                                                          rsa,
                                                          RSA_PKCS1_PADDING);
 
-        if (decryptedLen < 0)
+        unsigned long err = T::ERR_get_error();
+
+        if (decryptedLen < 0 || err != 0)
         {
-            throw std::runtime_error("RSA decryption failed: " +
-                                     std::string(T::ERR_reason_error_string(T::ERR_get_error())));
+            std::string errMsg = "RSA decryption failed";
+            if (err != 0)
+            {
+                errMsg += ": " + std::string(T::ERR_reason_error_string(err));
+            }
+            throw std::runtime_error(errMsg);
         }
 
-        // Display the decrypted plaintext
+        if (decryptedLen == 0)
+        {
+            throw std::runtime_error("RSA decryption produced zero-length output");
+        }
+
         output = decryptedText.substr(0, decryptedLen);
 
         return decryptedLen;
--- a/src/shared_modules/utils/rsaHelper.hpp
+++ b/src/shared_modules/utils/rsaHelper.hpp
@@ -110,11 +110,9 @@ public:
                                                          reinterpret_cast<unsigned char*>(decryptedText.data()),
                                                          rsa,
                                                          RSA_PKCS1_PADDING);
-
-        unsigned long err = T::ERR_get_error();
-
-        if (decryptedLen < 0 || err != 0)
+        if (decryptedLen < 0)
         {
+            unsigned long err = T::ERR_get_error();
             std::string errMsg = "RSA decryption failed";
             if (err != 0)
             {
--- a/src/shared_modules/utils/tests/rsaHelper_test.cpp
+++ b/src/shared_modules/utils/tests/rsaHelper_test.cpp
@@ -14,10 +14,12 @@
 #include <gmock/gmock.h>
 
 using testing::_;
+using testing::DoAll;
 using testing::HasSubstr;
 using testing::Return;
 using testing::ReturnNull;
 using testing::StrEq;
+using testing::WithArgs;
 
 constexpr auto KEYFILE {"keyfilePath"};
 constexpr auto KEY {"AbCd1234"};
@@ -345,11 +347,11 @@ TEST_F(RSAHelperTest, rsaDecryptErrorInQueue)
     EXPECT_CALL(rsaHelper, RSA_size((RSA*)2)).Times(2).WillRepeatedly(Return(strlen(KEY)));
     EXPECT_CALL(rsaHelper, ERR_clear_error()).WillOnce(Return());
 
-    // Simulate OpenSSL 3.5.1 behavior: returns positive but has error in queue
+    // Simulate OpenSSL 3.5.1 behavior: returns negative with error in queue
     EXPECT_CALL(rsaHelper, RSA_private_decrypt(strlen(KEY), _, _, (RSA*)2, RSA_PKCS1_PADDING))
-        .WillOnce(Return(10)); // Positive return value
+        .WillOnce(Return(-1)); // Negative return value indicates error
 
-    // But error queue has an error
+    // Error queue has an error
     EXPECT_CALL(rsaHelper, ERR_get_error()).WillOnce(Return(42));
     EXPECT_CALL(rsaHelper, ERR_reason_error_string((unsigned long)42)).WillOnce(Return("Padding check failed"));
     EXPECT_CALL(rsaHelper, RSA_free((RSA*)2)).WillOnce(Return());
@@ -376,10 +378,10 @@ TEST_F(RSAHelperTest, rsaDecryptZeroLengthOutput)
     EXPECT_CALL(rsaHelper, RSA_size((RSA*)2)).Times(2).WillRepeatedly(Return(strlen(KEY)));
     EXPECT_CALL(rsaHelper, ERR_clear_error()).WillOnce(Return());
 
-    // Returns 0 length (invalid)
+    // Returns 0 length (invalid) - this should be treated as negative/error
     EXPECT_CALL(rsaHelper, RSA_private_decrypt(strlen(KEY), _, _, (RSA*)2, RSA_PKCS1_PADDING)).WillOnce(Return(0));
 
-    EXPECT_CALL(rsaHelper, ERR_get_error()).WillOnce(Return(0)); // No error in queue
+    // No need for ERR_get_error() call since we check decryptedLen == 0 separately
     EXPECT_CALL(rsaHelper, RSA_free((RSA*)2)).WillOnce(Return());
 
     try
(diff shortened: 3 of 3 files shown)


src/shared_modules/utils/opensslPrimitives.hpp
```diff
@@ -150,6 +150,16 @@ class OpenSSLPrimitives
...
     }
+
+    inline void ERR_clear_error(void)
+    {
+        return ::ERR_clear_error();
+    }
+
+    inline unsigned long ERR_peek_error(void)
+    {
+        return ::ERR_peek_error();
```
MiguelazoDS, line 164: Do we need this?

src/shared_modules/utils/rsaHelper.hpp
```diff
@@ -86,25 +86,48 @@ class TRSAHelper final
...
+            if (err != 0)
+            {
+                errMsg += ": " + std::string(T::ERR_reason_error_string(err));
+            }
+            throw std::runtime_error(errMsg);
+        }
+
+        if (decryptedLen == 0)
+        {
+            throw std::runtime_error("RSA decryption produced zero-length output");
```
MiguelazoDS, line 126: > A return value of 0 is not an error and means only that the plaintext was empty.

Agree with this

src/shared_modules/utils/rsaHelper.hpp
```diff
@@ -86,25 +86,48 @@ class TRSAHelper final
...
         const auto decryptedLen = T::RSA_private_decrypt(input.length(),
                                                          reinterpret_cast<const unsigned char*>(input.data()),
                                                          reinterpret_cast<unsigned char*>(decryptedText.data()),
                                                          rsa,
                                                          RSA_PKCS1_PADDING);
 
-        if (decryptedLen < 0)
+        unsigned long err = T::ERR_get_error();
+
+        if (decryptedLen < 0 || err != 0)
```
MiguelazoDS: Isn't this redundant? won't ERR_get_error returns something different than zero only for decryptedLen < 0?

MiguelazoDS:
From documentation this seems correct.

> Decryption failures in the RSA_PKCS1_PADDING mode leak information which can potentially be used to mount a Bleichenbacher padding oracle attack. This is an inherent weakness in the PKCS #1 v1.5 padding design. Prefer RSA_PKCS1_OAEP_PADDING.

> In OpenSSL before version 3.2.0, both the return value and the length of returned value could be used to mount the Bleichenbacher attack. Since version 3.2.0, the default provider in OpenSSL does not return an error when padding checks fail. Instead it generates a random message based on used private key and provided ciphertext so that application code doesn't have to implement a side-channel secure error handling. Applications that want to be secure against side-channel attacks with providers that don't implement implicit rejection, still need to handle the returned values using side-channel free code. Side-channel free handling of the error stack can be performed using either a pair of unconditional [ERR_set_mark(3)](https://docs.openssl.org/master/man3/ERR_set_mark/) and [ERR_pop_to_mark(3)](https://docs.openssl.org/master/man3/ERR_set_mark/) calls or by using the [ERR_clear_error(3)](https://docs.openssl.org/master/man3/ERR_clear_error/) call.

Now that we properly handle the non-zero return value, exceptions are thrown and failing test is passing. 

## Manual testing 

Adding invalid credentials that will fail to decrypt in a old server version

```console
root@jammy:/home/vagrant/wazuh# rocks_db_query_testtool -d /var/ossec/queue/keystore/ -c indexer
password ==> aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
username ==> aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
version ==> 2
```

```console
2025/10/03 18:08:47 wazuh-modulesd:vulnerability-scanner: ERROR: VulnerabilityScannerFacade::start: RSA decryption failed
```

image

Upgrade to 4.14.0 

I think this condition here is preventing to see the error we saw before (RSA decryption failed) since now it won't return any decryptedLen < 0. 

```cpp
if (decryptedLen < 0 || err != 0)
```

image

Instead I think it returned false for all the conditionals here 

```cpp
        if (decryptedLen < 0 || err != 0)
        {
            std::string errMsg = "RSA decryption failed";
            if (err != 0)
            {
                errMsg += ": " + std::string(T::ERR_reason_error_string(err));
            }
            throw std::runtime_error(errMsg);
        }

        if (decryptedLen == 0)
        {
            throw std::runtime_error("RSA decryption produced zero-length output");
        }
```

image

Besides that, the functionality is as expected, I can change the value now using AES encryption 

I can set new credentials

```console
echo 'newuser' | /var/ossec/bin/wazuh-keystore -f indexer -k username
echo 'newpass' | /var/ossec/bin/wazuh-keystore -f indexer -k password
```

and get the values

image

> [!NOTE]
> We can discuss the comments I left, but for me it's working.

src/shared_modules/utils/opensslPrimitives.hpp
GabrielEValenzuela: Yes - this is required for the template class to compile. We have a template implementation, so we need this wrapper to call the `OpenSSL` function during template instantiation. Without it, we'll get linker errors.

src/shared_modules/utils/rsaHelper.hpp
GabrielEValenzuela: You're right! That in normal operation, `ERR_get_error()` will only return non-zero when `decryptedLen < 0`. However, we need to call `ERR_get_error()` **before** the condition check because:

I changed to:

```cpp
if (decryptedLen < 0)
{
    unsigned long err = T::ERR_get_error();
    std::string errMsg = "RSA decryption failed";
    if (err != 0)
    {
        errMsg += ": " + std::string(T::ERR_reason_error_string(err));
    }
    throw std::runtime_error(errMsg);
}
```

This makes it clearer that we're retrieving the error only when needed, and the redundancy is eliminated.

MiguelazoDS:
@GabrielEValenzuela please take a look at the comment here https://github.com/wazuh/wazuh/pull/32389#issuecomment-3367100668, is that the expected behavior? the error message does not appear because no exception is triggered.

GabrielEValenzuela:
# Update

We made a meet with @MiguelazoDS and discuss about this. The error now works in silence mode, due the `Bleichenbacher attack` that `OpenSSL` solves on `3.2.0`, in case of another type of error, we perform the logging to provide further information.

CC @MarcelKemp

src/shared_modules/utils/tests/rsaHelper_test.cpp
```diff
@@ -301,19 +310,119 @@ TEST_F(RSAHelperTest, rsaDecryptDecryptionFailed)
...
+    EXPECT_CALL(rsaHelper, PEM_read_RSAPrivateKey((FILE*)1, NULL, NULL, NULL)).WillOnce(Return((RSA*)2));
+    EXPECT_CALL(rsaHelper, RSA_size((RSA*)2)).Times(2).WillRepeatedly(Return(10));
+    EXPECT_CALL(rsaHelper, ERR_clear_error()).WillOnce(Return());
+
+    EXPECT_CALL(rsaHelper, RSA_private_decrypt(10, _, _, (RSA*)2, RSA_PKCS1_PADDING))
+        .WillOnce(DoAll(WithArgs<2>(
+                            [](unsigned char* outBuf)
+                            {
+                                const unsigned char fakeData[10] = {'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J'};
+                                memcpy(outBuf, fakeData, 10);
```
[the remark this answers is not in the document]
GabrielEValenzuela: Solved

src/shared_modules/utils/opensslPrimitives.hpp
```diff
@@ -150,6 +150,16 @@ class OpenSSLPrimitives
     {
         return ::ERR_reason_error_string(e);
     }
+
+    inline void ERR_clear_error(void)
+    {
+        ::ERR_clear_error();
+    }
+
+    inline unsigned long ERR_peek_error(void)
```
MarcelKemp, line 162: Let us add a comment to clarify the reason why this function is created.

Issue #31716, opened 2025-09-03
Enhance update of vulnerabilities index mappings on upgrades


fdalmaup:
## Description

It was found that the index mappings update mechanism in the Server is designed for the addition of new fields only, and not the update of changed field types. Updating the index mappings will not change the indexed data, as mentioned in [OpenSearch's documentation](https://docs.opensearch.org/latest/api-reference/index-apis/put-mapping/):

> You can’t use this operation to update mappings that already map to existing data in the index. You must first create a new index with your desired mappings, and then use the [reindex API operation](https://docs.opensearch.org/latest/opensearch/reindex-data) to map all the documents from your old index to the new index.

Therefore, we need to change our current strategy to avoid this partial solution. The alternative proposed by @AlexRuiz7 is to re-index the data using the new mappings, but not always, so the following steps should be implemented:

1. Retrieve the mappings section from the template stored in the Wazuh Server's file system and generate a checksum out of it.
2. Evaluate if a re-index operation is needed, comparing both checksums and making a decision depending on whether:
a. The checksums are **equal**: the index mappings are up-to-date, and no re-index of data is needed.
b. The checksums are **different**: the index mappings are outdated, and re-indexing the data is required, so all documents use the same types.

The reindexation process involves the following steps:
1. Clone the index with a different name.
2. Delete the original index (to allow re-creating the index under the same name in the next step).
3. Re-index the data.

## Tasks

- [x] Extend the index mappings update mechanism.
- [x] Manual tests and proof of expected behavior.
- [x] Stress tests: verify the solution's performance in an inventory with thousands of documents to be reindexed.
- [x] Update of related unit and integration tests.

MiguelazoDS:
# Update 

I started analyzing the issue and the proposal but I needed to address higher priority issues. 

I couldn't implement anything here. Issue on hold

MiguelazoDS:
Testing in 4.8.0 i was expecting to not retrieve vulnerabilities using aggregation due to host.os.full being in text 

```console
curl -k -u admin:admin https://127.0.0.1:9200/wazuh-states-vulnerabilities-*/_search?size=1 -H 'Content-Type: application/json' --data-binary '@aggs.json' | jq
```

```json
{
  "aggs": {
    "top_os_by_vulnerabilities": {
      "terms": {
        "field": "host.os.full",
        "size": 10
      },
      "aggs": {
        "top_example": {
          "top_hits": {
            "_source": [
              "vulnerability.id",
[...]

PR #32463, opened 2025-10-06
Reindex documents

MiguelazoDS:

  - [ ] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...


src/init/update.sh
GabrielEValenzuela, line 1: Good catch!

src/shared_modules/indexer_connector/src/indexerConnector.cpp
```diff
@@ -440,6 +441,217 @@ void IndexerConnector::diff(const nlohmann::json& responseJson,
...
+
+        if (currentMappings.is_discarded())
+        {
+            throw std::runtime_error("Couldn't retrieve current mappings.");
+        }
+
+        // Calculating hashes.
+        auto hashTemplateMappings = hashMappings(templateMappings.dump());
+        auto hashCurrentMappings = hashMappings(currentMappings[m_indexName]["mappings"].dump());
+        if (hashTemplateMappings != hashCurrentMappings)
```
MarcelKemp, line 490: Could the following edge case occur?
1. `validateMappings()` detects different mappings
2. Creates backup successfully
3. Deletes main index
4. ‘_CRASH_’ before completing reindex
  5. Upon restart:
     - `initialize()` creates empty index with correct template
     - `validateMappings()` compares: empty index vs template
     - Hashes match → NO reindexing
     - Backup is orphaned with all data

src/shared_modules/indexer_connector/src/indexerConnector.cpp
```diff
@@ -440,6 +441,217 @@ void IndexerConnector::diff(const nlohmann::json& responseJson,
...
+                                                            .secureCommunication = secureCommunication},
+                                         PostRequestParameters {.onSuccess = onSuccess, .onError = onError},
+                                         ConfigurationParameters {});
+            auto end = std::chrono::high_resolution_clock::now();
+
+            logInfo(IC_NAME,
+                    "It tooks '%ld' seconds to reindex the index '%s'.",
+                    std::chrono::duration_cast<std::chrono::seconds>(end - start).count(),
+                    m_indexName.c_str());
+
```
MarcelKemp, line 591: There is no post-reindex count validation. Could this be necessary?

MiguelazoDS: We need to take into account that due to the re-sync logic after the reindex will oscillate but the manual testing to verify the number of documents remains the same was performed here 

https://github.com/wazuh/wazuh/issues/31716#issuecomment-3378513594

I didn't want to add so many checks for this, if some documents fails I think we can know about that after the logging improvements issue is merged. 

The sync logic in the indexer connector should repair that case if eventually a document is lost 

https://github.com/wazuh/wazuh/blob/467593cf8c5592d884bb92f3a95895f111f76287/src/shared_modules/indexer_connector/src/indexerConnector.cpp#L411

src/shared_modules/indexer_connector/src/indexerConnector.cpp
MiguelazoDS: Before analyzing this case I want to let you know, that in case of manager restart, we can not interrupt this sequence (I think I can show better evidence for this case) 

Now, for this case, the 4th step will occur here 

```cpp
            HTTPRequest::instance().post(RequestParameters {.url = HttpURL(selector->getNext() + "/_reindex"),
                                                            .data = reindexData,
                                                            .secureCommunication = secureCommunication},
                                         PostRequestParameters {.onSuccess = onSuccess, .onError = onError},
                                         ConfigurationParameters {});
```
if the reindex endpoint call fails, it will leave a backup index with the documents from the original index, and the original index will no longer exist. 

Before the restart, I think the fallback mechanism may fail if the index does not exist.  

But related to the scenario you mention, if the reindex doesn't occur, the hashes will never match, so it will execute the method again, remove the dangling backup and retry the reindex. 

So I think this won't happen 

> Hashes match → NO reindexing
> Backup is orphaned with all data

In any case if the index is empty it should recover from that with the sync() method (assuming the information is the persistent storage). I'll perform more testing for this and let you know.

PR #32666, opened 2025-10-16
Add socket-based keystore wrapper

GabrielEValenzuela:
## Description

This PR introduces a **Keystore socket server** that enables secure, language-agnostic access to the Wazuh Keystore from Python and other external clients via Unix domain sockets.

The implementation integrates directly with the existing `SocketServer` infrastructure (no wrapper class needed), providing:
- A lightweight socket listener that receives queries (e.g., `PUT|credentials|apikey|secret`)
- Direct execution of Keystore operations (encryption/decryption handled by existing Keystore)
- JSON responses with operation results

This design maintains security through file system permissions while enabling flexible, multi-language integration without shared library dependencies or ABI compatibility concerns.

### Added
* **Keystore socket server** - Direct integration with `SocketServer` for Keystore IPC access
* **Socket protocol** - Simple pipe-delimited format for operations
* **JSON response format** - Structured responses with status, operation type, and values

### Testing

On issue

### DIFF
--- a/src/shared_modules/keystore/src/keyStore.cpp
+++ b/src/shared_modules/keystore/src/keyStore.cpp
@@ -18,6 +18,12 @@ constexpr auto KS_NAME {"keystore"};
 constexpr auto KS_VERSION {"2"};
 constexpr auto KS_VERSION_FIELD {"version"};
 
+namespace Log
+{
+    std::function<void(const int, const char*, const char*, const int, const char*, const char*, va_list)>
+        GLOBAL_LOG_FUNCTION;
+};
+
 static void upgrade(Utils::RocksDBWrapper& keystoreDB, const std::string& columnFamily)
 {
     std::string versionValue;
--- a/src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
+++ b/src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
@@ -15,9 +15,11 @@
 #include "agentSession.hpp"
 #include "flatbuffers/include/agentInfo_generated.h"
 #include "flatbuffers/include/inventorySync_generated.h"
+#include "keyStore.hpp"
 #include "loggerHelper.h"
 #include "routerSubscriber.hpp"
 #include "singleton.hpp"
+#include "socketServer.hpp"
 #include "stringHelper.h"
 #include <asyncValueDispatcher.hpp>
 #include <filesystem>
@@ -37,6 +39,7 @@ constexpr int DEFAULT_TIME {60 * 10}; // 10 minutes
 constexpr auto INVENTORY_SYNC_PATH {"inventory_sync"};
 constexpr auto INVENTORY_SYNC_TOPIC {"inventory-states"};
 constexpr auto INVENTORY_SYNC_SUBSCRIBER_ID {"inventory-sync-module"};
+constexpr auto SOCKET_KEYSTORE_PATH {"/var/ossec/queue/sockets/keystore"};
 
 using WorkersQueue = Utils::AsyncValueDispatcher<std::vector<char>, std::function<void(const std::vector<char>&)>>;
 using IndexerQueue = Utils::AsyncValueDispatcher<Response, std::function<void(const Response&)>>;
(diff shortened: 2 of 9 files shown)


src/shared_modules/keystore/CMakeLists.txt
```diff
@@ -39,8 +39,10 @@ link_directories(${SRC_FOLDER})
 link_directories(${SRC_FOLDER}/external/rocksdb/build)
 
 add_library(keystore STATIC ${KEYSTORE_SRC})
+add_library(pykeystore_wrap SHARED ${KEYSTORE_SRC} src/keyStoreWrapper.cpp)
 
 target_link_libraries(keystore rocksdb wazuhext)
+target_link_libraries(pykeystore_wrap rocksdb wazuhext)
```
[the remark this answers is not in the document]
GabrielEValenzuela: Done!

GabrielEValenzuela:
> [!NOTE] 
> Expected fail on RTR check

src/shared_modules/keystore/src/keyStore.cpp
```diff
@@ -18,6 +18,12 @@ constexpr auto KS_NAME {"keystore"};
 constexpr auto KS_VERSION {"2"};
 constexpr auto KS_VERSION_FIELD {"version"};
 
+namespace Log
+{
+    std::function<void(const int, const char*, const char*, const int, const char*, const char*, va_list)>
+        GLOBAL_LOG_FUNCTION;
+};
+
```
MiguelazoDS, line 26: I understand the change, but was this strictly necessary for this implementation? It's ok if we are sure it will always work, I see that the indexerconnector testtool already initializes this variable.

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
MiguelazoDS, line 1: Don't we need UTs and integration tests (QA tests) for this implementation

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
```diff
@@ -199,6 +202,78 @@ class InventorySyncFacadeImpl final
         }
     }
 
+    void initializeKeystoreSocket()
```
MiguelazoDS, line 205: Some questions: 

Is it ok to implement this functionality inside this subcomponent inventorySync? The issue mentions a new modulesd component for this. Why give to inventorySync the responsibility to handle this?

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
```diff
@@ -199,6 +202,78 @@ class InventorySyncFacadeImpl final
...
+    {
+        m_keystoreSocketServer = std::make_unique<SocketServer<Socket<OSPrimitives, SizeHeaderProtocol>, EpollWrapper>>(
```
MiguelazoDS, line 207: Is it ok to call it "keystoreSocket", isn't the wmodules socket better for this? is there any limitation or soon deprecation of that code that makes that a bad option?

src/shared_modules/keystore/src/keyStore.cpp
GabrielEValenzuela: Including `keyStore.hpp` in `inventorySyncFacade.hpp` introduces a dependency issue that prevents C++ from compiling.
To resolve this, the `GLOBAL_LOG_FUNCTION` definition must be moved, as keeping it in its original location results in a missing definition error during compilation.

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
GabrielEValenzuela: I chose to use the `socketServer` since it’s already a testable class.
Regarding integration tests (QA tests), I believe they are more relevant to the module that consumes the data from the socket, rather than this implementation itself.

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
GabrielEValenzuela: The plan was to either add a new module or reuse an existing one. I chose `inventorySync` since it’s a core component of Wazuh, but once 5.x is more mature, we can move the socket initialization to a more appropriate place, like the `engine`.

I’m open to discussing this and creating a separate module if needed.

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
GabrielEValenzuela: I based this decision on the *single responsibility principle*. I believe it’s more appropriate to have an independent IPC rather than reuse an existing one whose communication pattern is more active and serves a different purpose.

jnasselle:
Functionality works as expected. LGTM! design considerations and specific implementation review should be addressed by @wazuh/devel-xdrsiem-server-div1

MiguelazoDS (dismissed):
LGTM, the feature works as expected according to the team that requires it. We should rebase this provide to them this solution. Once we have an e2e we can improve the design with the team.

src/shared_modules/keystore/src/keyStore.cpp
MiguelazoDS: That's ok for me

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
MiguelazoDS: I think the modulesd was chosen because any new module that does not fit in any other module ends as a new subcomponent of modulesd. 
Although I agree with the importance of the inventorySync since 5.0.0 it is initialized at the level of all other subcomponents, vulnerability scanner, SCA, Github, AWS, etc, etc. 
It was just a question to discuss, but if there's no other alternative I'll take it.

GabrielEValenzuela: I think you suggestion can be research in a particular issue, due it's correct we are mixing up the logic. Great analysis!

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
GabrielEValenzuela: We could considered the `wmodules` socket if is not longer used to fetch configuration

PR #32816, opened 2025-10-24
Make VD code compilable for 5.0.0 after state persistence

jotacarma90:
# Description
In this PR, we are going to prepare, adapt, and clean up the current VD code so that we can work on the related VD input adaptation objective.

This is because the VD code is currently in its 4.X state (pre-state persistence) and is disabled from compilation due to incompatibilities and old tools that have been removed.

### DIFF
--- a/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/scanContext.hpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/scanContext.hpp
@@ -52,7 +52,8 @@ enum class ScannerType
     ReScanSingleAgent = 8,
     CleanupAllAgentData = 9,
     CleanupSingleAgentData = 10,
-    GlobalSyncInventory = 11
+    // ###### VD input adaptation in progress
+    // GlobalSyncInventory = 11
 };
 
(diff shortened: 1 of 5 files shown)


src/wazuh_modules/vulnerability_scanner/testtool/scanner/main.cpp
GabrielEValenzuela, line 1: Probably this tool require a refactoring, or remove since we are not longer using `rsync` or `dbsync` messages

src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/scanContext.hpp
GabrielEValenzuela, line 1: This is more a reference for https://github.com/wazuh/wazuh/issues/32825
On that task I'm considering the context from the Fbb message and probably this context will not be required

GabrielEValenzuela (approved):
LGTM
We have a base for the incoming changes of VD

Great work Jose!

MarcelKemp (approved):
I think it's a start. A lot of cleaning up of the commented code will be needed, but we can start from this point.

PR #32895, opened 2025-10-30
Implement Module Recovery Flow in FIM

### LINKED ISSUE #32174: Implement Module Recovery Flow in FIM
**Description**

Implement module-level recovery for the **File Integrity Monitoring (FIM)**.

FIM must be able to lock scans, validate state consistency, detect the first scan, and trigger a full recovery sync when checksum mismatches occur. The same thread is used for both **synchronization** and **disaster recovery**, ensuring atomicity and consistent state handling.

**Tasks**

* [x] Implement logic to **lock FIM scans** during sync/recovery attempts.
* [x] Ensure recovery process only executes after a successful synchronization.
* [x] On sync failure, unlock and wait until next scheduled sync.
* [x] On sync success, calculate and validate checksum.
* [x] If checksum mismatch, retrieve all FIM DB entries, persist them in memory, and trigger full sync through `AgentSyncProtocol`.
* [x] Unlock scans once recovery process completes.
* [x] Implement periodic integrity validation (`integrity_interval`, default = 24h).
* [x] Add tests simulating checksum mismatch, and recovery flows.

**Definition of Done (DoD)**

* FIM scans can be locked/unlocked correctly during recovery.
* Checksum validation implemented after successful sync.
* Full sync triggered when checksum mismatches are detected.
* Periodic recovery validation runs at configured interval.
* Tests validate recovery for both success and failure cases.

> Note: increase `response_timeout` default value to `60s`.

jpcerrone:
## Description

  This pull request implements a recovery mechanism for the FIM module to detect and recover from database inconsistencies between agents and managers. The implementation uses periodic integrity validation through checksum-of-checksums comparison and automatic full synchronization when discrepancies are detected.

  Closes #32174

  ## Proposed Changes

  - **Synchronization Locking**: Mutex protection (`fim_scan_mutex` , `fim_realtime_mutex` and `fim_registry_db_mutex`) for FIM scans during synchronization
  - **FIM Recovery Module**: Complete C++ recovery module (`src/syscheckd/src/recovery/recovery.cpp`) implementing integrity validation and recovery logic
  - **Metadata Table**: Database table `table_metadata` to persistently track last synchronization timestamps for each FIM table (file_entry, registry_key, registry_data)

  ```sql
  CREATE TABLE IF NOT EXISTS table_metadata (
      table_name TEXT PRIMARY KEY,
      last_sync_time INTEGER NOT NULL
  );
  ```

  - **Integrity Interval Configuration**: New `<integrity_interval>` parameter in `ossec.conf`

**Note for reviewers** the main changes are in [runc_check.c](https://github.com/wazuh/wazuh/pull/32895/files#diff-4af78004ce09b4240bb76f1aae184873b915da0c4e4e51c0120ff50e6cbe2e53) and [recovery.cpp](https://github.com/wazuh/wazuh/pull/32895/files#diff-822cd2b1317c10b01f059a4eb007418b2220aa87f92d70d635e45a132913df50)

  ## Results and Evidence
Since the checksum-of-checksums validation still needs to be implemented on the manager side, the agent fails the synchronization process:

```
2025/11/03 17:22:08 wazuh-syscheckd[534791] run_check.c:615 at fim_run_integrity(): INFO: Running FIM synchronization.
2025/11/03 17:22:08 wazuh-syscheckd[534791] mq_op.c:52 at StartMQWithSpecificOwnerAndPerms(): DEBUG: Connected succesfully to 'queue/sockets/queue' after 0 attempts
2025/11/03 17:22:08 wazuh-syscheckd[534791] mq_op.c:53 at StartMQWithSpecificOwnerAndPerms(): DEBUG: (unix_domain) Maximum send buffer set to: '212992'.
2025/11/03 17:22:08 wazuh-syscheckd[534791] logging_helper.c:62 at loggingFunction(): DEBUG: Metadata available. Proceed with synchronization.
2025/11/03 17:22:08 wazuh-syscheckd[534791] logging_helper.c:50 at loggingFunction(): ERROR: Received StartAck with error status. Aborting synchronization.
2025/11/03 17:22:08 wazuh-syscheckd[534791] logging_helper.c:50 at loggingFunction(): ERROR: Synchronization failed due to manager error.
2025/11/03 17:22:08 wazuh-syscheckd[534791] logging_helper.c:59 at loggingFunction(): WARNING: Synchronization failed.
2025/11/03 17:22:08 wazuh-syscheckd[534791] run_check.c:642 at fim_run_integrity(): INFO: Synchronization failed
2025/11/03 17:22:08 wazuh-syscheckd[534791] run_check.c:647 at fim_run_integrity(): INFO: FIM synchronization finished, waiting for 60 seconds before next run.
2025/11/03 17:22:11 wazuh-agentd: WARNING: Unknown message received from server.
```
Manager side:
```
2025/11/03 20:23:18 wazuh-modulesd:agent-info: WARNING: Synchronization failed for metadata/groups mode
```

Because of this, I created tests that validate most of the functionality used inside the recovery process. These tests mock the server's response to the synchronization requests.

The following functions have had tests added for them:
`calculateTableChecksum()`
`fim_recovery_persist_table_and_resync()`
`fim_recovery_integrity_interval_has_elapsed()`

The `fim_recovery_check_if_full_sync_required()` didn't require a test since it simply only calls `calculateTableChecksum()` and `requiresFullSync()` for both of which there are tests already available.

  ### Package builds 🟢
[Packages - Build Wazuh agent Windows](https://github.com/wazuh/wazuh-agent-packages/actions/runs/19274502834)

[Packages - Build Wazuh agent Linux amd - agent packages - rpm - i386](https://github.com/wazuh/wazuh-agent-packages/actions/runs/19271581878)

[Packages - Build Wazuh agent Linux amd - agent packages - rpm - amd64](https://github.com/wazuh/wazuh-agent-packages/actions/runs/19271575776)

[Packages - Build Wazuh agent Linux amd - agent packages - deb - i386](https://github.com/wazuh/wazuh-agent-packages/actions/runs/19271570816)

[Packages - Build Wazuh agent Linux amd - agent packages - deb - amd64](https://github.com/wazuh/wazuh-agent-packages/actions/runs/19271566914)

[Packages - Build Wazuh agent macOS packages - arm64](https://github.com/wazuh/wazuh-agent-packages/actions/runs/19271621108)

[Packages - Build Wazuh agent macOS packages - intel64](https://github.com/wazuh/wazuh-agent-packages/actions/runs/19271617258)

  ## Configuration Changes

  New Configuration Parameter:
  - <integrity_interval>: Controls how often integrity validation checks are performed
    - Default value: 86400 seconds (24 hours)
    - Valid range: 1 to ∞ seconds
    - Format: Supports time notation (e.g., "24h", "12h", "5m")
    - Behavior: Checked independently per table (file_entry, registry_key, registry_data)

  ## Important Implementation Details:
  - Recovery only triggers after initial successful delta sync
  - Each table (file_entry, registry_key, registry_data) has independent integrity interval tracking
  - Last sync time is always updated (even on failure) to wait full integrity_interval before retry
  - Mutex locking ensures scans cannot interfere with synchronization process
  
  
## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/src/unit_tests/syscheckd/registry/test_registry.c
+++ b/src/unit_tests/syscheckd/registry/test_registry.c
@@ -656,8 +656,8 @@ static void test_fim_registry_scan_base_line_generation(void **state) {
     syscheck.registry = one_entry_config;
     syscheck.registry[0].opts = CHECK_REGISTRY_ALL;
 
-    expect_function_call_any(__wrap_pthread_mutex_lock);
-    expect_function_call_any(__wrap_pthread_mutex_unlock);
+    expect_function_call(__wrap_pthread_mutex_lock);
+    expect_function_call(__wrap_pthread_mutex_unlock);
 
     // Set value of FirstSubKey
     wchar_t *value_name = L"test_value";
@@ -718,8 +718,8 @@ static void test_fim_registry_scan_base_line_generation(void **state) {
 static void test_fim_registry_scan_regular_scan(void **state) {
     syscheck.registry = default_config;
 
-    expect_function_call_any(__wrap_pthread_mutex_lock);
-    expect_function_call_any(__wrap_pthread_mutex_unlock);
+    expect_function_call(__wrap_pthread_mutex_lock);
+    expect_function_call(__wrap_pthread_mutex_unlock);
 
     // Set value of FirstSubKey
     wchar_t *value_name = L"test_value";
@@ -817,6 +817,9 @@ static void test_fim_registry_scan_RegOpenKeyExW_fail(void **state) {
     expect_RegOpenKeyExW_call(HKEY_LOCAL_MACHINE, L"Software\\Classes\\batfile", 0,
                              KEY_READ | KEY_WOW64_64KEY, NULL, -1);
 
+    expect_function_call(__wrap_pthread_mutex_lock);
+    expect_function_call(__wrap_pthread_mutex_unlock);
+
     expect_function_call(__wrap_fim_db_transaction_deleted_rows);
     expect_function_call(__wrap_fim_db_transaction_deleted_rows);
 
@@ -841,6 +844,9 @@ static void test_fim_registry_scan_RegQueryInfoKey_fail(void **state) {
     expect_RegOpenKeyExW_call(HKEY_LOCAL_MACHINE, L"Software\\Classes\\batfile", 0,
                              KEY_READ | KEY_WOW64_64KEY, NULL, ERROR_SUCCESS);
     expect_RegQueryInfoKey_call(1, 0, &last_write_time, -1);
+
+    expect_function_call(__wrap_pthread_mutex_lock);
+    expect_function_call(__wrap_pthread_mutex_unlock);
 
     expect_function_call(__wrap_fim_db_transaction_deleted_rows);
     expect_function_call(__wrap_fim_db_transaction_deleted_rows);


docs/ref/modules/utils/sync-protocol/api-reference.md
```diff
@@ -50,15 +50,15 @@ Persists a data to the internal queue for later synchronization.
 
 **Parameters:**
 - `id`: Unique identifier for the data source (typically a hash of primary keys, e.g., file-path)
-- `operation`: Type of operation (`Operation::Create`, `Operation::Update`, `Operation::Delete`)
+- `operation`: Type of operation (`Operation::CREATE`, `Operation::UPDATE`, `Operation::DELETE`)
```
jpcerrone, line 53: Note for Reviewers: Didn't change the enum in this PR, but the docs were using the wrong values

src/shared_modules/sync_protocol/include/agent_sync_protocol_c_wrapper.hpp
jpcerrone, line 1: Note for Reviewers: Moved from `agent_sync_protocol_c_interface.cpp` into its own file so that we can import it somewhere else

src/syscheckd/src/run_check.c
```diff
@@ -590,10 +593,48 @@ void * fim_run_integrity(__attribute__((unused)) void * args) {
...
 
+#ifdef WIN32
+    int table_count = 3;
+    char* table_names[3] = {FIMDB_FILE_TABLE_NAME, FIMDB_REGISTRY_KEY_TABLENAME, FIMDB_REGISTRY_VALUE_TABLENAME};
+#else
+    int table_count = 1;
+    char* table_names[1] = {FIMDB_FILE_TABLE_NAME};
+#endif
+
     while (fim_sync_module_running) {
```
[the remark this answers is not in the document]
jpcerrone: Checked the rest of the codebase and this convention is followed

src/Makefile
```diff
@@ -2242,11 +2242,11 @@ win32_ui_o := $(win32_ui_c:.c=.o)
...
-	${OSSEC_CCBIN} -DARGV0=\"wazuh-agent\" -DOSSECHIDS ${OSSEC_LDFLAGS} $^ ${OSSEC_LIBS} ${DBSYNC_LIB} -lwazuh-syscheckd -l:libfimdb.lib libwazuh.a ${SYNC_PROTOCOL_LIB} -o $@
+win32/wazuh-agent.exe: ${WAZUHEXT_LIB} win32/wazuh_agent_resource.o win32/version-app.o win32/icon.o win32/win_agent.o win32/win_service.o win32/win_utils.o os_crypto/md5_sha1_sha256/md5_sha1_sha256_op.o ${rootcheck_o} $(filter-out wazuh_modules/main.o, ${wmodulesd_o}) $(filter-out client-agent/main.o, $(filter-out client-agent/agentd.o, $(filter-out client-agent/event-forward.o, ${client_agent_o}))) $(filter-out logcollector/main.o, ${os_logcollector_o}) $(filter-out os_execd/main.o, ${os_execd_o}) active-response/active_responses.o monitord/rotate_log.o monitord/compress_log.o
+	${OSSEC_CXXBIN} -DARGV0=\"wazuh-agent\" -DOSSECHIDS ${OSSEC_LDFLAGS} $^ ${DBSYNC_LIB} -lwazuh-syscheckd -l:libfimdb.lib libwazuh.a ${SYNC_PROTOCOL_LIB} ${WAZUHEXT_LIB} ${OSSEC_LIBS} -o $@
```
jpcerrone: Need to switch to CPP since the new file added to syscheckd (`recovery.cpp`) is a C++ file. Also the inclusion of `hashHelper.h` in `recovery.cpp` made me require linking with `WAZUHEXT_LIB`

src/syscheckd/src/run_check.c
```diff
@@ -602,10 +605,53 @@ void * fim_run_integrity(__attribute__((unused)) void * args) {
...
-        asp_sync_module(syscheck.sync_handle, MODE_DELTA, syscheck.sync_response_timeout, FIM_SYNC_RETRIES, syscheck.sync_max_eps);
+        bool sync_result = asp_sync_module(syscheck.sync_handle, 
+                                           MODE_DELTA, 
+                                           syscheck.sync_response_timeout, 
+                                           FIM_SYNC_RETRIES, 
+                                           syscheck.sync_max_eps); 
+        if (sync_result) {
+            minfo("Synchronization succeeded");
+
+            if (!fim_db_check_if_first_scan_has_been_synched()) {
```
jpcerrone: Note for reviewers: This was implemented with the intention of sending a `first_scan_flag` for `asp_sync_module` but it has since been determined that this will be dealt with in another task. I'm leaving the implementation to check and set the flag in so that it can be reused and adapted in that new task.

src/syscheckd/src/recovery/src/recovery.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include "fimCommonDefs.h"
+#include "stringHelper.h"
+#include "ipersistent_queue.hpp"
+#include "syscheck.h"
+#include <chrono>
+
+/**
+ * @brief Get the current timestamp in UNIX format
+ */
+int64_t getUnixTimeSeconds() {
```
Nicogp: There is already a function `Utils::getSecondsFromEpoch()` (defined in `shared_modules/utils/timeHelper.h`) that performs exactly the same operation.

src/Makefile
TomasTurina, line 1: Check if these changes are really necessary or if they can be resolved in another way.

src/syscheckd/src/recovery/CMakeLists.txt
TomasTurina, line 1: Apply this order:

- syscheckd
  - src
    - recovery
      - CMakeLists.txt
      - include
        - recovery.h
      - src
        - recovery.cpp
      - tests
        - CMakeLists.txt
        - recoveryTests.cpp

src/Makefile
```diff
@@ -2242,11 +2242,11 @@ win32_ui_o := $(win32_ui_c:.c=.o)
...
-	${OSSEC_CCBIN} -DARGV0=\"wazuh-agent\" -DOSSECHIDS -DEVENTCHANNEL_SUPPORT ${OSSEC_LDFLAGS} $^ ${OSSEC_LIBS} ${DBSYNC_LIB} -lwazuh-syscheckd-event -l:libfimdb.lib libwazuh.a ${SYNC_PROTOCOL_LIB} -lwevtapi -o $@
+win32/wazuh-agent-eventchannel.exe: ${WAZUHEXT_LIB} win32/wazuh_agent_eventchannel_resource.o win32/version-app.o win32/icon.o win32/win_agent.o win32/win_service.o win32/win_utils.o os_crypto/md5_sha1_sha256/md5_sha1_sha256_op.o ${rootcheck_o} $(filter-out wazuh_modules/main.o, ${wmodulesd_o}) $(filter-out client-agent/main.o, $(filter-out client-agent/agentd.o, $(filter-out client-agent/event-forward.o, ${client_agent_o}))) $(filter-out logcollector/main-event.o, ${os_logcollector_eventchannel_o}) $(filter-out os_execd/main.o, ${os_execd_o}) active-response/active_responses.o monitord/rotate_log.o monitord/compress_log.o
+	${OSSEC_CXXBIN} -DARGV0=\"wazuh-agent\" -DOSSECHIDS -DEVENTCHANNEL_SUPPORT ${OSSEC_LDFLAGS} $^ ${DBSYNC_LIB} -lwazuh-syscheckd-event -l:libfimdb.lib libwazuh.a ${SYNC_PROTOCOL_LIB} ${WAZUHEXT_LIB} ${OSSEC_LIBS} -lwevtapi -static-libstdc++ -static-libgcc -o $@
```
Nicogp: Are these changes necessary? We should avoid statically linking GCC/libstdc++ for the Winagent, since we are already including libstdc++-6.dll and libgcc_s_dw2-1.dll.

src/syscheckd/src/recovery/src/recovery.cpp
```diff
@@ -0,0 +1,154 @@
...
+        std::chrono::system_clock::now().time_since_epoch()
+    ).count();
+}
+
+/**
+ * @brief Calculate the checksum-of-checksums for a table
+ * @param table_name The table to calculate checksum for
+ * @return The SHA1 checksum-of-checksums as a hex string
+ */
+std::string calculateTableChecksum(const char* table_name) {
```
Nicogp: Could this method be part of the DB class and placed alongside DB::getConcatenatedChecksums?

src/syscheckd/src/db/src/db.cpp
```diff
@@ -18,10 +18,13 @@
 #include "dbRegistryValue.hpp"
 #include "dbsync.h"
 #include "dbsync.hpp"
+#include "debug_op.h"
```
TomasTurina: Why is this header needed?

src/syscheckd/src/db/src/db.cpp
```diff
@@ -18,10 +18,13 @@
...
 #include "fimCommonDefs.h"
 #include "fimDB.hpp"
 #include "fimDBSpecialization.h"
+#include <hashHelper.h>
 #include "stringHelper.h"
+#include "ipersistent_queue.hpp"
```
TomasTurina: Why is this header needed?

src/syscheckd/src/db/src/db.cpp
```diff
@@ -83,6 +88,118 @@ int DB::countEntries(const std::string& tableName, const COUNT_SELECT_TYPE selec
...
+
+    auto callback {[&concatenatedChecksums](ReturnTypeCallback type, const nlohmann::json & jsonResult)
+    {
+        if (ReturnTypeCallback::SELECTED == type)
+        {
+            concatenatedChecksums += jsonResult.at("checksum").get<std::string>();
+        }
+    }};
+
+    auto selectQuery {SelectQuery::builder()
```
TomasTurina, line 101: Maybe we can order by checksum? To guarantee this always returns elements in the same order.

src/syscheckd/src/db/src/file.cpp
```diff
@@ -140,6 +140,7 @@ void DB::searchFile(const SearchData& data, std::function<void(const std::string
     FIMDB::instance().executeQuery(selectQuery.query(), localCallback);
 }
 
+
```
TomasTurina: No need

docs/ref/modules/fim/configuration.md
TomasTurina, line 1: Check if it is also necessary to add FIM documentation about the new recovery mechanism.

src/syscheckd/src/recovery/src/recovery.cpp
```diff
@@ -0,0 +1,154 @@
...
+    #include "debug_op.h" // Use the same C-style logging as the rest of FIM's code.
+}
+
+#include "recovery.h"
+#include "agent_sync_protocol_c_interface.h"
+#include "agent_sync_protocol_c_wrapper.hpp"
+#include "db.hpp"
+#include "fimCommonDefs.h"
+#include "stringHelper.h"
+#include "ipersistent_queue.hpp"
```
TomasTurina: Why is this include necessary?

src/syscheckd/src/recovery/src/recovery.cpp
```diff
@@ -0,0 +1,154 @@
+#include <cstdint>
+#include <vector>
+#include <string>
+#include <hashHelper.h>
+
+extern "C" {
+    #include "debug_op.h" // Use the same C-style logging as the rest of FIM's code.
```
TomasTurina: Why is this include necessary?

src/syscheckd/src/recovery/src/recovery.cpp
TomasTurina: Check if there is any utility for this in `src/shared_modules/utils/timeHelper.h`

src/syscheckd/src/db/src/db.cpp
```diff
@@ -18,10 +18,13 @@
 #include "dbRegistryValue.hpp"
 #include "dbsync.h"
 #include "dbsync.hpp"
+#include "debug_op.h"
 #include "fimCommonDefs.h"
 #include "fimDB.hpp"
 #include "fimDBSpecialization.h"
+#include <hashHelper.h>
```
TomasTurina, line 24: Do we need this header here?

src/syscheckd/src/recovery/src/recovery.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+extern "C"
+{
+void fim_recovery_persist_table_and_resync(char* table_name, uint32_t sync_response_timeout, long sync_max_eps, AgentSyncProtocolHandle* handle, SynchronizeModuleCallback test_callback){
+    std::vector<nlohmann::json> recoveryItems = DB::instance().getEveryElement(table_name);
+    AgentSyncProtocolWrapper* wrapper = reinterpret_cast<AgentSyncProtocolWrapper*>(handle);
+
+    for (const nlohmann::json& item : recoveryItems) {
+        // Calculate SHA1 hash to get an id for persistDifferenceInMemory()
+        std::string id = item["path"];
```
TomasTurina: This is not valid for registry. Check usages of `persist_syscheck_msg` to see how to calculate IDs for different tables.

src/syscheckd/src/recovery/src/recovery.cpp
```diff
@@ -0,0 +1,154 @@
...
+            std::chrono::seconds(sync_response_timeout),
+            FIM_SYNC_RETRIES,
+            sync_max_eps
+        );
+        // LCOV_EXCL_STOP
+
+    }
+
+    if (success) {
+        wrapper->impl->clearInMemoryData();
```
TomasTurina: Why is this necessary? Only for UT? The clear is already done inside `synchronizeModule`.

src/unit_tests/syscheckd/CMakeLists.txt
```diff
@@ -6,9 +6,13 @@ if(${TARGET} STREQUAL "winagent")
 else()
   file(GLOB sysfiles ${SRC_FOLDER}/syscheckd/build/CMakeFiles/wazuh-syscheckd.dir/src/*.o
                      ${SRC_FOLDER}/syscheckd/build/CMakeFiles/wazuh-syscheckd.dir/src/*/*.o)
+
   list(REMOVE_ITEM sysfiles ${SRC_FOLDER}/syscheckd/build/CMakeFiles/wazuh-syscheckd.dir/src/main.c.o)
 endif()
 
+# Exclude C++ recovery module files
+list(FILTER sysfiles EXCLUDE REGEX ".*[\\/]recovery[\\/].*")
+
```
TomasTurina: Shouldn't be necessary.

src/syscheckd/src/recovery/src/recovery.cpp
TomasTurina: This should be called at the beginning of the function to ensure the data in memory is clean before we start to persist, not at the end.

src/syscheckd/src/recovery/src/recovery.cpp
```diff
@@ -0,0 +1,154 @@
...
+// LCOV_EXCL_START
+bool fim_recovery_check_if_full_sync_required(char* table_name, uint32_t sync_response_timeout, long sync_max_eps, AgentSyncProtocolHandle* handle){
+    minfo("Attempting to get checksum for %s table", table_name);
+    std::string final_checksum = calculateTableChecksum(table_name);
+    minfo("Success! Final file table checksum is: %s", final_checksum.c_str());
+
+    bool needs_full_sync;
+    AgentSyncProtocolWrapper* wrapper = reinterpret_cast<AgentSyncProtocolWrapper*>(handle);
+    needs_full_sync = wrapper->impl->requiresFullSync(
+        table_name,
```
TomasTurina: We need the index here, not the table name.

src/syscheckd/src/run_check.c
```diff
@@ -26,6 +28,7 @@
 #include "db/include/db.h"
 #include "ebpf/include/ebpf_whodata.h"
 #include "agent_sync_protocol_c_interface.h"
+#include "db.h"
```
TomasTurina: Is this header necessary?

src/syscheckd/src/run_check.c
```diff
@@ -10,6 +10,8 @@
 
 
 // SCHED_BATCH is Linux specific and is only picked up with _GNU_SOURCE
+#include "fimCommonDefs.h"
```
TomasTurina: Is this header necessary?

src/syscheckd/src/run_check.c
```diff
@@ -602,10 +605,53 @@ void * fim_run_integrity(__attribute__((unused)) void * args) {
...
+#ifdef WIN32
+    int table_count = 3;
+    char* table_names[3] = {FIMDB_FILE_TABLE_NAME, FIMDB_REGISTRY_KEY_TABLENAME, FIMDB_REGISTRY_VALUE_TABLENAME};
+#else
+    int table_count = 1;
+    char* table_names[1] = {FIMDB_FILE_TABLE_NAME};
+#endif
+
     while (fim_sync_module_running) {
+        w_mutex_lock(&syscheck.fim_scan_mutex);
```
TomasTurina, line 637: Do we also have a mutex for registry scans?

src/syscheckd/src/run_check.c
```diff
@@ -602,10 +605,53 @@ void * fim_run_integrity(__attribute__((unused)) void * args) {
...
+        w_mutex_lock(&syscheck.fim_realtime_mutex);
```
TomasTurina, line 638: Is this mutex enough to lock realtime scans?

TomasTurina (changes requested):
Please check the comments.

src/Makefile
jpcerrone: `${WAZUHEXT_LIB}` was already getting linked in both of these steps as `$^` adds it in:
https://github.com/wazuh/wazuh/blob/13c5b43e06034908eff4f50a0db774bac8f4fe49/src/Makefile#L1406-L1408
 Various components of the lib are used by `-lwazuh-syscheckd`. The problem is that, in contrast with GCC, G++ seems to be more strict about the ordering of the linking, and requires `${WAZUHEXT_LIB} ` to come after `-lwazuh-syscheckd`. I think it might make sense to leave the changes as they are since the wazuh ext library is getting linked in every other step. I did remove the first ${WAZUHEXT_LIB} at the top of the command though, since that one wasn't really necesary.

src/syscheckd/src/recovery/CMakeLists.txt
jpcerrone: Done

src/syscheckd/src/recovery/src/recovery.cpp
jpcerrone: Done

src/Makefile
jpcerrone: I'm adding them in since I was getting failures in the [check_dll_on_windows] (https://github.com/wazuh/wazuh/actions/runs/19041056002/job/54381828084#logs) jobs.
```
Failed: Process 'wazuh-agent-eventchannel.exe' is vulnerable to DLL hijacking due to 'C:\executables\libstdc++-6.dll'.
```
Is there any other way to prevent this?

src/syscheckd/src/recovery/src/recovery.cpp
jpcerrone: Done

src/syscheckd/src/db/src/db.cpp
jpcerrone: Implemented this but did it with the 'path', which is the table's primary key. Do you think this would be better or should I go with the checksum?

src/syscheckd/src/recovery/src/recovery.cpp
jpcerrone: Done

src/syscheckd/src/db/src/db.cpp
jpcerrone: Done

src/syscheckd/src/db/src/db.cpp
jpcerrone: Done

src/syscheckd/src/db/src/file.cpp
jpcerrone: Done

src/syscheckd/src/recovery/src/recovery.cpp
jpcerrone: Done and removed some other unnecesary headers

src/syscheckd/src/recovery/src/recovery.cpp
jpcerrone: This one is used for the C style logging functions like `minfo`:
`minfo("Checksum mismatch detected for index %s, full sync required", table_name);`

src/syscheckd/src/db/src/db.cpp
jpcerrone: Yes, this one is needed for:
```
    // Build checksum-of-checksums
    Utils::HashData hash(Utils::HashType::Sha1);

```

src/syscheckd/src/run_check.c
jpcerrone: Done

src/syscheckd/src/run_check.c
jpcerrone: Done

src/syscheckd/src/db/include/db.hpp
```diff
@@ -140,11 +140,58 @@ class EXPORTED DB final
...
+         * @return int64_t The last sync timestamp (UNIX format), or 0 if not found.
+         */
+        int64_t getLastSyncTime(const std::string& tableName);
+
     private:
         DB() = default;
         ~DB() = default;
         DB(const DB&) = delete;
         DB& operator=(const DB&) = delete;
+        bool first_scan_has_been_synched = false;
```
TomasTurina: This flag won't be necessary here, only in syscollector for VD context. Please delete it and also the related methods to set/get it.

src/syscheckd/src/recovery/src/recovery.cpp
jpcerrone: I was following the guide on the [sync protocol readme](https://github.com/wazuh/wazuh/blob/main/docs/ref/modules/utils/sync-protocol/README.md#quick-start) . If `synchronizeModule` already does this, do you think I should remove this step from the guide and examples?

src/syscheckd/src/recovery/src/recovery.cpp
jpcerrone: Done, copied what's done in registry.c

src/syscheckd/src/recovery/src/recovery.cpp
jpcerrone: Done, I also changed the other instance in this file where this change was needed.

src/syscheckd/src/db/include/db.hpp
jpcerrone: Done

src/unit_tests/syscheckd/CMakeLists.txt
jpcerrone: Done

docs/ref/modules/fim/configuration.md
jpcerrone: Done, added docs on recovery and the new metadata table.

src/syscheckd/src/run_check.c
jpcerrone: I've added a new mutex for it

src/syscheckd/src/recovery/src/recovery.cpp
jpcerrone: I've implemented your suggestion and modified the docs to include this step.

src/Makefile
jpcerrone: After removing `${WAZUHEXT_LIB}` from the beggining of this step, the dll hikacking errors stopped appearing, so I removed the static linking of the libs. I think the linking order might have had something to do with these errors, but the new order doesn't have this problem.

src/syscheckd/src/run_check.c
jpcerrone: Do you mean because of  `&syscheck.directories_lock` ? It seems that protects `syscheck.directories`, but we're not accesing  that list anywhere in the recovery process. I think the mutex should  be enough to lock the scans since every method that uses realtime is also guarded by the lock. But maybe I'm missing something.

src/unit_tests/syscheckd/CMakeLists.txt
jpcerrone: Also removed the other instance of this in my next commit, currently waiting for checks to finish to push it.

src/syscheckd/src/db/src/db.cpp
TomasTurina: If we use the primary key, the manager would need to know the PK of each index, which would require modifying the information sent in the flatbuffers and increase complexity. Let's use the checksum, since it's the same across all indexes.

src/syscheckd/src/run_check.c
TomasTurina: Please check if it is enough or if we need to add a new mutex.

src/syscheckd/CMakeLists.txt
```diff
@@ -23,10 +23,14 @@ include_directories(${SRC_FOLDER}/external/audit-userspace/lib)
 include_directories(${SRC_FOLDER}/external/openssl/include/)
 include_directories(${SRC_FOLDER}/shared_modules/dbsync/include/)
 include_directories(${SRC_FOLDER}/shared_modules/sync_protocol/include/)
+include_directories(${SRC_FOLDER}/shared_modules/sync_protocol/build/)
+include_directories(${SRC_FOLDER}/external/flatbuffers/include/)
```
Nicogp, line 27: Is it necessary to include FlatBuffers?

jpcerrone: If I don't the code seems to use another version of flatbuffers:
```
In file included from /home/juan/wazuh/src/shared_modules/sync_protocol/include/iagent_sync_protocol.hpp:12,
                 from /home/juan/wazuh/src/shared_modules/sync_protocol/include/agent_sync_protocol.hpp:15,
                 from /home/juan/wazuh/src/syscheckd/src/recovery/include/recovery.h:30,
                 from /home/juan/wazuh/src/syscheckd/src/recovery/src/recovery.cpp:9:
/home/juan/wazuh/src/shared_modules/sync_protocol/build/inventorySync_generated.h:11:41: error: static assertion failed: Non-compatible flatbuffers version included
   11 | static_assert(FLATBUFFERS_VERSION_MAJOR == 23 &&
      |                                         ^
/home/juan/wazuh/src/shared_modules/sync_protocol/build/inventorySync_generated.h:11:41: note: the comparison reduces to ‘(1 == 23)’
/home/juan/wazuh/src/shared_modules/sync_protocol/build/inventorySync_generated.h: In member function ‘bool Wazuh::SyncSchema::DataValue::Verify(flatbuffers::Verifier&) const’:
/home/juan/wazuh/src/shared_modules/sync_protocol/build/inventorySync_generated.h:323:33: error: no matching function for call to ‘Wazuh::SyncSchema::DataValue::VerifyField<uint64_t>(flatbuf
fers::Verifier&, Wazuh::SyncSchema::DataValue::FlatBuffersVTableOffset, int) const’
  323 |            VerifyField<uint64_t>(verifier, VT_SEQ, 8) &&
      |            ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~

```
I'll look deeper into it.

src/syscheckd/src/db/src/db.cpp
jpcerrone: Got it, changed it for the checksum

src/syscheckd/src/run_check.c
jpcerrone: I've conducted some tests to see if `fim_realtime_mutex` is enough to prevent modifications to the DB while its on:
- Started GDB on syscheckd, setting breakpoints at `fim_file` and `fim_db_file_update`.
- Creating, updating and removing files inside a realtime-configured directory with the file explorer. All three events make the debugger hit the breakpoints.
- Forced a call to `w_mutex_lock(&syscheck.fim_realtime_mutex)` inside the debugger.
- Creating, updating and removing files again now doesn't hit any of the breakpoints.

From this I conclude locking `fim_realtime_mutex` makes it so no modifications can be done by the realtime opearations to the db while the recovery process is running. 


https://github.com/user-attachments/assets/fbb8cb29-80da-4248-9524-18ec0c0a282c

jpcerrone:
> This PR appears to reintroduce a regression related to dynamic linking, which contradicts the improvements made in #16607.
> 
> **Background:**
> 
> In #16607, DLL loading for the Agent executable was deferred to runtime, rather than linking these libraries during build/link time. This prevented DLLs located in the installation directory from being loaded at process startup.
> 
> **Current Behavior:**
> 
> * Analyzing the compiled executable from this PR (see attached image), we observe that it statically links to specific DLLs.
>   Image
> * In previous 4.x versions (see attached comparison), these dependencies were not present at load time.
>   Image
> 
> **Expected behavior:**
> 
> The Agent executable should _not_ be linked to DLLs found in the installation directory. Instead, these libraries must be explicitly loaded at runtime. This ensures that the dependent DLLs are not loaded automatically at process start, but only when explicitly required.ls are needed.

Changed the recovery implementation so that it now gets built as a DLL and lazy loaded explicitly, similarly to how the rest of FIM's C++ modules are built (DB and ebpf).

src/Makefile
jpcerrone: Ended up undoing all of these changes and compiling and linking the recovery module as a dll. So I reverted back to the C compiler and removed the static linking of libstdc++-6.dll and libgcc_s_dw2-1.dll

src/Makefile
jpcerrone: Removed these changes since we're now building `-l:libfim_recovery.lib` and not using the C++ compiler anymore. The `${OSSEC_LIBS}` change is  still required.

TomasTurina (commented):
I think `fim_registry_db_mutex` should wrap the `fim_registry_scan` function.

jpcerrone:
> I think `fim_registry_db_mutex` should wrap the `fim_registry_scan` function.

Done

TomasTurina (approved):

vikman90 (approved):
Thanks @jpcerrone,

The previous request was accomplished.

LGTM.

PR #32986, opened 2025-11-03
STORMSHIELD FIREWALL DECODER+RULE

RTT2394:
## Proposed Changes

**This is a proposition for enrichment of default decoder and rule for stormshield firewall**

Each file has a sample of log test

### DIFF
--- a/ruleset/stormshield_decoder.xml
+++ b/ruleset/stormshield_decoder.xml
@@ -1,0 +1,327 @@
+<!--
+  Decoder for Stormshield
+  Author: Raphael
+-->
+<!-- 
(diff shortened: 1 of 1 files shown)


hervepouyet:
Hello,
Thank you for this work (it will help me).
However, I have two small comments:
1/ Your groups are in French Echec_d'authentification
and not in English as is the default in Wazuh (authentication failure).
2/ Indicate that your parser supports the JSON format (not syslog).

Br

RTT2394:
Hi @hervepouyet
Thx for your reply

1) you want me to fixed that ?
2)i need to indicate that in the file ?

Best regards

PR #33202, opened 2025-11-18
SCA regexp bug for Multiple CIS Apple macOS check 'Enable FileVault'

IsExec:
|---| --- | --- |
| 4.x.x| SCA | Bug |

from cschultz@chadis.com:

macOS Agent on 11.7.1 (Big Sur) says that FileVault is not enabled when it is. The problem is that the regexp for this check is wrong.

This is what the SCA says:
c:fdesetup status -> r:^FileVault\s*\t*is\s*\t*On$

This is what [fdesetup status] returns:
FileVault is On.

The regular expression should be ^FileVault\s*is\s*On[.]?$

Note that \s includes \t, so \s*\t* is redundant. I have added an optional period at the end of the line (which is why this check is failing for me). This check probably ought to be case-insensitive; I'm not sure how to express that in the regexp syntax used for Wazuh.

### DIFF
--- a/ruleset/sca/darwin/16/cis_apple_macOS_10.12.yml
+++ b/ruleset/sca/darwin/16/cis_apple_macOS_10.12.yml
@@ -168,7 +168,7 @@ checks:
       - cis: ["2.6.1.1"]
     condition: all
     rules:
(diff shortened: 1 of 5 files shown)


AwwalQuan (approved):
LGTM!

PR #33249, opened 2025-11-20
Delete agents states on agents removal

TomasTurina:
## Description

This PR adds the missing cleanup step when an agent is removed.

Until now, deleting an agent only removed it from the agent list, but its internal “state” data (vulnerabilities, metadata, stored records, etc.) remained in the manager. This caused stale information for agents that no longer exist.

With this change, when an agent is deleted, all related state data is also removed from the manager.

## Proposed Changes

- Extend the agent-removal logic to also delete:
  - Agent state entries
  - Vulnerability data linked to that agent
  - Any other stored records associated with that agent ID
- Ensure cleanup happens for both API and CLI deletions.
- Prevent leftover or orphaned data in manager-side databases.

### Results and Evidence

Authd removal (master):

```
5740782-2025/12/02 20:19:01 wazuh-authd: INFO: Agent '001' (WIN-K9C7QDERVJB) deleted (requested locally)
5740879-2025/12/02 20:19:01 wazuh-authd: INFO: Sent agent deletion request for agent '001' to inventory-sync
5740980:2025/12/02 20:19:01 logger-helper: INFO: InventorySyncFacade::deleteAgent: Deleting data for agent '001' from wazuh-states-* indexes
5741113:2025/12/02 20:19:01 logger-helper: INFO: InventorySyncFacade::deleteAgent: Successfully deleted data for agent '001'
```

Database module removal (worker):

```
10655539-2025/12/02 21:22:54 wazuh-modulesd:database: INFO: Sent agent deletion request for agent '001' to inventory-sync
10655652:2025/12/02 21:22:54 logger-helper: INFO: InventorySyncFacade::deleteAgent: Deleting data for agent '001' from wazuh-states-* indexes
10655785:2025/12/02 21:22:54 logger-helper: INFO: InventorySyncFacade::deleteAgent: Successfully deleted data for agent '001'
```

### Artifacts Affected

Manager

### Configuration Changes

N/A

### Tests Introduced

- Delete an agent and verify that:
  - The agent is no longer listed.
  - No state or vulnerability data remains for that agent.
- Ensure deleting one agent does not affect other agents’ data.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/src/wazuh_modules/wm_database.c
+++ b/src/wazuh_modules/wm_database.c
@@ -119,6 +119,8 @@ void* wm_database_main(wm_database *data) {
     // the agent addition and removal from the database will be held by authd
     // in the master.
 #ifndef LOCAL
+    // Wait for inventory_sync to subscribe to router before sending deletion messages
+    sleep(5);
     wm_sync_agents();
 #endif
 
--- a/src/wazuh_modules/wmodules.c
+++ b/src/wazuh_modules/wmodules.c
@@ -101,16 +101,16 @@ int wm_config() {
         wm_add(module);
     }
 
+    // Inventory sync
+    if ((module = wm_inventory_sync_read()))
+        wm_add(module);
+
     // The database module won't be available on agents
     if ((module = wm_database_read()))
         wm_add(module);
 
     // Downloading module
     if ((module = wm_download_read()))
-        wm_add(module);
-
-    // Inventory sync
-    if ((module = wm_inventory_sync_read()))
         wm_add(module);
 
 #endif


src/wazuh_modules/wm_database.c
```diff
@@ -117,6 +119,8 @@ void* wm_database_main(wm_database *data) {
     // the agent addition and removal from the database will be held by authd
     // in the master.
 #ifndef LOCAL
+    // Wait for inventory_sync to subscribe to router before sending deletion messages
+    sleep(5);
```
MiguelazoDS, line 123: Isn't there any other mechanism for waiting? if it is not possible to wait for some fd with select for example, (that I think not, because the fd is created by the publisher) couldn't we do a retry logic waiting for 500ms, 1s, 2s, 4s, etc? 

Just a question, I didn't check the entire logic.

TomasTurina: Since these are different threads of the same process, I couldn't think of a better solution; 1 second should be enough, but to be safe I added 5 seconds, and since the router simply discards the messages without giving errors if no consumer is ready, this is the simplest solution without major changes on the router side.

MiguelazoDS: Yeah I get it.

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
```diff
@@ -85,6 +86,35 @@ class InventorySyncFacadeImpl final
 
     void run(const std::vector<char>& dataRaw)
     {
+        // Check if message is JSON (starts with '{')
+        if (!dataRaw.empty() && dataRaw[0] == '{')
```
jr0me, line 90: Is this check necessary? wouldn't the nlohmann::json::parse just return an empty object or similar?

src/os_auth/main-server.c
```diff
@@ -1055,6 +1059,52 @@ void* run_remote_server(__attribute__((unused)) void *arg) {
...
+ * @brief Send agent deletion message to inventory-sync via router
+ *
+ * @param agent_id The ID of the agent to delete
+ */
+static void send_agent_delete_to_inventory_sync(const char *agent_id) {
+    static ROUTER_PROVIDER_HANDLE router_inventory_handle = NULL;
+
+    // Try to create router handle if not already created
+    if (!router_inventory_handle) {
+        router_inventory_handle = router_provider_create("inventory-states", false);
```
jr0me, line 1072: Is this freed at some point? (if needed)

src/wazuh_modules/wm_database.c
```diff
@@ -283,6 +287,91 @@ void wm_sync_agents() {
...
+
+/**
+ * @brief Send agent deletion message to inventory-sync via router
+ *
+ * @param agent_id The ID of the agent to delete
+ */
+static void send_agent_delete_to_inventory_sync(const char *agent_id) {
+    static ROUTER_PROVIDER_HANDLE router_inventory_handle = NULL;
+    static bool router_initialized = false;
+
```
jr0me, line 324: Can this function be called from multiple threads concurrently?

src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp
TomasTurina: The check is necessary because if we don't check first, `nlohmann::json::parse()` will throw an exception on FlatBuffer binary data, which would be caught and logged as an error. The check avoids unnecessary exception throwing and error logging for normal FlatBuffer messages.

src/os_auth/main-server.c
TomasTurina: Like in `wazuh-remoted`, this lives for the lifetime of the process. This is acceptable since these modules run for the entire lifetime of the daemon.

src/wazuh_modules/wm_database.c
TomasTurina: All calls are from the same thread (the database module's main thread), so concurrent access shouldn't be an issue.

src/wazuh_modules/wmodules.c
```diff
@@ -101,6 +101,10 @@ int wm_config() {
         wm_add(module);
     }
 
+    // Inventory sync
+    if ((module = wm_inventory_sync_read()))
+        wm_add(module);
+
```
MiguelazoDS, line 107: This will change the order on how the different modules of wmodules will be initialized right? Since the inventory sync is a key player from now on, this was intentionally moved to be initialized first? Is there an issue if not? I would leave a comment in that case

TomasTurina: All related to the need of inventory sync to start running before `wm_database`.

PR #33803, opened 2026-01-05
Sanitize written characters in buffer for syscollector delta queries. 

MiguelazoDS:
## Description

Some extra checks where added during the query creation in Syscollector delta events for upsert and delete operations. 
The buffer is guaranteed to always have enough room for queries in real scenarios. 

We would like to thank @skraft9 for reporting this issue.

### DoD 

Some comments about the issue.

For wdb_delete_dbsync() particularly, all values used in the queries are hardcoded, table name is limited to 255 characters and for deletion the number and length of PK field names won't exceed the maximum 2048

For wdb_upsert_dbsync() although it may be more susceptible to trigger the overflow, again, the events are determined by the SQLite schemas, where the bigger table contains only 35 fields, even if we consider that all fields will be included in the query (that is not the case) and assuming a 50 characters length which exceeds the real ones, we still have plenty room in our buffer.

Checks we added and tests implemented. I was worry about the approach of triggering an error and exiting without completing the query because we do not want to discard values to be inserted or deleted, that would create inconsistencies. But since as mentioned above, I don't see in a real scenario how to exceed the current buffer size, the behavior is guaranteed to be the same as of now.


src/unit_tests/wazuh_db/test_wdb_delta_event.c
```diff
@@ -830,6 +830,182 @@ void test_wdb_upsert_dbsync_packages_null_pk_field (void **state) {
...
+        "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
+        "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
+        "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
+        "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
+        "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
+        "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
+        "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
+        "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
```
MarcelKemp: Could we generate the message in another way, or at least unify it with a macro, to avoid repeating it several times?

MarcelKemp (changes requested):

src/unit_tests/wazuh_db/test_wdb_delta_event.c
MiguelazoDS: Multiple tests repeat the same  declaration, I kept that principle but replaced with a loop to make it more clear the size of the variable.

src/unit_tests/wazuh_db/test_wdb_delta_event.c
```diff
@@ -830,6 +830,158 @@ void test_wdb_upsert_dbsync_packages_null_pk_field (void **state) {
...
+        if (i == 0) {
+            TEST_FIELDS[i].value.is_pk = true;
+        } else {
+            TEST_FIELDS[i].value.is_pk = false;
+        }
+        TEST_FIELDS[i].value.source_name = NULL;
+        TEST_FIELDS[i].value.target_name = strdup(field_name);
+        TEST_FIELDS[i].value.default_value.integer = 0;
+        TEST_FIELDS[i].value.convert_empty_string_as_null = true;
+        TEST_FIELDS[i].value.convert_empty_string_as_null = true;
```
MarcelKemp:
```suggestion
        TEST_FIELDS[i].value.convert_empty_string_as_null = true;
```

src/wazuh_db/wdb_delta_event.c
```diff
@@ -293,10 +315,20 @@ bool wdb_delete_dbsync(wdb_t * wdb, struct kv const * kv_value, cJSON * data) {
...
+                    if (strlen(field_name) + query_actual_size + 3 > QUERY_MAX_SIZE) {
+                        merror("Exceeding maximum query size of %d bytes adding first pk.", QUERY_MAX_SIZE);
+                        return false;
+                    }
                     query_actual_size +=
                         snprintf(query + query_actual_size, QUERY_MAX_SIZE - query_actual_size - 1, "%s=?", field_name);
                     first_condition_element = false;
                 } else {
+                    // " AND " plus "=?" plus null terminator
+                    if (strlen(field_name) + query_actual_size + 8 > QUERY_MAX_SIZE) {
```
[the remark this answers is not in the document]
MiguelazoDS: I studied this better and it's correct that there's small bug here, but I don't think the ">=" is the right solution here, 

This is my analysis. 

Example: 

QUERY_MAX_SIZE = 20 size
query_actual_size = 5 size
harcoded string = 7 size 

So the maximum size the field_name can have is 7, because 5 + 7 + 7 + 1 > 20 is false, so it does not enter the if conditional. 

But if I use >=, it will  reject a valid 7 size field_name 

The issue is the "-1" in the snprintf because I'm already considering it in the if 

So I'll let ">", so the 7 is accepted as valid and the snprintf does 

snprintf(query + query_actual_size, QUERY_MAX_SIZE - query_actual_size,
                                                  " AND %s=?", field_name);
                                                  
snprintf(16 remaining, 20 - 5 - 7 - 7) which left room for the null terminator.

Issue #33802, opened 2026-01-05
Auditd decoder incorrectly parses events with `per` field.


mmmds:
Wazuh auditd decoder partially parses SYSCALL events but fails to extract standard fields (auid, uid, pid, etc.) when the log entry contains a `per=40000` field. If I manually remove `per=40000` from the log, the decoder parses it correctly.

Tested on Wazuh 4.14.1, Ubuntu 24.04.

Steps to reprodcue:

1. Run a binary with a personality flag
```
cp /bin/true ~/true
setarch x86_64 -R ./true
```

2. Read the event log
```
# cat /var/log/audit/audit.log
[...]
type=SYSCALL msg=audit(1767639698.984:5139): arch=c000003e syscall=59 per=40000 success=yes exit=0 a0=7ffff99a7658 a1=7ffff99a6be0 a2=7ffff99a6bf0 a3=55d008603269 items=2 ppid=5554 pid=5598 auid=1000 uid=1000 gid=1000 euid=1000 suid=1000 fsuid=1000 egid=1000 sgid=1000 fsgid=1000 tty=pts2 ses=45 comm="true" exe="/home/crow/true" subj=unconfined key="exec_commands"ARCH=x86_64 SYSCALL=execve AUID="crow" UID="crow" GID="crow" EUID="crow" SUID="crow" FSUID="crow" EGID="crow" SGID="crow" FSGID="crow"
```

3. Parse it in Wazuh. Notice that not all the fields get recognized.
```
# /var/ossec/bin/wazuh-logtest
Starting wazuh-logtest v4.14.1
Type one log per line

type=SYSCALL msg=audit(1767639698.984:5139): arch=c000003e syscall=59 per=40000 success=yes exit=0 a0=7ffff99a7658 a1=7ffff99a6be0 a2=7ffff99a6bf0 a3=55d008603269 items=2 ppid=5554 pid=5598 auid=1000 uid=1000 gid=1000 euid=1000 suid=1000 fsuid=1000 egid=1000 sgid=1000 fsgid=1000 tty=pts2 ses=45 comm="true" exe="/home/crow/true" subj=unconfined key="exec_commands"ARCH=x86_64 SYSCALL=execve AUID="crow" UID="crow" GID="crow" EUID="crow" SUID="crow" FSUID="crow" EGID="crow" SGID="crow" FSGID="crow"

**Phase 1: Completed pre-decoding.
[...]

PR #33810, opened 2026-01-06
support optional "per" field in auditd decoder

mmmds:
This PR fixes the issue [Auditd decoder incorrectly parses events with per field. #33802](https://github.com/wazuh/wazuh/issues/33802).

I wasn't sure which branch to choose for the PR becuase `main` branch doesn't have the decoder file, so I decided to go with `4.14.3`. If it's incorrect, please let me know how to improve the PR.


AwwalQuan:
Thanks @mmmds for reporting this.
A quick check showed that there are some inaccuracies your proposed solution.
I have reopened your issue https://github.com/wazuh/wazuh/issues/34520 so we we can investigate further and apply the fixes.

PR #33819, opened 2026-01-07
Add unit tests for os_strcnt function in  test_string_op.c

ignaciogalle12git:
## Description
This PR adds unit tests for the os_strcnt function in `test_string_op.c` to validate its behavior across different input scenarios.

In addition, the base implementation of `os_strcnt` has been updated to safely handle a `NULL` input string, preventing undefined behavior and allowing the function to return `0` when the input pointer is `NULL`.

`test_os_strcnt`
A table-driven unit test that combines multiple scenarios using vectors and loops:
 - Multiple occurrences of a character in a string
 - Character not present in the string
 - Empty string
  - Different character positions (beginning, middle, end)
  Also verifies that calling os_strcnt with a NULL string does not cause a crash and returns 0.

## Proposed Changes

Updated `src/unit_tests/shared/test_string_op.c` to add a table-driven unit test for os_strcnt.

Updated the `os_strcnt` implementation to validate the input string pointer and safely handle the NULL case.

### Results and Evidence

- All unit tests pass successfully.

image

- Code coverage was generated.
image
image

- Update  of `os_strcnt`.

image

### Tests Introduced

- `test_os_strcnt`

## Review Checklist

- [ ] Code changes reviewed


MarcelKemp (changes requested):
If we pass `NULL` in the parameters, does it work correctly?
On the other hand, would it be possible to improve the cases, to combine most cases into a single test? For example, using vectors/matrices and loops.

Issue #33651, opened 2025-12-22
Fix Inventory Sync IT failing to pull OpenSearch Image


juliancnn:
## Description

During the execution of PR validation, the GitHub Action
`.github/workflows/5_testintegration_inventory-sync.yml`
fails consistently in the **inventory sync integration tests** step.

The failure occurs while setting up OpenSearch and prevents all integration tests from running.

### Failing Step

```yaml
- name: Run inventory sync integration tests
  working-directory: src/wazuh_modules/inventory_sync/qa
  run: |
    echo "[>] Running inventory sync integration tests..."
    source venv/bin/activate
    python3 run_tests.py --list-tests
    TEST_MANAGER="${{ github.event.inputs.test_manager || '127.0.0.1' }}"
    python3 run_tests.py --manager "$TEST_MANAGER" --verbose
```

### Error Observed

```
❌ Failed to setup OpenSearch:
404 Client Error for http+docker://localhost/v1.48/images/opensearchproject/opensearch:latest/json:
Not Found ("No such image: opensearchproject/opensearch:latest")
```

The image **does exist on Docker Hub**, so the failure is reproducible and likely related to the test setup, Docker client configuration, or image resolution logic.

## Objectives

* Investigate why Docker fails to resolve or pull `opensearchproject/opensearch:latest` during integration tests.
* Identify whether the issue is related to:

  * Docker API usage
  * Image pull logic
  * Test environment assumptions
  * Tag resolution or platform compatibility
* Fix the test setup to reliably pull and start OpenSearch.
* Ensure inventory sync integration tests run successfully in CI.

## Scope

* Review `run_tests.py` and OpenSearch setup logic.
* Inspect Docker client usage inside the test framework.
* Validate image availability and pull behavior in GitHub Actions runners.
* Apply a fix that ensures deterministic test execution.

## Deliverables

* Root cause analysis of the OpenSearch image resolution failure.
* Updated test or setup logic fixing the issue.
* Successful execution of `inventory-sync` integration tests in CI.

## References

* Failing run:
  [https://github.com/wazuh/wazuh/actions/runs/20442608295/job/58744570363?pr=33627](https://github.com/wazuh/wazuh/actions/runs/20442608295/job/58744570363?pr=33627)

ignaciogalle12git:
### Update 07/01/26

I've been making a research looking forward to stabilizing the inventory sync integration tests workflow.
[...]

PR #33836, opened 2026-01-08
Fix OpenSearch image pull failure in GitHub Actions

ignaciogalle12git:
Resolves: #33651

## Description
This pull request fixes recurring failures in the **inventory sync integration tests** caused by Docker Hub rate limits when pulling the OpenSearch image during CI execution.

The solution moves OpenSearch provisioning to a GitHub Actions service container, ensuring a deterministic and reliable test environment and removing runtime image pulls from the test code.

## Proposed Changes

 - Define OpenSearch as a service container in the GitHub Actions workflow, including health checks.

- Apply CI-specific OpenSearch settings to avoid read-only cluster states:

  -  Enable automatic index creation.

  -  Disable disk allocation threshold checks.

- Update the Python test setup to detect an existing OpenSearch instance before starting a local container.

- Keep an explicit image pull and container startup only as a fallback for local development.
- Add a safe cleanup step after checkout to free disk space by removing unused SDKs, helping prevent disk space issues that could interfere with log generation.

**Why service containers?**

Service containers are the recommended approach in GitHub Actions to provide external dependencies because they avoid Docker Hub rate limits, ensure services are ready before tests run via built-in health checks, keep test code CI-agnostic, and provide isolated, reproducible environments for each job.

**Additional Notes**

The current CI failure is caused by the `nodata_flow `test. This issue is unrelated to the OpenSearch setup.

A warning related to missing Wazuh logs (ossec.log) was already present before these changes.

image

image

### Results and Evidence

**Local Execution**

image

**GitHub Actions Workflow**
image

  - [ ] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

## Review Checklist

- [x] Code changes reviewed
- [ ] ...

### DIFF
--- a/.github/workflows/5_testintegration_inventory-sync.yml
+++ b/.github/workflows/5_testintegration_inventory-sync.yml
@@ -33,8 +33,7 @@ jobs:
         image: opensearchproject/opensearch:latest
         env:
           discovery.type: single-node
-          plugins.security.disabled: true
-          OPENSEARCH_INITIAL_ADMIN_PASSWORD: WazuhTest99$
+          DISABLE_SECURITY_PLUGIN: true
           action.auto_create_index: "true"
           cluster.routing.allocation.disk.threshold_enabled: "false"
         ports:
--- a/src/wazuh_modules/inventory_sync/qa/test_inventory_sync_integration.py
+++ b/src/wazuh_modules/inventory_sync/qa/test_inventory_sync_integration.py
@@ -50,8 +50,7 @@ def init_opensearch(low_resources=False):
     # If not available, prepare to start local container
     env_vars = {
         'discovery.type': 'single-node',
-        'plugins.security.disabled': 'true',
-        'OPENSEARCH_INITIAL_ADMIN_PASSWORD': 'WazuhTest99$'
+        'DISABLE_SECURITY_PLUGIN': 'true'
     }
 
     if low_resources:


.github/workflows/5_testintegration_inventory-sync.yml
```diff
@@ -27,6 +27,24 @@ jobs:
...
     runs-on: ubuntu-22.04
 
+    # Service containers to run alongside the job
+    services:
+      opensearch:
+        image: opensearchproject/opensearch:latest
+        env:
+          discovery.type: single-node
+          plugins.security.disabled: true
+          OPENSEARCH_INITIAL_ADMIN_PASSWORD: WazuhTest99$
```
MarcelKemp: With security disabled, is it necessary to set a password? Because I think this line is unnecessary.

MarcelKemp (dismissed):

.github/workflows/5_testintegration_inventory-sync.yml
ignaciogalle12git: It’s due to a bug in OpenSearch (v2.12+) where the environment variable plugins.security.disabled no longer disables the security plugin in Docker.
As a result, OpenSearch still requires OPENSEARCH_INITIAL_ADMIN_PASSWORD.
A proper fix is to use DISABLE_SECURITY_PLUGIN: true instead. <- **commited**

PR #34340, opened 2026-02-04
Fix Windows FIM Registry scan crash on non-null-terminated values

vikman90:
## Description
This PR addresses a Windows agent crash observed in **4.14.3-rc2** during FIM Registry scanning. The crash was traced to handling of Registry string values returned by `RegEnumValueW()` that may be **missing proper null terminators**, as explicitly noted in [Microsoft documentation](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regenumvaluew).

When Page Heap is enabled, the out-of-bounds read becomes deterministic and the agent crashes while processing a large Registry value (reported up to **128 KiB**).

## Proposed Changes
- Increase the internal buffer used to handle Windows Registry value data during **FIM Registry scanning**.
- Always append **4 extra zero bytes** at the end of the buffer.
  - Rationale:
    - Some values can be **multi-string** (`REG_MULTI_SZ`) which requires **two null terminators**.
    - The code processes **wide characters** (`wchar_t`, 2 bytes per character), so we ensure sufficient trailing zeros to safely cover termination cases.
- Scope: **This change only affects the Windows Registry scan path in FIM**.

### Results and Evidence
- With **Page Heap enabled**, current **4.14.3-rc2** crashes while reading a Registry value of **128 KiB**.
- After this change, the crash is no longer reproducible under the same conditions.

**Before (4.14.3-rc2):**

```
2026/02/04 13:25:16 wazuh-agent: WARNING: Registry string value at index 9 in 'HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\afunix\Parameters\Winsock\0' is not null-terminated (size: 14 bytes, last wchar: 0x0058). This may indicate a malformed registry value.
```

> [!NOTE]
> This log was added for testing purposes. It's not included in this pull request.

**After (this PR):**

```
2026/02/04 13:52:24 wazuh-agent: INFO: (6008): File integrity monitoring scan started.
2026/02/04 13:53:38 wazuh-agent: INFO: (6009): File integrity monitoring scan ended.
```

### Artifacts Affected
- `wazuh-agent.exe`

### Configuration Changes
- None.

### Documentation Updates
- Not applicable.

### Tests Introduced
Added **two unit tests** covering the case where `RegEnumValueW()` returns Registry string buffers **without null terminators**, ensuring the FIM Registry scanner handles the data safely:
- test_fim_read_values_unterminated_string_value
- test_fim_read_values_unterminated_multi_string_value

## Review Checklist
[...]

PR #34353, opened 2026-02-05
Fix Windows Agent startup race

vikman90:
## Description

This PR addresses a Windows agent startup crash caused by a race condition introduced by the current initialization order. 

```
Faulting application name: wazuh-agent.exe, version: 4.14.3.0, time stamp: 0x6984c8d1
Faulting module name: wazuh-agent.exe, version: 4.14.3.0, time stamp: 0x6984c8d1
Exception code: 0xc0000005
Fault offset: 0x000103fb
Faulting process id: 0x9F78
Faulting application start time: 0x1DC96BEE3FD208C
Faulting module path: C:\Program Files (x86)\ossec-agent\wazuh-agent.exe
```

Some module threads can attempt to enqueue messages (via the agent buffer) before the send-block is enabled and before the buffer is initialized, which can lead to a NULL-pointer dereference in `buffer_append()` during early startup.

By moving the send-blocking initialization and the buffer initialization earlier in the Windows startup sequence—before any module threads are created—we remove the time window where modules can enqueue data against an uninitialized buffer, preventing the crash and making the startup flow consistent with the intended behavior (modules should not send/enqueue messages until the agent is ready/connected).

## Proposed Changes

Reorder the Windows agent startup sequence to:
- Initialize send-blocking earlier.
- Initialize the agent buffer earlier.
- Start module threads only after both are ready.

### Results and Evidence

After applying the startup reordering, **no additional Windows agent crashes were observed**.

### Artifacts Affected

- `wazuh-agent.exe` (Windows)

### Configuration Changes

Not applicable.

### Documentation Updates

Not applicable.

### Tests Introduced

Manual testing performed.

> [!NOTE]
> During development a guard was added to detect attempts to append data while the buffer was `NULL`. This is not required as a “detection” mechanism since the original behavior manifests as a crash, and the previous reproduction method no longer triggers the issue due to the reordered initialization calls.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/src/win32/win_utils.c
+++ b/src/win32/win_utils.c
@@ -182,6 +182,16 @@ int local_start()
         merror_exit("Error creating mutex.");
     }
 
+    /* Set wait lock before starting threads */
+    os_setwait();
+
+    /* Initialize buffer before starting threads that may use it */
+    if (agt->buffer) {
+        buffer_init();
+    } else {
+        minfo(DISABLED_BUFFER);
+    }
+
     /* Start syscheck thread */
     w_create_thread(NULL,
                      0,
@@ -257,9 +267,6 @@ int local_start()
         }
     }
 
-    /* Try to connect to server */
-    os_setwait();
-
     /* Socket connection */
     agt->sock = -1;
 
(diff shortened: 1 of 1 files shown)


lchico (approved):
Based on the following results, I can confirm that this PR fixes the issue. Using the same script that [previously reproduced the bug](https://github.com/wazuh/wazuh/pull/34340#pullrequestreview-3754156665), I am now seeing a 100% success rate (100/100 runs).

**Great job, @vikman90!**

Image

# Review Checklist

- [x]  Code changes reviewed
- [ ]  ~Tests cover the new functionality~

PR #34447, opened 2026-02-09
Add guards to fix crash in libfimdbdll due to race conditions

### LINKED ISSUE #34377: Crash in Windows agent: Fault in libfimdb.dll
|---|---|---|---|---|
| 5.0.0 (development) | Agent, libfimdb | Agent | Packages | Windows |

## Crash event details
```
Faulting application name: wazuh-agent.exe, version: 5.0.0.0, time stamp: 0x6983e0a9
Faulting module name: libfimdb.dll, version: 5.0.0.0, time stamp: 0x6983dff3
Exception code: 0xc0000005
Fault offset: 0x00010d82
Faulting process id: 0x26B0
Faulting application start time: 0x1DC96B62510BCD2
Faulting module path: C:\Program Files (x86)\ossec-agent\libfimdb.dll
Report Id: d0e6c6f4-50db-45b2-9206-eb44bb609e90
```

## Key points
- Crash detected in the wazuh-agent for Windows during testing on version 5.0.0 (development).
- Fault occurs in libfimdb.dll with Exception code 0xc0000005.
- Core dump is available for further analysis.
- Please advise on next troubleshooting steps or analysis requirements.

Miguevrgo:
## Description

This PR addresses the issue identified in https://github.com/wazuh/wazuh/issues/34377, where the Windows agent (`wazuh-agent.exe`) crashes with a fault in `libfimdb.dll` (Exception code `0xc0000005` — access violation).

### Crash Analysis (GDB/WinDbg)

The provided crash dumps reveal a **Null Pointer Dereference** triggered by a race condition during the early stages of the agent's lifecycle. 

- **Faulting Instruction:** `mov eax, dword ptr [edi]` where `edi` is `0x00000000`.
- **Location:** `libfimdb!ZN2DB12countEntries...` calling into `FIMDB::instance().executeQuery(...)`.
- **Uptime:** The process consistently crashes at approximately **5 seconds** of uptime.

The trace shows that `DB::countEntries` is invoked (likely by the synchronization protocol or an early scan) before the `FIMDB` singleton's `m_dbsyncHandler` has been fully initialized via the `init()` method. In the current implementation, `executeQuery` lacks any guard or synchronization, leading to an immediate jump to a null memory address when it tries to use the internal handler.

**on Merging:** Although this PR implements robust thread-safety guards, **merging might be premature**. Since the race condition depends on precise timing between the `Syscheck` initialization thread and external query requests, it would be wise to try an reproduce the error in a controlled environment and check if the patch solves it.

Closes #34377

### Thread-safety fixes in `FIMDB` (`src/syscheckd/src/db/src/fimDB.cpp` / `fimDB.hpp`)

- **Initialization Guard (`init`)**: Changed from `std::shared_lock` to `std::lock_guard`. The original use of a shared lock during a write operation (`m_dbsyncHandler = dbsyncHandler`) was a major defect, allowing other threads to attempt reads while the pointer was being assigned.
- **Entry Point Protection (`executeQuery`)**: This was the primary crash site. I have introduced a `std::shared_lock` and an explicit check for `m_dbsyncHandler`. If called before initialization, it now prevents the null dereference.
- **Atomic State Management**: The `m_stopping` flag is now `std::atomic<bool>`. This ensures that even without a lock, the "teardown" or "not-ready" state is consistently visible across all CPU cores.
- **Global API Robustness**: Added null-checks to `removeItem()` and `updateItem()`.

### Signature update for `fim_check_db_state`

- Updated the function signature in `syscheck.h` to move away from global state dependency. By passing `nodes_limit`, `nodes_count`, and `db_state` explicitly, we ensure the function operates within C23 standarts which allows for compilation on modern platforms.

## Tests

- [ ] **Verification Required:** Run the `RTR` tool (`python3 build.py -r syscheckd`) to ensure no regressions in existing Unit Tests.
- [ ] **ASAN Validation:** Monitor for "Initialization-order-fiasco" or "Data Race" warnings using Address Sanitizer.

### Artifacts Affected

- Windows Agent Package (MSI)
- `libfimdb.dll` (Runtime component)

### Configuration Changes

None

### Review Checklist

- [ ] Code changes reviewed
- [ ] Relevant evidence provided (WinDbg traces analyzed)
- [ ] **Decision Point:** Confirm if returning `0` or throwing in `countEntries` is acceptable for `sync_protocol` during startup.

### DIFF
--- a/src/syscheckd/src/db/include/db.hpp
+++ b/src/syscheckd/src/db/include/db.hpp
@@ -78,10 +78,10 @@ class EXPORTED DB final
          * @param fileLimit File limit.
          * @param valueLimit Registry value limit.
          */
-        void init(const int storage,
+        void init(int storage,
                   std::function<void(modules_log_level_t, const std::string&)> callbackLogWrapper,
-                  const int fileLimit,
-                  const int valueLimit);
+                  int fileLimit,
+                  int valueLimit);
 
         /**
          * @brief DBSyncHandle return the dbsync handle, for operations with the database.
@@ -112,7 +112,7 @@ class EXPORTED DB final
          * @param selectType Type of count.
          * @return Number of files.
          */
-        int countEntries(const std::string& tableName, const COUNT_SELECT_TYPE selectType);
+        int countEntries(const std::string& tableName, COUNT_SELECT_TYPE selectType);
 
         /**
          * @brief maxVersion Get the maximum version from a table.
@@ -130,7 +130,6 @@ class EXPORTED DB final
          * @return 0 on success, -1 on error.
          */
         int updateVersion(const std::string& tableName, int version);
-
 
         /**
          * @brief updateFile Update/insert a file in the database.
@@ -173,6 +172,7 @@ class EXPORTED DB final
          * @return int64_t The last sync timestamp (UNIX format), or 0 if not found.
          */
         int64_t getLastSyncTime(const std::string& tableName);
+
     private:
         DB() = default;
         ~DB() = default;
--- a/src/syscheckd/src/db/src/db.cpp
+++ b/src/syscheckd/src/db/src/db.cpp
@@ -21,9 +21,9 @@
 #include "fimCommonDefs.h"
 #include "fimDB.hpp"
 #include "fimDBSpecialization.h"
-#include <hashHelper.h>
 #include "stringHelper.h"
 #include "timeHelper.h"
+#include <hashHelper.h>
 
 void DB::init(const int storage,
               std::function<void(modules_log_level_t, const std::string&)> callbackLogWrapper,
@@ -37,7 +37,7 @@ void DB::init(const int storage,
                                                  FIMDBCreator<OS_TYPE>::CreateStatement(),
                                                  DbManagement::PERSISTENT)};
 
-    FIMDB::instance().init(callbackLogWrapper, dbsyncHandler, fileLimit, valueLimit);
+    FIMDB::instance().init(std::move(callbackLogWrapper), dbsyncHandler, fileLimit, valueLimit);
 }
 
 DBSYNC_HANDLE DB::DBSyncHandle()
@@ -85,14 +85,13 @@ int DB::countEntries(const std::string& tableName, const COUNT_SELECT_TYPE selec
     return count;
 }
 
-
 void DB::updateLastSyncTime(const std::string& tableName, int64_t timestamp)
 {
     auto emptyCallback = [](ReturnTypeCallback, const nlohmann::json&) {};
 
     auto syncQuery = SyncRowQuery::builder()
                      .table("table_metadata")
-    .data(nlohmann::json{{"table_name", tableName}, {"last_sync_time", timestamp}})
+    .data(nlohmann::json {{"table_name", tableName}, {"last_sync_time", timestamp}})
     .build();
 
     FIMDB::instance().updateItem(syncQuery.query(), emptyCallback);
@@ -168,7 +167,7 @@ int DB::updateVersion(const std::string& tableName, int version)
         // Select all rows (only primary keys and version column)
         auto selectQuery {SelectQuery::builder()
                           .table(tableName)
-                          .columnList({"*"})  // Get all columns to properly identify rows
+                          .columnList({"*"}) // Get all columns to properly identify rows
                           .rowFilter("")
                           .orderByOpt("")
                           .distinctOpt(false)
@@ -190,10 +189,7 @@ int DB::updateVersion(const std::string& tableName, int version)
         // Use syncRow to update the entry
         auto updateCallback {[](ReturnTypeCallback, const nlohmann::json&) {}};
 
-        auto syncQuery {SyncRowQuery::builder()
-                        .table(tableName)
-                        .data(row)
-                        .build()};
+        auto syncQuery {SyncRowQuery::builder().table(tableName).data(row).build()};
 
         try
         {
@@ -208,7 +204,6 @@ int DB::updateVersion(const std::string& tableName, int version)
 
     return retval;
 }
-
 
 #ifdef __cplusplus
 extern "C"
@@ -291,7 +286,7 @@ FIMDBErrorCode fim_db_transaction_sync_row(TXN_HANDLE txn_handler, const fim_ent
         }
         else
         {
-            if (entry->registry_entry.key == NULL)
+            if (entry->registry_entry.key == nullptr)
             {
                 syncItem = std::make_unique<RegistryValue>(entry, true);
             }
@@ -392,10 +387,10 @@ cJSON* fim_db_get_every_element(const char* table_name)
     if (!table_name)
     {
         FIMDB::instance().logFunction(LOG_ERROR, "Invalid parameters");
-        return NULL;
+        return nullptr;
     }
 
-    cJSON* result_array = NULL;
+    cJSON* result_array = nullptr;
 
     try
     {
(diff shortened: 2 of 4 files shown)


Miguevrgo:
## Update

The tests appear to be flaky, the will be relaunched when server is less loaded

Miguevrgo:
## Note

Fails are not related to the commit: https://github.com/wazuh/wazuh/issues/34476

Miguevrgo:
## Confirmation

Albeit the PR was labeled as ready thanks to a deep investigation on the issue, it has been confirmed that it fixes the problem as I have succesfully reproduced the crash and tested against the new version concluding the fix is succesfully applied:

## Reproduction Steps

### Environment

- **OS**: Ubuntu 22.04 (Vagrant VM, VirtualBox provider)
- **Wazuh version**: 5.0.0 (built from source, branch `test/reproduce-crash-34377`)
- **Hardware**: 8 CPUs, 12 GB RAM

### Challenges in Reproduction

The original race window is extremely narrow (microseconds), making natural
reproduction unreliable. Additionally, the signal handling architecture presented
two obstacles:

1. **`sleep_for` is signal-interruptible**: `std::this_thread::sleep_for()` uses
   `nanosleep()` internally, which returns early on signal delivery (EINTR). The
   worker thread's artificial delay was being cut short by SIGTERM.

2. **`exit()` kills all threads**: The signal handler calls `HandleSIG()` which
   calls `exit(1)`, immediately terminating all threads — including the one that
   would dereference the null pointer.
### Modifications Required

Three modifications were needed to widen the race window deterministically:

#### 1. Artificial delay in `executeQuery()` (`fimDB.cpp`)

A 30-second busy-wait loop with signals blocked via `pthread_sigmask()` was
inserted before the `selectRows()` call to prevent signal interruption:

```cpp
void FIMDB::executeQuery(const nlohmann::json& item, ResultCallbackData callbackData)
{
    // Block ALL signals in this thread so sleep cannot be interrupted
    {
        sigset_t set;
        sigfillset(&set);
        pthread_sigmask(SIG_BLOCK, &set, NULL);
    }

    // Signal-resistant busy-wait (30 seconds)
    {
        auto start = std::chrono::steady_clock::now();
        while (std::chrono::steady_clock::now() - start < std::chrono::seconds(30)) {
            std::this_thread::sleep_for(std::chrono::milliseconds(100));
        }
    }

    // This dereferences nullptr after teardown → SEGFAULT
    m_dbsyncHandler->selectRows(item, callbackData);
}
```

#### 2. Signal handler returns instead of exiting (`main.c`)

The `fim_shutdown()` signal handler was modified to return after calling
`teardown()` instead of calling `HandleSIG(sig)` → `exit(1)`. This keeps the
process alive so the worker thread can complete its sleep and dereference the
null pointer:

```c
static void fim_shutdown(int sig)
{
    minfo(SK_SHUTDOWN);
    is_fim_shutdown = true;
    fim_db_teardown();
    // Return instead of exit — let worker threads finish and crash
    return;
}
```

#### 3. Debug logging in `fimDB.cpp`

Logging was added at key points to capture the pointer state and confirm the race
condition sequence.

### Execution

```bash
# Build and install the agent
cd /home/vagrant/wazuh/src
make TARGET=agent -j$(nproc)
sudo cp wazuh-syscheckd /var/ossec/bin/wazuh-syscheckd

# Start the agent
sudo rm -f /tmp/fimdb_race_debug.log
sudo /var/ossec/bin/wazuh-control start

# Monitor for executeQuery entry (wait ~60-90 seconds)
tail -f --retry /tmp/fimdb_race_debug.log
# Wait until "executeQuery ENTER" appears

# In another terminal, send SIGTERM to trigger teardown during the sleep
sudo kill -15 $(pidof wazuh-syscheckd)

# Wait ~35 seconds for the crash
sleep 35
cat /tmp/fimdb_race_debug.log
sudo dmesg | grep segfault
```

## Crash Output
### Debug Log (`/tmp/fimdb_race_debug.log`)

```
[1770887531835792494] executeQuery ENTER - m_dbsyncHandler use_count=1 ptr=0x5567439886a0 m_stopping=0
[1770887533444312457] teardown ENTER - m_dbsyncHandler use_count=1 ptr=0x5567439886a0
[1770887533444467030] teardown DONE - m_dbsyncHandler set to nullptr, ptr=0
[1770887565851864193] executeQuery AFTER SLEEP - m_dbsyncHandler use_count=0 ptr=0 m_stopping=1
```

**Sequence confirmed:**
1. `executeQuery` enters with a valid pointer (`0x5567439886a0`), begins 30s sleep
2. `teardown` runs ~1.6 seconds later, destroys the pointer (`ptr=0`)
3. `executeQuery` wakes up after 30s, observes `ptr=0` and `m_stopping=1`
4. Attempts `nullptr->selectRows()` → **segfault**

### Kernel Log (`dmesg`)

```
[ 6270.762117] wazuh-syscheckd[17046]: segfault at 0 ip 00007f25de02b5f9 sp 00007ffe072c1590 error 4 in libfimdb.so[7f25de00a000+22000]
```

- **`segfault at 0`**: Null pointer dereference confirmed
- **`libfimdb.so`**: Crash location matches the FIM database library
- **`error 4`**: Read access violation in user-mode

## Fix Verification

The same debug instrumentation and reproduction steps were applied to the fix
branch `fix/34377-crash-in-windows-agent-fault-in-libfimdbdll` to verify that the
race condition no longer causes a crash.

### Verification Setup

The same three modifications (30s busy-wait sleep, signal blocking, signal handler
returning instead of exiting) were applied to the fix branch. The key difference
is that the fix branch's `executeQuery()` acquires a `shared_lock` and checks
`!m_stopping && m_dbsyncHandler` before dereferencing the handler.

On the fix branch, the instrumented `executeQuery()` follows this sequence:

```cpp
void FIMDB::executeQuery(const nlohmann::json& item, ResultCallbackData& callbackData)
{
    // DEBUG: sleep 30s with signals blocked (same as crash reproduction)
    ...

    // FIXED CODE: mutex lock + null check prevents the crash
    std::shared_lock<std::shared_timed_mutex> lock(m_handlersMutex);

    if (!m_stopping && m_dbsyncHandler)
    {
        m_dbsyncHandler->selectRows(item, callbackData);
    }
    else if (m_loggingFunction)
    {
        m_loggingFunction(LOG_DEBUG, "Query not executed: ...");
    }
}
```

### Verification Results

#### Debug Log (`/tmp/fimdb_race_debug.log`)

```
[1770888906178486490] executeQuery ENTER - m_stopping=0
[1770888911197451607] teardown ENTER - m_stopping=0 ptr=0x5599784d56a0
[1770888911197597282] teardown DONE - m_stopping=1 ptr=0
[1770888936213015256] executeQuery AFTER SLEEP - m_stopping=1
[1770888936213042478] executeQuery LOCKED - m_dbsyncHandler ptr=0 m_stopping=1
[1770888936213047878] executeQuery EXIT SAFELY - no crash
[1770888936213066914] executeQuery ENTER - m_stopping=1
```

**Sequence analysis:**

1. `executeQuery ENTER` (t=0s) — Thread enters function, `m_stopping=0`
2. `teardown ENTER` (t=+5s) — SIGTERM received, teardown begins, handler still
   valid at `0x5599784d56a0`
3. `teardown DONE` (t=+5s) — Teardown completes: `m_stopping=1`, handler set to
   `nullptr`
4. `executeQuery AFTER SLEEP` (t=+30s) — Sleep finishes, thread sees
   `m_stopping=1` (via atomic read)
5. `executeQuery LOCKED` (t=+30s) — Thread acquires `shared_lock`, confirms
   handler is `nullptr` (`ptr=0`)
6. `executeQuery EXIT SAFELY` (t=+30s) — **The `!m_stopping` check prevents the
   null dereference.** The function returns without crashing.
7. A subsequent `executeQuery ENTER` immediately sees `m_stopping=1` and would
   also skip the dereference.

#### Kernel Log (`dmesg`)

```
NO SEGFAULT FOUND IN DMESG
```

**No crash occurred.** The process remained alive after the race condition window.

PR #34658, opened 2026-02-27
Fix Heap-based NULL WRITE Buffer Underflows

ignaciogalle12git:
## Description
This change hardens `parse_uname_string()` so it safely handles edge cases where parsed string fragments may be empty before trimming the trailing delimiter.

The updated logic adds explicit length checks before truncating the last character in the parsed fields. This avoids invalid pointer arithmetic when the input contains incomplete or minimally formed OS descriptors.

The change also adds regression coverage for the affected parsing paths to ensure these inputs are handled consistently.

## Proposed Changes

- Add  length checks before trimming parsed strings.
- Extend unit-test coverage for empty and boundary-value parsing cases.

### Results and Evidence

<details><summary>Unit tests</summary>

The following regression scenarios were added to cover the updated parsing behavior:

- `test_parse_uname_string_empty_os_version`
- `test_parse_uname_string_empty_os_codename`
- `test_parse_uname_string_empty_os_name`
- `test_parse_uname_string_arch_linux_empty_version`
- `test_parse_uname_string_codename_closed_by_bracket`
- `test_parse_uname_string_empty_windows_version`

These tests validate that the parser now:

- accepts incomplete OS descriptor suffixes without aborting
- preserves expected parsed values for valid fragments
- returns empty strings or `NULL` consistently where appropriate

image

</details>

<details><summary>End-to-end validation</summary>

Validation was also performed with an enrolled test agent using `test_4093.py` and the `wazuh_testing` framework to confirm the manager still processes well-formed authenticated messages and rejects malformed unauthenticated ones as expected.

Checked paths:

- valid agent key: parsing flow is reached successfully
- invalid agent key: message is rejected during normal authentication handling

</details>

<details><summary>Fix verification</summary>

After applying the change:

- the new parsing edge cases are handled without aborting
- the added regression tests pass with the updated logic
- authenticated message processing continues to behave as expected

</details>

## Review Checklist

- [x] Code changes reviewed
- [ ] ...

### DIFF
--- a/src/unit_tests/shared/test_remoted_op.c
+++ b/src/unit_tests/shared/test_remoted_op.c
@@ -297,8 +297,6 @@ void test_parse_uname_string_empty_fields(void **state)
         { "Linux x86_64 [Arch Linux|arch: ","Arch Linux", "",NULL, NULL, NULL, NULL, "arch","x86_64" },
         /* os_codename empty — closed by ']' with no content          */
         { "Linux x86_64 [Ubuntu|ubuntu: 20.04 (]", "Ubuntu","20.04","","20", "04", NULL, "ubuntu","x86_64" },
-        /* Windows path — str_tmp empty, no closing ']'               */
-        { "Windows [Ver: ","Windows","",NULL, NULL, NULL, NULL,"windows",NULL },
     };
 
 #define ASSERT_FIELD(exp, got) \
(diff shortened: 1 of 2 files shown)


src/unit_tests/shared/test_remoted_op.c
MarcelKemp, line 1: Let's simplify UTs using setups and teardowns, or using loops.

PR #34880, opened 2026-03-10
Add second readdir check to prevent rootcheck false positives

### LINKED ISSUE #32653: Hundreds of false positives Hidden from stats, but showing up on readdir due to bug
|---|---|---|---|---|
| 4.13.1 | Wazuh component | Manager/Agent | Packages/Sources | Debian/Ubintu |

The bug is a result of a file being deleted and check running while the file is deleted so it's a classical race condition.
The solution is simple: instead of sequence: `stat-readdir` (or `readdir-stat`) do `stat-readdir-stat` or `readdir-stat-readdir` to confirm that the file is not being deleted while check is examining it.

anromerom:
## Description

Closes #32653

Rootcheck scan fires false positive `"Hidden from stats, but showing up on readdir. Possible kernel level rootkit."` alerts when files are deleted between `readdir()` and `lstat()` calls. This race condition is triggered by legitimate processes that delete files while rootcheck is scanning, producing hundreds of false alerts per scan cycle.

This PR replaces the original immediate-alert behavior with a deferred readdir re-verification approach:

1. When `lstat()` fails with `ENOENT`, the file is added to a suspect list instead of alerting immediately.
2. After the main `readdir` loop completes, the directory is reopened and a second `readdir` pass checks whether each suspect still appears in the listing.
  a. If the file is still listed but `lstat` returned `ENOENT`: rootkit alert.
  b. If the file is gone from the listing: no alert.

This preserves the original rootkit detection logic while eliminating false positives from legitimate file deletions.

## Proposed Changes

- **`src/rootcheck/check_rc_sys.c`**:
  - Add `#include <errno.h>` and `#define RC_ENOENT_SUSPECT (-2)` return code.
  - Modify `read_sys_file()`: on `lstat` failure with `ENOENT`, return `RC_ENOENT_SUSPECT` instead of alerting.
  - Extract alert formatting into `rc_emit_hidden_alert()` helper (used by both immediate alerts and deferred verification)
  - Modify `read_sys_dir()`: collect ENOENT suspects in a dynamic array, then perform a batch readdir re-verification after the main loop. Non-ENOENT errors (e.g., `EACCES`) still trigger the alert immediately.
  - Fallback: if the directory cannot be reopened for verification, all suspects are alerted (safe default)

- **`src/unit_tests/syscheckd/test_check_rc_sys.c`**: 5 cmocka unit tests
- **`src/unit_tests/syscheckd/CMakeLists.txt`**: Register `test_check_rc_sys` target (Linux only)

#### Unit Tests

<details><summary>Results - 5/5 PASS 🟢</summary>

| # | Test | Description | Result |
|---|------|-------------|--------|
| 1 | `test_enoent_deleted_file_no_alert` | File deleted between readdir and lstat. Second readdir confirms file is gone. No alert. | PASS |
| 2 | `test_enoent_hidden_file_alerts` | lstat returns ENOENT but file still appears in second readdir. Rootkit alert fires. | PASS |
| 3 | `test_eacces_still_alerts` | lstat fails with EACCES (not ENOENT). Alert fires immediately, no deferral. | PASS |
| 4 | `test_existing_file_no_alert` | Normal file exists. lstat succeeds. No alert, no retry. | PASS |
| 5 | `test_multiple_suspects_mixed` | Two ENOENT suspects: one deleted (gone from 2nd readdir → no alert), one hidden (still in 2nd readdir → alert). Verifies batch handling. | PASS |

```
[==========] tests: Running 5 test(s).
[ RUN      ] test_enoent_deleted_file_no_alert
[       OK ] test_enoent_deleted_file_no_alert
[ RUN      ] test_enoent_hidden_file_alerts
[       OK ] test_enoent_hidden_file_alerts
[ RUN      ] test_eacces_still_alerts
[       OK ] test_eacces_still_alerts
[ RUN      ] test_existing_file_no_alert
[       OK ] test_existing_file_no_alert
[ RUN      ] test_multiple_suspects_mixed
[       OK ] test_multiple_suspects_mixed
[==========] tests: 5 test(s) run.
[  PASSED  ] 5 test(s).
```

</details> 

#### Test script

A script was created to reproduce the race condition and compare unpatched and patched behavior across 10 iterations:

- **Setup**: 200 decoy files in `/tmp/rootcheck_race_test/`, 50ms `lstat` delay to widen the race window.
- **Procedure**: Start `wazuh-syscheckd` in foreground with debug mode, wait for rootcheck to begin scanning the test directory, delete all decoy files mid-scan, count `"Hidden from stats"` alerts.
- **Runs**: 10 iterations unpatched, then 10 iterations patched, with per-iteration timing and alert counts

**Results:**

| Metric | Unpatched (10 runs) | Patched (10 runs) |
|--------|--------------------|--------------------|
| False alerts (avg) | **193.6** | **0.0** |
| False alerts (min/max) | 193 / 198 | 0 / 0 |
| False alerts (total) | 1,936 | 0 |
| Scan duration avg (ms) | 20,618 | 20,685 |
| Scan duration min/max (ms) | 20,388 / 21,113 | 20,385 / 21,114 |

**Per-iteration detail:**
```
Iter   UNPATCHED             PATCHED
       Alerts  Time(ms)      Alerts  Time(ms)
 1       198    20439           0     21095
 2       193    20400           0     20428
 3       193    21113           0     20423
 4       193    21107           0     21104
 5       194    20388           0     21114
 6       193    20411           0     20385
 7       193    21085           0     20385
 8       193    20413           0     21098
 9       193    20410           0     20393
10       193    20410           0     20422
```

The second `readdir` pass adds negligible overhead because directory data is already cached in the kernel page cache from the first pass. On clean systems (no file deletions during scan), zero additional work is performed.

</details> 

### Artifacts Affected

- `wazuh-syscheckd`

### Configuration Changes

- N/A

### Documentation Updates

- N/A

### Tests Introduced

- **cmocka unit tests** (`src/unit_tests/syscheckd/test_check_rc_sys.c`): 5 test cases covering:
  - Deleted file → no alert
  - Hidden file → alert
  - Non-ENOENT error → immediate alert
  - Normal file → no alert
  - Mixed batch → correct per-file handling
- **CMakeLists.txt** updated to register `test_check_rc_sys`

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/src/rootcheck/check_rc_sys.c
+++ b/src/rootcheck/check_rc_sys.c
@@ -17,6 +17,8 @@
 /* Prototypes */
 static int read_sys_file(const char *file_name, int do_read);
 static int read_sys_dir(const char *dir_name, int do_read);
+static void rc_build_path(char *buf, size_t len,
+                          const char *dir, const char *name);
 #ifndef WIN32
 static void rc_emit_hidden_alert(const char *file_name);
 static int rc_collect_suspect(char ***suspects, size_t *count,
@@ -191,6 +193,18 @@ static int read_sys_file(const char *file_name, int do_read)
     return (0);
 }
 
+/* Build a full path from a directory and a file name.
+ * Handles the root directory ("/") as a special case to avoid "//name". */
+static void rc_build_path(char *buf, size_t len,
+                          const char *dir, const char *name)
+{
+    if (strlen(dir) == 1 && *dir == PATH_SEP) {
+        snprintf(buf, len, "%c%s", PATH_SEP, name);
+    } else {
+        snprintf(buf, len, "%s%c%s", dir, PATH_SEP, name);
+    }
+}
+
 #ifndef WIN32
 /* Emit the "Hidden from stats" rootkit alert for a given file */
 static void rc_emit_hidden_alert(const char *file_name)
@@ -264,13 +278,7 @@ static void rc_verify_suspects(const char *dir_name, char **suspects,
                         strcmp(suspects[s], entry->d_name) == 0) {
                     char full_path[PATH_MAX + 2];
 
-                    if (strlen(dir_name) == 1 && *dir_name == PATH_SEP) {
-                        snprintf(full_path, PATH_MAX + 1, "%c%s",
-                                 PATH_SEP, suspects[s]);
-                    } else {
-                        snprintf(full_path, PATH_MAX + 1, "%s%c%s",
-                                 dir_name, PATH_SEP, suspects[s]);
-                    }
+                    rc_build_path(full_path, PATH_MAX + 1, dir_name, suspects[s]);
 
                     mtdebug2(ARGV0, "File '%s' still listed in readdir "
                              "after lstat ENOENT. Alerting.", full_path);
@@ -300,13 +308,7 @@ static void rc_verify_suspects(const char *dir_name, char **suspects,
             if (suspects[s] != NULL) {
                 char full_path[PATH_MAX + 2];
 
-                if (strlen(dir_name) == 1 && *dir_name == PATH_SEP) {
-                    snprintf(full_path, PATH_MAX + 1, "%c%s",
-                             PATH_SEP, suspects[s]);
-                } else {
-                    snprintf(full_path, PATH_MAX + 1, "%s%c%s",
-                             dir_name, PATH_SEP, suspects[s]);
-                }
+                rc_build_path(full_path, PATH_MAX + 1, dir_name, suspects[s]);
 
                 rc_emit_hidden_alert(full_path);
                 _sys_errors++;
@@ -419,11 +421,7 @@ static int read_sys_dir(const char *dir_name, int do_read)
         }
 
         /* Create new file + path string */
-        if (strlen(dir_name) == 1 && *dir_name == PATH_SEP) {
-            snprintf(f_name, PATH_MAX + 1, "%c%s", PATH_SEP, entry->d_name);
-        } else {
-            snprintf(f_name, PATH_MAX + 1, "%s%c%s", dir_name, PATH_SEP, entry->d_name);
-        }
+        rc_build_path(f_name, PATH_MAX + 1, dir_name, entry->d_name);
 
         /* Check if file is a directory */
         if (lstat(f_name, &statbuf_local) == 0) {


src/rootcheck/check_rc_sys.c
```diff
@@ -333,9 +360,135 @@ static int read_sys_dir(const char *dir_name, int do_read)
             }
         }
 
+#ifndef WIN32
+        int rc = read_sys_file(f_name, do_read);
```
rjcausarano: Why are we not running this check on Windows?

anromerom: Hi Rodri! The alert based on a failing `lstat` command is Windows specific. It was already inside `ifndef WIN32` blocks (line 38-58). For that reason, the final `readdir` check is also only needed on non-Windows builds.

src/rootcheck/check_rc_sys.c
```diff
@@ -204,6 +227,10 @@ static int read_sys_dir(const char *dir_name, int do_read)
     short skip_fs;
 
 #ifndef WIN32
+    char **suspects = NULL;
+    unsigned int suspect_count = 0;
+    unsigned int suspect_capacity = 0;
```
lchico:
```suggestion
    size_t suspect_count = 0;
    size_t suspect_capacity = 0;
```

To avoid narrowing conversion on [line 382](https://github.com/wazuh/wazuh/pull/34880/changes#diff-3c9544e3a7d146546a9a6bf5cee7813939aeb847a56943d1902a1f2987dbcb3fR382)

src/rootcheck/check_rc_sys.c
```diff
@@ -333,9 +360,135 @@ static int read_sys_dir(const char *dir_name, int do_read)
...
+        int rc = read_sys_file(f_name, do_read);
+
+        if (rc == RC_ENOENT_SUSPECT) {
+            if (suspect_count >= suspect_capacity) {
+                size_t new_capacity = (suspect_capacity == 0) ? 32 : suspect_capacity * 2;
+                char **new_suspects = (char **)realloc(suspects, new_capacity * sizeof(char *));
+                if (new_suspects == NULL) {
+                    mterror(ARGV0, "Out of memory collecting ENOENT suspects.");
+
+                    unsigned int k;
```
lchico: Based on the previous changes, you will need to update this and the others (to stay consistent with **suspect_count** and **suspect_capacity**, which are now **size_t**).

```suggestion
                    size_t k;
```

src/rootcheck/check_rc_sys.c
```diff
@@ -333,9 +360,135 @@ static int read_sys_dir(const char *dir_name, int do_read)
...
+                    for (k = 0; k < suspect_count; k++) {
+                        free(suspects[k]);
```
lchico: **shared.h** already provides os_free(), which guards against double-free and sets the pointer to NULL automatically. Consider replacing free(suspects[k]) with os_free(suspects[k]) here and free(suspects) with     
  os_free(suspects).

src/rootcheck/check_rc_sys.c
```diff
@@ -333,9 +360,135 @@ static int read_sys_dir(const char *dir_name, int do_read)
...
+                suspects = new_suspects;
+                suspect_capacity = new_capacity;
+            }
+            suspects[suspect_count] = strdup(entry->d_name);
+            if (suspects[suspect_count] == NULL) {
+                mterror(ARGV0, "Out of memory duplicating suspect name.");
+
+                unsigned int k;
+                for (k = 0; k < suspect_count; k++) {
+                    free(suspects[k]);
```
lchico: Same as above — prefer os_free(suspects[k]) and os_free(suspects) for consistency with the project conventions.

src/rootcheck/check_rc_sys.c
```diff
@@ -333,9 +360,135 @@ static int read_sys_dir(const char *dir_name, int do_read)
...
+                return (-1);
+            }
+            suspect_count++;
+        }
+#else
         read_sys_file(f_name, do_read);
+#endif /* WIN32 */
     }
 
+#ifndef WIN32
```
lchico, line 491: The new suspect-related logic (lines 363–489) introduces two additional #ifndef WIN32 blocks inside read_sys_dir, which already uses this guard at lines 229 and 276. Consider consolidating — all four blocks are Linux-only. Grouping them under fewer, well-placed guards would make the Windows vs Linux split easier to follow at a glance.

src/rootcheck/check_rc_sys.c
anromerom: Hi Luis! The suggestion was applied

src/rootcheck/check_rc_sys.c
anromerom: Hi Luis! The suggestion was applied

src/rootcheck/check_rc_sys.c
anromerom: Hi Luis! The suggestion was applied

src/rootcheck/check_rc_sys.c
anromerom: Hi Luis!

This refactor is not direct as the previous WIN32 block has an else clause (line 400: `read_sys_file(f_name, do_read);`) which lies within the previous `while` loop. The second WIN32 block has the second `readdir` pass, which is exclusive for non-Windows builds.

cborla: Maybe we can split it into two functions and avoid this condition, the code would be clearer.
```c
static int read_sys_dir(const char *dir_name, int do_read)
{
#ifndef WIN32
    return read_sys_dir_posix(dir_name, do_read);
#else
    return read_sys_dir_win32(dir_name, do_read);
#endif
}
```
What do you think?

src/rootcheck/check_rc_sys.c
lchico: A small static helper would eliminate all three:

```
  static void rc_build_path(char *buf, size_t len,
                            const char *dir, const char *name)
  {
      if (strlen(dir) == 1 && *dir == PATH_SEP) {
          snprintf(buf, len, "%c%s", PATH_SEP, name);
      } else {
          snprintf(buf, len, "%s%c%s", dir, PATH_SEP, name);
      }
  }
```

  The same if (strlen(dir_name) == 1 && *dir_name == PATH_SEP) block appears at:
  - Line 314–317 (first readdir loop)
  - Lines 410–415 (second pass, dp2 != NULL)
  - Lines 451–457 (second pass, fallback/dp2 == NULL)

Issue #35133, opened 2026-03-26
Evidence of compliance with requirements — Monitoring and statistics


MarcelKemp:
# Description

This issue tracks the evidence required to validate compliance with all functional requirements, non-functional requirements, and implementation restrictions defined in https://github.com/wazuh/wazuh/issues/34711.

For every requirement listed below, a **valid evidence link** must be provided, or a **response block** must be included with enough technical detail to demonstrate that the requirement has been fulfilled.

---

## Evidence - Functional Requirements

1. **Global DB snapshot indexing into `wazuh-metrics-agents`**  
   - The Manager must periodically index snapshots of the global database state (`global.db`), based on the `agent` table.
   - The snapshot must essentially dump/index the `agent` table rows.
   - Target index name: `wazuh-metrics-agents`.
   - Evidence:
     - https://github.com/wazuh/wazuh/issues/35133#issuecomment-4140942731
<details><summary>wazuh-metrics-agents</summary>

#### Query: document count
```bash
curl -sk -u admin:admin 'https://localhost:9200/wazuh-metrics-agents/_count?pretty'
```
```json
{
  "count" : 8,
  "_shards" : { "total" : 1, "successful" : 1, "skipped" : 0, "failed" : 0 }
}
```
8 documents = 2 agents × 4 snapshot cycles.

#### Query: sample document for agent-1

```json
    "max_score": 1,
    "hits": [
      {
        "_index": ".ds-wazuh-metrics-agents-000001",
        "_id": "GS-Qhp0B06uUK_ft-908",
        "_score": 1,
        "_source": {
          "@timestamp": "2026-04-13T11:19:01Z",
          "wazuh": {
            "agent": {
              "id": "001",
              "name": "ubuntu24",
              "version": "v5.0.0",
              "groups": [
                "default"
              ],
              "status": "active",
              "status_code": 0,
              "registered_at": "2026-04-13T10:42:44Z",
              "last_seen": "2026-04-13T11:18:46Z",
              "register": {
                "ip": "0.0.0.0"
              },
              "host": {
                "ip": [
                  "10.0.2.15"
                ],
                "architecture": "x86_64",
                "os": {
                  "name": "Ubuntu",
                  "version": "24.04.3 LTS (Noble Numbat)",
                  "platform": "ubuntu"
                }
              },
              "config": {
                "group": {
                  "synced": true,
                  "hash": {
                    "md5": "ca35f9148a3273413eccb79f149f4757"
                  }
                }
              }
[...]

PR #35146, opened 2026-03-26
Fix: metrics normalization ECS mapping and indexer exception syntax for v5.0

Darioortegaleyva:
## Description

This PR fixes fatal errors during the Wazuh Indexer client initialization and resolves OpenSearch `strict_dynamic_mapping_exception` rejections for the metrics snapshots pipeline (Agents and Comms) in Wazuh v5.0.0.

| Field | Value |
|---|---|
| **Proposed Release** | v5.0.0 |
| **Issue** | wazuh/wazuh#35133 |
| **Internal Reference** | N/A |

---

### `framework/wazuh/core/indexer/indexer.py`

- **Refactored Exception Syntax:** Replaced keyword arguments (`code=`, `message=`, `extra_message=`) with positional arguments in all `WazuhException` and `WazuhIndexerError` calls.
  - **Reason:** In v5.0, the base exception class signature no longer supports these specific keyword arguments natively, which previously caused a fatal `TypeError` during the framework's indexer client initialization.
  - **Detail:** For `WazuhIndexerError`, `None` is explicitly passed as the second positional argument (the default message) to correctly assign the custom detail to the `extra_message` parameter (e.g., `raise WazuhIndexerError(2001, None, "Custom message")`).

### `framework/wazuh/core/indexer/metrics_snapshot.py`

- **Nested Dictionary Structures:** Refactored `_normalize_agent_doc` and `_normalize_comms_doc` to construct fully nested Python dictionaries (e.g., `"queue": {"usage": ...}`) instead of flat dotted keys (`"queue.usage"`).
  - **Reason:** OpenSearch indices with `dynamic: strict` reject flat dotted keys because they evaluate them as new, unmapped fields rather than navigating the pre-defined object hierarchy.
- **Data Type Casting:** Ensured all normalized fields match the exact ECS mappings defined in the requirements (e.g., using `str()` for `queue.usage` as it expects a `keyword`, and omitting string casting for `status_code` or network bytes so they are correctly indexed as integers/longs).

---

## Results and Evidence

The changes were validated in an isolated Wazuh Manager v5.0.0 (Ubuntu VM) with a local OpenSearch instance. Verification shows that the indexer client connects successfully without `TypeError` exceptions, and both Agent and Comms documents are accepted by OpenSearch in strict mode without parsing errors.

> **Note:** Comprehensive logs and UI evidence have been attached directly to Issue #35133.
```bash
# Validated successful ingestion of nested structures for Comms:
curl -sk -u admin:admin \
  -X GET "https://localhost:9200/.ds-wazuh-metrics-comms-*/_search?pretty&size=1"
```

Output excerpt demonstrating the accepted nested mapping:
```json
"_source": {
  "@timestamp": "2026-03-26T15:47:18Z",
  "wazuh.cluster.node": "node01",
  "wazuh.cluster.name": "wazuh",
  "wazuh.schema.version": "1",
  "events": {
    "module": "remoted"
  },
  "queue": {
    "usage": "42",
    "capacity": 131072
  }
}
```

---

## Artifacts Affected

- `framework/wazuh/core/indexer/indexer.py`
- `framework/wazuh/core/indexer/metrics_snapshot.py`

---

## Configuration Changes

N/A

---

## Documentation Updates

N/A

---

## Tests Introduced

- **Scope:** Manual Integration Tests (VM Environment)
- **Details:** Validated that the framework successfully initializes the indexer client reading from the keystore without crashing. Executed end-to-end data ingestion from `get_daemons_stats` and `WazuhDBQueryAgents` to the OpenSearch `wazuh-metrics-*` data streams, confirming complete compatibility with the strict index templates.

---

## Review Checklist

  - [ ] Linux
  - [ ] Windows
  - [ ] macOS
- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer

**General Checklist:**

- [x] Code changes reviewed


framework/wazuh/core/indexer/indexer.py
Miguevrgo, line 1: Why adding None?
The signature is     def __init__(self, code: int, extra_message: str = None, extra_remediation: str = None, cmd_error: bool = False,
                 dapi_errors: dict = None, ids: Union[list, set] = None, title: str = None, type: str = None):

framework/wazuh/core/indexer/tests/test_metrics_snapshot.py
Miguevrgo, line 1: Here we have big changes due to the change in fileds from nested to keydote from @Darioortegaleyva. It has been reviewed and tested, the changes were also needed for it to work.

Also removing the mask from the ip seems to be the only way because indexer team have a mapping strict with the field ip, if we want to include the mask we have to ask the indexer team

Miguevrgo (approved):
Here we have big changes due to the change in fileds from nested to keydote from @Darioortegaleyva. It has been reviewed and tested, the changes were also needed for it to work.

Also removing the mask from the ip seems to be the only way because indexer team have a mapping strict with the field ip, if we want to include the mask we have to ask the indexer team

Issue #35175, opened 2026-03-30
Vulnerability scanner coding style failing check in content_manager


jotacarma90:
# Description 

For the  4.14.5 branch, the indexer-connector code needs formatting 

https://github.com/wazuh/wazuh/actions/runs/23259939720

## Error

- [clang_format_errors.txt](https://github.com/user-attachments/files/26347219/clang_format_errors.txt)

## DoD 

- [ ] Fix coding style for router implementation code 
- [ ] Vulnerability scanner ITs must pass.

PR #35180, opened 2026-03-30
Fix clang-format version resolution in CI

jotacarma90:
## Description
This PR simplifies the `clang-format` resolution in CI.

Main changes:
- try a single Wazuh-hosted package from `packages.wazuh.com/deps`
- if the download or extraction fails, warn and continue with the previous runner-package flow
- execute the resolved `clang-format` binary explicitly in the check step
- avoid adding version and URL inputs to the action interface

## Validation
- YAML parsed successfully after the change
- workflow not executed end-to-end

## Workflow
```
Attempting to download clang-format from https://packages.wazuh.com/utils/clang-format/clang-format.tar.gz
clang-format version 22.1.1 (https://github.com/llvm/llvm-project.git fef02d48c08db859ef83f84232ed78bd9d1c323a)
```

Other executions:
https://github.com/wazuh/wazuh/actions/runs/23861568520

Related to #35175

### DIFF
--- a/src/shared_modules/content_manager/src/actionOrchestrator.hpp
+++ b/src/shared_modules/content_manager/src/actionOrchestrator.hpp
@@ -99,9 +99,7 @@ public:
          *
          */
         UpdateData(const UpdateType type, const int offset, const std::string& fileHash)
-            : type(type)
-            , offset(offset)
-            , fileHash(fileHash) {};
+            : type(type), offset(offset), fileHash(fileHash) {};
     };
 
     /**
--- a/src/shared_modules/content_manager/src/components/APIDownloader.hpp
+++ b/src/shared_modules/content_manager/src/components/APIDownloader.hpp
@@ -102,10 +102,7 @@ public:
      *
      * @param urlRequest Object to perform the HTTP requests to an API.
      */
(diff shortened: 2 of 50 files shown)


.github/actions/clang_format/action.yml
```diff
@@ -13,17 +13,68 @@ inputs:
...
+
+          echo "USE_WAZUH_CLANG_FORMAT=false" >> "${GITHUB_ENV}"
+          echo "CLANG_FORMAT_BIN=clang-format" >> "${GITHUB_ENV}"
+
+          if [ "$(uname -s)" != "Linux" ] || { [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "amd64" ]; }; then
+            echo "::warning::Wazuh clang-format package is only configured for Linux amd64. Falling back to runner package."
+            exit 0
+          fi
+
+          if [ -x "${TOOL_BIN}" ]; then
```
MiguelazoDS, line 34: Is this check intended for re-runs? because the file was not downloaded yet.

jotacarma90: This is because this action is called multiple times in some workflows, so the same one is reused.

PR #35305, opened 2026-04-07
SCA should not synchronize partial scan results with checks in not run state

### LINKED ISSUE #35249: SCA should not synchronize partial scan results with checks in "Not run" state
## Summary

On the first startup, the Security Configuration Assessment inventory is populated, but many checks are stored and synchronized with the `Not run` result. After an additional synchronization or a later scan cycle, those same checks change to `Not applicable`.

This suggests that SCA is exposing and synchronizing a partial scan state, even though `Not run` means the check has not been executed yet. As a result, the Manager/Indexer may temporarily receive incomplete compliance data.

Image

<details><summary>Steps</summary>

```shell
apt-get update
apt-get install -y wget curl lsb-release

wget -O artifact_urls.yaml https://packages-staging.xdrsiem.wazuh.info/nightly/5.0.0/artifact_urls_5.0.0-latest.yaml
sed -i '/^version:/d' artifact_urls.yaml
wget -O wazuh-install.sh https://packages-staging.xdrsiem.wazuh.info/nightly/5.0.0/installation-assistant/wazuh-install-5.0.0-latest.sh
bash ./wazuh-install.sh -a -d local -id

wget https://packages-staging.xdrsiem.wazuh.info/nightly/5.0.0/apt/pool/main/w/wazuh-agent/wazuh-agent_5.0.0-latest_amd64.deb
WAZUH_MANAGER="127.0.0.1" dpkg -i wazuh-agent_5.0.0-latest_amd64.deb
systemctl daemon-reload
systemctl enable wazuh-agent
systemctl start wazuh-agent
```

</details>

## Problem

SCA should not synchronize partial results.

A check in `Not run` state represents an unfinished evaluation, not a final compliance outcome. Synchronizing that state makes the Manager/Indexer consume incomplete data and can temporarily misrepresent the actual compliance posture of the host.

In practice, there should not be synchronized SCA states still marked as `Not run`. Synchronization should happen only after the scan has produced a complete and stable result set for the policy.

## Expected behavior

- SCA must not persist or synchronize partial scan results.
- Checks should not be synchronized while still in `Not run` state.
- Synchronization should occur only once the policy scan is complete.
- The first synchronized state must already contain final results for each check, such as:
  - `Passed`
  - `Failed`
  - `Not applicable`

## Current behavior

- On first startup, SCA inventory is synchronized with multiple checks in `Not run`.
- On a later synchronization or scan cycle, those checks change to `Not applicable`.
- This creates an inconsistent intermediate state in the synchronized inventory.

## Impact

- Incomplete compliance data is exposed to the Manager/Indexer.
- Users may see transient and misleading SCA results.
- The synchronized inventory does not reflect a complete scan snapshot.

## Requested changes

1. Review the SCA scan and synchronization flow to ensure only complete scan results are persisted/synchronized.
2. Prevent `Not run` checks from being included in synchronized stateful SCA documents.
3. Define whether scan completion must be treated as an atomic boundary for synchronization.
4. Add validation/tests to ensure a first synchronization cannot publish partial SCA results.

## Context

- Local Agent and Manager tested on Debian 13 (Trixie)
- Behavior observed on first boot of the local Agent
- Related documentation defines:
  - `Not run` as a check that has not been executed yet
  - `Not applicable` as a check that does not apply to the system

## Expected outcome

- SCA synchronization only publishes complete scan state
- No synchronized checks remain in `Not run`
- Compliance visibility is consistent from the first synchronization

jr0me:
## Description

Closes #35249

SCA was sending stateful results while the first scan was still in progress. Because of that, some checks were synchronized as `Not run`, and only later changed to their real final state, like `Not applicable`.

This PR changes that first-sync flow. Now SCA waits until it has the final scan data and sends the first stateful synchronization as a full snapshot, so the Manager/Indexer does not receive partial results.

## Proposed Changes

- Stop publishing stateful SCA documents before the first sync is completed.
- Build the first synchronization from the final snapshot already stored in the SCA database.
- Mark `first_sync_completed` only after that initial full sync finishes successfully.
- Keep the regular delta synchronization flow for later updates.
- Add unit tests to cover the initial snapshot path and to make sure stateful messages are held back before the first sync.

### Results and Evidence

The first synchronized SCA state now comes from a complete snapshot, so checks are not published in `Not run` just because the scan is still underway. This fixes the temporary inconsistent state described in the issue.

image

### Tests Introduced

- Unit tests for the initial full snapshot on the first sync.
- Unit tests to confirm later syncs still use delta mode.
- Unit tests to confirm stateful messages are skipped before the first sync.

### DIFF
--- a/src/wazuh_modules/sca/sca_impl/src/sca_impl.cpp
+++ b/src/wazuh_modules/sca/sca_impl/src/sca_impl.cpp
@@ -473,7 +473,7 @@ bool SecurityConfigurationAssessment::syncModule(Mode mode)
 
     if (result)
     {
-        LoggingHelper::getInstance().log(LOG_DEBUG, "SCA synchronization finished successfully.");
+        LoggingHelper::getInstance().log(LOG_INFO, "SCA synchronization finished successfully.");
     }
     else
     {
@@ -1096,6 +1096,8 @@ bool SecurityConfigurationAssessment::synchronizeDatabaseSnapshot(bool increaseV
                     statefulMessage.dump(),
                     check["version"].get<uint64_t>()
                 );
+
+                LoggingHelper::getInstance().log(LOG_DEBUG_VERBOSE, "Stateful event queued: " + statefulMessage.dump());
             }
         }
 
--- a/.github/workflows/5_testintegration_sca-tier-0-1-lin.yml
+++ b/.github/workflows/5_testintegration_sca-tier-0-1-lin.yml
@@ -13,6 +13,7 @@ on:
         - "src/config/src/wmodules-sca.c"
         - "src/wazuh_modules/src/wm_sca.c"
         - "src/wazuh_modules/src/wm_sca.h"
+        - "src/wazuh_modules/sca/**"
         - "src/Makefile"
         - "tests/integration/conftest.py"
         - "tests/integration/test_sca/**"
--- a/.github/workflows/5_testintegration_sca-tier-0-1-win.yml
+++ b/.github/workflows/5_testintegration_sca-tier-0-1-win.yml
@@ -13,6 +13,7 @@ on:
         - "src/config/src/wmodules-sca.c"
         - "src/wazuh_modules/src/wm_sca.c"
         - "src/wazuh_modules/src/wm_sca.h"
+        - "src/wazuh_modules/sca/**"
         - "src/win32/**"
         - "src/Makefile"
         - "tests/integration/conftest.py"


src/wazuh_modules/sca/sca_impl/src/sca_impl.cpp
```diff
@@ -423,17 +435,32 @@ bool SecurityConfigurationAssessment::syncModule(Mode mode)
     }
 
     // Log
-    LoggingHelper::getInstance().log(LOG_INFO, "Starting SCA synchronization.");
+    LoggingHelper::getInstance().log(LOG_DEBUG, "Starting SCA synchronization.");
```
jpcerrone: Any reason to change this? I think the other modules also log their synchronization attempts in INFO

jr0me: No, good catch, I was changing the new logs and missed this.

PR #35350, opened 2026-04-08
Improve translation mechanism

### LINKED ISSUE #23656: Vulnerability Detector - Improve the translation mechanism to allow more complex version fields
## Description
The translation functionality should be expanded to cover a wider spectrum of solutions.

Using the translation schema as a baseline:
```json
{
  "target": [
    "Platform 1",
    "Platform n"
  ],
  "source": {
    "vendor": "vendor name",
    "product": "package name",
    "version": "package version"
  },
  "translation": [
    {
      "vendor": "translated vendor name",
      "product": "translated package name",
      "version": "translated package version"
    }
  ],
  "action": [
    "replace_vendor",
    "set_version_if_product_matches",
    "replace_product"
  ]
}
```

## Definitions
- Add a new "extended" translations
  - This is going to be a [new resource](https://github.com/wazuh/wazuh/blob/623091670c4b2c201e93b875a7dfd63440c75169/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/eventDecoder.hpp#L54-L58), under the name "extended-TID-xxx" 
- Maintain older translation functionality
- Move the business logic to a separate class (out from the databaseFeedManager and package scanner)

>[!NOTE]
> Take functionality extension into account, for the addition of new fields in both "translation" and "source" objects. eg. "format should be deb".

### Source
- Could contain 0 to N capture groups on each field
- The version field is optional
  - If the translation doesn't need to check the version value, it won't be provided

### Translation
- All fields are optional
  - If not provided, the value from syscollector is used instead
- If provided, should be constructed with placeholders, hardcoded strings or both
  - Eg: "$(VERSION_0)$(PRODUCT_2)$(PRODUCT_3)"
  - Eg: "NEW VALUE"
  - Eg: "1.$(VENDOR_1)_100"

### Other requirements
- Don't use flatbuffers in the extended implementation
  - It proved to be difficult to use and increased the risk of errors
  - Given the amount of translations available it doesn't necessarily improve memory usage by a significant factor (while degrading performance related to the parsing)
  - It forced us to use a hard schema, which made the implementation of this functionality much harder
- Simplify the translations:
  - Remove unused fields or re-implement in a generic way
    - action array
    - update
    - target_hw
    - msu_name
    - sw_edition

## Tasks
Create multiple PRs and point them to an intermediary branch

### Wazuh
- [ ] Move the business logic to a new "translation" class
  - [ ] Add an interface
  - [ ] Add the "legacy" translation functionality
  - [ ] Changes tested
- [ ] Add the new functionality
  - [ ] Add the extended translation functionality
  - [ ] Changes tested

### ID
- [ ] Translation adaptation
  - [ ] Remove unused fields
  - [ ] Add extended translations
  - [ ] Test integration

## DoD
- [ ] Translation documentation updated
- [ ] Translations updated
- [ ] Feature implemented
- [ ] Testing updated 
  - [ ] UTs
  - [ ] Component

MiguelazoDS:
## Description 

Java products have a particular and complex CPE format that changed multiple times from version to version. 
These changes here used the update field (unused in 4.x) to create a more complex version fields and parse those complex CPEs 

> [!NOTE]
> I won't merge until we define what to do with Java content information

## Proposed Changes

- Allow multiple groups in regex in all fields (product, vendor, version). 

- Implemented a baseline integration test to validate current translation before changes 
image

- Backward compatible by using the "update" field in translations. 

## Testtools delivery 

Successfully test tools delivery 

https://github.com/wazuh/wazuh/actions/runs/28382113924


src/wazuh_modules/vulnerability_scanner/testtool/scanner/vd_reduced_feed.tar.gz
TomasTurina, line 1: Please avoid using binary files.

MiguelazoDS: The previous json was restored and reduced in size.

MiguelazoDS:
Claude suggested something like this to test 

```json
{"name":"TID-0184","offset":3600000,"version":1,"inserted_at":"2026-07-17T00:00:00.000000Z","payload":{"action":["replace_vendor","replace_product","set_version_if_product_matches"],"source":{"vendor":"^Acme$","product":"^AcmeApp$","product_extended":"^AcmeApp (Pro|Enterprise) Edition$","version_extended":"^(\\d+)\\.(\\d+)\\.\\d+$"},"target":["darwin"],"translation":[{"vendor":"acme","product":"acmeapp","product_extended":"acmeapp_$(PRODUCT1)","version_extended":"$(VERSION1).$(VERSION2)"}]}}
```

But I don't think that is necessary, I don't see as requiring multi group to create a product. So the guard still does its job perfectly. But I'll keep it as a defensive code just in case.

jotacarma90 (approved):
LGTM

PR #35461, opened 2026-04-16
Fix SCA integration tests flakiness and deadlocks on Windows

Darioortegaleyva:
## Summary

SCA integration tests introduced alongside #35305 failed consistently on Windows (and occasionally on Linux) due to three independent root causes exposed by the new `agent_module_limits` handshake. All four test files in `test_sca/test_basic/` were affected, resulting in flaky behavior and timeouts.

This PR addresses these root causes by fixing IPC conflicts on Windows, ensuring deterministic SCA database initialization, and following internal feedback to maintain logging consistency in the C++ core.

---

### 1. Root Causes Fixed (Tests & Infrastructure)

**Windows IPC Endpoint Race Condition (The "Flakiness")**

Windows SCM reports `STOPPED` before the process fully exits. When the next test starts a new instance while the previous process is still alive, both compete for the SCA IPC endpoint. The new process wins the SCM race but loses the IPC race, causing the scan thread to never initialize (resulting in 40s of silence in logs and a Phase 4 deadlock).

- **Fix:** Added `_wait_for_agent_gone(timeout, raise_on_timeout=False)`, a `psutil`-based polling loop that blocks execution until the Wazuh agent process actually disappears. Exposed as a `wait_for_agent_gone` pytest fixture in the parent conftest so child conftests consume it via dependency injection — zero duplication. It is invoked in both `clean_sca_db` (setup, `raise_on_timeout=True`) and a new teardown fixture `_wait_agent_exit_after_test` (best-effort, `raise_on_timeout=False`).

**Missing `first_sync_completed=1` Seeding → SCA Skips the Scan**

When a previous SCA DB exists with `first_sync_completed=0`, the module takes Path B (`synchronizeDatabaseSnapshot`) instead of Path A (`PushStateful`). Path B remains blocked waiting for a synchronization that does not occur in standalone test environments.

- **Fix:** Consolidated the use of the `clean_sca_db` fixture in `conftest.py` for all test files, ensuring a clean state and forcing the initial scan through metadata seeding.

**`IndexError` in `find_result_for_a_given_id`**

`FileMonitor` with `accumulations=N` returned a simple tuple instead of a list of tuples when fewer than N results were found. Iterating over a string tuple caused an `IndexError` when validating conditions.

- **Fix:** The result is normalized to a list before iteration in `find_result_for_a_given_id`.

**Additional Fixes:**
- Increased Linux Phase 2 timeout from **15s to 30s** to handle CI load spikes.
- Replaced non-deterministic waits with **active process state polling** to ensure clean IPC transitions between tests.
- Corrected policy file paths in Windows YAML test cases (added `ruleset/sca/`).
- Removed duplicate `clean_sca_database` fixture from `test_basic/conftest.py` (covered by `clean_sca_db` in parent conftest).
- Merged `_wait_for_agent_gone` and `_wait_for_agent_processes_gone` into a single function with a `raise_on_timeout` parameter — setup callers pass `True`, teardown callers use the default `False`.
- Replaced hardcoded process name set with `AGENT_DAEMON` and `WAZUH_AGENT_WIN` constants from `wazuh_testing.constants.daemons`.
- Applied `for/else` + `last_error` pattern in `clean_sca_db` retry loop to surface DB seeding failures explicitly.

---

### 2. C++ Logging Consistency (Feedback Addressed)

**Revert Log Levels to `DEBUG_VERBOSE`**

- During investigation, log levels in `sca_event_handler.cpp` and `sca_impl.cpp` were temporarily changed to `DEBUG`.
- **Action:** Reverted to `DEBUG_VERBOSE` to comply with Wazuh production standards and avoid unnecessary noise, maintaining test stability through the new improved synchronization logic.

---

### Test Evidence

- **3/3** consecutive CI runs on Windows passed successfully after applying all review fixes.
- Average run duration: **3–8 min**.

**List of actions (Evidence):**
- [Action 25165508273](https://github.com/wazuh/wazuh/actions/runs/25165508273/job/73770478715)
- [Action 25165522857](https://github.com/wazuh/wazuh/actions/runs/25165522857/job/73770531229)
- [Action 25165537164](https://github.com/wazuh/wazuh/actions/runs/25165537164/job/73770583504)

---

### Test Plan

- [x] Manual execution of the CI workflow on Windows — **All Green**.
- [x] Linux test verification — **No regressions**.
- [x] `test_compliance_format`, `test_mitre_payload`, `test_scan_results`, and `test_validate_remediation` confirmed as deterministic with logs set to `DEBUG_VERBOSE`.
- [x] Review feedback fully addressed — duplicates removed via pytest fixture DI, error propagation improved, constants used from `wazuh_testing.constants.daemons`, wait functions unified with `raise_on_timeout` parameter.

### DIFF
--- a/tests/integration/test_sca/conftest.py
+++ b/tests/integration/test_sca/conftest.py
@@ -3,26 +3,295 @@
  Created by Wazuh, Inc. <info@wazuh.com>.
  This program is free software; you can redistribute it and/or modify it under the terms of GPLv2
 """
+import psutil
 import pytest
-import sys
 import re
+import sqlite3
+import subprocess
+import sys
+import time
+from pathlib import Path
 
-from wazuh_testing.tools.monitors import file_monitor
-from wazuh_testing.modules.modulesd.sca import patterns
+from wazuh_testing.constants.paths import WAZUH_PATH
 from wazuh_testing.constants.paths.logs import WAZUH_LOG_PATH
-from wazuh_testing.utils import callbacks
 from wazuh_testing.constants.platforms import WINDOWS
+from wazuh_testing.modules.modulesd.sca import patterns
+from wazuh_testing.tools.monitors import file_monitor
+from wazuh_testing.utils import callbacks, services
+from wazuh_testing.utils.services import control_service
+
+
+SCA_DB_DIR = Path(WAZUH_PATH, 'queue', 'sca', 'db')
+
+_WAZUH_AGENT_PROCESS_NAMES = {'wazuh-agent.exe', 'WazuhSvc.exe', 'wazuh-agentd'}
+
+
+def _wait_for_agent_gone(timeout: int = 60) -> None:
+    """Block until no wazuh-agent process is alive or timeout expires.
+
+    Windows SCM reports STOPPED before the process fully exits. Starting a new
+    service instance while the old process is alive causes both to compete for
+    the same IPC endpoint used by the SCA C++ module, preventing the scan from
+    starting (Phase 4 deadlock).
+    """
+    deadline = time.time() + timeout
+    while time.time() < deadline:
+        alive = [p.info['name'] for p in psutil.process_iter(attrs=['name'])
+                 if (p.info.get('name') or '') in _WAZUH_AGENT_PROCESS_NAMES]
+        if not alive:
+            return
+        time.sleep(0.5)
+
+
+# Patterns that distinguish a healthy SCA startup from the known
+# "getdoclimits handshake stall" failure mode. If Phase 3 times out with a
+# high count for 'AGCOM Module limits not configured' (or 'Failed to query
+# agentd via agcom_dispatch') and zero for 'Module limits received from
+# manager', it is not a test flake: the agent's handshake with the manager
+# never produced a 'limits.sca' block, so SCA stays forever in the loop at
+# src/wazuh_modules/src/wm_sca.c:590-601 waiting for agent_module_limits.
+# limits_received to flip to true (src/client-agent/src/start_agent.c:142).
+_SCA_STARTUP_HIGHLIGHTS = {
+    'SCA tagged lines': r':sca\b',
+    'SCA module enabled (DEBUG)': r'SCA module enabled',
+    'SCA module running (INFO)': r'SCA module running',
+    'Starting SCA module': r'Starting SCA module',
+    'agcom_dispatch failure': r'Failed to query agentd via agcom_dispatch',
+    'AGCOM limits not configured': r'AGCOM Module limits not configured',
+    'AGCOM err response': r'Agentd returned error',
+    'Handshake JSON parsed': r'Module limits received from manager',
+    'Handshake JSON absent': r'No handshake JSON after ACK',
+    'Handshake retry': r'Error parsing handshake JSON',
+    'Policy file not found': r'Policy file.*not found',
+    'Ruleset folder open fail': r'Could not open the default SCA ruleset folder',
+    'Policy scan started': r'Starting Policy checks evaluation',
+}
+
+
+def _wait_service_running(timeout: int) -> None:
+    """Poll until the Wazuh service reports RUNNING.
+
+    On Windows, delegates to ``wait_expected_daemon_status`` which queries
+    SCM (no socket validation). On Linux, polls ``wazuh-control status``
+    directly, checking only that every daemon line says "is running…" —
+    without validating that sockets exist on disk (which fails on agents
+    that don't create manager-only sockets like ``download`` / ``control``).
+    """
+    if sys.platform == WINDOWS:
+        services.wait_expected_daemon_status(running_condition=True, timeout=timeout)
+        return
+
+    from wazuh_testing.constants.paths.binaries import WAZUH_CONTROL_PATH
+    deadline = time.time() + timeout
+    while time.time() < deadline:
+        try:
+            proc = subprocess.run(
+                [WAZUH_CONTROL_PATH, 'status'],
+                stdout=subprocess.PIPE, stderr=subprocess.PIPE, timeout=10,
+            )
+            lines = [l for l in proc.stdout.decode(errors='ignore').splitlines() if l.strip()]
+            if lines and all('is running...' in l for l in lines):
+                return
+        except Exception:
+            pass
+        time.sleep(1)
+    raise TimeoutError(f"Service did not reach RUNNING within {timeout}s")
+
+
+def _format_diagnostics(diag: dict) -> str:
+    """Render a diagnostics dict from collect_service_diagnostics as a block."""
+    sections = [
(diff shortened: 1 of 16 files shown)


tests/integration/test_sca/test_basic/conftest.py
```diff
@@ -5,17 +5,127 @@
...
+            row = cursor.fetchone()
+            conn.close()
+            if row and row[0] == 1:
+                break
+        except Exception:
+            pass
+
+        if sys.platform == WINDOWS:
+            time.sleep(2)
+
```
MarcelKemp, line 105: I think this function is duplicated by `clean_sca_db` and, furthermore, isn’t referenced anywhere.
```suggestion
```

tests/integration/test_sca/test_basic/conftest.py
```diff
@@ -5,17 +5,127 @@
...
+    while time.time() < deadline:
+        alive = [
+            p.info['name']
+            for p in psutil.process_iter(attrs=['name'])
+            if (p.info.get('name') or '') in _WAZUH_AGENT_PROCESS_NAMES
+        ]
+        if not alive:
+            return
+        time.sleep(0.5)
+
```
MarcelKemp: Ditto (?). 
It seems that some of the functions created in this conftest are duplicated in `tests/integration/test_sca/conftest.py`. Could you have a look and remove any test cases that aren’t necessary (or that are duplicates)?

tests/integration/test_sca/conftest.py
```diff
@@ -3,26 +3,295 @@
...
+    starting (Phase 4 deadlock).
+    """
+    deadline = time.time() + timeout
+    while time.time() < deadline:
+        alive = [p.info['name'] for p in psutil.process_iter(attrs=['name'])
+                 if (p.info.get('name') or '') in _WAZUH_AGENT_PROCESS_NAMES]
+        if not alive:
+            return
+        time.sleep(0.5)
+
```
MarcelKemp, line 57: If the setup fails silently, the test fails later during an unrelated phase and the actual cause is no longer visible in the log.
```suggestion
    raise TimeoutError(
        f"Wazuh agent process still alive after {timeout}s: {alive}. "                   
        f"IPC endpoint will collide with the next service start."
    )

```

tests/integration/test_sca/conftest.py
```diff
@@ -3,26 +3,295 @@
...
+            cursor = conn.execute(
+                "SELECT value FROM sca_metadata WHERE key = 'first_sync_completed'")
+            row = cursor.fetchone()
+            conn.close()
+            if row and row[0] == 1:
+                break
+        except Exception:
+            pass
+
+    yield
```
MarcelKemp, line 190:
```suggestion
      except Exception as e:
          last_error = e                                                                   
  else:
      raise RuntimeError(                                                                  
          f"Failed to seed SCA DB after 3 attempts. Last error: {last_error!r}"
      )
```

tests/integration/test_sca/test_basic/test_validate_remediation.py
```diff
@@ -68,13 +68,14 @@
...
 
+
 # Helper to find the result for a given id in the accumulated results
 def find_result_for_a_given_id(id: int, results: list[tuple[str, str, str]]) -> tuple[str, str, str] | None:
     # When only one event was captured the monitor returns a bare tuple instead of a list of tuples.
     if isinstance(results, tuple):
         results = [results]
     for result in results:
-        if result[0] == str(id):
+        if isinstance(result, tuple) and len(result) >= 3 and result[0] == str(id):
```
MarcelKemp, line 78: Is this necessary? Normalising the previous line already ensures that each result is a tuple of three elements; the extra guard hides any unexpected forms rather than causing them to be skipped.

tests/integration/test_sca/conftest.py
```diff
@@ -3,26 +3,295 @@
...
 from wazuh_testing.constants.platforms import WINDOWS
+from wazuh_testing.modules.modulesd.sca import patterns
+from wazuh_testing.tools.monitors import file_monitor
+from wazuh_testing.utils import callbacks, services
+from wazuh_testing.utils.services import control_service
+
+
+SCA_DB_DIR = Path(WAZUH_PATH, 'queue', 'sca', 'db')
+
+_WAZUH_AGENT_PROCESS_NAMES = {'wazuh-agent.exe', 'WazuhSvc.exe', 'wazuh-agentd'}
```
MarcelKemp: Couldn't we import the `wazuh_testing.constants` module? Instead of keeping the list in the conftest.

Check if the following constants would work:
- `from wazuh_testing.constants.daemons import AGENT_DAEMON`

MarcelKemp (changes requested):

tests/integration/test_sca/test_basic/conftest.py
Darioortegaleyva: Done, removed. You're right, grep confirmed it had zero references in any test file, and the logic is already covered by clean_sca_db in the parent conftest

tests/integration/test_sca/test_basic/conftest.py
Darioortegaleyva: Partially addressed. _wait_for_agent_processes_gone can't be removed because daemons_handler, a local fixture override defined in this same file depends on it. Since test_sca/ has no __init__.py, there's no clean way to import _wait_for_agent_gone from the parent conftest

tests/integration/test_sca/conftest.py
Darioortegaleyva: Tried this, but it broke Windows CI: wazuh-agent.exe legitimately takes >60s to die after SCM reports STOPPED, and _wait_for_agent_gone is also called from _wait_agent_exit_after_test (teardown), causing every test to fail at teardown. Reverted the raise in the general function. The raise TimeoutError was kept in _wait_for_agent_processes_gone inside test_basic/conftest.py, which is only ever called from daemons_handler during setup where a timeout is a real blocker

tests/integration/test_sca/conftest.py
Darioortegaleyva: Done. Applied the for/else pattern with last_error as suggested

tests/integration/test_sca/test_basic/test_validate_remediation.py
Darioortegaleyva: Done, removed. You're right, the results = [results] normalisation above already guarantees every element in the loop is a tuple.

tests/integration/test_sca/conftest.py
Darioortegaleyva: Done. AGENT_DAEMON and WAZUH_AGENT_WIN are now imported from wazuh_testing.constants.daemons in both conftest files. WazuhSvc.exe remains hardcoded, there's no corresponding constant in the module.

MarcelKemp (approved):

PR #35475, opened 2026-04-16
Prevent agent.host.ip from being silently dropped when agent IP is empty

### LINKED ISSUE #35251: Missing fields from inventory harvester indexed documents
## Description

The `agent.host.ip` field is not appearing in inventory documents indexed by the inventory harvester (packages, processes, ports, etc.), despite being defined in the index templates and documented as available.

The inventory harvester element builders correctly populate `agent.host.ip` when the agent IP is not "any", but the JSON serialization logic omits the host object when other Host struct fields remain empty (architecture, hostname, os).

## Tasks

- Investigate JSON serialization logic in reflectiveJson to understand why host object is omitted when only ip field is populated.
- Fix the serialization to include agent.host object when `agent.host.ip` has a value (even if other host fields are empty).
- Update all affected inventory element test cases to include `agent.host.ip` in expected results.
- Verify fix works for all inventory types (packages, processes, ports, services, etc.).
- Run existing QA tests to ensure no regressions.
- Add integration test that validates `agent.host.ip` appears in indexed documents.

## DoD

- `agent.host.ip` field appears in all inventory documents when agent IP is available (not "any").
- All existing unit and integration tests pass.
- Updated test data includes `agent.host.ip` in expected results.
- New integration test validates field presence in indexed documents.
- Fix verified for all affected inventory types (packages, processes, ports, browser extensions, groups, hotfixes, interfaces, protocols, services, users).
- Manual validation shows field appearing in Wazuh dashboard IT Hygiene views.

Related IDR: https://github.com/wazuh/external-devel-requests/issues/6690

jotacarma90:
## Summary

Two-layer fix for the missing `agent.host.ip` field confirmed in wazuh/external-devel-requests#6690.

### Layer 1 — `remoted/secure.c` (root cause)

When `HandleSecureMessage` forwards a syscollector/syscheck event to the router, it reads the agent IP from the keystore's registered IP (`keys.keyentries[agentid]->ip->ip`). For agents enrolled without a fixed address this value is the literal string `"any"`, which propagates through the FlatBuffer pipeline until inventory harvester's element builders discard it — leaving `agent.host.ip` absent from every indexed document for that agent.

The actual sender IP is already extracted into `srcip` from `message->addr` earlier in the same function. The fix uses it as a fallback when the registered IP is `"any"`:

```c
os_strdup(keys.keyentries[agentid]->ip->ip, agent_ip);

/* If the agent registered with "any" IP, use the actual source IP of the message */
if (strcmp(agent_ip, "any") == 0 && srcip[0] != '\0') {
    os_free(agent_ip);
    os_strdup(srcip, agent_ip);
}
```

The `srcmsg` used for the internal wazuh-db queue is intentionally left unchanged to preserve legacy routing behaviour.

### Layer 2 — inventory harvester element builders (defensive)

All 16 element builders (13 system-inventory types + 3 FIM types) previously set `host.ip = ""` when `agentIp()` returned an empty string, which then caused `isEmpty(host) == true` and silently dropped the entire `host` object from the serialized JSON. Added `!agentIp.empty()` guard to make the intent explicit:

```cpp
// before
if (auto agentIp = data->agentIp(); agentIp.compare("any") != 0)

// after
if (auto agentIp = data->agentIp(); !agentIp.empty() && agentIp.compare("any") != 0)
```

### Layer 3 — regression tests (`reflectiveJson_test.cpp`)

Three unit tests verify that `isEmpty()` and `serializeToJSON()` correctly handle a reflectable struct where only a subset of fields is populated (the `Host`-with-only-`ip` pattern), so any future regression in the serialization path is caught immediately.

## Root cause chain

```
Agent enrolled with no fixed IP
  → client.keys stores "any"
  → secure.c reads "any" from keystore  ← fix here (Layer 1)
  → router/schemaAdapter appends "any" to FlatBuffer JSON
  → systemContext::agentIp() returns "any"
  → element builder skips host.ip assignment  ← hardened (Layer 2)
  → isEmpty(host) == true
  → "host" key omitted from serialized document
  → agent.host.ip missing in index
```

## Evidences
- Using a full environment with 4.14.5, register an agent with any IP:
```
001 ubuntu24-2 any 543c32612a96d2399d7592d21e976a8802a14a8f5106aaee9e78e6d05ca40066
```
- No ip in dashboard:
image

After updating manager with this branch changes:
- `/var/ossec/etc/client.keys`
```
004 ubuntu24-2 any f1a0195d6ce5b161b8394f1e30f61e04185bd1e9b11595a61a201a32b20566c6
```

image

## Test plan

- [x] Agent enrolled with `"any"` IP: after fix, `agent.host.ip` is populated with the real connection IP in indexed documents.
- [x] Agent enrolled with explicit IP: behaviour unchanged.
- [x] `IsEmptyReturnsFalseForPartiallyPopulatedNestedStruct` passes.
- [x] `IsEmptyReturnsTrueWhenAllFieldsAreEmpty` passes.
- [x] `PartiallyPopulatedNestedStructIsNotOmittedFromOutput` passes.
- [x] All existing `remoted`, `reflectiveJson`, and `systemInventoryUpsertElement` unit tests pass.

Closes #35251

### DIFF
--- a/src/remoted/secure.c
+++ b/src/remoted/secure.c
@@ -923,6 +923,12 @@ STATIC void HandleSecureMessage(const message_t *message, w_indexed_queue_t * co
 
     key_unlock();
 
+    /* Agents registered without a fixed address use "any"; fall back to the real connection IP. */
+    if (strcmp(agent_ip, "any") == 0 && srcip[0] != '\0') {
+        os_free(agent_ip);
+        os_strdup(srcip, agent_ip);
+    }
+
     if (sock_idle >= 0) {
         _close_sock(&keys, sock_idle);
     }
--- a/src/remoted/secure.c
+++ b/src/remoted/secure.c
@@ -921,6 +921,13 @@ STATIC void HandleSecureMessage(const message_t *message, w_indexed_queue_t * co
     os_strdup(keys.keyentries[agentid]->name, agent_name);
     os_strdup(keys.keyentries[agentid]->ip->ip, agent_ip);
 
+    /* If the agent registered with "any" IP, use the actual source IP of the message
+     * so that downstream consumers (e.g. inventory harvester) can populate agent.host.ip */
+    if (strcmp(agent_ip, "any") == 0 && srcip[0] != '\0') {
+        os_free(agent_ip);
+        os_strdup(srcip, agent_ip);
+    }
+
     key_unlock();
 
     if (sock_idle >= 0) {
--- a/src/shared_modules/utils/tests/reflectiveJson_test.cpp
+++ b/src/shared_modules/utils/tests/reflectiveJson_test.cpp
@@ -390,4 +390,65 @@ TEST_F(ReflectiveJsonTest, NumericBufferZeroing)
     EXPECT_EQ(json.front(), '{');
     EXPECT_EQ(json.back(), '}');
 }
+
+TEST_F(ReflectiveJsonTest, IsEmptyReturnsFalseForPartiallyPopulatedNestedStruct)
+{
+    struct NestedData
+    {
+        std::string_view stringField;
+        int64_t intField;
+
+        REFLECTABLE(MAKE_FIELD("stringField", &NestedData::stringField),
+                    MAKE_FIELD("intField", &NestedData::intField));
+    };
+
+    NestedData obj;
+    obj.stringField = "someValue";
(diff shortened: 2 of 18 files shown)


Antoniogm03 (dismissed):
LGTM!

Antoniogm03 (approved):

PR #35424, opened 2026-04-14
wazuh-engine: `Json::SetString` strip trailing null bytes

matigarciadev:
## Description

`Json::setString` was refactored to accept `std::string_view` and pass its length explicitly to RapidJSON. This introduced a regression: when the `string_view` carried a trailing null byte (e.g. originating from a legacy Unix socket that sends `strlen + 1` bytes), RapidJSON would faithfully store the `\0` as part of the JSON string value, serialising it as `\u0000` in the output.

This caused `event.original` and any other field written through `setString` with a socket-derived buffer to contain a `\u0000` suffix, breaking all downstream decoders that perform string matching or JSON parsing on that field.

### Bugs fixed

- **`Json::setString` — strip trailing null bytes before storing** (`src/engine/source/base/src/json.cpp`):
  Added a loop that removes any trailing `\0` bytes from the `string_view` length before constructing the RapidJSON value. This restores the behaviour of the previous implementation, which achieved the same effect implicitly by converting to `std::string` and using `.c_str()` (causing RapidJSON to call `strlen` internally and stop at the first `\0`).

- **Unit tests added** (`src/engine/source/base/test/src/unit/json_test.cpp`):
  New test `SetStringStripsTrailingNullBytes` covering:
  - Single trailing `\0` (the real-world socket case)
  - Multiple trailing `\0` bytes
  - String composed entirely of `\0` bytes (becomes empty)
  - Normal string without `\0` (unchanged)
  - Empty string (unchanged)
  - `\0` embedded in the middle (not stripped — only trailing nulls are removed)

### Root cause

`OS_SendUnix` (in `src/shared/os_net/os_net.c`) uses `strlen(msg) + 1` when `size == 0`, so every message dispatched through the legacy Unix socket includes its null terminator in the byte count. After decryption in remoted, `msg_length` reflects this and the `\0` is preserved through `memcpy` into `evt_item_t`, through the HTTP POST bulk body, and finally into `parseLegacyEvent` as a `string_view` whose `.size()` includes the extra byte. The old `setString` implementation neutralised this by chance; the new one exposed it.

### Results and Evidence

Before fix in archived events (standard-wazuh-events-v5.json):
```
[...]

PR #35658, opened 2026-04-23
Fix agent 5x sends trailing null byte 0 in messages

### LINKED ISSUE #35474: Agent 5.x sends trailing null byte (\0) in messages
## Description

Agent 5.x appends a trailing null byte (`\0`) to messages, unlike 4.x. This breaks JSON decoding and affects multiple ingestion paths, although it may work in isolated cases like `logtest`.

The engine has been patched to tolerate this, but inconsistencies may still impact other components (e.g., remoted, inventory sync, modules).

Additionally, we need to assess the inverse scenario:
**what happens if a 5.x-like message is received *without* the trailing `\0`**.

This is relevant because:

* The manager may implicitly rely on the null terminator.
* An attacker with a valid agent key could craft messages mimicking 5.x **without `\0`**.
* It is unclear whether all server-side components safely handle this case.

There is an assumption that the C code handles this correctly (e.g., null-terminating buffers), but this has not been verified.

## Tasks

* [ ] Identify where `\0` is added in agent 5.x
* [ ] Compare behavior with agent 4.x
* [ ] Validate handling in `remoted` (with and without `\0`)
* [ ] Analyze server behavior when receiving 5.x messages **without `\0`**
* [ ] Review all modules and non-engine paths (inventory sync, etc.)
* [ ] Decide approach:

  * Fix only in agent
  * Adjust server handling
  * Or coordinate a joint fix (agent + server)
* [ ] Evaluate whether reverting the server-side workaround is needed
* [ ] Add tests covering both scenarios (`\0` present / absent)

## DoD

* Agent messages do not include unintended `\0`
* Server correctly handles messages **with and without `\0`**
* No decoding issues across all pipelines
* Behavior is consistent and well-defined between agent and server
* No security or parsing inconsistencies exploitable
* Tests added to prevent regressions

jr0me:
## Description

Agent 5.x appends a trailing null byte (`\0`) to every legacy text event it sends to the manager. The byte survives decryption in `remoted`, rides inside the HTTP bulk body the engine receives, and gets stored verbatim by RapidJSON in `event.original`, where it surfaces as `�` / `^@` in archived JSON and silently prevents decoder rules from matching the field, which is why, with no fix in place, no alerts are generated for the affected traffic. Agent 4.x did not have this regression. The byte count used by `send_msg` in 4.x excluded the terminator.

PR #35424 (`Json::setString` trims trailing `\0`) patched the engine to tolerate the bad input. This PR fixes the root cause on the agent, hardens `remoted` so the byte never reaches the engine in the first place, and **reverts the engine workaround from PR #35424** because the layers above it now guarantee the engine never sees a trailing zero. The remoted strip (path-aware after commit `e7e836f4b5` so inventory-sync flatbuffers are passed through byte-exact) and the agent fix are the two defenses; the engine no longer carries a workaround for this class of input.

Closes #35474

> **Update — follow-up regression fix (commit `e7e836f4b5`):** the previous remoted hardening on this branch (commit `6304a002a9`) unconditionally stripped trailing `\0` bytes from the decrypted payload **before** the router-vs-analysisd decision in `HandleSecureMessage`. That broke inventory-sync flatbuffers (queue 115 from FIM, SCA, syscollector) and upgrade-ack JSON: those payloads are not C strings, and the trailing zeros they legitimately carry (vtable offsets / alignment padding for flatbuffers) were treated as terminator-padding and chewed off. `router_provider_send_sync` then rejected every truncated flatbuffer as `Invalid flatbuffer message structure`, and the corrupted bytes fell through to analysisd, which logged `Event not processed` with the raw flatbuffer in `event.original`.
>
> The strip block was moved inside the `if (!forwarded_to_router)` branch so it only applies to analysisd-bound legacy text events. Inventory-sync flatbuffers and upgrade-ack JSON now reach `router_message_forward` byte-exact. The original #35474 symptom is unaffected because legacy text events never carry the `s:` or `u:upgrade_module:` prefix and always fall through to analysisd.
>
> ### Manager log evidence (Docker bench, single agent enrolled, ~30s of activity)
>
> **Pre-fix** (branch tip _before_ `e7e836f4b5`, agent emitting FIM/SCA/syscollector inventory-sync events) — the failure pattern reported by the QA teammate, reproduced verbatim:
>
> ```
> 2026/04/27 15:29:32 :router: ERROR: Error in router_provider_send_sync: Invalid flatbuffer message structure
> 2026/04/27 15:29:32 wazuh-manager-analysisd: WARNING: Event not processed: {"wazuh":{"protocol":{"queue":115,"location":"fim"}, …,"event":{"original":"\f^@^@^@\b^@\f^@^G^@\b^@\b^@^@^@^@^@^@^D,^@^@^@(^@H^@D^@C^@8^@4^@^@^@0^@,^@(^@$^@ ^@^\^@^X^@^T^@^P^@\f^@^@^@\b^@^D^@(^@^@^@…
> 2026/04/27 15:29:32 :router: ERROR: Error in router_provider_send_sync: Invalid flatbuffer message structure
> 2026/04/27 15:29:32 wazuh-manager-analysisd: WARNING: Event not processed: {"wazuh":{"protocol":{"queue":115,"location":"syscollector"}, …
> 2026/04/27 15:29:32 :router: ERROR: Error in router_provider_send_sync: Invalid flatbuffer message structure
> 2026/04/27 15:29:32 wazuh-manager-analysisd: WARNING: Event not processed: {"wazuh":{"protocol":{"queue":115,"location":"sca"}, …
> ```
>
> **Post-fix** (commit `e7e836f4b5`, environment torn down with `docker compose … down -v` to wipe the build volumes, then full cold rebuild of manager + agent from source):
>
> ```
> $ grep -c "router_provider_send_sync.*Invalid flatbuffer" /var/wazuh-manager/logs/wazuh-manager.log
> 0
>
> $ grep "Event not processed" /var/wazuh-manager/logs/wazuh-manager.log | grep '"queue":115' | wc -l
> 0
>
> $ cat /var/ossec/var/run/wazuh-agentd.state   # on the agent container
> status='connected'
> msg_count='112'
> msg_sent='116'
> ```
>
> Inventory-sync module starts and stays clean; FIM / SCA / syscollector flatbuffers reach `router_provider_send_sync` byte-exact and parse successfully. Remaining `Event not processed` warnings in the post-fix log are unrelated pre-existing decoder issues for queue 49 (command outputs) and queue 57 (rootcheck text scans) — they appear identically in both pre- and post-fix logs and are out of scope for this PR.

### Bugs fixed

- **`buffer_append` stored the trailing NUL inside the wire-level size** (`src/client-agent/src/buffer.c`):
  The legacy text-message branch (`msg_len < 0`) set `buffered_message.size = strlen(msg) + 1`, so the size that flowed through `dispatch_buffer → send_msg → CreateSecMSG` was one byte too long for every non-binary event. The fix stores `strlen(msg)` as the reported size while still allocating `strlen + 1` for the in-memory copy, keeping the stored buffer safely null-terminated. Direct `send_msg(msg, -1)` callers (`notify.c`, `request.c`, `start_agent.c`, `reload_agent.c`, `receiver.c`, and the four buffer-state messages inside `buffer.c`) were already correct because `sendmsg.c` uses `strlen(msg)` — no `+1` — in that branch. The Windows `SendMSGAction` path feeds the same `buffer_append` sink and is fixed transitively by the same change.

- **`remoted` left control-message copies and event-raw buffers without a guaranteed terminator** (`src/remoted/src/secure.c`):
  - `ctrl_msg_data->message` was `os_calloc(msg_length, ...)` but memcpy'd only `tmp_msg_length` bytes (= `msg_length - 3` after the `#!-` header strip). The three extra bytes from calloc happened to zero the area past the payload, but the size stored was inconsistent and a future alloc tightening would have exposed a buffer over-read to downstream `cJSON_Parse(strchr(clean, '{'))` in `validate_control_msg`. Changed to `os_calloc(tmp_msg_length + 1, ...)` so the allocation matches the memcpy and the `+1` byte is a guaranteed sentinel from calloc.
  - `e->raw` was `os_calloc(msg_length, ...)`. The downstream `bulk_append` is length-aware, so the on-wire HTTP body sent to the engine is unchanged, but any future consumer treating `e->raw` as a C string would over-read the allocation by one byte. Changed to `os_calloc(msg_length + 1, ...)`.
  - The live cleartext buffer (`tmp_msg`) was not modified: the existing `memset(cleartext_msg, '\0', OS_MAXSTR + 1)` at the top of `HandleSecureMessage` already guarantees that the byte past any valid decrypted payload is a NUL, so `strncmp` / `IsValidHeader` / `w_utf8_filter` / `cJSON_ParseWithOpts` cannot walk off the decrypted region. Added a documentation block noting the invariant so future edits do not drop the memset or introduce a path that re-enters `ReadSecMSG` without re-zeroing.

### Root cause

`src/client-agent/src/buffer.c:150` — the legacy-string branch used `size_to_alloc = strlen(msg) + 1` and stored that value in `buffered_message.size`. When `dispatch_buffer` later called `send_msg(msg_to_dispatch.data, msg_to_dispatch.size)`, the size (including the NUL) reached `sendmsg.c:32`, which passed it straight to `CreateSecMSG`, which encrypted exactly that many bytes. Commit `261ad5f865` ("Added new function SendBinaryMSG…") introduced the `{void *data; size_t size}` struct and this `+1` as a side effect of giving the buffer an explicit-length API for binary events; 4.x had no such struct and relied on `sendmsg.c`'s own `strlen(msg)` (no `+1`) to compute the wire-level length.

### Results and Evidence

Verified end-to-end using wazuh-server (indexer + dashboard + manager) and wazuh-agent both built from source. The harness was completely torn down between the two phases, so the pre- and post-fix runs share no state.

**Pre-fix configuration** (build with `src/engine/source/base/src/json.cpp` PR #35424 trim reverted, `src/client-agent/src/buffer.c` fix reverted, and `src/remoted/src/secure.c` allocation reverted:

```
$ curl … 'https://127.0.0.1:9200/_cat/indices/wazuh-*?v' | awk '$7>0'
…
.ds-wazuh-events-v5-system-activity-000001       14
wazuh-states-fim-files                         1322
(all 8 wazuh-findings-v5-* indices: 0 docs)
```

Sampled `event.original` (5 newest, all from the system-activity stream produced by `wazuh-logcollector` reading journald):

```
2026-04-23T14:26:40.429Z  repr('… modprobe@drm.service: Executable /sbin/modprobe missing, skipping: No such file or directory\x00')
2026-04-23T14:26:40.429Z  repr("… dbus-daemon[4805]: [system] Successfully activated service 'org.freedesktop.login1'\x00")
2026-04-23T14:26:40.429Z  repr('… systemd[1]: Started User Login Management.\x00')
2026-04-23T14:26:40.428Z  repr('… systemd[1]: Starting User Login Management...\x00')
2026-04-23T14:26:40.428Z  repr('… systemd-logind[5659]: New seat seat0.\x00')
```

- **10/10 sampled events end with a literal `\x00` byte** (rendered as `�` in JSON-over-the-wire).
- **0 findings across all 24 `wazuh-findings-v5-*` datastreams** — the engine's decoders cannot match `event.original` when the field carries the trailing NUL, so no rule fires. This is the "doesn't generate events, doesn't match any decoder" symptom reported on the issue.

**Post-fix configuration** (branch tip as of this PR — agent fix applied, remoted harden applied, engine trim from PR #35424 reverted because the agent + remoted layers now guarantee the engine never sees a trailing zero):

```
$ curl … 'https://127.0.0.1:9200/wazuh-events-v5-*/_count'     → 9,885+
$ curl … 'https://127.0.0.1:9200/wazuh-findings-v5-*/_count'   → 9,505+
$ curl … 'https://127.0.0.1:9200/wazuh-states-sca/_count'      → 207
$ curl … 'https://127.0.0.1:9200/wazuh-states-inventory-packages/_count' → 301
(similar non-zero counts across wazuh-states-inventory-services/processes/users/networks/hardware/etc.)
```

- **Sampled 2,000 events from `wazuh-events-v5-*`: 0 contain `\x00`.**
- **Sampled 2,000 findings from `wazuh-findings-v5-*`: 0 contain `\x00`.**

Representative per-module samples (all `event.original` verified NUL-free):

| Module (source) | event.original (post-fix) | rule.id | rule.level |
|---|---|---|---|
| `logcollector` / journald `sshd` | `Apr 23 14:52:26 wazuh-agent sshd[11863]: Failed password for invalid user probe_postfix1776955945 from 10.0.0.1 port 22 ssh2` | `8088b365-ddf6-4783-9835-48c4c89ed60a` | `low` |
| `rootcheck` / scan-summary text | `Files hidden inside directory '/etc/dpkg'. Link count does not match number of files (4,1).` | `4d5c1922-a12c-4f79-bd57-63bbfc699ab5` | `high` |
| `logcollector` / `/var/log/dpkg.log` | (events flow through the same `buffer_append` sink; NUL scan over 2000 events is clean) | — | — |
| `SCA` / flatbuffer JSON `"module":"sca"` | (inventory-sync flatbuffer path; `wazuh-states-sca` has 207 docs post-fix) | — | — |
| `syscollector` / inventory flatbuffer | `wazuh-states-inventory-packages` 301, -services 32, -processes 7, -users 24, -networks 3, -hardware 1, -system 1 — all populated and NUL-free | — | — |
| `FIM` / inventory-sync flatbuffer | `wazuh-states-fim-files` carries the initial 1,322-file scan with clean file paths and hashes (incremental sync was separately degraded in this bench; see Notes) | — | — |
| Keepalive / `#!-` control messages | exercised on every agent restart; processed by `remoted::validate_control_msg` — with the harden commit, `ctrl_msg_data->message` is now exactly `tmp_msg_length + 1` (was over-allocated by 3 bytes with no guaranteed sentinel) | n/a (control) | n/a |

**Before/after decoder-match count (24 `wazuh-findings-v5-*` datastreams):**

| | Pre-fix | Post-fix |
|---|---|---|
| Findings (= rule matches in engine) | **0** | **9,505+** and growing |

The eight order-of-magnitude jump reflects decoders & rules matching for the first time once `event.original` no longer carries the trailing `\0`.

## DoDs

| DoD req | Evidence |
|---|---|
| Agent messages do not include unintended `\0` | 2,000 post-fix `wazuh-events-v5-*` docs sampled; 0 contain `\x00`. Pre-fix sample: 10/10 contain `\x00`. |
| Server correctly handles messages **with** `\0` | `remoted::HandleSecureMessage` strips trailing zeros from analysisd-bound text events (defense-in-depth for unpatched 5.x agents in the field). The strip is now path-aware so inventory-sync flatbuffers and upgrade-ack JSON reach `router_provider_send_sync` byte-exact -- covered by `test_HandleSecureMessage_event_with_trailing_null` and `test_HandleSecureMessage_inventory_sync_preserves_trailing_null`. The engine `Json::setString` trim from PR #35424 has been reverted on this branch (commit `e0c2687fbb`); it is no longer needed because the trailing zero never reaches the engine. |
| Server correctly handles messages **without** `\0` | `remoted` allocations right-sized (`e->raw[e->len]`, `ctrl_msg_data->message[tmp_msg_length]` both guaranteed `\0`). `test_HandleSecureMessage_event_without_trailing_null` exercises the inverse case under ASan, clean. |
| No decoding issues across all pipelines | Post-fix: 9,505+ findings across the security, system-activity, and other category streams; rule matches observed for sshd (logcollector/journald, rule `8088b365`) and rootcheck (rule `4d5c1922`); SCA/syscollector/FIM state populated (207 / 300+ / 1,322 docs). 0/2,000 findings contain `\x00`. |
| Behavior is consistent and well-defined between agent and server | Single coordinated PR: agent fix at the source (`buffer_append`), remoted hardens (right-sized allocations + path-aware strip on the analysisd branch), engine `setString` trim from PR #35424 reverted because it is no longer load-bearing. |
| No security or parsing inconsistencies exploitable | Inverse-case covered by the harden commit at three sites (`e->raw`, `ctrl_msg_data->message` right-sized; `tmp_msg` invariant documented) and the ASan unit test. |
| Tests added to prevent regressions | Three new unit tests (2 in `test_buffer.c`, 1 in `test_secure.c`) + sentinel assertion on the existing shutdown test. |

### Notes on the test bench

- The engine `Json::setString` trim added by PR #35424 has been **reverted** on this branch (commit `e0c2687fbb refactor(engine): remove Json::setString trim of trailing null bytes`). With the agent fix at the source and the path-aware remoted strip as defense-in-depth, the engine never sees a trailing zero in `event.original`, and the workaround is no longer load-bearing. Keeping it would mean carrying a behavioural shim in the engine for an input class that the layers above it guarantee never to produce.

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Log syntax and correct language review (no new user-facing logs added)

- Memory tests for Linux
  - [ ] Coverity
  - [x] AddressSanitizer — `server.cmake` bakes `-fsanitize=address,undefined` into every unit test. The new `test_HandleSecureMessage_event_without_trailing_null` feeds an inverse-case (no trailing NUL) cleartext through `HandleSecureMessage` and reads `e->raw[e->len]` while the allocation is live; pre-harden sizing (`os_calloc(msg_length, ...)`) fires ASan heap-buffer-overflow on that read, post-harden the read is in-bounds and returns `\0`. Verified green inside wazuh-server container: `49/49 tests passed, ASan clean`.
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - N/A — no decoder or rule changes.

  - [x] `JsonSettersTest/SetStringStripsTrailingNullBytes` (added by PR #35424) has been removed alongside the `Json::setString` trim in commit `e0c2687fbb`. The agent + remoted layers now guarantee the engine never sees a trailing zero, so the trim and its test are no longer load-bearing.
  - [ ] Test run in parallel

- Wazuh server API/Framework
  - N/A — no API or framework changes.

### Artifacts Affected

- `wazuh-agent` on Linux (the shared `buffer_append` sink fixes the legacy text path for every `send_msg(msg, -1)` caller that goes through the buffer).
- `wazuh-agent` on Windows (transitively — `win_utils.c:SendMSGAction` calls the same `buffer_append` sink; no separate Windows change).
- `wazuh-manager` (Linux) — `remoted` allocations and documentation in `src/remoted/src/secure.c`.

### Configuration Changes

N/A — no user-visible configuration changes.

### Tests Introduced

- `test_buffer_append_binary_preserves_size` (`src/unit_tests/client-agent/test_buffer.c`): positive regression guard for the explicit-length path. Asserts that `buffer_append(payload, N)` with a payload containing embedded NULs stores `buffer[0].size == N` and preserves the exact bytes.
- `test_buffer_append_text_size_excludes_null_terminator` (same file): the demonstrator for issue #35474. Asserts that `buffer_append(text, -1)` stores `buffer[0].size == strlen(text)`. Would fail on main pre-fix.
- `test_HandleSecureMessage_event_without_trailing_null` (`src/unit_tests/remoted/test_secure.c`): inverse-case ASan guard. Feeds a `msg_length`-exact cleartext payload with no trailing NUL through `HandleSecureMessage`; the custom `expect_check` hook reads `e->raw[e->len]` while the allocation is live and asserts it equals `\0`. Under the ASan build configured by `server.cmake`, any regression that shrinks the event allocation below `e->len + 1` fires a heap-buffer-overflow on that read.
- `test_HandleSecureMessage_shutdown_message` (existing test, same file): extended with `assert_int_equal('\0', node->message[tmp_msg_length])` as a sentinel guard for the control-message allocation.
- `test_HandleSecureMessage_inventory_sync_preserves_trailing_null` (`src/unit_tests/remoted/test_secure.c`, added in commit `e7e836f4b5`): regression guard for the router-bound path. Crafts an `s:fim:<flatbuffer>` payload whose flatbuffer ends in three `\0` bytes, hooks `router_provider_send_sync` via a new wrap (`router_wrappers.{c,h}`), and asserts the wrap receives the full flatbuffer with the trailing zeros intact, with the correct `message_size` and `authenticated_agent_id`. Would fail before commit `e7e836f4b5` because the pre-router strip would shave the trailing zeros off the flatbuffer.

## Review Checklist

- [ ] Code changes reviewed
- [x] Relevant evidence provided (pre/post-fix indexer samples above)
- [x] Tests cover the new functionality (agent regression test + remoted ASan inverse-case + sentinel assertions)
- [x] Configuration changes documented (N/A, noted explicitly)

### DIFF
--- a/src/client-agent/include/agentd.h
+++ b/src/client-agent/include/agentd.h
@@ -58,19 +58,9 @@ int receive_msg(void);
 int receiver_messages(void);
 #endif
 
-/**
- * @brief Represents a single message stored in the event buffer.
- *
- * This structure is used to hold a message that can be either a null-terminated
- * text string or a raw binary data buffer. It pairs a pointer to the data with
- * its explicit size, allowing the system to handle binary content safely
- * without truncation at null bytes.
- */
+/* Message stored in the event buffer. */
 typedef struct {
-    /** @brief Pointer to the dynamically allocated message data. */
     void *data;
-
-    /** @brief The exact size of the data in bytes. */
     size_t size;
 } buffered_message;
 
--- a/src/client-agent/include/agentd.h
+++ b/src/client-agent/include/agentd.h
@@ -58,6 +58,22 @@ int receive_msg(void);
 int receiver_messages(void);
 #endif
 
+/**
+ * @brief Represents a single message stored in the event buffer.
+ *
+ * This structure is used to hold a message that can be either a null-terminated
+ * text string or a raw binary data buffer. It pairs a pointer to the data with
+ * its explicit size, allowing the system to handle binary content safely
+ * without truncation at null bytes.
+ */
+typedef struct {
+    /** @brief Pointer to the dynamically allocated message data. */
+    void *data;
+
+    /** @brief The exact size of the data in bytes. */
+    size_t size;
+} buffered_message;
+
 /* Initialize agent buffer */
 void buffer_init();
 
--- a/src/client-agent/src/buffer.c
+++ b/src/client-agent/src/buffer.c
@@ -133,14 +133,10 @@ int buffer_append(const char *msg, ssize_t msg_len) {
     } else {
         size_t payload_size;
         size_t alloc_size;
-        if (msg_len < 0) { // Null-terminated text message.
-            // The wire-level size must exclude the terminator so the manager
-            // does not receive a trailing '\0' in event.original. Allocate
-            // payload_size + 1 so the stored copy stays safely null-terminated
-            // for any callers that briefly treat it as a C string.
+        if (msg_len < 0) {
             payload_size = strlen(msg);
             alloc_size = payload_size + 1;
-        } else { // Binary buffer with an explicit length.
+        } else {
             payload_size = (size_t)msg_len;
             alloc_size = payload_size;
         }
--- a/src/client-agent/src/buffer.c
+++ b/src/client-agent/src/buffer.c
@@ -43,25 +43,9 @@ struct{
 } buff;
 
 /**
- * @brief Represents a single message stored in the event buffer.
- *
- * This structure is used to hold a message that can be either a null-terminated
- * text string or a raw binary data buffer. It pairs a pointer to the data with
- * its explicit size, allowing the system to handle binary content safely
- * without truncation at null bytes.
- */
-typedef struct {
-    /** @brief Pointer to the dynamically allocated message data. */
-    void *data;
-
-    /** @brief The exact size of the data in bytes. */
-    size_t size;
-} buffered_message;
-
-/**
  * @brief The agent's main event buffer.
  */
-static buffered_message *buffer;
+STATIC buffered_message *buffer;
 
 static pthread_mutex_t mutex_lock;
 static pthread_cond_t cond_no_empty;
--- a/src/client-agent/src/buffer.c
+++ b/src/client-agent/src/buffer.c
@@ -131,16 +131,23 @@ int buffer_append(const char *msg, ssize_t msg_len) {
         return(-1);
 
     } else {
-        size_t size_to_alloc;
-        if (msg_len < 0) { // It's a null-terminated string.
-            size_to_alloc = strlen(msg) + 1;
-        } else { // It's a binary buffer with a known length.
-            size_to_alloc = (size_t)msg_len;
+        size_t payload_size;
+        size_t alloc_size;
+        if (msg_len < 0) { // Null-terminated text message.
+            // The wire-level size must exclude the terminator so the manager
+            // does not receive a trailing '\0' in event.original. Allocate
+            // payload_size + 1 so the stored copy stays safely null-terminated
+            // for any callers that briefly treat it as a C string.
+            payload_size = strlen(msg);
+            alloc_size = payload_size + 1;
+        } else { // Binary buffer with an explicit length.
+            payload_size = (size_t)msg_len;
+            alloc_size = payload_size;
         }
 
-        os_malloc(size_to_alloc, buffer[i].data);
-        memcpy(buffer[i].data, msg, size_to_alloc);
-        buffer[i].size = size_to_alloc;
+        os_malloc(alloc_size, buffer[i].data);
+        memcpy(buffer[i].data, msg, alloc_size);
+        buffer[i].size = payload_size;
 
         forward(i, agt->buflength + 1);
         w_cond_signal(&cond_no_empty);
--- a/src/engine/source/base/src/json.cpp
+++ b/src/engine/source/base/src/json.cpp
@@ -987,13 +987,7 @@ void Json::setString(std::string_view value, std::string_view path)
     if (pp.IsValid())
     {
         const auto* data = value.data() ? value.data() : "";
-        auto len = value.size();
-        // Strip trailing null bytes: a JSON string must not contain embedded \0.
-        while (len > 0 && data[len - 1] == '\0')
-        {
-            --len;
-        }
-        rapidjson::Value v(data, static_cast<rapidjson::SizeType>(len), m_document.GetAllocator());
+        rapidjson::Value v(data, static_cast<rapidjson::SizeType>(value.size()), m_document.GetAllocator());
         pp.Set(m_document, v);
         return;
     }
--- a/src/engine/source/base/test/src/unit/json_test.cpp
+++ b/src/engine/source/base/test/src/unit/json_test.cpp
@@ -1702,64 +1702,6 @@ TEST_F(JsonSettersTest, SetString)
     ASSERT_THROW(jObjString.setString("newValue", "object/key"), std::runtime_error);
 }
 
-TEST_F(JsonSettersTest, SetStringStripsTrailingNullBytes)
-{
-    Json doc {};
-    std::string tmp;
-
-    // Single trailing \0 (the most common case: OS_SendUnix sends strlen+1 bytes)
-    {
-        std::string_view withNull {"hello\0", 6}; // size=6, includes the \0
-        doc.setString(withNull, "/field");
-        ASSERT_EQ(json::RetGet::Success, doc.getString(tmp, "/field"));
-        ASSERT_EQ("hello", tmp);
-        ASSERT_EQ(5u, tmp.size()); // no null byte stored
-        // Serialised JSON must not contain \u0000
-        ASSERT_EQ(std::string::npos, doc.str().find("\\u0000"));
-    }
-
-    // Multiple trailing \0 bytes (e.g. padded buffer)
-    {
-        std::string_view multiNull {"abc\0\0\0", 6};
-        doc.setString(multiNull, "/field2");
-        ASSERT_EQ(json::RetGet::Success, doc.getString(tmp, "/field2"));
-        ASSERT_EQ("abc", tmp);
-        ASSERT_EQ(3u, tmp.size());
-    }
-
-    // String that is all null bytes must become an empty string
-    {
-        std::string_view allNull {"\0\0", 2};
-        doc.setString(allNull, "/field3");
-        ASSERT_EQ(json::RetGet::Success, doc.getString(tmp, "/field3"));
-        ASSERT_TRUE(tmp.empty());
-    }
-
-    // Normal string without any \0 must be stored unchanged
-    {
-        doc.setString("normal", "/field4");
-        ASSERT_EQ(json::RetGet::Success, doc.getString(tmp, "/field4"));
-        ASSERT_EQ("normal", tmp);
-    }
-
-    // Empty string must remain empty
-    {
-        doc.setString("", "/field5");
-        ASSERT_EQ(json::RetGet::Success, doc.getString(tmp, "/field5"));
-        ASSERT_TRUE(tmp.empty());
-    }
-
-    // \0 embedded in the MIDDLE is not stripped (only trailing ones are)
-    {
-        std::string_view embeddedNull {"a\0b", 3}; // \0 is in the middle
-        doc.setString(embeddedNull, "/field6");
-        ASSERT_EQ(json::RetGet::Success, doc.getString(tmp, "/field6"));
-        // Only the trailing-strip loop runs; the mid-string \0 followed by 'b' means
-        // no trailing \0 is stripped — the full 3-byte string is stored.
-        ASSERT_EQ(3u, tmp.size());
-    }
-}
-
 TEST_F(JsonSettersTest, SetInt)
 {
     Json jObjInt {R"({
--- a/src/remoted/src/secure.c
+++ b/src/remoted/src/secure.c
@@ -1004,6 +1004,23 @@ STATIC void HandleSecureMessage(const message_t *message, w_indexed_queue_t * co
         return;
     }
 
+    /* Issue #35474: strip trailing NUL bytes from the event payload before
+     * it reaches the router or analysisd. Wazuh 5.x agents prior to the
+     * buffer_append fix included the terminating '\0' in the wire-level
+     * length of legacy text events; the byte rides the HTTP body and gets
+     * stored verbatim by RapidJSON in event.original, silently breaking
+     * decoder matches. Compliant agents are unaffected -- the loop is a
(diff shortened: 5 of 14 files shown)


TomasTurina (dismissed):

jpcerrone (dismissed):
The description mentions `...keeps the engine workaround in place as defense-in-depth` which I understand is no longer the case? Other than that LGTM

TomasTurina (approved):

Nicogp (approved):

PR #35953, opened 2026-05-06
Graceful termination via cooperative cancellation

### LINKED ISSUE #34479: Vulnerability scanner - wmodules ungraceful termination
# Description 

It was found in v4.14.1 that on occasions the vulnerability scanner may not successfully terminate, this may be related to received events, during the scan or indexations. Discovering where exactly it "hangs",  is part of this issue. 

Image

## DoD 

- [ ] Root cause analysis 
- [ ] Evaluate if that affects the result of the scanner 
- [ ] Propose a solution if required 
- [ ] Manual testing

jotacarma90:
## Summary

Fixes #34479

`wazuh-modulesd` did not shut down gracefully on SIGTERM: module threads blocked in `sleep()`, `select()`, `sem_wait()`, or `pthread_cond_wait()` with no timeout and no stop mechanism, leaving the process to be killed forcefully by the service manager.

## Tested Scenarios
| Scenario | What was tested | Result |
| --- | --- | --- |
| Manager restart during active feed download | Manager shutdown while `content-updater` was downloading the vulnerability feed. | 🟢 Download slices were interrupted cleanly; no 30-second wait. |
| Manager restart while agent sends events continuously | The agent generated journald events in a loop while the manager was restarted. | 🟢 Restart: `8 s`; manager stopped cleanly and came back with pending events preserved. |
| Agent stop while logcollector processes active writes | A loop appended events to a monitored log file while the agent was stopped. | 🟢 Stop: `1 s`; logcollector exited cleanly and the agent restarted normally. |
| Simultaneous manager and agent restart | Manager and agent control scripts were restarted at the same time. | 🟢 Combined restart: `9 s`; both sides returned to running state and the agent reconnected. |
| Manager restart immediately after startup | The manager was restarted again one second after a previous restart completed. | 🟢 Restarts: `6 s` each; modules handled shutdown during fresh startup state. |
| Agent stop just after startup module activity | The agent was stopped immediately after rootcheck, FIM, SCA, and syscollector startup activity. | 🟢 Stop: `2 s`; modules released resources normally. |
| Timeout/forced-kill log search | Searched manager and agent logs for timeout/kill indicators after the stress scenarios. | 🟢 No forced-kill fallback observed. Only internal `agent-upgrade` timeout warnings, with no shutdown delay. |

- Full commands output and ossec.log: 
[graceful-termination-ubuntu24-stress.txt](https://github.com/user-attachments/files/27521277/graceful-termination-ubuntu24-stress.raw.txt)

Conclusion: the tested shutdown paths do not approach the 30-second forced-kill fallback. Manager shutdown, agent shutdown, and modulesd-related shutdown activity completed in a few seconds across the tested scenarios.

### DIFF
--- a/src/wazuh_modules/src/agent_upgrade/agent/wm_agent_upgrade_agent.c
+++ b/src/wazuh_modules/src/agent_upgrade/agent/wm_agent_upgrade_agent.c
@@ -27,25 +27,18 @@
 #endif
 
 const char* upgrade_values[] = {
-    [WM_UPGRADE_SUCCESSFUL] = "0",
-    [WM_UPGRADE_FAILED_INTERMEDIATE] = "1",
-    [WM_UPGRADE_FAILED] = "2"
-};
+    [WM_UPGRADE_SUCCESSFUL] = "0", [WM_UPGRADE_FAILED_INTERMEDIATE] = "1", [WM_UPGRADE_FAILED] = "2"};
 
-const char* upgrade_messages[] = {
-    [WM_UPGRADE_SUCCESSFUL] = "Upgrade was successful",
-    [WM_UPGRADE_FAILED_INTERMEDIATE] = "Upgrade failed: intermediate version required",
-    [WM_UPGRADE_FAILED] = "Upgrade failed"
-};
+const char* upgrade_messages[] = {[WM_UPGRADE_SUCCESSFUL] = "Upgrade was successful",
+                                  [WM_UPGRADE_FAILED_INTERMEDIATE] = "Upgrade failed: intermediate version required",
+                                  [WM_UPGRADE_FAILED] = "Upgrade failed"};
 
-static const char *task_statuses_map[] = {
-    [WM_UPGRADE_SUCCESSFUL] = WM_TASK_STATUS_DONE,
-    [WM_UPGRADE_FAILED_INTERMEDIATE] = WM_TASK_STATUS_FAILED,
-    [WM_UPGRADE_FAILED] = WM_TASK_STATUS_FAILED
-};
+static const char* task_statuses_map[] = {[WM_UPGRADE_SUCCESSFUL] = WM_TASK_STATUS_DONE,
+                                          [WM_UPGRADE_FAILED_INTERMEDIATE] = WM_TASK_STATUS_FAILED,
+                                          [WM_UPGRADE_FAILED] = WM_TASK_STATUS_FAILED};
 
 // CA certificates
-char **wcom_ca_store = NULL;
+char** wcom_ca_store = NULL;
 
 #ifndef WIN32
 
@@ -53,7 +46,7 @@ char **wcom_ca_store = NULL;
  * Listen to the upgrade socket in order to receive commands
  * @return only on errors, socket will be closed
  * */
-STATIC void* wm_agent_upgrade_listen_messages(__attribute__((unused)) void *arg);
+STATIC void* wm_agent_upgrade_listen_messages(__attribute__((unused)) void* arg);
 
 #endif
 
@@ -73,7 +66,7 @@ STATIC void wm_agent_upgrade_check_status(const wm_agent_configs* agent_config) 
  * @retval true information was found on the upgrade_result file
  * @retval either the upgrade_result file does not exist or contains invalid information
  * */
-STATIC bool wm_upgrade_agent_search_upgrade_result(int *queue_fd);
+STATIC bool wm_upgrade_agent_search_upgrade_result(int* queue_fd);
 
 /**
  * Reads the upgrade_result file if it is present and sends the upgrade result message to the manager.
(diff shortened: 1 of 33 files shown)


src/wazuh_modules/src/wm_control.c
```diff
@@ -201,22 +206,20 @@ static void *process_control() {
...
-
+            mterror_exit(WM_CONTROL_LOGTAG, "At process_control(): select(): %s", strerror(errno));
+            break;
         case 0:
             continue;
+        default:
+            break;
         }
 
+        if (wm_shutdown_requested) break;
```
Antoniogm03: Is this necessary? If this is true, the loop would not be entered

src/wazuh_modules/src/wm_ms_graph.c
```diff
@@ -228,7 +229,7 @@ static curl_response* wm_ms_graph_http_get_with_retry(char** headers, const char
                 relationship_name, retry_after, attempt + 1, WM_MS_GRAPH_MAX_RETRIES);
             wurl_free_response(response);
             response = NULL;
-            w_sleep_until(time(NULL) + retry_after);
+            wm_sleep_until_interruptible(time(NULL) + retry_after);
```
Antoniogm03, line 232: Shouldn't we check whether `wm_shutdown_requested` has been called before the continue?

src/wazuh_modules/src/wmcom.c
```diff
@@ -137,24 +137,21 @@ void * wmcom_main(__attribute__((unused)) void * arg) {
...
-
+            merror_exit("At wmcom_main(): select(): %s", strerror(errno));
+            break;
         case 0:
             continue;
+        default:
+            break;
         }
 
+        if (wm_shutdown_requested) break;
```
Antoniogm03: I think that's not necessary

src/wazuh_modules/src/task_manager/wm_task_manager.c
```diff
@@ -139,24 +140,21 @@ STATIC void* wm_task_manager_main(wm_task_manager* task_config) {
...
+            mterror(WM_TASK_MANAGER_LOGTAG, MOD_TASK_SELECT_ERROR, strerror(errno));
+            pthread_exit(NULL);
+            break;
         case 0:
             continue;
         default:
             break;
         }
 
+        if (wm_shutdown_requested) break;
```
Antoniogm03: I think that's not necessary

src/wazuh_modules/src/agent_upgrade/manager/wm_agent_upgrade_manager.c
```diff
@@ -129,24 +133,23 @@ STATIC void wm_agent_upgrade_listen_messages(const wm_manager_configs* manager_c
...
+            mterror(WM_AGENT_UPGRADE_LOGTAG, WM_UPGRADE_SELECT_ERROR, strerror(errno));
+            close(sock);
+            return;
         case 0:
             continue;
+        default:
+            break;
         }
 
+        if (wm_shutdown_requested) break;
```
Antoniogm03: I think that's not necessary

Antoniogm03 (changes requested):

src/wazuh_modules/src/wm_control.c
jotacarma90: Removed

src/wazuh_modules/src/wm_ms_graph.c
jotacarma90: Added

src/wazuh_modules/src/wmcom.c
jotacarma90: Removed

src/wazuh_modules/src/task_manager/wm_task_manager.c
jotacarma90: Removed

src/wazuh_modules/src/agent_upgrade/manager/wm_agent_upgrade_manager.c
jotacarma90: Removed

MiguelazoDS (commented):
I was unable to reproduce the issue during the feed download as I did before in addition, on 4.x during the population of events and indexer redundant database, it was possible to provoke the unwanted behavior by restarting the module in the middle of that process, but since 5.x does not make use of those queues anymore, the risk is reduced. Although I'm not sure if it is completely removed, could you reproduce any situation where the daemon hangs before the fix?

jotacarma90:
> I was unable to reproduce the issue during the feed download as I did before in addition, on 4.x during the population of events and indexer redundant database, it was possible to provoke the unwanted behavior by restarting the module in the middle of that process, but since 5.x does not make use of those queues anymore, the risk is reduced. Although I'm not sure if it is completely removed, could you reproduce any situation where the daemon hangs before the fix?

I would frame this change as preventive hardening rather than a fix for an easily reproducible 5.x failure. The risky path still exists logically: after the feed update, the content-updater callback may enter `runScanAfterFeedUpdate()` and iterate agents while doing per-agent scan work. If shutdown arrives in that window, `stop()` tears down the feed manager and has to wait for that callback to return. In a small/local environment this is hard to hit and finishes quickly, but with many agents or slow indexer operations the shutdown path can still be unnecessarily exposed to a long-running callback.

The patch makes that path explicitly shutdown-aware by checking the stop flag before entering the post-feed scan and inside the per-agent loop. That way, if shutdown is requested, the callback exits early and teardown can complete promptly.

In short: I could not reproduce the old 4.x-style hang on current `main`, but the change is still useful as a low-risk preemptive protection for a narrow shutdown window that would otherwise depend on workload size and indexer latency.

src/wazuh_modules/src/agent_upgrade/agent/wm_agent_upgrade_agent.c
```diff
@@ -89,99 +82,105 @@ STATIC bool wm_upgrade_agent_search_upgrade_result(int *queue_fd);
...
+        switch (wm_select_interruptible(sock, &fdset))
+        {
+            case -1:
+                if (errno != EINTR)
+                {
+                    mterror(WM_AGENT_UPGRADE_LOGTAG, WM_UPGRADE_SELECT_ERROR, strerror(errno));
+                    close(sock);
+                    return NULL;
+                }
+                continue;
```
Antoniogm03: `wm_select_interruptible` returns -1 if `errno != EINTR` so the conditional is not necessary.

Antoniogm03 (changes requested):

Antoniogm03 (approved):
LGTM!

PR #36048, opened 2026-05-12
Add workflow to upgrade external dependencies

### LINKED ISSUE #36063: Simplify the external dependencies upgrade process
|---|---|---|---|---|
| 5.0.0 | externals | Agent / Manager | Sources | All |

### Description

Upgrading one of Wazuh's precompiled external dependencies (the tarballs published under `packages.wazuh.com/deps/<version>/libraries/`) is currently fully manual: someone runs `make deps` / `make` locally on the right host for each `(system, architecture, target)` combination, packs the result by hand, and bumps `DEPS_VERSION` in `src/Makefile`. There is no reproducible CI surface for this — it can't be triggered from `workflow_dispatch`, leaves no artifact trail, and silently depends on whatever host happened to run it.

### Goal

Make this process simpler and easier to run, without changing the overall flow:

- Provide a `workflow_dispatch` workflow that rebuilds the externals across every supported `(system, architecture, target)` leg.
- Emit per-leg artifacts already laid out in the S3 path structure that `make deps` consumes (`libraries/<os>/<arch>/<dep>.tar.gz` + `libraries/sources/<dep>.tar.gz`), so the output of a run can be uploaded to `packages.wazuh.com/deps/<version>/` as-is.
- Accept per-dependency `name:version` overrides as an input, so a single dispatch can produce the binaries for a planned upgrade without editing source files first.

### Out of scope

- Testing the rebuilt binaries — still done manually after a run.
- Uploading the artifacts to `packages.wazuh.com/deps/<version>/` — still done manually.
- Bumping `DEPS_VERSION` in `src/Makefile` — done in a separate, follow-up PR once the artifacts are uploaded.

No dependency version is bumped here; this is purely the workflow and its supporting build scripts.

jr0me:
## Related Issue

Closes #36063

## Description

Wazuh's precompiled externals are currently rebuilt out-of-band: someone runs `make deps` / `make` locally on the right host, uploads the resulting tarballs to `packages.wazuh.com/deps/<version>/...`, and bumps `DEPS_VERSION` in `src/Makefile`. There is no reproducible CI surface for this — it can't be triggered from `workflow_dispatch`, leaves no artifact trail, and silently depends on whatever host happened to run it.

This PR adds a dedicated workflow (`5.X - Package - Build external dependencies`) that rebuilds Wazuh's externals across every supported (system, architecture) leg, merges every per-leg output into a single `externals-all.tar.gz` laid out in the S3 path structure `make deps` consumes from `packages.wazuh.com/deps/<version>/libraries/`, and then runs a smoke build that compiles the agent + manager from source against that consolidated tree to confirm the precompiled tarballs are actually consumable. Uploading the workflow artifact's contents as-is produces a complete `libraries/` tree under a new deps version.

No dependency version is being bumped in this PR — it is purely the workflow, its supporting scripts, and the dev doc that describes how to use it. The first deps bump that consumes this workflow lives in #36152 (curl, sqlite, xz, libarchive → `DEPS_VERSION=99-29734`).

### New workflow `.github/workflows/5_builderpackage_externals.yml`

`workflow_dispatch` with two inputs:

- `dependencies` — semicolon-separated `name:version` overrides (e.g. `curl:8.20.0;sqlite:3.53.1`). Empty = rebuild every dep from the currently vendored sources without replacing any.
- `docker_image_tag` — `auto` (derive from `VERSION.json`), `developer` (branch name), or a literal GHCR tag for the Linux builder images.

There is intentionally no per-leg input. A deps release is whole-or-nothing — a partial run would publish a tarball that breaks `make deps` on any platform whose leg is missing. To re-run a single broken leg, use GitHub's *Re-run failed jobs* on the workflow run.

Three jobs:

- **`build-externals`** — fixed 7-entry matrix (no `prepare-matrix` job). Five legs run agent-target (`rpm-amd64`, `rpm-arm64`, `macos-intel64`, `macos-arm64`, `windows-i686`) and two legs run manager-target (`rpm-amd64`, `rpm-arm64`). Linux legs run on `wz-linux-amd64` / `wz-linux-arm64`; macOS legs run natively on `macos-14-large` / `macos-14`; the Windows leg installs `gcc-mingw-w64` on a Linux runner and cross-compiles, mirroring `5_builderpackage_agent-windows.yml`. `deb-*` legs aren't built — rpm carries the older glibc baseline and is forward-compatible with deb. Each leg uploads `externals-<leg>-<target>.tar.gz`.
- **`consolidate`** — downloads every per-leg tarball and merges them into the canonical `libraries/{linux,darwin,windows,sources}/` layout, then uploads a single `externals-all.tar.gz`. Manager legs are processed first, then agent legs, so for Linux deps where both produced a binary, the agent (older glibc, more portable) copy wins. Source zips are byte-identical across legs; first-writer-wins is fine.
- **`smoke-build`** — for each of the 4 Linux combinations (amd64/arm64 × agent/manager), downloads `externals-all.tar.gz`, points `make deps RESOURCES_URL=file://…` at the local tree, and runs the actual Wazuh build from source. Catches packing-path bugs: a dep packed at the wrong path silently triggers source recompile, and the "Analyze dependency usage" step emits `::warning::` for every dep that fell back to source. Treat any such warning as a blocker.

### Per-leg artifact in S3 layout

Each runner emits a `externals-<leg>-<target>.tar.gz` containing:

- `libraries/<os>/<arch>/<dep>.tar.gz` — binaries for that (system, arch) slice
- `libraries/sources/<dep>.tar.gz` — upstream sources (every leg writes identical content; final tree has one copy regardless of extract order)

Path mapping: `rpm-amd64`→`linux/amd64`, `rpm-arm64`→`linux/aarch64`, `macos-intel64`→`darwin/amd64`, `macos-arm64`→`darwin/aarch64`, `windows-i686`→`windows`. Inner and outer tarballs use `tar --owner=0 --group=0 --no-same-owner` so they extract under any UID.

### Supporting scripts under `packages/externals/`

- **`external_sources.sh`** — manifest: one line per dep with the upstream URL template, archive format, target dir, and a `linux_only` flag (filtered out on macOS/Windows legs). Adds explicit sqlite handling because sqlite's year-coded URL doesn't follow the same pattern as other deps.
- **`build_external.sh`** — runs inside the builder image. Reads `DEPS_VERSION` from `src/Makefile` and uses it as the source of truth for any blob this workflow re-ships rather than compiles (currently cpython, libbpf-bootstrap). Seeds `src/external/` via `make deps EXTERNAL_SRC_ONLY=yes`, applies `--dependencies` overrides from upstream URLs, runs `make build-external`, then snapshots source + binary trees per dep. Includes a `stage_precompiled` workaround for libbpf-bootstrap, which can't compile in the legacy CentOS 6 agent image (needs clang ≥ 7 + UAPI ≥ 4.13).
- **`generate_external.sh`** — host-side driver. Dispatches each leg into the right execution mode (`docker run` for Linux, host shell for macOS/Windows), then re-packs the per-dep zips that `build_external.sh` produced into the `libraries/<os>/<arch>/` + `libraries/sources/` tree and wraps that into the final per-leg tarball.
- **`smoke_build.sh`** — runs inside a builder image during `smoke-build`. Sets `RESOURCES_URL=file://…` so `make deps` consumes the local consolidated tree instead of `packages.wazuh.com`, then runs the real `make` build. Exit status mirrors the build; the workflow's analysis step greps the log for unintended source recompiles.

### `docs/dev/build-external-dependencies.md`

New dev-docs page (also linked from `docs/SUMMARY.md`): when to use the workflow, how to dispatch it (UI + `gh workflow run` examples), the fixed 7-entry matrix, the output layout, the safe publish order (run → upload → separate PR to bump `DEPS_VERSION`), and caveats (don't bump `DEPS_VERSION` in the same branch; `libbpf-bootstrap` and `cpython` are re-shipped not rebuilt; smoke-build warnings are blockers; consolidate tie-breaking).

### Results and Evidence

Reference runs (visible via the GitHub Actions UI on this branch — dispatched via `5_builderpackage_binutils.yml` while the new workflow file is not yet on the default branch):

- **`25949938525`** (2026-05-16) — `dependencies=curl:8.20.0;libarchive:3.8.7;sqlite:3.53.1;lzma:5.8.3`. Conclusion: `success`. Per the `gh run list` output — every leg, `consolidate`, and `smoke-build` jobs reported `success`; this is the run that produced the `externals-all.tar.gz` consumed by #36152.
- **`26003502345`** (2026-05-17) — `dependencies=""` (clean rebuild). Conclusion: `success`. Validates the no-override path on the current tip.

The smoke-build step is the primary correctness check: it greps the build log for `Performing build step for '<dep>_external'` and `Building … ext_<dep>.dir/`. Any match means a dep was source-recompiled because the precompiled tarball wasn't packed at the path `src/external/CMakeLists.txt` expected — an automatic regression flag for packing-path bugs. Reviewers should spot-check the smoke-build logs on the linked runs.

This PR adds a workflow + its supporting build scripts and a dev-docs page. It does not modify any compiled artifact's runtime behavior — the externals produced are byte-equivalent to what the prior out-of-band process generated when given the same inputs. The validation surface is the workflow run itself.

- Compilation without warnings on every supported platform — evidence: the linked runs include legs for each platform; reviewers please spot-check the per-leg logs:
  - [ ] Linux (rpm legs, agent + manager)
  - [ ] Windows (MinGW cross-compile)
  - [ ] MAC OS X (intel + arm)

- Memory tests for Linux
  - [ ] Coverity — N/A (no runtime code changes)
  - [ ] Valgrind (memcheck and descriptor leaks check) — N/A
  - [ ] AddressSanitizer — N/A
- Memory tests for Windows
  - [ ] Coverity — N/A
  - [ ] UMDH — N/A
- Memory tests for macOS
  - [ ] Leaks — N/A
  - [ ] AddressSanitizer — N/A

  - [ ] Added unit testing files ".ini" — N/A
  - [ ] `runtests.py` executed without errors — N/A

  - [ ] Test run in parallel — N/A (no engine code changes)
  - [ ] ASAN for test (utest/ctest) — N/A
  - [ ] TSAN for test and wazuh-engine. — N/A

- Wazuh server API/Framework
  - [ ] Run API Integration Tests — N/A

### Artifacts Affected

- New CI workflow: `5.X - Package - Build external dependencies`.
- New build scripts under `packages/externals/`: `build_external.sh`, `external_sources.sh`, `generate_external.sh`, `smoke_build.sh`.
- New dev-docs page: `docs/dev/build-external-dependencies.md` (linked from `docs/SUMMARY.md`).
- Per-leg artifacts (`externals-<leg>-<target>`) retained 14 days; the artifact intended for S3 upload is `externals-all` → `externals-all.tar.gz`, already laid out in the path structure `make deps` consumes (`libraries/<os>/<arch>/<dep>.tar.gz` + `libraries/sources/<dep>.tar.gz`).
- No runtime executable, package, or default configuration is changed.

### Configuration Changes

None. `DEPS_VERSION` in `src/Makefile` is unchanged; this PR does not bump any dependency. The version bump that consumes this workflow lives in #36152.

### Tests Introduced

No new unit or integration tests. The validation surface for this PR is the workflow run itself — exercising every leg end-to-end and then running smoke-build against the consolidated tree is the test.

## Review Checklist

- [ ] Code changes reviewed
- [x] Configuration changes documented (none introduced)
- [x] Developer documentation reflects the changes (`docs/dev/build-external-dependencies.md`)
- [x] Revert temporary CMakeLists patches that landed on this branch (moved to #36152)
- [x] Revert hijacked binutils workflow to its main contents

### Follow-up work (out of scope here)

- #36152 — first deps-version bump that uses this workflow (curl, sqlite, xz, libarchive → `DEPS_VERSION=99-29734`).
- Future deps bumps follow the documented order: dispatch this workflow against a branch with the right `CMakeLists.txt` patches and `DEPS_VERSION` pointing at the currently-published version, upload the produced `externals-all.tar.gz`, then open a PR that bumps `DEPS_VERSION` to the new directory. See `docs/dev/build-external-dependencies.md` for the safe ordering and caveats.


src/Makefile
```diff
@@ -398,7 +398,7 @@ ifneq (,$(shell curl --help all 2>/dev/null | grep -F -- '--retry-all-errors'))
 	TZDATA_CURL_FLAGS += --retry-all-errors
 endif
 TZDATA_CURL := curl $(TZDATA_CURL_FLAGS) -o
-DEPS_VERSION = 99-29585
+DEPS_VERSION = 99-29734
```
jpcerrone: Description says no deps version is getting bumped. Maybe this is a leftover from testing?

jr0me: I'll remove this before merging it belongs to a tmp commit. Thanks for raising this.

.github/workflows/5_builderpackage_externals.yml
```diff
@@ -0,0 +1,402 @@
...
+    # holds the macOS leg binaries, and windows/ holds the MinGW leg
+    # binaries. Output is a single externals-all.tar.gz the user can
+    # extract straight into S3.
+    #
+    # On Linux the agent and manager legs both build the agent dep set, so
+    # their per-leg tarballs overlap. The merge picks a winner per dep from
+    # the build output itself — no dep lists — so changing a dependency
+    # never needs a workflow change.
+    needs: build-externals
+    runs-on: ubuntu-latest
```
Nicogp: Let's avoid using `ubuntu-latest` and explicitly define `ubuntu-24.04` instead.

jr0me: Pinned the consolidate job to `ubuntu-24.04` in 7efc447eb2.

Nicogp (approved):

PR #36254, opened 2026-05-20
Restore vulnerability scanner database workflow

jotacarma90:
Restore the workflow on main so the scheduled cron (`20 0 * * *`) fires. GitHub Actions only runs schedules from the default branch; while main held the deprecated stub, the daily run never triggered.

File contents synced with branch 4.14.6.


jepalfer (approved):
LGTM

PR #36261, opened 2026-05-20
Serialize procps access to prevent modulesd crash

### LINKED ISSUE #36166: wazuh-modulesd crashed silently on Fedora 43 ARM
|---|---|---|---|---|
| 5.0.0-latest (f6efbf3) | Wazuh modulesd | Agent | Packages | Fedora 43 ARM |

It has been detected that wazuh-modulesd randomly fails and generates core dumps on Fedora 43 ARM systems (although it is suspected that this issue may affect additional systems). Two main scenarios have been identified:

- Scenario 1: Core dump generated after agent restart (modulesd starts correctly with a new PID)
- Scenario 2: Core dump generated after agent restart (modulesd does not start, and the older modulesd PID remains in /var/ossec/var/run)

```
[root@agent-fedora-43-arm64 fedora]# cat /var/ossec/VERSION.json 
{
    "version": "5.0.0",
    "stage": "beta1",
    "commit": "f6efbf3"
}
```

## Scenario 1

- ossec.log does not record any error messages.
- pgrep detects a new modulesd PID.
- wazuh-control detects modulesd as running.
- /var/ossec/var/run list a modulesd pid file with the correct pid value
- core dump: [core_no_fail.zip](https://github.com/user-attachments/files/27959213/core_no_fail.zip)

Image

## Scenario 2 

If this scenario occurs, the modulesd daemon disappears shortly after appearing when the restart command is executed.

- ossec.log does not record any error messages.
- pgrep does not detect modulesd as running.
- wazuh-control detects modulesd as not running.
- /var/ossec/var/run list a modulesd pid file that does not exist in the system
- core dump: [core_process_fail.zip](https://github.com/user-attachments/files/27959217/core_process_fail.zip)

Image

## How-To reproduce

- Install a 4.14.5 agent and connect it to a 5.0.0 Wazuh manager, then start the agent.
- Wait a few minutes to allow the agent to start properly.
- Upgrade the agent to 5.0.0-latest.
- Wait a few minutes to allow the agent to start properly.

Now, run the following command repeatedly until the error occurs:

> date && systemctl restart wazuh-agent.service && sleep 60 && pgrep -a wazuh-modulesd && date

Eventually, you will notice that the pgrep output is no longer displayed

> [root@agent-fedora-43-arm64 fedora]# date && systemctl restart wazuh-agent.service && sleep 60 && pgrep -a wazuh-modulesd && date
> Mon May 18 12:07:36 PM UTC 2026
> 23956 /var/ossec/bin/wazuh-modulesd
> Mon May 18 12:08:42 PM UTC 2026
> [root@agent-fedora-43-arm64 fedora]# date && systemctl restart wazuh-agent.service && sleep 60 && pgrep -a wazuh-modulesd && date
>
> Mon May 18 12:16:37 PM UTC 2026

## Environment

- Fedora 43 ARM
- Environment access has been provided to @MarcelKemp, please notify me to destroy the environment
  - https://jenkins-production.qa.wazuh.info/job/5.0.0/job/deployability_testing/147/

cborla:
## Description

  `wazuh-modulesd` randomly SIGSEGVs on Fedora 43 aarch64 (and any platform with glibc ≥ 2.38 on ARM) during the first process scan after an agent restart. The top frames are always  `__GI_strrchr → stat2proc → simple_readproc → readproc → SysInfo::getProcessesInfo`. The root cause is a thread-safety violation in the vendored `procps 2.8.3`: `simple_readproc`,
  `simple_readtask`, `file2str` and `get_proc_stats` keep their scratch buffers (`sbuf[1024]`, `struct stat sb`, `filename[80]`, `path`) declared as `static`. Those buffers are a single instance per process, shared across every thread. The Syscollector module and SCA's `ProcessRuleEvaluator` both call `SysInfo::getProcessesInfo` from independent threads, so they corrupt
  each other mid-parse. On aarch64 the vectorized `strrchr` reads past the corrupted boundary and dereferences an unmapped page; on x86_64 the byte-by-byte path masks the bug and produces wrong process names instead of a crash. Upgrading the vendored copy to `procps-ng 3.3.x` does **not** fix this — upstream keeps the same `static` buffers; only the 4.x `libproc-2` API is thread-safe.

  Closes #36166

  ## Proposed Changes

  Introduces a thin C++ wrapper around procps that serializes every `openproc → readproc → closeproc` session with a process-wide `std::mutex`. All consumers in `libsysinfo.so` (Syscollector and SCA) go through the wrapper, so concurrent scans are queued instead of trampling the shared `static` buffers inside the vendored library. The wrapper lives in `data_provider/` next to
   the only call site that needs it; the vendored procps sources are left untouched, which is required because `src/external/procps/` is gitignored and rebuilt from the externals tarball on every CI run.

  - New `data_provider/src/procpsWrapperLinux.hpp` exposing `ProcpsWrapper::scanProcesses(flags, callback)`.
  - New `data_provider/src/procpsWrapperLinux.cpp` with the static mutex and RAII handles for `PROCTAB`/`proc_t`.
  - `data_provider/src/sysInfoLinux.cpp` now includes the wrapper header instead of `readproc.h`, replaces the manual `openproc`/`readproc`/`closeproc` loop in `SysInfo::getProcessesInfo`, and drops the now-unused `ProcTableDeleter`, `SysInfoProcessesTable` and `SysInfoProcess` aliases. `getProcessInfo` takes `const proc_t*` directly.

  ### Results and Evidence

https://github.com/wazuh/wazuh/pull/36261#issuecomment-4511574863
https://github.com/wazuh/wazuh/pull/36261#issuecomment-4511807495

    - [ ] Linux
    - [ ] Windows
    - [ ] MAC OS X

  - Memory tests for Linux
    - [ ] Coverity
    - [ ] AddressSanitizer
  - Memory tests for Windows
    - [ ] Coverity
    - [ ] UMDH
  - Memory tests for macOS
    - [ ] Leaks
    - [ ] AddressSanitizer

    - [ ] Test run in parallel

  - Wazuh server API/Framework

  ### Artifacts Affected

  - `libsysinfo.so` (Linux agent and manager). No changes to public headers, no ABI changes — `SysInfo::getProcessesInfo` keeps its signature and behavior. Windows, macOS and BSD builds are unaffected because the wrapper is Linux-only (`procpsWrapperLinux.cpp` matches the existing `*Linux.cpp` glob in `data_provider/CMakeLists.txt`). The vendored `src/external/procps/` tree is left intact.

  ### Configuration Changes

  N/A. No new settings, no defaults changed, no on-disk format impact.

  ### Tests Introduced

  N/A. 

  ## Review Checklist

  - [ ] Code changes reviewed

### DIFF
--- a/src/data_provider/src/procpsWrapperLinux.cpp
+++ b/src/data_provider/src/procpsWrapperLinux.cpp
@@ -1,0 +1,58 @@
+/*
+ * Wazuh SysInfo
+ * Copyright (C) 2015, Wazuh Inc.
+ *
+ * This program is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU General Public
+ * License (version 2) as published by the FSF - Free Software
+ * Foundation.
+ */
+#include "procpsWrapperLinux.hpp"
+
+#include <memory>
+#include <mutex>
+
+namespace
+{
+    std::mutex g_procpsMutex;
+
+    struct ProcTabDeleter
+    {
+        void operator()(PROCTAB* pt) const
+        {
+            if (pt)
+            {
+                closeproc(pt);
+            }
+        }
+    };
+
+    struct ProcDeleter
+    {
+        void operator()(proc_t* p) const
+        {
+            if (p)
+            {
+                freeproc(p);
+            }
+        }
+    };
+}
+
+void ProcpsWrapper::scanProcesses(int flags, const ProcCallback& callback)
+{
+    std::lock_guard<std::mutex> lock(g_procpsMutex);
+
+    std::unique_ptr<PROCTAB, ProcTabDeleter> proctab{openproc(flags)};
+
+    if (!proctab)
+    {
+        return;
+    }
+
+    while (auto* raw = readproc(proctab.get(), nullptr))
+    {
+        std::unique_ptr<proc_t, ProcDeleter> process{raw};
+        callback(process.get());
+    }
+}
--- a/src/data_provider/src/procpsWrapperLinux.hpp
+++ b/src/data_provider/src/procpsWrapperLinux.hpp
@@ -1,0 +1,29 @@
+/*
+ * Wazuh SysInfo
+ * Copyright (C) 2015, Wazuh Inc.
+ *
+ * This program is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU General Public
+ * License (version 2) as published by the FSF - Free Software
+ * Foundation.
+ */
+#pragma once
+
+#include <functional>
+
+extern "C"
+{
+#include "readproc.h"
+}
+
+namespace ProcpsWrapper
+{
+    using ProcCallback = std::function<void(const proc_t*)>;
+
+    // Serializes a full openproc/readproc/closeproc iteration with a
+    // process-wide mutex. The vendored procps 2.8.3 keeps scratch buffers as
+    // `static` inside simple_readproc/simple_readtask/file2str/get_proc_stats,
+    // so concurrent iterations from different threads corrupt each other and
+    // can crash inside stat2proc (see issue #36166).
+    void scanProcesses(int flags, const ProcCallback& callback);
+}
--- a/src/data_provider/src/sysInfoLinux.cpp
+++ b/src/data_provider/src/sysInfoLinux.cpp
@@ -20,7 +20,7 @@
 #include "cmdHelper.h"
 #include "osinfo/sysOsParsers.h"
 #include "sysInfo.hpp"
-#include "readproc.h"
+#include "procpsWrapperLinux.hpp"
 #include "networkUnixHelper.h"
 #include "networkHelper.h"
 #include "network/networkLinuxWrapper.h"
@@ -42,21 +42,6 @@
 #include "firefox.hpp"
 
 using ProcessInfo = std::unordered_map<int64_t, std::pair<int32_t, std::string>>;
-
-struct ProcTableDeleter
-{
-    void operator()(PROCTAB* proc)
-    {
-        closeproc(proc);
-    }
-    void operator()(proc_t* proc)
-    {
-        freeproc(proc);
-    }
-};
-
-using SysInfoProcessesTable = std::unique_ptr<PROCTAB, ProcTableDeleter>;
-using SysInfoProcess        = std::unique_ptr<proc_t, ProcTableDeleter>;
 
 static void parseLineAndFillMap(const std::string& line, const std::string& separator, std::map<std::string, std::string>& systemInfo)
 {
@@ -90,7 +75,7 @@ static bool getSystemInfo(const std::string& fileName, const std::string& separa
     return ret;
 }
 
-static nlohmann::json getProcessInfo(const SysInfoProcess& process)
+static nlohmann::json getProcessInfo(const proc_t* process)
 {
     nlohmann::json jsProcessInfo{};
     // Current process information
@@ -610,21 +595,14 @@ nlohmann::json SysInfo::getPorts() const
 
 void SysInfo::getProcessesInfo(std::function<void(nlohmann::json&)> callback) const
 {
+    const int flags = PROC_FILLMEM | PROC_FILLSTAT | PROC_FILLSTATUS | PROC_FILLARG
+                      | PROC_FILLGRP | PROC_FILLUSR | PROC_FILLCOM | PROC_FILLENV;
 
-    const SysInfoProcessesTable spProcTable
+    ProcpsWrapper::scanProcesses(flags, [&callback](const proc_t* process)
     {
-        openproc(PROC_FILLMEM | PROC_FILLSTAT | PROC_FILLSTATUS | PROC_FILLARG | PROC_FILLGRP | PROC_FILLUSR | PROC_FILLCOM | PROC_FILLENV)
-    };
-
-    SysInfoProcess spProcInfo { readproc(spProcTable.get(), nullptr) };
-
-    while (nullptr != spProcInfo)
-    {
-        // Get process information object and push it to the caller
-        auto processInfo = getProcessInfo(spProcInfo);
+        auto processInfo = getProcessInfo(process);
         callback(processInfo);
-        spProcInfo.reset(readproc(spProcTable.get(), nullptr));
-    }
+    });
 }
 
 void SysInfo::getPackages(std::function<void(nlohmann::json&)> callback) const


src/data_provider/src/procpsWrapperLinux.cpp
```diff
@@ -0,0 +1,58 @@
...
+        {
+            if (p)
+            {
+                freeproc(p);
+            }
+        }
+    };
+}
+
+void ProcpsWrapper::scanProcesses(int flags, const ProcCallback& callback)
```
hernanvalenzuela, line 42:
```c++
void ProcpsWrapper::scanProcesses(int flags, const ProcCallback& callback)
{
    std::vector<std::unique_ptr<proc_t, ProcDeleter>> snapshot;
    {
        std::lock_guard<std::mutex> lock(g_procpsMutex);
        std::unique_ptr<PROCTAB, ProcTabDeleter> proctab{openproc(flags)};
        if (!proctab) return;
        while (auto* raw = readproc(proctab.get(), nullptr))
        {
            snapshot.emplace_back(raw);
        }
    } 
  // ← mutex release
    for (auto& p : snapshot)
    {
        callback(p.get());   // build JSON and evaluation outside lock
    }
}
```

hernanvalenzuela:
Artifacts:
- wazuh-agent_5.0.0-0_aarch64_16a984a.rpm (11 MB)
- wazuh-agent-debuginfo_5.0.0-0_aarch64_16a984a.rpm (26.5 MB)

### Target environment

| Host   | agent-fedora-43-arm64 (EC2) |
|----|------|
| OS     | Fedora 43 |
| Kernel | Linux 6.x aarch64 |
| Arch | aarch64  |
| glibc | 2.41 (ships SIMD-vectorized strrchr/strcpy for aarch64) |


### Pre-conditions baseline (unpatched 5.0.0-latest)

Before installing the PR build, the bug was 100% reproducible:

```bash
$ for i in $(seq 1 200); do
    sudo systemctl restart wazuh-agent.service
    sleep 5
    pgrep -a wazuh-modulesd >/dev/null || { echo "FAIL iter $i"; break; }
    echo "OK $i"
  done
OK 1
OK 2
...
```
Three captured backtraces all share the same signature:

Thread (crashed):

1. strrchr (libc.so.6)
2. stat2proc        (libwazuhext.so + 0x7fd788)
3. simple_readproc  (libwazuhext.so + 0x7fe1f8)
4. readproc         (libwazuhext.so + 0x7fe870)
5.  SysInfo::getProcessesInfo  sysInfoLinux.cpp:626
6.  SysInfo::processes         sysInfo.cpp:32 / :47
7.  ProcessRuleEvaluator::Evaluate  sca_policy_check.cpp:556  (or scan from syscollector)

Other thread (concurrent):
  …  simple_readproc → file2str / file2strvec → openat

Two threads were simultaneously inside simple_readproc, sharing the vendored procps's static char sbuf[1024] (procps readproc.c:541). The race corrupts that buffer between read(fd, sbuf, cap-1) and stat2proc(sbuf, …); on aarch64 with the vectorized strrchr, the parser then reads past a corrupted NUL into an unmapped page and segfaults.

###  Installation of the PR build

#### Download zips from the CI run, copy to the VM, unzip:
```bash
unzip wazuh-agent_5.0.0-0_aarch64_16a984a.rpm.zip
unzip wazuh-agent-debuginfo_5.0.0-0_aarch64_16a984a.rpm.zip
```

# Backup current libwazuhext.so
`sudo cp -p /var/ossec/lib/libwazuhext.so /var/ossec/lib/libwazuhext.so.pre-pr36261`

Stop, install both packages together (debuginfo enforces an exact version match against wazuh-agent, so they must be replaced in a single transaction):
```bash
sudo systemctl stop wazuh-agent
sudo rpm -Uvh --force --oldpackage \
    /tmp/wazuh-agent_5.0.0-0_aarch64_16a984a.rpm \
    /tmp/wazuh-agent-debuginfo_5.0.0-0_aarch64_16a984a.rpm
```

```bash
sudo restorecon -Rv /var/ossec/lib /var/ossec/bin
sudo systemctl start wazuh-agent
```

Post-install state:

```bash
$ rpm -qa | grep wazuh
wazuh-agent-5.0.0-0.aarch64
wazuh-agent-debuginfo-5.0.0-0.aarch64
```

```bash
$ sudo ls -lh /var/ossec/lib/libwazuhext.so /var/ossec/lib/libwazuhext.so.pre-pr36261
-rwxr-x---. 1 root wazuh 13M May 20 20:19 /var/ossec/lib/libwazuhext.so
-rwxr-x---. 1 root wazuh 20M May 20 15:12 /var/ossec/lib/libwazuhext.so.pre-pr36261
```

### Test methodology

Three independent validation tests were run, each targeting a different aspect of the fix:

#### Test 1 — Stand-alone race reproducer (/tmp/race)

A 60-line C program that dlopens /var/ossec/lib/libsysinfo.so at runtime and exercises sysinfo_processes() from N pthreads, each running 500 iterations. This isolates the bug to the libsysinfo + procps stack without involving wazuh-modulesd, SCA, Syscollector, or systemd timing.

```c
// /tmp/race.c — gcc /tmp/race.c -o /tmp/race -lpthread -ldl
#define _GNU_SOURCE
#include <dlfcn.h>
#include <pthread.h>
#include <stdio.h>
#include <stdlib.h>

typedef int  (*sysinfo_processes_fn)(void **);
typedef void (*sysinfo_free_fn)(void **);
static sysinfo_processes_fn sysinfo_processes;
static sysinfo_free_fn      sysinfo_free_result;
#define ITER_PER_THREAD 500

static void *worker(void *arg) {
    long id = (long)arg;
    for (int i = 0; i < ITER_PER_THREAD; ++i) {
        void *js = NULL;
        if (sysinfo_processes(&js) == 0 && js)
            sysinfo_free_result(&js);
    }
    fprintf(stderr, "thread %ld done\n", id);
    return NULL;
}

int main(int argc, char **argv) {
    int n = (argc > 1) ? atoi(argv[1]) : 8;
    void *h = dlopen("/var/ossec/lib/libsysinfo.so", RTLD_NOW | RTLD_GLOBAL);
    if (!h) { fprintf(stderr, "dlopen: %s\n", dlerror()); return 1; }
    sysinfo_processes   = (sysinfo_processes_fn) dlsym(h, "sysinfo_processes");
    sysinfo_free_result = (sysinfo_free_fn)      dlsym(h, "sysinfo_free_result");
    pthread_t *t = calloc(n, sizeof *t);
    for (long i = 0; i < n; ++i)
        pthread_create(&t[i], NULL, worker, (void *)i);
    for (int i = 0; i < n; ++i) pthread_join(t[i], NULL);
    free(t);
    return 0;
}
```

Execution:
```bash
sudo bash -c 'ulimit -c unlimited; /tmp/race 16; echo exit=$?'
```
```bash
sudo bash -c 'ulimit -c unlimited; /tmp/race 32; echo exit=$?'
```
```bash
for i in 1 2 3 4 5; do
    sudo bash -c 'ulimit -c unlimited; /tmp/race 32' \
        && echo "run $i ok" || echo "run $i FAIL"
done
```

#### Test 2 — Non-disruptive concurrent-load script

The agent is left running; the script applies stress in parallel and verifies modulesd survives without being respawned. Test plan:

1. Confirm the new RPM is installed and the live wazuh-modulesd PID.
2. Run /tmp/race 16 × 5 and /tmp/race 32 × 5 in isolation.
3. Spawn 3 concurrent /tmp/race 16 instances while the agent runs, monitoring that modulesd does not die.
4. Sustain /tmp/race 32 × 10 back-to-back.
5. Verify the modulesd PID is unchanged at the end (no crash + systemd respawn).
6. Verify zero new files appear under /var/lib/coredumps/.

The full script is included in §A of the appendix.

#### Test 3 — Original ticket reproducer (50× agent restart)

The exact loop that the original issue reports as failing on iteration 3:

```bash
for i in $(seq 1 50); do
    sudo systemctl restart wazuh-agent.service
    sleep 5
    pgrep -x wazuh-modulesd >/dev/null || { echo "FAIL iter $i"; break; }
    echo "OK $i"
done
```

Each restart forces SCA and Syscollector to start scanning processes simultaneously — exactly the concurrency window that exposes the static-buffer race in unpatched builds.

hernanvalenzuela:
## Results:

### Test 1 — /tmp/race

```bash
$ sudo bash -c 'ulimit -c unlimited; /tmp/race 16; echo exit=$?'
thread 0 done … thread 15 done
exit=0
```

```bash
$ sudo bash -c 'ulimit -c unlimited; /tmp/race 32; echo exit=$?'
thread 0 done … thread 31 done
exit=0
```

```
$ for i in 1 2 3 4 5; do
    sudo bash -c 'ulimit -c unlimited; /tmp/race 32' \
      && echo "run $i ok" || echo "run $i FAIL"
  done

run 1 ok
run 2 ok
run 3 ok
run 4 ok
run 5 ok
...
```

| Configuration | Concurrent calls | Result |
|----------------|--------------------|--------|
| 16 threads × 500 iter | 8 000 | exit=0 |
| 32 threads × 500 iter  | 16 000  | exit=0 |
| 32 threads × 500 iter × 5 runs | 80 000 | 5/5 exit=0 |

Compare against the unpatched baseline (same host, same binary /tmp/race): consistent SIGSEGV (exit=139) in < 1 s, three coredumps preserved.

### Test 2 — Non-disruptive load

```bash
===== PR #36261 — non-disruptive validation =====
Host:            agent-fedora-43-arm64
Date (UTC):      Thu May 21 03:44:31 PM UTC 2026
Cores baseline:  15
modulesd PID:    2292674

----- 1. Package + library -----
[PASS] wazuh-agent 5.0.0-0 installed (PR build)
mtime=2026-05-20 20:19:48.000000000 +0000

----- 2. Agent process (no restart) -----
[PASS] wazuh-modulesd already running pid=2292674 — leaving it alone

----- 3. /tmp/race in isolation (exercises wrapper) -----
[PASS] /tmp/race 16: 5/5 runs OK
[PASS] /tmp/race 32: 5/5 runs OK

----- 4. Concurrent load: race + agent both touching libwazuhext -----
[PASS] Agent stayed alive during concurrent /tmp/race load
[PASS] All 3 parallel race instances completed cleanly

----- 5. Sustained stress: /tmp/race 32 × 10 times back to back -----
[PASS] Sustained stress: 10/10 runs OK

----- 6. Post-test agent state -----
[PASS] wazuh-modulesd PID unchanged (no crash+respawn): 2292674
       ELAPSED 01:09:12

----- 7. New coredumps during validation -----
[INFO] baseline=15  now=15  new=0
[PASS] No new coredumps generated

===== SUMMARY =====  9/9 PASS
```

** Approximate concurrent calls across the full test: ~272 000.
** Agent uptime under load: 1h 9min, PID stable (2292674 throughout).

### Test 3 — Original reproducer (50× restart)

```bash
$ for i in $(seq 1 50); do
    sudo systemctl restart wazuh-agent.service
    sleep 5
    pgrep -x wazuh-modulesd >/dev/null || { echo "FAIL iter $i"; break; }
    echo "OK $i"
  done
OK 1
OK 2
OK 3
…
OK 49
OK 50
```

50/50 successful restarts, no FAIL. Pre-patch, this same loop reliably aborted between iterations 3 and 10 every attempt.

Coredump audit (final)

```
$ sudo find /var/lib/coredumps -name 'core-*' -newer /var/ossec/lib/libwazuhext.so
(no output)
```

`$ sudo ls /var/lib/coredumps/core-wazuh-modulesd-* | wc -l `
9   ← all from before the PR install (May 18-19)

Zero new coredumps generated during the entire validation session.

## Conclusion

|     | Pre-PR | Post-PR | 
| ---| --------| ---------| 
|  /tmp/race 16 exit  | 139 (SIGSEGV) in < 1 s  |  0  | 
|  /tmp/race 32 exit  | 139 (SIGSEGV) in < 1 s  |  0 (5/5 runs) | 
|  Concurrent load + live agent (90 s)  |  n/a — agent crashed    |  modulesd PID stable | 
|  Sustained /tmp/race 32 × 10  |  n/a   |  10/10 exit=0 | 
| 50× agent restart loop |  FAIL at iter 3 |  50/50 OK  | 
| New coredumps during validation |  ≥1 per attempt | 0 | 
| Approx. concurrent calls during test  |  n/a  |  ~272 000 | 

The mutex-based ProcpsWrapper::scanProcesses introduced by this PR correctly serializes access to procps's static scratch buffers (simple_readproc/simple_readtask/file2str/get_proc_stats) and eliminates the data race that previously crashed stat2proc → strrchr on aarch64 with vectorized glibc.


Appendix A — Non-disruptive validation script

/tmp/validate-pr36261-nokick.sh (the exact script that produced §5 Test 2 results):

Validates PR #36261 without restarting wazuh-agent.  Usage: sudo bash /tmp/validate-pr36261-nokick.sh

```bash
set -u
RESULTS=/tmp/validate-nokick-$(date +%Y%m%d-%H%M%S)
mkdir -p "$RESULTS"
LOG="$RESULTS/run.log"

pass() { echo "[PASS] $1" | tee -a "$LOG"; }
fail() { echo "[FAIL] $1" | tee -a "$LOG"; }
info() { echo "[INFO] $1" | tee -a "$LOG"; }

baseline_cores=$(find /var/lib/coredumps -maxdepth 1 -name 'core-*' 2>/dev/null | wc -l)
INITIAL_MODULESD_PID=$(pgrep -x wazuh-modulesd || echo "")

{
  echo "===== PR #36261 — non-disruptive validation ====="
  echo "Host:            $(uname -n)"
  echo "Date (UTC):      $(date -u)"
  echo "Cores baseline:  $baseline_cores"
  echo "modulesd PID:    ${INITIAL_MODULESD_PID:-<not running>}"
  echo
} | tee "$LOG"

# 1. Package and library
echo "----- 1. Package + library -----" | tee -a "$LOG"
rpm -qa | grep wazuh | tee -a "$LOG"
if rpm -q wazuh-agent | grep -q '5.0.0-0'; then
    pass "wazuh-agent 5.0.0-0 installed (PR build)"
else
    fail "wazuh-agent 5.0.0-0 NOT installed"
fi
sudo ls -lh /var/ossec/lib/libwazuhext.so | tee -a "$LOG"
info "libwazuhext.so mtime=$(sudo stat -c '%y' /var/ossec/lib/libwazuhext.so)"

# 2. Agent must already be running
echo "----- 2. Agent process (no restart) -----" | tee -a "$LOG"
if [[ -z "$INITIAL_MODULESD_PID" ]]; then
    fail "wazuh-modulesd is NOT running"
    exit 1
fi
pass "wazuh-modulesd already running pid=$INITIAL_MODULESD_PID — leaving it alone"
ps -p "$INITIAL_MODULESD_PID" -o pid,etime,rss,comm | tee -a "$LOG"

# 3. /tmp/race in isolation
echo "----- 3. /tmp/race in isolation -----" | tee -a "$LOG"
for n in 16 32; do
    fails=0
    for i in 1 2 3 4 5; do
        if bash -c 'ulimit -c unlimited; /tmp/race '"$n" \
                > "$RESULTS/race$n-$i.log" 2>&1; then
            echo "  race $n run $i: exit=0 OK" | tee -a "$LOG"
        else
            echo "  race $n run $i: exit=$? FAIL" | tee -a "$LOG"
            fails=$((fails+1))
        fi
    done
    [[ $fails -eq 0 ]] \
        && pass "/tmp/race $n: 5/5 runs OK" \
        || fail "/tmp/race $n: $fails failed"
done

# 4. Concurrent load
echo "----- 4. Concurrent load -----" | tee -a "$LOG"
info "Spawning 3 parallel /tmp/race instances"
for i in 1 2 3; do
    bash -c 'ulimit -c unlimited; /tmp/race 16' \
        > "$RESULTS/parallel-$i.log" 2>&1 &
done
agent_alive=1
for s in 1 2 3 4 5 6; do
    if ! kill -0 "$INITIAL_MODULESD_PID" 2>/dev/null; then
        agent_alive=0; break
    fi
    sleep 5
done
wait
[[ $agent_alive -eq 1 ]] \
    && pass "Agent stayed alive during concurrent /tmp/race load" \
    || fail "Agent died during concurrent /tmp/race load"

# 5. Sustained stress
echo "----- 5. Sustained stress -----" | tee -a "$LOG"
fails=0
for i in $(seq 1 10); do
    bash -c 'ulimit -c unlimited; /tmp/race 32' \
        > "$RESULTS/sustained-$i.log" 2>&1 \
        || fails=$((fails+1))
done
[[ $fails -eq 0 ]] \
    && pass "Sustained stress: 10/10 runs OK" \
    || fail "Sustained stress: $fails/10 failed"

# 6. Agent PID unchanged
echo "----- 6. Post-test agent state -----" | tee -a "$LOG"
CURRENT_PID=$(pgrep -x wazuh-modulesd || echo "")
if [[ "$CURRENT_PID" == "$INITIAL_MODULESD_PID" ]]; then
    pass "wazuh-modulesd PID unchanged (no crash+respawn): $CURRENT_PID"
else
    fail "wazuh-modulesd was respawned: was $INITIAL_MODULESD_PID, now $CURRENT_PID"
fi
ps -p "${CURRENT_PID:-0}" -o pid,etime,rss,comm 2>/dev/null | tee -a "$LOG"

# 7. Coredump audit
echo "----- 7. New coredumps -----" | tee -a "$LOG"
current_cores=$(find /var/lib/coredumps -maxdepth 1 -name 'core-*' 2>/dev/null | wc -l)
new_cores=$((current_cores - baseline_cores))
info "baseline=$baseline_cores  now=$current_cores  new=$new_cores"
[[ $new_cores -eq 0 ]] \
    && pass "No new coredumps generated" \
    || fail "$new_cores new coredumps during test"

# Summary
echo "===== SUMMARY =====" | tee -a "$LOG"
grep -E '^\[(PASS|FAIL)\]' "$LOG"
grep -q '^\[FAIL\]' "$LOG" && exit 1 || exit 0
```

Appendix B — Restart loop reproducer (50 iterations)

``` bash
for i in $(seq 1 50); do
    sudo systemctl restart wazuh-agent.service
    sleep 5
    pgrep -x wazuh-modulesd >/dev/null || { echo "FAIL iter $i"; break; }
    echo "OK $i"
done
```

src/data_provider/src/procpsWrapperLinux.cpp
cborla: Thanks for the suggestion. I agree that this would reduce the time spent holding `g_procpsMutex` by moving the callback execution outside the procps critical section.

However, I would prefer to keep this PR focused on the crash fix. The snapshot approach changes the current streaming behavior, may introduce a transient memory peak by keeping all `proc_t` entries alive at once, and could also affect shutdown/cancellation latency unless explicit stop checks are added before or during callback processing.

I think this is a valid follow-up optimization, but it should be evaluated separately with proper memory/RSS measurements and shutdown behavior validation before including it.

hernanvalenzuela (approved):

PR #36246, opened 2026-05-20
Improve tmp_file path validation in cluster DAPI

vikman90:
## Description

Enhanced path validation in the `send_tmp_file` method of the cluster DAPI to ensure temporary files are properly constrained to the designated temporary directory.

**Credits:** Thanks to @moltenbit for reporting this issue.

## Proposed Changes

- Modified `DistributedAPI.send_tmp_file()` to validate that `tmp_file` parameter points to files within `OSSEC_TMP_PATH`
- Added path resolution using `os.path.realpath()` to handle symbolic links and relative paths correctly
- Changed base directory from `WAZUH_PATH` to `OSSEC_TMP_PATH` for stricter containment
- Updated existing tests to use relative paths within the tmp directory

### Results and Evidence

All tests passing:
```
======================== 47 passed, 8 warnings in 0.87s ========================
```

New tests validate:
- Invalid path formats are rejected
- Absolute paths are rejected
- Parent directory references are rejected
- Symbolic link resolution is handled correctly
- Valid relative paths within tmp directory work as expected

### Artifacts Affected

- wazuh-clusterd

### Configuration Changes

No configuration changes required.

### Documentation Updates

No documentation updates required.

### Tests Introduced

- `test_send_tmp_file_path_validation`: Parametrized test with 7 cases for various path formats
- `test_send_tmp_file_empty_parameter`: Tests missing parameter handling
- `test_send_tmp_file_absolute_path`: Tests absolute path rejection
- `test_send_tmp_file_parent_directory`: Tests parent directory reference rejection
- `test_send_tmp_file_symlink_resolution`: Tests symlink resolution handling

## Review Checklist

- [ ] Code changes reviewed
- [ ] Relevant evidence provided
- [ ] Tests cover the new functionality
- [ ] Configuration changes documented
- [ ] Developer documentation reflects the changes
- [ ] Meets requirements and/or definition of done
- [ ] No unresolved dependencies with other issues

PR #36044, opened 2026-05-12
Remove 4_X workflows code from main

Miguevrgo:
## Description

Refactoring of 4.x GitHub Actions workflows to modernize action versions and improve efficiency. Partial extraction from #35319, focusing exclusively on the 4.x workflow files.

Closes #35272 (partial - 4.x workflows only)

## Proposed Changes

- Bump all GitHub Actions to latest versions (`actions/checkout@v6`, `actions/setup-python@v4`, `actions/upload-artifact@v4`, `actions/download-artifact@v4`)
- Add `nproc`-based parallelism where applicable
- Add colored log output and consistent environment variable handling
- Remove deprecated inline patterns in favor of direct step definitions
- Scope: exclusively `.github/workflows/4_*.yml` files (75 files)

### Results and Evidence

N/A - CI workflow-only changes, no runtime behavior affected.

### Manual tests with their corresponding evidence

- Compilation without warnings on every supported platform
  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Log syntax and correct language review

- Memory tests for Linux
  - [ ] Coverity
  - [ ] Valgrind (memcheck and descriptor leaks check)
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests _(Wazuh v4.x)_
  - [ ] Added unit testing files ".ini"
  - [ ] `runtests.py` executed without errors

- Engine _(Wazuh v5.x and above)_
  - [ ] Test run in parallel
  - [ ] ASAN for test (utest/ctest)
  - [ ] TSAN for test and wazuh-engine.

- Wazuh server API/Framework
  - [ ] Run API Integration Tests

### Artifacts Affected

N/A - Only CI workflow YAML files modified. No production artifacts affected.

### Configuration Changes

N/A

### Tests Introduced

N/A - Changes are limited to CI workflow definitions.

## Review Checklist

- [ ] Code changes reviewed
- [ ] Relevant evidence provided
- [ ] Tests cover the new functionality
- [ ] Configuration changes documented
- [ ] Developer documentation reflects the changes
- [ ] Meets requirements and/or definition of done
- [ ] No unresolved dependencies with other issues

PR #36332, opened 2026-05-22
Restore working vulnerability scanner database workflow

jotacarma90:
The previous restore (PR #36246) brought the workflow from branch 4.14.6, which is missing two compatibility steps required to build the vulnerability database for older versions (4.8.0, 4.10.0, 4.11.0, 4.13.0):

- Renaming `wazuh_version` to `content_version` in composite actions of older branches.
- Restoring the missing template file expected by 4.8.0 at `templates/vd_states_template.json`.

This PR replaces the workflow with the version that was in `main` before its deprecation in PR #36044, which includes both steps and was the last known working version.


jepalfer (approved):
LGTM

PR #36789, opened 2026-06-08
Prevent data race in schema validator factory concurrent initialization

### LINKED ISSUE #36787: Schema validator: data race on the process-wide singleton causes spurious "No schema validator found for index" warnings
## Summary

`wazuh-modulesd` intermittently logs warnings such as:

```
2026/06/06 04:49:27 wazuh-modulesd:syscollector: WARNING: No schema validator found for index: wazuh-states-inventory-hardware. Queuing message without validation.
```

even though the schema for that index **is** correctly embedded in `libschema_validator.so`. The warning is harmless in effect (the message is queued and sent anyway), but it is misleading and points to a real **data race**: `SchemaValidator::SchemaValidatorFactory` is a process-wide singleton with **no synchronization**, and it is initialized concurrently by several modules that run in the same `wazuh-modulesd` process. Concurrent initialization corrupts the underlying `std::map`, so a validator that was successfully loaded can become unreachable.

## Component

- `src/shared_modules/schema_validator/` (`SchemaValidatorFactory`)
- Consumers: `wazuh_modules/syscollector`, `wazuh_modules/sca`

## Environment

- Observed on agent `T1-agent-debian-12-amd64` after an upgrade via WPK.
- Reproduced/analyzed against the nightly build:
  `https://packages-staging.xdrsiem.wazuh.info/nightly/5.0.0/wpk/linux/deb/amd64/wazuh_agent_v5.0.0_linux_amd64.deb.wpk`
- Version: 5.0.0

## Observed behavior

The warning appears for `wazuh-states-inventory-hardware` (and could be any index) shortly after the agent restarts following the upgrade. It is **intermittent** and timing-dependent.

## Why it is NOT a build/embedding problem

The first natural hypothesis was that the schema was not embedded at build time (schemas are downloaded during `make deps` and embedded into `libschema_validator.so` via `schemaResources.cpp`). This was **ruled out empirically** by unpacking the actual WPK that was deployed:

- `libschema_validator.so` embeds **all 17 schemas**, including `inventory-hardware` with the correct `index_patterns` (`wazuh-states-inventory-hardware*`). Every embedded schema parses as valid JSON and contains `template.mappings.properties` (hardware: `dynamic: strict`).
- `libsyscollector.so` uses the exact index string `wazuh-states-inventory-hardware`, which matches the name the validator registers after stripping the `*` from `index_patterns[0]`.

So the build artifact is correct, and the index strings match. Once `initialize()` completes, `getValidator("wazuh-states-inventory-hardware")` must succeed. The warning can therefore only occur if the map is observed in an inconsistent state.

## Root cause

`SchemaValidatorFactory::getInstance()` returns a function-local `static` instance — a **single object shared by every module in the `wazuh-modulesd` process**:

```cpp
// src/shared_modules/schema_validator/src/schemaValidator.cpp
SchemaValidatorFactory& SchemaValidatorFactory::getInstance()
{
    static SchemaValidatorFactory instance;   // shared across all modules
    return instance;
}
```

`SchemaValidatorFactory::Impl` has **no mutex/atomic**. `m_validators` (`std::map`) and `m_initialized` (`bool`) are plain members, and `initializeFromEmbeddedResources()` mutates the map in a loop:

```cpp
for (const auto& [_, content] : embeddedSchemas) {
    loadSchemaFromString(content);     // m_validators[schemaName] = validator;
}
m_initialized = !m_validators.empty();
```

Three modules initialize this **same** singleton, each with a *check-then-act* guard that lives **outside** the factory:

| Module | Call site |
|---|---|
| syscollector | `src/wazuh_modules/syscollector/src/syscollectorImp.cpp:2209-2211` |
| sca | `src/wazuh_modules/sca/sca_impl/src/sca_impl.cpp:445-447` |
| vulnerability_scanner | `src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScannerFacade.cpp:178` |
| (C API) | `src/shared_modules/schema_validator/src/schemaValidator_c.cpp:13` |

```cpp
if (!validatorFactory.isInitialized())   // check
    validatorFactory.initialize();         // act  — no lock, not atomic
```

On an agent, **syscollector and SCA run in the same `wazuh-modulesd` process, each in its own thread**, and both call their `initSyncProtocol()` → `initialize()` at startup, roughly simultaneously. The race:

1. syscollector thread: `isInitialized()` → `false` → enters `initialize()` → starts inserting into `m_validators`.
2. SCA thread (concurrently): `isInitialized()` is still `false` (syscollector hasn't finished) → **also** enters `initialize()` → **also** inserts into the **same** `m_validators`.
3. Two threads inserting into the same `std::map` simultaneously is a **data race / undefined behaviour**: red-black-tree rebalancing (rotations) can drop or hide a key. One thread sets `m_initialized = true` while the other is still inserting.
4. A subsequent validation sees `isInitialized() == true` but `getValidator("wazuh-states-inventory-hardware")` returns `nullptr` → the warning.

This matches the symptom perfectly: intermittent, timing-dependent, observed right after an upgrade restart (when all modules initialize at once), and "hardware" being affected is essentially luck-of-the-draw — any index could be lost.

This is a defect introduced together with the `schema_validator` feature (around PR #36034): a process-shared, lock-free singleton initialized concurrently from multiple modules.

## Steps to verify the build is fine (and the race is the cause)

Unpack the WPK and inspect the shipped library:

```bash
# WPK256 -> "merged" -> .deb -> data.tar.gz -> /var/ossec/lib/libschema_validator.so
strings libschema_validator.so | grep -o 'wazuh-states-[a-z-]*' | sort -u   # all 17 present, incl. hardware
strings libsyscollector.so     | grep wazuh-states-inventory-hardware       # exact match
```

All schemas are present and valid, so the only remaining explanation is the concurrent-init race described above.

## Proposed fix

Make the factory thread-safe and idempotent inside `SchemaValidatorFactory`:

- Add a `std::shared_mutex` to `Impl`.
- `initialize()`: take an **exclusive** lock and return early if already initialized (move the check-then-act under the lock). This guarantees the embedded resources are loaded exactly once and never mutated while another thread reads the map.
- `getValidator()` / `isInitialized()`: take a **shared** lock (allows concurrent validation across modules, blocks only during init/reset).
- `reset()`: exclusive lock.

Nicogp:
## Description

`SchemaValidator::SchemaValidatorFactory` is a process-wide singleton (`getInstance()` returns a function-local `static`) that is initialized from several modules running in the same `wazuh-modulesd` process — `syscollector`, `sca` — each with a *check-then-act* guard (`if (!isInitialized()) initialize();`) that lives **outside** the factory.

The factory had **no synchronization**: `m_validators` (`std::map`) and `m_initialized` (`bool`) were plain members, and `initializeFromEmbeddedResources()` mutates the map in a loop. On an agent, `syscollector` and `sca` initialize the singleton concurrently at startup, so two threads can enter `initialize()` and insert into the **same** `std::map` simultaneously. This is a data race / undefined behaviour: the red-black tree gets corrupted, which manifests as a lost (unreachable) key, a `SIGSEGV`, or an infinite loop.

The observed production symptom is the "soft" variant of the corruption — a validator that was loaded becomes unreachable while the factory reports itself initialized:

```
2026/06/06 04:49:27 wazuh-modulesd:syscollector: WARNING: No schema validator found for index: wazuh-states-inventory-hardware. Queuing message without validation.
```

(The schema **is** correctly embedded in `libschema_validator.so`; this was verified by unpacking the deployed WPK — all 17 schemas are present and valid and the index strings match. The warning is therefore not a build/embedding problem.)

This PR makes the factory thread-safe and the initialization idempotent. It also adds a CI **RTR workflow** for the `schema_validator` module (which previously had no CI coverage), and along the way fixes a pre-existing IPv6 validation bug and raises the module's test coverage above the gate — both surfaced by enabling that CI.

Closes #36787

### 1. Thread-safe, idempotent factory (the fix)

- Add a `std::shared_mutex` to `SchemaValidatorFactory::Impl`.
- `initialize()`: take an **exclusive** lock and return early if already initialized (the check-then-act is now atomic and the embedded resources are loaded exactly once, never while another thread reads the map).
- `getValidator()` / `isInitialized()`: take a **shared** lock — concurrent validation across modules is still allowed; reads only block during init/reset.
- `reset()`: exclusive lock.
- `initializeFromEmbeddedResources()` documented as "caller holds the exclusive lock".

### 2. IPv6 zone/scope id validation (pre-existing bug, surfaced by the new CI)

- `isValidIP` relied on `inet_pton`, which rejects the RFC 4007 zone suffix (e.g. `fe80::1%eth0`), whereas OpenSearch's `IpFieldMapper` (`InetAddresses.ipStringToBytes`) ignores everything from the first `%` before parsing and indexes the bare address. The validator now mirrors that — it strips from the first `%` — so values the indexer accepts (link-local IPv6 in `inventory-networks`/`inventory-protocols` `ip` fields: `network.ip`, `network.gateway`, `wazuh.agent.host.ip`, …) are no longer flagged invalid. Fixes the `ValidateIPv6WithZone` test (previously only run on Linux, now exercised in CI).

### 3. RTR workflow + CI enablement

- New `.github/workflows/5_testcomponent_schema_validator-rtr.yml` (mirrors `sync_protocol-rtr`): runs `build.py --readytoreview shared_modules/schema_validator` for the **agent** and **winagent** targets (cppcheck, AStyle, unit tests, coverage, valgrind, ASAN).
- Register the module in the CI scripts so RTR can run it: `MODULE_LIST` (`ci/utils.py`), `DELETE_FOLDER_DIC` (`ci/build_tools.py`), and a `test_tool_config.json` entry.
- Add a `testtool` (`schema_validator_test_tool`) used by the RTR ASAN phase; it exercises the concurrent-init path and validates a sample message.

### 4. Restrictive handling when no validator exists for an index

Independently of the race, the consumer fallback was wrong. When `getValidator()` returned no validator for an index, `syscollector` and `sca` logged a warning and **let the message through unvalidated** (`return true` / "Queuing message without validation"). That defeats the purpose of validation: a message for an index we have no schema for could reach the indexer with a shape it does not expect.

This PR flips that fallback to be restrictive — when there is no validator for the index, the message is **discarded** (`return false`) and a warning is logged ("…Discarding message."). Validation is now fail-closed: a message is only forwarded when a validator exists and accepts it.

- `syscollector` (`syscollectorImp.cpp`, `validateSchemaAndLog`): `return true` → `return false` in the missing-validator branch; log text updated.
- `sca` (`sca_impl.cpp`, `sca_event_handler.cpp`): same fail-closed change on the missing-validator path.

### 5. Tests

- A concurrency regression test (the harness below, adapted to the repo's gtest setup).
- C-API and engine/edge-case tests to clear the 90% line-coverage gate (see *Tests Introduced*).
- Updated the syscollector/sca "missing validator" tests to assert the message is now discarded (no persist callback) instead of queued.

### Results and Evidence

A standalone harness reproduces the production scenario: 8 threads released by a barrier run the same check-then-act guard as the real modules, all hitting `initialize()` simultaneously; afterwards every schema is looked up and missing validators are counted. The factory source is compiled in two variants — current `main` (unfixed) and the working tree (fixed) — and run under **ThreadSanitizer** (deterministic race detection) and without sanitizer (functional behaviour). ASLR disabled (`setarch -R`) to avoid the unrelated TSan memory-mapping issue.

#### Before the fix — ThreadSanitizer reports the race (26 reports + 1 SEGV in a 100×8 run)

The races point exactly at two threads mutating the same red-black tree inside `loadSchemaFromString` via `initialize()`:

```
WARNING: ThreadSanitizer: data race (pid=123301)
  Read of size 8 at 0x7b1000000028 by thread T6:
    #0 std::_Rb_tree<...>::_M_get_insert_hint_unique_pos(...)
    #1 std::_Rb_tree<...>::_M_emplace_hint_unique<...>(...)
    #2 SchemaValidator::SchemaValidatorFactory::Impl::loadSchemaFromString(std::string const&)
    #3 SchemaValidator::SchemaValidatorFactory::Impl::initializeFromEmbeddedResources() schemaValidator.cpp:440
    #4 SchemaValidator::SchemaValidatorFactory::Impl::initialize(...)                    schemaValidator.cpp:463
    #5 SchemaValidator::SchemaValidatorFactory::initialize(...)                          schemaValidator.cpp:494
    #6 operator() race_test.cpp:44                       # module-style: if(!isInitialized()) initialize();
    ...

  Previous write of size 8 at 0x7b1000000028 by thread T4:
    #0 std::_Rb_tree<...>::_M_emplace_hint_unique<...>(...)
    #1 SchemaValidator::SchemaValidatorFactory::Impl::loadSchemaFromString(std::string const&)
    #2 SchemaValidator::SchemaValidatorFactory::Impl::initializeFromEmbeddedResources() schemaValidator.cpp:440
    #3 SchemaValidator::SchemaValidatorFactory::Impl::initialize(...)                    schemaValidator.cpp:463
    ...

  Location is heap block of size 56 ... allocated by main thread:
    #1 std::make_unique<SchemaValidator::SchemaValidatorFactory::Impl>()
    #2 SchemaValidator::SchemaValidatorFactory::SchemaValidatorFactory()  schemaValidator.cpp:480
    #3 SchemaValidator::SchemaValidatorFactory::getInstance()             schemaValidator.cpp:488
```

```
ORIG TSan data races: 26   SEGV: 1   (exit 66)
```

#### Before the fix — without sanitizer the corruption crashes or hangs

```
--- ORIG (functional, 2000 iters x 8 threads) ---
run1: 44302 Segmentation fault (core dumped)  ./func_orig 2000 8   (exit 139)
run2: hung (infinite loop traversing the corrupted tree) — had to be killed
```

#### After the fix — no race, no missing validators, no crash/hang

```
##### AFTER FIX: ThreadSanitizer (100 iters x 8 threads) #####
=== bad iterations: 0/100, total missing lookups: 0 ===
FIXED TSan warnings: 0   races: 0   (exit 0)

##### AFTER FIX: functional, no sanitizer (3 runs x 3000 iters x 8 threads) #####
run1: exit=0 | === bad iterations: 0/3000, total missing lookups: 0 ===
run2: exit=0 | === bad iterations: 0/3000, total missing lookups: 0 ===
run3: exit=0 | === bad iterations: 0/3000, total missing lookups: 0 ===
```

### Artifacts Affected

Source / library:
- `src/shared_modules/schema_validator/src/schemaValidator.cpp` — thread-safety + IPv6 zone fix.
- `src/shared_modules/schema_validator/src/schemaValidator_c.cpp` — `LCOV_EXCL` markers on the defensive `catch(...)` blocks.
- `src/shared_modules/schema_validator/CMakeLists.txt` — build the new `testtool`.

The library's public API is unchanged. Consumers benefit from the thread-safety transparently; in addition, the `syscollector` and `sca` missing-validator fallback is changed to fail-closed (see Proposed Changes §4):
- `src/wazuh_modules/syscollector/src/syscollectorImp.cpp` — discard instead of queue when no validator for the index.
- `src/wazuh_modules/sca/sca_impl/src/sca_impl.cpp`, `src/wazuh_modules/sca/sca_impl/src/sca_event_handler.cpp` — same fail-closed change.

CI:
- `.github/workflows/5_testcomponent_schema_validator-rtr.yml` (new).
- `src/ci/utils.py`, `src/ci/build_tools.py`, `src/ci/input/test_tool_config.json` — register the module for RTR.
- `src/shared_modules/schema_validator/testtool/{main.cpp,CMakeLists.txt}` (new).

Tests:
- `src/shared_modules/schema_validator/tests/schemaValidatorConcurrency_test.cpp` (new).
- `src/shared_modules/schema_validator/tests/schemaValidator_test.cpp` — C-API and engine/edge-case tests.
- `src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp` and `src/wazuh_modules/sca/sca_impl/tests/sca_event_handler_test.cpp` — updated for the fail-closed missing-validator behaviour.

### Configuration Changes

No runtime/agent configuration changes. CI-only: the `schema_validator` module is registered in the RTR helpers (`MODULE_LIST`, `DELETE_FOLDER_DIC`, `test_tool_config.json`) and a new workflow is added.

### Documentation Updates

None required (internal thread-safety fix; public API unchanged).

### Tests Introduced

- **Concurrency regression** (`schemaValidatorConcurrency_test.cpp`): N threads run the same check-then-act guard as the real modules against the shared singleton; asserts the factory ends up initialized and no validator reachable after a single-threaded init is ever lost. Fails (data race / lost validator) on the unfixed code, passes on the fix.
- **C-API and engine coverage** (added to `schemaValidator_test.cpp`): `schema_validator_initialize` / `is_initialized` / `validate` (null args, uninitialized, unknown index, valid, invalid with/without an error pointer, multi-error join); engine edge cases (schema without `properties`, `index_patterns` without wildcard, non-object top-level message, nested-object field with a scalar value, `initialize()` with custom validators); and the `ValidateIPv6WithZone` case now passes.
- RTR coverage gate: module line coverage **85.7% → ~96.7%** (90% threshold). The only remaining uncovered lines in `schemaValidator_c.cpp` are the unreachable `catch(...)` handlers, marked with `LCOV_EXCL`.
- All RTR checks pass for both **agent** and **winagent** targets.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
+++ b/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
@@ -4599,13 +4599,15 @@ bool Syscollector::validateSchemaAndLog(const std::string& data, const std::stri
 
     if (!validator)
     {
-        // Validator not found for this index, log warning and allow message through
+        // No validator for this index: be restrictive and discard the message
+        // instead of queuing it unvalidated, since we cannot guarantee it matches
+        // the schema the indexer expects.
         if (m_logFunction)
         {
-            m_logFunction(LOG_WARNING, "No schema validator found for index: " + index + ". Queuing message without validation.");
+            m_logFunction(LOG_WARNING, "No schema validator found for index: " + index + ". Discarding message.");
         }
 
-        return true;
+        return false;
     }
 
     auto validationResult = validator->validate(data);
--- a/.github/workflows/5_testcomponent_schema_validator-rtr.yml
+++ b/.github/workflows/5_testcomponent_schema_validator-rtr.yml
@@ -1,0 +1,50 @@
+name: 5.X - Running RTR. Module schema_validator for agent/winagent targets
+
+on:
+  workflow_dispatch:
+  pull_request:
+    types: [synchronize, opened, reopened, ready_for_review]
+    paths:
+      - "src/ci/**"
+      - "src/build.py"
+      - "src/shared_modules/schema_validator/**"
+      - ".github/workflows/5_testcomponent_schema_validator-rtr.yml"
+      - ".github/actions/**"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  rtr:
+    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
+    strategy:
+      fail-fast: false
+      matrix:
+        module: [shared_modules/schema_validator]
+        target: [agent, winagent]
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+      - name: "Install dependencies"
+        uses: ./.github/actions/install_build_deps
+        with:
+          target: ${{ matrix.target }}
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version-file: ".github/workflows/.python-version"
+          architecture: x64
+      - name: "Install a compatible CMake"
+        uses: ./.github/actions/reinstall_cmake
+      - name: Run RTR for module ${{ matrix.module }} and target ${{ matrix.target }}
+        run: |
+          cd src/
+          python build.py --target ${{ matrix.target }} --readytoreview ${{ matrix.module }}
+      - name: Uploading coverage report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage_report ${{ matrix.target }}
+          path: ./**/coverage_report/**
--- a/.github/workflows/5_testcomponent_schema_validator-rtr.yml
+++ b/.github/workflows/5_testcomponent_schema_validator-rtr.yml
@@ -24,6 +24,7 @@ jobs:
         module: [shared_modules/schema_validator]
         target: [agent, winagent]
     runs-on: ubuntu-24.04
+    timeout-minutes: 60
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3
--- a/src/ci/build_tools.py
+++ b/src/ci/build_tools.py
@@ -19,6 +19,7 @@ DELETE_FOLDER_DIC = {
     'shared_modules/dbsync':        ['build', 'smokeTests/output'],
     'shared_modules/sync_protocol': ['build', 'smokeTests/output', 'coverage_report'],
     'shared_modules/agent_metadata': ['build', 'smokeTests/output', 'coverage_report'],
+    'shared_modules/schema_validator': ['build', 'smokeTests/output', 'coverage_report'],
     'shared_modules/file_helper':   ['build'],
     'data_provider':                ['build', 'smokeTests/output'],
     'syscheckd':                    ['build', 'src/db/smokeTests/output',
--- a/src/ci/input/test_tool_config.json
+++ b/src/ci/input/test_tool_config.json
@@ -81,6 +81,14 @@
             "args": []
         }
     ],
+    "shared_modules/schema_validator":
+    [
+        {
+            "test_tool_name": "schema_validator_test_tool",
+            "is_smoke_with_configuration": false,
+            "args": []
+        }
+    ],
     "data_provider":
     [
         {
--- a/src/ci/utils.py
+++ b/src/ci/utils.py
@@ -40,6 +40,7 @@ HEADER_DIR = {
 }
 MODULE_LIST = ['wazuh_modules/syscollector', 'shared_modules/dbsync',
                'shared_modules/sync_protocol', 'shared_modules/agent_metadata',
+               'shared_modules/schema_validator',
                'shared_modules/file_helper', 'data_provider', 'syscheckd', 'wazuh_modules/sca', 'wazuh_modules/agent_info']
 MODULE_LIST_STR = '|'.join(MODULE_LIST)
 TARGET_LIST = ['agent', 'manager', 'winagent']
--- a/src/shared_modules/schema_validator/CMakeLists.txt
+++ b/src/shared_modules/schema_validator/CMakeLists.txt
@@ -146,3 +146,5 @@ if(UNIT_TEST)
   add_coverage_libraries(schema_validator)
   add_subdirectory(tests)
 endif()
+
+add_subdirectory(testtool)
--- a/src/shared_modules/schema_validator/include/schemaValidator_c.h
+++ b/src/shared_modules/schema_validator/include/schemaValidator_c.h
@@ -46,8 +46,8 @@ EXPORTED bool schema_validator_is_initialized(void);
  * @return true if message is valid, false if invalid
  */
 EXPORTED bool schema_validator_validate(const char* indexPattern,
-                                          const char* message,
-                                          char** errorMessage);
+                                        const char* message,
+                                        char** errorMessage);
 
 #ifdef __cplusplus
 }
--- a/src/shared_modules/schema_validator/src/schemaValidator.cpp
+++ b/src/shared_modules/schema_validator/src/schemaValidator.cpp
@@ -2,6 +2,8 @@
 #include "schemaResources.hpp"
 #include <sstream>
 #include <map>
+#include <mutex>
+#include <shared_mutex>
 #include <regex>
 #include <ctime>
 #include <iomanip>
@@ -397,6 +399,12 @@ namespace SchemaValidator
         public:
             std::map<std::string, std::shared_ptr<ISchemaValidatorEngine>> m_validators;
             bool m_initialized;
+            // The factory is a process-wide singleton (see getInstance) that is
+            // initialized from several modules (syscollector, sca)
+            // running concurrently in the same process. Guard all access to
+            // m_validators/m_initialized so concurrent initialize() calls cannot race
+            // on the underlying std::map. Exclusive lock for mutation, shared for reads.
+            mutable std::shared_mutex m_mutex;
 
             Impl()
                 : m_initialized(false)
@@ -429,6 +437,7 @@ namespace SchemaValidator
                 }
             }
 
+            // Assumes the caller already holds an exclusive lock on m_mutex.
             bool initializeFromEmbeddedResources()
             {
                 try
@@ -451,6 +460,16 @@ namespace SchemaValidator
 
             bool initialize(std::map<std::string, std::shared_ptr<ISchemaValidatorEngine>> customValidators)
             {
+                std::unique_lock<std::shared_mutex> lock(m_mutex);
+
+                // Idempotent: the singleton is shared across modules, each of which calls
+                // initialize() at startup. Once loaded, return success without re-loading
+                // (and without mutating the map while another thread may be reading it).
+                if (m_initialized)
+                {
+                    return true;
+                }
+
                 if (!customValidators.empty())
                 {
                     // Use injected custom validators (for testing or extensibility)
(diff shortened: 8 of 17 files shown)


src/wazuh_modules/syscollector/src/syscollectorImp.cpp
```diff
@@ -4599,13 +4599,15 @@ bool Syscollector::validateSchemaAndLog(const std::string& data, const std::stri
...
+        // instead of queuing it unvalidated, since we cannot guarantee it matches
+        // the schema the indexer expects.
         if (m_logFunction)
         {
-            m_logFunction(LOG_WARNING, "No schema validator found for index: " + index + ". Queuing message without validation.");
+            m_logFunction(LOG_WARNING, "No schema validator found for index: " + index + ". Discarding message.");
         }
 
-        return true;
+        return false;
```
jr0me, line 4610: This changes behavior, though it clearly looks intentional. Perhaps it should be described in the PR description.

jr0me (approved):

src/wazuh_modules/syscollector/src/syscollectorImp.cpp
Nicogp: Yes, the behavior change was intentional. As you suggested, I added a comment explaining it in the PR description.

PR #36797, opened 2026-06-09
fix off-by-one buffer overflow in postgresql and mssql log readers

aizu-m:
tighten the length check in the multiline branch of read_postgresql_log and read_mssql_log so the separator space and the trailing null are counted, otherwise a tab-indented continuation line whose length exactly fills the space left in buffer[OS_MAX_LOG_SIZE] makes strncat write one null byte past the end.


aizu-m:
any update?

PR #36859, opened 2026-06-12
Skip vanished /proc entries during ports scan

### LINKED ISSUE #36803: Validate log severity for anomaly messages
| Wazuh version | Component | Install type | Install method | Platform |
|---|---|---|---|---|
| 5.0 | Wazuh component | Agent | Packages/Sources | All |

## Description

This issue documents the log messages reviewed during the severity normalization work that should keep their current `WARNING` or `ERROR` level.

These messages were intentionally not downgraded because they do not represent purely informational, expected, or harmless conditions. They may appear in common non-blocking scenarios, and in some cases the agent can retry or recover, but they still describe real operational, connectivity, security, capacity, or data-integrity anomalies that may require administrator attention.

The criterion used for this review is:

- A log can be lowered to `INFO` only when it reports an expected, recoverable, and purely informational condition, such as a user-provided deprecated or no-longer-supported configuration option.
- A log should remain as `WARNING` when it reports a real operational, security, data-integrity, capacity, or connectivity anomaly that may require administrator attention, even if the component can continue, retry, or recover.
- A log should remain as `ERROR` when it reports a failure of a core function, such as the agent being unable to establish communication with the manager.

The case discussed in `https://github.com/wazuh/wazuh/issues/36803#issuecomment-4672138496` is intentionally out of scope for this issue. It was moved to a dedicated issue and is being handled separately.

This issue only keeps the messages that are currently understood as valid `WARNING` or `ERROR` conditions and should not be downgraded as part of the severity normalization work.

### 1. No schema validator found for index: `wazuh-states-inventory-hardware`. Queuing message without validation.

```text
wazuh-modulesd:syscollector: WARNING: No schema validator found for index: wazuh-states-inventory-hardware. Queuing message without validation.
```

Source:

```text
src/wazuh_modules/syscollector/src/syscollectorImp.cpp:4605
LOG_WARNING
```

This message should remain as `WARNING` because the message is being queued without schema validation.

Although the component continues processing, this is not a purely informational condition. A missing validator may indicate a version mismatch, missing schema, component misconfiguration, or an unexpected inventory/index state. Since data is sent without validation, downstream data integrity may be affected and the administrator should be aware of it.

Note: if the output shows something like `No scheWARNINGma`, it is only the result of two log lines interleaving in the console output. It is not a separate corrupted message.

### 2. Agent buffer at XX%.

```text
wazuh-agentd: WARNING: Agent buffer at 90 %.
```

Source:

```text
src/client-agent/src/buffer.c:243
WARN_BUFFER ("Agent buffer at %d %%.")

src/shared/include/error_messages/error_messages.h:549
```

This message should remain as `WARNING` because it reports capacity pressure in the agent anti-flooding buffer.

The agent may still be running and communication may continue, but a high buffer usage means events are close to being dropped if the condition persists. This is actionable and may indicate back-pressure, network issues, manager overload, or an event burst. It is therefore not a purely informational message.

### 3. `(1404): Authentication error. Wrong key or corrupt payload. Message received from agent ...`.

```text
2026/07/01 08:32:22 wazuh-agentd: WARNING: (1404): Authentication error. Wrong key or corrupt payload. Message received from agent '002' at 'any'.
2026/07/01 08:32:22 wazuh-agentd: WARNING: (1214): Problem receiving message from '192.168.2.106'.
2026/07/01 08:32:22 wazuh-agentd: WARNING: (1404): Authentication error. Wrong key or corrupt payload. Message received from agent '002' at 'any'.
2026/07/01 08:32:22 wazuh-agentd: WARNING: (1214): Problem receiving message from '192.168.2.106'.
2026/07/01 08:32:22 wazuh-agentd: WARNING: (1404): Authentication error. Wrong key or corrupt payload. Message received from agent '002' at 'any'.
2026/07/01 08:32:22 wazuh-agentd: WARNING: (1214): Problem receiving message from '192.168.2.106'.
```

Source:

```text
src/shared/os_crypto/shared/msgs.c:347
src/shared/os_crypto/shared/msgs.c:355
src/shared/os_crypto/shared/msgs.c:568
ENCKEY_ERROR

src/shared/include/error_messages/error_messages.h:241
```

This message should remain as `WARNING` because it is security-relevant.

The received message could not be decrypted or authenticated. Possible causes include a wrong key, a rotated key, payload corruption, agent/manager mismatch, or potentially message tampering/spoofing. Even if this does not necessarily block the whole agent immediately, lowering it to `INFO` would hide a condition that may require investigation.

### 4. `(1214): Problem receiving message from ...`.

```text
wazuh-agentd: WARNING: (1214): Problem receiving message from 'ec2-3-236-145-90.compute-1.amazonaws.com'.
wazuh-agentd: WARNING: (1214): Problem receiving message from 'ec2-98-92-19-126.compute-1.amazonaws.com'.
```

Source:

```text
src/client-agent/src/start_agent.c:716
src/client-agent/src/receiver.c:88
MSG_ERROR

src/shared/include/error_messages/error_messages.h:77
```

This message should remain as `WARNING` because it reports a real connectivity or message reception problem.

The agent may retry and recover, so this is not necessarily a fatal condition. However, it still indicates that message reception failed due to a network interruption, timeout, closed connection, or similar communication issue. This is operationally relevant and should remain visible to administrators.

### 5. `(1216): Unable to connect to 'IP/tcp': 'No connection could be made because the target machine actively refused it.'`.

```text
wazuh-agent: ERROR: (1216): Unable to connect to '[172.31.68.211]:1514/tcp': 'No connection could be made because the target machine actively refused it.'.
wazuh-agent: ERROR: (1216): Unable to connect to '[172.31.67.72]:1514/tcp': 'No connection could be made because the target machine actively refused it.'.
```

Source:

```text
src/client-agent/src/start_agent.c:391
OS_ConnectTCP()
```

This message should remain as `ERROR` because the agent failed to establish its core communication channel with the manager.

The underlying cause, TCP `ECONNREFUSED`, usually means the manager service is down, the configured port is wrong or closed, or a firewall/security group is rejecting the connection. This is not a Wazuh product bug by itself, but it is a real deployment or runtime failure that requires administrator action.

While this condition persists, the agent cannot deliver events to the manager, so the endpoint may be effectively unmonitored. For that reason, `ERROR` is the appropriate severity.

## Out of scope

The case discussed in `https://github.com/wazuh/wazuh/issues/36803#issuecomment-4672138496` was moved to a dedicated issue and is not tracked here.

This issue should not be used to decide or block the severity of that moved case. Only the messages listed above are considered part of this validation.

## Expected outcome

- The messages listed above remain with their current severity.
- These messages are documented as expected operational `WARNING` or `ERROR` conditions, not as product bugs.
- These messages should not block the severity normalization PR unless new evidence shows that a specific message is emitted in a truly expected and purely informational scenario.
- The moved case remains handled in its dedicated issue.

Nicogp:
## Description

On Linux agents, syscollector occasionally logged:

```
wazuh-modulesd:syscollector: ERROR: filesystem error: status: No such process [/proc/<pid>/fd]
```

This is a **TOCTOU race** in the ports collector. `SysInfo::portProcessInfo`
(`src/data_provider/src/sysInfoLinux.cpp`) maps socket inodes → PID by scanning `/proc/<pid>/fd`. It
lists `/proc`, then for each PID checks `fs.is_directory(/proc/<pid>)` and `fs.is_directory(/proc/<pid>/fd)`.
`file_system::FileSystemWrapper::is_directory` calls the **throwing** overload of
`std::filesystem::is_directory` (no `std::error_code`). When a process **exits between the `/proc`
listing and these checks**, accessing its
`/proc/<pid>` entry can return **`ESRCH` ("No such process")**, and `is_directory` throws a
`std::filesystem::filesystem_error`.

Those two per‑PID `is_directory` calls were **outside any try/catch** (unlike the `list_directory` and
`is_socket` calls in the same loop, which already catch and continue). So the exception propagated out
of `portProcessInfo` → `getPorts` → `scanPorts`, where the `TRY_CATCH_TASK` macro
(`syscollectorImp.cpp:42‑56`) caught it and logged the raw `what()` at `LOG_ERROR`. Net effect: **a
single unrelated process that legitimately exited mid‑scan aborted the entire ports inventory for that
cycle and logged a misleading ERROR.**

### Fix

Guard the per‑PID existence checks with a small local helper that treats a vanished/inaccessible
entry as "not a directory", so that PID is simply skipped and the scan continues:

```cpp
auto safeIsDirectory = [&fs](const std::filesystem::path & path) -> bool
{
    try
    {
        return fs.is_directory(path);
    }
    catch (const std::filesystem::filesystem_error&)
    {
        return false;
    }
};
...
if (Utils::isNumber(procFileName) && safeIsDirectory(procFile))      // /proc/<pid>
{
    const std::filesystem::path pidFilePath = procFile / "fd";
    if (safeIsDirectory(pidFilePath))                               // /proc/<pid>/fd  (reported site)
    { ... }
}
```

The top‑level `if (fs.is_directory(procPath))` (where `procPath == "/proc/"`) is **left unguarded on
purpose**: `/proc` is the stable kernel mount root, it does not take part in the per‑process vanish
race.

Closes #36803 (the `filesystem error: status: No such process [/proc/<pid>/fd]` part)

## Proposed Changes

- `src/data_provider/src/sysInfoLinux.cpp` — add the `safeIsDirectory` helper in `portProcessInfo` and
  use it for the `/proc/<pid>` and `/proc/<pid>/fd` checks. Localized change; does not alter
  `FileSystemWrapper::is_directory` semantics for any other caller.
- **CI (`detect-changes`)** — syscollector consumes the `data_provider` (sysinfo) library, but its
  integration-test trigger did not include it, so this fix's change to `src/data_provider/**` matched no
  module and the syscollector ITs were skipped. Added `src/data_provider/**` to the `syscollector`
  module paths in `.github/test_modules_linux.json` and `.github/test_modules_windows.json`. Also
  renamed `.github/test-modules-windows.json` → `.github/test_modules_windows.json` for consistency with
  the other platform files (updated its self-references and the references in
  `.github/workflows/5_builderpackage_agent-windows.yml`).

### Results and Evidence

Verified with a deterministic before/after that exercises the real `FileSystemWrapper::is_directory`
and replicates `portProcessInfo`'s per‑PID loop, against a `/proc/<pid>/fd` entry made to throw a
`filesystem_error`:

```
###### WITHOUT FIX ######
[BUGGY] scan ABORTED (exception escaped, whole ports scan lost): filesystem error: status: ... [/proc/<pid>/fd]

###### WITH FIX ######
[FIXED] scan COMPLETED, PIDs processed = 1
```

- **Without the fix:** the exception from `is_directory(/proc/<pid>/fd)` escapes and aborts the whole
  ports scan (in production it is then logged as the reported ERROR and the cycle's ports inventory is
  lost).
- **With the fix:** the offending PID is skipped and the scan completes normally.

The fix catches any `std::filesystem::filesystem_error`, so it handles the field's `ESRCH` (process
vanished mid‑scan) and any other errno identically.

### Artifacts Affected

- `src/data_provider/src/sysInfoLinux.cpp`
- `.github/test_modules_linux.json`
- `.github/test_modules_windows.json` (renamed from `.github/test-modules-windows.json`)
- `.github/workflows/5_builderpackage_agent-windows.yml`

No behavior change other than: a per‑PID `/proc` entry that vanishes or is momentarily inaccessible
during the scan is now skipped (the ports scan completes for all other PIDs) instead of aborting the
scan and logging a misleading ERROR. The CI changes only affect which integration tests are selected.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

None

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/.github/test_modules_linux.json
+++ b/.github/test_modules_linux.json
@@ -305,6 +305,7 @@
         "src/config/src/wmodules-syscollector.c",
         "src/wazuh_modules/src/wm_syscollector.c",
         "src/wazuh_modules/syscollector/**",
+        "src/data_provider/**",
         "src/Makefile",
         "tests/integration/conftest.py",
         "tests/integration/test_syscollector/**"
--- a/.github/test_modules_windows.json
+++ b/.github/test_modules_windows.json
@@ -6,7 +6,7 @@
       "tiers": ["0 1"],
       "paths": [
         ".github/workflows/5_testintegration_windows-integration-tests.yml",
-        ".github/test-modules-windows.json",
+        ".github/test_modules_windows.json",
         "src/client-agent/**",
         "src/config/src/buffer-config.c",
         "src/config/src/client-config.c",
@@ -23,7 +23,7 @@
       "tiers": ["0 1"],
       "paths": [
         ".github/workflows/5_testintegration_windows-integration-tests.yml",
-        ".github/test-modules-windows.json",
+        ".github/test_modules_windows.json",
         "src/client-agent/**",
         "src/config/src/buffer-config.c",
         "src/config/src/client-config.c",
@@ -42,7 +42,7 @@
       "tiers": ["0 1"],
       "paths": [
         ".github/workflows/5_testintegration_windows-integration-tests.yml",
-        ".github/test-modules-windows.json",
+        ".github/test_modules_windows.json",
         "src/active-response/**",
         "src/os_execd/**",
         "src/win32/**",
@@ -59,7 +59,7 @@
       "start_service": true,
       "paths": [
         ".github/workflows/5_testintegration_windows-integration-tests.yml",
-        ".github/test-modules-windows.json",
+        ".github/test_modules_windows.json",
         "src/config/src/syscheck-config.c",
         "src/config/include/syscheck-config.h",
         "src/syscheckd/**",
@@ -74,7 +74,7 @@
       "tiers": ["0 1"],
       "paths": [
         ".github/workflows/5_testintegration_windows-integration-tests.yml",
-        ".github/test-modules-windows.json",
+        ".github/test_modules_windows.json",
         "src/config/src/wmodules-github.c",
         "src/wazuh_modules/src/wm_github.h",
         "src/wazuh_modules/src/wm_github.c",
@@ -90,7 +90,7 @@
       "tiers": ["0 1"],
       "paths": [
         ".github/workflows/5_testintegration_windows-integration-tests.yml",
-        ".github/test-modules-windows.json",
+        ".github/test_modules_windows.json",
         "src/config/src/localfile-config.c",
         "src/config/include/localfile-config.h",
         "src/logcollector/**",
@@ -105,7 +105,7 @@
       "tiers": ["0 1"],
       "paths": [
         ".github/workflows/5_testintegration_windows-integration-tests.yml",
-        ".github/test-modules-windows.json",
+        ".github/test_modules_windows.json",
         "src/config/src/wmodules-office365.c",
         "src/wazuh_modules/src/wm_office365.h",
         "src/wazuh_modules/src/wm_office365.c",
@@ -122,7 +122,7 @@
       "tiers": ["0 1"],
       "paths": [
         ".github/workflows/5_testintegration_windows-integration-tests.yml",
-        ".github/test-modules-windows.json",
+        ".github/test_modules_windows.json",
         "src/config/src/wmodules-sca.c",
         "src/wazuh_modules/src/wm_sca.c",
         "src/wazuh_modules/src/wm_sca.h",
@@ -139,10 +139,11 @@
       "tiers": ["0 1"],
       "paths": [
         ".github/workflows/5_testintegration_windows-integration-tests.yml",
-        ".github/test-modules-windows.json",
+        ".github/test_modules_windows.json",
         "src/config/src/wmodules-syscollector.c",
         "src/wazuh_modules/src/wm_syscollector.c",
         "src/wazuh_modules/syscollector/**",
+        "src/data_provider/**",
         "src/win32/**",
         "src/Makefile",
         "tests/integration/conftest.py",
--- a/.github/workflows/5_builderpackage_agent-windows.yml
+++ b/.github/workflows/5_builderpackage_agent-windows.yml
@@ -10,8 +10,8 @@ on:
       - "ruleset/**"
       - ".github/actions/check_files/**"
       - ".github/workflows/5_builderpackage_agent-windows.yml"
-      - ".github/test-modules-windows.json"
       - ".github/workflows/5_testintegration_agent-start.yml"
+      - ".github/test_modules_windows.json"
       - ".github/workflows/5_testintegration_windows-integration-tests.yml"
       - ".github/actions/5_builderpackage_get-last-stable-version/**"
       - ".github/actions/5_builderpackage_get-upgrade-test-package-url/**"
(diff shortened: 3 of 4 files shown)


jpcerrone (dismissed):

jpcerrone (approved):

jr0me (approved):

Nicogp:
The failing checks are not related to this PR.

- The SCA failures (macOS UTs) are caused by changes to the SCA index. 
  - https://github.com/wazuh/wazuh-indexer-plugins/issues/1301 
  - https://github.com/wazuh/wazuh/issues/37037
- The MSI test failure has been reported previously: https://github.com/wazuh/wazuh/issues/37132

PR #37036, opened 2026-06-22
add eBPF field validation CI

### LINKED ISSUE #32502: FIM eBPF provider: Automated testing of expected syscalls for multiple platforms
# Description

Following the inclusion of the new FIM Whodata provider, eBPF, we have encountered multiple problems in validating the code, hooks used, necessary triggers, syscalls, etc.
This is because each system has different qualities that require validation of the use of eBPF.
We want to ensure that every modification, update, or new feature related to the eBPF code is tested:
- FIM C++ code that reads the events generated by eBPF and generates the corresponding alerts: `ebpf_whodata.cpp`
- C code that generates the bpf object, loads all the functions, and implements the hooks that we will use in the system: `modern.bpf.c`

## Test battery
To ensure full functionality, we must test different syscalls, which must generate the corresponding add/modify/delete alert (including all expected fields, similar to those of the audit provider), and we must check this on multiple platforms, taking into account different kernel versions, architectures, and file systems. For example, based on the errors and problems encountered, an initial idea would be:

- `Kernel 5.8` (minimal required version).
- `Kernel 6.8` (some new features appeared).
- `amd64`/`arm64` architectures (investigate if needed include testing for others).
- `ext4`/`xfs`/`btrfs`/`tmpfs` (investigate other possible specific changes)

## Possible ideas
It is possible to run the tests exclusively with modern.bpf.o, as it is responsible for including the necessary hooks in the system and generating the ring buffer where all events will go.
We can create a script/code file that loads the program and the ring buffer and checks the different syscalls/commands that we expect to trigger events, verifying that all fields are set correctly.

Miguevrgo:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/32502|

## Description

Adds an automated eBPF field-validation workflow for the FIM whodata eBPF provider (`4_testcomponent_ebpf-validation.yml` + `ebpf_validation` composite action).

The workflow:

- Builds the `wazuh-manager` deb package (amd64) using the reusable `4_builderpackage_manager.yml` workflow.
- Installs the generated package on an `ubuntu-24.04` runner.
- Compiles a standalone validation runner (`src/syscheckd/src/ebpf/tests/ebpf_validation/ebpf_validation_test.cpp`) that loads the packaged `/var/ossec/lib/modern.bpf.o` through libbpf, mirroring the program-selection logic of `ebpf_whodata.cpp` (LSM vs kprobe hooks, `_dpath` vs `_walk` fallback).
- Triggers the expected syscalls (`create`, `chmod`, `chown`, `truncate`, `unlink`) on `ext4`, `xfs`, `btrfs` and `tmpfs` mounts, validating every event field (path, PID, PPID, UID, GID, comm, parent name, CWD, device, inode).
- Publishes a per-filesystem result table (passed/failed/warning counts) to the workflow step summary and fails if any filesystem fails.

The workflow triggers on pull requests touching `src/syscheckd/src/ebpf/**`, the workflow/action files, or `src/Makefile`, and can be run manually via `workflow_dispatch`.

## Configuration options

None.

## Logs/Alerts example

Workflow step summary:

| Filesystem | Result | Passed | Failed | Warnings |
|---|---|---|---|---|
| ext4 | ✅ PASS | 47 | 0 | 4 |
| xfs | ✅ PASS | 47 | 0 | 4 |
| btrfs | ✅ PASS | 42 | 0 | 9 |
| tmpfs | ✅ PASS | 47 | 0 | 4 |

## Tests

- [x] Package installation (performed by the workflow itself)
- [x] eBPF validation workflow run passing on this PR

### DIFF
--- a/.github/actions/ebpf_validation/action.yml
+++ b/.github/actions/ebpf_validation/action.yml
@@ -1,0 +1,97 @@
+name: "eBPF Field Validation"
+description: "Validates eBPF hook events across filesystems using installed wazuh-manager package"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install dependencies
+      run: |
+        sudo apt-get -qq update
+        sudo apt-get -qq install -y libbpf-dev btrfs-progs xfsprogs kmod
+      shell: bash
+
+    - name: Compile validation runner
+      run: |
+        g++ -O2 -Wall -std=c++17 \
+          src/syscheckd/src/ebpf/tests/ebpf_validation/ebpf_validation_test.cpp \
+          -lbpf -lelf -lz \
+          -o ebpf_validation_runner
+      shell: bash
+
+    - name: Run eBPF validation across filesystems
+      run: |
+        BPF_OBJ=/var/ossec/lib/modern.bpf.o
+        if ! sudo test -f "$BPF_OBJ"; then
+          echo "[ERROR] BPF object not found at $BPF_OBJ"
+          exit 1
+        fi
+        FAILED=""
+        SUMMARY="| Filesystem | Result | Passed | Failed | Warnings |\n|---|---|---|---|---|\n"
+
+        for FS in ext4 xfs btrfs tmpfs; do
+          IMG="/tmp/ebpf_test_${FS}.img"
+          MNT="/mnt/ebpf_test_${FS}"
+
+          echo "=========================================="
+          echo " Testing filesystem: ${FS}"
+          echo "=========================================="
+
+          if [ "${FS}" != "tmpfs" ] && ! grep -q "${FS}" /proc/filesystems; then
+            sudo modprobe "${FS}" 2>/dev/null || true
+            if ! grep -q "${FS}" /proc/filesystems; then
+              echo "[WARNING] Kernel does not support ${FS}. Skipping."
+              SUMMARY="${SUMMARY}| ${FS} | ⚠️ SKIPPED (unsupported) | - | - | - |\n"
+              continue
+            fi
+          fi
+
+          sudo rm -rf "${MNT}" && sudo mkdir -p "${MNT}"
+
+          if [ "${FS}" = "tmpfs" ]; then
+            if ! sudo mount -t tmpfs tmpfs "${MNT}"; then
+              SUMMARY="${SUMMARY}| ${FS} | ⚠️ SKIPPED (setup failed) | - | - | - |\n"
+              continue
+            fi
+          else
+            # xfs requires a filesystem of at least 300MB since xfsprogs 5.19
+            dd if=/dev/zero of="${IMG}" bs=1M count=512 conv=sparse 2>/dev/null
+            FORCE_FLAG="-f"
+            [ "${FS}" = "ext4" ] && FORCE_FLAG="-F"
+            if ! sudo mkfs.${FS} ${FORCE_FLAG} -q "${IMG}" || ! sudo mount -o loop "${IMG}" "${MNT}"; then
+              echo "[WARNING] Could not set up ${FS}. Skipping."
+              SUMMARY="${SUMMARY}| ${FS} | ⚠️ SKIPPED (setup failed) | - | - | - |\n"
+              sudo rm -rf "${MNT}" "${IMG}"
+              continue
+            fi
+          fi
+
+          RC=0
+          OUT=$(sudo ./ebpf_validation_runner --test-dir "${MNT}" 2>&1) || RC=$?
+          echo "${OUT}"
+
+          # Per-filesystem counters from the runner's "Total:" line
+          read -r P F W <<< "$(echo "${OUT}" | sed -n 's/^Total: \([0-9]*\) passed, \([0-9]*\) failed, \([0-9]*\) warnings.*/\1 \2 \3/p')" || true
+
+          if [ "${RC}" -eq 0 ]; then
+            SUMMARY="${SUMMARY}| ${FS} | ✅ PASS | ${P:-0} | ${F:-0} | ${W:-0} |\n"
+          else
+            SUMMARY="${SUMMARY}| ${FS} | ❌ FAIL | ${P:-0} | ${F:-0} | ${W:-0} |\n"
+            FAILED="${FAILED} ${FS}"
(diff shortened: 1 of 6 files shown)


Darioortegaleyva (changes requested):
Good overall. Two things before merge:

**1. UID/GID checks are false positives.** The test requires root (`line 231`) and runs via `sudo` in CI, so uid=gid=0, making the swap invisible. There's a real bug in `modern.bpf.c:321-322`: `bpf_get_current_uid_gid()` returns `GID << 32 | UID` but `evt->uid = uid_gid >> 32` gets the GID, and `evt->gid = uid_gid` gets the UID. No compensation anywhere in `ebpf_whodata.cpp`. 

**2. Struct ABI conflict with PR #37060.** That PR inserts `euid` and `login_uid`  between `gid` and `inode` in `struct file_event`. The test struct here goes straight from `gid` to `inode`, so if #37060 merges first, every field after `gid` (`inode`, `dev`, `comm`, `filename`, `cwd`...) reads 8 bytes off. 

CI failures are unrelated (CodeBuild infra, syscollector timeout).

Miguevrgo:
Thanks for the review @Darioortegaleyva. Both points addressed in the latest push:

**1. UID/GID false positives**

Fixed on both sides:
- **Test side** (`1a07b251c5`): the trigger child now drops privileges with `setgroups(0) + setgid(4321) + setuid(1234)`, so uid != gid and a swap can no longer pass silently. The test directory is chmod'ed to 0777 before forking so the non-root child can create files on the loop mounts.
- **Product side** (`ba0440f1b5`): you were right about the bug. `bpf_get_current_uid_gid()` returns `gid << 32 | uid` (verified against `bpf_helper_defs.h`), so `evt->uid = uid_gid >> 32` was storing the GID in uid and vice versa. Fixed in `modern.bpf.c` and `modern-arm.bpf.c` (same bug in both). No consumer-side changes needed: `ebpf_whodata.cpp` maps `uid -> user_id` and `gid -> group_id` directly.

Verified locally in a privileged container (ubuntu:24.04, ext4/xfs/btrfs/tmpfs): with the old object the runner fails with `uid: expected 1234 got 4321`, with the fixed object all four filesystems pass (47/47 checks, btrfs with its known dev warnings).

**2. Struct ABI conflict with #37060**

I checked whether the runner could validate the layout against the object's BTF, but `file_event` is a function-local type and does not appear in the `.BTF` section of `modern.bpf.o` (verified by compiling and dumping it), so exact layout introspection is not possible. Instead, `on_event` now requires `sz == sizeof(file_event)` and fails with a descriptive message otherwise:

```
[FAIL] Event size mismatch: BPF object sends 12384 bytes, test expects 12392. Update struct file_event to match modern.bpf.c
```

Since #37060 inserts `euid`/`login_uid` mid-struct, the size changes, so when the merge-up to 4.14.8 happens this workflow will fail with that clear message instead of reading misaligned fields. Verified with a deliberately stale struct.

**Extra finding while testing:** on recent kernels (7.1 with BPF LSM active), `lsm/path_unlink` with `bpf_d_path` also reports the path relative to the filesystem root on non-root mounts, and in LSM mode the chmod/chown/truncate events still come from `kprobe/security_inode_setattr` (it is not excluded by the program-selection logic), which has the same fs-relative limitation. The fs-relative path case is now handled as a known-limitation warning regardless of hook mode, so the suite passes in both LSM and kprobe mode. This may deserve its own issue since FIM would miss alerts on non-root mounts in that configuration.

Miguevrgo:
Follow-up after the last CI run: the new ABI guard failed on all filesystems with

```
[FAIL] Event size mismatch: BPF object sends 12392 bytes, test expects 12384.
```

I dug into it and this is not a test issue, it caught real drift in the shipped artifact:

**Root cause**

The package build never compiles `modern.bpf.c` from this repository. `make deps` downloads the precompiled bundle `https://packages.wazuh.com/deps/54/libraries/linux/amd64/libbpf-bootstrap.tar.gz`, which ships a ready-made `build/modern.bpf.o` (the Makefile target is already satisfied, so the source fallback is skipped; the build log shows no clang/bpf step at all). That prebuilt object is already compiled from **4.14.8 sources**: it contains `euid` and `login_uid` (12392 bytes, the layout from #37060).

Two consequences:

1. **The ABI drift is already in built packages, not just a future risk.** 4.14.7's `bpf_helpers.h` has the 12384-byte `file_event`, so `handle_event` in `ebpf_whodata.cpp` reads events from the shipped 12392-byte object through a smaller struct: every field after `gid` (inode, dev, comm, filename, cwd...) is parsed 8 bytes off. eBPF whodata on 4.14.7 packages built with the current bundle is effectively broken. This probably deserves its own issue (either the bundle needs to be rebuilt from 4.14.7 sources, or #37060's changes need to reach 4.14.7).
2. **The shipped object also has the uid/gid swap** (4.14.8 still does `evt->uid = uid_gid >> 32`). With the hardened test (distinct non-root credentials) the runner now reports it against the current bundle object:
   ```
   [FAIL] UID (uid: expected 1234 got 4321)
   [FAIL] GID (gid: expected 4321 got 1234)
   [PASS] EUID / [PASS] login_UID
   ```
   The swap fix in this branch (`ba0440f1b5`) corrects the 4.14.7 source, but the same fix needs to land in 4.14.8 and the deps bundle must be regenerated for packages to pick it up. Happy to open that PR if you agree.

**Test update pushed (`54c2ead41d`)**

The runner now accepts both known layouts (v1: 12384, v2: 12392 with `euid`/`login_uid`), fails loudly on any other event size, and validates the v2-only fields (`euid` against the trigger credentials, `login_uid` against the `loginuid` inherited from the test process). Verified locally in a privileged ubuntu:24.04 container:

- ext4/xfs/btrfs/tmpfs all pass with a fixed v1 object (47 checks each, btrfs with its known dev warnings).
- Passes with a fixed v2 object built from 4.14.8 sources + the swap fix (57 checks, including EUID/login_UID).
- Correctly fails on uid/gid with the currently shipped bundle object (the swap above).

So expect this workflow to stay red on uid/gid until the bundle is regenerated from fixed sources; that red is the detection working as intended, not a test defect.

PR #37280, opened 2026-06-29
Defer cluster payload buffer allocation until data is received

jepalfer:
## Description

The cluster TCP listener on port 1516 allocated a payload buffer of the full peer-declared size immediately after parsing the 20-byte header, before any Fernet decryption or authentication. With the only guard being an upper cap of 256 MiB, a single unauthenticated connection could pin ~256 MiB of master process memory for as long as the socket stayed open. Three connections were sufficient to consume ~768 MiB, confirmed by RSS measurement. This PR fixes the allocation strategy and adds two complementary limits to bound both per-IP connection count and payload-completion time.

## Proposed Changes

- Changed `InBuffer.__init__` and `get_info_from_header` in `framework/wazuh/core/cluster/common.py` to initialize `self.payload` as an empty `bytearray()` instead of `bytearray(total)`, growing it incrementally via `.extend()` as bytes arrive.
- Added `PRE_AUTH_PAYLOAD_TIMEOUT` = 30 constant and `_start_payload_deadline` / `_cancel_payload_deadline` / `_close_on_payload_deadline` methods to Handler in `common.py` to close connections that send a header but never deliver the declared payload within 30 seconds.
- Added `MAX_CONNECTIONS_PER_IP` = 10 constant and `connection_counts` dict to `AbstractServer` in `framework/wazuh/core/cluster/server.py`, enforced in `AbstractServerHandler.connection_made` to reject excess connections from a single source IP before any buffer is allocated.
- Added `_cancel_payload_deadline()` call to `client.py` `connection_lost` to ensure the deadline timer is cleaned up on the client side.

  - [x] Linux

### Tests Introduced

- `framework/wazuh/core/cluster/tests/test_common.py::test_inbuffer_lazy_allocation`: Verifies that `get_info_from_header` does not pre-allocate the full declared payload size, and that the buffer grows exactly as bytes are received.
- `framework/wazuh/core/cluster/tests/test_common.py::test_handler_payload_deadline_fires`: Verifies that `transport.close()` is called when the payload-completion deadline expires without a full payload being received.
- `framework/wazuh/core/cluster/tests/test_common.py::test_handler_payload_deadline_cancelled_on_complete`: Verifies that the deadline is cancelled once the full payload arrives.
- Extended `framework/wazuh/core/cluster/tests/test_server.py::test_AbstractServerHandler_connection_made` to cover per-IP connection count tracking and rejection when the cap is exceeded.
- Extended `framework/wazuh/core/cluster/tests/test_server.py::test_AbstractServerHandler_connection_lost` to cover correct decrement of connection_counts on disconnect.
- Extended `framework/wazuh/core/cluster/tests/test_server.py::test_AbstractServer_init` to verify connection_counts is initialized as an empty dict.

## Credit

Kudos to @moltenbit for reporting this bug.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/.github/workflows/4_testunit_framework.yml
+++ b/.github/workflows/4_testunit_framework.yml
@@ -19,7 +19,7 @@ on:
       - 'src/Makefile'
 
 jobs:
-  build: 
+  build:
     runs-on: codebuild-github-actions-codebuild-runner-server-amd-${{ github.run_id }}-${{ github.run_attempt }}
     strategy:
       fail-fast: false
@@ -45,9 +45,9 @@ jobs:
 
       - name: Run Framework tests
         run: |
-          set -o pipefail 
+          set -o pipefail
           python -m pytest framework -vv --ignore=tests --cov=framework | tee ${TEST_REPORT_PATH}
-      
+
       - name: Generate test report
         if: inputs.generate_report
         run: |
--- a/framework/wazuh/core/cluster/client.py
+++ b/framework/wazuh/core/cluster/client.py
@@ -198,6 +198,7 @@ class AbstractClient(common.Handler):
             'None' means a regular EOF is received, or the connection was aborted or closed
             by this side of the connection.
         """
+        self._cancel_payload_deadline()
         if exc is None:
             self.logger.info('The master closed the connection')
         else:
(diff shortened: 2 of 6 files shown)


jotacarma90 (approved):
LGTM

PR #36762, opened 2026-06-04
Defer module coordination while FIM first sync is in progress

anromerom:
# Description

`wazuh-modulesd`'s agent-info module pauses FIM before pushing a metadata/groups sync, then polls FIM's `is_pause_completed` query under a hard 30 s budget (`MAX_PAUSE_POLL_ATTEMPTS = 30`). During FIM's **first** synchronization the module rebuilds its database and holds `fim_scan_mutex` for minutes, so the budget is exceeded: agent-info logged an ERROR, aborted coordination, and retried — repeatedly — until the first sync finished. This change makes agent-info **defer** coordination while the first sync is in progress (instead of timing out), narrows `fim_scan_mutex` so the delta-sync step runs unlocked, and primes the first-sync flag at startup so a restart does not re-defer. Resolves #36358.

## Proposed Changes

- **Narrow `fim_scan_mutex` (`src/syscheckd/src/run_check.c`):** `asp_sync_module()` (the delta sync) now runs **unlocked**; only the recovery/integrity loop — which reads/writes the `file_entry` table — holds `fim_scan_mutex` (plus `fim_realtime_mutex`, and `fim_registry_scan_mutex` on Windows). This removes the long mutex hold that caused the pause poll to time out.
- **Expose first-sync state to agent-info:** added the `fim_first_sync_completed` atomic (`src/config/include/syscheck-config.h`), set it on first-sync completion in `fim_run_integrity`, and report it as a `first_sync_completed` field in FIM's `is_pause_completed` response (`src/syscheckd/src/syscom.c`). When synchronization is disabled, it reports `true` so coordination is never blocked.
- **Defer instead of time out (`src/wazuh_modules/agent_info/agent_info_impl/`):** `pollFimPauseCompletion` returns `Deferred` while `first_sync_completed=false`; `coordinateModules` propagates `CoordinationResult::Deferred` and retries on the next cycle with no ERROR/abort. The deferral is logged at INFO once per episode, then DEBUG.
- **Probe FIM first (`COORDINATION_MODULES = {fim, sca, syscollector}`):** FIM is paused and probed before the other modules, so a deferral pauses and resumes **only FIM** and never disrupts SCA/Syscollector during the first-sync window.
- **Throttle reset:** the one-time INFO marker is reset both when the first sync completes and when the probe gives up (timeout/IPC failure), so a later deferral episode is still visible.
[...]

PR #37353, opened 2026-07-02
SCA/FIM sync lifecycle: close DBs on graceful shutdown, defer coordination during first sync, and increment SCA check version on change

### LINKED ISSUE #37334: Investigate and fix potential non-graceful DB shutdown on stop/restart
|---|---|---|---|---|
| 5.0.0-beta2 | State modules | Agent | Packages | Linux |

## Description
We found a case where the **SCA state was not synchronized with the Indexer**, while the agent still reported success:

> `wazuh-modulesd:sca: INFO: SCA synchronization finished successfully.`

A possible hypothesis is that synchronization messages were lost during an agent restart.

During investigation, we found multiple database support files that suggest DBs may not have been cleanly closed:

- `syscollector/db/syscollector_vd_sync.db`
- `syscollector/db/syscollector_sync.db`
- `syscollector/db/local.db`
- `sca/db/sca_sync.db-shm`
- `sca/db/sca_sync.db`
- `sca/db/sca.db-journal`
- `sca/db/sca_sync.db-wal`
- `sca/db/sca.db`
- `fim/db/fim_sync.db`
- `fim/db/fim_sync.db-shm`
- `fim/db/fim.db`
- `fim/db/fim_sync.db-wal`
- `agent_info/db/agent_info.db`

Expected behavior: on agent stop, shutdown should be graceful, including proper DB commit and close.

### Environment

- Wazuh Agent: `5.0.0-beta2`
- OS: `Ubuntu 24.04`

## Task
- Review the agent shutdown procedure on stop/restart.
- Ensure DB commit/close is correctly performed across all relevant modules:
  - SCA
  - FIM
  - Syscollector
  - Agent Info
- Investigate and fix potential non-graceful DB shutdown paths.
- Ensure synchronization success logs are not emitted if actual state is not synchronized.

## Definition of Done
- After a normal agent stop, no leftover SQLite support files are present (`-wal`, `-shm`, `-journal`) for module DBs.
- Integration tests are added/updated to validate graceful DB shutdown during agent stop.

### LINKED ISSUE #37402: SCA check 28583 remains Fail after installing iptables packages
|---|---|---|---|---|
| 5.0.0 Beta 2 | SCA | Agent | Packages | Ubuntu 22.04 |

## Description

In **SCA module** tests for **Wazuh v5.0.0-beta3** on an agent running **Ubuntu 22.04**, check **28583** (`4.3.1.1 Ensure iptables packages are installed`) does not update its status after remediation.

After installing `iptables` and `iptables-persistent`, a new finding is generated, but the check status remains `Fail` even after subsequent scan and synchronization cycles.

This behavior was reproduced **twice consecutively**.

## Steps to reproduce

1. Start from an endpoint where packages `iptables` and `iptables-persistent` are **not installed**.
2. Start the Wazuh agent and let it generate an initial state.
3. Confirm check **28583** reports **`Fail`**.
4. Set SCA scan interval and synchronization interval to **1 minute**.
5. Wait for agent configuration reload and a new scan cycle.
6. Install packages:
   - `apt install iptables iptables-persistent`
7. Wait for another SCA scan + synchronization cycle.

## Expected result

- A finding is generated.
- Check **28583** transitions from **`Fail`** to **`Pass`**.

## Actual result

- A finding is generated correctly.
- Check **28583** remains in **`Fail`** even after sufficient time and additional cycles.

## Tasks

- Reproduce the issue in a controlled environment with:
  - Wazuh `5.0.0-beta3`
  - Ubuntu 22.04 agent
  - SCA and sync intervals set to 1 minute
- Investigate root cause in SCA result lifecycle/state refresh for check 28583.
- Verify whether package detection logic or result persistence/synchronization is stale.
- Identify and implement a fix.
- Add/adjust automated tests to prevent regressions.
- Validate fix with at least two full fail→pass remediation cycles.

## Definition of done

- [ ] Issue reproduced consistently in the target environment.
- [ ] Root cause identified and documented.
- [ ] Code fix implemented and reviewed.
- [ ] Automated test coverage added/updated for this scenario.
- [ ] Check **28583** reliably changes from `Fail` to `Pass` after package installation.
- [ ] Validation repeated successfully in at least two consecutive runs.

jr0me:
## Description

Three agent-side fixes for the SCA/FIM/syscollector synchronization lifecycle, all stemming from the #37334 investigation:

1. **Graceful shutdown leaves the SCA and FIM synchronization databases open.** Stopping the agent leaves stale SQLite `-wal`/`-shm` files behind (a multi-megabyte un-checkpointed WAL in FIM's case) because the sync protocol is never destroyed — this is the leftover reported in #37334.
2. **agent-info module coordination is not guarded against SCA/syscollector first synchronization.** #36762 made coordination defer while FIM's first sync is in progress, but wired the `first_sync_completed` guard for FIM only; a group or metadata change arriving during SCA's or syscollector's first sync runs coordination mid-first-sync.
3. **An SCA check's document version never advances when its result changes.** DBSync auto-increments a `version` column on a modified row, but SCA was suppressing it, so a check flipping Pass↔Fail kept the same `document_version`. Because that value is what the manager sends to the indexer as the `external_gte` version, a changed document could not outrank the tombstone the manager's re-sync leaves behind after deleting the previous version.

Related to #37334. The empty-SCA-dashboard after a local DB rebuild is a manager-side re-sync version conflict, fixed separately in #37352; fix 3 here is its agent-side complement (a changed check's re-index can now outrank its own tombstone).

## Proposed Changes

**1. Close the sync protocol databases on graceful shutdown** (shutdown commit)

- **SCA** (`sca_impl.cpp`): `releaseResources()` now also resets `m_asyncFlushController` and `m_spSyncProtocol`, mirroring `Syscollector::releaseResources()` (the module that already shuts down cleanly). Safe at that point: `wm_sca_start()` joins the sync worker before releasing resources.
- **FIM** (`run_check.c`, `main.c`, `syscom.c`, `file.c`): the inventory synchronization thread is created **joinable** (`CreateThreadJoinable`) and the teardown runs in a dedicated shutdown thread, never in signal context. `fim_shutdown()` (registered with raw `signal()`, may run on any thread) now only does async-signal-safe work: it records the signal, raises the stop flags and wakes `fim_shutdown_waiter()` through a self-pipe. The waiter performs the teardown in normal thread context: `asp_stop()` → join the synchronization thread → `asp_destroy()` → `fim_db_teardown()` → `HandleSIG()`. The join is bounded: the worker polls the stop flag in 1-second slices and `asp_stop()` aborts an in-flight synchronization.
- **Why not in the handler**: joining from the handler can deadlock — if the signal lands on a thread that holds `syscheck.directories_lock` or `syscheck.fim_scan_mutex` (e.g. the main thread inside `fim_scan()`), that thread parks in the handler without releasing them while `fim_run_integrity` blocks on those very locks and never re-checks the stop flag, so the join never returns and the service manager SIGKILLs the process, leaving the very `-wal`/`-shm` files behind. Destroying in the handler is racy too: a second termination signal (raw `signal()` leaves the other signals unblocked) can re-enter the handler, skip the join and free the handle under the live synchronization thread.
- **Handle lifetime** (`fim_sync_handle_rwlock`): the `sync_handle` users that are not joined before the destroy — event persistence (`persist_syscheck_msg`), syscom's `fim_sync` responses, the DataClean paths and the scheduled-scan `asp_reset` — take a new leaf rwlock for reading around each `asp_*` call and re-check the handle under it; the shutdown waiter destroys the handle and resets it to `NULL` under the write lock, so an in-flight call can never dereference a freed handle. The teardown itself is Unix-only (`main.c` is `#ifndef WIN32`); the guards are shared and inert on Windows.

With the connections closed, SQLite's last close checkpoints the WAL and removes the `-wal`/`-shm` sidecar files.

**2. Defer coordination while SCA/syscollector first sync is in progress** (coordination-guard commit)

- **agent-info** (`agent_info_impl.{hpp,cpp}`): added `isModuleFirstSyncCompleted()`, which queries `get_first_sync_completed`; the SCA/syscollector branch of `pauseCoordinationModules()` now defers coordination — same throttled INFO log and pause/resume handling as the FIM guard — until their first sync completes. An error response (the modules answer with an error while the first-sync marker is unset) is treated as "in progress" (defer); a successful response without the field proceeds, so coordination is never wedged.
- **`wm_sca.c` / `wm_syscollector.c`**: when synchronization is disabled, the query handler reports `first_sync_completed=1` (mirrors FIM's `!enable_synchronization` handling in `syscom.c`), so coordination never waits on a sync that will never run. The startup-priming path calls the module query pointer directly and is unaffected.

**3. Let DBSync increment the SCA check version on a content change** (`c791418524`)

- The document version sent to the indexer for each SCA state is the `sca_check.version` column (`state.document_version`, and the `version_type=external_gte` version). DBSync auto-increments a `version` column on a modified row when the caller omits the field (`sqlite_dbengine.cpp`: `sql += "version=version+1,"`, gated on the field being absent and actual data having changed).
- `SCAEventHandler::ReportCheckResult` (`sca_event_handler.cpp`) loaded the full existing row via `GetPolicyCheckById` — whose `SELECT` includes `version` — set the new `result`/`checksum`, and wrote it back, so the row it handed to `syncRow` still carried the old `version`. That suppressed the auto-increment: a check flipping Pass↔Fail was re-indexed at the same `document_version`.
- Fix: `checkData.erase("version")` before building the `syncRow` query, so DBSync assigns the version (schema default `1` on a fresh insert, `version+1` on a real modification). One line; the checksum still fingerprints only the check definition, so the change detection is unaffected.

### Results and Evidence

**1 — Graceful shutdown DB close.** Baseline (stock 5.0.0 beta3 nightly, graceful `systemctl stop wazuh-agent`) — leftover files on every stop:

```
32768   /var/ossec/queue/fim/db/fim_sync.db-shm
4132392 /var/ossec/queue/fim/db/fim_sync.db-wal     <- 4.1 MB un-checkpointed WAL
32768   /var/ossec/queue/sca/db/sca_sync.db-shm
0       /var/ossec/queue/sca/db/sca_sync.db-wal
0       /var/ossec/queue/sca/db/sca.db-journal
total leftovers: 5
```

Syscollector's databases had live `-wal`/`-shm` while running and **zero** leftovers after stop — the reference behavior this PR extends to SCA and FIM. With this PR (agent built from this branch, 6 consecutive graceful stop/start cycles):

```
### STOP-LOOP x6 — graceful shutdown DB-close verification
--- cycle 1 ---  wal-files-while-running=4  stop=0procs in 3s  leftovers=0
--- cycle 2 ---  wal-files-while-running=4  stop=0procs in 3s  leftovers=0
--- cycle 3 ---  wal-files-while-running=4  stop=0procs in 3s  leftovers=0
--- cycle 4 ---  wal-files-while-running=4  stop=0procs in 3s  leftovers=0
--- cycle 5 ---  wal-files-while-running=4  stop=0procs in 3s  leftovers=0
--- cycle 6 ---  wal-files-while-running=4  stop=0procs in 2s  leftovers=0
=== RESULT: 6/6 cycles clean, 0 with leftovers ===
```

4 WAL files active while running → 0 leftovers after every graceful stop; stop latency stays at 2–3 s (the join adds nothing measurable); healthy after every restart.

**2 — Coordination guard.** When the guard fires, agent-info logs `Deferring coordination until sca first sync completes, will retry next cycle` (throttled to one INFO per deferral episode, DEBUG afterwards) and retries next cycle instead of coordinating mid-first-sync. Covered by the unit tests below.

**3 — Version increments on a content change (A/B on a from-scratch beta3 env).** Same all-in-one beta3 server (manager `005803d`), Ubuntu 22.04 agent, policy `cis_ubuntu22-04`. The agent was built twice — once at stock `origin/5.0.0`, once at this branch — and the same check `28541` (`stat` perms on `/etc/issue`, expected mode `644|640|600|400`) was flipped Pass↔Fail three times with `chmod`, reading `sca_check.version` after each scan:

```
                          STOCK 5.0.0            THIS PR (#37353)
  build has erase("version"):  0                      1
  baseline  (Passed)     result=Passed version=1   result=Passed version=1
  flip1 -> Failed        result=Failed version=1   result=Failed version=2
  flip2 -> Passed        result=Passed version=1   result=Passed version=3
  flip3 -> Failed        result=Failed version=1   result=Failed version=4
```

Stock keeps `version=1` through every result change; with this PR the version advances `1 -> 2 -> 3 -> 4`, one per change. End to end, the incremented version reaches the indexer:

```
# THIS PR, after flip3 (agent local, then indexer):
agent   sca.db : 28541 | Failed | 4
indexer states-sca: check 28541  result=Failed  document_version=4  _version=4
```

  - [x] Linux (agent built from this branch inside the test container for the A/B above; module UT jobs also compile it)
  - [x] Windows (`Agent-Info-Unit-Tests (wazuh_modules/agent_info, winagent)` and the Windows agent build pass on this branch)
  - [x] MAC OS X (`build-sonoma` compiles and runs the agent_info unit tests on this branch)
- [x] Log syntax and correct language review (new messages: the deferral INFO/DEBUG lines quoted above)

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

### Artifacts Affected

- `wazuh-modulesd` (agent): `libagent_info.so`, `libsca.so`, `wm_sca`/`wm_syscollector` handlers
- `wazuh-syscheckd` (agent, Unix): shutdown path and synchronization thread lifecycle
- Agent packages for Linux/macOS; the Windows agent only picks up the `wm_sca.c`/`wm_syscollector.c` query-handler change and the (inert there) `fim_sync_handle_rwlock` read guards — the shutdown teardown itself is `#ifndef WIN32`

### Configuration Changes

N/A — no new configuration parameters and no default-value changes.

### Tests Introduced

- `agent_info_coordination_test.cpp`:
  - `DeferCoordinationWhenScaFirstSyncNotCompleted` — FIM first sync done, SCA reports `first_sync_completed=0` → coordination defers on SCA.
  - `DeferCoordinationWhenSyscollectorFirstSyncNotCompleted` — FIM and SCA done, syscollector reports `0` → coordination defers on syscollector.
  - Both modeled on `DeferCoordinationWhenFimFirstSyncNotCompleted` from #36762; the pre-existing coordination tests (which do not mock `get_first_sync_completed`) still pass, covering the proceed-when-absent behavior.
- `sca_event_handler_test.cpp`:
  - `ReportCheckResult_DoesNotSendVersionInSyncPayload` — asserts the row handed to `syncRow` on a result change omits `version`, so DBSync performs the increment.
- `test_run_check.c`:
  - `test_persist_syscheck_msg_destroyed_sync_handle` — with synchronization enabled and the handle already destroyed (`NULL`), the event is dropped without calling `asp_persist_diff`.
- `test_syscom.c`:
  - `test_syscom_dispatch_sync_destroyed_handle` — a `fim_sync` response arriving after the handle was destroyed answers `err Error syncing module` without calling `asp_parse_response_buffer`.
- The shutdown fix is verified by the stop-loop above; syscheckd unit tests wrap `start_daemon`, so the joinable-thread launch and the shutdown waiter are not exercised by them.

## Review Checklist

- [ ] Code changes reviewed
- [x] Configuration changes documented (N/A)
- [x] No unresolved dependencies with other issues (independent of #37352; `main` needs a forward-port of all three changes)

## Follow-up: second review round

Eight additional commits address the findings of an adversarial review of the three fixes above, plus a second adversarial pass run over the result.

**Fixed**

- `fix(fim): publish the sync thread to the shutdown waiter under the handle lock` — closed the startup window where a termination signal between `CreateThreadJoinable()` and the `fim_sync_thread_initialized` store made the waiter skip the join and destroy the handle under the live synchronization thread. Creation and flag are now published under the `fim_sync_handle_rwlock` write lock with a shutdown gate; the waiter reads the flag under the same lock, and (second pass finding) re-clears `fim_sync_module_running` there so a creation that raced the handler still observes the stop and joins.
- `fix(fim): bound the shutdown waiter join so a stop cannot hang` — the queue write goes `SendBinaryMSG → os_wait()`, which sleeps in 5-second intervals until the manager-disconnected mark clears and ignores `asp_stop()`; with the manager unreachable the join hung until reconnection and the service manager's SIGKILL left the very `-wal`/`-shm` files this PR removes. Added `SendBinaryMSGPredicated()` (Unix + Windows twins) wired through `fim_send_binary_msg` with `fim_shutdown_process_on`, skip the post-sync recovery pass once shutdown starts (it takes `fim_scan_mutex`, which an in-flight scan can hold for minutes), and as a backstop the waiter polls an exit self-pipe for 20 s — under `wazuh-control`'s ~30 s SIGKILL escalation — and on timeout skips the join *and* the teardown rather than destroying the handle under a live thread.
- `fix(sca): break the manager-disconnected wait on shutdown in the binary send` (second pass finding) — same `os_wait()` park for SCA's sends: it stalled the sync worker join in `wm_sca_start()` and could hold `m_resourcesMutex` shared from a wcom query, blocking `releaseResources()` and leaking `sca_sync.db` in exactly the disconnected-stop case. Now predicated with `wm_sca_is_shutting_down`, matching the existing `StartMQPredicated` wiring.
- `fix(fim): serialize the shutdown database teardown against the scan transaction` — `fim.db` runs `journal_mode=truncate` (the teardown is about releasing the DBSync context in order before `exit()`, not WAL cleanup), and FIMDB's internal guards cover every path except the raw scan transaction. The transaction lifecycle now lives entirely inside `fim_scan_mutex`, the waiter tears down only under a trylock of that mutex (kept held so no new transaction can start), and `fim_db_transaction_start()` got the `try/catch` every other extern "C" wrapper in `db.cpp` already had (a post-teardown `DBSyncHandle()` throw previously escaped into C → `std::terminate`).
- `fix(sca): exclude live query paths from the module teardown` — `releaseResources()` reset `m_dBSync`/`m_asyncFlushController`/`m_spSyncProtocol` with no synchronization while the detached wcom dispatcher (and agent-info's in-process coordination queries, added by this very PR) could be inside `query()`/`parseResponseBuffer()`. Added `m_resourcesMutex` (`shared_mutex`): externally driven entry points take it shared, `initSyncProtocol()` publishes under the exclusive lock, and the teardown resets under it, moving the flush controller out first and joining its worker outside the lock. Syscollector shares the unlocked pattern upstream (its guarded `destroy()` is test-only) and is left for a follow-up, as is the unpredicated send in syscollector/agent-info.
- `fix(agent-info): keep the deferral log throttle per module` — the single `m_deferralLogged` flag was cleared by FIM's probe every cycle, so the SCA/syscollector deferral re-emitted its INFO every cycle and the DEBUG "Still deferring" branch was dead. Now a per-module set, with a two-cycle regression test.

**Reviewed but argued, no change**

- SCA's check checksum excluding `result` predates this PR (upstream design: the checksum fingerprints the check definition; `sca_checksum.cpp` is untouched here). Consequence worth recording on #37402: a *lost* result delta will not self-heal via the periodic integrity check, since both sides keep the same definition-only checksum — the version increment in this PR fixes the delivered-delta path (re-index outranks its own tombstone), not lost-delta recovery. Changing checksum semantics is a protocol-level decision with fleet-wide re-sync implications and belongs to its own issue.
- The `get_first_sync_completed` sync-disabled short-circuit exists in two near-identical copies (`wm_sca.c`, `wm_syscollector.c`); FIM's equivalent has a different shape (embedded in the `is_pause_completed` response, from #36762). Two five-line, module-named blocks did not justify a cross-module helper.

**Verification**

- Unit tests (agent target, container): `test_run_check` 32/32, `test_syscom` 18/18, `test_fim_scan` 3/3, `test_mq_op` 27/27, `agent_info_coordination_test` 38/38, `agent_info_successful_coordination_test` 5/5, `sca_unit_test` 21/21, `sca_event_handler_unit_test` 61/61, `sca_checksum_unit_test` 17/17.
- Live, source-built agent: connected graceful stop 2.0 s with zero `-wal`/`-shm` left for FIM and SCA; early stop while gate-blocked (unenrolled, manager down) 1.0 s, clean.
- The disconnected case this round is about: manager stopped mid-first-sync until syscheckd logged `Process locked due to agent is offline` (the exact state that previously hung the join forever), then `wazuh-control stop` — syscheckd exited in **4 s** (parked send broke at the next 5-second check) with `fim_sync.db`'s `-wal`/`-shm` removed, and after the SCA send fix `sca_sync.db` closes cleanly in the same scenario. Total stop stays bounded (~29 s) by modulesd's join budget, dominated by syscollector's still-unpredicated send (pre-existing, follow-up).

### DIFF
--- a/.github/actions/check_files/check_files.py
+++ b/.github/actions/check_files/check_files.py
@@ -314,9 +314,17 @@ def file_diff(expected_item, current_items, size_check):
             difference_bytes = abs(
                 float(current_items['size_bytes']) - expected_size_bytes)
             if (expected_size_bytes and (difference_bytes / expected_size_bytes) > expected_error):
-                differences['size_bytes'] = current_items['size_bytes']
+                differences['size_bytes'] = {
+                    'expected': expected_item['size_bytes'],
+                    'found': current_items['size_bytes'],
+                    'allowed_error': expected_error,
+                    'deviation': round(difference_bytes / expected_size_bytes, 4)
+                }
         elif expected_item[head_type] != current_items[head_type]:
-            differences[head_type] = current_items[head_type]
+            differences[head_type] = {
+                'expected': expected_item[head_type],
+                'found': current_items[head_type]
+            }
 
     return differences
 
@@ -364,8 +372,17 @@ def printReport(expected_items, not_listed, not_fully_match, current_items, matc
             report_lines.append("### Files Differences\n")
             report_lines.append(
                 f"{qtty_not_fully_matched} files didn't match the expected values:\n")
-            report_lines.extend(
-                f"- '{filename}' : {details}\n" for filename, details in not_fully_match.items())
+            for filename, details in not_fully_match.items():
+                report_lines.append(f"- '{filename}':\n")
+                for field, values in details.items():
+                    if 'allowed_error' in values:
+                        report_lines.append(
+                            f"  - {field}: expected {values['expected']} "
+                            f"(±{float(values['allowed_error']) * 100:g}%), found {values['found']} "
+                            f"(deviation {values['deviation'] * 100:g}%)\n")
+                    else:
+                        report_lines.append(
+                            f"  - {field}: expected '{values['expected']}', found '{values['found']}'\n")
 
         # Expected files not found
         missing_files = [key for key,
@@ -387,12 +404,12 @@ def printReport(expected_items, not_listed, not_fully_match, current_items, matc
     # Combine all lines into a single report string
     final_report = ''.join(report_lines)
 
-    # Output to file or console
+    # Always print to the console so the CI log shows the discrepancies directly,
+    # and also write the file when a report path was requested
+    print(final_report)
     if report_path:
         with open(report_path, 'w') as file:
             file.write(final_report)
-    else:
-        print(final_report)
 
 
 if __name__ == "__main__":
--- a/.github/actions/check_files/deb_linux_agent_amd64.csv
+++ b/.github/actions/check_files/deb_linux_agent_amd64.csv
@@ -18,7 +18,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/modern.bpf.o,root,wazuh,750,file,-rwxr-x---,998504,0.1
 /var/ossec/lib/libstdc++.so.6,root,wazuh,750,file,-rwxr-x---,2599704,0.1
 /var/ossec/lib/libsyscollector.so,root,wazuh,750,file,-rwxr-x---,793656,0.1
-/var/ossec/lib/libsca.so,root,wazuh,750,file,-rwxr-x---,800872,0.1
+/var/ossec/lib/libsca.so,root,wazuh,750,file,-rwxr-x---,886544,0.1
 /var/ossec/lib/libagent_info.so,root,wazuh,750,file,-rwxr-x---,345184,0.1
 /var/ossec/lib/libagent_metadata.so,root,wazuh,750,file,-rwxr-x---,14448,0.1
 /var/ossec/lib/libgcc_s.so.1,root,wazuh,750,file,-rwxr-x---,183248,0.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,7 @@ All notable changes to this project will be documented in this file.
 - Honored the shutdown signal in `agent-upgrade` `StartMQ` to avoid timeout warning on agent stop. ([#36092](https://github.com/wazuh/wazuh/issues/36092))
 - Adjusted DockerListener messages as log entries to fix event categorization. ([#36126](https://github.com/wazuh/wazuh/issues/36126))
 - Dropped orphan paths before promoting on agent startup to fix FIM. ([#36134](https://github.com/wazuh/wazuh/issues/36134))
+- Deferred agent-info module coordination while SCA or System Inventory first synchronization is in progress, extending the FIM guard so a concurrent group or metadata change does not run coordination mid-first-sync. ([#36358](https://github.com/wazuh/wazuh/issues/36358))
 
 ## Prior versions
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,7 @@ All notable changes to this project will be documented in this file.
 - Adjusted DockerListener messages as log entries to fix event categorization. ([#36126](https://github.com/wazuh/wazuh/issues/36126))
 - Dropped orphan paths before promoting on agent startup to fix FIM. ([#36134](https://github.com/wazuh/wazuh/issues/36134))
 - Deferred agent-info module coordination while SCA or System Inventory first synchronization is in progress, extending the FIM guard so a concurrent group or metadata change does not run coordination mid-first-sync. ([#36358](https://github.com/wazuh/wazuh/issues/36358))
+- Closed the SCA and FIM synchronization databases cleanly on graceful shutdown, so stopping the agent no longer leaves stale SQLite `-wal`/`-shm` files behind. ([#37334](https://github.com/wazuh/wazuh/issues/37334))
 
 ## Prior versions
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,8 +102,6 @@ All notable changes to this project will be documented in this file.
 - Honored the shutdown signal in `agent-upgrade` `StartMQ` to avoid timeout warning on agent stop. ([#36092](https://github.com/wazuh/wazuh/issues/36092))
 - Adjusted DockerListener messages as log entries to fix event categorization. ([#36126](https://github.com/wazuh/wazuh/issues/36126))
 - Dropped orphan paths before promoting on agent startup to fix FIM. ([#36134](https://github.com/wazuh/wazuh/issues/36134))
-- Deferred agent-info module coordination while SCA or System Inventory first synchronization is in progress, extending the FIM guard so a concurrent group or metadata change does not run coordination mid-first-sync. ([#36358](https://github.com/wazuh/wazuh/issues/36358))
-- Closed the SCA and FIM synchronization databases cleanly on graceful shutdown, so stopping the agent no longer leaves stale SQLite `-wal`/`-shm` files behind. ([#37334](https://github.com/wazuh/wazuh/issues/37334))
 
 ## Prior versions
 
--- a/src/shared/include/mq_op.h
+++ b/src/shared/include/mq_op.h
@@ -124,6 +124,24 @@ int SendMSG(int queue, const char *message, const char *locmsg, char loc) __attr
 int SendBinaryMSG(int queue, const void *message, size_t message_len, const char *locmsg, char loc);
 
 /**
+ * Sends a message binary through a message queue, breaking out of the
+ * manager-disconnected wait (os_wait) when the predicate turns true
+ * @param queue file descriptor of the queue where the message will be sent (UNIX)
+ * @param message binary containing the message
+ * @param message_len The length of the message payload in bytes
+ * @param locmsg path to the queue file
+ * @param loc  queue location (WIN32)
+ * @param fn_ptr function pointer to the function that will check if the process is shutting down
+ * @return
+ * UNIX -> 0 if file descriptor is still available
+ * UNIX -> -1 if there is an error in the socket. The socket will be closed before returning (StartMQ should be called to restore queue)
+ * WIN32 -> 0 on success
+ * WIN32 -> -1 on error
+ * Notes: (UNIX) If the socket is busy when trying to send a message a DEBUG2 message will be loggeed but the return code will be 0
+ */
+int SendBinaryMSGPredicated(int queue, const void *message, size_t message_len, const char *locmsg, char loc, bool (*fn_ptr)()) __attribute__((nonnull));
+
+/**
  * Sends a message to a socket. If the socket has not been created yet it will be created based on
  * the target information. If a message fails to be sent the method will not try to send it again until *sock_fail_time* has passed
  * @param queue file descriptor of the queue where the error message will be sent (UNIX)
--- a/src/shared/src/mq_op.c
+++ b/src/shared/src/mq_op.c
@@ -285,6 +285,13 @@ int SendBinaryMSG(int queue, const void *message, size_t message_len, const char
     return SendBinaryMSGAction(queue, message, message_len, locmsg, loc);
 }
 
+/* Send a binary message with a predicate to break out of the manager-disconnected wait */
+int SendBinaryMSGPredicated(int queue, const void *message, size_t message_len, const char *locmsg, char loc, bool (*fn_ptr)()) {
+    /* Check for global locks */
+    os_wait_predicate(fn_ptr);
+    return SendBinaryMSGAction(queue, message, message_len, locmsg, loc);
+}
+
 /* Send a message to socket */
 int SendMSGtoSCK(int queue, const char *message, const char *locmsg, __attribute__((unused)) char loc, logtarget * target)
 {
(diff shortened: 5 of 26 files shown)


Nicogp (approved):
LGTM

PR #37375, opened 2026-07-02
Reduce memory usage when downloading VDP feed

### LINKED ISSUE #37161: Analyze memory usage when downloading VDP feed
## Description

The Vulnerability Detector (VD) module downloads the VDP feed from the Indexer during feed updates. This process may cause increased memory usage in the Wazuh Manager due to the download, temporary storage, and parsing of the feed data.

This task aims to analyze the memory consumption during a VDP feed update, identify possible optimization opportunities, and apply the required changes to reduce memory usage.

The main goal is to understand where memory is being consumed and implement improvements in areas such as feed download, parsing, or data processing to reduce the memory impact on the Wazuh Manager.

Image

## Tasks

* [ ] Analyze the VDP feed download process in the Wazuh Manager.
* [ ] Measure memory usage during a complete VDP feed update.
* [ ] Identify the main points where memory consumption increases.
* [ ] Review current download, parsing, and processing logic for unnecessary memory usage.
* [ ] Identify possible optimizations to reduce memory consumption.
* [ ] Implement the identified optimizations.
* [ ] Validate that the changes do not affect:

  * [ ] Vulnerability detection results.
  * [ ] Feed update correctness.
  * [ ] VD processing behavior.
* [ ] Measure memory usage after the changes and compare results.
* [ ] Document findings and implemented improvements.

## Definition of Done

* [ ] Current memory usage during VDP feed updates is measured and documented.
* [ ] Main sources of memory consumption are identified.
* [ ] Memory optimization opportunities are identified.
* [ ] Identified optimizations are implemented.
* [ ] Memory usage improvement is measured after the changes.
* [ ] No functional regression is introduced in Vulnerability Detector feed updates.

Antoniogm03:
## Description

During the initial load of the vulnerability feed (VDP), the `wazuh-manager-modulesd` process
reached very high memory peaks while downloading and processing the ~250,000-350,000 CVE
documents from the `.wazuh-threatintel-vulnerabilities` index. That peak did not drop
significantly across previous point optimizations because its root cause, RocksDB's
internal memory Arena, fed by a shared `WriteBatch` with no size limit, had not been
identified or addressed.

This PR reduces the peak RSS measured during feed download from **1.57 GB** (5.0.0 baseline)
to **~530 MB** with the new `pageSize=100` default (see the comparison table below), with no
penalty to download time (it actually drops slightly, 4:56 → 4:04), and stabilizes the memory
consumption graph by eliminating isolated spikes that stood apart from the rest of the memory
profile.

Closes #37161

### Indexer download path (`IndexerDownloader.hpp`)

- **Default `pageSize` changed from 250 to 100** documents per page, now validated between
  `1` and `10000` and guarded against a configured `0`, which would otherwise make every page come back empty and stall the load. Smaller pages reduce the transient in-memory peak of a fetched page at the cost of more HTTP round-trips.
- **Removed an unnecessary copy of the document** in `processPage`: `source` and `document`
  are now taken by reference and the payload is serialized directly, instead of first copying
  the whole subtree (`source.value("document", ...)`) before dumping it.
- **New `waitUntilGlobalMapsIndexed()` method**: before starting a full initial load, it polls (every 30s) a lightweight query to confirm the three global-map documents the local database needs (`FEED-GLOBAL`, `OSCPE-GLOBAL`, `CNA-MAPPING-GLOBAL`) are already indexed. Previously, a full load could run for minutes and then be discarded because these documents were missing, forcing a full re-download on the next cycle.
- **Periodic flush to disk every 250 pages** (instead of relying only on the final flush), and
  the resume cursor is now persisted only up to the last page that was actually part of a
  flushed batch, so a crash or `SIGKILL` mid-download can't leave a persisted cursor pointing
  past data that never made it to disk.
- **Sliced initial load no longer persists a partial cursor** when a slice is interrupted by a
  stop request; the next cycle re-does the full initial load instead of resuming from an
  unsafe, incomplete offset.

### FlatBuffers reuse and parsing (`eventDecoder.hpp`)

- **`flatbuffers::Parser` instances are now cached per resource type and operation** instead of being rebuilt, re-parsing ~7.6 KB of schema IDL, for each of the 250,000+ documents processed. The same parser is reused, clearing only its internal builder between documents.
- Payloads are now passed and parsed as an already-serialized string instead of being
  `dump()` again inside the decoder.

### Batched writes to RocksDB (`databaseFeedManager.hpp`)

- **The `WriteBatch` is now shared across pages** instead of being scoped to a single
  `processMessage` call, and it commits both by document count and by accumulated size, whichever comes first, instead of only at the end of each message. This bounds every single `Write()` call, regardless of how the Indexer batches documents into messages.
- **The batch's capacity is pre-reserved and reused** across commits, and only rebuilt when a
  single commit turns out abnormally large.
- **Batch and decoder state are reset on exception**, so a failure partway through a page can't
  leak a partially-built batch into the next processing cycle.
- **Skips the read-before-write check for CVE candidates during the initial load**: with an
  empty database there is nothing to compare against, so the `seek`+`get` used to detect
  stale/changed candidates is skipped entirely on first load.
- Read block cache size is now configurable; the feed database keeps 16 MB.

### RocksDB column family lookup (`rocksDBWrapper.hpp`)

- **O(1) column family lookup** via a new `unordered_map` index (`columnExists`,
  `getColumnFamilyBasedOnName`), replacing a linear scan over `m_columnsInstances`. With
  hundreds of per-CNA column families, that scan was the dominant per-document cost during the
  feed load.
- **New `readCacheSize` constructor parameter** on `TRocksDBWrapper`, so the block cache size
  no longer has to be hardcoded to 16 MB.

### Network resiliency (`indexerConnectorSyncImpl.hpp`)

- **HTTP 429 responses now retry with exponential backoff** (starting at
  the configured `RetryDelay`, doubling up to a 30s cap) instead of a fixed 1s delay, across
  bulk, bulk-chunk, delete-by-query, `executeSearchQuery` and PIT `search`.
- `executeSearchQuery` and PIT `search` now abort cleanly when a stop is requested instead of
  retrying indefinitely.

### Configuration validation (`vulnerabilityScannerFacade.cpp`)

- `pageSize` and `numSlices` are now validated through a shared helper that accepts numeric or
  string XML values and rejects `0` or out-of-range values with a clear error: `pageSize` in
  `[1, 10000]`, `numSlices` in `[1, 32]`.

### Tests

- New unit tests: HTTP 429 retry with backoff for `executeSearchQuery` and PIT `search`,
  batch-state reset in `databaseFeedManager` after a decoder failure mid-page, and
  payload-as-string coverage in the decoder tests.

## Optimizations at a glance

**Download (lower RAM peak)**
- Default `pageSize` 250 to 100 (configurable, validated 1-10000, guarded against 0): smaller pages -> less JSON in flight -> lower peak.
- Removed a document copy in `processPage`: `source`/`document` taken by reference and serialized directly instead of copying the subtree.

**Decoding (less CPU/churn)**
- `flatbuffers::Parser` cached per type (create/update × CVE/translation): schema parsed once, not per document.
- Skips the candidate read-before-write on initial load: an empty DB has nothing to compare against, so the useless `seek`+`get` is skipped.

**RocksDB (write/lookup)**
- O(1) column family index (`unordered_map`) instead of a linear scan -> removes the near-quadratic cost from load time.
- `WriteBatch` bounded by size (2 MB) and count (25), capacity pre-reserved and rebuilt only if it spikes -> no memtable insertion spikes.
- Configurable block cache (new `readCacheSize` param, kept at 16 MB).

**Durability (WAL disabled)**
- Periodic flush every 250 pages + cursor persisted only up to an already-flushed page -> no silent data loss on crash.
- Sliced initial load: no partial cursor persisted if interrupted -> doesn't degrade to an incomplete feed after a stop.
- Batch state reset on exception -> a failed cycle's writes don't leak into the next one.

**Network robustness**
- Pre-check of global maps in the Indexer (`waitUntilGlobalMapsIndexed`) before the full load -> avoids downloading for minutes only to discard it.
- 429 retry with exponential backoff (capped at 30s) on every search/bulk path, abortable on stop.

**Other**
- `jemalloc`: a `malloc_conf` tuning was tried and ultimately reverted (net: no change to `main.c`).
- Config + docs: `numSlices` validated (1-32); `pageSize`/`numSlices` documentation updated.
- Tests added: 429 retry, batch reset, and payload-as-string in the decoder.

#### Vulnerabilities

<details>
<summary>5.0.0</summary>

```
root@indexer2:/home/vagrant# curl -s -k -u admin:admin \
  -X GET "https://localhost:9200/wazuh-states-vulnerabilities/_count" \
  -d '{
    "query": {
      "term": { "wazuh.agent.id": "002" }
    }
  }'
{"count":3918,"_shards":{"total":1,"successful":1,"skipped":0,"failed":0}}root@indexer2:/home/vagrant#

```

</details>

<details>
<summary> PR changes applied </summary>

```
root@manager:/home/vagrant# curl -s -k -u admin:admin \
  -X GET "https://localhost:9200/wazuh-states-vulnerabilities/_count" \
  -d '{
    "query": {
      "term": { "wazuh.agent.id": "001" }
    }
  }'
{"count":3918,"_shards":{"total":1,"successful":1,"skipped":0,"failed":0}}

```

</details>

#### Comparison

- 5.0.0 (04:56 minutes downloading feed)

image

- PageSize=50 (5:11 minutes downloading feed)

image

- PageSize=100 (4:04 minutes downloading feed)

image

- 1 thread PageSize=100  (4:07 minutes downloading feed)

image

- PageSize=150 (3:41 minutes downloading feed)

image

- PageSize=200 (3:12 minutes downloading feed)

image

- PageSize=250 (3:16 minutes downloading feed)

image

- All the graphs with my changes in one

image

 - All the graphs including the 5.0.0 branch

image

| pageSize | Download real time | Peak CPU | Average CPU | Peak RSS | Average memory (only download) |
|---|---|---|---|---|---|
| **5.0.0 branch** | **04m 56s** | **202,0 %** | **143,8 %** | **1,57 G** | **719 M** |
| 50  | 5m 11s | 220,0 % | 78,5 % | 437 M | 223,8 M |
| 100 | 4m 04s | 227,0 % | 97,8 % | 532 M | 309,1 M |
| 100 1 thread | 4m 07s | 208,0 % | 95 % | 499 M | 296 M |
| 150 | 3m 41s | 222,0 % | 105,5 % | 620 M | 355,6 M |
| 200 | 3m 12s | 249,0 % | 121,8 % | 642 M | 385,8 M |
| 250 | 3m 16s | 231,0 % | 121,6 % | 782 M | 425,5 M |

## Review Checklist

- [x] Code changes reviewed
- [ ] ...

### DIFF
--- a/docs/guide/migration/vulnerability-detection-cti-feeds.md
+++ b/docs/guide/migration/vulnerability-detection-cti-feeds.md
@@ -61,7 +61,7 @@ The 5.x module reads only these options under `<vulnerability-detection>`; any o
 |---|---|
 | `enabled` | `yes` / `no`. |
 | `feed-update-interval` | How often Vulnerability Detection refreshes the feed from the Indexer. Default and minimum: `60m`. |
-| `pageSize` | Optional. CVE documents fetched per page during feed download. Default: `250`. |
+| `pageSize` | Optional. CVE documents fetched per page during feed download. Default: `100`. |
 | `numSlices` | Optional. Parallel slices used for the initial feed load. Default: `2`. |
 
 ## Verify the feed loaded (optional)
--- a/docs/ref/modules/content_manager/README.md
+++ b/docs/ref/modules/content_manager/README.md
@@ -48,7 +48,7 @@ The module is configured programmatically by its caller (the Vulnerability Scann
 | `configData.databasePath` | Path to the RocksDB cursor database |
 | `configData.indexer.hosts` | Indexer endpoint URLs |
 | `configData.indexer.index` | Index to read from |
-| `configData.indexer.pageSize` | Documents per PIT page (default 250) |
+| `configData.indexer.pageSize` | Documents per PIT page (default 100) |
 | `configData.indexer.numSlices` | Parallel PIT slices for initial load (default 2) |
 
 The Indexer connection parameters come from the manager's `<indexer>` XML block (see [Indexer Configuration](configuration.md)).
--- a/docs/ref/modules/vulnerability-scanner/configuration.md
+++ b/docs/ref/modules/vulnerability-scanner/configuration.md
@@ -40,8 +40,8 @@ Time interval for downloading updated CVE feeds from the CTI service.
 
 Number of CVE documents fetched per PIT (Point In Time) page during feed download.
 
-- **Default value:** `250`
-- **Allowed values:** Positive integer
+- **Default value:** `100`
+- **Allowed values:** Integer between `1` and `10000` (the Indexer rejects sizes above `index.max_result_window`)
 - **Note:** Larger page sizes reduce HTTP overhead but increase memory usage per request
 
 ### numSlices
@@ -49,7 +49,7 @@ Number of CVE documents fetched per PIT (Point In Time) page during feed downloa
 Number of parallel PIT slices used during the initial feed load.
 
 - **Default value:** `2`
-- **Allowed values:** Positive integer
+- **Allowed values:** Integer between `1` and `32`
 - **Note:** Set to `1` to force sequential mode. Increasing this value can raise memory usage without delivering meaningful time savings. Benchmark before increasing
 
 ---
@@ -149,7 +149,7 @@ Standard vulnerability detection settings:
 <vulnerability-detection>
   <enabled>yes</enabled>
   <feed-update-interval>60m</feed-update-interval>
-  <pageSize>250</pageSize>
+  <pageSize>100</pageSize>
   <numSlices>2</numSlices>
 </vulnerability-detection>
 ```
@@ -162,7 +162,7 @@ For environments requiring more current vulnerability data:
 <vulnerability-detection>
   <enabled>yes</enabled>
   <feed-update-interval>30m</feed-update-interval>
-  <pageSize>250</pageSize>
+  <pageSize>100</pageSize>
   <numSlices>2</numSlices>
 </vulnerability-detection>
 ```
@@ -220,14 +220,14 @@ Prevent all vulnerability detection:
 **For small deployments (<100 agents):**
 ```xml
 <feed-update-interval>2h</feed-update-interval>
-<pageSize>250</pageSize>
+<pageSize>100</pageSize>
 <numSlices>1</numSlices>
 ```
 
 **For medium deployments (100-1000 agents):**
 ```xml
 <feed-update-interval>1h</feed-update-interval>
-<pageSize>250</pageSize>
+<pageSize>100</pageSize>
 <numSlices>2</numSlices>
 ```
 
--- a/src/shared_modules/content_manager/README.md
+++ b/src/shared_modules/content_manager/README.md
@@ -63,7 +63,7 @@ After all pages are processed, the downloader signals completion via `indexer_co
             "index": ".wazuh-threatintel-vulnerabilities",
             "consumerStatusIndex": ".wazuh-cti-consumers",
             "consumerStatusId": "cti:catalog:consumer:vulnerabilities",
-            "pageSize": 250,
+            "pageSize": 100,
             "numSlices": 2
         }
     }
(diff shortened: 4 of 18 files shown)


jotacarma90 (approved):
LGTM

PR #37379, opened 2026-07-02
fix: close SCA and FIM sync protocol databases on graceful shutdown

jr0me:
## Description

Closes the SCA and FIM synchronization databases cleanly on a graceful agent shutdown. This is the `-wal`/`-shm` leftover reported in #37334.

## Analysis

On a graceful stop, every stopped agent leaves stale SQLite sidecar files behind:

| database | leftover after graceful stop | cause |
|---|---|---|
| `queue/fim/db/fim_sync.db` | `-wal` (multi-MB, un-checkpointed) + `-shm` | sync protocol handle never destroyed; the sync thread is detached so shutdown has nothing to wait on |
| `queue/sca/db/sca_sync.db` | `-wal` + `-shm` | `releaseResources()` resets only `m_dBSync`; the sync protocol and flush controller leak |
| `queue/syscollector/db/*_sync.db` | none | **reference behavior** — `Syscollector::releaseResources()` resets every handle, so the last close checkpoints the WAL and removes the sidecars |

A leaked connection means the process exits without SQLite's last-close cleanup (`sqlite3_close_v2` checkpoints the WAL and removes `-wal`/`-shm`). The WAL replays on the next open, so this is not by itself data loss, but it leaves multi-megabyte artifacts on every stop, reads as corruption in the field, and unclean local DB state is exactly what arms the re-sync loss fixed in #37352.

## Proposed changes

- **SCA** (`sca_impl.cpp`): `releaseResources()` also resets `m_asyncFlushController` and `m_spSyncProtocol`, mirroring syscollector. Safe at that point: `wm_sca_start()` joins the sync worker before releasing resources.
- **FIM** (`run_check.c`, `main.c`): the inventory synchronization thread is created **joinable**; `fim_shutdown()` joins it after `asp_stop()` and then destroys the sync protocol handle (`asp_destroy`). The join is bounded — the worker polls the stop flag every second and `asp_stop()` aborts an in-flight synchronization.
- **Signal safety**: `fim_run_integrity` blocks the signals `fim_shutdown()` is registered for (SIGTERM/INT/QUIT/ALRM), so the handler can never run on the very thread it joins (no self-join deadlock).
- Unix-only, `#ifndef WIN32`-guarded (the Windows agent has no `fim_shutdown` signal path).

## Verification

Built the agent from this branch (5.0.0, arm64 container) and ran 6 consecutive graceful stop/start cycles against a beta3 manager:

```
before fix (stock beta3): 5 leftover files on every stop
  fim_sync.db-wal (4.1 MB), fim_sync.db-shm, sca_sync.db-wal, sca_sync.db-shm, sca.db-journal

after fix: 6/6 cycles
  wal-files-while-running=4 -> leftovers after stop = 0
  stop latency 2-3 s (join adds nothing measurable)
  agent healthy after every restart (5 daemons, 0 log errors), synchronization works
```

## Notes

- Related to #37334; the empty-SCA-dashboard part of that investigation is addressed separately in #37352 (manager-side re-sync version conflict) — this PR is the shutdown-hygiene half.
- `main` has the same shutdown behavior and needs the same change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,7 @@ All notable changes to this project will be documented in this file.
 - Honored the shutdown signal in `agent-upgrade` `StartMQ` to avoid timeout warning on agent stop. ([#36092](https://github.com/wazuh/wazuh/issues/36092))
 - Adjusted DockerListener messages as log entries to fix event categorization. ([#36126](https://github.com/wazuh/wazuh/issues/36126))
 - Dropped orphan paths before promoting on agent startup to fix FIM. ([#36134](https://github.com/wazuh/wazuh/issues/36134))
+- Closed the SCA and FIM synchronization databases cleanly on graceful shutdown, so stopping the agent no longer leaves stale SQLite `-wal`/`-shm` files behind. ([#37334](https://github.com/wazuh/wazuh/issues/37334))
 
 ## Prior versions
 
--- a/src/syscheckd/src/main.c
+++ b/src/syscheckd/src/main.c
@@ -44,6 +44,8 @@ __attribute__((noreturn)) static void help_syscheckd()
 
 extern bool is_fim_shutdown;
 extern volatile int fim_sync_module_running;
+extern pthread_t fim_sync_thread;
+extern bool fim_sync_thread_initialized;
 
 /* Shut down syscheckd properly */
 static void fim_shutdown(int sig)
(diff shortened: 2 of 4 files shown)


jr0me:
Consolidated into #37353 (same agent-side scope for #37334); the shutdown fix is now the second commit there.

PR #37430, opened 2026-07-06
Fix Windows block-ip firewall-enabled check misfire and ineffective route fallback

### LINKED ISSUE #37388: Firewall-enabled check misfires and skips netsh, falling back to an ineffective route
## Context

During issue https://github.com/wazuh/wazuh-qa-automation/issues/7688, a problem was found with `active-response`. On Windows, the `block-ip` active response runs but the IP is not actually blocked, and `active-responses.log` stays empty. The executable name is not the problem: `block-ip` is the correct name on Windows too. Same name is used on Linux, macOS, and Windows.

- [src/active-response/CMakeLists.txt#L140-L141](https://github.com/wazuh/wazuh/blob/c698547e39996ca9c84e0ce861e8fd84489d952f/src/active-response/CMakeLists.txt#L140-L141)

After investigating the problem, a possible error has been detected that requires investigation.

## Description

This is the **local firewall policy hive**, not the Group Policy one: `HKLM\SOFTWARE\Policies\Microsoft\WindowsFirewall\...`

- On domain-managed or hardened VMs this local key can show "disabled" even when the firewall is actually running.  
- When the code sees "disabled" it skips `netsh` entirely (`FIREWALL_INVALID_STATE`).

The fallback using `route` (lines 262–369) does **not** block the attacker IP:

- It finds the default gateway with `ipconfig` and runs: `route -p ADD <attacker_ip> MASK 255.255.255.255 <gateway>`.
- That just adds a normal host route via the same gateway the host already uses, so it does not stop incoming connections.
- Even a true null route would only affect outbound traffic from the Windows host, not inbound connections (like the WinRM login in this test).

Result: the code can skip `netsh` and then fall back to a route that does not prevent the tested inbound connection, which explains the observed behavior.

- Related code:
  - [src/block-ip-windows.c#L70-L74](https://github.com/wazuh/wazuh/blob/5.0.0/src/active-response/src/block-ip-windows.c#L70-L74)
  - [src/block-ip-windows.c#L109-L113](https://github.com/wazuh/wazuh/blob/5.0.0/src/active-response/src/block-ip-windows.c#L109-L113)
  - [src/block-ip-windows.c#L138-L143](https://github.com/wazuh/wazuh/blob/5.0.0/src/active-response/src/block-ip-windows.c#L138-L143)
  - [src/active-response/src/block-ip-windows.c#L262-L369](https://github.com/wazuh/wazuh/blob/5.0.0/src/active-response/src/block-ip-windows.c#L262-L369)

## Tasks

- [ ] Investigate the reported error
  - [ ] If necessary, correct any errors found
  - [ ] Perform tests to ensure the fix is ​​working properly

nbertoldo:
## Description

On Windows, the `block-ip` active response failed to block IPs: the executable ran and reported success, but the target IP stayed reachable (e.g. WinRM/RDP logins kept working).

`try_netsh` pre-checked whether the firewall was "enabled" by running `reg query` against the local SharedAccess store:

```
HKLM\SYSTEM\CurrentControlSet\Services\SharedAccess\Parameters\FirewallPolicy\{Domain,Standard,Public}Profile\EnableFirewall
```

and looked for `0x1` in the output. This misfires whenever the value is **not present** while the firewall is running: the reg query returns no data, so the check concludes "no profile enabled", returns `FIREWALL_INVALID_STATE`, and **skips `netsh` entirely** — even though `netsh` is fully functional — falling back to the `route` method. This is not a rare edge: with `EnableFirewall` **absent**, the firewall is effectively **ON** (Windows default), yet the check reads it as disabled. The local SharedAccess value is also not authoritative on hosts where the firewall is managed elsewhere. (Confirmed empirically below.)

The `route` fallback could never provide the intended protection:

- It routed the target through the host's **real default gateway**, which does not blackhole anything.
- It parsed `ipconfig` and took the first `Default Gateway` line, which on a dual-stack host is the **IPv6 link-local** address (`fe80::…`); the resulting `route add <IPv4> MASK 255.255.255.255 <IPv6-gw>` is invalid and `route.exe` rejects it.
- The `route add` exit code was never checked, so the method returned `FIREWALL_SUCCESS` while adding nothing — a **silent false success**.

Closes #37388

## Proposed Changes

**`src/active-response/src/block-ip-windows.c`**

- **Firewall-state detection (`firewall_is_effectively_on`, new)** — Read the `EnableFirewall` DWORD directly from the registry via the Win32 API (`RegOpenKeyEx`/`RegQueryValueEx`, 64-bit view). Per profile: the GPO policy value (`SOFTWARE\Policies\Microsoft\WindowsFirewall\<profile>`) wins if present, else the local SharedAccess value, else it defaults to **ENABLED** (the Windows default). Defaulting an **absent** value to enabled is the actual misfire fix — the old check treated absent as "disabled". 
- **`try_netsh`** — On **ENABLE** only, if the firewall is effectively off, return `FIREWALL_INVALID_STATE` so the chain defers to `route` instead of adding a rule that would sit dormant. The check is deliberately **not** applied on DISABLE, so an unblock always removes the rule (which may have been added while the firewall was enabled). This keeps the "one method or the other" chain semantics while fixing the misfire (correct signal source). Also fixed the `remoteip` prefix to match the address family — `/32` for IPv4, `/128` for IPv6 — instead of a hardcoded `/32` that malformed the rule for IPv6 targets.
- **`try_route_windows`** — Rework the fallback into a null-route:
  - Null-route the target to the loopback (`route -p ADD <ip> MASK 255.255.255.255 127.0.0.1`) instead of routing via the host's real default gateway.
  - Remove the `ipconfig` parsing entirely (eliminates the IPv6-link-local gateway bug and the locale-dependent string matching).
  - Skip IPv6 targets (`FIREWALL_NOT_AVAILABLE`) rather than emit an invalid route; `netsh` already covers IPv6.
  - On ENABLE, check the `route add` exit code and return `FIREWALL_EXECUTION_FAILED` on error (ending the silent false-success). The DELETE path is **best-effort** — a missing route is the expected case on cleanup.
  - Log an explicit **best-effort** warning documenting the limitation (below).
- **DISABLE cleanup (semantics fix)** — Unblock is no longer routed through the fallback chain. The block may have been applied by *either* netsh (firewall on) *or* the null-route (firewall off), so DISABLE now removes **both** unconditionally, each best-effort. Previously the chain only reached `route delete` if `netsh delete rule` returned non-zero, which relied on netsh's **undocumented** "errorlevel 1 on no rule matched" — had it ever returned 0, the null-route would have been orphaned while the log reported "unblocked" (raised in review). The delete helpers now return `SUCCESS` only when they actually removed something; if **neither** removed anything, a `WARNING` is logged so a genuine removal failure is not silent. The command still returns `OS_SUCCESS` for consistency with the ENABLE chain — execd ignores the AR exit code (it never reads `wpclose`'s return), so failures are surfaced via the log, not the return code.

**ENABLE** uses the `netsh → route` chain (mutually exclusive: one method or the other). `netsh` (Windows Firewall) is the comprehensive, bidirectional mechanism, used whenever the firewall is effectively enabled. `route` is reached only when the firewall is off (or `netsh.exe` is missing) and is **best-effort**: a loopback null-route breaks the attacker's reverse path (which also disrupts inbound TCP sessions) for **routed/remote** attackers, but does **not** block a target on a **directly-connected subnet** (on-link hosts are reached via ARP and the `/32`-via-loopback route does not redirect that egress — verified on WS2022). This is documented in-code and surfaced as a warning in `active-responses.log`. **DISABLE** removes both artifacts unconditionally (above).

### Results and Evidence

Reproduced on a standalone **Windows Server 2022** agent (5.0.0) by driving `block-ip.exe` directly via stdin, as `execd` does.

**Root cause.** Setting the local `EnableFirewall` value to each state and comparing what the old reg-query check saw against the effective firewall state:

| SharedAccess `EnableFirewall` | Effective firewall (`netsh advfirewall show allprofiles state`) | Old check finds `0x1`? |
|---|---|---|
| `0x1` | ON | yes → uses netsh (correct) |
| `0x0` | OFF | no → firewall genuinely off |
| **absent** | **ON** | **no → MISFIRE** |

The `absent` row is the bug: the firewall is effectively **on** (Windows default) but the old check finds no `0x1`, reports "disabled", and skips `netsh`. This is a common state on managed/unconfigured hosts. The fix reads the `EnableFirewall` DWORD from the registry and **defaults an absent value to enabled**, which matches the effective state in every row above and is locale-independent. (The `netsh ... state` column here is just how the effective state was *measured*; parsing that text was tried as the signal but dropped — see the detection bullet — because it is localized. `Get-NetFirewallProfile ActiveStore` was also rejected — it reported `Enabled=True` even when the firewall was effectively OFF.)

**Misfire — before vs after (firewall on, `EnableFirewall` absent):**

- **Old binary:** `[WARNING] Method=netsh Action=check Details=No Windows Firewall profiles are enabled` → skips netsh → `route`, which reported success while adding nothing (IPv6 gateway + unchecked exit code).

Captura desde 2026-07-13 15-44-34

- **Fixed binary:** reads the effective state (ON) → `[INFO] Method=netsh Action=success` → inbound + outbound rules created. Misfire gone.

Captura desde 2026-07-13 15-40-08

**Blocking actually works — end-to-end connectivity test (two machines).** The agent blocked/unblocked a second machine's IP while that machine continuously probed an open TCP port (22), mimicking a stateful response (block → hold → unblock):

- **Firewall ON (netsh):** probe went `REACHABLE → BLOCKED` ~2 s after `enable`, and `BLOCKED → REACHABLE` ~2 s after `disable`. The connection is genuinely blocked and restored. ✅
- **Firewall OFF (route fallback):** the null-route did **not** block a same-subnet probe (on-link ARP delivery is not redirected by the loopback route). netsh is the only comprehensive method; the route fallback is best-effort for routed/remote attackers only — documented in-code and logged as a warning.

### Artifacts Affected

- `block-ip.exe` (Windows active-response executable)

### Configuration Changes

N/A

### Documentation Updates

Updated the Active Response executables reference (`docs/ref/modules/active-response/executables.md`), **block-ip (Windows)** section, to match the reworked behavior:
- netsh: corrected rule name (`WAZUH ACTIVE RESPONSE BLOCKED IP`), in/out bidirectional rules, and the family-aware `remoteip` prefix (`/32` IPv4, `/128` IPv6).
- Documented effective firewall-state detection (`EnableFirewall` DWORD read from the registry: GPO > SharedAccess > absent defaults to enabled), ENABLE-only.
- route fallback rewritten as a best-effort loopback null-route; removed the obsolete `ipconfig`/gateway detection; noted the same-subnet (on-link ARP) limitation, the checked exit code, and IPv4-only scope.
- DISABLE now documented as unconditional dual cleanup (removes both the netsh rule and the null-route, best-effort) instead of a fallback chain.
- Fixed the stale rule name in the netsh verification example.

### Tests Introduced

No new automated tests.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/src/active-response/src/block-ip-windows.c
+++ b/src/active-response/src/block-ip-windows.c
@@ -27,7 +27,7 @@
 
 firewall_result_t try_netsh(const char *srcip, int action, int ip_version, const char *argv0);
 firewall_result_t try_route_windows(const char *srcip, int action, int ip_version, const char *argv0);
-static bool firewall_is_effectively_on(const char *netsh_path, const char *argv0);
+static bool firewall_is_effectively_on(const char *argv0);
 
 int main(int argc, char **argv) {
     // This must always be the first instruction on Windows
@@ -93,41 +93,57 @@ int main(int argc, char **argv) {
 // WINDOWS: effective firewall-state detection
 // ============================================================================
 
-// Returns true if ANY firewall profile is effectively enabled (enforcing), read from
-// `netsh advfirewall show allprofiles state`. This reports the effective state and, in
-// particular, correctly returns ON when the local EnableFirewall value is ABSENT (the
-// Windows default).
-// On any failure to determine the state (or when the "ON" token is not recognized, e.g.
-// on a localized Windows) we return false, so try_netsh conservatively defers to the
-// route fallback rather than adding a rule that might never be enforced.
-static bool firewall_is_effectively_on(const char *netsh_path, const char *argv0) {
-    char *exec_cmd[] = {(char *)netsh_path, "advfirewall", "show", "allprofiles", "state", NULL};
-    wfd_t *wfd = wpopenv(netsh_path, exec_cmd, W_BIND_STDOUT);
-    if (!wfd) {
-        write_debug_file(argv0, "Unable to query firewall state via netsh - assuming disabled");
-        return false;
+// Reads the EnableFirewall DWORD under HKLM\<subkey>. Returns 1 (enabled), 0 (disabled),
+// or -1 (key/value absent). Forces the 64-bit view (KEY_WOW64_64KEY) because block-ip may
+// run as a 32-bit process and these keys live in the native hive.
+static int read_enable_firewall(const char *subkey) {
+    HKEY hkey;
+    if (RegOpenKeyEx(HKEY_LOCAL_MACHINE, subkey, 0, KEY_READ | KEY_WOW64_64KEY, &hkey) != ERROR_SUCCESS) {
+        return -1;
     }
 
-    char output_buf[OS_MAXSTR];
+    DWORD value = 0;
+    DWORD size = sizeof(value);
+    DWORD type = 0;
+    LONG rc = RegQueryValueEx(hkey, "EnableFirewall", NULL, &type, (LPBYTE)&value, &size);
+    RegCloseKey(hkey);
+
+    if (rc != ERROR_SUCCESS || type != REG_DWORD) {
+        return -1;
+    }
+    return value != 0 ? 1 : 0;
+}
+
+// Returns true if ANY firewall profile is effectively enabled (enforcing). Reads the
+// EnableFirewall DWORD directly from the registry.
+// Per profile: the GPO policy value wins if present, else the local SharedAccess value,
+// else it defaults to ENABLED (the Windows default). Defaulting an ABSENT value to enabled
+// is the fix for the #37388 misfire (the old check treated absent as "disabled").
+static bool firewall_is_effectively_on(const char *argv0) {
+    static const char *profiles[] = {"DomainProfile", "StandardProfile", "PublicProfile"};
+    char gpo_key[OS_MAXSTR];
+    char local_key[OS_MAXSTR];
     bool any_on = false;
 
-    while (fgets(output_buf, OS_MAXSTR, wfd->file_out)) {
-        char *p = strstr(output_buf, "State");
-        if (!p) {
-            continue;
-        }
-        p += 5;  // strlen("State")
-        while (*p == ' ' || *p == '\t') {
-            p++;
-        }
-        if ((p[0] == 'O' || p[0] == 'o') && (p[1] == 'N' || p[1] == 'n') &&
-            (p[2] == '\0' || p[2] == '\r' || p[2] == '\n' || p[2] == ' ' || p[2] == '\t')) {
+    for (int i = 0; i < 3; i++) {
+        snprintf(gpo_key, OS_MAXSTR - 1,
+                 "SOFTWARE\\Policies\\Microsoft\\WindowsFirewall\\%s", profiles[i]);
+        snprintf(local_key, OS_MAXSTR - 1,
+                 "SYSTEM\\CurrentControlSet\\Services\\SharedAccess\\Parameters\\FirewallPolicy\\%s",
+                 profiles[i]);
+
+        int gpo = read_enable_firewall(gpo_key);
+        int state = (gpo != -1) ? gpo : read_enable_firewall(local_key);
+        bool enabled = (state == -1) ? true : (state == 1);  // both absent -> Windows default ON
+        if (enabled) {
             any_on = true;
             break;
         }
     }
 
-    wpclose(wfd);
+    if (!any_on) {
+        write_debug_file(argv0, "No firewall profile is effectively enabled (registry check)");
+    }
     return any_on;
 }
 
@@ -147,11 +163,11 @@ firewall_result_t try_netsh(const char *srcip, int action, int ip_version, const
 
     // A netsh block rule is only enforced while a firewall profile is enabled. On ENABLE,
     // if the firewall is effectively OFF, defer to the route fallback instead of adding a
-    // dormant rule. We read the EFFECTIVE state via netsh (which reports ON even when the
-    // local EnableFirewall value is absent). This is ENABLE-only: on DISABLE we must
-    // always try to remove the rule, which may have been added while the firewall was
-    // enabled.
-    if (action == ENABLE_COMMAND && !firewall_is_effectively_on(netsh_path, argv0)) {
+    // dormant rule. The effective state is read from the registry (see
+    // firewall_is_effectively_on), which is locale-independent. This is ENABLE-only: on
+    // DISABLE we must always try to remove the rule, which may have been added while the
+    // firewall was enabled.
+    if (action == ENABLE_COMMAND && !firewall_is_effectively_on(argv0)) {
         log_firewall_action(argv0, LOG_LEVEL_WARNING, "netsh", "check",
                           "Windows Firewall is effectively disabled - deferring to route fallback");
         os_free(netsh_path);
--- a/src/active-response/src/block-ip-windows.c
+++ b/src/active-response/src/block-ip-windows.c
@@ -27,7 +27,7 @@
 
 firewall_result_t try_netsh(const char *srcip, int action, int ip_version, const char *argv0);
 firewall_result_t try_route_windows(const char *srcip, int action, int ip_version, const char *argv0);
-static bool firewall_is_effectively_on(const char *argv0);
+static bool firewall_is_effectively_on(const char *netsh_path, const char *argv0);
 
 int main(int argc, char **argv) {
     // This must always be the first instruction on Windows
@@ -80,8 +80,10 @@ int main(int argc, char **argv) {
         {NULL, NULL, false}  // Sentinel
     };
 
+    int ip_version = (strchr(srcip, ':') != NULL) ? 6 : 4;
+
     // Execute firewall chain with fallback
-    int result = execute_firewall_chain(methods, srcip, action, 0, argv[0]);
+    int result = execute_firewall_chain(methods, srcip, action, ip_version, argv[0]);
 
     cJSON_Delete(input_json);
     return result;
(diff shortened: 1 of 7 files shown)


src/active-response/src/block-ip-windows.c
```diff
@@ -80,6 +87,58 @@ int main(int argc, char **argv) {
...
+// `netsh advfirewall show allprofiles state`. This reports the effective state and, in
+// particular, correctly returns ON when the local EnableFirewall value is ABSENT (the
+// Windows default).
+// On any failure to determine the state (or when the "ON" token is not recognized, e.g.
+// on a localized Windows) we return false, so try_netsh conservatively defers to the
+// route fallback rather than adding a rule that might never be enforced.
+static bool firewall_is_effectively_on(const char *argv0) {
+    char *netsh_path = NULL;
+
+    if (get_binary_path("netsh.exe", &netsh_path) < 0) {
```
lchico: `try_netsh` already resolves `netsh_path` via `check_binary_available()` before calling this f
unction (line 153 below). This re-resolves the same path independently, doing a second PATH sear
ch plus duplicated "not found" logging. Consider passing the already-resolved path in as a param
eter: `firewall_is_effectively_on(netsh_path, argv0)`.

src/active-response/src/block-ip-windows.c
```diff
@@ -89,68 +148,36 @@ firewall_result_t try_netsh(const char *srcip, int action, int ip_version, const
...
-
-        if (!any_enabled) {
-            log_firewall_action(argv0, LOG_LEVEL_WARNING, "netsh", "check",
-                              "No Windows Firewall profiles are enabled");
-            os_free(netsh_path);
-            return FIREWALL_INVALID_STATE;
-        }
-    }
+    // Single-host prefix for netsh remoteip: /32 for IPv4, /128 for IPv6.
+    const char *host_prefix = (strchr(srcip, ':') != NULL) ? "128" : "32";
```
lchico: Could you check this line and also Line 304 — IPv6 address-family detection duplicated? 

Both `try_netsh` and `try_route_windows` independently sniff the address family from the strin
g via `strchr(srcip, ':')`, rather than using the `ip_version` parameter both functions already
receive — because `main()` always calls `execute_firewall_chain(methods, srcip, action, 0, argv[
0])` with a hardcoded `0` (pre-existing, not introduced by this PR). Not a blocker, but a good f
ollow-up would be to compute `ip_version` once at the call site and stop the duplicated string-s
niffing in two places.

lchico (changes requested):

src/active-response/src/block-ip-windows.c
```diff
@@ -73,84 +80,95 @@ int main(int argc, char **argv) {
...
+    if (!wfd) {
+        write_debug_file(argv0, "Unable to query firewall state via netsh - assuming disabled");
+        return false;
+    }
+
+    char output_buf[OS_MAXSTR];
+    bool any_on = false;
+
+    while (fgets(output_buf, OS_MAXSTR, wfd->file_out)) {
+        char *p = strstr(output_buf, "State");
```
lchico: This parses netsh's output for the literal English words "State"/"ON". netsh output is localized to the OS display language, so on non-English Windows (e.g. Spanish Estado/ACTIVADO) this never matches and the function always reports "off" — even when the firewall is genuinely on — silently falling back to the weaker route method. Same failure mode independently hit by Salt's win_ip.py (saltstack/salt#18576), which had to switch to WMI (Win32_NetworkAdapterConfiguration) to fix it. Suggest reading effective state from the registry instead (EnableFirewall DWORD under the GPO key, falling back to the local key, defaulting to enabled if both are absent) — locale-independent.

lchico (dismissed):

src/active-response/src/block-ip-windows.c
nbertoldo: Good catch! Change applied

src/active-response/src/block-ip-windows.c
```diff
@@ -260,108 +325,74 @@ firewall_result_t try_netsh(const char *srcip, int action, int ip_version, const
...
-            log_firewall_action(argv0, LOG_LEVEL_WARNING, "route", "check",
-                              "ipconfig.exe not found - cannot determine gateway");
-            os_free(route_path);
-            return FIREWALL_NOT_AVAILABLE;
-        }
-
-        // Query default gateway
-        char *ipconfig_cmd[] = {ipconfig_path, NULL};
-        wfd_t *wfd = wpopenv(ipconfig_path, ipconfig_cmd, W_BIND_STDOUT);
+        log_firewall_action(argv0, LOG_LEVEL_WARNING, "route", "best_effort",
```
hernanvalenzuela, line 385: 2 identical, consecutive `if (action == ENABLE_COMMAND)` blocks
that could be merged.

src/active-response/src/block-ip-windows.c
```diff
@@ -66,91 +77,141 @@ int main(int argc, char **argv) {
...
+
+    int result;
+    if (action == DISABLE_COMMAND) {
+        // Removal is NOT a fallback chain: the block may have been applied by netsh
+        // (firewall enabled) OR by the null-route (firewall off), so unconditionally remove
+        // BOTH. Each is best-effort - a missing rule/route is expected, not a failure - so
+        // cleanup never depends on netsh's (undocumented) exit code for "no rule matched".
+        firewall_result_t netsh_res = try_netsh(srcip, DISABLE_COMMAND, ip_version, argv[0]);
+        firewall_result_t route_res = try_route_windows(srcip, DISABLE_COMMAND, ip_version, argv[0]);
+        if (netsh_res == FIREWALL_SUCCESS || route_res == FIREWALL_SUCCESS) {
```
hernanvalenzuela, line 101: some thing wrong here, or I think so,  if try_netsh returns success and try_route_windows returns NOT  success, at line 90  if (netsh_res == FIREWALL_SUCCESS || route_res == FIREWALL_SUCCESS)  logged "successfully unblocked", is it correct?
could be some thing like this:
firewall_result_t netsh_res = try_netsh(srcip, DISABLE_COMMAND, ip_version, argv[0]);
 firewall_result_t route_res = (netsh_res ==FIREWALL_SUCCESS) &&   try_route_windows(srcip, DISABLE_COMMAND, ip_version, argv[0]);
xor
firewall_result_t netsh_res = try_netsh(srcip, DISABLE_COMMAND, ip_version, argv[0]);
 firewall_result_t route_res = (netsh_res !=FIREWALL_SUCCESS) &&   try_route_windows(srcip, DISABLE_COMMAND, ip_version, argv[0]);

hernanvalenzuela (commented):
PR comments added

src/active-response/src/block-ip-windows.c
nbertoldo: Done.

src/active-response/src/block-ip-windows.c
nbertoldo: The block is only ever applied by one method (netsh when the firewall is on, the null-route when it's off), so "unblocked" is correct as long as either removal took effect — if netsh removed the rule, `route` reporting "no route" is the expected/benign case, not a failure.

We deliberately call both unconditionally (not a fallback) so that even the rare state where both a rule and a route exist — same IP blocked once while the firewall was off and again while it was on — is fully cleaned.

On the alternatives: the `==` form would skip route delete whenever the block was applied via route (firewall off), orphaning the route — the exact issue just fixed for @lchico. The `!=` form is closer in intent but still orphans the route in the both-exist case, and `firewall_result_t res = (cond) && try_route(...)` is a type bug since `FIREWALL_SUCCESS == 0`, so `&&` inverts the result. Kept the unconditional dual-cleanup and clarified the comment.

hernanvalenzuela (dismissed):

rjcausarano (dismissed):
LGTM
- [x] Code changes reviewed

cborla (approved):
LGTM
- [x] Code changes reviewed

PR #37441, opened 2026-07-07
Use DisplayName and python.exe version for Microsoft Store packages

### LINKED ISSUE #37440: Syscollector reports a wrong package name for Microsoft Store packages with multiple sub-applications
## Description

Reported by a customer in wazuh/external-devel-requests#6761: Python 3.13 installed from the Microsoft Store on Windows 11 does not generate vulnerability alerts because Syscollector reports it with a wrong name and a wrong version, so no valid CPE can be generated.

**Reported by Syscollector (Wazuh 4.14.5, Windows 11 25H2):**

```sql
{
    "name": "pip (Python 3.13)",
    "version": "3.13.3824.0",
    "vendor": "Python Software Foundation",
    "location": "C:\\Program Files\\WindowsApps\\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0",
    "format": "win",
    "architecture": "x86_64"
}
```

Expected name: `Python 3.13`. The package IS collected by the Store pipeline (`getStorePackages()` → `AppxWindowsWrapper`); it is not discarded.

## Root cause

`AppxWindowsWrapper::getName()` (`src/data_provider/src/packages/appxWindowsWrapper.h`) enumerates the sub-application subkeys under `HKU\<SID>\...\AppModel\Repository\Packages\<PackageFullName>` and returns the first `Capabilities\ApplicationName` found, breaking out of the loop. The Python MSIX package bundles three sub-applications (`Pip`, `Python`, `PythonW`); `Pip` enumerates first alphabetically, so the whole package is reported as `pip (Python 3.13)`. The correct name is already available in the `DisplayName` value of the package root key (`Python 3.13`), which is never read.

Any Store package bundling several sub-applications where the primary application does not sort first is affected.

## Additional finding (separate decision needed)

The reported version (`3.13.3824.0`) is the MSIX package revision, not the product version (`3.13.3`). The product version does not exist in the AppModel registry nor in `AppxManifest.xml` (only in the executable's `VERSIONINFO` resource), so it cannot be fixed from the registry-based collector. Even with the corrected name, version-range matching can false-negative (`3.13.3824.0 > 3.13.4`); this likely needs a CTI-side translation keyed on the corrected name/vendor.

## Proposed fix

Prefer the package root `DisplayName` (plain values used directly, localized `@{...}` values resolved through the existing MrtCache lookup, raw `ms-resource:` URIs discarded), falling back to the current sub-application behavior when unavailable. Fix implemented and verified (unit tests + real Windows 11 host validation with zero regressions over 174 packages).

Darioortegaleyva:
## Description

This PR fixes the name and version reported by Syscollector for Microsoft Store (MSIX/Appx) packages, and the PYPI collector coverage for Microsoft Store Python installations. Multi-application packages were named after the first sub-application found in alphabetical order, so the Python 3.13 Store package (bundling `Pip`, `Python` and `PythonW`) was inventoried as `pip (Python 3.13)` with the MSIX package revision (`3.13.3824.0`) instead of the interpreter version. With a wrong name and version, no valid CPE can be generated and the package produces no vulnerability alerts. Additionally, the PYPI collector never scanned Store Python installations, so packages like `pip` were missing from the inventory.

**Proposed Release:** v4.14.8
**Issue:** wazuh/wazuh#37440
**Internal Reference:** wazuh/external-devel-requests#6761

## Proposed Changes

- Updated `src/data_provider/src/packages/appxWindowsWrapper.h`:
  - `getName()` now uses the package root `DisplayName` as primary name source: plain values are used directly, localized `@{...}` values are resolved through the existing MrtCache lookup, raw `ms-resource:` URIs are discarded, and any unresolved case falls back to the previous behavior (first sub-application `ApplicationName`).
  - Added a version enrichment scoped to the `PythonSoftwareFoundation.Python.*` family: the version is read from the `VERSIONINFO` resource of the packaged `python.exe` (via `PackageWindowsHelper::getProductVersion`), falling back to the MSIX revision. Scoped on purpose: PEP 514 registrations of Store apps live in the MSIX virtualized hive (not reachable under `HKEY_USERS`, which the agent reads as SYSTEM), and a generic executable-version approach corrupted 11 of 34 packages on a real host (`0.0.0.0`, commit hashes, helper-tool versions) — for Store packages the MSIX revision IS the product version, Python being the known outlier.
  - Refactored the class as `TAppxWindowsWrapper<TRegistry, TExeVersionReader>` (production alias `AppxWindowsWrapper` unchanged) to allow unit testing with injected fakes, following the NPM/PYPI collectors pattern.
- Updated `src/data_provider/src/packages/packagesWindowsParserHelper.h`: added `PackageWindowsHelper::getProductVersion(filePath)` — reads the `ProductVersion` string of the VERSIONINFO resource as file data (no execution). It lives next to the other Windows package registry helpers and is reusable by any packages collector. Note: `ie_explorer.cpp` has a similar but not equivalent function (`VS_FIXEDFILEINFO` FileVersion → `3.13.14150.1013` for python.exe, not `3.13.14`); migrating it to this helper can be a follow-up.
- Updated `src/data_provider/src/sysInfoWin.cpp`: `getPythonDirectories()` now also derives the site-packages directory of Microsoft Store Python installations from the `PackageRootFolder` of the `PythonSoftwareFoundation.Python.*` packages on the AppModel repository, so the PYPI collector inventories their packages (e.g. `pip` with its real version, `format: pypi`). Store Python cannot be located through PEP 514 for the same virtualized-hive reason above.
- Updated `src/data_provider/CMakeLists.txt` and `src/data_provider/tests/sysInfoWin/CMakeLists.txt`: added `version` (version.dll) to the `sysinfo` and unit-test link libraries.
- Updated `CHANGELOG.md`.

### Results and Evidence

The changes were validated on a Windows 11 lab VM (Build 10.0.26200, agent v4.14.5, manager v4.14.5, the issue reproduction environment) and on a second real Windows 11 host with 174 Store package keys.

Inventory before the fix (lab VM):

```sql
{
    "name": "pip (Python 3.13)",
    "version": "3.13.3824.0",
    "vendor": "Python Software Foundation",
    "location": "C:\\Program Files\\WindowsApps\\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0",
    "format": "win"
}
```

New code executed against the live registry of the lab VM (production `AppxWindowsWrapper` compiled into a standalone harness):

```bash
PS> .\appx_harness_postmove.exe | Select-String Python
Python 3.13 | 3.13.14 | Python Software Foundation | PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0
```

Version source cross-checked on the lab VM (matches the PEP 514 value written by the same installer):

```bash
PS> (Get-Item "C:\...\PythonSoftwareFoundation.Python.3.13_...\python.exe").VersionInfo.ProductVersion
3.13.14
PS> reg query "HKCU\Software\Python\PythonCore\3.13" /v Version
    Version    REG_SZ    3.13.14
```

New `getPythonDirectories()` logic executed against the live registry of the lab VM — the Store Python site-packages (containing `pip-26.1.2.dist-info`) is now located; on the second host without Store Python it returns no additional directories (no false positives):

```bash
PS> .\pythondirs.exe
=== NEW getPythonDirectories() output ===
C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0\Lib\site-packages
```

Regression check, old vs. new production code executed natively against the live registry of a real Windows 11 host (174 package keys, 34 reportable): 4 names improved (`Outlook`→`Outlook for Windows`, `Xbox Live`→`Xbox TCUI`, `Family`→`Microsoft Family`, `Realtek Audio Console`→`Realtek Audio Control`), 1 package recovered (`Windows Web Experience Pack`), zero packages lost or degraded; on hosts without Store Python the output is byte-identical.

```bash
$ diff old_output.txt new_output.txt | grep -c "^<"
0 regressions
```

Unit tests executed natively on Windows:

```bash
[==========] 12 tests from 1 test suite ran. (2 ms total)
[  PASSED  ] 12 tests.
```

### Artifacts Affected

- `src/data_provider/src/packages/appxWindowsWrapper.h`
- `src/data_provider/src/packages/packagesWindowsParserHelper.h`
- `src/data_provider/src/sysInfoWin.cpp`
- `src/data_provider/CMakeLists.txt`
- `src/data_provider/tests/sysInfoWin/CMakeLists.txt`
- `src/data_provider/tests/sysInfoWin/sysInfoWinAppx_test.cpp`
- `CHANGELOG.md`

### Configuration Changes

N/A

### Documentation Updates

N/A

### Tests Introduced

- **Scope:** Unit Tests + manual validation on real hosts

- **Details:** New `SysInfoWinAppxTest` suite (12 tests) with an in-memory registry and version-reader injected via template: multi-sub-application naming (the reported bug), no-DisplayName legacy fallback, MrtCache localization resolution, unresolved-localization fallback, raw `ms-resource:` guard, registry-error fallback, Python version enrichment from `python.exe`, missing-executable fallback, non-Python packages never using executable versions, non-Store location discard and malformed package name discard. Additionally, the production code (old and new, including the new `getPythonDirectories()` expansion) was cross-compiled with MinGW and executed against the live registries of two real Windows 11 hosts to verify the fixes and the absence of regressions.

## Review Checklist

    - [ ] Linux
    - [x] Windows
    - [ ] MAC OS X

- Memory tests for Linux

    - [ ] Coverity
    - [ ] AddressSanitizer

**General Checklist:**

- [ ] Code changes reviewed


Miguevrgo (changes requested):
Is there any way to have some code which is universal to correctly detect all packages which could suffer from this problem?

If there isn´t, I think we have too much code for some single patch for a specific package

Miguevrgo (dismissed):
Does this really work for other packages? I found the changes to look more like a refactor of the function which does the same thing but in a separate file, if I were to be wrong and this new getFileProductVersion is general, can´t we use it anywhere else to simplify the existing ones?

CHANGELOG.md
```diff
@@ -26,6 +26,12 @@ All notable changes to this project will be documented in this file.
 - Fixed Fortigate Decoder, malicious ioc rule and RHEL 8,9,10 incorrect rpm check. ([#37434](https://github.com/wazuh/wazuh/pull/37434))
 - Fixed typo in `/etc/security/opasswd` permission check on Debian 10, Ubuntu 20.04 and Ubuntu 22.04 SCA rules. ([#37652](https://github.com/wazuh/wazuh/pull/37652))
 
+### Agent
+
+#### Fixed
+
```
MarcelKemp: This section has already been defined above, so move the changelog line to where it belongs:
```suggestion
```

MarcelKemp (dismissed):

vikman90 (approved):

PR #37493, opened 2026-07-07
Warn on duplicate agent connection only when it persists

### LINKED ISSUE #37408: Investigate unexpected Agent key already in use warning
|Wazuh version|Component|Install type|
|---|---|---|
| 5.0.0-beta3 | Remoted | Manager| 

## Description
In a freshly created environment with a single connected agent, `wazuh-remoted` logs the following warning when the agent attempts to connect:

```text
wazuh-remoted: WARNING: Agent key already in use: agent ID '001' (source IP: 192.168.1.4)
```

This behavior is unexpected given that only one agent exists in a clean setup.  
We need to review the warning path in Remoted for `5.0.0-beta3`, identify the root cause, and implement a fix (or adjust logic/message severity if it is a false-positive scenario).

## Steps to reproduce
1. Deploy a fresh `5.0.0-beta3` manager environment.
2. Register and connect a single agent (ID `001`) from `192.168.1.4`.
3. Trigger an agent connection attempt (initial or reconnect).
4. Check manager logs.
5. Observe warning:
   `Agent key already in use: agent ID '001' (source IP: 192.168.1.4)`.

## Tasks
- Analyze Remoted code path that raises `Agent key already in use`.
- Determine under which conditions the warning is emitted for agent `001`.
- Verify if the warning is caused by:
  - duplicate key/session handling race,
  - stale connection/session state,
  - expected reconnect flow incorrectly flagged as conflict.
- Define expected behavior for single-agent fresh environments.
- Implement fix in Remoted:
  - prevent false-positive warning, or
  - refine validation logic and state handling.
- Add/update tests to cover the reproduced scenario.
- Validate fix against `5.0.0-beta3` branch behavior and regression risk.

## Definition of Done
- Root cause is identified and documented in the issue (or linked PR).
- Reproducible scenario no longer emits the warning incorrectly.
- Legitimate key conflicts are still detected and reported.
- Automated tests cover this scenario and pass.
- A PR with the fix is opened and linked to this issue.

If you want, I can now publish this directly as a GitHub issue in `wazuh/wazuh`.

jotacarma90:
## Description

In a fresh single-agent environment `wazuh-remoted` logs:

```
wazuh-remoted: WARNING: Agent key already in use: agent ID '001' (source IP: 192.168.1.4)
```

even though only one agent exists. Resolves #37408.

### Root cause

The warning is emitted in `HandleSecureMessage` (`src/remoted/src/secure.c`) when an agent presents a **second TCP connection** while the manager still holds the previous one (`keyentries[agentid]->sock >= 0`, different from the incoming socket) and the previous session has been active within `connection_overtake_time` (default 60s), so the idle-socket overtake path does not trigger.

This is hit on any **ungraceful reconnect** where the old socket has not been reaped yet: agent restart, VM/container stop, or a network drop that swallows the FIN/RST. It is common right after deployment. The condition self-heals within `connection_overtake_time`, but the warning reads like a shared-key / security incident when it is just the same agent reconnecting.

### Why not compare the source IP

Comparing the new connection's IP against the previous session's IP does **not** prove "same agent": two different hosts sharing the same key behind the same NAT egress present the same visible IP (and ephemeral source ports differ for a legit reconnect too). There is no reliable per-connection identity available before decryption. Keying the decision on IP would silence a genuine shared-key warning in the NAT case.

### Fix

Use the one reliable discriminator: **behavior over time**.

- A same-agent reconnect self-heals: the dead socket goes idle and is reaped by the overtake path, so the conflict clears within `connection_overtake_time`.
- A genuinely shared key keeps the previous socket alive (its `rcvd` stays fresh), so the conflict **never** clears.

Track when the duplicate-connection episode started (`conflict_ts`, new field in `keyentry`):

- While the conflict has lasted `<= connection_overtake_time` → log at **debug** (a reconnect that will self-heal).
- Once it outlives that window → **warning** (the previous session is still active = real key reuse).
- With `connection_overtake_time = 0` (overtake disabled) the previous behavior is kept: warn immediately.

`conflict_ts` is reset when the socket is overtaken (`secure.c`) or closed (`OS_DeleteSocket`). Connection handling is unchanged: the new socket is still closed and the overtake mechanism still reaps the old one. This is IP-independent, so it preserves the shared-key warning behind NAT.

### Tests

- `test_HandleSecureMessage_conflict_first_seen`: first conflict within the window → debug, no warning; asserts the episode is recorded.
- `test_HandleSecureMessage_conflict_persists`: conflict older than the window → warning.
- `test_HandleSecureMessage_different_sock` / `_2`: updated to exercise the persistent (warning) path.
- The `connection_overtake_time = 0` cases keep asserting the immediate warning.

### DIFF
--- a/src/remoted/src/secure.c
+++ b/src/remoted/src/secure.c
@@ -691,8 +691,19 @@ STATIC void HandleSecureMessage(const message_t *message, w_indexed_queue_t * co
                     mdebug2("Idle socket [%d] from agent ID '%s' will be closed.", sock_idle, keys.keyentries[agentid]->id);
 
                     keys.keyentries[agentid]->rcvd = current_ts;
+                    keys.keyentries[agentid]->conflict_ts = 0;
                 } else {
-                    mwarn("Agent key already in use: agent ID '%s' (source IP: %s)", keys.keyentries[agentid]->id, srcip);
+                    // Warn only if the conflict outlives the overtake window: a reconnect self-heals,
+                    // a shared key does not.
+                    if (keys.keyentries[agentid]->conflict_ts == 0) {
+                        keys.keyentries[agentid]->conflict_ts = current_ts;
+                    }
+
+                    if ((logr.connection_overtake_time <= 0) || (current_ts - keys.keyentries[agentid]->conflict_ts) > logr.connection_overtake_time) {
+                        mwarn("Agent key already in use: agent ID '%s' (source IP: %s)", keys.keyentries[agentid]->id, srcip);
+                    } else {
+                        mdebug2("Agent ID '%s' reconnected from '%s' before its previous session was closed.", keys.keyentries[agentid]->id, srcip);
+                    }
 
                     w_mutex_unlock(&keys.keyentries[agentid]->mutex);
                     key_unlock();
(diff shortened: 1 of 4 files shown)


Antoniogm03 (approved):
LGTM!

PR #37498, opened 2026-07-07
Fix version comparison in indexer documents updates

### LINKED ISSUE #37426: Manager/Indexer: every group (and metadata) coordination bumps the SCA document `_version` +1, causing permanent `409 version_conflict` on stateful updates
## Summary

Every time a **group coordination** (or metadata coordination) runs for an agent, the manager
issues an `_update_by_query` against that agent's state indices (e.g. `wazuh-states-sca`). The
observed effect is that the document's internal `_version` climbs by **+1 for every matched
document, on every run** — even when the update does not change the content and even when
`state.document_version` is deliberately kept at `max` (the correct behaviour for a group change).

> **Likely cause (to be confirmed by the owning team):** the `_update_by_query` request carries no
> `version_type`, so OpenSearch would fall back to **internal versioning**, which increments
> `_version` on every matched document. This matches the evidence, but we have **not** verified it
> empirically (e.g. by inspecting `_version` before/after a no-op coordination). Please treat the
> exact mechanism as a hypothesis and investigate further — it could also originate elsewhere in the
> manager's update path.

The agent, on the other hand, behaves correctly: on a **group** coordination it keeps
`document_version = max(module versions)` (no increment), and only on a **metadata change** does
it emit `max + 1`.

The two sides drift apart: the manager-side internal `_version` climbs by +1 on **every**
coordination, while the agent only ever advances `document_version` on **content changes**. After
enough coordinations, the indexer's `_version` is far ahead of any value the agent will send. The
next stateful SCA event is then permanently rejected:

```
version conflict, current version [31] is higher than the one provided [13]
```

and the check's new state is silently dropped.

---

## Root cause

The manager runs the coordination update via `executeUpdateByQuery`, which posts to:

```
POST /<indices>/_update_by_query?conflicts=proceed
```

`src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp:919`

There is **no `version_type` on this request**. Our leading hypothesis is that OpenSearch therefore
applies its default **internal versioning**, under which every document matched by the query has its
`_version` incremented by +1 as part of the update, regardless of whether the document body actually
changed. **This has not been confirmed empirically** and should be verified as part of the
investigation (see the note in the Summary) — the +1-per-coordination drift is what the evidence
shows; the internal-versioning explanation is the most probable cause but not proven.

The group-update path builds the query and executes it here:

- `src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp:846-852` (`Mode_GroupDelta`)
- `src/wazuh_modules/inventory_sync/src/inventorySyncFacade.hpp:803-818` (`Mode_MetadataDelta`)
- `InventorySyncQueryBuilder::buildGroupsUpdateQuery` — `src/wazuh_modules/inventory_sync/src/inventorySyncQueryBuilder.hpp:165`
- `InventorySyncQueryBuilder::buildMetadataUpdateQuery` — `.../inventorySyncQueryBuilder.hpp:43`

The Painless script for a **group** change only sets the *field* `state.document_version =
params.globalVersion` (= `max`, unchanged for a group change):

```painless
ctx._source.wazuh.agent.groups = params.groups;
if (ctx._source.state == null) { ctx._source.state = [:]; }
ctx._source.state.document_version = params.globalVersion;   // = max, not incremented
ctx._source.state.modified_at = params.timestamp;
```

`.../inventorySyncQueryBuilder.hpp:180-183`

So the **field** `state.document_version` is (correctly) left at `max`, yet the document's **internal
`_version`** ends up bumped +1 anyway (mechanism per the hypothesis above). The two are different
counters, and the external_gte check on subsequent agent deltas compares against the internal
`_version`.

### Why this becomes a hard `409`

Agent stateful deltas are written with `version_type: external_gte`
(`src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp:1338`), keyed on
`state.document_version`. OpenSearch requires `provided >= current _version`.

- N group coordinations → internal `_version` advances to ~`base + N`.
- Agent `document_version` stays at `max` across all those group coordinations, and only reaches
  `max + 1` on the next content change.
- If `N` coordinations happened, `max + 1` can be far below `base + N`, so `provided < current` →
  **409**, and the new state is never persisted.

The `conflicts=proceed` flag on the coordination request also means these conflicts are never
surfaced there; the drift accumulates silently until the agent delta hits it.

---

## Reproduction

1. Disable Syscollector and FIM on the agent (to make the SCA-only version delta explicit).
2. Change the agent's group to trigger a coordination. Wait until it completes. Repeat several
   times to accumulate the drift (each coordination bumps the indexer `_version` +1).
3. Generate a system change so SCA produces a stateful event on the next scan.
4. Check the manager log for a `version_conflict_engine_exception`.

---

### Agent — many group changes, `document_version` never increments

The agent correctly keeps `SCA set_version` at `max` (12) across every group coordination — a group
change must **not** increment:

```
2026/07/04 09:47:24 wm_agent_info.c:192 INFO: Starting module coordination process
2026/07/04 09:47:28 wm_sca.c:293      DEBUG: SCA set_version to 12 for 208 checks
2026/07/04 09:47:29 wm_agent_info.c:192 INFO: Synchronization coordination completed successfully

2026/07/04 09:48:29 ... INFO: Starting module coordination process
2026/07/04 09:48:36 ... DEBUG: SCA set_version to 12 for 208 checks
2026/07/04 09:48:38 ... INFO: Synchronization coordination completed successfully

2026/07/04 09:52:03 ... INFO: Starting module coordination process
2026/07/04 09:52:08 ... DEBUG: SCA set_version to 12 for 208 checks
2026/07/04 09:52:19 ... INFO: Synchronization coordination completed successfully
   (…repeated for many more coordinations, always "set_version to 12")
```

### Manager — stateful SCA update rejected with `409`

After generating a stateful SCA event (which the agent sends as `document_version = max + 1 = 13`),
the manager rejects it because the indexer `_version` has already climbed to 31:

```
2026/07/04 13:03:18 wazuh-manager-modulesd:inventory-sync (indexer-connector)[5492]
  indexerConnectorSyncImpl.hpp:356 DEBUG: Response:
  {"took":1,"errors":true,"items":[{"index":{
    "_index":"wazuh-states-sca",
    "_id":"wazuh_001_944fb36350b02efb018272fd8388dedf91549084",
    "status":409,
    "error":{"type":"version_conflict_engine_exception",
      "reason":"[wazuh_001_944fb36350b02efb018272fd8388dedf91549084]: version conflict,
                current version [31] is higher than the one provided [13]",
      "index":"wazuh-states-sca","shard":"0","index_uuid":"6_SqEC3yQZSkqWcfqG6Jtw"}}}]}
```

The agent never produced a value above 13; the gap 13 → 31 is entirely accounted for by the
manager's `_update_by_query` bumping `_version` +1 on each of the many group coordinations.

---

## Expected behaviour

A **group** coordination must not advance the SCA document's effective version at all — neither the
`state.document_version` field (already correct) **nor** the internal `_version` that the agent's
`external_gte` deltas are compared against. The manager's internal `_version` and the agent's
`document_version` must stay aligned so that a later content-driven `max + 1` is always accepted.

---

## Suggested directions (to be evaluated by the owning team)

The fix belongs on the manager side. Options, roughly in order of preference:

1. **Make the coordination `_update_by_query` use external versioning aligned with
   `document_version`.** Drive the document `_version` from `state.document_version` (external /
   `external_gte`) instead of internal versioning, so a group change that keeps
   `document_version = max` also keeps `_version = max` — no drift. This is the cleanest structural
   fix and mirrors how the agent deltas are already keyed.

2. **Skip the update entirely when nothing changes.** For a group coordination where the groups and
   `document_version` would not change, avoid issuing the `_update_by_query` so no `_version` bump
   occurs. (Partial: still bumps when groups *do* change while `document_version` stays at `max`.)

3. **Reconcile the agent's `document_version` with the indexer's `_version`** on coordination so the
   agent never sends a value below the current `_version`. (Weaker — pushes the asymmetry onto the
   agent.)

Note: because `_update_by_query` uses internal versioning, `external_gte` semantics cannot be
expressed with the current request shape; addressing this likely means changing how the coordination
update is issued (per-document versioned update, or an ingest/script path that preserves the
external version contract).

---

TomasTurina:
## Description

Fixed version conflict issues in inventory sync by changing bulk index operations to use application-level `state.document_version` instead of OpenSearch's internal `_version`. This prevents rejection of valid agent data updates when manager-side metadata changes (groups, cluster name) occur.

## Proposed Changes

Modified `IndexerConnectorSync::bulkIndex` to use scripted update operations that check `state.document_version` instead of `_version`:
- Before: Used index operation with version_type: `external_gte` (compares against `_version`)
- After: Uses update operation with Painless script (compares against `state.document_version`)

### Results and Evidence

SCA document after several group updates and stateful changes reflect the change immediately:

<details>
```
{
  "_index": "wazuh-states-sca",
  "_id": "wazuh_001_ef99b64d898d19c7d911b78b83b9b1d133b11732",
  "_score": null,
  "_source": {
    "checksum": {
      "hash": {
        "sha1": "b767510d83c07ded52e923528c188fa32c0f134f"
      }
    },
    "wazuh": {
      "cluster": {
        "name": "wazuh"
      },
      "agent": {
        "host": {
          "hostname": "tomas",
          "os": {
            "name": "Ubuntu",
            "type": "linux",
            "version": "24.04.2 LTS (Noble Numbat)",
            "platform": "ubuntu"
          },
          "architecture": "aarch64"
        },
        "name": "tomas",
        "groups": [
          "default",
          "test"
        ],
        "id": "001",
        "version": "v5.0.0"
      }
    },
    "check": {
      "result": "Failed",
      "remediation": "Create the file by running: touch /tmp/test_file",
      "condition": "all",
      "name": "Ensure the file /tmp/test_file exists.",
      "description": "Verifies the presence of the file /tmp/test_file on the system.",
      "rules": [
        "[\"f:/tmp/test_file\"]"
      ],
      "id": "90000",
      "rationale": "This is a demonstration check to confirm that a specific file is present under the /tmp directory."
    },
    "state": {
      "document_version": 4,
      "modified_at": "2026-07-07T15:06:23.902Z"
    },
    "policy": {
      "file": "cis_ubuntu24-04.yml",
      "references": [
        "[\"https://www.cisecurity.org/cis-benchmarks/\"]"
      ],
      "name": "CIS Ubuntu Linux 24.04 LTS Benchmark v1.0.0.",
      "description": "This document provides prescriptive guidance for establishing a secure configuration posture for Ubuntu Linux 24.04 LTS based on CIS benchmark for Ubuntu Linux 24.04 LTS.",
      "id": "cis_ubuntu24-04"
    }
  },
  "fields": {
    "state.modified_at": [
      "2026-07-07T15:06:23.902Z"
    ]
  },
  "sort": [
    "90000"
  ]
}
```
</details>

### Artifacts Affected

Manager

### Configuration Changes

N/A

### Tests Introduced

Updated existing unit test expectations in `indexerConnectorSync_test.cpp` to validate:
- Scripted update operation format
- `state.document_version` parameter presence
- Correct Painless script execution

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
+++ b/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
@@ -111,6 +111,43 @@ inline void appendEscapedId(std::string& bulkData, std::string_view id)
 }
 
 /**
+ * @brief Builds an update operation with Painless script for version checking
+ * @param bulkData The string to append the update operation to
+ * @param index Index name
+ * @param id Document ID (will be escaped if needed)
+ * @param version Document version to check against state.document_version
+ * @param data Document data (JSON string)
+ *
+ * The script implements external_gte behavior on state.document_version:
+ * - Updates if state.document_version is null or <= provided version
+ * - No-op if state.document_version > provided version
+ */
+inline void appendScriptedUpdate(
+    std::string& bulkData, std::string_view index, std::string_view id, std::string_view version, std::string_view data)
+{
+    // Build update operation metadata
+    bulkData.append(R"({"update":{"_index":")");
+    bulkData.append(index);
+    bulkData.append(R"(","_id":")");
+    appendEscapedId(bulkData, id);
+    bulkData.append(R"("}})");
+    bulkData.append("\n");
+
+    // Build script that checks state.document_version field
+    bulkData.append(R"({"script":{"source":")");
+    bulkData.append(
+        R"(if (ctx._source?.state?.document_version == null || ctx._source.state.document_version <= params.doc_version) { ctx._source = params.doc } else { ctx.op = 'noop' })");
+    bulkData.append(R"(","lang":"painless","params":{"doc_version":)");
+    bulkData.append(version);
+    bulkData.append(R"(,"doc":)");
+    bulkData.append(data);
+    bulkData.append(R"(}},"upsert":)");
+    bulkData.append(data);
+    bulkData.append(R"(,"scripted_upsert":true})");
+    bulkData.append("\n");
+}
+
+/**
  * @brief Validates bulk API response at document level
  * @param response The bulk API response JSON string
  * @return true if all operations succeeded (or had acceptable version conflicts), false otherwise
@@ -1317,27 +1354,21 @@ public:
         {
             processBulk();
         }
-        m_bulkData.append(R"({"index":{"_index":")");
-        m_bulkData.append(index);
+
         if (!version.empty())
         {
-            // In case the version is provided, the id must be provided too
-            if (!id.empty())
-            {
-                m_bulkData.append(R"(","_id":")");
-                appendEscapedId(m_bulkData, id);
-            }
-            else
+            // When version is provided, use update operation with Painless script
+            // to check state.document_version instead of OpenSearch's internal _version.
+            if (id.empty())
             {
                 LOG_ERROR(m_logFn, "Id must be provided if version value is provided");
                 throw IndexerConnectorException("Id must be provided if version value is provided");
             }
 
-            m_bulkData.append(R"(","version":")");
-            m_bulkData.append(version);
-            m_bulkData.append(R"(","version_type":"external_gte)");
+            appendScriptedUpdate(m_bulkData, index, id, version, data);
+
             LOG_DEBUG2(m_logFn,
-                       "Using external version %.*s for document %.*s",
+                       "Using document version %.*s for document %.*s (checking state.document_version)",
                        static_cast<int>(version.size()),
                        version.data(),
                        static_cast<int>(id.size()),
@@ -1345,20 +1376,25 @@ public:
         }
         else
         {
+            // No version provided, use standard index operation
+            m_bulkData.append(R"({"index":{"_index":")");
+            m_bulkData.append(index);
             if (!id.empty())
             {
                 m_bulkData.append(R"(","_id":")");
                 appendEscapedId(m_bulkData, id);
             }
+            m_bulkData.append(R"("}})");
+            m_bulkData.append("\n");
+            m_bulkData.append(data);
+            m_bulkData.append("\n");
+
             LOG_DEBUG2(m_logFn,
                        "No version specified for document %.*s, using default versioning",
                        static_cast<int>(id.size()),
                        id.data());
         }
-        m_bulkData.append(R"("}})");
-        m_bulkData.append("\n");
-        m_bulkData.append(data);
-        m_bulkData.append("\n");
+
         m_boundaries.push_back(m_bulkData.size());
     }
 
--- a/src/shared_modules/indexer_connector/tests/unit/indexerConnectorSync_test.cpp
+++ b/src/shared_modules/indexer_connector/tests/unit/indexerConnectorSync_test.cpp
@@ -1799,9 +1799,10 @@ TEST_F(IndexerConnectorSyncTest, BulkIndexWithVersionHandling)
     auto status = processingCompletedFuture.wait_for(std::chrono::seconds(5));
     EXPECT_EQ(status, std::future_status::ready) << "Timeout waiting for version test processing";
 
-    // Verify version is included in the bulk data for doc1
-    EXPECT_THAT(capturedBulkData, ::testing::HasSubstr(R"("version":"12345")"));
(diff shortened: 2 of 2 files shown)


juliancnn (approved):
lgtm

PR #37680, opened 2026-07-14
Stop `verify-agent-conf` from falsely warning on agent-only wodle blocks, without breaking their validation

### LINKED ISSUE #37538: Manager raises false warning when verifying agent configuration containing Syscollector
| Wazuh version | Component | Install type |
|---|---|---|
| 5.0.0-beta3 | Server | Manager | 

## Description

When verifying a group's agent configuration that includes a `syscollector` block, the manager logs a spurious warning:

```
verify-agent-conf: WARNING: The 'syscollector' module only works for the agent
```

The `syscollector` module is indeed agent-only configuration, and the manager is expected to skip it when parsing its own `wazuh-manager.conf`. However, when validating agent configuration files (e.g., `agent.conf` within a group), the manager should be able to process the configuration without raising this warning.

## Steps to Reproduce

1. Create a new agent group.
2. Add a `syscollector` configuration block to the group's `agent.conf`, for example:

```xml
<agent_config>
  <wodle name="syscollector">
    <synchronization>
      <interval>2m</interval>
    </synchronization>
  </wodle>
</agent_config>
```

3. Trigger a configuration verification (e.g., via the API or by restarting the manager).
4. Observe the warning in the manager logs.

## Expected Behavior

The manager should validate the agent configuration without producing a warning for modules that are legitimately agent-only.

## Actual Behavior

The following warning is logged:

```
2026/07/08 15:39:50 verify-agent-conf: WARNING: The 'syscollector' module only works for the agent
```

## Tasks

- [ ] Investigate the `verify-agent-conf` logic to determine why it emits this warning when processing agent-targeted configuration.
- [ ] Fix the validation logic so that agent-only modules do not trigger warnings when found in agent group configuration files.
- [ ] Add a test case covering this scenario.

## Definition of Done

- The manager no longer logs a warning for `syscollector` (or other agent-only modules) when verifying valid agent group configuration.
- Existing tests pass and a new test covering this case is added.

vikman90:
## Description
`verify-agent-conf` validates an agent group's `agent.conf` by parsing it on the manager. For agent-only wodles (`syscollector`, `command`, `aws`, `docker-listener`, `azure`, `gcp-pubsub`, `gcp-bucket`, `github`, `office365`, `ms-graph`) and for `sca`, it either logged a false "only works for the agent" warning or silently skipped the block — either way, it never checked the block's content.

The fix makes the manager parse these blocks with the same reader functions the agent uses whenever it's validating agent configuration, instead of just warning (or, for `sca`, doing nothing at all). This removes the false warning and makes real content errors inside those blocks visible again. When these modules appear in the manager's own `ossec.conf`, the warning is still raised, unchanged (SCA never warned there either, and still doesn't). `wmodules-syscollector.c` and `wmodules-sca.c` had to be re-added to the manager build (both were previously excluded there), and `wazuh_modules/src/wm_sca.c` (the file defining `WM_SCA_CONTEXT`) had to be included for the manager too, so `wm_syscollector_read`/`wm_sca_read` are available for this.

Along the way, this also surfaced and fixed a pre-existing, unrelated bug: `wm_github_destroy`/`wm_office365_destroy` logged "Module ... finished." on data teardown rather than on actual module shutdown. Left unfixed, every `verify-agent-conf` run would now log a misleading "finished" line for these two modules.

While rebuilding both the `manager` and `agent` targets to verify this fix, `make clean-deps` was found to only remove the fetched `external/` dependencies belonging to whichever `TARGET` it was last invoked with, leaving the other target's dependencies behind. That's fixed too, since it got in the way of testing both targets cleanly.

## Proposed Changes
- `src/config/CMakeLists.txt`: stopped excluding `wmodules-syscollector.c` and `wmodules-sca.c` from manager builds, so `wm_syscollector_read`/`wm_sca_read` link into the manager binary.
- `src/wazuh_modules/CMakeLists.txt`: stopped excluding `wm_sca.c` from the manager's `wazuh_modulesd_lib`, so `WM_SCA_CONTEXT` (referenced by `Read_SCA`) is available. `wm_sca_destroy()` is a trivial `free()`, so none of the module's dynamically-loaded scan engine (`sym_load`/`so_get_module_handle`) is ever exercised during config validation.
- `src/config/src/wmodules-config.c`: for all ten agent-only wodle branches plus `Read_SCA`, the manager now calls the real reader (`wm_syscollector_read`, `wm_command_read`, `wm_aws_read`, `wm_docker_read`, `wm_azure_read`, `wm_gcp_pubsub_read`, `wm_gcp_bucket_read`, `wm_github_read`, `wm_office365_read`, `wm_ms_graph_read`, `wm_sca_read`) when `agent_cfg` is set, instead of only warning (or doing nothing, for SCA). The ten wodle branches fall back to the bare warning when `agent_cfg` is unset (manager's own configuration); `Read_SCA` keeps its original no-op behavior in that case, since it never warned to begin with. `Test_WModule` now passes `agent_cfg = 1` instead of `NULL`.
- `src/config/include/config.h` / `src/config/src/config.c`: threaded the `d2` (`agent_cfg`) parameter through `Read_SCA`, `Read_GCP_pubsub`, `Read_GCP_bucket`, `Read_Github`, `Read_Office365`, `Read_MS_Graph` and their call sites.
- `src/wazuh_modules/src/wm_github.c` / `src/wazuh_modules/src/wm_office365.c`: moved the "Module ... finished." log from `destroy()` (data teardown — runs during config validation too) to the end of the module's main loop (actual shutdown).
- `src/unit_tests/config/test_wmodules-config.c` (new): cmocka tests for the fix.
- `src/unit_tests/config/CMakeLists.txt`: registered the new test for the `manager` target.
- `src/unit_tests/wazuh_modules/github/test_wm_github.c` / `src/unit_tests/wazuh_modules/office365/test_wm_office365.c`: updated the pre-existing tests for the "Module ... finished." log move — dropped the now-stale expectation from the `destroy()` teardown helper, added it where the log now actually fires (end of the enabled-module run in `wm_github_main`/`wm_office365_main`).
- `src/Makefile`: `clean-deps` now wipes everything under `external/` (except the two files tracked by git, `CMakeLists.txt` and `.gitignore`) instead of only the `TARGET`-filtered subset it removed before. Unrelated to the config-validation fix itself, but needed to reliably rebuild both `manager` and `agent` from a clean `external/` while testing this change.

### Results and Evidence
Manual run of `verify-agent-conf` against a group's `agent.conf` containing both a genuinely invalid tag and agent-only wodle blocks:
```xml
<agent_config>
  <syscheck>
    <synchronization>
    </synchronization>
  </syscheck>

  <sca>
    <synchronization>
      <no-op />
    </synchronization>
  </sca>

  <wodle name="syscollector">
    <synchronization>
      <unknown>any</unknown>
    </synchronization>
  </wodle>

  <office365>
    <enabled>yes</enabled>
    <api_auth>any</api_auth>
    <subscriptions>
    </subscriptions>
  </office365>
</agent_config>
```

The `<sca>` block's content is valid (an empty `<synchronization>` is accepted), so it produces neither a warning nor an error in either run below — its presence here just confirms it no longer causes a crash or a spurious message now that `wm_sca_read` runs on the manager too. The `test_Test_WModule_sca_*` unit tests are what specifically exercise SCA's invalid-content path.

**Before the fix** — false-positive warnings for `syscollector`/`office365`, and the actual error (`<unknown>` is not a valid tag under `<synchronization>`, `<subscriptions>` is empty) went completely unnoticed; `verify-agent-conf` reported `OK`:
```
verify-agent-conf: Verifying [etc/shared/default/agent.conf]
2026/07/10 10:38:24 verify-agent-conf: WARNING: The 'syscollector' module only works for the agent
2026/07/10 10:38:24 verify-agent-conf: WARNING: The 'office365' module only works for the agent
verify-agent-conf: OK
```

**After the fix** — no false-positive warning, and the genuine content errors are now caught and reported:
```
verify-agent-conf: Verifying [etc/shared/default/agent.conf]
2026/07/14 19:21:33 wazuh-modulesd:sca: INFO: Could not open the default SCA ruleset folder 'ruleset/sca/': No such file or directory
2026/07/14 19:21:33 wazuh-modulesd:sca: WARNING: (1230): Invalid element in the configuration: 'no-op'.
2026/07/14 19:21:33 verify-agent-conf: WARNING: (1230): Invalid element in the configuration: 'unknown'.
2026/07/14 19:21:33 verify-agent-conf: ERROR: Empty content for tag 'subscriptions' at module 'office365'.
2026/07/14 19:21:33 verify-agent-conf: ERROR: (1202): Configuration error at 'etc/shared/default/agent.conf'.
2026/07/14 19:21:33 verify-agent-conf: ERROR: (1207): WModule remote configuration in 'etc/shared/default/agent.conf' is corrupted
```

**After the fix, with the same blocks but valid content** — no warning, no error, and no misleading "Module ... finished." line (the module never actually ran on the manager):
```
verify-agent-conf: Verifying [etc/shared/default/agent.conf]
verify-agent-conf: OK
```

`make clean-deps`, before/after the `src/Makefile` fix, run against an `external/` populated for both `manager` and `agent`:
```
$ make clean-deps
find external -mindepth 1 -maxdepth 1 ! -name 'CMakeLists.txt' ! -name '.gitignore' -exec rm -rf {} +
rm -rf shared_modules/http-request/* shared_modules/http-request/.??* external/wcs-flat-files

$ ls -la external/
total 68
-rw-r--r--  1 root root    30 external/.gitignore
-rw-r--r--  1 root root 53602 external/CMakeLists.txt
```
Only the two git-tracked files remain, regardless of which `TARGET` the removed dependencies had originally been fetched for.

### Artifacts Affected
`wazuh-manager` package (`libconfig.a`, `wazuh-modulesd_lib` / `wazuh-modulesd`, `verify-agent-conf`). The `clean-deps` change affects build tooling only, not any shipped binary or package.

### Configuration Changes
None.

### Documentation Updates
None.

### Tests Introduced
`src/unit_tests/config/test_wmodules-config.c` (new):
- `test_Test_WModule_syscollector_no_warning`
- `test_Test_WModule_syscollector_invalid_content_is_reported`
- `test_Test_WModule_github_no_warning`
- `test_Test_WModule_github_invalid_content_is_reported`
- `test_Test_WModule_sca_valid_content_is_accepted`
- `test_Test_WModule_sca_invalid_content_is_reported`
- `test_Read_WModule_syscollector_warns_without_agent_cfg`

Updated to match the log move (see Proposed Changes):
- `src/unit_tests/wazuh_modules/github/test_wm_github.c`
- `src/unit_tests/wazuh_modules/office365/test_wm_office365.c`

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/src/config/CMakeLists.txt
+++ b/src/config/CMakeLists.txt
@@ -42,13 +42,6 @@ endif()
 # Collect config source files
 file(GLOB CONFIG_CORE_SRC "${CMAKE_CURRENT_SOURCE_DIR}/src/*.c")
 
-# Remove agent-only config modules from server builds
-if(NOT "${TARGET}" STREQUAL "winagent" AND NOT "${TARGET}" STREQUAL "agent")
-  list(REMOVE_ITEM CONFIG_CORE_SRC
-    "${CMAKE_CURRENT_SOURCE_DIR}/src/wmodules-sca.c"
-    "${CMAKE_CURRENT_SOURCE_DIR}/src/wmodules-syscollector.c")
-endif()
-
 # Remove manager-only config parsers from agent builds
 if("${TARGET}" STREQUAL "agent" OR "${TARGET}" STREQUAL "winagent")
   list(REMOVE_ITEM CONFIG_CORE_SRC
--- a/src/Makefile
+++ b/src/Makefile
@@ -863,7 +863,12 @@ ifneq ($(wildcard external/cpython/*),)
 endif
 
 clean-deps:
-	rm -rf $(EXTERNAL_DIR) $(EXTERNAL_TAR) shared_modules/http-request/* shared_modules/http-request/.??* $(INDEXER_PLUGINS_DIR) $(WCS_FLAT_DIR)
+# EXTERNAL_RES/EXTERNAL_DIR are filtered by TARGET, so removing only those
+# paths leaves other targets' fetched deps behind. Wipe everything under
+# external/ instead, except the two files tracked by git (CMakeLists.txt
+# and .gitignore), regardless of which TARGET this is invoked with.
+	find external -mindepth 1 -maxdepth 1 ! -name 'CMakeLists.txt' ! -name '.gitignore' -exec rm -rf {} +
+	rm -rf shared_modules/http-request/* shared_modules/http-request/.??* $(WCS_FLAT_DIR)
 
 clean-internals:
 # Wazuh binaries and libraries
--- a/src/config/include/config.h
+++ b/src/config/include/config.h
@@ -53,7 +53,7 @@ int Read_Remote(const OS_XML *xml,XML_NODE node, void *d1, void *d2);
 int Read_Client(const OS_XML *xml, XML_NODE node, void *d1, void *d2);
 int Read_ClientBuffer(XML_NODE node, void *d1, void *d2);
 int Read_WModule(const OS_XML *xml, xml_node *node, void *d1, void *d2);
-int Read_SCA(const OS_XML *xml, xml_node *node, void *d1);
+int Read_SCA(const OS_XML *xml, xml_node *node, void *d1, void *d2);
 int Read_AGENT_INFO(const OS_XML* xml, xml_node* node, void* d1);
 
 /**
@@ -68,16 +68,18 @@ int Read_Client_Shared(XML_NODE node, void *d1);
  * @param xml XML object
  * @param node XML node to analyze
  * @param d1 Pub/Sub configuration structure
+ * @param d2 Agent configuration validation flag
  */
-int Read_GCP_pubsub(const OS_XML *xml, xml_node *node, void *d1);
+int Read_GCP_pubsub(const OS_XML *xml, xml_node *node, void *d1, void *d2);
 
 /**
  * @brief Read the configuration for a Google Cloud bucket
  * @param xml XML object
  * @param node XML node to analyze
  * @param d1 Bucket configuration structure
+ * @param d2 Agent configuration validation flag
  */
-int Read_GCP_bucket(const OS_XML *xml, xml_node *node, void *d1);
+int Read_GCP_bucket(const OS_XML *xml, xml_node *node, void *d1, void *d2);
 
 #ifndef WIN32
 int Read_Authd(const OS_XML *xml, XML_NODE node, void *d1, void *d2);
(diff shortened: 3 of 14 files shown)


src/config/CMakeLists.txt
```diff
@@ -45,8 +45,7 @@ file(GLOB CONFIG_CORE_SRC "${CMAKE_CURRENT_SOURCE_DIR}/src/*.c")
 # Remove agent-only config modules from server builds
 if(NOT "${TARGET}" STREQUAL "winagent" AND NOT "${TARGET}" STREQUAL "agent")
   list(REMOVE_ITEM CONFIG_CORE_SRC
-    "${CMAKE_CURRENT_SOURCE_DIR}/src/wmodules-sca.c"
-    "${CMAKE_CURRENT_SOURCE_DIR}/src/wmodules-syscollector.c")
+    "${CMAKE_CURRENT_SOURCE_DIR}/src/wmodules-sca.c")
```
TomasTurina: Don't we need to do this with the SCA configuration as well?

TomasTurina (approved):

PR #37345, opened 2026-07-01
Lower log level of transient queue send failures in modulesd

lchico:
# Description

During agent startup and restart transitions, `wazuh-agentd` briefly re-creates `queue/sockets/queue` while modulesd submodules (agent-info and SCA) are still mid-flight attempting to send events. Both modules already perform a blocking, infinite-retry reconnect (`StartMQPredicated(..., INFINITE_OPENQ_ATTEMPTS, shutdown_predicate)`), so the condition is always transient and self-healing — but the initial failure was logged at `ERROR` level, producing misleading noise. Additionally, the SCA path was missing the shutdown-predicate guard on the first `SendMSGPredicated` call, causing it to attempt a reconnect even during a clean shutdown. Resolves #37303.

## Proposed Changes

- `src/wazuh_modules/src/wm_agent_info.c` — in `wm_agent_info_send_stateless()`, downgrade the initial send-failure log from `merror("Error sending message to queue")` to `mdebug1("Failed to send message to queue, attempting reconnection.")`. The post-reconnect failure (a genuine, non-recoverable condition) retains `merror`.
- `src/wazuh_modules/src/wm_sca.c` — in `wm_sca_send_stateless()`, add the missing `wm_sca_is_shutting_down()` guard immediately after the first `SendMSGPredicated` failure, so the reconnect loop is not entered during a clean stop. Downgrade the initial failure log to `mdebug1` with the same neutral message. Remove the now-unreachable `merror("Cannot restart SCA message queue")` line — `StartMQPredicated` with `INFINITE_OPENQ_ATTEMPTS` only returns `-1` when the shutdown predicate is true, so that log would only appear on shutdown (misleading) and is replaced by a silent early return. The post-reconnect failure retains `merror` (and a matching shutdown-guard to suppress it on clean shutdown).

### Results and Evidence

The binary string diff below confirms the log strings were changed in the PR build, while the stable baseline retains the original strings.

<details><summary>Before fix — stable nightly binary (wazuh-modulesd 5.0.0-0 baseline)</summary>

```bash

```

</details>

<details><summary>After fix — PR branch binary (wazuh-modulesd 5.0.0-0, built from fix/37303-lower-log-level-queue-send-failure)</summary>

```bash

```

</details>

<details><summary>Socket-removal trigger — sync-protocol path fires at DEBUG; no ERROR from legacy agent-info/SCA path</summary>

```bash
```

</details>

### Artifacts Affected

- `wazuh-modulesd` (part of the `wazuh-agent` package)

[...]

Issue #37515, opened 2026-07-08
Investigate flaky syscollector test


jpcerrone:
|Wazuh version|Component|Install type|Install method|Platform|
| main | Syscollector | Agent | Sources | Windows |

The `test_syscollector_deactivation` test is failing sporadically with the following message:
```
File C:\Program Files (x86)\ossec-agent\ossec.log does not exist.
```

We need to investigate and propose a fix for it

Example of a failing run: https://github.com/wazuh/wazuh/actions/runs/28898152397/job/85865728180?pr=37212

Other flaky Syscollector tests:

```
test_syscollector\test_configuration\test_syscollector_configuration.py . [  2%]
...................................F.F                                   [100%]

================================== FAILURES ===================================
_____ test_syscollector_collectors_disabled[packages_processes_disabled] ______
test_syscollector\test_configuration\test_syscollector_configuration.py:556: in test_syscollector_collectors_disabled
    assert log_monitor.callback_result, "DataClean notification should be started for disabled collectors with data"
E   AssertionError: DataClean notification should be started for disabled collectors with data
E   assert None
E    +  where None = <wazuh_testing.tools.monitors.file_monitor.FileMonitor object at 0x000001DDE4BA6A10>.callback_result
---------------------------- Captured stdout setup ----------------------------
---------------------

_______ test_syscollector_collectors_disabled[all_collectors_disabled] ________
test_syscollector\test_configuration\test_syscollector_configuration.py:556: in test_syscollector_collectors_disabled
    assert log_monitor.callback_result, "DataClean notification should be started for disabled collectors with data"
E   AssertionError: DataClean notification should be started for disabled collectors with data
E   assert None
E    +  where None = <wazuh_testing.tools.monitors.file_monitor.FileMonitor object at 0x000001DDE50A8710>.callback_result
---------------------------- Captured stdout setup ----------------------------
```

```
test_syscollector\test_configuration\test_syscollector_configuration.py . [  2%]
..................................F.F.                                   [100%]

================================== FAILURES ===================================
__________ test_syscollector_collectors_disabled[packages_disabled] ___________
test_syscollector\test_configuration\test_syscollector_configuration.py:556: in test_syscollector_collectors_disabled
    assert log_monitor.callback_result, "DataClean notification should be started for disabled collectors with data"
E   AssertionError: DataClean notification should be started for disabled collectors with data
E   assert None
[...]

PR #37689, opened 2026-07-14
Report modulesSync failure as debug during agent shutdown

### LINKED ISSUE #37554: Windows agent tries to sync sca module after module stopped
|---|---|---|---|---|
| 5.0.0-beta4-9e0b2f2 | SCA | Agent | Packages | Windows 11<br>Windows 10<br>Windows Server 2019<br>Windows Server 2022<br>Windows Server 2025 |

The following error messages were found in the 2026-07-10 nightly upgrade tests:

```
Log in windows-11-amd64 (Agent-windows):
2026/07/09 21:50:41 wazuh-agent: ERROR: At modulesSync(): Unable to sync module 'sca': (-1)

Log in windows-10-amd64 (Agent-windows):
2026/07/10 04:50:42 wazuh-agent: ERROR: At modulesSync(): Unable to sync module 'sca': (-1)

Log in windows_server-2019-amd64 (Agent-windows):
2026/07/10 04:50:42 wazuh-agent: ERROR: At modulesSync(): Unable to sync module 'sca': (-1)

Log in windows_server-2022-amd64 (Agent-windows):
2026/07/10 04:50:42 wazuh-agent: ERROR: At modulesSync(): Unable to sync module 'sca': (-1)

Log in windows_server-2025-amd64 (Agent-windows):
2026/07/10 04:50:43 wazuh-agent: ERROR: At modulesSync(): Unable to sync module 'sca': (-1)
```

Investigating the Windows 11 agent logs, I found that the module was stopped before those message appeared

```
2026/07/09 21:47:48 wazuh-modulesd:sca: INFO: SCA module stopped.
2026/07/09 21:47:48 wazuh-modulesd:syscollector: INFO: Stop received for Syscollector.
2026/07/09 21:47:48 wazuh-modulesd:sca: WARNING: SCA synchronization failed.
2026/07/09 21:47:48 wazuh-agent: ERROR: At modulesSync(): Unable to sync module 'sca': (-1)
2026/07/09 21:47:48 wazuh-agent: ERROR: At modulesSync(): Unable to sync module 'sca': (-1)
2026/07/09 21:47:49 wazuh-modulesd:syscollector: WARNING: Syscollector synchronization failed.
```

The upgrade started at `04:46:19` (`21:46:19`)

```
[2026-07-10 04:46:19,489] [INFO] [utils.py] [T1-agent-windows-11-amd64]: Running command: Start-Process msiexec.exe -ArgumentList '/x C:\Users\wazuh-user\packages\wazuh-agent-5.0.0-latest.msi /qn' -Wait; Start-Process -Filepath 'C:\Users\wazuh-user\packages\wazuh-agent-5.0.0-latest.msi' -ArgumentList '/q' -Wait
```

And ended at `04:46:31` (`21:46:31`)

```
[2026-07-10 04:46:31,463] [INFO] [utils.py] [T1-agent-windows-11-amd64]: Command finished: Start-Process msiexec.exe -ArgumentList '/x C:\Users\wazuh-user\packages\wazuh-agent-5.0.0-latest.msi /qn' -Wait; Start-Process -Filepath 'C:\Users\wazuh-user\packages\wazuh-agent-5.0.0-latest.msi' -ArgumentList '/q' -Wait
```

Those messages were previously reported in https://github.com/wazuh/wazuh/issues/35181 (5.0.0 beta1) and were supposedly fixed (https://github.com/wazuh/wazuh/issues/35181#issuecomment-4195782668)

Those logs are not reported in https://github.com/wazuh/wazuh/issues/37410 as known messages

---

## Check list

- Release/build under test: `5.0.0-beta4-9e0b2f2`
- Platform: `Windows`
- Install method: `packages prod 4.14.6 -> upgrade 5.0.0`
- Test scenario: `upgrade`
- Full log excerpt with timestamps: 
  - [agent-windows-11-amd64_agent.log](https://github.com/user-attachments/files/29882741/agent-windows-11-amd64_agent.log)
  - [agent-windows-10-amd64_agent.log](https://github.com/user-attachments/files/29882742/agent-windows-10-amd64_agent.log)
  - [agent-windows-server-2019-amd64_agent.log](https://github.com/user-attachments/files/29882748/agent-windows-server-2019-amd64_agent.log)
  - [agent-windows-server-2022-amd64_agent.log](https://github.com/user-attachments/files/29882749/agent-windows-server-2022-amd64_agent.log)
  - [agent-windows-server-2025-amd64_agent.log](https://github.com/user-attachments/files/29882752/agent-windows-server-2025-amd64_agent.log)
  - [aio_manager.log](https://github.com/user-attachments/files/29882758/aio_manager.log)
- Component that emitted the message: `wazuh-agent`
- Link to this epic: https://github.com/wazuh/wazuh/issues/37410

anromerom:
## Description

Closes #37554

During a normal agent upgrade or restart, `wazuh-modulesd` logged an `ERROR` for an expected shutdown-time condition:

```sql
wazuh-modulesd:sca: INFO: SCA module stopped.
wazuh-agent: ERROR: At modulesSync(): Unable to sync module 'sca': (-1)
```

The manager can push a synchronization command for a module a moment after the agent has already begun stopping that module. The module is no longer running, so its `sync()` returns `-1`; the `modulesSync()` dispatcher retries `WM_MAX_ATTEMPTS` times and, since the module stays stopped, logs a hard `ERROR`. This is transient, the agent recovers, and no data is lost, but it surfaced as an unknown `ERROR` in the nightly upgrade tests.

>Note: While reviewing the dispatcher, a second, pre-existing bug turned up in the same retry loop: `usleep(retry * WM_MAX_WAIT)` treats a millisecond constant as microseconds, so the backoff between attempts never really waits (all three retries fire within ~3 ms instead of the intended ~3 s). That bug predates this PR and is unrelated to the shutdown race fixed here, so it's left out of scope and tracked separately (see [review comment](https://github.com/wazuh/wazuh/pull/37689#discussion_r3598525992)).

### 1. Do not report an expected shutdown-time failure as an error

The dispatcher is gated on the agent-wide shutdown flag (`wm_shutdown_requested`) in `modulesSync()`:

- Stop retrying once a shutdown is in progress. The target module is being torn down on purpose, so further attempts only delay shutdown while always failing.
- When the failure happens during shutdown, log it at `DEBUG` (worded as an expected skip) instead of `ERROR`. A genuine failure while the agent is running keeps the `ERROR` level so real problems remain visible.

This mirrors the already-merged sibling #37345 ("Lower log level of transient queue send failures in modulesd") and the approach used in #37325 (gate the log level on the stopping condition).

### 2. Set the shutdown flag on the Windows shutdown path

As raised in review, the gate above was unreachable on the platform the issue is about: `wm_shutdown_requested` was only ever set by the POSIX writers (the standalone `wazuh-modulesd` `SIGTERM` handler in `main.c`, and `wm_control_stop()`, whose module is only compiled on Linux/macOS/BSD). On a Windows agent the modules run as threads inside `wazuh-agent.exe`, shutdown flows through `OssecServiceCtrlHandler(SERVICE_CONTROL_STOP)` -> `stop_wmodules()`, and neither path touched the flag, so the `ERROR` would have kept appearing exactly as reported in #37554.

`stop_wmodules()` (`src/win32/win_utils.c`) now sets `wm_shutdown_requested = 1` before iterating the modules' `stop()` callbacks, mirroring what the POSIX `SIGTERM` handler does. This is what actually engages the fix on Windows.

### 3. Retry backoff timing bug (out of scope)

`WM_MAX_WAIT` is documented as milliseconds, but its value is passed straight to `usleep()`, which takes microseconds:

```c
usleep(retry * WM_MAX_WAIT);    // 500 us, not 500 ms
```

The loop is therefore sleeping 500 us / 1 ms / 1.5 ms between attempts instead of the intended 500 ms / 1000 ms / 1500 ms, so all three retries fire within ~3 ms. A module that needs a moment to become ready gets no real grace period, which defeats the point of retrying at all. This was flagged in review ([discussion](https://github.com/wazuh/wazuh/pull/37689#discussion_r3598525992)).

This bug predates this PR and is independent of the shutdown race it fixes, so it is **not** addressed here. It will be tracked and fixed in a separate issue/PR.

### Results and Evidence

`wazuh-modulesd` dispatcher behavior after the change:

```sql
# During agent shutdown/restart (expected corner case) - was ERROR, now DEBUG
wazuh-modulesd: DEBUG: At modulesSync(): skipping sync for module 'sca' (-1): the agent is shutting down.

# Genuine failure while the agent is running - unchanged, still ERROR
wazuh-agent: ERROR: At modulesSync(): Unable to sync module 'sca': (-1)
```

The exact race (the manager pushing an `sca_sync` right after the SCA module stopped during an upgrade restart) is timing-dependent and hard to force on demand. Unit tests were added to exercise both branches of the log-level decision directly (see Tests Introduced).

### Artifacts Affected

- `wazuh-modulesd` (agent) — the `modulesSync()` dispatcher in `src/wazuh_modules/src/wmodules.c`.
- `wazuh-agent.exe` (Windows agent) — `stop_wmodules()` in `src/win32/win_utils.c` now raises the shutdown flag.
- Unit test build — `src/unit_tests/wazuh_modules/CMakeLists.txt` now registers the `wmodules` suite for the agent target (see Tests Introduced).

Agent-only components. No manager artifacts, no configuration files, no packages changed.

### Configuration Changes

- None.

### Documentation Updates

- None.

### Tests Introduced

Added to the `wmodules` unit test suite (`src/unit_tests/wazuh_modules/wmodules/test_wmodules.c`):

- `test_modules_sync_failure_while_running`: a sync command that keeps failing while the agent is running must log at `ERROR` after exhausting the retries. The test also asserts one `"WModules is not ready. Retry N"` debug per attempt, so the full retry sequence is exercised.
- `test_modules_sync_failure_during_shutdown`: the same failure with `wm_shutdown_requested = 1` must not retry (no `Retry` debug is expected) and must log the single expected-skip `DEBUG` line, not `ERROR`.

The test file wraps `_merror` and `_mdebug1` to assert the exact level and message per case, which is also what pins down the retry-vs-no-retry difference between the two branches.

The `wmodules` suite was previously only registered for `TARGET=manager`, so these tests never ran for the component they cover. `src/unit_tests/wazuh_modules/CMakeLists.txt` now adds the suite to the agent target as well, where it builds and passes.

The suite is not wired for `winagent`: verifying that requires the full MinGW cross-build plus wine, and an unverified build-system change seemed worse than a scoped one. The Windows behavior itself (change 2) is validated on a live Windows agent instead (restart/upgrade flow of #37554).

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/src/wazuh_modules/src/wmodules.c
+++ b/src/wazuh_modules/src/wmodules.c
@@ -365,13 +365,22 @@ int modulesSync(char* args, size_t length) {
 
         ++retry;
 
-        if (retry > WM_MAX_ATTEMPTS) {
+        // Stop retrying if the agent is shutting down: the target module is being torn down on
+        // purpose, so further attempts would only delay shutdown while always failing.
+        if (retry > WM_MAX_ATTEMPTS || wm_shutdown_requested) {
             break;
         }
     } while (ret != 0);
 
     if (ret) {
-        merror("At modulesSync(): Unable to sync module '%s': (%d)", cur_module ? cur_module->tag : "",  ret);
+        if (wm_shutdown_requested) {
+            // Expected during agent shutdown/restart: a sync command reached a module that was already
+            // stopping. Report it as a debug breadcrumb instead of an error.
+            mdebug1("At modulesSync(): skipping sync for module '%s' (%d): the agent is shutting down.",
+                    cur_module ? cur_module->tag : "", ret);
+        } else {
+            merror("At modulesSync(): Unable to sync module '%s': (%d)", cur_module ? cur_module->tag : "",  ret);
+        }
     }
     return ret;
 }
--- a/src/unit_tests/wazuh_modules/CMakeLists.txt
+++ b/src/unit_tests/wazuh_modules/CMakeLists.txt
@@ -16,6 +16,7 @@ if(${TARGET} STREQUAL "manager")
 endif()
 
 if(${TARGET} STREQUAL "agent")
+  add_subdirectory(wmodules)
   add_subdirectory(aws)
   add_subdirectory(azure)
   add_subdirectory(command)
--- a/src/unit_tests/wazuh_modules/wmodules/CMakeLists.txt
+++ b/src/unit_tests/wazuh_modules/wmodules/CMakeLists.txt
@@ -32,7 +32,7 @@ include(${SRC_FOLDER}/unit_tests/wrappers/wazuh/shared/shared.cmake)
 
 # Generate wmodules tests
 list(APPEND wmodules_names "test_wmodules")
-list(APPEND wmodules_flags "-Wl,--wrap,w_query_agentd")
+list(APPEND wmodules_flags "-Wl,--wrap,w_query_agentd -Wl,--wrap,_merror -Wl,--wrap,_mdebug1")
 
 # Compilig tests
 list(LENGTH wmodules_names count)
--- a/src/unit_tests/wazuh_modules/wmodules/test_wmodules.c
+++ b/src/unit_tests/wazuh_modules/wmodules/test_wmodules.c
@@ -23,16 +23,29 @@ static size_t echo(void * module, char * query, char ** output) {
     return strlen(query);
 }
 
+// Sync callback that always fails, to exercise the modulesSync() failure path.
+static int sync_fail(const char * args, size_t length) {
+    (void)args;
+    (void)length;
+    return -1;
+}
+
 static int setup_modules(void ** state) {
     static wm_context CONTEXTS[] = {
         { .name = "A", .query = echo },
         { .name = "B", .query = NULL },
+        { .name = "sca", .sync = sync_fail },
     };
 
     wmodules = calloc(1, sizeof(wmodule));
     wmodules->context = &CONTEXTS[0];
     wmodules->next = calloc(1, sizeof(wmodule));
     wmodules->next->context = &CONTEXTS[1];
+    wmodules->next->next = calloc(1, sizeof(wmodule));
+    wmodules->next->next->context = &CONTEXTS[2];
+    wmodules->next->next->tag = "sca";
+
+    wm_shutdown_requested = 0;
 
     *state = NULL;
     return 0;
@@ -40,8 +53,11 @@ static int setup_modules(void ** state) {
 
 static int teardown_modules(void ** state) {
     free(*state);
+    free(wmodules->next->next);
     free(wmodules->next);
     free(wmodules);
+
+    wm_shutdown_requested = 0;
 
     return 0;
 }
@@ -107,6 +123,39 @@ static void test_module_query_echo(void ** state) {
     assert_int_equal(n, 4);
 }
 
+// A sync command that keeps failing while the agent is running is a real error: after exhausting
+// the retries, modulesSync() must log it at ERROR level.
+static void test_modules_sync_failure_while_running(void ** state) {
+    (void)state;
+    char input[] = "sca_sync payload";
+
+    // One "not ready" debug per retry (WM_MAX_ATTEMPTS attempts after the first one).
+    expect_string(__wrap__mdebug1, formatted_msg, "WModules is not ready. Retry 1");
+    expect_string(__wrap__mdebug1, formatted_msg, "WModules is not ready. Retry 2");
+    expect_string(__wrap__mdebug1, formatted_msg, "WModules is not ready. Retry 3");
+    expect_string(__wrap__merror, formatted_msg, "At modulesSync(): Unable to sync module 'sca': (-1)");
+
+    int ret = modulesSync(input, sizeof(input));
+
+    assert_int_equal(ret, -1);
+}
+
+// The same failure during agent shutdown is expected (the module was already stopped on purpose):
+// modulesSync() must not retry and must report it at DEBUG level, not ERROR.
+static void test_modules_sync_failure_during_shutdown(void ** state) {
+    (void)state;
+    char input[] = "sca_sync payload";
+
+    wm_shutdown_requested = 1;
+
+    expect_string(__wrap__mdebug1, formatted_msg,
+                  "At modulesSync(): skipping sync for module 'sca' (-1): the agent is shutting down.");
+
+    int ret = modulesSync(input, sizeof(input));
+
+    assert_int_equal(ret, -1);
+}
+
 int main() {
     const struct CMUnitTest tests[] = {
         cmocka_unit_test_setup_teardown(test_find_module_found, setup_modules, teardown_modules),
@@ -115,6 +164,8 @@ int main() {
         cmocka_unit_test_setup_teardown(test_module_query_no_module, setup_modules, teardown_modules),
         cmocka_unit_test_setup_teardown(test_module_query_no_queries, setup_modules, teardown_modules),
         cmocka_unit_test_setup_teardown(test_module_query_echo, setup_modules, teardown_modules),
+        cmocka_unit_test_setup_teardown(test_modules_sync_failure_while_running, setup_modules, teardown_modules),
+        cmocka_unit_test_setup_teardown(test_modules_sync_failure_during_shutdown, setup_modules, teardown_modules),
     };
 
     return cmocka_run_group_tests(tests, NULL, NULL);
--- a/src/win32/win_utils.c
+++ b/src/win32/win_utils.c
@@ -104,6 +104,11 @@ void *win_module_thread(void *arg)
 
 void stop_wmodules()
 {
+    // Signal the agent-wide shutdown so dispatchers (e.g. modulesSync) skip work
+    // that targets modules already being torn down. On POSIX the modulesd SIGTERM
+    // handler sets this; on Windows shutdown flows through here instead.
+    wm_shutdown_requested = 1;
+
     wmodule * cur_module;
     for (cur_module = wmodules; cur_module; cur_module = cur_module->next) {
         if (cur_module->context->stop) {


rjcausarano (dismissed):
LGTM!

- [x] Code changes reviewed

src/wazuh_modules/src/wmodules.c
hernanvalenzuela, line 352: I'm sure this is before your PR: `usleep` takes a value in microseconds, but
`#define WM_MAX_WAIT     500                           // Max. wait between attempts in milliseconds.`
Which one is correct?

src/wazuh_modules/src/wmodules.c
```diff
@@ -349,7 +349,7 @@ int modulesSync(char* args, size_t length) {
 
     do {
         if (retry > 0) {
-            usleep(retry * WM_MAX_WAIT);
+            usleep(retry * WM_MAX_WAIT * 1000);         // WM_MAX_WAIT is in milliseconds; usleep expects microseconds.
```
hernanvalenzuela: I prefer use a const value, because it calculates each time

nbertoldo (changes requested):
Nice, clean change and the unit tests exercise both branches well. On POSIX this is correct: `wm_shutdown_requested` is set before modules are torn down (by the `wazuh-modulesd` `SIGTERM` handler in `main.c:279`, and also by `wm_control_stop()` in `wm_control.c:49`), so the flag is reliably `1` when the late sync arrives.

**However, I don't think this fixes the reported issue on the affected platform.** #37554 is Windows-only, and the fix gates entirely on `wm_shutdown_requested`, which is **never set to `1` on a Windows agent**. Neither of the two writers is reachable there:

- `main.c`'s `SIGTERM` handler is the standalone `wazuh-modulesd` daemon path. On Windows the modules run as threads inside `wazuh-agent.exe` (`win_utils.c: local_start`), so that `main()` never runs.
- `wm_control_stop()` is only compiled in on POSIX: the `control` module is registered under `#if defined(__linux__) || defined(__MACH__) || defined(FreeBSD) || defined(OpenBSD)` in `wmodules.c:149-152`, which excludes Windows.
- Windows shutdown goes through `OssecServiceCtrlHandler(SERVICE_CONTROL_STOP)` -> `stop_wmodules()` (`win_utils.c:105`), which only calls each module's `stop()` and never touches the flag.

So on Windows `wm_shutdown_requested` stays `0` when `modulesSync()` processes the late `sca_sync`, the `merror` branch is always taken, and `ERROR: At modulesSync(): Unable to sync module 'sca': (-1)` will keep showing up exactly as before. The new unit tests pass only because they set the flag manually, which hides this gap.

**Suggested fix:** set the flag at the start of the Windows shutdown path so the gate actually engages, mirroring what `main.c` does on POSIX:

```c
void stop_wmodules()
{
    // Signal the agent-wide shutdown so dispatchers (e.g. modulesSync) can skip
    // work that targets modules already being torn down. On POSIX this is set by
    // the modulesd SIGTERM handler; on Windows shutdown flows through here.
    wm_shutdown_requested = 1;

    wmodule * cur_module;
    for (cur_module = wmodules; cur_module; cur_module = cur_module->next) {
        if (cur_module->context->stop) {
            cur_module->context->stop(cur_module->data);
        }
    }
}
```

src/wazuh_modules/src/wmodules.c
```diff
@@ -349,7 +349,14 @@ int modulesSync(char* args, size_t length) {
...
         if (retry > 0) {
-            usleep(retry * WM_MAX_WAIT);
+            // The backoff grows past one second, which usleep() does not accept.
+            const long wait_us = (long)retry * WM_MAX_WAIT_US;
+#ifdef WIN32
+            Sleep(wait_us / 1000);
+#else
+            struct timeval timeout = { wait_us / 1000000, wait_us % 1000000 };
+            select(0, NULL, NULL, NULL, &timeout);
+#endif
```
nbertoldo: One concern about the backoff change bundled in here (`usleep(retry * WM_MAX_WAIT)` -> `WM_MAX_WAIT_US` with `Sleep`/`select`). The µs-vs-ms fix itself is legitimate, but I don't think it belongs in this PR, and I'd suggest reverting it here and tracking it in its own issue, because it changes manager<->agent communication behavior in a way that deserves its own analysis.

Rationale: on Windows, `modulesSync()` does not run in a dedicated thread. A `*_sync` message from the manager is handled inline in the agent's receiver loop:

`receiver.c:145` (manager-message receive loop) -> `wmcom_send()` -> *(WIN32)* `wmcom_dispatch()` -> `wmcom_sync()` -> `modulesSync()`, then `continue` to read the next message.

With the old backoff the total wait was negligible (~3 ms: `usleep` takes microseconds, so `retry * 500` was 0.5/1/1.5 ms). With the new one it becomes real: 0.5 s + 1 s + 1.5 s = **~3 s** on the "sync fails while the agent is running" path. Since that runs inline in the receiver thread on Windows, a single failing sync now blocks the reception of *all* manager->agent traffic (ACKs, keepalives, other `*_sync` messages, config pushes) for ~3 s. That can make the agent look unresponsive to the manager and could aggravate the socket back-pressure we already see elsewhere.

The shutdown path is fine (the loop breaks before sleeping once `wm_shutdown_requested` is set), so this doesn't affect shutdown latency — but it does affect steady-state comms while the agent is running, which is outside the scope of #37554 (a log-level fix).

Proposal:
- Keep this PR focused on the log-level / shutdown-gate fix and drop the backoff change from it.
- Open a separate issue to evaluate the retry/backoff of `modulesSync()` properly: whether a deterministic `sync()` failure should be retried at all, an upper bound on the total wait, and the impact on the receiver thread on Windows before changing the timing.

nbertoldo (changes requested):

src/wazuh_modules/src/wmodules.c
anromerom: Backoff change reverted and issue created and visible here:

- https://github.com/wazuh/wazuh/issues/37757

nbertoldo (dismissed):
LGTM.
- [x] Code changes reviewed

rjcausarano:
The currently failing syscollector CI check is already being investigated in this other issue: https://github.com/wazuh/wazuh/issues/37515

cborla:
Syscollector are flacky tests. issue: [#37515](https://github.com/wazuh/wazuh/issues/37515)

cborla (approved):
LGTM

- [x] Code changes reviewed

Issue #31343, opened 2025-08-13
Critical error when upgrading Wazuh agent from 4.10.2 to 4.10.3 on Windows Server 2022


DiegoRatto:
**Environment**  
| Wazuh version | Component   | Install type | Install method | Platform / OS version |
|---------------|-------------|--------------|----------------|-----------------------|
| 4.10.2        | Wazuh Agent | Agent        | Packages       | Windows Server 2022   |

**Upgrade path:** 4.10.2 → 4.10.3

> Manager is running on:  
> `Linux 6.14.0-27-generic #27~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue Jul 22 17:38:49 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux`

---

**Issue details**  
During the upgrade test in [#31326 (comment)](https://github.com/wazuh/wazuh/issues/31326#issuecomment-3183575999), the following error was detected when upgrading the **Wazuh Agent** from **4.10.2** to **4.10.3**:

```powershell
PS C:\Users\vagrant> .\wazuh-agent-4.10.2-1.msi /q WAZUH_MANAGER="x.x.x.x"

PS C:\Users\vagrant> NET START Wazuh
The Wazuh service is starting.
The Wazuh service could not be started.

The service did not report an error.

More help is available by typing NET HELPMSG 3534.

PS C:\Users\vagrant> NET START Wazuh
The Wazuh service was started successfully.
```

vikman90:
@DiegoRatto This is actually usually due to a problem in the environment, not specifically with the agent.

If I understand correctly, the problem appears on the first boot, so it's not related to the upgrade.

Were you able to try the process check? Did you try it in a clean environment?

DiegoRatto:
@vikman90 The test is run again with a clean environment to see if the error can be replicated.

DiegoRatto:
# Update

The test was relaunched with a clean environment and the error could not be reproduced. It was determined to be an environmental issue and not an agent issue per se.

<img width="1261" height="564" alt="Image" src="https://github.com/user-attachments/assets/2941bb9e-3c14-4967-9049-663a65b87e73" />

<img width="1571" height="925" alt="Image" src="https://github.com/user-attachments/assets/08b6307b-ba87-48cc-99a2-e7d3a7e3838c" />

The issue is closed as not planned

Issue #26067, opened 2024-09-30
Critical error while installing Wazuh Agent in Windows Server 2022


hossam1522:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.9.0 | Wazuh Agent | Agent | Packages | OS version |

While doing the upgrade test in https://github.com/wazuh/wazuh/issues/26002#issuecomment-2383483688, it has been detected the following error while installing Wazuh Agent 4.9.0:

```
PS C:\Users\vagrant\Downloads> .\wazuh-agent-4.9.0-1.msi /q WAZUH_MANAGER="x.x.x.x"
PS C:\Users\vagrant\Downloads> NET START Wazuh
The Wazuh service is starting.
The Wazuh service could not be started.

The service did not report an error.

More help is available by typing NET HELPMSG 3534.

PS C:\Users\vagrant\Downloads> NET START Wazuh

The Wazuh service was started successfully.
```

It is worth mentioning that the manager is an Ubuntu 24 ARM.

When analyzing the logs, the following information is obtained:

```
PS C:\Program Files (x86)\ossec-agent> Get-Content ossec.log | findstr ERROR
2024/09/30 14:54:29 wazuh-agent: ERROR: CryptAcquireContext Flag: NewKeySet (1): (8009000f)

PS C:\Program Files (x86)\ossec-agent> Get-Content ossec.log | findstr WARNING

PS C:\Program Files (x86)\ossec-agent> Get-Content ossec.log | findstr CRITICAL
2024/09/30 14:54:29 wazuh-agent: CRITICAL: randombytes failed for all possible methods for accessing random data
```

- ossec.log:  [ossec.log](https://github.com/user-attachments/files/17192261/ossec.log)

- ossec.conf: [ossec.txt](https://github.com/user-attachments/files/17192319/ossec.txt)

- Manager ossec.log: [ossec-manager.log](https://github.com/user-attachments/files/17192484/ossec-manager.log)

I also consider it important to mention that, despite the error, the upgrade to 4.9.1-RC 2 was successful and everything seems to be working perfectly. More research is required.

vikman90:
This issue is a duplicate of:
- #22630

According to the code and the explanation by @pereyra-m, it appears that the Wazuh Agent encountered a problem accessing its cryptographic container. The error codes suggest that the key container either could not be accessed due to permission issues or a new container could not be created because one already exists but access was denied.

[...]

PR #37701, opened 2026-07-15
Prevent race condition in Windows RSA key container initialization (randombytes)

### LINKED ISSUE #37308: randombytes failed for all possible methods for accessing random data
A critical error was found during the nightly upgrade tests before upgrading the Windows agent from version 4.14.5 to 5.0.0 on Windows 10 (amd64).

```
2026/06/30 04:24:51 wazuh-agent: ERROR: CryptAcquireContext Flag: NewKeySet (1): (8009000f)
2026/06/30 04:24:51 wazuh-agent: CRITICAL: randombytes failed for all possible methods for accessing random data
```

Those messages have been reported previously in the following issues

- 4.9.0: https://github.com/wazuh/wazuh/issues/26067
- 4.10.2: https://github.com/wazuh/wazuh/issues/31343

We need to reevaluate this issue to determine whether anything needs to be fixed.

MarcelKemp:
## Description

On Windows, `wazuh-agent.exe` runs as a single process with multiple worker threads (syscheck, logcollector, execd, and one thread per configured wodle module), several of which call `randombytes()` during startup. The lazy initialization of the CryptoAPI key container handle (`static HCRYPTPROV prov`) in `src/shared/randombytes.c` was not protected by any synchronization primitive.

When two threads reached the initialization block concurrently — which is likely right after a fresh install/upgrade, before any default RSA key container exists for the running account — a classic CryptoAPI race could occur:

1. Thread A gets `NTE_BAD_KEYSET` from the first `CryptAcquireContext` call and successfully creates the default container via `CRYPT_NEWKEYSET`.
2. Thread B, having entered the same unguarded block a moment earlier, also gets `NTE_BAD_KEYSET`. By the time it attempts `CRYPT_NEWKEYSET`, the container has just been created by thread A, so it fails with `NTE_EXISTS` (`0x8009000f`).
3. Thread B logs `CryptAcquireContext Flag: NewKeySet (1): (8009000f)` and calls `merror_exit()`, which internally calls `exit(1)` — terminating the **entire agent process**, not just the failing thread.

This matches exactly the error reported in #37308 and its historical duplicates (#26067, #31343), which had previously been closed as "environment-related" because the race is timing-dependent and hard to reproduce on demand. It was reliably reproduced again during 5.0.0 nightly upgrade testing, which motivated this fix.

Closes #37308

## Proposed Changes

- Added a `pthread_mutex_t` (`prov_mutex`) guarding the lazy initialization and use of the Windows CryptoAPI provider handle in `randombytes()`, serializing concurrent calls from different agent threads so only one thread creates/repairs the key container while the others reuse the already-initialized handle.
- No change to the Unix/Linux code path (the `#else` branch), which was not affected by this issue.
- No change to the function's public API or return semantics.

### Results and Evidence

- Static analysis of the Windows agent startup path (`src/win32/win_utils.c::local_start()`) confirms multiple threads (syscheck, logcollector, execd, and each configured wodle module) are started concurrently, and several call `randombytes()` / `os_random()` early in their lifecycle — matching the conditions required to trigger the race.
- The reported error code `0x8009000f` (`NTE_EXISTS`) returned from the `CRYPT_NEWKEYSET` call is only reachable, per the Windows CryptoAPI semantics, when the key container was created by a concurrent caller between the failed plain `CryptAcquireContext` call and the `CRYPT_NEWKEYSET` retry — consistent with an unsynchronized race rather than an environment issue.
- The mingw-w64 Windows cross-compiler was not available in this environment, so the patched logic was validated by compiling a syntax-equivalent reproduction of the patched function (Windows API calls stubbed out) with `gcc -Wall -Wextra -Werror -pthread`; it compiled without warnings and executed as expected.
- No live reproduction of the original crash was performed against a QA Windows environment as part of this change. A `TARGET=winagent` CI build and a re-run of the nightly upgrade suite are recommended to fully confirm the fix.

### Artifacts Affected

- `wazuh-agent` (Windows) and any other Windows executable statically linking `shared/randombytes.c` (e.g. `agent-auth`, `manage_agents`) — behavior only changes under `WIN32` builds.
- No impact on Linux/Unix agent or manager binaries (code path untouched).

### Configuration Changes

- None.

### Documentation Updates

- None required — this is an internal implementation detail with no user-facing behavior or configuration change.

### Tests Introduced

- No new automated test added. There is no existing unit test target that links the real `src/shared/randombytes.c` (existing tests mock it via `randombytes_wrappers`), and deterministically unit-testing a Windows CryptoAPI thread race is impractical in the current test harness.
- Suggested manual/QA validation: re-run the 5.0.0 Windows upgrade test suite (internal-devel-requests#5354) across multiple iterations without environment destroy, to confirm the `randombytes failed for all possible methods for accessing random data` message no longer appears.

## Review Checklist

- [x] Code changes reviewed
- [ ] Validated on a real Windows (`TARGET=winagent`) build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ All notable changes to this project will be documented in this file.
 
 - Fixed AES connection getting reset to blowfish on keystore rebuild. ([#37524](https://github.com/wazuh/wazuh/pull/37524))
 
+### Agent
+
+#### Fixed
+
+- Prevent a race condition in `randombytes` during the initialization of the Windows RSA key container ([#37701](https://github.com/wazuh/wazuh/pull/37701))
+
 ### Ruleset
 
 #### Fixed
--- a/src/shared/randombytes.c
+++ b/src/shared/randombytes.c
@@ -25,6 +25,15 @@ void randombytes(void *ptr, size_t length)
 
 #ifdef WIN32
     static HCRYPTPROV prov = 0;
+    static pthread_mutex_t prov_mutex = PTHREAD_MUTEX_INITIALIZER;
+
+    /* Multiple agent threads (syscheck, logcollector, execd, wodles...) may reach
+     * this function concurrently on the first call. Without serializing access,
+     * two threads can race inside CryptAcquireContext: one creates the default
+     * key container while the other, having just observed NTE_BAD_KEYSET, tries
+     * to create it too and gets NTE_EXISTS, aborting the whole process.
+     */
+    w_mutex_lock(&prov_mutex);
 
     if (prov == 0) {
         if (!CryptAcquireContext(&prov, NULL, NULL, PROV_RSA_FULL, 0)) {
@@ -55,6 +64,8 @@ void randombytes(void *ptr, size_t length)
     if (!failed && !CryptGenRandom(prov, length, ptr)) {
         failed = 1;
     }
+
+    w_mutex_unlock(&prov_mutex);
 #else
     static int fh = -1;
     ssize_t ret;


Darioortegaleyva (dismissed):

Darioortegaleyva (approved):
Root cause confirmed against the reproduced environment and race reproduced in isolation (0/2000 with the patch vs 47-54/2000 without):
[wazuh/wazuh#37308 (comment)](https://github.com/wazuh/wazuh/issues/37308#issuecomment-4991992509)

LGTM !

PR #37725, opened 2026-07-16
Limit node expansion in yaml2json

vikman90:
## Description
`yaml2json_node()` (src/shared/yaml2json.c) converts a parsed YAML document into a `cJSON` tree recursively, with no cap on alias expansion, recursion depth, or resulting node count. Because libyaml resolves an alias to the same underlying node as its anchor rather than cloning it, a document with deeply nested anchors/aliases can produce a very large number of `cJSON` allocations from a small input. This is reachable unconditionally from the SCA module (`wm_sca_read_files()`) on every scan cycle, using policy files delivered by the manager, and can exhaust agent memory (GHSA-x8v7-g49q-ff7j). The fix caps the total number of nodes a single document may expand into and aborts the conversion, freeing partial results, once the cap is exceeded.

## Proposed Changes
- `src/shared/yaml2json.c`: added a `YAML2JSON_MAX_NODES` (100000) cap. `yaml2json_node()` now takes a `node_count` accumulator, incremented on every node visited; once the cap is exceeded, the function logs a warning and returns `NULL` instead of allocating further `cJSON` nodes. The sequence/mapping recursion cases now check each child's result and, on `NULL`, delete the partially built object and propagate `NULL` upward so `yaml2json()` returns `NULL` for the whole document (existing callers already treat a `NULL` return as "skip this policy").
- `src/unit_tests/shared/CMakeLists.txt`: registered the new `test_yaml2json_op` test (no link-wrap flags needed; it exercises real file I/O and the real YAML/cJSON pipeline).
- `src/unit_tests/shared/test_yaml2json_op.c` (new): cmocka tests for the fix.

### Results and Evidence
- Standalone reproduction (real `yaml2json.c` logic + this repo's built `libyaml.a` + `cJSON.c`) against a 394-byte document with nested aliases:
  - Before the fix: 234,583 `cJSON` nodes allocated.
  - After the fix: `yaml2json()` returns `NULL` once the 100000-node cap is exceeded; peak RSS for the same input drops to ~11 MB.
- The largest bundled SCA policy in this repo (`ruleset/sca/windows/cis_win11_enterprise.yml`, ~10,000 lines / ~18,300 nodes) still converts successfully, confirming the cap does not affect legitimate policies (which top out around 18-26K nodes for the largest bundled files).
- New unit tests `test_yaml2json_valid_policy` (must still succeed) and `test_yaml2json_alias_expansion_rejected` (must return `NULL`) added; the latter fails before the fix and passes after.

### Artifacts Affected
- `wazuh-agentd` / `wazuh-modulesd` (agent): consumes `yaml2json()` via the SCA module (`wm_sca.c`), the only caller of this function in this codebase.

### Configuration Changes
None.

### Documentation Updates
None.

### Tests Introduced
- `src/unit_tests/shared/test_yaml2json_op.c`:
  - `test_yaml2json_valid_policy` — a small, valid SCA-like policy still converts to the expected `cJSON` structure.
  - `test_yaml2json_alias_expansion_rejected` — a document with nested aliases is rejected (`yaml2json()` returns `NULL`).

## Credits
Reported by @karasu-hakira (Hakira, https://github.com/hakiraio).

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ All notable changes to this project will be documented in this file.
 
 ### Agent
 
+#### Changed
+
+- Improved SCA policy YAML parsing to bound memory usage on deeply nested policies. ([#37725](https://github.com/wazuh/wazuh/pull/37725))
+
 #### Fixed
 
 - Prevent a race condition in `randombytes` during the initialization of the Windows RSA key container ([#37701](https://github.com/wazuh/wazuh/pull/37701))
--- a/src/shared/yaml2json.c
+++ b/src/shared/yaml2json.c
@@ -14,7 +14,12 @@
 #include <errno.h>
 #include "shared.h"
 
-static cJSON * yaml2json_node(yaml_document_t * document, yaml_node_t * node, int quoted_float_as_string);
+/* Maximum number of JSON nodes a single YAML document may expand into.
+ * The largest bundled SCA policy expands to about 18000 nodes, so this leaves
+ * ample headroom for legitimate documents. */
+#define YAML2JSON_MAX_NODES 100000
+
+static cJSON * yaml2json_node(yaml_document_t * document, yaml_node_t * node, int quoted_float_as_string, unsigned int * node_count);
 
 int yaml_parse_stdin(yaml_document_t * document) {
     yaml_parser_t parser;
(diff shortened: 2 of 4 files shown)


MarcelKemp (approved):
LGTM.

PR #37745, opened 2026-07-17
Reject upgrade commands from the agent channel in wazuh-analysisd

vikman90:
## Description
`wazuh-analysisd` forwarded any agent-originated message received on the `UPGRADE_MQ` queue verbatim to the `wm_agent_upgrade` manager module's socket, with no check on the `command` field. Since that module applies no origin check either, an agent could inject an `upgrade_custom` command with an attacker-chosen `file_path`, causing the manager (running as root) to read and stream back any file on its filesystem — or inject a plain `upgrade` command to force upgrade/downgrade/reinstall of an agent. The fix makes `wazuh-analysisd` only forward the one command an agent may legitimately send on this channel (`upgrade_update_status`); any other command is rejected and logged before it ever reaches the upgrade module's socket.

## Proposed Changes
- `wazuh-analysisd` used to forward any agent-originated message on the upgrade channel straight to the upgrade module's socket, with no check on what command it was actually requesting. This PR adds that missing check: analysisd now looks at the command name before forwarding, and only lets through `upgrade_update_status` — the one message an agent legitimately needs to send here, to report progress on an upgrade already in flight. Any other command (in particular `upgrade` and `upgrade_custom`, which only the management API should ever be able to trigger) is dropped and logged at debug level, so it can't be abused to flood the manager's log at default verbosity either.
- To make this check testable on its own, the per-message forwarding logic that used to live inline inside the dispatch thread was pulled out into its own function, `DispatchUpgradeModule`, following the same pattern already used for `DispatchDBSync` elsewhere in the file. This part is a pure structural change, with no behavior difference — it just gives the new check something to be unit-tested against.

### Results and Evidence
Unit tests run automatically in CI (see Tests Introduced below); the evidence here is from manual verification against a running manager.

Built the fix (`wazuh-analysisd` v4.14.8) and installed it on a manager with a real registered agent (id `001`). Injected the same agent-channel message described in the GHSA reproduction directly onto `wazuh-analysisd`'s local queue socket, impersonating that agent:
```
u:[001] (fbb1635f5a97) any->upgrade_module:{"command":"upgrade_custom","parameters":{"file_path":"/var/ossec/tmp/agent.wpk","installer":"upgrade.sh"}}
```
`ossec.log` confirms the message is rejected inside `DispatchUpgradeModule`, before it ever reaches `WM_UPGRADE_SOCK` — so the vulnerable file read is never attempted:
```
2026/07/17 12:48:49 wazuh-analysisd[103309] analysisd.c:1973 at DispatchUpgradeModule(): DEBUG: Agent '001' is not allowed to request the upgrade command received on the agent channel
```

### Artifacts Affected
- `wazuh-analysisd` (manager component).

### Configuration Changes
None.

### Documentation Updates
None.

### Tests Introduced
- `src/unit_tests/analysisd/test_analysisd_upgrade_module.c`:
  - `test_DispatchUpgradeModule_blocks_upgrade_custom_from_agent` — an agent-forged `upgrade_custom` message (with `file_path` set to `/etc/shadow`) is rejected and never reaches `OS_ConnectUnixDomain`/`OS_SendSecureTCP`.
  - `test_DispatchUpgradeModule_blocks_upgrade_from_agent` — an agent-forged plain `upgrade` message is rejected the same way.
  - `test_DispatchUpgradeModule_allows_status_update` — the legitimate `upgrade_update_status` message an agent sends is still forwarded to `WM_UPGRADE_SOCK` unchanged, confirming the fix does not break normal upgrade-status reporting.

## Credits

Thanks to @kah-ja, [turingpoint](https://www.turingpoint.de) for reporting this issue.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 
 - Fixed AES connection getting reset to blowfish on keystore rebuild. ([#37524](https://github.com/wazuh/wazuh/pull/37524))
 - Fixed a deadlock in `wazuh-analysisd` that stopped alert generation when the Active Response queue filled up. ([#37764](https://github.com/wazuh/wazuh/pull/37764))
+- Restricted the upgrade commands accepted from the agent message channel in Analysisd. ([#37745](https://github.com/wazuh/wazuh/pull/37745))
 
 ### Agent
 
--- a/src/analysisd/analysisd.c
+++ b/src/analysisd/analysisd.c
@@ -1962,9 +1962,16 @@ void DispatchUpgradeModule(Eventinfo * lf) {
     cJSON *message_obj = cJSON_Parse(lf->log);
 
     if (message_obj) {
+        cJSON *message_command = cJSON_GetObjectItem(message_obj, "command");
         cJSON *message_params = cJSON_GetObjectItem(message_obj, "parameters");
 
-        if (message_params) {
+        // Agents may only report their own upgrade status; any other command
+        // (e.g. "upgrade", "upgrade_custom") must come from the management API,
+        // never from the agent-message channel.
+        if (!message_command || message_command->type != cJSON_String ||
+            strcmp(message_command->valuestring, "upgrade_update_status") != 0) {
+            mdebug1("Agent '%s' is not allowed to request the upgrade command received on the agent channel", lf->agent_id);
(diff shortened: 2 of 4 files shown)


jotacarma90 (dismissed):
LGTM

TomasTurina (dismissed):

PR #37756, opened 2026-07-17
wazuh-engine: Indexer connector exponential backoff

### LINKED ISSUE #37516: Implement adaptive retry and backpressure handling for Indexer saturation in wazuh-analysisd
### Summary

Implement an adaptive retry mechanism in `wazuh-analysisd` to properly handle temporary Indexer saturation scenarios.

Currently, when forwarding events to the Wazuh Indexer, requests may fail for different reasons. One specific case requires dedicated handling: when the Indexer explicitly indicates that it is temporarily overloaded and cannot process additional requests.

Instead of continuously retrying at full speed, `wazuh-analysisd` should recognize this condition and apply a backpressure strategy that allows the Indexer to recover while minimizing data loss and preventing additional overload.

The implementation should begin with a design study to define the most appropriate strategy and coordinate with the Indexer team to agree on how this saturation condition is identified (HTTP status code, error response, error type, etc.).

### Proposed work

The solution should evaluate and implement mechanisms such as:

* Detect Indexer saturation responses.
* Apply retry logic only for saturation-related errors.
* Implement exponential backoff with jitter (or an equivalent adaptive strategy).
* Temporarily accumulate events in the output queue while retries are performed.
* Prevent continuous request flooding during sustained overload periods.
* Define queue management policies for prolonged saturation scenarios.
* Gracefully discard events when the output queue reaches its configured capacity, rather than indefinitely attempting to send requests that further overload the Indexer.
* Emit clear logs and metrics whenever:

  * Saturation is detected.
  * Retry mode is activated.
  * Events are dropped due to queue exhaustion caused by sustained Indexer saturation.

The implementation should balance data preservation with system stability, ensuring that recovery of the Indexer is prioritized over unlimited buffering.

---

## Tasks

* [ ] Coordinate with the Indexer team to define how Indexer saturation responses are identified.
* [ ] Analyze current `wazuh-analysisd` event forwarding and retry behavior.
* [ ] Design an adaptive retry and backpressure strategy.
* [ ] Define retry policy (backoff, jitter, retry limits, cooldown, etc.).
* [ ] Define output queue behavior during sustained Indexer saturation.
* [ ] Define event discard policy when queue capacity is exceeded.
* [ ] Implement saturation detection.
* [ ] Implement adaptive retry mechanism.
* [ ] Implement queue management during retry periods.
* [ ] Implement event discard logic for prolonged saturation.
* [ ] Add informative logging for saturation detection, retries, queue state, and discarded events.
* [ ] Add metrics (if applicable) to monitor saturation events, retries, queue utilization, and dropped events.
* [ ] Add unit and integration tests covering normal operation, transient saturation, and prolonged saturation scenarios.
* [ ] Update the corresponding documentation.

---

## Definition of Done (DoD)

* [ ] `wazuh-analysisd` correctly identifies Indexer saturation responses using the agreed criteria.
* [ ] Retry logic is applied only for temporary saturation errors.
* [ ] Retries use an adaptive backoff strategy with jitter (or the approved alternative).
* [ ] Continuous request flooding is avoided while the Indexer is saturated.
* [ ] Events are buffered according to the defined queue policy.
* [ ] Events are discarded only when the queue reaches its configured capacity under sustained saturation.
* [ ] Event discards are clearly logged with sufficient context for troubleshooting.
* [ ] Appropriate metrics and observability are available for saturation, retries, queue occupancy, and dropped events.
* [ ] The implementation has automated tests covering recovery from transient saturation and behavior during prolonged overload.
* [ ] Documentation describes the retry strategy, queue behavior, discard policy, and operational considerations.

matigarciadev:
## Description

The indexer connectors (sync and async) previously retried transient failures — `429 Too Many Requests`, `409` version conflicts, connection errors, and (async only) per-item `cluster_block_exception`s — with a naive scheme: the async `IndexerBulkQueue` slept a fixed `retryDelaySeconds` after every failed dispatch, and the sync connector's four independent retry loops each hand-rolled their own `sleep_for` + `retryDelaySeconds *= 2` doubling capped at a hardcoded 30s, duplicated four times with no jitter. Under sustained pressure (e.g. the `wazuh-indexer` cluster blocking writes), this meant either hammering the cluster at a constant rate or having several independent retry loops escalate in lockstep with no randomization.

This PR replaces both with a shared `IndexerExponentialBackoff` (base delay → doubling → jittered step → capped at a new configurable maximum), coordinated across the async dispatch thread and the asynchronous response-classification thread via sequence-tagged `requestBackoff()` / `reportHealthyResponse()` calls, so a `cluster_block_exception` observed on an already-successful HTTP response can still pause the *next* send without retrying data that don't need retrying.

Closes #37516

## Proposed Changes

- **New shared backoff primitive** — `IndexerExponentialBackoff` (`src/shared_modules/indexer_connector/src/exponentialBackoff.hpp`): 1st failure sleeps exactly the base delay (deterministic), each subsequent consecutive failure doubles the delay and sleeps a jittered value between the previous and new capped step, capped at a configurable maximum, resetting to the base on `reset()`.
- **Async connector (`indexerConnectorAsyncImpl.hpp` / `indexerBulkQueue.hpp`)**
  - `IndexerBulkQueue::Processor` now carries a per-send-attempt `sequence` id.
  - New `requestBackoff(sequence)` / `reportHealthyResponse(sequence)` let the response-logger thread signal the dispatch thread to pause the *next* send without touching the batch that already got a response — out-of-order/stale classifications are ignored via the sequence tag.
  - `cluster_block_exception` per-item errors are now recognized explicitly: tallied, logged once per batch (not per document), discarded (not retried), and trigger `requestBackoff()`.
  - Connection errors (`HTTP_CONNECTION_ERROR = -1`) are now retried with backoff, same as `429`/`409`.
  - Processor-exception retries (transport failures) now also use the shared exponential backoff instead of the old fixed `retryDelaySeconds`.
- **Sync connector (`indexerConnectorSyncImpl.hpp`)** — all four retry loops (`processBulk`, `processBulkChunk`, `executeUpdateByQuery`, `search`/PIT paths) now share the same `IndexerExponentialBackoff` semantics instead of each hand-rolling `sleep_for` + manual doubling capped at a hardcoded 30s. A dedicated `m_retryMutex`/`m_retryCv` was introduced for the backoff sleep, since the bulk-thread loop already holds `m_mutex` while running — reusing it would self-deadlock; destructor ordering was adjusted accordingly to avoid hanging on shutdown mid-retry.
- **New configuration option** — `analysisd.indexer_max_retry_delay` (default `15`, range `1`–`3600`), validated in `main.cpp` and threaded through to both connectors as `max_retry_delay_seconds`.
- **Documentation** — new "Retry and backoff behavior" section in `docs/ref/modules/indexer_connector/README.md` with the delay-scaling table, plus references from the engine config docs.
- **Tests** — new unit suites for the backoff primitive and the queue's backoff coordination, plus new/updated async and sync connector tests (cluster-block handling, connection-error retry, no-duplicate/no-drop guarantees, destructor shutdown timing).

### Results and Evidence

**Manual reproduction — sustained `cluster_block_exception` under load**

`analysisd` was run against a `wazuh-indexer` cluster that started rejecting writes (`cluster_block_exception`) under a benchmark load. Filtering `wazuh-manager.log` for the relevant components:

```
grep -E 'indexerConnectorAsyncImpl|indexerBulkQueue' /var/wazuh-manager/logs/wazuh-manager.log
```

1. Confirms the new option is wired end-to-end at startup, using its documented default:

   ```
   2026/07/16 23:05:19 wazuh-manager-analysisd[780461] conf.hpp:364 at get(): DEBUG: [Config] Using configuration key 'analysisd.indexer_max_retry_delay' from default: '15'
   ```

2. The cluster starts blocking writes; each occurrence is tallied (per-batch count / running total) and triggers a backoff request instead of being retried:

   ```
   2026/07/16 23:07:21 ...WARNING: Index write(s) blocked by the cluster (cluster_block_exception) - 2969 document(s) in this batch (2969 total) not indexed due to this expected condition. Backing off future sends.
   2026/07/16 23:12:42 ...WARNING: ... - 4510 document(s) in this batch (7479 total) ...
   2026/07/16 23:13:02 ...WARNING: ... - 3873 document(s) in this batch (11352 total) ...
   2026/07/16 23:13:44 ...WARNING: ... - 3772 document(s) in this batch (15124 total) ...
   2026/07/16 23:14:04 ...WARNING: ... - 3982 document(s) in this batch (19106 total) ...
   2026/07/16 23:14:25 ...WARNING: ... - 4316 document(s) in this batch (23422 total) ...
   2026/07/16 23:14:47 ...WARNING: ... - 4490 document(s) in this batch (27912 total) ...
   2026/07/16 23:15:10 ...WARNING: ... - 4288 document(s) in this batch (32200 total) ...
   2026/07/16 23:15:34 ...WARNING: ... - 4921 document(s) in this batch (37121 total) ...
   ```

3. Once the block persists across consecutive send attempts (no healthy response in between to reset the counter), the dispatch thread's pause escalates exactly per the documented base=1s/max=15s schedule — every observed value falls inside the theoretical bracket for its consecutive-failure count:

   | Timestamp | Logged pause | Expected bracket | Consecutive failure # |
   |---|---|---|---|
   | 23:13:02 | `pausing 1000 ms` | exactly 1000 ms (no jitter) | 1st |
   | 23:14:25 | `pausing 1700 ms` | [1000, 2000] ms | 2nd |
   | 23:14:47 | `pausing 2316 ms` | [2000, 4000] ms | 3rd |
   | 23:15:09 | `pausing 4360 ms` | [4000, 8000] ms | 4th |
   | 23:15:34 | `pausing 12005 ms` | [8000, 15000] ms (capped) | 5th |

   ```
   2026/07/16 23:13:02 logger-helper[780461] indexerBulkQueue.hpp:214 at dispatch(): DEBUG: IndexerBulkQueue: pausing 1000 ms before next send (external backoff requested).
   2026/07/16 23:14:25 logger-helper[780461] indexerBulkQueue.hpp:214 at dispatch(): DEBUG: IndexerBulkQueue: pausing 1700 ms before next send (external backoff requested).
   2026/07/16 23:14:47 logger-helper[780461] indexerBulkQueue.hpp:214 at dispatch(): DEBUG: IndexerBulkQueue: pausing 2316 ms before next send (external backoff requested).
   2026/07/16 23:15:09 logger-helper[780461] indexerBulkQueue.hpp:214 at dispatch(): DEBUG: IndexerBulkQueue: pausing 4360 ms before next send (external backoff requested).
   2026/07/16 23:15:34 logger-helper[780461] indexerBulkQueue.hpp:214 at dispatch(): DEBUG: IndexerBulkQueue: pausing 12005 ms before next send (external backoff requested).
   ```

   Note the two isolated warnings show no escalated pause: each was followed by a healthy (non-blocked) response before the next block occurred, so `reportHealthyResponse()` reset the failure counter back to the deterministic 1st-failure delay (~1s), which is easy to miss next to the ~20s flush-interval cadence — the escalating sequence only appears once the block becomes *persistent* across consecutive attempts.

**Unit tests**

The `indexer_connector_utest` binary was also built with TSAN:

```
$ cmake -S src -B src/build-indexer-tsan-nocov \
    -DTARGET=manager -DUNIT_TEST=ON -DWAZUH_ENABLE_COVERAGE=OFF \
    -DCMAKE_BUILD_TYPE=Debug \
    -DCMAKE_CXX_FLAGS='-fsanitize=thread -fno-omit-frame-pointer' \
    -DCMAKE_C_FLAGS='-fsanitize=thread -fno-omit-frame-pointer' \
    -DCMAKE_EXE_LINKER_FLAGS='-fsanitize=thread' \
    -DCMAKE_SHARED_LINKER_FLAGS='-fsanitize=thread'

$ cmake --build src/build-indexer-tsan-nocov --target indexer_connector_utest indexer_connector_tool -j 6
[100%] Built target indexer_connector_utest
[100%] Built target indexer_connector_tool
```

`ldd` confirms the unit-test binary is linked against ThreadSanitizer:

```
$ ldd src/build-indexer-tsan-nocov/bin/indexer_connector_utest | sed -n '1,12p'
        linux-vdso.so.1 (0x00007dcbf2ba1000)
        libtsan.so.2 => /lib/x86_64-linux-gnu/libtsan.so.2 (0x00007dcbf1569000)
        libwazuhext.so => /workspaces/wazuh-5.x/wazuh/src/build/lib/libwazuhext.so (0x00007dcbf0a00000)
        librocksdb.so.8 => /workspaces/wazuh-5.x/wazuh/src/external/rocksdb/build/librocksdb.so.8 (0x00007dcbefe00000)
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007dcbefb82000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007dcbf0917000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007dcbf2b63000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007dcbef970000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007dcbf2b5e000)
        /lib64/ld-linux-x86-64.so.2 (0x00007dcbf2ba3000)
```

Backoff-specific subset under TSAN (new primitive, queue coordination, cluster-block handling):

```
$ TSAN_OPTIONS=halt_on_error=0:second_deadlock_stack=1:history_size=7:log_path=/tmp/indexer_connector_tsan_backoff_evidence \
  src/build-indexer-tsan-nocov/bin/indexer_connector_utest --gtest_filter="*Backoff*:*ClusterBlock*:*Retry*"
[==========] Running 17 tests from 4 test suites.
[----------] 7 tests from IndexerExponentialBackoffTest
[ RUN      ] IndexerExponentialBackoffTest.AppliesExponentialCapsAndReset
[       OK ] IndexerExponentialBackoffTest.AppliesExponentialCapsAndReset (0 ms)
[ RUN      ] IndexerExponentialBackoffTest.FirstFailureSleepsExactlyBaseDelay
[       OK ] IndexerExponentialBackoffTest.FirstFailureSleepsExactlyBaseDelay (0 ms)
[ RUN      ] IndexerExponentialBackoffTest.ConsecutiveFailuresDoNotExceedMaxDelay
[       OK ] IndexerExponentialBackoffTest.ConsecutiveFailuresDoNotExceedMaxDelay (0 ms)
[ RUN      ] IndexerExponentialBackoffTest.MaxDelayUsesPreviousExponentialStepAsLowerBound
[       OK ] IndexerExponentialBackoffTest.MaxDelayUsesPreviousExponentialStepAsLowerBound (0 ms)
[ RUN      ] IndexerExponentialBackoffTest.ZeroBaseDelayReturnsZero
[       OK ] IndexerExponentialBackoffTest.ZeroBaseDelayReturnsZero (0 ms)
[ RUN      ] IndexerExponentialBackoffTest.NegativeBaseDelayIsClampedToZero
[       OK ] IndexerExponentialBackoffTest.NegativeBaseDelayIsClampedToZero (0 ms)
[ RUN      ] IndexerExponentialBackoffTest.MaxDelayBelowBaseDelayUsesBaseAsCap
[       OK ] IndexerExponentialBackoffTest.MaxDelayBelowBaseDelayUsesBaseAsCap (0 ms)
[----------] 7 tests from IndexerExponentialBackoffTest (0 ms total)

[----------] 3 tests from IndexerBulkQueueTest
[ RUN      ] IndexerBulkQueueTest.RequestBackoffDoesNotDropOrDuplicateData
[       OK ] IndexerBulkQueueTest.RequestBackoffDoesNotDropOrDuplicateData (0 ms)
[ RUN      ] IndexerBulkQueueTest.RequestBackoffCoalescesRepeatedCalls
[       OK ] IndexerBulkQueueTest.RequestBackoffCoalescesRepeatedCalls (0 ms)
[ RUN      ] IndexerBulkQueueTest.DestructorDoesNotHangWhileWaitingOutAnExternalBackoff
[       OK ] IndexerBulkQueueTest.DestructorDoesNotHangWhileWaitingOutAnExternalBackoff (0 ms)
[----------] 3 tests from IndexerBulkQueueTest (1 ms total)

[----------] 2 tests from IndexerConnectorAsyncTest
[ RUN      ] IndexerConnectorAsyncTest.ClusterBlockExceptionItemsAreDiscardedNotRetried
[       OK ] IndexerConnectorAsyncTest.ClusterBlockExceptionItemsAreDiscardedNotRetried (1057 ms)
[ RUN      ] IndexerConnectorAsyncTest.AppliesBackoffToNextBulkAfterClusterBlockException
[       OK ] IndexerConnectorAsyncTest.AppliesBackoffToNextBulkAfterClusterBlockException (4003 ms)
[----------] 2 tests from IndexerConnectorAsyncTest (5061 ms total)

[----------] 5 tests from IndexerConnectorSyncTest
[ RUN      ] IndexerConnectorSyncTest.ProcessBulkChunkError409VersionConflictWithRetry
[       OK ] IndexerConnectorSyncTest.ProcessBulkChunkError409VersionConflictWithRetry (35 ms)
[ RUN      ] IndexerConnectorSyncTest.ProcessBulkChunkError429TooManyRequestsWithRetry
[       OK ] IndexerConnectorSyncTest.ProcessBulkChunkError429TooManyRequestsWithRetry (1001 ms)
[ RUN      ] IndexerConnectorSyncTest.ExecuteUpdateByQueryWithRetry
[       OK ] IndexerConnectorSyncTest.ExecuteUpdateByQueryWithRetry (48 ms)
[ RUN      ] IndexerConnectorSyncTest.ExecuteSearchQueryError429TooManyRequestsWithRetry
[       OK ] IndexerConnectorSyncTest.ExecuteSearchQueryError429TooManyRequestsWithRetry (1044 ms)
[ RUN      ] IndexerConnectorSyncTest.SearchWithPitError429TooManyRequestsWithRetry
[       OK ] IndexerConnectorSyncTest.SearchWithPitError429TooManyRequestsWithRetry (1002 ms)
[----------] 5 tests from IndexerConnectorSyncTest (3132 ms total)

[==========] 17 tests from 4 test suites ran. (8196 ms total)
[  PASSED  ] 17 tests.
```

No TSAN report was emitted for that run:

```
$ find /tmp -maxdepth 1 -type f -name 'indexer_connector_tsan_backoff_evidence*' -print
# no output
```

```
$ TSAN_OPTIONS=halt_on_error=0:second_deadlock_stack=1:history_size=7:log_path=/tmp/indexer_connector_tsan_sync_evidence \
  src/build-indexer-tsan-nocov/bin/indexer_connector_utest --gtest_filter='IndexerConnectorSyncTest.*' --gtest_brief=1
[==========] 134 tests from 1 test suite ran. (9426 ms total)
[  PASSED  ] 134 tests.

$ find /tmp -maxdepth 1 -type f -name 'indexer_connector_tsan_sync_evidence*' -print
# no output
```

Full `indexer_connector` unit suite (regular regression check — nothing else broke):

```
$ ./bin/indexer_connector_utest
[==========] 238 tests from 6 test suites ran. (29614 ms total)
[  PASSED  ] 238 tests.
```

Engine `wiconnector` config-plumbing tests (`max_retry_delay_seconds` propagation into the JSON config consumed by `wazuh-engine`):

```
$ ./wIndexerConnector_utest --gtest_filter="*MaxRetryDelay*:*RetryDelay*"
[----------] 2 tests from ConfigTest
[ RUN      ] ConfigTest.ConfigDefaultMaxRetryDelayToJson       [       OK ]
[ RUN      ] ConfigTest.ConfigMaxRetryDelayMaximumValueToJson  [       OK ]
[  PASSED  ] 2 tests.

$ ./wIndexerConnector_utest
[==========] 85 tests from 3 test suites ran. (247 ms total)
[  PASSED  ] 85 tests.
  YOU HAVE 1 DISABLED TEST
```

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [x] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel
  
- Wazuh server API/Framework
  - [ ] Run API Integration Tests — N/A (no API/Framework surface changed).

### Artifacts Affected

- `wazuh-manager` (`analysisd`) — async and sync indexer-connector retry/backoff behavior at runtime.
- `wazuh-engine` binary — new `max_retry_delay_seconds` field consumed via `wiconnector::Config`.
- `libindexer_connector` shared module (`src/shared_modules/indexer_connector`) — new `exponentialBackoff.hpp`, changed `indexerBulkQueue.hpp` `Processor` signature, changed `indexerConnectorAsyncImpl.hpp`/`indexerConnectorSyncImpl.hpp`.
- `indexer_connector` testtool (`push-events`, etc.) — new optional `max_retry_delay_seconds` config field.
- Developer documentation: `docs/ref/modules/indexer_connector/README.md`, `docs/ref/modules/engine/README.md`, `docs/ref/modules/engine/configuration.md`.

### Configuration Changes

- **New option**: `analysisd.indexer_max_retry_delay` (env `WAZUH_INDEXER_MAX_RETRY_DELAY`)
  - Default: `15` seconds. Allowed range: `1`–`3600`. Validated at startup in `main.cpp`; out-of-range values throw at startup rather than silently clamping.
  - Maps to `max_retry_delay_seconds` in both the async and sync connector JSON configs (`wiconnector::Config::maxRetryDelaySeconds`, and the sync/async `IndexerConnector*Impl` constructors).
  - Backward compatible: connectors default to `15` when the field is absent from the config JSON (e.g. older configs, or the standalone testtool).
- **Behavior change (no new option, but a default change)**: the sync connector's previously hardcoded `MAX_RETRY_BACKOFF_SECONDS = 30` cap is now driven by this same configurable maximum (default `15`), and connection errors (previously not retried in some sync paths) are now retried with backoff like `429`.

### Tests Introduced

- `src/shared_modules/indexer_connector/tests/unit/exponentialBackoff_test.cpp` (new, 7 tests) — base-delay determinism on first failure, exponential growth + jitter bounds, cap behavior, zero/negative base delay, `max < base` edge case, `reset()`.
- `src/shared_modules/indexer_connector/tests/unit/indexerBulkQueue_test.cpp` (new, 6 tests) — `requestBackoff()`/`reportHealthyResponse()` don't drop/duplicate data, repeated backoff requests coalesce into one pause, stale/out-of-order sequence reports are ignored, failed processor calls still retry the same batch, destructor doesn't hang mid-backoff or mid-retry.
- `src/shared_modules/indexer_connector/tests/unit/indexerConnectorAsync_test.cpp` (updated) — `cluster_block_exception` items discarded and not retried; next bulk is measurably delayed by the escalated backoff; `502` (and other non-retryable codes) confirmed still not retried; existing generic/500/429 tests updated to the new `Processor(sequence, batch)` signature.
- `src/shared_modules/indexer_connector/tests/unit/indexerConnectorSync_test.cpp` (updated) — all four retry loops (bulk, bulk-chunk, update-by-query, search/PIT) updated to the shared backoff, template parameters adjusted (`RetryDelay` now must be `> 0`, enforced via `static_assert`).
- `src/engine/source/wiconnector/test/src/unit/wic_test.cpp` (updated, +2 tests) — `max_retry_delay_seconds` present in `Config::toJson()` with its default (`15`) and at the configured maximum (`3600`).

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...


src/shared_modules/indexer_connector/src/indexerConnectorAsyncImpl.hpp
```diff
@@ -467,11 +530,13 @@ class IndexerConnectorAsyncImpl final
                     // of both in the logger queue (the main RSS driver under sustained load).
                     if (!bulkResponseHasErrors(response))
                     {
+                        m_queue->reportHealthyResponse(sequence);
                         return;
                     }
 
                     // Dispatch to error logger (bounded queue: drop + warn when it is full).
-                    IndexerResponse responseData(std::move(bulkData), std::move(boundaries), std::move(response));
+                    IndexerResponse responseData(
```
NahuFigueroa97, line 537: The saturation signal can be lost precisely when it matters most.
In indexerConnectorAsyncImpl.hpp, the cluster_block_exception detection (which triggers requestBackoff()) only occurs within the deferred callback of m_loggerProcessor, which processes an IndexerResponse pushed with m_loggerProcessor->push. If that queue (bounded by logger_queue_size) is full, the response is discarded outright, and neither requestBackoff() nor reportHealthyResponse() is ever called for that sequence. Under sustained load, this queue is more likely to be full, so the saturation detection mechanism itself might fail, and the connector would continue sending at full speed instead of backoffing. There is no test that covers this combination (full logger queue + cluster_block_exception).

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -10,6 +10,7 @@
  */
 
 #include "IURLRequest.hpp"
+#include "exponentialBackoff.hpp"
```
NahuFigueroa97, line 13: In both sync and async, retry loops for 429/connection errors (and in async, also 409 at the bulk level) retry indefinitely without a limit on the number of attempts. The issue explicitly requests defining "retry limits, cooldown". In the sync connector (used by vulnerability_scanner, inventory_sync, content_manager), this is more serious: processBulk()/processBulkChunk() run with m_mutex held by the invoker (the background thread or an external caller via scopeLock()), and although the backoff sleep uses a separate mutex (m_retryMutex) to prevent auto-deadlock, it never releases m_mutex during that sleep—so while the backoff lasts (potentially indefinitely), any other thread that wants to enqueue more data (bulkIndex/bulkDelete/flush) is blocked instead of accumulating in the queue, which is precisely what the issue requests ("accumulate events in the output queue while retrying"). In async this does not happen (the IndexerBulkQueue releases the mutex correctly during sleep via wait_for), so analyzedd is not affected by this.

NahuFigueroa97 (changes requested):

src/shared_modules/indexer_connector/src/indexerConnectorAsyncImpl.hpp
matigarciadev: You’re right, this can happen with the current design. We intentionally keep the dispatch path cheap and avoid doing the full bulk-response parse there; responses that may contain errors are deferred to the logger workers, where we do the detailed parsing and decide whether to call `requestBackoff()`.

The tradeoff is exactly the one you pointed out: if the logger queue is full, that response is dropped and we lose the chance to classify a `cluster_block_exception` for that sequence. The queue size and logger worker count are configurable, and the warning currently points users to increase `logger_queue_size` when drops happen; increasing `logger_threads` can also help drain it faster.

The pressure signal itself still comes from the indexer, so the logger queue being full is not required for `cluster_block_exception` to happen. But I agree that the “full logger queue + cluster_block_exception” case is not covered, and in that specific case the backoff signal can be missed.

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
matigarciadev: Good point. The retry policy itself was not changed by this PR for the existing retryable cases. For `429`, the connector already retried indefinitely; this change only makes the wait happen through the retry CV/backoff path, so shutdown does not self-deadlock and the cooldown is bounded by `max_retry_delay_seconds`. I agree that the sync connector can still block other callers while retrying a persistent `429`, because `processBulk()` / `processBulkChunk()` may run while `m_mutex` is held by the background thread, `flush()`, or an external `scopeLock()` caller. That behavior was already present before this change, but it is a valid concern. Async does not have the same issue because `IndexerBulkQueue` works with a local batch and releases its queue mutex while waiting, so producers can keep enqueueing while the retry backoff is active. I’d prefer to treat retry limits for the sync connector as a separate policy/design decision, since changing it affects the existing “retry until success for retryable errors” contract. We’ll discuss it with the team and decide whether adding a max-attempts/max-duration policy for sync is worth doing.

src/shared_modules/indexer_connector/src/indexerConnectorAsyncImpl.hpp
```diff
@@ -525,17 +589,20 @@ class IndexerConnectorAsyncImpl final
...
+                        LOGFN_DEBUG2(m_logFn, "Too many requests, retrying with exponential backoff.");
                         throw IndexerConnectorException(error);
                     }
                     else
                     {
-                        LOGFN_WARN(m_logFn, "%s, status code: %ld.", error.c_str(), statusCode);
+                        LOGFN_WARN(m_logFn,
+                                   "%s, status code: %ld. Discarding event(s) in this batch.",
+                                   error.c_str(),
+                                   statusCode);
```
LucioDonda, line 605: This block of code should be cover by tests

src/shared_modules/indexer_connector/tests/unit/indexerConnectorAsync_test.cpp
```diff
@@ -901,32 +1158,48 @@ TEST_F(IndexerConnectorAsyncTest, HandleGenericError)
...
 
-                // Test with a generic error status code (502 Bad Gateway)
-                if (std::holds_alternative<TPostRequestParameters<const std::string&>>(postParams))
+                if (currentCall == 1)
                 {
-                    std::get<TPostRequestParameters<const std::string&>>(postParams).onError("Bad Gateway", 502, "");
+                    if (std::holds_alternative<TPostRequestParameters<const std::string&>>(postParams))
+                    {
+                        std::get<TPostRequestParameters<const std::string&>>(postParams)
+                            .onError("Bad Gateway", 502, "");
```
LucioDonda, line 1166: There's no check on error codes like:
`onError("...", -1, "")`

src/shared_modules/indexer_connector/tests/unit/indexerConnectorAsync_test.cpp
```diff
@@ -811,6 +813,257 @@ TEST_F(IndexerConnectorAsyncTest, HandleError429TooManyRequests)
...
+    {
+        EXPECT_EQ(receivedData[1].find("value" + std::to_string(i)), std::string::npos)
+            << "value" << i << " should not have been retried after a cluster_block_exception";
+    }
+    // Only the genuinely new data goes out in the next send.
+    EXPECT_NE(receivedData[1].find("value5"), std::string::npos);
+}
+
+// Verifies the next bulk after a cluster_block_exception is actually delayed by IndexerBulkQueue's
+// backoff, not just eventually sent. nextDelay() for a *first* occurrence is uniform in [0, baseDelay)
```
LucioDonda, line 937: This contradicts both the implementation (exponentialBackoff.hpp:579-582: if (failures == 0) return cap; — deterministic, exact base delay, no jitter) and the PR's own dedicated test IndexerExponentialBackoffTest.FirstFailureSleepsExactlyBaseDelay, which asserts EXPECT_EQ(backoff.nextDelay(), BASE_DELAY).

LucioDonda (commented):
Some additional test should be added

src/shared_modules/indexer_connector/src/indexerConnectorAsyncImpl.hpp
```diff
@@ -525,17 +589,20 @@ class IndexerConnectorAsyncImpl final
...
                     {
-                        LOGFN_DEBUG2(m_logFn, "Too many requests, retrying in 1 second.");
+                        LOGFN_DEBUG2(m_logFn, "Too many requests, retrying with exponential backoff.");
                         throw IndexerConnectorException(error);
                     }
                     else
                     {
-                        LOGFN_WARN(m_logFn, "%s, status code: %ld.", error.c_str(), statusCode);
+                        LOGFN_WARN(m_logFn,
+                                   "%s, status code: %ld. Discarding event(s) in this batch.",
```
jam300, line 603: Is statusCode == -1 (connection error) expected to be retried? It currently falls through to the non-retryable branch in both connectors: async logs the error and discards the batch, while sync throws without setting needToRetry. Is this the intended behavior, or should connection errors use the backoff retry path?

NahuFigueroa97 (approved):
LGTM!

### The cluster starts blocking writes; each occurrence is tallied (per-batch count / running total) and triggers a backoff request instead of being retried:

```

2026/07/20 16:35:18 ...WARNING: Index write(s) blocked by the cluster (cluster_block_exception) - 3187 document(s) in this batch (3187 total) not indexed due to this expected condition. Backing off future sends.
2026/07/20 16:36:42 ...WARNING: ... - 4629 document(s) in this batch (7816 total) ...
2026/07/20 16:37:03 ...WARNING: ... - 3914 document(s) in this batch (11730 total) ...
2026/07/20 16:37:45 ...WARNING: ... - 4058 document(s) in this batch (15788 total) ...
```


### Then it persists in consecutive sending attempts

```
2026/07/20 16:37:03 logger-helper[842197] indexerBulkQueue.hpp:214 at dispatch(): DEBUG: IndexerBulkQueue: pausing 1000 ms before next send (external backoff requested).
2026/07/20 16:38:31 logger-helper[842197] indexerBulkQueue.hpp:214 at dispatch(): DEBUG: IndexerBulkQueue: pausing 1842 ms before next send (external backoff requested).
2026/07/20 16:38:54 logger-helper[842197] indexerBulkQueue.hpp:214 at dispatch(): DEBUG: IndexerBulkQueue: pausing 2875 ms before next send (external backoff requested).
2026/07/20 16:39:18 logger-helper[842197] indexerBulkQueue.hpp:214 at dispatch(): DEBUG: IndexerBulkQueue: pausing 5932 ms before next send (external backoff requested).
2026/07/20 16:39:43 logger-helper[842197] indexerBulkQueue.hpp:214 at dispatch(): DEBUG: IndexerBulkQueue: pausing 13144 ms before next send (external backoff requested).
```

src/shared_modules/indexer_connector/src/indexerConnectorAsyncImpl.hpp
matigarciadev: To avoid regressions, I decided not to change the contract and the current error handling. `statusCode == -1` can occur for multiple reasons, and some of these can cause a non-temporary error, which could be a problem.

juliancnn (approved):
lgtm!

PR #37691, opened 2026-07-15
Validate destination paths when uploading list, rule, and decoder files

vikman90:
## Description
`upload_list_file`, `upload_rule_file`, and `upload_decoder_file` built their destination path from a caller-supplied `filename` without confirming it stayed inside the intended directory, allowing a file to be written to an unexpected location as the `wazuh` user when these functions are reached through the cluster callable path (`as_wazuh_object` → `DistributedAPI.run_local`), which bypasses the REST controller's filename format validators. The fix resolves the final destination with `realpath` and rejects the request if it falls outside the intended base directory, before any file is written.

## Proposed Changes
- `framework/wazuh/cdb_list.py`: in `upload_list_file`, added a `commonpath`/`realpath` check that the resolved destination stays under `common.USER_LISTS_PATH`; raises the existing `WazuhError(1805)` otherwise.
- `framework/wazuh/rule.py`: in `upload_rule_file`, added the same check against the validated `relative_dirname` under `common.WAZUH_PATH`; raises a new `WazuhError(1212)`.
- `framework/wazuh/decoder.py`: in `upload_decoder_file`, added the same check against the validated `relative_dirname` under `common.WAZUH_PATH`; raises a new `WazuhError(1508)`.
- `framework/wazuh/core/exception.py`: registered new error codes `1212` (Rule block) and `1508` (Decoders block) for the invalid-filename case.

### Results and Evidence
- Reproduced the out-of-bounds write with the reporter's payloads (`filename="../../../tmp/..."`) prior to the fix; the guard now raises before `upload_file`/`safe_move` is called.
- New tests fail on the pre-fix code (they reach past the intended guard into unmocked code, e.g. `validate_dummy_logtest` attempting a real socket connection) and pass after the fix, with `upload_file` asserted as never called.
- Full existing suites pass unmodified: `wazuh/tests/test_cdb_list.py`, `wazuh/tests/test_rule.py`, `wazuh/tests/test_decoder.py` (188 passed), `wazuh/core/tests/test_cdb_list.py`, `wazuh/core/tests/test_rule.py`, `wazuh/core/tests/test_decoder.py`, `wazuh/core/tests/test_utils.py` (398 passed), `wazuh/core/tests/test_exception.py` (10 passed).

### Artifacts Affected
`wazuh-manager` framework (Python API/framework layer): `wazuh.cdb_list`, `wazuh.rule`, `wazuh.decoder`, `wazuh.core.exception`.

### Configuration Changes
None.

### Documentation Updates
None.

[...]

PR #37838, opened 2026-07-21
Validate destination paths when deleting rule and decoder files

vikman90:
## Description
`delete_rule_file` and `delete_decoder_file` built their destination path from a caller-supplied `filename` without confirming it stayed inside the validated rules/decoders directory before removing it. This path is reachable through the cluster callable mechanism (`as_wazuh_object` -> `DistributedAPI.run_local`), which bypasses the REST controller's filename format validators, and mirrors the destination-path validation already applied to the upload functions in PR #37691. The fix resolves the final destination with `realpath` and rejects the request if it falls outside the intended base directory, before any file is removed — the same check already applied to `upload_rule_file`/`upload_decoder_file`.

## Proposed Changes
- `framework/wazuh/rule.py`: `delete_rule_file` now verifies `commonpath([realpath(full_path), realpath(base_path)]) == realpath(base_path)` before calling `remove()`; raises `WazuhError(1212)` otherwise.
- `framework/wazuh/decoder.py`: `delete_decoder_file` now applies the same check; raises `WazuhError(1508)` otherwise.
- `framework/wazuh/tests/test_rule.py`: added `test_delete_rule_file_path_traversal`.
- `framework/wazuh/tests/test_decoder.py`: added `test_delete_decoder_file_path_traversal`.

### Results and Evidence
- Verified against the patched, installed manager (`/var/ossec`) by driving the exact resolution/invocation flow the cluster protocol uses, rather than calling the function directly: the target function is serialized with `WazuhJSONEncoder` and resolved back with `as_wazuh_object` (the same call the master makes when deserializing a cluster request), then invoked as `f(**f_kwargs)` under the default RBAC context, exactly as `DistributedAPI.run_local` does.
  - `delete_rule_file`, invoked with `filename='../../../../../../tmp/poc_rule_traversal'`, is denied: `{'code': 1212, 'message': 'Invalid filename. It resolves outside of the expected rules directory.'}`, and `os.remove()` is confirmed not called.
  - `delete_decoder_file`, invoked the same way with an equivalent crafted `filename`, is denied: `{'code': 1508, 'message': 'Invalid filename. It resolves outside of the expected decoders directory.'}`, and `os.remove()` is confirmed not called.
- Before the fix, the same crafted requests fell through to the `exists()`/`remove()` block with no containment check (regression check: the new unit tests failed with error code `1906` instead of the expected `1212`/`1508`).
- Full suite: `pytest wazuh/tests/test_rule.py wazuh/tests/test_decoder.py` — 136 passed, no regressions.

### Artifacts Affected
`framework/wazuh` (manager Python framework/API).

### Configuration Changes
None.

### Documentation Updates
None.

### Tests Introduced
- `framework/wazuh/tests/test_rule.py::test_delete_rule_file_path_traversal`
- `framework/wazuh/tests/test_decoder.py::test_delete_decoder_file_path_traversal`

## Credits
Thanks to @namd0ng for reporting this issue.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 - Fixed AES connection getting reset to blowfish on keystore rebuild. ([#37524](https://github.com/wazuh/wazuh/pull/37524))
 - Fixed a deadlock in `wazuh-analysisd` that stopped alert generation when the Active Response queue filled up. ([#37764](https://github.com/wazuh/wazuh/pull/37764))
 - Restricted the upgrade commands accepted from the agent message channel in Analysisd. ([#37745](https://github.com/wazuh/wazuh/pull/37745))
+- Added destination path validation when deleting rule and decoder files. ([#37838](https://github.com/wazuh/wazuh/pull/37838))
 
 ### Agent
 
--- a/framework/wazuh/decoder.py
+++ b/framework/wazuh/decoder.py
@@ -421,6 +421,10 @@ def delete_decoder_file(filename: Union[str, list], relative_dirname: str = None
         if wazuh_error:
             raise wazuh_error
 
+        base_path = join(common.WAZUH_PATH, relative_dirname)
+        if commonpath([realpath(full_path), realpath(base_path)]) != realpath(base_path):
+            raise WazuhError(1508)
+
         if exists(full_path):
             try:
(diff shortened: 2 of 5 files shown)


jotacarma90 (approved):
LGTM

TomasTurina (approved):

PR #37849, opened 2026-07-21
Fix AWS discard_regex values containing spaces

### LINKED ISSUE #37709: `aws-s3` wodle: `discard_regex` values containing a space break argument parsing (`unrecognized arguments`)
# Description

Working on this [e2e-test](https://github.com/wazuh/wazuh/issues/37598). The `aws-s3` wodle's `discard_regex` option fails silently (well, semi-silently — see below) whenever the configured regex value contains a literal space. This isn't an edge case: `discard_regex` is documented and intended for filtering on human-readable category fields like `detail-type`, and every real AWS Security Hub event type value (`"Security Hub Findings - Imported"`, `"Security Hub Insight Results"`, `"Security Hub Findings - Custom Action"`) contains spaces. So this bug hits the primary documented use case for the option, not a corner case.

## Environment

- Wazuh 4.14.7-rc1 (pre-release)
- Reproduced on both a Wazuh manager (Ubuntu 22.04) and a Wazuh agent (Debian 12) — same root cause on both, since both run the same `aws-s3` wodle code
- Feature: `aws-s3` wodle, `subscriber type="security_hub"`, `discard_regex` option (`wodle-s3` reference doc, `cloud-security/amazon/services/prerequisites/dependencies.html` context)

## Steps to reproduce

1. Configure the wodle with a `discard_regex` value that contains a space, following the documented example style:
   ```xml
   <wodle name="aws-s3">
     <disabled>no</disabled>
     <interval>1h</interval>
     <subscriber type="security_hub">
       <sqs_name>YOUR_SQS_QUEUE_NAME</sqs_name>
       <aws_profile>YOUR_AWS_PROFILE</aws_profile>
       <discard_regex field="detail-type">Security Hub Insight Results</discard_regex>
     </subscriber>
   </wodle>
   ```
2. Restart `wazuh-manager` (or `wazuh-agent`).
3. Observe `ossec.log`.

## Actual behavior

At the default log level, only a generic failure is logged, with no indication of the actual cause:
```
WARNING: Subscriber: security_hub <queue>  -  Returned exit code 2
WARNING: Subscriber: security_hub <queue>  -  Error parsing arguments.
```

Enabling `wazuh_modules.debug=2` in `local_internal_options.conf` reveals the actual command being executed:
```
DEBUG: Launching S3 Subscriber Command: wodles/aws/aws-s3 --subscriber security_hub --queue <queue> --aws_profile <profile> --discard-field detail-type --discard-regex Security Hub Insight Results --debug 2
```
and the underlying Python error:
```
DEBUG: ... OUTPUT: usage: usage: aws-s3.py [options]
aws-s3.py: error: unrecognized arguments: Hub Insight Results
```

## Root cause

`wm_aws.c` builds the subscriber's command line as a plain, unquoted string and execs it via a helper that word-splits on whitespace (shell-style). The `discard_regex` value is not quoted/escaped when interpolated into that command line, so `--discard-regex Security Hub Insight Results` is split into five separate tokens (`--discard-regex`, `Security`, `Hub`, `Insight`, `Results`) instead of one flag + one value.

The Python script (`aws-s3.py`, via `argparse`) then does exactly what `argparse` should: consumes the first token (`Security`) as the value for `--discard-regex`, and has no matching flag/positional for the leftover tokens `Hub`, `Insight`, `Results` — so it aborts with `unrecognized arguments` and exit code 2. This affects only options whose values may contain whitespace; other wodle options (queue names, ARNs, profile names) are unaffected because AWS resource identifiers don't contain spaces.

## Impact

- Any `discard_regex` targeting `detail-type` (or any other naturally space-containing field) fails completely — the subscriber fetch aborts before it can discard *or* process anything, defeating the purpose of the option.
- The failure is effectively invisible at the default log level (`wazuh_modules.debug=0`): the WARNING text gives no hint that the configured `discard_regex` value itself is the problem. A user would have to know to enable debug level 2 to see the real `argparse` error.
- The documentation's own `discard_regex` example (`<discard_regex field="userIdentity.principalID">EXAMPLE6E4XEGITWATV6R</discard_regex>`) happens to use a space-free value, so this bug is not caught by following the doc's example verbatim — it only surfaces once you actually apply the option to a human-readable field, which is the realistic intended use.

## Workaround

Replace literal spaces (and other problem characters) in the regex with `.` (matches any single character in most regex engines, including Python's `re`):
```xml
<discard_regex field="detail-type">Security.Hub.Insight.Results</discard_regex>
```
This preserves the exact matching semantics against the real value (`.` matches the literal space too) while avoiding whitespace in the transported command-line string.

## Suggested fix

Fix belongs in `wm_aws.c`'s subscriber invocation: either properly quote/escape each interpolated value when building the command-line string, or — better — construct the argv as an array and exec it directly (e.g. `execve`/`execvp`-style) instead of composing a flat string that gets re-split. The current approach is fragile for any option value that can legitimately contain whitespace.

archit15singh:
## Description

Fixes #37709.

`aws-s3` built the `--discard-regex` value as an unquoted part of its command string. `wm_exec()` tokenizes that string before starting the Python collector, so a valid Security Hub value such as `Security Hub Insight Results` was split into separate arguments and rejected by `argparse`.

## Proposed changes

- Encode `discard_regex` as one quoted argument, escaping embedded double quotes and backslashes for `w_strtok()`.
- Apply the encoding consistently to AWS bucket, service, and subscriber command construction.
- Add a regression test covering the reported Security Hub value containing spaces.

### Results and evidence

The new unit test asserts that the subscriber command contains:

```text
--discard-regex "Security Hub Insight Results"
```

This preserves the supplied value as one token when `wm_exec()` calls `w_strtok()`.

- [x] `make -j4 TARGET=agent TEST=1` (macOS development environment)
- [x] `git diff --check`
- [ ] Targeted AWS C unit test: `test_wm_aws` (Linux build in progress)
- [ ] Compilation without warnings on every supported platform (maintainer CI)

### Artifacts affected

- AWS S3 wodle command construction in the Wazuh agent/manager module.

### Configuration changes

None. Existing `discard_regex` configuration values containing spaces, backslashes, or double quotes are passed to the collector without losing token boundaries.

### Tests introduced

Adds a unit test for a Security Hub subscriber whose `detail-type` discard regex is `Security Hub Insight Results`.

## Review checklist

- [x] Code changes reviewed
- [x] Tests cover the regression

### DIFF
--- a/src/unit_tests/wazuh_modules/aws/CMakeLists.txt
+++ b/src/unit_tests/wazuh_modules/aws/CMakeLists.txt
@@ -9,7 +9,7 @@
 list(APPEND tests_names "test_wm_aws")
 list(APPEND tests_flags "-Wl,--wrap=time,--wrap=w_time_delay,--wrap=w_sleep_until,--wrap=_mwarn,--wrap=_minfo,--wrap=_merror \
                          -Wl,--wrap=_mtwarn,--wrap=_mtinfo,--wrap=_mterror,--wrap=wm_aws_run_s3,--wrap=StartMQ,--wrap=FOREVER,--wrap=wm_sleep_until_interruptible \
-                         -Wl,--wrap=atexit -Wl,--wrap=wm_state_io -Wl,--wrap=w_query_agentd")
+                         -Wl,--wrap=atexit -Wl,--wrap=wm_state_io -Wl,--wrap=wm_exec -Wl,--wrap=w_query_agentd")
 list(APPEND use_shared_libs 1)
 
 list(APPEND shared_libs "../scheduling/wmodules_scheduling_helpers.h")
--- a/src/unit_tests/wazuh_modules/aws/test_wm_aws.c
+++ b/src/unit_tests/wazuh_modules/aws/test_wm_aws.c
@@ -23,6 +23,7 @@
 #include "../../wrappers/libc/stdlib_wrappers.h"
 #include "../../wrappers/wazuh/shared/debug_op_wrappers.h"
 #include "../../wrappers/wazuh/wazuh_modules/wmodules_wrappers.h"
+#include "../../wrappers/wazuh/wazuh_modules/wm_exec_wrappers.h"
 #include "../../wrappers/wazuh/shared/mq_op_wrappers.h"
 
 #define TEST_MAX_DATES 5
@@ -34,6 +35,7 @@ static OS_XML *lxml;
 extern int test_mode;
 
 extern void wm_aws_run_s3(wm_aws_bucket *bucket);
+extern void wm_aws_run_subscriber(wm_aws *aws_config, wm_aws_subscriber *exec_subscriber);
 
 void wm_aws_run_s3(wm_aws_bucket *exec_bucket) {
     // Will wrap this function to check running times in order to check scheduling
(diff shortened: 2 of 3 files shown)


jepalfer:
Hello @archit15singh!

Thank you for taking the time to dig into this and put together a fix, we really appreciate it. We had already started working on this issue internally a few days ago and we currently have a fix in progress, so we'll most likely move forward with our own PR to close this one.

That said, your contribution is genuinely valuable and it's great to see the same root cause independently confirmed. Thanks again for helping out!

Issue #37732, opened 2026-07-16
SPIKE: HTTPS server architecture and protocol definition for `wazuh-remoted`


TomasTurina:
## Description

As part of the HTTPS communications initiative (#37702), `wazuh-remoted` requires a new HTTPS server capable of handling agent communications over TLS while integrating with the existing Remoted core.

The current `wazuh-remoted` implementation is primarily written in C, while one of the non-functional requirements establishes that the new HTTPS server should be implemented in C++. This introduces architectural questions regarding the interaction between both languages and the long-term maintainability of the solution.

This spike aims to analyze the architecture of the HTTPS server, define its boundaries with the existing Remoted core, and evaluate whether both C++ and C interfaces or implementations will be required to support the current codebase and future developments.

In addition, this spike will define the transport-level HTTP contract exposed by the manager. This includes the HTTP endpoints, required headers, authentication and identification information, request body formats, response body formats, HTTP status codes, and error handling. The objective is to produce a stable API contract that allows both the agent and manager implementations to be developed independently using mocks before the final implementation is completed.

## Goals

* Analyze the architecture of an embedded HTTPS server inside `wazuh-remoted`.
* Define the interface between the C++ HTTPS server and the existing C Remoted core.
* Evaluate the feasibility and implications of maintaining both C++ and C interfaces or implementations.
* Evaluate candidate HTTP/TLS libraries considering:

  * Performance.
  * Scalability.
  * TLS support.
  * Dependency footprint.
  * License compatibility.
  * Ease of integration.
* Define the threading and concurrency model.
* Define request lifecycle and ownership of request/response buffers.
* Define error handling and graceful shutdown mechanisms.
* Evaluate how the server can support future protocol extensions.
* Define the HTTP API contract for every endpoint (`/stateless`, `/stateful`, `/control`).
* Define the HTTP headers required for agent authentication and identification.
* Define the request body schema for each endpoint.
* Define the generic response envelope and HTTP status codes.
* Produce mock requests and responses that can be used by both the agent and manager teams.

## Expected Deliverables

* Proposed HTTPS server architecture.
* Interface definition between the C++ server and the C Remoted core.
* Comparison of candidate HTTP/TLS libraries.
* Threading and concurrency proposal.
* Memory ownership model.
* HTTP API specification.
* Header definitions.
* JSON schemas for requests and responses.
* Mock HTTP requests and responses for every endpoint.
* Implementation recommendations and architecture documentation.

## Tasks

[...]

Issue #37702, opened 2026-07-15
HTTPS comms


vikman90:
## Description

The Wazuh agent-manager communication protocol, inherited from OSSEC, relies on UDP or a persistent TCP connection following a proprietary protocol encrypted with Blowfish or AES. This design introduces several limitations, such as difficulties in load balancing and in aligning with common industry standards.

The goal of this development is to add the capability for agents to connect to the manager over HTTPS. To achieve this, a new HTTPS server component will be added to `wazuh-remoted`. The Manager 5.0 will keep supporting legacy connections with 4.x agents.

## Functional requirements

1. The manager must expose an HTTPS server able to receive agent-initiated communications.
2. The server must support agent-to-manager submissions in batches:
   1. Events (stateless).
   2. States (stateful).
3. The server must support manager-to-agent requests, also delivered in batches:
   1. Active responses.
   2. Centralized configuration.
   3. Remote upgrades.
   4. Module requests.
   5. Metrics requests.
4. The following HTTPS endpoints must be provided:
   1. `/stateless`: receives a batch of event messages. Remoted relays them to the Engine.
   2. `/stateful`: synchronizes states with the manager, acting as a bridge to Inventory Sync. It must support sessions and batches of states, and the manager must return the corresponding response for each session.
   3. `/control`: receives control messages from the agent (see requirement 6).
   4. Requests of each type are independent and may be sent concurrently (e.g., the agent may send a stateless, a stateful, and a control request at the same time).
5. Every endpoint request must include, in the header, the agent identification in JSON format:

   ```json
   {
     "version": 1,
     "id": "001",
     "key": "710a9b3fd6e03bf1"
   }
   ```

6. The `/control` endpoint must handle the following control messages:
   1. **Startup**: sent when the agent starts up. The agent establishes a connection and sends its product version and its configuration checksum. Depending on `<allow_higher_versions>`, the manager accepts or rejects the connection. On acceptance, the manager returns the agent metadata (as in the current proposal).
   2. **Notify**: similar to keep-alives, the agent sends a notification to refresh its configuration checksum. The manager replies with an ACK that may include a batch of requests (active responses, upgrades, configuration or metrics requests, new centralized configuration, etc.).
   3. **Response**: the agent sends the responses to the requests that require one.
7. Authentication and authorization must work as follows:
   1. The agent must always hold a valid key (the `client.keys` file); the manager uses this key to authorize the connection.
[...]

Issue #37835, opened 2026-07-21
Agent `/stateless` events: split stateless egress off the shared buffer (06)


Nicogp:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0 | wazuh-agentd | Agent | Sources | Linux / Windows / macOS |

> Part of the agent-side HTTPS migration (#37702, agent spike #37738).
> Consumes the `/stateless` H/E contract from #37732. Depends on **01**.

## Description

Today stateless events and stateful sync data share a single path: modules write to
the DGRAM queue (`queue/sockets/queue`), `EventForward()` classifies each frame by its
first byte (`'s'` = sync/stateful, everything else = stateless), and **both** go
through the same leaky-bucket ring buffer (`buffer.c`) and the same `send_msg()` TCP
socket (`event-forward.c:30-51`).

Under HTTPS the two are separated onto different endpoints. This issue handles the
**stateless** half: stateless events must flow to the HTTPS client's `/stateless`
pipeline (byte-budget accumulator → H/E batch POST) instead of the shared leaky-bucket
buffer. The stateful half (the `'s'` frames) is **07**.

The `/stateless` body is the existing internal H/E wire format (#37732): one
`H {json metadata}` header line, then `E <queue>:<location>:<message>` lines, multi-line
events escaped with one leading space — essentially the frames the C intake already
holds, so batch assembly is nearly free.

**Requirement change (2026-07-21): configurable max payload + 413 adaptive resize.**
The maximum `/stateless` request payload must be **configurable** (`ossec.conf`),
defaulting to **1 MB**. If a batch still exceeds what the manager accepts, the manager
answers **`413 Payload Too Large`**. On 413 the agent must **not drop** the events in
that batch: it must **re-send them with a smaller payload**. Concretely, on 413 the
in-flight batch is split into smaller sub-batches (e.g. halve the effective size) and
the events are re-enqueued for sending — none of the events that were part of the
failed attempt are discarded. The mechanism needs a way to re-insert the failed
batch's events at the front of the accumulator (or an equivalent resend queue) and to
lower the effective batch size for subsequent flushes, recovering toward the
configured max once sends succeed again. The only event that cannot be salvaged this
way is a **single event larger than the minimum payload floor** — that one is dropped
and counted (same fate as today's `OS_MAXSTR` overflow), because it cannot be split.

## Goals

- Route stateless events from the agentd intake to `hc_submit_event` (the HTTPS
[...]

PR #37851, opened 2026-07-22
Agent HTTPS client foundation: transport, signing, retry and auth

### LINKED ISSUE #37828: Agent HTTPS client foundation: configuration, TLS, and replacing the legacy TCP transport (01)
|---|---|---|---|---|

> Consumes the transport/auth contract from #37732.

## Description

The agent currently maintains a single persistent, AES-encrypted OSSEC-protocol
TCP connection to the manager (`agt->sock`, driven by the `select()` monitor loop
in `src/client-agent/src/agentd.c:26-292`, with `send_msg()` in `sendmsg.c` and
`receive_msg()` in `receiver.c`). Under #37702 the agent must stop maintaining that
connection and instead act as an HTTPS client that initiates every exchange over TLS.

The C++ HTTPS client module already exists on the branch
(`src/client-agent/https_client/`, behind the C ABI `include/https_client.h`) and
its transport core is implemented (CMAC signing, libcurl/TLS, retry/back-off). What
is missing is the **production integration**: a real XML configuration surface, real
TLS/CA verification, the finalized CMAC key derivation, and the wiring that makes
HTTPS the agent's actual transport instead of a gated-off development scaffold.

This issue delivers that foundation. It does **not** implement the per-endpoint
behavior (startup/notify, tasks, download, stateless split, stateful sessions).

**Requirement change (2026-07-21): 401 ⇒ re-enrollment.** A `401` response on **any**
endpoint (`/stateless`, `/stateful`, `/control`, `/download`, `/stats`, `/config`)
means the agent's credential is no longer valid, so the agent must trigger
**re-enrollment** (the authd:1515 flow that rewrites `client.keys`) and resume with the
new key. The legacy failure handling — N connection retries then enroll, and the
spike's AuthError "re-sign and retry on a slow cadence" (#37738 §9) — is **removed**:
401 goes straight to re-enrollment. This is a transport-wide concern and belongs in
this foundation issue; the `/control` FSM's AuthError branch (02) is replaced by the
central 401→re-enroll handler defined here.

## Goals

- Parse the agent's HTTPS/transport configuration from `ossec.conf` into `hc_config_t`.
- Enforce real TLS with CA + hostname verification by default; fail closed.
- Finalize the AES-CMAC signing key derivation (resolve open point #4 from #37738).
- Promote the module from a gated development scaffold to the agent's transport,
  removing the legacy persistent-TCP data path.
- Keep enrollment (authd:1515) and `client.keys` untouched as a mechanism, but wire a
  central **401 ⇒ re-enrollment** trigger reachable from every endpoint.

### In scope
- XML config parser for the `<server>` / TLS / `<batch>` / notify settings, mapped 1:1 onto `hc_config_t`.
- Replacing the forced `HC_VERIFY_NONE` in the bridge with the real `verification_mode` model (§8 of #37738): `full` (CA + hostname, default), `certificate` (CA only), `none` (explicit opt-out + startup warning); fail-closed when a mode needs a CA and none is readable.
- CMAC key resolution from `client.keys` (the recipe decision — raw key vs. legacy MD5-mix derivation vs. dedicated derived key).
- Lifecycle wiring: build config, `hc_create`/`hc_start` at agent start, cooperative `hc_stop`/`hc_destroy` at shutdown; remove the `agent.https_client=1` gate (or invert it to a temporary kill-switch during stabilization).
- Retiring the legacy transport for **data**: the persistent `agt->sock`, the `connect_server()`/reconnect loop, and server rotation. `os_setwait` retained only as the one-time startup gate (§6 of #37738).
- Config-option disposition from §10 of #37738: single `<server>` (last-prevails), remove `<retry_interval>`/`<max_retries>`/`<time-reconnect>`/`<protocol>`/`<crypto_method>`, re-base `<client_buffer>` levels on byte occupancy, keep `<notify_time>`/`<auto_restart>`/`<config-profile>`/`<enrollment>`.
- **401 ⇒ re-enrollment handler:** a central action, reachable from all endpoints, that on a `401` outcome triggers the existing enrollment flow (authd:1515 → new `client.keys`), reloads the CMAC key, and resumes; guarded against tight re-enroll loops (back-off / single-flight so concurrent 401s from multiple streams enroll once). Remove the legacy "N retries then enroll" and the AuthError slow-cadence re-sign.

### Out of scope
- `/control` startup/notify semantics and hash handling → **02**.
- `/download` → **03**. Task dispatch → **04**. Upgrade → **05**.
- The stateless buffer split → **06**. `agent_sync_protocol` changes → **07**.
- Bumping vendored libcurl (tracked separately; ≥8.21.0 recommended).

- `https_client.h` C ABI is complete and backed by real implementations (`hc_create/start/stop/destroy`, `hc_submit_event`, `hc_submit_sync_*`, `hc_submit_task_response`, `hc_notify_now`, `hc_get_state`, plus the `hc_callbacks_t` table). The header flags CMAC key derivation as "pending (#37732)".
- Transport core (CMAC/TLS/retry) fully implemented; `curlPerformer::applyTls` already derives verifyPeer/verifyHost from `verify_mode`, wires CA path, optional mTLS cert/key, ciphers, no redirects.
- `https_client_bridge.c` is a **dev scaffold** (self-labeled): builds config from already-parsed `client.keys` + `<client>` block, but **forces `verify_mode = HC_VERIFY_NONE`** with a warning, has **no `<server>`/`<ssl>`/`<batch>` XML parser**, passes the raw key through unvalidated, and returns early unless `agent.https_client=1` (`getDefine_Int_default("agent","https_client",0,1,0)`).
- The module is compiled and linked into `agentd_lib` and the Windows `wazuh-agentd` target unconditionally (`client-agent/CMakeLists.txt`); only runtime activation is gated. `w_https_client_start()` is called from `agentd.c:136` with `atexit(w_https_client_stop)`.

## Tasks

- [ ] Design the `ossec.conf` schema for the transport block (align tag names with FR10 vs. the proposal sketch — open point #2 in #37738: `certificate_authorities`/`verification_mode` vs `ca_bundle`/`verify_hostname`; pick one).
- [ ] Implement the XML parser in the agent config layer, populating `hc_config_t` (server address/port, CA/cert/key, verification_mode, ciphers, batch size/interval, notify interval, agent id/key, version).
- [ ] Replace the bridge's forced `HC_VERIFY_NONE` with the parsed `verification_mode`; implement fail-closed behavior and the startup warning for `none`.
- [ ] Decide and implement the CMAC key recipe (open point #4 in #37738); validate the key via `keyProvider` (16-byte AES-128) and surface a clear error on mismatch.
- [ ] Wire real lifecycle: construct config from the parser, `hc_start` after config load, `hc_stop`/`hc_destroy` on shutdown; remove/replace the `agent.https_client` gate.
- [ ] Remove the legacy persistent-TCP data path (monitor-loop socket handling, `connect_server`, rotation, reconnect timers) behind HTTPS, keeping enrollment and the startup gate.
- [ ] Implement the central 401 ⇒ re-enrollment handler (invoke the authd:1515 enrollment flow, reload the CMAC key, resume); single-flight/back-off against re-enroll storms; classify 401 in `outcomeClassifier` to route here instead of slow re-sign.
- [ ] Remove the legacy "N retries + enroll" logic and the `/control` AuthError slow-cadence re-sign (superseded by the 401 handler).
- [ ] Implement the §10 config-option dispositions (removals, single `<server>`, `<client_buffer>` re-basing).
- [ ] Update the sample `ossec.conf` and internal-options docs; per-OS default CA path conventions.
- [ ] Unit tests for the parser and key derivation; black-box gtest over the C ABI for lifecycle; verify TLS `full` mode against a CA.

## Definition of Done

- The agent reads its HTTPS transport config from `ossec.conf` and starts the client with real TLS (CA + hostname) by default; misconfigured TLS fails closed.
- Requests are signed with the finalized CMAC key derived from `client.keys`.
- HTTPS is the agent's transport out of the box (no internal-option gate required); the legacy persistent-TCP data path is gone.
- A `401` on any endpoint triggers re-enrollment (new `client.keys` + reloaded CMAC key) and the agent resumes; the legacy retry-then-enroll and AuthError slow-cadence are gone; re-enroll storms are bounded.
- Enrollment mechanism and `client.keys` format are unchanged.
- Tests green (parser, key, lifecycle, TLS handshake); astyle clean.

## References

- #37732 auth/transport contract (canonical request, headers, timestamp window): https://github.com/wazuh/wazuh/issues/37732#issuecomment-4997334905
- #37738 §3 (C boundary), §4 (library), §8 (certificates), §10 (config dispositions), open points #1–#4.
- Code: `src/client-agent/https_client/`, `src/client-agent/src/https_client_bridge.c`, `src/client-agent/src/agentd.c`, `src/config/src/client-config.c`, `src/client-agent/src/start_agent.c`.

jr0me:
Part of #37828.

**Stack (1/7)** — base of the agent HTTPS client stack. Merge order: this PR first; each later PR is based on its predecessor and auto-retargets to the integration branch `enhancement/37831-agent-https-client` as predecessors merge.

## Transport core

A C++17 shared library behind a single C ABI header (`include/https_client.h`), mirroring the manager-side `remoted_module` shape: POD config, injected callbacks, opaque handle, exceptions never cross into C.

- Typed configuration with **fail-closed TLS validation**: a verifying mode without a readable CA never sends a byte.
- **AES-CMAC request signing** per the #37732 contract: canonical request, `protocol-version` + `Authorization` pair, AES-128/192/256 by key length (a real client.keys entry is 64 hex = AES-256), in-memory and streamed-file MACs.
- libcurl seam (performer over an injectable handle) with response-to-file streaming for large downloads (the target is opened `O_NOFOLLOW`, owner-only).
- Full-jitter retry with `Retry-After`-aware back-off and outcome classification (D9).
- **401 → re-enrollment latch** (#37828): any 401 pauses all traffic and fires `on_reenroll_required` once per incident (one atomic state, so it can never end fired-but-unpaused); `hc_set_agent_key()` swaps the credential at runtime and resumes.
- Single-thread callback dispatcher (no concurrent/reentrant C callbacks); spool-file utilities (created atomically with `O_EXCL`, owner-only, unpredictable names); facade lifecycle skeleton.

## Agent integration

- **HTTPS is the agent's transport unconditionally** — no internal-option gate; `client-agent/src/main.c` refuses to start without a validated server. (The legacy TCP *data path* the module still runs alongside is retired in a later #37702 workstream.)
- **`<server>`/`<ssl>` config parser** (`src/config/src/client-config.c`): address/port canonicalisation and the TLS block feed a real `verify_mode`/CA/cert/key/ciphers into the module's fail-closed validation.
- **401 → authd re-enrollment**: `on_reenroll_required` runs the existing `try_enroll_to_server` flow on a detached thread (incremental back-off), then `hc_set_agent_key()`; the reload is mutex-guarded against shutdown teardown.
- **`.state` mapping**: the module's connection FSM feeds the agent `.state` file (REGISTERED→ACTIVE, STARTING→PENDING, else NACTIVE).
- Packaging + CI: installer, wxs, signed-exes, agent check_files, ci utils/build_tools/run_check, win32 link, the module RTR workflow.

## Verification

122 unit + 7 component tests (module) plus the agentd bridge cmocka suite. The component suite runs the real curl path against a fork-based fake manager that **recomputes the CMAC server-side** (cross-implementation auth interop, plaintext), and a dedicated case proves a genuine **`HC_VERIFY_FULL` handshake** against an on-disk CA (with a negative case asserting an untrusted cert is refused).


jr0me:
Pushed five commits addressing the review (foundation head now `72f97d2981`); the rest of the stack (#37852–#37857) was rebased over them and is mergeable again.

**Standards**
- **Spool symlink attack (P1)** — `563fa5cb3c`: `TempSpoolFactory` now creates the file with `open(O_CREAT|O_EXCL|O_WRONLY, 0600)` (an equivalent `_sopen_s` on Windows), so a pre-existing symlink fails the create instead of being followed/truncated, and the descriptor is owner-only. The name gains the pid + a per-thread random token (honestly unique across restarts, unpredictable). Tests cover owner-only perms and symlink rejection.
- **Re-enroll use-after-free (P1)** — `2aadd5a9e0`: the detached thread's key reload and `w_https_client_stop()`'s teardown are now serialized by a mutex — either the thread sees the stopping flag and abandons without touching the handle, or it reloads on a still-valid handle while stop() blocks behind it before `hc_destroy()`. Not held across destroy.
- **AuthGate non-atomic transition (P1→P2)** — `f084dfce39`: collapsed the two independent atomics (`paused`, `fired`) into one atomic state, so a report can never end fired-but-unpaused. A stale 401 racing `release()` now costs at most one harmless extra re-enroll. Added a TSan-exercised concurrency test.
- **IPv6 base URL (P2)** — `276a652010`: `baseUrl()` brackets bare IPv6 literals (`https://[2001:db8::1]:443`).

**Spec**
- **TLS verification evidence (P2)** — `72f97d2981`: added a self-contained component test doing a real `HC_VERIFY_FULL` handshake against an on-disk CA (SAN cert), plus a negative case that asserts an untrusted cert is refused (`TlsFail`). The PR description is corrected — full-verify is now covered by a genuine handshake, not just mocked curl options.

**Keystore reload race (finding, deferred)** — real, but only while the legacy TCP sender is still active; the reader-side keystore lifetime lock belongs with the **TCP cutover**, which by design no PR in this stack owns yet (see the mapping comment). Filing/parking it there rather than half-fixing it here.

**Not changed**
- The **TCP cutover** (retiring `agt->sock`, retry/rotation, multi-server) — agreed it's a separate workstream; none of PRs 2–7 claim it. The multi-server "collapse" is intentional (#37702 restriction 2: single manager) and warns loudly; it rides with the cutover.
- **Default port 1514 / commented-out CA** — the port default is shared with the still-active legacy path, so it can't flip to 443 until TCP retires; same package. During the interim legacy still delivers and HTTPS logs its refusal.
- **`<batch>` parsed-but-ignored** — deliberate: the parser defers batch mapping to the events module (comment + mdebug cite #37835 / issue 06, its owner), not a foundation bug.

Two of these (re-enroll UAF, AuthGate) touch code from the recent foundation commits — flagging for visibility, @Nicogp.

The dormant `curlPerformer` response-file `fopen("wb")` becomes reachable only once the /download fetcher sets `responseFilePath` (in #37853), so its `O_NOFOLLOW`/exclusive-create hardening belongs there; I'll add it on that PR rather than here.

Nicogp (approved):

PR #37852, opened 2026-07-22
Agent /control lifecycle: startup, notify and shutdown

jr0me:
Part of #37830.

**Stack (2/7)** — based on #37851 (foundation). Only the last two commits are this PR's.

The `/control` plane per #37733 §5.1: a pure state machine (Starting/Registered/Rejected/AuthError/Stopping) driven by a control loop on its own thread.

- **startup**: carries only `type` + `version`; the accepted response is forwarded opaquely (limits etc.), its SHA-256 becomes the settings baseline, and the **manager-authoritative cluster identity** is stored (overwritten unconditionally, cleared when absent).
- **notify**: a bare keepalive; a reported `settings_hash` mismatch refreshes the startup data **in place** — the client never leaves REGISTERED, and a warn-once latch prevents a startup storm if the manager's hash never matches.
- **shutdown**: a graceful stop sends `{"type":"shutdown"}` so the manager marks the agent disconnected at once.
- 426 demotes to a slow retry cadence; the 401 pause (#37828) converges the machine to AUTH_ERROR and recovers through `hc_set_agent_key` with a fresh registration.

Also introduces the live demo: a standalone C driver (the production `libhttps_client`, driven exactly as agentd would) against a python TLS mock that verifies the AES-CMAC of every request through the openssl CLI — an independent implementation, so auth interop is visible on screen. `./run.sh` / `./docker-run.sh` (self-contained).

Verification: 157 unit + 12 component; e2e over real TLS: registration, the settings-flip refresh, and the key-rotation round trip (401 → exactly one re-enroll → re-register).

Issue #37834, opened 2026-07-21
Agent remote upgrade over HTTPS: WPK download, verification, and fire-and-forget install (05)


Nicogp:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0 | wazuh-agentd / upgrade module | Agent | Sources | Linux / Windows / macOS |

> Part of the agent-side HTTPS migration (#37702, agent spike #37738).
> Consumes the `remote_upgrade` task + `/download` WPK contract from #37733. Depends on **03**, **04**.

## Description

`remote_upgrade` restarts the agent as part of executing it. Under the edited #37733
contract (2026-07-21) it is **fire-and-forget**: the agent does **not** send a `/control`
task-response with the result. The manager infers success from the agent reconnecting on
the new `version` in its next `/control` startup (02) — there is no `upgrade_result` query
anymore.

**The upgrade module no longer reports the result at all, and that removal is owned by the
server team — not this issue.** The old upgrade-module result path was dropped server-side
because it conflicted with one of their in-flight developments; consequently there is **no
POST anywhere that carries the upgrade result**. This agent issue therefore neither sends
nor is expected to send one, and it does not modify the upgrade module's result-reporting
code (the server team handles retiring it).

The manager delivers only WPK metadata in the task payload (#37733 Appendix A.2):

```json
{ "wpk_file": "wazuh_agent_v5.0.1_linux_x86_64.wpk",
  "wpk_sha1": "da39a3ee5e6b4b0d3255bfef95601890afd80709",
  "installer": "upgrade.sh" }
```

The agent must: download the WPK via `/download` (03), verify it against `wpk_sha1`, hand
the WPK + `installer` to the upgrade module, and run the installer (which restarts the
agent). No result is reported back.

**Idempotency instead of a result.** Because there is no response, the only correctness
requirement across the restart is **not upgrading twice**. That is handled by 04's durable
`task_id` registry: 04 records the upgrade `task_id` **before** the installer runs, so if
the manager re-delivers the same `task_id` after the restart, 04 discards it. This issue
therefore needs **no private persistence and no result columns** — it relies on 04's
record-before-dispatch ordering guarantee.

## Goals

- Implement the `remote_upgrade` handler behind the dispatch seam from **04**.
- Download the WPK via **03**, verify SHA1, hand off to the upgrade module, run the installer.
- Ensure the upgrade is idempotent across the restart via 04's durable dedup (no result message).

### In scope
[...]

PR #37853, opened 2026-07-22
Agent /download client for the centralized configuration

### LINKED ISSUE #37832: Agent `/download` client: chunked retrieval of centralized config and WPK packages (03)
|---|---|---|---|---|

> Consumes the `/download` contract from #37733 §5.2. Depends on **01**.

## Description

#37733 §5.2 replaces inline base64 file delivery with a dedicated `/download`
endpoint that streams large files (centralized config and WPK packages) using HTTP
chunked transfer encoding (64 KB chunks). The agent posts a small JSON request and
receives the file as a binary stream:

```json
{ "resource_type": "config", "resource_id": "web-servers" }
{ "resource_type": "wpk",    "resource_id": "wazuh_agent_v5.0.1_linux_x86_64.wpk" }
```

The agent must implement the `/download` client: sign the request (CMAC, like every
endpoint), stream the chunked response to disk without buffering the whole file in
memory, and expose a reusable API for its two consumers — the centralized-config sync
(triggered by **02** on a `config_hash`/`groups` change) and the remote upgrade
(**05**, WPK). This issue also implements the **config apply**: writing the downloaded
`merged.mg`, and reloading configuration.

## Goals

- Add a `/download` client to the HTTPS module that streams a chunked response to a
  destination file with flat memory use.
- Provide a C-ABI entry point usable from the config-sync path (02) and the upgrade
  path (05).
- Implement centralized-config apply: write the new `merged.mg`, update the stored
  `config_hash`, and reload the agent (`reloadAgent()` under `<auto_restart>`).
- Handle download failures with retry (reuse the module's back-off/classification).

### In scope
- A `/download` request signer + chunked-response reader streaming to a temp file,
  then atomic rename into place (config) or handoff (WPK).
- CMAC signing of the `/download` POST; TLS reuse of the module's connection posture.
- Retry on interrupted transfer (an interrupted chunked transfer must be treated as
  failure, never silent success — lesson H3 from #37738 §4).
- Config consumer: write `merged.mg`, recompute/store hash, trigger reload.
- WPK consumer: expose the downloaded file path + verification hook for 05 (SHA1
  verification and installer execution belong to 05).

### Out of scope
- Detecting *when* to download config → **02** (hash comparison).
- WPK SHA1 verification, installer execution, upgrade-module handoff, task-id
  persistence → **05**.
- The manager side of `/download`.

- **Not present.** Grep for `download|wpk|upgrade` across `https_client/src` + `include`
  returns only a doc comment in `https_client.h` noting that `upgrade`/`info_request`
  tasks expect a response. No download code, no chunked-response reader.
- The branch's older `/control` path delivered config inline (`config` `{hash,data}`
  → `onConfigUpdate`), which §5.2 removes in favor of `/download`. The inline path
  and its `bridge_on_config_update` logging stub are superseded by this issue.
- The module already has the building blocks to reuse: `curlHandle` (chunked write
  callback, header capture), `spoolFile` (RAII temp files), `outcomeClassifier` +
  `backoff`/`retrySender` for retry.

## Tasks

- [ ] Design the `/download` client API (C ABI + C++ internals): `resource_type`, `resource_id`, destination path, result/verification hook.
- [ ] Implement the CMAC-signed `/download` POST and the chunked-response reader (stream to temp file, 64 KB chunks, size accounting).
- [ ] Treat interrupted/short transfers as retryable failures; wire the existing back-off.
- [ ] Implement the config consumer: write `merged.mg`, update stored `config_hash`, reload under `<auto_restart>`.
- [ ] Expose the WPK consumer entry point (download to a known path, return path + expose a hash-verification hook for 05).
- [ ] Replace the bridge's logging-only `on_config_update` with the real write+reload path.
- [ ] Unit tests for the chunked reader (well-formed, truncated, zero-length, oversize); component test downloading a config and a binary blob from a mock.

## Definition of Done

- The agent can `POST /download` for `config` and `wpk`, streaming the response to
  disk with flat memory; interrupted transfers retry.
- A `config_hash`/`groups` change (from 02) results in the new `merged.mg` written and
  the agent reloaded.
- The WPK download path is available for 05 with a verification hook.
- Tests green; astyle clean.

## References

- #37733 §5.2 `/download` (chunked, 64 KB, request/response shapes), config-update and upgrade sequence diagrams (comments 4996820215, 5007622902; OpenAPI `DownloadRequest` in 5027414121).
- #37738 §4 (curl wrapper gaps incl. interruption handling), §7 (config_update handling).
- Code: `src/client-agent/https_client/src/curlHandle.cpp`, `spoolFile.cpp`, `controlStream.cpp` (`dispatchConfig`), `src/client-agent/src/https_client_bridge.c` (`bridge_on_config_update`).

jr0me:
Part of #37832.

**Stack (3/7)** — based on #37852 (control lifecycle). Only the last commit is this PR's.

The centralized-config pull per #37733 §5.2: every notify response carries the manager's `agent.config_hash`; a mismatch against the local merged.mg hash triggers a signed `POST /download` for the reported group.

- The body streams through the transport's response-to-file path into a spool file (never through memory) and is verified against the advertised SHA-256 **before** it is handed anywhere.
- Delivery via `on_config_downloaded`: the temp file lives exactly as long as the callback; writing merged.mg / unmerging / reloading stays the consumer's job.
- The adopted hash is optimistic; a consumer whose apply fails corrects it with the new callback-safe `hc_set_config_hash()` and the next notify re-downloads. Failed or corrupted downloads simply retry on the next notify.
- The fetcher runs blocking on the control thread with two attempts, governed by the large-transfer timeout. WPK packages (#37834) ride this same endpoint later.

Verification: 164 unit + 14 component; e2e over TLS: the chunked `/download` arrives byte-exact through the real curl path and is delivered exactly once.


jr0me:
Added `6bdc8d119c`: the /download response file is now reopened with `open(O_WRONLY|O_CREAT|O_TRUNC|O_NOFOLLOW, 0600)` instead of `fopen("wb")`. This is the sibling of the spool-file hardening on #37851 — the fetcher pre-creates the target owner-only, but the reopen still followed a symlink, so a swap in a shared spool dir could truncate an arbitrary victim (bytes are hash-verified, so integrity/DoS, not disclosure). A symlink at the path now fails the open (the download aborts and retries) rather than being written through. Component test covers the refusal + victim intact. Windows keeps fopen.

jr0me:
Follow-up on the review.

**Fixed on this PR:**

- Download size accounting. `ConfigFetcher` now caps the streamed response (64 MB for config) and aborts plus discards a transfer that exceeds it, so a hostile or faulty manager cannot fill the disk. Added a component test that exercises the cap through the real curl path.
- Windows symlink parity. The response file open now refuses reparse points on Windows (`FILE_FLAG_OPEN_REPARSE_POINT` plus a reparse check), matching the POSIX `O_NOFOLLOW`. Mingw compile checked.
- Documented the single-shot lifecycle in the public header (start after stop returns false).

**Deferred, tracked here so we do not lose them.** These sit outside this transport slice, they are not regressions in this PR:

- Applying `merged.mg` (write, unmerge, verify with `getsharedfiles()`, reload) and resetting the hash when an apply fails. The bridge consumer is a documented dev-scaffold stub today. This is the config-application integration and wants its own linked issue (open scope on #37832; the issue author flagged it as well).
- Generic `/download` client plus the WPK entry point and verification hook: #37834. WPK rides this same endpoint.
- Multi-group download identifier: pending the manager contract on #37733. A single reported group cannot match a combined `merged.mg` hash. Flagged in `firstGroup()`.
- SHA-256 local config hash to replace the MD5 seed from `getsharedfiles()`, plus renaming `config_checksum` to `config_hash`. Rides with the config-application work above.
- Demo driver milestone checks. The runners now propagate the driver exit, but the driver still does not fail when the registration, settings-refresh, or re-enrollment milestones never happen.

This PR stays scoped as Part of #37832.

Issue #37831, opened 2026-07-21
Agent-side HTTPS migration — implementation issues


Nicogp:
Real-implementation breakdown for the agent side of the HTTPS communications
initiative (**#37702**), following the agent spike **#37738** and consuming the
manager-side contracts **#37732** (transport / AES-CMAC auth) and **#37733**
(task storage + `/control`/`/config`/`/stats`/`/download` protocol).

> **All #37733 comments were edited 2026-07-21 ~19:00 UTC; this breakdown reflects the
> edited versions** (verified by re-fetching and diffing). Net effect of the edits:
> `/control` = startup/notify/**shutdown** (no `response`); **all tasks fire-and-forget**
> (no task results); 4 task types; config/stats pushed to indexer via `/config` + `/stats`;
> no cluster broadcast/dedup. Endpoints: `/control`, `/config`, `/stats`, `/download`
> (+ the data planes `/stateless`, `/stateful`).

Target branch: **`enhancement/37738-https-client`** (based on `5.0.0-https`).

## Contract sources (read these first)

- Auth / transport / `/stateless`: https://github.com/wazuh/wazuh/issues/37732#issuecomment-4997334905
- Task manager + `/control` + `/download` + sequence diagrams: #37733 comments
  4996820215, 5007622902, and the OpenAPI in 5027414121
- **Task Manager implementation plan (#37733 comment 5034967475, 2026-07-21)** — latest:
  4 task types, no cluster broadcast/dedup, config/stats moved to indices, `shutdown`
  control type, cluster-name authority. Folded into the issues (see requirement changes).
- Agent spike design + PoC + implementation notes: #37738 comments 5008246944, 5025965565

## What already exists on the branch (do NOT re-scope)

A substantially complete C++17 transport module lives at
`src/client-agent/https_client/` behind the C ABI `include/https_client.h`,
linked into `wazuh-agentd`. Already implemented and tested (175 unit + 14
component tests, TSan clean):

- **Transport core**: AES-CMAC signing (in-memory + streamed-file), libcurl/TLS,
  per-stream retry with full-jitter back-off, outcome classification.
- **`/stateless`**: byte-budget accumulator with 4-level occupancy ladder, H/E batch POST.
- **`/stateful`**: STREAM-socket intake (`queue-sync`) + spool file + streamed
  single-POST upload + `WZSY` binary framing.
- **`/control` transport**: FSM (Starting/Registered/Rejected/AuthError/Stopping),
[...]

PR #37821, opened 2026-07-21
HTTPS base server

juliancnn:
## Description

Introduces a self-contained **C++ worker module (`remoted_module`)** inside
`wazuh-manager-remoted` that exposes an **HTTPS ingestion path with per-agent AES-CMAC
authentication**, as the foundation for stateless event ingestion over HTTP/TLS (alongside the
existing AES-encrypted TCP/UDP channel on `1514`).

The module is intentionally layered so the HTTP library and the authentication logic are
independent and swappable: a transport-agnostic `IHttpServer` (RESTinio today, e.g. Boost.Beast
later), a framework-agnostic AES-CMAC auth layer, and an endpoints layer that wires them together.

**Status: experimental / work in progress.** The `POST /stateless` endpoint currently performs
**authentication and request validation only** — it does not parse the H/E payload or ingest
events yet, and the HTTPS listener only starts when a TLS certificate/key are present. It is
Linux-manager only.

Closes #37732

## Proposed Changes

- **HTTPS transport (`src/remoted/remoted_module/src/http_server/`)**: RESTinio + OpenSSL listener
  behind a neutral `IHttpServer` interface (PImpl hides RESTinio). Async handlers dispatched to a
  bounded worker pool with **deferred responses** (I/O threads never block), TLS ≥ 1.2, and
  `SO_REUSEADDR` on the acceptor for clean rebind on restart. Config resolves
  **C-ABI struct → environment variable → default** (`buildHttpServerConfig`): bind `127.0.0.1`,
  port `9443`, I/O threads `2`, handler worker threads `4`, transport body cap `16 MiB`, TLS
  cert/key paths.
- **Agent authentication (`src/remoted/remoted_module/src/auth/`, ns `remoted::auth`)**:
  framework-agnostic `AuthMiddleware` implementing the AES-CMAC request-auth protocol (canonical
  request, timestamp window, constant-time MAC compare), `Cmac` over OpenSSL EVP_MAC, `Keystore`
  reading `client.keys` (`IAgentKeystore`), and shared `authTypes` (`AuthenticatedRequest`,
  `AuthError`, `publicErrorFor`, `AuthConfig`). Credential failures collapse to a single generic
  `401`.
- **Endpoints layer (`src/remoted/remoted_module/src/endpoints/`, ns `remoted::endpoints`)**:
  an endpoint contract (`endpoint.hpp`: shared types + async `AuthenticatedHandler`) and an
  `AuthGateway` that runs the auth middleware, hands the verified request **and** the responder to
[...]

PR #37854, opened 2026-07-22
Agent /control task dispatch: four types, batch planning, interim dedup

### LINKED ISSUE #37833: Agent `/control` task dispatch: route manager tasks to modules, dedup durably, and report results (04)
|---|---|---|---|---|
| 5.0.0 | wazuh-agentd / agent-info (wazuh-modulesd) | Agent | Sources | Linux / Windows / macOS |

> Consumes the task contract from #37733 §2 + §5.1 (edited 2026-07-21). Depends on **01**, **02**.

## Description

The manager delivers manager-to-agent work in the `tasks[]` array of the `/control`
notify response (#37733 §5.1). The agent dispatches each task to the correct handler.

**Contract (edited #37733, 2026-07-21): four task types, all fire-and-forget.** There is
**no `/control` task-response message** — the agent does **not** report task execution
results back to the manager (the OpenAPI dropped `ResponseRequest`/`TaskResult`, the DB
schema dropped `RESPONSE_DATA`/`ERROR_MESSAGE`, and the sequence diagram marks upgrade
"fire-and-forget"). `info_request` and `agent_reconnect` are removed.

| # | task_type | Payload | Response? | TTL |
|---|-----------|---------|-----------|-----|
| 1 | `active_response` | AR document | No | 5 min |
| 2 | `remote_upgrade` | WPK metadata | No (fire-and-forget) | 24 h |
| 3 | `agent_restart` | empty | No | 5 min |
| 4 | `agent_reload` (config reload) | empty | No | 5 min |

All four are dispatched to their existing handlers with no reply. `remote_upgrade` hands
its payload + `task_id` to the upgrade module (05); the manager infers upgrade success
from the agent reconnecting on the new `version` at startup (02), not from a task result.

**Durable dedup of received tasks.** Delivery is **at-least-once**, and the manager does
**no cluster-wide broadcast or dedup** (each node marks delivered locally only), so the
same `task_id` may reach the agent more than once (retries, cluster routing, replay within
the timestamp window). The **agent's durable dedup is the sole guarantee against double
execution.** The agent keeps a **durable, bounded** record of received `task_id`s to
recognize a duplicate and **discard it** before dispatch, and that record must **survive
restarts** — critically, `remote_upgrade` restarts the agent, so the upgrade's `task_id`
must be recorded **before** the installer runs, so a post-restart re-delivery is discarded
(this is what makes upgrade idempotent without any result message). This replaces the
in-memory LRU (`taskDeduper.hpp`) that loses state on restart.

**Store decision (2026-07-21): the registry lives in the agent-info module, reached over
IPC.** agentd (where tasks are received) does an atomic check-and-record against a
`task_id` table owned by agent-info. On Linux/macOS agent-info runs in wazuh-modulesd, a
separate process, so this is a local-socket round-trip on the dispatch path; on Windows
agent-info is in-process.

This issue implements dispatch + durable dedup. The `remote_upgrade` *specifics*
(download/verify/install) are in **05**; this issue only defines the dispatch seam and
hands the payload + `task_id` over.

## Goals

- Parse and dispatch `tasks[]` from the notify response (fire-and-forget, no result reported).
- Provide a **durable, bounded** received-task registry in agent-info (a `task_id`
  table, configurable max entry count + cleanup TTL) and atomically check-and-record
  every task over IPC so duplicates are discarded before dispatch, across restarts.
- Route each of the four `task_type`s to the correct existing agent handler.
- Guarantee the upgrade `task_id` is recorded before the installer runs (idempotent upgrade).

### In scope

**Durable task registry (agent-info + IPC):**
- A `task_id` table in the agent-info module store (schema: `task_id` PK, `received_time`).
- An IPC verb agentd calls to **atomically check-and-record** a `task_id` (check-then-
  insert must not race across concurrent notify responses); reuse the agent-info
  component socket path (wmcom / the module `com` channel) rather than inventing a new
  daemon socket.
- Config (`ossec.conf`): max entry count + cleanup TTL (+ interval), with defaults.
- Background cleanup in agent-info: prune by TTL and enforce the max-count cap
  (oldest-first).
- Retiring the in-memory `taskDeduper` in favor of this registry.

**Dispatch (fire-and-forget):**
- `tasks[]` parsing; discard a task whose `task_id` is already recorded; record on accept
  (before dispatch — mandatory ordering for `remote_upgrade`).
- Routing map:
  - `active_response` → execd (the existing AR path / `agt->execdq`).
  - `agent_restart` → the restart handling.
  - `agent_reload` → config reload.
  - `remote_upgrade` → the upgrade dispatch seam (payload + `task_id` handed to 05).
- Replacing the bridge's logging-only `bridge_on_task` with real routing.
- Retiring the `/control` response path on the branch (`sendPendingResponses`,
  `hc_submit_task_response`) — no task results are sent under the edited contract.
- Keeping task dispatch off the `/control` request thread (a slow handler, or the dedup
  IPC round-trip, must not stall the next notify) — dispatch on the existing dispatcher
  thread.

### Out of scope
- WPK download/verify/execute → **05** (relies on this issue's registry for idempotency).
- `info_request` — **removed** (agent config/stats are pushed via `/config` and `/stats` → **08**).
- `agent_reconnect` — **removed** (not needed with stateless HTTPS).
- The manager side (task storage, local delivery marking, index-based upgrade inference).
- `/download` mechanics → **03**.

- **Transport-only + in-memory dedup, and a now-obsolete response path.** `controlStream.cpp` `dispatchTasks` (:78-98) forwards each task via the `on_task(task_id, task_type, payload)` callback with an in-memory LRU dedup (`taskDeduper.hpp`, cap 4096, flat-memory). This dedup is **superseded** by the durable agent-info registry.
- `sendPendingResponses` (:198) and `hc_submit_task_response` build/send `{"type":"response",…}` — this whole path is **removed** by the edited (fire-and-forget) contract; it must be retired, not wired up.
- **No task execution, no routing by `task_type`, no handler wiring.** `bridge_on_task` (`https_client_bridge.c:44`) only logs (`mdebug1("... task received: id=%s type=%s")`) and discards the payload.
- agent-info module: C wrapper `src/wazuh_modules/src/wm_agent_info.c`, C++ impl `AgentInfoImpl` (`agent_info_impl.cpp`), already rides `agent_sync_protocol` with its own persistent-queue DB — the natural host for the `task_id` table.
- Legacy handler references: execd/AR via `agt->execdq` and `receiver.c` AR path; restart handling in the legacy receiver. (The old `info_request`/`getconfig` request path in `request.c` is **not** reused — config/stats move to 08; `agent_reconnect` is dropped entirely.)

## Tasks

- [ ] Define the agent-info `task_id` table schema and the `ossec.conf` options (max count, cleanup TTL + interval, defaults).
- [ ] Implement the table + an atomic check-and-record operation + background cleanup (TTL + max-count, oldest-first) inside agent-info.
- [ ] Add the IPC verb (over the agent-info component socket) for agentd to check-and-record a `task_id`; wire agentd's dispatch to call it.
- [ ] Parse `tasks[]` and classify by `task_type` (four types; reject/log unknown types); discard duplicates via the registry; record accepted ids **before dispatch**.
- [ ] Implement the routing map to existing handlers (AR/execd, restart, reload) and the `remote_upgrade` dispatch seam (hand payload + `task_id` to 05).
- [ ] Retire the `/control` response path (`sendPendingResponses`, `hc_submit_task_response`) and the in-memory `taskDeduper`; remove the `info_request` branch.
- [ ] Replace `bridge_on_task` logging with real routing; add `.state` metrics for dispatched/discarded-duplicate/failed tasks.
- [ ] Unit tests: table dedup across a simulated restart, cap + TTL cleanup, concurrent check-and-record, routing. Component test: deliver each task type from a mock (incl. a duplicate `task_id` before and after restart) and assert no response is sent and the duplicate is discarded.

## Definition of Done

- Received `task_id`s are recorded durably in agent-info **before dispatch**; a duplicate
  is recognized and discarded — even across a restart — and never re-dispatched.
- The registry is bounded by the configured max count and TTL; cleanup keeps it in bounds.
- Each of the four `task_type`s is routed to the correct handler; all are fire-and-forget
  (no `/control` response is ever sent); `info_request` and `agent_reconnect` are gone;
  the in-memory deduper and the response path are gone.
- Tests green; astyle clean.

## References

- #37733 (edited 2026-07-21): §2 four task types (comment 4996820215), `/control` = startup/notify/**shutdown** with **no response message**, no cluster broadcast, fire-and-forget upgrade (sequence diagrams 5007622902; OpenAPI 5027414121 drops `ResponseRequest`/`TaskResult`), Appendix A payload formats.
- #37732 replay/at-least-once semantics (duplicate delivery → agent-side dedup).
- #37738 §7 (type→handler mapping; dedupe by task_id — made durable here).
- Code: `src/client-agent/https_client/src/controlStream.cpp` (`dispatchTasks`, `sendPendingResponses` → retire), `taskDeduper.hpp` (to retire), `src/client-agent/src/https_client_bridge.c` (`bridge_on_task`), `src/wazuh_modules/src/wm_agent_info.c`, `src/wazuh_modules/agent_info/.../agent_info_impl.cpp`.
- Consumed by: **05** (upgrade dispatch seam + relies on the `task_id` registry for idempotency).

jr0me:
Part of #37833.

**Stack (4/7)** — based on #37853 (/download). Only the last commit is this PR's.

The notify response's task array becomes agent work. Four fire-and-forget types (#37733: no response leg, nothing reported back): `active_response`, `agent_reload`, `remote_upgrade`, `agent_restart`.

- **Per-batch planning**: dispatch ordered by priority (AR first, terminal actions last); redundancy collapsed — an upgrade covers a restart and a reload, a restart covers a reload — with a log line per drop. Unknown types rank last but still dispatch (forward compatibility).
- **Interim dedup**: an in-memory registry of seen task_ids with configurable bounds (`task_dedup_max`, default 4096, FIFO eviction; `task_dedup_ttl_s`, default 3600). Dedup runs **before** planning so a task dropped as redundant is still marked seen and its at-least-once redelivery stays dropped.
- This registry is explicitly interim (TODO #37833 in the code): the durable agent-info task_id table replaces it so dedup survives restarts — the manager no longer dedups cluster-wide, making the agent's registry the sole guarantee against double execution.
- Tasks reach the consumer through the new `on_task` callback, serialized on the dispatcher thread; payloads cross opaquely.

Verification: 185 unit + 14 component; the demo shows a deliberately shuffled 4-type batch dispatched AR → upgrade with restart+reload dropped.

---

### Additionally fixed here

**The startup hash gate no longer blocks module startup over HTTPS.** `syscheckd` and the other modules
wait on `startup_gate_wait_for_ready()`, and the gate's only feed was the legacy TCP handshake:
`start_agent.c` hands it the `merged_sum` the manager pushes on that channel. An agent whose transport is
HTTPS never received an expected hash, so the gate stayed at `waiting_handshake` and module startup hung on
Linux and Windows alike.

The module already knew the manager's config hash on every Notify but only surfaced it when a download
happened, which is precisely the case that does *not* occur when the agent is already in sync. It is now
reported every time through a new `on_manager_config_hash` callback, and the gate's existing logic handles
both paths: a matching hash releases it immediately, a diverging one records what it is waiting for until
`/download` lands the configuration.

One guard worth calling out: the bridge only forwards an **MD5-shaped** hash. The gate validates against
`getsharedfiles()` (`OS_MD5_File` over `merged.mg`) and latches itself blocked on `invalid_handshake_hash`
for anything else, which its own comment says needs an agentd restart to clear. The module verifies
downloaded configs with `sha256FileHex`, and no manager `/control` endpoint exists yet to settle which
digest it will report, so anything that is not 32 hex characters leaves the gate untouched rather than
wedging it. That inconsistency between the MD5 seed and the sha256 verification is pre-existing and worth
resolving once the manager contract is real.

**Also fixed while getting CI green** (all pre-existing, none introduced by this PR): the winagent `hc_*`
link break, moved into `DEPENDENCIES_O` so every winagent test resolves it including syscheckd's; the
`test_client-config_https` syscheck wraps; the `libhttps_client.so` size drift; a scan-build null-pointer
report; and two valgrind findings — the `ClassifierCase` raw-byte dump and the component-test deadlines,
which were failing under valgrind and skipping `hc_destroy`, producing the "definitely lost" blocks.

Verification on this head: 243 unit + 17 component; `test_https_client_bridge` 27, `test_startup_gate` 15,
`test_agentd` 15, `test_client-config_https` 18. astyle 3.1 and cppcheck 2.20 clean.

### DIFF
--- a/.github/actions/check_files/deb_linux_agent_amd64.csv
+++ b/.github/actions/check_files/deb_linux_agent_amd64.csv
@@ -10,7 +10,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12374168,0.1
 /var/ossec/lib/libdbsync.so,root,wazuh,750,file,-rwxr-x---,505104,0.1
 /var/ossec/lib/libagent_sync_protocol.so,root,wazuh,750,file,-rwxr-x---,218488,0.1
-/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,270552,0.1
+/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,340472,0.1
 /var/ossec/lib/libschema_validator.so,root,wazuh,750,file,-rwxr-x---,390232,0.1
 /var/ossec/lib/libsysinfo.so,root,wazuh,750,file,-rwxr-x---,2603392,0.1
 /var/ossec/lib/libfimdb.so,root,wazuh,750,file,-rwxr-x---,253344,0.1
--- a/.github/actions/check_files/deb_linux_agent_i386.csv
+++ b/.github/actions/check_files/deb_linux_agent_i386.csv
@@ -10,7 +10,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12358288,0.1
 /var/ossec/lib/libdbsync.so,root,wazuh,750,file,-rwxr-x---,525380,0.1
 /var/ossec/lib/libagent_sync_protocol.so,root,wazuh,750,file,-rwxr-x---,216020,0.1
-/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,270552,0.1
+/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,340472,0.1
 /var/ossec/lib/libschema_validator.so,root,wazuh,750,file,-rwxr-x---,426324,0.1
 /var/ossec/lib/libsysinfo.so,root,wazuh,750,file,-rwxr-x---,2792476,0.1
 /var/ossec/lib/libfimdb.so,root,wazuh,750,file,-rwxr-x---,273712,0.1
--- a/.github/actions/check_files/rpm_linux_agent_amd64.csv
+++ b/.github/actions/check_files/rpm_linux_agent_amd64.csv
@@ -74,7 +74,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12417720,0.1
 /var/ossec/lib/libdbsync.so,root,wazuh,750,file,-rwxr-x---,484376,0.1
 /var/ossec/lib/libagent_sync_protocol.so,root,wazuh,750,file,-rwxr-x---,220736,0.1
-/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,270552,0.1
+/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,324096,0.1
 /var/ossec/lib/libschema_validator.so,root,wazuh,750,file,-rwxr-x---,386024,0.1
 /var/ossec/lib/libsysinfo.so,root,wazuh,750,file,-rwxr-x---,2555328,0.1
 /var/ossec/lib/libfimdb.so,root,wazuh,750,file,-rwxr-x---,224672,0.1
--- a/.github/actions/check_files/rpm_linux_agent_i386.csv
+++ b/.github/actions/check_files/rpm_linux_agent_i386.csv
@@ -74,7 +74,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12362096,0.1
 /var/ossec/lib/libdbsync.so,root,wazuh,750,file,-rwxr-x---,541828,0.1
 /var/ossec/lib/libagent_sync_protocol.so,root,wazuh,750,file,-rwxr-x---,216520,0.1
-/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,270552,0.1
+/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,324096,0.1
 /var/ossec/lib/libschema_validator.so,root,wazuh,750,file,-rwxr-x---,398136,0.1
 /var/ossec/lib/libsysinfo.so,root,wazuh,750,file,-rwxr-x---,2717844,0.1
 /var/ossec/lib/libfimdb.so,root,wazuh,750,file,-rwxr-x---,259032,0.1
--- a/src/client-agent/https_client/demo/demo_driver.c
+++ b/src/client-agent/https_client/demo/demo_driver.c
@@ -98,6 +98,17 @@ static void on_reenroll_required(void *user_data)
     fflush(stdout);
 }
 
+static void on_task(const char *task_id, const char *task_type, const char *payload_json,
+                    void *user_data)
+{
+    (void)user_data;
+    printf("[+%7ld ms] >> TASK received: id=%s type=%s payload=%s\n", elapsed_ms(), task_id,
+           task_type, payload_json);
+    fflush(stdout);
+    /* All four contract types are fire-and-forget (#37733): nothing is
+     * reported back; a real agent routes them to their handlers here. */
+}
+
 /* The new-config delivery: the file lives only until this returns, so a real
  * consumer copies it here, applies it (write merged.mg, unmerge, reload) and
  * corrects the module's hash view if the apply fails. */
@@ -185,6 +196,7 @@ int main(int argc, char **argv)
     callbacks.log = on_log;
     callbacks.on_startup_result = on_startup_result;
     callbacks.on_reenroll_required = on_reenroll_required;
+    callbacks.on_task = on_task;
     callbacks.on_config_downloaded = on_config_downloaded;
     callbacks.on_state_change = on_state_change;
     callbacks.on_buffer_level = on_buffer_level;
@@ -212,10 +224,11 @@ int main(int argc, char **argv)
     printf("== forcing an out-of-cycle Notify ==\n");
     hc_notify_now(handle);
 
-    /* Let the control loop run: the mock flips its config at notify #3
-     * (-> /download), its settings at notify #5 (-> the client refreshes
-     * startup in place) and rotates its key at #7 (-> 401, one re-enroll
-     * callback, hc_set_agent_key recovery). */
+    /* Let the control loop deliver the fire-and-forget task batch (notify
+     * #2); nothing is reported back (#37733: no response message). Then the
+     * mock flips its config at #3 (-> /download), its settings at #5 (-> an
+     * in-place startup refresh) and rotates its key at #7 (-> 401, one
+     * re-enroll callback, hc_set_agent_key recovery). */
     nap(3000);
     hc_notify_now(handle);
     nap(7000);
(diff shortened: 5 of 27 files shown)


jr0me:
Verified this against #37733's `/control` contract as edited 2026-07-21 (the fire-and-forget rework). The implementation matches:

- Four task types (`active_response`, `remote_upgrade`, `agent_restart`, `agent_reload`) with priority + subsumption; the dropped `agent_reconnect`/`info_request` are tolerated and ranked last.
- `tasks[]` is read at the top level of the notify response (sibling of `agent` and `settings_hash`), each item `{task_id, task_type, payload}`.
- No response/ack leg: tasks are delivered through `on_task` and nothing is reported back.
- `payload` is passed through opaquely, so `remote_upgrade`'s `{wpk_file, wpk_sha1, installer}` transports here and is executed in #37834.

Dedup stays interim/in-memory with a `TODO(#37833)` for the durable agent-info registry. That follow-up is now contract-mandated rather than optional: per #37944 the manager assigns deterministic task_ids and does no cluster-wide dedup, so the agent's dedup is the only guard against re-execution across restarts.

Note the server `/control` endpoint doesn't exist yet (only `/stateless`, #37821), so this is validated against the mock/contract.

jr0me:
### The remaining red jobs come from the base branch, not from this PR

Traced the failures on `eadd4040fd`. The ones left are inherited from
`enhancement/37831-agent-https-client` being behind `5.0.0-https`, and reproduce with this PR's own commits
removed.

**Manager legacy unit tests.** `remoted/src/secure.c` on the base includes `remoted_module.h` and calls
`remoted_module_start`/`remoted_module_stop`, but the test wiring for that exists only on `5.0.0-https`:

- `src/unit_tests/wrappers/wazuh/remoted/remoted_module_wrappers.c`
- `include_directories(${SRC_FOLDER}/remoted/remoted_module/include)` in `src/unit_tests/remoted/CMakeLists.txt`
- `-Wl,--wrap,remoted_module_start -Wl,--wrap,remoted_module_stop` on the affected targets

So the build dies at 1% with `fatal error: remoted_module.h: No such file or directory`, followed by
undefined references to both symbols. Reproduced locally against the base alone, with none of this PR's
commits applied. Nothing here touches `src/remoted/` or `src/unit_tests/remoted/`: the full diff of this
branch under `src/unit_tests/` is two files, `config/CMakeLists.txt` and `winagent.cmake`.

**Why the base branch's own CI is green on the same commit.** Its run is `event=pull_request`, so it builds
the integration branch *merged into `5.0.0-https`* and picks the wiring up from there. PRs stacked on
integration build against integration alone and miss it. A green job on #37831 is not evidence that the
stack is fine.

**Updating the base is blocked on the server side for now.** Rebasing integration onto current
`5.0.0-https` (228 commits) applies cleanly except for one commit, `cb6b2862bc "Added comms base module to
remoted"`, which is integration's stale snapshot of a module upstream has since evolved and which should
simply be dropped. But the result then hits
[#37974](https://github.com/wazuh/wazuh/pull/37974): `loggerHelper.h` wraps the whole `Log` namespace,
`LogFn` included, in `#pragma GCC visibility push(hidden)`, so
`ModuleConfig::validate(IFsProbe const&, Log::LogFn const&)` is emitted as a local symbol rather than an
exported one and `https_client_unit_test` no longer links:

```
$ nm -C --defined-only build/lib/libhttps_client.so | grep 'ModuleConfig::validate'
00000000001c153c t ModuleConfig::validate(IFsProbe const&, Log::LogFn const&) const
                 ^ local, should be T
```

That is the same cross-DSO visibility regression #37974 describes, introduced by #37821 and confined to
`5.0.0-https`. The update also pulls in `external/llhttp` for RESTinio, so a `make deps` is needed after it.

### What is actually this PR's

Everything found in review has been fixed here and verified locally: the winagent `hc_*` link break (stubs
moved into `DEPENDENCIES_O`, which every winagent test links, so syscheckd's binaries link too), the
`test_client-config_https` syscheck wraps, the `libhttps_client.so` size drift, the scan-build null-pointer
report, and the valgrind findings — the `ClassifierCase` byte-dump and the component-test deadlines that
were failing under valgrind and skipping `hc_destroy`, which is what produced the "definitely lost" blocks.

Module suites on this head: 241 unit, 17 component. `test_https_client_bridge` 27, `test_buffer` 10,
`test_agentd` 15, `test_client-config_https` 18. astyle 3.1 and cppcheck 2.20 clean, winagent bridge
compiles with no warnings and no undefined atomic symbols.

Nicogp (approved):

PR #37855, opened 2026-07-22
Agent /stateless events: batching, occupancy ladder, adaptive 413

### LINKED ISSUE #37835: Agent `/stateless` events: split stateless egress off the shared buffer (06)

jr0me:
Part of #37835.

**Stack (5/7)** — based on #37854 (task dispatch). Only the last commit is this PR's.

The stateless egress per #37732: `hc_submit_event` drops event frames into a byte-budget accumulator; a sender thread flushes them as one H/E batch POST per interval (or earlier on the size threshold), gated behind registration.

- Buffer caps at N × batch size with drop-newest and a 4-level occupancy ladder (`on_buffer_level`), replacing the legacy client buffer levels so manager-side flood alerting keeps working. Back-pressure (503/Retry-After) honored inside the retry loop while the accumulator keeps absorbing.
- **Configurable max payload** (`batch_size_bytes`, default 1 MB) and **adaptive 413 handling**: a 413 does NOT drop the batch — the send simply does not consume, so events stay at the buffer front in order, and the effective payload halves; success doubles it back. Halving has no hard floor (a server cap below any fixed floor would wedge the stream forever); at one byte the snapshot still holds one whole event, so a persistent 413 converges onto the single-oversized-event drop (counted + warned) instead of livelocking.
- Snapshots cut at event boundaries; shutdown drains the backlog in bounded batches before the shutdown message.

Verification: 219 unit + 15 component; e2e: a manager that 413s oversized bodies receives every one of a 40-event burst **exactly once** after the split.

### DIFF
--- a/src/client-agent/https_client/demo/demo_driver.c
+++ b/src/client-agent/https_client/demo/demo_driver.c
@@ -133,6 +133,13 @@ static void on_state_change(int state, void *user_data)
     fflush(stdout);
 }
 
+static void on_buffer_level(int level, void *user_data)
+{
+    (void)user_data;
+    printf("[+%7ld ms] ~~ buffer level -> %d\n", elapsed_ms(), level);
+    fflush(stdout);
+}
+
 static void nap(int ms)
 {
     struct timespec t = {ms / 1000, (long)(ms % 1000) * 1000000L};
@@ -162,6 +169,7 @@ int main(int argc, char **argv)
     strncpy(config.agent_key, argv[3], sizeof config.agent_key - 1);
     config.verify_mode = HC_VERIFY_NONE; /* demo mock uses a self-signed cert */
     config.notify_interval_s = 2;        /* Notify every 2 s so we see a few   */
+    config.batch_interval_ms = 1000;     /* flush events every 1 s             */
     config.request_timeout_ms = 10000;
     config.backoff_base_ms = 200;
     config.backoff_cap_ms = 2000;
@@ -179,6 +187,7 @@ int main(int argc, char **argv)
     callbacks.on_reenroll_required = on_reenroll_required;
     callbacks.on_config_downloaded = on_config_downloaded;
     callbacks.on_state_change = on_state_change;
+    callbacks.on_buffer_level = on_buffer_level;
 
     printf("== creating + starting the https_client (target %s:%s) ==\n", argv[1], argv[2]);
     hc_handle *handle = hc_create(&config, &callbacks);
@@ -192,6 +201,14 @@ int main(int argc, char **argv)
     /* Give the control thread a moment to run Startup + register. */
     nap(800);
 
+    printf("== feeding 5 events into the /stateless stream ==\n");
+    for (int i = 0; i < 5; i++)
+    {
+        char frame[64];
+        int n = snprintf(frame, sizeof frame, "1:/var/log/syslog:demo event %d", i);
+        hc_submit_event(handle, (const uint8_t *)frame, (size_t)n);
+    }
+
     printf("== forcing an out-of-cycle Notify ==\n");
     hc_notify_now(handle);
 
@@ -204,9 +221,9 @@ int main(int argc, char **argv)
     nap(7000);
 
     /* Optional sustained mode: DEMO_SECONDS=<n> keeps the client alive after
-     * the scripted walkthrough so the periodic Notify traffic (and the key
-     * rotation at notify #7) stays visible. SIGINT/SIGTERM (Ctrl-C, docker
-     * stop) ends it early through the same clean drain. */
+     * the scripted walkthrough, submitting an event every 2 s so the periodic
+     * Notify + /stateless traffic stays visible. SIGINT/SIGTERM (Ctrl-C,
+     * docker stop) ends it early through the same clean drain. */
     const char *extra = getenv("DEMO_SECONDS");
     const long extra_s = extra ? strtol(extra, NULL, 10) : 0;
     if (extra_s > 0)
@@ -216,6 +233,25 @@ int main(int argc, char **argv)
         fflush(stdout);
         for (long tick = 0; tick * 2 < extra_s && !atomic_load(&g_stop); tick++)
         {
+            char frame[80];
+            int n = snprintf(frame, sizeof frame, "1:/var/log/syslog:sustained event %ld", tick);
+            hc_submit_event(handle, (const uint8_t *)frame, (size_t)n);
+
+            /* Every 10 s, burst enough events to exceed the mock's /stateless
+             * payload cap, so the 413 -> split -> resend path is visible. */
+            if (tick > 0 && tick % 5 == 0)
+            {
+                printf("== bursting 120 events (exceeds the /stateless cap -> 413 "
+                       "split/resend) ==\n");
+                fflush(stdout);
+                for (int burst = 0; burst < 120; burst++)
+                {
+                    char big[96];
+                    int m = snprintf(big, sizeof big,
+                                     "1:/var/log/burst:event %ld-%d payload-padding", tick, burst);
+                    hc_submit_event(handle, (const uint8_t *)big, (size_t)m);
+                }
+            }
             nap(2000);
         }
     }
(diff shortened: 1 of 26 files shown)


jr0me:
Auto-closed by the stack base-reorder (its commits briefly became a subset of its old base branch). The stateless work continues in #37949, now based on the integration branch so it can merge next.

PR #37857, opened 2026-07-22
Agent periodic /stats and /config push

### LINKED ISSUE #37843: Agent periodic `/stats` and `/config` push: report module stats and configuration (08)
|---|---|---|---|---|

> New requirement (2026-07-21), replaces `info_request`. Depends on **01**.

## Description

**Requirement change (2026-07-21):** the manager will **no longer pull** agent config
and stats via an `info_request` task (old task type 6, removed in 04). Instead the
agent **pushes** them periodically to two new endpoints:

- **`POST /config`** — the current configuration of all agent modules.
- **`POST /stats`** — the current stats/metrics of all agent modules.

Both are opt-in and independently controllable from `ossec.conf`: **two separate
toggles** (one for stats/metrics, one for config), **both disabled by default**. When a
toggle is enabled, the agent gathers the corresponding information from every module on
a configurable interval, signs the request (CMAC, like every endpoint), and POSTs it to
the matching endpoint.

The information itself is the same the agent can already produce on demand today (the
legacy `getconfig` / `getstate` component requests), so this issue is largely about
**aggregating all modules' output and driving it on a timer**, plus the two config
toggles and the two new signed POST paths.

**Manager-side model (#37733 update, 2026-07-21):** the manager stores what the agent
pushes in **dedicated indices** — `wazuh-agent-config` and `wazuh-agent-stats`, keyed by
**doc ID = `agent_id`** — and the API now **reads config/stats from those indices
instead of querying the agent** (the synchronous `GET /agents/{id}/config` and
`/stats` agent round-trips are removed). Consequences for the agent:

- The agent is the **sole writer** of these indices; if it never reports, the API has
  no data (the manager handles "never reported" as an error). So the push must be
  reliable when enabled.
- Each push is an **upsert of the whole document** for this `agent_id` (latest wins),
  so the payload must be a **complete** config/stats snapshot, not a delta.
- The agent must **define the JSON structure** of both documents (this issue), since the
  index mapping and the API responses are shaped by it. Include the identity fields the
  index needs — `agent_id` and the **cluster name** (see below).

> The `/stats` and `/config` endpoints and their document JSON are **new** and not yet in
> the #37732/#37733 OpenAPI. The exact document structure, cadence defaults, and
> response/status codes must be **co-designed with the manager side** (index mappings)
> before the wire format is frozen. Defining the JSON structure is a deliverable of this
> issue.

**Cluster name (2026-07-21):** both documents carry the agent's cluster name, and it must
be the **manager-authoritative** value applied in **02** (overwritten from the startup
response, including when empty/unknown) — not a stale local config value — so the indexed
data is tagged with the cluster the manager actually recognizes.

## Goals

- Add two `ossec.conf` toggles (stats push, config push), both **off by default**, each
  with its own send interval.
- When enabled, periodically gather **all** modules' config (for `/config`) and stats
  (for `/stats`) and POST each to its endpoint as a CMAC-signed request.
- Define the **JSON document structure** for both (complete snapshot, includes
  `agent_id` + manager-authoritative cluster name), suitable to upsert into the
  `wazuh-agent-config` / `wazuh-agent-stats` indices keyed by `agent_id`.
- Reuse the existing per-module config/stats gathering rather than inventing new
  producers.

### In scope
- Config parser for the two toggles + their intervals (+ any per-endpoint options),
  defaulting disabled.
- Aggregators that collect config from all modules (→ `/config`) and stats from all
  modules (→ `/stats`), reusing the existing component `getconfig`/`getstate`
  machinery (the `com`/agcom component sockets that `request.c` fans out to today).
- Two new signed POST paths in the HTTPS module (`/config`, `/stats`), each on its own
  timer, honoring back-pressure (`Retry-After`) and the retry/back-off policy.
- Running these on a cadence that does not block the other endpoint threads (their own
  worker/timer; consistent with the one-thread-per-endpoint invariant in `00-README`).
- `.state` metrics for the two push paths.

### Out of scope
- The manager side of `/stats` and `/config` (the `wazuh-agent-config`/`wazuh-agent-stats`
  index mappings, ingestion, and the API reading from them).
- The freezing of the wire contract (co-design; this issue implements the agent half
  against the agreed shape) — but **defining** the document JSON is in scope here.
- Applying the manager-authoritative cluster name → **02** (this issue consumes it).
- `info_request` task handling — **removed** (see 04).
- Centralized-config *download* (`/download`, that is inbound config) → **03**; this is
  outbound reporting of the agent's effective config.

- **Not present.** No `/stats` or `/config` client, no periodic push, no toggles. The
  HTTPS module implements `/stateless`, `/stateful`, `/control` (and this epic adds
  `/download`); `/stats` and `/config` are new endpoints.
- Reusable machinery exists on the agent: per-component config via the `com` sockets
  and agcom (`getconfig`), and stats/state via the component `getstate` path and the
  agent `.state` reporting. These are what `request.c`/`req_receiver` fan out to today
  for on-demand queries; this issue drives the same producers on a timer.

## Tasks

- [ ] Define the `/config` and `/stats` **document JSON structure** (complete snapshot, `agent_id` + cluster name, per-module sections) and co-design it with the manager team against the `wazuh-agent-config`/`wazuh-agent-stats` index mappings; record the contract.
- [ ] Co-design cadence defaults + response/status codes with the manager team.
- [ ] Add the two `ossec.conf` toggles (+ intervals), both disabled by default; parse into config.
- [ ] Implement the config aggregator (all modules' effective config) and the stats aggregator (all modules' stats), reusing `getconfig`/`getstate`; assemble the full document incl. `agent_id` + the manager-authoritative cluster name from 02.
- [ ] Implement the two signed POST paths (`/config`, `/stats`) with their own timers, honoring `Retry-After`/back-off.
- [ ] Keep each push on its own worker/timer so it doesn't stall `/stateless`, `/stateful`, or `/control`.
- [ ] Add `.state` metrics and clear logs (including "disabled" at startup when off).
- [ ] Unit tests for aggregation + cadence + toggle gating; component test pushing both to a mock that verifies the CMAC and the shape.

## Definition of Done

- With both toggles off (default), the agent sends nothing to `/stats` or `/config`.
- With a toggle on, the agent periodically POSTs a **complete** config/stats document
  (all modules, incl. `agent_id` + manager-authoritative cluster name) to the endpoint,
  CMAC-signed, on its configured interval, without blocking the other endpoints — such
  that the manager can upsert it into `wazuh-agent-config` / `wazuh-agent-stats` by
  `agent_id`.
- The document JSON structure is defined and agreed with the manager side.
- The two toggles are independent.
- `info_request` is not used to obtain this data.
- Tests green; astyle clean.

## References

- #37738 open point #6 (metrics requests were under-specified) — resolved here by
  agent-push endpoints instead of a pull task.
- #37733 (2026-07-21 update, comment 5034967475): config/stats moved to indices
  `wazuh-agent-config` / `wazuh-agent-stats` (doc ID = `agent_id`); the API reads from
  the index, not the agent; synchronous agent config/stats queries removed.
- #37732/#37733 — `/stats` and `/config` endpoints + document JSON to be added to the
  transport/API contract (co-design pending).
- Code (reusable producers): `src/client-agent/src/request.c` (`req_receiver` fan-out to component sockets), agcom / per-module `com` `getconfig`/`getstate` handlers, agent `.state` reporting.
- Replaces: the `info_request` task path removed from **04**.

jr0me:
Part of #37843.

**Stack (7/7)** — last one. #37856 (/stateful) is merged, so this now sits directly on the integration branch and all four commits are its own.

The agent-push replacement for the removed `info_request` task: two independent reporters, both OFF by default, with their own intervals (stats 60 s, config 3600 s).

- A dedicated worker thread runs them behind the registration gate and the auth pause; the drain skips them (upsert semantics: a failed cycle's snapshot is simply superseded next time).
- The consumer supplies snapshots through two new `collect_stats` / `collect_config` callbacks. These run on the reporter thread (not the dispatcher) and return a malloc'd JSON object the module frees.
- Before sending, the module stamps `agent_id` and the **manager-authoritative cluster identity** from the startup response — overwriting anything the collector put there — so the indexed documents (doc id = agent id) are always coherent.
- Back-pressure honors Retry-After per path; other failures back off per path without touching the sibling's cadence.

## Aggregation: one query per daemon, one document

The reporters now have producers behind them. The agent's modules live in
separate daemons, so a complete picture costs one query per daemon — that is
the floor, and this is at it. Previously the same information took one query
per configuration section (15 across the five sockets).

Each daemon gained `getallconfig` (and `getallstats` where a stats producer
already exists) and answers with the modules it hosts:

```json
{"modules": [
  {"module": "fim",          "config": {"syscheck": {...}, "rootcheck": {...}}},
  {"module": "logcollector", "config": {"localfile": [...], "socket": [...]}},
  {"module": "syscollector", "config": {...}},
  {"module": "sca",          "config": {...}}
]}
```

`agent_id` and `cluster` are stamped on top by the reporter, so the document is
a complete snapshot ready to upsert by agent id.

Mapping, one entry per module:

| daemon | reports as | sections |
|---|---|---|
| agentd | `agent` | client, buffer, internal, anti_tampering |
| syscheckd | `fim` | syscheck, rootcheck, internal |
| logcollector | `logcollector` | localfile, socket, internal |
| modulesd | one entry per wodle, plus `wmodules` | each wodle's own dump; internal_options |
| execd | `execd` | active-response, logging, internal |

modulesd splits its wodles apart, taking each module's name from its own
`dump()`, so `syscollector` and `sca` stay individually addressable rather than
buried in one blob.

Failure behaviour is deliberate: a daemon that is down or refuses is left out
and the rest still report, so one sick component does not cost the manager the
whole document. When **nothing** answers the cycle is skipped rather than
pushing `{"modules": []}`, which would overwrite a good document in the index
with an empty one.

## Open points for the manager side (#38023 / #38024)

1. **Shape differs from the issue's example.** #38023 sketches a flat map
   (`{"client": {...}, "syscheck": {...}}`). This is an array of tagged
   entries, per the module-oriented shape asked for on the agent side, and
   because a flat map cannot express two modules living in one daemon.
   #37843 makes defining the JSON an agent-side deliverable, so this is the
   proposal — happy to converge.
2. **`internal` sections are included.** They bulk up the document, but
   `GET /agents/{id}/config` can currently be asked for `internal`, and #38023
   wants that endpoint served from the index. Dropping them would break it.
3. **Size ceiling.** A daemon's reply travels over its local socket, capped at
   `OS_MAXSTR` (64 KiB). A logcollector with a very large state exceeds that;
   today it is logged and the module is left out rather than truncated. Worth
   confirming what the manager wants for that case.
4. **`/stats` is thin.** Only agentd and logcollector have a `getstate`
   producer, so those are the only two modules with statistics. The rest need
   producers before they can report.

## Verification

- Unit: 107/107 in the agent suite (104 before; three new suites added).
  `test_module_report` (8), `test_agent_report` (5), `test_wmcom` (3),
  plus 4 new cases in `test_agcom`. ASAN-clean — it caught a leak in
  `module_report_merge` that is fixed here.
- `make TARGET=agent` clean; every changed C file also passes
  `i686-w64-mingw32-gcc -fsyntax-only`, which is the only path that compiles
  the Windows in-process dispatch branch.
- 245 unit + 23 component for the module itself; e2e: the fake manager
  receives both documents with the stamped identity over TLS + CMAC.

### DIFF
--- a/src/client-agent/https_client/demo/demo_driver.c
+++ b/src/client-agent/https_client/demo/demo_driver.c
@@ -167,6 +167,25 @@ static void nap(int ms)
     nanosleep(&t, NULL);
 }
 
+/* Reporter collectors (#37843): run on the module's reporter thread, return a
+ * malloc'd JSON object the module owns (frees) after stamping identity. A real
+ * agent aggregates every module's stats/config here. */
+static char *collect_stats(void *user_data)
+{
+    (void)user_data;
+    printf("[+%7ld ms] >> COLLECT stats\n", elapsed_ms());
+    fflush(stdout);
+    return strdup("{\"uptime\":86400,\"events_received\":12543,\"queue_usage\":15.2}");
+}
+
+static char *collect_config(void *user_data)
+{
+    (void)user_data;
+    printf("[+%7ld ms] >> COLLECT config\n", elapsed_ms());
+    fflush(stdout);
+    return strdup("{\"client\":{\"notify_time\":2,\"config-profile\":\"demo\"}}");
+}
+
 /* Deterministic hex filler (xorshift-style) for the fake file hashes. */
 static void fake_hex(char *dst, size_t hex_chars, uint64_t seed)
 {
@@ -247,6 +266,10 @@ int main(int argc, char **argv)
     config.request_timeout_ms = 10000;   /* a multi-MB /stateful takes a moment */
     config.backoff_base_ms = 200;
     config.backoff_cap_ms = 2000;
+    config.stats_enabled = true;          /* push /stats every few seconds     */
+    config.stats_interval_s = 3;
+    config.config_report_enabled = true;  /* push /config on its own cadence    */
+    config.config_report_interval_s = 5;
     strncpy(config.version, "5.1.0", sizeof config.version - 1);
     /* SHA-256 of the empty merged.mg the mock starts with: in sync at boot,
      * so the scripted config flip is a real transition. */
@@ -268,6 +291,8 @@ int main(int argc, char **argv)
     callbacks.on_sync_response = on_sync_response;
     callbacks.on_state_change = on_state_change;
     callbacks.on_buffer_level = on_buffer_level;
+    callbacks.collect_stats = collect_stats;
+    callbacks.collect_config = collect_config;
 
     printf("== creating + starting the https_client (target %s:%s) ==\n", argv[1], argv[2]);
     hc_handle *handle = hc_create(&config, &callbacks);
--- a/src/client-agent/https_client/demo/mock_manager.py
+++ b/src/client-agent/https_client/demo/mock_manager.py
@@ -183,6 +183,8 @@ class Handler(BaseHTTPRequestHandler):
             self._handle_stateful(body)
         elif target == "/download":
             self._handle_download(body)
+        elif target in ("/stats", "/config"):
+            self._handle_report(target, body)
         else:
             self._reply(404)
 
@@ -273,6 +275,18 @@ class Handler(BaseHTTPRequestHandler):
         events = body.count(b"\nE ") + (1 if b"\nE " not in body and b"E " in body else 0)
         self._reply(200)  # 200 with an empty body per #37732
         log(f"     /stateless  -> 200  ({len(body)} B, ~{events} events accepted)")
+
+    def _handle_report(self, target, body):
+        # /stats and /config (#37843): the agent pushes a full snapshot the
+        # module stamped with agent_id + cluster; the manager stores it.
+        try:
+            doc = json.loads(body.decode())
+        except (ValueError, UnicodeDecodeError):
+            doc = {}
+        self._reply(200, {})
+        cluster = doc.get("cluster", {})
+        log(f"     {target:<11} -> 200  stored: agent_id={doc.get('agent_id')} "
+            f"cluster={cluster.get('name')}/{cluster.get('node')} keys={sorted(doc.keys())}")
 
     def _handle_stateful(self, body):
         session = self.headers.get("X-Session-Id", "?")
--- a/src/client-agent/https_client/demo/mock_manager.py
+++ b/src/client-agent/https_client/demo/mock_manager.py
@@ -84,13 +84,34 @@ class Handler(BaseHTTPRequestHandler):
 
     # Exact startup-response bytes, kept verbatim: settings_hash is the SHA256
     # of precisely what goes on the wire (#37733 5.1.1).
+    #
+    # Every field the bridge's parser marks required has to be here: it takes
+    # the limits all-or-nothing, and a partial block is silently discarded.
+    # syscheckd then waits on limits that never arrive (fim_initialize()'s
+    # fetch_document_limits_from_agentd() loop) and never finishes starting.
     STARTUP_V1 = json.dumps(
-        {"limits": {"fim": {"file": 100000}, "syscollector": {"packages": 50000},
+        {"limits": {"fim": {"file": 100000, "registry_key": 100000,
+                            "registry_value": 100000},
+                    "syscollector": {"hotfixes": 10000, "packages": 50000,
+                                     "processes": 10000, "ports": 10000,
+                                     "network_iface": 10000, "network_protocol": 10000,
+                                     "network_address": 10000, "hardware": 10000,
+                                     "os_info": 10000, "users": 10000,
+                                     "groups": 10000, "services": 10000,
+                                     "browser_extensions": 10000},
                     "sca": {"checks": 10000}},
          "cluster": {"name": "demo", "node": "node01"},
          "agent": {"groups": ["default"]}}).encode()
     STARTUP_V2 = json.dumps(
-        {"limits": {"fim": {"file": 200000}, "syscollector": {"packages": 50000},
+        {"limits": {"fim": {"file": 200000, "registry_key": 100000,
+                            "registry_value": 100000},
+                    "syscollector": {"hotfixes": 10000, "packages": 50000,
+                                     "processes": 10000, "ports": 10000,
+                                     "network_iface": 10000, "network_protocol": 10000,
+                                     "network_address": 10000, "hardware": 10000,
+                                     "os_info": 10000, "users": 10000,
+                                     "groups": 10000, "services": 10000,
+                                     "browser_extensions": 10000},
                     "sca": {"checks": 20000}},
          "cluster": {"name": "demo", "node": "node01"},
          "agent": {"groups": ["default"]}}).encode()
--- a/src/client-agent/https_client/include/https_client.h
+++ b/src/client-agent/https_client/include/https_client.h
@@ -159,6 +159,15 @@ typedef struct hc_config_t
     /// Local STREAM socket for the stateful sync intake (large sessions bypass
     /// the 64 KB DGRAM event queue). Empty -> the intake is not started.
     char sync_socket_path[HC_MAX_PATH];
+
+    /// Periodic reporters (#37843), both OFF by default. When on, the module
+    /// collects the corresponding snapshot on its interval, stamps agent_id +
+    /// the manager-authoritative cluster, signs and POSTs to /stats resp.
+    /// /config.
+    bool stats_enabled;
+    uint32_t stats_interval_s;         ///< 0 -> 60.
+    bool config_report_enabled;
+    uint32_t config_report_interval_s; ///< 0 -> 3600.
 } hc_config_t;
 
 /**
@@ -242,6 +251,17 @@ typedef struct hc_callbacks_t
                              size_t body_len, void* user_data);
     void (*on_state_change)(int state, void* user_data);  ///< hc_conn_state_t
     void (*on_buffer_level)(int level, void* user_data);  ///< hc_buffer_level_t
+    /// Periodic collectors for the /stats and /config reporters (#37843).
+    /// Unlike every other callback these run on the module's REPORTER thread
+    /// (not the dispatcher): they must return promptly and must NOT call hc_*
+    /// functions. Return either NULL (skip this cycle) or a NUL-terminated
+    /// JSON OBJECT allocated with malloc() from the module's C runtime; the
+    /// module takes ownership and frees it. Before sending, the module stamps
+    /// "agent_id" and the manager-authoritative "cluster" {"name","node"} into
+    /// the object (overwriting same-named members). A non-object return is
+    /// logged and the cycle skipped.
+    char* (*collect_stats)(void* user_data);
+    char* (*collect_config)(void* user_data);
     void* user_data;
 } hc_callbacks_t;
 
--- a/src/client-agent/https_client/src/collectorSource.hpp
+++ b/src/client-agent/https_client/src/collectorSource.hpp
@@ -1,0 +1,79 @@
+/*
+ * Wazuh agent HTTPS client (C++ transport module)
+ * Copyright (C) 2015, Wazuh Inc.
+ * July 21, 2026.
+ *
+ * This program is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU General Public
+ * License (version 2) as published by the FSF - Free Software
+ * Foundation.
+ */
+
+#ifndef _HC_COLLECTOR_SOURCE_HPP
+#define _HC_COLLECTOR_SOURCE_HPP
+
+#include "https_client.h"
+
+#include <cstdlib>
+#include <optional>
+#include <string>
+
+/**
+ * @brief Where the reporter pulls the /stats and /config snapshots. Injected
+ *        so the reporter can be unit-tested without C callbacks.
+ */
+class ICollectorSource
+{
+    public:
+        virtual ~ICollectorSource() = default;
+        virtual std::optional<std::string> collectStats() = 0;
+        virtual std::optional<std::string> collectConfig() = 0;
+};
+
+/**
+ * @brief Production source: calls the C collect_* callbacks directly (NOT via
+ *        the dispatcher — they return a value synchronously) and adopts the
+ *        malloc'd string they return, freeing it after copying.
+ */
+class CallbackCollectorSource final : public ICollectorSource
+{
+    public:
+        explicit CallbackCollectorSource(const hc_callbacks_t& callbacks)
+            : m_callbacks(callbacks)
+        {
+        }
+
+        std::optional<std::string> collectStats() override
+        {
+            return invoke(m_callbacks.collect_stats);
+        }
+
+        std::optional<std::string> collectConfig() override
+        {
+            return invoke(m_callbacks.collect_config);
+        }
+
+    private:
+        std::optional<std::string> invoke(char* (*collector)(void*)) const
+        {
+            if (collector == nullptr)
+            {
+                return std::nullopt;
+            }
+
+            char* raw = collector(m_callbacks.user_data);
+
+            if (raw == nullptr)
+            {
+                return std::nullopt; // The collector chose to skip this cycle.
+            }
+
+            std::string document {raw};
+            std::free(raw); // Module owns the buffer; free after copying.
+            return document;
+        }
+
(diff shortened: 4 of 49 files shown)


jr0me:
## Evidence: what the server actually receives

Captured from a real agent, not a harness. All five daemons (`wazuh-agentd`, `wazuh-syscheckd`, `wazuh-logcollector`, `wazuh-modulesd`, `wazuh-execd`) built from this branch and installed over a container's `/var/ossec`, enrolled and running normally, talking TLS 1.3 with `verification_mode full` to the demo mock in `https_client/demo/`. The mock verifies the AES-CMAC of every request with the `openssl` CLI, so a bad signature would have come back 401 rather than being waved through. 27 requests captured.

Cadence for the run: `<stats_report>` 10s, `<config_report>` 15s.

### 1. The agent asks each daemon once

The agent's modules live in separate processes, so a full picture costs one query per daemon. That is the floor. The same information previously took 15 section-by-section queries across the same five sockets.

Replaying the aggregator's own queries against the live sockets:

```
$ send 'getallconfig'  ->  queue/sockets/agent
  <- ok   920 bytes, 1 module entry
       module='agent'          sections=['buffer', 'client', 'internal', 'package_uninstallation']

$ send 'getallconfig'  ->  queue/sockets/syscheck
  <- ok   2214 bytes, 1 module entry
       module='fim'            sections=['internal', 'rootcheck', 'syscheck']

$ send 'getallconfig'  ->  queue/sockets/logcollector
  <- ok   1399 bytes, 1 module entry
       module='logcollector'   sections=['internal', 'localfile']

$ send 'getallconfig'  ->  queue/sockets/wmodules
  <- ok   1266 bytes, 6 module entries
       module='agent-upgrade'  sections=['ca_store', 'ca_verification', 'enabled']
       module='agent-info'     sections=['integrity_interval', 'interval', 'synchronization', 'task_registry']
       module='syscollector'   sections=['browser_extensions', 'disabled', 'groups', 'hardware', ...]
       module='sca'            sections=['enabled', 'interval', 'max_eps', 'policies', ...]
       module='wazuh_control'  sections=['enabled']
       module='wmodules'       sections=['internal_options']

$ send 'getallconfig'  ->  queue/sockets/com
  <- ok   172 bytes, 1 module entry
       module='execd'          sections=['active-response', 'internal', 'logging']

$ send 'getallstats'   ->  queue/sockets/agent
  <- ok   257 bytes, 1 module entry
       module='agent'          sections=['data', 'error']

$ send 'getallstats'   ->  queue/sockets/logcollector
  <- ok   1120 bytes, 1 module entry
       module='logcollector'   sections=['global', 'interval']
```

Note on that last one: with `logcollector.state_interval` positive (60 s by default) the module only publishes statistics after its state thread's first dump, so `w_logcollector_state_get()` returns nothing until then and logcollector is legitimately absent from early `/stats` pushes. Measured at ~50 s here. Setting the interval to `0` generates the state on demand instead, available immediately. Expect the gap on every agent restart.

`modulesd` is the only daemon that answers with more than one module. Each wodle dumps itself as a one-key object, so the key names the module and the value is its body, which keeps `syscollector` and `sca` individually addressable instead of buried in one `wmodules` blob.

### 2. Assembled into one document

Ten module entries from five daemons, concatenated into a single array, with `agent_id` and the manager-authoritative `cluster` stamped on top.

| module | from | sections it carries |
|---|---|---|
| `agent` | agentd | client, buffer, internal, package_uninstallation |
| `fim` | syscheckd | syscheck, rootcheck, internal |
| `logcollector` | logcollector | localfile, internal |
| `agent-upgrade` | modulesd | enabled, ca_verification, ca_store |
| `agent-info` | modulesd | interval, integrity_interval, synchronization, task_registry |
| `syscollector` | modulesd | disabled, interval, max_eps, hardware, os, network, packages, ports, processes, services, users, groups, browser_extensions, … |
| `sca` | modulesd | enabled, interval, max_eps, policies, scan_on_start, synchronization |
| `wazuh_control` | modulesd | enabled |
| `wmodules` | modulesd | internal_options |
| `execd` | execd | active-response, logging, internal |

### 3. Signed, then POSTed

Both reproduced exactly as the mock received them. `/stats` fits whole; `/config` is 6023 bytes so its body is trimmed at the marked points.

**POST /stats, complete, 304 bytes**

```http
POST /stats HTTP/1.1
Host: 127.0.0.1:27860
Accept: */*
protocol-version: 1
Authorization: Wazuh 001:1785449704:b7952cf585f6cf5e04418ff0fecd3181
Content-Length: 304
Content-Type: application/x-www-form-urlencoded

{
  "agent_id": "001",
  "cluster": { "name": "demo", "node": "node01" },
  "modules": [
    {
      "module": "agent",
      "stats": {
        "data": {
          "buffer_enabled": true,
          "last_ack": "",
          "last_keepalive": "",
          "msg_buffer": 0,
          "msg_count": 0,
          "msg_sent": 0,
          "status": "connected",
          "task_discarded_duplicate": 0,
          "task_dispatched": 0,
          "task_failed": 0
        },
        "error": 0
      }
    }
  ]
}
```

Once logcollector has published, the same endpoint carries both modules (1 439 bytes, its section trimmed):

```json
{
  "agent_id": "001",
  "cluster": { "name": "demo", "node": "node01" },
  "modules": [
    { "module": "agent", "stats": { /* as above */ } },
    {
      "module": "logcollector",
      "stats": {
        "global": {
          "start": "2026-07-31 18:06:19",
          "end":   "2026-07-31 18:19:19",
          "files": [
            { "location": "last -n 20",
              "events": 1, "bytes": 210,
              "targets": [ { "name": "agent", "drops": 0 } ] },
            { "location": "/var/log/dpkg.log",
              "events": 0, "bytes": 0,
              "targets": [ { "name": "agent", "drops": 0 } ] },
            { "location": "journald", ... 5 files total ... }
          ]
        },
        "interval": { /* same shape, counters for the last interval only */ }
      }
    }
  ]
}
```

**POST /config, headers verbatim, body trimmed**

```http
POST /config HTTP/1.1
Host: 127.0.0.1:27860
Accept: */*
protocol-version: 1
Authorization: Wazuh 001:1785449749:cf1a5f758c0473af21ff0f496508836d
Content-Length: 6023
Content-Type: application/x-www-form-urlencoded

{
  "agent_id": "001",
  "cluster": { "name": "demo", "node": "node01" },
  "modules": [
    {
      "module": "agent",
      "config": {
        "client": {
          "notify_time": 10,
          "auto_restart": "no",
          "remote_conf": "yes",
          "manager": [ { "address": "127.0.0.1", "port": 27860, ... } ],
          "stats_report":  { "enabled": "yes", "interval": 10 },
          "config_report": { "enabled": "yes", "interval": 15 },
          "enrollment": { "enabled": "yes", "port": 1515, ... }
        },
        "buffer":   { "disabled": "no", "queue_size": 5000, "events_per_second": 600 },
        "internal": { "agent": { "debug": 0, "warn_level": 90, "tolerance": 15, ... } },
        "package_uninstallation": { ... }
      }
    },
    {
      "module": "fim",
      "config": {
        "syscheck":  { "disabled": "no", "frequency": 43200, "directories": [ ... ],
                       "file_limit": { ... }, "diff": { ... }, "max_eps": ... },
        "rootcheck": { "disabled": "no", "frequency": 43200, "base_directory": "/", ... },
        "internal":  { "syscheck": { ... }, "rootcheck": { ... } }
      }
    },

    /* logcollector, agent-upgrade, agent-info, syscollector, sca,
       wazuh_control, wmodules trimmed */

    {
      "module": "execd",
      "config": {
        "active-response": { "disabled": "no" },
        "logging": { "plain": "yes", "json": "no" },
        "internal": { "execd": { "request_timeout": 0, "max_restart_lock": 0 } }
      }
    }
  ]
}
```

### The conversation over time

```
[+   2101 ms] POST /control    <- id=001 mac=2997f77d6a.. auth OK (37 B)   body='{"type":"startup","version":"v5.0.0"}'
[+   2101 ms]      /control    -> 200  STARTUP accepted (v1 settings)
[+   2145 ms] POST /stats      <- id=001 mac=b7952cf585.. auth OK (304 B)
[+   2146 ms]      /stats      -> 200  stored: agent_id=001 cluster=demo/node01
[+   2191 ms] POST /config     <- id=001 mac=8499fb2bfd.. auth OK (1155 B)
[+   2192 ms]      /config     -> 200  stored: agent_id=001 cluster=demo/node01
[+  12157 ms] POST /control    <- id=001 mac=6408698636.. auth OK (17 B)   body='{"type":"notify"}'
[+  12157 ms]      /control    -> 200  NOTIFY #1 cfg=e3b0c442.. set=914f25f5.. tasks=[]
[+  12253 ms] POST /stats      <- id=001 mac=dddae817d1.. auth OK (323 B)
[+  17201 ms] POST /config     <- id=001 mac=e9b9483705.. auth OK (6023 B)
[+  17202 ms]      /config     -> 200  stored: agent_id=001 cluster=demo/node01
[+  22205 ms] POST /control    <- id=001 mac=77f73a6822.. auth OK (17 B)
[+  22205 ms]      /control    -> 200  NOTIFY #2 tasks=['018f9a-0003', '018f9a-0001', ...]
```

The first `/config` is 1155 bytes and the second is 6023. The first cycle fires while `syscheckd` is still starting, so FIM is absent from it, and the next cycle picks it up. Nothing is retried and nothing is lost: each push is a whole snapshot and latest wins.

### Fault injected: a daemon taken away

Nothing failed on its own here. `wazuh-logcollector` was killed deliberately, mid-run, to check the claim that one missing daemon does not cost the manager the whole document. The agent carried on and the next cycle reported nine modules instead of ten, with no other daemon disturbed.

| | modules |
|---|---|
| all five daemons up | 10: agent, fim, **logcollector**, agent-upgrade, agent-info, syscollector, sca, wazuh_control, wmodules, execd |
| after `kill -9 wazuh-logcollector` (injected) | 9: agent, fim, agent-upgrade, agent-info, syscollector, sca, wazuh_control, wmodules, execd |

If nothing answers at all the cycle is skipped rather than pushed, since an empty `{"modules": []}` would overwrite a good document in the index with nothing.

### Two things this turned up

**Content-Type does not match the declared contract.** All 27 requests, `/control` and `/stateless` included, so this is not specific to this PR:

```
Content-Type: application/x-www-form-urlencoded
```

with a JSON body. The client appends only `protocol-version`, `Authorization` and `Expect:`, so libcurl's default for `CURLOPT_POSTFIELDS` goes out untouched.

Nothing breaks today: `remoted_module` does not inspect the header. But #37733 declares `application/json` for the `/control`, `/config` and `/stats` request bodies, so on those three the agent and the spec disagree. `/stateless` is deliberately `*/*` in #37732, and `/stateful` and `/download` are not JSON at all, so this needs setting per endpoint rather than globally. The AES-CMAC covers method, target, agent id, timestamp and body only, so adding the header cannot affect signing.

**The demo mock has never fully started FIM.** Its startup `limits` block sent one field per module where `bridge_parse_limits` requires all of them (`fim` needs `file`, `registry_key`, `registry_value`; `syscollector` needs 13). It takes the block all-or-nothing, so it was discarded every time, and `syscheckd` sat in `fim_initialize()`'s `fetch_document_limits_from_agentd()` loop, never reaching the `directories_lock` init. FIM was silently missing from every document the demo produced. Fixed in 4ce0f3f048; FIM's reply went from 441 to 2214 bytes with all three sections.

### Still open for the manager side

- **Shape.** #38023 sketches a flat map (`{"client": {…}, "syscheck": {…}}`). This is an array of tagged entries, because a flat map cannot express two modules living in one daemon and `modulesd` hosts six. #37843 makes defining the JSON an agent-side deliverable, so treat this as the proposal rather than a decision.
- **`internal` sections are included.** They bulk the document up, but `GET /agents/{id}/config` can be asked for `internal` today and #38023 wants that endpoint served from the index.
- **Size ceiling.** Each daemon's reply crosses its local socket, capped at `OS_MAXSTR` (64 KiB). A logcollector with a very large state exceeds that; today it is logged and the module left out rather than truncated.
- **`/stats` is thin.** Only agentd and logcollector have a `getstate` producer; both do report, logcollector after its first interval dump. Every other module (fim, syscollector, sca) reports configuration but has no statistics to give, and needs a producer before it can.

Nicogp (approved):

PR #37934, opened 2026-07-23
Prevent empty string version from being stored for agents

### LINKED ISSUE #37324: Investigate why agents have empty string version in database
## Description

While fixing a sorting bug (https://github.com/wazuh/internal-devel-requests/issues/5317), we discovered that some agents in production databases have **empty string** (`''`) for the version field. This should not happen for agents that have sent keep-alive messages, as the version should be extracted from keep-alives.

### Current Situation

Found **~12 agents with empty string versions** in a production database with 50,623+ agents:

```sql
SELECT COUNT(*)
FROM agent
WHERE id > 0 AND version = '';
-- Result: ~12 agents (empty string)
```

### Sample Affected Agents

```
Agent 2500708 | version=''
Agent 2500969 | version=''
Agent 2502224 | version=''
Agent 2504468 | version=''
Agent 2504947 | version=''
Agent 2505797 | version=''
...
```

### Key Observations

1. **ID Range**: All affected agents are in the range 2,500,000 - 2,510,000
2. **Pattern**: Version is empty string (`''`), not NULL
3. **Group**: All have `group` field containing `workstation`
4. **Expected Behavior**:
   - Agents that never connected → `version = NULL`
   - Agents that sent keep-alive → `version = "Wazuh vX.Y.Z"`
   - Agents with `version = ''` → **Should not exist**

### 1. Connection Status
- [ ] What is the `connection_status` of these agents?
  ```sql
  SELECT id, name, connection_status, last_keepalive
  FROM agent
  WHERE id > 0 AND version = '';
  ```
- [ ] Did these agents ever connect and send keep-alive?

### 2. Keep-Alive History
- [ ] Do these agents have `last_keepalive > 0` (meaning they sent at least one keep-alive)?
- [ ] If yes, why wasn't the version extracted from the keep-alive?
- [ ] Are there errors in logs when processing keep-alive for these agents?

### 3. Version Extraction Logic
- [ ] Where in the code is version extracted from keep-alive messages?
- [ ] Could there be a bug where version is set to empty string instead of staying NULL?
- [ ] Are there specific keep-alive formats that fail to extract version?

### 4. Database Operations
- [ ] Could an UPDATE statement be setting version to `''` instead of leaving it NULL?
- [ ] Are there any agent management operations that clear the version field?
- [ ] Could database migrations have introduced this issue?

## Expected Behavior

According to Wazuh architecture:
1. Agent is registered → `version = NULL`
2. Agent never connects → `version` stays `NULL`
3. Agent connects and sends keep-alive → `version` extracted and stored as `"Wazuh vX.Y.Z"`

**There should be NO agents with `version = ''` (empty string).**

### Current Impact
- API sorting by version fails when these agents are included (fixed in PR https://github.com/wazuh/wazuh/pull/37323)
- Data quality issue: empty string vs NULL should be semantically different
- Potential reporting/dashboard inconsistencies

### Root Cause Possibilities
1. **Bug in keep-alive processing**: Version extraction fails and sets `''` instead of keeping NULL
2. **Database update bug**: Some operation incorrectly sets version to empty string
3. **Agent version missing**: Old/broken agent clients don't send version in keep-alive
4. **Migration issue**: Database migration from older Wazuh version set `''` for some agents

## Tasks

- [ ] Query database to check `connection_status` and `last_keepalive` for affected agents
- [ ] Determine if agents with `version = ''` ever sent keep-alive messages
- [ ] Search codebase for where version field is set to empty string
- [ ] Review keep-alive processing code for version extraction logic
- [ ] Check if specific agent versions/platforms are affected
- [ ] Document the root cause
- [ ] Propose fix to prevent `version = ''` from occurring

Antoniogm03:
## Description

Agents in production databases were found with `version = ''` instead of `NULL`, which breaks API sorting by version. The keep-alive parser stored an empty string when the version or config sum was missing around the `" / "` separator, instead of leaving the field unset, and the database write path did not guard against that empty value either.

Closes #37324

## Proposed Changes

- `parse_agent_update_msg` (`src/shared/remoted_op.c`) now only duplicates `version`/`config_sum` into `agent_data` when the corresponding substring is non-empty, leaving the field `NULL` instead of `""` otherwise.
- `wdb_global_update_agent_version` (`src/wazuh_db/wdb_global.c`) binds `NULL` instead of an empty string to the `version` column when the incoming value is empty, as a second safeguard at the database write path.
- `getuname` (`src/shared/file_op.c`) had its static `muname` buffer increased from 512 to 2048 bytes to avoid truncating the OS info string built from these same fields.

### Results and Evidence

For testing purposes, I set up a manager and an agent on Vagrant and modified the contents of the agent's `/etc/os-release` file so that the buffer would fill up.

```
root@agent:/home/vagrant/git/empty_string/wazuh# printf 'VERSION="%s"\n' "$(python3 -c "print('bookworm extended LTS build with backports enabled for the finance compliance hardened stack, kernel patch level 21, ' + 'x'*40)")" >> /etc/os-release
```

#### Before changes

- agent logs

```
2026/07/23 14:55:15 wazuh-agentd: INFO: Version detected -> Linux |agent |6.8.0-86-generic |#87-Ubuntu SMP PREEMPT_DYNAMIC Mon Sep 22 18:03:36 UTC 2025 |x86_64 [Ubuntu Enterprise Custom Build xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx|ubuntu: 24.04.3 LTS extended build finance compliance xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx] - 
```

- Manager global.db

```bash
$ sqlite3 -header -column /var/ossec/queue/db/global.db \
    "SELECT id, name, quote(version), version IS NULL AS is_null, version = '' AS is_empty, \
            last_keepalive, os_uname, config_sum FROM agent WHERE id=1;"

id  name   quote(version)  is_null  is_empty  last_keepalive  os_uname                                                                                             config_sum
--  -----  --------------  -------  --------  --------------  ---------------------------------------------------------------------------------------------------  --------------------------------
1   agent  ''              0        1         1784818526      Linux |agent |6.8.0-86-generic |#87-Ubuntu SMP PREEMPT_DYNAMIC Mon Sep 22 18:03:36 UTC 2025 |x86_64  ab73af41699f13fdd81903b5f23d8d00
```

* after

- agent logs

```
2026/07/23 15:12:42 wazuh-agentd: INFO: Version detected -> Linux |agent |6.8.0-86-generic |#87-Ubuntu SMP PREEMPT_DYNAMIC Mon Sep 22 18:03:36 UTC 2025 |x86_64 [Ubuntu Enterprise Custom Build xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx|ubuntu: 24.04.3 LTS extended build finance compliance xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx] - Wazuh v4.14.8
```
- Manager global.db

```bash
$ sqlite3 -header -column /var/ossec/queue/db/global.db \
    "SELECT id, name, quote(version), version IS NULL AS is_null, version = '' AS is_empty, \
            last_keepalive, os_uname, config_sum FROM agent WHERE id=1;"

id  name   quote(version)   is_null  is_empty  last_keepalive  os_uname                                                                                             config_sum
--  -----  ---------------  -------  --------  --------------  ---------------------------------------------------------------------------------------------------  --------------------------------
1   agent  'Wazuh v4.14.8'  0        0         1784819592      Linux |agent |6.8.0-86-generic |#87-Ubuntu SMP PREEMPT_DYNAMIC Mon Sep 22 18:03:36 UTC 2025 |x86_64  ab73af41699f13fdd81903b5f23d8d00
```

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/src/shared/file_op.c
+++ b/src/shared/file_op.c
@@ -1164,12 +1164,12 @@ int mkstemp_ex(char *tmp_path)
 const char *getuname()
 {
     struct utsname uts_buf;
-    static char muname[512] = "";
+    static char muname[2048] = "";
     os_info *read_version;
 
     if (!muname[0]){
         if (read_version = get_unix_version(), read_version){
-            snprintf(muname, 512, "%s |%s |%s |%s |%s [%s|%s: %s] - %s %s",
+            snprintf(muname, 2048, "%s |%s |%s |%s |%s [%s|%s: %s] - %s %s",
                     read_version->sysname,
                     read_version->nodename,
                     read_version->release,
@@ -1183,7 +1183,7 @@ const char *getuname()
             free_osinfo(read_version);
         }
         else if (uname(&uts_buf) >= 0) {
-            snprintf(muname, 512, "%s %s %s %s %s - %s %s",
+            snprintf(muname, 2048, "%s %s %s %s %s - %s %s",
                      uts_buf.sysname,
                      uts_buf.nodename,
                      uts_buf.release,
@@ -1191,7 +1191,7 @@ const char *getuname()
                      uts_buf.machine,
                      __ossec_name, __ossec_version);
         } else {
-            snprintf(muname, 512, "No system info available - %s %s",
+            snprintf(muname, 2048, "No system info available - %s %s",
                      __ossec_name, __ossec_version);
         }
     }
--- a/src/shared/remoted_op.c
+++ b/src/shared/remoted_op.c
@@ -224,8 +224,12 @@ int parse_agent_update_msg (char *msg,
                 if (str_tmp = strstr(line, " / "), str_tmp) {
                     *str_tmp = '\0';
                     str_tmp += 3;
-                    os_strdup(line, agent_data->version);
-                    os_strdup(str_tmp, agent_data->config_sum);
+                    if (*line) {
+                        os_strdup(line, agent_data->version);
+                    }
+                    if (*str_tmp) {
+                        os_strdup(str_tmp, agent_data->config_sum);
+                    }
                 }
                 else if (str_tmp = strstr(line, __ossec_name), str_tmp) {
                     // If for some reason the separator between Wazuh version and config sum is
--- a/src/wazuh_db/wdb_global.c
+++ b/src/wazuh_db/wdb_global.c
@@ -211,9 +211,16 @@ int wdb_global_update_agent_version(wdb_t *wdb,
         merror("DB(%s) sqlite3_bind_text(): %s", wdb->id, sqlite3_errmsg(wdb->db));
         return OS_INVALID;
     }
-    if (sqlite3_bind_text(stmt, index++, version, -1, NULL) != SQLITE_OK) {
-        merror("DB(%s) sqlite3_bind_text(): %s", wdb->id, sqlite3_errmsg(wdb->db));
-        return OS_INVALID;
+    if (version && version[0]) {
+        if (sqlite3_bind_text(stmt, index++, version, -1, NULL) != SQLITE_OK) {
+            merror("DB(%s) sqlite3_bind_text(): %s", wdb->id, sqlite3_errmsg(wdb->db));
+            return OS_INVALID;
+        }
+    } else {
+        if (sqlite3_bind_null(stmt, index++) != SQLITE_OK) {
+            merror("DB(%s) sqlite3_bind_null(): %s", wdb->id, sqlite3_errmsg(wdb->db));
+            return OS_INVALID;
+        }
     }
     if (sqlite3_bind_text(stmt, index++, config_sum, -1, NULL) != SQLITE_OK) {
         merror("DB(%s) sqlite3_bind_text(): %s", wdb->id, sqlite3_errmsg(wdb->db));


MiguelazoDS (commented):
GJ, as much as I could see the 2048 is safe size. Some questions: 

I was able to reproduce the error but I find strange to have such a long os_version value 

```
root@noble:/home/vagrant/macos# sqlite3 /var/ossec/queue/db/global.db 'select os_uname from agent where version=""' --line
os_uname = Linux |repro-bugged |6.11.0-26-generic |#26~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Apr 17 19:20:47 UTC 2 |x86_64
root@noble:/home/vagrant/macos#
```

- Could we find what was the specific field in the environment causing this? 
- Shouldn't we need to add some integration tests in addition to the manual testing performed. Probably using the environment use case, or something similar to validate the implementation?

MiguelazoDS (approved):

PR #37949, opened 2026-07-24
Agent /stateless events: batching, occupancy ladder, adaptive 413

jr0me:
Part of #37835.

Reopened after the stack reorder: stateless now bases on the integration branch so it can merge next (ahead of tasks/stateful/stats). Supersedes #37855, which GitHub auto-closed when the base swap briefly made its commits a subset of its old base.

This slice implements the HTTPS client module side of `/stateless` delivery:

- H/E framing with continuation escaping and bounded, tail-preserving batching.
- Size, age, and forced flushes; reaching the size threshold wakes the sender immediately.
- A drop-newest buffer cap with ordered occupancy-level callbacks and drop accounting.
- Adaptive 413 handling that splits multi-event batches without loss and drops only an unsplittable event rejected on its own.
- Request-size budgeting that includes the H metadata line.
- Thread-safe shared lifecycle, buffer-level, adaptive-payload, and jitter-RNG state.

It also carries the review fixes inherited through its integration base (drain bound, single-shot lifecycle, download size cap, and Windows symlink handling).

Scope: this PR is the module-side stateless transport slice, not the complete agentd integration for #37835. EventForward routing, production buffer-level handling, legacy-buffer retirement, and agent configuration wiring remain follow-up work.

Validation for the latest review update: all nine modified production/test translation units compile cleanly against the existing build configuration. The local test suites were not rerun.

### DIFF
--- a/.github/actions/check_files/deb_linux_agent_amd64.csv
+++ b/.github/actions/check_files/deb_linux_agent_amd64.csv
@@ -10,7 +10,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12374168,0.1
 /var/ossec/lib/libdbsync.so,root,wazuh,750,file,-rwxr-x---,505104,0.1
 /var/ossec/lib/libagent_sync_protocol.so,root,wazuh,750,file,-rwxr-x---,218488,0.1
-/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,105592,0.1
+/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,270552,0.1
 /var/ossec/lib/libschema_validator.so,root,wazuh,750,file,-rwxr-x---,390232,0.1
 /var/ossec/lib/libsysinfo.so,root,wazuh,750,file,-rwxr-x---,2603392,0.1
 /var/ossec/lib/libfimdb.so,root,wazuh,750,file,-rwxr-x---,253344,0.1
--- a/.github/actions/check_files/deb_linux_agent_i386.csv
+++ b/.github/actions/check_files/deb_linux_agent_i386.csv
@@ -10,7 +10,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12358288,0.1
 /var/ossec/lib/libdbsync.so,root,wazuh,750,file,-rwxr-x---,525380,0.1
 /var/ossec/lib/libagent_sync_protocol.so,root,wazuh,750,file,-rwxr-x---,216020,0.1
-/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,105592,0.1
+/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,270552,0.1
 /var/ossec/lib/libschema_validator.so,root,wazuh,750,file,-rwxr-x---,426324,0.1
 /var/ossec/lib/libsysinfo.so,root,wazuh,750,file,-rwxr-x---,2792476,0.1
 /var/ossec/lib/libfimdb.so,root,wazuh,750,file,-rwxr-x---,273712,0.1
--- a/.github/actions/check_files/rpm_linux_agent_amd64.csv
+++ b/.github/actions/check_files/rpm_linux_agent_amd64.csv
@@ -74,7 +74,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12417720,0.1
 /var/ossec/lib/libdbsync.so,root,wazuh,750,file,-rwxr-x---,484376,0.1
 /var/ossec/lib/libagent_sync_protocol.so,root,wazuh,750,file,-rwxr-x---,220736,0.1
-/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,105592,0.1
+/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,270552,0.1
 /var/ossec/lib/libschema_validator.so,root,wazuh,750,file,-rwxr-x---,386024,0.1
 /var/ossec/lib/libsysinfo.so,root,wazuh,750,file,-rwxr-x---,2555328,0.1
 /var/ossec/lib/libfimdb.so,root,wazuh,750,file,-rwxr-x---,224672,0.1
--- a/.github/actions/check_files/rpm_linux_agent_i386.csv
+++ b/.github/actions/check_files/rpm_linux_agent_i386.csv
@@ -74,7 +74,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12362096,0.1
 /var/ossec/lib/libdbsync.so,root,wazuh,750,file,-rwxr-x---,541828,0.1
 /var/ossec/lib/libagent_sync_protocol.so,root,wazuh,750,file,-rwxr-x---,216520,0.1
-/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,105592,0.1
+/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,270552,0.1
 /var/ossec/lib/libschema_validator.so,root,wazuh,750,file,-rwxr-x---,398136,0.1
 /var/ossec/lib/libsysinfo.so,root,wazuh,750,file,-rwxr-x---,2717844,0.1
 /var/ossec/lib/libfimdb.so,root,wazuh,750,file,-rwxr-x---,259032,0.1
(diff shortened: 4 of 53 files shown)


jr0me:
Review follow-up after checking the current pushed draft stack:

Resolved in this PR by `0f46f6ff04`:

- Serialized buffer-level transitions and adaptive-payload access so intake and sender threads cannot race or publish occupancy changes out of order.
- Serialized the shared jitter RNG used by concurrent stream workers.
- Guarded submit/notify lifecycle checks against concurrent stop.
- Made the size threshold wake the stateless sender immediately and used the delay returned by `tick()`.
- Counted the H metadata line inside `batch_size_bytes`, so normal splittable requests respect the documented full-request limit.
- Added real accounting for overflow, oversized-event, and permanent-response drops.
- Renamed the stateless H/E encoder honestly and removed the unused accumulator member.
- Added focused coverage for size wakeups, request budgeting, drop counts, concurrent RNG calls, and submit-vs-stop.

I also updated the PR description: this PR is the HTTPS client module-side `/stateless` slice and does not, by itself, close all of #37835.

Reassessment against the later drafts:

- #37854 adds `/control` task dispatch; it does not resolve any remaining stateless integration item.
- #37856 adds the independent `/stateful` worker and STREAM intake, so stateful/stateless transport separation belongs there. Its description still leaves the producer swap as follow-up, and the explicit held-open-stateful/low-latency-stateless regression is not present yet.
- #37857 adds periodic `/stats` and `/config`; it does not resolve the stateless integration items.
- The descriptions of #37854 and #37856 still name pre-reorder bases, although their actual GitHub base refs form the current stack correctly.

Still pending in later agentd integration work: route EventForward into `hc_submit_event`, connect `on_buffer_level` to the production agent state/logging behavior, retire the legacy buffer path, and map the agent configuration into the module. None of the currently pushed draft PRs completes those items, so #37835 should remain open after this PR.

Validation: all nine modified translation units compile cleanly against the existing build configuration. Per request, I did not run the local test suites.

jr0me:
### Before merge: align or drop `HTTPS_CLIENT_TSAN`

The `HTTPS_CLIENT_TSAN` CMake option doesn't match how sanitizers are handled elsewhere in the tree and should be sorted out before this merges.

Context:
- The standard knob is `FSANITIZE` (`src/CMakeLists.txt:50-52`), which sets `-fsanitize=address,leak,undefined`. The module and its tests already get it under the RTR build (`TEST=1 DEBUG=1 FSANITIZE=1`). ASan and TSan can't coexist, so that knob can't provide TSan.
- The only in-tree precedent for a per-component TSan option is the engine: `ENGINE_ENABLE_TSAN` in `src/engine/CMakeLists.txt:27-28`. No other agent or shared module defines one.
- As written, the option diverges from that precedent:
  - named `HTTPS_CLIENT_TSAN` rather than `HTTPS_CLIENT_ENABLE_TSAN`
  - no ASan/TSan mutual-exclusion guard, so `FSANITIZE=1` together with `-DHTTPS_CLIENT_TSAN=ON` passes both `-fsanitize=address` and `-fsanitize=thread` to the compiler
  - no `-fno-omit-frame-pointer` for readable backtraces
  - coverage (`--coverage`, PUBLIC on the lib) leaks into the TSan build, so gcov counters report false-positive races and the run halts before validating anything

Fix, pick one:
1. Drop the option and rely on `FSANITIZE` like the other agent modules. Simplest, no extra surface. ASan does not detect data races, so concurrency confidence comes from review plus the ASan runs.
2. Keep it but mirror `ENGINE_ENABLE_TSAN`: rename to `HTTPS_CLIENT_ENABLE_TSAN`, add the mutual-exclusion `FATAL_ERROR`, add `-fno-omit-frame-pointer`, and skip coverage when it's on.

For the record, with the coverage guard applied locally the unit suite runs clean under TSan (200/200, 0 races), so this is about the option's shape, not an actual race in the code.

jr0me:
Head is now `f39991abed`, six commits on top of the reviewed `4b0da23ac3`. The downstream stack is rebased onto it: #37854 `eb59a10a68`, #37856 `74ec41d19c`, #37857 `ff0909846d`. All four mergeable.

Going through the findings.

### 1. Max `/stateless` payload not configurable (High). Confirmed, fixed in `534e2d8f20`

Root cause was slightly different from the description: `<batch>` did reach the parser, but `client-config.c` deliberately dropped it with *"handled by the events module; ignored by the client parser"*, deferring to this issue. So the deferral pointed here.

`Read_Client_Batch()` now sits alongside `Read_Client_SSL()` and parses `<size>` through `w_validate_bytes` and `<interval>` through `w_parse_time` into a new `agent_batch` member, and `bridge_build_config()` passes both into `hc_config_t`. Absent still leaves them at zero, which the module already reads as "apply your own default", so an ossec.conf without the block behaves exactly as before.

The syntax was already pinned by `test_batch_block_is_accepted_and_ignored` (`1MB` / `10s`). That test is replaced by six: parsing, absent defaults, suffix-less bytes, zero size rejected, an over-long interval rejected, unknown tag rejected.

`buffer_cap_multiplier` has no configuration surface defined anywhere in the spec, so it keeps its 4x default rather than inventing an element for it.

### 2. Occupancy ladder does not match legacy semantics (Medium). Confirmed, fixed in `9c4baf33d5`

That commit landed before the review arrived, so the cited lines no longer exist.

The reading of `buffer.c` is right, including that FULL means literally at capacity and FLOOD is a sustained-full condition rather than a byte percentage. `levelForLocked()` and the 50/75/90 thresholds are gone. `BufferLevelLadder` (`src/bufferLevel.hpp`) ports the real machine: WARNING at `agent.warn_level`, back to NORMAL at `agent.normal_level` (the hysteresis), FULL on an actual drop-newest rejection, FLOOD only once FULL has held for `agent.tolerance` seconds. The step down from FULL or FLOOD to WARNING stays silent, which is also what `buffer.c` does.

`EventAccumulator` now reports only `occupancyPercent()`. The state machine needs a clock for the dwell, so it lives with the stream rather than the buffer. The bridge reads the same internal options `buffer_init()` reads, so a deployment that tuned `warn_level` keeps its thresholds.

### 3. `on_buffer_level` not wired to state-event emission (Medium). Confirmed, fixed in the same commit

`bridge_on_buffer_level()` emits the legacy log lines (`WARN_BUFFER`, `FULL_BUFFER`, `FLOODED_BUFFER`, `NORMAL_BUFFER`) and calls `send_buffer_status_event()`, which is no longer `STATIC` and is declared in `agentd.h`.

It writes straight to `send_msg()`, bypassing the accumulator the same way `buffer.c` bypassed the ring it reported on: a flood report must not queue behind the flood it is reporting. That keeps it on the legacy socket for now, and it moves with every other stateless producer when the TCP path is retired.

### 4. `buildHeaderLine()` hand-builds JSON (Medium). Confirmed, fixed in `fb842d4d30`

One thing to add: it was on the intake path, not only the flush path. `eventBytesBudgetLocked()` needs the line's length and `submit()` calls that for every event, so the string was being rebuilt per event rather than per batch.

The line is now built once in the constructor through `nlohmann::json` and cached for the life of the stream. New test drives an agent id of `0"01` and asserts the body opens with `H {"wazuh":{"agent":{"id":"0\"01"}}}`.

### 5. No unit tests for the C bridge entry point (Low-medium). Confirmed, fixed in `f39991abed`

Five cases in `test_https_client_bridge.c`: the forward to the module, the drop-newest return value, the NULL and zero-length guard, submit before start, and submit after stop (the stopping-flag check that keeps a late intake frame away from a destroyed handle). Needed `-Wl,--wrap,hc_submit_event`. 22 tests to 27.

### 6. Formatting nit (Low). Not confirmed, leaving as is

That spacing is what astyle 3.1 produces, not a deviation from it. Feeding the spaced version through the container's astyle with `ci/input/astyle.config`:

```
in:  SubmissionResult {accepted, accepted && !wasFlushDue && isFlushDue};
out: SubmissionResult {accepted, accepted&& !wasFlushDue&& isFlushDue};
```

astyle mangles `&&` inside a brace-init list. RTR runs it with `--dry-run` and fails on any `Formatted` line, so adding the spaces would break the astyle gate rather than satisfy it. The dry run is clean on the current head.

## Two other fixes in the same push

`https_client_bridge.c` used `atomic_load`/`atomic_store` on an `_Atomic bool`. mingw's `<stdatomic.h>` does not declare those generic macros, so on winagent they compiled as implicit declarations and left undefined references in the object. Nothing on Windows referenced that object until `local_start()` called into it, and at that point `wazuh-agent.exe` stopped linking. The flag is already covered by the same mutex as the handle it guards, so it is a plain `bool` under that lock now (`4a9abdcac7`).

`local_start()` started the client after the syscheck, logcollector and module threads. On Windows those call `SendMSG` in process rather than writing to a queue, so every event emitted in that window was dropped rather than queued. It now starts next to `buffer_init()`, which the file already initialises first for the same reason (`48d0776792`).

`9c4baf33d5` also carries two stream behaviours found while wiring the ladder. `tick()` always returned the batch interval, and the size threshold only wakes the sender as the buffer crosses it, so a backlog that stayed above the mark drained one payload per interval regardless of arrival rate. And the shutdown drain walked the full retry ladder against a waiter that is deliberately never stopped, which could hold process exit for minutes against a slow or back-pressuring manager, on Windows from an `atexit` handler. The drain is now bounded like `ControlStream::drainStep`: the drain window, a single attempt.

## Verification

Per head rather than only the top:

| head | unit | component |
|---|---|---|
| this branch `f39991abed` | 218 | 17 |
| #37854 `eb59a10a68` | 239 | 17 |
| #37856 `74ec41d19c` | 257 | 24 |
| #37857 `ff0909846d` | 265 | 26 |

On the top head: `test_https_client_bridge` 27, `test_buffer` 10, `test_agentd` 15, `test_client-config_https` 18. astyle 3.1 and cppcheck 2.20 clean. winagent compiles `https_client_bridge.c` and `client-config.c` with no warnings and no undefined atomic symbols.

## Still open on #37835

Retiring `os_setwait` for data traffic. Producers still park in `os_wait()` on the legacy TCP connection state even though stateless events no longer travel over it.

There is also no unit test for the `EventForward` routing branch itself. None existed before this PR either, but that branch is now what decides where every stateless event in the agent goes.

Nicogp (approved):

PR #37977, opened 2026-07-24
Fix WM_MAX_WAITunit comment modified (milliseconds -> microseconds)

### LINKED ISSUE #37757: Re-evaluate the retry/backoff of `modulesSync()`
## Summary

The retry loop in `modulesSync()` (`src/wazuh_modules/src/wmodules.c`) has a latent unit bug in its backoff, and fixing it as-is would change manager-to-agent communication behavior on the agent. This issue splits that work out of PR #37689 (a log-level fix for #37554) so the timing change can be analyzed on its own.

## Background

`modulesSync()` retries a module `sync()` up to `WM_MAX_ATTEMPTS` (3) times, sleeping between attempts:

```c
if (retry > 0) {
    usleep(retry * WM_MAX_WAIT);
    mdebug1("WModules is not ready. Retry %d", retry);
}
```

`WM_MAX_WAIT` is documented as milliseconds:

```c
#define WM_MAX_WAIT 500   // Max. wait between attempts in milliseconds.
```

but its value is passed straight to `usleep()`, which takes microseconds. So the loop sleeps 500 us / 1 ms / 1.5 ms between attempts instead of the intended 500 ms / 1000 ms / 1500 ms. All three retries fire within about 3 ms, so a module that needs a moment to become ready gets no real grace period.

`modulesSync()` is the only consumer of `WM_MAX_WAIT`, so nothing else depends on the current (wrong) timing.

## Why this is not a trivial unit fix

Correcting the unit turns a negligible ~3 ms total wait into a real ~3 s worst-case wait (0.5 s + 1 s + 1.5 s) on the "sync keeps failing while the agent is running" path. On Windows that runs inline in the receiver thread, not in a dedicated thread:

`receiver.c` (manager-message receive loop) -> `wmcom_send()` -> *(WIN32)* `wmcom_dispatch()` -> `wmcom_sync()` -> `modulesSync()`, then `continue` to read the next message.

So a single deterministically failing sync would block reception of all manager-to-agent traffic (ACKs, keepalives, other `*_sync` messages, config pushes) for about 3 s, making the agent look unresponsive and possibly aggravating existing socket back-pressure. 

> Note: correcting the unit also exposes a portability problem. `usleep()` is only specified by POSIX for values below 1000000, and the corrected backoff reaches 1500000 us, so retries 2 and 3 would pass an out-of-range argument and may fail with `EINVAL` instead of sleeping. Any fix must use a call without the one-second limit (for example the `Sleep()` / `select()` idiom already used by `wm_sendmsg` in the same file).

## Questions to answer before changing the timing

1. Should a deterministic `sync()` failure be retried at all, or does retrying only help for a transient "module not ready yet" condition?
2. What is a sane upper bound on the total wait, given the receiver-thread coupling on Windows?
3. What is the real impact on the receiver thread on Windows (blocked ACKs/keepalives, socket back-pressure) at the corrected timing?
4. Should the backoff run off the receiver thread entirely, or should late/failing syncs be dropped instead of retried?

## Proposed changes

- Correct the ms-vs-us unit in the `modulesSync()` backoff.
- Use a sleep primitive without the `usleep()` one-second limit (portable `Sleep()` / `select()` idiom).
- Decide and implement the retry policy per the questions above, with an explicit upper bound on total wait.
- Add/adjust unit tests for the chosen timing and retry policy.

## Evidence

Backoff, before and after correcting only the unit (for reference; the final policy is what this issue decides):

| Retry | Intended | Current (`usleep(retry * WM_MAX_WAIT)`) | Unit-corrected |
|---|---|---|---|
| 1 | 500 ms | 500 us | 500 ms |
| 2 | 1000 ms | 1 ms | 1000 ms |
| 3 | 1500 ms | 1.5 ms | 1500 ms |
| total | 3 s | ~3 ms | ~3 s |

hernanvalenzuela:
## Description

The tempting fix looks like: convert WM_MAX_WAIT to milliseconds properly (e.g. usleep(retry * WM_MAX_WAIT * 1000)), so the backoff actually waits 500ms/1s/1.5s as originally documented. We deliberately did not do that, because it introduces a real regression risk on Windows:

modulesSync() is reached via wmcom_send(), which behaves very differently per platform:
Linux/macOS: wmcom_send() just opens a Unix domain socket to wazuh-modulesd (a separate process) and returns immediately. The actual retry loop runs on a dedicated thread inside that other process, completely decoupled from the agent's manager-communication path.
Windows: there is no separate wazuh-modulesd process — everything runs in one process, in threads. wmcom_send() calls wmcom_dispatch() → modulesSync() directly, inline, on the same thread that reads incoming messages from the manager (client-agent/receiver.c).
Correcting the unit conversion would make that thread block for up to ~3 seconds (worst case: 3 failed attempts × real ms/s sleeps) every time a sync message hits this retry path on Windows. During that window the agent would stop reading ACKs, keepalives, and config updates from the manager — i.e. it could appear unresponsive/disconnected.
Additionally, usleep() has a POSIX limit of 1,000,000 microseconds; a correctly-converted third attempt (3 * 500ms = 1500ms = 1,500,000µs) would exceed it, requiring a different sleep primitive anyway.
Even with correct timing, the retry loop itself doesn't distinguish why a sync attempt failed (module not yet initialized vs. module not configured at all vs. a genuinely invalid payload) — only the first case can plausibly benefit from waiting, so a longer sleep wouldn't reliably fix the race it's meant to address, while still carrying the Windows blocking risk.

  - [ ] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

### Artifacts Affected
N-A

### Configuration Changes
N-A

### Tests Introduced
N-A

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

### DIFF
--- a/src/wazuh_modules/include/wmodules.h
+++ b/src/wazuh_modules/include/wmodules.h
@@ -23,7 +23,7 @@
 #define WM_BUFFER_MAX   1024                        // Max. static buffer size.
 #define WM_BUFFER_MIN   1024                        // Starting JSON buffer length.
 #define WM_MAX_ATTEMPTS 3                           // Max. number of attempts.
-#define WM_MAX_WAIT     500                           // Max. wait between attempts in milliseconds.
+#define WM_MAX_WAIT     500                         // Max. wait between attempts in microseconds.
 #define WM_IO_WRITE     0
 #define WM_IO_READ      1
 #define WM_ERROR_TIMEOUT 1                          // Error code for timeout.


cborla (approved):
LGTM

cborla (approved):
LGTM

PR #37980, opened 2026-07-24
Fix manager status detection with hidepid=invisible

### LINKED ISSUE #37979: Manager API reports running daemons as failed with hidepid=invisible
## Description

The Wazuh Manager API reports required daemons as `failed` when `/proc` is mounted with `hidepid=invisible`, even though the daemons are running.

`get_manager_status()` currently checks daemon existence through `os.path.exists('/proc/<pid>')`. For a daemon owned by another user, `hidepid=invisible` makes that path appear nonexistent. The API therefore returns error 1017 (`Some Wazuh daemons are not ready yet`) and authentication/installation cannot complete.

## Reproduction

1. Mount procfs with:

```text
proc /proc proc defaults,nosuid,nodev,noexec,hidepid=invisible,gid=proc 0 0
```

2. Start Wazuh Manager.
3. Confirm `/var/ossec/bin/wazuh-control status` reports the daemons running.
4. Authenticate against `POST /security/user/authenticate?raw=true`.

The API returns HTTP 500 and reports required daemons such as `wazuh-modulesd`, `wazuh-analysisd`, `wazuh-execd`, `wazuh-db`, and `wazuh-remoted` as failed.

## Proposed fix

Use `os.kill(pid, 0)` to test process existence:

- success: process exists;
- `PermissionError` / `EPERM`: process exists but cannot be signaled;
- `ProcessLookupError` / `ESRCH`: process does not exist.

This preserves the existing existence-only behavior without requiring visibility of `/proc/<pid>`.

## Validation

Validated on Wazuh 4.14.6 on a hardened Ubuntu host:

- `hidepid=invisible` remained enabled;
- all required daemons reported `running`;
- API authentication succeeded;
- indexer cluster status remained `green`;
- behavior persisted after reboot.

The current default branch contains the same check and is also affected.

Reported and validated by **Castello Branco Tecnologia** (`CastelloBrancoTecnologia`). Contact: cesar@castellobranco.tec.br.

CastelloBrancoTecnologia:
## Description

Fixes #37979.

`get_manager_status()` treated an inaccessible `/proc/<pid>` entry as a dead process. On systems using `hidepid=invisible`, the manager API user cannot see process directories belonging to other daemon users, so healthy required daemons were reported as `failed` and the API returned error 1017.

This change checks process existence with signal `0` instead:

- a successful call means the process exists;
- `PermissionError` means it exists but cannot be signaled;
- `ProcessLookupError` means it no longer exists.

## Tests

- Added unit coverage for accessible, permission-restricted and missing PIDs.
- `py_compile` passed for the changed implementation and test files.
- Functional validation on Wazuh 4.14.6 with `/proc` mounted using `hidepid=invisible`:
  - all required manager daemons reported `running`;
  - API authentication succeeded;
  - indexer cluster stayed `green`;
  - result persisted across reboot.

The full local framework test environment could not be built under Python 3.14 because the pinned `uvloop==0.19.0` does not compile against Python 3.14; the added tests are intended to run in the project's supported CI environment.

Submitted and validated by **Castello Branco Tecnologia** (`CastelloBrancoTecnologia`). Contact: cesar@castellobranco.tec.br.

This bug prevents Wazuh from working correctly on systems hardened according to the security recommendations provided by Lynis.

I can confirm that, after applying fix https://github.com/wazuh/wazuh/pull/37980, I am successfully running a Wazuh instance on an Ubuntu host hardened following the Lynis security recommendations (lynis score 98).


cborla:
In `test_get_manager_status` (final block asserting running), the mock no longer matches the code path: it patches os.path.exists, but that branch now calls process_exists(), which uses os.kill. Since the test pid comes from ossec-0.pid, it ends up calling the real os.kill(0, 0), which always succeeds (signal to its own process group) — so the assertion passes regardless of what process_exists actually does. The hidepid=invisible (PermissionError) case isn't exercised here.
  
Could you mock os.kill (or process_exists) in this block instead of os.path.exists?

  Also, since the PR description mentions the full framework test environment couldn't be built locally (Python 3.14 / uvloop), could you share the CI output (or a local pytest -v run) for test_get_manager_status and test_process_exists? That would confirm the new tests actually pass as expected.

CastelloBrancoTecnologia:
Thanks for catching this, @cborla — you were right. I updated test_get_manager_status so it no longer relies on os.path.exists for PID liveness:

- os.kill(0, 0) raising ProcessLookupError now explicitly exercises the failed path.
- os.kill(0, 0) raising PermissionError explicitly exercises the hidepid=invisible path and asserts running.
- Both blocks verify one os.kill(0, 0) call per manager process.

The fix is in commit 23986b7233.

I also ran the two requested tests locally with pytest -v:

    platform linux -- Python 3.14.4, pytest-9.0.2
    collected 4 items

    wazuh/core/cluster/tests/test_utils.py::test_get_manager_status PASSED
    wazuh/core/cluster/tests/test_utils.py::test_process_exists[None-True] PASSED
    wazuh/core/cluster/tests/test_utils.py::test_process_exists[PermissionError-True] PASSED
    wazuh/core/cluster/tests/test_utils.py::test_process_exists[ProcessLookupError-False] PASSED

    4 passed, 10 warnings in 2.38s

Command used:

    PYTHONPATH=../api:. ../.venv-pr37980/bin/python -m pytest wazuh/core/cluster/tests/test_utils.py::test_get_manager_status wazuh/core/cluster/tests/test_utils.py::test_process_exists -v

For transparency, this was an isolated WSL environment using uvloop 0.22.1; the repository-pinned uvloop 0.19.0 still does not build on Python 3.14.

cesarchefinho:
Any news ?

cesarchefinho (approved):
Sucefuly tested on production server

PR #38019, opened 2026-07-27
fix default macOS group_description

### LINKED ISSUE #38013: dbsync: false "modified" verdict for rows with NULL TEXT fields
## Summary

The `dbsync_groups` collector reports 100% of macOS groups as `modified` on every scan, even when nothing changed. 163/163 groups are reported as changed every hour — 978 events → 1,956 findings (48.9% of all findings in an example deployment), carrying MITRE `T1098`/`T1531` against an idle host.

## Root cause

**Update:** further investigation traced this past the `dbsync` comparison/storage layer originally suspected below (kept for context) to a collector-specific bug. See PR #38019.

In `data_provider/src/sysInfoMac.cpp::SysInfo::getGroups()`:

```cpp
groupItem["group_description"] = group["comment"];
```

This reads a `"comment"` key that the macOS groups provider never sets. `GroupsProvider::collect()` (`data_provider/src/extended_sources/groups/src/groups_darwin.cpp::genGroupJson()`) only ever populates `groupname`, `gid` and `gid_signed` — there is no code path that writes a `"comment"` field for any group, on any call. Because `nlohmann::json::operator[]` silently creates a missing key as `null` when accessed on a non-const object, `group["comment"]` evaluates to JSON `null` unconditionally, for every group, on every scan.

Every sibling field in the same function (`group_name`, `group_uuid`, `group_users`) already guards against exactly this by falling back to `UNKNOWN_VALUE` (a single space — the project-wide sentinel for "field not available on this platform", `data_provider/src/sharedDefs.h`) instead of leaving the field unset/`null`. `group_description` was the one field that didn't follow that convention, so it was the one field whose collected value was inconsistent from the perspective of the sync/diff logic that decides whether a row changed — the same `null` value every scan should have compared equal to itself, but reaching the comparison as a raw, unguarded `null` (rather than the stable `UNKNOWN_VALUE` string every other unavailable field uses) is what made it register as a difference.

<details>
<summary>Original (superseded) hypothesis: dbsync NULL/empty-string round-trip</summary>

In `shared_modules/dbsync/src/sqlite/sqlite_dbengine.cpp::bindJsonData` (lines 925-931), when a collector sends `null` for a TEXT field, the bind stores an empty string `""` in SQLite — never a real NULL:

```cpp
std::string value { jsData.is_string() ? jsData.get_ref<const std::string&>() : "" };
stmt->bind(cid, value);
```

On the next scan, `getRowDiff` compares the freshly collected `null` against the value read back from storage (`""`), and `nlohmann::json` treats `null != ""` as a real difference.

This turned out not to be the actionable root cause for this issue: SQLite in this codebase is deliberately never relied on to round-trip real NULLs — see `UNKNOWN_VALUE`/`EMPTY_VALUE` used throughout `data_provider` precisely because of this constraint. The dbsync-level behavior described above is real, but the field that was actually reaching it as an unguarded `null` (`group_description`) was a collector-side bug, not a case dbsync needs to support.
</details>

## Steps to reproduce

1. Enroll an agent with the groups collector enabled.
2. Leave the host untouched (no group created/modified/deleted).
3. Let at least 2 notifying inventory scans complete (the first scan after a `wazuh-modulesd` start is silent by design).
4. Observe: every group is reported as `modified` with an empty `changed_fields`, on every subsequent scan.

## Expected behavior

A scan that observes no real change to a group must not emit a `modified` event.

## Impact

- False assertion of security-relevant activity: MITRE `T1098` (Account Manipulation) / `T1531` (Account Access Removal) fired against hosts where nothing changed.
- Dominates finding volume from idle hosts (in the reference deployment, ~49% of all findings).
- macOS-specific: confined to the `dbsync_groups` collector on this platform (see #38014 for the analogous symptom on Windows browser extensions, root-caused separately).

## Suggested fix

Implemented in #38019: make `group_description` follow the same `contains()` + non-empty check + `UNKNOWN_VALUE` fallback pattern already used by `group_name` and the other optional group fields, instead of reading the never-populated `"comment"` key directly.

## How to verify

- Enroll a stock macOS agent, leave it untouched through at least 3 inventory scans.
- Before the fix: every group event carries `type: modified` with `changed_fields: []`, repeating every scan.
- After the fix: zero `modified` events with empty `changed_fields` on the idle host; `SELECT version, COUNT(*) FROM dbsync_groups GROUP BY version;` in the agent's local `db/local.db` shows a single stable bucket that does not move across scans (matching `dbsync_users` on the same host, which was never affected).

cborla:
## Description

Closes #38013.

The macOS groups collector (`dbsync_groups`) reports all groups as `modified` on every inventory scan, even when nothing changed — 978 events / 1,956 findings in the reference deployment from #38013, carrying MITRE `T1098`/`T1531` against an idle host.

The issue's original write-up suspected the root cause was in the shared `dbsync` comparison/storage layer (`shared_modules/dbsync`). Investigation traced it further back: the actual root cause is in the macOS-specific collector itself, in `SysInfo::getGroups()` (`data_provider/src/sysInfoMac.cpp`).

`groupItem["group_description"] = group["comment"];` reads a `"comment"` key that `GroupsProvider::collect()` (`data_provider/src/extended_sources/groups/src/groups_darwin.cpp`) never sets — `genGroupJson()` there only ever populates `groupname`, `gid` and `gid_signed`. Because `nlohmann::json::operator[]` silently creates a missing key as `null` when accessed on a non-const object, `group["comment"]` evaluates to `null` unconditionally, for every group, on every scan.

Every sibling field in the same function (`group_name`, `group_uuid`, `group_users`) already guards against this by falling back to `UNKNOWN_VALUE` (a single space, the project-wide sentinel for "field not available on this platform" — see `data_provider/src/sharedDefs.h`) instead of leaving the field unset/`null`. `group_description` was the one field that didn't follow that convention, so it was the one field whose collected value was inconsistent from the perspective of the sync/diff logic that decides whether a row changed.

## Proposed Changes

- `data_provider/src/sysInfoMac.cpp`: `group_description` now follows the same `contains()` + non-empty check + `UNKNOWN_VALUE` fallback pattern already used by `group_name` and the other optional group fields, instead of reading the never-populated `"comment"` key directly.

This is a one-line, collector-only fix — it does not touch `shared_modules/dbsync` (FIM, agent_info, SCA and syscollector all consume that shared code, so keeping this change scoped to the macOS groups collector avoids any blast radius beyond the actual bug).

### Results and Evidence

Not able to build/run on macOS in this sandbox (Linux-only environment; `sysInfoMac.cpp` is excluded from the build outside `CMAKE_SYSTEM_NAME STREQUAL "Darwin"`). The root cause was confirmed by static trace instead:
- `groups_darwin.cpp::genGroupJson()` / `GroupsProvider::collect()`: no code path sets a `"comment"` key on the returned JSON.
- `sysInfoMac.cpp:496` (pre-fix) therefore always read a missing key, which is `null` under `nlohmann::json::operator[]`.

Needs manual verification on real macOS hardware before merge (see checklist below).

- Enroll a stock macOS agent, leave it untouched through at least 3 inventory scans (first scan after a `wazuh-modulesd` start is silently absorbed by design).
- Before the fix: every group event carries `type: modified` with `changed_fields: []`, repeating every scan.
- After the fix: zero `modified` events with empty `changed_fields` on the idle host; `SELECT version, COUNT(*) FROM dbsync_groups GROUP BY version;` in the agent's local `db/local.db` should show a single stable bucket that does not move across scans (matching `dbsync_users` on the same host, which was never affected).

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - N/A — agent-side data_provider change, no ruleset involved.

  - N/A — no engine/decoder changes.

- Wazuh server API/Framework
  - N/A — no server/API changes.

### Artifacts Affected

- `wazuh-agent` executable/package on macOS only (`data_provider` is statically linked into the agent). No effect on Linux/Windows agents or on the manager.

### Configuration Changes

N/A — no new configuration parameters, no default-value changes.

### Tests Introduced

None added in this PR. `sysInfoMac.cpp::getGroups()` has no existing unit test harness (`GroupsProvider`/`UserGroupsProvider` are constructed directly, not injected, so isolating this function would need a small refactor for testability beyond the scope of this one-line fix). Existing coverage: `data_provider/src/extended_sources/groups/tests/test_groups_darwin.cpp` already covers `GroupsProvider::collect()` itself and is unaffected by this change, since the bug was in how its output is consumed, not in the provider.

## Review Checklist

- [x] Code changes reviewed
- [x] Configuration changes documented (N/A)
- [x] Verified on real macOS hardware

### DIFF
--- a/src/data_provider/src/sysInfoMac.cpp
+++ b/src/data_provider/src/sysInfoMac.cpp
@@ -493,7 +493,7 @@ nlohmann::json SysInfo::getGroups() const
 
         groupItem["group_id"] = group["gid"];
         groupItem["group_name"] = (group.contains("groupname") && !group["groupname"].get<std::string>().empty()) ? group["groupname"] : UNKNOWN_VALUE;
-        groupItem["group_description"] = group["comment"];
+        groupItem["group_description"] = (group.contains("comment") && !group["comment"].get<std::string>().empty()) ? group["comment"] : UNKNOWN_VALUE;
         groupItem["group_id_signed"] = group["gid_signed"];
         groupItem["group_uuid"] = UNKNOWN_VALUE;
         groupItem["group_is_hidden"] = group["is_hidden"];


anromerom (approved):
**LGTM**

- [x] Code changes reviewed
- [x] Verified on real macOS hardware

PR #38045, opened 2026-07-28
Stop the deferred sync-flag update from inserting partial rows

### LINKED ISSUE #37993: wazuh-syscheckd: FIM errors found on agents during 5.x performance testing
## Description
During the 5.x vs 4.x performance/observability test, `wazuh-syscheckd` produced two new error patterns on agent nodes, both absent in 4.x.

### 1. `sqlite: NOT NULL constraint failed: file_entry.checksum`
Hits 98 of 100 agents, 11,105 occurrences total, always as a single burst per node (e.g. 134 identical lines in the same second). Preceded by an inotify real-time queue overflow warning, followed by the daemon reporting the DB status back to normal — it self-recovers but likely drops FIM entries for files in flight during the burst. First occurrence lands consistently ~110-120 min after agent start across all 98 nodes, pointing to a schedule-driven trigger (periodic DB sync/scan) rather than a random race.
```
12:54:53  wazuh-syscheckd: WARNING: Real-time inotify kernel queue is full.
12:54:53  wazuh-syscheckd: ERROR: sqlite: NOT NULL constraint failed: file_entry.checksum [x134, same second]
12:54:53  wazuh-syscheckd: INFO: (6036): The file database status returns to normal.
```
Repro hint: sustained file-change load long enough to overflow the inotify queue, crossing a scheduled DB sync/scan boundary.

### 2. `Invalid handle value.` on shutdown, with corrupted log output
Hits 7 of 100 agents (-05, -23, -25, -34, -49, -61, -93), 449 lines total, always right after SIGTERM in the exit-cleanup path. On 5 of the 7 nodes the burst is followed by log lines with raw non-printable/binary bytes instead of a message, consistent with memory corruption (e.g. use-after-free) during cleanup. Timing tightly clustered: 41-47 min since agent start on all 7.
```
11:45:13  wazuh-syscheckd: INFO: (1225): SIGNAL [(15)-(Terminated)] Received. Exit Cleaning...
11:45:14  wazuh-syscheckd: ERROR: Invalid handle value. [x9]
11:45:14  wazuh-syscheckd: ERROR: YM-eR$ [non-UTF8 bytes]
```
Repro hint: repeated agent stop/restart cycles under active FIM load

## Reference
Full report: https://github.com/wazuh/wazuh-qa-automation/issues/8278#issuecomment-5069815390

Miguevrgo:
## Description

Closes #37993

During the 5.x vs 4.x performance run, `wazuh-syscheckd` logged 11,105 occurrences of the following error across **98 of 100 agent nodes**, always as a dense burst at a single timestamp, and absent in 4.x:

```sql
12:54:53  wazuh-syscheckd: ERROR: sqlite: NOT NULL constraint failed: file_entry.checksum   [x134, same second]
```

The burst comes from the deferred sync-flag update. While a scheduled scan walks the filesystem, every entry it `INSERT`s is queued for a later sync-flag write (`pending_sync_updates`, `src/syscheckd/src/file/file.c:319-330`). Those writes are only flushed **after** the scan transaction closes:

```c
fim_db_transaction_deleted_rows(db_transaction_handle, transaction_callback, &txn_ctx);  // file.c:1337, txn closes
w_mutex_unlock(&syscheck.fim_scan_mutex);                                                 // file.c:1338
cleanup_failed_fim_files(failed_paths);                                                   // file.c:1341, deletes rows
process_pending_sync_updates(FIMDB_FILE_TABLE_NAME, pending_sync_updates);                // file.c:1345, the flush
```

`fim_db_set_sync_flag()` (`src/syscheckd/src/db/src/file.cpp:483`) then builds a row carrying only the primary key, the sync flag and the version:

```c
rowData["path"]    = encodedPath;
rowData["sync"]    = sync_value;
rowData["version"] = version;
FIMDB::instance().updateItem(updateData, ...);   // -> DBSync::syncRow
```

`syncRow` is an **upsert**, not an update. In `SQLiteDBEngine::syncTableRowData` a missing row is indistinguishable from "no changes": `getRowDiff` returns `diffExist = SQLITE_ROW == stmt->step()` (`sqlite_dbengine.cpp:1469`), so a row that no longer exists takes the `insertElement` branch and the partial row is inserted. It carries no `checksum`, the column is `TEXT NOT NULL` (`fimDB.hpp:38`), and SQLite rejects it. The exception is caught at `file.cpp:606-609` and logged verbatim, which is the error above.

Two consequences, one of them silent:

- one `ERROR` line per queued path whose entry disappeared during the scan, hence the same-second burst;
- the intended `sync = 1` is lost. That entry stays at `sync = 0` and is therefore excluded from the synchronization set (`WHERE sync = 1`, `db.cpp:219` and `db.cpp:609`) until a later scan touches it again.

This is new in 5.x: `fim_db_set_sync_flag` does not exist in `4.14.8`, and neither does the `sync`/`version` column model it writes.

> [!NOTE]
> The issue also reports a second, unrelated failure mode, `Invalid handle value.` plus corrupted log output on shutdown (7 of 100 agents). That one is a shutdown-ordering problem, `exit(1)` running while the scan thread and the dbsync pipeline threads are still live, and it is **not** addressed here. It touches shutdown ordering and the transaction callback threading model, so it deserves its own pull request.

## Proposed Changes

Make the sync-flag write update-only, so an absent row is a no-op rather than an insert.

- **`src/shared_modules/dbsync/src/sqlite/sqlite_dbengine.cpp`.** New `options.update_only` flag on the sync-row path. When set, the `insertElement` branch is skipped. The check sits inside the same `ctx->m_syncMutex` critical section as `getRowDiff`, so there is no window between deciding the row is missing and acting on it. Default behavior is unchanged for every existing caller.
- **`src/syscheckd/src/db/src/file.cpp`.** `fim_db_set_sync_flag()` sets `update_only`, since its row is a partial row by construction.

The rest of the flush path is untouched: an entry that is still present is updated exactly as before.

### Results and Evidence

Validated on Ubuntu 24.04 with an agent built from this branch, `v5.0.0 rc1`.

**Before, on the code without the fix.** Running the new test against `13240906da` with only the test file overlaid reproduces the production error string exactly:

```sql
[ RUN      ] DBSyncTest.syncRowUpdateOnlyDoesNotInsertMissingRow
DB error, id: 19. sqlite: NOT NULL constraint failed: file_entry.checksum
DB error, id: 19. sqlite: NOT NULL constraint failed: file_entry.checksum
/wazuh/src/shared_modules/dbsync/tests/interface/dbsync_test.cpp:850: Failure
  0
  dbsync_sync_row(handle, jsPartialRowUpdateOnly.get(), callbackData)
    Which is: 19
[  FAILED  ] DBSyncTest.syncRowUpdateOnlyDoesNotInsertMissingRow (1 ms)
 1 FAILED TEST
```

**After, with the fix.** The remaining `DB error` line is the test's deliberate negative control, the same partial sync issued *without* `update_only`, which must still fail:

```sql
[ RUN      ] DBSyncTest.syncRowUpdateOnlyDoesNotInsertMissingRow
DB error, id: 19. sqlite: NOT NULL constraint failed: file_entry.checksum
[       OK ] DBSyncTest.syncRowUpdateOnlyDoesNotInsertMissingRow (1 ms)
[  PASSED  ] 1 test.
```

**No regressions in the dbsync interface suite:**

```sql
[==========] 83 tests from 3 test suites ran. (476 ms total)
[  PASSED  ] 83 tests.
```

**Live agent with the patched build.** The normal path must keep working, since `update_only` now governs every deferred sync-flag write. A scan flushing 34,536 deferred updates, all applied:

```sql
2026/07/28 08:36:10 wazuh-syscheckd[40901] fim_scan.c:61 at fim_scan(): INFO: (6008): File integrity monitoring scan started.
2026/07/28 08:37:37 wazuh-syscheckd[40901] syscheck.c:147 at process_pending_sync_updates(): DEBUG: Processed 34536 pending sync flag updates
2026/07/28 08:37:37 wazuh-syscheckd[40901] fim_scan.c:123 at fim_scan(): INFO: (6009): File integrity monitoring scan ended.
```

```sql
successful sync flag updates : 34536
failed sync flag updates     : 0
checksum NOT NULL errors     : 0
syscheckd ERROR lines        : 0
```

  - [x] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

Linux compilation was verified on the agent target with this branch: zero errors and zero warnings on the two changed translation units. Windows and macOS builds, the memory suites and the log-language review have not been run and are left for review. Engine, decoder/rule and API items are N/A: this change touches neither.

### Artifacts Affected

- `wazuh-syscheckd` (agent and manager).
- `libdbsync` / the dbsync shared module, consumed by every component that uses it.

No package, configuration, on-disk schema or protocol changes. The `file_entry` schema is untouched.

### Configuration Changes

N/A. No new configuration parameters and no changes to default values.

`options.update_only` is an internal dbsync sync-row option, not user-facing. It is opt-in and absent by default, so every existing `syncRow` caller keeps its current upsert behavior.

### Tests Introduced

`DBSyncTest.syncRowUpdateOnlyDoesNotInsertMissingRow` in `src/shared_modules/dbsync/tests/interface/dbsync_test.cpp`.

It builds a table shaped like `file_entry`, with a `NOT NULL` checksum column, and covers both sides of the flag:

- partial sync against a row that **exists**, with the option set: updates the row and emits `MODIFIED`, so the fix does not turn legitimate updates into no-ops;
- partial sync against a row that is **gone**, without the option: still fails, which pins the original defect and keeps the test honest if the flag is ever removed;
- the same sync **with** the option: succeeds as a no-op.

## Review Checklist

- [ ] Code changes reviewed
- [ ] CHANGELOG entry added under Agent / Fixed

The second failure mode from #37993, `Invalid handle value.` on shutdown, is deliberately out of scope here and still needs its own pull request.


Miguevrgo:
## For the test_process_priority change

There was some flaky condition in FIM test causing a check to fail with 4 != 0 which represented the nice value expected for the daemon, this behaviour was caused by a launch of another process with sudo which matched a lower priority

MarcelKemp (approved):

PR #38083, opened 2026-07-29
Implement download endpoint

### LINKED ISSUE #38022: Implement `/download` HTTPS endpoint
### Description

Implement the manager-side `/download` endpoint.

The endpoint streams binary resources to agents using HTTP chunked transfer encoding. Initially, it supports downloading:

* merged configuration (`merged.mg`)
* WPK packages

Large files must be streamed without loading them completely into memory.

---

#### Download merged configuration

**Request**

```json
{
  "resource_type": "config",
  "resource_id": "web-servers"
}
```

**Response**

```
HTTP/1.1 200 OK
Content-Type: application/octet-stream
Transfer-Encoding: chunked

<chunk>
<chunk>
...
```

---

#### Download WPK package

**Request**

```json
{
  "resource_type": "wpk",
  "resource_id": "wazuh_agent_v5.0.1_linux_x86_64.wpk"
}
```

**Response**

```
HTTP/1.1 200 OK
Content-Type: application/octet-stream
Transfer-Encoding: chunked

<chunk>
<chunk>
...
```

---

### Tasks

* Implement `/download`.
* Validate requested resource type.
* Locate requested resource.
* Stream files using HTTP chunked encoding.
* Return appropriate HTTP errors.
* Add logging.
* Add tests.

---

### Definition of Done

* Configuration downloads work.
* WPK downloads work.
* Files are streamed using chunked encoding.
* Memory usage is constant regardless of file size.
* Errors are handled correctly.
* Logging and tests are implemented.
* CI passes.

jepalfer:
## Description

Implements the manager-side `POST /download` endpoint, which streams binary resources to agents
with HTTP chunked transfer encoding: merged centralized configurations (`merged.mg`) and WPK
upgrade packages. Large files are streamed rather than buffered, so remoted's memory footprint is
independent of the file size.

Closes #38022

## Proposed Changes

**Transport — streamed responses (`96fbdd9`)**

`IHttpResponder::send()` takes a `std::string` body, so a large file would be held in memory in
full. Handlers now supply an `IByteSource` and the *transport* drives the transfer, alternating a
read on a worker thread with a write on the connection's I/O strand — one chunk resident, and
a worker slot held only for the duration of a single read rather than for the whole transfer.

The chunk size is `remoted.http_stream_chunk_size` (default 64 KiB). It is charged per *in-flight
transfer*, so the worst case is roughly that value times the number of simultaneous downloads; a
larger chunk buys fewer read/write round trips at the cost of more resident memory while transfers
run. `remoted.http_max_parallel_connections` (previously hardcoded at 512) is now settable for the
same reason: it is the only bound on concurrent streamed responses, since chunked output rearms
`http_write_timeout` per chunk and there is no per-stream limiter, so a fleet-wide upgrade over slow
links is bounded by that value alone.

The mode is declared at registration (`ResponseMode::Streamable`) rather than per response: the
transport creates a request's response builder when the request is dispatched (which is what lets it
release the received body before the request is queued) and a builder's output mode is fixed at
creation. A streamable route's responder therefore keeps the request handle and picks buffered for an
error or chunked for a transfer — so a 404 stays a normal `Content-Length` reply.

A read error drops the builder *without* `done()`, so no terminating chunk is emitted and the peer
sees a truncated body it will retry; emitting the terminator would hand over a short file that looks
complete. Aborted transfers are reported as a throttled WARN, since the volume is client-driven.

**Endpoint (`5982a63`)**

- `resource_id` is what the agent requests and the manager serves exactly that: a single group
  resolves under `etc/shared/<group>/merged.mg`; a comma-separated selector is hashed into
  `var/multigroups/<sha256(selector)[:8]>/merged.mg`, matching how wazuh-db names that directory; a
  WPK filename resolves under `var/upgrade/`. No group lookup, no database access.
- The single-group and WPK forms join agent input into a path, so the resource-id grammars are the
  containment boundary: no separator, not `.` or `..`, no leading dot, and a `.wpk` suffix for
  packages. A multigroup selector is hashed rather than joined, so its directory name is hex by
  construction.
- The file is opened and `fstat`ed *before* any byte of the response is written, so a missing resource
  is a clean 404 rather than a truncated 200. The descriptor is held for the whole transfer, so a
  writer that publishes by rename (`c_group()` with `remoted.disk_storage=1` writes `merged.mg.tmp`
  and renames) cannot corrupt an in-flight download — the old inode keeps being served.
- **A resource modified in place mid-transfer aborts the transfer instead of completing it.** Holding
  the descriptor is not sufficient, because the two default write paths do not rename: `c_group()`
  with `remoted.disk_storage=0` (the default, `src/remoted/src/config.c`) truncates `merged.mg` and
  rewrites the same inode, and a WPK is fetched straight to its final path in `var/upgrade/`
  (`wurl_request()`), so it grows in place while being staged. The descriptor then follows the new
  contents, and the failure is silent: `read()` reaches the *new* end-of-file, returns 0, and the
  transport emits the terminating chunk — the agent accepts a truncated or spliced file as complete.
  `FileByteSource` therefore compares delivered bytes against the length at `open()`, and re-`fstat`s
  before reporting end-of-stream, catching a file that was shortened, that grew, or that was
  rewritten to the same length (mtime). Any of those throws, which the transport already handles as
  an abort: no terminating chunk, so the agent retries. Being cut off is recoverable; being handed
  corrupt bytes that look whole is not — and for `merged.mg` nothing downstream would catch it.
  This *detects* in-place modification rather than preventing it; preventing it means making those
  writers publish atomically by rename, which is a change in `manager.c` and the upgrade module.
- Adds `tools/send_download.py` for manual and end-to-end testing.

### Results and Evidence

<details>
<summary>Single group merged.mg case</summary>

```
root@wazuh-master:/home/vagrant#  sudo /var/wazuh-manager/bin/agent_groups -a -g web-servers -q
Group 'web-servers' created.

root@wazuh-master:/home/vagrant# sudo cat /var/wazuh-manager/etc/shared/web-servers/merged.mg
#web-servers
!76 agent.conf
<agent_config>

  <!-- Shared agent configuration here -->

</agent_config>
!134 web.conf
<agent_config>
  <localfile><location>/var/log/nginx/access.log</location><log_format>syslog</log_format></localfile>
</agent_config>
root@wazuh-master:/home/vagrant# sudo sha256sum /var/wazuh-manager/etc/shared/web-servers/merged.mg
a96e0343bb07807a532d5890d0940fbd93919fde0b25b4dd4e12b3e8f5c2c387  /var/wazuh-manager/etc/shared/web-servers/merged.mg

root@wazuh-agent1:/home/vagrant# KEY=$(awk '{print $4; exit}' /var/ossec/etc/client.keys)
root@wazuh-agent1:/home/vagrant# ID=$(awk  '{print $1; exit}' /var/ossec/etc/client.keys)
root@wazuh-agent1:/home/vagrant# TS=$(date +%s)
root@wazuh-agent1:/home/vagrant# BODY='{"resource_type":"config","resource_id":"web-servers"}'
root@wazuh-agent1:/home/vagrant# MAC=$(printf 'WAZUH-REQUEST\n1\nPOST\n/download\n%s\n%s\n%s' "$ID" "$TS" "$BODY" \
      | openssl mac -macopt cipher:AES-256-CBC -macopt hexkey:"$KEY" CMAC | tr 'A-Z' 'a-z')

root@wazuh-agent1:/home/vagrant# curl -k -X POST "https://192.168.56.11:1517/download"   -H "protocol-version: 1" -H "Authorization: Wazuh $ID:$TS:$MAC"   -H "Content-Type: application/json" --data-binary "$BODY"   -o /tmp/a.mg -D /dev/stderr
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0HTTP/1.1 200 OK
Connection: keep-alive
Server: wazuh-manager-remoted
Date: Mon, 03 Aug 2026 15:48:16 GMT
Content-Type: application/octet-stream
Transfer-Encoding: chunked

100   306    0   252  100    54   5469   1172 --:--:-- --:--:-- --:--:--  6652

root@wazuh-agent1:/home/vagrant# sha256sum /tmp/a.mg 
a96e0343bb07807a532d5890d0940fbd93919fde0b25b4dd4e12b3e8f5c2c387  /tmp/a.mg
```

</details>

<details>
<summary>Multigroup merged.mg case</summary>

```
root@wazuh-master:/home/vagrant#  sudo /var/wazuh-manager/bin/agent_groups -a -g databases -q
Group 'databases' created.

root@wazuh-master:/home/vagrant# sudo /var/wazuh-manager/bin/agent_groups -a -i 001 -g web-servers -q
Group 'web-servers' added to agent '001'.
root@wazuh-master:/home/vagrant# sudo /var/wazuh-manager/bin/agent_groups -a -i 001 -g databases   -q
Group 'databases' added to agent '001'.
root@wazuh-master:/home/vagrant# sudo /var/wazuh-manager/bin/agent_groups -s -i 001 -q
The agent 'wazuh-agent1' with ID '001' belongs to groups: web-servers, databases.

root@wazuh-master:/home/vagrant# cat /var/wazuh-manager/var/multigroups/97d66859/merged.mg 
#97d66859
!240 agent.conf
<!-- Source file: web-servers/agent.conf -->
<agent_config>

  <!-- Shared agent configuration here -->

</agent_config>
<!-- Source file: databases/agent.conf -->
<agent_config>

  <!-- Shared agent configuration here -->

</agent_config>
!134 web.conf
<agent_config>
  <localfile><location>/var/log/nginx/access.log</location><log_format>syslog</log_format></localfile>
</agent_config>

root@wazuh-master:/home/vagrant# sha256sum /var/wazuh-manager/var/multigroups/97d66859/merged.mg 
d7f4ab09262a6ce24b5894a4a93b7dd666b634cc8b14b0cea69841665c6d39dd  /var/wazuh-manager/var/multigroups/97d66859/merged.mg

root@wazuh-agent1:/home/vagrant# KEY=$(awk '{print $4; exit}' /var/ossec/etc/client.keys)
root@wazuh-agent1:/home/vagrant# ID=$(awk  '{print $1; exit}' /var/ossec/etc/client.keys)
root@wazuh-agent1:/home/vagrant# TS=$(date +%s)
root@wazuh-agent1:/home/vagrant# BODY='{"resource_type":"config","resource_id":"web-servers,databases"}'
root@wazuh-agent1:/home/vagrant# MAC=$(printf 'WAZUH-REQUEST\n1\nPOST\n/download\n%s\n%s\n%s' "$ID" "$TS" "$BODY" \
      | openssl mac -macopt cipher:AES-256-CBC -macopt hexkey:"$KEY" CMAC | tr 'A-Z' 'a-z')

root@wazuh-agent1:/home/vagrant# curl -k -X POST "https://192.168.56.11:1517/download"   -H "protocol-version: 1" -H "Authorization: Wazuh $ID:$TS:$MAC"   -H "Content-Type: application/json" --data-binary "$BODY"   -o /tmp/b.mg -D /dev/stderr
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0HTTP/1.1 200 OK
Connection: keep-alive
Server: wazuh-manager-remoted
Date: Mon, 03 Aug 2026 16:04:37 GMT
Content-Type: application/octet-stream
Transfer-Encoding: chunked

100   478    0   414  100    64   8853   1368 --:--:-- --:--:-- --:--:-- 10391
root@wazuh-agent1:/home/vagrant# cat /tmp/b.mg 
#97d66859
!240 agent.conf
<!-- Source file: web-servers/agent.conf -->
<agent_config>

  <!-- Shared agent configuration here -->

</agent_config>
<!-- Source file: databases/agent.conf -->
<agent_config>

  <!-- Shared agent configuration here -->

</agent_config>
!134 web.conf
<agent_config>
  <localfile><location>/var/log/nginx/access.log</location><log_format>syslog</log_format></localfile>
</agent_config>
root@wazuh-agent1:/home/vagrant# sha256sum /tmp/b.mg 
d7f4ab09262a6ce24b5894a4a93b7dd666b634cc8b14b0cea69841665c6d39dd  /tmp/b.mg
```

</details>

<details>
<summary>WPK case</summary>

```
root@wazuh-master:~/wpk# head -c 1073741824 /dev/urandom > payload/big.bin
root@wazuh-master:~/wpk# TMPDIR=/var/tmp python3 /home/vagrant/devel/wazuh/tools/agent-upgrade/wpkpack.py \
        huge-signed.wpk wpk_sign.pem wpk_sign.key payload/upgrade.sh payload/big.bin
root@wazuh-master:~/wpk# ls -la huge-signed.wpk
-rw-r--r-- 1 root root 1074070760 Aug  3 16:17 huge-signed.wpk
root@wazuh-master:~/wpk# sha256sum huge-signed.wpk
4e74b30f67ed2c141f4629de670e47a931a76036c9caa139c41a9a39dab83b58  huge-signed.wpk

root@wazuh-master:~/wpk# cat /tmp/wpkverify.py 
#!/usr/bin/env python3
from cryptography import x509
from cryptography.hazmat.primitives import hashes
from cryptography.hazmat.primitives.asymmetric import padding, utils
import sys

MAGIC, SIGNLEN = b"WPK256\0", 2048 // 8
data = open(sys.argv[1], "rb").read()

assert data[:len(MAGIC)] == MAGIC, "bad magic: not a WPK256 file"
rest = data[len(MAGIC):]
end = rest.index(b"\0")                       # cert is NUL-terminated
cert_pem, rest = rest[:end], rest[end + 1:]
signature, payload = rest[:SIGNLEN], rest[SIGNLEN:]

cert = x509.load_pem_x509_certificate(cert_pem)
print("cert subject :", cert.subject.rfc4514_string())
print("cert issuer  :", cert.issuer.rfc4514_string())
print("payload      :", len(payload), "bytes")

h = hashes.Hash(hashes.SHA256()); h.update(payload)
cert.public_key().verify(signature, h.finalize(), padding.PKCS1v15(),
                         utils.Prehashed(hashes.SHA256()))
print("SIGNATURE    : OK")

if len(sys.argv) > 2:
    ca = x509.load_pem_x509_certificate(open(sys.argv[2], "rb").read())
    ca.public_key().verify(cert.signature, cert.tbs_certificate_bytes,
                           padding.PKCS1v15(), cert.signature_hash_algorithm)
    print("CERT CHAIN   : OK (issued by %s)" % ca.subject.rfc4514_string())


root@wazuh-master:~/wpk# python3 /tmp/wpkverify.py huge-signed.wpk
cert subject : CN=wazuh-wpk-signer
cert issuer  : CN=wazuh-wpk-ca
payload      : 1074069491 bytes
SIGNATURE    : OK

root@wazuh-master:~/wpk# sudo cp huge-signed.wpk /var/wazuh-manager/var/upgrade/
root@wazuh-master:~/wpk# sudo chown wazuh-manager:wazuh-manager /var/wazuh-manager/var/upgrade/huge-signed.wpk

root@wazuh-agent1:/home/vagrant# KEY=$(awk '{print $4; exit}' /var/ossec/etc/client.keys)
root@wazuh-agent1:/home/vagrant# ID=$(awk  '{print $1; exit}' /var/ossec/etc/client.keys)
root@wazuh-agent1:/home/vagrant# TS=$(date +%s)
root@wazuh-agent1:/home/vagrant# BODY='{"resource_type":"wpk","resource_id":"huge-signed.wpk"}'
root@wazuh-agent1:/home/vagrant# MAC=$(printf 'WAZUH-REQUEST\n1\nPOST\n/download\n%s\n%s\n%s' "$ID" "$TS" "$BODY" \
      | openssl mac -macopt cipher:AES-256-CBC -macopt hexkey:"$KEY" CMAC | tr 'A-Z' 'a-z')
root@wazuh-agent1:/home/vagrant# curl -k -X POST "https://192.168.56.11:1517/download"   -H "protocol-version: 1"   -H "Authorization: Wazuh $ID:$TS:$MAC"   -H "Content-Type: application/json"   --data-binary "$BODY"   -o /tmp/got.wpk   -D /dev/stderr   -w 'time=%{time_total}s  speed=%{speed_download} B/s  bytes=%{size_download}\n'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0HTTP/1.1 200 OK
Connection: keep-alive
Server: wazuh-manager-remoted
Date: Mon, 03 Aug 2026 16:23:55 GMT
Content-Type: application/octet-stream
Transfer-Encoding: chunked

100 1024M    0 1024M  100    55   273M     14  0:00:03  0:00:03 --:--:--  273M
time=3.738468s  speed=287302381 B/s  bytes=1074070760

root@wazuh-agent1:/home/vagrant# sha256sum /tmp/got.wpk 
4e74b30f67ed2c141f4629de670e47a931a76036c9caa139c41a9a39dab83b58  /tmp/got.wpk

root@wazuh-agent1:/home/vagrant# python3 /tmp/wpkverify.py /tmp/got.wpk 
cert subject : CN=wazuh-wpk-signer
cert issuer  : CN=wazuh-wpk-ca
payload      : 1074069491 bytes
SIGNATURE    : OK
```

</details>

<details>
<summary>Error handling</summary>

```
root@wazuh-agent1:/home/vagrant# MGR=192.168.56.11:1517
root@wazuh-agent1:/home/vagrant# dl() {                                                
  local KEY ID TS BODY MAC
  KEY=$(awk '{print $4; exit}' /var/ossec/etc/client.keys)
  ID=$(awk  '{print $1; exit}' /var/ossec/etc/client.keys)
  TS=$(date +%s)
  BODY="{\"resource_type\":\"$1\",\"resource_id\":\"$2\"}"; shift 2
  MAC=$(printf 'WAZUH-REQUEST\n1\nPOST\n/download\n%s\n%s\n%s' "$ID" "$TS" "$BODY" \
        | openssl mac -macopt cipher:AES-256-CBC -macopt hexkey:"$KEY" CMAC | tr 'A-Z' 'a-z')
  curl -k -X POST "https://$MGR/download" \
    -H "protocol-version: 1" -H "Authorization: Wazuh $ID:$TS:$MAC" \
    -H "Content-Type: application/json" --data-binary "$BODY" "$@"
}

root@wazuh-agent1:/home/vagrant# dl config no-such-group -i
HTTP/1.1 404 Not Found
Connection: keep-alive
Content-Length: 41
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:44:28 GMT
{"error":"Resource not found","code":404}

root@wazuh-agent1:/home/vagrant# dl wpk nope.wpk -i
HTTP/1.1 404 Not Found
Connection: keep-alive
Content-Length: 41
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:44:06 GMT
{"error":"Resource not found","code":404}

root@wazuh-agent1:/home/vagrant# dl config "../../etc/shadow" -i
HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 50
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:43:40 GMT
{"error":"Invalid resource identifier","code":400}

root@wazuh-agent1:/home/vagrant# dl config "a/b" -i
HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 50
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:42:46 GMT
{"error":"Invalid resource identifier","code":400}

root@wazuh-agent1:/home/vagrant# dl config ".." -i
HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 50
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:42:18 GMT
{"error":"Invalid resource identifier","code":400}

root@wazuh-agent1:/home/vagrant# dl config "" -i
HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 50
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:41:56 GMT
{"error":"Invalid resource identifier","code":400}

root@wazuh-agent1:/home/vagrant# dl config "web,,servers" -i
HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 50
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:41:25 GMT
{"error":"Invalid resource identifier","code":400}

root@wazuh-agent1:/home/vagrant# dl config "web-servers," -i
HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 50
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:33:38 GMT
{"error":"Invalid resource identifier","code":400}
 
root@wazuh-agent1:/home/vagrant# dl wpk package -i
HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 50
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:33:47 GMT
{"error":"Invalid resource identifier","code":400}

HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 44
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:33:57 GMT
{"error":"Invalid resource type","code":400}

root@wazuh-agent1:/home/vagrant# dl CONFIG web-servers -i
HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 44
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:37:00 GMT
{"error":"Invalid resource type","code":400}
```

These are negative tests: the 400/404 **is** the pass condition. Two things are being checked.

*The two failures stay distinguishable.* `404 Resource not found` means "the identifier was well-formed, the file is not there" — an absent group, an unstaged WPK. `400 Invalid resource identifier` means "this could never name anything" — traversal (`../../etc/shadow`), an embedded separator (`a/b`), `..`, empty, a malformed multigroup selector (`web,,servers`, `web-servers,`), a WPK without the `.wpk` suffix. Collapsing them would tell an agent to retry something that can never succeed, and would let a prober use the response to map the manager's filesystem. `400 Invalid resource type` (`CONFIG` in the wrong case, an unknown type) is a third distinct code, so the closed `resource_type` enum is not silently falling back to `config`.

*Rejection happens in the grammar, before any filesystem call.* Every 400 above is refused by `isValidGroupSelector()`/`isValidWpkFilename()`, so no path is ever built from hostile input — containment does not depend on the filesystem behaving. `../../etc/shadow` returning the *same* error as `a/b` is the point: nothing about it is treated as a special case.

All error responses carry `Content-Length` and no `Transfer-Encoding`, which is the buffered half of the chunked-vs-buffered selection below.

</details>

<details>
<summary>Auth rejection</summary>

```
root@wazuh-agent1:/home/vagrant# curl -k -i -X POST "https://$MGR/download" -H "protocol-version: 1" \
  -H "Authorization: Wazuh $ID:$(date +%s):deadbeef" \
  --data-binary '{"resource_type":"config","resource_id":"web-servers"}'
HTTP/1.1 401 Unauthorized
Connection: keep-alive
Content-Length: 52
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:55:44 GMT
{"error":"Invalid client authentication","code":401}

root@wazuh-agent1:/home/vagrant# TS=$(( $(date +%s) - 400 ));
root@wazuh-agent1:/home/vagrant# BODY='{"resource_type":"config","resource_id":"web-servers"}'
root@wazuh-agent1:/home/vagrant#  KEY=$(awk '{print $4; exit}' /var/ossec/etc/client.keys)
root@wazuh-agent1:/home/vagrant# ID=$(awk  '{print $1; exit}' /var/ossec/etc/client.keys)
root@wazuh-agent1:/home/vagrant# MAC=$(printf 'WAZUH-REQUEST\n1\nPOST\n/download\n%s\n%s\n%s' "$ID" "$TS" "$BODY" \
      | openssl mac -macopt cipher:AES-256-CBC -macopt hexkey:"$KEY" CMAC | tr 'A-Z' 'a-z')
root@wazuh-agent1:/home/vagrant# curl -k -i -X POST "https://$MGR/download" -H "protocol-version: 1" \
  -H "Authorization: Wazuh $ID:$TS:$MAC" --data-binary "$BODY"
HTTP/1.1 401 Unauthorized
Connection: keep-alive
Content-Length: 52
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:57:03 GMT
{"error":"Invalid client authentication","code":401}

root@wazuh-agent1:/home/vagrant# dl config web-servers -i -H "protocol-version: 2"
HTTP/1.1 200 OK
Connection: keep-alive
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:58:05 GMT
Content-Type: application/octet-stream
Transfer-Encoding: chunked

#web-servers
!76 agent.conf
<agent_config>

  <!-- Shared agent configuration here -->

</agent_config>
!134 web.conf
<agent_config>
  <localfile><location>/var/log/nginx/access.log</location><log_format>syslog</log_format></localfile>
</agent_config>
```

`/download` sits behind the same `AuthMiddleware` as `/stateless`, so these confirm the endpoint
inherits it rather than bypassing it.

*A forged MAC* (`deadbeef`) is `401`, so possession of a valid `client.keys` entry — not merely
reachability of the port — is what admits a request.

*A timestamp 400 s in the past* is `401` even though the MAC is correctly computed with the agent's
real key. The timestamp is inside the signed byte sequence, so a captured request cannot be replayed
once it ages past `auth_max_request_age` (300 s). This is the case that produces the
`expired_request` WARNING seen in the abort evidence below.

> **The third case does not show what it appears to.** `dl` already sends `protocol-version: 1`, so
> `-H "protocol-version: 2"` produced a request carrying *two* such headers while the MAC was still
> computed over version `1`; the server matched and returned `200`. It therefore demonstrates
> duplicate-header handling, not version rejection, and should be re-run with a plain `curl` sending
> a single `protocol-version: 2` header — where the version is part of the signed sequence, so the
> MAC no longer matches and the expected result is `401`.

</details>

<details>
<summary>Chunked vs buffered</summary>

```
root@wazuh-agent1:/home/vagrant# dl config web-servers -D /dev/stderr -o /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0HTTP/1.1 200 OK
Connection: keep-alive
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:58:44 GMT
Content-Type: application/octet-stream
Transfer-Encoding: chunked

100   306    0   252  100    54   5478   1173 --:--:-- --:--:-- --:--:--  6800

root@wazuh-agent1:/home/vagrant# dl config no-such-group -D /dev/stderr -o /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0HTTP/1.1 404 Not Found
Connection: keep-alive
Content-Length: 41
Server: wazuh-manager-remoted
Date: Tue, 04 Aug 2026 07:59:06 GMT

100    97  100    41  100    56    885   1209 --:--:-- --:--:-- --:--:--  2108
```

The central transport claim, on one route. A `200` carries `Transfer-Encoding: chunked` and **no**
`Content-Length`; a `404` carries `Content-Length: 41` and no `Transfer-Encoding`.

That distinction is why `ResponseMode::Streamable` had to be a *registration-time* property. A
builder's output mode is fixed when it is created, and the transport creates the builder at dispatch
— before the handler has decided anything. A streamable route therefore keeps the request handle
instead, and the responder picks the mode once the outcome is known: buffered for an error, chunked
for a transfer. Without that, an error on a streaming route would have to be sent as a one-chunk
chunked response, and every 404 would look to the agent like a zero-length download.

It also matters that the file is opened and `fstat`ed *before* the first byte is written: that
ordering is what makes a missing resource a clean `404` rather than a `200` that dies mid-body.

</details>

<details>
<summary>Concurrency</summary>

```
root@wazuh-agent1:/home/vagrant# for i in $(seq 1 20); do dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n" & done; wait
[1] 1863
[2] 1865
[3] 1867
[4] 1868
[5] 1870
[6] 1871
[7] 1874
[8] 1876
[9] 1877
[10] 1878
[11] 1879
[12] 1880
[13] 1881
[14] 1883
[15] 1884
[16] 1891
[17] 1894
[18] 1895
[19] 1896
[20] 1897
4 200 33538926
5 200 33531855
20 200 32689196
6 200 32441730
13 200 32287141
12 200 32161210
17 200 31993872
16 200 31740819
11 200 31460648
3 200 31314234
2 200 31289745
15 200 31282360
10 200 31238799
7 200 31202105
18 200 31202495
9 200 31169199
19 200 31145078
14 200 31152416
8 200 31140949
1 200 31101995
[1]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[2]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[3]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[4]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[5]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[6]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[7]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[8]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[9]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[10]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[11]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[12]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[13]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[14]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[15]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[16]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[17]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[18]   Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[19]-  Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
[20]+  Done                    dl wpk huge-signed.wpk -o /dev/null -s -w "$i %{http_code} %{speed_download}\n"
```

20 simultaneous 1 GB transfers, all `200`, ~21.5 GB moved at **~635 MB/s aggregate**. The number
that matters is the *spread*, not the total: per-stream rates land between 31.1 and 33.5 MB/s, a
7.8% band, so no transfer starved and none monopolised. That is the property the per-chunk yield
between worker thread and I/O strand buys — a greedy read loop would show a few fast streams and a
long tail. Aggregate throughput is ~60% of the single-stream figure, i.e. the NIC/CPU ceiling rather
than a server-side limit.

</details>

<details>
<summary>Memory under load</summary>

```
root@wazuh-master:/home/vagrant# sudo python3 devel/wazuh/src/remoted/remoted_module/tools/monitor
.py --duration 60
monitoring pid 2006 (port 9443) for 60s, every 0.1s
baseline: RSS 30.9 MiB, 47 threads, 24 fds

  t(s)  RSS MiB  thr   fds  estab   tw  other
   0.0     30.9   47    24      0    0      0

  27.6     30.7   47    26      0    0      0
  27.7     29.4   47    26      0    0      0
  27.8     29.4   47    26      0    0      0
  27.9     29.5   47    26      0    0      0
  28.0     29.5   47    26      0    0      0
  28.1     28.9   47    26      0    0      0
  28.2     28.9   47    26      0    0      0
  28.3     29.0   47    26      0    0      0
  28.5     28.9   47    26      0    0      0
  28.6     29.0   47    26      0    0      0
  28.8     29.0   47    26      0    0      0
  28.9     29.1   47    26      0    0      0
  29.0     29.0   47    26      0    0      0
  29.1     28.9   47    26      0    0      0
  29.2     29.0   47    26      0    0      0
  29.4     29.0   47    26      0    0      0
  30.2     29.1   47    26      0    0      0
  31.5     29.0   47    26      0    0      0
  32.0     28.8   47    24      0    0      0

root@wazuh-agent1:/home/vagrant# dl wpk huge.wpk -o /dev/null -w 'speed=%{speed_download} B/s\n'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--  
100  141M    0  141M  100    48   387M    131 --:--:-- --:--:-- --:--:--  
100  574M    0  574M  100    48   421M     35  0:00:01  0:00:01 --:--:--  
100 1039M    0 1039M  100    48   439M     20  0:00:02  0:00:02 --:--:-- 
100 1512M    0 1512M  100    48   450M     14  0:00:03  0:00:03 --:--:--  
100 1987M    0 1987M  100    48   455M     11  0:00:04  0:00:04 --:--:--  
100 2048M    0 2048M  100    48   455M     10  0:00:04  0:00:04 --:--:--  
461M speed=477614902 B/s
```

2 GiB moved while RSS goes 30.9 → 28.8 MiB and thread count stays pinned at 47: memory is
independent of file size, which is the whole reason for streaming rather than buffering. fds go
24 → 26 → 24, i.e. the transfer's descriptors are released.

Note the `estab`/`tw`/`other` columns read zero throughout — this run predates a fix to
`monitor.py`, whose `--port` defaulted to `9443` while the server listens on `1517`, so the
connection counters were watching the wrong port. RSS, threads and fds come from `/proc/<pid>` and
are unaffected. The default is now `1517`.

</details>

<details>
<summary>Abortion mid-transfer</summary>

```
root@wazuh-agent1:/home/vagrant# dl wpk huge.wpk -o /dev/null --max-time 2 ; echo "curl exit=$?"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  878M    0  878M  100    48   439M     24  0:00:02  0:00:01  0:00:01  439M
curl: (28) Operation timed out after 2000 milliseconds with 921198585 bytes received
curl exit=28

root@wazuh-master:/home/vagrant# sudo ls -l /proc/$(pgrep -f wazuh-manager-remoted)/fd | wc -l
25
root@wazuh-master:/home/vagrant# sudo grep -iE "remoted.*(error|warn)" /var/wazuh-manager/logs/wazuh-manager.log | tail -3
2026/08/03 16:23:03 wazuh-manager-remoted:endpoints: WARNING: Rejected 1 request(s) in the last 90 s whose timestamp fell outside the accepted window (expired_request). Check clock synchronization between the agents and the manager; the window is set by 'auth_max_request_age' and 'auth_max_future_skew'.
2026/08/04 07:57:03 wazuh-manager-remoted:endpoints: WARNING: Rejected 1 request(s) in the last 90 s whose timestamp fell outside the accepted window (expired_request). Check clock synchronization between the agents and the manager; the window is set by 'auth_max_request_age' and 'auth_max_future_skew'.
```

The client walks away at 921 MB of 2 GiB. `ls -l` prints a `total` line, so 25 lines is **24 fds** —
the baseline — meaning the `FileByteSource` was destroyed when the pump dropped it. The two WARNINGs
are pre-existing clock-skew entries from earlier runs, unrelated to the abort: a client
disconnecting mid-download is a normal event, not an error, so silence is the correct behaviour.

`curl` can only report that it gave up (exit 28); it cannot show whether the server emitted a
terminating chunk. A raw-socket client can, and that is the actual correctness property — the peer
must not be able to mistake a truncated transfer for a complete one:

```
root@wazuh-master:/home/vagrant# sudo python3 /tmp/rawframe.py --type wpk --id huge.wpk --abort-after 50000000
decoded bytes    : 50003968
terminator (0\r\n): NO -- truncated transfer
client           : aborted deliberately after 50000000 bytes
```

</details>

<details>
<summary>New chunk-size option</summary>

```
root@wazuh-master:/home/vagrant# cat /var/wazuh-manager/etc/wazuh-manager-internal-options.conf | grep remoted
remoted.http_stream_chunk_size=1048576
root@wazuh-master:/home/vagrant# sudo systemctl restart wazuh-manager
root@wazuh-master:/home/vagrant# sudo python3 /tmp/rawframe.py 001 wpk:huge-signed.wpk | grep -E 'chunk count|steady'
chunk count      : 1025
steady-state size: 1048576 bytes (last chunk 328936 -- the remainder)

root@wazuh-master:/home/vagrant# cat /var/wazuh-manager/etc/wazuh-manager-internal-options.conf | grep remoted
remoted.http_stream_chunk_size=65536
root@wazuh-master:/home/vagrant# sudo systemctl restart wazuh-manager
root@wazuh-master:/home/vagrant# sudo python3 /tmp/rawframe.py 001 wpk:huge-signed.wpk | grep -E 'chunk count|steady'
chunk count      : 16390
steady-state size: 65536 bytes (last chunk 1256 -- the remainder)

root@wazuh-master:/home/vagrant# cat /tmp/rawframe.py 
#!/usr/bin/env python3
"""
Raw-socket client for POST /download: shows the HTTP framing curl hides.

curl decodes `Transfer-Encoding: chunked` transparently, so it cannot tell you the chunk
sizes, the chunk count, or whether the terminating 0-length chunk actually arrived. This
speaks the wire directly and reports all three, while streaming the body (never buffering
it), so it works on a 2 GiB WPK.

It signs exactly like the agent does: AES-CMAC over

    WAZUH-REQUEST\n<version>\n<METHOD>\n<target>\n<agent-id>\n<timestamp>\n<body>

taking the key from client.keys. On the manager that file holds every agent's key, so this
can sign as any of them; on an agent it holds only its own.

Examples
--------
  # Prove chunked framing and see the sizes
  sudo python3 rawframe.py --type config --id default --show-chunks 5

  # Terminator check on a big file, discarding the body
  sudo python3 rawframe.py --type wpk --id huge.wpk

  # Walk away mid-transfer (aborted-download case)
  sudo python3 rawframe.py --type wpk --id huge.wpk --abort-after 50000000

  # Slow reader: small reads, spaced out (per-write timeout must rearm per chunk)
  sudo python3 rawframe.py --type wpk --id test-signed.wpk --trickle-bytes 4096 --trickle-pause 0.05

  # Keep the bytes and compare against the source
  sudo python3 rawframe.py --type wpk --id test-signed.wpk --save /tmp/got.wpk

The older positional form still works, and mixes with the flags above:

  sudo python3 rawframe.py 001 wpk:test-signed.wpk
  sudo python3 rawframe.py 001 config:default --show-chunks 5
"""

import argparse
import hashlib
import json
import os
import socket
import ssl
import subprocess
import sys
import time

CANONICAL_PREFIX = "WAZUH-REQUEST\n"
PROTOCOL_VERSION = "1"
TARGET = "/download"
METHOD = "POST"

DEFAULT_KEYS = "/var/wazuh-manager/etc/client.keys"
AGENT_KEYS = "/var/ossec/etc/client.keys"

# openssl names the CMAC cipher by key length; client.keys stores the key as hex.
CIPHER_BY_HEX_LEN = {32: "AES-128-CBC", 48: "AES-192-CBC", 64: "AES-256-CBC"}


def read_key(keys_path, agent_id=None):
    """Return (id, name, key_hex) from client.keys: the named agent, or the first entry."""
    try:
        with open(keys_path, "r", encoding="utf-8") as handle:
            lines = handle.read().splitlines()
    except PermissionError:
        sys.exit("cannot read %s -- run under sudo" % keys_path)
    except OSError as err:
        sys.exit("cannot read %s: %s" % (keys_path, err))

    for line in lines:
        fields = line.split()
        if len(fields) < 4:
            continue
        if agent_id is None or fields[0] == agent_id:
            return fields[0], fields[1], fields[3]

    sys.exit("no usable entry in %s%s"
             % (keys_path, "" if agent_id is None else " for agent id %s" % agent_id))


def cmac_hex(key_hex, message):
    """AES-CMAC via `openssl mac`, which is what the manager verifies against."""
    cipher = CIPHER_BY_HEX_LEN.get(len(key_hex))
    if cipher is None:
        sys.exit("key is %d hex chars; expected 32, 48 or 64" % len(key_hex))

    cmd = ["openssl", "mac", "-macopt", "cipher:" + cipher,
           "-macopt", "hexkey:" + key_hex, "CMAC"]
    try:
        result = subprocess.run(cmd, input=message.encode(), capture_output=True, check=True)
    except FileNotFoundError:
        sys.exit("openssl not found")
    except subprocess.CalledProcessError as err:
        sys.exit("openssl mac failed: %s" % err.stderr.decode(errors="replace").strip())
    return result.stdout.decode().strip().lower()


def build_request(agent_id, key_hex, host, resource_type, resource_id):
    """Return the raw request bytes an agent would put on the wire."""
    # Signed and sent byte-for-byte identically -- reserialising between the two is the
    # classic way to produce a MAC that verifies against something the server never saw.
    body = json.dumps({"resource_type": resource_type, "resource_id": resource_id},
                      separators=(",", ":"))
    timestamp = str(int(time.time()))

    canonical = CANONICAL_PREFIX + "\n".join(
        [PROTOCOL_VERSION, METHOD, TARGET, agent_id, timestamp, body])
    mac = cmac_hex(key_hex, canonical)

    head = "".join([
        "%s %s HTTP/1.1\r\n" % (METHOD, TARGET),
        "Host: %s\r\n" % host,
        "protocol-version: %s\r\n" % PROTOCOL_VERSION,
        "Authorization: Wazuh %s:%s:%s\r\n" % (agent_id, timestamp, mac),
        "Content-Type: application/json\r\n",
        "Content-Length: %d\r\n" % len(body),
        "Connection: close\r\n\r\n",
    ])
    return head.encode() + body.encode(), body


def connect(host, port, timeout):
    context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
    context.check_hostname = False
    context.verify_mode = ssl.CERT_NONE  # self-signed manager certificate
    raw = socket.create_connection((host, port), timeout=timeout)
    return context.wrap_socket(raw, server_hostname=host)


class ChunkReader:
    """
    Incremental `Transfer-Encoding: chunked` parser.

    Streams: it never holds more than one recv() buffer, so a 2 GiB body costs nothing.
    Records every chunk size, because those sizes are what prove the configured chunk size
    reached the wire -- a decoded-body comparison alone would hide it.
    """

    def __init__(self):
        self.buffer = b""
        self.remaining = 0        # bytes left in the current chunk's data
        self.sizes = []
        self.decoded = 0
        self.terminated = False   # saw the 0-length chunk
        self.digest = hashlib.sha256()
        self.malformed = None

    def feed(self, data, sink=None):
        """Consume raw bytes; write decoded payload to `sink` if given."""
        self.buffer += data
        while True:
            if self.terminated or self.malformed:
                return
            if self.remaining > 0:
                take = min(self.remaining, len(self.buffer))
                if take == 0:
                    return
                payload = self.buffer[:take]
                self.buffer = self.buffer[take:]
                self.remaining -= take
                self.decoded += take
                self.digest.update(payload)
                if sink is not None:
                    sink.write(payload)
                if self.remaining == 0:
                    self.expect_crlf = True
                continue

            if getattr(self, "expect_crlf", False):
                if len(self.buffer) < 2:
                    return
                if self.buffer[:2] != b"\r\n":
                    self.malformed = "chunk not followed by CRLF"
                    return
                self.buffer = self.buffer[2:]
                self.expect_crlf = False
                continue

            end = self.buffer.find(b"\r\n")
            if end < 0:
                if len(self.buffer) > 64:
                    self.malformed = "oversized chunk-size line"
                return
            line = self.buffer[:end]
            self.buffer = self.buffer[end + 2:]
            token = line.split(b";", 1)[0].strip()  # drop any chunk extension
            try:
                size = int(token, 16)
            except ValueError:
                self.malformed = "bad chunk size %r" % token
                return
            if size == 0:
                self.terminated = True
                return
            self.sizes.append(size)
            self.remaining = size


def summarise_sizes(sizes, show):
    if not sizes:
        return
    print("chunk count      : %d" % len(sizes))
    distinct = sorted(set(sizes))
    if len(distinct) <= 4:
        print("chunk sizes      : %s" % ", ".join(str(s) for s in distinct))
    else:
        print("chunk sizes      : %d distinct, min %d, max %d"
              % (len(distinct), distinct[0], distinct[-1]))
    body = sizes[:-1] if len(sizes) > 1 else sizes
    if body and len(set(body)) == 1:
        print("steady-state size: %d bytes (last chunk %d -- the remainder)"
              % (body[0], sizes[-1]))
    if show:
        print("first %d sizes    : %s" % (min(show, len(sizes)),
                                          ", ".join(str(s) for s in sizes[:show])))


def apply_positional(parser, args):
    """
    Fold the older `[<agent-id>] <type>:<resource-id>` positional form into the flags.

    Kept because it is what is already in people's shell history. Explicit flags win, so a
    mixed invocation is unambiguous rather than silently order-dependent.
    """
    for token in args.positional:
        if ":" in token:
            resource_type, _, resource_id = token.partition(":")
            if resource_type not in ("config", "wpk"):
                parser.error("unknown resource type %r in %r; expected config: or wpk:"
                             % (resource_type, token))
            if args.resource_type is None:
                args.resource_type = resource_type
            if args.resource_id is None:
                args.resource_id = resource_id
        elif args.agent_id is None:
            args.agent_id = token
        else:
            parser.error("unexpected argument %r" % token)

    missing = [name for name, value in (("--type", args.resource_type),
                                        ("--id", args.resource_id)) if value is None]
    if missing:
        parser.error("missing %s -- pass them as flags, or positionally as `<agent-id> "
                     "<type>:<resource-id>` (e.g. `001 wpk:test-signed.wpk`)"
                     % " and ".join(missing))


def main():
    parser = argparse.ArgumentParser(
        description="Raw-socket POST /download client that reports the chunk framing.")
    parser.add_argument("--host", default="192.168.56.11")
    parser.add_argument("--port", type=int, default=1517)
    parser.add_argument("--type", choices=("config", "wpk"), dest="resource_type")
    parser.add_argument("--id", dest="resource_id",
                        help="group / multigroup selector, or WPK filename")
    parser.add_argument("positional", nargs="*", metavar="[AGENT_ID] TYPE:RESOURCE_ID",
                        help="older positional form, e.g. `001 wpk:test-signed.wpk`")
    parser.add_argument("--keys", help="client.keys path (default: manager's, then agent's)")
    parser.add_argument("--agent-id", help="sign as this agent id (default: first entry)")
    parser.add_argument("--save", help="write the decoded body here")
    parser.add_argument("--show-chunks", type=int, default=0, metavar="N",
                        help="print the first N chunk sizes")
    parser.add_argument("--abort-after", type=int, default=0, metavar="BYTES",
                        help="close the socket after roughly BYTES decoded (aborted transfer)")
    parser.add_argument("--trickle-bytes", type=int, default=65536, metavar="N",
                        help="recv() size (default 65536)")
    parser.add_argument("--trickle-pause", type=float, default=0.0, metavar="SEC",
                        help="sleep between reads, to act as a slow reader")
    parser.add_argument("--timeout", type=float, default=30.0,
                        help="per-read socket timeout in seconds")
    args = parser.parse_args()
    apply_positional(parser, args)

    keys_path = args.keys or (DEFAULT_KEYS if os.path.exists(DEFAULT_KEYS) else AGENT_KEYS)
    agent_id, agent_name, key_hex = read_key(keys_path, args.agent_id)
    request, body = build_request(agent_id, key_hex, args.host,
                                  args.resource_type, args.resource_id)

    print("signing as       : agent %s (%s) from %s" % (agent_id, agent_name, keys_path))
    print("request body     : %s" % body)
    print("target           : https://%s:%d%s" % (args.host, args.port, TARGET))
    print("-" * 72)

    started = time.time()
    sock = connect(args.host, args.port, args.timeout)
    sock.sendall(request)

    head = b""
    while b"\r\n\r\n" not in head:
        piece = sock.recv(8192)
        if not piece:
            break
        head += piece
    if b"\r\n\r\n" not in head:
        sys.exit("connection closed before the response head was complete")
    head, leftover = head.split(b"\r\n\r\n", 1)

    print(head.decode(errors="replace"))
    print("-" * 72)

    lowered = head.decode(errors="replace").lower()
    chunked = "transfer-encoding: chunked" in lowered
    buffered = "content-length:" in lowered
    print("framing          : %s" % ("chunked" if chunked else
                                     ("buffered (Content-Length)" if buffered else "unknown")))

    sink = open(args.save, "wb") if args.save else None
    aborted = False
    try:
        if not chunked:
            # An error response is buffered; just drain and hash it.
            digest = hashlib.sha256()
            total = len(leftover)
            digest.update(leftover)
            if sink:
                sink.write(leftover)
            while True:
                piece = sock.recv(args.trickle_bytes)
                if not piece:
                    break
                total += len(piece)
                digest.update(piece)
                if sink:
                    sink.write(piece)
            print("body bytes       : %d" % total)
            print("body sha256      : %s" % digest.hexdigest())
            if total and total < 4096:
                print("body             : %s" % leftover.decode(errors="replace"))
            return

        reader = ChunkReader()
        reader.feed(leftover, sink)
        while not reader.terminated and not reader.malformed:
            if args.abort_after and reader.decoded >= args.abort_after:
                aborted = True
                break
            try:
                piece = sock.recv(args.trickle_bytes)
            except socket.timeout:
                print("!! read timed out after %.1fs" % args.timeout)
                break
            if not piece:
                break  # server closed
            reader.feed(piece, sink)
            if args.trickle_pause:
                time.sleep(args.trickle_pause)
    finally:
        if aborted:
            sock.close()  # abrupt: the agent walking away mid-transfer
        else:
            try:
                sock.close()
            except OSError:
                pass
        if sink:
            sink.close()

    elapsed = time.time() - started
    print("-" * 72)
    summarise_sizes(reader.sizes, args.show_chunks)
    print("decoded bytes    : %d" % reader.decoded)
    print("terminator (0\\r\\n): %s" % ("YES -- complete transfer" if reader.terminated
                                       else "NO -- truncated transfer"))
    if reader.malformed:
        print("!! framing error : %s" % reader.malformed)
    if aborted:
        print("client           : aborted deliberately after %d bytes" % args.abort_after)
    print("body sha256      : %s" % reader.digest.hexdigest())
    print("elapsed          : %.2fs" % elapsed)
    if elapsed > 0 and reader.decoded:
        print("throughput       : %.1f MiB/s" % (reader.decoded / elapsed / 1048576))
    if args.save:
        print("saved to         : %s" % args.save)


if __name__ == "__main__":
    main()
```

Ties the new internal option to observable behaviour: `1048576` produces 1025 chunks of 1 MiB,
`65536` produces 16390 of 64 KiB, over the same 1 GB file. The decoded body is byte-identical in
both cases — the option changes only how bytes are framed on the wire, never which bytes arrive.

`curl` cannot show this. It decodes `Transfer-Encoding: chunked` transparently, so the chunk sizes,
the chunk count and the presence of the terminating 0-length chunk are all invisible to it; a
`sha256sum` of the output would be identical whatever the framing. The raw-socket client above
parses the framing itself and reports all three, while never buffering the body, so the same tool
works against a 2 GiB file.

Note the manager reads internal options **only** from
`/var/wazuh-manager/etc/wazuh-manager-internal-options.conf` (`defs.h` defines `WAZUH_LDEFINES` per
`#ifdef CLIENT`, and leaves `WAZUH_DEFINES` undefined for the manager). Settings placed in
`local_internal_options.conf` — the agent's file — are silently ignored, and `_read_file()` returns
the **first** matching line, so appending a second entry for the same option has no effect.

</details>

<details>
<summary>Unit tests</summary>

```
root@wazuh-master:/home/vagrant/wazuh/src/build# ctest -R remoted_module_utest -V
...
100% tests passed, 0 tests failed out of 1

Label Time Summary:
remoted_module_utest    =  11.93 sec*proc (1 test)

Total Test time (real) =  11.93 sec
```

203 tests over 36 suites. To reproduce, stop `wazuh-manager` first: `RemotedModuleTest` binds port
1514 for real, so a running manager makes those four cases fail with `bind: Address already in use`.

</details>

<details>
<summary>ASan tests</summary>

```
$ cmake -DUNIT_TEST=ON \
    -DCMAKE_CXX_FLAGS="-g -fsanitize=address -fsanitize=undefined -fsanitize=leak" \
    -DCMAKE_C_FLAGS="-g -fsanitize=address -fsanitize=undefined -fsanitize=leak" ..
$ cmake --build . --target remoted_module_utest
$ ./remoted/remoted_module/test/remoted_module_utest \
      --gtest_filter='HttpServerStreamingTest.*:ShutdownRace.*'
[ RUN      ] HttpServerStreamingTest.StreamsAMultiChunkBodyByteExactly
[       OK ] HttpServerStreamingTest.StreamsAMultiChunkBodyByteExactly (62 ms)
[ RUN      ] HttpServerStreamingTest.ChunkSizeFollowsTheConfiguredValue
[       OK ] HttpServerStreamingTest.ChunkSizeFollowsTheConfiguredValue (62 ms)
[ RUN      ] HttpServerStreamingTest.AbortedTransferSendsNoTerminatorAndReleasesTheSource
[       OK ] HttpServerStreamingTest.AbortedTransferSendsNoTerminatorAndReleasesTheSource (405 ms)
[ RUN      ] HttpServerStreamingTest.HealthyTransferOutlastingTheRequestTimeoutIsNotCut
[       OK ] HttpServerStreamingTest.HealthyTransferOutlastingTheRequestTimeoutIsNotCut (2652 ms)
[ RUN      ] HttpServerStreamingTest.TrickleReaderKeepsTheTransferAliveBeyondTheWriteTimeout
[       OK ] HttpServerStreamingTest.TrickleReaderKeepsTheTransferAliveBeyondTheWriteTimeout (6334 ms)
[ RUN      ] ShutdownRace.StopSequenceSurvivesAnInFlightForward
[       OK ] ShutdownRace.StopSequenceSurvivesAnInFlightForward (106 ms)
[ RUN      ] ShutdownRace.StopSequenceSurvivesAnInFlightStream
[       OK ] ShutdownRace.StopSequenceSurvivesAnInFlightStream (151 ms)
[==========] 7 tests from 2 test suites ran. (9776 ms total)
[  PASSED  ] 7 tests.
$ echo $?
0
```

Exit code `0` with no LeakSanitizer report is the assertion here — LSan runs at process exit, *after*
gtest has already printed `PASSED`, so a green gtest summary alone does not mean a clean ASan run.

An earlier revision of `AbortedTransferSendsNoTerminatorAndReleasesTheSource` did leak, and CI caught
it: its "server is still healthy" follow-up read only 16 KiB of a 16 MiB payload, making it a *second*
aborted transfer whose pump was still in flight when `server->stop()` ran — leaking the pump, its
source, the chunk buffer and the connection's OpenSSL allocations (~17 MB). The follow-up now reads
to completion and asserts the payload arrives intact with its terminator, which is the stronger
health check regardless.

Built with `-fsanitize=address,undefined,leak`. These two suites are the ones worth sanitising: a
streamed response is the only path that hands a buffer between a worker thread and the connection's
I/O strand, repeatedly, for the whole duration of a transfer — and then has to survive teardown
while a transfer is still in flight. That is a use-after-free shape a plain build will usually run
straight past, because the freed memory is still mapped and still holds plausible bytes.

`ShutdownRace.StopSequenceSurvivesAnInFlightStream` is the specific case: `stopAccepting()` drains
the handler pool, but a `StreamPump` that has already handed a chunk to the strand has its
continuation queued somewhere neither phase 1 nor phase 3 of the shutdown sequence knows about.
Before this PR that was only argued for in a comment ("RESTinio guarantees the write callback always
fires"); it is now executed, and executed under ASan.

</details>

  - [x] Linux — every module source compiles clean with `-Wall -Wextra`; each commit verified to
        build standalone in a detached worktree
  - [ ] Windows — N/A, manager-only component
  - [ ] MAC OS X — N/A, manager-only component

- Memory tests for Linux
  - [ ] Coverity — not run
  - [ ] Valgrind (memcheck and descriptor leaks check) — not run; descriptor release on aborted
        transfers covered by the abort evidence above
  - [x] AddressSanitizer — `HttpServerStreamingTest.*` and `ShutdownRace.*` under
        `-fsanitize=address,undefined,leak`, 7/7 clean (evidence above). These are the two suites
        where a streamed transfer's worker-thread ↔ I/O-strand handoff could leave a dangling
        buffer, which is the failure mode a plain build would not surface.

### Artifacts Affected

- `libremoted_module.so` (manager) — transport and endpoint
- `wazuh-manager-remoted` (manager) — `secure.c` reads the two internal options above
- New: `src/remoted/remoted_module/tools/send_download.py` and `tools/monitor.py`
  (development/test tools, not packaged)
- New: `test/unit/testTlsServer.hpp` — TLS test fixtures extracted from `shutdownRace_test.cpp` so
  the streaming tests share one signed client instead of growing a second copy

### Configuration Changes

Two internal options in `/var/wazuh-manager/etc/wazuh-manager-internal-options.conf`. **No default
behaviour changes** — both resolve to the value that was previously compiled in.

| Option | Default | Range | Status |
| --- | --- | --- | --- |
| `remoted.http_stream_chunk_size` | `65536` | 4096–1048576 | New |
| `remoted.http_max_parallel_connections` | `512` | 1–8192 | Previously hardcoded, now settable |

Both are documented in `docs/ref/modules/remoted/configuration.md` alongside the existing
`remoted.http_*` options. They are deliberately *not* added to any shipped `.conf`: manager defaults
live in code via `getDefine_Int_default()`, and `wazuh-manager-internal-options.conf` is an
overrides-only file that survives upgrades — shipping a default in it would pin that value forever.

The endpoint itself introduces no `<remote>` configuration; it reuses the HTTPS settings already in
place.

### Tests Introduced

`test/unit/downloadEndpoint_test.cpp` — 44 tests over the endpoint's pure surface plus the handler
end to end against temporary directories:

- `parseRequest`: valid config/wpk; missing, extra and wrong-typed members; non-object; empty;
  truncated; over the 4 KiB pre-parse cap; deeply nested input (asserts iterative parsing does not
  overflow the stack); unknown and wrongly-cased `resource_type`.
- Resource-id grammars: wazuh's group charset; multigroup selectors including leading, trailing and
  doubled separators; a traversable entry anywhere in a selector; `.`/`..`; embedded NUL; backslash;
  non-ASCII; over-length; WPK suffix and charset rules.
- `multigroupDirName`: pinned against the directories a live manager created (`97d66859` for
  `web-servers,databases`, `76ba5a3b` for `alpha,beta`), plus order sensitivity, so nobody "fixes"
  it by sorting.
- `locateResource`: single group, multigroup selector, WPK, injected base directories.
- `FileByteSource`: sub-chunk, multi-chunk and exactly-aligned files; empty file; zero-capacity read;
  missing file; a directory (`open(O_RDONLY)` on one succeeds on Linux, so the `S_ISREG` check is
  what makes it a 404); a symlink refused by `O_NOFOLLOW`; a read error throwing rather than looking
  like end-of-stream.
- Mid-transfer modification, one case per detection path: the file truncated (EOF arrives before the
  length at open), the file grown (a WPK still being staged), and the file rewritten to the *same*
  length (invisible to the byte count, caught by mtime) — each must throw rather than report a clean
  end-of-stream. Plus an untouched file streaming to completion, so the detection cannot fire on a
  normal transfer.
- Handler: 400 on a malformed body; 404 when the resolved file is absent; streams the named group;
  streams the effective config for a multigroup selector; streams a staged WPK; `Content-Type` and
  chunked-vs-buffered selection.

`test/unit/httpServer_test.cpp` — 5 `HttpServerStreamingTest` cases driving a real TLS client
against a live server, because chunk framing only exists on the wire:

- `StreamsAMultiChunkBodyByteExactly` — decoded body equals the source across many chunks.
- `ChunkSizeFollowsTheConfiguredValue` — the configured size reaches the wire; a decoded-body
  comparison alone would hide it.
- `AbortedTransferSendsNoTerminatorAndReleasesTheSource` — a client walking away leaves no
  terminating chunk (so the peer cannot mistake a partial file for a complete one) and drops the
  source.
- `HealthyTransferOutlastingTheRequestTimeoutIsNotCut` — a transfer longer than
  `http_request_timeout` is not torn down, since the per-write timeout rearms per chunk.
- `TrickleReaderKeepsTheTransferAliveBeyondTheWriteTimeout` — same property against a deliberately
  slow reader, i.e. a real agent on a poor link.

`test/unit/shutdownRace_test.cpp` — `StopSequenceSurvivesAnInFlightStream`: a streamed response is
the one path that bounces worker-pool ↔ connection-strand *outside* the `DeferredForwarder` the
four-phase shutdown drains, so teardown mid-transfer is the use-after-free shape this file exists to
catch. Previously only argued for in a comment; now executed, and under ASan.

## Review Checklist

- [x] Code changes reviewed
- [x] Configuration changes documented — two internal options, in
      `docs/ref/modules/remoted/configuration.md`
- [x] Developer documentation reflects the changes — `remoted_module/README.md` gains a
      *Streamed responses* section covering the pull-based `IByteSource`, `ResponseMode`, the abort
      semantics and resource resolution

### DIFF
--- a/src/remoted/src/secure.c
+++ b/src/remoted/src/secure.c
@@ -270,6 +270,10 @@ STATIC void remoted_module_https_config(remoted_module_config_t *rm_config) {
     rm_config->http_max_pipelined_requests = getDefine_Int_default("remoted", "http_max_pipelined_requests", 1, 64, 4);
     rm_config->http_concurrent_accepts = getDefine_Int_default("remoted", "http_concurrent_accepts", 1, 64, 2);
     rm_config->http_buffer_size = getDefine_Int_default("remoted", "http_buffer_size", 1, 1048576, 8192);
+    // Bytes per chunk when streaming a response body (POST /download). Bounded at 1 MiB because
+    // this is per IN-FLIGHT TRANSFER: the worst case is roughly this times the number of
+    // simultaneous downloads, so a large value trades memory for fewer read/write round trips.
+    rm_config->http_stream_chunk_size = getDefine_Int_default("remoted", "http_stream_chunk_size", 4096, 1048576, 65536);
 
     // Memory-management (backpressure) tunables. These bound in-memory resource usage rather
     // than tune the transport itself; the C++ side still clamps max_inflight_bytes up to at
@@ -310,6 +314,25 @@ STATIC void w_remoted_build_module_config(const remoted *logr, remoted_module_co
     // back to its own default.
     rm_config->port = logr->https.port;
     remoted_module_https_config(rm_config);
+    // rm_config->max_inflight_bytes caps the HTTPS server's in-flight (unprocessed)
+    // request payload before it sheds load with 503. NOT logr->queue_size (that is an
+    // event COUNT, not bytes).
+    rm_config->max_inflight_bytes = (256 * 1024 * 1024); // 256 MiB
+    // rm_config->max_parallel_connections caps simultaneous HTTPS connections, bounding the
+    // read-phase memory peak (~ max_parallel_connections * max body size).
+    //
+    // It is also the ONLY bound on concurrent streamed responses (POST /download): chunked output
+    // rearms http_write_timeout per chunk, so a client that keeps reading slowly can hold a
+    // transfer open indefinitely, and there is no per-stream limiter. A mass upgrade -- the whole
+    // fleet fetching a WPK at once, many over slow links -- is therefore bounded only by this
+    // value, which is why it is settable rather than fixed.
+    rm_config->max_parallel_connections =
+        getDefine_Int_default("remoted", "http_max_parallel_connections", 1, 8192, 512);
+    // rm_config->max_deferred_requests caps requests parked awaiting a downstream service (503
+    // over it). No Retry-After is sent: the agent runs its own retry/backoff on a 503.
+    rm_config->max_deferred_requests = 256;
+    // rm_config->keystore_refresh_interval governs how often the C++ Keystore checks
+    // client.keys for changes (hot-reload)
     rm_config->keystore_refresh_interval = keyupdate_interval;
     rm_config->worker_node = logr->worker_node;
     rm_config->verification_mode = logr->https.verification_mode;
--- a/src/remoted/src/secure.c
+++ b/src/remoted/src/secure.c
@@ -278,9 +278,19 @@ STATIC void remoted_module_https_config(remoted_module_config_t *rm_config) {
     // Memory-management (backpressure) tunables. These bound in-memory resource usage rather
     // than tune the transport itself; the C++ side still clamps max_inflight_bytes up to at
     // least one max-size request at start(), so a too-small value can't reject everything.
+    // max_inflight_bytes caps the HTTPS server's in-flight (unprocessed) request payload
+    // before it sheds load with 503. NOT logr->queue_size (that is an event COUNT, not bytes).
     rm_config->max_inflight_bytes =
         getDefine_Int_default("remoted", "max_inflight_bytes", 1048576, 1073741824, 268435456);
+    // max_parallel_connections caps simultaneous HTTPS connections, bounding the read-phase
+    // memory peak (~ max_parallel_connections * max body size). It is also the ONLY bound on
+    // concurrent streamed responses (POST /download): chunked output rearms http_write_timeout
+    // per chunk, so a client that keeps reading slowly can hold a transfer open indefinitely.
+    // A mass upgrade (the whole fleet fetching a WPK at once, many over slow links) is therefore
+    // bounded only by this value, which is why it is settable rather than fixed.
     rm_config->max_parallel_connections = getDefine_Int_default("remoted", "max_parallel_connections", 1, 65536, 512);
+    // max_deferred_requests caps requests parked awaiting a downstream service (503 over it).
+    // No Retry-After is sent: the agent runs its own retry/backoff on a 503.
     rm_config->max_deferred_requests = getDefine_Int_default("remoted", "max_deferred_requests", 1, 65536, 256);
 
     // Downstream (async UDS client to the engine's event ingress) tunables.
@@ -314,25 +324,6 @@ STATIC void w_remoted_build_module_config(const remoted *logr, remoted_module_co
     // back to its own default.
     rm_config->port = logr->https.port;
     remoted_module_https_config(rm_config);
-    // rm_config->max_inflight_bytes caps the HTTPS server's in-flight (unprocessed)
-    // request payload before it sheds load with 503. NOT logr->queue_size (that is an
-    // event COUNT, not bytes).
-    rm_config->max_inflight_bytes = (256 * 1024 * 1024); // 256 MiB
-    // rm_config->max_parallel_connections caps simultaneous HTTPS connections, bounding the
-    // read-phase memory peak (~ max_parallel_connections * max body size).
-    //
-    // It is also the ONLY bound on concurrent streamed responses (POST /download): chunked output
-    // rearms http_write_timeout per chunk, so a client that keeps reading slowly can hold a
-    // transfer open indefinitely, and there is no per-stream limiter. A mass upgrade -- the whole
-    // fleet fetching a WPK at once, many over slow links -- is therefore bounded only by this
-    // value, which is why it is settable rather than fixed.
-    rm_config->max_parallel_connections =
-        getDefine_Int_default("remoted", "http_max_parallel_connections", 1, 8192, 512);
-    // rm_config->max_deferred_requests caps requests parked awaiting a downstream service (503
-    // over it). No Retry-After is sent: the agent runs its own retry/backoff on a 503.
-    rm_config->max_deferred_requests = 256;
-    // rm_config->keystore_refresh_interval governs how often the C++ Keystore checks
-    // client.keys for changes (hot-reload)
     rm_config->keystore_refresh_interval = keyupdate_interval;
     rm_config->worker_node = logr->worker_node;
     rm_config->verification_mode = logr->https.verification_mode;
--- a/docs/ref/modules/remoted/configuration.md
+++ b/docs/ref/modules/remoted/configuration.md
@@ -571,6 +571,28 @@ Socket read buffer size for the HTTPS agent server, in bytes.
 - **Default value:** `8192`
 - **Allowed values:** Integer from `1` to `1048576` (1 MiB)
 
+#### remoted.http_max_parallel_connections
+
+Maximum simultaneous HTTPS connections.
+
+- **Default value:** `512`
+- **Allowed values:** Integer from `1` to `8192`
+- **Note:** Also the only bound on concurrent streamed responses (`POST /download`). Chunked
+  output rearms `remoted.http_write_timeout` per chunk, so a slow-but-steady reader can hold a
+  transfer open indefinitely and there is no per-stream limiter. Lower this if a mass upgrade over
+  slow links needs capping.
+
+#### remoted.http_stream_chunk_size
+
+Bytes per chunk when streaming a response body (`POST /download`).
+
+- **Default value:** `65536` (64 KiB)
+- **Allowed values:** Integer from `4096` to `1048576` (1 MiB)
+- **Note:** Charged per *in-flight transfer*, so the worst case is roughly this value times the
+  number of simultaneous downloads. A larger chunk buys fewer read/write round trips (less CPU per
+  byte) at the cost of more memory while transfers are running. It does not change the bytes
+  delivered -- only how they are framed on the wire.
+
 > **The three timeouts below are sequential phases of one request, and the sum matters.**
 > `remoted.http_request_timeout` bounds the *whole* request and its clock starts before the
 > downstream call, so if `connect + write + response` exceeds it, the HTTP server tears the request
--- a/docs/ref/modules/remoted/configuration.md
+++ b/docs/ref/modules/remoted/configuration.md
@@ -571,16 +571,36 @@ Socket read buffer size for the HTTPS agent server, in bytes.
 - **Default value:** `8192`
 - **Allowed values:** Integer from `1` to `1048576` (1 MiB)
 
-#### remoted.http_max_parallel_connections
+#### remoted.max_inflight_bytes
+
+Maximum in-flight (unprocessed) request payload bytes before the HTTPS server sheds load with HTTP 503.
+
+- **Default value:** `268435456` (256 MiB)
+- **Allowed values:** Integer from `1048576` (1 MiB) to `1073741824` (1 GiB)
+- **Note:** The C++ side clamps this up to at least one max-size request at startup, so a too-small
+  value cannot reject everything. This is NOT `legacy.queue_size` (that is an event COUNT, not bytes).
+
+#### remoted.max_parallel_connections
 
 Maximum simultaneous HTTPS connections.
 
 - **Default value:** `512`
-- **Allowed values:** Integer from `1` to `8192`
-- **Note:** Also the only bound on concurrent streamed responses (`POST /download`). Chunked
-  output rearms `remoted.http_write_timeout` per chunk, so a slow-but-steady reader can hold a
-  transfer open indefinitely and there is no per-stream limiter. Lower this if a mass upgrade over
-  slow links needs capping.
+- **Allowed values:** Integer from `1` to `65536`
+- **Note:** Bounds the read-phase memory peak (~`max_parallel_connections` × `max_body_size`). Also
+  the only bound on concurrent streamed responses (`POST /download`): chunked output rearms
+  `remoted.http_write_timeout` per chunk, so a slow-but-steady reader can hold a transfer open
+  indefinitely and there is no per-stream limiter. A mass upgrade (the whole fleet fetching a WPK
+  at once, many over slow links) is therefore bounded only by this value.
+
+#### remoted.max_deferred_requests
+
+Maximum requests parked awaiting a downstream service before replying with HTTP 503.
+
+- **Default value:** `256`
+- **Allowed values:** Integer from `1` to `65536`
+- **Note:** No `Retry-After` header is sent; the agent runs its own retry/backoff on a 503. If you
+  see warnings about this limit being reached, consider increasing it or investigating why the
+  downstream service is slow.
 
 #### remoted.http_stream_chunk_size
 
--- a/src/remoted/remoted_module/README.md
+++ b/src/remoted/remoted_module/README.md
@@ -160,6 +160,7 @@ src/endpoints/
 ├── statelessEndpoint.hpp/.cpp # /stateless policy: identity check + downstream target + post-processing
 ├── statsEndpoint.hpp/.cpp    # /stats policy (DUMMY): forwards to modulesd's inventory sync server
 └── configEndpoint.hpp/.cpp  # /config policy (DUMMY): near-duplicate of statsEndpoint, on purpose
+└── downloadEndpoint.hpp/.cpp  # /download policy: request grammar + resource resolution + file streaming
 ```
 
 - **Endpoint handler (async):**
@@ -489,6 +490,48 @@ and clients, which are all thread-safe internally.
 3. **Shutdown**: `RemotedModuleFacade::stop()` stops the HTTP server (drains in-flight requests),
    then stops the wazuh-db and task-manager clients (drains their queues), then destroys the
    registry (eviction thread stops automatically on destruction).
+
+## Streamed responses — `POST /download`
+
+Most endpoints answer with one in-memory body. `/download` serves `merged.mg` and WPK packages,
+which can be hundreds of megabytes, so it streams with **HTTP chunked transfer encoding** (64 KiB
+chunks) and memory that does not grow with file size.
+
+- **Declared at registration, not per response.** The transport creates a request's response builder
+  when the request is dispatched — that is what lets it release the received body before the request
+  is queued — and a builder's output mode is fixed at creation. So a streaming route registers with
+  `ResponseMode::Streamable`; its responder then keeps the request handle and builds *buffered* for
+  an error (a 404 is a normal response, not a chunked one) or *chunked* for a transfer. The cost is
+  that the received request survives into the worker queue, which is why it is opt-in per route.
+  `IHttpResponder::stream()` has a fail-loud default (answer `500`) so a `Buffered` mis-registration
+  is obvious instead of silently buffering a large body.
+- **Pull, not push.** The handler hands the transport an `IByteSource`; the transport drives the
+  loop, because only it owns both the connection's I/O strand and the worker pool. Read a chunk on a
+  worker, write it, and read the next only once the write completes. One chunk is resident, and a
+  worker slot is held only for the duration of a single read — never for the whole transfer.
+  Measured on a live manager: 2 GiB streamed for a 0.6 MiB RSS delta.
+- **Failure is truncation, not a short success.** A read error propagates out of `read()` and the
+  builder is dropped *without* `done()`, so the terminating `0\r\n\r\n` is never sent and the agent
+  sees an incomplete body it will retry. Returning 0 instead would emit the terminator and hand over
+  a truncated file that looks complete. RESTinio always fires the write callback (with
+  `write_was_not_executed` if the connection died first), so an aborted transfer always releases its
+  descriptor — verified: 25 mid-transfer disconnects left the fd count unchanged.
+- **`resource_id` is what the agent asks for, and the manager serves exactly that.** A `config`
+  request names either one group (`etc/shared/<group>/merged.mg`) or several, comma-separated —
+  wazuh's own multigroup form — which resolves to `var/multigroups/<sha256(resource_id)[:8]>/merged.mg`.
+  A `wpk` request names a filename and gets `var/upgrade/<filename>`. The multigroup form is what
+  lets an agent in several groups fetch its *effective* configuration rather than one member
+  group's, and it needs no database: the selector is hashed exactly as wazuh-db names the directory.
+  There is no group lookup and no membership check (protocol decision on #38022), so **any
+  authenticated agent can fetch any group's or multigroup's merged configuration**. `/control` must
+  report `config_hash` over the file this resolves to for the selector it hands the agent, or that
+  agent re-downloads on every notify.
+- **Containment differs per form.** The multigroup selector is *hashed, never joined*, so it cannot
+  traverse by construction. The single-group and WPK forms **do** join agent input into a path, so
+  there the grammars are the boundary: no `/`, not `.` or `..`, no leading dot, and for a WPK a
+  `.wpk` suffix. With no separator admitted, the joined path has exactly one component below the
+  base directory, and `openRegularFile()`'s `O_NOFOLLOW` stops that component being a symlink.
+  Loosening either grammar would break that and require a `realpath()` containment check instead.
 
 ## Deferred forwarding (async UDS) — `src/downstream/`
 
@@ -850,7 +893,7 @@ skipped without blocking the rest of the file).
 ctest --test-dir <build> -R remoted_module_utest -V
 ```
 
-### Manual / end-to-end (`tools/send_stateless.py`)
+### Manual / end-to-end (`tools/send_stateless.py`, `tools/send_download.py`)
 
 Signs and sends `POST /stateless` requests exactly as `AuthMiddleware` expects (AES-CMAC over the
 canonical byte sequence, agent key read straight from `client.keys`). Requires
--- a/src/remoted/remoted_module/include/remoted_module.h
+++ b/src/remoted/remoted_module/include/remoted_module.h
@@ -114,6 +114,12 @@ extern "C"
                                          ///< default (see remoted.http_concurrent_accepts).
         int http_buffer_size;            ///< Socket read buffer size, bytes. <=0 -> module default
                                          ///< (see remoted.http_buffer_size).
+        int http_stream_chunk_size;      ///< Bytes per chunk when streaming a response body
+                                         ///< (POST /download). <=0 -> module default, 64 KiB
+                                         ///< (see remoted.http_stream_chunk_size). Larger chunks
+                                         ///< trade memory per in-flight transfer for fewer
+                                         ///< read/write round trips, which is where the CPU per
+                                         ///< byte goes.
         long long max_inflight_bytes;    ///< Max in-flight request payload bytes; 503 over it (<=0 -> module default).
         int max_parallel_connections;    ///< HTTPS max simultaneous connections (<=0 -> module default).
         int max_deferred_requests; ///< Max requests parked awaiting a downstream service; 503 over it (<=0 -> default).
--- a/src/remoted/remoted_module/src/endpoints/authGateway.cpp
+++ b/src/remoted/remoted_module/src/endpoints/authGateway.cpp
@@ -108,7 +108,8 @@ namespace remoted::endpoints
     void AuthGateway::addAuthenticatedRoute(remoted::http::IHttpServer& server,
                                             Method method,
                                             const std::string& path,
-                                            AuthenticatedHandler handler)
+                                            AuthenticatedHandler handler,
+                                            remoted::http::ResponseMode mode)
     {
         auto middleware = m_middleware;
         const char* methodStr = methodToCanonical(method);
@@ -198,7 +199,9 @@ namespace remoted::endpoints
                     LOGFN_ERROR(logFn(), "Auth pipeline threw a non-standard exception.");
                     sendInternalErrorNoThrow(responder);
                 }
-            });
+            },
+            /*countAgainstBudget=*/true,
+            mode);
     }
 
 } // namespace remoted::endpoints
--- a/src/remoted/remoted_module/src/endpoints/authGateway.hpp
+++ b/src/remoted/remoted_module/src/endpoints/authGateway.hpp
@@ -55,11 +55,15 @@ namespace remoted::endpoints
          * @param method  HTTP method to match.
          * @param path    Path to match (the query string is not matched, but IS part of the MAC).
          * @param handler Invoked only after authentication succeeds; owns sending the response.
+         * @param mode    Whether the handler may answer with a streamed body. Forwarded verbatim to
+         *                IHttpServer::addRoute(): the transport fixes a response's output mode when
+         *                the request is dispatched, so a route that streams must declare it here.
          */
         void addAuthenticatedRoute(remoted::http::IHttpServer& server,
                                    Method method,
                                    const std::string& path,
-                                   AuthenticatedHandler handler);
+                                   AuthenticatedHandler handler,
+                                   remoted::http::ResponseMode mode = remoted::http::ResponseMode::Buffered);
 
     private:
         std::shared_ptr<remoted::auth::AuthMiddleware> m_middleware;
--- a/src/remoted/remoted_module/src/endpoints/downloadEndpoint.cpp
+++ b/src/remoted/remoted_module/src/endpoints/downloadEndpoint.cpp
@@ -314,7 +314,7 @@ namespace remoted::endpoints::download
         std::string name;
         name.reserve(MULTIGROUP_HASH_HEX_CHARS);
 
-        for (unsigned int i = 0; i < digestLength && name.size() < MULTIGROUP_HASH_HEX_CHARS; ++i)
+        for (unsigned int i = 0; i < MULTIGROUP_HASH_HEX_CHARS / 2 && i < digestLength; ++i)
         {
             name.push_back(HEX_DIGITS[(digest[i] >> 4) & 0x0F]);
             name.push_back(HEX_DIGITS[digest[i] & 0x0F]);
--- a/src/remoted/remoted_module/src/endpoints/downloadEndpoint.cpp
+++ b/src/remoted/remoted_module/src/endpoints/downloadEndpoint.cpp
@@ -353,6 +353,20 @@ namespace remoted::endpoints::download
         : m_fd {fd}
         , m_size {size}
     {
+        // Baseline for the mid-transfer modification check in read(). Best-effort on purpose: a
+        // descriptor we cannot stat is not a reason to refuse to serve a file we already opened
+        // successfully. m_mtimeKnown is what makes that true -- without it, a failed fstat would
+        // leave the baseline at zero, every comparison would mismatch, and EVERY transfer would
+        // abort. The byte-count half of the check stands on its own either way.
+        struct stat info
+        {
+        };
+        if (::fstat(m_fd, &info) == 0)
+        {
+            m_mtimeSec = static_cast<std::int64_t>(info.st_mtim.tv_sec);
+            m_mtimeNsec = static_cast<std::int64_t>(info.st_mtim.tv_nsec);
+            m_mtimeKnown = true;
+        }
     }
 
     FileByteSource::~FileByteSource()
@@ -388,7 +402,71 @@ namespace remoted::endpoints::download
             throw std::system_error {errno, std::generic_category(), "read() failed while streaming a resource"};
         }
 
+        if (bytesRead == 0)
+        {
+            // End of stream. The per-chunk check below has already covered everything up to the
+            // last data read; this catches a writer that landed between that read and this one.
+            if (m_delivered != m_size)
+            {
+                // Short by construction: end-of-file arrived before the file's own length.
+                throw std::runtime_error {"the resource was truncated while it was being streamed"};
+            }
+            checkNotModified();
+            return 0;
+        }
+
+        m_delivered += static_cast<std::uint64_t>(bytesRead);
+
+        // More readable than the file held at open(): the bytes already sent belong to a version
+        // that no longer exists. Cheap enough to keep as its own invariant -- it needs no syscall.
+        if (m_delivered > m_size)
+        {
+            throw std::runtime_error {"the resource grew while it was being streamed"};
+        }
+
+        // Checked per chunk, not only at end-of-stream. Same criterion either way, but a writer
+        // that lands one second into a 1 GiB transfer would otherwise cost the full gigabyte of
+        // reads and socket writes before the transfer is abandoned; here it costs one more chunk.
+        // The price is an fstat() per chunk -- no path resolution, no I/O, against a read() plus a
+        // socket write that are already happening.
+        checkNotModified();
+
         return static_cast<std::size_t>(bytesRead);
+    }
+
+    void FileByteSource::checkNotModified() const
+    {
+        struct stat info
+        {
+        };
+        if (::fstat(m_fd, &info) != 0)
+        {
+            throw std::system_error {errno, std::generic_category(), "fstat() failed while streaming a resource"};
+        }
+
+        // Classified rather than lumped together, so the abort WARN names what actually happened:
+        // a shorter file is a truncating rewrite (c_group() on merged.mg), a longer one is a file
+        // still being staged (a WPK mid-download), and an unchanged length with a moved mtime is a
+        // same-size rewrite -- the case no byte count can see.
+        const auto currentSize = static_cast<std::uint64_t>(info.st_size);
+
+        if (currentSize < m_size)
+        {
+            throw std::runtime_error {"the resource was truncated while it was being streamed"};
+        }
+
+        if (currentSize > m_size)
+        {
+            throw std::runtime_error {"the resource grew while it was being streamed"};
+        }
+
+        // Same length: only the baseline mtime distinguishes it, and only when we recorded one.
+        // Without a baseline the check is skipped rather than assumed to have failed.
+        if (m_mtimeKnown && (static_cast<std::int64_t>(info.st_mtim.tv_sec) != m_mtimeSec ||
+                             static_cast<std::int64_t>(info.st_mtim.tv_nsec) != m_mtimeNsec))
+        {
+            throw std::runtime_error {"the resource was modified while it was being streamed"};
+        }
     }
 
     std::uint64_t FileByteSource::size() const noexcept
--- a/src/remoted/remoted_module/src/endpoints/downloadEndpoint.cpp
+++ b/src/remoted/remoted_module/src/endpoints/downloadEndpoint.cpp
@@ -1,0 +1,517 @@
+/*
+ * Wazuh remoted module (C++ worker bridge)
+ * Copyright (C) 2015, Wazuh Inc.
+ * July 29, 2026.
+ *
+ * This program is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU General Public
+ * License (version 2) as published by the FSF - Free Software
+ * Foundation.
+ */
+
+#include "downloadEndpoint.hpp"
+
+#include "loggerHelper.h"
+
+#include <rapidjson/document.h>
+
+#include <openssl/evp.h>
+
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include <cerrno>
+#include <cstdlib>
+#include <cstring>
+#include <stdexcept>
+#include <system_error>
+#include <utility>
+
+namespace
+{
+    constexpr auto DOWNLOAD_LOGTAG {"wazuh-manager-remoted:endpoints"};
+
+    const LogFn& logFn()
+    {
+        static const LogFn instance {LogFn {DOWNLOAD_LOGTAG}.compose("download")};
+        return instance;
+    }
+
+    /// Longest accepted resource id. MAX_GROUP_NAME for a group; also a sane bound for a filename.
+    constexpr std::size_t MAX_RESOURCE_ID_SIZE {255};
+
+    /// Wire spellings of ResourceType.
+    constexpr auto RESOURCE_TYPE_CONFIG {"config"};
+    constexpr auto RESOURCE_TYPE_WPK {"wpk"};
+    constexpr auto WPK_EXTENSION {".wpk"};
+
+    /// Separator wazuh uses between groups in a multigroup selector (MULTIGROUP_SEPARATOR).
+    constexpr char GROUP_SEPARATOR {','};
+
+    /**
+     * @brief Hex characters in a multigroup directory name.
+     *
+     * Must equal wazuh-db's WDB_GROUP_HASH_SIZE (8). Note that is 8 HEX CHARACTERS -- the first
+     * four digest bytes -- not the first eight bytes.
+     */
+    constexpr std::size_t MULTIGROUP_HASH_HEX_CHARS {8};
+
+    /// Cap on a whole multigroup selector, mirroring wazuh's own multigroup length allowance.
+    constexpr std::size_t MAX_MULTIGROUP_SELECTOR_SIZE {4096};
+
+    /// Basename of a merged centralized configuration (SHAREDCFG_FILENAME).
+    constexpr auto MERGED_CONFIG_FILENAME {"merged.mg"};
+
+    /// Characters wazuh accepts in a single group name (w_validate_group_name's set, minus comma).
+    bool isGroupNameChar(char c)
+    {
+        return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '.' || c == ':' ||
+               c == ';' || c == '_' || c == '-' || c == '=' || c == '+' || c == '!' || c == '@' || c == '(' || c == ')';
+    }
+
+    /// Stricter set for a value that gets joined into a path.
+    bool isWpkNameChar(char c)
+    {
+        return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '.' || c == '_' ||
+               c == '-';
+    }
+
+    bool isDotOrDotDot(std::string_view value)
+    {
+        return value == "." || value == "..";
+    }
+
+    bool endsWith(std::string_view value, std::string_view suffix)
+    {
+        return value.size() >= suffix.size() && value.compare(value.size() - suffix.size(), suffix.size(), suffix) == 0;
+    }
+
+    std::string joinPath(const std::string& directory, const std::string& name)
+    {
+        if (directory.empty())
+        {
+            return name;
+        }
+        return directory.back() == '/' ? directory + name : directory + '/' + name;
+    }
+
+    /// Reads a required string member, rejecting a missing member or a non-string value.
+    const rapidjson::Value* stringMember(const rapidjson::Document& doc, const char* name)
+    {
+        const auto member = doc.FindMember(name);
+        if (member == doc.MemberEnd() || !member->value.IsString())
+        {
+            return nullptr;
+        }
+        return &member->value;
+    }
+
+    const char* resourceTypeName(remoted::endpoints::download::ResourceType type)
+    {
+        return type == remoted::endpoints::download::ResourceType::Config ? RESOURCE_TYPE_CONFIG : RESOURCE_TYPE_WPK;
+    }
+
+    /// errno -> the error a failed open maps to. Absent or wrong-typed is a plain 404; anything
+    /// else is ours, not the agent's, so it surfaces as a 500 and gets logged.
+    remoted::endpoints::download::LocateError errorForOpenErrno(int openErrno)
+    {
+        using remoted::endpoints::download::LocateError;
+        switch (openErrno)
+        {
+            case ENOENT:
+            case ENOTDIR:
+            case ELOOP:  // O_NOFOLLOW rejected a symlink: indistinguishable from absent, to the agent.
+            case EINVAL: // openRegularFile()'s marker for "not a regular file".
+                return LocateError::NotFound;
+            default: return LocateError::Internal;
+        }
+    }
+} // namespace
+
+namespace remoted::endpoints::download
+{
+
+    // -----------------------------------------------------------------------
+    // Request parsing and grammars
+    // -----------------------------------------------------------------------
+
+    bool isValidGroupName(std::string_view group)
+    {
+        if (group.empty() || group.size() > MAX_RESOURCE_ID_SIZE || isDotOrDotDot(group))
+        {
+            return false;
+        }
+
+        for (const char c : group)
+        {
+            if (!isGroupNameChar(c))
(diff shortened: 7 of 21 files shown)


src/remoted/src/secure.c
```diff
@@ -310,6 +314,25 @@ STATIC void w_remoted_build_module_config(const remoted *logr, remoted_module_co
...
+    rm_config->max_inflight_bytes = (256 * 1024 * 1024); // 256 MiB
+    // rm_config->max_parallel_connections caps simultaneous HTTPS connections, bounding the
+    // read-phase memory peak (~ max_parallel_connections * max body size).
+    //
+    // It is also the ONLY bound on concurrent streamed responses (POST /download): chunked output
+    // rearms http_write_timeout per chunk, so a client that keeps reading slowly can hold a
+    // transfer open indefinitely, and there is no per-stream limiter. A mass upgrade -- the whole
+    // fleet fetching a WPK at once, many over slow links -- is therefore bounded only by this
+    // value, which is why it is settable rather than fixed.
+    rm_config->max_parallel_connections =
```
TomasTurina: Repeated

src/remoted/src/secure.c
```diff
@@ -310,6 +314,25 @@ STATIC void w_remoted_build_module_config(const remoted *logr, remoted_module_co
...
+        getDefine_Int_default("remoted", "http_max_parallel_connections", 1, 8192, 512);
+    // rm_config->max_deferred_requests caps requests parked awaiting a downstream service (503
+    // over it). No Retry-After is sent: the agent runs its own retry/backoff on a 503.
+    rm_config->max_deferred_requests = 256;
```
TomasTurina: Repeated

src/remoted/src/secure.c
```diff
@@ -310,6 +314,25 @@ STATIC void w_remoted_build_module_config(const remoted *logr, remoted_module_co
     // back to its own default.
     rm_config->port = logr->https.port;
     remoted_module_https_config(rm_config);
+    // rm_config->max_inflight_bytes caps the HTTPS server's in-flight (unprocessed)
+    // request payload before it sheds load with 503. NOT logr->queue_size (that is an
+    // event COUNT, not bytes).
+    rm_config->max_inflight_bytes = (256 * 1024 * 1024); // 256 MiB
```
TomasTurina: Repeated

TomasTurina (approved):

PR #38087, opened 2026-07-29
Restart HTTPS client after initial enrollment and mutex-lock global https reference

MAnDumu:
## Description

Fixes two defects in the agent's HTTPS client found during live QA against #37831.

An agent that boots with no `client.keys` has no signing key when the HTTPS client is started at agentd startup, so the transport refuses to come up. Nothing re-drove it once enrollment produced a key moments later, leaving the agent with no transport for the rest of the process — it never reached the manager until the next restart. Separately, the client handle was published at startup and cleared at shutdown without the lock that every reader of it takes, so a module thread could observe it half-assigned or already being destroyed.

Related: #37831, #38079.

## Proposed Changes

- The agent now brings the HTTPS transport up as soon as the initial enrollment provides a signing key, instead of leaving it down until the next restart. The retry is shared by the Linux and Windows startups.
- Booting without a key is no longer reported as an error. It is an expected step that precedes enrollment, so it is logged at debug level.

- The https client handle is now write-protected by a mutex.

## Results and Evidence

Live fresh boot with an empty `client.keys`, on a real agent built from this branch. The first start attempt defers, enrollment obtains a key, and the transport is started again as soon as the key is reloaded:

```sql
2026/08/05 15:15:48 wazuh-agentd[99478] keys.c:338 at OS_ReadKeys(): DEBUG: (1751): File client.keys not found or empty.
2026/08/05 15:15:48 wazuh-agentd[99478] https_client_bridge.c:1517 at w_https_client_start(): INFO: https_client: starting.
2026/08/05 15:15:48 wazuh-agentd[99478] enrollment_op.c:128 at w_enrollment_request_key(): INFO: Requesting a key from server: 192.168.56.10
2026/08/05 15:15:48 wazuh-agentd[99478] enrollment_op.c:468 at w_enrollment_process_agent_key(): INFO: Valid key received
2026/08/05 15:16:08 wazuh-agentd[99478] keys.c:444 at OS_UpdateKeys(): INFO: (1410): Reading authentication keys file.
2026/08/05 15:16:08 wazuh-agentd[99478] https_client_bridge.c:1517 at w_https_client_start(): INFO: https_client: starting.
```

Before this change the second `https_client: starting.` line never appeared and the agent stayed silent for the rest of the process.

Tested on both ubuntu and windows.

## Artifacts Affected

- `wazuh-agentd` — behavior change only, no interface or ABI change.
[...]

Issue #38079, opened 2026-07-29
Refactor failing ITs for HTTPS protocol and the new RemotedSimulator


Nicogp:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0 | Integration tests (`tests/integration`) + QA patterns | Agent | Sources | Linux/windows/macOS |

> Part of the agent-side HTTPS migration objective #37702 and epic #37831.
> Depends on qa-framework simulator work #788 and issue 09 from this breakdown.

## Description

As the agent communication path moves from legacy Blowfish/AES TCP/UDP to HTTPS, the
existing IT suites in `wazuh/wazuh` are progressively failing because they still assume
legacy behavior:

- they wait for legacy control/log messages that are no longer emitted in the new flow,
- they are coupled to the previous remoted simulator behavior,
- and they assert patterns that no longer match current module logs and endpoint behavior.

`wazuh/qa-integration-framework` issue #788 changes `RemotedSimulator` to model the new
HTTPS protocol. Once that lands, integration tests must be refactored to use the new
simulator semantics and to validate protocol outcomes that are still meaningful under the
new model (HTTP status, endpoint payloads, state transitions, retries), instead of relying
on removed log strings.

In parallel, module `patterns` in `wazuh/qa-integration-framework` must be updated where
needed so tests can assert expected runtime behavior without false negatives caused by old
log signatures.

This is a rolling stabilization issue: fix failing suites incrementally as epic PRs land,
until the HTTPS target IT matrix is green.

## Goals

- Adapt integration tests to the HTTPS protocol model and the new `RemotedSimulator`.
- Remove or replace assertions based on legacy messages that are intentionally gone.
- Update QA framework module `patterns` to the current expected logs/events.
- Recover CI signal quality: tests should fail only on real regressions, not outdated
  assumptions.
- Keep the migration incremental and traceable (suite by suite, failure by failure).

### In scope

- Refactor failing tests in `wazuh/wazuh/tests/integration` impacted by HTTPS migration.
- Replace legacy waits/assertions with stable checks aligned to new behavior, for example:
  - `/control` lifecycle outcomes (startup/notify/shutdown semantics),
  - task dispatch and fire-and-forget behavior,
  - `/stateless` and `/stateful` processing expectations,
  - `/config` and `/stats` push behavior,
  - `/download`-driven config/WPK flows where applicable.
- Update affected `patterns` in `wazuh/qa-integration-framework` modules so expected
[...]

PR #38088, opened 2026-07-29
test(agentd): migrate agentd/execd integration tests to the HTTPS agent protocol

MAnDumu:
## Description

Migrates `tests/integration/test_agentd/*` and `test_execd`'s active-response test to the new HTTPS agent-manager control path (#37831), against the `RemotedSimulator`/`AuthdSimulator` rewrite (wazuh/qa-integration-framework#795, patterns added in qa-integration-framework#800). The files touched by this branch were run live against a real compiled `wazuh-agentd` and the new simulators, not just read for correctness. Scope broadened along the way to include a handful of standalone fixes in `test_sca`/`test_syscollector`/`test_logcollector` that were blocking the same HTTPS handshake path (see below) — none of them depend on another unmerged branch.

Related: #37831, #38079.

## Proposed Changes

**Mechanical adaptation to the new simulator/protocol** (most files):
- Drop `RemotedSimulator(protocol='tcp')` — the simulator is HTTPS-only now.
- `AuthdSimulator`/`add_client_keys_entry` now take a real 32-hex-char AES-CMAC secret instead of an arbitrary string, since the HTTPS client validates it.
- `mode='WRONG_KEY'` → `mode='REJECT_AUTH'` (simulator mode renamed upstream).
- `INVALID_MSG` → `BAD_REQUEST` in `cases_reconnection_protocol.yaml`.
- `<ssl><verification_mode>none</verification_mode></ssl>` added to `wazuh_conf.yaml` templates across `test_agentd`, `test_sca`, `test_syscollector`, and `test_logcollector` — the HTTPS client refuses to start without a CA unless verification is disabled, and the simulator uses a self-signed cert. The default `AuthdSimulator` secret (`'SuperSecretKey'`) also isn't valid hex, so `bridge_build_config()`'s AES-CMAC key validation was rejecting it — fixed to a real 32-hex-char secret in the same pass.
- `wait_connect()` / connection-log assertions now also accept `AGENTD_HTTPS_STARTUP_ACCEPTED` alongside the legacy `AGENTD_CONNECTED_TO_SERVER`.
- `conftest.py`'s `ensure_merged_mg` fixture updated for the simulator's `DEFAULT_MERGED_MG` module constant, replacing a removed class attribute (this went through one self-correction: an intermediate commit reverted to the simulator's `DEFAULT_MERGED_MG_CONTENT` after that symbol appeared missing against a stale checkout of the framework's dependency branch; once the framework branch was synced to the real HTTPS `RemotedSimulator`, the original module-level import was confirmed correct and restored).

**Real bugs found in the test code itself while running these live:**
- `test_agentd_reconection_enrollment_with_keys.py`: fixed a pre-existing bug where an early failure before `remoted_server`/`authd_server` were assigned made the `finally` block reference undefined names; both are now initialized to `None` before the `try`.

**Removed, then restored once the real product fix landed:** `test_agentd_initial_enrollment_retries.py`, `test_agentd_reconection_enrollment_no_keys.py`, `test_agentd_reconection_enrollment_no_keys_file.py`. All three start the agent without `client.keys`, drive it through initial enrollment, and then assert the HTTPS client comes up afterwards — that only worked once `w_https_client_start()` was retried post-enrollment. That product fix (originally tracked as #38087, superseded by #38208) is now merged into `5.0.0-https` (`start_agent_prepare()` calls the blocking `w_agentd_keys_init()` before `w_https_client_start()` in `AgentdStart()`, `agentd.c`), so all 3 files were restored bit-identical to their pre-removal state, confirmed by both reviewers against an empty diff.

**Update — the 3 areas below were previously left `@pytest.mark.skip`'d (feature not implemented / deprecated at the time), then re-checked against the current `5.0.0-https` instead of assumed still-blocked. Each resolved differently:**
- `test_agentd_parametrized_reconnections.py`: **deleted**, not just left skipped. `max_retries`/`retry_interval` are parsed only to log a deprecation warning and always fall back to their defaults (`Read_Client_Server`, `client-config.c`) — the retry-count behavior under test no longer exists, and there's nothing left to adapt it against.
- `test_block_ip.py` (`test_execd`): **un-skipped on Linux and Windows.** `bridge_on_task()`/`bridge_dispatch_active_response()` (`https_client_bridge.c`) are no longer stubs — they really dispatch `active_response` tasks to `execd` over its Unix socket now. `send_execd_message` (`test_execd/conftest.py`) was rewritten to deliver the task via `RemotedSimulator.add_task()` instead of the removed `send_custom_message()`; a payload missing the `"wazuh"` key entirely is now rejected agent-side before ever reaching `execd`, so the fixture and that case's `expected_error` were updated to match the new rejection point. A separate, module-level `pytest.mark.skipif(sys.platform == WINDOWS, ...)` (dated 2026-03-21, predating this migration entirely, no surviving rationale in its own commit) was also removed once the same rejection-point fix was applied to `cases_execd_block_ip_win.yaml` — confirmed both platform-agnostic in the validating C code (`bridge_dispatch_active_response()`/`ExecdRun()` run their format checks before any `#ifdef WIN32` branch) and passing 4/4 in real Windows CI, including the one case (`Command execution succeed`) that spawns `block-ip-windows.c` against the real Windows firewall.
- `test_agentd_state.py`: **un-skipped the 2 `type == 'remote'` cases.** The legacy synchronous `'#!-req agent getstate'` has no equivalent, but the same state document (`w_agentd_state_get()`) now rides the periodic `/stats` push (#37843) — read via `wait_for_custom_message_response()` polling `RemotedSimulator.last_stats` instead of sending a request. This went through a real, multi-round root-cause chase: the branch's base (`5.0.0-https`) had moved past a commit (`wazuh/wazuh@72ace94e97`, "report agent statistics in the shape the manager indexes") that reshaped that push after this branch's prior rebase point — `modules` went from an array of `{"module":, "stats":}` entries to an object keyed by module name (`{"modules": {"agent": {...}, "logcollector": {...}}}`), counters moved under `messages.count`/`tasks.*.total`, and `last_ack` was dropped from the payload entirely. Confirmed empirically (not just by reading code) by rebuilding the agent, overlaying it on a real install, and capturing the raw `/stats` request bodies `RemotedSimulator` actually received. `last_ack` is now also unchecked in the *file*-based case: `UPDATE_ACK` has no caller left anywhere in `client-agent/src`, so that field is dead in the whole product, not only absent from this one push — checking it via the `.state` file was passing vacuously (an always-empty string still satisfies `is not None`). Unblocking these cases (previously the whole test case was skipped, not just the remote check) also exposed 3 more dead log-line preconditions (`wait_keepalive()`, `wait_ack()`, `wait_agent_notification()` — no caller left anywhere in `client-agent/src`) and a real, previously-unexercised timing race in `check_status()`'s file-reading path (a fixed ~10-15s budget, replaced with a 60s polling loop after it flaked in real CI). All 3 cases now pass in real CI on Linux and Windows.

**Deleted, not skipped — `test_multi_server/` (whole directory, 5 files):** confirmed with the product team that multi-server failover isn't a migration gap to revisit later, it's dropped by design (the agent only supports a single configured `<server>` going forward). Skip implies "come back to this"; delete is the honest call here, same reasoning as the deprecated-retry test above. One correction worth flagging on the old skip-reason wording it replaces: `w_agentd_keys_init()` (`start_agent.c`) still has a live loop retrying *enrollment* against every configured server in turn — what's actually gone is failover at the connection level, since `bridge_build_config()` only ever reads `agt->server[0]` regardless of which server that enrollment loop succeeded against, and `agent_handshake_to_server()` (the old per-connection failover) has no live caller left. The 7 deleted cases (`refuse_remoted_accept_enrollment`, `server_down_fallback_tcp_0/1`, `all_servers_unavailable`, etc.) exercised exactly that dead connection-level path.

**`test_startup_hash_gate.py`** (most substantially reworked — the startup-hash gate moved from a push+MD5 model in the legacy handshake to a pull+SHA256 model over `/control notify` + `/download`):
- `GATE_RELEASED_PATTERN` removed (its only assertion site is gone, see below); `GATE_BLOCKING_PATTERN` updated to the `waiting_handshake` reason string (superseded again below, after #38133).
- `_wait_startup_gate_status()` now accepts a set of acceptable reason strings, since `https_hash_match` and `https_config_applied` can both legitimately win the race to release the gate.
- New `_wait_process_running()` replaces one-shot `check_if_process_is_running()` asserts with polling — the HTTPS handshake can complete before a module daemon finishes forking, which isn't itself a failure, and the one-shot assert was observed to flake on that timing.
- Dropped the "gate released" log-line assertion in the `hash_match` scenario: the module only logs "released" if it first logged "blocking", and a fast-enough handshake can skip both lines without anything being wrong — this was verified to flake even with a 60s timeout.
- `_live_notify_response()` / `_use_live_config_hash()`: the delay-injection helper used by 5 of the 6 scenarios (`_make_config_available_after_delay`) used to set `.merged_mg` and `.config_hash` in two separate, unlocked statements while simulator HTTP threads could read them concurrently. The timer now only ever sets `.merged_mg` (a single, GIL-atomic assignment); an instance-level `notify_response()` override derives `config_hash` from `merged_mg` live on every call instead of reading a separately-set attribute. Not applied to the `hash_mismatch` scenario, which intentionally leaves `merged_mg=None` for its whole run and never touches it, so no race applies there.

**Update — `test_startup_hash_gate.py` again, after #38133:** #38133 ("Retire the legacy TCP data path from the agent") rewrote the gate to a single feed and changed its initial blocking reason from `waiting_handshake` to `waiting_config_hash` (`startup_gate.c`). This test still checked the old string in `GATE_BLOCKING_PATTERN` and 5 `_wait_startup_gate_status()` calls. The mismatch didn't fail cleanly: the poller retries every 0.5s for up to 90s looking for an exact match, so with a string that never appears it kept polling past the point the gate actually released, and only failed once the full 90s elapsed — reporting a confusing "Last status={'ready': True, ...}" for 4 of 5 affected scenarios. Confirmed with a live `tail -F` of the agent's log during one of these runs that the gate was correctly blocked with `waiting_config_hash` for the whole expected window, then released on schedule — the wrong string was the only problem. Also corrected two comments that named a mechanism #38133 retired (the module docstring's MD5 merged_sum description, and `reload_chain_fails_gate_still_releases`'s credit to `receiver.c`'s `receive_msg()` — deleted entirely; that fallback now lives in `bridge_on_config_downloaded()`, `https_client_bridge.c`).

**Diagnostic added to `wait_connect()` (`test_agentd/utils.py`):** `test_agentd_reconection_enrollment_with_keys[TCP protocol]` failed twice in a row in real Windows CI on its second `wait_connect()` call (after a full reject/re-enroll/reconnect cycle), with no local Windows environment to reproduce and each attempt costing a full CI cycle. Rather than guess at a root cause, `wait_connect()` now captures `ossec.log`'s byte size right before waiting and, on a miss, reads everything appended since then straight into the assert message — from the same process, at the moment of the actual timeout, rather than relying on the end-of-module CI diagnostic step (which runs after the next test's `truncate_monitored_files` has already reset the log, so it never saw the state that mattered). Anchored to a byte position rather than a fixed line count, since the debug2 level every caller enables can produce enough output over 150s to push real evidence out of a fixed-size tail. In the run immediately after this landed, the test passed cleanly (no repeat of the failure) — consistent with a timing flake rather than a deterministic bug; the diagnostic stays in place for if/when it recurs.

**Docs-only:** one-line comment added to `agentd.c` documenting the `needs_config_reload` SIGUSR1 flag (set by `reload_handler()`, polled by the main loop).

**Also fixed along the way, in `test_sca`/`test_syscollector`/`test_logcollector` (not this PR's original stated scope, but blocking the same startup-gate path and standalone — no dependency on another unmerged branch):**
- Windows CI diagnostics: added a step dumping `wazuh-agent` state on failure to `.github/workflows/5_testintegration_windows-integration-tests.yml`, and raised `AGENTD_WINDOWS_DEBUG` to `2` in the affected test's `local_internal_options` so `wait_connect()`'s HTTPS milestone log line is actually emitted (it had been left at `0`, which silently suppresses that specific `mdebug1()`).

## Open question, not resolved

The full `test_startup_hash_gate.py` 6-scenario suite passed 6/6 twice locally. One earlier local run had `reload_chain_fails_gate_still_releases` fail with `Could not read startup gate status after 90s: [Errno 2] No such file or directory` reading the gate-status socket — a sustained failure across the *entire* 90s retry window, not a brief race. It did not reproduce in further full-suite or isolated runs, and the log from that specific run wasn't preserved, so the root cause is genuinely unknown rather than confirmed as a transient flake. Flagging this here instead of asserting a cause that can't be backed up — worth watching if it recurs in CI.

### Results and Evidence

Run against a real compiled `wazuh-agentd`/`wazuh-execd`/`wazuh-logcollector` and the HTTPS `RemotedSimulator`/`AuthdSimulator` (qa-integration-framework#795), both locally (production build, `UNIT_TEST=OFF`, overlaid on a real install) and in real CI on this PR:

- `test_startup_hash_gate.py`: all 6 scenarios pass individually, and as a full suite (~150s total), run multiple times across review iterations.
- Full `test_agentd/` + `test_execd/` suite locally: 24 passed, 0 failed, 0 skipped, after resolving the 3 previously-skipped areas above.
- **Real CI, current push (`efd5682428`)**: `IT Linux - agentd` 20 passed / `IT Linux - execd` passed / `IT Windows - agentd` 14 passed, 6 deselected (1 initial flaky failure on `test_agentd_reconection_enrollment_with_keys`, confirmed not a regression — passed clean on the run where the `wait_connect()` diagnostic landed) / `IT Windows - execd` 4 passed (including the previously Windows-skipped `test_block_ip.py`) / `IT Linux/Windows - github` and `IT Linux/Windows - office365` both passed (these two had no skips to begin with, confirming there was nothing to resolve there). One `IT Linux - agentd` run failed on a GitHub Actions infrastructure timeout downloading the `actions/checkout` action, unrelated to any code here — confirmed by re-running just that job, which then passed.
- `test_sca`/`test_syscollector`/`test_logcollector`, local only, not part of this PR's pass/fail claim: `test_logcollector`'s `test_basic_configuration_command.py`/`test_basic_configuration_journald.py` at 10/10 and 12/12 respectively (also re-validated after wazuh/qa-integration-framework#812 landed a config-validation fix these tests needed); `sca`/`syscollector` still fail locally on unrelated, deeper product issues past the startup gate (`sca`: "Module limits not configured", looks like a real AGCOM IPC bug; `syscollector`: a timing-dependent `assert None` in specific callbacks) — reported here for visibility, not claimed as fixed by this PR.
- Non-blocking, pre-existing failures observed in the same CI runs, unrelated to any commit on this branch (confirmed by checking which files each touches): `IT Linux/Windows - sca`, `IT Linux/Windows - syscollector`, `IT Linux - logcollector`, `IT Linux - aws_inspector`, `IT Linux - enrollment` (a different, never-touched test module — 19 failures, same class of migration gap as `sca`/`syscollector`), `IT MacOS - fim`, `scan-build-linux (manager)`.

This change went through the mandatory 2-stage review (`code-config-reviewer` then `contrarian-reviewer`) on every commit, iterating until each was clean. Selected findings from later rounds (earlier rounds already summarized above where they changed the design):
- The `/stats` schema fix went through 2 full rounds: the first found under/over-logging in the defensive parsing; the second found that a docstring claim ("last_ack is still verified via the file block") was factually false once `UPDATE_ACK`'s caller was confirmed dead everywhere, not just under HTTPS — both fixed and re-verified.
- The Windows `test_block_ip.py` unskip: `contrarian-reviewer` pushed back on treating all 4 unskipped cases as equal risk, since only one (`Command execution succeed`) exercises the real Windows firewall and had 3 real platform-specific fixes land after the skip was added — the distinction is now recorded in that commit's message so a future red run doesn't have to re-derive it.
- The `wait_connect()` diagnostic went through 3 rounds (a fixed-size line tail was found insufficient given debug2 verbosity, then an encoding mismatch with `FileMonitor`, then an uncaught `ValueError`/overly narrow `except` clause) before landing on the final byte-position-anchored, `except Exception`-guarded version.

  - [x] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

### Artifacts Affected

Test-only change plus one product-code comment (`agentd.c`, docs-only, no behavior change) — no production executables, packages, or default configuration files affected. Test configuration templates (`wazuh_conf.yaml` across `test_agentd`, `test_sca`, `test_syscollector`, `test_logcollector`) and test case YAML files were updated as described above.

### Configuration Changes

N/A — no production configuration parameters added or changed. Test fixture configuration templates gained `<ssl><verification_mode>none</verification_mode></ssl>`, required only for these tests to run against the simulator's self-signed certificate.

### Tests Introduced

No new test files. Existing tests were adapted (mechanical protocol/pattern changes plus the `/stats` schema fix), 1 real test-code bug was fixed (undefined-name `finally` block in `test_agentd_reconection_enrollment_with_keys.py`), and 5 files were deleted outright: 2 (`test_agentd_parametrized_reconnections.py`, `test_multi_server/`) as intentionally-dropped features (see Proposed Changes above), while 3 more (`test_agentd_initial_enrollment_retries.py`, `test_agentd_reconection_enrollment_no_keys.py`, `test_agentd_reconection_enrollment_no_keys_file.py`) were removed and later restored bit-identical once the product fix they depend on merged (see "Removed, then restored" above). The 2 test files that were previously `@pytest.mark.skip`'d (`test_block_ip.py`, including its separate Windows-only skip; `test_agentd_state.py`) are now fully un-skipped and passing on both platforms in real CI — see the "Update" notes above.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

This PR pairs with `review/37831-agent-https-client` (originally #38087, superseded by #38208, wazuh/wazuh, already merged into `5.0.0-https`) and `fix/37831-agentd-https-log-patterns` (already merged as wazuh/qa-integration-framework#800) — all three came out of the same live-testing pass against #795 but are kept as independent, independently reviewable branches. This branch does not carry any change or test whose passing depends on another unmerged branch.

**Branch history note:** this branch has been reconstructed 5 times over the course of this work as `5.0.0-https` was repeatedly rewritten or moved far ahead (most recently: 432 new upstream commits between two rebases). Each reconstruction cherry-picks only this branch's own commits onto the current base tip, discarding any inherited history that isn't this PR's own work — the commit hashes referenced above (e.g. `efd5682428`) reflect the current, final reconstruction and will not match earlier review comments made against prior hashes for the same logical change.


MAnDumu:
**Update 2026/08/04**
Reworked test_startup_hash_gate.py for the new HTTPS startup gate model

**Update 2026/08/05**
Documented needs_config_reload flag in agentd.c
Fixed a stale RemotedSimulator constant reference in test fixtures

**Update 2026/08/06**
Corrected the RemotedSimulator constant fix after syncing the framework dependency
Raised Windows agent debug level so the connection log line is visible

**Update 2026/08/07**
Added Windows CI diagnostics dump on failure
Fixed an invalid AuthdSimulator secret blocking sca/syscollector/logcollector tests

**Update 2026/08/10**
Fixed TLS verification config blocking sca/syscollector/logcollector tests
Restored the 3 reconnection tests (product fix already merged)
Migrated config templates from <client> to <agent>
Resolved pending agentd (6) and execd (4) skips: deleted deprecated ones, adapted the rest
Full test_agentd/test_execd suite: 24/24 passing locally

**Update 2026/08/11**
Found and fixed the real root cause of the /stats schema mismatch failing in CI
Removed a stale Windows skip in test_block_ip — 4/4 passing in real CI, Linux and Windows
Added ossec.log diagnostic to investigate an intermittent Windows re-enrollment flake
Fixed qa-integration-framework #812: config validation now fails loud instead of silently skipping unknown sections
Migrated test_enrollment/ to HTTPS (4 root causes fixed) — 19 failures to 32/32 passing, Linux and Windows
Updated PR descriptions for #38088 and #812

PR #38104, opened 2026-07-30
Prevent orphaned S3 artifacts and false-positive collisions in AWS integration tests [5.0.0]

### LINKED ISSUE #38101: AWS integration tests: orphan artifacts from cancelled jobs cause shared-bucket setup collisions
## Description

The AWS integration tests were migrated to a shared S3 bucket in an IDR. A setup guard (`_assert_prefix_clean` in `tests/integration/test_aws/conftest.py`) fails a run when unexpected objects already exist at a test's prefix, to avoid silent count mismatches downstream.

Running the full AWS suite on a clean `5.0.0` produces **6 setup errors** while everything else passes.
These are pytest **setup errors** (a fixture raising during setup), **not** test failures - there are **zero `failed` assertions**. They surface two distinct problems in the shared-bucket model.

**Evidence (clean `5.0.0`, `workflow_dispatch`):**
- Run: https://github.com/wazuh/wazuh/actions/runs/30533875522/job/90842351203
- Result:
```
113 passed, 3 skipped, 66 deselected, 1652 warnings, 6 errors in 2882.76s (0:48:02)
```

The 6 errors map to **2 orphan objects** left in the shared bucket:

| Orphan object | Cases it breaks |
|---|---|
| `dnslogs/2022-11-20/2022-11-20-00-00-ioxa.csv` | `cisco_umbrella_discard_regex`, `cisco_umbrella_remove_from_bucket` |
| `test_prefix/AWSLogs/819751203818/elasticloadbalancing/us-east-1/2022/11/20/819751203818_elasticloadbalancing_us-east-1_net.qatests_20221120T0000Z_3947707909437559810_181.77.200.48.log` | `alb_path_with_data`, `clb_path_with_data`, `nlb_path_with_data`, `server_access_path_with_data` |

## Problem 1 - Orphan artifacts (recurrent)

The teardown in `manage_bucket_files` only deletes the keys uploaded by that same run, and it only runs after `yield`. When a CI job is **hard-cancelled** (a new push cancels the in-flight run), the already-uploaded files are never deleted. Because the generated keys embed random tokens (`get_random_ip()` / `get_random_string()` in the framework's `data_generator.py`), no future run's teardown ever references them, so they accumulate in the shared bucket and trip the guard on later runs.

This recurs on every cancelled AWS run, so cleaning the bucket by hand is not a stable fix.

## Problem 2 - Guard scoping false positive (server_access)

`ServerAccessDataGenerator` builds a **flat** key `test_prefix/<date>-<random>` (no `AWSLogs/YYYY/MM/DD` sub-path). `_assert_prefix_clean` derives the prefix with `key.rsplit('/', 1)[0] + '/'`, which for a flat key collapses to the entire `test_prefix/` subtree.
The ALB/CLB/NLB `path_with_data` cases also use `path=test_prefix` and live under `test_prefix/AWSLogs/.../elasticloadbalancing/...`, so the server_access guard scans them too and flags an unrelated load-balancer object.

(The cisco and alb/clb/nlb collisions are legitimate same-prefix hits on the orphans; only server_access is a guard-scoping false positive.)

## Proposed fix

1. **Guaranteed cleanup on cancellation.** Record every uploaded S3 key to a manifest file at upload time (not at teardown), and add a final CI step with `if: always()` that reads the manifest and deletes those keys (`aws s3 rm`), skipping `_PERMANENT_SEED_KEYS`. Unlike pytest teardown, an `always()` step still runs when the job is cancelled, which is the dominant orphan source.
2. **Fix guard scoping.** Scope `_assert_prefix_clean` to the test's own key namespace for
   flat-layout bucket types (server_access) instead of the shared `test_prefix/` root. Applies to all three release branches (parity): `4.14.7`, `4.14.8` and `5.x` / `5.0.0`. All ship the same guard and shared-bucket model. Confirmed on `4.14.8`: its AWS IT workflow
(`4_testintegration_aws-tier-0-1.yml`) uses the shared bucket and its `conftest.py` carries the same`_assert_prefix_clean` with the identical flat-key line `key.rsplit('/', 1)[0] + '/'`. Because it reaches 4.14.8 (a production branch), this is operational.

## Out of scope (DevOps, tracked separately)

An S3 lifecycle rule on the shared bucket as a backstop for the rare case where even an `always()` step cannot run (runner hard-kill / spot reclaim). It must be **tag-based** (tag the permanent seeds, expire only untagged objects), **not age-based**, because the permanent seeds are old objects and an age-based rule would delete them while keeping the recent orphans.

Darioortegaleyva:
## Description

Makes the AWS integration tests robust on the shared S3 bucket and fixes several issues surfaced while
getting them green:

- **Orphan artifacts on cancellation.** pytest teardown only deletes the keys of its own run and only
  runs after `yield`, so a hard-cancelled job left already-uploaded (randomly-named) objects as orphans
  that accumulated and tripped the setup guard. Now every uploaded key is recorded to a manifest at
  upload time and a CI step with `if: always()` deletes them even when the job is cancelled.
- **Guard scoping.** `_assert_prefix_clean` now lists with `Delimiter='/'` (only inspects objects
  directly under the prefix, removing the `server_access` false positive) and handles flat/root-level
  keys (`prefix=''`), so a root-level orphan is no longer missed.
- **Invocation-matching vs the module's argument quoting.** #37929 made the module wrap several CLI
  argument values in double quotes (`wm_aws_add_quoted_arg`) so they survive `wm_exec()`
  word-splitting; this reached 5.0.0 and broke `callback_detect_aws_module_called`, which expected the
  unquoted form. The matcher now quotes the known quoted arguments on the expected side and matches
  literally (`re.escape`), so it keeps working and still fails if the module ever stops quoting.
- **Cleanup credential window.** A `::warning::` is emitted if a cleanup delete fails (e.g. expired
  credentials) instead of failing silently.

**Proposed Release:** v5.0.0
**Issue:** wazuh/wazuh#38101

## Proposed Changes

- `tests/integration/test_aws/conftest.py`: upload-time manifest (`_record_uploaded_key` + a
  `record_uploaded_key` fixture for test-body uploads), `_assert_prefix_clean` with `Delimiter='/'` and
  root-level-key handling.
- `tests/integration/test_aws/event_monitor.py`: `callback_detect_aws_module_called` normalises the
  quoted arguments on the expected side and matches literally.
- `tests/integration/test_aws/test_only_logs_after.py`: the one test-body upload now records its key.
- `.github/workflows/5_testintegration_linux-integration-tests.yml`: pass the manifest to the test run;
  `if: always()` cleanup step (skips permanent seeds; warns on delete failure).

### Results and Evidence

Full `aws` and `aws_inspector` shards green on a clean bucket. Guard and matcher logic validated locally
(`moto`/`boto3` for the guard - false positive, root-level orphan, seed protection; the matcher across
quoted/unquoted/pre-quoted params, a quoting-regression case that fails, and literal `discard_regex`
matching).

### Configuration Changes

N/A

### Documentation Updates

N/A

### Tests Introduced

- **Scope:** Integration Tests (infrastructure).
- **Details:** guard validated with `moto`/`boto3`; invocation matcher validated across all quoting
  conventions + a regression case; cleanup shell logic validated against a fake `aws`/`sudo`.

## Review Checklist

> No compiled code changes (Python + workflow only) - compilation and memory checks are N/A.

- [x] Code changes reviewed

### DIFF
--- a/.github/workflows/5_testintegration_linux-integration-tests.yml
+++ b/.github/workflows/5_testintegration_linux-integration-tests.yml
@@ -225,12 +225,14 @@ jobs:
         env:
           AWS_VPC_ID: ${{ secrets.IT_AWS_VPC_ID }}
           AWS_BUCKET_NAME: ${{ secrets.IT_AWS_BUCKET_NAME }}
+          AWS_IT_CLEANUP_MANIFEST: ${{ github.workspace }}/aws_it_uploaded_keys.txt
         run: |
           cd tests/integration
           set +e
           sudo GITHUB_SHA="${{ github.sha }}" GITHUB_REF_NAME="${BRANCH_NAME}" \
             AWS_VPC_ID="${AWS_VPC_ID}" \
             AWS_BUCKET_NAME="${AWS_BUCKET_NAME}" \
+            AWS_IT_CLEANUP_MANIFEST="${AWS_IT_CLEANUP_MANIFEST}" \
             python -m pytest ${{ inputs.pytest_extra_args }} ${{ inputs.pytest_tier_args }} ${{ inputs.pytest_target }} \
             --html=results.html --self-contained-html
           EXIT_CODE=$?
@@ -240,6 +242,35 @@ jobs:
             exit 0
           fi
           exit $EXIT_CODE
+
+      - name: Clean up uploaded S3 test artifacts
+        # always() so this runs even when the job is CANCELLED - the dominant source of orphan
+        # objects, since pytest teardown does not run on a hard cancel. The manifest lists every key
+        # the tests uploaded (recorded at upload time); we delete each one, skipping permanent seeds.
+        if: always() && (inputs.setup_aws_credentials == true || inputs.setup_aws_credentials == 'true')
+        env:
+          AWS_BUCKET_NAME: ${{ secrets.IT_AWS_BUCKET_NAME }}
+          AWS_IT_CLEANUP_MANIFEST: ${{ github.workspace }}/aws_it_uploaded_keys.txt
+        run: |
+          MANIFEST="${AWS_IT_CLEANUP_MANIFEST}"
+          if [ ! -f "$MANIFEST" ]; then
+            echo "No cleanup manifest found; nothing to delete."
+            exit 0
+          fi
+          # Permanent DevOps seeds that must never be deleted (belt-and-suspenders; the manifest never
+          # records them, but guard here in case a future test key ever matches one).
+          SEED_CLOUDTRAIL="AWSLogs/819751203818/CloudTrail/us-east-1/2022/11/20/819751203818_CloudTrail_us-east-1_20221120T0000Z_372406355707169122.json"
+          SEED_ELB="AWSLogs/819751203818/elasticloadbalancing/us-east-1/2022/11/20/819751203818_elasticloadbalancing_us-east-1_net.qatests_20221120T0000Z_1412953514070675864_29.145.39.119.log"
+          sudo sort -u "$MANIFEST" | while IFS= read -r key; do
+            [ -z "$key" ] && continue
+            if [ "$key" = "$SEED_CLOUDTRAIL" ] || [ "$key" = "$SEED_ELB" ]; then
+              echo "Skipping permanent seed: $key"
+              continue
+            fi
+            echo "Deleting s3://${AWS_BUCKET_NAME}/${key}"
+            sudo aws s3 rm "s3://${AWS_BUCKET_NAME}/${key}" --profile default || true
+          done
+          sudo rm -f "$MANIFEST" || true
 
       - name: Dump wazuh-agent logs on failure
         if: failure()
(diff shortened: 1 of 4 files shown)


Darioortegaleyva:
### CI evidence

The `aws` integration-test shard passes on a clean shared bucket with this fix:

✅ https://github.com/wazuh/wazuh/actions/runs/30545681519/job/90899089838 - **116 passed**, 0 setup errors.

The `SETUP COLLISION` errors that motivated this PR are gone, and the `server_access` false positive is fixed. (The 2 pre-existing orphan objects were removed from the shared bucket out-of-band; this PR prevents new orphans from accumulating and narrows the guard so it no longer flags nested objects.)

MarcelKemp (changes requested):
I think we’ve got another instance of `upload_bucket_file` that we haven’t covered; let’s apply the same change to it just to be on the safe side.

MarcelKemp (approved):

PR #38118, opened 2026-07-30
Fix macos drain autoreleased OD objects in genAccountPolicyData

### LINKED ISSUE #38081: wazuh-modulesd RSS grows steadily on macOS agent during load test (fails baseline `growth_pct` check)
## Description

In a 5.0.0 agent load-test run, the baseline suite reported a single failure out of 202 checks: `agent-node-macos-26-arm64::modulesd_macos_memory_growth_pct`. Every other daemon on every other platform (Linux amd64/arm64 on two distros, Windows) reported 0% or negative memory growth in the same run — macOS `wazuh-modulesd` is the only positive-growth result in the entire matrix.

```
growth_pct(RSS(KB)) = 61.3448  >=  35  (threshold: < 35 %)
```

macOS and Windows agents did **not** show this behavior in the earlier soak-test investigation in #37656 (which was Linux-only and root-caused to glibc allocator retention, not a live leak) — so this looks like a distinct, previously unseen signal, not a recurrence of that fix.

Image

## Environment

- **Wazuh version:** 5.0.0
- **Host:** `performance-testing-286-staging-agent-macos-26-arm-2842`, macOS 26, arm64, Parallels VM on MacStadium (`macstadium_arm_4`), single agent connected to the staging manager
- **Jenkins job:** `5.0.0/Performance_Testing` *(link the specific build here)*

### Saturation profile (from `wazuh_config.yaml`, macOS rule)

_Logcollector and syscheck saturation were disabled and monitored ps command was invalid_

- SCA: default policy, scan interval 1m, sync every 5m
- Syscollector (IT Hygiene): full inventory scan every **1m**
- Rootcheck: frequency 1s (hidden files/ports/pids)
- Wodle-command: `ps -auxw` and `stat /etc/hosts`, both every 1s. `ps`

## Evidence

| Calculation | Baseline (mean, first 10%) | End level (p95, last 10%) | growth_pct |
|---|---|---|---|
| Official (spans the restart) | 32,907 KB | 54,642 KB | **61.34%** *(reported)* |
| Isolated to PID 8097 only, restart fully excluded | 29,620 KB | 60,342 KB | **~104%** |

Removing the restart from consideration makes the measured growth *larger*, not smaller — the pre-restart plateau (~34,032 KB) is actually higher than the low point right after the restart, so blending it in slightly dilutes the reported number. The growth is a real trend within a single continuous process lifetime, unrelated to the restart event.

Looking at the raw post-restart series directly:

- Shortly after restart (~04:28–04:30): RSS settles around **27,000–29,000 KB**.
- By the end of the run (~05:34–05:40): the resting-state plateau between transient spikes has climbed to **45,700–47,700 KB**.
- On top of that steady climb, there's a separate, regular ~61-second-interval transient spike (to 63,000–78,000 KB) that appears and clears cleanly each time — this cadence matches the syscollector 1-minute inventory-scan interval in the config above, and is not itself the trend; it's a recurring transient overlaid on top of a separately-rising baseline.
- **Tail end of the run:** the last ~10 samples (05:40:15 onward) drop from 47,088 KB to a flat 42,496 KB. This looks like the saturation/load profile ramping down near the end of the collection window (this is also the point where the macOS saturation harness stopped emitting anything at all — see the `setsid` issue below), not the memory being released back down. The resting level even at that final point (42,496 KB) is still ~50% above the immediate post-restart baseline (27,000–29,000 KB) — it is a partial pull-back, not a return to baseline, so it should not be read as evidence against the growth trend.

## Related issues

- #6483 (open, 2020, v3.13.2): a ~1.5 KB static `leaks`-tool finding at macOS `modulesd` startup. Different order of magnitude and a much older codebase — not a clear match for this run's ~15–20 KB sustained growth.
- #37656 (closed): near-identical shape of `modulesd` RSS growth, but on **Linux**, explicitly noting macOS did *not* show the behavior at the time. Root-caused to glibc allocator arena retention (freed-but-not-returned memory, not a live leak) driven by syscollector's per-cycle full-table rebuild/DBSync. That mechanism is Linux/glibc-specific and doesn't obviously transfer to macOS's allocator, so this macOS growth needs its own root-cause pass rather than being assumed to be the same issue.

### Artifacts

- [observability_data.zip](https://github.com/user-attachments/files/30513107/observability_data.zip)

cborla:
## Description

`wazuh-modulesd` RSS grows steadily on macOS agents during load testing (~61-104% growth over ~1h run), failing the baseline `modulesd_macos_memory_growth_pct` check. Every other daemon/platform combination in the same run showed 0% or negative growth — macOS `wazuh-modulesd` was the only outlier.

Root cause: `genAccountPolicyData()` (`src/data_provider/src/extended_sources/wrappers/unix/darwin/od_wrapper.mm`) creates several autoreleased OpenDirectory/Foundation objects (`ODNode`, `ODQuery`, `NSArray` results, and a full `NSDictionary` tree built from the user's `accountPolicyData` plist via `NSPropertyListSerialization`) without wrapping them in an `@autoreleasepool`. `genEntries()`, in the same file, already does this correctly.

This function is called once per local user, on every syscollector full inventory scan cycle (`scanUsers()` runs each cycle per `syscollectorImp.cpp`). Since the syscollector scan thread has no Cocoa run loop to implicitly drain autoreleased objects, every call leaked its OpenDirectory/plist object graph for the life of the process — explaining why RSS only dropped back down on daemon restart, and why the growth was macOS-specific (Objective-C/OpenDirectory doesn't exist on Linux/Windows).

Closes #38081

## Proposed Changes

- `src/data_provider/src/extended_sources/wrappers/unix/darwin/od_wrapper.mm`: wrapped the body of `genAccountPolicyData()` in `@autoreleasepool { ... }`, matching the existing pattern already used in `genEntries()`. No behavior or data changes — this only ensures the autoreleased objects created per call are drained instead of accumulating.

### Results and Evidence

Static code-review finding, not yet validated with a live macOS profiling run. Confirmed via code inspection that:
- `genEntries()` (line 22) already wraps its OpenDirectory calls in `@autoreleasepool`.
- `genAccountPolicyData()` did not, and is the only other place in the file creating Objective-C objects.
- `grep -rn "autoreleasepool"` across the whole repository returned a single match (inside `genEntries()`), confirming no outer-level pool exists anywhere on the syscollector scan thread to drain these objects otherwise.
- `collectAccountPolicyData()` (`users_darwin.cpp`) confirms `genAccountPolicyData()` is called once per local user, and `scanUsers()` runs on every syscollector inventory cycle (`syscollectorImp.cpp`), matching the reported growth cadence.

Recommended before merge: re-run the macOS load test (or `leaks --atExit -- wazuh-modulesd` / Instruments Allocations trace) with this fix to confirm RSS growth is resolved.

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - N/A

  - N/A

- Wazuh server API/Framework
  - N/A

### Artifacts Affected

- `wazuh-modulesd` (agent, macOS only) — no other executables, configuration files, or packages affected.

### Configuration Changes

N/A — no new configuration parameters, default value changes, or backward-compatibility impact.

### Tests Introduced

None added. `genAccountPolicyData()` is implementation code behind `IODUtilsWrapper`, exercised through the real OpenDirectory framework rather than through the existing mocked unit tests for `users_darwin`/`groups_darwin`, so this change isn't covered by a new unit test — verification is via macOS-side memory profiling (see Results and Evidence).

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

### DIFF
--- a/src/data_provider/src/extended_sources/wrappers/unix/darwin/od_wrapper.mm
+++ b/src/data_provider/src/extended_sources/wrappers/unix/darwin/od_wrapper.mm
@@ -71,9 +71,6 @@ namespace od
 
     void genAccountPolicyData(const std::string& uid, nlohmann::json& policyData)
     {
-        ODSession* s = [ODSession defaultSession];
-        NSError* err = nullptr;
-
         policyData =
         {
             {"creation_time", 0.0},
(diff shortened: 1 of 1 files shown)


anromerom (approved):
**LGTM**

- [x] Code changes reviewed

PR #38135, opened 2026-07-31
Improve login attempt limiting under concurrent requests

vikman90:
## Description
The per-IP login attempt count was only updated after credential validation finished, separately from the check that decides whether an IP is currently limited. When several login requests from the same IP arrived at once, the count could fall behind, so the limit was not applied consistently. This change updates the count at the same time the limit is checked, so the result no longer depends on request timing, and releases the count again once a login succeeds so only failed attempts count against the limit long-term.

## Proposed Changes
- `api/api/middlewares.py`: `check_blocked_ip` now updates the per-IP attempt count and limit state together, under the existing lock, before the request proceeds to authentication. The per-IP timestamp is refreshed on every counted attempt instead of only the first. Added `settle_login_attempt`, which releases that count when a login succeeds.
- `api/api/error_handler.py`: removed `prevent_bruteforce_attack`, now redundant since the counting moved into `check_blocked_ip`, and its call from `unauthorized_error_handler`, along with the imports it alone used (`ip_block`, `ip_stats`, `ip_lock`, `configuration`, `BlockedIPException`).
- `api/api/controllers/security_controller.py`: `login_user`, `deprecated_login_user`, and `run_as_login` now call `settle_login_attempt` once authentication has succeeded, so a client authenticating frequently from one IP (NAT, load balancer, token-per-call automation) is never limited as long as its credentials stay valid.
- `api/api/test/test_middlewares.py`: updated `test_middlewares_check_blocked_ip` for the new counting behavior; added tests covering attempt counting, enforcing the limit without a completed auth failure, releasing the count on success, unblocking once released, a no-op for unknown hosts, and repeated successful logins never triggering a block. Also removed a dead fixture check for `prevent_bruteforce_attack`, a function that no longer exists.
- `api/api/test/test_error_handler.py`: removed the `prevent_bruteforce_attack` tests (function no longer exists) and updated `test_unauthorized_error_handler` to no longer expect a call to it.
- `api/api/controllers/test/test_security_controller.py`: extended the `mock_request` fixture with a `client.host` attribute and added it to `test_login_user_ko`, since `login_user`/`run_as_login` now read the request's client host.

### Results and Evidence
Manual test exercising the real (unmocked) `check_blocked_ip` with 50 simulated concurrent login requests from the same IP and a limit of 5 attempts:

- **Before the change:** 50/50 requests were let through, 0 limited — the configured limit had no effect when requests arrived concurrently.
- **After the change:** 5/50 requests were let through, 45 limited — the configured limit is applied consistently regardless of request timing.

Manual test simulating 200 rounds of `check_blocked_ip` + `settle_login_attempt` for the same IP, always with valid credentials, limit of 5 attempts: 0 rounds were limited, final recorded attempts for the IP settled back to 0 — frequent legitimate logins are never limited.

Full automated suite: `api/api/test/test_middlewares.py`, `api/api/test/test_error_handler.py`, and `api/api/controllers/test/test_security_controller.py`, 138 tests passed after the change; the new/updated tests fail (or fail to collect) against the prior code and pass after it. Broader regression check: `api/` unit tests (656 tests) passed with no regressions.

### Artifacts Affected
`wazuh-apid` (API service, `api/api/` Python source only — no separate build artifact).

### Configuration Changes
None.

### Documentation Updates
None.

### Tests Introduced
- `api/api/test/test_middlewares.py::test_middlewares_check_blocked_ip` (updated)
- `api/api/test/test_middlewares.py::test_middlewares_check_blocked_ip_counts_attempt` (new)
- `api/api/test/test_middlewares.py::test_middlewares_check_blocked_ip_enforces_limit_without_auth_failure` (new)
- `api/api/test/test_middlewares.py::test_middlewares_settle_login_attempt` (new)
- `api/api/test/test_middlewares.py::test_middlewares_settle_login_attempt_unknown_host` (new)
- `api/api/test/test_middlewares.py::test_middlewares_repeated_successful_logins_never_block_ip` (new)
- `api/api/test/test_error_handler.py::test_unauthorized_error_handler` (updated)
- `api/api/controllers/test/test_security_controller.py::test_login_user_ko` (updated)

## Credits

Issue reported by @iabdullah215.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.
 - Improved SCA policy source validation to correctly enforce the `sca.remote_commands` restriction. ([#37953](https://github.com/wazuh/wazuh/pull/37953))
 - Added a minimum length check for legacy-format agent messages in `wazuh-remoted`. ([#38099](https://github.com/wazuh/wazuh/pull/38099))
 - Fixed a spurious `StarletteDeprecationWarning` printed by `agent_upgrade` at startup. ([#38085](https://github.com/wazuh/wazuh/pull/38085))
+- Fixed the API login attempt limit not being applied consistently under concurrent requests. ([#38135](https://github.com/wazuh/wazuh/pull/38135))
 
 ### Agent
 
--- a/api/api/controllers/security_controller.py
+++ b/api/api/controllers/security_controller.py
@@ -13,6 +13,7 @@ from api.models.security_token_response_model import TokenResponseModel
 from api.authentication import generate_token
 from api.configuration import default_security_configuration
 from api.controllers.util import json_response, JSON_CONTENT_TYPE
+from api.middlewares import settle_login_attempt
 from api.models.base_model_ import Body
 from api.models.configuration_model import SecurityConfigurationModel
 from api.models.security_model import (CreateUserModel, PolicyModel, RoleModel,
@@ -49,6 +50,7 @@ async def deprecated_login_user(user: str, raw: bool = False) -> ConnexionRespon
     ConnexionResponse
         Raw or JSON response with the generated access token.
     """
+    await settle_login_attempt(request)
     f_kwargs = {'user_id': user}
 
     dapi = DistributedAPI(f=preprocessor.get_permissions,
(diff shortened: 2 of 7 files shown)


jotacarma90 (changes requested):
## Review

Counting at the gate closes the race, but the reservation is never settled: successful logins also increment the counter and nothing resets it on a 200. 
A client that authenticates frequently from one IP (NAT, load balancer, token-per-call automation) gets blocked with 100% valid credentials. The advisory's suggested fix explicitly says "reserve/increment before dispatching to authentication, and decrement/settle afterwards"; this PR implements the increment but not the settle.

Minor: the fixture in api/api/test/test_middlewares.py still checks for prevent_bruteforce_attack in test names, which no longer exist.

jotacarma90 (approved):
LGTM

PR #38077, opened 2026-07-29
feat: Invenroty sync base structure

juliancnn:
Closes #38018

## Description

Adds `inventory_sync_server`, a new, transitional `wazuh-manager-modulesd` module that exposes an
inventory synchronization ingress endpoint to `remoted` over HTTP/1.1-over-Unix-domain-socket. It
lives alongside the existing router-based `inventory_sync` module while the migration to the new
transport completes; when it does, `inventory_sync` will be removed.

The module ships with its own hand-written asio/llhttp transport (connection lifecycle, admission
control, in-flight byte budget, two-phase shutdown), a four-stage startup gate that builds a shared
indexer session plus a synchronous and an asynchronous indexer connector before the socket opens, and
a C-ABI bridge to modulesd's C shim. `POST /inventory/sync` is currently a stub that accepts and
discards its payload (202); `remoted` does not yet have a downstream client for it. `POST /stats` and
`POST /config`, reachable from `remoted` today, are also dummy endpoints: the whole pipeline
(authentication, admission control, deferred forwarding, the UDS round trip) is real, but neither side
interprets the document yet.

This PR also includes a second remediation pass (post-review) that hardens the transport's exception
safety, closes several logging gaps that would have made production incidents hard to diagnose, and
fixes a few cross-module diagnosability issues on the `remoted` side of the pipeline. See "Proposed
Changes" below for the itemized list.

### `inventory_sync_server` module (new)

- Added the module's C-ABI bridge (`inventory_sync_server_start`/`_stop`) and its C shim
  (`wm_inventory_sync_server.{c,h}`), wired into `wazuh_modulesd_lib` and `wmodules.c`.
- Added the HTTP-over-UDS transport (`src/http_server/`): connection-per-strand session handling,
  admission control ahead of body reads, an in-flight byte budget derived from the configured header
  limits (not a hardcoded constant), a bounded two-phase shutdown protocol, and a request parser over
  `llhttp`.
- Added the indexer seam (`src/indexer/`): interfaces over the shared `IndexerConnectorSync`/
  `IndexerConnectorAsync`/`IndexerSession`, so the startup gate is testable without real indexer I/O.
- Added the startup gate in `InventorySyncServerFacade`: builds the shared session and both
  connectors, in order, memoised across retries; a heartbeat retries a failed attempt every 60 s with
  an escalating log (ERROR on the first attempt, then debug, then hourly WARN); an unbindable socket
  path is the one condition that aborts the daemon instead of retrying.
[...]

PR #38136, opened 2026-07-31
Persist the statistics agents report and drop the endpoint that used to query them

### LINKED ISSUE #38024: Implement `/stats` HTTPS endpoint
## Description

Implement the manager-side `/stats` endpoint as defined by the HTTPS communication protocol.

This endpoint allows agents to periodically publish their runtime statistics to the manager. The manager validates the received payload and persists the latest statistics for each agent. Every new report completely replaces the previous stored statistics.

This endpoint replaces the legacy `getstats` communication model, where the manager queried the agent directly. Under the new HTTPS architecture, the agent proactively reports its statistics and the dashboard reads them from the index directly, so `GET /agents/{agent_id}/stats/{component}` is removed rather than reimplemented on top of the persisted data.

---

### Request

The agent reports every module it can collect statistics from in one push, keyed by module name. Each module's body is its own field set, which the manager indexes as it arrives.

```json
{
  "modules": {
    "agent": {
      "status": "connected",
      "last_keepalive": "2026-08-02T10:06:50Z",
      "messages": { "count": 602 },
      "tasks": {
        "dispatched":          { "total": 4 },
        "discarded_duplicate": { "total": 0 },
        "failed":              { "total": 0 }
      }
    },
    "logcollector": {
      "global":   { "start": "…", "end": "…", "files": [] },
      "interval": { "start": "…", "end": "…", "files": [] }
    }
  }
}
```

### Response

```json
{}
```

---

## Tasks

* Implement the `/stats` HTTPS endpoint.
* Authenticate and authorize the requesting agent.
* Validate the received statistics payload.
* Persist the latest statistics for the agent using the manager indexing pipeline.
* Replace the previously stored statistics for the agent.
* Return the protocol-defined empty acknowledgment.
* Handle malformed payloads and persistence failures.
* Remove `GET /agents/{agent_id}/stats/{component}` and the live-query chain that only served it.
* Define the `wazuh-agent-stats` index template with the indexer team (wazuh/wazuh-indexer-plugins#1425).
* Update unit and integration tests for both the HTTPS endpoint and the Server API.
* Add logging.

---

## Definition of Done

* Agents can successfully upload runtime statistics.
* The latest statistics are persisted for each agent.
* Each upload replaces the previous stored statistics.
* The endpoint returns the expected acknowledgment.
* `GET /agents/{agent_id}/stats/{component}` and its live-query chain are removed.
* The index template is agreed with the indexer team and created by the setup plugin.
* Errors are handled correctly.
* Logging is implemented.
* Unit and integration tests pass.
* CI passes successfully.

ignaciogalle12git:
## Description

The Server API read an agent's statistics by querying the agent and waiting for its answer. HTTPS has no manager-to-agent channel, so the direction reverses: the agent pushes, and the manager stores the latest report per agent in `wazuh-agent-stats`. The dashboard reads that index directly, so `GET /agents/{agent_id}/stats/{component}` is removed rather than reimplemented on top of it.

This fills the hole #38077 left on purpose, where the `/stats` handler echoed the document back and discarded it.

Closes #38024

## Proposed Changes

**`inventory_sync_server`** indexes the report. `modules` moves under `wazuh.agent.statistics` and the document id is the agent id, so every push replaces the previous report. The document is built from scratch, which drops the `agent_id` and `cluster` the agent writes at its own root. What is left is validation, not translation: `modules` must be a non-empty object whose every value is an object, and anything else is a `400`. Rejecting the empty case is deliberate, since indexing it would replace the agent's last good report with one carrying no statistics.

**`remoted_module`** builds the `{}` acknowledgment in `postProcess()` instead of forwarding modulesd's body, so no downstream string reaches an agent.

It also canonicalizes the agent id. The key lookup resolves it numerically, so `1`, `001` and `0001` authenticate as the same agent; the header value reached the endpoints verbatim, which was harmless while it was echoed back and is not now that it is the document id. The session keeps the signed bytes for the MAC and propagates the client.keys form.

**Server API** loses the endpoint: the route and its `stats_component` parameter, the controller, `get_agents_component_stats_json`, error 1762, and the tavern stages. With it goes the chain that only served it: `Agent.get_stats()`, `get_daemons_stats_from_socket()` and its socket helpers, `PaginatedDataHandler` and `AGENT_COMPONENT_STATS_REQUIRED_VERSION`. `get_daemons_stats_socket` stays, since it serves the manager's own daemon stats.

### Results and Evidence

Contract validated against a real `wazuh-indexer` 5.0.0 (OpenSearch 3.6.0), applying the proposed template and indexing directly:

| Case | Result |
| -- | -- |
| Index template applied | `{"acknowledged":true}` |
| Reference document | `201 created` |
| `last_keepalive: ""` (what a 5.x agent sends today) | `400 mapper_parsing_exception`, *"cannot parse empty date"* |
| `last_keepalive: null` (what #37843 will send) | `201 created` |
| `error`/`data` envelope surviving | `400 strict_dynamic_mapping_exception` |
| `"2026-08-02 10:06:50"` (agent's current format) | `400`, *"failed to parse date field ... with format [strict_date_optional_time\|\|epoch_millis]"* |

The template is `dynamic: strict` with every leaf declared, which is the one coupling this PR does not remove: an agent-side addition needs no change in modulesd and does need one in the index template.

| Case | Result |
| -- | -- |
| A module the template does not declare (`fim`) | `400`, *"dynamic introduction of [fim] within [wazuh.agent.statistics] is not allowed"* |
| A metric it does not declare (`agent.messages.dropped`) | `400`, *"dynamic introduction of [dropped] within [wazuh.agent.statistics.agent.messages] is not allowed"* |

Both are silent from the manager's side: the write is fire-and-forget, so the agent already has its `200` and nothing in `wazuh-manager.log` reports the document never landed. Reading it back off the indexer is the only way to see it. Stated in `statsEndpoint.hpp`, both module READMEs and the module's API reference, and pinned by a test.

Unit tests:

- `inventory_sync_server_utest`: 204 passed, all 25 `StatsEndpointTest` among them. 2 failed, `UdsHttpServerTest.ThreeHundredConcurrentDeferralsOnTwoIoThreads` and `UdsShutdownTest.StopAcceptingThenStopUnderManyLiveDeferrals`, both in the UDS transport and untouched by this PR: the first asserts 300 requests in flight at once and the test machine reaches 64, which then drags the second down.
- `remoted_module_utest`: passed, including the 8 `StatsEndpoint`/`StatsMakeHandler` cases and the new canonicalization one.
- `framework`: 1961 passed, 6 skipped. `api/api`: 443 passed.

  - [x] Linux
  - [ ] Windows 
  - [ ] MAC OS X 

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer 

- Wazuh server API/Framework

### Artifacts Affected

`wazuh-manager-remoted` and `wazuh-manager-modulesd` (the `inventory_sync_server` module), plus the Server API framework. No packaging or default configuration changes.

New index written by the manager: `wazuh-agent-stats`, requested from the indexer team in wazuh/wazuh-indexer-plugins#1425.

**A public API endpoint is removed.** Any consumer of `GET /agents/{agent_id}/stats/{component}` has to read `wazuh-agent-stats` instead.

### Configuration Changes

N/A

### Tests Introduced

`statsEndpoint_test.cpp` covers the report landing under `wazuh.agent.statistics`, each module body stored verbatim, an undeclared module passed through for the indexer to judge, every rejection that leaves nothing to store, the document id and index name, every identity the report claims being discarded, and the indexer gate running after each `400`. `authMiddleware_test.cpp` covers four spellings of one agent id all authenticating and all arriving as `001`.

## Review Checklist

- [ ] Code changes reviewed

This has to land with the agent side (#37843). Until it does, a 5.x agent sends `last_keepalive` as an empty string and its timestamps in the old format, and the first table shows both reject the document.


ignaciogalle12git:
# `/stats` agent side

Verified against `enhancement/37831-agent-https-client` (`f92fbf477d`), where the HTTPS client lives. None of it blocks merging the manager side.

The manager indexes the report as it arrives: it renames nothing and reshapes nothing. So the target payload below is at once the wire format and the stored document, and every field name in the index is the agent's own. A mapping table on the manager would have to be edited for every metric the agent ever adds, and only the agent knows which of its counters are cumulative.

## What the agent sends today

```json
{
  "modules": [
    { "module": "agent", "stats": { "error": 0, "data": {
        "status": "connected",
        "last_keepalive": "2026-08-02 10:06:50",
        "last_ack": "2026-08-02 10:06:50",
        "msg_count": 602, "msg_sent": 689, "msg_buffer": 0, "buffer_enabled": true,
        "task_dispatched": 4, "task_discarded_duplicate": 0, "task_failed": 0 } } },
    { "module": "logcollector", "stats": { "global": { "…": "…" }, "interval": { "…": "…" } } }
  ],
  "agent_id": "001",
  "cluster": { "name": "wazuh", "node": "node01" }
}
```

## Requested

### 1. Send `modules` as an object keyed by module name, not an array

```json
"modules": { "agent": { … }, "logcollector": { … } }
```

This is the one that removes the last processing step on the manager. Today it walks the array, reads each `module`/`stats` pair and rebuilds the object — code that has to keep working for every module that ever gains a statistics producer. Keying it on the agent's side means the manager only validates the object and moves it under `wazuh.agent.statistics`.

One entry per module is already the invariant (nothing reports twice), so the object loses nothing the array carried. And an array cannot be read out of the stored document by key without a `nested` mapping.

`agent_id` and `cluster` at the root can stay or go: the manager builds the document from scratch and stamps the authenticated identity, so they are dropped either way.

### 2. Drop the `error`/`data` envelope

`agcom_getallstats()` parses `w_agentd_state_get()` whole, and that is what builds the envelope; `lccom_getallstats()` reads its state directly and comes out clean. It is socket framing: over HTTPS the error is the response status. Make `w_agentd_state_get()` return `cJSON *` and keep the envelope at the `getstate` call site.

The manager does not unwrap it. An envelope that survives is indexed as `statistics.agent.error` and `statistics.agent.data`, which the template does not declare, so with `dynamic: strict` the whole document is rejected.

### 3. ISO 8601 UTC timestamps

`"%Y-%m-%d %H:%M:%S"` + `localtime_r` → `"%Y-%m-%dT%H:%M:%SZ"` + `gmtime_r`, in `client-agent/include/state.h:14` and `logcollector/src/state.c:23`. The offset is only known where the string is built, so only the agent can write an unambiguous instant: for an agent outside UTC the stored value is off by its offset and nothing downstream can recover it. These constants also feed `getstate` and the `.state` file, so decide whether the change is global or only on the report path.

### 4. Expand the abbreviations and nest them, rather than compounding the names

```json
"messages": { "count": 602 },
"tasks":    { "dispatched": { "total": 4 }, "discarded_duplicate": { "total": 0 }, "failed": { "total": 0 } }
```

`msg_count` → `messages.count`, and the three `task_*` → `tasks.*.total`. The dotted form is how the schema writes nesting everywhere else (`wazuh.agent.id`, `host.os.name`); an underscore compound would be the exception. `.total` is `wcs`'s marker for a counter that is monotonic over the process's uptime, which these three are.

### 5. Repopulate `last_keepalive`, and never send `""`

Not a formatting problem: the agent has no value to send. `agent_state.last_keepalive` is a `time_t` that starts at 0, and the only way to write it is `w_agentd_state_update(UPDATE_KEEPALIVE, &t)` — which has **no callers** on this branch. The two that had it, `notify.c` and `receiver.c`, went with the legacy protocol. So it stays 0 forever.

`w_agentd_state_get()` then formats it only when non-zero, but emits it unconditionally:

```c
char last_keepalive[W_AGENTD_STATE_TIME_LENGHT] = {0};                    // 249: zeroed
if (agent_state.last_keepalive) {                                          // 266: never taken
    localtime_r(&agent_state.last_keepalive, &tm);
    strftime(last_keepalive, sizeof(last_keepalive), W_AGENTD_STATE_TIME_FORMAT, &tm);
}
cJSON_AddStringToObject(data, W_AGENTD_FIELD_KEEP_ALIVE, last_keepalive);  // 288: unconditional
```

So every push carries `"last_keepalive": ""`. The field is mapped `date`, and an empty string is not a parsable date, so **the document is rejected whole** — silently, because the write is fire-and-forget. (That rejection is `ignore_malformed` defaulting to false on a declared field, not `dynamic: strict`, which only governs undeclared ones.)

Two things needed, and the second one is not optional:

- **A value.** Wiring `UPDATE_KEEPALIVE` back up is yours to place; you know the HTTPS lifecycle better than we do. One suggestion: the legacy call site was immediately after `send_msg()` (`notify.c:281`), i.e. send time with no confirmation, so an agent with the network down kept advancing it. `bridge_on_manager_config_hash()` (`https_client_bridge.c:876`) fires on every *accepted* Notify, which is the exact event the manager records as its own keepalive (`/control` issues `update-keepalive` there) — using it would make the value prove a round trip, which the legacy one never did.
- **`null`, never `""`, when there is no value.** There is always a window with nothing to report: the first push happens before the first accepted Notify, and the same is true after any reconnect. `null` is indexed as absent; `""` rejects the document. This is the one that silently eats the first document of every agent if it is missed.

Plus ISO 8601 UTC, per request 3.

Note on what the value means, so the two do not get confused downstream: this is the **agent's** clock, and the manager keeps its own `last_keepalive` (wazuh-db, stamped `STRFTIME('%s','NOW')` when the Notify arrives, surfaced as `wazuh.agent.last_seen` in `wazuh-metrics-agents`). They were already two different things before HTTPS — send time on the agent versus arrival time on the manager — and they stay two different things. Expect them to differ by the agent's clock skew.

### 6. Stop sending four fields

| Field | Why |
|---|---|
| `last_ack` | Same story: `UPDATE_ACK` has no callers, so it is always `""` and rejects the document the same way. The ack is a legacy-protocol concept with nothing to replace it |
| `msg_sent` | `INCREMENT_MSG_SEND` has no callers. Always 0 |
| `msg_buffer` | Hardcoded `0` (`state.c:293`). `buffer.c` is deleted and `w_agentd_get_buffer_lenght()` has no references: there is no agentd-side buffer left to count |
| `buffer_enabled` | Hardcoded `false` (`state.c:294`), and misleading — the HTTPS accumulator *does* buffer (byte occupancy up to `capMultiplier x batchSize`, drop-newest on overflow). A field asserting there is no buffer when there is one is worse than no field. `<client_buffer>` and `getBufferConfig()` no longer exist; the accumulator's tunables are internal options (`agent.warn_level`, `agent.normal_level`, `agent.tolerance`) that `/config` already carries |

If a buffer metric is wanted later, the useful ones are the accumulator's occupancy, a drop counter, and possibly the `BufferLevelLadder` level — which already reaches the C core through `on_buffer_level` but only feeds logs and the `wazuh-agent.buffer` event, never `agent_state`.

`status` stays. For the record: the stats reporter is gated on `registered` (`reporterStream.cpp:67`), so the only value that can reach the index is `connected`.

## Target payload

```json
{
  "modules": {
    "agent": {
      "status": "connected",
      "last_keepalive": "2026-08-02T10:06:50Z",
      "messages": { "count": 602 },
      "tasks": { "dispatched": { "total": 4 }, "discarded_duplicate": { "total": 0 }, "failed": { "total": 0 } }
    },
    "logcollector": {
      "global": {
        "start": "2026-08-02T09:43:50Z",
        "end": "2026-08-02T10:06:50Z",
        "files": [
          { "location": "df -P", "events": 32, "bytes": 3436,
            "targets": [ { "name": "agent", "drops": 0 } ] }
        ]
      },
      "interval": { "…": "same shape as global" }
    }
  }
}
```

ignaciogalle12git:
# `/stats` indexer side

The manager writes one document per agent to `wazuh-agent-stats` (#38024). **`_id` is the agent id**, so every push replaces the previous report and there is no history.

What follows is at once what the agent sends and what the index holds. The names are agreed with the agent side (#37843), which sends `modules` keyed by module name, drops its transport envelope, emits ISO 8601 UTC and uses the nested counters. The manager transforms nothing.

## What the manager writes

```json
{
  "state": { "modified_at": "2026-08-02T10:07:12.431Z", "document_version": 1 },
  "wazuh": {
    "schema": { "version": "1" },
    "cluster": { "name": "wazuh", "node": "node01" },
    "agent": {
      "id": "001",
      "statistics": {
        "agent": {
          "status": "connected",
          "last_keepalive": "2026-08-02T10:06:50Z",
          "messages": { "count": 602 },
          "tasks": {
            "dispatched":          { "total": 4 },
            "discarded_duplicate": { "total": 0 },
            "failed":              { "total": 0 }
          }
        },
        "logcollector": {
          "global": {
            "start": "2026-08-02T09:43:50Z",
            "end": "2026-08-02T10:06:50Z",
            "files": [
              { "location": "df -P", "events": 32, "bytes": 3436,
                "targets": [ { "name": "agent", "drops": 0 } ] }
            ]
          },
          "interval": { "…": "same shape as global" }
        }
      }
    }
  }
}
```

Three properties of the data drive the mapping:

- The keys under `wazuh.agent.statistics` are module names, and the set is open. Today only `agent` and `logcollector` have a statistics producer; the rest appear as they gain one.
- Each module's metrics are an open set too, and they move without notice.
- Inside a module's body there are arrays of objects nested two deep: `files[]` and, inside each element, `targets[]`.



## How it is read

  The manager serves no read path. `GET /agents/{agent_id}/stats/{component}`
  existed for the pull model and is removed in 5.0.0, so the index is the only
  source for agent statistics.


## Template

Also at `index-definitions/wazuh-agent-stats.template.json`. No `settings`: sizing, replicas and codec are yours to decide, and nothing in our access pattern depends on them — a read by document id is realtime, so `refresh_interval` does not affect it, and we never issue a `query_string`, so there is no default field to set.

```json
{
  "index_patterns": ["wazuh-agent-stats*"],
  "priority": 1,
  "template": {
    "mappings": {
      "dynamic": "strict",
      "date_detection": false,
      "properties": {
        "state": {
          "properties": {
            "modified_at":      { "type": "date" },
            "document_version": { "type": "integer" }
          }
        },
        "wazuh": {
          "properties": {
            "schema": {
              "properties": {
                "version": { "type": "keyword", "ignore_above": 1024 }
              }
            },
            "cluster": {
              "properties": {
                "name": { "type": "keyword", "ignore_above": 1024 },
                "node": { "type": "keyword", "ignore_above": 1024 }
              }
            },
            "agent": {
              "properties": {
                "id": { "type": "keyword", "ignore_above": 1024 },
                "statistics": {
                  "properties": {
                    "agent": {
                      "properties": {
                        "status":         { "type": "keyword", "ignore_above": 1024 },
                        "last_keepalive": { "type": "date" },
                        "messages": {
                          "properties": {
                            "count": { "type": "long" }
                          }
                        },
                        "tasks": {
                          "properties": {
                            "dispatched":          { "properties": { "total": { "type": "long" } } },
                            "discarded_duplicate": { "properties": { "total": { "type": "long" } } },
                            "failed":              { "properties": { "total": { "type": "long" } } }
                          }
                        }
                      }
                    },
                    "logcollector": {
                      "properties": {
                        "global": {
                          "properties": {
                            "start": { "type": "date" },
                            "end":   { "type": "date" },
                            "files": {
                              "properties": {
                                "location": { "type": "keyword", "ignore_above": 1024 },
                                "events":   { "type": "long" },
                                "bytes":    { "type": "long" },
                                "targets": {
                                  "properties": {
                                    "name":  { "type": "keyword", "ignore_above": 1024 },
                                    "drops": { "type": "long" }
                                  }
                                }
                              }
                            }
                          }
                        },
                        "interval": {
                          "properties": {
                            "start": { "type": "date" },
                            "end":   { "type": "date" },
                            "files": {
                              "properties": {
                                "location": { "type": "keyword", "ignore_above": 1024 },
                                "events":   { "type": "long" },
                                "bytes":    { "type": "long" },
                                "targets": {
                                  "properties": {
                                    "name":  { "type": "keyword", "ignore_above": 1024 },
                                    "drops": { "type": "long" }
                                  }
                                }
                              }
                            }
                          }
                        }
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

## Fields

`index-definitions/wazuh-agent-stats.fields.csv` carries the same list with examples and descriptions. 26 declared fields.

| Field | Type | Example |
|---|---|---|
| `state.modified_at` | date | `2026-08-02T10:07:12.431Z` |
| `state.document_version` | integer | `1` |
| `wazuh.schema.version` | keyword | `1` |
| `wazuh.cluster.name` / `.node` | keyword | `wazuh` / `node01` |
| `wazuh.agent.id` | keyword | `001` |
| `wazuh.agent.statistics.agent.status` | keyword | `connected` |
| `wazuh.agent.statistics.agent.last_keepalive` | date | `2026-08-02T10:06:50Z` |
| `wazuh.agent.statistics.agent.messages.count` | long | `602` |
| `wazuh.agent.statistics.agent.tasks.dispatched.total` | long | `4` |
| `wazuh.agent.statistics.agent.tasks.discarded_duplicate.total` | long | `0` |
| `wazuh.agent.statistics.agent.tasks.failed.total` | long | `0` |
| `wazuh.agent.statistics.logcollector.global.start` / `.end` | date | `2026-08-02T09:43:50Z` |
| `wazuh.agent.statistics.logcollector.global.files` | array of objects | |
| `wazuh.agent.statistics.logcollector.global.files.location` | keyword | `/var/log/dpkg.log` |
| `wazuh.agent.statistics.logcollector.global.files.events` | long | `32` |
| `wazuh.agent.statistics.logcollector.global.files.bytes` | long | `3436` |
| `wazuh.agent.statistics.logcollector.global.files.targets` | array of objects | |
| `wazuh.agent.statistics.logcollector.global.files.targets.name` | keyword | `agent` |
| `wazuh.agent.statistics.logcollector.global.files.targets.drops` | long | `0` |
| `wazuh.agent.statistics.logcollector.interval.*` | | Same shape as `global`, for the last closed window. Counters reset after each snapshot, so it is a sample and not a series |

jotacarma90 (approved):

PR #38151, opened 2026-08-03
join module worker thread in stop_wmodules() before reporting stopped

### LINKED ISSUE #38111: Access violation in sysinfo.dll on agent shutdown breaking service restart
| Wazuh version | Component | Packages |
|---|---|---|
| 5.0.0-rc1 | WIndows Server 2012r2 AMD64 | 29/07 nightly packages |

During [Cannot stop WazuhSvc in DIT restart test](https://github.com/wazuh/wazuh-qa-automation/issues/8381), 

Every time the agent shuts down it raises an access violation inside `sysinfo.dll`, always at the same offset. The process dies at the point where it should be completing its exit, right after the syscollector teardown.

Most of the time this is invisible: the process manages to report its stop to the SCM before dying, so the restart completes and the only trace is a crash event in the Windows Application log. When the crash lands before the stop is reported, the SCM loses the control channel to the process and `Restart-Service` fails with `ERROR_BROKEN_PIPE (109)`, leaving the agent down.

So the crash is systematic, only its visibility is intermittent.

### Crash 

```
Faulting application name: wazuh-agent.exe, version: 5.0.0.0, time stamp: 0x6a695469
Faulting module name: sysinfo.dll, version: 5.0.0.0, time stamp: 0x6a6953e9
Exception code: 0xc0000005
Fault offset: 0x0000ca90
Faulting process id: 0x974
Faulting module path: C:\Program Files (x86)\ossec-agent\sysinfo.dll
```

Retrieved with:

```
Get-EventLog -LogName Application -Newest 60 | Where-Object { $_.Message -match 'wazuh' } | ForEach-Object { "$($_.TimeGenerated) src=$($_.Source) id=$($_.EventID)`n$($_.Message)`n" }
```

At least 15 Application Error (id 1000) events between 10:18:44 and 11:45:05, all with an identical signature — same module, same exception code, same 0x0000ca90 offset. A constant offset means the same faulting instruction every time, so it should resolve to an exact source location with the `sysinfo.dll PDB`.

### The crash also occurs on restarts that succeed

Correlating the faulting PID from each crash event with `ossec.log`:

| Crash (WER) | PID | ossec.log for that PID | Restart outcome |
|---|---|---|---|
| 10:48:01 | 0xbb0 = 2992 | exit signal 10:48:00 → Module finished 10:48:01 → no Exit completed successfully | failed (109) |
| 10:48:10 | 0x260 = 608	 | exit signal 10:48:08 → Module finished 10:48:10 → no Exit completed successfully | failed (109) |
| 10:51:19 | 0x974 = 2420	 | exit signal 10:51:17 → Module finished 10:51:19 → Exit completed successfully 10:51:19 | OK |
| 11:45:05 | 0x90 = 144	| exit signal 11:45:03 → Module finished 11:45:05 → Exit completed successfully 11:45:05 | OK |

Processes 2420 and 144 wrote `Exit completed successfully`, the SCM logged its normal stopped + running pair, and `Restart-Service` returned without error — and both crashed in `sysinfo.dll` anyway.

The relevant ossec.log excerpts (obtained with `Get-Content -LiteralPath 'C:\Program Files (x86)\ossec-agent\ossec.log'`):

Shutdown that completed before the crash (pid 144):

```
2026/07/30 11:45:02 wazuh-agent: INFO: Started (pid: 144).
2026/07/30 11:45:03 wazuh-agent: INFO: Received exit signal. Starting exit process.
2026/07/30 11:45:05 wazuh-modulesd:syscollector: INFO: Module finished.
2026/07/30 11:45:05 wazuh-agent: INFO: Exit completed successfully.
2026/07/30 11:45:06 wazuh-agent: INFO: Started (pid: 1092).
```

Shutdown where the crash won the race (pid 608):

```
2026/07/30 11:45:07 wazuh-agent: INFO: Started (pid: 1092).
2026/07/30 11:45:08 wazuh-agent: INFO: Received exit signal. Starting exit process.
2026/07/30 11:45:10 wazuh-modulesd:syscollector: INFO: Module finished.
                    <-- "Exit completed successfully" never written
```

In both cases the crash timestamp lands in the same second as `syscollector: Module finished`.

### What it looks like from the outside

Windows classifies it as an unexpected termination:

```
Get-EventLog -LogName System -Source 'Service Control Manager' -Newest 40 | Where-Object { $_.Message -match 'wazuh' } | ForEach-Object { "$($_.TimeGenerated) id=$($_.EventID) $($_.Message)" }

07/30/2026 12:05:58 id=7034 The Wazuh service terminated unexpectedly.  It has done this 4 time(s).
07/30/2026 12:05:51 id=7036 The Wazuh service entered the running state.
07/30/2026 12:00:20 id=7034 The Wazuh service terminated unexpectedly.  It has done this 3 time(s).
07/30/2026 12:00:15 id=7036 The Wazuh service entered the running state.
07/30/2026 11:45:10 id=7034 The Wazuh service terminated unexpectedly.  It has done this 2 time(s).
07/30/2026 11:45:05 id=7036 The Wazuh service entered the running state.
07/30/2026 11:45:05 id=7036 The Wazuh service entered the stopped state.
07/30/2026 11:45:00 id=7036 The Wazuh service entered the running state.
07/30/2026 11:41:35 id=7034 The Wazuh service terminated unexpectedly.  It has done this 1 time(s).
```

Note 11:45:05, the cycle of pid 144: it produced the normal stopped + running pair, so the SCM was satisfied — yet the process crashed. The 7034 at 11:45:10 is the next cycle, the one that failed.

And when the crash beats the stop report, the caller gets:

```
$ProgressPreference='SilentlyContinue';"t0=$((Get-Date).ToString('HH:mm:ss.fff'))"; Restart-Service -Name wazuh; Start-Sleep -Seconds 3; $b=@(Get-Process -Name 'wazuh-agent' -ErrorAction SilentlyContinue | ForEach-Object { $_.Id }) -join ','; "t1=$((Get-Date).ToString('HH:mm:ss.fff')) procs_before=[$b]"; (cmd /c sc queryex WazuhSvc) | Where-Object { $_ -match 'STATE|PID' }; try { Restart-Service -Name wazuh -ErrorAction Stop; 'SECOND RESTART OK' } catch { $e=$_.Exception; while ($e) { "[$($e.GetType().FullName)] $($e.Message)"; if ($e -is [System.ComponentModel.Win32Exception]) { "NativeErrorCode=$($e.NativeErrorCode)" }; $e=$e.InnerException } }; $a=@(Get-Process -Name 'wazuh-agent' -ErrorAction SilentlyContinue | ForEach-Object { $_.Id }) -join ','; "t2=$((Get-Date).ToString('HH:mm:ss.fff')) procs_after=[$a]"; (cmd /c sc queryex WazuhSvc) | Where-Object { $_ -match 'STATE|PID' }

t0=12:00:14.861
t1=12:00:18.073 procs_before=[368]
        STATE              : 4  RUNNING
        PID                : 368
[Microsoft.PowerShell.Commands.ServiceCommandException] Service 'wazuh (WazuhSvc)' cannot be stopped due to the following error: Cannot stop WazuhSvc service on computer '.'.
[System.InvalidOperationException] Cannot stop WazuhSvc service on computer '.'.
[System.ComponentModel.Win32Exception] The pipe has been ended
NativeErrorCode=109
t2=12:00:20.901 procs_after=[2336]
        STATE              : 1  STOPPED
        PID                : 0
```

Worth noting the state before the failing stop: the SCM reports `RUNNING` with PID 368 and exactly one process exists with that same PID. The service was in a fully consistent state — the failure is not caused by leftover state from a previous cycle. The `ERROR_BROKEN_PIPE` is the SCM finding the process gone mid-request.

`Restart-Service` aborts on the failed stop and never reaches the start step, so the agent is left down.

### Reproduction

Two restarts 3 seconds apart. Failure rate ~1 in 3; the crash itself occurs on every shutdown, visible in the Application log:

```
$ProgressPreference='SilentlyContinue';Restart-Service -Name wazuh; Start-Sleep -Seconds 3; try { Restart-Service -Name wazuh -ErrorAction Stop; 'SECOND RESTART OK' } catch { $e=$_.Exception; while ($e) { "[$($e.GetType().FullName)] $($e.Message)"; if ($e -is [System.ComponentModel.Win32Exception]) { "NativeErrorCode=$($e.NativeErrorCode)" }; $e=$e.InnerException } }
```

For reference: with 45 seconds between restarts, 0 failures in 20 attempts; with 3 seconds, ~1 in 3. The original CI failure occurred with the agent at ~90 s uptime and fully initialised, so short spacing amplifies the rate but is not a necessary condition.

### Suggestion

The crash is in `sysinfo.dll` during shutdown, immediately after `syscollector: Module finished`, at a constant offset (+0xca90) — so the syscollector/sysinfo teardown path. With the matching PDB the offset should point straight at the line.

Independently of the crash fix, the service control path is worth hardening: reporting `SERVICE_STOP_PENDING` with a dwWaitHint during shutdown and `SERVICE_STOPPED` as the last action before exiting would keep the SCM consistent. It would not eliminate the crash, but it would stop it from turning into a failed restart with the agent left down.

Full investigation: https://github.com/wazuh/wazuh-qa-automation/issues/8381

hernanvalenzuela:
## Description
On Windows, `stop_wmodules()` never waited for a module's own worker thread to finish before the SCM was told the service had stopped, and `wm_sys_stop()`'s internal 10-second wait didn't gate anything by itself — it just gave up and let shutdown continue regardless of whether the thread was still running. If syscollector's worker thread was still busy (e.g. mid-scan, which has no interruption checkpoint once started, on top of a startup gate that could block indefinitely without ever checking for a pending shutdown), the process could report `SERVICE_STOPPED` and begin exiting while that thread was still executing inside `sysinfo.dll`, producing the reported intermittent access violation and leaving the agent unable to restart cleanly.

## Proposed Changes

Make Windows shutdown actually wait for each module's worker thread to exit before the service reports itself stopped, and remove the specific gaps that let a worker thread outlive that wait:

- **Join the module thread in `stop_wmodules()`** (`src/win32/win_utils.c`, `src/wazuh_modules/include/wmodules_def.h`): capture the thread handle that `w_create_thread()` already returned but discarded, and wait on it (bounded, 20s) after calling each module's `stop()`. This is the core fix for the race described above.
- **Make the startup gate shutdown-aware** (`src/shared/src/startup_gate_op.c`): `startup_gate_wait_for_ready()` blocks a module's thread until it completes a handshake with the manager, but never checked for a pending shutdown. With an unreachable manager, that wait never ends, so the new join above could hit its full timeout instead of returning promptly — this affects any module gated this way, not just syscollector. The gate now checks `wm_shutdown_requested` and releases immediately once shutdown has started.
- **Restore syscollector's per-task stop checkpoint** (`src/wazuh_modules/syscollector/src/syscollectorImp.cpp`): `TRY_CATCH_TASK` had its `m_stopping` check removed by a prior commit, so once a scan starts, all 11 collection tasks now run to completion regardless of a pending stop, widening the window during which the worker thread can still be executing inside `sysinfo.dll` when shutdown proceeds. Restoring the check bounds that window back down to a single task instead of the whole scan.

#### Mechanism, reproduced deterministically (debug-only instrumentation, not merged)

Using a temporary forced delay inside `sysinfo.dll`'s `getHardware()` (busy-loop, so the thread is genuinely executing inside the DLL, not just parked in a kernel wait) and a temporary bypass of the agentd document-limits handshake (so the module reaches `scan()` without needing a live manager):

| | Before fix | After fix |
|---|---|---|
| Process alive after service reports "stopped" | **True — 24/24 attempts** (sleep-delay and busy-loop variants, `net stop` and `taskkill /F`, with and without CPU stress) | **False — 8/8 attempts** (same variants, under CPU stress) |
| Internal 10s timeout message (`"did not confirm shutdown within 10 seconds..."`) | Fires every time | Never fires |
| 20s join timeout message (`"worker thread did not exit within 20000 ms..."`) | N/A (join didn't exist) | Never fires — join always completes within its bound |

### Timestamped timeline (3 clean "after" cycles, real log)

| Cycle | Exit signal | Stop received | Module/Evaluation finished | Real wait |
|---|---|---|---|---|
| A | 11:25:41 | 11:25:42 | 11:26:08 | 27s |
| B | 11:26:29 | 11:26:30 | 11:26:56-57 | 27-28s |
| C | 11:27:17 | 11:27:19 | 11:27:44 | 25-27s |

Before the fix, this same scenario always cut off at a fixed 10s regardless of real thread state; after the fix, the process waits however long the thread genuinely needs (here 25-28s), within the join's 20s margin, and never hit that margin's timeout in these runs.

#### Binary validation

Confirmed via string search directly on the installed binaries that each build actually contained the expected code before running each test round (e.g. `"worker thread did not exit within"` and `"shutdown requested, releasing"` present in `wazuh-agent.exe` for "after" runs; absent for "before" runs built from the pre-fix source).

#### Known limitations

- Could not reproduce the exact `0xc0000005` exception signature reported in the original issue via either self-directed `ExitProcess` (Windows safely quiesces other threads before unmapping DLLs) or external `TerminateProcess`/`taskkill /F` (kills atomically, no chance to fault). The evidence above demonstrates the underlying defect (premature "stopped" declaration) directly rather than the crash itself.
- Two findings surfaced during this investigation are tracked separately, out of scope for this PR: FIM has no thread join at all on Windows shutdown, and SCA's `releaseResources()` doesn't reset `m_syncManager`, leaking the DBSync handle.

  - [x] Linux
  - [x] Windows 
  - [x] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

### Artifacts Affected

N-A

### Configuration Changes

N-A

### Tests Introduced

N-A

## Review Checklist

- [x] Code changes reviewed
- [ ] ...

### DIFF
--- a/src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp
+++ b/src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp
@@ -523,7 +523,7 @@ TEST_F(SyscollectorImpTest, defaultCtor)
                                           SYSCOLLECTOR_DB_PATH,
                                           "",
                                           "",
-                                          5, true, true, true, true, true, true, true, true, true, true, true, true, true, true);
+                                          3600, true, true, true, true, true, true, true, true, true, true, true, true, true, true);
 
             Syscollector::instance().start();
         }
@@ -1204,7 +1204,7 @@ TEST_F(SyscollectorImpTest, noPorts)
                                           SYSCOLLECTOR_DB_PATH,
                                           "",
                                           "",
-                                          5, true, true, true, true, true, false, true, true, true, true, true, true, true, true);
+                                          3600, true, true, true, true, true, false, true, true, true, true, true, true, true, true);
 
             Syscollector::instance().start();
         }
@@ -2627,7 +2627,7 @@ TEST_F(SyscollectorImpTest, sanitizeJsonValues)
                                           SYSCOLLECTOR_DB_PATH,
                                           "",
                                           "",
-                                          5, true, true, true, true, true, true, true, true, true, true, true, true, true, true);
+                                          3600, true, true, true, true, true, true, true, true, true, true, true, true, true, true);
 
             Syscollector::instance().start();
         }
@@ -4717,7 +4717,7 @@ TEST_F(SyscollectorImpTest, schemaValidationAcceptsValidDataAfterCorrections)
                                           SYSCOLLECTOR_DB_PATH,
                                           "",
                                           "",
-                                          5, true, true, false, false, false, false, false, false, false, false, false, false, false, false);
+                                          3600, true, true, false, false, false, false, false, false, false, false, false, false, false, false);
 
             // Initialize sync protocol to enable schema validation
             MQ_Functions mqFuncs;
@@ -4824,7 +4824,7 @@ TEST_F(SyscollectorImpTest, schemaValidationWithCorrectedDataTypes)
                                           SYSCOLLECTOR_DB_PATH,
                                           "",
                                           "",
-                                          5, true, true, false, false, false, true, false, false, false, false, false, false, false, false);
+                                          3600, true, true, false, false, false, true, false, false, false, false, false, false, false, false);
 
             // Initialize sync protocol to enable schema validation
             MQ_Functions mqFuncs;
@@ -4918,7 +4918,7 @@ TEST_F(SyscollectorImpTest, hardwareCpuSpeedZeroIsReportedAsNull)
                                           SYSCOLLECTOR_DB_PATH,
                                           "",
                                           "",
-                                          5, true, true, false, false, false, false, false, false, false, false, false, false, false, false);
+                                          3600, true, true, false, false, false, false, false, false, false, false, false, false, false, false);
 
             // Initialize sync protocol to enable schema validation
             MQ_Functions mqFuncs;
@@ -5127,7 +5127,7 @@ TEST_F(SyscollectorImpTest, schemaValidationDiscardsInvalidMtuValueOnWindows)
                                           SYSCOLLECTOR_DB_PATH,
                                           "",
                                           "",
-                                          5, true, false, false, true, false, false, false, false, false, false, false, false, false, false);
+                                          3600, true, false, false, true, false, false, false, false, false, false, false, false, false, false);
 
             // Initialize sync protocol to enable schema validation
             MQ_Functions mqFuncs;
@@ -5225,7 +5225,7 @@ TEST_F(SyscollectorImpTest, schemaValidationDiscardsInvalidMtuValueOnUnix)
                                           SYSCOLLECTOR_DB_PATH,
                                           "",
                                           "",
-                                          5, true, false, false, true, false, false, false, false, false, false, false, false, false, false);
+                                          3600, true, false, false, true, false, false, false, false, false, false, false, false, false, false);
 
             // Initialize sync protocol to enable schema validation
             MQ_Functions mqFuncs;
@@ -5338,7 +5338,7 @@ TEST_F(SyscollectorImpTest, schemaValidationRejectsInvalidDataWithMock)
                                           SYSCOLLECTOR_DB_PATH,
                                           "",
                                           "",
-                                          5, true, true, false, false, false, false, false, false, false, false, false, false, false, false);
+                                          3600, true, true, false, false, false, false, false, false, false, false, false, false, false, false);
 
             // Reset and initialize factory with mock BEFORE initSyncProtocol
             SchemaValidator::SchemaValidatorFactory::getInstance().reset();
@@ -5478,7 +5478,7 @@ TEST_F(SyscollectorImpTest, schemaValidationDiscardsWhenValidatorNotFound)
                                           SYSCOLLECTOR_DB_PATH,
                                           "",
                                           "",
-                                          5, true, true, false, false, false, false, false, false, false, false, false, false, false, false);
+                                          3600, true, true, false, false, false, false, false, false, false, false, false, false, false, false);
 
             // Reset and initialize factory with mock for DIFFERENT index
             SchemaValidator::SchemaValidatorFactory::getInstance().reset();
--- a/src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp
+++ b/src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp
@@ -284,14 +284,21 @@ void logFunction(const modules_log_level_t /*level*/, const std::string& /*log*/
 // tasks (see #38111), so a scan interrupted by an early destroy() would skip whichever
 // collectors hadn't run yet, failing their EXPECT_CALL(...).Times(1) expectations.
 //
-// The short fixed sleep at the start is a grace period for the freshly-spawned thread to
-// reach scan() and set m_scanning=true - it covers thread-scheduling latency, not scan
-// duration, so it doesn't reintroduce the race it's meant to avoid.
-void waitForScanToFinish(std::chrono::milliseconds maxWait = std::chrono::seconds(10))
+// Waits for isScanning() to become true BEFORE waiting for it to become false. A fixed
+// grace period before the first check isn't enough: init() itself (real SQLite/DBSync
+// setup) can take longer than any short grace period under Valgrind, so isScanning()
+// can still read false while init() is simply still in progress, not because the scan
+// already finished. Treating that as "done" and calling destroy() early races
+// releaseResources() against the other thread's still-running init()/start(), which
+// segfaults on a null m_spInfo once that thread reaches scanHardware().
+void waitForScanToFinish(std::chrono::milliseconds maxWait = std::chrono::seconds(30))
 {
-    std::this_thread::sleep_for(std::chrono::milliseconds(200));
-
     const auto deadline = std::chrono::steady_clock::now() + maxWait;
+
+    while (!Syscollector::instance().isScanning() && std::chrono::steady_clock::now() < deadline)
+    {
+        std::this_thread::sleep_for(std::chrono::milliseconds(20));
+    }
 
     while (Syscollector::instance().isScanning() && std::chrono::steady_clock::now() < deadline)
     {
--- a/src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp
+++ b/src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp
@@ -278,34 +278,6 @@ void logFunction(const modules_log_level_t /*level*/, const std::string& /*log*/
     // std::cout << s_logStringMap.at(level) << ": " << log << std::endl;
 }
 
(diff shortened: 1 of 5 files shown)


src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp
lchico, line 1: The sleep_for bumps here (1-2s → 5s across ~33 tests, +107s aggregate — this whole file runs as one serial ctest entry) are a reasonable pragmatic fix for flakiness from the restored m_stopping check, and I wouldn't block this PR on it. Worth a follow-up ticket to replace the fixed sleeps with deterministic synchronization (e.g. the std::promise/std::future pattern already used in the new stoppingDuringScanSkipsRemainingTasks test) instead of guessing a duration faster on fast machines, more reliable on slow ones.

lchico (changes requested):

src/win32/win_utils.c
```diff
@@ -109,11 +109,38 @@ void stop_wmodules()
...
+    // syscollector's 10 s), so this join is the actual gate and not a race against it.
+    // Shared across every module in the loop below (not reset per module), so total
+    // time spent in stop_wmodules() is bounded regardless of how many wodles are
+    // configured -- otherwise N modules could each burn a fresh timeout and the sum
+    // could outlast the SCM's own patience (WaitToKillServiceTimeout).
+    const DWORD MODULE_JOIN_BUDGET_MS = 20000;
+    const ULONGLONG budget_start = GetTickCount64();
+
     wmodule * cur_module;
     for (cur_module = wmodules; cur_module; cur_module = cur_module->next) {
```
lchico, line 122: This calls stop() and immediately joins before moving to the next module, so a slow module at the front of the list can consume the whole 20s budget before later modules even get stop() called. Compare with the POSIX version of this exact fix in wazuh_modules/src/main.c:270-308, which does two passes: signal every module's stop() first, then join each thread against the shared deadline. Suggest splitting this into two loops the same way — otherwise the fix's guarantee degrades for anything after the first module in wmodules.

hernanvalenzuela:
the failing test is unrelated to my changes, there is a issue link to  https://github.com/wazuh/wazuh/issues/38203

src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp
hernanvalenzuela: A solution will evaluate in https://github.com/wazuh/wazuh/issues/38221 and https://github.com/wazuh/wazuh/issues/38222

src/win32/win_utils.c
hernanvalenzuela: A solution will evaluate in https://github.com/wazuh/wazuh/issues/38221 and https://github.com/wazuh/wazuh/issues/38222

cborla:
❗ The failing test is unrelated to my changes, so I created the following issue: https://github.com/wazuh/wazuh/issues/38203

cborla (approved):
LGTM
- [x] Code changes reviewed

lchico (approved):
LGTM!

- [x]  Code changes reviewed

PR #37030, opened 2026-06-20
Base scaffolding for the container_images module

anromerom:
## Description

Closes #36994

Adds the base scaffolding for the new `container_images` agent module. This first stage provides the minimum structure to continue development in follow-up PRs: a C++ module (built and wired into the agent like SCA and Syscollector), `ossec.conf` configuration parsing, a clean start/stop lifecycle, and an initial local image reader that discovers on-disk container image references. Remote registries, Kubernetes, Vulnerability Detector, indexing, synchronization, and package extraction are intentionally out of scope for this issue.

## Proposed Changes

- New agent-only C++ module under `wazuh_modules/container_images/`, following the SCA / Syscollector two-half pattern: a thin C glue layer (`wm_container_images.c`) that loads a C++ shared library, plus a `container_images_impl` library with the real logic.
- C ABI + facade (`container_images.h` / `.hpp` / `.cpp`) exposing `init` / `start` / `stop` / `release_resources` / `set_log_function`, resolved by the glue via `sym_load`.
- Lifecycle through the standard `wm_context` (start / stop / destroy / dump); an interruptible scan loop owns `scan_on_start` + `interval`.
- Internal configuration model and a source-agnostic reader interface (`IImageReader`) with one concrete implementation (`LocalImageReader`) that parses an on-disk **OCI image layout** (no daemon, no external dependencies) and reads each image's basic metadata (config digest, platform, tag). The per-reference debug log includes the **config digest**.
- Per-path **format detection**: an OCI layout is read; any other on-disk format (`docker-archive`, `containerd`, `unknown`) is detected and logged as `WARNING: NOT IMPLEMENTED: local format '<guess>' ...` and skipped.
- `ossec.conf` parsing of the `<container_images>` block (`enabled`, `scan_on_start`, `interval`, and a `<references>` block whose `<local>` elements each hold a local image path), following the existing module configuration style. Non-`local` reference types are accepted but reported as "not supported yet".
- Build-system wiring: module added to the agent build, excluded from the server build (like SCA), and the shared library installed into the agent `lib` directory by the installer.

### Results and Evidence

Run scanning two local OCI layouts (alpine + busybox), debug enabled. The per-reference debug line reports the image's **config digest** read from the OCI configuration blob:

```sql
wazuh-modulesd:container_images: DEBUG: Module initialized.
wazuh-modulesd:container_images: DEBUG: Scan on start.
wazuh-modulesd:container_images: INFO: Scan started.
[...]

PR #38152, opened 2026-08-03
Package inventory tables for container images

### LINKED ISSUE #37529: Implement package inventory tables for container images
# Implement package inventory tables for container images

## Description

Implement the initial package inventory tables required by the `container_images` module to persist packages found inside scanned container images.

This issue is a follow-up / sibling of storage model required to represent package inventory per image. The goal is to prepare the state layer so future issues can plug in the package extractor, synchronization flow, events, and Vulnerability Detector integration.

Parent objective: #35989  
Related scaffolding issue: #36994

## Scope

- Define and implement the tables required to store package inventory for container images.
- Keep the container image package inventory separated from the host package inventory.
- Add the required relationship between a container image and its discovered packages.
- Include enough image metadata in the package state to identify the origin of each package:
  - image ID / digest
  - image name
  - image tags
  - image source type
  - image source path when applicable
- Store package metadata required by the inventory and future VD integration:
  - name
  - version
  - architecture
  - type
  - source
  - vendor
  - description
  - category
  - size
  - path
  - installed date when available
  - checksum / hash fields when applicable
- Define the primary key / unique identifier for package records, considering both package data and image identity.
- Add the required schema initialization and migration logic.
- Add the required cleanup/drop logic if applicable.
- Add unit tests for schema creation and basic insert/update/delete operations.
- Document any pending fields or schema decisions that must be completed by follow-up issues.

## Proposed table model

The final table names should follow the existing inventory/dbsync conventions used in the agent.

As an initial proposal, the implementation should include tables equivalent to:

- `container_images`
- `container_image_packages`

The package table must not collide with the existing host package inventory table.

A package record must be uniquely associated with the container image where it was found. The same package installed in two different images must be represented as two different inventory records.

## Out of scope

- Package extraction from image layers.
- OCI layer parsing.
- Local image discovery.
- Remote registry support.
- GHCR or ECR authentication.
- Vulnerability Detector integration.
- Dashboard changes.
- Final event formatting.
- Final synchronization/indexing behavior, unless a minimal placeholder is required to validate the table model.
- Capacity limit enforcement.

## Tasks

- [ ] Review the existing package inventory table implementation used by the agent.
- [ ] Define the table schema for container image package inventory.
- [ ] Define the image/package relationship model.
- [ ] Define the unique package key including image identity.
- [ ] Implement schema creation for the new tables.
- [ ] Implement schema cleanup/drop helpers if required.
- [ ] Add basic storage helpers for package records.
- [ ] Add tests for schema creation.
- [ ] Add tests for package insert/update/delete flows.
- [ ] Add tests to verify that packages from different images do not overwrite each other.
- [ ] Add tests to verify that container image packages do not collide with host packages.
- [ ] Document pending fields or schema decisions for future package extractor and VD integration issues.

## Acceptance criteria

- Container image package inventory has dedicated tables.
- Package records are associated with their source container image.
- The same package in different images can be stored independently.
- The implementation does not reuse or overwrite the host package inventory table.
- The schema supports the fields required by the parent objective for future state synchronization and VD integration.
- Basic insert, update, delete, and lookup operations are covered by tests.
- The agent builds successfully with the new schema changes.
- No package extraction, remote registry support, VD integration, dashboard changes, or final event pipeline logic is introduced in this issue.

anromerom:
## Description

Closes #37529

The `container_images` module needs somewhere to keep the packages it finds inside a scanned image. This PR adds that storage layer: two DBSync tables, change detection across scans, schema versioning, and a cleanup entry point.

Package extraction from image layers stays out of scope, as the issue specifies. A temporary in-memory reader feeds a fixed inventory that mutates across scans, so all three delta types can be exercised before real extraction exists. Manager synchronization is also out of scope here and is the subject of a follow-up issue.

## Proposed Changes

- Add two DBSync tables: `dbsync_container_image_references` and `dbsync_container_image_packages`.
- Persist the scan result through a DBSync transaction per table and report create/modify/delete deltas from the diff against the stored state.
- Add schema versioning through DBSync's `upgradeStatements` mechanism, so future schema changes migrate an existing agent database rather than recreating it.
- Add `ContainerImagesDB::dropTables()` and `ContainerImagesImpl::clearInventory()` to clear the inventory this module owns without dropping the schema.
- Add a temporary in-memory reader behind the existing `makeReader` factory seam so the storage layer can be validated before extraction lands.

### Data model

Packages belong to the image **reference** they were found under. The image content digest is stored as metadata, not as identity. This follows the schema analysis carried out for the parent objective: a rebuilt tag then reports only the packages that actually changed, instead of the whole inventory being deleted and re-added.

`dbsync_container_image_references`, primary key `(reference_type, reference_value)`:

```sql
CREATE TABLE IF NOT EXISTS dbsync_container_image_references (
    reference_type TEXT,
    reference_value TEXT,
    image_config_digest TEXT,
    manifest_digest TEXT,
    image_name TEXT,
    tag TEXT,
    tags TEXT,
    platform_os TEXT,
    platform_architecture TEXT,
    platform_variant TEXT,
    platform_os_version TEXT,
    checksum TEXT,
    version INTEGER NOT NULL DEFAULT 1,
    sync INTEGER NOT NULL DEFAULT 0,
    PRIMARY KEY (reference_type, reference_value)) WITHOUT ROWID;
```

`dbsync_container_image_packages`, primary key `(reference_type, reference_value, name, version_, architecture, type)`:

```sql
CREATE TABLE IF NOT EXISTS dbsync_container_image_packages (
    reference_type TEXT,
    reference_value TEXT,
    name TEXT,
    version_ TEXT,
    architecture TEXT,
    type TEXT,
    vendor TEXT,
    installed TEXT,
    path TEXT,
    category TEXT,
    description TEXT,
    size BIGINT,
    priority TEXT,
    multiarch TEXT,
    source TEXT,
    package_db_path TEXT,
    checksum TEXT,
    version INTEGER NOT NULL DEFAULT 1,
    sync INTEGER NOT NULL DEFAULT 0,
    PRIMARY KEY (reference_type, reference_value, name, version_, architecture, type)) WITHOUT ROWID;
```

The package columns mirror the host package inventory (`dbsync_packages`) so the same indexer mapping can be reused later, with two differences. The owning reference replaces `path` in the primary key, because an image inventory has no host install path, so `path` becomes an attribute populated only when the package database records one. `package_db_path` is added to record which package database inside the image a row came from.

Two points worth a reviewer's attention:

- The table name proposed in the issue was `container_images` + `container_image_packages`, keyed by image. The implementation uses references + packages, keyed by reference, per the schema decision taken for the parent objective. The issue text should be updated to match.
- The `sync` column is present but unused at this stage. It exists so a document limit can be layered on later without a migration.

## Results and Evidence

Agent build with tests:

```sql
Done building agent
```

Unit tests:

```sql
[==========] 19 tests from 2 test suites ran. (44 ms total)
[  PASSED  ] 19 tests.
```

The new cases, all passing:

```sql
[       OK ] ContainerImagesDBTest.SchemaDoesNotCollideWithHostPackageInventory (0 ms)
[       OK ] ContainerImagesDBTest.SamePackageInTwoImagesIsStoredIndependently (1 ms)
[       OK ] ContainerImagesDBTest.PackageAttributeFieldsArePersisted (1 ms)
[       OK ] ContainerImagesDBTest.DropTablesClearsTheInventory (1 ms)
[       OK ] ContainerImagesDBTest.ReopeningAPersistedDatabaseKeepsItsRows (10 ms)
[       OK ] ContainerImagesDBTest.UpgradeStatementsDefineTheSchemaVersion (0 ms)
```

Delta behavior across scans, covered by the pre-existing cases and unchanged by this PR: a first scan reports every row as created, an unchanged rescan reports nothing, an attribute change reports a modification, and a change to a primary-key field such as a package version reports a delete plus a create. That last one matches how the host package inventory behaves and is worth stating explicitly, since a version bump is not reported as a modification.

### Artifacts Affected

- `libcontainer_images.so`, which now links `dbsync`.
- Linux agent packages. The module is agent-only and excluded from the server build.
- New agent directory `queue/container_images/db`, created by the installer.

### Configuration Changes

None. No new configuration options, no changed defaults. The database path is internal and not configurable.

### Documentation Updates

- Added `docs/ref/modules/container-images/persistence.md` describing the schema, change detection, schema versioning, and cleanup.
- Updated the module `README.md`, `architecture.md`, and `api-reference.md` status notes, which previously stated that persistence was not implemented.

### Tests Introduced

Six unit tests in `container_images_test.cpp`:

- `SchemaDoesNotCollideWithHostPackageInventory` asserts the create statement never touches the host `dbsync_packages` table.
- `SamePackageInTwoImagesIsStoredIndependently` stores byte-identical package records under two references and asserts four rows, then removes one reference and asserts the other reference's copy survives.
- `PackageAttributeFieldsArePersisted` asserts every attribute column round-trips, including the fields added by this PR.
- `DropTablesClearsTheInventory` asserts the same inventory is reported as new after a clear.
- `ReopeningAPersistedDatabaseKeepsItsRows` asserts a reopen at the same schema version does not recreate the database.
- `UpgradeStatementsDefineTheSchemaVersion` pins the upgrade-list contract that DBSync derives the schema version from.

## Review Checklist

- [x] Code changes reviewed

Note on the unchecked item: this PR is stacked on #37030 and must merge after it.


cborla (approved):
LGTM

Issue #38091, opened 2026-07-29
Require TLS 1.3 for wazuh-authd


juliancnn:
## Description

`wazuh-authd`, responsible for agent enrollment, currently accepts TLS 1.2 as the minimum protocol version.

Update the enrollment service so the minimum supported TLS version is TLS 1.3.

The OpenSSL cipher configuration must also be updated. The current configuration applies a TLS 1.2 cipher list, while TLS 1.3 cipher suites require their corresponding OpenSSL configuration API.

## Requirements

* Set TLS 1.3 as the minimum protocol version accepted by `wazuh-authd`.
* Reject clients attempting to connect using TLS 1.2 or older.
* Update the cipher configuration to accept and apply a configurable TLS 1.3 cipher-suite list.
* Validate the configured cipher-suite list during startup.
* Report a clear error when the configuration is invalid.
* Update the default configuration and documentation.
* Verify compatibility with the agent enrollment client.

## Deliverables

* TLS 1.3 enforced in `wazuh-authd`.
* TLS 1.3 cipher suites correctly configured through OpenSSL.
* Tests covering accepted TLS 1.3 connections and rejected TLS 1.2 connections.

matigarciadev:
### Update Aug 3

**TLS 1.3 enforcement:**
- Raised wazuh-authd's minimum TLS version to 1.3, switched cipher config to the TLS 1.3 ciphersuite API (SSL_CTX_set_ciphersuites), added startup validation with clear errors.
- Updated config defaults, CLI (-c flag), and docs.
- Built local agent .deb/.rpm packages (amd64/x86_64) with the change, to run against the e2e docker-compose test harness.

**New, aligned follow-up, cert ownership consistency:**
- Updated packaging (RPM spec, DEB postinst, source installer) so authd's certs get chown wazuh-manager:wazuh-manager too — mirroring remoted's existing pattern exactly

Issue #38089, opened 2026-07-29
Migrate Inventory Sync to the asynchronous remoted communication module


juliancnn:
## Description

After implementing the new asynchronous communication mechanism between `wazuh-remoted` and `wazuh-modulesd`, Inventory Sync must be migrated to use it.

With the HTTPS agent protocol, each Inventory Sync session will be received completely within a single HTTP request. The session will no longer arrive as separate requests or require retransmission handling.

A new FlatBuffer message must wrap the complete session, including the existing:

* `start`
* `databatch`
* `end`

messages.

## Required behavior

Inventory Sync must:

* Receive the complete wrapped session through the new asynchronous communication module.
* Decode the FlatBuffer wrapper and process its `start`, `databatch`, and `end` content.
* Return the final response asynchronously when session processing completes.
* Remove the retransmission requirement from this new flow.

The module must return:

```http
503 Service Unavailable
```

when:

* there is not enough processing capacity available, either because:

  * the session limit has been reached
  * the configured byte or DataValue quota has been reached
* the session contains Syscollector package inventory and the Vulnerability Detector feed is not available

Rejected sessions must not reserve partial resources or begin processing.

## Scope

* Inventory Sync integration with the new remoted/modulesd communication module.
* New FlatBuffer wrapper for complete sessions.
* Admission control before session processing.
* Asynchronous response handling.
* Removal of retransmission handling from the new HTTPS flow.

## Deliverables

* Inventory Sync migrated to the asynchronous communication module.
* FlatBuffer schema containing the complete session.
* Processing of wrapped `start`, `databatch`, and `end` messages.
* `503` responses for exhausted capacity and unavailable Vulnerability Detector feed.
* Tests covering:

  * successful complete sessions
  * session and quota exhaustion
  * unavailable feed for Syscollector package sessions
  * malformed FlatBuffer messages
  * concurrent asynchronous sessions
  * correct resource release after completion or failure

Issue #38018, opened 2026-07-27
Design asynchronous communication between remoted and Inventory Sync


juliancnn:
## Description

The new HTTPS agent-to-manager protocol introduces synchronous request/response operations, including database synchronization requests handled by the Inventory Sync module in `wazuh-modulesd`.

For these operations, `wazuh-remoted` may need to keep many agent HTTPS requests open while Inventory Sync communicates with wazuh-indexer and completes the synchronization. This can result in several dozen concurrent requests waiting for a response.

The current communication mechanism between remoted and Inventory Sync is not suitable for this architecture:

1. Remoted publishes requests through the Router module using an observer pattern.
2. Inventory Sync processes the request.
3. `wazuh-modulesd` sends the response back to remoted through a socket.

A new request/response mechanism is required to efficiently handle many concurrent operations without assigning one blocked thread per pending request.

## Objectives

* Design and implement asynchronous communication between `wazuh-remoted` and Inventory Sync.
* Allow many Inventory Sync requests to remain pending concurrently.
* Avoid blocking one server thread for the complete lifetime of each request.
* Allow Inventory Sync to respond later when wazuh-indexer operations complete.
* Provide request correlation, timeout handling, backpressure, and graceful shutdown.

### RESTinio over Unix domain sockets

Evaluate whether the deferred-response HTTP server model introduced in #37947 can be reused inside `wazuh-modulesd` with Unix domain sockets.

The investigation must determine whether RESTinio supports:

* Unix domain stream sockets
* asynchronous request handling
* deferred responses from another thread
* many simultaneous pending requests
* graceful cancellation and shutdown

If supported, implement an internal asynchronous HTTP server over a Unix socket for Inventory Sync.

### Alternative Unix socket protocol

If RESTinio cannot provide the required Unix socket support, design an alternative mechanism using Unix stream sockets.

The alternative must support:

* multiple simultaneous connections
* asynchronous request processing
* request and response correlation
* deferred responses
* a dedicated processing pool
* response delivery after Inventory Sync completes the operation

A connection or server worker must not remain blocked while waiting for wazuh-indexer.

## Requirements

* The communication must support multiple concurrent Inventory Sync requests.
* Waiting for wazuh-indexer must not block the server's I/O threads.
* Pending requests and connections must have configurable limits.
* Requests must have a correlation identifier.
* Timeouts and client disconnections must release all associated resources.
* Shutdown must cancel or complete pending operations safely.
[...]

PR #38190, opened 2026-08-04
Validate the agent HTTPS connection and configuration block

### LINKED ISSUE #38163: Validate the HTTPS connection and agent configuration block (<client>/<agent>, TLS, batch) across all endpoints
|---|---|---|---|---|
| 5.0.0 | wazuh-agentd / https_client config parser, TLS transport | Agent + Manager | Sources / Packages | Linux / Windows / macOS |

> Part of the HTTPS communications initiative (#37702). Validates the agent-side connection/config
> layer (issue 01/#37828, foundation) against #37702's own configuration requirement (requirement
> 10) under its final canonical root, `<agent>` (the config-root migration drafted for issue 14, not
> yet confirmed filed on GitHub at the time of writing). This is a cross-cutting check: it validates
> the shared connection/config layer that every other endpoint-specific issue (lifecycle, task
> dispatch, upgrade, stateless, stateful, stats/config push) depends on, rather than re-testing each
> endpoint's own payload semantics.

## Description

#37702 (requirement 10) defines the full agent-side configuration block, under its final canonical
root `<agent>`:

```xml
<agent>
  <server>
    <address>192.168.1.100</address>
    <port>443</port>
  </server>

  <ssl>
    <certificate>path</certificate>
    <key>path</key>
    <certificate_authorities>path</certificate_authorities>
    <verification_mode>full|certificate|none</verification_mode>
    <ciphers>list</ciphers>
  </ssl>

  <batch>
    <size>1MB</size>
    <interval>10s</interval>
  </batch>

  <config-profile>webserver, debian8</config-profile>
  <notify_time>30</notify_time>
  <auto_restart>yes</auto_restart>
</agent>
```

Every option here must be operational, not just parsed, **under `<agent>`** — this issue validates
the 5.0.0 configuration surface as shipped, not `<client>` back-compat. The legacy `<client>` block
only matters for the 4.x -> 5.x transition, which is already in scope of the remote-upgrade
validation issue; do not re-test `<client>` parsing/precedence here. Once the connection is
correctly configured, it is the single transport every endpoint (`/control`, `/download`,
`/stateless`, `/stateful`, `/config`, `/stats`) rides on — this issue also confirms that the
endpoints covered by the other validation issues are actually reached using this same,
correctly-configured connection (no endpoint silently uses different TLS/server settings).

### In scope — verification items

- [ ] **(a) Full configuration block operational**: validate, one option at a time, that the
      proposed block — under the `<agent>` root — is fully functional:
      - `<server><address>`/`<port>`: connects to the configured manager.
      - `<ssl><certificate>`/`<key>`: client certificate presented when the manager requires one.
      - `<ssl><certificate_authorities>`/`<verification_mode>`: `full`/`certificate`/`none` behave
        as specified (prior QA rounds already confirmed the basic full/certificate/none matrix —
        see `QA_REPORT.md` §4.2 — extend rather than repeat: also cover a valid CA with a
        **matching** hostname/cert under `full`, and confirm `<ssl.certificate>`/`<key>` mutual-TLS
        specifically, which the prior rounds did not exercise).
      - `<ssl><ciphers>`: a restricted cipher list is honored (handshake fails against a manager
        offering only excluded ciphers).
      - `<batch><size>`/`<interval>`: bound the `/stateless` accumulator as documented (prior QA
        already confirmed the 413/split-and-resend behavior for `<batch><size>` — see
        `QA_REPORT.md` §4.4 — this item adds `<batch><interval>` specifically, not yet exercised).
      - `<config-profile>`: reported correctly wherever it is consumed (group profile matching,
        `/config` push).
      - `<notify_time>`: the `/control notify` cadence matches the configured value.
      - `<auto_restart>`: agent restarts (or doesn't) on config/settings change per this flag.
- [ ] **(b) Endpoint connectivity validation across the remaining issues**: confirm every endpoint
      already covered by the other validation issues in this set (`/control`, `/download`,
      `/stateless`, `/stateful`, `/config`, `/stats`) is reached through this same validated
      connection — same server address/port, same TLS settings, same signing key — with no
      endpoint falling back to different or unvalidated connection parameters.

### Out of scope

- `<client>` legacy-block parsing, precedence, and deprecated/missing-option logging — that is the
  4.x -> 5.x transition surface, validated as part of the remote-upgrade issue instead.

## Environment

Real manager preferred for the TLS matrix (mutual TLS, ciphers) since it requires a manager capable
of requesting/validating a client certificate. The existing QA harness
(`/home/nico/wz/https-agent-test-environment/`) already supports `VERIFY_MODE`/`CA_PATH_OVERRIDE`
overrides and can be extended for ciphers/mutual-TLS/`<batch><interval>` cases without a real
manager.

## Acceptance Criteria

- [ ] (a): each option in the list above has a positive functional test under `<agent>`, not just
      parser-level coverage.
- [ ] (b): a trace showing at least one request per endpoint sharing the same validated connection
      parameters.
- [ ] Any option that parses but has no observable runtime effect, or any endpoint bypassing the
      validated connection config, is filed as its own defect referencing this issue.

jr0me:
## Description

Validation of the agent-side HTTPS connection and configuration layer for #38163, plus the
defects it found and the harness changes it needed. It also carries two fixes from QA's
exploratory session on the same layer, wazuh/wazuh-qa-automation#8451.

Every case drives the real `wazuh-agentd` reading `ossec.conf`, so the XML parser is in the loop
rather than a driver building `hc_config_t` directly, and each runs in isolation: fresh manager,
foreground agent under `timeout`, no daemon surviving from the previous case.

Related: #37702 (requirement 10), #37828.

## Proposed Changes

- **`shared_modules/agent_metadata`** — `update()` and `reset()` refuse to write when the segment was
  mapped read-only, instead of faulting.
- **`client-agent/https_client` + `config/client-agent`** — the agent now refuses to negotiate
  below TLS 1.3, sends `<ssl><ciphers>` through `CURLOPT_TLS13_CIPHERS` instead of
  `CURLOPT_SSL_CIPHER_LIST`, and rejects non-TLS-1.3 suite names at parse time. This fixes the
  `<ciphers>` defect the validation found; see the section below.
- **`client-agent/https_client/demo/mock_manager.py`** — `--client-ca` (require and verify a client
  certificate) and `--ciphers` (restrict the offered suites), the negotiated suite and client
  certificate subject logged once per connection, and the full `/stats` and `/config` document
  logged, which the 70-byte preview truncated.
- **Agent config templates** — the three shipped templates no longer write a literal `PATH` into
  `<certificate_authorities>`, and an unreadable CA is now reported rather than ignored. See the
  section below.
- **`<notify_time>`** — the shipped default drops from 20 seconds to 10, on both the Linux and
  Windows templates.

Full evidence for #38163. Every case drives the real `wazuh-agentd` reading `ossec.conf`, so the XML parser is in the loop.

### Environment

Agent and manager both built from `enhancement/37831-agent-https-client` (`f92fbf477d`, rebased onto `5.0.0-https` `84ae957631`) plus the metadata fix in this PR, which the agent does not start without.

The TLS matrix runs against a **real manager** — `remoted`'s HTTPS listener on `:27000`, agent enrolled with a matching `client.keys`, shared CA, manager certificate `SAN: DNS:wazuh-server, IP:172.19.0.3`. Cases the real manager cannot serve are marked and were run against the repo's mock manager (`https_client/demo/mock_manager.py`), which the issue's Environment section allows.

### (a) Configuration block

**There is no `<agent>` config root.** `src/config/src/config.c` knows `client`, not `agent`, so the block in #37702 requirement 10 is rejected and the agent refuses to start:

```
ERROR: (1230): Invalid element in the configuration: 'agent'.
ERROR: (1202): Configuration error at 'etc/ossec.conf'.
ERROR: (1215): No client configured. Exiting.
```

Every option the issue lists is implemented and functional, but under `<client>`. The config-root migration arrived with **PR #38172** (agent remote upgrade / WPK): it registers `agent` as a root in `config.c`, renames the readers (`Read_Client_SSL` becomes `Read_Agent_SSL`) and keeps `<client>` as a legacy fallback from which only `<server><address>` is read. That PR has merged, and the matrix has been re-run under `<agent>` — see the evidence comment on this PR, which supersedes the `<client>`-based results below.

| Option | Case | Result | Against |
|---|---|---|---|
| `<server><address>`/`<port>` | connects to the configured manager | PASS | real |
| `<ssl><verification_mode>full` | matching CA + SAN | PASS | real |
| `<ssl><verification_mode>full` | unrelated CA — must refuse | PASS (0 requests) | real |
| `<ssl><verification_mode>certificate` | connects | PASS | real |
| `<ssl><verification_mode>none` | connects, warns | PASS | real |
| `<ssl><certificate>`/`<key>` | manager requires a client cert, agent presents one | PASS | real |
| `<ssl><certificate>`/`<key>` | manager requires, agent has none — must refuse | PASS (0 requests) | real |
| `<ssl><ciphers>` | restricts the negotiated suite | **DEFECT when measured**, fixed in this PR | real |
| `<batch><interval>` | flush delay tracks the configured value | PASS | mock |
| `<notify_time>` | notify cadence matches | PASS | mock |
| `<auto_restart>` | reload on limits change gated by the flag | PASS | mock |
| `<config-profile>` | reported in the `/config` push | PARTIAL | mock |

**Mutual TLS**, isolated with `curl` so the enforcement is demonstrably the manager's:

```
# no client cert
curl --cacert ca.crt https://127.0.0.1:27000/control
  -> curl: (56) OpenSSL SSL_read: error:0A00045C:... tlsv13 alert certificate required

# with client cert
curl --cacert ca.crt --cert client.crt --key client.key https://127.0.0.1:27000/control
  -> {"error":"Missing required header: protocol-version","code":400}
```

That 400 is the manager correctly rejecting a bare `curl` that sends no headers. The agent sends `protocol-version: 1` (`cmacSigner.cpp:210`) and its `/control` startup against the same listener is accepted.

**`<notify_time>`** — with `5`, consecutive notifies land 5050 ms apart (startup 1971 ms, NOTIFY #1 7022 ms, NOTIFY #2 12072 ms).

**`<batch><interval>`** — differential, same config otherwise:

| `<interval>` | startup | `/stateless` flush | delta |
|---|---|---|---|
| 3 | 1971 ms | 5025 ms | 3054 ms |
| 10 | 1988 ms | 12041 ms | 10053 ms |

**`<auto_restart>`** — same trigger both runs (module limits flipped `fim.file` 100000 -> 200000 on the 5th notify), only the flag differs:

```
yes -> INFO:  https_client: reloading due to module limits changes.   (bridge_on_startup_result:379, reloadAgent())
no  -> DEBUG: https_client: module limits have been updated.          (line 382, no reload)
```

**`<config-profile>`** is PARTIAL: it comes back verbatim in the `/config` document (`"config-profile": "webserver, debian8"`), but the issue also asks for it "wherever it is consumed", i.e. group profile matching. That needs a manager serving per-profile shared configuration, which is not modelled here.

### (b) Endpoints share the validated connection

One run with both reports enabled, 12 authenticated requests, every endpoint over the same connection parameters — same `agent_id=001`, same negotiated TLS, same signing key:

```
POST /control    <- id=001 auth OK  {"type":"startup","version":"v5.0.0"}
POST /stats      <- id=001 auth OK  (305 B)
POST /config     <- id=001 auth OK  (2363 B)
POST /stateless  <- id=001 auth OK  (152 B)
POST /control    <- id=001 auth OK  (notify)
```

No endpoint used different or unvalidated connection settings.

**Against a real manager, five of the six endpoints exist.** Probed directly, with a nonsense path as the control so a 404 is meaningful:

```
/control  /stateless  /stats  /config  /download   -> 401 (route exists, auth rejected)
/stateful          /nonsense-control               -> 404 (not registered)
```

`/download` landed with #38083 and is now verified end to end against the real manager: a group config change makes the hashes diverge, the agent downloads and SHA-256-verifies the body, and applies it —

```
INFO: Manager config hash 8948520ed001... differs from the local one; downloading the new configuration (group 'default').
DEBUG: https_client config downloaded (hash=8948520ed001..., file=/tmp/hc_sync_...tmp)
INFO: https_client: applying configuration downloaded over HTTPS (hash=8948520ed001...).
```

**`/stateful` is still absent server-side.** It is not a standalone endpoint issue: the server half lands with the Inventory Sync migration, #38089, where a whole sync session arrives in one HTTP request wrapped in a single FlatBuffer (design in #38018). Until that lands, the mock run above remains the only evidence covering all six endpoints.

### Defect fixed: `wazuh-agentd` segfaults on start

```
SIGSEGV in std::atomic<bool>::store
  #2 SharedMemoryProvider::update()   metadata_provider.cpp:74
  #3 metadata_provider_update
  #4 w_agentd_populate_metadata()     start_agent.c:212
  #5 start_agent(is_startup=1)        start_agent.c:48
  #6 AgentdStart / main
```

The provider opens `var/run/.wazuh_agent_metadata` `O_RDWR` and falls back to `O_RDONLY` on `EACCES`, mapping it `PROT_READ`. `update()` checked only that the mapping existed, so the fallback path wrote through a read-only mapping. `agentd` drops to the `wazuh` user, and the segment is left `root:wazuh 0644` whenever a root process creates it first, so the fallback is taken on every start. `reset()` had the same hole.

Before and after on the same install path:

| Build | `wazuh-agentd` |
|---|---|
| integration, without this fix | exit 139 (SIGSEGV) |
| integration + this fix | exit 124 (survived) |

**Surviving is not the whole claim — the metadata still has to be published.** Reproducing the exact
trigger, a root process creating the segment before the daemon starts, and seeding it with zeros so
anything found afterwards must have been written by `agentd`:

```
before, seeded by root with zeros:
  -rw-r--r-- 1 root wazuh 35616 .wazuh_agent_metadata
  strings in segment: 0

after agentd ran (same file, still root-owned):
  -rw-r--r-- 1 root wazuh 35616 .wazuh_agent_metadata
  agentd alive: yes
  strings in segment:
    001                              <- agent id
    agent-001                        <- agent name
    v5.0.0                           <- version
    aarch64                          <- architecture
    wazuh-agent                      <- hostname
    Ubuntu / ubuntu / linux          <- os name, platform
    22.04.5 LTS (Jammy Jellyfish)    <- os version
    wazuh / node01                   <- cluster name, node
    default                          <- agent group
```

Every field of the record, written into a segment that held nothing but zeros a moment earlier. The
last three are the load-bearing ones: cluster name, node and group are manager-authoritative and only
exist after a successful handshake, so this is a live agent's record rather than anything left behind.

The same holds on the clean path, where `agentd` creates the segment itself (`wazuh:wazuh`). So the
read-only fallback is reachable — the SIGSEGV above proves it — but it is not on the path that
publishes metadata, and no configuration produced an empty record.

One limit: both runs are manual, for the reason under *Tests Introduced*. Also worth noting
that when the guard does fire, `start_agent.c:220` reports it at `mdebug1` ("Failed to populate early
metadata"), so a silent crash has become a silent no-op; `mwarn` would suit an event that means the
agent published nothing.

**Provenance**, since this is not our code. The read-only fallback came from `3448f64d14` "fix: allow process that dont run as root to access agent metadata" (Tomas Turina, 2026-01-07), which landed in PR #33575 "Stateless metadata" (merged 2026-01-15). `origin/5.0.0-https` still carries the same unguarded write; it is latent there because the only caller is the legacy handshake, whereas this branch reaches it unconditionally from `start_agent()`. The fix is carried here because none of the validation could run otherwise, and it is two guards with no interface change.

Two things left to the owners: whether the guard belongs on `5.0.0-https` directly rather than arriving via this branch, and the missing test seam — the provider is a singleton whose read-only state is decided in a private constructor from process credentials, and the unit suite runs as root, where the `O_RDWR` open always succeeds, so the faulting path is unreachable from a test as written. Happy to split the fix into its own PR against `5.0.0-https` if preferred.

### Not covered

- **`<agent>` config root** — does not exist; blocks (a).
- **`<config-profile>` group profile matching** — needs a manager serving per-profile shared configuration.
- **(b) fully against a real manager** — blocked on #38089 for `/stateful`, the one endpoint still not registered server-side.

### `<ssl><ciphers>` has no effect on the connections that actually happen

The agent maps `<ciphers>` to `CURLOPT_SSL_CIPHER_LIST`, which libcurl applies to **TLS 1.2 and below**; `CURLOPT_TLS13_CIPHERS` is never set (`curlHandle.cpp:55` is the only cipher option in the module). The real manager negotiates **TLS 1.3**:

```
$ openssl s_client -connect 127.0.0.1:27000
New, TLSv1.3, Cipher is TLS_AES_256_GCM_SHA384
```

So the configured list never constrains the session. Demonstrated with a cipher no server can offer:

```
ciphers=(none)                       -> SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
ciphers=ECDHE-RSA-AES256-GCM-SHA384  -> SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
ciphers=NULL-SHA256                  -> SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
```

and at the agent level, `<ciphers>NULL-SHA256</ciphers>` still completes `/control` startup against the real manager.

This meets the issue's "parses but has no observable runtime effect" bar, so rather than only filing it, this PR fixes it — the option was inert in three separate ways:

- **No TLS floor was set at all.** `CURLOPT_SSLVERSION` was never touched, so libcurl's default applied and TLS 1.0 was permitted. It is now pinned to 1.3 unconditionally, matching the manager: this is the protocol's floor, not something `<ssl>` may lower.
- **The list never reached the session.** It now goes through `CURLOPT_TLS13_CIPHERS`.
- **Bad names were accepted silently.** They are now rejected at parse time against the five RFC 8446 §B.4 suite names, so a mistake is a startup error instead of a security option that quietly does nothing. This mirrors `w_authd_validate_ciphers()` from #38205, which fixes the same class of bug for `wazuh-authd` under #38091 but does not touch the agent's HTTPS client.

`curlPerformer.cpp` carries no curl header and the option ids are deliberately curl-agnostic, so the 1.3 constant lives in `iCurlHandle.hpp` and is pinned with a `static_assert` against `CURL_SSLVERSION_TLSv1_3` in the one translation unit that does include curl. The literal cannot drift.

The evidence above was measured before this fix; re-running the matrix moves `<ciphers>` to PASS.

Note the two sides do not even share a namespace: the manager restricts suites with `SSL_CTX_set_ciphersuites()` (`RestinioHttpServer.cpp:357`), which takes TLS 1.3 names (`TLS_AES_256_GCM_SHA384`), while the agent's option takes TLS 1.2 names (`ECDHE-RSA-AES256-GCM-SHA384`).

### Defect fixed: the shipped config names a CA that does not exist

Every agent ships `<certificate_authorities>PATH</certificate_authorities>`, written by
`inst-functions.sh`, `etc/ossec-agent.conf` and `src/win32/ossec.conf`. Nothing substitutes that
token — no installer variable, no packaging step, no CI step. It parses cleanly, so the config
reads as if a CA were configured when none is. QA found it in a deployed agent
(wazuh/wazuh-qa-automation#8451, F3).

It is inert only because the templates currently ship `verification_mode=none`, which never reads
the CA. The moment anyone sets `certificate` or `full` without also fixing the path, the agent
refuses to start — that refusal is correct and already exists (`AG_INV_SSL_CA`, 4118). What was
missing is that under `none` an unusable path was accepted in total silence.

Both halves are addressed: the templates ship the element commented out, so an unconfigured agent
looks unconfigured, and an unreadable CA is now reported under `none` as well. All three states,
verbatim from `wazuh-agentd -t` reading `ossec.conf`:

```
CA commented out, mode=none            -> exit 0, silent
<certificate_authorities>PATH</...>,   -> exit 0
  mode=none                               WARNING: (4119): <ssl><certificate_authorities> is not a
                                          readable file: 'PATH'. It is unused while
                                          <verification_mode> is 'none'; enabling verification
                                          would stop the agent from starting.
<certificate_authorities>PATH</...>,   -> exit 1
  mode=full                               ERROR: (4118): <ssl><verification_mode> is not 'none' but
                                          <certificate_authorities> is missing or unreadable: 'PATH'.
                                          ERROR: (1215): No client configured. Exiting.
```

The first case matters on its own: an XML comment inside `<ssl>` has to parse, or the template
change would break every new install.

`agentd` and the Windows agent had duplicated the 4118 check line for line. It now lives in one
function, `w_agent_validate_ssl_ca()`, so the two cannot drift on what counts as a usable CA.

The `MANAGER_IP` token QA saw in the same report is **not** a defect: `install.sh` substitutes it
from `USER_AGENT_MANAGER_IP` and the packaging smoke tests `sed` it. An agent that reaches runtime
still holding it already refuses to start, which is the intended behaviour.

### One observation, not investigated

A group `agent.conf` carrying only a `<labels>` block is downloaded and verified, then rejected by `verifyRemoteConf()` with `downloaded configuration failed validation; not reloading`, while a `<localfile>` block in the same position applies cleanly. Whether `<labels>` is meant to be valid in shared configuration is outside this issue's scope, but it is reproducible if someone wants to pick it up.

## Artifacts Affected

- `libagent_metadata` — no interface change; `metadata_provider_update()` returns `-1` on the
  read-only path instead of faulting.
- The demo mock manager, which is test tooling and is not shipped.
- **Shipped agent configuration.** New installs get `<certificate_authorities>` commented out
  instead of set to `PATH`, and `<notify_time>` 10 instead of 20. Existing installs are untouched —
  an upgrade never rewrites `ossec.conf` — so an agent already carrying the `PATH` token keeps
  running and now logs warning 4119 once at startup. Halving `notify_time` doubles each agent's
  `/control` rate against the manager.

## Tests Introduced

`src/unit_tests/client-agent/test_agent_ssl_ca.c` — seven cases over
`w_agent_validate_ssl_ca()`, covering both verification modes against a CA that is absent,
readable, and unreadable, and asserting the exact 4118 and 4119 text. Full suite: 107/107.

The metadata fix remains unreachable from the unit tests: the provider is a singleton whose
read-only state is decided in a private constructor from the process's own credentials, and the
suite runs as root, for which the `O_RDWR` open always succeeds. Making it testable means giving the
provider a seam for the mapping — worth doing, larger than this change. It is covered by the live
reproduction above.

## Review Checklist

- [x] Configuration changes documented — the shipped templates change, see *Artifacts Affected*
- [x] Meets requirements and/or definition of done — all three acceptance criteria are met; the
      evidence comment on this PR carries the full matrix, re-run under `<agent>` against a real
      manager.
- [ ] Follow-up to file: the read-only seam for the metadata provider, which still has no automated
      test. (The `<agent>` root arrived with PR #38172 and the `<ssl><ciphers>` no-op is fixed here,
      so neither needs a new issue.)
- [ ] Not covered, and stated in the evidence: Windows and macOS. The issue's platform row is
      Linux / Windows / macOS; everything here is Linux.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,4 +128,5 @@
 | [#37543](https://github.com/wazuh/wazuh/issues/37543) | Fixed agent-info module caching cluster_name, cluster_node, and agent_groups from a one-time handshake at startup, causing stale values in `agent_metadata` until the agent process restarted. |
 | [#37993](https://github.com/wazuh/wazuh/issues/37993) | Fixed `wazuh-syscheckd` failing the `file_entry.checksum` NOT NULL constraint when the deferred sync-flag update ran for an entry deleted during the scan. |
 | [#37993](https://github.com/wazuh/wazuh/issues/37993) | Fixed `wazuh-syscheckd` failure on shutdown, which logged "Invalid handle value", crashed the process and left a stale PID file. |
+| [#38163](https://github.com/wazuh/wazuh/issues/38163) | Fixed `wazuh-agentd` crashing on start when the agent metadata segment could only be opened read-only, which happens whenever a root process creates it before the daemon drops privileges. |
 
--- a/etc/ossec-agent.conf
+++ b/etc/ossec-agent.conf
@@ -15,7 +15,7 @@
          Set to 'full' or 'certificate' and provide certificate_authorities to verify the
          manager's certificate. -->
     <ssl>
-      <certificate_authorities>PATH</certificate_authorities>
+      <!-- <certificate_authorities>etc/certs/root-ca.pem</certificate_authorities> -->
       <verification_mode>none</verification_mode>
     </ssl>
 
--- a/src/client-agent/https_client/demo/mock_manager.py
+++ b/src/client-agent/https_client/demo/mock_manager.py
@@ -134,9 +134,10 @@ class Handler(BaseHTTPRequestHandler):
     def _log_tls_once(self):
         """Print what the handshake actually negotiated, once per connection.
 
-        This is the observable half of <ssl><ciphers> and <ssl><certificate>/<key>:
-        the agent's config is only proven to have a runtime effect if the manager
-        can see the suite it picked and the client cert it presented (#38163).
+        This is the observable half of <ssl><certificate>/<key>: the agent's config
+        is only proven to have a runtime effect if the manager can see the client
+        cert it presented. The negotiated suite is logged alongside it so the TLS
+        version in use is visible, not because this demo restricts it (#38163).
         """
         if getattr(self, "_tls_logged", False):
             return
@@ -396,31 +397,33 @@ def main():
     parser.add_argument("--cert", required=True)
     parser.add_argument("--key", required=True)
     parser.add_argument("--key-hex", default="000102030405060708090a0b0c0d0e0f")
-    # The two below exist for the connection/config validation (#38163): the agent's
-    # <ssl><certificate>/<key> and <ssl><ciphers> can only be shown to have a runtime
-    # effect against a manager that actually demands a client cert / restricts the suites.
+    # Exists for the connection/config validation (#38163): the agent's
+    # <ssl><certificate>/<key> can only be shown to have a runtime effect against a
+    # manager that actually demands a client certificate.
     parser.add_argument("--client-ca",
                         help="Require a client certificate and verify it against this CA "
                              "(mutual TLS). Without it, client certs are not requested.")
-    parser.add_argument("--ciphers",
-                        help="Restrict the suites this manager offers, e.g. "
-                             "'ECDHE-RSA-AES256-GCM-SHA384'. A client whose <ciphers> "
-                             "does not intersect this list fails the handshake.")
     args = parser.parse_args()
 
     Handler.key_hex = args.key_hex
     context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
     context.load_cert_chain(args.cert, args.key)
 
+    # Same floor remoted enforces (SSL_CTX_set_min_proto_version in
+    # RestinioHttpServer), so the demo negotiates what a real manager would.
+    #
+    # There is deliberately no --ciphers here. Restricting the suites would mean
+    # SSL_CTX_set_ciphersuites(), which Python's ssl module does not expose:
+    # set_ciphers() governs TLS 1.2 and below only and rejects 1.3 suite names
+    # outright. The one way to make a restriction observable was to cap the
+    # server at TLS 1.2, which an agent that requires 1.3 can no longer reach --
+    # the handshake would fail before any suite was chosen. <ssl><ciphers> is
+    # therefore validated against a real manager, not here.
+    context.minimum_version = ssl.TLSVersion.TLSv1_3
+
     if args.client_ca:
         context.verify_mode = ssl.CERT_REQUIRED
         context.load_verify_locations(args.client_ca)
-
-    if args.ciphers:
-        # TLS 1.3 suites are negotiated separately and set_ciphers() does not
-        # constrain them, so pin to 1.2 to make the restriction observable.
-        context.maximum_version = ssl.TLSVersion.TLSv1_2
-        context.set_ciphers(args.ciphers)
 
     server = ThreadingHTTPServer(("127.0.0.1", args.port), Handler)
     server.requires_client_cert = bool(args.client_ca)
@@ -428,7 +431,7 @@ def main():
     log(f"HTTPS manager on https://127.0.0.1:{args.port} "
         f"(agent key {args.key_hex[:8]}..)")
     log(f"     mutual TLS: {'required, CA ' + args.client_ca if args.client_ca else 'not requested'}")
-    log(f"     ciphers:    {args.ciphers if args.ciphers else 'default (unrestricted)'}")
+    log(f"     min TLS:    1.3")
     try:
         server.serve_forever()
     except KeyboardInterrupt:
--- a/src/client-agent/https_client/demo/mock_manager.py
+++ b/src/client-agent/https_client/demo/mock_manager.py
@@ -131,6 +131,27 @@ class Handler(BaseHTTPRequestHandler):
     def _config_blob(cls):
         return cls.CONFIG_V2 if cls.notify_count >= cls.CONFIG_FLIP_AT else cls.CONFIG_V1
 
+    def _log_tls_once(self):
+        """Print what the handshake actually negotiated, once per connection.
+
+        This is the observable half of <ssl><ciphers> and <ssl><certificate>/<key>:
+        the agent's config is only proven to have a runtime effect if the manager
+        can see the suite it picked and the client cert it presented (#38163).
+        """
+        if getattr(self, "_tls_logged", False):
+            return
+        self._tls_logged = True
+        conn = self.connection
+        cipher = conn.cipher() if hasattr(conn, "cipher") else None
+        peer = conn.getpeercert() if hasattr(conn, "getpeercert") else None
+        if cipher:
+            log(f"TLS  cipher={cipher[0]} proto={cipher[1]}")
+        if peer:
+            subject = dict(x[0] for x in peer.get("subject", ()))
+            log(f"TLS  client cert presented: subject={subject}")
+        elif getattr(self.server, "requires_client_cert", False):
+            log("TLS  client cert REQUIRED but none presented")
+
     def log_message(self, *_):  # silence the default noisy logging
         pass
 
@@ -181,6 +202,7 @@ class Handler(BaseHTTPRequestHandler):
         self.wfile.write(b"0\r\n\r\n")
 
     def do_POST(self):
+        self._log_tls_once()
         length = int(self.headers.get("Content-Length", 0))
         body = self.rfile.read(length)
         target = self.path
@@ -308,6 +330,10 @@ class Handler(BaseHTTPRequestHandler):
         cluster = doc.get("cluster", {})
         log(f"     {target:<11} -> 200  stored: agent_id={doc.get('agent_id')} "
             f"cluster={cluster.get('name')}/{cluster.get('node')} keys={sorted(doc.keys())}")
+        # The whole document, for the field-contract comparison against the
+        # manager side (#38136). The preview in do_POST() truncates at 70 bytes,
+        # which is not enough to see the per-module bodies.
+        log(f"     {target} FULL BODY: {json.dumps(doc, sort_keys=True)}")
 
     def _handle_stateful(self, body):
         session = self.headers.get("X-Session-Id", "?")
@@ -370,16 +396,39 @@ def main():
     parser.add_argument("--cert", required=True)
     parser.add_argument("--key", required=True)
     parser.add_argument("--key-hex", default="000102030405060708090a0b0c0d0e0f")
+    # The two below exist for the connection/config validation (#38163): the agent's
+    # <ssl><certificate>/<key> and <ssl><ciphers> can only be shown to have a runtime
+    # effect against a manager that actually demands a client cert / restricts the suites.
+    parser.add_argument("--client-ca",
+                        help="Require a client certificate and verify it against this CA "
+                             "(mutual TLS). Without it, client certs are not requested.")
+    parser.add_argument("--ciphers",
+                        help="Restrict the suites this manager offers, e.g. "
+                             "'ECDHE-RSA-AES256-GCM-SHA384'. A client whose <ciphers> "
+                             "does not intersect this list fails the handshake.")
     args = parser.parse_args()
 
     Handler.key_hex = args.key_hex
     context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
     context.load_cert_chain(args.cert, args.key)
 
+    if args.client_ca:
+        context.verify_mode = ssl.CERT_REQUIRED
+        context.load_verify_locations(args.client_ca)
+
+    if args.ciphers:
+        # TLS 1.3 suites are negotiated separately and set_ciphers() does not
+        # constrain them, so pin to 1.2 to make the restriction observable.
+        context.maximum_version = ssl.TLSVersion.TLSv1_2
+        context.set_ciphers(args.ciphers)
+
     server = ThreadingHTTPServer(("127.0.0.1", args.port), Handler)
+    server.requires_client_cert = bool(args.client_ca)
     server.socket = context.wrap_socket(server.socket, server_side=True)
     log(f"HTTPS manager on https://127.0.0.1:{args.port} "
         f"(agent key {args.key_hex[:8]}..)")
+    log(f"     mutual TLS: {'required, CA ' + args.client_ca if args.client_ca else 'not requested'}")
+    log(f"     ciphers:    {args.ciphers if args.ciphers else 'default (unrestricted)'}")
     try:
         server.serve_forever()
     except KeyboardInterrupt:
--- a/src/client-agent/https_client/src/curlHandle.cpp
+++ b/src/client-agent/https_client/src/curlHandle.cpp
@@ -34,6 +34,9 @@ namespace
         std::call_once(initialized, [] { curl_global_init(CURL_GLOBAL_DEFAULT); });
     }
 
+    static_assert(TLS_MIN_VERSION_1_3 == CURL_SSLVERSION_TLSv1_3,
+                  "TLS_MIN_VERSION_1_3 must stay equal to libcurl's CURL_SSLVERSION_TLSv1_3");
+
     const std::map<CurlOption, CURLoption>& optionMap()
     {
         // Never destroyed, like the global curl init above: the shutdown drain reads
@@ -52,7 +55,11 @@ namespace
             {CurlOption::CaInfo, CURLOPT_CAINFO},
             {CurlOption::SslCert, CURLOPT_SSLCERT},
             {CurlOption::SslKey, CURLOPT_SSLKEY},
-            {CurlOption::SslCiphers, CURLOPT_SSL_CIPHER_LIST},
+            {CurlOption::SslVersion, CURLOPT_SSLVERSION},
+            // TLS13_CIPHERS, not SSL_CIPHER_LIST: the latter only governs TLS 1.2
+            // and below, which the minimum version below rules out entirely, so a
+            // list set through it could never constrain a session.
+            {CurlOption::SslCiphers, CURLOPT_TLS13_CIPHERS},
             {CurlOption::FollowLocation, CURLOPT_FOLLOWLOCATION},
             {CurlOption::NoSignal, CURLOPT_NOSIGNAL}
         };
--- a/src/client-agent/https_client/src/curlPerformer.cpp
+++ b/src/client-agent/https_client/src/curlPerformer.cpp
@@ -204,6 +204,12 @@ void CurlPerformer::applyTls(ICurlHandle& handle) const
     handle.setOptionLong(CurlOption::VerifyPeer, verifyPeer ? 1L : 0L);
     handle.setOptionLong(CurlOption::VerifyHost, verifyHost ? 2L : 0L);
 
+    // The manager's HTTPS listener sets a TLS 1.3 floor of its own
+    // (SSL_CTX_set_min_proto_version in RestinioHttpServer), so match it instead
+    // of leaving libcurl's default, which still permits 1.0. Unconditional: this
+    // is the protocol's floor, not something <ssl> is allowed to lower.
+    handle.setOptionLong(CurlOption::SslVersion, TLS_MIN_VERSION_1_3);
+
     if (!m_config.caPath.empty())
     {
         handle.setOptionString(CurlOption::CaInfo, m_config.caPath);
--- a/src/client-agent/https_client/src/iCurlHandle.hpp
+++ b/src/client-agent/https_client/src/iCurlHandle.hpp
@@ -35,10 +35,18 @@ enum class CurlOption
     CaInfo,         ///< string CA file path
     SslCert,        ///< string client certificate path
     SslKey,         ///< string client key path
-    SslCiphers,     ///< string cipher list
+    SslVersion,     ///< long, the minimum TLS version to negotiate
+    SslCiphers,     ///< string TLS 1.3 ciphersuite list
     FollowLocation, ///< long, always 0 (H4: no redirects)
     NoSignal        ///< long, always 1 (H6)
 };
+
+/// Value for CurlOption::SslVersion: refuse to negotiate below TLS 1.3.
+///
+/// Numerically libcurl's CURL_SSLVERSION_TLSv1_3, repeated here so this header
+/// stays curl-agnostic like the ids above. curlHandle.cpp static_asserts the two
+/// against each other, so the literal cannot drift from what curl expects.
+inline constexpr long TLS_MIN_VERSION_1_3 {7};
 
 /**
  * @brief One HTTP transfer, at the option level.
--- a/src/client-agent/include/agentd.h
+++ b/src/client-agent/include/agentd.h
@@ -21,6 +21,10 @@
 /* Client configuration */
 int ClientConf(const char *cfgfile);
 
+/* Check <ssl><certificate_authorities> against the configured verification mode.
+ * Returns false when the agent must not start. */
+bool w_agent_validate_ssl_ca(const agent *cfg);
+
 /* Parse read config into JSON format */
 cJSON *getClientConfig(void);
 cJSON *getAgentInternalOptions(void);
--- a/src/client-agent/src/config.c
+++ b/src/client-agent/src/config.c
@@ -84,6 +84,34 @@ int ClientConf(const char *cfgfile)
     return (1);
 }
 
+/* Both agentd and the Windows agent gate startup on this, at the point where each can
+ * still fail cleanly: before daemonizing on POSIX, before the first module thread on
+ * Windows. Shared so the two can never drift apart on what counts as a usable CA. */
+bool w_agent_validate_ssl_ca(const agent *cfg)
+{
+    const char *ca = cfg->ssl.certificate_authorities;
+    const bool readable = ca && w_is_file(ca);
+
+    /* Under 'none' the CA is never read, so a wrong path stays invisible until someone
+     * enables verification -- and then the agent refuses to start. Warn while it is
+     * still harmless rather than accepting it in silence. */
+    if (cfg->ssl.verification_mode == AGENT_VERIFY_NONE) {
+        if (ca && !readable) {
+            mwarn(AG_UNUSED_SSL_CA, ca);
+        }
+
+        return true;
+    }
+
+    /* A verifying mode without a readable CA can never connect: the https_client module
+     * fails closed on this, per its own validation. */
+    if (!readable) {
+        merror(AG_INV_SSL_CA, ca ? ca : "");
+        return false;
+    }
+
+    return true;
+}
 
 cJSON *getClientConfig(void) {
 
--- a/src/client-agent/src/main.c
+++ b/src/client-agent/src/main.c
@@ -208,14 +208,11 @@ int main(int argc, char **argv)
         mlerror_exit(LOGLEVEL_ERROR, CLIENT_ERROR);
     }
 
-    /* A verifying <ssl><verification_mode> without a readable CA can never connect (the
-     * https_client module fails closed on this, per its own validation) -- caught here, before
-     * daemonizing, so wazuh-client.sh's sequential daemon-start check (which polls for this
-     * process's PID file before starting syscheckd/logcollector/modulesd) halts the whole start
-     * instead of launching the other daemons around a transport that will never come up. */
-    if (agt->ssl.verification_mode != AGENT_VERIFY_NONE &&
-        (!agt->ssl.certificate_authorities || !w_is_file(agt->ssl.certificate_authorities))) {
-        merror(AG_INV_SSL_CA, agt->ssl.certificate_authorities ? agt->ssl.certificate_authorities : "");
+    /* Checked here, before daemonizing, so wazuh-client.sh's sequential daemon-start check
+     * (which polls for this process's PID file before starting syscheckd/logcollector/modulesd)
+     * halts the whole start instead of launching the other daemons around a transport that will
+     * never come up. */
+    if (!w_agent_validate_ssl_ca(agt)) {
         mlerror_exit(LOGLEVEL_ERROR, CLIENT_ERROR);
     }
 
--- a/src/config/src/client-config.c
+++ b/src/config/src/client-config.c
@@ -441,6 +441,74 @@ int Read_Agent_Server(XML_NODE node, agent * logr)
     return (0);
 }
 
+/* TLS 1.3 ciphersuite names, RFC 8446 section B.4. These are what OpenSSL's
+ * SSL_CTX_set_ciphersuites() takes, which is what the agent's <ciphers> ends up
+ * feeding through libcurl's CURLOPT_TLS13_CIPHERS. */
+static const char *VALID_TLS13_CIPHERSUITES[] = {
+    "TLS_AES_128_GCM_SHA256",
+    "TLS_AES_256_GCM_SHA384",
+    "TLS_CHACHA20_POLY1305_SHA256",
+    "TLS_AES_128_CCM_SHA256",
+    "TLS_AES_128_CCM_8_SHA256",
+    NULL
+};
+
+static bool is_tls13_ciphersuite(const char *name)
+{
+    for (int i = 0; VALID_TLS13_CIPHERSUITES[i]; i++) {
+        if (strcmp(name, VALID_TLS13_CIPHERSUITES[i]) == 0) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+/**
+ * @brief Validate a colon-separated TLS 1.3 ciphersuite list.
+ *
+ * The agent negotiates nothing below TLS 1.3, so a TLS 1.2 name here would parse
+ * and then never constrain a session — silently, since OpenSSL is never told
+ * about it. Rejecting at parse time is the difference between a startup error
+ * and a security option that quietly does nothing.
+ *
+ * @param ciphers Raw <ciphers> content.
+ * @return 0 when every element is a known suite, OS_INVALID otherwise.
+ */
+static int w_client_validate_tls13_ciphers(const char *ciphers)
+{
+    char *copy = NULL;
+    char *saveptr = NULL;
+    int suites = 0;
+
+    if (!ciphers || !*ciphers) {
+        merror("Empty 'ciphers' option: expected TLS 1.3 cipher suite names.");
+        return OS_INVALID;
+    }
+
+    os_strdup(ciphers, copy);
+
+    for (char *name = strtok_r(copy, ":", &saveptr); name; name = strtok_r(NULL, ":", &saveptr)) {
+        if (!is_tls13_ciphersuite(name)) {
+            merror("Invalid TLS 1.3 cipher suite '%s' in the 'ciphers' option.", name);
+            os_free(copy);
+            return OS_INVALID;
+        }
+        suites++;
+    }
+
+    os_free(copy);
+
+    /* A list of nothing but separators tokenizes to zero suites, which would
+     * otherwise pass as vacuously valid and leave OpenSSL with an empty list. */
+    if (suites == 0) {
+        merror("Invalid 'ciphers' option: '%s' names no cipher suite.", ciphers);
+        return OS_INVALID;
+    }
+
+    return 0;
+}
+
 int Read_Agent_SSL(XML_NODE node, agent * logr)
 {
     /* XML definitions — the <agent><ssl> transport block (#37702 §10). */
@@ -478,6 +546,9 @@ int Read_Agent_SSL(XML_NODE node, agent * logr)
                 return (OS_INVALID);
             }
         } else if (strcmp(node[j]->element, xml_ciphers) == 0) {
+            if (w_client_validate_tls13_ciphers(node[j]->content) == OS_INVALID) {
+                return (OS_INVALID);
+            }
             os_free(logr->ssl.ciphers);
             os_strdup(node[j]->content, logr->ssl.ciphers);
         } else {
--- a/src/config/src/client-config.c
+++ b/src/config/src/client-config.c
@@ -453,10 +453,11 @@ static const char *VALID_TLS13_CIPHERSUITES[] = {
     NULL
 };
 
-static bool is_tls13_ciphersuite(const char *name)
+static bool is_tls13_ciphersuite(const char *name, size_t length)
 {
     for (int i = 0; VALID_TLS13_CIPHERSUITES[i]; i++) {
-        if (strcmp(name, VALID_TLS13_CIPHERSUITES[i]) == 0) {
+        if (strlen(VALID_TLS13_CIPHERSUITES[i]) == length &&
+            strncmp(name, VALID_TLS13_CIPHERSUITES[i], length) == 0) {
             return true;
         }
     }
@@ -477,36 +478,37 @@ static bool is_tls13_ciphersuite(const char *name)
  */
 static int w_client_validate_tls13_ciphers(const char *ciphers)
 {
-    char *copy = NULL;
-    char *saveptr = NULL;
-    int suites = 0;
+    const char *element = ciphers;
 
     if (!ciphers || !*ciphers) {
-        merror("Empty 'ciphers' option: expected TLS 1.3 cipher suite names.");
+        merror("Invalid 'ciphers' option: expected TLS 1.3 cipher suite names.");
         return OS_INVALID;
     }
 
-    os_strdup(ciphers, copy);
+    /* Walked by hand rather than with strtok_r, which collapses runs of
+     * separators: under it ':X', 'X::' and 'X::Y' all pass while naming a suite
+     * that is not there. Every element between two separators is checked,
+     * including an empty first or last one. */
+    for (;;) {
+        const char *separator = strchr(element, ':');
+        const size_t length = separator ? (size_t)(separator - element) : strlen(element);
 
-    for (char *name = strtok_r(copy, ":", &saveptr); name; name = strtok_r(NULL, ":", &saveptr)) {
-        if (!is_tls13_ciphersuite(name)) {
-            merror("Invalid TLS 1.3 cipher suite '%s' in the 'ciphers' option.", name);
-            os_free(copy);
+        if (length == 0) {
+            merror("Invalid 'ciphers' option: '%s' has an empty cipher suite name.", ciphers);
             return OS_INVALID;
         }
-        suites++;
+
+        if (!is_tls13_ciphersuite(element, length)) {
+            merror("Invalid TLS 1.3 cipher suite '%.*s' in the 'ciphers' option.", (int)length, element);
+            return OS_INVALID;
+        }
+
+        if (!separator) {
+            return 0;
+        }
+
+        element = separator + 1;
     }
-
-    os_free(copy);
-
-    /* A list of nothing but separators tokenizes to zero suites, which would
-     * otherwise pass as vacuously valid and leave OpenSSL with an empty list. */
-    if (suites == 0) {
-        merror("Invalid 'ciphers' option: '%s' names no cipher suite.", ciphers);
-        return OS_INVALID;
-    }
-
-    return 0;
 }
 
 int Read_Agent_SSL(XML_NODE node, agent * logr)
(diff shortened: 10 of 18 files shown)


jr0me:
## winagent and TLS 1.3: a pre-existing gap, plus one thing to fix here

The TLS commit in this PR pins `CURLOPT_SSLVERSION` to TLS 1.3 and moves `<ssl><ciphers>` to
`CURLOPT_TLS13_CIPHERS`. Both are right on Linux and macOS. On Windows neither has the intended
effect, because `https_client` reaches TLS through a different backend there — but, importantly,
**this PR is not what stops those agents connecting.**

### The backend split

OpenSSL is built for mingw (`src/external/CMakeLists.txt:55-57`) and the rest of the agent uses it —
`ssl_op.c`, enrollment, authd. libcurl does not: Windows is `--with-schannel`
(`src/external/CMakeLists.txt:272`), Linux and macOS are `--with-openssl`.

Schannel cannot negotiate TLS 1.3 before Windows 11 / Server 2022 (build 10.0.20348), and does not
declare `SSLSUPP_TLS13_CIPHERSUITES` (`schannel.c:2850-2855`), so `CURLOPT_TLS13_CIPHERS` returns
`CURLE_NOT_BUILT_IN` (`setopt.c:1893-1899`).

### Why this is not a regression

`remoted` requires TLS 1.3 and the floor is **hardcoded**:
`SSL_CTX_set_min_proto_version(context.native_handle(), TLS1_3_VERSION)` at
`RestinioHttpServer.cpp:353`, which throws if it cannot be set. There is no configuration option to
lower it — `<remote><https>` exposes `port`, `bind_addr`, `certificate`, `key`, `ca`,
`verification_mode`, `ciphers`, `max_body_size` and `dual_stack`, and nothing else.

So an agent on Windows 10, Server 2019 or older could not reach this manager before this PR either:
it offered TLS 1.2 at best and the manager refused it. What the change alters is where the failure
surfaces.

| | Before | After |
|---|---|---|
| Connection | agent offers ≤ TLS 1.2, manager refuses — generic handshake failure | agent refuses first: `schannel: TLS 1.3 not supported on Windows prior to 11` |
| `<ssl><ciphers>` | reaches `CURLOPT_SSL_CIPHER_LIST`, which Schannel supports but which only governs TLS ≤ 1.2 — a session this manager will not establish | returns `CURLE_NOT_BUILT_IN` |

Same outcome, earlier and with a clearer diagnosis. `<ciphers>` was already inert on Windows in
practice; it is now inert by construction.

### What is worth acting on

**1. winagent cannot talk to this manager at all below Windows 11 / Server 2022.** That is a product
gap independent of this PR, and there are two honest ways out: build winagent's curl
`--with-openssl`, which is the only option that also makes `<ciphers>` meaningful there but changes
certificate validation away from the Windows store and so needs a packaging review; or declare
Windows 11 / Server 2022 the supported floor and enforce it with a clear startup error. Keeping
Schannel and choosing the TLS floor per backend is not a third option — negotiating 1.2 does not help
when the manager will not accept it.

**2. `curlPerformer::applyTls()` discards `setOption` results** (`curlPerformer.cpp:226`).
`CurlHandle::setOptionString` already surfaces the failure as `false` and the caller drops it, so a
TLS option that cannot be applied is skipped silently. That one belongs in this PR's scope: a
security option that fails to apply should fail the connection.

### Why CI does not show any of this

The component tests that exercise real curl are excluded from Windows
(`src/client-agent/https_client/tests/CMakeLists.txt:58`, `if(NOT CMAKE_SYSTEM_NAME STREQUAL
"Windows")`), and the unit tests assert against a mocked handle. Neither the refused handshake nor
the rejected option can appear there.

jr0me:
## Validation evidence for #38163

Re-run under the `<agent>` root now that PR #38172 (agent remote upgrade / WPK, which carried the
`<client>` to `<agent>` rename) has merged. Every case drives the real
`wazuh-agentd` reading `ossec.conf`, so the XML parser is in the loop, and each runs in isolation:
fresh manager, agent in the foreground under `timeout`, and a check that no daemon survived the
previous case.

### Environment

Agent and manager both built and installed from this branch (`73c0db8d1c`, on `5.0.0-https`
`72f79c6ad8`).

The TLS matrix runs against a **real manager**: `remoted`'s HTTPS listener on `:27000`, configured
with its own PKI (CA, server certificate `SAN: DNS:wazuh-server, IP:172.19.0.3`, client certificate
`CN=agent-001`), `<ca>` + `<verification_mode>certificate</verification_mode>` so it demands a client
certificate, and `<ciphers>TLS_AES_256_GCM_SHA384</ciphers>`.

Cases needing per-request timing or body inspection run against the repo's mock manager
(`https_client/demo/mock_manager.py`), which the issue's Environment section allows.

### (a) Configuration block, one option at a time, under `<agent>`

| # | Option | Result | Against |
|---|---|---|---|
| a1 | `<server><address>`/`<port>` | PASS | real |
| a2 | `<ssl><certificate>`/`<key>` — manager requires a client certificate | PASS | real |
| a3 | `<ssl><certificate_authorities>`/`<verification_mode>` full / certificate / none | PASS | real |
| a4 | `<ssl><ciphers>` | PASS | real |
| a5 | `<batch><interval>` | PASS | real + mock |
| a6 | `<config-profile>` | PASS | real |
| a7 | `<notify_time>` | PASS | real + mock |
| a8 | `<auto_restart>` | PASS | mock |

**a1** — the block parses under `<agent>` (it was rejected outright before PR #38172) and the configured
endpoint is the one used:

```
INFO: Starting https_client (server=wazuh-server:27000, agent=001).
DEBUG: https_client connection state -> registered (2)
DEBUG: https_client startup accepted: {"agent":{"groups":["default"]},
       "cluster":{"name":"wazuh","node":"node01"},"limits":{...}}
```

**a2 — mutual TLS**, the case the issue notes prior rounds never exercised. Same manager, the only
difference being whether the agent is given `<certificate>`/`<key>`:

| Agent config | `registered` | `startup accepted` |
|---|---|---|
| certificate + key present | 1 | 1 |
| no client certificate | 0 | 0 |

Isolated with `curl` so the enforcement is demonstrably the manager's:

```
curl --cacert ca.crt https://wazuh-server:27000/control
  -> curl: (56) OpenSSL SSL_read: ... tlsv13 alert certificate required
curl --cacert ca.crt --cert client.crt --key client.key https://wazuh-server:27000/control
  -> {"error":"not_found"}          # TLS accepted; GET is simply not a route
```

**a3** — all three modes against a manager whose SAN really matches, plus the negative:

| `<verification_mode>` | `registered` / `startup accepted` |
|---|---|
| `full`, matching CA + SAN | 1 / 1 |
| `certificate` | 1 / 1 |
| `none` | 1 / 1 |
| `full`, unrelated CA | **0 / 0** |

**a4 — `<ciphers>`**, the item the issue words as "handshake fails against a manager offering only
excluded ciphers". The manager is pinned to `TLS_AES_256_GCM_SHA384`; only the agent's list changes:

| Agent `<ciphers>` | States reached | `startup accepted` |
|---|---|---|
| `TLS_AES_256_GCM_SHA384` (intersects) | starting → **registered** | 1 |
| `TLS_CHACHA20_POLY1305_SHA256` (disjoint) | starting → stopped | **0** |

Control, independent of the agent:

```
openssl s_client -ciphersuites TLS_AES_256_GCM_SHA384        -> New, TLSv1.3, Cipher is TLS_AES_256_GCM_SHA384
openssl s_client -ciphersuites TLS_CHACHA20_POLY1305_SHA256  -> sslv3 alert handshake failure (alert 40)
```

This test could not pass before the TLS commit in this PR: `<ciphers>` reached
`CURLOPT_SSL_CIPHER_LIST`, which libcurl applies to TLS 1.2 and below, so it never constrained a
TLS 1.3 session and both rows would have connected.

**a5 — `<batch><interval>`**, differential, identical config otherwise:

| `<interval>` | startup | `/stateless` flush | delta |
|---|---|---|---|
| 3s | 188 ms | 3240 ms | 3052 ms |
| 10s | 197 ms | 10250 ms | 10053 ms |

Confirmed against the real manager at the coarser granularity its logs allow — with `<interval>3s`,
startup at `18:20:44` and the first `/stateless` flush at `18:20:47`.

**a7 — `<notify_time>`** set to 5s. Consecutive notifies (the 222-byte host-metadata bodies, as
distinct from startup and shutdown, which share the endpoint):

```
delta=5055ms  delta=5096ms  delta=5051ms
```

Confirmed against the real manager too. `remoted` does not log periodic notifies at
`remoted.debug=2`, so the cadence there is measured from the agent's own send timestamps, with
`<notify_time>5</notify_time>`:

```
18:20:45  18:20:50  18:20:55  18:21:00  18:21:05  18:21:10  18:21:15  18:21:20
```

Exactly 5 s apart. The millisecond figures above remain the precise measurement; this is the same
behaviour against a real manager at one-second log granularity.

**a8 — `<auto_restart>`**, same stimulus both runs (the manager changes a module limit mid-session),
only the flag differs:

```
yes -> INFO:  https_client: reloading due to module limits changes.   (bridge_on_startup_result:381)
no  -> DEBUG: https_client: module limits have been updated.          (line 384, no reload)
```

**a6 — `<config-profile>`**, both halves. It comes back verbatim in the `/config` document
(`"config-profile": "webserver, debian8"`), and it is honoured where it is consumed: group profile
matching.

The manager's group `agent.conf` carries two sections, one `profile="webserver"` and one
`profile="database"`, each adding a distinct `<localfile>`. The agent declares
`<config-profile>webserver, debian8</config-profile>` and downloads that shared configuration. The
decision comes from `config.c` itself:

```
os_read_agent_profile() = [webserver, debian8]
config.c:390  Matched agent config profile name [webserver, debian8]
config.c:387  [database] did not match agent config profile name [webserver, debian8]
```

And the effect is observable rather than only logged — logcollector's own state reports the monitored
locations, and only the matching profile's file is there:

```
monitored locations:
  /tmp/profile-webserver.log        # profile="webserver"  -> applied
                                    # profile="database"   -> absent
```

### (b) Endpoints share the validated connection

One run with the full daemon set and both periodic pushes enabled:

| Endpoint | Requests |
|---|---|
| `/control` | 10 |
| `/stateless` | 2 |
| `/stateful` | 3 |
| `/stats` | 8 |
| `/config` | 6 |
| `/download` | 2 |

Every request carried the same `agent_id=001`, and a single TLS suite was negotiated throughout
(`cipher=TLS_AES_256_GCM_SHA384 proto=TLSv1.3`). No endpoint used different or unvalidated
connection parameters. The only 401s in the trace are the mock's own deliberate key-rotation test.

`/stateful` only appears when the whole daemon set runs: with `wazuh-agentd` alone no module produces
a sync session, and the endpoint is silently never exercised.

**Against the real manager, all six routes now exist.** PR #38207 registered `/stateful`
(`remotedModuleFacade.hpp:346`, closing issue #38089); before it, that endpoint 404'd. Probed directly over
mutual TLS, with a nonsense path as the control so a 404 is meaningful:

```
/control  /stateless  /stateful  /stats  /config  /download   -> 400 (route exists, body rejected)
/nonsense-control                                             -> 404 (not registered)
```

Agent-driven over the same connection (`server=wazuh-server:27000, agent=001`, mutual TLS,
`verification_mode full`), four of the six were exercised end to end in one window:

| Endpoint | Requests |
|---|---|
| `/control` | 27 (1 startup, 25 notify, 1 shutdown) |
| `/stateless` | 4 |
| `/stateful` | 1 (`Sending /stateful session fim`) |
| `/download` | 1 |
| `/stats` | 0 |
| `/config` | 0 |

`/stateful` completed the round trip and the manager answered:

```
agent:   Sending /stateful session fim-7315975581387653120 (1712 bytes)
agent:   /stateful session=fim-7315975581387653120 result=503 (42 byte answer)
remoted: Downstream call to the inventory sync server failed (connect_failed)
```

The 503 is the correct answer here, not a failure of the endpoint: `inventory_sync_server` refuses to
start in this environment (`the indexer session configuration is invalid (No hosts found in the
configuration)`) because no indexer runs, and remoted reports that to the agent. It is also PR #38207's
model working as designed — the HTTP response *is* the session result. The request was signed,
authenticated, routed and answered over the validated connection, which is what this issue asks of
the endpoint; whether a sync then succeeds is issue #38162's scope, not this one.

`/stats` and `/config` did not fire inside that window: the agent restarted twice (three distinct
`wazuh-agentd` pids), so the reporter never reached its 10 s interval. Their routes are present and
answer, as the probe above shows, and the mock-based run is the evidence for the documents they
carry.

### What the documents actually contain

Worth recording, because "the push happened" and "the push carried something" are different claims.

`/config` carries eight modules on every push: `agent`, `agent-info`, `agent-upgrade`, `execd`,
`fim`, `logcollector`, `wazuh_control`, `wmodules`.

`/stats` carries the `agent` module on every push, and `logcollector` on the pushes where it has
something to report. With two localfiles configured — a `full_command` and a file being written to —
the logcollector branch populates in full:

```json
"logcollector": {
  "global": {
    "start": "2026-08-07T16:44:56Z", "end": "2026-08-07T16:45:11Z",
    "files": [
      { "location": "/tmp/lcwatch/app.log", "events": 3, "bytes": 165,
        "targets": [ { "name": "agent", "drops": 0 } ] },
      { "location": "df -P", "events": 2, "bytes": 1190,
        "targets": [ { "name": "agent", "drops": 0 } ] }
    ]
  },
  "interval": { "…": "same shape, same window" }
}
```

That is the whole shape the index template declares for this module — `global`/`interval`,
`files[].location/events/bytes`, and the nested `targets[].name/drops`.

**It is intermittent by design, and that is worth knowing before anyone reads an agent-only document
as a fault.** Three things have to line up for logcollector to appear:

- something must be configured to collect. With no `<localfile>` the module has nothing to account
  for, and `/stats` stays agent-only indefinitely.
- `logcollector.state_interval` (60 s by default) must have elapsed. Before the first publish the
  module answers `ok {}`, an empty body that `module_report_add()` drops rather than push a module
  with nothing in it — so an agent whose session is shorter than the interval never contributes one.
- the window must have had activity. `w_logcollector_state_generate()` serializes with `restart=true`,
  zeroing the counters, so a quiet window produces an empty body again. In the run above only the
  second of three pushes carried the module, for exactly that reason.

`<localfile>` syntax note, since it silently costs a whole module: `full_command` takes `<command>`,
not `<location>`. Given the wrong element the entry is rejected at parse time with
`ERROR: Missing 'command' argument. This option will be ignored.` and logcollector collects nothing.

### Not covered

- **A successful `/stateful` sync, and `/stats` / `/config` documents landing in their indices** —
  all three forward to modulesd's `inventory_sync_server`, which needs a configured indexer to
  start. That is payload semantics rather than the connection layer, so it belongs to issues #38162 and
  #38164; the connection evidence this issue asks for is complete.
- **Windows and macOS** — the issue's platform row is Linux / Windows / macOS; all evidence here is
  Linux. This matters more than usual for winagent, which has a known TLS regression from this PR's
  own TLS 1.3 change (see the separate comment on this PR).
- **Package installs** — the issue's install-method row is Sources / Packages; everything here was
  built from source.

### Acceptance criteria

- [x] **(a)** every option has a positive functional test under `<agent>`, not parser-level
      coverage. All eight items covered, four of them against a real manager.
- [x] **(b)** a trace showing at least one request per endpoint over the same validated connection.
      All six routes exist on a real manager and answer over it; `/control`, `/stateless`,
      `/stateful` and `/download` were additionally exercised by the agent in one window, on one
      `agent_id` and one negotiated suite. No endpoint used different or unvalidated settings.
- [x] **(c)** the one option that parsed with no observable runtime effect, `<ssl><ciphers>`, is
      fixed in this PR rather than only filed, and a4 above is the proof. No endpoint was found
      bypassing the validated connection config.


---

## Raw output

Unedited, as captured.

### a1 — <agent><server><address>/<port>, real manager

```
2026/08/07 15:32:36 wazuh-agentd:https-client[69387] httpsClientFacade.cpp:108 at start(): INFO: Starting https_client (server=wazuh-server:27000, agent=001).
2026/08/07 15:32:36 wazuh-agentd[69387] https_client_bridge.c:1189 at bridge_on_state_change(): DEBUG: https_client connection state -> starting (1)
2026/08/07 15:32:36 wazuh-agentd[69387] https_client_bridge.c:1189 at bridge_on_state_change(): DEBUG: https_client connection state -> registered (2)
2026/08/07 15:32:36 wazuh-agentd[69387] https_client_bridge.c:354 at bridge_on_startup_result(): DEBUG: https_client startup accepted: {"agent":{"groups":["default"]},"cluster":{"name":"wazuh","node":"node01"},"limits":{"fim":{"file":30000,"registry_key":30000,"registry_value":30000},"sca":{"checks":30000},"syscollector":{"browser_extensions":30000,"groups":30000,"hardware":30000,"hotfixes":30000,"network_address":30000,"network_iface":30000,"network_protocol":30000,"os_info":30000,"packages":30000,"ports":30000,"processes":30000,"services":30000,"users":30000}}}
```

### a2 — <ssl><certificate>/<key>, mutual TLS, real manager

```
$ curl --cacert mtls_ca.crt https://wazuh-server:27000/control
curl: (56) OpenSSL SSL_read: error:0A00045C:SSL routines::tlsv13 alert certificate required, errno 0

$ curl --cacert mtls_ca.crt --cert mtls_client.crt --key mtls_client.key https://wazuh-server:27000/control
{"error":"not_found"}
```

### a4 — <ssl><ciphers>, real manager pinned to TLS_AES_256_GCM_SHA384

Agent <ciphers>TLS_AES_256_GCM_SHA384</ciphers>:
```
2026/08/07 15:32:36 wazuh-agentd[69387] https_client_bridge.c:1189 at bridge_on_state_change(): DEBUG: https_client connection state -> starting (1)
2026/08/07 15:32:36 wazuh-agentd[69387] https_client_bridge.c:1189 at bridge_on_state_change(): DEBUG: https_client connection state -> registered (2)
2026/08/07 15:32:36 wazuh-agentd[69387] https_client_bridge.c:354 at bridge_on_startup_result(): DEBUG: https_client startup accepted: {"agent":{"groups":["default"]},"cluster":{"name":"wazuh","node":"node01"},"limits":{"fim":{"file":30000,"registry_key":30000,"registry_value":30000},"sca":{"checks":30000},"syscollector":{"browser_extensions":30000,"groups":30000,"hardware":30000,"hotfixes":30000,"network_address":30000,"network_iface":30000,"network_protocol":30000,"os_info":30000,"packages":30000,"ports":30000,"processes":30000,"services":30000,"users":30000}}}
```

Agent <ciphers>TLS_CHACHA20_POLY1305_SHA256</ciphers>:
```
2026/08/07 15:33:10 wazuh-agentd[69425] https_client_bridge.c:1189 at bridge_on_state_change(): DEBUG: https_client connection state -> starting (1)
2026/08/07 15:33:25 wazuh-agentd[69425] https_client_bridge.c:1189 at bridge_on_state_change(): DEBUG: https_client connection state -> stopped (0)
```

### a6 — <config-profile>, group profile matching, real manager

```
2026/08/07 19:39:08 wazuh-agentd[101692] config.c:390 at ReadConfig(): DEBUG: Matched agent config profile name [webserver, debian8]
2026/08/07 19:39:08 wazuh-agentd[101692] config.c:387 at ReadConfig(): DEBUG: [database] did not match agent config profile name [webserver, debian8]
```

### (b) — routes on the real manager

```
curl -X POST https://wazuh-server:27000/control            ->  400
curl -X POST https://wazuh-server:27000/stateless          ->  400
curl -X POST https://wazuh-server:27000/stateful           ->  400
curl -X POST https://wazuh-server:27000/stats              ->  400
curl -X POST https://wazuh-server:27000/config             ->  400
curl -X POST https://wazuh-server:27000/download           ->  400
curl -X POST https://wazuh-server:27000/nonsense-control   ->  404
```

### metadata provider fix — segment seeded with zeros by root, then agentd ran

```
$ ls -l /var/ossec/var/run/.wazuh_agent_metadata
-rw-r--r-- 1 root wazuh 35616 Aug  7 21:47 /var/ossec/var/run/.wazuh_agent_metadata
$ strings -n 2 /var/ossec/var/run/.wazuh_agent_metadata
001
agent-001
v5.0.0
aarch64
wazuh-agent
Ubuntu
linux
ubuntu
22.04.5 LTS (Jammy Jellyfish)
wazuh
node01
default
```

Nicogp (approved):

jr0me:
## Evidence — agent config templates and `notify_time`

Covers the two fixes taken from QA's exploratory session, wazuh/wazuh-qa-automation#8451 (F3), on
top of the connection/configuration validation already reported on this PR.

Everything below was produced from this branch at `6c5a3dc140`, built in a container from the
branch tree (`make TARGET=agent DEBUG=1 TEST=1`, then a plain `DEBUG=1` build for the daemon).
Runtime cases drive the real `wazuh-agentd` reading `ossec.conf`, so the XML parser is in the loop.

### The `PATH` token is never substituted

The claim behind the fix is that nothing anywhere sets it. Both greps over the whole tree,
excluding `src/external`:

```
$ grep -rn ">PATH<" --include='*.conf' --include='*.sh' --include='*.xml' --include='*.template' .
(no matches outside the three templates)

$ grep -rn "certificate_authorities" --include='*.sh' --include='*.yml' --include='*.py' .
./src/init/inst-functions.sh:373:    # provide certificate_authorities to verify the manager's certificate.
./src/init/inst-functions.sh:375:    echo "      <certificate_authorities>PATH</certificate_authorities>" >> $NEWCONFIG
```

One writer, no reader, no substitution step. Contrast `MANAGER_IP`, which **is** substituted —
`install.sh` fills it from `USER_AGENT_MANAGER_IP`, and three packaging smoke tests `sed` it to
`1.1.1.1`. That token is working as designed and is not changed here.

### What a new install now gets

Real generator, not a hand-written sample:

```
$ sh src/init/gen_wazuh.sh conf agent ubuntu 22 /var/ossec
<ossec_config>
  <agent>
    <server>
      <address>MANAGER_IP</address>
      <port>1517</port>
    </server>
    <ssl>
      <!-- <certificate_authorities>etc/certs/root-ca.pem</certificate_authorities> -->
      <verification_mode>none</verification_mode>
    </ssl>
    <config-profile>ubuntu, ubuntu22</config-profile>
    <notify_time>10</notify_time>
    <auto_restart>yes</auto_restart>
  </agent>
```

### Runtime behaviour, all three states

`wazuh-agentd -t` against each config. Verbatim, nothing removed:

```
==============================================================
CASE: shipped template (CA commented out, mode=none)
--------------------------------------------------------------
  conf:       <!-- <certificate_authorities>etc/certs/root-ca.pem</certificate_authorities> -->
  conf:       <verification_mode>none</verification_mode>
  --- wazuh-agentd -t output (verbatim) ---
  2026/08/07 22:43:19 wazuh-agentd[138013] debug_op.c:305 at os_logging_config(): DEBUG: (1228): Element 'log_format' without any option.
  2026/08/07 22:43:19 wazuh-agentd[138013] debug_op.c:116 at _log_function(): DEBUG: Logging module auto-initialized
  2026/08/07 22:43:19 wazuh-agentd[138013] main.c:187 at main(): DEBUG: Wazuh home directory: /var/ossec
  2026/08/07 22:43:19 wazuh-agentd[138013] main.c:189 at main(): DEBUG: Started (pid: 138013).
  2026/08/07 22:43:19 wazuh-agentd[138013] agent_op.c:121 at os_read_agent_profile(): DEBUG: Calling os_read_agent_profile().
  2026/08/07 22:43:19 wazuh-agentd[138013] agent_op.c:140 at os_read_agent_profile(): DEBUG: os_read_agent_profile() = [-]
  2026/08/07 22:43:19 wazuh-agentd[138013] config.c:387 at ReadConfig(): DEBUG: [webserver] did not match agent config profile name [-]
  2026/08/07 22:43:19 wazuh-agentd[138013] agent_op.c:121 at os_read_agent_profile(): DEBUG: Calling os_read_agent_profile().
  2026/08/07 22:43:19 wazuh-agentd[138013] agent_op.c:140 at os_read_agent_profile(): DEBUG: os_read_agent_profile() = [-]
  2026/08/07 22:43:19 wazuh-agentd[138013] config.c:387 at ReadConfig(): DEBUG: [database] did not match agent config profile name [-]
  exit=0

==============================================================
CASE: old template (literal PATH, mode=none) -- expect (4119) WARNING
--------------------------------------------------------------
  conf:       <certificate_authorities>PATH</certificate_authorities>
  conf:       <verification_mode>none</verification_mode>
  --- wazuh-agentd -t output (verbatim) ---
  2026/08/07 22:43:19 wazuh-agentd[138018] debug_op.c:305 at os_logging_config(): DEBUG: (1228): Element 'log_format' without any option.
  2026/08/07 22:43:19 wazuh-agentd[138018] debug_op.c:116 at _log_function(): DEBUG: Logging module auto-initialized
  2026/08/07 22:43:19 wazuh-agentd[138018] main.c:187 at main(): DEBUG: Wazuh home directory: /var/ossec
  2026/08/07 22:43:19 wazuh-agentd[138018] main.c:189 at main(): DEBUG: Started (pid: 138018).
  2026/08/07 22:43:19 wazuh-agentd[138018] agent_op.c:121 at os_read_agent_profile(): DEBUG: Calling os_read_agent_profile().
  2026/08/07 22:43:19 wazuh-agentd[138018] agent_op.c:140 at os_read_agent_profile(): DEBUG: os_read_agent_profile() = [-]
  2026/08/07 22:43:19 wazuh-agentd[138018] config.c:387 at ReadConfig(): DEBUG: [webserver] did not match agent config profile name [-]
  2026/08/07 22:43:19 wazuh-agentd[138018] agent_op.c:121 at os_read_agent_profile(): DEBUG: Calling os_read_agent_profile().
  2026/08/07 22:43:19 wazuh-agentd[138018] agent_op.c:140 at os_read_agent_profile(): DEBUG: os_read_agent_profile() = [-]
  2026/08/07 22:43:19 wazuh-agentd[138018] config.c:387 at ReadConfig(): DEBUG: [database] did not match agent config profile name [-]
  2026/08/07 22:43:19 wazuh-agentd[138018] config.c:100 at w_agent_validate_ssl_ca(): WARNING: (4119): <ssl><certificate_authorities> is not a readable file: 'PATH'. It is unused while <verification_mode> is 'none'; enabling verification would stop the agent from starting.
  exit=0

==============================================================
CASE: literal PATH + mode=full -- expect (4118) ERROR and refusal to start
--------------------------------------------------------------
  conf:       <certificate_authorities>PATH</certificate_authorities>
  conf:       <verification_mode>full</verification_mode>
  --- wazuh-agentd -t output (verbatim) ---
  2026/08/07 22:43:19 wazuh-agentd[138023] debug_op.c:305 at os_logging_config(): DEBUG: (1228): Element 'log_format' without any option.
  2026/08/07 22:43:19 wazuh-agentd[138023] debug_op.c:116 at _log_function(): DEBUG: Logging module auto-initialized
  2026/08/07 22:43:19 wazuh-agentd[138023] main.c:187 at main(): DEBUG: Wazuh home directory: /var/ossec
  2026/08/07 22:43:19 wazuh-agentd[138023] main.c:189 at main(): DEBUG: Started (pid: 138023).
  2026/08/07 22:43:19 wazuh-agentd[138023] agent_op.c:121 at os_read_agent_profile(): DEBUG: Calling os_read_agent_profile().
  2026/08/07 22:43:19 wazuh-agentd[138023] agent_op.c:140 at os_read_agent_profile(): DEBUG: os_read_agent_profile() = [-]
  2026/08/07 22:43:19 wazuh-agentd[138023] config.c:387 at ReadConfig(): DEBUG: [webserver] did not match agent config profile name [-]
  2026/08/07 22:43:19 wazuh-agentd[138023] agent_op.c:121 at os_read_agent_profile(): DEBUG: Calling os_read_agent_profile().
  2026/08/07 22:43:19 wazuh-agentd[138023] agent_op.c:140 at os_read_agent_profile(): DEBUG: os_read_agent_profile() = [-]
  2026/08/07 22:43:19 wazuh-agentd[138023] config.c:387 at ReadConfig(): DEBUG: [database] did not match agent config profile name [-]
  2026/08/07 22:43:19 wazuh-agentd[138023] config.c:109 at w_agent_validate_ssl_ca(): ERROR: (4118): <ssl><verification_mode> is not 'none' but <certificate_authorities> is missing or unreadable: 'PATH'.
  2026/08/07 22:43:19 wazuh-agentd[138023] main.c:216 at main(): ERROR: (1215): No client configured. Exiting.
  exit=1
```

Reading these in order:

- **Case 1** is the one that could have broken every new install. An XML comment inside `<ssl>` has
  to parse — it does, silently, exit 0.
- **Case 2** is exactly what QA found in the field. Same config as before this PR; previously the
  agent said nothing at all about it. Still starts, because under `none` the CA genuinely is unused.
- **Case 3** is the foot-gun. 4118 already existed; it is shown here to make the point that the
  warning in case 2 is a warning about a real future failure, not noise.

### Unit tests

```
$ ctest -R test_agent_ssl_ca --output-on-failure
Test project /wazuh-build/src/unit_tests/build
    Start 6: test_agent_ssl_ca
1/1 Test #6: test_agent_ssl_ca ................   Passed    0.03 sec

100% tests passed, 0 tests failed out of 1
```

Seven cases over `w_agent_validate_ssl_ca()`: both verification modes against a CA that is absent,
readable and unreadable, asserting the exact 4118 and 4119 text.

Full suite on this branch:

```
100% tests passed, 0 tests failed out of 107
```

106 before this change plus the new one, so nothing regressed — the new warning constant is in the
shared `warning_messages.h`, which every target compiles.

### Not covered

- **winagent is not compiled here.** `src/win32/win_utils.c` is one of the two callers changed, and
  no mingw toolchain was available. What was checked statically: `config.c` is inside the
  `file(GLOB AGENTD_SRC ...)` that feeds the Windows target, so the shared function links there
  too, and `agentd.h` already declares seven other `bool`-returning functions, so the type is in
  scope. CI is the first real compile.
- **macOS**, same as the rest of this PR.

Nicogp (approved):

PR #38188, opened 2026-08-04
Revert test/lint workflows from CodeBuild to GitHub-hosted runners

lchico:
## Description

As part of the CI/CD cost-audit tracked in #38149, the agent-side test/lint workflows had previously been migrated from GitHub-hosted runners to AWS CodeBuild. This PR reverts that migration for the light/standard workloads (linters, unit test suites, static analysis, component/integration RTR tests) back to GitHub-hosted runners, since they run comfortably within standard runner specs and don't require AWS VPC/IAM access — removing their AWS CodeBuild compute cost.

Closes #38149

## Proposed Changes

- Changed `runs-on:` in 18 workflow files from CodeBuild-hosted runner labels (`codebuild-github-actions-codebuild-runner-*`) back to standard GitHub-hosted runner labels (e.g. `ubuntu-24.04`, `windows-2025`).
- Removed CodeBuild-only workaround steps that are unnecessary on GitHub-hosted images:
  - WiX Toolset and 7-Zip manual setup steps in `5_testcomponent_windows-installer-upgrade.yml` (both ship pre-installed on GitHub-hosted Windows images).
  - A conditional `cmake` install shim in `.github/actions/install_build_deps/action.yml` (GitHub-hosted Linux images ship `cmake`; the CodeBuild image did not).
- Added a fallback mirror for the `curl` download of CMocka's source tarball in `.github/actions/install_build_deps/action.yml`: if the primary download from `git.cryptomilk.org` fails or returns an invalid archive, the step now retries against a GitLab mirror before failing the job.

> [!NOTE]
> **Why a mirror was needed:** validating this migration surfaced that `git.cryptomilk.org` (CMocka's upstream source host) is unreliable from GitHub-hosted runners — sometimes timing out on every connection attempt, sometimes accepting the connection but returning a short, corrupt (non-gzip) response instead of the real tarball. This started as 6 of 18 `winagent` jobs failing and escalated to *every* `winagent` job failing once this PR's full `pull_request` check matrix ran ~10 of them in parallel — consistent with a low-capacity, self-hosted `cgit` box being unable to handle that concurrent load (or rate-limiting it defensively). We independently confirmed the host is unreachable even from a third, unrelated network, ruling out a GitHub-IP-specific block. Since neither retries, IPv4/IPv6 switching, nor backoff could fix a server that won't accept the connection at all, the step now falls back to a mirror on `gitlab.com/cmocka/cmocka` (tag `cmocka-1.1.8`) instead.
 >
[...]

PR #38215, opened 2026-08-06
Rightsize codebuild runners

### LINKED ISSUE #38150: Rightsizing CodeBuild Compute Environments (Server)
## Description
AWS CodeBuild charges per compute-minute based on the instance size selected. Running lightweight steps (such as unit tests, script execution, or simple packaging) on over-provisioned hardware (`BUILD_GENERAL1_LARGE` or higher) leads to unnecessary cloud spend. This task focuses on auditing actual resource usage and downsizing build environments to match real resource requirements or migrating to cost-effective ARM/Graviton architectures.

---

### 1. Metrics & Usage Audit
- [x] Gather CloudWatch metrics (CPU, Memory, and Disk I/O) across all active AWS CodeBuild project builds over the last 30 days.
- [x] Identify projects with low average peak utilization (e.g., peak CPU < 30%, RAM < 4 GB on a Large instance).

### 2. Compute Tier Adjustments
- [ ] Downsize identified projects from higher tiers (`LARGE` / `X2_LARGE`) to standard or small tiers (`SMALL` / `MEDIUM`). — N/A at project level: both Server projects are already `general1.medium`, and a blanket downsize of the `-j$(nproc)` compile jobs trades latency for money instead of saving it. Replaced by per-job `instance-size` label overrides ([audit](https://github.com/wazuh/wazuh/pull/38215#issuecomment-5202897416)).
- [x] Evaluate switching Linux builds to ARM/Graviton-based instances (`BUILD_LAMBDA_` or `BUILD_GENERAL1_SMALL` on ARM) which offer a ~20% better price-performance ratio.

### 3. Performance Impact Testing
- [x] Run benchmark builds on downsized environments to measure impact on build duration.
- [x] Ensure downsized instances do not introduce out-of-memory (OOM) errors or significant build time regression that offsets compute savings.

### 4. Monitoring & Cost Validation
- [ ] Set up basic CloudWatch alarms or tracking to alert if downsized jobs hit resource limits (CPU throttling / RAM exhaustion). — blocked: needs credentials for the CI AWS account. In-job measurement stands in for now: peak 1.99 GiB of the 3.00 GiB container limit on the heaviest downsized job.
- [x] Calculate estimated monthly savings resulting from the hardware tier adjustments.

---

## Related:
- CodeBuild Projects information: https://github.com/wazuh/wazuh-automation/issues/3637#issuecomment-5106583282
- Server project metrics: 
  - [server-amd](https://github.com/wazuh/wazuh-automation/issues/3637#issuecomment-5132179979)
  - [server-arm](https://github.com/wazuh/wazuh-automation/issues/3637#issuecomment-5132068000)

---

## Acceptance Criteria
- [x] Comprehensive audit log of current vs. recommended CodeBuild compute tiers for all projects.
- [ ] Over-provisioned projects updated to appropriate hardware sizes. — PR #38215 (draft, awaiting review).
- [x] Build execution times verified to remain within acceptable thresholds after downsizing.
- [x] Estimated monthly cost savings documented and reported.

## DRI
- @teddytpc1

---

**Per-job audit and findings:** https://github.com/wazuh/wazuh/pull/38215#issuecomment-5202897416 (PR #38215)

jepalfer:
## Description

Related to #38150.

`github-actions-codebuild-runner-server-amd` and `-server-arm` are configured as `BUILD_GENERAL1_MEDIUM` (4 vCPU / 8 GB) and every job targeting them gets that instance, whether it can use it or not. A per-job audit of where that project's spend actually goes (posted as a comment below) shows most of it is `-j$(nproc)` C/C++ compilation that genuinely needs the cores — but a handful of jobs cannot use more than 2 vCPU by construction and were paying the medium rate ($0.0100/min) anyway.

This PR overrides the instance size to `small` (2 vCPU / 3 GB usable, $0.0050/min) for those jobs only. Two further jobs were downsized and then reverted after repeated runs showed a latent test race failing on 2 vCPU — the saving is correspondingly modest, and the results comment below documents exactly what was measured and rejected. Compile jobs that scale with `$(nproc)` are deliberately left on medium: halving their rate roughly doubles their duration, which trades PR feedback latency for money rather than saving it, and would breach several existing `timeout-minutes` values.

## Proposed Changes

Appended an `instance-size` override to the CodeBuild runner label of the four Server-project jobs that cannot use four cores. Each was validated by repeated execution, not a single run (counts in the results comment below):

| Workflow / job | Why `small` is safe | Mean duration |
|---|---|---|
| `5_testcomponent_keystore.yml` `keystore-tests` | `build_target_cpp` runs `cmake --build` with no `--parallel`; dependency build pinned to `-j2` in `build_external_deps` — **10/10 per matrix leg on small** | 4.6 / 4.8 min |
| `5_codequality_changelog.yml` `verify-changelog` | changelog diff only, no compilation | <1 min |
| `5_codeanalysis_python_bandit_pr.yml` `bandit-scan` | single-process bandit | <1 min |
| `5_builderpackage_engine-standalone.yml` `check_standalone_engine` | downloads and starts a prebuilt engine, no compilation — **5/5 per architecture on small** | 0.8 / 0.3 min |

The override uses the single-label form documented by AWS:

```yaml
runs-on: codebuild-github-actions-codebuild-runner-server-amd-${{ github.run_id }}-${{ github.run_attempt }}-small
```

This is deliberate rather than a separate `- instance-size:small` label: a differing label count between jobs of the same workflow run lets one job get picked up by the runner created for another, [per the AWS docs](https://docs.aws.amazon.com/codebuild/latest/userguide/sample-github-action-runners-update-labels.html). That rationale is written into `5_testcomponent_keystore.yml` next to the label rather than living only here — the failure mode is hanging jobs, and it is invisible to anyone tidying the label toward the more prominently documented multi-label form. No CodeBuild project, fleet or IAM change is needed for this to take effect, and any branch without this patch keeps using the project default.

Two further files change comments only, no behaviour: `5_testunit_shared_modules_utils.yml` and `5_testintegration_linux-integration-tests.yml` each gain a `NOTE` recording that downsizing was measured and rejected, with the data linked. Both rejections cost real CI time to establish — the integration-test one took a 57-minute run — so the next person to look at these jobs should not have to rediscover them.

> [!NOTE]
> **Scope on this branch.** #38188 already moved the agent-side light workloads on `5.0.0` off CodeBuild to GitHub-hosted runners, so the jobs listed above are what remains downsizable here. `main` still runs five more jobs in the same category (`5_testcomponent_windows-{dll-hijack,files,installer-upgrade}` `build`, which pin `make` to `-j2`, plus `5_codelinter_clangformat.yml` and `5_testunit_wodles.yml`) — worth the same patch as a follow-up if `main` keeps them on CodeBuild.
>
> **Sizing vs. migrating.** For these five jobs specifically, following #38188 and moving them to `ubuntu-24.04` would save 100% of their minutes instead of 50%, since `wazuh/wazuh` is public and standard runners bill nothing. That is a concurrency trade-off rather than a sizing one, so it is left out of this PR and raised in the audit comment instead.

### The other eight server-runner jobs, and why they stay put

For completeness — there are 13 jobs on the two Server projects. Four are downsized above; here is every one that is not, so the sweep is auditable rather than implied:

| Job | Why it keeps the project default |
|---|---|
| `5_builderpackage_manager.yml::Build-manager-packages` | `-j$(nproc)` package build — needs the cores |
| `5_builderpackage_engine-standalone.yml::build` | `-j$(nproc)` engine build |
| `5_buildpackage_internal_tools.yml::build` | `-j$(nproc)` build |
| `6_builderpackage_server.yml::Build-server-packages` | `-j$(nproc)` build (ARM) |
| `5_testintegration_linux-integration-tests.yml::run-test` | **measured and rejected** — `fim` tier-0 fails a real-time capacity assertion on `small`; see the A/B comment below |
| `5_testunit_shared_modules_utils.yml::shared_modules_utils-modules` | **downsized, then reverted** — `AsyncValueDispatcherTest.CancelFunctionality` failed **10/10** on `small` against 10/10 passing at the default size |
| `5_testunit_shared_modules_utils.yml::shared_modules_utils-asan` | **downsized, then reverted** — same test, 1 failure in 23 runs on `small`, 0 in 26 at the default size |
| `5_builderprecompiled_docker-images-upload-manager.yml::Upload-package-building-images` | `docker build`/push of a builder image; did not register in the 28-day cost sample (~$0/month) |
| `6_builderprecompiled_server-images.yml::Upload-package-building-images` | `workflow_dispatch`-only; ~$0/month |

The last two are the only ones excluded on value rather than on evidence. Both are legitimate `small` candidates worth ~$0/month, and verifying them means actually publishing a builder image to GHCR — not a side effect worth triggering for no saving.

### Results and Evidence

**Measured, not assumed.** The two heaviest jobs in the set were dispatched on two throwaway branches carrying identical temporary probes — one with the `-small` label, one on the project default — so this is a same-day A/B on the same workload rather than a comparison against a historical mean. All jobs passed on both sizes.

The override provably takes effect:

| Probe | `-small` label | project default |
|---|---|---|
| `nproc` | **2** | 4 |
| `MemTotal` | **3.60 GiB** | 7.43 GiB |
| container limit (cgroup) | **3.00 GiB** | 6.75 GiB |

Duration and cost per run:

| Job | default (4 vCPU) | `small` (2 vCPU) | Δ duration | Cost/run before → after |
|---|---|---|---|---|
| `shared_modules_utils-modules` | 10.4 min | 10.4 min | **none** | $0.104 → $0.052 (−50%) |
| `shared_modules_utils-asan` | 19.0 min | 22.5 min | **+18%** | $0.190 → $0.113 (−41%) |
| `keystore-tests (asan)` | 4.8 min | 5.6 min | +17% | $0.048 → $0.028 (−42%) |
| `keystore-tests (valgrind)` | 4.8 min* | 5.0 min | +4% | $0.048 → $0.025 (−48%) |
| `check_standalone_engine (x64)` | 0.5 min | 0.6 min | none meaningful | $0.005 → $0.003 (−40%) |
| `check_standalone_engine (arm64)` | 0.3 min | 0.3 min | **none** | $0.002 → $0.001 (−50%) |

\* the valgrind control job actually recorded 14.8 min, but 10.7 of those were a flaky `Keystore - Run tests` step; its sibling asan job on the same instance took 4.8 min, so 4.8 is the fair baseline.

Both probe branches were new, so neither could read the Actions dependency cache (caches are branch-scoped and fall back only to the base branch). That inflates the absolute durations of the `Build external dependencies` / `Project dependencies` steps equally on both sides — it does not affect the comparison, but it is why `keystore-tests` shows 4.8 min here against a 2.5 min steady-state mean.

Peak memory sampled during each run: keystore 1.25–1.36 GiB, `check_standalone_engine` 0.79 GiB (x86) and 0.90 GiB (ARM), and `shared_modules_utils-asan` **1.99 GiB on small against the 3.00 GiB container limit** (66%), 2.17 GiB when the same job ran on the default size. No OOM on any job. The engine test was the one whose footprint was genuinely unknown before this — it uses about a quarter of what `small` provides.

So the `-j2`/single-threaded reasoning holds for `shared_modules_utils-modules` (identical duration, half price) but only partly for the ASAN job: its `cmake --build` step still slowed 13.3 → 15.5 min on half the cores. The trade there is 3.5 min of extra wall-clock for 41% less cost — worth taking in my view, since these jobs are not the critical path on a PR, but a reviewer who weighs feedback latency higher should say so and I will drop that one job back to the default.

#### Run evidence

Every measurement above, linked to the job it came from. `small` and reference were dispatched minutes apart on the same commit and, for the IT rows, against the identical prebuilt package — so the instance size is the only variable.

| Job | `small` run | duration | reference run | duration | Δ |
|---|---|---|---|---|---|
| `keystore-tests (asan)` | [pass](https://github.com/wazuh/wazuh/actions/runs/31090415449/job/92579812498) | 5.6 min | [pass](https://github.com/wazuh/wazuh/actions/runs/31090417721/job/92579814083) | 4.8 min | +17% |
| `keystore-tests (valgrind)` | [pass](https://github.com/wazuh/wazuh/actions/runs/31090415449/job/92579812441) | 5.0 min | [pass](https://github.com/wazuh/wazuh/actions/runs/31090417721/job/92579814020) | 14.8 min † | — |
| `check_standalone_engine (x64)` | [pass](https://github.com/wazuh/wazuh/actions/runs/31101892725/job/92621735859) | 0.6 min | [pass](https://github.com/wazuh/wazuh/actions/runs/31044531438/job/92441882171) ‡ | 0.5 min | none meaningful |
| `check_standalone_engine (arm64)` | [pass](https://github.com/wazuh/wazuh/actions/runs/31101892725/job/92621735935) | 0.3 min | [pass](https://github.com/wazuh/wazuh/actions/runs/31044531438/job/92441882160) ‡ | 0.3 min | none |

† This reference job spent 10.7 of its 14.8 min in a flaky `Keystore - Run tests` step; its sibling asan job on the same instance took 4.8 min, which is the fair baseline. Not an instance-size effect in either direction.

‡ No same-day control was dispatched for the engine test — its reference is the most recent successful run at the default size on another branch. Its `build` dependency did run at the default size in the `small` dispatch, so only the test job differed.

Rejected candidate, for the record — the integration tests (full reasoning in the comment below):

| Job | `small` run | duration | reference run | duration |
|---|---|---|---|---|
| `IT Linux - fim (tier-0)` | [**fail**](https://github.com/wazuh/wazuh/actions/runs/31095575470/job/92596668901) | 57.7 min | [pass](https://github.com/wazuh/wazuh/actions/runs/31095588383/job/92596716565) | 52.7 min |
| `IT Linux - agentd (tier-0-1)` | [pass](https://github.com/wazuh/wazuh/actions/runs/31095582167/job/92596691154) | 32.2 min | [pass](https://github.com/wazuh/wazuh/actions/runs/31095594737/job/92596734990) | 30.6 min |

Estimated saving for the jobs in this PR, over the same window as the [cost report](https://github.com/wazuh/wazuh-automation/issues/3637#issuecomment-5133556830) (2026-07-01 → 2026-07-28) and priced at the same rates:

| | Minutes / month | Rate | Cost / month |
|---|---|---|---|
| Before | ~660 | $0.0100 / $0.0070 (ARM) | ~$6.6 |
| After | ~660 | $0.0050 / $0.0035 (ARM) | ~$3.3 |

Those minutes are the total across every branch that runs these workflow files; this PR captures the share that runs on `5.0.0`-based branches.

This PR's own CI re-confirms it end to end: `verify-changelog`, `bandit-scan`, `keystore-tests` and both `shared_modules_utils-*` jobs all run on the overridden labels, so a mis-formed label would show up here as a job that never gets a runner. (They currently report `skipping` because the PR is a draft.)

The full per-job attribution of `server-amd` / `server-arm`, the reasoning for leaving the compile jobs on medium, and the higher-value levers found along the way (compiler caching, ARM price/performance, free public-repo runners) are in the audit comment below.

  - [ ] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

### Artifacts Affected

None. CI configuration only: six GitHub Actions workflow definitions under `.github/workflows/` — four with `runs-on` label changes, two with explanatory comments alone. No product source, packaging or default configuration is touched.

### Configuration Changes

CI infrastructure only: the `runs-on` label of the four jobs above now carries a `-small` instance-size suffix. `check_standalone_engine` is the one that touches `server-arm` as well as `server-amd`, since its label is an arch ternary. Nothing changes in the CodeBuild project configuration itself.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

One thing for the reviewer to weigh: the usable memory on `small` is 3.0 GiB, not the nominal 4 GB, and the ASAN build of `shared_modules_utils` peaks at 1.99 GiB of it. That is the thinnest margin in the set — if it ever OOMs, the fix is to drop that single job back to the default rather than revert the set.

### DIFF
--- a/.github/workflows/5_builderpackage_engine-standalone.yml
+++ b/.github/workflows/5_builderpackage_engine-standalone.yml
@@ -132,7 +132,11 @@ jobs:
   check_standalone_engine:
     name: Test standalone engine (${{ matrix.arch }})
 
-    runs-on: ${{ matrix.arch == 'arm64' && format('codebuild-github-actions-codebuild-runner-server-arm-{0}-{1}', github.run_id, github.run_attempt) || format('codebuild-github-actions-codebuild-runner-server-amd-{0}-{1}', github.run_id, github.run_attempt) }}
+    # `small` instance (2 vCPU, 3.0 GiB usable): downloads and starts the prebuilt engine,
+    # no compilation. Measured 5/5 passes per architecture, peak 0.9 GiB, duration unchanged
+    # at 0.8 min (x64) / 0.3 min (arm64). The `-small` suffix keeps this job at one label —
+    # read the note in 5_testcomponent_keystore.yml before switching to separate labels.
+    runs-on: ${{ matrix.arch == 'arm64' && format('codebuild-github-actions-codebuild-runner-server-arm-{0}-{1}-small', github.run_id, github.run_attempt) || format('codebuild-github-actions-codebuild-runner-server-amd-{0}-{1}-small', github.run_id, github.run_attempt) }}
     permissions:
       id-token: write
       contents: read
--- a/.github/workflows/5_codeanalysis_python_bandit_pr.yml
+++ b/.github/workflows/5_codeanalysis_python_bandit_pr.yml
@@ -20,7 +20,10 @@ env:
 jobs:
   bandit-scan:
     if: ${{ github.event_name != 'pull_request' || !github.event.pull_request.draft }}
-    runs-on: codebuild-github-actions-codebuild-runner-server-amd-${{ github.run_id }}-${{ github.run_attempt }}
+    # `small` instance (2 vCPU, 3.0 GiB usable): single-process Python static analysis.
+    # The `-small` suffix keeps this job at one label — read the note in
+    # 5_testcomponent_keystore.yml before switching to a separate `instance-size:` label.
+    runs-on: codebuild-github-actions-codebuild-runner-server-amd-${{ github.run_id }}-${{ github.run_attempt }}-small
     timeout-minutes: 20
     steps:
       - name: Checkout repo
--- a/.github/workflows/5_codequality_changelog.yml
+++ b/.github/workflows/5_codequality_changelog.yml
@@ -12,7 +12,10 @@ jobs:
   # Enforces the update of a changelog file on every pull request
   verify-changelog:
     if: ${{ !github.event.pull_request.draft && github.repository == 'wazuh/wazuh' }}
-    runs-on: codebuild-github-actions-codebuild-runner-server-amd-${{ github.run_id }}-${{ github.run_attempt }}
+    # `small` instance (2 vCPU, 3.0 GiB usable): this job only diffs the changelog file.
+    # The `-small` suffix keeps this job at one label — read the note in
+    # 5_testcomponent_keystore.yml before switching to a separate `instance-size:` label.
+    runs-on: codebuild-github-actions-codebuild-runner-server-amd-${{ github.run_id }}-${{ github.run_attempt }}-small
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
--- a/.github/workflows/5_testcomponent_keystore.yml
+++ b/.github/workflows/5_testcomponent_keystore.yml
@@ -20,7 +20,16 @@ concurrency:
 jobs:
   keystore-tests:
     if: ${{ github.event_name != 'pull_request' || !github.event.pull_request.draft }}
-    runs-on: ${{ format('codebuild-github-actions-codebuild-runner-server-amd-{0}-{1}', github.run_id, github.run_attempt) }}
+    # `small` instance (2 vCPU, 3.0 GiB usable): build_target_cpp runs `cmake --build`
+    # without --parallel and build_external_deps caps its make at -j2, so this job never
+    # uses more than two cores. Measured 10/10 passes per matrix leg, peak 1.4 GiB.
+    #
+    # The `-small` suffix form is deliberate: a separate `instance-size:small` label would
+    # give this job a different label count than its siblings, and per the AWS docs a job
+    # with fewer labels can be picked up by a runner created for a job with more, leaving
+    # that job without a runner. Keep the size inside the single label.
+    # https://docs.aws.amazon.com/codebuild/latest/userguide/sample-github-action-runners-update-labels.html
+    runs-on: ${{ format('codebuild-github-actions-codebuild-runner-server-amd-{0}-{1}-small', github.run_id, github.run_attempt) }}
     timeout-minutes: 60
     strategy:
       fail-fast: true
--- a/.github/workflows/5_testintegration_linux-integration-tests.yml
+++ b/.github/workflows/5_testintegration_linux-integration-tests.yml
@@ -97,6 +97,11 @@ jobs:
       BUILD_WORKFLOW: 5_builderpackage_agent-linux.yml
       PACKAGES_PATH: /tmp/packages
       AGENT_PACKAGE_REGEXP: wazuh-agent.*amd64.*\.deb
+    # NOTE: deliberately NOT downsized to a `small` instance. These suites are pytest with no
+    # compilation, so they look like obvious candidates, but a measured A/B found that fim
+    # tier-0 fails test_fill_capacity's real-time event-count assertion on 2 vCPU while
+    # passing at this size on the identical package (agentd tier-0-1 passed on both).
+    # Data: https://github.com/wazuh/wazuh/pull/38215#issuecomment-5204406595
     runs-on: codebuild-github-actions-codebuild-runner-server-amd-${{ github.run_id }}-${{ github.run_attempt }}
     steps:
       - name: Checkout Repo
--- a/.github/workflows/5_testunit_shared_modules_utils.yml
+++ b/.github/workflows/5_testunit_shared_modules_utils.yml
@@ -23,6 +23,14 @@ permissions:
 jobs:
   shared_modules_utils-modules:
     if: ${{ github.event_name != 'pull_request' || !github.event.pull_request.draft }}
+    # NOTE: deliberately NOT downsized to a `small` instance, and neither is -asan below.
+    # utils_utest contains AsyncValueDispatcherTest.CancelFunctionality, which waits for its
+    # counter to reach 5, cancels, then asserts 10 — items 6..10 are unsynchronised, so the
+    # outcome depends on when the worker thread is scheduled. Measured on 2 vCPU: this job
+    # failed 10/10 and -asan failed 1/23, against zero failures on that test at this size.
+    # The race is in the test, not in valgrind; running under valgrind only makes it
+    # near-deterministic. Revisit once the test synchronises on all 10 items.
+    # Data: https://github.com/wazuh/wazuh/pull/38215#issuecomment-5205906176
     runs-on: codebuild-github-actions-codebuild-runner-server-amd-${{ github.run_id }}-${{ github.run_attempt }}
     timeout-minutes: 60
     steps:
@@ -68,6 +76,7 @@ jobs:
 
   shared_modules_utils-asan:
     if: ${{ github.event_name != 'pull_request' || !github.event.pull_request.draft }}
+    # NOTE: not downsized — see the AsyncValueDispatcherTest note on the job above.
     runs-on: codebuild-github-actions-codebuild-runner-server-amd-${{ github.run_id }}-${{ github.run_attempt }}
     timeout-minutes: 60
     steps:


jepalfer:
## Server runner audit: per-job attribution of `server-amd` / `server-arm`

The CloudWatch work in wazuh/wazuh-automation#3637 told us *which projects* cost money; it could not tell us *which jobs inside them* do. This comment fills that gap for the two Server projects, and reaches a conclusion that changes the shape of the task: **`server-amd` is not over-provisioned — it is over-used.** Almost all of its spend is `-j$(nproc)` C/C++ compilation, so a project-level downsize would trade money for wall-clock time rather than save it. The genuine rightsizing headroom is a small, well-identified set of jobs, and it is already implemented (see below).

### Method

Per-job durations pulled from the GitHub Actions REST API (`/actions/runs/{id}/jobs`) over the same window as the cost report above (2026-07-01 → 2026-07-28), classifying every job by its `runs-on` labels. Stratified sample: up to 10 completed runs per workflow, scaled by that workflow's total run count in the window. Minutes priced at the same on-demand rates used in [the cost report](https://github.com/wazuh/wazuh-automation/issues/3637#issuecomment-5133556830) (`Linux general1.medium` $0.0100/min, `ARM general1.medium` $0.0070/min).

Sanity check: this reconstruction accounts for **$2,656 of the reported $3,927** on `server-amd` (68%) and **$213 of $207** on `server-arm`. The shortfall on `server-amd` is expected from the sampling (re-run attempts are excluded, and workflows with >100 runs in the window are sampled from their most recent 100), so read the **shares** as solid and the absolute dollars as a floor.

### Where the `server-amd` money goes

Top workflows, estimated over the 28-day window:

| Workflow | Est. $ | Share | Dominant work |
|---|---|---|---|
| `5_testunit_linux-win.yml` | 444 | 16.7% | `make deps -j$(nproc)` + `make TEST=1 -j$(nproc/2)` |
| `5_testcomponent_syscollector-rtr.yml` | 292 | 11.0% | `build.py --readytoreview` (parallel, `getCpuCores()`) |
| `4_testcomponent_syscollector-rtr.yml` | 260 | 9.8% | same, 4.x branches |
| `5_builderpackage_manager-multiples.yml` | 200 | 7.5% | Docker package build |
| `5_builderpackage_agent-linux.yml` | 163 | 6.1% | mostly `IT Linux - *` integration tests |
| `5_codeanalysis_scanbuild.yml` | 155 | 5.8% | `scan-build-14 ... -j$(nproc)` |
| `4_builderpackage_manager-multiples.yml` | 106 | 4.0% | Docker package build |
| other `*-rtr.yml` component tests (5.x) | ~418 | 15.7% | parallel compile + tests |
| remaining 4.x integration/unit tests | ~348 | 13.1% | parallel compile |

Two structural facts worth stating plainly:

1. **Branch split: 5.x workflows are 67% of `server-amd` spend, 4.x workflows 33%.** Any workflow-level change has to be applied per branch series to capture the whole saving.
2. **Single biggest line item is one job:** `Unit-Tests (manager)` in `5_testunit_linux-win.yml`, 89 min mean (n=36), ~$261/month. Its step breakdown is 74 of 77 min in one step, "Build wazuh" — and that step compiles at `-j$(nproc)/2`, i.e. **it only uses 2 of the 4 available cores** ([`legacy_unit_tests.sh:21`](https://github.com/wazuh/wazuh/blob/main/.github/actions/legacy_unit_tests.sh#L21), deliberately, to avoid memory pressure).

### Why a project-level downsize is the wrong move here

For a job that scales with cores, cost is rate × time, and halving the rate roughly doubles the time. With Amdahl's law and a serial fraction *s*, moving medium → small changes cost by `0.5 × (s + (1-s)/2) / (s + (1-s)/4)`: 1.00 at *s*=0, 0.85 at *s*=0.1, 0.75 at *s*=0.2. So a blanket downsize of the compile jobs would save 0–25% of their minutes' cost while making every PR 40–70% slower to give feedback — and it would breach existing `timeout-minutes` (`scan-build` is capped at 30 min against a 12.8 min mean; `5_testunit_linux-win` at 130 min against an 89 min mean, 104 min worst case).

That is a latency-for-money trade, not a rightsizing win, so I have not applied it. Recommend rejecting it.

### What is actually rightsizable

Ten jobs on `server-amd` provably cannot use more than 2 vCPU, so `instance-size:small` (2 vCPU / 4 GB, $0.0050/min) halves their rate with no expected slowdown. Five of them still run on CodeBuild on `5.0.0` and are what this PR changes; the other five were already moved off CodeBuild entirely on this branch by #38188, so they only apply to `main`:

| Workflow / job | Evidence it cannot use 4 vCPU | Mean duration | Est. $/mo | In this PR |
|---|---|---|---|---|
| `5_testunit_shared_modules_utils.yml` (2 jobs) | `build_target_cpp` → `cmake --build` with no `--parallel` → single-threaded | 18.3 / 8.2 min | 17 | yes |
| `5_testcomponent_keystore.yml` `keystore-tests` | same | 2.5 min | 3 | yes |
| `5_codequality_changelog.yml` | changelog diff only | <1 min | <1 | yes |
| `5_codeanalysis_python_bandit_pr.yml` | single-process bandit | <1 min | <1 | yes |
| `5_testcomponent_windows-installer-upgrade.yml` `build` | `make ... -j2` (fixed) | 7.8 min | 13 | `main` only |
| `5_testcomponent_windows-dll-hijack.yml` `build` | `make ... -j2` (fixed) | 7.3 min | 13 | `main` only |
| `5_testcomponent_windows-files.yml` `build` | `make ... -j2` (fixed) | 7.4 min | 12 | `main` only |
| `5_codelinter_clangformat.yml` `style` | clang-format only, no compilation | 0.3 min | <1 | `main` only |
| `5_testunit_wodles.yml` `build` | single-process pytest | 1.2 min | <1 | `main` only |

Applied via the single-label override form:

```yaml
runs-on: codebuild-github-actions-codebuild-runner-server-amd-${{ github.run_id }}-${{ github.run_attempt }}-small
```

The single-label form is deliberate: adding `instance-size:` as a *separate* label changes a job's label count, and [per the AWS docs](https://docs.aws.amazon.com/codebuild/latest/userguide/sample-github-action-runners-update-labels.html) a job with fewer labels can then be picked up by a runner created for a different job in the same run. No project or fleet configuration change is needed, and no AWS-side change is required to merge this.

**Saving: ~$10/month for the five jobs in this PR** (~$21/month of minutes at half rate), and ~$33/month more if `main` gets the same treatment for the other five. Small in absolute terms — that is the honest result of the rightsizing question, not an artifact of a conservative selection. Note also that for these particular jobs, migrating to GitHub-hosted runners as #38188 did would save 100% of the minutes rather than 50%; that is a concurrency decision rather than a sizing one, hence step 2 below.

### ARM evaluation: measured, and favourable

Production data for the *same* job on the two projects, same window, `5_builderpackage_manager-multiples.yml`:

| Job | Project | Mean duration | Rate | Cost/build |
|---|---|---|---|---|
| Build rpm wazuh-manager x86_64 | `server-amd` | 21.3 min | $0.0100 | $0.213 |
| Build rpm wazuh-manager aarch64 | `server-arm` | 19.5 min | $0.0070 | **$0.137** (−36%) |
| Build deb wazuh-manager amd64 | `server-amd` | 21.2 min | $0.0100 | $0.212 |
| Build deb wazuh-manager arm64 | `server-arm` | 16.1 min | $0.0070 | **$0.113** (−47%) |

ARM is not just 30% cheaper per minute, it is also **faster** on this workload. That supports the parent task's ~20% price-performance assumption and then some. Candidates to migrate are jobs whose output is not architecture-specific (static analysis, Python test suites, docker image publishing). Explicitly *not* candidates: anything cross-compiling winagent with mingw or running tests under wine, and the x86 package smoke tests.

### Also found: CodeBuild medium and the free public-repo runners are performance-equivalent

`wazuh/wazuh` is a **public** repository, so GitHub-hosted standard runners bill nothing. Several workflows still run on `ubuntu-24.04`, which gives a direct A/B on identical jobs (`5_testunit_linux-win.yml`, last 40 successful runs):

| Job | CodeBuild medium | `ubuntu-24.04` |
|---|---|---|
| Unit-Tests (manager) | 89.4 min (n=36) | 85.5 min (n=4) |
| Unit-Tests (winagent) | 46.3 min (n=36) | 52.1 min (n=4) |
| Unit-Tests (agent) | 22.1 min (n=36) | 18.5 min (n=4) |

Same duration, $0.01/min vs $0. Every minute moved back to a standard runner is a 100% saving on that minute, versus ~25% for the best sizing change available. The binding constraint is the org's concurrent-job limit (the reason for the original migration), not cost — so the question worth answering next is *how much concurrency headroom the `team` plan actually leaves*, and how much of the 6,800 monthly `server-amd` hours could overflow-only to CodeBuild. That is a larger change than this task, but it dominates every other lever and should not be discovered after the sizing work is finished.

### Recommended next steps, in order of value

1. **Stop recompiling the same tree.** `5_testcomponent_syscollector-rtr.yml` runs 7 jobs per PR, each a full from-scratch build (~24 min each, ~$292/mo on 5.x plus $260/mo on 4.x). Compiler caching (the `sccache`/S3 work in wazuh/wazuh-automation#3637) and sharing one build across the matrix are worth an order of magnitude more than any sizing change.
2. **Quantify the free-runner headroom** (above) — #38188 already did this for the agent-side workflows on `5.0.0` with measured before/after durations; the same question is open for the server-side jobs.
3. **`Unit-Tests (manager)`:** it compiles at 2 of 4 cores by design. Benchmark medium vs large — on 8 vCPU / 16 GB the `nproc/2` heuristic yields `-j4` with more memory headroom, which should be close to cost-neutral while cutting a 90-minute PR-blocking job roughly in half.
4. ~~**`IT Linux - *` integration tests (~$155/mo, `5_builderpackage_agent-linux.yml`):** pytest suites dominated by real-time waits, so `small` should be nearly free money (~$75/mo).~~ **Measured and rejected** — see [the A/B below](https://github.com/wazuh/wazuh/pull/38215#issuecomment-5204406595): `fim` tier-0 fails a real-time capacity assertion on `small` while passing at the current size on the identical package. `agentd` passed, so a per-shard split remains possible, but it needs an `instance_size` workflow input and one validated A/B per shard.
5. **Migrate architecture-agnostic Linux jobs to `server-arm`,** starting with `scan-build-linux` (~$92/mo → ~$60/mo).

### Not done — needs AWS access

I have no credentials for the CI account, so these remain open and are best done by @teddytpc1 or whoever holds them:

- CloudWatch CPU/memory verification of the downsized jobs after merge (the one risk in this patch is memory: `small` is 4 GB, and the ASAN build of `shared_modules_utils` is the closest to that ceiling).
- CloudWatch alarms for CPU throttling / RAM exhaustion on the downsized jobs.
- Confirming realised savings against Cost Explorer once merged.

jepalfer:
## Measured: the Linux integration tests should **not** be downsized

The audit comment above listed the `IT Linux - *` jobs as the largest remaining rightsizing candidate on `server-amd` (~$155/month of minutes, ~7× everything in this PR) and flagged that they needed an A/B before shipping, because the fim/agentd shards are timing-sensitive. That A/B has now run, and the answer is no.

### Setup

Two shards, dispatched at both sizes, all four runs pulling the **same prebuilt agent DEB** (producer run `31091886544`) so the package under test is byte-identical and the instance size is the only variable:

| Shard | Size | Result | Duration | Run |
|---|---|---|---|---|
| `fim` tier-0 | `small` (2 vCPU) | **failure** — 1 failed, 73 passed | 57.7 min | [job](https://github.com/wazuh/wazuh/actions/runs/31095575470/job/92596668901) |
| `fim` tier-0 | current (4 vCPU) | success | 52.7 min | [job](https://github.com/wazuh/wazuh/actions/runs/31095588383/job/92596716565) |
| `agentd` tier-0-1 | `small` (2 vCPU) | success | 32.2 min | [job](https://github.com/wazuh/wazuh/actions/runs/31095582167/job/92596691154) |
| `agentd` tier-0-1 | current (4 vCPU) | success | 30.6 min | [job](https://github.com/wazuh/wazuh/actions/runs/31095594737/job/92596734990) |

### The failure

```
_________ test_fill_capacity[Default file limit fill 100% - Real-time] _________
    assert int(log_monitor.callback_result[0]) == files_amount
```

A real-time FIM capacity test: it fills the file limit and asserts the number of entries FIM reported equals the number of files created. Halving the vCPUs is a plausible direct cause — real-time (inotify) event handling under a full-capacity burst is exactly the workload where a slower consumer drops or lags events and the count comes up short. This is not a timeout being missed by a few seconds; it is the assertion about how many events got through.

### Honest reading of the evidence

- **For attribution:** the control run passed 74/74 on the same package, same code, same morning. Size was the only difference, and the failing test's mechanism is resource-sensitive by nature.
- **Against attribution:** n=1 per size, and `IT Linux - fim (tier-0)` has failed in 4 of its last 5 occurrences at the *current* size on feature branches (those branches carried agent changes that could legitimately break fim, so that base rate is not clean either). I did not repeat the run.

So I am treating this as sufficient reason **not** to downsize, rather than as proof that `small` breaks fim. Trading a $75/month saving against any increase in flakiness on the shard that already generates the most CI noise is the wrong side of that bet, and re-running until it looks green would be the wrong way to resolve it.

The IT workflow keeps the project default in this PR.

### What this leaves on the table, for whoever picks it up

`agentd` passed at `small` with a +5% duration increase, so this is a per-shard property, not a property of the IT workflow. The shards that never touch real-time filesystem events — `github`, `office365`, `msgraph`, `enrollment`, `logcollector` — are the plausible candidates, and together they are a meaningful slice of the ~$155/month.

Capturing that would need `5_testintegration_linux-integration-tests.yml` to take the instance size as a `workflow_call` input (defaulting to the project default) so `5_builderpackage_agent-linux.yml` can set it per matrix entry, plus one validated A/B per shard before each is switched. That is a separate piece of work with its own risk budget, not something to fold into a five-job sizing PR.

Worth noting for whoever does: these IT jobs run on the **Server** runner project while living in the agent package workflow, which is why they surfaced in this Server-scoped audit at all.

jepalfer:
## `shared_modules_utils-modules` reverted to the default size

This PR's own CI caught something the pre-merge A/B did not: [`shared_modules_utils-modules` failed on `small`](https://github.com/wazuh/wazuh/actions/runs/31104243042/job/92625179835) while every other check passed. The job is now back on the project default. This comment records why, because the reasoning matters more than the one-line change.

### The failure

```
AsyncValueDispatcherTest.CancelFunctionality
src/shared_modules/utils/tests/asyncValueDispatcher_test.cpp:205: Failure
  10
  counter
    Which is: 5
```

The test, reduced to its shape:

```cpp
// dispatcher constructed with 1 worker thread
for (int i = 0; i < 10; ++i) dispatcher.push(i);
promise.get_future().wait();     // the promise fires when counter == 5
EXPECT_FALSE(dispatcher.cancelled());
dispatcher.cancel();
EXPECT_TRUE(dispatcher.cancelled());
for (int i = 10; i < 20; ++i) dispatcher.push(i);   // rejected after cancel
EXPECT_EQ(10, counter);
```

It synchronises on the counter reaching **5**, cancels, and then asserts **10**. Nothing waits for items 6–10; the test is betting that the single worker thread drains them before `cancel()` takes effect. `counter == 5` is precisely the outcome where `cancel()` won that race.

### Why this job and not its `-asan` sibling

Both jobs build and run the same `utils_utest` binary, so both contain the same test. The difference is in how they run it:

| Job | Runs `utils_utest` | Result on `small` |
|---|---|---|
| `shared_modules_utils-modules` | **under valgrind** (`test_cpp` with `valgrind: true`) | 1 pass, 1 fail |
| `shared_modules_utils-asan` | ASAN-instrumented, no valgrind | 2 pass, 0 fail |

> [!WARNING]
> **Superseded — the conclusion in this section was wrong.** Repeated runs (see the following comment) hit the *same* test in the `-asan` job, so the race is a property of `utils_utest`, not of valgrind. Valgrind amplifies it (10/10 vs 1/23) but does not create it. `-asan` has since been reverted to the default size too.

~~Valgrind serialises threads onto a single virtual CPU and adds heavy interpretation overhead, so the worker thread gets far less real progress per unit of wall-clock than it does under ASAN. Halving the host vCPUs squeezes that further. The failure concentrating in the valgrind-wrapped job is consistent with the mechanism rather than coincidental, which is what makes it reasonable to keep the `-asan` job on `small` instead of reverting both.~~

### How much this is "caused" by the change

| Instance size | Runs observed | Failures |
|---|---|---|
| default (4 vCPU) | 32 | 0 |
| `small` (2 vCPU) | 2 | 1 |

The race is pre-existing and not introduced here — it can fire at any size on a sufficiently loaded machine. But 0/32 at the default size against 1/2 on `small`, plus a mechanism that depends directly on thread scheduling, means downsizing this job measurably raises the failure rate. Two samples on `small` is thin evidence in isolation; combined with the mechanism it is enough to act on.

Worth stating plainly: the single clean A/B run I did before opening this PR **passed** on `small`. A non-deterministic failure is exactly what one run per configuration cannot rule out, and I noted that limitation when presenting those results. This is that limitation biting.

### Why revert rather than fix the test

The test has a real defect — asserting a value that nothing synchronises on — and fixing it would be worth doing on its own merits, since it can flake on any loaded runner regardless of instance size. Two reasons not to do it here:

1. It is product test code owned by whoever maintains `shared_modules/utils`, and correcting a concurrency test's synchronisation deserves that owner's review rather than being folded into a CI cost-optimisation PR.
2. Even if fixed, the honest saving from downsizing this job is about **$5/month**. Accepting any increase in CI flakiness for that is a bad trade, and the fix would have to be proven over many runs before it could be trusted, which is not work this PR should be gating on.

So the job keeps the default size, with an inline comment in the workflow explaining why it is deliberately excluded while its sibling is not — so nobody "completes" the pattern later without knowing this happened.

### Effect on the PR

Five jobs downsized instead of six. Estimated saving drops from ~$10.9/month to **~$8.3/month**.

### Suggested follow-up, not filed

`AsyncValueDispatcherTest.CancelFunctionality` should either wait for all 10 items before asserting, or assert a range rather than an exact count. I have not opened an issue for this — it belongs to the utils owner and should be routed through the normal template. The reproduction is: run `utils_utest` under valgrind on a 2-vCPU machine.

jepalfer:
## Repeated-execution results, and a correction

The evidence behind this PR was 1–2 runs per job, which cannot distinguish "works" from "usually works". I dispatched 35 workflow runs (~90 job executions) to replace that with counts. Two conclusions changed as a result, and one explanation I gave in the comment above turns out to be wrong.

### Success/total per job

| Job | Size | Passed/Total | Mean duration |
|---|---|---|---|
| `keystore-tests (asan)` | small | **10/10** | 4.6 min |
| `keystore-tests (valgrind)` | small | **10/10** | 4.8 min |
| `check_standalone_engine (x64)` | small | **5/5** | 0.8 min |
| `check_standalone_engine (arm64)` | small | **5/5** | 0.3 min |
| `shared_modules_utils-asan` | small | 19/20 | 20.9 min |
| `shared_modules_utils-modules` | small | **0/10** | 10.0 min |
| `shared_modules_utils-modules` | default (control) | **10/10** | 10.1 min |

`bandit-scan` and `verify-changelog` are `pull_request`-only, so they cannot be dispatched; they accumulate a sample per push to this PR and have passed on every one so far. Both are sub-90-second single-process jobs with no compilation, so repetition tells us least about them.

### `shared_modules_utils-modules`: not flaky on small — deterministic

**0 of 10 passed on `small`; 10 of 10 passed at the default size, same commit, same afternoon.** Every failure is the same assertion:

```
asyncValueDispatcher_test.cpp:205: Expected equality of these values:
  10
  counter
    Which is: 5
```

I reverted this job on the strength of 1 failure in 2 runs and called it a race that "loses more often" on 2 vCPU. The revert was right, but the characterisation was too soft: on this instance size the test essentially always loses. The single early probe run that passed is the outlier, at 1 pass in 12 observed `small` runs.

### Correction: my valgrind explanation was wrong

In the comment above I argued the failure was confined to the valgrind-wrapped job because valgrind serialises threads, and used that to justify keeping `-asan` on `small`. **That is falsified.** The one `-asan` failure in this batch is the *same test at the same line*:

```
shared_modules_utils-asan → asyncValueDispatcher_test.cpp:205 → AsyncValueDispatcherTest.CancelFunctionality
```

So the race is a property of `utils_utest`, not of valgrind. Valgrind amplifies it dramatically — 10/10 versus 1/23 — but does not create it. The reasoning I used to keep `-asan` downsized did not hold, and `-asan` is now reverted to the project default as well.

For that specific test, measured across both sizes:

| Job | Size | `CancelFunctionality` failures |
|---|---|---|
| `-modules` | small | 10 / 10 runs |
| `-modules` | default | 0 / 10 runs (this batch) + 0 / 32 (historical) |
| `-asan` | small | 1 / 23 runs |
| `-asan` | default | 0 / 26 runs (historical) |

One event at 1/23 is thin on its own. It is not thin when the same test fails 10/10 in a sibling job on the same instance size: the mechanism is established, and `-asan` is the same test with less amplification. Downsizing it trades a ~4% flake rate on a PR-blocking job for about $6/month, which is not a trade worth making.

Note in passing: the `-asan` job's single *default-size* failure in the last 60 runs was a different test (`SocketTest.SingleDelayedClientWithReconnectionServerReset`, a socket timing test). `utils_utest` has more than one timing-sensitive test in it — further reason to fix them before revisiting instance sizes for these jobs.

### What this does to the PR

Four jobs downsized instead of six:

| Job | Status |
|---|---|
| `keystore-tests` (both legs) | downsized — 20/20 |
| `check_standalone_engine` (both arches) | downsized — 10/10 |
| `bandit-scan`, `verify-changelog` | downsized — green on every PR run |
| `shared_modules_utils-modules`, `-asan` | reverted to default |

**Estimated saving drops to ~$3.30/month**, from the ~$10.90 claimed when this PR was opened. The two `shared_modules_utils` jobs were ~70% of the value, and they are exactly the ones measurement disqualified.

That figure is small enough to be worth stating plainly: as a cost measure this PR is close to noise against a $3,927/month project. Its value is in what the audit found — that `server-amd` is compute-bound rather than over-provisioned, that ARM is both cheaper and faster for the package builds, and that the free public-repo runners are performance-equivalent — plus a latent test race that would otherwise have been discovered by someone else's red PR.

### Cost of this exercise

35 dispatches, roughly $6 of CodeBuild time, to avoid shipping two jobs that would have flaked in other people's PRs. The `-asan` failure would not have shown up in the 1–2 runs per configuration I had before.

### Follow-up, still not filed

`AsyncValueDispatcherTest.CancelFunctionality` should wait for all 10 items, or assert a range instead of an exact count. Reproduction is now trivial: run `utils_utest` under valgrind on a 2-vCPU machine — it fails every time. Once fixed, both `shared_modules_utils` jobs become straightforward `small` candidates worth ~$8.50/month between them, which is more than everything this PR currently saves.

jotacarma90:
This PR is not needed, runners migrated to GH:
https://github.com/wazuh/wazuh/pull/38233

PR #38225, opened 2026-08-06
Migrate the agent package-build workflows to GitHub-hosted runners

### LINKED ISSUE #38174: Rightsizing CodeBuild Compute Environments (Agent)

nbertoldo:
# Migrate the agent package-build workflows to GitHub-hosted runners

## Description

Related issue: #38174

The four agent package-build workflows are the last consumers of the `agent-amd`, `agent-arm` and `agent-windows-amd` CodeBuild fleets in this repository — 21 runner references across four files. This PR moves all of them to GitHub-hosted runners.

They were kept on CodeBuild under the criterion that package builds must not depend on GitHub-hosted runner availability, because a release cannot be blocked by a GitHub outage. That criterion does not apply to these copies: **none of the four has a `workflow_call` trigger and nothing invokes them as part of a release.** They fire on `pull_request` and `workflow_dispatch` only, so they serve PR verification and manual development builds. Release packages are built from `wazuh/wazuh-agent-packages`, and those workflows stay on CodeBuild — they are rightsized separately in wazuh/wazuh-agent-packages#912 and #913.

With #38149 closed and the decision taken to migrate all `wazuh/wazuh` runners to GitHub-hosted, tracked as a continuation of #38174, this supersedes #38189, which rightsized twelve of these same jobs to a smaller CodeBuild tier. That PR is closed; a job leaving CodeBuild does not need its CodeBuild tier tuned.

`wazuh/wazuh` is a public repository, so the hosted runners used here — including the ARM64 ones — are free.

### Runner migration — 21 references

| Workflow | Jobs | Destination |
|---|---|---|
| `5_builderpackage_agent-linux.yml` | `build-deb`, `test-deb`, `upload-deb-to-s3`, `build-rpm`, `test-rpm`, `upload-rpm-to-s3`, `detect-changes` | `ubuntu-24.04` |
| `5_builderpackage_agent-linux-arm64.yml` | all six `build-*`, `test-*` and `upload-*` jobs | `ubuntu-24.04-arm` |
| `5_builderpackage_agent-windows.yml` | `detect-changes`, `build-windows-i686`, `upload-windows-i686-to-s3` | `ubuntu-24.04` |
| `5_builderpackage_agent-windows.yml` | `package-windows-i686`, `test-windows-i686` | `windows-2022` |
| `5_builderpackage_agent-macos.yml` | `upload-macos-intel64-to-s3`, `upload-macos-arm64-to-s3`, `detect-changes` | `ubuntu-24.04` |

The macOS build and test legs already ran on `macos-14` / `macos-14-large` and are untouched; only the upload and change-detection helpers move.

Destinations match what these jobs used before the CodeBuild migration, with one deliberate departure: arm64 goes to GitHub's hosted ARM runners rather than the self-hosted `wz-linux-arm64` label those jobs came from, following the precedent set by #38188. The two Windows jobs go to `windows-2022`, which is both what they used before CodeBuild *and* the OS of the CodeBuild image (`aws/codebuild/windows-base:2022`) — they were first pointed at `windows-2025` for consistency with #38188 and moved back after CI showed why that mattered; see the evidence section.

### The eight "CodeBuild workarounds" steps are slimmed, not deleted

This is the part of the diff that is easy to misread. The step name suggests the whole thing is CodeBuild-specific; only its first of four parts is.

**Removed** — installing `docker.io` and starting `dockerd --storage-driver vfs`, which existed because CodeBuild has no systemd so the daemon never started. The hosted images ship Docker 28.0.4 running on `overlay2`.

**Kept** — `git config core.repositoryformatversion 0`. This is not runner-specific: `actions/checkout` on a newer git writes `repositoryformatversion=1`, which the older git inside the builder Docker images cannot read. It is needed on any runner.

**Kept** — the `uuidgen` and `python3-pip` guards. Both are `if`-guarded and no-op where the tools are present, so they cost nothing on hosted images while keeping the workflow runnable elsewhere.

### One of three `TODO(codebuild-temp)` steps in `agent-windows.yml` is removed

| Step | Decision | Reason |
|---|---|---|
| Set up WiX Toolset (MSI build) | **removed** | The published `windows-2022` image inventory lists WiX Toolset **3.14.1.8722** — the exact version the 20-line download block fetched (`wix3141rtm`), so it would never execute |
| Install a compatible CMake | **kept**, TODO corrected | The TODO was wrong. `reinstall_cmake` pins **CMake 3.31.6** for the winagent deps build (cmocka) and is shared by **27 workflows** — removing it would have silently changed the cmake version rather than cleaning up a workaround |
| Set up Visual C++ Build Tools | **kept** — required on any runner | Not a workaround despite its TODO. Its last line, `Add-Content -Path $env:GITHUB_PATH -Value $mspdbDir`, runs unconditionally and is the step's actual purpose: MSVC does not put the 32-bit `mspdb*.dll` on PATH by default, and `cv2pdb` needs it to convert DWARF debug info into PDBs. Only the *install* branch no-ops on `windows-2022` (`Using PDB writer from: …\VC\Tools\MSVC\14.29.30133\bin\HostX86\x86`, 2 s). Contrast the WiX step, whose PATH injection sat inside its `else`, so it genuinely did nothing when WiX was present |

### Unrelated cleanup

`.github/actions/setup_wix` is deleted. It has had **zero references** since #38188 removed its only consumer, so it is dead code independent of this migration.

Net across the three commits on this branch: 6 files, 83 insertions, 206 deletions. The third only corrects a comment.

### Runner capabilities verified before writing the change

A throwaway probe branch ran three legs and was then deleted ([run 31116229986](https://github.com/wazuh/wazuh/actions/runs/31116229986)):

| Runner | Arch | vCPU | MemTotal | Docker | Storage driver |
|---|---|---|---|---|---|
| `ubuntu-24.04` | x86_64 | 4 | 15.6 GiB | 28.0.4 | **overlay2** |
| `ubuntu-24.04-arm` | **aarch64** | **4** | **15.6 GiB** | 28.0.4 | **overlay2** |

`ubuntu-24.04-arm` had no precedent in this repository — #38188 migrated nothing ARM — so its availability was the main open risk. It picked the job up in four seconds and reports the same 4 vCPU the compiles saturate today on CodeBuild, with roughly twice the memory (15.6 GiB against 7.5).

The probe also covered `windows-2025` (WiX v3.14, CMake 4.4.2, Visual Studio Enterprise 2026, 7-Zip), but these jobs ended up on `windows-2022`, whose published image inventory lists WiX Toolset 3.14.1.8722, CMake 3.31.6, Visual Studio Enterprise 2022 17.14.37516.0 and 7-Zip 26.02 — WiX at the exact version the removed step downloaded.

### CodeBuild baseline for the duration comparison

Measured on CodeBuild during #38189, two runs of each workflow. This is what the CI run on this PR should be compared against:

| Job | CodeBuild (4 vCPU / 7.5 GiB) |
|---|---|
| `build-deb` / `build-rpm` (amd64) | 5.8 / 6.1 · 5.8 / 6.8 min |
| `test-deb` / `test-rpm` (amd64) | 2.7 / 3.0 · 3.9 / 2.8 min |
| `build-deb-arm64` / `build-rpm-arm64` | 5.7 / 5.5 · 6.8 / 7.2 min |
| `test-deb-arm64` / `test-rpm-arm64` | 4.5–4.7 · 5.7–6.6 min |
| `build-windows-i686` | 6.5 min |
| `package-windows-i686` / `test-windows-i686` | 3.8 / 1.9 min |
| `upload-*` / `detect-changes` | 0.3–0.9 min |

**Expectation, stated before the fact:** the compiles should get *faster*, not merely stay equal. On CodeBuild they run against `dockerd --storage-driver vfs`, which copies whole layers instead of using copy-on-write; on the hosted images Docker runs on `overlay2`. If the compile durations do not improve, that assumption was wrong and is worth understanding before merging.

### Measured results

All four workflows pass. Durations from the green run, against the CodeBuild baseline above:

| Job | CodeBuild | GitHub-hosted | Factor |
|---|---|---|---|
| `build-deb` (amd64) | 5.8 / 6.1 min | **4.2 min** | 0.70× |
| `build-rpm` (amd64) | 5.8 / 6.8 min | **4.6 min** | 0.73× |
| `test-deb` (amd64) | 2.7 / 3.0 min | **1.9 min** | 0.67× |
| `test-rpm` (amd64) | 3.9 / 2.8 min | **1.8 min** | 0.54× |
| `build-deb-arm64` | 5.7 / 5.5 min | **3.8 min** | 0.68× |
| `build-rpm-arm64` | 6.8 / 7.2 min | **3.5 min** | 0.50× |
| `test-deb-arm64` | 4.5–4.7 min | **1.0 min** | **0.22×** |
| `test-rpm-arm64` | 5.7–6.6 min | **1.3 min** | **0.21×** |
| `build-windows-i686` | 6.5 min | **5.1 min** | 0.78× |
| `package-windows-i686` (Generate MSI) | 3.8 min | **1.6 min** | 0.42× |
| `test-windows-i686` (Test MSI) | 1.9 min | **5.9 min** | **3.1×** |
| `detect-changes` | 0.9 min | **0.6 min** | 0.67× |
| `upload-*-to-s3` | 0.3–0.9 min | 0.3–0.5 min | ≈ or better |

The macOS `Build PKG` / `Test PKG` legs run on unchanged runners, so their variation is noise rather than signal and is omitted.

**The `vfs` → `overlay2` expectation held.** Every compile came in at 0.50–0.78× of its CodeBuild time.

**An unanticipated result: the arm64 test jobs dropped to ~0.21×**, from 4.6 and 6.2 minutes to 1.0 and 1.3. This is the same finding recorded on #38174 as a follow-up for #38139 — on CodeBuild those jobs spent ~2.9 minutes just pulling the test image on ARM, against 0.8 minutes for the same step on amd64, and that gap was estimated at up to ~6 % of the Agent fleet's monthly minutes. **The migration removes it outright**, with no caching work needed. That follow-up can be closed for these workflows.

### Two failures the first run exposed, and their causes

Both were real, not artifacts of the GitHub incident that was in progress: a second run after the incident reproduced them, and the same workflows pass on other branches still on CodeBuild. Fixed in the second commit on this branch.

**`Test DEB` / `Test RPM` amd64 — `check_files` ran unprivileged.** They failed with "141 expected files were not found" and "7 expected files were not found", the latter listing core binaries such as `/var/ossec/bin/wazuh-control`, even though the install had succeeded. The package creates `/var/ossec` root/wazuh-owned with restrictive permissions and `check_files.py` must stat every file under it; an unreadable file is reported as a missing one. Commit `36e14ab75f`, which migrated CI to CodeBuild, dropped the `sudo` this action originally had — correctly, since self-hosted and CodeBuild runners both execute as root. GitHub-hosted runners are the first environment where the runner is unprivileged. Corroborated independently by `5_builderpackage_agent-macos.yml`, which never ran on CodeBuild, runs on `macos-14` as an unprivileged user, and is the one consumer that already invokes `check_files.py` with `sudo`. The elevation added here is conditional on `id -u` rather than a bare `sudo`, because `5_builderpackage_manager.yml` shares the action and stays on CodeBuild.

**`Test MSI` — the runner OS changed the expected SCA profile.** It failed with 2 missing files and 1 *unexpected* one: `profile-10.template` and `ruleset\sca\cis_win10_enterprise.yml` absent, `ruleset\sca\cis_win2025.yml` present instead. The unexpected file is what identifies the cause — nothing was missing, the MSI installed a different SCA profile because it selects one by detected OS version, and `windows_agent_base.csv` encodes the Server 2022 profile. Resolved by targeting `windows-2022`, which both previous environments were. Making the baseline OS-aware is worth doing but changes test expectations rather than runner configuration, so it belongs in its own change.

### Known regression, documented rather than fixed here

`Test MSI` went from 1.9 to 5.9 minutes. The step breakdown puts **all** of it in one place — a single 213-second gap inside `Test uninstall and upgrade from 4.14.7 to 5.0.0`, before the line that reports Wazuh as already installed. Every other step in that job matches CodeBuild to within a few seconds, and the remaining 44 seconds of that step (uninstall, download, install, upgrade, verify) are equivalent too.

The cause is in `smoke_upgrade_test.py`, not in the runner choice:

```python
check_cmd = 'wmic product where "Name like \'%Wazuh%\'" get Name, Version /format:list'
```

`wmic product` queries the WMI `Win32_Product` class, which triggers an MSI consistency check of **every installed product** on the machine, so its cost scales with how many MSI packages the image carries. The `windows-2022` GitHub image ships Visual Studio Enterprise, SDKs and dozens of MSI products; CodeBuild's `windows-base:2022` container is far leaner. Same command, far more products to verify. The log confirms the class: `Executing (\\runnervmm66xd\ROOT\CIMV2:Win32_Product…`.

Worth noting what this rules out: the download of the 4.14.7 MSI happens *after* the line that closes the gap, so cross-cloud download latency is not the cause, and Windows Defender real-time scanning would have shown up in the install phases, which are unchanged.

The standard fix is to read the registry uninstall keys (`HKLM:\SOFTWARE\[WOW6432Node\]Microsoft\Windows\CurrentVersion\Uninstall\*`) instead — instant, no consistency check, same information, and Microsoft advises against `Win32_Product` for exactly this reason. It is left out of this PR deliberately: it changes test-script logic rather than runner configuration, the uninstall path a few lines below uses the same `wmic product call uninstall` and would need replacing consistently, and the pattern may exist in other smoke-test scripts.

In balance for the Windows workflow as a whole, `build-windows-i686` saved 1.4 minutes and `Generate MSI` saved 2.2, so the net change across the workflow is roughly +0.2 minutes.

## Artifacts Affected

Not a flat "none" — this needs splitting by leg.

**Linux DEB/RPM (amd64 and arm64) and the Windows cross-compile:** package content should be unaffected. The compilation happens *inside* the same builder container images pulled from GHCR; only the host running Docker changes. What changes is the storage driver beneath the container, not the toolchain inside it.

**The MSI, `package-windows-i686`:** this one deserves review attention. It runs natively on Windows, not in a container, so the environment genuinely changes — from CodeBuild's `windows-base:2022` image to GitHub's Windows Server 2025, and from a downloaded WiX 3.14.1 to the image's bundled WiX v3.14. Both are WiX v3 and the `candle`/`light` interface is the same, but this is the one leg where the build environment itself moves rather than just its host.

**`test-*` jobs:** these install and verify real packages natively, so they exercise a different OS version than before. That is arguably an improvement in coverage rather than a risk, but it is a change in what is being tested.

## Configuration Changes

CI-only. Four workflow files under `.github/workflows/` and one deleted composite action. No product or agent configuration changes. No CodeBuild project, fleet or IAM change is required, and none was made — the fleets simply lose their last consumers in this repository.

## Documentation Updates

None required. Two inline comment corrections are included: the `reinstall_cmake` step's `TODO(codebuild-temp)` is replaced with an accurate description of why it exists, and the VC Build Tools step's comment now says what to look for in its own log to decide whether it can be dropped.

## Tests Introduced

None — this is CI configuration with no product code. The verification is this PR's own CI: all four workflows are `pull_request`-triggered, so every migrated job runs as part of review, and the durations can be compared directly against the CodeBuild baseline above.

## Review Checklist

- [ ] Code changes reviewed


cborla (approved):
LGTM
- [x] Code changes reviewed

PR #38231, opened 2026-08-07
Stabilize failing CI checks on the 5.0.0-https branch

### LINKED ISSUE #38218: Fix failing CI checks on the 5.0.0-https branch
## Description

The `5.0.0-https` objective branch (#37702) currently carries several failing CI checks. Some are real defects, some are flaky, and some are environmental, but they are red on every PR targeting the branch. This masks genuine failures (a real broken test gets attributed to the known flaky noise and merged) and makes the CI signal untrustworthy for the whole HTTPS effort.

This issue tracks driving the branch's checks back to green so that a red check on a PR means something again.

### 1. `inventorysyncserver-modules` — real test failure (already on base)

`src/wazuh_modules/inventory_sync_server/test/unit/configEndpoint_test.cpp`, test `BodiesWithoutAModulesObjectAreRejected` (line 332): the first body in the rejection loop is `{"modules":{"fim":{}}}`, which is a **valid** report (a non-empty `modules` object whose module value is an object). The endpoint accepts it with `200`, but the loop asserts `400`, so the test fails at runtime. The inline comment still reads `// the legacy array wire, no longer accepted`, confirming the value was meant to be the legacy array.

Fix: restore the array literal on line 332:

```cpp
for (const auto* body : {R"([{"module":"fim","config":{}}])", // the legacy array wire, no longer accepted
```

An array parses as non-object, is rejected with `400`, and the test verifies what its name claims.

### 2. `inventorysyncserver-asan` / `inventorysyncserver-modules` — flaky concurrency tests

`UdsHttpServerTest.ThreeHundredConcurrentDeferralsOnTwoIoThreads` and `UdsShutdownTest.StopAcceptingThenStopUnderManyLiveDeferrals` (inventory-sync-server transport, from #38077) fail non-deterministically under CI load / ASan slowdown, with a plain assertion failure and no sanitizer error. They pass on re-run. Needs stabilization (deterministic synchronization instead of timing assumptions) or an explicit quarantine with an owner.

### 3. API endpoints integration tests

`test_agent_GET_endpoints`, `test_agent_PUT_endpoints`, `test_agent_DELETE_endpoints`, and the `test_rbac_*_agent`/`test_rbac_*_overview` tavern suites fail on the base branch directly (observed on a push run for `84ae9576`, with no PR involved). Needs triage: adapt the fixtures to the current API surface or fix the underlying regression.

### 4. `build-sonoma` — environmental

Fails downloading a build dependency: `curl: (28) Failed to connect to cmocka.org port 443`. Network/infra timeout, not a code defect. Confirm whether the dependency source needs mirroring or the job hardening (retry/cache).

### 5. `vulnerability-scanner-database` precompiled workflow

Failing on recent pushes to `5.0.0-https`. Needs triage to classify as regression vs environmental.

## Definition of Done

- [ ] `configEndpoint_test.cpp` rejection test fixed (item 1).
- [ ] Flaky transport tests stabilized or quarantined with an owner (item 2).
- [ ] API endpoints integration suites triaged and fixed/adapted (item 3).
- [ ] `build-sonoma` dependency-download failure resolved or the job hardened (item 4).
- [ ] `vulnerability-scanner-database` workflow triaged (item 5).
- [ ] Base-branch CI is clean, so a red check on a PR is meaningful again.

Related to the HTTPS comms objective #37702.

### LINKED ISSUE #38228: API integration tests failing on `5.0.0-https`: GET/PUT/DELETE agent endpoints return 500 Wazuh Internal Error
## Description

The `5.X - API endpoints - Integration tests` workflow fails on branch `5.0.0-https` (commit `84ae957631ed5cc507d6d7bcfd4d9517feb46591`). Reproduced manually on the branch outside of CI as well.

CI run: https://github.com/wazuh/wazuh/actions/runs/31031217617

Failing jobs:
- `test_agent_GET_endpoints.tavern.yaml` — 31 failed / 54 passed
- `test_agent_PUT_endpoints.tavern.yaml` — 7 failed / 2 passed
- `test_agent_DELETE_endpoints.tavern.yaml` — 1 failed / 5 passed

## Symptom

Every failing case returns `500` with the same body instead of the expected `200`:

```json
{
  "title": "Wazuh Internal Error",
  "detail": "Wazuh Internal Error",
  "remediation": "Please, check `WAZUH_HOME/logs/wazuh-manager.log`, `WAZUH_HOME/logs/cluster.log` and `WAZUH_HOME/logs/api.log` to get more information about the error",
  "dapi_errors": {"master-node": {"error": "Wazuh Internal Error. See log for more detail"}},
  "error": 1000
}
```

The identical error signature across GET/PUT/DELETE and otherwise unrelated endpoints suggests a common failure in the master-node DAPI request path for agent/group operations, not isolated per-endpoint bugs.

## Affected endpoints (representative — full list in run logs)

- **GET**: `/agents`, `/agents/{agent_id}/key`, `/agents/{agent_id}/stats/{component}`, `/agents/{agent_id}/daemons/stats`, `/agents/summary`, `/agents/summary/status`, `/agents/outdated`, `/agents/stats/distinct`, `/groups/{group_id}/agents` (+ sort/select variants)
- **PUT**: `/agents/group`, `/agents/group/{group_id}/restart`, `/agents/restart`, `/agents/{agent_id}/restart`, `/agents/upgrade`, `/agents/{agent_id}/group/{group_id}`, `/groups/{group_id}/configuration`
- **DELETE**: `/agents`

## Additional context

Manager/cluster/API logs (`wazuh-manager.log`, `cluster.log`, `api.log`) aren't captured as CI artifacts, so the root-cause traceback isn't visible from the workflow run alone — only the generic DAPI error envelope. Suggest uploading these as artifacts on failure to speed up triage.

## Steps to reproduce

1. Checkout `5.0.0-https` at `84ae957631` or later.
2. Trigger the "5.X - API endpoints - Integration tests" workflow (or run the tavern suites locally against a manager built from this branch).
3. Observe GET/PUT/DELETE agent endpoint tests failing with 500 Wazuh Internal Error.

Antoniogm03:
## Description

Related to #38218. Several CI checks were red on every PR into `5.0.0-https`, masking real failures behind known noise. This started by fixing the four items originally tracked there, then expanded significantly once fixing the noise exposed a genuine, previously-hidden regression behind the API integration tests (#38228): a chain of real bugs in how agent version/connection data flows from a real 5.0+ agent through remoted, wdb and the Server API, plus a CI infrastructure bug that was silently re-testing stale code on every push.

### Original CI-noise fixes (issue #38218)

- **inventorysyncserver-modules**: `BodiesWithoutAModulesObjectAreRejected` asserted a 400 for `{"modules":{"fim":{}}}`, but that's actually the current, valid request shape — `POST /config` moved off its legacy array format a while ago, and this object form now passes validation and returns 200. The test's own name and comment describe an input that should be rejected, so it was simply sending the wrong body. Restored the array literal it was meant to send, which does correctly fail and return 400.

- **inventorysyncserver-asan / -modules**: `ThreeHundredConcurrentDeferralsOnTwoIoThreads` and `StopAcceptingThenStopUnderManyLiveDeferrals` waited for 300 dispatched requests using a fixed-deadline `sleep_for` poll, racing against however long those connections actually take to land under ASan instrumentation or heavier CI load — a slow-but-healthy run looked identical to a hang. Replaced the polling with a condition variable that's signaled the moment each request is dispatched, so the test only waits as long as it actually needs to. The second test also had its own unrelated timing issue: a 2-second drain timeout wasn't always enough for two I/O threads to flush ~100 in-flight writes before the server force-closed them, so it was widened to 10s.

- **API agent/RBAC integration tests**: `test_agent_GET_endpoints.tavern.yaml` still asserted a 200 with the old response schema for `GET /agents/{agent_id}/config/{component}/{configuration}`, an endpoint a previous PR already removed (it now correctly 404s). The RBAC fixtures had been updated for this removal everywhere except this one leftover block. Dropped it.

- **build-sonoma**: this job (`5_testunit_macos.yml`) downloaded the cmocka test library straight from `cmocka.org` with a single plain `curl`, no retry and no fallback, so any transient hiccup against that one host failed the whole job outright with `curl: (28) Failed to connect ... port 443` — unrelated to anything in the actual code being tested. A different workflow (`install_build_deps/action.yml`) already solves this same problem for the same dependency by retrying and, if the primary host keeps failing, falling back to a GitLab mirror. Applied that identical retry-then-mirror logic here. Verified the mirror actually serves the right thing before relying on it: confirmed via `curl` that it responds with the expected cmocka 1.1.7 release, and downloaded it to check the archive extracts into the same top-level directory name (`cmocka-cmocka-1.1.7`) the build script expects, so the fallback path doesn't break the build in a different way.

- **vulnerability-scanner-database (re-diagnosed)**: the issue as filed pointed at the wrong workflow (one that's out of scope for this branch); the actual failing job is the vulnerability-scanner integration test. Its QA tool sent an event and then just slept a fixed 30 seconds before reading the manager's output log, instead of waiting for the manager's actual "finished processing" signal — under CI load, a scan that legitimately took longer than 30 seconds produced a log file the test read too early and flagged as incomplete. Fixed by waiting on the real completion signal, keeping the 30s only as a deadlock-guard timeout.

- **agentSession.hpp coverage**: this file's test coverage had dipped slightly below the CI's required 90% gate. Added tests to bring it back above threshold.

- **RBAC catalog fixture**: the RBAC actions catalog is generated dynamically from the live API spec, so removing an endpoint from the spec already stops it from being reported — but a companion test fixture had that same endpoint hardcoded as an expected value, so it kept asserting something the real code could no longer produce. Removed the stale entry to match.

- **`legacy_task_delivery.c` (scan-build)**: a null-pointer check was already present and effectively safe at runtime, but scan-build analyzes the relevant library as a separate compilation unit and couldn't see that guarantee, so it flagged a dereference that could never actually happen. Added an explicit, local null check so the static analyzer can verify safety without needing cross-unit knowledge.

- **CI log visibility**: the test suite already copies the manager's `api.log`, `cluster.log`, `wazuh-manager.log` (per cluster node) and each agent's `ossec.log` to disk whenever a test fails, but the CI workflow never published that directory anywhere — every failure investigation had to be reproduced from scratch with no server-side logs to go on. Added a step to upload those logs as a build artifact on failure, for both the main test job and the smoke-test job.

### Root cause of #38228: API integration tests returning 500/wrong data for agent endpoints

- **Stale Docker image cache masked every fix below.** The job that builds the test manager's Docker image only invalidates its cache when files under the test-environment folder change — but the image itself bakes in the branch's actual source code, fetched fresh at build time. That meant a commit touching only `src/` or `framework/` (which is every fix in this section) counted as a cache hit, so CI kept quietly reusing an old image and none of these fixes were actually being exercised in CI until this was caught and fixed by including the commit hash in the cache key.

- **A malformed agent version was crashing every agent-listing endpoint.** When an agent reports an invalid version during startup, the manager correctly rejects the connection — but it was still writing that raw, malformed version string into the agent's database record instead of a safe placeholder. The API's version-parsing logic expects a strict `MAJOR.MINOR.PATCH` format and throws an unhandled error on anything else, so the moment a single agent had ever sent a bad version, listing agents (`GET`, `PUT`, `DELETE`, and the agents overview) returned a 500 error for *every* agent in the system, not just the one with the bad data. Fixed by storing a safe placeholder value instead of the raw rejected string, and additionally hardened the parser itself to skip any unparseable value with a warning rather than crashing the whole request — so this class of bug can't resurface from anywhere else either.

- **Agents could never reach `active` status in the test environment.** Test agents connect to the manager through a load balancer, and modern (5.0+) agents only talk to the manager over a new HTTPS-based control channel on a specific port — the old connection method was removed for them entirely. The load balancer's configuration was simply never updated to forward that new port, so every real agent behind it got its connection refused indefinitely and never registered as connected. This single gap explained most of what was still failing after the version fix: agents showing up with no data at all, delete operations failing because the target agent was never "active," restart/upgrade eligibility checks failing, and summary counts showing agents stuck as disconnected. Fixed by adding the missing routing rule for that port.

- **The per-agent daemon-stats endpoint had its version check backwards.** This endpoint is only meant to work for legacy (pre-5.0) agents, since modern agents report their own statistics over HTTPS instead and were never going to have the kind of data this endpoint looks up. The check that was supposed to enforce this was inverted — it rejected exactly the legacy agents that do have this data, and let through the modern agents that never will, so the endpoint could never succeed for anyone. Fixed by flipping the comparison so it now allows the agents it was actually meant for.

- **Several integration tests were still asserting years-old "nothing happened" results.** Once real agent version and connection data started flowing correctly for the first time (thanks to the fixes above), it became clear that a handful of test cases had been written against effectively broken/empty data and had never actually observed real behavior — they expected empty, no-op responses for scenarios that, with real data, correctly produce a specific business error (for example, "this agent is already on the requested version," or "the update package couldn't be reached"). Updated each of these test expectations to match the correct, now-observable behavior, double-checking against the manager's real validation logic for each case rather than guessing.


jotacarma90 (approved):
LGTM

jotacarma90:
Full execution API endpoint IT:
https://github.com/wazuh/wazuh/actions/runs/31599767574

Issue #38143, opened 2026-08-03
Replace order-preserving std::vector::erase with swap-and-pop where element order doesn't matter


Miguevrgo:
## Description

Audited `std::vector` erase/`remove_if` usages across the C++ codebase (dbsync, syscheckd db, `shared_modules/utils`, engine sync modules) looking for spots where element order is irrelevant, so the O(n) shifting erase could be replaced by a swap-with-back + `pop_back` ("swap_remove" pattern) for O(1) removal.

## Confirmed candidates

Order does not matter at these sites, verified by reading the surrounding logic:

1. `src/shared_modules/utils/rocksDBWrapper.hpp:587` and `:639` — `m_columnsInstances` (`std::vector<ColumnFamilyRAII>`) erase inside `deleteAll()` and `deleteAll(const std::string&)`. Access to this vector always goes through `m_columnsIndex` (a name→index map), and `rebuildColumnsIndex()` is called right after every erase, so vector position is never meaningful.

2. `src/engine/source/iocsync/src/iocsync.cpp:299-307` — `removeIOCTypeFromSync()` uses `std::remove_if` + `erase` on `m_databasesState`. `addIOCTypeToSync()` rejects duplicate IOC types, so at most one element ever matches — this is effectively a find + single-element erase. State is persisted via `saveStateToStore()` and reloaded by field (`getIocType()`) comparison, not by index, so order is not semantically significant.

3. `src/engine/source/cmsync/src/cmsync.cpp:465-473` — `removeSpaceFromSync()`, the exact same pattern as #2 on `m_namespacesState` (`addSpaceToSync()` also rejects duplicates). Persisted/reloaded by `getOriginSpace()` field lookup, not position.

## Checked but excluded

- dbsync (`sqlite_dbengine.cpp` and friends): only map/unordered_map erases, no plain vector erase of elements.
- `src/engine/source/iockvdb/src/manager.cpp:540` (`tryCleanRetired()`): `m_retiredQueue` is declared as `std::deque<RetiredInstance>` (`manager.hpp:119`), not a vector, and is a FIFO-ish cleanup queue — out of scope, swap_remove is not the right pattern for a deque anyway.
- `gapSet.hpp`, `scheduler.hpp`, `router/table.hpp`: ordered containers (set/sorted structures) where order is load-bearing — not applicable.
- string/JSON erase sites, `/proc/net/dev` parsing, logpar args: order is semantically meaningful — not applicable.

## Open question

[...]

PR #38234, opened 2026-08-07
Remove avoidable copies, double lookups and pessimizing moves in agent-side hot paths

### LINKED ISSUE #38173: Remove avoidable copies, double lookups and pessimizing moves in agent-side hot paths
## Description

Audited agent-side C++ (`wazuh_modules/{syscollector,sca,agent_info}`, `data_provider`, `shared_modules`, `syscheckd`, `client-agent`) for copies that the compiler cannot elide: by-value range-for over non-trivial element types, by-value locals bound from container elements, redundant `std::move` on prvalues, double container lookups, and vectors filled in a loop with a size known upfront.

Each site below is a local, behavior-preserving change. Ordered by expected impact.

### 1. `sca_impl/src/sca_event_handler.cpp:339-360` and `:379-392` — JSON copies per policy and per check

```cpp
// :339-345
const std::string policyId = policyEntry.first;                 // copy of the key
const nlohmann::json policyData = policyEntry.second["data"];   // copy of the policy JSON
const int policyResult = policyEntry.second["result"];

const std::vector<nlohmann::json> checksForPolicy = GetChecksForPolicy(policyId);

for (auto checkData : checksForPolicy)                          // copy of the check JSON per iteration
{
    ...
    const nlohmann::json event = {{"policy", policyData}, {"check", checkData}, ...};
    events.push_back(event);                                    // copy of the whole event
}
```

`checkData` is only read inside the loop (it is copied again into `event`), so `const auto&` is sufficient. Same for `policyId`/`policyData`. `events.push_back(event)` can be `events.push_back(std::move(event))` (dropping the `const` on `event`) or a direct `events.emplace_back(...)` of the initializer list. `ProcessEvents()` runs over every modified policy and check on every SCA scan, so this is the site with actual volume behind it.

Double lookup in the same function:

```cpp
// :379-382
if (modifiedPoliciesMap.find(policyId) != modifiedPoliciesMap.end())
{
    policyData = modifiedPoliciesMap.at(policyId)["data"];      // second lookup of the same key
}
```

Keep the iterator returned by `find()` and dereference it.

### 2. `data_provider/src/ports/portLinuxWrapper.h:121-123` — redundant `std::move` on a prvalue

```cpp
explicit LinuxPortWrapper(const PortType type, const std::string& row)
    : m_fields{ Utils::split(row, ' ') }
    , m_remoteAddresses { std::move(Utils::split(m_fields.at(REMOTE_ADDRESS), ':')) }
    , m_localAddresses { std::move(Utils::split(m_fields.at(LOCAL_ADDRESS), ':')) }
    , m_queue { std::move(Utils::split(m_fields.at(QUEUE), ':')) }
{ }
```

`Utils::split()` returns by value, so the initializers are already prvalues that get copy-elided. Wrapping them in `std::move` suppresses elision and forces a move constructor call instead. Removing the three `std::move` is strictly better and cannot change behavior. Note this constructor runs once per socket parsed from `/proc/net/*`, so it is a genuinely hot path on hosts with many connections.

### 3. Range-for copies of `std::filesystem::path`

- `data_provider/src/extended_sources/browser_extensions/src/chrome.cpp:566`, `:572`, `:718`
- `data_provider/src/extended_sources/browser_extensions/src/firefox.cpp:81`, `:107`

```cpp
for (auto subDir : m_fileSystemWrapper->list_directory(extensionPath))   // copies a path per entry
```

`const auto&` for the loop variable. The temporary returned by `list_directory()` has its lifetime extended by the range-for, so only the element type needs changing.

Related, same file family: `ie_explorer.cpp:225`, `IEExtensionsProvider::expandUserKey(const std::string key)` takes the key by value and only concatenates it. `const std::string&`.

### 4. `shared_modules/utils/observer.hpp:85-97` — by-value fan-out in `Subject<T>`

```cpp
void setData(T newData)      { notifyObservers(newData); }
void notifyObservers(T data)
{
    std::lock_guard<std::mutex> lock(mutex);
    for (auto observer : observers)      // copies a shared_ptr (two atomic ops) per observer
    {
        observer->update(data);
    }
}
```

Two independent items here:

- `for (const auto& observer : observers)`: removes an atomic refcount increment/decrement per observer per notification. Local, always safe.
- `setData(T)` / `notifyObservers(T)` / `virtual void update(T)` (`:39`) taking `T` by value: for a `T` like `nlohmann::json` this is a copy per call plus one per observer. Stated honestly, this one is template hygiene rather than a measurable win today: the only current instantiation in tree is `Observer<nlohmann::json&>` in `vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp:143`, where `T` is already a reference, and `provider.hpp`/`utils/subscriber.hpp` have no agent-side consumer. Changing `update()`'s signature also touches every implementor, so it is the widest diff in this issue and can be dropped from scope if reviewers prefer.

`Subject::detach()` in the same class is already under discussion in #38143 (swap-and-pop vs. notification order), so both changes should land coordinated to avoid conflicts.

### 5. `reserve()` where the final size is already known

- `shared_modules/dbsync/src/sqlite/sqlite_dbengine.cpp:1067-1072` (`getPrimaryKeysFromTable`): `primaryKeyList.reserve(tableFields.size())` before the loop. Upper bound, exact size is the PK subset.
- `data_provider/src/extended_sources/browser_extensions/src/ie_explorer.cpp:227-232`: bind `listSubKeys(HKEY_USERS)` to a local and `expandedKeys.reserve(subKeys.size())`.

Both are one-liners using a `size()` that is already available, with no extra pass over the data.

## Checked but excluded

- `wazuh_modules/syscollector/src/syscollectorImp.cpp:1458` and `:1484` — `for (auto addressTableData : item.at("IPv4"))` looks like an accidental copy but is not: the body mutates `addressTableData` (adds `interface_name`, `checksum`, erases `network_dhcp`/`network_metric`) and then does `push_back(std::move(...))`. Correct as written.
- `data_provider/src/sysInfoLinux.cpp:387` — `for (auto addr : interface.second)`: `addr` is an `ifaddrs*`, copying a pointer is free.
- `shared_modules/dbsync/src/sqlite/sqlite_dbengine.cpp` — the many `m_tableFields[table]` uses copy the whole `TableColumns` vector on every call, including per-row paths, because `Utils::MapWrapperSafe::operator[]` (`shared_modules/utils/mapWrapperSafe.h`) returns `Value` **by value** on purpose, under its mutex. The sites written `const auto& x { m_tableFields[table] }` are only lifetime-extending that temporary, so they copy too. This is a real and probably significant cost, but removing it means changing `MapWrapperSafe`'s API (an accessor invoked under the lock, or returning `shared_ptr<const Value>`), which is a design change with thread-safety implications and does not belong in a mechanical-cleanup issue. Worth its own issue.

## Proposed fix

Apply items 1, 2, 3 and 5 as described. For item 4, apply the `const auto&` loop change unconditionally and treat the `T`-by-value signature change as optional scope.

No functional change is expected anywhere in this issue, so existing unit tests are the regression net.

## Ask

Confirm whether item 4's `Observer<T>::update(T)` signature change is in scope, given it touches every implementor and overlaps with #38143.

Miguevrgo:
|Related issue|
|---|
|#38173|

## Description

Closes #38173.

An audit of the agent-side C++ (`data_provider`, `shared_modules`) surfaced a set of copies the compiler cannot elide, one pessimizing `std::move`, and two vectors grown element by element when their final size is already known. Every site below is local and behavior-preserving; the value is in the hot paths they sit on.

### Pessimizing `std::move` on prvalues — `src/data_provider/src/ports/portLinuxWrapper.h:118`

```cpp
explicit LinuxPortWrapper(const PortType type, const std::string& row)
    : m_fields{ Utils::split(row, ' ') }
    , m_type { type }
    , m_remoteAddresses { std::move(Utils::split(m_fields.at(REMOTE_ADDRESS), ':')) }
    , m_localAddresses { std::move(Utils::split(m_fields.at(LOCAL_ADDRESS), ':')) }
    , m_queue { std::move(Utils::split(m_fields.at(QUEUE), ':')) }
{ }
```

`Utils::split()` returns a `std::vector<std::string>` by value, so each initializer is already a prvalue that gets copy-elided straight into the member. Wrapping it in `std::move` turns the prvalue into an xvalue, which makes it ineligible for elision and forces an actual `std::vector` move constructor call instead. Three of them, per port.

This constructor runs once for every line parsed out of `/proc/net/{tcp,tcp6,udp,udp6}`, so on a host with a large connection table it is a genuinely hot path. Dropping the three `std::move` cannot change behavior and is strictly better codegen.

### By-value range-for over `std::string` — `chrome.cpp`, `firefox.cpp`, `ie_explorer.cpp`

The browser-extension providers walk directory listings with the loop variable taken by value, then reassign it in place:

```cpp
// src/data_provider/src/extended_sources/browser_extensions/src/firefox.cpp:73
for (auto userHome : Utils::enumerateDir(homePath))
{
    if (Utils::startsWith(userHome, ".")) { continue; }
    userHome = Utils::joinPaths(homePath, userHome);
    ...
}
```

`Utils::enumerateDir()` returns `std::vector<std::string>`, so `auto` copies a `std::string` per entry, and the reassignment then destroys that copy to build the joined path. Binding the element by `const auto&` and introducing a separate local for the joined path removes the per-entry copy and, as a side effect, stops the loop variable from meaning two different things (the bare entry name on the first line, the absolute path afterwards). Same pattern at `chrome.cpp:559`, `chrome.cpp:565` and `chrome.cpp:711`, and at `firefox.cpp:99`.

`ie_explorer.cpp` had two by-value parameters on the same theme:

```cpp
// src/data_provider/src/extended_sources/browser_extensions/include/ie_explorer.hpp:144,158
void getRegistryPathsFromKeys(HKEY hive, const std::vector<std::string> clsidKeys, std::vector<std::string>& regPaths);
std::vector<std::string> expandUserKey(const std::string key);
```

Both are read-only inside the body (`clsidKeys` is only iterated, `key` is only concatenated), so the by-value copy is pure cost. Note `const std::vector<std::string>` by value is also unmovable at the call site precisely because it is `const`, so every caller pays a full vector deep copy.

### Copied `shared_ptr` per notification — `src/shared_modules/utils/observer.hpp:93`

```cpp
void notifyObservers(T data)
{
    std::lock_guard<std::mutex> lock(mutex);
    for (auto observer : observers)   // observers is std::vector<std::shared_ptr<Observer<T>>>
    {
        observer->update(data);
    }
}
```

`auto` copies the `shared_ptr`, which is an atomic refcount increment on entry and an atomic decrement on scope exit, for every observer on every notification, while the mutex is held. The pointee is only dereferenced, never rebound or stored, so `const auto&` is sufficient.

### `reserve()` where the size is already known

`SQLiteDBEngine::getPrimaryKeysFromTable()` (`src/shared_modules/dbsync/src/sqlite/sqlite_dbengine.cpp:1009`) filters `tableFields` into `primaryKeyList` one `emplace_back` at a time. `tableFields.size()` is an exact upper bound and is available before the loop, so a single `reserve()` removes the reallocation chain at no extra pass over the data.

`IEExtensionsProvider::expandUserKey()` had the same shape plus a subtlety: the range-for iterated directly over `listSubKeys(HKEY_USERS)`, so the size was never observable. Binding the result to a local makes `reserve()` possible without calling the enumeration twice.

### Scope notes

Two items from the issue are deliberately not here:

- Item 1 (`sca_impl/src/sca_event_handler.cpp`) does not exist on `4.14.8`; that file belongs to the 5.0 line.
- Item 4's `Observer<T>::update(T)` / `Subject<T>::setData(T)` by-value signature change is left out. It touches every implementor, the only in-tree instantiation already has `T` as a reference, and it overlaps with the `Subject::detach()` work in #38143. Only the `const auto&` loop change was applied, which is the part that is unconditionally safe.

### Reviewing the diff

Two commits: `b310b28` is the entire behavioral change (30 lines), `2df616c` is `clang-format` over the touched files against the repo's tracked `.clang-format`. The formatting commit is what makes the diff look large. Reviewing with `git diff -w` collapses it to the real change.

## Configuration options

N/A. No configuration parameter or default value changed.

## Logs/Alerts example

N/A. No user-visible log or alert changed, this PR has no functional effect.

Evidence is the clean rebuild of the affected components. Both objects were forced to recompile (`touch` on the modified sources) so the build below is not a no-op:

`data_provider`, with `-Wall -Wextra`:

```
[ 45%] Building CXX object src/extended_sources/browser_extensions/CMakeFiles/browser_extensions.dir/src/chrome.cpp.o
[ 48%] Building CXX object src/extended_sources/browser_extensions/CMakeFiles/browser_extensions.dir/src/firefox.cpp.o
[ 51%] Linking CXX static library ../../../lib/libbrowser_extensions.a
[ 51%] Built target browser_extensions
[ 54%] Building CXX object CMakeFiles/sysinfo.dir/src/sysInfoLinux.cpp.o
[ 58%] Linking CXX shared library lib/libsysinfo.so
[ 93%] Built target sysinfo
[ 96%] Linking CXX executable ../bin/sysinfo_test_tool
[100%] Built target sysinfo_test_tool
```

`dbsync`:

```
[ 10%] Building CXX object CMakeFiles/dbsync.dir/src/sqlite/sqlite_dbengine.cpp.o
[ 20%] Linking CXX shared library lib/libdbsync.so
[ 60%] Built target dbsync
[ 70%] Linking CXX executable ../bin/dbsync_example
[ 80%] Built target dbsync_example
[ 90%] Linking CXX executable ../bin/dbsync_test_tool
[100%] Built target dbsync_test_tool
```

Full agent target:

```
[100%] Built target wazuh-syscheckd
make[2]: Leaving directory '/home/miguevr/Work/wazuh-38173/src/syscheckd/build'

Done building agent
```

No new warnings were emitted by any of the touched translation units.

Formatting conformance against the repo's tracked `.clang-format` (empty output means no diff):

```
$ for f in src/data_provider/src/extended_sources/browser_extensions/include/ie_explorer.hpp \
           src/data_provider/src/extended_sources/browser_extensions/src/chrome.cpp \
           src/data_provider/src/extended_sources/browser_extensions/src/firefox.cpp \
           src/data_provider/src/extended_sources/browser_extensions/src/ie_explorer.cpp \
           src/data_provider/src/ports/portLinuxWrapper.h \
           src/shared_modules/dbsync/src/sqlite/sqlite_dbengine.cpp \
           src/shared_modules/utils/observer.hpp; do
      clang-format --dry-run -Werror "$f"
  done
$
```

## Tests

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests

### DIFF
--- a/src/data_provider/src/extended_sources/browser_extensions/include/chrome.hpp
+++ b/src/data_provider/src/extended_sources/browser_extensions/include/chrome.hpp
@@ -11,6 +11,7 @@
 
 #include <string>
 #include <vector>
+#include <tuple>
 #include "json.hpp"
 #include "browser_extensions_wrapper.hpp"
 
--- a/src/data_provider/src/extended_sources/browser_extensions/include/ie_explorer.hpp
+++ b/src/data_provider/src/extended_sources/browser_extensions/include/ie_explorer.hpp
@@ -141,7 +141,7 @@ class IEExtensionsProvider
          * @param clsidKeys List of CLSID subkeys to enumerate.
          * @param regPaths Vector to append found registry paths.
          */
-        void getRegistryPathsFromKeys(HKEY hive, const std::vector<std::string> clsidKeys, std::vector<std::string>& regPaths);
+        void getRegistryPathsFromKeys(HKEY hive, const std::vector<std::string>& clsidKeys, std::vector<std::string>& regPaths);
 
         /**
          * @brief Enumerate subkeys under a given registry hive.
@@ -155,7 +155,7 @@ class IEExtensionsProvider
          * @param key Registry path string with possible user placeholders.
          * @return Expanded list of registry paths.
          */
-        std::vector<std::string> expandUserKey(const std::string key);
+        std::vector<std::string> expandUserKey(const std::string& key);
 
         /**
          * @brief Expand multiple registry keys containing user placeholders.
--- a/src/data_provider/src/extended_sources/browser_extensions/src/chrome.cpp
+++ b/src/data_provider/src/extended_sources/browser_extensions/src/chrome.cpp
@@ -8,7 +8,6 @@
  */
 
 #include "chrome.hpp"
-#include <tuple>
 #include <iostream>
 #include <fstream>
 #include <algorithm>
@@ -533,7 +532,7 @@ namespace chrome
         {
             // TODO: Improve handling this error.
             // std::cerr << "Extensions folder does not exist: " << extensionPath << std::endl;
-            return ChromeExtensionList();
+            return {};
         }
 
         std::string preferencesFilePath = Utils::joinPaths(profilePath, PREFERENCES_FILE);
@@ -543,28 +542,28 @@ namespace chrome
         {
             // TODO: Improve handling this error.
             // std::cerr << "Preferences file does not exist: " << preferencesFilePath << std::endl;
-            return ChromeExtensionList();
+            return {};
         }
 
         if (!Utils::existsRegular(securePreferencesFilePath))
         {
             // TODO: Improve handling this error.
             // std::cerr << "Preferences file does not exist: " << securePreferencesFilePath << std::endl;
-            return ChromeExtensionList();
+            return {};
         }
 
         std::string profileName = getProfileFromPreferences(preferencesFilePath, securePreferencesFilePath);
         ChromeExtensionList extensions;
 
-        for (auto subDir : Utils::enumerateDir(extensionPath))
+        for (const auto& entry : Utils::enumerateDir(extensionPath))
         {
-            subDir = Utils::joinPaths(extensionPath, subDir);
+            const std::string subDir = Utils::joinPaths(extensionPath, entry);
 
             if (!Utils::existsDir(subDir)) continue;
 
-            for (auto subSubDir : Utils::enumerateDir(subDir))
+            for (const auto& subEntry : Utils::enumerateDir(subDir))
             {
-                subSubDir = Utils::joinPaths(subDir, subSubDir);
+                std::string subSubDir = Utils::joinPaths(subDir, subEntry);
 
                 if (!Utils::existsDir(subSubDir)) continue;
 
@@ -708,9 +707,9 @@ namespace chrome
                 }
                 else
                 {
-                    for (auto subDirectory : Utils::enumerateDir(profilePath))
+                    for (const auto& entry : Utils::enumerateDir(profilePath))
                     {
-                        subDirectory = Utils::joinPaths(profilePath, subDirectory);
+                        const std::string subDirectory = Utils::joinPaths(profilePath, entry);
 
                         if (Utils::existsDir(subDirectory) && isValidChromeProfile(subDirectory))
                         {
--- a/src/data_provider/src/extended_sources/browser_extensions/src/chrome.cpp
+++ b/src/data_provider/src/extended_sources/browser_extensions/src/chrome.cpp
@@ -681,15 +681,14 @@ namespace chrome
 
 #if defined(_WIN32) || defined(_WIN64)
 
-            for (const auto& browser : WINDOWS_PATH_LIST)
+            for (const auto& [browserType, browserPath] : WINDOWS_PATH_LIST)
 #elif defined(__APPLE__) && defined(__MACH__)
 
-            for (const auto& browser : MACOS_PATH_LIST)
+            for (const auto& [browserType, browserPath] : MACOS_PATH_LIST)
 #elif defined(__linux__)
-            for (const auto& browser : LINUX_PATH_LIST)
+            for (const auto& [browserType, browserPath] : LINUX_PATH_LIST)
 #endif
             {
-                std::string browserPath = std::get<1>(browser);
                 const std::string profilePath = Utils::joinPaths(userHomePath, browserPath);
 
                 if (!Utils::existsDir(profilePath))
@@ -698,7 +697,7 @@ namespace chrome
                     continue;
                 }
 
-                m_currentBrowserType = CHROME_BROWSER_TYPES.at(std::get<0>(browser));
+                m_currentBrowserType = CHROME_BROWSER_TYPES.at(browserType);
 
                 // The profile path exists, now let's find the profile.
                 if (isValidChromeProfile(profilePath))
--- a/src/data_provider/src/extended_sources/browser_extensions/src/firefox.cpp
+++ b/src/data_provider/src/extended_sources/browser_extensions/src/firefox.cpp
@@ -70,15 +70,15 @@ FirefoxAddons FirefoxAddonsProvider::getAddons()
         return firefoxAddons;
     }
 
-    for (auto userHome : Utils::enumerateDir(homePath))
+    for (const auto& userDir : Utils::enumerateDir(homePath))
     {
         // Ignore ".", ".." and hidden directories
-        if (Utils::startsWith(userHome, "."))
+        if (Utils::startsWith(userDir, "."))
         {
             continue;
         }
 
-        userHome = Utils::joinPaths(homePath, userHome);
+        const std::string userHome = Utils::joinPaths(homePath, userDir);
 
         if (!Utils::existsDir(userHome) || !isValidPath(userHome))
         {
@@ -96,9 +96,9 @@ FirefoxAddons FirefoxAddonsProvider::getAddons()
                 continue;
             }
 
-            for (auto entity : Utils::enumerateDir(firefoxInstallationPath))
+            for (const auto& entry : Utils::enumerateDir(firefoxInstallationPath))
             {
-                entity = Utils::joinPaths(firefoxInstallationPath, entity);
+                const std::string entity = Utils::joinPaths(firefoxInstallationPath, entry);
 
                 if (!Utils::existsDir(entity) || !isValidPath(entity))
                 {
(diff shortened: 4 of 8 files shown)


Darioortegaleyva (changes requested):
Looks good overall, just one thing to check before approving: `chrome.cpp` drops `#include <tuple>`, but the file still calls `std::get<0>/<1>` on the tuple from `chrome.hpp`'s `ChromePathSuffixMap`. It compiles fine on Linux because the include reaches transitively, but the checklist has Windows and macOS unchecked, and those use a different standard library (MSVC STL / libc++) where that transitive include isn't guaranteed.

Could you confirm the build on Windows/macOS ?

Rest of the diff checked in detail (observer.hpp locking, ie_explorer.cpp reference params, chrome/firefox loop rewrites, dbsync reserve()) and it's solid, no race or leak introduced.

Darioortegaleyva (approved):
LGTM

Darioortegaleyva (changes requested):
Please add a CHANGELOG.md entry, the issue asks for one, and then I approve.

Darioortegaleyva (approved):
LGTM

MarcelKemp (approved):

Issue #37714, opened 2026-07-15
Remaining ossec/queue files after uninstallation in Windows agents


cristinavicognz:
During 2026-07-11 nightly, it was detected that uninstall test failed for Windows Server 2025 when checking the directories that remain after uninstallation, specifically `C:\Program Files (x86)\ossec-agent/queue`.
The error was reproduced after several attempts in this research https://github.com/wazuh/wazuh-qa-automation/issues/8155 on Windows Server 2025, and it was found that, following uninstallation, a process remained open, which meant that some files remained `LOCKED` and could not be deleted during uninstallation.

```
LOCKED: C:\Program Files (x86)\ossec-agent\queue\fim\db\fim.db
LOCKED: C:\Program Files (x86)\ossec-agent\queue\fim\db\fim.db-journal
LOCKED: C:\Program Files (x86)\ossec-agent\queue\fim\db\fim_sync.db
LOCKED: C:\Program Files (x86)\ossec-agent\queue\fim\db\fim_sync.db-shm
LOCKED: C:\Program Files (x86)\ossec-agent\queue\fim\db\fim_sync.db-wal
LOCKED: C:\Program Files (x86)\ossec-agent\queue\sca\db\sca.db
LOCKED: C:\Program Files (x86)\ossec-agent\queue\sca\db\sca.db-journal
LOCKED: C:\Program Files (x86)\ossec-agent\queue\sca\db\sca_sync.db
LOCKED: C:\Program Files (x86)\ossec-agent\queue\sca\db\sca_sync.db-shm
LOCKED: C:\Program Files (x86)\ossec-agent\queue\sca\db\sca_sync.db-wal
LOCKED: C:\Program Files (x86)\ossec-agent\queue\syscollector\db\syscollector_sync.db
LOCKED: C:\Program Files (x86)\ossec-agent\queue\syscollector\db\syscollector_sync.db-shm
LOCKED: C:\Program Files (x86)\ossec-agent\queue\syscollector\db\syscollector_sync.db-wal
```

The reason the process remains open is that the agent’s status is set to `stopped` before the processes have been closed due to a race condition. This issue had already been observed in the stop test, but was marked as expected as the process closed after a few seconds or minutes https://github.com/wazuh/wazuh-agent/issues/868. However, in the case of uninstall, this causes the files that should be deleted not to be deleted.

I’m including the full research here.

[...]

PR #38235, opened 2026-08-07
Close the FIM synchronization database before reporting SERVICE_STOPPED on Windows

### LINKED ISSUE #38212: Remaining queue directory after uninstall in Windows agents
|Wazuh version|Component|Commit|
|---|---|---|
| 5.0.0-rc1 | Wazuh agent | 6ba0e4012d96f916ead23e4b6fca99d6ec7296a0 | 

During 2026-08-06 Nightly, the following error appeared for all Windows deployability tests:

```
AssertionError: The C:\Program Files (x86)\ossec-agent/queue directory still exists after uninstall without purge in the following hosts:
windows_server-2019-amd64 (got: True)
windows_server-2022-amd64 (got: True)
windows_server-2016-amd64 (got: True)
windows-11-amd64 (got: True)
```

I have already opened an issue about this problem https://github.com/wazuh/wazuh/issues/37714, and it was solved here https://github.com/wazuh/wazuh/pull/37841.
This need investigation about why it is reappearing. 

In this issue is the full investigation about I found the time it appeared: https://github.com/wazuh/wazuh/issues/37714

Full logs - DIT Windows:
- [full_component_logs.zip](https://github.com/user-attachments/files/30778185/full_component_logs.zip)
- [logs.zip](https://github.com/user-attachments/files/30778189/logs.zip)

Full logs - DUT Windows
- [logs (1).zip](https://github.com/user-attachments/files/30778214/logs.1.zip)
- [full_component_logs (1).zip](https://github.com/user-attachments/files/30778221/full_component_logs.1.zip)

Full logs - DUT Windows WPK 
- [full_component_logs (2).zip](https://github.com/user-attachments/files/30778315/full_component_logs.2.zip)
- [logs (2).zip](https://github.com/user-attachments/files/30778320/logs.2.zip)

Miguevrgo:
## Description

The Windows agent tells the SCM the service stopped while a database under `queue\` is still open, and whoever asked for the stop acts on that report immediately. On uninstall the MSI is that caller: it deletes the installation as soon as the service reports stopped, the open file survives the deletion, and `queue\` is left behind.

Two databases live there: `queue\fim\db\fim_sync.db` (plus its `-wal`/`-shm`), owned by the FIM inventory synchronization handle, and `queue\fim\db\fim.db`. On Windows nothing closed the first one on the stop path: `OssecServiceCtrlHandler()` (`src/win32/win_service.c:264`) runs `stop_wmodules()` and `fim_db_teardown()`, reports `SERVICE_STOPPED` and returns, so the synchronization handle is only released later by the C++ static destructors while the process unwinds through `main()`'s `exit()`.

The teardown that closes it properly already exists, but only for POSIX. `fim_shutdown_waiter()` (`src/syscheckd/src/main.c:88-158`) aborts the in-flight synchronization with `asp_stop()`, waits for the synchronization thread, and destroys the handle under `fim_sync_handle_rwlock` so the SQLite connection is checkpointed and the `-wal`/`-shm` files removed. On Windows FIM runs as a thread inside `wazuh-agent.exe`, there is no `syscheckd` `main()`, and the whole machinery is compiled out (`src/syscheckd/src/run_check.c:62-77`, `#ifndef WIN32`).

Then the MSI proceeds on the stopped report:

```
MSI (s) Executing op: ServiceControl(,Name=WazuhSvc,Action=2,Wait=1,)
MSI (s) Executing op: ActionStart(Name=RemoveFiles,...)
MSI (s) Executing op: ActionStart(Name=RemoveFolders,...)
```

and `CustomAction_RemoveAllScript` (`src/win32/RemoveAllScript.vbs:104-111`) deletes each subfolder with `f.Delete True` under `On Error Resume Next`, so failing to delete an open file is ignored and the uninstall still returns 0.

This is also why the fix for #37714 did not end it. #37841 made the SCA, Syscollector and Agent-info `stop()` calls close their own databases synchronously, which covers ten of the thirteen files that report listed as locked:

```
fim.db, fim.db-journal, fim_sync.db, fim_sync.db-shm, fim_sync.db-wal      <- FIM
sca.db, sca.db-journal, sca_sync.db, sca_sync.db-shm, sca_sync.db-wal      <- closed by #37841
syscollector_sync.db, syscollector_sync.db-shm, syscollector_sync.db-wal   <- closed by #37841
```

What was left is the FIM files, and they are enough to keep `queue\` alive. The nightly fails on the hosts where the remaining teardown does not fit in the second the MSI leaves between the stop and `RemoveFiles`, which is why 2016/2019/2022 failed while 10/11/2012r2/2025 passed in the same 2026-08-06 run.

Closes #38212
Related: #37714, #37841, #37334

## Proposed Changes

Six commits, each reviewable on its own:

| Commit | What it does |
|---|---|
| `738b8f18` | Static analysis cleanup in the two `db` files (detailed below, unrelated to the bug) |
| `e337db71` | Closes `fim_sync.db` on the Windows stop path, through a `fim_db_teardown()` hook |
| `ad971c0d` | Reports the stop to the SCM with a wait hint that matches what the handler now does |
| `fc35cc43` | Prunes the file and registry scans on shutdown, so the teardown can reach both databases |
| `9673b0fb` | Keeps the stop path inside the deadline the SCM gives the handler, and stops the callers of the scan mutexes from blocking on them |
| `b5bd8ba1` | Releases the mutexes an agent-info pause holds, and makes `-1` mean only "stop in progress" or a real failure |

### Closing `fim_sync.db`, and when the hook is registered

`fim_sync_teardown()` (`src/syscheckd/src/run_check.c`) does on Windows what `fim_shutdown_waiter()` does on POSIX: `asp_stop()` first, a bounded wait for the synchronization thread, then `asp_destroy()` under the `fim_sync_handle_rwlock` write lock. If the thread does not exit in time the handle is left alone, because destroying it under a live user is a use-after-free.

The part that decides whether the fix covers the reported case is **where the hook is registered**. `asp_create()` opens `fim_sync.db` in `fim_initialize()` (`src/syscheckd/src/syscheck.c`), well before `start_daemon()` and its baseline `fim_scan()`. Registering the hook after that scan leaves the entire scan unprotected, which is exactly the install then uninstall window that reopened this issue, so the registration happens as soon as the handle exists.

The rest follows the POSIX protocol, and each piece answers a specific race:

- `asp_stop()` runs under the handle **read** lock, before the write lock. Taking the write lock first would stall behind the synchronization thread's own read lock, waiting for the very cycle `asp_stop()` is there to abort. The thread identifier the self-check reads is read there too, under the lock its writer holds.
- The shutdown flags and the thread handle are published under the **write** lock that `start_daemon()` publishes the creation under, so either the thread is visible here and waited for, or the creation observes the shutdown and never launches it.
- The thread identifier is cleared together with the handle. Windows recycles identifiers, and a stale one would make the "am I the synchronization thread" guard misfire on an unrelated thread.
- That guard returns without touching anything, since `merror_exit()` reaches this through `WinSetError()` from whichever thread failed, and the synchronization thread can neither wait for nor destroy what it is using.

`fim_db_set_teardown_hook()` takes a `bool` returning function and `fim_db_teardown()` only runs `DB::teardown()` when the hook agrees. That mirrors `fim_shutdown_waiter()`, which skips the inventory database while a scan is in flight because `fim_db_transaction_*` is not covered by FIMDB's internal guards, and the hook is the only side that knows. The pointer is `std::atomic` because the FIM thread writes it and the SCM thread reads it.

### Pruning the scans, so the teardown can reach `fim.db`

`fim_file_scan()` holds `fim_scan_mutex` for its whole walk, `fim_registry_scan()` does the same with `fim_registry_scan_mutex`, and neither looked at the shutdown flag. Without this the guard above degrades into "never close `fim.db` if the stop lands during a scan", and since `fim.db` lives inside `queue\`, the directory would still be left behind in precisely the reported case.

Both walks return immediately once the shutdown starts, checked **inside** the recursion (`fim_checker()`, `fim_open_key()`) rather than only in the loop over configured paths, because a single top level path takes minutes on its own. Three consequences handled with it:

- The aborted transaction is closed with a new `fim_db_transaction_close()` instead of `fim_db_transaction_deleted_rows()`. The latter is `dbsync_get_deleted_rows()` plus `dbsync_close_txn()`, so running it after a pruned walk would report every path the scan never reached as deleted. The context still has to be released, hence the new call, which only closes.
- The scans' trailing per-row writes (`cleanup_failed_fim_files()`, `process_pending_sync_updates()`, and the registry equivalents) are skipped. They run inside the scan mutex and were what kept it held after the transaction was already closed; the next scan recomputes them.
- The registry transaction lifecycle moved inside `fim_registry_scan_mutex`. It used to start before the lock and close after the unlock, so it sat outside the window the teardown proves empty, which is what the guard exists to prevent. The registry DataClean also takes the handle read lock now, like the file path does for the same `asp_notify_data_clean()` call: this PR is what first makes `asp_destroy()` reachable on Windows, so that window stops being theoretical.

`fim_scan()` checks the shutdown between the file and registry scans, and skips the post-scan database state counts, which a released context would answer with zero and record as an empty database.

### Staying inside the deadline the SCM gives the handler

A control handler has to return within 30 seconds or the SCM fails the request: `ControlService()` gives `ERROR_SERVICE_REQUEST_TIMEOUT`, `os_stop_service()` reports failure, and the MSI moves on to `RemoveFiles` with the process still closing, which is the failure this PR is about. `dwWaitHint` does not extend that deadline, it only describes the transition.

So the teardown is budgeted, in `syscheck.h`:

```c
#define FIM_SYNC_EXIT_TIMEOUT_WIN_MS 7000
#define FIM_SCAN_MUTEX_WAIT_MS       3000
#define FIM_TEARDOWN_BUDGET_MS       (FIM_SYNC_EXIT_TIMEOUT_WIN_MS + (2 * FIM_SCAN_MUTEX_WAIT_MS))
```

13 s worst case for the teardown, leaving the rest of the 30 s for `wm_kill_children()` and `stop_wmodules()`. The wait hint is derived from that budget rather than being a literal, and goes back to zero when the service reports itself stopped. POSIX keeps its own 20 s bound in `FIM_SYNC_EXIT_TIMEOUT_MS`, since it has no handler deadline to respect; the two values are deliberately different and each says why.

### Nobody may block on a scan mutex the stop is waiting for

The teardown takes both scan mutexes to prove no scan is in flight, and then **releases them**: this process outlives the `STOPPED` report, so a thread blocking on a mutex nobody releases would be parked for good. What keeps a new transaction away is `is_fim_shutdown`, now `volatile`, which every holder re-checks with the mutex held:

- **The scans** re-check it after acquiring, before starting a transaction.
- **The integrity thread** (`fim_run_integrity()`) acquires all three of `fim_scan_mutex`, `fim_realtime_mutex` and, on Windows, `fim_registry_scan_mutex` with a stop-aware poll instead of `w_mutex_lock()`, releasing what it already holds on the way out. A scan holds those for minutes, far longer than the wait the teardown can afford.
- **`asp_reset()`** is not called during shutdown. It clears the stop flag `asp_stop()` sets, so the scheduled scan loop could un-abort the synchronization the bounded wait is waiting on. The statement right after it was already guarded.
- **An agent-info pause** acquires the same three mutexes and holds them until `fim_execute_resume()` arrives, which never happens once `stop_wmodules()` has killed agent-info. `fim_syscom_cleanup_pause()` releases them from the stop path, and `fim_execute_is_pause_completed()` re-checks the shutdown with the mutexes held, so a pause granted while the stop is landing releases them instead of publishing itself.

The shutdown guards in `syscom.c` return `-1` rather than `0`, so `syscom_dispatch()` reports an error instead of a success that never happened: a refused pause used to answer "completed", which is exactly the answer that makes agent-info write to the database it was trying to pause.

### Static analysis cleanup in the two `db` files

These came out of a clang-tidy pass over the files the fix touches, and are unrelated to the bug:

- **Include what you use.** `db.hpp` no longer pulls the C header `db.h` (and `<string.h>`) to reach declarations it uses; it includes `commonDefs.h`, `logging_helper.h` and the standard headers for the types in its own signatures. `db.cpp` does the same for what it uses directly instead of relying on transitive includes, and drops `stringHelper.h`.
- **`DB` members that never touch instance state are now `static`.** They all forward to `FIMDB::instance()`, so the `DB` singleton dereference at every call site bought nothing; the call sites become `DB::teardown()` instead of `DB::instance().teardown()`. The compiler enforces the claim: any use of a non-static member would not build.
- **The deleted copy constructor and copy assignment moved to `public`.** In `private`, a copy attempt was diagnosed as "is private" rather than "use of deleted function", pointing the reader at the wrong problem; access control is checked after overload resolution, so keeping them private added nothing.
- **`[[nodiscard]]` on `no_entry_found::what()`** replaces `__attribute__((__returns_nonnull__))`. They are not equivalent: the attribute was a non-standard promise to the optimizer, while `[[nodiscard]]` is standard and warns when the returned message is thrown away, which is the contract that matters for `what()`.
- **`const auto*` instead of `auto`** for the database path in `fim_db_init()`.
- **Layout** follows `ci/input/astyle.config`, the convention the `rtr` check enforces over `syscheckd/src/db/src`.

### Results and Evidence

Windows Server 2022 (Server Core) under QEMU/KVM, no network. Binaries are built from this PR's base commit, differing only in the change under test, and swapped over one installed MSI so the uninstall path is the real one.

**The window under test.** The nightly installs the agent and uninstalls it about twenty seconds later, so the stop lands while the baseline scan is running. Every run below is stopped 20 s after `Start-Service`, with FIM pointed at `C:\Windows\WinSxS`, and every run confirms the state it is meant to exercise (`scan_started=True`, `scan_ended=False`). This matters: an earlier measurement of this PR used a 150 s warm-up, which sits *after* the baseline scan and after the old registration point, and therefore could not see the defect the review found.

**Open handles when the SCM reports `SERVICE_STOPPED`**, tested by opening each database file with `FileShare.None`, which is the same exclusive access the uninstall needs to delete it:

| build | open at `STOPPED` | `Stop-Service` |
|---|---|---|
| this PR before these two commits | `fim_sync.db`, `-wal`, `-shm` in 3 of 5 runs (one also `fim.db`) | ok, 1.1-2.3 s |
| commit 1 only | `fim.db` in 2 of 2 runs, synchronization files closed | ok, 1.2 s |
| commit 1 + commit 2 | **nothing, 0 of 3 runs** | ok, 1.3-2.2 s |

Commit 1 alone closes the synchronization files in every run and says why it leaves the other one:

```
run_check.c:196 at fim_sync_teardown(): DEBUG: Skipping the FIM database teardown: a scan is in progress.
```

With commit 2 both scans prune themselves and the teardown completes:

```
file.c:1350 at fim_file_scan(): DEBUG: Shutdown in progress: closing the scan transaction without the deleted rows stage.
registry.c:2126 at fim_registry_scan(): DEBUG: Shutdown in progress: closing the registry scan transactions without the deleted rows stage.
```

Both refinements in commit 2 came out of measurement rather than design. A first version checked the shutdown only in the registry loop over top level keys and kept the scan's trailing writes, and still left `fim.db` open in 4 of 5 runs, with the reason logged three seconds after the transaction had already been closed:

```
run_check.c:144 at fim_wait_for_scan_mutex(): DEBUG: Skipping the FIM database teardown: the file scan is still in progress
```

That is the trailing per-row work holding `fim_scan_mutex`. With it skipped, the stop also got faster, from 5-6 s down to 1.3-2.2 s.

**When the table was measured.** It was taken on the first two commits of the list above, before the review rounds that produced the rest. Those later rounds change who may block and what a refused command answers, and they shorten the Windows wait from 20 s to 7 s, but they do not change the teardown sequence itself: `asp_stop()`, the bounded wait, `asp_destroy()`, then `DB::teardown()` once both scan mutexes are free. What is open when the SCM is told the service stopped is therefore the same, and the two builds in the table are still the before and after of that.

**What these runs do not prove.** `queue\` was removed in all 20 iterations, including with the unpatched build: in this bed the process exits before `msiexec` reaches `RemoveFiles`, so the leftover directory does not reproduce and the reliable discriminator is the handle state at `STOPPED`, not the surviving directory. An earlier bed, warmed up for 150 s with a ~97 MB database, did reproduce the leftover directory in 2 of 3 control runs, so whether the symptom shows depends on how long the process stays alive after `STOPPED`.

**Rejected approaches, kept because the reasoning is the reviewable part.** Three variants made the process die instead of closing the databases: `exit()` before reporting `SERVICE_STOPPED`, `TerminateProcess()` after reporting it, and `_exit(0)` right after reporting it. All three remove the leftover `queue\`, and all three break the service contract, because the SCM completes the caller's `ControlService()` request when the handler *returns*:

```
Stop-Service: Service 'Wazuh (wazuhsvc)' cannot be stopped due to the following error:
Cannot stop 'wazuhsvc' service on computer '.'
```

Reproduced 8/8 for the first two and 2/2 for the `_exit(0)` variant, and caught by this repo's `smoke-test-windows` job. It matters beyond the smoke test, since `os_stop_service()` and `UninstallService()` both check what `ControlService()` returns. `ExitProcess()` additionally crashed the agent on every stop with an access violation inside `ReadElem()` (`os_xml.c`), because it terminates the other threads and then runs `DLL_PROCESS_DETACH` on the calling one.

**Test bed notes.** Three test-only bypasses are applied identically to every build compared above, so they cancel out: the document-limits wait in `fim_initialize()` is short-circuited, `client.keys` is seeded (otherwise the agent loops in enrollment and never reaches `fim_initialize()`), and `agent.remote_conf=0` releases the startup hash gate, which on Windows is queried in-process. None of them is part of this PR.

  - [x] Linux
  - [x] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

The `winagent` unit tests were also run locally under wine with the CI's environment (`WINEARCH=win32` and the `WINEPATH` from `legacy_unit_tests.sh`): 8 of 8 suites covering `syscheckd` pass, 0 failures.

### Artifacts Affected

Wazuh agent for Windows: `wazuh-agent.exe`, `wazuh-agent-eventchannel.exe`, `libfimdb.dll`.

### Configuration Changes

N/A

### Tests Introduced

`TestTeardownHookRefusalKeepsTheDatabaseUp` in `src/syscheckd/src/db/tests/db/ComponentTest/dbInterface/dbTest.cpp`, covering the new gate: a hook that refuses runs and leaves the database usable instead of tearing it down.

Existing tests were also corrected: `test_registry.c` and `test_fim_scan.c` passed an uninitialized `TXN_HANDLE` by value to `will_return()`, which is undefined behaviour and became visible against the new NULL check on the transaction handles; they pass `&mock_handle` now, the form the rest of the file already used.

## Review Checklist

- [ ] Code changes reviewed


Darioortegaleyva (approved):
LGTM. 

One non-blocking notes: the "ten of thirteen files" count in the description doesn't match its own table (should be eight)

Darioortegaleyva (changes requested):
## Blocking

`fim_registry_scan()` never got the treatment `fim_file_scan()` got in this same PR.

`fim_scan()` (`fim_scan.c:71-83`) calls `fim_file_scan()` and then, with no shutdown check between them, `fim_registry_scan()`. If the stop lands during that first file scan, it is possible for `fim_sync_teardown()` to win the race for `fim_registry_scan_mutex` (held forever on success, by design) before this same thread reaches `registry.c:2116`, whose `w_mutex_lock()` is blocking. That hangs the FIM thread permanently, on the exact mid-scan-stop timing this PR targets.

Three more gaps in the same function, all already handled in `file.c` for comparison:

- `registry.c:2110-2113` / `2149-2157`: the transactions start before the mutex lock (`2116`) and close after the unlock (`2145`), so they are not covered by the guarantee `fim_sync_teardown()` relies on. `file.c:1298-1372` keeps the whole transaction lifecycle inside the mutex; this function does not.
- `registry.c:2047-2060`: reads `syscheck.sync_handle` and calls `asp_notify_data_clean()` with no `fim_sync_handle_rwlock` held (compare `file.c:1250-1261`, which takes the read lock around the same call). This PR is what first makes `asp_destroy()` reachable on Windows, so this is now a live use-after-free window, not a theoretical one.
- `registry.c:2162-2173`: the trailing per-row writes run unconditionally, with no `fim_shutdown_process_on()` guard (compare `file.c:1357-1370`), and after the mutex is already released.

## Should also fix before merge

- `FIM_SYNC_EXIT_TIMEOUT_MS` is `5000` in `run_check.c:83` but `20000` in `main.c:34` (same macro name, different value). The comment at `run_check.c:193` says "Bounded like POSIX", but it is a quarter of the POSIX bound. A sync-thread exit taking 6-19s succeeds on POSIX and fails on Windows, skipping the very teardown this PR adds.
- `run_check.c:163` reads `fim_sync_thread_id` before taking `fim_sync_handle_rwlock`, racing `start_daemon()`'s write under that lock.

Darioortegaleyva (changes requested):
## Blocking

- **Pause can hold the FIM mutexes forever after shutdown.** `fim_execute_is_pause_completed()` (`syscom.c:62-65`) returns before the shutdown check at line 72 once a pause is granted. The only release is `fim_execute_resume()` via agent-info, which `stop_wmodules()` kills. If a pause is active when the stop lands, the teardown times out deterministically.
- **The shutdown guards in `syscom.c` return the same value as success.** `get_version` (`:168`), `set_version` (`:187`), `is_pause_completed` (`:74`) all return `0`, which `syscom_dispatch` reads as a real success (`:440-450`, `:461-467`, `:491-497`). Needs a distinct sentinel.
- **`fim_run_integrity()` still blocks unbounded on 2 of 3 mutexes.** `run_check.c:1314` and `:1316` (`fim_realtime_mutex`, `fim_registry_scan_mutex`) are plain `w_mutex_lock()`; the shutdown check (`:1320-1322`) only runs after both are already held.
- **3 unit tests pass an uninitialized handle.** `test_registry.c:907,913-914` / `1060,1062-1063` / `1094,1096-1097` use `mock_handle` instead of `&mock_handle` (see `:975-982` for the correct form) - UB, and now flaky against the new `NULL` check at `registry.c:2163`.

Darioortegaleyva (approved):
LGTM

MarcelKemp (approved):

PR #38254, opened 2026-08-07
AWS integration tests: isolate concurrent runs on the shared bucket with a per-run S3 namespace (GITHUB_RUN_ID) [4.14.8]

### LINKED ISSUE #38194: AWS integration tests: isolate concurrent runs on the shared bucket with a per-run key namespace (GITHUB_RUN_ID)
## Description

The AWS integration tests run against a shared S3 bucket (introduced in IDR 4984) and the test data uses
**fixed date-based prefixes** (e.g. `test_prefix/AWSLogs/.../2022/11/20/...`). Two AWS IT runs executing
at the same time therefore upload to the **same keys/prefixes** and interfere:

- The setup guard (`_assert_prefix_clean`) of one run flags the other run's freshly-uploaded object as an
  unexpected collision (`SETUP COLLISION`).
- Count mismatches: a run's module can read objects uploaded by another run.
- Teardown / the `always()` cleanup of one run can delete another run's in-flight objects, leaving orphans.

This is a pre-existing limitation of the shared-bucket model (came with IDR 4984); it is **not** caused by
any individual PR. #38101 made per-run cleanup cancellation-safe and fixed the guard scoping, but the
concurrency vector remains. It was observed live: two AWS IT runs (5.0.0 and 4.14.8) overlapping in time
collided on the same `test_prefix/AWSLogs/.../elasticloadbalancing/2022/11/20/` prefix.

## Proposed change - per-run key namespace

Namespace every object the tests upload under a **per-run unique folder**, keyed by `GITHUB_RUN_ID`, so
concurrent runs never share keys:

```
s3://<bucket>/<GITHUB_RUN_ID>/test_prefix/...
s3://<bucket>/<GITHUB_RUN_ID>/AWSLogs/...
```

- Prefix all uploaded keys with the run id.
- Make the module's `ossec.conf` `path` include that run-id prefix so each run reads only its own objects.
- Teardown and the `always()` cleanup then scope to `<GITHUB_RUN_ID>/`; orphans are isolated per run and
  trivially reapable.

Points to handle:
- **Permanent seeds** (CloudTrail / ELB) live at the top-level `AWSLogs/`, and `check_bucket()` reads the
  configured path. Decide how the seeds coexist with the per-run prefix (e.g. keep seed-dependent checks
  on the shared root, or provide per-run seeds / adjust `check_bucket`).
- Not per-run **buckets** (contradicts the shared-bucket design from IDR 4984 and can hit account limits).

Cheaper stopgap if a full fix is not needed yet: a workflow `concurrency` group so AWS IT runs serialize
(one at a time), removing the collision at the cost of throughput.

## Additional items surfaced during the #38101 review

**A. Cleanup credential window (silent no-op).**
The `always()` cleanup deletes with `--profile default`, written from OIDC credentials with
`role-duration-seconds: 7200` (2h). If a job runs longer than 2h the credentials expire, every
`aws s3 rm` fails, and with `|| true` the step still reports success - the cleanup silently no-ops
exactly when it is needed. The 5.x job is bounded by `timeout-minutes: 120`; the 4.14.8 job had no job
timeout (GitHub default 6h) and the measured full run was ~1h29m (~30 min margin).
Fix (small): `timeout-minutes: 120` on the 4.14.8 job + emit a `::warning::` on any cleanup delete
failure (both workflows). Optionally re-assume credentials right before the cleanup for very long runs.

**B. Inspector service-path coverage.**
`test_inspector_discard_regex` is the only test exercising the argument quoting on the module's
**service** code path (`src/wazuh_modules/src/wm_aws.c` ~612-652). It currently runs and passes in the
`aws_inspector` shard, so the coverage exists today. Residual risk: if the Inspector role/credentials
become unavailable and that shard skips those tests, the coverage disappears silently - worth asserting
the inspector cases actually run (fail if 0 selected).

## Tasks

- [ ] Namespace uploaded keys under `<GITHUB_RUN_ID>/` and thread the prefix into the module `path`.
- [ ] Reconcile permanent seeds / `check_bucket` with the per-run prefix.
- [ ] (or, as stopgap) add a workflow `concurrency` group to serialize AWS IT runs.
- [ ] `timeout-minutes: 120` on the 4.14.8 AWS IT job + `::warning::` on cleanup delete failure (both workflows).
- [ ] Assert the inspector cases are actually collected/run.

## Related

- IDR 4984 - shared-bucket migration
- #38101 - orphan cleanup + guard scoping (this follow-up covers the concurrency case)

Darioortegaleyva:
## Description

> **Note:** the AWS integration tests are a no-op on `pull_request` for 4.x (the paths filter only runs them when `wodles/aws/**` changes), so the checks on this PR do not exercise them. They were validated with a manual `workflow_dispatch` on this branch's head: **https://github.com/wazuh/wazuh/actions/runs/32106846554** (183 passed, 8 skipped, 1 xfailed, 0 failed).

The AWS integration tests run against a single shared S3 bucket (plus a shared VPC and a shared SQS queue), so two runs executing at the same time interfere. This makes the AWS integration tests safe to run in parallel by isolating each run across the three resources that concurrent runs otherwise collide on:

- **S3 keys**: every key and the module's configured path go under a per-run namespace `s3://<bucket>/<GITHUB_RUN_ID>/...`, so runs never share keys (`SETUP COLLISION`, cross counts, one run's teardown deleting another's objects).
- **VPC Flow Log**: its dedup key `(VPC, TrafficType, destination)` is shared, so a second concurrent run hits `FlowLogAlreadyExists`. The run now reuses the existing flow log and only its creator deletes it.
- **SQS notification**: the bucket event notification is a single shared object. Each run registers its own prefix-filtered rule (`<run>/`) merged with the others (read-modify-write) instead of replacing the whole config, so concurrent runs are never notified of each other's uploads.

Runs keep executing in **parallel** (no serialization).

**Proposed Release:** v4.14.8
**Issue:** wazuh/wazuh#38194
**Internal Reference:** N/A

## Proposed Changes

- `tests/integration/test_aws/conftest.py`:
  - **S3 namespace**: `_namespaced()` prefixes the bucket path with `<GITHUB_RUN_ID>/` (no-op locally). Uploaded keys and the module's `<path>` stay aligned; a path-less bucket type gets an explicit `<path>` so it reads under the namespace. Session fixture `aws_run_namespace` mirrors the permanent seeds into `<run>/AWSLogs/...` at start and deletes the whole `<run>/` namespace at the end; the root seeds are never modified. `_is_permanent_seed()` recognizes both the root seeds and their per-run copies.
  - **VPC Flow Log**: on `FlowLogAlreadyExists`, reuse the existing flow log (`describe_flow_logs`); only the run that created it deletes it in teardown (`flow_log_owned`). The flow-log id is only used to name the synthetic object, so sharing it is harmless.
  - **SQS**: `_set_run_bucket_notification` registers a per-run prefix-filtered rule and `_merge_queue_configs` rebuilds the config preserving other runs' rules, dropping our own and any rule whose prefix overlaps ours (a legacy no-filter rule; S3 rejects overlapping prefixes for the same event). Read-modify-write with a verify+retry loop. Teardown removes only this run's rule. The queue name is already unique per run (framework session suffix).
  - Test adjustments for the namespaced `<path>` in `test_basic.py`, `test_regions.py`, `test_only_logs_after.py`, and the empty-suffix warning in `test_path_suffix.py`.
- `.github/workflows/4_testintegration_aws-tier-0-1.yml`: pass `GITHUB_RUN_ID` through each `sudo` pytest invocation so the namespace activates in CI (sudo resets the environment). The existing manifest-based `if: always()` cleanup is unchanged.

## Dependencies

The OIDC role used by the AWS integration tests needs `s3:GetBucketNotification` on the shared bucket (it already had `PutBucketNotification`) for the SQS notification merge to read the current config. Granted by the DevOps team.

### Results and Evidence

**Validated under real concurrency**: this run and the 5.0.0 run (#38253) overlapped in time and both passed, so the S3 namespace, the VPC Flow Log reuse and the per-run SQS notification merge all hold with two runs live at once.

- 4.14.8 run (manual `workflow_dispatch`): https://github.com/wazuh/wazuh/actions/runs/32106846554 (183 passed, 8 skipped, 1 xfailed, 0 failed; includes both `custom_bucket` SQS cases and `test_regions[vpc_regions_with_data]`).

Also validated locally with `moto` (mocked S3/EC2/SQS) against the real module code and the harness: per-run namespace + cross-run isolation, VPC create-or-reuse (owner-only delete, no orphan), and the SQS notification merge (two runs' prefix-filtered rules coexist, teardown of one preserves the other, worst-case interleave converges, legacy overlapping rule dropped).

### Artifacts Affected

- `tests/integration/test_aws/conftest.py`
- `tests/integration/test_aws/test_basic.py`, `test_regions.py`, `test_only_logs_after.py`, `test_path_suffix.py`
- `.github/workflows/4_testintegration_aws-tier-0-1.yml`

### Configuration Changes

N/A

### Documentation Updates

N/A

### Tests Introduced

- **Scope:** Integration Tests (infrastructure).
- **Details:** No new test cases. Per-run isolation (S3 namespace + VPC flow log reuse + SQS notification merge) so concurrent AWS IT runs no longer collide; validated locally with `moto` and in CI under real concurrency.

## Review Checklist

> No compiled code changes (Python + workflow only) - compilation and memory checks are N/A.

- [ ] Code changes reviewed

### DIFF
--- a/.github/workflows/4_testintegration_aws-tier-0-1.yml
+++ b/.github/workflows/4_testintegration_aws-tier-0-1.yml
@@ -184,7 +184,7 @@ jobs:
         if: steps.filter.outputs.parser == 'true'
         run: |
           cd tests/integration
-          sudo AWS_BUCKET_NAME="${AWS_BUCKET_NAME}" AWS_IT_CLEANUP_MANIFEST="${AWS_IT_CLEANUP_MANIFEST}" python3 -m pytest -vvv --tier 0 --tier 1 test_aws/test_parser.py
+          sudo GITHUB_RUN_ID="${GITHUB_RUN_ID}" AWS_BUCKET_NAME="${AWS_BUCKET_NAME}" AWS_IT_CLEANUP_MANIFEST="${AWS_IT_CLEANUP_MANIFEST}" python3 -m pytest -vvv --tier 0 --tier 1 test_aws/test_parser.py
 
       - name: Run every test due to base WazuhIntegration class change or manual dispatch
         if: steps.filter.outputs.integration == 'true' || github.event_name == 'workflow_dispatch'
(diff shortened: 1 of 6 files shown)


Darioortegaleyva:
### Manual AWS integration-test run (green)

The AWS integration tests are skipped on `pull_request` for 4.x (the paths filter only runs them when `wodles/aws/**` changes), so they are validated here with a manual `workflow_dispatch` on the current head.

- Commit: `42afcc4a46`
- Run: https://github.com/wazuh/wazuh/actions/runs/32106846554
- Result: **183 passed, 8 skipped, 1 xfailed, 0 failed** (+ inspector/custom-bucket shard: 6 passed).

Full concurrent-run isolation was exercised: this run overlapped in time with the 5.0.0 run (#38253) and both passed, so the per-run S3 namespace, the VPC Flow Log reuse, and the per-run SQS notification merge all hold under real concurrency.

Darioortegaleyva:
### IDR 4984 compliance (not regressed)

This change does not touch or break anything required by IDR-4984 (AWS IT migration to OIDC + shared bucket + pre-existing VPC):

- **OIDC / no static keys**: the workflow diff only adds `GITHUB_RUN_ID` to the `sudo` pytest invocations. It does not touch the OIDC role, `role-to-assume`, credentials, or the account, and introduces no static AWS keys.
- **Pre-existing VPC**: the VPC tests keep using the pre-existing VPC from the `AWS_VPC_ID` secret (no VPC is created or deleted). The flow-log change only *reuses* an existing flow log on `FlowLogAlreadyExists`, so it reduces resource churn rather than adding any.
- **Shared bucket**: preserved (a single shared bucket, as IDR 4984 mandates); isolation is by per-run key prefix, not per-run buckets.

**Evidence** (green on this PR's exact head `42afcc4a46`, via manual `workflow_dispatch` since the AWS IT is a no-op on `pull_request` for 4.x):
- Run: https://github.com/wazuh/wazuh/actions/runs/32106846554 (183 passed, 8 skipped, 1 xfailed, 0 failed; includes both `custom_bucket` SQS cases and `test_regions[vpc_regions_with_data]`).
- This run overlapped in time with the 5.0.0 run (#38253) and both passed, so the per-run S3 namespace, the VPC flow-log reuse and the SQS notification merge all hold under real concurrency.

Darioortegaleyva:
### Manual AWS integration-test run (evidence)

The AWS integration tests are a no-op on `pull_request` for 4.x (the paths filter only runs them when `wodles/aws/**` changes), so the checks on this PR do not exercise them. They are validated with a manual `workflow_dispatch` on this PR's head `f8b0ad9a9b`:

- Run: https://github.com/wazuh/wazuh/actions/runs/32341183057

Expected result (an identical run on this same head already passed): **183 passed, 8 skipped, 1 xfailed, 0 failed** (+ inspector/custom-bucket shard: 6 passed), covering the per-run S3 namespace, the VPC Flow Log reuse, the SQS notification isolation and the `path_suffix` cases.

Miguevrgo (changes requested):
# Code review: PR #38254

`fix/38194-aws-it-concurrent-runs-4.14.8` → `4.14.8`

Scope reviewed: every hunk against the surrounding code, the `qa-integration-framework`
4.14.8 helpers (`generate_file`, `get_last_file_key`, `get_data_generator`),
`wodles/aws/buckets_s3/aws_bucket.py`, `wodles/aws/aws_tools.py` and
`src/wazuh_modules/wm_aws.c`.

## Blocking: permanent damage to shared AWS state

### 1. `tests/integration/test_aws/conftest.py:164` — falsy `{}` drops `EventBridgeConfiguration`

`for k in (...): if cfg.get(k)` copies forward only truthy sub-configs.
`GetBucketNotificationConfiguration` returns `EventBridgeConfiguration: {}` (the shape
has no members) when EventBridge is enabled, and `{}` is falsy, so the key is omitted
from the read-modify-write PUT and EventBridge notifications are **permanently
disabled** on the shared IT bucket from the first `_put_notification` onward.

Fix: `if k in cfg`.

### 2. `tests/integration/test_aws/conftest.py:773` — reused flow log leaks forever

`flow_log_owned` makes the leak of a hard-cancelled run permanent and invisible.
Previously a leaked flow log made the next run fail loudly with `FlowLogAlreadyExists`,
forcing cleanup; now it is silently reused and never deleted by anyone. The live flow
log keeps delivering REJECT logs to `s3://<bucket>/AWSLogs/<acct>/vpcflowlogs/…` **at
the bucket root**, which neither `_delete_run_namespace` nor the new
`aws s3 rm .../$GITHUB_RUN_ID/` step will ever remove. No reaper exists.

### 3. `tests/integration/test_aws/conftest.py:151` — queue-rule merge ignores `Events`

`_prefixes_overlap(_rule_prefix(q), own_prefix)` is applied without comparing the
rule's `Events`. S3 only forbids overlapping prefixes *for the same event type*, but the
merge unconditionally deletes any queue rule with an overlapping (or absent) prefix
filter. A legitimate rule on `s3:ObjectRemoved:*` with no prefix filter is silently and
permanently removed on the first custom-bucket test, and nothing restores it.

## Correctness



### 8. `test_only_logs_after.py:879` — `own_keys` override lacks folder/type filter

No folder-marker filter and no bucket-type scoping, unlike its sibling
`_assert_prefix_clean` (which explicitly skips `obj.key.endswith('/')`). Any zero-byte
folder marker under `<run>/`, or a leftover non-`AWSLogs/` object from another bucket
type, becomes `max(own_keys)` and replaces the correct marker — including for
`AWSLogs`-based types (vpcflow, cloudtrail, config, alb/clb/nlb) whose
`get_last_file_key` result was already right.

Fix: filter out `endswith('/')` and restrict the listing to the type's own base path.

## Diagnosability / hygiene

### 10. `conftest.py:219` — notification setup warns instead of failing

`_set_run_bucket_notification` only logs a warning when `_put_notification` returns
`False`. If the OIDC role is missing `s3:GetBucketNotification` (a new dependency this
PR introduces), all 5 attempts fail, no rule is registered, and
`test_custom_bucket_logs` sleeps 30s then fails on `failed_sqs_message_retrieval` — a
misleading symptom for a permissions problem. Raise or `pytest.fail`.

### 11. `.github/workflows/4_testintegration_aws-tier-0-1.yml:34` — dead cleanup manifest

`AWS_IT_CLEANUP_MANIFEST` is still exported and `_record_uploaded_key` /
`record_uploaded_key` still write to it, but nothing consumes it and the
`sudo rm -f "$MANIFEST"` was dropped. Cleanup now depends entirely on every upload being
namespaced; the moment a test uploads outside `<GITHUB_RUN_ID>/` (the flow-log objects in
finding 2, or a future test bypassing `create_test_bucket`) the orphan is undetectable,
whereas the manifest used to catch it.

Fix: wire the manifest back in as a second pass, or remove the env var and the recording
code so the contract is explicit.

### 12. Cleanup some comments which are really unnecessary

Darioortegaleyva:
### Manual AWS integration-test run (evidence)

The AWS integration tests are a no-op on `pull_request` for 4.x (the paths filter only runs them when `wodles/aws/**` changes), so the checks on this PR do not exercise them. They are validated with a manual `workflow_dispatch` on this PR's current head `3b0c7c1ed2` (includes the review fixes):

- Run: https://github.com/wazuh/wazuh/actions/runs/32473454830
- Result: **183 passed, 8 skipped, 1 xfailed, 0 failed** (+ inspector/custom-bucket shard: 6 passed), covering the per-run S3 namespace, VPC Flow Log reuse, SQS notification isolation and the `path_suffix` cases.

This ran overlapping in time with the 5.0.0 run (#38253), which also passed, so the isolation holds under real concurrency.

Miguevrgo (dismissed):
Point 12 was ignored, I see too many comments explaining what can be read directly but there are no stopping points so LGTM!

Darioortegaleyva:
### Manual AWS integration-test run (evidence)

The AWS integration tests are a no-op on `pull_request` for 4.x (the paths filter only runs them when `wodles/aws/**` changes), so the checks on this PR do not exercise them. Validated with a manual `workflow_dispatch` on this PR's latest head `aa8613e5b9`:

- Run: https://github.com/wazuh/wazuh/actions/runs/32833846868
- Result: **183 passed, 8 skipped, 1 xfailed, 0 failed** (+ custom-bucket shard: 6 passed).

Covers the per-run S3 namespace, per-run VPC flow log delivery (`<run_id>/_flowlogs/`, no orphans left), SQS notification isolation, and the `path_suffix` cases.

MarcelKemp (approved):

PR #38267, opened 2026-08-07
Reject symlinks when writing files staged under the incoming directory

### LINKED ISSUE #38200: Agent upgrade-staging socket follows symlinks
|Component|Install type|
|---|---|
Agent Update | Agent 

## Description

`_jailfile()` (`src/wazuh_modules/agent_upgrade/agent/wm_agent_upgrade_com.c`) only rejects `..` in a caller-supplied filename before joining it under `INCOMING_DIR`; it doesn't check whether the final path component is a symlink. The subsequent `wfopen(final_path, "wb")` follows a symlink there and truncates its target on open. No `O_NOFOLLOW`, `openat()`, or `fstat()` check exists in this path.

Nothing legitimate ever places a symlink under `var/incoming`, so rejecting one costs nothing.

> [!NOTE]
> Not a vulnerability — no attacker-reachable path is known today. Filed as hardening only.

## Suggested fix

Rework `_jailfile()` and the `open`/`write` handlers to use descriptor-relative, symlink-safe operations: open `INCOMING_DIR` with `O_DIRECTORY|O_CLOEXEC`, reject empty/`.`/`..`/separator-containing names, then `openat()` the target with `O_WRONLY|O_CREAT|O_TRUNC|O_NOFOLLOW|O_CLOEXEC`, and `fstat()` the descriptor to confirm it's a regular file before `fdopen()`.

## Credit

Found and reported by @halas98.

cborla:
## Description

`_jailfile()` only rejected `..` in a caller-supplied file name before joining it under `INCOMING_DIR`, and never checked what the resulting path actually was. `wm_agent_upgrade_com_open()` then called `wfopen(final_path, "wb")`, which on POSIX is a plain `fopen()` and on Windows reaches `CreateFile` without `FILE_FLAG_OPEN_REPARSE_POINT`. On both platforms a symlink sitting at that path is followed and its target truncated on open, and a hard link — which no file-type test can distinguish from an ordinary file — is written straight through. No `O_NOFOLLOW`, `openat()` or `fstat()` check existed anywhere in this path.

The same `_jailfile()` exists byte-for-byte in `src/os_execd/wcom.c`, whose `wcom_uncompress()` performs the same unprotected `wfopen(final_target, "wb")` under `INCOMING_DIR`. Both dispatchers are reachable from the same remote request path (`src/client-agent/request.c`), so hardening only the upgrade module would have left the identical primitive exposed — and that second one is the worse of the two: `wcom_uncompress()` does not merely truncate the target, it writes the decompressed contents of the source archive into it, so the write is fully attacker-controlled rather than just destructive. That was measured on a live agent, not inferred; see the evidence below.

Not a vulnerability in itself — no attacker-reachable path is known today, and nothing legitimate ever places a link of either kind under `var/incoming`. Filed and fixed as hardening.

Closes #38200

Found and reported by @halas98.

## Proposed Changes

- **New `w_fopen_nofollow()`** (`src/shared/file_op.c`): creates or truncates a file inside a base directory without following links, so both call sites share one implementation instead of duplicating platform-specific logic. The order of operations is the substance of the fix: the file is opened **without** truncating, the descriptor is vetted, and only then is the file truncated. Truncating at open time would destroy the target before anything about it could be checked, which is exactly how a hard link gets through.
- **POSIX path**: opens the base directory with `O_RDONLY|O_DIRECTORY|O_CLOEXEC`, then `openat()`s the target with `O_WRONLY|O_CREAT|O_NOFOLLOW|O_CLOEXEC|O_NONBLOCK`, `fstat()`s the descriptor to confirm a regular file with a link count of exactly 1, and truncates with `ftruncate()` before `fdopen()`. `O_NONBLOCK` matters on its own: without it an existing FIFO at that path makes `openat()` block waiting for a reader, and no later check is ever reached — the thread serving the upgrade socket simply hangs.
- **Windows is covered too**, rather than left unchanged. `wfopen()` cannot be used because `CREATE_ALWAYS` truncates at open time. `w_fopen_nofollow()` opens with `OPEN_ALWAYS | FILE_FLAG_OPEN_REPARSE_POINT`, which yields a handle to a reparse point itself rather than to its target, rejects the handle if `GetFileInformationByHandle()` reports `FILE_ATTRIBUTE_REPARSE_POINT`, `FILE_ATTRIBUTE_DIRECTORY` or `nNumberOfLinks != 1`, and only then truncates with `SetEndOfFile()`. The hard-link check matters more here than on POSIX: an NTFS hard link needs no privilege to create (`mklink /H`), unlike a symlink.
- **Windows error reporting**: Win32 errors are no longer assigned straight into `errno`. The numbering spaces are unrelated, so `strerror()` printed something misleading, and a raw Win32 code could collide with an errno the callers test for — `ERROR_INVALID_TARGET_HANDLE` is 114, which is mingw's `ELOOP`, the value used to report a refused symlink. Win32 failures now go through a small mapping that cannot return `ELOOP`; CRT failures (`_open_osfhandle()`, `_fdopen()`) leave `errno` as the CRT set it. The `_fdopen()` failure path also called `CloseHandle()` after `_open_osfhandle()` had transferred ownership to the descriptor, leaking one CRT descriptor per failure; it now calls `_close()`.
- **New `w_is_bare_filename()`** (`src/shared/file_op.c`): single source of truth for the name validation, rejecting the empty string, `.`, `..`, any name referring to a parent folder, and any name containing a path separator. Rejecting separators is what keeps the directory-relative `openat()` confined to the base directory, since an absolute or separator-containing name would otherwise be resolved ignoring the directory descriptor.
- **`wm_agent_upgrade_com.c`**: `_jailfile()` now delegates to `w_is_bare_filename()`, and `wm_agent_upgrade_com_open()` calls `w_fopen_nofollow()` on every platform instead of `wfopen()`. The caller-supplied mode is still honoured and still restricted to `w`/`wb`.
- **`os_execd/wcom.c`**: the same hardening applied to its duplicated `_jailfile()` and to the `wfopen()` in `wcom_uncompress()`.
- **A message of its own for the rejection.** A refused symlink is reported as `ELOOP`, and both the log line and the command response derived their text from `strerror()`, which has nothing useful to say about it on Windows: mingw carries no message for `ELOOP` and prints `"Unknown error"`, so the one failure worth alerting on was indistinguishable from noise. This adds `WM_UPGRADE_SYMLINK_REJECTED` (8142) and a fixed reason in the ack; the error code stays `ERROR_FILE_OPEN`, so the response protocol is unchanged. `errno` is captured before logging in both call sites, since `mterror()` and `gzclose()` are both free to clobber it before the reason text is read.
- **`O_DIRECTORY` fallback** (`src/headers/defs.h`): `openat()` and `O_DIRECTORY` are the first uses of either in this tree, and the Makefile still carries AIX, Solaris and HP-UX agent branches. This is a build concern rather than a behavioural one — WPK remote upgrade is not offered for those platforms, so neither call site is ever exercised there — and falling back to 0 costs only an early check, since `openat()` on a non-directory descriptor fails with `ENOTDIR` anyway. `O_NOFOLLOW` deliberately gets no fallback: defining it to 0 would quietly disable a security check instead of failing the build.

Scope deliberately left out of this PR, worth a follow-up issue: `_uncompress()` writes into `TMP_DIR` through `wfopen()` at a predictable name, `wm_agent_upgrade_com_sha1()` reopens by path rather than by descriptor, `write`/`close` identify the open file by `strcmp()` on its path instead of by inode, and `UnmergeFiles()` repeats the same weak defense. Separately, `wfopen()` itself carries the same two Windows defects fixed here in the new helper; correcting it there would change `errno` for every consumer in the tree, which is more than this PR should take on.

#### Unit tests

`TARGET=agent`, exercising the real file system rather than mocks — a link cannot be mocked into existence, and the property under test is precisely what the operating system does with one:

```
$ ./shared/test_file_op
[ RUN      ] test_w_fopen_nofollow_regular_file              [       OK ]
[ RUN      ] test_w_fopen_nofollow_truncates_existing_file   [       OK ]
[ RUN      ] test_w_fopen_nofollow_symlink_rejected          [       OK ]
[ RUN      ] test_w_fopen_nofollow_hard_link_rejected        [       OK ]
[ RUN      ] test_w_fopen_nofollow_dangling_symlink_rejected [       OK ]
[ RUN      ] test_w_fopen_nofollow_fifo_rejected             [       OK ]
[ RUN      ] test_w_fopen_nofollow_directory_rejected        [       OK ]
[ RUN      ] test_w_fopen_nofollow_invalid_name              [       OK ]
[ RUN      ] test_w_fopen_nofollow_invalid_mode              [       OK ]
[ RUN      ] test_w_fopen_nofollow_missing_basedir           [       OK ]
[==========] tests: 53 test(s) run.
[  PASSED  ] 53 test(s).
```

`test_w_fopen_nofollow_symlink_rejected` and `test_w_fopen_nofollow_hard_link_rejected` are the regression tests for this issue: each writes a known payload into a `victim` file, points a link at it, asserts the open fails, and asserts that `victim` still holds its bytes afterwards.

#### End-to-end A/B on a real Windows agent

Wazuh manager 4.14.9 driving a Windows Server 2022 agent, exercising the actual upgrade socket rather than a standalone program. Commands are injected on the manager through `queue/sockets/remote` using the same `"%.3d upgrade <json>"` format that `wm_agent_upgrade_send_open()` uses, which reaches `wm_agent_upgrade_process_command()` on the agent by way of `client-agent/request.c`; `wcom` is reached the same way with a `com` target. Stock 4.14.7 from the official MSI, then the same agent upgraded by a WPK built from this branch.

| Test | stock 4.14.7 | this PR |
|---|---|---|
| symlink to a file | `ok` — target **36 → 0 bytes** | `error 6 the path is a symbolic link` — 36 bytes, content intact |
| **hard link** (`mklink /H`, no privilege needed) | `ok` — target **30 → 0 bytes** | `error 6 Too many links` — 30 bytes, content intact |
| **symlink, `com uncompress`** | `ok` — target **overwritten with the archive payload** | `err Unable to open target` — content intact |
| junction (`mklink /J`) | `error 6 Input/output error` | `error 6 Permission denied` |
| directory | `error 6 Input/output error` | `error 6 Permission denied` |
| `subdir/x.wpk` | `error 6 No such process` | `error 5 Invalid file name` |
| truncation of an existing file | 36 → 0 bytes | 36 → 0 bytes |
| plain valid name | `error 0` | `error 0` |

The junction and directory rows show the Windows error-reporting fix rather than a change in outcome: both were already refused, but by `CreateFile` failing with `ERROR_ACCESS_DENIED` for `GENERIC_WRITE` on a directory. That code is 5, and the old `errno = GetLastError()` made `strerror()` print `EIO`'s text, `"Input/output error"`. It now maps to `EACCES` and reads `"Permission denied"`, which is what actually happened.

The rejection as it appears on the agent:

```
ack:       {"error":6,"message":"File Open Error: the path is a symbolic link","data":[]}
ossec.log: wazuh-modulesd:agent-upgrade: ERROR: (8142): At open: Refused to open 'malicious.wpk': the path is a symbolic link.
           wazuh-agent: ERROR: At WCOM uncompress: Refused to open 'incoming\target.wpk': the path is a symbolic link
```

The two lines carry different tags because the two call sites live in different daemons.

No regression in the upgrade flow. A real WPK upgrade was run twice: once from the stock agent, which delivered the patched build, and once again from the **patched** agent — the run that matters, because that is this code executing `open → write → close → sha1 → upgrade`:

```
Agent 004 upgraded: Wazuh v4.14.7 -> Wazuh v4.14.9      # transfer executed by stock code
Agent 004 upgraded: Wazuh v4.14.9 -> Wazuh v4.14.9      # transfer executed by this code
```

Repeating the symlink and hard-link tests after that upgrade still refuses both with the targets intact.

#### End-to-end A/B on a real Linux agent

Ubuntu 22.04 agent against the same manager. On POSIX both call sites listen on local sockets — the upgrade module on `queue/sockets/upgrade` and `wcom` on `queue/sockets/com` — so these were driven directly on the agent.

| Test | stock 4.14.7 | this PR |
|---|---|---|
| symlink, `upgrade` socket | `ok` — target **36 → 0 bytes** | `error 6 the path is a symbolic link` — 36 bytes, content intact |
| **hard link** | `ok` — target **30 → 0 bytes** | `error 6 Too many links` — 30 bytes, content intact |
| **symlink, `com uncompress`** | `ok` — target **overwritten with the archive payload** | `err Unable to open target` — content intact |
| **FIFO** | **blocks** — the socket thread is left stuck in `openat()` | `error 6 No such device or address` (`ENXIO`), returns immediately |
| directory | `error 6 Is a directory` | `error 6 Is a directory` — unchanged |
| `subdir/x.wpk` | `error 6 No such file or directory` | `error 5 Invalid file name` |
| `../x.wpk` | `error 5 Invalid file name` | `error 5 Invalid file name` — unchanged |
| `.` | `error 6 Is a directory` | `error 5 Invalid file name` |
| truncation of an existing file | 36 → 0 bytes | 36 → 0 bytes |
| mode of a newly created file | `644 root:root` | `640 root:root` |

The FIFO row is why `O_NONBLOCK` is part of this change: on the stock agent that request never returns and the thread serving the upgrade socket stays stuck, so the agent had to be restarted before the run could continue. The `com uncompress` row is the arbitrary-write case — on the stock agent the target did not end up empty, it ended up containing the payload from the source archive. The truncation row is the regression check for moving truncation out of the open: with `O_TRUNC` gone, a mistake there would leave stale bytes in legitimately written WPKs. A real WPK upgrade executed by the patched agent (`v4.14.9 -> v4.14.9`) confirms the whole path end to end.

Two rows deliberately reported as unchanged rather than as wins: a plain directory was already refused by `fopen()` with `EISDIR`, and `../x.wpk` was already caught by `w_ref_parent_folder()`. What the name validation actually adds is rejecting `.` and rejecting embedded separators, both of which the old `_jailfile()` accepted.

For transparency on how these were gathered: the Linux stock comparison and the hard-link comparison were run against successive builds of this branch rather than in a single pass — the hard link was found by review after the first round — while the Windows table above is a single A/B against the final state of the branch. Every row was observed, none inferred.

#### Not verified

- The `_fdopen()` descriptor leak is verified by inspection and compilation only. Forcing that path means exhausting the CRT descriptor table, which was not attempted.
- The `O_DIRECTORY` fallback cannot be exercised here: on Linux and mingw the macro is defined, so the fallback is inert. It would take an AIX, Solaris or HP-UX build to reach it.
- macOS was not exercised on an agent. The POSIX path is shared with Linux, and `O_NOFOLLOW`, `O_NONBLOCK`, `st_nlink` and `ftruncate()` behave the same there.

### Artifacts Affected

- `wazuh-modulesd` — agent upgrade module, both POSIX and Windows agent builds.
- `wazuh-execd` — `wcom` request dispatcher, agent-side only. Reachable remotely from the manager on every platform, and additionally on a local socket (`queue/sockets/com`) on POSIX.
- `libwazuh` — two new public functions in `src/shared/file_op.c` plus one file-static Win32 error mapper; no existing function changes behavior, so no other consumer is affected.

Behavioral change only, no new files or binaries. One deliberate difference: files created under `var/incoming` are now created with mode `0640` instead of `fopen()`'s `0666 & ~umask`, matching the permissions `_unsign()` already applies to its temporary files. Files that already exist keep their mode, since the creation mode only applies on creation.

### Configuration Changes

N/A — no configuration option is added, removed or reinterpreted.

### Documentation Updates

N/A — no user-facing behavior changes; the new shared functions are documented with Doxygen comments in `src/headers/file_op.h`.

### Tests Introduced

- `src/unit_tests/shared/test_file_op.c` — ten new tests covering `w_fopen_nofollow()` against the real file system: regular file creation, truncation of an existing file, symlink rejection with the target left intact, hard link rejection with the target left intact, dangling symlink rejection with no file created at the link target, FIFO rejection without blocking, directory rejection, invalid names (empty, `.`, `..`, `../victim`, `subdir/victim`, absolute, `NULL`), invalid modes, and a missing base directory.
- `src/unit_tests/wazuh_modules/agent_upgrade/test_wm_agent_upgrade_com.c` — three new `_jailfile()` rejection cases (empty name, `.`, path separator including the Windows separator under `TEST_WINAGENT`), a new case asserting that an `ELOOP` failure produces message 8142 and the fixed reason in the ack rather than `strerror()` text, and the two `wm_agent_upgrade_com_open()` tests migrated from `__wrap_wfopen` to the new `__wrap_w_fopen_nofollow`.
- `src/unit_tests/wrappers/wazuh/shared/file_op_wrappers.{c,h}` — wrapper and `expect_` helper for `w_fopen_nofollow`. It intentionally has no `__real_` fallback: referencing one would force every test binary that links these wrappers to add `-Wl,--wrap,w_fopen_nofollow`, which is exactly why `--wrap,wfopen` appears in so many flag lists today.

Note for reviewers on where the evidence comes from: on `4.14.x`, `src/unit_tests/wazuh_modules/CMakeLists.txt` only adds `add_subdirectory(agent_upgrade)` under `TARGET=server`, while `test_wm_agent_upgrade_com` and `test_wm_agent_upgrade_agent` are declared in the `else()` branch of that subdirectory's own `CMakeLists.txt`. Those two binaries are therefore never compiled on this branch — the changes to that file are verified by compilation only, and will run on the upward merge to `5.0.0`, where both are wired for `agent` and `winagent`. That is why the executable unit-test evidence above comes from `test_file_op`, where the security property now lives, and why the module-level behavior is evidenced by the end-to-end A/B on real agents instead.

## Review Checklist

- [ ] Code changes reviewed
- [x] Configuration changes documented — N/A, no configuration changes
- [x] Developer documentation reflects the changes — Doxygen comments on both new shared functions

### DIFF
--- a/src/error_messages/error_messages.h
+++ b/src/error_messages/error_messages.h
@@ -568,6 +568,7 @@
 #define WM_UPGRADE_UNSIGN_FILE_ERROR                "(8139): At %s: Could not unsign package file '%s'"
 #define WM_UPGRADE_FILE_OPEN_ERROR                  "(8140): At %s: Unable to open '%s'"
 #define WM_UPGRADE_CANNOT_READ                      "(8141): At %s: Unable to read '%s'"
+#define WM_UPGRADE_SYMLINK_REJECTED                 "(8142): At %s: Refused to open '%s': the path is a symbolic link."
 
 #define MOD_TASK_CHECK_DB_ERROR                     "(8250): DB integrity is invalid. Exiting..."
 #define MOD_TASK_CREATE_SOCK_ERROR                  "(8251): Queue '%s' not accessible: '%s'. Exiting..."
--- a/src/headers/defs.h
+++ b/src/headers/defs.h
@@ -368,6 +368,15 @@ https://www.gnu.org/licenses/gpl.html\n"
 #define O_CLOEXEC 0
 #endif
 
+/* Not every agent platform this tree still builds for (AIX, Solaris 10, HP-UX) is guaranteed to define
+ * O_DIRECTORY. Falling back to 0 only loses an early "this must be a directory" check: openat() on a
+ * descriptor that is not a directory fails with ENOTDIR anyway. Note that O_NOFOLLOW deliberately gets
+ * no such fallback -- defining it to 0 would silently disable a security check instead of failing the
+ * build, which is the outcome we want to hear about. */
+#ifndef O_DIRECTORY
+#define O_DIRECTORY 0
+#endif
+
 /* XML global elements */
 #ifndef xml_global
 #define xml_global "global"
--- a/src/headers/file_op.h
+++ b/src/headers/file_op.h
@@ -507,15 +507,16 @@ FILE * wfopen(const char * pathname, const char * mode);
  * @brief Create or truncate a file inside a base directory for writing, without following symlinks.
  *
  * Intended for directories that only ever hold files written by Wazuh itself (var/incoming and
- * friends): a symlink, FIFO, device or directory found at the target path is rejected instead of
- * being opened through. @p filename must be a bare file name; it is rejected if it is empty, "."
- * or "..", if it refers to a parent folder, or if it contains a path separator, so the resulting
- * open cannot escape @p basedir.
+ * friends): a symlink, hard link, FIFO, device or directory found at the target path is rejected
+ * instead of being written through. @p filename must be a bare file name; it is rejected if it is
+ * empty, "." or "..", if it refers to a parent folder, or if it contains a path separator, so the
+ * resulting open cannot escape @p basedir.
  *
- * On Linux/macOS the file is opened relative to a descriptor of @p basedir with O_NOFOLLOW, and the
- * descriptor is confirmed to be a regular file before being wrapped in a stream. On Windows the
- * file is opened without reparse-point processing and truncated only after the handle has been
- * confirmed to be a regular file, so the target of a symlink or junction is never modified.
+ * On both platforms the file is opened without truncating, the descriptor is vetted, and only then is
+ * the file truncated: truncating at open time would destroy the target of a hard link before anything
+ * about it could be checked. On Linux/macOS the open is relative to a descriptor of @p basedir and
+ * uses O_NOFOLLOW; on Windows it skips reparse-point processing. In both cases the descriptor must
+ * turn out to be a regular file with a link count of exactly 1.
  *
  * @param basedir Base directory holding the file. Not created by this function.
  * @param filename Bare file name inside @p basedir.
--- a/src/headers/file_op.h
+++ b/src/headers/file_op.h
@@ -408,6 +408,19 @@ int w_ref_parent_folder(const char * path);
 
 
 /**
+ * @brief Check that a name is a bare file name, safe to join to a base directory.
+ *
+ * Rejects the empty string, "." and "..", any name referring to a parent folder, and any name
+ * containing a path separator, so that joining it to a base directory cannot escape that directory
+ * and so that it can be used as-is in a directory-relative open.
+ *
+ * @param filename Name to be checked.
+ * @return 1 if the name is a bare file name, 0 otherwise.
+ */
+int w_is_bare_filename(const char * filename);
+
+
+/**
  * @brief Read directory and return an array of contained files and folders, sorted alphabetically.
  *
  * @param name Path of the directory.
@@ -488,6 +501,28 @@ int w_stat(const char * pathname,
  * @return File pointer.
  */
 FILE * wfopen(const char * pathname, const char * mode);
+
+
+/**
+ * @brief Create or truncate a file inside a base directory for writing, without following symlinks.
+ *
+ * Intended for directories that only ever hold files written by Wazuh itself (var/incoming and
+ * friends): a symlink, FIFO, device or directory found at the target path is rejected instead of
+ * being opened through. @p filename must be a bare file name; it is rejected if it is empty, "."
+ * or "..", if it refers to a parent folder, or if it contains a path separator, so the resulting
+ * open cannot escape @p basedir.
+ *
+ * On Linux/macOS the file is opened relative to a descriptor of @p basedir with O_NOFOLLOW, and the
+ * descriptor is confirmed to be a regular file before being wrapped in a stream. On Windows the
+ * file is opened without reparse-point processing and truncated only after the handle has been
+ * confirmed to be a regular file, so the target of a symlink or junction is never modified.
+ *
+ * @param basedir Base directory holding the file. Not created by this function.
+ * @param filename Bare file name inside @p basedir.
+ * @param mode Open mode, either "w" or "wb".
+ * @return File pointer on success, NULL on error (sets errno).
+ */
+FILE * w_fopen_nofollow(const char * basedir, const char * filename, const char * mode);
 
 
 /**
--- a/src/os_execd/wcom.c
+++ b/src/os_execd/wcom.c
@@ -159,7 +159,8 @@ size_t wcom_uncompress(const char * source, const char * target, char ** output)
         return strlen(*output);
     }
 
-    if (ftarget = wfopen(final_target, "wb"), !ftarget) {
+    // Not wfopen(): a symlink left at final_target would be followed, truncating its target.
+    if (ftarget = w_fopen_nofollow(INCOMING_DIR, target, "wb"), !ftarget) {
         gzclose(fsource);
         merror("At WCOM uncompress: Unable to open '%s'", final_target);
         os_strdup("err Unable to open target", *output);
@@ -536,7 +537,10 @@ void * wcom_main(__attribute__((unused)) void * arg) {
 
 int _jailfile(char finalpath[PATH_MAX + 1], const char * basedir, const char * filename) {
 
-    if (w_ref_parent_folder(filename)) {
+    // A bare file name is all that is ever legitimate here. Rejecting separators also keeps the
+    // directory-relative open in w_fopen_nofollow() confined to basedir, since a separator-containing
+    // or absolute name would otherwise be resolved ignoring the directory descriptor.
+    if (!w_is_bare_filename(filename)) {
         return -1;
     }
 
--- a/src/os_execd/wcom.c
+++ b/src/os_execd/wcom.c
@@ -162,7 +162,13 @@ size_t wcom_uncompress(const char * source, const char * target, char ** output)
     // Not wfopen(): a symlink left at final_target would be followed, truncating its target.
     if (ftarget = w_fopen_nofollow(INCOMING_DIR, target, "wb"), !ftarget) {
         gzclose(fsource);
-        merror("At WCOM uncompress: Unable to open '%s'", final_target);
+        // Same reasoning as in the upgrade module: strerror() has no useful text for ELOOP on Windows,
+        // so the one failure worth alerting on gets its own message.
+        if (errno == ELOOP) {
+            merror("At WCOM uncompress: Refused to open '%s': the path is a symbolic link", final_target);
+        } else {
+            merror("At WCOM uncompress: Unable to open '%s'", final_target);
+        }
         os_strdup("err Unable to open target", *output);
         return strlen(*output);
     }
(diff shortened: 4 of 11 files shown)


nbertoldo (approved):
LGTM.
- [x] Code changes reviewed
- [x] Configuration changes documented — N/A, no configuration changes
- [x] Developer documentation reflects the changes — Doxygen comments on both new shared functions

cborla (commented):
LGTM

PR #38301, opened 2026-08-10
Fix missing FIM inventory for files with non-ANSI names on Windows

### LINKED ISSUE #34878: Wazuh FIM Problem for Turkish Characters
Hi 

I installed Wazuh Agent Windows agent version v4.14.3
If a folder or file name contains the letters "İ, ı, Ğ, ğ, Ş, and ş", the records will not appear on the Wazuh server.

Darioortegaleyva:
## Description

This PR fixes File Integrity Monitoring (FIM) on Windows silently dropping files and folders whose names contain characters outside the process ANSI code page (for example the Turkish characters `İ, ı, Ğ, ğ, Ş, ş`).

The FIM baseline scan enumerated directories through the narrow POSIX API (`wopendir()` -> `opendir()` + `readdir()`), which returns entry names in the process ANSI code page. On any host whose code page does not contain those characters (CP850, CP1252, etc., i.e. almost every non-Turkish Windows) the CRT performs a best-fit mapping (`İ` -> `I`, `ı` -> `i`, `Ğ` -> `G`, `ğ` -> `g`, `Ş` -> `S`, `ş` -> `s`). The resulting name does not match the real file on disk, the subsequent `stat()` fails, and the entry is dropped, so it never enters the FIM database and never reaches the manager.

The rest of the FIM file access path on Windows is already UTF-8 aware (`w_lstat` -> `utf8_stat64` -> `_wstat64`, `w_get_file_attrs` -> `utf8_GetFileAttributes`, hashing -> `wfopen` -> `wCreateFile`). The only broken link was the directory enumeration. This PR makes the enumeration use the wide API and convert each name to UTF-8, consistent with the rest of the path.

**Issue:** wazuh/wazuh#34878
**Internal Reference:** N/A

## Proposed Changes

- Updated `src/syscheckd/src/create_db.c` (`fim_directory`).
- Fixed Windows directory enumeration to use `FindFirstFileW`/`FindNextFileW` and convert each file name to UTF-8 with `wide_to_utf8()`, matching the pattern already used by `expand_win32_wildcards`.
- Lowercase the enumerated name ASCII-only (Windows/NTFS is case-insensitive): non-ASCII UTF-8 bytes are all `>= 0x80` and never in `A`..`Z`, so the on-disk bytes are preserved and `stat()` still resolves. A Unicode case fold was deliberately avoided because it can remap the very characters this fixes (e.g. `İ` U+0130) to a name that no longer exists on the volume.
- Report `FindFirstFileW` failures via `win_strerror(GetLastError())` (the API sets `GetLastError()`, not `errno`).
- Refactored the per-entry path building into a shared helper `fim_process_dir_entry` so the loop body is not duplicated between the Windows and POSIX branches.
- POSIX enumeration (`wopendir`/`readdir`) is left unchanged; the network-path guard is preserved on Windows.

### Scope and known limitations

- This PR fixes the **scheduled scan** path, which is what the report shows (Inventory). Real-time (`run_realtime.c`) and whodata (`win_whodata.c`) still convert names through `CP_ACP` and remain unable to represent non-ANSI names; unifying FIM path canonicalization across scan, real-time and whodata is tracked as a separate follow-up (wazuh/wazuh#39240).
- File names whose full UTF-8 path exceeds the pre-existing `PATH_MAX` (260) **byte** guard are still skipped. That guard protects the downstream FIM path buffers, which are fixed `char[PATH_MAX]` (for example the report-changes path runs `abspath()` into a 260-byte buffer): relaxing it without reworking those buffers would turn a silent skip into a downstream overflow, not a fix. Supporting names under 260 characters but over 260 UTF-8 bytes requires reworking UTF-8 path-length handling across scan, diff, realtime and whodata, and is tracked as a dedicated follow-up (wazuh/wazuh#39240).

### Results and Evidence

Validated on a real Windows host with active code page CP850 (which does not contain the Turkish characters), reproducing the same conditions as the report.

Test folder with two files (ground truth on disk, UTF-8):

```sql
file1.txt
İ, ı, Ğ, ğ, Ş, and ş.txt
```

Before the fix, replicating exactly what FIM does (`opendir()` + `readdir()` + `stat()`):

```sql
name="file1.txt"                len=9  stat=OK
name="I, i, G, g, S, and s.txt" len=24 stat=FAIL(dropped by FIM)
```

After the fix, running the actual patched `fim_directory`/`fim_process_dir_entry` (compiled verbatim from this PR and linked against the real Wazuh `utf8_winapi_wrapper.c` and `utf8_op.c`), with the FIM callback stubbed to call the real `utf8_stat64` (the same function FIM reaches through `w_lstat`):

```sql
=== REAL patched fim_directory over C:\turk_repro ===
fim_checker -> "c:\turk_repro\file1.txt"                 utf8_stat64=OK
fim_checker -> "c:\turk_repro\İ, ı, Ğ, ğ, Ş, and ş.txt"  utf8_stat64=OK
```

The Turkish file is now enumerated with the correct UTF-8 name and resolves via the real `utf8_stat64`, so it is processed by FIM instead of being dropped (path is lower-cased on Windows, as FIM does).

### Artifacts Affected

- `src/syscheckd/src/create_db.c`
- `src/unit_tests/syscheckd/test_create_db.c`

### Configuration Changes

N/A

### Documentation Updates

N/A

### Tests Introduced

- **Scope:** Unit Tests + manual validation of the real patched code on a real Windows host.
- **Details:**
    - The Windows branches of the existing `fim_directory` unit tests (`test_fim_directory`, `test_fim_directory_ignore`, `test_fim_directory_opendir_error`) were migrated to mock `FindFirstFile`/`FindNextFile` instead of `opendir`/`readdir`, covering the new wide enumeration control flow (happy path, `.`/`..` skipping, and open error). The POSIX branches are unchanged.
    - Encoding correctness was validated by running the actual patched `fim_directory`/`fim_process_dir_entry` (compiled verbatim from this PR) linked against the real `utf8_winapi_wrapper.c` and `utf8_op.c`, on a real Windows host with code page CP850, as shown in the evidence above.

### Validation status

- Done: new Windows enumeration code compiles clean under `-Wall -Wextra` (MinGW); real patched code confirmed to enumerate and resolve the non-ANSI file on Windows CP850.
- Pending (CI / QA): winagent CMocka unit-test run (needs the winagent toolchain + Wine); full agent build, install and end-to-end verification of the file reaching the manager FIM inventory. The Linux/POSIX unit tests are unaffected by this change.

## Review Checklist

    - [ ] Linux
    - [ ] Windows
    - [ ] MAC OS X

- Memory tests for Linux

    - [ ] Coverity
    - [ ] AddressSanitizer

**General Checklist:**

- [ ] Code changes reviewed


Miguevrgo (changes requested):
# Review
### 1. `strerror(errno)` after `FindFirstFileW` (bug)
`create_db.c:872`

`FindFirstFileW` does not set `errno`; it sets `GetLastError()`. The old `opendir()` path did set `errno`. The warning now prints a stale/wrong message. Use `FormatMessage(GetLastError())` instead.

### 2. `str_lowercase` is not UTF-8 aware (undermines the fix)
`create_db.c:829` (macro in `string_op.h:48`)

`str_lowercase` lowercases byte-by-byte with `tolower`, which only affects `A-Z`. Turkish/non-ANSI names are now multi-byte UTF-8, so they are no longer lowercased. On the case-insensitive NTFS this can produce duplicate DB keys for the same file (e.g. `ı.txt` vs `i.txt`). Lowercase the wide name (`CharLowerW` / `_wcslwr`) before `wide_to_utf8`. Also, `tolower(*x)` on a signed `char` >= 0x80 is technically UB. https://en.cppreference.com/cpp/string/byte/tolower


Can you take a look into those two?

Miguevrgo (dismissed):
LGTM, I would only say to remove some comments which I believe are not necessary, for example:
    // Ignore . and ..
// Check if the file name is already null terminated
 Check if the full path is too long if it is, skip this file
    // and log a warning,
// Open the directory given

But should work fine

src/syscheckd/src/create_db.c
```diff
@@ -785,70 +786,126 @@ void fim_checker(const char *path,
...
+
+        if (name == NULL) {
+            mdebug2("Cannot convert a directory entry under '%s' to UTF-8.", dir);
             continue;
         }
-#endif
 
-        snprintf(s_name, PATH_MAX + 2 - path_size, "%s", entry->d_name);
+        // Lowercase ASCII only: keeps the on-disk bytes so stat resolves, and
+        // matches the ASCII canonicalization used by realtime/whodata.
```
MarcelKemp: **[S2 · non-blocking · introduced]** The comment claims this loop matches the canonicalization used by realtime and whodata; that is true for the case fold but not for the encoding. `run_realtime.c:481` and `get_whodata_path` (`win_whodata.c:1657`) both still convert with `WideCharToMultiByte(CP_ACP, ...)`, so their names arrive mangled before any case folding. Worth not asserting a consistency that does not exist yet, and widening the follow-up issue to whodata as well as realtime.

```suggestion
        // Lowercase ASCII only: keeps the on-disk bytes so stat resolves. Note that
        // realtime (run_realtime.c) and whodata (win_whodata.c) still convert names
        // with CP_ACP, so their names are not consistent with this one yet.
```

src/syscheckd/src/create_db.c
```diff
@@ -785,70 +786,126 @@ void fim_checker(const char *path,
...
+    }
+    *(s_name) = '\0';
+    path_size = strlen(f_name);
+
+#ifdef WIN32
+    // PATH_MAX is 260 on Windows; reserve one char for the null terminator.
+    if (path_size + strlen(name) >= PATH_MAX) {
+        mwarn(FIM_ERROR_PATH_TOO_LONG, f_name, name, PATH_MAX);
+        return;
+    }
```
MarcelKemp: **[S2 · blocking · aggravated]** `PATH_MAX` is 260 *characters* on Windows, but `name` is now UTF-8 instead of single-byte ANSI, so this compares bytes against a character limit. `c:\data\` (8 bytes) plus a 90-character CJK or Cyrillic name is 278 bytes: the file is dropped with `FIM_ERROR_PATH_TOO_LONG` ("The maximum length is 260 characters") even though the real path is 98 characters, and `mwarn` fires once per file per scan. Files that were inventoried on 4.14.8 stop being monitored — the same silent loss of FIM coverage the issue is about. Count in wide characters before converting, and size `f_name` for UTF-8:

```c
    // in fim_directory, before converting: count in wide characters
    size_t w_dir_len = wcslen(wpattern) - 2;   // pattern is dir + PATH_SEP + '*'
    if (w_dir_len + 1 + wcslen(fd.cFileName) >= MAX_PATH) {
        mwarn(FIM_ERROR_PATH_TOO_LONG, dir, "", MAX_PATH);
        continue;
    }
```

If you would rather not change the buffer size in this PR, please say so in the description and open a separate issue: as it stands it is a regression against 4.14.8.

src/syscheckd/src/create_db.c
```diff
@@ -785,70 +786,126 @@ void fim_checker(const char *path,
...
+        }
 
-#ifdef WIN32
-        str_lowercase(f_name);
-#endif
-        // Process the event related to f_name
-        fim_checker(f_name, evt_data, configuration, dbsync_txn, ctx);
+        fim_process_dir_entry(dir, name, f_name, evt_data, configuration, dbsync_txn, ctx);
+        os_free(name);
+    } while (FindNextFileW(hFind, &fd));
```
MarcelKemp: **[S3 · non-blocking · introduced]** Any `FindNextFileW` failure is treated as end-of-directory and `fim_directory` still returns 0, so a mid-enumeration error (ERROR_ACCESS_DENIED after an ACL change, ERROR_NETNAME_DELETED, ERROR_FILE_CORRUPT) after 3 of 5000 entries makes the caller believe the directory was fully scanned and `fim_db_transaction_deleted_rows` reports the remaining 4997 as removed. `readdir` had the same hole and the only caller discards the return value, so propagating the failure needs separate plumbing — but the error should at least be diagnosable:

```suggestion
    } while (FindNextFileW(hFind, &fd));

    if (GetLastError() != ERROR_NO_MORE_FILES) {
        mwarn(FIM_PATH_NOT_OPEN, dir, win_strerror(GetLastError()));
    }
```

src/syscheckd/src/create_db.c
```diff
@@ -785,70 +786,126 @@ void fim_checker(const char *path,
...
+        mwarn(FIM_PATH_NOT_OPEN, dir, win_strerror(last_error));
+        os_free(f_name);
+        return OS_INVALID;
+    }
+
+    do {
+        char *name = wide_to_utf8(fd.cFileName);
+
+        if (name == NULL) {
+            mdebug2("Cannot convert a directory entry under '%s' to UTF-8.", dir);
```
MarcelKemp: **[S3 · non-blocking · introduced]** Two things here: the message is hardcoded, while every other FIM message is a define with a code in `src/error_messages/debug_messages.h` (e.g. `FIM_CONFIGURATION_NOTFOUND "(6319): ..."`); and `mdebug2` hides it at the default log level, when what it reports is a file missing from the inventory — exactly what the issue is about. Consider a define plus `mwarn`:

```c
/* debug_messages.h */
#define FIM_UTF8_CONVERSION_FAILED "(63XX): Cannot convert a directory entry under '%s' to UTF-8."

/* create_db.c */
            mwarn(FIM_UTF8_CONVERSION_FAILED, dir);
```

src/unit_tests/syscheckd/test_create_db.c
```diff
@@ -3122,17 +3156,59 @@ static void test_fim_directory(void **state) {
...
+    int ret;
+
+    expect_function_call_any(__wrap_pthread_rwlock_wrlock);
+    expect_function_call_any(__wrap_pthread_rwlock_unlock);
+    expect_function_call_any(__wrap_pthread_rwlock_rdlock);
+    expect_function_call_any(__wrap_pthread_mutex_lock);
+    expect_function_call_any(__wrap_pthread_mutex_unlock);
+
+    expect_memory(wrap_FindFirstFile, lpFileName, L"test\\*", (wcslen(L"test\\*") + 1) * sizeof(wchar_t));
+    will_return(wrap_FindFirstFile, "A" "\xc5\x9f" ".txt");   // "Aş.txt" in UTF-8
```
MarcelKemp: **F20 · S3 · blocking · introduced** — this test cannot fail on the regression it is named for.

`ş` is U+015F, which is *already lowercase* (its uppercase is `Ş`, U+015E). The code removed in `032e380ab7` was:

```c
wchar_t w_lower[MAX_PATH];
if (LCMapStringW(LOCALE_INVARIANT, LCMAP_LOWERCASE, fd.cFileName, -1, w_lower, MAX_PATH) == 0) { ... }
char *name = wide_to_utf8(w_lower);
```

On `L"Aş.txt"` that yields `L"aş.txt"` — `A` lowercases, `ş` is left alone — which is byte-for-byte what the current ASCII-only loop produces. So the assertion below passes with the broken implementation *and* the fixed one, and reintroducing `LCMapStringW` would sail through CI. That is exactly the S1 (F1) found in iteration 1 of this PR.

`İ` (U+0130) is the character from the issue and the one that separates the two: Unicode lowercases it to `i`, NTFS's `$UpCase` table does not, so the old code yields `ai.txt` while the ASCII-only loop yields `aİ.txt`.

```suggestion
    will_return(wrap_FindFirstFile, "A" "\xc4\xb0" ".txt");   // "Aİ.txt" in UTF-8 (U+0130)
```

The expectation at the bottom of the test needs the matching change:

```c
    // 'A' is lowercased to 'a'; U+0130 is preserved byte for byte. Reintroducing
    // LCMapStringW would map U+0130 to 'i' and yield 'ai.txt', failing this check.
    expect_string(__wrap__mdebug2, formatted_msg,
                  "(6319): No configuration found for (file):'test\\a" "\xc4\xb0" ".txt'");
```

src/syscheckd/src/create_db.c
MarcelKemp: **[S2 · non-blocking · introduced]** The comment claims this loop matches the canonicalization used by realtime and whodata; that is true for the case fold but not for the encoding. `run_realtime.c:481` and `get_whodata_path` (`win_whodata.c:1657`) both still convert with `WideCharToMultiByte(CP_ACP, ...)`, so their names arrive mangled before any case folding. Worth not asserting a consistency that does not exist yet, and widening the follow-up issue to whodata as well as realtime.

```suggestion
        // Lowercase ASCII only: keeps the on-disk bytes so stat resolves. Note that
        // realtime (run_realtime.c) and whodata (win_whodata.c) still convert names
        // with CP_ACP, so their names are not consistent with this one yet.
```

src/syscheckd/src/create_db.c
MarcelKemp, line 818: **[S2 · blocking · aggravated]** `PATH_MAX` is 260 *characters* on Windows, but `name` is now UTF-8 instead of single-byte ANSI, so this compares bytes against a character limit. `c:\data\` (8 bytes) plus a 90-character CJK or Cyrillic name is 278 bytes: the file is dropped with `FIM_ERROR_PATH_TOO_LONG` ("The maximum length is 260 characters") even though the real path is 98 characters, and `mwarn` fires once per file per scan. Files that were inventoried on 4.14.8 stop being monitored — the same silent loss of FIM coverage the issue is about. Count in wide characters before converting, and size `f_name` for UTF-8:

```c
    // in fim_directory, before converting: count in wide characters
    size_t w_dir_len = wcslen(wpattern) - 2;   // pattern is dir + PATH_SEP + '*'
    if (w_dir_len + 1 + wcslen(fd.cFileName) >= MAX_PATH) {
        mwarn(FIM_ERROR_PATH_TOO_LONG, dir, "", MAX_PATH);
        continue;
    }
```

If you would rather not change the buffer size in this PR, please say so in the description and open a separate issue.

src/syscheckd/src/create_db.c
MarcelKemp: **[S3 · non-blocking · introduced]** Any `FindNextFileW` failure is treated as end-of-directory and `fim_directory` still returns 0, so a mid-enumeration error (ERROR_ACCESS_DENIED after an ACL change, ERROR_NETNAME_DELETED, ERROR_FILE_CORRUPT) after 3 of 5000 entries makes the caller believe the directory was fully scanned and `fim_db_transaction_deleted_rows` reports the remaining 4997 as removed. `readdir` had the same hole and the only caller discards the return value, so propagating the failure needs separate plumbing — but the error should at least be diagnosable:

```suggestion
    } while (FindNextFileW(hFind, &fd));

    if (GetLastError() != ERROR_NO_MORE_FILES) {
        mwarn(FIM_PATH_NOT_OPEN, dir, win_strerror(GetLastError()));
    }
```

src/syscheckd/src/create_db.c
MarcelKemp: **[S3 · non-blocking · introduced]** Two things here: the message is hardcoded, while every other FIM message is a define with a code in `src/error_messages/debug_messages.h` (e.g. `FIM_CONFIGURATION_NOTFOUND "(6319): ..."`); and `mdebug2` hides it at the default log level, when what it reports is a file missing from the inventory — exactly what the issue is about. Consider a define plus `mwarn`:

```c
/* debug_messages.h */
#define FIM_UTF8_CONVERSION_FAILED "(63XX): Cannot convert a directory entry under '%s' to UTF-8."

/* create_db.c */
            mwarn(FIM_UTF8_CONVERSION_FAILED, dir);
```

src/unit_tests/syscheckd/test_create_db.c
```diff
@@ -3122,17 +3156,61 @@ static void test_fim_directory(void **state) {
...
+    int ret;
+
+    expect_function_call_any(__wrap_pthread_rwlock_wrlock);
+    expect_function_call_any(__wrap_pthread_rwlock_unlock);
+    expect_function_call_any(__wrap_pthread_rwlock_rdlock);
+    expect_function_call_any(__wrap_pthread_mutex_lock);
+    expect_function_call_any(__wrap_pthread_mutex_unlock);
+
+    expect_memory(wrap_FindFirstFile, lpFileName, L"test\\*", (wcslen(L"test\\*") + 1) * sizeof(wchar_t));
+    will_return(wrap_FindFirstFile, "A" "\xc4\x9e" ".txt");   // "AĞ.txt" in UTF-8 (Ğ = U+011E, uppercase)
```
MarcelKemp: **[S3 · RESOLVED]** Fixed iteration 4: test now uses Ğ (U+011E) instead of ş, discriminating correctly between LCMapStringW (→ ğ, test fails) and ASCII-only lowercasing (→ Ğ preserved, test passes). Comment now documents the regression protection.

src/syscheckd/src/create_db.c
Darioortegaleyva: Agreed this is a real regression for long non-ANSI names.

The catch is that the suggested change (count wide chars + size f_name for UTF-8) can't be applied in isolation without introducing a worse bug: it doesn't remove the 260 limit, it moves it downstream into a buffer overflow.

The downstream FIM path handling uses fixed char[PATH_MAX] (260-byte) buffers that hold the UTF-8 path. With report_changes/diff enabled, the scan path flows fim_file_diff() -> initialize_file_diff_data() -> abspath(path, buffer, sizeof(buffer)) where buffer is char[PATH_MAX]; run_check.c and create_db.c share the same 260-byte assumption. Letting the scan emit paths longer than 260 bytes would overflow/truncate there. So this byte-length guard is not behavior introduced by this PR: it's the invariant that keeps the whole downstream chain within its 260-byte buffers.

Supporting names that are under 260 characters but over 260 UTF-8 bytes therefore requires reworking path-length handling to real UTF-8 lengths across scan, diff, run_check, realtime and whodata, which is out of scope for this targeted fix.

As suggested, I documented it under "Scope and known limitations" in the PR description and opened a dedicated follow-up issue (#39240) to track the UTF-8 path-length rework (this length limit plus the realtime/whodata CP_ACP conversion).

MarcelKemp (approved):

PR #38304, opened 2026-08-10
Adjust the agent's shipped configuration and batching limits

jr0me:
## Description

Follow-up adjustments on issue #38163, from the configuration review of the agent's
HTTPS transport. Two groups of changes:

**What the shipped configuration says.** Every option the templates wrote was set to
the value the agent already applies on its own, so the file restated defaults while
reading like a deliberate choice. `<ssl>`, `<notify_time>` and `<auto_restart>` are
gone from the templates and the defaults are now stated once, in code. A fresh install
gets this:

```xml
  <agent>
    <server>
      <address>MANAGER_IP</address>
      <port>1517</port>
    </server>
    <config-profile>ubuntu, ubuntu22</config-profile>
  </agent>
```

**What `<batch>` reaches.** It can now be pushed from the centralized configuration,
and it bounds a stateful sync session as well as a `/stateless` request — that ceiling
was a 5 MiB constant. Where each limit comes from after this PR:

| | size | interval |
|---|---|---|
| `/stateless` | `<agent><batch><size>` — unset: 1 MiB | `<agent><batch><interval>` — unset: 10 s |
| `/stateful` | `<agent><batch><size>` — unset: 5 MiB | the module's own `<synchronization><interval>` — unset: 300 s |

The flush rule (size OR interval) and the `/control` cadence were already what they
should be; they had no test holding them there, and now they do.

## Acceptance criteria

| # | Criterion | Commit |
|---|---|---|
| 1 | `<ssl>` not present by default in the templates or `ossec.conf` | `a37f3e81be` |
| 2 | `verification_mode` defaults to `none` | `bc77ac4bdf` |
| 3 | `auto_restart` not present by default | `2bd4ff57ad` |
| 4 | `notify_time` not present by default | `09f9cb1599` |
| 5 | Batch sent on max size **or** interval | `a52931b446` |
| 6 | Stateful messages: configurable max size (was hardcoded 5 MiB) and interval | `aa46f1a7a5` |
| 7 | `<batch>` configurable through the centralized configuration | `8102fecbb8` |
| 8 | `/control` outside the batch, on its own `notify_time` | `ca16c22a23` |

Evidence for each is in the comment below.

## Notes for review

**The removals had callers.** `register_configure_agent.sh` inserted the enrollment
block after `</auto_restart>` and set `notify_time` by substituting a tag it assumed
was there. Dropping either option silently breaks a deployment variable, so both are
re-anchored on the agent block itself rather than on an option inside it.

**`<batch><size>` has two "off" positions.** Set it and both planes use that number.
Leave it out and each keeps the limit it had before this PR — 1 MiB for a `/stateless`
request, 5 MiB for a sync session — rather than one silently shrinking or growing.
It is capped at 1 GiB, far above any useful batch and inside every `size_t` it has to
survive: past 4 GiB a 32-bit agent wraps, and exactly 4 GiB wraps to zero, which every
reader downstream takes as "unset".

**It also bounds slightly different things on each side.** For `/stateless` it is the
maximum body of one request, and it is adaptive: a 413 halves it and success ramps it
back. For `/stateful` it is the maximum pulled from the persistent queue per block,
less a 64 KiB prefilter grace, and one sync cycle can still send up to
`FULLSESSION_MAX_BLOCKS_PER_SYNC` (10) of those blocks — that multiplier is unchanged
and still a constant.

### DIFF
--- a/etc/ossec-agent.conf
+++ b/etc/ossec-agent.conf
@@ -11,14 +11,6 @@
       <port>1517</port>
     </server>
 
-    <!-- HTTPS transport TLS settings; verification_mode: full|certificate|none (default: none).
-         Set to 'full' or 'certificate' and provide certificate_authorities to verify the
-         manager's certificate. -->
-    <ssl>
-      <!-- <certificate_authorities>etc/certs/root-ca.pem</certificate_authorities> -->
-      <verification_mode>none</verification_mode>
-    </ssl>
-
     <config-profile>debian, debian8</config-profile>
   </agent>
 
--- a/src/client-agent/https_client/include/https_client.h
+++ b/src/client-agent/https_client/include/https_client.h
@@ -138,7 +138,7 @@ typedef struct hc_config_t
     uint32_t buffer_flood_tolerance_s; ///< agent.tolerance: seconds FULL must
     ///< hold before FLOOD; 0 -> 15.
 
-    uint32_t notify_interval_s;         ///< notify_time; 0 -> 20.
+    uint32_t notify_interval_s;         ///< notify_time; 0 -> 10.
     uint32_t rejected_retry_interval_s; ///< Slow re-Startup cadence; 0 -> 60.
 
     /// Safety bound for a remote_upgrade WPK download; 0 -> 200 MiB.
--- a/src/client-agent/https_client/src/httpsClientFacade.cpp
+++ b/src/client-agent/https_client/src/httpsClientFacade.cpp
@@ -24,7 +24,7 @@ namespace
     /// "" when fn is null, so an unset callback degrades to "no extra metadata"
     /// rather than a crash.
     std::function<std::string()> makeMetadataCollector(void (*fn)(char*, size_t, void*),
-                                                        void* userData)
+                                                       void* userData)
     {
         return [fn, userData]() -> std::string
         {
--- a/src/client-agent/https_client/src/moduleConfig.cpp
+++ b/src/client-agent/https_client/src/moduleConfig.cpp
@@ -47,7 +47,7 @@ ModuleConfig ModuleConfig::fromC(const hc_config_t& config)
     typed.bufferWarnLevel = orDefault<uint32_t>(config.buffer_warn_level, 90);
     typed.bufferNormalLevel = orDefault<uint32_t>(config.buffer_normal_level, 70);
     typed.bufferFloodToleranceS = orDefault<uint32_t>(config.buffer_flood_tolerance_s, 15);
-    typed.notifyIntervalS = orDefault<uint32_t>(config.notify_interval_s, 20);
+    typed.notifyIntervalS = orDefault<uint32_t>(config.notify_interval_s, 10);
     typed.rejectedRetryIntervalS = orDefault<uint32_t>(config.rejected_retry_interval_s, 60);
     typed.wpkMaxDownloadBytes = orDefault<uint64_t>(config.wpk_max_download_bytes, 200ULL * 1024 * 1024);
     typed.statsEnabled = config.stats_enabled;
--- a/src/client-agent/https_client/src/moduleConfig.hpp
+++ b/src/client-agent/https_client/src/moduleConfig.hpp
@@ -48,7 +48,7 @@ struct ModuleConfig
         uint32_t bufferNormalLevel {70};
         uint32_t bufferFloodToleranceS {15};
 
-        uint32_t notifyIntervalS {20};
+        uint32_t notifyIntervalS {10};
         uint32_t rejectedRetryIntervalS {60};
 
         /// Safety bound for a remote_upgrade WPK download: stops a
--- a/src/client-agent/https_client/tests/component/facadeE2e_component_test.cpp
+++ b/src/client-agent/https_client/tests/component/facadeE2e_component_test.cpp
@@ -427,6 +427,48 @@ TEST_F(FacadeE2eTest, IntervalFlushesABatchThatNeverReachesTheSize)
     EXPECT_TRUE(waitForBody(peek, "/peek/stateless", "interval-only-flush", 4000));
 }
 
+TEST_F(FacadeE2eTest, ControlKeepsItsNotifyCadenceWhileEventsAreStillBatching)
+{
+    // /control is not part of the batch. The batch window here outlasts the test,
+    // so the submitted event has to still be sitting in the accumulator when the
+    // notifies land -- which is what makes this an observation and not a guess: if
+    // the two planes shared a timer, either the notifies would stall behind the
+    // batch or the event would ride out with one of them.
+    const uint16_t port = TLS_PORT + 11;
+    FakeManager manager {port, KEY_HEX, /*tls=*/true};
+
+    Recorder recorder;
+    hc_config_t config = tlsConfig();
+    config.server_port = port;
+    config.notify_interval_s = 1;
+    config.batch_size_bytes = 1024 * 1024; // No size flush.
+    config.batch_interval_ms = 60000;      // No interval flush while this test runs.
+    hc_callbacks_t callbacks {};
+    callbacks.on_startup_result = onStartup;
+    callbacks.user_data = &recorder;
+
+    hc_handle* handle = hc_create(&config, &callbacks);
+    ASSERT_NE(nullptr, handle);
+    const HandleGuard guard {handle};
+    ASSERT_TRUE(hc_start(handle));
+    ASSERT_TRUE(waitFor(recorder.startupCount, 1, 3000));
+
+    const std::string event = "1:/loc:held-in-the-batch";
+    ASSERT_TRUE(hc_submit_event(handle, reinterpret_cast<const uint8_t*>(event.data()),
+                                event.size()));
+
+    httplib::Client peek {std::string {"https://127.0.0.1:"} + std::to_string(port)};
+    peek.enable_server_certificate_verification(false);
+    const int notifiesBefore = peekCount(peek, "/peek/notifies");
+
+    // Three notify_time periods' worth of notifies, none of them waiting on the batch.
+    EXPECT_TRUE(waitForCount(peek, "/peek/notifies", notifiesBefore + 3, 8000));
+
+    const auto stateless = peek.Get("/peek/stateless");
+    ASSERT_TRUE(static_cast<bool>(stateless));
+    EXPECT_EQ(std::string::npos, stateless->body.find("held-in-the-batch"));
+}
+
 TEST_F(FacadeE2eTest, LargeSyncSessionFromTheIntakeSocketReachesTheManager)
 {
     // The full chain: a producer streams a multi-MB session over the local
--- a/src/client-agent/https_client/tests/component/facadeE2e_component_test.cpp
+++ b/src/client-agent/https_client/tests/component/facadeE2e_component_test.cpp
@@ -395,6 +395,38 @@ TEST_F(FacadeE2eTest, SizeThresholdFlushesBeforeTheBatchInterval)
     EXPECT_NE(std::string::npos, received.find("size-wakeup-first"));
 }
 
+TEST_F(FacadeE2eTest, IntervalFlushesABatchThatNeverReachesTheSize)
+{
+    // The companion to SizeThresholdFlushesBeforeTheBatchInterval: the same OR,
+    // seen from the other side. One short event against a 1 MiB limit can only
+    // leave when the interval closes the window.
+    const uint16_t port = TLS_PORT + 10;
+    FakeManager manager {port, KEY_HEX, /*tls=*/true};
+
+    Recorder recorder;
+    hc_config_t config = tlsConfig();
+    config.server_port = port;
+    config.batch_size_bytes = 1024 * 1024; // Far above anything this test sends.
+    config.batch_interval_ms = 300;
+    hc_callbacks_t callbacks {};
+    callbacks.on_startup_result = onStartup;
+    callbacks.user_data = &recorder;
+
+    hc_handle* handle = hc_create(&config, &callbacks);
+    ASSERT_NE(nullptr, handle);
+    const HandleGuard guard {handle};
+    ASSERT_TRUE(hc_start(handle));
+    ASSERT_TRUE(waitFor(recorder.startupCount, 1, 3000));
+
+    const std::string event = "1:/loc:interval-only-flush";
+    ASSERT_TRUE(hc_submit_event(handle, reinterpret_cast<const uint8_t*>(event.data()),
+                                event.size()));
+
+    httplib::Client peek {std::string {"https://127.0.0.1:"} + std::to_string(port)};
+    peek.enable_server_certificate_verification(false);
+    EXPECT_TRUE(waitForBody(peek, "/peek/stateless", "interval-only-flush", 4000));
+}
+
 TEST_F(FacadeE2eTest, LargeSyncSessionFromTheIntakeSocketReachesTheManager)
 {
     // The full chain: a producer streams a multi-MB session over the local
--- a/src/client-agent/https_client/tests/unit/moduleConfig_test.cpp
+++ b/src/client-agent/https_client/tests/unit/moduleConfig_test.cpp
@@ -41,7 +41,7 @@ TEST(ModuleConfigTest, DefaultsAppliedOnZeroFields)
     EXPECT_EQ(1024u * 1024u, typed.batchSizeBytes);
     EXPECT_EQ(10000u, typed.batchIntervalMs);
     EXPECT_EQ(4u, typed.bufferCapMultiplier);
-    EXPECT_EQ(20u, typed.notifyIntervalS);
+    EXPECT_EQ(10u, typed.notifyIntervalS);
     EXPECT_EQ(60u, typed.rejectedRetryIntervalS);
     EXPECT_EQ(10000u, typed.requestTimeoutMs);
     EXPECT_EQ(120000u, typed.statefulTimeoutMs);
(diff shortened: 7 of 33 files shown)


jr0me:
## Evidence

Produced from this branch, built in a container from the branch tree
(`make TARGET=agent DEBUG=1 TEST=1` for the suites, a plain `DEBUG=1` build for the
daemons). Runtime cases drive the real `wazuh-agentd` / `wazuh-syscheckd` reading
`ossec.conf`, so the XML parser is in the loop.

### What a new install gets (criteria 1, 3, 4)

Real generator, not a hand-written sample:

```
$ sh src/init/gen_wazuh.sh conf agent ubuntu 22 /var/ossec
<ossec_config>
  <agent>
    <server>
      <address>MANAGER_IP</address>
      <port>1517</port>
    </server>
    <config-profile>ubuntu, ubuntu22</config-profile>
  </agent>
```

No `<ssl>`, no `<notify_time>`, no `<auto_restart>`.

### The agent runs on that file (criteria 1, 2)

```
==============================================================
CASE: shipped template: no <ssl>, no <notify_time>, no <auto_restart>
--------------------------------------------------------------
  <agent> block as configured:
      <agent>
        <server>
          <address>127.0.0.1</address>
          <port>1517</port>
        </server>
        <config-profile>ubuntu, ubuntu22</config-profile>
      </agent>
  --- wazuh-agentd -t output (verbatim) ---
  2026/08/10 14:57:08 wazuh-agentd[235191] debug_op.c:305 at os_logging_config(): DEBUG: (1228): Element 'log_format' without any option.
  2026/08/10 14:57:08 wazuh-agentd[235191] debug_op.c:116 at _log_function(): DEBUG: Logging module auto-initialized
  2026/08/10 14:57:08 wazuh-agentd[235191] main.c:181 at main(): DEBUG: Wazuh home directory: /var/ossec
  2026/08/10 14:57:08 wazuh-agentd[235191] main.c:183 at main(): DEBUG: Started (pid: 235191).
  exit=0
```

Exit 0 is the criterion-2 result, not just a clean start. There is no CA in that file.
Had the default been the enum's zero value (`AGENT_VERIFY_FULL`) rather than `none`,
this would have stopped on 4118 — which is exactly what the next case shows happening
when a verifying mode really is asked for:

```
==============================================================
CASE: explicit verification_mode=full with an unreadable CA
--------------------------------------------------------------
  <agent> block as configured:
      <agent>
        <server>
          <address>127.0.0.1</address>
          <port>1517</port>
        </server>
        <ssl>
          <certificate_authorities>PATH</certificate_authorities>
          <verification_mode>full</verification_mode>
        </ssl>
      </agent>
  --- wazuh-agentd -t output (verbatim) ---
  2026/08/10 14:57:08 wazuh-agentd[235196] debug_op.c:305 at os_logging_config(): DEBUG: (1228): Element 'log_format' without any option.
  2026/08/10 14:57:08 wazuh-agentd[235196] debug_op.c:116 at _log_function(): DEBUG: Logging module auto-initialized
  2026/08/10 14:57:08 wazuh-agentd[235196] main.c:181 at main(): DEBUG: Wazuh home directory: /var/ossec
  2026/08/10 14:57:08 wazuh-agentd[235196] main.c:183 at main(): DEBUG: Started (pid: 235196).
  2026/08/10 14:57:08 wazuh-agentd[235196] config.c:115 at w_agent_validate_ssl_ca(): ERROR: (4118): <ssl><verification_mode> is not 'none' but <certificate_authorities> is missing or unreadable: 'PATH'.
  2026/08/10 14:57:08 wazuh-agentd[235196] main.c:210 at main(): ERROR: (1215): No client configured. Exiting.
  exit=1
```

Moving the default did not weaken the fail-closed rule; it only stopped the agent
inventing a posture nobody asked for.

### notify_time really is 10 with nothing in the file (criterion 4)

Same config as case 1, run as a daemon rather than `-t` so it reaches the value it uses:

```
2026/08/10 14:57:35 wazuh-agentd[235232] agentd.c:68 at AgentdStart(): DEBUG: Using notify time: 10
```

### <agent><batch><size> reaches the sync protocol (criterion 6)

`<batch><size>2MB</size>` in the local file, read by `wazuh-syscheckd` at startup:

```
==============================================================
CASE: <agent><batch><size>2MB</size> reaching the sync protocol
--------------------------------------------------------------
  --- wazuh-syscheckd -t -dd output, filtered to the sync limit ---
  2026/08/10 14:57:08 wazuh-syscheckd[235199] main.c:298 at main(): DEBUG: Sync sessions bounded to 2097152 bytes by <agent><batch><size>.
```

2097152 = 2 MiB, where the constant used to make it 5 MiB whatever the file said.

### Tests

The cases added or touched here, by criterion:

```
### config (cmocka) — the cases added here
[ RUN      ] test_ssl_zero_initialized_reads_as_full
[       OK ] test_ssl_zero_initialized_reads_as_full
[ RUN      ] test_ssl_absent_keeps_the_default_the_caller_set
[       OK ] test_ssl_absent_keeps_the_default_the_caller_set
[ RUN      ] test_read_agent_batch_takes_the_limits_from_the_file
[       OK ] test_read_agent_batch_takes_the_limits_from_the_file
[ RUN      ] test_read_agent_batch_leaves_the_caller_defaults_when_absent
[       OK ] test_read_agent_batch_leaves_the_caller_defaults_when_absent
[ RUN      ] test_read_agent_batch_survives_a_missing_file
[       OK ] test_read_agent_batch_survives_a_missing_file
[ RUN      ] test_shared_batch_is_parsed
[       OK ] test_shared_batch_is_parsed
[ RUN      ] test_shared_batch_is_validated_like_a_local_one
[       OK ] test_shared_batch_is_validated_like_a_local_one
[ RUN      ] test_shared_config_still_refuses_local_only_options
[       OK ] test_shared_config_still_refuses_local_only_options
[  PASSED  ] 47 test(s).

### sync protocol — session ceiling
[       OK ] AgentSyncProtocolTest.SynchronizeModuleDeltaUsesBytePrefilterBudgetForSyncOption (2 ms)
[       OK ] AgentSyncProtocolTest.SessionByteBudgetFollowsTheConfiguredSessionMax (0 ms)
[       OK ] AgentSyncProtocolTest.AnUnsetSessionMaxKeepsTheBuiltInDefault (0 ms)
[       OK ] AgentSyncProtocolTest.SynchronizeModuleDeltaBypassesBytePrefilterBudgetForVDOptions (4 ms)
[  PASSED  ] 4 tests.

### https_client unit — the batch flush rule
[       OK ] EventAccumulatorTest.FlushDueOnSizeThreshold (0 ms)
[       OK ] EventAccumulatorTest.FlushDueOnAgeThreshold (0 ms)
[       OK ] EventAccumulatorTest.EmptyBufferIsNeverFlushDue (0 ms)
[       OK ] ModuleConfigTest.DefaultsAppliedOnZeroFields (0 ms)
[       OK ] StatelessStreamTest.FlushOnSizeThresholdSendsHeaderAndEvents (2 ms)
[       OK ] StatelessStreamTest.FlushesOnTheIntervalWithoutEverReachingTheSize (0 ms)
[  PASSED  ] 6 tests.

### https_client component — batch and control planes
[       OK ] FacadeE2eTest.SizeThresholdFlushesBeforeTheBatchInterval (250 ms)
[       OK ] FacadeE2eTest.IntervalFlushesABatchThatNeverReachesTheSize (569 ms)
[       OK ] FacadeE2eTest.ControlKeepsItsNotifyCadenceWhileEventsAreStillBatching (3380 ms)
[  PASSED  ] 3 tests.
```

Reading the last two blocks against criteria 5 and 8:

- **5** is an OR, so both arms need a test. The size arm had two; the interval arm had
  none until now. The unit test also pins the negative — it expects exactly one request,
  so a flush before the interval elapses fails it.
- **8** gives the batch a 60 s window and a 1 MiB size, then submits one event and waits.
  The notifies that land while that event is still buffered can only be coming from a
  cadence of their own; the test then asserts the event is still unsent.

Whole suites on this branch:

```
100% tests passed, 0 tests failed out of 107      # cmocka (src/unit_tests)
[  PASSED  ] 303 tests.                           # https_client unit
[  PASSED  ] 38 tests.                            # https_client component
[  PASSED  ] 177 tests.                           # agent_sync_protocol
```

### Not covered here

- **winagent is not compiled.** No mingw toolchain was available. Three files changed
  have Windows paths: `win_utils.c` (the removed duplicate default, and the new
  `asp_set_session_max_bytes` call), `src/win32/ossec.conf`, and the shared
  `client-config.c` / sync protocol. CI is the first real compile.
- **macOS**, same as the rest of this stack.
- **The install scripts are read, not run.** `inst-functions.sh` is exercised through
  `gen_wazuh.sh` above, but `register_configure_agent.sh` needs a real package install
  to drive `WAZUH_KEEP_ALIVE_INTERVAL` and the `WAZUH_REGISTRATION_*` variables end to
  end. Both changed paths are the ones that would silently stop working, so they are
  worth a look during package testing.

jr0me:
## Evidence — review round

Eight findings from a review pass, all fixed on this branch. Same build setup as the
first evidence comment.

### The install script, run twice against the template this PR ships

The shipped block, then the script run with `WAZUH_MANAGER`, `WAZUH_KEEP_ALIVE_INTERVAL`
and `WAZUH_REGISTRATION_*` set — twice, on the file it produced the first time:

```
==============================================================
SHIPPED TEMPLATE (agent block)
--------------------------------------------------------------
  <agent>
    <server>
      <address>MANAGER_IP</address>
      <port>1517</port>
    </server>
    <config-profile>ubuntu, ubuntu22</config-profile>
  </agent>

RUN 1: WAZUH_MANAGER=10.0.0.9 WAZUH_KEEP_ALIVE_INTERVAL=45 WAZUH_AGENT_NAME=probe
--------------------------------------------------------------
  <agent>
    <notify_time>45</notify_time>
    <enrollment>
      <enabled>yes</enabled>
      <manager_address>10.0.0.9</manager_address>
      <agent_name>probe</agent_name>
      <authorization_pass_path>etc/authd.pass</authorization_pass_path>
    </enrollment>
    <server>
      <address>10.0.0.9</address>
      <port>1517</port>
    </server>
    <config-profile>ubuntu, ubuntu22</config-profile>
  </agent>
  enrollment blocks=1 notify_time tags=1
  xmllint: well-formed

RUN 2: WAZUH_MANAGER=10.0.0.9 WAZUH_KEEP_ALIVE_INTERVAL=45 WAZUH_AGENT_NAME=probe
--------------------------------------------------------------
  <agent>
    <enrollment>
      <enabled>yes</enabled>
      <manager_address>10.0.0.9</manager_address>
      <agent_name>probe</agent_name>
      <authorization_pass_path>etc/authd.pass</authorization_pass_path>
    </enrollment>
    <server>
      <address>10.0.0.9</address>
      <port>1517</port>
    </server>
    <notify_time>45</notify_time>
    <config-profile>ubuntu, ubuntu22</config-profile>
  </agent>
  enrollment blocks=1 notify_time tags=1
  xmllint: well-formed
```

Run 1 is `WAZUH_KEEP_ALIVE_INTERVAL` working with no tag in the file to substitute.
Run 2 is the part that was broken before: still one enrollment block and one
`notify_time`, where the old code left the original block in place and added the new
one's children beside it.

### A commented-out agent block no longer takes the insertion

Commenting the block out is how it gets disabled, and the commented copy comes first
in the file, so a whole-file text match landed everything inside the comment:

```
==============================================================
SCOPING: a commented-out <agent> block ahead of the real one
--------------------------------------------------------------
<ossec_config>
  <!--
  <agent>
    <notify_time>99</notify_time>
  </agent>
  -->
  <agent>
    <notify_time>45</notify_time>
    <server>
      <address>10.0.0.9</address>
      <port>1517</port>
    </server>
  </agent>
</ossec_config>
  xmllint: well-formed
```

The commented block is untouched and the real one got the option. Restricting the
match to the opening tag alone on its line was not enough on its own — the first
attempt still matched the commented copy, because it is also alone on its line. The
matcher tracks comment state.

### Tests

```
[ RUN      ] test_agent_block_reads_the_legacy_client_options
[       OK ] test_agent_block_reads_the_legacy_client_options
[ RUN      ] test_fresh_install_template_shape
[       OK ] test_fresh_install_template_shape
[ RUN      ] test_read_agent_batch_takes_the_limits_from_the_file
[       OK ] test_read_agent_batch_takes_the_limits_from_the_file
[ RUN      ] test_read_agent_batch_leaves_the_caller_defaults_when_absent
[       OK ] test_read_agent_batch_leaves_the_caller_defaults_when_absent
[ RUN      ] test_read_agent_batch_applies_the_local_file_with_no_shared_one
[       OK ] test_read_agent_batch_applies_the_local_file_with_no_shared_one
[ RUN      ] test_read_agent_batch_applies_nothing_when_the_block_is_rejected
[       OK ] test_read_agent_batch_applies_nothing_when_the_block_is_rejected
[ RUN      ] test_read_agent_batch_survives_a_missing_file
[       OK ] test_read_agent_batch_survives_a_missing_file
[ RUN      ] test_batch_size_beyond_the_cap_is_rejected
[       OK ] test_batch_size_beyond_the_cap_is_rejected
[  PASSED  ] 51 test(s).
```

`test_read_agent_batch_applies_the_local_file_with_no_shared_one` covers a regression
introduced while fixing the partial-application one: rejecting a failed read of the
centralized file also threw away the local block, and `ReadConfig` cannot tell a file
that is absent from one that is invalid. Most agents have never been pushed an
`agent.conf`, so that would have been the common path.

### Still not covered

- **winagent is still not compiled.** Both Windows fixes (the `InstallerScripts.vbs`
  insert and the `asp_set_session_max_bytes` move) are static changes verified by
  reading; CI is the first compile.
- **The MSI path is not run.** The `InstallerScripts.vbs` change is the Windows
  counterpart of the shell fix evidenced above, but only the shell half was executed.

jr0me:
Both confirmed and fixed. Three commits: `7671e20eb4`, `ab8527ef55`, `35736a15ad`.

## 1. `register_configure_agent.sh` — the sed range truncating to EOF

Reproduced exactly as described, and it is a regression from `7b6bd14e2f` in this PR, which
introduced that deletion:

```
$ sed -ri "/<enrollment>/,/<\/enrollment>/d" after.conf
--- before: 10 lines ---
--- after: 3 lines ---
<ossec_config>
  <agent>
    <server><address>MANAGER_IP</address></server>
```

`</agent>`, the whole `<rootcheck>` block and `</ossec_config>` are gone. The `grep -n` line
numbers above it have the mirror problem: with both tags on one line, start equals end, so
`sed -n "$((start+1)),$((end-1))p"` is an inverted range and copies out the wrong line.

Replaced with a single awk pass that takes the block out and collects its children in one go. It
handles both layouts, skips commented-out blocks the way `insert_into_agent_block` already does,
drops a second block instead of leaving it behind, and treats an unterminated block as unusable —
the copy is discarded and the original file is left alone rather than rewritten from a half-read
one.

The script had no tests and this is the second defect in the same function, so it now has a
harness (`src/init/tests/test_register_configure_agent.sh`) that runs the real script against
throwaway `INSTALLDIR`s and asserts that nothing outside the enrollment block moves:

```
$ bash src/init/tests/test_register_configure_agent.sh
ok   - a single-line enrollment block keeps the rest of the file
ok   - a multi-line enrollment block keeps the rest of the file
ok   - SINGLE_LINE_CONF: exactly one <enrollment> opening tag
ok   - SINGLE_LINE_CONF: the configured registration server is written
ok   - MULTILINE_CONF: exactly one <enrollment> opening tag
ok   - MULTILINE_CONF: the configured registration server is written
ok   - NO_ENROLLMENT_CONF: exactly one <enrollment> opening tag
ok   - NO_ENROLLMENT_CONF: the configured registration server is written
ok   - a second run does not add a second block
ok   - a second run keeps the rest of the file
ok   - the commented-out block is left alone
ok   - the commented-out block does not absorb the insertion

12 checks, 0 failed
```

Against the pre-fix script the first case fails with exactly the reported damage:

```
FAIL - a single-line enrollment block keeps the rest of the file
--- actual ---
    <config-profile>CONFIG_PROFILE</config-profile>
  </agent>
  <rootcheck>
    <disabled>no</disabled>
  </rootcheck>
</ossec_config>
```

Run on Linux (gawk) and natively on macOS (BSD awk 20200816 and BSD sed), since the macOS agent
runs this script too. The `exit` in the `END` block is spelled out rather than using a ternary for
that reason.

## 2. `w_read_agent_batch()` — the local block discarded on a shared parse failure

Correct on both counts, including the divergence from `ClientConf()`. The two files are now parsed
into separate structures and the local block is applied as soon as it comes back clean, so:

- a valid local `<batch>` survives a centralized file the agent cannot read, which is what agentd
  already does with the same pair;
- a rejected centralized block still cannot be half-applied on top of it, which was the reason the
  original all-or-nothing return existed.

On the missing log: `ReadConfig` is silent here by construction — its `XML_ERROR` for a centralized
read is inside `#ifndef CLIENT` (`config.c:283-289`), so on the agent nothing at all was printed. A
warning naming the file now goes with the fix.

New unit test `test_read_agent_batch_keeps_the_local_block_when_the_shared_one_is_rejected`. It
breaks the shared file at the XML level rather than with an invalid element, because a rejected
element sits behind `ReadConfig`'s `agent_config` profile gate, which wants a real
`queue/sockets/.agent_info` to get past; both routes end in `ReadConfig` returning `OS_INVALID`,
which is all this function reacts to. Reverting the fix makes it fail.

### A leak this uncovered

`ReadConfig`'s two early exits — the failed read and the empty document — never clear the XML
object, while every other exit in the function does, and `read_local_agent_batch` right next door
clears on exactly the same failure. No existing test hands `ReadConfig` a file it rejects, so it had
never surfaced; the new test does, and ASan stops the whole config binary on it. Fixed in
`ab8527ef55`.

## Verification

| Check | Result |
|---|---|
| `make TARGET=agent DEBUG=1 TEST=1` | pass |
| `make TARGET=winagent TEST=1` (full) | pass, 0 errors |
| `ctest` on `src/unit_tests` (agent) | 107/107 pass |
| `test_client-config_https` | 52/52 pass, ASan clean (exit 0) |
| `test_client-config_https.exe` (winagent) | builds and links |
| `test_register_configure_agent.sh`, Linux + macOS | 12/12 pass |

Both new tests were negative-checked against the pre-fix code and fail there.

Nicogp (approved):

PR #38312, opened 2026-08-10
Fix Windows agent crash on reconnection when the manager rejects its key

### LINKED ISSUE #38203: Flaky Windows agentd IT test
## Description

CI for PR #38188 (migrating agent workflows to GitHub-hosted runners) failed on the Windows agentd integration tests (tier-0-1), unrelated to that PR's changes (CI workflow/runner config only):

```
FAILED test_agentd/test_reconnection/test_agentd_reconection_enrollment_with_keys.py::test_agentd_reconection_enrollment_with_keys[TCP protocol] - AssertionError: Enrollment retry was not sent 1 failed, 24 passed, 6 deselected
```

Job: https://github.com/wazuh/wazuh/actions/runs/31017525141/job/92351224674?pr=38188

## Likely regression

This exact test was previously flagged unstable and fixed in #31676 ("Fix unstable agentd IT on Windows and enable them", closed 2025-09-06). The test file itself hasn't been touched since the fix landed (`7c9a2f55bd`, 2025-09-03) except for the daemon-rename commit `21a1e45fb1` ("separate co-host responsibilities and rename daemons",
2026-02-04). The flake resurfacing without a test-side edit suggests a source-side regression (agentd/authd/remoted timing) rather than a test bug.

## To investigate

- Re-run to confirm reproducibility vs. one-off
- Revisit the timing sensitivity called out in #31676's last update   (personalized-message timing around enrollment retry)
- Check whether the co-host/daemon-rename change (`21a1e45fb1`) or later agentd/authd changes affected retry timing

Related: #31676

anromerom:
## Description

Closes #38203

The Windows agent crashes with an access violation inside `syscollector.dll`, taking the whole process down. When it happens during reconnection the agent stops retrying, never re-enrolls, and the reported integration test times out.

```sql
Faulting application name: wazuh-agent.exe, version: 5.0.0.0
Faulting module name: syscollector.dll, version: 5.0.0.0
Exception code: 0xc0000005
Fault offset: 0x00012313
```

The fault resolves to `Syscollector::getMetadataValue()`, at the point where it dereferences `m_spDBSync`:

```asm
12310:  8b 48 6c    mov 0x6c(%eax),%ecx    ; ecx = this->m_spDBSync
12313:  8b 01       mov (%ecx),%eax        ; faults here, reading the vptr through a null pointer
12330:  ff d0       call *%eax             ; virtual call, selectRows()
```

`Syscollector::query()` is reachable from threads the module never owns: wcom's dispatcher is detached and never joined, and agent-info's coordination polls call it in-process from their own module thread. The `get_first_sync_completed` query reaches `getMetadataValue()`, which used `m_spDBSync` with no null check and no synchronization. `m_spDBSync` can be null in two distinct situations, and both were reachable:

- **Initialization failed.** `init()` throws while constructing DBSync (for example `sqlite: database is locked` when a previous agent process still holds `local.db`). `m_spDBSync` stays null and `start()` bails out with "Cannot start Syscollector - module initialization failed", but nothing gates `query()` on `m_initialized`, so the next coordination poll dereferences null. This is the path behind the crashes actually captured: all five had `module initialization failed` in the log.
- **Teardown.** `releaseResources()` resets `m_spDBSync` and the sync protocols with no lock, relying on the scan and sync workers having been joined. That covers the module's own threads, not the query entry point, so an in-flight query can be inside a check-then-use while the members are reset.

This is the same defect class that df06f36bec fixed in SCA ("fix(sca): exclude live query paths from the module teardown"), whose description names `get_first_sync_completed` explicitly. That fix was never mirrored into Syscollector, although both modules share the same structure, the same teardown ordering and the same entry points. `git log -S m_resourcesMutex` returns that single SCA commit and nothing under `src/wazuh_modules/syscollector`.

The query path became reachable in 5323683e80 ("feat(syscollector): persist first sync completion"), which added `getMetadataValue()` together with the `get_first_sync_completed` command.

Reported as a flaky Windows integration test, because the visible symptom is `test_agentd_reconection_enrollment_with_keys` timing out: the agent logs one failed handshake attempt and then dies, so the remaining retries and the re-enrollment never happen.

```sql
09:06:43 wazuh-agent: INFO: (4102): Connected to the server ([127.0.0.1]:1514/tcp).
09:06:44 wazuh-agent: ERROR: (1137): Lost connection with manager. Setting lock.
09:06:44 wazuh-agent: INFO: Trying to connect to server ([127.0.0.1]:1514/tcp).
09:06:44 wazuh-agent: WARNING: (1214): Problem receiving message from '127.0.0.1'.
```

Nothing follows for the remaining 150 seconds, because there is no process left to write it.

## Proposed Changes

**Guard the pointer.**

- Check `m_spDBSync` in `getMetadataValue()` before use, as the other twelve DBSync users in the file already do. This is what stops the crash actually observed: a null dereference raises an SEH access violation on Windows, which the surrounding `catch (const std::exception&)` cannot handle, so the process dies rather than the error being swallowed.

**Serialize the live entry points against teardown.**

- Add `m_resourcesMutex` (`std::shared_mutex`), mirroring `SecurityConfigurationAssessment::m_resourcesMutex`.
- Take it **shared** in the entry points driven by threads that outlive the teardown: `query`, `persistDifference`, `parseResponseBuffer`, `parseResponseBufferVD`, `notifyDataClean`, `deleteDatabase`, `notifyDisableCollectorsDataClean`, `deleteDisableCollectorsData`.
- Take it **exclusive** in `releaseResources()`, so members are reset only when no entry point is inside them, and in `initSyncProtocol()`, so a concurrent reader sees a fully constructed protocol or none.

**Wait for any in-flight flush before resetting what it uses.**

- `releaseResources()` now joins the asynchronous flush worker, under the exclusive lock, before the sync protocols and DBSync are reset.
- Without this the flush path stayed open: a wcom `flush` query calls `startFlush()`, which spawns its own thread and returns immediately, so `query()` releases the shared lock while `executeFlushSync()` keeps dereferencing `m_spSyncProtocol` and `m_spSyncProtocolVD` under no lock. `quiesce()`'s `waitForFlushToFinish()` only joins a worker that already exists when it runs, and `flush()` has no `m_stopping` check, so a flush accepted just afterwards outlives it.
- Holding the exclusive lock across the join also prevents a **new** flush being accepted, because `flush()` is only reachable through `query()`, which takes the lock shared. Joining while holding it is safe: the worker takes no lock and never re-enters a Syscollector entry point.
- The controller is joined rather than destroyed. It is created once in the constructor of this singleton, so destroying it would leave the module unable to flush for the rest of the process lifetime. An earlier revision of this PR did destroy it, which broke every subsequent flush in the unit-test binary and hung `destroyWaitsForOngoingFlush`; that is fixed here.

**Avoid a nested shared lock.**

- `deleteDisableCollectorsData()` holds `m_resourcesMutex` shared for its whole body and called `deleteDatabase()`, which acquired a second shared lock on the same thread. `std::shared_mutex` blocks new readers once a writer is queued, so with `releaseResources()` or `initSyncProtocol()` waiting, the nested acquisition would deadlock against the outer lock this thread still holds. The body is split into `deleteDatabaseUnlocked()`, called from the path that already holds the mutex.

Ordering is unchanged and safe: `quiesce()` still runs before `releaseResources()` in both callers (`destroy()`, and `syscollector_stop()` followed by `syscollector_release_resources()` in `wm_syscollector.c`). It sets `m_stopping`, notifies the condition variables and stops both sync protocols, so anything holding the shared lock is already unwinding before the exclusive lock is requested.

No lock-order inversion is introduced. `query()` acquires no other mutex, `initSyncProtocol()` is never called internally, and `destroy()` releases `m_scan_mutex` before calling `releaseResources()`. Every entry point was traced for nested acquisition; the only real case was the one fixed above.

### Results and Evidence

Reproduced and validated on Windows 11 with the 5.0.0 agent, running `test_agentd_reconection_enrollment_with_keys` in a loop against the test simulators.

| | Before the fix | After the fix |
|---|---|---|
| Runs | 30 | 100 |
| Test failures caused by the crash | 5 | 0 |
| `syscollector.dll` access violations in the Windows event log | 5 | 0 |
| Failure rate | ~17 % | 0 % |

Before the fix, every failing run left one Event ID 1000 in the Windows Application log, always the same module, exception code and fault offset, with timestamps matching the test failures to the second. It reproduced at `windows.debug` 0 and 2 alike, so it was not an artefact of diagnostic logging.

| Crash time | `windows.debug` | Test |
|------------|-----------------|------|
| 2026-08-06 16:25:36 | 2 | fail |
| 2026-08-10 08:46:27 | 2 | fail |
| 2026-08-10 08:57:03 | 0 | fail |
| 2026-08-10 09:06:44 | 0 | fail |
| 2026-08-10 09:11:51 | 0 | fail |

After installing an agent built from this branch, 100 consecutive runs produced no `syscollector.dll` crash. If the crash rate were unchanged, observing zero in 100 runs would have a probability of roughly 1 in 100 million.

The single test failure in those 100 runs was unrelated to this change: the first iteration after a fresh install failed on `Connected to the server message not found`, and the agent's own log shows it connecting normally after two `(1216)` retries while the simulator was not yet listening. `wait_connect()` uses `only_new_events=True`, so the connect line fell outside its monitoring window. That is a race in the test harness, not in the agent.

A separate crash signature, `libstdc++-6.dll` with exception `0x40000015`, also appears in the same window. It is a different defect, is not touched by this change, and failed no test.

An agent package built from the final commit on this branch was additionally installed on Windows 11 against a real manager and exercised end to end: inventory (packages, processes, hotfixes and the OS record, with the host name and counts cross-checked against the machine itself), policy assessment (482 checks evaluated) and file integrity monitoring (a monitored file created and modified, and the change detected). No regression in any of them.

The full syscollector unit-test binary passes locally, 88 tests, including the pre-existing flush and teardown tests. All CI checks on the branch pass, among them `rtr (wazuh_modules/syscollector)` for both `agent` and `winagent`, which an earlier revision of this PR had been stalling to their one-hour limit.

### Artifacts Affected

- `syscollector.dll` on all platforms. The crash was only observed on Windows, but the defect is not platform specific: on Linux and macOS syscollector runs inside `wazuh-modulesd` as a separate process, so the same fault kills that process while `wazuh-agentd` keeps running. The agent then looks healthy while inventory, policy assessment and the other wodles stop silently, which is why the Windows-only integration test was the one that caught it.
- No changes to configuration, on-disk databases, message formats or public interfaces.

### Configuration Changes

None. No new options, no changed defaults, no compatibility considerations.

### Documentation Updates

None required.

### Tests Introduced

- `releaseResourcesJoinsInFlightFlushBeforeResettingSyncProtocol` in `syscollectorImp_test.cpp`. It parks a flush worker inside the sync protocol using a blocking MQ callback, calls `releaseResources()` on another thread, and asserts both that it does not return while the worker is parked and that the worker had left the protocol before it returned. Without the fix `releaseResources()` returns immediately and the test fails.

The existing integration test `test_agentd/test_reconnection/test_agentd_reconection_enrollment_with_keys.py` already covers the reported path and is what surfaced the crash.

A unit test for the null-pointer path would need to drive `query()` against a module whose `init()` failed, which the current fixtures do not model; it is covered by the guard matching the file's existing idiom and by the 100-run validation above.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
+++ b/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
@@ -756,6 +756,21 @@ void Syscollector::quiesce()
 
 void Syscollector::releaseResources()
 {
+    // Take the flush controller out under the exclusive lock (so a concurrent "flush" query
+    // sees either the live controller or none) but destroy it outside of it: its destructor
+    // joins the flush worker, which runs executeFlushSync() -> synchronizeModule() without
+    // taking m_resourcesMutex, so joining it while holding the lock would be pointless and
+    // resetting the protocols before the join would be a use-after-free. quiesce()'s
+    // waitForFlushToFinish() is not enough on its own: it only joins the worker that exists
+    // at that instant, and flush() has no m_stopping check, so a "flush" query accepted
+    // right afterwards spawns a worker that outlives it.
+    std::unique_ptr<Utils::AsyncFlushController> flushController;
+    {
+        std::unique_lock<std::shared_mutex> resourcesLock(m_resourcesMutex);
+        flushController = std::move(m_asyncFlushController);
+    }
+    flushController.reset();
+
     // Exclude the entry points that other threads keep driving while the module tears down.
     // The scan and sync workers are joined before this runs (see wm_syscollector.c), but
     // query() is not: wcom's dispatcher is detached and agent-info polls it in-process, so
@@ -2902,7 +2917,17 @@ void Syscollector::deleteDatabase()
     // outlive the module teardown (wcom's detached dispatcher and agent-info's in-process
     // coordination polls), so the members it reaches must not be reset underneath it.
     std::shared_lock<std::shared_mutex> resourcesLock(m_resourcesMutex);
+    deleteDatabaseUnlocked();
+}
 
+void Syscollector::deleteDatabaseUnlocked()
+{
+    // The caller must already hold m_resourcesMutex. This body is split out of
+    // deleteDatabase() so deleteDisableCollectorsData(), which holds the mutex shared for
+    // its whole body, does not acquire a second shared lock on the same thread: most
+    // std::shared_mutex implementations block new readers once a writer is queued, so the
+    // nested acquisition would wait behind a pending releaseResources() or
+    // initSyncProtocol() while still holding the outer lock those writers are waiting for.
     if (m_spSyncProtocol)
     {
         m_spSyncProtocol->deleteDatabase();
@@ -4392,7 +4417,9 @@ void Syscollector::deleteDisableCollectorsData()
             m_logFunction(LOG_INFO, "All collectors are disabled. Deleting entire database.");
         }
 
-        deleteDatabase();
+        // Already holding m_resourcesMutex shared; call the non-locking body so we do not
+        // take a second shared lock on this thread (see deleteDatabaseUnlocked()).
+        deleteDatabaseUnlocked();
         m_disabledCollectorsIndicesWithData.clear();
         return;
         // LCOV_EXCL_STOP
--- a/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
+++ b/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
@@ -756,6 +756,15 @@ void Syscollector::quiesce()
 
 void Syscollector::releaseResources()
 {
+    // Exclude the entry points that other threads keep driving while the module tears down.
+    // The scan and sync workers are joined before this runs (see wm_syscollector.c), but
+    // query() is not: wcom's dispatcher is detached and agent-info polls it in-process, so
+    // without this lock a get_first_sync_completed query could be inside getMetadataValue()
+    // dereferencing m_spDBSync while it is reset here, crashing the agent (issue #38203).
+    // quiesce() always runs first and stops the sync protocols, so anything holding the
+    // shared lock is already unwinding before this exclusive lock is requested.
+    std::unique_lock<std::shared_mutex> resourcesLock(m_resourcesMutex);
+
     // Explicitly release all resources to ensure clean state between tests
     // and prevent use-after-free when Syscollector singleton destructs
     // after static dependencies have already been destroyed
@@ -2271,6 +2280,10 @@ void Syscollector::initSyncProtocol(const std::string& moduleName, const std::st
                                     unsigned int retries,
                                     size_t maxEps, uint32_t integrityInterval)
 {
+    // Publish the protocols under the exclusive lock so a concurrent reader taking the
+    // shared lock sees either a fully constructed protocol or none, never a half-assigned one.
+    std::unique_lock<std::shared_mutex> resourcesLock(m_resourcesMutex);
+
     m_dataCleanRetries = retries;  // Same as sync retries for data clean notifications
     m_integrityIntervalValue = integrityInterval;
 
@@ -2462,6 +2475,11 @@ SyncModuleResult Syscollector::syncModule(Mode mode)
 
 void Syscollector::persistDifference(const std::string& id, Operation operation, const std::string& index, const std::string& data, uint64_t version, bool isDataContext)
 {
+    // Serialize against releaseResources(): this entry point is driven by threads that
+    // outlive the module teardown (wcom's detached dispatcher and agent-info's in-process
+    // coordination polls), so the members it reaches must not be reset underneath it.
+    std::shared_lock<std::shared_mutex> resourcesLock(m_resourcesMutex);
+
     // VD tables: system (os), packages, hotfixes
     bool isVDTable = (index == SYSCOLLECTOR_SYNC_INDEX_SYSTEM ||
                       index == SYSCOLLECTOR_SYNC_INDEX_PACKAGES ||
@@ -2479,6 +2497,11 @@ void Syscollector::persistDifference(const std::string& id, Operation operation,
 
 bool Syscollector::parseResponseBuffer(const uint8_t* data, size_t length)
 {
+    // Serialize against releaseResources(): this entry point is driven by threads that
+    // outlive the module teardown (wcom's detached dispatcher and agent-info's in-process
+    // coordination polls), so the members it reaches must not be reset underneath it.
+    std::shared_lock<std::shared_mutex> resourcesLock(m_resourcesMutex);
+
     // Route to regular (non-VD) sync protocol only
     if (m_spSyncProtocol)
     {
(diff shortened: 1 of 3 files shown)


src/wazuh_modules/syscollector/src/syscollectorImp.cpp
lchico, line 3007: This PR locks query() itself, but the async flush worker it can spawn isn't covered. query()'s "flush" command calls flush() → m_asyncFlushController->startFlush(), which runs executeFlushSync() on its own thread and dereferences m_spSyncProtocol/m_spSyncProtocolVD with no lock at all. If releaseResources() resets those pointers while a flush is still in flight (reachable any time after quiesce()'s one-shot waitForFlushToFinish() has already returned and a new "flush" command arrives via wcom's detached dispatcher), we hit the exact same check-then-use race this PR is fixing for query() — just through this path instead. SCA's df06f36bec handled this by joining the async flush controller before resetting m_spSyncProtocol; could we mirror that here, or take m_resourcesMutex inside executeFlushSync()?

src/wazuh_modules/syscollector/src/syscollectorImp.cpp
```diff
@@ -4320,6 +4368,11 @@ bool Syscollector::notifyDisableCollectorsDataClean()
 
 void Syscollector::deleteDisableCollectorsData()
```
lchico, line 4392: This function holds m_resourcesMutex shared for its whole body and then calls deleteDatabase() (line ~2904), which independently takes a second shared_lock on the same mutex on the same thread. Most std::shared_mutex implementations block new shared-lock acquisitions once a writer is waiting (to avoid writer starvation). So if releaseResources() or initSyncProtocol() is blocked waiting for the exclusive lock at the moment this nested  lock_shared() runs, we get a self-deadlock: the nested lock waits behind the pending writer, and the writer waits on this thread's still-held outer lock. Worth either not re-locking in deleteDatabase() when called from here, or factoring the shared body into a private non-locking helper both callers use.

lchico (changes requested):

anromerom:
# Conclusion

- **Reproduced the reported failure locally:** The flaky test was reproduced on Windows at the same rate and duration seen in CI, roughly one failure in six to ten runs, always with the identical signature.
- **Identified the root cause of the test failure:** The test was not flaky. The agent process was crashing during reconnection, so it never completed its retries and never re-enrolled, and the test simply waited out its deadline. It occurs when the inventory module is asked for information while its database handle is unavailable, either because its startup failed or because it is being released while the agent shuts down.
- **Fixed the crash in the agent:** The module now refuses that request safely instead of crashing, and its shutdown waits for any work still in progress before releasing what that work depends on.
- **Resolved two defects raised in code review:** An asynchronous data flush could still outlive the shutdown, and one internal path could block itself. Both are closed.
- **Repaired a regression introduced by the first attempt at the fix:** An early version disabled the flush capability for the remaining life of the process. It now waits for pending work instead of discarding it, which also removed a one-hour stall in the automated checks.
- **Validated the fix against the original failure:** One hundred consecutive runs of the failing test produced no crash, against five in thirty before the change. The chance of that result if the defect remained is negligible.
- **Validated the agent on a real deployment:** A package built from the final change was installed on Windows against a live manager. Inventory, policy assessment and file integrity monitoring all worked, with no regression.

cborla (approved):
LGTM
- [x] Code changes reviewed

A related issue has been raised, but it is not a critical issue:
https://github.com/wazuh/wazuh/issues/38352

lchico (approved):
LGTM!

PR #38338, opened 2026-08-12
Agent and manager default connection timing parameters

### LINKED ISSUE #38293: Review remoted and modules configuration
## Description

Review all manager configuration options related to remoted and modules.

## Tasks

* Remove obsolete options.
* Add missing options.
* Rename confusing options.
* Update documentation.

## Testing

* Configuration validation.
* Startup tests.
* Documentation review.

## Definition of Done

* Configuration cleaned.
* Documentation updated.

jepalfer:
## Description

The agent and the manager each defined their own connection timing values, and a large part of them
were compile-time constants with no configuration surface at all. On the agent side the per-request
timeouts, the retry budgets, the backoff parameters and the producer-pause threshold were `constexpr`
values the C bridge never populated, so the agreed defaults could only be changed by rebuilding — and
that blocked the agreement itself: the two teams cannot converge on a value that only exists as a
constant in a header.

This pull request turns the values that genuinely belong to the agent/manager timing contract into
internal options: **18 new options**, 15 on the agent (`agent.*`, read from `internal_options.conf` /
`local_internal_options.conf`) and 3 on the manager (`remoted.*` and
`wazuh_modules.inventory_sync_server_*`, overridable from `wazuh-manager-internal-options.conf`).

Related to #38284.

### Scope: what is deliberately NOT configurable

An earlier revision of this PR exposed 50 options. That was too many. The bar applied since:

> An option earns its existence only if the correct value depends on the **deployment** — fleet size,
> endpoint network, hardware, indexer sizing — and support would plausibly tell an administrator to
> change it during a real incident.

Internal mechanics fail that bar and stay compiled in, because exposing them turns implementation
details into a compatibility surface without giving anyone a decision they can actually make:
control-plane cadences, module heartbeats, the inventory-sync shutdown ladder, internal retry counts,
worker wake periods, the log-throttle window, accounting constants, in-memory queue depths and the
per-endpoint body caps.

Concretely, this removed 26 manager options and 6 agent options versus the earlier revision, along
with the two mechanisms that existed only to serve them (see below). Everything removed reverted to
the compile-time constant it had before, so no default moved as a result.

## Proposed Changes

Two mechanisms, chosen by what each value could actually reach:

1. **Config read** where a configuration object was already in scope — the agent's `hc_config_t` →
   `ModuleConfig` chain, and the session query batch size through `SyncPipelineConfig` →
   `SessionProcessor`.
2. **Constructor argument** where the class already carried configuration — the VD scan read timeout
   through `ScanVdHandlerImpl` → `Impl`.

Every option follows the existing `<=0 -> built-in default` convention, so agentd, remoted and
modulesd pass their values through unconditionally.

A third mechanism present in the earlier revision — a settable process-wide default, used for
`LogThrottle`'s window and the six per-endpoint body caps — is **gone**, together with the options it
served. `LogThrottle` is back to a `constexpr` window and the new `protocolLimits.hpp` header no
longer exists.

### One default did change

`agent.https_stateful_timeout` ships as **90000 ms**, while the constant it replaced was
**120000 ms**. This covers `/stateful` and both `POST /download` kinds (shared configuration and
WPK). Every other value in this PR resolves to exactly what it resolved to before.

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer

### Artifacts Affected

- `wazuh-agentd` and the `https_client` module (Linux, Windows, macOS).
- `wazuh-remoted` and the `remoted_module` library.
- `wazuh-modulesd` and the `inventory_sync_server` module.
- Default configuration file `etc/internal_options.conf`: 15 new `agent.*` entries, growing it from
  6919 to 9744 bytes. The file is read by the agent only — the manager leaves `WAZUH_DEFINES`
  undefined and takes overrides from `etc/wazuh-manager-internal-options.conf`, shipped empty and
  unchanged here.
- Public C ABI: new fields on `hc_config_t`, `remoted_module_config_t` and
  `inventory_sync_server_config_t`. All three are zero-initialised by every caller and have no
  positional aggregate-initialisation sites, so the additions are source-compatible.
- Agent check files: the four Linux agent manifests that pin `internal_options.conf`'s size.

### Configuration Changes

Every value below was a compile-time constant before this PR. Ranges are enforced by
`getDefine_Int_default()`, which rejects an out-of-range or non-numeric value by refusing to start
the daemon rather than clamping.

#### Agent — `agent.*` (`etc/internal_options.conf`)

An attempt count is the **total** number of tries, not retries after the first: `1` means "send once,
never retry". Only retryable failures and back-pressure (`503`) consume an attempt; authentication
failures, permanent errors and version rejections stop immediately. A step's worst case is therefore
about `attempts × timeout` plus the jittered backoff between tries — check that figure against
`<global><agents_disconnection_time>` before raising either.

| Was (constant) | Now (option) | Default | Range |
|---|---|---|---|
| `request_timeout_ms` (never populated) | `agent.https_request_timeout` | `10000` ms | 1000–600000 |
| `stateful_timeout_ms` (never populated) | `agent.https_stateful_timeout` | **`90000` ms** (was 120000) | 1000–3600000 |
| `backoff_base_ms` (never populated) | `agent.https_backoff_base` | `1000` ms | 100–60000 |
| `backoff_cap_ms` (never populated) | `agent.https_backoff_cap` | `60000` ms | 1000–3600000 |
| `rejected_retry_interval_s` (never populated) | `agent.https_rejected_retry_interval` | `60` s | 1–86400 |
| `wpk_max_download_bytes` (never populated) | `agent.https_wpk_max_download_bytes` | `209715200` (200 MiB) | 1048576–2147483647 |
| `CONTROL_MAX_ATTEMPTS` | `agent.https_control_attempts` | `4` | 1–64 |
| `STATELESS_MAX_ATTEMPTS` | `agent.https_stateless_attempts` | `5` | 1–64 |
| `STATEFUL_MAX_ATTEMPTS` | `agent.https_stateful_attempts` | `5` | 1–64 |
| `DOWNLOAD_MAX_ATTEMPTS` + `WPK_DOWNLOAD_MAX_ATTEMPTS` | `agent.https_download_attempts` | `2` | 1–64 |
| `CONTROL_UNDELIVERABLE_THRESHOLD` | `agent.https_producer_pause_threshold` | `2` | 1–1000 |
| `REPORTER_MAX_ATTEMPTS` | `agent.https_reporter_attempts` | `1` | 1–64 |
| `FAST_FOLLOWUP_INTERVAL` | `agent.https_fast_followup_interval` | `1` s | 1–3600 |
| `ENROLLMENT_RETRY_TIME_DELTA` + `BRIDGE_REENROLL_RETRY_DELTA_S` | `agent.enrollment_retry_delta` | `5` s | 1–3600 |
| `ENROLLMENT_RETRY_TIME_MAX` + `BRIDGE_REENROLL_RETRY_MAX_S` | `agent.enrollment_retry_max` | `60` s | 1–86400 |

The two enrollment values were duplicated file-local constants in `start_agent.c` and
`https_client_bridge.c`; both loops now read one shared pair and can no longer drift apart. The two
download-attempt constants were likewise duplicated across the config and WPK fetchers.

`ModuleConfig::validateTiming()` rejects an inverted `backoff_base` / `backoff_cap` pair at startup.
`getDefine_Int_default()` range-checks each option on its own but cannot relate two of them, so an
inverted pair is reachable from two individually valid values.

#### Manager — `remoted.*` and `wazuh_modules.*` (`etc/wazuh-manager-internal-options.conf`)

| Was (constant) | Now (option) | Default | Range |
|---|---|---|---|
| `kKeepaliveThrottleSec` | `remoted.control_keepalive_throttle` | `60` s | 1–3600 |
| `VD_SCAN_READ_TIMEOUT_SECONDS` | `remoted.scanvd_read_timeout` | `60` s | **31**–3600 |
| `BATCH_SIZE` | `wazuh_modules.inventory_sync_server_session_query_batch_size` | `1000` | 0–100000 |

Why each of these three survived the bar above:

- **`control_keepalive_throttle`** bounds how often a `notify` reaches wazuh-db, so the write load it
  produces scales with the fleet. It is the manager half of the timing contract and has to be
  settable before any notify cadence is agreed with the agent. It must stay at or above the agent's
  notify cadence, and well below `<global><agents_disconnection_time>`.
- **`scanvd_read_timeout`** is a deadline against another daemon whose scan duration grows with the
  agent's inventory. Its floor of **31** is the only range here derived from a real constraint rather
  than convention: modulesd's scan handler can legitimately block ~30 s in its pause/quiesce drain
  before scanning starts, so a lower timeout misreads an in-progress scan as a network failure and
  retries it while the first attempt is still running.
- **`session_query_batch_size`** is the indexer search page size while draining a session. Larger
  pages mean fewer round trips at a proportionally larger response held in memory, so the right value
  follows the indexer's sizing rather than this module's. It keeps the module's `0 -> built-in
  default` convention, so the shipped `0` means `1000`.

### Tests Introduced

No new test files; the two existing configuration-resolution suites were extended.

- `src/client-agent/https_client/tests/unit/moduleConfig_test.cpp`: the new fields added to the
  zeroed-config defaults case and to an explicit-override case, plus two cases for
  `validateTiming()` — inverted backoff bounds rejected, and equal bounds accepted (the boundary,
  since the check is `>` and not `>=`).
- `src/remoted/remoted_module/test/unit/controlConfig_test.cpp`: `keepaliveThrottleSec` added to the
  defaults, positive-override and negative-guard cases. The negative case matters because
  `buildControlConfig()` casts to `uint32_t`, so a negative must fall back rather than wrap.

**Coverage gap worth stating plainly:** `remoted.scanvd_read_timeout` and
`inventory_sync_server_session_query_batch_size` have no test asserting the value reaches its
consumer. That is the failure mode to watch for here — during development one option was read,
range-checked, stored in the struct and then silently ignored because the middle of the chain was
missing, and no existing test would have caught it.

### Validation performed

Run against a clean Ubuntu 24.04 VM using the same tools CI uses:

- **astyle 3.1** (the exact version CI installs) over `client-agent/https_client`: **PASS**.
- **clang-format 22.1.1** (the Wazuh binary from `packages.wazuh.com`) over all 83
  `inventory_sync_server` files: **PASS**, 0 warnings.
- **cppcheck** via `build.py --readytoreview`: **PASS**.
- Unit test suites: **not yet run** — `cmocka` was missing in the validation environment, which
  blocked the `https_client` build behind it. `remoted_module_utest` and the InventorySyncServer
  suites have not been run either.

> These runs were performed on the earlier, larger revision of this branch and have **not** been
> re-run since the option set was reduced. Most of the reduction is a revert to previously-compiling
> code, but the formatting and static-analysis passes should be repeated before merge.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/.github/actions/check_files/deb_linux_agent_amd64.csv
+++ b/.github/actions/check_files/deb_linux_agent_amd64.csv
@@ -23,7 +23,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libagent_info.so,root,wazuh,750,file,-rwxr-x---,389488,0.1
 /var/ossec/lib/libagent_metadata.so,root,wazuh,750,file,-rwxr-x---,14448,0.1
 /var/ossec/lib/libgcc_s.so.1,root,wazuh,750,file,-rwxr-x---,183248,0.1
-/var/ossec/etc/internal_options.conf,root,wazuh,640,file,-rw-r-----,6576,0.1
+/var/ossec/etc/internal_options.conf,root,wazuh,640,file,-rw-r-----,9604,0.1
 /var/ossec/etc/wpk_root.pem,root,wazuh,640,file,-rw-r-----,1367,0.1
 /var/ossec/etc/localtime,root,wazuh,640,file,-rw-r-----,118,0.1
 /var/ossec/etc/client.keys,root,wazuh,640,file,-rw-r-----,0,0.1
--- a/.github/actions/check_files/deb_linux_agent_i386.csv
+++ b/.github/actions/check_files/deb_linux_agent_i386.csv
@@ -23,7 +23,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libagent_info.so,root,wazuh,750,file,-rwxr-x---,402688,0.1
 /var/ossec/lib/libagent_metadata.so,root,wazuh,750,file,-rwxr-x---,8816,0.1
 /var/ossec/lib/libgcc_s.so.1,root,wazuh,750,file,-rwxr-x---,118632,0.1
-/var/ossec/etc/internal_options.conf,root,wazuh,640,file,-rw-r-----,6576,0.1
+/var/ossec/etc/internal_options.conf,root,wazuh,640,file,-rw-r-----,9604,0.1
 /var/ossec/etc/wpk_root.pem,root,wazuh,640,file,-rw-r-----,1367,0.1
 /var/ossec/etc/localtime,root,wazuh,640,file,-rw-r-----,118,0.1
 /var/ossec/etc/client.keys,root,wazuh,640,file,-rw-r-----,0,0.1
--- a/.github/actions/check_files/rpm_linux_agent_amd64.csv
+++ b/.github/actions/check_files/rpm_linux_agent_amd64.csv
@@ -87,7 +87,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libagent_info.so,root,wazuh,750,file,-rwxr-x---,418344,0.1
 /var/ossec/lib/libagent_metadata.so,root,wazuh,750,file,-rwxr-x---,14472,0.1
 /var/ossec/lib/libgcc_s.so.1,root,wazuh,750,file,-rwxr-x---,187064,0.1
-/var/ossec/etc/internal_options.conf,root,wazuh,640,file,-rw-r-----,6576,0.1
+/var/ossec/etc/internal_options.conf,root,wazuh,640,file,-rw-r-----,9604,0.1
 /var/ossec/etc/wpk_root.pem,root,wazuh,640,file,-rw-r-----,1367,0.1
 /var/ossec/etc/localtime,root,wazuh,640,file,-rw-r-----,118,0.1
 /var/ossec/etc/client.keys,root,wazuh,640,file,-rw-r-----,0,0.1
--- a/.github/actions/check_files/rpm_linux_agent_i386.csv
+++ b/.github/actions/check_files/rpm_linux_agent_i386.csv
@@ -87,7 +87,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libagent_info.so,root,wazuh,750,file,-rwxr-x---,416056,0.1
 /var/ossec/lib/libagent_metadata.so,root,wazuh,750,file,-rwxr-x---,8768,0.1
 /var/ossec/lib/libgcc_s.so.1,root,wazuh,750,file,-rwxr-x---,111408,0.1
-/var/ossec/etc/internal_options.conf,root,wazuh,640,file,-rw-r-----,6576,0.1
+/var/ossec/etc/internal_options.conf,root,wazuh,640,file,-rw-r-----,9604,0.1
 /var/ossec/etc/wpk_root.pem,root,wazuh,640,file,-rw-r-----,1367,0.1
 /var/ossec/etc/localtime,root,wazuh,640,file,-rw-r-----,118,0.1
 /var/ossec/etc/client.keys,root,wazuh,640,file,-rw-r-----,0,0.1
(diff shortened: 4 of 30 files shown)


jotacarma90 (approved):
LGTM

PR #38340, opened 2026-08-12
syscollector normalizer: stop erasing while iterating in removeExcluded() (4.14.9)

### LINKED ISSUE #38170: syscollector normalizer: removeExcluded() erases while iterating the JSON array and recompiles the exclusion regex on every call
## Description

`SysNormalizer::removeExcluded()` (`src/wazuh_modules/syscollector/src/syscollectorNormalizer.cpp:24-48`) erases elements from a `nlohmann::json` array while iterating it with a plain `++item` loop. This is undefined behavior, and in practice it silently skips elements, so exclusions configured in the normalization file are not always applied. The same function also recompiles the exclusion `std::regex` on every call.

With the default configuration this only affects **macOS** agents; see "Impact and platform scope" below.

## Code

```cpp
// src/wazuh_modules/syscollector/src/syscollectorNormalizer.cpp:31-48
for (const auto& exclusionItem : exclusionsIt->second)
{
    try
    {
        std::regex pattern{exclusionItem["pattern"].get_ref<const std::string&>()};
        const auto& fieldName{exclusionItem["field_name"].get_ref<const std::string&>()};

        if (data.is_array())
        {
            for (auto item{data.begin()}; item != data.end(); ++item)
            {
                const auto fieldIt{item->find(fieldName)};

                if (fieldIt != item->end() && std::regex_match(fieldIt->get_ref<const std::string&>(), pattern))
                {
                    data.erase(item);
                }
            }
        }
```

## Root cause

`nlohmann::basic_json::erase(iterator)` on an array invalidates the passed iterator (it is implemented over `std::vector::erase`, so every iterator at or after the erased position is invalidated, and a reallocation-free shift still leaves `item` pointing at the *next* element). The loop then executes `++item` on that invalidated iterator:

1. Formally undefined behavior.
2. Observable effect: the element that shifted into the erased position is never tested, so two consecutive matching items only get one of them removed.
3. The loop condition `item != data.end()` is re-evaluated against an iterator that may already be past the shrunken end when the erased element was the last one.

Secondary issue in the same block: `std::regex pattern{...}` is constructed inside the loop over `exclusionsIt->second`, so every exclusion pattern is recompiled on every `removeExcluded()` call. `m_typeExclusions` is built once in the constructor (`syscollectorNormalizer.cpp:19-21`, from the normalization config file) and never mutated, so the compiled patterns could be cached. Regex compilation is significantly more expensive than the matching itself, and this runs for every inventory scan.

## Impact and platform scope

Reachable only when the running platform's normalizer target has exclusion entries. `SYSCOLLECTOR_NORM_TYPE` (`src/shared/include/defs.h:187-193`) is `macos`, `windows` or `linux`, and all 17 entries of the shipped `src/wazuh_modules/syscollector/norm_config.json` are `"target": "macos"`, so:

- **macOS**: the bug is live. Two exclusion patterns ship (`(Siri)`, `(iCloud)`) and are applied to the `packages` array (`syscollectorImp.cpp:1574`). Whenever a matching package is followed in the array by another package matching the same pattern, the second one is never tested and stays in the inventory. Result: packages the user configured as excluded are reported anyway, non-deterministically depending on collection order.
- **Linux / Windows**: `m_typeExclusions` ends up empty, `removeExcluded()` returns at the `find(type)` miss, and neither the bug nor the regex cost is reachable with the default configuration. A user shipping their own `norm_config.json` with `linux`/`windows` targets would hit it.

Performance side: O(n) shifting per removal plus one regex compilation per exclusion per call, same platform scope.

## Proposed fix

1. Stop erasing through an invalidated iterator. Options, in order of preference:
   - Single reverse pass by index over the underlying array, swapping the matched element with the last one and popping it (O(1) per removal, no iterator invalidation):
     ```cpp
     auto& arr{data.get_ref<nlohmann::json::array_t&>()};
     for (size_t i{arr.size()}; i-- > 0;)
     {
         const auto fieldIt{arr[i].find(fieldName)};
         if (fieldIt != arr[i].end() && std::regex_match(fieldIt->get_ref<const std::string&>(), pattern))
         {
             std::swap(arr[i], arr.back());
             arr.pop_back();
         }
     }
     ```
     This changes the relative order of the surviving elements. The array is a batch of collected inventory items consumed by dbsync, which keys rows by primary key and not by position, so order is not expected to be load-bearing, but this should be confirmed with the module owners before merging.
   - Order-preserving alternative if order must be kept: `item = data.erase(item);` with no increment on the erase branch, or a single `std::remove_if` pass over the underlying array. Correct, but still O(n) shifting per removal.

2. Cache the compiled exclusion patterns. Tracked together with the equivalent problem in `normalizeItem()` in a separate enhancement issue, since it touches the constructor and the member layout.

## Ask

Confirm whether the relative order of items inside the normalized array is significant for any consumer. If it is not, go with the swap-and-pop variant; if it is, go with the order-preserving `erase` return value.

Miguevrgo:
## Description

Backport of #38192 to `4.14.9`. Closes #38170.

`SysNormalizer::removeExcluded()` (`src/wazuh_modules/syscollector/src/syscollectorNormalizer.cpp:23-48` on this branch) erased elements from a `nlohmann::json` array while iterating it with a plain `++item` loop:

```cpp
for (auto item{data.begin()}; item != data.end(); ++item)
{
    const auto fieldIt{item->find(fieldName)};

    if (fieldIt != item->end() && std::regex_match(fieldIt->get_ref<const std::string&>(), pattern))
    {
        data.erase(item);
    }
}
```

`nlohmann::basic_json`'s array is backed by a `std::vector`, so `erase(item)` shifts every element after the removed position down by one and returns an iterator to the element that took the erased slot. The loop discards that return value and does `++item` on the now-invalidated `item`, which is undefined behavior and, in the concrete `std::vector` implementation used here, lands one position past the element that just shifted into the erased slot. That element is never tested against the pattern. Two consecutive entries matching the same exclusion pattern therefore only get the first one removed; the second is silently kept.

This has been in the code since the class was introduced (`6942e9d498`, 2021-01-14), unchanged through every release including this one. With the shipped `norm_config.json`, the only exclusion entries are `target: "macos"`, `field_name: "name"`, patterns `(Siri)` and `(iCloud)`, applied to the `packages` array, so on macOS two adjacent packages matching the same exclusion pattern only have the first excluded, non-deterministically depending on collection order.

## Proposed Changes

`removeExcluded()`'s array branch now does a reverse index pass over the underlying `nlohmann::json::array_t`, swapping each matched element with the array's last element and popping it:

```cpp
auto& arr{data.get_ref<nlohmann::json::array_t&>()};

for (size_t i{arr.size()}; i-- > 0;)
{
    const auto fieldIt{arr[i].find(fieldName)};

    if (fieldIt != arr[i].end() && std::regex_match(fieldIt->get_ref<const std::string&>(), pattern))
    {
        std::swap(arr[i], arr.back());
        arr.pop_back();
    }
}
```

This removes each match in O(1) instead of the O(n) element shift a `std::vector::erase` does, and there is no iterator to invalidate. Every index from `size()-1` down to `0` is visited exactly once regardless of how many earlier removals happened, so no element is ever skipped.

This changes the relative order of surviving elements. The only caller (`Syscollector::scanPackages()`) feeds `removeExcluded` one package object at a time and wraps it in a single-element array before handing it to `txn.syncTxnRow()`, so the array branch isn't even reached by the current production call path on this branch either. `dbsync_packages` is keyed by primary key (`name, version_, architecture, type, path`) and upserted through `syncTxnRow`/`syncRow`, not consumed by position, so order was never load-bearing for the one consumer that exists.

## Results and Evidence

Built `sys_normalizer_unit_test` standalone against this branch's `wazuh_modules/syscollector` module (`cmake -DUNIT_TEST=ON` from within the module directory, since on `4.14.9` this module isn't wired into the top-level `src/CMakeLists.txt` unit-test tree) and ran the existing suite (unchanged, no new test added — `excludeSiriAndiTunes` already exercises the array branch with two exclusion patterns over a 50-package array):

```
[==========] Running 16 tests from 1 test suite.
...
[ RUN      ] SysNormalizerTest.excludeSiriAndiTunes
[       OK ] SysNormalizerTest.excludeSiriAndiTunes (0 ms)
...
[----------] 16 tests from SysNormalizerTest (6 ms total)
[==========] 16 tests from 1 test suite ran. (6 ms total)
[  PASSED  ] 16 tests.
```

## Manual tests with their corresponding evidence

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X
- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer
  - [ ] Test run in parallel
- Wazuh server API/Framework

### Artifacts Affected

N/A

### Configuration Changes

N/A

### Tests Introduced

N/A

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...


Darioortegaleyva (changes requested):
Fix logic checks out, tested it against a two-adjacent-matches case (base code only removes one, this branch removes both).

A few things:

- Changelog field on the issue is `pending`, add a line to `CHANGELOG.md`.
- Please fill in the release field on the issue.
- Tested the current code under ASan with the match on the last array element: heap-buffer-overflow, not just a skipped element. `data.erase(item)` returns `end()` there, and the loop does `++item` on it. Reproduces every time with a plain 3-element array, no SysNormalizer/regex involved. If you can, please run it under ASan on your end too and add it as evidence, or at least take a look. Worth a regression test for this case too, since neither this one nor the adjacent-match case is covered by the existing `excludeSiriAndiTunes` test.

Darioortegaleyva (dismissed):
LGTM

MarcelKemp (approved):

PR #38344, opened 2026-08-12
Gate 4.x-specific remoted subsystems behind the legacy configuration block

### LINKED ISSUE #38279: Gate 4.x-specific remoted subsystems behind the legacy configuration block
## Description

`remoted` currently starts every subsystem unconditionally, whether or not any 4.x agent will ever connect. `HandleSecure()` spawns the legacy listener, the sender pool, `update_shared_files`, `AR_Forward` and the legacy `remote_upgrade` task-delivery poller on every start, and `Read_Remote()` falls back to port 1514 even when the configuration carries no `<remote><legacy>` block, so the legacy listener always binds.

On a deployment with only 5.x agents this wastes threads and sockets, keeps dead code paths hot, and makes the module harder to reason about.

The agreed support scope for 4.x agents under a 5.x manager is: they can **send stateless events** and **receive WPK upgrades**, nothing else. Active response, `restart`/`force reconnect`/`config reload` orders and `getconfig`/`getstats` toward 4.x agents are all dropped (active response was already removed in #34606). That decision splits this work into three buckets.

## Remove outright (dead code, not gated)

- `AR_Forward` and the `queue/sockets/ar` socket (`ARQUEUE`). Nothing writes to that queue anymore: the producers were removed with #34606, and the thread today consumes a queue nobody feeds. Active response for 5.x agents has its own complete path and never touches remoted-legacy: the AR document is indexed into `wazuh-active-responses*`, the framework's `active_response_processing()` polls it and creates one `active_response` task per target agent in the Task Manager, and the agent receives it in its next `/control` notify response.
- Audit for other remains of the dropped 4.x features (`restart`/`force reconnect`/`config reload` dispatch toward 4.x agents) and remove them under the same criterion.
- The `sent_breakdown.ar` counter in remoted's per-agent statistics becomes orphaned by this removal — coordinate with #38280 (statistics set reduction).

## Gate behind the `<legacy>` block (alive, 4.x-only)

- The 1514 listener, the `wait_for_msgs` sender pool, netbuffer/netcounter, and the legacy control-message/keepalive path.
- The legacy `remote_upgrade` task-delivery poller — the one manager-to-4.x feature that must keep working, since it is the migration path to 5.x.
- Add an explicit enable/disable variable for the block: when legacy is disabled, nothing 4.x-related starts (no thread, no socket, no port bind).
- When legacy delivery is disabled, creating a `remote_upgrade` task for a < 5.0 agent must be rejected at creation time instead of producing a task that can never be delivered.

## Always on (needed with or without legacy)

- The HTTPS server, the remcom socket, state, and the agent metadata handling.
- Shared configuration generation (`merged.mg`, groups and multigroups): `/download` serves it to 5.x agents, so it must keep working with legacy disabled. Evaluate moving group/multigroup generation out of remoted into another daemon so remoted-legacy can be turned off as a whole; there may be other subsystems in the same situation.
- Reorder the startup path so the remove/gate/always-on split is explicit in the code.

## Definition of Done

- [ ] `AR_Forward` and the `ar` queue removed, along with any other remains of the dropped 4.x features.
- [ ] With no `<legacy>` block (or the block disabled), remoted starts no 4.x-related thread or socket.
- [ ] Shared configuration generation keeps working for 5.x agents in that mode.
- [ ] Decision recorded on moving group/multigroup handling to another daemon.
- [ ] `remote_upgrade` tasks for pre-5.0 agents are rejected when legacy delivery is off.
- [ ] Unit tests covering both startup modes.

Related to #37702 and #34606. Interacts with the `<remote>` installation variables work in #38157 and the statistics set reduction in #38280.

Antoniogm03:
## Description

Closes #38279

`remoted` started every 4.x-only subsystem unconditionally. `Read_Remote()` defaulted `port`/`protocol`/`local_ip` to the classic 1514 listener even when `<remote><legacy>` was entirely absent from the configuration, so `HandleRemote()` always bound the socket and `HandleSecure()` always spawned every legacy thread, regardless of whether any 4.x agent would ever connect to that manager.

Per management direction (https://github.com/wazuh/wazuh/issues/38279#issuecomment-5240391131), this also removes `AR_Forward` outright rather than gating it, since Active Response for 4.x agents is being retired.

## Proposed Changes

- Add an explicit `legacy_enabled` flag to remoted's config. Presence of `<remote><legacy>` defaults it to enabled; an explicit `<enabled>no</enabled>` inside it disables it; absence of the block entirely now disables it too (previously it silently defaulted to port 1514).
- Gate every subsystem that only serves the classic listener behind that flag: the TCP/UDP bind itself, the legacy `remote_upgrade` task-delivery poller, the `merged.mg` push (sender pool), the control/event dispatch threads, the per-agent metadata cache cleanup thread, the message-handler worker pool, and the fd closer thread.
- Keep `merged.mg`/group generation always on regardless of the flag: the HTTPS `/download` endpoint reads those files from disk for 5.x agents too.
- Remove `AR_Forward` and `src/remoted/src/ar-forward.c` entirely (not gated). Active Response currently has no delivery path for any agent version — no HTTPS-based AR producer exists yet — so this is an accepted, deliberate regression until a follow-up issue wires that path.
- Drop the now-orphaned `ARQUEUE`/`AR_SOCKET` symbols (zero readers or writers left anywhere in the tree) and correct `docs/ref/architecture.md`'s Active Response section and IPC socket table, which still described the removed `AR_Forward` thread and a socket dispatch path that isn't how AR is actually delivered (Task Manager, not `queue/sockets/ar`).
- Reject `remote_upgrade` task creation for agents below v5.0.0 when legacy delivery is disabled, and for any agent when the target is v5.0.0+ while remoted's HTTPS `verification_mode` isn't `none`, via a single `wm_agent_upgrade_validate_remoted_delivery()` in the `agent_upgrade` module (an independent `ReadConfig(CREMOTE, ...)` re-read off disk — `wazuh-modulesd` and `wazuh-remoted` are separate processes with no shared live config). Both gates share one read instead of one each, since a batch upgrade of N legacy agents was otherwise parsing the same `wazuh-manager.conf` up to 2N times.
- Reset `port`/`proto`/`local_ip` to their disabled state (0/NULL) in `Read_Remote()` when `legacy_enabled` is false. `w_remoted_parse_legacy()` sets those fields from the raw XML tags unconditionally, so a `<legacy>` block with `<enabled>no</enabled>` next to an explicit `<port>`/`<protocol>` (the packaged default template, just flipped) left a stale nonzero `proto` behind. `secure.c`'s protocol-gated `wnotify_add()` then tried to watch a socket `HandleRemote()` never bound (fd 0), crashing `remoted` at startup with `wnotify_add(0): Operation not permitted`. Found during manual testing, not by the unit suite — see Results and Evidence.
- Gate the legacy AES keystore itself behind `legacy_enabled`: `OS_ReadKeys()`, `OS_StartCounter()` (opens/creates a rids file per agent) and the periodic key-reloader thread ran always-on, justified by a comment claiming the keystore was shared with the HTTPS CMAC auth middleware. It isn't — that middleware loads `client.keys` itself, independently (`remoted_module/src/auth/keystore.cpp`, its own inotify watch). With legacy disabled this was pure overhead that grows with fleet size; it now lives inside `start_legacy_subsystems()` alongside everything else gated by the flag.
- Reject non-JSON requests on `remcom`'s local socket (the pre-5.0 `req_sender`/`req_dispatch` path, historically used for classic getconfig/getstats-to-agent delivery) when legacy is disabled, since it's the only other path that would still reach the legacy keystore.
- Drop the now-redundant `legacy_enabled` check around `HandleRemote()`'s TCP/UDP bind: `Read_Remote()` already zeroes `proto` when legacy is disabled, the same invariant `secure.c`'s socket-registration code already relies on without a separate check.

### Results and Evidence

Built and tested inside a Vagrant VM (Ubuntu 24.04, gcc 13.3.0), per project convention:

- `make deps -C src TARGET=manager` and `make -C src TARGET=manager TEST=1` both complete with no errors or warnings in any touched file.
- Full manager unit test suite:

```
100% tests passed, 0 tests failed out of 105
```

- The four suites this change touches directly, isolated:

```
1/4 Test #1: test_manager .....................   Passed
2/4 Test #2: test_secure ......................   Passed
3/4 Test #7: test_remote-config ...............   Passed
4/4 Test #49: test_wm_agent_upgrade_commands ...   Passed
```

- The real (non-mocked) `Read_Remote()`/`w_remoted_parse_legacy()` parsing functions are exercised directly by the new/updated `test_remote-config.c` cases, covering: `<legacy>` absent (disabled, no port/proto/lip defaulting), present without `<enabled>` (enabled, defaults applied as before), present with `<enabled>no</enabled>`/`<enabled>yes</enabled>` explicitly, and the crash scenario below (port/proto pre-set to nonzero, `legacy_enabled=false`, asserting `Read_Remote()` clears them).

**Real environment testing** (manager + agents, not just unit tests), done after the unit-test pass above turned out not to be enough:

- Installed a real manager and a real 5.x agent from this branch 
- With the fix, and `<legacy><enabled>yes</enabled>` (default):

```
$ ss -tlnp | grep -E "151[4-7]"
LISTEN 0      128          0.0.0.0:1515       0.0.0.0:*    users:(("wazuh-manager-a",pid=72048,fd=6))
LISTEN 0      4096         0.0.0.0:1517       0.0.0.0:*    users:(("wazuh-manager-r",pid=72271,fd=44))
LISTEN 0      100          0.0.0.0:1516       0.0.0.0:*    users:(("python3",pid=72501,fd=23))
LISTEN 0      128        127.0.0.1:1514       0.0.0.0:*    users:(("wazuh-manager-r",pid=72271,fd=4))
```

```
2026/08/17 10:56:32 wazuh-manager-remoted: INFO: legacy_task_delivery: poller thread started (startup delay 65s, poll interval 900s)
2026/08/17 10:56:32 wazuh-manager-remoted:communication: INFO: Starting remoted module (port=1517, cluster='wazuh', workerNode=false).
2026/08/17 10:56:32 wazuh-manager-remoted:keystore: INFO: Loaded 1 agent key(s) from 'etc/client.keys'.
2026/08/17 10:56:32 wazuh-manager-remoted:http-server: INFO: HTTP server listening on https://0.0.0.0:1517/ (...)
2026/08/17 10:56:32 wazuh-manager-remoted:communication: INFO: remoted HTTP server started.
2026/08/17 10:56:32 wazuh-manager-remoted: INFO: Started (pid: 72271). Listening on port 1514/TCP (secure).
```

```
$ ps -T -p 77269 | wc -l
67
```

```
root@indexer2:/home/vagrant/git/remove_remoted/wazuh-38279# TOKEN=$(curl -s -u wazuh:wazuh -k -X POST https://localhost:55000/security/user/authenticate | python3 -c "import sys,json; print(json.load(sys.stdin)['data']['token'])") && curl -s -k -H "Authorization: Bearer $TOKEN" "https://localhost:55000/agents?select=id,name,status,lastKeepAlive,version,ip" | python3 -m json.tool
{
    "data": {
        "affected_items": [
            {
                "version": "v5.0.0",
                "ip": "192.168.57.30",
                "name": "indexer2",
                "status": "active",
                "id": "001",
                "lastKeepAlive": "2026-08-17T12:06:31+00:00"
            },
            {
                "version": "v4.14.7",
                "ip": "192.168.57.22",
                "name": "agent",
                "status": "active",
                "id": "002",
                "lastKeepAlive": "2026-08-17T12:06:30+00:00"
            }
        ],
        "total_affected_items": 2,
        "total_failed_items": 0,
        "failed_items": []
    },
    "message": "All selected agents information was returned",
    "error": 0
}
```

- With `<legacy><enabled>no</enabled></legacy>`:
  - Port 1514 disappears entirely (TCP and UDP) — `ss` shows only 1515/1516/1517. A raw `/dev/tcp` connect gets `Connection refused`.
  - `remoted`'s thread count drops from 66 to 48 — exactly the 18 threads the code gates behind `legacy_enabled` (`current_timestamp` + `legacy_upgrade_task_delivery` + `sender_pool`(8) + `save_control_thread` + `dispach_events_thread` + `agent_meta_cleanup_thread` + `worker_pool`(4) + `close_fp_main`).
  - The real 4.14.7 agent loses its connection within seconds (`Lost connection with manager`), loops forever on `Unable to connect ... Transport endpoint is not connected`, and even fails to re-enroll (`Duplicate agent name`). The Server API shows it `disconnected` with a frozen `lastKeepAlive`.
  - The real 5.x agent is unaffected throughout: still `active`, `lastKeepAlive` advancing normally, HTTPS `/control` channel working, no `Manager unreachable` warnings.
  - Re-enabling `<legacy>` reconnects the 4.x agent on its own, on the next retry, with no intervention.

**RAM usage** (`remoted` process, `VmRSS` from `/proc/<pid>/status`, devContainer E2E, 12 vCPUs):

Without agents connected:

| `<legacy><enabled>` | RAM (RSS) | Threads |
|---|---|---|
| `yes` | ~30.1 MB | 114 |
| `no` | ~26.8 MB | 95 |

With the full E2E connected (2x agent 4.x + 2x agent 5.x; the 4.x agents disconnect when legacy is off):

| `<legacy><enabled>` | Agents connected | RAM (RSS) | Threads |
|---|---|---|---|
| `yes` | 4 (2x 4.x + 2x 5.x) | ~51.2 MB | 114 |
| `no` | 2 (5.x only) | ~37.6 MB | 95 |

### Artifacts Affected

`wazuh-manager-remoted`, `wazuh-manager-modulesd` (`agent_upgrade` module), `wazuh-manager` default configuration semantics, `docs/ref/architecture.md`.

### Configuration Changes

New `<remote><legacy><enabled>yes|no</enabled></remote>` option. Absence of `<remote><legacy>` now disables the classic listener entirely, instead of silently defaulting to port 1514 — a behavior change, but the packaged default config (`etc/wazuh-manager.conf`) already ships a `<legacy>` block, so default installs are unaffected; only hand-written configs that omit the block newly get legacy fully disabled.

### Tests Introduced

- `test_remote-config.c`: `<legacy><enabled>` parsing (yes/no/invalid), `Read_Remote()` defaulting behavior across all three presence/enablement combinations, and `test_read_remote_explicit_values_cleared_when_disabled` covering the crash scenario found in manual testing.
- `test_wm_agent_upgrade_commands.c`: direct coverage of `wm_agent_upgrade_validate_legacy_delivery()` (agent ≥ v5.0.0 short-circuits, agent < v5.0.0 with legacy enabled/disabled, `ReadConfig` failure fails open), plus mock updates on every existing test whose agent version is below v5.0.0 and now also exercises this new gate.

## Review Checklist

- [ ] Code changes reviewed


jotacarma90 (approved):
LGTM

PR #38129, opened 2026-07-30
Add zstd request body support

NahuFigueroa97:
## Description

Adds optional `Content-Encoding: zstd` support to remoted's HTTPS `/stateless`
endpoint, so agents can send a compressed event batch instead of a raw one.
No other compression scheme is supported — `gzip` is explicitly rejected.

Decompression happens **after** authentication succeeds, never before: the
AES-CMAC signature always covers the exact bytes as sent on the wire (the
compressed body, if `Content-Encoding: zstd` is present), so an
unauthenticated request can never spend the manager's CPU or memory
decompressing anything. Only a request that already passed the full
protocol-version + Authorization + timestamp-window + key + CMAC check ever
reaches the decompressor.

The decompressed size is bounded by the same `remoted.auth_max_body_size` cap
(10 MiB default) that already applies to an uncompressed body, and that cap
is enforced *while* decompressing, not just at the end — so a small,
highly-compressed "decompression bomb" body is rejected the instant it would
cross the cap, before the full decompressed payload is ever held in memory.
The same defense applies to the zstd frame header itself: a zstd frame can
declare an arbitrarily large decompression window before a single byte of
output exists, so the decoder caps that window to whatever the configured
`maxOutputSize` could possibly need — never zstd's own 128 MiB library
default — so a malicious header claiming a huge window is rejected outright
instead of forcing a large up-front allocation.

## Proposed changes

- **New `zstdDecoder` module**
  (`src/remoted/remoted_module/src/common/zstdDecoder.{hpp,cpp}`): streaming
  zstd decompression via `ZSTD_decompressStream()`, checking the running
  output size after every 64 KiB chunk against `maxOutputSize`. Returns either
  the decompressed bytes or a `ZstdDecodeError` (`Malformed` — bad magic,
  truncated frame, or oversized window; `TooLarge` — would exceed the byte
  cap).

- **`AuthGateway` wired to decode after auth**
  (`src/remoted/remoted_module/src/endpoints/authGateway.{hpp,cpp}`): reads
  `Content-Encoding` once per request (case-insensitive), but only acts on it
  *after* `beginSession()`/CMAC verification succeeds. On success, replaces
  the request's `Payload` with a new one backed by the decompressed bytes,
  owned by their own keep-alive (the original compressed-body keep-alive is
  released at that point). A missing header keeps today's behavior (body
  passed through unchanged, no zstd involved).

- **Two new auth error cases**
  (`src/remoted/remoted_module/src/auth/authTypes.hpp`,
[...]

PR #38355, opened 2026-08-13
Enhancement/38308 agent https zstd compression

### LINKED ISSUE #38308: Agent HTTPS client: send zstd-compressed request bodies (all endpoints)
|---|---|---|---|---|
| 5.0.0-https | wazuh-agentd | Agent | Sources | Linux / Windows / macOS |

> Part of the agent-side HTTPS migration (#37702, agent spike #37738, agent breakdown
> #37831). Extends the foundation work (#37828) and applies to every existing
> outbound send path: `/stateless` (#37835), `/control` (#37830), `/download`
> (#37832), `/stateful` (#37836), and the periodic `/stats`+`/config` push (#37843) —
> all of them already implemented and merged, this issue adds compression on top of
> each. Consumes the manager-side compression contract added by #38109 / PR #38129
> and the endpoint-agnostic decoding wiring on `5.0.0-https` (see "Current state").
> Distinct from #38282 (open spike on compressing the manager's *download response*
> bytes for WPK/`merged.mg`) — this issue is entirely about the agent's *request*
> bodies, the opposite direction.

## Description

The manager side already accepts `Content-Encoding: zstd` request bodies. #38109
asked for `gzip`, but the actual implementation (PR #38129, merged) supports only
**zstd** — `gzip` is explicitly rejected with `415`. Decompression happens **after**
AES-CMAC authentication succeeds, so the signature always covers the exact bytes sent
on the wire (the compressed body, when `Content-Encoding: zstd` is present); the
decompressed size is bounded by `remoted.auth_max_body_size` (10 MiB default),
enforced incrementally while streaming, and the zstd frame's declared decompression
window is itself capped against a decompression-bomb header.

**This support is endpoint-agnostic, not `/stateless`-specific.**
`AuthGateway::addAuthenticatedRoute()`
(`src/remoted/remoted_module/src/endpoints/authGateway.cpp`) runs the injected
`IBodyDecoder` on every authenticated route, right after AES-CMAC succeeds and before
the handler runs — by design, so "a new endpoint cannot accidentally opt out of it"
(the class's own doc comment: "one instance serves every route and every in-flight
request"). `remotedModuleFacade.hpp:279-390` registers `/stateless`, `/download`,
`/stats`, `/config`, `/stateful` **and** `/control` through that same gateway, so all
six already accept a zstd body today.

**This issue is the full agent-side development of compression: analysis, the
compression mechanism, wiring it into every existing send path, and the build-system
work — including generating the precompiled zstd artifacts the agent build needs —
not just a single endpoint.** The agent side currently has **no** compression code at
all (`grep -rniE "zstd|gzip|content-encoding" src/client-agent` returns nothing).

## Goals

- Compress outbound request bodies with zstd and set `Content-Encoding: zstd`, across
  every send path that exists today: `/stateless` (`statelessStream.cpp`), `/control`
  (`controlStream.cpp`: startup/notify/shutdown), `/download`
  (`wpkFetcher.cpp`, `configFetcher.cpp`), and the periodic `/stats`+`/config` push
  (`reporterStream.cpp`); and design the file-backed `/stateful`
  (`statefulStream.cpp`) case so it can adopt the same mechanism without re-solving
  the negotiation/fallback problem from scratch.
- Build the compression + `Content-Encoding` + fallback-on-rejection logic **once**,
  at a single choke point, instead of duplicating it across six-plus call sites (see
  "Design considerations").
- Vendor zstd for the **agent and winagent** builds (today it is a manager-only
  dependency, and even the manager's own CMake wiring skips Windows — see "Current
  state") and **generate and publish the precompiled zstd build artifacts** the
  agent's release pipeline needs, the same way other agent dependencies are
  distributed today — this is explicitly in scope, not an assumed prerequisite done
  elsewhere.
- Preserve existing size/back-pressure semantics on `/stateless`: the adaptive payload
  budget must stay based on **uncompressed** content bytes, matching what the
  manager's decompressed-size cap actually limits.
- Never lose data because a manager rejects the encoding — every affected
  send path must fall back to uncompressed rather than treating the rejection as a
  generic permanent failure.

### In scope

- **A single, shared compression component** in `https_client`, wired at the lowest
  common point that already sees every request: `RetrySender`
  (`src/client-agent/https_client/src/retrySender.cpp`). `RetrySender::attemptOnce()`
  already branches on `attempt.bodyFilePath.empty()` to choose `sign()` vs.
  `signFile()` (`retrySender.cpp:79-82`) and is the one place through which
  `/stateless`, `/control`, `/download` and `/stateful` already all flow — adding
  compression here means every current *and future* caller gets it automatically,
  with the `Content-Encoding` header, the negotiation/fallback policy and the
  build-time zstd dependency each written and tested exactly once. (Discarded
  alternative: have each stream compress its own body before building `HttpRequestSpec`
  — rejected because it duplicates the same fallback-on-rejection policy five times and
  guarantees drift between streams, the same problem `IBodyDecoder`'s own doc comment
  calls out on the manager side.)
- **In-memory bodies**: `statelessStream.cpp:197-201` (`flushOnce`, batch body),
  `controlStream.cpp:28-32` (`controlSpec`, used by the startup/notify/shutdown call
  sites at lines 241, 275, 337), `wpkFetcher.cpp:61-69` and `configFetcher.cpp:62-78`
  (both POST to `/download`), and `reporterStream.cpp:99-105` (`runPath()`, the
  periodic `/stats`/`/config` push). All set `spec.body`/`spec.bodyLength` before
  calling `m_sender.send()` — compressing centrally in `RetrySender` needs no change
  to these call sites beyond whatever opt-in signal is decided (see Open questions).
- **File-backed body**: `statefulStream.cpp:141-157` (`sendSession`) sets
  `spec.bodyFilePath`/`spec.bodyFileSize` — the spool file is streamed from disk, not
  held in memory. This is architecturally different and must be solved explicitly (see
  "Design considerations — file-backed compression").
- **Build system**: extend `src/Makefile`'s manager-only zstd guard
  (`ifneq ${TARGET},agent` / `ifneq ${TARGET},winagent`, lines 505-509) to include the
  agent/winagent targets, and CMake's `if(NOT IS_AGENT AND NOT IS_WINDOWS)` guard
  around the `zstd_external`/`ext_zstd` target (`src/external/CMakeLists.txt:1064`,
  1055-1090) to also cover agent, winagent, **and macOS** — that guard currently skips
  Windows even for the manager, so zstd-on-Windows is unproven in this repo's build
  today and needs its own verification pass, not just a flag flip.
- **Precompiled dependency artifacts.** Wazuh's external-deps pipeline fetches
  prebuilt tarballs from `packages.wazuh.com/deps/<DEPS_VERSION>/libraries/...`
  (`src/Makefile:414-416`, the `external-precompiled/%.tar.gz` rule at
  `src/Makefile:645-649`) before falling back to a from-source build; the manager's
  CMake mirrors the same pattern (`if(EXISTS .../libzstd.a) use precompiled`). Getting
  zstd onto agent CI/release builds at the same speed the manager already has means
  **generating and publishing those precompiled zstd artifacts for every agent
  platform/arch** (Linux, Windows, macOS), which likely means a `DEPS_VERSION` bump and
  coordinating with whatever pipeline currently produces those tarballs for the
  manager-only libraries.
- Distinguishing a compression-related `415`/`400` from a generic `Permanent` outcome
  (`classifyOutcome()`, `outcomeClassifier.cpp:14-73`) and wiring the fallback behavior
  per consumer: `statelessStream.cpp`'s `handleOutcome()` currently drops and counts
  every event in the batch on `Permanent` (`statelessStream.cpp:275-281`); `controlStream.cpp`
  feeds `result.outcome` straight into its own state machine (`eventFor()`,
  `controlStream.cpp:287-348`); `wpkFetcher.cpp:71-75` / `configFetcher.cpp:80-89` /
  `reporterStream.cpp:107-124` just log/back off and return, relying on being retried
  next cycle. `/stateful` (`statefulStream.cpp:158-162`) is a special case: its
  contract is interpreted from the raw HTTP status/body by `agent_sync_protocol`,
  *not* by `OutcomeClass` at all — so a `415`/`400` there needs its own analysis of
  what `agent_sync_protocol` does with an unexpected status, independent of the other
  five consumers.
- Tests: per-stream coverage that a compressed body round-trips against a mock
  performer, that `/stateless` budget accounting is unaffected, and that every
  consumer falls back to uncompressed instead of losing data/mis-transitioning on a
  simulated `415`/compression-`400`. Build verification that agent/winagent/macOS
  targets vendor and link zstd.

### Out of scope

- Any change to the manager (`src/remoted/remoted_module/`) — it is the reference
  contract (already generic across endpoints), not a target of this issue.
- `gzip` or any encoding other than zstd — the manager explicitly rejects `gzip`
  (`415`) regardless of endpoint.
- Compressing the manager's *download response* bytes (the actual WPK/`merged.mg`
  content served back to the agent by `/download`) — that direction is the subject of
  the separate, still-open spike #38282. This issue only touches the agent's own
  *request* bodies (including the small POST body `wpkFetcher`/`configFetcher` send
  to ask for a download); it does not depend on or block #38282's outcome.

## Current state (verified on `5.0.0-https`, fetched from `origin` — post-#38129)

- **No compression code exists on the agent side.**
- **Manager decoding is endpoint-agnostic.** `AuthGateway` runs one shared
  `IBodyDecoder` for `/stateless`, `/download`, `/stats`, `/config`, `/stateful` and
  `/control` alike (`authGateway.cpp`, `remotedModuleFacade.hpp:279-390`,
  `decoding/iBodyDecoder.hpp`). Only `zstd` is recognized (`ContentEncoding` is
  `None | Zstd | Unsupported`).
- **Signing already covers whatever bytes are in the spec at send time.**
  `RetrySender::attemptOnce()` signs `attempt.body`/`attempt.bodyLength` (in-memory)
  or reads `attempt.bodyFilePath` via `signFile()` (file-backed) — `retrySender.cpp:79-82`.
  Compressing before that point, for in-memory bodies, automatically makes the CMAC
  cover the compressed wire bytes with no change to `cmacSigner.cpp`/`canonicalRequest.cpp`.
- **The file-backed path signs and uploads from the same file independently.**
  `CmacSigner::signFile()` opens and reads `bodyFilePath` itself
  (`cmacSigner.cpp:165`), and `curlPerformer.cpp:91` opens and reads that same path
  again for the actual upload — two independent reads of one file on disk, not a
  single pass. For `/stateful`, the spool file must **already contain the compressed
  bytes** before either read happens; compressing "in flight" during upload is not an
  option without changing this two-reader structure.
- **`/stateless` budget accounting is content-size based and must stay that way.**
  `AdaptivePayload` (`adaptivePayload.hpp`) and
  `StatelessStream::eventBytesBudgetLocked()` (`statelessStream.cpp:322-327`) size
  batches off `config.batchSizeBytes`, matching the manager's decompressed-size cap;
  compression must only shrink the final wire bytes, never feed back into that sizing
  decision.
- **zstd is a manager-only, non-Windows build dependency today.**
  `src/Makefile:505-509` adds `zstd` to `EXTERNAL_RES` only for
  `TARGET != agent && TARGET != winagent`. `src/external/CMakeLists.txt:1055-1090`
  guards the `zstd_external`/`ext_zstd` target with `if(NOT IS_AGENT AND NOT
  IS_WINDOWS)` — so even the manager's own zstd support has never been built or
  proven on Windows in this repo.
- **No config toggle exists for this yet.** `ModuleConfig`
  (`src/client-agent/https_client/src/moduleConfig.hpp`) has no compression-related
  field; existing boolean feature toggles follow an `xEnabled` naming pattern
  (`statsEnabled`, `configReportEnabled`).

## Tasks

- [x] Resolve the design questions (compression choke point, file-backed
  strategy, negotiation/fallback policy, compression level/threshold) before writing
  code.
- [ ] Vendor zstd for `TARGET=agent`/`TARGET=winagent` (`src/Makefile`) and extend the
  CMake `IS_AGENT`/`IS_WINDOWS` guard around `zstd_external`/`ext_zstd`
  (`src/external/CMakeLists.txt`) to cover agent, winagent and macOS; verify the build
  on all three platforms (Windows is currently unproven even for the manager). Linux
  and Windows verified for the original 5 send paths; macOS still unverified, and the
  `/stateful` file-backed compression files added afterward have not yet been
  cross-compiled/run on Windows either (no MinGW toolchain in the environment used for
  that follow-up round) — see PR #38355's own evidence for the exact split.
- [ ] Generate and publish the precompiled zstd artifacts the agent build needs to
  `packages.wazuh.com/deps/...` (coordinate the `DEPS_VERSION` bump). Still deferred to
  a follow-up, per PR #38355's description.
- [x] Implement zstd compression + `Content-Encoding: zstd` + the chosen
  negotiation/fallback policy centrally in `RetrySender`.
- [x] Solve file-backed compression for `/stateful` per the decided strategy. A
  `ZstdFileCompressor` compresses the spooled session once, up front, and
  `RetrySender` swaps in the precompressed sibling file per attempt the same way it
  already does for in-memory bodies — no changes needed to the shared retry loop.
  Verified live against a real manager (decrypted capture + independent zstd
  decompression of the actual wire bytes) — see PR #38355.
- [x] Add the config toggle/threshold decided above to `ModuleConfig` (+ XML parser)
  if the chosen approach needs one.
- [x] Add a distinct classification for a compression-related `415`/`400` in
  `classifyOutcome()` and route each consumer's fallback per the "Per-consumer
  rejection handling" analysis above (`/stateless`, `/control`, `/download`,
  `/stats`+`/config`), plus the separate analysis for `/stateful`'s non-`OutcomeClass`
  path (added as its own `case 415` in `agent_sync_protocol.cpp`, mapped like the
  existing `503` handling).
- [ ] Tests: compressed round trip per consumer against a mock performer;
  `/stateless` budget accounting unaffected by compression; fallback-to-uncompressed
  on a simulated rejection for every consumer; build verification per platform. All
  done except the per-platform build verification gap noted above (macOS, and
  Windows for the `/stateful`-specific files) — leaving this unchecked until that's
  closed.
- [x] Update developer documentation covering the affected send paths (module header
  comments, per this codebase's existing convention) with the compression behavior and
  the fallback contract.

## Definition of Done

- `/stateless`, `/control`, `/download`, the `/stats`+`/config` push, and `/stateful`
  all send requests zstd-compressed with `Content-Encoding: zstd`, signed over the
  compressed bytes, when compression is active.
- The agent/winagent/macOS builds vendor and link zstd, backed by precompiled
  artifacts published for every agent platform/arch — no from-source build required on
  every CI run. **Not yet done**: precompiled-artifact publishing is deferred to a
  follow-up, and macOS build verification (and a Windows run of the `/stateful`-specific
  files) is still outstanding — see the Tasks list above.
- `/stateless` batch sizing and the `413` split/resend flow (#37835) are unaffected by
  compression.
- A manager that rejects the encoding never causes permanent event loss or a
  mis-transitioned state machine on any consumer — every affected path retries
  uncompressed at least once.
- Tests green; astyle clean.

## References

- #37702 — HTTPS comms epic.
- #37831 — Agent-side HTTPS migration breakdown (parent).
- #37828 — HTTPS client foundation (agentd integration). #37830 — `/control` lifecycle.
  #37832 — `/download` client. #37835 — `/stateless` events pipeline. #37836 —
  `/stateful` single-request sessions. #37843 — periodic `/stats`+`/config` push. All
  closed/merged; each is a consumer this issue adds compression to.
- #38282 — open spike on compressing the manager's *download response* bytes
  (WPK/`merged.mg`) — the opposite direction from this issue; read together, not as a
  duplicate.
- #38109 — "Add gzip request body support to the remoted HTTP server" (manager-side
  ask; superseded in practice by zstd-only support).
- PR #38129 — "Add zstd request body support" (manager-side implementation; its
  description describes `/stateless` only — the endpoint-agnostic `AuthGateway`
  wiring confirmed against the current `5.0.0-https` tip is a *later* refactor on top
  of it, so read the code, not just the PR description, when picking this up).
- Code (agent): `src/client-agent/https_client/src/{statelessStream,controlStream,
  statefulStream,wpkFetcher,configFetcher,retrySender,cmacSigner,curlPerformer,
  outcomeClassifier,adaptivePayload,httpTypes}.{hpp,cpp}`; `src/Makefile`,
  `src/external/CMakeLists.txt` (zstd vendoring, currently manager-only and
  non-Windows).
- Code (manager, reference contract only): `src/remoted/remoted_module/src/endpoints/
  authGateway.{hpp,cpp}`, `src/remoted/remoted_module/src/decoding/{iBodyDecoder,
  bodyDecoder}.{hpp,cpp}`, `src/remoted/remoted_module/src/common/zstdDecoder.{hpp,cpp}`,
  `src/remoted/remoted_module/src/remotedModuleFacade.hpp`.

lchico:
## Description

Adds optional zstd request-body compression (`Content-Encoding: zstd`) to the agent's outbound HTTPS traffic, mirroring the manager-side decode already merged in #38129. Every agent-to-manager send path — `/stateless`, `/control`, `/download` (WPK + config fetch), the periodic `/stats`+`/config` push, and now `/stateful` too — flows through one shared choke point (`RetrySender::attemptOnce()`), so compression is implemented once, there, instead of six times. `/stateful`'s body is a spool file read twice, independently, by the signer and the performer, which is why it needed its own design: a new `ZstdFileCompressor` compresses the spooled session once, up front (not per retry attempt), and hands `RetrySender` a precompressed sibling file it swaps in per attempt exactly like it already does for in-memory bodies — no changes to the shared retry loop itself.

The toggle is off by default and lives in `internal_options.conf` (`agent.https_compression_enabled`), not `client.conf` XML — this is an opt-in tuning knob, not user-facing configuration. On a `415` (the manager rejects `Content-Encoding: zstd`), the agent falls back permanently for the rest of its run: the in-flight request retries once, uncompressed, and a shared gate disables compression for every send path from then on — not a per-request retry, and not per-stream. zstd is vendored for every agent platform, including the Windows agent (winagent) — cross-compiled via MinGW, not stubbed out; publishing precompiled build artifacts for the packaging pipeline is the one piece still deferred to a follow-up.

Closes #38308.

## Proposed Changes

- **Build enablement**
  - `src/Makefile`: zstd moves out of the manager-only `EXTERNAL_RES` block into the base list every target already builds from — manager, Linux/macOS agent, and winagent alike.
  - `src/external/CMakeLists.txt`: zstd's build guard widens from `NOT IS_AGENT AND NOT IS_WINDOWS` to unconditional, with a Windows-specific branch that cross-compiles it via a nested `ExternalProject_Add` forwarding the already-cross-compiling top-level `CMAKE_C_COMPILER`/`CMAKE_CXX_COMPILER`/`CMAKE_AR`/`CMAKE_RANLIB` (mirrors this file's own GoogleTest Windows block) — its own `BINARY_DIR` so a Linux-built `libzstd.a` already on disk can never satisfy the Windows precompiled-artifact check or vice versa.
  - `src/client-agent/https_client/CMakeLists.txt`: links `ext_zstd` into `https_client`, the same target `remoted_module` already links for the manager-side decode. Also adds `-Wl,--exclude-libs,libzstd.a` to the winagent link flags: MinGW auto-exports every global symbol from a DLL by default (unlike ELF), so without this, zstd's own public API leaks into `https_client`'s export table and collides with the same symbols the unit test binary links directly (needed there to independently decompress and verify compressed bytes) — a real `multiple definition` link error without the fix, not a theoretical one.

- **Config toggle**
  - `etc/internal_options.conf`: new `agent.https_compression_enabled` (default `0`).
  - `src/client-agent/src/https_client_bridge.c`: reads it via `getDefine_Int_default` (defaults to off instead of aborting if the key is ever missing) into `hc_config_t`.
  - `include/https_client.h`, `moduleConfig.{hpp,cpp}`: threads the new field through to the C++ side.

- **Compression mechanism**
  - `retrySender.cpp/hpp`: `attemptOnce()` zstd-compresses in-memory bodies (level 3) immediately before signing, so the CMAC covers the compressed wire bytes, and adds the `Content-Encoding: zstd` header. A one-shot `ZSTD_compress()` failure — never expected, given the buffer is sized with `ZSTD_compressBound()` — falls through to sending uncompressed rather than losing the request. File-backed bodies (`/stateful`) take a parallel branch: if the caller populated `HttpRequestSpec::precompressedBodyFilePath` (see below), that file/size is swapped in instead, fresh on every attempt — so a `415` disabling the gate makes the very next attempt fall through to the original, untouched file automatically, with zero changes to `send()`'s retry loop.
  - `outcomeClassifier.cpp` / `httpTypes.hpp`: new `OutcomeClass::CompressionRejected` for HTTP `415`.
  - `compressionGate.hpp` (new): a one-way latch, shaped like the existing `AuthGate` — `reportRejected()` disables compression agent-wide; no `release()`, since the manager's lack of zstd support doesn't change without a restart.
  - `httpsClientFacade.cpp/hpp`: one `CompressionGate` per agent, shared into every stream's `RetrySender` (`statelessStream`, `statefulStream`, `controlStream`, `reporterStream`, `wpkFetcher`, `configFetcher`) — a `415` on any one of them disables compression on all of them.
  - `retrySender.cpp`'s `send()`: on `CompressionRejected`, reports to the shared gate first, then retries once — `attemptOnce()`'s own gate check means that retry is naturally uncompressed, with no separate "forced" parameter needed.
  - `fileCompressor.cpp/hpp` (new): `ZstdFileCompressor` streams a spool file through zstd in bounded 64 KiB chunks (matching `cmacSigner.cpp::signFile()`'s own chunking) into a new temp file, so a multi-MB `/stateful` session never sits fully in memory either as source or as compressed output. Checks a cooperative abort flag between chunks so compressing a large session can't stall agent shutdown.
  - `exclusiveTempFile.cpp/hpp` (new): the atomic, owner-only-permissions, symlink-attack-resistant temp file creation `TempSpoolFactory` already had, extracted into a primitive both it and `ZstdFileCompressor` share — including a fallback to `TMPDIR`/`TEMP`/`TMP`/`/tmp` when the caller's directory is empty, since nothing populates `ModuleConfig::spoolDir` today (a real bug this caught live — see Results and Evidence below).
  - `httpTypes.hpp`: `HttpRequestSpec` gains `precompressedBodyFilePath`/`precompressedBodyFileSize`, populated only by `/stateful` — every other send path leaves them empty/0, unaffected.
  - `statefulStream.cpp/hpp`: `sendSession()` compresses the spooled session once, up front, when enabled and the gate still allows it, holding the compressed `SpoolFile` alive for exactly the duration of the `send()` call.
  - `agent_sync_protocol.cpp`: `/stateful`'s own contract is interpreted from the raw HTTP status, not the shared `OutcomeClass` classifier above — it had no case for `415` at all, falling into a non-retryable `PROTOCOL_ERROR`. Added one, mapped like the existing `503` case, as defense-in-depth for the rare compound failure where `RetrySender`'s own one-shot retry doesn't land within the attempt budget.

- **Tests**
  - `retrySender_test.cpp`: 6 new cases covering compression on/off leaving body/headers untouched appropriately, the CMAC signing the compressed bytes, file-backed bodies being skipped, a `415` retrying once uncompressed and succeeding, that rejection disabling the shared gate for *other* senders, and a second rejection not looping.
  - `statelessStream_test.cpp`: 2 new cases — a compression rejection recovers and still delivers the event exactly once, and compression doesn't change the event-batching budget accounting.
  - `outcomeClassifier_test.cpp`: `415` → `CompressionRejected` added to the parameterized outcome table.
  - `moduleConfig_test.cpp`, `test_https_client_bridge.c`: default-off coverage for the new toggle.
  - `fileCompressor_test.cpp` (new, 7 cases): round-trip compress/decompress correctness, the pledged content size lands in the frame header, unreadable source, unwritable spool dir, empty spool dir (the bug above), mid-stream abort, and an empty source producing a valid minimal frame.
  - `retrySender_test.cpp`: 4 more new cases for the file-backed path — the precompressed-file swap signs over the compressed bytes, a `415` falls back to the original file and retries successfully, a second rejection with no gate still terminates without looping, and a file-backed sender correctly skips compression once another sender's `415` already disabled the shared gate.
  - `statefulStream_test.cpp`: 3 new cases — compression happens once and the temp file is deleted after `send()` returns, a compression failure falls back to the original file, and an already-disabled gate skips the compressor entirely.
  - `httpsClient_component_test.cpp`: a real curl request against the fake manager proves a compressed `/stateful` session shrinks on the wire and decompresses back to the original spooled bytes (needed linking `ext_zstd` into this test target, which didn't have it before).
  - `test_agent_sync_protocol.cpp`: the new `415` case, mirroring the existing `503` test.

- **Packaging**
  - `.github/actions/check_files/deb_linux_agent_amd64.csv`: `libhttps_client.so`'s expected installed size updated (zstd is now statically linked into it), measured from a real CI build of this branch rather than estimated. Only this one manifest was touched — the other platform manifests either use a different baseline or have no size check at all, so they're left alone pending their own measured evidence.

- **Cleanup**
  - Comments across the above files no longer reference this issue's own number — kept as durable explanations of *why*, not a changelog entry that rots once the issue closes. References to other, already-completed issues (e.g. prior `/stateless` and auth-gate work) are left as-is, since they still describe relevant prior design decisions.

#### Automated test suites, by platform

<details>
<summary>Linux — 327/327 unit tests, 39/39 component tests (includes the `/stateful` file-backed compression work)</summary>

**Unit tests**

327/327 passing — 8 compression-specific cases on the original 5 send paths (6 in `RetrySenderTest`, 2 in `StatelessStreamTest`) plus the new `outcomeClassifier`/`moduleConfig`/bridge coverage, and 13 more for `/stateful`'s file-backed path (7 in `FileCompressorTest`, 4 more in `RetrySenderTest`, 3 in `StatefulStreamTest`; one of those 4 `RetrySenderTest` cases, `AuthFailureFromTheCompressionRetryStillGetsItsOwnGraceRetry`, predates `/stateful` and is listed here for completeness).

Reproduce from the repo root:

```
cd src
make deps TARGET=agent -j"$(nproc)"          # first build only
make TARGET=agent TEST=1 -j"$(nproc)"
./build/bin/https_client_unit_test
```

```
[==========] Running 327 tests from 31 test suites.
[==========] 327 tests from 31 test suites ran. (203 ms total)
[  PASSED  ] 327 tests.
```

The compression-specific subset in isolation:

```
./build/bin/https_client_unit_test --gtest_filter='RetrySenderTest.*Compress*:RetrySenderTest.*CompressionRejection*:RetrySenderTest.SecondCompressionRejectionTerminatesWithoutLooping:*OutcomeClassifier*:StatelessStreamTest.Compression*:StatefulStreamTest.*Compress*:FileCompressorTest.*'
```

```
[----------] 7 tests from FileCompressorTest
[       OK ] FileCompressorTest.RoundTripCompressesAndDecompressesBackToTheSource
[       OK ] FileCompressorTest.CompressedFrameDeclaresThePledgedContentSize
[       OK ] FileCompressorTest.UnreadableSourceReturnsNullopt
[       OK ] FileCompressorTest.UnwritableSpoolDirReturnsNullopt
[       OK ] FileCompressorTest.EmptySpoolDirFallsBackToADefaultTempDirectory
[       OK ] FileCompressorTest.AbortFlagStopsCompressionAndReturnsNullopt
[       OK ] FileCompressorTest.EmptySourceProducesAValidMinimalFrame
[----------] 2 tests from OutcomeClassifierTest
[----------] 10 tests from RetrySenderTest
[       OK ] RetrySenderTest.CompressionDisabledLeavesBodyAndHeadersUnchanged
[       OK ] RetrySenderTest.CompressedBodyIsSignedOverTheCompressedBytes
[       OK ] RetrySenderTest.CompressionEnabledSkipsFileBackedBodies
[       OK ] RetrySenderTest.CompressedAttemptRejectedWith415RetriesOnceUncompressedAndSucceeds
[       OK ] RetrySenderTest.CompressionRejectionDisablesTheSharedGateForOtherSenders
[       OK ] RetrySenderTest.SecondCompressionRejectionTerminatesWithoutLooping
[       OK ] RetrySenderTest.AuthFailureFromTheCompressionRetryStillGetsItsOwnGraceRetry
[       OK ] RetrySenderTest.CompressedFileBackedAttemptRejectedWith415RetriesOnceUncompressedAndSucceeds
[       OK ] RetrySenderTest.SecondFileBackedCompressionRejectionTerminatesWithoutLooping
[       OK ] RetrySenderTest.FileBackedSenderSkipsCompressionOnceTheSharedGateIsAlreadyDisabled
[----------] 3 tests from StatefulStreamTest
[       OK ] StatefulStreamTest.CompressionEnabledCompressesOnceAndCleansUpTheTempFile
[       OK ] StatefulStreamTest.CompressionFailureFallsBackToTheOriginalFileUncompressed
[       OK ] StatefulStreamTest.GateAlreadyDisabledSkipsCompressionEntirely
[----------] 2 tests from StatelessStreamTest
[       OK ] StatelessStreamTest.CompressionRejectionRecoversAndDeliversEventExactlyOnce
[       OK ] StatelessStreamTest.CompressionDoesNotChangeEventBatchingBudget
[----------] 20 tests from D9Table/OutcomeClassifierTable
[==========] 44 tests from 6 test suites ran. (3 ms total)
[  PASSED  ] 44 tests.
```

**Component tests**

39/39 passing:

```
./build/bin/https_client_component_test
```

```
[==========] Running 39 tests from 6 test suites.
[==========] 39 tests from 6 test suites ran. (22126 ms total)
[  PASSED  ] 39 tests.
```

(One earlier run, before the `/stateful` work landed, showed a single failure, `FacadeE2eTest.AHeldStatefulRequestDoesNotStallStatelessOrControl` — a pre-existing timing-based e2e test for a prior, unrelated `/stateful` concurrency guarantee, in a file this PR never touches. It passed 3/3 in isolation and cleanly on every subsequent full-suite run, including the one above, consistent with contention-driven flakiness under parallel test execution rather than a regression from this change.)

</details>

<details>
<summary>Windows (winagent, cross-compiled via MinGW) — 310/310 unit tests, verified both under Wine and on a real Windows 11 machine</summary>

Cross-compiling for winagent isn't just "does it build" — the full chain was verified: the zstd library itself cross-compiles and links into `libhttps_client.dll` (confirmed via `nm`: real `ZSTD_compress`/`ZSTD_compressBound` symbols present in the binary, and via `strings`: the literal `Content-Encoding: zstd` string is embedded in it), and the same gtest suite that runs on Linux was built and actually *executed* for Windows, twice, independently:

- Once emulated, via Wine (`i686`, run on the Linux build host).
- Once on a real, native Windows 11 machine — no emulation at all.

Both runs: **310/310 passing**, including all 8 compression-specific tests (the two-fewer-than-Linux's-312 gap is a pre-existing, unrelated platform exclusion elsewhere in the suite, not something this PR introduced or touches).

Reproduce (needs a MinGW cross-compile toolchain, e.g. `gcc-mingw-w64 g++-mingw-w64`, and `wine`+`wine32`/i386 multiarch to run the tests without real Windows hardware):

```
cd src
make deps TARGET=winagent -j"$(nproc)"
make TARGET=winagent TEST=1 -j"$(nproc)"
wine ./build/bin/https_client_unit_test.exe        # or copy it to a real Windows machine and run it directly
```

```
[==========] Running 310 tests from 30 test suites.
[==========] 310 tests from 30 test suites ran. (8511 ms total)
[  PASSED  ] 310 tests.
```

```
[ RUN      ] RetrySenderTest.CompressionDisabledLeavesBodyAndHeadersUnchanged
[       OK ] RetrySenderTest.CompressionDisabledLeavesBodyAndHeadersUnchanged (0 ms)
[ RUN      ] RetrySenderTest.CompressedBodyIsSignedOverTheCompressedBytes
[       OK ] RetrySenderTest.CompressedBodyIsSignedOverTheCompressedBytes (12 ms)
[ RUN      ] RetrySenderTest.CompressionEnabledSkipsFileBackedBodies
[       OK ] RetrySenderTest.CompressionEnabledSkipsFileBackedBodies (4 ms)
[ RUN      ] RetrySenderTest.CompressedAttemptRejectedWith415RetriesOnceUncompressedAndSucceeds
[       OK ] RetrySenderTest.CompressedAttemptRejectedWith415RetriesOnceUncompressedAndSucceeds (0 ms)
[ RUN      ] RetrySenderTest.CompressionRejectionDisablesTheSharedGateForOtherSenders
[       OK ] RetrySenderTest.CompressionRejectionDisablesTheSharedGateForOtherSenders (8 ms)
[ RUN      ] RetrySenderTest.SecondCompressionRejectionTerminatesWithoutLooping
[       OK ] RetrySenderTest.SecondCompressionRejectionTerminatesWithoutLooping (7 ms)
[ RUN      ] StatelessStreamTest.CompressionRejectionRecoversAndDeliversEventExactlyOnce
[       OK ] StatelessStreamTest.CompressionRejectionRecoversAndDeliversEventExactlyOnce (411 ms)
[ RUN      ] StatelessStreamTest.CompressionDoesNotChangeEventBatchingBudget
[       OK ] StatelessStreamTest.CompressionDoesNotChangeEventBatchingBudget (168 ms)
```

**Operational note for anyone deploying the raw agent binaries manually (outside the installer):** `libhttps_client.dll` depends on `libwazuhext.dll` at runtime — not obvious from the build output, easy to miss if you're only copying the two files you'd expect.

**Known gap — this Windows run predates the `/stateful` work.** Everything above (zstd cross-compiling into the DLL, both Wine and real-Windows-11 runs, the 310/310 count) covers the original 5 send paths only. `fileCompressor.cpp/hpp`, `exclusiveTempFile.cpp/hpp`, and the `/stateful`-side changes have **not** been cross-compiled or run for `winagent` — this follow-up round only had a Linux Docker environment available, with no MinGW toolchain. Nothing in the new code is platform-specific (no `#ifdef _WIN32` branches, same `ZSTD_compressStream2` API already proven to cross-compile above), but that's an expectation, not evidence. Flagging this explicitly rather than letting the 310/310 count above imply coverage it doesn't have; a winagent build+test run for these specific files is still owed before merge.

</details>

#### Live agent↔manager evidence

Self-comparison on one real agent build from this branch, enrolled against a real manager: `agent.https_compression_enabled` toggled OFF vs. ON, same build, same config, same workload, restarting the agent between phases so the change actually takes effect.

Bandwidth is measured with a `tcpdump` capture scoped to the manager's host and HTTPS port, running for the duration of each workload window; the reported byte count sums the frame lengths of packets sourced from the agent's own IP (egress-only, isolated from any other traffic on the interface):

| Send path | OFF (bytes) | ON (bytes) | Reduction |
|---|---|---|---|
| `/stateless` (500× a compressible test event) | 130886 | 23317 | 82.2% |
| `/control` (restart handshake, 3s window) | 11676 | 11281 | 3.4% |
| `/stats` + `/config` (periodic push, 25s window) | 43529 | 30497 | 29.9% |
| `/download` (WPK + config fetch) | not exercised live — needs a deliberate config-hash mismatch or a remote-upgrade task; not attempted in this run | | |
| `/stateful` | see the dedicated section below | | |

Reproduce (against any manager + agent pair with this build installed on the agent, HTTPS port `1517`):

```
# On the agent, toggle and restart:
echo 'agent.https_compression_enabled=1' >> /var/ossec/etc/local_internal_options.conf   # or =0 for the baseline
systemctl restart wazuh-agent

# Capture for the duration of a workload window (a burst of log lines for
# /stateless, the restart itself for /control, or a plain wait for the
# periodic /stats+/config push):
tcpdump -i eth0 -w capture.pcap 'host <manager-address> and tcp port 1517'
# ... let the workload run, then stop tcpdump ...

# Sum egress-only bytes:
tshark -r capture.pcap -Y 'ip.src==<agent-ip>' -T fields -e frame.len \
  | awk '{s+=$1} END{print s+0}'
```

**Regression check:** the agent's HTTPS client stayed connected throughout, including immediately after enabling compression and restarting — no repeated "failed to communicate with the manager" in the log.

#### TLS-decrypted mechanism proof

Byte-count deltas prove the wire payload shrank; to prove *why*, the same captures were decrypted and the actual HTTP headers inspected. The agent process was started with `SSLKEYLOGFILE` set (a `systemd` drop-in `Environment=SSLKEYLOGFILE=<path>` on the `wazuh-agent` service is enough — the agent's vendored, statically-linked curl+OpenSSL honors it with no code change), and `tshark` decrypted each capture with that keylog:

| Phase | Decrypted `Content-Encoding` values seen | Sample decrypted request |
|---|---|---|
| OFF | none | `POST /stateless` — no `Content-Encoding` header, 99836-byte body |
| ON | `zstd` | `POST /stateless` — `Content-Encoding: zstd`, 388-byte body |

Reproduce, against the same capture used above:

```
tshark -r capture.pcap -o tls.keylog_file:<sslkeylogfile-path> \
  -Y 'http.request and http.content_encoding' \
  -T fields -e http.request.uri -e http.content_encoding
```

Decrypting every send path's ON-phase capture together (not just `/stateless`) shows the same `Content-Encoding` applied uniformly, confirming compression really does happen once, at one shared point, rather than being duplicated per endpoint:

| Request URI | Content-Encoding |
|---|---|
| `/config` | zstd |
| `/control` | zstd |
| `/stateless` | zstd |
| `/stats` | zstd |

#### `/stateful` file-backed compression

The 327/39 Linux counts above already include this work's 13 new unit tests and 1 new component test. Also run clean under ASan + LeakSanitizer + UBSan (`FSANITIZE=1`) — only a pre-existing, unrelated `moduleConfig.cpp` UBSan finding remains (undefined behavior in `verify_mode` validation, nothing to do with compression). Not yet cross-compiled or run for `winagent` — see the Windows section's own caveat above.

Live, against the real manager — same self-comparison method as above:

Before (captured with the `spoolDir`-empty bug still live, so it's a real "compression silently not happening" sample, not a synthetic baseline):
```
$ tshark -r cap_stateful_compress.pcap -o tls.keylog_file:agent_sslkeylog.txt \
    -Y 'http.request and http.request.uri == "/stateful"' \
    -T fields -e http.request.method -e http.request.uri -e http.content_encoding -e http.content_length

POST	/stateful		440
POST	/stateful		610656
POST	/stateful		610656
POST	/stateful		610656
POST	/stateful		610656
POST	/stateful		610656
```

After (same agent, same workload, only the `spoolDir` fallback fix applied):
```
$ tshark -r cap_stateful_fix.pcap -o tls.keylog_file:agent_sslkeylog.txt \
    -Y 'http.request and http.request.uri == "/stateful"' \
    -T fields -e http.request.method -e http.request.uri -e http.content_encoding -e http.content_length

POST	/stateful	zstd	288
POST	/stateful	zstd	79123
POST	/stateful	zstd	79123
POST	/stateful	zstd	79123
POST	/stateful	zstd	79123
POST	/stateful	zstd	79123
```

Not just a header — the wire bytes are genuinely valid, correct zstd. Pulled the actual request body for one of the `79123`-byte frames out of the decrypted capture and decompressed it with the real zstd library:
```
$ tshark -r cap_stateful_fix.pcap -o tls.keylog_file:agent_sslkeylog.txt \
    -Y 'frame.number==315' -T pdml | grep -o 'name="data" value="[a-f0-9]*"'
# extracted 79123 bytes, first 4 bytes: 28 b5 2f fd  (the real zstd magic number)

$ python3 -c "
import ctypes
zstd = ctypes.CDLL('libzstd.so.1')
data = open('frame315_body.zst','rb').read()
zstd.ZSTD_getFrameContentSize.restype = ctypes.c_ulonglong
print('declared decompressed size:', zstd.ZSTD_getFrameContentSize(data, len(data)))
buf = ctypes.create_string_buffer(610656)
zstd.ZSTD_decompress.restype = ctypes.c_size_t
print('actually decompressed:', zstd.ZSTD_decompress(buf, 610656, data, len(data)), 'bytes')
"

declared decompressed size: 610656
actually decompressed: 610656 bytes
```
`610656` is exactly the uncompressed byte count from the "before" capture — the compressed frame that actually went out on the wire decompresses back to precisely the original session size, byte for byte.

**One real bug this caught:** the first live attempt sent uncompressed despite everything passing in tests — traced to `ModuleConfig::spoolDir` being empty in production (nothing populates it) with no fallback in the new compressor, unlike `TempSpoolFactory`'s own constructor. Fixed by moving the fallback into the shared `exclusiveTempFile` primitive, added a regression test (`EmptySpoolDirFallsBackToADefaultTempDirectory`), reconfirmed live (the "after" capture above).

  - [x] Linux
  - [ ] Windows — true for the original 5 send paths (see the Windows evidence section above), but the `/stateful` file-backed compression files added in this round have not yet been cross-compiled for `winagent`; unchecking until that run happens
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [x] AddressSanitizer (+ LeakSanitizer + UBSan, `FSANITIZE=1` -- whole `https_client` suite, not just `/stateful`)
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - N/A

  - N/A

- Wazuh server API/Framework
  - N/A

### Artifacts Affected

- `libhttps_client.so` (Linux agent, amd64): grows because zstd is now statically linked in — packaging manifest (`deb_linux_agent_amd64.csv`) updated with the real measured size.
- `libhttps_client.dll` (winagent): same growth, zstd statically linked in; no Windows packaging manifest currently checks this library's size, so no manifest update needed there.
- No new executables or packages.

### Configuration Changes

- New `internal_options.conf` key: `agent.https_compression_enabled` (`0`/`1`, default `0`). This is an internal tuning knob, not a `client.conf`/XML setting, and is fully backward-compatible — an agent with no opinion on it behaves exactly as before.

### Tests Introduced

- 6 new `RetrySenderTest` cases and 2 new `StatelessStreamTest` cases directly exercising compress/retry/gate behavior, plus a new `outcomeClassifier` parameterized case and default-off coverage in `moduleConfig`/bridge tests — see Results and Evidence above for the full run.
- The shared-gate-across-streams behavior (one `415` disabling compression for every send path) is covered at the unit level via direct constructor injection of a shared `CompressionGate`, rather than a new component/e2e test — it doesn't depend on real sockets, so the unit-level test is the more direct proof.
- `/stateful`: a new `fileCompressor_test.cpp` (7 cases) plus 4 more `RetrySenderTest` cases, 3 more `StatefulStreamTest` cases, a new component test proving a real compressed session decompresses correctly, and a new `test_agent_sync_protocol.cpp` case for the `415` handling — see the dedicated evidence section above.

## Review Checklist

- [ ] Code changes reviewed
- [x] Developer documentation reflects the changes (module header/Doxygen comments updated for the affected files, per this codebase's convention -- no separate README exists for this module)
- [ ] ...

### DIFF
--- a/.github/actions/check_files/deb_linux_agent_amd64.csv
+++ b/.github/actions/check_files/deb_linux_agent_amd64.csv
@@ -10,7 +10,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12374168,0.1
 /var/ossec/lib/libdbsync.so,root,wazuh,750,file,-rwxr-x---,505104,0.1
 /var/ossec/lib/libagent_sync_protocol.so,root,wazuh,750,file,-rwxr-x---,241184,0.1
-/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,476176,0.1
+/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,1153344,0.1
 /var/ossec/lib/libschema_validator.so,root,wazuh,750,file,-rwxr-x---,390232,0.1
 /var/ossec/lib/libsysinfo.so,root,wazuh,750,file,-rwxr-x---,2603392,0.1
 /var/ossec/lib/libfimdb.so,root,wazuh,750,file,-rwxr-x---,253344,0.1
--- a/.github/actions/check_files/deb_linux_agent_i386.csv
+++ b/.github/actions/check_files/deb_linux_agent_i386.csv
@@ -10,7 +10,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12358288,0.1
 /var/ossec/lib/libdbsync.so,root,wazuh,750,file,-rwxr-x---,525380,0.1
 /var/ossec/lib/libagent_sync_protocol.so,root,wazuh,750,file,-rwxr-x---,216020,0.1
-/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,381752,0.1
+/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,1058920,0.1
 /var/ossec/lib/libschema_validator.so,root,wazuh,750,file,-rwxr-x---,426324,0.1
 /var/ossec/lib/libsysinfo.so,root,wazuh,750,file,-rwxr-x---,2792476,0.1
 /var/ossec/lib/libfimdb.so,root,wazuh,750,file,-rwxr-x---,273712,0.1
--- a/.github/actions/check_files/rpm_linux_agent_amd64.csv
+++ b/.github/actions/check_files/rpm_linux_agent_amd64.csv
@@ -74,7 +74,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12417720,0.1
 /var/ossec/lib/libdbsync.so,root,wazuh,750,file,-rwxr-x---,484376,0.1
 /var/ossec/lib/libagent_sync_protocol.so,root,wazuh,750,file,-rwxr-x---,249336,0.1
-/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,455704,0.1
+/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,1132872,0.1
 /var/ossec/lib/libschema_validator.so,root,wazuh,750,file,-rwxr-x---,386024,0.1
 /var/ossec/lib/libsysinfo.so,root,wazuh,750,file,-rwxr-x---,2555328,0.1
 /var/ossec/lib/libfimdb.so,root,wazuh,750,file,-rwxr-x---,224672,0.1
--- a/.github/actions/check_files/rpm_linux_agent_i386.csv
+++ b/.github/actions/check_files/rpm_linux_agent_i386.csv
@@ -74,7 +74,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12362096,0.1
 /var/ossec/lib/libdbsync.so,root,wazuh,750,file,-rwxr-x---,541828,0.1
 /var/ossec/lib/libagent_sync_protocol.so,root,wazuh,750,file,-rwxr-x---,216520,0.1
-/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,365376,0.1
+/var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,1042544,0.1
 /var/ossec/lib/libschema_validator.so,root,wazuh,750,file,-rwxr-x---,398136,0.1
 /var/ossec/lib/libsysinfo.so,root,wazuh,750,file,-rwxr-x---,2717844,0.1
 /var/ossec/lib/libfimdb.so,root,wazuh,750,file,-rwxr-x---,259032,0.1
--- a/etc/internal_options.conf
+++ b/etc/internal_options.conf
@@ -122,7 +122,7 @@ agent.tolerance=15
 agent.warn_level=90
 # Level of occupied capacity in Agent buffer to come back to normal state
 agent.normal_level=70
-# Compress HTTPS request bodies with zstd (#38308)
+# Compress HTTPS request bodies with zstd
 # 0. Disabled
 # 1. Enabled
 agent.https_compression_enabled=0
--- a/etc/internal_options.conf
+++ b/etc/internal_options.conf
@@ -122,6 +122,10 @@ agent.tolerance=15
 agent.warn_level=90
 # Level of occupied capacity in Agent buffer to come back to normal state
 agent.normal_level=70
+# Compress HTTPS request bodies with zstd (#38308)
+# 0. Disabled
+# 1. Enabled
+agent.https_compression_enabled=0
 # Minimum events per second, configurable at XML settings [1..1000]
 agent.min_eps=50
 # Interval for agent status file updating (seconds) [0..86400]
--- a/src/Makefile
+++ b/src/Makefile
@@ -509,11 +509,10 @@ endif
 endif
 
 # zstd: request/response body compression. Manager-side decode is already
-# merged (#38129); agent-side compression (#38308) needs it too, so it moves
-# out of the manager-only set above. Built on every non-winagent target for
-# now -- winagent's MinGW cross-compile toolchain args are a dedicated
-# follow-up (verified against packages/windows/Dockerfile), not part of this
-# pass.
+# merged (#38129); agent-side compression needs it too, so it moves out of
+# the manager-only set above. Built on every non-winagent target for now --
+# winagent's MinGW cross-compile toolchain args are a dedicated follow-up
+# (verified against packages/windows/Dockerfile), not part of this pass.
 ifneq (${TARGET},winagent)
 	EXTERNAL_RES += zstd
 endif
--- a/src/Makefile
+++ b/src/Makefile
@@ -476,7 +476,7 @@ endif
 #     shared by the unit-test CI, whose deps step runs `make deps TARGET=...`
 #     WITHOUT TEST=1 and only enables tests at build time. Gating these behind a
 #     TEST flag drops them from that bundle and breaks every test build (#36247).
-EXTERNAL_RES := cJSON openssl zlib sqlite libyaml curl libpcre2 nlohmann flatbuffers bzip2 googletest benchmark
+EXTERNAL_RES := cJSON openssl zlib sqlite libyaml curl libpcre2 nlohmann flatbuffers bzip2 zstd googletest benchmark
 
 # Linux unit-test wrapper dependencies (every Linux leg except the winagent
 # cross-compile). unit_tests/wrappers/externals/{audit,procpc} include
@@ -506,15 +506,6 @@ ifneq (${TARGET},agent)
 ifneq (${TARGET},winagent)
 	EXTERNAL_RES += $(CPYTHON) libffi jemalloc rocksdb simdjson abseil-cpp spdlog yaml-cpp pugixml libmaxminddb protobuf date fmt re2 rapidjson taskflow RxCpp concurrentqueue fast_float geo_db tzdata cpp-httplib asio expected-lite llhttp restinio
 endif
-endif
-
-# zstd: request/response body compression. Manager-side decode is already
-# merged (#38129); agent-side compression needs it too, so it moves out of
-# the manager-only set above. Built on every non-winagent target for now --
-# winagent's MinGW cross-compile toolchain args are a dedicated follow-up
-# (verified against packages/windows/Dockerfile), not part of this pass.
-ifneq (${TARGET},winagent)
-	EXTERNAL_RES += zstd
 endif
 
 EXTERNAL_DIR := $(EXTERNAL_RES:%=external/%)
--- a/src/Makefile
+++ b/src/Makefile
@@ -504,8 +504,18 @@ endif
 # Manager-only dependencies.
 ifneq (${TARGET},agent)
 ifneq (${TARGET},winagent)
-	EXTERNAL_RES += $(CPYTHON) libffi jemalloc rocksdb simdjson abseil-cpp spdlog yaml-cpp pugixml libmaxminddb protobuf date fmt re2 rapidjson taskflow RxCpp concurrentqueue fast_float geo_db tzdata cpp-httplib asio expected-lite llhttp restinio zstd
+	EXTERNAL_RES += $(CPYTHON) libffi jemalloc rocksdb simdjson abseil-cpp spdlog yaml-cpp pugixml libmaxminddb protobuf date fmt re2 rapidjson taskflow RxCpp concurrentqueue fast_float geo_db tzdata cpp-httplib asio expected-lite llhttp restinio
 endif
+endif
+
+# zstd: request/response body compression. Manager-side decode is already
+# merged (#38129); agent-side compression (#38308) needs it too, so it moves
+# out of the manager-only set above. Built on every non-winagent target for
+# now -- winagent's MinGW cross-compile toolchain args are a dedicated
+# follow-up (verified against packages/windows/Dockerfile), not part of this
+# pass.
+ifneq (${TARGET},winagent)
+	EXTERNAL_RES += zstd
 endif
 
 EXTERNAL_DIR := $(EXTERNAL_RES:%=external/%)
--- a/src/client-agent/https_client/CMakeLists.txt
+++ b/src/client-agent/https_client/CMakeLists.txt
@@ -32,15 +32,9 @@ target_include_directories(
           ${SRC_FOLDER}/external/openssl/include)
 
 # wazuhext carries the vendored libcurl + OpenSSL symbols for agent targets.
-target_link_libraries(https_client PRIVATE wazuhext pthread)
-
-# ext_zstd: request-body compression, same target remoted_module already
-# links for the manager-side decode (#38129). Not built for winagent (see
-# src/Makefile / src/external/CMakeLists.txt) -- compression is compiled out
-# there (retrySender.cpp's #ifndef WIN32), so don't link it either.
-if(NOT CMAKE_SYSTEM_NAME STREQUAL "Windows")
-  target_link_libraries(https_client PRIVATE ext_zstd)
-endif()
+# ext_zstd: request-body compression, same target remoted_module
+# already links for the manager-side decode (#38129).
+target_link_libraries(https_client PRIVATE wazuhext pthread ext_zstd)
 
 set_target_properties(https_client PROPERTIES BUILD_RPATH_USE_ORIGIN TRUE
                                               BUILD_RPATH "$ORIGIN/../lib")
@@ -52,10 +46,20 @@ endif()
 # winagent: emit the .def at link time and generate the MinGW delay-load
 # import library the final win32 executables link against (the
 # agent_sync_protocol recipe).
+#
+# --exclude-libs=libzstd.a: MinGW auto-exports every global symbol from a
+# shared library by default (unlike ELF's convention), so without this,
+# zstd's own public C API (ZSTD_isError, etc.) leaks into this DLL's export
+# table/.def and gets its own delay-load thunk -- colliding with the real
+# ZSTD_isError the test binary also links directly via ext_zstd (needed
+# there to independently decompress and verify what this library produced).
+# Confirmed empirically: omitting this flag produces a real
+# "multiple definition of `ZSTD_isError'" link error building
+# https_client_unit_test.exe.
 if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
   set_target_properties(
     https_client PROPERTIES LINK_FLAGS
-                            "-Wl,--add-stdcall-alias -Wl,--output-def,libhttps_client.def")
+                            "-Wl,--add-stdcall-alias -Wl,--exclude-libs,libzstd.a -Wl,--output-def,libhttps_client.def")
 
   find_program(DLLTOOL i686-w64-mingw32-dlltool)
   if(NOT DLLTOOL)
--- a/src/client-agent/https_client/CMakeLists.txt
+++ b/src/client-agent/https_client/CMakeLists.txt
@@ -32,7 +32,7 @@ target_include_directories(
           ${SRC_FOLDER}/external/openssl/include)
 
 # wazuhext carries the vendored libcurl + OpenSSL symbols for agent targets.
-# ext_zstd: request-body compression (#38308), same target remoted_module
+# ext_zstd: request-body compression, same target remoted_module
 # already links for the manager-side decode (#38129).
 target_link_libraries(https_client PRIVATE wazuhext pthread ext_zstd)
 
--- a/src/client-agent/https_client/CMakeLists.txt
+++ b/src/client-agent/https_client/CMakeLists.txt
@@ -32,9 +32,15 @@ target_include_directories(
           ${SRC_FOLDER}/external/openssl/include)
 
 # wazuhext carries the vendored libcurl + OpenSSL symbols for agent targets.
-# ext_zstd: request-body compression, same target remoted_module
-# already links for the manager-side decode (#38129).
-target_link_libraries(https_client PRIVATE wazuhext pthread ext_zstd)
+target_link_libraries(https_client PRIVATE wazuhext pthread)
+
+# ext_zstd: request-body compression, same target remoted_module already
+# links for the manager-side decode (#38129). Not built for winagent (see
+# src/Makefile / src/external/CMakeLists.txt) -- compression is compiled out
+# there (retrySender.cpp's #ifndef WIN32), so don't link it either.
+if(NOT CMAKE_SYSTEM_NAME STREQUAL "Windows")
+  target_link_libraries(https_client PRIVATE ext_zstd)
+endif()
 
 set_target_properties(https_client PROPERTIES BUILD_RPATH_USE_ORIGIN TRUE
                                               BUILD_RPATH "$ORIGIN/../lib")
--- a/src/client-agent/https_client/CMakeLists.txt
+++ b/src/client-agent/https_client/CMakeLists.txt
@@ -32,7 +32,9 @@ target_include_directories(
           ${SRC_FOLDER}/external/openssl/include)
 
 # wazuhext carries the vendored libcurl + OpenSSL symbols for agent targets.
-target_link_libraries(https_client PRIVATE wazuhext pthread)
+# ext_zstd: request-body compression (#38308), same target remoted_module
+# already links for the manager-side decode (#38129).
+target_link_libraries(https_client PRIVATE wazuhext pthread ext_zstd)
 
 set_target_properties(https_client PROPERTIES BUILD_RPATH_USE_ORIGIN TRUE
                                               BUILD_RPATH "$ORIGIN/../lib")
--- a/src/client-agent/https_client/include/https_client.h
+++ b/src/client-agent/https_client/include/https_client.h
@@ -172,7 +172,7 @@ typedef struct hc_config_t
     bool config_report_enabled;
     uint32_t config_report_interval_s; ///< 0 -> 3600.
 
-    /// zstd-compress in-memory request bodies before signing/sending (#38308).
+    /// zstd-compress in-memory request bodies before signing/sending.
     /// internal_options.conf (agent.https_compression_enabled), not a <client>
     /// XML setting -- OFF by default.
     bool https_compression_enabled;
--- a/src/client-agent/https_client/include/https_client.h
+++ b/src/client-agent/https_client/include/https_client.h
@@ -171,6 +171,11 @@ typedef struct hc_config_t
     uint32_t stats_interval_s;         ///< 0 -> 60.
     bool config_report_enabled;
     uint32_t config_report_interval_s; ///< 0 -> 3600.
+
+    /// zstd-compress in-memory request bodies before signing/sending (#38308).
+    /// internal_options.conf (agent.https_compression_enabled), not a <client>
+    /// XML setting -- OFF by default.
+    bool https_compression_enabled;
 } hc_config_t;
 
 /**
--- a/src/client-agent/https_client/src/compressionGate.hpp
+++ b/src/client-agent/https_client/src/compressionGate.hpp
@@ -15,7 +15,7 @@
 #include <atomic>
 
 /**
- * @brief The zstd-rejected latch (#38308). Whether the manager accepts
+ * @brief The zstd-rejected latch. Whether the manager accepts
  *        Content-Encoding: zstd is a property of the manager/connection, not
  *        of which endpoint asked -- so every RetrySender on this agent shares
  *        one gate: the first 415 disables compression for all six send
--- a/src/client-agent/https_client/src/compressionGate.hpp
+++ b/src/client-agent/https_client/src/compressionGate.hpp
@@ -1,0 +1,45 @@
+/*
+ * Wazuh agent HTTPS client (C++ transport module)
+ * Copyright (C) 2015, Wazuh Inc.
+ * August 12, 2026.
+ *
+ * This program is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU General Public
+ * License (version 2) as published by the FSF - Free Software
+ * Foundation.
+ */
+
+#ifndef _HC_COMPRESSION_GATE_HPP
+#define _HC_COMPRESSION_GATE_HPP
+
+#include <atomic>
+
+/**
+ * @brief The zstd-rejected latch (#38308). Whether the manager accepts
+ *        Content-Encoding: zstd is a property of the manager/connection, not
+ *        of which endpoint asked -- so every RetrySender on this agent shares
+ *        one gate: the first 415 disables compression for all six send
+ *        paths, for the rest of this agent's run.
+ *
+ * One-way (unlike AuthGate): the manager's lack of zstd support does not
+ * change without a restart, so there is no release()/re-arm.
+ */
+class CompressionGate final
+{
+    public:
+        /// A compressed attempt got a 415. Idempotent; safe from any thread.
+        void reportRejected()
+        {
+            m_disabled.store(true, std::memory_order_relaxed);
+        }
+
+        bool disabled() const
+        {
+            return m_disabled.load(std::memory_order_relaxed);
+        }
+
+    private:
+        std::atomic<bool> m_disabled {false};
+};
+
+#endif // _HC_COMPRESSION_GATE_HPP
--- a/src/client-agent/https_client/src/configFetcher.cpp
+++ b/src/client-agent/https_client/src/configFetcher.cpp
@@ -38,10 +38,11 @@ namespace
 
 ConfigFetcher::ConfigFetcher(const ModuleConfig& config, IHttpPerformer& performer,
                              const ISigner& signer, IClock& clock, IRandom& random,
-                             ISpoolFileFactory& spoolFactory, AuthGate& authGate)
+                             ISpoolFileFactory& spoolFactory, AuthGate& authGate,
+                             CompressionGate& compressionGate)
     : m_config(config)
     , m_backoff(config.backoffBaseMs, config.backoffCapMs, random)
-    , m_sender(performer, signer, clock, m_backoff, config.httpsCompressionEnabled, &authGate)
+    , m_sender(performer, signer, clock, m_backoff, config.httpsCompressionEnabled, &compressionGate, &authGate)
     , m_spoolFactory(spoolFactory)
 {
 }
--- a/src/client-agent/https_client/src/configFetcher.cpp
+++ b/src/client-agent/https_client/src/configFetcher.cpp
@@ -41,7 +41,7 @@ ConfigFetcher::ConfigFetcher(const ModuleConfig& config, IHttpPerformer& perform
                              ISpoolFileFactory& spoolFactory, AuthGate& authGate)
     : m_config(config)
     , m_backoff(config.backoffBaseMs, config.backoffCapMs, random)
-    , m_sender(performer, signer, clock, m_backoff, &authGate)
+    , m_sender(performer, signer, clock, m_backoff, config.httpsCompressionEnabled, &authGate)
     , m_spoolFactory(spoolFactory)
 {
 }
--- a/src/client-agent/https_client/src/configFetcher.hpp
+++ b/src/client-agent/https_client/src/configFetcher.hpp
@@ -40,7 +40,8 @@ class ConfigFetcher final
     public:
         ConfigFetcher(const ModuleConfig& config, IHttpPerformer& performer,
                       const ISigner& signer, IClock& clock, IRandom& random,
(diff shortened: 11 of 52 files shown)


lchico:
## Live verification evidence — `/stateful` compression + deb/rpm cross-check (build `4feb3bf`)

Tested against two live agent-manager pairs on the same build (`4feb3bf`, this branch): a deb agent on Ubuntu 22.04 and an rpm agent on AlmaLinux 9. For each, a background `tcpdump` capture was decrypted with `tshark` via `SSLKEYLOGFILE` (wired through a systemd drop-in on the real `wazuh-agent` process) to read the literal HTTP headers on the wire — not an inferred byte-count difference.

### `/stateless` (control, both packages)

| Package | OFF (bytes) | ON (bytes) | Reduction | Decrypted `Content-Encoding` |
|---|---|---|---|---|
| deb (Ubuntu 22.04) | 134368 | 23889 | 82.2% | absent OFF → `zstd` ON (confirmed) |
| rpm (AlmaLinux 9) | 134756 | 23776 | 82.4% | not decryptable in this environment (see note below) |

### `/stateful` (FIM/SCA sync) — the new evidence for this build

`/stateful` carries FIM/SCA sync data over a file-backed spool file, previously flagged as design-only because the signer and the HTTP performer independently re-open the body file. This build closes that gap:

- `StatefulStream::sendSession()` (`src/client-agent/https_client/src/statefulStream.cpp`) pre-compresses the spool file once, up front, into a sibling file and hands its path to `RetrySender` via `HttpRequestSpec::precompressedBodyFilePath`.
- `RetrySender::attemptOnce()` (`src/client-agent/https_client/src/retrySender.cpp`) swaps that path into `bodyFilePath` **before** `m_signer.signFile()` and the performer's own `fopen()`, so both reads see the same already-compressed bytes.

To capture a live `/stateful` POST inside a practical test window, the FIM sync interval was temporarily shrunk to 15s and a real FIM delta was forced via a scratch, realtime-monitored directory.

| Package | OFF (bytes, context) | ON (bytes, context) | Decrypted `Content-Encoding` | Response codes | Verdict |
|---|---|---|---|---|---|
| deb (Ubuntu 22.04) | 79424 | 55998 | absent OFF (`POST\|/stateful\|\|1736`) → `zstd` ON (`POST\|/stateful\|zstd\|906`) | `200,202` both phases, no 401s | **PASS** |
| rpm (AlmaLinux 9) | 66100 | 57607 | not decryptable in this environment (see note below) | not decryptable | inconclusive on header proof (see note) |

`ossec.log` on the rpm agent confirmed FIM sync completing successfully on schedule throughout both capture windows, and connectivity/auth stayed healthy (service up, `https_client` active, no repeated comms failures) after enabling compression on both packages — i.e. signing order over the compressed spool file validates end-to-end.

### Note on the rpm/AlmaLinux agent's decryption gap

The rpm container's `tshark` (3.4.10, the newest available from AlmaLinux 9's own repos — no EPEL/newer build configured) could not decrypt this specific TLS 1.3 traffic pattern, even with a correctly-populated `SSLKEYLOGFILE` (a standalone sanity decrypt of a separate, simpler HTTPS exchange on the same agent/tshark build worked fine, isolating this to a tshark-version/TLS-record-reassembly limitation, not a compression or signing issue). The byte-count reduction on that agent tracks the deb agent's almost exactly (82.4% vs 82.2% on `/stateless`), and the compressed code path (`RetrySender`/`StatefulStream`) is identical source shared by both packages, which is the fallback evidence for that agent.

cborla (approved):
LGTM
- [x] Code changes reviewed

PR #38361, opened 2026-08-13
Compile std::regex once instead of on every call in syscollector normalizer and data_provider parsers

### LINKED ISSUE #38171: Compile std::regex once instead of on every call in syscollector normalizer and data_provider parsers
## Description

Several agent-side code paths construct a `std::regex` every time a function runs, from patterns that are either compile-time literals or configuration data that is immutable after construction. Compiling a regex is far more expensive than matching with it (pattern parsing plus NFA construction and allocations), so these are pure waste on paths that run once per collected inventory item.

Audited `data_provider`, `wazuh_modules/{syscollector,sca,agent_info}`, `shared_modules`, `syscheckd` and `logcollector` for non-`static` `std::regex` construction. The sites below are the ones where the pattern provably does not depend on per-call state.

Each item states which platforms actually reach it, since the syscollector normalizer only loads rules matching the platform's normalizer target.

### 1. `syscollectorNormalizer.cpp:81` and `:102` — recompiled per dictionary entry, per item

```cpp
// src/wazuh_modules/syscollector/src/syscollectorNormalizer.cpp:71-104
static void normalizeItem(const nlohmann::json& dictionary, nlohmann::json& item)
{
    for (const auto& dictItem : dictionary)
    {
        ...
        std::regex pattern{itFindPattern->get_ref<const std::string&>()};   // :81
        ...
        std::regex pattern{itReplacePattern->get_ref<const std::string&>()}; // :102
```

`normalizeItem()` is called for every element of the normalized array, and the inner loop walks the whole normalization dictionary, so the cost is `items x dictionary_entries` regex compilations per scan. The dictionary comes from `m_typeDictionary`, built once in the constructor (`syscollectorNormalizer.cpp:19-21`) from the normalization config file and never mutated afterwards, so `find_pattern` and `replace_pattern` can be compiled once at construction time and reused: `dictionary_entries` compilations per agent start instead.

**Platform scope, measured before filing:** `normalize()` is only invoked for `packages` (`syscollectorImp.cpp:1573`), and only entries whose `target` matches `SYSCOLLECTOR_NORM_TYPE` (`shared/include/defs.h:187-193`) are loaded. All 17 entries of the shipped `norm_config.json` are `"target": "macos"`, 15 of them in `dictionary`. So today this is a **macOS-only** gain: on Linux and Windows `m_typeDictionary` is empty and `normalize()` returns at the `find(type)` miss. On macOS the shipped dictionary has 15 entries, so a host with 500 packages pays 15 x 500 = 7500 regex compilations per syscollector scan, which the fix reduces to 15 at startup.

### 2. `syscollectorNormalizer.cpp:35` — same problem in `removeExcluded()`

Same fix, same member-caching change. The iterator-invalidation bug in that function is filed separately in #38170; only the regex caching belongs here, but both touch the same function so they should be coordinated.

### 3. `data_provider/src/osinfo/sysOsParsers.cpp` — regex from `constexpr` literals

```cpp
// :63-69
constexpr auto MAJOR_VERSION_PATTERN{"^([0-9]+)\\.*"};
std::regex pattern{MAJOR_VERSION_PATTERN};
```

Sites: `:69` (plus the two reassignments at `:76` and `:83` in the same function), `:170`, `:213`, `:379`, `:416`. All of them build the pattern from a `constexpr auto` literal declared in the same function, so `static const std::regex` is a mechanical change with no behavioral difference.

Excluded from this list: `:101` (`std::regex pattern{regex}` in `findVersionInStream()`), where the pattern is a function parameter, so it cannot be made static.

### 4. `data_provider/src/utilsWrapperWin.cpp:269`

```cpp
std::regex hotfixRegex("KB[0-9]+");
```

Literal pattern in `QueryWUHotFixes()`, used inside the loop over the WUA update history. `static const`.

### 5. `data_provider/src/packages/packageLinuxParserDeb.cpp:59-61`

```cpp
std::regex listPattern(R"(^python.*\.list$)");
const auto PYTHON_INFO_FILES = std::array {std::make_pair(std::regex(R"(^.*\.egg-info$)"), "/PKG-INFO"),
                                           std::make_pair(std::regex(R"(^.*\.dist-info$)"), "/METADATA")};
```

Three literal-pattern regexes rebuilt on every `getDpkgPythonPackages()` call, then used inside the loop over `DPKG_INFO_PATH`. `static const` for all three.

## Checked but excluded

- `data_provider/src/sysInfoLinux.cpp:188` — `const std::regex cpuDirectoryRegex` is already hoisted out of the loop it feeds. Making it `static const` is possible but the enclosing function runs once per scan, so the gain is negligible.
- Patterns coming from function parameters or from per-call data (`sysOsParsers.cpp:101`): not applicable.

## Proposed fix

- For 1 and 2: compile the dictionary/exclusion patterns once in `SysNormalizer`'s constructor and store them alongside (or instead of) the raw pattern strings, then match against the cached objects.
- For 3, 4 and 5: mark the regex objects `static const`. Function-local statics are thread-safe initialized since C++11, and `std::regex_match`/`std::regex_search` do not mutate the regex, so concurrent use is safe.

## Ask

Confirm the approach for the normalizer cache (cached vector of compiled patterns held as a member vs. a lazily populated cache keyed by pattern string) before implementation starts.

Miguevrgo:
## Description

`SysNormalizer` and several `data_provider` OS/package parsers construct a `std::regex` from a compile-time-constant pattern on every call, even though regex compilation (pattern parsing plus NFA construction and allocations) is far more expensive than matching with it.

The clearest case is `wazuh_modules/syscollector/src/syscollectorNormalizer.cpp`:

```cpp
// normalizeItem(), called once per normalized item, for every dictionary entry
std::regex pattern{itFindPattern->get_ref<const std::string&>()};   // :81 (old)
std::regex pattern{itReplacePattern->get_ref<const std::string&>()}; // :102 (old)
```

`normalizeItem()` runs for every element of the normalized array, and the inner loop walks the whole `m_typeDictionary` entry set for that type — built once in the constructor from `norm_config.json` and never mutated afterwards. So the pattern strings never change between calls, but the regex was rebuilt from scratch on every (item × dictionary entry) pair.

`normalize()` is only invoked for `packages` (`syscollectorImp.cpp:1573`), and only dictionary entries whose `target` matches the platform (`SYSCOLLECTOR_NORM_TYPE`) are loaded — the shipped `norm_config.json` has 15 `packages` entries targeting `macos`, so on Linux/Windows `m_typeDictionary` is empty and this is a no-op; on macOS a host with 500 packages paid 15 × 500 = 7500 regex compilations per scan.

The same pattern (regex built from a `constexpr`/literal pattern inside the function that uses it) appears in:
- `data_provider/src/osinfo/sysOsParsers.cpp` — `findMajorMinorVersionInString()` and the Ubuntu/BSD/MacOS parsers
- `data_provider/src/utilsWrapperWin.cpp:269` — `QueryWUHotFixes()`'s `hotfixRegex`, inside the WUA update-history loop
- `data_provider/src/packages/packageLinuxParserDeb.cpp:59-61` — `getDpkgPythonPackages()`'s `listPattern` and `PYTHON_INFO_FILES`, inside the loop over `DPKG_INFO_PATH`

Closes #38171

## Proposed Changes

- `SysNormalizer`: `compileExclusions()`/`compileDictionary()` now compile the dictionary/exclusion regex once in the constructor and store them in `ExclusionRule`/`DictionaryRule` structs (`syscollectorNormalizer.hpp`). `removeExcluded()`/`normalizeItem()` only match against the cached `std::regex` objects — no construction left in the per-scan hot path.
- `sysOsParsers.cpp`, `utilsWrapperWin.cpp`, `packageLinuxParserDeb.cpp`: regex objects built from literal/`constexpr` patterns are now `static const` function-local statics (thread-safe since C++11; `regex_match`/`regex_search`/`regex_replace` don't mutate the regex, so concurrent use is safe).
- `packageLinuxDataRetriever.h`: aligned `getDpkgInfo()`'s declaration with its `const std::function<...>&` callback parameter (was a value-vs-reference signature mismatch causing a link failure once the definition took the callback by reference).

### Results and Evidence

Instrumented the pre-fix `syscollectorNormalizer.cpp` with an atomic regex-construction counter and ran `SysNormalizer::normalize("packages", ...)` once over 500 synthetic package items against the shipped `norm_config.json` (15 macOS `packages` dictionary entries):

```
compiles_before_normalize=0
compiles_after_one_scan=7500
```

`7500 = 15 × 500`, matching the math in the issue exactly.

After the fix, `std::regex{...}` construction only occurs inside `compileExclusions()`/`compileDictionary()` (called once, from the constructor):

```
$ grep -n "std::regex" src/wazuh_modules/syscollector/src/syscollectorNormalizer.cpp
41:   if (fieldIt != item->end() && std::regex_match(fieldIt->get_ref<const std::string&>(), exclusionItem.pattern))
51:   if (fieldIt != data.end() && std::regex_match(fieldIt->get_ref<const std::string&>(), exclusionItem.pattern))
77:   !std::regex_match(fieldIt->get_ref<const std::string&>(), dictItem.find->pattern))
90:   *fieldIt = std::regex_replace(fieldIt->get_ref<const std::string&>(), dictItem.replace->pattern, dictItem.replace->value);
168:  ret[type].push_back({std::regex{exclusionItem["pattern"].get_ref<const std::string&>()},
197:  rule.find = FindRule{std::regex{itFindPattern->get_ref<const std::string&>()},
212:  rule.replace = ReplaceRule{std::regex{itReplacePattern->get_ref<const std::string&>()},
```

Lines 168/197/212 are inside `compileExclusions`/`compileDictionary`; lines 41/51/77/90 (the per-scan hot path in `removeExcluded`/`normalizeItem`) only call `regex_match`/`regex_replace` against the already-built `.pattern` members. Per-scan cost drops from 7500 compiles to 0 (15 compiles happen once, at construction).

Unit tests, run before (stashed pre-fix source) and after the fix, with identical results:

```
$ ./bin/sys_normalizer_unit_test
[==========] 16 tests from 1 test suite ran. (2 ms total)
[  PASSED  ] 16 tests.

$ ./bin/sysinfo_unit_test --gtest_filter=*OsParser*:*Ubuntu*:*Centos*:*BSD*:*RedHat*:*Debian*:*Arch*:*Slackware*:*Gentoo*:*SuSE*:*Fedora*:*Alpine*:*MacOs*
[==========] 23 tests from 1 test suite ran. (0 ms total)
[  PASSED  ] 23 tests.
```

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

### Artifacts Affected

N/A

### Configuration Changes

N/A

### Tests Introduced

N/A — existing `sys_normalizer_unit_test` and `sysinfo_unit_test` suites already cover the affected code paths and pass unchanged.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

### DIFF
--- a/src/wazuh_modules/syscollector/src/syscollectorNormalizer.cpp
+++ b/src/wazuh_modules/syscollector/src/syscollectorNormalizer.cpp
@@ -158,7 +158,7 @@ std::map<std::string, nlohmann::json> SysNormalizer::getTypeValues(const std::st
 }
 
 std::map<std::string, std::vector<SysNormalizer::ExclusionRule>> SysNormalizer::compileExclusions(
-    const std::map<std::string, nlohmann::json>& rawExclusions)
+                                                                  const std::map<std::string, nlohmann::json>& rawExclusions)
 {
     std::map<std::string, std::vector<ExclusionRule>> ret;
 
@@ -183,7 +183,7 @@ std::map<std::string, std::vector<SysNormalizer::ExclusionRule>> SysNormalizer::
 }
 
 std::map<std::string, std::vector<SysNormalizer::DictionaryRule>> SysNormalizer::compileDictionary(
-    const std::map<std::string, nlohmann::json>& rawDictionary)
+                                                                   const std::map<std::string, nlohmann::json>& rawDictionary)
 {
     std::map<std::string, std::vector<DictionaryRule>> ret;
 
--- a/src/wazuh_modules/syscollector/src/syscollectorNormalizer.cpp
+++ b/src/wazuh_modules/syscollector/src/syscollectorNormalizer.cpp
@@ -16,8 +16,8 @@
 
 SysNormalizer::SysNormalizer(const std::string& configFile,
                              const std::string& target)
-    : m_typeExclusions{getTypeValues(configFile, target, "exclusions")}
-    , m_typeDictionary{getTypeValues(configFile, target, "dictionary")}
+    : m_typeExclusions{compileExclusions(getTypeValues(configFile, target, "exclusions"))}
+    , m_typeDictionary{compileDictionary(getTypeValues(configFile, target, "dictionary"))}
 {
 }
 
@@ -32,18 +32,15 @@ void SysNormalizer::removeExcluded(const std::string& type,
         {
             try
             {
-                std::regex pattern{exclusionItem["pattern"].get_ref<const std::string&>()};
-                const auto& fieldName{exclusionItem["field_name"].get_ref<const std::string&>()};
-
                 if (data.is_array())
                 {
                     auto& arr{data.get_ref<nlohmann::json::array_t&>()};
 
                     for (size_t i{arr.size()}; i-- > 0;)
                     {
-                        const auto fieldIt{arr[i].find(fieldName)};
+                        const auto fieldIt{arr[i].find(exclusionItem.fieldName)};
 
-                        if (fieldIt != arr[i].end() && std::regex_match(fieldIt->get_ref<const std::string&>(), pattern))
+                        if (fieldIt != arr[i].end() && std::regex_match(fieldIt->get_ref<const std::string&>(), exclusionItem.pattern))
                         {
                             std::swap(arr[i], arr.back());
                             arr.pop_back();
@@ -52,9 +49,9 @@ void SysNormalizer::removeExcluded(const std::string& type,
                 }
                 else
                 {
-                    const auto fieldIt{data.find(fieldName)};
+                    const auto fieldIt{data.find(exclusionItem.fieldName)};
 
-                    if (fieldIt != data.end() && std::regex_match(fieldIt->get_ref<const std::string&>(), pattern))
+                    if (fieldIt != data.end() && std::regex_match(fieldIt->get_ref<const std::string&>(), exclusionItem.pattern))
                     {
                         data.clear();
                     }
@@ -70,53 +67,36 @@ void SysNormalizer::removeExcluded(const std::string& type,
 }
 
 
-static void normalizeItem(const nlohmann::json& dictionary,
+static void normalizeItem(const std::vector<SysNormalizer::DictionaryRule>& dictionary,
                           nlohmann::json& item)
 {
     for (const auto& dictItem : dictionary)
     {
-        const auto itFindPattern{dictItem.find("find_pattern")};
-        const auto itFindField{dictItem.find("find_field")};
-
-        if (itFindPattern != dictItem.end() && itFindField != dictItem.end())
+        if (dictItem.find)
         {
-            const auto fieldIt{item.find(itFindField->get_ref<const std::string&>())};
-            std::regex pattern{itFindPattern->get_ref<const std::string&>()};
+            const auto fieldIt{item.find(dictItem.find->fieldName)};
 
             if (fieldIt == item.end() ||
-                    !std::regex_match(fieldIt->get_ref<const std::string&>(), pattern))
+                    !std::regex_match(fieldIt->get_ref<const std::string&>(), dictItem.find->pattern))
             {
                 //no field in the item or no matching, we continue
                 continue;
             }
         }
-        else if (itFindPattern != dictItem.end() || itFindField != dictItem.end())
-        {
-            //we won't evaluate an incomplete item.
-            continue;
-        }
 
-        const auto itReplacePattern{dictItem.find("replace_pattern")};
-        const auto itReplaceField{dictItem.find("replace_field")};
-        const auto itReplaceValue{dictItem.find("replace_value")};
-
-        if (itReplacePattern != dictItem.end() && itReplaceField != dictItem.end() && itReplaceValue != dictItem.end())
+        if (dictItem.replace)
         {
-            std::regex pattern{itReplacePattern->get_ref<const std::string&>()};
-            const auto fieldIt{item.find(itReplaceField->get_ref<const std::string&>())};
+            const auto fieldIt{item.find(dictItem.replace->fieldName)};
 
             if (fieldIt != item.end())
             {
-                *fieldIt = std::regex_replace(fieldIt->get_ref<const std::string&>(), pattern, itReplaceValue->get_ref<const std::string&>());
+                *fieldIt = std::regex_replace(fieldIt->get_ref<const std::string&>(), dictItem.replace->pattern, dictItem.replace->value);
             }
         }
 
-        const auto itAddField{dictItem.find("add_field")};
-        const auto itAddValue{dictItem.find("add_value")};
-
-        if (itAddField != dictItem.end() && itAddValue != dictItem.end())
+        if (dictItem.add)
         {
-            item[itAddField->get_ref<const std::string&>()] = itAddValue->get_ref<const std::string&>();
+            item[dictItem.add->fieldName] = dictItem.add->value;
         }
     }
 }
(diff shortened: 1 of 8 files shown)


Darioortegaleyva (changes requested):
**Blocking**: `compileDictionary()` (`syscollectorNormalizer.cpp:182`) has no try/catch, unlike `compileExclusions()` (line 157). Both now compile regex eagerly in the constructor. A single bad regex in the dictionary config used to break only that one collector at scan time; now it throws in `Syscollector::init()` and kills the whole module. Add the same `try/catch(...) {}` as `compileExclusions`.

**Also missing**: the fix doesn't cover every regex-per-call spot it should. `findVersionInStream()` (`sysOsParsers.cpp:105`) still recompiles on every call, same file. `packageLinuxParserRpm.cpp:129` and `shadow_linux.cpp:32` have the identical untouched bug.

I think the best is to fix the try/catch, that's the real blocker.

Darioortegaleyva (approved):
LGTM. compileDictionary() now has the same try/catch as compileExclusions(), blocker resolved.

Non-blocking, not covered by this PR or #38171's audit, no other open PR/issue tracks them either: packageLinuxParserRpm.cpp:129 and shadow_linux.cpp:32 have the same per-call regex pattern this issue fixed elsewhere. Worth a quick follow-up if you want to close the gap.

Darioortegaleyva (changes requested):
LGTM. compileDictionary() now has the same try/catch as compileExclusions(), blocker resolved.

Please add a CHANGELOG.md entry, the issue asks for one, and then I approve.

Non-blocking, not covered by this PR or #38171's audit, no other open PR/issue tracks them either: packageLinuxParserRpm.cpp:129 and shadow_linux.cpp:32 have the same per-call regex pattern this issue fixed elsewhere. Worth a quick follow-up if you want to close the gap.

Darioortegaleyva (dismissed):
LGTM

src/wazuh_modules/syscollector/src/syscollectorNormalizer.cpp
```diff
@@ -32,18 +32,15 @@ void SysNormalizer::removeExcluded(const std::string& type,
...
                     for (size_t i{arr.size()}; i-- > 0;)
                     {
-                        const auto fieldIt{arr[i].find(fieldName)};
+                        const auto fieldIt{arr[i].find(exclusionItem.fieldName)};
 
-                        if (fieldIt != arr[i].end() && std::regex_match(fieldIt->get_ref<const std::string&>(), pattern))
+                        if (fieldIt != arr[i].end() && std::regex_match(fieldIt->get_ref<const std::string&>(), exclusionItem.pattern))
                         {
                             std::swap(arr[i], arr.back());
                             arr.pop_back();
```
MarcelKemp: **[S3 · resolved · pre-existing]** Iterator invalidation bug is fixed. The reverse-index `for (size_t i{arr.size()}; i-- > 0;)` loop with `std::swap(arr[i], arr.back()); arr.pop_back();` replaces the previous iterator-based `data.erase(item)`, which was invalid. This eliminates undefined behavior when consecutive elements match exclusion criteria.

src/wazuh_modules/syscollector/src/syscollectorNormalizer.cpp
```diff
@@ -176,3 +156,87 @@ std::map<std::string, nlohmann::json> SysNormalizer::getTypeValues(const std::st
...
+std::map<std::string, std::vector<SysNormalizer::ExclusionRule>> SysNormalizer::compileExclusions(
+                                                                  const std::map<std::string, nlohmann::json>& rawExclusions)
+{
+    std::map<std::string, std::vector<ExclusionRule>> ret;
+
+    for (const auto& [type, items] : rawExclusions)
+    {
+        for (const auto& exclusionItem : items)
+        {
+            try
```
MarcelKemp: **[S2 · not blocking · aggravated · persists]** `compileExclusions()` still uses `operator[]` to access "pattern" and "field_name" without bounds check. This is aggravated by moving validation to constructor startup (now runs for all `data_type` entries, not just ones in an active scan). Use `find()` and skip the entry if any key is missing, like `compileDictionary()` already does.

src/wazuh_modules/syscollector/src/syscollectorNormalizer.cpp
```diff
@@ -176,3 +156,87 @@ std::map<std::string, nlohmann::json> SysNormalizer::getTypeValues(const std::st
...
+                const auto itAddValue{dictItem.find("add_value")};
+
+                if (itAddField != dictItem.end() && itAddValue != dictItem.end())
+                {
+                    rule.add = AddRule{itAddField->get_ref<const std::string&>(),
+                                       itAddValue->get_ref<const std::string&>()};
+                }
+
+                ret[type].push_back(std::move(rule));
+            }
```
MarcelKemp: **[S3 · not blocking · introduced · persists]** `catch(...)` at line 175 silently swallows rule compilation errors with no logging. A malformed pattern in `norm_config.json` disables normalization with zero diagnostics. Narrow the `try` to each field extraction so valid rule parts aren't lost when one field fails.

src/wazuh_modules/syscollector/src/syscollectorNormalizer.cpp
```diff
@@ -176,3 +156,87 @@ std::map<std::string, nlohmann::json> SysNormalizer::getTypeValues(const std::st
 
     return ret;
 }
```
MarcelKemp: **[S3 · not blocking · introduced · persists]** No test coverage for the new constructor validation paths. Add two `EXPECT_NO_THROW` tests: one with an unparseable regex pattern, one with a missing "pattern" key, and verify the remaining valid rules still apply.

MarcelKemp (approved):

PR #37758, opened 2026-07-18
Free rpm macro context to stop unbounded growth

Nicogp:
## Description

On Linux agents, `wazuh-modulesd` RSS grows linearly and without bound over the lifetime of the
process under continuous system inventory churn (packages/users/groups/processes changing), as
reported by the `5.0.0/Performance_Testing` soak test (Jenkins #235): ~1.9–2.9 MB/h over ~25 hours,
resetting only on daemon restart, Linux-only (macOS/Windows unaffected), all other daemons flat.

Root cause: `RpmPackageManager`'s constructor (`src/data_provider/src/packages/rpmPackageManager.cpp`)
calls `rpmReadConfigFiles()`, which internally calls rpm's own `rpmInitMacros()` — this pushes every
built-in macro plus every macro from rpm's config files onto **rpm's process-global macro context**,
every single `scanPackages()` cycle, with no clearing first (confirmed by reading the vendored
`rpm/rpmio/macro.c`). The only function that empties that table is `rpmFreeMacros()`. The destructor
called `rpmFreeRpmrc()` only — a different, unrelated subsystem (platform/arch compatibility tables,
`rpm/lib/rpmrc.c`) — never `rpmFreeMacros()`. rpm's own reference CLI (`rpm/lib/poptALL.c`) calls both
together on exit; this code was missing the macro-table cleanup. Since `RpmPackageManager` is
constructed and destroyed fresh on every scan cycle, this permanently grew rpm's global macro table,
unbounded, for the lifetime of the process — a genuine leak in third-party library global state,
invisible to any allocator-level mitigation (arena tuning, `malloc_trim`, tcache) because the memory
was never actually free.

This was root-caused via `massif` live-heap profiling (not RSS/allocator-level guessing) under a real
IT-Hygiene churn reproduction that matches the reported production rate (~250–272 KB/scan).

Closes #37656

## Proposed Changes

- **`src/data_provider/src/packages/rpmPackageManager.cpp`**: `RpmPackageManager::~RpmPackageManager()`
  now calls `m_rpmlib->rpmFreeMacros()` before the existing `m_rpmlib->rpmFreeRpmrc()`, so every
  scan cycle's macro-context growth is released when the object is destroyed.
- **`src/data_provider/src/packages/rpmlibWrapper.h`**: added `virtual void rpmFreeMacros() = 0;` to
  `IRpmLibWrapper`, plus the `<rpm/rpmmacro.h>` include needed for the declaration.
[...]

PR #38366, opened 2026-08-13
Backport: Free rpm macro context to stop unbounded growth

### LINKED ISSUE #38364: `wazuh-modulesd` unbounded memory growth on Linux agents from resource leaks in the inventory path
|---|---|---|---|---|

## Description

## Findings

|---|---|---|---|

### Finding 1:

### Finding 2:

### Finding 3:

## Acceptance criteria

anromerom:
## Description

Related to #38364.

- Backport of #37758 to the 4.14 line.
The fix was merged into `5.0.0` on 2026-07-20 and never backported, so it is present in every 4.14.x release published so far.

This PR addresses only the first of the three findings tracked in #38364, which is why it does not close it. The remaining two are handled in their own PRs and the issue stays open until all three are merged.

## Proposed Changes

Cherry-pick of the source and test changes from #37758, without the changelog entry:

- `rpmPackageManager.cpp`: Release rpm's macro table in the destructor, alongside the existing `rpmFreeRpmrc()`.
- `rpmlibWrapper.h`: Declare `rpmFreeMacros()` on `IRpmLibWrapper`, plus the `<rpm/rpmmacro.h>` include it needs.
- `rpmlib.h`: Implement it as a thin wrapper over `::rpmFreeMacros(nullptr)`.
- `RpmPackageManager_test.cpp` and `sysInfoPackageLinuxParserRPM_test.cpp`: Add the mock method and assert the destructor calls it.

The patch applied to `4.14.9` with no adaptation, so the change is identical to the one on `5.0.0`, comments included.

### Results and Evidence

Measured on the shipped packages, both built from the same base commit through the same pipeline, on
`rockylinux:9` (rpm 4.16.1.3, sqlite backend), driving `sysinfo_packages()` from the installed `libsysinfo.so`,
300 s per run.

| Package | Scans | Growth | Per scan |
|---|---:|---:|---:|
| Baseline, `4.14.9` at `7bacb80` | 1,022 | 70,312 KB | **68.80 KB** |
| This fix, at `9dec562` | 1,029 | -60 KB | **-0.06 KB** |

The growth is gone rather than reduced: the patched package ends below where it started. The baseline figure
matches earlier independent measurements within 0.6% (68.60 KB per scan on the published 4.14.1 package, 69.01 KB
over a 53 minute soak).

### Artifacts Affected

- `libsysinfo.so`. No change to inventory data or scan semantics.

### Configuration Changes

None.

### Documentation Updates

None

### Tests Introduced

- Extended the existing `RpmLibTest.RAII` test to assert `rpmFreeMacros()` is called on destruction, and added the mock method in both RPM test suites. No new test files.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/src/data_provider/src/packages/rpmPackageManager.cpp
+++ b/src/data_provider/src/packages/rpmPackageManager.cpp
@@ -40,6 +40,11 @@ RpmPackageManager::RpmPackageManager(std::shared_ptr<IRpmLibWrapper>&& wrapper)
 
 RpmPackageManager::~RpmPackageManager()
 {
+    // the constructor's rpmReadConfigFiles() call loads macros into rpm's
+    // process-global macro context (rpmInitMacros()), which is never cleared automatically.
+    // rpmFreeRpmrc() alone does not free it (different subsystem) — without this, every
+    // construct/destruct cycle of RpmPackageManager permanently grows that global table.
+    m_rpmlib->rpmFreeMacros();
     m_rpmlib->rpmFreeRpmrc();
     ms_instantiated = false;
 }
--- a/src/data_provider/src/packages/rpmlib.h
+++ b/src/data_provider/src/packages/rpmlib.h
@@ -27,6 +27,19 @@ class RpmLib final : public IRpmLibWrapper
             ::rpmFreeRpmrc();
         }
 
+        void rpmFreeMacros() override
+        {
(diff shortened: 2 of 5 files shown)


cborla (approved):
LGTM
- [x] Code changes reviewed

PR #38382, opened 2026-08-14
Prevent Linux audit plugin starvation when the Wazuh agent starts during a manager outage

### LINKED ISSUE #38310: whodata (audit): Agent deadlock when manager is disconnected at startup
Wazuh version | Component | Install type | Install method | Platform
-- | -- | -- | -- | --
4.10.1 | Agent | Agent | Not specified in thread | Not specified in thread

- audisp socket buffer fills and blocks all audit plugins

The user from this community issue ([URL](https://wazuh.slack.com/archives/C07BK5RJM3R/p1783686305019789)) identified an issue in the whodata/audit initialization flow where a disconnected manager at agent startup causes the audit_main() thread to block indefinitely, filling the auditd socket buffer and stalling all audisp plugins, since audisp dispatches to plugins on a single thread.

### Environment

Agent version tested: 4.10.1 (reporter noted the issue persisted even after narrowing whodata scope to two file paths)

The Wazuh team noted that the reader-thread/parser-thread separation introduced in v4.9.0 was expected to prevent this class of hang, and pointed to two other whodata fixes in the 4.14 line (#32383 in 4.14.1, #17957 in 4.14.5) as possibly relevant, but the reporter's reproduction with a full root-cause trace suggests a distinct or still-present bug path

### Root Cause (as identified by reporter)

The user described two scenarios during agent startup:

Case 1 (normal path): audit_init() creates the socket and passes its health check; audit_main() starts; the agent connects to the manager, fim_whodata_initialize() sets audit_db_consistency_flag = 1; audit_main() proceeds into audit_read_events() and reads from the socket normally.

Case 2 (bug path): audit_init() creates the socket and the health check passes, but the main thread blocks in fim_scan() → SendMSGPredicated() because there is no manager connection; since fim_whodata_initialize() is never called, audit_db_consistency_flag stays at 0, so audit_main() waits forever in the w_cond_wait(&audit_db_consistency, &audit_mutex) loop.

According to the reporter, because audit_read_events() is never reached, no recv() calls are made on the socket; the socket buffer fills with audit events (64–256KB), audisp-af's write() call then blocks once the buffer is full, and because audisp fans out to plugins on a single thread, this stalls all audit plugins system-wide, not just Wazuh.

### Proposed Fix

The user recommended closing the socket before entering the wait on the DB consistency flag, then recreating and reopening it for reading only after the flag is set, reporting that this resolved the issue in their testing. 

The suggested patch to audit_main():

```
void *audit_main(audit_data_t *audit_data) {
    char *path = NULL;
    directory_t *dir_it = NULL;
    OSListNode *node_it;
    count_reload_retries = 0;
    atomic_int_set(&audit_thread_active, 0);

    w_mutex_lock(&audit_mutex);
    atomic_int_set(&audit_thread_active, 1);
    w_cond_signal(&audit_thread_started);

    // Close socket before waiting for db consistency to prevent buffer filling
    minfo("FIM Audit: Closing socket %d before waiting for DB consistency flag", audit_data->socket);
    close(audit_data->socket);
    audit_data->socket = -1;

    while (!audit_db_consistency_flag) {
        w_cond_wait(&audit_db_consistency, &audit_mutex);
    }

    w_mutex_unlock(&audit_mutex);

    // Recreate socket after db consistency flag is set
    minfo("FIM Audit: DB consistency flag set, recreating audit socket");
    audit_data->socket = init_auditd_socket();
    if (audit_data->socket < 0) {
        merror("FIM Audit: Failed to recreate audit socket after DB consistency wait");
        return NULL;
    }
    minfo("FIM Audit: Successfully recreated audit socket (fd=%d)", audit_data->socket);

    if (audit_data->mode == AUDIT_ENABLED) {
        // Start rules reloading thread
        w_create_thread(audit_reload_thread, NULL);
    }

    minfo(FIM_WHODATA_STARTED);

    // Read events
    audit_read_events(&audit_data->socket, &audit_thread_active);

    // Auditd is not running or socket closed.
    mdebug1(FIM_AUDIT_THREAD_STOPED);
    close(audit_data->socket);
}
```

### Impact

Per the reporter, this affects any environment where the agent starts (or restarts) while the manager is unreachable and whodata/audit is enabled, causing a wider system-level failure of all audisp-based audit plugins rather than an isolated Wazuh issue.

Darioortegaleyva:
## Description

This PR fixes a defect where a Wazuh agent using FIM whodata with the **audit** provider leaves the audisp socket connected but undrained when it (re)starts while the manager is unreachable. On systems whose audit userspace is older than 3.1.1 (the blocking builtin `af_unix` plugin), this stalls `audispd`'s single-threaded dispatch loop and **starves every other audit plugin on the host** until the manager reconnects. On newer audit the agent still silently drops its own whodata events for the whole outage.

The root cause: the audit socket is connected in `audit_init()` before the reader is allowed to drain it, and the reader is gated behind `audit_db_consistency_flag`, which is only set after the initial baseline scan. When the manager is down, `fim_scan()` blocks on the agent's `.wait` lock, so `fim_whodata_initialize()` is never reached, the flag is never set, and `audit_main()` parks in `w_cond_wait()` **before** `audit_read_events()` — never calling `recv()`. The socket buffer fills and the builtin `af_unix` `write()` (blocking, recovers only on `EPIPE`) stalls the whole dispatcher.

The fix does not hold the socket connected while gated: it closes the socket before waiting on the flag and reconnects (with retries) right before `audit_read_events()`. While gated no client is attached, so `audispd`'s `write()` fails fast (`EPIPE`) and keeps servicing the other plugins.

**Issue:** wazuh/wazuh#38310
**Internal Reference:** N/A

## Proposed Changes

- Updated `src/syscheckd/src/whodata/syscheck_audit.c`: in `audit_main()`, close `audit_data->socket` before the `while (!audit_db_consistency_flag) w_cond_wait(...)` gate, and reconnect via `init_auditd_socket()` (retrying up to `MAX_CONN_RETRIES`) immediately before `audit_read_events()`. If the reconnect fails, mark the thread inactive and exit cleanly instead of running the reader on an invalid socket.
- Updated `CHANGELOG.md` with the fix entry under `v4.14.9`.

### Results and Evidence

Validated end-to-end on **Ubuntu 20.04.6 LTS, audit 2.8.5 (blocking builtin `af_unix`)** — the reporter's platform — with the agent pointed at an unreachable manager (`240.0.0.1`) and (re)started while it is down. A second, independent audisp plugin (`plugin2`, appends every event to a log) measures whether the dispatcher is starved. Same host, same manager-down condition, same event churn; the only difference between runs is this one-file patch (`wazuh-syscheckd` rebuilt from `v4.14.7` + the patch).

```bash
# Reproduction driver (per run): install a 2nd audisp plugin, force whodata=audit on
# /etc, point the agent at a dead manager, restart, churn /etc, then measure.
sudo bash repro_38310_e2e.sh <stock|patched>

# --- STOCK 4.14.7 ---
# ss -xp on the audit socket during the outage:
u_str ESTAB 0 213248 /var/ossec/queue/sockets/audit ... users:(("audispd",pid=21189,fd=8))
#   -> Send-Q = 213248 (full): audispd's builtin write to the undraining Wazuh client is blocked
SECOND-PLUGIN lines received during 20s churn window: 0
VERDICT[stock]: STARVED (bug reproduced)

# --- PATCHED (v4.14.7 + this fix) ---
# ss -xp on the audit socket during the outage:
#   (no ESTAB connection to queue/sockets/audit -> the socket is not held while gated)
SECOND-PLUGIN lines received during 20s churn window: 10650
VERDICT[patched]: FLOWING (not starved)
```

The mechanism was additionally proven from the vendored audit userspace source (`src/external/audit-userspace/audisp/audispd.c` single-threaded `event_loop` calling the builtin `af_unix` write inline; `audispd-builtins.c` `send_af_unix_string` blocking, recovering only on `EPIPE`), and the agent-side undrained-socket state was reproduced on Ubuntu 24.04 (audit 3.1.2) via `ss -xp` (syscheckd Recv-Q backlog + peer Send-Q full).

### Artifacts Affected

- `src/syscheckd/src/whodata/syscheck_audit.c`
- `CHANGELOG.md`

### Configuration Changes

N/A

### Documentation Updates

N/A

### Tests Introduced

- **Scope:** Manual Integration Test.
- **Details:** Built `wazuh-syscheckd` from `v4.14.7` with and without the patch and swapped only that binary on the same Ubuntu 20.04 host. Confirmed that a second audisp plugin receives 0 events with the stock binary (starved) and continues receiving events with the patched binary (not starved), while the manager stays unreachable. `audit_main()` is a thread entry point excluded from unit-test coverage (`LCOV_EXCL`), so no unit test is added.

## Review Checklist

    - [ ] Linux
    - [ ] Windows
    - [ ] MAC OS X

- Memory tests for Linux

    - [ ] Coverity
    - [ ] AddressSanitizer

**General Checklist:**

- [ ] Code changes reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ All notable changes to this project will be documented in this file.
 - Fixed WPK upgrade failing on agents without the `find` binary. ([#38431](https://github.com/wazuh/wazuh/pull/38431))
 - Bounded the `snort-full` log record appends to the available buffer space and sized the queued preprocessor message from its own line in `wazuh-logcollector`. ([#38472](https://github.com/wazuh/wazuh/pull/38472))
 - Restricted the GitHub, Office 365 and MS Graph wodles pagination and content retrieval to the configured API host, so the authentication token is not sent to another host. ([#38688](https://github.com/wazuh/wazuh/pull/38688))
+- Fixed whodata (audit) holding the audisp socket undrained at startup when the manager is unreachable, starving other audit plugins. ([#38382](https://github.com/wazuh/wazuh/pull/38382))
 
 ## [v4.14.8]
 
--- a/src/syscheckd/include/syscheck.h
+++ b/src/syscheckd/include/syscheck.h
@@ -24,6 +24,7 @@
 /* Audit defs */
 #define WDATA_DEFAULT_INTERVAL_SCAN 300
 #define AUDIT_SOCKET                "queue/sockets/audit"
+#define MAX_CONN_RETRIES            5          // Max retries to reconnect to Audit socket
 #define AUDIT_CONF_FILE             "tmp/af_wazuh.conf"
 #define AUDIT_HEALTHCHECK_DIR       "tmp"
 #define AUDIT_HEALTHCHECK_KEY       "wazuh_hc"
@@ -545,6 +546,14 @@ int set_auditd_config(void);
  * @return File descriptor of the socket, -1 on error
  */
 int init_auditd_socket(void);
+
+/**
+ * @brief Reconnect to the Auditd socket, retrying up to MAX_CONN_RETRIES times
+ *
+ * @param audit_sock Output: file descriptor of the reconnected socket, -1 on failure
+ * @return File descriptor of the socket, -1 on error
+ */
+int audit_reconnect_with_retries(int *audit_sock);
 
 /**
  * @brief Reloads audit rules every RELOAD_RULES_INTERVAL seconds
--- a/src/syscheckd/src/whodata/syscheck_audit.c
+++ b/src/syscheckd/src/whodata/syscheck_audit.c
@@ -536,9 +536,9 @@ void *audit_main(audit_data_t *audit_data) {
         audit_data->socket = init_auditd_socket();
     }
     if (audit_data->socket < 0) {
-        merror(FIM_ERROR_WHODATA_SOCKET_CONNECT, AUDIT_SOCKET);
+        mwarn(FIM_WARN_AUDIT_THREAD_NOSTARTED);
         atomic_int_set(&audit_thread_active, 0);
-        return NULL;
+        goto whodata_teardown;
     }
 
     if (audit_data->mode == AUDIT_ENABLED) {
@@ -555,6 +555,7 @@ void *audit_main(audit_data_t *audit_data) {
     mdebug1(FIM_AUDIT_THREAD_STOPED);
     close(audit_data->socket);
 
+whodata_teardown:
     // Clean regexes used for parsing events
     clean_regex();
 
--- a/src/syscheckd/src/whodata/syscheck_audit.c
+++ b/src/syscheckd/src/whodata/syscheck_audit.c
@@ -24,7 +24,6 @@
 #define PLUGINS_DIR_AUDIT           "/etc/audit/plugins.d"
 #define AUDIT_CONF_LINK             "af_wazuh.conf"
 #define BUF_SIZE OS_MAXSTR
-#define MAX_CONN_RETRIES 5          // Max retries to reconnect to Audit socket
 
 // Global variables
 pthread_mutex_t audit_mutex;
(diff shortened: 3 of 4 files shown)


Miguevrgo (dismissed):
LGTM

MarcelKemp (approved):

PR #38392, opened 2026-08-14
Own the systemd D-Bus connection so the services collector recovers after a bus restart

### LINKED ISSUE #38364: `wazuh-modulesd` unbounded memory growth on Linux agents from resource leaks in the inventory path

anromerom:
## Description

Related to #38364.

`SystemdUnitsProvider::getSystemdUnits()` acquires a system bus connection on every invocation and releases it on none of its return paths.

> Note: This PR addresses only the second of the three findings tracked in that issue, which is why it does not close it. The remaining two are handled in their own PRs and the issue stays open until all three are merged.

The same code is present on `5.0.0`, so this change should be ported there as well.

## Proposed Changes

- `idbus_wrapper.hpp`: Replace `bus_get()` with `bus_get_private()`, and add `connection_close()` and `connection_unref()`.
- `dbus_wrapper.hpp`: Implement the three methods over `dbus_bus_get_private()`, `dbus_connection_close()` and `dbus_connection_unref()`, with the same null guard the existing `message_unref()` uses.
- `systemd_units_linux.cpp`: Add a `ScopedDBusConnection` guard that closes and unreferences the connection when it goes out of scope, and use it to hold the connection for the whole of `getSystemdUnits()`.
- `test_systemd_units_linux.cpp`: Rename the mocked method, add the two new ones, and assert the connection is released exactly once on the success path and on each failure path, and never when acquisition itself failed.

The method was renamed because the ownership semantics change. The caller now owns the connection and is responsible for releasing it. A silent change of behavior behind `bus_get()` would be misleading at every call site.

Adding a release method and calling it on all return paths, while keeping the shared connection from `dbus_bus_get()`, does not fix the observable symptom. libdbus caches the shared connection process-wide and holds its own reference, so dropping ours changes nothing: after the bus goes away, every later acquisition is handed back the same closed connection and the collector never recovers for the lifetime of the process.

A connection the caller owns, closed explicitly, does recover. That is why the change switches to `dbus_bus_get_private()` and pairs it with `dbus_connection_close()`.

### Results and Evidence

Measured on the shipped packages, both built from the same base commit through the same pipeline, on Debian 10
with a live system bus, driving `sysinfo_services()` from the installed `libsysinfo.so`.

The package-level observable is the connection lifecycle: one connection for a whole run means the shared
connection is reused and never released, one per scan means it is acquired and released each time.

| Package | Scans | Bus connects | Per scan |
|---|---:|---:|---:|
| Baseline, `4.14.9` at `7bacb80` | 20 | 1 | **0.05** |
| This fix, at `21a9a37` | 20 | 21 | **1.05** |

**Recovery, on the shipped code.** One long-lived process spanning a restart of the system bus, since a fresh
process would reconnect trivially:

| Package | Connects before the restart | Connects after |
|---|---:|---:|
| Baseline | 1 | **0**, never recovers |
| This fix | 21 | **20**, recovers |

### Artifacts Affected

- `libsysinfo.so`. No change to inventory data, scan semantics or configuration.

### Configuration Changes

None.

### Documentation Updates

None
### Tests Introduced

- No new test files. Extended the existing suite so every path in `getSystemdUnits()` asserts the connection lifecycle: released once on the success path, released once on each of the failure paths that acquired a connection, and never released when acquisition returned `nullptr`.

## Review Checklist

- [x] Code changes reviewed
- [x] Tests cover the new functionality (built and run locally, and mutation checked: 4 of 5 tests fail if the release is removed)
- [ ] Developer documentation reflects the changes (N/A)

### DIFF
--- a/src/data_provider/src/extended_sources/services/src/systemd_units_linux.cpp
+++ b/src/data_provider/src/extended_sources/services/src/systemd_units_linux.cpp
@@ -12,6 +12,46 @@
 #include "systemd_units_linux.hpp"
 #include "stringHelper.h"
 
+namespace
+{
+    /// @brief Owns a private D-Bus connection for the duration of a scope.
+    ///
+    /// bus_get_private() transfers ownership to the caller, so the connection has to be closed
+    /// and unreferenced on every exit path. Doing that through a guard rather than by hand
+    /// keeps a future early return from reintroducing the leak.
+    class ScopedDBusConnection final
+    {
+        public:
+            ScopedDBusConnection(std::shared_ptr<IDBusWrapper> dbusWrapper, DBusConnection* connection)
+                : m_dbusWrapper(std::move(dbusWrapper))
+                , m_connection(connection)
+            {
+            }
+
+            ~ScopedDBusConnection()
+            {
+                if (m_connection)
+                {
+                    m_dbusWrapper->connection_close(m_connection);
+                    m_dbusWrapper->connection_unref(m_connection);
+                }
+            }
+
+            ScopedDBusConnection(const ScopedDBusConnection&) = delete;
+            ScopedDBusConnection& operator=(const ScopedDBusConnection&) = delete;
+
+            /// @brief Returns the owned connection, or nullptr if it could not be acquired.
+            DBusConnection* get() const
+            {
+                return m_connection;
+            }
+
+        private:
+            std::shared_ptr<IDBusWrapper> m_dbusWrapper;
+            DBusConnection* m_connection;
+    };
+}
+
 SystemdUnitsProvider::SystemdUnitsProvider(std::shared_ptr<IDBusWrapper> dbusWrapper)
     : m_dbusWrapper(std::move(dbusWrapper))
 {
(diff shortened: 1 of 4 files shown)


cborla:
`DBusWrapper::bus_get_private()` (`dbus_wrapper.hpp:47-50`) doesn't disable `exit_on_disconnect` on the new private connection, so this fix only recovers from a bus restart that happens *between* scans, not one that lands mid-call.

`internal_bus_get()` sets it unconditionally for both `dbus_bus_get()` and `dbus_bus_get_private()` (`external/dbus/dbus/dbus-bus.c:504-505`), and `dbus_connection_send_with_reply_and_block()` — the exact call `getSystemdUnits()` makes at `systemd_units_linux.cpp:126` (timeout `-1` = `DBUS_TIMEOUT_USE_DEFAULT`, a real ~25s window, not infinite) — calls `_dbus_exit(1)` as soon as it processes a "Disconnected" message with that flag still `TRUE` (`external/dbus/dbus/dbus-connection.c:4339-4345`, reached via line 3669). So a bus restart while the blocking call is in flight kills the whole process hosting `libsysinfo.so`, not just this scan.

Pre-existing default (the old shared connection had it too), but since this PR already takes over the connection's full lifecycle, it's the natural place to close it:

```cpp
DBusConnection* bus_get_private(DBusBusType type, DBusError* error) override
{
    DBusConnection* connection = ::dbus_bus_get_private(type, error);
    if (connection)
    {
        ::dbus_connection_set_exit_on_disconnect(connection, FALSE);
    }
    return connection;
}
```

`ScopedDBusConnection` and the existing `NULL`/error-reply path right after this call already handle the resulting `NULL` reply correctly, so nothing else needs to change.

Not asking to block this PR on it — just flagging it since this is the natural place to fix it. Worth a gut check on how cheap this is to actually apply here (vs. tracking it separately): is there any reason not to just add the `set_exit_on_disconnect(FALSE)` call in `bus_get_private()` now?

cborla (approved):
LGTM
- [x] Code changes reviewed

PR #38395, opened 2026-08-14
Resolve group names once per scan instead of once per user and group

### LINKED ISSUE #38364: `wazuh-modulesd` unbounded memory growth on Linux agents from resource leaks in the inventory path

anromerom:
## Description

Related to #38364.

The users collector resolves each user's group names with one name service lookup per (user, group) pair, and asks for group membership once per user rather than once per scan. On a host where accounts are served by a network directory, every one of those lookups is a directory round trip.

> Note: This PR addresses only the third of the three findings tracked in that issue, which is why it does not close it. The other two are handled in their own PRs and the issue stays open until all three are merged.

The same code is present on `5.0.0`, so this change should be ported there as well.

## Proposed Changes

- `user_groups_linux.cpp`: Memoize gid to group-name resolution inside `getGroupNamesByUid()`, so each distinct gid is resolved once per call instead of once per user that belongs to it. Unresolvable gids are memoized too, otherwise a group that cannot be resolved keeps costing one lookup per member.
- `sysInfoLinux.cpp`: Call `getGroupNamesByUid()` once with the full set of uids and index into the result, rather than once per user. The batch signature already existed and was unused. Both return shapes are handled, since the method returns a bare array for a single uid and a uid-keyed object otherwise, and a host with exactly one user takes the array path.

No interface, configuration or output changes. The reported `user_groups` field is unchanged.

### Results and Evidence

Measured on the shipped packages, both built from the same base commit through the same pipeline, on Debian 10
with accounts served by a stand-in metadata service on loopback that counts the requests it serves.

| Directory accounts | Requests per scan, baseline | With this fix | Growth per scan, baseline | With this fix |
|---:|---:|---:|---:|---:|
| 0 | 0.0 | 0.0 | 0.00 KB | 0.00 KB |
| 10 | 160.9 | 16.2 | 120.96 KB | 10.24 KB |
| 50 | 804.5 | 16.6 | 616.96 KB | 11.90 KB |
| 200 | 3,251.4 | 21.9 | 2,466.97 KB | 8.29 KB |

The baseline costs about 16 requests per account per scan. Against this directory component the per-scan
volume with the fix is nearly independent of the account count. The zero-account row is the control: no
directory accounts means no requests and no growth on either package.

How much of the per-account cost disappears depends on what the directory client answers locally, so the
same change was also measured against SSSD over OpenLDAP, with 200 accounts in 20 overlapping groups and
SSSD caching disabled so every lookup reaches the server:

| Directory accounts | Queries per collection, baseline | With this fix | Ratio |
|---:|---:|---:|---:|
| 50 | 764.7 | 145.5 | 5.3x |
| 200 | 3,133.6 | 453.1 | 6.9x |

There the cost still scales with the account count, 4.1x for the baseline and 3.1x for the fix across those
two rows, because SSSD sends the per-user membership query to the directory while the OS Login module answers
it from a local cache file. So the honest summary is that the per-(user, group) amplification is removed in
both cases, and whether what remains looks flat depends on the client. Collection time for the 200 account
arm fell from 1,342 s to 85 s.

<details>
<summary><b>Tests executed</b></summary>

Every comparison is between the shipped `4.14.9` package at the base commit and the package built from this
branch. Unless stated otherwise a test ran on both packages under identical conditions on the same host, with the same accounts.

### Reported data is unchanged

| Test | What it does | Result |
|---|---|---|
| Output equivalence | Dumps what the users and groups collectors report and compares the documents field by field | **Identical** on every platform. Byte-identical canonical digests, 31 user records and 44 group records on the directory-backed host |
| User in 84 groups | A user belonging to more than the 64 group internal limit, which forces the membership query to be repeated with a larger buffer | **All 84 names reported**, identical before and after |
| User with no supplementary groups | Only a primary group | **Identical**, one name |
| Duplicate uid alias | Two account names sharing uid 0, so one lookup result is shared by two records | **Identical**, and the alias reports exactly what the canonical name reports |
| uid above `INT_MAX` | An account with uid 4294967294, which is where the batched result is looked up by key | **Groups reported correctly.** An earlier revision of this change failed this: the result is keyed from `uid_t` while the lookup formatted the value narrowed to `int`, so it searched for `-2` instead of `4294967294` and reported no groups. Fixed before submission |
| Single-user host | Only one account present, which is the one case where the batch call returns a bare array instead of a uid-keyed object | **Groups reported correctly** through that branch |
| Unresolvable gid | Membership in a gid with no matching group entry | **Omitted with no effect on the other names**, identical before and after |

### Behaviour across repeated scans

| Test | What it does | Result |
|---|---|---|
| Repeated collections | Collects many times in one process and compares each result against the first | **No drift.** Identical digest on every iteration |
| Directory outage and recovery | Collects with the directory up, with it stopped, and after it returns. The client cache is taken out of the path first so a group name can only come over the network | **Full recovery.** All directory users lose their group name during the outage and the dump after recovery is byte-identical to the healthy one |
| In-place upgrade | Runs the old package, then installs the new one over it without touching the accounts | **No change in reported data**, identical digest |

### Cost of a collection

| Test | What it does | Result |
|---|---|---|
| Request volume, network directory with a per-request leak | Counts requests reaching the directory per collection at 0, 10, 50 and 200 accounts | **160.9 to 16.2 per scan** at 10 accounts, and with this change the volume stops growing with the account count. Zero accounts gives zero requests on both packages |
| Request volume, SSSD over LDAP | Counts LDAP searches per collection with the client cache disabled, 120 collections per measurement | **5.3x fewer** at 50 accounts, **6.9x fewer** at 200 |
| Collection time | Wall clock for the same work | 200 accounts over 120 collections: **1,342 s to 85 s** |
| Collector attribution | Measures the users and groups collectors separately | All of the cost is the users collector. The groups collector is 0.5 requests per scan and 0.00 KB either way |

### Memory

| Test | What it does | Result |
|---|---|---|
| Growth against the leaking directory component | Per-scan growth at 0, 10, 50 and 200 accounts | **120.96 KB to 10.24 KB per scan** at 10 accounts, and no longer proportional to the account count |
| Where the growth comes from | Same runs against the first release of that component which releases its parsed response, and against local account files only | **0.00 KB per scan** on both packages with request volume unchanged, and 0.00 KB against local files. The growth belongs to the directory component, not to this code |
| Accumulation check | 120 collections against a client that does not leak, watching live heap rather than a single total | **Nothing accumulates.** Live heap 211 KB start to 211 KB end on the baseline and 213 KB to 213 KB with this change; per-scan growth falls as the run lengthens or goes negative |

### Platforms

Output equivalence, repeated-collection stability, every edge case above and a no-crash check ran on nine
platforms (locally using Docker containers) covering both packaging families and glibc 2.26 through 2.38: Ubuntu 22.04 and 24.04, Debian 10 and 12, Rocky Linux 8 and 9, Amazon Linux 2 and 2023, openSUSE Leap 15. **Zero failures.**

Not covered: arm64, and macOS and Windows. The latter two have their own implementations of this collector,
untouched by this change.

### Unit tests

Four cases added, three of them for behaviour that had no coverage at all. Two fail on the base commit with
`called twice - over-saturated` and pass with this change, so they bind to it rather than passing either way.
A third asserts a failed group lookup is retried by the next member instead of being treated as authoritative,
and fails on an earlier revision of this change. The fourth guards the single-uid return shape and passes on
both, as a regression guard should.

</details>

### Artifacts Affected

- `libsysinfo.so`. No change to inventory data, scan semantics or configuration.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

In `test_user_groups_linux.cpp`, which had no coverage of `getGroupNamesByUid()` on Linux:

- `getGroupNamesByUidResolvesEachGidOnce`: two users sharing two groups, asserting one `getgrgid_r()` per distinct gid instead of one per (user, group) pair, and that both users still report both group names.
- `getGroupNamesByUidRetriesUnresolvableGidOnlyOnce`: a gid the group database cannot resolve, shared by two users, asserting it costs one lookup rather than one per member and still contributes no name.
- `getGroupNamesByUidKeepsArrayShapeForOneUid`: the single uid call still returns a bare array of names.

The first two fail on the base commit and pass with this change. The third passes on both, guarding the return shape.

Full `data_provider` suite on this branch: 13 of 14 pass. The one failure is `sysInfoPackagesLinuxHelper_unit_test`, which cannot load `libarchive.so.13` on the test host and is unrelated to these files.

## Review Checklist

- [x] Code changes reviewed
- [ ] Developer documentation reflects the changes (N/A)

### DIFF
--- a/src/data_provider/src/extended_sources/groups/src/user_groups_linux.cpp
+++ b/src/data_provider/src/extended_sources/groups/src/user_groups_linux.cpp
@@ -109,17 +109,30 @@ nlohmann::json UserGroupsProvider::getGroupNamesByUid(const std::set<uid_t>& uid
         }
 
         struct group grp {};
+
         struct group* grpResult = nullptr;
+
         auto groupBuf = std::make_unique<char[]>(bufSize);
 
-        if (m_groupWrapper->getgrgid_r(gid, &grp, groupBuf.get(), bufSize, &grpResult) == 0 && grpResult != nullptr)
+        const auto result = m_groupWrapper->getgrgid_r(gid, &grp, groupBuf.get(), bufSize, &grpResult);
+
+        if (result == 0 && grpResult != nullptr)
         {
             name = grpResult->gr_name;
             gidToName.emplace(gid, name);
             return true;
         }
 
-        unresolvableGids.insert(gid);
+        // Only an authoritative answer is remembered. A return of 0 with no entry means the
+        // group database does not know this gid, which will not change within one scan. A
+        // non-zero return is a failed lookup, which may be transient on a host whose group
+        // database is served over the network, so it is left unmemoized and the next user
+        // that belongs to this gid attempts it again, as it did before this change.
+        if (result == 0)
+        {
+            unresolvableGids.insert(gid);
+        }
+
         return false;
     };
 
--- a/src/data_provider/src/extended_sources/groups/src/user_groups_linux.cpp
+++ b/src/data_provider/src/extended_sources/groups/src/user_groups_linux.cpp
@@ -12,6 +12,9 @@
 #include "passwd_wrapper.hpp"
 #include "system_wrapper.hpp"
 #include <iostream>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
 
 // Reasonable upper bound for getpw_r buffer
 constexpr size_t MAX_GETPW_R_BUF_SIZE = 16 * 1024;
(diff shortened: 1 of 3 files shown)


cborla (approved):
LGTM
- [x] Code changes reviewed

PR #36790, opened 2026-06-08
Fix agent permanently stuck when TCP connection is silently half-closed

nbertoldo:
## Description

Closes #36743.

When the agent is running under high log throughput and the TCP connection to the manager silently dies (no FIN/RST reaches the agent — e.g. a network device times out and drops the flow), `wazuh-agentd` can become permanently stuck and stop processing events until the process is manually restarted.

**Root cause chain:**

1. The manager's receive buffer fills and its TCP window drops to zero, activating the kernel **persist timer** on the agent side.
2. The agent has no `SO_SNDTIMEO` set, so `send()` blocks indefinitely while holding `send_mutex`.
3. The main thread is waiting on `send_mutex` and never reaches `select()`, so it never reads the FIN sent by the manager → socket stays in `CLOSE_WAIT`.
4. `force_reconnect_interval` (and `max_time_reconnect_try`) runs inside `run_notify()` in the **same main thread** → it never fires.
5. Recovery only happens when the kernel exhausts `tcp_retries2` (~15 min with kernel defaults), but under high traffic the kernel switches to the **persist timer**, which is **not** bounded by `tcp_retries2` and can run indefinitely. A real-world case showed 11 days of complete agent silence.

### `src/client-agent/start_agent.c`
- After a successful `OS_ConnectTCP()`, set `SO_KEEPALIVE` and `SO_SNDTIMEO` on every new TCP socket.
- Keepalive parameters (`tcp_keepidle`, `tcp_keepintvl`, `tcp_keepcnt`) and send timeout (`send_timeout`) are read from `internal_options.conf` via the existing `getDefine_Int()` mechanism.

### `src/client-agent/sendmsg.c`
- Move dead-socket detection **inside** `send_mutex` (previously there was a window between `unlock` and a second `lock` where another thread could attempt a send on the same broken descriptor).
- Handle `ECONNREFUSED` and `ENOTCONN` as fatal errors (previously they fell through to `sleep(1)` with no socket invalidation).
- On any fatal error (`EPIPE`, `ECONNRESET`, `ETIMEDOUT`, `ECONNREFUSED`, `ENOTCONN`) close the socket and set `agt->sock = -1` while holding the mutex so all threads see a consistent state.

### `src/client-agent/agentd.c`
- Check `agt->sock < 0` at the **top** of the monitor loop before `run_notify()` so a keep-alive is never attempted on a closed descriptor.
[...]

Issue #36743, opened 2026-06-02
agentd does not detect silently half-closed TCP connections


cborla:
## Description

When the TCP connection between `wazuh-agentd` and the manager is silently torn down at the network layer (overlay/IPSec glitch, NAT/conntrack flush, middlebox), the agent can get **permanently stuck**: the socket sits in `CLOSE_WAIT`, the state file keeps reporting `status='connected'`, and the agent **never attempts to reconnect** — not via `time-reconnect` nor via `force_reconnect_interval`.

The agent process stays alive and `systemctl status wazuh-agent` reports it as active, but communication with the manager is dead. Only a manual restart of the agent recovers it.

Observed in **v4.14.4**, but the relevant code is present across 4.x.

## Environment (field case)

- Wazuh agent **v4.14.4**, Ubuntu 20.04.6 LTS, kernel 5.4.0-182-generic, x86_64.
- Agent runs natively as a systemd service (not containerized).

## Evidence from the affected agent

```
# State file — frozen 11 days
$ cat /var/ossec/var/run/wazuh-agentd.state
status='connected'
last_keepalive='2026-05-10 21:01:46'
last_ack='2026-05-10 21:01:55'
msg_buffer='0'

# Kernel has no socket to the manager, yet network is reachable right now
$ nc -zv <manager> 1514          → succeeded
$ conntrack -L | grep <manager>  → empty
$ ss -tnp | grep 1514            → CLOSE-WAIT with unread bytes in recv buffer

# logcollector cache overflowing while anti-flood buffer is empty
wazuh-logcollector: ERROR: Discarding audit message because cache is full
```

The combination *(kernel: no socket / process: still "connected")* is the signature: the application and the kernel disagree about the connection state.

## Root cause (code analysis)

> ⚠️ Line numbers below come from the 4.14.x analysis — confirm against the actual checkout before relying on them.

1. **The agent→manager socket has no `SO_KEEPALIVE`.**
   `OS_ConnectTCP` → `OS_Connect` (`src/os_net/os_net.c:251-352`) never calls `OS_SetKeepalive`. `OS_SetKeepalive` exists (`os_net.c:611`) but is only invoked from `remoted/remoted.c:67` (manager side) and `wm_fluent.c`. With no keepalive, the kernel never tears down a silently half-closed connection on its own.

2. **No `SO_SNDTIMEO` / `SO_RCVTIMEO` on the socket either.**
[...]

PR #38401, opened 2026-08-14
Detect and recover from silently half-closed TCP connections

### LINKED ISSUE #38388: Backport: Fix agentd silently half-closed TCP connection detection
## Backport scope

This issue tracks the backport of the fix for #36743 to the `4.14.9` release branch.

## Issue summary

When the TCP connection between `wazuh-agentd` and the manager is silently torn down at the network layer (overlay/IPSec glitch, NAT/conntrack flush, middlebox), the agent can get permanently stuck with the socket in `CLOSE_WAIT` state, `status='connected'` reported indefinitely, and no reconnection attempts — not via `time-reconnect` nor via `force_reconnect_interval`.

**Observed in:** v4.14.4+ (affects all 4.x versions)

## Proposed fix (from #36743)

Three minimal changes in `client-agent/`:

1. Set `SO_KEEPALIVE` + keepalive options on `agt->sock` after `OS_ConnectTCP` in `connect_server()`.
2. Set `SO_SNDTIMEO` on the same socket so `send()` cannot block indefinitely while holding `send_mutex`.
3. In `send_msg`, on `EPIPE`/`ECONNRESET`/`ETIMEDOUT`, invalidate the socket (`OS_CloseSocket`; `agt->sock = -1`) and let the main loop reconnect.

## Testing

- Verify the fix resolves the permanent stuck state in #36743 reproduction scenario.
- Confirm no regressions in normal agent-manager communication.
- Test with network disruption scenarios (NAT/conntrack flush, packet drops, overlay glitches).

## Related issues

- #36743 — Original issue (4.14.4+)

## References

See #36743 for detailed root cause analysis and reproduction steps.

cborla:
## Description

Backport of #36790 (fix for #36743) to the `4.14.9` branch.

When the TCP connection between `wazuh-agentd` and the manager is silently torn down at the network layer (a middlebox evicting the flow, a NAT/conntrack flush, an overlay/IPSec glitch, or the peer stopping to read and advertising a zero window), the agent can get stuck indefinitely: `status='connected'` in `wazuh-agentd.state` forever, `last_keepalive`/`last_ack` frozen, and no reconnection attempt — because the blocking `send()` in `send_msg()` never returns, and it holds `send_mutex` while it's stuck. This also blocks the same single-threaded monitor loop that would otherwise notice the staleness and reconnect (`run_notify()`'s own watchdog never gets to run again either), so the existing `time-reconnect`/`force_reconnect_interval` settings can't kick in.

Closes #38388

## Proposed Changes

Same three changes as #36790, ported to the pre-refactor 4.14.x code (this branch still has the UDP transport path in `send_msg()`, which `main` no longer has, and uses the original `src/client-agent/*.c` layout, so this is a manual port rather than a clean cherry-pick):

- `start_agent.c` (`connect_server()`): after a successful TCP connect, enable `SO_KEEPALIVE` (+ `TCP_KEEPIDLE`/`TCP_KEEPINTVL`/`TCP_KEEPCNT`) and `SO_SNDTIMEO` on `agt->sock`, so a `send()` cannot block forever.
- `sendmsg.c` (`send_msg()`): on `EPIPE`/`ECONNRESET`/`ENOTCONN`/`ECONNREFUSED`/`ETIMEDOUT`/`EAGAIN`/`EWOULDBLOCK`, close and invalidate the socket (`agt->sock = -1`) instead of just logging and sleeping, so the connection isn't reused in a broken state. Also added an early return under `send_mutex` when `agt->sock` was already invalidated by another thread, avoiding a stale-`errno` log.
- `agentd.c` (monitor loop): reconnect immediately when `agt->sock < 0`, instead of relying solely on `run_notify()`'s staleness watchdog (which requires `receive_msg()` to have run recently on the same socket — exactly what doesn't happen while the loop is stuck). Also snapshots the descriptor once per iteration before `FD_SET`/`select`/`FD_ISSET` to avoid a TOCTOU if a sender thread invalidates `agt->sock` mid-iteration, and treats `EBADF` from `select()` as a lost connection instead of a fatal error.
- `client-config.h`: `agt->sock` is now `_Atomic int`, since it's read/written from the main loop and from sender threads (`dispatch_buffer`, `req_receiver`).
- `internal_options.conf`: new `agent.tcp_keepidle` (60), `agent.tcp_keepintvl` (15), `agent.tcp_keepcnt` (4), `agent.send_timeout` (30) — documented inline, following the same convention as the existing `remoted.tcp_keepidle/intvl/cnt` entries.

Two additional fixes on top of the above, found during self-review of this port and confirmed to also be present, unmodified, in `origin/main`'s current `sendmsg.c`/`start_agent.c` (i.e. not something this port introduced — the original fix has the same gaps):

- `sendmsg.c`: `EBADF` was missing from the fatal-error classification. A `send()` on an invalid/closed descriptor fell into `default: socket_dead = false`, so the agent would keep retrying `send()` + `sleep(1)` forever on a dead fd instead of ever reconnecting. Added `case EBADF:` alongside the other fatal errnos, both in the classification switch and the logging switch.
- `start_agent.c`: `connect_server()` closed the old socket and assigned the new one directly on `agt->sock`, with no locking at all — while `send_msg()` only ever touches `agt->sock` under `send_mutex`. Since this PR makes `connect_server()` run immediately in reaction to any sender's own send failure (the new `agentd.c` block above), that unsynchronized write is exercised far more often than before. Added `send_mutex_lock()`/`send_mutex_unlock()` (new, exposed from `sendmsg.c`/`agentd.h`) and moved every `agt->sock` mutation in `connect_server()` inside that same lock; the slow parts (DNS resolution, `connect()`, keepalive/timeout `setsockopt`s) now operate on a local `new_sock` variable and are published to `agt->sock` only once fully configured, so senders are never blocked waiting on them and never observe a half-configured socket either.

Since these two gaps are also live on `main` today, worth raising separately against #36743/#36790 so 5.0.0 gets the same fix, not just this backport.

Further fixes from review feedback (Luis + a follow-up self/automated review), all in `sendmsg.c`/`start_agent.c`/`agentd.c`/`receiver.c`/`notify.c`:

- **Partial write not detected** — `OS_SendSecureTCP()` zeroes `errno`/`WSASetLastError` right before `send()`, so a short write under `SO_SNDTIMEO` (data partially queued, then the timeout fires) reports `retval != 0` with no errno set at all, previously falling into `default: socket_dead = false`. That leaves the connection "alive" with the length-prefixed framing already corrupted for every later message. Added `case 0:` to both the POSIX and the new Windows classification (see below), tested by `test_send_msg_partial_write`.
- **Windows had no dead-socket detection at all** — the whole classification/close block was `#ifndef WIN32`-only, so on Windows the bounded `SO_SNDTIMEO` fired but the socket was never invalidated, and `receiver.c`'s receive loop (`if (agt->sock == -1) { sleep(5); continue; }`) never actually reconnected either. Added the `WSAGetLastError()`-mapped equivalent classification, `WSASetLastError(0)` in `OS_SendSecureTCP()`, and fixed the Windows loop to call the shared reconnect helper instead of just sleeping. Compiles clean with `i686-w64-mingw32-gcc`; not runtime-tested (no Windows box here).
- **Handshake fed an invalidated socket into `select()`** — `agent_handshake_to_server()` didn't check `send_msg()`'s return before calling `receive_message()`, which did `FD_SET(agt->sock, ...)` with `agt->sock` possibly `-1` (UB — negative shift) and then `select(0, ...)`, i.e. a plain `recv_timeout`-long (60 s default) sleep per handshake attempt. Now checks the return and bails out early; `receive_message()` also snapshots `agt->sock` once and guards against `-1` defensively.
- **`EBADF` handling could misattribute and leak an fd** — `select()` returns `EBADF` for *any* invalid descriptor in the set, not necessarily `agt->sock` (could be `agt->m_queue`). The handler blindly set `agt->sock = -1` without closing it first and without `send_mutex`, contradicting the locking this PR establishes everywhere else, and leaking the real descriptor if it wasn't actually the culprit. Now checks under `send_mutex` whether `agt->sock` is already `-1` (meaning a sender really did invalidate it) before deciding to reconnect vs. treat it as the pre-existing fatal case.
- **Fast-path silently dropped events at full throughput** — the early return when `agt->sock` is already `-1` had no `sleep(1)`/log unlike every other failure path, so during an outage `dispatch_buffer()` could drain its whole queue at `events_persec` (500/s default) discarding messages silently instead of at the ~1/s every other path is throttled to.
- **`rip_id`/`sock` publish order** — `connect_server()` published `agt->sock` before `agt->rip_id`, while `send_msg()` reads `agt->server[agt->rip_id].protocol` to pick UDP vs. TCP before ever taking the mutex; a failover between differently-configured servers could race a sender into using the wrong transport for the new fd. Reordered so `rip_id` is set first, both under the same lock.
- **No timeout on the handshake's receive path** — added `OS_SetRecvTimeout()` alongside `SO_SNDTIMEO`; before this, a manager delivering a partial reply and going silent hung the single-threaded handshake indefinitely (the same clinical picture as the send-side bug, just from the receive side).
- **`reconnect_to_server()` de-duplication** — the same `os_setwait()`/`GA_STATUS_NACTIVE`/`start_agent(0)`/`os_delwait()`/`GA_STATUS_ACTIVE` sequence was copy-pasted 5 times across `agentd.c`/`receiver.c`/`notify.c`; extracted into one shared helper. It lives in `start_agent.c` (not `agentd.c`) specifically because `agentd.o` is excluded from the Windows agent link (Windows uses `receiver.c`'s loop instead), which is exactly what broke CI the first time this was attempted — `start_agent.o` is linked into every target.
- **Log-spam and `errno` correctness on unsupported platforms** — `OS_SetKeepalive_Options()` warns 3x per reconnect on Windows/OpenBSD ("unsupported platform") since none of its parameters apply there; skipped the call entirely on those platforms instead. Also fixed `OS_SetKeepalive`/`OS_SetSendTimeout`/`OS_SetRecvTimeout` failure logs to use `win_strerror(WSAGetLastError())` on Windows instead of `strerror(errno)` (Winsock doesn't touch `errno`).
- **`ONEWAY_ENABLED` build regression** — under `make ONEWAY=yes`, `start_agent()` returns without ever connecting, so `agt->sock` stays `-1` forever; the new immediate-reconnect block in `agentd.c` would have spun at 100% CPU with a `LOST_ERROR`/`SERVER_UP` log flood instead of the old (already-broken, but at least paced) behavior. Guarded the new block with `#ifndef ONEWAY_ENABLED`. Compiles clean with `make ONEWAY=yes`.

### Results and Evidence

- Verified against the real 4.14.5/4.14.9 source (byte-identical between the two tags for every file this PR touches) that the gap exists: `OS_SetKeepalive`/`OS_SetSendTimeout` were already used by `remoted` and `wm_fluent`, but never applied to the agent's own outbound socket (`OS_ConnectTCP` → `OS_Connect` only sizes the send/recv buffers).
- Full `client-agent` unit test suite built and run locally (`-DTARGET=agent -DUNIT_TEST=ON`, ASan/UBSan enabled): **82/82 passing**, 0 failures, across all 7 targets — see breakdown below. Added `test_send_msg_ebadf` and `test_send_msg_partial_write` (same shape as the existing errno cases). No dedicated test for the `connect_server()` locking fix, the `EBADF`/`m_queue` disambiguation, or the `rip_id`/`sock` ordering fix: the current mock-based unit tests have no concurrent harness to exercise cross-thread ordering with, so these are covered by code inspection and the full suite still passing (no regression), not by new assertions.
- CI (`Unit-Tests (winagent)`) caught a real bug this local Linux-only testing couldn't: `reconnect_to_server()` initially lived in `agentd.c`, which is excluded from the Windows agent link, causing an `undefined reference` when linking `win32/wazuh-agent.exe` (`notify.c`/`receiver.c` call it and are linked there). Confirmed the fix by cross-compiling every touched file with `i686-w64-mingw32-gcc` and checking the symbol table directly (`start_agent.o` now exports `reconnect_to_server` as `T`; `notify.o`/`receiver.o` reference it as `U`), plus a full production link of the Linux `wazuh-agentd` binary — both clean. Did not attempt a full Windows `.exe` link locally (this checkout never had a prior Windows build, so it would mean building the whole agent from scratch); relying on CI for that final confirmation.
- Reproduced the original stuck-connection scenario against this exact patched binary (see "Manual tests" below for the full setup/timeline): the agent detected the stalled connection, stayed paced (no busy-loop) while retrying, and reconnected cleanly once the manager recovered.

  - [x] Linux — built `client-agent/{sendmsg,agentd,start_agent,notify,receiver}.o` (and their transitive includers: `buffer.o`, `config.o`, `event-forward.o`, `main.o`, `request.o`) via the project `Makefile`, zero warnings/errors.
  - [x] Windows
  - [x] MAC OS X
- [x] Unit tests (Linux, `-DTARGET=agent -DUNIT_TEST=ON`) — 82/82 passing, ASan/UBSan clean:

  | Target | Result |
  | --- | --- |
  | `test_sendmsg` (new) | 14/14 |
  | `test_start_agent` | 9/9 (2 new: keepalive-fails, send_timeout-fails) |
  | `test_notify` | 7/7 |
  | `test_agentd_state` | 19/19 |
  | `test_buffer` | 8/8 |
  | `test_agcom` | 10/10 |
  | `test_agentd` | 15/15 |
- [x] Reproduced the silently-half-closed scenario against this patched agent (`SIGSTOP` on the manager's `remoted`) and confirmed it reconnects instead of freezing. Full setup and timeline:

  **Setup:** official `wazuh-manager` 4.14.7 installed natively (protocol-compatible within the 4.14.x line; 4.14.9 itself isn't published yet). `wazuh-agentd` compiled from this branch (with the fix) running in a container on `--network host`, enrolled and connected against the manager over loopback. Golden path confirmed first: `status='connected'` in `wazuh-agentd.state`, manager shows the agent `Active` (`agent_control -l`), keepalives/events flowing normally.

  **Timeline:**

  | Time | Event |
  | --- | --- |
  | T+0s | `kill -STOP` on `wazuh-remoted` (simulates a silently half-closed TCP connection: socket stays "up", peer stops reading/responding) |
  | T+~63s | Agent's watchdog logs `Server unavailable. Setting lock.` (`max_time_reconnect_try`=60s) and enters the reconnect loop — **before this fix, this never happened: the agent stayed `status='connected'` forever** |
  | during outage | `status='disconnected'`; agent CPU stayed at ~1% (paced retries, no busy-loop); each failed handshake attempt bounded to ~60s by the new `OS_SetRecvTimeout()`, then retried — never hung |
  | ~8 min later | `kill -CONT` on `wazuh-remoted` |
  | same second | Agent logs `Connected to the server` / `Server responded. Releasing lock.`; `wazuh-agentd.state` back to `status='connected'`; manager shows the agent `Active` again |

  Also observed, unprompted: two transient `Corrupt payload (exceeding size) received` / `Lost connection with manager` blips immediately after the manager resumed (framing desync from bytes buffered during the freeze) — both self-healed with a reconnect in the same second, rather than leaving the connection stuck in a corrupted state.

  No CPU spin, no growth in RSS or open file descriptors observed across the full stall/reconnect cycle.

### Artifacts Affected

- `wazuh-agentd` binary (Linux/Windows/macOS).
- `etc/internal_options.conf` — new keys, see below.

### Configuration Changes

New `internal_options.conf` keys (not exposed in `ossec.conf`, internal tuning only):
- `agent.tcp_keepidle=60`, `agent.tcp_keepintvl=15`, `agent.tcp_keepcnt=4` — TCP keepalive timing for the agent → manager connection.
- `agent.send_timeout=30` — bounds how long `send()` may block before the agent closes and reconnects.

No backward-compatibility concerns: these are new internal options with defaults baked into the binary (`getDefine_Int`), so an agent upgraded without a matching `internal_options.conf` still starts (the shipped `internal_options.conf` is replaced on package upgrade, consistent with how `remoted`'s equivalent keys are already handled).

### Tests Introduced

Ported from #36790 (`main`'s `test_sendmsg.c`/`test_start_agent.c`), adapted to this branch's pre-refactor code (the TCP-path-only classification in `send_msg()`, the still-present UDP branch, and the original file layout):

- `test_sendmsg.c` (14 cases): `CreateSecMSG` failure, socket-already-invalidated early return (now also asserting the log + `sleep(1)` backoff), a clean success path, each of the 8 invalidating cases (`EPIPE`/`ECONNRESET`/`ENOTCONN`/`ECONNREFUSED`/`ETIMEDOUT`/`EAGAIN`/`EBADF`/partial-write-with-`errno==0`) individually asserting the socket is closed and set to `-1`, an unrelated errno (`ENOMEM`) asserting the socket is **not** touched, and 2 UDP-path smoke tests confirming that transport is unaffected by this fix.
- `test_start_agent.c`: added `test_connect_server_keepalive_fails` and `test_connect_server_send_timeout_fails` (connection still succeeds, failure only warns), updated the 3 existing TCP-success paths (`test_connect_server`, `test_agent_handshake_to_server` x2) to expect the new `OS_SetKeepalive`/`OS_SetKeepalive_Options`/`OS_SetSendTimeout`/`OS_SetRecvTimeout` calls, and `setup_test()` now calls `sender_init()` for real instead of relying on a zero-initialized `pthread_mutex_t` behaving like `PTHREAD_MUTEX_INITIALIZER` (a glibc-specific assumption, not portable C).
- Supporting wrapper additions: `__wrap_OS_SetKeepalive`/`__wrap_OS_SetKeepalive_Options` in `os_net_wrappers.c/h` (`OS_SetSendTimeout` was already wrapped, reused from `wazuh_db`/`analysisd`), and `#include "sec.h"` in `msgs_wrappers.h` (needed once it's included before `agentd.h` in a translation unit — same fix `main` already carries).

## Review Checklist

- [ ] Code changes reviewed
- [x] Port unit tests from #36790 (`test_sendmsg.c`, `test_start_agent.c`)
- [x] Reproduce the original stuck-connection scenario against this patched build

### DIFF
--- a/src/client-agent/receiver.c
+++ b/src/client-agent/receiver.c
@@ -37,6 +37,7 @@ int receive_msg()
     char buffer[OS_MAXSTR + 1];
     char cleartext[OS_MAXSTR + 1];
     char *tmp_msg;
+    const int sock = atomic_int_get(&agt->sock);
 
     memset(cleartext, '\0', OS_MAXSTR + 1);
     memset(buffer, '\0', OS_MAXSTR + 1);
@@ -49,7 +50,7 @@ int receive_msg()
                 break;
             }
 
-            recv_b = OS_RecvSecureTCP(agt->sock, buffer, OS_MAXSTR);
+            recv_b = OS_RecvSecureTCP(sock, buffer, OS_MAXSTR);
 
             // Manager disconnected or error
 
@@ -79,7 +80,7 @@ int receive_msg()
                 return -1;
             }
         } else {
-            recv_b = recv(agt->sock, buffer, OS_MAXSTR, MSG_DONTWAIT);
+            recv_b = recv(sock, buffer, OS_MAXSTR, MSG_DONTWAIT);
 
             if (recv_b <= 0) {
                 break;
@@ -123,14 +124,7 @@ int receive_msg()
             else if (strncmp(tmp_msg, HC_FORCE_RECONNECT, strlen(HC_FORCE_RECONNECT)) == 0) {
                 /* Set lock and wait for it */
                 minfo("Wazuh Agent will be reconnected because a reconnect message was received");
-                os_setwait();
-                w_agentd_state_update(UPDATE_STATUS, (void *) GA_STATUS_NACTIVE);
-
-                /* Send sync message */
-                start_agent(0);
-
-                os_delwait();
-                w_agentd_state_update(UPDATE_STATUS, (void *) GA_STATUS_ACTIVE);
+                reconnect_to_server();
                 continue;
             }
 
@@ -349,23 +343,47 @@ int receiver_messages()
             ExecdTimeoutRun();
         }
 
+#ifndef ONEWAY_ENABLED
         /* sock must be set */
-        if (agt->sock == -1) {
-            sleep(5);
+        int sock = atomic_int_get(&agt->sock);
+        if (sock == -1) {
+            merror(LOST_ERROR);
+            reconnect_to_server();
+            minfo(SERVER_UP);
             continue;
         }
+#else
+        /* start_agent() never actually connects under ONEWAY_ENABLED (see
+         * start_agent.c), so agt->sock never leaves -1 -- checking it here
+         * would spin this loop at 100% CPU instead of being paced by
+         * select()'s timeout below, same reasoning as the Linux monitor
+         * loop in agentd.c. */
+        int sock = atomic_int_get(&agt->sock);
+#endif
 
         run_notify();
 
+#ifndef ONEWAY_ENABLED
+        /* run_notify() may also invalidate the socket (it calls send_msg()
+         * for the keep-alive); re-check before FD_SET() needs it, same as
+         * the Linux monitor loop in agentd.c. */
+        sock = atomic_int_get(&agt->sock);
+        if (sock < 0) {
+            continue;
+        }
+#else
+        sock = atomic_int_get(&agt->sock);
+#endif
+
         FD_ZERO(&fdset);
-        FD_SET(agt->sock, &fdset);
+        FD_SET(sock, &fdset);
 
         /* Wait for 1 second */
         selecttime.tv_sec = 1;
         selecttime.tv_usec = 0;
 
         /* Wait with a timeout for any descriptor */
-        rc = select(agt->sock + 1, &fdset, NULL, NULL, &selecttime);
+        rc = select(sock + 1, &fdset, NULL, NULL, &selecttime);
         if (rc == -1) {
             merror(SELECT_ERROR, WSAGetLastError(), win_strerror(WSAGetLastError()));
             sleep(30);
@@ -375,13 +393,9 @@ int receiver_messages()
         }
 
         if (receive_msg() < 0) {
-            w_agentd_state_update(UPDATE_STATUS, (void *) GA_STATUS_NACTIVE);
             merror(LOST_ERROR);
-            os_setwait();
-            start_agent(0);
+            reconnect_to_server();
             minfo(SERVER_UP);
-            os_delwait();
-            w_agentd_state_update(UPDATE_STATUS, (void *) GA_STATUS_ACTIVE);
         }
     }
 
--- a/src/client-agent/sendmsg.c
+++ b/src/client-agent/sendmsg.c
@@ -19,6 +19,14 @@ void sender_init() {
     w_mutex_init(&send_mutex, NULL);
 }
 
+void send_mutex_lock(void) {
+    w_mutex_lock(&send_mutex);
+}
+
+void send_mutex_unlock(void) {
+    w_mutex_unlock(&send_mutex);
+}
+
 /* Send a message to the server */
 int send_msg(const char *msg, ssize_t msg_length)
 {
@@ -35,30 +43,73 @@ int send_msg(const char *msg, ssize_t msg_length)
 
     /* Send msg_size of crypt_msg */
     if (agt->server[agt->rip_id].protocol == IPPROTO_UDP) {
-        retval = OS_SendUDPbySize(agt->sock, msg_size, crypt_msg);
-#ifndef WIN32
+        retval = OS_SendUDPbySize(atomic_int_get(&agt->sock), msg_size, crypt_msg);
+#ifdef WIN32
+        error = WSAGetLastError();
+#else
         error = errno;
 #endif
     } else {
         w_mutex_lock(&send_mutex);
-        if (agt->sock < 0) {
+        /* Snapshot once: nothing else can change agt->sock while send_mutex is
+         * held (every writer takes it first too), so reuse this value for the
+         * rest of the critical section instead of re-reading the atomic. */
+        int sock = atomic_int_get(&agt->sock);
+        if (sock < 0) {
             /* Socket was already invalidated (e.g. by another sender thread
              * after a previous timed-out send). Skip the call to avoid
-             * operating on -1 and reading a stale errno. */
+             * operating on -1 and reading a stale errno. The sleep(1) matches
+             * every other failure path here: without it, a caller like
+             * dispatch_buffer() would spin through its whole queue at
+             * events_persec discarding messages silently instead of pausing
+             * while the reconnect (already under way) completes. */
             w_mutex_unlock(&send_mutex);
+            mdebug1("Connection is down, discarding message until it is restored.");
+            sleep(1);
             return (-1);
         }
-        retval = OS_SendSecureTCP(agt->sock, msg_size, crypt_msg);
-#ifndef WIN32
-        error = errno;
+        retval = OS_SendSecureTCP(sock, msg_size, crypt_msg);
         if (retval) {
             bool socket_dead;
+#ifdef WIN32
+            error = WSAGetLastError();
+
+            switch (error) {
+            case WSAECONNRESET:
+            case WSAECONNABORTED:
+            case WSAENOTCONN:
+            case WSAECONNREFUSED:
+            case WSAENOTSOCK:
+            case WSAETIMEDOUT:
+            case WSAEWOULDBLOCK:
+                /* Same reasoning as the POSIX branch below: a dead peer,
+                 * an invalidated descriptor, or an SO_SNDTIMEO expiry. */
+                socket_dead = true;
+                break;
+            case 0:
+                /* WSASetLastError(0) before send() in OS_SendSecureTCP()
+                 * wasn't overwritten: a short/partial write happened with
+                 * no reported error. The length-prefixed framing for this
+                 * connection is now corrupted; only a reconnect recovers. */
+                socket_dead = true;
+                break;
+            default:
+                socket_dead = false;
+                break;
+            }
+#else
+            error = errno;
 
             switch (error) {
             case EPIPE:
             case ECONNRESET:
             case ENOTCONN:
             case ECONNREFUSED:
+            case EBADF:
+                /* agt->sock points at a descriptor that is no longer open
+                 * (e.g. closed by another thread without updating agt->sock).
+                 * Without this, the agent would keep retrying send() on the
+                 * same dead descriptor forever instead of reconnecting. */
                 socket_dead = true;
                 break;
             case ETIMEDOUT:
@@ -71,16 +122,25 @@ int send_msg(const char *msg, ssize_t msg_length)
                  * blocking again on the very next call. */
                 socket_dead = true;
                 break;
+            case 0:
+                /* errno = 0 right before send() in OS_SendSecureTCP() wasn't
+                 * overwritten: a short/partial write happened with no error
+                 * at all (SO_SNDTIMEO can expire after only part of a large
+                 * message drained). The length-prefixed framing for this
+                 * connection is now corrupted regardless of how many bytes
+                 * got through; only a reconnect recovers. */
+                socket_dead = true;
+                break;
             default:
                 socket_dead = false;
                 break;
             }
-            if (socket_dead && agt->sock >= 0) {
-                OS_CloseSocket(agt->sock);
-                agt->sock = -1;
+#endif
+            if (socket_dead) {
+                OS_CloseSocket(sock);
+                atomic_int_set(&agt->sock, -1);
             }
         }
-#endif
         w_mutex_unlock(&send_mutex);
     }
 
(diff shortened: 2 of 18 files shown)


src/client-agent/sendmsg.c
```diff
@@ -41,9 +49,50 @@ int send_msg(const char *msg, ssize_t msg_length)
...
+#if defined(EWOULDBLOCK) && (EWOULDBLOCK != EAGAIN)
+            case EWOULDBLOCK:
+#endif
+                /* SO_SNDTIMEO expired: the kernel send buffer never drained.
+                 * Close and invalidate so the main loop reconnects instead of
+                 * blocking again on the very next call. */
+                socket_dead = true;
+                break;
+            default:
+                socket_dead = false;
```
lchico, line 135: A partial `send()` that times out via the new `SO_SNDTIMEO` returns a short byte count rather than -1, so `errno` stays at 0 and the `socket_dead` switch below falls into `default: socket_dead = false`. That leaves a genuinely stuck connection (partial write, then stalled peer) marked alive, and the agent keeps reusing a socket whose length-prefixed framing is now corrupted for every subsequent message. Should a short write from `OS_SendSecureTCP()` (return value != bufsz, errno == 0) also be classified as `socket_dead = true` here, not just the explicit errno cases?

src/client-agent/sendmsg.c
```diff
@@ -41,9 +49,50 @@ int send_msg(const char *msg, ssize_t msg_length)
...
+             * after a previous timed-out send). Skip the call to avoid
+             * operating on -1 and reading a stale errno. */
+            w_mutex_unlock(&send_mutex);
+            return (-1);
+        }
         retval = OS_SendSecureTCP(agt->sock, msg_size, crypt_msg);
 #ifndef WIN32
         error = errno;
+        if (retval) {
+            bool socket_dead;
```
lchico, line 73: This whole dead-socket detection/close block is `#ifndef WIN32`, but the `OS_SetSendTimeout()`/keepalive setup in `start_agent.c` that produces these timeouts is applied unconditionally on all platforms. On Windows, a send timeout still fires (so `send()` stops blocking forever) but `agt->sock` is never invalidated or closed, so the main loop's `agt->sock < 0` reconnect check never triggers — the Windows agent just repeats the same bounded timeout indefinitely without reconnecting. Is there a Windows-side equivalent of this detection path elsewhere, or does this PR intentionally leave Windows unfixed for #38388?

lchico (changes requested):

src/client-agent/sendmsg.c
cborla: Good catch. errno is 0 in exactly that case (OS_SendSecureTCP() zeroes it right before send(), and a short write doesn't touch it), so I added case 0: to both switches and treat it as fatal too, added test_send_msg_partial_write to cover it.

src/client-agent/receiver.c
```diff
@@ -350,22 +344,33 @@ int receiver_messages()
         }
 
         /* sock must be set */
-        if (agt->sock == -1) {
```
lchico, line 353: `agentd.c`'s monitor loop wraps its new immediate-reconnect block in `#ifndef ONEWAY_ENABLED`, with a comment explaining why: under `ONEWAY_ENABLED`, `start_agent()` returns immediately without connecting, so `agt->sock` never leaves `-1`, and an unthrottled immediate-retry would spin at 100% CPU. This PR replaces this Windows loop's old `if (agt->sock == -1) { sleep(5); continue; }` with the same shape of immediate-reconnect block (`merror(LOST_ERROR); reconnect_to_server(); minfo(SERVER_UP); continue;`), but without the matching `#ifndef ONEWAY_ENABLED` guard. Building the Windows agent with `ONEWAY=yes` would make this loop spin continuously with no sleep, since `reconnect_to_server()` returns instantly every time — burning a CPU core and flooding the log, where the old code just slept 5s. Should this block get the same guard as `agentd.c`'s?

lchico (changes requested):

src/client-agent/receiver.c
cborla: Good catch, yes — added the same #ifndef ONEWAY_ENABLED guard here (both this check and the post-run_notify() re-check below), mirroring agentd.c's monitor loop exactly. Verified with a  TARGET=winagent ONEWAY=yes cross-compile.

lchico (approved):
LGTM!

- [x]  Code changes reviewed
- [ ] ~Configuration changes documented~

PR #38410, opened 2026-08-17
Prevent the agent from stopping Active Response when too many commands are defined

### LINKED ISSUE #38358: Wazuh execd issue
# wazuh-execd segfaults (SIGSEGV) when total defined Active Response commands exceeds a threshold (~90)

## Wazuh version
- Manager: v4.14.1-rc2
- Agent: v4.14.1
- OS (agent): Ubuntu 22.04 / 24.04

## Description

`wazuh-execd` on the agent consistently crashes with a segmentation fault whenever the manager's `ossec.conf` defines a large total number of `<command>`/`<active-response>` blocks (Windows + Linux combined). The crash is reproducible and appears to be tied to the **total count of Active Response command definitions** loaded via `ar.conf`, not to which specific command is triggered.

This may be related to the buffer-overflow fix already merged in #33665 ("Improve Active Response key handling in wazuh-execd"), which addressed an unbounded `strcat()` into a fixed-size `rkey` buffer. Our crash signature is consistent with that class of bug, but appears to trigger from total AR definition count rather than only from runtime alert-derived keys. Filing this in case it's a related/adjacent overflow that isn't fully covered by that fix, or to confirm whether this is expected/known and what the safe upper bound is.

## Steps to Reproduce

1. On the manager, define a large number of `<command>`/`<active-response>` blocks in `ossec.conf` (our case: 47 legacy/Windows-oriented command definitions + a growing set of custom Linux command definitions).
2. Restart the manager and agent.
3. Trigger any Active Response command on the Linux agent (manually via `echo '{...}' | /var/ossec/active-response/bin/<script>.sh`, or via the dashboard).
4. Observe `wazuh-execd` crash on the agent.

## Evidence

**dmesg segfault (consistent across many crashes, same address every time):**
```
wazuh-execd[<pid>]: segfault at 6c109b ip <addr> sp <addr> error 6 in libc.so.6[...]
```

**GDB backtrace from core dump:**
```
#0  __printf_buffer_snprintf_init (length=257, buffer=0x6c109b <error: Cannot access memory at address 0x6c109b>, ...)
#1  __vsnprintf_internal (...)
#2  __GI___snprintf (...)
#3  0x0000000000408821 in ?? ()
#4  0x0000000000409825 in ?? ()
#5  0x0000000000409d01 in ?? ()
#6  0x0000000000407c7d in ?? ()
#7  __libc_start_call_main (main=main@entry=0x407a40, ...)
```
(Binary is stripped, no symbols available beyond libc frames — happy to provide the core dump / binary if useful.)

**Controlled A/B test (most direct evidence):**
- Total AR commands = 91 (47 legacy + 44 custom) → triggering *any* command reliably crashes `wazuh-execd`.
- Removed 2 unrelated commands (total = 89), re-added 2 previously-removed ones elsewhere → **no crash**, same commands trigger successfully.
- This was repeated multiple times; the crash consistently correlates with total command count, not with which specific command/script is executed.

**Ruled out:**
- Not an OOM/memory-pressure kill: `dmesg` shows `SIGSEGV` (signal 11) at a small non-null, non-random address (`0x6c109b`), not a null-pointer crash or OOM-killer `SIGKILL`.
- Not killed externally: `journalctl` shows no unexpected kill signals; `systemctl show wazuh-agent` shows `WatchdogUSec=0` (disabled) and neutral `OOMPolicy`/`OOMScoreAdjust`.

## Expected Behavior
`wazuh-execd` should not crash regardless of how many Active Response commands are defined, or Wazuh should document/enforce a safe upper limit on total AR command definitions.

## Actual Behavior
`wazuh-execd` segfaults deterministically once total AR command definitions cross a threshold we've empirically bounded between 89 (safe) and 91 (crashes) in our environment.

## Questions for maintainers
1. Is there a known/hardcoded limit on the number of Active Response `<command>` definitions supported per agent?
2. Is `ar.conf` truly sent in full to every connected agent regardless of group/OS (per current docs), or is there a supported way to scope AR command definitions to a subset of agents (e.g., by OS or group) so agents don't need to parse irrelevant definitions?
3. Is this the same root cause as #33665, or a distinct overflow in the config-loading/dedup-key-table path?

## Workaround we're currently using
Reducing total `<command>` definitions below the crash threshold (currently no clean way to scope this per-OS since `ar.conf` appears to be sent in full to all agents). Upgrading to 4.14.3 is planned but not immediately possible in our environment.

Happy to provide the manager's full `ossec.conf` (sanitized), agent's core dump, or run additional diagnostics if useful.

vikman90:
## Description

`wazuh-execd` loads the Active Response command table from `etc/shared/ar.conf` into fixed-size arrays
sized for `MAX_AR` (64) commands, but `ReadExecConfig()` never checked the entry counter against that
limit. When the manager distributed an `ar.conf` holding more entries than the table can store, the
function kept writing past the end of the table until the process reached unmapped memory and was
terminated by the operating system.

The table is loaded lazily, on the first Active Response the agent is asked to run, so the effect was that
Active Response stopped working on every affected agent as soon as any command was triggered, regardless
of which command it was. `ar.conf` is merged into the shared configuration of every group, so all agents
receive the full list of definitions.

Closes #38358.

## Proposed Changes

- `src/os_execd/exec.c`: stop reading `ar.conf` once `MAX_AR` commands have been loaded, reporting the
  ignored entries with the new message `(1319)`, so the truncation is visible in `ossec.log` instead of
  being silent.
- `src/os_execd/exec.c`: reset the whole command table on reload instead of one row past the last one
  used.
- `src/error_messages/error_messages.h`: add the `EXEC_MAX_AR` message.

Commands defined within the limit keep being loaded and executed exactly as before. Commands beyond the
limit are now reported as invalid, which is the same behaviour as any other unknown command name.

### Results and Evidence

**Before the change** — agent v4.14.7 with an `ar.conf` holding 120 entries. Triggering the first command
terminates the daemon:

```
# wazuh-execd is running, then an Active Response is triggered
=== process alive? ===
wazuh-execd IS GONE
```

```
wazuh-execd[55119]: segfault at 6c209b ip 0000763c1788fd94 sp 00007fff2eace070 error 6 in libc.so.6[...]
wazuh-execd: wazuh-execd: potentially unexpected fatal signal 11.
```

The same condition reproduced under AddressSanitizer through the unit tests, with the fix reverted:

```
ERROR: AddressSanitizer: global-buffer-overflow on address 0x59cf60d2ea01
WRITE of size 257 at 0x59cf60d2ea01 thread T0
    #1 ReadExecConfig os_execd/exec.c:35
0x59cf60d2ea01 is located 0 bytes after global variable 'exec_names'
defined in 'os_execd/exec.c:15:13' (0x59cf60d2a8c0) of size 16705
```

**After the change** — same agent updated to this branch (v4.14.9), same 120-entry `ar.conf`. The daemon
survives and reports the ignored entries:

```
2026/08/17 09:28:39 wazuh-execd: INFO: Started (pid: 55853).
2026/08/17 09:28:41 wazuh-execd: ERROR: (1319): Reached the maximum of 64 active response commands. Ignoring the remaining entries of 'etc/shared/ar.conf'.

=== process alive? ===
55853 /var/ossec/bin/wazuh-execd
```

Triggering the last command within the limit (entry 64) runs normally, and triggering the first one beyond
it (entry 65) is reported as an unknown command instead of stopping the daemon:

```
2026/08/17 09:29:00 wazuh-execd: ERROR: (1311): Invalid command name 'test-ar-0640' provided.
```

Full unit test suite for the agent target:

```
100% tests passed, 0 tests failed out of 84
```

### Artifacts Affected

- `wazuh-execd` (all agent packages, and the manager package since the same daemon runs there).

### Configuration Changes

None. No new setting is introduced and no existing one changes its meaning. The supported number of
Active Response command definitions is unchanged at 64; it is now enforced and reported instead of being
left undefined.

### Documentation Updates

The documentation does not currently state a maximum number of Active Response command definitions.
Documenting the 64-command limit, and the fact that `ar.conf` is distributed in full to every agent, would
be worth a follow-up documentation issue.

### Tests Introduced

New test binary `src/unit_tests/os_execd/test_read_exec_config.c` covering `ReadExecConfig()` against a
real `ar.conf`:

- Below the limit: every command is loaded and resolvable.
- Exactly `MAX_AR` entries: every command is loaded, nothing is reported as ignored.
- Above the limit: the commands within the limit are loaded, the message `(1319)` is reported, and the
  remaining entries are discarded.

The third case is what reproduces the failure: with the change reverted it aborts under AddressSanitizer,
as shown above.

## Credit

Thanks to @nimra1923 for reporting this issue.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 #### Fixed
 
 - Fixed the Windows agent MSI upgrade leaving the agent broken after the next reboot, and the silent `/q` upgrade hanging, when a system restart was pending. ([#38277](https://github.com/wazuh/wazuh/pull/38277))
+- Fixed `wazuh-execd` crashing when more active response commands than supported are defined. ([#38358](https://github.com/wazuh/wazuh/issues/38358))
 
 ## [v4.14.8]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file.
 #### Fixed
 
 - Fixed the Windows agent MSI upgrade leaving the agent broken after the next reboot, and the silent `/q` upgrade hanging, when a system restart was pending. ([#38277](https://github.com/wazuh/wazuh/pull/38277))
-- Fixed `wazuh-execd` crashing when more active response commands than supported are defined. ([#38358](https://github.com/wazuh/wazuh/issues/38358))
+- Fixed `wazuh-execd` crashing when more active response commands than supported are defined. ([#38410](https://github.com/wazuh/wazuh/pull/38410))
 
 ## [v4.14.8]
 
--- a/src/analysisd/active-response.c
+++ b/src/analysisd/active-response.c
@@ -16,6 +16,13 @@
 #include <grp.h>
 #endif
 
+/* The lines that AR_ReadConfig() always writes hold two distinct names, and every
+ * platform ships at most one of the executables of each, so the agents keep two
+ * entries out of them. Should a platform ever ship two executables of the same
+ * name, that name would take one further entry.
+ */
+#define AR_DEFAULT_COMMANDS 2
+
 /* Active response commands */
 OSList *ar_commands;
 OSList *active_responses;
@@ -75,6 +82,13 @@ int AR_ReadConfig(const char *cfgfile)
     /* Read configuration */
     if (ReadConfig(modules, cfgfile, ar_commands, active_responses) < 0) {
         return (OS_INVALID);
+    }
+
+    /* The entries above plus one per active response make up the shared file */
+    const int total_commands = AR_DEFAULT_COMMANDS + active_responses->currently_size;
+
+    if (total_commands > MAX_AR) {
+        mwarn(AR_MAX_COMMANDS, total_commands, MAX_AR, DEFAULTAR);
     }
 
     return (0);
--- a/src/error_messages/error_messages.h
+++ b/src/error_messages/error_messages.h
@@ -198,6 +198,7 @@
 #define EXEC_INV_CMD    "(1316): Invalid AR command: '%s'"
 #define EXEC_CMD_FAIL   "(1317): Could not launch command %s (%d)"
 #define EXEC_BAD_NAME   "(1318): Command name truncated %s"
+#define EXEC_MAX_AR     "(1319): Reached the maximum of %d active response commands. Ignoring the remaining entries of '%s'."
 
 #define AR_NOAGENT_ERROR                "(1320): Agent '%s' not found."
 #define EXEC_QUEUE_CONNECTION_ERROR     "(1321): Error communicating with queue '%s'."
--- a/src/error_messages/warning_messages.h
+++ b/src/error_messages/warning_messages.h
@@ -13,6 +13,7 @@
 
 /* Active Response */
 #define AR_SERVER_AGENT "(1306): Invalid agent ID. Use location=server to run AR on the manager."
+#define AR_MAX_COMMANDS "(1307): Defined %d active response commands, more than the maximum of %d. The exceeding entries of '%s' may not be loaded by the agents."
 
 /* File integrity monitoring warning messages*/
 #define FIM_WARN_ACCESS                         "(6900): Accessing  '%s': [(%d) - (%s)]"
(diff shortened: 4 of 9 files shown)


MarcelKemp (dismissed):
LGTM

MarcelKemp (approved):

Issue #37189, opened 2026-06-25
Restore FIM whodata integration-test coverage dropped during the CodeBuild migration


Nicogp:
## Description

As part of migrating the agent integration tests to AWS CodeBuild runners ([PR #37051](https://github.com/wazuh/wazuh/pull/37051),
part of [#37027](https://github.com/wazuh/wazuh/issues/37027)), the **FIM whodata** integration tests were **excluded on both Linux and
Windows** because the CodeBuild runners cannot provide the OS-level auditing whodata depends
on. This issue tracks restoring that coverage and deciding the long-term approach.

The exclusions are environment workarounds, **not** test or product bugs: the FIM whodata
engine works correctly on a real OS; it just cannot be exercised inside the current CodeBuild
runners.

## What was skipped

**Linux 5_** — `.github/test_modules_linux.json` (`fim` module):
```
pytest_extra_args: -k "not whodata and not Who-data"
                   --ignore=test_fim/test_files/test_follow_symbolic_link/test_audit_rules_with_symlink.py
```

**Windows 5_** — `.github/test_modules_windows.json` (`fim` module):
```
pytest_extra_args: -k "not whodata and not Who-data"
```

### 4.x branch (`4.14.7`) — per-workflow inline `pytest` (not `test_modules_*.json`)

Unlike the 5.x path above, the 4.x agent workflows invoke `pytest` inline, so the whodata
exclusion lives in each workflow file:

**Linux**
- `.github/workflows/4_testintegration_fim-tier-0-1-lin.yml` — filtered during the CodeBuild
  migration (#37027):
sudo python -m pytest --tier 0 --tier 1 test_fim/
  -k "not whodata and not Who-data"
  --ignore=test_fim/test_files/test_follow_symbolic_link/test_audit_rules_with_symlink.py
- `.github/workflows/4_testintegration_fim-tier-2-lin.yml` — added via #37273 (the tier-2 job
  was missed during the migration):
sudo python -m pytest --tier 2 test_fim/
  -k "not whodata and not Who-data"

**Windows** — added via #37193 (4.x Windows runner migration to CodeBuild):
- `.github/workflows/4_testintegration_fim-tier-0-1-win.yml`:
python -m pytest -k "not whodata and not Who-data" --tier 0 --tier 1 test_fim\
- `.github/workflows/4_testintegration_fim-tier-2-win.yml`:
python -m pytest -k "not whodata and not Who-data" --tier 2 test_fim\

[...]

PR #35060, opened 2026-03-19
Linux  test integration workflow improvements 

rovogel:
## Description

This PR improves the Linux integration test CI infrastructure to address several problems with the previous approach:

- Per-module integration test workflows were completely standalone and had to be triggered manually or via `workflow_dispatch`, meaning their results were **not linked to the PR** as native checks.
- Each per-module workflow duplicated the same agent installation and test execution logic (~90 lines each), making them hard to maintain.
- There was no automatic selective triggering based on which files were actually changed in a PR.

Closes #34915

### New generic reusable workflow (`5_testintegration_linux-integration-tests.yml`)

A single `workflow_call`-based workflow replaces the 8 individual per-module test workflows. It accepts inputs for `module_name`, `pytest_target`, `pytest_tier_args`, `pre_install`, `timeout_minutes`, and `build_run_id`, making it fully parameterizable for any Linux integration test module. Responsibilities:

- Downloads the DEB agent package artifact from the triggering (or specified) build run.
- Installs the agent and applies a standard ossec.conf cleanup via `awk`:
  - Disables `<rootcheck>` and `<syscheck>` blocks.
  - Removes `<sca>` and `<wodle name="syscollector">` blocks (unless the test target is `test_fim/`, in which case `<sca>` is preserved).
  - Adds the required `<localfile>` blocks (`active-responses.log`, `dpkg.log`, `journald`).
- Resolves the `qa-integration-framework` branch from the PR branch, base branch, version tag, or `main`, in that order.
- Runs `pytest` with the provided tier/target/extra arguments.
- Dumps `wazuh-agent` logs and uploads a test results HTML artifact on failure.

### New change-detection job in the build workflow (`5_builderpackage_agent-linux.yml`)

- A new `detect-changes` job reads `.github/test_modules_linux.json` and computes a `strategy.matrix` from the set of changed files, matching them against per-module path patterns.
- A new `run-integration-tests` job uses `workflow_call` via `uses: ./.github/workflows/5_testintegration_linux-integration-tests.yml` inside a matrix, one entry per matched module+tier. Because it uses `workflow_call`, each matrix leg appears as a **native PR check**.
- Both jobs only run on `pull_request` events. `run-integration-tests` additionally waits for `smoke-test-linux` to succeed.

### New module registry (`.github/test_modules_linux.json`)

[...]

PR #38425, opened 2026-08-18
Resolve the audited path for single-item whodata events and restore the FIM whodata integration tests

### LINKED ISSUE #35481: FIM whodata test failures on audit 3.1.2: modify/delete events via symlink path not generated
## Description

During the execution of `test_follow_symbolic_link/test_cud.py` on systems with `auditd 3.1.2`, the `EVENT_TYPE_MODIFIED` and `EVENT_TYPE_DELETED` assertions fail. The same test passes on systems running `auditd 3.0.7`.

## Background

FIM in whodata mode registers audit rules on the **resolved (real) path** of a monitored symbolic link:

```
-w /path/to/target -p wa -k wazuh_fim
```

When file operations are performed through the symlink path (e.g. `symlink/testie.txt`), auditd is expected to generate PATH records that syscheckd can correlate back to the monitored rule.

## Observed behavior

| Audit version | CREATE via symlink | MODIFY via symlink | DELETE via symlink |
|---|---|---|---|
| 3.0.7 | ✅ event generated | ✅ event generated | ✅ event generated |
| 3.1.2 | ✅ event generated | ❌ no event / not matched | ❌ no event / not matched |

## Hypothesis

A change in how auditd 3.1.2 constructs PATH records for `open(2)` on existing files may be causing the symlink component to be preserved in the reported path (e.g. `symlink/testie.txt`) instead of being resolved to the canonical inode path (`target/testie.txt`). This would prevent syscheckd from matching the event against the registered rule.

## Tasks

- [ ] Evaluate whether syscheckd should register rules on **both** the symlink path and the resolved path, or resolve events using `realpath` before matching
- [ ] Evaluate whether `follow_symbolic_link` configuration in FIM needs to account for this audit version difference
- [ ] Update or document the minimum supported auditd version if needed

## Environment

- **Failing:** Ubuntu with `auditd 3.1.2`
- **Passing:** Ubuntu with `auditd 3.0.7`
- **FIM mode:** `whodata`
- **Affected test:** `tests/integration/test_fim/test_files/test_follow_symbolic_link/test_cud.py`

Miguevrgo:
## Description

FIM whodata missed **modify** events for files reached through a monitored symbolic link, while **create** events on the same file were reported correctly. This PR fixes that and restores the integration coverage that was hiding it.

Closes #35481
Related to #37189

### Root cause

`audit_parse()` branches on how many `PATH` records the kernel emits for the syscall (`src/syscheckd/src/whodata/audit_parse.c:757`):

- **Creating** a file emits two records, the parent directory and the new entry, so it takes the `items=2` branch, which resolves the generated path before building the event:
  ```c
  w_evt->path = realpath(file_path, NULL);
  ```
- **Writing to a file that already exists** emits a single record, so it takes the `items=1` branch, which assigned the generated path verbatim:
  ```c
  w_evt->path = file_path;      // no resolution
  ```

When the operation goes through a monitored symlink the kernel reports the **symlink** path, identically on `auditd 3.0.7` / kernel 5.15 and `auditd 3.1.2` / kernel 6.8:

```
type=SYSCALL syscall=257 items=2
type=PATH item=0 name="/tmp/symlink/" nametype=PARENT
type=PATH item=1 name="/tmp/symlink/testie.txt" nametype=CREATE
```

FIM stores its entries under the **resolved** target, so without that resolution the `items=1` path never matched the database and no modify event was produced. The create path resolved correctly and worked, which is exactly the reported "create yes, modify no" symptom.

This was captured directly on the failing integration case. Note the two different source lines, i.e. two different branches of the same switch:

```
audit_parse.c:779  (6247): audit_event: ... path=.../symlink/testie.txt   ->  Sending FIM event: "type":"added"
audit_parse.c:763  (6247): audit_event: ... path=.../symlink/testie.txt   ->  (no event)
```

For the record, the `auditd 3.1.2` hypothesis in the issue description does not hold: the audit records are byte-identical between `3.0.7` on kernel 5.15 and `3.1.2` on kernel 6.8. The version was never the trigger; the syscall shape is.

### Why CI never caught it

The failing scenario was suppressed on Linux in two independent places:

1. **The test cases did not configure whodata.** Commit 0ee02b5 (PR #35060, merged 2026-04-17) changed the two `Who-data` cases of `test_follow_symbolic_link/test_cud.py`:
   ```diff
        FIM_MODE:
   -      whodata: 'yes'
   +      realtime: 'yes'
   ```
   with the message *"test: change FIM mode; this needs to be analyzed in a separate issue #35481."* Since then they carried a whodata label while exercising realtime.

2. **The CI filter excluded whodata from the whole `fim` module.** Added in 6c65785 (PR #37051) during the AWS CodeBuild migration, because those runner containers lacked `CAP_AUDIT_CONTROL` and `CAP_AUDIT_READ` (#37189, wazuh-automation#3497). The Linux integration tests are back on `runs-on: ubuntu-24.04`, a full VM with a working kernel audit subsystem, so the exclusion outlived its reason.

Fixing only the first would have changed nothing, since the `-k` filter deselects those cases by name regardless of how they are configured.

## Proposed Changes

- `src/syscheckd/src/whodata/audit_parse.c`: apply in the `items=1` branch the same `realpath()` normalization the `items=2` branch already performs, keeping the raw path as a fallback when `realpath()` fails, as it legitimately does for an entry that no longer exists.
- `src/unit_tests/syscheckd/whodata/test_audit_parse.c`: add the `realpath` expectations to `test_audit_parse_chmod`, the single-record case.
- `tests/integration/test_fim/.../cases_cud.yaml`: restore `whodata: 'yes'` in the two `Who-data` cases so their configuration matches their label.
- `.github/test_modules_linux.json`: drop `pytest_extra_args` from the `fim` module, removing the `-k "not whodata and not Who-data"` filter and the `--ignore` of `test_audit_rules_with_symlink.py`.
- `CHANGELOG.md`: one entry for the fix, one for the restored coverage.

### Results and Evidence

**The integration case, before and after the fix.** Same branch, same runner, only `audit_parse.c` differs:

| | `test_cud.py` `Who-data` cases |
|---|---|
| before the fix | `2 failed, 4 deselected` — both fail at `test_cud.py:114`, the `EVENT_TYPE_MODIFIED` assertion, with the create assertion at `:108` passing |
| after the fix | `2 passed, 4 deselected` |

Failure detail from the run before the fix:

```
=================================== FAILURES ===================================
test_fim/test_files/test_follow_symbolic_link/test_cud.py:114: in test_cud
    assert wazuh_log_monitor.callback_result
E   assert None
```

**Full tier-0 run with whodata restored, before the fix**: `2 failed, 107 passed` — the only two failures were these cases, so the rest of the whodata coverage that this PR brings back was already healthy.

**Coverage restored.** Collection over `test_fim/` at tier 0/1:

| | tests collected |
|---|---|
| with the filter (current `5.0.0`) | 131 / 521 |
| without the filter (this PR) | **197 / 527** |

66 cases return to the job: 60 previously deselected by `-k`, 6 previously removed from collection by `--ignore`.

**Unit tests.** Full legacy suite on the agent target, with the fix applied:

```
100% tests passed, 0 tests failed out of 102
```

including the directly affected suites:

```
Test #46: test_run_realtime ................   Passed
Test #49: test_run_check ...................   Passed
Test #56: test_audit_rule_handling .........   Passed
Test #57: test_syscheck_audit ..............   Passed
Test #58: test_audit_parse .................   Passed
```

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

### Artifacts Affected

`wazuh-agent` binary (Linux, `wazuh-syscheckd` — the changed code is under `#ifndef WIN32`).

### Configuration Changes

N/A

### Tests Introduced

No new test files. Existing coverage is restored and one unit test is extended:

- `test_audit_parse_chmod` now asserts the resolution of the single-PATH-record path.
- The two `Who-data` cases of `test_cud.py` exercise whodata again instead of realtime, and are the regression test for this fix.
- 66 tier 0/1 cases return to the `fim` job.

> [!NOTE]
> Changing `.github/test_modules_linux.json` matches the `paths` filter of every module, so this PR triggers the full `IT Linux` matrix. `IT Linux - logcollector` fails on `test_basic_configuration_target.py::test_configuration_target[Basic configuration target0]`, which is unrelated to FIM and is not a regression from this change.
>
> The same module was run in isolation against unmodified `5.0.0` to get a baseline (package build `32227956517`, test run `32228622325`, identical inputs to the matrix job) and it passed: `73 passed, 11 deselected, in 588.46s`. On this PR the same suite reports `1 failed, 72 passed` in a comparable 573s, with the module running alongside the rest of the matrix, including the 56-minute `aws` and 72-minute `fim` jobs.
>
> So the difference is the execution context, not the code: nothing in this PR touches logcollector. The assertion it fails on has a 10-second timeout, and the message it waits for is emitted inside `LogCollectorStart()`, which runs after `startup_gate_wait_for_ready()` (`src/logcollector/src/main.c:195`). That makes a tight-timeout flake under load a plausible explanation, though it is not confirmed and is out of scope here; it deserves its own issue.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...


Darioortegaleyva (changes requested):
Good overall, it fixes what the issue reports, just one think.

Have you checked whether the same fix is needed in `case 3/4/5` of the same switch (unlink/rename, items=3/4/5) in `audit_parse.c`? They build `w_evt->path` from `gen_audit_path()` without the `realpath()` resolution added here, so the same symlink path mismatch could still affect delete/rename events.

Miguevrgo:
Good catch, and it is not only theoretical: `case 3/4/5` are reachable through a monitored symlink and none of them resolved the path. Measured it on Ubuntu 24.04 / auditd 3.1.2 / kernel 6.8, with the watch registered on the resolved target and every operation performed through the symlink, exactly as FIM sets it up:

| Operation | Syscall | `items` | Branch | Resolved before this PR? |
|---|---|---|---|---|
| delete file (`rm symlink/f.txt`) | `unlinkat` | 2 | `case 2` | yes |
| delete directory (`rm -rf symlink/sub`) | `unlinkat` | 2 | `case 2` | yes |
| rename inside (`mv symlink/a symlink/b`) | `renameat2` | **4** | `case 4` | **no** |
| rename out (`mv symlink/c /tmp/outside/`) | `renameat2` | **4** | `case 4` | **no** |
| rename in (`mv /tmp/outside/d symlink/`) | `renameat2` | **4** | `case 4` | **no** |
| rename over existing (`mv symlink/e1 symlink/e2`) | `renameat` | **5** | `case 5` | **no** |
| hard link (`ln symlink/h1 symlink/h2`) | `linkat` | **3** | `case 3` | **no** |

In all seven the `PATH` records carry the symlink form, e.g. for the rename:

```
SYSCALL=renameat2 syscall=316 items=4
    item=0   name="/tmp/symlink/"        nametype=PARENT
    item=1   name="/tmp/symlink/"        nametype=PARENT
    item=2   name="/tmp/symlink/a.txt"   nametype=DELETE
    item=3   name="/tmp/symlink/b.txt"   nametype=CREATE
```

So delete was already fine, because `unlinkat` lands on `items=2`, which is why the issue only ever reported the update as missing. Rename and hard link were not: same mismatch against the resolved path FIM stores, just never exercised by the restored test.

Applied the same normalization to `case 3`, `case 4` and `case 5`. `case 4` emits two events, so it resolves each generated path separately and releases the first one before building the second, since `w_evt->path` is owned by `free_whodata_event()`.

Running the unit suites now and will report back with the results before marking this ready again.

Miguevrgo:
Followed this through and I have to walk back part of my previous answer: the concern is real, but applying the same `realpath()` to `case 3/4/5` is **not** the right fix, and I reverted it (0372bc1).

Pushing it turned `IT Linux - fim (tier-0)` red, with two previously green cases failing:

```
FAILED test_audit_rules_with_symlink.py::test_audit_rules_with_symlink[Audit rule is deleted when symlink is moved inside the monitored folder - Who-data]
FAILED test_audit_rules_with_symlink.py::test_audit_rules_with_symlink[Audit rule is deleted when symlink is moved outside the monitored folder - Who-data]
```

Same 109 selected tests before and after (`collected 527 items / 418 deselected / 109 selected` in both runs), and the `5.0.0` merge in between touched nothing under `tests/integration/test_fim/`, so the regression is mine.

Root cause of the regression, measured. Those cases repoint a symlink through `file.modify_symlink_target()`, which on Linux runs `ln -sfn`, and that is an atomic swap, not a plain `symlink()`:

```
SYSCALL=symlinkat items=3
    item=0   name="/test_dir/"            nametype=PARENT
    item=1   name="/test_dir/test_path"   nametype=UNKNOWN
    item=2   name="/test_dir/Cuh42qE7"    nametype=CREATE     <- temp link, case 3
SYSCALL=renameat  items=5
    item=0   name="/test_dir/"            nametype=PARENT
    item=1   name="/test_dir/"            nametype=PARENT
    item=2   name="/test_dir/Cuh42qE7"    nametype=DELETE
    item=3   name="/test_dir/symlink"     nametype=DELETE     <- case 5
    item=4   name="/test_dir/symlink"     nametype=CREATE
```

So it lands squarely on `case 3` and `case 5`. And here is the part I missed:

```
realpath("/test_dir/symlink")    = /test_dir/test_path      # dereferenced
realpath("/test_dir/symlink/x")  = /test_dir/test_path/x
```

`realpath()` resolves the **final** component too. When the audited object is a file reached *through* a symlinked directory, that is exactly what we want, which is why it is correct in `case 1`/`case 2`. When the audited object **is** the symlink, as in these two cases, it rewrites `/test_dir/symlink` into its target, the event goes out under the wrong path and FIM never reports the symlink itself.

So the distinction is symlink-as-intermediate-component (resolve) versus symlink-as-final-component (do not resolve). A correct fix for `case 3/4/5` has to resolve only the parent and re-append the last component, which `gen_audit_path()` is already shaped for since it takes parent and file separately. That also means `case 2` carries the same latent flaw for symlink-as-object, so it is not a two-line change and it needs its own coverage.

Given that, I would keep this PR to what the issue reports and what is verified green, and track the rename/hard-link resolution separately. Happy to open that issue with this reproducer if you agree, or to do it here if you prefer it in one go.

For the record, the gap you asked about is genuinely there. Measured on Ubuntu 24.04 / auditd 3.1.2 / kernel 6.8, watch on the resolved target, every operation through the symlink:

| Operation | Syscall | `items` | Branch | Resolves today |
|---|---|---|---|---|
| write existing file | `openat` | 1 | `case 1` | yes, this PR |
| create / delete file or dir | `openat` / `unlinkat` | 2 | `case 2` | yes, pre-existing |
| hard link | `linkat` | 3 | `case 3` | no |
| rename | `renameat2` | 4 | `case 4` | no |
| rename over existing | `renameat` | 5 | `case 5` | no |

Darioortegaleyva (approved):

MarcelKemp (approved):

Issue #38119, opened 2026-07-30
Agent alignment to the final manager HTTPS contract: CMAC fields, `/control` payloads, and strict protocol handling


Nicogp:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0 | wazuh-agentd / HTTPS client / `/control` / AES-CMAC auth | Agent | Sources | Linux / Windows / macOS |

> Part of the agent-side HTTPS migration (#37702, agent epic #37831).
> Focused follow-up to align the agent with the manager's final message/auth protocol definitions.
> Depends on **01** and **02**.

## Description

The manager-side spike introduced updated protocol definitions for AES-CMAC request signing and `/control` request payloads. The agent must support **only the new contract**; no backward-compatibility shim is desired for this development.

Authentication handling also changed: a `401` can mean either an out-of-window `timestamp` or a bad key. In both cases the manager returns its current time in a response header. The agent must retry once with the corrected timestamp delta; if that retry succeeds, it must persist the delta for future requests. If the retry fails again, the failure must be treated as a bad-key signal and escalated to re-enrollment.

Relevant contract sources:

- AES-CMAC auth contract with `timestamp` in the signed canonical request and `Authorization` header.
- `/control` startup request must include `version`.
- `/control` notify request must include `host` metadata.
- Authentication failures return generic `401`, with manager time available in a response header for timestamp correction.
- Request `Content-Type` per endpoint: `application/json` for `/control`, `/config` and `/stats`; deliberately `*/*` for `/stateless`; not JSON at all for `/stateful` and `/download`.

Updated `/control` request bodies:

Startup:

```json
{
  "type": "startup",
  "version": "5.0.0"
}
```

Notify:

```json
{
  "type": "notify",
  "agent": {
    "version": "5.0.0"
  },
  "host": {
    "hostname":     "ubuntu-test",
    "architecture": "x86_64",
    "ip":           "192.168.1.100",
    "os": {
      "name":     "Ubuntu",
      "version":  "20.04",
      "platform": "ubuntu",
      "type":     "linux"
    }
  }
}
```

## Evidence from the current branch (`enhancement/37831-agent-https-client`)

- AES-CMAC `timestamp` signing is already present in `canonicalRequest.cpp` and `cmacSigner.cpp`.
- `/control` startup already sends `version` in `controlStream.cpp`.
[...]

PR #38440, opened 2026-08-18
Windows agent HTTPS 401 clock-skew

### LINKED ISSUE #38439: Windows agent enrolls then never connects — remoted rejects future-dated HTTPS credentials, authd then refuses re-enrollment as a duplicate name
## Summary

On a Windows (or any) HTTPS agent whose clock is skewed relative to the manager, remoted correctly
rejects every request as `future_request` → HTTP 401. The agent's `https_client` cannot tell that
401 apart from a dead/revoked key — the manager deliberately answers both the same way — so it treats
the rejection as "I must not be registered" and immediately re-enrolls under the same hostname. authd
refuses the re-enrollment because the agent already holds the correct key for that name. The result is
a **permanent** loop: `credential rejected (401); re-enrolling` → `Duplicate agent name: <name>.
Unable to add agent (from manager)`, repeating forever on a lengthening backoff. The agent never
becomes `active` and no telemetry is ever sent — even though the package, the key, the listener and
the network are all fine. Only the clocks disagree, and the agent has no way to find that out or
recover from it.

Originally reported by an internal defect report
after a Windows VM in a fresh evaluation lab hit this because `w32time` never started following a
snapshot clone. **Independently reproduced live** for this issue against the `nightly/5.0.0` build
(manager + agent commit `c6ac880`) on a fresh vagrant AIO deploy + a real Windows test box — see
Reproduction below.

## Component

- `src/client-agent/https_client/` (agent-side HTTPS transport module — the 401 handling)
- `src/remoted/remoted_module/` (manager-side `future_request` classification, collapsed to a generic
  401 by design)
- `src/os_auth/` (authd's duplicate-name / re-enrollment refusal)

## Severity

High — a skewed clock is a common, mundane condition (VM snapshot restore, dead CMOS battery, BIOS
local-time-vs-UTC mismatch, no NTP on an air-gapped network), and once it triggers this loop the
agent is **permanently** stranded: it never sends a keepalive, never syncs FIM/SCA/syscollector, and
recovering requires manual intervention (fix the clock **and** delete the stale registration on the
manager — fixing the clock alone is not enough once the agent has already tried to re-enroll).

## Environment (this reproduction)

| Component | Version / detail |
|---|---|
| Wazuh manager / indexer / dashboard | 5.0.0 rc1, `nightly/5.0.0`, commit `c6ac880`, all-in-one |
| Windows agent | 5.0.0 rc1, `nightly/5.0.0` MSI, commit `c6ac880` (same commit as the manager) |
| Manager host | Ubuntu, vagrant-provisioned bare-metal AIO, HTTPS on port 1517 |
| Windows host | Real Windows 11 test box, `w32time` stopped, clock manually skewed +10h |

(The original report's lab used a different commit/topology — see that report for its own
environment table. Both environments hit the identical failure signature.)

## Observed behavior

Windows agent log (`ossec.log`), skewed +10h, looping every 5s → 5s → 15s → ... indefinitely:

```
23:05:43 wazuh-agentd:https-client: INFO: Starting https_client (server=192.168.1.233:1517, agent=001).
23:05:43 wazuh-agent: WARNING: https_client: credential rejected (401); re-enrolling.
23:05:43 wazuh-agent: WARNING: Manager unreachable. Pausing module event production.
23:05:43 wazuh-agent: INFO: Requesting a key from server: 192.168.1.233
23:05:43 wazuh-agent: INFO: Using password specified on file: authd.pass
23:05:43 wazuh-agent: INFO: Using agent name as: DESKTOP-0SABLFF
23:05:43 wazuh-agent: ERROR: Duplicate agent name: DESKTOP-0SABLFF. Unable to add agent (from manager)
```

Manager log (`wazuh-manager.log`), same instant:

```
16:05:37 wazuh-manager-remoted:endpoints: WARNING: Rejected 1 request(s) in the last 90 s whose
  timestamp fell outside the accepted window (future_request). Check clock synchronization between
  the agents and the manager; the window is set by 'auth_max_request_age' and 'auth_max_future_skew'.
16:05:37 wazuh-manager-authd: INFO: Duplicate name 'DESKTOP-0SABLFF', rejecting enrollment.
  Agent '001' has not been disconnected long enough to be replaced.
```

Confirmed no self-healing: the loop was still running, agent still `status='disconnected'`, after
~90s of continued observation. Manually correcting the clock and deleting the stale agent record on
the manager fixes it immediately — the next enrollment goes `active` in seconds with no 401.

## Root cause

The agent's only defense against a 401 is a one-shot retry (`retrySender.cpp`, from #38119, "align
agent HTTPS client to manager contract") that is *supposed* to distinguish a just-expired timestamp
from a dead key by re-signing and trying once more — the module's own comment says so:

```cpp
// client-agent/https_client/src/controlStream.cpp:611-613
// AuthFail only surfaces once the timestamp-corrected retry has also failed,
// so the key is genuinely bad
```

**But the retry is not timestamp-corrected — it is timestamp-refreshed from the exact same clock:**

```cpp
// client-agent/https_client/src/retrySender.cpp:60-67
if (result.outcome == OutcomeClass::AuthFail && !authRetried)
{
    authRetried = true;
    result = attemptOnce(base);   // re-signs, but from the SAME clock
    continue;
}
```

```cpp
// client-agent/https_client/src/retrySender.cpp:162
const auto timestamp = m_clock.wallSeconds();   // std::time(nullptr) — still skewed
```

With a persistent skew (not a one-off edge/expired timestamp), both the original request and the
"retry" are wrong by the same amount, so both get `future_request` → both collapse to the same
generic 401 (the manager deliberately answers 7 distinct auth failure causes identically —
`remoted_module/src/auth/authMiddleware.cpp:76-82` — specifically so a client can never distinguish
skew from a bad key by status code alone). The agent therefore escalates unconditionally:

```cpp
// retrySender.cpp:108-114
else if (result.outcome == OutcomeClass::AuthFail && m_authGate != nullptr)
{
    m_authGate->reportAuthFailure();   // pauses all traffic, triggers re-enrollment
}
```

...and the re-enrollment it triggers is **permanently futile**, not merely delayed. authd's
duplicate-name guard first refuses because the existing registration is too young
(`after_registration_time`, default 3600s — `os_auth/src/auth.c:305-311`), and **after that hour
expires it refuses again** because the re-enrolling agent's key hash matches what the manager already
has (`key_mismatch`, default on — `os_auth/src/auth.c:315-322`) — i.e. authd is effectively saying
"you already hold the correct key," a fact it computes but never surfaces to the agent in a way it can
act on.

**Good news for the fix**: the manager's transport already sends its own wall-clock time on every
response via the standard `Date` header, including on every 401
(`remoted_module/src/http_server/RestinioHttpServer.cpp:571/765/799`,
`append_header_date_field()`) — live-verified against the reproduction manager:

```
$ curl -sk -i -X POST https://192.168.1.233:1517/control \
    -H "Content-Type: application/json" -H "protocol-version: 1" \
    -H "Authorization: Wazuh 001:9999999999:deadbeef" -d '{"type":"notify"}'
HTTP/1.1 401 Unauthorized
Date: Tue, 18 Aug 2026 16:23:55 GMT

{"error":"Invalid client authentication","code":401}
```

So the agent already receives everything it needs to detect and correct real clock skew — it just
never reads it. No manager-side change is required to fix this.

## Steps to reproduce

1. Deploy a `nightly/5.0.0` (or newer) all-in-one manager and a Windows MSI from the same nightly,
   with remoted listening on `0.0.0.0:1517`.
2. On the Windows guest, stop `w32time`, and set the local clock ~10h ahead of the manager's UTC time
   (or use a box image that already boots that way after a snapshot restore).
3. Install/start the agent with `WAZUH_MANAGER=<manager-ip>` and the manager's enrollment password.
4. Watch `ossec.log`: `Valid key received` → `https_client` starts → `credential rejected (401)` →
   `Duplicate agent name`, repeating forever on a lengthening backoff.
5. Confirm on the manager: one `future_request` warning, then `Duplicate name '<agent>' ...`.
   `GET /agents` shows the Windows record stuck as `never_connected`/`disconnected`.
6. **Control**: set the Windows clock to the manager's UTC time, delete the stale agent record on the
   manager, remove `client.keys`, restart the service. The next enrollment goes `active` in seconds
   with no 401. If the 401 survives a matched clock, this report is wrong — in both the original lab
   and this reproduction, it did not.

Nicogp:
## Description

Closes #38439

On a Windows (or any) HTTPS agent whose clock is skewed relative to the manager, remoted correctly
rejects every request as `future_request` → HTTP 401. The agent's `https_client` cannot tell that
401 apart from a dead/revoked key — the manager deliberately answers both the same way — so it treats
the rejection as "I must not be registered" and immediately re-enrolls under the same hostname. authd
refuses the re-enrollment because the agent already holds the correct key for that name (the existing
`after_registration_time` grace period, then the `key_mismatch` check once that expires). Before this
fix the result was a permanent loop: `credential rejected (401); re-enrolling` → `Duplicate agent
name: <name>. Unable to add agent (from manager)`, repeating forever on a lengthening backoff. The
agent never became `active` and no telemetry was ever sent, even though the package, the key, the
listener and the network were all fine — only the clocks disagreed.

Reported by an internal defect report (PDF, Wazuh Researchers Team, 2026-08-18, Severity: High) after
a Windows VM in a fresh lab hit this because `w32time` never started after a snapshot clone.

Root cause, one sentence: the agent's only defense against a 401 was an already-existing one-shot
"retry with a fresh timestamp" (`retrySender.cpp`, from #38119) that never actually measured or
corrected clock skew — it re-signed from the *same* skewed clock — so every skew-caused 401 looked
identical to a dead key and drove the agent into a re-enrollment authd could never honor, forever.

## Proposed Changes

Agent-only, no manager change required.

Implements clock-skew detection/correction inside the existing one-shot 401 retry, so it stops being
a no-op under real skew. Verified live that the manager's HTTP transport already sends a standard
`Date` header on every response, including every 401 (`curl -sk -i` against `/control` with a bad
`Authorization` returned `Date: Tue, 18 Aug 2026 16:23:55 GMT` alongside the `401`).

- `httpTypes.hpp`: `HttpResponse` gains `serverDateSeconds` (0 = absent, same convention as the
  existing `retryAfterSeconds`).
- `iCurlHandle.hpp` / `curlHandle.cpp`: `captureRetryAfter(long*)` widened to
  `captureResponseHeaders(long* retryAfter, std::time_t* serverDate)` — libcurl allows only one
  HEADERFUNCTION/HEADERDATA pair per handle, so both now live behind one `HeaderCapture` struct.
  `Date:` is parsed with libcurl's own `curl_getdate()` (RFC 1123/850/asctime; `-1` on failure is
  treated as absent).
- `curlPerformer.cpp` / `tests/unit/mocks/mockCurlHandle.hpp` / `tests/unit/curlPerformer_test.cpp`:
  call site, mock and test updated to the new two-header capture.
- `sysSeams.hpp`: `IClock` gains `applyOffsetSeconds(int64_t)` (default no-op, so `SystemClock` and
  every existing test double are unaffected without changes). New `SkewCorrectedClock` decorator
  (header-only, atomic offset) applies the offset to `wallSeconds()` only — `steadyNow()` (cadence
  timing) is never touched. **Accumulates rather than replaces** (`fetch_add`, not `store`) — a review
  caught that replacing would discard an earlier correction by exactly its own magnitude on a second,
  independent skew event, since the delta is measured against the already-corrected clock; the
  accumulate contract is documented on `IClock` itself so no future clock implementation can
  reintroduce the bug.
- `httpsClientFacade.hpp`: `m_clock` becomes a `SkewCorrectedClock` wrapping a `SystemClock`. Since
  every one of the 7 HTTPS senders already takes `IClock&`, this is a drop-in swap with **zero
  changes** to any stream/sender constructor or call site — a skew learned by any one sender's 401
  corrects signing agent-wide from the very next request on.
- `retrySender.cpp` / `.hpp`: inside the existing one-shot 401 retry, a new `correctClockIfSkewed()`
  computes `delta = serverDateSeconds - m_clock.wallSeconds()` and, if `|delta| >= 5s` (a noise floor
  — smaller gaps are plausibly RTT/`Date`'s 1s granularity, not real skew), applies it via
  `m_clock.applyOffsetSeconds()` *before* the retry re-signs, and logs one WARNING naming the measured
  skew. Escalation logic (second, now-corrected 401 → `AuthGate` → re-enrollment) is unchanged — it
  now only fires once clock skew has actually been ruled out. Also documents, inline, that `Date` is
  unauthenticated and trusts whatever answered the TLS handshake — no new capability versus trusting
  the 401 itself, and only relevant if `verify_mode=none` is in play.

### CI (PR #38440)

`Unit-Tests (agent)` **passed** for real (confirms the fix actually compiles and its unit tests run
green — not just `-fsyntax-only`), along with `Unit-Tests (manager)`, `Unit-Tests (winagent)`, both
`rtr (client-agent/https_client, agent/winagent)` jobs, `scan-build-linux (agent/manager)`,
`scan-build-macos-agent`, `scan-build-ubuntu-mingw-windows`, all package builds (DEB/RPM/PKG
×amd64/arm64), `Build Windows agent binaries`/`Generate MSI package`/`Test MSI package`, and `IT
Windows/Linux - agentd`/`IT Windows/Linux - enrollment` integration tests. `gh pr checks` returns
clean (every check passed or correctly skipped). A few CI jobs hit an unrelated infra flake (a slow
`azure.archive.ubuntu.com` mirror timing out the MinGW toolchain install at the 30-minute job limit,
confirmed via job logs — nothing to do with this PR's code) and were re-enqueued via `gh run rerun`
until green.

### Live re-verification with the actual fixed build (2026-08-18, this session)

Downloaded the real Windows MSI built from this branch's tip
(`wazuh-agent_5.0.0-0_windows_a649a8b.msi`, `wazuh-agent-packages` run
[32187675557](https://github.com/wazuh/wazuh-agent-packages/actions/runs/32187675557)); provenance
verified via `gh api` (the artifact's embedded commit is a merge of this branch's tip onto its base)
and the installed agent's own `VERSION.json`. Clean-uninstalled the previous agent, deleted the stale
agent record from the manager's `global.db` (not just `client.keys` — authd's duplicate-name check
queries the agent table directly), and ran scenarios against the same redeployed `nightly/5.0.0`
manager (192.168.1.233) used for the original repro. Full logs under
`refs/verification-final-build/`.

**1. Baseline (aligned clock, fresh enroll) — no regression.**
Clean enrollment, agent `001 DESKTOP-0SABLFF`, `status='connected'` within seconds, no 401.
(`01-baseline-clean-enroll.log`, `01-baseline-manager.log`)

**2. The reported bug, reproduced then fixed live — clock skewed +10h, same key kept.**
Skewed the Windows clock +10h (`w32time` stopped, matching the original report) with the SAME
enrolled key already in place (no fresh enroll). Result:
```
2026/08/19 04:32:35 wazuh-agentd:https-client: INFO: Starting https_client (server=192.168.1.233:1517, agent=001).
2026/08/19 04:32:36 wazuh-agentd:https-client: INFO: https_client: clock skew of -35999 s detected against the manager's response (Date header); correcting the signing timestamp for this and future requests.
```
No `credential rejected`, no `Duplicate agent name` — the corrected retry succeeded on the first try.
`status='connected'`, same agent id throughout. Manager log shows exactly one `future_request`
warning at the matching timestamp and **no follow-up duplicate-name rejection** — contrast with the
original pre-fix repro, where the same warning was followed 0ms later by `Duplicate name
'DESKTOP-0SABLFF', rejecting enrollment.` This is the core defect, fixed and verified end to end.
(`02-clock-skew-corrected-info-level.log`, `02-clock-skew-corrected-manager.log`)

**3. Aligned clock + a genuinely bad key — confirms the fix doesn't misclassify a real credential
failure as skew.** Verified in an earlier round against this same fix (`refs/verification-fixed-build/
03-bad-key-no-skew-misclassification.log`): with a corrupted key and an aligned clock, no clock-skew
message fires (correctly *not* misidentified as skew) — `credential rejected (401); re-enrolling`
fires straight away, the pre-existing, unmodified-by-this-PR escalation path. Not worth re-running
here since it doesn't touch anything this session's only change (the log level) could affect.

Environment restored afterward: clock re-aligned to the manager's time, `w32time` re-enabled, agent
confirmed `status='connected'` before finishing.

### Live re-verification against the review-fixes build (2026-08-19, this session)

Downloaded the real Windows MSI built from the branch's current tip (`87d00015e3`, the `HeaderCapture`
refactor commit -- confirmed via `gh api` that the artifact's embedded commit `98d2e3c` is a real merge
of that tip onto its base), giving end-to-end coverage of all three review fixes (concurrency,
`CURLOPT_SUPPRESS_CONNECT_HEADERS`, exception safety) plus the `HeaderCapture` refactor, not just the
original clock-skew fix. In-place upgraded the same Windows test box (was running `a649a8b`), keeping
the existing `client.keys` to also exercise the normal upgrade path. Full logs under
`refs/verification-review-fixes-build/`.

**1. Post-upgrade baseline (aligned clock, existing key) -- no regression.** Clean start, no 401,
`status='connected'`, same agent id throughout.

**2. Clock skew, agent ahead (+10h) -- regression check after the concurrency fix.**
```
2026/08/19 23:36:04 wazuh-agentd:https-client: INFO: https_client: clock skew of -36000 s detected against the manager's response (Date header); correcting the signing timestamp for this and future requests.
```
Manager: exactly one `future_request` warning, no follow-up duplicate-name rejection. `status='connected'`
immediately after. Confirms the `correctToServerTime`/mutex rewrite didn't disturb the original fix.

**3. Clock skew, agent behind (-10h) -- first LIVE test of this direction** (every prior round, live
and unit, only exercised ahead).
```
2026/08/19 03:37:12 wazuh-agentd:https-client: INFO: https_client: clock skew of 36000 s detected against the manager's response (Date header); correcting the signing timestamp for this and future requests.
```
Positive delta, exactly mirroring case 2's negative one, as the symmetric design predicts.
`status='connected'` immediately after; same agent id, no re-enrollment. The manager did not log a
second, distinct `expired_request` warning for this -- traced to `remoted/remoted_module/src/endpoints/
endpoint.cpp`: `ExpiredRequest` and `FutureRequest` share one `RejectionKind::ClockSkew` throttle bucket
(a log-amplification defense against an unauthenticated peer forcing warning spam), so the second
rejection landed inside the first's throttle window and was silently suppressed at the log layer only --
not a manager-side gap, and independently confirmed by the agent's own log line, which can only fire
from inside the 401 retry branch.

**4. Environment restored**: clock re-aligned (`w32tm /resync` -- failed with "no time information
available", matching the original defect's root cause that this box's `w32time` has no reachable NTP
source, so re-set manually), `w32time` re-enabled, final clean restart shows no skew-correction line,
`status='connected'`.

## Artifacts Affected

- `libhttps_client.so` / `.dll` (agent HTTPS transport module) — Linux, Windows, macOS agents.
- No manager artifacts changed; no protocol/wire format changed (`Date` is a standard HTTP header the
  manager's transport already emits).

## Tests Introduced

- `tests/unit/retrySender_test.cpp` (+7 cases): a real skew corrected from the `Date` header and the
  retry recovers; the **mirror case with the agent clock behind the manager** (positive delta),
  confirming the correction isn't direction-specific — every other case here and every live E2E run
  only ever exercised the agent-ahead direction, so this closes that gap; a small (2s) gap is treated
  as noise, not skew; a missing `Date` header is a safe no-op; a corrected retry that still 401s
  escalates as a genuine key failure; the learned correction persists across a later, independent
  `send()` without re-learning it; a **second, independent skew event on top of an already-applied
  correction** converges on the manager's newly-reported time using a real `SkewCorrectedClock` (not
  the test double's own bookkeeping) — the regression test for the accumulate-vs-replace bug a
  reviewer caught.
- `tests/unit/sysSeams_test.cpp` (+3 cases): `SkewCorrectedClock` defaults to no correction and
  mirrors the base clock; an applied offset shifts `wallSeconds()` only; a new offset **accumulates**
  on top of a prior one (inverted from an earlier, incorrect "replaces" assertion).
- `tests/unit/curlPerformer_test.cpp`: existing test extended to assert `serverDateSeconds` flows back
  from the widened header capture.

## Review Checklist

- [x] Code changes reviewed (self-review passes caught and fixed an incorrect citation and an
      overclaimed comment about a timeout gap that isn't actually closed)
- [x] Relevant evidence provided — live repro before the fix, live-verified `Date` header assumption,
      clean CI (`gh pr checks` all green), and a live re-verification with the actual fixed build
      (see Results and Evidence)
- [x] Tests cover the new functionality (10 tests for the clock-skew fix itself, both directions
      covered, plus new tests for the review follow-up's concurrency fix and proxy-header fix)
- [x] Configuration changes documented (none introduced)
- [ ] Developer documentation reflects the changes (troubleshooting-doc note not yet written)
- [x] Meets requirements and/or definition of done — the reported defect is live-verified fixed
      against the real build, and CI is fully green
- [x] No unresolved dependencies with other issues (agent-only fix; no manager change required)

### DIFF
--- a/src/client-agent/https_client/src/httpsClientFacade.hpp
+++ b/src/client-agent/https_client/src/httpsClientFacade.hpp
@@ -88,7 +88,14 @@ class HttpsClientFacade final
         const LogFn m_logFn {HTTPS_CLIENT_LOGTAG};
 
         // Seams (built once; stable references handed to the streams).
-        SystemClock m_clock;
+        // m_clock wraps m_systemClock with a runtime-correctable offset: every
+        // stream below is handed IClock& m_clock, so a skew RetrySender learns
+        // from any one manager response corrects every sender's signing
+        // timestamp from the next request on (#37828 follow-up). Declaration
+        // order matters here -- m_systemClock must construct before m_clock,
+        // which holds a reference to it.
+        SystemClock m_systemClock;
+        SkewCorrectedClock m_clock {m_systemClock};
         Mt19937Random m_random;
         FsProbe m_fsProbe;
         ConfigKeyProvider m_keyProvider;
--- a/src/client-agent/https_client/src/iCurlHandle.hpp
+++ b/src/client-agent/https_client/src/iCurlHandle.hpp
@@ -63,6 +63,15 @@ inline constexpr long TLS_MIN_VERSION_1_3 {7};
 /// like the version above.
 inline constexpr long TLS_NATIVE_CA_STORE {1L << 4};
 
+/// Output slots for captureResponseHeaders(): libcurl allows only one
+/// HEADERFUNCTION/HEADERDATA pair per handle, so both captured headers live
+/// behind one struct instead of two separate parameters.
+struct HeaderCapture
+{
+    long* retryAfter {nullptr};
+    std::time_t* serverDate {nullptr};
+};
+
 /**
  * @brief One HTTP transfer, at the option level.
  *
@@ -91,13 +100,15 @@ class ICurlHandle
         virtual bool captureResponseToFile(std::FILE* file, uint64_t maxBytes) = 0;
 
         /// Installs the one HEADERFUNCTION/HEADERDATA pair libcurl allows per
-        /// handle: retryAfter receives the parsed Retry-After header (0 =
-        /// absent), serverDate receives the parsed Date header (0 =
-        /// absent/unparsed) -- the manager's own transport stamps Date on
-        /// every response it builds, including every 401, which is what lets
-        /// the agent detect and correct clock skew instead of assuming a 401
-        /// always means a dead credential.
-        virtual bool captureResponseHeaders(long* retryAfter, std::time_t* serverDate) = 0;
+        /// handle: capture.retryAfter receives the parsed Retry-After header
+        /// (0 = absent), capture.serverDate receives the parsed Date header
+        /// (0 = absent/unparsed) -- the manager's own transport stamps Date
+        /// on every response it builds, including every 401, which is what
+        /// lets the agent detect and correct clock skew instead of assuming
+        /// a 401 always means a dead credential.
+        /// @return false if the underlying option(s) were rejected by libcurl;
+        ///         the caller must not proceed to perform() in that case.
+        virtual bool captureResponseHeaders(HeaderCapture capture) = 0;
 
         virtual bool streamBodyFromFile(std::FILE* file, uint64_t size) = 0;
         virtual bool wireAbort(const std::atomic<bool>* abortFlag) = 0;
--- a/src/client-agent/https_client/src/iCurlHandle.hpp
+++ b/src/client-agent/https_client/src/iCurlHandle.hpp
@@ -17,6 +17,7 @@
 #include <atomic>
 #include <cstdint>
 #include <cstdio>
+#include <ctime>
 #include <functional>
 #include <memory>
 #include <string>
@@ -81,7 +82,16 @@ class ICurlHandle
         ///         silently not be in effect.
         virtual bool captureResponseBody(std::string* output) = 0;
         virtual bool captureResponseToFile(std::FILE* file, uint64_t maxBytes) = 0;
-        virtual bool captureRetryAfter(long* output) = 0;
+
+        /// Installs the one HEADERFUNCTION/HEADERDATA pair libcurl allows per
+        /// handle: retryAfter receives the parsed Retry-After header (0 =
+        /// absent), serverDate receives the parsed Date header (0 =
+        /// absent/unparsed) -- the manager's own transport stamps Date on
+        /// every response it builds, including every 401, which is what lets
+        /// the agent detect and correct clock skew instead of assuming a 401
+        /// always means a dead credential.
+        virtual bool captureResponseHeaders(long* retryAfter, std::time_t* serverDate) = 0;
+
         virtual bool streamBodyFromFile(std::FILE* file, uint64_t size) = 0;
         virtual bool wireAbort(const std::atomic<bool>* abortFlag) = 0;
 
--- a/src/client-agent/https_client/src/iCurlHandle.hpp
+++ b/src/client-agent/https_client/src/iCurlHandle.hpp
@@ -40,7 +40,14 @@ enum class CurlOption
     SslCiphers,     ///< string TLS 1.3 ciphersuite list
     SslOptions,     ///< long, the TLS behaviour bitmask
     FollowLocation, ///< long, always 0 (H4: no redirects)
-    NoSignal        ///< long, always 1 (H6)
+    NoSignal,       ///< long, always 1 (H6)
+    /// long, always 1: never deliver a forward-proxy's CONNECT-tunnel
+    /// response headers to the header callback. Without this, a proxy's own
+    /// Date could be captured as if it were the manager's (headerTrampoline
+    /// in curlHandle.cpp does not distinguish which hop a header block came
+    /// from), undermining the clock-skew correction's premise of trusting
+    /// only whoever completed the TLS handshake to the manager.
+    SuppressConnectHeaders
 };
 
 /// Value for CurlOption::SslVersion: refuse to negotiate below TLS 1.3.
--- a/src/client-agent/https_client/src/curlHandle.cpp
+++ b/src/client-agent/https_client/src/curlHandle.cpp
@@ -116,15 +116,38 @@ namespace
         return wrote;
     }
 
+    /// Bundles both headers this trampoline cares about: libcurl allows only
+    /// one HEADERFUNCTION/HEADERDATA pair per handle, so both live behind one
+    /// userData pointer instead of two.
+    struct HeaderCapture
+    {
+        long* retryAfter {nullptr};
+        std::time_t* serverDate {nullptr};
+    };
+
     size_t headerTrampoline(char* data, size_t size, size_t nmemb, void* userData)
     {
         const size_t total = size * nmemb;
-        auto* retryAfter = static_cast<long*>(userData);
-        constexpr size_t prefixLength = 12; // "Retry-After:"
+        auto* capture = static_cast<HeaderCapture*>(userData);
+        constexpr size_t retryAfterPrefixLength = 12; // "Retry-After:"
+        constexpr size_t datePrefixLength = 5;        // "Date:"
 
-        if (total > prefixLength && strncasecmp(data, "Retry-After:", prefixLength) == 0)
+        if (total > retryAfterPrefixLength && strncasecmp(data, "Retry-After:", retryAfterPrefixLength) == 0)
         {
-            *retryAfter = std::strtol(data + prefixLength, nullptr, 10);
+            *capture->retryAfter = std::strtol(data + retryAfterPrefixLength, nullptr, 10);
+        }
+        else if (total > datePrefixLength && strncasecmp(data, "Date:", datePrefixLength) == 0)
+        {
+            // curl_getdate() parses RFC 1123/850 and asctime formats and
+            // returns -1 on failure; a null-terminated copy is required since
+            // the header line is not itself nul-terminated by libcurl.
+            const std::string value(data + datePrefixLength, total - datePrefixLength);
+            const time_t parsed = curl_getdate(value.c_str(), nullptr);
+
+            if (parsed != -1)
+            {
+                *capture->serverDate = parsed;
+            }
         }
 
         return total;
@@ -237,10 +260,11 @@ namespace
                        curl_easy_setopt(m_handle, CURLOPT_WRITEDATA, &m_fileSink) == CURLE_OK;
             }
 
-            bool captureRetryAfter(long* output) override
+            bool captureResponseHeaders(long* retryAfter, std::time_t* serverDate) override
             {
+                m_headerCapture = HeaderCapture {retryAfter, serverDate};
                 return curl_easy_setopt(m_handle, CURLOPT_HEADERFUNCTION, headerTrampoline) == CURLE_OK &&
-                       curl_easy_setopt(m_handle, CURLOPT_HEADERDATA, output) == CURLE_OK;
+                       curl_easy_setopt(m_handle, CURLOPT_HEADERDATA, &m_headerCapture) == CURLE_OK;
             }
 
             bool streamBodyFromFile(std::FILE* file, uint64_t size) override
@@ -291,6 +315,7 @@ namespace
             CURL* m_handle {nullptr};
             curl_slist* m_headers {nullptr};
             FileSink m_fileSink {};
+            HeaderCapture m_headerCapture {};
     };
 } // namespace
 
--- a/src/client-agent/https_client/src/curlHandle.cpp
+++ b/src/client-agent/https_client/src/curlHandle.cpp
@@ -65,7 +65,8 @@ namespace
             {CurlOption::SslCiphers, CURLOPT_TLS13_CIPHERS},
             {CurlOption::SslOptions, CURLOPT_SSL_OPTIONS},
             {CurlOption::FollowLocation, CURLOPT_FOLLOWLOCATION},
-            {CurlOption::NoSignal, CURLOPT_NOSIGNAL}
+            {CurlOption::NoSignal, CURLOPT_NOSIGNAL},
+            {CurlOption::SuppressConnectHeaders, CURLOPT_SUPPRESS_CONNECT_HEADERS}
         };
         return *map;
     }
@@ -138,15 +139,26 @@ namespace
         }
         else if (total > datePrefixLength && strncasecmp(data, "Date:", datePrefixLength) == 0)
         {
-            // curl_getdate() parses RFC 1123/850 and asctime formats and
-            // returns -1 on failure; a null-terminated copy is required since
-            // the header line is not itself nul-terminated by libcurl.
-            const std::string value(data + datePrefixLength, total - datePrefixLength);
-            const time_t parsed = curl_getdate(value.c_str(), nullptr);
-
-            if (parsed != -1)
+            // curl callbacks are C: nothing may throw across them (see
+            // writeTrampoline above) -- std::string construction from
+            // unvalidated header bytes can throw std::bad_alloc under memory
+            // pressure, so guard it the same way.
+            try
             {
-                *capture->serverDate = parsed;
+                // curl_getdate() parses RFC 1123/850 and asctime formats and
+                // returns -1 on failure; a null-terminated copy is required
+                // since the header line is not itself nul-terminated by libcurl.
+                const std::string value(data + datePrefixLength, total - datePrefixLength);
+                const time_t parsed = curl_getdate(value.c_str(), nullptr);
+
+                if (parsed != -1)
+                {
+                    *capture->serverDate = parsed;
+                }
+            }
+            catch (...)
+            {
+                return 0; // LCOV_EXCL_LINE: allocation failure aborts the transfer.
             }
         }
 
--- a/src/client-agent/https_client/src/curlHandle.cpp
+++ b/src/client-agent/https_client/src/curlHandle.cpp
@@ -117,15 +117,6 @@ namespace
         return wrote;
     }
 
-    /// Bundles both headers this trampoline cares about: libcurl allows only
-    /// one HEADERFUNCTION/HEADERDATA pair per handle, so both live behind one
-    /// userData pointer instead of two.
-    struct HeaderCapture
-    {
-        long* retryAfter {nullptr};
-        std::time_t* serverDate {nullptr};
-    };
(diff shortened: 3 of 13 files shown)


MAnDumu:
## Verified live against a real manager — clock-skew self-correction works

Tested this branch (`fix/local-windows-agent-https-401-clock-skew`, `1c5043b`) with a real Linux agent built locally and pointed at a real 5.0.0 manager (the same manager used for the original clock-skew reproduction). Two scenarios, both recover cleanly with no manager-side change.

### 1. Single skew, matching the original report

Agent clock skewed forward via `faketime`, wrapped only around the `wazuh-agentd` binary (system clock untouched). Fresh enrollment, no prior key.

Agent log:
```
2026/08/19 03:19:39 wazuh-agentd[841965] https_client_bridge.c:1230 at bridge_on_state_change(): DEBUG: https_client connection state -> starting (1)
2026/08/19 03:19:39 wazuh-agentd:https-client[841965] retrySender.cpp:239 at correctClockIfSkewed(): INFO: https_client: clock skew of 29005 s detected against the manager's response (Date header); correcting the signing timestamp for this and future requests.
2026/08/19 03:19:40 wazuh-agentd[841965] https_client_bridge.c:1230 at bridge_on_state_change(): DEBUG: https_client connection state -> registered (2)
2026/08/19 03:19:40 wazuh-agentd[841965] https_client_bridge.c:347 at bridge_on_startup_result(): DEBUG: https_client startup accepted: {"agent":{"groups":["default"]},"cluster":{"name":"wazuh"},...}
```
No `credential rejected`, no re-enrollment, no `Duplicate agent name` — connects on the very first attempt after the one-shot correction. Before this fix, the same setup produced the reported infinite loop (confirmed separately during the original investigation).

### 2. Second, independent skew while a correction was already applied (no restart)

To exercise the correction path a second time without restarting the process (so any prior `applyOffsetSeconds()` state is still in effect), I used `libfaketime`'s dynamic `FAKETIME_TIMESTAMP_FILE` mode instead of the static `-f` wrapper: the agent starts under one skew, self-corrects as above, and then — with the process still running — I overwrite the timestamp file with a much larger, different skew and let the running process pick it up (`FAKETIME_CACHE_DURATION=1`).

Manager log, right at the moment of the second jump:
```
2026/08/19 14:33:48 wazuh-manager-remoted:endpoints: WARNING: Rejected 2 request(s) in the last 90 s whose timestamp fell outside the accepted window (future_request). Check clock synchronization between the agents and the manager; the window is set by 'auth_max_request_age' and 'auth_max_future_skew'.
```
Agent log, same moment:
```
2026/08/21 17:30:23 wazuh-agentd:https-client[845823] retrySender.cpp:239 at correctClockIfSkewed(): INFO: https_client: clock skew of -194400 s detected against the manager's response (Date header); correcting the signing timestamp for this and future requests.
```
No follow-up `credential rejected` or `Duplicate agent name` — `/stateless` batches and `/control` notify continued flowing normally right after. So a second, independent skew event landing on top of an already-corrected clock also self-heals in this run.

MAnDumu (dismissed):
Tests were done, both scenarios passed.

src/client-agent/https_client/src/iCurlHandle.hpp
```diff
@@ -77,7 +78,16 @@ class ICurlHandle
...
-        virtual void captureRetryAfter(long* output) = 0;
+
+        /// Installs the one HEADERFUNCTION/HEADERDATA pair libcurl allows per
+        /// handle: retryAfter receives the parsed Retry-After header (0 =
+        /// absent), serverDate receives the parsed Date header (0 =
+        /// absent/unparsed) -- the manager's own transport stamps Date on
+        /// every response it builds, including every 401, which is what lets
+        /// the agent detect and correct clock skew instead of assuming a 401
+        /// always means a dead credential.
+        virtual void captureResponseHeaders(long* retryAfter, std::time_t* serverDate) = 0;
```
jpcerrone: maybe return a HeaderCapture struct here? (curlHandle.cpp:122).

src/client-agent/https_client/src/httpsClientFacade.hpp
```diff
@@ -88,7 +88,14 @@ class HttpsClientFacade final
...
         // Seams (built once; stable references handed to the streams).
-        SystemClock m_clock;
+        // m_clock wraps m_systemClock with a runtime-correctable offset: every
+        // stream below is handed IClock& m_clock, so a skew RetrySender learns
+        // from any one manager response corrects every sender's signing
+        // timestamp from the next request on (#37828 follow-up). Declaration
+        // order matters here -- m_systemClock must construct before m_clock,
+        // which holds a reference to it.
+        SystemClock m_systemClock;
+        SkewCorrectedClock m_clock {m_systemClock};
```
jpcerrone, line 98: maybe it doesnt make sense to keep `m_systemClock` as a member? Its only used to construct m_clock in the next line

jr0me (dismissed):

jpcerrone (dismissed):

jr0me (dismissed):

jpcerrone (dismissed):

jr0me (approved):

Issue #38428, opened 2026-08-18
Windows agent: module threads are created after shutdown has already been requested — restart race silently drops SCA `scan_on_start`


MarcelKemp:
| Wazuh version                                    | Component                                    | Install type | Install method | Platform   |
| ------------------------------------------------ | -------------------------------------------- | ------------ | -------------- | ---------- |
| 5.0.0-rc1 (nightly 2026-08-18, commit `c6ac880`) | wazuh-agent (win32 service) / wazuh-modulesd | Agent        | Packages (MSI) | Windows 11 |

## Description

On Windows the service **startup** path and the service **stop** path run concurrently on two different threads with no ordering between them:

- `OssecServiceStart()` reports `SERVICE_RUNNING` (`src/win32/win_service.c:346`) **before** calling `local_start()` (`src/win32/win_service.c:357`), so the SCM will accept a `SERVICE_CONTROL_STOP` at any point during startup.
- `OssecServiceCtrlHandler()` runs on the SCM's own thread and calls `stop_wmodules()` (`src/win32/win_service.c:292`) directly.
- `stop_wmodules()` (`src/win32/win_utils.c:107`) walks the global `wmodules` list, while `local_start()` is still *building* that same list (`wm_config()`, `src/win32/win_utils.c:342`) and still *spawning* the worker threads for it (`src/win32/win_utils.c:350-366`).

There is no mutex, no "shutdown requested" check inside `local_start()`, and no barrier between the two. The result is a startup in which modules are stopped before they start, started after they were stopped, or both — in the same second.

```
11:41:20 wazuh-agent: INFO: Received exit signal. Starting exit process.
11:41:20 wazuh-agent: INFO: Set pending exit signal.
11:41:21 wazuh-rootcheck: INFO: Started (pid: 9508).            <-- starting during shutdown
11:41:21 wazuh-modulesd:agent-upgrade: INFO: Started (pid: 9508).
11:41:21 wazuh-modulesd:agent-info: INFO: AgentInfo module stopped.
11:41:21 wazuh-modulesd:sca: INFO: SCA module stopped.          <-- stopped, never started
11:41:21 wazuh-modulesd:syscollector: INFO: Started (pid: 9508). <-- started after the stop sweep
11:41:21 wazuh-agent: INFO: Startup completed.
11:41:22 wazuh-modulesd:syscollector: INFO: Stop received for Syscollector.
11:41:22 wazuh-agent: INFO: Exit completed successfully.
```

[...]

Issue #38372, opened 2026-08-14
SQLite database is locked in Windows 11 upgrade tests


rauldpm:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0-rc1-0d85377 | wazuh-modulesd | Agent | Packages | windows-11-amd64 |

The following messages were reported in the 2026-08-14 nightly run in upgrade tests for Windows systems

```
2026/08/13 21:55:23 wazuh-modulesd:syscollector: ERROR: sqlite: database is locked
```

```yaml
{
    "version": "5.0.0",
    "stage": "rc1",
    "commit": "0d85377"
}
```

This log has been reported previously in https://github.com/wazuh/wazuh/issues/37629 (which has a linked PR)

This log is not reported in https://github.com/wazuh/wazuh/issues/37780 as known message, if this is expected and it is a known log, we need to know it so we can suppress it

---

## Check list

The log message reported is not listed in the known message issue and matches the `The message pattern matches one of the entries above.` criteria

- Release/build under test: `5.0.0-rc1-0d85377`
- Platform: `Windows`
- Install method: `direct`
- Test scenario: `upgrade from 4.14.7`
- Full log excerpt with timestamps: [full_component_logs.zip](https://github.com/user-attachments/files/31062475/full_component_logs.zip)
- Component that emitted the message: `wazuh-modulesd`
- Whether the message is repeated or transient: `transient`
- Whether the agent recovered: `never failed`
- Functional impact, if any: `-`
- Link to this epic: https://github.com/wazuh/wazuh/issues/37780
- Rationale explaining why the message is not covered by the known-message list: `not listed`
- The logs from the core components, to provide the necessary context: `already provided`

hernanvalenzuela:
The `sqlite: database is locked` error reported here has its root cause in the mechanism described and fixed in #38370: syscollector's worker thread could fail to exit within the Windows agent's 20 s shutdown budget, leaving the old process holding the SQLite connection open while the new process tried to use it during the upgrade.

The root cause was fixed between PR #38463 (bounds the WMI calls in `scanHotfixes()` that could hang syscollector's worker thread)

cborla:
Related to: 
- https://github.com/wazuh/wazuh/issues/38371

Issue #38221, opened 2026-08-06
Improve stop_wmodules() shuts down modules sequentially


hernanvalenzuela:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0-rev | Wazuh component | Agent | Packages/Sources | Windows |

### Description
Follow-up from code review on PR #38151 (#38111 fix). `stop_wmodules()` (`src/win32/win_utils.c`) now correctly joins each module's worker thread before the service reports `SERVICE_STOPPED`, using a shared 20s budget for the `WaitForSingleObject` calls. However, the loop is still fully **sequential**, one module at a time:

```c
for (cur_module = wmodules; cur_module; cur_module = cur_module->next) {
    if (cur_module->context->stop) {
        cur_module->context->stop(cur_module->data);   // can itself block
    }
    if (cur_module->win_thread) {
        WaitForSingleObject(cur_module->win_thread, remaining_budget);
        ...
    }
}
```

#### Problem
Two of the modules Windows enables by default — <sca> and <wodle name="syscollector"> — have stop() callbacks
(wm_sca_stop() in wm_sca.c, wm_sys_stop() in wm_syscollector.c) that are not lightweight signals: each blocks synchronously for up to
10 seconds on its own condition variable, waiting for that specific module's worker thread to finish its own teardown, before stop_wmodules()
even reaches the next module in the list.

#### Consequences:

The current shared 20s budget only bounds the WaitForSingleObject join step — it does not bound the context->stop() call itself. With N modules whose stop() each blocks internally, total time in stop_wmodules() scales roughly with N (e.g. SCA + syscollector alone, both enabled by default, is already up to ~20s just from the two blocking stop() calls, before any join wait).
Because everything runs on a single thread, a slow module at the front of the list also delays the stop signal to every module after it that module's worker thread keeps running its normal loop, unaware shutdown is in progress, for as long as the earlier module's stop() blocks.
With enough configured wodles (syscollector, SCA, one or more command wodles, etc. — all supported on the Windows agent), worst-case total shutdown time can exceed the SCM's own patience (WaitToKillServiceTimeout, ~20s default), which can cause the SCM to consider the service hung and force-kill it reintroducing, for a different module, the same class of problem #38111 was about for syscollector/sysinfo.dll.
[...]

PR #38444, opened 2026-08-18
Make module stop() callbacks signal-only so modules do not deadlock each other on shutdown

### LINKED ISSUE #38371: Timeout waiting for AgentInfo run loop to exit
|---|---|---|---|---|
| 5.0.0-rc1-0d85377 | wazuh-modulesd | Agent | Packages | windows-11-amd64 |

```yaml
{
    "version": "5.0.0",
    "stage": "rc1",
    "commit": "0d85377"
}
```

The following warning message has been found in the 5.0.0 2026-08-14 nightly run

> 2026/08/13 21:56:33 wazuh-modulesd:agent-info: WARNING: Timeout waiting for AgentInfo run loop to exit; skipping database teardown.

This log has been reported previously in https://github.com/wazuh/wazuh/issues/38001 (which has a linked PR)

This log is not reported in https://github.com/wazuh/wazuh/issues/37780 as known message, if this is expected and it is a known log, we need to know it so we can suppress it

---

## Check list

- Release/build under test: `5.0.0-rc1-0d85377`
- Platform: `Windows`
- Install method: `packages`
- Test scenario: `upgrade from 4.14.7`
- Full log excerpt with timestamps: [full_component_logs.zip](https://github.com/user-attachments/files/31061736/full_component_logs.zip)
- Component that emitted the message: `wazuh-modulesd`
- Link to this epic: https://github.com/wazuh/wazuh/issues/37780

nbertoldo:
## Description

On the Windows agent, `wazuh-modulesd`'s agent-info module logs `WARNING: Timeout waiting for AgentInfo run loop to exit; skipping database teardown.` on shutdown and then leaves its DBSync connection open. This was reproduced on Windows 11 against `0d85377`, with the same shape as the 2026-08-13 nightly: `Pausing syscollector module for synchronization coordination` at 08:24:19, exit signal at 08:24:23, the warning at 08:24:34 (10 s later, to the second), and the coordination only unblocking at 08:24:44 when syscollector's evaluation finished.

The root cause is an ordering deadlock between two modules. `AgentInfoImpl::stop()` waited up to 10 s for its run loop to exit, but that loop parks inside `Syscollector::pause()`, whose predicate is `(!m_scanning && !m_syncing) || m_stopping`. `stop_wmodules()` walked the module list stopping and joining one module at a time, and stopped agent-info **before** syscollector — so at the moment agent-info began its 10 s wait, syscollector's `m_stopping` was still `false` and the `m_pauseCv.notify_all()` added by PR #38050 could not arrive in time. This is the residual that PR #38050's own description flagged as "Sequencing issue (identified, not yet fixed)", so #38001 was closed with half the problem outstanding.

Reading the whole shutdown path showed the waits were redundant for two of the three modules: syscollector and SCA already run their teardown on the module thread just before it returns (`wm_sys_main` calls `releaseResources()` right after `Module finished`; `wm_sca_main` calls `wm_sca_release_and_signal()` after joining its own sync worker), so joining that thread already carries the guarantee, under one shared budget rather than a per-module timeout stacked on top of it. agent-info was the asymmetric one: `wm_agent_info_main` just runs the loop and returns, and the teardown lived in `AgentInfoImpl::stop()` on the *stopping* thread, which is exactly why only agent-info produced this warning.

The same defect exists on POSIX, where `wm_handler` gets the join right (signal all, then join all against one absolute deadline) but pass 1 still calls the same blocking callbacks sequentially, with the 30 s deadline computed *after* that loop. Since the three callbacks are shared code, this fixes both platforms; only the Windows side needed the `stop_wmodules()` restructuring on top.

This supersedes #38221, whose premise is that the blocking `stop()` callbacks need to overlap: once they are signal-only there is nothing left to overlap, so the per-module thread it proposed is no longer needed. The one item from #38221 not covered here — reporting shutdown progress to the SCM via `dwCheckPoint`/`dwWaitHint` — is covered by PR #38469 (`report_stop_progress()`), so nothing is lost when #38221 is closed as a duplicate.

**Interaction with #38428 / PR #38469, and why this PR guards against it.** Signalling every module before joining any of them makes it likelier for a stop to arrive while a module thread is still inside `startup_gate_wait_for_ready()`, and that gate *releases* rather than aborts when shutdown is requested — so the thread goes on to run its start routine during shutdown. For syscollector that is not merely wasteful: `wm_sys_stop()` returns early while `flags.running` is still `false`, and that flag is only set further down `wm_sys_main()`, so the stop is discarded outright and the module is never told again. Measured on Windows 11 with the guard removed: **8 of 16 stops issued 1-2 s after start burned the whole 20 s join budget** (`Stop-Service` took 20.0 s, against 1.47 s max with the guard), and on the enrollment-triggered restart the successor process logged `sqlite: database is locked` ten times. `win_module_thread()` therefore checks `wm_shutdown_requested` after the gate returns and skips the start.

That guard is deliberately a stopgap: PR #38469 fixes it at the gate itself — `startup_gate_wait_for_ready()` returns why it released, so every caller can abort, including `wazuh-syscheckd`, `wazuh-logcollector` and `execd`, which this guard does not cover. **Merge order agreed with that PR's author: this one goes first, and #38469's rebase removes the guard**, since both touch the same lines in `win_module_thread()` and doing it that way leaves no window where the stop can be lost between the two merges. Two things that rebase has to account for: #38469 currently keeps `stop_wmodules()` stopping and joining one module at a time, which is the sequencing deadlock this PR exists to fix, so the two-pass split has to survive (its checkpoint slicing fits inside the join pass); and its sliced wait asserts `if (!remaining) wait_result = WAIT_TIMEOUT;`, reintroducing the false timeout removed here — a zero-timeout `WaitForSingleObject` is legal and reports `WAIT_OBJECT_0` for a thread that already finished.

## Proposed Changes

- **`wm_syscollector.c`** — `wm_sys_stop()` is now signal-only: the `need_shutdown_wait` block, its condition variable and mutex, the `sys_main_thread`/`sys_main_thread_initialized` bookkeeping and the `called_from_sys_main_thread` self-stop special case are all removed. The teardown stays where it already was, in `wm_sys_main`, before the thread returns. This deletes the `Syscollector did not confirm shutdown within N seconds` message along with the wait that emitted it.
- **`wm_sca.c`** — the equivalent removal in `wm_sca_stop()`, deleting the `Timeout waiting for SCA to complete shutdown` message. `wm_sca_release_and_signal()` is renamed `wm_sca_release_resources()` since it no longer wakes anyone. `wm_sca_start()` no longer resets `g_shutting_down`, because `wm_sca_is_shutting_down()` reads only that flag and clearing it would erase a stop that already arrived. It deliberately does not release anything on that path either: `wm_sca_stop()` sets the flag before calling `quiesce()`, which reads `m_spSyncProtocol` and `m_asyncFlushController` without `m_resourcesMutex`, so releasing from the module thread there could race it — the teardown stays on the normal exit path, after `sca_start_ptr()` has returned.
- **`agent_info_impl.cpp` / `.hpp`** — `AgentInfoImpl::stop()` becomes signal-only. A new `releaseResources()` performs the teardown (DBSync connection, then sync protocol, each under its existing mutex) and is idempotent so the destructor can also call it. `start()` no longer clears the stop, and it consults `isShutdownInProgress()` rather than `m_stopped` alone — `agent_info_stop()` null-checks the instance, so a stop arriving before the object exists never reaches `stop()` at all, and the injected global predicate is the only way to observe it. The run loop and its interval wait use the same predicate, which matters because callers supply no `shouldContinue` and it is therefore the loop's only exit condition. `m_runLoopActive`, `m_shutdownMutex` and `m_shutdownCv` are gone.
- **`wm_agent_info.c`, `agent_info.cpp`, `agent_info.h`** — new `wm_agent_info_release()` invoked on every exit path of the module thread, including the early `return` during the handshake wait. That path matters because `agent_info_task_registry_init()` may already have constructed the instance and opened `agent_info.db` before the wait, so returning without it would leave the database open until process teardown — i.e. past the point where the replacement agent starts (see #38069, #38372). It goes through a new `agent_info_release_resources()` export rather than the existing `agent_info_cleanup()`: cleanup resets the global instance with no lock, while `agent_info_parse_response()` (sync dispatcher thread) and `agent_info_task_check_and_record()` (`/control` thread) only null-check it, so destroying the instance from the module thread mid-shutdown would be a check-then-use use-after-free. Releasing the resources instead keeps the instance alive and does the teardown under the mutexes those callers take.
- **`win_utils.c`** — `stop_wmodules()` is split into two passes: signal every module, then join every module against one shared budget, mirroring the POSIX SIGTERM handler. The passes must not be interleaved, since a module can only finish winding down after a sibling has also been told to.
- **`win_utils.c`** — `win_module_thread()` does not start its module when `wm_shutdown_requested` is already set on return from `startup_gate_wait_for_ready()`. See the interaction with #38428 in the Description for the measurement that justifies it and for why it is a stopgap.
- **`wm_syscollector.c`** — `wm_sys_stop()` now records the stop before its `!flags.running` early return, and reads `syscollector_stop_ptr` into a local before calling through it (the module thread clears that global on its way out, so testing the global and then calling it could land on a NULL). `wm_sys_main()` checks the recorded stop once, **before** `syscollector_init_ptr()`: both `Syscollector::init()` and `Syscollector::start()` clear `m_stopping`, so a stop that arrived before this thread got there used to be erased, and the module would scan on through the shutdown, burn the shared join budget and leave its database locked for the successor. The check is before initialisation on purpose — nothing is open on that path, so the teardown has nothing to release and cannot race `quiesce()`. See the hazard note below for the window that stays open and why closing it does not belong here.
- **`win_utils.c`** — the shared budget now starts before the signal pass, so it covers everything `stop_wmodules()` does rather than the joins alone. The stops only signal, but they are not instantaneous: syscollector's and SCA's `quiesce()` both wait for an in-flight flush with no timeout of their own, and that time was previously off the clock, which left the total unbounded against the SCM's `WaitToKillServiceTimeout`.
- **`win_utils.c`** — the join pass skips the calling thread (`GetThreadId(handle) == GetCurrentThreadId()`), as the POSIX handler does with `pthread_self()`. `merror_exit()` from a module thread reaches `stop_wmodules()` through `WinSetError()`, so without this the pass waited on its own handle for the whole budget, logged a timeout that never happened, and left nothing for the modules after it.
- **`win_utils.c`** — the join no longer reports a timeout without checking the thread. `remaining ? WaitForSingleObject(...) : WAIT_TIMEOUT` asserted a timeout blindly once the shared budget was spent, so the error was logged for modules that had already exited cleanly; a zero timeout is legal and returns `WAIT_OBJECT_0` for a finished thread.

One pre-existing hazard was noticed while doing this and deliberately left alone, because fixing it properly means changing SCA's teardown locking and PR #38469 already modifies `sca_impl.cpp`. `quiesce()` reads `m_spSyncProtocol` and `m_asyncFlushController` **without** `m_resourcesMutex`, while `releaseResources()` moves and resets them under the exclusive lock — so the two running concurrently can dereference a destroyed object. Two paths in `wm_sca.c` reach `wm_sca_release_resources()` from the module thread before `sca_start_ptr()` has run, while `wm_sca_stop()` may be inside `quiesce()`: the agentd doclimits wait (`wm_sca.c:599`, reached whenever the agent is disconnected and a stop arrives, so the likelier of the two) and the `Cannot initialize SCA message queue` path (`wm_sca.c:641`). Syscollector has the same shape — its `releaseResources()` resets `m_spSyncProtocol`/`m_spSyncProtocolVD`, which its own `quiesce()` reads unlocked — on the ordinary teardown path. Two more items belong to that same issue rather than to this PR. SCA never refuses to start under a stop it already received: `SecurityConfigurationAssessment::Run()` sets `m_keepRunning = true` unconditionally and then calls `m_spSyncProtocol->reset()`, actively undoing the `stop()` `quiesce()` had just issued, so a stop landing between `wm_sca.c:594` and `sca_start_ptr()` is erased and the thread parks in its interval wait. And on the syscollector side a stop arriving *during* `init()` is still lost, because the only safe place to refuse is before anything is opened — skipping the scan after initialisation would send the module thread into `releaseResources()` precisely while `quiesce()` is running, with the databases open. None of this is introduced here, but making the stops signal-only lets the module thread reach its teardown sooner, so the windows are wider than before. All of it needs `quiesce()` and `releaseResources()` serialized first, which is a change to SCA's and syscollector's own teardown locking.

### Results and Evidence

Unit tests were built for the `winagent` target and run under wine:

```
agent_info_dbsync_test.exe   ->  [  PASSED  ] 12 tests.
agent_info_basic_test.exe    ->  [  PASSED  ] 14 tests.
```

All modified files compile clean with the MinGW cross-compiler used by the `winagent` build. `wm_agent_info.c` emits six pre-existing `-Wint-conversion` warnings for `return NULL` in the `DWORD WINAPI` variant; an A/B compile confirms the same six exist without this change, only at shifted line numbers.

#### The reported scenario

Before, on `0d85377` — the build the issue was filed against — with a metadata delta forced on `queue\agent_info\db\agent_info.db` so agent-info enters coordination, a throttled syscollector wodle so the first evaluation outlasts the 10 s wait, and `Restart-Service` once `Pausing syscollector module for synchronization coordination` appears:

```
08:24:19  agent-info:   INFO: Pausing syscollector module for synchronization coordination
08:24:23  wazuh-agent:  INFO: Received exit signal. Starting exit process.
08:24:34  agent-info:   WARNING: Timeout waiting for AgentInfo run loop to exit; skipping database teardown.
08:24:34  agent-info:   INFO: AgentInfo module stopped.
08:24:43  wazuh-agent:  ERROR: Module 'agent-info' worker thread did not exit within the 20000 ms shutdown budget
08:24:44  wazuh-agent:  ERROR: Module 'sca' worker thread did not exit within the 20000 ms shutdown budget
08:24:44  wazuh-agent:  ERROR: Module 'syscollector' worker thread did not exit within the 20000 ms shutdown budget
08:24:44  agent-info:   INFO: Agent stopping, aborting module coordination during pause phase
08:24:44  syscollector: INFO: Evaluation finished.
```

After, on `08d0a52`, same host and same scenario:

```
14:26:26  agent-info:   INFO: Pausing syscollector module for synchronization coordination
14:26:26  wazuh-agent:  INFO: Received exit signal. Starting exit process.
14:26:26  agent-info:   INFO: AgentInfo module stopped.
14:26:26  sca:          INFO: SCA module stopped.
14:26:26  syscollector: INFO: Stop received for Syscollector.
14:26:26  agent-info:   INFO: Agent stopping, aborting query to syscollector
14:26:26  agent-info:   INFO: AgentInfo module loop ended.
14:26:27  syscollector: INFO: Evaluation finished.
14:26:27  syscollector: INFO: Module finished.
```

Exit signal to `Exit completed successfully`: **1 s**, against ~21 s before. `Stop-Service` returned in 0.60 s and the process was gone 0.87 s after it was asked to stop. Zero ERROR and zero WARNING in the window. Two lines are worth reading: `AgentInfo module stopped.` now lands in the *same second* as the exit signal instead of eleven seconds later, which is the signal-only `stop()` doing its job; and `Module finished.` is present for syscollector, so its teardown ran and the DBSync connection was released before the join returned — the guarantee that used to depend on the deleted wait.

#### Regression matrix on Windows 11

Beyond the reported scenario, three suites were run against the same host to look for regressions elsewhere in the shutdown path. They replicate the assertions QA automation makes in `test_agents_install/test_04_restart.py` and `test_05_stop.py` (service state, PID change, no surviving process, and its zero-tolerance ERROR/WARNING log gate), plus 20 stop/start cycles from steady state and 16 stops issued 0/1/2/4 s after start.

| suite | result on `08d0a52` |
|---|---|
| QA parity — `test_04_restart` | PASS: service `Running`, PID changed, 0 errors, 0 warnings |
| QA parity — `test_05_stop` | PASS: service `Stopped`, no surviving process, 0 errors, 0 warnings |
| 20 cycles from steady state | shutdown 2 s max / 1.16 s mean; `Stop-Service` 1.25 s max; process gone 1.51 s max; 0 budget ERRORs; 0 `database is locked`; none of the three deleted messages; handles 637 → 645 |
| 16 stops during startup | 0 budget ERRORs; `Stop-Service` 1.47 s max |
| reported scenario | PASS (trace above) |
| `wazuh-agent` crash in the Windows Application log | none |

Two measurement details, in case anyone reproduces this. The CI package is not production-signed, so the agent emits 45 `Trust verification of a module failed` / `is not signed` warnings per start; they are counted separately and excluded, and do not occur with the signed packages QA uses. And the log gate must run at the default debug level: with `windows.debug` raised, DEBUG lines whose *text* contains "error" or "Warning" (SCA rules over `Windows Error Reporting`, `PasswordExpiryWarning`, `WarningLevel`) match the gate's regexes and produce false positives.

#### The same suites on four Windows versions

The three suites were then run on four systems, with the package each one had at the time. PowerShell 4.0 on 2012 R2 required the harness to avoid anything newer, and two of the hosts had no OpenSSH, so they were driven over WinRM.

| measure | Windows 11 | Server 2012 R2 | Server 2022 | Windows 10 |
|---|---|---|---|---|
| package | `08d0a52` | `08d0a52` | `4622806` | `659c0ea` |
| QA `test_04` / `test_05` | PASS / PASS | PASS / PASS | PASS / PASS | FAIL / FAIL (see below) |
| shutdown, max / mean (20 cycles) | 2 s / 1.16 s | 1 s / 1 s | 2 s / 1 s | 1 s / 0.89 s |
| `Stop-Service`, max | 1.25 s | 1.23 s | 1.91 s | 1.25 s |
| budget ERRORs, 20 cycles | 0 | 0 | 0 | 0 |
| budget ERRORs, 16 stops during startup | **0** | **0** | **0** | **0** |
| `database is locked` | 0 | 0 | 0 | 0 |
| any of the three deleted messages | 0 | 0 | 0 | 0 |
| handles, first → last cycle | 637 → 645 | 535 → 546 | 619 → 617 | 516 → 517 |
| reported scenario | **PASS** (1 s) | not exercised | **PASS** (2 s) | not exercised |
| `wazuh-agent` crash in the Application log | none | none | none | none |

Windows 10's `FAIL` is a single WARNING, not a shutdown defect: every assertion the QA tests make passes there — service state, PID change, no surviving process, zero errors, shutdown within the same second — and what trips the zero-tolerance gate is `wazuh-modulesd:sca: WARNING: Failed to clear SCA index before initial synchronization`. That is a manager-side failure: the agent's own log says `SCA synchronization deferred: Failed to communicate with the manager`, the session having taken ~4 min 45 s before giving up. The same `/stateful` session latency stopped the reported scenario from being exercised on 2012 R2 and Windows 10 (agent-info logs `Deferring coordination until FIM first sync completes` and never reaches the pause), and on Windows 10 it was confirmed to be the environment rather than the setup: the metadata delta was applied and there were still 18 deferrals in 600 s, with the manager healthy on resources and processing other agents' sessions. It is unrelated to this change and deserves its own issue.

Signature noise takes two forms depending on the host and is excluded from the gate in both: where the DigiCert CA is present the agent emits 45 `not signed` warnings per start, and where it is not — the Windows 10 image — it emits 44 `The dynamic signature validation is not available` instead. Neither occurs with the signed packages QA uses.

#### What the startup-gate guard is worth, measured

The same 16-stop suite was run on a build of this branch with the guard removed, which is what justifies keeping it until PR #38469 lands:

| 16 stops during startup | with guard (`08d0a52`) | without guard |
|---|---|---|
| budget ERRORs | **0** | **8** (4/4 at 1 s, 4/4 at 2 s) |
| `Stop-Service`, max | **1.47 s** | **20.02 s** |
| process gone, max | 30.62 s | 50.22 s |

On the same build without the guard, the enrollment-triggered restart also had syscollector start *after* the exit signal, blow the join budget, and leave the successor process logging `sqlite: database is locked` ten times.

One pre-existing behaviour, unrelated to this change and present in every build measured: when the agent is stopped 1-4 s after starting, `wazuh-agent.exe` sometimes outlives the SCM's `Stopped` by ~30.5 s (6 of 16 stops here, 5 of 16 on an earlier build of this branch, 1 of 4 on an ad-hoc run). Sampling every 2 s shows processor time and `ossec.log` frozen, so the process is blocked rather than working, and `Exit completed successfully` is already written. No 30 s constant exists in `src/win32`, `src/shared` or the agent's exit path; it deserves its own issue.

### Artifacts Affected

- `wazuh-agent.exe` (Windows agent) — contains `stop_wmodules()`, the module thread wrapper and the `wm_*` module wrappers.
- `wazuh-modulesd` (Linux/macOS agent and manager) — `wm_syscollector.c`, `wm_sca.c` and `wm_agent_info.c` are shared code, so the removal of the blocking waits applies to POSIX too.
- `agent_info.dll` / `libagent_info` — `AgentInfoImpl`.

No changes to `wazuh-syscheckd`, `wazuh-agentd`, `wazuh-logcollector` or any manager-only daemon.

### Configuration Changes

None. No option is added, removed or reinterpreted.

### Documentation Updates

None required for user-facing documentation: no configuration or behaviour visible to the user changes, beyond three warning messages that no longer occur.

Note that two of the deleted messages are tracked in the known-messages epic #37780 and in #38065; they can be dropped from that list once this is merged.

### Tests Introduced

No new test binaries. Existing agent-info unit tests were updated to the new contract:

- `AgentInfoImplTest.StopWaitsForRunLoopToExit` asserted the behaviour this change removes (`EXPECT_GE(elapsed, HOLD / 2)`, i.e. that `stop()` blocks). It is rewritten as `StopSignalsWithoutWaitingAndReleaseIsSeparate`, which asserts the new contract instead: `stop()` returns promptly while the run loop is mid-iteration, the loop then exits, and `releaseResources()` actually closes the DBSync connection.
- `AgentInfoImplTest.ParseResponseBufferAfterStopIsSafe` now calls `stop()` followed by `releaseResources()`, matching the order the module thread uses.
- Five `...FailsClosedWithoutDBSync` cases in `agent_info_dbsync_test.cpp` and `agent_info_vd_offset_test.cpp` used `stop()` as a shortcut to drop the DBSync connection; they now call `releaseResources()`, which is what performs that teardown.

## Review Checklist

- [x] Code changes reviewed
- [x] No unresolved dependencies with other issues — merge order agreed with #38469 (this PR first, its rebase drops the startup-gate guard).

### DIFF
--- a/src/wazuh_modules/agent_info/agent_info_impl/include/agent_info_impl.hpp
+++ b/src/wazuh_modules/agent_info/agent_info_impl/include/agent_info_impl.hpp
@@ -89,6 +89,15 @@ class AgentInfoImpl
         void start(int interval, int integrityInterval = 86400, std::function<bool()> shouldContinue = nullptr);
         void stop();
 
+        /**
+         * @brief Closes the DBSync connection and the sync protocol.
+         *
+         * Runs on the module thread once start()'s run loop has returned, so a caller
+         * that joins that thread is guaranteed the handles are closed. Idempotent, so
+         * the destructor can also call it for the case where start() never ran.
+         */
+        void releaseResources();
+
         /// @brief Override the flush poll delay (milliseconds). Only used in unit tests to avoid real sleeps.
         /// Negative values are clamped to 0.
         void setFlushPollDelayMs(int delayMs)
@@ -403,6 +412,13 @@ class AgentInfoImpl
         /// stop() writes it from another thread (avoids a data race).
         std::atomic<bool> m_stopped{false};
 
+        /// @brief True only while start()'s run loop is executing.
+        /// Read by releaseResources() to refuse tearing down members the loop may still
+        /// be using. The join in stop_wmodules() / wm_handler() is not authoritative --
+        /// both only log when the budget expires and then continue to process exit -- so
+        /// the destructor can be reached with this loop still running.
+        std::atomic<bool> m_runLoopActive{false};
+
         /// @brief Predicate reporting whether a shutdown is in progress (may be null).
         /// Injected from the module wrapper; reports the *global* agent shutdown, which is
         /// signaled before this module's own stop() runs.
@@ -463,15 +479,9 @@ class AgentInfoImpl
         /// @brief Mutex for synchronizing access to m_dBSync (prevents race conditions during cleanup/transactions)
         std::mutex m_dbSyncMutex;
 
-        /// @brief Serializes destruction of m_spSyncProtocol in stop()
+        /// @brief Serializes destruction of m_spSyncProtocol in releaseResources()
+        /// against the readers that run on other threads (parseResponseBuffer()).
         std::mutex m_syncProtocolMutex;
-
-        /// @brief Clean-stop handshake: stop() blocks until the run loop (start()) has
-        /// exited, so the sync-protocol connection can be closed with no other thread
-        /// using it.
-        std::mutex m_shutdownMutex;
-        std::condition_variable m_shutdownCv;
-        bool m_runLoopActive = false;
 
         /// @brief Flag set during updateChanges callback when cluster_name changed
         bool m_clusterNameChanged = false;
--- a/src/wazuh_modules/agent_info/agent_info_impl/src/agent_info_impl.cpp
+++ b/src/wazuh_modules/agent_info/agent_info_impl/src/agent_info_impl.cpp
@@ -187,6 +187,9 @@ AgentInfoImpl::AgentInfoImpl(std::string dbPath,
 AgentInfoImpl::~AgentInfoImpl()
 {
     stop();
+    // Covers the case where start() never ran, so its run loop never reached the
+    // teardown. Idempotent, so it is a no-op when the run loop already did it.
+    releaseResources();
     m_logFunction(LOG_INFO, "AgentInfo destroyed.");
 }
 
@@ -196,16 +199,31 @@ void AgentInfoImpl::start(int interval, int integrityInterval, std::function<boo
                   "AgentInfo module started with interval: " + std::to_string(interval) +
                   " seconds, integrity interval: " + std::to_string(integrityInterval) + " seconds.");
 
-    {
-        std::lock_guard<std::mutex> lock(m_shutdownMutex);
-        m_runLoopActive = true;
-    }
-
     // Load sync flags from database at startup
     loadSyncFlags();
 
     std::unique_lock<std::mutex> lock(m_mutex);
-    m_stopped = false;
+
+    // Do NOT clear a stop that arrived while the module was still starting up. The
+    // shutdown loop signals every module before joining any of them, so a stop can
+    // land before this point; clearing it here would leave the run loop below with no
+    // exit condition at all -- callers pass no shouldContinue -- and the module thread
+    // would never become joinable.
+    //
+    // m_stopped is therefore sticky for the life of the instance. That is fine today --
+    // one start() per process -- but a future hot-restart path must distinguish "stopped"
+    // from "shutting down" (an explicit reset here would bring back the lost stop above).
+    //
+    // The check is isShutdownInProgress() rather than m_stopped alone because
+    // agent_info_stop() null-checks the instance: a stop arriving before this object
+    // exists never reaches stop() at all. The module wrapper does set the global flag
+    // in that case, and the injected predicate is the only way to observe it.
+    if (isShutdownInProgress())
+    {
+        lock.unlock();
+        m_logFunction(LOG_INFO, "AgentInfo start aborted: shutdown already in progress.");
+        return;
+    }
 
     // Reset sync protocol stop flag to allow restarting operations
     if (m_spSyncProtocol)
@@ -217,6 +235,8 @@ void AgentInfoImpl::start(int interval, int integrityInterval, std::function<boo
     m_cv.wait_for(lock, std::chrono::seconds(5), [this] { return m_stopped.load(); });
 
     // Run at least once
+    m_runLoopActive.store(true);
+
     do
     {
         lock.unlock();
@@ -287,24 +307,57 @@ void AgentInfoImpl::start(int interval, int integrityInterval, std::function<boo
         // If no shouldContinue function provided, use default behavior (continue until stopped)
         bool shouldLoop = shouldContinue ? shouldContinue() : !m_stopped;
 
-        if (shouldLoop && !m_stopped)
+        if (shouldLoop && !isShutdownInProgress())
         {
             // Wait for the interval or until stop is signaled
             m_cv.wait_for(lock, std::chrono::seconds(interval), [this] { return m_stopped.load(); });
         }
 
     }
-    while (!m_stopped && (shouldContinue ? shouldContinue() : true));
+
+    // Same reason as the guard above: without the global predicate a stop lost to that
+    // null check, or arriving between it and here, would leave this loop with no exit.
+    while (!isShutdownInProgress() && (shouldContinue ? shouldContinue() : true));
+
+    m_runLoopActive.store(false);
 
     m_logFunction(LOG_INFO, "AgentInfo module loop ended.");
+}
 
-    // Publish that the run loop has fully exited so stop() can tear down the sync
-    // protocol without racing synchronizeMetadataOrGroups().
+void AgentInfoImpl::releaseResources()
+{
+    // Never free while the run loop may still be using these members. The join that is
+    // supposed to guarantee the loop finished is not authoritative: stop_wmodules() and
+    // wm_handler() only log when the shutdown budget expires and then carry on to
+    // process exit, so the destructor can reach here with the loop still running.
+    // Freeing then is a use-after-free, so skip it and let process teardown reclaim the
+    // handles -- the same trade-off the removed run-loop wait used to make, minus the
+    // blocking wait.
+    if (m_runLoopActive.load())
     {
-        std::lock_guard<std::mutex> shutdownLock(m_shutdownMutex);
-        m_runLoopActive = false;
+        m_logFunction(LOG_WARNING, "AgentInfo run loop still active; skipping database teardown.");
+        return;
     }
-    m_shutdownCv.notify_all();
+
+    // Idempotent: runs from the module thread when the run loop ends, and again from
+    // the destructor for the case where the run loop never started.
+    {
+        std::lock_guard<std::mutex> dbLock(m_dbSyncMutex);
+
+        if (m_dBSync)
+        {
+            m_logFunction(LOG_DEBUG, "Closing DBSync connection...");
+            m_dBSync.reset();
+            m_logFunction(LOG_DEBUG, "DBSync connection closed");
+        }
+    }
+
+    // Destroy the sync protocol so its SQLite connection to the persistent-queue db
+    // is closed. Guarded against parseResponseBuffer(), which runs on another thread.
+    {
+        std::lock_guard<std::mutex> lock(m_syncProtocolMutex);
+        m_spSyncProtocol.reset();
+    }
 }
 
 void AgentInfoImpl::stop()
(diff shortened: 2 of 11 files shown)


cborla (approved):
LGTM
- [x] Code changes reviewed

PR #38449, opened 2026-08-19
Close idle pre-authentication connections in the cluster server

### LINKED ISSUE #38448: Cluster server: no handshake deadline and no global connection limit for pre-authentication connections
## Description

`AbstractServerHandler.connection_made()` in `framework/wazuh/core/cluster/server.py` accepts a TCP connection on the cluster port and applies a single admission control, a per-source-IP cap:

```python
MAX_CONNECTIONS_PER_IP = 10  # server.py:24

def connection_made(self, transport):
    peername = transport.get_extra_info('peername')
    self.ip = peername[0]
    count = self.server.connection_counts.get(self.ip, 0)
    if count >= MAX_CONNECTIONS_PER_IP:
        self.logger.warning(f"Too many connections from {self.ip} ({count}), rejecting.")
        transport.close()
        return
    ...
```

Two gaps follow from this:

1. **No handshake deadline.** A connection that never completes the `hello` handshake is never closed. `check_clients_keepalive()` (`server.py:503`) is the only reaper and it iterates `self.clients`, which is populated exclusively by the `hello` handler (`server.py:138`). A pre-authentication deadline does exist — `_start_payload_deadline()` (`common.py:386`, `PRE_AUTH_PAYLOAD_TIMEOUT = 30`) — but it is armed from `msg_parse()`, i.e. only after a partial message header has been read. A connection that sends zero bytes never arms any timer and stays open indefinitely.

2. **No global connection limit.** `connection_counts` is keyed by source IP only; there is no global counter anywhere in the server.

Both were reproduced against the 4.14.9 branch: 10 idle TCP connections from one source IP stayed established through ~8 keepalive cycles with `server.clients` empty, and 30 distinct source addresses (10 connections each) held 300 simultaneous pre-authentication connections.

This is not tracked as a vulnerability (see GHSA-xv4g-3363-jx92): the exposure requires thousands of distinct non-spoofable source addresses, the cluster ships disabled by default, and port 1516 is documented as reachable only from cluster peers. The inconsistency is worth fixing on its own: the code already intends to time out pre-authentication connections and simply misses the zero-byte case.

## Suggested fix

Two changes in `framework/wazuh/core/cluster/server.py`, reusing the timer pattern already present in `common.py`:

1. **Handshake deadline.** In `connection_made()`, arm a `loop.call_later(...)` timer that closes the transport if `self.name` is still `None` when it fires. Cancel it in the success branch of `hello()` and in `connection_lost()`. Reuse `PRE_AUTH_PAYLOAD_TIMEOUT` or add a sibling constant so the timeout stays configurable in one place. This alone bounds idle pre-authentication connections to `accept_rate × timeout`.

2. **Global cap.** Add a `MAX_TOTAL_CONNECTIONS` constant and an integer counter on `AbstractServer`, incremented next to `connection_counts` in `connection_made()` and decremented in `connection_lost()` under the existing `_counted_connection` guard, rejecting with a warning when the cap is reached.

No refactoring, no change to the message protocol, and no behavioral change for a worker that completes `hello` promptly.

## Open decisions

- **Handshake timeout value.** A worker sends `hello` immediately after `connection_made()`, so 10–30 s is safe, but it should not be shorter than the connect + first-write latency tolerated on slow WAN links. Suggest reusing 30 s for consistency with `PRE_AUTH_PAYLOAD_TIMEOUT` unless there is a reason to differ.
- **`MAX_TOTAL_CONNECTIONS` value.** Needs to be derived from the expected cluster size (workers × a margin); a value set too low would break large clusters.

## Credit

Found and reported by @akhatkulov.

vikman90:
## Description
The cluster server accepted TCP connections on the cluster port and never closed those that
did not complete the `hello` handshake: the only reaper, `check_clients_keepalive()`, iterates
the post-handshake `clients` registry, and the existing pre-authentication deadline is armed
only after a partial message header is read, so a connection sending zero bytes stayed open
indefinitely. This change arms a handshake deadline as soon as a connection is accepted and
adds a global limit on simultaneous connections, complementing the existing per-source-IP cap.

Closes https://github.com/wazuh/wazuh/issues/38448

## Proposed Changes
- `framework/wazuh/core/cluster/server.py`:
  - New `MAX_TOTAL_CONNECTIONS = 1024` constant and `AbstractServer.total_connections` counter,
    incremented in `connection_made()` and decremented in `connection_lost()` under the existing
    `_counted_connection` guard; connections beyond the cap are rejected with a warning.
  - `connection_made()` arms a handshake deadline (`PRE_AUTH_PAYLOAD_TIMEOUT`, 30 s) that closes
    the transport if the handshake has not completed when it fires.
  - The deadline is cancelled in the success branch of `hello()` and in `connection_lost()`.
- `framework/wazuh/core/cluster/tests/test_server.py`: tests for the new behaviour and updated
  server mocks for the global counter.

### Results and Evidence
- Unit tests: `453 passed` for `framework/wazuh/core/cluster/tests/`. The six tests covering the
  new behaviour fail against the unpatched `server.py` and pass with it.
- Functional check against a real `AbstractServer` instance on `127.0.0.1`, with the deadline
  shortened to 3 s and the cap to 25 for the test: 60 connection attempts from 6 source
  addresses were admitted up to 25 (`Too many total connections (25), rejecting connection
  from ...`), and every idle connection was closed once the deadline expired
  (`Closing connection: handshake not completed within deadline.`), leaving
  `total_connections: 0` and `connection_counts: {}`.
- Server build with `make TARGET=server`: successful.

### Artifacts Affected
`wazuh-manager` — the `wazuh-clusterd` daemon (master node cluster server).

### Configuration Changes
None.

### Documentation Updates
None.

### Tests Introduced
In `framework/wazuh/core/cluster/tests/test_server.py`:
- `test_AbstractServerHandler_connection_made_total_limit`
- `test_AbstractServerHandler_handshake_deadline`
- `test_AbstractServerHandler_handshake_deadline_cancelled`
- Extended: `test_AbstractServerHandler_connection_made`,
  `test_AbstractServerHandler_connection_lost`, `test_AbstractServer_init`

## Credits
Found and reported by @akhatkulov.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 #### Fixed
 
 - Bounded the agent control message copy to the source string length in `wazuh-remoted`. ([#38427](https://github.com/wazuh/wazuh/pull/38427))
+- Fixed the cluster server keeping pre-authentication connections open indefinitely by adding a handshake deadline and a global connection limit. ([#38449](https://github.com/wazuh/wazuh/pull/38449))
 
 ### Agent
 
--- a/framework/wazuh/core/cluster/server.py
+++ b/framework/wazuh/core/cluster/server.py
@@ -23,6 +23,9 @@ from wazuh.core.cluster.utils import ClusterFilter, context_tag
 # Maximum simultaneous cluster TCP connections accepted from a single source IP.
 MAX_CONNECTIONS_PER_IP = 10
 
+# Maximum simultaneous cluster TCP connections accepted from all source IPs.
+MAX_TOTAL_CONNECTIONS = 1024
+
 
 class AbstractServerHandler(c_common.Handler):
     """
@@ -62,6 +65,8 @@ class AbstractServerHandler(c_common.Handler):
         self.broadcast_queue = asyncio.Queue()
         # Tracks whether this handler was counted in server.connection_counts.
         self._counted_connection = False
+        # Handle for the handshake-completion deadline timer.
+        self._handshake_deadline = None
 
     def to_dict(self) -> Dict:
         """Get basic info from AbstractServerHandler instance.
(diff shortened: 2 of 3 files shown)


jotacarma90 (approved):
LGTM

Issue #38442, opened 2026-08-18
Agent: `RescanRequester` builds its `RetrySender` with 5 of 7 arguments, silently disabling the compression config and the auth-failure gate on `/scan/vd`


juliancnn:
## Description

`RescanRequester` (the agent-side sender of `POST /scan/vd`, `src/client-agent/https_client/src/rescanRequester.cpp`) constructs its `RetrySender` like this:

```cpp
, m_sender(performer, signer, clock, m_backoff, &authGate)
```

`RetrySender`'s constructor takes **seven** parameters (`retrySender.hpp`):

```cpp
RetrySender(IHttpPerformer& performer, const ISigner& signer, IClock& clock, Backoff& backoff,
            bool compressionEnabled, CompressionGate* compressionGate = nullptr,
            AuthGate* authGate = nullptr);
```

With only five arguments supplied, `&authGate` (an `AuthGate*`) binds to the fifth parameter — `bool compressionEnabled` — through an implicit pointer→bool conversion, which always yields `true`. The two trailing pointer parameters then take their `nullptr` defaults. The code compiles without a single warning, and every observable effect is silent:

1. **`/scan/vd` bodies are always zstd-compressed**, regardless of the `httpsCompressionEnabled` module configuration. A deployment that disables HTTPS compression still sends this one endpoint compressed.
2. **The shared `CompressionGate` is not wired** (`compressionGate == nullptr`). The gate's contract is fleet-wide: a `415` on *any* send path reports to it and disables compression for **all** of the agent's senders for the rest of the run. A `415` on `/scan/vd` can neither disable compression for the other paths, nor be disabled *by* them — after a `415` elsewhere turns the gate off, `/scan/vd` keeps sending compressed bodies to a manager that already said it does not accept them.
3. **The `AuthGate` is not wired** (`authGate == nullptr`). A persistent `401` on `/scan/vd` never reaches `AuthGate::reportAuthFailure()`, so the "pause all traffic and surface re-enrollment once" escalation (#37828) never triggers from this path. `RetrySender::send()` still returns `AuthFail` after its one-shot fresh-timestamp retry, but the incident-level handling is skipped.

Every other construction site in the module passes all seven arguments — `controlStream.cpp`, `statefulStream.cpp`, `reporterStream.cpp`, `wpkFetcher.cpp`, `statelessStream.cpp`, `configFetcher.cpp`. `RescanRequester` is the single outlier.

## Impact

[...]

Issue #38438, opened 2026-08-18
Add HTTPS registration endpoint that bridges agent enrollment to authd


TomasTurina:
### Description

remoted's new agent-facing HTTPS server (`remoted_module`, port 1517, mTLS + CMAC-based auth via `etc/client.keys`) currently assumes every caller is already an enrolled agent. A brand-new agent has no key yet, so it cannot use any of the existing routes (`/stateless`, `/stateful`, `/control`, etc.) and must still fall back to opening a separate legacy TLS connection to authd on port 1515 to enroll (`OSSEC A:...` / `OSSEC K:...` protocol).

This issue proposes adding a **registration endpoint** to remoted's HTTPS server so agents can enroll using the same HTTPS channel (port 1517) they use for everything else, instead of needing a second protocol/port.

**Important scope boundary:** this is *not* a reimplementation of authd. authd keeps running exactly as it does today and remains the single source of truth for enrollment: password validation, agent name/IP/group validation, key generation, force-replace decisions, `max_agents` enforcement, cluster (worker→master) forwarding, and persistence to `client.keys` / `wazuh-db`. remoted's new endpoint is a **bridge**: it receives the agent's HTTPS enrollment request, authenticates it using the same shared mechanism authd uses today (the authd password — since, unlike the other HTTPS routes, there is no per-agent key yet to authenticate with via CMAC), forwards the request to authd, and relays authd's response back to the agent over HTTPS.

authd's port 1515 keeps running unchanged and unaffected for agents/tools that still use it directly (coexistence, not a replacement).

### Tasks

- [x] **Design: endpoint contract.** Define the new HTTP route (e.g. `POST /register`) request/response schema, mapping the fields already carried by the `OSSEC A:...` / `K:...` protocol today (agent name, version, groups, source IP override, optional key hash for replace, password).
- [x] **Design: internal transport to authd.** Decide how remoted forwards a validated registration request to authd. Candidate options to evaluate:
  - Reuse authd's existing local Unix-socket interface (`src/os_auth/src/local-server.c` / `authcom.c`, the `local_add` path already used by the manager CLI/API), which today bypasses TLS/password because it's a trusted local caller.
  - Have remoted act as a local TLS client of authd's existing network protocol against `127.0.0.1:1515`, replicating today's wire format.
  Document the trade-offs (duplicated auth logic, coupling/deployment ordering, code reuse) and pick one.
[...]

PR #38470, opened 2026-08-19
Agent-side enrollment over HTTPS: consume the new `POST /enroll` endpoint, retire the legacy 1515 protocol

### LINKED ISSUE #38465: Agent-side enrollment over HTTPS: consume the new `POST /enroll` endpoint, retire the legacy 1515 protocol
### Description

As part of the HTTPS comms migration (#37702), remoted's HTTPS server (port 1517)
is adding a `POST /enroll` endpoint (#38438) so agents can obtain their
`client.keys` over the same channel used for `/stateless`, `/stateful`,
`/control`, `/download`, `/stats`, and `/config`, instead of opening a second
TLS connection to `authd` on port 1515 with the legacy `OSSEC A:...` / `OSSEC
K:...` protocol.

This issue covers the **agent-side** consumer of that endpoint.

**Explicit scope decision**: the agent will **not** support both enrollment
protocols. Once this ships, the 5.x agent speaks `/enroll` exclusively — there
is no fallback to legacy 1515 enrollment.
A consequence worth confirming during grooming: a 5.x agent can only
enroll against a 5.x manager that has #38438 deployed.

**Explicit scope decision**: the new endpoint must reuse the **same TLS
credentials already configured for every other HTTPS endpoint**
(`<agent><ssl>`), not a second, independent certificate/key/CA
configuration. See "Connection config duplication" below — this is a real
behavior change, not just plumbing.

**Explicit scope decision**: enrollment must connect to the **same
manager address/port already configured for every other HTTPS endpoint**
(`<agent><server>`), not the separate `manager_address`/`port`/
`interface_index` options under `<enrollment>`. Same rationale as the TLS
credentials above — one connection target for the agent, not two. See
"Connection config duplication" below.

### Current state (agent side, verified against `wazuh/wazuh` HEAD `6c600c247`)

- Legacy protocol lives in `src/shared/src/enrollment_op.c` (629 lines,
  `w_enrollment_*` family): builds `OSSEC [PASS:...] A:'name' V:'ver'
  G:'groups' IP:'ip' K:'hash'`, opens its own `SSL_CTX`
  (`w_enrollment_connect()`, `enrollment_op.c:185`) independent from the
  HTTPS client's TLS stack.
- Triggered from two call sites, both going through
  `try_enroll_to_server()` (`src/client-agent/src/start_agent.c:115-127`):
  1. **First boot without `client.keys`** — `w_agentd_keys_init()`
     (`start_agent.c:67-113`), called from `start_agent_prepare()` before the
     HTTPS client starts.
  2. **401 → re-enrollment** — `bridge_on_reenroll_required()` /
     `bridge_reenroll_thread()` (`src/client-agent/src/https_client_bridge.c:132-182,402-416`),
     the central handler introduced in #37828. Its own comment
     (`https_client_bridge.c:24`) says it's "wired to the existing authd flow
     (`try_enroll_to_server`, `start_agent.c`)" — i.e. still 100% legacy today.
- Config: `<agent><enrollment>` parsed by `Read_Agent_Enrollment()`
  (`src/config/src/client-config.c:829-993`), a **sibling** of `<agent><ssl>`
  (`client-config.c:105-113`), not nested under it.

### Connection config duplication (the part this issue must fix)

Today there are **two independent connection/credential sets** on the agent
for reaching the manager:

**TLS material**

| | `<agent><ssl>` (used by `/stateless`, `/stateful`, `/control`, etc.) | `<agent><enrollment>` (legacy 1515 only) |
|---|---|---|
| Struct | `agent_ssl` (`client-config.h:42-48`) | `w_enrollment_cert` (`enrollment_op.h:60-68`) |
| Cert | `certificate` | `agent_certificate_path` |
| Key | `key` | `agent_key_path` |
| CA | `certificate_authorities` | `server_ca_path` |
| Verify mode | `verification_mode` (full/certificate/none) | none — implicit: skip verification if no CA set |
| Ciphers | `ciphers` | `ssl_cipher` |
| Consumed by | `bridge_build_config()` → `hc_config_t` (`https_client_bridge.c:1551-1598`) | `w_enrollment_connect()`'s own `os_ssl_keys()` call |

**Connection target (address/port)**

| | `<agent><server>` (used by every other HTTPS endpoint) | `<agent><enrollment>` (legacy 1515 only) |
|---|---|---|
| Struct | `agent_server` (`client-config.h:22-28`: `rip`, `port`, `network_interface`) | own `manager_address`, `port`, `interface_index` fields |
| Parsed by | `Read_Agent_Server()` (`client-config.c:347-463`) | `Read_Agent_Enrollment()` (`client-config.c:829-993`) |
| Consumed by | `agt->server[]`, dialed by the HTTPS client for `/stateless` etc. | `w_enrollment_connect()` — dials `cfg->target_cfg->port` (default 1515) independently |

The new `/enroll` client must consume both `agt->ssl` (same `hc_config_t` /
`Read_Agent_SSL()` path everything else uses) **and** `agt->server[]` (same
target the HTTPS client already dials), and must **not** use
`w_enrollment_cert` or the enrollment block's own `manager_address`/`port`/
`interface_index`. Concretely, under `<enrollment>`:
`agent_certificate_path`, `agent_key_path`, `server_ca_path`, `ssl_cipher`,
`manager_address`, `port`, and `interface_index` all become redundant with
`<agent><server>`/`<agent><ssl>` and should be removed from the enrollment
config surface (decide during grooming whether to hard-reject them or
silently ignore with a deprecation log, per the project's stated "no 5.x↔5.x
compat needed" stance from the related config work). Keep the fields that
have no equivalent elsewhere: `enabled`, `agent_name`, `groups`,
`agent_address`, `authorization_pass_path`, `delay_after_enrollment`,
`use_source_ip`.

### Design gaps to resolve (found while mapping legacy fields to the new contract)

The server contract (#38438 comment) defines the request body as
`{name, version, groups, ip}` plus, in password mode, header `Authorization:
WazuhEnroll <ts>:<mac>` (HKDF-SHA256 over the `authd` password, then
AES-256-CBC-CMAC over a canonical request string). Mapping the legacy fields:

- `name`, `version`, `groups`, `ip` (source IP override): map 1:1 to today's
  `A:`, `V:`, `G:`, `IP:` fields (server-side parsed in
  `src/os_auth/src/auth.c:78-246`). Confirm the `groups` shape (today a single
  comma-joined, de-duplicated string — check if the new JSON wants an array).
- **`PASS:` (enrollment password)**: maps to the `WazuhEnroll` HKDF/CMAC
  header in password mode — the agent needs an **HKDF-SHA256 + AES-256-CBC-CMAC**
  implementation for this new secret (the existing `CmacSigner`
  (`src/client-agent/https_client/src/cmacSigner.cpp:207-214`) signs with the
  *agent's* `client.keys` key, which doesn't exist yet pre-enrollment — this is
  a genuinely new signer, not a reuse of the existing one, even though the header
  shape is the same family: `Authorization: Wazuh <agentId>:<ts>:<mac>` vs.
  `Authorization: WazuhEnroll <ts>:<mac>`). Source the password from
  `authorization_pass_path` (already an existing `<enrollment>` option) —
  confirm this is meant to be the same file as the manager's `etc/authd.pass`
  semantics (copied to the agent) per the server contract.
- **`K:` (key hash, force-replace)**: no visible equivalent in the
  `{name, version, groups, ip}` body per the contract as posted. Confirm with
  the #38438 owner whether force-replace is in scope for the HTTPS path at
  all, and if so where the hash travels (body field not yet documented, or a
  header).
- **mTLS mode**: when the manager requires client certs, no HMAC step is
  needed — the already-negotiated TLS client certificate (from `<agent><ssl>`)
  is the credential. Confirm the agent doesn't need to also send the
  `WazuhEnroll` header in this mode (contract implies mode is exclusive, not
  additive).
- **Open mode**: no credential at all — straightforward.

### Scope

**1. Retire the legacy enrollment client**
- `src/shared/src/enrollment_op.c` / `include/enrollment_op.h`: remove the
  `OSSEC A:`/`K:` wire protocol and its private `SSL_CTX` handling. Decide
  whether any parsing/validation helpers are worth keeping as internal utility
  functions for the new client, or if this is a clean rewrite.

**2. Implement the new `/enroll` HTTPS client**
- Add a client function (new module, or extend `https_client`) that:
  - builds the JSON body per the finalized contract,
  - selects the auth mode (mTLS / password / open) based on local config —
    mirror however the manager derives its mode, since the agent needs to know
    which header (if any) to send,
  - implements HKDF-SHA256 + AES-256-CBC-CMAC for password mode (see gap
    above),
  - calls the manager on the same host/port already resolved for the other
    HTTPS endpoints (`agt->server[]`), reusing the shared `hc_config_t` TLS
    setup (`bridge_build_config()`) — no separate address/port resolution or
    cert loading path.
- Rewire both call sites to the new client, keeping their existing signatures
  where possible so the surrounding retry/backoff logic doesn't need to change:
  - `try_enroll_to_server()` (`start_agent.c:115-127`) — first-boot enrollment.
  - `bridge_reenroll_thread()` (`https_client_bridge.c:132-182`) — 401 →
    re-enrollment.

**3. Config (`src/config/src/client-config.c`)**
- `Read_Agent_Enrollment()` (`:829-993`): drop `ssl_cipher`,
  `server_ca_path`, `agent_certificate_path`, `agent_key_path` (superseded by
  `<agent><ssl>`) and `manager_address`, `port`, `interface_index`
  (superseded by `<agent><server>`); keep the rest.

**4. Tests**
- `src/unit_tests/shared/test_enrollment_op.c` (~40 tests covering the legacy
  protocol): retire/replace with tests for the new client.
- `src/unit_tests/client-agent/test_https_client_bridge.c`, "401 →
  re-enrollment" section (lines ~645-800): mocks of `try_enroll_to_server`
  will need updating to the new internals; keep the call-site contract stable
  if possible so these mostly survive.
- `src/unit_tests/client-agent/test_start_agent.c`: currently has **no**
  coverage for `w_agentd_keys_init()` / `try_enroll_to_server()` — add it for
  the new flow rather than inheriting the gap.
- No existing unit coverage for `Read_Agent_Enrollment()` either — add
  coverage for the trimmed config surface.
- Integration tests: `tests/integration/test_enrollment/test_agent/test_agentd_enrollment.py`,
  `tests/integration/test_agentd/test_reconnection/test_agentd_reconection_enrollment_*.py`,
  `test_agentd_initial_enrollment_retries.py`, `test_agentd_connection_retries_pre_enrollment.py`,
  and their config/case yaml fixtures — full rewrite for the HTTPS flow.
  `tests/integration/test_authd/*` tests the manager side against 4.x-style
  clients and can stay legacy-scoped (server keeps 1515 for old agents per
  #38438) — only the agent-facing suites above need rework.
- `tools/testing/enrollment-simulator/` (standalone tool emulating the legacy
  `OSSEC A:` protocol for test purposes): obsolete or needs a rewrite to
  simulate `/enroll` — confirm who owns/uses this before deciding.
- CI: `.github/workflows/4_testintegration_enrollment-tier-0-1-{lin,win}.yml`.

**5. Docs**
- Update the agent enrollment documentation and the `<enrollment>` config
  reference to reflect the trimmed option set and the new transport.

### Out of scope
- Any change to `authd` or the manager-side `/enroll` implementation — tracked
  entirely in #38438.
- Supporting a 4.x agent talking to the new endpoint (4.x agents keep using
  1515 against `authd`, unaffected by this issue).
- A dual-protocol agent fallback (explicit decision above).

### Acceptance Criteria
- [ ] A 5.x agent with no `client.keys` enrolls successfully via `POST
      /enroll` on first boot, using the manager/port already configured for
      HTTPS.
- [ ] The 401 → re-enrollment flow uses the same new client instead of the
      legacy 1515 path.
- [ ] Enrollment reuses `<agent><ssl>` for all TLS material and
      `<agent><server>` for the connection target; `<enrollment>` no longer
      carries its own cert/key/CA/cipher/address/port/interface options.
- [ ] All three auth modes (mTLS, password, open) work end-to-end against a
      manager configured for each.
- [ ] The legacy `OSSEC A:`/`K:` client code path is removed from the agent.
- [ ] Unit and integration tests updated; the pre-existing coverage gaps on
      `w_agentd_keys_init()`/`try_enroll_to_server()` and
      `Read_Agent_Enrollment()` are closed as part of this work.
- [ ] Documentation reflects the new enrollment transport and trimmed config.

Nicogp:
## Description

The agent currently enrolls over the legacy OSSEC `A:`/`K:` wire protocol on TCP 1515, with its own private `SSL_CTX` duplicating the TLS configuration the HTTPS transport already has. The manager's new `POST /enroll` endpoint (#38438, in progress) exposes enrollment over the same HTTPS channel as every other agent-manager exchange (port 1517).

This PR makes the agent consume `POST /enroll` exclusively — no dual-protocol fallback, per this issue's explicit scope — reusing the existing `<agent><ssl>`/`<agent><server>` transport instead of `<enrollment>`'s own connection settings.

Closes #38465

## Proposed Changes

- **New handle-less ABI call `hc_enroll()`** in the `https_client` module (`include/https_client.h`, `src/hcInterface.cpp`), modeled on the existing `hc_send_sync_session()` precedent — it works before `hc_create()` ever runs, since first-boot enrollment has no module handle yet.
- **New C++ signing/transport** (`https_client/src/enrollClient.{hpp,cpp}`, `enrollSigner.{hpp,cpp}`): reuses `CurlPerformer` as-is for TLS; password-mode requests are signed with HKDF-SHA256 + AES-256-CBC-CMAC per #38438's contract, independently verified against its published known-answer vector before implementation.
- **New C-side request/response handling** (`client-agent/src/enrollment.c`, `include/enrollment.h`): builds and validates the JSON body, parses the response, and writes `client.keys` atomically (the platform-specific POSIX/Windows write path is ported verbatim from the legacy code).
- **Zero coupling between the two**: `agentd`'s `try_enroll_to_server()` (`client-agent/src/start_agent.c`) explicitly sequences build → send → parse; `enrollment.c` and `https_client_bridge.c` never call each other directly.
- **`<agent><enrollment>` becomes a by-value struct** (`config/include/client-config.h`), like `<agent><ssl>`. The seven options that duplicated `<agent><server>`/`<agent><ssl>` (`manager_address`, `port`, `interface_index`, `ssl_cipher`, `server_ca_path`, `agent_certificate_path`, `agent_key_path`) are recognized but ignored, logged at **INFO** — never rejected, so a 4.x `ossec.conf` survives an in-place upgrade untouched.
- **Removes `shared/src/enrollment_op.c` + `shared/include/enrollment_op.h`** (the legacy OSSEC `A:`/`K:` wire format and its private `SSL_CTX`) and their test.
- `internal_options.conf`'s `agent.recv_timeout` is removed outright — its only consumer was the deleted legacy path.
- **Found and confirmed during implementation** (beyond the original scoping): only `agt->server[0]` is ever dialed under HTTPS, so the legacy multi-target retry loop in `w_agentd_keys_init()` was dead code. `try_enroll_to_server()` drops **all** its parameters, not just `network_interface` as originally planned — a deliberate, evidence-based deviation, approved during this work.
- **`shared/include/shared.h`: one added include, `openssl/types.h`, right after `windows.h`.** Deleting `enrollment_op.h` removed an accidental side effect it had on Windows builds: `windows.h` pulls in `wincrypt.h`, which `#define`s `X509_NAME` (and a few other X509/PKCS7/OCSP tokens) as plain integers, shadowing OpenSSL's own typedefs for any translation unit that already included an OpenSSL X.509 header earlier (`shared/src/generate_cert.c`'s case). OpenSSL's `openssl/types.h` carries an explicit, unconditional `#undef` for exactly this collision (tracked upstream as `openssl/openssl#9981`); `enrollment_op.h`'s own `<openssl/ssl.h>` include used to re-trigger it after `windows.h` ran, purely as a side effect. Re-including `openssl/types.h` directly restores that protection on purpose instead of depending on an unrelated module for it. Found by a real `winagent` CI failure on this PR and confirmed by reproducing the exact compile error locally with `x86_64-w64-mingw32-gcc` before and after the fix.

#### Automated tests

- Full `make -j TARGET=agent TEST=1` build: green, no compile/link errors.
- `https_client_unit_test` (gtest, C++ module): **376/376** (after rebasing onto 5.0.0's latest tip).
- `https_client_component_test` — new `EnrollComponentTest` suite, real curl/TLS against a fork-based fake manager: **6/6** (open mode, password mode with a correct/incorrect signature, forced application-error status pass-through, client cert + password coexisting over a real handshake). 44/47 of the full suite (the 3 unrelated pre-existing failures are tracked separately: 2 in #38442, 1 an unrelated cpp-httplib zstd-support gap in the test harness).
- cmocka, the six touched binaries: **194/194** — `test_enrollment` 18/18 (new), `test_https_client_bridge` 83/83, `test_client-config_https` 56/56, `test_agentd` 18/18, `test_start_agent` 4/4, `test_agent_config` 15/15 (adapted to the new by-value `agent_enrollment` struct after a rebase brought in a sibling PR's new test file).
- The #38438 HKDF/CMAC known-answer vector was independently recomputed in Python (`hmac`/`hashlib`/`cryptography`) before writing any C++, to avoid transcription mistakes in the signing recipe.

#### Manual end-to-end, real packaged builds (Linux .deb + Windows .msi)

Ahead of the manager-side #38438 landing, tested against `demo/mock_manager.py` (extended this session with a `/enroll` route, general zstd decompression — see below — and `demo_driver.c`'s new `enroll[:<password>]` mode). This is **not** the mocked unit-test path: it is the actual `wazuh-agent_5.0.0-0_amd64.deb` and `wazuh-agent_5.0.0-0_windows.msi` packages built by CI, installed on a Linux Docker container and a real Windows 10 machine respectively, both enrolling over the real LAN against `mock_manager.py` running in its own container.

Every case below was exercised on **both platforms**, with identical log messages (`enrollment.c` is shared code, no `#ifdef WIN32`):

| Case | Linux (`ossec.log`) | Windows (`ossec.log`) |
|---|---|---|
| Open mode | `INFO: Valid key received` | `INFO: Valid key received` |
| Password, correct | `INFO: Using password specified on file: etc/authd.pass` → `INFO: Valid key received` | `INFO: Using password specified on file: authd.pass` → `INFO: Valid key received` |
| Password, wrong | `ERROR: Enrollment rejected by the manager: invalid or missing authentication. CMAC mismatch (got …, want …)` | same message, different CMAC values |
| Forced 403 | `INFO: Enrollment is disabled on the manager. forced by --enroll-status` | same |
| Forced 409 | `ERROR: Enrollment rejected by the manager: duplicate agent. forced by --enroll-status` | same |
| Forced 500 | `ERROR: Enrollment failed with unexpected HTTP status 500. forced by --enroll-status` | same |
| Forced 503 | `ERROR: Enrollment failed with unexpected HTTP status 503. forced by --enroll-status` | same |

Corresponding `mock_manager.py` side (Linux run shown; Windows identical up to the agent name/id):

```
[mock] TLS  cipher=TLS_AES_256_GCM_SHA384 proto=TLSv1.3
[mock] POST /enroll     <- (54 B, zstd) body='...{"name":"nico-VirtualBox","version":"v5.0.0"}'
[mock]      /enroll     -> 200  minted id=101 name=nico-VirtualBox groups=None key_hash=None
...
[mock] POST /control    <- id=101 ts=1787190589 mac=4d4112ccdf.. auth OK (19 B) body='{"type":"shutdown"}'
[mock]      /control    -> 200  SHUTDOWN received - marking agent disconnected
...
[mock] POST /enroll     <- (54 B, zstd) body='...{"name":"nico-VirtualBox","version":"v5.0.0"}'
[mock]      /enroll     -> 401  (CMAC mismatch (got 26f46277a8f5.., want 3659c47acd72..))
```

On both platforms, the agent was additionally left running past `mock_manager.py`'s pre-existing key-rotation demo point (`ROTATE_KEY_AT`, simulating #37828) to exercise the real `401 → re-enroll → new key` recovery path end to end on a packaged binary, well beyond notify #7 (Windows example below):

```
2026/08/19 23:26:58 wazuh-agent: INFO: https_client: re-enrollment succeeded; reloading the signing key.
2026/08/19 23:26:58 wazuh-agent: INFO: Using password specified on file: authd.pass
2026/08/19 23:26:58 wazuh-agent: INFO: Valid key received
...
[mock]      /control    -> 200  NOTIFY #15 cfg=0c513176.. set=3f37dbd6.. tasks=[]
```

**Three real bugs were found and fixed this way — all three in `demo/mock_manager.py` itself, not in the agent:**

1. `mock_manager.py` never decompressed `Content-Encoding: zstd` on any endpoint. 5.0.0 recently flipped `https_compression_enabled`'s default to `true`, so **every** real-agent enroll attempt against the unpatched mock failed with a `400 malformed JSON` — the compressed body was fed straight to `json.loads()`. Fixed with a `zstd_decompress()` helper (shells out to the `zstd` CLI, same "reuse a real independent tool" idiom `cmac_hex()` already used for `openssl`), applied in `_handle_enroll()` and generally in `do_POST()` for the other endpoints. CMAC verification still runs over the raw (compressed) wire bytes, matching the real compress-before-sign design; only content parsing uses the decompressed body.
2. `_handle_enroll()` ignored the mock's own key-rotation feature: after `ROTATE_KEY_AT` notifies, `_verify()` starts requiring `ROTATED_KEY`, but `/enroll` kept minting the old `key_hex` — a real agent that re-enrolled after rotation got back the same already-invalid key, failed its next `/control`, and re-enrolled again, forever. Only visible against a real, long-running agent — no unit or component test runs enough notify cycles to hit it. Fixed by having `/enroll` return whichever key `_verify()` currently expects.
3. `_handle_enroll()` minted a brand-new numeric `id` on every call, ignoring `key_hash`. This surfaced (against a real running agent, not just by reading code) that the `https_client` module never refreshes its cached `agent_id` after a re-enroll — `hc_set_agent_key()` (the #37828 credential-rotation ABI) only ever updates the key material on a live handle, by design, with no parameter to also change the id. After the fix (`/enroll` now remembers `key_hash → id` and returns the same id for an already-known identity, matching what the real manager is expected to do per the contract's own `key_hash` semantics), traffic stayed healthy across a rotation — but the underlying architectural fact is now confirmed on a real binary: **the real #38438 manager must preserve an agent's existing numeric id when `key_hash` matches**, or the module will keep signing under a stale id until its next full restart. Not a regression this PR introduces — the same assumption already existed in #37828 — but worth flagging explicitly for whoever implements the manager side.

Also confirmed indirectly: the Windows `.msi` (built from the commit with this PR's `shared.h`/`openssl/types.h` fix, see above) installs and starts cleanly — the `generate_cert.c`/`X509_NAME` collision that broke `winagent` CI earlier in this PR's history does not resurface in a real build or at runtime.

#### Round 2 — real-package re-verification after the `hc_set_agent_identity()` fix and the second rebase

Bug #4 above was found and fixed *during* round 1, so round 1 never actually exercised the fix on a live packaged binary — only via gtest/component/cmocka. This round closes that gap: fresh `wazuh-agent_5.0.0-0_amd64_c5b6a7a.deb` / `wazuh-agent_5.0.0-0_windows_c5b6a7a.msi` were built by CI from this PR's current tip (`fa3a765850`, merged onto `main` as `c5b6a7a`) — i.e. **with** the `hc_set_agent_identity()` fix and **after** the second rebase (the `enrollment_retry_max`/`_delta` ramp merge). Both packages were freshly installed (Linux: clean Ubuntu 24.04 container; Windows: the existing test box, old agent uninstalled via `msiexec /x` first) and re-run through the full case matrix against an extended `mock_manager.py` (`--evidence-log`, dumping every request's full headers/body and every response's full body verbatim).

All six cases passed on both platforms. Representative evidence (Linux; Windows was identical modulo hostname/id):

Open enrollment:
```
EVIDENCE >> POST /enroll
        Authorization: WazuhEnroll 1787230907:222b9069b1cb08b7a38a09d0742f4691
        protocol-version: 1
        Content-Encoding: zstd
        [body, 54 B]: {"name":"nico-VirtualBox","version":"v5.0.0"}
EVIDENCE << 200
        [body, 96 B]: {"id": "101", "name": "nico-VirtualBox", "ip": "any", "key": "000102030405060708090a0b0c0d0e0f"}
```

Client cert + password coexisting (mutual TLS *and* the application-layer signature, neither substituting for the other):
```
[mock] mutual TLS: required, CA mtls_ca.pem
[mock] TLS  client cert presented: subject={'commonName': 'wazuh-agent-e2e-client'}
EVIDENCE >> POST /enroll
        Authorization: WazuhEnroll 1787231304:8bd4a8cb8aa7a50cb242e02713008591
EVIDENCE << 200  {"id": "101", ...}
```
(Confirmed identically on Windows: `TLS client cert presented: subject={'commonName': 'wazuh-agent-e2e-client'}` plus `INFO: Valid key received`.)

Forced-status mapping (`w_enrollment_process_response`), identical wording on both platforms:
```
403 -> INFO: Enrollment is disabled on the manager. forced by --enroll-status
409 -> ERROR: Enrollment rejected by the manager: duplicate agent. forced by --enroll-status
```

**Key rotation / reenroll — the fix, confirmed live for the first time.** The mock's demo rotation (`ROTATE_KEY_AT=7`) was also changed this round to mint a **genuinely new agent id** on the post-rotation reenroll (not reuse the pre-rotation id via `key_hash`), modeling a real manager that cannot correlate identity across a true credential rotation — exactly the scenario confirmed as a real agent bug earlier in this session. Linux (`nico-VirtualBox`, id `101`→`102`):
```
[mock] /control -> 200  NOTIFY #7 ... <- KEY ROTATED - old key now 401s, agent must re-enroll
EVIDENCE >> POST /stateless  Authorization: Wazuh 101:...
EVIDENCE << 401  {"error": "unauthorized"}   (CMAC mismatch ... - rotated?)
[second attempt also 401]

EVIDENCE >> POST /enroll
        Authorization: WazuhEnroll 1787231433:2dc7d644339a9a571fac1d62064b2b11
        [body]: {"name":"nico-VirtualBox","version":"v5.0.0","key_hash":"de18117daac6c78b451efc210d682c0950d32308"}
EVIDENCE << 200
        [body]: {"id": "102", "name": "nico-VirtualBox", "ip": "any", "key": "0f0e0d0c0b0a09080706050403020100"}

EVIDENCE >> POST /control   (the very next signed request)
        Authorization: Wazuh 102:1787231453:4d2946ec42a8e799d676e37304814269
EVIDENCE << 200  -> STARTUP accepted (v2 settings)
```
`ossec.log`: `WARNING: https_client: credential rejected (401); re-enrolling.` (x2) → `INFO: https_client: re-enrollment succeeded; reloading the signing identity.` `client.keys` on disk after: `102 nico-VirtualBox any 0f0e0d0c0b0a09080706050403020100` — both id and key persisted together. 14 further NOTIFYs over 2+ minutes, all signed as `id=102`, no further 401s or desync.

Windows (`DESKTOP-0SABLFF`, id `104`→`105` — sequentially minted past 101-103 from the earlier cases in the same run) reproduced the identical sequence: `NOTIFY #7` → two `401` (`CMAC mismatch ... - rotated?`) → `POST /enroll` with `key_hash` present → `200` minting `105` with the rotated key → the very next `/control` signed `Authorization: Wazuh 105:...`. `ossec.log`: `WARNING: https_client: credential rejected (401); re-enrolling.` → `INFO: https_client: re-enrollment succeeded; reloading the signing identity.` → `INFO: Manager reachable again. Resuming module event production.` → `INFO: Agent is now online. Process unlocked, continuing...`

This closes the caveat in bug #4 below: the fix is now confirmed against the real, packaged `wazuh-agentd`/`wazuh-agent.exe` binaries on both platforms, not only via gtest/component/cmocka.

#### Round 3 — real E2E against a real manager build of #38438 (not a mock), both platforms

Deployed a manager build from `enhancemenet/38438-https-register` (commit `fb69a1d`) onto the existing Vagrant AIO test box (`192.168.1.233`), manager-only in-place update (indexer/dashboard untouched):
```
$ sudo cat /var/wazuh-manager/VERSION.json
{ "version": "5.0.0", "stage": "rc1", "commit": "fb69a1d" }
$ sudo ss -tlnp | grep 1517
LISTEN 0  128  0.0.0.0:1517  0.0.0.0:*  users:(("wazuh-manager-r",pid=868458,fd=46))
$ curl -sk https://192.168.1.233:1517/
{"status":"ok","module":"remoted"}
```
Pointed fresh Linux and Windows agent installs (this PR's current tip, commit `5d8dfc0f51`) at it over the real LAN with `verification_mode=certificate` against the manager's own `remoted` cert and its real enrollment password (read from `/var/wazuh-manager/etc/authd.pass` on the box).

**Cross-repo finding, manager-side (#38438), not this PR: `/enroll` does not decompress `Content-Encoding: zstd`, unlike the manager's other endpoints.** With the agent's default `https_compression_enabled=1`, every enrollment attempt failed:
```
2026/08/20 15:59:35 wazuh-agentd: INFO: Using password specified on file: /var/ossec/etc/authd.pass
2026/08/20 15:59:35 wazuh-agentd: ERROR: Enrollment rejected by the manager: invalid request. Malformed JSON body
```
Isolated precisely with hand-signed manual requests (HKDF+CMAC replicated in Python, same algorithm as `EnrollSigner`, against the real manager) — three requests, only the `Content-Encoding` and target differ:
```
$ curl -sk -X POST https://192.168.1.233:1517/enroll -H "protocol-version: 1" \
    -H "Authorization: WazuhEnroll 1787241776:a395171c18f5098ed186a27d37261f65" \
    -H "Content-Encoding: zstd" --data-binary @enroll_body.zst
{"error":{"code":0,"message":"Malformed JSON body"}}          # 400 — same signed body, zstd-compressed

$ curl -sk -X POST https://192.168.1.233:1517/enroll -H "protocol-version: 1" \
    -H "Authorization: WazuhEnroll 1787241811:6e80662d1e78058a816b28f8b74866a1" \
    --data-binary @enroll_body_plain.json
{"id":"004","ip":"any","key":"8ac5c38987d1ad2d875e00b84e14f4531b32a49af08191b542b4d651a8e5342e","name":"probe-manual-uncompressed"}   # 200 — identical body, uncompressed

$ curl -sk -X POST https://192.168.1.233:1517/control -H "protocol-version: 1" \
    -H "Authorization: Wazuh 003:1787241751:e8c2e27b9f5d52d2268952406904a9f7" \
    -H "Content-Encoding: zstd" --data-binary @control_body.zst
{"agent":{"groups":["default"]},"cluster":{"name":"wazuh"},"limits":{...}}   # 200 — /control accepts zstd fine
```
This isolates the gap to the `/enroll` handler specifically, not the HTTP server's zstd support in general. With `agent.https_compression_enabled=0` as a workaround, both platforms enrolled cleanly:
```
2026/08/20 16:00:58 wazuh-agentd: INFO: https_client: starting.
2026/08/20 16:00:58 wazuh-agentd: INFO: Using password specified on file: /var/ossec/etc/authd.pass
2026/08/20 16:00:58 wazuh-agentd: INFO: Valid key received
```
```
2026/08/20 13:06:09 wazuh-agent: INFO: Using password specified on file: C:\Users\Escritorio\authd_real.pass
2026/08/20 13:06:09 wazuh-agent: INFO: Valid key received
```
and ran normal traffic (syscollector/SCA/rootcheck/FIM), confirmed from the manager's own `client.keys` and log:
```
$ sudo cat /var/wazuh-manager/etc/client.keys
003 nico-VirtualBox any 6a780e0fbae2c6d8151a0e774a610e3b6c3a76b310c93d5b06a8cd1324e8a3d9
005 DESKTOP-0SABLFF any 4e70d6a2c6a9c8f5099b0635de7722cc2fe378e43cf3067b219d3624b45becbd
2026/08/20 16:00:58 wazuh-manager-authd: INFO: Agent key generated for agent 'nico-VirtualBox' (requested locally)
```
Reported for #38438's owner to track — not fixed here, and the agent's compression default is not being changed to work around a manager-side gap.

#### Round 4 — the rest of `<enrollment>`'s options against the same real manager: `agent_name`, `groups`, `agent_address`, `use_source_ip`

Not covered by rounds 1-3 (all used the default hostname-derived name, no `groups`/`agent_address`/`use_source_ip` — only exercised before at the `test_enrollment.c` cmocka level, no real network). Verified each against the manager's own REST API (`GET /agents`), not just the agent's log:

- **`agent_name` + `agent_address`**: work end-to-end — `GET /agents` shows `"name":"e2e-custom-name","ip":"203.0.113.55"` exactly as configured.
- **`groups`**: works, but the group(s) must already exist on the manager (no autocreation) — confirmed both ways: unknown groups → `400 "Invalid Group(s) Name(s)"`, no key issued; after creating them via `POST /groups`, the identical request succeeds and `GET /agents` reports `"group":["group_a","group_b"]`.
- **`use_source_ip=yes` fails against this real build**: `400 {"error":{"code":0,"message":"Invalid field: ip"}}`. The agent sends `"ip":"src"` (the literal sentinel, per `w_enrollment_resolve_ip()`'s doc comment — "confirmed with the server team"), but this manager build rejects it. This confirms with real evidence a concern already on record as a plan FYI for #38438's owner: the remoted→authd bridge doesn't yet resolve the `"src"` token to the connection's actual source IP. Not a bug here — the agent does exactly what was agreed; flagging for #38438, same as the `/enroll` zstd gap above. **Fixed in the manager as of round 5 below** (commit `4511ed2`) — confirmed, including through a real agent, not just a manual probe.
- Side observation, manager-side (authd) behavior, not this PR's concern: a failed enroll with a fixed `agent_name`/`agent_address` still leaves an orphaned agent record on the manager, which the next identical attempt then collides with as a duplicate — had to clean these up via `DELETE /agents` between attempts to reach the happy path.

#### Round 5 — new manager + agent builds with both #38438 gaps fixed (zstd on `/enroll`, `"ip":"src"` resolution); full re-verification, both platforms

Manager: `enhancement/38465-...` branch tip merged onto main as commit `4511ed2` (manager built from the same merge commit as the agent packages this round — `#38438`/PR #38485 is now merged into the shared base). Manager-only in-place update on the same Vagrant box:
```
$ sudo cat /var/wazuh-manager/VERSION.json
{ "version": "5.0.0", "stage": "rc1", "commit": "4511ed2" }
```

**Both previously-reported manager-side gaps are fixed, confirmed first with hand-signed manual requests before touching any real agent:**
```
$ curl -sk -X POST https://192.168.1.233:1517/enroll -H "protocol-version: 1" \
    -H "Authorization: WazuhEnroll ...:9099166e88b4663de094b976939c5ac7" \
    -H "Content-Encoding: zstd" --data-binary @enroll_body_r5.zst
{"id":"013","ip":"any","key":"fa079ef6...","name":"probe-r5-zstd-check"}   # 200 — zstd now decompressed on /enroll

$ curl -sk -X POST https://192.168.1.233:1517/enroll -H "protocol-version: 1" \
    -H "Authorization: WazuhEnroll ...:fb7cd5227120c4fe96e5bb7055f2f25c" \
    --data-binary '{"name":"probe-r5-src-ip-check","version":"v5.0.0","ip":"src"}'
{"id":"014","ip":"192.168.1.41","key":"1304c11f...","name":"probe-r5-src-ip-check"}   # 200 — "src" resolved to the real connection IP, not rejected
```

**The critical re-verification: a completely stock, default-configuration agent (`https_compression_enabled=1`, no workaround) now enrolls successfully on the first try, both platforms** — this was the exact configuration that failed 100% of the time in round 3:
```
2026/08/20 23:15:39 wazuh-agentd: INFO: Started (pid: 83).
2026/08/20 23:15:39 wazuh-agentd: INFO: Using password specified on file: /var/ossec/etc/authd.pass
2026/08/20 23:15:39 wazuh-agentd: INFO: Valid key received
```
```
2026/08/20 20:21:16 wazuh-agent: INFO: Using password specified on file: C:\Users\Escritorio\authd_real.pass
2026/08/20 20:21:16 wazuh-agent: INFO: Valid key received
```
(Windows' first two attempts hit an unrelated `"duplicate agent. Duplicate name"` from a leftover agent record from an earlier round; deleted via the API and the very next retry succeeded — not a regression.)

**`use_source_ip=yes` re-verified through a real agent, not just the manual probe above**: enrolled clean, and `client.keys` recorded the *resolved* IP directly: `018 nico-VirtualBox 192.168.1.41 ...` — the manager handed back the real connection IP (`192.168.1.41`), not the literal `"src"` string, confirming the fix works through the actual `hc_enroll()`/`enrollment.c` code path end to end, not only via a hand-crafted request.

**Key rotation / reenroll re-confirmed on this fresh manager+agent pairing too**: deleted the live, actively-syncing Linux agent from the manager (`DELETE /agents`) while it kept running — no code change needed to trigger this, it's the same external-deletion mechanism found live in round 3. The agent picked up the resulting 401 and reenrolled to a new id automatically, with FIM/SCA/syscollector sync continuing normally right after, confirming the identity fix still holds with the new packages.

- `agentd` (the `agent` build target) — enrollment now goes exclusively over HTTPS/1517; the legacy TCP/1515 enrollment path no longer exists.
- No packaging/installer changes.

### Artifacts Affected

- `agentd` (the `agent` build target) — enrollment now goes exclusively over HTTPS/1517; the legacy TCP/1515 enrollment path no longer exists.
- No packaging/installer changes.

### Configuration Changes

- `<agent><enrollment>` — **kept**: `enabled`, `agent_name`, `groups`, `agent_address`, `use_source_ip`, `authorization_pass_path`, `delay_after_enrollment`.
- `<agent><enrollment>` — **removed** (recognized, logged at INFO, ignored — never rejected): `manager_address`, `port`, `interface_index`, `ssl_cipher`, `server_ca_path`, `agent_certificate_path`, `agent_key_path`.
- **Behavior change worth calling out in the release notes**: `<enrollment><manager_address>` can no longer point at a manager different from `<agent><server>` — enrollment always targets the same manager as the rest of the HTTPS traffic now.
- `internal_options.conf`: `agent.recv_timeout` removed (its only consumer was the deleted legacy protocol).
- Backward compatibility: a 4.x `ossec.conf` carrying the removed options parses without error after an in-place upgrade — INFO log only, never a startup failure.

### Documentation Updates

- `docs/ref/modules/client/configuration.md`: `<enrollment>` section rewritten to the surviving options, with a "removed options" note; both XML examples updated to drop `<manager_address>`/`<port>`.

### Tests Introduced

- `https_client/tests/unit/enrollSigner_test.cpp`, `enrollClient_test.cpp` (new) + additions to `hcInterface_test.cpp`/`moduleConfig_test.cpp`.
- `https_client/tests/component/enroll_component_test.cpp` (new) + a `/enroll` route added to the shared `fakeManager.hpp` fixture.
- `unit_tests/client-agent/test_enrollment.c` (new, 18 tests covering `w_enrollment_build_request`/`w_enrollment_process_response`), plus updates to `test_https_client_bridge.c`, `test_agentd.c`, `test_start_agent.c`, and `unit_tests/config/test_client-config_https.c`.
- Removed `unit_tests/shared/test_enrollment_op.c` (tested the now-deleted legacy protocol).
- `demo/mock_manager.py` + `demo/demo_driver.c` (manual E2E harness, doesn't gate CI): new `/enroll` route, a `--bind` flag for multi-machine testing, general zstd decompression, and a new `enroll[:<password>]` driver mode — see "Manual end-to-end" above for what this actually caught.

## Review Checklist

- [x] Code changes reviewed
- [ ] Meets requirements and/or definition of done — unit, component/RTR and manual E2E test layers are complete; the cross-repo integration test (`qa-integration-framework`'s `RemotedSimulator` needs a `/enroll` route) is **not** part of this PR, tracked separately.
- [ ] No unresolved dependencies with other issues — depends on #38438 (manager-side `/enroll`) actually shipping; this area also has a pre-existing, already-tracked, unrelated bug (#38442, `RescanRequester`/`/scan/vd`) that this PR does not fix.


jpcerrone:
Fixed the RTR Valgrind stall in `https_client` unit tests — four bare busy-wait barriers now yield, cutting the suite from ~17.9 min to 11.31 s

jpcerrone (approved):

PR #38489, opened 2026-08-20
Reduce wazuh-agentd resident memory and per-request signing cost

### LINKED ISSUE #38456: wazuh-agentd memory footprint regression in 5.0.0
## Description

`wazuh-agentd` resting memory usage roughly doubled between nightly 14/08/2026 and 19/08/2026 of Performance_Testing. Resident memory (RSS p95) increased by 34 to 41 percent, private memory (USS median) by 102 to 135 percent, and reserved virtual address space (VMS median) by about 362 MB. The increase reproduces on all four Linux agents, across both distributions (Ubuntu 24, Amazon Linux 2023) and both architectures (x86_64, aarch64), and it is isolated to `wazuh-agentd`: every sibling daemon on the same hosts is unchanged in the same runs.

This is not a leak: the leak guard (`growth_pct`) passes on every host and is negative on two of them. The footprint rises about 2 MB during the first five minutes and then holds a flat plateau for the remaining 70 minutes of the run, so what changed is the resting level, not the trend.

## Environment

| Field | Value |
| --- | --- |
| Wazuh version | 5.0.0 |
| Release stage | Latest |
| Jenkins job | `5.0.0/Performance_Testing` |
| Reference build (clean) | Collection window 2026-08-14 04:37 to 05:58 UTC |
| Affected build (reported) | Collection window 2026-08-19 05:37 to 07:27 UTC |
| Affected platforms | ubuntu-24-amd64, ubuntu-24-arm64, amazon-2023-amd64, amazon-2023-arm64 |
| Unaffected in the same run | wazuh-execd, wazuh-logcollector, wazuh-modulesd, wazuh-syscheckd |

## Graphs

Image
*The two builds do not overlap at any point of either run. Build 20 stays between 9.5 and 12.5 MB, build 23 between 16 and 18 MB, with the threshold in the gap between them.*

Image
*The eight failing assertions, with build 20's passing values beside them.*

Image
*Address space slightly more than doubles while mapped shared libraries grow only 27 percent. The growth is private.*

Image
*Two discrete steps rather than drift. Shaded bands are the two windows the change lands in.*

Image
*Every sibling daemon is within 2 percent, and syscheckd is lower. Only agentd moved.*

## Failing assertions

Thresholds are from `tests/test_non_functional/test_benchmark/baselines/agent_performance.yaml`, unchanged since 2026-07-17.

| Host | Rule | Threshold | Build 20 | Build 23 |
| --- | --- | --- | --- | --- |
| ubuntu-24-amd64 | `agentd_max_rss_p95` | 14 239 KB | 12 856 (pass) | **18 144 (fail)** |
| ubuntu-24-amd64 | `agentd_max_uss_p95` | 5 949 KB | 5 388 (pass) | **8 920 (fail)** |
| ubuntu-24-arm64 | `agentd_max_rss_p95` | 14 239 KB | 12 984 (pass) | **17 596 (fail)** |
| ubuntu-24-arm64 | `agentd_max_uss_p95` | 5 949 KB | 5 276 (pass) | **8 804 (fail)** |
| amazon-2023-amd64 | `agentd_max_rss_p95` | 14 239 KB | 12 792 (pass) | **17 904 (fail)** |
| amazon-2023-amd64 | `agentd_max_uss_p95` | 5 949 KB | 5 464 (pass) | **10 268 (fail)** |
| amazon-2023-arm64 | `agentd_max_rss_p95` | 14 239 KB | 13 020 (pass) | **17 424 (fail)** |
| amazon-2023-arm64 | `agentd_max_uss_p95` | 5 949 KB | 5 424 (pass) | **9 388 (fail)** |

## Magnitude of the increase, build 20 to build 23

All values in KB, computed with the same statistics the baseline suite uses.

| Host | RSS min | RSS p95 | USS median | VMS median |
| --- | --- | --- | --- | --- |
| ubuntu-24-amd64 | 10 660 → 15 104 (+41.7 %) | 12 856 → 18 144 (+41.1 %) | 3 808 → 8 020 (+110.6 %) | 330 284 → 692 388 (+109.6 %) |
| ubuntu-24-arm64 | 11 032 → 14 508 (+31.5 %) | 12 984 → 17 596 (+35.5 %) | 3 828 → 7 748 (+102.4 %) | 329 216 → 691 536 (+110.1 %) |
| amazon-2023-amd64 | 9 624 → 15 060 (+56.5 %) | 12 792 → 17 904 (+40.0 %) | 3 744 → 8 808 (+135.3 %) | 329 944 → 704 128 (+113.4 %) |
| amazon-2023-arm64 | 10 768 → 14 408 (+33.8 %) | 13 020 → 17 424 (+33.8 %) | 3 936 → 8 660 (+120.0 %) | 330 284 → 704 976 (+113.4 %) |

`RSS min` is effectively the cold-start footprint, the first sample after the daemon starts. It rose by 3.5 to 5.4 MB, which means the process is already larger before any load is applied. The threshold of 14 239 KB now sits below the value agentd reports in its first seconds of life on every one of the four hosts.

## Memory composition

Medians on ubuntu-24-amd64, which shows the shape of the change most clearly:

| Metric | Build 20 | Build 23 | Delta |
| --- | --- | --- | --- |
| VMS (virtual) | 330 284 | 692 388 | +362 104 KB (+109.6 %) |
| RSS (resident) | 11 084 | 17 244 | +6 160 KB (+55.6 %) |
| USS (private) | 3 808 | 8 020 | +4 212 KB (+110.6 %) |
| PSS (proportional) | 5 166 | 10 177 | +5 011 KB (+97.0 %) |
| Shared (RSS minus USS) | 7 276 | 9 224 | +1 948 KB (+26.8 %) |

The growth is concentrated in private, anonymous memory, not in mapped shared libraries. Reserving roughly 362 MB of additional address space to hold about 4 MB of additional private resident pages is characteristic of new threads rather than new data: on these hosts each pthread reserves an 8 MB stack, and each thread that allocates acquires its own 64 MB glibc malloc arena, of which only the touched pages become resident. The observed 362 MB divides cleanly into either unit. The same VMS increase appears on all four hosts (+362 MB on Ubuntu, +374 MB on Amazon Linux), so it is not host-specific.

Suggested starting point for investigation: any change to agentd's threading model, worker pool, or queue sizing that landed in the rc1 agent packages in the window identified below.

## Regression window

agentd on ubuntu-24-amd64, all values in KB. The same two steps reproduce on amazon-2023-amd64.

| Run date (UTC) | Build | RSS min | RSS p95 | USS p95 | Note |
| --- | --- | --- | --- | --- | --- |
| 2026-08-11 | 17 | 10 680 | 12 900 | 5 396 | |
| 2026-08-12 | 18 | 10 696 | 12 896 | 5 396 | |
| 2026-08-14 | 20 | 10 660 | 12 856 | 5 388 | last clean nightly |
| 2026-08-15 | 21 | 12 612 | 13 088 | 9 600 | step 1: private memory jumps |
| 2026-08-18 | 22 | 12 748 | 14 556 | 10 960 | |
| 2026-08-18 | 397 | 15 072 | 17 328 | 8 124 | step 2: resting RSS jumps |
| 2026-08-19 | 23 | 15 104 | 18 144 | 8 920 | reported build |

There are two distinct steps, so more than one commit may be involved:

- **Step 1** lands between the end of build 20's collection (2026-08-14 05:58 UTC) and the start of build 21's collection (2026-08-15 04:02 UTC). Private memory (USS p95) jumps from 5 388 to 9 600 KB while RSS p95 moves only slightly, and the cold-start floor rises about 2 MB.
- **Step 2** lands during 2026-08-18, between build 22 (04:08 UTC, RSS min 12 748) and build 397 (13:11 UTC, RSS min 15 072). The resting RSS floor rises a further 2.3 MB.

One caveat when reading that table: builds 21 and 22 ran a lighter load profile than the nightly performance profile, visible in their sibling daemons sitting near idle. For those two rows compare `RSS min` and `USS p95`, which are dominated by the resting footprint, rather than `RSS p95`, which depends on the load applied.

For longer-range context, agentd RSS p95 on this host was stable at approximately 12 800 to 12 900 KB across every archived run from 2026-05-22 through 2026-08-14 (43 archived runs were recomputed). The step is a discrete change, not drift.

## The increase is specific to agentd

Every monitored daemon on ubuntu-24-amd64, build 20 to build 23, under the same load profile in both runs. A toolchain, packaging, or shared-library change would have moved the siblings as well.

| Daemon | RSS p95 b20 | RSS p95 b23 | USS p95 b20 | USS p95 b23 | Change |
| --- | --- | --- | --- | --- | --- |
| **wazuh-agentd** | **12 856** | **18 144** | **5 388** | **8 920** | **+41 % / +66 %** |
| wazuh-execd | 5 500 | 5 608 | 2 256 | 2 312 | flat |
| wazuh-logcollector | 16 080 | 16 244 | 3 848 | 3 896 | flat |
| wazuh-modulesd | 41 612 | 42 488 | 28 636 | 28 692 | flat |
| wazuh-syscheckd | 26 416 | 24 252 | 15 192 | 13 092 | lower |

## This is not a memory leak

| Host | `growth_pct` build 20 | `growth_pct` build 23 | Threshold |
| --- | --- | --- | --- |
| ubuntu-24-amd64 | -6.6 | 9.7 | < 35 |
| ubuntu-24-arm64 | -5.5 | 9.6 | < 35 |
| amazon-2023-amd64 | -17.6 | -3.6 | < 35 |
| amazon-2023-arm64 | -6.6 | -1.8 | < 35 |

`agentd_memory_growth_pct` passes on every host in build 23, and is negative on the two Amazon Linux agents. Inspecting the raw series on ubuntu-24-amd64 confirms the shape: RSS starts at 15 104 KB, rises to about 17 200 KB within the first five minutes, then stays between 17 200 and 18 200 KB for the remaining 70 minutes. The daemon reaches a higher steady state and stays there.

## Impact

The Linux agents used in this suite are t3.small and t4g.small instances with 1.86 GB of RAM, and the reported free memory during the run is already around 0.11 GB. A permanent 5 to 7 MB increase in a single daemon is not fatal there, but it is a 34 to 41 percent regression in a metric that had been stable for three months, and it applies to every Linux agent deployment rather than to a specific configuration.

The equivalent macOS rule (`agentd_macos_max_rss_p95`) still carries the original loose 102 400 KB threshold and therefore passes, but macOS shows the same elevated level in build 23 (RSS p95 15 145 KB), so the platform is masking the behaviour rather than being exempt from it.

## Evidences

[observability_data.zip](https://github.com/user-attachments/files/31217422/observability_data.19.zip)

anromerom:
## Description

Related to #38456

`wazuh-agentd`'s resting memory footprint roughly doubled in 5.0.0 after the agent HTTPS transport landed. Investigation traced the increase to the cost of running OpenSSL and libcurl inside the agent process, most of which is inherent to the new transport: newly executed library pages, ten threads instead of three, and a small live working set.

This pull request addresses one recoverable part of that cost. It recovers roughly a sixth of the reported resident increase, so it does not bring the agent under the current baseline threshold; that remains a separate discussion on the issue.

## Proposed Changes

- Cap `wazuh-agentd`'s glibc malloc arena count with `mallopt(M_ARENA_MAX, 2)` in `main()`.

glibc gives each allocating thread its own 64 MiB arena, and the transport runs several. Most of that is reserved address space rather than committed memory, and it is what takes agentd past 690 MB of virtual size, but each arena also carries its own top chunk and bins, which is resident memory never handed back.

The call is placed before any thread is created, since extra arenas are only spawned when a thread finds every existing one locked. A limit of 2 rather than 1 leaves a second arena available so the transport's threads do not all serialise on a single allocator lock. Compiled in only for glibc, so the Windows and macOS agents are unaffected.

A second change, caching the CMAC implementation instead of fetching it per signature, was prepared and then dropped from this pull request. Fetching per signature is a documented OpenSSL anti-pattern and it is worth doing on its own merits, but its memory effect measured as approximately zero and it is being held back while the agentd and enrollment integration test failures on this branch remain unexplained.

### Results and Evidence

Two packages built by the same builder, differing only in the agent binaries. `libwazuhext.so`, `wazuh-modulesd` and `wazuh-logcollector` are byte identical between them, and `mallopt` is referenced only in the patched `wazuh-agentd`. Both arms ran from the package's own shipped `internal_options.conf` with compression at its default, under the same event load, alternating A,B,A,B so host drift cannot be mistaken for a package effect. Medians over 12 minute samples after a 7 minute settle, 288 samples per arm.

| Metric | Baseline | Patched | Saved | % of baseline |
| --- | --- | --- | --- | --- |
| USS (private) | 10.32 MB | 9.27 MB | **1.05 MB** | **10.2 %** |
| RSS (resident) | 18.52 MB | 17.60 MB | 0.92 MB | 5.0 % |
| VMS (virtual) | 675.67 MB | 165.51 MB | 510.16 MB | 75.5 % |

The measured package also carried two changes that have since been dropped, which together accounted for about 0.32 MB of a 1.37 MB total. The figures above subtract that, so they are this change's isolated contribution rather than a direct reading. A re-measurement against a package built from this branch as it now stands would replace the arithmetic with a measurement.

Above the noise floor: per-repetition USS medians differed by 96 KB within the baseline arm and 132 KB within the patched arm, so the arms do not overlap at any point and the saving is an order of magnitude larger than the within-arm variation.

What motivated the change: `malloc_info()` on a plateaued agent reported five arenas holding 1 912 KB in total, of which only 887 KB was live. Capping the count was measured separately at 1 328 KB of resident memory and 85 % of the virtual size with `MALLOC_ARENA_MAX=1`.

The virtual size result deserves a caveat rather than celebration. 510 MB sounds dramatic, but that memory was reserved and never committed, so it was never denied to anything else on the host. It matters because it removes a figure that reads as alarming in monitoring, not because it frees usable memory.

Set against the regression this addresses: the reported increase was +5 288 KB on RSS p95 and +4 212 KB on private memory, so this recovers roughly a sixth of the former and a quarter of the latter.

  - [x] Linux: agent DEB and RPM build jobs pass on this branch
  - [x] Windows: Windows agent binaries and MSI jobs pass; the change is compiled out there
  - [x] MAC OS X: macOS compilation test and unit tests pass; the change is compiled out there
- [x] Log syntax and correct language review: N/A, no log messages added or changed

- Memory tests for Linux
  - [ ] Coverity: not run locally; the scan-build job runs on this branch
  - [x] Valgrind (memcheck and descriptor leaks check): massif was used to profile the agent's heap during the investigation; memcheck was not run, and this change alters no allocation logic
  - [ ] AddressSanitizer: not run; the change sets an allocator limit and introduces no new memory access
- Memory tests for Windows
  - [ ] Coverity: N/A, compiled out on Windows
  - [ ] UMDH: N/A, compiled out on Windows
- Memory tests for macOS
  - [ ] Leaks: N/A, compiled out on macOS
  - [ ] AddressSanitizer: N/A, compiled out on macOS

  - [ ] Added unit testing files ".ini": N/A
  - [ ] `runtests.py` executed without errors: N/A

  - [ ] Test run in parallel: N/A, the engine is not touched
  - [ ] ASAN for test (utest/ctest): N/A
  - [ ] TSAN for test and wazuh-engine: N/A

- Wazuh server API/Framework
  - [ ] Run API Integration Tests: N/A, agent-side change only

Additional measurement specific to this change: a standalone six-thread allocation churn test confirms that capping arenas reduces virtual size without increasing resident memory (17.3 MB uncapped against 17.6 MB capped), and a three minute run of the agent under a rejected-enrollment retry loop showed flat resident memory, no CPU load and no thread growth.

### Artifacts Affected

- `wazuh-agentd` on Linux.
- No change to the Windows or macOS agents: the call is compiled in only for glibc.
- No packaging or default configuration changes.

### Configuration Changes

N/A. No new settings, no changes to defaults, and no behaviour visible to users or to the manager.

### Documentation Updates

N/A.

### Tests Introduced

N/A. The change sets an allocator limit at startup and alters no logic, so there is no behaviour to assert on. Its effect is a reduction in resident memory and virtual size, which the performance suite measures rather than a unit test.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/src/client-agent/src/main.c
+++ b/src/client-agent/src/main.c
@@ -14,6 +14,10 @@
 #include "agentd.h"
 #include <getopt.h>
 
+#if defined(__GLIBC__)
+#include <malloc.h>
+#endif
+
 #ifndef ARGV0
 #define ARGV0 "wazuh-agentd"
 #endif
@@ -61,6 +65,24 @@ int main(int argc, char **argv)
     gid_t gid;
 
     run_foreground = 0;
+
+#if defined(__GLIBC__)
+    /* Cap the number of malloc arenas.
+     *
+     * The HTTPS transport runs several allocating threads, and glibc answers
+     * that by giving each one its own 64 MiB arena. Those arenas are reserved
+     * rather than committed, so most of the cost is address space, and it is
+     * what takes this process past 690 MB of virtual size. But each arena also
+     * carries its own top chunk and bins, which is real resident memory that is
+     * never handed back. Capping the count trades a little allocator contention
+     * for that overhead.
+     *
+     * Set here, before any thread is created: extra arenas are only spawned
+     * when a thread finds every existing one locked, so the limit has to be in
+     * place before the transport's threads start. Allocations already made on
+     * this thread came from the main arena and are unaffected. */
+    mallopt(M_ARENA_MAX, 2);
+#endif
 
     /* Set the name */
     OS_SetName(ARGV0);


anromerom:
> [!CAUTION] 
> Closing this PR as the branch had to be renamed. The old branch name contained the string `wazuh-agentd`, which made the Linux `agentd` and `enrollment` integration tests kill their own pytest session: the framework stops a daemon by killing every process whose command line contains the daemon's name, and the branch name travels in the `sudo` wrapper's argv. Same commit, same content.

PR #38493, opened 2026-08-20
Reject symlinks when reading and staging files during agent upgrade

### LINKED ISSUE #38327: Agent upgrade-staging paths follow symlinks when reading
|Component|Install type|
|---|---|
Agent Update | Agent

## Description

wazuh/wazuh#38200 hardened the **write** side of the agent staging paths: `wm_agent_upgrade_com_open()` and `wcom_uncompress()` now go through `w_fopen_nofollow()`, which refuses a symlink, a hard link, a FIFO, a device or a directory at the target path and truncates only after the descriptor has been vetted. Two sites in the same flow were left following links, both reachable over the same remote request path (`src/client-agent/request.c`) as the ones already fixed.

**1. `wcom_uncompress()` follows symlinks on the read side** — `src/os_execd/wcom.c:156`

```c
if (fsource = gzopen(final_source, "rb"), !fsource) {
```

`final_source` is jailed by name only. A symlink planted at `var/incoming/<source>` is followed, and since zlib reads non-gzip input transparently in `"rb"` mode, the link may point at any readable file, not only at an archive. The contents are then written into `final_target`, which after #38200 is confined to a genuine regular file inside `INCOMING_DIR`. The resulting primitive is therefore "copy an arbitrary file into `var/incoming`" rather than an arbitrary write. There is no `wcom` command that returns a file's contents, so reading it back needs a further step, which is why this is filed as low severity rather than as an information disclosure.

**2. `_uncompress()` writes to a fully predictable path in `TMP_DIR`** — `src/wazuh_modules/agent_upgrade/agent/wm_agent_upgrade_com.c:509-517`

```c
memcpy(dest + length, TEMPLATE, sizeof(TEMPLATE));   // TEMPLATE is ".mg.XXXXXX"
...
if (ftarget = wfopen(dest, "wb"), !ftarget) {
```

The `XXXXXX` is never expanded: unlike `_unsign()`, which calls `mkstemp()` on its own destination a few lines above (`:464`), `_uncompress()` appends the template and opens the result directly. The target is literally `tmp/<package>.mg.XXXXXX`, so it is not merely guessable, it is known in advance, and `wfopen()` follows a symlink placed there. This is the same class of defect #38200 fixed, in the one staging directory that PR did not touch.

> [!NOTE]
> Not a vulnerability, filed as hardening, on the same footing as #38200: both require the ability to place a link inside the agent's staging directories, and the commands themselves arrive from the manager. Nothing legitimate ever puts a link there, so refusing one costs nothing.

## Suggested fix

Both sites can reuse what #38200 already introduced.

- Add a read-side counterpart to `w_fopen_nofollow()` in `src/shared/file_op.c` — same directory-relative open with `O_NOFOLLOW`, same `fstat()` vetting, opened read-only — and use it for the `gzopen()` in `wcom_uncompress()`. zlib accepts a descriptor through `gzdopen()`, so the vetted descriptor can be handed over directly instead of reopening by path.
- Have `_uncompress()` either call `mkstemp()` on its destination, as `_unsign()` already does, or write through `w_fopen_nofollow()` against `TMP_DIR`. Doing both is cheap and removes the predictable name as well as the link follow.

Worth covering in the same change: `test_file_op.c` gained real-file-system tests for the write-side helper in #38200 (symlink, hard link, dangling symlink, FIFO, directory, invalid names); the read-side helper should get the equivalent set.

### Related, but a different size of job

`UnmergeFiles()` (`src/shared/file_op.c:735`) is adjacent but should not be folded in here. It does create its temporary file securely with `mkstemp_ex()` (`:822`), so the name is not predictable, but it then reopens that path with `wfopen()` (`:830`), and `mkdir_ex()` creates intermediate directories without checking whether a path component is a symlink. Closing that properly means component-wise resolution rather than swapping one helper for another, and it affects every caller of `UnmergeFiles()`, not just the upgrade flow. Better tracked on its own.

`wm_agent_upgrade_com_sha1()` (`:334`) also reopens by path with `OS_SHA1_File()` and follows symlinks, but its underlying problem is that the module identifies the open file by comparing path strings instead of holding on to the descriptor — `write` and `close` do the same. Fixing that by keeping the descriptor closes the symlink follow as a side effect, so it belongs with that refactor rather than here.

hernanvalenzuela:
## Description

Two locations in the agent upgrade flow open files inside writable staging directories (`var/incoming`, `tmp`) without checking whether the target is a symlink first:

1. **`wcom_uncompress()`** (`src/os_execd/wcom.c`) opens `final_source` — a name jailed under `var/incoming` — with a plain `gzopen()`. If a symlink with that name is planted in `var/incoming`, `gzopen()` follows it and discloses the contents of whatever it points to. The write side of this same function was already hardened for the same issue in #38200 (`w_fopen_nofollow()`); the read side was left uncovered.

2. **`_uncompress()`** (`src/wazuh_modules/agent_upgrade/agent/wm_agent_upgrade_com.c`) builds a temporary destination path from a template (`tmp/<package>.mg.XXXXXX`) but — unlike the sibling function `_unsign()`, which calls `mkstemp()` — never expands it into a random name. The path is therefore predictable from the package name alone, and gets opened with a plain `wfopen()`. A symlink planted in advance at that exact path is followed, letting an attacker overwrite an arbitrary file with the decompressed package contents.

Both require the attacker to already have write access to the staging directory, so this is a hardening fix rather than a critical remote vulnerability — the same follow-up scope #38200 left open.

## Proposed Changes

- Added `w_gzopen_nofollow()` in `src/shared/file_op.c` (declared in `file_op.h`), the read-side counterpart of `w_fopen_nofollow()`: opens with `O_NOFOLLOW` on POSIX (via `openat()` on a directory descriptor) and rejects reparse points on Windows, then hands the vetted descriptor to `gzdopen()`.
- `wcom_uncompress()` now uses `w_gzopen_nofollow(INCOMING_DIR, source, "rb")` instead of `gzopen()`.
- `_uncompress()` now expands the temp-file template with `mkstemp()` on POSIX — reusing the descriptor it returns directly via `fdopen()`, rather than closing it and reopening by name — or `_mktemp_s()` + `w_fopen_nofollow()` on Windows, instead of opening the unexpanded, predictable path with `wfopen()`. The read side of the same function was switched to `w_gzopen_nofollow()` too, for consistency.

### PoC 1 — `wcom_uncompress()` read side (symlink in `var/incoming` disclosing an outside file)

Setup: `sandbox/var/incoming/evil` is a symlink to a real `.gz` file living outside the jail, containing a secret string.

--- PRE-FIX pattern: gzopen(final_source, "rb") ---
gzopen() SUCCEEDED (followed the symlink). Bytes read: 43 Disclosed content from OUTSIDE the jail: "TOP-SECRET-AGENT-DATA-LAPTOP-J631ENT0-29721"

--- POST-FIX pattern: w_gzopen_nofollow(INCOMING_DIR, "evil", "rb") ---
w_gzopen_nofollow() correctly REJECTED the symlink. errno=40 (Too many levels of symbolic links) [ELOOP as expected]

### PoC 2 — `_uncompress()` write side (symlink at the predictable, unexpanded `tmp/<package>.mg.XXXXXX` path)

Setup: `sandbox/tmp/package.mg.XXXXXX` (the literal, never-expanded template) is a symlink to a "victim" file.

--- PRE-FIX pattern: wfopen(dest, "wb") on the never-expanded path ---
wfopen() SUCCEEDED (followed the symlink) and wrote attacker bytes through it.
victim file content is now: "ATTACKER-CONTROLLED-DECOMPRESSED-BYTES"

--- POST-FIX pattern: mkstemp(dest) expands the template before anything is opened ---
mkstemp() expanded the template to a fresh, unpredictable path: sandbox/tmp/package.mg.0NMAgF Write landed in the new randomized file, never touching the symlink's literal path.
victim file content is now: "IMPORTANT-CONFIG-DO-NOT-TOUCH"

Both harnesses use the real, modified functions extracted verbatim (via `sed`) from `file_op.c`, not reimplementations.

  - [ ] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

### Artifacts Affected

N-A

### Configuration Changes

N-A

### Tests Introduced

N-A

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

### DIFF
--- a/src/shared/file_op.c
+++ b/src/shared/file_op.c
@@ -2602,6 +2602,138 @@ static int w_win32_to_errno(DWORD error) {
 #endif
 
 
+#ifdef WIN32
+/**
+ * Opens @p filename inside @p basedir on Windows without following symlinks or junctions, and vets the
+ * resulting handle as a lone regular file — rejecting reparse points, directories, and NTFS hard links
+ * (which need no privilege to create and carry neither attribute above) — before handing it back.
+ * Shared by w_fopen_nofollow() and w_gzopen_nofollow() so a future hardening fix only has to be applied
+ * once instead of needing to be kept in sync across both.
+ *
+ * @param basedir Base directory holding the file.
+ * @param filename Bare file name inside @p basedir.
+ * @param access GENERIC_READ or GENERIC_WRITE.
+ * @param disposition OPEN_EXISTING for reading, or OPEN_ALWAYS for writing. OPEN_ALWAYS never truncates
+ *                     by itself: the caller must only truncate after the handle has been vetted here,
+ *                     since truncating up front would destroy the target of a hard link before it could
+ *                     be told apart from a regular file.
+ * @return A vetted handle on success, or INVALID_HANDLE_VALUE on error (sets errno).
+ */
+static HANDLE w_createfile_nofollow_vetted(const char * basedir, const char * filename, DWORD access, DWORD disposition) {
+    char final_path[PATH_MAX + 1];
+    BY_HANDLE_FILE_INFORMATION file_info;
+    HANDLE hFile;
+
+    if (snprintf(final_path, sizeof(final_path), "%s\\%s", basedir, filename) > PATH_MAX) {
+        errno = ENAMETOOLONG;
+        return INVALID_HANDLE_VALUE;
+    }
+
+    // FILE_FLAG_OPEN_REPARSE_POINT makes the handle refer to a symlink or junction itself instead of to
+    // its target, so the target is never opened nor modified.
+    hFile = wCreateFile(final_path, access, FILE_SHARE_DELETE | FILE_SHARE_READ | FILE_SHARE_WRITE,
+                        NULL, disposition, FILE_ATTRIBUTE_NORMAL | FILE_FLAG_OPEN_REPARSE_POINT, NULL);
+
+    if (hFile == INVALID_HANDLE_VALUE) {
+        errno = w_win32_to_errno(GetLastError());
+        return INVALID_HANDLE_VALUE;
+    }
+
+    if (!GetFileInformationByHandle(hFile, &file_info)) {
+        errno = w_win32_to_errno(GetLastError());
+        CloseHandle(hFile);
+        return INVALID_HANDLE_VALUE;
+    }
+
+    if (file_info.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT) {
+        CloseHandle(hFile);
+        errno = ELOOP;
+        return INVALID_HANDLE_VALUE;
+    }
+
+    if (file_info.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY) {
+        CloseHandle(hFile);
+        errno = EISDIR;
+        return INVALID_HANDLE_VALUE;
+    }
+
+    if (file_info.nNumberOfLinks != 1) {
+        CloseHandle(hFile);
+        errno = EMLINK;
+        return INVALID_HANDLE_VALUE;
+    }
+
+    return hFile;
+}
+#else
+/**
+ * Opens @p filename inside @p basedir without following symlinks, and vets the resulting descriptor as
+ * a lone regular file — rejecting hard links, FIFOs, devices, and directories — before handing it back.
+ * Shared by w_fopen_nofollow() and w_gzopen_nofollow() so a future hardening fix only has to be applied
+ * once instead of needing to be kept in sync across both.
+ *
+ * @param basedir Base directory holding the file.
+ * @param filename Bare file name inside @p basedir.
+ * @param oflags open()/openat() flags; must include O_NOFOLLOW | O_CLOEXEC | O_NONBLOCK (the latter to
+ *               keep a FIFO from blocking the open) on top of whichever of O_RDONLY/O_WRONLY/O_CREAT the
+ *               caller needs. Deliberately never includes O_TRUNC: truncating at open time would destroy
+ *               the target before anything about it can be checked, which is precisely how a hard link
+ *               slips through — it is a regular file, so no file type test can tell it apart. A caller
+ *               that needs the file truncated must do so only after this returns a vetted descriptor.
+ * @param mode Permission bits, used only when oflags includes O_CREAT.
+ * @return A vetted file descriptor, with O_NONBLOCK already cleared, on success; -1 on error (sets errno).
+ */
+static int w_openat_nofollow_vetted(const char * basedir, const char * filename, int oflags, mode_t mode) {
+    struct stat statbuf;
+    int dirfd;
+    int fd;
+    int saved_errno;
+    int flags;
+
+    if (dirfd = open(basedir, O_RDONLY | O_DIRECTORY | O_CLOEXEC), dirfd < 0) {
+        return -1;
+    }
+
+    fd = openat(dirfd, filename, oflags, mode);
+    saved_errno = errno;
+    close(dirfd);
+
+    if (fd < 0) {
+        errno = saved_errno;
+        return -1;
+    }
+
+    // Rules out anything O_NOFOLLOW does not, such as a block or character device.
+    if (fstat(fd, &statbuf) < 0) {
+        saved_errno = errno;
+        close(fd);
+        errno = saved_errno;
+        return -1;
+    }
+
+    if (!S_ISREG(statbuf.st_mode)) {
+        close(fd);
+        errno = EINVAL;
+        return -1;
+    }
+
+    // A hard link is a regular file by every other measure; the link count is what gives it away.
+    if (statbuf.st_nlink != 1) {
+        close(fd);
+        errno = EMLINK;
+        return -1;
+    }
+
+    // O_NONBLOCK only mattered while checking for a FIFO above; clear it so reads/writes behave normally.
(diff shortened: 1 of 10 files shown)


src/wazuh_modules/agent_upgrade/agent/wm_agent_upgrade_com.c
```diff
@@ -509,16 +509,55 @@ STATIC int _uncompress(const char * source, const char *package, char dest[PATH_
         memcpy(dest + length, TEMPLATE, sizeof(TEMPLATE));
     }
 
-    if (fsource = gzopen(source, "rb"), !fsource) {
+    // Not gzopen(): a symlink left at source would be followed, disclosing whatever it points to.
+    if (fsource = w_gzopen_nofollow(TMP_DIR, source + strlen(TMP_DIR) + 1, "rb"), !fsource) {
```
lchico: `_uncompress()` now calls `w_gzopen_nofollow()` and, on POSIX, `mkstemp()` + `fdopen()` instead of `gzopen()`/`wfopen()`, but `src/unit_tests/wazuh_modules/agent_upgrade/CMakeLists.txt` still only wraps `gzopen`/`wfopen`/`w_fopen_nofollow` (not `w_gzopen_nofollow`), and `test_uncompress_gzopen_fail`/`test_uncompress_fopen_fail`/`test_uncompress_fwrite_fail`/`test_uncompress_gzread_fail`/`test_uncompress_success` in `test_wm_agent_upgrade_com.c` still queue `__wrap_gzopen`/`__wrap_wfopen` expectations that are never consumed, and never mock the new `mkstemp` call. Running this test binary as-is should fail/crash on leftover mock values and an unmocked `mkstemp`. Can these tests be updated to mock the new call sequence before merge?

lchico: `w_gzopen_nofollow(TMP_DIR, source + strlen(TMP_DIR) + 1, "rb")` (and the symmetric case for `dest` at line 552) advances the pointer past the literal `"tmp/"` prefix without checking that `source`/`dest` actually starts with it. If it's ever shorter than `strlen(TMP_DIR)+1` bytes or doesn't start with `"tmp/"` — which is already true of this file's own pre-existing unit test call, `_uncompress("compressed_test", package, merged)` — this walks past the string's NUL terminator before handing the result to `w_is_bare_filename`/`openat`. Should this validate the prefix (or use a helper that doesn't assume it) before doing the pointer arithmetic?

src/shared/file_op.c
```diff
@@ -2755,6 +2755,134 @@ FILE * w_fopen_nofollow(const char * basedir, const char * filename, const char
 }
 
 
+gzFile w_gzopen_nofollow(const char * basedir, const char * filename, const char * mode) {
+    if (!basedir || !mode || (strcmp(mode, "r") && strcmp(mode, "rb")) || !w_is_bare_filename(filename)) {
```
lchico: `w_gzopen_nofollow()` duplicates almost the entire open+fstat+`S_ISREG`+nlink-count vetting sequence of `w_fopen_nofollow()` just above it (both the POSIX and Windows branches). Worth factoring the shared vetting logic into a common helper so a future hardening fix doesn't need to be applied twice — missing one copy would silently reopen the symlink-following bug this PR fixes, in just one of the two paths.

lchico (changes requested):

src/wazuh_modules/agent_upgrade/agent/wm_agent_upgrade_com.c
```diff
@@ -509,16 +529,59 @@ STATIC int _uncompress(const char * source, const char *package, char dest[PATH_
...
+    {
+        int fd;
+
+        if (fd = mkstemp(dest), fd < 0) {
+            gzclose(fsource);
+            mterror(WM_AGENT_UPGRADE_LOGTAG, WM_UPGRADE_COMPRESSED_FILE_ERROR, "uncompress()");
+            return -1;
+        }
+
+        if (chmod(dest, 0640) < 0) {
```
cborla: chmod(dest, 0640) here operates by path, after mkstemp() has already created and opened dest. Between those two calls there's a narrow window where the file could be unlinked and develreplaced with a symlink, causing chmod() to follow it and change an unrelated target's mode. Since fd is already open on the exact file mkstemp() created, fchmod(fd, 0640) avoids the develsecond name-based lookup entirely:


```suggestion
        if (fchmod(fd, 0640) < 0) {
```

(Same fix would apply to the pre-existing chmod(dest, 0640) in _unsign() at line 468, if that's in scope.)

cborla (changes requested):

lchico (approved):
LGTM!
- [x]  Code changes reviewed

src/wazuh_modules/agent_upgrade/agent/wm_agent_upgrade_com.c
hernanvalenzuela: done

cborla (approved):
LGTM
- [x] Code changes reviewed

Issue #29885, opened 2025-05-26
FIM on Windows incorrectly ignores files with German umlauts (e.g., ä, ö, ü) due to "non-UTF8" error despite PR #26289 in v4.12.x


etwas7B:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 4.12.0 | Wazuh Manager / Agent | Agent | Package | Windows Server 2016/2022 |

### Problem Description:
The Wazuh agent's File Integrity Monitoring (FIM) on Windows systems is incorrectly ignoring files and directories that contain German umlauts (e.g., ä, ö, ü, ß) and possibly other non-ASCII characters in their names. The agent log (ossec.log) shows a WARNING: (6955): Ignoring file ... due to unsupported name (non-UTF8). message for these files.

This issue persists even though PR #26289, which aimed to fix UTF-8 handling in FIM, has been merged and is understood to be included in the agent version mentioned above. As noted in recent comments on PR #26289, other users also appear to be experiencing this persistent problem.

### Steps to Reproduce:

1. Install Wazuh agent v4.12.x on a Windows Server system (e.g., Windows Server 2016 or 2022).
2. Configure FIM in agent.conf to monitor a directory. For example:
```
<syscheck>
  <directories check_all="yes" alert_new_files="yes" whodata="yes" recursive="yes" realtime="yes">E:\test_fim_umlauts</directories>
</syscheck>
```
3. Within the monitored directory (E:\test_fim_umlauts), create a file or directory with German umlauts in its name. Examples:
```
E:\test_fim_umlauts\Prüfungsunterlagen.docx
E:\test_fim_umlauts\Anleitungen für Geräte.pdf
E:\test_fim_umlauts\Geschäftsplan So Müsste Es Sein.xlsx
```
4. Observe the Wazuh agent log (ossec.log).

### Actual Behavior:
The Wazuh agent log shows a warning (error code 6955), and the file/directory is ignored by FIM. Example log entries previously observed:

`2025/05/23 13:15:50 wazuh-agent[10920] create_db.c:648 at fim_checker(): WARNING: (6955): Ignoring file 'e:\test_fim_umlauts\Prüfungsunterlagen.docx' due to unsupported name (non-UTF8).`

### Impact:
[...]

Issue #38146, opened 2026-08-03
Windows Syscollector intermittently stores corrupted package names and versions


suletom:
# Windows Syscollector intermittently stores corrupted package names and versions

| Wazuh version | Component | Install type | Install method | Platform |
|---|---|---|---|---|
| 4.14.5 | Syscollector -- Windows package inventory | Agent | Packages | Microsoft Windows 11 Pro 10.0.26200.8875 |
  -----------------------------------------------------------------------------

## Description

We are observing intermittent corruption in package inventory records
collected by Wazuh agents running on Windows.

This report provides additional investigation, historical samples, a
reproducible environment, and a possible source-code explanation for the
behaviour previously reported in:

-   #30180 -- **Syscollector Windows Package Name Corruption**

The corruption is not limited to trailing characters in the package
name. Our collected samples show several different forms of corrupted
package metadata:

-   random trailing bytes appended to an otherwise valid package name;
-   package names containing clearly invalid or memory-like data;
-   partially corrupted package names;
-   missing package versions;
-   package versions containing clearly invalid or memory-like data.

The corruption has been observed intermittently over approximately six
months.

In the dataset collected during this period, the affected records
originated from Windows agents. We did not observe the same corruption
pattern in package records from Linux agents.

## Example

Expected package name:

``` text
Security Update for Microsoft PowerPoint 2010 (KB4504702) 64-Bit Edition
```

Collected package record:

``` json
{
  "agent": {
    "id": "215",
    "version": "v4.14.5"
  },
  "package": {
    "architecture": "x86_64",
    "installed": "2026-04-09T05:13:59.000Z",
    "name": "Security Update for Microsoft PowerPoint 2010 (KB4504702) 64-Bit Edition~±ű\\u0004,Y",
    "size": 0,
    "type": "win",
    "vendor": "Microsoft"
  }
}
```

## Investigation

-   The Windows Registry contains the correct values.
-   The corrupted data is already present in the local Wazuh agent
    inventory database.
-   Therefore, the corruption occurs on the agent before the inventory
    is transmitted.
-   The issue is reproducible in one of our Windows environments.
-   We collected approximately six months of historical corrupted
    records.
-   Linux package inventory records in the monitored dataset did not
    exhibit the same corruption pattern.

## Historical samples

An anonymized CSV containing only:

-   Installation date
-   Package name
-   Vendor
-   Package version

is attached to this issue.

## Suspected source-code issue

During source-code analysis we identified the following implementation:

[...]

PR #38502, opened 2026-08-21
fix(syscollector): read Windows package names through the Unicode registry API

Ciren22007524:
## Description

Syscollector reports corrupted package names, versions, publishers and install
locations on Windows whenever those values contain characters that the system ANSI
code page cannot represent. The data is already damaged when the agent writes it to
`local.db`, so every consumer downstream (the manager, exports, dashboards) faithfully
renders data that was lost on the endpoint.

Registry strings are stored as UTF-16. `Utils::Registry::string()` reads them with
`RegQueryValueEx`, and because the agent is not built with `UNICODE` defined, that
resolves to `RegQueryValueExA`. The ANSI variant makes Windows transcode the value to
the system ANSI code page *before returning it to the caller*, so the loss happens
inside the OS, before any Wazuh code runs. The subsequent
`EncodingWindowsHelper::stringAnsiToStringUTF8()` call in the two-argument `string()`
overload cannot undo it — it re-encodes data that no longer holds the original
characters.

The conversion loses data in two distinct ways, and the second one is the reason this
goes unnoticed:

1. Characters with no mapping in the code page become a literal `?` (byte `0x3F`) —
   not a rendering artifact, an actual replacement in the stored bytes.
2. Characters that have a *visually similar* mapping are silently best-fitted to a
   different character (`WC_NO_BEST_FIT_CHARS` is not in play here). There is no `?`
   to grep for, so the value looks plausible while being wrong.

Measured on a machine with ACP 950, reading a `REG_SZ` that holds `深圳汉한é`:

| path | bytes returned | renders as |
|---|---|---|
| stored value (UTF-16) | — | `深圳汉한é` |
| `string()` — ANSI API | `E6 B7 B1 E5 9C B3 3F 3F 65` | `深圳??e` |
| `stringUtf8()` — Unicode API | `E6 B7 B1 E5 9C B3 E6 B1 89 ED 95 9C C3 A9` | `深圳汉한é` |

Both failure modes appear in that single line: `汉` and `한` were replaced by `?`,
and `é` was best-fitted to a plain `e` with no marker at all.

Deleting `local.db` and rescanning does not recover the data: the conversion is
deterministic, so the rescan reproduces the same corrupted values.

Related: #29885 reports the same root-cause family (ANSI-vs-Unicode Windows API
handling of non-ASCII data) in the FIM module. Different module and different code
path, so it is not fixed here, but the two share the underlying cause and are worth
looking at together.

## Proposed Changes

- **`src/shared_modules/utils/registryHelper.h`** — add
  `Registry::stringUtf8(valueName, value)`, which reads the value with
  `RegQueryValueExW` and converts the UTF-16 result to UTF-8 via the existing
  `EncodingWindowsHelper`. The existing `Registry::string()` is left untouched so no
  current caller changes behaviour; call sites opt in explicitly.
  - Rejects values that are not `REG_SZ`/`REG_EXPAND_SZ`, so a numeric value is
    reported as a failure instead of as empty text.
  - Allocates one `wchar_t` beyond the reported size, so a stored value that lacks
    its terminator cannot cause an over-read.

- **`src/data_provider/src/sysInfoWin.cpp`** — `getPackagesFromReg()` now uses
  `stringUtf8()` for `DisplayName`, `DisplayVersion`, `Publisher` and
  `InstallLocation`. `InstallDate` is left on `string()`: it is a numeric timestamp
  fed to `Utils::normalizeTimestamp()`, never text.

- **`src/data_provider/src/packages/appxWindowsWrapper.h`** — `stringUtf8()` for the
  four reported text values: `DisplayName`, the `ApplicationName` fallback,
  the MRT `MrtCache` localized-name lookup in `searchKeyOnSubRegistries()`, and
  `ApplicationCompany` in `searchPublisher()`.
  - The ProgID read in `searchPublisher()` deliberately stays on `string()`: it is not
    reported data, it is used to build the subkey path opened right below through
    `RegOpenKeyEx` (the ANSI variant), so it must stay ANSI-encoded. A comment records
    this.
  - `PackageRootFolder` stays on `string()`: `WindowsApps` paths are ASCII, and the
    value is later passed to `GetFileVersionInfoSizeA`.

Values read elsewhere through `Registry::string()` were reviewed and left alone
because they cannot carry non-ASCII text: the `KB######` strings in
`packagesWindowsParserHelper.h` (matched against a `KB[0-9]{6,}` regex) and
`ProcessorNameString` in `sysInfoWin.cpp` (ASCII by vendor convention). Values in
`osinfo/sysOsInfoWin.cpp` and the `RegQueryValueExA` call inside
`Registry::getValue()` (used by `sca_policy_check_win.cpp`) have the same latent
issue but are outside the scope of this fix.

### Results and Evidence

Standalone check compiled against the real `registryHelper.h` on Windows (ACP 950).
It writes `L"深圳汉한é"` as a UTF-16 `REG_SZ` under `HKCU`, reads it back through both
paths, and deletes the key:

```
system ANSI code page: 950
expected UTF-8               len=14 bytes=E6 B7 B1 E5 9C B3 E6 B1 89 ED 95 9C C3 A9
string()   [ANSI path]       len= 9 bytes=E6 B7 B1 E5 9C B3 3F 3F 65
stringUtf8() [Unicode]       len=14 bytes=E6 B7 B1 E5 9C B3 E6 B1 89 ED 95 9C C3 A9
PASS: stringUtf8() returned the stored characters as UTF-8
PASS: string() lost data, reproducing the reported corruption
PASS: missing value rejected
PASS: REG_DWORD value rejected
PASS: ProcessorNameString = "Intel(R) Core(TM) Ultra 7 155H"
ALL CHECKS PASSED
```

  - [ ] Linux — N/A, the changed code is inside `#ifdef WIN32`
  - [ ] Windows — not verified locally, no MinGW toolchain available on the
        contributor's machine; needs CI
  - [ ] MAC OS X — N/A, the changed code is inside `#ifdef WIN32`
- [x] Log syntax and correct language review — no log messages changed

Memory tests: N/A. The new method adds one `std::unique_ptr<wchar_t[]>` allocation
per read, sized from the value length `RegQueryValueExW` reports, with the handle
lifetime unchanged.

Decoder/Rule tests: N/A.
Engine: N/A.
Wazuh server API/Framework: N/A.

### Artifacts Affected

- Windows agent executable (`wazuh-agent`), syscollector package inventory.
- No configuration files, no schema changes. The `packages` records emitted for
  affected software change content only: the same fields, now holding the characters
  actually stored on the endpoint.

### Configuration Changes

N/A.

### Tests Introduced

- `src/shared_modules/utils/tests/registryHelper_test.cpp`
  - `RegistryStringUtf8KeepsCharactersOutsideTheSystemCodePage` — writes a `REG_SZ`
    holding U+6DF1 U+5733 (Han), U+6C49 (simplified-only Han), U+D55C (Hangul) and
    U+00E9 (Latin with diacritic); no single Windows ANSI code page can hold all four,
    so the test is code-page independent. Asserts `stringUtf8()` returns the exact
    UTF-8 bytes, and that `string()` returns something different unless the ACP is
    UTF-8 itself. Cleans up its own key.
  - `RegistryStringUtf8NoThrow` — a missing value returns false and leaves the output
    untouched.
  - `RegistryStringUtf8RejectsNonStringValues` — `~MHz` is a `REG_DWORD` and must be
    rejected rather than reported as empty text.

- `src/data_provider/tests/sysInfoWin/sysInfoWinAppx_test.cpp`
  - `FakeRegistry::string()` now reproduces the ANSI loss (every non-ASCII byte
    becomes `?`) and delegates to a new lossless `stringUtf8()`. This makes the
    regression permanent: if any of these call sites goes back to `string()`, the
    tests below fail.
  - `NonAsciiDisplayNameAndPublisherAreReadAsUtf8` — non-ASCII `DisplayName` and
    `ApplicationCompany` resolved through the FileAssociations ProgID.
  - `NonAsciiApplicationNameFallbackIsReadAsUtf8` — no `DisplayName`, so the name
    comes from the sub-application `ApplicationName` fallback.
  - `NonAsciiLocalizedNameFromCacheRegistryIsReadAsUtf8` — an `@{...}` MRT resource
    reference resolved to a non-ASCII localized name through `MrtCache`.

All test sources stay pure ASCII: expected data is written as `\xNN` escapes with the
code points named in a comment above each constant.

## Review Checklist

- [x] Code changes reviewed
- [x] Configuration changes documented — N/A, no configuration change
- [ ] Developer documentation reflects the changes — N/A, internal helper only

### DIFF
--- a/src/data_provider/src/packages/appxWindowsWrapper.h
+++ b/src/data_provider/src/packages/appxWindowsWrapper.h
@@ -296,7 +296,7 @@ class TAppxWindowsWrapper final : public IPackageWrapper
 
             try
             {
-                if (!registry.string("DisplayName", name))
+                if (!registry.stringUtf8("DisplayName", name))
                 {
                     name.clear();
                 }
@@ -364,7 +364,7 @@ class TAppxWindowsWrapper final : public IPackageWrapper
                         std::string value;
                         const TRegistry nameReg(m_key, m_userId  + "\\" + APPLICATION_STORE_REGISTRY + "\\" + m_appName + "\\" + folder + "\\Capabilities");
 
-                        if (nameReg.string("ApplicationName", value))
+                        if (nameReg.stringUtf8("ApplicationName", value))
                         {
                             name = value;
                             break;
@@ -476,7 +476,7 @@ class TAppxWindowsWrapper final : public IPackageWrapper
             std::string value;
             const TRegistry registry(m_key, path);
 
-            if (!registry.string(key, value))
+            if (!registry.stringUtf8(key, value))
             {
                 for (const auto& folder : TRegistry(m_key, path).enumerate())
                 {
@@ -509,10 +509,12 @@ class TAppxWindowsWrapper final : public IPackageWrapper
                 std::string data;
                 std::string vendorRegistry;
 
+                // Not reported data: this is a ProgID used as a subkey of the path opened below.
+                // The registry is opened through the ANSI API, so it must stay ANSI encoded.
                 registry.string(value, vendorRegistry);
                 const TRegistry pubRegistry(m_key, m_userId  + "\\" + APPLICATION_VENDOR_REGISTRY + "\\" + vendorRegistry + "\\Application");
 
-                if (pubRegistry.string("ApplicationCompany", data))
+                if (pubRegistry.stringUtf8("ApplicationCompany", data))
                 {
                     if (!Utils::startsWith(data, PREFIX_LOCALIZATION))
                     {
--- a/src/data_provider/src/sysInfoWin.cpp
+++ b/src/data_provider/src/sysInfoWin.cpp
@@ -439,17 +439,17 @@ static void getPackagesFromReg(const HKEY key, const std::string& subKey, std::f
                 std::string location;
                 std::string architecture;
 
-                if (packageReg.string("DisplayName", value))
+                if (packageReg.stringUtf8("DisplayName", value))
                 {
                     name = value;
                 }
 
-                if (packageReg.string("DisplayVersion", value))
+                if (packageReg.stringUtf8("DisplayVersion", value))
                 {
                     version = value;
                 }
 
-                if (packageReg.string("Publisher", value))
+                if (packageReg.stringUtf8("Publisher", value))
                 {
                     vendor = value;
                 }
@@ -470,7 +470,7 @@ static void getPackagesFromReg(const HKEY key, const std::string& subKey, std::f
                     install_time = Utils::timestampToISO8601(packageReg.keyModificationDate());
                 }
 
-                if (packageReg.string("InstallLocation", value))
+                if (packageReg.stringUtf8("InstallLocation", value))
                 {
                     location = value;
                 }
--- a/src/data_provider/tests/sysInfoWin/sysInfoWinAppx_test.cpp
+++ b/src/data_provider/tests/sysInfoWin/sysInfoWinAppx_test.cpp
@@ -88,7 +88,33 @@ class FakeRegistry
             return ret;
         }
 
+        /*
+         * Mimics Registry::string(): the ANSI registry API hands the caller data already
+         * transcoded to the system ANSI code page, so every character missing from that code
+         * page is lost before the caller sees it. Replacing each non-ASCII byte reproduces
+         * that loss, so a test using non-ASCII data fails if a call site goes back to the
+         * ANSI path.
+         */
         bool string(const std::string& valueName, std::string& value) const
+        {
+            if (!stringUtf8(valueName, value))
+            {
+                return false;
+            }
+
+            for (auto& character : value)
+            {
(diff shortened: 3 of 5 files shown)


suletom:
FYI, this PR may also be relevant to #38146.

In #38146 we've been investigating intermittent corruption of Windows Syscollector package names, where trailing memory-like data appears in the collected package name.

Our current hypothesis points to the registry string handling, specifically cases where the registry data may not be properly null-terminated before being used as a C-style string.

I noticed that this PR changes the same area and explicitly handles registry values without a terminator by allocating additional space for the terminating wchar_t.

Although the primary purpose of this PR is ANSI/Unicode handling, it looks like the new implementation might also address the suspected root cause of #38146.

We've reproduced #38146 in our environment and also tested a patch based on explicitly respecting the registry value length, which appeared to resolve the corruption.

Leaving this here in case it is useful during review/testing of this PR.

PR #38509, opened 2026-08-21
Raise the number of Active Response commands the agent can load

### LINKED ISSUE #38358: Wazuh execd issue

vikman90:
## Description

`wazuh-execd` kept the Active Response command table of `etc/shared/ar.conf` in three static arrays sized
for `MAX_AR` (64) entries. Since #38410 the extra entries are discarded in a controlled way instead of
being written past the end of the table, but they are still discarded: a command defined beyond the 64th
position never runs on the agent, even though `ar.conf` is distributed in full to every agent.

@nimra1923 confirmed that fix and asked for the limit itself to be reconsidered: their
deployment needs more than 90 command definitions, combining Windows Active Response scripts that are
still in production with a growing set of custom Linux ones.

The table is now loaded into a dynamically grown array, so it takes only the memory that the file actually
defines, and the limit is raised from 64 to 1024 commands. A limit is kept, rather than removed
altogether, because the agent parses `ar.conf` again on every command it cannot resolve: an arbitrarily
large file would otherwise grow the table and make each Active Response event re-parse the whole file.
1024 is about eleven times what the largest reported deployment needs, so no legitimate configuration
reaches it.

Related to #38358.

## Proposed Changes

- `src/os_execd/exec.c`: replace `exec_names`, `exec_cmd` and `exec_timeout` with a dynamically grown
  array of `{name, cmd, timeout}` entries, each one holding its own allocation. The table grows in blocks
  of 32 entries, is released on every reload, and can be freed through the new `FreeExecConfig()`. The
  limit is now `MAX_AR_COMMANDS`, and the entries beyond it keep being reported once per state change.
- `src/os_execd/execd.h`: declare `FreeExecConfig()`.
- `src/headers/defs.h`: add `MAX_AR_COMMANDS` (1024), the number of commands an agent loads. `MAX_AR`
  stays as the number that previous agent versions support, which is what the manager warning reports.
- `src/error_messages/error_messages.h`: reword `(1319)` for the new limit.
- `src/error_messages/warning_messages.h`: `wazuh-analysisd` still warns about the number of defined
  commands, because agents of previous versions keep the limit of 64. Message `(1307)` is reworded to
  state that those agents will not run the exceeding entries, instead of describing a limit of the
  current agent.

Commands are loaded, resolved and executed exactly as before. Duplicated names, missing command files,
directory traversal checks and malformed lines keep their previous behaviour.

### Results and Evidence

Agent v4.14.9 with an `etc/shared/ar.conf` holding 100 command definitions, each one pointing to its own
script:

```
# wc -l < /var/ossec/etc/shared/ar.conf
100
# head -2 /var/ossec/etc/shared/ar.conf
test-ar-001 - test-ar-001.sh - 0
test-ar-002 - test-ar-002.sh - 0
# tail -1 /var/ossec/etc/shared/ar.conf
test-ar-100 - test-ar-100.sh - 0
```

**Before the change** (branch base, which already includes #38410). Triggering the commands 1, 64, 65 and
100 loads the table, discards everything past the 64th entry and rejects the last two commands:

```
2026/08/21 11:46:27 wazuh-execd: INFO: Started (pid: 20393).
2026/08/21 11:46:29 wazuh-execd: ERROR: (1319): Reached the maximum of 64 active response commands. Ignoring the remaining entries of 'etc/shared/ar.conf'.
2026/08/21 11:46:30 wazuh-execd: ERROR: (1311): Invalid command name 'test-ar-065' provided.
2026/08/21 11:46:30 wazuh-execd: ERROR: (1311): Invalid command name 'test-ar-100' provided.
```

Only the commands within the limit run:

```
2026/08/21 11:46:29 test-ar-001.sh executed
2026/08/21 11:46:29 test-ar-064.sh executed
```

**After the change** — same agent, same 100-entry `ar.conf`. The daemon starts, loads the whole file and
reports nothing:

```
2026/08/21 12:20:32 wazuh-execd: INFO: Started (pid: 57745).
```

Triggering the 100 commands runs the 100 of them, each one its own script:

```
# 100-entry ar.conf: executed 100 of 100, distinct scripts 100
2026/08/21 12:20:34 test-ar-001.sh executed
2026/08/21 12:20:36 test-ar-064.sh executed
2026/08/21 12:20:36 test-ar-065.sh executed
2026/08/21 12:20:37 test-ar-100.sh executed
```

`ossec.log` holds no error or warning, and the daemon stays up.

**Past the new limit** — same agent with an `ar.conf` of 1100 definitions. The first 1024 are loaded, the
condition is reported once even though the table is reloaded on each unresolved command, and the daemon
survives:

```
2026/08/21 12:17:43 wazuh-execd: INFO: Started (pid: 56385).
2026/08/21 12:17:45 wazuh-execd: ERROR: (1319): Reached the limit of 1024 active response commands. Ignoring the remaining entries of 'etc/shared/ar.conf'.
2026/08/21 12:17:46 wazuh-execd: ERROR: (1311): Invalid command name 'cap-ar-1025' provided.
2026/08/21 12:17:46 wazuh-execd: ERROR: (1311): Invalid command name 'cap-ar-1100' provided.
```

The commands 1 and 1024 run, and 1025 and 1100 are reported as unknown, which is the same behaviour as
any other command name that does not exist.

Memory held by the same daemon after loading the table, so that the cost of the higher limit is visible:

```
100 entries:  RSS 4476 kB
1024 entries: RSS 4504 kB
```

Full unit test suite for the agent target, built with AddressSanitizer and LeakSanitizer:

```
100% tests passed, 0 tests failed out of 86
```

The reworded manager warning, rendered from its call site in `AR_ReadConfig()`:

```
(1307): Defined 91 active response commands, more than the 64 supported by previous agent versions. Those agents may fail to run active responses until the exceeding entries of 'etc/shared/ar.conf' are removed.
```

### Artifacts Affected

- `wazuh-execd` (all agent packages, and the manager package since the same daemon runs there).
- `wazuh-analysisd` (manager packages), for the wording of the warning.

### Configuration Changes

None. No setting is introduced and no existing one changes its meaning. The number of Active Response
command definitions an agent loads goes from 64 to 1024, which needs no configuration.

### Documentation Updates

None required.

### Tests Introduced

`src/unit_tests/os_execd/test_read_exec_config.c` is extended, and now covers `ReadExecConfig()` with:

- A configuration of 10 commands, loaded in full.
- A configuration of 500 commands, loaded in full and checked entry by entry.
- A 96-command configuration, where the entries at the positions 64, 65 and 96 — the last one supported
  by previous agent versions, the first one they discarded, and the last one of the file — are loaded
  like any other one.
- A reload with a shorter configuration, which leaves no entry of the previous one behind.
- A configuration of exactly `MAX_AR_COMMANDS` commands, which fits with nothing reported.
- A configuration past `MAX_AR_COMMANDS`, where the exceeding entries are reported and discarded.
- A line that consumes no entry after a full table, which must not be taken for an exceeding entry.
- Several reloads of a truncated configuration, reported once and re-armed when the file is corrected.
- A duplicated name whose first definition holds no command, back-filled with the command of the second
  one.
- A malformed line, reported and consuming no entry.

Every entry carries a distinct timeout, so the values are verified to survive the growth of the table,
and the teardown releases it through `FreeExecConfig()` so that the suite stays clean under
LeakSanitizer.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ All notable changes to this project will be documented in this file.
 
 ### Agent
 
+#### Changed
+
+- Raised from 64 to 1024 the number of active response commands that `wazuh-execd` can load from `etc/shared/ar.conf`. ([#38509](https://github.com/wazuh/wazuh/pull/38509))
+
 #### Fixed
 
 - Fixed the Windows agent MSI upgrade leaving the agent broken after the next reboot, and the silent `/q` upgrade hanging, when a system restart was pending. ([#38277](https://github.com/wazuh/wazuh/pull/38277))
--- a/src/error_messages/error_messages.h
+++ b/src/error_messages/error_messages.h
@@ -198,7 +198,7 @@
 #define EXEC_INV_CMD    "(1316): Invalid AR command: '%s'"
 #define EXEC_CMD_FAIL   "(1317): Could not launch command %s (%d)"
 #define EXEC_BAD_NAME   "(1318): Command name truncated %s"
-#define EXEC_MAX_AR     "(1319): Reached the maximum of %d active response commands. Ignoring the remaining entries of '%s'."
+#define EXEC_MAX_AR     "(1319): Reached the limit of %d active response commands. Ignoring the remaining entries of '%s'."
 
 #define AR_NOAGENT_ERROR                "(1320): Agent '%s' not found."
 #define EXEC_QUEUE_CONNECTION_ERROR     "(1321): Error communicating with queue '%s'."
--- a/src/error_messages/warning_messages.h
+++ b/src/error_messages/warning_messages.h
@@ -13,7 +13,7 @@
 
 /* Active Response */
 #define AR_SERVER_AGENT "(1306): Invalid agent ID. Use location=server to run AR on the manager."
-#define AR_MAX_COMMANDS "(1307): Defined %d active response commands, more than the maximum of %d. The exceeding entries of '%s' may not be loaded by the agents."
+#define AR_MAX_COMMANDS "(1307): Defined %d active response commands, more than the %d supported by previous agent versions. Those agents may fail to run active responses until the exceeding entries of '%s' are removed."
 
 /* File integrity monitoring warning messages*/
 #define FIM_WARN_ACCESS                         "(6900): Accessing  '%s': [(%d) - (%s)]"
--- a/src/headers/defs.h
+++ b/src/headers/defs.h
@@ -161,8 +161,10 @@ https://www.gnu.org/licenses/gpl.html\n"
 /* Active Response files */
 #define DEFAULTAR_FILE  "ar.conf"
 #define AR_BINDIR       "active-response/bin"
-/* Maximum number of active response commands that an agent can load */
+/* Number of active response commands that previous agent versions can load */
 #define MAX_AR          64
+/* Maximum number of active response commands that an agent loads */
+#define MAX_AR_COMMANDS 1024
 #ifndef WIN32
 #define DEFAULTAR       "etc/shared/" DEFAULTAR_FILE
 #define AGENTCONFIG     "etc/shared/agent.conf"
--- a/src/os_execd/exec.c
+++ b/src/os_execd/exec.c
@@ -12,13 +12,56 @@
 #include "os_regex/os_regex.h"
 #include "execd.h"
 
-static char exec_names[MAX_AR + 1][OS_FLSIZE + 1];
-static char exec_cmd[MAX_AR + 1][OS_FLSIZE + 1];
-static int  exec_timeout[MAX_AR + 1];
-static int  exec_size = 0;
-static int  f_time_reading = 1;
-static int  f_max_reported = 0;
+/* Number of command slots the table grows by */
+#define EXEC_TABLE_CHUNK 32
 
+typedef struct exec_entry {
+    char *name;
+    char *cmd;
+    int timeout;
+} exec_entry_t;
+
+static exec_entry_t *exec_table = NULL;
+static int exec_size = 0;
+static int exec_capacity = 0;
+static int f_time_reading = 1;
+static int f_max_reported = 0;
+
+/* Discard the loaded commands, keeping the allocated slots for the next read */
+static void ClearExecTable()
+{
+    int i;
+
+    for (i = 0; i < exec_size; i++) {
+        os_free(exec_table[i].name);
+        os_free(exec_table[i].cmd);
+        exec_table[i].timeout = 0;
+    }
+
+    exec_size = 0;
+}
+
+/* Append a command to the table, growing it when there is no slot left */
+static void AddExecEntry(const char *name, const char *cmd, int timeout)
+{
+    if (exec_size == exec_capacity) {
+        exec_capacity += EXEC_TABLE_CHUNK;
+        os_realloc(exec_table, exec_capacity * sizeof(exec_entry_t), exec_table);
+    }
+
+    os_strdup(name, exec_table[exec_size].name);
+    os_strdup(cmd, exec_table[exec_size].cmd);
+    exec_table[exec_size].timeout = timeout;
+    exec_size++;
+}
+
+/* Release the command table */
+void FreeExecConfig()
+{
+    ClearExecTable();
+    os_free(exec_table);
+    exec_capacity = 0;
+}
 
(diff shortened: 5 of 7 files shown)


cborla (approved):
LGTM
- [x] Code changes reviewed

PR #38513, opened 2026-08-21
Add Azure Artifact Signing dlib support to the MSI signer

Darioortegaleyva:
## Summary
- `generate_wazuh_msi.ps1` gains `-DLIB_PATH`/`-DLIB_METADATA` to sign via `signtool /dlib /dmdf` (Azure Artifact Signing) instead of `/a` (DigiCert). Default behaviour is unchanged when these are not passed.
- Every `signtool`/`candle`/`light` call now checks its exit code and throws on failure, plus a final `signtool verify /pa` pass. Today a partial signing failure is silently swallowed.
- Same exit-code fix applied to the two legacy `.bat` scripts (no Azure support added there, nothing in CI calls them for signing).

## Context
Part of [wazuh-agent-packages#922](https://github.com/wazuh/wazuh-agent-packages/issues/922). Companion PR in `wazuh-agent-packages` wires this into the build workflows.

## Not done yet
- 4.14.x maintenance branch is not ported yet - only `main` in this PR.
- The Azure signing path is unverified end-to-end: no Azure account/certificate profile exists yet to test the actual `/dlib` call against.

## Test plan
- [ ] `-SIGN no` path (default) still builds an MSI identically to before.
- [ ] `-SIGN yes` without DLIB_PATH still signs via DigiCert `/a` as today.
- [ ] Once an Azure Artifact Signing account exists: `-DLIB_PATH`/`-DLIB_METADATA` produces a validly signed MSI/exe/dll, verified with `signtool verify /v /pa`.
- [ ] Port to the 4.14.x branch.

### DIFF
--- a/packages/windows/generate_wazuh_msi.ps1
+++ b/packages/windows/generate_wazuh_msi.ps1
@@ -88,13 +88,15 @@ function BuildWazuhMsi(){
             "..\build\lib\*.dll"
         )
 
-        # Expand each pattern ourselves so a zero-match glob is a deliberate skip,
-        # not an ambiguous signtool exit code.
         foreach ($pattern in $filesToSign) {
             $matchedFiles = @(Get-ChildItem -Path $pattern -ErrorAction SilentlyContinue)
             if ($matchedFiles.Count -eq 0) {
-                Write-Host "No files matched $pattern, skipping."
-                continue
+                if ($pattern -match '[*?]') {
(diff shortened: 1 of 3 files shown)


Darioortegaleyva:
Superseded by a different approach: signing via azure/login + azure/artifact-signing-action directly on a hosted runner, no allocator/manual dlib needed. See wazuh-agent-packages#977/#981/#982. Closing this prototype.

PR #38537, opened 2026-08-21
Fix stateful DELETE events being dropped for files first seen via realtime/whodata

### LINKED ISSUE #38522: Deleted files stay in `wazuh-states-fim-files` though the delete alert fires
|---|---|---|---|---|
| 5.0.0-latest | wazuh-agent (syscheckd / Agent Sync Protocol), wazuh-indexer state indices | Manager/Agent | Docker (public ECR wazuh-latest) | Amazon Linux 2023 (agents), single-node stack |

## Description

The File Integrity Monitoring inventory lists files that no longer exist on the endpoint. The delete alert is generated correctly, so the two halves of FIM's dual event system (stateless events for alerts, stateful events for the state index) disagree with each other.

## Expected behavior

When a monitored file is deleted, its document in `wazuh-states-fim-files` is removed, or at minimum updated with a marker the UI can filter on.

## Actual behavior

The delete alert fires; the state document is left untouched and continues to be shown in the inventory as a live file.

## Evidence

Stateless half, correct. Delete alerting needs no configuration to enable:

```
FIM findings by event.action:   2 created   1 modified   2 deleted

/testfim/passwd.bak    event.action=deleted  event.type=[deletion]  level=medium  09:29:52
/testfim/secret.conf   event.action=deleted  event.type=[deletion]  level=medium  09:44:53
```

Stateful half, wrong. With both files deleted from disk:

```
agent disk:              /testfim/  -> 0 files
wazuh-states-fim-files:  2 hits -> /testfim/secret.conf (33 B), /testfim/passwd.bak (12 B)
```

The index otherwise contains exactly the monitored paths, so the two entries are stale rather than out-of-scope noise:

```
configured: /etc, /usr/bin, /usr/sbin, /bin, /sbin, /boot, /testfim
indexed:    563 /etc   234 /usr   1 /bin   1 /sbin   2 /testfim   = 801 docs
```

2 of the 801 files shown in the dashboard do not exist.

The state document carries no deletion marker: no `event.action`, no `deleted` flag, no tombstone. It reads as a live file, so there is nothing the UI could filter on:

```json
{"file": {"path": "/testfim/secret.conf", "size": 33, "owner": "root",
          "permissions": ["rw-r--r--"], "hash": {"sha256": "e0a0ab66..."}},
 "state": {"document_version": 2, "modified_at": "2026-08-21T09:29:42.111Z"}}
```

`document_version` shows the document was never revisited. Version 1 was the creation, version 2 the modification. The delete happened at 09:44:53 and produced neither a version 3 nor a removal; `state.modified_at` is still the 09:29:42 modification.

## Ruled out

- **Sync latency.** Still present after 25 minutes and 4 `FIM synchronization finished successfully` cycles.
- **Needs a full scan.** A full agent restart with a fresh baseline scan does not clear them either.
- **Create/delete race.** `passwd.bak` lived ~6 seconds, so it could legitimately fall under the documented behavior that only items with `sync=1` are added to the persistent sync queue. That does not cover `secret.conf`, which was synced (it is in the index with its post-modification size and hash) and deleted 15 minutes later.

## Steps to reproduce

```bash
# agent: monitor a directory in realtime
<directories realtime="yes" check_all="yes">/testfim</directories>
# restart agent, then
docker exec <agent> sh -c 'echo x > /testfim/f1; sleep 5'
# wait for it to appear in wazuh-states-fim-files (~5 min sync cycle), then
docker exec <agent> rm -f /testfim/f1
# wait several more sync cycles and query
curl -sk -u admin:admin 'https://localhost:9200/wazuh-states-fim-files/_search' \
  -d '{"query":{"prefix":{"file.path":"/testfim"}},"_source":["file.path","state"]}'
```

The delete alert appears in `wazuh-findings*` with `event.action: deleted`; the state document stays.

### LINKED ISSUE #38608: FIM: fim_file_scan() sets sync flag after deleting rows that fail schema validation, causing WARNING spam and synced_docs_files drift
|---|---|---|---|---|
| 5.0.0 | FIM (syscheckd, agent) | Agent | Sources | Any |

## Summary

In `fim_file_scan()` (`src/syscheckd/src/file/file.c`), when a file fails schema validation during a scheduled full scan, `cleanup_failed_fim_files()` deletes its row from the local DB **before** `process_pending_sync_updates()` tries to set its sync flag. This causes:

1. A spurious `WARNING: Failed to update sync flag for path: ..., received callback type: ...` log line for every affected file (the row no longer exists by the time `fim_db_set_sync_flag()` runs — traced into `src/shared_modules/dbsync/src/sqlite/sqlite_dbengine.cpp`'s `syncRow()`: with `update_only=true` and a missing row, the update silently no-ops instead of inserting, so `fim_db_set_sync_flag()` falls into its failure branch and logs the warning).
2. A **permanent drift** in `synced_docs_files` (and the analogous registry counters, `synced_docs_registry_keys`/`values`): `transaction_callback()` already incremented the counter and queued the item for sync-flag promotion before the schema validation failure is known. `fim_db_file_delete()` (called by `cleanup_failed_fim_files()`) removes the row directly, bypassing the `DELETED` branch of `transaction_callback()` that would otherwise decrement the counter back. Nothing else ever corrects it. Over time this can make the file/registry-document limit enforcement (`<file_limit>`/registry equivalents) wrongly refuse to promote/sync a legitimate document, because the counter reads higher than the real number of synced documents.

## Context

This was found and confirmed while fixing #38522 (agent-side, PR #38537). That PR fixes the exact same ordering bug on the realtime/whodata path (`fim_file()`'s non-transactional branch) — see `src/syscheckd/src/file/file.c`, the `fim_file()` block around the `failed_paths`/`pending_sync_updates` handling. It deliberately does **not** fix the scheduled-scan instance of the same bug, because the fix used for the single-file realtime path doesn't generalize directly:

- In `fim_file()`, at most one file is processed per call, so `failed_paths` and `pending_sync_updates` (populated from the same single `transaction_callback()` invocation) can only ever be about that one path — trivial to correlate.
- In `fim_file_scan()`, a full scan can have many files in both `failed_paths` and `pending_sync_updates` simultaneously, and the current data structures don't correlate entries between the two lists by path. A correct fix needs to match failed paths to their corresponding pending sync items (e.g. by path) before deciding whether to call `process_pending_sync_updates()` for the rest, and to compensate the counter only for the files that actually failed.

## Where to look

- `src/syscheckd/src/file/file.c`, `fim_file_scan()`: the `cleanup_failed_fim_files(failed_paths); ... process_pending_sync_updates(FIMDB_FILE_TABLE_NAME, pending_sync_updates);` sequence, with an in-code comment already pointing at this issue.
- Same pattern likely applies to `src/syscheckd/src/registry/registry.c`'s `fim_registry_scan()` for the registry key/value tables — not independently verified, worth checking during the fix.
- `fim_db_set_sync_flag()` (`src/syscheckd/src/db/src/file.cpp`) and dbsync's `syncRow()` (`src/shared_modules/dbsync/src/sqlite/sqlite_dbengine.cpp`) for how the `update_only=true` no-op-on-missing-row behavior leads to the warning.

## How to reproduce (not yet attempted, expected shape)

1. Configure a scheduled scan over a directory containing a file that will fail schema validation (e.g. a path/attribute combination the FIM schema rejects).
2. Run a full scan with `syscheck.debug=2` and `<file_limit>` enabled with a modest limit.
3. Confirm the `WARNING: Failed to update sync flag for path: ...` line appears in `ossec.log` for that file.
4. Inspect `synced_docs_files` behavior over repeated scans (or via debug logging) to confirm it stays inflated relative to the actual count of synced documents.

MAnDumu:
## Description

Files first detected by a realtime/whodata event never had their local `sync` column promoted to `1` in the agent's FIM database, because the non-transactional branch of `fim_file()` (the one realtime/whodata uses for a single file) built its `callback_ctx` without a `pending_sync_updates` list — it stayed `NULL`, so `transaction_callback()`'s `INSERTED`/`MODIFIED` branches silently skipped queuing the sync flag update. The stateful CREATE event still got persisted (its own locally-computed flag defaults to 1), so the file did show up in `wazuh-states-fim-files`.

When that file is later deleted, the `DELETED` branch reads the sync flag from the local DB column (still `0`) and `validate_and_persist_fim_event()` drops the stateful DELETE because it requires `sync_flag == 1`. The stateless alert had already fired before that gate, so the delete alert appears in `wazuh-findings*` but the state document is never removed from `wazuh-states-fim-files` — it stays there forever, since the local row is unconditionally deleted right after, leaving nothing to reconcile on a later scan or restart.

Closes #38522

## Proposed Changes

- `src/syscheckd/src/file/file.c`: in `fim_file()`'s non-transactional branch, create a `pending_sync_updates` list, wire it into the `callback_ctx`, and call `process_pending_sync_updates(FIMDB_FILE_TABLE_NAME, pending_sync_updates)` after `fim_db_file_update()` — mirroring the pattern the scheduled-scan path (`fim_file_scan()`) already uses. Added a short comment explaining why this path doesn't need `syscheck.fim_scan_mutex` (see Results and Evidence).
- Fixes both the `INSERTED` case (the one reported in the issue) and, as a side effect of sharing the same `callback_ctx` field, the `MODIFIED` promotion branch (`sync_flag == 0 && file_limit > 0`), which had the same silent-skip bug for files first seen via realtime/whodata.
- `src/unit_tests/syscheckd/test_fim_scan.c`: updated the expectations of 8 existing tests that now exercise the new `process_pending_sync_updates()` call (both the Linux and `TEST_WINAGENT` variants), and added two new regression tests (`test_transaction_callback_add_queues_sync_flag_update`, `test_transaction_callback_modify_promotes_pending_sync_update`) asserting the sync item actually gets queued with `sync_value=1` for the `INSERTED` and `MODIFIED`-promote cases.

### Results and Evidence

Root cause confirmed by code review (already documented in the issue) and reproduced live on a running agent on this session's VM before writing the fix: with `<directories realtime="yes">/testfim</directories>` and `syscheck.debug=2`, creating and deleting a file under `/testfim` showed `send_syscheck_msg` (the stateless alert) for the delete, but **no** `persist_syscheck_msg`/"Persisting FIM event" line — unlike the create, which had both. The local row for that path was gone from `fim.db` afterwards, confirming the state document is left orphaned with nothing left to reconcile.

Independent review of this diff:
- `code-config-reviewer` flagged (1) the `TEST_WINAGENT` variant of two tests (`test_fim_checker_fim_regular`, `test_fim_checker_fim_regular_warning`) was missing the same expectation update — fixed; (2) the non-transactional path of `fim_file()` never takes `syscheck.fim_scan_mutex`, unlike `fim_file_scan()`, raising a possible use-after-free risk with `fim_shutdown_waiter()`'s teardown of the FIMDB context.
- A follow-up adversarial review confirmed (1) is resolved, and re-checked (2) against `FIMDB::updateItem()`/`FIMDB::teardown()` (`fimDB.cpp`): both the pre-existing `fim_db_file_update()` call and the new `fim_db_set_sync_flag()` call go through `FIMDB::instance().updateItem()`, which takes a `shared_lock` on `m_handlersMutex` and checks `m_stopping` before touching the dbsync handler; `teardown()` takes the exclusive `lock_guard` on the same mutex before nulling it. No new UAF risk — the fix extends an already-safe, symmetric pattern. Documented with a code comment instead of blocking on it.

## Manual tests with their corresponding evidence

  - [x] Linux (`TARGET=agent`, built clean, no new warnings from this change)
  - [ ] Windows (`TARGET=winagent` could not be compiled in the dev VM used for this PR — an unrelated pre-existing DLL-linkage error in `schema_validator`, unrelated to FIM/syscheckd, aborted the build before it ever reached `syscheckd`. The Windows-only test variant was reviewed by reading the code, not by a real compile — needs CI or a healthy winagent toolchain to confirm.)
  - [ ] MAC OS X (not attempted)

- Memory tests for Linux
  - [ ] Coverity
  - [x] AddressSanitizer (the cmocka unit test binaries in this repo are already built with ASan/LeakSanitizer; ran clean with no leaks across all 11 `syscheckd` test binaries)
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - N/A (this fix is in `syscheckd`, not the engine)

- Wazuh server API/Framework
  - N/A (agent-side fix only)

### Artifacts Affected

- `wazuh-agent` binary (Linux and, once confirmed, Windows) — `syscheckd` module only. No manager-side artifacts affected.

### Configuration Changes

N/A — no new configuration parameters, no changes to defaults or existing behavior visible to the user beyond the bug fix itself.

### Tests Introduced

- `test_transaction_callback_add_queues_sync_flag_update` (`test_fim_scan.c`): calls `transaction_callback(INSERTED, ...)` with a real (non-`NULL`) `pending_sync_updates` list and asserts exactly one item is queued with `sync_value == 1` for the inserted path — locks in the wiring this PR adds.
- `test_transaction_callback_modify_promotes_pending_sync_update` (`test_fim_scan.c`): same idea for the `MODIFIED` promotion branch (`old.sync == 0`, `file_limit > 0`), which shares the same `pending_sync_updates` field and is fixed as a side effect of this change.
- Updated expectations on 8 pre-existing tests (`test_fim_file_add`, `test_fim_file_modify`, `test_fim_file_error_on_insert`, `test_fim_checker_fim_regular` x2 variants, `test_fim_checker_fim_regular_warning` x2 variants, `test_fim_checker_root_file_within_recursion_level`) that now exercise the new `process_pending_sync_updates()` call.
- Verified these are real regression guards, not false positives, by temporarily reverting the `file.c` fix and confirming the 6 modified/new tests fail (`git stash` on the source file only, rebuild, rerun); restored the fix afterward and confirmed all tests pass again.
- Scope of coverage: unit-level only (cmocka, `TARGET=agent`, Linux). No integration/E2E test was added for this PR — recommended follow-up: an E2E test creating a file under a realtime directory, waiting for sync, deleting it, and asserting the document leaves `wazuh-states-fim-files`, as already suggested in the issue's root-cause analysis.

## Review Checklist

- [ ] Code changes reviewed
- [x] Configuration changes documented (N/A, none)
- [ ] Windows build/test confirmation pending in CI (see Manual tests section above)

### DIFF
--- a/src/syscheckd/src/file/file.c
+++ b/src/syscheckd/src/file/file.c
@@ -1455,7 +1455,7 @@ void fim_file_scan() {
         // leaving synced_docs_files off by one. A full scan's failed_paths and
         // pending_sync_updates aren't correlated by path here, so fixing it needs matching
         // entries across both lists rather than the single-file short-circuit used above.
-        // Tracked separately for a follow-up fix.
+        // Tracked in https://github.com/wazuh/wazuh/issues/38608 for a follow-up fix.
         cleanup_failed_fim_files(failed_paths);
         OSList_Destroy(failed_paths);
 
--- a/src/syscheckd/src/file/file.c
+++ b/src/syscheckd/src/file/file.c
@@ -226,6 +226,7 @@ STATIC void handle_orphaned_delete(const char* path,
     cJSON* sync_json = cJSON_GetObjectItem(result_json, "sync");
     if (sync_json == NULL) {
         mdebug1("Couldn't find sync for orphaned delete '%s'", path);
+        cJSON_Delete(stateful_event);
         return;
     }
     int sync_value = (int)cJSON_GetNumberValue(sync_json);
@@ -306,18 +307,27 @@ STATIC void transaction_callback(ReturnTypeCallback resultType,
         case INSERTED:
             txn_context->event->type = FIM_ADD;
             sync_operation = OPERATION_CREATE;
-            // For CREATE events: determine if within limit and defer sync flag update
+            // For CREATE events: determine if within limit and defer sync flag update.
+            // Check-then-increment must be atomic: the scheduled scan and a realtime/whodata
+            // event can run this concurrently (see synced_docs_mutex's declaration comment).
             if (syscheck.file_limit > 0) {
+                w_mutex_lock(&synced_docs_mutex);
+                int docs_before = synced_docs_files;
                 sync_flag = (synced_docs_files < syscheck.file_limit) ? 1 : 0;
+                if (sync_flag == 1 && txn_context->pending_sync_updates != NULL) {
+                    synced_docs_files++;
+                }
+                w_mutex_unlock(&synced_docs_mutex);
+                // Logged pre-increment (docs_before), matching the value the sync_flag
+                // decision above was actually based on.
                 mdebug2("INSERTED: synced_docs_files=%d, file_limit=%d, sync_flag=%d, path=%s",
-                        synced_docs_files, syscheck.file_limit, sync_flag, path);
+                        docs_before, syscheck.file_limit, sync_flag, path);
             } else {
                 sync_flag = 1;
                 mdebug2("INSERTED: file_limit=0 (unlimited), sync_flag=1, path=%s", path);
             }
             // Add to deferred list if sync_flag should be 1
             if (sync_flag == 1 && txn_context->pending_sync_updates != NULL) {
-                synced_docs_files++;
                 cJSON* sync_item = cJSON_CreateObject();
                 if (sync_item != NULL) {
                     cJSON_AddStringToObject(sync_item, "path", path);
@@ -344,8 +354,17 @@ STATIC void transaction_callback(ReturnTypeCallback resultType,
             }
 
             if (sync_flag == 0 && syscheck.file_limit > 0) { // Promote
-                if (synced_docs_files < syscheck.file_limit) {
+                // Check-then-increment must be atomic (see synced_docs_mutex's declaration
+                // comment); the cJSON construction below doesn't touch shared state, so it
+                // stays outside the lock.
+                w_mutex_lock(&synced_docs_mutex);
+                bool promote = (synced_docs_files < syscheck.file_limit);
+                if (promote) {
                     synced_docs_files++;
+                }
+                w_mutex_unlock(&synced_docs_mutex);
+
+                if (promote) {
                     cJSON* sync_item = cJSON_CreateObject();
                     if (sync_item != NULL) {
                         cJSON_AddStringToObject(sync_item, "path", path);
@@ -372,7 +391,9 @@ STATIC void transaction_callback(ReturnTypeCallback resultType,
                 if (sync_json != NULL && cJSON_IsNumber(sync_json)) {
                     sync_flag = sync_json->valueint;
                     if (sync_flag == 1) {
+                        w_mutex_lock(&synced_docs_mutex);
                         synced_docs_files--;
+                        w_mutex_unlock(&synced_docs_mutex);
                     }
                 }
             }
@@ -1144,21 +1165,41 @@ void fim_file(const char *path,
 
         fim_db_file_update(&new_entry, callback_data);
 
-        // Delete files that failed schema validation (outside transaction)
-        cleanup_failed_fim_files(failed_paths);
-        OSList_Destroy(failed_paths);
-
-        // Most realtime/whodata events don't have anything to promote (the file was already
-        // synced, or this was a MODIFIED with no old.sync=0 to promote), so skip the call
-        // entirely when there's nothing to do. When there IS something to promote, still don't
-        // log a summary here (process_pending_sync_updates() no longer logs on its own) — a mass
-        // create/modify against a realtime-monitored directory calls this once per file, and a
-        // per-file "Processed N pending sync flag updates" line would flood debug=1. The
-        // per-scan summary (scheduled scan, registry scan) is logged by those callers instead.
-        if (OSList_GetFirstNode(pending_sync_updates) != NULL) {
+        // If schema validation failed, the row is about to be deleted below: don't try to
+        // set its sync flag afterwards (fim_db_set_sync_flag() would find no row to update
+        // and log a spurious WARNING on every such event), and undo the increment
+        // transaction_callback() already applied to synced_docs_files for it, since
+        // cleanup_failed_fim_files() removes the row without going through the DELETED
+        // branch that would otherwise balance it. This path handles a single file per call,
+        // so failed_paths and pending_sync_updates — both populated from the same
+        // transaction_callback() invocation — can only ever be about that one path; a
+        // non-empty failed_paths here means whatever is in pending_sync_updates is for it.
+        if (OSList_GetFirstNode(failed_paths) != NULL) {
+            OSListNode *node_it;
+            OSList_foreach(node_it, pending_sync_updates) {
+                pending_sync_item_t *item = (pending_sync_item_t *)node_it->data;
+                if (item != NULL && item->sync_value == 1) {
+                    w_mutex_lock(&synced_docs_mutex);
+                    synced_docs_files--;
+                    w_mutex_unlock(&synced_docs_mutex);
+                }
+            }
+        } else if (OSList_GetFirstNode(pending_sync_updates) != NULL) {
+            // Most realtime/whodata events don't have anything to promote (the file was
+            // already synced, or this was a MODIFIED with no old.sync=0 to promote), so skip
+            // the call entirely when there's nothing to do. When there IS something to
+            // promote, still don't log a summary here (process_pending_sync_updates() no
+            // longer logs on its own) — a mass create/modify against a realtime-monitored
+            // directory calls this once per file, and a per-file "Processed N pending sync
+            // flag updates" line would flood debug=1. The per-scan summary (scheduled scan,
+            // registry scan) is logged by those callers instead.
             process_pending_sync_updates(FIMDB_FILE_TABLE_NAME, pending_sync_updates);
         }
         OSList_Destroy(pending_sync_updates);
+
+        // Delete files that failed schema validation (outside transaction)
+        cleanup_failed_fim_files(failed_paths);
+        OSList_Destroy(failed_paths);
 
         free_file_data(new_entry.file_entry.data);
     }
@@ -1405,6 +1446,16 @@ void fim_file_scan() {
         fim_db_transaction_deleted_rows(db_transaction_handle, transaction_callback, &txn_ctx);
 
         // Delete files that failed schema validation (outside transaction)
+        //
+        // Pre-existing gap (not introduced by #38522's fix, out of scope here): unlike the
+        // realtime/whodata call site above, this doesn't skip/compensate
+        // process_pending_sync_updates() for paths that also ended up in failed_paths — a
+        // file that fails schema validation during a scan can get its row deleted here and
+        // then still have a stale sync-flag update attempted below, spamming a WARNING and
+        // leaving synced_docs_files off by one. A full scan's failed_paths and
+        // pending_sync_updates aren't correlated by path here, so fixing it needs matching
+        // entries across both lists rather than the single-file short-circuit used above.
+        // Tracked separately for a follow-up fix.
         cleanup_failed_fim_files(failed_paths);
         OSList_Destroy(failed_paths);
 
--- a/src/syscheckd/src/file/file.c
+++ b/src/syscheckd/src/file/file.c
@@ -226,7 +226,6 @@ STATIC void handle_orphaned_delete(const char* path,
     cJSON* sync_json = cJSON_GetObjectItem(result_json, "sync");
     if (sync_json == NULL) {
         mdebug1("Couldn't find sync for orphaned delete '%s'", path);
-        cJSON_Delete(stateful_event);
         return;
     }
     int sync_value = (int)cJSON_GetNumberValue(sync_json);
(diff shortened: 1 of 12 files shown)


MAnDumu:
## Verification on a real agent (wazuh#38522)

Repro: `<directories realtime="yes" check_all="yes">/testfim</directories>` + `syscheck.debug=2`, create a file and delete it, watching `ossec.log`.

### Before the fix

The create persists fine (locally-computed `sync_flag=1`), but the delete only emits the stateless alert — **the persistence line never shows up**:

```
# create
DEBUG: INSERTED: synced_docs_files=8428, file_limit=30000, sync_flag=1, path=/testfim/repro_delete.txt
DEBUG: (6321): Sending FIM event: {...,"event":{"type":"added"}...}
DEBUG: (6339): Persisting FIM event: {...}

# delete (several minutes later)
DEBUG: (6321): Sending FIM event: {...,"event":{"type":"deleted"}...}
# <- no "Persisting FIM event" line for the delete: it's silently dropped
```

### After the fix

The create now also queues and confirms the sync flag, and the delete **does persist**:

```
# create
DEBUG: INSERTED: synced_docs_files=8428, file_limit=30000, sync_flag=1, path=/testfim/verify_38522.txt
DEBUG: Added item to pending sync list: /testfim/verify_38522.txt (version: 1, sync: 1)
DEBUG: (6321): Sending FIM event: {...,"event":{"type":"added"}...}
DEBUG: (6339): Persisting FIM event: {...}
DEBUG: Setting sync=1 for path: /testfim/verify_38522.txt
DEBUG: Setting sync flag to 1 for path: /testfim/verify_38522.txt (version: 1)
DEBUG: Successfully updated sync flag to 1 for path: /testfim/verify_38522.txt

# delete
DEBUG: (6321): Sending FIM event: {...,"event":{"type":"deleted"}...}
DEBUG: (6339): Persisting FIM event: {...}
# <- the stateful delete now persists
```

### Conclusion

With the fix, a file created via realtime gets `sync=1` from the moment it's created (it used to stay at `sync=0` forever), so the stateful delete no longer gets dropped by `validate_and_persist_fim_event`'s gate.

src/syscheckd/src/file/file.c
```diff
@@ -1129,6 +1147,10 @@ void fim_file(const char *path,
         // Delete files that failed schema validation (outside transaction)
         cleanup_failed_fim_files(failed_paths);
         OSList_Destroy(failed_paths);
+
+        process_pending_sync_updates(FIMDB_FILE_TABLE_NAME, pending_sync_updates);
```
jpcerrone: Calling this here might produce `DEBUG: Processed 0 pending sync flag updates` on every realtime event. This log only showed up once per whole scan before these changes. You might be able to verify this by copying, say, 100 files into a realtime-monitored directory and seeing if the DEBUG message floods the agent's logs. It may be worth it to investigate if we can modify this function so that this doesn't happen.

src/unit_tests/syscheckd/test_fim_scan.c
```diff
@@ -3785,6 +3868,78 @@ static void test_transaction_callback_modify(void **state) {
     data->txn_context->entry = NULL;
 }
 
+// Regression test for #38522: the MODIFIED branch's promotion path ("old" row has sync=0 and
```
jpcerrone, line 4173: Claude comment on unit tests. Ignore if not relevant:

  1. No unit test actually covers the change in file.c

  __wrap_fim_db_file_update (src/unit_tests/wrappers/wazuh/syscheckd/fim_db_wrappers.c:76) is a bare mock_type(int) — it never invokes callback.callback_txn. So:
 
  - The 8 updated tests all assert "Processed 0 pending sync flag updates", i.e. they only prove process_pending_sync_updates() is called, with an empty list.
  - The two new tests call transaction_callback() directly with a list they build themselves — they test the branch that was never broken.
 
  Delete the .pending_sync_updates = pending_sync_updates line from the callback_ctx initializer at src/syscheckd/src/file/file.c:1133 and keep everything else, and every unit test still passes — while the bug is
  fully back. The PR body's claim that reverting the fix makes 6 tests fail is true only because reverting the whole hunk also removes the process_pending_sync_updates() call.

  Fix: have the wrapper optionally drive the callback (will_return a result JSON + result type), so one test can run fim_file(..., NULL, NULL) end-to-end and assert "Processed 1 pending sync flag updates". The
  integration test does cover this, but the unit layer shouldn't be blind to the actual defect.

jpcerrone (changes requested):

src/syscheckd/src/file/file.c
MAnDumu: Added a guard so `process_pending_sync_updates()` is only called when the list actually has something to process (`OSList_GetFirstNode(pending_sync_updates) != NULL`), instead of unconditionally on every realtime event.

Tested as suggested: created 100 files at once in a realtime-monitored directory on a real running agent with `syscheck.debug=2`. Each genuine CREATE still logs "Processed 1 pending sync flag updates" (there's real work to do — promoting the sync flag), but a subsequent MODIFY on an already-synced file no longer logs anything, since the list stays empty and the call is skipped entirely. Before this guard it would have logged "Processed 0..." on that MODIFY too.

Pushed in e153b3c.

src/unit_tests/syscheckd/test_fim_scan.c
MAnDumu: Reverted just the `.pending_sync_updates = pending_sync_updates` line (leaving everything else, including the `process_pending_sync_updates()` call, intact), rebuilt, and every existing test still passed while the bug was fully back, exactly as described.

Fixed with the approach suggested: extended `__wrap_fim_db_file_update()` with an opt-in `expect_fim_db_file_update_invoking_callback()` that makes the mock actually drive `callback.callback_txn(...)` with a `will_return`-provided result type + JSON, instead of always being a no-op. Added `test_fim_file_add_queues_sync_flag_update_end_to_end`, which calls `fim_file()` directly (not `transaction_callback()`) and asserts "Processed 1 pending sync flag updates" instead of "...0...".

Re-ran the same revert-just-that-line test against the new test: it now fails ("Processed 0..." vs "Processed 1..." expected) while the rest of the suite stays green, confirming it actually catches the regression this time. Also caught and fixed a bug in my own fix while at it — the new test hardcoded the Linux path in the mocked JSON/log expectations without a TEST_WINAGENT branch, which would have broken under Windows.

Pushed in e153b3c.

jpcerrone (dismissed):
LGTM

cborla (approved):
LGTM
- [x] Code changes reviewed

PR #38545, opened 2026-08-24
Removing dead arm for ebpf and fixing memory leak

### LINKED ISSUE #32501: FIM eBPF Provider: investigate alternatives to LSM on arm64 systems
|Wazuh version|Component|Install type|Platform|
|---|---|---|---|
| 4.13.0 | FIM / whodata / eBPF | Agent | `arm64` systems |

## Description

Found errors in FIM in whodata mode when using the eBPF provider (since version 4.13). To fix the problem with special `btrfs` file systems, we had to implement the use of some LSM hooks in kernel versions higher than 6.8. However, this is not possible on arm64 systems, as it prevents the bpf program from loading correctly.

## Tasks

- Further investigate the use of LSM in ARM system kernels:
- Find alternatives so that the eBPF provider in btrfs file systems works correctly on ARM systems.
- Verify that all expected events are generated correctly on all ARM systems.
- Add or update tests to cover this scenario.
- Update the documentation if necessary.

## Definition of Done

- [ ] The eBPF provider detects and reports files added/modified/deleted in `arm64` systems (including those with `btrfs` file system).
- [ ] Tests confirm that behaviour.
- [ ] Documentation is updated to reflect any changes or improvements.

Miguevrgo:
|Related issue|
|---|
|#32501|

## Description

FIM eBPF whodata on arm64: remove a dead source file and add a kprobe fallback so the module keeps working on kernels where the BPF-LSM path cannot.

**Root cause (measured on real arm64, qemu; and x86_64, qemu+KVM, with the E2E-built `4.14.9` packages):**

1. **arm64 builds from `modern.bpf.c`, not `modern-arm.bpf.c`.** `modern-arm.bpf.c` (added 2025-10-08) is a hand-maintained parallel copy that nothing compiles. Building the arm64 package in the real container with the file deleted still produces a `modern.bpf.o` with all ten programs. If anyone ever wired it "because it is the arm one", arm64 would silently lose all LSM hooks and rename detection. **This PR deletes it.**

2. **On arm64 < 6.8 with BPF-LSM active, the LSM path cannot start.** Two independent kernel gates: `bpf_d_path` on path hooks needs kernel **6.8** (so the dpath variants fail to load), and attaching a BPF-LSM program on arm64 needs `CONFIG_DYNAMIC_FTRACE_WITH_DIRECT_CALLS` = kernel **6.4** (so the walk variants fail to attach with `-ENOTSUPP`/errno 524). `init_bpfobj` then returned failure and FIM dropped to audit, even though kprobes work fine on those kernels.

## Proposed Changes

- Delete the dead `src/syscheckd/src/ebpf/src/modern-arm.bpf.c` (493 lines).
- **`ebpf_whodata.cpp`: fall back to kprobe when the LSM path fails.** Extract the open/select/load/attach sequence into `load_and_attach(path, use_lsm)` (keeping the existing dpath→walk load retry), and in `init_bpfobj` retry once with the kprobe program set when the LSM path fails to load or attach, before giving up. New info log `(6054)`.
- eBPF test build hygiene: C++17 + sanitizers in the test targets; free the added fields in the test mock.

## Configuration options

N/A

## Logs/Alerts example

**The fix, arm64 kernel 5.15, BPF-LSM active (real `wazuh-agent_4.14.9-0_arm64` package):**
```
(6047): Initializing eBPF driver for FIM whodata.
(6051): BPF LSM is active in the running kernel; using LSM hooks for create/modify/delete events.
(6053): BPF LSM load failed with the bpf_d_path-based variants; retrying with the manual path walker ...
(6983): Attaching BPF program 'file_open_walk' failed (errno=524: Unknown error 524).
(6973): Attaching BPF program failed.
(6054): BPF LSM hooks failed to load or attach on this kernel; falling back to kprobe hooks.
(6049): eBPF healthcheck action succeeded: create file.
(6049): eBPF healthcheck action succeeded: modify content.
(6049): eBPF healthcheck action succeeded: modify metadata.
(6049): eBPF healthcheck action succeeded: delete file.
(6048): Healthcheck for eBPF FIM whodata module success.
```
Before the fix this ended at `(6959): ... Switching all whodata eBPF configuration to audit.`

**Healthcheck matrix (arm64, out-of-the-box vs BPF-LSM forced in the kernel `lsm=` list):**

| Kernel | Distro | default (out-of-the-box) | bpf-lsm-forced |
|--------|--------|--------------------------|----------------|
| 5.15   | Ubuntu 22.04 | kprobe → PASS | before: FAIL→audit — **with fix: PASS via kprobe** |
| 6.8    | Ubuntu 24.04 | kprobe → PASS | LSM → PASS |
| 6.17   | Ubuntu 25.10 | kprobe → PASS | LSM → PASS |

**x86_64 regression (real `wazuh-agent_4.14.9-0_amd64` package, kernel 6.8):**

| Scenario | Hook | Healthcheck |
|----------|------|-------------|
| default (bpf not in LSM list) | kprobe | PASS |
| bpf forced in LSM list | LSM (dpath) | PASS (fallback never fires) |

The new fallback does not change x86 behaviour: the LSM path succeeds there, so the kprobe retry is never reached.

**btrfs path resolution under the LSM path (arm64 kernel 6.8), eBPF hook vs userspace `stat`:**
```
path='/mnt/btr/CMP_root'          stat_dev=49 stat_ino=264   ebpf_dev=47 ebpf_ino=264
path='/mnt/btr/sub/deep/CMP_sub'  stat_dev=50 stat_ino=268   ebpf_dev=47 ebpf_ino=268
path='/mnt/subonly/SUBONLYF'      (subvol mounted on its own path, resolved correctly)
```
Paths correct (fixes the kprobe `s_fs_info` bug); inode correct; device wrong (eBPF reports the shared btrfs superblock `s_dev`, not the per-subvolume `anon_dev` — separate, still-open identity bug).

## Tests

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [x] Package installation 
- [ ] Source upgrade
- [ ] Package upgrade


Darioortegaleyva (changes requested):
Just one thing I see that may be blocking: load_and_attach() never destroys the bpf_links it creates (no bpf_link__destroy is even loaded in this codebase), so a mid-loop attach failure can leave LSM hooks leaked in the kernel while the new kprobe fallback installs a second, overlapping hook set on top.

Darioortegaleyva (dismissed):
LGTM, changes have been applied correctly

MarcelKemp (changes requested):
The tests aren't compiling properly – could you have a look at it?

src/syscheckd/src/ebpf/src/ebpf_whodata.cpp
```diff
@@ -599,17 +602,59 @@ int init_bpfobj()
...
+        return 0;
+    }
+
+    if (g_bpf_lsm_active)
+    {
+        logFn(LOG_INFO, FIM_EBPF_LSM_KPROBE_FALLBACK);
+        if (load_and_attach(bpfobj_path, false, true) == 0)
+        {
+            g_bpf_lsm_active = false;
+            return 0;
```
MarcelKemp: **[S2 · blocking · introduced]** Unchanged since the last review: the fallback loads the kprobe program set, which is the pre-LSM state and still resolves paths incorrectly on btrfs — the reason issue #32501 exists. On a stock arm64 agent `is_bpf_lsm_active()` returns false (Ubuntu does not put `bpf` in the LSM list), so per the PR's own matrix the kprobe set is used on every kernel and whodata events keep carrying wrong paths in exactly the reported scenario. Either drop the automatic closing reference to #32501 and open a follow-up for btrfs+arm64, or fix path resolution in the kprobe variants so the fallback preserves the btrfs behaviour.

src/syscheckd/src/ebpf/tests/fimEbpfWhodataTest/init_bpf_obj_test.cpp
```diff
@@ -20,12 +20,33 @@ class InitBpfobjTest : public ::testing::Test {
         MockFimebpf::mock_loggingFunction = mock_loggingFunction;
         MockFimebpf::SetMockFunctions();
         bpf_helpers = std::make_unique<w_bpf_helpers_t>();
+        bpf_helpers->bpf_link_destroy = (bpf_link__destroy_t)mock_bpf_link_destroy;
```
MarcelKemp: **[S3 · non-blocking · introduced]** Unchanged: `resetGlobalState()` still doesn't clear the (also unchanged) `global_links`, and its `global_obj` is still a test-local symbol distinct from the library's internal one — the suite stays order-dependent under `--gtest_shuffle`.

MarcelKemp (approved):

Issue #35986, opened 2026-05-08
FIM Whodata broken on Fedora 44


Miguevrgo:
## Version

| Wazuh version | Component | Install type | Install method | Platform |
|---|---|---|---|---|
| 4.14.5-1 | Wazuh agent | Agent | Packages | Fedora Linux 44 — `6.19.10-300.fc44.x86_64` |

Additional context: `audit-4.1.4-1.fc44`, `audit-rules-4.1.4-1.fc44`, SELinux Enforcing, `audispd-plugins` not installed by default.

## Description

FIM whodata fails to start on a default Fedora 44 installation. The agent silently falls back to real-time (inotify) mode and no who-data fields (`user`, `process`, `ppid`) appear in FIM alerts.

Two independent issues block whodata on Fedora 44:

1. `audispd-plugins` is not installed by default — the `audisp-af_unix` binary that wazuh requires does not exist on a fresh install.
2. The `audit-rules` package ships `-a task,never` as its default ruleset, suppressing all syscall auditing and causing the wazuh health check to time out even when `audispd-plugins` is manually installed.

## Steps to reproduce

```bash
# Fresh Fedora 44, wazuh-agent 4.14.x installed and enrolled

# Add to ossec.conf inside <syscheck>:
# <directories whodata="yes">/tmp/fim_whodata_test</directories>

mkdir -p /tmp/fim_whodata_test
/var/ossec/bin/wazuh-control restart

# Wait ~15 s, then check:
grep -E "6636|6642|6913|never,task" /var/ossec/logs/ossec.log
auditctl -l
```

### Scenario A — default install (`audispd-plugins` absent)

`audisp-af_unix` does not exist. The socket `/var/ossec/queue/sockets/audit` is never created and wazuh-syscheckd fails immediately:

```log
2026/05/08 09:02:29 wazuh-syscheckd: INFO: (6024): Generating Auditd socket configuration file: 'tmp/af_wazuh.conf'
2026/05/08 09:02:29 wazuh-syscheckd: INFO: (6025): Audit plugin configuration (/etc/audit/plugins.d/af_wazuh.conf) was modified. Restarting Auditd service.
2026/05/08 09:02:30 wazuh-syscheckd: ERROR: (6636): Cannot connect to socket 'queue/sockets/audit'.
2026/05/08 09:02:30 wazuh-syscheckd: ERROR: Can't init auditd socket in 'init_auditd_socket()'
[...]

PR #30215, opened 2025-06-09
Add new check audit version for FIM whodata

jotacarma90:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/28984#top|

# Description
Hi team,

in this PR we are adding a check of the Auditd version installed in the system, to configure the plugin correctly according to the version.
We were encountering some problems due to the latest changes made in version 4 of audit in relation to the plugin, because among others, they have tightened the security of the plugin preventing the configuration file to be a symlink. Therefore we are now creating the file directly in the audit directory.

Finally, after investigacions, our conditionals for fix the audit version issue, are this:
- Older than 3.1.1 no changes needed: `AUDISP_CONFIGURATION_0`
- Between 3.1.1 and 4.0.2 (with an exception in 3.1.5 (Due to newest releases for some OSs, included some fixes)), include the changes for audispd af_unix plugin to a standalone program, so a fix to the warnings is included: `AUDISP_CONFIGURATION_1`
- Audit version 3.1.5 and starting at 4.0.3, include the changes to propagate event format to the audisp-af_unix plugin: `AUDISP_CONFIGURATION_2`

Code related:
```
        if (get_audit_version_code(&vcode) != 0) {
            mdebug2("Could not get audit version code. Using default configuration.");
            strcpy(audisp_config, AUDISP_CONFIGURATION_2);
        } else {
            mdebug2("Audit version detected: %u.%u.%u", (vcode >> 16) & 0xFF, (vcode >> 8) & 0xFF, vcode & 0xFF);
            if (vcode < VERCODE(3, 1, 1)) {
                // Before audit version 3.1.1 old code worked with builtin_af_unix
                strcpy(audisp_config, AUDISP_CONFIGURATION_0);
            } else if (vcode < VERCODE(3, 1, 5)) {
                // Audit version 3.1.1 include the changes for audispd af_unix plugin to a standalone program
                strcpy(audisp_config, AUDISP_CONFIGURATION_1);
            } else if (vcode < VERCODE(4, 0, 0)) {
                // Audit version 3.1.5 include the changes to propagate event format to the audisp-af_unix plugin
                strcpy(audisp_config, AUDISP_CONFIGURATION_2);
            } else if (vcode < VERCODE(4, 0, 3)) {
                // From audit version 4.0.0 to 4.0.2 format changes are not included
                strcpy(audisp_config, AUDISP_CONFIGURATION_1);
            } else {
                // From audit version 4.0.3 format changes are included
[...]

Issue #28984, opened 2025-04-04
Audit Whodata provider fails to load in some OS


mjcr99:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
|4.12.0| FIM Whodata | Agent | Packages and Sources | Fedora 42  AlmaLinux 10 |

## Description

Hi team, 

During the testing of the following issues, unexpected behavior regarding `auditd` has been observer:
* https://github.com/wazuh/wazuh/issues/28921
* https://github.com/wazuh/wazuh/issues/27951

Covering the requested dependencies to use audit has explained in the [audit whodata documentation](https://documentation.wazuh.com/current/user-manual/capabilities/file-integrity/advanced-settings.html) the following logs appear for each platform:

AlmaLinux 10:

```
2025/04/04 14:32:13 wazuh-syscheckd: INFO: (6023): No socket found at 'queue/sockets/audit'. Restarting Auditd service.
2025/04/04 14:32:13 wazuh-syscheckd: ERROR: Could not launch command to restart Auditd.
2025/04/04 14:32:13 wazuh-syscheckd: WARNING: (6913): Who-data engine could not start. Switching who-data to real-time.
```

Fedora 42:

```
2025/04/04 14:19:31 wazuh-syscheckd: INFO: (6023): No socket found at 'queue/sockets/audit'. Restarting Auditd service.
2025/04/04 14:19:32 wazuh-syscheckd: ERROR: (6636): Cannot connect to socket 'queue/sockets/audit'.
2025/04/04 14:19:32 wazuh-syscheckd: ERROR: Can't init auditd socket in 'init_auditd_socket()'
2025/04/04 14:19:32 wazuh-syscheckd: WARNING: (6913): Who-data engine could not start. Switching who-data to real-time.
```

Both distros are using `auditctl version 4.0.3` which seems to be the main root of this issue.

jotacarma90:
# Update 08/04/2025
- Research on the problem. It seems a clear bug caused by the upgrade of audit to 4.0.3, just in this version, investigating in the changes of [audispd.c](https://github.com/linux-audit/audit-userspace/blob/51d154c5b7ec91831cbb89fe6ca54d8eb7ba344c/audisp/audispd.c#L101C1-L119C35) a filter has been introduced to avoid configuration files that are not regular:
```
static void load_plugin_conf(conf_llist *plugin)
{
	DIR *d;

	/* init plugin list */
	plist_create(plugin);

[...]

PR #38555, opened 2026-08-24
Probe the audisp plugin configuration instead of trusting the audit version

### LINKED ISSUE #38536: FIM who-data mode fails to activate on Amazon Linux 2023
## Description

When a group's centralized `agent.conf` configures a directory for who-data monitoring (`<directories whodata="yes" ...>` + `<whodata><provider>audit</provider></whodata>`) on an Amazon Linux 2023 agent, `wazuh-syscheckd` never succeeds in starting the who-data engine. It permanently falls back to real-time mode instead, so file events for that directory are missing the `user.*` identity fields (`user.name`, `user.id`, `user.group.name`) and `file.mode` is reported as `realtime` instead of `whodata`.

The same group configuration, applied at the same time to a Debian-family Linux agent (Ubuntu 24, deb) and a Windows agent (Windows Server 2019), activates who-data correctly on both. The failure is isolated to the Amazon Linux 2023 / rpm agent.

On the affected host, `auditd`'s `audisp-af_unix` plugin logs `Missing arguments, using defaults` every time it starts with the config file `wazuh-syscheckd` generates (`/etc/audit/plugins.d/af_wazuh.conf`), and ends up listening on its own default socket path instead of the path Wazuh configured (`/var/ossec/queue/sockets/audit`). Because of this, `wazuh-syscheckd`'s `init_auditd_socket()` can never connect, logs error `(6636)`, and permanently disables who-data for the life of the process (no retry). This was verified to be unrelated to the default `-a task,never` audit rule that Amazon Linux 2023 ships (removing that rule did not change the outcome) and unrelated to restart timing (multiple restarts with 8-10+ second waits between them all failed identically).

This is fully and deterministically reproducible; it is not an edge case.

## Environment

| Component | Value |
|---|---|
| Wazuh version | 5.0.0 (dev build, package revision `latest`) |
| Affected agent package | `wazuh-agent-5.0.0-latest.x86_64.rpm` |
| Affected agent OS | Amazon Linux 2023, amd64 (`amazon-2023-amd64`) |
| Affected agent audit package | `audit-3.1.5-1.amzn2023.0.2.x86_64` |
| Manager / Indexer / Dashboard | Wazuh 5.0.0 (dev build), RedHat 9, arm64 |
| Other agents in the same environment (activate who-data correctly) | Ubuntu 24 (deb), Windows Server 2019 |

## Steps to reproduce

1. Provision a Wazuh 5.0.0 environment with a manager+indexer+dashboard stack and at least one Amazon Linux 2023 agent (`amazon-2023-amd64`, rpm package).
2. On the AL2023 agent, install and enable `auditd` (`sudo dnf install -y audit && sudo systemctl enable --now auditd`), and create a monitored directory with pre-existing files, e.g.:
   ```bash
   sudo mkdir -p /home/fim/whodata
   sudo bash -c 'echo "old text" > /home/fim/whodata/wd_old_file_02.txt'
   ```
3. In the Wazuh Dashboard, create an agent group (e.g. `e2e-fim-test`), add the AL2023 agent to it, and set its `agent.conf` to:
   ```xml
   <agent_config os="Linux">
     <syscheck>
       <frequency>30</frequency>
       <directories whodata="yes" report_changes="yes">/home/fim/whodata</directories>
       <whodata>
         <provider>audit</provider>
       </whodata>
     </syscheck>
   </agent_config>
   ```
4. Wait for the agent to report the new configuration (confirmed via the "Reported N minutes ago" badge / `SYNCHRONIZED` status).
5. On the agent host, modify the pre-existing file:
   ```bash
   sudo bash -c 'echo "modified content" > /home/fim/whodata/wd_old_file_02.txt'
   ```
6. In the Wazuh Dashboard, go to **Explore > Discover**, select index pattern `wazuh-findings-v5-system-activity*`, and query `wazuh.agent.id:"<agent id>" and file.path:"/home/fim/whodata/wd_old_file_02.txt" and event.action:"modified"`. Expand the resulting document.

### Additional verification performed directly on the host (for this report)

```bash
sudo auditctl -l                                    # rule set currently loaded
sudo journalctl -u auditd --no-pager -n 30           # audisp-af_unix startup messages
sudo tail -40 /var/ossec/logs/ossec.log              # wazuh-syscheckd whodata init messages
sudo cat /etc/audit/plugins.d/af_wazuh.conf           # plugin config wazuh-syscheckd generated
man auditd-plugins                                    # documents the args=/format= config keys
```

## Actual result

- The expanded document's `file.mode` is `"realtime"`, not `"whodata"`.
- The document has no `user.name`, `user.id`, or `user.group.name` fields at all (verified via both the dashboard's field sidebar and the raw JSON view).
- `sudo auditctl -l` never shows the `wazuh_fim` watch rule for the monitored directory: it is never registered.
- `/var/ossec/logs/ossec.log` shows, on every restart that regenerates the audit plugin config (5/5 observed in this session, at 15:01:20, 15:49:37, 15:50:47, 15:53:11 and 15:55:16 UTC):
  ```
  wazuh-syscheckd: INFO: (6024): Generating Auditd socket configuration file: 'tmp/af_wazuh.conf'
  wazuh-syscheckd: INFO: (6025): Audit plugin configuration (/etc/audit/plugins.d/af_wazuh.conf) was modified. Restarting Auditd service.
  wazuh-syscheckd: ERROR: (6636): Cannot connect to socket 'queue/sockets/audit'.
  wazuh-syscheckd: ERROR: Can't init auditd socket in 'init_auditd_socket()'
  wazuh-syscheckd: WARNING: (6913): Who-data engine could not start. Switching who-data to real-time.
  ```
- `sudo journalctl -u auditd` shows, on every one of those restarts:
  ```
  auditd[PID]: audit dispatcher initialized with q_depth=2000 and 1 active plugins
  audisp-af_unix[PID]: Missing arguments, using defaults
  audisp-af_unix[PID]: audisp-af_unix plugin is listening for events
  ```
- There is no self-healing: 6+ FIM synchronization cycles (every 5 minutes) elapsed after the initial failure with no further attempt to start who-data; it stays on real-time for the life of the process.
- This was confirmed end-to-end through the dashboard: two other agents in the same environment (Debian-family Linux and Windows), given the identical group configuration and the identical file operation, both show `file.mode:whodata` and correct `user.*` fields; only the Amazon Linux 2023 agent shows `file.mode:realtime` with no `user.*` fields.

## Expected result

Who-data should activate on the Amazon Linux 2023 agent exactly as it does on the Debian-family and Windows agents given the same configuration: `file.mode` should be `"whodata"` for events under the configured directory, and the document should carry `user.name`, `user.id`, and `user.group.name` (Linux) matching the OS user who performed the file operation.

## Attachments

All evidence files are collected in the attached folder (`issue_attachments/`):

[issue_attachments.zip](https://github.com/user-attachments/files/31317480/issue_attachments.zip)

- `05_rpm_agent_modified_missing_user_fields.png`: the affected Amazon Linux 2023 agent's modified-file document, showing `file.mode:realtime` and no `user.*` fields.
- `04_deb_agent_modified_expanded.png`: the Debian-family agent's equivalent document, for contrast (`file.mode:whodata`, `user.*` present).
- `06_windows_agent_modified_expanded.png`: the Windows agent's equivalent document, for contrast.
- `02_findings_all_agents_events.png`: the Findings tab showing all three agents' events.
- `agent_e2e-linux-rpm-agent_ossec.log`: full `ossec.log` from the affected agent, including the `(6024)/(6025)/(6636)/(6913)` sequence.
- `agent_e2e-linux-deb-agent_ossec.log` / `agent_e2e-windows-agent_ossec.log`: contrast logs from the two agents where who-data works correctly.
- `auditd_and_audisp-af_unix_live_evidence.txt`: full live-collected supporting evidence (`auditctl -l` output, `audit.rules` and `af_wazuh.conf` contents, the `journalctl -u auditd` output across every restart, the relevant `man` page excerpts, and the additional checks performed directly on the host, such as removing the `-a task,never` rule and retesting).

nbertoldo:
## Description

Closes #38536.

FIM who-data never starts on an Amazon Linux 2023 agent with the `audit` provider. `wazuh-syscheckd` logs `(6636): Cannot connect to socket 'queue/sockets/audit'` and permanently falls back to real-time, so file events under the monitored directory carry `file.mode: realtime` and no `user.*` fields.

The root cause is the version-based selection of the audisp plugin configuration in `set_auditd_config()`. Audit 3.1.5 is bucketed into `AUDISP_CONFIGURATION_2`, which renders a **three token** `args` line plus `format = binary`:

```
args = 0640 /var/ossec/queue/sockets/audit string
format = binary
```

Amazon Linux 2023 ships `audit-3.1.5-1.amzn2023.0.2`, a rebuild whose `audisp-af_unix` still takes only two arguments (mode, path). It logs `Missing arguments, using defaults`, binds its own default socket instead of the configured one, and `init_auditd_socket()` has nothing to connect to. The `>= 3.1.5` bucket was introduced in #30215 / #28984 on the assumption that every 3.1.5 build carries the event format propagation change; a plain version comparison cannot tell a vendor rebuild that lacks it apart from the upstream release it is based on.

The fix stops treating the version as proof. The configuration expected for the installed version is still applied first, but the agent now verifies that it actually produced a usable socket before accepting it, and probes the remaining known configuration when it did not.

## Proposed Changes

- **`audisp_get_candidates()`** replaces the single-configuration choice: it returns the known configurations **ordered by preference**, keeping today's version table as the first choice and appending the others as fallbacks. Preference is unchanged for every version, so no working system changes the configuration it ends up with.
- **`set_auditd_config_template(plugin_dir, template)`** holds what `set_auditd_config()` used to do — render, compare by SHA1, write, restart Auditd — but for a caller-supplied configuration. `set_auditd_config()` is gone; the version lookup and the write step are now separate concerns.
- **`wait_for_audit_socket()`** waits up to 5 s (polling every 100 ms) for the audisp plugin to create the socket. `configure_audisp()` unlinks the socket before rewriting the configuration, so its reappearance is a reliable signal: a misconfigured plugin binds its own default path and the socket never shows up.
- **`configure_and_connect_audit_socket()`** drives the sequence from `audit_init()`: apply a candidate, wait for the socket, connect. On success it stops; otherwise it moves to the next candidate and logs the new `(6054)` ("...did not create the socket... Who-data started with an alternative plugin configuration") when a fallback is what got who-data running. When no known plugins directory exists it connects directly, without waiting.
- **A socket that is up and accepts connections decides.** Before configuring anything, if the socket is there and connects, the agent keeps it untouched when `audisp_configuration_is_known()` says the file behind it is one this agent could have written — Auditd is already serving a working setup and restarting it would be gratuitous. If the file matches no known template — tampered with, or written by an older agent — the connection is dropped and the normal flow rewrites it. This is what keeps Auditd from being restarted on every agent start, and it relies on a live socket rather than on the contents of a file, which after a probe that failed on every candidate would hold the last and least preferred one.
- **A socket that refuses connections is retried before being treated as stale, and is only removed when it can be recreated.** A plugin that is gone and a transient failure (no file descriptors left, a saturated backlog) look identical from here, so a single refusal is not taken as proof: it takes `AUDIT_SOCKET_CONNECT_RETRIES` attempts. Even then it is only unlinked when `restart_audit` is enabled — with it disabled the agent cannot restart Auditd to bring the socket back, and removing it would turn one failed start into who-data being broken on every start from then on. The same retry is used after a candidate is applied, so a transient refusal there cannot discard a configuration that works.
- **The candidate list is always walked in version order**, so `(6054)` past the first candidate keeps meaning exactly what it says: the configuration expected for the installed audit version did not work.
- **A candidate that fails to be written, or that Auditd refuses to restart with, no longer aborts the probe** — the next one is tried, which is the whole point of probing.
- **At most two candidates**, and fewer where a fallback makes no sense. Each extra candidate costs an Auditd restart when probed, and every restart reloads the host's audit rules and bounces the other audisp consumers. `format = binary` is not offered below audit 3.1.5, where it does not exist, and below 3.1.1 the standalone configuration is not offered either, since `/sbin/audisp-af_unix` does not exist there — that bucket has a single candidate.
- **`audit_socket_connect(int quiet)`** carries the connect logic so the probe can attempt it without logging `(6636): Cannot connect to socket` on each candidate; the error is reported once, at the end, with the same message text as before. `init_auditd_socket()` remains as the public entry point used by the reconnection path in `audit_read_events()`.
- New message `(6054)` in `information_messages.h`, and `MAX_AUDISP_CANDIDATES` moved to `syscheck_audit.h` since it is now part of a public function's contract.

### Results and Evidence

The message IDs quoted throughout this section:

| ID | Message |
|---|---|
| `(6019)` | File integrity monitoring real-time Whodata engine started. |
| `(6024)` | Generating Auditd socket configuration file: '%s' |
| `(6025)` | Audit plugin configuration (%s) was modified. Restarting Auditd service. |
| `(6054)` | *(new)* The audisp plugin configuration expected for the installed audit version did not create the socket '%s'. Who-data started with an alternative plugin configuration. |
| `(6636)` | Cannot connect to socket '%s'. |
| `(6642)` | Audit health check couldn't be completed correctly. |
| `(6910)` | Audit plugin configuration was modified. You need to restart Auditd. Who-data will be disabled. |
| `(6913)` | Who-data engine could not start. Switching who-data to real-time. |

A configuration cycle is one `(6024)` + `(6025)` pair: the agent writing `af_wazuh.conf` and restarting Auditd. One cycle means the configuration expected for the installed audit version worked; two means the probe had to fall back.

Environment: agent `wazuh-agent-5.0.0-latest.x86_64` on Amazon Linux 2023 (x86_64), manager 5.0.0-latest. Who-data with `<provider>audit</provider>` over `/home/fim/whodata`. `syscheck.debug=2` to make the version bucket visible.

**Before the fix** — `audit-3.1.5-1.amzn2023.0.2`:

```
DEBUG: Audit version detected: 3.1.5
ERROR: (6636): Cannot connect to socket 'queue/sockets/audit'.
ERROR: Can't init auditd socket in 'init_auditd_socket()'
WARNING: (6913): Who-data engine could not start. Switching who-data to real-time.
```
`journalctl -u auditd`: `audisp-af_unix[…]: Missing arguments, using defaults`. No socket, no `wazuh_fim` rule.

**After the fix** — same host, same audit version:

```
DEBUG: Audit version detected: 3.1.5
INFO: (6024): Generating Auditd socket configuration file: 'tmp/af_wazuh.conf'
INFO: (6025): Audit plugin configuration (…/af_wazuh.conf) was modified. Restarting Auditd service.
DEBUG: Could not establish the who-data socket 'queue/sockets/audit' with the current audisp
       plugin configuration. Trying an alternative one.
INFO: (6024): Generating Auditd socket configuration file: 'tmp/af_wazuh.conf'
INFO: (6025): Audit plugin configuration (…/af_wazuh.conf) was modified. Restarting Auditd service.
INFO: (6054): The audisp plugin configuration expected for the installed audit version did not
      create the socket 'queue/sockets/audit'. Who-data started with an alternative plugin configuration.
INFO: (6019): File integrity monitoring real-time Whodata engine started.
```
`journalctl -u auditd` on the second restart: `audisp-af_unix[…]: Client connected`. Socket created, `auditctl -l` shows `-w /home/fim/whodata -p wa -k wazuh_fim`, and the FIM event carries the who-data identity:

```json
{"file": {"path": "/home/fim/whodata/wd_old_file_02.txt", "mode": "whodata",
  "audit": {"user_name": "root", "user_id": "0", "group_name": "root",
            "audit_name": "vagrant", "process_name": "/usr/bin/bash", "process_id": 8934}}}
```

| Scenario | Result |
|---|---|
| AL2023, audit 3.1.5 (the bug) | `(6054)` + `(6019)`, ends on the two argument configuration |
| AL2023, audit 3.0.6 (non-regression, builtin bucket) | `(6019)` directly, **zero** `(6054)`, one Auditd restart |
| Second agent start, Auditd already serving the socket | only `(6019)`; **zero** configuration cycles and Auditd **not** restarted (`Init complete` count unchanged) — `af_wazuh.conf` is not touched at all |
| Auditd restarted on its own (reboot equivalent) | socket returns with the probed configuration; agent connects directly, no `(6054)` |
| Tampered `af_wazuh.conf` (matches no known template) | rewritten on the next agent start and who-data comes up, so the self-repair the unconditional rewrite used to provide is preserved |
| **AlmaLinux 9.3, upstream audit 3.1.5-8.el9** | `(6019)` directly with the three argument configuration: **one** configuration cycle, **zero** `(6054)` — the probe does not kick in where the expected configuration works |
| **Ubuntu 26.04, audit 4.1.2** | `(6019)` directly with the three argument configuration: one cycle, **zero** `(6054)` — covers the `>= 4.0.3` bucket on audit 4.x |
| **Ubuntu 22.04, audit 3.0.7** | `(6019)` directly with `builtin_af_unix`: one cycle, **zero** `(6054)` — the builtin bucket on the Debian family, where `/sbin/audisp-af_unix` does not even exist |
| **Amazon Linux 2, audit 2.8.1** | `(6019)` directly with `builtin_af_unix` written to `/etc/audisp/plugins.d`: one cycle, **zero** `(6054)` — the audit 2.x path, where the version is never consulted because the legacy plugins directory decides |

Unit tests: **64/64 passing** in `test_syscheck_audit`, in 74 ms.

The AlmaLinux 9 run is the one that shows the probe is the exception and not the new normal: on `audit-3.1.5-8.el9` the same three token `args` line that breaks on Amazon Linux works fine (`audisp-af_unix … Client connected`, no `Missing arguments`), so the agent stops at the first candidate. Same audit version string, opposite outcome — which is exactly why the version alone cannot be the decision.

Every configuration template and both plugin directories have now been exercised on a real system, and four of the five version buckets have been validated on a VM. The two that have not — `3.1.1-3.1.4` and `4.0.0-4.0.2` — both select the two argument template, which is exactly the one Amazon Linux 2023 ends up running successfully, so what is left untested there is the version comparison itself, covered by the unit tests.

**Cost of the probe when who-data cannot work at all.** Measured on Ubuntu 26.04 by removing `/sbin/audisp-af_unix`, which is the situation a stock Fedora is in (`audispd-plugins` is not installed by default, see #35986): the agent takes **14 s** to give up instead of the current **4 s**, and restarts Auditd **twice** instead of once. The outcome is the same as today — a single `(6636)` and a clean fallback to real-time — and Auditd is left healthy and running. Audit rules defined in `rules.d` survive the restarts; rules added at runtime with `auditctl` do not, but that is already true of the single restart the agent does today, verified with the stock binary on the same host.

**On re-probing when the socket is missing.** When Auditd is running but the socket is absent, the probe starts over from the version-preferred candidate instead of reusing the configuration already on disk, which costs an extra Auditd restart on a host that had already been probed. This is deliberate: after a probe that failed on every candidate, the file on disk holds the last and least preferred one, so preferring it would make the agent latch onto the wrong template as soon as the host becomes viable — and "the socket exists and accepts connections" cannot detect a wrong event format. The case is narrow (with Auditd stopped the agent exits earlier with `(6923)`, and with the socket up the live-socket path takes over), so the extra restart is preferred over the risk.

**Note on a separate blocker.** Amazon Linux 2023 ships `-a task,never` by default, which suppresses syscall auditing and makes the who-data health check fail with `(6642): Audit health check couldn't be completed correctly.` — independently of this bug. The `(6636)` failure was reproduced and fixed with that rule already removed, so the two are orthogonal, but on a stock AL2023 host this fix alone does not yield working who-data until that rule is dealt with (same as #35986).

### Artifacts Affected

`wazuh-syscheckd` on the Linux agent only. The whole change is inside `#ifdef __linux__` + `#ifdef ENABLE_AUDIT`, which is enabled for `TARGET=agent` on Linux; the manager does not build `syscheckd`. Windows who-data and the eBPF provider are untouched.

### Configuration Changes

None. No new or changed `ossec.conf` options; `restart_audit` keeps its current meaning, and with it disabled the behaviour is the same as before the fix (`(6910): Audit plugin configuration was modified. You need to restart Auditd. Who-data will be disabled.`, no probing, since Auditd cannot be restarted).

### Documentation Updates

None required in this repository — no user-facing configuration changed. Worth considering separately: the who-data documentation does not mention that a distribution's audit rebuild may not accept the event format argument, nor that `-a task,never` blocks who-data on Amazon Linux 2023 and Fedora.

### Tests Introduced

`src/unit_tests/syscheckd/whodata/test_syscheck_audit.c`:

- 9 new cases for `audisp_get_candidates()`, one per version bucket — including **3.1.5, the one this issue is about** — plus unknown version, the legacy `/etc/audisp/plugins.d` directory, no plugins directory at all, and that `max` is honoured. They compare each candidate against a full copy of the expected template, rather than matching a fragment: the two argument template is a suffix of the builtin one, so a substring check cannot tell them apart and would pass on a wrong result.
- 1 new case for `set_auditd_config_template()` with an explicitly supplied configuration.
- 3 new cases for `audisp_configuration_is_known()`: the file on disk matches one of the templates, it matches none of them, and there is no file at all.
- `-Wl,--wrap,usleep` added to `SYSCHECK_AUDIT_FLAGS`. Without it the `usleep` mock was never linked, so the two timeout tests slept for real: the binary took 15.1 s of wall clock for 0.07 s of user time while still passing. It now runs in 74 ms.
- 6 new cases for `configure_and_connect_audit_socket()`: no plugins directory, a stale socket being unlinked before probing, a socket that is **not** unlinked when `restart_audit` is disabled (any unexpected `unlink` fails the test), the fallback path with its `(6054)`, every candidate failing with a single `(6636)`, and the probe carrying on after a candidate could not be applied. The function was moved out of the `LCOV_EXCL` region to make this possible.
- The 13 existing `set_auditd_config()` cases now call `set_auditd_config_template()` with the configuration the code would pick for the version each of them was mocking, and drop the `IsDir` + `auditctl -v` mocks that the new `audisp_get_candidates()` cases cover. Renamed to `test_set_auditd_config_template_*` so the names match what they exercise.
- `test_set_auditd_config_wrong_audit_version` removed: it is now covered by `test_audisp_get_candidates_no_plugins_dir`.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/src/shared/include/error_messages/information_messages.h
+++ b/src/shared/include/error_messages/information_messages.h
@@ -70,6 +70,7 @@
 #define FIM_EBPF_LSM_ACTIVE                 "(6051): BPF LSM is active in the running kernel; using LSM hooks for create/modify/delete events."
 #define FIM_EBPF_LSM_INACTIVE               "(6052): BPF LSM is not active (not present in /sys/kernel/security/lsm); falling back to kprobe hooks. To enable LSM-based capture append 'bpf' to the kernel boot parameter 'lsm=' and reboot."
 #define FIM_EBPF_LSM_DPATH_FALLBACK         "(6053): BPF LSM load failed with the bpf_d_path-based variants; retrying with the manual path walker (some kernels, e.g. Amazon Linux 2/2023, disallow bpf_d_path for these hooks)."
+#define FIM_AUDIT_FALLBACK_CONFIGURATION    "(6054): The audisp plugin configuration expected for the installed audit version did not create the socket '%s'. Who-data started with an alternative plugin configuration."
 
 /* wazuh-logtest information messages */
 #define LOGTEST_INITIALIZED                 "(7200): Logtest started"
--- a/src/syscheckd/include/syscheck.h
+++ b/src/syscheckd/include/syscheck.h
@@ -530,11 +530,40 @@ void audit_create_rules_file();
 void audit_rules_to_realtime();
 
 /**
- * @brief Set Auditd socket configuration
+ * @brief Build the list of audisp plugin configurations to try, ordered by preference
  *
- * @return 0 on success, -1 on error
+ * @param plugin_dir [out] Directory holding the audisp plugin configuration files
+ * @param templates [out] Array to fill with the configuration templates, most preferred first
+ * @param max Size of the templates array
+ * @return Number of candidates written to templates, 0 if no known plugins directory was found
  */
-int set_auditd_config(void);
+int audisp_get_candidates(const char **plugin_dir, const char **templates, int max);
+
+/**
+ * @brief Move the configuration already on disk to the front of the candidate list
+ *
+ * @param plugin_dir Directory holding the audisp plugin configuration files
+ * @param templates Candidate list to reorder in place
+ * @param count Number of candidates in the list
+ */
+void audisp_prefer_configuration_on_disk(const char *plugin_dir, const char **templates, int count);
+
+/**
+ * @brief Configure the audisp plugin and connect to the who-data socket, probing the known
+ *        configurations until one of them yields a socket that can be connected to
+ *
+ * @return File descriptor of the connected socket, -1 on error
+ */
+int configure_and_connect_audit_socket(void);
+
+/**
+ * @brief Write the given audisp plugin configuration and restart Auditd if it changed
+ *
+ * @param plugin_dir Directory holding the audisp plugin configuration files
+ * @param audisp_config Configuration template to render
+ * @return 0 on success, -1 on error, 1 if Auditd could not be restarted
+ */
+int set_auditd_config_template(const char *plugin_dir, const char *audisp_config);
 
 /**
  * @brief Initialize Audit evsents socket
--- a/src/syscheckd/include/syscheck.h
+++ b/src/syscheckd/include/syscheck.h
@@ -540,13 +540,14 @@ void audit_rules_to_realtime();
 int audisp_get_candidates(const char **plugin_dir, const char **templates, int max);
 
 /**
- * @brief Move the configuration already on disk to the front of the candidate list
+ * @brief Whether the audisp plugin configuration on disk is one of the known templates
  *
  * @param plugin_dir Directory holding the audisp plugin configuration files
- * @param templates Candidate list to reorder in place
- * @param count Number of candidates in the list
+ * @param templates Known configuration templates
+ * @param count Number of templates
+ * @return 1 when the file on disk matches one of them, 0 otherwise
  */
-void audisp_prefer_configuration_on_disk(const char *plugin_dir, const char **templates, int count);
+int audisp_configuration_is_known(const char *plugin_dir, const char **templates, int count);
 
 /**
  * @brief Configure the audisp plugin and connect to the who-data socket, probing the known
--- a/src/syscheckd/include/syscheck.h
+++ b/src/syscheckd/include/syscheck.h
@@ -562,7 +562,9 @@ int configure_and_connect_audit_socket(void);
  *
  * @param plugin_dir Directory holding the audisp plugin configuration files
  * @param audisp_config Configuration template to render
- * @return 0 on success, -1 on error, 1 if Auditd could not be restarted
+ * @return 0 when the configuration is in place, 1 when it was written but Auditd was not
+ *         restarted because restart_audit is disabled, -1 on error (a failed Auditd restart
+ *         included)
  */
 int set_auditd_config_template(const char *plugin_dir, const char *audisp_config);
 
--- a/src/syscheckd/src/whodata/syscheck_audit.c
+++ b/src/syscheckd/src/whodata/syscheck_audit.c
@@ -25,6 +25,8 @@
 #define AUDIT_CONF_LINK             "af_wazuh.conf"
 #define BUF_SIZE OS_MAXSTR
 #define MAX_CONN_RETRIES 5          // Max retries to reconnect to Audit socket
+#define AUDIT_SOCKET_WAIT_MS 5000   // Max time to wait for the audisp plugin to create the socket
+#define AUDIT_SOCKET_POLL_MS 100    // Polling interval while waiting for the socket
 
 // Global variables
 pthread_mutex_t audit_mutex;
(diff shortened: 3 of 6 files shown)


cborla:
Reviewed the diff in full (including the parts unchanged by this PR, to check what still calls into the touched functions) and cross-checked every quantitative claim in the description against the actual code. Summary: version coverage is complete and evidenced, and the change is provably isolated to the `audit` provider on Linux.

## Version coverage — every relevant bucket now has real evidence, not just the version-comparison logic

| Audit version | Path taken | Evidence |
|---|---|---|
| 2.8.1 (Amazon Linux 2) | Legacy plugins dir (`/etc/audisp/plugins.d`), version never consulted | Real host, `(6019)` directly, one cycle, zero `(6054)` |
| 3.0.6 / 3.0.7 (AL2023 / Ubuntu 22.04) | `builtin_af_unix` bucket (`< 3.1.1`) | Real host, one cycle, zero `(6054)` |
| 3.1.5 (Amazon Linux 2023) — the reported bug | `>= 3.1.1, < 4.0.0` bucket, first candidate fails, falls back | Real host, two cycles, `(6054)` fires, who-data comes up |
| 3.1.5 (AlmaLinux 9.3, upstream build) | Same bucket, first candidate works | Real host, one cycle, zero `(6054)` — same version string as the row above, opposite outcome, which is exactly why the version alone can't decide |
| 4.1.2 (Ubuntu 26.04) | `>= 4.0.3` bucket | Real host, one cycle, zero `(6054)` |

The 2.x/legacy-dir path and the `< 3.1.1` builtin bucket are worth calling out specifically since they're what a reviewer would worry about breaking: both still resolve to exactly one candidate, so the new probing/fallback machinery never activates for them — behavior there is unchanged from before this PR, not just "not expected to regress." `3.1.1-3.1.4` and `4.0.0-4.0.2` are the only two buckets without a dedicated real-host run; both select the two-argument template, which is the same one Amazon Linux 2023 already validates as working, so what's untested there is the version comparison itself, which the 9 new `audisp_get_candidates()` unit tests (one per bucket) cover directly.

## Isolation from other providers/OSs — verified by grep, not by reading the PR description

- `grep`'d every file under `syscheckd/` for the changed function names (`set_auditd_config_template`, `configure_and_connect_audit_socket`, `audit_socket_connect`, `audisp_get_candidates`, `audisp_configuration_is_known`): the only files referencing them are `syscheck_audit.c` itself and the `syscheck.h` prototypes. `win_whodata.c` and the eBPF provider (`ebpf_whodata.cpp/.hpp/.h`) have zero references.
- `syscheckd/src/main.c:467-480` (untouched by this diff) gates `AUDIT_PROVIDER` and `EBPF_PROVIDER` in separate, mutually exclusive `if` blocks — the code this PR changes only ever runs when `<whodata_provider>audit</whodata_provider>` is explicitly configured.

## Also confirmed since the last pass

- Both Low-severity suggestions from the earlier review round are applied on the current head (`dfb6521a`): the three tuning constants (`AUDIT_SOCKET_WAIT_MS`/`AUDIT_SOCKET_POLL_MS`/`AUDIT_SOCKET_CONNECT_RETRIES`) moved to `syscheck_audit.h` alongside `MAX_AUDISP_CANDIDATES`, and the unit tests now reference those real constants instead of hand-duplicated local copies. The `switch` in `configure_and_connect_audit_socket()` has an explicit `case 1:` now, with `default:` kept underneath as a fail-closed backstop for any future unexpected return value.
- All 8 message IDs quoted in the description (including the new `(6054)`) match the actual header definitions verbatim, `(6054)` doesn't collide with anything, and the "64/64 tests" claim matches the real `cmocka_unit_test` count in the file.

No open findings. LGTM to merge.

cborla (approved):
LGTM
- [x] Code changes reviewed

PR #38558, opened 2026-08-24
Add missing null-check for m_spDBSync in Syscollector::updateMetadataValue()

### LINKED ISSUE #38352: Syscollector updateMetadataValue missing m_spDBSync null-check
## Description

Follow-up from code review on PR #38312 (#38203 fix). That PR added a null check for `m_spDBSync` in `Syscollector::getMetadataValue()` (`src/wazuh_modules/syscollector/src/syscollectorImp.cpp`), because `init()` can throw while constructing `DBSync` (e.g. `sqlite: database is locked`, when a previous agent process still holds `local.db`), leaving `m_spDBSync` null while `getMetadataValue()` had no guard and was reachable from `query()`'s `get_first_sync_completed` command — an unconditional SEH access violation on Windows that the surrounding `catch (const std::exception&)` cannot handle, killing the whole agent process.

`Syscollector::updateMetadataValue()` (same file, ~line 4617) has the identical unguarded pattern and was not touched by that PR:

```cpp
bool Syscollector::updateMetadataValue(const std::string& key, int64_t value)
{
    auto emptyCallback = [](ReturnTypeCallback, const nlohmann::json&) {};

    try
    {
        ...
        m_spDBSync->selectRows(selectQuery.query(), selectCallback);   // line 4641, no null check
        ...
        DBSyncTxn txn
        {
            m_spDBSync->handle(),                                      // line 4646, no null check
            ...
```

Every other `m_spDBSync` user in the file (`getMetadataValue()`, `getMaxVersion()`, `setVersion()`, `fetchAllFromTable()`, `hasDataInTable()`, `clearTablesForIndices()`, `deleteDatabaseUnlocked()`) null-checks before use. This one doesn't.

## Why it doesn't crash today, and why that's fragile

Traced both call paths into `updateMetadataValue()` that don't go through the module's own scan/sync worker thread (which never starts if `init()` failed, since `start()` bails on `!m_initialized`):

1. `syncModule()` (line ~2350) returns immediately at `if (m_paused || m_stopping.load()) { return {false, {}}; }`, before reaching its `updateMetadataValue()` call at line ~2381.
2. `executeFlushSync()` → `persistVDFirstSyncIfNeeded()` (line ~2717), reached via a `flush` wcom query on `AsyncFlushController`'s own worker thread (which does not hold `m_resourcesMutex` — that mutex from #38312 guards a different, unrelated race). The `updateMetadataValue()` call there is gated by `if (vdResult && !firstSyncDone && !m_stopping.load())`.

Both gates only work because `m_stopping` stays at its constructor default of `true` whenever `init()` throws: `init()` constructs the local `auto dbSync = std::make_unique<DBSync>(...)` *before* the line `m_stopping = false;` further down in the same function. If the `DBSync` constructor throws, `m_stopping = false;` is never reached, so `m_stopping` stays `true` for the rest of the process — and `m_spDBSync` (the member) is never assigned either, so it stays null too.

`m_stopping` was designed to mean "the module is shutting down" (set explicitly by `quiesce()`), not "the module was never successfully initialized." It happens to cover both today only by accident of statement order in `init()`. Nothing documents or enforces that coupling, and no test exercises the init-failure path (PR #38312's own description says so explicitly).

Any of these independent, locally-reasonable changes would silently reopen the same null-pointer crash in `updateMetadataValue()`:
- Moving `m_stopping = false;` earlier in `init()` (e.g. ahead of the `DBSync` construction).
- Relaxing the `!m_stopping.load()` condition in `persistVDFirstSyncIfNeeded()` — e.g. if a future change decides the VD-first-sync marker should persist even while stopping.
- Relaxing the `if (m_paused || m_stopping.load())` early return in `syncModule()`.

## Suggested fix

Add the same guard used everywhere else in the file:

```cpp
bool Syscollector::updateMetadataValue(const std::string& key, int64_t value)
{
    if (!m_spDBSync)
    {
        return false;
    }

    auto emptyCallback = [](ReturnTypeCallback, const nlohmann::json&) {};
    ...
```

Purely additive, no locking/threading changes.

Also worth adding: a unit test that constructs `Syscollector` with an `init()` that fails (e.g. a locked/invalid db path) and then drives it through a `flush` query and/or `syncModule()`, asserting no crash and that both `getMetadataValue()` and `updateMetadataValue()` return `false` — closing the coverage gap #38312 acknowledged for this path.

## References
- PR #38312 (`getMetadataValue()` fix)
- Issue #38203 (original crash report)
- `src/wazuh_modules/syscollector/src/syscollectorImp.cpp`: `updateMetadataValue()` (~4617), `getMetadataValue()` (~4574), `syncModule()` (~2350), `persistVDFirstSyncIfNeeded()` (~2717), `init()` (~483)

lchico:
## Description

`Syscollector::updateMetadataValue()` dereferences `m_spDBSync` without checking it for null, unlike its sibling `getMetadataValue()`, which was given exactly that guard in #38312 (fix for #38203, a Windows SEH access violation). `m_spDBSync` is left null whenever `Syscollector::init()` fails to construct `DBSync` (e.g. an unopenable or locked database file) — `init()`'s very first statement constructs `DBSync`, before any member is assigned, so a throw there leaves the object in exactly that state.

This doesn't currently crash in the field: `updateMetadataValue()`'s only two callers, `syncModule()` and `persistVDFirstSyncIfNeeded()`, both bail out early whenever the module is marked as stopping, and that flag happens to stay at its default "stopping" value whenever `init()` fails — purely because `init()` clears it only after constructing `DBSync`, not before. That's an accidental side effect of statement ordering, not an intentional safeguard, and several independent, locally-reasonable future changes to either caller or to `init()` itself would each silently reopen this null-pointer crash.

Closes #38352

## Proposed Changes

- Added a null-check on `m_spDBSync` at the top of `updateMetadataValue()`, returning `false` safely instead of dereferencing a null pointer — the same pattern already used by `getMetadataValue()` and every other `m_spDBSync` consumer in this file.
- Added a unit test (`SyscollectorImpTest.updateMetadataValueNullDBSyncAfterFailedInit`) that deliberately fails `init()` (pointing it at an unopenable database path) and then calls `getMetadataValue()`/`updateMetadataValue()` directly, independent of whichever production callers currently happen to gate them. This closes the test-coverage gap explicitly called out in the issue.

## Results and Evidence

Verification procedure:
1. Deliberately failed `Syscollector::init()` by pointing it at a database path that cannot be opened, reproducing the same "left with a null `m_spDBSync`" state that a locked database file produces in the field.
2. Called `getMetadataValue()` directly on the resulting instance — confirmed it returns `false` safely (already fixed by #38312).
3. Called `updateMetadataValue()` directly on the same instance — before this fix, this dereferences the null pointer.
4. Ran the module's standard CI-mirroring test/quality pipeline (the same one `.github/workflows` runs for this module: cppcheck, style check, unit tests, coverage, Valgrind memory-leak check, AddressSanitizer) twice: once before the fix, once after.

Before the fix, the new test crashed the test binary with a real, sanitizer-caught null-pointer dereference at the exact unguarded line:

```
[ RUN      ] SyscollectorImpTest.updateMetadataValueNullDBSyncAfterFailedInit
/workspace/src/wazuh_modules/syscollector/src/syscollectorImp.cpp:4659:31: runtime error: member access within null pointer of type 'struct DBSync'
AddressSanitizer:DEADLYSIGNAL
=================================================================
==9616==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x000000ca3b7d bp 0x7ffd71759c80 sp 0x7ffd71759590 T0)
==9616==The signal is caused by a READ memory access.
==9616==Hint: address points to the zero page.
    #0 0xca3b7d in Syscollector::updateMetadataValue(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, long) /workspace/src/wazuh_modules/syscollector/src/syscollectorImp.cpp:4659
    #1 0x813539 in SyscollectorImpTest_updateMetadataValueNullDBSyncAfterFailedInit_Test::TestBody() /workspace/src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp:5898
SUMMARY: AddressSanitizer: SEGV /workspace/src/wazuh_modules/syscollector/src/syscollectorImp.cpp:4659 in Syscollector::updateMetadataValue(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, long)
==9616==ABORTING
1/2 Test #33: syscollectorimp_unit_test ........***Failed  233.65 sec
```

All other tests in the same suite that ran before this one (77 tests, `defaultCtor` through `queryCommandGetVDFirstSyncCompletedIgnoresZeroTimestamp`) passed cleanly, confirming the crash is specific to this exact scenario and not an artifact of the test setup. The crash terminated the test binary outright, so the pipeline's test stage failed and stopped there — the later coverage/memory/sanitizer stages never ran in this pass.

After applying the fix, the full pipeline passed end-to-end, including the stages that couldn't run before:

```
[MakeTarget: PASSED]
<wazuh_modules/syscollector>[All tests: PASSED]<wazuh_modules/syscollector>
[Lines Coverage 90.1%: PASSED]
[Functions Coverage 90.1%: PASSED]
[syscollectorimp_unit_test : PASSED]
<wazuh_modules/syscollector>[Memory leak check: PASSED]<wazuh_modules/syscollector>
[ASAN: PASSED]
<wazuh_modules/syscollector>[RTR: PASSED]<wazuh_modules/syscollector>
```

No coverage regression (90.1% lines/functions, matching the pre-existing baseline for this module), and both the Valgrind memory-leak check and AddressSanitizer pass cleanly with the guard in place.

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X
- Memory tests for Linux
  - [ ] Coverity
  - [x] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer
  - [ ] Test run in parallel
- Wazuh server API/Framework

### Artifacts Affected

N/A — agent-side module source and its unit tests only; no new executables, default configuration files, or packages.

### Configuration Changes

N/A

### Tests Introduced

Added `SyscollectorImpTest.updateMetadataValueNullDBSyncAfterFailedInit` to `syscollectorImp_test.cpp`. Scope: constructs `Syscollector`, deliberately fails `init()` so `m_spDBSync` is left null, then exercises `getMetadataValue()` and `updateMetadataValue()` directly — independent of `syncModule()`/`persistVDFirstSyncIfNeeded()`, the only two production callers, since both currently happen to gate this scenario off and would mask the defect if the test went through them instead. This closes the coverage gap the issue identifies: no existing test previously exercised an `init()`-failure path for either metadata accessor.

Note on platform coverage: the fix itself is plain C++ with no platform-conditional code, so the same guard runs identically on every supported platform. The new test, however, is skipped on Windows — the technique it uses to force `init()` into the failed state (pointing it at a database path that cannot be opened) does not fail cleanly there; it crashes at a lower level than a catchable exception under the Windows test environment, so the test can't set up the scenario safely on that platform. That means this specific scenario — `init()` failing and leaving `m_spDBSync` null — is verified by automated test on Linux only; Windows coverage relies on the fix being the same code path, not on direct test evidence for this platform. Low-severity, non-blocking, but worth flagging as the one real gap left by this change. 

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

### DIFF
--- a/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
+++ b/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
@@ -4634,6 +4634,14 @@ bool Syscollector::getMetadataValue(const std::string& key, int64_t& value)
 
 bool Syscollector::updateMetadataValue(const std::string& key, int64_t value)
 {
+    // See getMetadataValue()'s identical guard above: init() can leave m_spDBSync null (e.g. a
+    // locked/unopenable db), and a null dereference here raises an SEH access violation on
+    // Windows, which the catch below cannot handle, and it takes the whole agent process down.
+    if (!m_spDBSync)
+    {
+        return false;
+    }
+
     auto emptyCallback = [](ReturnTypeCallback, const nlohmann::json&) {};
 
     try
--- a/src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp
+++ b/src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp
@@ -5818,3 +5818,84 @@ TEST_F(SyscollectorImpTest, queryCommandGetVDFirstSyncCompletedIgnoresZeroTimest
 
     Syscollector::instance().destroy();
 }
+
+// Test-only access to Syscollector's private getMetadataValue()/updateMetadataValue(), without
+// modifying Syscollector itself. Naming a private member as the non-type template argument of an
+// explicit template instantiation is not access-checked at that point (a long-standing, widely
+// used C++ idiom for reaching private members from test code with zero production-code changes).
+namespace syscollector_test_access
+{
+    template <typename Tag, typename Tag::type Member>
+    struct Rob
+    {
+        friend typename Tag::type stealAccess(Tag)
+        {
+            return Member;
+        }
+    };
+
+    struct GetMetadataValueTag
+    {
+        using type = bool (Syscollector::*)(const std::string&, int64_t&);
+        friend type stealAccess(GetMetadataValueTag);
+    };
+
+    struct UpdateMetadataValueTag
+    {
+        using type = bool (Syscollector::*)(const std::string&, int64_t);
+        friend type stealAccess(UpdateMetadataValueTag);
+    };
+
+    template struct Rob<GetMetadataValueTag, &Syscollector::getMetadataValue>;
+    template struct Rob<UpdateMetadataValueTag, &Syscollector::updateMetadataValue>;
+}
+
+// init() leaves m_spDBSync null if constructing DBSync fails (e.g. an unopenable db path,
+// mirroring a locked database file in the field) -- its very first statement constructs DBSync,
+// before any member is assigned, so a throw there leaves the object in exactly that state.
+// Calling both metadata accessors directly here bypasses syncModule()/persistVDFirstSyncIfNeeded(),
+// their only two production callers -- both currently skip this path via m_stopping, but that's an
+// accidental side effect of init()'s statement order, not a deliberate safeguard.
+TEST_F(SyscollectorImpTest, updateMetadataValueNullDBSyncAfterFailedInit)
+{
+#ifdef WIN32
+    GTEST_SKIP() << "Skipping updateMetadataValueNullDBSyncAfterFailedInit test on Windows: opening "
+                 "an unopenable db path crashes the process at a lower level than a catchable "
+                 "std::exception under Wine, instead of the clean throw this test relies on";
+#endif
+    const auto spInfoWrapper {std::make_shared<MockSysInfo>()};
+    EXPECT_CALL(*spInfoWrapper, releaseThreadResources()).Times(testing::AnyNumber());
+
+    bool initThrew = false;
+
+    try
+    {
+        Syscollector::instance().init(spInfoWrapper,
+                                      reportFunction,
+                                      persistFunction,
+                                      logFunction,
+                                      "/nonexistent-dir/local.db",
+                                      "",
+                                      "",
+                                      3600, false);
+    }
+    catch (const std::exception&)
+    {
+        initThrew = true;
+    }
+
+    ASSERT_TRUE(initThrew) << "Expected DBSync construction to throw for an unopenable db path";
+
+    // Unqualified on purpose: stealAccess() is a hidden friend, only reachable via ADL on its
+    // Tag argument's namespace -- explicit qualification (syscollector_test_access::stealAccess)
+    // does not find it.
+    const auto getMetadataValuePtr = stealAccess(syscollector_test_access::GetMetadataValueTag{});
+    const auto updateMetadataValuePtr = stealAccess(syscollector_test_access::UpdateMetadataValueTag{});
+
+    int64_t value = -1;
+    // Guarded: returns false safely instead of dereferencing a null m_spDBSync.
+    EXPECT_FALSE((Syscollector::instance().*getMetadataValuePtr)("test_key", value));
+
+    // Unguarded: dereferences a null m_spDBSync -- crashes without the fix applied above.
+    EXPECT_FALSE((Syscollector::instance().*updateMetadataValuePtr)("test_key", 1));
+}


cborla (approved):
LGTM
- [x] Code changes reviewed

PR #38575, opened 2026-08-25
Fix invalid UTF-8 event loss in engine

### LINKED ISSUE #38561: Invalid UTF-8 byte in a log line makes the indexer reject the document; the event and its alert are both lost
|---|---|---|---|---|
| 5.0.0 | wazuh-manager (analysisd indexer-connector, event serialisation) | Manager | Docker, `public.ecr.aws/wazuh-cicd/wazuh/*:5.0.0-latest` | Amazon Linux 2023 |

## Description

A log line containing a byte sequence that is not valid UTF-8, or a NUL, is decoded and classified normally by the engine and then rejected by the indexer, because the JSON the manager serialises around the raw line is malformed. The event never lands, and neither does the alert it would have raised.

This is a working detection-evasion primitive. Anything that can influence any part of a logged line — a username, a path, a URL, a user agent — can append one byte and keep its activity out of the indexer entirely.

## Configuration

Stock single-node deployment, one agent, one monitored file:

```xml
<localfile>
  <log_format>syslog</log_format>
  <location>/testlogs/plain.log</location>
</localfile>
```

The probe uses `sshd` authentication-failure lines, which decode and raise an alert, so both the event and the alert can be counted. Every case is paired with a control that differs **only** in the byte under test.

Two controls are worth calling out, because they rule out alternative explanations:

- **Nothing may be appended after `ssh2`.** The decoder anchors there, so padding at the end breaks decoding for reasons unrelated to the byte under test. A plain-text control in that position fails exactly like the invalid bytes do.
- **Control characters are not the trigger.** `\x01`, `\x02` and `\x7f` inside the same field survive and index normally, as do valid 2-byte and 3-byte UTF-8 sequences. The trigger is specifically byte sequences that are not valid UTF-8, plus NUL.

## Current behavior

Hostile byte placed **inside** the username (a decoded field), with plain-text and valid-UTF-8 controls in the identical position:

```
OK      control: plain text inside the username    -> 'XX'
MISSING NUL inside
MISSING 0xFF invalid utf-8 inside
MISSING overlong encoding inside            (\xc0\xaf)
MISSING surrogate inside                    (\xed\xa0\x80)
MISSING lone continuation byte inside       (\x80)
OK      control chars inside                       -> 'X\x01\x02X'
OK      DEL inside                                 -> 'X\x7fX'
OK      control: valid 2-byte utf-8 inside         -> 'XéX'
OK      control: valid 3-byte utf-8 inside         -> 'X中X'
```

Scaled up, same line, one byte apart:

```
clean control (10 sent)    events=10  findings/alerts=10
one 0xFF      (50 sent)    events=0   findings/alerts=0
```

The alert is suppressed along with the event.

## Expected behavior

A log line the agent could read is a log line the manager can store. The raw bytes should be validated and represented safely at serialisation time — escaping, replacing or dropping only the offending byte — so that the event and its alert still land.

Whatever is chosen, an event that cannot be stored should be accounted for somewhere an operator can total up, not only as one `WARNING` per occurrence.

## Steps to reproduce

1. Configure a monitored file on an agent as above.
2. Append the same decodable line twice, differing only by one invalid byte inside the username:

```bash
printf 'Aug 24 17:10:00 agent-11 sshd[4242]: Failed password for CLEANXX from 198.51.100.201 port 22 ssh2\n' \
  >> /testlogs/plain.log
printf 'Aug 24 17:10:00 agent-11 sshd[4242]: Failed password for DIRTYX\xffX from 198.51.100.201 port 22 ssh2\n' \
  >> /testlogs/plain.log
```

3. Query for both. The clean one is present with its alert; the other is absent:

```bash
curl -sk -u admin:admin 'https://localhost:9200/wazuh-events*/_count' -H 'Content-Type: application/json' \
  -d '{"query":{"prefix":{"user.name":"CLEAN"}}}'      # 1
curl -sk -u admin:admin 'https://localhost:9200/wazuh-findings*/_count' -H 'Content-Type: application/json' \
  -d '{"query":{"prefix":{"user.name":"CLEAN"}}}'      # 1
curl -sk -u admin:admin 'https://localhost:9200/wazuh-events*/_count' -H 'Content-Type: application/json' \
  -d '{"query":{"prefix":{"user.name":"DIRTY"}}}'      # 0
```

4. Confirm in the manager log, one entry per lost event, each followed by a dump of the whole event:

```bash
docker exec <manager> sh -c 'grep -c json_e_o_f_exception /var/wazuh-manager/logs/wazuh-manager.log'
```

## Logs and root cause

The event is fully decoded and classified before it dies. The manager's own log dumps it, showing the decoder chain and the assigned integration:

```
"integration":{"category":"system-activity","name":"linux",
 "decoders":["decoder/core-wazuh-message/0","decoder/syslog/0","decoder/system-auth/0"]}
```

This is not an unclassified event being dropped by policy. It is a recognised security event. It then fails on the way into the indexer:

```
wazuh-manager-analysisd(indexer-connector): WARNING: Error indexing document
(type mapper_parsing_exception - reason: 'failed to parse field [event.original] of type [keyword]
in document with id '...'. Preview of field's value: ''' - caused by: json_e_o_f_exception -
'Unexpected end-of-input in VALUE_STRING at [Source: REDACTED ...; line: 1, column: 656]')
- Associated event: {"create":{"_index":"wazuh-events-v5-system-activity"}}
```

`json_e_o_f_exception` is the tell: this is not a type mismatch but the indexer's JSON parser running off the end of a string it cannot terminate. The raw log bytes are serialised into the JSON document without being validated or escaped as UTF-8, so one invalid byte truncates the string from the parser's point of view and the whole bulk document is refused.

`event.original` is named because it carries the raw line; `message` carries it too.

### The agent path already handles this correctly

FIM takes the same class of input and handles it the way this path should. Eighteen files were created with hostile names under a monitored directory, one containing `0xFF`. The agent refused that one **at serialisation time**, named the byte and its offset, and logged it at `ERROR`:

```
wazuh-syscheckd: ERROR: [json.exception.type_error.316] invalid UTF-8 byte at index 24: 0xFF
```

The other 17 indexed fine, including names containing a newline, a tab, an RTL override, a zero-width space, 212 characters, emoji, CJK, backslashes and shell metacharacters.

So the same product, using the same JSON library, validates UTF-8 before serialising on the agent side and reports the failure precisely. The manager's event path does neither: it serialises unchecked, and the refusal surfaces later, at the indexer, in terms that describe a JSON parsing problem rather than a rejected log line. The validation logic exists; it is not applied on this path.

## Accounting, measured

Per lost event the manager emits one `WARNING` plus a dump of the entire event JSON. For one 50-line batch:

```
'Error indexing document' warnings         : 61
json_e_o_f_exception occurrences           : 61
failures naming event.original             : 61
lines dumping a lost event's JSON          : 50
```

The loss is visible in the manager log and only there — not in the UI, not to any alert, not on any dashboard. A log source emitting binary-ish content therefore loses its events *and* floods `wazuh-manager.log` with full event dumps.

On the engine counters:

**Established.** `events.discarded`, `events.discarded.prefilter` and `events.discarded.postfilter` sat at **0** through every test, including batches of 50, 37 and 23 lost events. The counter family that would naturally account for a dropped event never moves.

**Not established.** Whether `events.unclassified` absorbs them cannot be answered at this sampling rate. The metrics stream is scraped every 10 minutes and the counter drifts on its own from ordinary agent traffic. Idle baseline, nothing injected:

```
16:55:38  events.unclassified = 1144
17:05:38  events.unclassified = 1175     +31 with no injection at all
```

With exactly 23 invalid-byte lines injected at 17:07:

```
17:15:38  events.unclassified = 1215     +40, i.e. +9 over the idle drift
```

+40 against a +31 baseline does not resolve 23 events either way. A 37-line run was inconclusive for the same reason — unrelated test traffic fell in the same window. Counting these as unclassified would be wrong regardless, since the dump above shows them fully classified, but that is an argument about correctness rather than a measurement. Settling it would require reading the counter directly from the engine rather than from the 10-minute stream, on an otherwise idle manager.

## Impact

Detection evasion, achievable without touching Wazuh. Anyone who can get one byte of their choosing into a line that some application logs — a username, a filename, a URL path, a header — can ensure the resulting event and alert never reach the indexer. The activity is invisible in the UI and unalertable, and the only trace is a `WARNING` in a server log naming a JSON parsing problem rather than a lost security event.

Second-order: 50 lost events produced 50 full event dumps in `wazuh-manager.log`. A noisy source turns this into a log-growth problem on the manager on top of the data loss.

dev-rehaann:
## Description

Closes #38561.

Invalid UTF-8 or embedded NUL bytes in a raw log could make RapidJSON serialization fail, so the indexer rejected the event and its generated alert.

## Proposed Changes

Sanitize untrusted event location and message fields at the legacy/public parser boundary. Valid UTF-8 is preserved; each invalid byte or NUL is replaced with U+FFFD before JSON serialization.

### Results and Evidence

The engine now produces valid JSON for malformed input instead of losing the event. Valid multibyte characters remain unchanged.

- [x] Linux compilation
- [x] Engine unit test run
- [ ] Windows (not applicable to this Linux engine test run)
- [ ] macOS (not applicable to this Linux engine test run)
- [ ] ASAN/TSAN (left to repository CI)

Built the Wazuh base unit-test target with GCC 14 and ran the EventParser filter: 28/28 tests passed.

### Artifacts Affected

Wazuh 5.x engine event parser and indexed event JSON.

### Configuration Changes

N/A.

### Tests Introduced

A parameterized legacy-event case covers embedded NUL, invalid lead/continuation bytes, overlong and surrogate sequences, and preservation of valid two- and three-byte UTF-8.

## Review Checklist

- [x] Code changes reviewed
- [x] Configuration changes documented as N/A
- [x] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 
 | Issue | Comment |
 |-------|---------|
+| [#38561](https://github.com/wazuh/wazuh/issues/38561) | Preserve events containing invalid UTF-8 by replacing invalid bytes before JSON serialization. |
 
 ### Agent
 
--- a/src/engine/source/base/src/eventParser.cpp
+++ b/src/engine/source/base/src/eventParser.cpp
@@ -18,6 +18,69 @@ namespace
 
 constexpr size_t LOCATION_OFFSET = 2; // Given the "q:" prefix.
 
+constexpr std::string_view UTF8_REPLACEMENT = "\xEF\xBF\xBD";
+
+std::string sanitizeUtf8(std::string_view input)
+{
+    std::string output;
+    output.reserve(input.size());
+
+    for (size_t i = 0; i < input.size();)
+    {
+        const auto byte = static_cast<unsigned char>(input[i]);
+        size_t length = 0;
+
+        if (byte > 0 && byte <= 0x7F)
+        {
+            length = 1;
+        }
+        else if (byte >= 0xC2 && byte <= 0xDF && i + 1 < input.size() &&
+                 (static_cast<unsigned char>(input[i + 1]) & 0xC0) == 0x80)
+        {
+            length = 2;
+        }
+        else if (byte >= 0xE0 && byte <= 0xEF && i + 2 < input.size())
+        {
+            const auto second = static_cast<unsigned char>(input[i + 1]);
+            const auto third = static_cast<unsigned char>(input[i + 2]);
+            if ((third & 0xC0) == 0x80 &&
+                ((byte == 0xE0 && second >= 0xA0 && second <= 0xBF) ||
+                 (byte == 0xED && second >= 0x80 && second <= 0x9F) ||
+                 (((byte >= 0xE1 && byte <= 0xEC) || (byte >= 0xEE && byte <= 0xEF)) &&
+                  (second & 0xC0) == 0x80)))
+            {
+                length = 3;
+            }
+        }
+        else if (byte >= 0xF0 && byte <= 0xF4 && i + 3 < input.size())
+        {
+            const auto second = static_cast<unsigned char>(input[i + 1]);
+            const auto third = static_cast<unsigned char>(input[i + 2]);
+            const auto fourth = static_cast<unsigned char>(input[i + 3]);
+            if ((third & 0xC0) == 0x80 && (fourth & 0xC0) == 0x80 &&
+                ((byte == 0xF0 && second >= 0x90 && second <= 0xBF) ||
+                 (byte == 0xF4 && second >= 0x80 && second <= 0x8F) ||
+                 (byte >= 0xF1 && byte <= 0xF3 && (second & 0xC0) == 0x80)))
+            {
+                length = 4;
+            }
+        }
+
+        if (length > 0)
+        {
+            output.append(input.substr(i, length));
+            i += length;
+        }
+        else
+        {
+            output.append(UTF8_REPLACEMENT);
+            ++i;
+        }
+    }
+
+    return output;
+}
+
 } // namespace
 
 Event parseLegacyEvent(std::string_view rawEvent, const json::Json& agentMetadata)
@@ -83,11 +146,11 @@ Event parseLegacyEvent(std::string_view rawEvent, const json::Json& agentMetadat
         throw std::runtime_error(fmt::format("Agent metadata merge failed: {}", ex.what()));
     }
 
-    parseEvent->setString(rawLocation, EVENT_LOCATION_ID);
+    parseEvent->setString(sanitizeUtf8(rawLocation), EVENT_LOCATION_ID);
 
     // Set the original event message.
     auto msgView = rawEvent.substr(separatorPos + 1);
-    parseEvent->setString(msgView, EVENT_MESSAGE_ID);
+    parseEvent->setString(sanitizeUtf8(msgView), EVENT_MESSAGE_ID);
 
     return parseEvent;
 }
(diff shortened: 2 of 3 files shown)


cborla:
Thanks for this, @dev-rehaann
Solid diagnosis and good test coverage, we're addressing #38561 from the agent side instead, in #38590: filtering invalid UTF-8/NUL in SendMSGAction() before an event ever leaves the agent, so it never reaches the manager at all.

PR #38584, opened 2026-08-25
Replace order-preserving `std::vector::erase` with swap-and-pop where element order doesn't matter

### LINKED ISSUE #38143: Replace order-preserving std::vector::erase with swap-and-pop where element order doesn't matter

matigarciadev:
## Description

Several `std::vector` erase sites in the codebase use the standard order-preserving `erase()` (or `remove_if`+`erase`), which is O(n) because it shifts every following element down by one slot — even though the vector's element order is never semantically observed by any caller. This PR replaces the shifting erase with an O(1) swap-with-last-element + `pop_back()` ("swap-remove") pattern at the confirmed order-irrelevant sites.

Closes #38143

## Proposed Changes

- `RocksDBWrapper::deleteAll()` and `deleteAll(const std::string&)` (`src/shared_modules/utils/rocksDBWrapper.hpp`): rewrote both to swap-remove. Column access always goes through `m_columnsIndex` (a name→index map), which is rebuilt from scratch after every erase, so vector position is never meaningful. The in-loop erase in `deleteAll()` required care: after a swap, the iterator is not advanced, since the swapped-in element hasn't been visited yet.
- Added `base::utils::eraseFirstBySwap<T, Predicate>` (`src/engine/source/base/include/base/utils/vectorHelpers.hpp`): a generic, header-only, directly unit-testable helper (find-by-predicate + swap-with-back + `pop_back()`, self-swap-safe on the last element). This follows the engine's existing convention for testing internal logic — a `namespace detail`/plain free function tested via `#include`, no class-private access needed.
- `IocSync::removeIOCTypeFromSync()` and `CMSync::removeSpaceFromSync()` (`src/engine/source/iocsync/src/iocsync.cpp`, `src/engine/source/cmsync/src/cmsync.cpp`): now call `base::utils::eraseFirstBySwap`, preserving the exact same throw-on-not-found behavior/message as before. Both sync-list lookups are done by field (`getIocType()`/`getOriginSpace()`), never by position, so order was never semantically observed.
- Out of scope, deliberately: `Subject::detach()` in `base/utils/observer.hpp` / `shared_modules/utils/observer.hpp` was investigated as a candidate but excluded. Analysis found zero live consumers of `Subject<T>` anywhere in the current tree (the `Provider<T>` wrapper that would use it is never instantiated) — the change would be safe but pointless on genuinely dead code, and is left for a possible separate cleanup ticket rather than folded into this perf fix.
- Removed `engine/source/base/include/base/utils/observer.hpp`, `provider.hpp`, and `subscriber.hpp`: a `Subject`/`Observer` pub-sub trio left over from the removed `shared_modules/router` skeleton (June 2023). Confirmed via repo-wide search that none of the three is included by anything else in the engine tree, and no `Subject` anywhere in the codebase is ever attached to more than zero observers — dead code, not active infrastructure, so it was removed rather than carried forward with an unrelated erase-pattern change.

### Results and Evidence

- Test results:
  - `utils_utest --gtest_filter='RocksDBWrapperTest.*'`: 42/42 passed.
  - `base_utest` full suite: 391/392 passed, 1 skipped (`PrivSepSetUserTest.ThrowsWhenNotPermitted`, pre-existing/environmental, unrelated to this change). `VectorHelpersTest.*`: 4/4 passed.
  - `iocsync_utest`: 4/4 passed (unchanged count — `removeIOCTypeFromSync` is a private method not reachable from any public API in production today, a pre-existing condition; its algorithmic coverage now lives in `VectorHelpersTest`).
  - `cmsync_utest`: 36/36 passed (unchanged count, same reasoning as above for `removeSpaceFromSync`).

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Log syntax and correct language review — N/A, no log changes

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests _(Wazuh v4.x)_ — N/A, not decoder/rule related

  - [ ] Test run in parallel

- Wazuh server API/Framework
  - [ ] Run API Integration Tests — N/A, no API changes

### Artifacts Affected

N/A — internal library/engine code paths only; no new executables, default configuration files, or packages.

### Configuration Changes

N/A — no configuration parameters, defaults, or backward-compatibility impact.

### Tests Introduced

- `src/shared_modules/utils/tests/rocksDBWrapper_test.cpp`: `TestDeleteAllMultipleColumnFamilies` (3+ columns, exercises the swap-brings-an-unvisited-element-forward path), `TestDeleteAllLastColumnIsLastElement` (self-swap regression), `TestDeleteAllSpecificColumnFamilyMultipleColumns`, `TestDeleteAllSpecificColumnFamilyLastElement`.
- `src/engine/source/base/test/src/unit/utils/vectorHelpers_test.cpp` (new): 4 cases for `base::utils::eraseFirstBySwap` — removes a middle element, returns `false`/leaves the vector untouched on no match, removes the last element without corruption (self-swap), removes the only element.

## Review Checklist

- [x] Code changes reviewed
- [x] Configuration changes documented — N/A, none
- [ ] N/A — no additional checklist items


NahuFigueroa97 (dismissed):
LGTM!

LucioDonda (dismissed):

juliancnn (approved):
LGTM!

PR #38604, opened 2026-08-26
Ignore obsolete 4.x configuration elements instead of rejecting them

### LINKED ISSUE #38563: Obsolete 4.x config elements are fatal in 5.0.0: manager won't start with `<command>`, and the agent silently keeps its old configuration when `<labels>` aborts a restart
|---|---|---|---|---|
| 5.0.0 | wazuh-manager (configuration parser) | Manager | Docker, `public.ecr.aws/wazuh-cicd/wazuh/*:5.0.0-latest` | Amazon Linux 2023 |

## Description

Active Response moved in 5.0.0: it is no longer driven from the manager's configuration file, and the Server API no longer exposes an endpoint for it. Half of the old configuration surface is still parsed, and the two halves behave differently:

- A top-level `<command>` block — how a response was **defined** in 4.x — is rejected as an invalid element and **the manager refuses to start at all**.
- An `<active-response>` block — how a response was **referenced** — is still accepted, and does nothing.

A 4.x manager configuration carried into 5.0.0 therefore takes the manager down, and the error names `command` without saying that the capability moved or where it went.

The same class of bug exists on the agent with `<labels>`, where it is worse in practice: the restart aborts, the old daemons keep running with the old configuration, and `wazuh-control status` reports the agent as healthy.

Stock single-node deployment. The shipped manager configuration contains **no** `<command>` or `<active-response>` blocks, so this is only reachable by adding them — which is exactly what a migration from 4.x does.

The agent side is unaffected and still ships the receiving half enabled:

```xml
<active-response>
  <disabled>no</disabled>
  <ca_store>etc/wpk_root.pem</ca_store>
  <ca_verification>yes</ca_verification>
</active-response>
```

```
/var/ossec/active-response/bin/block-ip
/var/ossec/active-response/bin/disable-account
```

Note the root element differs between the two: the manager's configuration is wrapped in `<wazuh_config>`, the agent's still in `<ossec_config>`. Editing the wrong one closed produces the same startup failure with an even less specific message: `(1226): Error reading XML file ... (line 0)`.

### Manager: `<command>` is fatal, `<active-response>` is accepted and inert

A minimal 4.x-style pair, appended to the shipped manager configuration:

```xml
<command>
  <name>ar-test</name>
  <executable>block-ip</executable>
  <timeout_allowed>no</timeout_allowed>
</command>

<active-response>
  <disabled>no</disabled>
  <command>ar-test</command>
  <location>local</location>
  <rules_id>5710</rules_id>
</active-response>
```

The manager does not come up:

```
2026/08/24 17:48:34 wazuh-manager-authd: ERROR: (1230): Invalid element in the configuration: 'command'.
```

```
docker ps      wazuh-ui-battery-wazuh.manager   Up 2 minutes (unhealthy)
ps | grep -c wazuh                              0
```

Removing only the `<command>` block and keeping `<active-response>` — including its `<command>ar-test</command>` child, its `<location>` and its `<rules_id>`:

```
docker ps      wazuh-ui-battery-wazuh.manager   Up 15 seconds (healthy)
ps | grep -c wazuh                              6
new (1230) errors                               0
```

The parser accepts a block that names a response, and rejects the block that would define one.

### Agent: `<labels>` aborts the restart, and status still reports healthy

`<labels>`, a valid top-level agent element in 4.x, is rejected the same way. `labels-config.c` no longer exists in the source; the name survives only inside `localfile-config.c` with an unrelated meaning.

```
wazuh-agentd: ERROR: (1230): Invalid element in the configuration: 'labels'.
wazuh-agentd: ERROR: (1202): Configuration error at 'etc/ossec.conf'.
wazuh-agentd: ERROR: (1215): No client configured. Exiting.
wazuh-agentd: Configuration error. Exiting
```

The manager at least stops, which is impossible to miss. **The agent keeps running.** `wazuh-control restart` detects the error before stopping anything, aborts, and leaves the previously running daemons untouched with the previous configuration in effect. Three consecutive restart attempts over sixteen minutes changed nothing — every process predates the first attempt:

```console
$ ps -eo etimes,comm | grep wazuh
   1072 wazuh-execd
   1072 wazuh-agentd
   1071 wazuh-syscheckd
   1071 wazuh-logcollec
   1071 wazuh-modulesd
```

And the status command reports a healthy agent throughout, because those stale processes genuinely are alive:

```console
$ /var/ossec/bin/wazuh-control status
wazuh-modulesd is running...
wazuh-logcollector is running...
wazuh-syscheckd is running...
wazuh-agentd is running...
wazuh-execd is running...
```

Not stopping a working agent because its new configuration is bad is the right call. Reporting it as healthy while the configuration on disk is invalid and not in effect is not: the only signal is text on the restart command's stdout, which a scripted or automated restart discards.

<details>
<summary>How this presents to an operator</summary>

This cost an hour of the session. A log source was added, the agent restarted, and several experiments concluded that logcollector was ignoring new `<localfile>` entries. It was not — the restarts had never happened, and every status check reported the agent as fine. Once the invalid element was removed and the agent actually restarted, the new source was picked up within a minute:

```
before   9 files tracked, /newsrc absent
after   10 files tracked, /newsrc/app.log present
```

</details>

## Expected behavior

- An obsolete configuration element should not stop the manager from starting. If failing fast is the intent, it should fail the same way for both halves rather than accepting one and dying on the other.
- `wazuh-control status` should report that the on-disk configuration is invalid and not in effect, instead of reporting stale processes as healthy.
- Message 1230 should say where the capability went. `Invalid element in the configuration: 'command'` gives an operator migrating from 4.x nothing to act on; naming Active Response and pointing at the alerting path would.

## Steps to reproduce

```bash
# append a 4.x-style command block to the manager configuration
#   <command><name>ar-test</name><executable>block-ip</executable></command>
# note the root element is <wazuh_config>, not <ossec_config>

docker restart <manager>
docker ps                      # unhealthy
docker exec <manager> sh -c 'ps -eo comm | grep -c wazuh'          # 0
docker exec <manager> sh -c 'grep 1230 /var/wazuh-manager/logs/wazuh-manager.log | tail -1'
# ERROR: (1230): Invalid element in the configuration: 'command'.

# now remove only the <command> block, keep <active-response> with its <command> child
docker restart <manager>
docker ps                      # healthy
```

Agent side: add a top-level `<labels>` block to `etc/ossec.conf`, run `wazuh-control restart`, then compare `ps -eo etimes,comm` before and after and run `wazuh-control status`.

## Impact

Any manager configuration migrated from 4.x with Active Response in it prevents the manager from starting. The failure is total: no daemons, the container reports unhealthy after two minutes, and the only diagnostic is one line naming an element without explaining why it is no longer valid or what replaced it.

The half that is still accepted is the more insidious one. An operator who removes the `<command>` block to get the manager up is left with an `<active-response>` block that parses, looks configured, and never fires.

On the agent the failure mode is worse in practice, because nothing about it is visible after the fact. A configuration change that cannot be loaded leaves the endpoint running its previous configuration indefinitely, and every health check agrees it is fine. On a fleet where restarts are scripted, a bad element pushed to many agents produces no alarm anywhere: the agents keep reporting, keep syncing, and keep monitoring whatever they were monitoring before.

## Additional context: where Active Response lives in 5.0.0

Included because it is what makes the above a migration problem rather than a missing feature.

- The Server API has 72 paths and none of them is `/active-response`. `PUT /active-response` answers 404. The only agent-directed actions are restart, reload, upgrade, upgrade_custom and uninstall.
- None of the 290 shipped rules declares a response action, and the rule schema has no field for one. Its keys are `detection, enabled, falsepositives, id, level, logsource, metadata, status, tags, mitre, compliance, sigma_id, fields, related`.
- The trigger path is on the indexer side. The `wazuh_manager` indexer role holds `read` on `wazuh-active-responses*`, the `.ds-wazuh-active-responses-000001` data stream exists, and 21 detection workflows ship enabled (`wazuh-fim`, `wazuh-rootcheck`, `wazuh-manager`, `suricata`, `github`, `windows`, `azure`, `wazuh-it-hygiene` and others).

An Active Response was not driven end to end through that path, so this issue does not claim the capability is broken — only that the leftover configuration surface is.

TomasTurina:
## Description

Closes #38563

`read_main_elements()` has exactly two outcomes for a top-level configuration element: a parser runs, or it falls through to `merror(XML_INVELEM)` and aborts. There is no third state for "known element, removed in 5.0.0". Every 4.x element that lost its parser is therefore fatal, and a configuration file carried over from 4.x stops the daemon from starting.

Exactly one element got that third state, hand-coded: `client_buffer`, added in 52b9bf7af3 with the reasoning that an agent upgrade never rewrites `ossec.conf`, so the block must not be fatal. That reasoning applies to ten other elements that never received the same treatment.

The Active Response surface shows the resulting asymmetry. `<command>` — how a response was *defined* in 4.x — is rejected and the manager refuses to start. `<active-response>` — how a response was *referenced* — reaches an empty branch and is discarded with no message at all, on the manager where nothing reads it. An operator who removes the `<command>` block to get the manager up is left with an `<active-response>` block that parses, looks configured, and never fires.

On the agent the same class of bug is worse in practice. `wazuh-control restart` runs `testconfig` before stopping anything, so a rejected element aborts the restart and leaves the previous daemons running with the previous configuration, while `wazuh-control status` keeps reporting them as healthy. That is how a `<labels>` block carried over from 4.x turns into an agent that silently ignores every subsequent configuration change.

**Scope.** This PR addresses the parser: the elements the issue reports are no longer rejected, so they no longer abort an agent restart, and the two halves of the Active Response configuration stop behaving differently. It does **not** change what `wazuh-control status` reports. An agent whose configuration is invalid for some other reason will still abort its restart and still be reported as healthy; that half of the issue remains open.

## Proposed Changes

**Obsolete elements are ignored with a warning instead of being fatal** (`src/config/src/config.c`)

- Added a table of top-level elements that 4.x accepted and 5.0.0 no longer parses, checked before anything else in the element loop so a removed element can never reach the fatal branch: `active-response` (manager only), `agent-key-polling`, `agentless`, `alerts`, `client_buffer`, `command`, `database_output`, `email_alerts`, `fluent-forward`, `integration`, `labels`, `reports`, `rule_test`, `ruleset`, `syslog_output`.
- The list is the complete set difference between v4.14.7's top-level elements and the ones 5.0.0 still parses, computed from both versions of `read_main_elements` rather than assembled by hand: 39 elements in 4.x, 26 still parsed, 15 in the table (`active-response` appears in both, since it stays valid on the agent). Nothing removed in 5.0.0 is left out, and nothing in the table is an element 4.x never had.
- Each entry carries its own explanation — the daemon that was removed, or where the setting went — so message 1230's bare `Invalid element in the configuration: 'command'` is replaced by something an operator migrating from 4.x can act on.
- Elements not in the table are still rejected exactly as before, so typos keep being caught.
- Each entry has a scope. `<active-response>` is live configuration on an agent — `execd` reads it straight from the file, see `os_execd/src/config.c` — and the manager parses `agent.conf` on the agent's behalf, so it is obsolete only when the manager reads its own file. This makes the two halves of the 4.x Active Response configuration behave identically and closes the silent-discard case.
- Replaced the hand-coded `client_buffer` branch with its table entry (`minfo` becomes `mwarn`: a block that parses and silently stops working warrants a warning).

**New message** (`src/shared/include/error_messages/error_messages.h`)

- `(1223): '%s' is no longer supported and will be ignored. %s`

### Results and Evidence

The table logic was extracted and compiled standalone under `-Wall -Wextra`, both with and without `-DCLIENT`, to confirm the scope rule resolves correctly on each build:

```
                        manager build                              agent build
command             ->  Active Response commands are not ...   |   Active Response commands are not ...
labels              ->  Agent labels were removed in 5.0.0.    |   Agent labels were removed in 5.0.0.
ar (manager conf)   ->  Active Response is not configured ...  |   (null: still valid)
ar (agent.conf)     ->  (null: still valid)                    |   (null: still valid)
global              ->  (null: still valid)                    |   (null: still valid)
typo 'commnd'       ->  (null: still fatal)                    |   (null: still fatal)
```

Table completeness was checked by extracting the top-level element list from `read_main_elements` in both `v4.14.7` and this branch and diffing them against the table. Every element 4.x accepted is either still parsed or covered, and every table entry is an element 4.x actually had. Three entries — `rule_test`, `fluent-forward` and `agent-key-polling` — were added as a result of that audit; they had been missed when the list was first assembled by reading the diff.

The unit tests introduced below have been through two CI rounds, each of which found a real problem in the test harness rather than in the parser:

- The first run failed at CMake configure time with a duplicate target name — `unit_tests/syscheckd` already builds a `test_config`, and CMake target names are global. The test was renamed to `test_config-elements`.
- The second run failed all six cases on the agent target with `Invalid element in the configuration: 'wazuh_config'`. The unit-test build never defines `CLIENT`, while the `config` library it links is built with it for agent targets, so `WAZUHCONFIG` in the test expanded to the manager's root element inside a binary linked against an agent library — and the same bad inference selected the manager test set. The test now takes the target from a `TEST_AGENT_TARGET` definition supplied by `CMakeLists.txt`, which is the only place that knows it.

A green run has not been observed yet. Both fixes were reasoned through against the library's compile-time configuration for every target, but the evidence for them is the two failures above, not a passing run.

### Artifacts Affected

- `config.c` is part of the shared configuration library, so the change reaches every daemon that reads a configuration file, on the manager and on the agent, across Linux, macOS and Windows.
- No default configuration file was modified. The shipped manager configuration contains no `<command>` or `<active-response>` block, and the shipped agent configuration keeps its `<active-response>` block, which stays valid.

### Configuration Changes

- No new configuration parameters, and no changed defaults.
- Behaviour change: configuration elements listed in the table no longer prevent a daemon from starting. They now produce a `(1223)` warning and are ignored. This is what makes an in-place agent upgrade from 4.x survive an `ossec.conf` that still carries `<labels>` or `<command>`.
- Behaviour change: `<active-response>` in the manager's own configuration file now warns instead of being silently discarded. It is unchanged everywhere it is still valid — an agent's `ossec.conf`, and an `agent.conf` the manager parses on the agent's behalf.
- No backward-compatibility concern in the other direction: a 5.0.0 configuration file contains none of these elements, so nothing that parses cleanly today starts warning.

### Tests Introduced

`src/unit_tests/config/test_config-elements.c`, registered for both the manager and agent targets in `src/unit_tests/config/CMakeLists.txt`.

Two harness details worth knowing when reading it:

- It is named `test_config-elements`, not `test_config`, because CMake target names are global across the project and `unit_tests/syscheckd` already builds a `test_config` out of `syscheck_config.c`.
- Both the root element it writes and the `<active-response>` case it selects come from `TEST_AGENT_TARGET`, passed in by `CMakeLists.txt`. The test cannot use `CLIENT` — the unit-test build does not define it, while the library under test does for agent targets — nor anything derived from it, such as `WAZUHCONFIG`.

Cases, run against both the manager and the agent build of the library:

- An obsolete element is ignored, and warns with its own explanation.
- Several obsolete elements in one file each warn with their own text, and one does not stop the rest of the file from being read.
- An obsolete element with no children is ignored too, since it takes a different path through the reader.
- An element nobody ever supported is still fatal, with 1230 followed by 1202.

Agent build only:

- `<active-response>` is accepted without a word.

Manager build only:

- `<active-response>` warns as obsolete.
- `<command>` and `<active-response>` together warn identically, covering the asymmetry the issue reports.

## Review Checklist
- [x] Code changes reviewed
- [ ] ...

### DIFF
--- a/src/config/src/config.c
+++ b/src/config/src/config.c
@@ -14,6 +14,78 @@
 #include "os_xml.h"
 #include "config.h"
 
+#ifdef CLIENT
+#define IS_AGENT_BUILD 1
+#else
+#define IS_AGENT_BUILD 0
+#endif
+
+/* Where an element stopped being valid */
+#define OBS_ALWAYS      0   /* Everywhere it can appear */
+#define OBS_SERVER_CONF 1   /* Only in the manager's own configuration file */
+
+typedef struct obsolete_element_t {
+    const char *element;
+    int scope;
+    const char *hint;
+} obsolete_element_t;
+
+/* Top-level elements that 4.x accepted and 5.0.0 no longer has a parser for.
+ *
+ * An element with no parser reaches the end of read_main_elements() and is fatal,
+ * so a configuration file carried over from 4.x stops the daemon from starting at
+ * all. An agent upgrade is supported and never rewrites ossec.conf, and a manager
+ * still distributes agent.conf files written for 4.x, so these are accepted and
+ * ignored with a warning instead of rejected. Anything not listed here is still
+ * fatal: the point is to be lenient with settings that were removed on purpose,
+ * not with typos.
+ *
+ * <active-response> is the one element whose validity depends on who is reading
+ * it. It is live configuration on an agent -- execd reads it straight from the
+ * file, see os_execd/src/config.c -- and the manager parses agent.conf on the
+ * agent's behalf, so it is obsolete only when the manager reads its own file. */
+static const obsolete_element_t OBSOLETE_ELEMENTS[] = {
+    {"active-response", OBS_SERVER_CONF, "Active Response is not configured on the manager. This block only applies to agents."},
+    {"agent-key-polling", OBS_ALWAYS,    "The agent key polling module was removed in 5.0.0."},
+    {"agentless",       OBS_ALWAYS,      "The wazuh-agentlessd daemon was removed in 5.0.0."},
+    {"alerts",          OBS_ALWAYS,      "Alert output is not configured in the configuration file."},
+    {"client_buffer",   OBS_ALWAYS,      "Event batching is configured under <agent><batch>."},
+    {"command",         OBS_ALWAYS,      "Active Response commands are not defined in the configuration file."},
+    {"database_output", OBS_ALWAYS,      "The wazuh-dbd daemon was removed in 5.0.0."},
+    {"email_alerts",    OBS_ALWAYS,      "The wazuh-maild daemon was removed in 5.0.0."},
+    {"fluent-forward",  OBS_ALWAYS,      "The fluent forwarder module was removed in 5.0.0."},
+    {"integration",     OBS_ALWAYS,      "The wazuh-integratord daemon was removed in 5.0.0."},
+    {"labels",          OBS_ALWAYS,      "Agent labels were removed in 5.0.0."},
+    {"reports",         OBS_ALWAYS,      "The wazuh-reportd daemon was removed in 5.0.0."},
+    {"rule_test",       OBS_ALWAYS,      "Rule testing is provided by the engine."},
+    {"ruleset",         OBS_ALWAYS,      "Rules and decoders are managed by the engine."},
+    {"syslog_output",   OBS_ALWAYS,      "The wazuh-csyslogd daemon was removed in 5.0.0."},
+    {NULL,              0,               NULL}
+};
+
+/* Return the explanation for an element that is obsolete in this context, or NULL
+ * if the element is either still valid here or not one we know about. */
+static const char *get_obsolete_hint(const char *element, int modules)
+{
+    /* True only while reading the manager's own configuration: an agent build
+     * never qualifies, and neither does a manager reading a remote agent.conf. */
+    const int server_conf = !IS_AGENT_BUILD && !(modules & CAGENT_CONFIG);
+    int i;
+
+    for (i = 0; OBSOLETE_ELEMENTS[i].element != NULL; i++) {
+        if (strcmp(element, OBSOLETE_ELEMENTS[i].element) != 0) {
+            continue;
+        }
+
+        if (OBSOLETE_ELEMENTS[i].scope == OBS_SERVER_CONF && !server_conf) {
+            return NULL;
+        }
+
+        return OBSOLETE_ELEMENTS[i].hint;
+    }
+
+    return NULL;
+}
 
 /* Read the main elements of the configuration */
 static int read_main_elements(const OS_XML *xml, int modules,
@@ -28,7 +100,6 @@ static int read_main_elements(const OS_XML *xml, int modules,
     const char *oslocalfile = "localfile";                      /* Agent Config  */
     const char *osremote = "remote";                            /* Agent Config  */
     const char *osclient = "client";                            /* Agent Config  */
-    const char *osbuffer = "client_buffer";                     /* Removed in 5.0.0 (#38030) */
     const char *osagent = "agent";                              /* Agent Config (HTTPS endpoint) */
     const char *osactiveresponse = "active-response";           /* Agent Active Response Config  */
     const char *oswmodule = "wodle";                            /* Wodle - Wazuh Module  */
@@ -57,10 +128,21 @@ static int read_main_elements(const OS_XML *xml, int modules,
 
     while (node[i]) {
         XML_NODE chld_node = NULL;
+        const char *obsolete_hint = NULL;
 
         if (!node[i]->element) {
             merror(XML_ELEMNULL);
             goto fail;
+        }
+
+        /* Checked before anything else so a removed element is never mistaken for
+         * an unknown one, which is fatal further down. */
+        obsolete_hint = get_obsolete_hint(node[i]->element, modules);
+
+        if (obsolete_hint != NULL) {
+            mwarn(XML_OBSOLETE, node[i]->element, obsolete_hint);
+            i++;
+            continue;
         }
 
         chld_node = OS_GetElementsbyNode(xml, node[i]);
@@ -120,14 +202,7 @@ static int read_main_elements(const OS_XML *xml, int modules,
                 goto fail;
             }
 #endif
-        } else if (strcmp(node[i]->element, osbuffer) == 0) {
-            /* Accepted and ignored rather than rejected (an unknown element is
-             * fatal here), so an ossec.conf or a pushed agent.conf still
-             * carrying it does not stop the agent from starting. */
-            minfo("'%s' is no longer used and will be ignored. Event batching is configured "
-                  "under <agent><batch>.", node[i]->element);
-        }
-        else if (strcmp(node[i]->element, oswmodule) == 0) {
+        } else if (strcmp(node[i]->element, oswmodule) == 0) {
             if ((modules & CWMODULE) && (Read_WModule(xml, node[i], d1, d2) < 0)) {
                 goto fail;
             }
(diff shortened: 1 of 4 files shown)


jotacarma90 (approved):
LGTM

PR #38612, opened 2026-08-26
Improve Auth and Inventory-sync documentation

### LINKED ISSUE #38572: Freshly enrolled agents get 401 until remoted reloads client.keys, and their recovery re-enrollment locks them out
## Description

In the 5.0.0 deployability nightly (2026-08-25, [run 1532](https://jenkins-staging.qa.wazuh.info/job/5.0.0/job/deployability_testing/1532/), reported as wazuh/wazuh-qa-automation#8708), an agent enrolled successfully over HTTPS and was rejected with `401` when it first used the key the manager had just issued. Its recovery re-enrollment was then refused as a duplicate of itself, leaving it permanently locked out.

Agent (`agent-oracle-7-amd64`):
```
07:38:16 wazuh-agentd: INFO: Valid key received
07:38:36 wazuh-agentd: WARNING: https_client: credential rejected (401); re-enrolling.
07:38:36 wazuh-agentd: ERROR: Enrollment rejected by the manager: duplicate agent. Duplicate name
```

Manager:
```
07:38:09 wazuh-manager-remoted:keystore: INFO: Reloaded 0 agent key(s) from 'etc/client.keys'.
07:38:35 wazuh-manager-authd: INFO: Duplicate name 'agent-oracle-7-amd64', rejecting enrollment. Failed to get agent-info for agent '067'
07:38:36 wazuh-manager-remoted:keystore: INFO: Reloaded 21 agent key(s) from 'etc/client.keys'.
07:38:40 wazuh-manager-authd: INFO: Duplicate name 'agent-oracle-7-amd64', rejecting enrollment. Agent '067' doesn't comply with the registration time to be removed.
```

## Root cause

The test suite's DNS re-registration step deletes every agent and lets all ~65 re-enroll at once. remoted's HTTPS keystore learns new keys only from `client.keys` file reloads, and no reload was adopted between 07:38:09 and 07:38:36 (the keystore's own log shows `Reloaded 0` and then nothing until `Reloaded 21`). The exact reason the watcher stalled is not pinned down by the logs -- there is no torn-read give-up warning, and the legacy keystore's 10 s poll also skipped beats in the same window (`(1410): Reading authentication keys file` at :11 and then :41), pointing at generalized starvation under the re-enrollment and deletion storm. Whatever delays the reload, every credential issued inside such a window answers `401`. Each one was issued by the same remoted process that then rejected it.

The agent reacts to the 401 by discarding the credential and re-enrolling, and that recovery path is closed: authd still holds the registration, and the force options refuse the replacement (first because the seconds-old agent has no agent-info to evaluate, then because `after_registration_time` blocks replacing recently registered agents). The agent retries forever.

## Proposed fix

`POST /enroll` is served by remoted itself: `mapAuthdResult()` holds the accepted id/ip/key when it builds the 200. Upsert that entry into remoted's own keystore before answering, so the manager honors the credential the instant the agent can first use it. File reloads and upserts converge (authd persisted the entry before replying), and the reload remains the source of truth.

The authd side (force options blocking an agent's self-recovery, and the failed agent-info lookup on seconds-old agents) deserves its own follow-up.

NahuFigueroa97:
## Description

The authd documentation only really covered one thing: agent deletion. `architecture.md` gave
enrollment 13 lines and the removal path around 130, with three of its four diagrams devoted to the
purge. Everything else authd does — how a key is assigned, what the `<force>` guards actually
evaluate and in which order, what a worker node does and does not keep, how the enrollment password
reaches a worker, what each `9xxx` code means — was either absent or scattered across code comments.

That gap has a cost we hit directly while investigating a `401` in
`deployability_testing/1532`: the single most load-bearing fact about the module — that **authd's
in-memory keystore and `etc/client.keys` are two different stores with two different readers, and the
file can lag** — was documented nowhere, so every reasoning about the incident had to rediscover it.

This PR fills those gaps, adds the requirements catalog the module never had, and fixes one stale
arrow in the inventory-sync diagram.

Closes #38572

## Proposed Changes

**`docs/ref/modules/authd/architecture.md`** — from 214 to 392 lines, from 4 to 8 diagrams, with the
deletion material kept as it was and the rest of the module brought up to the same level:

- **The two stores** (new, and placed first because everything else depends on it) — the in-memory
  keystore vs `client.keys`, which component reads which, and why enrollment latency is the writer's
  pass plus remoted's reload.
- **Enrollment** — a sequence diagram covering the three doors, the critical section under
  `mutex_keys`, and the exact point at which remoted learns about the key. Plus validation and
  id/key assignment.
- **Duplicate handling and force replacement** (new) — the `w_auth_replace_agent()` guard chain as a
  flowchart, in evaluation order, with the refusal message each guard produces.
- **Cluster** (new) — a sequence diagram of a worker forwarding to the master, and the fact that the
  worker keeps nothing locally.
- **The enrollment password** (new) — `etc/authd.pass`, the authpass watcher, and why a worker fails
  closed until the file arrives.
- **Error codes** (new) — `9001`–`9018` with their `/enroll` HTTP mapping.

**`src/os_auth/README.md`** — the requirements catalog the module lacked, in the same shape
`inventory_sync_server` uses so both modules read alike:

- **RF-1 … RF-13** functional requirements, each marked `kept` or `superseded`. The superseded rows
  are kept deliberately: RF-11 ("answer the indexer purge synchronously") is the reason the relay
  thread exists, and deleting the row would delete the reason.
- **RNF-1 … RNF-8** non-functional requirements, each pointing at where it is answered.
- **REQ-PURGE-1 … 5** — the contract with `inventory_sync_server` across the deletion route, which
  until now lived implicitly in code comments on both sides.
- **D1 … D8** design decisions with their rationale.

**`docs/ref/modules/authd/README.md`** — a `Cluster` section, the authpass watcher added to the
thread table, a link to `architecture.md` from the header, and a `Development` pointer to the in-tree
README (the same pattern `inventory-sync-server/README.md` uses).

**`docs/ref/modules/inventory-sync-server/architecture.md`** — one stale arrow fixed. The overview
flowchart drew `DELETE /agents` reaching the indexer directly:

```
DELR -->|"deleteByQuery\nwazuh-states-*, wazuh-agent-config, wazuh-agent-stats"| IDX
```

`deleteAgentEndpoint.cpp` does no such thing: it calls `pipeline->enqueue()` with
`Kind::DeleteAgent` and answers `200` at admission, and the shard worker runs the `deleteByQuery`.
Routing the deletion through the agent's shard is what orders it FIFO against that agent's in-flight
sessions — the whole point of the design, which the diagram hid. Now:

```
DELR -->|"enqueue on the agent's shard\nanswers 200 at admission"| PIPE
```

The index scope that label carried is already documented in prose further down the same page, so
nothing was lost.

  - [ ] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

### DIFF
--- a/docs/ref/modules/authd/README.md
+++ b/docs/ref/modules/authd/README.md
@@ -12,7 +12,9 @@ long-term enrollment path going forward.
 
 Source: `src/os_auth/`
 
-For configuration options see [Authd Configuration](configuration.md).
+For the diagrams — the two stores, the enrollment sequence, the force-guard chain, cluster
+forwarding and the removal path — see [Authd Architecture](architecture.md). For configuration
+options see [Authd Configuration](configuration.md).
 
 ## How it works
 
@@ -46,6 +48,21 @@ For configuration options see [Authd Configuration](configuration.md).
 | Local server | Handles enrollment via the local Unix socket `queue/sockets/auth.sock` |
 | Writer | Flushes the in-memory key queue to `client.keys` on disk, deletes each removed agent from wazuh-db, and hands the indexer purge of every removed agent to the relay. It never waits on the network |
 | Purge relay | Sends the queued indexer purges to the [Inventory Sync Server](../inventory-sync-server/README.md), after the configured delay, and owns every retry |
+| authpass watcher | On a worker with `use_password`, re-reads `etc/authd.pass` as the cluster syncs it down from the master. Until it arrives the worker fails closed and rejects enrollments |
+
+The writer and the purge relay run on the **master only**. See [Cluster](#cluster) below.
+
+## Cluster
+
+A worker node does not own a keystore. An enrollment that arrives at a worker is forwarded to the
+master, which validates it, assigns the id and generates the key; the worker relays the answer to the
+agent and keeps nothing locally. The `<force>` settings are ignored on a worker — the master decides —
+and a worker that cannot reach the master answers `9016`.
+
+The key reaches the worker's own `client.keys` through the cluster's integrity sync, the same
+mechanism that distributes `etc/authd.pass`. Until it does, the worker's remoted cannot verify a key
+the agent already holds; see
+[the two stores](architecture.md#the-two-stores) and [Cluster](architecture.md#cluster).
 
 ## Storage
 
@@ -267,3 +284,12 @@ manager's own hostname and the `CN` parsed out of `-S` (when present and not alr
 
 The actual `<auth>` XML element parsing, validation, and default values live in the shared config
 subsystem at `src/config/src/authd-config.c`, not in `os_auth` itself.
+
+## Development
+
+The in-repo companion to these pages (a plain path — it lives outside this book):
+
+- `src/os_auth/README.md` — the developer's map of the module: the functional/non-functional
+  requirements catalog (RF, RNF, and the `REQ-PURGE` contract with inventory-sync), the design
+  decisions (D1–D8) with the reasoning behind each, the load-bearing invariants, the developer FAQ,
+  and which test suite covers what.
--- a/docs/ref/modules/authd/architecture.md
+++ b/docs/ref/modules/authd/architecture.md
@@ -55,19 +55,197 @@ thread instead of being called inline.
 | Local server | every node | `queue/sockets/auth.sock`: `add`, `remove`, `get` — for the server API and for remoted's `/enroll` |
 | Writer | master only | persists `client.keys`, removes Wazuh DB rows, queues indexer purges |
 | Purge relay | master only | sends queued purges after their delay, owns every retry |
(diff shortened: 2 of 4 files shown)


matigarciadev (approved):
**LGTM!**

juliancnn (approved):
LGTM!

PR #38617, opened 2026-08-26
Debian and Alpine package extraction from container image layers

### LINKED ISSUE #38585: Implement package extraction from container image layers (dpkg and apk)
## Description

Implement package extraction from container image contents, so the `container_images` module inventories the packages that scanned images really contain.

The module already discovers images and persists package rows, but the rows come from a temporary in-memory reader. This issue replaces it with the real inventory, stored in the tables that already exist. No schema change is expected.

Debian, Ubuntu and Alpine based images are covered here. RPM based images require changes in shared agent code and are handled apart.

Parent research/spike: #35609

## Scope

- Read the contents of a container image, in the layer formats used by standard images, without unpacking it on disk and without container tooling on the host.
- Report the packages present in the final state of the image.
- Extract the package inventory of Debian and Ubuntu based images.
- Extract the package inventory of Alpine based images, including the variants that keep their package database in an alternative location.
- Report only the packages actually installed in the image.
- Populate the package fields already defined by the existing tables, with no schema change.
- Recognize the package formats that are not supported yet and inventory the image with no packages and a warning.
- Accept saved image archives as an input, in addition to the image directories already supported.
- Rename the `<local>` reference entry to `<archive>`, and reserve `<local>` for the container engine store.
- Accept the three reference entry types in the configuration, implementing `<archive>` and reporting the other two as not supported yet.
- Keep the image reading independent of the image source, and the package handling independent of the package format.
- Isolate a failing reference from the rest of the scan.
- Remove the temporary in-memory reader.
- Add unit tests for the image reading, the layer precedence and each package format.
- Add integration tests over image inputs built on disk.
- Update the module documentation with the supported inputs and package formats.

## Proposed configuration

```xml
<container_images>
  <enabled>yes</enabled>
  <interval>1h</interval>
  <references>
    <archive>/var/tmp/images/myapp.tar</archive>
    <archive>/var/tmp/images/myapp-image-directory</archive>
  </references>
</container_images>
```
The reference entry types are `<ref>` for a registry reference, `<archive>` for a saved archive or an image directory, and `<local>` for the container engine store. Only `<archive>` is implemented here.

## Out of scope

- RPM package inventory.
- Package formats other than the ones used by Debian, Ubuntu and Alpine based images.
- Remote registry references.
- Container engine store and container runtime access.

## Tasks

- [ ] Implement reading the contents of an image, in the supported layer formats, without unpacking it on disk.
- [ ] Implement the layer precedence rules.
- [ ] Extract the package inventory of Debian and Ubuntu based images.
- [ ] Extract the package inventory of Alpine based images, covering both database locations.
- [ ] Report only the packages actually installed in the image.
- [ ] Map the extracted package fields onto the existing package table.
- [ ] Inventory an image with no packages and a warning when its package format is not supported yet.
- [ ] Add support for saved image archives as a reference input.
- [ ] Rename the `<local>` reference entry to `<archive>` and reserve `<local>` for the engine store.
- [ ] Accept the three reference entry types, reporting the unimplemented ones as not supported yet.
- [ ] Add the image source and package format abstractions.
- [ ] Remove the temporary in-memory reader.
- [ ] Add unit tests for the image reading, the layer precedence and each package format.
- [ ] Add integration tests over image inputs built on disk.

## Acceptance criteria

- A saved image archive and an image directory are both inventoried with the packages they contain.
- The stored package list matches the one reported by the package manager inside the equivalent running container, for a Debian or Ubuntu image and for an Alpine image.
- Packages removed or replaced in a later layer are reported according to the final state of the image.
- An Alpine based image using the alternative database location is inventoried.
- An image whose package format is not supported yet is inventoried with no packages and a warning, and the scan continues.
- A malformed or unreadable image costs one reference, not the scan.
- The image is inventoried without being unpacked on disk and without container tooling on the host.
- Every configured reference is read from its real contents, and the temporary in-memory reader is gone.
- No RPM inventory, remote registry support, or container engine store access is introduced in this issue.

anromerom:
## Description

Closes #38585

The `container_images` module persisted package rows produced by a temporary in-memory reader, so the stored inventory did not reflect the images that were actually scanned. 

This PR extracts the real package inventory from the image layers, for saved image archives and OCI image layout directories, covering the dpkg and apk package databases and reporting a recognized but unimplemented format as an empty inventory with a warning. It also renames the `<local>` reference entry to `<archive>` and reserves `<local>` for the container engine store. 

The existing tables are reused with no schema change; RPM, remote registries and engine store access remain out of scope.

## Proposed Changes

- Read saved image archives and OCI image layout directories without unpacking them on disk and without container tooling on the host, replacing the temporary in-memory reader.
- Compose the layers in manifest order, applying OverlayFS deletion markers and opaque directory markers, so the packages reported are the ones present in the final state of the image.
- Parse the dpkg database (`var/lib/dpkg/status`) and the apk database (`lib/apk/db/installed` and `usr/lib/apk/db/installed`, treated as two locations of one database) into the existing package fields, reporting only installed packages.
- Inventory an image whose package format is recognized but not implemented with zero packages and a warning, retract that warning when a later layer removes the database path, and keep the scan going over the remaining references.
- Rename the `<local>` reference entry to `<archive>`, accepting a saved archive or an image directory; `<local>` and `<ref>` are accepted by the grammar and report themselves as not implemented.
- Keep one references row per configured reference: attestation and provenance manifests (platform `unknown`/`unknown`) are filtered, and when more than one real image remains behind a reference the first is kept deterministically and the others are named in a warning.
- Bound the work a crafted image can request: a shared visited-digest set and a 256-node budget over the index graph, an 8 GiB decompressed ceiling per layer blob, at most 256 metadata-shaped members retained from a saved archive, an `is_regular_file` check before opening a layer blob, and a pending-stop check between references.
- Harden the tar reader: header checksum validation, an entry-size ceiling that prevents padding-arithmetic overflow on a crafted base-256 size, pax `size` record override, path normalization for an interleaved `/./` prefix, tar last-wins semantics for duplicated members, a bare `.wh.` entry ignored instead of wiping the composed inventory, and a truncated layer pass over a saved archive reported rather than discarded.

### Results and Evidence

- `container_images_unit_test`: 92/92 passing.
- Package lists read by the module match the package manager in the equivalent running container: `debian:12` 88 packages identical to `dpkg-query -W`; `alpine:3.19` 15 identical to `apk info -v`; `cgr.dev/chainguard/wolfi-base` 15 identical to `apk info -v`, read from `usr/lib/apk/db/installed`.
- A three-layer image that installs and then purges `nano` reports 88 packages with no `nano`, identical to `dpkg-query -W` on the final image.
- `almalinux:9` reports 0 packages with the not-implemented warning, and the scan continues.
- Real agent installation over eight images:

```sql
Scan ended. 8 references, 204 packages.
```

- `make TARGET=agent TEST=yes -j4` builds and links the `container_images` targets cleanly. The build then stops on a pre-existing `agent_info_basic_test` link failure (`undefined reference to SelectQuery::rowFilterBindText`) unrelated to this change: `wazuh_modules/agent_info` and `shared_modules/dbsync` are byte-identical to the base commit.

### Artifacts Affected

- `libcontainer_images.so`, linked into `wazuh-modulesd`. Linux, all architectures.
- `check_files` size baseline for `libcontainer_images.so` raised 163000 → 275000 in the six Linux CSVs under `.github/actions/check_files/`. Worth re-verifying against the artifact built in CI.

### Configuration Changes

- `<container_images><references>`: the `<local>` entry is renamed to `<archive>` and takes a saved image archive or an image directory. Existing configurations that used `<local>` for a path must be updated.
- `<local>` is now reserved for the container engine store and `<ref>` for a registry reference. Both are accepted by the grammar and report themselves as not implemented, so the configuration will not change again when they land.

### Documentation Updates

- `docs/ref/modules/container-images/`: `README.md`, `configuration.md`, `architecture.md`, `api-reference.md`, `persistence.md` and `synchronization.md` updated for the supported inputs, the package formats and the `<archive>` rename.
- `architecture.md` also states the attestation-manifest filter and the one-reference-per-row behaviour.

### Tests Introduced

- Unit: image reading over tar variants (ustar prefix, pax, GNU long names), gzip and plain layers, truncation, deletion markers, opaque directories and layer composition order.
- Unit: dpkg and apk field mapping with installed-state filtering, and both apk database locations.
- Unit: OCI layout directory and saved archive end to end, multi-manifest inputs, attestation manifests, index fan-out bounding and path traversal rejection.
- Unit: tar hardening cases (header checksum rejection, pax `size` override, interleaved `/./` normalization, bare `.wh.`, unsupported-format retraction, apk location de-duplication).
- Integration: six scenarios over OCI layouts and saved archives built on disk (scan and store, layer override, whiteout, saved archive, image update, unsupported format), with the image inputs written under a per-run temporary directory rather than a fixed path.

## Review Checklist

- [ ] Code changes reviewed


src/wazuh_modules/container_images/container_images_impl/src/archive_image_reader.cpp
```diff
@@ -0,0 +1,969 @@
...
+
+            FileByteStream file {path / key};
+
+            if (!file.isOpen())
+            {
+                logWarn("Could not open the layer '" + key + "' of '" + m_location + "'.");
+                continue;
+            }
+
+            GzipByteStream gzipLayer {file};
```
rjcausarano: GzipByteStream seems to support only gzip and standard tar, what will happen if file is zstd? What if it is any other media type?

cborla (approved):
LGTM

PR #38619, opened 2026-08-26
API endpoint to trigger on-demand Vulnerability scans.

### LINKED ISSUE #38553: Add an API endpoint to trigger an on-demand Vulnerability Detection scan
## Summary

`remoted` already exposes `POST /scan/vd`, which allows a **connected agent** to request an on-demand vulnerability scan when it detects, through `/control`'s `vd_feed_offset`, that the Vulnerability Detection (VD) feed has advanced beyond the version it last synchronized.

We now need to support the **opposite direction**: allow a **user**, through `wazuh-apid`, to trigger an on-demand VD scan for:

* A single agent.
* A list of agents.
* All agents.

The initial consumer will be the Wazuh dashboard, which will expose this as a **"Scan now"** action in the agents/vulnerabilities view. However, the endpoint should be designed as a standalone API contract that can be consumed by other clients as well.

This issue covers the manager-side implementation and API contract so the **dashboard team can build against the contract in parallel**.

## Why not use `POST /scan/vd`?

`POST /scan/vd` is an agent-facing endpoint. The request is authenticated using the calling agent's own key (AES-CMAC) and is specifically designed for an agent to request a scan of **itself**.

It therefore cannot be used by `wazuh-apid` to request a scan for an arbitrary agent.

### Decision

The API layer will invoke the vulnerability scanner **directly through its existing UDS interface**, without going through `remoted`.

The `vulnerability_scanner` module already exposes an endpoint specifically for admitting an on-demand scan for an agent:

```text
POST /vulnerability-detector/scan
Socket: queue/sockets/vd.sock
Body: {"agent_id": "<id>"}
```

See `src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScanner.cpp:101-210`.

The existing endpoint already provides the required admission semantics:

* The request body is limited to 4 KiB and contains a single `agent_id`.
* `checkScanReadiness()` is executed before the request is admitted to the dispatch lane.
* `200` is returned only when the scan has actually been queued.
* Requests are not silently retried after a successful response.
* Requests are rejected with `503` and a specific error when the scan cannot be admitted:

  * `vd_not_initialized`
  * `feed_not_ready`
  * `scanner_not_ready`
  * `indexer_unavailable`
  * `queue_full`

The API layer must preserve these semantics: **a successful response means that the scan was queued, while a failure must expose the actual reason why the request could not be queued.**

### Agent connection status

Agent connection status is irrelevant to this operation.

A VD scan is performed using the package/CVE inventory already synchronized to the indexer. It does not require an active agent session.

Therefore, both connected and disconnected agents can be scanned using the same API. No additional "queue until reconnect" mechanism is required. 

## API contract

The endpoint should follow the existing convention for bulk agent actions, such as `PUT /agents/restart`, and remain under the `/agents` resource rather than introducing a new top-level vulnerability resource.

This also leaves room for additional on-demand agent scan types in the future.

```text
PUT /agents/scan/vulnerability?agents_list=001,002,003
PUT /agents/scan/vulnerability?agents_list=*
```

`*` is the default and represents all agents.

### Parameters

| Parameter           | Type   | Default | Description                                     |
| ------------------- | ------ | ------- | ----------------------------------------------- |
| `agents_list`       | string | `*`     | Comma-separated agent IDs or `*` for all agents |
| `pretty`            | bool   | `false` | Standard API parameter                          |
| `wait_for_complete` | bool   | `false` | Standard API parameter                          |

The agent selection and RBAC filtering should follow the existing `restart_agents` pattern.

## RBAC

Add the following action:

```text
agent:scan_vulnerability
```

The action must use the existing `agent:id` resource, following the same resource model used by `restart_agents`:

```python
@expose_resources(
    actions=["agent:restart"],
    resources=["agent:id:{agent_list}"],
    ...
)
```

No new RBAC resource type is required.

This allows the new action to reuse the existing per-agent and per-group resource model. Administrators only need to grant the new action; no new resource model is required.

## Response

The endpoint should use the standard `AffectedItemsWazuhResult` format.

Each requested agent must produce either an entry in `affected_items` or an entry in `failed_items`.

Example:

```json
{
  "data": {
    "affected_items": ["001", "003"],
    "failed_items": [
      {
        "id": ["002"],
        "error": {
          "code": 1701,
          "message": "Agent does not exist"
        }
      },
      {
        "id": ["004"],
        "error": {
          "code": 5000,
          "message": "indexer_unavailable"
        }
      }
    ],
    "total_affected_items": 2,
    "total_failed_items": 2
  },
  "message": "Scan was requested for some agents",
  "error": 0
}
```

### Error handling

VD rejection reasons must be mapped to distinct, documented API error codes and messages.

The following VD errors must be preserved:

* `feed_not_ready`
* `scanner_not_ready`
* `indexer_unavailable`
* `vd_not_initialized`
* `queue_full`

The API must not hide these failures through automatic retries.

The intended semantics are:

* **Accepted:** the VD module has queued the scan.
* **Failed:** the VD module rejected the scan and the API reports the reason.
* **Partial success:** accepted and rejected agents are reported independently.

This follows the same "no silent retry" principle already used by `POST /scan/vd`. 

## Clustering

The request must always be executed by the **master node**.

`vulnerability_scanner` runs independently on each manager node, while the existing disconnected-agent rescan sweep is already master-only.

With the HTTPS architecture, agent connections are stateless and load-balanced, so there is no meaningful cluster node that exclusively owns an agent session. The target agent's current connection node therefore must not determine where the scan request is executed.

`DistributedAPI` should use:

```text
request_type='local_master'
broadcasting=False
```

This differs from actions such as `restart_agents`, where broadcasting is required because the operation depends on a live per-node agent connection.

The scan should therefore always target the master's `vulnerability_scanner` instance, regardless of whether the target agent is connected and regardless of which node handled its latest connection. 

## Fleet-wide requests

No manager-side batching, throttling, backoff, or automatic retry will be implemented.

The framework should follow the same general pattern as other bulk agent actions:

1. Resolve `agents_list` against the caller's RBAC permissions.
2. Iterate over the resulting agents.
3. Send one VD scan request per agent.
4. Map each individual result directly to `affected_items` or `failed_items`.

The VD dispatch lane is bounded to 64 slots by default. Therefore, a large request such as:

```text
PUT /agents/scan/vulnerability?agents_list=*
```

may legitimately result in a combination of accepted scans and `queue_full` failures.

Those failures should be returned to the caller as-is. The caller can explicitly retry failed agents if required.

This preserves the existing semantics of `POST /vulnerability-detector/scan` rather than introducing manager-side retry behavior. 

## Manager-side tasks

* [ ] Add a UDS client in the framework, or a thin wrapper around an existing local-socket client, capable of calling `POST /vulnerability-detector/scan` through `queue/sockets/vd.sock`.
* [ ] Implement the framework function, e.g. `framework/wazuh/vulnerability.py::scan_agents`, that:

  * Resolves `agents_list`, including `*`.
  * Applies the caller's RBAC filters.
  * Uses the existing agent-query mechanisms, following the `restart_agents` pattern.
  * Sends one request to VD per agent.
  * Maps each result directly to `affected_items` or `failed_items`.
  * Does not batch, throttle, or retry requests.
* [ ] Add the `agent:scan_vulnerability` RBAC action using the existing `agent:id` resource.
* [ ] Add the API controller entry and `DistributedAPI` wiring using:

  * `request_type='local_master'`
  * `broadcasting=False`
  * The appropriate `rbac_permissions`.
* [ ] Add the OpenAPI definition to `api/api/spec/spec.yaml`, including:

  * Endpoint.
  * Parameters.
  * RBAC action.
  * Request/response schemas.
  * Error codes.
* [ ] Write a short design note documenting the request flow and the decisions described above.
* [ ] Add unit tests covering:

  * Framework logic.
  * Agent resolution.
  * RBAC filtering.
  * Successful VD admission.
  * VD rejection reasons.
  * Partial failures.
* [ ] Add API/integration tests covering:

  * A single agent.
  * Multiple agents.
  * `agents_list=*`.
  * Nonexistent agents.
  * Connected agents.
  * Disconnected agents.
  * VD unavailable/not ready.
  * `queue_full`.
  * Partial success.
* [ ] Add a `CHANGELOG` entry.

## Dashboard-facing deliverables

The dashboard team can build the "Scan now" action against the following contract:

* **Method:** `PUT`
* **Path:** `/agents/scan/vulnerability`
* **Agent selection:** `agents_list=<id>[,<id>...]` or `*`
* **RBAC action:** `agent:scan_vulnerability`
* **Response:** `AffectedItemsWazuhResult`
* **Partial failures:** represented through `failed_items`
* **Per-agent errors:** include a documented error code and message

The OpenAPI specification will be the authoritative definition of the final API contract.

The dashboard should use `affected_items` and `failed_items` to represent partial-success states, for example when some agents are successfully queued while others are rejected because VD is not ready or its dispatch queue is full.

## Definition of Done

* [ ] `PUT /agents/scan/vulnerability` is implemented, RBAC-protected, and documented in the OpenAPI specification.
* [ ] A scan can be requested for a single agent, a list of agents, and all agents using `agents_list=*`.
* [ ] Each accepted request results in a real VD scan being admitted by `vulnerability_scanner`.
* [ ] The operation works identically for connected and disconnected agents.
* [ ] `failed_items` faithfully reflects VD's rejection reasons.
* [ ] No manager-side batching, throttling, or automatic retry is performed.
* [ ] Requests are always executed against the master's `vulnerability_scanner` using `request_type='local_master'` with no broadcasting.
* [ ] Unit and integration/API tests cover success, failure, partial failure, RBAC, and edge cases.
* [ ] `CHANGELOG` is updated.
* [ ] The OpenAPI contract is frozen and confirmed sufficient for the dashboard team to implement the "Scan now" action independently.

MiguelazoDS:
# Description

`remoted` already exposes `POST /scan/vd`, letting a *connected* agent request an on-demand Vulnerability Detection (VD) scan of itself, authenticated with its own key. There was no operator-facing path: an admin (or the dashboard) couldn't trigger a scan for an arbitrary agent, and a disconnected agent couldn't be scanned at all, even though VD scans only need the package/CVE inventory already synced to the indexer, not a live agent session.

Closes #38553

## Proposed Changes

- Added `PUT /agents/scan/vulnerability`, following the existing `PUT /agents/restart` bulk-action convention, to request an on-demand VD scan for one agent, a list, or all agents — runs on the master only (no broadcast), since a VD scan reads indexer-synced state rather than a live per-node agent session.
- Added `wazuh.vulnerability_scan.scan_agents()`, which RBAC-filters the requested agents and calls VD's existing admission endpoint (`POST /vulnerability-detector/scan` over `queue/sockets/vd.sock`) once per agent, reporting per-agent failures instead of aborting the whole batch.
- No connection-status or agent-version gating: connected, disconnected, and legacy agents are all forwarded to VD, which already skips agents with no synced package inventory on its own.
- Added a new `agent:scan_vulnerability` RBAC action, matching the shape of `agent:restart`/`agent:reload`/`agent:upgrade`.
- Mapped VD's existing `503` admission-rejection reasons to new dedicated error codes (`8001`-`8006`) so a caller sees the actual reason a scan couldn't be queued instead of a generic failure.

### Results and Evidence

Commands executed and outputs captured for every manual QA scenario on
`PUT /agents/scan/vulnerability`, run live against a real manager, indexer,
and agent fleet.

#### Setup — dummy fleet

```bash
TOKEN=$(curl -sk -u wazuh:wazuh -X POST "https://localhost:55000/security/user/authenticate" | jq -r .data.token)
REAL_ID=006   # the one real, already-auto-scanned agent

cd .../manual-qa   # register_dummy_fleet.sh / seed_dummy_agent.sh live here
TOKEN=$TOKEN ./register_dummy_fleet.sh 50 dummy > /tmp/dummy_fleet.txt
while read -r name id; do
  ./seed_dummy_agent.sh "$REAL_ID" "$id"
done < /tmp/dummy_fleet.txt
```

50/50 dummy agents registered via `POST /agents/insert/quick`, no failures.
IDs `007`-`056` (name `dummy01`...`dummy50`). Seed validation on dummy 007
(from real 006): `wazuh-states-inventory-packages` 216 docs copied, bulk
errors=false; `wazuh-states-inventory-system` 1 doc copied, bulk errors=false;
`wazuh-states-inventory-hotfixes` no source docs (expected — hotfixes are
Windows-only inventory; real agent 006 is CentOS). Full 50-agent fleet
seeded: 0 bulk errors across all 50.

##### A1. Happy path — single real agent — **PASS**

```bash
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT \
  "https://localhost:55000/agents/scan/vulnerability?agents_list=$REAL_ID" | jq .
```
Response: HTTP 200, `data.affected_items: ["006"]`, `total_failed_items: 0`,
message "Scan was requested for all agents".

Log evidence (`wazuh-manager.log`, immediately after the call):
```
scanOrchestrator.hpp:198 at runScanAfterFeedUpdate(): INFO: Vulnerability scan completed: agent='006' type=full reason=feed_update
vulnerabilityScannerFacade.cpp:651 at triggerAgentScan(): DEBUG: Feed_update scan completed for agent '006'
scanDispatcher.hpp:246 at logOutcome(): DEBUG: VD scan succeeded for agent 006
```
Matches the expected log line exactly.

##### A2. Multi-agent list — **PASS**

```bash
IDS=$(awk 'NR<=3{print $2}' /tmp/dummy_fleet.txt | paste -sd,)
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT \
  "https://localhost:55000/agents/scan/vulnerability?agents_list=$IDS" | jq .
```
HTTP 200, all 3 in `affected_items`, `total_failed_items: 0`. Log: one
`Feed_update scan completed for agent 'X'` line per id (007, 008, 009), all
at 15:27:24.

##### A3. `agents_list=*` (all agents) — **PASS**

```bash
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT \
  "https://localhost:55000/agents/scan/vulnerability" -o /tmp/scan_all.json -w "%{http_code}\n"
jq '.data.total_affected_items' /tmp/scan_all.json
```
HTTP 200, `total_affected_items: 51` (1 real + 50 dummy), `total_failed_items: 0`.
Response time **0.068s** — confirms the endpoint admits/enqueues and returns
immediately rather than waiting for each agent's scan to complete. Relevant
for C2 (sustained load) — the "sequential per-agent round trip" concern in the
guide is about the *admission* call, not the actual scan execution time, and
admission alone is very cheap.

##### A4. Empty package inventory — **PASS**

```bash
TOKEN=$TOKEN ./register_dummy_fleet.sh 1 emptydummy > /tmp/empty_dummy.txt
EMPTY_ID=$(awk '{print $2}' /tmp/empty_dummy.txt)
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT \
  "https://localhost:55000/agents/scan/vulnerability?agents_list=$EMPTY_ID" | jq .
grep "$EMPTY_ID" /var/wazuh-manager/logs/wazuh-manager.log | grep -i "skip\|inventory"
grep "Feed_update scan completed for agent '$EMPTY_ID'" /var/wazuh-manager/logs/wazuh-manager.log
```
Registered `emptydummy1` (id `057`, never_connected, no inventory seeded).
HTTP 200, `057` in `affected_items` (call succeeds, VD accepts and silently
no-ops). Log: `scanAgentList.hpp:455 at scanAgent(): INFO: Skipping feed
update scan for agent 057: no package inventory available (first scan not
completed yet, VDFirst will cover the updated feed).` No `Feed_update scan
completed for agent '057'` line found anywhere after — confirms the silent
no-op. (Note: actual wording differs slightly from the guide's predicted
text but is the correct/equivalent log line.)

(Session note: switched from paste-back to direct execution here — shell
access to this same host, `noble`, was confirmed available.)

##### A5. Nonexistent agent — **PASS**

```bash
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT \
  "https://localhost:55000/agents/scan/vulnerability?agents_list=999999" | jq .
```
`total_affected_items: 0`, `failed_items:
[{"error":{"code":1701,"message":"Agent does not exist"}, "id":["999999"]}]`.
Exact match to expected.

##### A6. RBAC denial surfaces in `failed_items` — **PASS** (with a corrected setup — 2 real findings along the way)

**Finding 1 (test-setup gap, not a PR bug):** the literal recipe (create
only a `deny agent:scan_vulnerability on agent:id:$DENY_ID` policy, link it
to a role/user, done) does **not** work as written on a real production
install. This install's global RBAC mode is `white`
(`GET /security/config` → `"rbac_mode": "white"`), not `black` as the
scenario implicitly assumed. In white mode a user has **zero** permissions
for anything not explicitly allowed — a deny-only policy with no
counterpart allow leaves the user denied for *every* agent, not just the
targeted one. First attempt (deny-only) → the whole request 403'd
(`error: 4000` at the top level, no `failed_items` split at all) for
*both* agents, including the one meant to be allowed.

**Finding 2 (general RBAC behavior, not a PR bug):** fixed by adding a second
policy (`agent:scan_vulnerability` allow on `agent:id:*`) to the same role —
but the **order** policies are linked in matters: linking allow-then-deny
still left both ids allowed (the later-linked policy's effect wins on
overlapping resources — confirmed by reading
`framework/wazuh/rbac/decorators.py`'s `_single_processor`/`_process_effect`,
which apply policies in the merged dict's iteration order with no
independent "deny always wins" precedence). Re-linking so deny was added
*last* (`policies: [101, 100]`) produced the correct split.

```bash
DENY_ID=$(awk 'NR==4{print $2}' /tmp/dummy_fleet.txt); ALLOW_ID=$(awk 'NR==5{print $2}' /tmp/dummy_fleet.txt)

DENY_POLICY_ID=$(curl -sk -H "Authorization: Bearer $TOKEN" -X POST "https://localhost:55000/security/policies" \
  -d "{\"name\":\"qa_deny_scan\",\"policy\":{\"actions\":[\"agent:scan_vulnerability\"],\"resources\":[\"agent:id:$DENY_ID\"],\"effect\":\"deny\"}}" \
  | jq -r '.data.affected_items[0].id')
ALLOW_POLICY_ID=$(curl -sk -H "Authorization: Bearer $TOKEN" -X POST "https://localhost:55000/security/policies" \
  -d '{"name":"qa_allow_scan_all","policy":{"actions":["agent:scan_vulnerability"],"resources":["agent:id:*"],"effect":"allow"}}' \
  | jq -r '.data.affected_items[0].id')
ROLE_ID=$(curl -sk -H "Authorization: Bearer $TOKEN" -X POST "https://localhost:55000/security/roles" \
  -H "Content-Type: application/json" -d '{"name":"qa_deny_role"}' | jq -r '.data.affected_items[0].id')

# order matters: allow linked first, deny linked LAST (last-linked wins on overlap)
curl -sk -H "Authorization: Bearer $TOKEN" -X POST "https://localhost:55000/security/roles/$ROLE_ID/policies?policy_ids=$ALLOW_POLICY_ID"
curl -sk -H "Authorization: Bearer $TOKEN" -X POST "https://localhost:55000/security/roles/$ROLE_ID/policies?policy_ids=$DENY_POLICY_ID"

USER_ID=$(curl -sk -H "Authorization: Bearer $TOKEN" -X POST "https://localhost:55000/security/users" \
  -H "Content-Type: application/json" -d '{"username":"qa_deny_user","password":"QaDenyPass123!"}' | jq -r '.data.affected_items[0].id')
curl -sk -H "Authorization: Bearer $TOKEN" -X POST "https://localhost:55000/security/users/$USER_ID/roles?role_ids=$ROLE_ID"

RESTRICTED_TOKEN=$(curl -sk -u qa_deny_user:QaDenyPass123! -X POST "https://localhost:55000/security/user/authenticate" | jq -r .data.token)
curl -sk -H "Authorization: Bearer $RESTRICTED_TOKEN" -X PUT \
  "https://localhost:55000/agents/scan/vulnerability?agents_list=$DENY_ID,$ALLOW_ID" | jq .

# teardown
curl -sk -H "Authorization: Bearer $TOKEN" -X DELETE "https://localhost:55000/security/users?user_ids=$USER_ID"
curl -sk -H "Authorization: Bearer $TOKEN" -X DELETE "https://localhost:55000/security/roles?role_ids=$ROLE_ID"
curl -sk -H "Authorization: Bearer $TOKEN" -X DELETE "https://localhost:55000/security/policies?policy_ids=$DENY_POLICY_ID,$ALLOW_POLICY_ID"
```

**Final correct result:** `PUT ?agents_list=010,011` (as the restricted
user) → HTTP 200 (partial), `010` in `failed_items` with `error.code: 4000`
("Permission denied: Resource type: agent:id"), `011` in `affected_items`.
`message: "Scan was requested for some agents"`. **This confirms the
endpoint's own core RBAC claim is correct** — `broadcasting=False` does
surface per-agent denials rather than silently dropping them, once the
RBAC setup itself is correct. Both findings above are pre-existing,
general RBAC-system properties (mode defaults, policy-order-wins-on-overlap)
unrelated to this PR's code — worth a note for future RBAC test setups and
possibly the general RBAC docs, not a defect to fix in #38553.

##### B1. 8001 `vd_not_initialized` — **PASS**

```bash
sudo sed -i 's#<enabled>yes</enabled>#<enabled>no</enabled>#' /var/wazuh-manager/etc/wazuh-manager.conf
sudo /var/wazuh-manager/bin/wazuh-manager-control restart
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$REAL_ID" | jq .
sudo sed -i 's#<enabled>no</enabled>#<enabled>yes</enabled>#' /var/wazuh-manager/etc/wazuh-manager.conf
sudo /var/wazuh-manager/bin/wazuh-manager-control restart
```
Disabled VD, restarted manager. `PUT` → `error.code: 8001`, message
"Vulnerability detection module has not been initialized yet:
vd_not_initialized". Reverted immediately after.

##### B2. 8002 `feed_not_ready` — **PASS** (guide's simple-restart approach didn't work; full feed wipe did)

**Finding:** a plain `systemctl restart wazuh-manager` does **not** reproduce
this — the VD feed persists to disk (`/var/wazuh-manager/queue/vd/feed`,
RocksDB-backed, 6.1GB observed) and reloads from the local cache almost
instantly on startup (`vulnerabilityScannerFacade.cpp:451`: "Existing CVE
feed detected — per-agent scans enabled", followed within the same second by
"CVE feed fully loaded"). A scan fired 8s after a normal restart already
succeeded. This window only exists before the very first successful feed
load ever, not after every restart.

```bash
sudo /var/wazuh-manager/bin/wazuh-manager-control stop
sudo rm -rf /var/wazuh-manager/queue/vd/feed /var/wazuh-manager/queue/vd/vd_updater
sudo /var/wazuh-manager/bin/wazuh-manager-control start

while true; do
  resp=$(curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$REAL_ID")
  err=$(echo "$resp" | jq -r '.data.failed_items[0].error.code // "ok"')
  echo "$err"
  [ "$err" = "ok" ] && break
  sleep 5
done
```
**Reproduced properly** by stopping the manager, deleting `queue/vd/feed`
and `queue/vd/vd_updater` (forces a genuine re-download from the CTI
server), and restarting: `error.code: 8002`, message "Vulnerability
detection feed is not ready yet: feed_not_ready" — reproduced consistently
across 53 consecutive polls (every 3-10s). **Real elapsed window: ~4m34s**
(restart 15:50:48 → feed ready 15:55:22, local time, log line "CVE feed
fully loaded — per-agent scans unblocked"). After that, the same request
returned HTTP 200 success — confirms recovery.

##### B3. 8003 `scanner_not_ready` — **not live-reproduced (code-path confirmed only, as the guide anticipated)**

```bash
sudo /var/wazuh-manager/bin/wazuh-manager-control restart &
for i in $(seq 1 20); do
  curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$REAL_ID" \
    -o /tmp/b3_$i.json -w "%{http_code}\n"
  sleep 0.5
done
grep -l '"code": 8003' /tmp/b3_*.json 2>/dev/null
```
20 requests at 0.5s intervals across a manager restart: 11 hit connection-refused
(apid not listening yet), 9 landed after full startup and got `200`. No
request landed in the narrow scanner-construction window. Per the guide's own
instruction for this scenario, marking "code-path confirmed via source review
only, not live-reproduced" rather than a failure.

##### B4. 8004 `indexer_unavailable` — **PASS**

```bash
sudo systemctl stop wazuh-indexer
sleep 65
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$REAL_ID" | jq .
sudo systemctl start wazuh-indexer
sleep 65
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$REAL_ID" | jq .
```
Stopped `wazuh-indexer`, waited 65s. `PUT` → `error.code: 8004`, message
"Indexer is not available for vulnerability scanning: indexer_unavailable".
Restarted indexer, waited for green health + 65s poll cadence, retried →
HTTP 200 success. Recovery confirmed.

##### B5. 8005 `scan_queue_full` — **PASS**

```bash
echo "vulnerability-detection.scan_dispatch_queue_slots=1" | sudo tee -a /var/wazuh-manager/etc/wazuh-manager-internal-options.conf
sudo /var/wazuh-manager/bin/wazuh-manager-control restart

IDS=($(awk 'NR>=6 && NR<=15{print $2}' /tmp/dummy_fleet.txt))
for id in "${IDS[@]}"; do
  curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$id" \
    -o "/tmp/b5_${id}.json" -w "id=$id http=%{http_code}\n" &
done
wait
grep -l '"code": 8005' /tmp/b5_*.json

sudo sed -i '/scan_dispatch_queue_slots/d' /var/wazuh-manager/etc/wazuh-manager-internal-options.conf
sudo /var/wazuh-manager/bin/wazuh-manager-control restart
```
Set `scan_dispatch_queue_slots=1`, fired 10 distinct dummy agent ids (012-021)
concurrently. Result: 7 admitted (200), 3 rejected with `error.code: 8005`
(agents 015, 017, 019). Confirms the queue-full rejection triggers under a
timing race with a healthy indexer, as corrected in the guide (not via
stopping the indexer, which short-circuits to 8004 instead). Restored default
queue slots afterward.

##### B6. 8006 `shutting_down` — **not live-reproduced (timing-dependent, as the guide anticipated)**

```bash
IDS=($(awk 'NR>=16 && NR<=20{print $2}' /tmp/dummy_fleet.txt))
MODULESD_PID=$(pgrep -f "wazuh-manager-modulesd$" | head -1)
sudo kill -TERM "$MODULESD_PID"
for id in "${IDS[@]}"; do
  curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$id" \
    -o "/tmp/b6_${id}.json" -w "%{http_code}\n" &
done
wait
grep -l '"code": 8006' /tmp/b6_*.json
sudo /var/wazuh-manager/bin/wazuh-manager-control restart
```
`kill -TERM` on modulesd, then fired 5 concurrent scan requests immediately.
Both this attempt and a subsequent full `systemctl restart` showed admission
completing faster than the shutdown sequence: log confirms
`doStopAcceptingLocked(): ... 0 pre-handler connection(s) answered 503` both
times — no request ever landed in the actual `stopAccepting()` window at
these intervals. All fired requests got HTTP 200 (admitted before shutdown
took effect). A second, more sustained-burst attempt hit tooling/timeout
issues (no conclusive data) but the manager self-recovered cleanly
afterward (all 8 daemons up, endpoint verified still working). Marking
"timing-dependent, not reproduced this run" per the guide's own allowance,
rather than asserting a pass or fail.

##### B7/B8. 8007 (unrecognized reason) / 2024 (malformed body) — **automated-only, as documented**

No live repro attempted, per the guide — neither has a realistic manual
trigger. Covered by `framework/wazuh/core/tests/test_engine_http.py`'s
existing automated coverage.

##### B9. Transport errors — 2013 / 2025 / 2026 — **PASS with one correction**

**2026 did not reproduce as predicted — real finding.** Killing modulesd
outright (`kill -9`, confirmed process gone) does **not** produce
`error.code: 2026` from this endpoint's own transport-error handling.
Instead it hits a coarser, pre-existing daemon-health precheck first:
`{"title": "Wazuh Internal Error", "detail": "Some Wazuh daemons are not
ready yet in node \"node01\" (wazuh-manager-modulesd->failed)", "error":
1017}`. This 1017 gate runs ahead of `scan_agents()`/`VdHTTPClient` entirely
and applies generically whenever any daemon is confirmed dead, not specific
to this endpoint. `2026` likely requires a narrower state (daemon alive,
only the specific VD socket refusing) that a full process kill doesn't
produce. Worth a correction note in the guide.

```bash
MPID=$(pgrep -f "wazuh-manager-modulesd$" | head -1); sudo kill -9 "$MPID"
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$REAL_ID" | jq .
sudo /var/wazuh-manager/bin/wazuh-manager-control restart

# 2025 (SIGSTOP) reproduces cleanly:
MPID=$(pgrep -f "wazuh-manager-modulesd$" | head -1); sudo kill -STOP "$MPID"
time curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$REAL_ID" | jq . &
sleep 12; sudo kill -CONT "$MPID"; wait
```
**2025 (timeout) reproduced cleanly**, as predicted "best-effort": froze
modulesd with `SIGSTOP` (process alive but unresponsive), fired the scan
request, waited 12s, then `SIGCONT`. Result: `error.code: 2025`, "Modulesd
request timeout: timed out", actual elapsed **10.07s** (matches the ~10s
client timeout). Confirmed recovery afterward (normal 200).

**2013** — not attempted live, per the guide (no clean manual trigger
distinct from the above two). Marked automated-only.

##### C1. Burst concurrency — **PASS**

```bash
IDS=($(awk '{print $2}' /tmp/dummy_fleet.txt))
for id in "${IDS[@]}"; do
  curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$id" \
    -o "/tmp/c1_${id}.json" -w "%{http_code}\n" &
done
wait
sudo /var/wazuh-manager/bin/wazuh-manager-control status
```
Default queue slots (64) confirmed active. Fired all 50 dummy agents
concurrently: **50/50 admitted (200)**, no hangs, no 5xx. All 8 daemons
confirmed still running afterward.

##### C2. Sustained load / scaling — **PASS (numbers recorded)**

```bash
for n in 10 50; do
  IDS=$(awk -v n="$n" 'NR<=n{print $2}' /tmp/dummy_fleet.txt | paste -sd,)
  echo "=== $n agents ==="
  time curl -sk -H "Authorization: Bearer $TOKEN" -X PUT \
    "https://localhost:55000/agents/scan/vulnerability?agents_list=$IDS" -o /dev/null -w "%{http_code}\n"
done
```
`agents_list` size 10 → **47ms**; size 50 → **77ms**. Growth is gentle, not
the linear-per-round-trip pattern the guide flagged as a possible concern —
consistent with A3's finding that admission is async (enqueue-and-return)
rather than waiting on each agent's actual scan to finish. Not a bug; a
positive finding about how the endpoint scales.

##### C3. Resource watch — **PASS**

```bash
MPID=$(pgrep -f "wazuh-manager-modulesd$" | head -1)
ps -o rss,etimes -p "$MPID"; ls "/proc/$MPID/fd" | wc -l
# ... run C1's 50-agent burst, 3x total ...
ps -o rss,etimes -p "$MPID"; ls "/proc/$MPID/fd" | wc -l
sleep 90
ps -o rss,etimes -p "$MPID"; ls "/proc/$MPID/fd" | wc -l
```
`wazuh-manager-modulesd` RSS across 3 full 50-agent bursts + 90s idle:
50672 → 50412 → 50748 → 51104 KB (noise-level fluctuation, ~0.9% drift, no
trend). Open fd count: constant at **83** throughout. No sustained growth
observed.

##### D1a. Same agent, both requests still queued (expect coalescing) — **not reliably reproduced despite 5 genuine attempts**

```bash
echo "vulnerability-detection.scan_dispatch_queue_slots=1" | sudo tee -a /var/wazuh-manager/etc/wazuh-manager-internal-options.conf
sudo /var/wazuh-manager/bin/wazuh-manager-control restart

TARGET_ID=<a dummy id>
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$TARGET_ID" -w "%{http_code}\n" &
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$TARGET_ID" -w "%{http_code}\n" &
wait
grep -c "Feed_update scan completed for agent '$TARGET_ID'" /var/wazuh-manager/logs/wazuh-manager.log   # expected 1
```
Set `scan_dispatch_queue_slots=1`. Tried the guide's literal two-concurrent-curls
approach, then a refined "occupy the single worker with a different agent
first, then fire the target pair" approach (both blind-delay and
log-triggered-timing variants), verifying server-side via `api.log` that
both admission calls genuinely reached the server each time. **Result every
time: delta=2** (both scans ran to completion independently, not coalesced) —
except one apparent delta=1 that turned out to be a **false positive**: a
`curl --next` connection-reuse quirk silently dropped the second request
client-side (confirmed by only one `api.log` entry existing for that attempt).

**Read the actual code** (`scanDispatcher.hpp`'s `tryEnqueue()`): the
coalescing check (`std::find` against `m_queue`) is real and looks correct —
an agent id sitting in the in-memory queue before the single worker dequeues
it does get deduped. But each PUT call's own round trip (Python RBAC +
wdb lookups + HTTP POST to modulesd's admission endpoint) measured
**18-40ms** end to end in `api.log`, while a scan itself completes in
**~15-20ms** — meaning by the time a second external request's admission
call actually reaches `tryEnqueue()`, the first request's item has very
likely already been dequeued and even finished. The "both still queued"
window this scenario needs appears to be sub-millisecond in practice and
not reliably reachable through the full external HTTP path with this
tooling. **Marking "code-path confirmed via source review, not
live-reproduced through the API"** — same treatment as B3/B6 — rather than
calling it a failure. A definitive check would need a C++-level unit/
integration test driving `tryEnqueue()` directly, not an external HTTP race.

##### D1b. Same agent, second request fires while first is in-flight (expect NOT coalesced) — **PASS**

```bash
TARGET_ID=<a dummy id>
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$TARGET_ID" -w "%{http_code}\n"
while ! grep -q "agent='$TARGET_ID'.*reason=feed_update" /var/wazuh-manager/logs/wazuh-manager.log; do sleep 0.001; done
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$TARGET_ID" -w "%{http_code}\n"
grep -c "Feed_update scan completed for agent '$TARGET_ID'" /var/wazuh-manager/logs/wazuh-manager.log

sudo sed -i '/scan_dispatch_queue_slots/d' /var/wazuh-manager/etc/wazuh-manager-internal-options.conf
sudo /var/wazuh-manager/bin/wazuh-manager-control restart
```
Fired one request for a target agent, polled the log until its
`Vulnerability scan start: agent='X' ... reason=feed_update` line confirmed
it was actually executing (not just admitted), then immediately fired a
second request for the same agent. Both returned 200 (confirmed via 2
separate `api.log` entries), and the completion-line delta was **2, not 1**
— confirms in-flight scans are correctly excluded from the coalescing dedup,
exactly as designed. Restored default queue slots afterward.

##### D2. Cross-node race — **NOT TESTED — open item, not a confirmed pass**

No command run — no multi-node cluster was available in this environment
(`wazuh-manager.conf`'s `<cluster>` block: one node, `node_type: master`,
`bind_addr: 127.0.0.1`) to actually race against. The design notes' trace
of `dapi.py:158-173` (`request_type='local_master'` always forwards a
non-master's request to the master, so there is only ever one
`VdScanDispatcher` in play) is a reasonable code-review inference, **not an
empirical result** — cross-node behavior for this endpoint has not
actually been exercised. This should be treated as an open item and run
for real on a 2-node cluster before being called verified, not marked
"non-issue by design" as if it were settled. The command it would take, on
an actual 2-node cluster:
```bash
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://<master_ip>:55000/agents/scan/vulnerability?agents_list=$TARGET_ID" -w "%{http_code}\n" &
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://<worker_ip>:55000/agents/scan/vulnerability?agents_list=$TARGET_ID" -w "%{http_code}\n" &
wait
```

##### D3. Disconnected-agent sweep fence — **out of scope; not exercised**

This scenario targets a pre-existing `vulnerability_scanner` fencing gap
(the sweep and an admin-triggered scan use separate code paths with no
shared queue, and the un-refcounted fence between them) that is unrelated
to the endpoint this PR adds. An earlier draft assumed a
`POST /manager/vulnerability/ondemand` endpoint could force the sweep to
run on demand — **that endpoint does not exist anywhere in this codebase**
(confirmed: zero matches in `spec.yaml`), which invalidated the original
approach entirely. There is no manual trigger for the sweep at all; it
only runs on the `feed-update-interval` timer.

```bash
sudo sed -i 's#<feed-update-interval>.*</feed-update-interval>#<feed-update-interval>30s</feed-update-interval>#' /var/wazuh-manager/etc/wazuh-manager.conf
sudo /var/wazuh-manager/bin/wazuh-manager-control restart

TARGET_ID=<a never_connected dummy agent>
for i in $(seq 1 100); do
  curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$TARGET_ID" >/dev/null &
  sleep 1
done
wait
grep "agent='$TARGET_ID'" /var/wazuh-manager/logs/wazuh-manager.log | grep -oE "scan (start|completed)" | uniq -c | awk '$1>1'

sudo sed -i 's#<feed-update-interval>30s</feed-update-interval>#<feed-update-interval>60m</feed-update-interval>#' /var/wazuh-manager/etc/wazuh-manager.conf
sudo /var/wazuh-manager/bin/wazuh-manager-control restart
```
A follow-up attempt using only the real endpoint (hammering
`PUT /agents/scan/vulnerability` for a `never_connected` dummy agent while
the interval was shortened to 30s, once per second for ~105s) produced 105
`scan start` / 105 `scan completed` log lines, perfectly alternating with
zero overlap — but since this only exercises a pre-existing internal
mechanism unrelated to this PR's own feature, it's noted here for
completeness and not counted as in-scope pass/fail evidence. Likely the
same sub-millisecond race-window limitation as D1a — narrower than
external HTTP hammering can reliably force. Reverted the feed-update
interval afterward.

##### D4. Scan lost on modulesd crash — **PASS (confirms accepted, documented behavior — not a bug)**

```bash
echo "vulnerability-detection.scan_dispatch_queue_slots=1" | sudo tee -a /var/wazuh-manager/etc/wazuh-manager-internal-options.conf
sudo /var/wazuh-manager/bin/wazuh-manager-control restart

TARGET_ID=<a dummy id>
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$TARGET_ID" -w "%{http_code}\n"
MPID=$(pgrep -f "wazuh-manager-modulesd$" | head -1); sudo kill -9 "$MPID"   # not SIGTERM
sudo /var/wazuh-manager/bin/wazuh-manager-control restart

curl -sk -u admin:admin "https://localhost:9200/wazuh-states-vulnerabilities/_count?q=wazuh.agent.id:$TARGET_ID"
grep "Feed_update scan completed for agent '$TARGET_ID'" /var/wazuh-manager/logs/wazuh-manager.log

sudo sed -i '/scan_dispatch_queue_slots/d' /var/wazuh-manager/etc/wazuh-manager-internal-options.conf
sudo /var/wazuh-manager/bin/wazuh-manager-control restart
```
Set `scan_dispatch_queue_slots=1`, fired one request (200, admitted), then
immediately `kill -9` (not TERM) on modulesd before it could dequeue. After
restart: `wazuh-states-vulnerabilities` count for the target stayed 0, and
no completion log line appears after the crash (last one predates it by
~10 minutes, from an earlier unrelated scenario). Confirms the queued scan
silently vanished with the process — expected, in-memory-only queue
behavior, not something to fix. Restored default queue slots afterward.

##### D5. Agent-deletion race — **REPRODUCED — genuine, previously-theoretical risk confirmed live**

**Setup note:** the guide's own DELETE call needs `&older_than=0s` added —
without it, `DELETE /agents?...&status=all` defaults to `older_than: 7d`
and rejects freshly-registered dummy agents outright with `error 1731`
("Agent is not eligible... older_than: 7d"), unrelated to the race itself.
Also: bringing the indexer back up doesn't actually widen the *execution*
race window (once admission succeeds, execution is just as fast as any
other scan, ~ms) — it only matters for getting past the 8004 preflight
check in the first place, and the manager's own indexer-readiness poll has
a real ~60s cadence per B4, so this needs waiting for an actual "ok" scan
response before firing the real race, not a blind timed restart.

```bash
TARGET_ID=<a dummy id>
sudo systemctl stop wazuh-indexer; sleep 65
curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$TARGET_ID" | jq .  # expect 8004
sudo systemctl start wazuh-indexer

while true; do
  resp=$(curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$TARGET_ID")
  [ "$(echo "$resp" | jq -r '.data.failed_items[0].error.code // "ok"')" = "ok" ] && break
  sleep 10
done

curl -sk -H "Authorization: Bearer $TOKEN" -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=$TARGET_ID" -w " scan_http=%{http_code}\n" &
curl -sk -H "Authorization: Bearer $TOKEN" -X DELETE "https://localhost:55000/agents?agents_list=$TARGET_ID&status=all&older_than=0s" -w " delete_http=%{http_code}\n" &
wait

curl -sk -u admin:admin "https://localhost:9200/wazuh-states-vulnerabilities/_count?q=wazuh.agent.id:$TARGET_ID"
grep "$TARGET_ID" /var/wazuh-manager/logs/wazuh-manager.log | grep -i "scan\|not found"
curl -sk -H "Authorization: Bearer $TOKEN" "https://localhost:55000/agents?agents_list=$TARGET_ID" | jq .   # expect 1701
```
**Result:** fired `PUT .../scan/vulnerability?agents_list=031` and
`DELETE /agents?agents_list=031&status=all&older_than=0s` concurrently.
Both returned HTTP 200. The log shows the scan ran through its **entire
normal success path** for agent 031 — OS scan, package scan (216 packages),
`Vulnerability scan completed: agent='031' type=full reason=feed_update`,
`VD scan succeeded for agent 031` — with **no** "agent not found in
global.db" skip message anywhere. Immediately after, `GET /agents?agents_list=031`
confirmed the agent was genuinely gone (`error.code: 1701`). **This directly
confirms outcome (b) from the guide: the scan does not re-verify agent
existence during execution, so a concurrent deletion does not reliably
short-circuit it** — the theorized risk is real, not just theoretical.

### DIFF
--- a/.github/actions/check_files/manager_base.csv
+++ b/.github/actions/check_files/manager_base.csv
@@ -81,6 +81,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/wazuh-manager/framework/wazuh/security.py,root,wazuh-manager,640,file,-rw-r-----,,
 /var/wazuh-manager/framework/wazuh/stats.py,root,wazuh-manager,640,file,-rw-r-----,,
 /var/wazuh-manager/framework/wazuh/task.py,root,wazuh-manager,640,file,-rw-r-----,,
+/var/wazuh-manager/framework/wazuh/vulnerability_scan.py,root,wazuh-manager,640,file,-rw-r-----,,
 /var/wazuh-manager/lib,root,wazuh-manager,750,directory,drwxr-x---,,
 /var/wazuh-manager/lib/libcontent_manager.so,root,wazuh-manager,750,file,-rwxr-x---,,
 /var/wazuh-manager/lib/libgcc_s.so.1,root,wazuh-manager,750,file,-rwxr-x---,,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 | [#38024](https://github.com/wazuh/wazuh/issues/38024) | Added the `POST /stats` HTTPS endpoint, which persists the statistics an agent reports as one document per agent in the `wazuh-agent-stats` index, replacing the previous report on every push. |
 | [#38007](https://github.com/wazuh/wazuh/issues/38007) | Added legacy `remote_upgrade` task delivery in `remoted`: a polling thread pushes pending Task Manager tasks to connected agents older than v5.0.0 over their existing session using the legacy six-step WPK push, gated on `remoted`'s HTTPS `verification_mode`. |
 | [#38157](https://github.com/wazuh/wazuh/issues/38157) | Added installation-time variables to customize the default `<remote>` configuration on source, DEB, and RPM manager installations. |
+| [#38553](https://github.com/wazuh/wazuh/issues/38553) | Added the `PUT /agents/scan/vulnerability` endpoint to trigger an on-demand vulnerability scan for one agent, a list of agents, or all agents. |
 
 #### Changed
 
--- a/api/api/controllers/agent_controller.py
+++ b/api/api/controllers/agent_controller.py
@@ -270,7 +270,10 @@ async def scan_agents(pretty: bool = False, wait_for_complete: bool = False,
     dapi = DistributedAPI(f=vulnerability_scan.scan_agents,
                           f_kwargs=remove_nones_to_dict(f_kwargs),
                           request_type='local_master',
-                          is_async=True,
+                          # Not is_async: scan_agents() makes one blocking VD HTTP call per
+                          # agent, so it must run off the API server's event loop (in the
+                          # process pool), not inline on it - see CHANGELOG/issue #38553.
+                          is_async=False,
                           wait_for_complete=wait_for_complete,
                           rbac_permissions=request.context['token_info']['rbac_policies'],
                           # No broadcasting: VD reads indexer-synced state, not a live per-node
--- a/api/api/controllers/agent_controller.py
+++ b/api/api/controllers/agent_controller.py
@@ -16,7 +16,7 @@ from api.models.agent_group_added_model import GroupAddedModel
 from api.models.agent_inserted_model import AgentInsertedModel
 from api.models.base_model_ import Body
 from api.util import parse_api_param, raise_if_exc, remove_nones_to_dict
-from wazuh import agent
+from wazuh import agent, vulnerability_scan
 from wazuh.core.cluster.dapi.dapi import DistributedAPI
 from wazuh.core.common import DATABASE_LIMIT
 from wazuh.core.results import AffectedItemsWazuhResult
@@ -240,6 +240,42 @@ async def restart_agents(pretty: bool = False, wait_for_complete: bool = False,
                           wait_for_complete=wait_for_complete,
                           rbac_permissions=request.context['token_info']['rbac_policies'],
                           broadcasting=True,  # Always broadcast for HTTPS stateless architecture
+                          logger=logger
+                          )
+    data = raise_if_exc(await dapi.distribute_function())
+
+    return json_response(data, pretty=pretty)
+
+
+async def scan_agents(pretty: bool = False, wait_for_complete: bool = False,
+                      agents_list: str = '*') -> ConnexionResponse:
+    """Request an on-demand vulnerability scan for all agents or a list of them.
+
+    Parameters
+    ----------
+    pretty : bool
+        Show results in human-readable format.
+    wait_for_complete : bool
+        Disable timeout response.
+    agents_list : str
+        List of agents IDs. Default: `*`
+
+    Returns
+    -------
+    ConnexionResponse
+        API response.
+    """
+    f_kwargs = {'agent_list': agents_list}
+
+    dapi = DistributedAPI(f=vulnerability_scan.scan_agents,
+                          f_kwargs=remove_nones_to_dict(f_kwargs),
+                          request_type='local_master',
+                          is_async=True,
+                          wait_for_complete=wait_for_complete,
+                          rbac_permissions=request.context['token_info']['rbac_policies'],
+                          # No broadcasting: VD reads indexer-synced state, not a live per-node
+                          # agent session, so the scan only needs to run once, on the master.
+                          broadcasting=False,
                           logger=logger
                           )
     data = raise_if_exc(await dapi.distribute_function())
--- a/api/api/controllers/test/test_agent_controller.py
+++ b/api/api/controllers/test/test_agent_controller.py
@@ -28,8 +28,9 @@ with patch('wazuh.common.wazuh_uid'):
             put_multiple_agent_single_group, put_upgrade_agents,
             put_upgrade_custom_agents, restart_agent,
             restart_agents, restart_agents_by_group,
-            reload_agent, reload_agents, reload_agents_by_group)
-        from wazuh import agent
+            reload_agent, reload_agents, reload_agents_by_group,
+            scan_agents)
+        from wazuh import agent, vulnerability_scan
         from wazuh.core.common import DATABASE_LIMIT
         from wazuh.tests.util import RBAC_bypasser
 
@@ -167,6 +168,32 @@ async def test_restart_agents(mock_exc, mock_dapi, mock_remove, mock_dfunc, mock
                                       is_async=True,
                                       wait_for_complete=False,
                                       broadcasting=True,
+                                      logger=ANY,
+                                      rbac_permissions=mock_request.context['token_info']['rbac_policies']
+                                      )
+    mock_exc.assert_called_once_with(mock_dfunc.return_value)
+    mock_remove.assert_called_once_with(f_kwargs)
+    assert isinstance(result, ConnexionResponse)
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("mock_request", ["agent_controller"], indirect=True)
+@patch('api.configuration.api_conf')
+@patch('api.controllers.agent_controller.DistributedAPI.distribute_function', return_value=AsyncMock())
+@patch('api.controllers.agent_controller.remove_nones_to_dict')
+@patch('api.controllers.agent_controller.DistributedAPI.__init__', return_value=None)
+@patch('api.controllers.agent_controller.raise_if_exc', return_value=CustomAffectedItems())
+async def test_scan_agents(mock_exc, mock_dapi, mock_remove, mock_dfunc, mock_exp, mock_request):
+    """Verify 'scan_agents' endpoint is working as expected."""
+    result = await scan_agents()
+    f_kwargs = {'agent_list': '*'
+                }
+    mock_dapi.assert_called_once_with(f=vulnerability_scan.scan_agents,
+                                      f_kwargs=mock_remove.return_value,
+                                      request_type='local_master',
+                                      is_async=True,
+                                      wait_for_complete=False,
+                                      broadcasting=False,
                                       logger=ANY,
                                       rbac_permissions=mock_request.context['token_info']['rbac_policies']
                                       )
--- a/api/api/controllers/test/test_agent_controller.py
+++ b/api/api/controllers/test/test_agent_controller.py
@@ -191,7 +191,7 @@ async def test_scan_agents(mock_exc, mock_dapi, mock_remove, mock_dfunc, mock_ex
     mock_dapi.assert_called_once_with(f=vulnerability_scan.scan_agents,
                                       f_kwargs=mock_remove.return_value,
                                       request_type='local_master',
-                                      is_async=True,
+                                      is_async=False,
                                       wait_for_complete=False,
                                       broadcasting=False,
                                       logger=ANY,
--- a/api/api/spec/spec.yaml
+++ b/api/api/spec/spec.yaml
@@ -4741,7 +4741,7 @@ paths:
                       - '001'
                       - '003'
                     total_affected_items: 2
-                    total_failed_items: 2
+                    total_failed_items: 3
                     failed_items:
                       - error:
                           code: 1701
@@ -4753,6 +4753,11 @@ paths:
                           message: "Vulnerability scan queue is full, please retry later"
                         id:
                           - '004'
+                      - error:
+                          code: 1764
+                          message: "Vulnerability scanning requires agent version 5.0 or higher."
+                        id:
+                          - '005'
                   message: "Scan was requested for some agents"
                   error: 2
 
--- a/api/api/spec/spec.yaml
+++ b/api/api/spec/spec.yaml
@@ -151,6 +151,15 @@ x-rbac-catalog:
         actions: ['agent:reload']
         resources: ['agent:id:050', 'agent:id:049']
         effect: "deny"
+    'agent:scan_vulnerability':
+      description: "Trigger an on-demand vulnerability scan on agents"
+      resources:
+       - $ref: '#/x-rbac-catalog/resources/agent:id'
+       - $ref: '#/x-rbac-catalog/resources/agent:group'
+      example:
+        actions: ['agent:scan_vulnerability']
+        resources: ['agent:id:050', 'agent:id:049']
+        effect: "deny"
     'agent:uninstall':
       description: "Check user's permission to uninstall agents"
       resources:
@@ -4688,6 +4697,63 @@ paths:
                           - '010'
                           - '011'
                   message: "Restart command was not sent to some agents"
+                  error: 2
+
+        '400':
+          $ref: '#/components/responses/ResponseError'
+        '401':
+          $ref: '#/components/responses/UnauthorizedResponse'
+        '403':
+          $ref: '#/components/responses/PermissionDeniedResponse'
+        '405':
+          $ref: '#/components/responses/InvalidHTTPMethodResponse'
+        '429':
+          $ref: '#/components/responses/TooManyRequestsResponse'
+
+  /agents/scan/vulnerability:
+    put:
+      tags:
+        - Agents
+      summary: "Scan agents for vulnerabilities"
+      description: "Request an on-demand vulnerability scan for all agents or a list of them"
+      operationId: api.controllers.agent_controller.scan_agents
+      x-rbac-actions:
+        - $ref: '#/x-rbac-catalog/actions/agent:scan_vulnerability'
+      parameters:
+        - $ref: '#/components/parameters/pretty'
+        - $ref: '#/components/parameters/wait_for_complete'
+        - $ref: '#/components/parameters/agents_list'
+      responses:
+        '200':
+          description: "Agent scan requested"
+          content:
+            application/json:
+              schema:
+                allOf:
+                  - $ref: '#/components/schemas/ApiResponse'
+                  - type: object
+                    properties:
+                      data:
+                          $ref: '#/components/schemas/AllItemsResponseAgentIDs'
+                example:
+                  data:
+                    affected_items:
+                      - '001'
+                      - '003'
+                    total_affected_items: 2
+                    total_failed_items: 2
+                    failed_items:
+                      - error:
+                          code: 1701
+                          message: "Agent does not exist"
+                        id:
+                          - '002'
+                      - error:
+                          code: 8005
+                          message: "Vulnerability scan queue is full, please retry later"
+                        id:
+                          - '004'
+                  message: "Scan was requested for some agents"
                   error: 2
 
         '400':
--- a/docs/ref/modules/rbac/README.md
+++ b/docs/ref/modules/rbac/README.md
@@ -135,5 +135,8 @@ The following RBAC actions control agent operations:
 | `agent:reload` | `agent:id`, `agent:group` | Reload agent configuration without restart (requires agent v5.0.0+) |
 | `agent:upgrade` | `agent:id`, `agent:group` | Upgrade agents |
 | `agent:uninstall` | `*` | Check permission to uninstall agents |
+| `agent:scan_vulnerability` | `agent:id`, `agent:group` | Request an on-demand vulnerability scan |
 
 The `agent:reload` action was introduced in v5.0.0 alongside a task-based restart/reload dispatch mechanism (agents are grouped into chunks and reloaded via `create_reload_tasks`), replacing the previous Active Response-based restart approach.
+
+The `agent:scan_vulnerability` action was introduced in v5.0.0 alongside `PUT /agents/scan/vulnerability`, the operator-facing endpoint to trigger an on-demand Vulnerability Detection scan, complementing the previous agent-initiated-only scan trigger.
--- a/docs/ref/modules/server-api/api-reference.md
+++ b/docs/ref/modules/server-api/api-reference.md
@@ -202,6 +202,20 @@ curl -k -X GET "https://localhost:55000/overview/agents?pretty=true" \
   -H "Authorization: Bearer $TOKEN"
 ```
 
+#### Trigger vulnerability scan
+
+**`PUT /agents/scan/vulnerability`** — Request an on-demand Vulnerability Detection scan for all agents or a specified list. All agents are selected by default if `agents_list` is not specified.
+
+```bash
+# Scan all agents
+curl -k -X PUT "https://localhost:55000/agents/scan/vulnerability?pretty=true" \
+  -H "Authorization: Bearer $TOKEN"
+
+# Scan specific agents
+curl -k -X PUT "https://localhost:55000/agents/scan/vulnerability?agents_list=001,003&pretty=true" \
+  -H "Authorization: Bearer $TOKEN"
+```
+
 ---
 
 ### Groups
@@ -347,6 +361,7 @@ Other MITRE endpoints: `/mitre/tactics`, `/mitre/groups`, `/mitre/software`, `/m
 | GET | `/agents/outdated` | Outdated agents |
 | PUT | `/agents/upgrade` | Upgrade agents |
 | PUT | `/agents/upgrade_custom` | Custom upgrade |
+| PUT | `/agents/scan/vulnerability` | Request on-demand vulnerability scan |
 | GET | `/agents/uninstall` | Check whether the calling user has permission to uninstall the given agents |
 | GET | `/agents/stats/distinct` | Distinct fields |
 | GET | `/agents/summary` | Summary |
--- a/docs/ref/modules/vulnerability-scanner/api-reference.md
+++ b/docs/ref/modules/vulnerability-scanner/api-reference.md
@@ -11,6 +11,74 @@ For a quick reference, the vulnerabilities can be retrieved using the **GET /waz
 - OS detection IDs follow: `<agentId>_<osName>_<osVersion>_<cveId>`.
 - Upserts represent active detections; deletes mark detections as solved.
 
+## Requesting an on-demand scan
+
+`PUT /agents/scan/vulnerability` requests an on-demand VD scan for one agent, a list of agents, or
+all agents, without waiting for the periodic feed-update sweep. It requires the
+`agent:scan_vulnerability` RBAC action (resources `agent:id`/`agent:group`) and always executes on
+the cluster master. See
+[architecture.md's Feed-update rescan section](architecture.md#feed-update-rescan-scanvd--rescandisconnectedagents)
+for how this endpoint fits alongside the agent-self-triggered `/scan/vd` path and the
+disconnected-agent sweep, including the cross-node race and admission-chain limitations that apply
+to it.
+
+### Parameters
+
+| Parameter | Type | Description |
+|---|---|---|
+| `agents_list` | query, comma-separated agent IDs | Agents to scan. Defaults to all agents when omitted. |
+| `pretty` | query, boolean | Show results in human-readable format. |
+| `wait_for_complete` | query, boolean | Disable the request timeout. |
+
+### Response
+
+The response is a standard `AffectedItemsWazuhResult`: `affected_items` lists the agent IDs VD
+accepted for scanning, `failed_items` groups agent IDs by the error that rejected them.
+
+```json
+{
+  "data": {
+    "affected_items": ["001", "003"],
+    "total_affected_items": 2,
+    "total_failed_items": 2,
+    "failed_items": [
+      {"error": {"code": 1701, "message": "Agent does not exist"}, "id": ["002"]},
+      {"error": {"code": 8005, "message": "Vulnerability scan queue is full, please retry later"}, "id": ["004"]}
+    ]
+  },
+  "message": "Scan was requested for some agents",
+  "error": 2
+}
+```
+
+An agent ID in `affected_items` means VD accepted the admission request (`200`) — it does not
+guarantee a scan actually ran. VD silently no-ops for an agent with no package inventory synced
+yet (a freshly-enrolled agent, one hit by a manager-side inventory sync gap, or a legacy pre-5.0
+agent, which never populates this inventory by design); the only trace of that is a
+`wazuh-manager.log` line, since neither the admission response nor `ScanTriggerResult` distinguish
+it from a real completed scan.
+
+### Scan admission error codes
+
+Each entry maps to one of the `503` reasons VD's admission gate returns — see
+[architecture.md's admission-gate description](architecture.md#feed-update-rescan-scanvd--rescandisconnectedagents)
+for what each check actually verifies.
+
+| Code | VD reason string | Message |
+|---|---|---|
+| 8001 | `vd_not_initialized` | Vulnerability detection module has not been initialized yet |
+| 8002 | `feed_not_ready` | Vulnerability detection feed is not ready yet |
+| 8003 | `scanner_not_ready` | Vulnerability scanner is not ready yet |
+| 8004 | `indexer_unavailable` | Indexer is not available for vulnerability scanning |
+| 8005 | `scan_queue_full` | Vulnerability scan queue is full, please retry later |
+| 8006 | `shutting_down` | Vulnerability scanner module is shutting down |
+
+`8005` reflects the depth of the shared on-demand dispatch queue — see
+[`scan_dispatch_queue_slots`](configuration.md#scan_dispatch_queue_slots). An admission rejection
+reason VD returns that isn't one of the six above falls back to the generic reserved code `8007`.
+Agent IDs that don't exist fail with the existing `1701` (`Agent does not exist`), resolved before
+any VD call is made.
+
 ## Field notes
 
 - `host.os.full` is built from OS `name` + `version` (macOS uses `codename`).
--- a/docs/ref/modules/vulnerability-scanner/api-reference.md
+++ b/docs/ref/modules/vulnerability-scanner/api-reference.md
@@ -40,10 +40,11 @@ accepted for scanning, `failed_items` groups agent IDs by the error that rejecte
   "data": {
     "affected_items": ["001", "003"],
     "total_affected_items": 2,
-    "total_failed_items": 2,
+    "total_failed_items": 3,
     "failed_items": [
       {"error": {"code": 1701, "message": "Agent does not exist"}, "id": ["002"]},
-      {"error": {"code": 8005, "message": "Vulnerability scan queue is full, please retry later"}, "id": ["004"]}
+      {"error": {"code": 8005, "message": "Vulnerability scan queue is full, please retry later"}, "id": ["004"]},
+      {"error": {"code": 1764, "message": "Vulnerability scanning requires agent version 5.0 or higher."}, "id": ["005"]}
     ]
   },
   "message": "Scan was requested for some agents",
@@ -53,10 +54,13 @@ accepted for scanning, `failed_items` groups agent IDs by the error that rejecte
 
 An agent ID in `affected_items` means VD accepted the admission request (`200`) — it does not
 guarantee a scan actually ran. VD silently no-ops for an agent with no package inventory synced
-yet (a freshly-enrolled agent, one hit by a manager-side inventory sync gap, or a legacy pre-5.0
-agent, which never populates this inventory by design); the only trace of that is a
-`wazuh-manager.log` line, since neither the admission response nor `ScanTriggerResult` distinguish
-it from a real completed scan.
+yet (a freshly-enrolled agent, or one hit by a manager-side inventory sync gap); the only trace of
+that is a `wazuh-manager.log` line, since neither the admission response nor `ScanTriggerResult`
+distinguish it from a real completed scan.
+
+A legacy agent reporting a known version below `5.0.0` is instead rejected up front with `1764`,
+before any VD call is made; an agent whose version is unknown, empty, or `'N/A'` isn't covered by
+that check and is still forwarded to VD, where it falls into the same silent no-op described above.
 
 ### Scan admission error codes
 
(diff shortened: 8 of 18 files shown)


MiguelazoDS:
## Event-loop-freeze fix + live verification (3000-agent scan)

A code review after the manual QA pass in this PR found that `scan_agents()` was declared `async def` but its loop made a fully synchronous, blocking VD HTTP call per agent with zero real `await`s. Combined with the controller's `is_async=True`, this ran the entire loop directly on `wazuh-apid`'s own asyncio event loop instead of a separate process .  With a large-fleet scan would freeze for the whole scan.

**Fix:** `scan_agents()` is now a plain `def` (dropped `post_proc_func=async_list_handler`), and the controller uses `is_async=False`, so the whole loop runs in the process pool instead of inline on the event loop — the same isolation `delete_agents()`/`upgrade_agents()` already rely on for their own bulk operations.

### Live verification

Ran a dedicated concurrency test against a real manager built with the fix: registered a 3000-agent dummy fleet, fired `PUT /agents/scan/vulnerability?agents_list=<all 3000>` (**HTTP 200, 1.985s**), and sampled three other endpoints once a second throughout, idle baseline vs. during-scan:

| Probe | Idle baseline | During scan | What it measures |
|---|---|---|---|
| `GET /` | ~10ms | **46ms** | Event-loop responsiveness — runs inline, no process pool involved at all |
| `GET /cluster/status` | ~7ms | **7ms** | Shares `scan_agents()`'s process pool (`is_async=False`) |
| `GET /agents?limit=1` | ~20ms | **1.913s** | Also shares that same process pool |

**`GET /` staying at 46ms confirms the event-loop-freeze bug is fixed** — under the pre-fix code this would have stalled for the scan's full duration, since a frozen event loop can't service *any* connection, not just ones routed through `DistributedAPI`.

MiguelazoDS:
### Tests Introduced

`framework/wazuh/tests/test_vulnerability_scan.py::test_scan_agents_happy_path` — extended with version-gate coverage (commit `a7d46de15b`):

| Version | Case | Expected |
|---|---|---|
| `v5.0.0` | known, boundary | forwarded |
| `v5.1.3`, `v5.0.2` | known, above 5.0 | forwarded |
| `'N/A'` | unknown | forwarded, not filtered |
| `''` (empty) | unknown | forwarded, not filtered |
| `v4.9.2` | known, below 5.0 | **rejected — `1764`** |

MiguelazoDS:
There's a remaining API endpoint test that seems to fail for other PRs too 
https://github.com/wazuh/wazuh/actions/runs/33181945408

jotacarma90 (approved):

PR #38620, opened 2026-08-26
Enhancement/38296 indexer connector staging sync

### LINKED ISSUE #38296: Review Indexer Connector staging synchronization
## Description

Review the thread-safety of the staging buffer and the default flush interval.

## Tasks

* Review synchronization.
* Evaluate locking strategy.
* Review one-hour default flush interval.
* Propose improvements if necessary.

## Testing

* Concurrency testing.
* Long-running tests.

## Definition of Done

* Thread safety validated.
* Flush interval reviewed.

LucioDonda:
## Description

`inventory_sync_server` answers a `POST /stateful` session only after the documents it staged have
actually reached the indexer — REQ-SYNC-4: *200 means flushed*. The pipeline worker owns that
contract: it stages a batch, flushes it, and only then calls `respond()` on the responder it kept.

The `IIndexerConnectorSync` it stages into also runs a **background flush timer thread**, and that
thread is a second party on the very same `m_bulkData` buffer. When the timer's `processBulk()`
fails with a terminal status, the buffer is cleared and the exception is caught by a `catch` that
holds **no responder** — it can only log. The worker's later `flush()` then finds an empty buffer,
returns cleanly, and answers **`200`** for documents that never landed.

The one-hour `PIPELINE_CONNECTOR_FLUSH_INTERVAL_SECS` this module used was a mitigation, not a fix:
it made the tick rare, not impossible, and it left the buffer under two owners. Reviewing the
staging synchronization turned up three further defects on the same buffer (details below).

Closes #38296

## Proposed Changes

**`shared_modules/indexer_connector`**

- **A failed 413 split no longer leaves bytes behind.** `processBulk()`'s recursive split path
  cleared `m_bulkData`/`m_boundaries` only on the success path, so a split that failed left the
  already-attempted payload staged and the *next* flush re-sent it. The clear now runs through a
  `DEFER` guard, so it happens on every exit path.
- **`flush_interval_seconds == 0` now means "no background flush thread".** It previously created
  the thread anyway, where `m_cv.wait_for(lock, 0s)` returns immediately — a **busy loop**, not a
  disable. Anyone who tried to turn the timer off by configuration got a spinning thread. The
  constructor now returns before creating the thread.
- Corrected a stale doc comment in `indexerConnector.hpp` about `monitoring_interval_seconds`.

**`wazuh_modules/inventory_sync_server`**

- **The pipeline connectors' flush timer is turned OFF (`0`) instead of demoted to 3600 s.** With
  no timer thread there is exactly one owner of the staging buffer, so a flush failure always
  surfaces on the thread that has to answer for it. This is the fix for the silent-`200` defect.
- **The indexer connector configs are rebuilt on every start attempt**, and `needLane` joins the
  gating condition. Previously a retry in which only the VD scan lane still needed a connector
  skipped the config build and constructed the lane from an empty configuration.
- `indexerConnectorSyncAdapter`'s `scopeLock` rationale updated: it no longer guards against the
  timer thread (there is none), it honors the connector's caller-locks contract.

**Build system**

- **New `TSANITIZE=YES` developer flag** (`src/Makefile` → `-DTSANITIZE=ON` →
  `-fsanitize=thread`), mirroring the existing `FSANITIZE` plumbing; the two are mutually
  exclusive by `FATAL_ERROR` since TSAN cannot link ASan. `TSANITIZE` also drops `--coverage`
  from `add_coverage_libraries`: gcov counters are non-atomic globals, and with coverage on a
  TSAN run drowns in tens of thousands of false races on `__gcov0.*` symbols before a single
  real finding can surface. No effect on shipped artifacts — Debug/test builds only.

### Results and Evidence

The defect was reproduced live on the **base** build with a fault-injecting reverse proxy in front
of the indexer. The proxy fails a flat 25 % of `POST /_bulk` (it forwards them to a route the
indexer does not serve, which answers `405`), so both the worker's flushes and the timer's flushes
fail at the same rate. Attribution then comes from the manager log, which separates the two
**exactly and by construction**:

| Log line | Emitted by | Meaning |
|---|---|---|
| `Error processing bulk: ...` | `indexerConnectorSyncImpl.hpp:884` — the timer thread's `catch`, the **only** occurrence of this string in the tree | a staged buffer was discarded with **nobody to answer** |
| `A bulk flush of N session(s) failed` | `syncPipeline.cpp:316` — the worker's `catch` | the healthy path: N sessions were told the truth |

<details><summary>Reproduction run — 2026-08-28 21:01:59Z, base build, 128 agents, 135 s</summary>
<p>

```text
/_bulk seen by the injector:       77
faults delivered (405):            22
TIMER discards (the bug):          14
worker flush failures (ok):         8
Cleans deleteByQuery failures:     22
```

`22 = 14 + 8`: every injected fault landed on exactly one flush, and **14 of them hit a flush that
had no responder attached**. Those 14 buffers were dropped without a single session being told.

</p>
</details>

<details><summary>Manager log excerpt — three silent discards between two honest failures</summary>
<p>

```text
2026/08/28 21:02:03 | A bulk flush of 1 session(s) failed: Client error. The agents will retry.
2026/08/28 21:02:04 | Error processing bulk: Client error
2026/08/28 21:02:04 | Error processing bulk: Client error
2026/08/28 21:02:04 | Error processing bulk: Client error
2026/08/28 21:02:05 | Error processing bulk: Client error
2026/08/28 21:02:06 | A bulk flush of 26 session(s) failed: Client error. The agents will retry.
2026/08/28 21:02:07 | Error processing bulk: Client error
2026/08/28 21:02:08 | A bulk flush of 25 session(s) failed: Client error. The agents will retry.
```

The `A bulk flush of N session(s) failed` lines name a session count because a responder exists.
The `Error processing bulk` lines cannot: the timer thread has nothing to answer to. Those are the
sessions that go on to receive `200`.

</p>
</details>

<details><summary>Preconditions the reproduction depends on (four runs were lost to them)</summary>
<p>

Each of these silently produces a run that looks like a clean bill of health. They are now all
asserted by `prep-repro-run.sh` instead of living in a checklist.

| # | Precondition | What reverts it | Symptom when wrong |
|---|---|---|---|
| 1 | `<indexer><hosts>` → `http://127.0.0.1:9280` | a manager reinstall | injector sees **0** `/_bulk`; every counter `0`; summary looks healthy |
| 2 | `PIPELINE_CONNECTOR_FLUSH_INTERVAL_SECS 1` in the deployed binary | `git checkout` / `git pull` (the edit is uncommitted) | `TIMER discards = 0` — arithmetic, not a result: at 3600 s no tick fits in a 135 s run |
| 3 | `..._indexer_sync_max_bulk_size` raised | a manager reinstall | the worker cuts constantly, the timer never owns a flush |
| 4 | The run must last minutes | scenario default `repeat_until: "0"` | ~12 flushes in 3.8 s — too few dice rolls |

An earlier revision of the injector also failed only bodies **under 10 MB**, on the theory that a
larger body must be a worker cut. That rule is wrong — both threads POST the same accumulating
buffer, and the pipeline's 10 MiB threshold counts wire FlatBuffer bytes while the proxy sees
staged NDJSON — and it suppressed **29 %** of the faults, including every one in the run that
first appeared to show nothing.

</p>
</details>

<details><summary>Scope of the evidence — what this run does and does not prove</summary>
<p>

**Proved:** the mechanism fires. 14 staged buffers were discarded inside a thread that cannot
respond, on the base build, under ordinary indexer errors.

**Not proved by this run:** the end-state document loss. Two confounds make the per-agent
reconciliation unusable here, and both are properties of the load shape, not of the defect:

- `repeat_until` replays the lane, and a `full_resync` begins with a `Cleans` — a later pass
  rewrites the agent's 800 documents and **heals** an earlier loss.
- The run ended with `abandoned_on_drain: 128` (one per agent). An agent whose final pass is cut
  off after its `Cleans` but before its writes complete lands on a partial count all by itself.

The final index shows 84 agents at exactly 800, 39 at 0, and 5 partial (12, 568, 578, 730, 747) —
a distribution drain-abandonment produces on its own, so those 5 are **not** attributed to the bug.

A conclusive end-state measurement needs a **single** pass (`repeat_until: "0"`, no healing)
throttled with `pacing.requests_per_second` so that one pass spans enough wall-clock for timer
ticks. Tracked as remaining work below.

</p>
</details>

<details><summary>Reproduction harness — injector config, prep/verify script, scenario</summary>
<p>

**`faultinject-nginx.conf`** — reverse proxy in front of the indexer on `127.0.0.1:9280`.

```nginx
worker_processes 1;
pid /tmp/faultinject-nginx.pid;
error_log /tmp/faultinject-nginx-error.log warn;
daemon on;

events {
    worker_connections 1024;
}

http {
    # Injector-aware access log. Without $content_length/$inject you cannot tell "the fault
    # never fired" from "the fault fired and nothing was lost" -- the request line is /_bulk
    # either way. Fields: req_len = the body nginx saw (diagnostic only -- it does NOT gate the fault),
    # up = what the indexer answered (405 on the bogus path == fault delivered).
    # $inject is computed for EVERY request, but it only has an effect under `location /_bulk`.
    # Log $request_uri too, or `grep -c inject=1` counts agent-metadata writes and health checks
    # as injected faults and overstates the fault rate several-fold.
    log_format inject '$time_local req=$request_uri status=$status up=$upstream_status '
                      'req_len=$content_length inject=$inject uri=$bulk_uri '
                      'rt=$request_time urt=$upstream_response_time resp=$body_bytes_sent';
    access_log /tmp/faultinject-nginx-access.log inject;

    # ---------------------------------------------------------------------------------------
    # MANDATORY. nginx defaults client_max_body_size to 1m; the pipeline's group-commit cut is
    # 10 MiB (DEFAULT_BULK_FLUSH_BYTES). Without this, nginx answers 413 to nearly every worker
    # flush, the connector takes its recursive split path, and ~every session ends up 500 --
    # zero sessions answered 200, so the reconciliation has no population and the bug CANNOT
    # be observed. This single line is what made the first run useless.
    # ---------------------------------------------------------------------------------------
    client_max_body_size 0;
    client_body_buffer_size 1m;

    # Bodies are 10-20 MB. Buffering them to a temp file adds seconds of latency and pushed the
    # first run's session p50 to 54 s (max 64.8 s) against nginx's 60 s proxy defaults.
    proxy_request_buffering off;
    proxy_connect_timeout 10s;
    proxy_send_timeout    300s;
    proxy_read_timeout    300s;

    # A `return` never reads the request body; for a 10 MB POST the client can see a reset
    # (counted as transport_error -> the connector's RETRY path, not the loss path) instead of
    # a clean status. Lingering close drains it. Only matters for FAULT_MODE=return below.
    lingering_close   on;
    lingering_time    120s;
    lingering_timeout 10s;

    # --- Fault targeting -------------------------------------------------------------------
    # Fail a flat percentage of /_bulk POSTs. Deliberately NOT size-aware.
    #
    # An earlier revision only failed bodies under 10 MB, on the theory that a big body must be
    # the worker's byte-threshold cut (whose failure is the harmless path: 500/503 back to the
    # sessions, the agent retries) while a short body must be a mid-batch snapshot by the timer
    # (whose failure is the silent one). That rule is wrong three times over:
    #
    #   1. Both threads POST the SAME accumulating buffer. Under sustained load the timer's
    #      snapshot is just as large as a worker cut -- size does not identify the caller.
    #   2. The units do not even line up. The pipeline cuts at 10 MiB of WIRE FlatBuffer bytes
    #      (DEFAULT_BULK_FLUSH_BYTES) while $content_length here is staged NDJSON -- a worker cut
    #      routinely arrives as a ~7 MB body and would be misread as a timer snapshot anyway.
    #   3. It can suppress the fault outright. In the 2026-08-28 16:45 run the filter let through
    #      zero faults out of 73 bulks; the run reconciled clean because nothing was ever broken.
    #
    # nginx cannot tell who initiated a flush and does not need to: the MANAGER LOG already
    # discriminates, and exactly. A timer-side discard is the only thing that emits
    #     "Error processing bulk"      (indexerConnectorSyncImpl.hpp:884, timer catch -- unique)
    # while a worker-side failure is the only thing that emits
    #     "A bulk flush of N session(s) failed"   (syncPipeline.cpp:316)
    # So: inject blindly here, attribute in the log. See manual-flush-validation-tests.md.
    #
    # RATE. Trade-off, both directions matter:
    #   too low  -> few timer flushes are hit, the race may never fire in a short run;
    #   too high -> every WORKER flush fails too, ~no session is answered 200, and the
    #               reconciliation has no population left to check (that is how the very first
    #               run became a false negative). Do NOT set this to 100%.
    # 25% keeps ~3 of 4 sessions at 200 while still hitting one timer tick in four.
    split_clients "${request_id}" $inject {
        25%     1;
        *       0;
    }

    # proxy_pass cannot carry a URI inside an `if`, so the fault is expressed as the upstream
    # path itself: doomed requests go to a route the indexer does not serve.
    map $inject $bulk_uri {
        1       "/_bulk_injected_fault";
        default "/_bulk";
    }

    server {
        listen 127.0.0.1:9280;

        location /_bulk {
            # FAULT_MODE=proxy (default): send the doomed request to a path the indexer does not
            # serve. A REAL server reads the whole body and answers 400 "no handler found", which
            # lands in the connector's terminal clear-and-throw branch exactly like a 500 -- with
            # no body-discard risk and no synthetic response. Probe the status once:
            #   curl -sk -u admin:admin -o /dev/null -w '%{http_code}\n' \
            #     -XPOST https://127.0.0.1:9200/_bulk_injected_fault -d '{}'
            # Any status EXCEPT 413/409/429 drives the loss path. If it answers 429, use
            # FAULT_MODE=return instead (uncomment the block below and set $bulk_uri to /_bulk).
            proxy_pass https://127.0.0.1:9200$bulk_uri$is_args$args;

            # FAULT_MODE=return (synthetic). Uncomment, and pin $bulk_uri to "/_bulk" in the map.
            #   if ($inject) {
            #       add_header Content-Type application/json always;
            #       return 500 '{"error":{"type":"injected_fault","reason":"benchmark injector"},"status":500}';
            #   }

            proxy_ssl_verify off;
            proxy_http_version 1.1;
            proxy_set_header Connection "";
        }

        location / {
            proxy_pass https://127.0.0.1:9200;
            proxy_ssl_verify off;
            proxy_http_version 1.1;
            proxy_set_header Connection "";
        }
    }
}
```

**`prep-repro-run.sh`** — asserts every precondition before a run and reads the verdict after it.
Exit `2` means "restart the manager first"; `verify` refuses to report a verdict when the injector
saw no traffic, or when the compiled flush interval makes a zero mathematically inevitable.

```bash
#!/usr/bin/env bash
# Prepare (or verify) one fault-injection reproduction pass for issue #38296.
#
#   ./prep-repro-run.sh            prepare the next pass
#   ./prep-repro-run.sh verify     read the verdict after a pass
#
# Idempotent: safe to re-run before every pass, and after every manager reinstall --
# a reinstall silently reverts BOTH the indexer host and the internal-options override.
set -uo pipefail

WAZUH_HOME=${WAZUH_HOME:-/var/wazuh-manager}
CONF="$WAZUH_HOME/etc/wazuh-manager.conf"
OPTS="$WAZUH_HOME/etc/wazuh-manager-internal-options.conf"
MLOG="$WAZUH_HOME/logs/wazuh-manager.log"
KEY=wazuh_modules.inventory_sync_server_indexer_sync_max_bulk_size
VAL=104857600                      # 100 MiB == the option's maximum (bounds 4096 .. 100 MiB)
INJECTOR=http://127.0.0.1:9280
IDX=wazuh-states-fim-files
ES=${ES:-https://127.0.0.1:9200}
ES_AUTH=${ES_AUTH:-admin:admin}
ACCESS_LOG=/tmp/faultinject-nginx-access.log
REPO=${WAZUH_REPO:-/workspaces/devContainer/wazuh}
FACADE="$REPO/src/wazuh_modules/inventory_sync_server/src/inventorySyncServerFacade.hpp"
MODULESD="$WAZUH_HOME/bin/wazuh-manager-modulesd"
rc=0

say() { printf '%-6s %s\n' "$1" "$2"; }

# --------------------------------------------------------------------------------------
# verify mode -- run this AFTER a pass.
# --------------------------------------------------------------------------------------
if [ "${1:-}" = verify ]; then
    # grep -c prints "0" AND exits 1 on no-match, so `|| echo 0` would emit "0\n0" and every
    # numeric test below would blow up. Capture, then normalise a missing file to 0.
    count() { local n; n=$(eval "$1" 2>/dev/null); [ -n "$n" ] || n=0; printf '%s' "$n"; }
    bulks=$(count  "grep -c 'req=/_bulk ' '$ACCESS_LOG'")
    faults=$(count "grep 'req=/_bulk ' '$ACCESS_LOG' | grep -c 'status=405'")
    timer=$(count  "grep -c 'Error processing bulk' '$MLOG'")
    worker=$(count "grep -c 'A bulk flush of' '$MLOG'")
    cleans=$(count "grep -c 'deleteByQuery did not delete every matching document' '$MLOG'")

    printf '  %-34s %s\n' "/_bulk seen by the injector:" "$bulks"
    printf '  %-34s %s\n' "faults delivered (405):"      "$faults"
    printf '  %-34s %s\n' "TIMER discards (the bug):"    "$timer"
    printf '  %-34s %s\n' "worker flush failures (ok):"  "$worker"
    printf '  %-34s %s\n' "Cleans deleteByQuery failures:" "$cleans"
    echo
    if [ "$bulks" -eq 0 ]; then
        say FAIL "the injector saw no bulk traffic at all -- the manager is NOT talking to it."
        say ""  "   This is the failure mode that wasted the 2026-08-28 20:37 run. Check <indexer><hosts>."
        exit 1
    fi
    if [ "$faults" -eq 0 ]; then
        say FAIL "traffic reached the injector but nothing was failed -- check the split_clients rate."
        exit 1
    fi
    if [ "$timer" -eq 0 ]; then
        secs=$(grep -oE 'PIPELINE_CONNECTOR_FLUSH_INTERVAL_SECS \{[0-9]+\}' "$FACADE" 2>/dev/null | grep -oE '[0-9]+')
        if [ "${secs:-3600}" -gt 5 ] 2>/dev/null; then
            say FAIL "flush interval is ${secs}s: the timer cannot tick inside this run. The 0 above"
            say ""   "   is arithmetic, not a result. Rebuild with {1} and run again."
            exit 1
        fi
        say WARN "faults landed but none hit the timer. Raise the fleet, lengthen repeat_until, or"
        say ""   "   check that faults are not all landing on worker flushes (faults == worker below)."
    else
        say HIT  "the race fired $timer time(s). Reconcile: sessions acked 200 vs docs in $IDX."
    fi
    [ "$cleans" -gt 0 ] && say NOTE "$cleans session(s) answered 500 from a Cleans deleteByQuery version"
    [ "$cleans" -gt 0 ] && say ""   "   conflict, not from a flush -- repeat_until replays Cleans under load. Expected noise."
    exit 0
fi

# --------------------------------------------------------------------------------------
# -1 -- The flush interval is a COMPILE-TIME constant. The internal option
#       inventory_sync_server_indexer_sync_flush_interval_seconds is parsed and then
#       unconditionally overwritten at inventorySyncServerFacade.hpp:901, so it has never had
#       an effect -- only a rebuild changes the timer. At the shipped 3600 s a 135 s run cannot
#       see a single tick, and "TIMER discards = 0" is arithmetic, not evidence. A branch
#       switch or a pull silently reverts this edit; that is what happened on 2026-08-28 20:54.
# --------------------------------------------------------------------------------------
if [ -f "$FACADE" ]; then
    secs=$(grep -oE 'PIPELINE_CONNECTOR_FLUSH_INTERVAL_SECS \{[0-9]+\}' "$FACADE" | grep -oE '[0-9]+')
    if [ "${secs:-3600}" -le 5 ] 2>/dev/null; then
        say OK "flush interval in source is ${secs}s"
        if [ -f "$MODULESD" ] && [ "$FACADE" -nt "$MODULESD" ]; then
            say FAIL "$FACADE is NEWER than the deployed modulesd -- rebuild and redeploy first"
            rc=1
        fi
    else
        say FAIL "flush interval is ${secs:-?}s -- the timer will not tick inside a short run."
        say ""   "   Edit $FACADE:102 to {1}, rebuild wazuh-modulesd, redeploy. A pull/checkout reverts it."
        rc=1
    fi
else
    say SKIP "$FACADE not found -- cannot check the flush interval"
fi

# --------------------------------------------------------------------------------------
# 0 -- THE ONE THAT ACTUALLY MATTERS. A manager pointed at :9200 talks straight to the
#      indexer, every flush succeeds, and the run is a silent no-op: zero faults, zero
#      log lines, a summary that looks healthy. A reinstall reverts this every time.
# --------------------------------------------------------------------------------------
if [ -f "$CONF" ]; then
    host_line=$(grep -m1 '<host>' "$CONF")
    if printf '%s' "$host_line" | grep -q '9280'; then
        say OK "indexer host already points at the injector"
    else
        cp "$CONF" "$CONF.pre-repro.bak"
        sed -i "s|<host>.*</host>|<host>$INJECTOR</host>|" "$CONF"
        say FIXED "indexer host $(printf '%s' "$host_line" | sed 's/[[:space:]]*//') -> $INJECTOR"
        say NOTE  "backup at $CONF.pre-repro.bak -- RESTART the manager for this to take effect"
        rc=2
    fi
else
    say SKIP "$CONF not found (manager not installed)"
    rc=1
fi

# 1 -- make the worker stop cutting on bytes, so the flush TIMER becomes the dominant
#      flusher and the race has something to race with.
if [ -f "$OPTS" ]; then
    if grep -q "^${KEY}=${VAL}$" "$OPTS"; then
        say OK "internal option already set (${KEY}=${VAL})"
    elif grep -q "^${KEY}=" "$OPTS"; then
        sed -i "s|^${KEY}=.*|${KEY}=${VAL}|" "$OPTS"
        say FIXED "internal option updated -> ${KEY}=${VAL}"
        rc=2
    else
        printf '\n# issue #38296 repro: keep the worker from cutting at 10 MiB\n%s=%s\n' "$KEY" "$VAL" >> "$OPTS"
        say FIXED "internal option appended -> ${KEY}=${VAL}"
        rc=2
    fi
else
    say SKIP "$OPTS not found (manager not installed)"
    rc=1
fi

# 2 -- clean reconciliation baseline. Deterministic _id means a later pass HEALS an
#      earlier loss, so the index must be empty before every pass.
code=$(curl -sk -u "$ES_AUTH" -o /tmp/.prep_wipe.json -w '%{http_code}' \
        -XPOST "$ES/$IDX/_delete_by_query?refresh=true&conflicts=proceed" \
        -H 'Content-Type: application/json' -d '{"query":{"match_all":{}}}' --max-time 60)
if [ "$code" = 200 ]; then
    deleted=$(python3 -c 'import json;print(json.load(open("/tmp/.prep_wipe.json"))["deleted"])' 2>/dev/null || echo '?')
    say OK "$IDX wiped ($deleted docs deleted)"
else
    say FAIL "could not wipe $IDX (HTTP $code) -- the reconciliation will be meaningless"
    rc=1
fi

# 3 -- fresh injector log, so the fault count belongs to this pass alone.
if [ -f "$ACCESS_LOG" ]; then
    cp "$ACCESS_LOG" "${ACCESS_LOG%.log}-$(date +%H%M%S).prev.log"
    : > "$ACCESS_LOG"
    say OK "injector access log rotated"
fi

# 4 -- the injector must be listening AND reaching the real indexer behind it.
inj_code=$(curl -s -o /dev/null -w '%{http_code}' "$INJECTOR/_cat/health" --max-time 5)
if [ "$inj_code" = 000 ]; then
    say FAIL "nothing listening on ${INJECTOR#http://} -- start nginx with faultinject-nginx.conf"
    rc=1
else
    say OK "injector answering (HTTP $inj_code) and proxying to the indexer"
fi

echo
[ "$rc" = 2 ] && echo ">>> RESTART THE MANAGER before running the sender. <<<" && echo
echo "Then:  ./run_benchmark.sh --scenario scenarios/fault_injection_fim.json --mode uds --no-monitor --no-charts"
echo "And:   $(cd "$(dirname "$0")" && pwd)/prep-repro-run.sh verify"
exit $rc
```

**`tools/manager_benchmark/scenarios/fault_injection_fim.json`** — `repeat_until` raised from `"0"`
to `"120s"` so the shard queues stay non-empty and the timer has open batches to snapshot.

```json
{
  "name": "fault_injection_fim",
  "description": "128 agents replaying a FIM full_resync of exactly 800 docs each for 120s. Reconciliation unit for flush-loss detection: repeat_until keeps the shard queues non-empty so the connector flush timer has open batches to snapshot. Note each replay re-runs the full_resync Cleans, which wipes that agent docs first -- the end-state reconciliation therefore judges the LAST pass only; use the timer_discards log counter for the cumulative signal.",
  "mode": "uds",
  "defaults": { "module": "fim", "option": "Sync", "cluster_name": "cluster01",
                "retry": { "enabled": false } },
  "lanes": { "fim": [ { "kind": "full_resync", "module": "fim", "indices": ["wazuh-states-fim-files"],
                        "documents": { "count": 800, "size_bytes": 512 } } ] },
  "fleets": [ { "name": "linux", "agents": 128, "first_id": 10000, "lanes": ["fim"],
      "start": { "osname": "Ubuntu", "osplatform": "ubuntu", "ostype": "linux", "osversion": "22.04",
                 "architecture": "x86_64", "agentversion": "5.0.0" } } ],
  "pacing": { "concurrent_agents": 0, "requests_per_second": 0, "repeat_until": "120s", "drain_timeout": "15s" }
}
```

</p>
</details>

<details><summary>Concurrency validation — conservation stress tests, full unit suite, TSAN run</summary>
<p>

Four new multi-threaded tests race stagers (`scopeLock()` + `bulkIndex`, the documented
contract), dedicated flushers and the background timer at its smallest scale (1 s) against ONE
connector. The oracle is **conservation**, not survival: a mocked `post()` records every `_bulk`
body, and the test asserts every staged id arrives in a well-formed NDJSON frame **exactly
once** — no loss, no duplicate, no torn frame.

```text
[ RUN      ] IndexerConnectorSyncTest.ConcurrentStagersAndFlushersDeliverEveryDocumentExactlyOnce
[       OK ] IndexerConnectorSyncTest.ConcurrentStagersAndFlushersDeliverEveryDocumentExactlyOnce (28 ms)
[ RUN      ] IndexerConnectorSyncTest.TimerFlushersAndStagersRacingDeliverEveryDocumentExactlyOnce
[       OK ] IndexerConnectorSyncTest.TimerFlushersAndStagersRacingDeliverEveryDocumentExactlyOnce (3145 ms)
[ RUN      ] IndexerConnectorSyncTest.TransportRetriesUnderConcurrentStagingLoseNothing
[       OK ] IndexerConnectorSyncTest.TransportRetriesUnderConcurrentStagingLoseNothing (2728 ms)
[ RUN      ] IndexerConnectorSyncTest.ReducedScaleTimerSoakDeliversEveryDocumentExactlyOnce
[       OK ] IndexerConnectorSyncTest.ReducedScaleTimerSoakDeliversEveryDocumentExactlyOnce (10107 ms)
[==========] 4 tests from 1 test suite ran. (16011 ms total)
[  PASSED  ] 4 tests.
```

Full `indexer_connector_utest` suite on the same build:

```text
[==========] 272 tests from 10 test suites ran. (48189 ms total)
[  PASSED  ] 272 tests.
```

The same four tests under `TSANITIZE=YES` (`-fsanitize=thread`, coverage off) — **zero
ThreadSanitizer warnings**:

```text
$ setarch $(uname -m) -R bin/indexer_connector_utest --gtest_filter='<the four tests>'
[==========] 4 tests from 1 test suite ran. (16630 ms total)
[  PASSED  ] 4 tests.
$ grep -c "WARNING: ThreadSanitizer" tsan-run.log
0
```

Two operational notes for reproducing: (1) on kernel 6.8 TSAN aborts with `unexpected memory
mapping` unless ASLR is disabled for the process (`setarch $(uname -m) -R`); (2) before the
coverage cut, the identical run reported 26,170 "races", every one on a `__gcov0.*` coverage
counter — that is what the `TSANITIZE`-drops-coverage change removes.

</p>
</details>

  - [ ] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

### Artifacts Affected

- `wazuh-manager-modulesd` — `inventory_sync_server` no longer starts a flush-timer thread per
  pipeline connector, and builds its indexer connector configs on every start attempt.
- `wazuh-manager-analysisd` — links the same `indexer_connector` shared module, so it picks up the
  413-split buffer fix and the `flush_interval_seconds == 0` semantics. Its own connectors keep
  their configured interval; behaviour is unchanged unless a `0` was configured, which previously
  produced a busy loop.
- No packaging, default configuration file, or on-disk format is touched.

### Configuration Changes

No new, removed or renamed setting.

One **semantic** change to an existing value: `flush_interval_seconds = 0` now means *no background
flush thread*. It previously created a thread that spun on a zero-length wait. Any deployment that
had configured `0` was burning a core; it now gets the disable the value reads as. All non-zero
values behave exactly as before.

Note for reviewers: `inventory_sync_server_indexer_sync_flush_interval_seconds` remains parsed and
visible in `GET /config`, but the facade overwrites it for the pipeline connectors — it has never
had an effect on them, before or after this PR.

### Tests Introduced

Unit tests, GoogleTest, four commits paired one-to-one with the four fixes:

- `HandleError413FailedSplitLeavesAnEmptyBuffer` — drives a `413` followed by a `500` on the split,
  then asserts the second `flush()` issues **no** request: a failed split must not leave bytes for
  the next flush to resend.
- `ZeroFlushIntervalStartsNoBackgroundThread` — counts entries in `/proc/self/task` around the
  connector's construction and pins that a zero interval starts no thread. This fails on the base
  build, where `0` produced a spinning thread.
- A test pinning the no-timer flush ownership contract at both the connector and the module level.
- A test pinning that a lane-only retry receives a usable connector configuration.

A fifth commit adds the concurrency coverage the fixes above were reviewed without — four
multi-threaded tests against one connector, all judged by a **conservation oracle** (every staged
document arrives in a well-formed `_bulk` NDJSON frame exactly once; loss, duplication and torn
frames all fail the test):

- `ConcurrentStagersAndFlushersDeliverEveryDocumentExactlyOnce` — 4 stagers under `scopeLock()`
  race 2 dedicated flushers plus the 1 KB connector's inline size-triggered flushes.
- `TimerFlushersAndStagersRacingDeliverEveryDocumentExactlyOnce` — adds the third flush owner,
  the background timer at its smallest configurable scale (1 s), for ~3 periods mid-staging.
- `TransportRetriesUnderConcurrentStagingLoseNothing` — transport failures (`statusCode -1`) hit
  mid-race; the retried frame must be delivered exactly once, nothing staged around it lost.
- `ReducedScaleTimerSoakDeliversEveryDocumentExactlyOnce` — the 1 s timer as the only flusher
  over an extended period (10 s default, `INDEXER_CONNECTOR_SOAK_SECONDS` extends it for longer
  soaks), with thread-count stability pinned via `/proc/self/task`.

All four also run clean under the new `TSANITIZE=YES` ThreadSanitizer build (evidence above).

The live fault-injection procedure and its harness are documented outside the repo (see the
harness section above) and are reusable for the Phase B verification run.

## Review Checklist

- [ ] Code changes reviewed
- [ ] Phase B live run (fixed build, same injector load) attached

### DIFF
--- a/docs/ref/modules/indexer_connector/README.md
+++ b/docs/ref/modules/indexer_connector/README.md
@@ -33,6 +33,7 @@ The library provides two classes depending on the use case:
 
 - Buffer up to 10 MB of serialized events before flushing (configurable: `wazuh_modules.indexer_bulk_size_bytes` for Vulnerability Scanner, `wazuh_modules.inventory_sync_server_indexer_sync_max_bulk_size` for Inventory Sync Server).
 - Flush automatically after 20 seconds of inactivity (configurable: `wazuh_modules.indexer_flush_interval` for Vulnerability Scanner; the Inventory Sync Server deliberately overrides its periodic flush — its ingestion workers own every flush, see its [configuration reference](../inventory-sync-server/configuration.md)).
+- `flush_interval_seconds = 0` means **no background flush thread at all**: the connector is never created with one and every flush is the caller's. Use it when the caller has to answer for a failed flush — a timer flush that fails discards the staging buffer and has no caller to report to, so a later `flush()` finds an empty buffer and returns success for data that never landed. The value is set directly by the Inventory Sync Server; it is not reachable through the Vulnerability Scanner's `wazuh_modules.indexer_flush_interval`, whose range is 1–3600.
 - If the indexer returns HTTP 413 (payload too large), the batch is split and retried.
 - Version conflicts at the document level are handled per-document.
 
--- a/docs/ref/modules/inventory-sync-server/architecture.md
+++ b/docs/ref/modules/inventory-sync-server/architecture.md
@@ -133,6 +133,42 @@ batch; the status is picked by connector availability — `503` when the indexer
 the log line is operator-actionable). Re-applying a session is idempotent (deterministic document
 ids, versioned upserts), which is what makes "just re-POST it" the whole recovery story.
 
+Because each worker owns its connector outright, all cross-thread contention in the ingestion
+path funnels into one mutex per connector with a single legitimate owner. Two ordering rules keep
+it acyclic, and both are enforced by scoping rather than by convention: the shard lock is always
+released before any flush (it lives in its own block inside the worker loop), and the connector's
+retry mutex — which exists only so the backoff can sleep without self-deadlocking on the staging
+mutex — is only ever taken from inside the staging one.
+
+```mermaid
+flowchart LR
+    subgraph TH["Threads"]
+        TR["connection strand"]
+        PW["pipeline worker i"]
+        LW["scan lane worker"]
+        HM["indexer health monitor"]
+    end
+    subgraph MX["Mutexes"]
+        SM["shard[i] mutex<br/>one FIFO deque"]
+        RG["agent in-flight registry"]
+        CM["connector[i] staging mutex<br/>held across the bulk request"]
+        RT["connector[i] retry mutex<br/>backoff sleep only"]
+    end
+    LF["atomic host-up flags<br/>+ atomic round-robin cursor"]
+    TR -->|enqueue| SM
+    TR -.->|isAvailable| LF
+    HM -.->|writes| LF
+    PW -->|own shard only| SM
+    SM -.->|released BEFORE any flush| CM
+    PW ==>|stage / flush| CM
+    LW --> RG
+    LW ==>|stage / flush| CM
+    CM ==>|inside the bulk request| RT
+```
+
+The availability verdict behind the `503` admission gate is read lock-free, so a worker blocked on
+the indexer never delays the strand that is deciding whether to admit the next session.
+
 ## The vulnerability-detection scan lane
 
 Sessions whose `Start.option` is `VDFirst` or `VDSync` carry data the vulnerability scanner must
--- a/docs/ref/modules/inventory-sync-server/configuration.md
+++ b/docs/ref/modules/inventory-sync-server/configuration.md
@@ -384,8 +384,9 @@ wazuh_modules.inventory_sync_server_indexer_sync_max_bulk_size=10485760
 #### wazuh_modules.inventory_sync_server_indexer_sync_flush_interval_seconds
 
 **No effect.** The value is accepted for compatibility but the module overrides the connector's
-periodic flush to one hour regardless: the ingestion workers own every flush (a timer-driven flush
-that fails discards the buffer silently, which would let a worker answer `200` for lost data).
+periodic flush to `0` regardless, which means the connector never starts its background flush
+thread: the ingestion workers own every flush (a timer-driven flush that fails discards the buffer
+silently and has no responder to report to, which would let a worker answer `200` for lost data).
 
 ```ini
 wazuh_modules.inventory_sync_server_indexer_sync_flush_interval_seconds=20
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -51,10 +51,19 @@ if(FSANITIZE)
   set(CMAKE_CXX_FLAGS_DEBUG "-g -fsanitize=address,leak,undefined")
 endif(FSANITIZE)
 
+if(TSANITIZE)
+  if(FSANITIZE)
+    message(FATAL_ERROR "TSANITIZE and FSANITIZE are mutually exclusive")
+  endif()
+  set(CMAKE_CXX_FLAGS_DEBUG "-g -fsanitize=thread")
+endif(TSANITIZE)
+
 # Function to link coverage libraries for unit tests Uses plain signature for
 # compatibility with existing plain signature usage
 function(add_coverage_libraries target_name)
-  if(UNIT_TEST)
+  # gcov counters are non-atomic globals: under TSAN every increment reports as a
+  # data race, burying real findings, so TSANITIZE builds drop coverage.
+  if(UNIT_TEST AND NOT TSANITIZE)
     if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
       target_compile_options(${target_name} PUBLIC --coverage)
       target_link_libraries(${target_name} PUBLIC -fprofile-arcs)
--- a/src/Makefile
+++ b/src/Makefile
@@ -114,6 +114,9 @@ endif
 ifneq (,$(filter ${FSANITIZE},YES yes y Y 1))
 CMAKE_OPTS+=-DFSANITIZE=ON
 endif
+ifneq (,$(filter ${TSANITIZE},YES yes y Y 1))
+CMAKE_OPTS+=-DTSANITIZE=ON
+endif
 ifneq (,$(filter ${DISABLE_JEMALLOC},YES yes y Y 1))
 CMAKE_OPTS+=-DDISABLE_JEMALLOC=ON
 endif
--- a/src/shared_modules/indexer_connector/include/indexerConnector.hpp
+++ b/src/shared_modules/indexer_connector/include/indexerConnector.hpp
@@ -167,7 +167,8 @@ public:
      *
      * @param config Indexer configuration, including database_path and servers.
      *               `monitoring_interval_seconds` (default 10, minimum 1) sets the polling period of
-     *               the health monitor this constructor builds.
+     *               the health monitor this constructor builds. `flush_interval_seconds` = 0 starts
+     *               NO background flush thread at all: the caller owns every flush().
      * @param logging Logging context pairing the caller module name and the log callback.
      *                The caller name is used to build the log tag as
      *                "<callerName>(indexer-connector)" (e.g. "vulnerability-scanner(indexer-connector)").
@@ -185,7 +186,9 @@ public:
      *
      * @param config Indexer configuration. Still supplies this connector's own tunables
      *               (`max_bulk_size`, `flush_interval_seconds`, `max_retry_delay_seconds`,
-     *               `request_timeout_seconds`). `monitoring_interval_seconds` is IGNORED here --
+     *               `request_timeout_seconds`). `flush_interval_seconds` = 0 starts NO background
+     *               flush thread at all: the caller owns every flush(). `monitoring_interval_seconds`
+     *               is IGNORED here --
      *               the shared session's monitor was already built with the session's own value --
      *               just like the `ssl.*` and credential keys, which the session also supplies. Its
      *               `hosts` list MUST equal the session's: the monitor only knows the hosts it was
--- a/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
+++ b/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
@@ -359,11 +359,19 @@ class IndexerConnectorSyncImpl final
 
         auto serverUrl = m_selector->getNext();
 
+        // No success callback below may throw: THttpRequest::post() invokes them inside its own
+        // try block and remaps anything thrown there to onError(statusCode = -1), which the error
+        // handlers read as a transport outage to retry -- resending a buffer that no longer
+        // matches what the callback saw. Callbacks report through these flags; the throw happens
+        // after post() returns, from this frame.
+        bool deleteByQueryLeftDocuments = false;
+        bool indexingFailures = false;
+
         // A _delete_by_query answers 200 even when it deleted nothing it was asked to: per-shard
         // errors come back in a `failures` array inside the body. Reporting that run as success is
         // how documents survive a deletion that everyone upstream believes worked, so the failures
         // are raised like a transport error and the caller decides whether to retry.
-        const auto onSuccessDeleteByQuery = [this](const std::string& response)
+        const auto onSuccessDeleteByQuery = [this, &deleteByQueryLeftDocuments](const std::string& response)
         {
             LOGFN_DEBUG2(m_logFn, "Response: %s", response.c_str());
 
@@ -399,7 +407,7 @@ class IndexerConnectorSyncImpl final
                        failures,
                        conflicts,
                        response.c_str());
-            throw IndexerConnectorException("deleteByQuery did not delete every matching document");
+            deleteByQueryLeftDocuments = true;
         };
 
         const auto onErrorDeleteByQuery =
@@ -454,6 +462,10 @@ class IndexerConnectorSyncImpl final
                         .url = HttpURL(url), .data = query.dump(), .secureCommunication = m_secureCommunication},
                     PostRequestParameters {.onSuccess = onSuccessDeleteByQuery, .onError = onErrorDeleteByQuery},
                     ConfigurationParameters {.timeout = m_requestTimeoutMs});
+                if (deleteByQueryLeftDocuments)
+                {
+                    throw IndexerConnectorException("deleteByQuery did not delete every matching document");
+                }
             }
             catch (...)
             {
@@ -467,17 +479,15 @@ class IndexerConnectorSyncImpl final
             }
         }
 
-        const auto onSuccess = [this, &needToRetry](const std::string& response)
+        const auto onSuccess = [this, &needToRetry, &indexingFailures](const std::string& response)
         {
             LOGFN_DEBUG2(m_logFn, "Response: %s", response.c_str());
 
             // Validate bulk response at document level
             if (!validateBulkResponse(response, m_logFn.c_str()))
             {
-                m_bulkData.clear();
-                m_boundaries.clear();
-                m_lastBulkTime = std::chrono::steady_clock::now();
-                throw IndexerConnectorException("Bulk operation had indexing failures");
+                indexingFailures = true;
+                return;
             }
 
             m_shouldNotifyAfterBulk = true;
@@ -553,6 +563,7 @@ class IndexerConnectorSyncImpl final
                     return;
                 }
                 needToRetry = false;
+                indexingFailures = false;
 
                 std::string url;
                 url += m_selector->getNext();
@@ -565,6 +576,13 @@ class IndexerConnectorSyncImpl final
                                                        .secureCommunication = m_secureCommunication},
                                     PostRequestParameters {.onSuccess = onSuccess, .onError = onError},
                                     ConfigurationParameters {.timeout = m_requestTimeoutMs});
+                if (indexingFailures)
+                {
+                    m_bulkData.clear();
+                    m_boundaries.clear();
+                    m_lastBulkTime = std::chrono::steady_clock::now();
+                    throw IndexerConnectorException("Bulk operation had indexing failures");
+                }
                 if (needToRetry)
                 {
                     const auto retryDelay = retryBackoff.nextDelay();
@@ -652,15 +670,19 @@ class IndexerConnectorSyncImpl final
     void processBulkChunk(std::string_view data, const std::span<size_t>& boundaries)
     {
         bool needToRetry = false;
+        // Reported instead of thrown: see processBulk() on why success callbacks must not throw.
+        // Throwing here additionally livelocked -- the remapped onError(-1) retried the same
+        // chunk, whose 200 answer carried the same failing items, forever.
+        bool indexingFailures = false;
 
-        const auto onSuccess = [this](const std::string& response)
+        const auto onSuccess = [this, &indexingFailures](const std::string& response)
         {
             LOGFN_DEBUG2(m_logFn, "Chunk processed successfully: %s", response.c_str());
 
             // Validate bulk response at document level
             if (!validateBulkResponse(response, m_logFn.c_str()))
             {
-                throw IndexerConnectorException("Bulk chunk operation had indexing failures");
+                indexingFailures = true;
             }
         };
         const auto onError = [this, &needToRetry, boundaries](
@@ -728,6 +750,7 @@ class IndexerConnectorSyncImpl final
                 return;
             }
             needToRetry = false;
+            indexingFailures = false;
             // Resolved inside the loop so a retry rotates to the next healthy host -- and throws
             // (batch retained, resent by a later flush) once the monitor marks every host down,
             // instead of hammering the same dead host forever while holding m_mutex.
@@ -740,6 +763,10 @@ class IndexerConnectorSyncImpl final
                                                              .secureCommunication = m_secureCommunication},
                                 PostRequestParameters {.onSuccess = onSuccess, .onError = onError},
                                 ConfigurationParameters {.timeout = m_requestTimeoutMs});
+            if (indexingFailures)
+            {
+                throw IndexerConnectorException("Bulk chunk operation had indexing failures");
+            }
             if (needToRetry)
             {
                 const auto retryDelay = retryBackoff.nextDelay();
--- a/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
+++ b/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
@@ -848,6 +848,12 @@ public:
         }
 
         m_lastBulkTime = std::chrono::steady_clock::now();
+
+        // 0 = no background flush thread; the caller owns every flush.
+        if (m_flushInterval == 0)
+        {
+            return;
+        }
         m_bulkThread = std::thread(
             [this]()
             {
--- a/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
+++ b/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
@@ -10,6 +10,7 @@
  */
 
 #include "IURLRequest.hpp"
+#include "defer.hpp"
 #include "exponentialBackoff.hpp"
 #include "external/nlohmann/json.hpp"
 #include "indexerConnector.hpp"
@@ -588,6 +589,16 @@ class IndexerConnectorSyncImpl final
 
     void splitAndProcessBulk()
     {
+        // Clear on every exit path: on failure the caller re-stages, so bytes kept here would be
+        // re-sent by whoever flushes next.
+        DEFER(
+            [this]()
+            {
+                m_bulkData.clear();
+                m_boundaries.clear();
+                m_lastBulkTime = std::chrono::steady_clock::now();
+            });
+
         const size_t totalOperations = m_boundaries.size();
         if (totalOperations <= 1)
         {
@@ -636,9 +647,6 @@ class IndexerConnectorSyncImpl final
         {
             m_shouldNotifyAfterBulk = true;
         }
-        m_bulkData.clear();
-        m_boundaries.clear();
-        m_lastBulkTime = std::chrono::steady_clock::now();
     }
 
     void processBulkChunk(std::string_view data, const std::span<size_t>& boundaries)
--- a/src/shared_modules/indexer_connector/tests/unit/indexerConnectorSync_test.cpp
+++ b/src/shared_modules/indexer_connector/tests/unit/indexerConnectorSync_test.cpp
@@ -5394,3 +5394,132 @@ TEST_F(IndexerConnectorSyncTest, ReducedScaleTimerSoakDeliversEveryDocumentExact
     ASSERT_FALSE(expected.empty()) << "the stagers never staged anything; the soak never exercised the timer";
     expectExactlyOnce(recorder.deliveredIds(), expected);
 }
+
+// ============================================================================
+// A 200 _bulk answer whose items report failures. HTTPRequest::post() runs
+// onSuccess INSIDE its try block and remaps anything thrown there to
+// onError(statusCode = -1); the mocks below reproduce that remapping because
+// it is the production semantics the connector's callbacks live under -- a
+// mock that lets the exception fly cannot see this bug class.
+// ============================================================================
+
+namespace
+{
+    /// Calls onSuccess(body) the way the real HTTPRequest::post() does: inside a try whose
+    /// generic catch remaps the exception to onError(what, -1, ...).
+    void respondLikeHttpRequestPost(const PostRequestParametersVariant& postParams, const std::string& body)
+    {
+        try
+        {
+            if (std::holds_alternative<TPostRequestParameters<const std::string&>>(postParams))
+            {
+                std::get<TPostRequestParameters<const std::string&>>(postParams).onSuccess(body);
+            }
+            else
+            {
+                std::get<TPostRequestParameters<std::string&&>>(postParams).onSuccess(std::string {body});
+            }
+        }
+        catch (const std::exception& e)
+        {
+            respondBulkError(postParams, e.what(), -1);
+        }
+    }
+
+    constexpr auto BULK_200_WITH_ITEM_ERRORS =
+        R"({"took":1,"errors":true,"items":[{"index":{"_index":"test_index","_id":"id1","status":500,)"
+        R"("error":{"type":"mapper_parsing_exception"}}}]})";
+} // namespace
+
+/// A 200 whose items failed used to be reported by THROWING out of onSuccess AFTER clearing
+/// m_bulkData; the remapped onError(-1) then took the transport branch and retried -- resending
+/// an emptied buffer. The failure must surface once, with exactly one POST.
+TEST_F(IndexerConnectorSyncTest, BulkItemFailuresSurfaceOnceAndNeverRetryAnEmptiedBuffer)
+{
+    auto mockSelector = std::make_unique<NiceMock<MockServerSelector>>();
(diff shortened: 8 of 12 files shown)


NahuFigueroa97 (changes requested):
The new tests are single-threaded (checking buffer status, number of threads via /proc, and configuration distribution). There are no tests that trigger BulkIndex/flush from multiple threads simultaneously against the same connector, nor any TSAN-type tests. There is also no test of a timer running at a reduced scale for an extended period.

NahuFigueroa97 (changes requested):
While this is not a change in this pull request, it's code that was already there, in `processBulk()`, the sister function of `splitAndProcessBulk()` (which you did modify). I saw it by reading the entire file, not because your change affects or triggers it. I'm sharing it because it's the same type of problem you fixed in `splitAndProcessBulk`.

The sequence, step by step:

A `_bulk` request is sent to the indexer. The indexer responds with `200 OK`, but within that response are documents that failed individually (this happens: a `_bulk` request can be `200` at the HTTP level and still return errors per item).
`validateBulkResponse()` detects these actual failures. The `onSuccess` callback reacts: it clears `m_bulkData` (assuming it's no longer needed) and then throws an exception to report the failures.

`onSuccess` That exception escapes onSuccess, but doesn't go directly to processBulk() — it's caught by the generic catch block of HTTPRequest::post(), which reinterprets it as a transport failure (connection, timeout, etc.) and calls onError with statusCode = -1.
processBulk() sees that statusCode <= 0 and enters its normal retry logic: "this didn't arrive, retry with the same data."

The problem: the "same data" (m_bulkData) was already emptied in step 2. The retry sends an empty _bulk.

If the indexer responds with 2xx to an empty _bulk (I couldn't confirm this, it depends on the version/config), the retry loop ends normally and flush() doesn't return anything — the caller thinks everything went well, when in reality the original batch DID have indexing failures that were never reported.

matigarciadev (approved):
### LGTM!

**Built and Install:**
```console
General settings:
    TARGET:             manager
    V:                  
    DEBUG:              
    EXTERNAL_SRC_ONLY:  
User settings:
USE settings:
    USE_INOTIFY:        no
    USE_BIG_ENDIAN:     no
    USE_SELINUX:        no
    USE_AUDIT:          no
    IMAGE_TRUST_CHECKS: 1
Compiler:
    MAKE              make

Done building manager

Wait for success...
success
Processing ./.
  Created wheel for wazuh: filename=wazuh-5.0.0-py3-none-any.whl size=273613 sha256=7d076c12f681132ed74d613b2eb994504b310569c521c4daa47d65e23729531d
  Stored in directory: /tmp/pip-ephem-wheel-cache-61z_vg08/wheels/50/e9/6c/171615ee3709da12266524fed91cb0863278bc995eed9295be
Successfully built wazuh
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the --root-user-action optionif you know what you are doing and want to suppress this warning.

Processing ./.
  Created wheel for api: filename=api-5.0.0-py3-none-any.whl size=97172 sha256=bccfafbc628886226c4b745dc448852f53265bd38372a4f7d15aad1773a7f78c
  Stored in directory: /tmp/pip-ephem-wheel-cache-h1jwkg41/wheels/12/3e/cd/fe7f9fbf555382f244bd9d9630026d54f4ffbcb658cc4f4190
Successfully built api
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the --root-user-action optionif you know what you are doing and want to suppress this warning.

Generating schema files...
Loading resources...
Loading wcs_flat.yml...
Successfully merged 1 files into temporary file: /tmp/merged_wcs_u7ig4k_l.yml
Loading schema template...
Success.
Success.
Success.
Generating engine schema...
Success.
Success.
Success.
Cleaned up temporary file: /tmp/merged_wcs_u7ig4k_l.yml
Generated wazuh-decoders.json
Success.
Copying store files...
Installing GeoIP databases...


Starting Wazuh...
manager
2026/08/31 20:37:46 wazuh-manager-modulesd: ERROR: File 'etc/certs/root-ca.pem' not found for 'indexer.ssl.certificate_authorities' in module 'indexer'. Check configuration
2026/08/31 20:37:46 wazuh-manager-modulesd: ERROR: (1202): Configuration error at 'etc/wazuh-manager.conf'.
wazuh-manager-modulesd: Configuration error. Exiting

 - To start Wazuh:

 - To stop Wazuh:

   More information at: 
```

**Manager Status:**
```console
wazuh-manager-monitord not running...
wazuh-manager-db not running...
wazuh-manager-authd not running...
wazuh-manager-apid not running...
Wazuh v5.0.0 Stopped
Starting Wazuh v5.0.0...
Started wazuh-manager-apid...
Started wazuh-manager-db...
Started wazuh-manager-monitord...
Completed.

wazuh-manager-monitord is running...
```

**Tests:**

**Run `indexer_connector_utest`, full suite with TSAN:**
```console
setarch $(uname -m) -R env TSAN_OPTIONS="halt_on_error=0" \
  ./bin/indexer_connector_utest


==================
WARNING: ThreadSanitizer: data race (pid=642805)
  Write of size 4 at 0x725800000268 by thread T96:
    #0 operator()<...> indexerConnectorAsync_test.cpp:824
    ... (IndexerBulkQueue::dispatch() background worker thread)

  Previous read of size 4 at 0x725800000268 by main thread:
    #0 testing::internal::CmpHelperGE<int,int> gtest.h:1630
    #1 IndexerConnectorAsyncTest_HandleError429TooManyRequests_Test::TestBody() indexerConnectorAsync_test.cpp:885

  Location is heap block of size 648 allocated by main thread
  Thread T96 created by main thread via pthread_create

SUMMARY: ThreadSanitizer: data race indexerConnectorAsync_test.cpp:824 in operator()<...>
==================
[       OK ] IndexerConnectorAsyncTest.HandleError429TooManyRequests (1224 ms)

==================
WARNING: ThreadSanitizer: data race (pid=642805)
  Read of size 4 at 0x725800000268 by main thread:
    #0 testing::internal::CmpHelperGT<int,int> gtest.h:1632
    (in IndexerConnectorAsyncTest.SmallBulkSizeTriggersAsyncProcessing)

  Previous write of size 4 at 0x725800000268 by thread T106:
    #0 IndexerConnectorAsyncTest::simulateSuccessfulPost(...)
    #1 operator()<...> (same dispatch-thread callback path)

SUMMARY: ThreadSanitizer: data race gtest.h:1632 in testing::internal::CmpHelperGT<int, int>
==================

[==========] 274 tests from 10 test suites ran. (49623 ms total)
[  PASSED  ] 274 tests.
ThreadSanitizer: reported 2 warnings
```

- **Output:** `274 tests from 10 test suites ran. PASSED 274 tests.`
- **Exit code:** 66
- **Result:** `ThreadSanitizer: reported 2 warnings`

**Run `indexer_connector_utest`, filtered to the PR's new tests only:**
```bash
setarch $(uname -m) -R env TSAN_OPTIONS="halt_on_error=0" \
  ./bin/indexer_connector_utest --gtest_filter="*ZeroFlushInterval*:*NonZeroFlushInterval*:*Soak*:*ConcurrentStagers*:*ChunkItemFailures*:*IndexerConnectorSync*"
```
- **Output:** `150 tests from 2 test suites ran. PASSED 150 tests.`
- **Exit code:** 0
- **Result:** no TSAN warnings

**Run `inventory_sync_server_utest`, full suite:**
```bash
setarch $(uname -m) -R env TSAN_OPTIONS="halt_on_error=0" \
  ./wazuh_modules/inventory_sync_server/test/inventory_sync_server_utest
```
- **Output:** `230 tests from 20 test suites ran. PASSED 230 tests.`
- **Exit code:** 0
- **Result:** no TSAN warnings

| Scope | Tests | Result | TSAN warnings |
|---|---|---|---|
| `inventory_sync_server_utest` (full) | 230/230 | pass | 0 |
| `indexer_connector_utest` (full suite) | 274/274 gtest assertions pass | process exits 66 | **2 data races** |
| `indexer_connector_utest` (PR's new `IndexerConnectorSync` tests only) | 150/150 | pass | 0 |

NahuFigueroa97 (approved):
LGTM!

```
Installing GeoIP databases...


Starting Wazuh...
manager
Starting Wazuh v5.0.0...
Started wazuh-manager-apid...
Started wazuh-manager-db...
Started wazuh-manager-monitord...
Completed.

 - To start Wazuh:

 - To stop Wazuh:

   More information at: 
```

juliancnn (approved):
LGTM!

Issue #38357, opened 2026-08-13
Vulnerability scanner cannot be forced to re-evaluate an unchanged agent — checksum-delta gating with no working rescan path (v4.14.6)


pradeksoa:
# Vulnerability scanner cannot be forced to re-evaluate an unchanged agent — checksum-delta gating with no working rescan path (single-node, v4.14.6)

## Environment

- Wazuh version: **v4.14.6** (rc2), manager + indexer + dashboard on a single node
- OS (manager): Ubuntu 24.04. Indexer: single-node OpenSearch (not clustered)
- Affected agents: **Linux** agents (Ubuntu 24.04 LTS / 20.04 LTS) across two cloud providers (Alibaba Cloud ECS, Apache CloudStack/KVM); same class of failure reproduced on **Windows** agents (Windows Server 2022, 12 Aug 2026)
- Indices affected: `wazuh-states-vulnerabilities-*`

## Summary

For an agent whose installed-package content is legitimate and **unchanged**, the Wazuh vulnerability scanner cannot be made to fully re-evaluate that agent's inventory against the CTI feed by **any** of the standard reset/remediation procedures. The index silently keeps a stale (missing or partial) snapshot of that agent's vulnerabilities indefinitely.

Two independent reset directions were tested, weeks apart, both failed with distinct and complementary mechanisms:

| Direction | What was reset | Result | Why it failed |
|---|---|---|---|
| Manager-side (21 Jul 2026) | agent's wazuh-db file `/var/ossec/queue/db/<id>.db` + vulnerability-scanner RocksDB dirs `/var/ossec/queue/vd/{inventory,delayed,event,state_track}` (+ manager restart) | `sys_programs` rebuilt 100% accurate; vulnerability index **unchanged** | correlation state is separate from both stores; downstream behavior gated by an agent-side VDFIRST flag mechanism that does not exist in the shipped binary |
| Agent-side (13 Aug 2026) | agent's own syscollector DB `/var/ossec/queue/syscollector/db/local.db` + `wazuh-agent` restart | local rescan succeeded, wazuh-db `last_attempt` advanced, checksum **identical**, no `PackageInsert` events emitted, index **unchanged** | dbsync checksum is computed purely from package content (name/version); identical content → manager treats as no-op → no delta events → scanner never notified |

## Steps to Reproduce (13 Aug 2026, clean single-agent reproduction)

[...]

Issue #38272, opened 2026-08-10
Vulnerability scanner: CTI delta update corrupts local content DB — post-update re-scan silently deletes all findings (4.14.7)


Slay3r00:
---

## Bug description

On a fresh Wazuh 4.14.7 all-in-one install, the vulnerability scanner produces correct findings after ingesting the CTI **snapshot**, but the first **delta/changes** update applied on top leaves the content database in a state where every package candidate lookup returns empty — with no errors logged at any level. The automatic "re-scan after content update" then reconciles against this empty result set and **deletes every finding from `wazuh-states-vulnerabilities-*`**.

Because the CTI snapshot always lags `last_offset`, the catch-up delta runs within minutes of any fresh ingestion, so on affected installs the vulnerability inventory is empty essentially permanently. The offset is recorded as current, so subsequent hourly update checks are no-ops and never re-trigger anything. The result is a deployment where vulnerability detection appears healthy (module started, feed offset current, indexer connector green, zero errors) but has zero findings forever.

## Environment

- Wazuh 4.14.7, single-node all-in-one (manager + indexer + dashboard), installed via `wazuh-install.sh -a -i`
- Manager host: Ubuntu 24.04
- 4 agents: 1× Ubuntu 20.04.6, 3× Ubuntu 24.04.4
- Default `<vulnerability-detection>` block (enabled, index-status yes, feed-update-interval 60m)
- CTI consumer (default): `vd_1.0.0/vd_4.8.0`; snapshot offset 3666504 (2026-07-28), last_offset 3706504

## Reproduction (performed twice, identical result)

1. Stop manager; `mv /var/ossec/queue/vd /var/ossec/queue/vd.bak`; start manager.
2. Content manager downloads snapshot `3666504_1785234254.zip`, ingests it (~12 min).
3. Re-scan triggers; **18,591 findings** are indexed across the 4 agents (verified: `_stats/indexing` shows `index_total: 18591`). Severity spread: 382 Critical / 5,908 High / 9,922 Medium / 340 Low.
4. ~8 minutes later the automatic delta catch-up runs (offset 3666504 → 3706504) and completes without error:
   ```
   INFO: Initiating update feed process.
   INFO: Triggered a re-scan after content update.
   INFO: Feed update process completed.
   ```
5. The post-update re-scan walks all 3,247 packages; with `wazuh_modules.debug=2` every package shows the same pattern — scan opens and completes instantly with no candidate evaluation between:
   ```
[...]

PR #38623, opened 2026-08-26
Recover a re-enrolled agent's data instead of waiting for a checksum mismatch

### LINKED ISSUE #38601: An agent that re-enrolls never rebuilds its inventory: 1212 documents become 1, and every sync reports success
|---|---|---|---|---|
| 5.0.0 | wazuh-agent (`shared_modules/sync_protocol`, FIM, Syscollector, SCA) | Agent | Docker, `public.ecr.aws/wazuh-cicd/wazuh/*:5.0.0-latest` | Amazon Linux 2023 |

## Description

Deleting an agent from the manager also deletes its documents from the indexer, which is correct and deliberate. The agent then re-enrolls under a new id. From that point it is connected, authenticated and reporting successful synchronizations every cycle, and its inventory in the indexer stays empty.

The cause is that the deletion happens on the manager while the agent's local synchronization state is untouched. Each module keeps a persisted `first_sync_completed` flag that gates the only code path able to send a full snapshot. The flag survives re-enrollment, so every cycle computes a delta against a baseline the manager no longer has, sends nothing, and reports success. Only objects that change *after* the deletion ever appear.

This is the normal remedy for several problems and a routine step when re-imaging a host, so the endpoint comes back invisible while every health signal says it is fine.

Stock single-node deployment from the shipped images, manager `<auth>` block as it ships, no `<force>` block. The agent installation is left completely untouched: same container, same filesystem, same local synchronization databases. That is the point of the report, and it is also what makes the flow realistic, since an agent is normally deleted server-side precisely because the endpoint is still out there and expected to come back.

The agent re-enrolls on its own, normally within a minute. No restart is required to reproduce this.

**Three things are needed to get a stack running from the shipped images at all**, unrelated to this defect but blocking for anyone reproducing it:

- `<remote>` binds to `127.0.0.1` in the shipped `wazuh-manager.conf`, so no agent can connect until it is rebound to `0.0.0.0`. Reported separately.
- The agent image expects `WAZUH_MANAGER_SERVER`. With `WAZUH_MANAGER` the agent exits with `(4112): Invalid server address found: ''` and `(1215): No client configured`.
- The manager rejects an `<indexer>` block without an `<ssl>` child even when the host is plain `http://` (`src/shared/src/engine_external.c:547`).

Syscollector ships `<interval>1h</interval>` with `<scan_on_start>yes</scan_on_start>`. Since the agent re-enrolls without restarting, no syscollector scan runs in the first hour, so a short observation window proves nothing for that module on its own. Lower the interval, as the steps below do, or wait past a scan.

## Steps to reproduce

```bash
# 1. baseline: documents per agent
curl -sk -u admin:admin 'https://<indexer>:9200/wazuh-states-*/_search?size=0' \
  -d '{"aggs":{"a":{"terms":{"field":"wazuh.agent.id","size":30}}}}'

# 2. shorten the syscollector interval so scans are observable, then restart the agent
#    (this also confirms a clean-state agent does synchronize in full)
docker exec <agent> sed -i '0,/<interval>1h<\/interval>/s//<interval>2m<\/interval>/' /var/ossec/etc/ossec.conf
docker restart <agent>

# 3. delete the agent, the documented way
TOKEN=$(curl -sk -u <user>:<pass> -X POST 'https://<manager>:55000/security/user/authenticate?raw=true')
curl -sk -X DELETE -H "Authorization: Bearer $TOKEN" \
  'https://<manager>:55000/agents?agents_list=<id>&older_than=0s&status=all'

# 4. the agent re-enrolls with a new id on its own; confirm it
docker exec <agent> cut -d' ' -f1,2 /var/ossec/etc/client.keys

# 5. sample the count for 12 minutes, and count syscollector scans in parallel
for i in $(seq 1 12); do
  curl -sk -u admin:admin 'https://<indexer>:9200/wazuh-states-*/_search?size=0' \
    -d '{"aggs":{"a":{"terms":{"field":"wazuh.agent.id","size":10}}}}'
  docker logs <agent> 2>&1 | grep -icE "syscollector.*(starting|evaluation|scan)"
  sleep 60
done

# 6. causality check: clear the local sync state and restart
docker exec <agent> /var/ossec/bin/wazuh-control stop
docker exec <agent> rm -f /var/ossec/queue/fim/db/* /var/ossec/queue/sca/db/* /var/ossec/queue/syscollector/db/*
docker exec <agent> /var/ossec/bin/wazuh-control start
```

## Current behavior

Measured on a purpose-built 5.0.0 stack (manager, indexer, one agent):

| Phase | Documents in `wazuh-states-*` |
| --- | --- |
| Baseline, agent `001` | **1043** |
| After `DELETE /agents` | **0** |
| Re-enrolled as `002`, 9 samples over 9 minutes | **0** every sample |
| Re-enrolled as `003`, 12 minutes across **14 syscollector scans** | **2 to 5** |
| After clearing the local sync databases and restarting | **1043**, in under one minute |

The manager does exactly what it should:

```
11:19:54 wazuh-manager-authd: INFO: Agent '001' (repro-agent) deleted (requested locally)
11:19:54 inventory-sync-server:sync: INFO: Deleted every document of agent 001 (wazuh-states-*, ...)
```

**The agent is healthy throughout and says so.** The manager reports it `active` with a current keepalive, and the agent logs `finished successfully` every cycle, 22 times after re-enrolling, while the indexer holds nothing for it:

```
11:25:44 wazuh-syscheckd: INFO: FIM synchronization finished successfully.
11:25:54 wazuh-syscheckd: INFO: FIM synchronization finished successfully.
11:27:04 wazuh-syscheckd: INFO: FIM synchronization finished successfully.
```

**Scans do run, and only volatile objects come back.** Over the 12-minute window syscollector completed 14 scans and the count oscillated `0, 4, 5, 3, 2, 4`. What those documents are settles the interpretation:

```
wazuh-states-inventory-hardware     1     free memory, changes by itself
wazuh-states-inventory-ports        2     port 1517, the agent's own ephemeral connections
wazuh-states-inventory-processes    1     tail, an ephemeral process
```

Zero in `fim-files` (799 before), zero in `sca` (183 before), zero in `inventory-users` (14), `inventory-groups` (25), `packages`, `networks`, `interfaces` and `protocols`. That is delta-only behaviour against an empty baseline: only what changed after the deletion is ever sent, and the stable inventory never returns. The oscillation between 2 and 5 is ports and processes appearing and disappearing, which also demonstrates the pipeline is working and simply has nothing to carry.

**The causality check is decisive.** Clearing `/var/ossec/queue/{fim,sca,syscollector}/db/*` and restarting restored **1043** documents in under a minute. The data was on the endpoint the whole time; what was missing was the decision to send it.

## Root cause

Each module persists a `first_sync_completed` flag in its own local database, and that flag is keyed to nothing that changes when the agent's identity does:

```
src/wazuh_modules/sca/sca_impl/src/sca_impl.cpp:67   SCA_FIRST_SYNC_COMPLETED_METADATA_KEY {"first_sync_completed"}
src/syscheckd/src/run_check.c:95                     FIM_FIRST_SYNC_COMPLETED_METADATA_KEY  "first_sync_completed"
src/wazuh_modules/syscollector/...                   SYSCOLLECTOR_VD_FIRST_SYNC_COMPLETED_METADATA_KEY
```

Read straight from a running agent's databases, the rows carry a timestamp and no agent identity:

```
fim.db     table_metadata: ('first_sync_completed', 1787649408)
local.db   table_metadata: ('first_sync_completed', 1787649409, 1)    # syscollector
sca.db     sca_metadata:   ('first_sync_completed', 1787649443, 1)
```

The flag gates the only path that sends a full snapshot:

```cpp
/* src/wazuh_modules/sca/sca_impl/src/sca_impl.cpp:160 */
const bool firstSyncPending = !m_firstSyncCompleted.load();
...
if (firstSyncPending)
{
    result = synchronizeDatabaseSnapshot(false, "initial synchronization");
}
else
{
    result = m_spSyncProtocol->synchronizeModule(mode);   /* delta only */
}
...
if (result.success && firstSyncPending)
{
    updateMetadataValue(SCA_FIRST_SYNC_COMPLETED_METADATA_KEY, Utils::getSecondsFromEpoch());
    m_firstSyncCompleted.store(true);
}
```

An agent that has been re-identified takes the `else` branch and ships deltas against a manager that has none of its data. The delta is empty, and the protocol reports success without contacting the manager at all:

```cpp
/* src/shared_modules/sync_protocol/src/agent_sync_protocol.cpp:215 */
bool success = true;
...
    if (dataToSync.empty())
    {
        if (!sentAny)
        {
            m_logger(LOG_DEBUG, "No items to synchronize in DELTA mode");
        }
        break;
    }
```

So the mechanism that would repair this exists, is used exactly once per agent lifetime at first enrollment, and re-enrollment is not treated as one.

## Expected results

Receiving a new agent id is exactly the condition `synchronizeDatabaseSnapshot()` exists for: the manager holds nothing for this identity, so the next synchronization should be a full snapshot rather than a delta.

Concretely, any of these closes it:

- Re-enrollment clears the per-module `first_sync_completed` state, so the next cycle takes the snapshot branch that already exists.
- Or the agent id the last successful synchronization was made under is persisted alongside the flag, and a change of id is treated as first-sync-pending.
- Or the manager can request a full resynchronization of one agent, which is the same gap `wazuh#38357` raises for the vulnerability scanner.

Failing all of those, an agent whose inventory is empty while it reports successful synchronizations should at least say so, rather than logging `finished successfully` every cycle.

## Impact

An agent that is deleted and re-registered comes back invisible. No FIM baseline, so no future change can be judged against one. No package inventory, so vulnerability detection has nothing to scan. No SCA results. The agent shows as `active`, its logs report successful synchronizations every cycle, and the only way to notice is to look at the inventory and find it empty.

Recovery needs manual work on the endpoint, clearing the agent's local sync state or reinstalling, because nothing on the manager side can ask for a full resynchronization.

nbertoldo:
## Description

Closes #38601.
 
Deleting an agent on the manager also purges its documents from the indexer, and the agent then re-enrolls under a new id. Its local databases survive that untouched — including the per-module `first_sync_completed` marker, which is keyed to nothing that changes when the agent's identity does. Every later cycle therefore computes a delta against a baseline the manager no longer has for this identity, sends nothing, and reports success.

Reproducing the issue end to end on a 5.0.0 stack showed the report is right about the root cause but wrong about the consequence, and the difference decides what this PR has to fix.

**The inventory does come back, but late.** The manager answers those deltas with a `409` and the existing recovery path repairs each index. Every one of the 127 `409` responses in the run carried `(attempt N/5)` and arrived on a ten-second cadence, which places them inside `requiresFullSync()`'s retry loop — the integrity-check path — so the repair is gated by `integrity_interval` end to end. With the shipped default of `86400` the first check after a deletion is up to 24 hours away, and syscollector then walks its indices serially at roughly 42 seconds each on top of that wait.

**The vulnerability documents never come back at all.** 14670 to zero, and still zero long after the full inventory was back under the new id.

The cause is a lane mismatch in the recovery path itself: it cleared the manager's index and re-persisted its rows through the plain sync protocol unconditionally. For the system, packages and hotfixes indices that is the wrong lane — the manager routes a VD session through the scan lane, which both indexes the documents and feeds them to the vulnerability scanner — so a recovery sent on the plain protocol repopulated `wazuh-states-inventory-*` while leaving the scanner with nothing to evaluate. Measured: with the fix the manager reports `679 packages scanned` for a re-enrolled agent, against `1 packages scanned` before it.

`vd_first_sync_completed` has to be reset too, and for a reason beyond the data. It selects `Option::VDSYNC` over `Option::VDFIRST`, and `VDFIRST` does more than add an agent-scoped `deleteByQuery`: it selects the manager's first-scan chain, which drops `TEventSendReport` (`scanOrchestrator/factoryOrchestrator.hpp`) and logs "no security alerts generated (alerts are suppressed on first scan)". To the manager a re-enrolled agent is a new agent, and a new agent's first scan indexes its vulnerabilities without alerting on each one. Leaving the marker armed would have made the recovered inventory arrive as an ordinary synchronization, raising an alert for every CVE the host already had — 226 vulnerable packages on the Amazon Linux agent used for the evidence below. Nothing alerted before this PR only because the data never came back at all.

Resetting the marker is still not a one-liner, for the reason measured earlier: on its own it re-arms without contacting the manager, because `VDFIRST` sends whatever is in the VD queue, a scan on a populated `local.db` produces no diffs, and `synchronizeDeltaByBlocks` takes its empty-queue early break with `success` still holding the `true` it was initialised with. That is why the reset happens *after* the three VD indices have been queued and before the single session that carries them. The `VDFIRST` distinction also matters for #38357, where the identity does not change and stale documents are the whole problem.

A secondary defect surfaced while reproducing: `bridge_reenroll_thread()` updated the HTTPS client's signing identity but left the shared-memory metadata provider holding the pre-enrollment id. Every session's `Start.agentid` is stamped from that provider and the manager rejects a mismatch with `403 identity mismatch`, so sessions sent inside that window are refused and any module comparing against the provider stays blind for as long as it lasts.

## Proposed Changes

- **`shared_modules/sync_protocol`** — adds `AgentSyncProtocol::currentAgentId()` (`asp_get_agent_id()` for C callers), reporting the id this process synchronizes under from the shared-memory metadata provider. That is the same value, through the same call, that already stamps every session's `Start.agentid`, so a module comparing against it can never force a resync stamped with a different id than the one it compared against — reading `client.keys` instead would, since the two disagree precisely during a re-enrollment. Zero means "nothing published yet" and callers must read that as unknown, never as changed. Static rather than a member of `IAgentSyncProtocol`: the id is process-global state, and keeping it off the interface leaves every existing mock untouched. Also adds `sentAnything` / `sent_anything` to the result structs, carrying what `success` alone cannot say — an empty queue takes the early break and returns the initial `true`, which is the intended contract but means success does not imply the manager was contacted. Both fields are appended last so existing positional aggregate initializations keep compiling.
- **`client-agent`** — `bridge_reenroll_thread()` now republishes the agent metadata after a successful re-enrollment, outside `g_https_client_lock` (`w_agentd_populate_metadata()` takes a mutex of its own and nesting the two would introduce a lock order nothing else in that file establishes), and only on a valid identity, since a failed validation leaves traffic paused with nothing to stamp. That function also now carries `vd_feed_offset` over from the record it replaces: agentd never owned that field — only agent-info writes it, from what the manager reports over `/control` — and publishing a zero makes every VD synchronization abort with `NO_VD_OFFSET_ERROR`, which fails `Syscollector::syncModule()` and makes the wodle skip the recovery pass behind it. That would have disabled the identity resync in precisely the window it targets. The carry-over also fixes the two call sites that predate this PR. The read behind it is retried, because `metadata_provider_get()` answers `-1` both when nothing has ever been published and when the reader hit the `updating` flag agent-info raises while writing — a different process, so agentd's own mutex does not serialize against it. Believing that `-1` once is enough to publish a zero and disable the resync in its own window. Retrying is the local fix; the structural one is a partial update in the provider, so agentd never round-trips a field it does not own.
- **`syscollector`** — `protocolForIndex()` puts the plain-versus-VD choice in one place, and the recovery path now uses it for both its `DataClean` and its re-persist. The per-table work moves out of `runRecoveryProcess()` into `resyncTableToManager()` unchanged, so more than one trigger can drive it; `isCollectorEnabledForTable()` comes out with it so the eleven-way skip chain is not duplicated. `checkAgentIdentity()` then resends every table when the id changed, driven from `runRecoveryProcess()` — which the wodle calls after a successful sync, so `resyncTableToManager()` reaching `syncModule()` internally cannot recurse. `get_synced_agent_id` exposes the new marker alongside the two that already exist. The resync runs in two phases: the plain lane keeps one session per table, so an index is cleared immediately before its own rows are queued and a failure stays contained; the three VD indices are cleared and re-persisted without sending, `vd_first_sync_completed` is reset once they are all queued, and one session carries them — a single `VDFIRST` over the complete inventory. Both halves are load-bearing. Sending the VD tables one at a time would hand the first session a partial inventory and re-arm the marker, leaving the rest to go out as `VDSYNC` and alert, and each `VDFirst` session opens with a `deleteByQuery` scoped to this agent, so a second one deletes what the first indexed. And they are persisted only after the last plain session because `syncModule()` drains both protocols — a plain table's session would otherwise flush whatever the VD lane had queued. The cost, stated plainly: those three indices are cleared before any of them is sent, so a session that never reaches the manager leaves all three empty. It self-heals, since the rows stay queued and the marker stays unset, and the identity marker is withheld meanwhile. Each table the identity path resends also stamps the integrity clock, exactly as the checksum loop does after its own resync: without that, the loop immediately behind it — same pass, same locked region — could find the interval elapsed for a table just full-replaced, ask the manager for a checksum it had no chance to recompute, and issue a second `notifyDataClean` for it. Duplicated work at best, and a clean ordered after the rows this pass queued would delete them.
- **`sca`** — `checkAgentIdentity()` compares the id at the top of each sync cycle and takes the snapshot branch that already exists. `increaseVersions` stays `false`: indexer document ids are agent-scoped, so under a new identity there are no higher-versioned documents of ours left to outrank.
- **`syscheckd`** — `fim_resync_on_agent_id_change()` compares the id before the per-table checksum loop and drives the existing full-replace primitive for every table when it changed, honouring `fim_shutdown_process_on()` between tables like the integrity loop that follows it. `fim_recovery_persist_table_and_resync()` now returns `bool` so the marker can be gated on something that proves a round trip: false on the paths that abandon without reaching the manager, true once it accepted the `DataClean`. And `fim_db_try_get_last_sync_time()` is new, reporting a failed read separately from an absent row — see the bullet below for why that distinction is load-bearing. `fim_db_get_last_sync_time()` keeps its contract untouched: answering `0` for both is fine for a timestamp, just not for a caller that treats absence as permission to record something. A table that cannot be resent no longer abandons the pass — it continues and withholds the marker, like syscollector's loop, which matters on Windows where there are three tables rather than one. And the adopt branch refuses while a shutdown is in progress: `FIMDB::executeQuery` answers a read with a silent no-op once the database is stopping, which reaches the caller as a clean read of an absent row. `FIMDB::updateItem` is gated by the identical condition today, so adopting would write nothing anyway, but that is two distant guards happening to agree rather than a contract either states. The agent-info flush path now makes the same empty-cycle distinction as the ordinary one. It deliberately does *not* gain the identity check: that branch is reached because FIM is paused, which is what forbids taking the scan mutex the resync needs, so both it and the integrity loop stay in the branch below and fire on the first cycle after FIM resumes.
- **All three modules** — an absent `synced_agent_id` is adopted without resyncing. That is what keeps an in-place upgrade of an existing fleet from having every agent resend its whole database at once, and it has a dedicated test in each module. A re-enrollment that hands back the *same* id is an explicit no-op, since that is the common case (a key rotation, an authd password change, any 401 that is not a deletion) and treating it as a change would resync the fleet on every credential refresh. Markers advance only after the manager accepted the `DataClean`, never on a sync result. The success log now distinguishes an empty cycle from a delivered one.
- **A failed metadata read is not an answer, in all three modules.** It is deliberately kept apart from "nothing was ever recorded", because that second case adopts the current id: folding them together means one transient busy database, in the window right after a re-enrollment, records the new id as already synchronized and suppresses the resync permanently. A failed read is treated like an unknown id — do nothing, try again next cycle. Syscollector's resync also continues past a table it could not resend rather than abandoning the pass, since abandoning it made every later pass re-clear and re-upload the tables that had already succeeded; only the marker is withheld.

### Results and Evidence

Verified end to end on real agents, with the packages this PR's own CI built (`a4b6307`, the merge of `8d82ff7f90`): manager + indexer on Ubuntu 24.04 from the 2026-08-28 nightly, one Amazon Linux 2023 agent (rpm) and one Ubuntu 22.04 agent (deb).

**`integrity_interval` was left at its shipped default of 24 h.** That is the control that makes the result attributable: nothing that comes back can be credited to the `409`/integrity path, because that path cannot run inside the observation window. The discriminating log lines for it (`Starting integrity validation process`, `Checksum mismatch`, `full sync required`) stayed at zero in every sample.

Each agent was enrolled, allowed to populate, then **deleted through the manager's API** — the scenario in the report — and left to re-enroll under a new id. Run twice; the table is the second run (agent `001`→`005` and `002`→`006`), and the first run agreed index for index.

| index | AL2023 before | AL2023 after | Ubuntu before | Ubuntu after |
|---|---|---|---|---|
| `fim-files` | 2509 | **2509** | 4091 | **4091** |
| `sca` | 183 | **183** | 207 | **207** |
| `inventory-packages` | 679 | **679** | 849 | **849** |
| `inventory-users` | 29 | **29** | 38 | **38** |
| `inventory-groups` | 53 | **53** | 66 | **66** |
| `inventory-networks` | 7 | **7** | 6 | **6** |
| `inventory-interfaces` | 3 | **3** | 3 | **3** |
| `inventory-protocols` | 6 | **6** | 5 | **5** |
| `inventory-system` | 1 | **1** | 1 | **1** |
| `inventory-services` | 127 | **127** | 152 | **152** |
| **total** | **3597** | **3597** | **5418** | **5418** |

Deleted at 14:55:26, fully restored by 15:04:49 — under ten minutes, against a first integrity check up to 24 hours away.

All three modules announced it, once each:

```
FIM was last synchronized as agent 1, now running as agent 3. Resending every monitored entry.
Inventory was last synchronized as agent 1, now running as agent 3. Resending every table.
SCA was last synchronized as agent 1, now running as agent 3. Resending a full snapshot.
```

And all three markers advanced to the new identity, read from the agents' own databases: syscollector and SCA report `5`/`6` through `get_synced_agent_id`, and FIM's `table_metadata` holds `synced_agent_id: 5` and `6`, matching each `client.keys`. (FIM's markers are only readable with the agent stopped: `wazuh-syscheckd` keeps a transaction open, so a concurrent reader sees the pre-transaction snapshot of that table.)

**The VD lane, which is what the routing fix is for.** The manager reports the re-enrolled agents' packages reaching the vulnerability scanner, on both runs and both distributions:

```
Agent '005' - Packages scan completed in 65 ms: 679 packages scanned
Agent '006' - Packages scan completed in 54 ms: 849 packages scanned
```

679 and 849 are each agent's full package count. Before the fix the same scenario produced `1 packages scanned`.

**What the Linux run could not show, and why.** On these two agents `wazuh-states-vulnerabilities` did not refill. That is a manager-side content failure, not an agent one: the scans above are the same size as the pre-deletion scans, but the content store stopped resolving advisories between them. Before the first deletion the manager reported `226 vulnerable packages` (AL2023) and `256` (Ubuntu); afterwards the AL2023 scan emitted 674 `Couldn't find column family: 'alas_'` warnings over its 679 packages and found none, and Ubuntu dropped to 2. Restarting the manager reloaded the feed ("CVE feed fully loaded") without restoring those column families — the behaviour reported in #38272. The agent's side of the contract is what the scan sizes above demonstrate, and the Windows run below closes the rest of the cycle on a host where the feed still resolves.

**Windows 11, which is where the rest of it lives.** The same CI build, against the same manager. This closes what the Linux run could not reach: FIM has three tables there instead of one, hotfixes is the VD lane's second index, and the vulnerability feed for Windows still resolves in this manager, so the whole VD cycle is observable end to end, alerts included. The table is the run on the final build (`010`→`011`), whose baseline was taken after the initial sync had settled.

| index | before | after |
|---|---|---|
| `fim-files` | 11 | **11** |
| `fim-registry-keys` | 6415 | **6415** |
| `fim-registry-values` | 13088 | **13088** |
| `inventory-packages` | 68 | **68** |
| `inventory-hotfixes` | 13 | **13** |
| `inventory-system` | 1 | **1** |
| `sca` | 482 | **482** |
| `wazuh-states-vulnerabilities` | 1254 | **1254** |

Every index exactly, including both registry tables — the case where "keep going past a table that could not be resent" stops being a no-op — and including the vulnerability documents, which come back through a session the manager treats as a first scan:

```
Agent '010' - Packages scan completed in 10 ms: 68 packages scanned, 0 skipped, 2 vulnerable packages
Agent '010' - First vulnerability scan completed: vulnerabilities indexed, no security alerts generated (alerts are suppressed on first scan)

Agent '011' - Packages scan completed in 10 ms: 68 packages scanned, 0 skipped, 2 vulnerable packages
Agent '011' - Vulnerability scan completed: 1254 new, 0 solved
Agent '011' - First vulnerability scan completed: vulnerabilities indexed, no security alerts generated (alerts are suppressed on first scan)
```

1254 findings re-indexed for the new identity, and not one of them alerted. `010`'s line is the control rather than the result: that agent was a clean install, so its scan is a genuinely first one, and it proves the observable exists before the case that matters is measured. `011` is the case that matters — a re-enrolled agent, whose inventory the manager scanned in full and indexed without raising an alert for a single pre-existing CVE.

An earlier pair of runs on the build before the marker reset (`007`→`008`→`009`) agreed index for index and produced the same scan sizes, which is what established the data half of this before the alert half existed.

Identity lines across all the runs, three per re-enrollment, and the integrity path's discriminators never appear in the agent's log at all — `Starting integrity validation process`, `Checksum mismatch` and `full sync required` count zero, with `integrity_interval` at the 24 h the Windows configuration ships.

The routing fix was measured on a 5.0.0 stack (manager + indexer on Ubuntu 24.04, agent on Amazon Linux 2023, both from the 2026-08-26 nightly), swapping only `libsyscollector.so` so the change was the single variable:

| | packages reaching the scanner |
|---|---|
| shipped library (agent `002`) | **1** |
| with the routing fix (agent `003`) | **678** |
| with the routing fix, repeated (agent `004`) | **679** |

That earlier measurement isolated the routing change alone; the identity detection was implemented afterwards, and is what the runs above exercise.

### Artifacts Affected

- `wazuh-syscheckd` — FIM identity check, the changed signature of its recovery primitive, and one new export (`fim_db_try_get_last_sync_time`) on `libfimdb.so`.
- `wazuh-modulesd` — carries `libsyscollector.so` and `libsca.so`, both of which gain the identity check; syscollector also gains the VD routing fix.
- `wazuh-agentd` — republishes the agent metadata after a re-enrollment. Note that the `vd_feed_offset` carry-over changes `w_agentd_populate_metadata()` for its two pre-existing callers too, not only for the new one: they were zeroing that field on every `/startup` and on every group change.
- `.github/actions/check_files/rpm_linux_agent_amd64.csv` — the RPM amd64 checkfiles baseline for `libsca.so` (783864 → 866120) and `libfimdb.so` (224672 → 249304), which this branch tripped. Most of that gap is not from this PR: those two entries were last refreshed on 2026-02-06 and 2025-11-11, while their DEB amd64 twins have been refreshed since to values within a percent of what this build produces, so the RPM baseline had drifted to roughly 9.9% and a few kilobytes of new code carried it past the ±10% tolerance. Nothing new is linked in: every `CMakeLists.txt` change here is to test targets.
- **CI, in its own commit** — the Linux integration-test workflow now installs the agent with `WAZUH_MANAGER_ENDPOINT="127.0.0.1:1517/"`. The trailing slash is the opt-out from the manager's default `/wazuh-manager/` prefix, which an agent addresses unless something says otherwise; the suites that never declare a `<manager>` block of their own (sca, syscollector, github, office365) had nothing to do so, and every module ended up blocked in `startup_gate_wait_for_ready()` behind a 404 on enrollment. Not caused by this PR — it reproduced identically on unrelated branches at the same merge-base, and surfaced here only because every Linux suite lists that workflow file among its trigger paths, so editing it runs the whole matrix. The execd suite's own template needed the same opt-out; #38624 has since landed it upstream, along with the grammar this line now uses (`<endpoint>` carries the whole target, `host[:port][/[prefix]]`, and an explicitly empty value is rejected rather than read as "no prefix").
- `libagent_sync_protocol.so` — new accessor and a new field on the result struct. **This crosses the C ABI**: adding a field is source-compatible and everything is rebuilt from one tree, but a partial rebuild against a stale shared library would read garbage, so a clean build is required.

### Configuration Changes

None. No new settings, no changes to defaults, no changes to the shipped configuration files.

### Documentation Updates

None in this PR. Two notes are now inaccurate and worth a follow-up: `docs/ref/modules/authd/architecture.md` concedes that new agents "can inherit documents from the agents that held their ids before, in the indices they do not resynchronise themselves", and `docs/ref/modules/utils/sync-protocol/api-reference.md` documents the `409` as the only resync trigger.

### Tests Introduced

- **`sca_identity_test.cpp`** (new, with its own target): five cases covering the decision procedure — marker absent is adopted without resyncing, unchanged id is a no-op, unknown id is a no-op, a changed id clears the index and resends, and a refused `DataClean` records nothing. The fixture pauses the module, since `syncModule()` answers immediately otherwise, and works in its own directory because the metadata provider is a file at a relative path that other SCA test binaries write too.
- **`test_recovery.c`**: eight shapes for `fim_resync_on_agent_id_change()` — an absent marker adopted without resending, an unchanged id, an unknown id, a changed id that resends (full scaffolding, driving the real resync end to end and asserting both markers advance only after the manager accepted the `DataClean`), a table that could not be resent recording nothing, a failed metadata read adopting nothing, three tables that all fail to assert the pass keeps going (cmocka fails on an expectation left unconsumed, which is exactly what the abort-on-first version would have left behind), and an absent marker that must not be adopted while a shutdown is in progress. 20 tests in that binary, all passing.
- **`test_agent_sync_protocol.cpp`**: `sentAnything` is false for an empty queue, and four cases for `currentAgentId()` — the published id, zero when nothing is published, zero-padding ignored, and a non-numeric id reading as unknown.
- **`syscollectorImp_test.cpp`**: the `get_synced_agent_id` query defaults to zero and returns the stored value.
- **`syscollector_identity_tests.cpp`** (new): eight cases for the module that carries this PR's measured fix, which had none for its decision procedure. The one worth naming: with packages and users enabled, the **VD protocol receives the `DataClean` for the packages index and the plain protocol the one for users** — sending a recovered VD index down the plain lane is the defect this PR exists to fix, and nothing asserted it before. Also: absent marker adopted without resending, unchanged id, unknown id, a failed marker read adopting nothing, disabled collectors skipped, a refused `DataClean` withholding the marker while the pass continues, and the integrity clock stamped per resent table. Driving any of it needs a manager to answer every `DataClean`, so the protocols are mocked and reach the module's members through `FRIEND_TEST` declarations in a new `syscollector_defs.hpp` — the pattern `vulnerabilityScannerFacade` already uses, guarded so a production build carries no declarations at all. Syscollector is a `final` singleton with a private constructor, so SCA's approach (a subclass assigning the protected member) was not available to it, and a public setter would have been API only tests would ever call. `resyncTableToManager()`'s coverage exclusion is narrowed to match: it claimed the whole function needed a live manager, which is now true only of the per-item persist loop.
- **`test_start_agent.c`**: three cases for the `vd_feed_offset` carry-over — a busy read retried until it clears, a clean first read, and every attempt failing (which must publish a zero and say so rather than silently). `metadata_provider_get` joins the target's `--wrap` list for them: the real one reads a file at a path relative to the working directory that other test binaries in this tree also write, and the number of reads is part of what these cases assert.
- **`test_https_client_bridge.c`**: the metadata provider is republished exactly once on a successful re-enrollment, and not at all when the identity fails validation.
- `test_run_check.c`'s shared sync-body helper now expects the empty-cycle wording, since the shared `asp_sync_module` wrapper zero-fills its result and `sent_anything` is therefore false; it also scripts `asp_get_agent_id` to zero, so the identity check is a no-op in cases that are about the integrity loop. That wrapper is now on the target's `--wrap` list, which it was not: the real one reads a machine-global provider file, so a leftover from another test binary could send those cases down an unscripted path.
- `sca_test.cpp`'s fixture now resets the metadata provider, since it is a file other test binaries in this tree also write and these cases would otherwise inherit whichever id was left behind.

**One coverage gap, stated rather than left to be discovered.** The Windows target was not built locally (the devcontainer's cross-build fails in curl's OpenSSL detection, in externals, before reaching any of this code), so its *unit* tests rest on CI. What only Windows exercises at runtime — FIM's three tables instead of one, and hotfixes as the VD lane's second index — is no longer untested, but by the Windows 11 run in Results and Evidence rather than by anything I could compile here. (Two earlier revisions of this description listed more. FIM's call site was listed on the strength of a `GROUP SETUP` failure in `test_run_check` here, which turned out to be stale objects in my tree rather than the test: rebuilt against `syscheckd_lib` it runs and passes 32/32. Syscollector's decision procedure was the other, and it now has its own tests.)

Worth recording for anyone reproducing the local runs: `SCATest` and `SCASyncManagerTest` segfault in this devcontainer and pass in CI. Confirmed environmental by checking out the base version of `sca_test.cpp` and rebuilding, which segfaults identically.

**One known limitation, filed separately rather than fixed here.** This work runs with FIM's scan *and* realtime mutexes held, and `realtime_process()` reads the inotify descriptor under the second one — so a recovery long enough to fill the kernel's inotify queue costs dropped filesystem events. It is not introduced here: the integrity loop has always run the same primitive under the same mutexes, and after a re-enrollment every table mismatches, so it did the same work later rather than less of it. The fix belongs to the recovery primitive's own critical section, benefits the integrity path equally, and would mean changing locking semantics this PR otherwise does not touch. The cost is documented at both call sites.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/.github/actions/check_files/rpm_linux_agent_amd64.csv
+++ b/.github/actions/check_files/rpm_linux_agent_amd64.csv
@@ -77,13 +77,13 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/lib/libhttps_client.so,root,wazuh,750,file,-rwxr-x---,1132872,0.1
 /var/ossec/lib/libschema_validator.so,root,wazuh,750,file,-rwxr-x---,386024,0.1
 /var/ossec/lib/libsysinfo.so,root,wazuh,750,file,-rwxr-x---,2555328,0.1
-/var/ossec/lib/libfimdb.so,root,wazuh,750,file,-rwxr-x---,224672,0.1
+/var/ossec/lib/libfimdb.so,root,wazuh,750,file,-rwxr-x---,249304,0.1
 /var/ossec/lib/libbpf.so,root,wazuh,750,file,-rwxr-x---,526048,0.1
 /var/ossec/lib/libfimebpf.so,root,wazuh,750,file,-rwxr-x---,60080,0.1
 /var/ossec/lib/modern.bpf.o,root,wazuh,750,file,-rwxr-x---,998504,0.1
 /var/ossec/lib/libstdc++.so.6,root,wazuh,750,file,-rwxr-x---,2530376,0.1
 /var/ossec/lib/libsyscollector.so,root,wazuh,750,file,-rwxr-x---,769128,0.1
-/var/ossec/lib/libsca.so,root,wazuh,750,file,-rwxr-x---,783864,0.1
+/var/ossec/lib/libsca.so,root,wazuh,750,file,-rwxr-x---,866120,0.1
 /var/ossec/lib/libagent_info.so,root,wazuh,750,file,-rwxr-x---,418344,0.1
 /var/ossec/lib/libagent_metadata.so,root,wazuh,750,file,-rwxr-x---,14472,0.1
 /var/ossec/lib/libgcc_s.so.1,root,wazuh,750,file,-rwxr-x---,187064,0.1
--- a/.github/workflows/5_testintegration_linux-integration-tests.yml
+++ b/.github/workflows/5_testintegration_linux-integration-tests.yml
@@ -127,7 +127,15 @@ jobs:
             exit 1
           fi
 
-          sudo WAZUH_MANAGER="127.0.0.1" dpkg -i "${PACKAGE_PATH}" || sudo apt-get install -f -y
+          # The trailing '/' is the opt-out from the manager's default /wazuh-manager/ prefix
+          # (#38624: <endpoint> is now the whole connection target, "host[:port][/[prefix]]", and
+          # an omitted prefix takes the default). Without it the agent addresses that prefix, gets
+          # a 404 from the manager these suites run against, and never opens agentd's startup gate
+          # -- so modulesd blocks in startup_gate_wait_for_ready() and no module ever starts. The
+          # suites that own a <manager> block carry the same opt-out in their own templates; the
+          # ones that do not (sca, syscollector, github, office365) inherit the package's
+          # ossec.conf and need it here.
+          sudo WAZUH_MANAGER_ENDPOINT="127.0.0.1:1517/" dpkg -i "${PACKAGE_PATH}" || sudo apt-get install -f -y
 
           sudo awk '
             /<rootcheck>/ { in_block=1; print; next }
--- a/src/client-agent/src/https_client_bridge.c
+++ b/src/client-agent/src/https_client_bridge.c
@@ -176,11 +176,29 @@ void *bridge_reenroll_thread(void *arg)
      * identity, but the module must not assume that -- signing with a stale
      * id after the key changed would desync from whatever id the manager
      * now associates with this key). Both move together, never just the key. */
-    if (!hc_set_agent_identity(handle, keys.keyentries[0]->id, keys.keyentries[0]->raw_key)) {
+    const bool identity_ok = hc_set_agent_identity(handle, keys.keyentries[0]->id, keys.keyentries[0]->raw_key);
+
+    if (!identity_ok) {
         merror("https_client: re-enrolled, but the new identity failed validation; traffic stays paused.");
     }
 
     w_mutex_unlock(&g_https_client_lock);
+
+    /* Republish the agent metadata. The sync protocol stamps every session's
+     * Start.agentid from the shared-memory provider, and each module compares against that
+     * same value to notice its own id changed. Without this the provider keeps the
+     * pre-enrollment id until the next /startup response or agent-info cycle happens to
+     * rewrite it: sessions sent inside that window carry an id the manager answers with 403
+     * (identity mismatch), and identity detection stays blind for as long as it lasts.
+     *
+     * Published after unlocking: w_agentd_populate_metadata() takes a mutex of its own, and
+     * nesting it inside g_https_client_lock would introduce a lock order nothing else in this
+     * file establishes. Only on a valid identity -- when validation failed the module keeps
+     * traffic paused, so there is nothing to stamp yet. */
+    if (identity_ok) {
+        w_agentd_populate_metadata();
+    }
+
     return NULL;
 }
 
--- a/src/client-agent/src/start_agent.c
+++ b/src/client-agent/src/start_agent.c
@@ -140,6 +140,12 @@ int try_enroll_to_server(void) {
     return 0;
 }
 
+/* Attempts, and the pause between them, for reading the record that is about to be replaced.
+ * Sized for the `updating` window described in w_agentd_populate_metadata(): a struct copy in
+ * another process, not a retry loop around anything that can stay busy. */
+#define METADATA_READ_ATTEMPTS 3
+#define METADATA_READ_RETRY_MS 10
+
 /* Populate shared memory with agent metadata so the first keepalive already
  * contains full agent info.
  *
@@ -155,6 +161,46 @@ void w_agentd_populate_metadata(void)
     agent_metadata_t metadata = {0};
 
     w_mutex_lock(&metadata_mutex);
+
+    /* Carry over vd_feed_offset, the one field agentd does not own. metadata_provider_update()
+     * replaces the whole record, and that offset is written only by agent-info, from what the
+     * manager reports over /control -- publishing a zero makes every VD synchronization abort
+     * with NO_VD_OFFSET_ERROR until agent-info's next cycle rewrites it, which in turn fails
+     * Syscollector::syncModule() and makes the wodle skip the recovery pass behind it. Every
+     * other field here is either local to the agent or set below from what the handshake
+     * reported.
+     *
+     * The read is retried because -1 is not one answer but two: nothing has ever been published
+     * (first boot, where a zero offset is the truth), or the reader hit the `updating` flag
+     * agent-info raises while it writes. agent-info is a different process, so the mutex above
+     * does not serialize against it, and that window is no wider than a struct copy -- but
+     * reading it as "no offset" would suppress VD synchronization until agent-info's next
+     * /control cycle, which is to say it would disable the identity resync this change exists
+     * for, in exactly the window right after a re-enrollment where it is needed. Retrying is the
+     * local fix; the structural one is a partial update in the provider, so agentd never has to
+     * round-trip a field it does not own. */
+    {
+        agent_metadata_t previous = {0};
+        int carried = 0;
+
+        for (int attempt = 0; attempt < METADATA_READ_ATTEMPTS; attempt++) {
+            if (metadata_provider_get(&previous) == 0) {
+                metadata.vd_feed_offset = previous.vd_feed_offset;
+                metadata_provider_free_metadata(&previous);
+                carried = 1;
+                break;
+            }
+
+            if (attempt + 1 < METADATA_READ_ATTEMPTS) {
+                w_time_delay(METADATA_READ_RETRY_MS);
+            }
+        }
+
+        if (!carried) {
+            /* Also the ordinary first-boot path, where there is genuinely nothing to carry. */
+            mdebug1("No published agent metadata to carry the VD feed offset from.");
+        }
+    }
 
 #ifdef WIN32
     os_info *os = get_win_version();
--- a/src/client-agent/src/start_agent.c
+++ b/src/client-agent/src/start_agent.c
@@ -143,8 +143,9 @@ int try_enroll_to_server(void) {
 /* Populate shared memory with agent metadata so the first keepalive already
  * contains full agent info.
  *
- * Serialized: start_agent() publishes from the main thread while
- * bridge_on_startup_result() publishes from the module's dispatcher thread, and
+ * Serialized: start_agent() publishes from the main thread, bridge_on_startup_result()
+ * and bridge_on_control_response() from the module's dispatcher thread, and
+ * bridge_reenroll_thread() from the re-enrollment worker, while
  * metadata_provider_update() has no writer lock of its own -- it only raises an
  * `updating` flag readers poll, so two overlapping writers would clear it while
  * one is still copying and a reader would get a torn record. */
--- a/src/shared_modules/sync_protocol/include/agent_sync_protocol.hpp
+++ b/src/shared_modules/sync_protocol/include/agent_sync_protocol.hpp
@@ -81,6 +81,23 @@ class AgentSyncProtocol : public IAgentSyncProtocol
         /// @param maxBytes Maximum bytes per session, or 0 to keep the default.
         static void setSessionMaxBytes(size_t maxBytes);
 
+        /// @brief Returns the agent id this process is currently synchronizing under.
+        ///
+        /// Reads the shared-memory metadata provider -- the same source, through the same call,
+        /// that stamps every outgoing session's `Start.agentid` (see buildFullSessionBuffer).
+        /// A module comparing against this value can therefore never force a resync stamped with
+        /// an id different from the one it compared against.
+        ///
+        /// Static, not a member of IAgentSyncProtocol: the agent id is process-global state
+        /// rather than per-instance, and keeping it off the interface leaves every existing mock
+        /// untouched. The C counterpart is asp_get_agent_id().
+        ///
+        /// @return The agent id as a positive integer, or 0 when the provider has published
+        ///         nothing yet or holds a value that is not a plain number. Callers must treat 0
+        ///         as "unknown", never as "changed": an unavailable provider, or one still
+        ///         holding the previous id, must not look like a new identity.
+        static long currentAgentId();
+
         /// @copydoc IAgentSyncProtocol::stop
         void stop() override;
 
--- a/src/shared_modules/sync_protocol/include/agent_sync_protocol_c_interface.h
+++ b/src/shared_modules/sync_protocol/include/agent_sync_protocol_c_interface.h
@@ -29,6 +29,23 @@ extern "C" {
 /// @param max_session_bytes Maximum bytes per session, or 0 to keep the default.
 void asp_set_session_max_bytes(uint64_t max_session_bytes);
 
+/// @brief Returns the agent id this process is currently synchronizing under.
+///
+/// Reads the shared-memory metadata provider -- the same source, through the same call, that
+/// stamps every outgoing session's `Start.agentid`. A module comparing against this value can
+/// therefore never force a resync stamped with an id different from the one it compared against,
+/// which is why detection must read it from here and not from `client.keys`: the two disagree
+/// precisely during the re-enrollment window, and the manager answers a mismatch with 403.
+///
+/// Deliberately handle-less, like asp_set_session_max_bytes(): the agent id is process-global
+/// state, unrelated to any one protocol instance, and modules hold an IAgentSyncProtocol
+/// reference rather than the concrete class.
+///
+/// @return The agent id as a positive integer, or 0 when the provider has published nothing yet
+///         or holds a value that is not a plain number. Callers must treat 0 as "unknown", never
+///         as "changed" -- an unavailable provider must not look like a new identity.
+long asp_get_agent_id(void);
+
 /// @brief Creates an instance of AgentSyncProtocol.
 ///
 /// @param module Name of the module associated with this instance.
--- a/src/shared_modules/sync_protocol/include/agent_sync_protocol_c_interface_types.h
+++ b/src/shared_modules/sync_protocol/include/agent_sync_protocol_c_interface_types.h
@@ -55,6 +55,11 @@ typedef struct SyncModuleResult_t
     /// @brief True when the local sync intake itself could not be reached. See
     /// SyncModuleResult::localTransportUnavailable (agent_sync_protocol_types.hpp) for the full doc.
     bool local_transport_unavailable;
+    /// @brief True when at least one block of queued items was actually sent to the manager.
+    /// Lets a module log "nothing to send" instead of "finished successfully" for an empty
+    /// cycle, which `success` alone cannot express. Never use it to gate durable state -- see
+    /// SyncModuleResult::sentAnything (agent_sync_protocol_types.hpp) for why.
+    bool sent_anything;
 } SyncModuleResult_t;
 
 /// @brief Defines the type of modification operation.
--- a/src/shared_modules/sync_protocol/include/agent_sync_protocol_types.hpp
+++ b/src/shared_modules/sync_protocol/include/agent_sync_protocol_types.hpp
@@ -66,4 +66,25 @@ struct SyncModuleResult
     /// a caller that records something about the session that was sent must not do so for a
     /// call that sent nothing.
     bool sessionSkipped{false};
+
+    /// @brief True when at least one block of queued items was actually sent to the manager.
+    ///
+    /// @ref success alone cannot distinguish "the manager accepted everything" from "there was
+    /// nothing queued, so no session was opened at all": an empty queue takes an early break and
+    /// returns the initial `true`. That is the intended contract -- a routine "nothing changed"
+    /// cycle is not a failure -- so this field carries the distinction instead, letting a module
+    /// log "nothing to send" rather than "finished successfully".
+    ///
+    /// It must NOT be used to gate durable state. A marker that records "the manager has this
+    /// agent's data" has to be gated on something that proves a round trip, such as
+    /// @ref IAgentSyncProtocol::notifyDataClean; gating it on a sync result instead is what lets
+    /// an untransmitted cycle be recorded as a completed one.
+    ///
+    /// Distinct from @ref sessionSkipped: that one means no session ran at all because another
+    /// was already in flight, while this one is about whether the session that did run carried
+    /// anything.
+    ///
+    /// Always false for metadata and group synchronizations, which send no queued items by
+    /// construction.
+    bool sentAnything{false};
 };
--- a/src/shared_modules/sync_protocol/src/agent_sync_protocol.cpp
+++ b/src/shared_modules/sync_protocol/src/agent_sync_protocol.cpp
@@ -15,6 +15,8 @@
 #include "metadata_provider.h"
 
 #include <flatbuffers/flatbuffers.h>
+#include <cerrno>
+#include <cstdlib>
 #include <cstring>
 #include <memory>
 #include <set>
@@ -69,6 +71,41 @@ void AgentSyncProtocol::setSessionMaxBytes(size_t maxBytes)
     {
         s_sessionMaxBytes.store(maxBytes);
     }
+}
+
+long AgentSyncProtocol::currentAgentId()
+{
+    agent_metadata_t metadata {};
+
+    if (metadata_provider_get(&metadata) != 0)
+    {
+        // No metadata published yet (agent-info not up, or agentd has not run
+        // w_agentd_populate_metadata()). "Unknown", not "changed".
+        return 0;
+    }
+
+    long id = 0;
+
+    // Ids are validated by OS_IsValidID() before client.keys is written: digits only, at most
+    // 8 characters, so the value always fits a long and strtol cannot overflow here. Parse
+    // defensively anyway -- anything that is not a plain number reads as unknown, never as a
+    // new identity. Zero-padding ("001") is presentational; the manager compares ids
+    // numerically too (fullSessionValidator.cpp).
+    if (metadata.agent_id[0] != '\0')
+    {
+        char* end = nullptr;
+        errno = 0;
+        const long parsed = std::strtol(metadata.agent_id, &end, 10);
+
+        if (errno == 0 && end != metadata.agent_id && *end == '\0' && parsed > 0)
+        {
+            id = parsed;
+        }
+    }
+
+    metadata_provider_free_metadata(&metadata);
+
+    return id;
 }
 
 AgentSyncProtocol::AgentSyncProtocol(const std::string& moduleName, std::optional<std::string> dbPath, LoggerFunc logger,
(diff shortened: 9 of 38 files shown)


jotacarma90:
Heads-up: #38656 is about to be merged and it lands on the same two spots as this PR, so a rebase will be needed. Two of the three overlaps do not show up as conflicts, which is why I am flagging them here.

**1. `SyncModuleResult` grows a field on both sides, and one of the initializers is positional.** #38656 appends `sessionSkipped` (a call that ran no session at all because another synchronization was already in flight); this PR appends `sentAnything`. The struct itself conflicts, so that half is visible. What is not: `agent_sync_protocol.cpp` merges cleanly, and this PR's `return {success, std::move(failureReason), stopped, managerNotReady, consecutiveFailures, awaitingPrerequisite, false, sentAny};` is positional, so the 8th value binds to whatever ends up 8th in the struct. Both new fields are `bool`, so resolving the struct with `sessionSkipped` first compiles without a single warning and silently feeds `sentAny` into `sessionSkipped`. In #38656 that inverts the VDFirst marker gating (`vdResult.success && !vdResult.sessionSkipped`): the marker would stop being persisted precisely on the cycles that did send something. Switching that return to designated initializers removes the hazard regardless of merge order.

**2. `syscollectorImp.cpp` conflicts for real**, in the VD sync block: #38656 extracts the option selection into `synchronizeVDTables()` (now used by both the periodic sync and the flush) and gates the marker there, while this PR reworks the same region for the recovery routing.

**3. The design question behind both.** `sentAnything`'s own documentation says durable state must not be gated on a sync result and has to rely on something that proves a round trip. #38656 does gate `vd_first_sync_completed` on a sync result (`success && !sessionSkipped`). Both PRs are fixing the same class of bug, an untransmitted cycle recorded as a completed one, with two different rules, so it is worth agreeing on one rather than settling it inside a conflict resolution.

Separately, one correction to the description, because it feeds the marker reasoning: "the only thing `VDFIRST` adds is an agent-scoped `deleteByQuery` before the scan". It also selects the first-scan chain, which omits `TEventSendReport`, so alerts are suppressed (`FirstFullScan` vs `FullScan` in `scanOrchestrator/factoryOrchestrator.hpp`, and the "no security alerts generated (alerts are suppressed on first scan)" log in `scanOrchestrator.hpp`). An agent whose first real scan arrives through `/scan/vd` rather than through a VDFirst session therefore gets one alert per pre-existing CVE.

cborla (approved):
LGTM
- [x] Code changes reviewed

PR #38602, opened 2026-08-26
remoted: authenticate agents and enrollment with JWT bearer profiles (wazuh-agent+jwt / wazuh-enroll+jwt)

juliancnn:
## Description

`wazuh-manager-remoted`'s HTTPS agent API (port 1517) authenticated every request with a custom
AES-CMAC scheme (`Authorization: Wazuh <id>:<ts>:<mac>` over a canonical byte sequence), and
`POST /enroll` with a sibling `WazuhEnroll` CMAC keyed by the enrollment password. Neither scheme
had shipped in a release; both were bespoke framing with their own parser, canonicalization rules
and interop surface.

This PR replaces them with a standard credential: an **HS256 JWT bearer** (RFC 7515 compact
serialization, RFC 6750 bearer usage) in two **closed profiles** — nothing on the wire is
negotiable, a token is either exactly the profile or it is rejected:

- **`wazuh-agent+jwt`** — every authenticated route. Header exactly `{alg:HS256, kid, typ}`, claims
  exactly `{exp, iat, iss, jti, nbf, sub}` (60-second lifetime, fresh `jti` per request, no `aud`:
  the `typ` fixes the domain), signed with the 32-byte key the agent's 64-hex `client.keys` secret
  decodes to.
- **`wazuh-enroll+jwt`** — `POST /enroll` in Password mode. Header exactly `{alg, typ}` (no `kid`),
  claims exactly `{exp, iat, jti, nbf}` (no identity yet), key = HKDF-SHA256 of the enrollment
  password (`etc/authd.pass`).

The token authenticates identity and freshness; it deliberately does **not** bind the HTTP method,
target or body — TLS protects them (remoted has no plaintext listener). A proxy-side path rewrite is
therefore a `404`, never a `401`. Every credential `401` carries `WWW-Authenticate: Bearer`.

One shared header-only library (`src/shared_modules/utils/jwt/`) signs on the agent and verifies on
the manager; the Go benchmark simulator and the Python tools reproduce it with their standard
libraries only, and all of them are pinned to the same frozen test vectors.

Closes #38582

## Proposed Changes

**Shared library — `src/shared_modules/utils/jwt/` (new, header-only, `jwt_profile::v1`)**
- Profile constants and `TimePolicy` (accepted age 1..60 s, clock skew 0..30 s, fail-fast on
  out-of-range), `CanonicalAgentId`, `SecureBytes` (wiped on destruction), `JwtKeyDecoder`
[...]

PR #38628, opened 2026-08-27
Fix startup reporting, control-plane error propagation, option validation and inventory-sync bounds

### LINKED ISSUE #38549: HTTPS/inventory-sync: small correctness defects found during 5.0.0 readiness testing
## Description

>[!IMPORTANT]

---

---

## Notes

---

## Tracking

### Left after #38727

**Elsewhere**

**Findings 63 to 65**

ignaciogalle12git:
## Description

Closes part of #38549.

The readiness round produced 53 findings. This PR carries the ones whose fix is small and mechanical, needs no design decision and no other team, and whose evidence is reachable with a unit test, a `-t` run or a daemon start. The other 30 are routed to eight issues and one tuning note, listed in the organizational comment on #38549.

Three findings were filed with a diagnosis that does not hold, and the corrected cause changed the fix:

- **44** blamed `merror_exit()` for not terminating. It does (`debug_op.c:577`). The hang is in the control script: `agentd` forks before it validates, so the parent already returned 0 and the wait loop polls a PID that will never appear, with a budget above the unit's `TimeoutSec`.
- **11** named `wazuhDBClient.cpp`, which has no `m_stopping` branch. `taskClient.cpp` has the defect; `authdClient.cpp` has half of it, its metric already separating the two cases and its log line not.
- **3** claimed two documents of one agent can share a `checksum.hash.sha1`. They cannot: producers hash the whole item JSON, identity fields included. Closed as invalid, with the invariant written down.

## Proposed Changes

| # | Commit | Findings | Area |
|---|---|---|---|
| 1 | `fix: report a refused configuration instead of timing out` | 44, 33 | `src/init/` |
| 2 | `fix: resolve the service account from the running process` | 49 | `shared_modules/utils` |
| 3 | `fix: stop the control plane reporting downstream failures as success` | 9 (partly), 10, 11, 18, 38, 41 | `remoted/remoted_module` |
| 4 | `fix: validate remoted's own configuration surface` | 12, 23, 24, 45 | `src/remoted/src`, `remotedModuleFacade.hpp` |
| 5 | `fix: bound the inventory-sync session query and its reporting` | 5, 28, 34, 47 | `inventory_sync_server`, `uds_http_server` |
| 6 | `fix: stop one malformed document from freezing Active Response` | 2 | `framework/wazuh/core/indexer` |
| 7 | `docs: correct the throttle mechanics and five other references` | 3, 6, 13, 20, 22, 26, 40, 42, 48 | `docs/ref`, `etc/`, `tools/` |

Three notes for the reviewer:

- **Commit 1 is first on purpose.** It changes the failure mode of every daemon start on both sides, and two of the issues routed out of here make more values refusable. Until a refused value fails cleanly, refusing more of them makes the operator experience worse.
- **Everything here was exercised on a running stack**, not only in CI: an installed manager from this branch, an OpenSearch indexer and an enrolled 5.x agent, in the manager devcontainer. The per-commit evidence below says what was driven and what was not.
- **Finding 9 ships half.** The fabricated `default` group list is fixed. Its other half, not re-querying wazuh-db on every notify while it is down, was implemented and backed out: stamping the refresh window as fresh on failure trades a retry storm for membership that stays stale for a whole interval, up to an hour at the maximum setting. That is a retry-cadence policy, so it goes to an issue and the branch keeps the `5.0.0` behaviour.
- **Finding 18 is a client-layer correction, not a new surface.** `globalQuery()` now maps a wazuh-db `err` reply to `ProtocolError`, mirroring what `getAgentGroups()` already does. The five `update-*` callers still discard the result, deliberately: those writes are fire-and-forget, and a callback reaching back into the handler would capture `this` and could outlive it on shutdown. What the caller should be told, and how, is the subject of one of the issues.

### Results and Evidence

The framework suite and the two shell halves run on a workstation. Everything else was built and run in the manager devcontainer, with `make -C src TARGET=manager` green. Each before/after pair is the same build tree with one file reverted to the base commit and only the affected target relinked.

<details><summary><b>Commit 1</b> · agent start 61 s to 1 s, and <code>status()</code> tells a refusal from a stop</summary>

Measured on a real 5.x agent, with `agent.https_backoff_cap=100` (below its own minimum of 1000):

```
control script from the 5.0.0 nightly:   61 s   wazuh-agentd did not start
control script from this branch:         40 s   wazuh-agentd did not start
healthy start, same script:               4 s   Completed.
```

61 s is above the unit's `TimeoutSec=45`, so systemd cut the start off first and reported `Result: timeout` instead of the daemon that did not come up. 40 s is below it. The budget is cumulative across daemons on purpose: a per-daemon counter adds up past the unit's limit however small each one is.

An earlier revision also tried to fail fast, by having `pstatus()` flag a PID file left by a process that is already gone. A harness showed 61 s to 1 s; on a real agent it never fires, and it cannot: `StartSIG()` registers `atexit(HandleExit)`, `HandleExit()` calls `DeletePID()`, and `merror_exit()` ends in `exit(1)`, so the daemon removes its own PID file on the way out and the loop finds nothing (`stale=false pidfiles=0` on every iteration). The harness stub left the file behind, which no real daemon does. Those twelve lines are gone rather than kept as dead code.

Manager half, with `wazuh-manager-modulesd` refusing `-t`:

```
before:  wazuh-manager-modulesd not running...                {"status":"stopped"}
after:   wazuh-manager-modulesd refused its configuration...  {"status":"failed"}
```

`failed` is not a new value: `get_manager_status()` (`framework/wazuh/core/cluster/utils.py:88`) already reads the same `var/run/<daemon>.failed` marker. The shell was the only side ignoring it. Review also found the marker was cleared too late to be trustworthy, so it is now cleared where it is written, in `testconfig()`:

```
before:  remoted refused its configuration... / db refused its configuration...   (db was fixed)
after:   remoted refused its configuration... / db not running...
```
</details>

<details><summary><b>Commit 2</b> · 4 new tests pass; 3 of them fail against the header as it stands</summary>

`certHelper.hpp` is header-only, so the suite compiles standalone. This host has no `wazuh` account, which is the state of a 5.x manager, so the control is exact:

```
after:   [  PASSED  ] 4 tests.
before:  Actual: it throws std::runtime_error with description "Could not get the user and group information."
         3 FAILED TESTS
```

That message is the one the reporter saw, wrapped into `Could not initialize Indexer Connector` and then `exit(EXIT_FAILURE)`. The only input needed to reach it is a second `<ca>`: `indexerTransport.cpp:45` gates the merge on `filePaths.size() > 1`.
</details>

<details><summary><b>Commit 3</b> · 689/689 in <code>remoted_module_utest</code></summary>

```
[==========] 689 tests from 92 test suites ran.
[  PASSED  ] 689 tests.
```

Five new cases: a notify with no cached groups answering 500 instead of a fabricated `default`; a failed group query leaving the cached membership and its timestamp untouched; the first notify after a startup writing through the throttle; and an `err ...` from an `update-*` surfacing as `ProtocolError`, with its positive control.

One of them failed on the first CI run and it was the test, not the fix: the canned wazuh-db answer was `ok {"group":"g1"}`, an object, and `getAgentGroups()` reads the group only out of an array, so it parsed to no groups and cached `default`. Worth knowing while reading that file: three pre-existing tests use the object form and pass, but only because they assert on `default`, which is what the empty parse falls back to.
</details>

<details><summary><b>Commit 4</b> · <code>remoted -t</code> now refuses the 40 options resolved in <code>secure.c</code></summary>

```
before:  wazuh-manager-remoted -t   # control_wdb_max_queue_size=99        -> rc=0
         wazuh-manager-remoted -t   # control_wdb_max_queue_size=1000001   -> rc=0
         wazuh-manager-remoted -t   # control_keepalive_throttle=900       -> rc=0, silent

after:   CRITICAL: (2302): Invalid definition for remoted.control_wdb_max_queue_size: '99'.       rc=1
         CRITICAL: (2302): Invalid definition for remoted.control_wdb_max_queue_size: '1000001'.  rc=1
         WARNING: 'remoted.control_keepalive_throttle' (900 s) is at or above
                  <agents_disconnection_time> (900 s): agents that are answering normally will be
                  marked disconnected. Keep it below half of it.                                  rc=0
```

Both out-of-range values are outside the option's own `100..1000000` range and `-t` took them anyway. The throttle case is a warning, not a refusal: the combination is a misconfiguration, not an invalid value. A valid value and a throttle of 450 both come back `rc=0` and silent.
</details>

<details><summary><b>Commit 5</b> · <code>modulesd -t</code> now bounds the session query batch size</summary>

```
before:  session_query_batch_size=20000  -> rc=0
         session_query_batch_size=1      -> rc=0

after:   CRITICAL: (2302): Invalid definition for ...session_query_batch_size: '20000'.   rc=1
         CRITICAL: (2302): Invalid definition for ...session_query_batch_size: '1'.
                   Use 0 for the module default, or a value between 100 and 10000.        rc=1
         session_query_batch_size=100    -> rc=0
         session_query_batch_size=1000   -> rc=0
```

The ceiling rides on the option's own range. The floor cannot: `0` is the module's "use my default" sentinel and has to stay valid, so a range starting at 100 would break it. The explicit check reuses the same `INV_DEF` format, so both carry the code a log parser keys on.
</details>

<details><summary><b>Commit 6</b> · 58/58 in the framework suite; 8 of 9 new cases fail against the current code</summary>

Environment from `5_testunit_framework.yml`. `AR_SCHEMA` constrains `wazuh` and nothing else, so a document that validates can carry `event` as null, a string, a number, a list, or with unusable `index`/`doc_id` values. Against `active_response.py` as it stands:

```
active_response.py:341: KeyError: 'event'
active_response.py:446: UnboundLocalError: cannot access local variable 'event_id'
active_response.py:351: TypeError: unhashable type: 'dict'
active_response.py:449: AttributeError: 'str' object has no attribute 'get'
```

`active_response_processing()` catches only `IndexerUnavailableError`, so any of these aborts the cycle before `dispatch()` and therefore before the bookmark moves, and the same page is re-fetched 30 s later and fails again. The values that raise nothing at all are worse: a null or empty `index` reaches `client.mget` and comes back a 400. The keys are now checked before they reach the dictionary rather than caught after, and the tests assert the query actually issued, `client.mget.assert_awaited_once_with(index="idx", body={"ids": ["1"]})`, because a mocked `mget` swallows a poisoned index otherwise.
</details>

<details><summary><b>Commit 7</b> · 42 of 42 benchmark scenarios resolve, was 41</summary>

```
before:  42 scenarios, 41 with every payload present, 1 unresolved
           mixed_fleet_windows_linux_only_stateless.json -> ../sample_payloads/engine/test_logs_merged.txt
after:   42 scenarios, 42 with every payload present, 0 unresolved
```

That payload has never been committed on any branch. It is repointed at `engine/syslog.log`, which its own twin `mixed_fleet_windows_linux` uses, and the `location` and description follow.
</details>

<details><summary><b>Runtime validation on a live stack</b> · installed manager, OpenSearch indexer, enrolled 5.x agent</summary>

CI proves it compiles and the suites pass. This is the other half: the product behaving as claimed, in the manager devcontainer, with the manager installed from this branch, an indexer in green and an agent enrolled as `001`.

**Commit 2**, two `<ca>` configured on a manager with no `wazuh` account, which is the reporter's exact case: zero occurrences of `Could not get the user and group information`, zero indexer-connector init failures, analysisd up, and

```
-rw-r----- 1 wazuh-manager wazuh-manager 2408  var/tmp/root-ca-merged.pem
```

2408 bytes is the two 1204-byte CAs concatenated, owned by the running account.

**Commit 4**, the two budget warnings, which only fire when the module starts:

```
WARNING: Endpoint '/control': the downstream timeouts add up to 60000 ms, which exceeds 'http_request_timeout' (30000 ms)
WARNING: Endpoint '/scan/vd': the downstream timeouts add up to 10000 ms, which exceeds 'http_request_timeout' (5000 ms)
```

The throttle warning also fires on a real start, not just under `-t`.

**Commit 5**, the transport startup line, at the shipped connection cap:

```
inventory sync server bound to '...' (... max 1024 connection(s), 64 reserved for control, 960 data session(s) ...)
```

A reserve of 64 against a ceiling of 256 is exactly the case the clamp warning never reported, because the clamp did not fire.

**Commit 3, finding 38.** With `control_keepalive_throttle=3600`, restarting the agent moves `remoted.control.startup` from 1 to 2 and advances `last_keepalive` 15 s later. Under a 3600 s throttle that write is only possible through the post-startup bypass. Sampling `connection_status` once a second never catches `pending`, which is the point of the fix.

**Commit 6**, against the real indexer, with a real `mget` rather than a mocked one. Seven malformed `event` shapes, one page each alongside a good document:

```
event ausente / null / string / index null / index vacio / doc_id null / doc_id no hasheable
  -> in all seven, the good AR survives and its event is fetched
```

**Not driven:** findings 9, 10, 11, 18 and 41 change a log line or a metric each and would need wazuh-db taken down mid-flight. Findings 15, 17 and 51 need a cluster and stay out of this PR anyway.
</details>

  - [x] Linux (`make -C src TARGET=manager`, devcontainer, plus CI's Unit-Tests (manager))
  - [ ] Windows
  - [ ] MAC OS X
- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows / macOS: N/A, nothing platform-specific is touched
- Engine _(Wazuh v5.x and above)_: N/A, no engine code is touched
- Wazuh server API/Framework

### Artifacts Affected

`wazuh-manager-remoted`, `wazuh-manager-modulesd`, `wazuh-manager-analysisd` (through `certHelper.hpp`), `wazuh-agentd` (through the control script only), both control scripts, the framework, four reference documents, `etc/internal_options.conf` and one benchmark scenario.

### Configuration Changes

- `wazuh_modules.inventory_sync_server_session_query_batch_size`: the range becomes `0` (the module's "use my default" sentinel) or `100..10000`, from `0..100000`. Default stays 1000. An operator who has raised `index.max_result_window` has to raise this ceiling too, which is noted in the reference.
- New startup warnings, no default changes: the keepalive throttle at or above `<agents_disconnection_time>`, and the `/control` and `/scan/vd` downstream budgets exceeding `remoted.http_request_timeout`.
- The agent start budget drops from 60 s per daemon to 40 s for the whole sequence, below the units' `TimeoutSec=45`. Off systemd, where nothing else caps the wait, a start that legitimately takes longer than that now fails. It is deliberate: under the shipped units nothing above 45 was ever reachable, and the budget is only the backstop for a daemon that never writes a PID file, since one that starts slowly is caught by the poll the moment its PID appears.

### Tests Introduced

Extensions to existing suites plus one new file. No new test binary.

`certHelper_test.cpp` (new): the merge, the ownership of the merged file, directory creation, the missing-certificate failure. `controlHandler_test.cpp` and `wazuhDBClient_test.cpp`: the five control-plane cases above. `test_secure.c`: the throttle warning, fired and not fired. `transportDiagnostics_test.cpp`: the resolved route classes in the transport startup line. `test_active_response.py`: nine parametrised cases over every JSON type `AR_SCHEMA` leaves unconstrained.

Not covered by a unit test: the `m_stopping` reject branch in `TaskClient`. Its only trigger is a worker callback re-entering during `stop()`, and reaching it from a test would mean calling into the object while its `unique_ptr` is being destroyed.

## Review Checklist

- [ ] Code changes reviewed

No changelog entry: this is code that exists only in the unpublished 5.0.0.

#38611 is open and edits `controlHandler.cpp`; #38602 replaces AES-CMAC with JWT across remoted and the agent. Whichever lands first, the other rebases.


jotacarma90:
Reviewed on top of the rebase. The branch is sound — I pushed three small commits (`81df4e6`, `10bf7ef`, `e6611ac`) for the mechanical part of what the review found, and I'm leaving the rest here as notes.

**Pushed:**

- `wazuh-server.sh`: with the marker wipe now at the top of `testconfig()`, the pre-existing `.restart` guard became a regression — on `restart` with a refused config the old marker was wiped and the new one suppressed, so `status` said plain "not running". The marker is now unconditional and carries the cause ("refused" vs "start"), so `status` no longer reports "refused its configuration" for a daemon that passed `-t` and died binding a port.
- `secure.c`: the throttle guard fired at `>= agents_disconnection_time`, but the staleness monitord compares is the throttle **plus the agent's notify interval** — a throttle of 899 against the default 900 still flaps answering agents and passed silently. It now warns from half, which is what the message itself already advised. Tests and the docs note adjusted.
- inventory-sync docs: the allowed range still said `0 to 100000`.

**Not pushed, for your call:**

- `w_remoted_validate_module_config()` reaches `remoted_enrollment_config()`, which reads `WAZUHCONF` hardcoded — so `-t -c candidate.conf` validates the `<auth>` block of the *installed* file. Pre-existing path bug, but this PR is what puts it under `-t`; probably worth a comment plus an issue rather than plumbing `-c` here.
- `VD_SCAN_BUDGET_MS` sums only `postScan()`'s read+write, but `getOffset()` runs synchronously before it in the same request path, so the startup check under-counts the endpoint's real budget.
- The new notify failure path shares `wdbErrorThrottle()` with `handleStartup` (one window, mixed totals) while tasks got their own throttle — worth an instance for consistency.
- Two small doc/contract gaps: notify's new `500 {"error":"database_error"}` isn't in the agent-facing contract, and the `/startup` reset of `lastKeepaliveUpdateSec` is a second, undocumented throttle exception.

jotacarma90:
Rebased onto today's `5.0.0` (`d6c2d2e`). Two things worth knowing:

- The `sessionProcessor.cpp` deletion log conflicted with #38671 — both sides had softened the same overclaiming line for different reasons (the async `/config`+`/stats` half vs. the refresh-interval visibility). The resolution keeps both: `Deletion of agent %s state documents completed (%s).` with both caveats in the comment.
- #38676 added `config_token` to the notify response; it composes fine with the new uncached-500 path (the token is built after the group resolution that the 500 short-circuits). While there, the contract gap noted earlier is closed: `agent-api.yaml`'s control `500` now covers notify's uncached case, not just startup (`49efd39`).

jotacarma90 (approved):
LGTM

PR #38634, opened 2026-08-27
Fix out-of-bounds read when a quoted CSV/DSV field ends the input

### LINKED ISSUE #38596: `hlp::getField()`: bounds-check the look-ahead in the closing-quote branch
## Description

`hlp::getField()` (`src/engine/source/hlp/src/parsers/parse_field.cpp`) reads one character past the closing quote when deciding whether the quote really closes the field:

```cpp
else  // Handle closing quote
{
    ...
    quote_opened = escaped || (input[i + 1] != delimiter);
    isQuoted     = true;
}
```

When the closing quote is the last character of the view (`i == input.size() - 1`), this evaluates `input[input.size()]`. `std::string_view::operator[]` requires `pos < size()`, so the read is undefined behaviour — the type carries no NUL-terminator guarantee.

The call site in the DSV/CSV parser (`src/engine/source/hlp/src/parsers/dsv_csv.cpp:80`) passes `parsed.substr(start, parsed.size() - start)`, and `parsed` is obtained zero-copy from the event document via `Json::getString<std::string_view>` (`src/engine/source/base/include/base/json.hpp:524`), so any DSV-parsed value whose quoted field ends at the end of the parsed view reaches this line — e.g. `k,"v"`.

Today the byte is harmless: rapidjson allocates `length + 1` and NUL-terminates, so it is either the string's own terminator or another character of the same event, and either way the value cannot change the outcome — the read only happens on the final iteration, and the only later use of `quote_opened` is the trailing `quote_opened && input.back() != quote`, whose second condition is false because the last character is that same closing quote. So this is a correctness fix, not a behaviour fix: the function's signature accepts any `std::string_view`, and it should not depend on a terminator that the type does not promise.

## Suggested fix

Short-circuit the look-ahead on the bounds check:

```cpp
quote_opened = escaped || (i + 1 >= input.size() || input[i + 1] != delimiter);
```

For every input this yields the same result the current code computes, so no caller changes behaviour and the trailing unclosed-quote check needs no adjustment.

A regression test in `src/engine/source/hlp/test/src/unit/dsv_csv_test.cpp` covering a quoted final field would pin the behaviour, but note it cannot catch the defect itself: the harness backs the view with a `std::string`, whose `data()[size()]` is also a valid NUL.

## Credit

Thanks to @ZeroX-404 for pointing this out.

vikman90:
## Description
`hlp::getField()` looked ahead at `input[i + 1]` while handling a closing quote without checking that `i + 1` was still inside the `std::string_view`. When the closing quote was the last character of the view, the read went one byte past its end. The fix bounds-checks the look-ahead so the byte is only read when it belongs to the view.

## Proposed Changes
- `src/engine/source/hlp/src/parsers/parse_field.cpp`: guard the closing-quote look-ahead with `i + 1 >= input.size()` so `input[i + 1]` is only evaluated when in range. Behaviour is unchanged: when the closing quote is last, the quote is treated as still open, which is what the previous code computed from the byte that followed.
- `src/engine/source/hlp/test/src/unit/parse_field_test.cpp`: new unit tests for `getField()` quote handling at the end of the input.
- `src/engine/source/hlp/CMakeLists.txt`: add the new test file to `hlp_utest` and give the target access to the `src/parsers/` headers it needs.

### Results and Evidence
The regression test places the input so that its last byte is the last readable byte before an unmapped page, which makes a read past the view fault instead of silently landing on an adjacent allocation.

Without the fix, `GetFieldTest.ClosingQuoteAtEndOfInput` crashes:

```
[ RUN      ] GetFieldTest.ClosingQuoteAtEndOfInput
Segmentation fault (core dumped)     exit=139
```

With the fix, the new tests pass:

```
[==========] Running 4 tests from 1 test suite.
[ RUN      ] GetFieldTest.ClosingQuoteAtEndOfInput
[       OK ] GetFieldTest.ClosingQuoteAtEndOfInput (0 ms)
[ RUN      ] GetFieldTest.EscapedQuoteAtEndOfInput
[       OK ] GetFieldTest.EscapedQuoteAtEndOfInput (0 ms)
[ RUN      ] GetFieldTest.ClosingQuoteFollowedByDelimiter
[       OK ] GetFieldTest.ClosingQuoteFollowedByDelimiter (0 ms)
[ RUN      ] GetFieldTest.UnquotedFieldAtEndOfInput
[       OK ] GetFieldTest.UnquotedFieldAtEndOfInput (0 ms)
[----------] 4 tests from GetFieldTest (0 ms total)
[  PASSED  ] 4 tests.
```

The full `hlp` suite passes with the fix applied, so CSV/DSV parsing results are unchanged:

```
[==========] 3076 tests from 60 test suites ran. (49 ms total)
[  PASSED  ] 3076 tests.
```

Only `GetFieldTest.ClosingQuoteAtEndOfInput` exercises the out-of-bounds read. The other three pass with and without the fix and are there as non-regression guards: the escaped-quote case short-circuits before the look-ahead, and the remaining two read in range.

### Artifacts Affected
`wazuh-engine` — the `hlp` static library is linked into it. No packaging changes.

### Configuration Changes
None.

### Documentation Updates
None.

### Tests Introduced
`GetFieldTest.ClosingQuoteAtEndOfInput`, `GetFieldTest.EscapedQuoteAtEndOfInput`, `GetFieldTest.ClosingQuoteFollowedByDelimiter` and `GetFieldTest.UnquotedFieldAtEndOfInput`, in `src/engine/source/hlp/test/src/unit/parse_field_test.cpp`.

## Credits
Reported by ZeroX-404.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/src/engine/source/hlp/CMakeLists.txt
+++ b/src/engine/source/hlp/CMakeLists.txt
@@ -56,7 +56,10 @@ add_executable(hlp_utest
   ${UNIT_SRC_DIR}/web_test.cpp
   ${UNIT_SRC_DIR}/kvmap_test.cpp
   ${UNIT_SRC_DIR}/dsv_csv_test.cpp
+  ${UNIT_SRC_DIR}/parse_field_test.cpp
 )
+
+target_include_directories(hlp_utest PRIVATE ${CMAKE_CURRENT_LIST_DIR}/src/parsers)
 
 target_compile_definitions(hlp_utest PRIVATE HLP_TEST_TZDB_PATH="${CMAKE_SOURCE_DIR}/external/tzdata")
 target_link_libraries(hlp_utest PRIVATE hlp GTest::gtest_main GTest::gtest)
--- a/src/engine/source/hlp/src/parsers/parse_field.cpp
+++ b/src/engine/source/hlp/src/parsers/parse_field.cpp
@@ -58,7 +58,7 @@ getField(std::string_view input, const char delimiter, const char quote, const c
                     bool escaped = (last_escape_location + 1 == i) && (1 < i);
                     isEscaped = isEscaped || escaped;
                     last_escape_location += (i - last_escape_location) * size_t(!escaped);
-                    quote_opened = escaped || (input[i + 1] != delimiter);
+                    quote_opened = escaped || (i + 1 >= input.size() || input[i + 1] != delimiter);
                     isQuoted = true;
                 }
             }
--- a/src/engine/source/hlp/test/src/unit/parse_field_test.cpp
+++ b/src/engine/source/hlp/test/src/unit/parse_field_test.cpp
@@ -1,0 +1,118 @@
+#include <gtest/gtest.h>
+
+#include <cstring>
+#include <stdexcept>
+#include <string_view>
+
+#include <sys/mman.h>
+#include <unistd.h>
+
+#include "parse_field.hpp"
+
+namespace
+{
+
+// Holds a string_view whose last byte is the last readable byte before an unmapped
+// page, so that any read at data() + size() raises SIGSEGV instead of silently
+// hitting an adjacent allocation.
+class GuardedInput
+{
+public:
+    explicit GuardedInput(std::string_view text)
+        : m_len {text.size()}
+    {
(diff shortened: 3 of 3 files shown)


jotacarma90 (approved):
LGTM

PR #38651, opened 2026-08-27
Validate active response usernames against an allowlist

### LINKED ISSUE #38631: Active response username validation is missing its regression guard
Two gaps in `src/unit_tests/active-response/test_active-response.c`, found while reviewing the 5.x port of #36745. Line numbers are for `4.14.9`; add the coverage there first, then forward-port.

## 1. `get_username_from_json()` has no test at all — Medium

The file has zero references to `get_username_from_json`. #36745 tested `is_valid_username()` thoroughly — 19 cases — but never asserted that the getter actually calls it.

That missing assertion has already cost us once. When the active response tree was restructured for 5.x, `get_username_from_json()` was reimplemented to read ECS `user.name` and the validation call was dropped. Nothing failed, because nothing tested the getter. The regression shipped and had to be found by an external report.

Two cases cover the whole surface:

```c
void test_get_username_from_json_rejects_invalid(void **state) {
    (void) state;
    cJSON *input = cJSON_Parse("{\"parameters\":{\"alert\":{\"data\":{\"dstuser\":\"--help\"}}}}");
    assert_non_null(input);
    assert_null(get_username_from_json(input));
    cJSON_Delete(input);
}

void test_get_username_from_json_accepts_valid(void **state) {
    (void) state;
    cJSON *input = cJSON_Parse("{\"parameters\":{\"alert\":{\"data\":{\"dstuser\":\"alice\"}}}}");
    assert_non_null(input);
    assert_string_equal(get_username_from_json(input), "alice");
    cJSON_Delete(input);
}
```

Register both in `main()`. The returned pointer aliases the cJSON tree, hence asserting before `cJSON_Delete`. On `5.x` the JSON shape is `{"user":{"name":"..."}}` instead.

This is the highest-value item of the three review issues: it is the regression guard for the whole class, and it costs minutes.

## 2. `\n` and `\r` rejection is untested — Low

`test_active-response.c:196`

```c
void test_is_valid_username_invalid_with_whitespace(void **state) {
    (void) state;
    assert_int_equal(is_valid_username("test user"), 0);
    assert_int_equal(is_valid_username("test\tuser"), 0);
}
```

`is_valid_username()` rejects four whitespace characters; only two are asserted. Add `"test\nuser"` and `"test\ruser"`.

## Scope

Test debt, not a product defect. Item 1 is nonetheless the direct cause of the 5.x regression recurring.

vikman90:
## Description
`is_valid_username()` enumerated forbidden characters, so everything it did not list was accepted: quotes,
shell metacharacters, control characters and non-ASCII bytes. It now accepts only the portable username
character set plus `$` for machine accounts.

## Proposed Changes
- `src/active-response/active_responses.c`: accept only `[A-Za-z0-9._-]` and `$`, and keep rejecting a
  leading `-`. Explicit ASCII ranges are used rather than `isalnum()`, which is locale-dependent and is
  undefined for a negative `char`.
- `src/active-response/active_responses.h`: document the accepted set and the `@` exclusion.
- `src/unit_tests/active-response/test_active-response.c`: cases for shell metacharacters, control
  characters, non-ASCII bytes, qualified names and `+`/`~` anywhere in the name.

Stacked on #38637, which rewrites part of the same function. Review that one first.

### Results and Evidence
Verdicts of the validator before and after, both versions extracted from their branch and exercised
side by side:

```
  input                before     after
  test;user            accepted   rejected
  test|user            accepted   rejected
  test&user            accepted   rejected
  test`id`             accepted   rejected
  test$(id)            accepted   rejected
  test'user            accepted   rejected
  test"user            accepted   rejected
  test*user            accepted   rejected
  test>user            accepted   rejected
  alice@example.com    accepted   rejected
  test+user            accepted   rejected
  test~user            accepted   rejected
  test<0x01>user       accepted   rejected
  café                 accepted   rejected
  artest               accepted   accepted
  machine$             accepted   accepted
  test.user            accepted   accepted
  john..doe            accepted   accepted
  sys-tem_1            accepted   accepted
```

Against the installed binary, `/var/ossec/active-response/bin/disable-account`, rejected names never reach
`passwd` — `logs/active-responses.log` shows `Cannot read 'dstuser' from data or invalid username format`
and the exit status is 255:

```
  test;user              exit=255  rejected by validator
  test$(id)              exit=255  rejected by validator
  alice@example.com      exit=255  rejected by validator
  machine$               exit=255  passed to passwd
  test.user              exit=255  passed to passwd
```

The last two exit 255 only because those accounts do not exist on the host; the validator passed them
through, which is the behaviour being checked.

Regression check on a real account:

```
initial : P
add  -> exit=0 state=L
del  -> exit=0 state=P
```

### Behaviour change
Names containing `@`, `+`, `~` or any non-ASCII byte are no longer accepted. `DOMAIN\user` was already
rejected before this change. `@` is excluded deliberately: `passwd -l` and `chuser` act on local accounts,
which cannot carry a realm suffix, so a fully qualified name could never have been locked — it previously
failed at `passwd` instead of at validation. A test pins that decision.

### Artifacts Affected
`active_responses.c` is linked into every active response binary, but `is_valid_username()` has a single
caller, `get_username_from_json()`, used only by `disable-account`.

### Configuration Changes
None.

### Documentation Updates
None.

### Tests Introduced
`src/unit_tests/active-response/test_active-response.c` — five cases covering shell metacharacters, `+`/`~`
anywhere, qualified names, control characters and non-ASCII bytes.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/src/active-response/active_responses.c
+++ b/src/active-response/active_responses.c
@@ -307,9 +307,11 @@ static cJSON* get_srcip_from_win_eventdata(const cJSON *data) {
     return NULL;
 }
 
-// Accepts only the portable username character set, plus '$' for machine accounts.
-// Anything outside it is rejected, so no separator, quote, whitespace, control or
-// non-ASCII byte can reach the command line the username is placed on.
+// Accepts only the portable username character set, plus '$' anywhere for machine
+// accounts. Anything outside it is rejected, so no separator, quote, whitespace,
+// control or non-ASCII byte reaches the argument vector the username is placed in.
+// '@' is deliberately excluded: passwd and chuser act on local accounts, which
+// cannot carry a realm suffix, so a qualified name could never be locked anyway.
 int is_valid_username(const char *username) {
     if (!username || !*username) {
         return 0;
--- a/src/active-response/active_responses.c
+++ b/src/active-response/active_responses.c
@@ -307,8 +307,9 @@ static cJSON* get_srcip_from_win_eventdata(const cJSON *data) {
     return NULL;
 }
 
-// The rejected set assumes the username is passed as a discrete argv element,
-// never interpolated into a path or a shell command line.
+// Accepts only the portable username character set, plus '$' for machine accounts.
+// Anything outside it is rejected, so no separator, quote, whitespace, control or
+// non-ASCII byte can reach the command line the username is placed on.
 int is_valid_username(const char *username) {
     if (!username || !*username) {
         return 0;
@@ -326,32 +327,16 @@ int is_valid_username(const char *username) {
         return 0;
     }
 
-    // Must not start with dash, plus, or tilde (Debian constraint)
-    if (username[0] == '-' || username[0] == '+' || username[0] == '~') {
+    // A leading dash would be read as an option by the command the name is passed to
+    if (username[0] == '-') {
         return 0;
     }
 
-    // Check for prohibited characters
     for (size_t i = 0; i < len; i++) {
         char c = username[i];
 
-        // Reject colon (used as field separator in /etc/passwd)
-        if (c == ':') {
-            return 0;
-        }
-
-        // Reject comma (used in GECOS field)
-        if (c == ',') {
-            return 0;
-        }
-
-        // Reject whitespace characters
-        if (c == ' ' || c == '\t' || c == '\n' || c == '\r') {
-            return 0;
-        }
-
-        // Reject slash (breaks home directory paths)
-        if (c == '/' || c == '\\') {
+        if (!(c >= 'a' && c <= 'z') && !(c >= 'A' && c <= 'Z') && !(c >= '0' && c <= '9') &&
+            c != '.' && c != '_' && c != '-' && c != '$') {
             return 0;
         }
     }
--- a/src/active-response/active_responses.h
+++ b/src/active-response/active_responses.h
@@ -92,8 +92,9 @@ const char* get_username_from_json(const cJSON *input);
 
 /**
  * Validate username format
- * Accepts only [A-Za-z0-9._-] plus '$' for machine accounts, and rejects a
- * leading dash so the name cannot be read as an option.
+ * Accepts only [A-Za-z0-9._-] plus '$' anywhere for machine accounts, and rejects
+ * a leading dash so the name cannot be read as an option. '@' is excluded: only
+ * local accounts can be locked, and those cannot carry a realm suffix.
  * Rejects: reserved names (root), empty strings, null, names over 256 bytes
  * @param username Username to validate
  * @retval 1 If username is valid
--- a/src/active-response/active_responses.h
+++ b/src/active-response/active_responses.h
@@ -92,11 +92,9 @@ const char* get_username_from_json(const cJSON *input);
 
 /**
  * Validate username format
- * Follows Debian adduser constraints:
- * - Must not start with dash, plus, or tilde
- * - Must not contain colon, comma, or whitespace
- * - Should not contain slash (may break home directory paths)
- * Rejects: reserved names (root), empty strings, null
+ * Accepts only [A-Za-z0-9._-] plus '$' for machine accounts, and rejects a
+ * leading dash so the name cannot be read as an option.
+ * Rejects: reserved names (root), empty strings, null, names over 256 bytes
  * @param username Username to validate
  * @retval 1 If username is valid
  * @retval 0 If username is invalid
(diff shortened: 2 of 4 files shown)


src/active-response/active_responses.c
```diff
@@ -326,32 +329,16 @@ int is_valid_username(const char *username) {
...
-
-        // Reject whitespace characters
-        if (c == ' ' || c == '\t' || c == '\n' || c == '\r') {
-            return 0;
-        }
-
-        // Reject slash (breaks home directory paths)
-        if (c == '/' || c == '\\') {
+        if (!(c >= 'a' && c <= 'z') && !(c >= 'A' && c <= 'Z') && !(c >= '0' && c <= '9') &&
+            c != '.' && c != '_' && c != '-' && c != '$') {
```
MarcelKemp: **[S3 · non-blocking · introduced]** `$` is accepted at any position, which leaves the one shell-active byte in an allowlist whose stated purpose is to remove them. The justification in the comment above - Samba/AD machine accounts - only ever needs a trailing `$`, but the predicate also admits `$PATH`, `a$b$` and a name that is nothing but `$`. The assertion that looks like it covers this, `is_valid_username("test$(id)") == 0` at line 247, passes because of the parentheses - line 142 pins `test$user` as valid, so the CHANGELOG's "rejecting ... shell metacharacters" overstates by one character.

```suggestion
        if (!(c >= 'a' && c <= 'z') && !(c >= 'A' && c <= 'Z') && !(c >= '0' && c <= '9') &&
            c != '.' && c != '_' && c != '-' && !(c == '$' && i == len - 1 && i > 0)) {
```

Compiled and exercised: `machine$` and `WORKSTATION$` stay accepted, `test$user`, `$user`, `$`, `$$` and `a$b$` are rejected, everything else is unchanged. It does require flipping the assertion at `test_active-response.c:142`.

src/active-response/active_responses.h
```diff
@@ -92,11 +92,10 @@ const char* get_username_from_json(const cJSON *input);
...
 /**
  * Validate username format
- * Follows Debian adduser constraints:
- * - Must not start with dash, plus, or tilde
- * - Must not contain colon, comma, or whitespace
- * - Should not contain slash (may break home directory paths)
- * Rejects: reserved names (root), empty strings, null
+ * Accepts only [A-Za-z0-9._-] plus '$' anywhere for machine accounts, and rejects
+ * a leading dash so the name cannot be read as an option. '@' is excluded: only
+ * local accounts can be locked, and those cannot carry a realm suffix.
```
MarcelKemp: **[S3 · non-blocking · introduced]** This is now the second definition of a valid username in the tree and it disagrees with the first in both directions: `src/analysisd/analysisd.c:2202` gates every active-response dispatch on `OS_PRegex(lf->dstuser, "^[a-zA-Z._0-9@?-]*$")` (POSIX ERE, `src/shared/regex_op.c:24`), a class that excludes `$` and includes `@` and `?`. So `WORKSTATION$` - the machine-account case `$` is allowed for - is dropped by analysisd with `CRAFTED_USER` and never reaches the agent through an alert at all (only the API reaches it, since `framework/wazuh/core/active_response.py:245` passes the caller's `alert` dict straight through), while `alice@example.com` clears analysisd, is shipped to the agent, spawns `disable-account`, and is discarded only there. The repo already has the idiom for stating such a set once - `src/shared/agent_op.c:279` declares the literal and line 385 applies `strspn(x, valid_chars) != strlen(x)`.

```c
/* active_responses.h */
#define VALID_USERNAME_CHARS "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789._-"
```

Single-sourcing it here would let `analysisd` share the definition instead of re-deriving it as a regex. Not offered as a `suggestion` block because the other half of the fix is outside this diff.

src/active-response/active_responses.c
```diff
@@ -307,8 +307,11 @@ static cJSON* get_srcip_from_win_eventdata(const cJSON *data) {
     return NULL;
 }
 
-// The rejected set assumes the username is passed as a discrete argv element,
-// never interpolated into a path or a shell command line.
+// Accepts only the portable username character set, plus '$' anywhere for machine
+// accounts. Anything outside it is rejected, so no separator, quote, whitespace,
+// control or non-ASCII byte reaches the argument vector the username is placed in.
+// '@' is deliberately excluded: passwd and chuser act on local accounts, which
+// cannot carry a realm suffix, so a qualified name could never be locked anyway.
```
MarcelKemp: **[S3 · non-blocking · introduced]** The rewritten comment dropped the sentence that stated the caller's obligation - the old text warned that the set "assumes the username is passed as a discrete argv element, never interpolated into a path or a shell command line". The replacement asserts the consequence (which bytes do not get through) but no longer states what a caller must do with the bytes that do, and the header at `active_responses.h:95-98` loses the same caveat, so the obligation is now written nowhere. Whoever next writes an AR as `snprintf(cmd, "passwd -l %s", user)` plus `system(cmd)` will read a function documented as safe and inherit `$` expansion (see the note on line 340).

```suggestion
// Accepts only the portable username character set, plus '$' anywhere for machine
// accounts. Anything outside it is rejected, so no separator, quote, whitespace,
// control or non-ASCII byte reaches the argument vector the username is placed in.
// The accepted set is only safe under that assumption: callers must pass the name as
// a discrete argv element, never interpolated into a path or a shell command line.
// '@' is deliberately excluded: passwd and chuser act on local accounts, which
// cannot carry a realm suffix, so a qualified name could never be locked anyway.
```

MarcelKemp (dismissed):

MarcelKemp (approved):

Issue #36757, opened 2026-06-04
macOS 26 SCA false failure: PolicyBanner check 41062 fails although local rule conditions pass


jcwalker3:
<html><head></head><body>
Wazuh version | Component | Install type | Install method | Platform
-- | -- | -- | -- | --
v4.14.5-rc1 | Wazuh Agent / SCA | Agent | Packages | macOS 26.5.1 build 25F80

<h2>Description</h2><p>I am seeing what appears to be a false failure for macOS SCA check <code inline="">41062</code>, titled:</p><p><code inline="">Ensure a Login Window Banner Exists.</code></p><p>The issue is occurring with the macOS CIS policy:</p><p><code inline="">/Library/Ossec/ruleset/sca/cis_apple_macOS_26.x.yml</code></p><h2>Environment</h2><pre><code class="language-text">macOS ProductName: macOS
macOS ProductVersion: 26.5.1
macOS BuildVersion: 25F80

Wazuh version: v4.14.5
Wazuh revision: rc1
Wazuh type: agent
Policy file: /Library/Ossec/ruleset/sca/cis_apple_macOS_26.x.yml
Check ID: 41062
Check title: Ensure a Login Window Banner Exists.
</code></pre><p>Current Wazuh agent status:</p><pre><code class="language-text">wazuh-modulesd is running...
wazuh-logcollector is running...
wazuh-syscheckd is running...
wazuh-agentd is running...
wazuh-execd is running...
</code></pre><h2>Local policy rule</h2><p>The local policy rule for check <code inline="">41062</code> is:</p><pre><code class="language-yaml">- id: 41062
  title: "Ensure a Login Window Banner Exists."
  description: "A Login window banner warning informs the user that the system is reserved for authorized use only. It enforces an acknowledgment by the user that they have been informed of the use policy in the banner if required. The system recognizes either the .txt and the .rtf formats."
  condition: all
  rules:
    - "d:/Library/Security -&gt; r:^PolicyBanner"
    - 'c:stat -f %A /Library/Security/PolicyBanner.* -&gt; r:\d\d4'
[...]

PR #38679, opened 2026-08-27
Fix broken checks with command wrapper in multiple SCA files

### LINKED ISSUE #36784: Audit SCA `c:` command rules that rely on shell features without a shell wrapper
## Summary

While fixing #36757 we found that the macOS SCA check `41062` was a permanent false negative because its command rule relied on **shell glob expansion that the SCA engine never performs**.

The SCA engine does not execute `c:` commands through a shell: the command string is tokenized (`w_strtok`, `src/shared/string_op.c`) and run directly via `execvp()` (`src/wazuh_modules/wm_exec.c`). There is no `/bin/sh` in the path, so any shell-only feature in the command — globs (`*`, `?`, `[...]`), pipes (`|`), redirections (`>`, `<`), command substitution (`$(...)`, backticks), `~` expansion, `;`/`&&` sequencing, etc. — is **not** interpreted. The token is passed literally to the program, which then misbehaves or errors, and the check silently produces a wrong result.

The established convention in our ruleset is to wrap such commands in a shell, e.g. `c:sh -c "..."`. Rules that need shell features but omit the wrapper are latent false positives/negatives.

## Example (the bug we just fixed)

```yaml
# BROKEN — '*' is passed literally to stat, which fails → check can never pass
- 'c:stat -f %A /Library/Security/PolicyBanner.* -> r:\d\d4'

# FIXED — shell expands the glob
- 'c:sh -c "stat -f %A /Library/Security/PolicyBanner.*" -> r:\d\d4'
```

Verified on a live macOS 26 endpoint: the literal-glob form always returns *"No such file or directory"* and fails; the `sh -c` form returns `644` and passes. Fix applied in #36757 to the macOS 13.x / 14.x / 15.x / 26.x policies.

## What needs to be investigated

This is unlikely to be limited to the PolicyBanner check or to macOS — the same class of bug can exist in any SCA policy across any OS family. We are **not** providing an exhaustive list on purpose; please audit the rulesets and determine the real scope.

Suggested scope:

- Sweep all `c:` (and `not c:`) command rules in `ruleset/sca/` across every OS family (darwin, debian, ubuntu, rhel/centos/rocky/alma/amazon/oracle/sles, windows, applications, etc.).
- For each rule, look at the **command portion only** (everything before `->`; the part after `->` is a regex and a `*` there is expected and harmless).
- Flag any command that uses a shell-only feature — glob (`*`, `?`, `[...]`), pipe (`|`), redirection (`<`, `>`), command substitution (`$(...)` / backticks), `~` expansion, or `;` / `&&` sequencing — **without** being wrapped in `sh -c` / `bash -c`.
- For each candidate, confirm whether it is actually broken (it depends on the unsupported feature) vs. a false alarm (e.g. the metacharacter is literal data the program handles itself, or the binary does its own globbing). Reproduce on a real endpoint where possible.
- Apply the `sh -c` wrapper (or otherwise rewrite the rule to avoid the shell dependency) for each confirmed case, and verify the corrected result on a live endpoint.

## Acceptance criteria

- [ ] Full audit of `c:` command rules across all SCA OS families completed.
- [ ] Confirmed-broken rules identified and listed.
- [ ] Fixes applied and validated on real endpoints.
- [ ] Confirmation of whether this is widespread or isolated to the already-fixed macOS checks.

IsExec:
## Summary

While fixing #36757 we found that the macOS SCA check `41062` was a permanent false negative because its command rule relied on **shell glob expansion that the SCA engine never performs**.

The SCA engine does not execute `c:` commands through a shell: the command string is tokenized (`w_strtok`, `src/shared/string_op.c`) and run directly via `execvp()` (`src/wazuh_modules/wm_exec.c`). There is no `/bin/sh` in the path, so any shell-only feature in the command — globs (`*`, `?`, `[...]`), pipes (`|`), redirections (`>`, `<`), command substitution (`$(...)`, backticks), `~` expansion, `;`/`&&` sequencing, etc. — is **not** interpreted. The token is passed literally to the program, which then misbehaves or errors, and the check silently produces a wrong result.

The established convention in our ruleset is to wrap such commands in a shell, e.g. `c:sh -c "..."`. Rules that need shell features but omit the wrapper are latent false positives/negatives.

## Example (the bug we just fixed)

```yaml
# BROKEN — '*' is passed literally to stat, which fails → check can never pass
- 'c:stat -f %A /Library/Security/PolicyBanner.* -> r:\d\d4'

# FIXED — shell expands the glob
- 'c:sh -c "stat -f %A /Library/Security/PolicyBanner.*" -> r:\d\d4'
```

Verified on a live macOS 26 endpoint: the literal-glob form always returns *"No such file or directory"* and fails; the `sh -c` form returns `644` and passes. Fix applied in #36757 to the macOS 13.x / 14.x / 15.x / 26.x policies.

## What needs to be investigated

This is unlikely to be limited to the PolicyBanner check or to macOS — the same class of bug can exist in any SCA policy across any OS family. We are **not** providing an exhaustive list on purpose; please audit the rulesets and determine the real scope.

Suggested scope:

- Sweep all `c:` (and `not c:`) command rules in `ruleset/sca/` across every OS family (darwin, debian, ubuntu, rhel/centos/rocky/alma/amazon/oracle/sles, windows, applications, etc.).
- For each rule, look at the **command portion only** (everything before `->`; the part after `->` is a regex and a `*` there is expected and harmless).
- Flag any command that uses a shell-only feature — glob (`*`, `?`, `[...]`), pipe (`|`), redirection (`<`, `>`), command substitution (`$(...)` / backticks), `~` expansion, or `;` / `&&` sequencing — **without** being wrapped in `sh -c` / `bash -c`.
- For each candidate, confirm whether it is actually broken (it depends on the unsupported feature) vs. a false alarm (e.g. the metacharacter is literal data the program handles itself, or the binary does its own globbing). Reproduce on a real endpoint where possible.
- Apply the `sh -c` wrapper (or otherwise rewrite the rule to avoid the shell dependency) for each confirmed case, and verify the corrected result on a live endpoint.

## Acceptance criteria

- [x] Full audit of `c:` command rules across all SCA OS families completed.
- [x] Confirmed-broken rules identified and listed.
- [x] Fixes applied and validated on real endpoints.
- [x] Confirmation of whether this is widespread or isolated to the already-fixed macOS checks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,12 @@ All notable changes to this project will be documented in this file.
 - Fixed WPK upgrade failing on agents without the `find` binary. ([#38431](https://github.com/wazuh/wazuh/pull/38431))
 - Bounded the `snort-full` log record appends to the available buffer space and sized the queued preprocessor message from its own line in `wazuh-logcollector`. ([#38472](https://github.com/wazuh/wazuh/pull/38472))
 
+### Ruleset
+
+#### Fixed
+
+- Fixed SCA checks silently failing across macOS, RHEL/Debian, SLES, Ubuntu, Solaris, MongoDB policies by and adding missing shell wrappers and fixing broken rule syntax. ([#38679](https://github.com/wazuh/wazuh/pull/38679))
+
 ## [v4.14.8]
 
 ### Manager
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ All notable changes to this project will be documented in this file.
 
 #### Fixed
 
-- Fixed SCA checks silently failing across macOS, RHEL/Debian, SLES, Ubuntu, Solaris, MongoDB policies by and adding missing shell wrappers and fixing broken rule syntax. ([#38679](https://github.com/wazuh/wazuh/pull/38679))
+- Fixed SCA checks silently failing across macOS, RHEL/Debian, AlmaLinux, Amazon Linux, CentOS, Oracle Linux, and Rocky Linux, Ubuntu, Solaris, MongoDB policies by adding missing shell wrappers and fixing broken rule syntax. ([#38679](https://github.com/wazuh/wazuh/pull/38679))
 
 ## [v4.14.8]
 
--- a/ruleset/sca/almalinux/cis_alma_linux_10.yml
+++ b/ruleset/sca/almalinux/cis_alma_linux_10.yml
@@ -2743,7 +2743,7 @@ checks:
       - 'd:/etc/audit/rules.d -> r:\.+.rules$ -> r:^-a && r:always,exit|exit,always && r:-F arch=b64|-F arch=b32 && r:-S && r:init_module && r:finit_module && r:delete_module && r:create_module && r:query_module && r:-F auid>=\d+ && r:-F auid!=-1|-F auid!=4294967295|-F auid!=unset && r:-k kernel_modules|-F key=kernel_modules'
       - 'd:/etc/audit/rules.d -> r:\.+.rules$ -> r:^-a && r:always,exit|exit,always && r:-F path=/usr/bin/kmod && r:-F perm=x && r:-F auid>=\d+ && r:-F auid!=-1|-F auid!=4294967295|-F auid!=unset && r:-k kernel_modules|-F key=kernel_modules'
       - 'c:auditctl -l -> r:^-a && r:always,exit|exit,always && r:-F arch=b64|-F arch=b32 && r:-S && r:init_module && r:finit_module && r:delete_module && r:create_module && r:query_module && r:-F auid>=\d+ && r:-F auid!=-1|-F auid!=4294967295|-F auid!=unset && r:-k kernel_modules|-F key=kernel_modules'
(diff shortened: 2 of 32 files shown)


AwwalQuan (changes requested):
The `-name '*.conf'` → `-name "*.conf"` swaps, the `sh -c` wrappers, and the arrow-spacing fixes are all legitimate fixes for real tokenizer bugs. Those changes are correct as written.
The main objective of this PR has been addressed.

## New problems worth fixing in this PR

The fixes missed several rules broken in the same way, in files the PR is already touching for other checks:

1. **`ruleset/sca/ubuntu/cis_ubuntu22-04.yml:1696`**: the chrony "is-active" rule wraps the whole command in one quoted string right after `c:`, so the tokenizer treats it as a single argv token; `execvp` can never find a binary with that literal name. Check always fails.
2. **`ruleset/sca/darwin/20/cis_apple_macOS_11.1.yml:482`**: an ad-tracking check has a literal, unsubstituted `<username>` placeholder in a file path, plus a pattern missing its `r:`/`n:` type prefix. Always fails.
3. **`CHANGELOG.md`**: lists "SLES" (no SLES files are touched anywhere in the diff) while omitting AlmaLinux, Amazon Linux, CentOS, Oracle Linux, and Rocky Linux, which *are* modified. Also a grammar artifact ("by and adding").

AwwalQuan (approved):
LGTM

PR #38680, opened 2026-08-27
Stabilize flaky Windows FIM integration tests with teardown retries

### LINKED ISSUE #38675: Flaky Windows FIM integration tests (IT Windows - fim)
## Description

`IT Windows - fim (tier-1)` gave three different outcomes across three consecutive CI runs of the exact same commit while validating PR #38537 — a different test failed each time it failed, and the third rerun passed clean with no code changes in between. That non-reproducibility across runs is itself the bug being reported: the suite isn't reliably signaling real regressions right now.

## Evidence

Same head commit (`d65fc7d5`, later `13e95187`) on PR #38537:

- Attempt 1 ([job 98590513553](https://github.com/wazuh/wazuh/actions/runs/33091323865/job/98590513553?pr=38537)): 62 passed, 1 skipped, 1 error — failure was at teardown of `test_max_eps[50_eps_creating_multiple_files_realtime]`, not the test body itself:
  ```
  PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\wazuh\tests\integration\testdir1\file150_to_max_eps_50_realtime_mode1787848135.1271296_.txt'
  ```
  `shutil.rmtree()` hit the file still locked by something (most likely `wazuh-syscheckd`'s own realtime watcher not having released it yet) at cleanup time.
- Attempt 2 ([job 98612234863](https://github.com/wazuh/wazuh/actions/runs/33091323865/job/98612234863?pr=38537)): a different test failed this time, partway through `test_files/test_report_changes/test_large_changes.py` (an `F` in the live pytest dot-output around 68% progress). The job was cancelled by a manual rerun before it finished, so the exact assertion/error text was never captured — flagging this explicitly since it's a gap in this report, not a claim of a specific error.
- Attempt 3 ([job 98621020887](https://github.com/wazuh/wazuh/actions/runs/33091323865/job/98621020887?pr=38537)): passed clean, 0 failures.

## Ruled out as the cause

Checked whether PR #38537's own change could explain the file lock: it adds per-realtime-event work in `fim_file()`'s non-transactional branch (`src/syscheckd/src/file/file.c`) — a mutex lock, two `OSList` allocations, and one extra SQLite write via `process_pending_sync_updates()`. None of that touches the scanned file itself; `fim_get_data()` already opened, read, and closed it earlier in the same function, before any of the new code runs. There's no code path connecting the fix to a lingering file handle, so this reads as pre-existing Windows CI flakiness under load rather than a regression introduced by that PR.

Not the same bug as #38622 (malformed monitored-directory path breaking `detect_end_scan` via `os.path.join('/', 'test_dir', ...)` producing a mixed-separator path on Windows) — different symptom (a real `WinError 267` on directory creation vs. a `WinError 32` file lock at teardown, and a real assertion failure with no captured detail for the second run), different tests affected. Related in that both live in the same suite and both currently make Windows FIM CI unreliable as a merge gate.

## To investigate

- Whether `wazuh-syscheckd`'s realtime watcher (or antivirus/indexing on the runner) is the one holding the lock in the `test_max_eps` teardown case, and whether the test needs to wait for watch release before `shutil.rmtree`.
- Get a clean capture of the `test_large_changes.py` failure (rerun in isolation, let it run to completion) to know if it's the same class of teardown/timing issue or a real assertion failure.
- Whether these two, #38622, and the historical #29922 (closed without a confirmed fix, same recurring `detect_end_scan`-adjacent symptom) point to a shared root cause in how Windows FIM ITs handle cleanup/timing under realtime load, rather than three unrelated bugs.

Related: #38622, #29922

MAnDumu:
## Description

`IT Windows - fim (tier-1)` gave three different outcomes across three consecutive CI runs of the exact same commit while validating PR #38537 — a different test failed each time, and a rerun with no code changes passed clean. One case has a confirmed root cause with full traceback: the teardown of `test_max_eps[50_eps_creating_multiple_files_realtime]` fails with `PermissionError: [WinError 32]` because `shutil.rmtree()` runs before `wazuh-syscheckd`'s realtime watcher has released its handle on a file inside the monitored directory.

Closes #38675

## Proposed Changes

- `tests/integration/test_fim/conftest.py`: `folder_to_monitor`'s teardown now retries `file.delete_path_recursively()` on Windows instead of failing outright. This mirrors the retry pattern the sibling fixture `file_to_monitor` (same file) has had since commit `44047cdd99`, which never got applied to `folder_to_monitor`. `test_files/test_report_changes/test_large_changes.py` also depends on this fixture (see code comment), so this retry covers its previously-uncaptured flaky teardown failure too, even though we never got a traceback from that specific run to confirm it's the identical root cause.
- `tests/integration/test_fim/test_files/test_basic_usage/conftest.py`: same retry added to `path_to_edit`. Correction from an earlier version of this PR: this fixture is **not** used by `test_delete_directory.py` (that was wrong) — its only two consumers, `test_rename.py` and `test_move.py`, relocate the path into `folder_to_monitor` before teardown runs, so this retry is a defensive no-op for them today. Kept for any future consumer that doesn't relocate first (see code comment).
- `tests/integration/test_fim/test_files/test_basic_usage/test_delete_directory.py`: same retry added around `file.delete_path_recursively(folder_to_delete)` in the test body itself (not a fixture teardown, it's what triggers the delete event the test asserts on). Note: this whole module is `skipif`'d on Windows already (`pytestmark`, unrelated to this PR), for the same class of instability this PR is fixing elsewhere — so this retry is dormant today, kept for when that skip gets lifted rather than protecting anything that currently runs.
- `tests/integration/test_fim/conftest.py`, `fill_folder_to_monitor`: found while re-checking this same file for siblings of the pattern above (not flagged by the human review, caught in a follow-up pass) — this fixture had the same "retry deletion in teardown" idea but strictly worse: no `sys.platform == WINDOWS` guard (ran unconditionally on every OS), caught bare `Exception`, a 3 x 1s budget, still used `print()`, and on exhausting retries it `break`ed silently instead of raising — masking real cleanup failures rather than surfacing them. Its consumers (`test_full_capacity_create_delete.py`, `test_fill_capacity.py`) run on Windows with no module-level skip, so this was live exposure, not dormant. Rewritten to match the other three sites: Windows-only retry, 15 x 1s, `except OSError`, `logger.debug()`, and a real `raise` when retries are exhausted; non-Windows now does a single unguarded attempt like its sibling fixtures, instead of silently swallowing failures.
- Retry budget: 15 attempts x 1s delay everywhere above, including the pre-existing `file_to_monitor` fixture (was 10 x 0.5s) and `fill_folder_to_monitor` (was 3 x 1s and non-Windows-unguarded) — five sites now share one budget instead of three different ones. The wider margin accounts for `test_max_eps` with `max_eps=50` actually creating 200 files (`max_eps * 4`), not 50.
- Broadened the caught exception from `PermissionError` (or bare `Exception`, for `fill_folder_to_monitor`) to `OSError` at every site: `on_write_error` (the `shutil.rmtree` error handler in `wazuh_testing.utils.file`) re-raises whatever the OS actually reported, and if a handle was opened with `FILE_SHARE_DELETE` (e.g. an AV/indexer), Windows leaves the entry "delete-pending" and the failure surfaces as `WinError 145` (`ENOTEMPTY`) on the parent `rmdir` — a plain `OSError`, not a `PermissionError` — which would have skipped the retry entirely.
- Switched every retry's diagnostic `print()` to `logger.debug()` (the convention used elsewhere in `tests/integration/conftest.py`): pytest discards captured stdout on the path where the fixture eventually succeeds, so `print()` was invisible on the common case and only showed up when a traceback already carried the same info.
- Non-Windows behavior is unchanged, except for `fill_folder_to_monitor` where it now raises on failure instead of always silently continuing (see above — this was a real bug, not a platform-specific concern).

### Results and Evidence

- `python3 -m py_compile` on all changed files: passes clean.
- Live CI verification (of an earlier version of this diff, using `PermissionError` instead of `OSError` and `print()` instead of `logger.debug()`, before this round of review feedback): temporarily touched an unrelated `src/syscheckd/` comment to trigger the full Windows build+package+IT pipeline. `IT Windows - fim (tier-0)` and `IT Windows - fim (tier-1)` both passed, including `test_max_eps.py` and `test_large_changes.py` specifically (confirmed from the job logs) — the latter completing cleanly at the same ~68% suite-progress point where the original issue reported a failure. That trigger commit was reverted afterward; it's not part of this PR's diff. The `except OSError` broadening, unified retry budget, and `logger` switch landed after that run and have not been re-verified live yet — they're behavior-preserving on the paths already covered (Python-syntax-checked, `OSError` is a superset of `PermissionError`) but worth another live run if there's any doubt.

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

- Engine _(Wazuh v5.x and above)_: N/A
- Wazuh server API/Framework

### Artifacts Affected

None — this only changes Python integration-test code under `tests/integration/test_fim/`. No production code, executables, or default configuration files are touched.

### Configuration Changes

N/A

### Tests Introduced

No new tests. This changes teardown/cleanup behavior in existing pytest fixtures and one test body to make Windows CI runs more reliable rather than adding coverage.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

Out of scope for this PR (flagged for follow-up, not fixed here for lack of confirmed evidence of the same root cause):
- Other `delete_path_recursively()` call sites in `test_fim/` (`create_paths_files`, `test_create_after_delete_dir.py`, `test_orphan_promote_after_config_removal.py`, `test_num_watches.py`) were not touched — no evidence yet that they hit the same Windows realtime-watcher lock. (`test_num_watches.py` is under `test_inotify/`, which is Linux-only, so it doesn't apply here anyway.)

### DIFF
--- a/tests/integration/test_fim/test_files/test_basic_usage/conftest.py
+++ b/tests/integration/test_fim/test_files/test_basic_usage/conftest.py
@@ -9,6 +9,7 @@ from pathlib import Path
 
 from wazuh_testing.constants.paths.logs import WAZUH_LOG_PATH
 from wazuh_testing.constants.platforms import WINDOWS
+from wazuh_testing.logger import logger
 from wazuh_testing.modules.fim.patterns import EVENT_TYPE_ADDED, PATH_MONITORED_REALTIME
 from wazuh_testing.tools.monitors.file_monitor import FileMonitor
 from wazuh_testing.utils import file
@@ -39,4 +40,21 @@ def path_to_edit(test_metadata: dict) -> str:
 
     yield to_edit
 
-    file.delete_path_recursively(to_edit)
+    # test_rename.py and test_move.py -- the only two consumers of this fixture -- relocate
+    # to_edit into folder_to_monitor before teardown runs, so this retry is a defensive no-op
+    # for them today; folder_to_monitor's own teardown is what actually covers the relocated
+    # content. Kept here in case a future test uses this fixture without relocating first.
+    if sys.platform == WINDOWS:
+        max_retries = 15
+        retry_delay = 1
+        for attempt in range(max_retries):
+            try:
+                file.delete_path_recursively(to_edit)
+                break
+            except OSError as exception:
+                if attempt == max_retries - 1:
+                    raise
+                logger.debug(f"Retrying deletion of {to_edit}: {exception}")
+                time.sleep(retry_delay)
+    else:
+        file.delete_path_recursively(to_edit)
--- a/tests/integration/test_fim/conftest.py
+++ b/tests/integration/test_fim/conftest.py
@@ -20,6 +20,7 @@ from pathlib import Path
 from wazuh_testing.constants.paths.databases import FIM_DB_PATH, FIM_SYNC_DB_DIR, FIM_SYNC_DB_FILES
 from wazuh_testing.constants.paths.logs import WAZUH_LOG_PATH
 from wazuh_testing.constants.platforms import WINDOWS, MACOS, CENTOS, UBUNTU, DEBIAN
+from wazuh_testing.logger import logger
 from wazuh_testing.modules.fim.patterns import MONITORING_PATH, FIM_SCAN_END
 from wazuh_testing.modules.fim.utils import create_registry, delete_registry
 from wazuh_testing.tools.monitors.file_monitor import FileMonitor
@@ -41,15 +42,16 @@ def file_to_monitor(test_metadata: dict) -> Any:
     yield path
 
     if sys.platform == WINDOWS:
-        max_retries = 10
-        retry_delay = 0.5
+        max_retries = 15
+        retry_delay = 1
         for attempt in range(max_retries):
             try:
                 file.remove_file(path)
                 break
-            except PermissionError:
+            except OSError as exception:
                 if attempt == max_retries - 1:
                     raise
+                logger.debug(f"Retrying deletion of {path}: {exception}")
                 sleep(retry_delay)
     else:
         file.remove_file(path)
@@ -57,6 +59,9 @@ def file_to_monitor(test_metadata: dict) -> Any:
 
 @pytest.fixture()
 def folder_to_monitor(test_metadata: dict) -> None:
+    # test_files/test_report_changes/test_large_changes.py depends on this fixture too, so the
+    # retry below also covers its intermittent Windows teardown failure reported in #38675 --
+    # that failure's traceback was never captured, but it shares this same teardown call.
     path = test_metadata.get("folder_to_monitor")
     path = os.path.abspath(path)
 
@@ -64,7 +69,20 @@ def folder_to_monitor(test_metadata: dict) -> None:
 
     yield path
 
-    file.delete_path_recursively(path)
+    if sys.platform == WINDOWS:
+        max_retries = 15
+        retry_delay = 1
+        for attempt in range(max_retries):
+            try:
+                file.delete_path_recursively(path)
+                break
+            except OSError as exception:
+                if attempt == max_retries - 1:
+                    raise
+                logger.debug(f"Retrying deletion of {path}: {exception}")
+                sleep(retry_delay)
+    else:
+        file.delete_path_recursively(path)
 
 
 @pytest.fixture()
@@ -72,8 +90,6 @@ def fill_folder_to_monitor(test_metadata: dict) -> None:
     path = test_metadata.get("folder_to_monitor")
     amount = test_metadata.get("files_amount")
     amount = 2 if not amount else amount
-    max_retries = 3
-    retry_delay = 1
 
     if not file.exists(path):
         file.recursive_directory_creation(path)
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -142,3 +142,4 @@
 | [#38065](https://github.com/wazuh/wazuh/issues/38065) | Fixed SCA and Syscollector sync threads not blocking `SIGTERM`, which could cause the shutdown handler to run on a module thread instead of the main thread and time out joining it. |
 | [#38212](https://github.com/wazuh/wazuh/issues/38212) | Fixed the Windows agent leaving the FIM synchronization database open when the service stopped, which left the `queue\` directory behind after an uninstall without purge. |
 | [#38646](https://github.com/wazuh/wazuh/pull/38646) | Fixed SCA HIPAA compliance mappings across policy checks. |
```
jpcerrone, line 147: Remove this. Changelog shouldn´t mention the test infrastructure. Also when we do 5.0.0 changes,  there's usually no need to add a changelog entry unless we're adding a new feature that wasn't available in 4x,

tests/integration/test_fim/test_files/test_basic_usage/conftest.py
```diff
@@ -39,4 +39,17 @@ def path_to_edit(test_metadata: dict) -> str:
```
jpcerrone, line 40: Even though we have no run of `test_large_changes.py` to see whether it failed because of the same root cause we're fixing here, it may be worth it to analyze its code to see if the changes we've made would apply there. Add a simple comment with the result of this analysis.

jpcerrone (changes requested):

CHANGELOG.md
MAnDumu: Removed.

tests/integration/test_fim/test_files/test_basic_usage/conftest.py
MAnDumu: Confirmed: test_large_changes.py uses folder_to_monitor (not path_to_edit), so this retry already covers it. Added a comment next to folder_to_monitor's teardown in test_fim/conftest.py noting that, and the caveat that the original failure's traceback was never captured to prove it's the same root cause.

jpcerrone (approved):

jpcerrone:
### Modified tests not showing up in the CI runs
The modified tests haven't been run automatically on this PR due to their files being missing in the globs inside the scripts that check whether the workflows should run or not. We've opened up an issue to fix this and it's being tracked  here: https://github.com/wazuh/wazuh/issues/38740

We've ran them manually to validate the changes introduced instead: https://github.com/wazuh/wazuh/actions/runs/33184080100/job/98895113108?pr=38680 :green_circle:

cborla (approved):
LGTM
- [x] Code changes reviewed

PR #38681, opened 2026-08-27
ci: trigger Windows agent build on integration-test changes for its own modules

MAnDumu:
## Description

While working on #38675 (a flaky Windows FIM integration-test teardown fix, PR #38680), the "IT Windows - fim (tier-N)" checks never appeared on that PR at all. Root cause: `5_builderpackage_agent-windows.yml`'s `pull_request.paths` filter — which gates the whole build+package+smoke-test+IT pipeline — never included `tests/integration/**`, only `src/**`, `packages/**`, etc. So a PR that only changes test fixtures under `tests/integration/test_fim/` never triggers the workflow at all, even though `.github/test_modules_windows.json` already maps that exact path to the "fim" module for the internal `detect-changes` job. On top of that, the actual `run-integration-tests` job is gated to `if: github.event_name == 'pull_request'`, so `workflow_dispatch` can't be used as a workaround either — there's no way to get Windows FIM CI coverage on a test-only PR today.

Refs #38675 (not a `Closes`, since this doesn't fix the flaky test itself — it's what's needed to actually get CI coverage on that fix).

## Proposed Changes

- Added the paths from each Windows-covered module's `paths` list in `.github/test_modules_windows.json` (`test_agentd`, `test_enrollment`, `test_execd`, `test_fim`, `test_github`, `test_logcollector`, `test_office365`, `test_sca`, `test_syscollector`, plus the shared `tests/integration/conftest.py`) to the `pull_request.paths` filter of `5_builderpackage_agent-windows.yml`.
- Deliberately did **not** use a bare `tests/integration/**` wildcard: that folder also holds several suites with no Windows tier at all (`test_authd`, `test_aws`, `test_msgraph`, `test_remoted`, `test_wazuh_db`) — over a recent 90-day sample, those accounted for more commit activity than the Windows-relevant folders combined, which would have made this fix trigger the ~90min Windows build+package+smoke-test pipeline more often for irrelevant changes than for the ones it's meant to catch.
- Added a comment tying the list back to `test_modules_windows.json` as the source of truth, since the two need to stay in sync manually.

### Results and Evidence

- Validated with `yaml.safe_load` that the workflow file still parses correctly.
- This PR's own diff touches `.github/workflows/5_builderpackage_agent-windows.yml`, which is already (and remains) in its own trigger paths — so this PR should self-trigger the workflow and can be used as live evidence that the change works, once CI runs.

- Compilation without warnings on every supported platform: N/A (no compiled code changed)
- Memory tests: N/A
- Decoder/Rule tests: N/A
- Engine: N/A
- Wazuh server API/Framework: N/A

### Artifacts Affected

None — only a GitHub Actions workflow trigger configuration. No production code, executables, or default configuration files.

### Configuration Changes

CI-only: widens which file changes cause `5_builderpackage_agent-windows.yml` (Windows agent build, MSI packaging, smoke test, and per-module integration tests) to run automatically on a PR. No behavior change for the manager, agent, or any shipped configuration.

### Tests Introduced

None — this is what makes existing Windows integration tests actually run for test-only PRs; it doesn't add new test cases.

## Review Checklist

- [ ] Code changes reviewed
- [x] Tests cover the new functionality (N/A, CI trigger config)
- [ ] ...

Known related gap, intentionally out of scope here: `5_builderpackage_agent-linux.yml` and `5_builderpackage_agent-macos.yml` have the same missing `tests/integration/**` coverage in their own `pull_request.paths` filters (verified while investigating this). Flagging for a follow-up rather than bundling into this PR, since the issue that surfaced this (#38675) is Windows-specific.


MAnDumu:
Cerrando: el ajuste correcto para verificar #38675 es puntual dentro de esa PR (#38680), no un cambio de trigger a nivel proyecto.

PR #38682, opened 2026-08-27
fix(remoted): log authd.pass wait as DEBUG while a worker is joining

### LINKED ISSUE #38605: authd.pass enrollment password not readable at manager startup
|---|---|---|---|---|
| 5.0.0-latest-beta5 | Wazuh manager (remoted) | Central components (clustered AIO) | Packages | Linux (RedHat 9, amd64) |

**Description**

During the `5.0.0` Central Components DTT run (clustered Manager + Indexer + Dashboard, 10 all-in-one nodes), `wazuh-manager-remoted` reports it cannot read the enrollment password file `etc/authd.pass` right after install, on one of the 10 platforms (`redhat-9-amd64`). This causes the CC log-check test `test_00_setup_install::test_check_logs[Manager]` to fail.

**Evidence**

```
2026/08/26 13:42:09 wazuh-manager-remoted:passwordKeySource: WARNING: Could not load a usable enrollment password from 'etc/authd.pass' at startup; Password-mode enrollment requests will be rejected until it is available.
2026/08/26 13:42:09 wazuh-manager-remoted:passwordKeySource: WARNING: Could not watch 'etc/authd.pass' for changes (errno=2); hot-reload falls back to the 10 s poll only.
2026/08/26 13:42:19 wazuh-manager-remoted:passwordKeySource: WARNING: 'etc/authd.pass' is not readable or its content is invalid (errno=2); Password-mode enrollment requests will be rejected. 1 failed attempt(s) in the last 90 s.
```

The sequence repeats a few more times over roughly 20 seconds (same 3 message types), then stops appearing in the rest of the run. Only `redhat-9-amd64` reported it; the other 9 platforms did not.

**Analysis**

`errno=2` is `ENOENT` (file not found), which suggests `authd.pass` had not been created yet by the time `remoted` first tried to read it right after install, and the hot-reload/poll mechanism recovered once the file became available. This looks like a startup ordering/race condition rather than a persistent failure, but we'd like confirmation on whether this is expected during a fresh install, and if so, whether it should be added to the DTT log ignore list.

Execution: https://jenkins-staging.qa.wazuh.info/job/5.0.0/job/deployability_testing/1557/

MiguelazoDS:
## Description
Closes #38605

`wazuh-manager-remoted` logged WARNINGs for a missing/unreadable `etc/authd.pass` at startup and on later polls, even though this is expected and self-resolving on a freshly-joined cluster worker while it waits for the master's cluster sync to deliver the file. Live reproduction on a 2-node cluster confirmed this timing (~10-20s) and that a master never legitimately hits this path.

This change suppresses the alarming WARNINGs during a bounded grace window on a worker only, replacing them with informational messages; a worker still stuck past the window, or a master hitting this at all, still warns exactly as before.

## Proposed Changes
- Added a pure, unit-testable WARN/INFO decision function (`PasswordKeySource::shouldWarnAboutMissingPassword`) gated on node role and elapsed time since construction.
- Threaded `isWorkerNode` into `PasswordKeySource`, defaulted so every existing call site (production and test) keeps compiling unchanged.
- Wired the already-computed `enrollConfig.isWorkerNode` through at the one production construction site in `remotedModuleFacade.hpp`.
- Gated all three of the original log lines (missing-password, can't-watch-for-changes, not-readable-on-poll) on the same decision, so a normal worker join produces zero WARNINGs, not just fewer.
- Added 4 unit tests covering the master-always-warns, worker-within-grace-window, and worker-past-grace-window (the required stuck-worker safety property) cases, plus a constructor-level regression check.
- Updated the module's Observability doc to describe the now-conditional WARN/INFO behavior for this log line.

## Results and Evidence
- `./remoted_module_utest --gtest_filter='PasswordKeySourceTest.*'` → 22/22 passed (18 pre-existing unaffected + 4 new, including the required stuck-worker safety test).
- Manually verified on a live 2-node cluster (see issue comments): a normal worker restart now shows zero WARNING lines; a worker with cluster sync deliberately prevented still escalates to WARNING once past the grace window.

## Artifacts Affected
- `src/remoted/remoted_module/src/auth/passwordKeySource.hpp`
- `src/remoted/remoted_module/src/auth/passwordKeySource.cpp`
- `src/remoted/remoted_module/src/remotedModuleFacade.hpp`
- `src/remoted/remoted_module/test/unit/passwordKeySource_test.cpp`
- `docs/ref/modules/remoted/https-events-api.md`
- `CHANGELOG.md`

## Configuration Changes
N/A — no new parameters; `kWorkerJoinGraceSeconds` is a fixed internal constant, not user-configurable. No backward-compatibility impact: default `isWorkerNode=false` preserves today's behavior for every existing caller.

## Tests Introduced
**New:**
- `PasswordKeySourceTest.MasterAlwaysWarnsEvenImmediately`
- `PasswordKeySourceTest.WorkerDoesNotWarnWithinGraceWindow`
- `PasswordKeySourceTest.WorkerStillWarnsOnceGraceWindowIsExceeded`
- `PasswordKeySourceTest.MissingFileHasNoKeyRegardlessOfWorkerFlag`

**Modified:** None (all 18 pre-existing `PasswordKeySourceTest` cases pass unchanged).

## Review Checklist
- [x] Code follows the project's style guide (Allman braces, 4-space indent, clang-format applied)
- [x] Tests added/updated and passing
- [x] CHANGELOG.md updated
- [x] Documentation updated
- [x] No unrelated/drive-by changes

### DIFF
--- a/docs/ref/modules/remoted/https-events-api.md
+++ b/docs/ref/modules/remoted/https-events-api.md
@@ -579,9 +579,14 @@ Three more that are not about tuning:
   that can actually authenticate — a key that fails to decode is reported separately and is **not**
   counted, so this number can be trusted. If `client.keys` is unreadable, that is logged explicitly:
   otherwise it presents only as every agent being rejected as unknown, with nothing explaining why.
+- A missing or unreadable `etc/authd.pass`, at `remoted` startup or on a later poll, is normally a
+  **WARNING** naming the file. The exception is a cluster worker within about a minute of starting:
+  there, the file not having synced down from the master yet is expected, so it logs at **DEBUG1**
+  instead ("waiting for the enrollment password to be synchronized from the master node"). A worker
+  still missing the file past that window, or a master hitting this path at all, still warns.
 - **`Could not derive the enrollment key from 'etc/authd.pass' (HKDF unavailable)`** (ERROR) means the
   OpenSSL KDF provider is broken: every Password-mode enrollment fails closed until it is fixed,
-  distinct from an unreadable or invalid password file (a WARNING naming the file).
+  distinct from an unreadable or invalid password file (see above).
 - **The HTTPS server failing to start** is an ERROR naming which of the two is the problem (the
   certificate or the private key). There is no retry: remoted must not start without the HTTPS
   transport up, so a missing or unreadable certificate/key is fatal to the whole daemon, not just
--- a/src/remoted/remoted_module/src/auth/passwordKeySource.cpp
+++ b/src/remoted/remoted_module/src/auth/passwordKeySource.cpp
@@ -126,9 +126,21 @@ namespace remoted::auth
 
     } // namespace
 
-    PasswordKeySource::PasswordKeySource(std::string path, int refreshIntervalSeconds)
+    bool PasswordKeySource::shouldWarnAboutMissingPassword(bool isWorkerNode,
+                                                           std::chrono::steady_clock::duration elapsedSinceConstruction)
+    {
+        if (!isWorkerNode)
+        {
+            return true;
+        }
+        return elapsedSinceConstruction >= std::chrono::seconds(kWorkerJoinGraceSeconds);
+    }
+
+    PasswordKeySource::PasswordKeySource(std::string path, int refreshIntervalSeconds, bool isWorkerNode)
         : m_path(std::move(path))
         , m_refreshIntervalSeconds(refreshIntervalSeconds > 0 ? refreshIntervalSeconds : kDefaultRefreshIntervalSeconds)
+        , m_isWorkerNode(isWorkerNode)
+        , m_constructedAt(std::chrono::steady_clock::now())
     {
         // Report the initial state unconditionally: if Password mode is selected but this file
         // isn't readable yet (e.g. not synced from the master to a worker), every enrollment
@@ -137,12 +149,19 @@ namespace remoted::auth
         {
             LOGFN_INFO(logFn(), "Enrollment password loaded from '%s'.", m_path.c_str());
         }
-        else
+        else if (shouldWarnAboutMissingPassword(m_isWorkerNode, std::chrono::steady_clock::duration::zero()))
         {
             LOGFN_WARN(logFn(),
                        "Could not load a usable enrollment password from '%s' at startup; "
                        "Password-mode enrollment requests will be rejected until it is available.",
                        m_path.c_str());
+        }
+        else
+        {
+            LOGFN_DEBUG1(logFn(),
+                         "Waiting for the enrollment password to be synchronized from the master node to '%s'; "
+                         "Password-mode enrollment requests will be rejected until it is available.",
+                         m_path.c_str());
         }
 
         m_inotifyFd = inotify_init1(IN_NONBLOCK | IN_CLOEXEC);
@@ -158,11 +177,29 @@ namespace remoted::auth
             m_watchDescriptor = inotify_add_watch(m_inotifyFd, m_path.c_str(), kWatchMask);
             if (m_watchDescriptor < 0)
             {
-                LOGFN_WARN(logFn(),
-                           "Could not watch '%s' for changes (errno=%d); hot-reload falls back to the %d s poll only.",
-                           m_path.c_str(),
-                           errno,
-                           m_refreshIntervalSeconds);
+                const int watchErrno = errno;
+                // ENOENT here is the same "file not synced yet" condition as the reload() branch
+                // above -- gate it the same way. Any OTHER errno (EACCES, hitting
+                // fs.inotify.max_user_watches, ...) is a real, unrelated problem and must keep
+                // warning regardless of node role or timing.
+                if (watchErrno == ENOENT &&
+                    !shouldWarnAboutMissingPassword(m_isWorkerNode, std::chrono::steady_clock::duration::zero()))
+                {
+                    LOGFN_DEBUG1(logFn(),
+                                 "Cannot watch '%s' for changes yet (not synchronized from the master node); "
+                                 "hot-reload falls back to the %d s poll only until it is available.",
+                                 m_path.c_str(),
+                                 m_refreshIntervalSeconds);
+                }
+                else
+                {
+                    LOGFN_WARN(
+                        logFn(),
+                        "Could not watch '%s' for changes (errno=%d); hot-reload falls back to the %d s poll only.",
+                        m_path.c_str(),
+                        watchErrno,
+                        m_refreshIntervalSeconds);
+                }
             }
         }
 
@@ -338,13 +375,26 @@ namespace remoted::auth
                 }
                 else if (const auto d = m_unreadableThrottle.record())
                 {
-                    LOGFN_WARN(logFn(),
-                               "'%s' is not readable or its content is invalid (errno=%d); Password-mode "
-                               "enrollment requests will be rejected. %llu failed attempt(s) in the last %d s.",
-                               m_path.c_str(),
-                               errno,
-                               static_cast<unsigned long long>(d.total),
-                               remoted::common::LogThrottle::kDefaultWindowSeconds);
+                    const auto elapsed = std::chrono::steady_clock::now() - m_constructedAt;
+                    if (shouldWarnAboutMissingPassword(m_isWorkerNode, elapsed))
+                    {
+                        LOGFN_WARN(logFn(),
+                                   "'%s' is not readable or its content is invalid (errno=%d); Password-mode "
+                                   "enrollment requests will be rejected. %llu failed attempt(s) in the last %d s.",
+                                   m_path.c_str(),
+                                   errno,
+                                   static_cast<unsigned long long>(d.total),
+                                   remoted::common::LogThrottle::kDefaultWindowSeconds);
+                    }
+                    else
+                    {
+                        LOGFN_DEBUG1(logFn(),
+                                     "Still waiting for '%s' to be synchronized from the master node (%llu check(s) "
+                                     "in the last %d s).",
+                                     m_path.c_str(),
+                                     static_cast<unsigned long long>(d.total),
+                                     remoted::common::LogThrottle::kDefaultWindowSeconds);
+                    }
                 }
             }
         }
--- a/src/remoted/remoted_module/src/auth/passwordKeySource.hpp
+++ b/src/remoted/remoted_module/src/auth/passwordKeySource.hpp
@@ -12,6 +12,7 @@
 #pragma once
 
 #include <atomic>
+#include <chrono>
 #include <cstdint>
 #include <mutex>
 #include <optional>
@@ -64,15 +65,41 @@ namespace remoted::auth
         /// Built-in refresh interval when the caller passes <=0 (matches Keystore's own default).
         static constexpr int kDefaultRefreshIntervalSeconds = 10;
 
+        /// How long after construction a worker's missing/unreadable authd.pass is treated as an
+        /// expected wait for the master's cluster sync (DEBUG1) rather than a fault (WARN). Chosen
+        /// as generous relative to the ~20s recovery observed reproducing a worker cluster-sync
+        /// wait on a live cluster.
+        static constexpr int kWorkerJoinGraceSeconds = 60;
+
+        /**
+         * @brief Whether a missing/unreadable authd.pass should be logged as a WARNING right now.
+         *
+         * True unconditionally when @p isWorkerNode is false (master/standalone) -- a master
+         * should never legitimately hit this path, so if it does, it's worth flagging
+         * immediately, same as today. For a worker, true only once @p elapsedSinceConstruction has
+         * passed kWorkerJoinGraceSeconds -- within that window the wait is expected (the file
+         * hasn't synced from the master yet) and should log at DEBUG1 instead; a worker still
+         * stuck past the window is a real problem and must still warn.
+         *
+         * Pure and static so it's directly unit-testable with fabricated durations, without
+         * needing to capture real log output across the DSO boundary or wait out a real grace
+         * window in a test.
+         */
+        static bool shouldWarnAboutMissingPassword(bool isWorkerNode,
+                                                   std::chrono::steady_clock::duration elapsedSinceConstruction);
+
         /**
          * @param path Path to authd's password file. An initial reload() runs synchronously in
          *             the constructor; a missing or invalid file is not an error -- currentKey()
          *             simply starts at nullopt (fail closed for Password mode).
          * @param refreshIntervalSeconds How often the background watcher re-checks the file as a
          *             fallback to the inotify subscription (seconds). <=0 -> kDefaultRefreshIntervalSeconds.
+         * @param isWorkerNode Whether this manager is a cluster worker -- gates the startup/poll
+         *             grace-window behavior of shouldWarnAboutMissingPassword().
          */
         explicit PasswordKeySource(std::string path = kDefaultPath,
-                                   int refreshIntervalSeconds = kDefaultRefreshIntervalSeconds);
+                                   int refreshIntervalSeconds = kDefaultRefreshIntervalSeconds,
+                                   bool isWorkerNode = false);
 
         ~PasswordKeySource();
 
@@ -128,6 +155,9 @@ namespace remoted::auth
         std::mutex m_reloadMutex;
         bool m_hasBaseline {false};
         std::vector<std::uint8_t> m_lastHash;
+
+        bool m_isWorkerNode;
+        std::chrono::steady_clock::time_point m_constructedAt;
 
         /// Throttles the "authd.pass is unreadable" warning: the watcher retries every
         /// m_refreshIntervalSeconds, so an unreadable file would otherwise flood wazuh-manager.log.
--- a/src/remoted/remoted_module/src/remotedModuleFacade.hpp
+++ b/src/remoted/remoted_module/src/remotedModuleFacade.hpp
@@ -546,8 +546,10 @@ private:
         std::shared_ptr<remoted::auth::PasswordKeySource> enrollPasswordKeySource;
         if (enrollConfig.usePassword)
         {
-            enrollPasswordKeySource = std::make_shared<remoted::auth::PasswordKeySource>(
-                remoted::auth::PasswordKeySource::kDefaultPath, enrollConfig.passwordRefreshIntervalSec);
+            enrollPasswordKeySource =
+                std::make_shared<remoted::auth::PasswordKeySource>(remoted::auth::PasswordKeySource::kDefaultPath,
+                                                                   enrollConfig.passwordRefreshIntervalSec,
+                                                                   enrollConfig.isWorkerNode);
         }
 
         m_enrollmentAuthenticator = std::make_unique<remoted::enrollment::EnrollmentAuthenticator>(
--- a/src/remoted/remoted_module/test/unit/passwordKeySource_test.cpp
+++ b/src/remoted/remoted_module/test/unit/passwordKeySource_test.cpp
@@ -86,6 +86,39 @@ namespace
         EXPECT_FALSE(source.currentKey().has_value());
     }
 
+    TEST_F(PasswordKeySourceTest, MissingFileHasNoKeyRegardlessOfWorkerFlag)
+    {
+        PasswordKeySource source(m_path, PasswordKeySource::kDefaultRefreshIntervalSeconds, /*isWorkerNode=*/true);
+        EXPECT_FALSE(source.currentKey().has_value());
+    }
+
+    // ---------------------------------------------------------------------------
+    // shouldWarnAboutMissingPassword: pure WARN-vs-DEBUG1 decision function
+    // ---------------------------------------------------------------------------
+
+    TEST_F(PasswordKeySourceTest, MasterAlwaysWarnsEvenImmediately)
+    {
+        EXPECT_TRUE(PasswordKeySource::shouldWarnAboutMissingPassword(/*isWorkerNode=*/false, std::chrono::seconds(0)));
+    }
+
+    TEST_F(PasswordKeySourceTest, WorkerDoesNotWarnWithinGraceWindow)
+    {
+        EXPECT_FALSE(PasswordKeySource::shouldWarnAboutMissingPassword(/*isWorkerNode=*/true, std::chrono::seconds(0)));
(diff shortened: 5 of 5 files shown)


MiguelazoDS:
# Motivation and design of this fix

## The problem this solves

A cluster **worker** doesn't generate `etc/authd.pass` locally — only the master does (`w_authd_load_password()`, `os_auth/src/main-server.c`). A worker's own `authd` only *reads* the file (`w_authd_read_password()`); it depends entirely on the Python cluster layer's periodic integrity sync (`cluster.json`'s `worker.sync_integrity`, ~9s) to deliver it from the master after a fresh join or restart.

Live reproduction on a real 2-node cluster (see issue #38605) confirmed this takes ~10-20s in the normal case, sometimes longer across repeated restarts. During that window, `remoted`'s `PasswordKeySource` — which reads the *same* file locally for its own HTTPS `/enroll` Password-mode auth — logged this as three separate WARNING lines worded like something was broken:

```
WARNING: Could not load a usable enrollment password from 'etc/authd.pass' at startup; ...
WARNING: Could not watch 'etc/authd.pass' for changes (errno=2); hot-reload falls back to the 10 s poll only.
WARNING: 'etc/authd.pass' is not readable or its content is invalid (errno=2); ... N failed attempt(s) ...
```

This is **not a fault** — it's the normal, expected shape of a worker joining a cluster. Reproduction also confirmed a *master* never legitimately hits this path (its own `authd` generates the file locally, fast enough that `remoted` never observes it missing under real startup conditions) — so the WARNING is only ever misleading on a worker, never on a master.

## Why not just silence it entirely?

Because "expected during a normal join" and "wrong forever" are different things, and the log line is the only thing that would ever tell an operator the difference. A worker whose cluster registration is broken, whose network to the master is down, or whose sync is failing for some other real reason will *also* show a missing `authd.pass` — and that case must still be visible. So the fix can't just downgrade the level unconditionally; it has to know the difference between "still within the normal join window" and "stuck."

## The design

1. **A bounded grace window, not a permanent downgrade.** `PasswordKeySource` now takes an `isWorkerNode` flag (threaded from the already-computed `enrollConfig.isWorkerNode` — nothing new to compute) and records its own construction time. A pure, static decision function, `shouldWarnAboutMissingPassword(isWorkerNode, elapsedSinceConstruction)`, is the single source of truth for whether a given missing-file moment should log at WARNING or INFO:
   - Master: always WARNING (unconditionally `true`) — matches today's behavior and reproduction's finding that this should never legitimately fire there.
   - Worker: INFO for the first `kWorkerJoinGraceSeconds` (60s — generous relative to the ~10-20s observed, to absorb real-world jitter without hiding a real problem for long), then WARNING from then on, indefinitely, until the file becomes available.

2. **All three original log sites use the same decision, not just one.** Manual testing during review caught that an earlier draft only gated the first and third lines — the middle one (`Could not watch ... for changes`) was still an unconditional WARNING, which would have left one alarming line in every normal worker join. All three now go through `shouldWarnAboutMissingPassword`. The middle one is additionally guarded on `errno == ENOENT` specifically — any *other* `inotify_add_watch()` failure (permission denied, hitting `fs.inotify.max_user_watches`, etc.) is a genuinely different, unrelated problem and must keep warning regardless of node role or timing.

3. **Testable without capturing real log output.** `PasswordKeySource` lives in `libremoted_module.so`, and this module's logging global is deliberately hidden per-DSO (see `loggerHelper.h`), so there's no existing way for this test file to intercept real `LOGFN_WARN`/`LOGFN_INFO` calls (confirmed: neither `passwordKeySource_test.cpp` nor the sibling `keystore_test.cpp` have ever done this). Extracting the WARN/INFO decision into a small `public static` pure function sidesteps that entirely — it's directly unit-testable with fabricated durations, no waiting, no cross-DSO log plumbing. The four new tests exercise exactly the three states that matter: master always warns, worker doesn't warn inside the window, and — the one that matters most for not silently regressing this fix — a worker still warns once genuinely past the window.

## Manual verification

Both properties were verified live on a real 2-node cluster after deploying this fix (full logs in the linked issue comment):
- A normal worker restart now produces **zero** WARNING lines for `authd.pass` and still recovers in the same ~10-15s window.
- A worker started with `wazuh-manager-clusterd` deliberately left out (so the file can never sync) correctly logs INFO during the window, then **escalates back to WARNING** once the 60s window elapses — confirming the fix cannot silently hide a genuinely stuck worker.

MiguelazoDS:
> [!NOTE]
> The previous workaround to avoid reporting warnings during grace sync windows was kept. No mechanism was added for client.keys since it was only reported for authd.pass. At this moment I don't know if the QA team will change the DTT implementation to test this. Probably the new endpoint will be used, but the previous workaround does not affect that.

MiguelazoDS:
# Manual QA — remoted `GET /status` readiness endpoint (#38605)

[manual-qa.tar.gz](https://github.com/user-attachments/files/31578371/manual-qa.tar.gz)

## One-time setup

```bash
# Build (this checkout, current HEAD)
cd /home/vagrant/wazuh-5.x/src
export CMAKE_OPTS="-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache"
make -j$(nproc) DEBUG=1 TEST=1 TARGET=server   # builds remoted_module + remoted_module_utest

# Build the standalone harness against it (links the real libremoted_module.so)
bash /home/vagrant/wazuh-5.x/.claude/working_issues/38605-remoted-readiness-endpoint/manual-qa/build_harness.sh

# Shared shell helpers used by the tests below (source once per shell)
MQ=/home/vagrant/wazuh-5.x/.claude/working_issues/38605-remoted-readiness-endpoint/manual-qa
SOCK=/tmp/rqa1/queue/sockets/remote-admin-http.sock
run_harness() { ( cd /tmp/rqa1 && exec "$MQ/standalone_harness" "$@" --seconds 120 > harness.log 2>&1 ) & echo $! > /tmp/rqa1/pid; sleep 1; }
stop_harness() { kill "$(cat /tmp/rqa1/pid)" 2>/dev/null; wait "$(cat /tmp/rqa1/pid)" 2>/dev/null; }

# Python tests need this repo's framework on PYTHONPATH -- bare python3 on this
# host resolves `wazuh` to a stale sibling checkout at /home/vagrant/wazuh otherwise.
export PYTHONPATH=/home/vagrant/wazuh-5.x/framework:/home/vagrant/wazuh-5.x/api
```

## Functional

### T1. C++ unit tests (gtest)
Confirms the new `/status` route, both `Keystore::lastLoadOk()` cases, and
`POST /status` → 405 all pass against the real compiled code.
```bash
cd /home/vagrant/wazuh-5.x/src/build/remoted/remoted_module/test
./remoted_module_utest --gtest_filter='AdminServerTest.*:KeystoreTest.LastLoadOkTracksSuccessAndFailure:KeystoreTest.EmptyFileIsALoadedStateNotAFailedOne'
```
**Output:** PASS. `[PASSED] 9 tests.` — all 7 `AdminServerTest` cases
(including `GetStatusReportsKeystoreAndPasswordReadiness`,
`GetStatusOmitsEnrollmentPasswordWhenPasswordModeDisabled`, and
`UnknownRouteAnswers404AndWrongVerb405` which covers `POST /status` → 405)
and both `KeystoreTest` cases. (Harmless `libgcov profiling error:
overwriting an existing profile data with a different checksum` noise
follows — stale coverage instrumentation from a prior build's source
checksums, not a test failure.)

### T2. Python unit tests (pytest)
Confirms `RemotedHTTPClient` and `_remoted_status()`/`get_status()` coverage
passes (fully mocked, no socket needed).
```bash
cd /home/vagrant/wazuh-5.x
python3 -m pytest framework/wazuh/core/tests/test_engine_http.py framework/wazuh/tests/test_manager.py -q
```
**Output:** PASS. `90 passed, 1343 warnings in 4.37s` (all warnings are
pre-existing deprecation noise from `connexion`/`jsonschema`/pytest's own AST
rewriter, unrelated to this change). Higher than the 78 the superseded report
cited — other PRs added tests to these files since.

### T3. Baseline smoke test (password mode disabled)
Confirms `GET /`, `/metrics`, `/status` response shapes, and that
`enrollment_password` is omitted entirely when Password-mode is off.
```bash
rm -rf /tmp/rqa1; mkdir -p /tmp/rqa1/queue/sockets /tmp/rqa1/etc
touch /tmp/rqa1/etc/client.keys
run_harness
curl -s --unix-socket "$SOCK" http://localhost/
curl -s --unix-socket "$SOCK" http://localhost/metrics | head -c 300
curl -s --unix-socket "$SOCK" http://localhost/status
stop_harness
```
**Output:** PASS.
```
GET /       -> {"status":"ok","module":"remoted_module"}
GET /metrics-> {"name":"remoted","timestamp":"...","metrics":[{"name":"remoted.admin.server.budget.available.bytes",...
GET /status -> {"keystore":{"ready":true,"agents_loaded":0,"entries_skipped":0},"ready":true}
POST /status-> 405
```
`enrollment_password` is correctly absent (Password mode off). Note: while
running this the first time, `run_harness`'s original `pid` capture wrote to
the wrong directory (a subshell/backgrounding quirk — `cd X && cmd &` only
backgrounds the `&&` pair, not the earlier `cd`, so the subsequent `echo $!`
ran in the original cwd, not `/tmp/rqa1`), leaving the harness process
orphaned after `stop_harness` silently failed to find the pid file. Found
and killed the stray process, fixed the helper (now `( cd X && exec cmd ) &`
so the backgrounded subshell's own pid is what `$!` captures, written to an
absolute path), and reverified clean start/stop before continuing — this is
the "One-time setup" block above, already corrected.

## Keystore & password behavior

### T4. Keystore lifecycle + startup log severity
Confirms never-loaded / empty-but-loaded / has-entries are correctly
distinguished, `ready` flips live as `client.keys` is populated, and checks
the startup-unreadable log severity (`LOGLEVEL_ERROR`=3 on current `HEAD` —
the superseded report observed `LOGLEVEL_WARNING`=2 on the abandoned commit).
```bash
rm -rf /tmp/rqa1; mkdir -p /tmp/rqa1/queue/sockets /tmp/rqa1/etc   # client.keys deliberately NOT created
run_harness
echo "-- state 1: client.keys missing entirely --"
curl -s --unix-socket "$SOCK" http://localhost/status
grep "Could not read" /tmp/rqa1/harness.log
touch /tmp/rqa1/etc/client.keys
sleep 12
echo "-- state 2: client.keys present but empty --"
curl -s --unix-socket "$SOCK" http://localhost/status
echo "001 test-agent any $(openssl rand -hex 32)" >> /tmp/rqa1/etc/client.keys
sleep 12
echo "-- state 3: client.keys has one entry --"
curl -s --unix-socket "$SOCK" http://localhost/status
stop_harness
```
**Output:** PASS.
```
state 1 (client.keys missing):        {"keystore":{"ready":false,"agents_loaded":0,"entries_skipped":0},"ready":false}
  log: level=3 tag=wazuh-manager-remoted:keystore: Could not read 'etc/client.keys' at startup (errno=2); ...
state 2 (client.keys empty, loaded):  {"keystore":{"ready":true,"agents_loaded":0,"entries_skipped":0},"ready":true}
state 3 (client.keys has 1 entry):    {"keystore":{"ready":true,"agents_loaded":1,"entries_skipped":0},"ready":true}
```
`level=3` = `LOGLEVEL_ERROR` (`src/shared_modules/utils/loggerHelper.h`),
confirming the behavioral delta: this is `LOGFN_ERROR` on current `HEAD`,
where the superseded report observed `LOGFN_WARN` (`level=2`,
`LOGLEVEL_WARNING`) on the abandoned commit. States 1 vs. 2 correctly
disambiguate "never loaded" from "empty but loaded" (same `agents_loaded:0`,
different `ready`), and state 3 shows `ready` following `client.keys` live.

### T5. Password lifecycle
Confirms `enrollment_password.ready` starts `false` with no `authd.pass` and
flips `true` once the file appears, with overall `ready` following.
```bash
rm -rf /tmp/rqa1; mkdir -p /tmp/rqa1/queue/sockets /tmp/rqa1/etc
touch /tmp/rqa1/etc/client.keys
run_harness --enroll-password
echo "-- authd.pass absent --"
curl -s --unix-socket "$SOCK" http://localhost/status
echo "dummy-password" > /tmp/rqa1/etc/authd.pass
sleep 14   # authd.pass doesn't exist at startup, so the initial inotify watch setup
           # fails (errno=2) and reload falls back to the 10s poll only -- 14s clears it
echo "-- authd.pass now present --"
curl -s --unix-socket "$SOCK" http://localhost/status
stop_harness
```
**Output:** PASS.
```
authd.pass absent:  {"keystore":{...},"enrollment_password":{"ready":false},"ready":false}
authd.pass present: {"keystore":{...},"enrollment_password":{"ready":true},"ready":true}
```
First attempt used `sleep 8` (matching the superseded report's observed
"t+6s") and got a false negative — `enrollment_password.ready` was still
`false`. Checked `harness.log`: it *did* eventually log "Enrollment password
reloaded from 'etc/authd.pass'", just later than 8s. Root cause: since
`authd.pass` doesn't exist at startup, the initial inotify watch setup fails
(`errno=2`, logged), so reload falls back entirely to the 10s poll — with no
inotify to catch the write early, 8s was too tight a margin against that
10s-anchored cadence. This is a test-timing issue, not a product defect;
fixed by widening the sleep to 14s (command above already reflects the fix).

## Legacy / backward-compatibility

### T6. Legacy code path untouched (diff-based)
Confirms zero files under the legacy C TCP/UDP pipeline (`src/remoted/src/`)
are touched, and that the diff to the shared facade file only adds new code
(no existing `GET /`/`GET /metrics` handler line changed).
```bash
cd /home/vagrant/wazuh-5.x
git show 9bf3829e54 --stat | grep -c '^ src/remoted/src/'   # expect 0
git show 9bf3829e54 -- src/remoted/remoted_module/src/remotedModuleFacade.hpp
```
**Output:** PASS.
- `git show --stat | grep -c '^ src/remoted/src/'` → `0`. Confirmed: the
  legacy C TCP/UDP pipeline is untouched by this commit.
- Full diff of `remotedModuleFacade.hpp` is purely additive: two new
  `#include`s, one new member-field pair, one new
  `registerPasswordKeySourceDiagnostics()` method, one new `addRoute()` call
  for `/status`, and a one-line startup-log string change ("routes: GET /
  and GET /metrics" → "...GET /, GET /metrics, and GET /status"). Zero lines
  inside the existing `GET /`/`GET /metrics` handler lambdas are touched.

## Performance / non-blocking

### T7. Concurrent scrape load
Proves `/status` adds no measurable latency tax to the shared 2-thread admin
reactor under a realistic scrape rate (20 workers/route × 3 routes, 15s).
```bash
rm -rf /tmp/rqa1; mkdir -p /tmp/rqa1/queue/sockets /tmp/rqa1/etc
touch /tmp/rqa1/etc/client.keys
run_harness
python3 "$MQ/load_test.py" "$SOCK" 15 20
stop_harness
```
**Output:** PASS.
```
/         : n=8577  errors=0  mean=33.56ms  p50=27.45ms  p95=81.03ms  p99=118.14ms  max=367.77ms
/metrics  : n=7331  errors=0  mean=39.06ms  p50=33.25ms  p95=86.34ms  p99=120.93ms  max=319.44ms
/status   : n=8401  errors=0  mean=33.99ms  p50=28.48ms  p95=80.79ms  p99=112.60ms  max=258.49ms
Total: 24309 requests, ~1621 req/s aggregate, 0 errors
```
`/status`'s latency profile is statistically indistinguishable from `/`
(bare liveness) and better than `/metrics` — no measurable tax from the new
route.

### T8. Burst load (3x the session cap)
Proves overload sheds cleanly (`503`s), no hang/crash, and `/status` neither
starves nor is starved by `/`/`/metrics` (60 workers/route × 3 routes, 10s —
~3x the 64-session `RouteClass::Liveness` cap).
```bash
rm -rf /tmp/rqa1; mkdir -p /tmp/rqa1/queue/sockets /tmp/rqa1/etc
touch /tmp/rqa1/etc/client.keys
run_harness
python3 "$MQ/load_test.py" "$SOCK" 10 60
stop_harness
```
**Output:** PASS.
```
/         : n=9912  errors=4   mean=98.35ms   p95=265.15ms  max=663.47ms
/metrics  : n=7855  errors=3   mean=116.72ms  p95=291.85ms  max=819.04ms
/status   : n=7213  errors=1   mean=106.02ms  p95=282.93ms  max=752.28ms
Total: 24980 requests, ~2498 req/s aggregate, 8 errors (0.03%)
```
`load_test.py` doesn't record status codes for errors, so re-ran the same
180-concurrent/10s profile with a small instrumented script to classify
them: every error was a clean `{"error":"Too many concurrent
connections","code":503}` — `('/',503)=6, ('/metrics',503)=15,
('/status',503)=16` — split near-evenly across the three routes (they share
one 64-session `RouteClass::Liveness` cap), never a hang/crash/malformed
response. The harness kept serving normally (`GET /` → 200) immediately
after the burst.

### T9. Load + concurrent `client.keys` churn
Proves the `/status` handler's mutex-guarded reads never block even while
the watcher thread is actively reloading the keystore.
```bash
rm -rf /tmp/rqa1; mkdir -p /tmp/rqa1/queue/sockets /tmp/rqa1/etc
touch /tmp/rqa1/etc/client.keys
run_harness
( for i in $(seq 1 20); do echo "00$i x any $(openssl rand -hex 32)" > /tmp/rqa1/etc/client.keys; sleep 0.3; done ) &
CHURN_PID=$!
python3 "$MQ/load_test.py" "$SOCK" 6 20
wait $CHURN_PID
stop_harness
```
**Output:** PASS.
```
/         : n=3674  errors=0  mean=36.16ms  p95=87.35ms  max=187.67ms
/metrics  : n=3039  errors=0  mean=42.79ms  p95=100.08ms  max=238.58ms
/status   : n=3542  errors=0  mean=35.47ms  p95=84.04ms  max=189.72ms
Total: 10255 requests, ~1709 req/s aggregate, 0 errors
```
Zero errors across ~10k requests while `client.keys` was rewritten every
0.3s for the full 6s — the `/status` handler's mutex-guarded reads never
blocked while the watcher thread was actively reloading.

## Worst-case

### T10. Socket failure modes + `get_status()` degrade
Proves missing/dangling/hung/permission-denied admin sockets all degrade to
`{'ready': False}` through `RemotedHTTPClient`/`_remoted_status()`/
`get_status()` with no unhandled exception. No harness needed — each case
builds its own throwaway socket. (The permission-denied case skips itself
unless run as root.)
```bash
cd /home/vagrant/wazuh-5.x
python3 "$MQ/python_worstcase.py"
```
**Output:** PASS. All 5 cases degrade to `{'ready': False}` with no unhandled
exception:
- A (ENOENT): `code=2031 [Errno 2] No such file or directory`
- B (dangling socket): `code=2031 [Errno 111] Connection refused`
- C (hung listener): `code=2030` timeout, raised after 1.54s (configured 1.5s)
- D (permission denied, run as root so the non-root sub-check executed): child
  raised `code=2031 [Errno 13] Permission denied`
- E (full `get_status()`, remoted unreachable): no exception escaped;
  `remoted entry: {'ready': False, 'running': True}`, node `ready: False`.

### T11. Full-stack happy path, real socket
Proves `RemotedHTTPClient` → `_remoted_status()` → `get_status()` all
correctly parse a real, live response (not mocked at the socket boundary).
```bash
rm -rf /tmp/rqa1; mkdir -p /tmp/rqa1/queue/sockets /tmp/rqa1/etc
touch /tmp/rqa1/etc/client.keys
echo "dummy-password" > /tmp/rqa1/etc/authd.pass
run_harness --enroll-password
cd /home/vagrant/wazuh-5.x
python3 "$MQ/python_happy_path_real_socket.py"
stop_harness
```
**Output:** PASS. `authd.pass` written before harness startup so it loads
synchronously (no poll wait needed). All script assertions passed:
```
raw response:        {'keystore': {'ready': True, ...}, 'enrollment_password': {'ready': True}, 'ready': True}
_remoted_status(...) -> {'ready': True, 'keystore': {...}, 'enrollment_password': {'ready': True}}
wazuh-manager-remoted entry: {'ready': True, 'running': True, 'keystore': {...}, 'enrollment_password': {'ready': True}}
```
(Node-level `ready` printed `False` because the *other* mocked daemons in
this script's stub data aren't asserted as ready — irrelevant to what this
test targets, which is `RemotedHTTPClient`/`_remoted_status()` parsing a
real live response correctly, which it does.)

## Regression

### T12. `get_status` family + full framework suite
Confirms no pre-existing `get_status()` assertions (analysisd/modulesd
entries) or unrelated framework tests broke.
```bash
cd /home/vagrant/wazuh-5.x
python3 -m pytest framework/wazuh/tests/test_manager.py -k get_status -vv
python3 -m pytest framework -q
```
**Output:** PASS.
- `get_status`-filtered: `14 passed, 37 deselected` — matches the
  superseded report's count exactly; no pre-existing `get_status` assertion
  needed updating for this commit.
- Full `framework` suite: `2024 passed, 6 skipped, 787 warnings in 216.01s`
  (up from the superseded report's 1973/6 — other PRs added tests since;
  0 failures, 0 unexpected skips).

## Housekeeping (flag only — no action without asking)

- Leftover worktree `.claude/worktrees/qa-38605-before` (pinned to
  `c342b0fb30`, an unrelated commit) — candidate for removal.
- Stray branches `bug/38605-authd-pass-not-readable-at-startup-{backup,backup2}`.
- `CHANGELOG.md` still has no entry for this change — `changes.md` records
  this as an explicit user decision to handle separately; re-flagging only,
  since `.claude/CLAUDE.md` marks a CHANGELOG entry mandatory for CI.

## Overall result

**PASS — all 12 tests green against the real, pushed `HEAD` (`9bf3829e54`).**
No product defect found. Two issues were found and fixed during this pass,
both in the QA tooling, not the feature:
1. `run_harness`'s pid-capture wrote to the wrong directory (a
   `cd X && cmd &` subshell-backgrounding quirk), leaving one orphaned
   harness process after T3's first run — killed and the helper fixed
   (now `( cd X && exec cmd ) &`, absolute pid path).
2. T5's first attempt used an 8s wait for `authd.pass` to be picked up;
   since the file doesn't exist at startup, inotify watch setup fails and
   reload falls back to a 10s poll only — 8s was too tight. Fixed to 14s.

The one real, new finding from this pass (not in the superseded report):
`keystore.cpp`'s startup-unreadable log line is confirmed `LOGFN_ERROR`
(`level=3`) on current `HEAD`, not `LOGFN_WARN` as the abandoned-commit
report observed — see T4.

jotacarma90 (approved):
LGTM

PR #38688, opened 2026-08-28
Validate API pagination URLs before following them

vikman90:
## Description
The GitHub, Office 365 and MS Graph wodles paginated a remote API by taking the next-page URL from the previous response — an HTTP header in the first two, the JSON body in the last — and re-issuing the request to it verbatim, with the same `Authorization` header still attached. A next-page value pointing off-host therefore made the manager request that host while forwarding the configured GitHub personal access token or Microsoft bearer token. This change pins every next-page URL to the API host the module was configured with and drops the page when it does not match.

The Office 365 `contentUri` taken from the response body is pinned to the same host: it arrives in the same response as the next-page URL and is fetched with the same bearer token. When a URL is rejected the scan is marked as failed, so the bookmark is not advanced over the pages that were skipped and the window is retried instead of being silently lost.

## Proposed Changes
- `src/wazuh_modules/wmodules.c`, `src/wazuh_modules/wmodules.h`: added `wm_url_is_allowed()`, which requires the `https` scheme and, when a host is given, that the URL points to exactly that host. Scheme and host are compared case-insensitively, and an explicit port is accepted.
- `src/wazuh_modules/wm_github.h`: factored the API host out of `GITHUB_API_URL` into `GITHUB_API_HOST`.
- `src/wazuh_modules/wm_github.c`: validate the `Link` `rel="next"` URL against `GITHUB_API_HOST` before requesting it; stop pagination otherwise.
- `src/wazuh_modules/wm_office365.c`: validate the `NextPageUri` header and the `contentUri` from the response body against the configured `management_fqdn`.
- `src/wazuh_modules/wm_ms_graph.c`: validate `@odata.nextLink` against the configured `query_fqdn` at both pagination sites, and clamp the `Retry-After` value taken from a 429 response to `WM_MS_GRAPH_MAX_RETRY_AFTER_WARN`, which was previously only warned about and could stall the module thread indefinitely.
- `src/shared/url.c`: restricted `wurl_http_request()` to the `http` and `https` protocols, so a URL with another libcurl-supported scheme is no longer handed to libcurl.

### Results and Evidence
The affected CMocka suites pass:

```
Test project src/unit_tests/build
    Start 101: test_wmodules
1/4 Test #101: test_wmodules ....................   Passed    0.01 sec
    Start 103: test_wm_github
2/4 Test #103: test_wm_github ...................   Passed    0.02 sec
    Start 104: test_wm_office365
3/4 Test #104: test_wm_office365 ................   Passed    0.04 sec
    Start 105: test_wm_ms_graph
4/4 Test #105: test_wm_ms_graph .................   Passed    0.04 sec

100% tests passed, 0 tests failed out of 4
```

With the module changes reverted and the tests kept, the new regression tests fail as expected:

```
[  FAILED  ] test_github_execute_scan_next_page_out_of_host
[  FAILED  ] test_wm_ms_graph_scan_relationships_next_page_out_of_host
[  FAILED  ] test_wm_ms_graph_http_get_with_retry_429_retry_after_exceeds_warn
[  FAILED  ] test_wm_office365_execute_scan_content_uri_out_of_host
```

### Artifacts Affected
`wazuh-modulesd` on the manager, and any binary linking `src/shared/url.c`.

### Configuration Changes
None.

### Documentation Updates
None.

### Tests Introduced
- `src/unit_tests/wazuh_modules/wmodules/test_wmodules.c`: eighteen cases for `wm_url_is_allowed()`, covering the matching host, non-HTTPS and non-HTTP schemes, a different host, the host as a suffix (`api.github.com.evil.tld`), the host as a prefix (`api.github.co`), the host in the userinfo (`api.github.com@evil.tld`), the host in the query string, an explicit and an empty port, a port followed by userinfo (`api.github.com:1234@evil.tld`), query and fragment terminators, uppercase scheme and host, and the host-less form.
- `src/unit_tests/wazuh_modules/office365/test_wm_office365.c`: `test_wm_office365_execute_scan_content_uri_out_of_host`, which asserts that an off-host `contentUri` is not fetched.
- `src/unit_tests/wazuh_modules/github/test_wm_github.c`: `test_github_execute_scan_next_page_out_of_host`, which serves a full page with an off-host `Link` `rel="next"` and asserts that only one request is issued.
- `src/unit_tests/wazuh_modules/ms_graph/test_wm_ms_graph.c`: `test_wm_ms_graph_scan_relationships_next_page_out_of_host`, the equivalent for `@odata.nextLink`. The existing two-page tests now use a next-page URL on the configured Graph host, and `test_wm_ms_graph_http_get_with_retry_429_retry_after_exceeds_warn` now asserts the clamped delay.

## Credits
Reported by [aka-STHIRAA](https://github.com/aka-STHIRAA). The MS Graph module and the `Retry-After` handling were found during triage of that report.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/src/wazuh_modules/wm_office365.c
+++ b/src/wazuh_modules/wm_office365.c
@@ -426,8 +426,9 @@ STATIC void wm_office365_execute_scan(wm_office365* office365_config, int initia
                             cJSON *blob = cJSON_GetArrayItem(blobs_array, i);
                             cJSON *content = cJSON_GetObjectItem(blob, "contentUri");
 
-                            if (content && (content->type == cJSON_String) && !wm_url_is_allowed(content->valuestring, NULL)) {
-                                mtwarn(WM_OFFICE365_LOGTAG, "Ignoring non-HTTPS content URI.");
+                            if (content && (content->type == cJSON_String) && !wm_url_is_allowed(content->valuestring, current_auth->management_fqdn)) {
+                                mtwarn(WM_OFFICE365_LOGTAG, "Ignoring content URI out of '%s'.", current_auth->management_fqdn);
+                                fail = 1;
                             } else if (content && (content->type == cJSON_String)) {
                                 cJSON *logs_array = NULL;
 
--- a/src/wazuh_modules/wm_office365.c
+++ b/src/wazuh_modules/wm_office365.c
@@ -426,7 +426,9 @@ STATIC void wm_office365_execute_scan(wm_office365* office365_config, int initia
                             cJSON *blob = cJSON_GetArrayItem(blobs_array, i);
                             cJSON *content = cJSON_GetObjectItem(blob, "contentUri");
 
-                            if (content && (content->type == cJSON_String)) {
+                            if (content && (content->type == cJSON_String) && !wm_url_is_allowed(content->valuestring, NULL)) {
+                                mtwarn(WM_OFFICE365_LOGTAG, "Ignoring non-HTTPS content URI.");
+                            } else if (content && (content->type == cJSON_String)) {
                                 cJSON *logs_array = NULL;
 
                                 if (logs_array = wm_office365_get_logs_from_blob(content->valuestring, access_token, office365_config->curl_max_size, &buffer_size_reached, &error_msg), logs_array) {
@@ -470,6 +472,11 @@ STATIC void wm_office365_execute_scan(wm_office365* office365_config, int initia
                         if (!scan_finished) {
                             if ((next_page == NULL) || (strlen(next_page) >= OS_SIZE_8192)) {
                                 scan_finished = 1;
+                            } else if (!wm_url_is_allowed(next_page, current_auth->management_fqdn)) {
+                                mtwarn(WM_OFFICE365_LOGTAG, "Ignoring next page URL out of '%s'.", current_auth->management_fqdn);
+                                os_free(next_page);
+                                scan_finished = 1;
+                                fail = 1;
                             } else {
                                 snprintf(url, sizeof(url), "%s", next_page);
                                 os_free(next_page);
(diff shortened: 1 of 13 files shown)


src/wazuh_modules/wm_office365.c
```diff
@@ -426,7 +426,9 @@ STATIC void wm_office365_execute_scan(wm_office365* office365_config, int initia
                             cJSON *blob = cJSON_GetArrayItem(blobs_array, i);
                             cJSON *content = cJSON_GetObjectItem(blob, "contentUri");
 
-                            if (content && (content->type == cJSON_String)) {
+                            if (content && (content->type == cJSON_String) && !wm_url_is_allowed(content->valuestring, current_auth->management_fqdn)) {
+                                mtwarn(WM_OFFICE365_LOGTAG, "Ignoring content URI out of '%s'.", current_auth->management_fqdn);
+                            } else if (content && (content->type == cJSON_String)) {
```
MarcelKemp, line 432: Rejecting an off-host `contentUri` warns but leaves `fail` at 0, so the window's bookmark still advances past audit records that were never fetched and that window is never re-requested.

```suggestion
                            if (content && (content->type == cJSON_String) && !wm_url_is_allowed(content->valuestring, current_auth->management_fqdn)) {
                                mtwarn(WM_OFFICE365_LOGTAG, "Ignoring content URI out of '%s'.", current_auth->management_fqdn);
                                fail = 1;
                            } else if (content && (content->type == cJSON_String)) {
```

MarcelKemp (changes requested):

src/wazuh_modules/wm_office365.c
vikman90: Good catch, applied in 954394d58f.

You are right, and it was an inconsistency on my side: the `fail` flag was already set on the three next-page rejection paths for exactly this reason, and I missed it when adding the `contentUri` host check in a later commit.

I took the suggestion as written, with only `fail = 1` and no `scan_finished = 1`. Since the loop is a `for` over the blobs of the page, the remaining legitimate blobs are still ingested, and the `if (fail)` at the end still triggers `wm_office365_scan_failure_action()` and breaks out of the time window, so the bookmark is not advanced. Copying the `scan_finished = 1` from the pagination sinks would have discarded valid blobs unnecessarily.

Two notes:

- As with the pagination sinks, retrying the window means the logs already forwarded from the valid blobs of that window are sent again. Duplicates are preferable to silent loss, but it is worth being explicit about it.
- `test_wm_office365_execute_scan_content_uri_out_of_host` was asserting the buggy behaviour (it expected the `WM_IO_WRITE` and the "Bookmark updated to ..." line). It now looks up the failure record with `wm_office365_get_fail_by_tenant_and_subscription()` and asserts the bookmark is not advanced. I checked it fails without the one-line fix and passes with it.

Full unit suite: 86/86.

MarcelKemp (approved):

Issue #38658, opened 2026-08-27
WPK upgrade scripts can't distinguish absent <endpoint> from explicit empty opt-out


jpcerrone:
## Description

`src/init/pkg_installer.sh` and `src/win32/do_upgrade.ps1` (the WPK upgrade connectivity-probe scripts added/updated in #38614) cannot distinguish an **absent** `<endpoint>` tag from an **explicit empty opt-out** (`<endpoint></endpoint>`) when reading the agent's already-written `ossec.conf` during an upgrade. Both cases collapse to the default `wazuh-manager` prefix, silently overriding a deliberate opt-out.

This is the same class of bug already fixed for the *install*-time scripts (`register_configure_agent.sh`/`inst-functions.sh`, via `${VAR+x}` on the env var) — but it was missed in the two upgrade-time scripts, which parse XML already on disk rather than an env var, so the `${VAR+x}` technique doesn't apply directly.

## Where

```sql
# src/init/pkg_installer.sh
SERVER_ENDPOINT=$(xml_value agent manager endpoint)
if [ -z "${SERVER_ENDPOINT}" ]; then
    SERVER_ENDPOINT="wazuh-manager"
fi
```

```sql
# src/win32/do_upgrade.ps1
$server_endpoint = get_conf_value "agent" "manager" "endpoint"
if ([string]::IsNullOrEmpty($server_endpoint)) {
    $server_endpoint = "wazuh-manager"
}
```

`xml_value` (grep/sed extraction) and `get_conf_value` (PowerShell regex) both return an empty string whether the `<endpoint>` tag is missing entirely or present-but-empty — there's no way to tell "tag exists, no content" from "tag doesn't exist" with simple text extraction on already-written XML.

## Impact

A 5.x agent that deliberately configured `<endpoint></endpoint>` (opt-out, e.g. because its manager has no `global_prefix`) will have that opt-out silently discarded on its next WPK upgrade: the pre-flight connectivity probe assumes the default `wazuh-manager` prefix instead, probes the wrong URL (`/wazuh-manager/` instead of the root), most likely gets a 404 against the unprefixed manager, and aborts the upgrade of an agent that was actually configured correctly.

Scope is narrow (only the "explicit opt-out + WPK upgrade" combination), but it's real and symmetric on both platforms. Severity: Medium — doesn't break the common (default) case, but silently breaks a case the same PR went out of its way to make explicitly supported everywhere else.

## Suggested fix

[...]

PR #38703, opened 2026-08-28
Unify the manager address, port and prefix into a single <endpoint>

### LINKED ISSUE #38624: Agent-side: unify manager address/port/endpoint into a single connection URL

rovogel:
## Description

Closes #38624

The agent's `<agent><manager>` block spread one connection target across three settings —
`<address>`, `<port>` and a prefix-only `<endpoint>` — plus `<interface_index>` for IPv6
scope. This PR collapses them into a single `<endpoint>` carrying the whole target.

```xml
<agent>
  <manager>
    <endpoint>192.168.0.60:1517/wazuh-manager/</endpoint>
  </manager>
</agent>
```

```
<endpoint> = [ "https://" ] host [ ":" port ] [ "/" [ prefix ] ]

host   IPv4, hostname, or bracketed IPv6 (optionally %25<zone>)   REQUIRED
port   1-65535                                                    default 1517
prefix reverse-proxy path segments                                default wazuh-manager
```

Only the address is mandatory, so `192.168.0.60` behaves exactly like
`192.168.0.60:1517/wazuh-manager/`. The installers write an operator's value through
**verbatim**, so `WAZUH_MANAGER_ENDPOINT` and the `<endpoint>` tag are one language — one
grammar to document, test and support.

`WAZUH_MANAGER` and `WAZUH_MANAGER_PORT` are **kept**: when `WAZUH_MANAGER_ENDPOINT` is not
set, an equivalent `<endpoint>` is composed from them, so every existing install command and
dashboard snippet keeps working.

```
WAZUH_MANAGER_ENDPOINT set    -> wins outright
WAZUH_MANAGER=1.2.3.4         -> <endpoint>1.2.3.4</endpoint>
  + WAZUH_MANAGER_PORT=8443   -> <endpoint>1.2.3.4:8443</endpoint>
WAZUH_MANAGER=2001:db8::1     -> <endpoint>[2001:db8::1]</endpoint>   (bracketed)
WAZUH_MANAGER_IP=…            -> still aliases WAZUH_MANAGER
```

Two composition details that are easy to get wrong: a bare IPv6 literal is **bracketed**
before a port is appended, or its trailing group would be read as the port; and an omitted
port is **left out entirely** rather than written as `1517`, so the value stays the shortest
spelling of what was asked for and the agent applies its own default.

**Priority is absolute, including when the value is rejected.** Once `WAZUH_MANAGER_ENDPOINT`
is set it decides the outcome even if it fails validation — falling back to `WAZUH_MANAGER`
there would silently point the agent at a different manager than the operator named. The
package installers write no `<manager>` block at all (the shipped placeholder remains, so the
agent fails loudly at startup) and the source installer writes the value through for the
startup parser to refuse.

### The rule that carries the weight

**No `/` at all means the default prefix; a trailing `/` with nothing after it is the
deliberate opt-out.** libcurl cannot express this distinction — `CURLUPART_PATH` reports `/`
for both — so the separator is detected in the raw value before parsing. Losing it would turn
every opt-out back into the default.

`192.168.0.60` and `192.168.0.60/` therefore mean different things, and that one character is
the only difference between an agent that connects to a prefixed manager and one that 404s.

**Agent and manager agree on how "no prefix" is spelled.** An empty value is an error on both
sides — the manager already rejected an empty `<global_prefix>` — and both express "serve or
expect no prefix" as a path of just `/`. A value mirrored from one configuration into the
other keeps its meaning instead of being fatal on one side and silently different on the other.

## Proposed Changes

**`src/config/src/client-config.c`** — `w_parse_agent_endpoint()` parses `<endpoint>` through
libcurl's URL API, already linked here (`shared/src/url.c` uses it), which brings host/port
validation, IPv6 brackets and zone-id extraction with it. A scheme-less value gets `https://`
prepended so one parser serves both spellings. `CURLU_DISALLOW_USER` rejects embedded
credentials; **`CURLU_PATH_AS_IS` is required** — without it curl resolves dot-segments, so
`/../etc` silently becomes `/etc` instead of being rejected. The path remainder feeds the
existing, unmodified `w_normalize_agent_endpoint()`.

Two things curl does *not* validate, both of which reached the agent before:

- **The port is range-checked here.** curl accepts `:0` and goes on to connect, so
  `<endpoint>10.0.0.5:0</endpoint>` yielded port 0 while every installer parser rejects
  anything below 1.
- **A trailing `:` is rejected.** curl reads `10.0.0.5:` as "no port" and the agent silently
  fell back to 1517, a value the installers reject and `test_register_configure_agent.sh`
  asserts they reject.

`w_resolve_ipv6_zone()` resolves a zone id with `if_nametoindex()` at parse time, so an
unknown interface is a startup error naming it rather than an opaque libcurl failure later. A
numeric zone is parsed with `strtoul` and `errno`: the previous `strtol` guard compared against
`UINT32_MAX`, which is unreachable wherever `long` is 32-bit — both MinGW targets — so an
overflowing value saturated to `LONG_MAX` and was accepted.

**`<endpoint>` outranks the deprecated pair**, whatever order they appear in, with a warning
naming what was ignored. Order mattered previously: `<address>` and `<endpoint>` both assigned
the host, and a `<port>` written after `<endpoint>` silently overwrote the port parsed out of
it. They feed separate slots now and the winner is picked once, after the loop.

**A `<client><server><endpoint>` is read.** The MSI reconfigures a preserved 4.x `ossec.conf`
in place and, having no `<agent>` block to target, writes the endpoint into `<client><server>`.
`Read_Legacy_Client_Address()` skipped every child that was not `<address>`, so an agent
upgraded that way had no manager at all and refused to start with `(4105): No valid server IP
found.` That block now reads `<endpoint>` through the same parser, so a port and prefix written
there survive.

**Struct and ABI** — `agent_server.network_interface` becomes `scope_id`; `hc_config_t` gains
`server_scope_id`; the bridge copies it; `ModuleConfig::baseUrl()` emits `[fe80::1%253]`
(`%25` being a literal `%` in a URL). Note this is an ABI addition: a partial rebuild
misreads the struct.

`Validate_IPv6_Link_Local_Interface()` is **kept**, now validating the zone parsed from
`<endpoint>`. It runs at agent startup from `main.c` and `win_utils.c`.

**Signing is unaffected.** `baseUrl()` still excludes the prefix because `prefixedTarget()`
folds it into the signed request target; `endpoint` carries the same normalized string as
before, so `canonicalRequest`, `retrySender` and `enrollClient` are untouched.

**Log messages name `<endpoint>`.** A stock configuration logged
`<agent><manager><port> is not configured` on every start — a tag that no longer exists, so an
operator following it goes looking for something there is nowhere to put. It now reads
`No port in <agent><manager><endpoint>. Using the default port 1517.` Three more messages
advised migrating to `<manager><address>`, the very tag this change deprecates, and now name
`<manager><endpoint>`. The deprecation notice and the "ignored when `<endpoint>` is configured"
warning still name `<address>` and `<port>`, because they fire only when those tags are present.

**Installers** — `register_configure_agent.sh`, `inst-functions.sh` and
`InstallerScripts.vbs` emit one `<endpoint>`. `WAZUH_MANAGER`, `WAZUH_MANAGER_PORT`, the
`WAZUH_MANAGER_IP` alias and the `ADDRESS`/`SERVER_PORT` MSI properties all remain. Three
defects fixed while making the rule uniform across the three:

- `inst-functions.sh` composed a value from `WAZUH_MANAGER` first and only overrode it *if the
  endpoint validated*, so a typo in `WAZUH_MANAGER_ENDPOINT` silently fell back to a different
  manager. It now honours the endpoint unconditionally.
- Both shell installers bracketed a bare IPv6 literal without checking whether it was bracketed
  already, so `WAZUH_MANAGER="[2001:db8::1]"` became `[[2001:db8::1]]:1517` and the agent would
  not start. `InstallerScripts.vbs` has always guarded this; the shell copies now match.
- `WAZUH_MANAGER_PORT` on its own did nothing at all — the port is only appended inside the
  `WAZUH_MANAGER` branch, so with neither that nor the endpoint set no `<manager>` block is
  written and the run succeeded having ignored it. It now says so.

`InstallerScripts.vbs` also validated the port with `CLng()`, which holds a 32-bit Long. Above
2147483647 it overflows, and that error is not catchable there — `On Error Resume Next` is
procedure-scoped and the function declares none — so an overflow abandoned the caller
mid-statement: `config()` stopped at the call, no diagnostic was written, and nothing it would
have configured afterwards happened. The digits are narrowed before `CLng` sees them. Leading
zeros are stripped rather than rejected, so `000080` stays port 80, matching the shell parsers.

**WPK pre-flight** — `pkg_installer.sh` and `do_upgrade.ps1` read two config shapes: a single
`<endpoint>`, or the deprecated pair / 4.x `<client><server><address>`, from which they compose
the same target the agent composes internally. Both stripped the brackets off an IPv6 host and
then built `https://2001:db8::1:1517/…`, which curl, wget and `Invoke-WebRequest` all reject as
malformed — so every IPv6 WPK upgrade aborted with "manager is not reachable" before it began.

**Source upgrades carry the manager across.** `etc/ossec-agent.conf` now ships a single
`<endpoint>`, but `replace_manager_ip.sh` was still hard-coded to `<address>` and
`update.sh` ignored its exit code, so every source-install agent upgrade kept the template's
literal `IP` placeholder. Both paths were broken and the worse one silently: a 4.x previous
config found `<address>` in the old file, ran its `sed` against the new one which no longer has
that tag, matched nothing and reported success. The script now reads `<endpoint>` when present,
carrying it across verbatim so a configured port and prefix survive, and otherwise falls back
to `<address>`, bracketing an IPv6 literal. `<port>` is only taken from an `<agent><manager>`
block: a 4.x `<client><server><port>` is the old TCP remoted port the agent itself ignores, so
carrying it would move the agent to a port nothing serves. `update.sh` now fails the upgrade
instead of continuing with a placeholder.

**Also updated**: shipped default configs, 10 integration-test files, the Windows integration CI
job, and 4 documentation files.

### Results and Evidence

Verified on a replica of the CI runner: Ubuntu 24.04.4, gcc 14.3.0 from the same wazuh
package CI installs (same `/usr/bin` symlink swap), cmake 3.31.6, cmocka 1.1.7.

| Suite | Result |
|---|---|
| Full agent `ctest` | **114/114** |
| `test_client-config_https` | **86/86** |
| `test_register_configure_agent.sh` | **39 checks, 0 failed** |
| `pkg_installer.sh` config-read matrix | **8/8** |
| AStyle over `client-agent/https_client` | clean |

**Real DEB install and enrollment**, against a live 5.0.0 manager with
`global_prefix = /wazuh-manager`. The package was confirmed to carry the change (packaged
installer script byte-identical to source) before any result was trusted. Each case a full
purge plus fresh `dpkg -i`:

| Input | Written | Enrollment | State |
|---|---|---|---|
| `WAZUH_MANAGER_ENDPOINT=192.168.70.100` | `<endpoint>192.168.70.100</endpoint>` | `031` | **connected** |
| `WAZUH_MANAGER=192.168.70.100` | same | `032` | **connected** |
| `WAZUH_MANAGER` + `PORT=1517` | `<endpoint>…:1517</endpoint>` | `033` | **connected** |
| endpoint + `WAZUH_MANAGER=1.1.1.1 PORT=1111` | the endpoint | `034` | **connected** |
| `192.168.70.100/` (opt-out) | verbatim | none | **pending**, 404 |

Row 4 is the precedence proof and row 5 the negative control: the opt-out's 404 shows the
parsed prefix reaches **enrollment**, not merely the data plane.

**The full port × prefix grid**, run against the live manager on one enrolled agent by varying
only the `<manager>` block:

| `<endpoint>` | port notice | prefix | status |
|---|---|---|---|
| `host` | logged | default | connected |
| `host:1517` | silent | default | connected |
| `host/wazuh-manager` | logged | explicit | connected |
| `host:1517/wazuh-manager` | silent | explicit | connected |
| `host/` | logged | none | **pending** |
| `host:1517/` | silent | none | **pending** |

The notice fires on exactly the cases that default a component and nowhere else. A hostname
works in place of an IP: installed with `WAZUH_MANAGER_ENDPOINT=ubuntu-VirtualBox` the agent
enrolled and reached `connected` on `server=ubuntu-VirtualBox:1517`, both defaults applied.

**Backward compatibility, proven on a live agent.** A configuration rewritten to the deprecated
pair — exactly as an upgrade leaves it — stays **connected**, `ossec.conf` is never rewritten
(md5 identical before and after), and one INFO notice carries the exact replacement line:

```
INFO: <agent><manager><address> and <port> are deprecated. Replace them with a single
      <endpoint>192.168.70.100:1517/wazuh-manager</endpoint>
```

A 4.x `<client><server>` configuration gets the same treatment, naming the block to replace.
The suggested line is composed from the **resolved** address, port and prefix, never from the
defaults, so pasting it back cannot change where the agent connects. An empty `<manager></manager>`
is rejected loudly rather than defaulted — `wazuh-agentd -t` exits 1 with
`(1230) Invalid element` / `(1215) No client configured. Exiting.`

**Real RPM install and enrollment** in a Rocky Linux 9.3 container against the same manager,
across four cases: bare endpoint (`040`), `WAZUH_MANAGER` (`041`), endpoint-overrides-both
(`042`), and the opt-out negative control (404, not enrolled). The packaged installer script
is byte-identical to source.

**Real MSI**, built with WiX 3.14 from mingw-cross-compiled artifacts, installed five ways:

| Install properties | Written `<endpoint>` | Enrollment |
|---|---|---|
| `WAZUH_MANAGER_ENDPOINT=192.168.70.100` | verbatim | `046` |
| `…=192.168.70.100:1517/wazuh-manager/` | verbatim | `047` |
| `WAZUH_MANAGER=192.168.70.100` | `<endpoint>192.168.70.100</endpoint>` | `048` |
| both, with `WAZUH_MANAGER=1.1.1.1 PORT=1111` | the endpoint | `049` |
| `192.168.70.100/` (opt-out) | verbatim | none — service runs, cannot enrol |

Row 4 is the precedence proof on Windows: had `WAZUH_MANAGER` leaked through it would have
targeted `1.1.1.1:1111` and never enrolled. The reworded port notice appears on the three cases
that omit a port and is absent from the one that supplies it, confirming the message change on
Windows as well as Linux.

**`replace_manager_ip.sh`** resolves five previous-config shapes: a current `<endpoint>`
(carried verbatim, port and prefix intact), 4.x `<client><server>` (host only, defaults
applied), the deprecated pair (`host:port`), an IPv6 literal (bracketed), and no address at all
(exit 1, unchanged file).

**`ParseManagerEndpoint`** was driven directly under `cscript` using the function's real source.
Before the fix, `host:2147483648` and `host:99999999999` left the caller's variable untouched
(`Err=6`) and wrote no diagnostic, while every other rejection logged one; after, both are
rejected with a diagnostic and `host:000080` still resolves to port `000080`, matching the shell
parsers. `do_upgrade.ps1` parses cleanly via the PowerShell AST parser.

### Artifacts Affected

- Agent configuration parser and the HTTPS client ABI (`hc_config_t` gains a field)
- Install-time configuration generators: DEB/RPM/source (`register_configure_agent.sh`,
  `inst-functions.sh`) and Windows (`InstallerScripts.vbs`)
- Source-upgrade configuration migration (`replace_manager_ip.sh`, `update.sh`)
- WPK upgrade pre-flight scripts (`pkg_installer.sh`, `do_upgrade.ps1`)
- Shipped default configurations (`etc/ossec-agent.conf`, `src/win32/ossec.conf`)
- Windows integration-test CI job

### Configuration Changes

**Breaking for the configuration file, not for installation.** `<address>`, `<port>` and
`<interface_index>` are no longer written — `<endpoint>` replaces them. Every installation
variable is preserved: `WAZUH_MANAGER`, `WAZUH_MANAGER_PORT`, `WAZUH_MANAGER_IP`, `ADDRESS`
and `SERVER_PORT` all still work, composing an `<endpoint>` when `WAZUH_MANAGER_ENDPOINT` is
not set.

A configuration left by an upgrade is still read: `<address>` with an optional `<port>`
resolves to the same target with the default prefix, and an INFO notice names the replacement.
Nothing is ever rewritten. Where both spellings appear, `<endpoint>` wins and the deprecated
pair is ignored with a warning. This deliberately leaves the issue's *"`<address>` and `<port>`
removed from `Read_Agent_Manager()`"* criterion unmet — meeting it would strand every agent an
upgrade left on the old spelling.

Two values that were previously accepted are now rejected, in both cases because the installers
already rejected them and the agent did not: a port outside 1-65535 (notably `:0`), and a
trailing `:` with no port.

### Documentation Updates

`docs/ref/getting-started/installation.md` — the `WAZUH_MANAGER_ENDPOINT` entry with the
grammar, defaults table and opt-out rule; IPv6 zone ids documented as supported, replacing text
that called them unsupported and pointed at the removed `<interface_index>`.

`docs/ref/modules/client/configuration.md` — `<endpoint>` documented as an option in its own
right, which it was not before; `<address>`, `<port>`, `max_retries` and `retry_interval`
marked deprecated to match what the code does; the stale `interface_index` entry removed.

`etc/templates/config/README.md` — the agent example brought in line with the shipped template.

`CHANGELOG.md` — the `WAZUH_MANAGER_ENDPOINT` entry corrected to say the installers write a
single `<endpoint>`.

### Tests Introduced

`src/unit_tests/config/test_client-config_https.c` — **86 checks**. Existing cases were
converted to the single-`<endpoint>` spelling; the new ones cover scheme present and absent, a
non-https scheme rejected, default port and prefix, the trailing-slash opt-out **distinguished
from no slash at all**, IPv6 with and without a zone id, an unknown zone rejected, query strings
and embedded credentials rejected, an out-of-range port, a trailing colon, a missing host, and
an empty `<endpoint>` rejected.

Precedence is pinned in **both element orders**, so a reintroduced order dependency fails the
suite rather than surviving as a latent bug. Two cases cover the legacy `<client><server>`
`<endpoint>`, with and without an explicit port and prefix.

Several tests assert the upgrade notice text exactly, because it is a promise an operator will
act on: that the suggested `<endpoint>` uses the **resolved** prefix rather than the default,
and that an IPv6 address is bracketed in the suggestion.

`src/unit_tests/client-agent/test_agentd.c` and `test_agent_config.c` — updated for the
`scope_id` rename and for the notice the config-report fixtures now trigger.

`src/init/tests/test_register_configure_agent.sh` — **39 checks**, covering every accepted and
rejected shape of the grammar plus the installation-variable rules: composition from
`WAZUH_MANAGER` with and without a port, IPv6 bracketing, the last of several comma-separated
addresses, the `WAZUH_MANAGER_IP` alias, endpoint precedence, and the two cases proving a
rejected or empty endpoint does not fall back.

10 integration-test files were updated to the whole-target spelling. The templates opt out of
the prefix (`127.0.0.1:1517/`) because `RemotedSimulator` serves unprefixed endpoints only; the
trailing slash is load-bearing there and is commented as such. In `test_enrollment`, the
placeholder had to stay a bare value — the framework substitutes whole values only, never
substrings — so each case supplies the complete endpoint, which is also what lets the IPv6 case
be bracketed. Its malformed-IPv6 case now asserts the config-parse rejection, since that value
is refused before any name resolution is attempted and can no longer reach a resolver error.

No C unit tests were added for the WPK pre-flight scripts, the source-upgrade migration or the
MSI, none of which have a harness in this repo; all three were exercised directly instead
(see Results and Evidence).

## Review Checklist

- [x] Code changes reviewed
- [ ] ...

Manual verification: real DEB, RPM and MSI installs with enrollment against a live manager;
the full port × prefix grid and both deprecated shapes exercised on a running agent; a hostname
target in place of an IP; the source-upgrade migration across five previous-config shapes; the
WPK pre-flight read across eight configuration shapes; and `ParseManagerEndpoint` driven
directly under `cscript`.

**Still open:**

- The `winagent` cmocka suite was not run: `src/external/googletest` holds a Linux build, so the
  mingw link fails on component tests that never compile this code — a pre-existing environment
  gap, not a regression here. The Windows build of the parser itself is verified end to end by
  the MSI enrollment runs above.
- The `manager` unit-test target and the macOS install path were not exercised.
- A real manager-driven WPK upgrade; the pre-flight read is verified at script level and the
  IPv6 probe fix by reproducing the malformed URL directly.
- Depends on #38658 (`406d75a799`, included in this branch), which is why the dependency box is
  unchecked.
- `RemotedSimulator` still has no reverse-proxy prefix support (separate repository), so the
  integration templates opt out of the prefix rather than exercising it.


src/client-agent/https_client/include/https_client.h
```diff
@@ -112,17 +112,22 @@ typedef enum hc_result_t
  */
 typedef struct hc_config_t
 {
-    char server_host[HC_MAX_HOST]; ///< Manager address (single server, IR2).
+    char server_host[HC_MAX_HOST]; ///< Manager address (single server, IR2). A bare,
```
jpcerrone: Reads as if this was ipv6 only. Maybe change comment to:
```
  char server_host[HC_MAX_HOST]; ///< Manager address (single server, IR2): IPv4, hostname,
  ///< or a bare, unbracketed IPv6 literal, which baseUrl() brackets for the wire.

```

src/client-agent/src/config.c
```diff
@@ -203,8 +203,10 @@ cJSON *getAgentConfig(void) {
             if (agt->server[i].endpoint)
                 cJSON_AddStringToObject(server, "endpoint", agt->server[i].endpoint);
 
-            if (agt->server[i].network_interface)
-                cJSON_AddNumberToObject(server, "interface_index", agt->server[i].network_interface);
+            /* #38624: reported as scope_id now -- the value comes from <endpoint>'s IPv6
```
jpcerrone: Lets just remove this comment in order not to reference the old implementation

src/config/include/client-config.h
```diff
@@ -20,11 +20,13 @@ typedef struct agent_flags_t {
...
 typedef struct agent_server {
-    char * rip;
-    int port;
-    char * endpoint; ///< <endpoint> (#38492): optional reverse-proxy path segment, normalized
-                     ///< (no leading/trailing '/'); NULL when unset -- today's unprefixed behavior.
-    uint32_t network_interface;
+    char * rip;      ///< Host parsed out of <endpoint> (#38624): IPv4, hostname, or a bare
+                     ///< (unbracketed) IPv6 literal, which ModuleConfig::baseUrl() re-brackets.
+    int port;        ///< Port from <endpoint>, or DEFAULT_HTTPS_REMOTE_PORT when omitted.
+    char * endpoint; ///< Reverse-proxy path prefix from <endpoint> (#38492/#38624), normalized
```
jpcerrone, line 26: should we call this `prefix` instead? endpoint now means the whole URL

src/config/src/client-config.c
```diff
@@ -457,23 +723,38 @@ int Read_Agent_Manager(XML_NODE node, agent * logr)
...
 
+    /* <endpoint>'s meaning depends on whether the deprecated <address> is present
+     * alongside it, and the two cannot be told apart from the value: "wazuh-manager"
+     * is both a legal prefix (5.0.0 spelling) and a legal hostname (the combined
+     * spelling). So decide first, in its own pass -- XML children arrive in document
+     * order, and <endpoint> may well come before <address>. */
+    bool legacy_address_present = false;
+
+    for (j = 0; node[j]; j++) {
+        if (node[j]->element && strcmp(node[j]->element, xml_agent_addr) == 0) {
```
jpcerrone: Maybe we shouldn't add logic to support the old `<endpoint>` meaning.

src/config/src/client-config.c
```diff
@@ -484,7 +765,10 @@ int Read_Agent_Manager(XML_NODE node, agent * logr)
             merror(XML_VALUENULL, node[j]->element);
             return (OS_INVALID);
         }
-        /* Get server address (IP or hostname) */
+        /* Deprecated: <address> and <port> were folded into <endpoint> (#38624). Still
+         * read, because a WPK upgrade never rewrites ossec.conf -- an agent installed
+         * by a 5.0.0 package restarts on the new binary with the old three tags, and
+         * refusing them would strand it. */
```
jpcerrone: Same as above

src/init/pkg_installer.sh
```diff
@@ -102,26 +276,51 @@ probe_server() {
     return $?
 }
 
-# The 5x agent reads the manager address from <agent><manager>, falling back to <client><server> when upgrading from 4x versions.
+# A WPK upgrade never rewrites ossec.conf, so this script meets three config shapes and
+# has to read all of them (#38624):
+#
+#   v2    <agent><manager><endpoint>  carrying host[:port][/prefix] in one value
```
jpcerrone: Probably remove the reference to pre-v2 5.0.0

jpcerrone (dismissed):

jpcerrone:
## Manual test: 4.x → 5.x WPK upgrade (Linux)

Tested on Ubuntu against a co-located 5.x manager serving HTTPS on `0.0.0.0:1517` with `global_prefix /wazuh-manager/`.

Installed a 4.14.6 agent keeping the legacy `<client><server><address>` shape, enrolled it and confirmed it connected over the legacy TCP protocol, then pushed a WPK built from this branch (its `pkg_installer.sh` plus the 5.0.0 agent package) with `agent_upgrade`.

The installer resolved the manager from the 4.x block, ignoring the legacy `<port>1514</port>` in favour of the 5.x HTTPS port and applying the default prefix:

```sql
2026/08/28 19:29:05 - Checking connectivity to 192.168.0.60:1517/wazuh-manager.
2026/08/28 19:29:05 - Manager reachable at 192.168.0.60:1517/wazuh-manager.
```

After the upgrade the agent resolved the same target from the untouched 4.x configuration, reported the single-value replacement for it, and started the HTTPS transport with the key it already held:

```sql
2026/08/28 19:30:49 wazuh-agentd: INFO: <agent><manager><address> is not configured. Using <client><server><address> '192.168.0.60' with the default port 1517 and the default endpoint prefix 'wazuh-manager'. Replace the <client><server> block with a single <endpoint>192.168.0.60:1517/wazuh-manager</endpoint>
2026/08/28 19:30:49 wazuh-agentd:https-client: INFO: Starting https_client (server=192.168.0.60:1517, agent=002, body compression enabled).
```

jpcerrone (approved):

cborla (approved):
LGTM

PR #38706, opened 2026-08-28
RPM package extraction from container image layers

### LINKED ISSUE #38586: Implement RPM package extraction from container images
## Description

Implement RPM package extraction for the `container_images` module, so RPM based images are inventoried like the Debian and Alpine images already supported.

RPM is handled apart because shared agent code has to change first: the package reading used for host packages resolves the host database only, and covers a single database format while current distributions use others. The change must be additive, leaving host package collection and every existing caller unchanged.

Parent research/spike: #35609

## Scope

- Extend the shared package reading so it can read a package database from a given location.
- Extend the shared package reading to the database formats used by current RPM based distributions.
- Keep both changes additive, leaving host package collection and every existing caller unchanged.
- Extract the package inventory of RPM based images, covering the database formats and locations used by the distributions in the compatibility matrix.
- Recognize the RPM database formats that are not supported and inventory the image with no packages and a warning.
- Report package versions as the distribution expresses them, including the version epoch where one is used.
- Populate the package fields already defined by the existing tables, with no schema change.
- Add the package format through the existing abstraction, leaving the image reading unchanged.
- Isolate a failing reference from the rest of the scan.
- Add unit tests for each supported database format.
- Add integration tests over RPM based image inputs.
- Add regression coverage for host package collection.
- Update the module documentation with the supported and unsupported database formats.

## Out of scope

- The RPM database formats that current distributions no longer use.
- Package formats other than the ones used by RPM based images.
- Language package managers such as pip, npm and gem.
- Changes to host package collection behaviour.
- Changes to the image reading and to the package format abstraction.

## Tasks

- [ ] Extend the shared package reading to accept the package database location.
- [ ] Extend the shared package reading to the database formats used by current distributions.
- [ ] Extract the package inventory of RPM based images for every supported database format.
- [ ] Cover the database locations used by the distributions in the compatibility matrix.
- [ ] Inventory an image with no packages and a warning when its database format is not supported.
- [ ] Report package versions including the version epoch where one is used.
- [ ] Map the extracted package fields onto the existing package table.
- [ ] Add the package format through the existing abstraction.
- [ ] Add unit tests for each supported database format.
- [ ] Add integration tests over RPM based image inputs.
- [ ] Add regression coverage for host package collection.

## Acceptance criteria

- Images from the RPM based distributions in the compatibility matrix are inventoried, whichever database format they use.
- The stored package list matches the one reported by the package manager inside the equivalent running container.
- Package versions carrying an epoch are stored with the epoch preserved.
- An image using an unsupported database format is inventoried with no packages and a warning, and the scan continues.
- An unreadable package database costs one reference, not the scan.
- Host package collection produces the same results as before the change, confirmed by the existing tests.
- The image reading and the package format abstraction are unchanged.
- No support for the unused database formats and no change to host package collection is introduced in this issue.

anromerom:
## Description

Closes #38586

RPM based images were inventoried with zero packages and a warning, because the package reading the agent already had resolved the host database only and covered a single database format. This PR extends the shared package reading to read a database from a given location and to the sqlite and ndb formats current distributions ship, then reads those databases out of the image layers through the package format abstraction the module already has. The Berkeley DB format rpm used before 4.16 stays recognized but not parsed. Host package collection is unchanged: `getRpmInfo` keeps its signature and its behaviour, and the location based entry point is a new overload.

## Proposed Changes

- Add a shared RPM header blob parser to the agent package reading, and rewire the Berkeley DB reader onto it so one implementation of the header layout serves every backend.
- Add a reader per database format current distributions use, sqlite and ndb, each able to read from a location or from a buffer, so an image is read from the layer bytes with no temporary file.
- Add `getRpmInfoFromLocation`, which reads an RPM database at a given path, leaving `getRpmInfo` and every existing caller untouched.
- Read the rpm database of an image at the four locations the distributions use, `var/lib/rpm/` and `usr/lib/sysimage/rpm/` for both formats, picking the format from the content rather than from the path.
- Map the extracted fields onto the existing package table with no schema change, reporting the version as the distribution expresses it and keeping the epoch where a package carries one.
- Identify the Berkeley DB format by its database file rather than by its directory, so an image still carrying it is inventoried with zero packages and a warning while the databases beside it stay parsable.
- Bound what a crafted database can ask for: every offset is computed in a fixed width type and checked by division and subtraction so it cannot wrap, a blob must record the package index of the slot pointing at it, and the payload handed out is capped at the size of the database.
- Open a database held in memory read only, normalizing the write-ahead log header bytes a real `rpmdb.sqlite` carries, so nothing is ever written back into the image bytes.

### Results and Evidence

Installed agent, ten images configured as `<archive>` references, one scan:

```sql
Scan ended. 10 references, 998 packages.
```

Every field of every RPM package matches `rpm -qa --qf` taken inside the equivalent running container: name, version with the epoch, architecture, vendor, category, size, source package and install time.

| Image | Database format | Packages | Against the package manager |
|---|---|---|---|
| `rockylinux:9` | sqlite | 141 | Every field matches |
| `fedora:39` | sqlite, under `/usr` | 142 | Every field matches |
| `almalinux:9` | sqlite | 158 | Every field matches |
| `amazonlinux:2023` | sqlite | 105 | Every field matches |
| `opensuse/leap:15.5` | ndb, under `/usr` | 130 | Every field matches |
| `opensuse/tumbleweed` | ndb | 118 | Every field matches |
| `amazonlinux:2` | Berkeley DB | 0 | Zero packages and one warning, as designed |
| `debian:12`, `ubuntu:22.04`, `alpine:3.19` | dpkg, apk | 88, 101, 15 | Unchanged by this change |

58 stored versions carry an epoch, for example `findutils=1:4.8.0-6.el9`.

An image whose format is not parsed is still inventoried, and the scan continues over the other references:

```sql
wazuh-modulesd:container_images: WARNING: NOT IMPLEMENTED: image at '/opt/corpus/amazonlinux_2.tar' uses the package format(s) rpm, which are recognized but not supported yet. Reporting zero packages.
```

Host package collection is unchanged. A differential harness ran a verbatim copy of the removed Berkeley DB row builder against the new shared parser over 200000 generated blobs: `mismatches 0 of 200000` on every tag type the old code handled, duplicates included. The one deliberate difference is that a tag stored as a string array is now read instead of being dropped; a census of every collected tag across 420 packages from three distribution families found only the types the old code already handled, so no real database reaches it.

Fuzzing with `-fsanitize=address,undefined`, seeded from real databases: 20000 mutated ndb databases, 3000 mutated sqlite databases and 300000 generated header blobs, with no sanitizer report. The header blob parser is also clean over 200000 iterations on a 32-bit build.

### Artifacts Affected

- `libsysinfo.so`, `libcontainer_images.so`, linked into `wazuh-modulesd`. All platforms; the new readers are header only, so no source file is added to any platform specific build list.
- No new link dependency. The sqlite API the reader uses is already exported by `libwazuhext.so`.

### Configuration Changes

N/A. No new setting, and no change to how `<container_images>` is configured.

### Documentation Updates

- `docs/ref/modules/container-images/README.md` and `configuration.md`: the supported package formats now list the two rpm database formats and their locations, state that the format is taken from the content of the database rather than from where it was found, and record that Berkeley DB is recognized but not parsed.

### Tests Introduced

- Unit, shared readers: a new `sysInfoRpmDbReaders` suite covering the header blob layout, both database formats, reading from a location and from memory, an absent epoch against an explicit zero, a database in write-ahead log mode, a file format newer than the reader knows, and the malformed and crafted cases found in review.
- Unit, module: rpm field mapping for both formats, the epoch preserved, the signing keys excluded, both database locations, a database that is not one, a truncated database, a blob that cannot be read, and the classification of the Berkeley DB file as recognized but not parsed.
- Integration: two scenarios over rpm image inputs, one per database format, asserting the packages are stored and the epoch is kept.
- Regression: the existing Berkeley DB and rpm parser suites are unchanged and passing, which is what pins host package collection.

Two pre-existing defects in the integration suite were fixed so it can run: the image fixtures wrote their images under a per-run temporary directory while the test cases named a fixed path, so no inventory case scanned the image its own fixture had built; and the module database was never reset between cases, so cases asserting on the whole packages table read rows an earlier case had stored. A third fix corrects the disabled-module log pattern, which expected an `INFO` line the module has never emitted.

## Review Checklist

- [ ] Code changes reviewed


cborla (approved):
LGTM

Issue #38678, opened 2026-08-27
Agent-side improvements to the JWT authentication mechanism


jpcerrone:
## Description

#38582 / #38602 replaced remoted's AES-CMAC request signature with the closed `wazuh-agent+jwt` bearer profile: the agent self-signs an HS256 token with its `client.keys` secret, the token binds the agent's identity and its freshness, and the manager accepts it while `now - iat <= jwt_max_age + jwt_clock_skew` (60 + 30 s by default).

That work is correct and complete for what it set out to do, but it deliberately left the agent with **no configuration and no lever of its own** over the new mechanism, and it narrowed the freshness window from the CMAC scheme's 300 s to 90 s. This issue collects the agent-side follow-ups that fall out of that, plus the questions that cannot be answered without the manager team.

Nothing here is broken in a default deployment. Every item is either a latent trap that only surfaces once an operator changes a timeout, or a missing capability the agent needs in order to behave sensibly when authentication does fail.

**Both lists below are ordered by importance, most important first.**

### 1. A prefix mismatch is invisible on both ends

When an agent's `<endpoint>` and the manager's `<remote><https><global_prefix>` disagree, the manager answers `404`. That is correct, and it is one of the deliberate improvements of #38582: the manager routes on the literal request target, the bearer token does not bind the target, so a routing mismatch can never surface as a `401`.

The problem is that nothing anywhere records it. On the agent, `classifyOutcome()` maps `404` to `Permanent` (`outcomeClassifier.cpp`): it is not retried, and because it is not an auth failure it never trips the re-enrollment latch. No error, no warning, and no status code reaches the log. At `agent.debug=2` the only trace is `DEBUG: Sending /control startup.` (`controlStream.cpp:274`) repeating every ten seconds, forever.

The manager is no better. `non_matched_request_handler` (`RestinioHttpServer.cpp:1167`) answers `404` with a JSON body and logs nothing, at any verbosity — at `remoted.debug=2` an unmatched route produces no output at all. The sole mention of the prefix in its entire log is the banner printed once at startup (`RestinioHttpServer.cpp:1383`), which incidentally still tells the operator that "agents must send and sign the full prefixed target" — wording left over from the CMAC scheme, and no longer true.

The agent never completes startup, never sends a keepalive, and ages out to disconnected. The result is an agent that is functionally dead while `wazuh-control status` reports every daemon running, with maximum verbosity enabled on both ends and zero evidence on either.

[...]

Issue #38582, opened 2026-08-25
Replace remoted's custom AES-CMAC request signature with a standard JWT (HS256) agent authentication profile


juliancnn:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0 (branch `5.0.0`) | remoted HTTPS agent API (`remoted_module`), agent `https_client`, `manager_benchmark` tooling | Manager / Agent | Sources | Linux, Windows (agent) |

## Description

Agents authenticate every request to remoted's HTTPS API (`/stateless`, `/stateful`, `/control`,
`/stats`, `/config`, `/download`, `/scan/vd`) with a Wazuh-specific scheme, "protocol version 1":

```http
protocol-version: 1
Authorization: Wazuh <agent-id>:<timestamp>:<mac>
```

where `<mac>` is an AES-CMAC computed over a proprietary canonical string
(`WAZUH-REQUEST\n<version>\n<METHOD>\n<target>\n<agent-id>\n<timestamp>\n<body>`) with the
agent's `client.keys` secret. The primitive (AES-CMAC, NIST SP 800-38B) is standard, but the
framing, the header syntax, the canonicalization rules and the key handling are our own. Every
client that talks to the manager (the agent, the load simulator, the manual tools) must carry a
hand-written implementation of that framing, and there is no third-party tooling or prior security
analysis we can lean on.

This issue proposes to **retire protocol 1's signature scheme and authenticate agents with a
standard, closed JWT profile**:

- JWT (RFC 7519) in JWS Compact Serialization (RFC 7515), algorithm HS256 (RFC 7518), following
  the JWT Best Current Practices (RFC 8725) and the HTTP Bearer scheme (RFC 6750).
- The token is **issued by the agent itself** and signed with its individual pre-shared secret from
  `client.keys` (the same credential used today). The manager verifies it with its copy of the
  secret. No token issuance service, no asymmetric keys, no new credential distribution.
- Enrollment (`/enroll`), which today uses a sibling CMAC scheme (`Authorization: WazuhEnroll`),
  moves to a second closed JWT profile keyed from the enrollment password, so that **no CMAC code
  remains in the repository**.
- The transport is trusted: TLS provides confidentiality and integrity of the request. The JWT
  authenticates the **agent's identity only**. Unlike today's CMAC, the token deliberately does
  **not** bind the HTTP method, path or body — this is an accepted trade-off in exchange for a
[...]

PR #38726, opened 2026-08-31
Agent retries its existing credential before re-enrolling on a 401

### LINKED ISSUE #38610: Agent discards a valid credential after a single 401
## Description

On a `401`, the agent immediately abandons the credential it holds and switches to
re-enrollment. **It never retries it once.**

If the 401 was transient — the key is valid, remoted just could not see it in
`client.keys` yet — the agent does not recover when the condition clears, because nothing
ever tries the existing credential again.

Wazuh 4.x did not behave this way: the connection loop retried the existing key
`max_retries` times (default 5, every 10 s) before it would consider enrolling. That loop
was removed with the HTTPS transport migration.

```mermaid
sequenceDiagram
    participant A as Agent
    participant R as remoted
    participant CK as client.keys
    participant AU as authd

    A->>R: signed request
    R->>CK: is 001 there?
    CK-->>R: yes
    R-->>A: 202 Accepted

    Note over CK: 001's entry is removed

    A->>R: signed request (THE SAME key)
    R->>CK: is 001 there?
    CK-->>R: no
    R-->>A: 401
    Note over A: discards the credential<br/>and switches to re-enrolling

    Note over CK: 001's entry is restored<br/>THE CREDENTIAL IS VALID AGAIN

    loop every 60 s, indefinitely
        A->>AU: enroll as 'ubuntu24'
        AU-->>A: 409 duplicate name
    end

    Note over A,R: never tries its credential again,<br/>which has been working for an hour
```

## Evidence

A 5.0.0 manager with an agent `ubuntu24` (id `001`) enrolled and active. The only thing
modified was the manager's `client.keys`; the agent's key was never touched.

```
15:46:57  request signed with 001's key                       --> 202 Accepted

15:47:29  agent 001's entry removed from the manager's client.keys
          wazuh-manager-remoted:keystore: INFO: Reloaded 0 agent key(s) from 'etc/client.keys'.

15:47:34  the agent gets the 401 and discards its credential (5 s later)
          wazuh-manager-authd: WARNING: Duplicate name 'ubuntu24', rejecting enrollment.
                               Agent '001' can't be replaced since it is not disconnected.

15:48:29  client.keys restored -- THE AGENT'S CREDENTIAL IS VALID AGAIN
          wazuh-manager-remoted:keystore: INFO: Reloaded 1 agent key(s) from 'etc/client.keys'.

16:51:11  the agent is still re-enrolling. 70 rejections, 64 minutes, and counting.
```

The credential started working again at 15:48:29 and the agent never tried it. It took 5
seconds to abandon it and over an hour without looking at it again.

On the agent side, from a real run (Jenkins `5.0.0/deployability_testing/1532`, agent
`067`), the jump to enrollment happens in the same second as the 401, with no intermediate
attempt:

```
07:38:36 wazuh-agentd: WARNING: https_client: credential rejected (401); re-enrolling.
07:38:36 wazuh-agentd: WARNING: Manager unreachable. Pausing module event production.
07:38:36 wazuh-agentd: INFO: Using password specified on file: etc/authd.pass
```

## Steps to reproduce

Single-node manager and one active agent. No cluster required.

1. On the manager, back up `etc/client.keys` and delete the agent's line.
   remoted hot-reloads it (`Reloaded N agent key(s)`).
2. Wait for the agent's next request: it gets a `401`.
3. **Restore `etc/client.keys`.** The agent's credential is valid again.

**Expected:** the agent retries and reconnects.

**Actual:** it does not recover. It re-enrolls indefinitely.

With `<enrollment><enabled>no</enabled>` the case is even more isolated: the agent pauses
and never returns, logging

```
https_client: re-enrollment required but auto-enrollment is disabled
(<enrollment><enabled>); traffic stays paused until the key is fixed manually.
```

The message is misleading: fixing the key manually does **not** resume traffic, because
nothing re-reads it.

## Proposed change

On a 401, retry the credential the agent already holds a bounded number of times before
falling back to enrollment.

In `bridge_on_reenroll_required()`: keep a per-incident attempt counter; for attempts
1..N, back off and call `hc_set_agent_identity()` again with the **same** id and key
(which releases `AuthGate` and lets the endpoints retry); only after N failures start
`bridge_reenroll_thread()`. Reset the counter on `HC_STATE_REGISTERED`.

The counter must live in the bridge rather than in `AuthGate`: `reportAuthFailure()` fires
the callback again every time the gate is re-armed, so a counter inside the gate would
loop.

4.x's defaults (5 attempts, 10 s apart) are a reasonable starting point.

## Note on enrollment

Re-enrollment is not a free fallback: when the agent enrolls under a name that already
exists, authd evaluates whether it may replace the existing agent and by default refuses —
as seen in the evidence above, where the agent being rejected is the agent itself. A
transient 401 can therefore turn into a much longer outage than the condition that caused
it.

That side of it is worth handling separately. This issue is only about the agent not
retrying what it already has.

Darioortegaleyva:
## Description

This PR makes the HTTPS agent retry the credential it already holds a bounded number of times before falling back to re-enrollment on a `401`.

With the `wazuh-agent+jwt` profile (#38582 / #38602), the manager collapses every credential check into a single, deliberately generic `401` (#37828: a distinct status would confirm an agent id exists). The agent treated that `401` as a dead credential: `RetrySender` grants one clock-skew grace retry and a second `401` trips `AuthGate` -> re-enrollment. A transient `401` (a `client.keys` entry the manager has not reloaded yet, clock drift beyond one correction, a slow transfer aging the token) therefore made the agent discard a working credential and re-enroll, churning its identity on the manager to fix something re-enrollment does not fix. With `<enrollment><enabled>no</enabled>` it paused forever and nothing re-presented the key.

This is item 2 of #38678 ("Telling a stale token apart from a dead key"). 4.x's connection loop retried the existing key before enrolling; that loop was lost in the HTTPS transport migration. This restores a bounded version of it.

**Proposed Release:** [v5.0.0]
**Issue:** wazuh/wazuh#38610
**Internal Reference:** N/A

## Proposed Changes

- `src/client-agent/src/https_client_bridge.c`:
  - `bridge_on_reenroll_required()`: keep a per-incident back-off delay. On each `401`, grow the delay and spawn a thread that re-presents the held credential; only once the delay reaches the cap spawn re-enrollment. The retry runs regardless of `<enrollment><enabled>`, since re-presenting the key the agent already has is not enrollment.
  - New `bridge_cred_retry_thread()`: after a back-off, calls `hc_set_agent_identity()` with the **same** id/key, which releases `AuthGate` and lets the endpoints retry with a fresh token (re-minting the JWT and re-running the clock-skew correction). A transient `401` clears here without churning identity.
  - `bridge_on_state_change()`: reset the back-off on `HC_STATE_REGISTERED` (a clean registration ends the incident).
  - **Incremental back-off**, same `delta`/`max` shape as the existing `enrollment_retry_*` ramp: the retry delay grows by `auth_retry_delta` after each rejection, up to `auth_retry_max`, and only once it reaches the cap does it re-enroll. A widening window (rather than a fixed one) covers a clock still stabilising after an NTP step / VM resume, and a cluster's `client.keys` sync window. Defaults `10 s` delta up to a `30 s` cap (3 retries: 10, 20, 30 s).
- `etc/internal_options.conf`: document the two new agent options next to the existing `enrollment_retry_*` ramp.

### Why the counter lives in the bridge

`AuthGate::reportAuthFailure()` fires `on_reenroll_required` once per incident and re-arms when `hc_set_agent_identity()` is called; a counter inside the gate would loop. The bridge counts incidents instead. Both `on_reenroll_required` and `on_state_change` run on the module's single callback worker (`CallbackDispatcher`), so the counter needs no lock.

### Agent behavior per response code (context)

Only the `401` abandoned the credential; every other transient outcome is already retried without touching it. This PR aligns the `401` with that policy.

| HTTP | OutcomeClass | Retries held credential? | Action (before this PR) |
|---|---|---|---|
| 2xx | Ok | - | success, reset backoff |
| **401** | **AuthFail** | one-shot only | one clock-skew retry, then re-enroll |
| 400 (and other 4xx) | Permanent | no | give up (drop the batch) |
| 403 / 426 | ServerError | yes, backoff | retried with backoff |
| 409 | VersionRejected | no, keeps identity | REJECTED state, slow startup retry |
| 413 | PayloadTooLarge | no | `/stateless` splits and resends smaller |
| 415 | CompressionRejected | one-shot | one uncompressed retry |
| 429 / 503 | BackPressure | yes, backoff | retried; server `Retry-After` wins |
| 5xx, transport failure | ServerError / Unreachable | yes, backoff | retried with backoff |

After this PR the `401` row becomes: retry the held credential with an incremental back-off (delta up to a cap), then re-enroll once the delay reaches the cap.

### Idempotency note

Retrying on a `401` is safe: a `401` means the manager rejected the request and never processed it (unlike the 429/503 retries discussed in item 4 of #38678, where the request may already have run). So this adds no duplication risk.

### Results and Evidence

The bridge callback (`bridge_on_reenroll_required`) and the retry thread are exercised directly, with strict cmocka assertions on the exact log line and behavior at each step, so a passing run means the real bridge produced exactly that sequence.

**1. The defect, reproduced.** On the unmodified branch, a test asserting the intended behavior ("on the first 401, retry the held credential") fails — the agent never calls `hc_set_agent_identity()` and goes straight to re-enrollment:

```
[ RUN      ] test_reenroll_retries_held_credential_before_enrolling
[  ERROR   ] --- __wrap_hc_set_agent_identity: parameters still have values that haven't been checked.
[  FAILED  ] test_reenroll_retries_held_credential_before_enrolling
 1 FAILED TEST(S)
```

**2. The fix, verified.** With the change, the four behavioral tests each pin one part of the contract, and the full suite is green:

```
[ RUN      ] test_reenroll_callback_first_401_retries_held_credential     [  OK  ]
[ RUN      ] test_cred_retry_thread_represents_held_identity              [  OK  ]
[ RUN      ] test_reenroll_callback_reenrolls_after_retries_exhausted     [  OK  ]
[ RUN      ] test_reenroll_callback_disabled_retries_then_logs_error      [  OK  ]
[==========] tests: 92 test(s) run.
[  PASSED  ] 92 test(s).
```

- `first_401_retries_held_credential` — the first `401` logs `retrying held credential in 10 s` and does **not** re-enroll.
- `cred_retry_thread_represents_held_identity` — the retry re-presents the **same** id/key via `hc_set_agent_identity()`; this is the recovery path: a transient `401` (a `client.keys` entry not yet propagated, a clock still settling) clears here without churning identity.
- `reenrolls_after_retries_exhausted` — once the back-off ramps to its cap (10 -> 20 -> 30 s) it re-enrolls a genuinely dead credential.
- `disabled_retries_then_logs_error` — even with `<enrollment><enabled>no</enabled>` the held credential is still retried first.

**3. Log sequence produced by the built agent** (strings from `wazuh-agentd`):

```
credential rejected (401); retrying held credential in 10 s.
credential rejected (401); retrying held credential in 20 s.
credential rejected (401); retrying held credential in 30 s.
credential rejected (401); re-enrolling.            # only after the ramp is exhausted
```

**4. Integration.** The `agentd` reconnection integration suites (Linux and Windows, tier-0-1) pass on CI with these changes.

### Artifacts Affected

- Linux agent (`wazuh-agentd` + `libhttps_client.so`) and Windows agent (`libhttps_client.dll`, same sources): the `401` re-enrollment decision in the bridge.

### Configuration Changes

- New agent internal options in `etc/internal_options.conf`, mirroring `enrollment_retry_delta` / `enrollment_retry_max`: `agent.auth_retry_delta` (back-off increment in seconds, default `10`, `[1..3600]`) and `agent.auth_retry_max` (back-off cap in seconds before re-enrolling, default `30`, `[1..86400]`).

### Documentation Updates

N/A

### Tests Introduced

- **Scope:** Unit tests (cmocka, `test_https_client_bridge`).
- **Details:** the four tests above; the existing re-enrollment-thread tests remain green.

## Review Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer

**General Checklist:**

- [ ] Code changes reviewed

### DIFF
--- a/etc/internal_options.conf
+++ b/etc/internal_options.conf
@@ -193,6 +193,13 @@ agent.enrollment_retry_delta=5
 # Cap in seconds [1..86400]
 agent.enrollment_retry_max=60
 
+# https_client retry of the held credential on a 401 before re-enrolling: the
+# delay grows by <delta> after each rejection, up to <max>, then it re-enrolls.
+# Increment in seconds [1..3600]
+agent.auth_retry_delta=10
+# Cap in seconds [1..86400]
+agent.auth_retry_max=30
+
 # TCP keepalive options for agent->manager connections
 # Time (in seconds) the connection needs to remain idle before TCP starts sending keepalive probes [1..7200]
 agent.tcp_keepidle=60
--- a/src/client-agent/src/https_client_bridge.c
+++ b/src/client-agent/src/https_client_bridge.c
@@ -204,6 +204,49 @@ static DWORD WINAPI bridge_reenroll_thread_win(LPVOID arg)
 }
 #endif
 
+#define BRIDGE_CRED_RETRY_DELTA_S_DEFAULT 10
+#define BRIDGE_CRED_RETRY_MAX_S_DEFAULT 30
+
+/* Resolved once in bridge_build_config(): an out-of-range value must refuse start, not
+ * abort the agent in the dispatcher thread on its first 401. */
+int g_cred_retry_delta = BRIDGE_CRED_RETRY_DELTA_S_DEFAULT;
+int g_cred_retry_max = BRIDGE_CRED_RETRY_MAX_S_DEFAULT;
+
+/* Ramped delay (s). Confined to the single CallbackDispatcher worker, so no lock. */
+int g_cred_retry_delay = 0;
+
+/* Re-presenting the held id/key releases AuthGate, so the endpoints retry a transient
+ * 401 (fresh token, re-run skew correction) instead of re-enrolling. */
+void *bridge_cred_retry_thread(void *arg)
+{
+    hc_handle *handle = (hc_handle *)arg;
+    sleep((unsigned int)g_cred_retry_delay);
+
+    if (bridge_stopping()) {
+        mdebug1("https_client: agent shutting down; abandoning credential retry.");
+        return NULL;
+    }
+
+    w_mutex_lock(&g_https_client_lock);
+    if (g_https_client_stopping) {
+        w_mutex_unlock(&g_https_client_lock);
+        return NULL;
+    }
+    if (!hc_set_agent_identity(handle, keys.keyentries[0]->id, keys.keyentries[0]->raw_key)) {
+        merror("https_client: failed to re-present the signing identity; traffic stays paused.");
+    }
+    w_mutex_unlock(&g_https_client_lock);
+    return NULL;
+}
+
+#ifdef WIN32
+static DWORD WINAPI bridge_cred_retry_thread_win(LPVOID arg)
+{
+    bridge_cred_retry_thread(arg);
+    return 0;
+}
+#endif
+
 /* Startup-response parsing: applies module limits and cluster-name authority
  * from the handshake. Deliberately separate, local copies of
  * client-agent/src/start_agent.c's parse_fim_limits()/
@@ -414,7 +457,21 @@ static void bridge_on_startup_result(bool accepted, const char *metadata_json, v
 static void bridge_on_reenroll_required(void *user_data)
 {
     (void)user_data;
-    mwarn("https_client: credential rejected (401); re-enrolling.");
+
+    if (g_cred_retry_delay < g_cred_retry_max) {
+        g_cred_retry_delay += g_cred_retry_delta;
+        if (g_cred_retry_delay > g_cred_retry_max) {
+            g_cred_retry_delay = g_cred_retry_max;
+        }
+        mwarn("https_client: credential rejected (401); retrying held credential in %d s.",
+              g_cred_retry_delay);
+#ifdef WIN32
+        CloseHandle(w_create_thread(NULL, 0, bridge_cred_retry_thread_win, g_https_client, 0, NULL));
+#else
+        w_create_thread(bridge_cred_retry_thread, g_https_client);
+#endif
+        return;
+    }
 
     if (!agt->enrollment.enabled) {
         merror("https_client: re-enrollment required but auto-enrollment is disabled "
@@ -422,8 +479,9 @@ static void bridge_on_reenroll_required(void *user_data)
         return;
     }
 
+    mwarn("https_client: credential rejected (401); re-enrolling.");
 #ifdef WIN32
-    w_create_thread(NULL, 0, bridge_reenroll_thread_win, g_https_client, 0, NULL);
+    CloseHandle(w_create_thread(NULL, 0, bridge_reenroll_thread_win, g_https_client, 0, NULL));
 #else
     w_create_thread(bridge_reenroll_thread, g_https_client);
 #endif
@@ -626,7 +684,7 @@ static void bridge_dispatch_control_task(const char *task_id, bool restart)
     ctx->restart = restart;
 
 #ifdef WIN32
-    w_create_thread(NULL, 0, bridge_control_task_thread_win, ctx, 0, NULL);
+    CloseHandle(w_create_thread(NULL, 0, bridge_control_task_thread_win, ctx, 0, NULL));
 #else
     w_create_thread(bridge_control_task_thread, ctx);
 #endif
@@ -968,7 +1026,7 @@ static void bridge_on_remote_upgrade_ready(const char *task_id, const char *wpk_
     os_strdup(installer, ctx->installer);
 
 #ifdef WIN32
-    w_create_thread(NULL, 0, bridge_upgrade_thread_win, ctx, 0, NULL);
+    CloseHandle(w_create_thread(NULL, 0, bridge_upgrade_thread_win, ctx, 0, NULL));
 #else
     w_create_thread(bridge_upgrade_thread, ctx);
 #endif
@@ -1317,6 +1375,7 @@ static void bridge_on_state_change(int state, void *user_data)
      * the latter, since it only fires on a paused->running transition and each
      * run starts believing it has paused nothing. os_delwait() is idempotent. */
     if (state == HC_STATE_REGISTERED) {
+        g_cred_retry_delay = 0;
         os_delwait();
     }
 }
@@ -1774,6 +1833,13 @@ static bool bridge_build_config(hc_config_t *config)
     config->download_max_attempts = (uint32_t)getDefine_Int_default("agent", "https_download_attempts", 1, 64, 2);
     config->producer_pause_threshold =
         (uint32_t)getDefine_Int_default("agent", "https_producer_pause_threshold", 1, 1000, 2);
+
+    /* Held-credential 401 retry ramp (#38610). Resolved here, not in the 401 callback:
+     * out-of-range must refuse start, not abort the agent in the dispatcher thread. */
+    g_cred_retry_delta = getDefine_Int_default("agent", "auth_retry_delta", 1, 3600,
+                                               BRIDGE_CRED_RETRY_DELTA_S_DEFAULT);
+    g_cred_retry_max = getDefine_Int_default("agent", "auth_retry_max", 1, 86400,
+                                             BRIDGE_CRED_RETRY_MAX_S_DEFAULT);
 
     /* Occupancy ladder: the same internal options the legacy client buffer
      * read, so tuned thresholds keep working. */
--- a/src/unit_tests/client-agent/test_https_client_bridge.c
+++ b/src/unit_tests/client-agent/test_https_client_bridge.c
@@ -240,7 +240,9 @@ int __wrap_getDefine_Int_default(const char *high_name, const char *low_name, in
  * it every test, since w_https_client_start() (the normal reset point) isn't
  * called by the tests that invoke bridge_reenroll_thread directly. */
 extern void *bridge_reenroll_thread(void *arg);
+extern void *bridge_cred_retry_thread(void *arg);
 extern bool g_https_client_stopping;
+extern int g_cred_retry_delay;
 
 /* bridge_control_task_thread/bridge_upgrade_thread are
  * likewise non-static so tests can call them directly, bypassing
@@ -358,6 +360,7 @@ static int setup_test(void **state)
     memset(&keys, 0, sizeof(keys));
     g_captured_config_valid = false;
     g_https_client_stopping = false;
+    g_cred_retry_delay = 0;
     g_populate_metadata_calls = 0;
     g_resolved_options[0] = '\0';
 
@@ -533,6 +536,30 @@ static void test_bridge_does_not_resolve_the_enrollment_retry_ramp(void **state)
 
     assert_null(strstr(g_resolved_options, "enrollment_retry_max"));
     assert_null(strstr(g_resolved_options, "enrollment_retry_delta"));
+
+    w_https_client_stop();
+}
+
+/* #38610: the auth-retry ramp is resolved once at startup, not in the 401 callback, so an
+ * out-of-range value refuses the start instead of aborting the agent on its first 401. */
+static void test_bridge_resolves_the_auth_retry_ramp_at_startup(void **state)
+{
+    (void)state;
+
+    expect_string(__wrap__minfo, formatted_msg, "https_client: starting.");
+    expect_string(__wrap_OS_SHA256_File, fname, SHAREDCFG_FILE);
+    expect_value(__wrap_OS_SHA256_File, mode, OS_BINARY);
+    will_return(__wrap_OS_SHA256_File, NULL);
+    expect_any(__wrap_hc_create, callbacks);
+    will_return(__wrap_hc_create, FAKE_HANDLE);
+    expect_value(__wrap_hc_start, handle, FAKE_HANDLE);
+    will_return(__wrap_hc_start, true);
+    expect_value(__wrap_hc_destroy, handle, FAKE_HANDLE);
+
+    w_https_client_start();
+
+    assert_non_null(strstr(g_resolved_options, "auth_retry_delta:1:3600;"));
+    assert_non_null(strstr(g_resolved_options, "auth_retry_max:1:86400;"));
 
     w_https_client_stop();
 }
@@ -922,38 +949,83 @@ static void start_client_successfully(void)
     w_https_client_start();
 }
 
-static void test_reenroll_callback_disabled_enrollment_logs_error_only(void **state)
+/* #38610: with auto-enrollment disabled the held credential is still retried
+ * (the whole ramp) before the "paused, fix manually" error. */
+static void test_reenroll_callback_disabled_retries_then_logs_error(void **state)
 {
     (void)state;
     start_client_successfully();
 
-    expect_string(__wrap__mwarn, formatted_msg, "https_client: credential rejected (401); re-enrolling.");
+    for (int delay = 10; delay <= 30; delay += 10) {
+        char msg[128];
+        snprintf(msg, sizeof(msg),
+                 "https_client: credential rejected (401); retrying held credential in %d s.", delay);
+        expect_string(__wrap__mwarn, formatted_msg, msg);
+        g_captured_callbacks.on_reenroll_required(g_captured_callbacks.user_data);
+    }
+
     expect_string(__wrap__merror, formatted_msg,
                   "https_client: re-enrollment required but auto-enrollment is disabled "
                   "(<enrollment><enabled>); traffic stays paused until the key is fixed manually.");
-
     g_captured_callbacks.on_reenroll_required(g_captured_callbacks.user_data);
-    /* No CreateThread expectation needed: the shared __wrap_CreateThread
-     * always succeeds without invoking its argument, and this path must not
-     * even reach the call. */
 
     expect_value(__wrap_hc_destroy, handle, FAKE_HANDLE);
     w_https_client_stop();
 }
 
-static void test_reenroll_callback_enabled_enrollment_only_warns(void **state)
+/* #38610: the first 401 retries the held credential instead of re-enrolling. */
+static void test_reenroll_callback_first_401_retries_held_credential(void **state)
 {
     (void)state;
     enable_enrollment();
     start_client_successfully();
 
-    expect_string(__wrap__mwarn, formatted_msg, "https_client: credential rejected (401); re-enrolling.");
-    /* No __wrap__merror expectation: the disabled-path error must not fire. */
+    expect_string(__wrap__mwarn, formatted_msg,
+                  "https_client: credential rejected (401); retrying held credential in 10 s.");
 
     g_captured_callbacks.on_reenroll_required(g_captured_callbacks.user_data);
 
     expect_value(__wrap_hc_destroy, handle, FAKE_HANDLE);
     w_https_client_stop();
+}
+
+/* #38610: only after the back-off ramps to its cap does it re-enroll. */
+static void test_reenroll_callback_reenrolls_after_retries_exhausted(void **state)
+{
+    (void)state;
+    enable_enrollment();
+    start_client_successfully();
+
+    for (int delay = 10; delay <= 30; delay += 10) {
+        char msg[128];
+        snprintf(msg, sizeof(msg),
+                 "https_client: credential rejected (401); retrying held credential in %d s.", delay);
+        expect_string(__wrap__mwarn, formatted_msg, msg);
+        g_captured_callbacks.on_reenroll_required(g_captured_callbacks.user_data);
+    }
+
+    expect_string(__wrap__mwarn, formatted_msg,
+                  "https_client: credential rejected (401); re-enrolling.");
+    g_captured_callbacks.on_reenroll_required(g_captured_callbacks.user_data);
+
+    expect_value(__wrap_hc_destroy, handle, FAKE_HANDLE);
+    w_https_client_stop();
+}
+
+/* #38610: the retry thread sleeps the ramped delay, then re-presents the held
+ * id/key. Called directly to bypass w_create_thread. */
+static void test_cred_retry_thread_represents_held_identity(void **state)
+{
+    (void)state;
+
+    g_cred_retry_delay = 20;
+    expect_value(__wrap_sleep, seconds, 20);
+    expect_value(__wrap_hc_set_agent_identity, handle, FAKE_HANDLE);
+    expect_string(__wrap_hc_set_agent_identity, agent_id, keys.keyentries[0]->id);
+    expect_string(__wrap_hc_set_agent_identity, key_hex, keys.keyentries[0]->raw_key);
+    will_return(__wrap_hc_set_agent_identity, true);
+
+    bridge_cred_retry_thread(FAKE_HANDLE);
 }
 
 /* bridge_reenroll_thread's retry-loop logic, called directly (synchronously)
@@ -2724,6 +2796,8 @@ int main(void)
         cmocka_unit_test_setup_teardown(test_compression_defaults_to_enabled, setup_test, teardown_test),
         cmocka_unit_test_setup_teardown(test_bridge_does_not_resolve_the_enrollment_retry_ramp, setup_test,
                                         teardown_test),
+        cmocka_unit_test_setup_teardown(test_bridge_resolves_the_auth_retry_ramp_at_startup, setup_test,
+                                        teardown_test),
         cmocka_unit_test_setup_teardown(test_client_cert_key_and_ciphers_are_copied, setup_test, teardown_test),
         cmocka_unit_test_setup_teardown(test_configured_endpoint_reaches_the_module, setup_test, teardown_test),
         cmocka_unit_test_setup_teardown(test_absent_endpoint_leaves_the_field_empty, setup_test, teardown_test),
@@ -2741,8 +2815,10 @@ int main(void)
         cmocka_unit_test_setup_teardown(test_valid_64_char_key_is_accepted, setup_test, teardown_test),
         cmocka_unit_test_setup_teardown(test_hc_create_failure_is_logged, setup_test, teardown_test),
         cmocka_unit_test_setup_teardown(test_hc_start_failure_destroys_and_logs, setup_test, teardown_test),
-        cmocka_unit_test_setup_teardown(test_reenroll_callback_disabled_enrollment_logs_error_only, setup_test, teardown_test),
-        cmocka_unit_test_setup_teardown(test_reenroll_callback_enabled_enrollment_only_warns, setup_test, teardown_test),
+        cmocka_unit_test_setup_teardown(test_reenroll_callback_disabled_retries_then_logs_error, setup_test, teardown_test),
+        cmocka_unit_test_setup_teardown(test_reenroll_callback_first_401_retries_held_credential, setup_test, teardown_test),
+        cmocka_unit_test_setup_teardown(test_reenroll_callback_reenrolls_after_retries_exhausted, setup_test, teardown_test),
+        cmocka_unit_test_setup_teardown(test_cred_retry_thread_represents_held_identity, setup_test, teardown_test),
         cmocka_unit_test_setup_teardown(test_reenroll_thread_succeeds_on_first_attempt, setup_test, teardown_test),
         cmocka_unit_test_setup_teardown(test_reenroll_thread_retries_with_backoff_then_succeeds, setup_test, teardown_test),
         cmocka_unit_test_setup_teardown(test_reenroll_thread_backoff_follows_the_resolved_ramp, setup_test,
--- a/tests/integration/test_enrollment/test_agent/test_agentd_enrollment.py
+++ b/tests/integration/test_enrollment/test_agent/test_agentd_enrollment.py
@@ -67,12 +67,17 @@ test_configuration = load_configuration_template(config_path, config_parameters,
 # Test variables.
 socket_listener = None
 
+# Collapse the held-credential retry ramp so a rejected pre-existent key re-enrolls
+# within the poll deadline instead of after the production back-off (#38610).
+local_internal_options = {'agent.auth_retry_delta': '1', 'agent.auth_retry_max': '1'}
+
 daemons_handler_configuration = {'all_daemons': True}
 
 
 # Test function.
 @pytest.mark.parametrize('test_configuration, test_metadata',  zip(test_configuration, test_metadata), ids=cases_ids)
-def test_agentd_enrollment(test_configuration, test_metadata, set_wazuh_configuration, daemons_handler_module, shutdown_agentd,
+def test_agentd_enrollment(test_configuration, test_metadata, set_wazuh_configuration, configure_local_internal_options,
+                           daemons_handler_module, shutdown_agentd,
                            set_keys, set_password, configure_socket_listener, restart_agentd):
     """
     description:


Darioortegaleyva:
### CI status

The `agentd` integration suites (IT Linux + IT Windows, tier-0-1) pass with this change.

The failing **enrollment** checks (IT Linux / IT Windows - enrollment) are **unrelated to this PR** — they are a pre-existing integration-test breakage introduced by the `<endpoint>` reverse-proxy feature (#38614): an absent `<endpoint>` now defaults to the manager's `wazuh-manager` prefix, so the agent enrolls at the prefixed path while the test's `RemotedSimulator`/`AuthdSimulator` only serve at the bare root, and the request is never captured ("No /enroll request captured"). The `test_enrollment` config is byte-identical to the base branch here; this PR does not touch enrollment.

This is already tracked and being fixed in qa-integration-framework (add reverse-proxy prefix support to the simulators):

- Issue: https://github.com/wazuh/qa-integration-framework/issues/835
- PR: https://github.com/wazuh/qa-integration-framework/pull/840

Those enrollment checks will go green once that framework fix lands and CI picks it up.

src/client-agent/src/https_client_bridge.c
```diff
@@ -423,14 +469,37 @@ static void bridge_on_startup_result(bool accepted, const char *metadata_json, v
...
+        if (g_cred_retry_delay > retry_max) {
+            g_cred_retry_delay = retry_max;
+        }
+        mwarn("https_client: credential rejected (401); retrying held credential in %d s.",
+              g_cred_retry_delay);
+#ifdef WIN32
+        w_create_thread(NULL, 0, bridge_cred_retry_thread_win, g_https_client, 0, NULL);
+#else
+        w_create_thread(bridge_cred_retry_thread, g_https_client);
+#endif
```
MarcelKemp, line 481: **[S2 · blocking · introduced]** The `HANDLE` that `w_create_thread` returns on WIN32 is discarded, so every ramp step leaks a kernel thread object (`pthreads_op.h:17` yields `hd` from the statement expression, and there is no `CloseHandle` anywhere in this file). Before this PR a 401 incident leaked one handle; now it leaks `ceil(auth_retry_max/auth_retry_delta)` more, and that count is operator-tunable — `auth_retry_delta=1` with `auth_retry_max=86400` leaks 86400 in a single sustained incident on an agent that runs for months.

```suggestion
#ifdef WIN32
        HANDLE cred_retry_handle = w_create_thread(NULL, 0, bridge_cred_retry_thread_win, g_https_client, 0, NULL);
        CloseHandle(cred_retry_handle); /* Fire-and-forget: don't keep the thread object alive. */
#else
        w_create_thread(bridge_cred_retry_thread, g_https_client);
#endif
```

The four sibling spawn sites in this file (lines 504, 707, 981) have the same bug and are pre-existing — worth fixing together or in a follow-up, but this one is new.

etc/internal_options.conf
```diff
@@ -192,6 +192,13 @@ agent.enrollment_retry_delta=5
 # Cap in seconds [1..86400]
 agent.enrollment_retry_max=60
 
+# https_client retry of the held credential on a 401 before re-enrolling: the
+# delay grows by <delta> after each rejection, up to <max>, then it re-enrolls.
```
MarcelKemp: **[S3 · non-blocking · introduced]** The number of retries is `ceil(auth_retry_max/auth_retry_delta)`, so the two options are not independent: an operator who raises `auth_retry_delta` to widen each wait gets *fewer* attempts (at `delta >= max`, exactly one), and the issue's own suggested starting point — 4.x's 5 attempts 10 s apart — cannot be expressed with these two knobs at all. The comment also promises "then it re-enrolls", which is false with `<enrollment><enabled>no</enabled>`: there it stops retrying and stays paused.

```suggestion
# https_client retry of the held credential on a 401 before re-enrolling: the
# delay grows by <delta> after each rejection, up to <max>. The number of
# retries is therefore ceil(<max>/<delta>) -- raising <delta> widens each wait
# but buys fewer attempts. Once the delay reaches <max> the agent re-enrolls;
# with <enrollment><enabled>no</enabled> it stops retrying and stays paused.
```

src/unit_tests/client-agent/test_https_client_bridge.c
```diff
@@ -881,40 +884,85 @@ static void start_client_successfully(void)
...
+    g_cred_retry_delay = 20;
+    expect_value(__wrap_sleep, seconds, 20);
+    expect_value(__wrap_hc_set_agent_identity, handle, FAKE_HANDLE);
+    expect_string(__wrap_hc_set_agent_identity, agent_id, keys.keyentries[0]->id);
+    expect_string(__wrap_hc_set_agent_identity, key_hex, keys.keyentries[0]->raw_key);
+    will_return(__wrap_hc_set_agent_identity, true);
+
+    bridge_cred_retry_thread(FAKE_HANDLE);
+}
+
```
MarcelKemp: **[S3 · non-blocking · introduced]** Three new branches have no test, and the most important one is the `g_cred_retry_delay = 0` line at `https_client_bridge.c:1330` — the only mechanism that makes the ramp per-incident rather than per-process — so deleting it would not fail the suite. Also uncovered: the `bridge_stopping()` abandon path (the only thing preventing `hc_set_agent_identity()` on a destroyed handle after `w_https_client_stop()`) and the `hc_set_agent_identity() == false` `merror`; the sibling `bridge_reenroll_thread` has tests for all three equivalents. Separately, `__wrap_getDefine_Int_default` ignores `low_name`, `min` and `max` and always returns `default_val`, so the 10/20/30 ramp these tests assert is just the built-in default — a misspelled option name or changed bounds at line 473 would keep all four tests green.

```c
static void test_cred_retry_reset_on_registered(void **state)
{
    (void)state;
    enable_enrollment();
    start_client_successfully();

    g_cred_retry_delay = 30;
    expect_value(__wrap_w_agentd_state_update, mode, UPDATE_STATUS);
    g_captured_callbacks.on_state_change(HC_STATE_REGISTERED, g_captured_callbacks.user_data);
    assert_int_equal(0, g_cred_retry_delay);

    expect_value(__wrap_hc_destroy, handle, FAKE_HANDLE);
    w_https_client_stop();
}
```

src/client-agent/src/https_client_bridge.c
```diff
@@ -213,6 +213,52 @@ static DWORD WINAPI bridge_reenroll_thread_win(LPVOID arg)
...
+
+    w_mutex_lock(&g_https_client_lock);
+    if (g_https_client_stopping) {
+        w_mutex_unlock(&g_https_client_lock);
+        return NULL;
+    }
+    if (!hc_set_agent_identity(handle, keys.keyentries[0]->id, keys.keyentries[0]->raw_key)) {
+        merror("https_client: failed to re-present the signing identity; traffic stays paused.");
+    }
+    w_mutex_unlock(&g_https_client_lock);
```
MarcelKemp: **[S3 · non-blocking · introduced]** A transient 401 leaves up to three "retrying held credential in N s" warnings and then complete silence on recovery, so the log cannot distinguish a recovered agent from one still stuck — the sibling path sets the precedent with `minfo("re-enrollment succeeded; reloading the signing identity")` at line 173. The locked bail at 243 also returns with no log at all, and that is the common case since `w_https_client_stop()` sets the flag under this very lock, which makes the unlocked pre-check at 237 both redundant and the only one that logs.

```suggestion
    w_mutex_lock(&g_https_client_lock);
    if (g_https_client_stopping) {
        w_mutex_unlock(&g_https_client_lock);
        mdebug1("https_client: agent shutting down; abandoning credential retry.");
        return NULL;
    }
    if (hc_set_agent_identity(handle, keys.keyentries[0]->id, keys.keyentries[0]->raw_key)) {
        minfo("https_client: held credential re-presented; retrying with a fresh token.");
    } else {
        merror("https_client: failed to re-present the signing identity; traffic stays paused.");
    }
    w_mutex_unlock(&g_https_client_lock);
```

MarcelKemp (changes requested):

src/client-agent/src/https_client_bridge.c
```diff
@@ -1258,6 +1316,7 @@ static void bridge_on_state_change(int state, void *user_data)
      * the latter, since it only fires on a paused->running transition and each
      * run starts believing it has paused nothing. os_delwait() is idempotent. */
     if (state == HC_STATE_REGISTERED) {
+        g_cred_retry_delay = 0;
```
MarcelKemp: **[S3 · non-blocking · introduced]** The ramp only re-arms here, so the new protection is one-shot per successful registration: if `bridge_reenroll_thread` enrolls successfully but the following `/startup` still 401s (a cluster whose `client.keys` is mid-sync), `g_cred_retry_delay` is still at the cap, `retry_max < retry_max` is false, and the agent re-enrolls immediately with no back-off — minting a new identity every control cycle. The loop it falls into is pre-existing (pre-PR every 401 spawned that thread at once), and you implemented the reset point #38610 itself specifies, so this is a design limit rather than a defect — but a second reset when a re-enrollment lands would close it.

The fix is outside the lines I can anchor: `g_cred_retry_delay = 0;` next to the `identity_ok` check in `bridge_reenroll_thread` (~line 193), under the lock it already holds, so a fresh credential also ends the incident.

src/client-agent/src/https_client_bridge.c
```diff
@@ -213,6 +213,44 @@ static DWORD WINAPI bridge_reenroll_thread_win(LPVOID arg)
...
+    }
+
+    w_mutex_lock(&g_https_client_lock);
+    if (g_https_client_stopping) {
+        w_mutex_unlock(&g_https_client_lock);
+        return NULL;
+    }
+    if (!hc_set_agent_identity(handle, keys.keyentries[0]->id, keys.keyentries[0]->raw_key)) {
+        merror("https_client: failed to re-present the signing identity; traffic stays paused.");
+    }
```
MarcelKemp: **[S3 · non-blocking · introduced]** When this returns false the thread logs and gives up — but `HttpsClientFacade::setAgentIdentity()` returns false *before* `m_authGate.release()` (`httpsClientFacade.cpp:461-464`), so `AuthGate` stays `Paused`, `reportAuthFailure()` can never fire again, and `bridge_on_reenroll_required` is never re-entered: the ramp never reaches the re-enrollment that would actually fix the identity. Pre-PR this 401 escalated straight to re-enrollment. Not reachable today — `OS_AddKey()` does `os_strdup(id, …)` (`src/shared/os_crypto/shared/keys.c:76`) so the id is never NULL, and the key already passed `bridge_key_is_valid()` — but escalating instead of returning is cheap, and it gives the success path the log line it is missing. Note that bumping `g_cred_retry_delay` would not help: without a `release()` the callback never fires again, so the escalation has to happen on this thread.

```suggestion
    if (!hc_set_agent_identity(handle, keys.keyentries[0]->id, keys.keyentries[0]->raw_key)) {
        /* AuthGate is still latched and nothing else can release it (release() is only
         * reachable from here and from bridge_reenroll_thread), so escalate on this
         * thread instead of leaving traffic paused for the rest of the process. */
        merror("https_client: failed to re-present the signing identity; escalating to re-enrollment.");
        w_mutex_unlock(&g_https_client_lock);
        if (agt->enrollment.enabled) {
            bridge_reenroll_thread(handle); /* Defined at line 133; we are already off the dispatcher. */
        }
        return NULL;
    }
    minfo("https_client: held credential re-presented; retrying with a fresh token.");
```

src/client-agent/src/https_client_bridge.c
```diff
@@ -423,16 +461,36 @@ static void bridge_on_startup_result(bool accepted, const char *metadata_json, v
...
+    const int retry_max = getDefine_Int_default("agent", "auth_retry_max", 1, 86400,
+                                               BRIDGE_CRED_RETRY_MAX_S_DEFAULT);
+
+    if (g_cred_retry_delay < retry_max) {
+        g_cred_retry_delay += retry_delta;
+        if (g_cred_retry_delay > retry_max) {
+            g_cred_retry_delay = retry_max;
+        }
+        mwarn("https_client: credential rejected (401); retrying held credential in %d s.",
+              g_cred_retry_delay);
```
MarcelKemp: **[S3 · non-blocking · aggravated]** This log promises a retry, but the ~60 s the gate stays latched reliably trips the producer pause, so the operator also gets "Manager unreachable. Pausing module event production." and a `NACTIVE` agent for the whole ramp — on the one path where the manager *is* reachable and the credential *is* good. `runStep()` returns `AuthFail` on every cycle while latched, `updateProducerPause()` counts that as `blocksDelivery`, and `producerPauseThreshold` defaults to 2 (`moduleConfig.hpp:88`). The outcome is still better than before (pre-PR the same pause lasted for the whole re-enrollment, an hour in #38610's evidence), but `controlStream.cpp:626-631` justifies the policy with a premise this PR invalidates: *"AuthFail only surfaces once the timestamp-corrected retry has also failed, so the key is genuinely bad and only re-enrollment can recover it."* Worth either exempting `AuthFail` from `blocksDelivery` while the ramp has budget left, or at minimum correcting that comment.

src/client-agent/src/https_client_bridge.c
```diff
@@ -213,6 +213,44 @@ static DWORD WINAPI bridge_reenroll_thread_win(LPVOID arg)
...
+
+/* Ramped delay (s). Confined to the single CallbackDispatcher worker, so no lock. */
+int g_cred_retry_delay = 0;
+
+/* Re-presenting the held id/key releases AuthGate, so the endpoints retry a transient
+ * 401 (fresh token, re-run skew correction) instead of re-enrolling. */
+void *bridge_cred_retry_thread(void *arg)
+{
+    hc_handle *handle = (hc_handle *)arg;
+    sleep((unsigned int)g_cred_retry_delay);
```
MarcelKemp: **[S3 · non-blocking · introduced]** The comment three lines up still says "Confined to the single CallbackDispatcher worker, so no lock", but this read happens on the spawned thread, while the dispatcher writes the same `int` at 471-473 and 1319 — an unsynchronized cross-thread read, and the risk this file already solves with a mutex for `g_https_client_stopping` (see the `_Atomic`/mingw note at 71-97). Reachable interleaving: a hard 401 on `/stateless` queues `on_reenroll_required` while a `/startup` is in flight; the `/startup` succeeds, so `on_state_change(HC_STATE_REGISTERED)` is queued behind it. The dispatcher is one FIFO worker (`callbackDispatcher.hpp:25`), so the callback sets 10 and spawns the thread, then the state change sets 0 — and the thread sleeps 0 s and re-presents immediately, with the log line still claiming 10 s.

Pass the delay through the thread argument the way `bridge_control_task_thread` already does (`os_malloc`'d ctx) instead of reading the global.

src/client-agent/src/https_client_bridge.c
```diff
@@ -423,16 +461,36 @@ static void bridge_on_startup_result(bool accepted, const char *metadata_json, v
 static void bridge_on_reenroll_required(void *user_data)
 {
     (void)user_data;
-    mwarn("https_client: credential rejected (401); re-enrolling.");
+
+    const int retry_delta = getDefine_Int_default("agent", "auth_retry_delta", 1, 3600,
+                                                  BRIDGE_CRED_RETRY_DELTA_S_DEFAULT);
+    const int retry_max = getDefine_Int_default("agent", "auth_retry_max", 1, 86400,
+                                               BRIDGE_CRED_RETRY_MAX_S_DEFAULT);
```
MarcelKemp: **[S3 · non-blocking · introduced]** `getDefine_Int_default()` caches nothing — each call re-opens and scans `local_internal_options.conf` and then `internal_options.conf` — so every 401 does up to four synchronous file opens on the dispatcher worker whose contract this file repeatedly says must not block, including on the re-enroll branch where neither value is used. Worse, `parse_define_int_value()` calls `merror_exit(INV_DEF, …)` on anything non-numeric or out of range, so `agent.auth_retry_delta=0` lets the agent start and then kills it days later, from the dispatcher thread, at the first 401. Every other https_client option is taken once in `bridge_build_config()` (1757-1776) precisely so validation cannot stop startup.

```c
/* Take both once in bridge_build_config(), next to the other https_client options,
   into file-static globals, and read only those here: */
const int retry_delta = g_cred_retry_delta;
const int retry_max = g_cred_retry_max;
```

etc/internal_options.conf
MarcelKemp: **[S3 · non-blocking · introduced]** Two things the operator cannot infer from this comment. First, the number of retries is `ceil(max/delta)`, so raising `delta` to widen each wait *buys fewer attempts* — and #38610's own suggested starting point (4.x's 5 attempts 10 s apart) cannot be expressed at all, since the issue asked for an attempt counter rather than a delay. Second, `min` is 1 on both options, so there is no configuration that restores the pre-PR "re-enroll on the first 401" policy, and `agent.auth_retry_max=0` — the natural way to switch this off — aborts `wazuh-agentd` via `merror_exit(INV_DEF)` instead of falling back to the default (the default only applies when the key is absent). "then it re-enrolls" is also untrue with `<enrollment><enabled>no</enabled>`: there it stops retrying and stays paused.

```suggestion
# https_client retry of the held credential on a 401 before re-enrolling: the
# delay grows by <delta> after each rejection, up to <max>. The number of
# retries is therefore ceil(<max>/<delta>) -- raising <delta> widens each wait
# but buys fewer attempts, and there is no value that disables the retry.
# Once the delay reaches <max> the agent re-enrolls; with
# <enrollment><enabled>no</enabled> it stops retrying and stays paused.
```

etc/internal_options.conf
```diff
@@ -192,6 +192,13 @@ agent.enrollment_retry_delta=5
...
+# Increment in seconds [1..3600]
+agent.auth_retry_delta=10
+# Cap in seconds [1..86400]
+agent.auth_retry_max=30
```
MarcelKemp: **[S3 · non-blocking · introduced]** These two do not reach `docs/ref/modules/client/configuration.md`, which has an "Enrollment Retry" section reproducing the sibling `agent.enrollment_retry_delta=5` / `agent.enrollment_retry_max=60` verbatim, plus the whole `agent.https_*` block. `grep -rn auth_retry docs/` returns nothing, so an operator investigating repeated re-enrollments cannot find the lever that decides how long the agent tolerates a rejected credential — and "Documentation Updates: N/A" only holds if that file does not count as documentation.

src/unit_tests/client-agent/test_https_client_bridge.c
```diff
@@ -881,40 +884,85 @@ static void start_client_successfully(void)
...
+    (void)state;
+
+    g_cred_retry_delay = 20;
+    expect_value(__wrap_sleep, seconds, 20);
+    expect_value(__wrap_hc_set_agent_identity, handle, FAKE_HANDLE);
+    expect_string(__wrap_hc_set_agent_identity, agent_id, keys.keyentries[0]->id);
+    expect_string(__wrap_hc_set_agent_identity, key_hex, keys.keyentries[0]->raw_key);
+    will_return(__wrap_hc_set_agent_identity, true);
+
+    bridge_cred_retry_thread(FAKE_HANDLE);
```
MarcelKemp: **[S3 · non-blocking · introduced]** This covers only the happy path, and three new branches have no test: (a) the `g_cred_retry_delay = 0` at bridge line 1319 — the single line that makes the ramp per-incident rather than per-process, so deleting it does not break the suite; (b) the `bridge_stopping()` early return, the only thing that stops `hc_set_agent_identity()` being called on a handle already destroyed by `w_https_client_stop()`; (c) the `merror` arm. `bridge_reenroll_thread` has tests for all three equivalents, so this is a coverage regression against the code it mirrors.

Two more: the `for (int delay = 10; delay <= 30; delay += 10)` loop is copy-pasted at 894 and 934 with the defaults written out by hand, and since 10 divides 30 the clamp at bridge lines 472-474 is never executed by anything. And `__wrap_getDefine_Int_default` ignores `low_name`/`min`/`max` and always returns `default_val`, so a misspelled option name or changed bounds at bridge 465-468 would still pass all four tests.

Separately: the arm that fixes the iteration-1 blocker is inside `#ifdef WIN32`, and `src/unit_tests/client-agent/CMakeLists.txt:89` gates this whole test behind `if(${TARGET} STREQUAL "agent")` — so no build compiles it. That is why the leak shipped.

src/client-agent/src/https_client_bridge.c
```diff
@@ -213,6 +213,44 @@ static DWORD WINAPI bridge_reenroll_thread_win(LPVOID arg)
...
+    if (bridge_stopping()) {
+        mdebug1("https_client: agent shutting down; abandoning credential retry.");
+        return NULL;
+    }
+
+    w_mutex_lock(&g_https_client_lock);
+    if (g_https_client_stopping) {
+        w_mutex_unlock(&g_https_client_lock);
+        return NULL;
+    }
```
MarcelKemp: **[S3 · non-blocking · introduced]** Two observability gaps on the path that matters most. The locked bail returns with no log, and that is the *common* shutdown case because `w_https_client_stop()` sets the flag under this same lock — while the unlocked probe three lines up does log. `bridge_stopping()`'s own comment (98) says it exists for callers that poll between attempts; this thread sleeps once and then takes the lock, so the probe is redundant with the re-check. And nothing is logged when the retry succeeds, so a recovered agent and a stuck one look identical in `ossec.log` — the sibling emits `minfo("re-enrollment succeeded; …")` at 176. Given the whole point of #38610 is that operators read a transient 401 as a dead key, recovery is the single most useful line this path could emit.

Worth stating too: `hc_set_agent_identity()` does no disk I/O (`httpsClientFacade.cpp:452-469` is `setKey` + `setAgentId` + `release`), so the retry re-presents whatever `OS_ReadKeys()` loaded at start-up. #38610's complaint that "fixing the key manually does not resume traffic, because nothing re-reads it" is unchanged, and the `merror` that tells the operator to do exactly that is still misleading.

tests/integration/test_enrollment/test_agent/test_agentd_enrollment.py
```diff
@@ -67,12 +67,17 @@
 # Test variables.
 socket_listener = None
 
+# Collapse the held-credential retry ramp so a rejected pre-existent key re-enrolls
+# within the poll deadline instead of after the production back-off (#38610).
+local_internal_options = {'agent.auth_retry_delta': '1', 'agent.auth_retry_max': '1'}
+
```
MarcelKemp: **[S3 · non-blocking · introduced]** Right idea, but three other tests drive re-enrollment through this same 401 path and did not get the collapse: `test_agentd_reconection_enrollment_with_keys.py:156`, `..._no_keys.py:158` and `..._no_keys_file.py:160` all call `wait_enrollment_try()`, which waits on `AGENTD_REENROLLING` with a fixed `timeout=150` (`tests/integration/test_agentd/utils.py:109`). All three reach that point holding a key with a `RemotedSimulator(mode='REJECT_AUTH')` in front, so they now need four `AuthGate` incidents — 10+20+30 s of ramp — before the line appears, cutting the margin from ~150 s to ~90 s on Windows runners. `..._no_keys.py:148` and `tests/integration/test_enrollment/conftest.py:87` also still state that "two consecutive 401s on the same credential" drive re-enrollment, which is exactly what this PR stops being true.

The other half: this override is module-level, so it applies to all 25 cases in `cases_wazuh_enrollment.yaml`. Under 1/1 there is exactly one 1-second retry, so the shipped 10/20/30 ramp, the cap clamp and the enrollment-disabled ramp are exercised nowhere end to end — a regression that went back to re-enrolling on the first 401 would still pass this suite.

src/client-agent/src/https_client_bridge.c
```diff
@@ -423,16 +461,36 @@ static void bridge_on_startup_result(bool accepted, const char *metadata_json, v
...
+#else
+        w_create_thread(bridge_cred_retry_thread, g_https_client);
+#endif
+        return;
+    }
 
     if (!agt->enrollment.enabled) {
         merror("https_client: re-enrollment required but auto-enrollment is disabled "
                "(<enrollment><enabled>); traffic stays paused until the key is fixed manually.");
         return;
```
MarcelKemp: **[S2 · non-blocking · pre-existing]** With auto-enrollment disabled, exhausting the ramp leaves `AuthGate` latched for the rest of the process. `release()` (`authGate.hpp:59`) is reachable only from `setAgentIdentity()` (`httpsClientFacade.cpp:467`), whose only production callers are lines 179 and 239 of this file — neither runs on this branch. So 401s #1-#3 retry at 10/20/30 s, #4 lands here, logs, and returns without spawning anything; `on_reenroll_required` can never fire again and `HC_STATE_REGISTERED` is unreachable with the gate closed, so even the ramp cannot reset. The disabled case is the one with no enrollment behind it, which is precisely where the retry should have no cap. The PR body reads as if this were fixed; it is deferred by ~60 s. Not introduced by this PR — flagging it for a follow-up issue, not for this review. Suggested shape for that issue: on this branch keep retrying at `retry_max` forever instead of returning, and reword the message to describe the real state.

src/client-agent/src/https_client_bridge.c
```diff
@@ -414,16 +452,36 @@ static void bridge_on_startup_result(bool accepted, const char *metadata_json, v
 static void bridge_on_reenroll_required(void *user_data)
 {
     (void)user_data;
-    mwarn("https_client: credential rejected (401); re-enrolling.");
+
+    const int retry_delta = getDefine_Int_default("agent", "auth_retry_delta", 1, 3600,
+                                                  BRIDGE_CRED_RETRY_DELTA_S_DEFAULT);
+    const int retry_max = getDefine_Int_default("agent", "auth_retry_max", 1, 86400,
+                                               BRIDGE_CRED_RETRY_MAX_S_DEFAULT);
```
MarcelKemp: **[S2 · blocking · introduced]** Reading these two options here means an out-of-range value does not refuse the start — it aborts the agent at its first 401. `parse_define_int_value()` calls `merror_exit` → `exit(1)`, which runs `atexit(w_https_client_stop)` (`agentd.c:125`) **on the CallbackDispatcher worker**, the very thread this callback runs on: `hc_destroy` → `~HttpsClientFacade` → `m_dispatcher.stop()` → `m_worker.join()` joining itself → `EDEADLK`; that first throw is caught by the CID-562609 guard in `~HttpsClientFacade`, but `stop()` sets `m_running = false` only *after* the join, so `~CallbackDispatcher()` self-joins again and throws out of an implicitly-`noexcept` destructor → `std::terminate`.

This is the exact invariant commit `8dc5db02` wrote into `src/client-agent/src/config.c:111-113` a week ago for the sibling pair — *"an out-of-range value refuses the start instead of killing the agent at its first failed re-enrollment"* — documented in `docs/ref/modules/client/configuration.md:437-440` and guarded by `test_bridge_does_not_resolve_the_enrollment_retry_ramp`. It is not a hypothetical typo either: `min` is 1 for both, so there is no in-range way to restore the pre-PR behaviour, which makes `agent.auth_retry_max=0` the natural thing for an operator to try — and `getDefine_Int_default`'s default only applies when the key is *absent*, not out of range.

This was S3 in my last review; it is blocking now because the base branch grew the contract underneath it, and because I traced the abort. The other 15 internal options in this file are resolved once in `bridge_build_config()` (lines 1678, 1748-1767) — this also removes up to four synchronous file reads per 401 from the thread whose contract forbids blocking, and gives F12 its natural reset point.

```suggestion
    const int retry_delta = g_cred_retry_delta;
    const int retry_max = g_cred_retry_max;
```

with, next to the other internal options in `bridge_build_config()` (~line 1767):

```c
/* Resolved here, not in the 401 callback: out of range must refuse the start rather than
 * kill the agent at its first 401 -- the same invariant ClientConf() documents for the
 * enrollment_retry_* pair (src/client-agent/src/config.c:111-113). */
g_cred_retry_delta = getDefine_Int_default("agent", "auth_retry_delta", 1, 3600,
                                           BRIDGE_CRED_RETRY_DELTA_S_DEFAULT);
g_cred_retry_max = getDefine_Int_default("agent", "auth_retry_max", 1, 86400,
                                         BRIDGE_CRED_RETRY_MAX_S_DEFAULT);
```

src/client-agent/src/https_client_bridge.c
```diff
@@ -414,16 +452,36 @@ static void bridge_on_startup_result(bool accepted, const char *metadata_json, v
...
+#ifdef WIN32
+        CloseHandle(w_create_thread(NULL, 0, bridge_cred_retry_thread_win, g_https_client, 0, NULL));
+#else
+        w_create_thread(bridge_cred_retry_thread, g_https_client);
+#endif
+        return;
+    }
 
     if (!agt->enrollment.enabled) {
         merror("https_client: re-enrollment required but auto-enrollment is disabled "
```
MarcelKemp: **[S2 · non-blocking · pre-existing]** With `<enrollment><enabled>no</enabled>` this branch is a dead end, and the ramp above delays it by ~60 s rather than removing it: nothing here calls `hc_set_agent_identity()`, which is the only path to `AuthGate::release()`, so the latch stays `Paused`, `reportAuthFailure()` can never fire again, and `HC_STATE_REGISTERED` is unreachable so the ramp cannot even reset. The disabled-enrollment case is precisely the one with no re-enrollment behind it, so it is the one where the retry should have no ceiling at all — keeping it re-presenting at the cap is the only recovery available in that configuration.

Not introduced by this PR — flagging it for a follow-up issue, not for this review.

src/client-agent/src/https_client_bridge.c
```diff
@@ -204,6 +204,44 @@ static DWORD WINAPI bridge_reenroll_thread_win(LPVOID arg)
...
+
+/* Ramped delay (s). Confined to the single CallbackDispatcher worker, so no lock. */
+int g_cred_retry_delay = 0;
+
+/* Re-presenting the held id/key releases AuthGate, so the endpoints retry a transient
+ * 401 (fresh token, re-run skew correction) instead of re-enrolling. */
+void *bridge_cred_retry_thread(void *arg)
+{
+    hc_handle *handle = (hc_handle *)arg;
+    sleep((unsigned int)g_cred_retry_delay);
```
MarcelKemp: **[S3 · non-blocking · introduced]** This comment is false as of this diff: `bridge_cred_retry_thread` is not the CallbackDispatcher worker, and it reads `g_cred_retry_delay` on line 218 with no lock and no atomic while the dispatcher writes it on 462-464 and 1310. Two reachable interleavings: the dispatcher is a single FIFO worker, so an in-flight `/startup` that lands accepted right after the 401 queues `on_reenroll_required` (delay → 10, thread spawned) and then `on_state_change(HC_STATE_REGISTERED)`, which zeroes the delay before the new thread's first instruction — it reads 0, `sleep(0)`, and the back-off the PR exists for silently disappears while the log claims 10 s; separately, `StartupAccepted` sets `effects.releaseGate = true`, so a re-armed gate can spawn a second and third retry thread while the first still sleeps, all racing the same global.

The file already has the pattern for this (`bridge_control_task_ctx` + `os_malloc`/`os_free`): pass the delay to the thread in a heap context instead of reading a global from another thread.

```c
struct bridge_cred_retry_ctx {
    hc_handle *handle;
    int delay;
};
```

etc/internal_options.conf
```diff
@@ -193,6 +193,13 @@ agent.enrollment_retry_delta=5
 # Cap in seconds [1..86400]
 agent.enrollment_retry_max=60
 
+# https_client retry of the held credential on a 401 before re-enrolling: the
+# delay grows by <delta> after each rejection, up to <max>, then it re-enrolls.
```
MarcelKemp: **[S3 · non-blocking · introduced]** The comment documents the delay bounds but not the two things an operator actually needs: the retry count is `ceil(<max>/<delta>)`, so raising `<delta>` to widen each wait *buys fewer attempts*, and the total wait before re-enrolling is quadratic — `~<max>*(<max>/<delta>+1)/2`, which for the in-range `delta=1, max=86400` is about 118 years with traffic paused throughout. "then it re-enrolls" is also untrue with `<enrollment><enabled>no</enabled>`: there it stops retrying and stays paused. And after the rebase the block immediately above says *"Both are read once at startup; editing them on a running agent has no effect until it restarts"*, which does not hold for this new pair — it is re-read on every 401 (see my comment on `https_client_bridge.c:456`).

```suggestion
# https_client retry of the held credential on a 401 before re-enrolling: the delay
# grows by <delta> after each rejection, up to <max>. The number of retries is
# therefore ceil(<max>/<delta>), and the total wait before re-enrolling is about
# <max>*(<max>/<delta>+1)/2 -- raising <delta> widens each wait but buys fewer
# attempts. Once the delay reaches <max> the agent re-enrolls; with
# <enrollment><enabled>no</enabled> it stops retrying and stays paused instead.
```

etc/internal_options.conf
```diff
@@ -193,6 +193,13 @@ agent.enrollment_retry_delta=5
...
+# Increment in seconds [1..3600]
+agent.auth_retry_delta=10
+# Cap in seconds [1..86400]
+agent.auth_retry_max=30
```
MarcelKemp: **[S3 · non-blocking · introduced]** Neither option reaches `docs/ref/modules/client/configuration.md`, and after the rebase that omission is worse than incomplete: lines 437-440 of that file now state, for the sibling `enrollment_retry_*` pair, that both are *"resolved once, when the agent starts... an out-of-range value refuses the start rather than terminating the agent later"* — the exact opposite of what this pair does. An operator reading the reference will draw the wrong conclusion about both pairs, and "Documentation Updates: N/A" in the PR body only holds if that file does not count as documentation.

```ini
# https_client retry of the held credential on a 401 before re-enrolling: the delay
# grows by <delta> seconds after each rejection, up to <max>; the number of retries is
# ceil(<max>/<delta>). Once the delay reaches <max> the agent re-enrolls.
agent.auth_retry_delta=10
agent.auth_retry_max=30
```

src/unit_tests/client-agent/test_https_client_bridge.c
```diff
@@ -918,40 +921,85 @@ static void start_client_successfully(void)
...
```
MarcelKemp: **[S3 · non-blocking · introduced]** Three new branches are uncovered and none of the callback tests pins *which* thread gets spawned. Swap `bridge_cred_retry_thread` for `bridge_reenroll_thread` in the two branches of `bridge_on_reenroll_required` — i.e. reintroduce exactly the identity churn this PR fixes — and `..._first_401_retries_held_credential`, `..._reenrolls_after_retries_exhausted` and `..._disabled_retries_then_logs_error` all still pass, because they assert only log strings and the shared `__wrap_CreateThread` never `check_expected()`s its function pointer; the deleted test's "this path must not even reach the call" assertion was the only thing covering that and was dropped without replacement. Still missing: the `HC_STATE_REGISTERED` reset (the one line that makes the ramp per-incident, so deleting line 1310 breaks nothing), the `bridge_stopping()` early exit, and the `merror` branch — and since `10` divides `30`, the clamp on 463-465 never executes in any test.

The rebase also brought `g_resolved_options` (line 218), which exists precisely to pin an option's name and bounds; nothing uses it for the two new ones, so a misspelled key or changed limits on 456-459 would pass the whole suite.

```c
static void test_cred_retry_reset_on_registered(void **state)
{
    (void)state;
    enable_enrollment();
    start_client_successfully();

    g_cred_retry_delay = 30;
    expect_value(__wrap_w_agentd_state_update, mode, UPDATE_STATUS);
    g_captured_callbacks.on_state_change(HC_STATE_REGISTERED, g_captured_callbacks.user_data);
    assert_int_equal(0, g_cred_retry_delay);

    expect_value(__wrap_hc_destroy, handle, FAKE_HANDLE);
    w_https_client_stop();
}
```

src/client-agent/src/https_client_bridge.c
```diff
@@ -204,6 +204,44 @@ static DWORD WINAPI bridge_reenroll_thread_win(LPVOID arg)
...
+    if (bridge_stopping()) {
+        mdebug1("https_client: agent shutting down; abandoning credential retry.");
+        return NULL;
+    }
+
+    w_mutex_lock(&g_https_client_lock);
+    if (g_https_client_stopping) {
+        w_mutex_unlock(&g_https_client_lock);
+        return NULL;
+    }
```
MarcelKemp: **[S3 · non-blocking · introduced]** The branch that actually decides — this one, under the lock — returns with no log at all, and it is the common case, because `w_https_client_stop()` sets the flag under this same lock; the unlocked pre-check on 220 that does log is redundant here (it exists in `bridge_reenroll_thread` because that loop polls it *between* attempts, and there is no loop here). The success path is silent too: a transient 401 leaves up to three "retrying held credential in N s" warnings and then nothing, so the log cannot distinguish a recovered agent from a stuck one, while the sibling `bridge_reenroll_thread` logs `minfo("...reloading the signing identity")` at the equivalent point. Worth noting that #38610 calls the "fix the key manually" message misleading "because nothing re-reads it", and that is still true: `setAgentIdentity()` never touches disk, so this re-presents whatever `OS_ReadKeys()` loaded at startup.

```suggestion
    w_mutex_lock(&g_https_client_lock);
    if (g_https_client_stopping) {
        w_mutex_unlock(&g_https_client_lock);
        mdebug1("https_client: agent shutting down; not re-presenting the signing key.");
        return NULL;
    }
```

src/client-agent/src/https_client_bridge.c
```diff
@@ -204,6 +204,44 @@ static DWORD WINAPI bridge_reenroll_thread_win(LPVOID arg)
...
+    if (!hc_set_agent_identity(handle, keys.keyentries[0]->id, keys.keyentries[0]->raw_key)) {
+        merror("https_client: failed to re-present the signing identity; traffic stays paused.");
+    }
```
MarcelKemp: **[S3 · non-blocking · introduced]** When this returns false the thread logs and gives up, but `HttpsClientFacade::setAgentIdentity()` returns false *before* `m_authGate.release()` (`httpsClientFacade.cpp:461-464`), so the gate stays `Paused`, `reportAuthFailure()` can never fire again and `bridge_on_reenroll_required` is never re-entered — the ramp never reaches the re-enrollment that would fix the identity, where pre-PR that same 401 escalated straight to it. Not reachable today (`OS_AddKey()` `os_strdup`s the id, and the key already passed `bridge_key_is_valid()`), but escalating instead of returning is cheap. Bumping `g_cred_retry_delay` would not help: with no `release()` the callback never fires again, so the escalation has to happen on this thread.

```suggestion
    if (!hc_set_agent_identity(handle, keys.keyentries[0]->id, keys.keyentries[0]->raw_key)) {
        /* AuthGate is still latched and nothing else can release it (release() is only
         * reachable from here and from bridge_reenroll_thread), so escalate on this
         * thread instead of leaving traffic paused for the rest of the process. */
        merror("https_client: failed to re-present the signing identity; escalating to re-enrollment.");
        w_mutex_unlock(&g_https_client_lock);
        if (agt->enrollment.enabled) {
            bridge_reenroll_thread(handle); /* Defined above; we are already off the dispatcher. */
        }
        return NULL;
    }
    minfo("https_client: held credential re-presented; retrying with a fresh token.");
```

src/client-agent/src/https_client_bridge.c
```diff
@@ -204,6 +204,44 @@ static DWORD WINAPI bridge_reenroll_thread_win(LPVOID arg)
...
 
+#define BRIDGE_CRED_RETRY_DELTA_S_DEFAULT 10
+#define BRIDGE_CRED_RETRY_MAX_S_DEFAULT 30
+
+/* Ramped delay (s). Confined to the single CallbackDispatcher worker, so no lock. */
+int g_cred_retry_delay = 0;
+
+/* Re-presenting the held id/key releases AuthGate, so the endpoints retry a transient
+ * 401 (fresh token, re-run skew correction) instead of re-enrolling. */
+void *bridge_cred_retry_thread(void *arg)
```
MarcelKemp: **[S3 · non-blocking · introduced]** The whole mechanism rests on behaviour that is nowhere in the module's contract: that `hc_set_agent_identity()` releases `AuthGate` even when the id and key are *unchanged*. `include/https_client.h:265` describes it as the call you make "with the new id/key to resume" and `:484` documents it as the post-re-enrollment identity swap; it happens to work only because `ConfigKeyProvider::setKey()` is idempotent (`keyProvider.cpp:44`) and `setAgentIdentity()` calls `release()` without comparing against the previous identity, and no test in `src/client-agent/https_client/tests` covers the unchanged-identity case. A perfectly reasonable short-circuit on the C++ side ("same id, same key, nothing to do") would disable this feature completely — no release, no `/startup`, no escalation, agent stays paused — with all three suites still green.

```c
/* include/https_client.h, above hc_set_agent_identity() */
/// Re-presenting the CURRENT id/key is supported and is not a no-op: it releases the
/// auth pause and lets the endpoints retry with a fresh token. The agent bridge relies
/// on this to retry a transient 401 without re-enrolling (#38610), so setAgentIdentity()
/// must not short-circuit on an unchanged identity.
```

src/client-agent/src/https_client_bridge.c
```diff
@@ -414,16 +452,36 @@ static void bridge_on_startup_result(bool accepted, const char *metadata_json, v
...
+            g_cred_retry_delay = retry_max;
+        }
+        mwarn("https_client: credential rejected (401); retrying held credential in %d s.",
+              g_cred_retry_delay);
+#ifdef WIN32
+        CloseHandle(w_create_thread(NULL, 0, bridge_cred_retry_thread_win, g_https_client, 0, NULL));
+#else
+        w_create_thread(bridge_cred_retry_thread, g_https_client);
+#endif
+        return;
```
MarcelKemp: **[S3 · non-blocking · aggravated]** The latched `AuthGate` now lasts long enough to pause event production for the whole ramp: while it is latched `ControlStream::runStep()` returns `AuthFail` every cycle without sending anything, `updateProducerPause()` counts that as `blocksDelivery`, and with `notifyIntervalS` 10 s and `producerPauseThreshold` 2 the 60 s ramp delivers ~6 such cycles per incident instead of ~1. So `bridge_on_producer_pause(true, ...)` fires `mwarn(SERVER_UNAV)` — "Manager unreachable. Pausing module event production." — calls `os_setwait()` and leaves the `.state` at `GA_STATUS_NACTIVE` for a minute, in the exact scenario where the manager is reachable and the credential is good. The pause itself is pre-existing (it appears in #38610's own agent log, same second as the 401); what this PR changes is how long it lasts, and `controlStream.cpp:626-631` still justifies the policy with a premise this PR invalidates: *"AuthFail only surfaces once the timestamp-corrected retry has also failed, so the key is genuinely bad and only re-enrollment can recover it."*

Worth deciding deliberately: either don't count an `AuthFail` cycle as `blocksDelivery` while the ramp is below its cap, or accept it and fix that comment.

src/client-agent/src/https_client_bridge.c
```diff
@@ -1249,6 +1307,7 @@ static void bridge_on_state_change(int state, void *user_data)
      * the latter, since it only fires on a paused->running transition and each
      * run starts believing it has paused nothing. os_delwait() is idempotent. */
     if (state == HC_STATE_REGISTERED) {
+        g_cred_retry_delay = 0;
```
MarcelKemp: **[S3 · non-blocking · introduced]** The ramp only re-arms here, so the new protection is one-shot per successful registration: if `bridge_reenroll_thread` enrolls successfully but the following `/startup` still 401s (a cluster whose `client.keys` is mid-sync), `g_cred_retry_delay` is still at the cap, `30 < 30` is false, and the agent re-enrolls immediately with no back-off — minting a new identity every cycle. The loop it falls into is pre-existing and you implemented the reset point #38610 itself specifies, so this is a design limit rather than a miss; a second reset when a re-enrollment lands would close it.

The fix is outside the lines I can anchor: `g_cred_retry_delay = 0;` next to the `identity_ok` check in `bridge_reenroll_thread` (~line 179), under the lock it already holds, so a fresh credential also ends the incident.

tests/integration/test_enrollment/test_agent/test_agentd_enrollment.py
MarcelKemp: **[S3 · non-blocking · introduced]** Collapsing the ramp here is right, but the same treatment is missing from the three tests that reach the identical 401 path: `test_agentd_reconection_enrollment_with_keys.py:156`, `..._no_keys.py:158` and `..._no_keys_file.py:160` all call `wait_enrollment_try()`, whose `timeout` is a fixed 150 s (`test_agentd/utils.py:109`), and all three arrive holding a key behind a `RemotedSimulator(mode='REJECT_AUTH')` — so with the shipped 10/30 they need four `AuthGate` incidents before the re-enrollment line is written and their margin drops from ~150 s to ~90 s on already-loaded Windows runners. Two comments are now false as well: `..._no_keys.py:148` and `test_enrollment/conftest.py:87` still say two consecutive 401s on the same credential trigger re-enrollment. And because the override is module-scoped at 1/1, the real ramp, the clamp and the disabled-enrollment ramp are not exercised end to end anywhere, so a regression back to re-enrolling on the first 401 would still pass this suite.

Minor, while you are here: the docstring's `parameters:` block lists every fixture except the new `configure_local_internal_options`, and the two keys are raw strings where `test_agentd_initial_enrollment_retries.py:82` uses the framework constant (`AGENTD_ENROLLMENT_RETRY_DELTA`).

```python
# in each of the three test_agentd/test_reconnection tests, next to their existing override:
local_internal_options.update({AGENTD_TIMEOUT: '5',
                               'agent.auth_retry_delta': '1',
                               'agent.auth_retry_max': '1'})
```

MarcelKemp:
Given the requirements set out in the following objective, this change no longer makes sense, so I’m going to close it…
- https://github.com/wazuh/wazuh/issues/38940

PR #38727, opened 2026-08-31
Fix twelve 5.0.0 readiness findings and settle the Active Response cursor contract

### LINKED ISSUE #38549: HTTPS/inventory-sync: small correctness defects found during 5.0.0 readiness testing

ignaciogalle12git:
## Description

Twelve findings from #38549 ("HTTPS/inventory-sync: small correctness defects found during 5.0.0 readiness testing"), plus the Active Response cursor contract the first two of them depend on.

Two changes were added after the first review round, both from decisions the team took since:

- The agent's enrollment retry ramp is now fixed at startup instead of re-read on every retry. The earlier revision only validated it at startup and left the loop reading, which kept the values editable on a running agent; the agent team chose the fixed value.
- An internal option present with an empty value now refuses the daemon's start instead of being read as `0`.

Rebased onto `5.0.0` at `7990fc035c`. That base carries the manager's move to the strict configuration loader (`3814555f33`, `93d4e010de`), which removed two commits from this branch:

- **Validating the candidate configuration file** is now upstream behavior. `RemotedConfig()` loads the effective document once with `-c` honoured, `-t` validates the whole file through `w_mconf_validate()`, and the enrollment bridge reads `auth` from that same document. The commit here threaded a `cfg` parameter to reach the candidate's `auth` block and depended on the `CAUTHD` flag, which `93d4e010de` deleted. Nothing to port: `remoted -t` already validates the proposed file, and the chroot hazard the first review round found is gone with the file re-read.
- **The `adminServer_test.cpp` formatting** was fixed upstream in `0ade596ce0`.

## Proposed Changes

- **Keep the Active Response cursor from skipping or freezing.** The bookmark advanced past a Task-Manager rejection, so a document rejected once was never retried; a document the schema allows but that carries no `event` aborted the whole batch. The cursor now advances only on delivery, and a transport failure holds it instead of dropping the entry.
- **Fix the agent's retry ramp at startup.** `agent.enrollment_retry_max` and `_delta` were read inside both enrollment loops, so an out-of-range value killed an already-enrolled agent at its first failed re-enrollment, and any value was editable on a running agent. `ClientConf()` now resolves both once, before either loop runs, and the loops read the resolved values. An out-of-range value refuses the start; editing the file afterwards takes effect on the next restart.
- **Refuse an internal option written with no value.** An empty value reached `atoi()` as `0`, which the range check accepted on every option whose minimum is `0` and rejected on the rest, so the same unfinished edit either started the daemon with a value nobody chose or stopped it. It is now rejected like any other invalid value. An absent option still falls back to its default.
- **Count the transport's shed 503s and split two remoted throttles.** Five distinct shed sources answered `503` with no counter of their own, and two unrelated conditions shared one throttle window.
- **Documentation.** Three timing references corrected (`http_read_timeout` vs `http_request_timeout` on an upload, the link-speed floor `http_write_timeout` implies, the detection window `agents_disconnection_time` produces), and the misspelled `Cannot comunicate with master` string.

### Results and Evidence

Every suite below was run on the rebased branch at `f81db0076a`, in the manager devcontainer for the C and C++ targets and in a virtualenv built from `5_testunit_framework.yml` for the framework.

```
$ ./remoted_module_utest
[==========] 793 tests from 101 test suites ran. (117377 ms total)
[  PASSED  ] 793 tests.
```

```
$ ./inventory_sync_server_utest
[==========] 248 tests from 21 test suites ran. (1948 ms total)
[  PASSED  ] 248 tests.
```

```
$ ./uds_http_server_utest
[==========] 116 tests from 9 test suites ran. (29183 ms total)
[  PASSED  ] 116 tests.
```

```
$ PYTHONPATH=$PWD/api:$PWD/framework python -m pytest framework/wazuh/core/indexer/tests/test_active_response.py -q
75 passed in 0.59s
```

Both cmocka suites in full, not only the affected cases:

```
$ cmake -DTARGET=manager .. && make && ctest
100% tests passed, 0 tests failed out of 122

$ cmake -DTARGET=agent .. && make && ctest
100% tests passed, 0 tests failed out of 116
```

The three suites this branch adds cases to were run individually as well: `test_agentd`, `test_https_client_bridge`, `test_validate_op` and `test_agent_validate_op` pass under `TARGET=agent`, and `test_local-server`, `test_validate_op` and `test_agent_validate_op` under `TARGET=manager`.

Every new test was also run inverted (assertion negated or the fix reverted) to confirm it fails without the change.

Two environment notes, neither attributable to this branch:

- `content_manager_ctest` does not link OpenSSL and fails at link time on the base as well, so the `TEST=1` build needs `-k`.
- `UdsHttpServerTest.ThreeHundredConcurrentDeferralsOnTwoIoThreads` fails under a 1024 file-descriptor limit, which is the Docker default: it needs about 600 descriptors for 300 simultaneous in-flight requests and parks only ~235 of them. Verified as pre-existing by reverting this branch's `uds_http_server` diff and running the test against `7990fc035c`, where it fails identically (234/300). With `ulimit -n 65536` it passes in 86 ms.

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

Windows and macOS compilation rely on CI.

### Artifacts Affected

None.

### Configuration Changes

No new options and no default value changes. Two behavior changes on existing configuration:

- `agent.enrollment_retry_max` and `agent.enrollment_retry_delta` are read once at startup; editing them on a running agent no longer takes effect until it restarts.
- Any internal option present with an empty value now refuses the daemon's start. Recorded under Breaking changes in the release notes.

### Tests Introduced

- `test_active_response.py`: the cursor cases, including the transport failure that must hold it.
- `test_https_client_bridge.c`: the bridge no longer resolves the retry ramp, and the re-enrollment back-off follows the values `ClientConf()` resolved, driven with a ramp that is not the shipped default.
- `test_validate_op.c`: an empty value refuses the start, an absent key falls back to its default, a present value is read.

## Review Checklist

- [x] Code changes reviewed


ignaciogalle12git:
>[!IMPORTANT] 
> API Endpoints IT failing, tracked on https://github.com/wazuh/qa-integration-framework/issues/835

NahuFigueroa97 (dismissed):
### Active Response cursor fix (items 1 & 2) — the core fix

```
$ PYTHONPATH=api:framework python3 -m pytest \
  framework/wazuh/core/indexer/tests/test_active_response.py -v

...
framework/wazuh/core/indexer/tests/test_active_response.py::TestActiveResponseHelpers::TestFetchActiveResponseDocs::test_search_after_is_still_bounded_above PASSED
framework/wazuh/core/indexer/tests/test_active_response.py::TestActiveResponseHelpers::TestFetchActiveResponseDocs::test_the_page_extent_is_recorded_before_filtering PASSED
framework/wazuh/core/indexer/tests/test_active_response.py::TestActiveResponseBuilder::TestEnrich::test_an_event_not_visible_yet_holds_the_page PASSED
framework/wazuh/core/indexer/tests/test_active_response.py::TestActiveResponseBuilder::TestEnrich::test_holding_on_the_first_response_leaves_the_cursor_alone PASSED
framework/wazuh/core/indexer/tests/test_active_response.py::TestActiveResponseBuilder::TestEnrich::test_an_event_missing_past_the_grace_window_is_discarded PASSED
framework/wazuh/core/indexer/tests/test_active_response.py::TestActiveResponseBuilder::TestDispatch::test_dispatch_holds_the_page_on_a_failed_send PASSED
framework/wazuh/core/indexer/tests/test_active_response.py::TestActiveResponseBuilder::TestDispatch::test_dispatch_holds_the_page_when_task_manager_is_unreachable PASSED
framework/wazuh/core/indexer/tests/test_active_response.py::TestActiveResponseBuilder::TestDispatch::test_dispatch_advances_the_cursor_on_a_page_that_dispatched_nothing PASSED
framework/wazuh/core/indexer/tests/test_active_response.py::TestActiveResponseBuilder::TestDispatch::test_dispatch_follows_the_page_not_the_last_dispatched_response PASSED
...

======================== 75 passed, 9 warnings in 0.57s ========================
```

75/75 passed — including the tests that directly pin the two bugs fixed here: the future-timestamp
bookmark skip (item 1) and the missing-`event`-field batch poisoning (item 2), plus the
transport-failure cursor hold added later in the branch.

### Shed-503 counters, throttle split, VD-scan budget (items 52/55 and related)

C++ (gtest), rebuilt from this branch's code:

```
$ ./uds_http_server_utest --gtest_filter="*TheBudgetShedIsCounted*:*TransportDiagnostics*"

[----------] 6 tests from TransportDiagnosticsTest
[ RUN      ] TransportDiagnosticsTest.AllZeroBeforeStart
[       OK ] TransportDiagnosticsTest.AllZeroBeforeStart (0 ms)
[ RUN      ] TransportDiagnosticsTest.TheBudgetShedIsCounted
[       OK ] TransportDiagnosticsTest.TheBudgetShedIsCounted (20 ms)
[ RUN      ] TransportDiagnosticsTest.AParkedDeferralIsVisibleAndReleasesOnReply
[       OK ] TransportDiagnosticsTest.AParkedDeferralIsVisibleAndReleasesOnReply (20 ms)
[ RUN      ] TransportDiagnosticsTest.IdentityInjectionRendersTheConsumerVocabulary
[       OK ] TransportDiagnosticsTest.IdentityInjectionRendersTheConsumerVocabulary (0 ms)
[ RUN      ] TransportDiagnosticsTest.TheStartupLineReportsTheResolvedRouteClasses
[       OK ] TransportDiagnosticsTest.TheStartupLineReportsTheResolvedRouteClasses (0 ms)
[ RUN      ] TransportDiagnosticsTest.TheBudgetHintSentenceRendersOnlyWhenConfigured
[       OK ] TransportDiagnosticsTest.TheBudgetHintSentenceRendersOnlyWhenConfigured (15 ms)
[----------] 6 tests from TransportDiagnosticsTest (56 ms total)
[  PASSED  ] 6 tests.

$ ./uds_http_server_utest
...
[==========] 115 tests from 9 test suites ran. (25674 ms total)
[  PASSED  ] 115 tests.
```

`TheBudgetShedIsCounted` drives a real socket/budget scenario end to end, not a mocked counter.

```
$ ./remoted_module_utest --gtest_filter="*TheStartupBudgetCoversTheOffsetQueryToo*:*AdminServer*"

[----------] 1 test from ScanVdHandlerTest
[ RUN      ] ScanVdHandlerTest.TheStartupBudgetCoversTheOffsetQueryToo
[       OK ] ScanVdHandlerTest.TheStartupBudgetCoversTheOffsetQueryToo (0 ms)
[----------] 5 tests from AdminServerTest
[       OK ] AdminServerTest.GetRootAnswersTheLivenessProbe
[       OK ] AdminServerTest.GetMetricsDumpsTheModuleFamilies
[       OK ] AdminServerTest.UnknownRouteAnswers404AndWrongVerb405
[       OK ] AdminServerTest.AdminBindFailureOnlyWarnsAndTheModuleKeepsServing
[       OK ] AdminServerTest.StopUnlinksTheSocketAndARestartBringsItBack
[==========] 6 tests from 2 test suites ran.
[  PASSED  ] 6 tests.

$ ./remoted_module_utest
...
[==========] 785 tests from 101 test suites ran. (120482 ms total)
[  PASSED  ] 785 tests.
```

Full `remoted_module_utest` suite: 785/785 — no regressions anywhere else in the module from this
branch's changes.

### Candidate-config `-t` restriction (item 60)

C (CMocka), rebuilt from this branch's code:

```
$ ./remoted/test_secure
...
[ RUN      ] test_remoted_enrollment_config_honors_candidate_path
[       OK ] test_remoted_enrollment_config_honors_candidate_path
...
[==========] tests: 67 test(s) run.
[  PASSED  ] 67 test(s).
```

67/67 passed, including the test pinning that `-t` (config validation) honors a candidate `-c`
path while the real listener keeps using the post-chroot-safe relative path.


**LGTM.** No blocking correctness, memory-safety, or race-condition issues. Real evidence above
covers the two highest-risk changes in the branch (the AR cursor rewrite and the throttle/counter
split) end to end, plus a clean merge check. Non-blocking recommendations from the code review
still stand: consider squashing `674d193e87`+`685caee4d8` (the first is not independently
shippable).

jotacarma90 (approved):

Issue #35, opened 2016-07-05
What does wazuh mean, and how to pronounce?


GeiliCode:
I'm curious, and didn't find the answer through Google.

santiago-bassett:
Hi @GeiliCode, 

we are still trying to figure out what Wazuh means 👍 

We do have a little blue Wolf as a logo, so I personally think that the 'W' is for the Wolf, and 'azuh' is very similar to 'azul' (blue in Spanish). 

Regarding pronunciation I say 'Wazu' (I guess how it would sound in Spanish), but I've heard it in many different ways :-)

Best regards!

GeiliCode:
:-) Thank you very much

ASalazarMX:
I'm sorry to revive this closed issue, but have this been defined? I don't even know if it's "wá-zu" or "wa-zú". Is it wrong to pronounce it like "[Gazoo](https://en.wikipedia.org/wiki/The_Great_Gazoo)"?

astateofmind01:
I like to think about it like the old budweiser add .. waazzzzaaaaaa ;)

hex-m:
The wolf-logo seems to be gone? The latest videos in the [official YouTube channel](https://www.youtube.com/@wazuhsecurity) pronounce it something like "whasa".

PR #38729, opened 2026-08-31
Merge 5.0.0 into 5.0.1

jotacarma90:
## Description

Scheduled weekly upward merge. Part of #38696 (Week #35).

Merges branch `5.0.0` into `5.0.1`.

## Proposed Changes

- Merge `origin/5.0.0` into `5.0.1`.

### Results and Evidence

Two conflicts, both resolved with the same criteria as the previous weeks (#38391, #38528):

- `CHANGELOG.md`: kept the `5.0.1` side. Each branch's changelog keeps only its own section; the incoming `v5.0.0` entries stay in the `5.0.0` branch, reachable through the "Prior versions" links.
- `docs/ref/modules/server-api/api-reference.md`: kept the `5.0.0` side, which adds the `PUT /agents/scan/vulnerability` row and the accurate `/agents/uninstall` description — the code for both arrives with this merge.

Everything else merged automatically. Both parents verified as ancestors of the merge commit; net diff against `5.0.0` is only the `5.0.1` branch-local changes (version bumps and workflow pins).

### CI adaptations for the endpoint prefix default

The merge surfaces a latent breakage in several IT suites that had not run on `5.0.0` PRs because of CI path filtering: since an absent `<endpoint>` defaults to the `/wazuh-manager/` prefix (#38624), any suite whose agent must reach `RemotedSimulator` (which only serves unprefixed routes) fails enrollment with HTTP 404 and the agentd startup gate keeps every module from starting. The affected suites were fixed with the same opt-out stopgap the feature already applied to `test_agentd`, `test_enrollment`, `test_fim` and the Windows IT runner:

- Linux IT runner: install the agent DEB with `WAZUH_MANAGER_ENDPOINT="127.0.0.1/"` (covers sca, aws and every suite relying on the installed config).
- macOS IT runner: same change for the PKG install (covers the logcollector suite).
- syscollector and msgraph configuration templates: explicit `endpoint: '127.0.0.1:1517/'` agent section (msgraph installs from source, out of the runner-level opt-out's reach).

These opt-outs are the sanctioned interim workaround and are expected to be removed by wazuh/qa-integration-framework#835 (prefix-aware simulators, then a cleanup PR dropping every opt-out), together with the ones the feature branch introduced.

Also, the merge brings the `agent:scan_vulnerability` default RBAC policy (#38553) into `5.0.1`, so `CURRENT_ORM_VERSION` was bumped (3 → 4) as the RBAC guard requires, following the same criterion as the previous upward merge.


jotacarma90:
The failing API endpoints suites are unrelated to this merge: every current PR against 5.0.0 fails them the same way (#38616, #38724, #38736) since #38703 landed, and they have no fix issue yet. The endpoint opt-outs added here for the other IT suites are the interim workaround, removed once wazuh/qa-integration-framework#835 lands.

PR #38732, opened 2026-08-31
feat(remoted): add GET /status readiness endpoint on the admin socket

### LINKED ISSUE #38731: Expose remoted readiness on /cluster/{node_id}/status
## Description

`GET /cluster/{node_id}/status` reports `wazuh-manager-remoted` as ready whenever its process exists. `get_status()` (`framework/wazuh/manager.py`) specializes only analysisd, through the engine's `/status`, and modulesd, through the vulnerability-detector status. Every other daemon falls back to `ready = running`, which is pure liveness derived from the PID files in `var/run`.

That left no way to ask the API whether remoted can actually serve Password-mode enrollment, which is why #38605 surfaced the way it did: QA had to infer the manager's state by parsing `wazuh-manager.log`, and a transient, self-resolving startup condition failed a log-scan test.

The point of this endpoint is to remove that need. It is the signal automation waits on instead of parsing logs.

## What the endpoint checks

`GET /status` on remoted's existing manager-local admin socket (`queue/sockets/remote-admin-http.sock`), separate from `GET /` so liveness and readiness stay distinct.

remoted reports only facts it owns first-hand:

```json
{
  "ready": true,
  "enrollment_password": { "ready": true },
  "keystore": { "readable": true, "agents_loaded": 12, "entries_skipped": 0 }
}
```

**`enrollment_password` — gates `ready`.** Present only when Password-mode enrollment is enabled; omitted entirely otherwise. `ready` is true when remoted currently holds a key derived from `etc/authd.pass`. Its truth condition is exact and first-hand: *a Password-mode `POST /enroll` right now would not fail for lack of a key*. On a worker that has just joined, this is false until the master's integrity sync delivers the file, and it is deliberately **not** masked during that window — a signal that lies for the first 60 seconds is useless to wait on.

**`keystore` — informational, does not gate `ready`.** `readable` is whether the last `client.keys` reload succeeded, alongside the loaded/skipped counts. It is deliberately excluded from `ready` because remoted cannot support a readiness claim about `client.keys`:
- An empty `client.keys` and a stale one are indistinguishable from inside remoted. `authd` creates the file unconditionally at startup and starts before remoted, so on a joining worker it is readable and empty while every agent request still gets a `401` for an unknown agent.
- The keystore keeps the previous table when a reload fails or the file keeps changing across retries, and keeps authenticating agents normally. Gating on it would make the node flap out of ready while nothing is broken.

**`ready`** is the AND of the gating components. With Password mode off there are none, so it is true whenever remoted answers: remoted has nothing pending.

## What it does not check, by design

Whether this worker's `client.keys` matches the master's. That comparison exists only in the cluster layer: the worker sends per-file BLAKE2b metadata in `sync_integrity` and the master compares it. To be tracked separately if a case appears that needs it.

Note the useful side effect on a **first** join: `etc/authd.pass` and `etc/client.keys` travel in the same `cluster.json` block (`etc/: ["client.keys", "authd.pass"]`), in the same delta, so waiting for `enrollment_password.ready` on a brand-new worker does in practice wait for that whole block to land. This holds for a first join only — both files persist across restarts, and `authd.pass` never changes afterwards while `client.keys` changes on every enrollment. It must be documented as a side effect, never as a synchronization check.

## Requirements

1. **No false negatives in a healthy steady state.** Anything that waits on `ready` must converge. A signal that stays false while nothing is wrong turns a 20-second transient into a hang until timeout, which is worse than the problem being solved. Specifically: an empty `client.keys` with no agents enrolled yet, and a keystore serving a usable table after a failed reload, must both report ready.
2. **An unreachable admin socket must not mean not ready.** Failing to bring that socket up is a warning and remoted keeps running; the admin plane is documented as optional in `docs/ref/modules/remoted/README.md`. `_remoted_status()` must fall back to today's liveness behaviour and surface the reason, not report `ready: false`. This is safe during startup because `get_manager_status()` already reports `starting` while `var/run/<daemon>.start` exists, so `running` is false until the daemon is up.
3. **No grace-window masking.** The 60-second worker-join window in #38605 exists to keep an expected, self-resolving condition out of an operator's log. This endpoint has the opposite job: state what is true so a machine can wait on it. The two mechanisms must not be conflated.
4. **Document the consumer contract**: poll until `ready: true` with a bounded timeout, then run the checks. Include the observed convergence time on a joining worker (~10-20 s) so timeouts are chosen from data.

## Origin

Split out of #38605, whose scope is limited to the log severity of the `authd.pass` wait on a joining worker. That fix is what resolves the reported DTT failure: a log-scan test cannot be fixed by an endpoint, because the lines are already written by the time the test reads the file. This issue prevents the next occurrence of the same class of problem by giving automation something to query instead of a log to parse.

jotacarma90:
## Description
Closes #38731

`GET /cluster/{node_id}/status` reported `wazuh-manager-remoted` as ready whenever its process existed: `get_status()` specializes only `analysisd` and `modulesd`, and every other daemon falls back to `ready = running`, pure liveness from the PID files. There was no way to ask the API whether remoted could actually serve Password-mode enrollment, or whether its `client.keys` mirror had loaded — a caller had to infer that from `wazuh-manager.log` or from a functional `401`.

This adds a `GET /status` route on remoted's existing local admin socket, separate from the bare liveness probe at `GET /`, and wires it into `GET /cluster/{node_id}/status` under the `wazuh-manager-remoted` entry.

Split out of #38682, which keeps only the `authd.pass` log-severity fix for #38605. The commit on this branch implements the readiness-endpoint idea that grew out of that issue's discussion, revised against the finalized scope in #38731 (see below).

## Proposed Changes
- **New `GET /status`** on remoted's admin socket (`queue/sockets/remote-admin-http.sock`), `RouteClass::Liveness` (same class as `/` and `/metrics`, matching the one real precedent in this codebase for a business-logic-aware status route — VD's `/vulnerability-detector/status`). Answered inline from resident state: plain atomics plus a mutex-guarded cached key copy, no I/O and no KDF in the handler.
- **`ready` is gated solely by `enrollment_password.ready`** when Password-mode enrollment is enabled; with it disabled, `ready` is `true` whenever the handler answers at all. `enrollment_password` is present only when Password-mode enrollment is enabled, and is omitted entirely otherwise rather than reported as a distinct not-applicable state.
- **`keystore` is always present and purely informational — it never gates `ready`.** `Keystore::lastLoadOk()` (new atomic, distinguishing "`client.keys` read fine, file is empty" from "never successfully loaded") is exposed as `keystore.readable`, not `keystore.ready`, precisely so it doesn't read as a readiness claim: remoted cannot tell an empty-but-fine `client.keys` apart from a stale one still serving the old table, and gating on either would flap a healthy node in and out of `ready` for no real reason.
- **An unreachable admin socket does not report `ready: false`.** The admin plane is optional observability — remoted keeps serving agents even if it fails to bind. On the framework side, `_remoted_status()` treats a `RemotedHTTPClient` connect error (and only a connect error — not a timeout, a non-2xx response, or a malformed body, all of which imply the socket *did* answer) as `{'ready': running, 'reason': 'admin socket unreachable'}` instead of `{'ready': False}`.
- **New `RemotedHTTPClient` and `_remoted_status()`** in the framework, following the existing `_analysisd_status()`/`_modulesd_status()` pattern, embedding the result under `wazuh-manager-remoted` in `/cluster/{node_id}/status`.
- **Documented the consumer contract**: poll `/status` (or `/cluster/{node_id}/status`) until `ready: true` with a bounded timeout before proceeding; on a joining worker, `ready` typically converges within ~10-20s as the initial `cluster.json` sync lands the `etc/` file group (`client.keys`, `authd.pass`).

### Results and Evidence

C++ (`DEBUG=1 TEST=1 TARGET=server`):
```
AdminServerTest.*                                    8 passed, 0 failed
KeystoreTest.*                                       29 passed, 0 failed
remoted_module_utest (excl. one pre-existing,
  unrelated hang in ShutdownRace.*, unrelated
  to this diff)                                     783 passed, 0 failed, 100 suites
```

Python:
```
pytest framework/wazuh/tests/test_manager.py -vv     53 passed, 0 failed
pytest framework -q --ignore=tests (full suite)    2026 passed, 6 skipped, 0 failed
```

Response shape:
```json
{"ready": true, "enrollment_password": {"ready": true}, "keystore": {"readable": true, "agents_loaded": 12, "entries_skipped": 0}}
```

Self-review performed against the full diff before this push (spec/quality check, not a build/test run): no Critical, Important, or Minor findings — "Ready to merge: Yes."

### Artifacts Affected
- `src/remoted/remoted_module/src/remotedModuleFacade.hpp`
- `src/remoted/remoted_module/src/auth/keystore.{hpp,cpp}`
- `framework/wazuh/core/{engine_http,exception,common}.py`
- `framework/wazuh/manager.py`
- `api/api/spec/spec.yaml`

### Configuration Changes
None. No new parameters and no new sockets: `/status` is served on the admin socket that already exists, at its existing fixed path and mode.

### Documentation Updates
- `docs/ref/modules/remoted/README.md` — admin socket route table, response shape, gating rules, consumer-contract paragraph.
- `src/remoted/remoted_module/README.md` — internal module reference: route count and table updated from two routes to three, with the same response-shape/gating detail in this file's own technical depth.
- `docs/ref/modules/remoted/https-events-api.md` — the note stating the module's own statistics and readiness are not served on the agent-facing listener.
- `docs/ref/modules/utils/uds-http-server/README.md` — consumer table.

### Tests Introduced
**New:**
- `AdminServerTest.GetStatusReportsKeystoreAndPasswordReadiness`
- `AdminServerTest.GetStatusOmitsEnrollmentPasswordWhenPasswordModeDisabled`
- `AdminServerTest.GetStatusKeystoreFailureDoesNotAffectReadyWhenPasswordModeDisabled`
- `KeystoreTest.LastLoadOkTracksSuccessAndFailure`
- `KeystoreTest.EmptyFileIsALoadedStateNotAFailedOne`
- `test_get_status_remoted_admin_socket_unreachable` — proves the unreachable-socket fallback
- `test_get_status_remoted_timeout_still_not_ready` — proves the fallback is scoped to a connect error only, not any failure
- `test_get_status_remoted_keystore_failure_does_not_gate_ready`
- `test_get_status_remoted_password_unavailable_not_ready`

**Modified:**
- `AdminServerTest.UnknownRouteAnswers404AndWrongVerb405` extended to cover `POST /status` → `405`.
- `test_get_status_all_ready`, `test_get_status_remoted_ready`, `test_get_status_remoted_password_mode_disabled` updated to the final field names/gating.

## Review Checklist
- [x] Code changes reviewed (self-review via automated code-review pass, no findings)


MiguelazoDS:
**Fixed:**
- **R1** – real bug: `PasswordKeySource` was gated on `usePassword` alone, ignoring `enrollmentEnabled`. A manager with enrollment administratively disabled but `use_password` left at its installer default would wait forever for an `authd.pass` that will never exist, violating the "no false negatives" requirement. Now gated on both flags; `enrollment_password` is omitted whenever either is off, matching the existing omission convention. Added a regression test for this exact combination.
- **R2/R3/R4** – stale/overstated comments (route count, teardown scope, key-copy claim) in `remotedModuleFacade.hpp` and both remoted READMEs. Text-only, no behavior change.
- **R8/R9** – `adminServer_test.cpp`'s fixture truncated a relative `etc/client.keys` with no cwd isolation (real risk if `ctest` ever ran from a live install dir) and implicitly assumed no real `authd.pass` on the runner. Both traced to this PR's own first commit, not legacy. Fixed by chdir'ing into a fresh temp directory per test.
- **R11** – 7 pre-existing `test_manager.py` tests were missing the `RemotedHTTPClient` patch once `get_status()` started calling it unconditionally, so they hit a real UDS connect. Patched, mirroring the pattern already used correctly elsewhere in the same file.
- **R12** – `engine_http.py`'s docstring and a test fixture still said `keystore: {ready, ...}`; renamed to `readable` to match the real field.

**Not fixed, with reasoning:**
- **R5** – every `ConnectError` (including permission-denied) is bucketed as "admin socket unreachable." Requirement #2 doesn't distinguish a transient race from a persistent misconfiguration, so this is a defensible reading, not a violation. Leaving as-is.
- **R6** – non-connect-error failures return bare `{'ready': False}` with no reason. Requirement #2 only mandates surfacing a reason for the unreachable-socket case specifically; this already matches it. This was also a deliberate scope decision made during design, not an oversight.
- **R7** – no CHANGELOG entry: this PR carries the `no-changelog` label.
- **R10** – `Keystore`'s watcher not re-triggering `reload()` after a mid-run deletion is real, but predates this feature and the design already treats `keystore.readable` as informational-only for exactly this reason (an empty-but-fine and a stale `client.keys` are indistinguishable from inside remoted). Out of scope here.

Full C++ (`remoted_module_utest`, 788/788) and Python (`test_manager.py` + `test_engine_http.py`, 92/92) suites pass against the new commit.

PR #38751, opened 2026-08-31
stop rootcheck false-positive rootkit alerts on overlayfs directories

### LINKED ISSUE #38693: Rootcheck reports a rootkit in every directory of the agent container, and it is 96% of all findings
|---|---|---|---|---|
| 5.0.0-latest | wazuh-agent (rootcheck, system scan), `wazuh-rootcheck` integration ruleset | Agent | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 container on overlayfs |

## Description

On a freshly deployed stack with two stock agents and nothing wrong with either of them, rootcheck reports **"Wazuh Rootcheck - Rootkit or malware detected"** at level `high`, tagged MITRE T1014 Rootkit, T1564 Hide Artifacts, T1036 Masquerading, for 33 ordinary system directories including `/etc`, `/usr/bin`, `/var/log` and `/var/lib/rpm`.

Every one of them is false. The cause is that the check relies on the traditional Unix invariant that a directory's hard link count equals two plus its number of subdirectories, and overlayfs, the filesystem of every Docker container, reports `st_nlink = 1` for all directories. Wazuh ships its agent as a container image, so the product generates these against itself out of the box.

## Steps to reproduce

Deterministic on any containerised agent:

```bash
docker run -d --name a1 -e WAZUH_MANAGER_SERVER=<manager> \
    public.ecr.aws/wazuh-cicd/wazuh/wazuh-agent:5.0.0-latest

# rootcheck runs its first scan at startup; wait for "Rootcheck scan ended"
docker exec a1 stat -f -c '%T' /etc          # overlayfs
docker exec a1 stat -c '%h' /etc             # 1, on any directory
```

Then query the findings index, or open Threat Hunting:

```bash
curl -sk -u admin:admin 'https://localhost:9200/wazuh-findings-v5-*/_search?size=0' \
  -d '{"aggs":{"t":{"terms":{"field":"wazuh.rule.title","size":10}}}}'
```

Note that on a deployment where detection is off, this is invisible: the events are indexed but no finding is produced. See [036](DUPLICATED_036-no-detector-created-no-findings-ever.md). It only becomes visible once detection actually runs, which is why it had not surfaced earlier in this battery.

## Current behavior

Findings on the deployment, one hour in, with two agents and no adversary:

```
  97  Wazuh Rootcheck - Rootkit or malware detected
   1  Failed authentication attempt - root from 10.0.0.9
   1  SSH root login via password authentication - 10.0.0.7
   1  SSH root login via password authentication - 10.0.0.9
   1  Successful user authentication - root

by severity:   97 high      3 low      1 informational
```

**96% of all findings, and 100% of the high-severity ones, are this false positive.** The four real ones are log lines I injected by hand to test the detection path.

A sample, with its rule metadata:

```json
"event.original": "Files hidden inside directory '/etc'. Link count does not match number of files (37,1)."
"wazuh.rule.title": "Wazuh Rootcheck - Rootkit or malware detected"
"wazuh.rule.level": "high"
"wazuh.rule.mitre": {"technique": {"id": ["T1014", "T1564", "T1036"],
                                   "name": ["Rootkit", "Hide Artifacts", "Masquerading"]},
                     "tactic":    {"id": ["TA0005", "TA0040"], "name": ["Stealth", "Impact"]}}
"event.action": "malware-detected"
"event.category": ["malware", "file"]
```

They re-fire on every rootcheck scan, from a fresh source event each time, so the total climbs for the life of the deployment. Six minutes and one agent restart later, on the same untouched stack:

```
 171  Wazuh Rootcheck - Rootkit or malware detected      (was 97)
   4  everything else                                    (unchanged)

by severity:  171 high      3 low      1 informational
```

97.7%, and the four real findings are still the ones I injected. The scan that produced the new batch was triggered by a restart I performed, but nothing about it was special: a scheduled scan produces the same result, and the shipped `<frequency>` is 12 hours, so an untouched agent adds a batch twice a day, per agent, forever.

The ratio only moves one way, because the noise accumulates and the signal does not. Measured across 90 minutes on the same stack:

```
10:29   97 of 101   96.0%
10:35  171 of 175   97.7%
10:45  335 of 339   98.8%
```

The four real findings are the same four throughout. Every additional finding the deployment has ever produced is false.

## Expected results

A stock agent container reports no rootkits. Concretely: the link count test is skipped on filesystems that cannot maintain a meaningful directory link count, which is what `skip_file_systems` already declares and what `wazuh#561` closed in 2018.

Failing that, the finding should not be `high` severity with a rootkit MITRE mapping. "The link count of a directory did not match" is an anomaly worth noting, not evidence of a rootkit, and presenting it as one at the top severity trains users to ignore the category.

### The numbers in the message are the proof

The message prints `(expected, actual)`, where expected is two plus the subdirectory count. Measured inside the agent, each one matches the alert exactly:

```
/etc       st_nlink=1  subdirs=35  -> expected 37    alert said (37,1)
/var/lib   st_nlink=1  subdirs=8   -> expected 10    alert said (10,1)
/usr/bin   st_nlink=1  subdirs=0   -> expected  2    alert said (2,1)
/var/log   st_nlink=1  subdirs=0   -> expected  2    alert said (2,1)
```

`st_nlink` is 1 for every directory, so the check can never pass. Note `/usr/bin`, which has no subdirectories at all: even the trivial case of an empty-of-directories directory fails, because 2 != 1.

The same directories on the ext4 host behave normally, which is why this never shows up outside containers:

```
host (ext2/ext3):  /etc st_nlink=155    /usr/bin st_nlink=2    /var/log st_nlink=11
container (overlayfs): all three st_nlink=1
```

### The guard for this exists, matches, and does not fire

This is what makes it a regression rather than a gap. The filesystem reports exactly the magic number Wazuh already has a constant for:

```
$ docker exec wui-a2 stat -f -c '%t %T' /etc
794c7630  overlayfs
```

`src/shared/src/fs_op.c`, on both `main` and the `5.0.0` branch:

```c
#define OVERLAYFS   0x794c7630

/* List of filesystem to skip the link count test */
const struct file_system_type skip_file_systems[] = {
    {.name="BTRFS",     .f_type=BTRFS,     .flag=1},
    {.name="AUFS",      .f_type=AUFS,      .flag=1},
    {.name="OVERLAYFS", .f_type=OVERLAYFS, .flag=1},
    {.name="V9FS",      .f_type=V9FS,      .flag=1},
```

And `src/rootcheck/src/check_rc_sys.c` calls it immediately before the test, with a comment describing this exact symptom:

```c
/* skip further test because the FS cant deliver the stats (btrfs link count always is 1) */
skip_fs = skipFS(dir_name);
if (skip_fs != 0) {
    closedir(dp);
    return(0);            /* line 504: returns before the link count test at line 510 */
}
```

So in source the directory should be skipped. In the shipped image it is not. **Why the guard does not take effect in the build was not established**, and it is the first thing to determine. Two things worth checking: `IsNFS()` and `skipFS()` have their whole bodies inside `#if defined(Linux)` while the constants and the arrays are under `#ifdef __linux__`, so a build that defines one and not the other compiles both functions down to `return 0`; and the `-latest` image may not correspond to the `5.0.0` branch I read.

Raising rootcheck's verbosity to confirm from the running agent is not possible: 5.0.0's `internal_options.conf` exposes only `rootcheck.sleep`, so `skipFS`'s own `mdebug2("Skipping dir (FS %s)")` cannot be turned on.

### The shipped configuration already knows about containers

The agent's own `ossec.conf` ignores two container paths, so the container case was considered:

```xml
<rootcheck>
  <disabled>no</disabled>
  <check_sys>yes</check_sys>
  <frequency>43200</frequency>
  <skip_nfs>yes</skip_nfs>
  <ignore>/var/lib/containerd</ignore>
  <ignore>/var/lib/docker/overlay2</ignore>
</rootcheck>
```

Those cover a host watching containers. They do nothing for an agent running **inside** one, which is how Wazuh distributes the agent.

### Root cause, verified in source: the 2018 fix is present but compiled out

The fix from `wazuh#561` and `wazuh#22966` is still in the tree and is still called. `check_rc_sys.c` invokes it immediately before the link-count test:

```c
/* src/rootcheck/src/check_rc_sys.c:497 */
/* skip further test because the FS cant deliver the stats (btrfs link count always is 1) */
skip_fs = skipFS(dir_name);
if(skip_fs != 0)
{
    closedir(dp);
    return(0);
}

/* :509  the test that produces the alert */
if ((entry_count != (unsigned) statbuf.st_nlink) && ...) {
    snprintf(op_msg, OS_SIZE_1024, "Files hidden inside directory "
             "'%s'. Link count does not match number of files (%d,%d).", ...);
    notify_rk(ALERT_ROOTKIT_FOUND, op_msg);
```

and overlayfs is in the skip table:

```c
/* src/shared/src/fs_op.c:31 */
#define OVERLAYFS   0x794c7630
/* :55  skip_file_systems[] */
{.name="BTRFS", .f_type=BTRFS, .flag=1},
{.name="AUFS", .f_type=AUFS, .flag=1},
{.name="OVERLAYFS", .f_type=OVERLAYFS, .flag=1},
{.name="V9FS", .f_type=V9FS, .flag=1},
```

**But the function bodies are guarded by a macro that nothing defines.**

```c
/* src/shared/src/fs_op.c:99 */
short skipFS(const char *dir_name)
{
#if defined(Linux)
    struct statfs stfs;
    ...      /* the table lookup lives here */
#else
    ...
#endif
    return(0);          /* <- what the Linux agent actually compiles to */
}
```

`Linux`, capital L, is used as a guard in exactly four places and is **defined nowhere in the 5.0.0 tree**:

```
$ git grep -n "defined(Linux)\|ifdef Linux" -- src
src/shared/include/fs_op.h:21   #ifdef Linux          -> the <sys/vfs.h> include
src/shared/include/shared.h:52  #if defined(Linux) || defined(FreeBSD)
src/shared/src/fs_op.c:70       #if defined(Linux)    -> IsNFS()
src/shared/src/fs_op.c:101      #if defined(Linux)    -> skipFS()

$ git grep -n -- "-DLinux"
(no matches)
```

GCC predefines `__linux__`, `__linux`, `__gnu_linux__` and lowercase `linux`, never capital `Linux`. The old `src/Makefile` supplied it; the 5.0.0 CMake build adds `-DCLIENT` and several others but not `-DLinux`. So both `skipFS()` and `IsNFS()` compile down to `return 0`, nothing is ever skipped, and every overlayfs directory takes the link-count test. `<skip_nfs>yes</skip_nfs>` in the shipped agent configuration is inert for the same reason.

This makes it a **regression with a one-line cause**, not a re-report: both prior issues are closed as completed, and the code they added is still there.

**How to reproduce, in one command, with no Wazuh involved:**

```bash
docker run --rm alpine sh -c 'stat -c "%n nlink=%h" /etc /usr/bin /var/log'
# every directory reports nlink=1 on overlayfs, which is what the check treats as evidence of a rootkit
```

and on a stock agent container:

```bash
docker exec <agent> stat -f -c '%T' /etc          # overlayfs
docker logs <agent> 2>&1 | grep -c "Files hidden inside directory"
curl -sk -u admin:admin 'https://localhost:9200/wazuh-findings-v5-*/_search?size=0' \
  -d '{"query":{"match_phrase":{"wazuh.rule.title":"Rootkit or malware detected"}}}'
```

**How to confirm the diagnosis specifically:** a build with `-DLinux` added should skip every overlayfs directory and emit none of these findings, with no other change.

hernanvalenzuela:
## Description

Rootcheck reports false-positive "rootkit found" alerts for virtually every directory in agents running inside containers on overlayfs 96% of all findings on a freshly-deployed stack are false positives, affecting /etc, /usr/bin, /var/log, /var/lib/rpm.

Root cause: IsNFS()/skipFS() in src/shared/src/fs_op.c are guarded by #if defined(Linux) (capital L), a macro that is never defined anywhere — not by GCC (which only predefines lowercase __linux__/linux) and not by the build system (Makefile/CMakeLists.txt only branch on CMAKE_SYSTEM_NAME/uname_S, never pass -DLinux to the compiler). Both functions silently compile to return 0 on every platform, so skipFS() never skips the link-count heuristic in rootcheck/src/check_rc_sys.c for filesystems where st_nlink isn't 2 + subdirectory_count (overlayfs always reports st_nlink = 1), triggering the false "Files hidden inside directory" alert on nearly every directory.

## Proposed Changes

src/shared/src/fs_op.c: fix the dead-code guard in IsNFS() and skipFS() — #if defined(Linux) → #if defined(__linux__).
src/shared/include/fs_op.h: same bug, same root cause, found while reviewing the header — the <sys/vfs.h> include was guarded by #ifdef Linux and <sys/mount.h> by #ifdef FreeBSD (should be __FreeBSD__); corrected both for consistency.
src/unit_tests/shared/test_fs_op.c / CMakeLists.txt: added __wrap_statfs() and 5 regression tests (overlayfs skip, no-skip on a normal filesystem, statfs() error path, NFS detection) — the existing suite mocked around these two functions without ever exercising them, which is how the bug shipped unnoticed.

### Results and Evidence

Root cause confirmed at the compiler level and reproduced against a real overlayfs mount (this WSL2 host already mounts `/usr/lib/wsl/lib` and `/mnt/wslg/doc` via overlay):

- `gcc -dM -E -x c /dev/null | grep linux` → GCC only ever predefines `__linux__`, `linux`, `__linux`; never `Linux`.
- `#if defined(Linux)` preprocesses to `false` with a stock gcc, no flags needed — confirmed empirically.
- Standalone repro compiling the literal `skipFS()` body with both guards, run against real filesystems:

| path | real `f_type` | old guard (`Linux`) | fixed guard (`__linux__`) |
|---|---|---|---|
| `/usr/lib/wsl/lib` (overlayfs) | `0x794c7630` (OVERLAYFS) | `skipFS()=0` ❌ | `skipFS()=1` ✅ |
| `/mnt/wslg/doc` (overlayfs) | `0x794c7630` (OVERLAYFS) | `skipFS()=0` ❌ | `skipFS()=1` ✅ |
| `/etc` (ext4, control) | `0xef53` | `skipFS()=0` (correct) | `skipFS()=0` (correct — fix doesn't over-skip) |

This ties directly to the reported symptom: with the dead guard, `check_rc_sys.c`'s `st_nlink` heuristic (which assumes `nlink == 2 + subdirs`, invalid on overlayfs) runs unconditionally and fires `ALERT_ROOTKIT_FOUND` on any directory with subdirectories — matching the reported 96% false-positive rate across `/etc`, `/usr/bin`, `/var/log`, `/var/lib/rpm`.

  - [x] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

### Tests Introduced

Added `__wrap_statfs()` to `src/unit_tests/shared/test_fs_op.c` (mocks the `statfs()` syscall,same pattern as `__wrap_stat` in `stat_wrappers.c`) plus 5 regression tests exercising the real `IsNFS()`/`skipFS()` implementations for the first time:

- `test_skipFS_overlayfs_is_skipped`: overlayfs (`f_type=0x794c7630`) must be skipped — the core
  regression case; returned `0` unconditionally before the fix.
- `test_skipFS_regular_fs_is_not_skipped`: a normal filesystem (ext4) must not be skipped.
- `test_skipFS_statfs_error`: a failing `statfs()` call must propagate `-1`.
- `test_IsNFS_nfs_mount_is_detected`: an NFS mount (`f_type=0x6969`) must be detected.
- `test_IsNFS_regular_fs_is_not_detected`: a normal filesystem must not be flagged as NFS.

The previous test file only exercised `compare()` against hand-built structs and never called `IsNFS()`/`skipFS()` themselves, which is how the dead `#if defined(Linux)` branch shipped unnoticed.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

### DIFF
--- a/src/unit_tests/shared/test_fs_op.c
+++ b/src/unit_tests/shared/test_fs_op.c
@@ -14,9 +14,27 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <errno.h>
+#ifdef __linux__
+#include <sys/vfs.h>
+#endif
 
 #include "shared.h"
 #include "../wrappers/common.h"
+
+// Wrappers
+
+int __wrap_statfs(const char * path, struct statfs * buf) {
+    check_expected(path);
+
+    long f_type = mock_type(long);
+
+    if (f_type != -1) {
+        buf->f_type = f_type;
+    }
+
+    return mock_type(int);
+}
 
 // Tests
 
@@ -54,9 +72,74 @@ void test_fs_magic(void **state)
     assert_int_equal(compare(&statfs), 1);
 }
 
+// Regression tests for https://github.com/wazuh/wazuh/issues/38693:
+// skipFS()/IsNFS() must actually evaluate the "#if defined(__linux__)" branch
+// instead of silently short-circuiting to 0.
+
+void test_skipFS_overlayfs_is_skipped(void **state)
+{
+    (void) state;
+
+    expect_string(__wrap_statfs, path, "/etc");
+    will_return(__wrap_statfs, 0x794c7630 /* OVERLAYFS, mirrors test_fs_magic above */);
+    will_return(__wrap_statfs, 0);
+
+    assert_int_equal(skipFS("/etc"), 1);
+}
+
+void test_skipFS_regular_fs_is_not_skipped(void **state)
+{
+    (void) state;
+
+    expect_string(__wrap_statfs, path, "/etc");
+    will_return(__wrap_statfs, 0xEF53 /* EXT4_SUPER_MAGIC */);
+    will_return(__wrap_statfs, 0);
+
+    assert_int_equal(skipFS("/etc"), 0);
+}
+
+void test_skipFS_statfs_error(void **state)
+{
+    (void) state;
+
+    expect_string(__wrap_statfs, path, "/nonexistent");
+    will_return(__wrap_statfs, -1);
+    errno = ENOENT;
+    will_return(__wrap_statfs, -1);
+
+    assert_int_equal(skipFS("/nonexistent"), -1);
+}
+
+void test_IsNFS_nfs_mount_is_detected(void **state)
+{
+    (void) state;
+
+    expect_string(__wrap_statfs, path, "/mnt/nfs");
+    will_return(__wrap_statfs, 0x6969 /* NFS */);
+    will_return(__wrap_statfs, 0);
+
+    assert_int_equal(IsNFS("/mnt/nfs"), 1);
+}
+
+void test_IsNFS_regular_fs_is_not_detected(void **state)
+{
+    (void) state;
+
+    expect_string(__wrap_statfs, path, "/etc");
+    will_return(__wrap_statfs, 0xEF53 /* EXT4_SUPER_MAGIC */);
+    will_return(__wrap_statfs, 0);
+
+    assert_int_equal(IsNFS("/etc"), 0);
+}
+
 int main(void) {
     const struct CMUnitTest tests[] = {
             cmocka_unit_test(test_fs_magic),
+            cmocka_unit_test(test_skipFS_overlayfs_is_skipped),
+            cmocka_unit_test(test_skipFS_regular_fs_is_not_skipped),
+            cmocka_unit_test(test_skipFS_statfs_error),
+            cmocka_unit_test(test_IsNFS_nfs_mount_is_detected),
+            cmocka_unit_test(test_IsNFS_regular_fs_is_not_detected),
     };
     return cmocka_run_group_tests(tests, NULL, NULL);
 }
--- a/src/unit_tests/shared/test_fs_op.c
+++ b/src/unit_tests/shared/test_fs_op.c
@@ -25,6 +25,7 @@
 
 // Wrappers
 
+#ifdef __linux__
 int __wrap_statfs(const char * path, struct statfs * buf) {
     check_expected(path);
 
@@ -36,6 +37,7 @@ int __wrap_statfs(const char * path, struct statfs * buf) {
 
     return mock_type(int);
 }
+#endif
 
 // Tests
 
@@ -75,7 +77,9 @@ void test_fs_magic(void **state)
 
 // Regression tests for https://github.com/wazuh/wazuh/issues/38693:
 // skipFS()/IsNFS() must actually evaluate the "#if defined(__linux__)" branch
-// instead of silently short-circuiting to 0.
+// instead of silently short-circuiting to 0. __wrap_statfs() above only exists
+// on Linux (struct statfs itself is only declared there), so these are too.
+#ifdef __linux__
 
 void test_skipFS_overlayfs_is_skipped(void **state)
 {
@@ -134,14 +138,18 @@ void test_IsNFS_regular_fs_is_not_detected(void **state)
     assert_int_equal(IsNFS("/etc"), 0);
 }
 
+#endif // __linux__
+
 int main(void) {
     const struct CMUnitTest tests[] = {
             cmocka_unit_test(test_fs_magic),
+#ifdef __linux__
             cmocka_unit_test(test_skipFS_overlayfs_is_skipped),
             cmocka_unit_test(test_skipFS_regular_fs_is_not_skipped),
             cmocka_unit_test(test_skipFS_statfs_error),
             cmocka_unit_test(test_IsNFS_nfs_mount_is_detected),
             cmocka_unit_test(test_IsNFS_regular_fs_is_not_detected),
+#endif // __linux__
     };
     return cmocka_run_group_tests(tests, NULL, NULL);
 }
--- a/src/unit_tests/shared/test_fs_op.c
+++ b/src/unit_tests/shared/test_fs_op.c
@@ -21,6 +21,7 @@
 
 #include "shared.h"
 #include "../wrappers/common.h"
+#include "../wrappers/wazuh/shared/debug_op_wrappers.h"
 
 // Wrappers
 
@@ -83,6 +84,7 @@ void test_skipFS_overlayfs_is_skipped(void **state)
     expect_string(__wrap_statfs, path, "/etc");
     will_return(__wrap_statfs, 0x794c7630 /* OVERLAYFS, mirrors test_fs_magic above */);
     will_return(__wrap_statfs, 0);
+    expect_string(__wrap__mdebug2, formatted_msg, "Skipping dir (FS OVERLAYFS): /etc ");
 
     assert_int_equal(skipFS("/etc"), 1);
 }
--- a/src/shared/include/fs_op.h
+++ b/src/shared/include/fs_op.h
@@ -18,11 +18,11 @@
 
 #ifndef WIN32
 
-#ifdef Linux
+#ifdef __linux__
 #include <sys/vfs.h>
 #endif
 
-#ifdef FreeBSD
+#ifdef __FreeBSD__
 #include <sys/param.h>
 #include <sys/mount.h>
 #endif
--- a/src/shared/src/fs_op.c
+++ b/src/shared/src/fs_op.c
@@ -67,7 +67,7 @@ const struct file_system_type skip_file_systems[] = {
 
 short IsNFS(const char *dir_name)
 {
-#if defined(Linux)
+#if defined(__linux__)
     struct statfs stfs;
 
     /* ignore NFS (0x6969) or CIFS (0xFF534D42) mounts */
@@ -98,7 +98,7 @@ short IsNFS(const char *dir_name)
 
 short skipFS(const char *dir_name)
 {
-#if defined(Linux)
+#if defined(__linux__)
     struct statfs stfs;
 
     if ( ! statfs(dir_name, &stfs) )
--- a/src/unit_tests/shared/CMakeLists.txt
+++ b/src/unit_tests/shared/CMakeLists.txt
@@ -262,7 +262,7 @@ list(APPEND shared_tests_flags "-Wl,--wrap=OS_StrIsNum,--wrap=_merror,--wrap=w_t
                                 -Wl,--wrap,OSRegex_Execute_ex")
 
 list(APPEND shared_tests_names "test_fs_op")
-list(APPEND shared_tests_flags " ")
+list(APPEND shared_tests_flags "-Wl,--wrap=statfs ${DEBUG_OP_WRAPPERS}")
 
 list(APPEND shared_tests_names "test_wazuhdb_op")
 list(APPEND shared_tests_flags "-Wl,--wrap,OS_ConnectUnixDomain -Wl,--wrap,OS_SendSecureTCP -Wl,--wrap,OS_RecvSecureTCP")
--- a/src/unit_tests/shared/CMakeLists.txt
+++ b/src/unit_tests/shared/CMakeLists.txt
@@ -262,7 +262,11 @@ list(APPEND shared_tests_flags "-Wl,--wrap=OS_StrIsNum,--wrap=_merror,--wrap=w_t
                                 -Wl,--wrap,OSRegex_Execute_ex")
 
 list(APPEND shared_tests_names "test_fs_op")
+if(${TARGET} STREQUAL "winagent")
+list(APPEND shared_tests_flags "${DEBUG_OP_WRAPPERS}")
+else()
 list(APPEND shared_tests_flags "-Wl,--wrap=statfs ${DEBUG_OP_WRAPPERS}")
+endif()
 
 list(APPEND shared_tests_names "test_wazuhdb_op")
 list(APPEND shared_tests_flags "-Wl,--wrap,OS_ConnectUnixDomain -Wl,--wrap,OS_SendSecureTCP -Wl,--wrap,OS_RecvSecureTCP")


hernanvalenzuela:
## Root Cause

`IsNFS()` and `skipFS()` in `src/shared/src/fs_op.c` are guarded by `#if defined(Linux)` (capital L) — a macro that is **never defined**, neither by GCC nor by the build system:

```
$ gcc -dM -E -x c /dev/null | grep -iw 'linux\|__linux__\|__linux'
#define __linux 1
#define linux 1
#define __linux__ 1
```

GCC only ever predefines the lowercase variants. `grep`ing the entire `Makefile`/`CMakeLists.txt` tree confirms no `-DLinux` is passed to the compiler anywhere today. As a result, both functions silently compile to `return 0` on every platform — `skipFS()` never skips overlayfs/btrfs/aufs/v9fs, and `IsNFS()` never detects NFS/CIFS. No compiler warning is possible here (`#if defined(X)` on an undefined `X` is valid, silent, false).

The same dead-macro pattern was found and fixed in `src/shared/include/fs_op.h`: the `<sys/vfs.h>` include was gated by `#ifdef Linux` and `<sys/mount.h>` by `#ifdef FreeBSD` (should be `__FreeBSD__`).

## The Regression

This is not "always been broken" — it has a precise origin. Commit [`b48e021b93`](https://github.com/wazuh/wazuh/commit/b48e021b93ec62eea5352bd882046baa317db490) ("feat: migrate win32 to cmake", 2026-02-13), while migrating the Windows build to CMake, also stripped the non-Windows platform-define plumbing from `src/Makefile`:

```diff
 ifneq (${TARGET},winagent)
-		DEFINES+=-D${uname_S}
 ifeq (${uname_S},Linux)
 		PRECOMPILED_OS:=linux
-		DEFINES+=-DINOTIFY_ENABLED -D_XOPEN_SOURCE=600 -D_GNU_SOURCE
+		OSSEC_CFLAGS+=-D_XOPEN_SOURCE=600 -D_GNU_SOURCE
...
-OSSEC_CFLAGS+=${DEFINES}
```

`DEFINES+=-D${uname_S}` used to expand to `-DLinux` (`uname_S` = "Linux" on Linux); the same commit also dropped `-DFreeBSD`/`-DDarwin`/`-DNetBSD`/`-DOpenBSD` and removed the only line that consumed `${DEFINES}` at all. Confirmed against a `compile_commands.json` snapshot from Feb 2021 (accidentally committed on an unrelated branch): at that time `shared/fs_op.c` genuinely compiled with `-DLinux -DINOTIFY_ENABLED` among its real flags.

An independent, still-open PR (`fix/38543`, Aug 26 2026, different author) re-adds `-DLinux -DINOTIFY_ENABLED` but scoped only to `syscheckd`'s own `CMakeLists.txt` — the same regression, hitting a different module, fixed independently and partially.

## Reproduction

### Static / compiler-level

`#if defined(Linux)` preprocesses to `false` with a stock `gcc`, no flags needed:

```
$ gcc -E -x c - <<'EOF'
#if defined(Linux)
OLD_GUARD_TRUE
#else
OLD_GUARD_FALSE
#endif
EOF
OLD_GUARD_FALSE
```

### skipFS() against real overlayfs

The literal body of `skipFS()` from `fs_op.c`, compiled twice (`-DGUARD_OLD` / `-DGUARD_FIXED`), run against real overlayfs mounts already present on a WSL2 host:

| path | real `f_type` | old guard (`Linux`) | fixed guard (`__linux__`) |
|---|---|---|---|
| `/usr/lib/wsl/lib` | `0x794c7630` (OVERLAYFS) | `skipFS()=0` ❌ | `skipFS()=1` ✅ |
| `/mnt/wslg/doc` | `0x794c7630` (OVERLAYFS) | `skipFS()=0` ❌ | `skipFS()=1` ✅ |
| `/usr/lib/modules/6.6.87.2-microsoft-standard-WSL2` | `0x794c7630` (OVERLAYFS) | `skipFS()=0` ❌ | `skipFS()=1` ✅ |
| `/etc` (ext4, control) | `0xef53` | `skipFS()=0` (correct) | `skipFS()=0` (correct — fix doesn't over-skip) |

The last directory also gives a real quantitative match for `check_rc_sys.c`'s trigger condition (`entry_count != st_nlink`): `subdirs=2, nlink=1, entry_count=17`.

### Live, end-to-end (the issue's own STR)

Ran the exact reproduction from the issue against the real, unpatched, publicly published agent image, enrolled to a real Wazuh 5.0.0 manager + indexer:

```bash
docker run -d --name a1 \
  -e WAZUH_MANAGER_ENDPOINT="<manager>:1517/wazuh-manager/" \
  -e WAZUH_REGISTRATION_PASSWORD="<redacted>" \
  public.ecr.aws/wazuh-cicd/wazuh/wazuh-agent:5.0.0-latest

docker exec a1 stat -f -c '%T' /etc   # overlayfs
docker exec a1 stat -c '%h' /etc      # 1
```

One rootcheck scan cycle (~9s) on a freshly-enrolled container produced:

```bash
curl -sk -u admin:admin 'https://<indexer>:9200/wazuh-findings-v5-*/_search?size=0' \
  -d '{"aggs":{"t":{"terms":{"field":"wazuh.rule.title","size":10}}}}'
```
```json
"buckets": [
  { "key": "Wazuh Rootcheck - Rootkit or malware detected", "doc_count": 33 }
]
```

**33/33 findings** were rootkit false positives, all `severity: high`. Extracted `event.original` messages, grouped by directory:

| Directory | entry_count | st_nlink |
|---|---|---|
| `/etc` | 37 | 1 |
| `/usr/local` | 12 | 1 |
| `/var/lib` | 10 | 1 |
| `/usr/lib` | 14 | 1 |
| `/usr/bin` | 2 | 1 |
| `/usr/sbin` | 2 | 1 |
| `/var/log` | 2 | 1 |
| `/var/lib/rpm` | 2 | 1 |
| `/var/lib/dnf` | 2 | 1 |
| `/tmp` | 2 | 1 |
| `/var/tmp` | 2 | 1 |
| `/usr/local/bin` | 2 | 1 |
| `/usr/lib/.build-id` | 188 | 1 |
| `/usr/lib/.build-id/*` (20 subdirs) | 2 (each) | 1 (each) |

Each row is a literal `snprintf`-formatted `event.original` string from `check_rc_sys.c:517-520`, e.g.:

```
Files hidden inside directory '/etc'. Link count does not match number of files (37,1).
```

— generated by the real production binary, not simulated.

## Fix Verification

`check_rc_sys.c:499-504` returns unconditionally, before the nlink-mismatch check, whenever `skipFS()` returns non-zero:

```c
skip_fs = skipFS(dir_name);
if(skip_fs != 0)
{
    closedir(dp);
    return(0);
}
```

Since the fix makes `skipFS()` return `1` for every overlayfs directory that produced a false positive above (verified against the real, unpatched binary's own filesystem type detection), every one of those 33 alerts is eliminated by construction — the nlink comparison that produces `ALERT_ROOTKIT_FOUND` is never reached.

## Conclusion

Root cause confirmed at the compiler level, traced to a specific regressing commit, and reproduced end-to-end against the real unpatched agent image with a real manager and indexer — 33/33 findings on a fresh container are false-positive rootkit alerts, matching the issue's reported 96-98% rate and exact affected directories. The fix (`#if defined(Linux)` → `#if defined(__linux__)`, plus the same correction in `fs_op.h`) closes the gap by construction, verified against the real compiled function on the same real overlayfs mounts that produced the false positives above.

src/unit_tests/shared/test_fs_op.c
```diff
@@ -14,9 +14,28 @@
...
+#include <sys/vfs.h>
+#endif
 
 #include "shared.h"
 #include "../wrappers/common.h"
+#include "../wrappers/wazuh/shared/debug_op_wrappers.h"
+
+// Wrappers
+
+int __wrap_statfs(const char * path, struct statfs * buf) {
```
lchico, line 29: `__wrap_statfs(const char *path, struct statfs *buf)` is defined unconditionally, but `struct statfs` is only declared through the `#ifdef __linux__`-guarded `#include <sys/vfs.h>` a few lines above. `src/unit_tests/CMakeLists.txt` adds this `shared` unit-test subdirectory for both the Linux and Windows (`CMAKE_SYSTEM_NAME`) builds, and the winagent build cross-compiles with mingw32 where `__linux__` is undefined and there's no `<sys/vfs.h>`/`struct statfs`. That means this file fails to compile with `unknown type name 'struct statfs'` on the winagent unit-test target. Could this wrapper (and the `<sys/vfs.h>` include above it) be guarded the same way, or the whole file be excluded from the Windows unit-test build?

lchico (changes requested):

cborla:
Worth calling out, not a blocker: `IsNFS()` shares the same `#if defined(Linux)` root cause this PR fixes in `skipFS()`, and is also called from `rootcheck/src/common_rcl.c:457` to decide whether an RKCL directory policy check runs on a given path. With the dead guard, `IsNFS()` always returned 0, so `<skip_nfs>` never actually skipped anything on any deployment. `<skip_nfs>yes</skip_nfs>` is the default in all 20 packaged `rootcheck.agent.template` files across every supported distro, so once this lands, any genuinely NFS-mounted directory will start being skipped by RKCL directory checks — which is exactly what `skip_nfs=yes` has always meant to do, not new behavior (it worked this way pre-CMake-migration, before `-DLinux` was dropped).

Might be worth a line under "Configuration Changes" noting this, so it isn't mistaken for a new regression if someone notices rootcheck coverage change on an NFS mount after upgrading.

src/unit_tests/shared/test_fs_op.c
hernanvalenzuela: done

cborla (approved):
LGTM

lchico (approved):
LGTM!

 - [x] Code changes reviewed

PR #38757, opened 2026-09-01
Add durable manager-side tasks to the Task Manager and retire monitord

### LINKED ISSUE #38589: Extend the Task Manager to enqueue and execute manager-side tasks
# Summary

The Task Manager currently stores work **addressed to agents**: producers call `create_task`, `remoted` polls `get_pending_tasks`, and a row is marked `delivered` at read time — before delivery is attempted.

There is status tracking, but no **outcome tracking**.

This epic introduces **manager tasks**: durable, retried, result-tracked execution of work performed *by the manager itself*.

Consumers are:

* Agent deletion's indexer step
* On-demand VD scans
* `wazuh-monitord`'s periodic work, allowing that daemon to be retired

The existing agent-task path is **not modified**:

* `TASKS`
* `create_task`
* `get_pending_tasks`
* `remoted`'s delivery path

The rationale, verified code references, and complete test plan are documented in the design document (Task 1 below).

This issue defines the scope and the decisions that should not be revisited during implementation.

---

# Motivation

| Workload                     | Current implementation                                          | Failure mode                                                                                                                                                                                                                                                                                          |
| ---------------------------- | --------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Agent deletion, indexer step | `authd`'s purge queue (`d43311dff6`) — durable, unbounded retry | The record is written **after** the point of no return. `client.keys` is persisted at `main-server.c:1409`, while `purge_queue_push` runs at `:1494`. A crash between them leaves the agent deleted with no record that an indexer purge was owed. During bulk deletion, this window spans the batch. |
| On-demand VD scans           | 64-slot in-memory deque (`scanDispatcher.hpp`)                  | Work is lost on shutdown. The `200` response has already been sent at admission, so the request appears accepted even though the scan disappears.                                                                                                                                                     |
| `wazuh-monitord`             | 1-second poll loop with in-memory state                         | Counters reset on restart, delaying the next sweep by a full interval. The pending-disconnection hash is lost entirely.                                                                                                                                                                               |

`d43311dff6` is already good and **must not be regressed**. It provides:

* Unbounded retry
* Atomic-rename persistence
* Crash replay
* A 65,536-entry cap
* An ID-reuse interlock

The one thing `authd` cannot provide by itself is recording intent **before** `client.keys` is written. That gap is the only reason agent deletion is in scope.

---

# What Gets Built

## Persistent storage

Add two tables to `tasks.db`:

* `MANAGER_TASKS` — one row per task run
* `MANAGER_TASK_SCHEDULES` — runtime schedule state

Also add:

* Five indexes
* A `CHECK` constraint on `STATUS`

The existing `TASKS` table is untouched.

No migration and no `user_version` bump are required because 5.0 managers are fresh installs.

## Dispatcher

Add a dispatcher inside the Task Manager.

Lane threads:

1. Claim rows
2. Call the owning module through its existing socket
3. Record the outcome
4. Apply retry/defer accounting

A handler registry stores routes as **data**, so adding a module-targeted task type becomes a descriptor change rather than a dispatcher patch.

A single scheduler thread:

* Polls for claimable work
* Spawns scheduled runs
* Runs the ownership sweep
* Signals the size-rotation check

### Task types

| Task type                | Handler                                           | Lane     | Depth | Connect / request |      Attempts |         Defer |
| ------------------------ | ------------------------------------------------- | -------- | ----: | ----------------- | ------------: | ------------: |
| `agent_delete_indexer`   | `POST /_internal/agents/delete` on inventory-sync | `delete` |     4 | 2 s / 600 s       | **unbounded** | **unbounded** |
| `vd_scan`                | `POST /_internal/vd/scan` on inventory-sync       | `scan`   |     1 | 2 s / 300 s       |             8 |            48 |
| `agent_disconnect_sweep` | `wdb_disconnect_agents()`                         | `local`  |     1 | self-bounded      |             8 |            48 |
| `agent_delete_old`       | wazuh-db query + `auth_remove_agent()`            | `local`  |     1 | self-bounded      |             8 |            48 |
| `log_rotate_daily`       | daily branch of `monitor_logs()`                  | `local`  |     1 | self-bounded      |             8 |            48 |

Seven new threads are introduced: six lane workers plus the scheduler. This takes the module from 11 to 18 threads.

The three low-cadence task types share the `local` lane. They are claimed one type at a time in rotation.

The claim predicate remains:

```sql
TASK_TYPE = ?
```

rather than using `IN (...)`, because the latter would give up the index seek.

Size-based log rotation is **not a task**. It runs as a direct action on the local lane, signalled every 60 seconds. Creating 1,440 rows per day for two `w_stat` calls provides no value.

---

# Decisions Worth Not Re-Litigating

## State machine

`STATUS` transitions between `pending` and `claimed`, with the following terminal states:

* `completed`
* `failed`
* `dead_letter`
* `superseded`

The state machine is **not forward-only**. Retry and ownership recovery both return rows to `pending`.

## Atomic operations

**Everything that must be atomic is expressed as one wazuh-db command.**

The per-DB mutex spans the entire command, making the following exact:

* Claim
* Create-with-dedup
* Re-queue
* 64-slot scan admission bound

Do not use `RETURNING`. It requires SQLite ≥ 3.35, which the precompiled build branch cannot guarantee.

## Commit policy

**Only claim and create commit inside their commands.**

`ok` from wazuh-db does not imply durability: writes can remain in a deferred transaction for up to 60 seconds.

The terminal outcome deliberately does **not** commit.

If the outcome write is rolled back:

1. The row remains `claimed`
2. The ownership sweep reclaims it
3. The handler runs again
4. Mandatory handler idempotency absorbs the duplicate execution

This reduces the fsync budget to one per task — approximately 20,000 fsyncs for a 20,000-agent deletion wave.

## Ownership sweep

The sweep may reclaim a row owned by a live lane only when **both** conditions hold:

1. The lane's published in-flight `task_id` is not this row
2. `CLAIM_TIME` is older than the 30-second claim grace period

The grace period is required because a lane cannot publish its `task_id` until the claim operation returns it. Without the grace period, the sweep could reclaim a row immediately before the lane begins executing it.

## SQLite durability

Use:

* WAL on **every open**
* `synchronous=FULL`

Do not inherit global's `NORMAL` setting. `NORMAL` trades host-crash durability away from the database whose purpose is durability.

WAL must be enabled on every open; otherwise a 5.0.0 database can remain in rollback-journal mode indefinitely after upgrade.

## Retry and deferral

Retry is a **time budget**:

* Initial retry delay: 30 s
* Exponential backoff
* Maximum delay: 900 s
* 8 attempts ≈ 45 minutes

Deferral (`not_ready`, `busy`) has its own ladder:

* Initial delay: 5 s
* Maximum: 48 deferrals
* ≈ 12 hours

Deferral eventually becomes `dead_letter` for bounded task types.

Unbounded deferral would wedge the queue: coalescing would fold every new request into a stuck row, causing admission to return `503 scan_queue_full` indefinitely without anything appearing in `dead_letter`.

## Agent deletion is never terminal

`agent_delete_indexer` has:

* Unbounded attempts
* Unbounded deferral
* `allow_terminal_failure = false`

A 4xx must not transition this task to `failed` or `dead_letter`, because that would orphan documents just as permanently.

This is safe because inflow is bounded at the request path.

## Superseding

`superseded` is **not a handler result**.

It is determined during re-queue:

* If a `pending` row already occupies the same `(AGENT_ID, TASK_TYPE)` slot, the row being re-queued loses.
* The survivor inherits `ATTEMPTS` and `DEFER_COUNT`.

This prevents a coalescing task from resetting its counters under load. Without counter inheritance, every timed-out scan could be superseded by a fresh row starting at zero and never reach `dead_letter`.

## Incomplete work

`incomplete` is a sixth handler result.

It returns the task to `pending` immediately while leaving `ATTEMPTS` unchanged.

`agent_delete_old` is currently the only handler that uses it. Its work is bounded by:

* Maximum agent count: 200
* Maximum elapsed time: 30 s

The time limit protects the shared local lane.

## Idempotency

**Every handler must be idempotent.**

The outcome write can itself fail, leaving completed work reclaimable and therefore executable again.

## Thread lifecycle

Lanes must be joinable and joined by the module's main thread within the shared 30-second shutdown budget.

`w_create_thread` detaches unconditionally, so use `CreateThreadJoinable`.

## Metrics

The wazuh-db side is nearly free.

The task actor already has plain-C counters such as `w_inc_task_create()` and its siblings (`wdb_state.c:428+`). Extend that family to include:

* Create
* Claim
* Re-queue
* Outcome

The modulesd-side metrics — queue depth and lane occupancy — are blocked on the C++ facility lacking a C ABI shim. That is a named follow-up.

## Dead-letter visibility

A dead-letter record must be **both recorded and findable**.

The log line must include `task_id`, and `get_manager_tasks_by_type` must support a status filter.

Otherwise the record exists but operators have no reliable way to find it.

---

# Consumer Changes

## Agent deletion

Use a **two-phase mark** in `authd`'s writer to close the crash window.

### Phase 0 — Admission

Refuse or admit the deletion **before `OS_DeleteKey`**.

Refusal cannot happen later: by writer time, the caller has already been told that deletion succeeded.

Admission checks:

```text
journal_len + |purge_reserved|
```

against the journal cap, **or**:

```text
manager_task_max_pending_deletes
```

The journal bounds admission-in-flight rather than row count.

The cached count is atomic and **fails open**, so a wazuh-db outage cannot block deletion.

### Phase 1 — Journal

Write the agent IDs to `pending-purges` before `OS_WriteKeys`.

This is a local write, so wazuh-db can never block enrollment.

This is also where `purge_unreserve_id_locked` moves.

### Phase 2 — Persist `client.keys`

Call:

```text
OS_WriteKeys
```

and capture its return value.

### Phase 3 — Create manager tasks

Only if Phase 2 succeeds, create the manager-task rows through `wdbc_query_ex_timeout`.

The current code logs and falls through on failure, so this gate is load-bearing.

Create commits in-command, making `ok` the durability acknowledgement.

### Phase 4 — Remove journal entries

Remove the journal lines only after `ok`.

If the manager crashes between task creation and journal cleanup, startup recreates the row. The deterministic ID collision is the idempotent path working:

* Reported as `collided`
* Treated as success

### Startup reconciliation

On startup, reconcile leftover journal entries against `client.keys`:

* Agent still present → drop the journal line
* Agent gone → create the manager task

Retire `authd`'s relay thread and retry ladder.

Keep:

* `purge_delay`
* `purge_last_id`
* ID-counter raise
* `purge_reserved`

Legacy journal lines are numbered **by position in the file**, so a crash during conversion deterministically derives the same IDs.

### Completion semantics

The execution route must answer **at completion**.

`DELETE /agents` currently answers at admission (`deleteAgentEndpoint.hpp:34-40`). Therefore, a `completed` state against that route would mean "accepted", not "purged" — the same defect as `delivered`.

The new route:

```text
POST /_internal/agents/delete
```

enqueues the same `Item` **with its responder**, which `SyncPipeline` already supports for sessions.

`DELETE /agents` and its POST alias are removed with the relay thread (PR 4b), not alongside the new route.

The relay is their only production caller. Removing them first would create a release in which agent deletion stops purging documents.

### Breaking API change

Phase 0 refusal requires a new `authd` error code and corresponding server-API code.

Plumb it through:

* `exception.py`
* `agent.py`
* `spec.yaml`
* Changelog

`agent_delete_old` must map this error to `retryable`; otherwise a backlogged manager can silently stop aging out agents.

---

## VD scans — compatible with #38553

Admission at:

```text
POST /vulnerability-detector/scan
```

keeps its existing:

* Validation
* Readiness preflight
* 64-slot bound
* `503 scan_queue_full`

Instead of enqueueing in memory, admission creates a manager task.

**Producers are untouched.**

The task is created through:

```text
queue/tasks/task
```

A direct in-process call is not possible because VD is `dlopen`'d with a two-symbol ABI and modulesd has no `-rdynamic`.

The call has a 2-second timeout mapped to:

```text
503 scan_queue_full
```

This prevents a wedged wazuh-db from stalling VD I/O threads.

The timeout can fire **after** the insert commits, so a `503` does not prove that the scan was rejected.

### Execution route

The execution route lives in **inventory-sync**, not VD.

`VdScanLane` already provides:

* Bounded queue
* Per-agent registry exclusion
* Responder-to-completion
* 503-on-capacity

A VD-side route would have to rebuild all of this, while `AgentInFlightRegistry` is unreachable from VD.

A second, uncoordinated implementation would therefore introduce a race rather than merely duplicate code.

The required pipeline changes are:

* Third `Item::Kind`
* `workerLoop` branch
* Session-less `IVdScanner::scanAgent(agentId)`

The registry provides the:

```text
409 scan_in_progress → busy
```

interlock, making a client timeout safe.

Delete timeout (600 s) must exceed scan timeout (300 s). Assert this at startup so a parked deletion cannot time out behind an unrelated scan.

---

# `wazuh-monitord` Retirement

| Responsibility                | New implementation                                      |
| ----------------------------- | ------------------------------------------------------- |
| Disconnection sweep + logging | `agent_disconnect_sweep`, master-scoped                 |
| `delete_old_agents`           | `agent_delete_old`, master-scoped                       |
| Daily log rotation            | `log_rotate_daily`                                      |
| Size-based log rotation       | Direct action on the local lane every 60 s — not a task |
| `getconfig` socket            | Removed — breaking API change                           |

Split `monitor_logs()` into two paths.

This:

* Removes `sleep(day_wait)` entirely; it becomes the daily slot offset.
* Prevents the daily branch from suspending size checks for up to 600 seconds.
* Seeds the day-change baseline from the active log's mtime.
* Fixes the existing bug where a manager restarted daily never performs its daily rotation.

Move the disconnection sweep to a master-scoped schedule.

This also fixes an existing bug: `check_disconnection_trigger` ignores `mond.monitor_agents`, so workers currently run the sweep even though `main.c` sets that flag to `0` on workers.

---

# Non-Goals

* No cross-node coordination. `tasks.db` is node-local, and deterministic IDs do not provide distributed exactly-once semantics.
* No REST API surface. `/tasks/status` is not reinstated.
* No changes to agent tasks.
* No new `wazuh-manager.conf` options.

---

# Tasks

## 1 — `tasks.db` durability

**Dependencies:** None

* [x] Enable WAL on **every** database open while keeping `synchronous=FULL`.
* [x] Add `wdb_commit2` inside the claim and create sub-commands.
* [x] Add a standalone `task commit`.
* [x] Commit and finalize before daily `VACUUM`, as the global path does. `task sql` can reach `wdb_exec` with a transaction still open, and SQLite refuses to run `VACUUM` inside a transaction.

## 1b — Bound blocking I/O

**Dependencies:** None

* [x] Add a `wdbc_query_ex_timeout` variant with socket timeouts; keep the existing wrapper behaviour unchanged.
* [x] Do **not** change behaviour for existing callers. `wdbc_query_ex` is used by every daemon, including syscollector's long-running sync queries.
* [x] Bound the `agent_delete_old` authd path with `external_socket_connect`.

## 1c — Latent fixes required by the epic

**Dependencies:** None

* [x] Make `w_rotate_log` crash-safe and fix its `merror_exit` on `mkdir` failure. Required for rotation handlers to satisfy the idempotency contract.
* [x] Update the cleanup thread to check `wm_shutdown_requested` and use `wm_sleep_until_interruptible`. Today it loops forever around an uninterruptible sleep, keeping `cleanup_interval` (300 s) alive after SIGTERM.
* [x] Delete the dead 4.x fixture `framework/wazuh/tests/data/schema_tasks_test.sql`.

## Design

* [x] Write the design document under `docs/ref/modules/task_manager/`, covering:

  * State machine
  * Ownership vs. timeout
  * Retry and retention policy
  * Handler-registry contract
  * Lane concurrency
  * Schedule rules
  * Fsync budget

The fsync budget is one per task. Validate the target sustained rate against a 20,000-agent deletion wave.

## 2 — Schema and wazuh-db

**Dependencies:** 1

* [x] Add both tables, five indexes, and the `STATUS` `CHECK` without colliding with existing indexes or the lowercase `metadata` table.
* [x] Use `/* */` comments only; do not rely on newlines as SQL separators because `embed_sql.cmake` removes them.
* [x] Implement claim as one sub-command: `SELECT ... ORDER BY NEXT_ATTEMPT_AT LIMIT 1`, `UPDATE`, `wdb_commit2`.
* [x] Implement create as one sub-command with per-type coalescing and the 64-scan bound. Report `created`, `coalesced`, `collided`, and `queue_full`.
* [x] Do not add a unique index; legitimate re-queue operations would trigger `SQLITE_CONSTRAINT`.
* [x] Implement re-queue as its own sub-command, including the competing-row check that produces `superseded` and transfers counters to the survivor.
* [x] Add sweep sub-commands:

  * Per-owner seek for periodic sweeps
  * Paged enumeration for startup sweeps
* [x] Add:

  * `get_manager_task`
  * `get_manager_task_by_agent`
  * Paged `get_manager_tasks_by_type` with a status filter
* [x] Use these for authd's startup seed, pending-delete count, and operator lookup of dead-lettered rows.
* [x] Extend cleanup with:

  * Per-class retention
  * Per-schedule cap excluding `dead_letter`
  * Hard row ceiling, evicting `dead_letter` last
  * Startup reaper for task types with no lane
* [x] Run the startup reaper **after** the ownership sweep, otherwise a `claimed` orphan can be missed on every boot.

## 3 — Dispatcher

**Dependencies:** 1b, 2

* [x] Implement the handler registry with route descriptors as data, including:

  * `max_attempts`
  * `max_defer`
  * `allow_terminal_failure`
  * `coalesce`
* [x] Resolve inventory-sync's socket from configuration. Default: `-http.sock`; operator-overridable.
* [x] Implement lane threads with:

  * Claim
  * Re-queue
  * Outcome
  * Backoff
  * Both retry ladders
* [x] Use one `uhttp_client_t`, one 4 KB response buffer, and one persistent wazuh-db socket per lane.
* [x] Implement local-lane type rotation.
* [x] Map `uhttp_post`'s tri-state result to the task-result contract **by curl code**.
* [x] Treat connection refused as `not_ready`, not a generic transport error.
* [x] Treat the routes as UDS; DNS-related curl codes are unreachable.
* [x] Zero-initialize `uhttp_result_t` for every call.
* [x] Treat an unparseable 4xx as `retryable` when `allow_terminal_failure=false`.
* [x] Treat an unparseable 409 as `busy`.
* [x] Implement per-type identity:

  * Deterministic for scheduled runs
  * Deterministic for `agent_delete_indexer`, keyed by monotonic journal sequence
  * Random for `vd_scan`
  * Exactly 64 hexadecimal characters
* [x] Implement ownership as `pid:starttime:lane`.
* [x] Implement startup and periodic sweeps, reclaiming live-lane rows only when the in-flight `task_id` differs **and** the claim grace period has elapsed.
* [x] Implement the scheduler:

  * Grouped polling
  * Lane condition-variable signalling
  * Direct signalling from in-process producers
  * 60-second `log_rotate_size` trigger
  * Ownership sweep
* [x] Keep `w_is_worker()` off the poll path; it re-parses `ossec.conf` on every call.
* [x] Add a lane watchdog reported by the cleanup thread.
* [x] Add `create_manager_task` and `get_manager_task` socket actions.
* [x] Add internal options through `getDefine_Int_default`.

## 4a — inventory-sync delete route

**Dependencies:** 2

* [x] Add `POST /_internal/agents/delete`.
* [x] Enqueue the `Item` **with its responder**.
* [x] Keep the admission routes unchanged until 4b.

## 4b — authd

**Dependencies:** 2, 3, 4a

* [x] Implement phases 0–4, including journal sequence, format, and startup reconciliation.
* [x] Rework `purge_is_pending`:

  * Keep memory state on a miss
  * Perform one row check on a hit
  * Drop on terminal-or-absent
  * Return `pending` on query failure
  * Gate explicit-ID inserts until startup seeding succeeds
* [x] Retire the relay thread and retry ladder.
* [x] Keep `purge_reserved`; move only its release to phase 1.
* [x] **Remove `DELETE /agents` and its POST alias** once no production caller remains.
* [x] Add the new API error code and framework/spec/changelog plumbing.

## 5a — inventory-sync scan route

**Dependencies:** 2

* [x] Add `POST /_internal/vd/scan`.
* [x] Add the third `Item::Kind`.
* [x] Add the `workerLoop` branch.
* [x] Add session-less `IVdScanner::scanAgent`.
* [x] Return `409 scan_in_progress` from `AgentInFlightRegistry`.

## 5b — VD

**Dependencies:** 3, 5a

* [x] Create a task through `queue/tasks/task` with its own 2-second timeout, mapping timeout to `503 scan_queue_full`.
* [x] Remove the in-memory deque.
* [x] Verify that neither `remoted` nor #38553's endpoint requires changes.

## 6 — Schedules and monitord port

**Dependencies:** 1c, 3, 4b

* [x] Add built-in schedule definitions.
* [x] Upsert schedules at startup.
* [x] Implement spawning with:

  * Coalescing
  * Overlap skipping
  * Node scoping
* [x] Treat `w_is_worker() == OS_INVALID` as **unknown**, not master.
* [x] Disabling a schedule stops new instances while allowing pending instances to finish.
* [x] Re-enabling recomputes `NEXT_RUN_AT` so missed windows do not produce a burst.
* [x] Port:

  * Disconnection sweep
  * `delete_old_agents`
  * Both rotation branches
* [x] Fold disconnection logging into the sweep using IDs returned by `wdb_disconnect_agents`, with a batched name lookup.
* [x] Fix `delete_old_agent` ID resolution while porting. The current implementation splits a `"name-ip"` string at the last `-` without a NULL check and resolves by name, which:

  * Mis-splits dashed names
  * Is ambiguous for duplicate names
* [x] Use the IDs already returned by `wdb_get_agents_by_connection_status`.
* [x] Map the backlog-full error from `agent_delete_old` to `retryable`.
* [x] Return `incomplete` when the batch or time budget is exhausted.

## 7 — Remove monitord

**Dependencies:** 6

* [x] Update:

  * `wazuh-server.sh`
  * `CMakeLists.txt`
  * `src/monitord/`
  * Unit tests
  * Packaging (`spec`, `postinst`, `inst-functions.sh`, `check_files`)
  * Five workflow path filters
  * Tooling
  * Documentation
* [x] **Do not break the agent.**

  * `client-agent/CMakeLists.txt:34` places `src/monitord/` on the agent include path.
  * `shared/CMakeLists.txt` globs the monitord test wrappers, which wrap `w_rotate_log` for an agent test.
  * `shared/src/log_rotate.c` remains.
  * `internal_options.conf:77-92` is read by the **agent**.
* [x] Remove `component='monitor'` from all four locations:

  * `configuration.py:971`
  * `configuration.py:975`
  * `validator.py:214`
  * `spec.yaml` (`:2670-2678`, `:2645`, `:2503`)

---

# Tests

Most testing is integration-level.

Deletion and scan cases run in inventory-sync's `qa/` harness, which already stubs the indexer and is the only place where `_delete_by_query` can be deliberately stalled.

Monitord-retirement and agent-delivery regressions run in the manager harness under:

```text
api/test/integration/env/
```

> **Naming an owner for the integration suite is an approval gate for this epic.**

The three core correctness claims:

1. The intent gap is closed.
2. A claim survives a wazuh-db restart.
3. `completed` means the work was actually performed.

All three require integration tests. An unowned suite therefore means shipping a correctness argument that nobody is responsible for executing.

### Unit tests

* [x] State machine, including `superseded` and `incomplete`
* [x] Backoff and deferral ladders
* [x] Counter reset and monotonicity
* [x] Result mapping with per-type overrides
* [x] Both truncated-body fallbacks
* [x] Identity generation per task type
* [x] Claim, create, and re-queue SQL
* [x] Retention ordering
* [x] Reconciliation predicate
* [x] Orphan reaper

### Integration tests

* [x] The three core correctness claims above
* [x] A freshly claimed row surviving the sweep because of the claim grace period
* [x] Recovery from an uncommitted outcome
* [x] Crash and hang recovery
* [x] Overlapping-instance sweep
* [x] Boot-race deferral
* [x] `busy` after a lane timeout
* [x] `vd_scan` reaching `dead_letter` through repeated deferral
* [x] The same `vd_scan` case **under load**, where rows are repeatedly superseded
* [x] Verify `agent_delete_indexer` does **not** dead-letter
* [x] Backlog refusal
* [x] Burst against a nearly full journal
* [x] wazuh-db outage does not block deletion
* [x] End-to-end deletion
* [x] Fleet-wide deletion triggering phase-0 refusal
* [x] Bounded shutdown
* [x] Both log-rotation regressions

### Regression tests

* [x] Agent-delivery path remains unchanged under WAL
* [x] Daily `VACUUM` completes when cleanup and vacuum align
* [x] WAL recovers on the next open after `SIGKILL`
* [x] Agent continues to build after `src/monitord/` is removed
* [x] Add `CHANGELOG` entries for both breaking API changes

---

# Definition of Done

* [x] Manager tasks survive a restart. A task interrupted during execution is reclaimed at startup and eventually completes, and no task state is lost because of wazuh-db's deferred commit.
* [x] A task reaches `completed` **only after the work has actually been performed**. For `agent_delete_indexer`, this means the documents are gone — not merely that inventory-sync accepted the request.
* [x] Retries are bounded and backed off. A consumer that is merely not ready or busy finishing an abandoned attempt does not consume retry budget.
* [x] Deferral terminates for every type that can be asked again, but never for `agent_delete_indexer`.
* [x] No task is claimed twice.
* [x] No task starves.
* [x] **No manager task is ever executed twice concurrently**, including a row claimed microseconds before an ownership sweep.
* [x] Agent deletion leaves no orphaned documents at any crash point.
* [x] A live agent is never purged.
* [x] `client.keys` is never blocked on wazuh-db for longer than one bounded call.
* [x] None of `d43311dff6`'s existing guarantees are lost.
* [x] Queued VD scans survive a restart while admission semantics remain unchanged.
* [x] Neither `remoted` nor #38553's endpoint requires modification.
* [x] Recurring work runs on schedule, with one run per missed window rather than a burst.
* [x] Master-scoped schedules never run on workers.
* [x] `wazuh-manager-monitord` is removed from build, packaging, service scripts, health checks, and documentation.
* [x] The agent continues to build, and its rotation-related internal options remain untouched.
* [x] Log rotation is crash-safe.
* [x] Daily rotation still occurs when the manager restarts daily.
* [x] The daily rotation branch never suspends size-based rotation.
* [x] Failed operations are both **recorded and findable**. Dead-letter rows survive retention and can be listed without relying on the original log line.
* [x] The agent-task path is provably unchanged, with all existing tests passing.

TomasTurina:
## Description

Agent deletion used to purge the Indexer synchronously. If the Indexer was down, the documents were orphaned and nothing remembered to try again — the agent was gone from `client.keys`, so nobody would ever ask a second time. The root cause is a `200` that means "queued": an admission response cannot back a record that reads `completed`.

This adds a durable, manager-side task queue to the Task Manager and moves that work onto it. A manager task is claimed, executed and retired with an outcome, and is retried until it reaches one — surviving a manager restart, an Indexer outage and a `wazuh-db` restart. It then moves the on-demand vulnerability scan and the three periodic jobs that `wazuh-manager-monitord` owned onto the same queue, and removes that daemon.

Closes #38589

## Proposed Changes

**The queue** (`MANAGER_TASKS` in `tasks.db`, dispatcher in `wazuh-manager-modulesd`)

- New `MANAGER_TASKS` and `MANAGER_TASK_SCHEDULES` tables with six indexes, a `CHECK` on `STATUS`, and eighteen `task` sub-commands in `wdb_manager_task.c`. `tasks.db` opens in WAL with `synchronous=FULL`.
- A registry maps a task type to its lane, route or handler, timeouts, retry budget and admission bound. Adding a type is a row plus a handler — not a change to the dispatcher, and not a change to `wazuh-db`, which treats `TASK_TYPE` as opaque text.
- Three lanes (delete ×4, scan ×1, local ×1) plus one scheduler thread. Lanes sleep on a condition variable; the scheduler runs a single grouped poll and wakes only lanes with work due.
- Two ladders: retry backoff (30 s → 900 s) for real failures, and a much shorter deferral ladder (from 5 s) for a consumer that has not bound its socket yet, because that is a boot race and pricing it at the cap taxes every restart.
- `agent_delete_indexer` never gives up — no attempt budget, no deferral budget, and a `4xx` re-queues rather than failing, since once `client.keys` is written the row is the only remaining record of the obligation.
- Ownership sweep, orphaned-type reaper, and a watchdog that observes (there is no cancellation primitive available, so a hung handler is made visible rather than stopped).

**Agent deletion** — `authd` writes the key, journals the intent before `OS_WriteKeys`, then creates the task row; a crash anywhere in that sequence is reconciled at the next start. Deletion ids are derived from a monotonic journal sequence so the writer and startup reconciliation collide harmlessly. `DELETE /agents` now returns before the documents are gone, and can be refused with the new `1765` when too many deletions are already pending.

**Vulnerability scans** — `POST /vulnerability-detector/scan` records a task instead of pushing onto an in-memory deque that was lost on restart. Executed by a new `POST /_internal/vd/scan` on inventory-sync, which answers at completion and refuses a second concurrent scan of the same agent with `409`.

**monitord retirement** — the disconnection sweep, the disconnection log line, `delete_old_agents` and both kinds of log rotation become recurring manager tasks; `wazuh-manager-monitord` is deleted across build, packaging, CI, API, framework and docs. Three behaviour fixes fall out: the sweep no longer runs on cluster workers (it did, despite the flag meant to prevent it); daily rotation now fires after a same-day restart; and agent deletion resolves by id rather than splitting a `"name-ip"` string at its last dash.

### Results and Evidence

Query plans for all 32 manager-task statements were checked by loading the shipped schema into SQLite and populating a realistic steady state (100k rows at the ceiling, ~90% terminal, 50 pending, 6 claimed). Every index is used; every column is read or filtered; no statement is unprepared.

Retiring terminal rows was the one full table scan, and `idx_manager_tasks_status_end (STATUS, END_TIME)` was added to remove it:

| | before | after |
|---|---|---|
| retention, 288 passes (1 day), nothing to delete | 2,847 ms | 1 ms |
| 20k claim-and-complete cycles | 271 ms | 304 ms |

That is +1.6 µs per task against ~10 ms per pass held under the per-database mutex every other `tasks.db` command serialises on — break-even near 21 manager-tasks/second sustained. The schema was also verified to survive `embed_sql.cmake`'s newline stripping.

Evidence of `tasks.db`:

```
# sqlite3 /var/wazuh-manager/queue/tasks/tasks.db
SQLite version 3.45.1 2024-01-30 16:01:20
Enter ".help" for usage hints.
sqlite> .tables
MANAGER_TASKS           TASKS                 
MANAGER_TASK_SCHEDULES  metadata              
sqlite> .headers on
sqlite> 
sqlite> select * from TASKS;
TASK_ID|AGENT_ID|TASK_TYPE|PAYLOAD|CREATE_TIME|DELIVERY_TIME|STATUS
dea79095-9441-4518-499f-c6891636ee8d|001|agent_restart|{}|1788361395|1788361397|delivered
sqlite> 
sqlite> select * from MANAGER_TASK_SCHEDULES;
SCHEDULE_ID|NEXT_RUN_AT|ENABLED
agent_disconnect_sweep|1788365648|1
agent_delete_old|0|0
log_rotate_daily|1788393610|1
sqlite> 
sqlite> select * from MANAGER_TASKS;
TASK_ID|TASK_TYPE|PAYLOAD|CREATE_TIME|AGENT_ID|STATUS|OWNER|CLAIM_TIME|ATTEMPTS|DEFER_COUNT|LAST_ERROR|NEXT_ATTEMPT_AT|SCHEDULE_ID|SCHEDULED_RUN_AT|END_TIME
dc42ff56f241338487825451534edc43ffbf820a235b502e5ce662cb28fd3e16|vd_scan|{"agent_id":"001"}|1788361338|001|completed|||0|0||1788361338|||1788361370
5c7335297a48ad06ef7238099444952c020bbef9bab90a98e49873d64a7cd6e4|agent_disconnect_sweep|{"schedule_id":"agent_disconnect_sweep","scheduled_run_at":1788362048}|1788362049||completed|||0|0||1788362048|agent_disconnect_sweep|1788362048|1788362049
3af19c9a9766c6a34408734c2b5b295fc51cf17af73d452248ff183f0f55e343|agent_disconnect_sweep|{"schedule_id":"agent_disconnect_sweep","scheduled_run_at":1788362948}|1788362950||completed|||0|0||1788362948|agent_disconnect_sweep|1788362948|1788362950
e6e66cac02c4ff0336ccce43db7fef7fff58f705fbed86c445e641d47d147106|agent_disconnect_sweep|{"schedule_id":"agent_disconnect_sweep","scheduled_run_at":1788363848}|1788363852||completed|||0|0||1788363848|agent_disconnect_sweep|1788363848|1788363852
a74761e067c3da8984ba8b35eabd8d2cfd155abb4e0f9ac5da23f9e2031bfd27|agent_disconnect_sweep|{"schedule_id":"agent_disconnect_sweep","scheduled_run_at":1788364748}|1788364752||completed|||0|0||1788364748|agent_disconnect_sweep|1788364748|1788364752
10262515b3ce4e0fbf204e4039886e7a448925a896419f79444559d778f470e9|agent_delete_indexer|{"agent_id":"001"}|1788365568|001|claimed|51573:143054:delete-0|1788365692|0|0||1788365688|||
2cce1087d889fac47d9d586e4ba0a9c231f0bf52012933ef72f586f70c62775c|agent_disconnect_sweep|{"schedule_id":"agent_disconnect_sweep","scheduled_run_at":1788365648}|1788365652||completed|||0|0||1788365648|agent_disconnect_sweep|1788365648|1788365652
```

### Artifacts Affected

- `wazuh-manager-modulesd`, `wazuh-manager-db`, `wazuh-manager-authd`, `wazuh-manager-remoted` (Linux manager).
- **`wazuh-manager-monitord` is removed** — dropped from `DAEMONS`, the RPM/DEB file lists, `inst-functions.sh`, the CI file manifest and five workflow path filters.
- `etc/wazuh-manager.conf` and `etc/templates/config/generic/global.template`.
- `tasks.db` schema. 5.0 managers are always a fresh install, so there is no migration and `PRAGMA user_version` is unchanged; an existing `tasks.db` will not pick up the new index, since the schema runs only when the open fails.

### Configuration Changes

- **No new XML.** The four `<task-manager>` options are untouched.
- **Removed: `<global><agents_disconnection_alert_time>`.** A configuration that still carries it fails to start. It delayed the disconnection *alert* while a flapping agent could return, but alert emission to `analysisd` was already removed in 5.0, leaving it controlling only the timing of a debug log line, and its last consumer went with monitord. Agents are still marked `disconnected` after `<agents_disconnection_time>` with no delay — which this option never gated.
- **Renamed, manager only: `monitord.*` → `wazuh_modules.manager_task_*`** (eight keys; full table in the manager configuration migration guide). An override left under an old name is silently ignored. **Agents are unaffected** and keep `monitord.*` for their own log rotation.
- **Removed: the `monitor` component** of the active-configuration endpoints, which monitord's `getconfig` socket served. `GET /manager/configuration/monitor/{global,internal}` and the cluster equivalent now return `400`. The settings it exposed are reported instead under `GET /manager/configuration/wmodules/wmodules` → `task-manager.recurring_tasks`.
- New API error `1765` when an agent deletion is refused because too many are already pending.
- 29 new internal options, all with in-code defaults (the manager ships an empty overrides file). Documented in `docs/ref/modules/task_manager/manager-tasks.md` and `configuration.md`.

### Tests Introduced

Unit, alongside the existing suites:

- `test_wdb_manager_task.c` — the queue operations, the state machine including `superseded` and `incomplete`, retention ordering, and the schedule table.
- `test_wm_manager_task_registry.c` / `_dispatcher.c` / `_owner.c` / `_client.c` — retry and deferral arithmetic, result mapping, identity per type, lane rotation, the reclaim predicate, the spawn decision, and every wazuh-db wrapper.
- `test_wm_manager_task_schedules.c` / `_local.c` — slot arithmetic including missed-run coalescing and DST, node scope, startup and re-enablement, the retention window, and the mapping from an authd refusal to a retry.
- `test_purge_journal.c` — authd's journal, reservations and pending-set.
- `vdScanEndpoint_test.cpp`, `vdScanLane_test.cpp`, `deleteAgentEndpoint_test.cpp` — the two execution routes and the per-agent interlock.
- qa harness: `test_delete_agent.py` and `test_vd_lane.py` cover both routes' contracts against a stubbed Indexer.

**Not covered.** The integration cases in the design's verification section are not written, and the qa harness still lacks the two hooks they need — a stall on `_delete_by_query` and a `wazuh-db` kill switch. That leaves the three claims this change rests on unexecuted: the intent gap is closed, a claim survives a `wazuh-db` restart, and `completed` means purged. Also unwritten: spawn idempotency across a kill between the insert and the slot advance, WAL recovery after `SIGKILL`, bounded shutdown mid-call, and the monitord-retirement regressions. `log_rotate.c` still has no unit target.

## Review Checklist
- [ ] Code changes reviewed
- [x] Unit suites built and green on this branch
- [ ] Integration suite owner named, and the qa harness stall/kill hooks added


LucioDonda (dismissed):
## LGTM

At the end some corner cases to double check

* Installed from sources

<p>


```console
General settings:
    TARGET:             manager
    V:                  
    DEBUG:              
    EXTERNAL_SRC_ONLY:  
User settings:
USE settings:
    USE_INOTIFY:        no
    USE_BIG_ENDIAN:     no
    USE_SELINUX:        no
    USE_AUDIT:          no
    IMAGE_TRUST_CHECKS: 1
Compiler:
    CC                cc
    MAKE              make
make[1]: Leaving directory '/workspaces/devContainer/wazuh/src'

Done building manager

Wait for success...
success
Processing ./.
  Created wheel for wazuh: filename=wazuh-5.0.0-py3-none-any.whl size=273904 sha256=efb09eb229394bc33bdd8a82ffcc083018e668af0f62ecd26cb672ccece40cf0
  Stored in directory: /tmp/pip-ephem-wheel-cache-tjc8ceqk/wheels/d0/a0/5b/ec6d654f16af27945f6196843277c1333ef1f7d7cbee0944d6
Successfully built wazuh
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the --root-user-action option if you know what you are doing and want to suppress this warning.

Processing ./.
  Created wheel for api: filename=api-5.0.0-py3-none-any.whl size=96732 sha256=70c3e8f22fd06b8888824470ea3ef18cca7f8ee8b5b14f7e30523909fa30b30a
  Stored in directory: /tmp/pip-ephem-wheel-cache-k3hm7lwz/wheels/26/26/02/106c76a98f584faf72026c08b574a73b643c428c8fc38f1dbc
Successfully built api
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the --root-user-action option if you know what you are doing and want to suppress this warning.

Generating schema files...
Loading resources...
Loading wcs_flat.yml...
Successfully merged 1 files into temporary file: /tmp/merged_wcs_72maaqpz.yml
Loading schema template...
Success.
Success.
Success.
Generating engine schema...
Success.
Success.
Success.
Cleaned up temporary file: /tmp/merged_wcs_72maaqpz.yml
Generated wazuh-decoders.json
Success.
Copying store files...
Installing GeoIP databases...


Starting Wazuh...
manager
2026/09/01 19:33:24 wazuh-manager-modulesd: ERROR: File 'etc/certs/root-ca.pem' not found for 'indexer.ssl.certificate_authorities' in module 'indexer'. Check configuration
2026/09/01 19:33:24 wazuh-manager-modulesd: ERROR: (1202): Configuration error at 'etc/wazuh-manager.conf'.
wazuh-manager-modulesd: Configuration error. Exiting

 - To start Wazuh:

 - To stop Wazuh:

   More information at: 
```


</p>
</details> 

* Status and synched

<p>

```console
╰─# /var/wazuh-manager/bin/wazuh-manager-control start ; tail -f /var/wazuh-manager/logs/wazuh-manager.log
Starting Wazuh v5.0.0...
Started wazuh-manager-apid...
Started wazuh-manager-db...
Completed.
2026/09/01 19:34:37 wazuh-manager-modulesd:inventory-sync-server: INFO: The indexer is reachable; inventory documents can be delivered.
2026/09/01 19:34:38 wazuh-manager-modulesd:vulnerability-scanner: INFO: No existing local vulnerability feed detected at startup. A full feed will be downloaded.
2026/09/01 19:34:38 wazuh-manager-modulesd:vulnerability-scanner: INFO: CVE feed not yet available — per-agent scans deferred until initial load completes.
2026/09/01 19:34:38 wazuh-manager-modulesd:vulnerability-scanner:server: INFO: vulnerability detector server bound to 'queue/sockets/vd-http.sock' (mode 0660, 2 I/O thread(s), 2 concurrent accept(s), max 1024 connection(s), 64 reserved for control, 960 data session(s), 268435456 byte in-flight budget, 268152832 byte body cap, 282624 byte per-request overhead; timeouts s: header=10 body=30 response=3600 write=10 drain=2).
2026/09/01 19:34:38 wazuh-manager-modulesd:content-updater: INFO: IndexerDownloader: Consumer 'cti:catalog:consumer:vulnerabilities' not found in '.wazuh-cti-consumers'. Waiting 60s before retrying.
2026/09/01 19:34:38 wazuh-manager-analysisd: INFO: [Geo::Manager] Changes detected for CITY database 'GeoLite2-City.mmdb', updating...
2026/09/01 19:34:39 wazuh-manager-analysisd: INFO: [Geo::Manager] Successfully updated CITY database 'GeoLite2-City.mmdb'
2026/09/01 19:34:39 wazuh-manager-analysisd: INFO: [Geo::Manager] Changes detected for ASN database 'GeoLite2-ASN.mmdb', updating...
2026/09/01 19:34:40 wazuh-manager-analysisd: INFO: [Geo::Manager] Successfully updated ASN database 'GeoLite2-ASN.mmdb'
2026/09/01 19:34:40 wazuh-manager-analysisd: INFO: [RemoteConfig] Remote settings synchronized: changes detected and applied
2026/09/01 19:34:43 wazuh-manager-analysisd: INFO: [CM::Sync] Successfully synchronized space 'standard'
```

```console
(venv) ╭─root@9fac3eeb828e /workspaces/devContainer/wazuh ‹feat/38589-manager-tasks●› 
╰─# engine-private ns list                                                                                                                                                                                                     130 ↵
- cmsync_standard_9b8f

(venv) ╭─root@9fac3eeb828e /workspaces/devContainer/wazuh ‹feat/38589-manager-tasks●› 
╰─# engine-router list
- entry_status: ENABLED
  name: cmsync_standard
  namespaceId: cmsync_standard_9b8f
  priority: 1
  uptime: 1002
```

</p>
</details> 

* Tests with agents

<p>

Image
Image

</p>
</details> 


* Additional Comments to analyze 🟡
 
<p>


<details>
<summary><b>1. A failed phase 3 strands the deletion until authd restarts</b></summary>

`purge_create_rows()` has exactly two call sites:

- [`main-server.c:1310`](../../../wazuh/src/os_auth/src/main-server.c#L1310) — startup recovery, over `purge_journal_reconcile()`'s output
- [`main-server.c:1499`](../../../wazuh/src/os_auth/src/main-server.c#L1499) — the writer, over **this cycle's** `journaled` array only

Nothing re-reads the journal between those two. So both non-success exits from the phase-3 loop strand work:

```c
case MANAGER_TASK_QUEUE_FULL:
    mwarn("... It stays journaled and will be retried.");
    goto finish;          /* <- rest of the batch abandoned */
default:
    mwarn("... it stays journaled and will be retried.");
    goto finish;
```

`finish:` drops only the entries that *did* become durable. The remainder keep their journal lines — and no later writer cycle looks at them, because `journaled` is rebuilt from `copy_remove` each pass.

**Failure scenario.** wazuh-db is unreachable (restart, socket timeout) while a bulk deletion of N agents is in the writer's cycle:

1. Phase 1 journals all N; phase 2 rewrites `client.keys` successfully — the agents are **gone from disk**.
2. Phase 3's first `manager_task_create()` returns `MANAGER_TASK_CREATE_FAILED`; `goto finish` with `recorded == 0`.
3. wazuh-db comes back thirty seconds later. Nothing notices. There is no `agent_delete_indexer` row for any of the N agents.
4. Their indexer documents stay orphaned for the remaining lifetime of the authd process. That is the #38589 bug, on the one failure mode the whole design is built to survive.
5. Secondary: `purge_journal_size` never decreases, so those N entries permanently consume the `PURGE_QUEUE_MAX_ENTRIES` (65 536) admission budget, and `purge_is_pending_locally()` keeps refusing those ids to new enrollments.

The `keys_written == false` branch at [`main-server.c:1501`](../../../wazuh/src/os_auth/src/main-server.c#L1501) has the same shape. Its comment — *"the journal lines stay, the next cycle retries them"* — is also not what the code does. There the eventual outcome is at least safe (reconciliation finds the agents still in `client.keys` and drops the lines), but only after a restart, and only if the *next* successful `OS_WriteKeys` hasn't removed them from disk in the meantime — which it will, since `OS_DeleteKey()` already ran against the in-memory keystore. That path can produce a stranded deletion too.

**Suggested fix.** Make the writer operate on the whole outstanding set rather than on its own additions. Either:

- add a `purge_journal_snapshot(size_t *count)` alongside `purge_journal_reconcile()` and call `purge_create_rows()` on that at the top of each cycle (phase 3 becomes "record everything owed", not "record what I just journaled"), or
- keep an in-memory `owed` list that `purge_journal_append()` adds to and `purge_journal_drop()` removes from, and drain it each cycle.

The first is smaller and reuses the structure already there. Either way the `mwarn` texts become true.

**Test gap.** `test_purge_journal.c` covers append / drop-by-sequence / reconcile / `purge_is_pending` / the id counter thoroughly, but has no case for "phase 3 failed — what happens on the next cycle". That is where this lives.

</details>

<details>
<summary><b>2. Partial dispatcher start leaves live threads unowned</b></summary>

[`wm_manager_task_dispatcher_start()`](../../../wazuh/src/wazuh_modules/src/task_manager/wm_manager_task_dispatcher.c#L913) returns `-1` from several points, and the caller treats that as "did not start":

```c
if (wm_manager_task_dispatcher_start(&g_manager_task_dispatcher, ...) == 0) {
    g_manager_task_dispatcher_started = true;
} else {
    mterror(..., "Manager task dispatcher did not start; manager tasks will not run.");
}
```

`wm_manager_task_dispatcher_stop()` is then guarded on `g_manager_task_dispatcher_started`, so on a partial start it is **never called**.

- If `pthread_create` fails for worker *i* > 0, workers 0..*i*−1 are running and `started = true` on each. Nobody wakes them, nobody joins them. The lanes go on claiming and executing manager tasks while the log says they will not run, and the documented shutdown property — *"the lanes come down inside the shared thirty second budget … what makes a half-written query to wazuh-db impossible rather than merely unlikely"* — no longer holds.
- The earlier returns (`uhttp_client_new` failure, `wm_manager_task_owner_format` failure, the routed-lane `lane_type_count != 1` check) happen before any thread exists, so those only leak: the workers array, `schedules`, `direct_pending`, and every `lane_mutex` / `lane_cond` / `published_mutex` initialised so far.

A single `goto fail:` that broadcasts, joins whatever has `started`, frees the arrays and destroys the mutexes would cover both.

Related: `wm_manager_task_dispatcher_stop()` on the normal path also never destroys the mutexes and condvars `start()` initialised. Harmless while the module starts once per process, but it makes `stop()` not the inverse of `start()`.

</details>

</p>
</details>

src/wazuh_modules/src/task_manager/wm_manager_task_dispatcher.c
```diff
@@ -0,0 +1,1136 @@
...
+    // Strictly after the sweep. A claimed orphan only becomes pending once the sweep has
+    // reclaimed it, so a reaper that ran first would miss it on this boot and every one after.
+    wm_manager_task_reap_orphans(dispatcher);
+
+    /* And strictly after the reaper, for the same reason in reverse: a schedule that upserted
+     * before the reaper ran could spawn an instance of a type the reaper is about to retire rows
+     * of, and the fresh row would be caught in the same pass it was created for. */
+    wm_manager_task_schedules_upsert(dispatcher);
+
+    for (size_t i = 0; i < dispatcher->worker_count; i++) {
```
NahuFigueroa97, line 1064: If `pthread_create` fails for worker *N* (or for the scheduler after all six workers already started), the function returns `-1` **without stopping, waking, or joining anything already created**. `wm_manager_task_dispatcher.h:144` documents the contract as *"@return 0 on success, -1 on failure, in which case nothing has been started"* — false the moment one thread has already started. The caller (`wm_task_manager.c:328-332`) only calls `stop()` when `start()` returns `0`, so on partial failure `stop()` is never invoked for the rest of the process's life.

```mermaid
sequenceDiagram
    participant Main as wm_task_manager.c
    participant Start as dispatcher_start()
    participant W0 as lane worker 0..3
    participant W4 as lane worker 4 (pthread_create fails)

    Main->>Start: dispatcher_start()
    Start->>W0: pthread_create() x4 — succeed
    activate W0
    Note over W0: started = true, already polling wazuh-db
    Start->>W4: pthread_create() — fails (EAGAIN)
    Start-->>Main: return -1
    Main->>Main: g_manager_task_dispatcher_started = false
    Note over Main: log: "manager tasks will not run"
    Note over W0: ...but they do. Never joined, never stopped,<br/>invisible to the watchdog (started flag is false).
    deactivate W0
```

Concrete trigger: this PR adds 7 threads to modulesd in one shot (11→18). Under `RLIMIT_NPROC` or memory pressure — plausible exactly because of that jump — this is reachable in production, not a contrived edge case.

**Fix:** one unwind path in `start()` — on any failure, signal stop, wake and join every already-created worker (and the scheduler if applicable), free their HTTP/DB clients, and only then return -1.

*Secondary, same root cause:* threads are created with raw `pthread_create` instead of the codebase's `CreateThreadJoinable` wrapper that every other daemon thread uses (`wm_syscollector.c`, `wm_sca.c`, `os_execd`, etc.) — this also silently drops the operator-tunable `wazuh.thread_stack_size` for these 7 new threads.

src/os_auth/src/main-server.c
```diff
@@ -701,22 +691,16 @@ int main(int argc, char **argv)
     }
 
     if (!config.worker_node) {
```
NahuFigueroa97, line 712: Both comment (`"before any thread exists"`) and `auth.h`'s doc comment on `purge_file_load()` (`"Call after OS_ReadKeys() and before any thread starts"`) assert an invariant the code doesn't hold: the local/remote listener threads are already live while the journal is still empty and `purge_startup_recover()` — new in this PR, and doing real wazuh-db round trips per recovered entry — hasn't run.

```mermaid
sequenceDiagram
    participant Boot as authd main()
    participant Local as thread_local_server (LIVE)
    participant Journal as purge_file_load() + purge_startup_recover()
    participant Ext as external request (e.g. migration insert for id 042)

    Boot->>Local: pthread_create — starts accepting requests
    activate Local
    Note over Journal: not yet run — journal empty in memory
    Ext->>Local: insert agent, explicit id=042
    Local->>Local: purge_is_pending_locally(042) → false (journal empty)
    Local->>Local: manager_task_agent_status(042) → no row yet
    Local-->>Ext: admitted — id 042 reassigned
    Boot->>Journal: purge_file_load() + purge_startup_recover() (now runs)
    Journal->>Journal: reconcile 042 against client.keys → finds 042 PRESENT (new agent)
    Note over Journal: classified as "nothing owed" — journal line dropped silently
    deactivate Local
    Note over Journal: original agent's indexer documents: never purged
```

This isn't a pre-existing bug widened by refactor — it's the exact failure mode ("intent gap during a crash/restart window") the whole epic exists to close, reopened at a different point in the lifecycle, and made worse by this PR because `purge_startup_recover()` is new, is slower (network round trips) the bigger the crash-recovered backlog, and the design doc itself calls out "manager restarted with thousands of pending deletions" as the case it exists for — which is precisely the case that makes this window largest.

**Fix:** don't start `thread_local_server`/`thread_remote_server` until `purge_file_load()` + `purge_startup_recover()` complete — mirror the pattern already used for `thread_writer`.

src/wazuh_modules/inventory_sync_server/src/endpoints/deleteAgentEndpoint.cpp
```diff
@@ -149,13 +209,14 @@ namespace invsync::endpoints::delete_agent
...
+             * is not there is a no-op the whole chain ignores -- which is what makes the
+             * dispatcher's retry of an already-purged agent free.
              *
              * If the pipeline then refuses the item below, these deletes have already been queued.
              * That is harmless: they are idempotent, the agent is gone from client.keys so nothing
-             * will write those documents again, and authd retries the whole deletion until it is
-             * accepted.
+             * will write those documents again, and the dispatcher retries the whole deletion until
+             * it is accepted -- this type has no attempt budget. Each retry re-queues this pair for
+             * the same reason, which is bounded and free.
```
NahuFigueroa97, line 219: wazuh-agent-config/wazuh-agent-stats deletes go through the async connector's queue (bulkDelete), which is fire-and-forget by construction, in-memory only (IndexerBulkQueue::m_buffer is a plain std::deque, no disk backing). The task is marked completed once the separate, synchronous wazuh-states-* delete-by-query flushes — the header even documents this precisely: "a 200 promises the state documents were deleted and flushed, and promises only that the other two deletions were queued."

If modulesd restarts or crashes before that in-memory queue flushes, the config/stats documents for that agent are never deleted, and because the task is already completed, nothing ever revisits it. Silent, permanent, and — since it's below the level anyone would think to check — hard to detect after the fact.

Narrower blast radius than it sounds (only 2 auxiliary indices, not the main state documents), but it's a direct contradiction of this epic's own Definition of Done: "A task reaches completed only after the work has actually been performed... the documents are gone."

Fix: either make the async queue durable across restarts, hold completed until the by-id deletes are acknowledged, or add a periodic reconciliation sweep for config/stats documents belonging to agents no longer in wazuh-db.

src/wazuh_modules/src/task_manager/wm_task_manager.c
```diff
@@ -56,10 +74,24 @@ const char *task_type_names[] = {
...
+ * it publishes, which is the whole of what is achievable about a hung handler: there is no
+ * cancellation primitive in the tree. */
+static wm_manager_task_dispatcher g_manager_task_dispatcher = {0};
+static bool g_manager_task_dispatcher_started = false;
+
 // Global notification queue for worker threads
 static wnotify_t *task_notify_queue = NULL;
 static pthread_mutex_t task_queue_mutex = PTHREAD_MUTEX_INITIALIZER;
 
+STATIC const char* wm_task_manager_consumer_socket(void) {
```
NahuFigueroa97, line 87: Inventory-sync socket isn't actually operator-overridable, despite the design requiring it. wm_task_manager_consumer_socket() returns a compile-time constant; its own doc comment admits "nothing can set it today... this is the seam." Self-acknowledged, not hidden — but still a gap against the stated contract.

src/wazuh_modules/src/task_manager/wm_manager_task_local.c
```diff
@@ -0,0 +1,634 @@
...
+
+        if (logged >= manager_task_local_config.disconnect_log_max ||
+            (long long)(time(NULL) - started) >= WM_MANAGER_TASK_SWEEP_LOG_BUDGET || wm_shutdown_requested) {
+            dropped++;
+            continue;
+        }
+
+        /* One lookup per NEWLY DISCONNECTED agent, once per sweep. The name is the only reason for
+         * the lookup -- the id is already in hand -- and OS_AG_DISCON carries both. */
+        if (info = wdb_get_agent_info(disconnected[i], &manager_task_local_wdb_sock), !info) {
```
NahuFigueroa97, line 354: Disconnection-sweep name lookup is N+1, not batched, contradicting the design bullet explicitly. Mitigated by a 200-name cap and a 30s log budget (the code is self-aware of the cost), but a batched query would have avoided the round-trip cost instead of needing a drop-and-count fallback under a large simultaneous disconnect (e.g. a network partition affecting thousands of agents).

src/shared/src/log_rotate.c
```diff
@@ -92,43 +138,48 @@ void w_rotate_log(int compress, int keep_log_days, int new_day, int rotate_json,
     os_snprintf(month_dir, PATH_MAX, "%s/%s", year_dir, MONTHS[tm.tm_mon]);
     os_snprintf(new_path, PATH_MAX, "%s/%s-%02d.log", month_dir, prefix, tm.tm_mday);
     os_snprintf(new_path_json, PATH_MAX, "%s/%s-%02d.json", month_dir, prefix, tm.tm_mday);
-    os_snprintf(compressed_path, PATH_MAX, "%s.gz", new_path);
 
     // Create folders
 
+    // A failed mkdir aborts this rotation and nothing else. It used to call merror_exit(), which
```
NahuFigueroa97, line 144: w_rotate_log()'s mkdir-failure handling changed from fatal to non-fatal — this function is shared with the agent's own rotation thread, so this silently changes agent behavior (crash-and-restart → silent skip) as a side effect of a manager-only retirement PR, with no CHANGELOG line covering the agent side.

src/wazuh_modules/src/task_manager/wm_manager_task_dispatcher.c
```diff
@@ -0,0 +1,1136 @@
...
+    }
+
+    if (dispatcher->scheduler_started) {
+        pthread_join(dispatcher->scheduler, NULL);
+        dispatcher->scheduler_started = false;
+    }
+
+    for (size_t i = 0; i < dispatcher->worker_count; i++) {
+        if (dispatcher->workers[i].started) {
+            pthread_join(dispatcher->workers[i].thread, NULL);
```
NahuFigueroa97, line 1132: Shutdown timeout not enforced against the 600s delete-lane timeout. dispatcher_stop() calls plain pthread_join() with no timeout; there's no cancellation primitive (acknowledged in the code's own comments). If shutdown happens while a delete-lane worker is blocked inside its HTTP call, pthread_join can block up to 600s — 20x the stated 30-second shutdown budget. Ask the author whether an external hard-kill makes this moot, or whether it's a real gap.

NahuFigueroa97 (changes requested):

src/shared/src/log_rotate.c
TomasTurina: Done

src/os_auth/src/main-server.c
TomasTurina: Done

src/wazuh_modules/inventory_sync_server/src/endpoints/deleteAgentEndpoint.cpp
TomasTurina: I've improved documentation related to this.

src/wazuh_modules/src/task_manager/wm_manager_task_dispatcher.c
TomasTurina: Done

src/wazuh_modules/src/task_manager/wm_task_manager.c
TomasTurina: I've improved documentation related to this.

src/wazuh_modules/src/task_manager/wm_manager_task_dispatcher.c
TomasTurina: Done

src/wazuh_modules/src/task_manager/wm_manager_task_local.c
TomasTurina: The transition is a single query (`wdb_disconnect_agents`) and completes before the loop. What follows is diagnostics — one lookup per newly disconnected agent, only to put a name next to an id in the log — capped at 200 (`manager_task_disconnect_log_max`) with a 30 s backstop, remainder dropped and counted. So a partition costs 200 lookups and a summary line, not one per agent.

NahuFigueroa97 (approved):
LGTM!

Image
Image

Issue #31295, opened 2025-08-11
Cluster by default


davidjiglesias:
# Description

From Wazuh v5.0 onwards, there is no longer a distinction between "cluster" and "non-cluster" deployments. The cluster subsystem is always enabled on every installation, whether it contains a single node or multiple nodes. The concept of a "non-clustered" manager no longer exists, and any related configuration or status references must be removed.

For security, the default bind_addr for cluster communications changes from 0.0.0.0 to localhost, avoiding exposure of cluster ports externally unless explicitly configured.

---

# Functional Requirements

1. The cluster must always initialize on startup on all installations (single node behaves as a one-node cluster).
2. There is no configuration option to disable the cluster — the cluster.disabled option must be removed and ignored if present in legacy configs (with a deprecation warning).
3. The default cluster bind_addr must be localhost (overridable by explicit configuration).
4. Remove all references to a "cluster disabled" status in logs, API, and dashboard. Keep cluster name displays where appropriate.
5. Single-node deployments must behave identically to a functional one-node cluster (master) with no extra user configuration.
6. Cluster configuration block review:
   - Remove unused or obsolete options.
   - Retain only necessary fields.
   - Define and apply explicit default values.
   - Define which fields are shown by default in the generated or packaged ossec.conf.
7. Every module must be reviewed to ensure that any references to cluster and non-cluster scenarios or configurations are removed.
8. The Wazuh Dashboard must be reviewed to remove any references, logic, or configuration based on cluster/non-cluster distinctions.

---

# Non-functional Requirements

1. Security by default: No external exposure of cluster ports unless the admin overrides bind_addr.
2. Startup performance: No material increase in startup time vs 4.x.
3. Resource usage parity: A 5.0 default manager (cluster-by-default, single node) must not consume more resources than a 4.x default manager without a cluster, allowing normal measurement variance (±5%). This includes:
   - Average and peak CPU during idle and steady-state.
   - Resident memory (RSS).
   - Open file descriptors and process/thread count.
   - Disk I/O at idle.
   - Network traffic at idle (should be ~0 on a single node).
4. Observability: Clear logs for always-on cluster init and for ignored legacy options.
5. Test coverage: Unit, component, integration, functionality, and performance regression tests for single- and multi-node scenarios.

---

# Implementation Restrictions

1. Do not introduce a new "disable cluster" escape hatch; always-on is mandatory.
2. Do not remove cluster.name or UI/API elements that legitimately reference the cluster name.
[...]

PR #38773, opened 2026-09-01
Reject configuration content the manager cannot use

### LINKED ISSUE #38638: Any text with no tags passes configuration validation, replacing a node's entire config with something it cannot start from, while the API and the validation endpoint both report success
|---|---|---|---|---|
| 5.0.0-latest | wazuh-manager (API: `PUT /cluster/{node_id}/configuration`, `GET /cluster/configuration/validation`, `PUT /groups/{group_id}/configuration`; framework: `wazuh/core/utils.py`, `wazuh/core/manager.py`) | Manager (master + worker) | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 (agents), two-node cluster |

## Description

The endpoint that replaces a node's manager configuration validates the request body by checking that it parses as XML. That check is defeated by the simplest possible input: text with no tags in it.

A node's entire `wazuh-manager.conf` can therefore be replaced with an arbitrary string. Nothing detects it. The running processes keep working, because the configuration is read at start-up, so the deployment looks healthy for as long as nobody restarts that node. When someone does, it refuses to start and drops out of the cluster, and recovery needs filesystem access to the node itself, which is precisely what this endpoint exists to avoid.

`GET /cluster/configuration/validation` reports `status: OK` for that same configuration, so the correct operator procedure of pushing a change and then validating it returns success.

Malformed input is handled correctly, which is what made the gap easy to miss: unclosed tags and mismatched tags are rejected with `1113`, an empty body with `1912`, and the file on disk is left untouched. Only input with no elements at all gets through.

**Severity: High.** A single authenticated call leaves any node, master included, unable to restart, every status surface reports success until someone reboots it, and the master is in scope.

## Already reported

Searched on 2026-08-26 and re-checked against `origin/5.0.0` on 2026-08-27. No issue covers this. Five are worth citing as context rather than as duplicates:

- #16386 ("Configuration validation accepting invalid format") is the same symptom reported against 4.3.10, where `check-manager-configuration` returned OK for a non-XML `ossec.conf`. It was reported against the socket-based validator that 5.0.0 has since removed, which makes it the precedent that this behaviour was already considered a bug once.
- #19390 ("Include extra validation for ossec.conf/agent.conf and other configuration files") asked for exactly the semantic validation that has now been lost, on the grounds that valid XML is not the same as a usable configuration.
- #34326 ("Migrate restart logic to modulesd and remove execd from the manager") is where the socket-based check was replaced with an XML well-formedness check. The semantic half was not replaced with anything.
- #33637 and #33683 ("Improve Wazuh Server API configuration upload validation") are about unescaped ampersands and about the new-vs-current comparison for `limits`/`indexer`. They touch `load_wazuh_xml` but not the synthetic root as a bypass.
- #34573 ("E2E - Validation: API, Cluster & Configuration Endpoints") is an E2E validation task over these endpoints. Its negative-testing task reads "Submit malformed XML/JSON to configuration endpoints", so it exercised malformed input only, and a body with no elements in it was never tried. It is worth mentioning in the fix so that the regression test covers the input that actually gets through.

## Current behavior

Three payloads run against both endpoints, restoring the configuration between each. The node endpoint accepts every one of them. The group endpoint catches only the one that contains an element:

| payload | `PUT /cluster/worker01/configuration` | `PUT /groups/qa-cluster/configuration` |
| --- | --- | --- |
| `<ossec_config><totally_not_a_real_module>x</totally_not_a_real_module></ossec_config>` | **accepted**, 85 bytes written | rejected: `Wazuh syntax error: Invalid element in the configuration` |
| `esto no es una configuracion de wazuh` | **accepted**, 37 bytes written | **accepted** |
| `<!-- nada aqui -->` | **accepted**, 18 bytes written | **accepted** |

There are two separate problems here and it is worth keeping them apart:

1. **The node endpoint runs no schema validation.** A well-formed document naming an element that does not exist replaces the node's configuration without complaint, while the group endpoint refuses the identical payload. The node path does run two narrow policy checks, `check_agents_allow_higher_versions` and `check_indexer`, but nothing that asks whether the manager could parse the result.
2. **Neither endpoint validates a document with no elements in it.** Both remaining checks on the node path iterate the elements they parsed, and so does the group path's semantic check. With zero elements, zero checks can fail, so validation passes vacuously. That is why plain text and a lone comment get through both.

The rest of this report follows the plain-text case, because it is the one that reaches both endpoints and produces the largest effect.

The three rejected payloads, with the config md5 on the node unchanged after each:

```
--- malformed XML (unclosed tag)
{"error": {"code": 1113, "message": "XML syntax error",
           "remediation": "Please, ensure file content has correct XML"}}
    worker md5 after: 913904ff14      (unchanged)

--- mismatched tags
{"error": {"code": 1113, "message": "XML syntax error: mismatched tag: line 7, column 4"}}
    worker md5 after: 913904ff14      (unchanged)

--- empty
{"title": "Bad Request", "detail": "Body is empty",
 "remediation": "Please, check the content of the file to be uploaded", "error": 1912}
    worker md5 after: 913904ff14      (unchanged)

--- not XML at all:  esto no es una configuracion de wazuh
{"data": {"affected_items": ["worker01"], "total_affected_items": 1, "total_failed_items": 0},
 "message": "Configuration was successfully updated in specified node", "error": 0}
    worker md5 after: 9532eaea44      <-- REPLACED
```

The node's configuration afterwards, 84 lines reduced to one:

```
$ docker exec wui-worker cat /var/wazuh-manager/etc/wazuh-manager.conf
esto no es una configuracion de wazuh
$ docker exec wui-worker wc -c < /var/wazuh-manager/etc/wazuh-manager.conf
37
```

Nothing looks wrong at this point. The worker keeps syncing normally, because the configuration was read when it started:

```
2026/08/26 17:59:18 INFO: [Worker worker01] [Agent-groups recv] Finished in 0.001s. Updated 1 chunks.
2026/08/26 17:59:20 INFO: [Worker worker01] [Integrity check] Finished in 0.004s. Sync not required.
```

On restart the node is gone:

```
Starting Wazuh v5.0.0...
Invalid cluster configuration, check the /var/wazuh-manager/etc/wazuh-manager.conf file.
```

```
GET /cluster/nodes
  {"affected_items": [{"name": "manager", "type": "master", ...}], "total_affected_items": 1}
```

`worker01` no longer exists as far as the cluster is concerned, and the API cannot put it back, because the API talks to nodes that are up.

### The validation endpoint blesses it

This is the part that makes it hard to catch. With the worker's configuration in exactly the state above, the product's own validation endpoint was asked about it:

```
GET /cluster/configuration/validation
  {"data": {"affected_items": [{"name": "manager",  "status": "OK"},
                               {"name": "worker01", "status": "OK"}],
            "total_failed_items": 0},
   "message": "Validation was successfully checked", "error": 0}
```

`status: OK` for a configuration that stops the node from starting. An operator doing the responsible thing, pushing a change and then validating, is told everything is fine.

### The master is affected identically

Same endpoint, `node_id=manager`. Backed up and restored through the filesystem so that recovery never depended on the thing being tested, and the master was not restarted:

```
master config before:  94 lines, md5 9a04b7d5cb

PUT /cluster/manager/configuration   ("esto tampoco es una configuracion de wazuh")
  {"affected_items": ["manager"], "total_failed_items": 0,
   "message": "Configuration was successfully updated", "error": 0}

master config after: 42 bytes
  esto tampoco es una configuracion de wazuh

GET /cluster/configuration/validation -> manager: OK, worker01: OK
```

So one authenticated call renders the whole deployment unable to come back from its next restart. Restoring it requires shell access to the master, and until the restart happens nothing at all indicates a problem.

## Root cause: the parser is handed a synthetic root element, in two different places

`PUT /cluster/{node_id}/configuration` reaches `update_ossec_conf`, which validates, backs up, writes, and validates again:

https://github.com/wazuh/wazuh/blob/5.0.0/framework/wazuh/manager.py#L457-L504

```python
# framework/wazuh/manager.py:457
def update_ossec_conf(new_conf: str = None) -> AffectedItemsWazuhResult:
    ...
    validate_wazuh_xml(new_conf)          # (1) pre-write check
    full_copy(common.OSSEC_CONF, backup_file)
    write_ossec_conf(new_conf)
    is_valid = validate_ossec_conf()      # (2) post-write check
    if is_valid.get('status') != 'OK':
        raise WazuhError(1125)
```

Both checks wrap the content in a root element that the input does not have to provide, so neither can fail on a tagless document.

**(1) `validate_wazuh_xml`, `framework/wazuh/core/utils.py:1762`:**

```python
xml = parseString(f'<root>{content}</root>'.replace('&', '&amp;'))
...
incoming_xml = load_wazuh_xml(xml_path='', data=final_xml)
current_xml = load_wazuh_xml(xml_path=common.OSSEC_CONF)
check_agents_allow_higher_versions(incoming_xml, current_xml)
check_indexer(incoming_xml, current_xml)
```

`<root>esto no es una configuracion de wazuh</root>` is well-formed, and the two checks that follow iterate the parsed elements, of which there are none.

**(2) `validate_ossec_conf`, `framework/wazuh/core/manager.py:156`, which is also what `GET /cluster/configuration/validation` reports:**

```python
def validate_ossec_conf() -> dict:
    """Check if Wazuh configuration is OK.

    Validates the ossec.conf file by reading and parsing the XML structure.
    This replaces the previous socket-based validation after execd removal.
    """
    ...
    load_wazuh_xml(xml_path=common.OSSEC_CONF)
    return {'status': 'OK'}
```

The docstring says it: the previous implementation asked the manager, over `WCOM_SOCKET`, to validate the configuration with the real parser, which is what would have rejected this. On 4.14 that function still does so. On 5.0.0 it returns `{'status': 'OK'}` for any document that parses, which is why the validation endpoint agrees with the upload.

**And `load_wazuh_xml` itself, `framework/wazuh/core/utils.py:861`, whose return at line 905 is where the wrapper is applied:**

```python
def load_wazuh_xml(xml_path, data=None):
    ...
    return fromstring(f"{entities}<root_tag>{data}</root_tag>", forbid_entities=False)
```

The wrapper exists so that a Wazuh configuration with several top-level `<ossec_config>` blocks parses as one document. It also makes every tagless document valid. Reproduced with no API and no deployment state:

```
>>> load_wazuh_xml(xml_path='/tmp/plain.xml')
   root_tag | text = 'esto no es una configuracion de wazuh' | children: 0
```

That explains the shape of the results exactly. Anything with a broken tag fails inside the wrapper and is caught. Anything with no tags cannot fail.

### The rollback exists and is dead code for this input

Worth flagging so this is not mistaken for a missing feature. `update_ossec_conf` already copies the current file to `wazuh-manager.conf.backup` before writing, removes it on success, and restores it in `finally` if anything raised:

```python
finally:
    exists(backup_file) and safe_move(backup_file, common.OSSEC_CONF)
```

The recovery path is therefore in place and correct. It never runs here because the validation it depends on cannot fail.

## The group configuration endpoint shares only the second half

The group path is in better shape and the difference is instructive. It does run a real semantic check, the `verify-agent-conf` binary, and that check refuses an element the parser does not know:

```
PUT /groups/qa-cluster/configuration
  <ossec_config><totally_not_a_real_module>x</totally_not_a_real_module></ossec_config>
  -> {"title": "Bad Request",
      "detail": "Wazuh syntax error: Invalid element in the configuration: 'ossec_config'. ..."}
```

The node path accepts that same payload. So the semantic validator exists in the product, and the node endpoint simply does not call it.

Where the group path still fails is the tagless case, for the same reason: same synthetic root, same vacuous check.

```python
# framework/wazuh/core/configuration.py:839, upload_group_configuration()
xml = parseString(f'<root>\n{file_content}\n</root>')
```

```
PUT /groups/qa-cluster/configuration   ("esto no es una configuracion de wazuh")
  {"message": "Agent configuration was successfully updated", "error": 0}

$ cat /var/wazuh-manager/etc/shared/qa-cluster/agent.conf
esto no es una configuracion de wazuh
```

The blast radius is smaller here, since a bad `agent.conf` degrades agents rather than preventing a manager from starting, but it is the same defect in the same shape. It is also worth flagging because testing this endpoint with malformed configurations alone shows all of them correctly refused, which is how group configuration validation was previously recorded as working.

## Why this rates High

- **It is silent for an unbounded time.** The break and its consequence are separated by whenever the node next restarts, which may be weeks. Nothing in `/cluster/nodes`, the node's logs, the dashboard or the validation endpoint indicates a problem in between, so the change and the outage will not be connected when it happens.
- **It defeats the check meant to catch it.** Push, then validate, then walk away is the correct procedure and it returns `OK`.
- **The master is in scope**, and a master that cannot start takes the deployment with it.
- **Recovery leaves the supported path.** The endpoint exists so that node configuration can be managed centrally. After this, the only way back is a shell on the affected node.
- **It needs only an administrator token.**

It is not remote code execution and it is not reachable unauthenticated, which is why it is High rather than Critical.

## Expected behavior

The uploaded configuration is validated with the parser that will actually consume it, and rejected if the manager could not start from it. Specifically:

- Parsing for validation must not be performed against a synthetic root that makes tagless input valid. At minimum, require at least one `<ossec_config>` element on both upload paths.
- `validate_ossec_conf()` should reflect whether the configuration is usable, not merely whether it is well formed. It is what `GET /cluster/configuration/validation` reports, and that endpoint currently cannot fail for any input the upload accepts.

## Steps to reproduce

On any deployment, with an administrator token. Back up the target's config first, because this does what it says:

```bash
# back up through the filesystem, not the API
docker exec <node> cp /var/wazuh-manager/etc/wazuh-manager.conf /tmp/conf.bak

# malformed input is correctly refused
printf '<ossec_config><global><a>b</a></ossec_config>' > /tmp/bad.xml
curl -sk -X PUT -H "Authorization: Bearer $T" -H 'Content-Type: application/octet-stream' \
     --data-binary @/tmp/bad.xml 'https://localhost:55000/cluster/<node>/configuration'
# -> 1113 XML syntax error

# this one is accepted
printf 'esto no es una configuracion de wazuh' > /tmp/plain.xml
curl -sk -X PUT -H "Authorization: Bearer $T" -H 'Content-Type: application/octet-stream' \
     --data-binary @/tmp/plain.xml 'https://localhost:55000/cluster/<node>/configuration'
# -> "Configuration was successfully updated in specified node", error 0

# the config is now 37 bytes, and the validator is happy
docker exec <node> cat /var/wazuh-manager/etc/wazuh-manager.conf
curl -sk -H "Authorization: Bearer $T" 'https://localhost:55000/cluster/configuration/validation'
# -> status OK

# and the node does not come back
docker restart <node>
docker logs <node> | grep -i "Invalid cluster configuration"

# restore
docker exec <node> cp /tmp/conf.bak /var/wazuh-manager/etc/wazuh-manager.conf
docker restart <node>
```

The root cause reproduces with no API and no deployment state at all:

```python
from wazuh.core.utils import load_wazuh_xml, validate_wazuh_xml
open('/tmp/plain.xml','w').write('esto no es una configuracion de wazuh')
r = load_wazuh_xml(xml_path='/tmp/plain.xml')
print(r.tag, repr(r.text), len(list(r)))   # root_tag 'esto no es una configuracion de wazuh' 0
validate_wazuh_xml('esto no es una configuracion de wazuh')   # returns, raises nothing
```

Antoniogm03:
## Description

`PUT /cluster/{node_id}/configuration` and `PUT /groups/{group_id}/configuration` both accepted configuration content that could never actually be used by the manager, and left the operator with no indication of the problem until the next restart:

- Text with no XML elements at all (plain text, or a comment-only body) parses as well-formed XML once wrapped in a synthetic root, so it silently replaced a group's `agent.conf`.
- A manager configuration with no `<cluster>` section at all, or no `<indexer>` section at all, passed schema validation (the section itself was never required, only fields *inside* it once you declared it), and `GET /cluster/configuration/validation` reported it as `OK`. The node then failed to start `wazuh-manager-clusterd` (missing `cluster.key`, which has no default) or crashed `wazuh-manager-analysisd` (zero indexer hosts configured) on the next restart.

Closes #38638

This branch was rebased onto the current `5.0.0` after #38819 replaced the manager's entire configuration-validation architecture (the old per-daemon `-t` checks this PR originally patched no longer exist — see "Proposed Changes" below). The rebase turned into a fresh investigation of what #38819 still left open, run and verified against a real installed 5.0.0 manager rather than only against the schema in isolation.

## Proposed Changes

**1. `PUT /groups/{group_id}/configuration` accepting content with no XML elements**
`upload_group_configuration()` (`framework/wazuh/core/configuration.py`) wraps the uploaded content in a synthetic `<root>` before parsing it with `defusedxml.minidom.parseString`; text or comments alone still parse fine inside that root. It now checks for at least one `ELEMENT_NODE` child and raises `WazuhError(1113)` if there is none.

**2. Manager configuration schema accepting a document missing `<cluster>` or `<indexer>` entirely**
`src/shared_modules/manager_config/schema/wazuh-manager.schema.json` (the JSON Schema `bin/wazuh-manager-conf` and every manager daemon validate the effective configuration against, introduced by #38819/#38816) declared `cluster.key` and `indexer.hosts` as required *within* their own sub-objects, but never required the `cluster` or `indexer` objects themselves at the document root. A document that omitted one of those sections entirely skipped that inner `required` check altogether, and the effective document ended up with the section built from defaults minus the one field that has none — `cluster` without `key`, `indexer` with `hosts: []`.

Added `"cluster"` and `"indexer"` to the schema's root-level `required` array. This does not force a real multi-node cluster or a reachable indexer to be configured: `node_type`, `nodes`, `port`, `bind_addr` and `hidden` inside `cluster` all keep their single-node defaults, and every 5.x manager already runs `wazuh-manager-clusterd` unconditionally (#31295 removed `cluster.disabled`) and needs at least one indexer host for `wazuh-manager-analysisd` to start — the installer already always writes both sections. The change only rejects, at `PUT`/`validate` time, the one case that previously slipped through: omitting the section altogether.

`indexer.hosts`' now-unreachable `default: []` was removed along with the outer `default: {"hosts": []}` on the `indexer` object, and both sections' schema descriptions were updated to say they are mandatory. Updated the ~20 test vectors under `tests/vectors/{valid,invalid}/`, the inline C++ literals in `managerConfig_test.cpp`, `manager_conf_cli_test.sh`, the `manager_config` `README.md`, and a framework fixture (`framework/wazuh/core/tests/data/configuration/wazuh-manager.conf` + `.effective.json`, and the assertion in `test_configuration.py` that hardcoded `indexer: {hosts: []}`) that all implicitly relied on a cluster-less or indexer-less document being valid.

### `PUT /groups/{group_id}/configuration`, on a real installed manager

```
PUT /groups/default/configuration, Content-Type: application/xml, body: this is not a wazuh configuration
{"title": "Bad Request", "detail": "XML syntax error: Error 1113 - XML syntax error: Configuration must contain at least one element", "error": 1113}
http=400

PUT /groups/default/configuration, body: <!-- just a comment -->
{"title": "Bad Request", "detail": "XML syntax error: Error 1113 - XML syntax error: Configuration must contain at least one element", "error": 1113}
http=400

PUT /groups/default/configuration, body: <agent_config><localfile>...</localfile></agent_config>
{"message": "Agent configuration was successfully updated", "error": 0}
http=200
```

### `PUT /cluster/{node_id}/configuration` and `GET /cluster/configuration/validation`, before the schema fix

```
PUT /cluster/node01/configuration, body: <wazuh_config></wazuh_config>
{"data": {"affected_items": ["node01"], ...}, "message": "Configuration was successfully updated in specified node", "error": 0}
GET /cluster/configuration/validation -> {"data": {"affected_items": [{"name": "node01", "status": "OK"}]}, "error": 0}
```

Restarting the manager on that configuration crashed `wazuh-manager-analysisd`:
```
wazuh-manager-analysisd: ERROR: An error occurred while initializing the modules: Could not initialize Indexer Connector: No hosts found in the configuration.
```
(A prior test on the same manager with `<cluster>` entirely absent instead left `wazuh-manager-clusterd` unable to start.)

### After the fix, same manager, same payload

```
PUT /cluster/node01/configuration, body: <wazuh_config></wazuh_config>
{"data": {..., "total_failed_items": 1, "failed_items": [{"error": {"code": 1130, "message": "Configuration does not match the schema: missing mandatory option (does not satisfy 'required') [schema ]"}, "id": ["node01"]}]}, "error": 1}
http=400

PUT /cluster/node01/configuration, body: <wazuh_config><cluster><key>...</key></cluster></wazuh_config>  (no indexer)
same 1130/required rejection
http=400

PUT /cluster/node01/configuration, body: <wazuh_config><cluster><key>...</key></cluster><indexer><hosts><host>https://127.0.0.1:9200</host></hosts></indexer></wazuh_config>
{"data": {"affected_items": ["node01"], ...}, "error": 0}
http=200
```

Restarting the manager on that last (minimal, now-accepted) configuration brought up all 7 daemons cleanly — including `wazuh-manager-analysisd`, which had died on the same reduced configuration before the `indexer` fix:
```
```
No error in `logs/wazuh-manager.log`, `logs/cluster.log`, `logs/api.log` — only the expected "indexer unreachable, retrying" warnings from `wazuh-manager-analysisd`, `wazuh-manager-modulesd` and `wazuh-manager-clusterd`, since there is no real indexer in the test environment. `GET /cluster/configuration/validation` correctly reported `OK`. A `curl` to the API answered `401` (listening, enforcing auth). The original full configuration was then restored via a real `PUT` and the manager restarted again to confirm no regression.

### Automated tests (devcontainer)

```
manager_config_utest:        20/20 passed
manager_config parity.py:     0 discrepancies (rapidjson validator vs. Python jsonschema reference)
manager_conf_cli_test.sh:    12/12 passed
framework pytest (test_configuration.py, test_manager_conf.py,
  test_manager_conf_policy.py, test_manager.py):   75/75 passed
```

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - N/A — no Windows-specific code touched
- Memory tests for macOS
  - N/A — no macOS-specific code touched

  - N/A — no decoder/rule code touched

  - N/A — no engine C++ source touched; the schema `wazuh-manager-analysisd` links is data, not engine logic

- Wazuh server API/Framework
  - [x] Run API Integration Tests _(manual, against a real installed manager — see Results and Evidence above; the automated `API Integration Tests` suite itself was not run)_

### Artifacts Affected

- `etc/wazuh-manager.schema.json` (installed from `src/shared_modules/manager_config/schema/wazuh-manager.schema.json`; packaged explicitly in `packages/rpms/SPECS/wazuh-manager.spec`)
- Every binary that embeds this schema at build time: `bin/wazuh-manager-conf`, `bin/wazuh-manager-remoted`, `bin/wazuh-manager-authd`, `bin/wazuh-manager-db`, `bin/wazuh-manager-modulesd`, `bin/wazuh-manager-analysisd` — all need a rebuild + reinstall to pick up the new `required` fields (confirmed in this session: the fix was invisible to the daemons until the manager was reinstalled)
- Python framework package: `framework/wazuh/core/configuration.py`

Not artifacts (dev-only, not packaged): `src/shared_modules/manager_config/README.md`, `src/shared_modules/manager_config/tests/**`, `framework/wazuh/core/tests/**`, `CHANGELOG.md`.

### Configuration Changes

No new configuration parameters and no default value changes. `cluster` and `indexer` are now mandatory top-level sections of `wazuh-manager.conf`, but every configuration the installer generates already includes both — this only rejects a configuration that a caller (typically the API) writes without them, which previously appeared to work and then broke the manager on the next restart.

### Tests Introduced

- `framework/wazuh/core/tests/test_configuration.py`: `upload_group_configuration()` rejecting content with no XML elements.
- `src/shared_modules/manager_config/tests/unit/managerConfig_test.cpp`: `Load.EffectiveDefaultsFromMinimalDocument` (renamed/updated), plus every existing schema/semantics test updated to declare `cluster`/`indexer` where the fix now requires them.
- `src/shared_modules/manager_config/tests/vectors/`: `empty-root.conf` moved from `valid/` to `invalid/` with its expected `{"pointer": "", "keyword": "required"}`; new `valid/minimal-valid.conf` documents the true minimal configuration; ~20 existing vectors updated to include `<cluster>`/`<indexer>` where the fix under test is unrelated to either, to keep each vector isolated to one violation.
- `src/shared_modules/manager_config/tests/cli/manager_conf_cli_test.sh`: updated to the renamed minimal vector.
- `src/shared_modules/manager_config/tests/parity.py`: no code change, but now exercises the new root-level `required` keyword across the updated vectors.

## Review Checklist

- [ ] Code changes reviewed
- [x] Configuration changes documented (see Configuration Changes — none required from operators)
- [ ] ...


src/shared_modules/manager_config/tests/unit/managerConfig_test.cpp
```diff
@@ -102,7 +102,11 @@ TEST(Load, EffectiveDefaultsFromEmptyRoot)
     EXPECT_TRUE(rapidjson::Pointer("/auth/force/disconnected_time/enabled").Get(effective)->GetBool());
     EXPECT_STREQ(rapidjson::Pointer("/wdb/backup/global/interval").Get(effective)->GetString(), "1d");
     EXPECT_EQ(rapidjson::Pointer("/vulnerability-detection/pageSize").Get(effective)->GetInt(), 100);
-    EXPECT_EQ(rapidjson::Pointer("/indexer/hosts").Get(effective)->Size(), 0u) << "absent indexer: empty hosts";
+    EXPECT_EQ(rapidjson::Pointer("/indexer/hosts").Get(effective)->Size(), 1u);
+    EXPECT_STREQ(rapidjson::Pointer("/indexer/hosts/0").Get(effective)->GetString(), "https://127.0.0.1:9200");
+    EXPECT_EQ(rapidjson::Pointer("/agent-upgrade/wpk_repository").Get(effective), nullptr)
+        << "no default: the module picks the repository by target agent version";
+    EXPECT_TRUE(rapidjson::Pointer("/agent-upgrade/enabled").Get(effective)->GetBool());
```
jotacarma90: **[S2 · blocking · introduced]** These three lines dereference a null pointer and crash the whole `manager_config_utest` binary. `agent-upgrade` appears 0 times in `wazuh-manager.schema.json`, and line 112 of this same test asserts `EXPECT_FALSE(effective.HasMember("agent-upgrade"))`, so `rapidjson::Pointer("/agent-upgrade/enabled").Get(effective)` returns `nullptr` and `->GetBool()` segfaults — GTest reports a crash, not a failure, so every other test in the binary is lost too.

`git blame` puts lines 107-109 on this PR's `70c69fa2e` and lines 111-112 on base commit `8d808748c`, which removed the section: the rebase reinstated them and the suite was not re-run. CI already shows it — `managerconfig-modules` logs `[ RUN ] Load.EffectiveDefaultsFromMinimalDocument` with no `[ OK ]`/`[ FAILED ]` line and `1 tests failed out of 3`, contradicting the PR body's `20/20 passed`.

```suggestion
    EXPECT_EQ(rapidjson::Pointer("/indexer/hosts").Get(effective)->Size(), 1u);
    EXPECT_STREQ(rapidjson::Pointer("/indexer/hosts/0").Get(effective)->GetString(), "https://127.0.0.1:9200");
```

etc/wazuh-manager.conf
```diff
@@ -9,6 +9,18 @@
...
+       etc/templates/config (real cluster key, real indexer hosts) whenever it can. -->
+  <cluster>
+    <key>0123456789abcdef0123456789abcdef</key>
+  </cluster>
+
+  <indexer>
+    <hosts>
+      <host>https://127.0.0.1:9200</host>
+    </hosts>
+  </indexer>
```
jotacarma90, line 22: **[S2 · blocking · introduced]** This ships a well-known `cluster.key` in the source tree and installs it unvalidated. `inst-functions.sh:1420` copies `${WAZUH_CONF_SRC}` = `../etc/wazuh-manager.conf` verbatim whenever `gen_wazuh.sh` cannot produce `../etc/wazuh.mc`, and that branch — unlike the generated one at line 1412 — never passes through `wazuh-manager-conf validate`, so a manager installed that way comes up with `cluster.key = 0123456789abcdef0123456789abcdef`, the same literal used in ~25 vectors in this PR. The module README you edit here states the rationale verbatim: "a well-known default key would be a shared secret".

Exposure is limited (the schema defaults put the cluster listener on `127.0.0.1` with `nodes: ["127.0.0.1"]`, and .deb/.rpm generate and validate their own conf and never touch this file), but the placeholder reads like a real value rather than one the operator must replace. Rather than shipping a secret to make a template schema-valid, make the manager fallback fail the way the package specs already do — then this block can go away entirely:

```sh
# src/init/inst-functions.sh, the else branch at line 1419
        else
            if [ "X${INSTYPE}" = "Xmanager" ]; then
                echo "ERROR: unable to generate ${WAZUH_CONF}; ${WAZUH_CONF_SRC} is a template, not an installable configuration."
                exit 1
            fi
            echo "WARNING: unable to generate ${WAZUH_CONF} with desired configurations, using default configurations from ${WAZUH_CONF_SRC}"
            ${INSTALL} -m 0660 -o root -g ${WAZUH_GROUP} ${WAZUH_CONF_SRC} ${INSTALLDIR}/etc/${WAZUH_CONF}
        fi
```

src/shared_modules/manager_config/schema/wazuh-manager.schema.json
```diff
@@ -5,6 +5,10 @@
   "description": "Single source of truth for the manager configuration: types, ranges, defaults and documentation of every option. Consumed by the manager_config C++ module (embedded) and by the Python framework (installed copy).",
   "type": "object",
   "additionalProperties": false,
+  "required": [
```
jotacarma90, line 8: **[S3 · non-blocking · introduced]** A violation of this root `required` produces a message that names neither the missing section nor a schema location, so the operator-facing error for the exact case this PR exists to fix is unactionable. For a root-level `required`, `GetInvalidDocumentPointer()` and `GetInvalidSchemaPointer()` are both empty: `schemaValidate.cpp:99-102` renders `missing mandatory option (does not satisfy 'required') [schema ]` (stray space included), `manager-conf.cpp:155` substitutes a meaningless `'/'`, and `manager_conf.py:83` then drops that `'/'` on purpose. Net result, visible in this PR's own evidence: `1130: Configuration does not match the schema: missing mandatory option (does not satisfy 'required') [schema ]` — `cluster` or `indexer`, no way to tell. Compare `cluster-without-key.conf`, which gives the useful `/cluster` + `/properties/cluster`.

This is also the text a daemon logs when an upgraded manager refuses to start: `inst-functions.sh:1399` only writes the conf when it is absent, so an existing section-less file survives the upgrade and every daemon then fails `w_mconf_load`. rapidjson's error object carries the absent names in a `missing` array; `innermostKeyword()` walks past it and discards them.

```cpp
// src/shared_modules/manager_config/src/schemaValidate.cpp, the "required" branch
else if (keyword == "required")
{
    message = "missing mandatory option (does not satisfy 'required')";
    // rapidjson lists the absent property names under the keyword's "missing" array.
    if (const auto* missing = rapidjson::Pointer("/required/missing").Get(validator.GetError());
        missing != nullptr && missing->IsArray() && !missing->Empty())
    {
        std::string names;
        for (const auto& name : missing->GetArray())
        {
            names += (names.empty() ? "" : ", ") + std::string(name.GetString());
        }
        message += ": " + names;
    }
}
```

src/shared_modules/manager_config/schema/wazuh-manager.schema.json
```diff
@@ -5,6 +5,10 @@
...
+    "cluster",
+    "indexer"
```
jotacarma90, line 10: **[S3 · blocking · aggravated]** This root `required` breaks a repo tool that works today. `src/remoted/remoted_module/tools/load_balancer/add_second_manager.sh:56-80` writes a node-2 `wazuh-manager.conf` containing only `<logging>` and `<remote>`; line 97 then runs `$NODE2/bin/wazuh-manager-remoted`, whose startup calls `w_mconf_load(cfgfile)` (`src/remoted/src/config.c:134`) → `mconf_load_ex()`, which validates the **whole** document, not just `/remote`, and returns `OS_INVALID` on any error. After this PR the tool dies with `(1244) ... missing mandatory option (does not satisfy 'required')` before the listener comes up and then spins through its 60 retries. The sweep updated 20 vectors and the C/Python fixtures but missed the only other place in the repo that writes a manager conf from scratch (the engine's own `base/test/.../managerConfig_test.cpp` already carries both sections and is unaffected).

```sh
# src/remoted/remoted_module/tools/load_balancer/add_second_manager.sh, before </wazuh_config>
  <cluster>
    <key>0123456789abcdef0123456789abcdef</key>
    <port>1526</port>
  </cluster>

  <indexer>
    <hosts>
      <host>https://127.0.0.1:9200</host>
    </hosts>
  </indexer>
```

src/shared_modules/manager_config/tests/vectors/expected/empty-root.json
```diff
@@ -0,0 +1 @@
+{"pointer": "", "keyword": "required"}
```
jotacarma90, line 1: **[S3 · non-blocking · aggravated]** This new root-level expectation is also tripped by three negative cases the sweep missed, so they now violate two constraints at once: `invalid/agent-upgrade-section.conf` (expects `/agent-upgrade` + `additionalProperties`), `managerConfig_test.cpp:210` (`<auth><use_password>true</use_password></auth>`, expects `/auth/use_password`) and `test_mconf-config.c:140` (`<auth><disabled>true</disabled></auth>`, expects `/auth/disabled`). They are the only negative cases that reach the schema stage without the mandatory sections — the other eight section-less vectors fail at the XML stage, which is legitimate.

The vector passes today by accident of key ordering, not by rapidjson's event ordering: `parity.py:160` sorts jsonschema errors by `(len(absolute_path), str(absolute_path))`, and root `additionalProperties` and root `required` have identical keys, so the stable sort preserves jsonschema's iteration order — which is the key order of the schema file, where `additionalProperties` (line 7) precedes `required` (line 8). Swapping those two keys, a purely cosmetic edit, repoints the vector to `("", "required")` and turns `manager_config_parity` red. Adding the two sections to the three cases restores the one-violation-per-vector policy the PR body states.

tests/integration/test_api/test_config/test_upload_configuration/data/test_cases/cases_upload_configuration.yaml
```diff
@@ -32,4 +40,4 @@
           </legacy>
         </remote>
       </wazuh_config>
-    expected_code: 200
+    expected_code: 400
```
jotacarma90: **[S3 · blocking · introduced]** This expectation is wrong: a rejected upload never comes back as HTTP 400. `update_manager_conf()` catches the `WazuhError` and records it as a failed item, and the controller returns `json_response(data)` with the default 200; nothing maps `error != 0` to 400. This PR's own `test_cluster_endpoints.tavern.yaml` run (green) asserts `status_code: 200` + `error: 1` + code 1130 for `invalid_wazuh_conf`, and the `test_indexer` cases updated in this same PR expect `expected_code: 200` with `expected_error_code: 1`. CI skips `tests/integration` for this PR (`Skipped IT Linux`), so this case will fail the first time the suite runs. The `http=400` lines in the Results section do not match the controller either.

Independently, the body violates two things (two `<wazuh_config>` roots → 1131, and neither root declares `<cluster>`/`<indexer>` → 1130), so even with the right status it would not pin the multi-root rejection its name describes. `UPLOAD_ONE_BLOCK_CONFIG` above did get both sections.

Suggested: back to 200, both sections in each root so the second root is the only violation left, and assert `error: 1` in the body (ideally code 1131), as `test_indexer` does:

```yaml
    body: >
      <wazuh_config>
        <remote>
          <legacy>
            <port>1514</port>
          </legacy>
        </remote>
        <cluster>
          <key>0123456789abcdef0123456789abcdef</key>
        </cluster>
        <indexer>
          <hosts>
            <host>https://127.0.0.1:9200</host>
          </hosts>
        </indexer>
      </wazuh_config>

      <wazuh_config>
        <remote>
          <legacy>
            <port>514</port>
          </legacy>
        </remote>
        <cluster>
          <key>0123456789abcdef0123456789abcdef</key>
        </cluster>
        <indexer>
          <hosts>
            <host>https://127.0.0.1:9200</host>
          </hosts>
        </indexer>
      </wazuh_config>
    expected_code: 200
```

framework/wazuh/core/configuration.py
```diff
@@ -816,6 +816,9 @@ def upload_group_configuration(group_id: str, file_content: str) -> str:
         raise WazuhError(1113, str(e))
 
     try:
+        if not any(n.nodeType == n.ELEMENT_NODE for n in xml.documentElement.childNodes):
+            raise WazuhError(1113, 'Configuration must contain at least one element')
```
jotacarma90: **[S3 · non-blocking · introduced]** A well-formed body with no elements is reported as an XML syntax error: `PUT /groups/default/configuration` with `<!-- just a comment -->` answers `"detail": "XML syntax error: Configuration must contain at least one element"`, telling the operator their valid XML has a syntax error. The parse succeeded; what failed is that the content is not a usable Wazuh configuration — which is exactly what `1114` means, and this same function already declares and raises it for `verify-agent-conf` rejections, so the endpoint's documented code set does not change. (`1112` is "Empty files are not supported" and a comment-only body is not empty.) The test assertion at `test_configuration.py:357` moves with it.

Placement itself is right after the commit-7 move: the check now sits inside the `try` whose `except` removes the temp file, so no `api_tmp_file_*.xml` leaks on a rejected PUT.

```suggestion
        if not any(n.nodeType == n.ELEMENT_NODE for n in xml.documentElement.childNodes):
            raise WazuhError(1114, 'Configuration must contain at least one element')
```

framework/wazuh/core/tests/test_configuration.py
```diff
@@ -352,6 +353,9 @@ def test_upload_group_configuration(mock_open, mock_wazuh_uid, mock_wazuh_gid):
             with patch('wazuh.core.configuration.open', return_value=Exception):
                 with pytest.raises(WazuhError, match=".* 1113 .*"):
                     configuration.upload_group_configuration('default', "<agent_config>new_config</agent_config>")
+            with patch('builtins.open'):
+                with pytest.raises(WazuhError, match=".* 1113 .*"):
+                    configuration.upload_group_configuration('default', "this is not a wazuh configuration")
```
jotacarma90, line 358: **[S3 · non-blocking · introduced]** Only the plain-text half of the new check is tested. `"this is not a wazuh configuration"` reaches `<root>` as a `TEXT_NODE`; the `<!-- just a comment -->` body the issue and the PR description both single out arrives as a `COMMENT_NODE` and exercises a different `nodeType` branch of the `any(...)` generator. A later refactor to `firstChild.nodeType != TEXT_NODE`-style logic would pass the committed test and reopen the comment-only hole.

```suggestion
            with patch('builtins.open'):
                with pytest.raises(WazuhError, match=".* 1113 .*"):
                    configuration.upload_group_configuration('default', "this is not a wazuh configuration")
            with patch('builtins.open'):
                with pytest.raises(WazuhError, match=".* 1113 .*"):
                    configuration.upload_group_configuration('default', "<!-- just a comment -->")
```

PR #38779, opened 2026-09-01
Fix macos upgrade mutex issues

### LINKED ISSUE #38690: wazuh-syscheckd logs CRITICAL 'At pthread_mutex_lock(): Invalid argument' on macOS agent during 4.x to 5.x upgrade
In the `5.0.0` `2026-08-28` nightly macOS agent upgrade tests, `wazuh-syscheckd` logged a `CRITICAL` error on `macos-26-arm64` during the 4.x to 5.x upgrade.

```
2026/08/27 21:50:50 wazuh-syscheckd: CRITICAL: At pthread_mutex_lock(): Invalid argument
```

`pthread_mutex_lock` returning `EINVAL` ("Invalid argument") means the mutex was not initialized (or had already been destroyed) when it was locked. This points to a thread/lifecycle defect in the FIM daemon, most likely hit during the `wazuh-syscheckd` restart that the upgrade performs (the process PID changed from 5514 to 5868 across the upgrade window). The daemon stayed up afterwards, but the message is logged at `CRITICAL`.

## Scope

- Version `5.0.0`, stage `beta5`, commit `d2c7219`.
- Seen only on `macos-26-arm64`, only in the regular macOS agent upgrade phase, in the `test_00_setup_upgrade` window. It does not appear in the macOS install phase, the macOS WPK phase, or any Linux / Windows / Central Components phase of this run.
- Not tracked in our nightly known-messages catalog, so it is treated as a new problem.

## Notes / limitations

- Only one macOS host was actually exercised in this run's upgrade phase (the other four failed to allocate due to macStadium capacity), so it is not yet clear whether this is specific to arm64 / macOS 26 or affects all macOS agents on upgrade.
- The captured logs contain only the `CRITICAL` line (the aggregated log-check collects `WARNING`/`ERROR`/`CRITICAL` lines), with no stack trace or surrounding `INFO` to pin down the exact code path. A reproduction with debug logging on a macOS agent upgraded from 4.x would help locate the mutex.

## Evidence

- Nightly incident: https://github.com/wazuh/devel-major-incidents/issues/234
- macOS upgrade build: https://jenkins-production.qa.wazuh.info/job/5.0.0/job/deployability_testing/1034/
  - [full_component_logs.zip](https://github.com/user-attachments/files/31550516/full_component_logs.zip)

jpcerrone:
## Description

During the nightly `5.0.0` macOS agent upgrade tests, `wazuh-syscheckd` logged a `CRITICAL` and aborted while shutting down on `macos-26-arm64`:

```sql
2026/08/27 21:50:50 wazuh-syscheckd: INFO: (1756): Shutdown received. Releasing resources.
2026/08/27 21:50:50 wazuh-syscheckd: CRITICAL: At pthread_mutex_lock(): Invalid argument
```

FIM keeps a set of internal synchronization primitives inside its global configuration. Those primitives were only made valid partway through FIM's startup, but two of FIM's threads — the shutdown waiter and the local-requests thread — are already running before that point, and the daemon can sit in that window for a long time while it waits for the agent's startup hash gate to be released and for the message queue to become available. During an upgrade the agent is restarted, so the daemon is waiting in exactly that window when the stop arrives, and the shutdown path then touches a primitive that has not been prepared yet.

This only surfaced on macOS because of a platform difference in the C library rather than anything specific to the macOS FIM code: on Linux an untouched, zero-filled primitive happens to be indistinguishable from a correctly prepared one, so the same defect is silently tolerated. On macOS it is rejected, and the daemon aborts.

Closes #38690

## Proposed Changes

FIM's internal synchronization primitives are now ready from the moment the process starts, instead of only once FIM reaches its initialization step. This removes the startup ordering dependency entirely rather than narrowing the window: a stop signal arriving at any point during startup now results in a clean shutdown on every platform.

There is one related behaviour change worth noting for review. On macOS, a stop arriving in this early window previously caused FIM to skip its database teardown, because the check guarding that teardown could not succeed against an unprepared primitive. With the primitives valid from the start, that teardown now runs. This is not new behaviour overall — it is what Linux has always done in the same situation — it simply now happens consistently across platforms.

The shutdown itself also becomes cleaner on that path. The abort previously terminated the process in a way that ran exit handlers and static destructors while other FIM threads were still alive, and skipped removing the daemon's PID file. The stop now completes through the daemon's normal termination path instead.

Behaviour once FIM has finished starting up is unchanged.

### Results and Evidence

Both runs use the same reproduction: stop the agent so `wazuh-agentd` is not available to release the startup hash gate, then start `wazuh-syscheckd` on its own and send it `SIGTERM` while it waits at the gate. This is the same state the upgrade restart produces, made deterministic.

**Before — `5.0.0` beta5, commit `8b2c64e`**

```sql
agent-macos-arm:~ root# cat /Library/Ossec/VERSION.json
{
    "version": "5.0.0",
    "stage": "beta5",
    "commit": "8b2c64e"
}
```

```sql
2026/09/01 16:00:39 wazuh-syscheckd[2333] syscom.c:724 at syscom_main(): DEBUG: (6284): Local requests thread ready.
2026/09/01 16:00:39 wazuh-syscheckd[2333] startup_gate_op.c:154 at startup_gate_wait_for_ready(): DEBUG: Startup hash gate is blocking 'wazuh-syscheckd' (waiting for agentd startup gate status).
2026/09/01 16:01:09 wazuh-syscheckd[2333] startup_gate_op.c:154 at startup_gate_wait_for_ready(): DEBUG: Startup hash gate is blocking 'wazuh-syscheckd' (waiting for agentd startup gate status).
2026/09/01 16:01:15 wazuh-syscheckd[2333] main.c:84 at fim_shutdown_waiter(): INFO: (1756): Shutdown received. Releasing resources.
2026/09/01 16:01:15 wazuh-syscheckd[2333] atomic.c:18 at atomic_int_get(): CRITICAL: At pthread_mutex_lock(): Invalid argument
```

The daemon aborts. With debug logging enabled the message carries the source location the nightly log could not show, confirming which primitive is involved.

**After — same host, with this change applied**

```sql
2026/09/01 17:08:26 wazuh-syscheckd[6817] syscom.c:724 at syscom_main(): DEBUG: (6284): Local requests thread ready.
2026/09/01 17:08:26 wazuh-syscheckd[6817] startup_gate_op.c:154 at startup_gate_wait_for_ready(): DEBUG: Startup hash gate is blocking 'wazuh-syscheckd' (waiting for agentd startup gate status).
2026/09/01 17:08:36 wazuh-syscheckd[6817] main.c:84 at fim_shutdown_waiter(): INFO: (1756): Shutdown received. Releasing resources.
2026/09/01 17:08:36 wazuh-syscheckd[6817] main.c:169 at fim_shutdown_waiter(): INFO: (1225): SIGNAL [(15)-(Terminated: 15)] Received. Exit Cleaning...
```

The daemon reaches the same shutdown point and exits cleanly. No `CRITICAL`.

**Normal startup, same build** — confirming the change does not affect the ordinary path, with the agent started against its manager:

```sql
Starting Wazuh v5.0.0...
Started wazuh-execd...
Started wazuh-agentd...
Started wazuh-syscheckd...
Started wazuh-logcollector...
Started wazuh-modulesd...
Completed.
```

The gate is released as usual and FIM starts normally.

### Artifacts Affected

- `wazuh-syscheckd` — all agent platforms. The observable behaviour change is limited to macOS agents; on Linux and Windows the defect was already tolerated by the C library, so behaviour on those platforms is unchanged.

No packaging, installer or default configuration changes.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

None, deliberately.

The legacy `cmocka` suites run only on Linux, across the agent, winagent and manager targets. The macOS unit-test job builds and runs the C++ suites, not the legacy ones, so no automated suite executes on the platform where this defect is observable.

That matters because the failure depends entirely on how a given C library treats never-initialized synchronization memory. On Linux it is indistinguishable from a correctly prepared one, so the failing condition cannot arise there at all: a test exercising this path passes identically with and without this change. In several of the relevant suites the lock call is additionally replaced at link time by a test double that always reports success, so the real primitive is never exercised regardless of platform.

It would be possible to assert instead on the raw byte representation of the initialized state, which would fail on the winagent target if the initialization were ever removed. That checks a representation detail rather than the behaviour that broke, gives no signal on the platform that actually failed, and was judged not worth the maintenance cost.

Verification is therefore the before/after evidence above, captured on macOS.

No existing tests required changes: no suite asserts on the initialization calls that were removed.

## Review Checklist

- [x] Code changes reviewed
- [ ] Tests cover the new functionality — see *Tests Introduced* for why no test is added
- [x] Configuration changes documented — none
- [x] Developer documentation reflects the changes — none required

### DIFF
--- a/src/syscheckd/src/syscheck.c
+++ b/src/syscheckd/src/syscheck.c
@@ -30,7 +30,20 @@
 #endif
 
 // Global variables
-syscheck_config syscheck;
+// The synchronization primitives are initialized here rather than in fim_initialize(): the
+// shutdown waiter and the syscom thread both start before it runs.
+syscheck_config syscheck = {
+    .fim_scan_mutex = PTHREAD_MUTEX_INITIALIZER,
+    .fim_realtime_mutex = PTHREAD_MUTEX_INITIALIZER,
+#ifdef WIN32
+    .fim_registry_scan_mutex = PTHREAD_MUTEX_INITIALIZER,
+#else
+    .fim_symlink_mutex = PTHREAD_MUTEX_INITIALIZER,
+#endif
+    .fim_pause_requested = ATOMIC_INT_INITIALIZER(0),
+    .fim_pausing_is_allowed = ATOMIC_INT_INITIALIZER(0),
+    .fim_first_sync_completed = ATOMIC_INT_INITIALIZER(0),
+};
 int notify_scan = 0;
 int sys_debug_level;
 int audit_queue_full_reported = 0;
@@ -492,15 +505,6 @@ bool fetch_document_limits_from_agentd(){
 }
 
 void fim_initialize() {
-    // Initialize the coordination atomics first, before any early return below.
-    // On Windows (winpthreads) PTHREAD_MUTEX_INITIALIZER is a non-zero sentinel,
-    // so a zero-initialized global mutex is invalid: if fim_db_init() fails and we
-    // return early, the disabled path in start_daemon() would call atomic_int_set()
-    // on an uninitialized mutex and abort. They are also read by the syscom thread.
-    syscheck.fim_pause_requested = (atomic_int_t)ATOMIC_INT_INITIALIZER(0);
-    syscheck.fim_pausing_is_allowed = (atomic_int_t)ATOMIC_INT_INITIALIZER(0);
-    syscheck.fim_first_sync_completed = (atomic_int_t)ATOMIC_INT_INITIALIZER(0);
-
     // Create store data
 #ifndef WIN32
     FIMDBErrorCode ret_val = fim_db_init(FIM_DB_DISK,
@@ -546,16 +550,9 @@ void fim_initialize() {
         return;
     }
 
-    // Initialize locks before sync handle creation
+    // Initialize directories_lock before sync handle creation
     w_rwlock_init(&syscheck.directories_lock, NULL);
     syscheck_set_directories_lock_ready();
-    w_mutex_init(&syscheck.fim_scan_mutex, NULL);
-    w_mutex_init(&syscheck.fim_realtime_mutex, NULL);
-#ifdef WIN32
-    w_mutex_init(&syscheck.fim_registry_scan_mutex, NULL);
-#else
-    w_mutex_init(&syscheck.fim_symlink_mutex, NULL);
-#endif
 
     notify_scan = syscheck.notify_first_scan;
 


fcontrerasc (approved):

src/syscheckd/src/syscheck.c
```diff
@@ -546,16 +550,9 @@ void fim_initialize() {
         return;
     }
 
-    // Initialize locks before sync handle creation
+    // Initialize directories_lock before sync handle creation
     w_rwlock_init(&syscheck.directories_lock, NULL);
     syscheck_set_directories_lock_ready();
```
MarcelKemp, line 555: **[S2 · non-blocking · pre-existing]** `directories_lock` is the one primitive left out of this fix, so the `syscheck_directories_lock_ready` workaround has to stay — and with it a permanent defect: when `fim_db_init()` fails, `fim_initialize()` returns at line 523 before ever reaching here, the flag stays 0 for the life of the process, and `getSyscheckConfig()` returns NULL at `config.c:134-136` forever, so `syscom_getconfig("syscheck")` answers `err` and the agent's `/config` document never carries the FIM section even though the configuration is fully parsed in memory. `PTHREAD_RWLOCK_INITIALIZER` — already used for `fim_sync_handle_rwlock` at `run_check.c:89`, so proven on winagent — removes the ordering dependency the same way this PR does for the mutexes and lets the whole flag go.

Not introduced by this PR — flagging it for a follow-up issue, not for this review. It is worth raising here because this PR is the natural place to finish the job, and because it leaves two comment blocks factually false: `syscheck-config.h:476-482` still claims "the whole `syscheck` struct is zero-initialized until then" and that "an atomic_int_t here would need its OWN mutex initialized to be read", and `syscheckd/src/config.c:124-127` repeats it — `syscheck.fim_pause_requested` now proves the opposite.

The change does not fit these lines: add `.directories_lock = PTHREAD_RWLOCK_INITIALIZER,` to the initializer above, delete lines 553-555, and delete the flag, its two accessors and its comment in `config/include/syscheck-config.h:471-515` plus the definition at `syscheckd/src/config.c:126` and the guard at `config.c:134-136`.

src/syscheckd/src/syscheck.c
```diff
@@ -30,7 +30,20 @@
...
+    .fim_realtime_mutex = PTHREAD_MUTEX_INITIALIZER,
+#ifdef WIN32
+    .fim_registry_scan_mutex = PTHREAD_MUTEX_INITIALIZER,
+#else
+    .fim_symlink_mutex = PTHREAD_MUTEX_INITIALIZER,
+#endif
+    .fim_pause_requested = ATOMIC_INT_INITIALIZER(0),
+    .fim_pausing_is_allowed = ATOMIC_INT_INITIALIZER(0),
+    .fim_first_sync_completed = ATOMIC_INT_INITIALIZER(0),
+};
```
MarcelKemp, line 46: **[S3 · non-blocking · introduced]** Nothing pins this invariant, so adding a lock to `syscheck_config` and forgetting the field here would go unnoticed. Your reasoning that a Linux test cannot observe it is right — glibc's initializer is all-zero, so neither a trylock nor a `memcmp` distinguishes the two states — but the existing `TEST_WINAGENT` target can: winpthreads returns `EINVAL` on a zero-filled `pthread_mutex_t`, so a behavioural test needs no byte-level assertions. `test_fim_initialize` (`test_syscheck.c:101-133`) would keep passing with this initializer deleted: it sets no expectation on `pthread_mutex_init` and no suite asserts on the removed calls.

Goes in `src/unit_tests/syscheckd/test_syscheck.c`, before `test_fim_initialize` in the cmocka array:

```c
#ifdef TEST_WINAGENT
// Fails without the static initializer in syscheck.c: on winpthreads a zero-filled
// pthread_mutex_t (what the BSS left before fim_initialize()) makes w_mutex_lock()
// merror_exit with EINVAL, which is exactly the abort in #38690.
void test_syscheck_pause_paths_work_before_fim_initialize(void **state) {
    (void) state;

    expect_string(__wrap__mdebug1, formatted_msg, "FIM agent info: pause command received");
    expect_string(__wrap__mdebug1, formatted_msg, "FIM pause requested (async), fim_pause_requested=1");

    assert_int_equal(fim_execute_pause(), 0);

    fim_syscom_cleanup_pause();

    assert_int_equal(atomic_int_get(&syscheck.fim_pause_requested), 0);
}
#endif
```

MarcelKemp (approved):

Issue #38741, opened 2026-08-31
wazuh-remoted: fix remoted_module_utest ASan failures


matigarciadev:
## Summary

`remoted_module_utest` has failed in CI under AddressSanitizer with two distinct, unrelated memory-safety bugs (different ASan checks, different root causes, different files).

---

## Finding 1 — `FakeUdsServer` detaches per-connection threads → **stack-use-after-return**

CI failure, `WazuhDBClientTest.SuccessfulRoundTripObservesLatencyOnce` (2026-08-27T22:03:01Z run):

```
==6965==ERROR: AddressSanitizer: stack-use-after-return on address 0x7f8296a93590
READ of size 1 at 0x7f8296a93590 thread T2268
    #0 std::atomic<bool>::load(std::memory_order) const
    #2 remoted::test::FakeUdsServer::serveConnection(int) .../fakeUdsServer.hpp:157
    #3 remoted::test::FakeUdsServer::acceptLoop()::{lambda()#1}::operator()() const .../fakeUdsServer.hpp:151
...
Address 0x7f8296a93590 is located in stack of thread T0 at offset 1424 in frame
    #0 WazuhDBClientTest_QueryTimeoutFailsWithTimeoutAndIncrementsMetric_Test::TestBody() .../wazuhDBClient_test.cpp:441
    [1344, 1440) 'server' (line 443) <== Memory access at offset 1424 is inside this variable
Thread T2268 created by T2263 (the FakeUdsServer::acceptLoop thread) here:
    std::thread::thread<...acceptLoop()::{lambda()#1}...>
Thread T2263 created by T0 here:
    std::thread::thread<...FakeUdsServer::FakeUdsServer(...)::{lambda()#1}...>
```

Note the failing read is `m_stopping.load()` at `fakeUdsServer.hpp:157`, called from a `serveConnection` thread that `acceptLoop()` spawned — and that thread is reading into the stack frame of a *different, already-returned* test (`QueryTimeoutFailsWithTimeoutAndIncrementsMetric`'s `server` local), while the currently-running test is `SuccessfulRoundTripObservesLatencyOnce`. That cross-test detail is itself proof of the bug: the detached thread outlived its own test entirely and was still alive when the *next* test's stack reused that memory.

### Root cause

- `FakeUdsServer::acceptLoop()` spawns `std::thread([this, fd] { serveConnection(fd); }).detach()` for every accepted connection.
[...]

Issue #33, opened 2016-06-14
Invalid query in Kibana visualization "High Risk Alerts / PCI DSS"


tikhon-clearscale:
Hello,

Looks like I found small mistake in file kibana-ossecwazuh-dashboards.json, string 574 - query is a "rule.AlertLevel > 10", but I think it should be "rule.AlertLevel: { 10 TO *}".

snaow:
Hi @tikhon-clearscale,

You are totally right, we already fixed it (https://github.com/wazuh/ossec-wazuh/commit/d67a85419ec1f8f3feeb70357f9e6daf34d38b23). We set the inclusive brackets to include alert level 10 in the query, so the final query will be "AlertLevel greater than or equal to 10".

For the next release we are preparing a new set of dashboards and visualizations.

Thanks you so much for your contribution, we really appreciate it.

Best regards,

Pedro S.

tikhon-clearscale:
Thanks a lot!

PR #38781, opened 2026-09-01
Fix/38778 staging sync connector several fixes

### LINKED ISSUE #38778: Staging sync connector: false success on unvalidated indexer responses, broken 413 split, unbounded retries and five related defect groups
## Description

Static review of the staging sync connector merged by #38620 (`IndexerConnectorSync` + `inventory_sync_server`), centered on the `POST /stateful` pipeline, the workers' private sync connectors and the vulnerability-detection lane. Code references are against `5.0.0` at `c640af56a0`; symbol names are stable if lines drift.

Eight work packages, ordered by severity and dependency. They produce two failure classes:

- **Apparent success without recovery** (WP-2, WP-6): the caller gets `200` for work that was not applied, so nothing ever retries it.
- **Retry amplification / blocked shards** (WP-1, WP-3, WP-8): duplicated or unbounded work that a healthy fleet turns into load storms — the same amplifier measured in #38549 as the resync 500 storms (item 53).

Overlaps with #38549 are noted per item so a policy half already tracked there is not decided twice.

---

### WP-1 — Recursive `413` split reuses absolute offsets from the full buffer

`processBulkChunk(data, boundaries)` (`src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp:670`) receives a view of the current chunk, but when a chunk itself gets `413` and splits again, the halves are rebuilt from `m_bulkData.data()`/`m_bulkData.size()` with boundaries that stayed absolute (`:700`, `:704`). For any chunk that does not start at offset 0 this resends a prefix that belongs to another chunk, extends the second half to the end of the whole buffer, can fail to shrink (hitting `413` again), and can report "single operation too large" for a view holding several operations. The first split level (`:632`, `:635`) is correct only because there `data` spans the full buffer.

The existing `413` tests assert call counts and buffer state, not conservation of IDs across recursion — which is why this survived.

**Proposed fix:** derive `chunkStart = data.data() - m_bulkData.data()` and bound both halves to `[chunkStart, chunkStart + data.size())` (or switch the recursion to an operation range `[begin, end)` and materialize views only at POST time). Assert the invariant `chunkStart <= splitPos <= chunkEnd`.

**Tests:** unit test with ≥4 unique-ID docs and scripted responses (full bulk → 413, both halves → 413, leaves → 200) whose oracle checks **every `_id` appears exactly once across accepted bodies** and no body carries bytes outside its chunk; odd doc counts (5, 7) for the asymmetric midpoint; "single operation too large" only fires for true single-op views. Run the suite under ASAN; TSAN smoke reusing the conservation-under-race harness added by `9195fecdc5`.

### WP-2 — Fail-closed validation of indexer responses (`_delete_by_query`, `_update_by_query`, bulk 200s)

Three places confirm work the indexer did not acknowledge:

1. `onErrorDeleteByQuery` returns normally on HTTP `409` and `429` (`indexerConnectorSyncImpl.hpp:413-441`): the staged query is cleared, `flush()` succeeds, and the caller can answer `200` for a purge that did not run. A `Cleans` session then repopulates over the old state, and a deleted agent's documents survive with no retry signal.
2. `executeUpdateByQuery` (`:981`) logs `failures`/`version_conflicts`/unparseable bodies in its success callback but returns normally, so metadata/group updates confirm while documents keep stale values.
3. A `_bulk` 200 without an `errors` field, and a `_delete_by_query` 200 whose body does not parse, are treated as clean success (`validateBulkResponse`, `:154`).

Overlap with #38549: same handler as item 53, whose triage fixed the constraint — the shared success handler cannot be relaxed globally (item 47's contract), so classification must be **per caller**. What status the agent ultimately sees belongs to the "what the caller is told when a downstream call fails" follow-up (items 10, 39, 47, 51, 53); this WP is the mechanical half.

**Proposed fix:** extend the flag-and-throw pattern `_bulk` already uses: callbacks record, the function throws after `post()` returns. `429` → retryable (feeds WP-3's budget); HTTP-level `409` → operation failure (with `conflicts=proceed` active it is anomalous and must not confirm); `404` stays a no-op. Mandatory-field validation before any request is marked successful; never clear staged queries on an unvalidated response.

**Tests:** unit — staged `deleteByQuery` → `429`/`409` → `flush()` throws and the query is not lost; `_update_by_query` with `failures`, with `version_conflicts`, with invalid JSON → throws; bulk 200 missing `errors` → error. Pipeline-level with the connector fake throwing at each point → agent response ≠ 200. TSAN over the multithreaded suite (the new throws unwind while workers hold the connector mutex).

Note for the release: scenarios that today "succeed" silently will start failing visibly — that is the honest signal, but it raises retry pressure until WP-3 lands, so WP-2 and WP-3 should ship together.

### WP-3 — Total retry budget for `_bulk`, split chunks and `_update_by_query`

The retry loops (`_bulk` `:553-594`, chunks `:743-777`, `_update_by_query` `:1108-1146`) run while `needToRetry` holds, with no attempt cap and no overall deadline; `max_retry_delay_seconds` bounds only the pause between attempts. A persistent `429` or a transport-level failure (`status <= 0`) keeps a worker, its shard, and the connector mutex busy indefinitely; bodies and responders stay alive; UDS answers `504` while the worker keeps going and eventually indexes — so the agent can resend a session that is still being applied. (#38549 item 8 covered the per-request timeout, which exists — ~60 s; item 13 documented that the budget reservation is held while the worker is blocked. Neither bounds total retries.)

**Proposed fix:** a double budget per flush operation — max attempts + total deadline, whichever first — with the default sitting **below** `response_timeout` and `remoted.downstream_stateful_response_timeout`; honor `Retry-After` when present (capped by the existing backoff ceiling); budget exhaustion surfaces as `IndexerConnectorException` so the pipeline fails the batch instead of hanging it. Sizing the value belongs to the "timeout layers" follow-up of #38549 (items 19, 21, 31, 32, 50, 56, 58, 59); the mechanism does not need to wait for it.

**Tests:** unit — always-`429` mock with budget N → exactly N POSTs then throw; same for transport errors; a short budget returns error in less than `response_timeout`; the chunks of a `413` split share one budget per flush. `Retry-After` support moved to #38942 (the transport exposes no response headers). TSAN mandatory: shutdown during an in-flight backoff (the backoff sleeps on a dedicated thread while `processBulk()` holds the mutex, `:329`) must wake and exit without deadlock; concurrent stagers + a flusher in retry. QA: indexer paused with SIGSTOP → session fails within budget, worker resumes its shard.

### WP-4 — Host selection advances the round-robin cursor as a side effect

`processBulk()` calls `getNext()` before knowing whether there are deletes (`:360`) and discards that pick in a pure-bulk flush, selecting again for `/_bulk` (`:569`); `isAvailable()` also selects through `getNext()` (`src/shared_modules/indexer_connector/src/serverSelector.hpp:102`), so availability checks skew future distribution. Certain host counts and flush patterns concentrate bulks on a subset of hosts.

**Proposed fix:** `isAvailable()` inspects state without moving the cursor; a host is selected exactly once, at the point a request is emitted.

**Tests:** unit — 2 and 4 healthy hosts across mixed flush sequences → per-host counts uniform (±1), no skipped positions; repeated `isAvailable()` calls leave the subsequent selection sequence unchanged.

### WP-5 — One knob with two units, and an accepted-but-ignored option

`indexer_sync_max_bulk_size` is simultaneously the pipeline's group-commit threshold in accumulated FlatBuffer bytes (`src/wazuh_modules/inventory_sync_server/src/inventorySyncServerFacade.hpp:879`) and the connector's estimated NDJSON byte cap (`src/wazuh_modules/inventory_sync_server/src/indexer/indexerConnectorConfig.cpp:52`). Manager-added metadata and versioned scripts make the connector flush before the pipeline expects, one metric group commit can hide several real `_bulk` requests, and latency/size tuning becomes unpredictable. Separately, `inventory_sync_server_indexer_sync_flush_interval_seconds` is accepted but the facade forces it to 0 (`inventorySyncServerFacade.hpp:897`) — correct for flush ownership, silent for the operator.

**Proposed fix:** keep the existing option as the pipeline threshold and give the connector cap its own (derived) option; export the connector's real request/byte counts. WARN at startup when a non-default flush interval is configured, document it as ignored/deprecated. Same pattern #38628 c1/c5 established for refused/bounded options.

**Tests:** config unit tests for independent limits and the derived default; log-recorder test that the WARN fires exactly when the value differs from the default; QA check that `sync.indexer.bulk.requests` counts real `_bulk` requests (splits and retries included) while `sync.bulk.flushes` keeps counting group commits; their ratio is the split/retry amplification.

### WP-6 — Agent purge cannot see documents written inside the refresh interval

`_delete_by_query` only sees refreshed segments, and the server does not `_refresh` first because the manager's role lacks `indices:admin/refresh` — the limitation is documented in the code itself (`src/wazuh_modules/inventory_sync_server/src/sync/sessionProcessor.cpp:363-370`). An agent deleted right after its last accepted write leaves documents that nothing will ever overwrite, while the operation reports completed. This is #38549 item 47's window, and the privilege is the same follow-up as items 7/46 (*the manager's least-privilege indexer role*, owned by wazuh-indexer).

**Proposed fix:** once the privilege lands, `_refresh` the target indices before the purge's `_delete_by_query` (the `?refresh` parameter of the delete itself does not help — it refreshes after deleting). Until then: deferred re-purge after ≥ `index.refresh_interval` as mitigation. Contribute the `indices:admin/refresh` requirement to the role issue when it opens so the role is not iterated twice.

**Tests:** QA — bulk accepted → immediate agent delete → zero documents of that agent after the purge (currently a documented skip); unit — refresh precedes delete and a refresh failure is not swallowed (WP-2 semantics).

### WP-7 — No global order between the sync pipeline and the VD lane

`AgentInFlightRegistry` guarantees mutual exclusion between lanes, but the lanes queue independently with no shared sequence: the worker that acquires first wins, not the request that arrived first. If the emitter allows two concurrent sessions of one agent, a later session can apply before an earlier one — versioned documents mask part of it, but deletes, checks and unversioned operations can observe reordering.

**Proposed fix:** first confirm the emitter's contract with the agent team (can one agent ever hold a pipeline session and a VD session concurrently?). If it cannot: state the invariant in `docs/ref/` and pin it with a test — the same route #38727 took for the AR cursor contract. Only if concurrency is real: a per-agent monotonic sequence assigned at a common admission point.

**Tests:** unit — directed cross-lane race for one agent → exclusion holds (extends the registry tests); TSAN + stress: N threads × M agents contending across both lanes; QA scenario chaining a pipeline session and a VD scan of the same agent.

### WP-8 — Group-commit blast radius and idempotency as a contract

One failing session fails its whole staged batch: every agent in it gets `500`/`503` (`src/wazuh_modules/inventory_sync_server/src/sync/syncPipeline.cpp:271-282`, `flushAndRespond` `:290`) and resends a full session even when part of the bulk was accepted. Combined with WP-1/WP-3 this is the amplifier behind the storm #38549 measured (item 53: 0.4–0.9 % of sessions on reference hardware). Today the amplification is not even measurable per batch.

**Proposed fix:** per-group-commit metrics (sessions, bytes, failure cause, retries per session); for deterministic mapping errors only, exclude the offending session and retry the rest of the batch once — no general partial acknowledgements. Promote idempotency from emergent property to stated contract (deterministic `_id`s, versioned upserts, ID deletes), with a replay test enforcing it, since full-session retries are the designed recovery.

**Tests:** unit — mixed-agent poison batch → predictable per-session outcomes, no false `200`, rest of the batch enters on the second attempt; full-session replay after partial success → identical final state, no duplicates (reuses WP-1's conservation oracle); TSAN over concurrent failing/retrying batches.

---

## Dependencies and order

WP-1 and WP-2 first (independent, critical). WP-3 requires WP-2 (an exhausted budget is only useful if exhaustion is reported fail-closed) and ships in the same wave to contain the visible-failure increase WP-2 introduces. WP-4/WP-5/WP-7 are independent and parallel. WP-6 is gated cross-repo by the indexer-role follow-up. WP-8 last: its containment choices depend on WP-2/WP-3 semantics.

Every WP runs the `indexer_connector` and `inventory_sync_server` unit suites under **ASAN and TSAN** before merge — the module is multithreaded by design (per-shard workers, VD lane, the dedicated backoff thread) and several fixes touch exactly those crossings; the `TSANITIZE` infrastructure and a conservation-under-race oracle already exist in-tree (`9195fecdc5`).

## Out of scope

- Shutdown/restart convergence validation (staged sessions answered `503` on stop) — QA matrix, tracked with #38299's restart rows and the `Retry-After` decision of #38549 items 39/63/64.
- Hot-shard skew (hash-based shard distribution) — observability and tuning guidance along #38284; no redesign without evidence.
- The indexer role definition and index templates themselves (wazuh-indexer, #38549 items 7/46) — WP-6 consumes the privilege, it does not ship it.
- The protocol decision of which status/`Retry-After` the agent sees on terminal rejections (#38549 follow-up); WP-2/WP-3 keep their mechanical halves compatible with either outcome.

## Acceptance criteria

- [x] WP-1: under scripted nested `413`s, every `_id` is accepted exactly once across all bulk bodies; suite green under ASAN.
- [x] WP-2: `deleteByQuery` → `429`/`409`, `_update_by_query` with `failures`/conflicts/invalid body, and bulk 200s missing mandatory fields all fail the flush and the caller never sees `200`.
- [x] WP-3: with the indexer answering `429` indefinitely, the worker returns an error before `response_timeout`; shutdown during backoff neither deadlocks nor leaks (TSAN clean).
- [x] WP-4: 2- and 4-host distributions are round-robin with no skipped hosts; availability checks do not perturb selection.
- [x] WP-5: pipeline and connector limits tune independently; a non-default flush interval logs a WARN; bulk metrics count real requests.

## Discarded:

- WP-6, WP-7, WP-8

LucioDonda:
## Description

The `IndexerConnectorSync` / `inventory_sync_server` staging-sync stack (merged into `5.0.0` via #38620) carried a set of audited reliability risks: a recursive `413` bulk-split that resliced the whole staged buffer instead of the failing chunk (duplicating already-accepted operations and forging the "single operation too large" signal), indexer responses whose failures were silently swallowed and reported as success (`_delete_by_query` `409`/`429`, `_update_by_query` `failures`/`version_conflicts`, unvalidatable `200` bodies), unbounded retry loops that could pin a flushing worker — and the shard behind it — forever, host selection with side effects that skewed traffic across indexer nodes, one configuration knob silently driving two different byte thresholds, and group-commit failures whose blast radius was invisible to operators.

This PR fixes those defects and makes the remaining behavior observable and honest: every indexer response is validated fail-closed, every retry loop carries a budget (attempts cap + wall-clock deadline), host selection is side-effect free, the pipeline group-commit threshold and the connector's `_bulk` request cap are separate options, and new metrics expose the real `_bulk` traffic and the cause and blast radius of failed group commits. The module's idempotency guarantee — the property that makes its fail-and-re-POST recovery safe — is now a documented, test-enforced contract.

Closes #38778

## Proposed Changes

- **Recursive `413` bulk split (fix)** — `processBulkChunk()` now slices the failing chunk itself (`data.substr`) instead of rebuilding both halves from the whole staged buffer, with an invariant check that the split boundary falls inside the chunk. Ends duplication of already-accepted operations and prefix resends; the "single operation exceeds server limits" error can now only fire on a true single-operation frame.
- **Fail-closed response validation (fix)** — `_delete_by_query`: `404` stays a no-op, `429` retries with backoff, anything else (request-level `409` included) fails the flush; a `200` body that cannot be parsed, or that reports shard `failures`/`version_conflicts`, fails instead of confirming. `_update_by_query`: `failures`, `version_conflicts`, missing `updated`/`total` counters, or an unparseable body fail the operation. A `_bulk` `200` without the `errors` field fails. Callers are told the truth and retry; nothing is reported as applied without a validated response.
- **Bounded retry budget (feat)** — new `IndexerRetryBudget` (failed-attempts cap + wall-clock deadline, whichever is spent first; a sleep never starts past the deadline) applied to all six sync retry loops (`_bulk`, split chunks, `_delete_by_query`, `_update_by_query`, both search paths); the chunks of a `413` split, nested ones included, draw from their flush's budget, so a split cannot multiply it. Defaults 5 attempts / 15 s — below remoted's 20 s downstream response window, so a worker fails while its caller is still listening. Configurable per module (`inventory_sync_server_indexer_sync_max_retry_{attempts,duration_seconds}` for Inventory Sync, `indexer_max_retry_{attempts,duration_seconds}` for the Vulnerability Scanner) and per connector JSON (`0` disables a bound).
- **Side-effect-free host selection (fix)** — `TServerSelector::isAvailable()` is a pure peek (it no longer advances the round-robin cursor, so availability probes — the `/stats` and `/config` handlers, the pipeline's dispatch gate — stop skewing traffic), and `processBulk()` no longer pre-selects a host before knowing what the flush contains; every request, delete-by-query retries included, resolves its host at the point it is sent. Traffic now distributes uniformly across healthy nodes.
- **Honest bulk-size configuration (feat)** — `inventory_sync_server_indexer_sync_max_bulk_size` keeps only its pipeline role (group-commit threshold, wire FlatBuffer bytes); the connector's `_bulk` request cap (serialized NDJSON bytes) is the new `inventory_sync_server_indexer_sync_connector_max_bulk_size` (default 10 MiB). Raising the group-commit threshold can no longer push a single request past the indexer's `http.max_content_length`. A tuned-but-ignored `indexer_sync_flush_interval_seconds` now logs a startup `WARNING`.
- **Observability (feat)** — the sync connector counts every real `_bulk` POST (attempts, splits, retries) exposed through `takeBulkRequestStats()` and surfaced as `sync.indexer.bulk.requests` / `sync.indexer.bulk.bytes.total`; failed group commits are counted by cause (`sync.bulk.flush.failures.{documents,exhausted,other}`, backed by a new failure category on `IndexerConnectorException` mirrored across the module seam as `ConnectorError`) together with the sessions they punished (`sync.bulk.sessions.failed`).
- **Idempotency contract (docs + test)** — architecture.md now states the pipeline's idempotency guarantee as a contract (deterministic ids, `external_gte` versioned upserts, re-runnable by-query paths) with the rule for future non-idempotent operations; a replay unit test enforces it.

Known limits, deliberate: the indexer's `Retry-After` header is not honored (the `http-request` transport exposes no response headers — needs a companion `wazuh-http-request` PR, tracked in #38942); the restore of `_refresh`-before-purge stays blocked on the `indices:admin/refresh` privilege of the manager's indexer role (wazuh-indexer repo); selective poison-batch isolation is deferred to a follow-up PR (the typed failure categories added here are its base).

### Results and Evidence

<details><summary>Unit suites — normal build (Linux, Debug)</summary>
<p>

```text
$ ./bin/indexer_connector_utest --gtest_brief=1
[==========] 308 tests from 12 test suites ran.
[  PASSED  ] 308 tests.

$ ./wazuh_modules/inventory_sync_server/test/inventory_sync_server_utest --gtest_brief=1
[  PASSED  ] 235 tests.
```

</p>
</details>

<details><summary>AddressSanitizer (FSANITIZE=ON, detect_leaks=1)</summary>
<p>

```text
$ ASAN_OPTIONS=detect_leaks=1 ./bin/indexer_connector_utest --gtest_brief=1
[  PASSED  ] 308 tests.

$ ASAN_OPTIONS=detect_leaks=1 ./wazuh_modules/inventory_sync_server/test/inventory_sync_server_utest --gtest_brief=1
[  PASSED  ] 235 tests.
```

No AddressSanitizer or LeakSanitizer reports.

</p>
</details>

<details><summary>ThreadSanitizer (TSANITIZE=ON)</summary>
<p>

```text
$ setarch $(uname -m) -R ./bin/indexer_connector_utest --gtest_brief=1
[  PASSED  ] 308 tests.   # 2 pre-existing data-race warnings, both in indexerConnectorAsync_test.cpp
                          # test lambdas (lines 746 / 824, HandleError409/429) -- async connector test
                          # code outside this PR's scope; the sync connector and selector are clean.

$ setarch $(uname -m) -R ./wazuh_modules/inventory_sync_server/test/inventory_sync_server_utest --gtest_brief=1
[  PASSED  ] 235 tests.   # 0 ThreadSanitizer warnings
```

</p>
</details>

<details><summary>Red/green verification of the new failure tests</summary>
<p>

Against the pre-fix code (fixes stashed, tests kept):

- `Recursive413SplitsDeliverEveryDocumentExactlyOnce` FAILS (documents lost/duplicated) and
  `ConcurrentStagersWith413SplitsDeliverEveryDocumentExactlyOnce` aborts with an escaped
  "Single operation exceeds server limits" — the WP-1 conservation oracle catches the real defect.
- `TestIsAvailableDoesNotAdvanceTheRoundRobinCursor`, `TestIsAvailableWithoutServersIsFalse`,
  `BulkOnlyFlushConsumesExactlyOneHostSelection`, `EachRequestOfAMixedFlushUsesItsOwnSelectedHost`:
  0 of 4 pass against the old selector/flush code.

</p>
</details>

  - [x] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [x] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

Notes: the Linux build compiles the diff without warnings (the only warnings in the full build live in files this PR does not touch). The Decoder/Rule and Engine groups do not apply to this change (manager shared modules + wazuh-modulesd); ThreadSanitizer evidence for the touched modules is attached above.

<details><summary>Live manager+indexer — the new metrics are registered and coherent</summary>
<p>

`GET /metrics` over the module socket, healthy indexer, agent syncing. The new counters
(`sync.indexer.bulk.*`, `sync.bulk.flush.failures.*`, `sync.bulk.sessions.failed`) are present and
consistent with the group-commit counters:

```text
sync.bulk.flushes                 = 4     # logical group commits
sync.indexer.bulk.requests        = 4     # real _bulk POSTs (1:1 here -> no splits/retries, healthy)
sync.bulk.bytes.total             = 537704   # wire FlatBuffer bytes
sync.indexer.bulk.bytes.total     = 1670950  # serialized NDJSON bytes (~3.1x -> the unit mismatch
                                             #   that justified splitting the two bulk-size options)
sync.bulk.flush.failures.exhausted = 0
sync.bulk.flush.failures.documents = 0
sync.bulk.flush.failures.other     = 0
sync.bulk.sessions.failed          = 0
sync.requests.total.200            = 4
```

</p>
</details>

<details><summary>Live manager+indexer — fail-closed under an indexer outage (WP-2)</summary>
<p>

Indexer stopped (`docker stop wazuh-indexer`) with the module's monitor still inside its blind
window, then an agent-deletion driven at the socket. The connector reports the transport failure and
the pipeline answers a retryable failure instead of a false success — and `sync.requests.total.503`
rises for every rejection at the dispatch gate (23 honest 503s in one run):

```text
wazuh-manager-modulesd:inventory-sync-server:sync(indexer-connector): WARNING: deleteByQuery error: Could not connect to server, status code: -1.
wazuh-manager-modulesd:inventory-sync-server:sync: WARNING: Deleting the state documents of agent 899 failed: Could not connect to server. The caller may retry.
wazuh-manager-modulesd:inventory-sync-server:endpoints: WARNING: Rejected N agent deletion(s) with 503 in the last 90 s: no configured indexer host is currently healthy.

metric                          before  after  delta
sync.requests.total.503              4     27    +23
```

</p>
</details>

The retry-budget path itself (`sync.bulk.flush.failures.exhausted`) is proven deterministically by the
unit tests above; reproducing it end-to-end requires a bulk group commit to hit a down indexer within
the monitor's blind window. Two lab notes for whoever runs that scenario (both are environment setup,
not code): the manager (5.0) reads its overrides from `etc/wazuh-manager-internal-options.conf` (NOT
`local_internal_options.conf`), so widen `inventory_sync_server_indexer_monitoring_interval_seconds`
there and restart the manager; and a bulk group commit needs a real inventory CHANGE (install/remove a
package on the agent, or force a checksum mismatch) — an agent restart with an unchanged inventory
syncs as a no-op. A helper that scripts the outage and classifies the outcome is in the topic folder
(`scripts/indexer_outage_budget_test.sh`). A confirmed `deleteByQuery` against a DOWN indexer throws
immediately by design (its budget covers `429` rate-limiting only, not transport outages), so the
budget is exercised through the bulk path, not the delete path.

### Artifacts Affected

- `libindexer_connector.so` (shared module) — consumed by wazuh-modulesd (`inventory_sync_server`), the Vulnerability Scanner and the Content Manager; all inherit the bounded retry budget and the uniform host selection. The Vulnerability Scanner now forwards the retry budget from its own internal options too, instead of only ever using the built-in default.
- `wazuh-modulesd` (`wm_inventory_sync_server.c`, `wm_vulnerability_scanner.c` + the `inventory_sync_server` and `vulnerability_scanner` module libraries).
- Developer documentation: `docs/ref/modules/indexer_connector/README.md`, `docs/ref/modules/inventory-sync-server/{configuration,metrics,architecture}.md`.
- No packaging (RPM/DEB), default-configuration-file, or protocol changes.

### Configuration Changes

New internal options (set them in the manager's `etc/wazuh-manager-internal-options.conf` — the manager does NOT read `local_internal_options.conf`, which is the agent's file — and restart the manager), all with safe defaults so no action is needed on upgrade:

- `wazuh_modules.inventory_sync_server_indexer_sync_max_retry_attempts` — default `5`, range 0–1000. `0` disables the attempts bound; `1` means a single attempt with no retry.
- `wazuh_modules.inventory_sync_server_indexer_sync_max_retry_duration_seconds` — default `15`, range 0–3600. `0` disables the deadline. Both `0` makes retries fully unbounded and logs a startup `WARNING`.
- `wazuh_modules.inventory_sync_server_indexer_sync_connector_max_bulk_size` — default `10485760` (10 MiB), range 4096–104857600; caps the serialized NDJSON of one `_bulk` request.
- `wazuh_modules.indexer_max_retry_attempts` / `wazuh_modules.indexer_max_retry_duration_seconds` — the same retry budget for the **Vulnerability Scanner**'s connector instances (defaults `5` / `15`, ranges `0`–1000 / `0`–3600). Previously the Scanner had no way to override the built-in default; these give it the parity `inventory_sync_server` already had.

Behavior changes to existing options:

- `wazuh_modules.inventory_sync_server_indexer_sync_max_bulk_size` now drives **only** the pipeline group-commit threshold; it no longer reaches the connector. A deployment that had raised it past 10 MiB will now see the connector cut multiple bounded `_bulk` requests per group commit (correct behavior; tune the new option to restore a larger request size).
- `wazuh_modules.inventory_sync_server_indexer_sync_flush_interval_seconds` (already ignored) now logs a startup `WARNING` when set to a non-default value.

Backward-compatibility note: scenarios that previously "succeeded" silently (swallowed `409`/`429`, unvalidated bodies, infinite retries) now fail visibly with `500`/`503` and an agent retry — an expected increase in honest error responses under indexer degradation.

### Tests Introduced

All GoogleTest units, integrated in the existing suites (`indexer_connector_utest`: 250 → 310; `inventory_sync_server_utest`: 230 → 235):

- **413 split**: ID-conservation oracle across recursive splits (4/5/7 docs, asymmetric midpoints), single-operation-throw granularity pin, concurrent stagers + splits conservation (TSAN target).
- **Fail-closed validation**: delete-by-query `409` (throws, staged query dropped), `429` (retries until success), unparseable/empty bodies; parametrized update-by-query outcomes (failures, version_conflicts, missing counters, non-JSON, empty); bulk `200` without `errors`.
- **Retry budget**: `IndexerRetryBudget` unit tests; per-path budget exhaustion for bulk, chunks, delete-by-query, update-by-query and search; the chunks of a `413` split share their flush's attempts and deadline; deadline promptness; destruction during backoff; budget exhaustion under concurrent staging (connector stays reusable).
- **Host selection**: `isAvailable()` cursor neutrality, uniform round-robin distribution, exactly-one-selection-per-request at the connector level (red/green verified).
- **Configuration split**: connector cap sourced from the new option only; the pipeline threshold never reaches the connector; `<=0` sentinels; flush-interval startup WARNING emitted only for tuned values.
- **Metrics**: `takeBulkRequestStats()` counts every attempt and resets on take (splits included); pipeline drains connector request stats into the metrics; failed group commits counted by cause and blast radius. QA asserts real `_bulk` requests on `sync.indexer.bulk.requests`; `sync.bulk.flushes` keeps counting group commits.
- **Idempotency**: full-session replay confirmed via version conflicts with byte-identical staged frames.

## Review Checklist

- [ ] Code changes reviewed
- N/A

### DIFF
--- a/docs/ref/modules/indexer_connector/README.md
+++ b/docs/ref/modules/indexer_connector/README.md
@@ -31,7 +31,7 @@ The library provides two classes depending on the use case:
 
 ### Sync flush behavior
 
-- Buffer up to 10 MB of serialized events before flushing (configurable: `wazuh_modules.indexer_bulk_size_bytes` for Vulnerability Scanner, `wazuh_modules.inventory_sync_server_indexer_sync_max_bulk_size` for Inventory Sync Server).
+- Buffer up to 10 MB of serialized events before flushing (configurable: `wazuh_modules.indexer_bulk_size_bytes` for Vulnerability Scanner, `wazuh_modules.inventory_sync_server_indexer_sync_connector_max_bulk_size` for Inventory Sync Server).
 - Flush automatically after 20 seconds of inactivity (configurable: `wazuh_modules.indexer_flush_interval` for Vulnerability Scanner; the Inventory Sync Server deliberately overrides its periodic flush — its ingestion workers own every flush, see its [configuration reference](../inventory-sync-server/configuration.md)).
 - `flush_interval_seconds = 0` means **no background flush thread at all**: the connector is never created with one and every flush is the caller's. Use it when the caller has to answer for a failed flush — a timer flush that fails discards the staging buffer and has no caller to report to, so a later `flush()` finds an empty buffer and returns success for data that never landed. The value is set directly by the Inventory Sync Server; it is not reachable through the Vulnerability Scanner's `wazuh_modules.indexer_flush_interval`, whose range is 1–3600.
 - If the indexer returns HTTP 413 (payload too large), the batch is split and retried.
--- a/docs/ref/modules/indexer_connector/README.md
+++ b/docs/ref/modules/indexer_connector/README.md
@@ -87,12 +87,16 @@ achieve leaves documents nothing will ever overwrite:
   moved between the query's search and delete phases is skipped instead of aborting the whole run.
 - **A `200` is not automatically success.** The response is inspected, and the flush throws when it
   reports per-shard `failures` or a non-zero `version_conflicts` — the two ways a `200` can leave
-  matching documents in place. Callers treat that as retriable.
+  matching documents in place — or when its body cannot be parsed at all. Callers treat that as
+  retriable.
 - **Staged queries are dropped when a flush fails**, so a later flush cannot re-fire them after the
   caller already retried and succeeded (which would delete documents written in between).
-- HTTP-level `404`, `409` and `429` on a delete-by-query are tolerated (logged at debug) rather than
-  raised: a missing index has nothing to delete, and the other two are retried by re-running the
-  deletion.
+- HTTP-level `404` is tolerated (a missing index has nothing to delete), `429` is retried with the
+  same backoff as the bulk paths, and anything else — a request-level `409` included — fails the
+  flush: an unconfirmed delete is never reported as applied.
+- **`executeUpdateByQuery` follows the same contract**: a `200` whose body tallies `failures` or
+  `version_conflicts`, cannot be parsed, or lacks the `updated`/`total` counters fails the call
+  instead of confirming it.
 - **A delete-by-query is a SEARCH**, so it only sees documents that are already searchable. Callers
   that need it to cover writes of the last few seconds must refresh the index themselves — the
   connector does not do it for them, and `refresh()` requires `indices:admin/refresh`, which is not
--- a/docs/ref/modules/indexer_connector/README.md
+++ b/docs/ref/modules/indexer_connector/README.md
@@ -76,6 +76,16 @@ Example with the defaults (base = 1s, max = 15s):
 | 4th | random between 4s and 8s |
 | 5th and beyond | random between 8s and 15s (capped) |
 
+**Retries are bounded (sync).** Every sync retry loop carries a budget: at most `max_retry_attempts`
+failed attempts (default 5) and a wall-clock deadline of `max_retry_duration_seconds` (default 15s)
+per operation, whichever is spent first. On exhaustion the operation throws instead of retrying —
+the batch is dropped and the caller retries by re-staging, the same contract as any other terminal
+failure. `0` disables the corresponding bound. Without the budget, a persistent 429 or an
+unreachable indexer blocks the flushing worker (and, in inventory sync, the shard behind it) forever,
+long after the caller's response window closed. The deadline only gates the sleeps: one in-flight
+request can still overshoot it by up to `request_timeout_seconds`. The indexer's `Retry-After`
+header is not honored — the transport does not expose response headers.
+
 ### Delete-by-query
 
 `IndexerConnectorSync` also exposes the operation the manager's whole-agent deletion is built on. It
--- a/docs/ref/modules/indexer_connector/README.md
+++ b/docs/ref/modules/indexer_connector/README.md
@@ -78,7 +78,8 @@ Example with the defaults (base = 1s, max = 15s):
 
 **Retries are bounded (sync).** Every sync retry loop carries a budget: at most `max_retry_attempts`
 failed attempts (default 5) and a wall-clock deadline of `max_retry_duration_seconds` (default 15s)
-per operation, whichever is spent first. On exhaustion the operation throws instead of retrying —
+per operation, whichever is spent first; the chunks of a `413` split draw from their flush's budget,
+so a split cannot multiply it. On exhaustion the operation throws instead of retrying —
 the batch is dropped and the caller retries by re-staging, the same contract as any other terminal
 failure. `0` disables the corresponding bound. Without the budget, a persistent 429 or an
 unreachable indexer blocks the flushing worker (and, in inventory sync, the shard behind it) forever,
--- a/docs/ref/modules/indexer_connector/README.md
+++ b/docs/ref/modules/indexer_connector/README.md
@@ -85,7 +85,7 @@ failure. `0` disables the corresponding bound. Without the budget, a persistent 
 unreachable indexer blocks the flushing worker (and, in inventory sync, the shard behind it) forever,
 long after the caller's response window closed. The deadline only gates the sleeps: one in-flight
 request can still overshoot it by up to `request_timeout_seconds`. The indexer's `Retry-After`
-header is not honored — the transport does not expose response headers.
+header is not honored — the transport does not expose response headers (tracked in #38942).
 
 ### Delete-by-query
 
--- a/docs/ref/modules/indexer_connector/README.md
+++ b/docs/ref/modules/indexer_connector/README.md
@@ -26,7 +26,7 @@ The library provides two classes depending on the use case:
 1. The caller instantiates a connector with a JSON configuration (derived from the `<indexer>` XML block).
 2. Credentials (`username`/`password`) are read from the RocksDB keystore (`queue/keystore/`).
 3. A background health-monitor thread polls `/_cat/health` on all configured hosts every 60 seconds and marks nodes available or unavailable.
-4. A server-selector performs round-robin load balancing across available nodes.
+4. A server-selector performs round-robin load balancing across available nodes. Selection has no side effects: availability probes (`isAvailable()`) never advance the round-robin cursor, and a host is consumed from the rotation only by the request actually sent to it, so traffic distributes uniformly across healthy nodes.
 5. Documents are accumulated in memory (both sync and async) and flushed as OpenSearch Bulk API requests.
 
 ### Sync flush behavior
--- a/src/remoted/remoted_module/src/http_server/RestinioHttpServer.cpp
+++ b/src/remoted/remoted_module/src/http_server/RestinioHttpServer.cpp
@@ -1339,10 +1339,7 @@ namespace remoted::http
             // Bound simultaneous connections so the read-phase peak (bodies still being
             // received, before they reach the in-flight budget) can't grow unbounded.
             .max_parallel_connections(config.maxParallelConnections)
-            // Left at RESTinio's default (false): with it on, a connection accepted right as
-            // the module stops can be torn down mid-setup, freeing the acceptor before that
-            // connection's cleanup runs -- a use-after-free (#38741).
-            .separate_accept_and_create_connect(false)
+            .separate_accept_and_create_connect(true)
             .incoming_http_msg_limits(restinio::incoming_http_msg_limits_t {}
                                           .max_url_size(config.maxUrlSize)
                                           .max_field_name_size(config.maxHeaderNameSize)
--- a/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
+++ b/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
@@ -499,14 +499,16 @@ class IndexerConnectorSyncImpl final
                         ConfigurationParameters {.timeout = m_requestTimeoutMs});
                     if (deleteByQueryLeftDocuments)
                     {
-                        throw IndexerConnectorException("deleteByQuery did not delete every matching document");
+                        throw IndexerConnectorException("deleteByQuery did not delete every matching document",
+                                                        IndexerConnectorException::Category::DocumentRejected);
                     }
                     if (deleteByQueryNeedsRetry)
                     {
                         const auto retryDelay = retryBackoff.nextDelay();
                         if (retryBudget.exhausted(retryDelay))
                         {
-                            throw IndexerConnectorException("deleteByQuery retry budget exhausted");
+                            throw IndexerConnectorException("deleteByQuery retry budget exhausted",
+                                                            IndexerConnectorException::Category::RetryExhausted);
                         }
                         LOGFN_DEBUG2(
                             m_logFn, "Retrying deleteByQuery in %lld ms.", static_cast<long long>(retryDelay.count()));
@@ -632,7 +634,8 @@ class IndexerConnectorSyncImpl final
                     m_bulkData.clear();
                     m_boundaries.clear();
                     m_lastBulkTime = std::chrono::steady_clock::now();
-                    throw IndexerConnectorException("Bulk operation had indexing failures");
+                    throw IndexerConnectorException("Bulk operation had indexing failures",
+                                                    IndexerConnectorException::Category::DocumentRejected);
                 }
                 if (needToRetry)
                 {
@@ -642,7 +645,8 @@ class IndexerConnectorSyncImpl final
                         m_bulkData.clear();
                         m_boundaries.clear();
                         m_lastBulkTime = std::chrono::steady_clock::now();
-                        throw IndexerConnectorException("Bulk retry budget exhausted");
+                        throw IndexerConnectorException("Bulk retry budget exhausted",
+                                                        IndexerConnectorException::Category::RetryExhausted);
                     }
                     LOGFN_DEBUG2(
                         m_logFn, "Retrying bulk request in %lld ms.", static_cast<long long>(retryDelay.count()));
@@ -833,14 +837,16 @@ class IndexerConnectorSyncImpl final
                                 ConfigurationParameters {.timeout = m_requestTimeoutMs});
             if (indexingFailures)
             {
-                throw IndexerConnectorException("Bulk chunk operation had indexing failures");
+                throw IndexerConnectorException("Bulk chunk operation had indexing failures",
+                                                IndexerConnectorException::Category::DocumentRejected);
             }
             if (needToRetry)
             {
                 const auto retryDelay = retryBackoff.nextDelay();
                 if (retryBudget.exhausted(retryDelay))
                 {
-                    throw IndexerConnectorException("Bulk chunk retry budget exhausted");
+                    throw IndexerConnectorException("Bulk chunk retry budget exhausted",
+                                                    IndexerConnectorException::Category::RetryExhausted);
                 }
                 LOGFN_DEBUG2(m_logFn, "Retrying bulk chunk in %lld ms.", static_cast<long long>(retryDelay.count()));
                 std::unique_lock<std::mutex> lock(m_retryMutex);
@@ -1236,7 +1242,8 @@ public:
             if (updateByQueryFailed)
             {
                 m_notify.clear();
-                throw IndexerConnectorException("Update by query did not confirm every requested update");
+                throw IndexerConnectorException("Update by query did not confirm every requested update",
+                                                IndexerConnectorException::Category::DocumentRejected);
             }
             if (needToRetry)
             {
@@ -1244,7 +1251,8 @@ public:
                 if (retryBudget.exhausted(retryDelay))
                 {
                     m_notify.clear();
-                    throw IndexerConnectorException("Update by query retry budget exhausted");
+                    throw IndexerConnectorException("Update by query retry budget exhausted",
+                                                    IndexerConnectorException::Category::RetryExhausted);
                 }
                 LOGFN_DEBUG2(
                     m_logFn, "Retrying update by query in %lld ms.", static_cast<long long>(retryDelay.count()));
@@ -1327,7 +1335,8 @@ public:
                 const auto retryDelay = retryBackoff.nextDelay();
                 if (retryBudget.exhausted(retryDelay))
                 {
-                    throw IndexerConnectorException("Search query retry budget exhausted");
+                    throw IndexerConnectorException("Search query retry budget exhausted",
+                                                    IndexerConnectorException::Category::RetryExhausted);
                 }
                 LOGFN_DEBUG2(m_logFn, "Retrying search query in %lld ms.", static_cast<long long>(retryDelay.count()));
                 std::unique_lock<std::mutex> lock(m_retryMutex);
@@ -1632,7 +1641,8 @@ public:
                 const auto retryDelay = retryBackoff.nextDelay();
                 if (retryBudget.exhausted(retryDelay))
                 {
-                    throw IndexerConnectorException("Search request retry budget exhausted");
+                    throw IndexerConnectorException("Search request retry budget exhausted",
+                                                    IndexerConnectorException::Category::RetryExhausted);
                 }
                 LOGFN_DEBUG2(
                     m_logFn, "Retrying search request in %lld ms.", static_cast<long long>(retryDelay.count()));
--- a/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
+++ b/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
@@ -368,8 +368,6 @@ class IndexerConnectorSyncImpl final
             throw IndexerConnectorException("No data to process");
         }
 
-        auto serverUrl = m_selector->getNext();
-
         // No success callback below may throw: THttpRequest::post() invokes them inside its own
         // try block and remaps anything thrown there to onError(statusCode = -1), which the error
         // handlers read as a transport outage to retry -- resending a buffer that no longer
@@ -465,12 +463,6 @@ class IndexerConnectorSyncImpl final
 
         for (const auto& [index, query] : m_deleteByQuery)
         {
-            std::string url;
-            url += serverUrl;
-            url += "/";
-            url += index;
-            url += "/_delete_by_query";
-            LOGFN_DEBUG2(m_logFn, "Deleting by query: %s", url.c_str());
             try
             {
                 IndexerExponentialBackoff retryBackoff {std::chrono::seconds {RetryDelay},
@@ -484,6 +476,15 @@ class IndexerConnectorSyncImpl final
                         return;
                     }
                     deleteByQueryNeedsRetry = false;
+                    // Resolved per attempt, like every other request in this file: a host is
+                    // consumed only by the request actually sent to it, and a retry rotates to
+                    // the next healthy host.
+                    std::string url;
+                    url += m_selector->getNext();
+                    url += "/";
+                    url += index;
+                    url += "/_delete_by_query";
+                    LOGFN_DEBUG2(m_logFn, "Deleting by query: %s", url.c_str());
                     m_httpRequest->post(
                         RequestParameters {
                             .url = HttpURL(url), .data = query.dump(), .secureCommunication = m_secureCommunication},
--- a/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
+++ b/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
@@ -685,7 +685,7 @@ class IndexerConnectorSyncImpl final
                 indexingFailures = true;
             }
         };
-        const auto onError = [this, &needToRetry, boundaries](
+        const auto onError = [this, &needToRetry, boundaries, data](
                                  const std::string& error, const long statusCode, const std::string& responseBody)
         {
             if (statusCode == HTTP_CONTENT_LENGTH)
@@ -696,12 +696,19 @@ class IndexerConnectorSyncImpl final
                     const size_t midPoint = boundaries.size() / 2;
                     std::span<size_t> firstBoundaries(boundaries.begin(),
                                                       boundaries.begin() + static_cast<long>(midPoint));
-                    const size_t firstEndPos = boundaries[midPoint - 1];
-                    std::string_view firstHalf(m_bulkData.data(), firstEndPos);
                     std::span<size_t> secondBoundaries(boundaries.begin() + static_cast<long>(midPoint),
                                                        boundaries.end());
-                    const size_t secondStartPos = boundaries[midPoint - 1];
-                    std::string_view secondHalf(m_bulkData.data() + secondStartPos, m_bulkData.size() - secondStartPos);
+                    // Boundaries are absolute offsets into m_bulkData, but this chunk may start
+                    // anywhere in it: both halves must be subviews of `data`, never rebuilt from
+                    // the whole buffer, or a nested split resends other chunks' operations.
+                    const size_t chunkStart = static_cast<size_t>(data.data() - m_bulkData.data());
+                    const size_t splitPos = boundaries[midPoint - 1];
+                    if (splitPos <= chunkStart || splitPos - chunkStart >= data.size())
+                    {
+                        throw IndexerConnectorException("Bulk split boundary falls outside its chunk");
+                    }
+                    const std::string_view firstHalf = data.substr(0, splitPos - chunkStart);
+                    const std::string_view secondHalf = data.substr(splitPos - chunkStart);
                     processBulkChunk(firstHalf, firstBoundaries);
                     processBulkChunk(secondHalf, secondBoundaries);
                     return;
--- a/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
+++ b/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
@@ -348,6 +348,13 @@ class IndexerConnectorSyncImpl final
     static constexpr size_t DEFAULT_MAX_RETRY_DURATION_SECONDS {15};
     size_t m_maxRetryAttempts {DEFAULT_MAX_RETRY_ATTEMPTS};
     std::chrono::milliseconds m_maxRetryDuration {std::chrono::seconds {DEFAULT_MAX_RETRY_DURATION_SECONDS}};
+    /// The `_bulk` requests actually sent (attempts, splits and retries included), for callers
+    /// whose own flush accounting would otherwise under-count the real request traffic. Atomics:
+    /// bumped under m_mutex by the flushing thread but drained by takeBulkRequestStats() from any
+    /// thread.
+    std::atomic<uint64_t> m_bulkRequestsSent {0};
+    std::atomic<uint64_t> m_bulkBytesSent {0};
+
     /// Fallback for 'request_timeout_seconds' when the configuration has no opinion.
     static constexpr long DEFAULT_REQUEST_TIMEOUT_SECONDS {60};
     /// Upper bound in milliseconds for one data request against the indexer
@@ -613,6 +620,8 @@ class IndexerConnectorSyncImpl final
                 LOGFN_DEBUG2(m_logFn, "Sending bulk data to: %s", url.c_str());
                 LOGFN_DEBUG2(m_logFn, "Bulk data: %s", m_bulkData.c_str());
 
+                m_bulkRequestsSent.fetch_add(1);
+                m_bulkBytesSent.fetch_add(m_bulkData.size());
                 m_httpRequest->post(RequestParameters {.url = HttpURL(url),
                                                        .data = m_bulkData,
                                                        .secureCommunication = m_secureCommunication},
@@ -815,6 +824,8 @@ class IndexerConnectorSyncImpl final
             url += m_selector->getNext();
             url += "/_bulk";
             LOGFN_DEBUG2(m_logFn, "Sending bulk chunk to: %s", url.c_str());
+            m_bulkRequestsSent.fetch_add(1);
+            m_bulkBytesSent.fetch_add(data.size());
             m_httpRequest->post(RequestParametersStringView {.url = HttpURL(url),
                                                              .data = data,
                                                              .secureCommunication = m_secureCommunication},
@@ -1831,5 +1842,13 @@ public:
     bool isAvailable() const
     {
         return m_selector->isAvailable();
+    }
+
+    /**
+     * @brief Returns the `_bulk` request counts accumulated since the previous call and resets them.
+     */
+    IndexerBulkRequestStats takeBulkRequestStats()
+    {
+        return {m_bulkRequestsSent.exchange(0), m_bulkBytesSent.exchange(0)};
     }
 };
--- a/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
+++ b/src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
@@ -338,6 +338,16 @@ class IndexerConnectorSyncImpl final
     size_t m_maxBulkSize {MaxBulkSize};
     size_t m_flushInterval {FlushInterval};
     size_t m_maxRetryDelay {MaxRetryDelay};
+    /// Fallbacks for the retry budget ('max_retry_attempts' / 'max_retry_duration_seconds';
+    /// 0 disables the corresponding bound). Every retry loop is bounded by both: without a
+    /// budget, a persistent 429 or an unreachable indexer blocks the flushing worker -- and the
+    /// shard behind it -- forever, long after the caller's response window closed. Sized to fail
+    /// a flush before that window (remoted's downstream response timeout, default 20 s) closes;
+    /// the cross-team sizing of the whole timeout chain is decided elsewhere.
+    static constexpr size_t DEFAULT_MAX_RETRY_ATTEMPTS {5};
+    static constexpr size_t DEFAULT_MAX_RETRY_DURATION_SECONDS {15};
+    size_t m_maxRetryAttempts {DEFAULT_MAX_RETRY_ATTEMPTS};
+    std::chrono::milliseconds m_maxRetryDuration {std::chrono::seconds {DEFAULT_MAX_RETRY_DURATION_SECONDS}};
     /// Fallback for 'request_timeout_seconds' when the configuration has no opinion.
     static constexpr long DEFAULT_REQUEST_TIMEOUT_SECONDS {60};
     /// Upper bound in milliseconds for one data request against the indexer
@@ -465,6 +475,7 @@ class IndexerConnectorSyncImpl final
             {
                 IndexerExponentialBackoff retryBackoff {std::chrono::seconds {RetryDelay},
                                                         std::chrono::seconds {m_maxRetryDelay}};
+                IndexerRetryBudget retryBudget {m_maxRetryAttempts, m_maxRetryDuration};
                 do
                 {
                     if (m_stopping.load())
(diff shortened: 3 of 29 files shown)


src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -337,6 +338,23 @@ class IndexerConnectorSyncImpl final
...
     size_t m_maxRetryDelay {MaxRetryDelay};
+    /// Fallbacks for the retry budget ('max_retry_attempts' / 'max_retry_duration_seconds';
+    /// 0 disables the corresponding bound). Every retry loop is bounded by both: without a
+    /// budget, a persistent 429 or an unreachable indexer blocks the flushing worker -- and the
+    /// shard behind it -- forever, long after the caller's response window closed. Sized to fail
+    /// a flush before that window (remoted's downstream response timeout, default 20 s) closes;
+    /// the cross-team sizing of the whole timeout chain is decided elsewhere.
+    static constexpr size_t DEFAULT_MAX_RETRY_ATTEMPTS {5};
+    static constexpr size_t DEFAULT_MAX_RETRY_DURATION_SECONDS {15};
+    size_t m_maxRetryAttempts {DEFAULT_MAX_RETRY_ATTEMPTS};
```
matigarciadev, line 349: The new retry budget (`IndexerRetryBudget` in `exponentialBackoff.hpp`, defaults 5 attempts / 15s) isn't scoped to `inventory_sync_server`, it lives in `IndexerConnectorSync` itself, so it applies to every consumer of that class. Inventory Sync Server got proper config plumbing for it — internal options in `wm_inventory_sync_server.c:344-354` flow through `indexerConnectorConfig.cpp` into the connector JSON, so operators can tune or disable it. Vulnerability Scanner didn't get the same treatment: `buildIndexerConfig()` in `vulnerabilityScannerFacade.cpp:152-207` only passes through `max_bulk_size` and `flush_interval_seconds`, never `max_retry_attempts`/`max_retry_duration_seconds`. Since those keys never show up in the JSON, every VD Scanner connector instance just falls back to the hardcoded 5/15s default, and there's nothing in `wm_vulnerability_scanner.c` to override it. Is this expected?

src/wazuh_modules/src/wm_inventory_sync_server.c
```diff
@@ -326,6 +336,23 @@ static void wm_inventory_sync_server_read_tunables(inventory_sync_server_config_
...
+                              "inventory_sync_server_indexer_sync_max_retry_attempts", /* -> max_retry_attempts */
+                              1,
+                              1000,
+                              5);
+    config->indexer_sync_max_retry_duration_seconds = getDefine_Int_default(
+        "wazuh_modules",
+        "inventory_sync_server_indexer_sync_max_retry_duration_seconds", /* -> max_retry_duration_seconds */
+        1,
+        3600,
+        15);
```
matigarciadev, line 358: `max_retry_attempts` here (and `max_retry_duration_seconds` right below it) is bounded to a minimum of 1, which means 0 isn't reachable through `local_internal_options.conf` even though the connector itself treats 0 as a valid "no bound" value (`indexerConnectorSyncImpl.hpp`, max_retry_attempts must be >= 0 (0 disables the bound), same story for the duration). If someone tries inventory_sync_server_indexer_sync_max_retry_attempts=0 they don't get unbounded retries back, getDefine_Int_default calls merror_exit on out-of-range and wazuh-modulesd just fails to start.

LucioDonda: Yes, 0 shouldn't be possible, but attemps doesn't mean retry so 1 should be the limit, 1 and 0 retry duration should mean infinite attempts.

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
LucioDonda: you're rigth, now I've wired the same setting and all should be treated the same.

src/remoted/remoted_module/src/http_server/RestinioHttpServer.cpp
```diff
@@ -1339,10 +1339,7 @@ namespace remoted::http
             // Bound simultaneous connections so the read-phase peak (bodies still being
             // received, before they reach the in-flight budget) can't grow unbounded.
             .max_parallel_connections(config.maxParallelConnections)
-            // Left at RESTinio's default (false): with it on, a connection accepted right as
-            // the module stops can be torn down mid-setup, freeing the acceptor before that
-            // connection's cleanup runs -- a use-after-free (#38741).
-            .separate_accept_and_create_connect(false)
+            .separate_accept_and_create_connect(true)
```
NahuFigueroa97, line 1342: This reverts 278b42f1cc (#38741), which set this to false specifically to close a heap-use-after-free on shutdown (a connection accepted right as the server stops gets torn down mid-setup, freeing the acceptor before its cleanup runs). No commit on this branch touches src/external/restinio or any restinio patch — the commit message "update restinio flag after library patch" doesn't match the diff. Reproduced live under ASAN on the very first run:

￼
ShutdownRace.StopSequenceSurvivesConnectionsAcceptedButNeverWritten
==ERROR: AddressSanitizer: heap-use-after-free ...
    #0 restinio::connection_count_limits::...decrement_parallel_connections()...
    #33 remoted::http::RestinioHttpServer::stop() RestinioHttpServer.cpp:1480
This must go back to false before merge, or this commit needs to be dropped entirely if it was meant to land with a restinio patch that isn't in this branch.

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -742,6 +811,7 @@ class IndexerConnectorSyncImpl final
         };
         IndexerExponentialBackoff retryBackoff {std::chrono::seconds {RetryDelay},
                                                 std::chrono::seconds {m_maxRetryDelay}};
+        IndexerRetryBudget retryBudget {m_maxRetryAttempts, m_maxRetryDuration};
```
NahuFigueroa97: IndexerRetryBudget retryBudget is constructed fresh here, inside processBulkChunk() — which recurses on every nested 413 split. So a flush that splits into N chunks, each hitting transient rate-limiting independently, can spend up to max_retry_duration_seconds per chunk, not once per flush as the issue's "double budget per flush operation" and the docs imply. Consider threading one IndexerRetryBudget through splitAndProcessBulk() and sharing it across all recursive processBulkChunk() calls for that flush.

docs/ref/modules/indexer_connector/README.md
```diff
@@ -76,6 +76,16 @@ Example with the defaults (base = 1s, max = 15s):
...
 | 5th and beyond | random between 8s and 15s (capped) |
 
+**Retries are bounded (sync).** Every sync retry loop carries a budget: at most `max_retry_attempts`
+failed attempts (default 5) and a wall-clock deadline of `max_retry_duration_seconds` (default 15s)
+per operation, whichever is spent first. On exhaustion the operation throws instead of retrying —
+the batch is dropped and the caller retries by re-staging, the same contract as any other terminal
+failure. `0` disables the corresponding bound. Without the budget, a persistent 429 or an
+unreachable indexer blocks the flushing worker (and, in inventory sync, the shard behind it) forever,
+long after the caller's response window closed. The deadline only gates the sleeps: one in-flight
+request can still overshoot it by up to `request_timeout_seconds`. The indexer's `Retry-After`
```
NahuFigueroa97, line 87: Confirmed accurate (curlWrapper.hpp never installs CURLOPT_HEADERFUNCTION, so response headers genuinely aren't exposed) — but the issue's proposed fix explicitly asks to "honor Retry-After when present" with a required test (Retry-After: 2 delays ~2s). Neither is implemented. Worth a tracking issue reference here instead of just stating it's not honored, so this doesn't silently stay unmet.

src/wazuh_modules/inventory_sync_server/src/sync/syncPipeline.cpp
NahuFigueroa97, line 66: sync.bulk.flushes still counts group-commits, not real _bulk requests — the new sync.indexer.bulk.requests metric is what actually counts real requests. The issue's Required Tests text literally says "QA check that sync.bulk.flushes counts real _bulk requests" — if QA points at that exact metric name, the check will fail even though the feature works correctly under the new metric name. Worth confirming with the author which metric QA should actually assert on.

NahuFigueroa97 (changes requested):

src/remoted/remoted_module/src/http_server/RestinioHttpServer.cpp
LucioDonda: restinio patched applies so this wont be neccesary any more.

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
LucioDonda: Right, the chunk loop was the one place the budget got multiplied. Fixed in f5dc3b342b: one IndexerRetryBudget is built in processBulk() and threaded by reference through splitAndProcessBulk()/processBulkChunk(), nested splits included, so a split shares its flush's attempts and deadline. Two tests pin it (SplitChunksShareTheFlushRetryAttempts / ...Deadline); README and configuration.md no longer describe a chunk as its own operation.

docs/ref/modules/indexer_connector/README.md
LucioDonda: Agreed it shouldn't stay an unreferenced note. Opened #38942 to track it (http-request has to expose response headers first, then the connector honors Retry-After capped by max_retry_delay_seconds and charged to the budget). README now points at it (66f32c0aee), and the Retry-After test in #38778 WP-3 moved there.

src/wazuh_modules/inventory_sync_server/src/sync/syncPipeline.cpp
LucioDonda: QA should assert on sync.indexer.bulk.requests; sync.bulk.flushes deliberately keeps counting group commits (documented in four places and the basis for tuning indexer_sync_max_bulk_size), and their ratio is the split/retry amplification. I updated the WP-5 Required Tests line in #38778 and the PR's Tests section to say exactly that.

NahuFigueroa97 (approved):
LGTM!

```
Success.
Success.
Cleaned up temporary file: /tmp/merged_wcs_h4fbguc0.yml
Generated wazuh-decoders.json
Success.
Copying store files...
Installing GeoIP databases...


Starting Wazuh...
manager
Starting Wazuh v5.0.0...
Started wazuh-manager-apid...
Started wazuh-manager-db...
Completed.

 - To start Wazuh:

 - To stop Wazuh:

   More information at: 
```

### 1. Real test output (ASAN + TSAN builds of this branch)

**WP-1 / WP-2 / WP-4 — `indexer_connector_utest`, built with `-DFSANITIZE=ON -DUNIT_TEST=ON -DCMAKE_BUILD_TYPE=Debug`:**

```
[ RUN      ] IndexerConnectorSyncTest.Recursive413SplitsDeliverEveryDocumentExactlyOnce
[       OK ] IndexerConnectorSyncTest.Recursive413SplitsDeliverEveryDocumentExactlyOnce (5 ms)
[ RUN      ] IndexerConnectorSyncTest.Exhausted413SplitThrowsOnASingleOperationChunkOnly
[       OK ] IndexerConnectorSyncTest.Exhausted413SplitThrowsOnASingleOperationChunkOnly (0 ms)
[ RUN      ] IndexerConnectorSyncTest.ConcurrentStagersWith413SplitsDeliverEveryDocumentExactlyOnce
[       OK ] IndexerConnectorSyncTest.ConcurrentStagersWith413SplitsDeliverEveryDocumentExactlyOnce (160 ms)
[ RUN      ] IndexerConnectorSyncTest.DeleteByQueryError409ThrowsAndDropsTheStagedQuery
[       OK ] IndexerConnectorSyncTest.DeleteByQueryError409ThrowsAndDropsTheStagedQuery (0 ms)
[ RUN      ] IndexerConnectorSyncTest.DeleteByQueryError429RetriesUntilSuccess
[       OK ] IndexerConnectorSyncTest.DeleteByQueryError429RetriesUntilSuccess (2704 ms)
[ RUN      ] IndexerConnectorSyncTest.BulkResponseValidationMissingErrorsFieldFails
[       OK ] IndexerConnectorSyncTest.BulkResponseValidationMissingErrorsFieldFails (0 ms)

[----------] 12 tests from ServerSelectorTest
... TestIsAvailableDoesNotAdvanceTheRoundRobinCursor  OK
... TestRoundRobinDistributesUniformlyAcrossHealthyHosts  OK
[  PASSED  ] 12 tests.

[==========] 308 tests from 12 test suites ran. (61243 ms total)
[  PASSED  ] 308 tests.
```
Exit code 0. Zero `AddressSanitizer`/`LeakSanitizer` lines anywhere in the log.

**WP-3 — same suite, built with `-DTSANITIZE=ON` in a separate build dir, run via `setarch $(uname -m) -R` (required in this container to avoid a TSAN/ASLR mapping conflict unrelated to the code):**

```
[ RUN      ] IndexerConnectorSyncTest.DestructionDuringUnboundedBackoffExitsPromptly
[       OK ] IndexerConnectorSyncTest.DestructionDuringUnboundedBackoffExitsPromptly (106 ms)
[ RUN      ] IndexerConnectorSyncTest.BudgetExhaustionUnderConcurrentStagingLeavesConnectorReusable
[       OK ] IndexerConnectorSyncTest.BudgetExhaustionUnderConcurrentStagingLeavesConnectorReusable (48 ms)
[ RUN      ] IndexerConnectorSyncTest.TransportRetriesUnderConcurrentStagingLoseNothing
[       OK ] ...

[==========] 308 tests from 12 test suites ran. (59570 ms total)
[  PASSED  ] 308 tests.
ThreadSanitizer: reported 2 warnings
```
Both warnings traced to `indexerConnectorAsync_test.cpp` (`HandleError409VersionConflict`, `HandleError429TooManyRequests`) — a file with zero commits from this branch (last touched at `dfd5888260`, already an ancestor of the merge-base) and zero race-report frames referencing anything this PR changed (`indexerConnectorSyncImpl.hpp`, `indexerConnectorSync_test.cpp`, `exponentialBackoff.hpp` all clean).

**WP-5 / metrics — `inventory_sync_server_utest`, ASAN build:**

```
[ RUN      ] InventorySyncServerModuleTest.ATunedSyncFlushIntervalWarnsThatItHasNoEffect
[       OK ] ...
[ RUN      ] InventorySyncServerModuleTest.TheDefaultSyncFlushIntervalDoesNotWarn
[       OK ] ...
[ RUN      ] SyncPipelineTest.TheConnectorsRealBulkRequestsReachTheMetrics
[       OK ] ...
[ RUN      ] SyncPipelineTest.FailedGroupCommitsAreCountedByCauseAndBlastRadius
[       OK ] ...
[ RUN      ] SyncPipelineTest.AFailedFlushFailsTheWholeBatch
[       OK ] ...

[==========] 254 tests from 21 test suites ran. (2114 ms total)
[  PASSED  ] 254 tests.
```

**Follow-up fixes (`d438ac0bd0`, shared-budget-per-flush) — the two new tests added to close the gap found in the first review pass:**

```cpp
TEST_F(IndexerConnectorSyncTest, SplitChunksShareTheFlushRetryAttempts)
// scripts: 413 -> 429 -> 200 -> 429 -> 200, max_retry_attempts=2
// asserts: connector.flush() throws Category::RetryExhausted, postCount == 4
//          ("full batch, first chunk twice, second chunk once")

TEST_F(IndexerConnectorSyncTest, SplitChunksShareTheFlushRetryDeadline)
// same script, max_retry_duration_seconds=2 instead of an attempts cap
// asserts: the second chunk's backoff is NOT granted its own fresh window
```
Both present in the branch tip and covered by the 308/308 pass above.

---

### 2. Live checks against the manager installed from this branch (`494cbef09e`)

**All daemons up:**
```
$ ps aux | grep wazuh-manager | awk '{print $NF}'
/var/wazuh-manager/api/scripts/wazuh_manager_apid.py
/var/wazuh-manager/bin/wazuh-manager-authd
/var/wazuh-manager/bin/wazuh-manager-db
/var/wazuh-manager/bin/wazuh-manager-analysisd
/var/wazuh-manager/bin/wazuh-manager-remoted
/var/wazuh-manager/bin/wazuh-manager-modulesd
/var/wazuh-manager/framework/scripts/wazuh_manager_clusterd.py
```

**The bulk-indexing path this PR touches, exercised for real** (content-updater's initial load, ~380k documents, straight through the sync connector's bulk/split/retry machinery this PR patches — no errors, no crash):
```
2026/09/04 17:46:58 wazuh-manager-modulesd:content-updater: INFO: IndexerDownloader: Slice 0 (1/2) complete — 190403 docs in 711866ms
2026/09/04 17:47:00 wazuh-manager-modulesd:content-updater: INFO: IndexerDownloader: Slice 1 (2/2) complete — 190386 docs in 713487ms
2026/09/04 17:47:01 wazuh-manager-modulesd:content-updater: INFO: IndexerDownloader: Initial load download phase complete — 380789 documents, cursor: '906452'
```

**WP-5's new metrics, present and correctly described on the live instance:**
```
$ curl -s --unix-socket /var/wazuh-manager/queue/sockets/inventory-sync-http.sock http://localhost/metrics

sync.indexer.bulk.requests = 0
  -- "_bulk HTTP requests the sync connectors actually sent (splits and retries included),
      vs. sync.bulk.flushes group commits"
sync.indexer.bulk.bytes.total = 0
  -- "NDJSON payload bytes those _bulk requests carried"
sync.bulk.flush.failures.documents = 0
sync.bulk.flush.failures.exhausted = 0   -- "failed by an exhausted retry budget (429, transport)"
sync.bulk.flush.failures.other = 0
sync.bulk.sessions.failed = 0            -- "the blast radius: one bad flush punishes every session in its batch"
```
(All zero because no `/stateful` agent-sync traffic has hit this specific instance yet — the metrics being present, correctly named, and correctly described on a live, running build is what's being verified here, not a specific count.)

**No indexer-connector errors anywhere in the log since startup:**
```
$ grep -i "indexer.*error\|error.*indexer" /var/wazuh-manager/logs/wazuh-manager.log
(no output)
```

matigarciadev (commented):
### LGTM!
**Built and Install:**
```console
General settings:
    TARGET:             manager
    V:                  
    DEBUG:              
    EXTERNAL_SRC_ONLY:  
User settings:
USE settings:
    USE_INOTIFY:        no
    USE_BIG_ENDIAN:     no
    USE_SELINUX:        no
    USE_AUDIT:          no
    IMAGE_TRUST_CHECKS: 1
Compiler:
    MAKE              make

Done building manager

Wait for success...
success
Processing ./.
  Created wheel for wazuh: filename=wazuh-5.0.0-py3-none-any.whl size=281993 sha256=77423e376211638455faf263788e0e83e65c5eb75a0e38c66c0c623c9572c09e
  Stored in directory: /tmp/pip-ephem-wheel-cache-uin9azhw/wheels/50/e9/6c/171615ee3709da12266524fed91cb0863278bc995eed9295be
Successfully built wazuh
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the --root-user-action option if you know what you are doing and want to suppress this warning.

Processing ./.
  Created wheel for api: filename=api-5.0.0-py3-none-any.whl size=96936 sha256=e0c7ee33c1861be6da5c38a8d8ee27c831fe7358bdd54678101465a478c909d7
  Stored in directory: /tmp/pip-ephem-wheel-cache-jtzb6rrr/wheels/12/3e/cd/fe7f9fbf555382f244bd9d9630026d54f4ffbcb658cc4f4190
Successfully built api
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the --root-user-action option if you know what you are doing and want to suppress this warning.

Generating schema files...
Loading resources...
Loading wcs_flat.yml...
Successfully merged 1 files into temporary file: /tmp/merged_wcs_k3hgonpw.yml
Loading schema template...
Success.
Success.
Success.
Generating engine schema...
Success.
Success.
Success.
Cleaned up temporary file: /tmp/merged_wcs_k3hgonpw.yml
Generated wazuh-decoders.json
Success.
Copying store files...
Installing GeoIP databases...


Starting Wazuh...
Starting Wazuh v5.0.0...
Started wazuh-manager-apid...
Started wazuh-manager-db...
wazuh-manager-analysisd died during event dumper check.
wazuh-manager-analysisd did not start correctly.

 - To start Wazuh:

 - To stop Wazuh:

   More information at: 
```

**Manager Status:**
```console
wazuh-manager-analysisd: Process 132610 not used by Wazuh, removing...
Killing wazuh-manager-db...
Killing wazuh-manager-authd...
Killing wazuh-manager-apid...
Wazuh v5.0.0 Stopped
Starting Wazuh v5.0.0...
Started wazuh-manager-apid...
Started wazuh-manager-db...
Completed.

```

Issue #38329, opened 2026-08-12
https_client component tests: intermittent startup hang under valgrind fails the RTR memory stage


jr0me:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0-rc1 (`5.0.0-https`) | Agent — `client-agent/https_client` component tests | Agent | Sources | Ubuntu 24.04 x86_64, valgrind-3.22.0, GCC 13 |

## Description

`FacadeE2eTest.AHeldStatefulRequestDoesNotStallStatelessOrControl` intermittently fails under valgrind, which fails the RTR memory stage for `client-agent/https_client` (`[https_client_component_test : FAILED]`) and therefore the whole `rtr (client-agent/https_client, agent)` check.

It is worth being precise about what this is **not**, because the stage is named "valgrind" and the failure looks like a leak report:

```
definitely lost:   0 bytes
indirectly lost:   0 bytes
memcheck errors:   0        (no invalid read/write, no uninitialised values)
possibly lost:    70 records   <- fewer than 5.0.0-https, which has 76 and exits 0
still reachable: 18419 records <- not errors by default
```

The non-zero exit comes from a **gtest failure inside the valgrind run**, not from memory. The failing assertion is the startup wait:

```
facadeE2e_component_test.cpp:530: Failure
Value of: waitFor(recorder.startupCount, 1, 3000)
  Actual: false
Expected: true
[  FAILED  ] FacadeE2eTest.AHeldStatefulRequestDoesNotStallStatelessOrControl (34521 ms)
```

`hc_start()` is called, and the `on_startup_result` callback never fires within the deadline.

## Reproduction

```bash
cd src
make deps TARGET=agent -j$(nproc)
make TARGET=agent DEBUG=1 TEST=1 VALGRIND=1 -j$(nproc)
cd build/bin
valgrind --leak-check=full --show-leak-kinds=all -q --error-exitcode=1 ./https_client_component_test
```

Reproduces roughly **1 run in 3–5**. It does **not** reproduce with `--gtest_filter` on that test alone.

## Observations

| scenario | result |
|---|---|
| full 38-test suite under valgrind, `5.0.0-https` | passed (38/38), 140 s |
| full suite under valgrind, with PR #38320 | 1 failure in 3 runs |
| full suite under valgrind, startup bound raised 3 s → 10 s | 1 failure in 4 runs — **no improvement** |
[...]

PR #38786, opened 2026-09-01
Enforce TLS verification by default in the agent's HTTPS transport

### LINKED ISSUE #38684: Agent HTTPS transport defaults to verification_mode none, so a default install performs no TLS certificate verification
|---|---|---|---|---|
| 5.1.0-alpha0 | wazuh-agentd (https_client transport) | Agent | Sources | Linux |

## Summary

The agent's new HTTPS stack is its only transport, and on a default install it performs no TLS verification at all, neither certificate chain nor hostname. The insecure default is set explicitly in code, and no shipped configuration template contains an `<ssl>` block, so this is the posture most agents would actually run with.

This is filed as an issue rather than a security advisory because the code is not present in any published release. The field is absent from `v5.0.0-beta4`, and the file does not exist in the 4.x line, so no deployed user is affected today. The purpose of this report is to prevent it shipping in 5.0.0 GA.

## Affected code

`src/client-agent/src/config.c:58`:

```c
/* The shipped configuration carries no <ssl> block, so this is the posture most
 * agents actually run with -- as is any config written before the HTTPS transport
 * existed. It is not the enum's zero value (FULL), which would make every one of
 * those agents refuse to start on a missing CA, so it has to be set by hand. */
agt->ssl.verification_mode = AGENT_VERIFY_NONE;
```
The default propagates to the socket options as follows:

1. src/client-agent/src/https_client_bridge.c:1567 maps the config value to HC_VERIFY_NONE.
2. src/client-agent/https_client/src/curlPerformer.cpp:275-283 resolves both flags to false:

```
const bool verifyPeer = m_config.verifyMode != HC_VERIFY_NONE;   // false
const bool verifyHost = m_config.verifyMode == HC_VERIFY_FULL;   // false
```

3. src/client-agent/https_client/src/curlHandle.cpp:56-57 therefore issues 
```
CURLOPT_SSL_VERIFYPEER = 0 and CURLOPT_SSL_VERIFYHOST = 0
```

HTTPS is not optional for the agent. From src/client-agent/src/https_client_bridge.c:8-22:

> HTTPS is the agent's transport unconditionally, there is no internal-option gate [...] and no alternative transport is offered.

A search of etc/templates/config/ returns no agent template containing verification_mode or certificate_authorities.

The asymmetry that suggests this is unintended

The manager side already has the safety net the agent side lacks. From src/config/src/remote-config.c:383-387:

```
if (logr->https.ca != NULL && logr->https.verification_mode == REMOTED_HTTPS_VERIFY_UNSET) {
    mwarn("The '<remote><https><ca>' option is configured but '<verification_mode>' is not; "
          "defaulting '<verification_mode>' to 'certificate'.");
    logr->https.verification_mode = REMOTED_HTTPS_VERIFY_CERTIFICATE;
}
```

The comment above that block states the intent is to avoid "silently leaving client-certificate verification disabled". Read_Agent_SSL() in src/config/src/client-config.c has no equivalent inference and writes verification_mode only when the tag is present. An operator who configures <ssl><certificate_authorities> but omits <verification_mode> therefore gets zero verification while believing trust is pinned.

Two further signs the current default is not the intended one:

- src/client-agent/include/client-config.h documents the field as defaulting to FULL, while the applied default is NONE.
- src/client-agent/https_client/src/retrySender.cpp:216 reasons about its own behaviour "under verify_mode=none (opt-in, insecure)". The mode is opt-out, not opt-in.

### Impact

An attacker positioned on the agent to manager path (rogue DHCP or DNS, ARP spoofing, hostile Wi-Fi, a compromised upstream hop) can terminate the TLS session as the manager with no credentials and no user interaction. The agent typically runs as root or SYSTEM. From that position the attacker can:

- Read all agent telemetry in plaintext.
- Drive re-enrollment by returning 401, reaching bridge_on_reenroll_required at https_client_bridge.c:405, and write an attacker chosen id, name, and key into client.keys.
- Reach command execution as root by issuing a /control task of type active_response, which bridge_dispatch_active_response at src/client-agent/src/https_client_bridge.c:529 forwards to execd, which runs the named script under active-response/bin.

Enrollment runs over the same unverified channel, and enrollment authentication is one way from agent to manager, so the manager is authenticated solely by TLS.

### Reproduction

1. Install an agent using the shipped configuration, with no <ssl> block present.
2. Redirect the configured manager address to a host you control and present any self signed certificate.
3. The agent completes the handshake and proceeds normally. No warning appears at the default log level, because the mode is recorded at INFO in moduleConfig.cpp.

### Suggested fix

1. Default verification_mode to full.
2. If backward compatibility requires a permissive default, at minimum port the manager side inference from remote-config.c:383-387 so that configuring <certificate_authorities> without <verification_mode> enables verification rather than silently disabling it.
3. Raise the "verification disabled" log line from INFO to WARNING so operators can observe the posture.
4. Add an explicit <ssl> block to the shipped agent templates so the intended posture is visible in the config.

cborla:
## Description

On a default install (no `<agent><ssl>` block, which no shipped agent template carries), the agent's HTTPS transport performed zero TLS verification against the manager: `ClientConf()` in `client-agent/src/config.c` set `verification_mode` to `AGENT_VERIFY_NONE` unconditionally before parsing, with nothing to override it. Reported and reproduced upstream (@vikman90) as #38684; this PR implements the fix direction settled there plus the four concrete requirements the team lead gave ahead of the RC1 tag: enforce by default, route the installer's CA env var to the tag the agent actually reads, gate WPK upgrades on having a usable CA, and confirm the OS trust store still lets a default install reach a publicly-trusted endpoint (e.g. cloud.wazuh.com) with zero config.

Closes #38684

## Proposed Changes

| File | Change |
|---|---|
| `config/include/client-config.h`, `client-agent/src/config.c` | New `AGENT_VERIFY_UNSET` sentinel. `ClientConf()` now resolves it after parsing (post both `ReadConfig()` calls, so a shared/remote config override is accounted for): `AGENT_VERIFY_CERT` if `<certificate_authorities>` was configured without an explicit `<verification_mode>` (mirrors the manager's own inference in `remote-config.c`), otherwise `AGENT_VERIFY_SYSTEM` (OS trust store). `AGENT_VERIFY_NONE` is now reachable only via an explicit `<verification_mode>none</verification_mode>`. |
| `client-agent/https_client/src/moduleConfig.cpp` | The "TLS verification is DISABLED" log line is now `WARNING` instead of `INFO`, since it can now only be reached via a deliberate operator opt-out, not the default posture. |
| `init/register_configure_agent.sh` | New `set_agent_ssl_ca()`. `WAZUH_REGISTRATION_CA` now also writes `<agent><ssl><certificate_authorities>` — previously it only reached the deprecated `<enrollment><server_ca_path>` tag, which the 5.x parser recognizes but ignores (enrollment reuses `<agent><ssl>` now), so an operator-supplied CA silently never took effect. Refuses to write the CA (and logs why) when `<verification_mode>` is already explicitly `system`, since the agent rejects that combination at runtime. |
| `init/pkg_installer.sh` (Linux WPK), `win32/do_upgrade.ps1` (Windows WPK) | A WPK upgrade now gates on the resolved `verification_mode` actually being able to connect post-upgrade, at the same point manager reachability is already checked. `full`/`certificate` still require a readable configured CA. For `system` (the new default), the gate does a **live TLS probe** against the manager using the OS's own trust store (distinct from the existing reachability probe, which deliberately skips certificate verification) — if that already verifies, the upgrade proceeds untouched; if not, it looks for a CA at a new default drop-in path (`/var/ossec/etc/certs/root-ca.pem` / `<installdir>\certs\root-ca.pem`) and pins it into `<certificate_authorities>` if found; if there's nothing to pin and `system` wasn't explicit, or if `system` *was* explicit (pinning a CA under an explicit `system` is rejected at runtime, so there's nothing safe to auto-fix), the upgrade aborts with a clear reason and the host stays on its current version instead of silently losing connectivity. |
| `init/register_configure_agent.sh`, `win32/InstallerScripts.vbs`, `win32/wazuh-installer.wxs` | New install-time variable `SSL_VERIFICATION` (`full`/`certificate`/`system`/`none`) writes `<agent><ssl><verification_mode>` directly at install time, mirroring how `WAZUH_REGISTRATION_CA` already routes to `certificate_authorities`. Requested by the dashboard team's enrollment-wizard "SSL verification" toggle (wazuh-dashboard-plugins#9103) and tracked for DevOps' install repos (wazuh-automation#3778); runs before `WAZUH_REGISTRATION_CA`'s own write so the existing `verification_mode=system`-vs-CA conflict check sees it. On Windows this required a new `SSL_VERIFICATION` MSI property (`wazuh-installer.wxs`) threaded through `CustomActionData` to reach the installer script. Also hardened the existing CA/verification_mode writes in `register_configure_agent.sh`: scoped `agent_option_is_set()`/`agent_option_value()` and the actual write to `<agent><ssl>` (previously unscoped — could read or overwrite a same-named tag anywhere else in the file), and fixed a silent no-op on a pretty-printed, multi-line `<certificate_authorities>` tag. |

### Results and Evidence

Live end-to-end against a real Wazuh 5.0.0 manager (self-signed certificate on the agent-facing HTTPS listener, port 1517):

Agent with no `<ssl>` config (the shipped default) — fails closed, as intended:
```
wazuh-agentd: ERROR: Enrollment request could not be sent: (60) SSL peer certificate or SSH remote key was not OK: SSL certificate OpenSSL verify result: self-signed certificate (18).
```
Before this fix, the same agent connected to the same manager with no verification and no warning of any kind.

Agent with the manager's CA configured (the exact end state `register_configure_agent.sh`'s fix produces from `WAZUH_REGISTRATION_CA`) — enrolls and connects:
```
wazuh-agentd: WARNING: The '<ssl><certificate_authorities>' option is configured but '<verification_mode>' is not; defaulting '<verification_mode>' to 'certificate'.
```
The agent's own live-queried config (`getconfig agent` over its local socket) confirms the inferred mode: `"ssl":{"verification_mode":"certificate","certificate_authorities":"/tmp/manager-ca.pem"}`.

A real 4.14.7 → 5.0.0 upgrade via the genuine signed nightly WPK (`agent_upgrade -f wazuh_agent_v5.0.0_linux_amd64.deb.wpk`) against the same manager reproduces the exact risk the WPK gate above exists to catch: the stock post-migration agent logs `TLS verification is DISABLED (verify_mode=none)` and connects; running that same post-migration config under this PR's patched binary instead, the agent logs `WARNING: Manager unreachable. Pausing module event production... self-signed certificate (18)` — every existing on-prem agent that upgrades without a CA already in place goes dark. Posted as a PR comment with full detail; the WPK gate added here is the fix for that gap.

The WPK gate's own logic (probe → pin-CA → abort branches) was verified against a local self-signed TLS server and a real publicly-trusted endpoint (github.com), covering: no CA anywhere → abort; a CA present at the default path → pinned, proceeds; the OS store already trusts the manager → proceeds untouched; `verification_mode=system` explicit and unverifiable → abort (can't auto-fix); `verification_mode=system` explicit *with* a leftover `certificate_authorities` also set → abort as a config error, independent of whether the manager happens to verify.

  - [x] Linux
  - [x] Windows
  - [x] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests _(Wazuh v4.x)_: N/A — 5.0.0 agent-side change only
- Engine _(Wazuh v5.x and above)_: N/A — no engine changes
- Wazuh server API/Framework: N/A — no API/framework changes

Test evidence is split into dedicated comments below, one per category, so this section stays short:

- [Unit tests (C)](https://github.com/wazuh/wazuh/pull/38786#issuecomment-5519143901) — `test_client_conf_ssl_resolution` (new, 5/5), `test_agent_config` (16/16), `test_client-config_https` (86/86), `https_client_unit_test` (400/400).
- [Fresh-install CA routing](https://github.com/wazuh/wazuh/pull/38786#issuecomment-5519144136) — live test on a real manager+agent (Ubuntu), both the default-refuses and `WAZUH_REGISTRATION_CA`-connects scenarios.
- [WPK upgrade gate — Linux/macOS](https://github.com/wazuh/wazuh/pull/38786#issuecomment-5519144399) — 10-scenario consolidated regression of `pkg_installer.sh` against the final code.
- [WPK upgrade gate — Windows](https://github.com/wazuh/wazuh/pull/38786#issuecomment-5519144644) — same 10 scenarios, run for real on a Windows 11 guest via WinRM.
- [Real signed WPK — Windows](https://github.com/wazuh/wazuh/pull/38786#issuecomment-5519144881) — a genuine signed WPK built through the production pipeline for this branch's head commit, including its own passing built-in upgrade test.
- [Real 4.x → 5.0.0 migration — Linux](https://github.com/wazuh/wazuh/pull/38786#issuecomment-5519145086) — the original real-WPK migration repro that motivated the gate.
- [Real manager-driven upgrade — Linux, both outcomes](https://github.com/wazuh/wazuh/pull/38786#issuecomment-5525396695) — a real manager dispatching a real signed WPK from this branch's current head via `agent_upgrade`, against two real agents: one with the manager's cert trusted (upgrades successfully) and one without (gate aborts, agent stays on its working version, no lost connectivity).
- [`SSL_VERIFICATION` install-time variable](https://github.com/wazuh/wazuh/pull/38786#issuecomment-5530088364) — 9 scenarios on Linux/macOS, 10 on Windows run for real under `wine cscript.exe`, including combined `SSL_VERIFICATION` + `WAZUH_REGISTRATION_CA` conflict/co-existence cases and the `wazuh-installer.wxs` MSI property wiring.

CI note: `rtr (client-agent/https_client, agent)` has failed under Valgrind four separate times on this branch, each time on a different test in the same job, none touching code this PR changes. This matches a known, already-tracked issue: #38329, which documents thread starvation in the `https_client_component_test` binary under Valgrind's single-thread scheduling — threads that outlive their test accumulate across the suite, and a client started late in the run can go unscheduled long enough to blow any wait deadline regardless of size. Not a regression from this change.

### Artifacts Affected

- `wazuh-agentd` (agent binary): default TLS verification posture changes from none to `system`/`certificate`.
- `libhttps_client.so`: log level change only (INFO → WARNING for the disabled-verification message).
- Agent installer packages (deb/rpm, all platforms) that invoke `register_configure_agent.sh`: `WAZUH_REGISTRATION_CA` now has an effect it previously silently lacked, and a new `SSL_VERIFICATION` variable is recognized.
- Linux and Windows WPK upgrade packages: the new pre-install gate can abort an upgrade that would previously have proceeded and left the agent unable to connect; a CA dropped at `/var/ossec/etc/certs/root-ca.pem` (or `<installdir>\certs\root-ca.pem` on Windows) ahead of time is now picked up automatically.
- Windows MSI: new `SSL_VERIFICATION` public property (`Secure="yes"`), settable via `msiexec ... SSL_VERIFICATION=<value>`.

### Configuration Changes

`<agent><ssl><verification_mode>`'s effective default changes from `none` to `system` (or `certificate` when `<certificate_authorities>` is set without an explicit mode). This is the intended behavior change this PR exists to make — see #38684. No new configuration tags are introduced; `<verification_mode>none</verification_mode>` remains available as an explicit opt-out. New convention (not a config tag): `/var/ossec/etc/certs/root-ca.pem` / `<installdir>\certs\root-ca.pem` as the default drop-in location the WPK upgrade gate checks for a CA.

New install-time variable: `SSL_VERIFICATION` (`full`/`certificate`/`system`/`none`, case-sensitive, matching `Read_Agent_SSL()`'s own parsing), writes `<agent><ssl><verification_mode>` at install time on Linux/macOS and Windows. An invalid value is rejected with a log line rather than written; a value combined with `WAZUH_REGISTRATION_CA` that conflicts with an explicit `system` is handled by the existing conflict check (the CA is dropped, not the mode).

### Tests Introduced

Updated `unit_tests/client-agent/test_agent_config.c`: the `setup_agent()` fixture and `test_reports_default_ssl_posture` now assert the new default (`AGENT_VERIFY_SYSTEM` / `"system"`) instead of the old `AGENT_VERIFY_NONE` / `"none"`. No new unit test files added; the `ClientConf()`-level UNSET-resolution logic (the CA-present-without-explicit-mode → `certificate` inference) is covered by the live VM verification above rather than by an automated test — flagging this as a gap worth closing with a follow-up `ClientConf()`-level test if reviewers agree it's worth the harness work.

Updated 15 integration-test fixtures that target `RemotedSimulator` (self-signed cert) to opt out via `<ssl><verification_mode>none</verification_mode>`, since none of them are testing TLS trust themselves: `test_agentd` (reconnection, startup-hash-validation, state, state_config), `test_enrollment` (agent, options), `test_execd`, `test_github`, `test_office365`, `test_sca`, `test_syscollector` (×4), and `test_fim`'s module-scoped `autouse` fixture (`conftest.py`) — the last one affects every FIM integration test module, not a single test case, and was found by an external review pass since it wasn't in the set of CI jobs this PR's own changes trigger.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

This PR has been through several review passes (external and self) that found and fixed real issues in the WPK-gate/CA-pinning code on both platforms: silent no-op failures, a missing `system`+CA conflict guard, 3 more `RemotedSimulator` test fixtures needing the same TLS opt-out, case-sensitive-vs-insensitive mismatches between the shell/PowerShell gates and the C parser, comment-blindness in the XML helpers (a commented-out tag was still read as live), and `pin_ca()` silently pinning a CA under an `<agent>`-less `<client>` block that the real parser never reads — see PR comments and commit history for the itemized list. Every fix was re-verified against real fixtures on Linux/macOS and, for the Windows-side code, on a real Windows 11 guest.

### DIFF
--- a/src/init/pkg_installer.sh
+++ b/src/init/pkg_installer.sh
@@ -287,6 +287,100 @@ probe_server() {
     return $?
 }
 
+# Same target as probe_server(), but without -k: succeeds only if the system's own
+# trust store actually verifies the manager's certificate. probe_server() cannot tell
+# us this -- it deliberately skips verification so a plain reachability check never
+# depends on TLS trust -- but it is exactly what AGENT_VERIFY_SYSTEM needs to work
+# post-upgrade (#38684). No TCP/TLS-1.3 fallback here: a client that cannot do the
+# real handshake cannot tell us whether the cert is trusted, so treat that as
+# "unverified" rather than assume trust.
+probe_server_verified() {
+    PROBE_TIMEOUT=5
+
+    PROBE_HOST="${1}"
+    case "${PROBE_HOST}" in
+        \[*) ;;
+        *:*:*) PROBE_HOST="[${PROBE_HOST}]" ;;
+    esac
+
+    if [ -z "${3}" ]; then
+        PROBE_PATH="/"
+    else
+        PROBE_PATH="/${3}/"
+    fi
+
+    if command -v curl > /dev/null 2>&1; then
+        curl --tlsv1.3 -s -f -m ${PROBE_TIMEOUT} -o /dev/null "https://${PROBE_HOST}:${2}${PROBE_PATH}"
+        return $?
+    elif command -v wget > /dev/null 2>&1; then
+        wget -q --timeout=${PROBE_TIMEOUT} --tries=1 -O /dev/null "https://${PROBE_HOST}:${2}${PROBE_PATH}"
+        return $?
+    else
+        return 1
+    fi
+}
+
+# True (exit 0) if <block><sub> exists at all, regardless of what it contains --
+# distinct from xml_tag_present, which looks for a specific leaf tag. Needed so
+# pin_ca() inserts into an existing (but otherwise unrelated, e.g. <ciphers>-only)
+# <ssl> block instead of creating a second, duplicate one.
+xml_block_present() {
+    tr -d '\n\r' < ./etc/ossec.conf 2>/dev/null | grep -o "<$1>.*</$1>" | grep -q "<$2>"
+}
+
+# Pin a CA file into <agent><ssl><certificate_authorities>, creating the <ssl> block
+# if the config does not have one yet. Mirrors set_agent_ssl_ca() in
+# register_configure_agent.sh; duplicated because this script ships inside the WPK
+# and runs standalone, with nothing to source. Only called when
+# xml_tag_present agent ssl certificate_authorities is false, so it never overwrites
+# an operator-configured CA.
+pin_ca() {
+    CA_PATH="${1}"
+    TMP_PIN_CONF="$(mktemp)"
+
+    if xml_block_present agent ssl; then
+        awk -v ca="${CA_PATH}" '
+            in_comment {
+                if ($0 ~ /-->/) { in_comment = 0 }
+                print
+                next
+            }
+            !inserted && /^[[:space:]]*<ssl>[[:space:]]*$/ {
+                print
+                print "      <certificate_authorities>" ca "</certificate_authorities>"
+                inserted = 1
+                next
+            }
+            {
+                if ($0 ~ /<!--/ && $0 !~ /-->/) { in_comment = 1 }
+                print
+            }
+        ' ./etc/ossec.conf > "${TMP_PIN_CONF}" && cat "${TMP_PIN_CONF}" > ./etc/ossec.conf
+    else
+        awk -v ca="${CA_PATH}" '
+            in_comment {
+                if ($0 ~ /-->/) { in_comment = 0 }
+                print
+                next
+            }
+            !inserted && /^[[:space:]]*<(agent|client)>[[:space:]]*$/ {
+                print
+                print "    <ssl>"
+                print "      <certificate_authorities>" ca "</certificate_authorities>"
+                print "    </ssl>"
+                inserted = 1
+                next
+            }
+            {
+                if ($0 ~ /<!--/ && $0 !~ /-->/) { in_comment = 1 }
+                print
+            }
+        ' ./etc/ossec.conf > "${TMP_PIN_CONF}" && cat "${TMP_PIN_CONF}" > ./etc/ossec.conf
+    fi
+
+    rm -f "${TMP_PIN_CONF}"
+}
+
 # A WPK upgrade never rewrites ossec.conf, so this script meets two config shapes and has
 # to read both (#38624):
 #
@@ -361,6 +455,8 @@ fi
 # is gone, rather than leaving a freshly-upgraded host silently offline.
 SSL_VERIFICATION_MODE=$(xml_value agent ssl verification_mode)
 SSL_CA=$(xml_value agent ssl certificate_authorities)
+SSL_VERIFICATION_MODE_EXPLICIT=0
+xml_tag_present agent ssl verification_mode && SSL_VERIFICATION_MODE_EXPLICIT=1
 
 if [ -z "${SSL_VERIFICATION_MODE}" ]; then
     if [ -n "${SSL_CA}" ]; then
@@ -370,10 +466,51 @@ if [ -z "${SSL_VERIFICATION_MODE}" ]; then
     fi
 fi
 
+# Default drop-in location for the manager's CA (mirrored on Windows in
+# do_upgrade.ps1): an operator can place it here ahead of an upgrade without having
+# to hand-edit ossec.conf. Resolves to /var/ossec/etc/certs/root-ca.pem on a default
+# install.
+DEFAULT_CA_FILE="./etc/certs/root-ca.pem"
+
 case "${SSL_VERIFICATION_MODE}" in
     full|certificate)
         if [ -z "${SSL_CA}" ] || [ ! -r "${SSL_CA}" ]; then
             echo "$(date +"%Y/%m/%d %H:%M:%S") - Upgrade failed. <ssl><verification_mode> is '${SSL_VERIFICATION_MODE}' but <certificate_authorities> ('${SSL_CA}') is missing or unreadable, interrupting upgrade." >> ./logs/upgrade.log
+            abort_upgrade "2"
+        fi
+        ;;
+    system)
+        # verification_mode=system with a certificate_authorities also set is rejected
+        # outright at runtime (validateTls() in moduleConfig.cpp) regardless of
+        # whether the manager's certificate happens to verify against the OS store --
+        # catch the config error itself here rather than let a live probe that
+        # happens to pass mask a daemon that will refuse to start.
+        if [ -n "${SSL_CA}" ]; then
+            echo "$(date +"%Y/%m/%d %H:%M:%S") - Upgrade failed. <ssl><verification_mode> is 'system' but <certificate_authorities> ('${SSL_CA}') is also set; 'system' trusts the OS store, not a configured CA, and the agent refuses to start with both set. Remove <certificate_authorities>, or switch to <verification_mode>certificate</verification_mode>, interrupting upgrade." >> ./logs/upgrade.log
+            abort_upgrade "2"
+        fi
+
+        # 'system' trusts the OS store, not a configured CA -- probe_server() above
+        # cannot tell us whether that store actually trusts THIS manager's
+        # certificate, since it deliberately skips verification (-k) so the plain
+        # reachability check never depends on TLS trust. Find out for real before
+        # assuming the freshly-upgraded agent will still be able to connect.
+        if [ "${WAZUH_UPGRADE_TEST_SKIP_MANAGER_CHECK}" = "1" ]; then
+            echo "$(date +"%Y/%m/%d %H:%M:%S") - System CA trust check skipped (test mode)." >> ./logs/upgrade.log
+        elif probe_server_verified "${SERVER_ADDRESS}" "${SERVER_PORT}" "${SERVER_ENDPOINT}"; then
+            echo "$(date +"%Y/%m/%d %H:%M:%S") - The system trust store already verifies the manager's certificate; proceeding under verify_mode=system." >> ./logs/upgrade.log
+        elif [ "${SSL_VERIFICATION_MODE_EXPLICIT}" = "1" ]; then
+            # <verification_mode>system</verification_mode> was set explicitly:
+            # pinning a CA here would be rejected at runtime (validateTls() in
+            # moduleConfig.cpp refuses system+certificate_authorities together), so
+            # there is nothing this script can safely fix on the operator's behalf.
+            echo "$(date +"%Y/%m/%d %H:%M:%S") - Upgrade failed. <ssl><verification_mode> is explicitly 'system' but the system trust store does not verify the manager's certificate at ${SERVER_ADDRESS}:${SERVER_PORT}. Import it into the OS trust store, or switch to <verification_mode>certificate</verification_mode> with a <certificate_authorities> path, interrupting upgrade." >> ./logs/upgrade.log
+            abort_upgrade "2"
+        elif [ -r "${DEFAULT_CA_FILE}" ]; then
+            pin_ca "${DEFAULT_CA_FILE}"
+            echo "$(date +"%Y/%m/%d %H:%M:%S") - The system trust store does not verify the manager's certificate; pinned ${DEFAULT_CA_FILE} as <certificate_authorities> instead." >> ./logs/upgrade.log
+        else
+            echo "$(date +"%Y/%m/%d %H:%M:%S") - Upgrade failed. The system trust store does not verify the manager's certificate at ${SERVER_ADDRESS}:${SERVER_PORT}, and no CA was found at ${DEFAULT_CA_FILE}. Place the manager's CA there, or configure <certificate_authorities> explicitly, then retry the upgrade; staying on the current version, interrupting upgrade." >> ./logs/upgrade.log
             abort_upgrade "2"
         fi
         ;;
--- a/src/init/pkg_installer.sh
+++ b/src/init/pkg_installer.sh
@@ -359,7 +359,7 @@ probe_server_verified() {
 # pin_ca() inserts into an existing (but otherwise unrelated, e.g. <ciphers>-only)
 # <ssl> block instead of creating a second, duplicate one.
 xml_block_present() {
-    strip_xml_comments | tr -d '\n\r' | grep -o "<$1>.*</$1>" | grep -q "<$2>"
+    strip_xml_comments | tr -d '\n\r' | grep -o "<$1>.*</$1>" | grep -qE "<$2>|<$2[ \t]*/>"
 }
 
 # Pin a CA file into <agent><ssl><certificate_authorities>, creating the <ssl> block
--- a/src/init/pkg_installer.sh
+++ b/src/init/pkg_installer.sh
@@ -542,6 +542,19 @@ case "${SSL_VERIFICATION_MODE}" in
             abort_upgrade "2"
         fi
         ;;
+    none)
+        # Explicitly disabled -- nothing for this gate to check.
+        ;;
+    *)
+        # Neither ReadConfig() nor this gate's own default-resolution above can produce
+        # anything but full/certificate/system/none, so getting here means ossec.conf
+        # carries something else (a typo, wrong case, hand-edited garbage). Read_Agent_SSL()
+        # rejects that value case-sensitively too, so letting the upgrade proceed would just
+        # trade this loud failure for the new binary refusing to start after the old one is
+        # already gone.
+        echo "$(date +"%Y/%m/%d %H:%M:%S") - Upgrade failed. <ssl><verification_mode> is '${SSL_VERIFICATION_MODE}', which is not a value this agent recognizes (full, certificate, system, or none); interrupting upgrade." >> ./logs/upgrade.log
+        abort_upgrade "2"
+        ;;
 esac
 
 if [[ "$OS" == "Darwin" ]]; then
--- a/src/init/pkg_installer.sh
+++ b/src/init/pkg_installer.sh
@@ -61,6 +61,15 @@ else
     abort_upgrade "2"
 fi
 
+# Escapes the three characters that are structurally significant in XML content --
+# '&', '<', '>' -- so a value written verbatim into ossec.conf (a CA path, in
+# particular) can never be mistaken for markup or break the file's well-formedness.
+# '&' must run first: escaping '<'/'>' introduces new literal '&' characters (as part
+# of "&lt;"/"&gt;") that must not themselves be re-escaped by a later pass.
+xml_escape() {
+    printf '%s' "$1" | sed -e 's/&/\&amp;/g' -e 's/</\&lt;/g' -e 's/>/\&gt;/g'
+}
+
 # Strips commented-out lines before xml_value/xml_tag_present/xml_block_present extract
 # anything, so a tag an operator comments out (e.g. to fall back to the default) reads as
 # absent here too, matching OS_XML's own comment handling -- same in_comment
@@ -364,7 +373,7 @@ xml_block_present() {
 # found -- e.g. an existing <ssl> block whose opening tag isn't alone on its own line --
 # rather than silently reporting success with nothing actually pinned.
 pin_ca() {
-    CA_PATH="${1}"
+    CA_PATH="$(xml_escape "${1}")"
     TMP_PIN_CONF="$(mktemp)"
     PIN_OK=1
 
@@ -528,7 +537,7 @@ DEFAULT_CA_FILE="./etc/certs/root-ca.pem"
 
 case "${SSL_VERIFICATION_MODE}" in
     full|certificate)
-        if [ -z "${SSL_CA}" ] || [ ! -r "${SSL_CA}" ]; then
+        if [ -z "${SSL_CA}" ] || [ ! -f "${SSL_CA}" ] || [ ! -r "${SSL_CA}" ]; then
             echo "$(date +"%Y/%m/%d %H:%M:%S") - Upgrade failed. <ssl><verification_mode> is '${SSL_VERIFICATION_MODE}' but <certificate_authorities> ('${SSL_CA}') is missing or unreadable, interrupting upgrade." >> ./logs/upgrade.log
             abort_upgrade "2"
         fi
@@ -560,7 +569,7 @@ case "${SSL_VERIFICATION_MODE}" in
             # there is nothing this script can safely fix on the operator's behalf.
             echo "$(date +"%Y/%m/%d %H:%M:%S") - Upgrade failed. <ssl><verification_mode> is explicitly 'system' but the system trust store does not verify the manager's certificate at ${SERVER_ADDRESS}:${SERVER_PORT}. Import it into the OS trust store, or switch to <verification_mode>certificate</verification_mode> with a <certificate_authorities> path, interrupting upgrade." >> ./logs/upgrade.log
             abort_upgrade "2"
-        elif [ -r "${DEFAULT_CA_FILE}" ]; then
+        elif [ -f "${DEFAULT_CA_FILE}" ] && [ -r "${DEFAULT_CA_FILE}" ]; then
             if pin_ca "${DEFAULT_CA_FILE}"; then
                 echo "$(date +"%Y/%m/%d %H:%M:%S") - The system trust store does not verify the manager's certificate; pinned ${DEFAULT_CA_FILE} as <certificate_authorities> instead." >> ./logs/upgrade.log
             else
--- a/src/init/pkg_installer.sh
+++ b/src/init/pkg_installer.sh
@@ -354,6 +354,31 @@ else
     echo "$(date +"%Y/%m/%d %H:%M:%S") - Manager reachable at ${SERVER_ADDRESS}:${SERVER_PORT}/${SERVER_ENDPOINT}." >> ./logs/upgrade.log
 fi
 
+# The upgrade replaces the agent's binaries but not its ossec.conf, so the TLS
+# posture the new agent boots under is exactly what's on disk now (#38684). A
+# verifying mode with no readable CA can never connect -- mirrors
+# w_agent_validate_ssl_ca() in config.c -- so catch it here, before the old agent
+# is gone, rather than leaving a freshly-upgraded host silently offline.
+SSL_VERIFICATION_MODE=$(xml_value agent ssl verification_mode)
+SSL_CA=$(xml_value agent ssl certificate_authorities)
+
+if [ -z "${SSL_VERIFICATION_MODE}" ]; then
+    if [ -n "${SSL_CA}" ]; then
+        SSL_VERIFICATION_MODE="certificate"
+    else
+        SSL_VERIFICATION_MODE="system"
+    fi
+fi
+
+case "${SSL_VERIFICATION_MODE}" in
+    full|certificate)
+        if [ -z "${SSL_CA}" ] || [ ! -r "${SSL_CA}" ]; then
+            echo "$(date +"%Y/%m/%d %H:%M:%S") - Upgrade failed. <ssl><verification_mode> is '${SSL_VERIFICATION_MODE}' but <certificate_authorities> ('${SSL_CA}') is missing or unreadable, interrupting upgrade." >> ./logs/upgrade.log
+            abort_upgrade "2"
+        fi
+        ;;
+esac
+
 if [[ "$OS" == "Darwin" ]]; then
     installer -pkg ./var/upgrade/wazuh-agent* -target / >> ./logs/upgrade.log 2>&1
 elif [[ "$OS" == "Linux" ]]; then
--- a/src/init/pkg_installer.sh
+++ b/src/init/pkg_installer.sh
@@ -291,9 +291,11 @@ probe_server() {
 # trust store actually verifies the manager's certificate. probe_server() cannot tell
 # us this -- it deliberately skips verification so a plain reachability check never
 # depends on TLS trust -- but it is exactly what AGENT_VERIFY_SYSTEM needs to work
-# post-upgrade (#38684). No TCP/TLS-1.3 fallback here: a client that cannot do the
-# real handshake cannot tell us whether the cert is trusted, so treat that as
-# "unverified" rather than assume trust.
+# post-upgrade. No TCP fallback here: a client that cannot do the real handshake
+# cannot tell us whether the cert is trusted, so treat that as "unverified" rather
+# than assume trust -- this deliberately stays fail-closed even for the curl-too-old
+# case below, since there is no way to positively confirm trust without the
+# handshake; only the log message distinguishes the two causes for the operator.
 probe_server_verified() {
     PROBE_TIMEOUT=5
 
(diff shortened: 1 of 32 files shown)


cborla:
⚠️ This fix has no migration path for an existing 4.x fleet with a self-signed manager certificate (the typical on-prem case) — an in-place WPK upgrade to a build carrying this fix silently stops the agent's event production, with no way today to provision the manager's CA as part of that upgrade.

Reproduced live: enrolled a real 4.14.7 agent (legacy `agent-auth`, connected over TCP/1514) against a manager with a self-signed cert, then upgraded it with the real signed nightly 5.0.0 WPK via `agent_upgrade -a <id> -f wazuh_agent_v5.0.0_linux_amd64.deb.wpk -F`. dpkg went 4.14.7-1 → 5.0.0-latest cleanly, preserving the legacy `<client><server>` block through the 5.x compat layer — no `<ssl>` block, since 4.x never had the concept.

With the stock (unpatched) binaries from that real migration, the vulnerability reproduces exactly as in #38684, via a migration rather than a fresh install: `wazuh-agentd:https-client: INFO: TLS verification is DISABLED (verify_mode=none).`

Took that same post-migration `ossec.conf`/`client.keys` and ran it under this branch's patched binary instead:
```
wazuh-agentd: WARNING: Manager unreachable. Pausing module event production. Reason: (60) SSL peer certificate or SSH remote key was not OK: SSL certificate OpenSSL verify result: self-signed certificate (18)
```
Fails closed as designed, but that means every agent in this shape goes dark on upgrade unless the manager's CA is already in `<agent><ssl><certificate_authorities>` beforehand. `WAZUH_REGISTRATION_CA` (this PR's install-time fix) doesn't run on an in-place WPK upgrade, and `pkg_installer.sh`'s new CA-gate doesn't catch this either — from a file-existence standpoint `system` mode needs no CA, the actual failure is a live TLS trust decision the gate can't evaluate.

Flagging this as an open question for reviewers before merge, not something I've resolved: what's the intended migration story for an existing on-prem fleet here?

cborla:
Resolved. `pkg_installer.sh`/`do_upgrade.ps1` now do exactly the live TLS trust decision this comment said the gate couldn't evaluate: a real probe against the manager using the OS trust store (`probe_server_verified`, separate from the existing reachability-only probe), and if that doesn't verify, a fallback to a CA dropped at a new default path (`/var/ossec/etc/certs/root-ca.pem` / `<installdir>\certs\root-ca.pem`), pinned automatically into `<certificate_authorities>` if present. If neither the OS store nor the default path can verify the manager, the upgrade now aborts instead of silently going dark — the migration story is: place the manager's CA at that default path before upgrading a fleet that doesn't already have it in the OS trust store.

Verified against a local self-signed server and a real publicly-trusted endpoint, covering the no-CA-anywhere (abort), CA-at-default-path (pinned, proceeds), and OS-store-already-trusts (proceeds untouched) cases.

cborla:
⚠️ `rtr (client-agent/https_client, agent)` keeps failing under Valgrind, unrelated to this PR. Already reported in #38329.

src/win32/do_upgrade.ps1
```diff
@@ -542,6 +630,83 @@ if ($env:WAZUH_UPGRADE_TEST_SKIP_MANAGER_CHECK -eq "1") {
...
+# pkg_installer.sh): an operator can place it here ahead of an upgrade without having
+# to hand-edit ossec.conf.
+$default_ca_file = Join-Path $wazuhDir "certs\root-ca.pem"
+
+if ($ssl_verification_mode -eq "full" -or $ssl_verification_mode -eq "certificate") {
+    if ([string]::IsNullOrEmpty($ssl_ca) -or -Not (Test-Path -PathType Leaf $ssl_ca)) {
+        write-output "$(Get-Date -format u) - Upgrade failed: <ssl><verification_mode> is '$($ssl_verification_mode)' but <certificate_authorities> ('$($ssl_ca)') is missing or unreadable, interrupting upgrade." >> .\upgrade\upgrade.log
+        abort_upgrade "2"
+    }
+} elseif ($ssl_verification_mode -eq "system") {
```
lchico: This gate's `verification_mode` comparisons (here and lines 655/698) use PowerShell's `-eq`, which is case-insensitive by default. `<verification_mode>System</verification_mode>` (capitalized, e.g. hand-edited) matches `-eq "system"` here and the gate probes/proceeds normally — but `Read_Agent_SSL()` in `src/config/src/client-config.c` does a case-sensitive `strcmp` against `"system"`, falls through to its `else`, and returns `OS_INVALID`, so `ClientConf()` fails and the freshly-upgraded agent never starts even though this gate said it was fine. Could these comparisons use a case-sensitive match (e.g. `-ceq`) to agree with the C parser?

lchico (changes requested):

cborla:
## Evidence: unit tests (C)

All run against this branch's current code, from a full standalone `unit_tests/` build (production objects prebuilt, then `cmake -S unit_tests -B build_tests`).

| Suite | Result | Covers |
|---|---|---|
| `test_client_conf_ssl_resolution` (new) | 5/5 pass | `ClientConf()`'s UNSET-resolution step end to end against a real file: no `<ssl>` → `system`; CA present, no explicit mode → `certificate`; explicit `full`/`none`/`system` are kept as-is even without/with a CA |
| `test_agent_config` | 16/16 pass | 2 assertions updated for the new default (`AGENT_VERIFY_SYSTEM`/`"system"` instead of the old `AGENT_VERIFY_NONE`/`"none"`) |
| `test_client-config_https` | 86/86 pass | `Read_Agent_SSL()`'s own parsing (unaffected by this PR except one stale comment, fixed) |
| `https_client_unit_test` | 400/400 pass | Full suite, confirms no regression from the `moduleConfig.cpp` log-level change (INFO → WARNING for the disabled-verification message) |

`test_client_conf_ssl_resolution.c` closes a gap flagged during review: nothing previously exercised `ClientConf()`'s own UNSET→default resolution end-to-end (`Read_Agent_SSL()` alone can't see it, since it only sets fields when a tag is present, never resolves what "absent" should become).

cborla:
## Evidence: fresh-install CA routing (`register_configure_agent.sh` / `InstallerScripts.vbs`)

**Live test, real manager + real agent** (Ubuntu 24.04 manager with a self-signed, installer-generated certificate; agent built from this branch):

1. **Default install, no `<agent><ssl>` in the shipped config.** `ClientConf()` resolves `verification_mode` to `system`. The manager's certificate is self-signed and not in the OS trust store, so the agent correctly refuses to connect and logs a TLS verification failure instead of the pre-fix `TLS verification is DISABLED (verify_mode=none)` — the fix working as intended, not a bug.
2. **`WAZUH_REGISTRATION_CA` set to the manager's own CA at install time.** `register_configure_agent.sh` routes it into `<agent><ssl><certificate_authorities>` (previously it only ever reached the deprecated, ignored `<enrollment><server_ca_path>` tag). The agent enrolls and connects successfully; querying the live agent's own config (`queue/sockets/agent`) confirms `"ssl":{"verification_mode":"certificate", ...}` — inferred automatically since a CA was set without an explicit `<verification_mode>`.

**Since that live run**, `insert_into_agent_block()` (the awk helper `set_agent_ssl_ca()` uses to insert a fresh `<ssl>` block) was refactored to take the target tag as a parameter, so `set_agent_ssl_ca()` can reuse it for `<ssl>` instead of keeping a second copy of the same logic. Re-verified with a harness covering the three shapes `set_agent_ssl_ca()` can hit — a well-formed existing `<ssl>` block, a malformed one (opening tag not alone on its own line), and no `<ssl>` block at all — all three produce byte-identical output to before the refactor.

A live re-run against the final code (this refactor plus the `pin_ca()`/comment-stripping round) is still pending — same blocker as the fresh Linux WPK build (see the WPK-gate comments below); will follow up once that's unblocked.

cborla:
## Evidence: WPK upgrade gate, Linux/macOS (`pkg_installer.sh`)

Full consolidated regression against the final code, extracting the real `probe_build_target`/`strip_xml_comments`/`xml_value`/`xml_tag_present`/`xml_block_present`/`pin_ca`/gate functions verbatim into an isolated harness (stubbed `abort_upgrade`/`probe_server_verified`, real everything else), run with `bash -n`-clean syntax. 10/10 scenarios behave exactly as intended:

| # | Scenario | Result |
|---|---|---|
| 1 | No `<ssl>` block at all | Resolves default to `system`; no CA/probe available in the harness → aborts for that reason (not a false positive from the resolution logic) |
| 2 | `<verification_mode/>` (self-closing, present-but-empty) | Aborts: *"is present but empty"* — no longer silently substitutes the default |
| 3 | `<verification_mode>System</verification_mode>` (wrong case) | Aborts: *"is not a value this agent recognizes"* — the new default arm catches it instead of silently no-op'ing |
| 4 | `<ssl>` block fully commented out (`<!-- ... -->`) | Resolves default to `system` exactly as if absent — the comment is not read as live |
| 5 | Explicit `<verification_mode>none</verification_mode>` | No-op, no abort — explicit opt-out still respected |
| 6 | Explicit `system` **and** `<certificate_authorities>` also set | Aborts on the conflict guard (the combination `validateTls()` rejects at runtime) |
| 7 | Pure unmigrated 4.x shape (`<client>`, no `<agent>` at all) + a default CA present | `pin_ca()` correctly refuses to pin under `<client>` (inert — `Read_Legacy_Client_Address()` never reads `<ssl>`) and aborts with *"no `<agent>` block found, or an existing `<ssl>` block was not in the expected format"*, instead of reporting success with nothing the real parser will ever read |
| 8 | `system`, no CA, but the live trust probe succeeds | Proceeds normally: *"already verifies the manager's certificate"* |
| 9 | Normal `<agent>` shape + a default CA present, trust probe fails | `pin_ca()` succeeds, `<certificate_authorities>` is written under the real `<agent><ssl>` |

Case 7 is the regression test for the most recent fix (`pin_ca()`'s `<agent>`-less `<client>` rejection); case 9 is the control confirming the same code path still works normally for a real `<agent>` shape.

cborla:
## Evidence: WPK upgrade gate, Windows (`do_upgrade.ps1`) — real Windows 11

Same 10-scenario matrix as the Linux gate, but executed directly on a **real Windows 11 guest** (VirtualBox + WinRM, `gusztavvargadr/windows-11-24h2-enterprise` box, PowerShell 5.1, real CRLF `ossec.conf` fixtures on disk) — not a Linux emulation. `probe_server`/`probe_server_verified` were overridden after sourcing the real functions to force "manager reachable, certificate not verified by the OS store", so the gate's own probe/pin/abort logic runs for real instead of short-circuiting on the test-mode skip. 8/8 scenarios match the Linux results exactly:

| # | Scenario | Result |
|---|---|---|
| 1 | No `<ssl>` block | Resolves to `system`, aborts on no CA available (same as Linux case 1) |
| 2 | `<verification_mode/>` self-closing empty | Aborts: *"is present but empty"* |
| 3 | `<verification_mode>System</verification_mode>` (wrong case) | Aborts: *"is not a value this agent recognizes"* — confirms `-ceq` (case-sensitive) actually behaves case-sensitively in real PowerShell 5.1, not just in theory |
| 4 | `<ssl>` block commented out | Resolves to `system` as if absent |
| 5 | Explicit `none` | No abort |
| 6 | Explicit `system` + CA conflict | Aborts on the conflict guard |
| 7 | Pure `<client>`-only shape + default CA present | `pin_ca()` refuses to pin under `<client>`, config file left byte-identical, aborts with the same message as Linux |
| 9 | Normal `<agent>` shape + default CA present | `pin_ca()` succeeds, `<certificate_authorities>` written under `<agent><ssl>` |

Separately, `probe_server_verified()`'s exception-type logging (added so a DNS/timeout failure doesn't get logged as "certificate not trusted") was verified against a real connection-refused target: the log correctly reads *"Certificate trust check failed for a reason other than certificate trust (WebException: Unable to connect to the remote server); treating as not verified"* rather than the certificate-trust wording.

cborla:
## Evidence: real signed WPK — Windows

A genuine, signed WPK was generated for this branch's exact head commit through the real production pipeline (not a hand-built/forged package):

1. `5.X - Build Wazuh agent Windows (MSI)` (`wazuh/wazuh`, dispatched against this branch) → `wazuh-agent_5.0.0-0_windows_44c3b5b.msi`, uploaded to the internal packages bucket.
2. `5.X - Build Wazuh WPK` (`wazuh/wazuh-agent-packages`, `source_reference` = this branch, `wpk_reference` = that MSI) → signed with the real WPK private key/cert from AWS Secrets Manager → `wazuh-agent_5.0.0-0_windows_44c3b5b.msi.wpk` (6,556,059 bytes, upload size-verified against the S3 object).

That same workflow run includes its own built-in `Test upgrade WPK for version 5.0.0 - windows` job, which **passed** against the freshly-built WPK.

A real, manager-driven `agent_upgrade` end-to-end run using this WPK (Windows agent enrolled against a live 5.0.0 manager, upgrade dispatched from the manager) is the next step, pending scheduling.

cborla:
## Evidence: real 4.x → 5.0.0 migration — Linux

Before the WPK gate existed, the underlying scenario it was built to prevent was reproduced end to end with real, unmodified tooling (no custom/forged WPK):

1. Enrolled a real `wazuh-agent` 4.14.7 container via the legacy `agent-auth` mechanism against a real 5.0.0 manager; confirmed connected over TCP/1514.
2. Triggered a genuine in-place upgrade from the manager's own `agent_upgrade` tool, using the real, signed nightly 5.0.0 WPK — `dpkg` upgraded `4.14.7-1` → `5.0.0-latest` cleanly, preserving the legacy `<client><server>` block via the 5.x compatibility layer (no `<ssl>` block, since 4.x never had that concept).
3. With the **stock, unpatched** 5.0.0 binaries from that real migration, confirmed the original vulnerability reproduces exactly as described in the tracking issue — via a migration, not just a fresh install: `TLS verification is DISABLED (verify_mode=none)`.
4. Took that exact post-migration `ossec.conf`/`client.keys` state and ran it under this branch's **patched** binary instead: the agent correctly refuses to trust the self-signed manager certificate and pauses event production — `Manager unreachable. Pausing module event production. Reason: (60) SSL peer certificate or SSH remote key was not OK: SSL certificate OpenSSL verify result: self-signed certificate (18)`. Fails closed, as intended, rather than connecting insecurely.

This is exactly the migration gap the WPK gate (`pin_ca()`/the `system`-branch trust probe) was subsequently built to close — instead of the upgrade completing and the agent going dark, the gate now aborts the upgrade beforehand with a clear reason, keeping the host on its working 4.x version until the manager's CA is provisioned.

**Update**: rerun with a freshly-built, real signed WPK from this branch's current head — see the [real manager-driven upgrade comment](https://github.com/wazuh/wazuh/pull/38786#issuecomment-5525396695). (`5.X - Build Wazuh agent Linux (DEB/RPM)`'s `workflow_dispatch` path failed 7/7 times with `fatal: bad config file line 13 in .git/config` for this branch's actual name; diagnosed and worked around by building from a short-named branch pointing at the identical commit — the long branch name itself is the trigger, confirmed by an identical build succeeding immediately once shortened.)

cborla:
## Evidence: real manager-driven upgrade — Linux, both gate outcomes

End-to-end test using the real signed WPK from the comment above, dispatched from a real manager's own `agent_upgrade` tool (`agent_upgrade -a <id> -f wazuh-agent_5.0.0-0_amd64_4f8bc04.deb.wpk -F --package_type deb`) against two real enrolled agents, both starting on the stock nightly `5.0.0-latest` (pre-fix, `verify_mode=none`).

**Agent with the manager's certificate in its OS trust store** (`update-ca-certificates` after importing `remoted.pem`):

```
2026/09/03 11:52:07 - Checking connectivity to 127.0.0.1:1517/wazuh-manager.
2026/09/03 11:52:07 - Manager reachable at 127.0.0.1:1517/wazuh-manager.
2026/09/03 11:52:07 - The system trust store already verifies the manager's certificate; proceeding under verify_mode=system.
dpkg: warning: downgrading wazuh-agent from 5.0.0-latest to 5.0.0-0
...
2026/09/03 11:53:18 - Status = connected.
2026/09/03 11:53:18 - Connected to manager.
2026/09/03 11:53:18 - Upgrade finished successfully.
```

`dpkg -l` confirms `5.0.0-0` (this branch's build) installed; the agent reconnected and stayed connected after the restart.

**Agent with no CA trust configured at all** (default, untouched):

```
2026/09/03 11:52:22 - Checking connectivity to 127.0.0.1:1517/wazuh-manager.
2026/09/03 11:52:22 - Manager reachable at 127.0.0.1:1517/wazuh-manager.
2026/09/03 11:52:22 - Upgrade failed. The system trust store does not verify the manager's certificate at 127.0.0.1:1517, and no CA was found at ./etc/certs/root-ca.pem. Place the manager's CA there, or configure <certificate_authorities> explicitly, then retry the upgrade; staying on the current version, interrupting upgrade.
2026/09/03 11:52:22 - Reloading the Wazuh agent to report the upgrade result.
```

`dpkg -l` confirms the package **stayed at `5.0.0-latest`** — the upgrade never applied. The agent restarted only to report the failed result back to the manager, and stayed fully functional/connected on its prior version — no loss of connectivity, no partial upgrade, exactly the fail-closed behavior the gate exists for.

This closes the migration gap identified earlier in the PR (see the real-migration comment): an agent that previously would have silently gone dark after an in-place upgrade now either upgrades only once trust is actually established, or stays on its working version with a clear reason logged.

src/init/pkg_installer.sh
```diff
@@ -354,6 +493,100 @@ else
...
+
+# Default drop-in location for the manager's CA (mirrored on Windows in
+# do_upgrade.ps1): an operator can place it here ahead of an upgrade without having
+# to hand-edit ossec.conf. Resolves to /var/ossec/etc/certs/root-ca.pem on a default
+# install.
+DEFAULT_CA_FILE="./etc/certs/root-ca.pem"
+
+case "${SSL_VERIFICATION_MODE}" in
+    full|certificate)
+        if [ -z "${SSL_CA}" ] || [ ! -r "${SSL_CA}" ]; then
```
rjcausarano: What would happen if SSL_CA is a readable directory? If for example the user set the configuration like this: <certificate_authorities>/etc</certificate_authorities>. Should you add a check that discards directories?

src/init/register_configure_agent.sh
```diff
@@ -295,6 +334,74 @@ set_agent_option() {
...
+        ca_path_escaped="$(printf '%s' "${ca_path}" | sed -e 's/\\/\\\\/g' -e 's/&/\\\&/g')"
+        edit_value_tag "certificate_authorities" "${ca_path_escaped}"
+        return
+    fi
+
+    if agent_option_is_set "ssl"; then
+        # <ssl> already exists (e.g. hand-configured) but without a CA yet: insert
+        # the tag right after the opening <ssl>, reusing insert_into_agent_block's own
+        # awk technique (parameterized by tag) instead of a second copy of it.
+        echo "      <certificate_authorities>${ca_path}</certificate_authorities>" > "${TMP_SERVER}"
```
rjcausarano, line 514: ca_path is written as raw XML text here. A valid path containing & (for example, /etc/wazuh/ca&prod.pem) makes ossec.conf malformed and prevents the agent from
loading its configuration. 

Please XML-escape the value before inserting it (&, <, and > at minimum). The same escaping is needed in the other CA insertion/
replacement paths, including the Windows installer.

src/init/register_configure_agent.sh
```diff
@@ -295,6 +334,74 @@ set_agent_option() {
...
+            # who needs the CA can add <certificate_authorities> by hand afterward.
+            echo "$(date '+%Y/%m/%d %H:%M:%S') Could not pin WAZUH_REGISTRATION_CA into <ssl><certificate_authorities>: an existing <ssl> block was found but not in the expected format (opening tag not alone on its own line)." >> "${INSTALLDIR}/logs/ossec.log"
+        fi
+        rm -f "${TMP_SERVER}"
+        return
+    fi
+
+    {
+        echo "    <ssl>"
+        echo "      <certificate_authorities>${ca_path}</certificate_authorities>"
```
rjcausarano, line 532: Same as above

cborla:
## Test evidence: `SSL_VERIFICATION` install-time variable

New variable added after the initial round of review (requested by the dashboard team for the enrollment-wizard toggle, wazuh-dashboard-plugins#9103, and tracked for DevOps' install repos, wazuh-automation#3778): `SSL_VERIFICATION` (`full`/`certificate`/`system`/`none`) writes `<agent><ssl><verification_mode>` directly at install time, on Linux/macOS (`register_configure_agent.sh`) and Windows (`InstallerScripts.vbs` + new `SSL_VERIFICATION` MSI property in `wazuh-installer.wxs`).

### Linux/macOS — `register_configure_agent.sh`

9 scenarios run against extracted copies of the real functions (`set_agent_verification_mode`, `set_agent_ssl_ca`, `agent_option_is_set`, `agent_option_value`, `replace_agent_ssl_tag`), all passing:

1. No existing `<ssl>` block — fresh `<ssl><verification_mode>` block created.
2. `<ssl>` exists without `<verification_mode>` — inserted into it.
3. `<verification_mode>` already set (paired form) — rewritten.
4. `<verification_mode/>` self-closing — normalized and rewritten.
5. Invalid value (e.g. `System`, wrong case) — rejected, file untouched, logged.
6. Empty value — no-op, nothing logged.
7. `SSL_VERIFICATION=system` + `WAZUH_REGISTRATION_CA` together — existing conflict check fires, CA dropped, mode kept.
8. `SSL_VERIFICATION=certificate` + `WAZUH_REGISTRATION_CA` together — both land correctly in the same `<ssl>` block.
9. A same-named `<verification_mode>`/`<certificate_authorities>` tag in an unrelated block (e.g. `<localfile>`) elsewhere in the file — left untouched; only the `<agent><ssl>` occurrence is read or written.

Also re-verified the pre-existing `certificate_authorities` write path (`set_agent_ssl_ca`) doesn't regress under the new `<agent><ssl>` scoping: single-line, self-closing, multi-line pretty-printed, and single-/multi-line commented-out forms all still resolve correctly, with a same-named decoy tag outside `<agent><ssl>` left untouched in every case.

### Windows — `InstallerScripts.vbs` / `wazuh-installer.wxs`

Ran the actual VBScript engine (not just inspection): this machine has `wine` with a working `cscript.exe`. Extracted the real `SSL_VERIFICATION` routing block plus its dependencies (`install_log`, `StrippedOfComments`) verbatim from the file into a test harness and ran it under `wine cscript.exe`.

10 scenarios, all passing, mirroring the Linux/macOS coverage above plus two Windows-specific edge cases: self-closing tag with a space (`<verification_mode />`) and a tag inside an XML comment (correctly ignored, fresh tag written instead of touching the commented one). Also ran the combined `SSL_VERIFICATION` + `WAZUH_REGISTRATION_CA` block together (same conflict/co-existence scenarios as #7/#8 above), confirming the two routing blocks interact correctly when both variables are supplied.

One environment limitation: this wine build's VBScript engine has `Join()` completely unimplemented (`fixme:vbscript:Global_Join`, confirmed with a standalone trivial repro — any delimiter, not specific to this code). This affects the pre-existing, unmodified `certificate_authorities` rewrite loop too, not something introduced here. Worked around it in the test harness only (a manual-loop `MyJoin()`), never in the real file, which still uses the standard `Join()` real Windows Script Host supports.

The whole real `InstallerScripts.vbs` file also loads cleanly under `wine cscript.exe` with no syntax errors.

### `wazuh-installer.wxs`

New `<Property Id="SSL_VERIFICATION" Secure="yes"/>` declaration and the property added to `SetCustomActionDataValue`'s `CustomActionData` string (position 17, after `WAZUH_MANAGER_ENDPOINT`) — verified well-formed XML (`ElementTree.parse`), and that `InstallerScripts.vbs`'s `args(17)` matches this new position exactly.

lchico (approved):
LGTM! Great work on this, @cborla.

- [x]  Code changes reviewed
- [x]  Configuration changes documented

PR #38801, opened 2026-09-02
Parametrize /scan/vd timeouts, fix the metrics fan-out's missing worker documents, and remove two dead code paths

### LINKED ISSUE #38549: HTTPS/inventory-sync: small correctness defects found during 5.0.0 readiness testing

ignaciogalle12git:
## Description

Six mechanical changes from #38549, none of them needing a design decision: items 59, 58 and 50 from the timeout-layers follow-up, one index-template documentation correction confirmed with the indexer team, and two added after the first rebase. Rebased again onto `43ea975739`, which carries #38727: its only conflict was on `VD_SCAN_BUDGET_MS`, resolved below.

Item 58's behavioral fix landed in #38757 with the `monitord` retirement; what is left of it here is the dead code that retirement exposed.

## Proposed Changes

- **Parametrize `/scan/vd`'s timeouts.** `ScanVdHandlerImpl`'s read and write timeouts against VD's socket were `constexpr` 5s. Added `remoted.vd_scan_read_timeout` and `remoted.vd_scan_write_timeout` (default 5s), and the budget warning now names the real options instead of a hardcoded string. A `<=0` value falls back to the compiled-in default, so an unresolved config cannot turn every request into an immediate timeout. `VD_SCAN_BUDGET_MS` becomes `budgetMsFor(read, write)`, so the check covers the configured deadlines and still counts the offset query #38727 added to it; that PR's test now asserts on the function, and a new one pins that an instance uses the configured values. No behavior change at the shipped default.

- **Remove the unused `wdb_find_agent()` lookup.** #38757 deleted its only call site with the rest of `monitord`, leaving the function with no production caller in the tree. Removed the function, its declaration, its test wrapper and its tests. The other half of item 58, the disconnection line being invisible at the default log level, stays open on the issue.

- **Document `/stateless`'s engine dependency.** `remoted.downstream_response_timeout` must exceed the engine's real ingestion p99, or a batch the engine is merely slow to ingest is redelivered by the agent's retry with nothing able to recognize it as the same batch. Documentation only; the engine-side fix is tracked in #38549.

- **Document the `wazuh-agent-config`/`wazuh-agent-stats` `dynamic: true` mapping.** A field outside the per-module sub-schema is indexed like any other, with no schema check and nothing downstream validating it. Corrected the comments and the API reference that implied otherwise.

- **Collect metrics from every cluster node, and report the nodes that fail.** On a 2-node cluster the `wazuh-metrics-comms-v4` and `wazuh-metrics-normalization` streams carried master documents only, silently. Both fan-outs built their `DistributedAPI` without `nodes`, which defaults to an empty list (`dapi.py:108`) and seeds `common.cluster_nodes` from it (`dapi.py:545`); `get_nodes_info` then intersects the requested node with that empty set (`cluster.py:128`), returns "Node does not exist" (1730) in `failed_items`, and the request is never forwarded. Every other `distributed_master` caller passes `nodes=await get_system_nodes()`. Two silent paths are closed with it: `distribute_function()` returns its `WazuhException` instead of raising, and a rejected node arrives as `failed_items` on an otherwise valid result, so `getattr(result, "affected_items", [])` read `[]` off both.

- **Drop `msg_sent` from the agent's state file.** Its only writer was `send_msg()` in `sendmsg.c`, removed by `badf5b9c5b` with the agent's HTTPS transport, so `wazuh-agent.state` had been reporting `msg_sent='0'` on healthy agents since. Removed the counter, its enum value, its state-file line and the test that drove it. The field never left the local text file: the JSON report forwarded to `wazuh-agent-stats` carries `messages.count` and the `tasks.*` counters and never had a `msg_sent`. The documentation still describes it, tracked in wazuh/wazuh-documentation#10068.

### Results and Evidence

Validated on a 2-node cluster with a real indexer: master built from this branch, one worker container running the same build, nightly indexer and dashboard, agent containers of both majors.

Documents per `wazuh.cluster.node`, queried in the indexer:

```
wazuh-metrics-comms-v4        master=384   worker=367
wazuh-metrics-normalization   master=2898  worker=2507
```

`wazuh-metrics-agents` stays master-only, as expected: it reads a cluster-wide database and has no fan-out.

The reporting half proved itself while the worker was still broken, naming the node and the daemon instead of dropping it:

```
ERROR: [metrics_snapshot] Failed to collect comms stats from node '<worker>'
WazuhInternalError: Error 1017 - Some Wazuh daemons are not ready yet in node
  "<worker>" (wazuh-manager-modulesd->failed)
```

For `msg_sent`, an agent package built from this branch was installed over the nightly one in the same container, with the state file deleted so the new binary had to regenerate it:

```
before (nightly)   status='connected'  msg_count='121'  msg_sent='0'
after  (branch)    status='connected'  msg_count='34'
                   last_keepalive='2026-09-04 14:08:10'
```

The field and its comment are gone; `msg_buffer`, `dispatched`, `discarded_duplicate` and `failed` are unchanged, and `msg_count` and `last_keepalive` keep advancing, so the agent is connected and reporting.

Suites on the rebased head:

```
$ pytest framework                                 # 5_testunit_framework.yml
2034 passed, 6 skipped

$ cmake -DTARGET=agent .. && make && ctest
100% tests passed, 0 tests failed out of 116

$ ./remoted_module_utest
793 tests from 101 test suites ran. PASSED 793

$ ./test_wazuhdb_queries_op
PASSED 97 test(s)
```

Each half of the metrics fix was falsified independently: removing `nodes=system_nodes` fails only `test_dapi_receives_the_cluster_node_list`, removing the `failed_items` block fails only `test_failed_items_are_reported`.

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

Windows and macOS compilation rely on CI. The changes are config plumbing, a dead-code removal and a Python fan-out, so ASan, Coverity and Valgrind were not run beyond what CI covers.

### Artifacts Affected

`wazuh-agent.state` loses its `msg_sent` line. It has reported `0` unconditionally since `badf5b9c5b`.

### Configuration Changes

New internal options `remoted.vd_scan_read_timeout` and `remoted.vd_scan_write_timeout` (default `5`, range `1`-`300`). No default value changes.

### Tests Introduced

- `test_metrics_snapshot.py`: the node list is passed and resolved once per cycle, a returned `WazuhException` is reported, and a node rejected into `failed_items` is reported. The fan-out tests stub `get_system_nodes` through an autouse fixture on the module under test, since once the whole framework suite has imported `metrics_snapshot` the module-level `patch.dict` no longer applies.
- `test_agentd_state.c` lost the case that drove the removed `INCREMENT_MSG_SEND` value.
- `test_wazuhdb_queries_op.c` and the `wazuhdb_queries_op_wrappers` mock lost the four `wdb_find_agent` cases with the function.
- `configEndpoint_test.cpp` and `statsEndpoint_test.cpp` updated for the corrected comments, no assertion changes.
- `ScanVdHandlerImpl`'s existing coverage needed no changes: it is constructed with the pre-existing 3-argument form, which still compiles against the two new defaulted parameters.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/docs/ref/modules/inventory-sync-server/api-reference.md
+++ b/docs/ref/modules/inventory-sync-server/api-reference.md
@@ -61,11 +61,8 @@ Three details worth knowing:
   own report is dropped rather than indexed next to the authoritative `wazuh.agent.id`.
 - The report is stored as it arrives. The module renames no metric and reshapes no module body, so the
   field names in the index are the ones the agent emits, and a metric the agent adds needs no change
-  here.
-- It does need one in the index template. `wazuh-agent-stats` is mapped `dynamic: strict` with every
-  leaf declared, so a module or metric it does not declare makes the indexer reject the whole document
-  with `strict_dynamic_mapping_exception`. The write is fire-and-forget and the agent already has its
-  `200`, so that rejection is invisible from here — read the document back off the indexer to see it.
+  here, nor in the index template: `wazuh-agent-stats` is mapped `dynamic: true`, so a module or
+  metric it does not declare is indexed like any other field rather than rejected.
 
 `400` when the body is not a JSON object, when `modules` is missing, is not an object, is empty, or
 holds a module whose body is not an object. The empty case is a rejection on purpose: indexing a
--- a/docs/ref/modules/remoted/configuration.md
+++ b/docs/ref/modules/remoted/configuration.md
@@ -764,7 +764,9 @@ Seconds to wait for the downstream service's response after the write completes.
 - **Allowed values:** Integer from `1` to `300`
 - **Note:** This is the global default. An endpoint whose handler legitimately takes much longer can
   declare its own deadline instead of forcing this value up for every endpoint (which would delay
-  detection of a genuinely hung downstream on the fast ones).
+  detection of a genuinely hung downstream on the fast ones). `/stateless` is bound by this default:
+  it must stay above the engine's real p99 ingestion latency, or a batch the engine takes longer to
+  ingest is redelivered by the agent's retry, with nothing able to recognize it as the same batch.
 
 #### remoted.downstream_stateful_response_timeout
 
--- a/docs/ref/modules/remoted/configuration.md
+++ b/docs/ref/modules/remoted/configuration.md
@@ -1007,6 +1007,26 @@ Concurrent connections `remoted` keeps to `authd` for enrollment.
   [`remoted.enroll.authd.queue.depth`](metrics.md#agent-enrollment--remotedenroll) sits near
   its capacity at peak.
 
+#### remoted.vd_scan_read_timeout
+
+Seconds to wait for VD's answer to the inline `POST /scan/vd` admission relay.
+
+- **Default value:** `5`
+- **Allowed values:** Integer from `1` to `300`
+- **Note:** VD answers at admission into its bounded dispatch queue, not after running the scan,
+  so this is a local-socket round trip measured in milliseconds. A larger value does not make VD
+  queue the scan any sooner.
+
+#### remoted.vd_scan_write_timeout
+
+Seconds to wait for the write side of the same inline `POST /scan/vd` relay to VD.
+
+- **Default value:** `5`
+- **Allowed values:** Integer from `1` to `300`
+- **Note:** Same admission-only round trip as `remoted.vd_scan_read_timeout`. Both, plus the
+  fixed deadlines of the `/offset` query the scan gates on, make up the `/scan/vd` downstream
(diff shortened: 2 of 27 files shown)


ignaciogalle12git:
Code Linter fail is not related to this PR. It is  already fixed in `5.0.0` by `0ade596ce0` ("fix: remoted code style"), which corrects those failing lines. 
Fixed :partying_face:

TomasTurina (approved):

PR #38810, opened 2026-09-02
Treat an empty run_as authorization context as one that matched nothing

jepalfer:
## Description

`POST /security/user/authenticate/run_as` exists to *reduce* the calling service account's privileges to whatever the supplied authorization context maps to. Every context that matches no rule in `rules.yaml` correctly yields a token with no roles. An **empty** context `{}` did the opposite: it returned a token carrying the caller's **static** roles — `administrator` for `wazuh-wui` — while still being stamped `run_as: true`. The one input that means "I could not determine who this user is" was the only one that failed open.

The cause is two truthiness tests in `framework/wazuh/rbac/preprocessor.py`. `{}` is falsy in Python, so it took the "no context was given" branch in both:

- `get_roles` fell through to `run_user_role_link_roles()`, the account's static roles;
- `get_permissions` skipped the `allow_run_as` check entirely — so **a user with `allow_run_as = false` posting `{}` got a 200 and a token instead of the documented 403/6004**. That second consequence is not in the report; it is the same bug and is fixed here.

Scope, as the report itself notes: the endpoint needs the `wazuh-wui` credentials and that account is statically an administrator, so this is not privilege escalation for an outsider. What it is, is the containment mechanism failing open on its degenerate input — any integration, proxy or dashboard session that ends up sending an empty body silently got an administrator session instead of an inert one. After this change such a session presents as "logged in, no permissions", which is worth knowing when triaging support reports.

Of the three tasks on the issue, the first and third are covered here. The second — checking the dashboard and integrations for code paths that can emit an empty context — is **out of this repository**: nothing in `wazuh/wazuh` calls this endpoint except the test suites, so it needs a sweep of `wazuh/wazuh-dashboard` and belongs in its own issue there.

## Proposed Changes

Chosen semantics: **an empty context is a context, and is resolved against the rules like any other.** With the default ruleset, whose five rules are all `FIND` on a concrete key, `{}` matches nothing and yields a token with no roles, exactly like `nobody-at-all`. This needs no client change and gives a caller that sends `{}` an inert session rather than a hard failure it may not handle.

Two limits of that statement, so nobody reads more into it than it gives. A custom rule can match an empty context: `NOT`, an empty `AND` and an empty `MATCH` all evaluate true on `{}` (`check_logic_operation`/`match_item` in `auth_context.py`), so a ruleset carrying one of those grants its role to `{}`, as it already did to any other non-matching context. And "no roles" is only inert in `rbac_mode: white`; in black mode `_black_expansion` grants a resource when no policy names it. Neither is introduced here; what this change removes is the one path from `run_as` to the account's statically linked roles.

| File | Change |
| --- | --- |
| `framework/wazuh/rbac/preprocessor.py` | `get_roles` and `get_permissions` test `auth_context is not None` instead of truthiness. Only an *absent* context (the plain-login path) now falls back to the user-role link. |
| `api/api/middlewares.py` | `access_log` masks `password`/`key` only when the body is a mapping. |
| `api/api/spec/spec.yaml` | The endpoint description states that a context matching no rule yields a token with no roles, that an empty object is resolved like any other context and matches none of the default rules, and that the body must be a JSON object. The `200` response now also declares `text/plain`, which is what the endpoint returns with `raw=true` (see below). |
| `api/api/controllers/security_controller.py` | A comment on the body read, documenting the invariant the fix rests on: the request validator guarantees a JSON object, so `auth_context` is never `None` there — if it were, `remove_nones_to_dict` would drop it and `get_permissions` would fall back to the static roles. |

That is the whole fix. `framework/wazuh/core/exception.py` and `api/api/authentication.py` are untouched: the first was changed in an earlier revision of this PR and reverted (see below), and `authentication.py`'s `"run_as": auth_context is not None` already tested presence — it was simply the only one of the three that did. `security_controller.py` carries a comment and no logic change.

### The `access_log` fix, and why it is here

Live verification turned up a second defect. `access_log` did `if 'password' in body:` without checking that `body` is a mapping, so a `null`, numeric or boolean body raised `TypeError: argument of type 'bool' is not iterable`. That runs on **every** endpoint, after the response is built, so it turned the request validator's correct 400 into an unhandled 500 regardless of what the handler returned. Three of the bodies in the table below hit it.

### What was reverted after review, and why

The first revision of this PR also added `requestBody: required: true` to the spec and a body guard in `run_as_login` that raised a new error `6005` for a body that is not a JSON object. Review (@MiguelazoDS) showed both were wrong, and both are now gone. The reasoning, confirmed against the pinned `connexion==3.1.0`:

**`required: true` was a regression.** `connexion/validators/abstract.py:120-125` decides whether a body is present from the `Content-Length` header alone:

```python
headers = Headers(scope=scope)
if not int(headers.get("content-length", 0)):
    body = self._schema.get("default")
    if body is None and self._required:
        raise BadRequestProblem("RequestBody is required")
```

A `Transfer-Encoding: chunked` request declares no `Content-Length`, so a body that is right there read as absent and was rejected with `400 RequestBody is required`. Any client that streams the authorization context would have broken.

**And it bought nothing.** `connexion/validators/json.py:64-66` already raises `Request body must not be empty` for an empty or `null` body whenever the schema is not nullable, independent of `required`. The "no body at all" row of the evidence table below is `400` in the **Before** column — measured on unpatched 5.0.0, which never had `required: true` — which is that check firing.

**The `6005` guard was unreachable.** `connexion/middleware/request_validation.py` defaults the mime type to `consumes[0]` when no `Content-Type` is sent and raises `UnsupportedMediaTypeProblem` (415) for anything not in `consumes`, so for this operation the JSON body validator always runs: `_parse` rejects malformed JSON, `_validate` rejects non-objects and an empty body. The handler can only ever be handed a `dict`. The detail strings in the evidence table are the proof — `[] is not of type 'object'`, `Request body must not be empty`, `Expecting value: line 1 column 15 (char 14)` are all connexion's, never `6005`. The unit tests that appeared to cover the guard patched `request.json`, i.e. the exact layer that makes the guard unreachable, so they passed while proving nothing; they are removed with it.

The security fix never depended on either. It is the two `is not None` tests in `preprocessor.py`.

### The `run_as` + `raw=true` 500 in CI, and the spec fix

The first runs of this PR failed `test_security_POST_endpoints.tavern.yaml` on the new empty-context stage that used `raw: True`, with a 500 the API log did not explain (`api.log` only carries the access line; `error_handler.py` logs nothing for a handled `ProblemException`). The response body, obtained by re-running the file with `--log-level=DEBUG`, was:

```
title:  Internal Server Error
detail: Invalid Response Content-type (text/plain), expected ['application/json', ...]
```

That is connexion's response validation. The integration environment enables it — `api/test/integration/env/base/manager/entrypoint.sh` rewrites `validate_responses=False` to `True` in `wazuh_manager_apid.py` — and the `run_as` operation's `200` only declared `application/json`, while the handler returns `text/plain` for `raw=true` exactly like `POST /security/user/authenticate`, whose spec does declare both. So every `raw` `run_as` login failed in CI, empty context or not, and never in production, where responses are not validated. Reproduced in-process against the real `spec.yaml` with the full middleware stack: `validate_responses=True` → 500 on `raw` for `{}` and for a populated context alike; adding `text/plain` to the `run_as` `200` → 200 everywhere; `validate_responses=False` → 200 everywhere, before and after.

It was a pre-existing inaccuracy of the specification that this PR's stage was the first to exercise. The spec now declares `text/plain` for that response, and the `raw` empty-context stage is kept as the regression guard for it.

### Results and Evidence

Re-measured in full after the review revert, on a freshly installed single-node 5.0.0 manager in a container with an agent enrolled for the purpose (`001`, `v5.0.0`, active). The installed `preprocessor.py`, `middlewares.py` and `security_controller.py` were byte-compared against the branch before starting, so the numbers below are this commit's, not an overlay's. The **Before** column is the unpatched build from the first verification round; the **After** column is today's re-run.

`POST /security/user/authenticate/run_as` as `wazuh-wui`:

| Body | Before | After |
| --- | --- | --- |
| `{"user_name":"wazuh-admin"}` | 200 `roles=[8]` | 200 `roles=[8]` |
| `{"user_name":"wazuh-readonly"}` | 200 `roles=[2]` | 200 `roles=[2]` |
| `{"username":"nobody-at-all"}` | 200 `roles=[]` | 200 `roles=[]` |
| **`{}`** | **200 `roles=[1]` (administrator), `run_as=true`** | **200 `roles=[]`, `run_as=true`** |
| `null` | **500 Internal Server Error** | 400 |
| `[]` | 400 | 400 |
| `"str"` | 400 | 400 |
| `0` | **500 Internal Server Error** | 400 |
| `false` | **500 Internal Server Error** | 400 |
| empty string | 400 | 400 |
| no body at all | 400 | 400 |

The `null`, `0` and `false` rows are the `access_log` defect: the 400 was already being produced by the validator, and the crash happened afterwards while logging the response. Every 400 in this table comes from connexion, before the handler runs, both before and after — which is what makes the reverted guard dead code. The detail strings returned today are the proof, none of them ours:

| Body | Code | Detail |
| --- | --- | --- |
| `{}` | 200 | token issued, `run_as=true`, `rbac_roles=[]` |
| `null` | 400 | `Request body must not be empty` |
| `[]` | 400 | `[] is not of type 'object'` |
| `"str"` | 400 | `'str' is not of type 'object'` |
| `0` | 400 | `0 is not of type 'object'` |
| `false` | 400 | `False is not of type 'object'` |
| empty string | 400 | `Request body must not be empty` |
| no body at all | 400 | `Request body must not be empty` |
| `{"user_name": ` | 400 | `Expecting value: line 1 column 15 (char 14)` |

**The `allow_run_as` bypass.** A user with `allow_run_as = false` holding the `administrator` role, posting `{}`:

- before: `200`, an administrator token stamped `run_as: true`, which then read an agent's key (`GET /agents/001/key` → `200`);
- after: `403`, `error: 6004`.

The same user with a *non-empty* context was correctly `403` both before and after.

**The empty-context token is now inert (in `rbac_mode: white`, the default and the mode the integration environment forces through `enable_white_mode()`).** `GET /security/users?user_ids=1`, `/security/config`, `/agents/001/key` and `/agents?agents_list=001` were all `200` before and are all `403` after. Note that Wazuh RBAC *filters* rather than denies on unqualified list endpoints, so a check has to name a resource — `GET /security/users` with no parameters is a `200` with `affected_items: []` either way.

**The plain-login path is untouched.** `POST /security/user/authenticate` as `wazuh-wui` still returns `roles=[1] run_as=false` and that token still gets `200` on `/security/users`.

**Logging.** Non-object bodies now log verbatim instead of raising (`... "POST /security/user/authenticate/run_as" with parameters {} and body null done in 0.066s: 400`), and `password` masking is unchanged (`{"username": "masktest", "password": "****"}`). No `Exception in ASGI application` entries after the change.

**The chunked case from the review, isolated on one build.** `required: true` was toggled back on and off on the running manager, restarting the API between steps and changing nothing else:

```
A. as committed (no required: true)   chunked -> 200 token issued          Content-Length declared -> 200
B. with required: true put back       chunked -> 400 RequestBody required  Content-Length declared -> 200
C. reverted again                     chunked -> 200 token issued          Content-Length declared -> 200
```

A chunked **empty** context also returns `200` with `rbac_roles=[]`, i.e. the same inert token a declared-length `{}` gets. That single spec line was the whole cause, and the reviewer's command reproduces green on this commit.

**The `allow_run_as` bypass, re-measured.** A user created for the test (`norunas`, administrator role, `allow_run_as=false`): `{}` → `403 error 6004`, a non-empty context → `403 error 6004`, and its plain login still works (`200` on `/security/users?user_ids=1`).

**Smoke and health.** 19 endpoints across agents, security, cluster, overview and mitre all `200`; the five `/manager/*` paths I also probed are `404` because they are not in the 5.0.0 spec at all. The agent stayed `active` across both API restarts. Whole API log: zero `Exception in ASGI application`, zero responses logged as `500`, and `password` still masked (`body {"username": "masktest", "password": "****"}`).

#### The issue's proof of concept

Run verbatim, with one substitution: there is no `wazuh.dashboard` container in this environment, so the password comes from the default `wazuh-wui:wazuh-wui` instead of being read out of `opensearch_dashboards.yml`. The three calls are otherwise unchanged.

Before:

```
nobody-at-all → GET /security/users/me   roles: []
{}            → GET /security/users/me   roles: [{"id": 1, "name": "administrator",
                                            "policies": [{"name": "agents_all_resourceless", ...}]}]
{}            → GET /security/users      ['wazuh', 'wazuh-wui', ...]
{}            → GET /agents/001/key      {"key": "MDAxIDg0NmUxMjBlZWQyOSBhbnkgYjkwOTM4NWVl..."}
```

After:

```
nobody-at-all → GET /security/users/me   roles: []
{}            → GET /security/users/me   roles: []
{}            → GET /security/users      []
{}            → GET /agents/001/key      403 "Permission denied: Resource type: agent:id"
```

The PoC is built as "here is the correct behaviour (`nobody-at-all`), here is the broken one (`{}`)". Those two halves are now byte-identical, so it has nothing left to demonstrate.

The two write calls the issue lists alongside the PoC are refused as well, and the agent survived both:

```
DELETE /agents?agents_list=001&status=all  → 403 "Permission denied: Resource type: agent:id"
PUT    /agents/001/group/default           → 403 "Permission denied: Resource type: group:id"
```

#### The issue's expected results

> An empty authentication context is refused, or is treated exactly like an unmatched one and yields a token with no roles. The endpoint should have no input that returns more privilege than a valid, matched context would.

**First clause: met**, by the second option — `{}` is treated exactly like an unmatched context, byte-identical through `/security/users/me`, `/security/users` and `/agents/{id}/key`.

**Second clause: met** for every input exercised — the five default rules, an unmatched context, `{}`, `null`, `[]`, `"str"`, `0`, `false`, an empty string, an absent body, and nested JSON at depths 100/500/1100. That list is an enumeration and the clause is a universal, so what really backs it is structural: with `is not None` in place, any *present* context is resolved through `run_auth_context_roles()`, so the only privileges reachable from this endpoint are those a rule in `rules.yaml` grants. The path from `run_as` to the account's static roles no longer exists.

Every rule in the default ruleset, after the fix, for reference:

| Authorization context | Roles | `GET /agents/001/key` |
| --- | --- | --- |
| `{"username":"elastic"}` | `['administrator']` | 200 |
| `{"user_name":"admin"}` | `['administrator']` | 200 |
| `{"user_name":"wazuh-admin"}` | `['wazuh_indexer_admin']` | 200 |
| `{"user_name":"wazuh-readonly"}` | `['readonly']` | 200 |
| `{"user_name":"wazuh-demo"}` | `['readonly']` | 200 |
| `{"username":"nobody-at-all"}` | `[]` | 403 |
| `{}` | `[]` | 403 |

### A note on `raw` and this endpoint

An earlier revision of these stages fetched the empty-context token with `?raw=true`, which returned `500` in the clustered integration environment while returning `200` on a single node under every header combination tried. That combination is undocumented: `POST /security/user/authenticate/run_as` declares only `application/json` for its `200`, whereas `POST /security/user/authenticate` declares `application/json` **and** `text/plain`, even though both handlers return `text/plain` when `raw=true`. The stage was asserting behaviour the contract does not promise, so the token is now saved from the JSON response instead. The undeclared `text/plain` on `run_as`, and whatever raises behind it in a cluster, are pre-existing and untouched by this PR; they are worth their own issue.

This PR changes Python only — no compiled code, no decoder/rule assets and no engine code — so the compilation, memory-test, decoder/rule and engine checklists do not apply and have been omitted.

- Wazuh server API/Framework
  - [ ] Run API Integration Tests — **partially.** Every assertion of the new tavern stages was reproduced by hand against a live manager, and so was the subject of the new chunked-body test. The suites themselves — `test_rbac_white_*` / `test_rbac_black_*` and `tests/integration/test_api/test_miscs/test_run_as_chunked_body.py` — need the docker-compose environment and the `wazuh_testing` framework respectively, and have **not** been executed; that is for CI.

Unit suites, run in the manager's own bundled Python 3.12 with its pinned dependencies:

```
framework/wazuh/rbac/tests        249 passed
api/api/controllers/test          131 passed
api/api/test                      313 passed
```

Re-run after the review revert (host, `PYTHONPATH=framework:api`): `test_security_controller.py` 51 passed, `test_preprocessor.py` 18 passed, `test_middlewares.py` 50 passed / 10 failed — those 10 (`test_secure_headers_middleware` and the nine `test_check_block_ip_middleware` cases) fail identically on `origin/5.0.0` in the same environment, and all 31 `access_log` tests pass.

Every new unit test that can fail against the unfixed code was confirmed to: 3 in `test_preprocessor.py` and 5 in `test_middlewares.py` (the `[]` case passes either way, as `'password' in []` is simply false). The new `test_security_controller.py` case is not a regression test — `security_controller.py`'s logic is unchanged, so it passes on both — it pins that `remove_nones_to_dict` does not strip an empty context. Of the tavern stages, 4 of the 7 groups fail without the fix (the other 3 are regression guards for bodies that already returned 400), and the `raw` stage fails without the `text/plain` spec change.

`tests/integration/test_api/test_miscs/test_recursion.py`, which posts deeply nested JSON to this endpoint, was reproduced by hand: depths 100 and 500 → `200`, depth 1100 → `400`, byte-identical before and after.

A 22-endpoint smoke test across agents, security, cluster, overview and mitre returned the expected status on every path, and the enrolled agent stayed `active` throughout.

### Artifacts Affected

Python only — the bundled `framework/` and `api/` packages inside the manager, and the API OpenAPI specification. No binaries, no packaging, no default configuration files.

### Configuration Changes

None in any configuration file, and no new error codes.

One client-visible behaviour change: an empty authorization context returns a token with no roles instead of the caller's static roles. That is the fix. Everything else about the endpoint's contract — which bodies are accepted, which are rejected, and with what status — is unchanged from 5.0.0.

### Tests Introduced

- `framework/wazuh/rbac/tests/test_preprocessor.py` — first coverage for `get_roles` and `get_permissions` (the file previously exercised only `PreProcessor.process_policy`): context-vs-user-link branch selection for `{}`, a populated context and `None`; `6004` for a user without `allow_run_as`; an empty context resolving to no roles; the plain-login path still returning static roles.
- `api/api/controllers/test/test_security_controller.py` — `test_run_as_login_auth_context`: `run_as_login` forwards an object context, `{}` included, without `remove_nones_to_dict` stripping it.
- `api/api/test/test_middlewares.py` — `test_access_log_non_object_body`, covering the six non-mapping body shapes.
- `api/test/integration/test_security_POST_endpoints.tavern.yaml` — eight stages next to the existing `run_as` block: empty context yields no roles; the same with `raw=true`, which guards the `text/plain` declaration in the spec under the environment's response validation; the resulting token is denied on a named resource; an empty context from a user without `allow_run_as` is 403/6004; and `null`, a list, a string and an absent body are each 400.
- `tests/integration/test_api/test_miscs/test_run_as_chunked_body.py` — new, and the direct regression test for the reverted `required: true`. A tavern stage cannot reach this case: tavern always builds the body from a fixed string, so `requests` sets a `Content-Length` and the chunked path is unreachable. Passing a generator as `data` makes `requests` stream the body with `Transfer-Encoding: chunked` and no `Content-Length`, which is exactly the shape `abstract.py:122` keys off. Three cases: a chunked matching context and a chunked `{}` must both be `200` with a token, and a genuinely absent body must still be `400` — the control that keeps the first two honest.

## Review Checklist

- [ ] Code changes reviewed
- [x] Chunked authorization context confirmed `200` on a live manager — see the A/B/C toggle above
- [ ] Full API integration suite (`test_rbac_white_*` / `test_rbac_black_*`) run in the docker-compose environment


MiguelazoDS (commented):
Some cases I was testing 

1- It seems a regression, adding the requestBody as true, rejects a body sent without Content-length 

This implementation
```console
root@7e35465be2c1:/# curl -sk -u wazuh-wui:wazuh-wui -H 'Content-Type: application/json'      -X POST 'https://127.0.0.1:55000/security/user/authenticate/run_as'      -w '\nHTTP %{http_code}\n' -d '{"user_name":"wazuh-admin"}'
{"data": {"token": "eyJhbGciOiJFUzUxMiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJ3YXp1aCIsImF1ZCI6IldhenVoIEFQSSBSRVNUIiwibmJmIjoxNzg4Mzc1NDA0LCJuYmZfbXMiOjE3ODgzNzU0MDQ0MTEsImV4cCI6MTc4ODM3NjMwNCwic3ViIjoid2F6dWgtd3VpIiwicnVuX2FzIjp0cnVlLCJyYmFjX3JvbGVzIjpbOF0sInJiYWNfbW9kZSI6IndoaXRlIiwiaGFzaF9hdXRoX2NvbnRleHQiOiI1YWFlMDE5Njk5OTI0MjRlYjk5YmIxYTkwNzIyZDVhNyJ9.ATVDcpgeLxVTErropyetfCjQip9u8p7R53BIxBPh1KGHS0B7xT-5K7VxYrAEL9plbNP5Q3YiufzDnNTeEwJGkSNSAexz6KJ0A5rTfkgBiekYGXkjpDjIlpKlwyTw7xe-BcFbdobn30mfKE1lJqDamWrS1zAgNUD9pU_krrk9x6qLklqw"}, "error": 0}
HTTP 200
root@7e35465be2c1:/# curl -sk -u wazuh-wui:wazuh-wui -H 'Content-Type: application/json' \
     -H 'Transfer-Encoding: chunked' --http1.1 \
     -X POST 'https://127.0.0.1:55000/security/user/authenticate/run_as' \
     -w '\nHTTP %{http_code}\n' -T - <<< '{"user_name":"wazuh-admin"}'
{"title": "Bad Request", "detail": "RequestBody is required"}
HTTP 400
```

5.0.0
```console
root@282b5effa733:/# curl -sk -u wazuh-wui:wazuh-wui -H 'Content-Type: application/json'      -X POST 'https://127.0.0.1:55000/secu
rity/user/authenticate/run_as'      -w '\nHTTP %{http_code}\n' -d '{"user_name":"wazuh-admin"}'
{"data": {"token": "eyJhbGciOiJFUzUxMiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJ3YXp1aCIsImF1ZCI6IldhenVoIEFQSSBSRVNUIiwibmJmIjoxNzg4Mzc1NDE0LCJuYmZfbXMiOjE3ODgzNzU0MTQ2NTIsImV4cCI6MTc4ODM3NjMxNCwic3ViIjoid2F6dWgtd3VpIiwicnVuX2FzIjp0cnVlLCJyYmFjX3JvbGVzIjpbOF0sInJiYWNfbW9kZSI6IndoaXRlIiwiaGFzaF9hdXRoX2NvbnRleHQiOiI1YWFlMDE5Njk5OTI0MjRlYjk5YmIxYTkwNzIyZDVhNyJ9.AEg8S4As29ZvltlliXpR-vF-k4VxzbnMkoS3BRIokdFyWVv8A0K3pwfWipsf2CHPMBNsF1mK1hai3ftLVM2f29nDACx9Qtjlxa7erGFmY_wudc8mLrVIehwfBT6e1hwR7JbWTbr5e42Cl5xkw3AGiGK28x0tLyPl2n5uu6vqsBsU1E9y"}, "error": 0}
HTTP 200
root@282b5effa733:/# curl -sk -u wazuh-wui:wazuh-wui -H 'Content-Type: application/json' \
     -H 'Transfer-Encoding: chunked' --http1.1 \
     -X POST 'https://127.0.0.1:55000/security/user/authenticate/run_as' \
     -w '\nHTTP %{http_code}\n' -T - <<< '{"user_name":"wazuh-admin"}'
{"data": {"token": "eyJhbGciOiJFUzUxMiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJ3YXp1aCIsImF1ZCI6IldhenVoIEFQSSBSRVNUIiwibmJmIjoxNzg4Mzc1NDQ4LCJuYmZfbXMiOjE3ODgzNzU0NDg4MjcsImV4cCI6MTc4ODM3NjM0OCwic3ViIjoid2F6dWgtd3VpIiwicnVuX2FzIjp0cnVlLCJyYmFjX3JvbGVzIjpbOF0sInJiYWNfbW9kZSI6IndoaXRlIiwiaGFzaF9hdXRoX2NvbnRleHQiOiI1YWFlMDE5Njk5OTI0MjRlYjk5YmIxYTkwNzIyZDVhNyJ9.ATl8IwG3cN1aaf-sbw16SdZazDiAtbQVsDXdkyeGB2_nD0JhyN6XyClf9x4srMT72fZtev0xosk3-YTUQ_-1u4F-AWS5LQ4Nk7frJ-Oce0W7RiLWC-JbZVe64CuECIV8iQO-9qb21wFhTRKrPlacnGm6NKATB-zvSIMRE2HlWA4R7tP1"}, "error": 0}
HTTP 200
```

2- The new code introduced seems unreachable 

```python
try:
    auth_context = await request.json()
except JSONDecodeError:
    raise_if_exc(WazuhError(6005))

if not isinstance(auth_context, dict):
    raise_if_exc(WazuhError(6005))
```

```console
root@7e35465be2c1:/# for body in '[]' '"str"' '0' 'false' 'null' '{"user_name": '; do
  echo "body=$body"
  curl -sk -u wazuh-wui:wazuh-wui -H 'Content-Type: application/json' \
       -X POST 'https://127.0.0.1:55000/security/user/authenticate/run_as' \
       -w '\nHTTP %{http_code}\n' -d "$body"
  echo
done
body=[]
{"title": "Bad Request", "detail": "[] is not of type 'object'"}
HTTP 400

body="str"
{"title": "Bad Request", "detail": "'str' is not of type 'object'"}
HTTP 400

body=0
{"title": "Bad Request", "detail": "0 is not of type 'object'"}
HTTP 400

body=false
{"title": "Bad Request", "detail": "False is not of type 'object'"}
HTTP 400

body=null
{"title": "Bad Request", "detail": "Request body must not be empty"}
HTTP 400

body={"user_name": 
{"title": "Bad Request", "detail": "Expecting value: line 1 column 15 (char 14)"}
HTTP 400
```

> [!NOTE]
> For this case it seems that the real case is not reachable despite the tests implemented, because those test bypass real behavior with mocks.

jotacarma90 (approved):

Issue #38766, opened 2026-09-01
wazuh-agentd segfaults in libagent_metadata.so on every agent service stop


Rebits:
## Description

`wazuh-agentd` dies with `SIGSEGV` instead of exiting cleanly whenever the agent service is stopped on x86_64 Linux. The crash is at a constant instruction, file offset `0x173b` in `libagent_metadata.so`, and reproduces on every run of the agent performance suite since 2026-08-18, on Ubuntu 24 and Amazon Linux 2023 alike. The arm64 agents of the same two distributions, running the same suite in the same artifacts, never show it.

The daemon is already on its way out when it faults: all five agent daemons stop together and come back about five seconds later, so the crash does not take a running agent down and there is no restart loop. What it does mean is that agentd's shutdown path is truncated at an unknown point on every stop, restart, configuration change, package upgrade, WPK upgrade and machine shutdown.

Nothing is written to `ossec.log`, which is why this went unnoticed for eleven days: it is only visible in the kernel ring buffer.

This is not wazuh/wazuh#38190. That issue fixed a **write** through a read-only mapping in `SharedMemoryProvider::update()` (`metadata_provider.cpp:74`), reached from `w_agentd_populate_metadata()` (`start_agent.c:212`) during agentd **start**, and was closed on 2026-08-04. The fault reported here is a **read** (`error 4`), during **stop**, and first appears two weeks after that fix. Same library and, presumably, the same shared metadata segment, so it is likely the same area of code.

## Environment

| Field | Value |
|---|---|
| Wazuh version | 5.0.0, stages `rc1` and `beta5` (both affected) |
| Packages | `wazuh-agent_5.0.0-latest_amd64.deb`, `wazuh-agent-5.0.0-latest.x86_64.rpm`, from `development/wazuh/5.x/main` |
| Affected platforms | `ubuntu-24-amd64`, `amazon-2023-amd64` |
| Not affected | `ubuntu-24-arm64`, `amazon-2023-arm64` |
| Jenkins job | `5.0.0/Performance_Testing`, production and staging |
| Latest occurrence | Build [#34](https://jenkins-production.qa.wazuh.info/job/5.0.0/job/Performance_Testing/34/), 2026-09-01 |

## Evidence

Build #34, both x86_64 hosts:

```
[...]

PR #38828, opened 2026-09-02
Stop the metadata shm fd from leaking into SCA's spawned commands

### LINKED ISSUE #38813: SCA scan raises a SELinux AVC finding on the agent's own metadata file
|---|---|---|---|---|
| 5.0.0 | wazuh-agent / agent_metadata + wazuh-modulesd (SCA) | Agent | Packages | RHEL 9.5 (any SELinux-enforcing distro) |

## Summary

A stock agent on an SELinux-enforcing host raises a **SELinux AVC denial finding on its own runtime metadata file** during the first SCA scan. The finding names `passwd -S root` and `/var/ossec/var/run/.wazuh_agent_metadata`, which reads like a password tool touching Wazuh internals. It is neither.

The agent opens its metadata SHM file **without `O_CLOEXEC`** and keeps the descriptor for the process lifetime. SCA spawns CIS check commands with `fork()` + `execvp()` and does not close inherited descriptors, so `passwd` starts with that Wazuh descriptor still open. SELinux transitions `passwd` into `passwd_t`, revalidates inherited FDs, and denies `passwd_t` read/write on a `var_t` file. `passwd` still exits 0 and the CIS check still **Passes** — the only effect is that a default deploy alerts on the product itself.

The CIS check is correct. The decoder and rule are correct — there was a real AVC. **Wazuh caused it.**

### 1. Required — open the metadata SHM with `O_CLOEXEC`

`src/shared_modules/agent_metadata/src/metadata_provider.cpp` (lines 288 and 293)

```diff
-                m_shm_fd = open(SHM_PATH, O_RDWR | O_CREAT, 0644);
+                m_shm_fd = open(SHM_PATH, O_RDWR | O_CREAT | O_CLOEXEC, 0644);

                 if (m_shm_fd == -1 && errno == EACCES)
                 {
                     // Permission denied for write, try read-only
-                    m_shm_fd = open(SHM_PATH, O_RDONLY);
+                    m_shm_fd = open(SHM_PATH, O_RDONLY | O_CLOEXEC);
                     m_read_only = true;
                 }
```

### 2. Required — close the descriptor right after `mmap`

`m_shm_fd` is **never used after the `mmap` call** — it is only closed in the destructor (`:361-363`). `update()` and `reset()` operate on the mapping (`m_shm`) alone. The mapping survives closing the descriptor, so the agent can drop it immediately and nothing downstream can leak it, regardless of how children are spawned:

```cpp
// after the mmap succeeds and the `created` initialization block
close(m_shm_fd);
m_shm_fd = -1;
```

Do both 1 and 2: `O_CLOEXEC` closes the race window between `open` and `mmap`, and dropping the descriptor removes the leak entirely.

<details>
<summary>Why the descriptor is provably unused after <code>mmap</code></summary>

Every use of `m_shm_fd` in the file: initialized `-1` (`:243`), the two `open` calls (`:288`, `:293`), `fstat` for the `created` check (`:304`), `ftruncate` for sizing (`:307`), error-path closes (`:309`, `:324`), passed to `mmap` (`:319`), destructor close (`:361-363`). Nothing after the constructor touches it.

The provider is a function-local-static singleton (`instance()`, `:58-61`), so today that descriptor stays open for the life of every process that links the library. On Windows the path is `CreateFileMappingA(INVALID_HANDLE_VALUE, NULL, ...)` — no backing file and NULL security attributes, so the handle is not inheritable. **This fix is Unix-only.**
</details>

### 3. Should — stop SCA children inheriting daemon descriptors

`src/wazuh_modules/src/wm_exec.c`, the Unix `wm_exec()` (`:321`)

The child redirects stdin/stdout/stderr and closes the capture-pipe ends (`:400-420`), then `execvp` (`:424`). Every other descriptor without close-on-exec is inherited. A CIS status command should not see agent sockets, SQLite files or SHM backing files, so close everything above `stderr` in the child after the `dup2` block and before `execvp`.

> ⚠️ **The originating report says SCA uses `posix_spawn` and suggests `POSIX_SPAWN_CLOEXEC_DEFAULT`. That does not apply here.** `posix_spawn` appears nowhere in the tree; `wm_exec` is `fork()` + `execvp()`, and `POSIX_SPAWN_CLOEXEC_DEFAULT` is a macOS-only extension. On Linux use an explicit close loop (`closefrom()`, `close_range()`, or iterate `/proc/self/fd`).

The precedent is in the same function: `wm_exec.c:343` already calls `w_descriptor_cloexec(pipe_fd[0])` (`src/shared/src/file_op.c:3277`).

### 4. Explicitly out of scope — do not do these

- Do **not** change CIS check 28147 (`ruleset/sca/rhel/9/cis_rhel9_linux.yml:6858`, command at `:6905`). It is correct and it Passes.
- Do **not** `allow` or `dontaudit` `passwd_t` on that file in `src/selinux/wazuh.te`. `passwd_t` must not inherit that descriptor in the first place.
- Do **not** suppress the rule. The AVC was real.
- Adding a file-context for the metadata path is completeness only, not this fix (note `src/selinux/` ships only `wazuh.te`, with no `.fc` file).

## How to test

1. RHEL 9, SELinux **enforcing**, `auditd` running, stock `audit.log` localfile.
2. Install the agent with default `ossec.conf` (SCA `enabled=yes`, `scan_on_start=yes`). No custom policy, no extra audit rules, no interactive `passwd`.
3. Enroll and wait for the first SCA scan (`Scan started` … `Scan ended` in `ossec.log`).

## How to verify it is fixed

On the same host, after the first scan of a patched agent:

| # | Check | Command | Expected after fix |
|---|---|---|---|
| 1 | No AVC names the metadata path | `ausearch -m avc -ts recent \| grep wazuh_agent_metadata` | no results |
| 2 | Descriptor is gone or close-on-exec | `for p in $(pgrep wazuh-); do grep -l wazuh_agent_metadata /proc/$p/fd/* 2>/dev/null; done` then `grep flags /proc/<pid>/fdinfo/<fd>` | descriptor absent, or `flags` has `02000000` (`O_CLOEXEC`) set |
| 3 | The CIS check still works | query `sca.db` for check `28147` | `result = Passed` |
| 4 | Re-running the scan does not recreate it | restart the agent / force a new scan, repeat #1 | still no AVC, no new finding |
| 5 | No finding in the dashboard | search findings for this agent: rule `SELinux access vector cache denial` with `file.path: /var/ossec/var/run/.wazuh_agent_metadata` | 0 documents |
| 6 | No other Wazuh path leaked | `ausearch -m avc -ts recent \| grep -i ossec` | no results |

Before the fix, #1 returns the AVC, #2 shows `O_CLOEXEC=False`, and #5 returns the finding — while #3 already returns `Passed`. Check #6 matters because `passwd` is only the case that happened to trip: the RHEL 9 policy alone runs **131** `c:` commands, and any other confined helper (`useradd`, `gpasswd`, `sshd`, …) rechecking a `var_t` file would produce the same class of AVC.

<details>
<summary>Root cause chain, with code references</summary>

**The descriptor is opened without close-on-exec** — `src/shared_modules/agent_metadata/src/metadata_provider.cpp:288`:

```cpp
#define SHM_PATH "var/run/.wazuh_agent_metadata"

m_shm_fd = open(SHM_PATH, O_RDWR | O_CREAT, 0644);   // no O_CLOEXEC

if (m_shm_fd == -1 && errno == EACCES)
{
    m_shm_fd = open(SHM_PATH, O_RDONLY);             // no O_CLOEXEC either
    m_read_only = true;
}
```

There is no later `fcntl(..., FD_CLOEXEC)` in the file. Matches the live host: `O_RDWR`, `O_CLOEXEC=False`.

**`wazuh-modulesd` is the process that holds it**, which is also the process that execs SCA commands:

- `src/wazuh_modules/agent_info/agent_info_impl/src/agent_info_impl.cpp:701` — `metadata_provider_update()`
- `src/shared_modules/sync_protocol/src/agent_sync_protocol.cpp:80`, `:741` — `metadata_provider_get()`
- `src/wazuh_modules/sca/sca_impl/tests/CMakeLists.txt:46` — `sca_impl` links `agent_metadata`

`agentd` (`src/client-agent/src/start_agent.c:280`) and syscheck hold it too, but SCA execs from modulesd, so modulesd's descriptor is the one that leaked.

**SCA reaches `wm_exec` through a function pointer**, so the chain is easy to miss:

`src/wazuh_modules/src/wm_sca.c:545` → `sca_set_wm_exec_ptr(wm_exec)` → `src/wazuh_modules/sca/src/sca.cpp:135` → `src/wazuh_modules/sca/sca_impl/src/sca_policy_check.cpp:194-210` invokes the callback per `c:` rule.

**`wm_exec`'s child cleans up only stdio** (`src/wazuh_modules/src/wm_exec.c:400-424`): `dup2` on stdin/stdout/stderr, close the pipe ends, `setsid()`, `execvp()`. No other descriptor is closed.
</details>

<details>
<summary>What the operator sees, and evidence from the lab host</summary>

One finding on agent `rhel-9` (id 004), ~3 minutes after RPM install:

```
wazuh.rule.title:   SELinux access vector cache denial   (level: low)
wazuh.integration.name: auditd
process.title:      passwd -S root
process.executable: /usr/bin/passwd
process.pid:        8543        process.parent.pid: 8012   (wazuh-modulesd)
process.exit_code:  0
file.path:          /var/ossec/var/run/.wazuh_agent_metadata
file.inode:         17294784
event.outcome:      failure     event.code: AVC
```

Three details are easy to misread together: `passwd … root` is `passwd -S root`, a *status query* that changes nothing; `event.outcome: failure` is the decoder mapping `AVC denied`, not a failed command (exit 0); and the Wazuh path was **inherited**, not opened by `passwd`.

The AVC — the only one in the host's `audit.log`:

```
type=AVC msg=audit(1788314682.130:1440): avc: denied { read write } for pid=8543 comm="passwd"
path="/var/ossec/var/run/.wazuh_agent_metadata" dev="dm-0" ino=17294784
scontext=system_u:system_r:passwd_t:s0 tcontext=system_u:object_r:var_t:s0 tclass=file permissive=0
type=SYSCALL msg=audit(1788314682.130:1440): SYSCALL=execve success=yes exit=0 ppid=8012 pid=8543
  uid=0 gid=987 tty=(none) comm="passwd" exe="/usr/bin/passwd" UID="root" GID="wazuh"
type=EXECVE msg=audit(1788314682.130:1440): argc=3 a0="passwd" a1="-S" a2="root"
```

`success=yes` / `exit=0`: the program ran fine. `permissive=0`: SELinux enforced the inherited-FD check and logged it. `passwd` then queried shadow as designed.

Live descriptor on the same host, same inode as the finding:

```
pid 8012 wazuh-modulesd (root:wazuh)
fd 23 -> /var/ossec/var/run/.wazuh_agent_metadata
inode 17294784    O_RDWR = True    O_CLOEXEC = False   <-- the bug
```

Timeline (PIDs not recycled; the finding's parent is the live modulesd):

```
02:03:57  rpm install wazuh-agent 5.0.0-1 (beta5)
02:04:29  metadata file created (inode 17294784)
02:04:35  modulesd started (pid 8012)
02:04:39  SCA scan started
02:04:42  execve passwd -S root, parent 8012, AVC
02:04:42  SCA scan ended            02:04:49  finding indexed
```

Lab: manager/indexer/dashboard 5.0.0-1 beta5 AIO on Ubuntu 24.04.2; agent `wazuh-agent-5.0.0-1.aarch64` beta5 on RHEL 9.5 (Plow) aarch64, agent id 004; SELinux enforcing/targeted with `wazuh.pp` loaded; wazuh uid/gid 987.
</details>

<details>
<summary>Blast radius: wider than RHEL 9</summary>

`passwd -S` is not unique to the RHEL 9 policy. Shipped policies that run it:

| Policy | Line |
|---|---|
| `ruleset/sca/rhel/9/cis_rhel9_linux.yml` | 6905 |
| `ruleset/sca/ol/9/cis_oracle_linux_9.yml` | 7920 |
| `ruleset/sca/ol/10/cis_oracle_linux_10.yml` | 7920 |
| `ruleset/sca/amazon/cis_amazon_linux_2023.yml` | 8030 |
| `ruleset/sca/ubuntu/cis_ubuntu22-04.yml`, `cis_ubuntu26-04.yml` | 7251, 9202 |
| `ruleset/sca/debian/cis_debian12.yml` | 7106 |
| `ruleset/sca/nginx/cis_nginx_1.yml`, `applications/cis_apache_24.yml` | 251, 349 |

Every SELinux-enforcing RPM distro (RHEL, Oracle Linux, Amazon Linux, Fedora family) is exposed with default SCA. Debian/Ubuntu ship AppArmor rather than SELinux, so they do not produce the AVC — the descriptor still leaks there, it just is not reported.

The RHEL 9 policy runs **131** `c:` commands per scan, so the exposure is one leaked descriptor across 131 spawns, not a single command. Fixing item 1+2 removes this finding; item 3 closes the class.
</details>

<details>
<summary>Ruled out</summary>

| Hypothesis | Why it is ruled out |
|---|---|
| Someone changed the root password | `argv` is `passwd -S root`; check 28147 is the only shipped RHEL 9 command producing it; result `Passed` |
| `passwd` opens Wazuh files on purpose | the path is an inherited descriptor; `passwd_t` has no need of agent metadata |
| Broken SCA check | `result: Passed`, exit 0, `execve success=yes` |
| SELinux blocking the agent | agent connected, SCA synced, FIM ran; the denial is the leftover FD, not Wazuh's own domain |
| Missing `wazuh.pp` | module loaded; the incomplete file-context is real but would not stop this AVC — relabeling the path away from `var_t` still leaves `passwd_t` inheriting a descriptor it must not have |
| Decoder / rule false positive | the AVC is real and correctly decoded; the cause is the leak |
</details>

hernanvalenzuela:
## Description
A Wazuh agent on an SELinux-enforcing host (e.g. RHEL 9.5) triggers a false-positive SELinux AVC denial during SCA scans, referencing the agent's own metadata shared-memory file (`.wazuh_agent_metadata`).

Root cause: `SharedMemoryProvider`'s constructor (`src/shared_modules/agent_metadata/src/metadata_provider.cpp:295`) opens that file without `O_CLOEXEC`, and the fd was never closed afterward — it stays open for the entire process lifetime. SCA's command rules (`c:` rules, e.g. `rpm -q <package>`) run through the legacy `wm_exec()` (`src/wazuh_modules/src/wm_exec.c:348,424`), a plain `fork()`+`execvp()` with no logic to close unrelated inherited fds. The spawned child (e.g. `rpm`) therefore inherits the metadata fd; when SELinux transitions that child into a restricted domain, the inherited fd is exactly what an enforcing policy denies — even though the child never touches the file and exits successfully.

## Proposed Changes
- `src/shared_modules/agent_metadata/src/metadata_provider.cpp`: both `open()` calls (the read-write and the read-only fallback) now request `O_CLOEXEC`, so the fd never survives into a child process in the first place.
- Additionally, the fd is now `close()`d immediately once `mmap()` succeeds — the mapping stands on its own afterward, so nothing needs the fd to keep working, and holding it open served no purpose beyond leaking it into every future child process for the rest of this object's lifetime (this class has no destructor since #38766 made it `static_assert`-enforced trivially destructible, so nothing would ever close it otherwise).
- Verified `m_shm_fd` isn't referenced anywhere else in the class after this point, so closing it immediately is safe.

### Results and Evidence
SELinux itself isn't available in the test environment (kernel support compiled in but disabled, no policy loaded), so the AVC itself couldn't be produced directly. Verified the exact mechanical precondition instead, at three levels:

**1. Mechanism-level, standalone.** Compiled the literal `open()`/`fork()+execvp()` pattern from both real files, toggling `O_CLOEXEC` on/off, with a child probe checking `fcntl(fd, F_GETFD)` on the known fd number:

```
=== BUGGY (no O_CLOEXEC) ===
[child_probe] fd=3 is STILL OPEN in this process after exec (FD_CLOEXEC was NOT set)

=== FIXED (O_CLOEXEC + close-after-use) ===
[child_probe] fd=3 is CLOSED in this process (fcntl errno=9 Bad file descriptor)
```

**2. Kernel-level, on the real unpatched binary.** Ran the real, unpatched, publicly published agent image (`public.ecr.aws/wazuh-cicd/wazuh/wazuh-agent:5.0.0-latest`), enrolled to a real Wazuh 5.0.0 manager. `/proc/<modulesd_pid>/fdinfo/<fd>` on the live process:

```
flags:  0100002
```

(`O_RDWR`, no `O_CLOEXEC` bit — that would be `02000000`.)

**3. Live, real SCA-spawned children.** Shadowed `/usr/bin/rpm` (used by the default CIS Amazon Linux 2023 SCA policy) with a wrapper dumping `/proc/self/fd/` before exec'ing the real binary. Across 80 real child processes spawned by the real `wm_exec()` during an SCA scan, every one showed the metadata fd open and inherited:

```
lrwx------ 1 root wazuh 64 ... 21 -> /var/ossec/var/run/.wazuh_agent_metadata
```

— in real `rpm -q libselinux`, `rpm -q setroubleshoot`, `rpm -q gdm`, `rpm -q chrony`, etc. — exactly the condition an enforcing SELinux policy denies.

`O_CLOEXEC` is a POSIX flag with kernel-guaranteed, deterministic semantics, not something that can pass "sometimes." Since the mechanism-level repro above already exercises the exact real code pattern both with and without the fix, and the live reproduction already established the real binary's leaked fd and the exact flags it lacked, re-running the full live reproduction against a rebuilt, patched agent image wasn't necessary to validate correctness.

  - [x] Linux
  - [x] Windows 
  - [x] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

### DIFF
--- a/src/shared_modules/agent_metadata/src/metadata_provider.cpp
+++ b/src/shared_modules/agent_metadata/src/metadata_provider.cpp
@@ -292,12 +292,17 @@ namespace
 #else
                 // Unix/Linux: Use mmap on a file
                 // Try read-write first (for writers), fallback to read-only (for readers)
-                m_shm_fd = open(SHM_PATH, O_RDWR | O_CREAT, 0644);
+                // O_CLOEXEC: this fd must not survive into children spawned via fork()+exec()
+                // (e.g. SCA's command rules through wm_exec()) -- an inherited fd here is
+                // exactly what an SELinux policy denies once the child transitions into a
+                // restricted domain (e.g. passwd_t), producing a false-positive AVC even though
+                // the child never touches the file (issue #38813).
+                m_shm_fd = open(SHM_PATH, O_RDWR | O_CREAT | O_CLOEXEC, 0644);
 
                 if (m_shm_fd == -1 && errno == EACCES)
                 {
                     // Permission denied for write, try read-only
-                    m_shm_fd = open(SHM_PATH, O_RDONLY);
+                    m_shm_fd = open(SHM_PATH, O_RDONLY | O_CLOEXEC);
                     m_read_only = true;
                 }
 
@@ -333,6 +338,13 @@ namespace
                     m_shm = nullptr;
                     return;
                 }
+
+                // The mapping stands on its own once mmap() succeeds; the fd itself is not
+                // needed for the mapping to keep working, so drop it now instead of holding it
+                // open (and leaking it into every future child process) for the rest of this
+                // object's lifetime. Belt-and-suspenders with O_CLOEXEC above.
+                close(m_shm_fd);
+                m_shm_fd = -1;
 
                 if (created && !m_read_only)
                 {
--- a/src/shared_modules/agent_metadata/tests/test_metadata_provider.cpp
+++ b/src/shared_modules/agent_metadata/tests/test_metadata_provider.cpp
@@ -19,6 +19,8 @@
 #include <cstdlib>
 #include <sys/wait.h>
 #include <unistd.h>
+#include <dirent.h>
+#include <climits>
 #endif
 
 class MetadataProviderTest : public ::testing::Test
@@ -382,6 +384,39 @@ TEST_F(MetadataProviderTest, ReadFromAtexitAfterProviderTeardownDoesNotCrash)
                                       << ": the provider was torn down while a later exit handler still read it";
     ASSERT_TRUE(WIFEXITED(status));
     EXPECT_EQ(WEXITSTATUS(status), 0);
+}
+
+// #38813: the shm fd was left open for the provider's whole lifetime, with no O_CLOEXEC,
+// so it leaked into every child process spawned afterward (e.g. SCA's command rules via
+// wm_exec()) -- exactly what an enforcing SELinux policy denies once such a child
+// transitions into a restricted domain, even though the child never touches the file.
+// mmap() doesn't need the fd once it succeeds, so the fix closes it right away instead of
+// relying on O_CLOEXEC alone; this asserts that steady state directly.
+TEST_F(MetadataProviderTest, MetadataFdNotLeftOpenAfterConstruction)
+{
+    agent_metadata_t metadata = createSampleMetadata();
+    ASSERT_EQ(metadata_provider_update(&metadata), 0); // triggers singleton construction
+
+    DIR* fdDir = opendir("/proc/self/fd");
+    ASSERT_NE(fdDir, nullptr);
+
+    struct dirent* entry;
+    char linkTarget[PATH_MAX];
+
+    while ((entry = readdir(fdDir)) != nullptr)
+    {
+        std::string path = std::string("/proc/self/fd/") + entry->d_name;
+        ssize_t len = readlink(path.c_str(), linkTarget, sizeof(linkTarget) - 1);
+
+        if (len > 0)
+        {
+            linkTarget[len] = '\0';
+            EXPECT_EQ(std::string(linkTarget).find(".wazuh_agent_metadata"), std::string::npos)
+                    << "fd " << entry->d_name << " still points at the metadata shm file";
+        }
+    }
+
+    closedir(fdDir);
 }
 
 #endif
--- a/src/shared_modules/agent_metadata/tests/test_metadata_provider.cpp
+++ b/src/shared_modules/agent_metadata/tests/test_metadata_provider.cpp
@@ -392,6 +392,10 @@ TEST_F(MetadataProviderTest, ReadFromAtexitAfterProviderTeardownDoesNotCrash)
 // transitions into a restricted domain, even though the child never touches the file.
 // mmap() doesn't need the fd once it succeeds, so the fix closes it right away instead of
 // relying on O_CLOEXEC alone; this asserts that steady state directly.
+//
+// Linux-only: relies on /proc/self/fd, which macOS (no procfs) doesn't have. SELinux is
+// Linux-only too, so the regression this guards against can't occur on macOS anyway.
+#ifdef __linux__
 TEST_F(MetadataProviderTest, MetadataFdNotLeftOpenAfterConstruction)
 {
     agent_metadata_t metadata = createSampleMetadata();
@@ -418,5 +422,6 @@ TEST_F(MetadataProviderTest, MetadataFdNotLeftOpenAfterConstruction)
 
     closedir(fdDir);
 }
+#endif // __linux__
 
 #endif


cborla (dismissed):
LGTM

- [x] Code changes reviewed

cborla (approved):
LGTM
- [x] Code changes reviewed

lchico (approved):
LGTM!

- [x] Code changes reviewed

PR #38830, opened 2026-09-02
Add support for remote image references (GHCR)

### LINKED ISSUE #38587: Implement remote image references from GHCR
## Description

Implement remote image references for the `container_images` module, so the agent inventories images hosted on the GitHub Container Registry without container tooling on the host.

The module reads images available on the host today. This issue adds the `<ref>` reference entry type, already accepted by the configuration, and everything behind it: resolve the reference, authenticate against the registry, retrieve the image, and inventory it with the package extraction that already exists.

Parent research/spike: #35608

## Scope

- Implement the `<ref>` reference entry type, resolving a reference and retrieving the image it points at.
- Authenticate against the registry using the method it advertises.
- Support private GHCR repositories with a user name and an access token, and public ones with no credential configured.
- Accept references pinned by tag and references pinned by content identifier.
- Select the image variant matching the agent platform, and report a reference with no matching variant.
- Inventory the retrieved image with the existing package extraction, unchanged.
- Add a registry credentials configuration block that refers to the agent keystore by key name, so no credential is written in `ossec.conf`.
- Keep credentials out of every log line and every error message.
- Skip retrieving an image that has not changed since the previous scan.
- Bound a single scan, with a maximum download size per image and per scan and a limit on how long a request may take.
- Retry with a back-off when the registry asks the agent to slow down or a request fails transiently, and give up naming the reference.
- Isolate a reference that cannot be resolved, authenticated or retrieved from the rest of the scan.
- Add unit tests for the authentication, the platform selection and the unchanged-image case.
- Add integration tests against a registry served locally, covering public access, authenticated access, a wrong credential, and a reference with several platform variants.
- Update the module documentation with the reference syntax, the credential setup and the network requirements.

## Proposed configuration

```xml
<container_images>
  <enabled>yes</enabled>
  <interval>1h</interval>
  <references>
    <ref>ghcr.io/owner/app:1.4</ref>
    <ref>ghcr.io/owner/other@sha256:2f3a...</ref>
  </references>
  <registry_auth>
    <registry>
      <host>ghcr.io</host>
      <user_keystore_key>ghcr_user</user_keystore_key>
      <passkey_keystore_key>ghcr_token</passkey_keystore_key>
    </registry>
  </registry_auth>
</container_images>
```

Credentials are deposited beforehand, outside the configuration file:

```
wazuh-keystore -f container_images -k ghcr_user  -v "owner"
wazuh-keystore -f container_images -k ghcr_token -v "<access token>"
```

A reference whose registry has no matching entry is attempted without credentials.

## Out of scope

- Registries other than GHCR.
- Changes to the package extraction and to the image reading.
- Modifying anything in the registry.
- Image signature and attestation verification.

## Tasks

- [ ] Implement the `<ref>` reference entry type and the resolution of a reference to its image.
- [ ] Implement authentication against the registry using the advertised method.
- [ ] Support private repositories with a user name and an access token, and public ones with no credential.
- [ ] Accept references pinned by tag and by content identifier.
- [ ] Implement the platform variant selection.
- [ ] Inventory the retrieved image with the existing package extraction.
- [ ] Add the registry credentials configuration block and its validation.
- [ ] Read the credentials from the agent keystore and keep them out of the logs.
- [ ] Skip retrieving an image that has not changed since the previous scan.
- [ ] Add the per-image and per-scan download bounds and the request time limit.
- [ ] Implement the retry and back-off behaviour.
- [ ] Add unit tests for the authentication, the platform selection and the unchanged-image case.
- [ ] Add integration tests against a locally served registry.

## Acceptance criteria

- A public GHCR reference is inventoried with no credential configured.
- A private GHCR reference is inventoried with a valid credential read from the keystore.
- A private GHCR reference with a missing or invalid credential is reported with one clear log line, and the remaining references are still scanned.
- The packages stored for a remote reference match the ones stored for the same image scanned from a saved archive.
- A reference pointing at several platform variants is inventoried from the one matching the agent, and a reference with no matching variant is reported.
- A reference pinned by content identifier is inventoried.
- A second scan of an unchanged reference retrieves no image contents.
- No credential value appears in the agent log at any log level.
- A registry asking the agent to slow down is retried with a back-off and does not stall the scan.
- No support for any registry other than GHCR, and no change to the package extraction or the image reading, is introduced in this issue.

anromerom:
## Description

Closes #38587

The `<ref>` reference entry type was accepted by the configuration and reported as unimplemented, so an image in a registry could not be inventoried at all. This PR implements it for GitHub Container Registry: a reference is resolved against the registry, authenticated by the method the registry advertises, matched to the agent's own platform, and its layers are streamed through the package extraction that already exists, so a remote image and the same image read from a saved archive produce the same inventory. Credentials are named by keystore key in the configuration and read at scan time from a new agent credential store, so no credential value is written into `ossec.conf`.

## Proposed Changes

- Add a registry reader behind the existing reader interface, resolving a reference, obtaining a token from the endpoint the `401` challenge advertises, and reading the image with the layer composition and package parsers unchanged.
- Add a registry transport owning its own cURL handle, because the shared HTTP wrapper exposes neither response headers, without which the advertised authentication method and `Retry-After` cannot be read, nor incremental delivery, without which a layer would be held in memory in full.
- Stream a layer blob rather than download it, bridging cURL's write callback to the pull-model byte stream with the multi interface and `CURL_WRITEFUNC_PAUSE`, on the caller's thread with no extra thread and no locking.
- Select the image variant matching the agent, skipping the build attestation entries an index carries, and report a reference that offers no matching variant along with the platforms it does offer.
- Add the `<registry_auth>` block and the `<ca_bundle>` element to the configuration grammar, and an agent credential store with a `wazuh-agent-keystore` tool, so the configuration names keys and never values.
- Resolve the certificate bundle at run time instead of trusting the one compiled into the HTTP library, which is whichever path existed on the machine that built the package, and refuse the reference when none is found rather than connecting unverified.
- Bound a scan: a byte ceiling per image and per scan, a wall-clock ceiling per layer transfer, a retry with back-off honouring `Retry-After`, and a stop request polled between reads so a slow registry cannot hold the shutdown budget every module shares.
- Fix a pre-existing defect in `loadStored()`, which brace-initialized the parsed tag list and so recovered none of it: an image known by more than one tag lost its tags on the first rescan that found it unchanged.

### Results and Evidence

A real multi-platform image pulled end to end on an agent host, `ghcr.io/astral-sh/uv:python3.12-bookworm-slim`:

```sql
agent platform: linux/amd64
selected manifest: sha256:5d275ca5f0da33c3368ac8fbb85fafabad023b3b8a7cff39a94ac0baecfd9a50 (linux/amd64)
layer 1 sha256:4831516dd0cb gzip streamed=28228487 complete=1 dbs=1
layer 2 sha256:f59651dbc0ea gzip streamed=3517407  complete=1 dbs=1
layer 3 sha256:322e9e97e47b gzip streamed=13667910 complete=1 dbs=1
layer 4 sha256:15155ecba1d2 gzip streamed=250      complete=1 dbs=0
layer 5 sha256:590f2846d3a7 gzip streamed=22786668 complete=1 dbs=0

total streamed bytes = 68200722
composed databases   = 1
  var/lib/dpkg/status (90360 bytes)
packages extracted   = 105
peak RSS kB          = 17456
```

**68.2 MB streamed at 17.4 MB peak resident memory.** Buffering the layers would have put peak memory above the streamed total; the cost of reading a layer is the 256 KiB suspend threshold, not the image size. The composed database appears in three of the five layers and the result is the last one, so the composition rules are exercised rather than bypassed. The 105 packages carry the versions Debian 12 ships, including `apt 2.6.1` and `base-files 12.4+deb12u13`.

The same reader driven against a registry served locally and answering as `ghcr.io`, which is what the integration cases use:

| Case | Result |
|---|---|
| Public reference, no credential configured | Inventoried, `linux/amd64` selected out of an index offering amd64 and arm64 |
| Layer retrieved through the redirect the registry answers with | The content host received `Authorization: None`; the token is not carried across it |
| Private reference, valid credential from the store | Inventoried, credential accepted |
| Private reference, wrong credential | Reported, and no image content retrieved |
| Private reference, nothing in the store | Reported, and no image content retrieved |
| Index offering only a foreign platform | Declined from the index alone, no blob requested |
| Manifest outside an index for a foreign platform | Configuration blob read to learn the platform, then declined, no layer requested |
| **Second scan of an unchanged reference** | **`unchanged`, 0 bytes, zero blob requests** |
| Registry answering `429` with `Retry-After` | Retried and resolved, scan not stalled |
| Reference naming another registry | Refused with no request attempted |
| Configured certificate bundle absent | Refused rather than connecting unverified |

No credential value reaches the log at any level, and the configuration dump reports key names only.

### Artifacts Affected

- `libcontainer_images.so`, loaded by `wazuh-modulesd`. DEB, RPM and macOS. The library still builds for Windows, where registry support is not offered.
- `wazuh-agent-keystore`, a new agent binary, with `queue/credentials/` created root-only by the installer. Not built for Windows.
- `wazuh-modulesd` and the shared configuration library, for the new grammar. The configuration parser for this module is compiled into the manager as well, which validates a block pushed through a shared `agent.conf`, so the manager binary changes even though it does not run the module.
- No new link dependency on any platform. libcurl already reaches the agent through `libwazuhext.so`, and the credential store reuses the OpenSSL that arrives the same way.

### Configuration Changes

New, all optional and all backward compatible. An existing `<container_images>` block behaves exactly as before.

| Element | Type | Default | Purpose |
|---|---|---|---|
| `references/ref` | string | empty | Image in GHCR, by tag or by digest. Already accepted by the grammar; now implemented. |
| `registry_auth/registry/host` | string | empty | Registry the credential entry applies to. |
| `registry_auth/registry/user_keystore_key` | string | empty | Name of the credential store key holding the user name. |
| `registry_auth/registry/passkey_keystore_key` | string | empty | Name of the credential store key holding the access token. |
| `ca_bundle` | string | detected | Certificate bundle used to verify a registry. Probed at run time when unset. |

A registry with no entry in `<registry_auth>` is attempted with no credential, which is what a public repository needs. An entry whose keys are absent from the store fails that reference rather than falling back to anonymous, so an operator's mistake is reported instead of being hidden behind a `404`.

Anything malformed inside `<registry_auth>` is reported and skipped rather than rejected, matching the existing behaviour of `<references>`: rejecting invalidates the whole module configuration, and the control script tests every daemon before starting any of them.

The scan bounds are fixed defaults rather than configuration: 2 GiB per image, 8 GiB per scan, 30 s per metadata request, 5 min per layer transfer, four attempts with a doubling back-off.

### Documentation Updates

- `docs/ref/modules/container-images/configuration.md`: the reference syntax for both pinning forms, the credential block, what the credential store does and does not protect, the certificate resolution order, the bounds on a scan, and the network requirement naming both `ghcr.io` and `pkg-containers.githubusercontent.com`.
- `docs/ref/modules/container-images/architecture.md`: the registry reader and the registry byte stream, including why the transport is separate from the shared HTTP wrapper.
- `docs/ref/modules/container-images/README.md`: the capability line corrected.

The credential store is documented as keeping the credential out of `ossec.conf`, and therefore out of configuration backups, shared-configuration pushes and support bundles, with the file's permissions as the control that protects it locally. It is not documented as encryption at rest, because the stored bytes are produced by the same mechanism the manager keystore uses and that mechanism keeps its key beside the value.

### Tests Introduced

**Unit, `container_images_unit_test`: 118 to 245.** Reference parsing including path traversal, an implicit registry and a registry port mistaken for a tag; the credential store round trip, malformed and truncated values, corrupted ciphertext and the stored permissions; certificate resolution order and its fail-closed outcome; `WWW-Authenticate` parsing including a value containing a comma, and a token URL that refuses a plain-http realm or an injected parameter; the retry schedule and `Retry-After` including its clamp; platform name normalization and variant matching; transport and byte-stream guards; and eighteen reader cases over an injected transport covering the anonymous and authenticated token paths, a credential missing from the store, a token endpoint outside the reference's registry, the unchanged-digest short circuit asserting zero blob requests, and both foreign-platform shapes.

**Unit, `test_wm_container_images`: 14 to 25.** The credential block, several registries, one key name only, a missing host, no key names at all, unknown tags, an empty block, the certificate element with its empty and repeated forms, and the defaults.

**Integration, `tests/integration/test_container_images/test_registry/`: five cases** against a registry served locally that answers as `ghcr.io`, covering public access, authenticated access, a wrong credential, an absent credential, platform variant selection in both the index and the bare-manifest shape, and the unchanged rescan asserted as a blob count rather than as a duration. The registry is a fixture in the suite, so nothing reaches the network and no container engine is involved.

The two older test packages in that suite now take their log patterns and database helpers from the suite's own `framework_module/`, which is where they were always documented as living. They previously imported them from `wazuh_testing.modules.modulesd.container_images`, which the published `qa-integration-framework` does not ship, so those imports only resolved on a host whose framework had been modified by hand.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/.github/actions/check_files/deb_linux_agent_amd64.csv
+++ b/.github/actions/check_files/deb_linux_agent_amd64.csv
@@ -5,6 +5,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/bin/wazuh-logcollector,root,root,750,file,-rwxr-x---,690248,0.1
 /var/ossec/bin/wazuh-control,root,root,750,file,-rwxr-x---,9331,0.1
 /var/ossec/bin/manage_agents,root,root,750,file,-rwxr-x---,174264,0.1
+/var/ossec/bin/wazuh-agent-keystore,root,root,750,file,-rwxr-x---,0,0
 /var/ossec/bin/wazuh-modulesd,root,root,750,file,-rwxr-x---,537016,0.1
 /var/ossec/bin/wazuh-agentd,root,root,750,file,-rwxr-x---,661568,0.1
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12374168,0.1
--- a/.github/actions/check_files/deb_linux_agent_arm64.csv
+++ b/.github/actions/check_files/deb_linux_agent_arm64.csv
@@ -5,6 +5,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/bin/wazuh-logcollector,root,root,750,file,-rwxr-x---,690248,0.5
 /var/ossec/bin/wazuh-control,root,root,750,file,-rwxr-x---,9067,0.1
 /var/ossec/bin/manage_agents,root,root,750,file,-rwxr-x---,174264,0.5
+/var/ossec/bin/wazuh-agent-keystore,root,root,750,file,-rwxr-x---,0,0
 /var/ossec/bin/wazuh-modulesd,root,root,750,file,-rwxr-x---,597472,0.5
 /var/ossec/bin/wazuh-agentd,root,root,750,file,-rwxr-x---,661568,0.5
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12374168,0.5
--- a/.github/actions/check_files/deb_linux_agent_i386.csv
+++ b/.github/actions/check_files/deb_linux_agent_i386.csv
@@ -5,6 +5,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/bin/wazuh-logcollector,root,root,750,file,-rwxr-x---,1055440,0.1
 /var/ossec/bin/wazuh-control,root,root,750,file,-rwxr-x---,9331,0.1
 /var/ossec/bin/manage_agents,root,root,750,file,-rwxr-x---,236032,0.1
+/var/ossec/bin/wazuh-agent-keystore,root,root,750,file,-rwxr-x---,0,0
 /var/ossec/bin/wazuh-modulesd,root,root,750,file,-rwxr-x---,916748,0.1
 /var/ossec/bin/wazuh-agentd,root,root,750,file,-rwxr-x---,1011784,0.1
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12358288,0.1
--- a/.github/actions/check_files/macos_agent_arm64.csv
+++ b/.github/actions/check_files/macos_agent_arm64.csv
@@ -5,6 +5,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /Library/Ossec/bin/wazuh-execd,root,wheel,750,file,-rwxr-x---,,
 /Library/Ossec/bin/wazuh-syscheckd,root,wheel,750,file,-rwxr-x---,,
 /Library/Ossec/bin/manage_agents,root,wheel,750,file,-rwxr-x---,,
+/Library/Ossec/bin/wazuh-agent-keystore,root,wheel,750,file,-rwxr-x---,,
 /Library/Ossec/bin/wazuh-agentd,root,wheel,750,file,-rwxr-x---,,
 /Library/Ossec/bin/wazuh-modulesd,root,wheel,750,file,-rwxr-x---,,
 /Library/Ossec/etc/internal_options.conf,root,wazuh,640,file,-rw-r-----,,
--- a/.github/actions/check_files/macos_agent_x86_64.csv
+++ b/.github/actions/check_files/macos_agent_x86_64.csv
@@ -5,6 +5,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /Library/Ossec/bin/wazuh-execd,root,wheel,750,file,-rwxr-x---,,
 /Library/Ossec/bin/wazuh-syscheckd,root,wheel,750,file,-rwxr-x---,,
 /Library/Ossec/bin/manage_agents,root,wheel,750,file,-rwxr-x---,,
+/Library/Ossec/bin/wazuh-agent-keystore,root,wheel,750,file,-rwxr-x---,,
 /Library/Ossec/bin/wazuh-agentd,root,wheel,750,file,-rwxr-x---,,
 /Library/Ossec/bin/wazuh-modulesd,root,wheel,750,file,-rwxr-x---,,
 /Library/Ossec/etc/internal_options.conf,root,wazuh,640,file,-rw-r-----,,
--- a/.github/actions/check_files/rpm_linux_agent_amd64.csv
+++ b/.github/actions/check_files/rpm_linux_agent_amd64.csv
@@ -69,6 +69,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/bin/wazuh-logcollector,root,root,750,file,-rwxr-x---,653320,0.1
 /var/ossec/bin/wazuh-control,root,root,750,file,-rwxr-x---,9331,0.1
 /var/ossec/bin/manage_agents,root,root,750,file,-rwxr-x---,167456,0.1
+/var/ossec/bin/wazuh-agent-keystore,root,root,750,file,-rwxr-x---,0,0
 /var/ossec/bin/wazuh-modulesd,root,root,750,file,-rwxr-x---,508352,0.1
 /var/ossec/bin/wazuh-agentd,root,root,750,file,-rwxr-x---,624632,0.1
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12417720,0.1
--- a/.github/actions/check_files/rpm_linux_agent_arm64.csv
+++ b/.github/actions/check_files/rpm_linux_agent_arm64.csv
@@ -69,6 +69,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/bin/wazuh-logcollector,root,root,750,file,-rwxr-x---,653320,0.5
 /var/ossec/bin/wazuh-control,root,root,750,file,-rwxr-x---,9067,0.1
 /var/ossec/bin/manage_agents,root,root,750,file,-rwxr-x---,167456,0.5
+/var/ossec/bin/wazuh-agent-keystore,root,root,750,file,-rwxr-x---,0,0
 /var/ossec/bin/wazuh-modulesd,root,root,750,file,-rwxr-x---,564896,0.5
 /var/ossec/bin/wazuh-agentd,root,root,750,file,-rwxr-x---,624632,0.5
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12417720,0.5
--- a/.github/actions/check_files/rpm_linux_agent_i386.csv
+++ b/.github/actions/check_files/rpm_linux_agent_i386.csv
@@ -69,6 +69,7 @@ full_filename,owner_name,group_name,mode,type,prot_permissions,size_bytes,size_e
 /var/ossec/bin/wazuh-logcollector,root,root,750,file,-rwxr-x---,821964,0.1
 /var/ossec/bin/wazuh-control,root,root,750,file,-rwxr-x---,9331,0.1
 /var/ossec/bin/manage_agents,root,root,750,file,-rwxr-x---,190936,0.1
+/var/ossec/bin/wazuh-agent-keystore,root,root,750,file,-rwxr-x---,0,0
 /var/ossec/bin/wazuh-modulesd,root,root,750,file,-rwxr-x---,711912,0.1
 /var/ossec/bin/wazuh-agentd,root,root,750,file,-rwxr-x---,789888,0.1
 /var/ossec/lib/libwazuhext.so,root,wazuh,750,file,-rwxr-x---,12362096,0.1
--- a/docs/ref/modules/container-images/README.md
+++ b/docs/ref/modules/container-images/README.md
@@ -100,7 +100,7 @@ wazuh-modulesd:container_images: INFO: Scan ended. 1 references, 132 packages.
 - RPM package extraction is not implemented; an RPM-based image is inventoried with zero packages and a warning.
 - Agent Sync Protocol synchronization is not implemented, so the stored inventory stays on the agent.
 - Vulnerability Detector integration is not implemented.
-- Container engine, registry, Windows, and Kubernetes integrations are not implemented.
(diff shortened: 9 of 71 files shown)


anromerom:
## Validation


### Unit tests

`container_images_unit_test`, the module's own suite, grown from 118 cases to 245:

```
[==========] 245 tests from 18 test suites ran. (435 ms total)
[  PASSED  ] 245 tests.
```

`test_wm_container_images`, the configuration parser compiled into both the agent and the manager, grown from 14 cases to 25:

```
[==========] tests: 25 test(s) run.
[  PASSED  ] 25 test(s).
```

New coverage: reference parsing including path traversal, an implicit registry and a registry port mistaken for a tag; the credential store round trip, malformed and truncated values, corrupted ciphertext and the stored permissions; certificate resolution order and its fail-closed outcome; `WWW-Authenticate` parsing including a value containing a comma, and a token URL that refuses a plain-http realm or an injected parameter; the retry schedule and `Retry-After` including its clamp; platform matching and its variants; transport and byte-stream guards; and eighteen reader cases over an injected transport, covering the anonymous and authenticated token paths, a credential missing from the store, a token endpoint outside the reference's registry, the unchanged-digest short circuit asserting zero blob requests, and both foreign-platform shapes.

A from-scratch rebuild of the module and its tests produces no compiler warnings.

### Registry protocol, against a registry answering as `ghcr.io`

Two harnesses drive the real reader against a TLS registry served locally that presents itself as `ghcr.io` on port 443, so the code under test takes exactly the path it takes in production, including the token exchange and the redirect to a separate content host.

| Group | Result |
|---|---|
| Public repository, anonymous token | Pass |
| Private repository, token from a stored credential | Pass |
| Credential wrong | Pass, refusal reported |
| Credential absent from the store | Pass, reference failed rather than retried anonymously |
| Platform variant selected out of an index | Pass |
| Foreign-platform index, no matching variant | Pass, no layer requested |
| Bare manifest with no platform | Pass, declined after reading its configuration blob |
| Layer arriving through the content-host redirect | Pass, token withheld from the redirect target |
| Unchanged reference rescanned | Pass, 0 bytes retrieved |
| Throttled response | Pass, retried on the back-off |

All ten groups pass. The withheld-token check is the one worth naming: GHCR redirects a blob to `pkg-containers.githubusercontent.com`, and the harness asserts the registry credential does not travel to that host.

### Logging isolation

The module's logging helper shared its exported symbols with the SCA module's copy of the same class, so whichever library `wazuh-modulesd` loaded first owned the singleton and the other module's lines came out under its tag. Namespacing the module's copy ends it.

Symbols shared between the two libraries, before and after:

```
$ comm -12 <(nm -D --defined-only libcontainer_images.so | grep LoggingHelper | awk '{print $3}' | sort) \
           <(nm -D --defined-only libsca.so | grep LoggingHelper | awk '{print $3}' | sort) | wc -l
0
```

A regression test loads both libraries in one process and asserts each module's lines reach its own callback. It cannot live in either unit suite, because the defect is a property of two shared objects loaded together:

```
$ ./logging_tag_isolation build/lib
container images callback: 1 line(s)
sca callback:              0 line(s)
PASS: each module's lines reach its own callback
```

It fails with the namespace removed and passes with it restored, ordered the way the failure appeared in the field.

### Manual end to end, real agent against real GHCR

Run on a Linux agent built from this branch, against three images published to GHCR for the purpose: a public multi-platform one, a private one, and one built only for a foreign platform.

**Public reference, no credential configured.**

```sql
INFO: Configuration loaded: enabled=yes, scan_on_start=yes, interval=3600, references=1.
DEBUG: No credential is configured for ghcr.io, requesting an anonymous token.
INFO: Scan ended. 1 references, 110 packages.
```

The variant selected, and the digest recorded, matching the `config.digest` read from the registry by hand:

```
reference_value                        platform_os  platform_architecture
-------------------------------------  -----------  ---------------------
ghcr.io/OWNER/wazuh-ci-public:1.0      linux        amd64

image_config_digest
sha256:8a163245f01b132ff5f2f70ef5098666d1cba552e9a1d4884fefd0aaabaad5fa
```

Real packages, not placeholders:

```
name         version_       architecture  type
-----------  -------------  ------------  ----
adduser      3.134          all           deb
apt          2.6.1          amd64         deb
base-files   12.4+deb12u15  amd64         deb
base-passwd  3.6.1          amd64         deb
bash         5.2.15-2+b13   amd64         deb
```

**Private reference with a stored credential.** Both references inventoried, both `amd64`, the private one yielding the same 110 packages from a real transfer:

```sql
DEBUG: Reference 'ghcr.io/OWNER/wazuh-ci-private:1.0' yielded 110 package(s) from 34219092 streamed byte(s).
```

**Credential absent from the store.** Reported, and the stored inventory kept rather than deleted, which is the behaviour that matters most:

```sql
WARNING: Reference 'ghcr.io/OWNER/wazuh-ci-private:1.0' could not be resolved: the credential configured for ghcr.io is missing from the agent credential store.
WARNING: Reference 'ghcr.io/OWNER/wazuh-ci-private:1.0' could not be read. Keeping the 110 package(s) already stored for it, which are left as they were.
```

The 110 packages are still there afterwards.

**Credential wrong.** Reported as the registry's own answer, with no credential value in the line:

```sql
WARNING: Reference 'ghcr.io/OWNER/wazuh-ci-private:1.0' could not be resolved: the registry refused to issue a token, answering 403.
WARNING: Reference 'ghcr.io/OWNER/wazuh-ci-private:1.0' could not be read. Keeping the 110 package(s) already stored for it, which are left as they were.
```

**Reference offering no matching variant.** Named, with what it does offer, and no layer requested:

```sql
WARNING: Reference 'ghcr.io/OWNER/wazuh-ci-foreign:1.0' was not inventoried: no image variant matches linux/amd64, the reference offers linux/ppc64le.
```

**Reference pinned by digest.** Inventoried, with the digest recorded as its name:

```
reference_value                                            tag
---------------------------------------------------------  ---------------------------------------------------------
ghcr.io/OWNER/wazuh-ci-public@sha256:f1830d85156cafa2c...   sha256:f1830d85156cafa2c...
```

**Second scan of an unchanged reference.** Metadata read, then stopped:

```sql
DEBUG: Reference 'ghcr.io/OWNER/wazuh-ci-public:1.0' still reports sha256:8a163245f01b132ff5f2f70ef5098666d1cba552e9a1d4884fefd0aaabaad5fa, so no image contents were retrieved.
```

Across the whole run the streamed-byte count stayed at 3 and the stored package count at 110, so the rescans transferred nothing.

**Remote compared against the same image on disk.** The packages stored for the remote reference and for the same image read through `<archive>` are identical:

```
$ diff /tmp/remote.txt /tmp/archive.txt && echo "identical"
identical
```

**Certificate bundle that does not exist.** Refused rather than fetched unverified:

```sql
WARNING: Reference 'ghcr.io/OWNER/wazuh-ci-public@sha256:f1830d85...' cannot be verified: the configured certificate bundle '/etc/ssl/certs/does-not-exist.crt' does not exist.
```

**Log tag, after the logging fix, on the same agent.** The installed library was verified to carry the namespaced symbols before the test, then the agent was started six times:

```
$ nm -DC /var/ossec/lib/libcontainer_images.so | grep -c "containerimages::LoggingHelper"
5

$ grep "wazuh-modulesd:sca" ossec.log | grep -E "Scan ended|Reference '|yielded|streamed byte|image contents"
$
```

No output: no container-images line under the `sca` tag across six starts. The same command against the previous library returned scan lines wearing that tag.

No credential value reaches the log at any level, and the configuration dump reports key names only.

PR #38836, opened 2026-09-02
fix(vulnerability-scanner): wait for in-flight feed update before teardown on stop()

### LINKED ISSUE #38812: wazuh-manager-modulesd SIGSEGV in StoreModel::handleRequest() when stopped during initial vulnerability feed load
## Description

During the OVA validation for [Release 5.0.0 - Beta 5 - Specific systems](https://github.com/wazuh/wazuh/issues/38793), `wazuh-manager-modulesd` received `SIGSEGV` while the manager was being restarted during the initial vulnerability feed load.

This is not only a kernel-log string match: Linux Audit recorded an `ANOM_ABEND` event with signal 11 for `/var/wazuh-manager/bin/wazuh-manager-modulesd`. Matching Beta 5 symbols resolve the fault to `StoreModel::handleRequest()` / `AbstractHandler::handleRequest()` in the vulnerability scanner.

The manager restarted successfully and remained functional. A later restart after the initial feed had completed was clean, so the observed result is `1/2` restarts under different feed states. The failure appears timing-dependent.

## Environment

| Field | Value |
|---|---|
| Artifact | `wazuh-5.0.0-beta5.ova` |
| Artifact SHA-512 | `8ebf3a8a64354038c374160e3af3a715afc4bd4d599f7cf03d2fc22b216cee1c192a5b95866c2fee77cb6bef686c17eb194a6f792b5b75c175632515952e3be3` |
| OVA OS | Amazon Linux 2023 |
| Kernel | `6.1.182-227.379.amzn2023.x86_64` |
| Hypervisor | VirtualBox 7.0 |
| VM allocation used | 4 vCPU, 8192 MiB RAM, 50 GiB disk |
| Manager package | `wazuh-manager-5.0.0-1.x86_64` |
| Manager build | `5.0.0`, stage `beta5`, commit `4ad980de569fec1fb1578f953bbb24fdc6a3a736` |
| Library SHA-256 | `1983ccf1c33351ce46a95760f898c291739bae06cf5d22b0c889d6511a21898f` |
| Vulnerability detection | Enabled |

`rpm -V wazuh-manager` did not report modifications under `/var/wazuh-manager/bin` or `/var/wazuh-manager/lib`. No OOM-killer or out-of-memory event was present in the kernel log.

## Configuration

```xml
<vulnerability-detection>
  <enabled>yes</enabled>
  <feed-update-interval>60m</feed-update-interval>
</vulnerability-detection>
```

## Steps to reproduce

1. Import and start the official Wazuh 5.0.0 Beta 5 OVA.
2. Wait for `IndexerDownloader` to begin the initial full vulnerability feed load:

   ```console
   wazuh-manager-modulesd:content-updater: INFO: IndexerDownloader: Starting initial full load (slices=2)
   ```

3. While that initial load is still running, restart the manager:

   ```console
   sudo systemctl restart wazuh-manager
   ```

4. Inspect the kernel and audit logs:

   ```console
   sudo journalctl -k -b | grep -E 'general protection|segfault|libvulnerability_scanner'
   sudo grep 'type=ANOM_ABEND' /var/log/audit/audit.log
   ```

## Actual behavior

The kernel recorded a general protection fault in the vulnerability-scanner library:

```console
Sep 02 10:02:27 wazuh kernel: traps: wazuh-manager-m[7241] general protection fault ip:7f57398f928b sp:7f57141c7710 error:0 in libvulnerability_scanner.so[7f5739878000+17a000]
```

Linux Audit independently confirms that the Wazuh process terminated with `SIGSEGV`:

```console
type=ANOM_ABEND msg=audit(1788343347.287:998): auid=4294967295 uid=992 gid=991 ses=4294967295 pid=2975 comm="wazuh-manager-m" exe="/var/wazuh-manager/bin/wazuh-manager-modulesd" sig=11 res=1
```

PID `2975` was the process leader and TID `7241` was the faulting thread:

```console
2026/09/02 09:55:59 wazuh-manager-modulesd: INFO: Started (pid: 2975).
2026/09/02 09:55:59 wazuh-manager-modulesd:vulnerability-scanner: INFO: Started (pid: 2975).
2026/09/02 09:59:59 wazuh-manager-modulesd:content-updater: INFO: IndexerDownloader: Starting initial full load (slices=2)
2026/09/02 10:02:27 wazuh-manager-modulesd:vulnerability-scanner: INFO: Module finished.
2026/09/02 10:02:27 wazuh-manager-modulesd:vulnerability-scanner:server: INFO: vulnerability detector server fully stopped.
```

After systemd started the manager again, the next initial load completed successfully:

```console
2026/09/02 10:10:28 wazuh-manager-modulesd:content-updater: INFO: IndexerDownloader: Initial load download phase complete — 380080 documents, cursor: '905742'
2026/09/02 10:10:29 wazuh-manager-modulesd:vulnerability-scanner: INFO: Feed update process completed (changed=true).
2026/09/02 10:10:29 wazuh-manager-modulesd:vulnerability-scanner: INFO: CVE feed fully loaded — per-agent scans unblocked.
```

One later restart, after the feed was fully available, did not generate another fault.

## Symbolization

The offset shown by the kernel, `0x8128b`, is relative to the executable RX mapping. The executable ELF segment begins at virtual address `0x78000`, so the corresponding ELF virtual address is `0xf928b`.

Using the matching Beta 5 symbols:

```text
AbstractHandler<std::shared_ptr<EventContext>>::handleRequest()
  src/shared_modules/utils/chainOfResponsability.hpp:119
inlined from:
StoreModel::handleRequest()
  src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/storeModel.hpp:103
```

The faulting instruction is:

```asm
mov (%rbx), %rax
```

This is the virtual dispatch to the next handler after a non-null check. The fault is consistent with an invalid or dangling handler pointer.

## Evidence

Image

MiguelazoDS:
## Description

`wazuh-manager-modulesd` could crash with `SIGSEGV` if the manager was restarted while the vulnerability scanner's initial feed load was still in progress. During shutdown, `VulnerabilityScannerFacade::stop()` destroyed the feed manager while a content-updater worker thread could still be mid-call inside its handler chain, so the running call kept executing on freed memory. This was reported during 5.0.0 Beta 5 release QA, reproducing 1 out of 2 restarts.

Closes #38812

## Proposed Changes

- Made `stop()` wait (bounded, 100ms poll / 3s timeout) to acquire the feed manager's existing internal mutex before resetting it, so an in-flight feed update finishes and releases the mutex before the object it's running on is destroyed. Reuses the same mutex `processMessage()` already holds for the duration of one page, mirroring the bounded-wait pattern already used elsewhere in this codebase (`syscheckd`) for the same class of shutdown race.
- Added a unit test proving `processMessage()` holds the mutex for its entire call — the property the fix's wait depends on — and a component-level non-regression test for `stop()`'s disabled-module path.
- **Accepted trade-off on timeout expiry**: if the wait never acquires the mutex within 3s, `stop()` still proceeds with the reset — the original race is still theoretically possible in that specific case. This matches the same trade-off `syscheckd`'s existing bounded-wait precedent already makes: refusing to proceed would hang the whole module's shutdown indefinitely, which is worse than a rare crash under a condition (a worker genuinely wedged, not just slow) that's already abnormal. Real measured `processMessage()` durations (see below) stay ~12x under the 3s bound, so hitting the timeout in practice would itself be a sign of a separate problem, not this logic.
- Extracted the wait loop into `waitToAcquireMutex()` (same 100ms/3s behavior, no functional change) and added direct unit tests for its acquired/timeout branches — CI's coverage gate caught that the original non-regression test never actually executed the new code, since it only exercised the disabled-module path.
- **Kept the timeout log at `WARN`, not `debug1`**: it should never fire in normal operation, but if it ever does, it's the one signal an operator would have that the residual crash risk above was live at that shutdown — worth surfacing by default rather than requiring debug logging to notice.

**Update: the wait above was guarding the wrong `reset()` call.** `m_databaseFeedManager` has more than one owner — `m_scanOrchestrator`, and every scanner node inside it, each hold their own `shared_ptr` copy — so `m_databaseFeedManager.reset()` never actually drops the object to zero references while the module is running normally. The real destructor only ran later, at the completely unguarded `m_scanOrchestrator.reset()` a few lines below, so the reported crash remained fully reproducible against the fix above. Additionally, an in-flight agent scan can read the same scan-orchestration chains through the same mutex, which the original wait never accounted for either.

- `TDatabaseFeedManager::teardown()` now explicitly and synchronously stops the content-updater (`m_contentRegistration.reset()`) before returning, instead of relying on the class's member-declaration order to eventually stop it — too late — as an implicit side effect of destruction. This closes the race regardless of which `shared_ptr` ends up dropping the last reference.
- `stop()`'s bounded wait now guards `m_databaseFeedManager.reset()` and `m_scanOrchestrator.reset()` together, closing the second race: an in-flight agent scan reading `PackageScanner`/`OsScanner`/`EventDetailsBuilder` while they're destroyed.
- `m_shouldStop` in `TDatabaseFeedManager` is now atomic — it was a plain `bool` read and written across threads with no synchronization.
- Fixed a flaky readiness signal and a missing exception guard in the existing mutex test, added a test proving `teardown()` actually stops the updater, and removed a component test that duplicated an existing one without exercising the code it claimed to.

### Results and Evidence

- `vulnerability_scanner_unit_tests`: full suite passed, 0 failures (new `ProcessMessageBlocksConcurrentMutexAcquisition`, `WaitToAcquireMutex_AcquiresWhenReleasedBeforeTimeout`, `WaitToAcquireMutex_TimesOutWhenLockNeverReleased` tests included).
- `vulnerability_scanner_component_tests`: 3 passed / 3 skipped (pre-existing, unrelated `IndexerConnector`-mocking gap) / 0 failed (new `StopWaitsForInFlightFeedUpdate` test included).
- Crash reproduced pre-fix via GDB against the exact crashing binary, confirming the fault at `StoreModel::handleRequest()` / `AbstractHandler::handleRequest()`, matching the issue's reported crash site.
- Real timing measurement across 759 production feed pages: p50=7.3ms, p99=85.8ms, max=242ms — all comfortably under the 3s bound.
- **Update**: `vulnerability_scanner_unit_tests` full suite: 512/512 passed (new `TeardownStopsContentUpdaterBeforeReturning` test included). `vulnerability_scanner_component_tests`: 2/2 passed, 3 skipped (same pre-existing, unrelated gap as above).
- The new `TeardownStopsContentUpdaterBeforeReturning` test was mutation-tested, not just run once: reverting `teardown()` to the earlier `m_shouldStop = true`-only behavior makes it fail (a concrete gmock cardinality violation), and restoring the fix makes it pass again — confirming it actually catches the regression it targets.
- Independently reproduced both races in a standalone, self-contained program mirroring the real object-lifetime shapes (the handler chain, the multiple-`shared_ptr`-owner structure, the member-declaration-order hazard): the unfixed shapes crash with a real, deterministic `SIGSEGV` (backtraces matching the issue's own crash site) on every run, and the corrected shapes run clean on every run under the identical stress.

### DIFF
--- a/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
@@ -25,6 +25,7 @@
 #include "vulnerabilityDescription_generated.h"
 #include "vulnerabilityRemediations_generated.h"
 #include "vulnerabilityScanner.hpp"
+#include <atomic>
 #include <chrono>
 #include <cstdint>
 #include <external/nlohmann/json.hpp>
@@ -200,7 +201,7 @@ public:
             {
                 for (const auto& resource : parsedMessage.at("data"))
                 {
-                    if (m_shouldStop)
+                    if (m_shouldStop.load(std::memory_order_acquire))
                     {
                         m_activeBatch.Clear();
                         m_activeBatchCount = 0;
@@ -392,7 +393,8 @@ public:
                         processMessageResult = processMessage(std::move(message), feedDatabase);
 
                         // If the process was interrupted or it failed, we return prematurely.
-                        if (m_shouldStop || !std::get<FileProcessingResultFields::STATUS>(processMessageResult))
+                        if (m_shouldStop.load(std::memory_order_acquire) ||
+                            !std::get<FileProcessingResultFields::STATUS>(processMessageResult))
                         {
                             logDebug2(WM_VULNSCAN_LOGTAG,
                                       "Feed update process interrupted or failed. Offset: %d, Hash: %s.",
@@ -961,10 +963,20 @@ public:
 
     /**
      * @brief Teardown the database feed manager.
+     *
+     * Explicitly (and synchronously) stops the content-updater here, rather than leaving it to
+     * be destroyed implicitly as a side effect of this object's member declaration order.
+     * ~ContentRegister -> ~Action drains any in-flight on-demand callback (OnDemandManager::
+     * removeEndpoint()'s exclusive lock blocks until one finishes) and then stops and joins the
+     * scheduler thread. By the time this call returns, no thread -- scheduled or on-demand -- can
+     * still be inside processMessage(). This must run before any lock this object's caller might
+     * hold is taken, since the join here can otherwise deadlock against a thread trying to
+     * reacquire that same lock inside processMessage().
      */
     void teardown()
     {
-        m_shouldStop = true;
+        m_shouldStop.store(true, std::memory_order_release);
+        m_contentRegistration.reset();
     }
 
     /**
@@ -1033,7 +1045,7 @@ private:
     std::unique_ptr<TContentRegister> m_contentRegistration;
     std::shared_ptr<TRocksDBWrapper> m_feedDatabase;
     std::unique_ptr<TranslationLRUCache> m_translationL2Cache;
-    bool m_shouldStop = false;
+    std::atomic<bool> m_shouldStop {false};
 
     static constexpr size_t BATCH_COMMIT_MAX_BYTES = 2 * 1024 * 1024;
 
--- a/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
@@ -1042,7 +1042,6 @@ private:
     std::shared_mutex& m_mutex;
     nlohmann::json m_updaterConfig;
     uint32_t m_translationLruSize;
-    std::unique_ptr<TContentRegister> m_contentRegistration;
     std::shared_ptr<TRocksDBWrapper> m_feedDatabase;
     std::unique_ptr<TranslationLRUCache> m_translationL2Cache;
     std::atomic<bool> m_shouldStop {false};
@@ -1053,6 +1052,19 @@ private:
     rocksdb::WriteBatch m_activeBatch {BATCH_COMMIT_MAX_BYTES};
     size_t m_activeBatchCount = 0;
     bool m_initialLoad = false;
+
+    /**
+     * Declared last so it is destroyed *first* (members are destroyed in reverse declaration
+     * order). Its destructor (~ContentRegister -> ~Action) synchronously joins the content-updater's
+     * scheduler thread and drains any in-flight on-demand callback before returning, and that
+     * callback (registered in this class's own constructor) reaches back into m_feedDatabase,
+     * m_activeDecoder and m_activeBatch through `this`. Declaring it last makes that safe
+     * unconditionally, for any path that destroys this object -- teardown() already forces this
+     * same early destruction explicitly, so this ordering changes nothing for the one real caller
+     * (VulnerabilityScannerFacade::stop()); it only removes the dependency on every future caller
+     * remembering to call teardown() first.
+     */
+    std::unique_ptr<TContentRegister> m_contentRegistration;
 
     /**
      * @brief Reads the global maps from the local feed database and loads them into memory.
--- a/src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScannerFacade.cpp
+++ b/src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScannerFacade.cpp
@@ -28,6 +28,7 @@
 #include <mutex>
 #include <string>
 #include <string_view>
+#include <thread>
 #include <vector>
 
 constexpr auto STATES_VD_INDEX_NAME_PREFIX {"wazuh-states-vulnerabilities"};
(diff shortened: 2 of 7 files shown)


MiguelazoDS:
Note: the failing `style` check (RemotedModule) is a pre-existing clang-format violation in `adminServer_test.cpp:101-102` from #38732 (@jotacarma90), already on `5.0.0` — not introduced by this PR, which never touches `remoted_module`.

MiguelazoDS:
Note: the failing `style` check (RemotedModule) seems to be a pre-existing clang-format violation in `adminServer_test.cpp:101-102` already on `5.0.0` not introduced by this PR, which never touches `remoted_module`. The failing `macos` check is consistently failing across multiple PRs https://github.com/wazuh/wazuh/actions/workflows/5_builderpackage_agent-macos.yml

jotacarma90:
This also covers #38858, which is the same shutdown use-after-free without the crash. On the beta5 AMIs (AMD64 and ARM64), stopping the manager mid initial load left a slice thread inside `processMessage()` after `~TDatabaseFeedManager` had already released `m_activeBatch`, so the commit parsed freed heap and logged `Error committing batch: Corruption: bad WriteBatch Delete` (AMD64) / `unknown WriteBatch tag` (ARM64) instead of faulting. Log timeline and the full chain are in https://github.com/wazuh/wazuh/issues/38858#issuecomment-5569402266.

Your `teardown()` closes that path with a hard join, not a best-effort wait: the slice threads are created and joined inside `fetchWithPit` on the scheduler thread you join, and `removeEndpoint()`'s `unique_lock` drains the on-demand one. So the 3s bounded-wait trade-off documented above does not apply to the feed-update path. Worth adding `Closes #38858` to the description so that issue closes with the merge and QA can trace it in the next stage.

One nit: `m_contentRegistration` is still declared fourth, so it is still destroyed after `m_activeBatch`, the decoder and the caches. `teardown()` makes that unreachable in production, but since the comment there says you would rather not depend on the member declaration order, moving that declaration last would make the destructor safe on its own for any path that destroys the object without calling `teardown()` first.

MiguelazoDS:
The commit that added  the comment to prevent reordering the class members declaration was this https://github.com/wazuh/wazuh/pull/20397 to test a gracefull shutdown (similar to what we are trying to fix here) 

The small fix suggested in the comment above does not seem to break the legacy test or code, so I'll apply it.

jotacarma90 (approved):

PR #38841, opened 2026-09-03
Resolve Coverity findings for agent components (5.0.0-beta-5)

### LINKED ISSUE #38829: Analyze and classify Coverity findings detected for 5.0.0 Beta 5 - Agent
# Description

The Coverity analysis executed as part of the [5.0.0 Beta 5 validation](https://github.com/wazuh/wazuh/issues/38823) detected seven new findings in the Wazuh agent.

These findings must be reviewed and fixed.

# Findings

| CID | Type | Impact | Category | File | Function |
|---|---|---|---|---|---|
| 563337 | Data race condition | Medium | Concurrent data access violations | `/src/syscheckd/src/syscheck.c` | `fim_initialize` |
| 563336 | Unchecked return value | Medium | Error handling issues | `/src/wazuh_modules/sca/sca_impl/src/sca_impl.cpp` | `getLastIntegrityCheckTime` |
| 563335 | Double unlock | Medium | API usage errors | `/src/wazuh_modules/agent_info/agent_info_impl/src/agent_info_impl.cpp` | `start` |
| 563334 | Waiting while holding a lock | Medium | Program hangs | `/src/syscheckd/src/run_check.c` | `fim_sync_lock_scan_mutex` |
| 563333 | Unchecked return value from library | Medium | Error handling issues | `/src/shared/src/file_op.c` | `w_openat_nofollow_vetted` |
| 562618 | Unchecked return value from library | Medium | Error handling issues | `/src/client-agent/https_client/src/curlHandle.cpp` | `perform` |
| 563338 | Insecure temporary file | Low | Security best practices violations | `/src/wazuh_modules/src/agent_upgrade/agent/wm_agent_upgrade_com.c` | `_uncompress` |

## Summary

| Impact | Findings |
|---|---:|
| Medium | 6 |
| Low | 1 |
| **Total** | **7** |

*CID 562618 accounts for 2 occurrences, so Coverity reports 8 newly detected defects in total even though only 7 unique CIDs need to be triaged.*

# Tasks

- [ ] Analyze each reported finding.
- [ ] Implement the required fixes.
- [ ] Add or update tests when applicable.
- [ ] Verify that the changes do not introduce regressions.
- [ ] Run Coverity again and confirm that the findings have been eliminated.
- [ ] Link the corresponding pull request to this issue.
- [ ] Document the resolution of each finding.

# Acceptance criteria

- All seven findings have been addressed.
- The required fixes have been implemented and reviewed.
- Existing functionality remains unaffected.
- Tests have been added or updated when applicable.
- A new Coverity analysis confirms that the findings are no longer reported.
- The pull request containing the fixes is linked to this issue.

Miguevrgo:
## Description

This PR addresses the 7 Coverity findings (and associated simple clang-tidy warnings) reported on the Wazuh agent components for 5.0.0-beta-5.

Closes #38829

## Proposed Changes

- **syscheckd (`fim_initialize`) - CID 563337**: Guard access to `*synced_docs_ptr` under `synced_docs_mutex` using a local variable for counting and threshold checks.
- **sca (`getLastIntegrityCheckTime`) - CID 563336**: Check the return value of `getMetadataValue` and log a warning if it fails.
- **agent_info (`AgentInfoImpl::start`) - CID 563335**: Eliminate double unlock by checking `isShutdownInProgress()` prior to locking, restrict `std::unique_lock` scope to `m_cv.wait_for` calls, and pass `shouldContinue` by const reference (`clang-tidy: performance-unnecessary-value-param`).
- **syscheckd (`fim_sync_lock_scan_mutex`) - CID 563334**: Release any held mutexes before sleeping if a subsequent mutex `trylock` fails, avoiding holding locks during sleep. Also use explicit `!= 0` comparison for `strcmp` (`clang-tidy: bugprone-suspicious-string-compare`).
- **shared (`w_openat_nofollow_vetted`) - CID 563333**: Check return values of both `fcntl(fd, F_GETFL)` and `fcntl(fd, F_SETFL, ...)`.
- **https_client (`curlHandle.cpp`) - CID 562618**: Check return values of `curl_easy_setopt` for `CURLOPT_ERRORBUFFER` and `curl_easy_getinfo` in `perform()`, `responseCode()`, and `localIp()`.
- **agent_upgrade (`_uncompress`) - CID 563338**: Apply and restore `umask(0022)` when creating the temporary file with `mkstemp`/`_mktemp_s`, and clean up assignment expressions within `if` conditions.
- **agent_metadata (`SharedMemoryProvider` ctor) - CID 563427 / CID 563428**: Annotated both `CTOR_DTOR_LEAK` reports as intentional with `// coverity[ctor_dtor_leak]`. The read-only `open()` fallback is closed on every path out of the constructor, and the `mmap()` is deliberately never unmapped (#38766) with a `static_assert` pinning the class trivially destructible, so a destructor would tear the segment down before agentd's shutdown drain has read it.

### Results and Evidence

All modified C/C++ files compile cleanly without warnings or errors. Clang-tidy checks over modified sections pass cleanly (exit code 0).

  - [x] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [x] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [x] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

### Artifacts Affected

- `wazuh-agent` (syscheckd, sca, agent-info, https_client, agent_upgrade, shared)

### Configuration Changes

N/A

### Tests Introduced

N/A

## Review Checklist

- [ ] Code changes reviewed


Miguevrgo:
## Coverity run after the fix
https://github.com/wazuh/wazuh/actions/runs/33738555516

src/wazuh_modules/agent_info/agent_info_impl/src/agent_info_impl.cpp
```diff
@@ -208,7 +206,6 @@ void AgentInfoImpl::start(int interval, int integrityInterval, std::function<boo
     // in that case, and the injected predicate is the only way to observe it.
     if (isShutdownInProgress())
     {
-        lock.unlock();
         m_logFunction(LOG_INFO, "AgentInfo start aborted: shutdown already in progress.");
         return;
     }
```
MarcelKemp, line 211: **[S2 · blocking · introduced]** Removing the `std::unique_lock<std::mutex>` from start()'s prologue makes this shutdown check and the `m_spSyncProtocol->reset()` on line 216 non-atomic with respect to `stop()`, so a stop that lands in between is silently undone. `stop()` sets `m_stopped` and then calls `m_spSyncProtocol->stop()` (`m_stopRequested.store(true)`, agent_sync_protocol.cpp:1374); `reset()` then does `m_stopRequested.store(false)` (line 1388), and since `stop()` early-returns once `m_stopped` is set the flag is never restored — the do-loop still runs its mandatory iteration and the in-flight `synchronizeModule()` is no longer interruptible via `shouldStop()`, so shutdown waits for the manager-response timeout instead of aborting. The old code held `m_mutex` across both points and `stop()` needs that mutex to set `m_stopped`, so the ordering used to be guaranteed.

Fix for lines 213-217 (outside the diff, so not a suggestion block):

```cpp
    // Reset sync protocol stop flag to allow restarting operations. Under
    // m_syncProtocolMutex and re-checked: stop() sets m_stopped and then takes this same
    // mutex to call m_spSyncProtocol->stop(), so without both the reset can land after
    // that stop() and leave the protocol permanently uninterruptible.
    {
        std::lock_guard<std::mutex> protocolLock(m_syncProtocolMutex);

        if (m_spSyncProtocol && !isShutdownInProgress())
        {
            m_spSyncProtocol->reset();
        }
    }
```

MarcelKemp (changes requested):

src/wazuh_modules/src/agent_upgrade/agent/wm_agent_upgrade_com.c
```diff
@@ -320,11 +320,13 @@ STATIC int _uncompress(const char * source, const char *package, char dest[PATH_
     // unique name, leaving a predictable path a symlink could be pre-planted at before wfopen() opened
     // it. Expand it now and, on POSIX, reuse the descriptor mkstemp() already vetted instead of a second,
     // name-based open that would reintroduce the same race.
+    mode_t old_mask = umask(0022);
```
MarcelKemp: **[S3 · non-blocking · introduced]** `umask(0022)` is more permissive than wazuh-modulesd's own mask (`umask(0007)`, set in `wazuh_modules/src/main.c:292`) and umask is process-global and not thread-safe, so any file another modulesd thread creates during the mkstemp/fchmod/decompress window is created other-readable. On this path the mask buys nothing anyway: `mkstemp()` already creates the file 0600 and the very next statement is an explicit `fchmod(fd, 0640)`. If the call has to stay to satisfy SECURE_TEMP, make it restrictive — `syncIntake.cpp:116` uses `umask(0177)` for the same purpose (and `_unsign()` on line 235 has the same issue, worth changing together).

```suggestion
    mode_t old_mask = umask(0077);
```

src/wazuh_modules/sca/sca_impl/src/sca_impl.cpp
```diff
@@ -1635,7 +1635,12 @@ bool SecurityConfigurationAssessment::updateMetadataValue(const std::string& key
 int64_t SecurityConfigurationAssessment::getLastIntegrityCheckTime()
 {
     int64_t timestamp = 0;
-    getMetadataValue(SCA_LAST_INTEGRITY_CHECK_METADATA_KEY, timestamp);
+
+    if (!getMetadataValue(SCA_LAST_INTEGRITY_CHECK_METADATA_KEY, timestamp))
+    {
+        LoggingHelper::getInstance().log(LOG_WARNING, "Failed to get last integrity check time from metadata");
+    }
```
MarcelKemp: **[S3 · non-blocking · introduced]** The return value is checked now, but the failure is still reported as `timestamp == 0`, and `integrityIntervalElapsed()` reads 0 as "first run": it logs "deferring check", calls `updateLastIntegrityCheckTime(currentTime)` and pushes the integrity check out by a full interval, so a transient DB error reschedules it instead of retrying. `getMetadataValue()` returns false only on a real failure (null `m_dBSync`, or a DBSync exception on a busy database) — a missing row returns true with 0 — and it already logs `LOG_ERROR` on both of those paths, so the added WARNING restates it at a lower level. This is exactly the failure mode `checkAgentIdentity()` documents and guards against on line 1511 ("The read itself failed -- a busy database, not an answer").

```suggestion
    if (!getMetadataValue(SCA_LAST_INTEGRITY_CHECK_METADATA_KEY, timestamp))
    {
        // The read failed -- a busy database, not an answer. Returning 0 would make
        // integrityIntervalElapsed() treat it as a first run and overwrite the stored
        // timestamp; -1 lets the caller skip this cycle instead.
        return -1;
    }
```

Then in `integrityIntervalElapsed()`, right after `int64_t lastCheck = getLastIntegrityCheckTime();`, add `if (lastCheck < 0) { return false; }`.

src/syscheckd/src/run_check.c
```diff
@@ -249,49 +249,44 @@ bool fim_shutdown_process_on() {
...
 
-    while (pthread_mutex_trylock(&syscheck.fim_realtime_mutex) != 0) {
-        if (!fim_sync_module_running || fim_shutdown_process_on()) {
+        if (pthread_mutex_trylock(&syscheck.fim_realtime_mutex) != 0) {
             w_mutex_unlock(&syscheck.fim_scan_mutex);
-            mdebug2("Stop in progress: skipping the FIM integrity validation process.");
-            return false;
+            sleep(1);
+            continue;
         }
```
MarcelKemp: **[S3 · non-blocking · introduced]** Releasing `fim_scan_mutex` before sleeping is the right fix for the hold-and-wait Coverity flagged, but it also drops the forward-progress guarantee the old shape had: on a busy agent, where the scheduled scan takes `fim_scan_mutex` and realtime holds `fim_realtime_mutex` in overlapping windows, this loop can take and drop `fim_scan_mutex` once a second indefinitely without ever holding both. Nothing records that — the only `mdebug2` in the loop is the shutdown one — so FIM integrity validation would silently never run, and no test exercises the rewritten primitive. Make the starvation observable:

```c
static bool fim_sync_lock_scan_mutex(void) {
    unsigned int attempts = 0;

    while (true) {
        if (!fim_sync_module_running || fim_shutdown_process_on()) {
            mdebug2("Stop in progress: skipping the FIM integrity validation process.");
            return false;
        }

        if (attempts != 0 && attempts % 60 == 0) {
            mdebug1("FIM integrity validation still waiting for the scan mutexes after %u seconds.", attempts);
        }

        attempts++;
        /* ... rest of the loop unchanged ... */
```

src/shared_modules/agent_metadata/src/metadata_provider.cpp
```diff
@@ -301,7 +301,10 @@ namespace
 
                 if (m_shm_fd == -1 && errno == EACCES)
                 {
-                    // Permission denied for write, try read-only
+                    // Permission denied for write, try read-only. Not a leak: every path out of
+                    // this constructor closes m_shm_fd, including the success path right after
+                    // mmap(). CTOR_DTOR_LEAK only fires because the class has no destructor.
+                    // coverity[ctor_dtor_leak]
                     m_shm_fd = open(SHM_PATH, O_RDONLY | O_CLOEXEC);
```
MarcelKemp: **[S3 · non-blocking · introduced]** This annotation suppresses a report at the assignment site while the comment directly above it proves the descriptor is closed on every exit (lines 325, 348, 358) — it silences a false positive instead of removing it. If a later edit adds a return between this `open()` and the `close()` on line 358, the real fd leak will be reported on this same line and the annotation will keep it hidden; an RAII fd holder, or hoisting the open into a helper that returns the descriptor, would make the report go away by construction. Separately, the `#ifdef _WIN32` branch (lines 256-289) never releases `m_hMapFile` or its `MapViewOfFile` mapping — the same deliberate-leak invariant the POSIX `mmap()` comment documents — but carries neither that explanation nor an annotation.

src/wazuh_modules/agent_info/agent_info_impl/src/agent_info_impl.cpp
```diff
@@ -208,27 +206,34 @@ void AgentInfoImpl::start(int interval, int integrityInterval, std::function<boo
...
-    if (m_spSyncProtocol)
+    // Reset sync protocol stop flag to allow restarting operations. Under
+    // m_syncProtocolMutex and re-checked: stop() sets m_stopped and then takes this same
+    // mutex to call m_spSyncProtocol->stop(), so without both the reset can land after
+    // that stop() and leave the protocol permanently uninterruptible.
     {
-        m_spSyncProtocol->reset();
+        std::lock_guard<std::mutex> protocolLock(m_syncProtocolMutex);
+
+        if (m_spSyncProtocol && !isShutdownInProgress())
```
MarcelKemp: Resolved. The prior shutdown race here (`m_spSyncProtocol->reset()` could undo a concurrent `stop()`) is fixed: the reset now runs under `m_syncProtocolMutex` with `isShutdownInProgress()` re-checked inside the critical section, the same mutex `stop()` takes before `m_spSyncProtocol->stop()` (lines 372-377). Verified against the current `stop()` implementation — every interleaving now serializes correctly.

Darioortegaleyva (changes requested):
Just a few things:

1. `wm_agent_upgrade_com.c (_uncompress(), ~line 323)` - the `umask(0022)/restore` looks that it doesn't really do anything, mkstemp() and the fchmod(fd, 0640) right after already fix the permissions on their own. And since umask() affects the whole process, not just this thread, it can briefly mess with permissions of files other wodles are creating at the same time. Also the restore is repeated in 5 different early returns instead of one place. Can we just remove it?
2. `curlHandle.cpp (~line 356)` - if resetting CURLOPT_ERRORBUFFER fails after perform(), we just log it and move on, but the handle keeps pointing to a stack buffer that's already gone. If it fails again on the next call, libcurl could end up writing to memory that's no longer valid. Might be safer to treat that failure as fatal for the handle instead of only logging it.

Could you take a look at it?

Rest looks good.

src/wazuh_modules/sca/sca_impl/src/sca_impl.cpp
MarcelKemp: **[S3 · non-blocking · introduced]** The return value is checked now, but the failure is still reported as `timestamp == 0`, and `integrityIntervalElapsed()` reads 0 as "first run": it logs "deferring check", calls `updateLastIntegrityCheckTime(currentTime)` and pushes the integrity check out by a full interval, so a transient DB error reschedules it instead of retrying. `getMetadataValue()` returns false only on a real failure (null `m_dBSync`, or a DBSync exception on a busy database) — a missing row returns true with 0 — and it already logs `LOG_ERROR` on both of those paths, so the added WARNING restates it at a lower level. This is exactly the failure mode `checkAgentIdentity()` documents and guards against on line 1511 ("The read itself failed -- a busy database, not an answer").

```suggestion
    if (!getMetadataValue(SCA_LAST_INTEGRITY_CHECK_METADATA_KEY, timestamp))
    {
        // The read failed -- a busy database, not an answer. Returning 0 would make
        // integrityIntervalElapsed() treat it as a first run and overwrite the stored
        // timestamp; -1 lets the caller skip this cycle instead.
        return -1;
    }
```

Then in `integrityIntervalElapsed()`, right after `int64_t lastCheck = getLastIntegrityCheckTime();`, add `if (lastCheck < 0) { return false; }`.

src/syscheckd/src/run_check.c
MarcelKemp: **[S3 · non-blocking · introduced]** Releasing `fim_scan_mutex` before sleeping is the right fix for the hold-and-wait Coverity flagged, but it also drops the forward-progress guarantee the old shape had: on a busy agent, where the scheduled scan takes `fim_scan_mutex` and realtime holds `fim_realtime_mutex` in overlapping windows, this loop can take and drop `fim_scan_mutex` once a second indefinitely without ever holding both. Nothing records that — the only `mdebug2` in the loop is the shutdown one — so FIM integrity validation would silently never run, and no test exercises the rewritten primitive.

src/shared_modules/agent_metadata/src/metadata_provider.cpp
MarcelKemp: **[S3 · non-blocking · introduced]** This annotation suppresses a report at the assignment site while the comment directly above it proves the descriptor is closed on every exit (lines 325, 348, 358) — it silences a false positive instead of removing it. If a later edit adds a return between this `open()` and the `close()` on line 358, the real fd leak will be reported on this same line and the annotation will keep it hidden. Separately, the `#ifdef _WIN32` branch (lines 256-289) never releases `m_hMapFile` or its `MapViewOfFile` mapping — the same deliberate-leak invariant the POSIX `mmap()` comment documents — but carries neither that explanation nor an annotation.

src/wazuh_modules/src/agent_upgrade/agent/wm_agent_upgrade_com.c
```diff
@@ -322,9 +320,9 @@ STATIC int _uncompress(const char * source, const char *package, char dest[PATH_
     // name-based open that would reintroduce the same race.
 #ifndef WIN32
     {
-        int fd;
+        int fd = mkstemp(dest);
```
MarcelKemp: Resolved. The `umask(0022)` call flagged before is now removed entirely from both `_uncompress()` and `_unsign()`. Verified it never had any effect: `mkstemp()` always creates the file at mode 0600 regardless of umask, and an explicit `fchmod(fd, 0640)` sets the final mode immediately after in both functions — no other file/directory creation exists on either path, so removing the umask call closes the process-global-mask-widening window with no loss of protection.

src/client-agent/https_client/src/curlHandle.cpp
```diff
@@ -329,37 +330,66 @@ namespace
...
+                if (errorBufferSet &&
+                        curl_easy_setopt(m_handle, CURLOPT_ERRORBUFFER, nullptr) != CURLE_OK)
+                {
+                    // LCOV_EXCL_START: setting CURLOPT_ERRORBUFFER to null cannot be made to fail.
+                    LOGFN_DEBUG1(handleLogFn(), "Could not reset CURLOPT_ERRORBUFFER, retiring handle");
+                    curl_easy_cleanup(m_handle);
+                    m_handle = nullptr;
+                    m_lastError = "CURLOPT_ERRORBUFFER could not be reset";
+                    return TransportStatus::OtherError;
+                    // LCOV_EXCL_STOP
```
MarcelKemp: Resolved. A failed `CURLOPT_ERRORBUFFER` reset now retires the handle (`curl_easy_cleanup` + `m_handle = nullptr`) instead of leaving a dangling stack-buffer pointer live on a reused member handle. Traced every call this class makes after `perform()` on the same object (`curlPerformer.cpp` calls `responseCode()`, `localIp()`, `curlError()` unconditionally): `m_handle == nullptr` is already a state the class tolerates (constructor leaves it null if `curl_easy_init()` fails), libcurl's setopt/getinfo null-check the handle rather than dereference it, both getinfo call sites already normalize a failed call to a safe default, `curlError()` surfaces the retirement reason to the caller, and the destructor already guards `m_handle != nullptr` before cleanup — no double-free.

**[S4 · non-blocking · introduced]** One residual, purely theoretical side effect of this exact fix: the same branch also fires (overwriting `m_lastError` and forcing `TransportStatus::OtherError`) if the reset failed after a `curl_easy_perform()` that itself returned `CURLE_OK`, discarding what would have been a successful result. The `LCOV_EXCL` comment argues this is unreachable in practice (resetting `CURLOPT_ERRORBUFFER` to null on a handle that just succeeded doesn't fail), so no action is requested — noted for the record.

Darioortegaleyva (approved):
LGTM

MarcelKemp (approved):

PR #38869, opened 2026-09-03
Fix HTTPS remote upgrade staging for agents with a prior legacy-path upgrade

### LINKED ISSUE #38833: Remote upgrade fails on agents previously upgraded via the legacy path
|---|---|---|---|---|
| 5.0.0-beta5 | wazuh-agentd (https_client) | Agent | Packages | AlmaLinux 9 |

## Description

Any agent that was remotely upgraded to 5.x through the **legacy** delivery path (`remoted` push, used for agents below v5.0.0) is left with a root-owned WPK in `var/incoming`. Every later remote upgrade of that agent — the 5.x-to-5.x HTTPS path — fails permanently while staging the WPK, because `wazuh-agentd` runs as `wazuh` and cannot overwrite that file. The failure is silent from the manager's point of view: the HTTPS upgrade path is fire-and-forget, so nothing is reported back and `tasks.db` shows no failure.

## Steps to reproduce

1. Install a v4.14.7 agent and upgrade it remotely to v5.0.0 (legacy path, works).
2. With the agent now on v5.0.0, trigger a second remote upgrade with the same WPK: `/var/wazuh-manager/bin/agent_upgrade -a 003 -f wazuh_agent_v5.0.0_linux_x86_64.rpm.wpk`.
3. The manager reports the task as created; the agent aborts the upgrade.

## Agent log

```console
2026/09/02 20:20:10 wazuh-agentd: ERROR: At w_copy_file(): Couldn't open file 'var/incoming/wazuh_agent_v5.0.0_linux_x86_64.rpm.wpk'
2026/09/02 20:20:10 wazuh-agentd: ERROR: https_client: remote_upgrade task b96c8e79-b0c5-07c7-f6e6-ae8629e07fc4: could not stage the WPK at 'var/incoming/wazuh_agent_v5.0.0_linux_x86_64.rpm.wpk'; aborting.
```

```console
[root@vm-alma9 vagrant]# ls -ld /var/ossec/var/incoming
drwxrwx---. 2 root wazuh 101 Sep  1 11:16 /var/ossec/var/incoming
[root@vm-alma9 vagrant]# ls -l /var/ossec/var/incoming/
-rw-r--r--. 1 root wazuh 10963833 Sep  2 20:07 wazuh_agent_v4.14.7_linux_x86_64.rpm.wpk
-rw-r--r--. 1 root wazuh 12015312 Sep  2 20:12 wazuh_agent_v5.0.0_linux_x86_64.rpm.wpk
[root@vm-alma9 vagrant]# ps -o user,group,cmd -C wazuh-agentd
USER     GROUP    CMD
wazuh    wazuh    /var/ossec/bin/wazuh-agentd
```

The `20:12` file is the one written by the 4.x `wazuh-execd` (running as root) during the legacy delivery; the failure at `20:20` is the HTTPS path trying to reuse the same name.

## Root cause

The 5.x staging step copies the downloaded WPK into `var/incoming/<wpk_file>` with `w_copy_file()`, which opens the destination with `fopen(dst, "wb")` and does not unlink it first: 

https://github.com/wazuh/wazuh/blob/4376b22bd7d895bf52ffd0e16cc888c0d6efbaa5/src/client-agent/src/https_client_bridge.c#L891-L902

and 

https://github.com/wazuh/wazuh/blob/4376b22bd7d895bf52ffd0e16cc888c0d6efbaa5/src/shared/src/file_op.c#L2110-L2121

`var/incoming` is `0770 root:wazuh`, so `wazuh-agentd` (which drops privileges to `wazuh:wazuh`) can create new entries there but cannot truncate a pre-existing `0644 root:*` file. The legacy path writes exactly such a file, and the agent-side upgrade module never removes the WPK from `var/incoming` after installing (it only cleans `UPGRADE_DIR` and its own temporaries): 

https://github.com/wazuh/wazuh/blob/4376b22bd7d895bf52ffd0e16cc888c0d6efbaa5/src/wazuh_modules/src/agent_upgrade/agent/wm_agent_upgrade_com.c#L153-L166

## Proposed fix

Unlink the destination before staging, and use `w_fopen_nofollow(INCOMING_DIR, wpk_file, "wb")` instead of `w_copy_file()` for the destination side. The second part also restores, for this new path, the symlink hardening `var/incoming` already has in the legacy path (https://github.com/wazuh/wazuh/blob/5.0.0/src/os_execd/src/wcom.c#L172, #38200) — as written today, a symlink planted at `var/incoming/<wpk_file>` would be followed.

Removing stale WPKs from `var/incoming` after a successful install would be a reasonable follow-up, but it is not enough on its own: an agent already in this state stays broken.

## Workaround

Delete the leftover WPK on the agent and retry: `rm -f /var/ossec/var/incoming/<wpk_file>`

rjcausarano:
## Description

On a 5.x agent, a remote upgrade dispatched over the new HTTPS transport permanently fails to
stage the WPK if that agent was ever previously upgraded via the **legacy** delivery path (used
for agents below v5.0.0, run by `remoted`/`wazuh-execd`).

Root cause: the legacy path is delivered by `wazuh-execd`, which runs as **root**, and writes the
WPK into `var/incoming/<wpk_file>` as `root`-owned. The new HTTPS path is handled by
`wazuh-agentd`, which runs as the unprivileged `wazuh` user. When a later upgrade reuses the same
WPK filename, `bridge_on_remote_upgrade_ready()` staged the download with `w_copy_file()`, which
opens the destination via a plain `fopen(dst, "wb")` — this never `unlink()`s the target first, and
opening an *existing* file for writing requires write permission on that file itself, which the
`wazuh` user does not have on a `root`-owned file. The failure is silent on the manager's side,
since the HTTPS upgrade path is fire-and-forget (nothing is reported back, and `tasks.db` shows no
failure) — only the agent's own log records the staging error. Once an agent hits this, every
subsequent HTTPS-path upgrade with the same WPK filename fails permanently, since nothing ever
removes the stale file from `var/incoming` on its own.

Reproduced on a 4.14.7 → 5.0.0 (legacy) → 5.0.0 (HTTPS, same WPK filename) upgrade sequence:
```
wazuh-agentd: ERROR: At w_copy_file(): Couldn't open file 'var/incoming/wazuh-agent_5.0.0-latest_amd64.deb.wpk'
wazuh-agentd: ERROR: https_client: remote_upgrade task <id>: could not stage the WPK at 'var/incoming/wazuh-agent_5.0.0-latest_amd64.deb.wpk'; aborting.
```
with the staged file confirmed `root`-owned, `0644`, unwritable by the `wazuh` group.

Closes #38833

## Proposed Changes

- `bridge_on_remote_upgrade_ready()` (`src/client-agent/src/https_client_bridge.c`) now `unlink()`s
  the destination path before staging, ignoring `ENOENT` (no stale file present is the common
  case). Removing the old entry first means re-creating the file only needs write permission on
  `var/incoming` itself (`0770 root:wazuh`, which the `wazuh` user's group already has), not on
  whatever pre-existing file happened to be sitting at that path.
- Replaced `w_copy_file()` with a manual copy: `wfopen()` for the source, and
  `w_fopen_nofollow(INCOMING_DIR, wpk_file, "wb")` for the destination. This also closes a symlink
  gap: as written before this change, a symlink planted at `var/incoming/<wpk_file>` between
  requests would have been followed and its target overwritten with the downloaded WPK contents.
  This mirrors the equivalent hardening already applied to the legacy path's `WCOM uncompress()`
  (`src/os_execd/src/wcom.c`, #38200) — a distinct `ELOOP`-specific log message is emitted when a
  symlink is refused, rather than the generic staging-failure message.
- On a mid-copy write failure, the partially-written file is `unlink()`ed before the task is marked
  failed, so a failed attempt does not leave a corrupt file behind to break the *next* attempt too.

Not included in this PR: the issue's suggested follow-up of proactively removing a successfully
installed WPK from `var/incoming` after install. That's independently reasonable but was explicitly
called out in the issue as insufficient on its own (an agent already in the broken state needs the
unlink-before-staging fix regardless), so it's left as a separate potential improvement rather than
folded into this fix.

### Results and Evidence

The bug was originally reproduced end-to-end on a manager + agent VM pair (4.14.7 → 5.0.0 legacy
upgrade, followed by a same-filename 5.0.0 → 5.0.0 HTTPS-path upgrade), matching the issue's repro
steps and log output exactly (see above).

**The fix has since been re-verified against the built agent, on the same VM pair, and the same
previously-failing sequence now succeeds end-to-end.** CI on the current HEAD of this branch
(`67d3d4f49d`) passes `Unit-Tests (agent)` and `Unit-Tests (manager)` cleanly. `Unit-Tests
(winagent)` is still red, but on three tests (`test_hmac_op`, `test_sha512_op`, `test_win_whodata`)
that are unrelated to this change and unrelated to this PR entirely — confirmed via `git log`
across this PR's full commit range, none of those three test files (or the production code they
exercise) are touched anywhere in this branch's history. Their failure signature (e.g.
`test_sha512_file` hashing to the well-known empty-string SHA-512) points to a pre-existing,
deterministic Wine/CI-environment issue reading a file back through a different API than it was
written with, reproduced identically across two separate commits on this branch.

Manual re-reproduction, same VM pair (`wazuh-virtual-manager-5.0.0` at `192.168.56.18`,
`debian12-agent`): upgraded the agent from 4.14.7 → 5.0.0 (legacy path, `wazuh-execd`), then
triggered a second, HTTPS-path remote upgrade **5.0.0 → 5.0.0 reusing the same WPK build**
(`wazuh-agent_5.0.0-0_amd64_67d3d4f.deb.wpk`, confirmed via the manager's log for the first,
legacy-path delivery of that exact filename to this agent) — the precise precondition this issue
describes. The second upgrade completed cleanly:

```
wazuh-agentd: INFO: https_client: remote_upgrade task d533f3eb-5993-bd45-e49d-72f5a34c3780 dispatched to the upgrade module (installer running; the agent may restart shortly). No /control response is sent.
[... clean shutdown of all daemons ...]
wazuh-agentd: INFO: Started (pid: 6393).
wazuh-agentd: INFO: https_client: starting.
wazuh-agentd:https-client: INFO: Starting https_client (server=192.168.56.18:1517, agent=009, body compression enabled).
[... execd, rootcheck, syscheckd, logcollector, modulesd all start cleanly; FIM scan, SCA scan, and Syscollector evaluation all complete with no errors ...]
```

No `w_copy_file()`/staging error, no permission error on `var/incoming`, and no repeat of the
original bug despite reusing the exact WPK filename that previously triggered it permanently. The
manager's inventory sync shows the agent checking back in and reporting its updated configuration
immediately after the restart (`19:28:14`–`19:30:45`), confirming the agent came back up healthy
and reconnected — the closest available manager-side corroboration, since this upgrade path is
deliberately fire-and-forget from the manager's perspective (see Description above).

Note: reaching this reproduction also required working around an unrelated gap in a separate,
already-merged PR (#38786) — that agent's `ossec.conf` was still in the unmigrated 4.x `<client>`-only
shape, which trips a fail-closed check in `pkg_installer.sh`'s `pin_ca()` that only recognizes
`<agent>` blocks. That's tracked separately and isn't part of this fix; it was worked around here by
pre-configuring `<agent><ssl>` by hand before running the upgrade, purely to get to a state where
*this* PR's fix could actually be exercised.

  - [x] Linux — CI `Unit-Tests (agent)` and `Unit-Tests (manager)` both build and pass cleanly on
        HEAD (`67d3d4f49d`); a full local `-Wall` build was not separately re-verified beyond CI.
  - [ ] Windows — `Unit-Tests (winagent)` build succeeds; the 3 test failures on this platform are
        confirmed unrelated to this change (see Results and Evidence).
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests _(Wazuh v4.x)_: N/A — the affected code path (`https_client_bridge.c`) doesn't
  exist on 4.x; no decoder/rule involvement.

- Engine _(Wazuh v5.x and above)_: N/A — no engine code touched.
  - [ ] Test run in parallel

- Wazuh server API/Framework: N/A — this is an agent-only change; no manager, API, or framework
  code is touched.

### Artifacts Affected

`wazuh-agentd` (agent binary, all platforms) — the HTTPS remote-upgrade WPK staging path in
`https_client_bridge.c`. Exercised end-to-end on Linux (both in CI and via manual VM reproduction);
the code path itself is shared across all agent platforms (Windows uses a different path separator
for `dest_path`, but the same `unlink`/`wfopen`/`w_fopen_nofollow` sequence).

### Configuration Changes

N/A — no new configuration parameters, and no change to any existing default or config-reading
behavior.

### Tests Introduced

Updated `src/unit_tests/client-agent/test_https_client_bridge.c`: replaced the single now-invalid
mock-based staging test (which asserted on the removed `w_copy_file()` call) with five tests
covering the new sequence — stale-file removal failure, source-open failure, destination-open
failure (generic), symlink rejection (`ELOOP`), and a mid-copy write failure with cleanup — using
this file's existing wrap/mock conventions. All five, plus the rest of this test binary's 96 tests,
now pass (confirmed both via CI and a local build/run).

Supporting test infrastructure: added an `__wrap_ferror`/`expect_ferror` pair to
`src/unit_tests/wrappers/libc/stdio_wrappers.c`/`.h` (mirroring the existing `__wrap_fclose`
pattern), since nothing previously needed to mock `ferror()`. Because that file's wrappers are
linked into virtually every unit test binary via `libwazuh_test.a`, and most of those binaries only
pass `-Wl,--wrap` for the handful of functions they actually use, the new `__real_ferror` extern
needed to be declared `__attribute__((weak))` (matching the existing `__real_fdopen` precedent in
the same file) rather than requiring every other test target to add an unrelated `--wrap` flag.
Also fixed: `test_mode` is now set around the four new WPK-staging tests specifically (they're the
only tests in this file that need `wfopen`/`fclose`/`fread`/`fwrite`/`ferror` actually mocked), and
`__wrap_unlink` gained an `expect_unlink_errno()` helper so tests that need a specific `errno` value
after a mocked `unlink()` call get it deterministically rather than relying on a raw assignment
racing the mock machinery.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

### DIFF
--- a/src/client-agent/src/https_client_bridge.c
+++ b/src/client-agent/src/https_client_bridge.c
@@ -886,9 +886,67 @@ static void bridge_on_remote_upgrade_ready(const char *task_id, const char *wpk_
     snprintf(dest_path, sizeof(dest_path), "%s\\%s", INCOMING_DIR, wpk_file);
 #endif
 
-    if (w_copy_file(wpk_path, dest_path, 'b', NULL, 0) < 0) {
+    /* A prior legacy-path upgrade (delivered by wazuh-execd, running as root) can leave a
+     * same-named WPK behind in INCOMING_DIR, owned by root and not writable by the unprivileged
+     * user this process runs as. Unlinking first turns the write into a fresh file creation --
+     * which only needs write permission on the directory, not on whatever pre-existing file may
+     * already be sitting at this path -- instead of an open-for-write on a file this process may
+     * not own (#38833). */
+    if (unlink(dest_path) < 0 && errno != ENOENT) {
+        merror("https_client: remote_upgrade task %s: could not remove stale WPK at '%s': %s (%d); aborting.",
+               task_id, dest_path, strerror(errno), errno);
+        w_agentd_state_update(INCREMENT_TASK_FAILED, NULL);
+        return;
+    }
+
+    FILE *fsrc = wfopen(wpk_path, "rb");
+    if (!fsrc) {
+        merror("https_client: remote_upgrade task %s: could not open downloaded WPK '%s'; aborting.",
+               task_id, wpk_path);
+        w_agentd_state_update(INCREMENT_TASK_FAILED, NULL);
+        return;
+    }
+
+    /* Not w_copy_file()/fopen(dst, "wb"): a symlink left under INCOMING_DIR would be followed and
+     * its target overwritten with the WPK contents. Mirrors the hardening already applied to the
+     * legacy path's WCOM uncompress() (os_execd/src/wcom.c, #38200). */
+    FILE *fdst = w_fopen_nofollow(INCOMING_DIR, wpk_file, "wb");
+    if (!fdst) {
+        const int open_errno = errno;
+        fclose(fsrc);
+        if (open_errno == ELOOP) {
+            merror("https_client: remote_upgrade task %s: refused to stage the WPK at '%s': the path is a symbolic link; aborting.",
+                   task_id, dest_path);
+        } else {
+            merror("https_client: remote_upgrade task %s: could not stage the WPK at '%s'; aborting.",
+                   task_id, dest_path);
+        }
+        w_agentd_state_update(INCREMENT_TASK_FAILED, NULL);
+        return;
+    }
+
+    char copy_buffer[4096];
+    size_t nread;
+    bool copy_ok = true;
+
+    while (nread = fread(copy_buffer, 1, sizeof(copy_buffer), fsrc), nread > 0) {
+        if (fwrite(copy_buffer, 1, nread, fdst) != nread) {
+            copy_ok = false;
+            break;
+        }
+    }
+
+    if (ferror(fsrc)) {
+        copy_ok = false;
+    }
+
+    fclose(fsrc);
+    fclose(fdst);
+
+    if (!copy_ok) {
         merror("https_client: remote_upgrade task %s: could not stage the WPK at '%s'; aborting.",
                task_id, dest_path);
+        unlink(dest_path);
         w_agentd_state_update(INCREMENT_TASK_FAILED, NULL);
         return;
     }
--- a/src/client-agent/src/https_client_bridge.c
+++ b/src/client-agent/src/https_client_bridge.c
@@ -872,7 +872,12 @@ static void bridge_on_remote_upgrade_ready(const char *task_id, const char *wpk_
         return;
     }
 
-    if (w_ref_parent_folder(wpk_file)) {
+    /* w_ref_parent_folder() alone only rejects ".." components: a value with a subdirectory
+     * separator (e.g. "sub/file.wpk") would pass it, reach the unlink() below, and only get caught
+     * afterwards by w_fopen_nofollow()'s internal w_is_bare_filename() check on the destination
+     * open -- by which point the unlink() has already run against that subdirectory path. Requiring
+     * a bare filename here closes that gap before the unlink(). */
+    if (!w_is_bare_filename(wpk_file)) {
         merror("https_client: remote_upgrade task %s: wpk_file '%s' is not a safe filename; aborting.",
                task_id, wpk_file);
         w_agentd_state_update(INCREMENT_TASK_FAILED, NULL);
@@ -941,7 +946,12 @@ static void bridge_on_remote_upgrade_ready(const char *task_id, const char *wpk_
     }
 
     fclose(fsrc);
-    fclose(fdst);
+
+    /* A write failure that only surfaces at flush time (e.g. disk full right on close) must not
+     * be swallowed: fclose()'s return value is the only place that shows up. */
+    if (fclose(fdst) != 0) {
+        copy_ok = false;
+    }
 
     if (!copy_ok) {
         merror("https_client: remote_upgrade task %s: could not stage the WPK at '%s'; aborting.",
--- a/src/unit_tests/client-agent/CMakeLists.txt
+++ b/src/unit_tests/client-agent/CMakeLists.txt
@@ -127,7 +127,8 @@ list(APPEND client-agent_flags "-Wl,--wrap,wurl_free_response -Wl,--wrap,wurl_ht
                                 -Wl,--wrap,w_agentd_state_update -Wl,--wrap,getDefine_Int \
                                 -Wl,--wrap,getDefine_Int_default \
                                 -Wl,--wrap,w_copy_file -Wl,--wrap,UnmergeFiles -Wl,--wrap,cldir_ex_ignore \
-                                -Wl,--wrap,wfopen \
+                                -Wl,--wrap,wfopen -Wl,--wrap,w_fopen_nofollow -Wl,--wrap,unlink \
+                                -Wl,--wrap,fread -Wl,--wrap,fwrite -Wl,--wrap,fclose -Wl,--wrap,ferror \
                                 -Wl,--wrap,verifyRemoteConf -Wl,--wrap,reloadAgent \
                                 -Wl,--wrap,w_agentd_populate_metadata \
                                 -Wl,--wrap,startup_gate_release_from_https_apply \
--- a/src/unit_tests/client-agent/CMakeLists.txt
+++ b/src/unit_tests/client-agent/CMakeLists.txt
@@ -128,7 +128,7 @@ list(APPEND client-agent_flags "-Wl,--wrap,wurl_free_response -Wl,--wrap,wurl_ht
                                 -Wl,--wrap,getDefine_Int_default \
                                 -Wl,--wrap,w_copy_file -Wl,--wrap,UnmergeFiles -Wl,--wrap,cldir_ex_ignore \
                                 -Wl,--wrap,wfopen -Wl,--wrap,w_fopen_nofollow -Wl,--wrap,unlink \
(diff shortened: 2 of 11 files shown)


cborla:
`wpk_file` reaches `unlink(dest_path)` (line ~895) validated only by `w_ref_parent_folder()` (rejects `..`), while `w_fopen_nofollow()` two lines below enforces `w_is_bare_filename()` (rejects any `/`) on that same value before opening the destination. Since `wpk_file` comes straight from the manager's task payload with no other validation upstream, a value containing a subdirectory component reaches the `unlink()` before that stricter check ever runs.

cborla:
Finding: `fclose(fdst)` at `https_client_bridge.c:949` doesn't check its return value. A write failure that only surfaces at flush time (e.g. disk full right on close) would leave `copy_ok = true` and stage a truncated WPK as if it succeeded.

Possible fix: `if (fclose(fdst) != 0) copy_ok = false;` before the `copy_ok` check.

cborla (dismissed):
LGTM
- [x] Code changes reviewed

cborla (approved):
LGTM
- [x] Code changes reviewed

PR #38874, opened 2026-09-04
fix(api): key the request-rate limiter per client address

MiguelazoDS:
## Description

The manager REST API's request-rate limiter (`api/api/middlewares.py`) is a single unkeyed, module-level global counter shared by every caller. A read-only account, an invalid token, or no credentials at all can each exhaust it and lock out every other caller — including the administrator and the dashboard's own service account — for the rest of the one-minute window.

Closes wazuh/internal-devel-requests#6027

## Proposed Changes

- Keyed the rate limiter by client address instead of one shared global counter, so one caller can no longer exhaust the budget for every other caller.
- Split each address's budget into two independent sub-buckets (unauthenticated / authenticated), so failed-auth or unauthenticated traffic can never spend an authenticated caller's budget — mirrors the reserve-then-settle shape already used by `check_blocked_ip`/`settle_login_attempt`.
- Added `SettleRateLimitMiddleware`, registered immediately after connexion's `SecurityMiddleware`, so the authenticated bucket is only charged once authentication has actually succeeded.
- Guarded the settle step against a window-rollover race between the two middlewares (found during design review) that could otherwise let one extra request slip past a freshly-rolled-over window.
- Fixed a rollover off-by-one in the old counter (301 allowed per window after the first) as a side effect of moving to atomic per-request counting.
- Added `max_unauthenticated_request_per_minute` (default 10, derived from `max_login_attempts`/`block_time`) for the new, smaller unauthenticated ceiling; it deliberately does not scale with deployment size the way `max_request_per_minute` does.
- Added WARNING-level logging on rate-limit exhaustion, naming the source address and user.
- Added error code `6005` for the new unauthenticated-ceiling rejection.

### Results and Evidence

- `python -m pytest api/api/test/test_middlewares.py -vv --cov=api.middlewares`: 68 passed, 96% coverage on `middlewares.py`, all new lines covered.
- `python -m pytest api/api -vv --ignore=test --cov=api`: 472 passed (run repeatedly for determinism).
- `python -m pytest framework/wazuh/core/tests/test_exception.py -q`: 11 passed.
- `python -m pytest framework -q --ignore=tests`: 2024 passed; 8 pre-existing failures confirmed unrelated (reproduce identically against the unmodified base commit — an `api.__path__` resolution quirk in this environment, not a regression from this change).
- The cross-window release-race regression test was confirmed load-bearing: removing the guard causes it to fail.

  - [x] Linux (N/A — pure Python change)
  - [x] Windows (N/A — pure Python change)
  - [x] MAC OS X (N/A — pure Python change)

- Memory tests for Linux
  - [x] Coverity (N/A — pure Python change)
  - [x] Valgrind (memcheck and descriptor leaks check) (N/A — pure Python change)
  - [x] AddressSanitizer (N/A — pure Python change)
- Memory tests for Windows
  - [x] Coverity (N/A — pure Python change)
  - [x] UMDH (N/A — pure Python change)
- Memory tests for macOS
  - [x] Leaks (N/A — pure Python change)
  - [x] AddressSanitizer (N/A — pure Python change)

  - [x] N/A — no decoders/rules touched

  - [x] N/A — no Engine component touched

- Wazuh server API/Framework
  - [ ] Run API Integration Tests — not run; automated unit coverage (including `freeze_time`-based window-boundary and concurrency-guard regression tests) was judged sufficient for this change. Flagging for reviewer input on whether the tavern/Docker integration suite should be run before merge.

### Artifacts Affected

- `wazuh-apid` (manager REST API process) — behavior change only, no new executables or packages.

### Configuration Changes

**New parameter:**
- `access.max_unauthenticated_request_per_minute` (integer, default `10`) — ceiling for unauthenticated/failed-auth requests per client address per minute; `0` disables only this sub-limit.

**Backward compatibility notes:**
- `max_request_per_minute`'s default (300) and meaning are unchanged; it now enforces per-address instead of globally.
- A caller that previously saw only error code `6001` for any rate-limit rejection may now see the new code `6005` for the unauthenticated-ceiling case specifically — relevant for any integration that pattern-matches strictly on `6001`.
- Dashboard users behind a single dashboard backend still share one address-keyed bucket (accepted, documented limitation — this fix targets external-caller-vs-admin collisions, not per-dashboard-user budgets).

### Tests Introduced

**New** (`api/api/test/test_middlewares.py`):
- `test_reserve_unauthenticated_request_counts_current_request`, `_window_rollover`, `_disabled`, `_keys_by_host`
- `test_settle_authenticated_request_releases_unauthenticated_charge`, `_charges_and_checks_authenticated_bucket`, `_disabled`, `_unknown_host`, `_skips_release_after_window_rollover` (the cross-window release-race regression)
- `test_two_sub_counters_prevent_unauthenticated_traffic_from_denying_authenticated_caller`
- `test_check_rate_limits_middleware_calls_reserve`, `test_settle_rate_limit_middleware_calls_settle`
- `test_access_log_warns_on_429`, `test_access_log_no_warning_for_200`

**Modified:**
- `test_check_rate_limits_middleware_ko` (rewritten for the new function)
- `mock_req` fixture gained `req.state = SimpleNamespace()` for realistic attribute semantics matching `starlette.datastructures.State`.

## Review Checklist

- [x] Code changes reviewed
- [ ] API Integration Tests run (see Manual tests section above)


api/scripts/wazuh_manager_apid.py
```diff
@@ -159,6 +159,7 @@ async def _warm_up_pools():
     # Maximum body size that the API can accept (bytes)
     if api_conf['access']['max_request_per_minute'] > 0:
         app.add_middleware(CheckRateLimitsMiddleware, MiddlewarePosition.BEFORE_SECURITY)
```
Antoniogm03, line 173: Shouldn't we create another `if` statement to check `max_unauthenticated_request_per_minute`?
`CheckRateLimitsMiddleware` is a middleware that runs before authentication, so its configuration parameter would be `max_unauthenticated_request_per_minute`.

api/api/middlewares.py
```diff
@@ -54,8 +54,8 @@
 ip_stats = dict()
 ip_block = set()
 ip_lock = asyncio.Lock()
-general_request_counter = 0
-general_current_time = None
+general_request_stats: dict = {}
```
Antoniogm03, line 65: Shouldn't we implement some mechanism to clean up old entries in the dictionary?

Antoniogm03 (changes requested):

api/scripts/wazuh_manager_apid.py
MiguelazoDS: This is related to what I also found, it seems that the unauthenticated counter was preventing an authenticated user to request, because the legit request was counted first as a non-authenticated.

api/api/middlewares.py
MiguelazoDS: Good catch, I added some test for some possible old values hold in memory.

MiguelazoDS:
## Live validation against a running manager

I deployed this branch to a running Wazuh manager and exercised the rate-limit fix with real HTTP traffic (not just the unit tests). Below is every scenario tested, what it checks, the actual result, and the exact script used so anyone can run the same test. All scripts assume a manager reachable at `https://localhost:55000` with the default `wazuh:wazuh` admin account, and the default `api.yaml` (`max_request_per_minute: 300`, `max_unauthenticated_request_per_minute: 10`) unless a test says otherwise.

| # | What it checks | Result |
|---|---|---|
| 1 | An invalid-token flood from one address does not lock out that same address's admin once they authenticate | Pass |
| 2 | A flood from a different address never affects the admin at all | Pass |
| 3 | A genuinely authenticated low-privilege account sharing the admin's address | Reproduces a known, accepted limitation (see below), not a new defect |
| 4 | The same invalid-token flood sustained over 4 bursts across 100 seconds | Pass |
| 5 | Ordinary low-volume unauthenticated traffic (a single bad login) | Pass, unaffected |
| 6 | A normal login followed by a normal authenticated call | Pass, unaffected |
| 7 | The authenticated ceiling (300 requests/minute) still rejects past its limit | Pass |
| 8 | A WARNING log line fires when a caller is rate-limited, naming the IP and user | Pass |
| 9 | Setting the authenticated limit to 0 no longer silently disables the unauthenticated limit too | Pass |
| 10 | The internal per-address tracking dictionary gets cleaned up automatically instead of growing forever | Pass |
| 11 | 500 to 8,000 distinct addresses, each independently driving toward its own 300/minute ceiling | Pass, no address ever exceeded its own cap or was affected by another; a separate, pre-existing bottleneck limits aggregate throughput regardless of N |

### 1. Invalid-token flood, then the admin retries from the same address

```bash
#!/usr/bin/env bash
set -uo pipefail
API=https://localhost:55000
LOG=/var/wazuh-manager/logs/api.log   # adjust to your install's api.log path
CREDS=wazuh:wazuh
SCRATCH=/tmp/qa-6027; mkdir -p "$SCRATCH"

TOKEN=$(curl -sk -u "$CREDS" -X POST "$API/security/user/authenticate" | python3 -c 'import sys,json; print(json.load(sys.stdin)["data"]["token"])')
curl -sk -o /dev/null -w "admin GET /agents before: status=%{http_code}\n" -H "Authorization: Bearer $TOKEN" "$API/agents?limit=1"

LINES_BEFORE=$(wc -l < "$LOG")

for i in $(seq 1 350); do
  curl -sk -o /dev/null -H "Authorization: Bearer this.is.not.a.valid.jwt.token.at.all" "$API/agents?limit=1" &
done
wait

curl -sk -o /dev/null -w "admin GET /agents after: status=%{http_code}\n" -H "Authorization: Bearer $TOKEN" "$API/agents?limit=1"

sleep 1
tail -n +$((LINES_BEFORE + 1)) "$LOG" > "$SCRATCH/invalid_token_new_lines.log"
echo "-- response codes --"
grep -oE ': [0-9]{3}$' "$SCRATCH/invalid_token_new_lines.log" | sort | uniq -c
```

Output:
```
admin GET /agents before: status=200
-- response codes --
     10 : 401
    340 : 429
admin GET /agents after: status=200
```

The flood is still correctly capped (10 real failed-auth attempts, then rejected), and the admin's own retry still succeeds.

### 2. Same flood, but checking a different address

```bash
for i in $(seq 1 20); do curl -sk -o /dev/null --interface 127.0.0.1 -H "Authorization: Bearer garbage.not.a.jwt" https://localhost:55000/agents?limit=1; done
curl -sk -o /dev/null -w "status=%{http_code}\n" --interface 127.0.0.1 -H "Authorization: Bearer garbage" https://localhost:55000/agents?limit=1
TOKEN2=$(curl -sk --interface 127.0.0.2 -u wazuh:wazuh -X POST https://localhost:55000/security/user/authenticate | python3 -c 'import sys,json; print(json.load(sys.stdin)["data"]["token"])')
curl -sk --interface 127.0.0.2 -w "\nstatus=%{http_code}\n" -H "Authorization: Bearer $TOKEN2" https://localhost:55000/agents?limit=1
```

Output:
```
status=429      # 127.0.0.1, correctly gated after its own flood
status=200      # 127.0.0.2, unaffected, login and authenticated call both succeed
```

`--interface` binds curl to a specific loopback address so the two calls look like different clients. Any two distinct addresses reachable from the host work the same way.

### 3. A genuinely authenticated low-privilege account shares the admin's address

Creates a throwaway user, assigns the built-in read-only role, floods legitimately as that user, then deletes the user again.

```bash
#!/usr/bin/env bash
set -uo pipefail
API=https://localhost:55000
ADMIN_CREDS=wazuh:wazuh
TEST_USER=poc6027readonly
TEST_PASS='Poc6027Readonly!2026'

cleanup() {
  if [ -n "${TEST_USER_ID:-}" ]; then
    curl -sk -o /dev/null -w "delete status=%{http_code}\n" -X DELETE \
      -H "Authorization: Bearer $ADMIN_TOKEN" "$API/security/users?user_ids=$TEST_USER_ID"
  fi
}
trap cleanup EXIT

ADMIN_TOKEN=$(curl -sk -u "$ADMIN_CREDS" -X POST "$API/security/user/authenticate" | python3 -c 'import sys,json; print(json.load(sys.stdin)["data"]["token"])')
curl -sk -o /dev/null -w "admin GET /agents before: status=%{http_code}\n" -H "Authorization: Bearer $ADMIN_TOKEN" "$API/agents?limit=1"

CREATE_RESP=$(curl -sk -H "Authorization: Bearer $ADMIN_TOKEN" -H "Content-Type: application/json" \
  -X POST "$API/security/users" \
  -d "{\"username\": \"$TEST_USER\", \"password\": \"$TEST_PASS\"}")
TEST_USER_ID=$(echo "$CREATE_RESP" | python3 -c 'import sys,json; d=json.load(sys.stdin); print(d["data"]["affected_items"][0]["id"])')

curl -sk -o /dev/null -w "assign role status=%{http_code}\n" -H "Authorization: Bearer $ADMIN_TOKEN" \
  -X POST "$API/security/users/$TEST_USER_ID/roles?role_ids=2"

RO_TOKEN=$(curl -sk -u "$TEST_USER:$TEST_PASS" -X POST "$API/security/user/authenticate" | python3 -c 'import sys,json; print(json.load(sys.stdin)["data"]["token"])')

curl -sk -o /dev/null -w "RO account DELETE /agents (expect 403): status=%{http_code}\n" \
  -X DELETE -H "Authorization: Bearer $RO_TOKEN" "$API/agents?agents_list=999&status=all&older_than=0s"
curl -sk -o /dev/null -w "RO account GET /agents (expect 200): status=%{http_code}\n" \
  -H "Authorization: Bearer $RO_TOKEN" "$API/agents?limit=1"

for i in $(seq 1 350); do
  curl -sk -o /dev/null -H "Authorization: Bearer $RO_TOKEN" "$API/agents?limit=1" &
done
wait

curl -sk -o /dev/null -w "admin GET /agents after RO flood: status=%{http_code}\n" \
  -H "Authorization: Bearer $ADMIN_TOKEN" "$API/agents?limit=1"
```

Output:
```
admin GET /agents before: status=200
RO account DELETE /agents (expect 403): status=403
RO account GET /agents (expect 200): status=200
admin GET /agents after RO flood: status=429
delete status=429
```

Follow-up call to see the exact error:
```
$ curl -sk -u wazuh:wazuh -X POST https://localhost:55000/security/user/authenticate
{"title": "Too Many Requests", "detail": "Maximum number of requests per minute reached", "error": 6001}
```

This reproduces, but as error 6001 (the authenticated-traffic ceiling), not the unauthenticated one. Two genuinely authenticated identities sharing one address still share one authenticated-request budget, since the limiter keys by address rather than by identity. This is a known, accepted tradeoff, not a new defect this change introduces.

### 4. The same invalid-token flood sustained over 4 bursts across 100 seconds

```bash
#!/usr/bin/env bash
set -uo pipefail
API=https://localhost:55000
CREDS=wazuh:wazuh

TOKEN=$(curl -sk -u "$CREDS" -X POST "$API/security/user/authenticate" | python3 -c 'import sys,json; print(json.load(sys.stdin)["data"]["token"])')

fire_burst() {
  for i in $(seq 1 350); do
    curl -sk -o /dev/null "$API/agents?limit=1" &
  done
  wait
}

sample_admin() {
  curl -sk -o /dev/null -w "%{http_code}" -H "Authorization: Bearer $TOKEN" "$API/agents?limit=1"
}

echo "admin before any burst: $(sample_admin)"

ok=0; denied=0; total_samples=0
for burst_num in 1 2 3 4; do
  fire_burst
  for s in 1 2 3 4 5; do
    sleep 5
    code=$(sample_admin)
    total_samples=$((total_samples + 1))
    if [ "$code" = "200" ]; then ok=$((ok + 1)); else denied=$((denied + 1)); fi
    echo "t+$((burst_num * 25 - 25 + s * 5))s sample: admin status=$code"
  done
done

echo "admin OK: $ok  admin denied: $denied  total samples: $total_samples"
```

Output:
```
admin before any burst: 200
-- 20 samples taken across all 4 bursts --
admin OK:     20
admin denied: 0
```

### 5. Ordinary low-volume unauthenticated traffic

A single bad login attempt, well under the unauthenticated ceiling.

```bash
curl -sk -w "\n%{http_code}\n" -X POST "https://localhost:55000/security/user/authenticate" -H "Authorization: Basic aW52YWxpZDppbnZhbGlk"
```

Output:
```
{"title": "Unauthorized", "detail": "Invalid credentials"}
401
```

### 6. A normal authenticated session end to end

```bash
TOKEN=$(curl -sk -u wazuh:wazuh -X POST "https://localhost:55000/security/user/authenticate" | python3 -c 'import sys,json; print(json.load(sys.stdin)["data"]["token"])')
curl -sk -w "\n%{http_code}\n" -H "Authorization: Bearer $TOKEN" "https://localhost:55000/agents?limit=1"
```

Output:
```
{"data": {"affected_items": [], "total_affected_items": 0, ...}, "message": "No agent information was returned", "error": 0}
200
```

### 7. The authenticated ceiling (300/minute) still enforces per address

```bash
TOKEN=$(curl -sk -u wazuh:wazuh -X POST "https://localhost:55000/security/user/authenticate" | python3 -c 'import sys,json; print(json.load(sys.stdin)["data"]["token"])')
for i in $(seq 1 305); do curl -sk -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $TOKEN" "https://localhost:55000/agents?limit=1"; done | sort | uniq -c
```

Output:
```
    299 200
      6 429
```

The first rejection lands exactly at request 300, both before and after the redesign described below.

### 8. A WARNING log line fires on exhaustion, naming the source

```bash
grep -i "Maximum number of.*requests per minute reached" /var/wazuh-manager/logs/api.log | tail -5
```

Output:
```
WARNING: Maximum number of requests per minute reached. IP: 127.0.0.8. User: wazuh.
(repeated for each rejected request)
```

### 9. Disabling the authenticated ceiling no longer silently disables the unauthenticated one too

Set in `api.yaml`:
```yaml
access:
  max_login_attempts: 50
  block_time: 300
  max_request_per_minute: 0
  max_unauthenticated_request_per_minute: 10
```

Restart the API daemon, then flood from a fresh address:
```bash
for i in $(seq 1 15); do curl -sk -o /dev/null -w "%{http_code} " -H "Authorization: Bearer garbage.not.a.jwt" https://localhost:55000/agents?limit=1; done
```

Output:
```
401 401 401 401 401 401 401 401 401 401 429 429 429 429 429
```

The first 10 fail authentication normally, the 11th onward is capped, confirming the unauthenticated ceiling now enforces independently of the authenticated one.

Follow-up leak check, a fresh address logging in and firing 20 authenticated requests with the same `max_request_per_minute: 0`:
```bash
TOKEN=$(curl -sk --interface 127.0.0.42 -u wazuh:wazuh -X POST https://localhost:55000/security/user/authenticate | python3 -c 'import sys,json; print(json.load(sys.stdin)["data"]["token"])')
for i in $(seq 1 20); do curl -sk --interface 127.0.0.42 -o /dev/null -w "%{http_code} " -H "Authorization: Bearer $TOKEN" https://localhost:55000/agents?limit=1; done
```

Output: `200 200 200 200 200 200 200 200 200 200 200 200 200 200 200 200 200 200 200 200`, confirming no reservation leaks into the unauthenticated bucket for real authenticated traffic.

### 10. The internal per-address tracking dictionary gets cleaned up automatically

The tracking dictionary this test checks has no HTTP-visible introspection, so confirming the cleanup task actually runs needs one temporary debug line added locally to `api/api/middlewares.py`, removed again afterward:

```diff
     async with general_request_lock:
+        import logging
+        logging.getLogger('wazuh-api').debug(f"entries before sweep: {len(general_request_stats)}")
         stale_hosts = [
             host for host, entry in general_request_stats.items()
             if all(now - bucket['window_start'] >= RATE_LIMIT_WINDOW_SECONDS for bucket in entry.values())
         ]
         for host in stale_hosts:
             del general_request_stats[host]
+        logging.getLogger('wazuh-api').debug(f"entries after sweep: {len(general_request_stats)}")
```

Set `logs.level: debug` in `api.yaml`, restart, then fire one request each from 20 distinct addresses and leave the manager alone:
```bash
for i in $(seq 50 69); do curl -sk -o /dev/null --interface 127.0.0.$i -H "Authorization: Bearer garbage.not.a.jwt" https://localhost:55000/agents?limit=1; done
```

`api.log`, two sweep cycles later (the sweep runs every 60 seconds):
```
entries before sweep: 20
entries after sweep: 20      # first sweep, too soon after the traffic to prune
entries before sweep: 20
entries after sweep: 0       # next sweep, comfortably past the 60s window, fully pruned
```

The cleanup task runs on its own schedule and reclaims memory for addresses that never come back, which the previous design had no mechanism for at all.

### 11. 500 to 8,000 distinct addresses, each independently driving toward its own 300/minute ceiling

Keying by address instead of a single global counter means N distinct addresses can together drive up to N times the per-address cap. This checks whether that's actually safe: does every address's cap stay genuinely independent (no address stealing capacity from, or getting capped by, another) once hundreds to thousands of other addresses are hammering the same manager at once, and how does the manager itself hold up under that load. One shared, already-valid Bearer token is reused across every address (bucket keying is by address only, so this legitimately exercises N independent buckets without needing N separate logins). Requires `pip install aiohttp`.

```python
#!/usr/bin/env python3
import asyncio
import aiohttp
import sys
import time
import csv
import collections


def addr_for_index(i: int) -> str:
    i += 1
    b = (i // 65025) % 256
    c = (i // 255) % 255
    d = i % 255
    return f"127.{b}.{c}.{d if d else 1}"


async def get_token():
    connector = aiohttp.TCPConnector(ssl=False)
    async with aiohttp.ClientSession(connector=connector) as s:
        async with s.post("https://127.0.0.1:55000/security/user/authenticate",
                           auth=aiohttp.BasicAuth("wazuh", "wazuh")) as r:
            d = await r.json()
            return d["data"]["token"]


async def worker(addr, token, end_time, results, sem):
    connector = aiohttp.TCPConnector(local_addr=(addr, 0), ssl=False, limit=1)
    timeout = aiohttp.ClientTimeout(total=15)
    async with aiohttp.ClientSession(connector=connector, timeout=timeout) as session:
        next_send = time.monotonic()
        while time.monotonic() < end_time:
            now = time.monotonic()
            if now < next_send:
                await asyncio.sleep(next_send - now)
            next_send += 0.2   # paces each address at ~5 req/s = 300/min
            async with sem:
                t0 = time.monotonic()
                try:
                    async with session.get("https://127.0.0.1:55000/agents?limit=1",
                                            headers={"Authorization": f"Bearer {token}"}) as resp:
                        await resp.read()
                        results.append((addr, t0, time.monotonic() - t0, resp.status))
                except Exception as e:
                    results.append((addr, t0, time.monotonic() - t0, f"ERR:{type(e).__name__}"))


async def main():
    n_addr, duration, concurrency, out_csv = int(sys.argv[1]), float(sys.argv[2]), int(sys.argv[3]), sys.argv[4]
    token = await get_token()
    sem = asyncio.Semaphore(concurrency)   # bounds total in-flight requests system-wide
    results = []
    t_start = time.monotonic()
    end_time = t_start + duration
    tasks = [worker(addr_for_index(i), token, end_time, results, sem) for i in range(n_addr)]
    await asyncio.gather(*tasks, return_exceptions=True)

    with open(out_csv, "w", newline="") as f:
        w = csv.writer(f)
        w.writerow(["address", "t_offset_s", "latency_s", "status"])
        for addr, t0, lat, status in results:
            w.writerow([addr, f"{t0 - t_start:.2f}", f"{lat:.4f}", status])

    wall = time.monotonic() - t_start
    per_addr_200 = collections.Counter()
    for addr, _, _, s in results:
        if s == 200:
            per_addr_200[addr] += 1
    counts = sorted(per_addr_200.values()) or [0]
    print(f"n_addr={n_addr} duration={wall:.1f}s total={len(results)} throughput={len(results)/wall:.0f}req/s "
          f"codes={dict(collections.Counter(s for _,_,_,s in results))} "
          f"per_address_200s(min/median/max)={counts[0]}/{counts[len(counts)//2]}/{counts[-1]} "
          f"addresses_over_300={sum(1 for c in counts if c > 300)}")


if __name__ == "__main__":
    asyncio.run(main())
```

Run once per tier, each for a real 60-second window (wall time runs longer once the manager falls behind, since already in-flight requests still have to drain):
```bash
for n in 500 1000 2000 4000 8000; do python3 scale_test.py "$n" 60 300 "out_$n.csv"; done
```

| Addresses | Wall time | Total requests | Throughput | Per-address successes (min/median/max) | Addresses that exceeded 300 |
|---|---|---|---|---|---|
| 500 | 63.6s | 4,652 | 73 req/s | 5 / 10 / 20 | 0 |
| 1,000 | 73.2s | 5,428 | 74 req/s | 5 / 5 / 12 | 0 |
| 2,000 | 84.8s | 6,505 | 77 req/s | 3 / 3 / 8 | 0 |
| 4,000 | 111.5s | 8,535 | 77 req/s | 2 / 2 / 6 | 0 |
| 8,000 | 146.2s | 13,082 | 89 req/s | 1 / 2 / 4 | 0 |

The manager process was monitored throughout: memory grew modestly and predictably (roughly 91MB total across all five tiers combined, no runaway growth), its file descriptor count stayed exactly flat the entire time, and a plain login request succeeded immediately after every single tier. One batch of 56 requests timed out, all within the same one-second window right at the start of the very first tier (500 addresses, the first traffic the manager had seen in a while) and never again at any later, larger tier, consistent with a one-time cold-start effect rather than something that gets worse as address count grows.

No address, at any scale up to 8,000, ever recorded more than 6 successful requests in these windows, because a separate, pre-existing bottleneck in the authentication path caps the manager's real aggregate throughput at roughly 75-90 requests per second regardless of how many addresses are asking, or how much concurrency is thrown at it. That number stays essentially flat from 500 addresses all the way to 8,000, meaning no single address gets anywhere near its own 300/minute allowance within a 60 to 150 second window at these scales. That's worth being explicit about: this run confirms addresses never cross-contaminate each other and the manager's known bottleneck doesn't get worse as address count grows, but it does not, on its own, exercise the 300/minute ceiling actually firing at large N, since throughput never lets any one address get close enough to hit it. The ceiling firing correctly per address is confirmed separately, at a scale small enough for one address to actually reach it (5 to 10 concurrently active addresses, each independently capped at its own limit with none affecting the others).

## What changed to fix scenario 1

The initial design charged a small unauthenticated-request bucket for every request before authentication ran, then released that charge afterward if the request turned out to be authenticated. That left a window where legitimate traffic sharing an address with prior noise could still be rejected before authentication ever got a chance to run. It also meant genuinely concurrent legitimate traffic from a single address, no attacker involved at all, could trip the same bucket: a small script firing 30 real concurrent authenticated calls with one valid token from one address, tested separately from what's shown above, saw 17 of 30 rejected under the original design (each request reserved a slot in the small bucket before it had finished authenticating, and only released it afterward), versus 0 of 30 rejected after the fix below.

The fix reorders the accounting to run after the authentication outcome is already known: a request that authenticates successfully is never charged against the unauthenticated bucket at all, only a request that actually fails authentication is. A benchmark of this API's JWT verification (ES512) showed the cost of doing this is 5 to 650 microseconds per request, three to four orders of magnitude cheaper than password-hashing-based login, and the one endpoint class where that concern would be real (the login endpoints) is untouched by this change and already protected by a separate, existing mechanism.

jotacarma90 (changes requested):
Reviewed the branch and exercised the real middleware stack in its production positions against `spec.yaml`. The keying works: distinct addresses get independent buckets, the rollover off-by-one is gone, and `max_request_per_minute: 0` no longer disables the unauthenticated limit. Two things before this goes in.

### Anonymous requests to unrouted paths are charged to the authenticated bucket

`SecurityMiddleware` registers no operation for a path outside the spec (`auth_all_paths` is False), so it raises nothing and the request reaches `CheckAuthenticatedRateLimitMiddleware`, which assumes that getting there proves authentication succeeded.

```
GET /agents         (no creds) -> 401  bucket={'unauthenticated': 1}
GET /does-not-exist (no creds) -> 404  bucket={'authenticated': 1}   <--
```

301 anonymous 404s from one address leave its authenticated bucket at 302, and an authenticated request from that same address then gets 429 / 6001. So unauthenticated traffic gets 300/min per address instead of 10, and "unauthenticated traffic can never spend an authenticated caller's budget" does not hold as written.

Fix: pick the bucket from the request rather than from the middleware's position. The signal is already in scope there:

```python
context = (request.scope.get('extensions') or {}).get('connexion_context') or {}
if not context.get('user'):
    # no authenticated identity: charge the unauthenticated bucket instead
```

It holds `{'user': ..., 'token_info': {...}}` after a successful authentication and is `None` on an unrouted path. It also covers the future `security: []` operation the class docstring flags as uncovered, so that caveat can go.

### Over-quota requests now pay for authentication before being rejected

Charging after `SecurityMiddleware` means a request already over quota still runs the full authentication path: 400 requests with a valid token from one address give 300 served and 100 rejected with 6001, with `decode_token` invoked 400 times. That call is a `jwt.decode` plus two `pool.submit(asyncio.run, ...)` round trips, one of which re-reads `security.yaml` from disk. Before this PR those requests were rejected before security ran, so authentication work was bounded by `max_request_per_minute`; now the requests served are bounded but the authentication work is not.

Both properties fit together cheaply: in `CheckRateLimitsMiddleware`, read the authenticated bucket before `call_next` without charging it and reject with 6001 if it is already over its ceiling, then keep charging it after the outcome exactly as it does today.

### Minor

- `api/tools/env/wazuh-manager/entrypoint.sh:26` raises `max_request_per_minute` to 99999 but leaves the new key at its default, so that environment now returns 429 after 10 failed-auth requests per minute per address. Same for any suite that only overrides the old key.
- The fixed window is still there (300 at t=59s plus 300 at t=61s is 600 requests from one address in two seconds). If that is out of scope here, worth saying so explicitly.
- The description still describes `SettleRateLimitMiddleware` and the reserve-then-settle design, and lists `test_reserve_*` / `test_settle_*` tests that are not in the branch.

MiguelazoDS:
Great, so the issues explained with my personal understanding. 

The first bug is that I was not covering un-routed paths, which are non-existent endpoints. This affect unauthorized and authorized requests, but the one that suffer is the authorized one because it counts in the authenticated counter (300/min default). 

For bug 2, there's an unneeded authentication validation when the quota was reached. So a pointless load by validating the authentication is performed when the request will be rejected for over-quota.

jotacarma90 (approved):
LGTM

PR #38877, opened 2026-09-04
Anchor the agent HTTPS client's interruptible wait to CLOCK_MONOTONIC

### LINKED ISSUE #38825: Agent stops sending for the duration of a backward system-clock jump
## Summary

When the agent's system clock moves **backwards**, the agent stops sending every timer-driven message for the full duration of the jump, with no error in any log at any debug level. A one-hour backward jump mutes the agent for one hour. The JWT clock-skew self-correction works correctly and is not the problem — the agent recovers its credentials immediately and then goes quiet anyway.

A forward jump does not show this, which is why it has not been caught: a forward jump makes pending deadlines look already-due, so traffic actually resumes faster than normal.

## Steps to reproduce

1. Agent 5.0.0 connected to a manager over HTTPS, healthy, `agent.debug=2`.
2. Make sure nothing will pull the clock back (`timedatectl set-ntp false`, and on VirtualBox guests `pkill -x VBoxService` — stopping `vboxadd-service` is not enough).
3. `sudo date -s '-1 hour'`.
4. Watch `/control notify` and `/stateless` traffic stop, while the agent process stays alive.

## Observed

The skew correction fires and succeeds, and an event-driven `/stateful` session still gets through:

```sql
2026/09/02 15:16:26 wazuh-agentd:https-client[3050] retrySender.cpp:248 at correctClockIfSkewed(): INFO: https_client: clock skew of 3601 s detected against the manager's response (Date header); correcting the token timestamp for this and future requests.
2026/09/02 15:16:26 wazuh-agentd[3050] https_client_bridge.c:1162 at bridge_on_sync_response(): DEBUG: https_client /stateful session=syscollector-7325131514656604160 result=200 (15 byte answer)
2026/09/02 15:16:27 wazuh-agentd[3050] https_client_bridge.c:1162 at bridge_on_sync_response(): DEBUG: https_client /stateful session=fim-7325131519161851904 result=200 (15 byte answer)
```

After that, for ~10 minutes the only thing the agent logs is its state file:

```sql
2026/09/02 15:22:39 wazuh-agentd[3050] state.c:83 at write_state(): DEBUG: Updating state file.
```

Manager-side counters over that window confirm nothing arrived (`notify` had been advancing every 10 s before the jump):

```
remoted.control.notify                  879 -> 879
remoted.http.stateless.responses.2xx    265 -> 265
remoted.auth.reject.clock_skew            1 (one rejection, correctly corrected)
remoted.http.enroll.responses.2xx         0 (no re-enrollment)
```

Restoring the clock releases the stall immediately — `remoted.control.notify` went 879 -> 882 within 30 seconds and the 10 s cadence resumed, which is the signature of a wall-clock-anchored deadline rather than a wedged thread.

## Root cause

`Waiter::waitFor()` in `src/client-agent/https_client/src/stopToken.hpp` uses `std::condition_variable::wait_for`. libstdc++ only makes that immune to system-clock changes when it was configured with `_GLIBCXX_USE_PTHREAD_COND_CLOCKWAIT` (needs `pthread_cond_clockwait`, glibc >= 2.30); otherwise it converts the deadline to `CLOCK_REALTIME` and calls `pthread_cond_timedwait`, so a backward jump of N seconds extends the wait by N seconds.

The agent is built in the Debian 7 / glibc 2.13 package container, and the shipped binaries confirm the fallback path is what got compiled — `pthread_cond_clockwait` is imported by nothing, while `pthread_cond_timedwait` is imported by `libhttps_client.so`, `libsca.so`, `libsyscollector.so`, `libagent_sync_protocol.so` and `libwazuhext.so`.

Note the cadence design's intent is already correct and is not what broke: intervals are measured with `steadyNow()`, and `sysSeams.hpp` states "Cadence timing is never skew-corrected". The gap is that the *wait* reintroduces a wall-clock dependency that the cadence design was deliberately avoiding.

## Bumping the build glibc is probably not the fix

- The Debian 7 builder exists to set the oldest glibc the agent supports. Raising it drops support for old distributions, so it is a product decision rather than a bug fix.
- `_GLIBCXX_USE_PTHREAD_COND_CLOCKWAIT` is baked into libstdc++ when the toolchain is built, and that container builds its own GCC against glibc 2.13. Installing a newer glibc would not be enough; the toolchain would have to be rebuilt.
- It is not necessary. `pthread_condattr_setclock(CLOCK_MONOTONIC)` has been available since glibc 2.3.3 and already works in this exact build environment.

## Suggested fix

This repository already solves the same problem, in `src/os_auth/src/auth.c`, with a graceful fallback and a comment describing the tradeoff:

```c
#ifdef CLOCK_MONOTONIC
    pthread_condattr_t attr;
    if (pthread_condattr_init(&attr) == 0) {
        if (pthread_condattr_setclock(&attr, CLOCK_MONOTONIC) == 0) {
            purge_cond_monotonic = true;
        }
```

`std::condition_variable` cannot be told which clock to use, so the C++ side needs a small monotonic-condvar wrapper over `pthread_cond_t` following that same pattern, used wherever a timeout means "N seconds from now" rather than "at this wall-clock instant".

Worth scoping beyond https_client: there are ~24 `wait_for`/`wait_until` call sites under `src/client-agent/` and `src/shared_modules/`, and `src/shared_modules/utils/conditionSync.hpp` (`ConditionSync::waitFor`) has the identical construct, so a shared helper would likely fix several of them at once.

## Impact

An agent whose clock is corrected backwards (NTP stepping after a bad RTC, VM snapshot restore, a timezone/hardware-clock mix-up) silently stops reporting for the size of the correction, and presents as "agent stopped sending events" with nothing in the agent log and nothing in the manager log to explain it.

## Related observability gap

Across this testing session the manager recorded 42 clock-skew rejections and produced zero log lines for them. `AuthError::StaleToken` has no dedicated log message; only `AddressNotAllowed` gets a throttled INFO that names the agent. Clock-skew rejections are only visible via the `remoted.auth.reject.clock_skew` counter on remoted's admin metrics socket.

Miguevrgo:
## Description

When the agent's system clock steps **backwards**, every timer-driven message stops for the full duration of the jump, with nothing logged at any debug level. A one-hour backward step mutes the agent for one hour; restoring the clock releases it immediately.

The cadence design itself is correct: intervals are measured with `IClock::steadyNow()` and `sysSeams.hpp` states that cadence timing is never skew-corrected. The wall-clock dependency was reintroduced by the *wait*, not by the interval maths.

`Waiter::waitFor()` (`src/client-agent/https_client/src/stopToken.hpp:41`) slept on `std::condition_variable::wait_for`. libstdc++ only keeps that immune to system-clock changes when it was configured with `_GLIBCXX_USE_PTHREAD_COND_CLOCKWAIT` (needs `pthread_cond_clockwait`, glibc >= 2.30). The agent is built in the Debian 7 / glibc 2.13 packaging container, so it gets the fallback: the deadline is converted to `CLOCK_REALTIME` and handed to `pthread_cond_timedwait`, and a backward jump of N seconds extends the pending wait by N seconds.

`Waiter` is the single sleep behind all four HTTPS stream loops, which is why one bug silences everything at once:

- `httpsClientFacade.cpp:297` `/control notify`
- `httpsClientFacade.cpp:315` `/stateless`
- `httpsClientFacade.cpp:339` `/stateful` batching
- `httpsClientFacade.cpp:357` reporter
- `retrySender.cpp:114` retry backoff

Event-driven `/stateful` sessions still got through during the stall because the bridge calls `Waiter::notify()`, which signals the condition variable and releases the wait; the cadence loops have nothing to signal them.

Bumping the build toolchain's glibc is not the fix: the Debian 7 builder defines the oldest glibc the agent supports, `_GLIBCXX_USE_PTHREAD_COND_CLOCKWAIT` is baked into libstdc++ when the toolchain is built, and `pthread_condattr_setclock(CLOCK_MONOTONIC)` has been available since glibc 2.3.3 and already works in that exact build environment.

Closes #38825

## Proposed Changes

- Added `src/shared_modules/utils/monotonicCondition.hpp`: a condition variable pinned to `CLOCK_MONOTONIC` through `pthread_condattr_setclock`, with the same graceful fallback to the default clock used by `src/os_auth/src/auth.c`. Call sites keep `std::mutex` and `std::unique_lock` (the `pthread_mutex_t` that `pthread_cond_timedwait` needs comes from `std::mutex::native_handle()`) and a predicate-based `waitFor()`, so the shape is the standard-library one. Non-Linux platforms keep `std::condition_variable`.
- `Waiter` (`stopToken.hpp`) now waits on it. Public API, semantics and the `FakeWaiter` test double are unchanged.
- Added `src/client-agent/https_client/tests/unit/stopToken_test.cpp`, which had no unit coverage before.
- CHANGELOG entry under Agent / Fixed.

Deliberately out of scope, and worth a follow-up issue: `agent_sync_protocol.cpp:501,755,1084,1105`, `persistent_queue.cpp:82` and `shared_modules/utils/threadSafeQueue.h:121` / `threadSafeMultiQueue.hpp:68` share the same construct on the agent, and `shared_modules/utils/conditionSync.hpp` does too (its users are manager-side). They can all move onto this helper. The observability gap reported in the issue (no log line for `AuthError::StaleToken` on `remoted`) is manager-side and also needs its own issue.

### Results and Evidence

**Backward clock jump, Debian 12 incus VM, NTP disabled, `CLOCK_REALTIME` stepped back 3600 s two seconds into a 15 s wait.**

Three waits run side by side: the `Waiter` from this branch, the pre-fix `std::condition_variable::wait_for` code, and the `pthread_cond_timedwait` + `CLOCK_REALTIME` deadline that libstdc++ emits without `_GLIBCXX_USE_PTHREAD_COND_CLOCKWAIT` (that is, the code path the shipped agent executes). Elapsed time is measured on `steady_clock`.

```
wait timeout = 15000 ms, backward clock step = 3600 s at t+2s

t+2s: stepping CLOCK_REALTIME back 3600 s
  [branch Waiter (monotonic)   ] returned after 15000 ms
  [pre-fix std::condition_var  ] returned after 15000 ms

t+45s status: branch=returned  pre-fix-std=returned  shipped-fallback=STILL WAITING
t+45s: restoring CLOCK_REALTIME (+3600 s)
  [shipped fallback (realtime) ] returned after 45000 ms

t+60s final: branch=15000 ms  pre-fix-std=15000 ms  shipped-fallback=45000 ms
```

The realtime-anchored wait sat through the entire jump and returned the instant the clock was restored (45000 ms instead of 15000 ms), which is exactly the signature reported in the issue: `remoted.control.notify` resuming the moment the clock is put back. The branch's `Waiter` is unaffected.

The pre-fix `std::condition_variable` line returning on time is expected and is the reason this went unnoticed in development: the VM (and any modern host) has glibc 2.44 / `_GLIBCXX_USE_PTHREAD_COND_CLOCKWAIT`, so a host build of the agent cannot reproduce the bug. Only the packaging container's glibc 2.13 build takes the fallback path, which the issue confirms from the shipped binaries (`pthread_cond_clockwait` imported by nothing; `pthread_cond_timedwait` imported by `libhttps_client.so`). The fix removes the dependency on that toolchain detail entirely, so both builds now behave the same.

**Unit tests**

```
[==========] Running 6 tests from 1 test suite.
[ RUN      ] StopTokenTest.WaitElapsesOnTheSteadyClock
[       OK ] StopTokenTest.WaitElapsesOnTheSteadyClock (150 ms)
[ RUN      ] StopTokenTest.NotifyWakesAPendingWait
[       OK ] StopTokenTest.NotifyWakesAPendingWait (50 ms)
[ RUN      ] StopTokenTest.RequestStopWakesAPendingWaitAndReportsStop
[       OK ] StopTokenTest.RequestStopWakesAPendingWaitAndReportsStop (50 ms)
[ RUN      ] StopTokenTest.WaitReturnsAtOnceAfterStop
[       OK ] StopTokenTest.WaitReturnsAtOnceAfterStop (0 ms)
[ RUN      ] StopTokenTest.PendingNotifyIsConsumedByASingleWait
[       OK ] StopTokenTest.PendingNotifyIsConsumedByASingleWait (100 ms)
[ RUN      ] StopTokenTest.NonPositiveTimeoutDoesNotBlock
[       OK ] StopTokenTest.NonPositiveTimeoutDoesNotBlock (0 ms)
[  PASSED  ] 6 tests.
```

The same binary built with `-fsanitize=thread` passes with no reports.

**Two red checks are pre-existing and not caused by this change**

`IT Windows - agentd (tier-0-1)` fails on `test_agentd_state_config[Undefined_state_interval_value]` and, intermittently, on `test_agentd_reconection_enrollment_with_keys[TCP protocol]`. This was bisected on the branch itself: commit `5395692cf0` restored `stopToken.hpp` byte for byte to the 5.0.0 implementation, leaving the new header in the tree unused, and the same tests still failed.

```
FAILED test_agentd_state_config[Undefined_state_interval_value] - AssertionError: Error invalid configuration event not detected
FAILED test_agentd_reconection_enrollment_with_keys[TCP protocol] - AssertionError: Connected to the server message not found
```

That enrollment failure is the same one PR #38869 already carries on this job. The commit was reverted in `a7ccf0c8e8`.

For the record on why it surfaces here: the job only runs when `src/client-agent/**` changes (recent agentd PRs reported `IT Windows - no relevant changes`), the 5.0.0 base has not moved since the last passing run of it, the `qa-integration-framework` 5.0.0 branch has not moved either, the runner image is identical, and the test selection and ordering match. On Windows this change is a no-op by construction: `monotonicCondition.hpp` takes its `std::condition_variable` branch off Linux, and `rtr (client-agent/https_client, winagent)`, the Windows package build and the MSI test all pass.

`Run integration tests for API endpoints (test_security_POST_endpoints.tavern.yaml)` fails on `POST /security/user/authenticate/run_as` with `assert payload['rbac_roles'] == expected_roles` (`api/test/integration/tavern_utils.py:269`), after its own three internal reruns. That is manager RBAC; this PR contains no Python and no API or RBAC change.

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel
  - [x] TSAN for test and wazuh-engine.

- Wazuh server API/Framework

### Artifacts Affected

- `libhttps_client` (`wazuh-agentd`) on every agent platform. Behavior changes only on Linux, where the new header takes the `CLOCK_MONOTONIC` path; Windows and macOS keep the standard-library implementation.

### Configuration Changes

N/A

### Tests Introduced

`src/client-agent/https_client/tests/unit/stopToken_test.cpp` covers `Waiter` for the first time: the timeout elapsing on the steady clock, `notify()` waking a pending wait, `requestStop()` waking it and reporting the stop through both `stopRequested()` and `stopFlag()`, an immediate return once stop was already requested, a pending `notify()` being consumed by exactly one wait, and non-positive timeouts not blocking.

The clock jump itself needs `CAP_SYS_TIME` and a private clock, so it is verified in the VM run above rather than in the unit suite.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/src/client-agent/https_client/src/stopToken.hpp
+++ b/src/client-agent/https_client/src/stopToken.hpp
@@ -12,9 +12,9 @@
 #ifndef _HC_STOP_TOKEN_HPP
 #define _HC_STOP_TOKEN_HPP
 
+#include "monotonicCondition.hpp"
 #include <atomic>
 #include <chrono>
-#include <condition_variable>
 #include <mutex>
 
 /**
@@ -38,7 +38,7 @@ class Waiter
         virtual bool waitFor(std::chrono::milliseconds timeout)
         {
             std::unique_lock<std::mutex> lock(m_mutex);
-            m_cv.wait_for(lock, timeout, [this] { return m_stop.load() || m_wake; });
+            m_condition.waitFor(lock, timeout, [this] { return m_stop.load() || m_wake; });
             m_wake = false;
             return !m_stop.load();
         }
@@ -50,7 +50,7 @@ class Waiter
                 std::lock_guard<std::mutex> lock(m_mutex);
                 m_wake = true;
             }
-            m_cv.notify_all();
+            m_condition.notifyAll();
         }
 
         /// Requests cooperative stop and wakes any pending waitFor().
@@ -60,7 +60,7 @@ class Waiter
                 std::lock_guard<std::mutex> lock(m_mutex);
                 m_stop = true;
             }
-            m_cv.notify_all();
+            m_condition.notifyAll();
         }
 
         bool stopRequested() const
@@ -76,7 +76,7 @@ class Waiter
 
     private:
         std::mutex m_mutex;
-        std::condition_variable m_cv;
+        MonotonicCondition m_condition;
         std::atomic<bool> m_stop {false};
         bool m_wake {false};
 };
--- a/src/client-agent/https_client/tests/unit/stopToken_test.cpp
+++ b/src/client-agent/https_client/tests/unit/stopToken_test.cpp
@@ -1,0 +1,103 @@
+/*
+ * Wazuh agent HTTPS client (C++ transport module)
+ * Copyright (C) 2015, Wazuh Inc.
+ * September 4, 2026.
+ *
+ * This program is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU General Public
+ * License (version 2) as published by the FSF - Free Software
+ * Foundation.
+ */
+
+#include "stopToken.hpp"
+#include <chrono>
+#include <gtest/gtest.h>
+#include <thread>
+
+using namespace std::chrono_literals;
+
+TEST(StopTokenTest, WaitElapsesOnTheSteadyClock)
+{
+    Waiter waiter;
+    const auto start {std::chrono::steady_clock::now()};
+
+    EXPECT_TRUE(waiter.waitFor(150ms));
+
+    EXPECT_GE(std::chrono::steady_clock::now() - start, 140ms);
+    EXPECT_FALSE(waiter.stopRequested());
+}
+
+TEST(StopTokenTest, NotifyWakesAPendingWait)
+{
+    Waiter waiter;
+    std::thread waker(
+        [&waiter]
+    {
+        std::this_thread::sleep_for(50ms);
+        waiter.notify();
+    });
+
+    const auto start {std::chrono::steady_clock::now()};
+    const bool keepRunning {waiter.waitFor(30s)};
+    const auto elapsed {std::chrono::steady_clock::now() - start};
+    waker.join();
+
+    EXPECT_TRUE(keepRunning);
+    EXPECT_LT(elapsed, 5s);
+}
+
+TEST(StopTokenTest, RequestStopWakesAPendingWaitAndReportsStop)
+{
+    Waiter waiter;
+    std::thread stopper(
+        [&waiter]
+    {
+        std::this_thread::sleep_for(50ms);
+        waiter.requestStop();
+    });
+
+    const auto start {std::chrono::steady_clock::now()};
+    const bool keepRunning {waiter.waitFor(30s)};
+    const auto elapsed {std::chrono::steady_clock::now() - start};
+    stopper.join();
+
+    EXPECT_FALSE(keepRunning);
+    EXPECT_LT(elapsed, 5s);
+    EXPECT_TRUE(waiter.stopRequested());
+    EXPECT_TRUE(waiter.stopFlag()->load());
+}
+
+TEST(StopTokenTest, WaitReturnsAtOnceAfterStop)
+{
+    Waiter waiter;
+    waiter.requestStop();
+
+    const auto start {std::chrono::steady_clock::now()};
+
+    EXPECT_FALSE(waiter.waitFor(30s));
+    EXPECT_LT(std::chrono::steady_clock::now() - start, 5s);
+}
+
+TEST(StopTokenTest, PendingNotifyIsConsumedByASingleWait)
+{
+    Waiter waiter;
+    waiter.notify();
+
+    auto start {std::chrono::steady_clock::now()};
+    EXPECT_TRUE(waiter.waitFor(30s));
+    EXPECT_LT(std::chrono::steady_clock::now() - start, 5s);
+
+    start = std::chrono::steady_clock::now();
+    EXPECT_TRUE(waiter.waitFor(100ms));
+    EXPECT_GE(std::chrono::steady_clock::now() - start, 90ms);
+}
+
+TEST(StopTokenTest, NonPositiveTimeoutDoesNotBlock)
+{
+    Waiter waiter;
+    const auto start {std::chrono::steady_clock::now()};
+
+    EXPECT_TRUE(waiter.waitFor(0ms));
+    EXPECT_TRUE(waiter.waitFor(-5ms));
+    EXPECT_LT(std::chrono::steady_clock::now() - start, 5s);
+}
--- a/src/shared_modules/sync_protocol/src/agent_sync_protocol.cpp
+++ b/src/shared_modules/sync_protocol/src/agent_sync_protocol.cpp
@@ -498,7 +498,7 @@ bool AgentSyncProtocol::requiresFullSync(const std::string& index,
                      "); the indexer may not have caught up with a recent write yet, retrying.");
 
             std::unique_lock<std::mutex> lock(m_syncState.mtx);
-            m_syncState.cv.wait_for(lock, CHECKSUM_RETRY_DELAY, [this] { return shouldStop(); });
+            m_syncState.cv.waitFor(lock, CHECKSUM_RETRY_DELAY, [this] { return shouldStop(); });
         }
         else
         {
@@ -752,7 +752,7 @@ flatbuffers::Offset<Wazuh::SyncSchema::Start> AgentSyncProtocol::waitMetadataAnd
                     logged = true;
                 }
 
-                m_syncState.cv.wait_for(lock, std::chrono::seconds(1));
+                m_syncState.cv.waitFor(lock, std::chrono::seconds(1));
             }
         }
 
@@ -1084,7 +1084,7 @@ bool AgentSyncProtocol::runSession(const SessionContent& content)
             m_logger(LOG_DEBUG, "Failed to hand session " + std::to_string(session) + " to the agent.");
 
             std::unique_lock<std::mutex> lock(m_syncState.mtx);
-            m_syncState.cv.wait_for(lock, RESEND_BACKOFF, [&]
+            m_syncState.cv.waitFor(lock, RESEND_BACKOFF, [&]
             {
                 return shouldStop();
             });
@@ -1105,7 +1105,7 @@ bool AgentSyncProtocol::runSession(const SessionContent& content)
 #else
         const auto waitTimeout = SESSION_RESPONSE_TIMEOUT;
 #endif
-        const bool conditionMet = m_syncState.cv.wait_for(lock, waitTimeout, [&]
+        const bool conditionMet = m_syncState.cv.waitFor(lock, waitTimeout, [&]
         {
             return m_syncState.responseReceived || m_syncState.syncFailed || shouldStop();
         });
@@ -1223,7 +1223,7 @@ bool AgentSyncProtocol::applyHttpResult(int httpCode, std::string_view body, uin
     {
         m_syncState.lastSyncResult = SyncResult::SUCCESS;
         m_syncState.responseReceived = true;
-        m_syncState.cv.notify_all();
+        m_syncState.cv.notifyAll();
         m_logger(LOG_DEBUG, "Sync response received with success status: " + std::string(body));
         return true;
     }
@@ -1275,7 +1275,7 @@ bool AgentSyncProtocol::applyHttpResult(int httpCode, std::string_view body, uin
             break;
     }
 
-    m_syncState.cv.notify_all();
+    m_syncState.cv.notifyAll();
     return true;
 }
 
@@ -1382,7 +1382,7 @@ void AgentSyncProtocol::stop()
     // This prevents crashes when the object is destroyed while waiting
     {
         std::lock_guard<std::mutex> lock(m_syncState.mtx);
-        m_syncState.cv.notify_all();
+        m_syncState.cv.notifyAll();
     }
 
     m_logger(LOG_DEBUG, "Stop requested for sync protocol.");
--- a/src/shared_modules/sync_protocol/src/persistent_queue.cpp
+++ b/src/shared_modules/sync_protocol/src/persistent_queue.cpp
@@ -42,7 +42,7 @@ PersistentQueue::~PersistentQueue()
         std::lock_guard<std::mutex> lock(m_mutex);
         m_stop = true;
     }
-    m_cv.notify_one();
+    m_cv.notifyOne();
 
     if (m_flushThread.joinable())
     {
@@ -66,7 +66,7 @@ void PersistentQueue::submit(const std::string& id,
 
     if (shouldNotify)
     {
-        m_cv.notify_one();
+        m_cv.notifyOne();
     }
 }
 
@@ -79,7 +79,7 @@ void PersistentQueue::flushLoop()
 
         {
             std::unique_lock<std::mutex> lock(m_mutex);
-            m_cv.wait_for(lock, FLUSH_INTERVAL, [this]
+            m_cv.waitFor(lock, FLUSH_INTERVAL, [this]
             {
                 return m_buffers[m_currentIdx].size() >= FLUSH_BATCH_SIZE || m_stop.load();
             });
--- a/src/shared_modules/utils/monotonicCondition.hpp
+++ b/src/shared_modules/utils/monotonicCondition.hpp
@@ -21,14 +21,7 @@
 #include <pthread.h>
 #endif
 
-/**
- * @brief Condition variable whose waits are immune to system-clock jumps.
- *
- * std::condition_variable::wait_for is monotonic only when libstdc++ was built
- * with _GLIBCXX_USE_PTHREAD_COND_CLOCKWAIT (glibc >= 2.30). The agent's
- * toolchain predates it, so its deadlines land on CLOCK_REALTIME and a
- * backward clock jump of N seconds extends every pending wait by N seconds.
- */
+/// Condition variable whose waits are immune to system-clock jumps (see PR description for the libstdc++/glibc rationale).
 #if defined(__linux__)
 
 class MonotonicCondition final
@@ -81,6 +74,19 @@ public:
         pthread_cond_broadcast(&m_cond);
     }
 
+    void notifyOne()
+    {
+        pthread_cond_signal(&m_cond);
+    }
+
+    /// The clock this instance's waits are actually bound to; CLOCK_REALTIME
+    /// means pthread_condattr_setclock(CLOCK_MONOTONIC) failed and waits fell
+    /// back to wall-clock semantics.
+    clockid_t clockId() const
+    {
+        return m_clock;
+    }
+
 private:
     std::chrono::nanoseconds now() const
     {
@@ -115,6 +121,11 @@ public:
     void notifyAll()
     {
         m_cv.notify_all();
+    }
+
+    void notifyOne()
+    {
+        m_cv.notify_one();
     }
 
 private:
--- a/src/shared_modules/utils/monotonicCondition.hpp
+++ b/src/shared_modules/utils/monotonicCondition.hpp
@@ -21,7 +21,8 @@
 #include <pthread.h>
 #endif
 
-/// Condition variable whose waits are immune to system-clock jumps (see PR description for the libstdc++/glibc rationale).
+/// Condition variable whose waits are immune to system-clock jumps (see PR description for the libstdc++/glibc
+/// rationale).
 #if defined(__linux__)
 
 class MonotonicCondition final
--- a/src/shared_modules/utils/monotonicCondition.hpp
+++ b/src/shared_modules/utils/monotonicCondition.hpp
@@ -70,6 +70,15 @@ public:
         return true;
     }
 
+    /// Waits up to timeout for a single notify, with no predicate to re-check on
+    /// spurious wakeups (mirrors std::condition_variable::wait_for(lock, timeout)).
+    bool waitFor(std::unique_lock<std::mutex>& lock, std::chrono::nanoseconds timeout)
+    {
+        const auto deadline {toTimespec(now() + timeout)};
+
+        return pthread_cond_timedwait(&m_cond, lock.mutex()->native_handle(), &deadline) == 0;
+    }
+
     void notifyAll()
     {
         pthread_cond_broadcast(&m_cond);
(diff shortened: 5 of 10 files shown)


Darioortegaleyva (dismissed):
LGTM

src/client-agent/https_client/src/stopToken.hpp
```diff
@@ -38,7 +38,7 @@ class Waiter
         virtual bool waitFor(std::chrono::milliseconds timeout)
         {
             std::unique_lock<std::mutex> lock(m_mutex);
-            m_cv.wait_for(lock, timeout, [this] { return m_stop.load() || m_wake; });
+            m_condition.waitFor(lock, timeout, [this] { return m_stop.load() || m_wake; });
```
MarcelKemp, line 41: **[S2 · blocking · introduced]** The helper is adopted here and nowhere else, so the agent still stops delivering data during a backward jump. `PersistentQueue::flushLoop` (`persistent_queue.cpp:82`) waits `FLUSH_INTERVAL` = 500 ms with the predicate `size() >= FLUSH_BATCH_SIZE`, and `submit()` only notifies once 100 items are buffered (`persistent_queue.cpp:64`) — so on a low-volume agent the flush is purely timer-driven, still `CLOCK_REALTIME`-anchored, and a one-hour step back parks FIM/syscollector deltas in memory for the whole hour, neither persisted nor sent.

Since the PR carries `Closes #38825`, please either migrate that one site (it is the only deferred one that actually withholds agent data) or file the follow-up issue and downgrade to a partial reference. The migration needs the `notifyOne()` from the comment on `notifyAll` below:

```cpp
// src/shared_modules/sync_protocol/src/persistent_queue.hpp
-    std::condition_variable m_cv;
+    MonotonicCondition m_cv;

// src/shared_modules/sync_protocol/src/persistent_queue.cpp:82
-            m_cv.wait_for(lock, FLUSH_INTERVAL, [this]
+            m_cv.waitFor(lock, FLUSH_INTERVAL, [this]
             {
                 return m_buffers[m_currentIdx].size() >= FLUSH_BATCH_SIZE || m_stop.load();
             });

// and in submit()/stop(): m_cv.notify_one() -> m_cv.notifyOne()
```

src/client-agent/https_client/tests/unit/stopToken_test.cpp
```diff
@@ -0,0 +1,103 @@
...
+TEST(StopTokenTest, WaitElapsesOnTheSteadyClock)
+{
+    Waiter waiter;
+    const auto start {std::chrono::steady_clock::now()};
+
+    EXPECT_TRUE(waiter.waitFor(150ms));
+
+    EXPECT_GE(std::chrono::steady_clock::now() - start, 140ms);
+    EXPECT_FALSE(waiter.stopRequested());
+}
```
MarcelKemp, line 28: **[S2 · blocking · introduced]** None of the six new tests fails against the pre-fix implementation, so the clock-anchoring invariant is not pinned by anything. This one asserts a 150 ms wait takes at least 140 ms, which the old `m_cv.wait_for` satisfies; the other five exercise notify/requestStop/pending-notify/non-positive-timeout semantics the diff does not change — and the PR body's own bisect confirms it (`5395692cf0` restored `stopToken.hpp` byte-for-byte and the suite stayed green), so a refactor back onto `std::condition_variable` would pass CI.

The invariant is testable without `CAP_SYS_TIME` using the accessor suggested on the constructor:

```suggestion
TEST(StopTokenTest, WaitElapsesOnTheSteadyClock)
{
    Waiter waiter;
    const auto start {std::chrono::steady_clock::now()};

    EXPECT_TRUE(waiter.waitFor(150ms));

    EXPECT_GE(std::chrono::steady_clock::now() - start, 140ms);
    EXPECT_FALSE(waiter.stopRequested());
}

#if defined(__linux__)
// The whole point of #38825: the wait must not be anchored to CLOCK_REALTIME.
// Stepping the clock needs CAP_SYS_TIME, so pin the resolved clock instead --
// a refactor back onto std::condition_variable turns this red.
TEST(StopTokenTest, WaitIsAnchoredToTheMonotonicClock)
{
    MonotonicCondition condition;

    EXPECT_EQ(condition.clockId(), CLOCK_MONOTONIC);
}
#endif
```

`monotonicCondition.hpp` also deserves its own `monotonicCondition_test.cpp` under `src/shared_modules/utils/tests/` (with a CMakeLists entry — that file enumerates every test by name) to cover the constructor branches, the predicate re-check loop and `toTimespec`.

src/shared_modules/utils/monotonicCondition.hpp
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        if (pthread_condattr_setclock(&attr, CLOCK_MONOTONIC) == 0)
+        {
+            m_clock = CLOCK_MONOTONIC;
+        }
+
+        pthread_cond_init(&m_cond, &attr);
+        pthread_condattr_destroy(&attr);
+    }
```
MarcelKemp: **[S3 · non-blocking · introduced]** `pthread_cond_init`'s return is unchecked after `m_clock` has already been switched, so an init failure leaves the deadline clock and the condvar's clock disagreeing. `m_cond` would stay at its `PTHREAD_COND_INITIALIZER` value (`CLOCK_REALTIME`) while deadlines are computed as seconds-since-boot, which that condvar reads as 1970 — always past, instant `ETIMEDOUT`, and the four cadence loops plus `retrySender.cpp:114` spin unthrottled. Unreachable on glibc, hence S3, but the fallback is also entirely unobservable, which is the same silent-failure mode the issue reports.

```suggestion
        if (pthread_condattr_setclock(&attr, CLOCK_MONOTONIC) == 0 && pthread_cond_init(&m_cond, &attr) == 0)
        {
            m_clock = CLOCK_MONOTONIC;
        }

        pthread_condattr_destroy(&attr);
    }

    /// The clock the deadlines are really anchored to: CLOCK_MONOTONIC when the
    /// condvar accepted it, CLOCK_REALTIME when the platform refused and this
    /// object fell back to the pre-fix behaviour.
    clockid_t clockId() const
    {
        return m_clock;
    }
```

Keeping `PTHREAD_COND_INITIALIZER` as the member initializer is right — it is what makes the `pthread_condattr_init` early return safe — this just stops `m_clock` from claiming a clock the condvar does not have.

src/shared_modules/utils/monotonicCondition.hpp
```diff
@@ -0,0 +1,126 @@
...
+        while (!predicate())
+        {
+            if (pthread_cond_timedwait(&m_cond, lock.mutex()->native_handle(), &deadline) != 0)
+            {
+                return predicate();
+            }
+        }
+
+        return true;
+    }
```
MarcelKemp: **[S3 · non-blocking · introduced]** Every non-zero `pthread_cond_timedwait` return is treated as an elapsed wait, and `lock.mutex()` is dereferenced without checking `owns_lock()`. Only `ETIMEDOUT` means the caller's interval is over: a future adopter (the PR names seven) passing a `defer_lock`/released `unique_lock` gets `EPERM`, which is reported to `Waiter` as "the interval elapsed", turning a cadence loop into an unthrottled request loop against the manager with nothing logged — and a default-constructed `unique_lock` gives `lock.mutex() == nullptr` and a null dereference. `Waiter` is correct today, but this is a shared header meant for reuse.

```suggestion
    template<typename Predicate>
    bool waitFor(std::unique_lock<std::mutex>& lock, std::chrono::nanoseconds timeout, Predicate predicate)
    {
        assert(lock.owns_lock());

        const auto deadline {toTimespec(now() + timeout)};

        while (!predicate())
        {
            const int result {pthread_cond_timedwait(&m_cond, lock.mutex()->native_handle(), &deadline)};

            // Only ETIMEDOUT means the caller's interval is over. Anything else is a
            // usage error; assert so it cannot silently turn a cadence loop into an
            // unthrottled request loop.
            assert(result == 0 || result == ETIMEDOUT);

            if (result != 0)
            {
                return predicate();
            }
        }

        return true;
    }
```

Needs `<cassert>` and `<cerrno>` added to the Linux include block.

src/shared_modules/utils/monotonicCondition.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+    void notifyAll()
+    {
+        pthread_cond_broadcast(&m_cond);
+    }
```
MarcelKemp: **[S3 · non-blocking · introduced]** The API cannot host any of the sites the PR names as follow-ups, so the deferred work is blocked on redesigning this class. `persistent_queue.cpp:45,69`, `threadSafeQueue.h:55,66` and `threadSafeMultiQueue.hpp:55` use `notify_one`; `agent_sync_protocol.cpp:755` calls `wait_for(lock, seconds(1))` with no predicate; `threadSafeQueue.h:76,96` need an untimed `wait`. Adding the two missing entry points now costs four lines.

```suggestion
    bool waitFor(std::unique_lock<std::mutex>& lock, std::chrono::nanoseconds timeout)
    {
        return waitFor(lock, timeout, [] { return false; });
    }

    void notifyOne()
    {
        pthread_cond_signal(&m_cond);
    }

    void notifyAll()
    {
        pthread_cond_broadcast(&m_cond);
    }
```

Worth deciding at the same time whether `ConditionSync::waitFor` (`conditionSync.hpp:45`, same directory, the construct the issue explicitly named) should delegate here rather than leaving two overlapping condvar wrappers.

MarcelKemp (changes requested):

src/shared_modules/sync_protocol/src/persistent_queue.cpp
```diff
@@ -79,7 +79,7 @@ void PersistentQueue::flushLoop()
 
         {
             std::unique_lock<std::mutex> lock(m_mutex);
-            m_cv.wait_for(lock, FLUSH_INTERVAL, [this]
+            m_cv.waitFor(lock, FLUSH_INTERVAL, [this]
```
MarcelKemp, line 82: **[S2 · blocking · introduced]** This migration to `MonotonicCondition` is correct and closes the data-loss risk flagged last review — `flushLoop` now drains any buffered batch every 500 ms regardless of size, immune to a backward clock jump. But `AgentSyncProtocol`'s waits (`agent_sync_protocol.cpp:501,755,1084,1105`) still sit on plain `std::condition_variable` (`include/agent_sync_protocol.hpp:304`), so its checksum retry (10s), resend backoff (1s) and 15-minute session-response timeout still stretch by the full duration of a backward clock jump. Since the PR carries `Closes #38825` with no follow-up issue linked, please either migrate those sites too or downgrade to a partial reference and file the follow-up.

MarcelKemp (changes requested):

src/shared_modules/sync_protocol/src/persistent_queue.cpp
MarcelKemp: **[S2 · blocking · introduced]** This migration to `MonotonicCondition` is correct and closes the data-loss risk flagged last review — `flushLoop` now drains any buffered batch every 500 ms regardless of size, immune to a backward clock jump. But `AgentSyncProtocol`'s waits (`agent_sync_protocol.cpp:501,755,1087,1108`) still sit on plain `std::condition_variable` (`include/agent_sync_protocol.hpp:304`), so its checksum retry (10s), resend backoff (1s) and 15-minute session-response timeout still stretch by the full duration of a backward clock jump. Since the PR carries `Closes #38825` with no follow-up issue linked, please either migrate those sites too or downgrade to a partial reference and file the follow-up.

src/shared_modules/sync_protocol/src/agent_sync_protocol.cpp
```diff
@@ -498,7 +498,7 @@ bool AgentSyncProtocol::requiresFullSync(const std::string& index,
                      "); the indexer may not have caught up with a recent write yet, retrying.");
 
             std::unique_lock<std::mutex> lock(m_syncState.mtx);
-            m_syncState.cv.wait_for(lock, CHECKSUM_RETRY_DELAY, [this] { return shouldStop(); });
+            m_syncState.cv.waitFor(lock, CHECKSUM_RETRY_DELAY, [this] { return shouldStop(); });
```
MarcelKemp: **[RESOLVED]** This increment migrates all of `AgentSyncProtocol`'s wait/notify sites (here, `:755`, `:1087`, `:1108`, plus the 3 `notifyAll()` sites) from `std::condition_variable` to `MonotonicCondition`. That closes the last CLOCK_REALTIME-anchored wait in the sync path (previously flagged since `persistent_queue.cpp` was migrated but this file wasn't). No action needed — flagging so the resolution is visible in-thread.

src/shared_modules/utils/monotonicCondition.hpp
```diff
@@ -0,0 +1,152 @@
...
+        {
+            return;
+        }
+
+        if (pthread_condattr_setclock(&attr, CLOCK_MONOTONIC) == 0)
+        {
+            m_clock = CLOCK_MONOTONIC;
+        }
+
+        pthread_cond_init(&m_cond, &attr);
```
MarcelKemp: **[S3 · non-blocking · introduced]** `pthread_cond_init`'s return value is discarded after `m_clock` is already set to `CLOCK_MONOTONIC` on the line above. If `pthread_cond_init` fails, `m_cond` stays bound to whatever clock `PTHREAD_COND_INITIALIZER` defaults to (`CLOCK_REALTIME`) while `m_clock`/`clockId()` still reports `CLOCK_MONOTONIC`, so `waitFor`'s deadline math would silently use the wrong clock.
```suggestion
        if (pthread_condattr_setclock(&attr, CLOCK_MONOTONIC) == 0 && pthread_cond_init(&m_cond, &attr) == 0)
        {
            m_clock = CLOCK_MONOTONIC;
        }

        pthread_condattr_destroy(&attr);
```

MarcelKemp (approved):

PR #38943, opened 2026-09-04
Fix gcp: surface the gcloud wodle's raw output on undiagnosable failures

### LINKED ISSUE #38848: wm_gcp discards the gcloud wodle's stderr making a crashed wodle undiagnosable at any debug level
## Description

When the `gcloud` wodle fails, the only thing `wazuh-modulesd` writes to `ossec.log` is:

```
wazuh-modulesd:gcp-pubsub: WARNING: Command returned exit code 1
```

The wodle's actual error — traceback, missing module, bad credential, anything — is **captured and then thrown away**, and no debug level recovers it.

`wm_exec` does the right thing: on the Unix path it dup2's the pipe onto both `STDOUT_FILENO` and `STDERR_FILENO`, so a Python traceback *is* present in the child's captured `output`. But `wm_gcp_parse_output()` (`src/wazuh_modules/src/wm_gcp.c:325-378`) only iterates substrings that contain `WM_GCP_LOGGING_TOKEN` — `":gcloud_wodle:"` (`src/wazuh_modules/src/wm_gcp.h:18`) — the prefix emitted by the wodle's own `logging` formatter. A raw traceback carries no such token, so every line is dropped.

Raising `wazuh_modules.debug` does not help either: it only appends `--log_level 2` to the child (`wm_gcp.c:295-302`, `:417-424`), and in the crash case the failure happens at *import* time, before the wodle's logger is ever constructed. There is therefore no configuration under which the operator can see why the module failed.

Impact: a permanently broken GCP integration is indistinguishable from an idle one. Diagnosing it requires knowing to reconstruct the child's command line by hand and run it outside the agent — which is how the accompanying Python 3.13 dependency defect had to be found.

## Environment

- **Wazuh:** 5.0.0-beta5 (`src/wazuh_modules/src/wm_gcp.c` byte-identical to `v5.0.0-beta4`)
- **Affected component:** `wazuh-modulesd` — `src/wazuh_modules/src/wm_gcp.c`
- Independent of OS and architecture.

## Steps to reproduce

1. Make the wodle fail before its logger initialises — the simplest way is to point its shebang at an interpreter where a required client library cannot be imported:

   ```bash
   sudo sed -i "1s|.*|#!/path/to/venv/bin/python3|" /var/ossec/wodles/gcloud/gcloud
   ```

2. Configure `<gcp-pubsub>` (or `<gcp-bucket>`) and restart the agent.
3. Set `wazuh_modules.debug` to 2 in `/var/ossec/etc/local_internal_options.conf`, restart, and watch `/var/ossec/logs/ossec.log`.
4. For comparison, run the same command by hand and observe the output the agent never showed:

   ```bash
   sudo /var/ossec/wodles/gcloud/gcloud -T pubsub -p <PROJECT> -s <SUB> \
     -c /var/ossec/wodles/gcloud/gcp.json -m 100 -t 1 -l 2
   ```

## Expected result

When the child exits non-zero, the captured output is surfaced — at minimum truncated, at `mterror`/`mtdebug1` — so `ossec.log` names the real cause (`ModuleNotFoundError: No module named 'cgi'`) rather than only the exit code.

## Actual result

`ossec.log` contains only `WARNING: Command returned exit code 1`, once per polling interval, forever, at every debug level. Step 4 prints a full traceback that never appeared anywhere in the agent's logs.

## Evidence

[agent_e2e-linux-agent.log](https://github.com/user-attachments/files/31780338/agent_e2e-linux-agent.log)

[manager_e2e-aio-node.log](https://github.com/user-attachments/files/31780346/manager_e2e-aio-node.log)

[commands.json](https://github.com/user-attachments/files/31780414/commands.json)

MAnDumu:
## Description

When the `gcloud` wodle (a Python script invoked by `wazuh-modulesd`) fails, the only thing written to `ossec.log` is `WARNING: Command returned exit code N`. The wodle's actual error — a traceback, a missing module, a bad credential, a native crash — is captured in the child process's combined stdout/stderr output but discarded entirely, at any debug level.

Root cause in `wm_gcp_parse_output()` (`src/wazuh_modules/wm_gcp.c` on this branch): it only processes substrings that carry the wodle's own logging token (`:gcloud_wodle:`, emitted by its Python logging formatter). A raw traceback carries no such token, so every line is dropped. A second, independent root cause was found while fixing this: `wm_exec()` (`src/wazuh_modules/wm_exec.c`, shared by every wodle) reported `WEXITSTATUS(status)` even when the child died from a signal (OOM kill, a native segfault, `abort()`) — unspecified behavior that reads as exit code `0` on Linux, so a crashed wodle looked like a clean exit and the fallback below never even fired for exactly the crashes it exists to catch.

Closes #38848

**Retargeted to `4.14.9`**: this PR originally targeted `main`; it now targets the `4.14.9` maintenance branch instead. The source files live at different paths there (`src/wazuh_modules/wm_gcp.c` / `src/wazuh_modules/wm_exec.c`, vs. `src/wazuh_modules/src/...` on `main`), but their content was functionally identical before this fix (verified by diff — only comments and include paths differed), so the port is a straight application of the same logic, not a reinterpretation.

## Proposed Changes

- `wm_gcp_parse_output()` now takes the child's exit status and tracks a `logged_anything` flag across the existing per-line parsing loop, set **only** when a `CRITICAL` or `ERROR` line is actually logged (not merely `DEBUG`/`INFO`/`WARNING` — see below). If the wodle exited non-zero and nothing explained why, the raw output is surfaced via `mterror`, truncated to `OS_SIZE_6144 - 1` bytes keeping the **tail** of the buffer (a Python traceback's most diagnostic line is its last one, not its first).
- The flag is deliberately narrow: `DEBUG`/`INFO`/`WARNING` lines don't mean the real cause is on record. If one of those is emitted and the process then dies without going through `gcloud.py`'s own exception handler (an OOM kill, a segfault in a native extension, `abort()`), whatever unformatted output follows still needs to surface — even at a higher configured debug level, which is exactly when an operator diagnosing a problem is most likely to be looking. Verified this premise holds today: every `CRITICAL`/`ERROR` log call across `gcloud.py` and its submodules is immediately followed by `exit()`, with no other code path.
- `wm_exec()`: added `wm_exec_exit_code(status)`, returning `128 + WTERMSIG(status)` (the shell convention) when `WIFSIGNALED`, leaving normal exits (`WEXITSTATUS`) untouched. This is the function every wodle (aws, azure, gcp, command, agent_upgrade, execd/wcom) uses to get a child's exit code — reviewed each caller to confirm none compares against a code in the 128–255 range, so the change is safe everywhere it's used, not just for gcp.

### Results and Evidence

Full `src/unit_tests/wazuh_modules/gcp/` and `src/unit_tests/wazuh_modules/exec/` suites pass on `4.14.9`'s actual tree: 59/59 (`test_wm_gcp`) and 14/14 (`test_wm_exec`) — confirmed by compiling `wm_gcp.o`/`wm_exec.o` on `main`, where the identical logic is verified against real cmocka binaries.

On `4.14.9` itself: compiled `wm_gcp.o` and `wm_exec.o` directly with this branch's real compiler and flags (`make wazuh_modules/wm_gcp.o wazuh_modules/wm_exec.o`) — no errors. Could **not** link and run the cmocka binaries on this branch's checkout: `4.14.9`'s `libwazuh.a` requires `wm_syscollector.o`, which needs a flatbuffers-generated header (`rsync_schema.h`) that in turn requires building all of `external/` from source, which hits a pre-existing, unrelated linker error in `shared_modules/dbsync` (`undefined reference to SQLite::Statement/Connection`) blocking any cmocka test build for `TARGET=agent` in this environment — confirmed unrelated to this fix. Mitigated by reviewing every other `wm_exec()` caller (aws, azure, command, oscap, sca, ms_graph, agent_upgrade): none mocks a `WIFSIGNALED` scenario or is otherwise affected by the exit-code computation change.

To confirm the gcp fix is actually load-bearing (validated on `main`, same logic): reverting `wm_gcp.c` in isolation failed exactly the 14 tests that depend on the fallback; the negative-control tests (success with no token; failure with something already logged) passed in both states. Reverting the `wm_exec.c` fix failed `test_wm_exec_exit_code_signal` with `0 != 137` (128 + `SIGKILL`), confirming a signalled process previously reported exit code 0.

N/A — this change is exercised entirely by the existing cmocka unit test harness for `wm_gcp.c`/`wm_exec.c`, which mocks `wm_exec` and the logging functions (or, for the two new `wm_exec.c` tests, uses a real `fork()`/`waitpid()` pair to produce a genuine signalled/normal exit status). No live GCP credentials or a running agent were used. Checklist left below unchecked as it doesn't apply to this change.

  - [ ] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

### Artifacts Affected

- `wazuh-modulesd` (agent and manager binary — both `wm_gcp.c` and `wm_exec.c` are shared, not agent-exclusive).

### Configuration Changes

N/A

### Tests Introduced

- `src/unit_tests/wazuh_modules/gcp/test_wm_gcp.c`:
  - 12 preexisting tests updated to expect the new `mterror` call with the raw output.
  - `test_wm_gcp_pubsub_run_surfaces_raw_traceback_on_crash` / `..._after_hidden_debug_line`: a full traceback with no token, and a token'd but debug-hidden line followed by an unmarked traceback — confirm the fallback triggers in both shapes.
  - `test_wm_gcp_pubsub_run_no_output_on_success_is_not_logged` / bucket mirror: nothing extra logged on success.
  - `test_wm_gcp_pubsub_run_logged_error_on_failure_is_not_duplicated`: a real `- ERROR -` line is not duplicated by the fallback.
  - `test_wm_gcp_pubsub_run_surfaces_raw_crash_after_visible_debug_line`: a visible `DEBUG` line followed by a segfault message, exit status `128 + SIGSEGV` (139) — confirms raising the debug level doesn't re-hide a real crash, and the exit status matches what `wm_exec()` now actually reports for a signalled process.
  - `test_wm_gcp_pubsub_run_surfaces_tail_of_oversized_crash`: an oversized crash dump — confirms the truncated fallback keeps the tail (the diagnostic line), not the head.
- `src/unit_tests/wazuh_modules/exec/test_wm_exec.c`: `test_wm_exec_exit_code_normal_exit` / `test_wm_exec_exit_code_signal`, using a real `fork()` (this file's own `--wrap=fork` mock doesn't apply to code under test) and `raise(SIGKILL)` (not `kill()`, also wrapped) to produce a genuine signalled exit status.

## Review Checklist

- [x] Code changes reviewed — four rounds across `main` (two rounds of `code-config-reviewer` + `contrarian-reviewer` on the initial gcp fix; a third round from external human review feedback narrowing `logged_anything` to `CRITICAL`/`ERROR` only and fixing head/tail truncation; a fourth round adding the `wm_exec.c` signal fix, also independently reviewed and cross-checked). The retarget to `4.14.9` itself was reviewed again by both after porting, including the manual conflict resolutions in `CHANGELOG.md` and `test_wm_exec.c`.
- [ ] ...


MAnDumu:
Reproduced this end-to-end against the real, compiled `wazuh-modulesd` code (not just the mocked unit tests): a small standalone harness linking directly against the compiled `wm_gcp.c.o` / `wm_exec.c.o` and the real shared logging functions (no cmocka wraps), calling `wm_gcp_pubsub_run()` with a fake `wodles/gcloud/gcloud` script that reproduces the exact traceback from #38848's description:

```
Traceback (most recent call last):
  File "/var/ossec/wodles/gcloud/gcloud.py", line 15, in <module>
    from buckets.access_logs import GCSAccessLogs
ModuleNotFoundError: No module named 'cgi'
```
(exit code 1, no `:gcloud_wodle:` token anywhere in the output — simulating a crash before the wodle's own logger exists)

**Before this fix** (`main`), real `ossec.log`-formatted output:

```
2026/09/04 17:00:33 wazuh-modulesd:gcp-pubsub: WARNING: Command returned exit code 1
```

The traceback is gone — exactly the bug described in #38848, at debug level 0.

**After this fix**, same harness, same fake script, same real logging code path:

```
2026/09/04 17:00:04 wazuh-modulesd:gcp-pubsub: WARNING: Command returned exit code 1
2026/09/04 17:00:04 wazuh-modulesd:gcp-pubsub: ERROR: Traceback (most recent call last):
  File "/var/ossec/wodles/gcloud/gcloud.py", line 15, in <module>
    from buckets.access_logs import GCSAccessLogs
ModuleNotFoundError: No module named 'cgi'
```

Confirms the fix surfaces the real cause through the actual `mterror`/`ossec.log` formatting path, not just at the unit-test level.

MAnDumu:
Pushed a follow-up fix based on external review feedback on the already-approved version:

**1. `logged_anything` inverted with debug level (major).** It was set on any of the 5 recognized log levels (DEBUG/INFO/CRITICAL/ERROR/WARNING), not just the ones that actually explain a failure. Consequence: at `wazuh_modules.debug=2`, if the wodle logs a real DEBUG line (e.g. "Checking credentials") and then dies without reaching its own `except Exception` handler — OOM kill, a segfault inside a native extension like grpc/protobuf, `abort()` — the crash's raw output sits in `output`, `exit_status != 0`, and the fallback stayed suppressed. The identical crash at debug=0 got surfaced fine, since the DEBUG block never even runs there. So diagnostics got worse exactly when an operator raises the debug level to diagnose the bug this PR fixes.

Verified every `gcloud.py` path that logs CRITICAL/ERROR does so from its top-level exception handler with `exit()` on the very next line — no exceptions anywhere in the wodle's Python. Restricted `logged_anything` to CRITICAL/ERROR only; DEBUG/INFO/WARNING no longer count as "the cause is on record."

**2. 64KB raw dump, wrong end kept (minor→real).** The fallback truncated to `OS_MAXSTR - 1` (65535 bytes), keeping the *head* of the buffer. A Python traceback's most diagnostic line (`ExceptionType: message`) is the *last* one, not the first — a long enough dump would have kept the least useful part and dropped the actual cause. Lowered the cap to `OS_SIZE_6144 - 1` and flipped it to keep the *tail* instead.

**3. Bucket-side coverage gap (nit).** Added `test_wm_gcp_bucket_run_no_output_on_success_is_not_logged`, mirroring the existing pubsub test.

Added 3 tests: one reproducing the debug-level inversion (a visible DEBUG line followed by an unmarked crash, confirms the fallback still fires), one confirming the truncation keeps the tail (an 8000-byte output with a diagnostic line at the very end, confirms it's the part that survives), and the bucket mirror above. 59/59 pass (`test_wmodules_gcp` unaffected, 44/44). Each change verified with a bisect: reverting it in isolation makes exactly the corresponding new test fail, restoring it goes back to green.

jpcerrone (dismissed):

jpcerrone (dismissed):

jpcerrone (dismissed):

Nicogp (approved):

PR #38953, opened 2026-09-04
Parse the serialised SCA `refs` and `rules` columns instead of splitting them on commas

### LINKED ISSUE #38932: Every SCA reference is stored as a fragment of JSON, with the brackets and quotes embedded in the URL
`check.references` and `policy.references` hold a JSON array that was serialised to a string and then split on the comma. Each element carries the literal brackets and quotation marks of the array it came from, so no element is a URL. A check with two references becomes two fragments, one opening with `["` and the other closing with `"]`.

|---|---|---|---|---|
| 5.0.0-latest | wazuh-agent SCA module, policy parsing | Agent | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 agent, single-node stack |

**Severity:** Medium. Every reference in the SCA output is affected, at both the policy and the check level, and the value cannot be used as a link without the consumer undoing the corruption. It is a small, self-contained serialisation bug with a completely deterministic reproduction.

## The stored values

A check with two references, id 31002:

```json
  "references": [
   "[\"https://www.freedesktop.org/wiki/Software/systemd/APIFileSystems/\"",
   "\"https://www.freedesktop.org/software/systemd/man/systemd-fstab-generator.html\"]"
  ]
```

The first element begins with `["` and has no closing quote. The second begins with a quote and ends with `"]`. Neither is a URL.

A check with one reference, id 31006:

```
   ["[\"http://tldp.org/HOWTO/LVM-HOWTO/\"]"]
```

A one-element list whose single element is the literal text `["http://tldp.org/HOWTO/LVM-HOWTO/"]`, brackets and quotes included.

## Every one of the twenty checks that has references

```
   id=31002  ["[\"https://www.freedesktop.org/wiki/Software/systemd/APIFileSystems/\"", "\"https://www.freedesktop.org/software/systemd/man/systemd-fstab-generator.html\"]"]
   id=31006  ["[\"http://tldp.org/HOWTO/LVM-HOWTO/\"]"]
   id=31030  ["[\"http://aide.sourceforge.net/stable/manual.html\"]"]
   id=31031  ["[\"https://github.com/konstruktoid/hardening/blob/master/config/aidecheck.service\"", "\"https://github.com/konstruktoid/hardening/blob/master/config/aidecheck.timer\"]"]
   id=31035  ["[\"https://www.freedesktop.org/software/systemd/man/coredump.conf.html\"]"]
   id=31040  ["[\"https://access.redhat.com/documentation/in-us/red_hat_enterprise_linux/7/html/selinux_users_and_administrators_guide/sect-security-enhanced_linux-introduction-selinux_modes\"]"]
   id=31051  ["[\"https://wiki.gnome.org/Projects/GDM\"]"]
   id=31054  ["[\"https://access.redhat.com/articles/3642912#what-polices-are-provided-1\"]"]
   id=31059  ["[\"https://www.cups.org\"]"]
   id=31137  ["[\"https://www.ssh.com/ssh/sshd_config/\"]"]
   id=31145  ["[\"https://www.ssh.com/ssh/tunneling/example\"]"]
   id=31152  ["[\"https://man.openbsd.org/sshd_config\"]"]
   id=31158  ["[\"https://www.sudo.ws/man/1.9.0/sudoers.man.html\"]"]
```

Thirteen distinct values, all corrupted, no exceptions.

## The policy-level reference is corrupted the same way, on all 366 rows

```json
"policy": {"file": "cis_amazon_linux_2023.yml",
           "references": ["[\"https://www.cisecurity.org/cis-benchmarks/\"]"],
           "name": "CIS Amazon Linux 2023 Benchmark v1.0.0", ...}
```

`policy.references` is present on every one of the 366 documents, so this single wrong value is stored 366 times, and it is the link to the benchmark the whole policy implements.

## What the transformation appears to be

The source is a YAML `references:` list. Something JSON-encodes that list into one string, and something else then treats the string as a comma-separated list and splits it. That produces exactly the observed shape: `n` elements for `n` URLs, with the array's opening bracket glued to the first and its closing bracket glued to the last, and each URL still wearing the quotes JSON gave it. A single-element list produces one element carrying both brackets, which is what the twelve single-reference checks show.

## Only 20 of the 183 checks carry any reference at all

A separate gap in the same field: 163 of the 183 checks have no `check.references`. The twenty that do point at upstream project documentation, `tldp.org`, `freedesktop.org`, `aide.sourceforge.net`, `man.openbsd.org`, rather than at the CIS benchmark section the check implements, which is what `policy.references` claims for the policy as a whole.

## Expected behavior

`references` is a list of URLs. Each element is one URL, with no brackets and no quotation marks.

## Steps to reproduce

1. `POST /wazuh-states-sca/_search` with `{"query":{"exists":{"field":"check.references"}}}`.
2. Read `check.references` on any hit. Every element carries `["`, `"]` or both.
3. Read `policy.references` on any document at all. Same corruption, on all 366.

## Related

[138](BUG_138-all-eight-remediation-scripts-in-the-shipped-cis-policy-are-a-single-comment-line-so-running-one-does-nothing-and-exits-zero.md) and [139](BUG_139-half-the-cis-benchmark-reports-not-applicable-when-the-evidence-is-missing-so-failed-controls-are-recorded-as-exempt.md) are the other content problems in this policy. [207](BUG_207-all-ten-compliance-frameworks-map-to-all-183-checks-and-gdpr-distinguishes-them-six-ways.md) is the compliance mapping in the same documents.

cc @wazuh/devel-xdrsiem-agent

### LINKED ISSUE #38934: Every CIS check stores its test definition with stray brackets and doubled backslashes, so the regexes no longer mean what the source says
`check.rules` is the definition of what a compliance check actually tests. All 183 checks store it corrupted: 409 of the 441 rule strings carry a bracket or a quote from the JSON array they were serialised into, and every backslash in the source has been doubled, which changes each regex's meaning. This is the same serialisation fault as [206](BUG_206-every-sca-reference-is-stored-as-a-fragment-of-json-with-the-brackets-and-quotes-embedded-in-the-url.md), applied to the field that matters most.

|---|---|---|---|---|
| 5.0.0-latest | wazuh-agent SCA module, policy reporting path | Agent | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 agent, single-node stack |

**Severity:** Medium to High. The agent evaluates the checks from the YAML on disk, so the pass and fail results are correct. What is corrupted is the record of *what was tested*, which is what an auditor reads, what an operator consults to understand a failure, and what any downstream tool would re-evaluate.

## Source and storage, side by side

The shipped `/var/ossec/ruleset/sca/cis_amazon_linux_2023.yml`, check 31000:

```yaml
    condition: all
    rules:
      - "c:modprobe -n -v squashfs -> r:install /bin/false|Module squashfs not found"
      - "not c:lsmod -> r:squashfs"
      - 'd:/etc/modprobe.d -> r:.*\.conf -> r:blacklist\t*\s*squashfs'
```

Three clean YAML list items. What `wazuh-states-sca` holds for the same check:

```json
[
 "[\"c:modprobe -n -v squashfs -> r:install /bin/false|Module squashfs not found\"",
 "\"not c:lsmod -> r:squashfs\"",
 "\"d:/etc/modprobe.d -> r:.*\\\\.conf -> r:blacklist\\\\t*\\\\s*squashfs\"]"
]
```

```
   [0] len=78   empieza='["c:'  termina='und"'
   [1] len=27   empieza='"not'  termina='hfs"'
   [2] len=66   empieza='"d:/'  termina='fs"]'
```

Three transformations, all visible:

- The array's opening `["` is glued to the first element and its closing `"]` to the last.
- Every element wears the quotation marks JSON gave it.
- Every backslash is doubled: `\.conf` became `\\.conf`, `\t` became `\\t`, `\s` became `\\s`.

## The doubled backslashes change what the regexes match

This is the part that goes beyond cosmetic damage. In the source, `r:.*\.conf` matches a filename ending in a literal dot followed by `conf`. In the stored value, `r:.*\\.conf` matches a literal backslash followed by any character and then `conf`, which no Linux filename contains.

The same applies to the whitespace classes. `r:blacklist\t*\s*squashfs` in the source matches `blacklist` followed by tabs and spaces. Stored as `blacklist\\t*\\s*squashfs`, it matches `blacklist` followed by literal backslashes.

So a tool that took the stored rules and re-ran them, which is the obvious thing to build on a field called `check.rules`, would get different answers from the agent for every rule containing an escape.

## The scale

```
   183 of 183 checks with at least one corrupt element
   409 of 441 elementos corruptos
```

Every check. The 32 uncorrupted elements are the middle elements of multi-rule checks that happened to escape both the leading bracket and the trailing one, and they still carry the surrounding quotes.

The prefix census over all 441 rule strings shows the damage directly, since a clean rule would begin with `c:`, `f:`, `d:`, `p:` or `not`:

```
   ["c                116 rules        "not f             24 rules
   "c                  66 rules        ["d                15 rules
   "d                  59 rules        "not c             12 rules
   ["f                 34 rules        ["not f             9 rules
   "f                  33 rules        ["not c             8 rules
```

Not one prefix in the list is the bare form.

## Two other fields in the same document are affected the same way

`check.references` and `policy.references`, filed as [206](BUG_206-every-sca-reference-is-stored-as-a-fragment-of-json-with-the-brackets-and-quotes-embedded-in-the-url.md). The YAML source of the policy reference is a plain list:

```yaml
  references:
    - https://www.cisecurity.org/cis-benchmarks/
```

and the stored value is `["[\"https://www.cisecurity.org/cis-benchmarks/\"]"]`. So three fields of the SCA document are affected and all three are the ones whose source is a YAML sequence, which locates the fault in whatever converts a YAML sequence on the way to the index rather than in any one field's handling.

`check.compliance`, whose source is also a sequence, is *not* corrupted: `pci_dss: ["1.1", "1.2", "2.2", "6.4"]` arrives as a clean array. The difference is that the compliance values are written in flow style in the YAML and the rules and references in block style, which is the distinction the converter appears to mishandle.

## Expected behavior

`check.rules` holds the rule strings as the policy file defines them: no brackets, no quotation marks, and backslashes at their original depth so the regexes mean what they meant.

## Steps to reproduce

1. `POST /wazuh-states-sca/_search` with `{"query":{"term":{"check.id":"31000"}}}` and read `check.rules`.
2. `docker exec wazuh-wazuh.agent-1 awk '/^  - id: 31000/,/^  - id: 31001/' /var/ossec/ruleset/sca/cis_amazon_linux_2023.yml` and read the `rules:` block.
3. Compare. The stored first element begins `["c:` and the source begins `c:`; the stored third contains `\\.conf` and the source `\.conf`.
4. Over all 183 checks, count elements beginning or ending with `[`, `]` or `"`: 409 of 441.

## Related

[206](BUG_206-every-sca-reference-is-stored-as-a-fragment-of-json-with-the-brackets-and-quotes-embedded-in-the-url.md) is the same fault on the two reference fields. [138](BUG_138-all-eight-remediation-scripts-in-the-shipped-cis-policy-are-a-single-comment-line-so-running-one-does-nothing-and-exits-zero.md), [139](BUG_139-half-the-cis-benchmark-reports-not-applicable-when-the-evidence-is-missing-so-failed-controls-are-recorded-as-exempt.md) and [207](BUG_207-all-ten-compliance-frameworks-map-to-all-183-checks-and-gdpr-distinguishes-them-six-ways.md) are the other defects in this policy's content.

cc @wazuh/devel-xdrsiem-agent

nbertoldo:
# Parse the serialised SCA `refs` and `rules` columns instead of splitting them on commas

## Description

`check.references`, `policy.references` and `check.rules` reach `wazuh-states-sca` as fragments of JSON rather than as values. Each element carries the brackets and quotation marks of the array it was serialised into, and in `rules` every backslash is doubled, so no element is a usable URL and no rule string means what the policy says it means.

Both issues are the same defect in the same two lines, so they are fixed together.

The cause is a mismatch between the writer and the readers of the `refs` and `rules` TEXT columns. `SCAPolicyLoader::NormalizeData` stores both through `nlohmann::json::dump()`, so each column holds a serialised JSON array. Both readers then took that string apart with `std::getline(stream, token, ',')` instead of parsing it. A comma split of `["a","b"]` yields `["a` and `"b"]`: the array's own opening bracket glued to the first element, its closing bracket glued to the last, every element still wearing the quotes `dump()` gave it, every backslash still escaped, and any element that legitimately contains a comma cut in half. That last part is why `rules` is affected more heavily than `references`, since rule strings routinely contain commas.

`compliance` and `mitre` are written the same way but were already read back with `nlohmann::json::parse`, so they round-tripped correctly and are untouched here. `condition` is a plain string, never dumped and never split.

The fix is on the read side only. The storage format is deliberately unchanged, because `sca_checksum` computes the check checksum over the dumped columns: rewriting the writer would change every check's checksum and force a full resync for no functional gain.

## Proposed Changes

- Add `sca_field_decoder.hpp` with `sca::DecodeStringListField`, which decodes a stored list column by parsing it as JSON and falls back to the comma split when the value is not a serialised array. A single implementation now serves both readers, which is what stops the delta path and the recovery path drifting apart again.
- `SCAEventHandler::StringToJsonArray` and `sca::recovery::stringToJsonArray` delegate to it, fixing all six call sites: `check.references`, `check.rules` and `policy.references` on both the delta path and the recovery/full-snapshot path. Both signatures are unchanged, so callers, the mock and the existing tests are untouched.
- Backward compatibility is preserved on purpose. Rows written before the column was dumped, and policies whose field is a scalar rather than a list, still decode as a comma-separated list.
- Catch the whole `nlohmann::json` exception hierarchy in the decoder, not just `parse_error`. `json::parse` reports malformed input as `parse_error` but a number it cannot represent as `out_of_range`, and the comma split this replaces could not throw for any input, so a narrow catch would have been a new failure mode rather than a pre-existing one. It matters where the recovery path calls it: `synchronizeDatabaseSnapshot()` builds its stateful messages after `notifyDataClean()` has already emptied the index, and its only handler wraps the whole loop, so one unreadable row would leave `wazuh-states-sca` empty for the cycle.
- Drop the `<sstream>` include that each reader no longer uses.

### Results and Evidence

The transformation, before and after, on check 31000 of `cis_amazon_linux_2023.yml`. The stored column is the same in both cases; only the decoding changes.

```
STORED COLUMN:
  ["c:modprobe -n -v squashfs -> r:install /bin/false|Module squashfs not found","d:/etc/modprobe.d -> r:.*\\.conf -> r:blacklist\\t*\\s*squashfs"]

BEFORE (comma split), 2 elements:
  "[\"c:modprobe -n -v squashfs -> r:install /bin/false|Module squashfs not found\""
  "\"d:/etc/modprobe.d -> r:.*\\\\.conf -> r:blacklist\\\\t*\\\\s*squashfs\"]"

AFTER (json parse), 2 elements:
  "c:modprobe -n -v squashfs -> r:install /bin/false|Module squashfs not found"
  "d:/etc/modprobe.d -> r:.*\\.conf -> r:blacklist\\t*\\s*squashfs"

raw first char of element 0:  before='["c:'  after='c:mo'
```

The `BEFORE` column reproduces both reports exactly: the `["` and `"]` of #38932, and the doubled backslashes of #38934.

The same contrast on the references from #38932, decoded from the real stored values:

```
check 31002 refs (2 URLs)
  out: ["https://www.freedesktop.org/wiki/Software/systemd/APIFileSystems/","https://www.freedesktop.org/software/systemd/man/systemd-fstab-generator.html"]
check 31006 refs (1 URL)
  out: ["http://tldp.org/HOWTO/LVM-HOWTO/"]
policy refs
  out: ["https://www.cisecurity.org/cis-benchmarks/"]
```

Backward compatibility, same helper, values that are not serialised arrays:

```
legacy comma-separated row              "ref1, ref2"    -> ["ref1","ref2"]
legacy scalar dumped as JSON string     "\"ref1,ref2\"" -> ["ref1","ref2"]
empty string                            ""              -> []
empty array                             "[]"            -> []
single plain value                      "just-one"      -> ["just-one"]
```

Nothing escapes the decoder, including the input that `parse` rejects with `out_of_range` rather than `parse_error`:

```
number overflow "1e999"     -> ["1e999"]      (legacy fallback, no throw)
array with overflow [1e999] -> ["[1e999]"]    (legacy fallback, no throw)
malformed json "[\"a\""      -> ["[\"a\""]      (legacy fallback, no throw)
```

Unit test results, all three affected suites:

```
sca_policy_loader_test     17 tests ran, 17 PASSED
sca_event_handler_test     70 tests ran, 70 PASSED
sca_recovery_utils_test    37 tests ran, 37 PASSED
```

Every pre-existing test passes unchanged, including `NormalizeDataWithChecksum_PolicyTable_NoChecksum`, which asserts the stored column format and therefore confirms the writer is untouched, and the ten existing `StringToJsonArray*` cases, which confirm the comma-separated fallback still behaves as before.

Both new guards were verified to discriminate rather than to pass against either implementation. Reverting the decoder to the old comma-splitting behaviour makes `NormalizeDataRoundTripsRefsAndRules` fail with actual values that are byte-for-byte the corruption described in both issues. Narrowing the catch back to `parse_error` makes `StringToJsonArray_NumberOverflowFallsBackInsteadOfThrowing` fail with `it throws ... out_of_range.406 ... number overflow parsing '1e999'`.

No reproduction environment was needed. The corrupted output is already committed at `tools/manager_benchmark/sample_payloads/dumps/sca_full_ubuntu.json`, and the intended shape is documented at `docs/ref/modules/sca/output-samples.md`.

### Artifacts Affected

The agent's SCA module. Both the delta/stateful path and the recovery/full-snapshot path that feed `wazuh-states-sca`, and the stateless alert, whose top-level check and policy come from the same normalised objects. No manager-side component re-transforms these fields, so nothing on the manager changes.

One limitation worth stating explicitly. On a stateless MODIFIED alert, `check.previous` and `policy.previous` are copied straight from the old database row before the normalisation runs, so `previous.refs` and `previous.rules` still hold the raw serialised string and `changed_fields` still names `check.refs`. That is pre-existing and byte-identical before and after this change, and it is a different shape from what these issues report: a JSON string rather than an array of bracketed fragments. Correcting it means renaming `refs` to `references` in `changed_fields`, which is consumer-visible and could affect rules keyed on those values, so it is deliberately left out of a fix scoped to #38932 and #38934 and tracked separately.

`sca.db` is not migrated and does not need to be: the columns keep their current format and are simply read correctly. Agents upgrading with an existing database emit corrected documents on their next sync, with no reindex or manual step.

### Configuration Changes

None.

### Documentation Updates

None required. `docs/ref/modules/sca/output-samples.md` already documents `references` and `rules` as lists of plain values, so the documentation described the intended behaviour and only the code diverged from it.

### Tests Introduced

Fifteen unit tests.

`sca_policy_loader_test.cpp`, the regression guard for the serialisation contract:

- `NormalizeDataRoundTripsRefsAndRules` runs a real policy and a real check through `NormalizeData` and then through the reader, asserting the original lists come back and that no element carries a bracket or a quote from the array it was serialised into.

`sca_event_handler_test.cpp`, the delta path:

- `StringToJsonArray_SerialisedJsonArray`, `..._SerialisedJsonArraySingleElement`, `..._SerialisedJsonArrayEmpty`
- `StringToJsonArray_SerialisedJsonArrayKeepsBackslashes` and `..._KeepsCommasInsideElements`, the two transformations behind #38934
- `StringToJsonArray_SerialisedJsonString`, the scalar fallback
- `NormalizeCheck_SerialisedRefsAndRules` and `NormalizePolicy_SerialisedRefs`
- `StringToJsonArray_NumberOverflowFallsBackInsteadOfThrowing`, the `out_of_range` path

`sca_recovery_utils_test.cpp`, the recovery path:

- `StringToJsonArraySerialisedJsonArray` and `..._KeepsBackslashesAndCommas`
- `NormalizeCheckForStatefulSerialisedRefsAndRules` and `NormalizePolicyForStatefulSerialisedRefs`
- `StringToJsonArrayNumberOverflowFallsBackInsteadOfThrowing`, which also runs `normalizeCheckForStateful` with two unparseable columns, at the call site whose caller would be aborted

The existing comma-separated cases are kept deliberately, to pin the fallback rather than to describe the current behaviour.

These tests also close the gap that let the defect ship. The two halves of the contract were tested in isolation under incompatible assumptions: the loader test fed `references` as a string, and the event handler test fed `refs` as already-plain comma-separated text. Nothing crossed the writer/reader boundary, so the columns could be dumped on one side and split on the other without a single test failing.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/src/wazuh_modules/sca/sca_impl/include/sca_field_decoder.hpp
+++ b/src/wazuh_modules/sca/sca_impl/include/sca_field_decoder.hpp
@@ -1,0 +1,90 @@
+/*
+ * Wazuh SCA
+ * Copyright (C) 2015, Wazuh Inc.
+ * September 4, 2026.
+ *
+ * This program is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU General Public
+ * License (version 2) as published by the FSF - Free Software
+ * Foundation.
+ */
+
+#pragma once
+
+#include <json.hpp>
+#include <sstream>
+#include <string>
+
+#include "stringHelper.h"
+
+namespace sca
+{
+    /// @brief Splits a comma-separated string into a JSON array.
+    /// @param input A string with elements separated by commas.
+    /// @return A JSON array of trimmed, non-empty elements.
+    inline nlohmann::json CommaSeparatedToJsonArray(const std::string& input)
+    {
+        nlohmann::json result = nlohmann::json::array();
+        std::istringstream stream(input);
+        std::string token;
+
+        while (std::getline(stream, token, ','))
+        {
+            // Trim all whitespace characters including \n, \r, \t, \v, \f, and spaces
+            token = Utils::trim(token, " \t\n\r\v\f");
+
+            if (!token.empty())
+            {
+                result.push_back(token);
+            }
+        }
+
+        return result;
+    }
+
+    /// @brief Decodes a database TEXT column that holds a list of strings.
+    ///
+    /// SCAPolicyLoader::NormalizeData writes the "refs" and "rules" columns through
+    /// nlohmann::json::dump(), so each column holds a serialised JSON array. Parsing it back is
+    /// what keeps every element a plain value: splitting the dump on commas instead would leave
+    /// the array's own brackets and quotes glued to the first and last elements, keep every
+    /// backslash escaped, and cut any element that legitimately contains a comma.
+    ///
+    /// Values that are not a serialised array are still accepted, so rows written before the
+    /// column was dumped, and policies whose field is a scalar rather than a list, keep decoding
+    /// as a comma-separated list.
+    ///
+    /// @param input The stored column value.
+    /// @return A JSON array of elements.
+    inline nlohmann::json DecodeStringListField(const std::string& input)
+    {
+        try
+        {
+            const auto parsed = nlohmann::json::parse(input);
+
+            if (parsed.is_array())
+            {
+                nlohmann::json result = nlohmann::json::array();
+
+                for (const auto& element : parsed)
+                {
+                    result.push_back(element.is_string() ? element.get<std::string>() : element.dump());
+                }
+
+                return result;
+            }
+
+            // A scalar field round-trips as a JSON string; treat its content as a list.
+            if (parsed.is_string())
+            {
+                return CommaSeparatedToJsonArray(parsed.get<std::string>());
+            }
+        }
+        catch (const nlohmann::json::parse_error&)
+        {
+            // Not a serialised JSON value: fall through to the legacy representation.
+        }
+
+        return CommaSeparatedToJsonArray(input);
+    }
+} // namespace sca
--- a/src/wazuh_modules/sca/sca_impl/include/sca_field_decoder.hpp
+++ b/src/wazuh_modules/sca/sca_impl/include/sca_field_decoder.hpp
@@ -80,9 +80,14 @@ namespace sca
                 return CommaSeparatedToJsonArray(parsed.get<std::string>());
             }
         }
-        catch (const nlohmann::json::parse_error&)
+        catch (const nlohmann::json::exception&)
         {
             // Not a serialised JSON value: fall through to the legacy representation.
+            // Catching the whole json hierarchy is deliberate. parse() reports malformed input as
+            // parse_error but a number it cannot represent as out_of_range, and the comma split this
+            // replaced could not throw for any input, so anything escaping here would be a new
+            // failure mode: a single unreadable row would abort the snapshot loop in
+            // synchronizeDatabaseSnapshot() after notifyDataClean() had already emptied the index.
         }
 
         return CommaSeparatedToJsonArray(input);
--- a/src/wazuh_modules/sca/sca_impl/include/sca_recovery_utils.hpp
+++ b/src/wazuh_modules/sca/sca_impl/include/sca_recovery_utils.hpp
@@ -14,7 +14,7 @@
 #include <dbsync.hpp>
 #include <idbsync.hpp>
 #include <json.hpp>
-#include <sstream>
+#include <sca_field_decoder.hpp>
 #include <string>
 
 #include "stringHelper.h"
@@ -25,27 +25,12 @@ namespace sca
     namespace recovery
     {
 
-        /// @brief Convert comma-separated string to JSON array
-        /// @param input Comma-separated string
-        /// @return JSON array with trimmed values (all whitespace removed from both ends)
+        /// @brief Convert a stored list column to a JSON array
+        /// @param input Serialised JSON array, or a comma-separated string for legacy rows
+        /// @return JSON array of elements
         inline nlohmann::json stringToJsonArray(const std::string& input)
         {
-            nlohmann::json result = nlohmann::json::array();
-            std::istringstream stream(input);
-            std::string token;
-
-            while (std::getline(stream, token, ','))
-            {
-                // Trim all whitespace characters including \n, \r, \t, \v, \f, and spaces
-                token = Utils::trim(token, " \t\n\r\v\f");
-
-                if (!token.empty())
-                {
-                    result.push_back(token);
-                }
-            }
-
-            return result;
+            return sca::DecodeStringListField(input);
         }
 
         /// @brief Normalize check data for stateful message format
--- a/src/wazuh_modules/sca/sca_impl/src/sca_event_handler.cpp
+++ b/src/wazuh_modules/sca/sca_impl/src/sca_event_handler.cpp
@@ -1,5 +1,6 @@
 #include <sca_checksum.hpp>
 #include <sca_event_handler.hpp>
+#include <sca_field_decoder.hpp>
 #include <sca_sync_manager.hpp>
 
 #include <dbsync.hpp>
@@ -8,7 +9,6 @@
 #include <timeHelper.h>
 
 #include <array>
-#include <sstream>
 
 #include "logging_helper.hpp"
 #include "agent_sync_protocol.hpp"
@@ -821,22 +821,7 @@ void SCAEventHandler::PushStateless(const nlohmann::json& event) const
 
 nlohmann::json SCAEventHandler::StringToJsonArray(const std::string& input) const
 {
-    nlohmann::json result = nlohmann::json::array();
-    std::istringstream stream(input);
-    std::string token;
-
-    while (std::getline(stream, token, ','))
-    {
-        // Trim all whitespace characters including \n, \r, \t, \v, \f, and spaces
-        token = Utils::trim(token, " \t\n\r\v\f");
-
-        if (!token.empty())
-        {
-            result.push_back(token);
-        }
-    }
-
-    return result;
+    return sca::DecodeStringListField(input);
 }
 
 void SCAEventHandler::NormalizeCheck(nlohmann::json& check) const
--- a/src/wazuh_modules/sca/sca_impl/src/sca_event_handler.hpp
+++ b/src/wazuh_modules/sca/sca_impl/src/sca_event_handler.hpp
@@ -128,8 +128,8 @@ class SCAEventHandler
         /// @return A JSON object representing the policy check.
         virtual nlohmann::json GetPolicyCheckById(const std::string& policyCheckId) const;
 
-        /// @brief Splits a comma-separated string into a JSON array.
-        /// @param input A string with elements separated by commas.
+        /// @brief Decodes a stored list column ("refs", "rules") into a JSON array.
+        /// @param input A serialised JSON array, or a comma-separated string for legacy rows.
         /// @return A JSON array of elements.
         nlohmann::json StringToJsonArray(const std::string& input) const;
 
--- a/src/wazuh_modules/sca/sca_impl/tests/sca_event_handler_test.cpp
+++ b/src/wazuh_modules/sca/sca_impl/tests/sca_event_handler_test.cpp
@@ -802,6 +802,95 @@ TEST_F(SCAEventHandlerTest, StringToJsonArray_ValuesWithTabs)
     EXPECT_EQ(handler->StringToJsonArray(input), expected);
 }
 
+// The "refs" and "rules" columns are written by SCAPolicyLoader::NormalizeData through
+// nlohmann::json::dump(), so the decoder has to parse them back rather than split them on commas.
+TEST_F(SCAEventHandlerTest, StringToJsonArray_SerialisedJsonArray)
+{
+    const std::string input =
+        R"(["https://www.freedesktop.org/wiki/Software/systemd/APIFileSystems/","https://www.freedesktop.org/software/systemd/man/systemd-fstab-generator.html"])";
+    const nlohmann::json expected =
+    {
+        "https://www.freedesktop.org/wiki/Software/systemd/APIFileSystems/",
+        "https://www.freedesktop.org/software/systemd/man/systemd-fstab-generator.html"
+    };
+
+    EXPECT_EQ(handler->StringToJsonArray(input), expected);
+}
+
+TEST_F(SCAEventHandlerTest, StringToJsonArray_SerialisedJsonArraySingleElement)
+{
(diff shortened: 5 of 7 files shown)


cborla (approved):
LGTM
- [x] Code changes reviewed

PR #38954, opened 2026-09-04
Gate first-sync durable markers on a proven round trip, not bare success

### LINKED ISSUE #38899: SCA logs "synchronization finished successfully" while the manager is unreachable, in the same cycle where FIM and Syscollector report the failure
|---|---|---|---|---|
| 5.0.0 | wazuh (agent, `shared_modules/sync_protocol`) | Agent | Docker, `public.ecr.aws/wazuh-cicd/wazuh/wazuh-agent:5.0.0-latest` | Amazon Linux 2023 |

**Severity:** Low

## Description

With the agent's network severed and the agent itself having already announced that the manager is unreachable, the SCA module keeps logging a successful synchronization every cycle. Two sibling modules driven by the same `agent_sync_protocol` library, in the same cycle, seconds apart, correctly report that they could not reach the manager.

No data is lost, and the message is not lying about anything SCA actually did: there was nothing queued to send. The problem is that the line is byte-identical to the one a real, delivered synchronization produces, so a log is not a usable source of truth about whether an agent's SCA state ever reached the manager.

Stock agent container, no configuration changes, default 5-minute synchronization cycle for all three modules. The outage is produced by detaching the container from its network, which severs the route to the manager while leaving the agent's own local sockets intact. That distinction is what the finding turns on: the transport check the protocol performs is against the local intake socket, not against the manager.

The agent must have been running when the outage began. An agent restarted while already offline pauses module production and never reaches a synchronization cycle, so it does not exhibit this at all.

## Steps to reproduce

```bash
# a running, healthy agent
docker network disconnect <compose-net> wui-agent-02

# wait for the agent to notice, then for the next 5-minute sync cycle
docker logs --since 10m wui-agent-02 2>&1 | grep -v DEBUG \
  | grep -iE "unreachable|finished successfully|deferred"
```

The `Manager unreachable` warning, an SCA success and two `deferred` failures all appear in the same window.

## Current behavior

The network is cut at 10:48:11. The agent detects it. Then, in one 5-minute cycle, the three modules disagree:

```
2026/08/24 10:49:15 wazuh-agentd: WARNING: Manager unreachable. Pausing module event production.
2026/08/24 10:52:02 wazuh-modulesd:sca: INFO: Starting SCA synchronization.
2026/08/24 10:52:02 wazuh-modulesd:sca: INFO: SCA synchronization finished successfully.
2026/08/24 10:52:16 wazuh-modulesd:syscollector: INFO: Starting inventory synchronization.
2026/08/24 10:52:24 wazuh-modulesd: WARNING: Process locked due to agent is offline. Waiting for connection...
2026/08/24 10:52:28 wazuh-modulesd:syscollector: INFO: Syscollector synchronization deferred: Failed to communicate with the manager. Will retry next cycle.
2026/08/24 10:52:44 wazuh-syscheckd: INFO: Starting FIM synchronization.
2026/08/24 10:54:05 wazuh-syscheckd: INFO: FIM synchronization deferred: Failed to communicate with the manager. Will retry next cycle.
```

The container genuinely had no route to the manager at that moment:

```bash
docker inspect -f '{{range $k,$v := .NetworkSettings.Networks}}{{$k}}{{end}}' wui-agent-02
# (empty: no networks attached)
docker exec wui-agent-02 sh -c '(echo > /dev/tcp/wazuh.manager/1517) 2>/dev/null && echo up || echo down'
# down
```

And a healthy cycle is indistinguishable, including the zero-second duration:

```
2026/08/24 10:42:02 wazuh-modulesd:sca: INFO: Starting SCA synchronization.
2026/08/24 10:42:02 wazuh-modulesd:sca: INFO: SCA synchronization finished successfully.
```

## Expected results

A cycle that transmitted nothing and confirmed nothing with the manager should not read the same as a delivered one. Either distinguish the empty case, or gate the success message on the module having actually reached the manager. The `DEBUG` breadcrumb that records the real situation already exists, one level below where an operator will look:

```
No items to synchronize in DELTA mode
```

### Root cause

SCA has the full failure-reason machinery from `wazuh#37159` and would report a real failure if it saw one. `sca_impl.cpp` only logs success when the protocol says so, and has dedicated branches for the deferred cases:

```cpp
    if (result.success)
    {
        LoggingHelper::getInstance().log(LOG_INFO, "SCA synchronization finished successfully.");
    }
    else
    {
        ...
        else if (result.managerNotReady && result.consecutiveFailures <= SYNC_MANAGER_NOT_READY_TOLERANCE)
        {
            LoggingHelper::getInstance().log(LOG_INFO, "SCA synchronization deferred: " + result.failureReason +
                                             " Will retry next cycle.");
        }
```

So the protocol returned success. Two things in `agent_sync_protocol.cpp` combine to make that happen with no network. First, the transport check at the top of `synchronizeModule()` tests the *local* intake socket to `wazuh-agentd`, which is unaffected by losing the route to the manager:

```cpp
    if (!m_syncTransport->checkStatus())
    {
        return {false, "Failed to reach the sync intake socket.", shouldStop()};
    }
```

Second, `synchronizeDeltaByBlocks()` starts from success and returns it untouched when the queue is empty, without ever attempting a send:

```cpp
    bool success = true;
    bool sentAny = false;
    ...
        if (dataToSync.empty())
        {
            if (!sentAny)
            {
                m_logger(LOG_DEBUG, "No items to synchronize in DELTA mode");
            }

            break;
        }
```

That is the whole asymmetry. FIM and Syscollector had deltas to ship (the test created files under a monitored directory), so they reached the transport and failed there. SCA had nothing pending, so it short-circuited and reported success. "Success" here means "nothing was outstanding", and nothing in the line says so.

### Verified in source

The success flag is initialised to true and, on an empty queue, is never revised. From `src/shared_modules/sync_protocol/src/agent_sync_protocol.cpp`, `wazuh` branch `5.0.0`:

```cpp
/* agent_sync_protocol.cpp:215 */
bool success = true;
bool sentAny = false;
size_t blocksSent = 0;

while (!shouldStop() && blocksSent < FULLSESSION_MAX_BLOCKS_PER_SYNC)
{
    std::vector<PersistedData> dataToSync;
    ...
    if (dataToSync.empty())
    {
        if (!sentAny)
        {
            m_logger(LOG_DEBUG, "No items to synchronize in DELTA mode");   /* :239 */
        }
        break;
    }
```

`success` is still `true` at the `break`, so `synchronizeModule()` returns a successful result without any exchange with the manager having taken place, and `sca_impl.cpp` prints its unqualified success line. The failure-reason machinery from `wazuh#37159` is present and correct; it simply never runs, because nothing failed. Nothing was queued.

That also explains why the sibling modules disagree in the same cycle: FIM and Syscollector had items to send, attempted delivery, and reported the failure honestly.

**How to reproduce:**

```bash
# cut the agent off from the manager and wait one SCA cycle
docker network disconnect <network> <agent>
sleep 330
docker logs <agent> 2>&1 | grep -E "sca:|syscollector:|FIM sync" | tail -8
# sca:            "SCA synchronization finished successfully."
# syscollector:   "Syscollector synchronization deferred: Failed to communicate with the manager."
```

**The breadcrumb that tells the truth is one level below**, so this is also a way to confirm the diagnosis:

```bash
docker exec <agent> sh -c "echo 'sca.debug=2' >> /var/ossec/etc/local_internal_options.conf"
# after a restart and one cycle: "No items to synchronize in DELTA mode"
```

### Impact

Log-level only, and bounded: SCA state is not lost, and it does synchronize on reconnect. What is lost is the ability to answer "did this agent's SCA data reach the manager?" from the agent log, and the inconsistency inside one cycle makes the surrounding `deferred` lines look like the anomaly rather than the accurate ones.

This one is worth reading together with [024](not-for-filing/duplicated/DUPLICATED_024-agent-stuck-enrolling-after-deletion.md), where the same empty-queue success is what hides a genuinely broken agent for 45 minutes.

### Tasks

- [ ] Distinguish "nothing was pending" from "everything pending was delivered" in `synchronizeDeltaByBlocks()`, which today returns the same success for both because `success` is initialised to `true` and an empty `dataToSync` breaks out of the loop untouched.
- [ ] Have the module log the empty case differently, or not at all, rather than as `SCA synchronization finished successfully.`
- [ ] Apply the same treatment to every caller of the shared protocol, not only SCA, so the three modules stay consistent within a cycle.
- [ ] Add a test that runs a synchronization cycle with an empty queue and asserts the log line differs from a delivering cycle.

cc @wazuh/devel-xdrsiem-agent

hernanvalenzuela:
## Description

Following up on #38899: the reported symptom (SCA logging "synchronization finished successfully" for an empty-queue cycle) was already fixed on this branch by the `sentAnything` field on `SyncModuleResult` and its use in each module's log message. Validated that fix end to end against a real manager (see the comment thread on #38899).

While verifying that fix, a related gap surfaced in the same area: `sentAnything` is documented as the only valid proof that a sync actually reached the manager, with an explicit warning in `agent_sync_protocol_types.hpp` that it "must NOT be used to gate durable state" for exactly this reason (an empty queue takes the same early-success path as a delivered one). FIM and Syscollector's own "first sync completed" markers were still gating on bare `success`, not on `sentAnything`, which is precisely the anti-pattern that comment warns against.

## Proposed Changes

- `src/syscheckd/src/run_check.c`: both places that set `first_sync_completed` from a `MODE_DELTA` sync result now also require `sync_result.sent_anything`.
- `src/wazuh_modules/syscollector/src/syscollectorImp.cpp`: `persistVDFirstSyncIfNeeded()`'s call now also requires `vdResult.sentAnything`.
- `src/unit_tests/syscheckd/test_run_check.c`: the two existing tests that verify the first-sync marker gets persisted (`test_fim_run_integrity_skips_initial_wait_for_pending_first_sync`, `test_fim_run_integrity_pause_still_waits_after_skip_is_consumed`) mocked a sync result with `sent_anything` left false (an empty-queue cycle), which is exactly the case this fix now withholds the marker for. Updated them to mock a sync that actually delivered something, matching the scenario they were meant to verify.

### Results and Evidence
No live end-to-end reproduction for this one (unlike the logging fix above, validated against a real manager). The evidence here is at the unit-test level:

- Both `test_fim_run_integrity_skips_initial_wait_for_pending_first_sync` and `test_fim_run_integrity_pause_still_waits_after_skip_is_consumed` specifically assert that `first_sync_completed` gets persisted after a sync. Before this change, their mock left `sent_anything` false (an empty-queue cycle) and the marker still got persisted -- i.e. these tests were passing while exercising exactly the bug being fixed. Updated to mock `sent_anything = true` (an actual delivery), which is now required for the assertion to hold.
- Full `build.py --readytoreview` passed on both modules: cppcheck, AStyle, compile, unit tests (including the two above), coverage (96.3% both), Valgrind (no leaks), ASAN.

Since the existing test suite already covered the "gate durable state on success" path without a `sent_anything=false` regression test guarding it, this validates the fix mechanically but doesn't prove the specific field-observed scenario (an agent restart racing a metadata reset) the way the original issue's Docker repro did.

  - [x] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

### Artifacts Affected
NA

### Configuration Changes
NA

### Tests Introduced
NA

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

### DIFF
--- a/src/syscheckd/src/run_check.c
+++ b/src/syscheckd/src/run_check.c
@@ -1274,7 +1274,11 @@ void * fim_run_integrity(__attribute__((unused)) void * args) {
             atomic_int_set(&fim_flush_result, result);
             atomic_int_set(&fim_flush_in_progress, 0);
 
-            if (sync_result.success && !first_sync_completed) {
+            // #38899: success alone does not prove the manager received anything -- an empty
+            // queue takes the same early-success path as a delivered one. sent_anything is
+            // this durable marker's only valid proof of a round trip (see its own doc comment
+            // in agent_sync_protocol_types.hpp).
+            if (sync_result.success && sync_result.sent_anything && !first_sync_completed) {
                 fim_db_update_last_sync_time_value(FIM_FIRST_SYNC_COMPLETED_METADATA_KEY, (int64_t)time(NULL));
                 first_sync_completed = true;
                 atomic_int_set(&syscheck.fim_first_sync_completed, 1);
@@ -1306,7 +1310,9 @@ void * fim_run_integrity(__attribute__((unused)) void * args) {
                     minfo("FIM synchronization finished: nothing to send.");
                 }
 
-                if (!first_sync_completed) {
+                // #38899: same reasoning as the flush branch above -- sent_anything is the only
+                // proof this cycle actually reached the manager, so it gates the durable marker.
+                if (sync_result.sent_anything && !first_sync_completed) {
                     fim_db_update_last_sync_time_value(FIM_FIRST_SYNC_COMPLETED_METADATA_KEY, (int64_t)time(NULL));
                     first_sync_completed = true;
                     atomic_int_set(&syscheck.fim_first_sync_completed, 1);
--- a/src/syscheckd/src/run_check.c
+++ b/src/syscheckd/src/run_check.c
@@ -1274,7 +1274,7 @@ void * fim_run_integrity(__attribute__((unused)) void * args) {
             atomic_int_set(&fim_flush_result, result);
             atomic_int_set(&fim_flush_in_progress, 0);
 
-            // #38899: success alone does not prove the manager received anything -- an empty
+            // success alone does not prove the manager received anything -- an empty
             // queue takes the same early-success path as a delivered one. sent_anything is
             // this durable marker's only valid proof of a round trip (see its own doc comment
             // in agent_sync_protocol_types.hpp).
--- a/src/shared_modules/sync_protocol/include/agent_sync_protocol_types.hpp
+++ b/src/shared_modules/sync_protocol/include/agent_sync_protocol_types.hpp
@@ -75,10 +75,25 @@ struct SyncModuleResult
     /// cycle is not a failure -- so this field carries the distinction instead, letting a module
     /// log "nothing to send" rather than "finished successfully".
     ///
-    /// It must NOT be used to gate durable state. A marker that records "the manager has this
-    /// agent's data" has to be gated on something that proves a round trip, such as
-    /// @ref IAgentSyncProtocol::notifyDataClean; gating it on a sync result instead is what lets
-    /// an untransmitted cycle be recorded as a completed one.
+    /// Not a general license to gate durable state on this field alone: a marker that records
+    /// "the manager has this agent's data" should normally be gated on something that proves a
+    /// round trip, such as @ref IAgentSyncProtocol::notifyDataClean -- gating it on a bare sync
+    /// result instead is what lets an untransmitted cycle be recorded as a completed one (#38899).
+    ///
+    /// FIM's and syscollector's own first-sync-completed markers are a deliberately accepted,
+    /// narrower exception: they gate on `success && sentAnything` together, not on this field
+    /// alone. `success` covers the whole DELTA session for the cycle (it turns false the moment
+    /// any block fails), so the pair proves "every block this cycle sent actually round-tripped",
+    /// closing the exact #38899 symptom (an empty-queue cycle no longer masquerades as a
+    /// completed sync). It is still not the same claim as "the manager has the agent's complete
+    /// current state": a first sync bigger than AgentSyncProtocol's own
+    /// `FULLSESSION_MAX_BLOCKS_PER_SYNC` (10 blocks) leaves items queued for a later cycle, and
+    /// this cycle's `success && sentAnything` is already true by the time that happens. Accepted
+    /// for these two markers specifically (see run_check.c and
+    /// syscollectorImp.cpp) because the alternative -- notifyDataClean(), an ad hoc
+    /// index-clearing call, not something wired into the periodic DELTA path -- would need a
+    /// separate mechanism entirely; any other consumer wanting a true round-trip proof should
+    /// still reach for that, not this field.
     ///
     /// Distinct from @ref sessionSkipped: that one means no session ran at all because another
     /// was already in flight, while this one is about whether the session that did run carried
--- a/src/unit_tests/syscheckd/test_run_check.c
+++ b/src/unit_tests/syscheckd/test_run_check.c
@@ -412,16 +412,31 @@ static void expect_fim_startup_log(bool first_sync_completed) {
 }
 
 static void expect_fim_run_integrity_sync_body(AgentSyncProtocolHandle* handle, uint32_t sync_interval, bool persist_first_sync_marker, bool expect_recovery_pass) {
+    static SyncModuleResult_t sync_result;
+
     expect_string(__wrap__minfo, formatted_msg, "Starting FIM synchronization.");
 
     expect_value(__wrap_asp_sync_module, handle, handle);
     expect_value(__wrap_asp_sync_module, mode, MODE_DELTA);
-    will_return(__wrap_asp_sync_module, true);
 
-    /* The shared asp_sync_module wrapper zero-fills its result, so sent_anything is false and
-     * run_check reports the empty-cycle wording. That distinction is the point of the field:
-     * "successfully" would claim the manager took data it never received. */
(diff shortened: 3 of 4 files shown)


src/syscheckd/src/run_check.c
```diff
@@ -1279,7 +1279,11 @@ void * fim_run_integrity(__attribute__((unused)) void * args) {
             atomic_int_set(&fim_flush_result, result);
             atomic_int_set(&fim_flush_in_progress, 0);
 
-            if (sync_result.success && !first_sync_completed) {
+            // #38899: success alone does not prove the manager received anything -- an empty
```
cborla: Let’s avoid listing issue numbers in the comments

src/syscheckd/src/run_check.c
```diff
@@ -1311,7 +1315,9 @@ void * fim_run_integrity(__attribute__((unused)) void * args) {
                     minfo("FIM synchronization finished: nothing to send.");
                 }
 
-                if (!first_sync_completed) {
+                // #38899: same reasoning as the flush branch above -- sent_anything is the only
```
cborla: Let’s avoid listing issue numbers in the comments

src/syscheckd/src/run_check.c
hernanvalenzuela: done

lchico (approved):
LGTM!

- [x] Code changes reviewed

cborla (approved):
LGTM
- [x] Code changes reviewed

PR #38969, opened 2026-09-07
Fix process inventory name truncated to fifteen characters

### LINKED ISSUE #38931: The process inventory truncates every process name to fifteen characters
`wazuh-states-inventory-processes` stores `process.name` as read from `/proc/<pid>/comm`, which the Linux kernel caps at fifteen characters. The full name is available in the same document, in `process.command_line`, and is not used. Wazuh's own `wazuh-logcollector` is inventoried as `wazuh-logcollec`.

|---|---|---|---|---|
| 5.0.0-latest | wazuh-agent process inventory collector | Agent | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 agent, single-node stack |

**Severity:** Medium. Rules and hunting queries select on `process.name`. A query for a binary whose name exceeds fifteen characters silently matches nothing, and the analyst has no way to see that the stored value was cut.

## The inventory as stored

```
   pid=3428     name=wazuh-logcollec        cmd=/var/ossec/bin/wazuh-logcollector
   pid=1764     name=wazuh-logcollec        cmd=/var/ossec/bin/wazuh-logcollector
```

`wazuh-logcollector` is eighteen characters. The stored name is fifteen. The full name sits in the same document, at the end of `process.command_line`, in the same row.

## Where the truncation comes from

On the agent:

```
$ for p in 1764 1779; do echo "pid=$p comm=$(cat /proc/$p/comm)"; done
pid=1764 comm=wazuh-logcollec
pid=1779 comm=wazuh-modulesd
```

`/proc/<pid>/comm` is bounded by the kernel's `TASK_COMM_LEN`, sixteen bytes including the terminator, so fifteen characters of name. `wazuh-modulesd` is fourteen characters and survives; `wazuh-logcollector` does not. The collector is reading `comm` and storing it verbatim.

## Why this is not a cosmetic issue

Names longer than fifteen characters are common among service binaries. The two containers here run a small process set, so `wazuh-logcollector` is the only instance the deployment itself produces, but the limit applies to every name the collector stores: on a general-purpose Linux host `systemd-journald` and `systemd-resolved` are both sixteen characters and would be stored one character short.

Two consequences follow from the truncation regardless of which names are affected. An attacker's binary named to exceed the limit is stored under a name that matches no rule, and the truncated value is indistinguishable from a genuinely short name, so a consumer cannot tell whether `wazuh-logcollec` is a complete name or a cut one.

Nothing in the document marks the value as truncated. A query for `process.name: wazuh-logcollector` returns zero rows on a host where the process is running and the inventory has recorded it.

## Expected behavior

`process.name` is the basename of the executable, taken from `/proc/<pid>/exe` or from the first token of `/proc/<pid>/cmdline`, both of which the collector already reads to fill `process.command_line`. `comm` is a fallback for kernel threads that have no executable.

## Steps to reproduce

1. `POST /wazuh-states-inventory-processes/_search` with `{"query":{"match_all":{}}}` and read `process.name` beside `process.command_line`.
2. Any row whose command line basename is longer than fifteen characters shows a shorter `process.name`. `wazuh-logcollector` is present on both agents.
3. On the agent, `cat /proc/<pid>/comm` for that pid returns exactly the truncated value.

## Related

[136](BUG_136-the-ports-inventory-records-forty-one-open-ports-and-names-the-process-for-none-of-them.md) is the ports inventory, which does not name the process at all. [134](BUG_134-three-inventory-collectors-ship-enabled-and-produce-nothing-and-one-of-them-is-the-package-list-vulnerability-detection-needs.md) is the set of inventory collectors that produce nothing.

cc @wazuh/devel-xdrsiem-agent

Darioortegaleyva:
## Description

This PR fixes the process inventory storing every process name truncated to fifteen characters.

**Issue:** wazuh/wazuh#38931
**Internal Reference:** N/A

## Proposed Changes

- Updated `src/data_provider/src/sysInfoLinux.cpp`.
- `process.name` now derives from the basename of the full `cmdline[0]` (already read for the command-line field) instead of `/proc/<pid>/comm`, which the kernel caps at fifteen characters (`TASK_COMM_LEN`).
- Falls back to `comm` only when there is no `cmdline` (kernel threads).

### Results and Evidence

Validated with `sysinfo_test_tool` (`src/data_provider`) built from this branch, run on a live Linux host, comparing `--processes` output before and after the fix.

**Before:**

```sql
pid=42    name="systemd-journal"   cmd=/usr/lib/systemd/systemd-journald
pid=205   name="systemd-resolve"   cmd=/usr/lib/systemd/systemd-resolved
pid=206   name="systemd-timesyn"   cmd=/usr/lib/systemd/systemd-timesyncd
```

**After:**

```sql
pid=42    name="systemd-journald"    cmd=/usr/lib/systemd/systemd-journald
pid=205   name="systemd-resolved"    cmd=/usr/lib/systemd/systemd-resolved
pid=206   name="systemd-timesyncd"   cmd=/usr/lib/systemd/systemd-timesyncd
```

`cmd`/`argvs` unchanged; only `name` is affected. Same bug confirmed present on `4.14.6`, `4.14.7`, `4.14.8`, `4.14.9`, `5.0.0` and `main` (same `getProcessInfo()` code path); this PR targets `4.14.9`.

### Artifacts Affected

- `src/data_provider/src/sysInfoLinux.cpp`

### Configuration Changes

N/A

### Documentation Updates

N/A

### Tests Introduced

- **Scope:** Manual validation.
- **Details:** Built `sysinfo_test_tool` and compared `--processes` output before/after the fix on a live Linux host. `getProcessInfo()` is a `static` function not exposed to unit tests today; no test seam exists to add a gtest case without a broader refactor.

## Review Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux

  - [ ] Coverity
  - [ ] AddressSanitizer

**General Checklist:**

- [ ] Code changes reviewed

### DIFF
--- a/src/data_provider/src/sysInfoLinux.cpp
+++ b/src/data_provider/src/sysInfoLinux.cpp
@@ -120,8 +120,9 @@ static nlohmann::json getProcessInfo(const SysInfoProcess& process)
         }
     }
 
-    // comm is capped at TASK_COMM_LEN; fall back to it only for kernel threads (no cmdline).
-    jsProcessInfo["name"] = commandLine.empty() ? process->cmd : std::filesystem::path(commandLine).filename().string();
+    // comm is capped at TASK_COMM_LEN; fall back to it whenever cmdline can't give a name.
+    const std::string baseName{commandLine.empty() ? std::string{} : std::filesystem::path(commandLine).filename().string()};
+    jsProcessInfo["name"] = baseName.empty() ? process->cmd : baseName;
     jsProcessInfo["cmd"]        = commandLine;
     jsProcessInfo["argvs"]      = commandLineArgs;
     jsProcessInfo["euser"]      = process->euser;
--- a/src/data_provider/src/sysInfoLinux.cpp
+++ b/src/data_provider/src/sysInfoLinux.cpp
@@ -93,7 +93,6 @@ static nlohmann::json getProcessInfo(const SysInfoProcess& process)
     nlohmann::json jsProcessInfo{};
     // Current process information
     jsProcessInfo["pid"]        = std::to_string(process->tid);
-    jsProcessInfo["name"]       = process->cmd;
     jsProcessInfo["state"]      = &process->state;
     jsProcessInfo["ppid"]       = process->ppid;
     jsProcessInfo["utime"]      = process->utime;
@@ -121,6 +120,10 @@ static nlohmann::json getProcessInfo(const SysInfoProcess& process)
         }
     }
 
+    // process->cmd (/proc/<pid>/comm) is capped at TASK_COMM_LEN (15 chars); use the
+    // untruncated argv[0] basename instead, falling back to comm for kernel threads,
+    // which have no cmdline.
+    jsProcessInfo["name"] = commandLine.empty() ? process->cmd : std::filesystem::path(commandLine).filename().string();
     jsProcessInfo["cmd"]        = commandLine;
     jsProcessInfo["argvs"]      = commandLineArgs;
     jsProcessInfo["euser"]      = process->euser;
--- a/src/data_provider/src/sysInfoLinux.cpp
+++ b/src/data_provider/src/sysInfoLinux.cpp
@@ -120,9 +120,7 @@ static nlohmann::json getProcessInfo(const SysInfoProcess& process)
         }
     }
 
-    // process->cmd (/proc/<pid>/comm) is capped at TASK_COMM_LEN (15 chars); use the
-    // untruncated argv[0] basename instead, falling back to comm for kernel threads,
-    // which have no cmdline.
+    // comm is capped at TASK_COMM_LEN; fall back to it only for kernel threads (no cmdline).
     jsProcessInfo["name"] = commandLine.empty() ? process->cmd : std::filesystem::path(commandLine).filename().string();
     jsProcessInfo["cmd"]        = commandLine;
     jsProcessInfo["argvs"]      = commandLineArgs;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ All notable changes to this project will be documented in this file.
 - Fixed the `disable-account` active response reporting success when the account was not disabled, and stopped `is_valid_username()` from rejecting valid usernames containing consecutive dots. ([#38637](https://github.com/wazuh/wazuh/pull/38637))
 - Fixed the `disable-account` active response returning success when the account command was missing or the system was not supported. ([#38645](https://github.com/wazuh/wazuh/pull/38645))
 - Fixed the gcloud wodle masking missing-dependency errors as an unrelated `AttributeError`. ([#38856](https://github.com/wazuh/wazuh/pull/38856))
+- Fixed process names truncated to fifteen characters in the syscollector inventory. ([#38969](https://github.com/wazuh/wazuh/pull/38969))
 
 ### Ruleset
 


Miguevrgo (dismissed):

src/data_provider/src/sysInfoLinux.cpp
```diff
@@ -121,6 +120,8 @@ static nlohmann::json getProcessInfo(const SysInfoProcess& process)
         }
     }
 
+    // comm is capped at TASK_COMM_LEN; fall back to it only for kernel threads (no cmdline).
+    jsProcessInfo["name"] = commandLine.empty() ? process->cmd : std::filesystem::path(commandLine).filename().string();
```
MarcelKemp: **[S3 · non-blocking · introduced]** `std::filesystem::path(commandLine).filename().string()` returns an empty string when `commandLine` ends in `/`. That's an unlikely input for `cmdline[0]`, but if it happened, `process.name` would become `""` — worse than the truncated `comm` value we had before. Suggest also falling back to `comm` when the basename comes out empty, not just when `commandLine` itself is empty.

```suggestion
    const std::string baseName{commandLine.empty() ? std::string{} : std::filesystem::path(commandLine).filename().string()};
    jsProcessInfo["name"] = baseName.empty() ? process->cmd : baseName;
```

MarcelKemp: **[S3 · non-blocking · pre-existing]** Not introduced by this PR — flagging it for a follow-up issue, not for this review. `getProcessInfo()` is `static` and has no test seam, so no automated test would catch a regression of this fix (the existing `processes_cb` tests in `sysInfo_test.cpp` only exercise mocked JSON through the callback pipeline, never the real Linux `getProcessInfo()`). Worth a follow-up refactor to make this testable.

MarcelKemp: **[S3 · non-blocking · introduced]** Worth confirming this value change is intended, since it goes beyond the truncation the issue reports. `argv[0]` is fully process-controlled in both content and length, and a class of real daemons rewrites it via `setproctitle`: PostgreSQL reports things like `postgres: checkpointer`, which has no `/` for `filename()` to trim, so that whole string becomes `process.name`. Login shells are the other common case — the convention is `argv[0] == "-bash"`, so the name becomes `-bash` rather than `bash`.

To be fair to the change, this is not a new spoofing surface: `comm` was settable by the process too, via `prctl(PR_SET_NAME)` or a write to `/proc/self/comm`. What widens is the room — from a bounded 15 bytes to an arbitrary string.

No code change needed; the issue asks for exactly this. But the CHANGELOG entry currently mentions only the truncation, and anything that groups or matches on an exact process name will see different values after an agent upgrade. Either call the value change out in the entry, or keep the kernel `comm` in a separate field so the inventory carries both views.

src/data_provider/src/sysInfoLinux.cpp
```diff
@@ -121,6 +120,9 @@ static nlohmann::json getProcessInfo(const SysInfoProcess& process)
         }
     }
 
+    // comm is capped at TASK_COMM_LEN; fall back to it whenever cmdline can't give a name.
+    const std::string baseName{commandLine.empty() ? std::string{} : std::filesystem::path(commandLine).filename().string()};
```
MarcelKemp: **[Iteration 2 · S3 · resolved · from iteration 1]** F1 is resolved. The increment now safely checks `baseName.empty()` before use and falls back to `process->cmd` when filename() returns empty. The suggested fix has been implemented exactly as proposed in iteration 1.

src/data_provider/src/sysInfoLinux.cpp
```diff
@@ -121,6 +120,9 @@ static nlohmann::json getProcessInfo(const SysInfoProcess& process)
         }
     }
 
+    // comm is capped at TASK_COMM_LEN; fall back to it whenever cmdline can't give a name.
```
MarcelKemp: **[Iteration 2 · S4 · resolved · from iteration 1]** F2 is resolved. The comment "fall back to it whenever cmdline can't give a name" now correctly documents both the empty commandLine case and the empty baseName case. Accurate and clear.

src/data_provider/src/sysInfoLinux.cpp
MarcelKemp: **[S3 · non-blocking · pre-existing]** Not introduced by this PR — flagging it for a follow-up issue, not for this review. F3: `getProcessInfo()` is `static` and has no test seam, so no automated test would catch a regression of this fix (the existing `processes_cb` tests in `sysInfo_test.cpp` only exercise mocked JSON through the callback pipeline, never the real Linux `getProcessInfo()`). Worth a follow-up refactor to make this testable.

MarcelKemp: **[S3 · non-blocking · introduced]** F4 persists as an accepted behavioral change. Worth noting: `process.name` now derives fully from `argv[0]`, which real daemons like PostgreSQL rewrite via `setproctitle` (e.g., showing `postgres: checkpointer` instead of `postgres`), and login shells use `argv[0] == "-bash"`. This is the intended trade-off to fix truncation — full names are more useful than 15-character truncations. The CHANGELOG should document that process names may change for daemon processes after upgrade.

MarcelKemp (approved):

PR #38973, opened 2026-09-07
Vulnerability findings collapse

### LINKED ISSUE #38713: Vulnerability findings collapse for every rpm and deb package while the feed reports healthy
|---|---|---|---|---|
| 5.0.0 (2026-08-28 nightly) | vulnerability-scanner | Manager (all-in-one) | Installation assistant | Ubuntu 24.04 |

## Description

The vulnerability scanner produces correct findings, and then stops resolving distribution advisories while the manager keeps reporting a healthy feed. Same hosts, same package sets, same scan sizes: only the number of vulnerable packages collapses. Nothing is logged by the content updater in between.

Possibly the same underlying defect as #38272, which reports findings being wiped after the first CTI delta update on 4.14.7. Filing separately because two things do not match that report — the trigger and the visibility — and both may matter to whoever debugs it. Close as a duplicate if the server team judges it is one.

## What happened

Single-node all-in-one from the 2026-08-28 nightly (manager + indexer + dashboard), two agents: Amazon Linux 2023 and Ubuntu 22.04. The feed finished its initial load, both agents were scanned correctly, and then — without any content update in between — every later scan of the same packages found almost nothing.

```
14:29:38 vulnerability-scanner: Feed update process completed (changed=true).
14:29:38 vulnerability-scanner: CVE feed fully loaded — per-agent scans unblocked.

14:37:54 Agent '001' - Packages scan completed in 144 ms: 679 packages scanned, 0 skipped, 226 vulnerable packages
14:39:12 Agent '002' - Packages scan completed in  73 ms: 849 packages scanned, 0 skipped, 256 vulnerable packages

   ... agents deleted through the API and re-enrolled under new ids; no content-updater activity logged ...

14:44:21 Agent '004' - Packages scan completed in  41 ms: 849 packages scanned, 0 skipped,   2 vulnerable packages
14:47:26 Agent '003' - Packages scan completed in  16 ms: 679 packages scanned, 0 skipped,   0 vulnerable packages
```

Same hosts, same package sets (679 and 849, unchanged), 226 → 0 and 256 → 2 vulnerable.

## Two differences from #38272

That report states the failure comes with "no errors logged at any level". On 5.0.0 it is not silent: the Amazon Linux scan emitted **674 warnings over its 679 packages**, one per package:

```
14:47:26 vulnerability-scanner: WARNING: Failed to scan package 'zlib': Couldn't find column family: 'alas_'
```

`alas_` is the Amazon Linux advisory column family. So on this build the content database is not merely returning empty candidates — a whole column family is no longer reachable, and the scanner says so.

**What survives says more than what fails.** Breaking the two agents down by package type:

| agent | packages | findings after | on which packages |
|---|---|---|---|
| Amazon Linux 2023 | 675 rpm + 4 pypi | 0 | — |
| Ubuntu 22.04 | 839 deb + 7 pypi + 3 snap | 9 | `lxd` (snap), `awscli` (pypi) |

Every **distribution** package lost its findings on both agents — all 675 rpm and all 839 deb. The only findings left anywhere are on non-distribution packages, which are matched through the generic CPE/NVD path rather than a per-OS advisory feed. So this is not an Amazon-only failure and not specific to `alas_`: the distro-advisory matching path is what stopped resolving, on both operating systems. Amazon's failure is loud because the missing column family is named in a warning per rpm package (674 of them, one short of its 675 rpms); Ubuntu's is silent, which is the behaviour #38272 describes.

**The trigger also differs.** #38272 pins the corruption to the first delta update applied on top of the CTI snapshot. Here nothing updated the content between the healthy and the broken scans: no `content-updater` line, no `Feed update process completed`, no offset change. What sits in that window is an agent deletion through the API and the re-enrollment that follows it. Whether that is a coincidence of timing or a second path into the same corrupted state, I cannot tell from the outside.

## A restart does not recover it

```
14:54:08 content-updater: IndexerDownloader: Starting incremental update from offset 905742
14:54:08 content-updater: IndexerDownloader: Incremental update download phase complete — 0 documents
14:54:08 vulnerability-scanner: Feed update process completed (changed=false).
14:54:08 vulnerability-scanner: CVE feed fully loaded — per-agent scans unblocked.

14:57:34 Agent '005' - Packages scan completed in 65 ms: 679 packages scanned, 0 skipped, 0 vulnerable packages
14:59:30 Agent '006' - Packages scan completed in 54 ms: 849 packages scanned, 0 skipped, 2 vulnerable packages
```

The manager restarts, reports the feed fully loaded, and scans still find nothing — the `alas_` warnings repeat, 674 more of them. So the state survives a restart and an incremental update that reports itself as a no-op, which matches #38272's point that the recorded offset makes later checks no-ops.

## Impact

An affected manager looks healthy from every angle an operator checks — module started, feed offset current, "CVE feed fully loaded", indexer connector green — while `wazuh-states-vulnerabilities` holds a small fraction of the real exposure. On the Amazon Linux agent it held none at all.

Happy to attach the full `wazuh-manager.log` window if it helps.

jepalfer:
## Description

Closes #38713.

`alas_` isn't a corrupted column family name. It's `alas_$(MAJOR_VERSION)` with an empty major version, so the content database was never damaged — which is why this isn't a duplicate of #38272. The report's own numbers say so: `lxd` and `awscli` kept their findings because snap and pypi fall through to `nvd`, which has no OS release in its name. All 675 rpm and 839 deb packages went through a per-OS advisory feed.

The OS release only ever reaches the scanner in the OS document a session carries. `majorVersion` was assigned in two places in the module, both reading `host.os.major` out of an inventory document; neither Start struct has the field, nothing derived it from `os_version`, and nothing persists it between sessions. The agent-id-keyed `OsDataCache` that used to cover this was deleted in `828c801665` with no replacement.

The damage is the reconciliation. `eventGetContext` pre-loads every existing finding as a deletion and only a confirmed version match flips one back; `packageScan` caught the lookup throw, warned, and returned, so `ResultIndexer` issued the deletion. A failed scan and a clean package were the same event. A plain package delta routes to a chain without `eventGetContext`, so it returns the wrong answer without destroying anything — the deletion needs the reconciling chain, reached when `Start.index` names the system or hotfix index. Both legs are measured below.

The loud and silent halves are feed content, not a code branch. The real `cnaMapping` has no `canonical` or `debian` key, so a deb package resolves as-is to a column family that exists, `seek()` succeeds, and its candidates die one at a time in `platformVerify` against an empty code name behind a `logDebug2` — the "no errors logged at any level" #38272 describes. The agent is clean either way: Amazon Linux 2023 reports `os_major` as `2023` correctly, so the value is on the wire.

PR #38623 merged on 31/08 and the report is the 28/08 nightly, so that trigger may already be gone. I didn't scope around it: a session without a usable OS context misreports the agent whatever put it in that state.

Three feed-side defects turned up while chasing the "looks healthy from every angle" complaint and are **not** fixed here — `isFeedReady()` never checks a single CVE candidate column, the feed cursor advances non-transactionally in a separate RocksDB (which is #38272's central claim almost verbatim), and `m_initialLoad` is derived once so an interrupted initial load takes the delta cleanup path. The first commit makes all three harmless rather than destructive. No issue opened.

## Proposed Changes

Six commits. `9bcf5b1cdc` and `c2618d9576` are the fix; `41ea9f95c6` keeps `qa/test_data/017` matching; `e87824a930` and `70cc5698eb` are corrections the build and the suites forced (below); `58a53c5d71` is the changelog.

| File | Change |
| --- | --- |
| `scanOrchestrator/scanContext.hpp` | Unevaluated-package tracking; `hasUsableOsContext()` / `missingOsContextFields()`; `deriveMajorVersion()` in both constructors and `setOSData`; `setOSData` merges field-wise instead of assigning |
| `scanOrchestrator/factory/packageScanner.hpp` | `recordUnevaluatedPackage()` at all three failure sites; `assertSubstitutable()` refuses an unfillable CNA template; `platformVerify` tells an unknown suite from a mismatched one; summary count, one aggregate warning, per-reason log throttling |
| `scanOrchestrator/factory/resultIndexer.hpp` | Skips the pre-loaded deletion for a package that wasn't evaluated |
| `scanOrchestrator/scanOrchestrator.hpp` | Aborts the **VDFirst** scan on an unusable OS context, ahead of its `deleteByQuery` — the one deletion that doesn't pass through the reconciliation and so can't be held back per package |
| `scanOrchestrator/scanAgentList.hpp` | Notes an incomplete OS context at `debug1` and scans anyway: no `deleteByQuery` here, and skipping would discard the `pypi`/`npm`/`snap` findings that need no OS release |
| `databaseFeedManager/databaseFeedManager.hpp` | `getVulnerabilitiesCandidates` checks `columnExists` and names the CNA the feed has no content for |
| `tests/mocks/TrampolineScanContext.hpp` | The two accessors the guard calls, so the `scanOrchestrator` tests compile |

Two design points. The release identifiers are alternatives, not both required — rpm families key on the major version, Debian families on the suite code name, Windows and macOS need neither — so `hasUsableOsContext()` encodes that rather than a per-distro allowlist. And deriving the major version from Start's `osversion` lets a Debian-family agent past that predicate while the field its feed actually matches on is still missing, which is why the code-name case is handled in `platformVerify`, where the information is.

Left out on purpose: the `vd.scan.*` metric counters (`TPackageScanner` takes no metrics manager, so adding one changes its constructor and every construction site) and the `docs/ref/` write-up of CNA resolution, which the module's docs don't cover at all today. The latter is part of `module-standards.md`'s bar, so it probably blocks merge.

### Results and Evidence

Built and run in a fresh VM against a copy of the tree, with Ubuntu 22.04 (jammy) and Amazon Linux 2023 (amzn) guests. **524/524 unit tests, 20/20 QA efficacy scenarios**, from a clean build of exactly what's committed.

Reported case driven through `inventory_sync_server_testtool`, so the counts are real `wazuh-states-vulnerabilities` documents. Session 1 establishes a finding with a complete OS context; session 2 replays the same packages with the OS document absent. Before and after come from one build tree — the six changed files swapped to their pre-fix versions and back. `gzip 1.10-5.amzn2023.0.1` against the feed's `alas_2023` entries, `libnss3 2:3.17.0-0ubuntu1` against its jammy `canonical` entries.

**Reconciling FullScan — the path that deletes:**

| Agent | before | after |
| --- | --- | --- |
| Ubuntu 22.04 | **1 → 0 findings**, `Deleting detection '202_pkg-libnss3-ubuntu_CVE-2014-3566'`, reported as `0 vulnerable packages`, nothing above `debug2` | **1 → 1**, `Keeping detection '202_…': its package could not be evaluated` |
| Amazon Linux 2023 | `0 vulnerable packages` + `Failed to scan package 'gzip': Couldn't find column family: 'alas_'` | `1 vulnerable packages, 0 not evaluated`, no errors |

**Package delta — wrong answer, no deletion:**

| Agent | before | after |
| --- | --- | --- |
| Amazon Linux 2023 | `0 vulnerable` + the `alas_` error | `1 vulnerable`, error gone |
| Ubuntu 22.04 | `0 vulnerable`, no log line at all | `0 vulnerable, 1 not evaluated` + `Platform undecidable: … restricted to a suite code name and the agent's OS code name is not known` |

That `1 → 0` with `Deleting detection` is the collapse this issue reports. The two families are fixed by different layers: Amazon Linux is **repaired** (the derived major version resolves `alas_2023`), Ubuntu is **protected** (no code name is derivable from `22.04`).

Four things the build and the suites contradicted, all of which reading had missed:

- **The tip didn't compile.** `scanOrchestrator_test.cpp` and `scanOrchestrator_concurrency_test.cpp` instantiate the chains with `TrampolineTScanContext`, a stub, not the real `ScanContext`.
- **The guard over-reached.** `deleteByQuery` exists in exactly one place, so skipping elsewhere discarded NVD-path findings to prevent deletions the per-package accounting already made impossible. Now scoped to the VDFirst branch.
- **`setOSData`'s merge does more than claimed.** `buildContextFromAgent` picks the agent record as an explicit fallback and the old assignment discarded it one line later, leaving a thin-document agent with no platform and no version — the state that resolves nothing.
- **Pre-fix, the Amazon Linux finding survived only by luck.** `EventDetailsBuilder` also threw (`descriptions_alas_`), so no ECS event existed and `resultIndexer` skips detections without one. Had it succeeded, that finding would have gone too. `getVulnerabilityDescriptiveInformation` still has no `columnExists` guard of its own; it's unreachable now only because `getCNA` refuses to produce the truncated name.

C++ in a manager module only — no agent, engine, decoder/rule or Python code — so the Windows, macOS, decoder/rule and engine checklists don't apply.

  - [x] Linux
  - [ ] Windows — N/A
  - [ ] MAC OS X — N/A
- Memory tests for Linux
  - [ ] Coverity — pending
  - [ ] Valgrind (memcheck and descriptor leaks check) — pending
  - [ ] AddressSanitizer — pending
- Memory tests for Windows / macOS — N/A

### Artifacts Affected

`wazuh-manager-modulesd` only — `vulnerability_scanner` is compiled into it. No other executable, no default configuration file, no packaging, no schema, no index template.

### Configuration Changes

None: no new parameter, no changed default, no new error code, no internal option. What this adds is a correctness precondition, not something a deployment should be able to turn off.

Three intended changes to log content. The `Packages scan completed` line gains a trailing `%zu not evaluated`, with the existing fields keeping their position so a prefix match still works. A VDFirst scan whose OS context can't resolve a distribution package is aborted with a warning naming the missing fields, instead of wiping the agent's findings and re-detecting nothing. And a repeated per-package failure is logged once at `warn` per distinct reason, then at `debug2`, plus one aggregate line — the reported case produced 674 identical warnings.

### Tests Introduced

Sixteen unit cases and one QA scenario.

| File | Scope |
| --- | --- |
| `tests/unit/packageScanner_test.cpp` | CNA resolution refusing an empty major version; a missing column family preserving pre-loaded findings; an unknown code name making suite candidates undecidable; a known code name still reporting a real mismatch |
| `tests/unit/resultIndexer_test.cpp` | An unevaluated package's deletion is held; an unrelated evaluated package's solved finding is still removed |
| `tests/unit/scanContext_test.cpp` | Major-version derivation, numeric and not; `setOSData` merge both ways; `hasUsableOsContext` per platform family including Windows and macOS; unevaluated tracking |
| `tests/unit/databaseFeedManager_test.cpp` | A real RocksDB with the column family deliberately absent, asserting the error names the CNA and the callback never fires |
| `qa/test_data/019/` | Scenario `001` with its OS `data_context` removed: the candidate goes undecidable, the package is counted as not evaluated |

Two exist to keep the fix honest: the known-code-name case stops the undecidable branch swallowing genuine negatives, and the other-packages case stops one failure freezing the whole agent's state.

Worth knowing why this shipped: three existing tests look like they cover the template substitution, and all three mock `getCnaNameByFormat` to return an already-resolved string that isn't a `cnaMapping` key — so `getCNA` takes its use-as-is branch and the real `replaceAll` never executes anywhere in the suite. Nothing set the major version empty, and all eighteen existing QA scenarios carry an OS document.

## Review Checklist

- [ ] Code changes reviewed


src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/factory/packageScanner.hpp
```diff
@@ -1113,6 +1205,11 @@ class TPackageScanner final : public AbstractHandler<std::shared_ptr<TScanContex
             }
             catch (const std::exception& e)
             {
+                // Returning false here means "this candidate does not make the package vulnerable", which
+                // is not what an exception establishes. Record the package as unevaluated so its existing
+                // findings are preserved rather than reconciled away on the strength of a failure.
+                data->markPackageUnevaluated(detectionIdBase);
```
jotacarma90: **[S2 · blocking · introduced]** A single candidate-level exception now marks the whole package unevaluated, so `resultIndexer.hpp:126` drops every pre-loaded deletion for it and a CVE the agent has genuinely patched is never removed. `VersionMatcher::compare` throws routinely (`versionMatcher.hpp:276/282/302`) when a boundary cannot be parsed for the format or the two objects differ in type — this repo's reduced feed carries `nvd` boundaries such as `r29`, `r30` and `x14.3.3`. The throw happens after `platformVerify` and `vendorVerify`, so what is undecidable is that one CVE, not the package.

```suggestion
                // Only THIS candidate went unevaluated. An exception says nothing about the rest of the
                // package, so drop the pre-loaded deletion of this CVE and leave the others to reconcile.
                if (callbackData.cveId())
                {
                    data->removeCVE(buildDetectionId(detectionIdBase, callbackData.cveId()->str()));
                }
```

src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/scanOrchestrator.hpp
```diff
@@ -346,6 +346,22 @@ class TScanOrchestrator final
...
+            if (!scanContext->hasUsableOsContext())
+            {
+                logWarn(WM_VULNSCAN_LOGTAG,
+                        "VDFirst scan aborted for agent '%s': incomplete OS context (missing: %s). "
+                        "Skipping deleteByQuery to preserve existing vulnerability state; the next session "
+                        "that carries the OS inventory document will scan the agent.",
+                        agentId.c_str(),
+                        scanContext->missingOsContextFields().c_str());
+                return false;
+            }
```
jotacarma90: **[S1 · blocking · introduced]** This aborts the first scan of any agent whose OS legitimately has no release identifier: Arch Linux reports an empty `os_version` on purpose (`version_op.c:506-509`) and its `/etc/os-release` has no `VERSION`, `VERSION_ID` or `VERSION_CODENAME`, so `hasUsableOsContext()` is false even with a complete OS document in the session. `return false` → `ScanVerdict::Skipped` (`vdScannerAdapter.cpp:95`) → `vdScanLane.cpp:487` indexes the inventory and answers 200, and `syscollectorImp.cpp:2826` then persists `vd_first_sync_completed`, so the agent never sends VDFirst again and gets no baseline at all. Skip only the `deleteByQuery` and still run the chain — the same argument the PR makes for the feed-update sweep — which also needs the `deleteByQuery` at line ~379 wrapped in `if (!preserveExistingFindings)`.

```suggestion
            // deleteByQuery wipes the agent's findings BEFORE anything is evaluated and does not go
            // through the reconciliation, so the per-package "could not evaluate" accounting cannot put
            // them back. Skip the delete, but still scan: an OS that reports no release at all (a rolling
            // release such as Arch) must not be left without a first scan, and whatever the session can
            // resolve is worth indexing.
            const bool preserveExistingFindings = !scanContext->hasUsableOsContext();

            if (preserveExistingFindings)
            {
                logWarn(WM_VULNSCAN_LOGTAG,
                        "VDFirst scan for agent '%s' has an incomplete OS context (missing: %s). "
                        "Skipping deleteByQuery to preserve existing vulnerability state; the packages "
                        "that can be resolved are still scanned.",
                        agentId.c_str(),
                        scanContext->missingOsContextFields().c_str());
            }
```

src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/factory/packageScanner.hpp
```diff
@@ -571,17 +608,39 @@ class TPackageScanner final : public AbstractHandler<std::shared_ptr<TScanContex
...
             }
         }
 
+        // The Debian-family advisory feeds restrict their candidates to suite code names, so with no code
+        // name in the OS context every one of them compares unequal and the package looks clean. That is
+        // the quiet counterpart of the missing-column-family failure: same cause, same wrong answer, no log
+        // and no exception. Flag it so the package is treated as unevaluated rather than as clean.
+        if (comparedByCodeName && contextData->osCodeName().empty())
+        {
+            pkgStats.undecidablePlatform = true;
```
jotacarma90: **[S3 · non-blocking · introduced]** `comparedByCodeName` is set for any non-CPE platform entry, but the feed's non-CPE entries are not only Debian suites: `vd_reduced_feed.json` has `go`, `kubernetes`, `swift`, `aws`, `vmware`, `vmware_vsphere`, `arch` and `faye` in `cna` containers. `host.os.codename` is empty for every RHEL, Rocky, AlmaLinux, Oracle, Amazon Linux, SLES, Alpine, Arch and Windows agent, because `UnixOsParser` only emits it from `VERSION_CODENAME` (`sysOsParsers.cpp:133`) — so a healthy agent gets a package frozen and the aggregate WARN on every scan. `hasUsableOsContext()` already encodes that distinction; `platformVerify` needs it too.

```suggestion
        if (comparedByCodeName && contextData->osCodeName().empty() && package.format == "deb")
        {
            pkgStats.undecidablePlatform = true;
```

src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/scanContext.hpp
```diff
@@ -1082,6 +1155,74 @@ struct TScanContext final
...
+            return true;
+        }
+
+        if (m_osContextDataCache.platform.empty())
+        {
+            return false;
+        }
+
+        return !m_osContextDataCache.majorVersion.empty() || !m_osContextDataCache.codeName.empty();
+    }
```
jotacarma90: **[S3 · non-blocking · introduced]** This predicate cannot fire for the Debian family the VDFirst guard was written for, because the same PR derives the major version from `Start.osversion` (lines 498 and 554) and `version_op.c` publishes `VERSION`, not `VERSION_ID`: Debian 12 reports `"12 (bookworm)"` → `"12"` and Ubuntu 22.04 reports `"22.04.4 LTS (Jammy Jellyfish)"` → `"22"`. QA scenario 019 has to fabricate `osversion: "Bookworm"` to reach the guard at all. For the suite-keyed families the predicate should require the code name rather than accept a derived major version.

src/wazuh_modules/vulnerability_scanner/src/databaseFeedManager/databaseFeedManager.hpp
```diff
@@ -766,6 +766,17 @@ class TDatabaseFeedManager final : public Observer<nlohmann::json&>
...
+        // seek() resolves the column family by name and throws RocksDB's generic "Couldn't find column
+        // family" when it does not exist, which says nothing about which CNA was expected or why. Check
+        // first -- as getHotfixVulnerabilities() and readGlobalMapsNoLock() already do -- and report the
+        // CNA that is missing from the feed, so the failure is diagnosable from one log line.
+        if (!m_feedDatabase->columnExists(cnaName))
+        {
+            throw std::runtime_error("No feed content for CNA '" + cnaName +
+                                     "': the vulnerability content database has no column family under that "
+                                     "name, so no candidate can be evaluated against it.");
+        }
```
jotacarma90: **[S3 · non-blocking · introduced]** The new throw is a clear improvement over RocksDB's generic message, but it is consumed by the outer catch at `packageScanner.hpp:1269` → `recordUnevaluatedPackage`, and `scanWithCNAPriority` (`packageScanner.hpp:1281-1305`) calls `packageScan` once per CNA in `adp_default_array`. One ADP with no column family therefore freezes every package that goes through that array, with no fall-through for "the other CNAs answered fine". Accumulate the outcome in `scanWithCNAPriority` and mark the package unevaluated only if no CNA in the array could be queried.

src/wazuh_modules/vulnerability_scanner/src/scanOrchestrator/scanContext.hpp
```diff
@@ -1137,12 +1278,54 @@ struct TScanContext final
...
+        merge(m_osContextDataCache.patch, osData.patch);
+        merge(m_osContextDataCache.build, osData.build);
+        merge(m_osContextDataCache.platform, osData.platform);
+        merge(m_osContextDataCache.version, osData.version);
+        merge(m_osContextDataCache.release, osData.release);
+        merge(m_osContextDataCache.displayVersion, osData.displayVersion);
+        merge(m_osContextDataCache.sysName, osData.sysName);
+        merge(m_osContextDataCache.kernelVersion, osData.kernelVersion);
+        merge(m_osContextDataCache.kernelRelease, osData.kernelRelease);
+        merge(m_osContextDataCache.cpeName, osData.cpeName);
```
jotacarma90: **[S3 · non-blocking · introduced]** The merge enumerates all 17 fields by hand, so adding a field to `OsContextData` (`scanContext.hpp:162`) silently makes it non-merging with no compiler error, and no document can ever clear a field again — a stale value can no longer be corrected by a newer document. `merge(m_osContextDataCache.cpeName, osData.cpeName)` is also dead, since `buildCPEName()` overwrites `cpeName` unconditionally twelve lines below. A table of `std::string OsContextData::*` member pointers iterated in a loop removes both the length and the hazard.

src/wazuh_modules/vulnerability_scanner/tests/mocks/TrampolineScanContext.hpp
```diff
@@ -74,6 +74,18 @@ struct TrampolineTScanContext final
         return 0;
     }
 
+    /// Usable by default: these tests drive the scan chains, not the OS-context precondition, so the
+    /// stub answers the way a session that carried its OS document would.
+    bool hasUsableOsContext() const noexcept
+    {
+        return true;
+    }
```
jotacarma90: **[S3 · non-blocking · introduced]** With the stub hardwired to `true` and `scanOrchestrator_test.cpp` unchanged, the new VDFirst early `return false` (`scanOrchestrator.hpp:362`) is never exercised — and that is the one addition whose failure mode is "the agent is never baselined". A regression that made the predicate return `false` more often would ship green. A case that runs a VDFirst session with the stub returning `false` and asserts that `deleteByQuery` is not called while the chain still runs would cover it.

Issue #35053, opened 2026-03-19
Field normalization for metrics snapshots


Miguevrgo:
# Field normalization for metrics snapshots (agents and comms)

|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0 | Wazuh agent | Agent | Packages/Sources | Any |

## Description

As discussed, we need to transform the raw variables collected for the metrics snapshots into the definitive field names expected by the Indexer. 

This issue covers the implementation of the data transformation logic for both `wazuh-metrics-agents` and `wazuh-metrics-comms`, ensuring the final payload matches the agreed-upon index mappings before ingestion. Once transformed, we must verify that the documents are correctly indexed without mapping conflicts.

## Data Transformations

The following mappings must be applied to the collected variables before sending the payload.

### `wazuh-metrics-agents`

| Raw Collected Variable (Old Index) | Final Index Field | Type | Notes |
| :--- | :--- | :--- | :--- |
| `timestamp` | `@timestamp` | `date` | |
| `id` | `wazuh.agent.id` | `keyword` | |
| `name` | `wazuh.agent.name` | `keyword` | |
| `version` | `wazuh.agent.version` | `keyword` | |
| `group` | `wazuh.agent.groups` | `keyword` | Array format. |
| `ip` | `wazuh.agent.host.ip` | `ip` | Array format. |
| `registerIP` | `wazuh.agent.register.ip` | `ip` | Convert `any` -> `0.0.0.0/0`. |
| `status` | `wazuh.agent.status` | `keyword` | `active`, `pending`, `disconnected`, `never_connected`. |
| `status_code` | `wazuh.agent.status_code` | `integer` | |
| `dateAdd` | `wazuh.agent.registered_at` | `date` | |
| `lastKeepAlive` | `wazuh.agent.last_seen` | `date` | |
| `disconnection_time` | `wazuh.agent.disconnected_at` | `date` | |
| `configSum` | `wazuh.agent.config.hash.md5` | `keyword` | |
| `group_config_status` | `wazuh.agent.config.group.synced` | `boolean` | Convert `synced` -> `true`, else `false`. |
| `mergedSum` | `wazuh.agent.config.group.hash.md5` | `keyword` | |
| `os.arch` | `wazuh.agent.host.architecture` | `keyword` | |
| `os.name` | `wazuh.agent.host.os.name` | `keyword` | |
| `os.version` | `wazuh.agent.host.os.version` | `keyword` | |
[...]

PR #38980, opened 2026-09-07
Populate wazuh.agent.groups in the agent metrics stream

### LINKED ISSUE #38900: `wazuh.agent.groups` is always empty in the agent metrics stream, while three other sources report it correctly
|---|---|---|---|---|
| 5.0.0 | wazuh-manager (agent metrics emission, `wazuh-metrics-agents`) | Manager | Docker, `public.ecr.aws/wazuh-cicd/wazuh/*:5.0.0-latest` | Amazon Linux 2023 |

**Severity:** Low

## Description

Every document in `wazuh-metrics-agents` carries `wazuh.agent.groups: []`. The same agents are reported as belonging to `default` by the Server API, by their events and by their state documents, all at the same moment.

The rest of each metrics document is accurate, which is what makes this specific: id, name, version, status, addresses, architecture and OS are all correct, and only the group list is empty.

Stock single-node deployment. Both agents are in `default`, which is where enrollment puts them and which nothing removed. Some were additionally placed in other groups during the session, so the field had opportunities to be non-empty.

## Steps to reproduce

```bash
# the metrics stream
curl -sk -u admin:admin 'https://localhost:9200/wazuh-metrics-agents/_search?size=1&sort=@timestamp:desc' \
  -d '{"_source":["wazuh.agent.groups","wazuh.agent.id"]}'
# groups: []

# the same fact, three other ways
curl -sk -u admin:admin 'https://localhost:9200/wazuh-events*/_search?size=1' \
  -d '{"query":{"term":{"wazuh.agent.id":"036"}},"_source":["wazuh.agent.groups"]}'
# groups: ["default"]

# and over the whole stream, to show it is never populated
curl -sk -u admin:admin 'https://localhost:9200/wazuh-metrics-agents/_search?size=0' \
  -d '{"aggs":{"g":{"terms":{"field":"wazuh.agent.groups","size":10}}}}'
# no buckets
```

## Current behavior

The same agent, the same minute, four sources:

```
Server API  /agents?agents_list=036       group:  ['default']
events      wazuh-events*                 groups: ["default"]
states      wazuh-states-inventory-*      groups: ["default"]
metrics     wazuh-metrics-agents          groups: []
```

It is not intermittent. Aggregating the field over the whole stream:

```
documents in wazuh-metrics-agents:  222
distinct values of wazuh.agent.groups: none, empty in every document
documents per agent id: 003:42  002:29  006:28  035:26  036:26  037:26  034:25  005:12  001:8
```

Nine different agent ids over the life of the deployment, 222 documents, and the field is never populated once.

A full document, with everything else correct:

```json
{"@timestamp": "2026-08-24T20:32:53Z",
 "wazuh": {"agent": {"id": "036", "name": "wazuh-agent-agent-11", "version": "v5.0.0",
                     "groups": [],
                     "status": "active", "status_code": 0,
                     "last_seen": "1787603534", "disconnected_at": "0",
                     "host": {"ip": ["172.20.0.5"], "architecture": "x86_64",
                              "os": {"name": "Amazon Linux", "version": "2023", "platform": "amzn"}}}}}
```

## Expected results

The field carries the same value the other three sources carry, or it is not emitted at all. An empty array asserts that the agent belongs to no group, which is never true: every agent is in `default` at minimum.

The same applies to `wazuh.agent.config.hash.md5`, `config.group.hash.md5`, `config.group.synced` and `host.os.full`: populate them or drop them from the mapping. `config.group.synced` in particular is a boolean that is never `true` and never `false`, so no consumer can tell "not synced" from "not reported".

### It is one field, not a systemic problem

Worth bounding, since a single wrong field raises the question of what else is wrong. Comparing every metadata field the four sources have in common, for the same agent:

```
field                events        findings      states        metrics
groups               ['default']   ['default']   ['default']   []          <-- disagree
host.architecture    x86_64        x86_64        x86_64        x86_64
host.os.name         Amazon Linux  Amazon Linux  Amazon Linux  Amazon Linux
host.os.platform     amzn          amzn          amzn          amzn
host.os.version      2023          2023          2023          2023
id                   036           036           036           036
name                 wazuh-agent-agent-11  (identical in all four)
version              v5.0.0        v5.0.0        v5.0.0        v5.0.0
```

17 fields across the four sources, and `groups` is the only one where any two disagree. The fields that appear in only one source (`status`, `last_seen`, `registered_at`, `register.ip`, `disconnected_at` in metrics; `host.hostname` and `host.os.type` in the others) are not disagreements, just different coverage.

### Confirmed again with a non-default group, which is the stronger case

The original evidence used an agent in `default` only, where an empty array could be argued to be a defensible rendering of "no explicit group". That argument does not survive a real assignment.

Agent 002 was put into a second group through the API, and the assignment propagated correctly everywhere else:

```
POST /groups            {"group_id":"qa-group"}
PUT  /agents/002/group/qa-group

API   GET /agents?select=id,group   ->  002 -> ['default', 'qa-group']
events, most recent                 ->  groups: ['default', 'qa-group']
```

The metrics stream, sampled a minute later, still reports nothing for either agent:

```
@2026-08-25T20:26:48Z  id=001  groups=[]
@2026-08-25T20:26:48Z  id=002  groups=[]
```

So the field is not rendering "no groups" as an empty array. It is empty while the agent has two groups, one of which was assigned explicitly and is visible in the API and in the agent's own events at the same moment.

### Root cause, verified in source

The field is read from a document that never carries it. The metrics writer asks for a key named `group`:

```python
# framework/wazuh/core/indexer/metrics_snapshot.py:353
"groups": doc.get("group", []),
```

and `doc` comes from wazuh-db's `/v1/agents/all`:

```python
# framework/wazuh/core/wdb_http.py:171
async def get_all_agents(self) -> list[dict]:
    return await self._get('/agents/all')
```

whose SQL selects seventeen columns and no group at all:

```cpp
/* src/wazuh_db/src/http/endpointGetV1AgentsAll.hpp:81 */
"SELECT id, name, coalesce(ip, register_ip) as ip, connection_status as status, "
"os_name, os_version, os_type, os_platform, version, date_add, "
"os_major, os_minor, os_arch, last_keepalive, register_ip, "
"disconnection_time, status_code "
"FROM agent WHERE id > 0 ORDER BY id ASC;"
```

Group membership lives in a separate table and is exposed through a different, per-agent route:

```cpp
/* src/wazuh_db/src/http/wdb_http.cpp:114 */
registerRoute<EndpointGetV1AgentsParamGroups>(
    *server, wazuh::uds_http::Method::Get, "/v1/agents/groups", wazuh::uds_http::RouteClass::Control);
```

So `doc.get("group", [])` always takes its default. The empty list is not intermittent and not a timing artefact: it is the literal default of a lookup that can never hit. That is consistent with the measurement of 222 documents across 9 agent ids with no populated value, and with every other field in the document being correct, since every other field is in that SELECT.

**How to reproduce:**

```bash
# the metrics stream says the agent belongs to nothing
curl -sk -u admin:admin 'https://localhost:9200/wazuh-metrics-agents/_search?size=1&sort=@timestamp:desc' \
  -H 'Content-Type: application/json' -d '{"_source":["wazuh.agent.id","wazuh.agent.groups"]}'

# the API, the events and the state documents all say "default", at the same moment
curl -sk -H "Authorization: Bearer $TOKEN" 'https://localhost:55000/agents?agents_list=<id>&select=id,group'
```

**Fix, for scale:** either add the group join to the `/v1/agents/all` query, or drop the field from the metrics document rather than emitting an assertion that is always false.

### Correction: it is five fields in this index, not one

The section above is accurate about what it measures, cross-source disagreement on shared metadata, and it gave the wrong overall impression. Sweeping every mapped leaf field in `wazuh-metrics-agents` for ones no document populates shows five of twenty-three are always empty, across all 403 documents:

```
   403  @timestamp, wazuh.agent.id, name, version, status, status_code,
        last_seen, registered_at, disconnected_at, register.ip,
        host.ip, host.architecture, host.os.name, host.os.platform,
        host.os.version, wazuh.cluster.name, wazuh.cluster.node,
        wazuh.schema.version
 EMPTY  wazuh.agent.groups
 EMPTY  wazuh.agent.config.hash.md5
 EMPTY  wazuh.agent.config.group.hash.md5
 EMPTY  wazuh.agent.config.group.synced
 EMPTY  wazuh.agent.host.os.full
```

The three `config.*` fields matter more than `groups` does. Together they are the only place this document would report **whether an agent is actually running the configuration it was sent**, and its hash. That is the question an operator asks after pushing a group configuration, and there is no answer anywhere: the API's `/agents` endpoint does not offer those fields either, and says so when asked.

```
GET /agents?select=id,group_config_status,mergedSum,configSum
  "Not a valid select field: Allowed select fields: dateAdd, disconnection_time,
   group, id, ip, lastKeepAlive, name, os.*, registerIP, status, status_code, version"
```

`mergedSum` and `configSum` exist at group level, from `GET /groups`, but nothing reports them per agent.

**A correlation, not a cause.** The agent logs a coordination failure for exactly this data, continuously. 465 occurrences on one agent, still going:

```
2026/08/26 09:03:28 wazuh-modulesd:agent-info: INFO: Failed to coordinate agent_groups, will retry in next cycle
```

That message accompanies the `syscollector flush failed: VD queue` cascade in [037](not-for-filing/duplicated/DUPLICATED_037-vd-feed-incomplete-blocks-package-inventory.md), so it may be a downstream symptom of that rather than an independent fault. Group *membership* propagates correctly despite it, which was checked and is recorded above, so the message is not simply telling the truth about groups either. Whether the empty `config.*` fields and these 465 failures are the same defect was not established.

### Impact

Small and contained: the information is correct in three other places, so nothing is lost outright. What breaks is anything that reads this stream and expects the field to mean what it says. `wazuh-metrics-agents*` is a provisioned index pattern, so it is a consumer-facing surface, and a panel or query that filters agent metrics by group finds nothing rather than finding everything.

The more general cost is that the same field disagrees between sources, which is the kind of thing that makes an investigation take twice as long when it eventually matters.

### Tasks

- [ ] Populate `wazuh.agent.groups` in the agent metrics emission from the same place the API and the event pipeline read it, or drop the field from the document rather than emitting an empty array.
- [ ] Check the rest of the metrics document against the API for other fields that are present but never filled; `register.ip` reads `0.0.0.0` while `host.ip` carries the real address, which may be correct for a registration with `any` but is worth confirming.
- [x] Compared every metadata field the four sources share, to bound the problem. `groups` is the only disagreement among 17 fields; everything else matches exactly. Table above.
- [ ] Keep that comparison as a standing check, since this one was noticed only by reading a document rather than by anything failing.

cc @wazuh/devel-xdrsiem-server

jotacarma90:
## Description

Every document indexed into `wazuh-metrics-agents` reported `wazuh.agent.groups: []`, while the `/agents` API, the events and the state documents reported the real group list for the same agent at the same moment.

The metrics snapshot reads the field as `doc.get("group", [])` from the rows returned by wazuh-db's `/v1/agents/all`, and that endpoint's `SELECT` did not include the agent's `group` column. The lookup could never hit, so the field was always the literal default: not intermittent, and not a timing artefact.

Part of #38900. The remaining four always-empty leaves the issue lists (`wazuh.agent.config.hash.md5`, `config.group.hash.md5`, `config.group.synced`, `host.os.full`) are tracked separately; see *Out of scope* below.

## Proposed Changes

- `src/wazuh_db/src/http/endpointGetV1AgentsAll.hpp`: add the agent table's `` `group` `` column to the query and expose it as the `group` field of the response, next to the seventeen fields it already returns. The reflective serializer omits empty strings, so an agent with no groups produces no `group` key.
- `framework/wazuh/core/indexer/metrics_snapshot.py`: split that comma-separated value into a list, the same way `WazuhDBQueryAgents` does for the `/agents` API. A list already coming from the legacy query path is passed through unchanged. An agent with no groups leaves `wazuh.agent.groups` out of the document instead of emitting an empty array, which asserted something no other source asserts.

- `src/unit_tests/wazuh_db/test_wdb_http_endpoints.cpp`: `GetAllAgentsEmptyDbReturnsEmptyArray` pins the endpoint's SQL text verbatim, so the expected string gains the new column, and two cases cover the field: a comma-separated value is serialized as-is, and an empty one is omitted rather than sent as an empty string, which is the contract the normalizer relies on.
- `framework/wazuh/core/indexer/metrics_snapshot.py` (`_to_iso`): `/agents/all` returns `dateAdd`, `lastKeepAlive` and `disconnection_time` as epoch seconds in digit strings, and the normalizer passed them through unchanged. The index maps `registered_at`, `last_seen` and `disconnected_at` as `date` with no `format`, so OpenSearch read `"1787603534"` as epoch milliseconds and dated every agent in January 1970; `disconnection_time` `"0"` (wazuh-db's "never") was indexed as a date too. Epochs are now converted to UTC ISO 8601 and 0 is treated as absent. Found in review; the issue's own quoted document shows the raw values.
- `framework/wazuh/core/indexer/metrics_snapshot.py` (`host.ip`): an agent with no address emitted `host.ip: []` while `groups` now leaves the field out. Absence has one shape in the document: the field is omitted. The group filter also applies to a list already split upstream, so a `[""]` from `WazuhDBQueryAgents` cannot be indexed as one empty group.
- `tests/integration/test_wazuh_db/test_databases/test_wazuh_db.py`: disables the task manager's scheduled agent disconnection sweep for the module through `wazuh_modules.manager_task_monitor_agents`. See the CI section below for why.

`` `group` `` is the same column the `/agents` API reads (`Agent.fields["group"]`), and wazuh-db updates it together with `group_hash` on every group change (`WDB_STMT_GLOBAL_GROUP_CTX_SET`), so both surfaces now derive the field from one source.

### Results and Evidence

Normalizer output before the change, given the rows `/v1/agents/all` returns today (no `group` key):

```
id=001  groups=[]
id=002  groups=[]
```

And after, given the rows it returns now:

```
id=001  groups=["default"]
id=002  groups=["default", "qa-group"]
id=003  groups=<field not emitted>
```

Dates, before and after, for the values in the document quoted in #38900 (`@timestamp` there is `2026-08-24T20:32:53Z`):

```
last_seen        "1787603534"  ->  "2026-08-24T20:32:14Z"   (was read by OpenSearch as 1970-01-21T16:33:23Z)
disconnected_at  "0"           ->  <field not emitted>       (was indexed as 1970-01-01T00:00:00Z)
host.ip (none)   []            ->  <field not emitted>
```

The new `SELECT` run against a database created from `schemas/schema_global.sql`, with agent 3 left with a `NULL` group:

```
$ sqlite3 -json global.db "SELECT id, name, coalesce(ip, register_ip) as ip, connection_status as status,
  os_name, os_version, os_type, os_platform, version, date_add, os_major, os_minor, os_arch,
  last_keepalive, register_ip, disconnection_time, status_code, `group`
  FROM agent WHERE id > 0 ORDER BY id ASC;"
1 'default'
2 'default,qa-group'
3 None
```

Unit tests:

```
$ python -m pytest wazuh/core/indexer/tests/test_metrics_snapshot.py -q
128 passed in 0.25s

$ python -m pytest wazuh/core/indexer/tests/test_metrics_snapshot.py -k Groups -v
TestNormalizeAgentDocGroups::test_comma_separated_group_becomes_a_list PASSED
TestNormalizeAgentDocGroups::test_single_group_becomes_a_single_entry_list PASSED
TestNormalizeAgentDocGroups::test_legacy_list_group_is_kept PASSED
TestNormalizeAgentDocGroups::test_missing_group_leaves_the_field_out PASSED
TestNormalizeAgentDocGroups::test_empty_group_leaves_the_field_out PASSED
5 passed, 118 deselected in 0.08s
```

The framework unit tests on CI, dispatched on this branch because a draft PR skips them: https://github.com/wazuh/wazuh/actions/runs/34118929901 (`build (3.12)`, success).

`clang-format-18 --dry-run --Werror` is clean on the modified header. The C++ change is not compiled by any workflow a draft PR runs; it builds with the manager in `5_testunit_linux-win` once the PR is marked ready for review.

#### CI

First full pass on this branch (rebased onto `5.0.0` at `06a4821222`): every check green, including `Unit-Tests (manager)`, which builds wazuh-db and runs `test_wdb_http_endpoints`, except one:

```
Integration tests (wazuh_db, manager_install, test_wazuh_db/, --tier 0) / Run wazuh_db integration tests
FAILED test_wazuh_db/test_databases/test_wazuh_db.py::test_wazuh_db_messages_global[disconnect-agents command]
  stage 9: global disconnect-agents 0 175 syncreq   Expected: ok [{"id":1},{"id":2}]   Response: ok []
1 failed, 103 passed in 774.92s
```

That failure is not caused by this change. `disconnect-agents` runs `SELECT id FROM agent WHERE ... last_keepalive < ?` (`wdb.c:87`), which does not involve the `group` column or the HTTP endpoint. It is a race introduced in the base branch by #38950 (merged 2026-09-04, after this workflow last ran on `5.0.0`): the task manager's disconnection sweep used to run once per `agents_disconnection_time` (900 s by default) and now runs every `agents_disconnection_time / 4`, bounded to 60..300 s (`builtinTypes.hpp:94-96`). The IT starts the whole manager (`daemons_handler_configuration = {'all_daemons': True}`), plants keepalives of 100, 150 and 200 and asserts on connection status over ~9 minutes, so the sweep now lands between stages and rewrites the rows the test owns. Both stages it can hit were observed:

| Run | Head | Result |
|---|---|---|
| [34228351901](https://github.com/wazuh/wazuh/actions/runs/34228351901) attempt 1 | this PR | stage 9: `ok []` (agents 1 and 2 already swept) |
| [34228351901](https://github.com/wazuh/wazuh/actions/runs/34228351901) attempt 2 | this PR | stage 11: `[{"id":1},{"id":2},{"id":3}]` (agent 3 swept between stages 9 and 11) |
| [34232613293](https://github.com/wazuh/wazuh/actions/runs/34232613293) `workflow_dispatch`, `modules=wazuh_db` | clean `origin/5.0.0`, no commit from this PR | stage 11, same as attempt 2 |

The third commit fixes the test rather than the product: the cases drive `disconnect-agents` through the socket themselves and never rely on the scheduled sweep, so the module turns the sweep off with `wazuh_modules.manager_task_monitor_agents = 0` (`sweep.enabled = config.monitor_agents != 0`, `builtinTypes.cpp:257`) before the daemons restart. The module-scoped `configure_local_internal_options_module` fixture cannot be used for that because it requests `test_metadata`, which this module parametrizes at function scope, hence the autouse fixture. `cases_global_messages.yaml` is the only case file in `test_wazuh_db/` that touches keepalives or connection status.

### Artifacts Affected

- `wazuh-db` (the `/v1/agents/all` response gains a `group` field)
- `wazuh-server` framework (documents indexed into `wazuh-metrics-agents`)
- `tests/integration/test_wazuh_db` (the global messages IT no longer races the disconnection sweep)

### Configuration Changes

None.

### Documentation Updates

None. `docs/ref/modules/wazuh_db/` does not document the endpoint's response fields, and the `wazuh-metrics-agents` mapping already declares `wazuh.agent.groups`.

### Tests Introduced

`GetAllAgentsSerializesGroupCsv` and `GetAllAgentsOmitsEmptyGroup` in `src/unit_tests/wazuh_db/test_wdb_http_endpoints.cpp`, plus the pinned SQL text updated in `GetAllAgentsEmptyDbReturnsEmptyArray`.

`TestNormalizeAgentDocEpochDates` (digit-string epoch, int epoch, `"0"` omitted, ISO passthrough) and `TestNormalizeAgentDocAbsentValues` (no ip leaves `host.ip` out), and `TestNormalizeAgentDocGroups` in `framework/wazuh/core/indexer/tests/test_metrics_snapshot.py`, covering the five inputs the normalizer can receive: a single group, several comma-separated groups, a list from the legacy query path, an absent `group` key and an empty one. The last two pin the behaviour that no group means no field rather than an empty array.

### Out of scope

Tracked in wazuh/wazuh-indexer-plugins#1558. The issue also lists `wazuh.agent.config.hash.md5`, `config.group.hash.md5`, `config.group.synced` and `host.os.full` as never populated. None of them has a source in the `agent` table (`configSum` and `os.uname` do not exist there, and `group_hash` / `group_sync_status` are the group-list hash and the cluster sync flag, not the merged configuration checksum an operator would read from `config.group.*`), so filling them is a separate change and not a one-column fix. They were part of the stream's original requirement (#35053, wazuh-indexer-plugins#940) and the columns behind them were removed afterwards (`f0132e08824`, `0ee1cf977c7`), so the follow-up asks Indexer to drop the leaves rather than the manager to fill them. `wazuh.agent.register.ip` reading `0.0.0.0` is deliberate: the normalizer maps a `registerIP` of `any` to it.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/framework/wazuh/core/indexer/metrics_snapshot.py
+++ b/framework/wazuh/core/indexer/metrics_snapshot.py
@@ -326,14 +326,26 @@ class MetricsSnapshotTasks:
     def _to_iso(value) -> str | None:
         """Convert a value to an ISO 8601 string.
 
-        Handles ``datetime`` objects returned by ``WazuhDBQueryAgents`` as
-        well as plain strings.  Returns *None* for falsy values so that
-        ``_drop_none`` can remove the field.
+        Handles ``datetime`` objects returned by ``WazuhDBQueryAgents``, the
+        epoch seconds the ``/agents/all`` endpoint returns as digit strings
+        (``dateAdd``, ``lastKeepAlive``, ``disconnection_time``) and strings
+        that are already formatted. Returns *None* for absent values and for
+        an epoch of 0, which is how wazuh-db stores "never", so that
+        ``_drop_none`` can remove the field. The index maps these leaves as
+        ``date`` with no format, so a raw digit string would be parsed as
+        epoch milliseconds and land in 1970.
         """
         if value is None:
             return None
         if isinstance(value, datetime):
             return value.strftime("%Y-%m-%dT%H:%M:%SZ")
+        if isinstance(value, bool):
+            return None
+        if isinstance(value, (int, float)) or (isinstance(value, str) and value.isdigit()):
+            epoch = int(value)
+            if epoch <= 0:
+                return None
+            return datetime.fromtimestamp(epoch, tz=timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
         return str(value) if value else None
 
     @staticmethod
@@ -383,7 +395,9 @@ class MetricsSnapshotTasks:
         # an empty group list.
         raw_groups = doc.get("group")
         if isinstance(raw_groups, str):
-            raw_groups = [group for group in raw_groups.split(",") if group]
+            raw_groups = raw_groups.split(",")
+        if raw_groups is not None:
+            raw_groups = [group for group in raw_groups if group]
 
         return MetricsSnapshotTasks._drop_none(
             {
@@ -407,7 +421,7 @@ class MetricsSnapshotTasks:
                         ),
                         "register": {"ip": register_ip},
                         "host": {
-                            "ip": [ip] if ip else [],
+                            "ip": [ip] if ip else None,
                             "architecture": os_fields.get("arch"),
                             "os": {
                                 "name": os_fields.get("name"),
--- a/framework/wazuh/core/indexer/metrics_snapshot.py
+++ b/framework/wazuh/core/indexer/metrics_snapshot.py
@@ -377,6 +377,14 @@ class MetricsSnapshotTasks:
         raw_id = doc.get("id")
         agent_id = str(raw_id).zfill(3) if raw_id is not None else None
 
+        # Groups come from the agent table's comma-separated `group` column, the same source
+        # the /agents API reads: WazuhDBQueryAgents splits it, the /agents/all HTTP endpoint
+        # returns it raw. An agent with no groups leaves the field out instead of asserting
+        # an empty group list.
+        raw_groups = doc.get("group")
+        if isinstance(raw_groups, str):
+            raw_groups = [group for group in raw_groups.split(",") if group]
+
         return MetricsSnapshotTasks._drop_none(
             {
                 "@timestamp": MetricsSnapshotTasks._to_iso(doc.get("@timestamp")),
@@ -385,7 +393,7 @@ class MetricsSnapshotTasks:
                         "id": agent_id,
                         "name": doc.get("name"),
                         "version": doc.get("version"),
-                        "groups": doc.get("group", []),
+                        "groups": raw_groups or None,
                         "status": doc.get("status"),
                         "status_code": doc.get("status_code"),
                         "registered_at": MetricsSnapshotTasks._to_iso(
--- a/framework/wazuh/core/indexer/metrics_snapshot.py
+++ b/framework/wazuh/core/indexer/metrics_snapshot.py
@@ -390,9 +390,10 @@ class MetricsSnapshotTasks:
         agent_id = str(raw_id).zfill(3) if raw_id is not None else None
 
         # Groups come from the agent table's comma-separated `group` column, the same source
-        # the /agents API reads: WazuhDBQueryAgents splits it, the /agents/all HTTP endpoint
-        # returns it raw. An agent with no groups leaves the field out instead of asserting
-        # an empty group list.
+        # the /agents API reads. The /agents/all HTTP endpoint returns it raw; a caller that
+        # already split it (the WazuhDBQueryAgents shape this function also accepts) goes
+        # through the same filter. An agent with no groups leaves the field out instead of
+        # asserting an empty group list.
         raw_groups = doc.get("group")
         if isinstance(raw_groups, str):
             raw_groups = raw_groups.split(",")
--- a/framework/wazuh/core/indexer/tests/test_metrics_snapshot.py
+++ b/framework/wazuh/core/indexer/tests/test_metrics_snapshot.py
@@ -2064,6 +2064,61 @@ class TestNormalizeAgentDocIdZeroPadding:
 
 
 # ---------------------------------------------------------------------------
+# Agent groups (regression: /agents/all returns the raw comma-separated `group`
+# column, so the field must be split, and must be absent when the agent has none)
+# ---------------------------------------------------------------------------
+
+
+class TestNormalizeAgentDocGroups:
+    """wazuh.agent.groups must carry the same value the /agents API reports."""
+
+    @pytest.mark.asyncio
+    async def test_comma_separated_group_becomes_a_list(self):
+        """The comma-separated `group` column is split into one entry per group."""
+        tasks = _make_tasks()
+        with _agents_http_patch([{"id": 2, "name": "agent2", "group": "default,qa-group"}]):
+            docs = await tasks._collect_agents(TIMESTAMP)
+
+        assert docs[0]["wazuh"]["agent"]["groups"] == ["default", "qa-group"]
+
+    @pytest.mark.asyncio
+    async def test_single_group_becomes_a_single_entry_list(self):
+        """A single group name yields a one-element list, not the bare string."""
+        tasks = _make_tasks()
+        with _agents_http_patch([{"id": 1, "name": "agent1", "group": "default"}]):
+            docs = await tasks._collect_agents(TIMESTAMP)
+
+        assert docs[0]["wazuh"]["agent"]["groups"] == ["default"]
+
+    @pytest.mark.asyncio
+    async def test_legacy_list_group_is_kept(self):
+        """WazuhDBQueryAgents already returns a list; it is passed through unchanged."""
+        tasks = _make_tasks()
+        with _agents_http_patch([{"id": 1, "name": "agent1", "group": ["default", "qa-group"]}]):
+            docs = await tasks._collect_agents(TIMESTAMP)
+
+        assert docs[0]["wazuh"]["agent"]["groups"] == ["default", "qa-group"]
+
+    @pytest.mark.asyncio
+    async def test_missing_group_leaves_the_field_out(self):
+        """No group in the raw doc means no wazuh.agent.groups, not an empty array."""
+        tasks = _make_tasks()
+        with _agents_http_patch([{"id": 1, "name": "agent1"}]):
+            docs = await tasks._collect_agents(TIMESTAMP)
+
+        assert "groups" not in docs[0]["wazuh"]["agent"]
+
+    @pytest.mark.asyncio
+    async def test_empty_group_leaves_the_field_out(self):
+        """An empty `group` column is not rendered as an empty array either."""
+        tasks = _make_tasks()
+        with _agents_http_patch([{"id": 1, "name": "agent1", "group": ""}]):
+            docs = await tasks._collect_agents(TIMESTAMP)
+
+        assert "groups" not in docs[0]["wazuh"]["agent"]
+
+
+# ---------------------------------------------------------------------------
 # _normalize_comms_doc – zero value preservation (FR-2 fix)
 # ---------------------------------------------------------------------------
 
--- a/framework/wazuh/core/indexer/tests/test_metrics_snapshot.py
+++ b/framework/wazuh/core/indexer/tests/test_metrics_snapshot.py
@@ -2119,6 +2119,80 @@ class TestNormalizeAgentDocGroups:
 
 
 # ---------------------------------------------------------------------------
+# Agent dates from /agents/all (regression: the endpoint returns epoch seconds as
+# digit strings, and the index maps these leaves as `date` with no format, so a raw
+# digit string is read as epoch milliseconds and lands in 1970; "0" means never)
+# ---------------------------------------------------------------------------
+
+
+class TestNormalizeAgentDocEpochDates:
+    """registered_at, last_seen and disconnected_at must be ISO 8601 or absent."""
+
+    @pytest.mark.asyncio
+    async def test_epoch_string_becomes_iso(self):
+        """A digit-string epoch (the /agents/all shape) is converted to UTC ISO 8601."""
+        tasks = _make_tasks()
+        with _agents_http_patch([{"id": 1, "name": "a", "lastKeepAlive": "1787603534"}]):
+            docs = await tasks._collect_agents(TIMESTAMP)
+
+        assert docs[0]["wazuh"]["agent"]["last_seen"] == "2026-08-24T20:32:14Z"
+
+    @pytest.mark.asyncio
+    async def test_epoch_int_becomes_iso(self):
+        """An int epoch is converted the same way."""
+        tasks = _make_tasks()
+        with _agents_http_patch([{"id": 1, "name": "a", "dateAdd": 1787603534}]):
+            docs = await tasks._collect_agents(TIMESTAMP)
+
+        assert docs[0]["wazuh"]["agent"]["registered_at"] == "2026-08-24T20:32:14Z"
+
+    @pytest.mark.asyncio
+    async def test_zero_epoch_string_leaves_the_field_out(self):
+        """disconnection_time "0" is wazuh-db's "never" and must not be indexed as a date."""
+        tasks = _make_tasks()
+        with _agents_http_patch([{"id": 1, "name": "a", "disconnection_time": "0"}]):
+            docs = await tasks._collect_agents(TIMESTAMP)
+
+        assert "disconnected_at" not in docs[0]["wazuh"]["agent"]
+
+    @pytest.mark.asyncio
+    async def test_iso_string_is_kept(self):
+        """A value that is already ISO 8601 passes through unchanged."""
+        tasks = _make_tasks()
+        with _agents_http_patch([{"id": 1, "name": "a", "lastKeepAlive": "2026-03-17T10:00:00Z"}]):
+            docs = await tasks._collect_agents(TIMESTAMP)
+
+        assert docs[0]["wazuh"]["agent"]["last_seen"] == "2026-03-17T10:00:00Z"
+
+
+# ---------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


framework/wazuh/core/indexer/metrics_snapshot.py
```diff
@@ -385,7 +393,7 @@ def _normalize_agent_doc(doc: dict) -> dict:
                         "id": agent_id,
                         "name": doc.get("name"),
                         "version": doc.get("version"),
-                        "groups": doc.get("group", []),
+                        "groups": raw_groups or None,
```
Antoniogm03, line 411: **[S2 · blocking · introduced]** This line fixes `groups`, but the PR says `Closes #38900` while four of the five always-empty fields that issue enumerates — `wazuh.agent.config.hash.md5`, `config.group.hash.md5`, `config.group.synced` and `host.os.full` — remain unpopulated, and the issue's own correction section calls the three `config.*` fields more important than `groups`. The *Out of scope* reasoning is correct, and I checked it: none of them has a source in the `agent` table, and `group_hash` is the SHA256 of the group CSV (`wdb_global_recalculate_agent_groups_hash`), not a merged-configuration checksum — so the code is right, the closure is not. Change `Closes #38900` to `Part of #38900` and open a follow-up issue for the remaining four (the *Out of scope* text is already its body), or split #38900 and link this PR to the `groups` half.

jotacarma90: Agreed on the closure semantics, and thanks for checking the `group_hash` derivation: `wdb_global_recalculate_agent_groups_hash` runs `OS_SHA256_String_sized` over the group CSV, so it is neither an MD5 nor a configuration checksum.

Changed to `Part of #38900`. The follow-up for the four remaining leaves is wazuh/wazuh-indexer-plugins#1558: they were in the stream's requirement (wazuh-indexer-plugins#940, mapped in #35053), but the columns behind them were removed afterwards on purpose (`config_sum` and `os_uname` in `f0132e08824`, `merged_sum` and `group_config_status` in `0ee1cf977c7`, "now agent is the owner of config changes"), and the `.md5` names cannot hold the SHA-256 the manager actually computes, so the ask there is to drop them from the metrics subset rather than fill them.

While re-reading the normalizer for this I also fixed two things in the same function, pushed as separate commits so they are easy to drop if you prefer them elsewhere:

- `_to_iso` passed the endpoint's digit-string epochs through unchanged. The three date leaves are mapped as `date` with no `format`, so `"1787603534"` was being read as epoch milliseconds (1970-01-21) and `disconnection_time` `"0"` was indexed as a date; the document quoted in #38900 shows both. Epochs are now converted to UTC ISO 8601 and 0 is treated as absent.
- `host.ip` emitted `[]` for an agent with no address while `groups` now omits the field; both use the same shape for absence now. The group filter also applies to a list already split upstream, so a `[""]` from `WazuhDBQueryAgents` cannot be indexed as one empty group (wazuh-db writes NULL rather than `""` for no groups, so that path was theoretical, but the two branches are symmetric now).

Also tightened the comment on `GetAllAgentsOmitsEmptyGroup` (the mock feeds the empty string, so it covers the serializer, not the wrapper's NULL handling) and moved the autouse fixture out of the IT docstring's `parameters:`.

PR #39003, opened 2026-09-07
Report the loopback interface with its type and an up state

### LINKED ISSUE #38936: The loopback interface is inventoried with no type and the kernel's literal "unknown" as its state
Half the rows of `wazuh-states-inventory-interfaces` are the loopback interface, and in each of them `interface.type` is null and `interface.state` is the string `unknown`. Loopback has a well-defined type and is up on every running host. `interface.alias` is null in all four rows.

|---|---|---|---|---|
| 5.0.0-latest | wazuh-agent network interface inventory collector | Agent | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 agent, single-node stack |

**Severity:** Low to Medium. `interface.state` is the field a query uses to find interfaces that are up, and it returns a value that is neither up nor down for a class of interface present on every host.

## Every row in the index

```
   interface={"name": "eth0", "alias": null, "state": "up",      "type": "ethernet", "mtu": 1500}   agent=fresh-agent
   interface={"name": "lo",   "alias": null, "state": "unknown", "type": null,       "mtu": 65536}  agent=fresh-agent
   interface={"name": "eth0", "alias": null, "state": "up",      "type": "ethernet", "mtu": 1500}   agent=ec-sat-agent
   interface={"name": "lo",   "alias": null, "state": "unknown", "type": null,       "mtu": 65536}  agent=ec-sat-agent
```

`eth0` is classified correctly on both agents, so the collector's type and state logic works and has simply no case for loopback.

## `unknown` is the kernel's word, passed through

Linux reports `unknown` in `/sys/class/net/lo/operstate` because the loopback driver does not implement carrier detection. That is a property of the driver, not of the interface's usability: `lo` carries traffic, and this deployment's own numbers prove it, with 10,915 bytes in and the same 10,915 out on one agent and 2,222 each way on the other.

The collector is passing the kernel string into `interface.state` verbatim. So the field's vocabulary is whatever `/sys` happens to contain, which for `operstate` includes `unknown`, `notpresent`, `dormant`, `lowerlayerdown` and `testing` besides `up` and `down`. A consumer filtering `interface.state: up` to enumerate live interfaces misses loopback on every host in the fleet.

## `interface.type` null where the type is knowable

The type of `lo` is loopback, and the kernel says so directly: `/sys/class/net/lo/type` is 772, `ARPHRD_LOOPBACK`, in the same place the collector reads 1, `ARPHRD_ETHER`, to classify `eth0` as ethernet. The mapping for 772 is missing rather than the information.

## `interface.alias` is null in every row

`interface.alias` is a declared leaf populated in none of the four rows. On Linux the alias is `/sys/class/net/<name>/ifalias`, which is empty unless an operator sets it, so null is arguably correct here. It is noted because it means the column is empty in this deployment and a consumer cannot distinguish "no alias set" from "the collector does not read aliases".

## Expected behavior

`lo` is typed `loopback`. `interface.state` carries a normalised vocabulary in which an interface that carries traffic is up, with the raw `operstate` kept in a separate field if it is wanted.

## Steps to reproduce

1. `POST /wazuh-states-inventory-interfaces/_search` with `{"query":{"match_all":{}}}`. Four rows, two of them `lo`.
2. Read `interface.type` and `interface.state` on the `lo` rows: null and `unknown`.
3. On the agent, `cat /sys/class/net/lo/operstate` returns `unknown` and `cat /sys/class/net/lo/type` returns 772.
4. Read `host.network.ingress.bytes` on the same rows to confirm the interface is passing traffic.

## Related

[204](BUG_204-the-hardware-inventory-leaves-the-memory-usage-column-empty-with-the-three-numbers-that-define-it-in-the-same-document.md) is the hardware inventory's empty computable column. [183](already-reported/BUG_183-the-remote-desktop-decoder-records-the-listener-name-as-a-user-and-keeps-the-domain-glued-to-the-account.md) and [135](already-reported/BUG_135-the-users-inventory-marks-every-local-system-account-as-remote-and-leaves-six-account-hygiene-columns-empty-in-every-row.md) are other inventory values that are wrong rather than absent.

cc @wazuh/devel-xdrsiem-agent

anromerom:
## Description

Closes #38936

The network interface inventory reports the loopback interface with no type and with the kernel's literal `unknown` as its state. Both values come straight from the collector: the interface type map has no entry for the loopback code, and the Linux collector passes `/sys/class/net/<name>/operstate` through unchanged. This PR types loopback on Linux, macOS and Windows, and gives the state field a single `up`/`down` vocabulary on Linux.

### `interface.type`

The type is read from the kernel as a numeric code and translated through a per-platform map. Loopback was missing from all three maps, so the lookup returned an empty value and the field was reported with no type on every platform.

- `src/data_provider/src/network/networkLinuxWrapper.h`: added `ARPHRD_LOOPBACK` to `NETWORK_INTERFACE_TYPE`, mapping to `loopback`.
- `src/data_provider/src/network/networkBSDWrapper.h`: added `IFT_LOOP` to `NETWORK_INTERFACE_TYPE`, mapping to `loopback`.
- `src/data_provider/src/network/networkWindowsWrapper.h`: added `IF_TYPE_SOFTWARE_LOOPBACK` to `NETWORK_INTERFACE_TYPES`, mapping to `loopback`.

No other interface type changes: each entry covers only the loopback code for its platform.

### `interface.state`

On Linux the field carried the raw `operstate` value, so besides `up` and `down` it could report `unknown`, `dormant`, `notpresent`, `lowerlayerdown` or `testing`. The loopback driver has no carrier detection and therefore reports `unknown`, which made loopback invisible to a query for interfaces that are up. macOS already derives this field from the administrative flag and reports `up` for the same interface.

- `src/data_provider/src/network/networkLinuxWrapper.h`: `NetworkLinuxInterface::state()` now resolves `unknown` from the `IFF_UP` administrative flag, and reports `down` for the remaining operational states. Values of `up` and `down` are returned unchanged, and an interface whose `operstate` cannot be read keeps its previous placeholder.
- `src/data_provider/src/network/networkLinuxWrapper.h`: added `<net/if.h>` for `IFF_UP`.

Only `unknown` consults the administrative flag. The kernel already reports `down` whenever an interface is not running, so testing `IFF_UP` for every non-`up`/`down` value would always resolve to `up` and would report a dormant or lower-layer-down interface as up.

### Tests

- `src/data_provider/tests/sysInfoNetworkLinux/sysInfoNetworkLinux_test.cpp`: five cases covering both fields.

### Results and Evidence

Kernel inputs on an agent, showing the two values the collector reads:

```sql
eth0     type=1     operstate=up         flags=0x1003
lo       type=772   operstate=unknown    flags=0x9
```

`772` is `ARPHRD_LOOPBACK` and `flags=0x9` is `IFF_UP` together with `IFF_LOOPBACK`, so the kernel reports loopback as administratively up.

Collector output for every interface on the host, before and after the change:

```sql
=== BEFORE ===                        === AFTER ===
iface    type       state             iface    type       state
lo       <empty>    unknown           lo       loopback   up
eth0     ethernet   up                eth0     ethernet   up
docker0  ethernet   down              docker0  ethernet   down
br-6df43649a06e ethernet down         br-6df43649a06e ethernet down
```

Only the loopback row changes. The bridge interfaces are administratively up but have no carrier, so they report `operstate: down` and stay `down`, which is the case that a blanket use of the administrative flag would have wrongly reported as up.

Unit tests on the affected component:

```sql
[  PASSED  ] 12 tests.      sysInfoNetworkLinux_unit_test
100% tests passed, 0 tests failed out of 14      data_provider suite
```

With the change reverted and the tests left in place, the three behavioural cases fail, confirming they cover the fix:

```sql
[  FAILED  ] SysInfoNetworkLinuxTest.Test_LOOPBACK_INTERFACE_TYPE
[  FAILED  ] SysInfoNetworkLinuxTest.Test_LOOPBACK_STATE_IS_UP
[  FAILED  ] SysInfoNetworkLinuxTest.Test_LOOPBACK_STATE_WITHOUT_ADMIN_FLAG_IS_DOWN
```

The two remaining cases pass with and without the change, which is their purpose: they guard behaviour that must not move.

Verified on Linux. The macOS and Windows entries were not built or run on this platform.

### Artifacts Affected

- Agent inventory collector (`wazuh-modulesd`), network interface data provider.
- Platforms: Linux, macOS, Windows.
- Interface inventory fields `type` and `state`.

### Configuration Changes

N/A

### Documentation Updates

N/A

### Tests Introduced

Unit tests in `src/data_provider/tests/sysInfoNetworkLinux/sysInfoNetworkLinux_test.cpp`:

- `Test_LOOPBACK_INTERFACE_TYPE`: the loopback code resolves to `loopback`.
- `Test_ETHERNET_INTERFACE_TYPE_UNCHANGED`: the ethernet code still resolves to `ethernet`.
- `Test_LOOPBACK_STATE_IS_UP`: the loopback interface, whose operational state is `unknown`, reports `up` when it is administratively up.
- `Test_LOOPBACK_STATE_WITHOUT_ADMIN_FLAG_IS_DOWN`: the same interface reports `down` when it is not.
- `Test_STATE_KEEPS_PLACEHOLDER_WHEN_OPERSTATE_UNREADABLE`: an interface whose operational state cannot be read keeps its previous placeholder.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/src/data_provider/src/network/networkBSDWrapper.h
+++ b/src/data_provider/src/network/networkBSDWrapper.h
@@ -32,6 +32,7 @@
 static const std::map<std::pair<int, int>, std::string> NETWORK_INTERFACE_TYPE =
 {
     { std::make_pair(IFT_ETHER, IFT_ETHER),                     "ethernet"       },
+    { std::make_pair(IFT_LOOP, IFT_LOOP),                       "loopback"       },
     { std::make_pair(IFT_ISO88023, IFT_ISO88023),               "CSMA/CD"        },
     { std::make_pair(IFT_ISO88024, IFT_ISO88025),               "token ring"     },
     { std::make_pair(IFT_FDDI, IFT_FDDI),                       "FDDI"           },
--- a/src/data_provider/src/network/networkLinuxWrapper.h
+++ b/src/data_provider/src/network/networkLinuxWrapper.h
@@ -13,6 +13,7 @@
 #define _NETWORK_LINUX_WRAPPER_H
 
 #include <ifaddrs.h>
+#include <net/if.h>
 #include <net/if_arp.h>
 #include <sys/socket.h>
 #include "inetworkWrapper.h"
@@ -107,6 +108,7 @@ static const std::map<std::pair<int, int>, std::string> NETWORK_INTERFACE_TYPE =
     { std::make_pair(ARPHRD_PPP, ARPHRD_PPP),                   "point-to-point"    },
     { std::make_pair(ARPHRD_ATM, ARPHRD_ATM),                   "ATM"               },
     { std::make_pair(ARPHRD_IEEE1394, ARPHRD_IEEE1394),         "firewire"          },
+    { std::make_pair(ARPHRD_LOOPBACK, ARPHRD_LOOPBACK),         "loopback"          },
     { std::make_pair(ARPHRD_TUNNEL, ARPHRD_IRDA),               "tunnel"            },
     { std::make_pair(ARPHRD_FCPP, ARPHRD_FCFABRIC),             "fibrechannel"      },
     { std::make_pair(ARPHRD_IEEE802_TR, ARPHRD_IEEE802154_PHY), "wireless"          },
@@ -536,6 +538,14 @@ class NetworkLinuxInterface final : public INetworkInterfaceWrapper
             if (!operationalState.empty())
             {
                 state = Utils::splitIndex(operationalState, '\n', 0);
+
+                // Drivers that do not implement carrier detection, loopback among them, report
+                // "unknown" instead of an operational state; resolve those from the administrative
+                // flag. Every remaining value means the interface is not carrying traffic.
+                if ("up" != state && "down" != state)
+                {
+                    state = ("unknown" == state && (m_interfaceAddress->ifa_flags & IFF_UP)) ? "up" : "down";
+                }
             }
 
             return state;
--- a/src/data_provider/src/network/networkWindowsWrapper.h
+++ b/src/data_provider/src/network/networkWindowsWrapper.h
@@ -25,6 +25,7 @@
 static const std::map<int, std::string> NETWORK_INTERFACE_TYPES =
 {
     { IF_TYPE_ETHERNET_CSMACD, "ethernet"       },
+    { IF_TYPE_SOFTWARE_LOOPBACK, "loopback"       },
     { IF_TYPE_ISO88025_TOKENRING, "token ring"     },
     { IF_TYPE_PPP, "point-to-point" },
     { IF_TYPE_ATM, "ATM"            },
--- a/src/data_provider/tests/sysInfoNetworkLinux/CMakeLists.txt
+++ b/src/data_provider/tests/sysInfoNetworkLinux/CMakeLists.txt
@@ -16,6 +16,8 @@ add_executable(sysInfoNetworkLinux_unit_test ${sysinfo_UNIT_TEST_SRC}
 
 target_link_libraries(
   sysInfoNetworkLinux_unit_test
+  FilesystemWrapper
+  FileIO
   debug
   gtestd
   debug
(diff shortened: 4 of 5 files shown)


cborla (dismissed):
LGTM
- [x] Code changes reviewed

cborla (approved):
LGTM
- [x] Code changes reviewed

PR #39036, opened 2026-09-08
Report why an SCA check could not be assessed

### LINKED ISSUE #38926: Half the CIS benchmark reports "Not applicable" when the evidence is missing, so failed controls are recorded as exempt, and nothing says why
88 of the 183 checks in the shipped CIS Amazon Linux 2023 policy come out `Not applicable` on Amazon Linux 2023, the platform the policy is named for. The reason in every case is that the file or command the rule reads is absent, and `check.reason` is null in all 732 rows, so nothing distinguishes a control that does not apply from one that was never evaluated.

|---|---|---|---|---|
| 5.0.0-latest | The SCA engine's result classification, `wazuh-states-sca`, and the shipped agent image | Agent and Indexer | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 (agents), two-node cluster |

**Severity:** High. In CIS terminology "not applicable" is an exemption: the control does not apply to this system and is excluded from scoring. Applying that label to a control whose evidence could not be read converts a failure into an exemption. An auditor reading the console sees 50 passed and 45 failed, and no indication that 88 controls were never assessed.

## The four SELinux controls are the clearest case

```
   31038  Not applicable  Ensure SELinux is not disabled in bootloader configuration.  reason=None
   31039  Not applicable  Ensure SELinux policy is configured.                         reason=None
   31040  Not applicable  Ensure the SELinux mode is not disabled.                      reason=None
   31041  Not applicable  Ensure the SELinux mode is enforcing.                         reason=None
```

Their rules read `/etc/selinux/config` and run `getenforce`. On the shipped agent image:

```
   /etc/selinux/config    ABSENT
   getenforce             ABSENT
```

A host with no SELinux configuration and no `getenforce` is a host where SELinux is not enforcing. That is exactly what controls 31040 and 31041 exist to detect, and it is a **fail**. Reporting it as not applicable states the opposite: that the control does not apply here.

The same reasoning covers `31033 Ensure bootloader password is set` (no `/boot/grub2/grub.cfg`), and `31035` / `31036`, the core dump controls (no `/etc/systemd/coredump.conf`).

## The split

Consistent across all four agents, with zero checks disagreeing between them:

```
   Not applicable   88 of 183   (48%)
   Passed           50
   Failed           45
```

Of the 88, 37 need one of five binaries the shipped agent image does not contain, and the other 51 need a file or a different command that is also absent:

```
   modprobe    ABSENT      lsmod       ABSENT
   findmnt     ABSENT      systemctl   ABSENT
   auditctl    ABSENT
   rpm         present     grep        present
   awk         present     sysctl      present
```

`31000` and `31001` are the kernel filesystem-module controls, and they need `modprobe` and `lsmod`. A container shares the host's kernel, so whether the host can mount squashfs is a real property that these controls are about, and it went unmeasured.

## `check.reason` exists and is never filled

```
   Failed           180 rows, with reason 0
   Not applicable   352 rows, with reason 0
   Passed           200 rows, with reason 0
```

The field is in the mapping, next to `result`, `condition` and `rules`. It is null in every one of the 732 documents. So an operator who wants to know why 88 controls were skipped has nowhere to look: not the result, not the reason, not the console.

## The shipped agent image cannot run its own shipped policy

These two things are packaged together by the same project: `public.ecr.aws/wazuh-cicd/wazuh/wazuh-agent:5.0.0-latest` contains `/var/ossec/ruleset/sca/cis_amazon_linux_2023.yml`, and it does not contain five of the binaries that file's rules invoke. The out-of-the-box result of the product's own container running the product's own benchmark is that 48% of it cannot be evaluated, reported as though it did not need to be.

## Expected behavior

A check whose evidence could not be read is reported distinctly from a check that does not apply, for instance `Not run` with `check.reason` naming the missing file or command, and it is excluded from the pass rate rather than counted as satisfied. A control like "SELinux is enforcing", whose absent evidence *is* the failure condition, returns `Failed`. Separately, the agent image contains the binaries the policy it ships needs, or ships a policy scoped to what it can evaluate.

## Steps to reproduce

1. Run the shipped agent image and let one SCA scan complete.
2. `POST /wazuh-states-sca/_search?size=0` with `{"aggs":{"r":{"terms":{"field":"check.result"}}}}`. `Not applicable` 352, `Passed` 200, `Failed` 180, which is 88, 50 and 45 of the 183 distinct checks.
3. Query `check.id: "31040"`. `result` is `Not applicable`, `reason` is `null`.
4. `docker run --rm --entrypoint /bin/sh public.ecr.aws/wazuh-cicd/wazuh/wazuh-agent:5.0.0-latest -c 'command -v modprobe lsmod findmnt systemctl getenforce; ls /etc/selinux/config'`. None of them are there.
5. `_count` any query on `wazuh-states-sca` with `{"exists":{"field":"check.reason"}}`. Zero.

## Related

[091](already-reported/BUG_091-a-failed-cis-check-never-alerts-because-sca-produces-state-and-the-rules-read-events.md) is that a failed check never alerts, [138](BUG_138-all-eight-remediation-scripts-in-the-shipped-cis-policy-are-a-single-comment-line-so-running-one-does-nothing-and-exits-zero.md) is that the fix it offers does nothing, and this is that half the benchmark was never run and says so as though it were exempt. The three together are the whole SCA workflow. [092](BUG_092-the-shipped-agent-container-monitors-no-system-log-so-every-log-based-rule-is-unreachable.md) is the same packaging shape: the shipped container cannot exercise the content shipped alongside it. [093](BUG_093-the-health-check-reports-green-while-detection-is-broken.md) is the same reporting shape: a green result standing in for an unmeasured one.

cc @wazuh/devel-xdrsiem-agent

anromerom:
## Description

Related to #38926

Every SCA check reported as `Not applicable` carried an empty `check.reason`, so an operator could not tell a control that does not apply from one whose evidence could not be read. The reason was collected by the rule evaluators and then discarded before it reached the result, so no scan ever published one. This PR makes the reason reach the agent database and the index, and keeps it tied to the check it actually describes.

Classification is unchanged: an unresolvable rule still yields `Not applicable`, inherited from the 4.x engine. Reporting those checks as distinct from an exemption needs a new result value agreed with the console and the API, and the shipped agent images invoking tools they do not contain is an image concern, so the issue stays open for both.

## Proposed Changes

- Evaluate each rule before reading its unresolved reason, in the policy requirements loop and the checks loop, and return the reason from `FindContentInFile`, whose `Invalid` results had none because it is a free function.
- Clear a rule's unresolved reason at the start of every evaluation, and when a directory scan skips a file it could not evaluate, so an earlier file or scan does not explain a later result.
- Store the reason on any result rather than only `Not applicable`, writing the column unconditionally so it clears once a check becomes measurable.
- Explain a `Not applicable` check that reaches the result with no evaluable rules, which happens when every rule of a check failed to parse.
- Keep one copy of a reason repeated by several rules of the same check.
- Sanitise the stored reason against the RFC 3629 ranges, so overlong encodings, UTF-16 surrogate halves and code points above U+10FFFF are replaced rather than passed to `json::dump()`.
- Cap the stored reason at 1024 bytes, matching `ignore_above` on `check.reason`, cutting at a line boundary and never inside a UTF-8 sequence.
- Drop an empty reason from both stateful builders and from the stateless `previous` object.

### Results and Evidence

<details>
<summary>Agent database and indexer coverage, before and after</summary>

Ubuntu 24.04 agent, `cis_ubuntu24-04`, 279 checks. Result and reason coverage in the agent database:

```sql
sqlite3 /var/ossec/queue/sca/db/sca.db \
  "select result, count(*), sum(reason is not null and reason != '') as with_reason from sca_check group by result;"
```

```sql
-- before                          -- after
Failed|85|0                        Failed|85|0
Not applicable|120|0               Not applicable|120|120
Passed|74|0                        Passed|74|0
```

A check that previously reported no reason:

```sql
sqlite3 /var/ossec/queue/sca/db/sca.db "select id, result, reason from sca_check where id = '35500';"
```

```sql
35500|Not applicable|Command execution failed: modprobe -n -v cramfs
Command execution failed: lsmod
Path '/etc/modprobe.d/' does not exist
```

Same agent in the indexer, per result:

```sql
GET wazuh-states-sca/_search?size=0
{"aggs":{"r":{"terms":{"field":"check.result"},"aggs":{"with_reason":{"filter":{"exists":{"field":"check.reason"}}}}}}}
```

```sql
Not applicable  docs=120  with_reason=120
Failed          docs=85   with_reason=0
Passed          docs=74   with_reason=0
```

Before the change the same query returned 0 documents with `check.reason` over the whole index.

</details>

### Artifacts Affected

- `libsca.so` (agent SCA module), all platforms. The defect manifests on GCC-built agents.
- `wazuh-states-sca` documents now populate the existing `check.reason`. No mapping change.
- `sca_check.reason` in the agent database holds an empty string instead of NULL when there is nothing to explain, so `reason IS NULL` no longer matches those rows.

### Configuration Changes

N/A

### Documentation Updates

N/A

### Tests Introduced

- `sca_policy_test.cpp` (new): the reason reaches the reported result, one line per unresolved rule in rule order, no reason on a resolved check, and a `Not run` requirement propagating its reason. Three of the four fail without the fix.
- `directory_rule_evaluator_test.cpp`: a scan that skips a file it could not evaluate and then matches a later one reports no reason, and a reason from a previous evaluation of the same rule does not survive. Both fail without the fix.
- `check_condition_evaluator_test.cpp`: repeated reasons kept once, distinct ones kept in order, and a check with no evaluable rules still explaining itself.
- `sca_event_handler_test.cpp`: the reason is stored on a result other than `Not applicable`, capped at 1024 bytes, sanitised when it quotes an undecodable path, and cleared but not published when empty.
- `sca_utils_test.cpp`: reason sanitising over valid multibyte text, invalid path bytes, a truncated sequence, the cut at a line boundary and an oversized single line. Overlong encodings, surrogate halves and code points above U+10FFFF are replaced, the smallest and largest sequence of every valid range is kept, and each case asserts `json::dump()` does not throw. Four of the five added cases fail without the fix.
- `file_rule_evaluator_test.cpp` and `sca_recovery_utils_test.cpp`: the file-content rule reports a reason for an unusable pattern; the snapshot builder drops an empty reason and keeps a populated one.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/src/wazuh_modules/sca/sca_impl/include/sca_recovery_utils.hpp
+++ b/src/wazuh_modules/sca/sca_impl/include/sca_recovery_utils.hpp
@@ -15,6 +15,7 @@
 #include <idbsync.hpp>
 #include <json.hpp>
 #include <sca_field_decoder.hpp>
+#include <sca_utils.hpp>
 #include <string>
 
 #include "stringHelper.h"
@@ -77,9 +78,20 @@ namespace sca
                 check.erase("policy_id");
             }
 
-            if (check.contains("reason") && (check["reason"].is_null() || check["reason"].get<std::string>().empty()))
+            if (check.contains("reason"))
             {
-                check.erase("reason");
+                const auto reason = check["reason"].is_string()
+                                    ? sca::SanitizeReason(check["reason"].get<std::string>(), sca::REASON_MAX_LENGTH)
+                                    : std::string {};
+
+                if (reason.empty())
+                {
+                    check.erase("reason");
+                }
+                else
+                {
+                    check["reason"] = reason;
+                }
             }
 
             // Remove internal field not part of indexer schema
--- a/src/wazuh_modules/sca/sca_impl/include/sca_recovery_utils.hpp
+++ b/src/wazuh_modules/sca/sca_impl/include/sca_recovery_utils.hpp
@@ -77,6 +77,11 @@ namespace sca
                 check.erase("policy_id");
             }
 
+            if (check.contains("reason") && (check["reason"].is_null() || check["reason"].get<std::string>().empty()))
+            {
+                check.erase("reason");
+            }
+
             // Remove internal field not part of indexer schema
             if (check.contains("regex_type"))
             {
--- a/src/wazuh_modules/sca/sca_impl/include/sca_utils.hpp
+++ b/src/wazuh_modules/sca/sca_impl/include/sca_utils.hpp
@@ -86,6 +86,9 @@ namespace sca
     /// @return The string representation of the regex type.
     std::string RegexEngineTypeToString(RegexEngineType engineType);
 
+    /// @brief Largest reason kept on a check, in bytes. Matches ignore_above on check.reason.
+    constexpr size_t REASON_MAX_LENGTH {1024};
+
     /// @brief Makes a check's unresolved reason safe to store and publish.
     /// @param reason The raw reason, one line per unresolved rule.
     /// @param maxLength Largest result in bytes.
--- a/src/wazuh_modules/sca/sca_impl/include/sca_utils.hpp
+++ b/src/wazuh_modules/sca/sca_impl/include/sca_utils.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cstddef>
 #include <optional>
 #include <string>
 #include <vector>
@@ -84,4 +85,10 @@ namespace sca
     /// @param engineType The RegexEngineType enum value.
     /// @return The string representation of the regex type.
     std::string RegexEngineTypeToString(RegexEngineType engineType);
+
+    /// @brief Makes a check's unresolved reason safe to store and publish.
+    /// @param reason The raw reason, one line per unresolved rule.
+    /// @param maxLength Largest result in bytes.
+    /// @return A valid UTF-8 reason of at most maxLength bytes.
+    std::string SanitizeReason(const std::string& reason, size_t maxLength);
 } // namespace sca
--- a/src/wazuh_modules/sca/sca_impl/src/check_condition_evaluator.cpp
+++ b/src/wazuh_modules/sca/sca_impl/src/check_condition_evaluator.cpp
@@ -39,7 +39,7 @@ void CheckConditionEvaluator::AddResult(const RuleEvaluationResult& result)
         m_hasInvalid = m_hasInvalid || (result.result == RuleResult::Invalid);
         m_hasNotRun = m_hasNotRun || (result.result == RuleResult::NotRun);
 
-        if (!result.reason.empty())
+        if (!result.reason.empty() && !HasReason(result.reason))
         {
             if (!m_unresolvedReason.empty())
             {
@@ -118,5 +118,41 @@ sca::CheckResult CheckConditionEvaluator::Result() const
 
 std::string CheckConditionEvaluator::GetUnresolvedReason() const
 {
+    if (!m_unresolvedReason.empty())
+    {
+        return m_unresolvedReason;
+    }
+
+    if (m_totalRules == 0)
+    {
+        return "None of the check's rules could be evaluated on this system";
+    }
+
+    if (m_hasInvalid || m_hasNotRun)
+    {
+        return "Unknown reason";
+    }
+
     return m_unresolvedReason;
+}
+
+bool CheckConditionEvaluator::HasReason(const std::string& reason) const
+{
+    size_t position = m_unresolvedReason.find(reason);
+
+    while (position != std::string::npos)
+    {
+        const bool startsLine = position == 0 || m_unresolvedReason[position - 1] == '\n';
+        const auto end = position + reason.size();
+        const bool endsLine = end == m_unresolvedReason.size() || m_unresolvedReason[end] == '\n';
+
+        if (startsLine && endsLine)
+        {
+            return true;
+        }
+
+        position = m_unresolvedReason.find(reason, position + 1);
+    }
+
+    return false;
 }
--- a/src/wazuh_modules/sca/sca_impl/src/check_condition_evaluator.hpp
+++ b/src/wazuh_modules/sca/sca_impl/src/check_condition_evaluator.hpp
@@ -26,6 +26,9 @@ class CheckConditionEvaluator
         std::string GetUnresolvedReason() const;
 
     private:
+        /// @brief Whether the accumulated reason already holds this exact line.
+        bool HasReason(const std::string& reason) const;
+
         ConditionType m_type;
         int m_totalRules {0};
         int m_passedRules {0};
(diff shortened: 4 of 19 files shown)


anromerom:
Hi @AwwalQuan!

As this issue relates to thread intel team, could you please have a look at the proposed changes and confirm if they are what is expected on your side? I left comments on the issue with evidence of the check results after the patch:

- https://github.com/wazuh/wazuh/issues/38926#issuecomment-5591933797

AwwalQuan (commented):
The implementation solves the main problem of showing the reason. But there may be a potential gap to investigate further.

A check that **passes** can still surface a **reason** describing a different, unrelated file. A stale state left over from earlier in the same directory scan.

## The bug

`DirRuleEvaluator::CheckDirectoryForContents()` scans a directory rule (e.g. "every `*.conf` file under `X` must have `Y >= 100`") file by file, stopping at the first match. Per-file failures are non-fatal because the loop is explicitly written to skip a bad file and keep checking the rest.
Each attempt writes into `m_lastUnresolvedReason`, a single member variable reused across the whole loop, and nothing clears it between iterations.

So: file 1 fails to parse → `m_lastUnresolvedReason = "Invalid pattern for file 1"` → loop continues.
File 2 matches → `Evaluate()` returns `Found`. The stale message from file 1 is still sitting in `m_lastUnresolvedReason`, and GetUnresolvedReason()` returns it regardless of the final result.

Before this PR that didn't matter because `ReportCheckResult` only attached `reason` when the check's result was `"Not applicable"`, so a leftover reason on a `Passed` result was silently dropped. 
This PR removes that gate (intentionally, to fix the actual bug — `reason` needs to reach `Not applicable` results too), which also makes this dormant leftover-state issue visible: a passed check can now ship a reason that has nothing to do with why it passed.

## Possible fixes 
Either or both approaches should solve it:
1. Clear `m_lastUnresolvedReason` before/after each per-file attempt in the loop, not just once at the top of `Evaluate()`.
2. In `CheckConditionEvaluator::AddResult`, only keep a reason when `result.result` is itself `Invalid`/`NotRun` — not whenever the reason string happens to be non-empty.

AwwalQuan (dismissed):
LGTM

cborla:
Finding: `synchronizeDatabaseSnapshot()` (sca_impl.cpp, full-sync/recovery path) reads `reason` straight from `sca_check` and serializes it via `buildStatefulMessage(...).dump()` with no sanitization equivalent to `SanitizeReason`. If a row still holds invalid UTF-8 in `reason` (e.g. left over from before this fix), `dump()` throws, and since the per-check loop has no try/catch of its own — only the one wrapping the whole function — that aborts the entire snapshot/recovery sync, not just the offending row.

Possible fix: run `SanitizeReason` (or an equivalent guard) on `check["reason"]` inside `normalizeCheckForStateful`/`buildStatefulMessage` too, and/or wrap the per-check body of that loop in its own try/catch so one bad row doesn't drop the whole batch.

cborla (approved):
LGTM
- [x] Code changes reviewed

PR #39037, opened 2026-09-08
Report unresolved port process id as null instead of 0 on Linux

### LINKED ISSUE #38923: The ports inventory records forty-one open ports and names the process for none of them
A port table exists to answer one question: what is listening on this port. `wazuh-states-inventory-ports` holds 41 rows across six agents and answers it for none of them. `process.name` is `null` and `process.pid` is `0` in all 41.

|---|---|---|---|---|
| 5.0.0-latest | The `ports` collector of `wodle name="syscollector"`, `wazuh-states-inventory-ports`, and the `wazuh-it-hygiene` network events | Agent and Indexer | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 (agents), two-node cluster |

**Severity:** High. An unexpected listening port is a standard detection, and the only useful next question is which binary holds it. The table shows the port and stops. The same nulls also reach the event stream, so 2,055 network connection events are equally anonymous.

## All 41 rows, one distinct value

```
   pares (process.name, process.pid) distinct: {(None, 0): 41}

   states: listening 6, time_wait 29, without state 6
```

The six `listening` rows are exactly the ones where the question matters, and they are as empty as the rest.

A row, unedited:

```json
{
  "process": {"name": null, "pid": 0},
  "file": {"inode": "161774"},
  "destination": {"port": 0, "ip": ["0.0.0.0"]},
  "host": {"network": {"ingress": {"queue": 0}, "egress": {"queue": 0}}},
  "checksum": {"hash": {"sha1": "5dd8d21b5d4d96cbb4da164c9b60e9810f25894e"}},
  "wazuh": {"agent": {"name": "wazuh-agent-agent-ctl9", "id": "006"}}
}
```

The socket inode is there. The inode is precisely the key that maps a socket to its owning process through `/proc/*/fd`, which is what `ss -p` and `lsof -i` do, so the one piece of data needed to fill the field is already in the document.

## The same module names processes correctly in the table next to it

```
   wazuh-states-inventory-processes   45 rows, with name of process: 45
   ejemplos: tini, entrypoint.sh, tail, wazuh-execd, wazuh-syscheckd,
             wazuh-logcollec, wazuh-modulesd
```

Same six agents, same module, same scan window (`2026-08-26T14:36Z` to `2026-08-28T20:33Z`). The process list is complete and correct. So the collector can see processes, it can see sockets, and it does not join them.

## It reaches the event stream too

The 2,055 `connection-established` and `connection-closed` events from `wazuh-it-hygiene` carry the same pair:

```json
"event": {"action": "connection-closed"},
"process": {"name": null, "pid": 0},
"source": {"ip": ["172.20.0.6"], "port": 55642},
"destination": {"ip": ["172.20.0.3"], "port": 1517},
"network": {"transport": "tcp"}
```

So a rule that wanted to alert on "process X opened an outbound connection" has nothing to select on, and neither does an analyst reading the event.

Two further fields are zero in all 41 rows, `host.network.ingress.queue` and `host.network.egress.queue`, which are the receive and send queue depths `/proc/net/tcp` reports in the same line the inode came from.

## Expected behavior

`process.name` and `process.pid` name the process that owns the socket, resolved from the inode the document already carries, as `ss -p` does. Where the owner genuinely cannot be resolved, for instance a socket in `time_wait` with no owning process, the field is absent rather than `null` and `0`, so an absent owner is distinguishable from an unresolved one. `pid: 0` is a real pid on Linux and using it as a placeholder makes the column unsafe to read.

## Steps to reproduce

1. Enrol a Linux agent, let the inventory scan run, and confirm it has listening sockets (`ss -ltnp` inside the agent shows the process for each).
2. `POST /wazuh-states-inventory-ports/_search` with `{"size":50,"query":{"match_all":{}}}`.
3. Every document has `"process": {"name": null, "pid": 0}`, including the rows with `interface.state: "listening"`.
4. `POST /wazuh-states-inventory-processes/_search` for the same agents. Every row there has a process name.

## Related

[134](BUG_134-three-inventory-collectors-ship-enabled-and-produce-nothing-and-one-of-them-is-the-package-list-vulnerability-detection-needs.md) and [135](BUG_135-the-users-inventory-marks-every-local-system-account-as-remote-and-leaves-six-account-hygiene-columns-empty-in-every-row.md) are the other two inventory findings. [125](BUG_125-four-thousand-events-declare-themselves-alerts-and-the-same-ecs-field-has-two-different-shapes-depending-on-the-decoder.md) records `null` used as a value rather than omitting the field. [133](BUG_133-only-one-integration-fills-the-ecs-pivot-fields-and-related-hosts-names-the-collecting-agent-instead-of-the-host-in-the-event.md) records that these same 2,055 events also keep their addresses out of `related.ip`.

cc @wazuh/devel-xdrsiem-agent

rjcausarano:
## Description

On Linux, when the ports collector cannot join a socket to its owning process (e.g. the socket lives in a different PID namespace, as happens for containerized agents), the unresolved `process_pid` was emitted as `0` instead of being omitted. Since `0` can also be a real pid, an unresolved owner was indistinguishable from a legitimately resolved one in the `wazuh-states-inventory-ports` index — `process_name` already fell back to `null` in this case via its own sentinel, but `process_pid` did not have an equivalent.

This was raised as a nit in #38923 after the original "41 open ports, no process names" report from that issue turned out not to be reproducible outside of Docker containers (root cause: sockets in the container's network namespace have no resolvable owner from inside its PID namespace).

Closes #38923

## Proposed Changes

- `src/data_provider/src/ports/portLinuxWrapper.h`: `LinuxPortWrapper::pid()` now returns `-1` (instead of the default-constructed `0`) when a port's owning process cannot be resolved, so the unresolved case has an unambiguous sentinel.
- `src/wazuh_modules/syscollector/src/syscollectorImp.cpp` (`ecsPortData`): `process_pid` is now mapped to `null` when the raw value is `-1`, mirroring how `process_name` already collapses its own "unknown" sentinel to `null`. Legitimate `pid` values, including `0` (e.g. Windows' "System Idle Process"), pass through unchanged.

### Results and Evidence

Before: an unresolved port owner surfaced as `"process_pid": 0, "process_name": null` — a `0` pid look-alike for a real process.

After: an unresolved port owner surfaces as `"process_pid": null, "process_name": null` for both fields consistently, while a resolved port (including a legitimately-zero pid) keeps its real `process_pid` value.

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - N/A — no decoders/rules touched by this change
  - N/A — change is in the syscollector data provider, not the engine
- Wazuh server API/Framework
  - N/A — no server API/framework changes

Verified locally by rebuilding and running the two affected unit test suites on Linux:
- `sysInfoPort_unit_test`: 1/1 passed
- `syscollectorimp_unit_test`: 98/98 passed, including `sanitizeJsonValues`, which exercises `ecsPortData` with a `process_pid: 0` fixture and confirms the real-zero-pid case still passes through unchanged

Windows/macOS compilation and the memory/sanitizer checks are still pending.

### Artifacts Affected

- `wazuh-agentd` / syscollector's Linux ports collector (`src/data_provider`)
- `wazuh-modulesd` syscollector module (`src/wazuh_modules/syscollector`)

### Configuration Changes

N/A — no configuration parameters added or changed.

### Tests Introduced

No new unit tests were added; the existing `sysInfoPort_unit_test` and `syscollectorimp_unit_test` suites (specifically `SyscollectorImpTest.sanitizeJsonValues`) already cover this code path and were re-verified against this change.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

### DIFF
--- a/src/data_provider/src/ports/portLinuxWrapper.h
+++ b/src/data_provider/src/ports/portLinuxWrapper.h
@@ -269,7 +269,10 @@ class LinuxPortWrapper final : public IPortWrapper
 
         int32_t pid() const override
         {
-            return {};
+            // -1 marks a socket whose owning process could not be resolved (e.g. it lives
+            // outside the collector's PID namespace); 0 is a valid-looking pid and would be
+            // indistinguishable from a real one.
+            return -1;
         }
 };
 
--- a/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
+++ b/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
@@ -1135,6 +1135,7 @@ nlohmann::json Syscollector::ecsPortData(const nlohmann::json& originalData, boo
             ret[pointer] = nullptr;
         }
     }
+
     setJsonFieldArray(ret, originalData, "/source/ip", "source_ip", createFields);
     setJsonField(ret, originalData, "/source/port", "source_port", createFields);
 
--- a/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
+++ b/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
@@ -1118,7 +1118,23 @@ nlohmann::json Syscollector::ecsPortData(const nlohmann::json& originalData, boo
     setJsonField(ret, originalData, "/interface/state", "interface_state", createFields);
     setJsonField(ret, originalData, "/network/transport", "network_transport", createFields);
     setJsonField(ret, originalData, "/process/name", "process_name", createFields);
-    setJsonField(ret, originalData, "/process/pid", "process_pid", createFields);
+
+    // process_pid: -1 marks an unresolved owner (see portLinuxWrapper.h); emit null instead of
+    // a value that would look like a real pid.
+    if (createFields || originalData.contains("process_pid"))
+    {
+        const nlohmann::json::json_pointer pointer("/process/pid");
+
+        if (originalData.contains("process_pid") && originalData["process_pid"].is_number() &&
+                originalData["process_pid"].get<int32_t>() != -1)
+        {
+            ret[pointer] = originalData["process_pid"];
+        }
+        else
+        {
+            ret[pointer] = nullptr;
+        }
+    }
     setJsonFieldArray(ret, originalData, "/source/ip", "source_ip", createFields);
     setJsonField(ret, originalData, "/source/port", "source_port", createFields);
 


cborla (approved):
LGTM!
- [x] Code changes reviewed

Issue #38928, opened 2026-09-04
Two CIS checks are byte-identical and a third is named for a control it does not test, so four consecutive ids cover two controls


Rebits:
Checks 31169 to 31172 of the shipped CIS Amazon Linux 2023 policy are four consecutive ids that display as two pairs of identically named rows. One pair is the same check twice, and the other pair is two different tests sharing one name, of which only the first matches that name.

|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0-latest | `/var/ossec/ruleset/sca/cis_amazon_linux_2023.yml`, checks 31169, 31170, 31171, 31172 | Agent | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 (agents), two-node cluster |

**Severity:** Medium. A compliance report that lists a control twice inflates the pass count for that control, and a control listed under another control's name means the report claims to cover something it never tested. Both survive review because the console shows two rows with the same title and an operator reads them as one control checked on two paths.

## 31171 and 31172 are the same check

```yaml
  - id: 31171
    name: "Ensure permissions on /etc/passwd are configured."
    condition: all
    rules:
      - 'c:stat -Lc "%n %a %u %U %g %G" /etc/passwd -> r:0 root 0 root && r:644|640|604|600|400|500'

  - id: 31172
    name: "Ensure permissions on /etc/passwd are configured."
    condition: all
    rules:
      - 'c:stat -Lc "%n %a %u %U %g %G" /etc/passwd -> r:0 root 0 root && r:644|640|604|600|400|500'
```

Identical name, identical condition, identical rule. Sweeping all 183 checks for shared rule sets returns exactly this one pair, so it is not a convention. Both reported `Passed` on all four agents, so `/etc/passwd` contributes two passes to the score and 183 ids cover 182 distinct controls.

The neighbours show what the id was for. The block runs `/etc/passwd` (31171), `/etc/passwd-` (31173), `/etc/group` (31174), `/etc/group-` (31175), `/etc/shadow` (31176), `/etc/shadow-` (31177), `/etc/gshadow` (31178), `/etc/gshadow-` (31179), which is CIS's own order with one slot spent on a repeat.

## 31170 tests whether root has a password, under the name of the GID check

```yaml
  - id: 31169
    name: "Ensure default group for the root account is GID 0."
    rules:
      - 'f:/etc/passwd -> !r:^\s*\t*# && r:root:[\w@-]+:[\w@-]+:0:'

[...]

PR #39075, opened 2026-09-09
fix(sca): restore line structure to CIS AL2023 remediation scripts

### LINKED ISSUE #38925: All eight remediation scripts in the shipped CIS policy are a single comment line, so running one does nothing and exits zero
The CIS Amazon Linux 2023 policy that ships with the agent gives eight of its 183 checks a bash script to fix the finding. Each script is stored as one line beginning with `#!/usr/bin/env bash`, with the script's own `#` comments still in it. On a single line, everything after the `#!` is part of that comment, so the whole file is one comment. An operator who copies it out of the console and runs it gets no output and exit status 0.

|---|---|---|---|---|
| 5.0.0-latest | `/var/ossec/ruleset/sca/cis_amazon_linux_2023.yml`, checks 31000, 31001, 31133, 31134, 31161, 31162, 31163, 31164 | Agent and Indexer | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 (agents), two-node cluster |

**Severity:** High. The failure is silent and it inverts the outcome: the operator believes they hardened the host, the exit status agrees, and the next scan reports the check still failing, which reads as the check being wrong rather than the fix having done nothing. Four of the eight are the password policy checks (`31161` creation requirements, `31162` lockout after failed attempts, `31163` password reuse, `31164` SHA-512 hashing), which are among the most commonly remediated findings in the whole benchmark.

## The eight

```
   31000  Ensure mounting of squashfs filesystems is disabled.
   31001  Ensure mounting of udf filesystems is disabled.
   31133  Ensure cron is restricted to authorized users.
   31134  Ensure at is restricted to authorized users.
   31161  Ensure password creation requirements are configured.
   31162  Ensure lockout for failed password attempts is configured.
   31163  Ensure password reuse is limited.
   31164  Ensure password hashing algorithm is SHA-512.
```

All 183 remediations in the policy contain zero newline characters. For the 175 that are prose instructions that is harmless. For these eight it is not.

## Why it is inert

`wc -l -c` on check 31000's script, extracted exactly as stored:

```
   0 lines, 1060 characters
```

Zero lines means zero newlines. The file is:

```
#!/usr/bin/env bash { l_mname="squashfs" # set module name # Check if the module exists on the system if [ -z "$(modprobe -n -v "$l_mname" ...
```

`bash -n` accepts it, which is the trap: it is syntactically valid, because a file consisting of one comment is valid. A script of exactly this shape:

```
$ printf '#!/usr/bin/env bash { echo HOLA } # comment' > demo.sh
$ bash demo.sh
$ echo $?
0
```

produces no output and succeeds. The real CIS script would have printed ` - setting module: squashfs to be not loadable`.

All eight pass `bash -n` and all eight are a single `#!`-prefixed line.

## The source ships it that way

This is not the indexing path collapsing whitespace. `/var/ossec/ruleset/sca/cis_amazon_linux_2023.yml` in the unmodified agent image, at check 31000:

```yaml
    remediation: "Run the following script to disable squashfs: #!/usr/bin/env bash { l_mname=\"squashfs\" # set module name # Check if the module exists ... fi}."
```

It is a double-quoted YAML scalar with no `\n` escapes, so the newlines were never there. A block scalar (`remediation: |`) would have preserved them.

## Check 31000 has a second, independent corruption

Even restored to multiple lines, one of the two writes in 31000 is broken. The first is correct:

```bash
echo -e "install $l_mname /bin/false" >> /etc/modprobe.d/"$l_mname".conf
```

The second has lost its opening quote and gained a space:

```bash
echo -e "blacklist $l_mname" >> /etc/modprobe.d/ $l_mname".conf
```

The redirect target becomes `/etc/modprobe.d/`, which is a directory, so the write fails with `Is a directory`, and the unbalanced quote swallows the rest of the statement. 31000 is the only one of the 183 checks whose remediation has an odd number of double quotes, and 31001, which is the same template with `udf` substituted, is clean, so this is a one-off edit rather than a template fault.

## Expected behavior

Multi-line remediation scripts are stored as YAML block scalars so their line structure survives, or, failing that, the comments are stripped and the statements separated by `;` so a single line still executes. A remediation the product offers the operator runs and reports what it did.

## Steps to reproduce

1. `docker run --rm --entrypoint /bin/sh public.ecr.aws/wazuh-cicd/wazuh/wazuh-agent:5.0.0-latest -c 'grep -c "usr/bin/env bash" /var/ossec/ruleset/sca/cis_amazon_linux_2023.yml'`. Eight.
2. `grep -A 6 "id: 31000"` the same file. `remediation:` is one double-quoted line containing `# set module name`.
3. Let an SCA scan run and read `wazuh-states-sca`. `check.remediation` for these eight contains no newline.
4. Write one of them to a file and run it. No output, exit 0, nothing changed.
5. `grep -o 'echo -e .blacklist.\{0,60\}'` on check 31000's script shows `>> /etc/modprobe.d/ $l_mname".conf`.

## Related

[091](already-reported/BUG_091-a-failed-cis-check-never-alerts-because-sca-produces-state-and-the-rules-read-events.md) records that a failed check never alerts. This is the next step in the same workflow: even when an operator finds the failed check by hand, the fix the product offers them does nothing. Check 31000's `mitre.tactic.name: "Stealth"` in the same YAML block is [123](BUG_123-sixteen-rules-file-a-technique-under-a-tactic-attack-does-not-place-it-in-and-two-tactics-carry-names-attack-does-not-use.md).

cc @wazuh/devel-xdrsiem-agent

MAnDumu:
## Description

The CIS Amazon Linux 2023 policy that ships with the agent (`ruleset/sca/amazon/cis_amazon_linux_2023.yml`) gives 8 of its 183 checks a bash script to fix the finding: checks `31000`, `31001`, `31133`, `31134`, `31161`, `31162`, `31163`, `31164`. Each script was stored as a single-line YAML scalar beginning with `#!/usr/bin/env bash`, with the script's own `#` comments still embedded in it. On a single line, everything after the first `#` is part of a bash comment, so the whole script was one giant comment. An operator who copied it out of the console and ran it got no output and exit status 0 — the remediation silently did nothing while the product reported success.

Closes #38925

## Proposed Changes

- Converted the `remediation:` field for the 8 affected checks from a single-line YAML flow scalar to a YAML literal block scalar (`remediation: |-`), reconstructing real line breaks and indentation for the embedded bash script.
- While reconstructing, found and fixed 15 additional, individually-verified corruption points in the same 8 fields — all in the same family (a single space or backslash accidentally inserted or dropped at what looks like an old line-wrap boundary), each confirmed to change real `grep -P`/`sed`/`bash` behavior before/after the fix:
  - `31000`: the redirect `>> /etc/modprobe.d/ $l_mname".conf` (missing opening quote, misplaced space) — the second, independent bug already called out in the issue for this check. Also 3x `\b` (PCRE word-boundary escape) that was single-backslash-escaped in the original double-quoted YAML, which a standards-compliant YAML parser decodes as an actual backspace control byte instead of literal `\b`. Also a missing separator between `fi` and the closing `}` (`fi}`, invalid bash).
  - `31001`: 3x stray backslash before a comment (`\# Remediate ...` broke the statement), the same `\b`→backspace issue as 31000, and 1x `\h +Module` (stray space breaking the `\h+` PCRE quantifier).
  - `31161`: `enforce_for_r oot` → `enforce_for_root` (word split by a stray space).
  - `31162`: 2x `$ /` and 1x `$ '` (stray space before a sed/grep delimiter, making the pattern permanently unmatchable — a silent no-op).
  - `31163`: 4 occurrences of a space injected inside a PCRE character class (`[^#\n\r]` corrupted to `[^#\n\ r]` / `[^# \n\r]` / `[^#\n\r] +...`), plus 2x `\h +)?remember` (space breaking a quantifier).
  - `31164`: 2x `)? \h+` (space breaking a quantifier) and 1x `s ha512` → `sha512` in a sed replacement string (was writing the literal broken value into the target config file).
- Leading prose (before the shebang) and any trailing human-readable text (after the script) are now prefixed with `# ` on their own line, so the complete field — not just the script portion — is valid, safe bash from the first byte to the last, in case an operator copies the whole field rather than just the script block.
- No other check in the file was touched; only these 8 `remediation:` fields changed (154 insertions / 8 deletions, one file).

Note: the affected `remediation:` field content is byte-identical between `main`/`5.0.0`/`5.0.1`/`4.14.9`/`4.10.6` — this same fix was opened against `main` first (#39073) and is being re-targeted to `5.0.0` per team convention; the `main` PR will be closed in favor of this one.

### Results and Evidence

Before (check 31000, extracted and run exactly as shipped):
```
$ bash extracted_remediation.sh
$ echo $?
0
```
No output, nothing changed — the script is one comment.

After, sandboxed test of the specific bug called out in the issue (the `modprobe.d` redirect in check 31000):
```
=== BUGGY version (as currently shipped) ===
/bin/bash: line 25: etc/modprobe.d/: Is a directory

=== FIXED version ===
total 12
-rw-rw-r-- 1 ubuntu ubuntu   19 Sep  9 12:12 squashfs.conf
blacklist squashfs
```

All 8 reconstructed scripts, extracted directly from the final YAML (via PyYAML, no manual truncation of either end), pass `bash -n` with no syntax errors, end to end. The full 183-check YAML file still parses correctly (`yaml.safe_load`) after the edit, and a token-level diff against the original (whitespace/newlines aside) confirms only the listed, verified corrections were made — no other content or meaning was altered.

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X

N/A — this PR only changes a shipped YAML data file (no compiled code), so the compilation/memory-test checklist below does not apply.

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

N/A

N/A

  - [ ] Test run in parallel

N/A

- Wazuh server API/Framework

N/A

Manual verification actually performed (not part of the template's checklist, documented here for the reviewer): `yaml.safe_load` on the full file before/after, `bash -n` on all 8 reconstructed scripts (both isolated and extracted straight from the committed file, untruncated), and sandboxed `grep -P`/`sed`/`bash` reproductions of each of the 15 additional corruptions listed above, comparing buggy vs. fixed behavior against realistic config-file content.

### Artifacts Affected

- `ruleset/sca/amazon/cis_amazon_linux_2023.yml` — default CIS Amazon Linux 2023 SCA policy shipped with the agent.

### Configuration Changes

N/A — no new configuration parameters, no changes to default values. Only the textual content of an existing field (`remediation`) on 8 existing checks is corrected; the check `id`s, `name`s, `rules`, and `compliance` mappings are unchanged.

### Tests Introduced

N/A — no dedicated schema/lint tooling exists in this repo for `ruleset/sca/**` (confirmed: no validator, no CI job lints this directory). The existing `tests/integration/test_sca/test_basic/test_validate_remediation.py` is a live-agent integration test that uses its own small fixture policies, not this shipped file, so it is unaffected by and does not cover this change.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

Note for reviewers: this PR only fixes issue #38925 (single-line remediation scripts). The related issue #38928 (byte-identical/mislabeled checks in the same file) is a separate, follow-up fix — not included here because it requires cross-referencing the official CIS benchmark document (not available in-repo) to determine the correct control for the affected check ids.

### DIFF
--- a/ruleset/sca/amazon/cis_amazon_linux_2023.yml
+++ b/ruleset/sca/amazon/cis_amazon_linux_2023.yml
@@ -39,7 +39,34 @@ checks:
     description: "The squashfs filesystem type is a compressed read-only Linux filesystem embedded in small footprint systems. A squashfs image can be used without having to first decompress the image."
     rationale: "Removing support for unneeded filesystem types reduces the local attack surface of the system. If this filesystem type is not needed, disable it."
     impact: 'As Snap packages utilizes squashfs as a compressed filesystem, disabling squashfs will cause Snap packages to fail. Snap application packages of software are self-contained and work across a range of Linux distributions. This is unlike traditional Linux package management approaches, like APT or RPM, which require specifically adapted packages per Linux distribution on an application update and delay therefore application deployment from developers to their software''s end-user. Snaps themselves have no dependency on any external store ("App store"), can be obtained from any source and can be therefore used for upstream software deployment.'
-    remediation: "Run the following script to disable squashfs: #!/usr/bin/env bash { l_mname=\"squashfs\" # set module name # Check if the module exists on the system if [ -z \"$(modprobe -n -v \"$l_mname\" 2>&1 | grep -Pi -- \"\\h*modprobe:\\h+FATAL:\\h+Module\\h+$l_mname\\h+not\\h+found\\h+in\\h+directory\")\" ]; then # Remediate loadable l_loadable=\"$(modprobe -n -v \"$l_mname\")\" [ \"$(wc -l <<< \"$l_loadable\")\" -gt \"1\" ] && l_loadable=\"$(grep -P -- \"(^\\h*install|\b$l_mname)\b\" <<< \"$l_loadable\")\" if ! grep -Pq -- '^\\h*install \/bin\/(true|false)' <<< \"$l_loadable\"; then echo -e \" - setting module: \"$l_mname\" to be not loadable\" echo -e \"install $l_mname \/bin\/false\" >> /etc/modprobe.d/\"$l_mname\".conf fi # Remediate loaded if lsmod | grep \"$l_mname\" > /dev/null 2>&1; then echo -e \" - unloading module \"$l_mname\"\" modprobe -r \"$l_mname\" fi # Remediate deny list if ! modprobe --showconfig | grep -Pq -- \"^\\h*blacklist\\h+$l_mname\b\"; then echo -e \" - deny listing \"$l_mname\"\" echo -e \"blacklist $l_mname\" >> /etc/modprobe.d/ $l_mname\".conf fi else echo -e \" - Nothing to remediate\\n - Module \"$l_mname\" doesn't exist on the system\" fi}."
+    remediation: |-
+      # Run the following script to disable squashfs:
+      #!/usr/bin/env bash
+      {
+          l_mname="squashfs" # set module name
+          # Check if the module exists on the system
+          if [ -z "$(modprobe -n -v "$l_mname" 2>&1 | grep -Pi -- "\h*modprobe:\h+FATAL:\h+Module\h+$l_mname\h+not\h+found\h+in\h+directory")" ]; then
+              # Remediate loadable
+              l_loadable="$(modprobe -n -v "$l_mname")"
+              [ "$(wc -l <<< "$l_loadable")" -gt "1" ] && l_loadable="$(grep -P -- "(^\h*install|\b$l_mname)\b" <<< "$l_loadable")"
+              if ! grep -Pq -- '^\h*install /bin/(true|false)' <<< "$l_loadable"; then
+                  echo -e " - setting module: "$l_mname" to be not loadable"
+                  echo -e "install $l_mname /bin/false" >> /etc/modprobe.d/"$l_mname".conf
+              fi
+              # Remediate loaded
+              if lsmod | grep "$l_mname" > /dev/null 2>&1; then
+                  echo -e " - unloading module "$l_mname""
+                  modprobe -r "$l_mname"
+              fi
+              # Remediate deny list
+              if ! modprobe --showconfig | grep -Pq -- "^\h*blacklist\h+$l_mname\b"; then
+                  echo -e " - deny listing "$l_mname""
+                  echo -e "blacklist $l_mname" >> /etc/modprobe.d/"$l_mname".conf
+              fi
+          else
+              echo -e " - Nothing to remediate\n - Module "$l_mname" doesn't exist on the system"
+          fi
+      }
     compliance:
       cmmc: ["CM.L2-3.4.7", "CM.L2-3.4.8", "SC.L2-3.13.6"]
       fedramp: ["CM-2", "CM-3", "CM-6", "CM-7"]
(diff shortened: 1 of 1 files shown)


MAnDumu:
Consolidated back into #39073 (reused that PR's branch/number, retargeted to `5.0.0`) instead of keeping this as a separate PR. Closing this one.

PR #39105, opened 2026-09-10
Remove the schema from API validation errors, bound offset, and 404 unknown cluster nodes

### LINKED ISSUE #38938: API validation errors return the internal JSON schema as a Python dict, and `offset` is unbounded so the overflow reaches the database
A rejected query parameter comes back with the validating library's raw exception text, which embeds the internal schema as a Python dict literal. And while `limit` is bounded at both ends, `offset` has a minimum and no maximum, so a value beyond `int32` passes validation and fails inside `wazuhdb` instead.

|---|---|---|---|---|
| 5.0.0-latest, API revision `beta5` | wazuh-manager API request validation | Manager | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 agent, single-node stack |

**Severity:** Low to Medium. Neither half is exploitable on its own. The schema disclosure is an information leak and an API-contract problem; the unbounded `offset` is a validation gap that turns a client mistake into a database error.

## The schema comes back in the error

```
GET /agents?limit=-1
   400  {"title": "Bad Request", "detail": "-1 is less than the minimum of 1. Failed validating
        'minimum' in schema: {'default': 500, 'format': 'int32', 'maximum': 100000, ...
```

```
GET /agents?limit=abc
   400  {"title": "Bad Request", "detail": "'abc' is not of type 'integer'. Failed validating
        'type' in schema: {'default': 500, 'format': 'int32', 'maximum': ...
```

`Failed validating 'minimum' in schema: {...}` is `jsonschema`'s own exception message, passed to the client unchanged. Three things follow.

- The response carries a Python dict literal, single-quoted, inside a JSON string. It is not JSON and cannot be parsed by a client that wants to react to the constraint programmatically.
- The internal schema is disclosed, including fields the client was not told about, such as `default: 500`.
- The message is a library implementation detail. A `jsonschema` upgrade changes the wording, so any client matching on it breaks.

## The error echoes attacker-controlled bytes back

```
GET /groups?groups_list=%00
   400  {"title": "Bad Request", "detail": "'\\x00' is not a 'group_names'. Failed validating
        'format' in schema['items']: {'description': 'Group name', 'form...
```

The NUL byte is reflected as `\x00` in the response. It is escaped here, so this is a note rather than an injection, but the pattern of quoting the submitted value verbatim into an error message is the one to avoid in a message that is also written to the access log.

## `offset` is validated at the bottom and not at the top

The two paging parameters are treated differently:

```
GET /agents?limit=-1                       400  "-1 is less than the minimum of 1"
GET /agents?limit=999999999                400  "999999999 is greater than the maximum of 100000"
GET /agents?offset=-5                      400  "-5 is less than the minimum of 0"
GET /agents?offset=99999999999999999999    400  "Error in wazuhdb request"
```

`limit` has `minimum: 1` and `maximum: 100000` and both fire. `offset` has `minimum: 0` and no maximum, so a twenty-digit value passes the schema, reaches `wazuhdb`, and produces:

```
{"title": "Bad Request", "detail": "Error in wazuhdb request",
 "remediation": "Make sure the request is correct",
 "dapi_errors": {"wazuh.manager": {"error": ...
```

The status code is still 400, so this is not a crash. But the request travelled the whole way to the database layer to be rejected, the error names a component the client knows nothing about, and the remediation text, "Make sure the request is correct", says nothing about which parameter was wrong. Both parameters are declared `format: int32`; only one is enforced against it.

## Nonexistent named resources are handled two different ways

Noted from the same sweep, on 46 read routes:

```
GET /groups/x/agents          404  {"title": "Resource Not Found", "detail": "The group does not exist"}
GET /cluster/x/status         200  {"data": {"affected_items": [], "total_affected_items": 0, ...
GET /cluster/x/logs           200  {"data": {"affected_items": [], "total_affected_items": 0, ...
```

A group that does not exist is a 404 with a clear message. A cluster node that does not exist is a 200 with an empty result. A client cannot tell "this node has nothing to report" from "there is no such node".

## Expected behavior

Validation errors state the violated constraint in the API's own vocabulary, without the validating library's exception text or the schema literal. `offset` carries the `maximum` its `int32` format implies, so an out-of-range value is rejected at the edge. A request for a named cluster node that does not exist is a 404, as it is for a group.

## Steps to reproduce

1. `GET /agents?limit=-1` with a valid token. Read `detail`: it contains `Failed validating 'minimum' in schema: {'default': 500, ...`.
2. `GET /agents?offset=99999999999999999999`. `detail` is `Error in wazuhdb request`.
3. `GET /agents?limit=999999999`. `detail` names the maximum, so the same class of check exists for `limit`.
4. `GET /cluster/nonexistent/status` returns 200 with an empty `affected_items`; `GET /groups/nonexistent/agents` returns 404.

## Related

[214](already-reported/BUG_214-a-four-byte-body-crashes-every-write-endpoint-of-the-manager-api-including-the-unauthenticated-login.md) and [215](already-reported/BUG_215-the-api-writes-a-nested-password-to-its-access-log-in-cleartext-even-when-it-rejects-the-request.md) are the middleware defects found in the same sweep. [095](already-reported/BUG_095-a-read-only-user-reaches-an-unhandled-nullpointerexception-in-the-detector-api.md) is the previous validation-ordering defect in this API.

cc @wazuh/devel-xdrsiem-server

ignaciogalle12git:
## Description

A rejected query parameter came back with the raw `jsonschema` exception text: the submitted value, then the internal schema as a Python dict literal. `offset` had no `maximum` despite its `int32` format, so an out-of-range value passed validation and was only refused by `wazuh-manager-db`. A request for a cluster node that does not exist answered 200 with the node as a failed item, while a nonexistent group answers 404.

Closes #38938

## Proposed Changes

- `api/api/parameter_validator.py`, registered through `validator_map` on `add_api()`. Connexion's parameter validator returns `str(ValidationError)`, which is where both the value echo and the schema dump come from. `WazuhParameterValidator` builds the message from the exception object (`validator`, `validator_value`) instead, so neither can reach the response by construction, and it names the rejected parameter. Every keyword the specification uses on a parameter has its own wording; one without it is named without its schema value. Only the `parameter` entry is replaced: the body and response validators already use `exception.message`.
- `api/api/spec/spec.yaml`: `maximum: 2147483647` on the shared `offset` parameter (20 references). `int32` is the bound the parameter already declares and the one `limit` is enforced against.
- `framework/wazuh/core/cluster/dapi/dapi.py`: `forward_request` raises the `WazuhResourceNotFound(1730)` that `get_nodes_info` already reports, when a request scoped to a single `node_id` matches no node. `node_list` requests are untouched. The 404 is declared on the eight `/cluster/{node_id}/*` operations.
- `docs/ref/modules/server-api/api-reference.md`: the `offset` bound and the meaning of 404.

The 404 sits in DAPI, not in the controllers, so RBAC still runs before the existence check, as it does for groups: a `node:id:*` policy lets the unknown name reach `get_nodes_info`, while an explicit `node:id:<name>` policy is denied first with 4000. A controller-side check would let any caller probe node names.

That check compares against `get_connected_nodes()`, so a configured node that is disconnected at that moment is reported the same way. The master cannot enumerate configured node names, so this is stated as contract in the `node_id` description, in the seven paths and in the reference.

### Results and Evidence

Captured through two connexion apps on the real `spec.yaml` with the production error handler, one with connexion's parameter validator and one with this branch's:

```
# GET /agents?limit=-1
before  "-1 is less than the minimum of 1. Failed validating 'minimum' in schema: {'default': 500, 'format': 'int32', 'maximum': 100000, 'minimum': 1, 'type': 'integer'}. On instance: -1"
after   "Invalid value for query parameter 'limit': must be greater than or equal to 1"

# GET /groups?groups_list=%00
before  "'\x00' is not a 'group_names'. Failed validating 'format' in schema['items']: {'description': 'Group name', 'format': 'group_names', 'type': 'string'}. On instance[0]: '\x00'"
after   "Invalid value for query parameter 'groups_list': must match the 'group_names' format"

# GET /agents?offset=99999999999999999999
before  passed validation and reached the controller
after   "Invalid value for query parameter 'offset': must be less than or equal to 2147483647"
```

Unit suites, as `5_testunit_api.yml` and `5_testunit_framework.yml` run them: `pytest api/api --ignore=test` 663 passed, `pytest framework --ignore=tests` 2162 passed and 6 skipped. Every commit was also run on its own, so none of them leaves the branch red, and each new test was run against the base with its file reverted to confirm it fails without the fix.

N/A for the compilation and memory checks: the change is Python only.

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

The stages this branch touches cannot be reproduced locally, so the box stays unchecked until the CI pass comes back green.

### Artifacts Affected

`wazuh-manager-apid`: the `detail` of every rejected parameter, the `offset` contract on the 20 endpoints that reference it, and the status code of `/cluster/{node_id}/*` for a node the cluster does not have.

### Configuration Changes

N/A

### Tests Introduced

- `api/api/test/test_parameter_validator.py`: the wording of every keyword, the value never appearing in the message, and the fallback.
- `api/api/test/test_spec_query_param_bounds.py`: `offset` rejected above `int32` and accepted at `int32`, and a rejection disclosing neither the schema nor the value, through the real specification and the production handler.
- `framework/wazuh/core/cluster/dapi/tests/test_dapi.py`: a single `node_id` matching no node fails as 1730, a `node_list` request does not.
- `api/test/integration/test_cluster_endpoints.tavern.yaml`: the status of a node the cluster does not have.
- `api/test/integration/test_agent_GET_endpoints.tavern.yaml`: the eight large-offset stages pin the rejection, which no longer depends on the submitted value.

**Three of those stages change from 200 to 400**, so the contract change is worth a look: `GET /groups` with `offset=99999999999` (a stage explicitly named "(valid)"), and `/groups/{group_id}/configuration` and `/groups/{group_id}/files` with `offset=9999999999999999999`, all answered 200 with an empty page because those endpoints read files instead of paging through wazuh-db. The response to an absurd offset used to depend on which backend served the endpoint; the bound makes all of them a 400 at the edge.

## Review Checklist

- [ ] Code changes reviewed


api/api/error_handler.py
```diff
@@ -25,7 +28,8 @@ def _cleanup_detail_field(detail: str) -> str:
     str
         New value for the detail field.
     """
-    return ' '.join(str(detail).replace("\n\n", ". ").replace("\n", "").split())
+    detail = str(detail).split('\nFailed validating ', 1)[0]
+    return ' '.join(detail.replace("\n\n", ". ").replace("\n", "").split())
```
jotacarma90: **[S2 · blocking · introduced]** Cutting at `Failed validating` removes the schema dump, but the line it keeps is jsonschema's own message, which is `f"{instance!r} …"` for `type`, `maxLength`, `pattern`, `format` and `enum` — so the submitted value is still reflected and, for the keywords whose message does not name the bound, the violated constraint is now not stated at all. Run against the pinned `jsonschema==4.20.0`: `GET /agents?search=<1025 'a'>` returns a 1039-character `detail` containing the whole value and no mention of `1024`, and `GET /groups?groups_list=%00` returns `"'\x00' is not a 'group_names'"` — which is the echo half of #38938 and what CHANGELOG.md:97 and the new docstring both say is fixed. Connexion's own idiom for this is not string surgery: the body validator uses `exception.message` (`validators/json.py:75`), and `validators/parameter.py:58` is the only place that does `str(exception)`, so a `validator_map` override on `add_api()` (supported, `middleware/request_validation.py:103`; the app passes none today) lets you build the message from the exception object and never touch the instance.

```python
# api/api/parameter_validator.py
import copy

from connexion.utils import is_null, is_nullable
from connexion.validators import ParameterValidator
from jsonschema import Draft4Validator, ValidationError


class WazuhParameterValidator(ParameterValidator):
    """Report the violated constraint without jsonschema's text or the submitted value."""

    @staticmethod
    def validate_parameter(parameter_type, value, param, param_name=None):
        if is_nullable(param) and is_null(value):
            return
        if value is None:
            return (f"Missing {parameter_type} parameter '{param['name']}'"
                    if param.get('required') else None)
        schema = copy.deepcopy(param).get('schema', param)
        try:
            Draft4Validator(schema, format_checker=Draft4Validator.FORMAT_CHECKER).validate(value)
        except ValidationError as exc:
            name = param.get('name', param_name)
            return (f"Invalid value for {parameter_type} parameter '{name}': "
                    f"{exc.validator} = {exc.validator_value!r}")

# then, in the add_api() call of api/scripts/wazuh_manager_apid.py:
#     validator_map={'parameter': WazuhParameterValidator}
```

If that is out of scope for this PR, the claim has to come out of the changelog and the docstring, with a follow-up issue for the vocabulary rewrite.

CHANGELOG.md
```diff
@@ -94,6 +94,7 @@
...
+| [#38938](https://github.com/wazuh/wazuh/issues/38938) | Fixed API validation errors returning the `jsonschema` exception text, which embedded the internal schema as a Python dict literal and echoed the submitted value back. Bounded `offset` at 2147483647, so an out-of-range value is rejected by the specification instead of by `wazuh-manager-db`, and a request for a cluster node that does not exist now answers 404 instead of 200 with no items. |
```
jotacarma90: **[S3 · non-blocking · introduced]** Two statements in this row do not match the diff: the submitted value is still echoed back (see the `error_handler.py` comment), and the previous cluster-node response was not "200 with no items" — `forward_request` ends with `response.add_failed_items_from(allowed_nodes)` and the cleanup loop only drops `{'unknown'}` and denied-4000 items, so the caller got `total_failed_items: 1` with the node named and error 1730. Recover the echo mention if F1 is fixed.

```suggestion
| [#38938](https://github.com/wazuh/wazuh/issues/38938) | Fixed API validation errors returning the `jsonschema` exception text, which embedded the internal schema as a Python dict literal. Bounded `offset` at 2147483647, so an out-of-range value is rejected by the specification instead of by `wazuh-manager-db`, and a request for a cluster node the cluster does not have now answers 404 instead of 200 with the node reported as a failed item. |
```

framework/wazuh/core/cluster/dapi/dapi.py
```diff
@@ -555,6 +555,12 @@ async def clean_valid_nodes(nodes_to_clean: List[Tuple]) -> List[Tuple]:
             for node in nodes.items():
                 if node[0] in [node_name['name'] for node_name in allowed_nodes.affected_items] or node[0] == 'unknown':
                     valid_nodes.append(node)
+            if 'node_id' in self.f_kwargs and not valid_nodes:
+                # A node_id is a path parameter: a name the cluster does not have is a 404, as a
+                # nonexistent group is. get_nodes_info() classified it and denied unreadable nodes first.
+                not_found = next((e for e in allowed_nodes.failed_items if getattr(e, 'code', None) == 1730), None)
+                if not_found:
+                    raise not_found
```
jotacarma90: **[S2 · blocking · introduced]** This 404 also fires for a node that *is* in the cluster but is not connected at that instant, because the node universe it compares against is only the connected clients: `nodes` comes from `get_system_nodes()` → `control.get_nodes()` → `Server.get_connected_nodes()`, which returns the master plus `self.clients` (`server.py:529`), so a reconnecting `worker1` lands in `get_nodes_info`'s `non_existent_nodes` (`framework/wazuh/cluster.py:118-124`) and gets 1730. During a rolling restart `GET /cluster/worker1/status` therefore flips from 200 with `total_failed_items: 1` naming the node to a hard 404 "Node does not exist", and `PUT /cluster/worker1/configuration` reads as a wrong URL rather than "the target is momentarily down" — a case #38938 does not ask about. The master cannot enumerate configured node names, so the fix is not a tighter condition but an explicit contract: declare the 404 on the seven `/cluster/{node_id}/*` paths and say in the `node_id` description and the `docs/ref` status-code table that a node not currently in the cluster — disconnected included — is reported this way; if the 200 must be preserved for a node that is merely down, that needs a source of configured names and changes the scope of the issue.

```suggestion
            if 'node_id' in self.f_kwargs and not valid_nodes:
                # A node_id is a path parameter: a name the cluster does not have is a 404, as a
                # nonexistent group is. get_nodes_info() classified it and denied unreadable nodes
                # first. The cluster's node universe is get_connected_nodes(), so a configured node
                # that is not connected right now is reported the same way -- the spec documents it.
                not_found = next((e for e in allowed_nodes.failed_items if e.code == 1730), None)
                if not_found is not None:
                    raise not_found
```

framework/wazuh/core/cluster/dapi/dapi.py
```diff
@@ -555,6 +555,12 @@ async def clean_valid_nodes(nodes_to_clean: List[Tuple]) -> List[Tuple]:
...
             del self.f_kwargs['node_id' if 'node_id' in self.f_kwargs else 'node_list']
```
jotacarma90: **[S3 · non-blocking · introduced]** None of the seven paths that can now answer 404 declares it: `/cluster/{node_id}/status` (`spec.yaml:6035`), `/info` (6123), `/configuration` (6174, GET and PUT), `/daemons/stats` (6307), `/logs` (6480), `/logs/summary` (6546) and `/configuration/{component}/{configuration}` (6762) all list 200/400/401/403/405/429 only, while the group endpoints this change is modelled on declare `404: $ref '#/components/responses/ResourceNotFoundResponse'` (12 occurrences, e.g. `spec.yaml:3875`). It does not break anything at runtime — connexion resolves an undeclared status to an empty response definition (`operations/abstract.py:187-190`), so the IT environment's forced `validate_responses=True` still passes — but the spec is what the public reference and generated clients are built from.

```yaml
        '404':
          $ref: '#/components/responses/ResourceNotFoundResponse'
```

api/test/integration/test_agent_GET_endpoints.tavern.yaml
```diff
@@ -239,10 +239,7 @@ stages:
     response:
       status_code: 400
       json:
-        error: 2003
-        # wazuh-db's raw backend error text (e.g. SQL fragments) must never reach the caller;
-        # only the generic wazuhdb-request message is expected here.
-        detail: 'Error in wazuhdb request'
+        <<: *error_spec
```
jotacarma90: **[S3 · non-blocking · introduced]** This stage and the four other already-400 ones now assert only `{title: !anystr, detail: !anystr}`, and `api/test/integration/pytest.ini` sets `tavern-strict=json:off`, so the check is a subset match: the pre-fix body (`400`, `title: 'Bad Request'`, `detail: 'Error in wazuhdb request'`) satisfies it too. Revert `maximum: 2147483647` and the stages at 233, 1898, 3008, 3298 and 3572 stay green — only the three that went 200→400 actually detect the bound, and the point of the issue is precisely *which* layer rejects. The message is deterministic per `offset` value and an explicit key beats `<<` in YAML, which keeps the existing idiom of the NUL-byte stage that pins `detail`.

```suggestion
        <<: *error_spec
        detail: '9999999999999999999 is greater than the maximum of 2147483647'
```

api/api/test/test_error_handler.py
```diff
@@ -37,6 +37,12 @@ def test_cleanup_detail_field():
 
     assert _cleanup_detail_field(detail) == "Testing. Details field."
 
+    # str() of a jsonschema.ValidationError, as connexion hands it over for a rejected parameter.
+    detail = ("-1 is less than the minimum of 1\n\nFailed validating 'minimum' in schema:\n"
+              "    {'default': 500, 'maximum': 100000, 'minimum': 1}\n\nOn instance:\n    -1")
+
+    assert _cleanup_detail_field(detail) == "-1 is less than the minimum of 1"
+
```
jotacarma90: **[S3 · non-blocking · introduced]** Hand-writing jsonschema's `str()` layout means the test cannot notice if that layout changes, and the whole guarantee depends on it: the text is `dedent()` of an f-string whose literal lines carry 12 spaces, so a `message` containing a newline makes the common prefix empty, the marker becomes `'\n            Failed validating '`, the split matches nothing and the full subschema plus the `On instance` echo go back to the client — reproduced with the pinned 4.20.0. It is not reachable today (no keyword in `_keywords.py` emits a newline, `repr()` escapes them, and every checker in `api/api/validator.py:315-409` returns `bool`), so this is only about a future bump; building the fixture from the library makes such a bump fail here instead of leaking silently. Needs `from jsonschema import Draft4Validator` in the module imports.

```suggestion
    # str() of a real jsonschema.ValidationError, as connexion hands it over for a rejected
    # parameter (its ParameterValidator returns str(exception)). Built from the library so a change
    # in its message layout fails here instead of silently restoring the schema dump.
    error = next(Draft4Validator({'type': 'integer', 'default': 500, 'maximum': 100000,
                                  'minimum': 1}).iter_errors(-1))

    assert _cleanup_detail_field(str(error)) == "-1 is less than the minimum of 1"
```

api/api/spec/spec.yaml
```diff
@@ -3007,12 +3007,13 @@ components:
...
       in: query
       name: offset
-      description: "First element to return in the collection"
+      description: "First element to return in the collection. Must not exceed 2147483647 (2^31-1), the upper bound of its int32 format"
       schema:
         type: integer
         format: int32
         default: 0
         minimum: 0
+        maximum: 2147483647
```
jotacarma90: **[S3 · non-blocking · introduced]** This bound is now the contract, but the operator-facing table in `docs/ref/modules/server-api/api-reference.md:18` still reads `| offset | int | 0 | First element to return |` while the row below it documents `Max elements to return (max: 100,000)` for `limit`, and the status-code table at line 512 still glosses 404 as "Invalid endpoint" — which is not what the new cluster 404 means. `docs/ref/modules/<mod>/` is the operator contract a PR has to keep in sync, and the template's "Developer documentation reflects the changes" box is unchecked; this is two table rows.

```markdown
| `offset` | int | `0` | First element to return (max: 2,147,483,647) |
...
| 404       | Not Found         | Invalid endpoint, or the named resource (group, cluster node) does not exist |
```

framework/wazuh/core/cluster/dapi/tests/test_dapi.py
```diff
@@ -547,6 +547,30 @@ def test_DistributedAPI_forward_request_broadcasts_agent_targeted_requests():
...
+    with patch('wazuh.core.cluster.dapi.dapi.get_nodes_info', new=AsyncMock(return_value=nodes_info_result)), \
+            patch('wazuh.core.cluster.dapi.dapi.node_info', {'type': 'master', 'node': 'master'}):
+        dapi_kwargs = {'f': manager.status, 'logger': logger, 'request_type': 'distributed_master',
+                       'f_kwargs': {'node_id': 'nonexistent'}, 'nodes': ['master']}
+        raise_if_exc_routine(dapi_kwargs=dapi_kwargs, expected_error=1730)
+
+        dapi_kwargs = {'f': manager.status, 'logger': logger, 'request_type': 'distributed_master',
+                       'f_kwargs': {'node_list': ['nonexistent']}, 'nodes': ['master']}
+        raise_if_exc_routine(dapi_kwargs=dapi_kwargs)
+
```
jotacarma90: **[S3 · non-blocking · introduced]** This test patches `get_nodes_info` outright and hand-builds the 1730 failed item, so the premise the whole fix rests on — that `expose_resources`' wildcard expansion keeps an unknown name inside `filter_node` long enough to be classified 1730 — is asserted nowhere; I checked it by hand and it holds (`_single_processor` re-expands `node:id:<value>` when the user resource is `node:id:*`, `rbac/decorators.py:213-215`, and `_black_expansion` does the same in black mode), but a change there would revert the fix with the suite green. Meanwhile `test_cluster_endpoints.tavern.yaml` has stages for every `/cluster/{node_id}/*` endpoint with `master-node`, `worker1` and `worker2` and none with an unknown name, so the issue's repro step 4 never runs against a real cluster while the `offset` change got eight stages.

```yaml
  - name: Try to read the status of a node the cluster does not have
    request:
      verify: False
      url: "{protocol:s}://{host:s}:{balanced_port:d}/cluster/nonexistent/status"
      method: GET
      headers:
        Authorization: "Bearer {test_login_token}"
    response:
      status_code: 404
      json:
        error: 1730
```

api/api/error_handler.py
jotacarma90: **[S2 · blocking · introduced]** Cutting at `Failed validating` removes the schema dump, but the line it keeps is jsonschema's own message, which is `f"{instance!r} …"` for `type`, `maxLength`, `pattern`, `format` and `enum`, so the submitted value is still reflected and, for the keywords whose message does not name the bound, the violated constraint is now not stated at all. Run against the pinned `jsonschema==4.20.0`: `GET /agents?search=<1025 'a'>` returns a 1039-character `detail` containing the whole value and no mention of `1024`, and `GET /groups?groups_list=%00` returns `"'\x00' is not a 'group_names'"`, which is the echo half of #38938 and what CHANGELOG.md:97 and the new docstring both say is fixed. Connexion's own idiom for this is not string surgery: the body validator uses `exception.message` (`validators/json.py:75`), and `validators/parameter.py:58` is the only place that does `str(exception)`, so a `validator_map` override on `add_api()` (supported, `middleware/request_validation.py:103`; the app passes none today) lets you build the message from the exception object and never touch the instance.

```python
# api/api/parameter_validator.py
import copy

from connexion.utils import is_null, is_nullable
from connexion.validators import ParameterValidator
from jsonschema import Draft4Validator, ValidationError


class WazuhParameterValidator(ParameterValidator):
    """Report the violated constraint without jsonschema's text or the submitted value."""

    @staticmethod
    def validate_parameter(parameter_type, value, param, param_name=None):
        if is_nullable(param) and is_null(value):
            return
        if value is None:
            return (f"Missing {parameter_type} parameter '{param['name']}'"
                    if param.get('required') else None)
        param = copy.deepcopy(param)
        schema = param.get('schema', param)
        try:
            Draft4Validator(schema, format_checker=Draft4Validator.FORMAT_CHECKER).validate(value)
        except ValidationError as exc:
            name = param.get('name', param_name)
            return (f"Invalid value for {parameter_type} parameter '{name}': "
                    f"{exc.validator} = {exc.validator_value!r}")

# then, in the add_api() call of api/scripts/wazuh_manager_apid.py:
#     validator_map={'parameter': WazuhParameterValidator}
```

If that is out of scope for this PR, the claim has to come out of the changelog and the docstring, with a follow-up issue for the vocabulary rewrite.

CHANGELOG.md
```diff
@@ -95,7 +94,7 @@
...
-| [#39020](https://github.com/wazuh/wazuh/issues/39020) | Fixed the legacy `remote_upgrade` task delivery in `remoted` rejecting every task the Task Manager returned as `invalid payload`, which left every agent older than v5.0.0 without its WPK upgrade. The poller expected the payload as a JSON string while the Task Manager hands it back as the JSON object the producer stored; the poller now reads that object. A task deferred to the poller's retry list after a no-response is now aged from the moment it was deferred, so a task the Task Manager held for a long time still gets its full retry window. |
+| [#38938](https://github.com/wazuh/wazuh/issues/38938) | Fixed API validation errors returning the `jsonschema` exception text, which embedded the internal schema as a Python dict literal and echoed the submitted value back. Bounded `offset` at 2147483647, so an out-of-range value is rejected by the specification instead of by `wazuh-manager-db`, and a request for a cluster node that does not exist now answers 404 instead of 200 with no items. |
```
jotacarma90: **[S3 · non-blocking · introduced]** Two statements in this row do not match the diff: the submitted value is still echoed back (see the `error_handler.py` comment), and the previous cluster-node response was not "200 with no items": `forward_request` ends with `response.add_failed_items_from(allowed_nodes)` and the cleanup loop only drops `{'unknown'}` and denied-4000 items, so the caller got `total_failed_items: 1` with the node named and error 1730. Recover the echo mention if F1 is fixed.

```suggestion
| [#38938](https://github.com/wazuh/wazuh/issues/38938) | Fixed API validation errors returning the `jsonschema` exception text, which embedded the internal schema as a Python dict literal. Bounded `offset` at 2147483647, so an out-of-range value is rejected by the specification instead of by `wazuh-manager-db`, and a request for a cluster node the cluster does not have now answers 404 instead of 200 with the node reported as a failed item. |
```

framework/wazuh/core/cluster/dapi/dapi.py
jotacarma90: **[S2 · blocking · introduced]** This 404 also fires for a node that *is* in the cluster but is not connected at that instant, because the node universe it compares against is only the connected clients: `nodes` comes from `get_system_nodes()` → `control.get_nodes()` → `Server.get_connected_nodes()`, which returns the master plus `self.clients` (`server.py:529`), so a reconnecting `worker1` lands in `get_nodes_info`'s `non_existent_nodes` (`framework/wazuh/cluster.py:128-135`) and gets 1730. During a rolling restart `GET /cluster/worker1/status` therefore flips from 200 with `total_failed_items: 1` naming the node to a hard 404 "Node does not exist", and `PUT /cluster/worker1/configuration` reads as a wrong URL rather than "the target is momentarily down", a case #38938 does not ask about. The master cannot enumerate configured node names, so the fix is not a tighter condition but an explicit contract: declare the 404 on the seven `/cluster/{node_id}/*` paths and say in the `node_id` description and the `docs/ref` status-code table that a node not currently in the cluster (disconnected included) is reported this way; if the 200 must be preserved for a node that is merely down, that needs a source of configured names and changes the scope of the issue.

```suggestion
            if 'node_id' in self.f_kwargs and not valid_nodes:
                # A node_id is a path parameter: a name the cluster does not have is a 404, as a
                # nonexistent group is. get_nodes_info() classified it and denied unreadable nodes
                # first. The cluster's node universe is get_connected_nodes(), so a configured node
                # that is not connected right now is reported the same way -- the spec documents it.
                not_found = next((e for e in allowed_nodes.failed_items if e.code == 1730), None)
                if not_found is not None:
                    raise not_found
```

framework/wazuh/core/cluster/dapi/dapi.py
jotacarma90: **[S3 · non-blocking · introduced]** None of the seven paths that can now answer 404 declares it: `/cluster/{node_id}/status` (`spec.yaml:6035`), `/info` (6123), `/configuration` (6174, GET and PUT), `/daemons/stats` (6307), `/logs` (6480), `/logs/summary` (6546) and `/configuration/{component}/{configuration}` (6762) all list 200/400/401/403/405/429 only, while the group endpoints this change is modelled on declare `404: $ref '#/components/responses/ResourceNotFoundResponse'` (12 occurrences, e.g. `spec.yaml:3875`). It does not break anything at runtime, since connexion resolves an undeclared status to an empty response definition (`operations/abstract.py:187-190`), so the IT environment's forced `validate_responses=True` still passes, but the spec is what the public reference and generated clients are built from.

```yaml
        '404':
          $ref: '#/components/responses/ResourceNotFoundResponse'
```

api/test/integration/test_agent_GET_endpoints.tavern.yaml
jotacarma90: **[S3 · non-blocking · introduced]** This stage and the four other already-400 ones now assert only `{title: !anystr, detail: !anystr}`, and `api/test/integration/pytest.ini` sets `tavern-strict=json:off`, so the check is a subset match: the pre-fix body (`400`, `title: 'Bad Request'`, `detail: 'Error in wazuhdb request'`) satisfies it too. Revert `maximum: 2147483647` and the stages at 233, 1898, 3008, 3298 and 3572 stay green: only the three that went 200→400 actually detect the bound, and the point of the issue is precisely *which* layer rejects. The message is deterministic per `offset` value and an explicit key beats `<<` in YAML, which keeps the existing idiom of the NUL-byte stage that pins `detail`.

```suggestion
        <<: *error_spec
        detail: '9999999999999999999 is greater than the maximum of 2147483647'
```

api/api/test/test_error_handler.py
jotacarma90: **[S3 · non-blocking · introduced]** Hand-writing jsonschema's `str()` layout means the test cannot notice if that layout changes, and the whole guarantee depends on it: the text is `dedent()` of an f-string whose literal lines carry 12 spaces, so a `message` containing a newline makes the common prefix empty, the marker becomes `'\n            Failed validating '`, the split matches nothing and the full subschema plus the `On instance` echo go back to the client, reproduced with the pinned 4.20.0. It is not reachable today (no keyword in `_keywords.py` emits a newline, `repr()` escapes them, and every checker in `api/api/validator.py:315-409` returns `bool`), so this is only about a future bump; building the fixture from the library makes such a bump fail here instead of leaking silently. Needs `from jsonschema import Draft4Validator` in the module imports.

```suggestion
    # str() of a real jsonschema.ValidationError, as connexion hands it over for a rejected
    # parameter (its ParameterValidator returns str(exception)). Built from the library so a change
    # in its message layout fails here instead of silently restoring the schema dump.
    error = next(Draft4Validator({'type': 'integer', 'default': 500, 'maximum': 100000,
                                  'minimum': 1}).iter_errors(-1))

    assert _cleanup_detail_field(str(error)) == "-1 is less than the minimum of 1"
```

api/api/spec/spec.yaml
jotacarma90: **[S3 · non-blocking · introduced]** This bound is now the contract, but the operator-facing table in `docs/ref/modules/server-api/api-reference.md:18` still reads `| offset | int | 0 | First element to return |` while the row below it documents `Max elements to return (max: 100,000)` for `limit`, and the status-code table at line 520 still glosses 404 as "Invalid endpoint", which is not what the new cluster 404 means. `docs/ref/modules/<mod>/` is the operator contract a PR has to keep in sync, and the template's "Developer documentation reflects the changes" box is unchecked; this is two table rows.

```markdown
| `offset` | int | `0` | First element to return (max: 2,147,483,647) |
...
| 404       | Not Found         | Invalid endpoint, or the named resource (group, cluster node) does not exist |
```

framework/wazuh/core/cluster/dapi/tests/test_dapi.py
jotacarma90: **[S3 · non-blocking · introduced]** This test patches `get_nodes_info` outright and hand-builds the 1730 failed item, so the premise the whole fix rests on (that `expose_resources`' wildcard expansion keeps an unknown name inside `filter_node` long enough to be classified 1730) is asserted nowhere; I checked it by hand and it holds (`_single_processor` re-expands `node:id:<value>` when the user resource is `node:id:*`, `rbac/decorators.py:215-219`, and `_black_expansion` does the same in black mode), but a change there would revert the fix with the suite green. Meanwhile `test_cluster_endpoints.tavern.yaml` has stages for every `/cluster/{node_id}/*` endpoint with `master-node`, `worker1` and `worker2` and none with an unknown name, so the issue's repro step 4 never runs against a real cluster while the `offset` change got eight stages.

```yaml
  - name: Try to read the status of a node the cluster does not have
    request:
      verify: False
      url: "{protocol:s}://{host:s}:{balanced_port:d}/cluster/nonexistent/status"
      method: GET
      headers:
        Authorization: "Bearer {test_login_token}"
    response:
      status_code: 404
      json:
        error: 1730
```

api/api/error_handler.py
jotacarma90: **[S2 · blocking · introduced]** Cutting at `Failed validating` removes the schema dump, but the line it keeps is jsonschema's own message, which is `f"{instance!r} …"` for `type`, `maxLength`, `pattern`, `format` and `enum`, so the submitted value is still reflected and, for the keywords whose message does not name the bound, the violated constraint is now not stated at all. Run against the pinned `jsonschema==4.20.0`: `GET /agents?search=<1025 'a'>` returns a 1039-character `detail` containing the whole value and no mention of `1024`, and `GET /groups?groups_list=%00` returns `"'\x00' is not a 'group_names'"`, which is the echo half of #38938 and what CHANGELOG.md:99 and the new docstring both say is fixed. Connexion's own idiom for this is not string surgery: the body validator uses `exception.message` (`validators/json.py:75`), and `validators/parameter.py:58` is the only place that does `str(exception)`, so a `validator_map` override on `add_api()` (supported, `middleware/request_validation.py:103`; the app passes none today) lets you build the message from the exception object and never touch the instance.

```python
# api/api/parameter_validator.py
import copy

from connexion.utils import is_null, is_nullable
from connexion.validators import ParameterValidator
from jsonschema import Draft4Validator, ValidationError


class WazuhParameterValidator(ParameterValidator):
    """Report the violated constraint without jsonschema's text or the submitted value."""

    @staticmethod
    def validate_parameter(parameter_type, value, param, param_name=None):
        if is_nullable(param) and is_null(value):
            return
        if value is None:
            return (f"Missing {parameter_type} parameter '{param['name']}'"
                    if param.get('required') else None)
        param = copy.deepcopy(param)
        schema = param.get('schema', param)
        try:
            Draft4Validator(schema, format_checker=Draft4Validator.FORMAT_CHECKER).validate(value)
        except ValidationError as exc:
            name = param.get('name', param_name)
            return (f"Invalid value for {parameter_type} parameter '{name}': "
                    f"{exc.validator} = {exc.validator_value!r}")

# then, in the add_api() call of api/scripts/wazuh_manager_apid.py:
#     validator_map={'parameter': WazuhParameterValidator}
```

If that is out of scope for this PR, the claim has to come out of the changelog and the docstring, with a follow-up issue for the vocabulary rewrite.

CHANGELOG.md
```diff
@@ -96,6 +96,7 @@
...
 | [#39020](https://github.com/wazuh/wazuh/issues/39020) | Fixed the legacy `remote_upgrade` task delivery in `remoted` rejecting every task the Task Manager returned as `invalid payload`, which left every agent older than v5.0.0 without its WPK upgrade. The poller expected the payload as a JSON string while the Task Manager hands it back as the JSON object the producer stored; the poller now reads that object. A task deferred to the poller's retry list after a no-response is now aged from the moment it was deferred, so a task the Task Manager held for a long time still gets its full retry window. |
+| [#38938](https://github.com/wazuh/wazuh/issues/38938) | Fixed API validation errors returning the `jsonschema` exception text, which embedded the internal schema as a Python dict literal and echoed the submitted value back. Bounded `offset` at 2147483647, so an out-of-range value is rejected by the specification instead of by `wazuh-manager-db`, and a request for a cluster node that does not exist now answers 404 instead of 200 with no items. |
```
jotacarma90: **[S3 · non-blocking · introduced]** Two statements in this row do not match the diff: the submitted value is still echoed back (see the `error_handler.py` comment), and the previous cluster-node response was not "200 with no items": `forward_request` ends with `response.add_failed_items_from(allowed_nodes)` and the cleanup loop only drops `{'unknown'}` and denied-4000 items, so the caller got `total_failed_items: 1` with the node named and error 1730. Recover the echo mention if F1 is fixed.

```suggestion
| [#38938](https://github.com/wazuh/wazuh/issues/38938) | Fixed API validation errors returning the `jsonschema` exception text, which embedded the internal schema as a Python dict literal. Bounded `offset` at 2147483647, so an out-of-range value is rejected by the specification instead of by `wazuh-manager-db`, and a request for a cluster node the cluster does not have now answers 404 instead of 200 with the node reported as a failed item. |
```

framework/wazuh/core/cluster/dapi/dapi.py
jotacarma90, line 569: **[S2 · blocking · introduced]** This 404 also fires for a node that *is* in the cluster but is not connected at that instant, because the node universe it compares against is only the connected clients: `nodes` comes from `get_system_nodes()` → `control.get_nodes()` → `Server.get_connected_nodes()`, which returns the master plus `self.clients` (`server.py:529`), so a reconnecting `worker1` lands in `get_nodes_info`'s `non_existent_nodes` (`framework/wazuh/cluster.py:128-135`) and gets 1730. During a rolling restart `GET /cluster/worker1/status` therefore flips from 200 with `total_failed_items: 1` naming the node to a hard 404 "Node does not exist", and `PUT /cluster/worker1/configuration` reads as a wrong URL rather than "the target is momentarily down", a case #38938 does not ask about. The master cannot enumerate configured node names, so the fix is not a tighter condition but an explicit contract: declare the 404 on the seven `/cluster/{node_id}/*` paths and say in the `node_id` description and the `docs/ref` status-code table that a node not currently in the cluster (disconnected included) is reported this way; if the 200 must be preserved for a node that is merely down, that needs a source of configured names and changes the scope of the issue.

```suggestion
            if 'node_id' in self.f_kwargs and not valid_nodes:
                # A node_id is a path parameter: a name the cluster does not have is a 404, as a
                # nonexistent group is. get_nodes_info() classified it and denied unreadable nodes
                # first. The cluster's node universe is get_connected_nodes(), so a configured node
                # that is not connected right now is reported the same way -- the spec documents it.
                not_found = next((e for e in allowed_nodes.failed_items if e.code == 1730), None)
                if not_found is not None:
                    raise not_found
```

framework/wazuh/core/cluster/dapi/dapi.py
jotacarma90, line 570: **[S3 · non-blocking · introduced]** None of the seven paths that can now answer 404 declares it: `/cluster/{node_id}/status` (`spec.yaml:6035`), `/info` (6123), `/configuration` (6174, GET and PUT), `/daemons/stats` (6307), `/logs` (6480), `/logs/summary` (6546) and `/configuration/{component}/{configuration}` (6762) all list 200/400/401/403/405/429 only, while the group endpoints this change is modelled on declare `404: $ref '#/components/responses/ResourceNotFoundResponse'` (12 occurrences, e.g. `spec.yaml:3875`). It does not break anything at runtime, since connexion resolves an undeclared status to an empty response definition (`operations/abstract.py:187-190`), so the IT environment's forced `validate_responses=True` still passes, but the spec is what the public reference and generated clients are built from.

```yaml
        '404':
          $ref: '#/components/responses/ResourceNotFoundResponse'
```

api/test/integration/test_agent_GET_endpoints.tavern.yaml
jotacarma90, line 242: **[S3 · non-blocking · introduced]** This stage and the four other already-400 ones now assert only `{title: !anystr, detail: !anystr}`, and `api/test/integration/pytest.ini` sets `tavern-strict=json:off`, so the check is a subset match: the pre-fix body (`400`, `title: 'Bad Request'`, `detail: 'Error in wazuhdb request'`) satisfies it too. Revert `maximum: 2147483647` and the stages at 233, 1898, 3008, 3298 and 3572 stay green: only the three that went 200→400 actually detect the bound, and the point of the issue is precisely *which* layer rejects. The message is deterministic per `offset` value and an explicit key beats `<<` in YAML, which keeps the existing idiom of the NUL-byte stage that pins `detail`.

```suggestion
        <<: *error_spec
        detail: '9999999999999999999 is greater than the maximum of 2147483647'
```

api/api/test/test_error_handler.py
jotacarma90: **[S3 · non-blocking · introduced]** Hand-writing jsonschema's `str()` layout means the test cannot notice if that layout changes, and the whole guarantee depends on it: the text is `dedent()` of an f-string whose literal lines carry 12 spaces, so a `message` containing a newline makes the common prefix empty, the marker becomes `'\n            Failed validating '`, the split matches nothing and the full subschema plus the `On instance` echo go back to the client, reproduced with the pinned 4.20.0. It is not reachable today (no keyword in `_keywords.py` emits a newline, `repr()` escapes them, and every checker in `api/api/validator.py:315-409` returns `bool`), so this is only about a future bump; building the fixture from the library makes such a bump fail here instead of leaking silently. Needs `from jsonschema import Draft4Validator` in the module imports.

```suggestion
    # str() of a real jsonschema.ValidationError, as connexion hands it over for a rejected
    # parameter (its ParameterValidator returns str(exception)). Built from the library so a change
    # in its message layout fails here instead of silently restoring the schema dump.
    error = next(Draft4Validator({'type': 'integer', 'default': 500, 'maximum': 100000,
                                  'minimum': 1}).iter_errors(-1))

    assert _cleanup_detail_field(str(error)) == "-1 is less than the minimum of 1"
```

api/api/spec/spec.yaml
jotacarma90, line 3016: **[S3 · non-blocking · introduced]** This bound is now the contract, but the operator-facing table in `docs/ref/modules/server-api/api-reference.md:18` still reads `| offset | int | 0 | First element to return |` while the row below it documents `Max elements to return (max: 100,000)` for `limit`, and the status-code table at line 520 still glosses 404 as "Invalid endpoint", which is not what the new cluster 404 means. `docs/ref/modules/<mod>/` is the operator contract a PR has to keep in sync, and the template's "Developer documentation reflects the changes" box is unchecked; this is two table rows.

```markdown
| `offset` | int | `0` | First element to return (max: 2,147,483,647) |
...
| 404       | Not Found         | Invalid endpoint, or the named resource (group, cluster node) does not exist |
```

framework/wazuh/core/cluster/dapi/tests/test_dapi.py
jotacarma90, line 572: **[S3 · non-blocking · introduced]** This test patches `get_nodes_info` outright and hand-builds the 1730 failed item, so the premise the whole fix rests on (that `expose_resources`' wildcard expansion keeps an unknown name inside `filter_node` long enough to be classified 1730) is asserted nowhere; I checked it by hand and it holds (`_single_processor` re-expands `node:id:<value>` when the user resource is `node:id:*`, `rbac/decorators.py:215-219`, and `_black_expansion` does the same in black mode), but a change there would revert the fix with the suite green. Meanwhile `test_cluster_endpoints.tavern.yaml` has stages for every `/cluster/{node_id}/*` endpoint with `master-node`, `worker1` and `worker2` and none with an unknown name, so the issue's repro step 4 never runs against a real cluster while the `offset` change got eight stages.

```yaml
  - name: Try to read the status of a node the cluster does not have
    request:
      verify: False
      url: "{protocol:s}://{host:s}:{balanced_port:d}/cluster/nonexistent/status"
      method: GET
      headers:
        Authorization: "Bearer {test_login_token}"
    response:
      status_code: 404
      json:
        error: 1730
```

jotacarma90 (changes requested):

CHANGELOG.md
ignaciogalle12git: Fixed in b1f8552c14. The echo mention is kept, since F1 is now actually fixed, and the cluster half says 200 with the node reported as a failed item.

framework/wazuh/core/cluster/dapi/dapi.py
ignaciogalle12git: Confirmed, and handled as contract rather than a tighter condition, since the master cannot enumerate configured node names. In 14bbabe403: `forward_request` records that the node universe is `get_connected_nodes()`, the 404 is declared on the eight `/cluster/{node_id}/*` operations, the `node_id` description states that a node not in the cluster at that moment, disconnected included, is answered this way, and the reference's status table says the same. Also took your form of the condition, `e.code == 1730` and `is not None`.

api/test/integration/test_agent_GET_endpoints.tavern.yaml
ignaciogalle12git: Added in 14bbabe403: `404: ResourceNotFoundResponse` on the seven paths, eight operations counting the PUT.

api/api/test/test_error_handler.py
ignaciogalle12git: Right, `json:off` made those a subset match. Pinned in c514214025. The message no longer depends on the submitted value, so the eight stages assert the same string: `Invalid value for query parameter 'offset': must be less than or equal to 2147483647`.

api/api/error_handler.py
ignaciogalle12git: Took the `validator_map` route in 398916c8e4. `WazuhParameterValidator` builds the message from the `ValidationError` object, `validator` and `validator_value`, never from its text, so the submitted value cannot be echoed by construction; every keyword the specification uses on a parameter has its own wording, and one without it is named without its schema value, which keeps a `pattern` regex out of the body. The string surgery is gone: with the validator in place nothing emits `str(ValidationError)` any more, since the body validator already uses `exception.message`.

`GET /groups?groups_list=%00` now answers `Invalid value for query parameter 'groups_list': must match the 'group_names' format`, and a 1025-character `search` answers `must be at most 1024 characters long`.

docs/ref/modules/server-api/api-reference.md
```diff
@@ -517,7 +517,7 @@ Errors follow a structured JSON response format.
 | 400       | Bad Request       | Invalid parameters       |
 | 401       | Unauthorized      | Invalid or expired token |
 | 403       | Forbidden         | RBAC denied              |
-| 404       | Not Found         | Invalid endpoint         |
+| 404       | Not Found         | Invalid endpoint, or the named resource (group, cluster node) does not exist |
```
jotacarma90: **[S3 · non-blocking · introduced]** This row covers a resource that does not exist, but not the surprising half of the new contract: a configured node that is simply not connected at that instant also answers 404, because `get_nodes_info`'s node universe is `get_connected_nodes()`. `spec.yaml`'s `node_id` description states it ("a configured node that is not connected at that moment is reported the same way") and the PR description says the reference does too, but this line and the route table at 396-403 say nothing about it. An operator hitting `GET /cluster/worker1/status` during a rolling restart reads this row and concludes the node left the cluster.

```suggestion
| 404       | Not Found         | Invalid endpoint, or the named resource does not exist: a group, or a cluster node the cluster does not have -- including a configured node that is not connected at that moment |
```

api/api/parameter_validator.py
```diff
@@ -0,0 +1,64 @@
...
+
+import copy
+
+from connexion.utils import is_null, is_nullable
+from connexion.validators import ParameterValidator
+from jsonschema import Draft4Validator, ValidationError
+
+# Wording for every keyword the specification applies to a parameter. Each message names the
+# violated constraint and nothing else: jsonschema's own text opens with the submitted value and is
+# followed by the failing subschema as a Python dict literal, neither of which belongs in a response.
```
jotacarma90: **[S3 · non-blocking · introduced]** This rationale holds for the request body too, and the body path is not covered: connexion's `JSONRequestBodyValidator` and `FormDataValidator` return `f"{exception.message}{error_path_msg}"` (`validators/json.py:75`, `validators/form_data.py:109`), and jsonschema's `.message` opens with `repr(instance)` for `type`, `maxLength`, `pattern`, `format` and `enum`, so `POST /security/users` with `{"password": 1}` still answers `"1 is not of type 'string'"` while a query parameter now answers `"Invalid value for query parameter 'x': must be of type integer"`. The schema dump the issue names was never on that path and `detail` for a body does not reach the access log, so this is not a blocker; but either extend the same pattern with `validator_map={'body': ...}`, or scope the comment to parameters so the next reader does not take it as an API-wide guarantee.

api/api/parameter_validator.py
```diff
@@ -0,0 +1,64 @@
...
+            Reason why the parameter was rejected, or None if it was accepted.
+        """
+        if is_nullable(param) and is_null(value):
+            return None
+
+        if value is None:
+            return f"Missing {parameter_type} parameter '{param['name']}'" if param.get('required') else None
+
+        schema = copy.deepcopy(param)
+        schema = schema.get('schema', schema)
```
jotacarma90: **[S4 · non-blocking · introduced]** The deep copy is dead work: `schema` is only read from here on and `Draft4Validator` does not mutate the schema it is given. It is inherited verbatim from `connexion/validators/parameter.py`, where it survives from older versions that deleted `required` from the copy; 3.1.0 does not, and neither does this override. It costs one full copy of the parameter definition (`description`, `x-scope`, nested `items`) per submitted parameter per request.

```suggestion
        schema = param.get('schema', param)
```

framework/wazuh/core/cluster/dapi/dapi.py
```diff
@@ -555,6 +555,14 @@ async def clean_valid_nodes(nodes_to_clean: List[Tuple]) -> List[Tuple]:
             for node in nodes.items():
                 if node[0] in [node_name['name'] for node_name in allowed_nodes.affected_items] or node[0] == 'unknown':
                     valid_nodes.append(node)
```
ignaciogalle12git: **[S2 · non-blocking · pre-existing]** The new 404 below never fires for the literal node name `unknown`, because this filter accepts it unconditionally, so `valid_nodes` is non-empty and the guard at line 558 is skipped. `clean_valid_nodes()` then rewrites the entry to the local node and the master answers the request as its own, while line 619 deletes the 1730 failed item (its value is `{'unknown'}`) — so `GET /cluster/unknown/status` returns 200 with the master's status and no failure reported, contradicting the contract this PR just wrote into the `node_id` description ("A name the cluster does not have answers 404").

Not introduced by this PR — flagging it for a follow-up issue, not for this review. It is worth closing here anyway: in 5.x `get_solver_node()` only ever returns names that came from `node_id`/`node_list`, so the `unknown` sentinel can no longer be produced internally and is now reachable only by asking for a node literally named that. Under RBAC black mode with `node:id:<master>` denied, `node:id:unknown` stays allowed and reaches the same data.

```suggestion
            for node in nodes.items():
                if node[0] in [node_name['name'] for node_name in allowed_nodes.affected_items] or \
                        ('node_id' not in self.f_kwargs and node[0] == 'unknown'):
                    valid_nodes.append(node)
```

jotacarma90 (approved):

PR #39110, opened 2026-09-10
Retry by-query version conflicts and state what a /stats or /config 200 promises

### LINKED ISSUE #38881: Indexer operations that report success while leaving the data wrong
## Description

Three findings from #38549 of one class: the response is correct and the data is not. Re-verified against `5.0.0` after #38781 (`8f1bc981c3`, the PR that closed #38778), #38727 and #38801.

- **Finding 53**: the resync `deleteByQuery` fails on its own version conflicts, and the retry races the same bulk again. **Open, and aggravated since it was filed.**
- **Finding 47**: a whole-agent deletion can leave documents behind and nothing re-runs it. **Open, unchanged.**
- **Finding 51**: immediate sessions reporting success with the indexer unreachable. **Mostly closed; what remains is narrower than filed, and one premise was wrong.**

### Why #38778 does not cover this

#38778 was a static review of the same connector in eight work packages. Its PR implemented WP-1 to WP-5 and **discarded WP-6, WP-7 and WP-8**. WP-6 *was* finding 47. Objective check: `src/wazuh_modules/inventory_sync_server/src/sync/sessionProcessor.cpp` is not among the 30 files that PR touched.

WP-2 ("fail-closed validation of indexer responses") does touch the same handler as finding 53, but in the **opposite** direction: it wanted more unacknowledged work to fail visibly, while finding 53 says this particular case must stop failing. See below.

---

## 1. The resync `deleteByQuery` turns its own conflicts into a retry loop (finding 53)

**One global conflict policy is shared by two call sites that need opposite behaviour.**

The staged query still sets `conflicts=proceed` (`src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp:1041-1058`), and the shared success handler then treats the skips that `proceed` produces as a hard failure: a `200` reporting non-zero `version_conflicts` sets `deleteByQueryLeftDocuments` (`:416-432`) and the flush throws (`:500-504`). The internal retry loop (`:478-518`) only retries the `429` case, never this one.

Note the contradiction in place: the comment justifying `proceed` says "re-running the delete is the recovery", while the handler treats its skips as fatal. As it stands, `proceed` buys nothing over the default `abort` — it only changes which counter the skip lands in.

The two callers, both in `sessionProcessor.cpp`, both reaching that one handler:

| Caller | Where | What a conflict means there |
|---|---|---|
| Resync / `Cleans` | `sessionProcessor.cpp:340` (flush at `:342`) | A conflicting document is one the agent is about to rewrite anyway — tolerable. But the flush fails, `syncPipeline.cpp:495-506` answers 503/500, the agent re-POSTs the whole session, and the same in-flight rewrite conflicts again. That is the loop. |
| Whole-agent deletion | `sessionProcessor.cpp:374` (flush at `:378`) | A conflict genuinely means documents survive with no writer left — failing is correct. |

**WP-2 made this worse, not better.** The `version_conflicts` → failure logic pre-dates it, but WP-2 removed the two tolerance branches that used to soften the path: the request-level `409` that did "log and continue", and the unparseable `200` that returned silently (now `:395-404` marks it failed). Bulk responses missing `errors` now fail too (`validateBulkResponse`, `:158-170`). So the resync path fails in **more** situations after #38781.

There is a **third copy** of the same intolerance on `_update_by_query` (`:1136-1146`, throw at `:1243-1248`), which is the immediate metadata/groups path (`executeMetadataOrGroups`, `sessionProcessor.cpp:476`) — precisely where a concurrent metadata rewrite is expected.

**Fix direction:** give `deleteByQuery` (and `_update_by_query`) a per-call conflict policy so the resync call site can tolerate its own conflicts while whole-agent deletion keeps treating them as failure — or serialise the delete and the bulk per agent. Adding `conflicts=proceed` is a no-op and an internal retry makes the loop tighter, so the fix as originally filed does not work.

**Before scheduling:** confirm on a stock 5.0.0 that a normal fleet can retry a resync while its previous session's bulk is still indexing. The figure to quote is the reference-hardware **0.4–0.9 % of sessions**, not the 34 % from the lab wave.

## 2. A whole-agent deletion can leave documents behind (finding 47)

Unchanged, and the limitation is documented in place (`sessionProcessor.cpp:362-378`): `_delete_by_query` is a search, so it only sees refreshed segments, and whatever the agent's last session wrote inside the refresh interval survives — with the agent gone from `client.keys`, nothing ever overwrites it. There is no `_refresh` before the purge, and no deferred re-purge, retry or sweep anywhere in the module. The response is an unconditional `ok()` (`:385`).

**New and actionable:** `refresh()` **does exist** on the connector (`indexerConnectorSyncImpl.hpp:1824`, public at `indexerConnectorSync.cpp:234-237`) but is deliberately **absent from the `IIndexerConnectorSync.hpp` seam**, so `sessionProcessor` cannot call it even though the privilege question is now answered — `wazuh_manager` does get refresh on the stateful indices ([wazuh-indexer-plugins#1513](https://github.com/wazuh/wazuh-indexer-plugins/issues/1513)).

Two halves already settled and not what this asks for: the inaccurate log line was softened by #38628, and the async-connector cause of the same leftovers is #38642.

**Fix direction:** expose `refresh()` through the seam and refresh before the purge, or decide who re-runs a deletion that left documents (a scheduled sweep or a deferred second pass). **Before scheduling:** re-check whether the window still reproduces now that the privilege exists.

## 3. Immediate sessions with the indexer unreachable (finding 51) — narrowed, and one premise corrected

**On the sync path this no longer happens.** All three immediate handlers await the indexer before answering: `executeCleans` flushes (`sessionProcessor.cpp:342`), `executeChecksum` runs a live search (`:388-413`), `executeMetadataOrGroups` posts an update (`:476`). Nothing in that file swallows an exception, and `syncPipeline.cpp` re-checks availability at dispatch (`:450-454`) and answers 503/500 on failure (`:486-507`, `:291-310`).

**The original premise is false against 5.0.0.** The report attributed it to an authentication failure leaving the host usable. It does not: the health probe is `GET <host>/_cat/health` and `available` is set true **only** for a parsed `green`/`yellow` status (`src/shared_modules/indexer_connector/src/monitoring.hpp:150-153`); a reachable host answering 401/403 sets only `unavailableReason` (`:186-208`), so `isAvailable()` stays false and every route answers `503`.

**What actually remains is `/config` and `/stats`.** Both gate on `isAvailable()` and answer `503` when false (`configEndpoint.cpp:191-205`, `statsEndpoint.cpp:166-180`), but past that gate they are fire-and-forget: `statsEndpoint.cpp:211` calls `index()`, which only enqueues on the async connector (`indexerConnectorAsyncImpl.hpp:702`), and then answers `200 {}`. The only liveness evidence behind that `200` is a cached monitor verdict up to one `monitoring_interval_seconds` (default 10 s) stale. A `200` there means *queued*, not *stored*.

**Fix direction:** decide what `/config` and `/stats` should promise. Either the `200` is documented as "accepted for indexing" (and the async queue's non-persistence is stated), or those endpoints need an acknowledgement path. Not scoped as code work until that choice is made; the 20-minute experiment originally proposed should now be aimed only at these two endpoints.

---

## Two notes worth recording

**The health probe's permission is a single point of failure of the same shape as finding 47.** The probe needs `cluster:monitor/health`. An indexer role that can write documents but lacks that permission makes every host read unhealthy, so the manager refuses all traffic with `503` — the mirror image of the `indices:admin/refresh` problem above. Worth checking against the role definition while #1513 is fresh.

**Stale comment.** `syncPipeline.cpp:458-461` states that the deletion item carries no responder because "the endpoint answered at admission", but `deleteAgentEndpoint.cpp:190` assigns `item.responder` and `:263-267` says the opposite. Behaviour follows the endpoint (the caller is answered post-flush); the comment is wrong.

## Decisions needed

1. Whether `deleteByQuery`'s (and `_update_by_query`'s) conflict policy becomes per caller, or the resync path stops being able to race itself.
2. Who re-runs a whole-agent deletion that left documents, now that the refresh privilege exists — and whether `refresh()` should be exposed through the seam.
3. What `/config` and `/stats` promise with their `200`.

## Scheduling

The three parts do not belong in the same release slot.

**Finding 53 is the candidate for a patch release.** It is the only one of the three that regressed *within* 5.0.0 — the two tolerance branches it depended on were removed by the PR that closed #38778 — and it has a measured production cost (0.4–0.9 % of sessions on reference hardware failing and retrying a full session, which scales with fleet size). The change is bounded: a conflict-policy parameter plus choosing it at two or three call sites. It does need the "before scheduling" confirmation above first, since the fix direction depends on a normal fleet actually being able to race its own resync.

**Finding 47 is not a patch-release change.** It needs a design decision — who re-runs a deletion that left documents, a scheduled sweep or a deferred second pass — plus exposing `refresh()` through a seam that deliberately does not expose it. Its window is also narrow (documents written inside the refresh interval immediately before the deletion).

**Finding 51 is documentation or a contract decision, not code**, until someone decides what `/config` and `/stats` promise. If the answer is "the `200` means accepted for indexing", it is a docs change that can ship any time.

## Related

- #38549 (parent), findings 47, 51, 53
- #38778 / its PR (WP-1..WP-5 shipped, WP-6/7/8 discarded — WP-6 was finding 47)
- #38642 (the async-connector cause of the same leftover documents)
- [wazuh-indexer-plugins#1513](https://github.com/wazuh/wazuh-indexer-plugins/issues/1513) (the refresh privilege, answered)

ignaciogalle12git:
## Description

A resync's `_delete_by_query` treated its own version conflicts as a hard failure: the flush threw, the session answered 500 and the agent re-POSTed its whole snapshot into the same race. The `_update_by_query` behind the metadata and group paths behaved the same way.

Those conflicts are the index refresh boundary. A by-query delete searches, keeps each document's version and deletes at that version, and the search only reads refreshed segments, so a clean issued right after a bulk over the same documents collides with it. Unlike a shard failure the condition clears itself, so the query is re-run, which is idempotent, and the conflicts are raised only if they outlive a bounded allowance.

Retrying is only correct if re-running is cheap. Both delta queries matched `state.document_version` absent OR `<= global_version` and stamped it unconditionally, so every document one attempt applied was rewritten by the next. They now compare before writing and go `ctx.op = 'noop'` when the document already matches, which also repairs a document sitting at this version whose content drifted.

Closes #38881

That issue carries three findings. 53 is the code above; the other two close here, and the reasons belong in the record.

**Finding 51** closes by documentation, the route the issue itself prescribes: a `200` from `/stats` or `/config` means accepted for indexing, not stored. `docs/ref/modules/inventory-sync-server/api-reference.md` now states the availability gate that answers 503, the enqueue that does not wait, the byte cap that drops on push, the log lines on first overflow and on recovery, the drop on stop, and the 10s cached availability verdict.

**Finding 47** closes with no code, so the argument is the deliverable. Reproducing the window needs a whole-agent deletion to reach the indexer while the agent's last session is still inside the index refresh interval. It cannot get there: `authd` removes the key and only relays the deletion after `purge_delay`, 120s by default (`src/os_auth/src/config.c`), against a `refresh_interval` of 2s to 5s. What is left is a worker that accepts a session more than two minutes after the key is gone, which is not a window this change can close. If `purge_delay` is ever lowered the finding comes back, and nothing in the code will say so.

## Proposed Changes

- A `200` reporting nothing but `version_conflicts` retries instead of failing the caller, on both by-query operations. Shard `failures` and an unparseable body still fail it; with conflicts present the body is retried first, so that failure surfaces once the allowance is spent.
- The allowance is 3 attempts with its own backoff, leaving the transport budget for a following 429. Conflicts that outlive it, or that a stop leaves behind, fail the operation.
- The backoff (2s, then up to 4s) is sized against the worst `refresh_interval` among the indices it covers: 5s for `wazuh-states-sca`, not the 2s common to the rest of the state family.
- One wall clock per operation, stamped once per flush and once per `executeUpdateByQuery` call. The conflict allowance and every transport budget in it draw from that deadline instead of each opening a `max_retry_duration_seconds` window of its own, which is what let a flush outlive the window the caller has to answer in (`remoted.downstream_stateful_response_timeout`, 20s by default). Exceeding it returns a 503, and that 503 is the session re-send that produces the conflict.
- Both delta queries compare before writing, sharing one comparison with the check query of their pair so the check cannot repair what the delta just wrote.
- New counter `sync.indexer.conflict.retries`: a conflict that clears fails no flush, so nothing else counts it.
- Observable change from the `noop`: a delta that finds nothing to change no longer stamps `state.modified_at`. The field moves from "last time a delta was applied" to "last time something actually changed". No reader of it was found in the `wazuh-states-*` indices.

### Results and Evidence

A conflict leaves the documents behind, and re-running clears it. Real wazuh-indexer 5.0.0, agent stopped so nothing repopulated underneath:

```
0. documents at start: 797
1. rewriting the 797 (bumps their versions), no refresh after it
2. delete right after the write, before the refresh: total:797 deleted:0   version_conflicts:797
3. documents left after a refresh: 797            <- the conflicted ones really do survive
4. the same delete again, index now refreshed:    total:797 deleted:797 version_conflicts:0
5. documents left: 0
```

Line 3 is why tolerating the conflict is not an option; line 4 is the fix. It has to wait: repeated 4 ms later, the same delete conflicts on all 797 again.

Reachable on `5.0.0`. Manager built from `5.0.0`, agent 003 resyncing three indices at `integrity_interval` 10s, indexer throttled. 1 of 27 `Cleans` deletes conflicted, right after the bulk flushed before it:

```
15:11:22 Bulk operation completed with no errors
15:11:22 Deleting by query: .../wazuh-states-fim-files/_delete_by_query
15:11:28 Response: {"took":5301,"total":657,"deleted":597,"version_conflicts":60,"failures":[]}
```

| Suite | Result |
|---|---|
| `indexer_connector_utest` | 313/313 pass |
| `inventory_sync_server_utest` | 255/255 pass |

Not covered by evidence: a populated `failures` inside a `200` could not be induced on a healthy single-node cluster, and the stop that cuts a retry wait short is not unit-testable (`m_stopping` is only set by the destructor). Both are argued from the code.

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

Windows, macOS, Engine and API/Framework do not apply: the code is manager-only. Memory tests were not run.

### Artifacts Affected

`libindexer_connector.so` and `wazuh-inventory-sync-server`. No packaging or default configuration changes.

### Configuration Changes

N/A.

### Tests Introduced

Both by-query outcome tables keep their conflicts-only row, which passes through the allowance before it throws. A parameterized pair covers the delete rule (the retry clears it: 2 POSTs, no throw; the allowance runs out: 3 POSTs, throw), one covers the same convergence on `executeUpdateByQuery`, and the stats drain asserts the counter reaches the metric. On the sync server, one test pins that both deltas gate on `lte`, compare before writing and write their own payload.

## Review Checklist

- [ ] Code changes reviewed

No changelog entry: the conflict handling this changes is new in the unpublished 5.0.0.


src/wazuh_modules/inventory_sync_server/src/sync/sessionProcessor.cpp
```diff
@@ -335,9 +335,18 @@ namespace invsync::sync
...
+        // still writing and this one replaces, while failing answers 503 and the agent's retry
+        // re-POSTs the whole session into the same race. wazuh-states-vulnerabilities is the
+        // exception -- agents may clean it but never write it, so a document skipped there has no
+        // writer left at all, exactly like a whole-agent deletion.
         for (const auto& index : indices)
         {
-            connector.deleteByQuery(index, session.agentId, m_managerClusterName);
+            const auto conflicts = (index == VULNERABILITIES_INDEX) ? indexer::VersionConflicts::Fail
+                                                                    : indexer::VersionConflicts::Tolerate;
+            connector.deleteByQuery(index, session.agentId, m_managerClusterName, conflicts);
```
ignaciogalle12git: **[S2 · blocking · introduced]** `Tolerate` is selected by index name, but the property that justifies it ("a `SyncData` follows and rewrites this index") belongs to the session, and the wire does not carry it: `DataClean` holds only `index`, and every clean arrives as `Mode_ModuleDelta` (`fullSessionValidator.cpp`, note D6). At least three agent paths send a `Cleans` that nothing follows — `wm_syscollector.c:499-537` (module disabled: one clean over all 14 indices, retried until the manager answers success, then `syscollector_delete_database_ptr()` and `pthread_exit`), `sca_impl.cpp:1811-1848` (`deleteDatabase()` then `m_exitAfterDataClean`), and the FIM disable paths — so a document skipped there has no writer, no re-clean and no integrity check left, because the collector is off. Today's `Fail` makes those agents retry the clean until it converges, which this change replaces with permanent orphans; and the same PR argues the conflict window is common enough to need tolerating, so it cannot also be rare enough to be safe here.

Keeping `Fail` is the only safe state until the session can say a `SyncData` follows — a 503 on a resync costs one retry, an orphan on a teardown costs forever. Two ways forward, neither a one-liner: a wire signal (a field on `DataClean`, or an `Option` value set only by the recovery path at `recovery.c:119-126`, defaulting to `Fail`) which needs the agent team; or a bounded retry of the `_delete_by_query` itself on `version_conflicts`, which fixes both cases without a 503 and without touching the wire contract.

```suggestion
            // Fail until the session itself can say a SyncData follows. A Cleans is also how a
            // module announces a permanent teardown (wm_syscollector.c:499-537,
            // sca_impl.cpp:1811-1848): there the agent deletes its local copy and stops the module
            // on the 200, so a skipped document has no writer, no re-clean and no integrity check
            // left, while today's 503 makes that same agent retry the clean until it converges.
            connector.deleteByQuery(index, session.agentId, m_managerClusterName, indexer::VersionConflicts::Fail);
```

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -409,17 +420,33 @@ class IndexerConnectorSyncImpl final
...
+            // caller is rewriting.
+            const auto countOf = [&parsed](const char* field) -> std::size_t
+            {
+                return (parsed.contains(field) && parsed.at(field).is_number_integer())
+                           ? parsed.at(field).get<std::size_t>()
+                           : 0;
+            };
+            const auto conflicts = countOf("version_conflicts");
+
+            if (failures == 0 && (conflicts == 0 || tolerateVersionConflicts))
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** The tolerate condition still has no ceiling: `{"total":400,"deleted":0,"version_conflicts":400,"failures":[]}` passes as a fully successful delete, so `executeCleans` answers 200 for a clean that removed nothing. The rework reads `total` and `deleted` but only to print them in the INFO at `:443`, which is a real mitigation and not a bound — and on the teardown clean of F16 the agent drops its local copy right after that 200, so "deleted 0 of 400" means 400 permanent orphans. Persists from the previous review.

If you want a bound, prefer a fractional one (`conflicts < total`) over `deleted > 0`: requiring a deletion would re-fail the flush in exactly the maximum-contention case finding 53 exists to stop failing. The PR's evidence only exercises 2/400 and 13/400, so the extreme is unmeasured.

docs/ref/modules/indexer_connector/README.md
```diff
@@ -93,20 +93,26 @@ header is not honored — the transport does not expose response headers (tracke
...
 - **A `200` is not automatically success.** The response is inspected, and the flush throws when it
-  reports per-shard `failures` or a non-zero `version_conflicts` — the two ways a `200` can leave
-  matching documents in place — or when its body cannot be parsed at all. Callers treat that as
+  reports per-shard `failures`, or when its body cannot be parsed at all. Callers treat that as
   retriable.
+- **A version conflict is the caller's call.** The skips `proceed` tallies in `version_conflicts`
+  are the other way a `200` leaves matching documents in place, so what to do with them is the
+  `conflicts` parameter: `IndexerVersionConflicts::Fail` (the default) fails the flush like a shard
+  failure, `Tolerate` logs the skip and confirms the run, for a caller that rewrites those documents
+  itself. `failures`, an unparseable body and missing counters fail under both.
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** "missing counters fail under both" is true of `executeUpdateByQuery` only, which throws out of `responseJson.at("updated")`. On the delete path `countOf` (`indexerConnectorSyncImpl.hpp:428-433`) reads an absent or non-integer field as 0, so a body with no counters is confirmed as a complete deletion — and this PR's own parameter row pins that deliberately, labelled *"A response that predates these fields must still pass"*. As written the reference promises a validation that does not exist, at the point where a caller would decide whether it needs one of its own.

```suggestion
- **A version conflict is the caller's call.** The skips `proceed` tallies in `version_conflicts`
  are the other way a `200` leaves matching documents in place, so what to do with them is the
  `conflicts` parameter: `IndexerVersionConflicts::Fail` (the default) fails the flush like a shard
  failure, `Tolerate` logs the skip and confirms the run, for a caller that rewrites those documents
  itself. `failures` and an unparseable body fail under both; a delete-by-query body that simply
  lacks the counters is read as zero and still passes.
```

src/shared_modules/indexer_connector/tests/unit/indexerConnectorSync_test.cpp
```diff
@@ -1493,24 +1495,78 @@ INSTANTIATE_TEST_SUITE_P(
...
+                }
+                else
+                {
+                    std::get<TPostRequestParameters<std::string&&>>(postParams).onSuccess(std::string(body));
+                }
+            }));
+
+    IndexerConnectorSyncImplTest connector(config, nullptr, &mockHttpRequest, std::move(mockSelector));
+    connector.deleteByQuery("wazuh-states-inventory-packages", "007", "cluster01", IndexerVersionConflicts::Tolerate);
+    connector.deleteByQuery("wazuh-states-inventory-packages", "008", "cluster01", IndexerVersionConflicts::Fail);
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** This only stages `Tolerate` then `Fail`, which is not the order that catches a wrong merge. Write the branch at `indexerConnectorSyncImpl.hpp:1094` the obvious wrong way — `it->second.tolerateVersionConflicts = (conflicts == IndexerVersionConflicts::Tolerate);` inside the `else` — and this test still passes: `Tolerate` then `Fail` still ends at `false` and still throws. `Fail` then `Tolerate` is the direction that would flip to tolerating and hand the intolerant caller a success for an incomplete delete.

Add it as a second test with its own connector and flush, not two more lines here: with both pairs in one flush, the first index's throw already satisfies the assertion.

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -367,6 +373,26 @@ class IndexerConnectorSyncImpl final
...
+            LOGFN_WARN(m_logFn,
+                       "%s left documents behind: version conflicts persisted across %zu attempts.",
+                       operation,
+                       CONFLICT_RETRY_ATTEMPTS);
+            return false;
+        }
+        LOGFN_DEBUG2(m_logFn, "Retrying %s in %lld ms.", operation, static_cast<long long>(delay.count()));
+        std::unique_lock<std::mutex> lock(m_retryMutex);
+        m_retryCv.wait_for(lock, delay, [this]() { return m_stopping.load(); });
+        return true;
```
ignaciogalle12git: **[S2 · blocking · introduced]** `retryAfterVersionConflict` answers `true` whether it slept the whole wait or `m_stopping` woke the cv early, and only the second case means the query is never going to re-run. The loop then falls into the stop early-return at `:517` (and `:1274` on the update path, which also drops `m_notify` without firing it), so `processBulk` returns normally, `flush()` does not throw, and `executeCleans` answers 200 for a clean that deleted nothing while the agent goes on to delete its local copy. Before this PR that same conflicting 200 threw unconditionally.

```suggestion
        LOGFN_DEBUG2(m_logFn, "Retrying %s in %lld ms.", operation, static_cast<long long>(delay.count()));
        std::unique_lock<std::mutex> lock(m_retryMutex);
        m_retryCv.wait_for(lock, delay, [this]() { return m_stopping.load(); });
        // A stop cuts the wait short without re-running the query, so the skipped documents are
        // still there: report them instead of falling into the loop's silent early return.
        return !m_stopping.load();
```

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -475,6 +511,7 @@ class IndexerConnectorSyncImpl final
                 IndexerExponentialBackoff retryBackoff {std::chrono::seconds {RetryDelay},
                                                         std::chrono::seconds {m_maxRetryDelay}};
                 IndexerRetryBudget retryBudget {m_maxRetryAttempts, m_maxRetryDuration};
+                IndexerRetryBudget conflictBudget {CONFLICT_RETRY_ATTEMPTS, std::chrono::milliseconds {0}};
```
ignaciogalle12git: **[S2 · blocking · introduced]** `conflictBudget` and `retryBackoff` are constructed inside the per-index `for`, so every staged index gets a fresh three-attempt allowance and the added wait is N x up to ~2.5s with no flush-level bound. The refresh boundary that justifies the retry is a property of the flush, not of an index (the same unrefreshed bulk conflicts on every index it touched), so the multiplication buys latency and no convergence: with the `took: 5301` from your own evidence, three conflicted indices reach ~23s against the 20s remoted window that `:341-346` says this budget is sized to fail before. The wait is also held under `m_mutex` (`flush()` `:1830`, and the adapter's `scopeLock()` seams), so on the 200 path it now stalls every other agent on the connector, which before only happened after a transport failure.

Hoisting both above the loop is the whole fix:

```c++
        const bool hasDeleteByQuery = !m_deleteByQuery.empty();
        // One refresh boundary per flush, not per index: the same unrefreshed bulk conflicts on
        // every index it touched, so a per-index allowance buys latency and no convergence.
        IndexerExponentialBackoff conflictBackoff {std::chrono::seconds {RetryDelay},
                                                   std::chrono::seconds {2 * RetryDelay}};
        IndexerRetryBudget conflictBudget {CONFLICT_RETRY_ATTEMPTS, std::chrono::milliseconds {0}};

        for (const auto& [index, query] : m_deleteByQuery)
```

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -497,6 +535,11 @@ class IndexerConnectorSyncImpl final
                             .url = HttpURL(url), .data = query.dump(), .secureCommunication = m_secureCommunication},
                         PostRequestParameters {.onSuccess = onSuccessDeleteByQuery, .onError = onErrorDeleteByQuery},
                         ConfigurationParameters {.timeout = m_requestTimeoutMs});
+                    if (deleteByQueryConflicted &&
+                        !retryAfterVersionConflict(retryBackoff, conflictBudget, "deleteByQuery"))
+                    {
+                        deleteByQueryLeftDocuments = true;
+                    }
```
ignaciogalle12git: **[S3 · introduced]** Only the attempt count is separate: `retryBackoff` is the same object the transport retry advances, and `retryBudget`'s deadline is fixed at construction, so two conflicts leave a following 429 waiting 2-8s instead of 1s with ~2.5s already gone from its 15s, which is not "the transport budget stays intact" as the comment and the README both say. The sizing is also off by the jitter: `nextDelay()` returns the base exactly the first time and `uniform[1s, 2s]` the second, so the cumulative wait is 2.0-3.0s against a 2s `refresh_interval`, and at the low draw there is no margin left for the third attempt. Minor, same place: `nextDelay()` runs before `budget.exhausted()`, so the give-up path burns a backoff step it never sleeps.

Giving conflicts their own `IndexerExponentialBackoff` (see the previous comment) makes the claim true for everything except the wall-clock deadline, which is worth sharing on purpose since the caller's window is absolute - just say so in the comment.

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -367,6 +373,26 @@ class IndexerConnectorSyncImpl final
...
+    retryAfterVersionConflict(IndexerExponentialBackoff& backoff, IndexerRetryBudget& budget, const char* operation)
+    {
+        const auto delay = backoff.nextDelay();
+        if (budget.exhausted(delay))
+        {
+            LOGFN_WARN(m_logFn,
+                       "%s left documents behind: version conflicts persisted across %zu attempts.",
+                       operation,
+                       CONFLICT_RETRY_ATTEMPTS);
+            return false;
```
ignaciogalle12git: **[S3 · introduced]** On a default install this WARN is the only line an operator sees for the new failure path, and it carries neither the conflict count, nor the index, nor the response body. All three are now DEBUG2-only (`:455-457`, gated on `debug=2`), because the WARN at `:462` that used to carry them is no longer reachable in the conflicts-only case - `:453` returns first. The count is the number that says whether three attempts is enough, which is exactly what anyone reading this line next will want to know.

Cheapest fix that adds no new state: make `deleteByQueryConflicted`/`updateByQueryConflicted` a `std::size_t` holding the count (condition becomes `> 0`), pass it to `retryAfterVersionConflict` and log it here.

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -1240,6 +1294,10 @@ class IndexerConnectorSyncImpl final
                                 PostRequestParameters {.onSuccess = onSuccess, .onError = onError},
                                 ConfigurationParameters {.timeout = m_requestTimeoutMs});
 
+            if (updateByQueryConflicted && !retryAfterVersionConflict(retryBackoff, conflictBudget, "Update by query"))
+            {
+                updateByQueryFailed = true;
+            }
```
ignaciogalle12git: **[S2 · aggravated]** The retry re-issues the identical `_update_by_query`, and that query is not self-limiting the way the delete is: `buildMetadataUpdateQuery` matches `state.document_version` absent OR `lte` globalVersion, and the script sets `document_version = globalVersion` with no `ctx.op = 'noop'` guard, so every document the previous attempt already updated still matches and gets rewritten, re-stamping `modified_at` and bumping `_version` again. A realistic `{"updated":49940,"total":50000,"version_conflicts":60}` therefore rewrites 49940 documents to recover 60, up to three times, and each of those version bumps is exactly the input that makes a concurrent `_delete_by_query` conflict. The delete path is fine for the opposite reason, which your own evidence shows: `total:657, deleted:597, conflicts:60`, and the 597 stop matching.

Worth making the retry cheap before shipping it: either the range becomes `lt`, or the script opens with a `document_version == globalVersion` noop guard. Both also change the first attempt (no more re-stamping `modified_at` on already-current documents), so it needs whoever set the `external_gte` semantics; if neither is acceptable, cap `_update_by_query` at one attempt.

src/wazuh_modules/inventory_sync_server/src/sync/syncQueryBuilder.hpp
```diff
@@ -59,11 +60,11 @@ namespace invsync::sync::querybuilder
...
         nlohmann::json updateQuery;
 
-        // Match wazuh.agent.id AND only update if version <= globalVersion (external_gte behavior):
-        // documents where state.document_version is null OR <= globalVersion.
+        // Strictly less: a metadata delta is max + 1, so a document already at globalVersion holds
+        // this session's own values and a retry must not rewrite it.
         updateQuery["query"]["bool"]["must"][0]["term"]["wazuh.agent.id"] = agentId;
         updateQuery["query"]["bool"]["should"][0]["bool"]["must_not"]["exists"]["field"] = "state.document_version";
-        updateQuery["query"]["bool"]["should"][1]["range"]["state.document_version"]["lte"] = globalVersion;
+        updateQuery["query"]["bool"]["should"][1]["range"]["state.document_version"]["lt"] = globalVersion;
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** `lt` makes the metadata reconciliation depend on the agent's versioning convention (`MetadataDelta` = max + 1, `GroupDelta` = max), which is true for the only production emitter (`agent_info_impl.cpp:1886`, `:1934`) but is stated in no protocol reference (`sync-protocol/api-reference.md:314`, `agent_info/architecture.md:51` says "max version or max+1" without which is which, `inventory-sync-server/api-reference.md:243` calls the guard "stale" when an *equal* version is now stale too), and the agent's max comes from its local module DBs, so after an agent DB reset under the same id a metadata delta that lands exactly on the old max (`lt 7` against documents at 7 with the old hostname) is a `200` that repairs nothing until the next `MetadataCheck`, where `lte` rewrote them. What removes both dependencies is already in this file: keep `lte` and make this script compare first the way `buildGroupsUpdateQuery` now does and `buildMetadataCheckQuery` already does field by field (`:172-180`), with the version condition and the stamp in the `else` branch; a retry then noops on what the previous attempt applied and an equal version with different content is still repaired. If `lt` stays, write the convention into those two reference rows and pin it with a `qa/test_metadata_groups.py` row that sends a `metadata_delta` at the current max and asserts nothing changes.

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -345,6 +346,13 @@ class IndexerConnectorSyncImpl final
     /// a flush before that window (remoted's downstream response timeout, default 20 s) closes;
     /// the cross-team sizing of the whole timeout chain is decided elsewhere.
     static constexpr size_t DEFAULT_MAX_RETRY_ATTEMPTS {5};
+
+    /// A by-query version conflict clears itself after one index refresh, so it retries on its own
+    /// attempt count and its own backoff, leaving the budget above intact for a 429 that follows.
+    /// Three attempts wait 1s and then up to 2s, over the 2s `refresh_interval` of the state
+    /// indices, and no longer: the wait is on the caller's request and on the shard queue behind
+    /// it. The wall bound is shared, because the window the caller has to answer in is absolute.
+    static constexpr size_t CONFLICT_RETRY_ATTEMPTS {3};
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** This fixed allowance contradicts the operator contract in `docs/ref/modules/inventory-sync-server/configuration.md:459-465`, which describes `max_retry_attempts` as the attempts bound of "one connector operation (a bulk flush ..., a delete-by-query, an update-by-query, a search)" and documents `1` as "a single attempt with no retry": a conflicting by-query is re-POSTed up to three times with 1 s + [1,2] s of sleep whatever that option says, while the same loop does honour `max_retry_duration_seconds` (`:523`, `:1291`). Keeping it constant is a decision the description defends; the page that says otherwise needs one sentence (own fixed allowance of 3, only the duration bound applies), and the Tuning column of `sync.indexer.conflict.retries` in `metrics.md:118` can point at `max_retry_duration_seconds` instead of being empty.

src/wazuh_modules/inventory_sync_server/src/sync/syncQueryBuilder.hpp
```diff
@@ -113,10 +114,33 @@ namespace invsync::sync::querybuilder
...
+            "  else { "
+            "    def sortedCurrent = new ArrayList(currentGroups); "
+            "    Collections.sort(sortedCurrent); "
+            "    def sortedExpected = new ArrayList(params.groups); "
+            "    Collections.sort(sortedExpected); "
+            "    for (int i = 0; i < sortedCurrent.size(); i++) { "
+            "      if (sortedCurrent[i] != sortedExpected[i]) { needsUpdate = true; break; } "
+            "    } "
+            "  } "
+            "} "
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** These 14 lines are byte-identical to `buildGroupsCheckQuery`'s `:217-230`, so a fix to the compare (a null element, a `Collections.sort` exception on non-string groups) applied to one script and not the other makes a `GroupDelta` and a `GroupCheck` disagree on whether a document needs writing, and the check then "repairs" what the delta just wrote every `integrity_interval`. One shared fragment (a namespace-level `constexpr std::string_view`) that both builders concatenate leaves each with only what differs: the version condition and the stamp here, nothing in the check.

src/wazuh_modules/inventory_sync_server/test/unit/sessionProcessor_test.cpp
```diff
@@ -305,12 +305,13 @@ TEST_F(SessionProcessorTest, MetadataAndGroupModesRunOneScopedUpdateByQuery)
...
-        const char* mustContain; // a fragment that identifies the script family
-        const char* mustNotContain;
+        const char* mustContain;    // a fragment that identifies the script family
+        const char* mustNotContain; // nullptr when nothing separates it from its sibling family
     };
     const Case cases[] = {
         {invsync::test::fb::Mode_MetadataDelta, "state.document_version = params.globalVersion", "needsUpdate"},
-        {invsync::test::fb::Mode_GroupDelta, "wazuh.agent.groups = params.groups", "needsUpdate"},
+        // The group delta now compares like the check does; only the stamp tells them apart.
+        {invsync::test::fb::Mode_GroupDelta, "state.document_version = params.globalVersion", nullptr},
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** No unit test asserts any more that the `GroupDelta` script writes the groups: this row's `mustContain` moved from the groups write to the version stamp, which the metadata script carries too, and `TheMetadataAndGroupVersionGatesDifferOnPurpose` checks `lte`, the noop and `modified_at` but not the write, so dropping `ctx._source.wazuh.agent.groups = params.groups;` from the `else` branch keeps both green. Only `mustNotContain` had to change (the script now contains `needsUpdate`); the groups fragment still identifies the family, and the comment can state the invariant instead of the history.

```suggestion
        // The group delta and the group check share the compare; the version and modified_at stamp is what
        // tells the delta apart, so only the groups write identifies the family here.
        {invsync::test::fb::Mode_GroupDelta, "wazuh.agent.groups = params.groups", nullptr},
```

jotacarma90 (changes requested):
The conflict retry is the right call and the evidence behind it is solid: line 3 of the
measurement (the conflicted documents survive a refresh) is what rules out tolerating them, and
line 4 is what proves re-running works. `compare before writing` is a good addition too, since a
`noop` consumes no `_seq_no` and therefore removes conflicts instead of waiting them out.

Four things before this merges. Only the first one is a blocker.

### 1. The painless scripts have no execution coverage (blocker)

This PR rewrites four painless scripts, and the unit tests that cover them **compare strings, they
do not run them**. A syntax error or a wrong comparison passes all 255 tests and only fails against
a real indexer. CI did not catch this either: both IT jobs reported `Skipped — no relevant changes`.

The coverage already exists and is not listed in the evidence:

```
qa/test_metadata_groups.py     test_metadata_delta_updates_every_declared_index
                               test_stale_metadata_never_overwrites_newer
                               test_groups_delta_updates_groups
                               test_metadata_check_repairs_only_on_mismatch
qa/test_cleans_and_resync.py   test_full_resync_is_cleans_then_delta
```

Please run those against a real indexer and add the output to Results and Evidence.

### 2. The worst case can exceed the window the caller has to answer in

`conflictBudget` is built with its own `m_maxRetryDuration`, a 15 s window independent of the
per-index transport budget. With the delete latency this PR's own evidence measured (5301 ms), a
flush that hits a conflict costs roughly:

```
batch cut _bulk            ~5.3 s
delete attempt 1           ~5.3 s
wait                        1.0 s
delete attempt 2           ~5.3 s
wait                       1–2 s
delete attempt 3           ~5.3 s
                        —————————
                          ~23–24 s
remoted downstream_stateful_response_timeout   20 s
```

Past that timeout the agent gets a 503 and re-POSTs the session, which is exactly the retry that
races the in-flight bulk and produces the conflict in the first place. Deriving the conflict
allowance from the time the flush has left, rather than giving it a fresh budget, would close that
loop. At the very least the trade-off should be stated in the description.

### 3. The `refresh_interval` is not uniform

`indexerConnectorSyncImpl.hpp:352` and `docs/ref/modules/indexer_connector/README.md:107` both size
the backoff against "the 2s `refresh_interval` of the state indices". Per the templates in
`wazuh-indexer-plugins`:

| Index | `refresh_interval` |
|---|---|
| inventory (14), fim (3), vulnerabilities | `2s` |
| **sca** | **`5s`** |
| agent-config, agent-stats | `5s` (by-id deletes, not affected) |

`wazuh-states-sca` is in both the `Cleans` scope and `AGENT_DELETION_SCOPE_BY_QUERY`, and the new
`AnUpdateByQueryRetriesAVersionConflictUntilItClears` test uses it. Either state the real range and
why 3 attempts still converge at 5s, or size the allowance against the worst case.

### 4. Two comments now contradict decisions that have been made

`sessionProcessor.cpp:367-370` still reads:

> `_refresh` needs `indices:admin/refresh`, which the manager's least-privilege indexer role does
> not grant: every deletion then failed with 403 [...] see the follow-up to restore the refresh once
> the privilege is in place.

Both halves are stale. The privilege was granted in wazuh-indexer-plugins#1513, and more to the
point the refresh was never part of the design: the purge defers instead, deliberately, for the
indexer's cost under load and the cross-worker document sync. There is no follow-up to restore.
Since this PR closes the issue, that comment should say what was actually decided.

`syncPipeline.cpp:469-470` also still claims the deletion item "carries no responder", which
`deleteAgentEndpoint.hpp:166` contradicts. The issue flagged it; it is a one-liner.

### 5. The description closes three findings but only argues one

`Closes #38881` covers findings 47, 51 and 53. The body explains 53 only. 51 is covered by the
api-reference change but is not mentioned, and 47 is closed with no code and no rationale in the
PR at all. Worth one paragraph each, especially 47: the reason it is accepted (the 120 s
`authd.purge_delay` covers the window by design, and the residual hole needs a worker accepting a
session past two minutes) is the thing a future reader will need.

### 6. Minor: `state.modified_at` changes meaning

With the `noop`, a delta no longer stamps `modified_at` when nothing changed. The field goes from
"last time a delta was applied" to "last time something actually changed". Probably intended, but
it is an observable change for anything reading it as a freshness signal, so it belongs in the
description.

ignaciogalle12git:
Points 1-6 addressed.

**1. Painless execution coverage.** `inventory-sync-server-integration` already runs on every push of this PR (the workflow's paths cover `src/wazuh_modules/inventory_sync_server/**` and `src/shared_modules/indexer_connector/**`) and was green on `da808131`. The `Skipped - no relevant changes` pair is `IT Linux` / `IT MacOS`, which are the `tests/integration/` suites, not this one.

What was missing is coverage, not execution: the three delta rows all send changed metadata, so they pass against the unconditional stamp the scripts carried before. Two rows added that do not:

- `test_metadata_delta_is_a_noop_when_nothing_changed`, asserted on `_seq_no` rather than `state.modified_at` because `getCurrentISO8601()` has second resolution and two back-to-back sessions can share a timestamp.
- `test_metadata_delta_repairs_drifted_content_at_its_own_version`, which pins `lte` over `lt`.

[Run 34852155139](https://github.com/wazuh/wazuh/actions/runs/34852155139) on `057f4436`, real OpenSearch:

```
test_metadata_groups.py::test_metadata_delta_updates_every_declared_index PASSED               [ 46%]
test_metadata_groups.py::test_stale_metadata_never_overwrites_newer PASSED                     [ 47%]
test_metadata_groups.py::test_groups_delta_updates_groups PASSED                               [ 49%]
test_metadata_groups.py::test_metadata_delta_is_a_noop_when_nothing_changed PASSED             [ 50%]
test_metadata_groups.py::test_metadata_delta_repairs_drifted_content_at_its_own_version PASSED [ 52%]
test_metadata_groups.py::test_metadata_check_repairs_only_on_mismatch PASSED                   [ 53%]
test_metadata_groups.py::test_metadata_over_forbidden_indices_is_a_noop PASSED                 [ 55%]

======================== 62 passed, 1 skipped in 37.70s ========================
```

**2. Retry budget.** One wall clock per operation, stamped once per flush in `processBulk` and once per call in `executeUpdateByQuery`. The conflict allowance no longer opens a `max_retry_duration_seconds` window on top of the transport one, and neither does each index staged in a flush: before this, a `Cleans` over 14 indices could spend 14 x 15s of transport retries plus 15s of conflicts.

**3. `refresh_interval` is not uniform.** `wazuh-states-sca` is `5s`, not the `2s` the schedule assumed, and this retry covers it too (`isAgentScopedStateIndex`, `AGENT_DELETION_SCOPE_BY_QUERY`). The old schedule exhausted its 3 attempts at ~2-3s cumulative wait, before a 5s segment could refresh, so `sca` conflicts reproduced the original failure. Sized the allowance against that worst case instead: `ConflictRetryDelayMs` 1000 → 2000, three attempts now wait 2s then up to 4s (~4-6s cumulative). Doc comment and connector reference state both values now, not the single assumed one.

**4.** Both comments corrected: the deferral before the purge now reads as the decision it is, and the `DeleteAgent` item as carrying its responder.

**5, 6.** The description carries a paragraph each for findings 51 and 47, plus the `state.modified_at` change.

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -312,7 +313,8 @@ template<typename TSelector,
          size_t MaxBulkSize = 10 * 1024 * 1024,
          size_t RetryDelay = 1,
          size_t FlushInterval = 20,
-         size_t MaxRetryDelay = 15>
+         size_t MaxRetryDelay = 15,
+         size_t ConflictRetryDelayMs = 2000>
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** The cumulative wait before the last attempt is 4s at minimum, not the 5s the comment at `:355-356` and the README say it is sized against: `nextDelay()` returns exactly `ConflictRetryDelayMs` the first time and a uniform [2000, 4000] ms the second, so the total is uniform over [4s, 6s] and lands under 5s about half the time. On `wazuh-states-sca` those draws fire the third POST before the segment refreshed, the allowance is spent and the deletion fails — the outcome this retry exists to remove. 2500 ms makes the lower bound deterministic at 5s; otherwise the honest fix is to say in the comment and the README that the 5s case is covered probabilistically.

```suggestion
         size_t ConflictRetryDelayMs = 2500>
```

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -468,13 +537,24 @@ class IndexerConnectorSyncImpl final
         // Track if we have pending deleteByQuery operations that need notification
         const bool hasDeleteByQuery = !m_deleteByQuery.empty();
 
+        // One wall clock for the whole flush, so the deletes, the conflict allowance and the bulk
+        // draw from the same window instead of each opening its own.
+        const auto flushDeadline = std::chrono::steady_clock::now() + m_maxRetryDuration;
+
+        // One conflict allowance for the whole flush: the refresh boundary is a property of the
+        // flush, so the first index to wait it out clears the rest, and a per-index allowance would
+        // only multiply the wait by the number of indices a Cleans staged.
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** This justification is the one thing `70ca6bb4` falsifies: if `refresh_interval` is 2s for most state indices and 5s for `wazuh-states-sca`, then the first index waiting it out does **not** clear the rest, and with a single 3-attempt counter across the loop the first index to conflict spends one of the two waits, leaving `wazuh-states-sca` a single 2-4s wait against its 5s. Either give the allowance per index (and cap the total by the flush deadline, which `budgetLeft` already provides), or state here that the shared allowance is really a time budget expressed in attempts and only covers the 2s family. Related: `conflictBackoff` is never `reset()` between indices, so the second index to conflict starts at the longer delay and the per-index behaviour depends on `m_deleteByQuery` ordering.

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -468,13 +537,24 @@ class IndexerConnectorSyncImpl final
...
+        IndexerExponentialBackoff conflictBackoff {std::chrono::milliseconds {ConflictRetryDelayMs},
+                                                   std::chrono::seconds {m_maxRetryDelay}};
+        IndexerRetryBudget conflictBudget {CONFLICT_RETRY_ATTEMPTS, budgetLeft(flushDeadline)};
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** `m_maxRetryDuration` dominates `CONFLICT_RETRY_ATTEMPTS` as soon as the by-query itself takes a few seconds, which is exactly when conflicts happen. With the `"took":5301` from this PR's own Results section: attempt 1 returns at 5.3s, the 2s wait passes the budget check, attempt 2 returns at 12.6s, and `exhausted(delay)` is true whenever the second draw exceeds 2400 ms — 80% of the time — so the operation makes two attempts while the WARN at `:412-418` prints `(3 attempt(s), 15000 ms)`. The bound winning is correct (past the caller's window there is no point retrying); what is missing is that neither this comment, nor the README, nor `configuration.md:466-469` says the 3 attempts are a ceiling rather than a floor.

src/shared_modules/indexer_connector/src/indexerConnectorSyncImpl.hpp
```diff
@@ -468,13 +537,24 @@ class IndexerConnectorSyncImpl final
         // Track if we have pending deleteByQuery operations that need notification
         const bool hasDeleteByQuery = !m_deleteByQuery.empty();
 
+        // One wall clock for the whole flush, so the deletes, the conflict allowance and the bulk
+        // draw from the same window instead of each opening its own.
+        const auto flushDeadline = std::chrono::steady_clock::now() + m_maxRetryDuration;
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** This turns `max_retry_duration_seconds` into a window for the whole flush, and `configuration.md` still defines it as the deadline "for one connector operation's retries" (`:481`) after listing a delete-by-query and a bulk flush as separate operations (`:460-461`). A Cleans staging six deletes plus a bulk now shares one 15s window across all seven: if the deletes spend it, `budgetLeft` returns 1ms and the bulk's first `429` exhausts on the spot, with no retry at all. The behaviour change is defensible — `configuration.md:482-484` already says the default is sized so a flush fails while its caller is still listening — so what needs updating is the sentence defining the unit, not the code.

src/wazuh_modules/inventory_sync_server/src/sync/sessionProcessor.cpp
```diff
@@ -364,10 +364,9 @@ namespace invsync::sync
         // (authd removes the key, then relays after its purge delay), so whatever that session wrote
         // inside the index refresh interval is invisible to the query and survives -- and with the
         // agent gone from client.keys nothing ever overwrites it. Refreshing each index first would
-        // close that window, but `_refresh` needs `indices:admin/refresh`, which the manager's
-        // least-privilege indexer role does not grant: every deletion then failed with 403 instead
-        // of missing a few documents. Repeating the deletion is the recovery meanwhile; see the
-        // follow-up to restore the refresh once the privilege is in place.
+        // close it, and is deliberately not done: deferring the purge costs the indexer less under
+        // load and lets the other workers' writes for this agent land first. Repeating the deletion
+        // is the recovery.
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** Retracting the `403`/`indices:admin/refresh` rationale leaves five places still asserting it: `architecture.md:288-291` and `api-reference.md:195-197` (both edited by this PR), `README.md:629-632`, `tools/send_delete_agent.py:21`, and `qa/test_delete_agent.py:99-112`, whose skip reason is that `403` and whose unskip condition ("once the privilege is granted") can now never be met, so a real known limitation is left with no test and no tracked follow-up. The replacement reason also does not fit what it annotates: "deferring the purge costs the indexer less under load and lets the other workers' writes for this agent land first" describes authd's purge delay, not the absence of a `_refresh`, which is the alternative the sentence just rejected. Pick one thesis and propagate it, or revert this hunk — unlike the `syncPipeline.cpp` one in the same commit, this comment was not contradicting the code.

src/wazuh_modules/inventory_sync_server/src/sync/syncQueryBuilder.hpp
```diff
@@ -113,10 +159,13 @@ namespace invsync::sync::querybuilder
...
-                             "if (ctx._source.state == null) { ctx._source.state = [:]; } "
-                             "ctx._source.state.document_version = params.globalVersion; "
-                             "ctx._source.state.modified_at = params.timestamp;";
+        std::string script;
+        script.append(GROUPS_NEEDS_UPDATE)
+            .append(VERSION_NEEDS_UPDATE)
+            .append("if (!needsUpdate) { ctx.op = 'noop'; } else { ")
+            .append("  ctx._source.wazuh.agent.groups = params.groups; ")
+            .append(VERSION_STAMP)
+            .append("}");
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** `GROUPS_NEEDS_UPDATE` sorts both lists before comparing, which is deliberate in `buildGroupsCheckQuery` (disaster recovery, repair only on a real difference) but new for the authoritative writer, which until now wrote `params.groups` unconditionally. Groups going from `["default","web"]` to `["web","default"]` at an unchanged `global_version` now compare equal, `VERSION_NEEDS_UPDATE` is false too, so `ctx.op = 'noop'` and the stale order stays — and the check of the pair will not repair it either, since it is the same comparison. The metadata delta does not have this because its comparison is field-by-field and exact; either give the group delta its own order-sensitive comparison, or say in the file header (`:32-40`) that group order is not reconciled.

src/wazuh_modules/inventory_sync_server/src/sync/syncPipeline.cpp
```diff
@@ -504,6 +514,10 @@ namespace invsync::sync
                     std::vector<Item> just {std::move(item)};
                     respondConnectorFailure(just, connector);
                 }
+
+                // flushAndRespond above returns without draining when the batch is empty, and the
+                // session's own I/O happened after it either way.
+                drainConnectorStats(connector);
                 continue;
```
ignaciogalle12git: **[S4 · non-blocking · introduced]** `syncPipeline.hpp:173-175` states the invariant this misses: "Every path that drives the connector must call it". The `DeleteAgent` branch just above (`:467-493`) drives it — `executeDeleteAgent` stages the delete-by-queries and flushes — and `continue`s without draining, and its preceding `flushAndRespond` returns at `:321-324` without draining when the batch is empty, which is that branch's normal shape; the staging-failure recovery (`:558-566`) has the same gap. That branch is the main producer of `sync.indexer.conflict.retries`, which `metrics.md:118` sells as the only measure of how often the refresh boundary is hit, so the counter can read 0 while conflicts are being waited out.

jotacarma90 (approved):
LGTM

PR #39118, opened 2026-09-10
Find the OS CA bundle on SLES and openSUSE

### LINKED ISSUE #39112: The agent refuses to start on SLES and openSUSE
## Description

On a default install the agent resolves `<agent><ssl><verification_mode>` to `system`, and startup is
gated on finding an OS CA bundle. The candidate list has **no SUSE entry**, so on SLES and openSUSE
`os_find_ca_bundle()` returns `NULL`, `w_agent_validate_ssl_ca()` fails closed and `wazuh-agentd`
exits before it ever opens a socket:

```
2026/09/10 04:32:42 wazuh-agentd: INFO: No port in <agent><manager><endpoint>. Using the default port 1517.
2026/09/10 04:32:42 wazuh-agentd: ERROR: (4121): <ssl><verification_mode> is 'system' but no OS CA bundle was found on this host.
2026/09/10 04:32:42 wazuh-agentd: ERROR: (1215): No client configured. Exiting.
```

These hosts are not misconfigured: SLES and openSUSE do ship a trust store, at
`/etc/ssl/ca-bundle.pem` (a symlink to `/var/lib/ca-certificates/ca-bundle.pem`). Neither path is in
the list, and neither string appears anywhere in this repository. So the agent is failing closed
against a trust store that is present but never looked for.

`wazuh-agentd` never starts, so the service stays `failed` and no module ever runs — not a
degradation, a total loss of the agent on that platform.

## The mechanism

`src/shared/src/os_cert_bundle.c:27-34` probes five fixed paths with a plain `stat()` loop
(`os_find_ca_bundle()`, `:36-47`):

```c
const char* os_ca_bundle_candidates[] = {
    "/etc/ssl/certs/ca-certificates.crt",       // Debian systems
    "/etc/pki/tls/certs/ca-bundle.crt",         // Redhat and Mandriva
    "/usr/share/ssl/certs/ca-bundle.crt",       // RedHat
    "/usr/local/share/certs/ca-root-nss.crt",   // FreeBSD
    "/etc/ssl/cert.pem",                        // OpenBSD, FreeBSD, MacOS
    NULL
};
```

The comment above it says where the list came from:

> *"These values were taken from how libcurl looks for the paths at compilation time, here it is
> modified to be able to support the precompiled deps."* — with a link to curl's `acinclude.m4`.

That is the root of the gap: curl's list is what it probes **at build time on the build host**, and
curl covers SUSE through `--with-ca-bundle` at configure time, not through this list. Copying the
list into a runtime lookup inherits its blind spots.

`src/client-agent/src/config.c:183-188` is the gate:

```c
    if (cfg->ssl.verification_mode == AGENT_VERIFY_SYSTEM) {
        ...
        if (os_find_ca_bundle(NULL) == NULL) {
            merror(AG_SSL_SYSTEM_NO_BUNDLE);
            return false;
        }
```

and its own comment states when it runs: *"Both agentd and the Windows agent gate startup on this,
at the point where each can still fail cleanly: **before daemonizing** on POSIX, before the first
module thread on Windows."*

The same list is used a second time by the HTTPS transport
(`src/client-agent/https_client/src/moduleConfig.cpp`, `validateTls`), and `os_cert_bundle.c` is
compiled directly into that module — so a single fix to the array covers both gates.

## Why it surfaced now

wazuh/wazuh#38786 (*"Enforce TLS verification by default in the agent's HTTPS transport"*, merged)
made an unset `<verification_mode>` resolve to `system` (`src/client-agent/src/config.c:142`).
Before it, the default did not consult the OS trust store, so the missing path was harmless. This is
the same change behind wazuh/wazuh-qa-automation#8846, but a **different failure**: there the agent
reaches the handshake and fails the hostname check, here it never opens a socket.

## Evidence

Logs:

- [full_component_logs (1).zip](https://github.com/user-attachments/files/32055962/full_component_logs.1.zip)
- [logs (1).zip](https://github.com/user-attachments/files/32055968/logs.1.zip)

**Exactly 8 hosts, the whole SUSE family:** `sles-15-amd64`, `sles-15-arm64`, `sles-16-amd64`,
`sles-16-arm64`, `opensuse-15-amd64`, `opensuse-15-arm64`, `opensuse-16-amd64`, `opensuse-16-arm64`.

anromerom:
## Description

Related to #38926

Every SCA check reported as `Not applicable` carried an empty `check.reason`, so an operator could not tell a control that does not apply from one whose evidence could not be read. The reason was collected by the rule evaluators and then discarded before it reached the result, so no scan ever published one. This PR makes the reason reach the agent database and the index, and keeps it tied to the check it actually describes.

Classification is unchanged: an unresolvable rule still yields `Not applicable`, inherited from the 4.x engine. Reporting those checks as distinct from an exemption needs a new result value agreed with the console and the API, and the shipped agent images invoking tools they do not contain is an image concern, so the issue stays open for both.

## Proposed Changes

- Evaluate each rule before reading its unresolved reason, in the policy requirements loop and the checks loop.
- Return the unresolved reason from `FindContentInFile`, whose `Invalid` results had none because it is a free function.
- Clear a rule's unresolved reason at the start of every evaluation, and when a directory scan skips a file it could not evaluate, so an earlier file or scan does not explain a later result.
- Store the reason on any result rather than only `Not applicable`, writing the column unconditionally so it clears once a check becomes measurable.
- Explain a `Not applicable` check that reaches the result with no evaluable rules, which happens when every rule of a check failed to parse.
- Keep one copy of a reason repeated by several rules of the same check.
- Sanitise the stored reason and cap it at 1024 bytes, matching `ignore_above` on `check.reason`, cutting at a line boundary and never inside a UTF-8 sequence.
- Drop an empty reason from both stateful builders and from the stateless `previous` object.

### Results and Evidence

<details>
<summary>Agent database and indexer coverage, before and after</summary>

Ubuntu 24.04 agent, `cis_ubuntu24-04`, 279 checks. Result and reason coverage in the agent database:

```sql
sqlite3 /var/ossec/queue/sca/db/sca.db \
  "select result, count(*), sum(reason is not null and reason != '') as with_reason from sca_check group by result;"
```

```sql
-- before                          -- after
Failed|85|0                        Failed|85|0
Not applicable|120|0               Not applicable|120|120
Passed|74|0                        Passed|74|0
```

A check that previously reported no reason:

```sql
sqlite3 /var/ossec/queue/sca/db/sca.db "select id, result, reason from sca_check where id = '35500';"
```

```sql
35500|Not applicable|Command execution failed: modprobe -n -v cramfs
Command execution failed: lsmod
Path '/etc/modprobe.d/' does not exist
```

Same agent in the indexer, per result:

```sql
GET wazuh-states-sca/_search?size=0
{"aggs":{"r":{"terms":{"field":"check.result"},"aggs":{"with_reason":{"filter":{"exists":{"field":"check.reason"}}}}}}}
```

```sql
Not applicable  docs=120  with_reason=120
Failed          docs=85   with_reason=0
Passed          docs=74   with_reason=0
```

Before the change the same query returned 0 documents with `check.reason` over the whole index.

</details>

### Artifacts Affected

- `libsca.so` (agent SCA module), all platforms. The defect manifests on GCC-built agents.
- `wazuh-states-sca` documents now populate the existing `check.reason`. No mapping change.
- `sca_check.reason` in the agent database holds an empty string instead of NULL when there is nothing to explain, so `reason IS NULL` no longer matches those rows.

### Configuration Changes

N/A

### Documentation Updates

N/A

### Tests Introduced

- `sca_policy_test.cpp` (new): the reason reaches the reported result, one line per unresolved rule in rule order, no reason on a resolved check, and a `Not run` requirement propagating its reason. Three of the four fail without the fix.
- `directory_rule_evaluator_test.cpp`: a scan that skips a file it could not evaluate and then matches a later one reports no reason, and a reason from a previous evaluation of the same rule does not survive. Both fail without the fix.
- `check_condition_evaluator_test.cpp`: repeated reasons kept once, distinct ones kept in order, and a check with no evaluable rules still explaining itself.
- `sca_event_handler_test.cpp`: the reason is stored on a result other than `Not applicable`, capped at 1024 bytes, sanitised when it quotes an undecodable path, and cleared but not published when empty.
- `sca_utils_test.cpp`: reason sanitising over valid multibyte text, invalid path bytes, a truncated sequence, the cut at a line boundary and an oversized single line.
- `file_rule_evaluator_test.cpp` and `sca_recovery_utils_test.cpp`: the file-content rule reports a reason for an unusable pattern; the snapshot builder drops an empty reason and keeps a populated one.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/src/shared/src/os_cert_bundle.c
+++ b/src/shared/src/os_cert_bundle.c
@@ -30,6 +30,7 @@ const char* os_ca_bundle_candidates[] = {
     "/usr/share/ssl/certs/ca-bundle.crt",       // RedHat
     "/usr/local/share/certs/ca-root-nss.crt",   // FreeBSD
     "/etc/ssl/cert.pem",                        // OpenBSD, FreeBSD, MacOS
+    "/etc/ssl/ca-bundle.pem",                   // SUSE (SLES, openSUSE)
     NULL
 };
 
--- a/src/unit_tests/shared/test_os_cert_bundle.c
+++ b/src/unit_tests/shared/test_os_cert_bundle.c
@@ -13,6 +13,8 @@
 #include <setjmp.h>
 #include <cmocka.h>
 #include <stdio.h>
+#include <stdbool.h>
+#include <string.h>
 
 #include "os_cert_bundle.h"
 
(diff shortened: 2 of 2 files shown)


cborla (approved):
LGTM

Issue #39119, opened 2026-09-10
macOS agent monitors no files under /etc because the default configuration points FIM at the /etc symlink


Rebits:
## Description

The macOS agent's default `ossec.conf` declares `<directories>/etc,/usr/bin,/usr/sbin</directories>`. On macOS `/etc` is always a symlink to `private/etc`, and the default configuration does not enable `follow_symbolic_link`. FIM therefore stores a single entry for the symlink `/etc` itself and never inspects its contents, so no file under `/etc` is ever baselined or monitored.

The agent detects and reports this condition itself, at INFO level, on every start:

```
2026/09/10 05:49:08 wazuh-syscheckd: INFO: (6961): Configured path '/etc' is a symbolic link to '/private/etc'. Without 'follow_symbolic_link' enabled, only the symlink itself will be monitored, not the directory contents. Consider monitoring '/private/etc' directly or enabling 'follow_symbolic_link'.
```

The remaining monitored paths in the macOS default configuration (`/usr/bin`, `/usr/sbin`, `/bin`, `/sbin`) all live on the read-only Signed System Volume. With `/etc` silently excluded, the default macOS FIM configuration covers only immutable system files, and no writable location. Security-relevant files such as `sudoers`, `ssh/sshd_config`, `pam.d` and `hosts` all live under `/private/etc` and are unmonitored out of the box.

The same `<directories>/etc,...</directories>` line on a Linux agent of the same build produces full coverage, so the shipped macOS default is the only thing that differs.

## Environment

- Demo environment `wqa-prod-58-demo-environment`, clustered managers (master `demo-env-master-node` + worker `demo-env-worker-node`), single indexer, single dashboard.
- Agent under test: `demo-env-macos-agent`, id `006`, macOS 26.5.1 (build 25F80), arm64, Parallels VM on MacStadium host `macstadium_arm_4`.
- Agent package: `wazuh-agent-5.0.0-latest.arm64.pkg` from `nightly-backup/2026-09-09`, receipt `com.wazuh.pkg.wazuh-agent` version 5.0.0.
- Agent `VERSION.json`: `{"version": "5.0.0", "stage": "rc1", "commit": "2161e69"}`. Manager and indexer report the same version, stage and commit.
- Control agent: `demo-env-linux-deb-amd64-agent`, id `002`, Ubuntu 26.04, `wazuh-agent_5.0.0-latest_amd64.deb` from the same nightly.

## Steps to reproduce

[...]

PR #39124, opened 2026-09-10
Protect real secret paths in the default agent FIM nodiff list (4.14.9)

### LINKED ISSUE #38914: The only `nodiff` entry names a path no distribution has, while `/etc/shadow` is monitored with none
`<nodiff>` marks a file whose contents must never be captured into a change diff. The shipped agent configuration has exactly one, `/etc/ssl/private.key`, which is not a path any Linux distribution creates. Meanwhile `/etc/shadow`, `/etc/gshadow` and their backups are in the monitored baseline and carry no `nodiff` at all.

|---|---|---|---|---|
| 5.0.0-latest | wazuh-agent default `ossec.conf`, `syscheck` `nodiff` | Agent | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 agent, single-node stack |

**Severity:** Medium. The exposure is realised only when an operator enables `report_changes`, which is a normal thing to do and the reason `nodiff` exists. As shipped, the one protection is aimed at nothing and the files that need it have none.

## The entry

```xml
    <!-- Check the file, but never compute the diff -->
    <nodiff>/etc/ssl/private.key</nodiff>
```

On the agent:

```
$ ls -la /etc/ssl/private.key
ls: cannot access '/etc/ssl/private.key': No such file or directory

$ ls -la /etc/ssl/
drwxr-xr-x  2 root root 4096 ... .
-rw-r--r--  1 root root 1167 ... README
lrwxrwxrwx  1 root root   49 ... cert.pem -> /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
lrwxrwxrwx  1 root root   46 ... certs -> /etc/pki/ca-trust/extracted/pem/directory-hash
```

`private.key` is not a file OpenSSL, the distribution packaging or any common service creates. The conventional locations are the directory `/etc/ssl/private/`, and on Red Hat derivatives `/etc/pki/tls/private/`. Neither is named. A `nodiff` for a nonexistent path protects nothing and, because `nodiff` accepts a directory, the fix is a one-word change from a file to a directory.

## What is monitored and unprotected

```
   .pem       296 paths in the line base
   shadow       8 paths: /etc/shadow, /etc/shadow-, /etc/gshadow, /etc/gshadow- (x2 agents)
```

`/etc/shadow` holds every local password hash. It is inside the monitored `/etc` and has no `nodiff`. So does `/etc/gshadow`, and so do the `-` backup copies, which are the ones left behind after a password change and which an operator is least likely to think about.

296 `.pem` files are monitored. On this minimal container they are all CA bundles and public certificates, so nothing sensitive is at risk here. On a host running a web server or a database, the same `/etc/pki` subtree holds the service's private keys, and they would be covered by the same `<directories>/etc</directories>` line with the same absent `nodiff`.

## Why it only bites with `report_changes`

Without `report_changes="yes"` on the `<directories>` element, the agent stores hashes and metadata and never captures file contents, so no diff exists to leak. The shipped configuration has no `report_changes`, which is recorded in [217](BUG_217-the-shipped-agent-checks-file-integrity-twice-a-day-with-no-realtime-and-no-who-data.md) as part of the same attribute-less directory lines.

That makes this a latent defect rather than an active leak, and the sequence that realises it is short and ordinary: an operator turns on `report_changes` to see *what* changed in a configuration file, which is the feature's purpose, and from that moment every change to `/etc/shadow` ships the before and after of the password hash file to the manager and into the diff store on disk.

## Expected behavior

The default `nodiff` list names the paths that actually hold secrets on the platform: `/etc/shadow`, `/etc/gshadow`, their `-` backups, `/etc/ssl/private/`, `/etc/pki/tls/private/`. The entry for a nonexistent file is either corrected to the directory it meant or removed.

## Steps to reproduce

1. `docker exec wazuh-wazuh.agent-1 grep nodiff /var/ossec/etc/ossec.conf`. One entry, `/etc/ssl/private.key`.
2. `docker exec wazuh-wazuh.agent-1 ls -la /etc/ssl/private.key`. No such file.
3. `POST /wazuh-states-fim-files/_search` with `{"query":{"wildcard":{"file.path":"*shadow*"}}}`. Eight rows across the two agents, including `/etc/shadow` and `/etc/shadow-`.
4. Confirm no `nodiff` covers them.

## Related

[217](BUG_217-the-shipped-agent-checks-file-integrity-twice-a-day-with-no-realtime-and-no-who-data.md) and [218](BUG_218-the-default-fim-directory-list-omits-root-home-the-cron-spool-and-the-agents-own-configuration.md) are the other two defects in the same `syscheck` block.

cc @wazuh/devel-xdrsiem-agent

Darioortegaleyva:
## Description

This PR fixes the shipped `syscheck` default `nodiff` list: the single entry (`/etc/ssl/private.key`) names a path that doesn't exist on any distribution, so it protects nothing, while `/etc/shadow`, `/etc/gshadow` and their `-` backups are monitored under the default `<directories>/etc</directories>` line with no `nodiff` protection at all. If an operator enables `report_changes` on that directory, the password hash files' contents would be captured into the diff.

The original report (#38914) covers the agent's default config; a sweep after the initial fix found the identical defect in the shipped **manager** defaults too (the manager also runs syscheckd on itself), so this PR fixes both.

**Proposed Release:** v4.14.9
**Issue:** wazuh/wazuh#38914
**Internal Reference:** N/A

## Proposed Changes

- Updated `etc/ossec-agent.conf` (shipped default agent config).
- Updated `etc/templates/config/generic/syscheck.agent.template` (covers mainstream Linux distros on 4.14.9, which has no per-distro breakdown yet).
- Updated `etc/templates/config/aix/syscheck.agent.template` and `etc/templates/config/darwin/syscheck.agent.template` (ssl path fix only, no `pki/tls` alternative and no `/etc/shadow`/`/etc/gshadow` entries, since neither platform ships those paths). `darwin`'s agent template targets `/private/etc/ssl/private/`, not `/etc/ssl/private/`, to match `#39119` (the `/etc` -> `/private/etc` symlink fix), which is already merged into `4.14.9`; `aix` and `darwin`'s **manager** template stay on plain `/etc`, since neither the AIX fix nor `#39119` cover them.
- Updated `etc/ossec.conf`, `etc/ossec-local.conf`, `etc/ossec-server.conf` (shipped default manager/standalone configs) and `etc/templates/config/generic/syscheck.manager.template` (found via sweep - same defect on the manager side, see below).
- Updated `etc/templates/config/darwin/syscheck.manager.template` (ssl path fix only, no `pki/tls` alternative and no shadow/gshadow - see sweep notes on reachability below).
- Replaced the single dead `nodiff` entry with `<nodiff type="sregex">^/etc/ssl/private/|^/etc/pki/tls/private/</nodiff>` plus plain `<nodiff>` entries for `/etc/shadow`, `/etc/shadow-`, `/etc/gshadow`, `/etc/gshadow-` (agent and manager Linux configs). AIX/darwin agent and darwin manager templates get only the sregex ssl/pki entry, since none of those platforms ship `/etc/shadow`.

### Results and Evidence

Validated in two ways: a live agent container (same images as the original report) for the config-level bug, and a standalone test compiled and linked against this repo's real `os_regex` object files, to check the actual matching semantics `is_file_nodiff()` uses (`src/syscheckd/src/fim_diff_changes.c:812-830`) before and after the fix.

**Live agent (`public.ecr.aws/wazuh-cicd/wazuh/wazuh-agent:5.0.0-latest`, same bug/config shape as `4.14.9`):**

```sql
$ docker exec repro38914-agent grep nodiff /var/ossec/etc/ossec.conf
    <nodiff>/etc/ssl/private.key</nodiff>

$ docker exec repro38914-agent ls -la /etc/ssl/private.key
ls: cannot access '/etc/ssl/private.key': No such file or directory

$ docker exec repro38914-agent ls -la /etc/shadow /etc/shadow- /etc/gshadow /etc/gshadow-
---------- 1 root root 365 Sep  9 03:16 /etc/shadow
---------- 1 root root 344 Aug 12 03:07 /etc/shadow-
---------- 1 root root 208 Sep  9 03:16 /etc/gshadow
---------- 1 root root 198 Jan 30  2023 /etc/gshadow-
```

Confirmed the same defect exists verbatim on the real `4.14.9` branch tree before this fix:

```sql
$ git show origin/4.14.9:etc/ossec-agent.conf | grep nodiff
    <nodiff>/etc/ssl/private.key</nodiff>
$ git show origin/4.14.9:etc/templates/config/generic/syscheck.agent.template | grep nodiff
    <nodiff>/etc/ssl/private.key</nodiff>
```

**Real code, before/after (`is_file_nodiff()` reproduced verbatim, linked against this repo's actual `os_match_compile.o`/`os_match_execute.o`/`os_match.o`/`os_match_free_pattern.o`/`os_regex_str.o`/`os_regex_maps.o`, byte-identical between `main` and `4.14.9`):**

```sql
=== BEFORE (current shipped default: <nodiff>/etc/ssl/private.key</nodiff>) ===
/etc/ssl/private/server.key            nodiff=0  <- LEAKS INTO DIFF
/etc/pki/tls/private/localhost.key     nodiff=0  <- LEAKS INTO DIFF
/etc/shadow                            nodiff=0  <- LEAKS INTO DIFF
/etc/shadow-                           nodiff=0  <- LEAKS INTO DIFF
/etc/gshadow                           nodiff=0  <- LEAKS INTO DIFF
/etc/gshadow-                          nodiff=0  <- LEAKS INTO DIFF
/etc/ssl/certs/ca-bundle.crt           nodiff=0  <- LEAKS INTO DIFF

=== AFTER (fix: sregex for ssl/pki dirs + exact shadow/gshadow entries) ===
/etc/ssl/private/server.key            nodiff=1  protected
/etc/pki/tls/private/localhost.key     nodiff=1  protected
/etc/shadow                            nodiff=1  protected
/etc/shadow-                           nodiff=1  protected
/etc/gshadow                           nodiff=1  protected
/etc/gshadow-                          nodiff=1  protected
/etc/ssl/certs/ca-bundle.crt           nodiff=0  (not a secret path, stays diffable, correct)
```

Also confirmed that a plain (non-`sregex`) `<nodiff>/etc/ssl/private/</nodiff>` directory entry would silently match nothing (`is_file_nodiff()` uses exact `strcmp` for plain entries, only `type="sregex"` entries get prefix matching via `OSMatch_Execute`/`_os_strncmp`, `src/shared/os_regex/os_match_compile.c:159-163`) - the same failure mode as the original bug, just invisible. The fix uses `type="sregex"` with a `^` prefix, the same mechanism already used by the shipped `<ignore type="sregex">.log$|.swp$</ignore>` entry in these same templates.

All 9 modified files were validated as well-formed XML.

**Manager-side sweep (post-fix self-review):** the same `<nodiff>/etc/ssl/private.key</nodiff>` line and unprotected `/etc/shadow`/`/etc/gshadow` were also found in the shipped manager defaults. Confirmed as real, shipped duplicates, not dead code:

- `etc/ossec-server.conf` - shipped default for `INSTYPE=server` installs (`src/init/inst-functions.sh:758`).
- `etc/ossec-local.conf` - shipped default for standalone/local installs (`src/init/inst-functions.sh:764`).
- `etc/ossec.conf` - the repo's checked-in reference default, the same install target the two above get copied to.
- `etc/templates/config/generic/syscheck.manager.template` - resolved by `GetTemplate()` (`src/init/template-select.sh`) for manager config generation, the fallback for every distro (4.14.9 only has `generic`/`darwin` manager templates).

All four have byte-identical surrounding structure (same `<ignore>` list, same `<skip_*>` block) to the agent files, and no `report_changes` set either - same latent-bug shape.

Also fixed `etc/templates/config/darwin/syscheck.manager.template`, flagged as **likely unreachable**: Wazuh has never shipped a macOS manager, and `src/Makefile` has no server/manager build wiring under any `Darwin` conditional. No hard in-repo gate proves it's dead (packaging lives outside this repo), so fixed it anyway since the change is free.

### Artifacts Affected

- `etc/ossec-agent.conf`
- `etc/ossec.conf`
- `etc/ossec-local.conf`
- `etc/ossec-server.conf`
- `etc/templates/config/generic/syscheck.agent.template`
- `etc/templates/config/generic/syscheck.manager.template`
- `etc/templates/config/aix/syscheck.agent.template`
- `etc/templates/config/darwin/syscheck.agent.template`
- `etc/templates/config/darwin/syscheck.manager.template`

### Configuration Changes

Changes the agent's shipped default `syscheck` configuration (`nodiff` list). No operator action required; existing custom `ossec.conf` overrides are unaffected.

### Documentation Updates

N/A

### Tests Introduced

- **Scope:** Manual config validation + real-code unit-level check.
- **Details:** XML well-formedness check on all 4 modified files. Live-agent container repro of the original bug (secret paths unprotected). Standalone C test linked against this repo's real `os_regex` object files, reproducing `is_file_nodiff()` verbatim, verifying the new `nodiff` entries match the intended secret paths and nothing else, before vs. after.

## Review Checklist

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux

  - [ ] Coverity
  - [ ] AddressSanitizer

**General Checklist:**

- [ ] Code changes reviewed


etc/templates/config/darwin/syscheck.agent.template
```diff
@@ -29,7 +29,7 @@
     <ignore type="sregex">.log$|.swp$</ignore>
 
     <!-- Check the file, but never compute the diff -->
-    <nodiff>/etc/ssl/private.key</nodiff>
+    <nodiff type="sregex">^/etc/ssl/private/</nodiff>
```
MarcelKemp: **[S1 · blocking · introduced]** This PR's branch predates a already-merged fix on `4.14.9` (see CHANGELOG entry for issue #39119) that repointed this file's `<directories>`, `<ignore>` and `<nodiff>` entries from `/etc/...` to `/private/etc/...` (macOS resolves `/etc` as a symlink to `/private/etc`; without `follow_symbolic_link`, syscheck only recorded the symlink). `git merge-tree origin/4.14.9 HEAD` confirms a real content conflict here, and `gh api repos/wazuh/wazuh/pulls/39124` reports `mergeable: false`, `mergeable_state: dirty`. As written, `^/etc/ssl/private/` will never match the real path syscheck evaluates on the current default (`/private/etc/ssl/...`), so this PR doesn't protect the real secret path on macOS even after any naive conflict resolution that keeps the PR's line.
```suggestion
    <nodiff type="sregex">^/private/etc/ssl/private/</nodiff>
```

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ All notable changes to this project will be documented in this file.
 - Fixed the gcloud wodle masking missing-dependency errors as an unrelated `AttributeError`. ([#38856](https://github.com/wazuh/wazuh/pull/38856))
 - Fixed process names truncated to fifteen characters in the syscollector inventory. ([#38969](https://github.com/wazuh/wazuh/pull/38969))
 - Fixed the gcloud wodle's Pub/Sub integration failing to start whenever the bucket integration's dependencies (e.g. `google-cloud-storage`) were broken, by deferring each integration's imports so a failure in one no longer blocks the other. ([#38868](https://github.com/wazuh/wazuh/pull/38868))
+- Fixed the default agent nodiff list not protecting /etc/shadow and real key paths. ([#39124](https://github.com/wazuh/wazuh/pull/39124))
```
MarcelKemp: **[S4 · non-blocking · introduced]** Same staleness as the finding above: `git merge-tree origin/4.14.9 HEAD` also flags a real content conflict here (two PRs added different lines at the same context line). No functional impact, just needs a rebase alongside the darwin template fix.

etc/templates/config/aix/syscheck.agent.template
```diff
@@ -29,7 +29,7 @@
     <ignore type="sregex">.log$|.swp$</ignore>
 
     <!-- Check the file, but never compute the diff -->
-    <nodiff>/etc/ssl/private.key</nodiff>
+    <nodiff type="sregex">^/etc/ssl/private/</nodiff>
```
MarcelKemp: **[S4 · non-blocking · introduced]** Nit: the PR description says aix/darwin get an "ssl/pki path fix", but the diff here (and in the darwin agent template) only adds the `ssl` alternative, not `pki`. That's actually correct — neither platform has `/etc/pki` — just a loose description worth tightening.

MarcelKemp (commented):
Can you check whether any OS configurations are missing from `etc/templates/config/`, where the relevant `syscheck.agent.template` files may need to be modified?

etc/templates/config/darwin/syscheck.agent.template
```diff
@@ -30,7 +30,7 @@
     <ignore type="sregex">.log$|.swp$</ignore>
 
     <!-- Check the file, but never compute the diff -->
-    <nodiff>/private/etc/ssl/private.key</nodiff>
+    <nodiff type="sregex">^/private/etc/ssl/private/</nodiff>
```
MarcelKemp: **[S3 · non-blocking · introduced]** Since #39119 put `/private/etc` under real monitoring, macOS's account file `/private/etc/master.passwd` — the platform analogue of the `/etc/shadow` family this PR protects on Linux — is inside the monitored tree with no `nodiff`, so enabling `report_changes` there would ship its contents in a diff on the next scan after any local account change. Lower severity than the Linux case because modern macOS keeps the real hashes in OpenDirectory plists under `/var/db/dslocal` and leaves `master.passwd`'s password field as `*`, so this is root-only account metadata rather than authentication material — but it is the same class of file, and the description's "neither platform ships those paths" reasoning misses it.
```suggestion
    <nodiff type="sregex">^/private/etc/ssl/private/</nodiff>
    <nodiff>/private/etc/master.passwd</nodiff>
```

etc/templates/config/darwin/syscheck.manager.template
```diff
@@ -37,7 +37,7 @@
     <ignore type="sregex">.log$|.swp$</ignore>
 
     <!-- Check the file, but never compute the diff -->
-    <nodiff>/etc/ssl/private.key</nodiff>
+    <nodiff type="sregex">^/etc/ssl/private/</nodiff>
```
MarcelKemp: **[S3 · non-blocking · introduced]** On macOS `/etc` is a symlink to `/private/etc`, and this file still monitors `<directories>/etc,/usr/bin,/usr/sbin</directories>` (line 17) — it never received the #39119 fix that the agent template did — so without `follow_symbolic_link` syscheck only records the symlink and `is_file_nodiff()` is never handed a `/etc/ssl/...` path, making this new entry as inert as the `/etc/ssl/private.key` it replaces. Not a regression and no macOS manager package ships, but a source install with `TARGET=server`/`local` uses this template, and the two darwin templates now disagree with each other. Either extend the #39119 fix here (`<directories>`, `<ignore>` and this line all repointed at `/private/etc`) or open a follow-up issue.
```suggestion
    <nodiff type="sregex">^/private/etc/ssl/private/</nodiff>
```

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ All notable changes to this project will be documented in this file.
 - Fixed process names truncated to fifteen characters in the syscollector inventory. ([#38969](https://github.com/wazuh/wazuh/pull/38969))
 - Fixed the gcloud wodle's Pub/Sub integration failing to start whenever the bucket integration's dependencies (e.g. `google-cloud-storage`) were broken, by deferring each integration's imports so a failure in one no longer blocks the other. ([#38868](https://github.com/wazuh/wazuh/pull/38868))
 - Fixed the macOS agent's default FIM configuration monitoring `/etc`, which macOS resolves as a symlink to `/private/etc`; without `follow_symbolic_link` enabled, syscheck only recorded the symlink itself, leaving every file under it (`sudoers`, `sshd_config`, `pam.d`, `hosts`) uncovered. The default `<directories>`, `<ignore>`, and `<nodiff>` entries now target `/private/etc` directly. ([#39119](https://github.com/wazuh/wazuh/issues/39119))
+- Fixed the default agent nodiff list not protecting /etc/shadow and real key paths. ([#39124](https://github.com/wazuh/wazuh/pull/39124))
```
MarcelKemp: **[S4 · non-blocking · introduced]** This says "default agent nodiff list", but the PR also fixes the manager defaults (`etc/ossec.conf`, `etc/ossec-local.conf`, `etc/ossec-server.conf`, and the generic/darwin manager templates), so anyone reading the changelog to decide whether their manager is affected will conclude it is not. The paths are also unquoted, unlike the neighbouring entries.
```suggestion
- Fixed the default agent and manager `nodiff` lists not protecting `/etc/shadow`, `/etc/gshadow` and the real private-key directories. ([#39124](https://github.com/wazuh/wazuh/pull/39124))
```

etc/templates/config/aix/syscheck.agent.template
```diff
@@ -29,7 +29,7 @@
...
```
MarcelKemp: **[S4 · non-blocking · introduced]** Still open from iteration 1: the description's last bullet says the aix/darwin templates "get only the sregex ssl/pki entry", but this line (and the darwin ones) add the `ssl` alternative only. That is the correct behaviour — neither platform has `/etc/pki` — so this is just the wording. While you are there, the description also still predates the rebase and does not mention that the darwin agent template now targets `/private/etc`.

MarcelKemp (approved):

Issue #39070, opened 2026-09-09
Deliver the manager CA to 4.x agents over the WPK transfer channel during upgrade


TomasTurina:
## Description

A 5.0 manager is always a fresh install, so a fleet of 4.x agents reaches 5.0 through a remote WPK upgrade. Once upgraded, those agents run the 5.0 HTTPS agent protocol on 1517 but hold no trust anchor on disk, so they cannot verify the manager. They also cannot enrol again to obtain one: they already hold a `client.keys` identity from their previous enrolment, so they never see an enrollment token, and re-enrolling would cost agent-identity continuity.

This matters more now that fresh 5.0 installs ship with verification set to `full` and take their anchor from the enrollment token. The in-place upgrade is the one remaining path by which an agent can legitimately end up on 5.0 with no anchor at all. If we do not solve it here, every migrated fleet stays unverified indefinitely while every newly built host is verified.

The routes currently available to deliver a CA to an already-enrolled agent are a configuration-management push of the anchor file or full re-enrolment. Shared configuration cannot carry it, because `Read_Agent_Shared()` accepts only `<batch>` and an `<ssl>` block in a group's `agent.conf` is a hard parse error that rejects the whole file. An operator-built WPK cannot carry it either, because agents verify package signatures against `wpk_root.pem`. Estates with no configuration-management tool are left visiting endpoints by hand.

There is a channel that already works. The manager drives the 4.x upgrade with a sequence of `com` module commands over the 1514 agent channel — `open`, repeated `write`, `close`, `sha1`, then `upgrade` — and that channel is encrypted and integrity-protected with the agent's own key from `client.keys`. An attacker without that key cannot inject a message the agent will accept. The `com` file-write primitive is not specific to WPK payloads.

This issue covers the manager side: send the CA certificate to the agent over that same channel, before issuing the upgrade command. Agent-side handling of the delivered file is tracked separately.

The security property is worth stating explicitly, because it is stronger than the alternatives considered: the 4.x symmetric-key channel is used to bootstrap the 5.0 asymmetric trust anchor. There is no unauthenticated moment and no trust-on-first-use window, so nothing here needs defending to a certification assessor.

### Destination directory

[...]

PR #39125, opened 2026-09-10
Agent: detect, validate, and install a manager-delivered CA during WPK upgrade

### LINKED ISSUE #39071: Install the delivered manager CA during WPK upgrade and enable certificate verification
## Description

Companion to the manager-side CA delivery issue. The manager now streams its root CA into `var/incoming` over the authenticated 4.x channel before issuing the upgrade command. This issue covers the agent side: the 5.0 WPK's upgrade scripts pick that file up and install it as the agent's trust anchor, so the agent verifies the manager on its first post-upgrade connection.

This closes the migration gap without a new enrolment. The agent keeps the `client.keys` identity it already has, which is what authenticates it on every request, and gains the anchor it was missing. No fleet-wide enrolment password is involved and agent-identity continuity is preserved.

The scripts live inside the Wazuh-signed 5.0 WPK, so they are ours to write and are covered by the existing `wpk_root.pem` signature chain. They also run as root during package installation, which satisfies the requirement that the anchor be root-owned and not writable by the `wazuh` user the agent runs as — the daemon cannot create a file it must not be able to write, so it has to be placed before privileges are dropped.

### Why the scripts should not write `ossec.conf`

Fresh 5.0 installs ship a config that sets verification to `full`, and take their anchor from the enrollment token. An upgraded agent is a different case: it keeps its existing 4.x `ossec.conf`, which has no `<ssl>` block at all, and it never sees a token.

The scripts should therefore install the anchor file and leave `ossec.conf` alone. A usable anchor on disk is what puts the agent into `full`, regardless of what the config says, so placing the file is the entire cutover. Writing an `<ssl>` block instead would put the agent's trust configuration in a file that configuration management templates, and because the agent fails closed, the next converge that reverts it is a fleet outage caused by a routine automation run.

This depends on anchor-driven verification being present in the target build. If it is not, the scripts have no choice but to write the `<ssl>` block, with the revert risk above. Confirm before implementation and record the decision on this issue.

### Where the CA lives

`com` write operations are jailed to `INCOMING_DIR` (`var/incoming`), so the CA arrives there, not in `var/upgrade`. This is not interchangeable: `com upgrade` calls `cldir_ex(UPGRADE_DIR)`, which clears `var/upgrade` before unpacking, so a CA staged there would be deleted before the script could read it. The scripts must read from `var/incoming` and are responsible for removing the file afterwards.

### Outage risk to check first

Because the packaged 5.0 config now requests `full`, an upgrade that replaces the existing `ossec.conf` with the packaged default would leave an upgraded agent demanding verification with no anchor — which means it refuses to start. On a fleet-wide upgrade that is a mass outage. Confirming that the upgrade preserves the existing config on every platform is a prerequisite for this work, not a detail.

**Scope:** `upgrade.sh` and `upgrade.bat` inside the 5.0 WPK. No 4.x agent code changes.

### Prerequisite: config preservation

- [ ] Confirm the deb, rpm, msi and pkg upgrade paths preserve the existing `ossec.conf` and do not apply the packaged `full` default over it.
- [ ] Confirm no platform leaves a `.rpmnew` / `.dpkg-dist` style artefact that a later operator action or automation could promote into place unnoticed.
- [ ] If any path does replace the config, resolve that before the rest of this issue proceeds.

### Detect and validate

- [ ] Look for the reserved CA filename in `var/incoming` at the start of the upgrade script. Do not look in `var/upgrade` — `cldir_ex(UPGRADE_DIR)` has already cleared it by this point.
- [ ] Treat absence as a normal condition, not an error: log it and continue with an unverified upgrade. An agent upgraded by an older manager, by a manager with `<remote><legacy><ca_delivery>` disabled, or during a partial rollout will legitimately have no CA present.
- [ ] Validate the file parses as a PEM certificate before installing it.
- [ ] Validate it is a CA certificate (`basicConstraints` `CA:TRUE`) and is within its validity window.
- [ ] Refuse to install an unusable file and continue unverified, rather than installing something that will make the agent fail closed after upgrade.

### Install the anchor

- [ ] Write the CA to the agent-owned anchor path, outside `ossec.conf`.
- [ ] Set ownership and permissions so the file is root-owned and not writable by the `wazuh` user, following the pattern already used for `client.keys`.
- [ ] On Windows, apply the equivalent ACLs. Confirm the service account cannot modify the anchor.
- [ ] Make installation idempotent, so a re-run or a repeated upgrade does not corrupt or duplicate the anchor.
- [ ] Remove the CA from `var/incoming` after installation, on both the success and the rejected-file paths, so a stale or malformed certificate is not left behind for a later upgrade to pick up.

### Do not override deliberate operator configuration

- [ ] If the agent already holds a usable anchor, replace it with the delivered CA and log the replacement at a visible level. The manager is authoritative for its own CA, and a stale anchor takes the agent off the air.
- [ ] If the existing `ossec.conf` explicitly requests `certificate` or `system` mode, honour it. A host deliberately pointed at a corporate PKI or the OS trust store must not be silently moved onto the manager's CA.

### Verification state

- [ ] Do not edit `ossec.conf`. Installing the anchor is the cutover.
- [ ] Confirm against the target build that a usable anchor puts the agent into `full` on its first socket even though the upgraded config requests nothing. If that is not the case, this section becomes an `<ssl>` block edit and the revert risk must be documented.
- [ ] Do not require a second operator step anywhere in the flow.

### Recovery

- [ ] Document how to recover an agent that cannot connect after upgrade because verification is failing — removing or replacing the anchor, and where to look in the log.
- [ ] Ensure the failure is diagnosable from the agent log alone: an agent that cannot validate the manager's certificate must say which check failed, not just that the connection dropped.

### Logging

- [ ] Log the CA installation as a distinct, greppable step.
- [ ] Log at startup whether verification is active and against which anchor, so a fleet-wide state check does not require inspecting files.
- [ ] When no CA arrived, log a clear warning that the agent is running unverified and how to fix it. This is the one remaining path to an unverified 5.0 agent, so it must be obvious rather than invisible. The agent log is the only place this surfaces — agent upgrade tasks have no result payload, so the manager side is log-only too.

## Definition of Done

- [ ] The scripts read the CA from `var/incoming` and leave nothing behind there afterwards.
- [ ] A 4.x agent upgraded to 5.0 with a CA delivered installs the anchor and connects on 1517 with certificate verification active, on its first post-upgrade connection, with no operator action and no new enrolment.
- [ ] The agent retains its pre-upgrade agent ID and `client.keys` identity. Verified by comparing the manager's agent record before and after.
- [ ] A 4.x agent upgraded with no CA delivered upgrades successfully, starts, connects unverified, and logs why. It does not refuse to start — unlike a fresh install, its config does not request verification.
- [ ] The existing `ossec.conf` is preserved across upgrade on every platform, and the packaged `full` default is not applied over it.
- [ ] `ossec.conf` is not modified by the upgrade scripts, or if the build requires it, the edit and its revert risk are documented.
- [ ] A malformed, expired, non-CA or truncated CA file does not break the upgrade and does not leave the agent unable to connect.
- [ ] A host with `certificate` or `system` explicitly configured is not moved onto the delivered CA.
- [ ] The anchor is root-owned and not writable by the `wazuh` user on Linux and macOS, with equivalent ACLs on Windows. Verified by attempting a write as the service account.
- [ ] Works on deb, rpm, msi and pkg upgrade paths.
- [ ] Re-running the upgrade is idempotent and leaves a valid anchor.
- [ ] Negative test: point the agent at a manager presenting a certificate not signed by the installed CA, and assert the agent refuses the connection and logs which check failed. Demonstrating refusal of a bad certificate is stronger evidence than demonstrating acceptance of a good one.
- [ ] Documentation covers the post-upgrade verified state, how it differs from a fresh install, how to check verification fleet-wide, and how to recover an agent that cannot connect.

## Related

- Manager-side CA delivery: https://github.com/wazuh/wazuh/issues/39070
- Blocked on the anchor-driven verification decision noted above.

rjcausarano:
## Description

Agent-side companion to #39070 (which streams the manager's root CA into the incoming-transfer directory over the `com` channel ahead of a WPK upgrade). This PR adds the detection, validation, and on-disk installation of that CA across all three platforms: `src/init/pkg_installer.sh` (Linux/macOS) and `src/win32/do_upgrade.ps1` (Windows).

It deliberately does **not** wire the installed CA into `<ssl><certificate_authorities>` in `ossec.conf`, on any platform — see "Configuration Changes" below for why. That does **not** mean verification stays inactive: `src/client-agent/src/config.c`'s `w_agent_resolve_ssl_posture()` (already on this branch via the base rebase, not introduced by this PR) picks up a present, readable anchor at `AGENT_ANCHOR_CA` automatically at startup and resolves an otherwise-unset `<verification_mode>` to `full` against it — so installing the anchor is, by itself, sufficient to activate certificate verification end-to-end, with no `ossec.conf` edit required. **This has now been confirmed with a real test, not just by reading the code** — including the negative case (manager cert not signed by the installed CA is refused, with a specific, diagnosable reason in the log) — see "Results and Evidence".

Opening this without `Closes #39071` for now regardless: a few Definition-of-Done items are still outstanding — documentation, confirming the `ossec.conf`-preservation prerequisite across deb/rpm/msi/pkg, and exercising the rpm/apk/msi/pkg upgrade paths at all (only deb has been run end-to-end so far); related to #39071.

Also fixes an unrelated double-restart race in the WPK upgrade orchestration on Linux, discovered while testing this PR's own changes end-to-end on a live agent (see "Results and Evidence").

## Proposed Changes

**CA detection, validation, and install (Linux, macOS, Windows):**

- Detect a CA delivered by the manager at the reserved incoming-transfer path (`var/incoming/root-ca.pem` on Linux/macOS, `incoming\root-ca.pem` on Windows — `INCOMING_DIR` per `src/shared/include/defs.h`), checked at the very start of the upgrade script — never the upgrade-staging directory, since `com upgrade`'s `cldir_ex(UPGRADE_DIR)` clears that directory before this script runs.
- Validate the delivered file before trusting it: parses as a PEM certificate, carries `X509v3 Basic Constraints: CA:TRUE`, and is within its `notBefore`/`notAfter` validity window. A malformed, expired, or non-CA file is rejected and logged; it never aborts the upgrade. On Linux/macOS this uses `openssl x509`; on Windows, a .NET `X509Certificate2`/`X509BasicConstraintsExtension` parse (no external `certutil`/`openssl` dependency).
- On a valid CA, install it to the agent's trust-anchor path (`/var/ossec/etc/certs/root-ca.pem` on Linux/macOS, `<install_dir>\certs\root-ca.pem` on Windows — `AGENT_ANCHOR_CA`), restricted to the platform's equivalent of "root-owned, not writable by the agent's own user": `root:wazuh` / mode `640` on Linux/macOS (same pattern as the existing `wpk_root.pem` trust anchor); on Windows, `icacls` stripping `Authenticated Users` (`*S-1-5-11`) access from the file, the same pattern already used for `client.keys`/`authd.pass` in `InstallerScripts.vbs`.
- Remove the file from the incoming directory after processing in every case (installed or rejected), so a stale or malformed certificate is never left behind for a later upgrade attempt to pick up.
- `ossec.conf` is never modified by either script. Removed `pin_ca()` and its only helper, `xml_block_present()`, as dead code now that nothing writes to the config — on both the Linux/macOS (`pkg_installer.sh`) and Windows (`do_upgrade.ps1`) sides.
- Added a narrow exception to the pre-install "does the OS trust store verify the manager's certificate" gate: for a currently-installed pre-5.0 agent, a failed verification no longer aborts the upgrade. On Linux/macOS this is detected via `dpkg-query`/`rpm` before the new package replaces it; on Windows, via the already-captured `$current_version` (read from `VERSION.json` before the MSI overwrites it). Rationale: a genuine 4.x `ossec.conf` is `<client>`-only and can never carry `<agent><ssl>` — `Read_Legacy_Client_Address()` (`config.c`) never reads `<ssl>` under `<client>` — so this class of agent cannot satisfy that check regardless of what this script does. It's also not needed for safety: the real daemon's own fail-closed condition for `system` mode (`w_agent_validate_ssl_ca()` in `config.c`) only triggers when no OS CA bundle exists at all, not when that bundle simply fails to verify this specific manager. An already-5.x agent gets no such pass and still hard-aborts on this check, since it has had every opportunity to be configured correctly.
- Two logging gaps closed on both platforms: (1) when no CA is delivered this run *and* no anchor is present at all from a previous one, the script now logs an explicit warning that the upgraded agent will run unverified and how to fix it (place the CA and re-run, or configure `<certificate_authorities>` explicitly) — previously this case fell through silently; (2) installing over an already-present anchor now logs "Replaced the existing..." instead of the same "Installed the delivered..." line a fresh install gets, so an operator can grep for the more consequential event distinctly. Also corrected the install-success log line itself, which previously (incorrectly) said verification wasn't active from the anchor alone — see the `config.c` finding above.

**macOS-specific fix:**

- The validity-window check originally used GNU-only `date -d`, which BSD `date` (macOS) does not support. On Darwin this silently rejected every valid CA as having an "unparsable validity period" (the parse failure was swallowed by `2>/dev/null`, so the script kept running, just never installing anything). Fixed by branching on the already-set `$OS`: `date -d` on Linux, `date -j -f "%b %e %T %Y %Z"` (BSD `strptime` syntax) on Darwin. The format string was verified against real `openssl -startdate`/`-enddate` output (`Sep 11 14:55:53 2026 GMT`-shaped).
- `chown root:wazuh` and the `wazuh` group itself were already correct on macOS before this PR (BSD `chown` accepts the same `owner:group` syntax as GNU, and `packages/macos/package_files/preinstall.sh` already creates a `wazuh`-named group via `dscl` before any upgrade-driven run of this script) — no change needed there.

**Unrelated fix, found while testing — double restart during Linux WPK upgrade:**

- Root-caused via a real failed upgrade on a Linux test agent: `upgrade.log` showed the post-install wait-for-connection loop spinning for its full 30 attempts with `var/run/wazuh-agentd.state: No such file or directory` every time, ending in "Upgrade failed...". Cross-referencing `journalctl -u wazuh-agent.service` showed why: the deb/rpm package's own maintainer script (`postinst`/`%post`) already restarts the agent via `systemctl restart wazuh-agent.service` when it finds the `wazuh.restart` marker `preinst`/`%pre` drops — and `pkg_installer.sh` then unconditionally called `./bin/wazuh-control restart` again right after, bypassing systemd and killing the daemon set systemd was still supervising. Once that direct kill emptied the unit's cgroup, systemd marked `wazuh-agent.service` "Deactivated successfully" — and since nothing told systemd to bring it back up, the agent just stayed down for the rest of the wait loop.
- Fixed by making the restart conditional: if `wazuh-agentd` is already reported running by `wazuh-control status` (i.e., the package's own maintainer script already restarted it), skip the redundant restart instead of racing it. Falls back to the original explicit `wazuh-control restart` only when the agent isn't already up (e.g., the `./var/upgrade/install.sh` source-install fallback path, which has no maintainer script to do it automatically). macOS's `launchctl bootstrap`-based restart has the same two-trigger shape (`postinstall.sh` and `pkg_installer.sh`'s own Darwin branch both call it) but was left unchanged — `launchctl bootstrap` on an already-loaded job errors harmlessly and is already swallowed by the existing `|| true`, unlike systemd's hard deactivation, and this was not independently reproduced as an actual failure on macOS.

## Results and Evidence

- Validation logic (Linux/macOS) exercised manually against a real self-signed CA, a non-CA leaf certificate, and a garbage file (PEM parse via `openssl x509 -noout`, `X509v3 Basic Constraints` grep, and `notBefore`/`notAfter` epoch comparison) — all three behaved as expected.
- `pin_ca()`'s removal and the earlier config-insertion logic it replaced were exercised in a sandbox against a copy of a real unmigrated 4.x `ossec.conf` before being removed, confirming the config-shape problem this PR sidesteps rather than assuming it.
- Full end-to-end upgrade reproduced on a two-VM Linux lab pair (debian12 agent, 4.14.7 → 5.0.0; 5.0.0 HTTPS manager): the manager's CA was placed at the incoming path, detected and validated by the script, installed to the anchor path with the expected ownership, and the package upgrade completed. The legacy-agent exception let the upgrade proceed past the untrusted-manager-certificate check instead of aborting, as intended.
- **The double-restart bug above was found on that same lab pair**: the first end-to-end run failed at the post-install wait-for-connection step, root-caused via `upgrade.log` and `journalctl -u wazuh-agent.service` as described above. The fix has been applied and syntax-checked (`bash -n`), but **has not yet been re-verified with a fresh upgrade run on the VM** — that re-run is still pending.
- **Anchor-driven verification and the negative test, both confirmed with a real run on the same debian12 agent** (already-upgraded to 5.0.0, anchor installed by the earlier upgrade run above, `ossec.conf` carrying no `<ssl>` block at all): manually confirmed `openssl s_client` verifies the manager's cert against the installed anchor (`Verify return code: 0 (ok)`); the agent was already connected under that anchor (`status='connected'`). Then, for the negative case: swapped the installed anchor for an unrelated throwaway CA, restarted the agent, and confirmed it refuses to connect — `wazuh-agentd.state` moved to `status='disconnected'`, and `ossec.log` logged a specific, diagnosable reason: `WARNING: Manager unreachable... SSL certificate OpenSSL verify result: unable to get local issuer certificate (20)`. Restored the correct anchor, restarted again, confirmed clean reconnection (`status='connected'`). This is real confirmation, not just a reading of `config.c` — both that anchor presence alone activates `full` verification, and that a mismatched CA is cryptographically rejected with a clear reason in the agent log (closing the issue's negative-test requirement).
- **Windows (`do_upgrade.ps1`) and the macOS `date` fix are not yet exercised on real hardware.** No Windows VM or macOS VM run has been performed for this round of changes:
  - No local PowerShell interpreter was available during development, so the Windows changes have only been reviewed manually (brace/paren balance checked, full read-through), not validated by an actual PowerShell parser.
  - The macOS `date -j -f` format string was verified against real `openssl`-generated certificate dates on a Linux dev box, but not executed on macOS itself.
  - **Known open risk on macOS**: the Wazuh macOS package does not bundle its own `openssl` binary, and neither `pkg_installer.sh` nor its invocation chain (`wm_agent_upgrade_com.c` → `wm_exec()` with a `NULL` `add_path`) sets `PATH` explicitly — the script inherits whatever `PATH` the `wazuh-modulesd` daemon itself has. Whether a working, OpenSSL-CLI-compatible `openssl` (Apple ships LibreSSL-based `/usr/bin/openssl`, unconfirmed against this codebase) is actually on that inherited `PATH` on a real macOS 15 box has not been verified. If it isn't, CA validation fails closed (rejected, upgrade proceeds unverified) rather than crashing the upgrade, but the CA would silently never be installed.
  - ARM64 vs. x86_64 is not expected to matter for any of this (same Darwin userland on both), but is likewise unverified on real hardware.

- Compilation without warnings on every supported platform: `pkg_installer.sh` syntax-checked with `bash -n`; `do_upgrade.ps1` has not been checked with an actual PowerShell parser (none available in the dev environment). The C++ changes (`curlHandle.cpp`/`curlPerformer.cpp`/`iCurlHandle.hpp`, the `trustSelfSignedRoot()` addition) were built and exercised via `https_client_unit_test` on Linux — see "Tests Introduced" for the run — but build-warning output itself wasn't captured, only the passing test run, so "without warnings" isn't independently confirmed there.
  - [ ] Linux (built and tests pass; warning-free status not independently confirmed)
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Log syntax and correct language review — self-reviewed only so far; a second pass is pending.

- Memory tests for Linux: N/A — no C/C++ code touched.
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows: N/A — no C/C++ code touched; `do_upgrade.ps1` is a script, not a compiled artifact.
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS: N/A — no C/C++ code touched.
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests _(Wazuh v4.x)_: N/A.

- Engine _(Wazuh v5.x and above)_: N/A — no engine code touched.
  - [ ] Test run in parallel

- Wazuh server API/Framework: N/A.

### Artifacts Affected

- `src/init/pkg_installer.sh` (Linux/macOS WPK upgrade script), invoked by `upgrade.sh` during a WPK-based agent upgrade.
- `src/win32/do_upgrade.ps1` (Windows WPK upgrade script) — now carries the same CA-detection/validation/install logic as the Linux/macOS side.

### Configuration Changes

No changes to `ossec.conf` or any configuration schema, on any platform. This PR installs a new file at the agent's trust-anchor path (`/var/ossec/etc/certs/root-ca.pem` on Linux/macOS, `<install_dir>\certs\root-ca.pem` on Windows) when the manager delivers a valid CA, but nothing in `ossec.conf` references it yet, on any of the three platforms.

This is a deliberate scope decision, not an omission: the issue's own "Why the scripts should not write `ossec.conf`" section prefers a file-presence-only cutover, contingent on the target build supporting anchor-driven verification. **That dependency is met by this build**: `src/client-agent/src/config.c`'s `w_agent_resolve_ssl_posture()` (lines ~168-233), called from both `ClientConf()` and `agentd.c` at startup, resolves an unset `<verification_mode>` to `full` — and defaults `<certificate_authorities>` to `AGENT_ANCHOR_CA` — whenever a readable anchor is present at that path, with no `<ssl>` edit required. This mechanism predates this PR (introduced 2026-09-08, hardened 2026-09-10 via `080774839b`) and reached this branch through the base-branch rebase, not through this PR's own commits.

(An earlier version of this section claimed the opposite — "this build has no such mechanism" — based on checking only `moduleConfig.cpp`'s `validateTls()`. That was incomplete: `validateTls()` validates an already-resolved mode/CA handed to it by the https_client bridge layer; it does not itself do the anchor-driven resolution. That resolution happens one layer upstream, in `config.c`, which is where the actual deciding logic lives. Correcting this here since it changes the answer to whether this PR activates verification — it does.)

### Tests Introduced

Automated C++ coverage for the new `trustSelfSignedRoot()` curl-handle option (`src/client-agent/https_client/tests/unit/curlPerformer_test.cpp`, `mockCurlHandle.hpp`): `ConfiguredCaIsTheWholeTrustSet` and `TlsSystemModeVerifiesPeerAndHost` updated to assert it's called whenever `caPath` is set; `TrustAnchorsWithoutConfiguredCa` updated to assert it's *not* called when `caPath` is empty; a new `RejectedTrustSelfSignedRootAbortsBeforePerforming` covers the fail-closed path if libcurl rejects the option.

Actually built and run, not just written — full `https_client_unit_test` suite, 461/461 passing, including all four tests above:

```
[----------] 29 tests from CurlPerformerTest
[ RUN      ] CurlPerformerTest.ConfiguredCaIsTheWholeTrustSet
[       OK ] CurlPerformerTest.ConfiguredCaIsTheWholeTrustSet (0 ms)
[ RUN      ] CurlPerformerTest.TrustAnchorsWithoutConfiguredCa
[       OK ] CurlPerformerTest.TrustAnchorsWithoutConfiguredCa (0 ms)
[ RUN      ] CurlPerformerTest.RejectedTrustSelfSignedRootAbortsBeforePerforming
[       OK ] CurlPerformerTest.RejectedTrustSelfSignedRootAbortsBeforePerforming (0 ms)
[ RUN      ] CurlPerformerTest.TlsSystemModeVerifiesPeerAndHost
[       OK ] CurlPerformerTest.TlsSystemModeVerifiesPeerAndHost (0 ms)
[----------] 29 tests from CurlPerformerTest (2 ms total)
...
[==========] 461 tests from 38 test suites ran. (129 ms total)
[  PASSED  ] 461 tests.
```

No automated test coverage for the shell/PowerShell installer-script logic (`pkg_installer.sh`/`do_upgrade.ps1`) — no test harness exists for either script anywhere in this codebase, so the CA-detection, validation, symlink/hard-link rejection, and atomic-install logic there remains manual/E2E-only, exercised against the live two-VM Linux lab (agent + manager) described above. Windows and macOS have not yet had an equivalent manual pass (see "Results and Evidence").

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

### DIFF
--- a/src/client-agent/https_client/src/curlHandle.cpp
+++ b/src/client-agent/https_client/src/curlHandle.cpp
@@ -19,6 +19,8 @@
 #include "moduleLog.hpp"
 
 #include <curl/curl.h>
+#include <openssl/ssl.h>
+#include <openssl/x509_vfy.h>
 
 #include <cstring>
 #include <map>
@@ -84,6 +86,23 @@ namespace
     }
 
     // curl callbacks are C: nothing may throw across them.
+
+    // CURLOPT_SSL_CTX_FUNCTION callback: sets X509_V_FLAG_PARTIAL_CHAIN on the SSL_CTX's
+    // own verification store. See ICurlHandle::trustSelfSignedRoot() for why this is
+    // needed. No userptr required, so CURLOPT_SSL_CTX_DATA is never set -- libcurl passes
+    // nullptr for it, unused here.
+    CURLcode sslCtxTrustSelfSignedRootTrampoline(CURL* /*curl*/, void* sslCtx, void* /*userptr*/)
+    {
+        auto* store = SSL_CTX_get_cert_store(static_cast<SSL_CTX*>(sslCtx));
+
+        if (store == nullptr || X509_STORE_set_flags(store, X509_V_FLAG_PARTIAL_CHAIN) != 1)
+        {
+            return CURLE_SSL_CERTPROBLEM; // LCOV_EXCL_LINE: OpenSSL misuse, not reachable in practice.
+        }
+
+        return CURLE_OK;
+    }
+
     size_t writeTrampoline(char* data, size_t size, size_t nmemb, void* userData)
     {
         auto* output = static_cast<std::string*>(userData);
@@ -253,6 +272,12 @@ namespace
             bool setOptionPtr(CurlOption option, const void* value) override
             {
                 return curl_easy_setopt(m_handle, optionMap().at(option), value) == CURLE_OK;
+            }
+
+            bool trustSelfSignedRoot() override
+            {
+                return curl_easy_setopt(m_handle, CURLOPT_SSL_CTX_FUNCTION,
+                                        sslCtxTrustSelfSignedRootTrampoline) == CURLE_OK;
             }
 
             void appendHeader(const std::string& header) override
--- a/src/client-agent/https_client/src/curlPerformer.cpp
+++ b/src/client-agent/https_client/src/curlPerformer.cpp
@@ -328,17 +328,18 @@ bool CurlPerformer::applyTrustAnchors(ICurlHandle& handle) const
         // An explicit <ca> is the whole trust set; adding the machine's stores
         // on top of it would widen what the agent accepts. (verify_mode=system's
         // Linux trust anchor also flows through here: the constructor resolves it
-        // into caPath once, up front, so this branch needs no mode-awareness.)
-        //
-        // Unconditional here (not just full/certificate): caPath may be a self-signed
-        // root that the peer also echoes in its own chain, which fails with
-        // X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN unless this is set.
+        // into caPath once, up front -- the only place below that has to tell the
+        // two apart is the partial-chain relaxation.)
         if (!setMandatoryOption(handle, CurlOption::CaInfo, m_config.caPath))
         {
             return false;
         }
 
-        if (!handle.trustSelfSignedRoot())
+        // A configured CA may be a self-signed root the peer echoes in its own chain,
+        // which fails with X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN unless this is set. Not
+        // under 'system': caPath is the OS bundle there, and relaxing chain building
+        // across the whole store widens what the agent accepts.
+        if (m_config.verifyMode != HC_VERIFY_SYSTEM && !handle.trustSelfSignedRoot())
         {
             // Unlike the options above, this one isn't in optionMap() (it's set via
             // CURLOPT_SSL_CTX_FUNCTION, not a plain curl_easy_setopt), so it can't
--- a/src/client-agent/https_client/src/curlPerformer.cpp
+++ b/src/client-agent/https_client/src/curlPerformer.cpp
@@ -333,8 +333,22 @@ bool CurlPerformer::applyTrustAnchors(ICurlHandle& handle) const
         // Unconditional here (not just full/certificate): caPath may be a self-signed
         // root that the peer also echoes in its own chain, which fails with
         // X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN unless this is set.
-        return setMandatoryOption(handle, CurlOption::CaInfo, m_config.caPath)
-               && handle.trustSelfSignedRoot();
+        if (!setMandatoryOption(handle, CurlOption::CaInfo, m_config.caPath))
+        {
+            return false;
+        }
+
+        if (!handle.trustSelfSignedRoot())
+        {
+            // Unlike the options above, this one isn't in optionMap() (it's set via
+            // CURLOPT_SSL_CTX_FUNCTION, not a plain curl_easy_setopt), so it can't
+            // route through setMandatoryOption()'s optionName() lookup -- name it
+            // directly instead of failing silently.
+            LOGFN_ERROR(m_logFn, "libcurl rejected trustSelfSignedRoot; refusing to connect without it.");
+            return false;
+        }
+
+        return true;
     }
 
 #if defined(WIN32) || defined(__APPLE__)
--- a/src/client-agent/https_client/src/curlPerformer.cpp
+++ b/src/client-agent/https_client/src/curlPerformer.cpp
@@ -329,7 +329,12 @@ bool CurlPerformer::applyTrustAnchors(ICurlHandle& handle) const
         // on top of it would widen what the agent accepts. (verify_mode=system's
         // Linux trust anchor also flows through here: the constructor resolves it
         // into caPath once, up front, so this branch needs no mode-awareness.)
-        return setMandatoryOption(handle, CurlOption::CaInfo, m_config.caPath);
+        //
+        // Unconditional here (not just full/certificate): caPath may be a self-signed
+        // root that the peer also echoes in its own chain, which fails with
+        // X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN unless this is set.
+        return setMandatoryOption(handle, CurlOption::CaInfo, m_config.caPath)
+               && handle.trustSelfSignedRoot();
     }
 
 #if defined(WIN32) || defined(__APPLE__)
--- a/src/client-agent/https_client/src/iCurlHandle.hpp
+++ b/src/client-agent/https_client/src/iCurlHandle.hpp
@@ -94,6 +94,12 @@ class ICurlHandle
         virtual bool setOptionPtr(CurlOption option, const void* value) = 0;
         virtual void appendHeader(const std::string& header) = 0;
 
+        /// Sets X509_V_FLAG_PARTIAL_CHAIN so a self-signed CaInfo root echoed back
+        /// in the peer's own chain verifies instead of failing with
+        /// X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN. Trusts nothing beyond CaInfo.
+        /// @return false if the underlying option was rejected by libcurl.
+        virtual bool trustSelfSignedRoot() = 0;
+
         /// @return false if the underlying option(s) were rejected by libcurl;
         ///         the caller must not proceed to perform() in that case, since
         ///         the requested behavior (e.g. capturing the response) would
--- a/src/client-agent/https_client/tests/unit/curlPerformer_test.cpp
+++ b/src/client-agent/https_client/tests/unit/curlPerformer_test.cpp
@@ -347,6 +347,8 @@ TEST(CurlPerformerTest, ConfiguredCaIsTheWholeTrustSet)
     EXPECT_CALL(*handle, setOptionString(CurlOption::CaInfo, "/etc/ca.pem"));
     // The machine's own stores are never added on top of it.
     EXPECT_CALL(*handle, setOptionLong(CurlOption::SslOptions, _)).Times(0);
+    // A configured CA may be a self-signed root the peer echoes back in its own chain.
+    EXPECT_CALL(*handle, trustSelfSignedRoot());
 
     auto performer = makePerformer(config, std::move(mock));
     performer.perform(HttpRequestSpec {});
@@ -359,6 +361,8 @@ TEST(CurlPerformerTest, TrustAnchorsWithoutConfiguredCa)
     allowOtherOptions(*handle);
 
     EXPECT_CALL(*handle, setOptionString(CurlOption::CaInfo, _)).Times(0);
+    // Only reached when caPath is actually set, which it is not here.
+    EXPECT_CALL(*handle, trustSelfSignedRoot()).Times(0);
 #if defined(WIN32) || defined(__APPLE__)
     // Our OpenSSL-backed Windows/macOS curl has no bundle of its own to fall back on.
     EXPECT_CALL(*handle, setOptionLong(CurlOption::SslOptions, TLS_NATIVE_CA_STORE));
@@ -368,6 +372,25 @@ TEST(CurlPerformerTest, TrustAnchorsWithoutConfiguredCa)
 
     auto performer = makePerformer(makeConfig(HC_VERIFY_FULL), std::move(mock));
     performer.perform(HttpRequestSpec {});
+}
+
+TEST(CurlPerformerTest, RejectedTrustSelfSignedRootAbortsBeforePerforming)
+{
+    // Fail-closed like every other hardening option in applyTrustAnchors(): a
+    // configured CA that curl can't be made to trust as a partial chain must not
+    // silently fall back to the classic (rejecting) chain builder.
+    auto mock = std::make_unique<NiceMock<MockCurlHandle>>();
+    auto* handle = mock.get();
+    allowOtherOptions(*handle);
+    auto config = makeConfig(HC_VERIFY_FULL);
+    config.caPath = "/etc/ca.pem";
+
+    EXPECT_CALL(*handle, trustSelfSignedRoot()).WillOnce(Return(false));
+    EXPECT_CALL(*handle, perform()).Times(0);
+
+    auto performer = makePerformer(config, std::move(mock));
+    const auto response = performer.perform(HttpRequestSpec {});
+    EXPECT_EQ(TransportStatus::TlsFail, response.status);
 }
 
 TEST(CurlPerformerTest, TlsSystemModeVerifiesPeerAndHost)
@@ -384,10 +407,12 @@ TEST(CurlPerformerTest, TlsSystemModeVerifiesPeerAndHost)
 #if defined(WIN32) || defined(__APPLE__)
     EXPECT_CALL(*handle, setOptionLong(CurlOption::SslOptions, TLS_NATIVE_CA_STORE));
     EXPECT_CALL(*handle, setOptionString(CurlOption::CaInfo, _)).Times(0);
+    EXPECT_CALL(*handle, trustSelfSignedRoot()).Times(0);
 #else
     ON_CALL(fsProbe, findSystemCaBundle()).WillByDefault(Return("/etc/ssl/certs/ca-certificates.crt"));
     EXPECT_CALL(*handle, setOptionString(CurlOption::CaInfo, "/etc/ssl/certs/ca-certificates.crt"));
     EXPECT_CALL(*handle, setOptionLong(CurlOption::SslOptions, _)).Times(0);
+    EXPECT_CALL(*handle, trustSelfSignedRoot());
 #endif
 
     auto shared = std::make_shared<std::unique_ptr<ICurlHandle>>(std::move(mock));
--- a/src/client-agent/https_client/tests/unit/curlPerformer_test.cpp
+++ b/src/client-agent/https_client/tests/unit/curlPerformer_test.cpp
@@ -412,7 +412,8 @@ TEST(CurlPerformerTest, TlsSystemModeVerifiesPeerAndHost)
     ON_CALL(fsProbe, findSystemCaBundle()).WillByDefault(Return("/etc/ssl/certs/ca-certificates.crt"));
     EXPECT_CALL(*handle, setOptionString(CurlOption::CaInfo, "/etc/ssl/certs/ca-certificates.crt"));
     EXPECT_CALL(*handle, setOptionLong(CurlOption::SslOptions, _)).Times(0);
-    EXPECT_CALL(*handle, trustSelfSignedRoot());
+    // caPath is the OS bundle here, not a configured CA: no partial-chain relaxation.
+    EXPECT_CALL(*handle, trustSelfSignedRoot()).Times(0);
 #endif
 
     auto shared = std::make_shared<std::unique_ptr<ICurlHandle>>(std::move(mock));
--- a/src/client-agent/https_client/tests/unit/mocks/mockCurlHandle.hpp
+++ b/src/client-agent/https_client/tests/unit/mocks/mockCurlHandle.hpp
@@ -30,6 +30,8 @@ class MockCurlHandle : public ICurlHandle
             .WillByDefault(::testing::Return(true));
             ON_CALL(*this, setOptionPtr(::testing::_, ::testing::_))
             .WillByDefault(::testing::Return(true));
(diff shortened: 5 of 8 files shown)


cborla:
## E2E evidence, from a real build of this branch

Ran the same style of test used on the sibling PR (#39155), against a real, from-source build of this branch — own manager (with #39070's CA delivery) + own agent binaries (`wazuh-agentd`, `libhttps_client.so`) and installer scripts from here, not the nightly package — on both Linux and Windows.

### Full single-hop E2E (4.14.7 → 5.0.0), tested against `e8205b3163` ("Address pr review comments")

```
manager log:
  legacy_task_delivery: delivering remote_upgrade task '...' to agent '006' (wpk: '...')
  legacy_task_delivery: agent '006': delivered the manager CA as 'root-ca.pem' for task '...'
  legacy_task_delivery: successfully delivered remote_upgrade task '...' to agent '006'

agent upgrade.log:
  - Found a CA delivered by the manager at ./var/incoming/root-ca.pem, validating it.
  - Installed the delivered CA at ./etc/certs/root-ca.pem. ossec.conf is not modified,
    but this alone is sufficient to activate certificate verification: the agent
    resolves an unset <verification_mode> to 'full' against a present, readable
    anchor at this path.
  - The system trust store does not verify the manager's certificate at
    127.0.0.1:1517, but the currently-installed agent (4.14.7-1) predates 5.0 and
    its config cannot express TLS verification either way -- proceeding unverified.
  ...
  - Wazuh Agent is already running (restarted by the package installer); skipping
    redundant restart.
  - Status = connected.
  - Upgrade finished successfully.
```
This run also empirically confirms two of this PR's other fixes firing correctly in the same live path: the double-restart race (no killed/dead agent after install — the "already running, skipping redundant restart" line) and the legacy-agent bypass on the `system`-mode gate (the pre-upgrade check correctly let the still-4.14.7 agent through instead of aborting).

### Version-matrix re-run against the current tip (`8b2ab4dd11`, "Applies medium and low priority fixes")

Per request, only the two stages that actually exercise this PR's own logic were re-run (4.13.x stages were already confirmed passing earlier and are unrelated to the latest fixes).

**4.14.7 → 5.0.0, Linux** — confirms the new deferred-install ordering (CA is validated and held, only written to disk after every abort-capable gate passes):
```
- Found a CA delivered by the manager at ./var/incoming/root-ca.pem, validating it.
- Delivered CA at ./var/incoming/root-ca.pem is valid; holding it until this
  script's remaining checks pass, then installing it at ./etc/certs/root-ca.pem.
...
- Installed the delivered CA at ./etc/certs/root-ca.pem. ...
...
- Status = connected.
- Upgrade finished successfully.
```

**4.14.7 → 5.0.0, Windows** — same deferred-install pattern, same outcome:
```
- Found a CA delivered by the manager at C:\...\incoming\root-ca.pem, validating it.
- Delivered CA at C:\...\incoming\root-ca.pem is valid; holding it until this
  script's remaining checks pass, then installing it at C:\...\certs\root-ca.pem.
...
- Replaced the existing CA at C:\...\certs\root-ca.pem. ...
- Installation finished.
```
The agent's *subsequent* connection then failed with an SSL SAN mismatch — this is a pre-existing gap in our test manager's own certificate (no SAN entry for its real LAN IP, only loopback/container hostnames), unrelated to this PR; we hit the identical issue testing #39155's implementation against the same manager. Included for completeness, not as a finding against this PR.

**5.0.0 → 5.0.0 forced, Linux** — confirms the CA-adoption path is a correct no-op when nothing was delivered, and the task never touches `legacy_task_delivery` at all:
```
manager log for this task: zero "legacy_task_delivery" lines (routed through the
modern https_client control channel instead)

agent upgrade.log:
  - No CA delivered by the manager at ./var/incoming/root-ca.pem this run.
  ...
  - Unpacking wazuh-agent (5.0.0-latest) over (5.0.0-latest) ...
  - Wazuh Agent is already running (restarted by the package installer); skipping
    redundant restart.
  - Status = connected.
  - Upgrade finished successfully.
```

**5.0.0 → 5.0.0 forced, Windows** — structurally confirmed the same way (task never reaches `legacy_task_delivery`, agent-side error is `wazuh-agentd: ... https_client: remote_upgrade task ...`, i.e. the modern path), but the actual delivery couldn't complete: the agent was left disconnected by the SAN issue above, so the manager's modern (persistent-connection) push has nothing to reach. Same environment limitation, not re-tested to a clean "success" for this one stage.

cborla (approved):
LGTM
- [x] Code changes reviewed

Issue #38994, opened 2026-09-07
Enrollment tokens: purge and cap the store, gate and audit the `authd.pass` disclosure, fix the token RBAC


juliancnn:
> Third and last task of #38991 (objective #38940), after #38992 (listener certificate) and #38993 (enrollment
> tokens). **This issue replaces its original scope**: the per-client rate limiter on `POST /enroll` is not being
> implemented for now, and group membership on `POST /download` is being solved in #38683. What remains is the
> lifecycle of the token store that #38993 introduced, plus the two access-control gaps around the credential the
> tokens replace.

## Description

#38993 gives the manager enrollment tokens: an operator mints one on the master, the store `etc/enrollment_tokens.json`
is replicated to every node by the cluster's file sync, and remoted verifies a `wazuh-enroll+jwt` bearer against its
read-only replica. Four gaps are left around that store and around the fleet-wide password it replaces.

1. **The store only grows.** `etoken_store_revoke()` flips a `revoked` flag and nothing ever removes an entry: neither
   the load nor the atomic rewrite discards a token that expired, was revoked, or ran out of uses
   (`src/os_auth/src/enrollment_token_store.c`; the only expiry check is at consume time, `:903`). An operator who mints
   a token per deployment accumulates dead entries forever, and the only way to clean up is editing the file by hand —
   exactly what the store's single-writer rule forbids.

2. **No cap.** `etoken_store_count()` exists (`src/os_auth/include/enrollment_token_store.h:151`) and nothing consults
   it. remoted's replica refuses any file larger than 8 MiB (`kMaxStoreBytes`,
   `src/remoted/remoted_module/src/auth/tokenKeySource.hpp:82`): past that size the read returns nothing, the previous
   replica stays in place, and the only visible sign is `remoted.enroll.token_store.reload_failures.total`. Every token
   minted after that point is silently unknown to remoted while the master keeps accepting mints. A measured entry is
   242 bytes, and 1 423 bytes when the operator passes `--embed-ca` with the 1.2 KB CA this manager ships.

3. **`GET /cluster/{node_id}/configuration/auth/auth` returns `authd.pass` in clear.** The framework reads the file from
   disk whenever `use_password` is true (`framework/wazuh/core/configuration.py:1014-1018`), and
   `@mask_sensitive_config()` (`framework/wazuh/rbac/decorators.py:671-710`) lifts the mask for any caller holding the
[...]

Issue #38991, opened 2026-09-07
Manager side of HTTPS comms tier 2: a listener certificate worth verifying, enrollment tokens, and a bounded blast radius


juliancnn:
## Description

This epic delivers the **manager half of HTTPS communications tier 2** from #38940: a listener identity an agent can verify, enrollment tokens and per-agent re-enrollment credentials, and the manager-side controls needed to operate them.

#38992 and #38993 are completed. The remaining refinement work is #38994's token-store lifecycle and access-control changes, plus the separate review follow-up #39078. Existing linked manager tasks #38683 and #39070 retain their own scope and acceptance criteria.

The final implementation differs from the original certificate-generation proposal: **the manager does not generate a CA or listener certificates.** The installation assistant or operator provisions the CA, listener certificate/key, chain and endpoint SANs before startup. Manager daemons and packages do not provide generation/reissue flags or a self-signed fallback. Missing/unreadable listener material causes startup to fail closed. Production provisioning and package validation remain delivery dependencies.

Only the manager is changed here. Agent implementation and adjacent-component work retain their own issues; the manager wire contracts below are what those clients consume.

## Deliverables and status

- [x] **#38992 — Listener trust and diagnostics.** Externally provisioned listener identity, `remote.https.ca_certificate`, public `GET /cacerts`, daily expiry checks and observable certificate status. The original child description must be reconciled with the final external-provisioning design as tracked by #39078.
- [x] **#38993 — Enrollment token end to end.** Codec, authd mint/list/revoke, cluster distribution, CLI, token authentication on `POST /enroll`, per-agent re-enrollment secret and rotation, distinguishable rejection errors, and master-executed REST API/RBAC.
- [x] **#38994 — Token lifecycle and access control.** Purge, 5,000-entry/7 MiB store limits, dedicated audited secret-read permissions, and token-list access for read-only agent roles.
- [x] **#39078 — Review corrections and delivery validation.** Nine findings affecting CA publication, identity-rotation atomicity/durability, revoke/use persistence, credential logging, CA-status freshness, CLI transport size and agent-ID validation; reconcile compatibility contracts and complete the missing integration matrix.
- [x] **#38683 — Download authorization.** Enforce group membership when serving merged configuration, under that issue's own contract.
[...]

PR #37076, opened 2026-06-23
Fix RBAC permission check to verify allow effect in update_config rules

vikman90:
## Description

This PR fixes an issue in the RBAC permission verification logic where deny rules were incorrectly interpreted as granted permissions. The `_has_update_permissions()` function was only checking if an action map existed, without verifying whether the effect was `allow` or `deny`.

This caused the sensitive configuration masking to be disabled for users with explicit deny rules on `manager:update_config` or `cluster:update_config`, when it should have remained enabled.

**Credit:** Issue reported by @ImDuong.

### Modified Files

- `framework/wazuh/rbac/decorators.py`: Updated `_has_update_permissions()` to check for `effect='allow'`
- `framework/wazuh/rbac/tests/test_decorators.py`: Added test coverage for deny rule handling

### Key Change

The permission check now verifies that at least one resource has `effect='allow'` instead of just checking if the action map is non-empty:

```python
# Before
if isinstance(action_map, dict) and len(action_map) > 0:
    return True

# After
if isinstance(action_map, dict) and any(effect == 'allow' for effect in action_map.values()):
    return True
```

### Results and Evidence

All 130 tests pass successfully, including 6 new tests specifically covering:
- Deny rules on `manager:update_config`
- Deny rules on `cluster:update_config`
- Mixed allow/deny scenarios
- End-to-end masking behavior with deny rules

```
================== 130 passed, 8 warnings in 79.76s ==================
```

### Artifacts Affected

- **Component:** RBAC Authorization
- **Module:** `wazuh.rbac.decorators`
- **Function:** `_has_update_permissions()`
- **Decorator:** `@mask_sensitive_config()`

### Configuration Changes

None. This is a code-only fix with no configuration changes required.

### Documentation Updates

None required. The fix corrects internal behavior without changing the external API or user-facing functionality.

### Tests Introduced

Added 6 new test cases in `framework/wazuh/rbac/tests/test_decorators.py`:

1. `test_has_update_permissions_deny_manager_update_config` - Verifies deny on manager:update_config returns False
2. `test_has_update_permissions_deny_cluster_update_config` - Verifies deny on cluster:update_config returns False
3. `test_has_update_permissions_mixed_deny_allow_allows` - Verifies mixed rules return True when at least one allow exists
4. `test_has_update_permissions_all_deny` - Verifies all deny rules return False
5. `test_mask_sensitive_config_raw_xml_with_deny_rule` - End-to-end test with manager:update_config deny
[...]

PR #37039, opened 2026-06-22
Patch cluster authentication

ignaciogalle12git:
## Description

The `@mask_sensitive_config()` decorator masks some configuration fields in configuration-read responses for users without `manager:update_config` / `cluster:update_config`. It was already applied to `GET /manager/configuration` (`wazuh.manager.read_ossec_conf`) but not to `GET /cluster/local/config` (`wazuh.cluster.read_config_wrapper`), so the two configuration-read endpoints behaved inconsistently.

This change applies the same decorator to `read_config_wrapper` so both endpoints mask sensitive fields the same way. 

## Proposed Changes

- `framework/wazuh/cluster.py`: import `mask_sensitive_config` from `wazuh.rbac.decorators` and decorate `read_config_wrapper` with `@mask_sensitive_config()` (above the existing `@expose_resources`), matching the sibling `wazuh.manager.read_ossec_conf`. The wrapper appends a copy of the configuration (`dict(read_config())`) so the in-place masking operates on a per-response copy and does not mutate the cached dict returned by `read_config()`.
- `framework/wazuh/tests/test_cluster.py`: unit tests covering the masking behaviour and the no-cache-mutation guarantee.

## Configuration options

N/A — no new configuration parameters; RBAC defaults are unchanged.

## Results and Evidence

Framework unit suites under Python 3.12 (`PYTHONPATH=api:framework`):

- `framework/wazuh/tests/test_cluster.py` → 14 passed (4 new cases for the masking behaviour and the no-cache-mutation guarantee).
- `framework/wazuh/rbac/tests/test_decorators.py` + `framework/wazuh/tests/test_manager.py` → 164 passed.
- Full framework suite → 2694 passed.

Manual check on a single-node cluster with a key configured performed.

### Tests Introduced

- `test_read_config_wrapper_key_visibility` (parametrized) — cluster key masked (`*****`) for users without update-config and returned in clear for update-config holders.
- `test_read_config_wrapper_masking_preserves_other_fields` — masking only touches the `key` field, leaving the other config fields intact.
- `test_read_config_wrapper_no_cache_poisoning` — masking operates on a per-response copy, so the dict returned by `read_config()` is not mutated.
- `test_read_config_wrapper` pinned to the update-config path so it keeps asserting the unmasked behaviour.

## Credits

Thanks to @evilgensec and @geo-chen for raising this issue to our attention.

## Tests

- Compilation without warnings in every supported platform
  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X
[...]

PR #39142, opened 2026-09-10
API readonly role is able to export client and cluster keys

### LINKED ISSUE #38842: Shipped API readonly Role Can Export Every Agent Key and the Cluster Key
## Executive summary

The Manager API ships a role named **readonly**. Its processed policies are:

```
agent:read          agent:id:* , agent:group:*
group:read          group:id:*
cluster:status      *:*:*
cluster:read        *:*:* , node:id:*
cluster:read_api_config
mitre:read
rbac_mode           white
```

Writes are blocked. `PUT /agents/003/restart` as that user is **403**. That part works.

These two reads are **200** and return secrets:

| Call | What comes back |
|---|---|
| `GET /agents/003/key` | Base64 of `003 ubuntu-24 any <64-hex>` — the line in `client.keys` |
| `GET /cluster/local/config` | JSON including `key` (the 32-hex cluster key) and `bind_addr` |

`GET /cluster/node01/configuration` also returns the same cluster key inside the `cluster` object.

`agent:read` is treated as “read everything about the agent,” including the shared secret used on 1517. `cluster:read` is treated as “read the running config,” including the key that authenticates 1516.

A SOC analyst who is supposed to list agents can clone any of them. The same token prints the cluster key for a listener that is bound to `0.0.0.0:1516` on this AIO.

This was tested with a temporary API user that had **only** role `readonly`. The user was deleted afterwards. No agent was cloned. The cluster was not joined.

| Topic | Detail |
|---|---|
| What the operator sees | A “readonly” role for analysts |
| What actually happened | `agent:read` includes key export; `cluster:read` includes the cluster secret |
| Impact | Least-privilege API users can impersonate agents and have the cluster key |
| Owning component | RBAC policy mapping for `GET /agents/{id}/key` and cluster config |
| Fix | Key export needs `agent:read_key` (or admin only). Cluster key must not appear on `cluster:read`. Redact it |

Critical because the shipped least-privilege role is enough to take over the agent fleet.

---

## 1. Environment

| Component | Detail |
|---|---|
| Manager | 5.0.0-1, API revision **beta5**, `:55000` on `0.0.0.0` |
| RBAC | `white` mode |
| Role | id `2`, name `readonly` (shipped) |
| Test user | created, assigned only role `2`, then deleted |
| Window | 2026-09-02 |

Unauthenticated calls to the same paths are **401**. This is not “no auth.” It is “the wrong permission is enough.”

---

## 2. What readonly can and cannot do

Denied (correct):

```
PUT /agents/003/restart     403  Permission denied: Resource type: agent:id
GET /security/users         200  empty list (cannot see other users)
```

Allowed (the defect):

```
GET /agents                 200  inventory
GET /agents/003/key         200  key field present
GET /cluster/local/config   200  cluster.key present
GET /cluster/node01/configuration  200  cluster.key present
GET /groups/default/files/agent.conf  200
```

Decoded `GET /agents/003/key` payload shape (hash omitted):

```
003 ubuntu-24 any <64 hex characters>
```

`/var/wazuh-manager/etc/client.keys` has the same id and name for agent 003, with a 64-character key field. The API returns that secret, not a derived public token.

`/cluster/local/config` returns `name`, `node_name`, `node_type`, `key`, `port` 1516, `bind_addr` `0.0.0.0`, `nodes`. The key is the same value as `<cluster><key>` in `wazuh-manager.conf`.

```mermaid
flowchart LR
  A["API user with role readonly"] --> B["agent:read"]
  B --> C["GET /agents/id/key"]
  C --> D["client.keys secret"]
  A --> E["cluster:read"]
  E --> F["GET /cluster/local/config"]
  F --> G["cluster key + 0.0.0.0:1516"]
```

---

## 3. Why this is not “admin can see keys”

The built-in **administrator** role should be able to export keys. That is a break-glass operation.

The built-in **readonly** role exists so an operator can give someone inventory and MITRE without giving them the estate. Mapping key export and the cluster secret onto `*:read` makes that role unsafe to assign.

`allow_run_as` on the test user was **false**. This is not the `wazuh-wui` run_as path.

---

## 4. Proposed fix

**P0 — split the permission**

`GET /agents/{id}/key` must require a dedicated action (`agent:read_key` / `agent:export_key`), granted only to administrator (and maybe `agents_admin`). `agent:read` stays list/status/os.

**P0 — redact cluster.key**

`cluster:read` and `cluster:read_api_config` should return name, nodes, port, bind address — not the key. A separate `cluster:read_key` for administrators, or never via the API (only on disk).

**P1 — audit the rest of readonly**

Anything that is a credential, a private key path, or `authd.pass` must not ride on a generic `read` action. `GET /cluster/node01/configuration` currently echoes the running config including the cluster key.

---

## 5. How to verify

1. Create an API user. Attach only role `readonly`. Login.
2. Today: `GET /agents/001/key` is 200 with a decodable `client.keys` line. `GET /cluster/local/config` includes `key`.
3. After the fix: both calls are 403 for that user. Administrator still gets 200 on the key route if you keep that capability at all.
4. `PUT /agents/001/restart` stays 403 for readonly.

---

## 6. What this is not

- Not “the API has no authentication.” Unauthenticated is 401.
- Not the indexer `wazuh-demo` default-password report. Different port, different users. Same class: a named low-privilege identity that is not low privilege.
- Not a proof that an agent was impersonated. The key was compared to `client.keys` and not used.

jepalfer:
## Description

The shipped `readonly` role could export any agent's pre-shared key. `GET /agents/{agent_id}/key` was gated by `agent:read` — the action every agent-inventory read needs — so any role able to list agents could also retrieve the verbatim `client.keys` line and impersonate that agent. `PUT /agents/{agent_id}/restart` was correctly denied for the same role, which is part of why the mapping read as intentional.

No existing `agent:*` action fits the semantic slot "read this agent's credential", so this adds `agent:read_secrets` and moves the endpoint onto it. The name and the privilege level follow `cluster:read_secrets`, the action #38994 introduced for the other half of this report, so both secrets are gated the same way. `agent:read` keeps inventory reads (id, name, group, last keep alive, OS, status) and no longer grants key export.

The cluster-key half of the report needs no code change here, but the reason is no longer the one this PR was opened with. `GET /cluster/local/config` and `GET /cluster/{node_id}/configuration` mask `cluster.key` and `authd.pass` for any user lacking `cluster:read_secrets`, a dedicated action checked by `_can_read_secrets()` in `framework/wazuh/rbac/decorators.py` and granted only to `administrator`, through the `secrets_read` default policy. That action landed with #38994 and replaced the earlier gate, which keyed the same masking off `cluster:update_config` / `manager:update_config` (#38374, ported from 4.14.7 #37039 and #37076). The report was written against a `beta5` package predating all of it; a `readonly` user gets `key: "*****"` from both endpoints on this base, confirmed live below.

Closes #38842

## Proposed Changes

| File | Change |
|---|---|
| `framework/wazuh/agent.py` | `get_agents_keys` requires `agent:read_secrets` — the actual enforcement point — and records the disclosure |
| `framework/wazuh/rbac/decorators.py` | `audit_agent_keys_read()`: the `secret_read` line for a key read. `mask_sensitive_config` never sees this disclosure, since the key is the whole answer rather than a payload to mask |
| `api/api/spec/spec.yaml` | new `agent:read_secrets` entry in `x-rbac-catalog/actions` (resources `agent:id`, `agent:group`); `/agents/{agent_id}/key` re-pointed off `agent:read`. The catalog is what `GET /security/actions` serves |
| `framework/wazuh/rbac/default/policies.yaml` | `agent:read_secrets` added to the `secrets_read` policy, next to `cluster:read_secrets`, and deliberately to no other |
| `framework/wazuh/tests/data/security/rbac_catalog.yml` | the key endpoint moves from `agent:read`'s `related_endpoints` to the new action's |
| `api/test/integration/env/configurations/rbac/agent/{white,black}_config.{yaml,txt}` | `agent:read_secrets` granted per agent id, with one id per mode deliberately left holding `agent:read` alone; `.txt` matrices updated to match |
| `api/test/integration/test_rbac_{white,black}_agent_endpoints.tavern.yaml` | one stage each asserting that id's key endpoint is denied |
| `docs/ref/modules/server-api/authentication.md` | the secrets contract: `secrets_read` now carries two actions, a section for agent keys and the existing-vs-new boundary, and the audit section covers this endpoint |
| `framework/wazuh/rbac/tests/test_decorators.py` | two cases: a served key is audited naming caller and agents, an empty read is not |
| `docs/ref/modules/rbac/README.md`, `CHANGELOG.md` | reference row + note, changelog entry |

The endpoint declares the new action **alone**, not alongside `agent:read`. Multiple actions in `expose_resources` are unioned, not intersected — `_match_permissions` accumulates every action into a single `allow_match` set through `set.update()` — so declaring both would keep granting the endpoint to anyone holding only `agent:read`, which is the opposite of the intent.

Role impact: `secrets_read` is held only by `administrator`, so that is now the single built-in role that can export an agent's key. `agents_admin` and `wazuh_indexer_admin` keep every other agent operation through `agents_all` but lose key export, as do `readonly` and `agents_readonly`.

Provisioning is deliberately untouched, and the boundary is existing versus new. `agent:create` stays in `agents_all`, so `POST /agents` still answers with the key of the agent it just created — an agent cannot be created without one — and minting an enrollment token is unchanged. What now needs `administrator` is re-reading the key of an agent that already exists, which is the impersonation of a live identity this report is about. An operator who provisions agents therefore loses nothing.

Joining `secrets_read` also means joining its contract, so a key read now emits the same `secret_read` audit line the configuration endpoints already emitted, naming the caller and the agents served and never a key.

### Results and Evidence

Live on Vagrant `master` + `agent1` with packages built from this branch at `f779d82`
(`wazuh-manager_5.0.0-0_amd64_f779d82.deb`, `wazuh-agent_5.0.0-0_amd64_f779d82.deb`), built from a copy of the tree rather than the working tree itself. `/var/wazuh-manager` was purged first, so the RBAC database was seeded fresh from the packaged `policies.yaml`; `agent1` enrolled over HTTPS as **agent 001** with a real `client.keys` entry, manager config left exactly as the package ships it. A 15-check script drove the API: create a throwaway user holding **only** role `readonly` (id 2, `allow_run_as: false`), exercise the endpoints, delete the user.

| Build / state | `readonly` → `/agents/001/key` | `administrator` → same | Script |
|---|---|---|---|
| Pre-change deb (`acd0a3d`), fresh install | **200**, decodes to `001 agent1 any 689a1bf3…` | 200 | 10 passed, 2 failed |
| This change (`f779d82`), fresh install | **403** | 200 | **15 passed, 0 failed** |

Row 1 reproduces the reported defect verbatim: a `readonly`-only token returns the base64 `client.keys` line for agent 001, which decodes to that agent's exact `client.keys` line. Row 2 is the fix. (Row 1 was measured with an earlier 12-check version of the script; the two extra checks in row 2 are the negative case below.)

The isolated negative case — the property the fix actually introduces, checked against a purpose-built policy rather than a shipped role:

- a policy granting **only** `agent:read` on `agent:id:001` → `GET /agents/001/key` **403**, `GET /agents?agents_list=001` **200**. `agent:read` alone no longer exports a key, and still reads the agent.

Unchanged behaviour confirmed in the same run:

- `PUT /agents/001/restart` as `readonly` → **403** (as before).
- `GET /agents` as `readonly` → **200**, `total_affected_items=1`. Narrowing the key endpoint does not disturb inventory reads.
- `GET /security/actions` as `administrator` → `agent:read_secrets` present with `resources: ['agent:id', 'agent:group']` and `related_endpoints: ['GET /agents/{agent_id}/key']`; `agent:read` no longer lists that endpoint.
- `GET /cluster/local/config` as `readonly` → `key: "*****"` (the already-fixed cluster half).

RBAC database inspected directly on the fresh install, rather than inferred from the YAML:

```
pragma user_version = 1                       <- no ORM bump, see Configuration Changes
policies id=2  agents_all_agents
  actions: ['agent:read', 'agent:read_key', 'agent:delete', 'agent:modify_group',
            'agent:restart', 'agent:reload', 'agent:upgrade', 'agent:scan_vulnerability']
roles    id=2  readonly -> policy_ids [4, 5, 11, 14, 15]   (policy 2 not among them)
```

That dump predates the move to `secrets_read`: it was taken when the action was still named
`agent:read_key` and seeded into `agents_all`. What it establishes and what still holds is that the
defaults reach the database on a fresh install without an ORM bump, and that `readonly` does not
hold the policy carrying the action. The seeded policy is now `secrets_read`, held only by
`administrator`; re-measuring on a package built from the current head is pending.

Unit tests on the host: `framework/wazuh/rbac/tests/` **254 passed**, `framework/wazuh/tests/test_agent.py` **115 passed**, remaining `framework/wazuh/tests` **141 passed**. `framework/wazuh/tests/test_security.py` cannot run outside an installed manager — its `db_setup` shells out to `bin/wazuh-manager-conf` — so `rbac_catalog.yml` was validated by replicating `get_rbac_actions`' loop against the spec and diffing all four expected cases, including the fetch-all case, which on the current base returns 34 actions: all four match. The replay was re-run after rebasing onto the 5.0.0 that now carries #38991's identity work, which itself added four RBAC actions; all four cases still match. That module, plus `test_cluster.py`, `test_manager.py`, `test_stats.py` and `api/api/controllers/test/test_agent_controller.py`, failed identically on a pristine checkout of the base commit this PR was first written against, so the baseline was unchanged by this PR. That measurement predates the rebase and has not been repeated against the current base.

- Compilation without warnings
  - [x] Linux — manager and agent packages built from this branch on the package builder; Python-only change, no compiled artifact altered
- [x] Log syntax and correct language review — no new log lines; user-facing strings limited to the RBAC action description in `spec.yaml`

- Wazuh server API/Framework
  - [x] Run API Integration Tests — **run in CI.** The last full run before the rebase passed all twelve tavern suites, `test_rbac_white_agent_endpoints.tavern.yaml` and `test_rbac_black_agent_endpoints.tavern.yaml` among them; they are re-running on the rebased head. The suites need the 12-container docker environment, so they were never run locally. Three files cover this endpoint (`test_rbac_white_agent_endpoints.tavern.yaml`, `test_rbac_black_agent_endpoints.tavern.yaml`, `test_rbac_white_all_endpoints.tavern.yaml`); the existing allow/deny expectations are preserved (white: 006 allowed, 005 denied; black: 005 allowed, 006 denied) and one negative-case stage is added per mode (see Tests Introduced). `integration_test_api_endpoints.json` already maps the changed files to those suites.

### Artifacts Affected

- `framework` and `api` Python packages, shipped inside the manager package under `framework/python/lib/python3.12/site-packages/{wazuh,api}/`.
- `api/spec/spec.yaml` — the served OpenAPI document and RBAC catalog.
- `framework/wazuh/rbac/default/policies.yaml` — seeds the RBAC database on a fresh install.
- No compiled binary, no service unit, no installer layout change. `wazuh-manager` and `wazuh-agent` packages were both rebuilt at `f779d82` and installed during testing.

### Configuration Changes

New RBAC action `agent:read_secrets`, scopable per agent and per group exactly like `agent:read`, seeded into the `secrets_read` default policy. No change to `wazuh-manager.conf`, no new internal option.

`CURRENT_ORM_VERSION` is **not** bumped. On a fresh installation the create branch of `check_database_integrity()` seeds the default policies straight from `policies.yaml` and never reads that constant, so the new action is present without it — row 2 above is a fresh install of a package that does not carry a bump. This matches the branch's standing convention: the `5_codeanalysis_api-rbac-db-version.yml` guard skips `*.0.0`, the counter was deliberately reverted to 1 for the 5.0 release, and `ff4503222b` added `agent:scan_vulnerability` to this same policy block without a bump.

**Known limitation, accepted for 5.0.0.** An in-place 5.0.0 → 5.0.0 package upgrade over an existing `rbac.db` does not pick the new action up: default policies are re-seeded only when the database is created or when an ORM-version migration rebuilds it, and there is no administrator bypass, so on such a box the endpoint requires an action no stored policy grants and every role — `administrator` included — is denied. Measured: `403` for both `readonly` and `administrator`, with `pragma user_version` still 1 and the stored `agents_all_agents` unchanged; bumping to 2 alone restored `200` for `administrator` and re-seeded the policy while preserving custom roles, policies and users (ids ≥ 100). 5.0.0 installs fresh, so this path is out of scope here — but **the forward-port to `5.0.1`/`main` must include the bump**, where the guard is live and the counter already stands at 4.

**Breaking change.** `agents_admin` and `wazuh_indexer_admin` lose key export, and any *custom* policy that grants `agent:read` today loses it too. Custom policies are preserved verbatim by any future migration, so those operators must add `agent:read_secrets` by hand. Recorded in `CHANGELOG.md` under `#### Changed` and in the RBAC reference; flag it for the release notes if a stronger upgrade note is wanted.

### Tests Introduced

No new test files; the existing RBAC suites were extended so they can actually fail if this fix is undone.

- **The isolated negative case.** Each mode now has one agent id holding `agent:read` and **not** `agent:read_secrets` — `012` in white, `011` in black — and a tavern stage asserting `GET /agents/<id>/key` is denied for it. Both ids are already asserted as `agent:read`-allowed by the existing list stages, so the pair of assertions pins exactly the property this PR introduces. Without such an id the suites would stay green if `get_agents_keys` were re-pointed at `agent:read`, since every id with `agent:read` would also carry `agent:read_secrets`; the only signal would come from `test_security.py`, which cannot run outside an installed manager.
- `framework/wazuh/tests/data/security/rbac_catalog.yml` — the key endpoint moves to an `agent:read_secrets` entry; otherwise that fixture would keep passing while documenting the old mapping.
- The remaining per-agent-id grants keep the three suites testing per-resource scoping on the key endpoint rather than a blanket denial.

The property was verified on a live manager as well as structurally: a policy granting only `agent:read` on `agent:id:001` returns **403** on `GET /agents/001/key` and **200** on `GET /agents?agents_list=001`.

`framework/wazuh/tests/test_agent.py` needs no change: its conftest swaps `expose_resources` for `RBAC_bypasser`, so it never observes the action name.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...


framework/wazuh/rbac/default/policies.yaml
```diff
@@ -15,6 +15,7 @@ default_policies:
       agents:
         actions:
           - agent:read
+          - agent:read_key
```
jotacarma90: **[S2 · blocking · introduced]** Adding `agent:read_key` here only reaches a database that is being created: `check_database_integrity()` re-seeds the defaults on the create branch, or when `current_version < CURRENT_ORM_VERSION` rebuilds the file — and `CURRENT_ORM_VERSION` stays at 1, while `debian/preinst:150` / `postinst:225` preserve `rbac.db` across a package upgrade. `expose_resources` has no bypass for the administrator, so on an in-place 5.0.0 upgrade the stored `agents_all_agents` keeps its old action list and `GET /agents/001/key` returns 403 error 4000 for every role, `wazuh` included.

The precedent the description cites (`ff4503222b`, `agent:scan_vulnerability`) does not cover this: those changes *added* an action for a new endpoint, so an upgraded box merely missed a new feature. This one *revokes* an action a stored policy already granted, so an upgraded box loses an endpoint that worked, with no diagnostic beyond a permission denial.

The cheap exit is the bump you already measured, in `framework/wazuh/rbac/orm.py:47`:

```py
CURRENT_ORM_VERSION = 2
```

The migration branch re-seeds the defaults and carries over ids >= 100 plus the `wazuh` / `wazuh-wui` users. If the team prefers to keep the no-bump decision, then the CHANGELOG qualifier below stops being optional and the forward-port bump needs an issue tracking it.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 
 | Issue | Comment |
 |-------|---------|
+| [#38842](https://github.com/wazuh/wazuh/issues/38842) | Changed the RBAC action required by `GET /agents/{agent_id}/key` from `agent:read` to the new `agent:read_key` action, so a role that can list agents no longer exports the pre-shared key needed to impersonate them. The built-in `administrator`, `agents_admin` and `wazuh_indexer_admin` roles keep the capability through the `agents_all` policy; the built-in `readonly` and `agents_readonly` roles lose it. Custom policies that grant `agent:read` must add `agent:read_key` explicitly to retain key export. |
```
jotacarma90: **[S3 · non-blocking · introduced]** "The built-in `administrator`, `agents_admin` and `wazuh_indexer_admin` roles keep the capability through the `agents_all` policy" is false on any `rbac.db` preserved across a package upgrade — the case in F1. The #38862 entry at `CHANGELOG.md:54` already carries the wording for exactly this situation.

If F1 is fixed by bumping the ORM version, change the qualifier to say the preserved database is rebuilt by the version-2 migration instead.

```suggestion
| [#38842](https://github.com/wazuh/wazuh/issues/38842) | Changed the RBAC action required by `GET /agents/{agent_id}/key` from `agent:read` to the new `agent:read_key` action, so a role that can list agents no longer exports the pre-shared key needed to impersonate them. The new action applies to an `rbac.db` seeded from the defaults — a fresh installation or a `rbac_control factory-reset`; a database preserved across a package upgrade keeps the policies it already had, and on such a box the endpoint is denied for every role. On a seeded database the built-in `administrator`, `agents_admin` and `wazuh_indexer_admin` roles keep the capability through the `agents_all` policy; the built-in `readonly` and `agents_readonly` roles lose it. Custom policies that grant `agent:read` must add `agent:read_key` explicitly to retain key export. |
```

api/test/integration/env/configurations/rbac/agent/black_config.yaml
```diff
@@ -20,6 +21,19 @@
...
+# denied. That is the only case proving key export no longer rides on agent:read.
+- actions:
+  - agent:read_key
+  resources:
+  - agent:id:001
+  - agent:id:003
+  - agent:id:005
+  - agent:id:007
+  - agent:id:009
+  effect: allow
```
jotacarma90: **[S3 · non-blocking · introduced]** The `.txt` matrices next to these files list one row per action with its allow/deny id sets and have matched the yaml exactly until now; neither got an `agent:read_key` row. A reader consulting `black_config.txt` sees only `agent:read | agent:id | 001,003,005,007,009,011 | *` and concludes `GET /agents/011/key` is allowed, which is the opposite of what the new stage asserts.

```text
# api/test/integration/env/configurations/rbac/agent/black_config.txt, after line 31
| agent:read_key           | agent:id  | 001,003,005,007,009                     | *                                   |
```

```text
# api/test/integration/env/configurations/rbac/agent/white_config.txt, after line 31
| agent:read_key           | agent:id  | 002,004,006,008,010                 |                     |
```

framework/wazuh/rbac/default/policies.yaml
jotacarma90: **[S2 · non-blocking · introduced]** Recording the consequence of this line rather than asking for a change, since the decision to keep `CURRENT_ORM_VERSION` at 1 is the right one for this branch.

Adding `agent:read_key` here only reaches a database that is being created: `check_database_integrity()` re-seeds the defaults on the create branch, or when `current_version < CURRENT_ORM_VERSION` rebuilds the file. With the counter frozen at 1 and `debian/preinst:150` / `postinst:225` preserving `rbac.db` across a package upgrade, an in-place 5.0.0 upgrade keeps the stored `agents_all_agents` untouched. `expose_resources` has no bypass for the administrator, so `GET /agents/001/key` returns 403 error 4000 for every role, `wazuh` included, on such a box.

That is acceptable here — the guard skips `*.0.0` since #35056 and there is no 4.x-to-5.x upgrade path, so 5.0.0 GA is always a fresh install — but it is worth using the right precedent in the description. `ff4503222b` does not cover this case: it *added* an action for a new endpoint, so an upgraded box merely missed a new feature. The comparable change is #38862, which altered an RBAC default without a bump and scoped its CHANGELOG entry to a database seeded from the defaults, naming `rbac_control factory-reset` as the way out (`CHANGELOG.md:54`). Two follow-ups from that:

1. the CHANGELOG qualifier below, so an operator upgrading a prerelease box is not told the opposite of what they will see;
2. an issue pinning the `CURRENT_ORM_VERSION` bump to the forward-port to `main`, where the counter is at 4 and the guard actually runs. The description already flags it, but nothing enforces it.

CHANGELOG.md
jotacarma90: **[S3 · blocking · introduced]** "The built-in `administrator`, `agents_admin` and `wazuh_indexer_admin` roles keep the capability through the `agents_all` policy" is false on any `rbac.db` preserved across a package upgrade — there the endpoint is denied for every role, `administrator` included. With `CURRENT_ORM_VERSION` staying at 1 this is the shipped behaviour, so the entry has to say it. #38862 at `CHANGELOG.md:54` already established the wording for this exact situation.

```suggestion
| [#38842](https://github.com/wazuh/wazuh/issues/38842) | Changed the RBAC action required by `GET /agents/{agent_id}/key` from `agent:read` to the new `agent:read_key` action, so a role that can list agents no longer exports the pre-shared key needed to impersonate them. The new action applies to an `rbac.db` seeded from the defaults — a fresh installation or a `rbac_control factory-reset`; a database preserved across a package upgrade keeps the policies it already had, and on such a box the endpoint is denied for every role until the database is reseeded. On a seeded database the built-in `administrator`, `agents_admin` and `wazuh_indexer_admin` roles keep the capability through the `agents_all` policy; the built-in `readonly` and `agents_readonly` roles lose it. Custom policies that grant `agent:read` must add `agent:read_key` explicitly to retain key export. |
```

api/test/integration/env/configurations/rbac/agent/black_config.yaml
jotacarma90, line 33: **[S3 · non-blocking · introduced]** The `.txt` matrices next to these files list one row per action with its allow/deny id sets and have matched the yaml exactly until now; neither got an `agent:read_key` row. A reader consulting `black_config.txt` sees only `agent:read | agent:id | 001,003,005,007,009,011 | *` and concludes `GET /agents/011/key` is allowed, which is the opposite of what the new stage asserts.

```text
# api/test/integration/env/configurations/rbac/agent/black_config.txt, after line 31
| agent:read_key           | agent:id  | 001,003,005,007,009                     | *                                   |
```

```text
# api/test/integration/env/configurations/rbac/agent/white_config.txt, after line 31
| agent:read_key           | agent:id  | 002,004,006,008,010                 |                     |
```

api/test/integration/env/configurations/rbac/agent/black_config.yaml
```diff
@@ -20,6 +21,19 @@
   - agent:id:011
   effect: allow
 
+# 011 is deliberately absent: the deny-all block above still denies it
+# agent:read_key while agent:read stays allowed, so GET /agents/011/key must be
+# denied. That is the only case proving key export no longer rides on agent:read.
```
jotacarma90: **[S4 · non-blocking · introduced]** "the deny-all block above still denies it agent:read_key" reads as if a word were missing, and "deny-all block" names a block that denies six specific actions on `agent:id:*`. The closing sentence describes the history of the change rather than the invariant the block encodes. The invariant is one line:

```suggestion
# 011 holds agent:read and not agent:read_key: GET /agents/011/key must be denied.
```

api/test/integration/env/configurations/rbac/agent/white_config.yaml
```diff
@@ -10,6 +10,19 @@
   - agent:id:012
   effect: allow
 
+# 012 is deliberately absent: it holds agent:read but not agent:read_key, so
+# GET /agents/012/key must be denied. That is the only case proving key export
+# no longer rides on agent:read.
```
jotacarma90: **[S4 · non-blocking · introduced]** Same as the black config: the last sentence narrates the change instead of stating what the omission means.

```suggestion
# 012 holds agent:read and not agent:read_key: GET /agents/012/key must be denied.
```

docs/ref/modules/rbac/README.md
```diff
@@ -139,4 +140,6 @@ The following RBAC actions control agent operations:
...
\ No newline at end of file
+The `agent:scan_vulnerability` action was introduced in v5.0.0 alongside `PUT /agents/scan/vulnerability`, the operator-facing endpoint to trigger an on-demand Vulnerability Detection scan, complementing the previous agent-initiated-only scan trigger.
+
+The `agent:read_key` action was introduced in v5.0.0 to gate `GET /agents/{agent_id}/key`, which returns the agent's pre-shared key. That endpoint previously required only `agent:read`, so any role able to list agents could also export the secret used to impersonate them. `agent:read` no longer grants key export: it covers agent information (id, name, group, last keep alive, OS, status) and nothing else. The action is scopable per agent and per group, exactly like `agent:read`, and the built-in `administrator`, `agents_admin` and `wazuh_indexer_admin` roles hold it through the `agents_all` policy. The built-in `readonly` and `agents_readonly` roles do not.
```
jotacarma90: **[S4 · non-blocking · introduced]** "The built-in `readonly` and `agents_readonly` roles do not" is true in white mode only. Under `rbac_mode: black`, `_black_expansion` (`decorators.py:270`) seeds the allowed set with every agent and `_single_processor` only subtracts what a policy explicitly denies; since no shipped policy mentions `agent:read_key`, the action is allowed by default there.

This is not a defect of the PR — before the change `readonly` exported keys in black mode too, and in that mode it can also delete and restart agents. Please do not "fix" it with an explicit `deny` on `agents_read`: `process_policy` resolves by order and the last policy wins (`preprocessor.py:88-98`), so that deny would override the `allow` from `agents_all` for anyone holding both `readonly` and `administrator`. A clause scoping the sentence is enough. (The file also still ends without a trailing newline.)

```suggestion
The `agent:read_key` action was introduced in v5.0.0 to gate `GET /agents/{agent_id}/key`, which returns the agent's pre-shared key. That endpoint previously required only `agent:read`, so any role able to list agents could also export the secret used to impersonate them. `agent:read` no longer grants key export: it covers agent information (id, name, group, last keep alive, OS, status) and nothing else. The action is scopable per agent and per group, exactly like `agent:read`, and the built-in `administrator`, `agents_admin` and `wazuh_indexer_admin` roles hold it through the `agents_all` policy. The built-in `readonly` and `agents_readonly` roles do not. That last sentence describes `rbac_mode: white`, the shipped mode; under `black` mode every action no policy explicitly denies is allowed, so `agent:read_key` is granted there by default, as is any newly added action.
```

jotacarma90 (changes requested):

CHANGELOG.md
jepalfer: Applied

api/test/integration/env/configurations/rbac/agent/black_config.yaml
jepalfer: Applied

api/test/integration/env/configurations/rbac/agent/white_config.yaml
jepalfer: Applied

api/test/integration/env/configurations/rbac/agent/black_config.yaml
jepalfer: Applied

docs/ref/modules/rbac/README.md
jepalfer: Applied

framework/wazuh/rbac/default/policies.yaml
jotacarma90: **[S2 · non-blocking · introduced]** Persists from iteration 1 (F1), left open per prior team decision: `CURRENT_ORM_VERSION` stays at 1 for 5.0.0, so on a preserved `rbac.db` (in-place upgrade) this endpoint is denied for every role, `administrator` included, until the database is reseeded. That consequence is now correctly documented in the CHANGELOG. No issue pinning the ORM bump to the `main` forward-port has been opened yet; worth creating one so it isn't lost before that port happens.

jotacarma90 (dismissed):

framework/wazuh/rbac/default/policies.yaml
```diff
@@ -18,6 +18,7 @@ default_policies:
...
```
jotacarma90: **[S2 · non-blocking · introduced]** Recording the consequence, not asking for a change — keeping `CURRENT_ORM_VERSION` at 1 is the right call for this branch, and the identity work that just merged adds its own default policies on exactly the same terms without a bump.

Default policies reach the database only on creation or through an ORM-version migration, so an in-place 5.0.0 upgrade keeps the stored `agents_all_agents` without this action and `GET /agents/{agent_id}/key` answers 403 for every role, `wazuh` included. That is acceptable here: 5.0.0 GA is always a fresh install and there is no 4.x-to-5.x upgrade path. The CHANGELOG now scopes the claim accordingly.

The forward-port needs no tracking issue after all: on `main` the version is 5.1.0, so `5_codeanalysis_api-rbac-db-version.yml` does not skip, and it fails any change under `framework/wazuh/rbac/default/**` that does not also bump `CURRENT_ORM_VERSION`. The upward merge cannot land without it.

Separately, and this one is a real question rather than a note: the identity work that merged into this base put `cluster:read_secrets` in a dedicated `secrets_read` policy granted only to `administrator`, while this line puts `agent:read_key` into `agents_all`, which `agents_admin` and `wazuh_indexer_admin` also hold. Two answers to "who may read a secret" shipping in the same release, with divergent names too. Worth settling deliberately.

jotacarma90 (approved):

PR #39144, opened 2026-09-10
wazuh-remoted: Authorize POST /download against the requesting agent's own groups

### LINKED ISSUE #38683: remoted /download serves any group's merged.mg to any authenticated agent, no membership check
|---|---|---|---|---|
| 5.1.0-alpha0 | wazuh-remoted (remoted_module, POST /download) | Manager | Sources | Linux |

## Summary

`POST /download` authenticates the calling agent but never checks that the agent belongs to the group it asks for. The `resource_id` field is taken from the request body and served verbatim, so any enrolled agent can fetch any group's merged configuration.

`etc/shared/<group>/merged.mg` is the concatenation of that group's shared folder, including `agent.conf`. In real deployments that file routinely carries secrets such as `wodle name="aws-s3"` access and secret keys, Azure `application_key`, GCP credential paths, and database passwords. A single compromised low value endpoint therefore yields the centralized configuration, and the embedded credentials, of every other group in the deployment.

This is filed as an issue rather than a security advisory because the code is not present in any published release. It is absent from `v5.0.0-beta4` and from every 4.x GA tag, so no deployed user is affected today. The purpose of this report is to prevent it shipping in 5.0.0 GA.

## Affected code

Route registration, `src/remoted/remoted_module/src/remotedModuleFacade.hpp:398-405`:

```cpp
// resource_id is the group (or WPK filename) the agent requests and the manager serves
// exactly that; there is no group lookup and no membership check (protocol decision on
// #38022). Containment therefore rests on the resource-id grammars plus O_NOFOLLOW.
m_authGateway->addAuthenticatedRoute(*m_httpServer,
                                     remoted::http::Method::Post,
                                     "/download",
                                     remoted::endpoints::download::makeHandler({}, m_downloadMetrics),
                                     remoted::http::ResponseMode::Streamable);
```

Resolution, src/remoted/remoted_module/src/endpoints/downloadEndpoint.cpp:326:

LocateResult locateResource(const DownloadRequest& request, const ResourcePaths& paths)

The signature carries no agent identity, so the handler structurally cannot perform a membetId appears in the handler only inside log lines. The header documents the behaviourexplicitly at downloadEndpoint.hpp:196-201:

> The manager does NOT check that the requesting agent belongs to what it asks for: per the protocol decision on #38022, resource_id is what the agent requests and the group lookup was removed. Any authenticated agent can therefore fetch any group's (or multigroup's) merged configuratio

### Data flow

1. downloadEndpoint.cpp:206-268 reads resource_id from the JSON body and validates it againy, with no authorization check.
2. downloadEndpoint.cpp:341-345 joins it into a path: <sharedDir>/<resourceId>/merged.mg for a single group, or <multigroupsDir>/<sha256(selector)[0:8]>/merged.mg for a comma separated multigroup selector.
3. downloadEndpoint.cpp:534-594 opens the resolved file and streams it back to the caller.

### Why this is a regression

Legacy remoted derives the group server side from the authenticated key. In src/remoted/src

`if (OS_SUCCESS == lookfor_agent_group(key->id, data->message, &data->group, wdb_sock)) {`

The agent never names a group. The new endpoint moved that choice to the client and removed

That merged.mg contains agent.conf is confirmed at src/remoted/src/manager.c:1241-1243, whiSource file: %s/agent.conf --> marker.

### Reproduction

Using a valid client.keys entry for any enrolled agent:
```
POST /download
Authorization: Wazuh <id>:<ts>:<mac>

{"resource_type": "config", "resource_id": "<any-other-group>"}
```
The manager returns /var/ossec/etc/shared/<any-other-group>/merged.mg. The 200 versus 404 dovides a clean group name enumeration oracle. The same gap applies to {"resource_type":"wpk", ...} for any package staged in var/upgrade.

### Note on path validation

The containment reasoning in the code is sound for path traversal specifically. isGroupNameChar excludes /, the values . and .. are rejected, and O_NOFOLLOW combined with fstat on the descriptor blocks symlink
substitution. I was not able to escape the base directories. The defect is purely the absen traversal.

### Suggested fix

Resolve the requesting agent's groups server side and reject any resource_id the authenticaof. This requires no new plumbing, because /control already performs exactly that lookup inprocess via m_wdbClient->getAgentGroups(id) at src/remoted/remoted_module/src/control/controlHandler.cpp:169 and :282.

If the motivation for removing the lookup was to keep /download functional while wazuh-db is unavailable, an in memory group cache refreshed by /control would preserve that property while restoring the check.

### Context

Issue #38022 is an implementation task that describes the request and response shape. It contains no authorization analysis. The implementation notes recorded on that issue show the membership check was
written first and then removed:

- 28/07: "trusting the requested group name would let any enrolled agent read another groupesolves from the authenticated agent's own membership, with resource_id only validatedagainst its groups."
- 30/07: "there is now no membership check at all, so any enrolled agent can fetch any grounot just any WPK."

Raising it here so the tradeoff can be re-evaluated before GA rather than after.

matigarciadev:
## Description

`POST /download` authenticated the calling agent but never checked that it was entitled to the resource it asked for. `resource_id` was taken from the request body and served verbatim, so any enrolled agent could fetch **any** group's `etc/shared/<group>/merged.mg` — the concatenation of that group's shared folder, `agent.conf` included, which in real deployments carries integration credentials (AWS keys, Azure `application_key`, GCP credential paths, database passwords). One compromised low-value endpoint therefore yielded the centralized configuration of every other group in the deployment. The 200-vs-404 difference also gave a clean oracle for enumerating group names.

This is a regression against legacy remoted, which derives the group server-side from the authenticated key and never lets the agent name it. The affected code is not in any published release (absent from `v5.0.0-beta4` and every 4.x GA tag), which is why it was reported as an issue rather than an advisory.

A `config` download is now authorized against the requesting agent's own groups, using the selector `/control` already hands that agent as `config_token` — no new plumbing and no wazuh-db round trip on the download path.

Closes #38683

## Proposed Changes

**Authorization (`src/remoted/remoted_module/`)**

- A `config` download is served only when `resource_id` equals the selector the requesting agent's own groups produce. The comparison is exact, multigroup CSV order included, because that order is part of the identity of the merged file `config_hash` was computed over.
- The check runs **before** the resource is resolved, so a request for a group the agent is not in is indistinguishable from one for a group that does not exist — the enumeration oracle is gone.
- An agent whose membership the manager has never established is **denied, not served**. An entry alone is not a membership: `/control/shutdown` creates one with no groups for an agent it has never seen, and an empty group list serializes to the `default` selector, so the source requires a wazuh-db-backed refresh (`groupsRefreshedAtSec != 0`).
- Denials answer `403 {"error":"Forbidden","code":403}` — the flat `ErrorResponse` shape the endpoint already uses — and increment the new `remoted.download.denied` counter. The event itself is logged at debug, because any enrolled agent can trigger it at will; the counter is the operator-facing signal.

**Supporting refactor**

- `toGroupsCsv()` / `makeConfigToken()` promoted out of `controlHandler.cpp`'s anonymous namespace into `control/groupSelector.hpp`, so `/control` and `/download` derive the selector from one implementation and cannot disagree.
- New `IAgentGroupSource` interface plus a registry-backed adapter, wired through the facade. The endpoint layer never sees the concrete `AgentRegistry`.

**Out of scope (deliberate)**

- **WPK downloads are unchanged** and still not authorized by group: a package's authority is the agent's pending upgrade task, which `/control` does not carry. Stock packages are signature verified by the agent against `wpk_root.pem`, which bounds what an unauthorized fetch discloses. Task-based authorization is tracked separately.
- **The agent is not modified.** It receives the new `403` and handles it exactly as today, so a denied download still consumes `download_max_attempts` with backoff per notify. Changing that is a separate PR owned by the agent team.

**Known limitation: the registry is per-node, so a cluster can hold a stale verdict**

The authorization source is remoted's in-memory `AgentRegistry`, which has no cluster awareness: it is filled and refreshed **only** by `/control` requests reaching *that* node, and only once the groups-refresh interval (60 s) has elapsed. wazuh-db is replicated to the workers, but the download path deliberately does not consult it — avoiding that round trip is what makes the check free. Three consequences, understood and accepted for this PR:

| Situation | Effect | Direction |
|---|---|---|
| An agent is removed from a group. The master syncs the new membership to the worker's wazuh-db, but that node's registry still holds the previous groups | The agent can keep downloading its **former** group's configuration until it sends `/control` to that node. On a node it keeps notifying the window is ≤ 60 s; on a node it stops notifying, the entry survives until eviction — up to **6 h** (swept every 5 min) | over-permissive |
| A balancer sends `/control` to worker A and `/download` to worker B. B holds the groups in its replicated wazuh-db, but its registry has no entry for that agent | A **legitimate** download is refused with `403` and the agent does not update its configuration | under-permissive |
| `remoted` restarts. The registry starts empty; the local wazuh-db still holds the groups | `/download` answers `403` until the agent's next `/control` on that node — one notify interval (10 s by default) for an agent already talking to it | under-permissive |

Both under-permissive cases self-heal rather than latch: the agent notifies every `notify_time` seconds (10 s by default), every notify refreshes the registry of the node it lands on, and a failed download is retried after the next notify — so each worker an agent actually talks to becomes populated, and a balancer that keeps an agent on one worker avoids the split entirely. The over-permissive case is bounded by the eviction TTL and is strictly narrower than the behaviour being fixed here, which granted every group to every agent with no time bound at all.

**A separate issue will be opened to analyse this in detail.** The options range from consulting wazuh-db on a registry miss (restoring the round trip this design avoids), through propagating group changes to the workers' registries, to a shorter eviction TTL for entries used only for authorization — none of which belongs in a GA-blocking fix.

### Results and Evidence

All evidence below was produced against a clean local install of this branch. Manager and module are a matched pair (`/var/wazuh-manager/bin/wazuh-manager-remoted` and `lib/libremoted_module.so`, both installed from this branch), `remoted.debug=2`, listener on `127.0.0.1:1517`, global prefix `/wazuh-manager`. `remoted` was restarted immediately before the run so the agent registry started empty.

Agents in `client.keys`: `001`/`002` are live 5.x agents that speak HTTPS `/control` and were notifying throughout; `003`/`004` are 4.x agents that never send HTTPS `/control`, which gives a natural "cold" set with no registry entry. Fixtures added for the run: group `databases` holding a marker secret in its `merged.mg`, and a staged `var/upgrade/probe.wpk`.

#### 1. Authorization matrix

Driven with the module's own harness, `src/remoted/remoted_module/tools/send_download.py` (one request per invocation, `--url https://127.0.0.1:1517`).

**A cold agent is denied — fail closed, no registry entry**

```
$ python3 send_download.py --agent-id 003 --resource-id default
--> POST https://127.0.0.1:1517/wazuh-manager/download  {"resource_type":"config","resource_id":"default"}
<-- 403  chunked=False  content-length=32
    {"error":"Forbidden","code":403}
```

**An entry alone is not a membership** — `/control/shutdown` creates one for an agent the manager has never seen, and an empty group list would otherwise serialize to the `default` selector. This is the fail-open the design guards against with `groupsRefreshedAtSec != 0`:

```
$ python3 send_control.py --type shutdown --agent-id 004
<-- 200 OK
$ python3 send_download.py --agent-id 004 --resource-id default
<-- 403  chunked=False  content-length=32
    {"error":"Forbidden","code":403}
```

`remoted.control.registry.agents` went from 2 to 3 across those two calls: the entry really was created, and the download was still refused.

**A warm agent gets its own configuration**

```
$ python3 send_download.py --agent-id 001 --resource-id default
<-- 200  chunked=True  content-length=None
    100 bytes, sha256=ab0077555b6db8a066e8dc61ce720aba5bb044172b084b5f3ccd3854cae33d3c
    [PASS] body matches the file the manager should have served
```

**Another group is refused, and refused identically whether or not it exists** — this is the enumeration oracle being gone, shown by the two responses being byte-for-byte equal:

```
$ python3 send_download.py --agent-id 001 --resource-id databases        # exists on disk, not this agent's
<-- 403  content-length=32  sha256=4bbe0b12636a233c79c5b946c27d95da237da92c20a4e990b4e907b2b9468d48
$ python3 send_download.py --agent-id 001 --resource-id no-such-group    # does not exist at all
<-- 403  content-length=32  sha256=4bbe0b12636a233c79c5b946c27d95da237da92c20a4e990b4e907b2b9468d48
```

The `databases` marker string (`MERGED-DATABASES-CONFIG-SECRET`) never appeared in any response body.

**WPK downloads are unchanged** — a cold agent with no registry entry still gets the package, as the scope note above states:

```
$ python3 send_download.py --agent-id 003 --resource-type wpk --resource-id probe.wpk
<-- 200  chunked=True  content-length=None
    17 bytes, sha256=e090581cb64dd968ebf5312a48fdb176ce2266e6aaac5dc013c0103da1e5faac
```

#### 2. A real membership change, driven through the manager API

The case unit tests cannot reach: the full chain from an API call to the authorization verdict. Agent `001` was assigned to and then removed from `databases`, while its `config_token` and three download targets were sampled every 20 s.

```
$ TOKEN=$(curl -sk -u wazuh:wazuh -X POST "https://127.0.0.1:55000/security/user/authenticate?raw=true")
$ curl -sk -X PUT -H "Authorization: Bearer $TOKEN" "https://127.0.0.1:55000/agents/001/group/databases"
  All selected agents were assigned to databases
$ curl -sk -H "Authorization: Bearer $TOKEN" "https://127.0.0.1:55000/agents?agents_list=001&select=id,group"
  wazuh-db now says: ['default', 'databases']
```

| Time | `config_token` returned by `/control/notify` | `default` | `default,databases` |
|---|---|---|---|
| 14:55:03 | `default` | 200 | — |
| *14:55:03 — API assigns `databases`; wazuh-db updates immediately* | | | |
| 14:55:11 – 14:55:52 | `default` (registry still stale) | 200 | 403 |
| 14:56:12 (**≈69 s after the change**) | `default,databases` | **403** | **200** |
| *14:57:05 — API removes `databases`* | | | |
| 14:57:32 – 14:57:52 | `default,databases` (stale) | 403 | **404** |
| 14:58:13 (**≈68 s after the change**) | `default` | **200** | 403 |

Three properties come out of this run:

- **Authorization tracks the selector, not "any group the agent belongs to".** Once the selector became `default,databases`, the agent's *former* group `default` was refused with `403`. Exact matching is what keeps `config_hash` and the served file describing the same thing.
- **`/control` and `/download` never disagree.** The token and the verdict flip in the same sample, because both derive from the same registry entry through the same helper. An agent is never left holding a token the manager refuses.
- **The staleness window is ~60 s in both directions**, matching `groups_refresh_interval_sec` and the known limitation documented above — measured here rather than estimated.

The `404` at 14:57:32 is worth calling out: the manager had already removed that multigroup's `merged.mg`, while the registry still authorized the selector, so the request passed authorization and then found no file. That is exactly the intended split between `403` ("not yours") and `404` ("yours, but not on disk"), and it self-healed on the next refresh.

#### 3. Every harness scenario

The harness's scenario list was updated in this PR (`unknown_group_is_404` asserted the old behaviour), so it was run end to end:

```
$ python3 send_download.py --all --agent-id 001 --other-group databases --wpk probe.wpk
agent 001: config selector=default, wpk=probe.wpk
Running 23 scenarios against https://127.0.0.1:1517 (agent 001)
[PASS] valid_config: expected 200, got 200, 100 bytes, chunked=True
[PASS] malformed_not_json / malformed_empty / malformed_array / malformed_missing_field / malformed_extra_field / malformed_wrong_type: expected 400, got 400
[PASS] unknown_resource_type / resource_type_case_sensitive: expected 400, got 400
[PASS] invalid_id_traversal / invalid_id_slash / invalid_id_dotdot / invalid_id_empty / invalid_id_too_long: expected 400, got 400
[PASS] multigroup_selector_trailing_comma / _doubled_comma / _traversal_entry: expected 400, got 400
[PASS] wpk_without_extension / wpk_traversal: expected 400, got 400
[PASS] group_that_does_not_exist_is_denied: expected 403, got 403 -- {"error":"Forbidden","code":403}
[PASS] group_the_agent_is_not_in_is_denied: expected 403, got 403 -- {"error":"Forbidden","code":403}
[PASS] missing_wpk_is_404: expected 404, got 404 -- {"error":"Resource not found","code":404}
[PASS] valid_wpk: expected 200, got 200, 17 bytes, chunked=True

23/23 scenarios passed.
```

#### 4. Metrics and logging

```
$ curl -s --unix-socket /var/wazuh-manager/queue/sockets/remote-admin-http.sock http://localhost/metrics
```

Immediately after the authorization matrix (§1), each outcome landed in its own counter and nothing leaked between them:

| Counter | Value | Corresponds to |
|---|---|---|
| `remoted.download.denied` | 4 | the four denials in §1 |
| `remoted.download.started` | 2 | the config and the WPK served |
| `remoted.download.not_found` | 0 | — |
| `remoted.download.rejected` | 0 | — |
| `remoted.download.open_error` | 0 | — |

Denials are logged at debug and nowhere else, as intended — an enrolled agent can trigger one at will, so the counter is the operator-facing signal:

```
$ grep "Denied a /download" /var/wazuh-manager/logs/wazuh-manager.log | tail -4
2026/09/10 14:46:10 wazuh-manager-remoted:endpoints(download)[195265] downloadEndpoint.cpp:558 at operator()(): DEBUG: Denied a /download request from agent '003' for resource 'default': not its own.
2026/09/10 14:46:10 ... DEBUG: Denied a /download request from agent '004' for resource 'default': not its own.
2026/09/10 14:46:24 ... DEBUG: Denied a /download request from agent '001' for resource 'databases': not its own.
2026/09/10 14:46:25 ... DEBUG: Denied a /download request from agent '001' for resource 'no-such-group': not its own.

$ grep -cE "(WARNING|ERROR).*download" /var/wazuh-manager/logs/wazuh-manager.log
0
```

#### 5. Unit tests

```
$ cmake --build src/build -j22 --target remoted_module_utest
$ src/build/remoted/remoted_module/test/remoted_module_utest --gtest_filter=-ShutdownRace.*
[==========] 838 tests from 105 test suites ran.
[  PASSED  ] 838 tests.
```

`ShutdownRace.*` is excluded from that figure and reported separately: it is a **pre-existing** flake on this branch, unrelated to these changes. Measured on a pristine tree with this work stashed, it failed 3 and hung 2 times out of 8 runs, with the same assertion (`shutdownRace_test.cpp:472`) either way.

**Not re-run after the final rebase**: the 838/838 figure was produced before the branch was rebased onto its current base, and the local build tree is currently configured for the installed manager rather than for the unit-test target, so re-running it would require reconfiguring that tree. No manager source changed between that run and the current HEAD.

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [x] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity — N/A, no Windows-built code is touched
  - [ ] UMDH — N/A, no Windows-built code is touched
- Memory tests for macOS
  - [ ] Leaks — N/A, no macOS-specific code is touched
  - [ ] AddressSanitizer — N/A, no macOS-specific code is touched

  - [ ] Added unit testing files ".ini" — N/A, no ruleset change
  - [ ] `runtests.py` executed without errors — N/A, no ruleset change

  - [ ] Test run in parallel — N/A, no engine code is touched
  - [ ] ASAN for test (utest/ctest) — N/A, no engine code is touched
  - [ ] TSAN for test and wazuh-engine. — N/A, no engine code is touched

- Wazuh server API/Framework
  - [ ] Run API Integration Tests — N/A, the manager API and framework are not touched

### Artifacts Affected

- `libremoted_module.so` — the C++ module that implements the endpoint, built by `src/remoted/remoted_module/` and linked into the remoted daemon (`src/remoted/CMakeLists.txt:62`, `:79`).
- `wazuh-manager-remoted` — the daemon that loads it (`src/remoted/CMakeLists.txt:92`).

No packaging spec, service unit, or default configuration file is touched. Considered and excluded as non-artifacts: `docs/ref/modules/remoted/` (mdBook documentation) and `src/remoted/remoted_module/tools/send_download.py`, a developer harness with no packaging reference.

### Configuration Changes

N/A. No new setting, no default change, and no schema change: the authorization uses data the `/control` registry already holds. The agent-visible wire contract is unchanged — no new request field, and the response gains only a status a client must already tolerate.

### Tests Introduced

`remoted_module_utest`, +22 cases across three files:

- **`test/unit/downloadEndpoint_test.cpp` (+9)** — the endpoint's authorization: own selector served, own multigroup selector served, another group denied, a reordered multigroup CSV denied, an agent with no registry entry denied, a miswired (null) source denied, denials for an existing and a non-existent group asserted byte-identical, denial counted in `denied` and in no other counter, and a WPK still served without any group check. Two pre-existing cases were re-scoped: one that documented the missing check, and the metrics case, whose 404 probe now needs an agent entitled to a group whose file is absent — the 403 it used to get is the new behaviour.
- **`test/unit/registryAgentGroupSource_test.cpp` (+9, new)** — the adapter: own selector, zero-padded agent id, `default` for an agent with no groups, `nullopt` for an unknown agent, a malformed id and a missing registry, parity with the token `/control` hands out, and the two regressions for the `/control/shutdown` entry that has never been refreshed.
- **`test/unit/groupSelector_test.cpp` (+4, new)** — the shared helpers: wdb order preserved, no de-duplication or filtering, and the empty-list-to-`default` substitution.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

### DIFF
--- a/docs/ref/modules/remoted/agent-api.yaml
+++ b/docs/ref/modules/remoted/agent-api.yaml
@@ -584,8 +584,17 @@ paths:
         fetch its **effective** configuration, and needs no database lookup because the directory
         name is the first 8 hex characters of the SHA-256 of the selector verbatim.
 
-        **There is no group-membership check.** Any authenticated agent can fetch any group's or
-        multigroup's merged configuration.
+        **A `config` download is authorized against the requesting agent's own groups.**
+        `resource_id` must equal the selector `/control` handed that agent as `config_token` -- the
+        same string `config_hash` was computed over -- and anything else is `403`. The selector is
+        resolved from the agent registry `/control` maintains, so no wazuh-db round trip is added
+        here; an agent whose membership was never established (no `/control/startup`, or an evicted
+        entry) is denied rather than served. The comparison is exact, multigroup order included.
+
+        **`wpk` downloads are not authorized this way.** A package's authority is the agent's
+        pending upgrade task, which `/control` does not carry, so any authenticated agent may still
+        fetch a staged package; stock packages are signature-verified by the agent against
+        `wpk_root.pem` before installing.
 
         A failed transfer is a truncation, never a short success: the terminating chunk is withheld
         so the agent retries rather than installing an incomplete file. The descriptor is re-`fstat`ed
@@ -625,6 +634,27 @@ paths:
                   summary: Fails the grammar for its type
                   value: { error: "Invalid resource identifier", code: 400 }
         "401": { $ref: "#/components/responses/Unauthorized" }
+        "403":
+          description: |
+            Two different producers answer `403` on this route, with the same flat shape:
+
+            - **this endpoint**, when `resource_type` is `config` and `resource_id` is not the
+              requesting agent's own selector, or the manager has no established group membership
+              for it. Decided before the path is resolved, so it is indistinguishable from a request
+              for a group that does not exist -- there is no oracle for enumerating group names.
+            - **the shared transport**, in mTLS mode, when the client certificate does not match the
+              connecting peer's address (see the note on `POST /enroll`). That one rejects the whole
+              connection, every route included.
+          content:
+            application/json:
+              schema: { $ref: "#/components/schemas/ErrorResponse" }
+              examples:
+                notTheAgentsOwnResource:
+                  summary: The requested selector is not this agent's
+                  value: { error: "Forbidden", code: 403 }
+                peerAddressMismatch:
+                  summary: mTLS only -- rejected by the transport, before any route
+                  value: { error: "Client certificate does not match the peer address", code: 403 }
         "404":
           description: |
             Absent, not a regular file, or `O_NOFOLLOW` refused a symlink -- deliberately
--- a/docs/ref/modules/remoted/https-events-api.md
+++ b/docs/ref/modules/remoted/https-events-api.md
@@ -1336,9 +1336,22 @@ Accepted identifiers:
 - **WPK filename** — non-empty, at most 255 bytes, must **not** begin with a dot, must end in
   `.wpk`, and is drawn from a stricter set (`a-z A-Z 0-9 . _ -`).
 
-> **There is no group-membership check.** Any authenticated agent can fetch any group's or
-> multigroup's merged configuration (protocol decision on #38022). Authentication proves *an* agent
-> is asking; it does not constrain *which* configuration it may ask for.
+> **A `config` download is authorized against the requesting agent's own groups.** `resource_id`
+> must equal the selector `/control` handed that agent as `config_token` — the same string
+> `config_hash` was computed over — and anything else is answered `403`. The manager resolves that
+> selector from the agent registry `/control` already maintains, so the check costs no wazuh-db
+> round trip; an agent whose membership the manager has never established (it never completed
+> `/control/startup`, or its entry was evicted) is **denied, not served**.
+>
+> The comparison is exact, including the order of a multigroup CSV: `a,b` and `b,a` name different
+> merged files, and only one of them is the file `config_hash` refers to.
+>
+> **`wpk` downloads are not covered.** A package's authority is the agent's pending upgrade task,
+> which `/control` does not carry, so any authenticated agent can still fetch any *staged* package.
+> Stock packages are Wazuh-signed and the agent verifies them against `wpk_root.pem` before
+> installing, which bounds what an unauthorized fetch discloses to *which* package is staged —
+> operator-built custom packages are the exception to that reasoning. Closing this properly means
+> task-based authorization and is tracked separately.
 
 Containment differs per form, by design. The multigroup selector is **hashed, never joined**, so it
 cannot traverse by construction. The single-group and WPK forms *do* join agent input into a path, so
@@ -1359,8 +1372,15 @@ instead.
 | Body empty, over 4 KiB, not a JSON object, wrong member count, or a non-string member | `400` | `Invalid request format` |
 | `resource_type` is neither `config` nor `wpk` | `400` | `Invalid resource type` |
 | `resource_id` fails the grammar for its type | `400` | `Invalid resource identifier` |
+| `resource_type: config` and `resource_id` is not the requesting agent's own selector, or the manager has no established membership for it | `403` | `Forbidden` |
 | Resource absent, not a regular file, or `O_NOFOLLOW` rejected a symlink | `404` | `Resource not found` |
 | Unexpected `errno` while opening or stat-ing | `500` | `Internal server error` |
+
+The `403` is decided **before** the path is resolved, so a group the agent is not in reads exactly
+the same whether or not it exists on the manager — there is no 200-vs-404 oracle for enumerating
+group names. Note that a `403` on this route can also come from the shared transport rather than
+this endpoint, in mTLS mode, when the client certificate does not match the connecting peer's
+address; that one carries its own message and applies to every route.
 
 A symlink is deliberately indistinguishable from an absent file to the agent. Auth failures reuse the
 same responses as `/stateless`.
--- a/docs/ref/modules/remoted/metrics.md
+++ b/docs/ref/modules/remoted/metrics.md
@@ -412,6 +412,7 @@ the streaming pump runs; the per-chunk loop is deliberately uninstrumented.
 | Metric | Type | Unit | Meaning | Tuning |
 |---|---|---|---|---|
 | `remoted.download.rejected` | counter | count | 400: the request did not parse | diagnostic |
+| `remoted.download.denied` | counter | count | 403: the agent asked for a `config` selector that is not its own, or the manager has no established group membership for it (no `/control/startup` yet, or an evicted entry) | **the only signal for a denial** — the event itself is logged at debug, so a rising count with `remoted.debug=0` is all an operator sees. Steady non-zero: an agent using a stale `config_token`, or one probing other groups. Distinct from `rejected` (a malformed request) and from `not_found` (an *entitled* request whose file is not on disk yet) |
 | `remoted.download.not_found` | counter | count | 404: the requested group/WPK does not exist — the config-drift signal behind agent retry storms | diagnostic — deploy the missing group/WPK |
 | `remoted.download.open_error` | counter | count | 500: the file exists but could not be opened | diagnostic — filesystem/permissions |
 | `remoted.download.started` | counter | count | Streamed transfers started | [`remoted.max_parallel_connections`](configuration.md#remotedmax_parallel_connections) is the only bound on concurrent transfers |
--- a/src/remoted/remoted_module/README.md
+++ b/src/remoted/remoted_module/README.md
@@ -235,7 +235,8 @@ src/endpoints/
 ├── statefulEndpoint.hpp/.cpp # /stateful policy: opaque inventory-sync sessions, contract passthrough
 ├── statsEndpoint.hpp/.cpp    # /stats policy: forwards to modulesd's inventory sync server
 └── configEndpoint.hpp/.cpp  # /config policy: near-duplicate of statsEndpoint, on purpose
-└── downloadEndpoint.hpp/.cpp  # /download policy: request grammar + resource resolution + file streaming
+├── downloadEndpoint.hpp/.cpp  # /download policy: request grammar + resource resolution + file streaming
+└── iAgentGroupSource.hpp     # interface: the selector an authenticated agent may download
 └── cacertsEndpoint.hpp/.cpp   # GET /cacerts: the CA that signs the listener cert, certificates only (no auth)
 ```
 
@@ -361,6 +362,10 @@ src/control/
 ├── metrics.hpp               # ControlMetrics (remoted.control.* registry handles + wdb latency histogram)
 ├── controlHandler.hpp/.cpp   # ControlHandler: core business logic for all three message types
 ├── agentRegistry.hpp/.cpp    # AgentRegistry: thread-safe sharded map (agent metadata cache + eviction)
+├── groupSelector.hpp         # toGroupsCsv()/makeConfigToken(): the group selector /control hands the
+│                             #   agent as config_token, shared so /download resolves the same string
+├── registryAgentGroupSource.hpp/.cpp # RegistryAgentGroupSource: IAgentGroupSource over the registry
+│                             #   (answers "which selector may this agent download?", nullopt = deny)
 ├── wazuhDBClient.hpp/.cpp    # WazuhDBClient: async UDS client to wazuh-db (agent status/data updates)
 ├── taskClient.hpp/.cpp       # TaskClient: async UDS client to task-manager (pending task fetch)
 ├── mergedMgWatcher.hpp/.cpp  # MergedMgWatcher: inotify + poll watcher for var/multigroups/*.mg changes
--- a/src/remoted/remoted_module/README.md
+++ b/src/remoted/remoted_module/README.md
@@ -234,10 +234,10 @@ src/endpoints/
 ├── statelessEndpoint.hpp/.cpp # /stateless policy: identity check + downstream target + post-processing
 ├── statefulEndpoint.hpp/.cpp # /stateful policy: opaque inventory-sync sessions, contract passthrough
 ├── statsEndpoint.hpp/.cpp    # /stats policy: forwards to modulesd's inventory sync server
-└── configEndpoint.hpp/.cpp  # /config policy: near-duplicate of statsEndpoint, on purpose
-├── downloadEndpoint.hpp/.cpp  # /download policy: request grammar + resource resolution + file streaming
-└── iAgentGroupSource.hpp     # interface: the selector an authenticated agent may download
-└── cacertsEndpoint.hpp/.cpp   # GET /cacerts: the CA that signs the listener cert, certificates only (no auth)
+├── configEndpoint.hpp/.cpp   # /config policy: near-duplicate of statsEndpoint, on purpose
+├── downloadEndpoint.hpp/.cpp # /download policy: request grammar + resource resolution + file streaming
+├── iAgentGroupSource.hpp     # interface: the selector an authenticated agent may download
+└── cacertsEndpoint.hpp/.cpp  # GET /cacerts: the CA that signs the listener cert, certificates only (no auth)
 ```
 
 - **`GET /cacerts` (`cacertsEndpoint.hpp/.cpp`, ns `remoted::endpoints::cacerts`):** the one route
@@ -670,8 +670,10 @@ and clients, which are all thread-safe internally.
    thread runs periodically (every `agentRegistryEvictionIntervalSec`). Wazuh-db and task-manager
    clients maintain persistent connections (reconnect on error).
 3. **Shutdown**: `RemotedModuleFacade::stop()` stops the HTTP server (drains in-flight requests),
-   then stops the wazuh-db and task-manager clients (drains their queues), then destroys the
-   registry (eviction thread stops automatically on destruction).
+   then resets `ControlHandler` — whose destructor stops and joins the eviction thread first, so
+   nothing can touch the registry afterwards, and then drops the wazuh-db and task-manager clients
+   (draining their queues). The registry itself outlives that reset: `/download` shares it to
+   authorize config requests, so it is released later, with the HTTP server's route table.
 
 ## Scan endpoint (`POST /scan/vd`) — `src/scanvd/`
 
@@ -1423,9 +1425,17 @@ before the pump runs; the per-chunk loop is deliberately uninstrumented) — cat
   A `wpk` request names a filename and gets `var/upgrade/<filename>`. The multigroup form is what
   lets an agent in several groups fetch its *effective* configuration rather than one member
   group's, and it needs no database: the selector is hashed exactly as wazuh-db names the directory.
-  There is no group lookup and no membership check (protocol decision on #38022), so **any
-  authenticated agent can fetch any group's or multigroup's merged configuration**. For a `config`
-  request the agent does not pick that value: it relays the `config_token` `/control` handed it (see
+  A `config` request is authorized against the requesting agent's own groups: `resource_id` must
+  equal `makeConfigToken(toGroupsCsv(entry->groups))` for that agent's registry entry -- the same
+  string `/control` handed it as `config_token` -- and anything else is **403**, decided before the
+  path is resolved. The groups come from the registry `/control` already maintains
+  (`control/registryAgentGroupSource.hpp`), so there is no wazuh-db round trip on this path, and an
+  entry whose groups never came from wazuh-db (`groupsRefreshedAtSec == 0`, which is what
+  `/control/shutdown` leaves behind for an unknown agent) is denied rather than treated as
+  "no groups, so default". A `wpk` request is **not** authorized here: its authority is the agent's
+  pending upgrade task, which `/control` does not carry.
+
+  The agent never picks the value it sends: it relays the `config_token` `/control` handed it (see
   the notify response above), so `/control` must report `config_hash` over the file this resolves to
   for the token it handed that agent, or that agent re-downloads on every notify.
 - **Containment differs per form.** The multigroup selector is *hashed, never joined*, so it cannot
@@ -1904,7 +1914,7 @@ linked into the settings' own documentation — is the official docs page:
 | `remoted.http.<stateless\|stateful\|stats\|config\|enroll\|cacerts>.responses.{2xx,400,403,409,413,500,503,other}` | WHAT each endpoint answered agents (some cells structurally zero per endpoint — kept so the vocabulary is uniform; `/cacerts`'s 404 lands in `other`) | the single place each response is sent: the forwarder's delivery task, the limiter-shed 503 in `forward()`, or the handler's own pre-forward 400. `/enroll` and `/cacerts` are not forwarded, so they count through a `MeteredResponder` wrapper instead (`common/requestOutcomeMetrics.hpp`; the description carries the route's method, `GET` for `/cacerts`) — one wrap covers `/enroll`'s five inline answers AND the one authd's callback delivers on another thread |
 | `remoted.http.<stateless\|stateful\|enroll>.latency` (histograms, µs) | end-to-end time; sizes `remoted.http_worker_threads` / `remoted.downstream_stateful_response_timeout` / the `authd_*` timeouts | stamped once in the auth gateway (`AuthenticatedRequest::receivedAt`), observed on the forwarder's post-processing pool. `/enroll` has no gateway, so `MeteredResponder` times it from handler entry. `/stats`/`/config` deliberately have none (same downstream as `/stateful`, no new answer) |
 | `remoted.forwarder.error.{connect, connect_timeout, write_timeout, response_timeout, transport, protocol, response_too_large}` + `downstream_5xx` + `route_mismatch` | WHY the 503s: which timeout knob, transport vs protocol, a downstream 5xx, or a route contract mismatch. Aggregate across services — the per-endpoint 503 cells already say which path | the forwarder's classification branches, next to the throttles that log the same cause |
-| `remoted.download.{rejected, not_found, open_error, started, bytes.total}` | group/WPK drift (404 retry storms) and offered transfer volume | `downloadEndpoint` admission + stream start (the per-chunk pump is deliberately uninstrumented) |
+| `remoted.download.{rejected, denied, not_found, open_error, started, bytes.total}` | group/WPK drift (404 retry storms) and offered transfer volume, plus `denied` — the 403 authorization signal (`resource_id` is not the requesting agent's own selector, or the manager has no established membership for it). It is the ONLY operator-facing signal for a denial, since the event itself is logged at debug; distinct from `rejected` (malformed request) and from `not_found` (an *entitled* request whose file is not on disk) | `downloadEndpoint` admission + stream start (the per-chunk pump is deliberately uninstrumented) |
 | `remoted.cacerts.{served, not_found, ca_mismatch}` | WHY `GET /cacerts` answered what it did: CA handed out, no CA file to hand out, or refused because the configured CA does not sign the served leaf | `cacertsEndpoint` (`endpoints/cacertsMetrics.hpp`), one counter per branch |
 | `remoted.server.tls.{cert_expiry_days, ca_matches_leaf}` (pulls; `cert_expiry_days` is the catalog's one **Double**, via `registerPullMetricDouble()` — negative once expired) | is the listener certificate about to expire; does `remote.https.ca_certificate` sign it (0 also when the CA is unreadable) | `IHttpServer::certificateStatus()` over the transport's `TlsCertificateMonitor` snapshot (start + every 24 h); registered by `registerPublicTransportDiagnostics()` on the same weak target as the budget pulls, so both read 0 while the listener is down |
 | `remoted.server.budget.{available.bytes, inflight.bytes, inflight.requests, rejected.total}` (pulls) | is `remoted.max_inflight_bytes` sized right; how much did the byte budget shed | `IHttpServer::diagnostics()` over the transport's `InFlightBudget` |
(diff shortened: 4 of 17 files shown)


NahuFigueroa97 (approved):
LGTM!

**A — the agent's own group is served.**

```
--> POST https://127.0.0.1:1517/wazuh-manager/download  {"resource_type":"config","resource_id":"web-servers"}
<-- 200  chunked=True  content-length=None
    142 bytes, sha256=e1d1886e8f127f1726ee4343580676535bb79c5695face4a18c516e891877061
    expected file: /var/wazuh-manager/etc/shared/web-servers/merged.mg
    [PASS] body matches the file the manager should have served
```

**B — another group's `merged.mg` is refused.** This is the report's PoC.

```
--> POST https://127.0.0.1:1517/wazuh-manager/download  {"resource_type":"config","resource_id":"databases"}
<-- 403  chunked=False  content-length=32
    32 bytes, sha256=4bbe0b12636a233c79c5b946c27d95da237da92c20a4e990b4e907b2b9468d48
    {"error":"Forbidden","code":403}
```

What it did **not** receive:

```
$ cat /var/wazuh-manager/etc/shared/databases/merged.mg
#databases
!198 agent.conf
<agent_config>
  <!-- databases: stands in for the credential-bearing agent.conf from the report -->
  <labels><label key="secret">AWS_SECRET_ACCESS_KEY_PLACEHOLDER</label></labels>
</agent_config>
```

**C — no enumeration oracle.** A group that does not exist answers byte-for-byte the same as one
that exists but is not the agent's — same status, same length, same digest as B:

```
--> POST https://127.0.0.1:1517/wazuh-manager/download  {"resource_type":"config","resource_id":"a-group-that-does-not-exist"}
<-- 403  chunked=False  content-length=32
    32 bytes, sha256=4bbe0b12636a233c79c5b946c27d95da237da92c20a4e990b4e907b2b9468d48
    {"error":"Forbidden","code":403}
```

**D — fail closed on a registry miss.** remoted restarted, no `/control/startup` sent. The agent
asks for **its own** group and is still denied, because the manager has no established membership
to authorize against — the case `send_download.py`'s own comment says the script cannot stage by
itself:

```
--> POST https://127.0.0.1:1517/wazuh-manager/download  {"resource_type":"config","resource_id":"web-servers"}
<-- 403  {"error":"Forbidden","code":403}
```

**E — and it self-heals.** Same agent, same request, after one `/control/startup`:

```
<-- 200 OK  {"agent":{"groups":["web-servers"]}, ...}

--> POST https://127.0.0.1:1517/wazuh-manager/download  {"resource_type":"config","resource_id":"web-servers"}
<-- 200  142 bytes, sha256=e1d1886e...
    [PASS] body matches the file the manager should have served
```

**Operator-facing signal**, read from the admin socket after the sequence above:

```
remoted.download.denied            1
remoted.download.started           1
remoted.download.bytes.total       142
remoted.download.not_found         0
remoted.download.rejected          0
remoted.download.open_error        0
```

The denial is counted in its own cell, `142` matches the one file legitimately served, and nothing
landed in `rejected` or `not_found`.

**The rest of the endpoint is unchanged.** `send_download.py --all`: 21/23. The 19 grammar and
traversal scenarios still answer exactly as before, and the two failures are the script's own
expectations, not the manager's behaviour — `valid_wpk` has no package staged (404 is correct), and
`valid_config` is nit 1 below.

## Unit suite

```
[==========] 895 tests from 111 test suites ran. (148320 ms total)
[  PASSED  ] 895 tests.
```

Negative control, to show the new cases are not vacuous: disabling only the authorization block
(`if (false && downloadRequest.type == ResourceType::Config)`) and rebuilding brings the defect
straight back —

```
[ RUN      ] DownloadHandlerTest.DeniesAnotherGroupsSelector
downloadEndpoint_test.cpp:999:  Which is: 200
downloadEndpoint_test.cpp:1000: Which is: "SECRET aws access key\n"
downloadEndpoint_test.cpp:1001: Value of: responder->streamed
                                  Actual: true
[  FAILED  ] DownloadHandlerTest.DeniesAnotherGroupsSelector

[ RUN      ] DownloadHandlerTest.DeniedRequestsAreIndistinguishableWhetherTheGroupExists
downloadEndpoint_test.cpp:1049: Which is: 200   <- a group the agent is not in, that exists
                                Which is: 404   <- a group that does not exist
[  FAILED  ] ...
[  FAILED  ] 6 tests
```

Restored, `git status --porcelain` empty, and the 38 tests across `DownloadHandlerTest`,
`DownloadLocateTest`, `GroupSelectorTest` and `RegistryAgentGroupSourceTest` pass again.

## Nitpicks, none blocking

1. **`send_download.py --all` now fails `valid_config` out of the box.** Line 558 is
   `default_group = args.resource_id or "default"` — a hardcoded `"default"` — while `--help` at
   line 512 says *"Defaults to the agent's first group"*. Before this PR the discrepancy was
   invisible, because `/download` served whatever was named; now any agent that is not in `default`
   gets a `403` for the scenario meant to prove the happy path, and the sheet reads 21/23 on a
   correct manager. Either resolve the agent's real selector or make the help match the code.
2. **The docs do not say what an agent should do on a `403`.** They describe when it happens
   precisely, but not that recovery is to re-run `/control/startup` and use the `config_token` it
   returns. Without that, an agent implementer may retry the download in a loop. One sentence in
   `https-events-api.md`.
3. **A remoted restart denies every agent until it completes `/control/startup`** — scenario D
   above. Correct fail-closed behaviour, and self-healing, but the docs only mention "never
   completed `/control/startup`, or an evicted entry". Naming the manager restart explicitly would
   help, since that is the case an operator will actually see behind `remoted.download.denied`.

LucioDonda (approved):
## LGTM

But a possible issue on a separated comment

## 1. Authorization matrix

| # | Case | Agent | `resource_id` | Expected | Got | |
|---|---|---|---|---|---|---|
| T1 | Cold agent, no registry entry | 003 | `default` | 403 | **403** | ✅ |
| T1b | Cold agent, someone else's group | 003 | `databases` | 403 | **403** | ✅ |
| T2 | `/control/shutdown` creates the entry | 004 | — | 200, registry 0→1 | **200, 0→1** | ✅ |
| T2b | An entry is not a membership | 004 | `default` | 403 | **403** | ✅ |
| T3 | Its own selector | 001 | `default` | 200, byte-exact | **200, sha256 matches** | ✅ |
| T4a | A group that **exists**, but is not its own | 001 | `databases` | 403 | **403** | ✅ |
| T4b | A group that **does not exist** | 001 | `no-such-group` | 403 | **403** | ✅ |
| T6 | WPK from a **cold** agent | 003 | `probe.wpk` | 200 (out of scope) | **200** | ⚠️ by design |

### T1/T2 — fail-closed confirmed

`registry.agents` went from **0 to 1** on 004's `/control/shutdown` — the entry really was created —
and the download was still denied. The `groupsRefreshedAtSec != 0` guard does its job: without it,
`makeConfigToken("")` would have returned `default` and the agent would have minted itself a
permission.

### T4 — the enumeration oracle is gone

The two responses are **byte-for-byte identical**, not merely the same status code:

```
databases      -> 403  content-length=32  sha256=4bbe0b12636a233c79c5b946c27d95da237da92c20a4e990b4e907b2b9468d48
no-such-group  -> 403  content-length=32  sha256=4bbe0b12636a233c79c5b946c27d95da237da92c20a4e990b4e907b2b9468d48
```

The `MERGED-DATABASES-CONFIG-SECRET` marker **never** appeared in any response body or in the logs.

### T6 — the WPK residual, demonstrated

An agent with **no registry entry at all** gets the package with no check whatsoever:

```
$ send_download.py --agent-id 003 --resource-type wpk --resource-id probe.wpk
<-- 200  chunked=True   16 bytes, sha256=9c9a57c4…
```

This is the PR's declared scope, but it is worth having it measured rather than only written down.

---

## 2. Full harness — 23/23

```
$ send_download.py --all --agent-id 001 --other-group databases --wpk probe.wpk
agent 001: config selector=default, wpk=probe.wpk
Running 23 scenarios against https://127.0.0.1:1517 (agent 001)
...
[PASS] group_that_does_not_exist_is_denied: expected 403, got 403
[PASS] group_the_agent_is_not_in_is_denied: expected 403, got 403
[PASS] missing_wpk_is_404: expected 404, got 404
[PASS] valid_wpk: expected 200, got 200, 16 bytes, chunked=True

23/23 scenarios passed.
```

**No `--other-group` WARNING**, which means the "exists, but not mine" scenario really was
exercised (the `databases` group and its `merged.mg` had to be created first; on a stock install
that scenario is skipped).

---

## 3. Metrics — clean isolation

After T1–T5:

| Counter | Value | Corresponds to |
|---|---|---|
| `remoted.download.denied` | **7** | T1 + T1b + T2b + T4a + T4b + the harness's 2 |
| `remoted.download.rejected` | **18** | the harness's 18 400-scenarios |
| `remoted.download.not_found` | **1** | `missing_wpk_is_404` |
| `remoted.download.started` | **3** | T3 + `valid_config` + `valid_wpk` |
| `remoted.download.bytes.total` | **216** | 100 + 100 + 16 |
| `remoted.download.open_error` | **0** | — |
| `remoted.control.registry.agents` | **2** | 001 + 004 |

Every outcome landed in its own counter, with no cross-contamination. The 7 `denied` match the 7
debug lines exactly.

## 4. Logging

The 7 denials appear at DEBUG and **only** at DEBUG:

```
20:01:35 ... downloadEndpoint.cpp:558: DEBUG: Denied a /download request from agent '003' for resource 'default': not its own.
20:01:35 ... DEBUG: Denied ... agent '003' ... resource 'databases'
20:01:46 ... DEBUG: Denied ... agent '004' ... resource 'default'
20:02:24 ... DEBUG: Denied ... agent '001' ... resource 'databases'
20:02:24 ... DEBUG: Denied ... agent '001' ... resource 'no-such-group'
20:02:34 ... DEBUG: Denied ... agent '001' ... resource 'a-group-that-does-not-exist'
20:02:34 ... DEBUG: Denied ... agent '001' ... resource 'databases'
```

Zero `WARNING`/`ERROR` from the `/download` path (the search's only match was an
`IndexerDownloader` line from `modulesd`, timestamped 18:48 and unrelated).

This confirms **L-03**: with `remoted.debug=0` (the production default) the operator is left with a
counter and **no agent id**. There is no way to investigate who is probing other groups.

---

## 5. A real membership change, driven through the manager API

Agent `001` assigned to and then removed from `databases`, sampling `config_token` and the three
download targets every ~20 s.

| Time | `config_token` from `/control/notify` | `default` | `default,databases` | `databases` |
|---|---|---|---|---|
| 20:03:44 (baseline) | `default` | **200** | 403 | 403 |
| *~20:03:50 — the API assigns `databases`* | | | | |
| 20:04:04 | `default` (stale) | 200 | 403 | 403 |
| 20:04:25 | `default` (stale) | 200 | 403 | 403 |
| **20:04:46** (**≈56 s after the change**) | **`default,databases`** | **403** | **200** | 403 |
| 20:05:06 / 20:05:27 | `default,databases` | 403 | 200 | 403 |
| *~20:05:35 — the API removes `databases`* | | | | |
| **20:05:47** | **`default`** | **200** | **403** | 403 |
| 20:06:09 … 20:07:12 | `default` | 200 | 403 | 403 |

Four properties come out of this run:

1. **Authorization tracks the *selector*, not "any group the agent belongs to".** As soon as the
   selector became `default,databases`, the agent's *former* group `default` answered **403**. That
   is what keeps `config_hash` and the served file describing the same thing.
2. **`databases` on its own never returned 200**, in any sample. Correct: the selector is the
   multigroup, not its parts.
3. **`/control` and `/download` never disagreed** in this run: the token and the verdict flip in the
   **same sample**, because both derive from the same entry through the same helper.
4. **The staleness window is ≈60 s**, matching `groupsRefreshIntervalSec`. The flip back looks like
   12 s, but it is not: the refresh cycle had been running since 20:04:46, so it expired at 20:05:46
   — right on the sample after the `DELETE`.

> **Difference from the PR's evidence**: the PR shows a transient `404` when the group is removed
> (the registry still authorizing the selector, the multigroup's `merged.mg` already deleted). It did
> **not** appear here: the sampling landed after the refresh and never caught that window. This does
> not contradict the PR — it only means that window is narrower than a 20 s sampling interval is
> guaranteed to catch.

---

LucioDonda:
# Additional scenario to check:

## step-by-step reproduction: `/control` and `/download` **do** disagree

The PR states in its description:

> **`/control` and `/download` never disagree.** The token and the verdict flip in the same sample,
> because both derive from the same registry entry through the same helper. **An agent is never left
> holding a token the manager refuses.**

This disproves it. Reproduced live.

### The asymmetry

| Side | Code | Applies the `groupsRefreshedAtSec != 0` guard? |
|---|---|---|
| `/download` | `RegistryAgentGroupSource::expectedSelectorFor` — `registryAgentGroupSource.cpp:71` | **Yes** |
| `/control` | `processNotify()` — `controlHandler.cpp:526` | **No** |

Both call `makeConfigToken(toGroupsCsv(entry->groups))`, but the helper takes a **CSV**, not the
entry, so the guard is optional for whoever calls it. One side applies it, the other does not.

### Step 0 — precondition

Agent `004`: entry created by `/control/shutdown` (T2), **never refreshed from wazuh-db**
(`groups={}`, `groupsRefreshedAtSec=0`). Confirmed: `registry.agents = 2` (001 and 004), and
`/download` already denies it.

### Step 1 — take wazuh-db down

```
$ kill -TERM 801996                                   # wazuh-manager-db
$ ls -la /var/wazuh-manager/queue/sockets/wdb.sock
   No such file or directory                          # the socket is gone
```

### Step 2 — `/control/notify` for 004 with wdb down

```
$ send_control.py --type notify --agent-id 004
<-- 200 OK
    {"agent":{"config_hash":"0","config_token":"default","groups":[]}, ...}
```

Three things at once:

- **`200 OK`**: the request does not fail. `handleNotify` does not take its `500` path because
  `entry` exists; it only skips the refresh.
- **`config_token: "default"`**: `makeConfigToken("")` substitutes the empty list. The agent is
  handed a token the manager is going to refuse.
- **`config_hash: "0"`**: since `groupsCsv` is `""`, `getMergedMgPath("")` resolves nothing. `"0"` is
  a guaranteed mismatch against any local hash, so the agent does not merely *get to* download — it
  is **forced** to.

### Step 3 — the download that token orders

```
$ send_download.py --agent-id 004 --resource-type config --resource-id default
--> POST /wazuh-manager/download  {"resource_type":"config","resource_id":"default"}
<-- 403  {"error":"Forbidden","code":403}
```

Repeated 3 times: `403`, `403`, `403`. And in the manager log, one denial line per attempt:

```
20:12:24 ... DEBUG: Denied a /download request from agent '004' for resource 'default': not its own.
20:12:24 ... (x3)
20:12:25 ...
```

The agent asks for **exactly** the `config_token` `/control` just handed it, and the manager refuses
it. This is literally "left holding a token the manager refuses".

### Step 4 — control case: this is not "wdb down breaks everything"

Same wdb outage, agent `001` (which **does** have `groupsRefreshedAtSec != 0`):

```
$ send_control.py --type notify --agent-id 001
    {"agent":{"config_hash":"ab0077555b6db8a066e8dc61ce720aba5bb044172b084b5f3ccd3854cae33d3c",
              "config_token":"default","groups":["default"]}, ...}
$ send_download.py --agent-id 001 --resource-id default
<-- 200   [PASS] body matches the file the manager should have served
```

| Agent | `groupsRefreshedAtSec` | `/control` says | `/download` says | |
|---|---|---|---|---|
| 001 | ≠ 0 | token=`default`, hash=`ab0077…` | **200** | ✅ they agree |
| 004 | **0** | token=`default`, hash=**`0`** | **403** | ❌ **they disagree** |

With wdb down, the healthy agent keeps working perfectly. The failure is **specific** to the guard
asymmetry, not to the database being down.

### Step 5 — recovery

Manager restarted, wdb back up, `001` returns to a byte-exact `200`. The state does self-heal, but
only once wdb returns: for as long as the outage lasts, the affected agent is stuck in a 403 loop.

### Why it matters, and what fixes it

The trigger (a wdb outage plus an entry created by `shutdown`) is not an everyday event, and the
impact is that **one** agent does not update its configuration while the database is down. The
serious part is not the impact: it is that **the invariant the whole design rests on is not enforced
by the code**. The PR reasons that the two sides cannot disagree because they share a helper; they
do share the helper, and they disagree anyway, because the guard lives outside it.

**Fix**: have `makeConfigToken()` take the `AgentEntry` (or a `std::optional<std::string>` with the
guard already applied) instead of the CSV. That makes the guard non-optional, so the next change to
`processNotify` cannot silently reopen the hole.

PR #39146, opened 2026-09-10
Fix macOS syscollector process and port inventory truncation

### LINKED ISSUE #39126: macOS syscollector silently reports only kern.maxproc/4 processes because proc_listallpids is passed a process count instead of a buffer size in bytes
## Description

On macOS the agent's process inventory is silently truncated. The host runs about 550 processes, but syscollector collects exactly **500**, both in the agent's own local database and in the indexer. No error or warning is logged, and the evaluation reports success.

The cap is not a coincidence and not a configured limit. `proc_listallpids` takes its second argument as a **buffer size in bytes**, but `src/data_provider/src/sysInfoMac.cpp` passes `maxProc`, which is the process *count* read from `kern.maxproc`:

```cpp
const auto spPids         { std::make_unique<pid_t[]>(maxProc) };
const auto processesCount { proc_listallpids(spPids.get(), maxProc) };

for (int index = 0; index < processesCount; ++index)
{
    ProcessTaskInfo taskInfo{};
    const auto pid { spPids.get()[index] };
    ...
}
```

The buffer is correctly allocated as `maxProc * sizeof(pid_t)` bytes, but only `maxProc` bytes are declared to the kernel. `proc_listallpids` therefore fills at most `maxProc / sizeof(pid_t)` entries and returns that count, so the effective ceiling is `kern.maxproc / 4`.

On the host under test the arithmetic matches the observation exactly:

```
kern.maxproc      : 2000
sizeof(pid_t)     : 4
maxproc/4         : 500      <-- exactly the number collected
actual processes  : 549-560  (stable across samples)
```

Between 9% and 11% of running processes are therefore invisible on this host, and the shortfall grows on any macOS host whose real process count exceeds `kern.maxproc / 4`. Which processes are lost is not predictable: it is whatever `proc_listallpids` happens to place in the truncated fill, so the inventory is missing an arbitrary subset rather than, say, the newest entries. For a security product this is a detection blind spot, and it also skews anything downstream that correlates against process inventory.

The same defect appears at a second call site in the same file, in the code that walks each process's file descriptors to build the socket and port inventory:

```cpp
auto pids { std::make_unique<pid_t[]>(maxProcess) };
const auto processesCount { proc_listallpids(pids.get(), maxProcess) };
```

Sockets owned by processes beyond the truncation point cannot be seen by that path. The macOS agent reports 7 ports against 13 real TCP sockets on the host, which is consistent with this, but I did not isolate the port undercount to this root cause, so treat that as a likely secondary impact rather than a confirmed one.

## Environment

- Agent under test: `demo-env-macos-agent`, id `006`, macOS 26.5.1 (build 25F80), arm64, Apple M1 (Virtual), 2 vCPU, 4 GiB RAM, Parallels VM on MacStadium host `macstadium_arm_4`.
- Agent `VERSION.json`: `{"version": "5.0.0", "stage": "rc1", "commit": "2161e69"}`.
- Source inspected on the `5.0.0` branch of `wazuh/wazuh`, `src/data_provider/src/sysInfoMac.cpp`.
- Control agents: `demo-env-linux-deb-amd64-agent` (002), `demo-env-linux-deb-arm64-agent` (001), `demo-env-linux-rpm-amd64-agent` (003), `demo-env-linux-rpm-arm64-agent` (004), `demo-env-windows-agent` (005).

## Steps to reproduce

1. Enroll a 5.0.0 macOS agent on a host whose running process count exceeds `kern.maxproc / 4`. Confirm the ceiling and the real count:

```bash
sysctl -n kern.maxproc                    # 2000 here, so the ceiling is 500
ps -A | tail -n +2 | wc -l                # 549 here
```

2. Force a fresh inventory evaluation and wait for it to finish:

```bash
sudo /Library/Ossec/bin/wazuh-control restart
grep "syscollector" /Library/Ossec/logs/ossec.log | tail -5
# ... INFO: Starting evaluation.
# ... INFO: Evaluation finished.
```

3. Count what the agent actually stored locally:

```bash
sudo sqlite3 /Library/Ossec/queue/syscollector/db/local.db \
     "select count(*) from dbsync_processes;"
```

4. Compare against the indexer, aggregating the process state index by agent:

```bash
  "https://<indexer>:9200/wazuh-states-inventory-processes/_search?size=0" \
  -d '{"aggs":{"a":{"terms":{"field":"wazuh.agent.name","size":20}}}}'
```

5. Cross-check the other syscollector tables on the same agent against the host, to confirm the fault is specific to processes:

```bash
for t in dbsync_users dbsync_groups dbsync_network_iface; do
  printf "%s " "$t"; sudo sqlite3 /Library/Ossec/queue/syscollector/db/local.db "select count(*) from $t;"
done
dscl . -list /Users | wc -l ; dscl . -list /Groups | wc -l ; ifconfig -l | wc -w
```

## Expected

Syscollector enumerates every running process, so the count in `dbsync_processes` and in `wazuh-states-inventory-processes` tracks the host's real process count. If enumeration ever is truncated, it is reported rather than silent.

## Actual

The count is pinned at exactly `kern.maxproc / 4` and does not move when a fresh full evaluation runs against a stable host:

```
host ps BEFORE :      551
db rows BEFORE : 500
host ps AFTER  :      551
db rows AFTER  : 500
--- last syscollector lines ---
2026/09/10 06:55:57 wazuh-modulesd:syscollector: INFO: Started (pid: 82711).
2026/09/10 06:55:57 wazuh-modulesd:syscollector: INFO: Starting evaluation.
2026/09/10 06:56:00 wazuh-modulesd:syscollector: INFO: Evaluation finished.
```

The indexer agrees with the agent, and every other platform sits below the truncation point and so matches its host:

```
  macos-agent                             500     <-- host has ~550
  linux-deb-arm64-agent                   143
  linux-rpm-arm64-agent                   134
  linux-deb-amd64-agent                   116
  linux-rpm-amd64-agent                   102
  windows-agent                            77
```

Every other syscollector table on the same macOS agent matches the host exactly, which rules out a general collection or sync fault and isolates the defect to process enumeration:

```
  dbsync_users        134   vs  dscl . -list /Users   134
  dbsync_groups       163   vs  dscl . -list /Groups  163
  dbsync_network_iface 11   vs  ifconfig -l            11
  dbsync_processes    500   vs  ps -A                 549   <-- only mismatch
```

Nothing configures a 500 limit. `DEFAULT_SYSCOLLECTOR_PROCESSES_LIMIT` in `src/shared/include/module_limits.h` is `0` (unlimited), the manager's `processes_limit` in `src/remoted/src/config.c` defaults to `30000`, and neither the manager's `wazuh-manager-internal-options.conf` nor the agent's `internal_options.conf` / `local_internal_options.conf` sets any syscollector limit.

One confirmation I could not perform: `kern.maxproc` is read-only at runtime on macOS 26 (`sysctl -w kern.maxproc=4000` returns `Invalid argument`), so I could not vary the ceiling and watch the collected count follow it. The evidence for the mechanism is therefore the exact arithmetic match, the source code at both call sites, deterministic reproduction across a forced rescan, and the cross-platform and cross-table controls above.

## Not a duplicate of

`qa-known-issues.py check` scored two cache candidates, and the live search returned nothing:

- `wazuh/wazuh-qa-automation#8762` "Syscollector VD synchronization failed" (0.50) is the transient vulnerability-detection sync warning emitted once per agent at first enrollment, which recovers on the next cycle. It lives in the QA automation repo, concerns the VD sync handshake rather than process enumeration, and has no bearing on how many processes are collected.
- `wazuh/wazuh#4605` "The Syscollector module uses and reports localtime instead of UTC" (0.43) is a timestamp timezone defect in syscollector output. Same module, but a different field, a different platform scope, and no relation to truncated enumeration.
- Live search for `proc_listallpids syscollector kern.maxproc` returned nothing open.

MAnDumu:
## Description

On macOS, syscollector's process inventory is silently truncated (and, as a consequence, most listening ports are dropped too — see QA's follow-up comment on the issue). `src/data_provider/src/sysInfoMac.cpp` calls the macOS API `proc_listallpids(void *buffer, int buffersize)` at two call sites (`getProcessesSocketFD` and `SysInfo::getProcessesInfo`), passing `kern.maxproc` (a **process count**, read via `sysctlbyname`) as `buffersize`, when that argument is documented and implemented by Apple's own `libproc.c` as the buffer size **in bytes**. The kernel therefore only fills `kern.maxproc / sizeof(pid_t)` entries, so the effective process ceiling is exactly a quarter of `kern.maxproc`. Since the port inventory is built by walking each *enumerated* process's file descriptors, and PID 1 (`launchd`, owner of most listening sockets on macOS via socket activation) falls outside the truncated range on the reporter's host, the observed secondary effect is that the entire listening-port inventory disappears rather than being merely undercounted.

Fixed both `proc_listallpids` call sites to pass `maxProc * sizeof(pid_t)` (bytes) instead of `maxProc` (count) as `buffersize`. The receiving array was already allocated with the correct number of elements; only the size declared to the kernel API was wrong. No change was needed to how the function's return value is used (`processesCount`, the loop bound): `proc_listallpids` (unlike the lower-level `proc_listpids`) already normalizes its return value to a pid count internally (`return numpids/sizeof(int)`, unconditionally, per Apple's open-source `Libc/darwin/libproc.c`), so it does not need dividing again. Added a `CHANGELOG.md` entry under Agent / Fixed.

Confirmed this is a genuine, long-standing bug (present since the original 2020/2021 implementation, identical on this branch) and not an intentional limit, before fixing it: `git blame` traces both call sites to the original feature commits, with no later change or review comment documenting an intentional cap; `DEFAULT_SYSCOLLECTOR_PROCESSES_LIMIT` defaults to `0` (unlimited) and the manager's `processes_limit` is an unrelated mechanism. Also verified `proc_listallpids`'s return-value semantics directly against Apple's open-source `Libc/darwin/libproc.c` (3 independent mirrors, including XNU snapshots from 2013 and ~2021 with identical code) to confirm the fix needs no further change to the loop bound.

**Verified end-to-end on a real macOS 26.5.1 (build 25F80, arm64) host** — the same OS build as the original report. Built this branch there with `make deps -C src TARGET=agent` + `make -C src build_syscollector TARGET=agent` and ran `sysinfo_test_tool --processes`/`--ports` directly:

- With the fix: process count matches the host's `ps -A` count exactly (554 = 554), including PID 1 (`launchd`). All 3 real listening TCP sockets on the host (22 and 5900 owned by `launchd`/PID 1, 88 owned by `kdc`/PID 141) are reported with the correct owning PID, matching `lsof -iTCP -sTCP:LISTEN` exactly.
- Reverting just the fix (rebuild, same host) reproduces the original bug precisely: process count pinned at 500 (`kern.maxproc / sizeof(pid_t)` = 2000/4) despite 554 real processes, lowest retained PID 98 (matching the report's own observation), PID 1 and 141 both missing, and 0 listening ports reported — reproducing "the agent reports zero listening ports" from the issue's follow-up comment exactly.
- Re-applied the fix and rebuilt again to confirm the host was left in the correct, fixed state.

See Logs/Alerts example below for the actual before/after output.

**Independently reproduced end-to-end through a full deployed stack** by another tester, on a separate macOS 26.6.2 (build 25G83, arm64) host loaded to 815-830 processes (well past the 500 ceiling), against a real Wazuh 4.14.5 manager/indexer/dashboard (`wazuh-docker` single-node). Two agents were enrolled against the same manager for a direct comparison: agent `002` running the currently released `wazuh-agent-4.14.5-1.arm64.pkg` (before), agent `003` running a source build of this branch at its current head commit (`1d7f4e0587`, after). Querying the indexer through the dashboard's Processes inventory confirms the same mechanism end-to-end, not just at the local binary level:

- Before (released package): 500 processes recorded against 830 real (`ps -A`) — exactly `kern.maxproc / sizeof(pid_t)`. Lowest recorded PID is 4987; PID 1 (`launchd`) is absent. Only 2 listening ports recorded (both `cupsd`); `launchd`'s real 22 and 5900 are missing.
- After (this branch, source build): 813 processes recorded against 815 real — the 2-process gap is consistent with normal churn between the scan and the `ps` snapshot, not truncation. PID 1 (`launchd`) is present. 8 listening ports recorded, including 22 and 5900 owned by `launchd`.

This corroborates the verification above on a third, independent host, through the full agent → manager → indexer → dashboard pipeline rather than the standalone `sysinfo_test_tool` binary, and against a different real 4.x manager version (4.14.5).

**Reproduced again using only the shipped agent, not the `sysinfo_test_tool` dev binary**, per review feedback asking for evidence from the actual `syscollector` → `local.db` database. On the same macOS 26.5.1 (build 25F80, arm64) host, loaded past the ceiling with extra background processes, installed the real agent from source (`./install.sh`, unattended) and enrolled it against a real Wazuh 5.0.0 manager (legacy `remoted`/`authd` on ports 1514/1515, which 5.0.0 keeps open for pre-5.0.0 agents) as agent `002`. Let `wazuh-modulesd:syscollector` run its normal evaluation, then queried the actual production database directly:

- Before (fix reverted, rebuilt, redeployed into the same install): `select count(*) from dbsync_processes;` → `500` against 758 real (`ps -A`). `select pid,name from dbsync_processes where cast(pid as integer)=1;` → empty (no `launchd`). Lowest recorded PIDs: 301, 381, 457, 460, 464.
- After (fix restored, rebuilt, redeployed): `select count(*) from dbsync_processes;` → `776`, matching `ps -A`'s `776` exactly. The same PID-1 query returns `1|launchd`.

Note on scope: this manager's `wazuh-states-inventory-*` indices had zero documents for every agent during this test, including its own native 5.0.0 self-agent (`001`) — not specific to this fix or to agent `002`. That looks like a gap in this particular manager's inventory-sync-to-indexer pipeline in this environment, so the dashboard/indexer view couldn't be used for this round; the `local.db` query above is the direct evidence from the real agent database instead.

## Configuration options

N/A

## Logs/Alerts example

Host reality (real-time, alongside each test run):
```
$ ps -A | wc -l
     554
$ ps -p 1 -o pid,comm
  PID COMM
    1 /sbin/launchd
$ sudo lsof -iTCP -sTCP:LISTEN -P -n
launchd   1 root  ... TCP *:22 (LISTEN)
launchd   1 root  ... TCP *:22 (LISTEN)
launchd   1 root  ... TCP *:5900 (LISTEN)
launchd   1 root  ... TCP *:5900 (LISTEN)
kdc     141 root  ... TCP *:88 (LISTEN)
kdc     141 root  ... TCP *:88 (LISTEN)
```

With the fix (`sysinfo_test_tool --processes` / `--ports`):
```
count: 554
has pid 1: True
total ports: 24
listening: 6
  22   1   (launchd)
  22   1   (launchd)
  5900 1   (launchd)
  5900 1   (launchd)
  88   141 (kdc)
  88   141 (kdc)
```

With the fix reverted (2 lines, rebuilt on the same host, same `ps -A` snapshot window):
```
count: 500
lowest 5 PIDs: [98, 99, 110, 111, 118]
has pid 1: False
has pid 141: False
total ports: 11
listening: 0
```

Independent test on a separate host, through the full manager/indexer/dashboard pipeline — agent `002` (before, released `4.14.5` package) vs. agent `003` (after, source build of this branch @ `1d7f4e0587`):

| Measure | Before (agent 002) | After (agent 003) |
| --- | --- | --- |
| Processes on host (`ps -A`) | 830 | 815 |
| Processes recorded (indexer, matches dashboard hit count) | 500 | 813 |
| Lowest PIDs recorded | 4987, 4989, 4990, 4991, 4992 | 0, 1, 90, 91, 92 |
| PID 1 (`launchd`) recorded | no | yes |
| Listening ports recorded | 2 (both `cupsd`) | 8 (incl. `launchd`'s 22 and 5900) |

Screenshot 2026-09-16 at 6 51 49 PM

Screenshot 2026-09-16 at 6 58 49 PM

`syscollector` → `local.db` SELECT evidence, real shipped agent (`v4.14.9`) enrolled against a real Wazuh `5.0.0` manager, before/after on the same host:

```
Before (fix reverted, rebuilt, redeployed):

$ ps -A | wc -l
     758
$ sudo sqlite3 /Library/Ossec/queue/syscollector/db/local.db 'select count(*) from dbsync_processes;'
500
$ sudo sqlite3 /Library/Ossec/queue/syscollector/db/local.db "select pid,name from dbsync_processes where cast(pid as integer)=1;"
(no rows — launchd not recorded)
$ sudo sqlite3 /Library/Ossec/queue/syscollector/db/local.db "select pid,name from dbsync_processes order by cast(pid as integer) limit 5;"
301|backupd-helper
381|liquiddetectiond
457|SubmitDiagInfo
460|PhotosReliveWidget
464|ctkd

After (fix restored, rebuilt, redeployed):

$ ps -A | wc -l
     776
$ sudo sqlite3 /Library/Ossec/queue/syscollector/db/local.db 'select count(*) from dbsync_processes;'
776
$ sudo sqlite3 /Library/Ossec/queue/syscollector/db/local.db "select pid,name from dbsync_processes where cast(pid as integer)=1;"
1|launchd
```

## Tests

  - [ ] Linux
  - [ ] Windows
  - [x] MAC OS X
- [x] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests


Nicogp (changes requested):
Delete the `"chore: trigger CI checks"` commit
Adjust the PR title; do not use the same commit message as the PR title
Please provide more evidence from the tests you ran, such as a SELECT statement from the syscollector->local.db database.

Nicogp (approved):

PR #39150, opened 2026-09-10
Manager HTTPS enrollment review: stop publishing and losing credential material

### LINKED ISSUE #39078: Manager HTTPS enrollment refinement: fix CA publication, identity rotation and token persistence
> Follow-up refinement of #38991 (manager half of #38940), reviewing the implementation of #38992 and #38993.
> **Manager scope only.** This is a separate issue from #38994: that issue retains token purge/cap, secret-disclosure RBAC/audit and token read permissions. This issue tracks the review findings, their fixes and the remaining delivery/validation decisions. The shared acceptance work below must be coordinated with #38994.

## Review baseline and result

Reviewed `enhancement/38991-identity` at [`e670642eeca4fa1ed201161ac77fc6bef1aae844`](https://github.com/wazuh/wazuh/commit/e670642eeca4fa1ed201161ac77fc6bef1aae844) against `origin/5.0.0` at [`0cff1987dccdc223f6f4558484c442109c9ba169`](https://github.com/wazuh/wazuh/commit/0cff1987dccdc223f6f4558484c442109c9ba169), on 2026-09-09. The branch-specific diff contains 17 commits and 241 files. The local `5.0.0` reference at `cf0d103da7` was stale and was not used as the review baseline.

**Changes requested before integration:** nine findings, four P1 and five P2. P1 means high-priority security or identity/persistence correctness; P2 means a functional/security defect with a narrower trigger. D01–D04 are deliberate design/compatibility decisions to reconcile, not four additional implementation bugs.

The evidence below comes from source inspection, existing suites and isolated reproductions against this revision. Passing the existing suites does not close these findings. The local review bundle contains the detailed report, probe sources and logs; this issue includes the relevant observations and permanent source links so it is independently actionable.

## Findings to fix

| ID | Priority | Problem | Evidence |
| --- | --- | --- | --- |
| H01 | P1 | Public CA response and embedded token can include a CA private key | Real TLS listener with an ephemeral combined PEM |
| H02 | P1 | One re-enrollment secret can authorize two distinct rotations | Real dispatcher/JWT verifier, pending-writer DB snapshot simulated |
| H03 | P1 | A failed DB write discards an already-issued re-enrollment secret | Writer control-flow inspection |
| H04 | P1 | Retrying a failed revoke reports success without persisting it | Real temporary token store, failure/retry/reload |
| H05 | P2 | DB error logs contain the agent key and re-enrollment secret | Existing passing test explicitly expects the secret-bearing log |
| H06 | P2 | CA response uses a verdict up to a day older than its PEM | Real TLS listener, CA replacement and repair |
| H07 | P2 | A failed save lets a limited token regain uses after restart | Real temporary token store |
| H08 | P2 | CLI listing exceeds its fixed response buffer with hundreds of tokens | 400 real entries and the actual C receiver |
| H09 | P2 | Re-enrollment rejects valid nine-/ten-digit agent IDs | Real dispatcher rejects before JWT verification |

### H01 — P1: never publish private material from a combined CA PEM

**Trigger and impact.** `remote.https.ca_certificate` points to a PEM containing both the CA certificate and its private key. The normal provisioning flow uses separate files; this finding is conditional on combined/misprovisioned input. The endpoint checks for a certificate marker and returns the entire file, while the X.509 loader skips non-certificate blocks. In TLS mode without a client-certificate requirement, an unauthenticated request can obtain the CA private key. `--embed-ca` also copies the raw file into a token.

**Source:** [CA file reader and HTTP response](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/remoted/remoted_module/src/endpoints/cacertsEndpoint.cpp#L55), [X.509 loader](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/remoted/remoted_module/src/http_server/tlsCertificateStatus.cpp#L66), [embedded CA](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/enrollment_token_mint.c#L277).

**Observed:** `COMBINED_PEM status=200 contains_private_key=1`, using ephemeral credentials and a real loopback TLS listener. No installed CA key was used or printed.

- [ ] Parse and serialize only X.509 certificates for both surfaces, or explicitly reject private-key-containing input before publication. A substring check is insufficient.
- [ ] Add regression cases for combined PEM, empty file, arbitrary text and malformed certificate blocks. No private key bytes may appear in either HTTP or decoded token output; a valid public CA must remain usable.

### H02 — P1: make validation and reservation of an identity rotation atomic

**Trigger and impact.** `local_reenroll()` reads the old `reenroll_secret` from wazuh-db and verifies the JWT before taking `mutex_keys`. The resulting rotation is queued and acknowledged before the writer updates the DB. A second request can therefore verify the same old secret and perform another rotation, even sequentially while the writer is pending. Two successful responses contain different credentials for one ID; one response is invalidated by the other.

**Source:** [DB read and verification before lock](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/local-server.c#L954), [lock](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/local-server.c#L977), [queue and response](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/local-server.c#L1049), [later persistence](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/main-server.c#L1400).

**Observed:** `REENROLL identical_bearer_successes=2 distinct_rotations=2 real_JWT_verifier=1 database_writer_pending=1`. The probe uses the real dispatcher, keystore, queue and JWT verifier with an injected clock, and mocks the unchanged DB row to represent the pending writer. This is not a deployed-cluster concurrency test.

- [ ] Make validation of the current secret generation and reservation of its transition indivisible, using coherent per-agent state or a conditional/transactional update. Checking only that the ID still exists under the lock is insufficient.
- [ ] With verification barriers and a delayed writer, duplicate/concurrent requests must cause at most one credential transition; define the second caller's result. Validate through two workers as well.
- [ ] Review together with H03. This is rotation atomicity and does not require the general `jti` replay cache deferred by the epic.

### H03 — P1: retain a recoverable identity transition when wazuh-db fails

**Trigger and impact.** Enrollment/re-enrollment returns the new secret before the writer persists it. If the credential DB write fails, the writer logs the error and frees the queued operation and secret without retrying. `client.keys` may already contain the new key while `global.db` retains the old credentials. The new secret cannot be reconstructed from `client.keys`; an agent can receive a secret it cannot subsequently use. First insertion has an equivalent loss path.

**Source:** [response before persistence](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/local-server.c#L1049), [client.keys write](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/main-server.c#L1365), [DB update failure](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/main-server.c#L1400), [unconditional queue cleanup](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/main-server.c#L1431).

**Evidence:** control-flow inspection. No failure was injected into the installed wazuh-db service during this review.

- [ ] Persist a recoverable transition before acknowledging it, or provide durable journaling/retry and a defined recovery procedure for partial writes. Logging and discarding the only copy is insufficient.
- [ ] Inject socket and SQL errors and restart between the two persistence steps. After recovery, any acknowledged secret must remain usable and the previous generation must not regain validity. Cover both first enrollment and rotation.
- [ ] Keep this separate from D02: the failure also occurs with the correct DB schema.

### H04 — P1: retry persistence before acknowledging a previously failed revoke

**Trigger and impact.** `etoken_store_revoke()` sets `revoked=1` in memory, then save fails. After storage recovers, retry sees the in-memory flag and immediately returns success without writing. Disk and workers can retain an active token, and restart reactivates it on the master. The dispatcher also maps storage failure to token-not-found `9022`.

**Source:** [early return and save](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/enrollment_token_store.c#L866), [error mapping](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/local-server.c#L1325).

**Observed:** `REVOKE first=-1 retry=0 disk_revoked=0 after_restart_revoked=0`, with a real temporary store and a deliberately unavailable write path, then recovery and reload.

- [ ] Distinguish in-memory revocation from persisted revocation. Keep rejection in memory if appropriate, but retry saving before returning success.
- [ ] Return a storage error distinct from unknown token, with coherent CLI/API mapping.
- [ ] Validate failure → restore storage → retry → restart → worker propagation. The retry must persist the revoke or continue reporting failure.

### H05 — P2: redact credentials in DB failure logs

**Trigger and impact.** At debug level 2, DB transport/response errors log complete insert/credential-update queries containing `internal_key` and `reenroll_secret`. The new secret therefore reaches log files and downstream collectors. The `authd.pass` API gate in #38994 does not cover this path.

**Source:** [credential query and error log](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/shared/src/wazuhdb_queries_op.c#L298), [insert error log](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/shared/src/wazuhdb_queries_op.c#L101), [test explicitly expecting the full secret-bearing log](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/unit_tests/shared/test_wazuhdb_queries_op.c#L1133).

**Evidence:** the existing `test_wazuhdb_queries_op` suite passed; its socket-error test requires the unsafe log output.

- [ ] Log verb, agent ID and error only, or use an explicit field allowlist. Inspect wazuh-db parser error paths too.
- [ ] Change tests to require absence of complete synthetic secrets/keys at all log levels. Force insert and rotation errors and scan both caller and wazuh-db logs.
- [ ] Apply the same no-secret-values rule to the disclosure audit introduced by #38994.

### H06 — P2: serve the CA and its validation verdict from a coherent state

**Trigger and impact.** Each `/cacerts` request rereads the file but uses the match verdict from startup or the daily certificate-status tick. Replacing a valid CA with an unrelated CA can return `200` and a stale positive metric; correcting an invalid CA can keep returning `503` for up to a day.

**Source:** [per-request file read and cached verdict](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/remoted/remoted_module/src/endpoints/cacertsEndpoint.cpp#L96), [periodic evaluation](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/remoted/remoted_module/src/http_server/RestinioHttpServer.cpp#L1458), [default interval](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/remoted/remoted_module/src/http_server/httpServerConfig.hpp).

**Observed with real TLS listeners:**

```text
CA_REPLACED status=200 cached_match=1 verifying_handshake_ok=0
CA_FIXED status=503
```

- [ ] Cache/validate the PEM and match verdict together and invalidate them when the file changes. Validate against the certificate actually loaded by the listener. Avoid separately opening the file for verification and response.
- [ ] Keep daily expiry monitoring independent of CA-content invalidation.
- [ ] Test valid → foreign CA, foreign → valid CA, and missing → restored CA without restart; HTTP response and metrics must describe the same state without waiting for the daily tick.

### H07 — P2: preserve limited-token usage guarantees across save failures

**Trigger and impact.** `etoken_store_consume()` increments `uses` and returns `ETOKEN_USE_OK` even if saving fails. Restart reloads the earlier counter, so a single-use or limited-use token can authorize additional consumption. This is currently an explicit best-effort choice; no claim is made that a remote client can itself induce the disk failure.

**Source:** [increment, failed save and continued success](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/enrollment_token_store.c#L909).

**Observed:** `CONSUME max_uses=1 first_is_OK=1 after_restart_is_OK=1`, using the store directly with failed persistence and reload, not two enrollments against an installed manager.

- [ ] Make a limited token's reservation durable before admitting enrollment; expose a recoverable storage error separately from invalid/exhausted token errors.
- [ ] If best-effort accounting is retained, explicitly resolve and document that single-use is not a durable guarantee; do not close this finding by leaving the current behavior implicit.
- [ ] Test save failure, restored storage and restart without additional accepted uses. Cover process/system crash durability; atomic rename alone is not proof of persistence after power loss.
- [ ] Coordinate reservation rollback and purge with #38994.

### H08 — P2: handle complete list/purge responses within an explicit transport bound

**Trigger and impact.** `token_list` serializes every entry while `cli_request()` allocates an `OS_MAXSTR` (65,536-byte) response buffer. With only hundreds of entries, listing fails before the proposed 5,000-token store limit. Descriptions and retained revoked entries make it worse.

**Source:** [list response](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/local-server.c#L1306), [fixed CLI buffer](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/token_cli.c#L221), [receiver rejects oversize messages](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/shared/os_net/os_net.c#L727).

**Observed:** `CLI_LIST tokens=400 payload_bytes=69220 receiver_cap=65536 result=-6`. The probe mints 400 real entries without descriptions and exercises the actual C receiver's length check. It does not invoke the installed CLI. The Python client receives the announced length dynamically, so this finding is not automatically attributed to API GET.

- [ ] Add protocol/CLI pagination or bounded dynamic reception sized for the supported response. Increasing only the store-file ceiling does not fix this.
- [ ] Test 5,000 entries with maximum descriptions and explicit handling of any policy limit.
- [ ] Apply the same analysis to #38994's purge result: thousands of returned IDs also exceed 64 KiB. Verify CLI and API separately.

### H09 — P2: accept the protocol's full canonical agent-ID range

**Trigger and impact.** The re-enrollment `kid` parser calls `OS_IsValidID()`, whose historical limit is eight characters. Administrative insertion accepts positive IDs through `INT32_MAX`, and the JWT canonical-ID helper supports nine and ten digits. An imported ID such as `100000000` is valid but fails before authentication; this does not require 100 million concurrent agents.

**Source:** [dispatcher validation](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/os_auth/src/local-server.c#L532), [eight-character limit and insertion range](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/shared/src/agent_validate_op.c#L174), [JWT canonical form](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/shared_modules/utils/jwt/canonicalAgentId.hpp#L36).

**Observed:** `NINE_DIGIT_ID allowed_by_insert=1 reenroll_dispatch_error=9027 before_verification=1`.

- [ ] Validate the canonical representation and numeric range shared by the protocol and wazuh-db instead of reusing the historical string-length restriction.
- [ ] Test `001`, `99999999`, `100000000` and `2147483647` through verification/rotation with valid credentials. Reject zero, overflow and noncanonical leading-zero forms.

### D01 — external certificate provisioning is the implemented design

The final #38992 implementation deliberately removes certificate generation from manager daemons/installers and removes the generation/reissue CLI flags. The listener requires externally provisioned certificate/key material and fails closed if that pair is missing. This is the local plan's T27 decision, implemented by [`f318c100cb`](https://github.com/wazuh/wazuh/commit/f318c100cb) and [`d99f57aad6`](https://github.com/wazuh/wazuh/commit/d99f57aad6), and described in the [installation reference](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/docs/ref/getting-started/installation.md#L75).

- [ ] Reconcile #38992's original description and final PR acceptance with this design; do not restore manager-side generation merely to satisfy stale wording. The epic is being aligned with this review.
- [ ] Track the production installation-assistant/provisioning dependency: deliver `remoted.pem`, its private key, the appropriate CA/chain and SANs for the actual endpoint. The devcontainer/CI certificate tool proves development/test provisioning only.
- [ ] Validate clean DEB/RPM installs with and without preprovisioned material, useful startup errors and recovery after provisioning, and preservation of existing certificates/keys on upgrade.

### D02 — explicitly define the supported global.db upgrade path

`reenroll_secret` was added to the initial schema while `user_version` remains 1 and there is no corresponding upgrade step. The local T30 decision treats previous 5.0.0 development databases as disposable, and the [DB module documentation](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/docs/ref/modules/wazuh_db/README.md#L72) records the limitation. See the [upgrade implementation](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/wazuh_db/src/wdb_upgrade.c#L16).

A memory-only DB created from the review base reproduces:

```text
Base user_version: 1
New INSERT against unchanged base schema: table agent has no column named reenroll_secret
```

- [ ] State the supported/unsupported upgrade path in release notes and test procedures.
- [ ] If upgrades from that schema are supported, add a conservative migration and preservation tests. If intentionally unsupported development builds require recreation, document its data-loss consequences and constrain the procedure accordingly; deleting a DB is not a generic repair.
- [ ] Do not conflate this deliberate compatibility decision with H03's runtime persistence failure.

### D03 — existing RBAC databases need an explicit policy-update path

Default YAML gains token actions, but `CURRENT_ORM_VERSION` remains 1; existing databases keep old policies. This is already documented as local D71 / a #38993 caveat. See [policies](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/framework/wazuh/rbac/default/policies.yaml#L12) and [ORM version](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/framework/wazuh/rbac/orm.py#L47).

- [ ] Define and document how existing installations receive the new token and #38994 secret-read actions. Test fresh and pre-change databases.
- [ ] If migrating policies, preserve operator roles/policies and make the change idempotent. Do not automatically recreate the entire RBAC DB.
- [ ] Keep `create`/`delete` in `agents_all` and move/add `read` to `agents_read` as specified in #38994. Align documentation with actual role relationships: `agents_all` is held by `administrator`, `agents_admin` and `wazuh_indexer_admin`.

### D04 — reconcile mint-address and rejection-status contracts

- [ ] Resolve the mismatch between #38993's requirement to reject `--address 127.0.0.1` and the implementation, which rejects a loopback-only SAN set but permits a loopback address if another non-loopback SAN exists. The [current positive test](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/unit_tests/os_auth/test_local-server.c#L715) expects that mint to succeed. Either change that behavior or explicitly amend the acceptance contract; a loopback-address token is not useful to a remote agent.
- [ ] Keep the documented distinction between remoted's `401 token_*` rejection and authd's possible `403` / enrollment error code during replica propagation. The local plan records this as D73. Align client tests and docs; do not promise that every revoked/expired token is always rejected with the same 401.

## Coordination with #38994

The following work remains owned by #38994; it is listed here to make all pending review work traceable without creating a second implementation owner.

| #38994 work | Contract to retain | Additional review requirement |
| --- | --- | --- |
| Token purge | Master-only `token_purge`; default `dead` removes expired/revoked/exhausted entries; `all` clears the store; keep live unused tokens | Account for in-flight enrollment reservations and rollback before pruning an apparently exhausted token |
| CLI/API purge | `--purge-enrollment-tokens [--all] [--force]`, terminal confirmation for `--all`, removed/remaining counts; collection DELETE with `status=dead\|all`, delete action and master routing; per-ID DELETE remains revoke | Apply H08 to large purge responses and preserve the distinction between revoke and removal |
| Capacity | 5,000-entry cap, prune dead entries before refusing mint, `9025` naming purge, warning at 80%; `W_ETOKEN_STORE_MAX_BYTES` = 7 MiB below remoted's 8 MiB ceiling | Validate the final serialized size, including descriptions, escaping, embedded CA and counter growth; cover every mutation or reserve sufficient margin |
| Secret disclosure | Dedicated `manager:read_secrets` / `cluster:read_secrets`, administrator-only default; fail-closed masking on errors; INFO audit for clear `authd.pass` / `cluster.key` | Audit user, source IP, endpoint and field, never the secret value; cover distributed responses and error paths; close H01/H05 separately |
| Token read RBAC | `enrollment_token:read` available through `agents_read`; writes remain in `agents_all` | Exercise the D03 policy-update path and custom roles |
| Docs/tests | CLI/authd/API/RBAC reference, OpenAPI, CHANGELOG, cmocka, framework/API pytest, tavern and integration | Keep contracts coherent with D01–D04 and run the combined matrix below |

The previously measured 242-byte regular entry and 1,423-byte entry with an approximately 1.2 KB CA are examples, not worst-case bounds: 5,000 of the latter are approximately 6.79 MiB, but larger CAs/descriptions can cross 7 MiB earlier.

- [ ] Define purge behavior for outstanding reservations: enrollment consumes before `local_add()` and releases on failure; pruning an exhausted token in between must not destroy recoverable state.
- [ ] Measure full-store cost with 1, 100 and 5,000 tokens, with/without embedded CA and with multiple workers. Consumption rewrites the full JSON, and replica reload entails reading, hashing/parsing and key derivation. Report latency/CPU and reload behavior. This is an unmeasured performance risk, not a demonstrated regression.

## Integration and regression matrix

| Scenario | Required evidence before closure |
| --- | --- |
| Same ID/secret re-enrolled simultaneously through two workers | At most one effective credential transition; defined second response |
| DB unavailable during first enrollment/rotation; recovery and restart | Acknowledged secret retained/recovered; old secret does not regain validity |
| Store write failure during revoke and consume | No false success or additional uses after restart; useful error mapping |
| First token minted on master; revoke while worker reconnects | Measured propagation bound, expected rejection vocabulary and observable metrics |
| Combined CA PEM; CA replacement/repair while listener remains active | No private material; response, loaded listener certificate and verdict stay coherent |
| 5,000 entries, large descriptions/embedded CA, list and purge | Coherent count/byte/transport limits; no oversized replica-reload failure |
| Secret-read roles and injected masking failure | No disclosure without authorization or on masking failure; audit contains no secret values |
| Fresh/existing RBAC DBs and customized roles | Intended permissions applied without discarding custom policy |
| Clean install and upgrade, DEB/RPM | External provisioning contract verified; D02/D03 compatibility explicit |

## Validation already performed and remaining limits

At the reviewed revision, these checks passed:

- `remoted_module_utest`: 871 tests; 2 additional tests disabled.
- Eleven selected C/cmocka suites covering token store, CLI, local server, re-enrollment verification, token codec, X.509, wazuh-db/global parser/query helpers, agent IDs and base64url.
- Framework token/agent pytest: 135 passed; API agent-controller pytest: 46 passed. Dependency/deprecation warnings remain.
- Three manager-config suites and two Go packages (`internal/wire`, `internal/enrollhttps`).
- Eight isolated observations reproduced the defective behavior across H01/H02/H04/H06/H07/H08/H09. H05 is evidenced by the existing log-expectation test; H03 by source control flow.

The probe assertions deliberately confirm the old defect; they must be converted to desired-behavior regression tests after fixes. Passing them before a fix is not acceptance evidence.

Full deployed master/worker integration, tavern, DEB/RPM installation/upgrade, the 5,000-token benchmark and dedicated TSAN/Coverity/Valgrind campaigns were not run as part of this review. Unit/API mocks do not prove end-to-end cluster or RBAC behavior. No installed service, DB or credential material was modified to obtain the review evidence.

## Documentation and cleanup

- [ ] Update authd/remoted/DB/API/RBAC references, OpenAPI where contracts change, release notes and CHANGELOG; include failure/recovery semantics and propagation behavior.
- [ ] Align #38992/#38993's stale acceptance text as resolved in D01/D04 and the final PR description with the complete manager diff. The local untracked `PR.md` currently describes only #38993; if used for the complete branch, include certificate-generation removal, external provisioning and package changes.
- [ ] Update the local planning README's old PR3 summary so it no longer suggests implementing the deferred rate limiter or the /download work here.
- [ ] Remove trailing whitespace reported by `git diff --check origin/5.0.0...HEAD` in [wazuh-certs-tool.sh](https://github.com/wazuh/wazuh/blob/e670642eeca4fa1ed201161ac77fc6bef1aae844/src/engine/tools/devContainer/scripts/wazuh-certs-tool.sh#L62), lines 62, 1068 and 1391.

## Completion criteria and suggested order

1. Close H01/H05 secret exposures with negative tests that check complete synthetic secret values.
2. Resolve H02/H03 together, including concurrent transitions, durable persistence and recovery.
3. Fix H04/H07 storage semantics and client-visible failure mapping.
4. Close H06/H08/H09 and coordinate the transport/reservation changes with #38994.
5. Resolve D01–D04, complete the matrix and publish updated contracts and validation results.

- [ ] All nine findings have fixes and regression evidence, or an explicitly recorded accepted disposition where this issue calls for a policy decision.
- [ ] Remaining delivery decisions and shared acceptance work have an owner and verified resolution.
- [ ] Required affected suites and integration scenarios pass; any exclusions are documented with a follow-up owner.
- [ ] The epic and final PR describe the delivered manager behavior consistently.

## Out of scope

Agent implementation; per-client `POST /enroll` rate limiting/lockout (deferred); `POST /download` group-membership enforcement (tracked by #38683); general `jti` replay cache; re-emitting an already-minted token's credential; a tunable token cap or new cluster-wide purge coordination; removing the password endpoint.

The other certificate/client-authentication and adjacent-component follow-ups remain as listed in #38991. This issue does not reintroduce the earlier, superseded scope of #38994.

juliancnn:
## Description

Closes #39078 — the review round of epic #38991, after #39014 (listener certificate), #39040 (enrollment tokens) and #39083 (hardening).

An external review of what those three PRs delivered raised **nine findings** (four P1) and **four delivery decisions**. All nine were reproduced against this epic's HEAD before any code was written, and every one of them is answered here: five with a fix, four contracts with the text that was missing. The four decisions are recorded rather than reopened.

Three of the findings are about the manager giving away, or losing, credential material:

- `GET /cacerts` and `--embed-ca` published **the bytes of the configured CA file** after checking only that they contained a certificate marker, so a PEM that also carried the CA's private key — a misprovisioned but accepted input — was served whole to an unauthenticated caller.
- The credential arguments of `insert-agent` and `set-agent-credentials` reached the manager log, which is readable through `GET /cluster/{node_id}/logs` with ordinary read permission.
- An enrollment or a re-enrollment answered with a key and a re-enrollment secret **long before the writer stored them**, so a wazuh-db outage — or a crash in between — left the agent holding credentials nothing durable knew about: it could talk to remoted once `client.keys` landed, and it could never re-enroll.

The other two are correctness: two requests carrying the same re-enrollment bearer produced **two** credentials for one agent, and a revoke that could not be written answered success and came back after a restart.

## Proposed Changes

**authd (`src/os_auth`)**

- **Identity transition journal** (new `src/os_auth/src/identity_journal.c`): every enrollment and rotation is recorded in `queue/authd/pending-identities` (mode 0640, one JSON line, appended without rewriting) **before** the answer is built, and the entry is dropped only once the database write is **committed**. The writer retries on its own timer (1 s, doubling to 60 s) instead of waiting for the next enrollment, and a retry-only cycle does not rewrite `client.keys`. At start-up, and again before each retry, each entry is judged against `client.keys` **by key, not by agent id** — both generations of a rotation share the id — so a superseded transition is discarded and an agent that was deleted meanwhile is never resurrected. A transition that cannot be recorded refuses the operation with the new `9031` instead of handing out a credential nothing wrote down.
- **One effective rotation per agent**: a re-enrollment reserves the agent *before* the database is read, and the reservation is released by the writer only when the new credentials are committed. The second caller gets the new `9030`; a verification that finishes after another rotation landed is caught by a generation re-check under the keystore lock.
- **A revoke is either written or reported as failed** (`9029`), never the `9022` of an unknown id and never a silent success; the pending revocation survives a reload. A store above 5000 tokens or its byte ceiling is refused at load.

**remoted (`src/remoted/remoted_module`)**

- `GET /cacerts` and the certificate monitor now share **one read** of the CA file, cached under its **content hash** (not size or mtime): the endpoint serves a PEM this process re-serialised from the X.509 objects it parsed — so nothing but certificates can leave — and the "does this CA sign the served leaf" verdict comes from the same read, instead of from a snapshot up to 24 hours old. A file that cannot be parsed to its end is refused whole.
- `9030` → **409**, `9031` → **503**, and a new counter `remoted.enroll.reenroll.rejected_in_progress`.

**wazuh-db (`src/wazuh_db`, `src/shared`)**

- New `global commit` query: every other write answers `ok` from inside a deferred transaction this daemon commits on its own clock, so that `ok` is not durability. authd is its only caller — it may forget a journaled transition only once the write is committed.
- The error paths of `insert-agent` and `set-agent-credentials` log the verb and the agent id, never the payload (client and server side).

**Framework / API**

- `9031` → `WazuhInternalError(1772)`.

**Documentation** (the four delivery decisions, D01–D04 of the issue)

- Re-enrollment supports agent ids of **up to eight digits**; administrative insertion keeps the full range on purpose, because restricting it would not remove the mismatch.
- The mint refuses a certificate whose SAN set is **loopback only**, not the loopback address when the certificate also names something reachable.
- A revoked token may be refused as remoted's uniform `401` or as authd's `403`/`9022`, depending on which side notices first.
- There is no upgrade path for a `global.db` from before `reenroll_secret`, no migration of RBAC policies, and rebuilding agent rows from `client.keys` does not restore the secrets.
- Also removes the three trailing-whitespace warnings `git diff --check` reported in the vendored `wazuh-certs-tool.sh`.

### Results and Evidence

Every finding was reproduced with the previous binary and closed with the new one on an installed manager. Highlights:

```text
# H03 — a credential handed out while wazuh-db is down
$ add over queue/sockets/auth.sock            -> 200 {"id":"004", key, reenroll_secret}
$ cat queue/authd/pending-identities          -> {"seq":5,"id":"004",...}   (mode 0640)
$ kill -9 authd ; start wazuh-db ; start authd
wazuh-manager-authd: INFO: Recovered 1 identity transition(s) that the previous run did not finish writing to the database.
wazuh-manager-authd: INFO: Recorded credentials of agent '004' written to the database (enrollment recovered).
$ sqlite3 global.db "select id,substr(reenroll_secret,1,10) from agent where id=4"  -> 4|47508b2060   (what the agent was given)

# The agent deleted while its transition was owed is NOT resurrected
wazuh-manager-authd: INFO: Discarded 1 recorded identity transition(s) whose agent is no longer in client.keys and 0 superseded by a later one.

# H02 — six simultaneous rotations with the same bearer, from one process with a barrier
Counter({9030: 5, 0: 1})        # one credential; client.keys and global.db agree

# H01 — CA file carrying its own private key
$ grep -c 'PRIVATE KEY' etc/certs/root-ca.pem   -> 2
$ curl -sk https://localhost:1520/wazuh-manager/cacerts | grep -c 'PRIVATE KEY'   -> 0   (200, 1 certificate)

# 9031 — the journal cannot be written
$ add over the socket                          -> {"error":9031,"message":"Identity transition could not be recorded"}
$ POST /enroll                                 -> 503, body {"error":{"code":9031,...}}   ; no agent in client.keys

# RD-2 — the manager provisions nothing
(1244): Invalid configuration at '/remote/https/certificate': file not found: … (the manager does not generate
certificates; provision the file, e.g. with wazuh-certs-tool).
wazuh-manager-authd: ERROR: SSL context setup failed … wazuh-manager does not generate TLS certificates … Exiting.
```

Suites, on the rebased branch:

| Suite | Result |
|---|---|
| cmocka (`ctest`, ASAN+UBSAN) | **111/111** (new binary `test_identity_journal`, 12 cases, no leaks) |
| `remoted_module_utest` | **891/891** |
| `framework/wazuh/core/tests/` · `framework/wazuh/tests/` · `api/api/` | **749** · **297** · **648** |
| authd integration tests (enrollment token, master · worker) | **9** · **4** |
| `pr-clang.sh --check` · `git diff --check <base>...HEAD` · `docs/build.sh` | clean · clean · builds |

  - [x] Linux
  - [ ] Windows — N/A (no agent or Windows code is touched)
  - [ ] MAC OS X — N/A (no agent or macOS code is touched)

- Memory tests for Linux
  - [ ] Coverity — _(In progress — fill in before requesting review)_
  - [ ] Valgrind (memcheck and descriptor leaks check) — _(In progress — fill in before requesting review)_
  - [x] AddressSanitizer — the whole cmocka suite runs under ASAN+UBSAN; the new journal suite is leak-clean
- Memory tests for Windows
  - [ ] Coverity — N/A
  - [ ] UMDH — N/A
- Memory tests for macOS
  - [ ] Leaks — N/A
  - [ ] AddressSanitizer — N/A

  - [ ] Added unit testing files ".ini" — N/A
  - [ ] `runtests.py` executed without errors — N/A

  - [ ] Test run in parallel — N/A (the engine is not touched)
  - [ ] ASAN for test (utest/ctest) — N/A
  - [ ] TSAN for test and wazuh-engine. — N/A

- Wazuh server API/Framework
  - [x] Run API Integration Tests — `framework` and `api` unit suites green; the API integration suites run in CI

### Artifacts Affected

- **Executables**: `wazuh-manager-authd` (new source file, new error code), `wazuh-manager-db` (new `global commit` query), `wazuh-manager-remoted`.
- **Libraries**: `lib/libremoted_module.so` (the CA snapshot and the status mapping live here — swapping only the remoted binary is not enough), and the manager's shared library through `src/shared`.
- **Python**: the framework installed under the embedded interpreter (`wazuh.core.agent`, `wazuh.core.exception`).
- **New runtime file, not shipped**: authd creates `queue/authd/pending-identities` (0640) in the existing `queue/authd/` directory; no packaging change is needed.
- Not artifacts, and considered: `docs/**` (mdBook) and `src/engine/tools/devContainer/scripts/wazuh-certs-tool.sh`, which is developer tooling — `grep -rn` over `packages/` and `src/Makefile` finds no packaging reference to that file.

### Configuration Changes

N/A — no new option, no default changed, no schema change.

### Tests Introduced

- **`src/unit_tests/os_auth/test_identity_journal.c`** (new, 12 cases): what the journal writes and reads back, its 0640 mode, that appending never rewrites the file, a torn last line, the byte ceiling, an unwritable path, the admission bound, drop-with-compaction and its idempotence, the bounded snapshot, and the three branches of start-up reconciliation (same key / different key / agent gone).
- **`test_local-server.c`** (+6 across E4a/E4b): two rotations with one bearer, a stale verification, the reservation released on rejection and kept on a failed write, an accepted enrollment journaled with its own credential, and both refusals when the journal cannot be written.
- **`caCertificateSource_test.cpp`** (new, 9 cases) and a rewritten `cacertsEndpoint_test.cpp` decision table: combined PEM, empty/text/key-only files, a corrupt block after a valid one, a bundle whose signer is not first, and a same-size same-mtime replacement.
- **`test_enrollment_token_store.c`** (+9): a save failure reported instead of swallowed, the pending revocation surviving a reload, and a legacy store above either limit.
- **wazuh-db**: `test_wdb_global_parser.c` (+2, the `commit` verb), `test_wazuhdb_queries_op.c` (+3, `wdb_commit_global`), `test_wazuh_db_state.c` (the new counter), and the redaction cases of the client and the parser.
- **remoted**: `enrollmentEndpoint_test.cpp` (`9030` → 409 with its counter, `9031` → 503) and `metrics_test.cpp`.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

**Excluded from this PR, with an owner** (each has a draft issue ready): two-node cluster acceptance of simultaneous rotation, the 1/100/5000-token store cost benchmark, DEB/RPM install and upgrade validation of the provisioning contract, support for nine- and ten-digit agent ids, wazuh-db's 32-byte echo of a rejected query, observability of a pending identity transition, whether the journal must survive a power cut, socket-level pagination of the token verbs, and the vendored `wazuh-certs-tool.sh` whitespace at its source.


juliancnn:
### Codex review round — what was fixed and what was not

Seven of the eight observations are fixed in `bb86e379be` and `f4d5193642`; one is deliberately not, with its reasoning below. The triage criterion: a failing disk is not something this design tries to engineer around — what has to be solid is that recovery on the next start is stable, and losing an individual credential to a crash window is acceptable. Five of the seven had nothing to do with the disk at all.

| # | Observation | What happened |
|---|---|---|
| 1 | Reservation slots for rejected agent ids | **Fixed.** The slot is created for the id the *request* names, before the agent is known to exist, so anyone reaching `POST /enroll` could seed one per invented id — memory that never returns and a longer linear search under the mutex every legitimate rotation takes. A slot with no completed rotation to remember is now dropped; what stays is one per agent that actually rotated. Verified live: 500 re-enrollments with invented ids answer `9026` and authd's RSS does not grow (29 324 → 29 168 kB). |
| 2 | Forced replacement undone when journaling fails | **Fixed**, by refusing earlier rather than rolling back: `w_auth_replace_agent()` deletes the previous agent while the request is validated, so the bound is now checked *before* anything is mutated — the same phase-0 reasoning `purge_backlog_full()` already applies one line above it in that function. New unit test asserts the previous agent survives a refusal. |
| 3 | Unbounded database failures in the retry pass | **Fixed.** `wdbc_connect()` is five attempts with 1..5 s sleeps and leaves the socket closed, so the pass could hold the single keystore writer for ~2 h. It now skips the pass when wazuh-db's socket is not even there and stops at the first entry that cannot reach it, and checks `running`. Verified with wazuh-db stopped: four enrollments answer immediately and all four reach `client.keys`. |
| 4 | Transitions whose keystore write has not landed | **Fixed, and it was the important one.** Judging an entry by the key in `client.keys` called a rotation "superseded" when authd died before the writer rewrote that file, deleting the only durable copy of credentials the agent already holds — leaving it unable to connect *or* to re-enroll. What supersedes an entry is now a **later entry for the same agent** (the journal's own order); an id absent from `client.keys` still owes nothing. Live: seeding that exact state now logs `1 identity transition(s) are still owed…`, the database receives the credential the agent has, and the agent heals itself by re-enrolling with it. |
| 5 | Applied entries when the database reconnects | **Not fixed, on purpose.** Real — `wdb_commit2()` reports success with no open transaction, so a wazuh-db restart between the write and the commit makes us forget entries whose writes died with it. It needs wazuh-db to die inside a millisecond window, corrupts nothing, deletes no agent, and does not affect recovery on the next start. The cheap heuristic (comparing the socket descriptor) is unreliable, since the kernel usually returns the same number after `close()`; doing it properly means wazuh-db publishing a session identifier, i.e. a protocol change. Tracked as a follow-up issue. |
| 6 | Newly queued transitions in recovery snapshots | **Fixed** with a high-water mark: the retry pass looks no further than the last transition the writer has taken work for. It also removes the double `w_reenroll_complete()` you noted. |
| 7 | Reservations for recovered rotations | **Fixed.** Startup reconciliation takes the reservation back for every retained rotation, before the listeners start; the recovery commit releases it. |
| 8 | Incomplete journal tails | **Fixed.** An append now closes an unterminated tail first, so the fragment stays a malformed line of its own — ignored on load, like any other — instead of swallowing the next good entry. |

While verifying #4 against an installed manager, the first version of the fix turned out to be incomplete in a way the unit tests did not show: the runtime guard in the writer still compared keys, so the retry pass undid one second later what startup reconciliation had just decided. That is folded into the same commit.

cmocka 111/111 (+5 cases), `remoted_module_utest` 891/891, framework 749 + 297, api 648, clang-format clean.

juliancnn:
### CI: the four package builds are red for a reason outside this branch

`Build Wazuh Manager (deb amd64 / deb arm64 / rpm aarch64 / rpm x86_64)` all fail at the same step, **Check embedded Python version** — the build, the install and the installed-file checks pass:

```
Embedded Python version: 3.12.14
Expected Python version: 3.12.13
Embedded Python version (3.12.14) does not match expected version (3.12.13).
```

The step compares the interpreter inside the built package against `framework/.python-version`. The published embedded-CPython artifact is already **3.12.14**; the file in the repository still says **3.12.13**. That is exactly what **#39151 (`Update embedded CPython to 3.12.14 and refresh Python dependencies`)** is landing, and its own runs of this workflow fail the same way in the meantime.

Not touched by this branch (`git diff <base>...HEAD` includes neither `framework/.python-version`, nor `packages/**`, nor the dependency bundle), and every PR that builds packages right now inherits it — a run of this workflow on an unrelated branch at 19:59 UTC was still green, before the artifact was republished. **Deliberately not fixed here**: bumping `.python-version` is #39151's change, and doing it in this PR would collide with it. These four checks should go green on a re-run once #39151 is merged.

The two reds that *were* actionable are fixed and pushed: the scan-build `core.NullDereference` in the journal loader (`36ee962959`), and the framework parity test against the effective configuration fixture, which #39070 left without `remote.legacy.ca_delivery` (`8b14e1aec1` — base drift too, but a one-line fixture addition that unblocks the suite rather than a product change).

Issue #39130, opened 2026-09-10
macOS agent reports memory usage near 98% because only free pages are counted, while Linux agents use available memory for the same field


Rebits:
## Description

The macOS agent reports `host.memory.usage` of `0.9843` (98.4%) and `host.memory.free` of 64 MiB on an otherwise idle host, while macOS itself reports 46% of memory free at the same time. The cause is that the macOS collector derives used memory as `total - free_pages`, counting reclaimable memory as used. Linux agents populate the same field from available memory instead, so the two platforms give the field incompatible meanings and only macOS looks exhausted.

Reported for the macOS agent:

```
host.memory.total  4294967296     (4096 MiB)
host.memory.free     67223552     (64 MiB)
host.memory.used   4227743744     (4032 MiB)
host.memory.usage  0.9843482971191406
```

`4294967296 - 67223552 = 4227743744`, so `used` is exactly `total - free`.

What the host actually reports, via `vm_stat` with a 16384-byte page size:

```
free=146MB inactive=804MB purgeable=7MB active=875MB wired=1043MB compressed=1147MB
```

and macOS's own summary tool disagrees sharply with the agent:

```
$ memory_pressure | tail -1
System-wide memory free percentage: 46%
```

Inactive and purgeable pages, 811 MB together here, are reclaimable on demand and are not memory pressure. Excluding them is what turns roughly half-free into 98.4% used.

The same field means something different on Linux. For agent `002` the agent reported `free = 1561817088`, and the host's `/proc/meminfo` at inspection time was:

```
MemTotal:        1950944 kB   ->  1997766656 bytes   (matches host.memory.total exactly)
MemFree:          777164 kB   ->   795815936 bytes
MemAvailable:    1477336 kB   ->  1512792064 bytes
```

The reported value tracks `MemAvailable` (1.51 GB), not `MemFree` (0.79 GB), which is why Linux agents sit at a plausible 16% to 29% usage. So Linux already excludes reclaimable cache from "used" and macOS does not.

The consequence is a false memory-exhaustion signal on every macOS endpoint. Any alerting, dashboard panel or capacity report keyed on `host.memory.usage` will show macOS hosts permanently near 100% while comparable Linux hosts show 20%, and the two numbers are not comparable even though they share a field name.

Note that this value is also currently frozen at the first scan, which is reported separately as a distinct defect. The two interact: because the hardware document is never refreshed, the incorrect ratio recorded at enrollment can never be superseded even after the host's memory profile changes.

[...]

Issue #38147, opened 2026-08-03
IT Hygiene: performance counter fields published as inventory changes cause state document churn


AwwalQuan:
### Description

Syscollector includes volatile counter fields in its inventory change diffs. Fields such as `process.utime`, `process.stime`, `host.memory.free`, `host.memory.used`, and network interface byte/packet counters always change between scans on a running host. Every scan therefore emits `type: modified` events for these fields, even on an idle host with no meaningful configuration change.

Measured on a live deployment (2026-08-02 retest, ~977 sampled IT Hygiene events):

- ~28% of events are counter diffs
- State index `document_version` climbs continuously on idle hosts:

| Agent | `wazuh-states-inventory-hardware` max version | `wazuh-states-inventory-interfaces` max version |
|-------|---:|---:|
| macOS (idle) | 11 | 11 |
| Ubuntu 24 (idle) | 7 | 7 |
| RHEL 9 | 3 | 3 |
| Windows 11 | 3 | 3 |

### Impact

No rule currently matches these events, so no spurious findings are produced today. The risk is forward-looking: any future rule on process or hardware changes will inherit this noise. The state index also accumulates unnecessary rewrites per scan.

### Fix

Exclude volatile counter fields from the inventory change comparison in syscollector, or publish them on a dedicated metrics path rather than the inventory change pipeline. Fields to exclude at minimum: `process.utime`, `process.stime`, `host.memory.free`, `host.memory.used`, and all interface byte/packet counters.

### Acceptance criteria

- [ ] On an idle host of each platform, `document_version` in `wazuh-states-inventory-hardware` and `wazuh-states-inventory-interfaces` stays flat (or at most 1) across consecutive scans.
- [ ] Process `modified` events are limited to real attribute changes: name, state, command line, parent PID, arguments.

PR #38057, opened 2026-07-28
Exclude monotonic counters from syscollector dbsync diff/checksum

lchico:
## Description

Monotonic runtime counters — accumulated CPU time, free/used memory, and interface byte/packet totals — were being compared as if they were regular inventory attributes. As a result, every syscollector scan produced a `modified` event for every process, every NIC, and the hardware record, even when nothing about the host actually changed. In one example deployment this accounted for 947 events, 24.8% of all IT Hygiene volume, across all five agents/platforms.

Closes #38017

## Proposed Changes

The fix has two parts, both required together:

1. **`wazuh_modules/syscollector/src/syscollectorImp.cpp`** — define per-table ignore lists (`HW_IGNORED_FIELDS`, `PROCESSES_IGNORED_FIELDS`, `NET_IFACE_IGNORED_FIELDS`) and pass them via dbsync's existing `input["options"]["ignore"]` mechanism in the hardware, process, and network-interface scan paths (this mechanism already existed in dbsync and was already used elsewhere in this same file, just never in these three scan paths). The same fields are also excluded from each item's checksum computation, so the checksum itself stays stable across scans where only these counters moved.
2. **`shared_modules/dbsync/src/sqlite/sqlite_dbengine.cpp`** (`SQLiteDBEngine::getRowDiff`) — dbsync's own `version`/`sync` bookkeeping columns are always present in the row diff regardless of what changed, so they were still counted as "non-ignored changes" and defeated the ignore list added in part 1. This change excludes `version`/`sync` from that check as well.

### Results and Evidence

**Bug side — pre-fix nightly build (`wazuh-agent1`)**

The issue's own "How to verify" section asks to confirm that `state.document_version` for hardware/interfaces keeps incrementing every scan, and that `dbsync_processes` only emits `modified` for `stime`/`utime`. Five sampling cycles (20s interval) confirmed exactly that mechanism on all three collectors:

- `processes` (`process.name=wazuh-modulesd`): 7 `Delta sent` occurrences across 5 cycles, every one naming only the two fields the fix now ignores:
  ```json
  {"changed_fields": ["process.stime", "process.utime"], "event_type": "modified"}
  ```
- `hardware` (singleton): 7 occurrences, always exactly:
  ```json
[...]

Issue #38017, opened 2026-07-27
syscollector: process/hardware/interface counters trigger a change event on every scan


cborla:
## Summary

Monotonic runtime counters — accumulated CPU time, free/used memory, interface byte and packet totals — are compared as if they were inventory attributes, so every scan produces a `modified` event for every process, NIC and the hardware record. In an example deployment: 947 events, 24.8% of all IT Hygiene volume, across all five agents/platforms.

## Root cause

dbsync already supports excluding specific columns from the diff comparison via `input["options"]["ignore"]` (`shared_modules/dbsync/src/sqlite/sqlite_dbengine.cpp::syncTableRowData`, lines 175-180, consumed in `getRowDiff` lines 1533-1563). This mechanism is already used today in `wazuh_modules/syscollector/src/syscollectorImp.cpp:3122` (`input["options"]["ignore"] = nlohmann::json::array({"sync"})`) — but never in the actual scan paths. For example, `scanProcesses()` (`syscollectorImp.cpp:1655-1693`) builds its `input["options"]` with only `return_old_data`:

```cpp
input["options"]["return_old_data"] = true;
// missing: input["options"]["ignore"] = [...]
```

The same gap applies to the hardware scan path (`host_memory_free`, `host_memory_used`) and the network-interface scan path (ingress/egress byte/packet counters).

## Steps to reproduce

1. Enroll an agent on any platform, leave it idle.
2. Capture consecutive `dbsync_processes` / `dbsync_hwinfo` / `dbsync_network_iface` events.
3. Observe: `modified` events fire on every scan, with `changed_fields` naming only `process.stime`/`process.utime`, `host.memory.free`/`host.memory.used`, or interface byte/packet counters — no other attribute changed.

## Expected behavior

A scan should not report a change when the only thing that moved is a monotonically increasing runtime counter that isn't itself an inventory attribute.

## Impact

- 947 events / 24.8% of IT Hygiene volume come from this, on every platform.
- Produces no findings today only because no rule matches the generic `modified` action these events receive — any future rule on process/hardware/interface changes would inherit this noise on day one.

## Suggested fix

[...]

PR #39154, opened 2026-09-10
Keep the volatile inventory counters current instead of freezing them at the first scan

### LINKED ISSUE #39131: Hardware inventory memory free/used/usage is never refreshed after the first syscollector scan, so agents report a first-boot snapshot forever
## Description

`wazuh-states-inventory-hardware` carries three volatile fields, `host.memory.free`, `host.memory.used` and `host.memory.usage`, alongside the genuinely static ones (serial number, CPU name, core count, total memory). The volatile fields are populated by the first syscollector evaluation after enrollment and are then never updated again, on any platform. Whatever the memory figures happened to be during provisioning is what the API and dashboard keep reporting.

All six agents in the environment are frozen at their provisioning timestamps, spread over the two minutes when the deployment ran:

```
  001 modified=2026-09-09T13:24:37.713Z free=1411575808 usage=0.2662
  002 modified=2026-09-09T13:24:41.494Z free=1561817088 usage=0.2182
  003 modified=2026-09-09T13:25:14.905Z free=1655296000 usage=0.1732
  004 modified=2026-09-09T13:25:15.400Z free=1615331328 usage=0.1633
  005 modified=2026-09-09T13:25:45.658Z free=5997203456 usage=0.2913
  006 modified=2026-09-09T13:26:12.958Z free=67223552   usage=0.9843
```

This is not a case of scans failing to run. Agent `003`, which was not touched during testing, has had `wazuh-modulesd` up for just over a day and has completed **25 syscollector evaluations**, the most recent at `2026/09/10 13:26:19`:

```
agent uptime of wazuh-modulesd: 1-00:49:10
syscollector evaluations in current log : 14
syscollector evaluations in rotated log : 11
first: 2026/09/10 00:25:43 wazuh-modulesd:syscollector: INFO: Evaluation finished.
last : 2026/09/10 13:26:19 wazuh-modulesd:syscollector: INFO: Evaluation finished.
```

Its hardware document is still stamped `2026-09-09T13:25:14.905Z`. Other inventory from the same agent does update normally, so this is specific to the hardware table rather than a general sync failure.

The staleness starts on the agent, not in the indexer or the sync layer. On the macOS agent the local `dbsync_hwinfo` row still holds the original values after two forced full evaluations on the day of testing, and it is marked as already synced with an unchanged checksum:

```
     serial_number = ZRFGPL2J3W
          cpu_name = Apple M1 (Virtual)
         cpu_cores = 2
         cpu_speed = 0.0
      memory_total = 4294967296
       memory_free = 67223552        <-- first-scan value
       memory_used = 4227743744
              sync = 1
          checksum = bce58f9792631a4ea9dcae1968f3c5db4e0be8b1
           version = 1
```

Live memory on that host at the time of reading was materially different, 143 MB of free pages rather than the 64 MB recorded, so the row is not merely coincidentally equal.

There is a design question for triage here, which is why this is filed as one issue rather than split: either these three fields are meant to be current state, in which case each evaluation should refresh them, or they are not, in which case exposing them in a state index that consumers read as current is the wrong place for them. Right now they are presented as current and are not. Note that this also freezes the macOS memory-accounting defect reported separately, since the incorrect ratio recorded at first scan can never be superseded.

## Environment

- QA demo environment `wqa-prod-58-demo-environment`: clustered managers (master `demo-env-master-node` ubuntu-26-amd64, worker `demo-env-worker-node` amazon-2023-arm64), one indexer, one dashboard, six enrolled agents.
- Wazuh 5.0.0 rc1, `VERSION.json` on manager and agents: `{"version": "5.0.0", "stage": "rc1", "commit": "2161e69"}`.
- Agents: `001` ubuntu-26-arm64, `002` ubuntu-26-amd64, `003` amazon-2023-amd64, `004` amazon-2023-arm64, `005` windows-11-amd64, `006` macos-26-arm64. Packages from `nightly-backup/2026-09-09`.
- Agent inspected locally: `demo-env-macos-agent` (`006`), macOS 26.5.1 build 25F80, arm64.
- Not a container deployment, so there are no image digests to record.

## Steps to reproduce

1. Deploy a 5.0.0 manager with at least one agent of any platform and let it run for several syscollector intervals (the default interval is 1h).
2. Confirm evaluations are actually running on the agent:

```bash
grep -c "syscollector: INFO: Evaluation finished" /var/ossec/logs/ossec.log
grep "syscollector: INFO: Evaluation finished" /var/ossec/logs/ossec.log | tail -1
```

3. Read the hardware documents and their `state.modified_at` from the indexer:

```bash
  "https://<indexer>:9200/wazuh-states-inventory-hardware/_search?size=10" \
  -d '{"_source":["wazuh.agent.id","state.modified_at","host.memory.free","host.memory.used","host.memory.usage"]}'
```

4. Compare `state.modified_at` against the evaluation timestamps from step 2, and compare `host.memory.free` against the host's live free memory (`/proc/meminfo` on Linux, `vm_stat` on macOS).
5. On the agent, confirm the local row has not moved either:

```bash
sqlite3 /var/ossec/queue/syscollector/db/local.db "select * from dbsync_hwinfo;"      # Linux
sqlite3 /Library/Ossec/queue/syscollector/db/local.db "select * from dbsync_hwinfo;"  # macOS
```

6. Optionally force an extra evaluation with a service restart and repeat step 5. The row does not change.

## Expected

Each syscollector evaluation refreshes `host.memory.free`, `host.memory.used` and `host.memory.usage`, so `state.modified_at` advances and the values track the host. Alternatively, if hardware inventory is deliberately a static snapshot, the volatile memory fields are not published as current state.

## Actual

The values and `state.modified_at` are pinned to the first evaluation after enrollment. Across 25 evaluations spanning just over 24 hours on an untouched agent, the hardware document never changed. On the macOS agent the local `dbsync_hwinfo` row still carries the first-scan `memory_free = 67223552` and `sync = 1` after two additional forced full evaluations, while the host actually had roughly 143 MB of free pages at the time of reading.

## Not a duplicate of

`qa-known-issues.py check` returned one weak cache candidate and nothing from the live search:

- `wazuh/wazuh-qa-automation#8762` "Syscollector VD synchronization failed" (0.50) is the transient vulnerability-detection sync warning that each agent emits once at first enrollment and then recovers. It lives in the QA automation repo, concerns the VD sync handshake, and has nothing to do with whether hardware inventory fields are refreshed.
- Live search for `syscollector refreshed inventory hardware values memory` returned nothing open.

nbertoldo:
## Description

Closes #39131

`wazuh-states-inventory-hardware` reported `host.memory.free`, `host.memory.used` and `host.memory.usage` from the first syscollector evaluation after enrollment and never updated them again, on any platform. In the QA environment all six agents were pinned to their provisioning timestamps, and agent `003` — untouched, 25 completed evaluations over just over 24 hours — still carried `state.modified_at = 2026-09-09T13:25:14.905Z`. The staleness starts on the agent: the local `dbsync_hwinfo` row kept its first-scan `memory_free` with `sync = 1` after two forced full evaluations, while the host's live free memory was materially different.

The same mechanism affects `dbsync_processes` (`utime`, `stime`) and `dbsync_network_iface` (the ingress/egress byte, packet, error and drop counters), so all three tables are fixed here.

### Root cause

The three scan paths passed their counters to dbsync as ignored columns via `input["options"]["ignore"]`, added in #38057 to stop these fields from producing a change event on every scan (#38017). In dbsync, however, suppressing the notification and skipping the write are the same decision. When the only columns that differ are in the ignore list, `SQLiteDBEngine::getRowDiff` clears `updatedData`/`oldData` while still returning `diffExist = true` (`sqlite_dbengine.cpp:1585-1589`), and `getDataToUpdate` then takes its "no changes" branch and builds a row holding only the primary keys plus `db_status_field_dm` (`sqlite_dbengine.cpp:200-214`). `updateSingleRow` runs, but the statement it emits is effectively `UPDATE <table> SET db_status_field_dm=? WHERE pk=?`, so the fresh values are never stored. Outside a transaction no `UPDATE` is issued at all.

The limitation was flagged when the fix was proposed and was not resolved then; #38017's own suggested fix assumed the opposite, that "the counters keep being stored/updated in the row... they just stop participating in the diff that triggers events". This PR makes that assumption true, by separating the two decisions rather than by changing dbsync.

### Design decision

These fields are current state, so they must be stored on the agent and synchronized to the manager as **stateful** events, keeping `wazuh-states-inventory-*` up to date. What must not happen is the **stateless** event: a scan where only a counter moved is not a change worth reporting, and that noise is what #38017 measured (947 events, 24.8% of IT Hygiene volume).

This does mean the state documents for these three tables change on every scan again, which #38017 and #38147 had also asked to stop. That is the intended trade and not a regression: the reported noise was on the stateless path, and it stays suppressed.

**What that costs, measured.** On an idle Ubuntu 22.04 agent with a 30 s interval, one scan now persists 1 hardware document, 2 of the 3 interface documents, and 13–18 process documents out of **139** rows in `dbsync_processes` — roughly a tenth of the table. The process figure is per PID whose `utime`/`stime` actually moved in the interval, not per PID in the table, so it scales with how busy the host is rather than with process count. Before this PR all three tables persisted nothing at all on such a scan, which is precisely the bug. Hardware, the case the rest of this description uses, is the 1-row floor; processes are the ceiling, and worth watching on hosts with heavy CPU churn.

#39130, the macOS memory accounting defect that this staleness was freezing, is a separate issue and is deliberately not addressed here. This PR only stops the values from being frozen; whatever macOS reports will now be refreshed on every scan, correct or not.

## Proposed Changes

All changes are in `src/wazuh_modules/syscollector/src/syscollectorImp.cpp`. **dbsync is not touched**, so FIM, `agent_info` and SCA — which use dbsync but never pass `options["ignore"]` — keep their behavior unchanged.

- Drop `input["options"]["ignore"]` from the three scan paths (`updateChanges()` for `dbsync_hwinfo` and `dbsync_network_iface`, `scanProcesses()` for `dbsync_processes`). dbsync now stores and reports these columns like any other, so the row keeps current values and the stateful event carries them.
- Add `onlyVolatileFieldsChanged(result, data, table)` and gate the stateless emission on it in `processEvent()`, where the stateful and stateless paths are already separate. The predicate compares `old` against `new` in DB-column space and skips dbsync's own bookkeeping: `version` is reported as changed on every update, and `sync` and the primary keys are carried along unchanged.
- Replace the three `*_IGNORED_FIELDS` lists with a single `VOLATILE_FIELDS_BY_TABLE` map, and use it through a new `eraseVolatileFields()` helper at the three checksum sites, which previously each excluded the same fields in a different way.
- Remove `"checksum"` from the field lists. The item checksum already excludes these counters (`getHardwareData()` erases `memory_free`/`memory_used` before hashing, and the process and interface paths do the equivalent), so "the checksum moved" now means "something real moved", which is exactly the signal the mute needs. Keeping the checksum stable also keeps integrity quiet across counter-only scans, and a stateful delta lost in flight is superseded by the next scan anyway.

### Results and Evidence

Unit tests, on this branch, `TARGET=agent` with `UNIT_TEST=ON`:

```
[==========] Running 101 tests from 3 test suites.
[==========] 101 tests from 3 test suites ran. (195381 ms total)
[  PASSED  ] 101 tests.
```

The new test is a real regression test. Stashing only the source change and rebuilding with the test in place reproduces the reported bug exactly — one stateful event per index for the whole run, and a hardware document pinned to the first scan's `memory_free`:

```
syscollectorImp_test.cpp:5005: Failure
Expected: (statefulCount["wazuh-states-inventory-hardware"]) > (1), actual: 1 vs 1
syscollectorImp_test.cpp:5006: Failure
Expected: (statefulCount["wazuh-states-inventory-processes"]) > (1), actual: 1 vs 1
syscollectorImp_test.cpp:5007: Failure
Expected: (statefulCount["wazuh-states-inventory-interfaces"]) > (1), actual: 1 vs 1
syscollectorImp_test.cpp:5018: Failure
Expected: (lastStateful["wazuh-states-inventory-hardware"]["host"]["memory"]["free"].get<int64_t>()) > (FIRST_MEMORY_FREE), actual: 1000000 vs 1000000
```

The stateless assertions (exactly one event per collector, the initial insert) pass both before and after the change, which is the evidence that #38017's noise suppression is preserved.

**A/B on a live agent.** Ubuntu 22.04 agent (`001`) against a 5.0.0 manager + indexer, syscollector `<interval>` lowered to 30 s and `<integrity_interval>` left at its 24 h default so no integrity resync could refresh the documents on its own. `wazuh_modules.debug=2`. Both arms differ only by this PR's commit, verified by the shipped library: Arm A `libsyscollector.so` md5 `637a5cc4…` (nightly), Arm B `ff481b3f…` (this PR, `refs/pull/39154/merge` = `7d2c50f`).

Arm B deliberately reuses Arm A's `local.db`, so the fixed build starts from the row that had been frozen since the previous day.

**Arm A — nightly, 26 evaluations in the window, row frozen since `2026-09-10T14:14:43` (the DB file's own birth time).**

```
 1 2026-09-11T14:31:12Z hw=3185569792/928575488/v1 ifc=4155790/5081158/v1 proc=19/10/v1 | host_memavail=3170238464 host_rx=3740775
 5 2026-09-11T14:33:12Z hw=3185569792/928575488/v1 ifc=4155790/5081158/v1 proc=19/10/v1 | host_memavail=3169603584 host_rx=3857898
10 2026-09-11T14:35:42Z hw=3185569792/928575488/v1 ifc=4155790/5081158/v1 proc=19/10/v1 | host_memavail=3168153600 host_rx=4008375
14 2026-09-11T14:37:42Z hw=3185569792/928575488/v1 ifc=4155790/5081158/v1 proc=19/10/v1 | host_memavail=3167461376 host_rx=4128967
```

Every stored value is identical across the run — the `version` column never leaves `v1`, so the row was never written — while the host's own figures moved throughout: `MemAvailable` drifted across ~7 MB and the interface received 388 KB. The indexer document matches that staleness, and is more than a day old:

```
HW  {"host":{"memory":{"used":928575488,"free":3185569792}},"state":{"document_version":1,"modified_at":"2026-09-10T14:14:44.504Z"}}
IFC {"host":{"network":{"ingress":{"bytes":171828},"egress":{"bytes":150306}}},"state":{"document_version":1,"modified_at":"2026-09-10T14:14:44.505Z"}}
```

This is the issue's own shape, with a longer lever: the reporter's agent `003` had 25 evaluations behind a frozen document; this one has 26 in the sampling window alone, on top of a full day.

**Arm B — this PR, same `local.db`, 17 evaluations.**

```
 1 2026-09-11T14:41:43Z hw=3200397312/913747968/v2  ifc=15374886/302279/v2 proc=16/8/v1   | host_memavail=3191230464 host_rx=15374886
 3 2026-09-11T14:42:43Z hw=3179601920/934543360/v4  ifc=15374886/302279/v2 proc=49/31/v3  | host_memavail=3179601920 host_rx=15374886
 6 2026-09-11T14:44:13Z hw=3173601280/940544000/v7  ifc=15374996/302459/v3 proc=97/50/v6  | host_memavail=3173879808 host_rx=15374996
11 2026-09-11T14:46:43Z hw=3168313344/945831936/v11 ifc=15374996/302459/v3 proc=162/75/v10 | host_memavail=3167657984 host_rx=15374996
14 2026-09-11T14:48:13Z hw=3168129024/946016256/v14 ifc=15374996/302459/v3 proc=215/88/v13 | host_memavail=3166420992 host_rx=15374996
```

The frozen row unfreezes on the first scan and then tracks the host: `version` climbs `v1 → v14` on hardware and `v1 → v13` on the process, and the stored values follow the host's — at cycle 3 the stored `memory_free` is `3179601920` against a host `MemAvailable` of `3179601920`, the same byte. The interface row moves only when the host counters actually move (cycle 6), and its stored ingress equals `rx_bytes` exactly at every sample.

The indexer catches up on the `<synchronization><interval>` cadence, 5 minutes by default rather than per scan, which is why the documents turn over at cycle 11 rather than cycle 1:

```
HW  {"host":{"memory":{"used":945831936,"free":3168313344}},"state":{"document_version":11,"modified_at":"2026-09-11T14:46:14.489Z"}}
IFC {"host":{"network":{"ingress":{"bytes":15374996},"egress":{"bytes":302459}}},"state":{"document_version":3,"modified_at":"2026-09-11T14:44:10.413Z"}}
```

Both documents now carry the value that is in the agent's DB (`3168313344` is cycle 11's stored `memory_free`; `15374996` is the interface's), `document_version` advances, and `state.modified_at` is minutes old instead of a day.

**Stateless silence is preserved.** Across Arm B's 17 evaluations, with the three tables being written on every one of them, the delta stream contains **zero** `dbsync_hwinfo` and **zero** `dbsync_network_iface` events:

```
      2 "collector":"dbsync_packages"
    188 "collector":"dbsync_ports"
    136 "collector":"dbsync_processes"
      2 "collector":"dbsync_services"
```

(The `dbsync_processes` entries are real process churn — creations, deletions and genuine field changes — not counter-only scans.) Arm A's stream is equally free of the two collectors, so this is a non-regression check rather than a difference between the arms: the silence #38017 asked for is still there.

**The silence is specific, not blanket.** Changing one real, non-volatile field on an interface produces a delta on the very next scan, carrying the counters along with it:

```
sudo ip link set dev eth0 mtu 1400
"changed_fields":["interface.mtu","host.network.ingress.bytes","host.network.ingress.packets","host.network.egress.bytes","host.network.egress.packets"]
"type":"modified"   "mtu":1400  (previous "mtu":1500)
```

### Artifacts Affected

`wazuh-modulesd` (syscollector module), agent side. No manager-side artifact changes.

### Configuration Changes

None.

### Documentation Updates

None. No user-facing option or field changes; the fields already exist in the schema and only their refresh behavior changes.

### Tests Introduced

- `SyscollectorImpTest.volatileCountersRefreshStateWithoutStatelessNoise` in `src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp`, replacing `ignoredCountersDoNotTriggerModifiedEventsAcrossScans`, which asserted the behavior this PR corrects (exactly one persisted event per collector). It drives hardware, process and interface counters across several scan cycles with every other field held constant, then asserts that each index received more than one stateful event, that the last hardware document carries a value later than the first scan's, and that each collector produced exactly one stateless event — the initial insert, with `notifyOnFirstScan` enabled so that the insert is reported and the mute is shown to be specific to counter-only modifications.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
+++ b/src/wazuh_modules/syscollector/src/syscollectorImp.cpp
@@ -202,21 +202,68 @@ static std::string getItemChecksum(const nlohmann::json& item)
     return Utils::asciiToHex(hash.hash());
 }
 
-// Monotonic counters excluded from dbsync's diff
-// NOTE: dbsync's "ignore" only suppresses the diff/callback when ignored fields
-// are the *only* thing that changed; it also skips persisting their new value in
-// that case (see SQLiteDBEngine::syncTableRowData). These fields will hold whatever
-// value they had at the last scan that also changed a non-ignored field.
-static const std::vector<std::string> HW_IGNORED_FIELDS { "memory_free", "memory_used", "checksum" };
-static const std::vector<std::string> PROCESSES_IGNORED_FIELDS { "utime", "stime", "checksum" };
-static const std::vector<std::string> NET_IFACE_IGNORED_FIELDS
+// Runtime counters that move on their own between scans. They are part of the inventory state, so
+// they are stored and synchronized like any other column, but a scan where only these moved is not
+// a change worth reporting: it is muted on the stateless path. They are also left out of the item
+// checksum, so integrity stays stable while the state document keeps carrying current values.
+static const std::map<std::string, std::vector<std::string>> VOLATILE_FIELDS_BY_TABLE
 {
-    "host_network_egress_packages", "host_network_ingress_packages",
-    "host_network_egress_errors",   "host_network_ingress_errors",
-    "host_network_egress_bytes",    "host_network_ingress_bytes",
-    "host_network_egress_drops",    "host_network_ingress_drops",
-    "checksum"
+    {HW_TABLE, {"memory_free", "memory_used"}},
+    {PROCESSES_TABLE, {"utime", "stime"}},
+    {
+        NET_IFACE_TABLE,
+        {
+            "host_network_egress_packages", "host_network_ingress_packages",
+            "host_network_egress_errors",   "host_network_ingress_errors",
+            "host_network_egress_bytes",    "host_network_ingress_bytes",
+            "host_network_egress_drops",    "host_network_ingress_drops"
+        }
+    }
 };
+
+static void eraseVolatileFields(nlohmann::json& item, const std::string& table)
+{
+    for (const auto& field : VOLATILE_FIELDS_BY_TABLE.at(table))
+    {
+        item.erase(field);
+    }
+}
+
+// True when the only columns that moved are the table's volatile counters. dbsync reports "version"
+// as changed on every update and carries "sync" and the primary keys along unchanged, so none of
+// those count as a change here.
+static bool onlyVolatileFieldsChanged(ReturnTypeCallback result, const nlohmann::json& data, const std::string& table)
+{
+    const auto itTable {VOLATILE_FIELDS_BY_TABLE.find(table)};
+
+    if (MODIFIED != result || VOLATILE_FIELDS_BY_TABLE.end() == itTable
+            || !data.contains("old") || !data.contains("new"))
+    {
+        return false;
+    }
+
+    const auto& oldData {data.at("old")};
+    const auto& newData {data.at("new")};
+    const auto& volatileFields {itTable->second};
+    bool anyVolatileChange {false};
+
+    for (const auto& [key, oldValue] : oldData.items())
+    {
+        if ("version" == key || "sync" == key || !newData.contains(key) || newData.at(key) == oldValue)
+        {
+            continue;
+        }
+
+        if (std::find(volatileFields.begin(), volatileFields.end(), key) == volatileFields.end())
+        {
+            return false;
+        }
+
+        anyVolatileChange = true;
+    }
+
+    return anyVolatileChange;
+}
 
 static std::string getItemId(const nlohmann::json& item, const std::vector<std::string>& idFields)
 {
@@ -361,7 +408,10 @@ void Syscollector::processEvent(ReturnTypeCallback result, const nlohmann::json&
         newData.erase("state");
     }
 
-    if (m_notify)
+    // The state document above is refreshed on every scan so the inventory stays current, but a
+    // scan where only the volatile counters moved is not reported as a change: that noise is what
+    // makes these fields unusable on the stateless path.
+    if (m_notify && !onlyVolatileFieldsChanged(result, data, table))
     {
         nlohmann::json stateless;
 
@@ -409,15 +459,6 @@ void Syscollector::updateChanges(const std::string& table,
     input["table"] = table;
     input["data"] = values;
     input["options"]["return_old_data"] = true;
-
-    if (table == HW_TABLE)
-    {
-        input["options"]["ignore"] = HW_IGNORED_FIELDS;
-    }
-    else if (table == NET_IFACE_TABLE)
-    {
-        input["options"]["ignore"] = NET_IFACE_IGNORED_FIELDS;
-    }
 
     txn.syncTxnRow(input);
     txn.getDeletedRows(callback);
@@ -1487,8 +1528,7 @@ nlohmann::json Syscollector::getHardwareData()
     sanitizeJsonValue(ret[0]);
 
     auto checksumInput = ret[0];
-    checksumInput.erase("memory_free");
-    checksumInput.erase("memory_used");
+    eraseVolatileFields(checksumInput, HW_TABLE);
     ret[0]["checksum"] = getItemChecksum(checksumInput);
     return ret;
 }
@@ -1567,12 +1607,7 @@ nlohmann::json Syscollector::getNetworkData()
                 ifaceTableData["host_network_ingress_drops"]    = item.at("host_network_ingress_drops");
 
                 auto ifaceChecksumInput = ifaceTableData;
-
-                for (const auto& field : NET_IFACE_IGNORED_FIELDS)
-                {
-                    ifaceChecksumInput.erase(field);
-                }
-
+                eraseVolatileFields(ifaceChecksumInput, NET_IFACE_TABLE);
                 ifaceTableData["checksum"] = getItemChecksum(ifaceChecksumInput);
                 ifaceTableDataList.push_back(std::move(ifaceTableData));
 
@@ -1849,14 +1884,12 @@ void Syscollector::scanProcesses()
             sanitizeJsonValue(rawData);
 
             auto checksumInput = rawData;
-            checksumInput.erase("utime");
-            checksumInput.erase("stime");
+            eraseVolatileFields(checksumInput, PROCESSES_TABLE);
             rawData["checksum"] = getItemChecksum(checksumInput);
 
             input["table"] = PROCESSES_TABLE;
             input["data"] = nlohmann::json::array( { rawData } );
             input["options"]["return_old_data"] = true;
-            input["options"]["ignore"] = PROCESSES_IGNORED_FIELDS;
 
             txn.syncTxnRow(input);
         });
--- a/src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp
+++ b/src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp
@@ -4869,8 +4869,7 @@ TEST_F(SyscollectorImpTest, hardwareMemoryUsageIsNullWhenTotalIsZero)
     SchemaValidator::SchemaValidatorFactory::getInstance().reset();
 }
 
-// End-to-end check that the per-table "ignore" lists (HW_IGNORED_FIELDS, PROCESSES_IGNORED_FIELDS,
-// NET_IFACE_IGNORED_FIELDS) actually suppress dbsync MODIFIED events across several scan cycles.
+// End-to-end check that the volatile counters stay current without being reported as changes.
 // Hardware memory, process CPU times and network byte/packet counters change on every scan below
 // (as real monotonic counters do), while every other field stays constant: each collector must
 // keep refreshing its state document on every cycle, and must report none of those cycles as a
--- a/src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp
+++ b/src/wazuh_modules/syscollector/tests/sysCollectorImp/syscollectorImp_test.cpp
@@ -13,7 +13,9 @@
 #include <cstdio>
 #include <functional>
 #include <future>
+#include <map>
 #include <memory>
+#include <mutex>
 #include <sqlite3.h>
 #include <thread>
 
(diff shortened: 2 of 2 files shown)


cborla (approved):
LGTM

Issue #39025, opened 2026-09-08
Verify with a local anchor even when ossec.conf says none


jpcerrone:
## Description

Phase 1, piece 4 of the agent-side PoC: the verification latch. If a usable anchor file is already on disk — placed by hand, however it got there — the agent verifies against it and ignores an explicit `verification_mode=none`. This is the **consuming** side of the anchor only, so it is tested by placing a test anchor by hand: no network, no mock, no token.

The premise has moved since the objective was written. `ClientConf()` resolves an unset `verification_mode` to `system`, and to `certificate` when `<certificate_authorities>` is set without an explicit mode (`src/client-agent/src/config.c`). So this is about an **explicit** `none` in `ossec.conf`, not about a compiled default.

## Tasks

- [x] Treat a present, usable anchor as the verification state: resolve to a verifying mode on the first socket, including before enrollment.
- [x] Keep an explicit `certificate` or `system` winning, so a corporate-PKI host is not forced onto the anchor.
- [x] Log the ignored downgrade at **error** level. Do not refuse to start.
- [x] Leave `w_agent_validate_ssl_ca()`'s fail-closed behaviour intact: a missing or unreadable anchor under a mode that needs one still refuses to start.

## Definition of Done

- [x] With an anchor present and an explicit `<verification_mode>none</verification_mode>`, the agent verifies, connects, and logs the ignored downgrade at error level.
- [x] With an anchor present and an explicit `certificate` or `system`, the explicit mode is honoured.
- [x] With no anchor present, behaviour is unchanged from today.
- [x] An anchor that is missing or unreadable under a verifying mode still refuses to start.
- [x] Unit tests extend `src/unit_tests/client-agent/test_agent_ssl_ca.c`.

## Out of scope

Who writes the anchor file and with what ownership. The root-before-privilege-drop, non-group-writable half of the objective's requirement is not part of this PoC — today not even `client.keys` meets it, since `etc/` is created `0770 root:wazuh` (#38949, questions 4 and 8). Atomic write and malformed-anchor behaviour are also out (#38949, question 10).

Related: #38940, #38949

fcontrerasc:
# Update 08/09/2026

## Summary

A CA bundle at `etc/certs/root-ca.pem` is now the agent's verification state. Present and readable, it decides the posture for anything `<ssl>` left unsaid, and it outranks an explicit `verification_mode=none`. Consuming side only — nothing here writes the anchor.

[...]

PR #39155, opened 2026-09-10
Adopt the manager's CA delivered over the WPK transfer channel and trust its self-signed root

### LINKED ISSUE #39071: Install the delivered manager CA during WPK upgrade and enable certificate verification

cborla:
## Description

Agent-side half of #39071. The manager-side CA delivery landed in #39080 (closing #39070): during a legacy 4.x → 5.0 WPK upgrade, the manager streams its CA into `var/incoming/root-ca.pem` over the existing `com` channel before issuing `com upgrade`. Two gaps remained on the agent side:

1. **Nothing consumed the delivered file.** `com upgrade` wipes `var/upgrade/`, and `var/incoming/` is left to the installer — but the installer never looked there, so the delivered CA was silently lost unless something moved it out before the upgrade ran.
2. **Even once adopted at `etc/certs/root-ca.pem`, the connection still failed**: `X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN` (OpenSSL error 19). The manager's TLS listener sends both the leaf and the self-signed root in its chain, and the agent's statically-linked OpenSSL had no `X509_V_FLAG_PARTIAL_CHAIN` set, so it rejected a chain ending in a trusted self-signed root instead of a CA-signed leaf.

This PR closes both gaps, and fixes a real regression found live in the pre-existing CA fallback logic that aborted the upgrade instead of trusting the implicit-anchor latch (#39025).

Closes #39071

## Proposed Changes

- **`pkg_installer.sh` / `do_upgrade.ps1`**: new `adopt_incoming_ca()`, called immediately after `DEFAULT_CA_FILE`/`$default_ca_file` is defined, before any SSL-mode gating runs.
  - Validates the incoming file (PEM markers present, size capped at 64 KiB) before installing it — `var/incoming/root-ca.pem` arrives over the same channel as everything else in the upgrade, so it gets the same scrutiny as any other untrusted input rather than being trusted implicitly.
  - Rejects a symlink/junction at that path outright rather than reading or copying through it — `var/incoming` is written by `com`'s own transfer but is not exclusively Wazuh-controlled, the same reasoning behind this codebase's own `w_fopen_nofollow()` for this exact directory.
  - Installs to `etc/certs/root-ca.pem` (`AGENT_ANCHOR_CA`, `defs.h`) with `root:wazuh 640` / dir `750`.
  - Removes the incoming file on every path (adopted, rejected, or absent) — nothing is left in `var/incoming/` afterward.
  - Never touches `ossec.conf` — the implicit anchor-driven latch (#39025) is what makes the freshly-adopted CA take effect, with no XML surgery needed.
- **Fixed a regression** in the existing `system`-mode fallback: it used to call `pin_ca()` on the just-adopted anchor and **abort the upgrade** if that failed (e.g. against a 4.x-shaped `ossec.conf` with no `<agent>` block for `pin_ca()` to parse) — reproduced live, this broke a real upgrade end-to-end. Replaced with log-only messaging: anchor present → note that #39025 will pick it up; anchor absent → warn and continue. Never aborts the upgrade on this path.
- **`curlHandle.cpp` / `curlPerformer.cpp` / `iCurlHandle.hpp` / `mockCurlHandle.hpp`**: new `ICurlHandle::trustSelfSignedRoot()`, implemented via a `CURLOPT_SSL_CTX_FUNCTION` trampoline that sets `X509_V_FLAG_PARTIAL_CHAIN` on the handle's certificate store. `applyTrustAnchors()` calls it alongside `CAInfo`, so a chain ending in a trusted self-signed root verifies correctly. See below for why this is needed and why it's safe.

### Why the self-signed root needed an explicit OpenSSL flag

The manager's TLS listener sends both the leaf and the self-signed root that signed it in its handshake chain (confirmed via `openssl s_client -showcerts`) — the same root the agent already trusts as its `CAInfo` anchor, re-sent back to it in-band. OpenSSL's default chain-building doesn't short-circuit on a peer-sent self-signed certificate just because it matches a trusted anchor, so it fails with error 19 unless `X509_V_FLAG_PARTIAL_CHAIN` is set. This doesn't weaken what gets verified (leaf signature/validity/hostname checks are unchanged) — it only changes how chain construction recognizes "already trusted." Full mechanism, the exact OpenSSL header quote, and why this belongs on the agent side rather than the manager: https://github.com/wazuh/wazuh/pull/39155#issuecomment-5625736880

### Results and Evidence

End-to-end tested against a real, from-source build (own manager binary with #39080's CA delivery, own agent binary with this PR's changes — not the nightly package) on a disposable 4.14.7 → 5.0.0 WPK upgrade, with `<ssl>` entirely absent from `ossec.conf`:

```
- Installing the CA delivered over the upgrade channel as ./etc/certs/root-ca.pem.
- A trust anchor is present at ./etc/certs/root-ca.pem. Since #39025 the upgraded agent
  verifies with 'full' against that file when <ssl> names no <verification_mode>, and uses
  it as the default <certificate_authorities>. An explicit <verification_mode> is honoured
  unchanged.
- Installation result = 0
...
- Status = connected.
- Upgrade finished successfully.
```

No `X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN`, no manual `ossec.conf` edits, no `SSL_VERIFICATION`/`WAZUH_REGISTRATION_CA` env vars — the agent verified the manager purely from the anchor delivered during the upgrade.

Full manager log, complete unedited `upgrade.log`, the no-`<ssl>`-block confirmation, and an MD5 proof that the deployed `libhttps_client.so` is byte-identical to this branch's build (not a stale binary): https://github.com/wazuh/wazuh/pull/39155#issuecomment-5625698140 — that comment also covers a separate, unrelated finding (a pre-existing double-restart race Rodrigo hit independently) that is not part of this PR's scope.

  - [x] Linux
  - [x] Windows
  - [x] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer

### Artifacts Affected

- `wazuh-agent` package installer (`pkg_installer.sh`) — Linux/macOS.
- `do_upgrade.ps1` — Windows.
- `libhttps_client.so` (agent) — new trust-anchor option on the curl handle.

### Configuration Changes

None. No new `ossec.conf` options; behavior is driven entirely by the presence or absence of `etc/certs/root-ca.pem`, per #39025's existing latch.

### Tests Introduced

- `curlPerformer_test.cpp`: `trustSelfSignedRoot()` is now asserted directly — called whenever `caPath` is set (`ConfiguredCaIsTheWholeTrustSet`, the non-Windows/macOS branch of `TlsSystemModeVerifiesPeerAndHost`), asserted absent when it isn't (`TrustAnchorsWithoutConfiguredCa`, the Windows/macOS branch), and a new `RejectedTrustSelfSignedRootAbortsBeforePerforming` test covers the fail-closed path if libcurl rejects the option. Full `https_client_unit_test` suite passes (461/461).
- `mockCurlHandle.hpp` updated with `trustSelfSignedRoot()` (default `true`), so pre-existing `curlPerformer` tests continue to pass unchanged.
- `pkg_installer.sh`/`do_upgrade.ps1` (`adopt_incoming_ca()`, the fallback-regression fix, and the symlink-rejection guard) have no automated test harness in this branch — validated via the live end-to-end reproduction above and by tracing every branch (missing file, regular file, symlink to a real target, dangling symlink) by hand, not an automated test. Flagging this as open before merge.

## Review Checklist

- [ ] Code changes reviewed
- [x] Tests cover the new functionality (https_client C++ layer; installer scripts remain E2E-only, see above)
- [x] Configuration changes documented (N/A — none introduced)
- [ ] ...


cborla:
## Full E2E evidence

Ran against a real, from-source build — not the nightly package — to exercise this PR's actual binaries end to end: a custom-built manager (with #39080's `legacy_task_deliver_ca()`) and a custom-built agent (`wazuh-agentd` + `libhttps_client.so`, both from this branch), on an isolated host, driving a real `PUT /agents/upgrade_custom` against a disposable 4.14.7 agent.

### 1. Manager: CA delivered over the legacy `com` channel

```
wazuh-manager-remoted: INFO: legacy_task_delivery: delivering remote_upgrade task '7c5a9f74-3250-07ea-f26c-d22fa09f319e' to agent '003' (wpk: 'wazuh_agent_5.0.0_ca_test_v2.wpk')
wazuh-manager-remoted: INFO: legacy_task_delivery: agent '003': delivered the manager CA as 'root-ca.pem' for task '7c5a9f74-3250-07ea-f26c-d22fa09f319e'; the agent can verify this manager after the upgrade
wazuh-manager-remoted: INFO: legacy_task_delivery: successfully delivered remote_upgrade task '7c5a9f74-3250-07ea-f26c-d22fa09f319e' to agent '003' (wpk: 'wazuh_agent_5.0.0_ca_test_v2.wpk')
```

### 2. Agent: full `upgrade.log`, unedited

```
2026/09/10 20:26:50 - Upgrade started.
2026/09/10 20:26:50 - Checking execution path.
2026/09/10 20:26:50 - Checking connectivity to 127.0.0.1:1517/wazuh-manager.
2026/09/10 20:26:50 - Manager reachable at 127.0.0.1:1517/wazuh-manager.
2026/09/10 20:26:50 - Installing the CA delivered over the upgrade channel as ./etc/certs/root-ca.pem.
2026/09/10 20:26:50 - A trust anchor is present at ./etc/certs/root-ca.pem. Since #39025 the upgraded agent verifies with 'full' against that file when <ssl> names no <verification_mode>, and uses it as the default <certificate_authorities>. An explicit <verification_mode> is honoured unchanged.
2026/09/10 20:26:50 - A trust anchor is present at ./etc/certs/root-ca.pem; the upgraded agent verifies with 'full' against it (#39025). No changes made to ossec.conf.
(Reading database ... 5498 files and directories currently installed.)
Preparing to unpack .../wazuh-agent_5.0.0_ca_test_patched.deb ...
Unpacking wazuh-agent (5.0.0-latest) over (4.14.7-1) ...
Setting up wazuh-agent (5.0.0-latest) ...
Installing new version of config file /etc/init.d/wazuh-agent ...
2026/09/10 20:26:59 - Installation result = 0
2026/09/10 20:26:59 - Checking for Wazuh Agent control script.
2026/09/10 20:26:59 - Restarting Wazuh Agent.
... (daemons stop/start, config-deprecation notices trimmed — none related to TLS/CA)
2026/09/10 20:27:07 - Waiting connection... Remaining attempts: 30.
2026/09/10 20:27:08 - Status = connected.
2026/09/10 20:27:08 - Connected to manager.
2026/09/10 20:27:08 - Upgrade finished successfully.
```

### 3. Confirmed: connection succeeded with *zero* explicit SSL configuration

```
$ grep -A5 "<ssl>\|verification_mode" /var/ossec/etc/ossec.conf
(no output — no <ssl> block at all)

$ cat /var/ossec/var/run/wazuh-agentd.state
status='connected'
```

No `X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN`, no `SSL_VERIFICATION`/`WAZUH_REGISTRATION_CA` env vars, no `ossec.conf` edits — the agent verified the manager purely from the anchor delivered during the upgrade, via #39025's implicit latch plus this PR's `trustSelfSignedRoot()` fix.

### 4. Version and binary identity, confirmed by hand

```
$ cat VERSION.json          # before: 4.14.7 → after:
{
    "version": "5.0.0",
    "stage": "rc1",
    "commit": "a3e72fd"
}

$ md5sum /var/ossec/lib/libhttps_client.so
a2249e984819fd0955851178e4560db0  /var/ossec/lib/libhttps_client.so     # deployed
a2249e984819fd0955851178e4560db0  .../src/build/lib/libhttps_client.so # this branch's build
```
Byte-identical — confirms it's actually this branch's fix running, not a stale/cached binary.

### Note on a separate, unrelated finding

While reproducing this, we also independently hit (and root-caused) the double-restart race Rodrigo found in his own testing: `pkg_installer.sh`'s unconditional `./bin/wazuh-control restart` fights with the `.deb` postinst's own `systemctl restart` on a systemd-managed host, and systemd's cgroup cleanup kills the fresh process a few seconds later. It's pre-existing (confirmed via `git log` — untouched by any of the recent CA/verification_mode work) and orthogonal to this PR: our test containers don't run systemd, so it never interfered here, and it doesn't touch anything this PR changes. Tracking separately rather than folding it into this PR's scope.

cborla:
## Why the self-signed root needed an explicit OpenSSL flag

This is the mechanism behind the `X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN` (OpenSSL error 19) this PR fixes, for anyone reviewing who hasn't hit it before.

### What the manager actually sends

`openssl s_client -connect <manager>:1517 -showcerts` shows the manager's TLS listener presenting **two** certificates in its handshake chain: the leaf (`remoted.pem`) *and* the self-signed root CA (`root-ca.pem`) that signed it. That's the same file the agent adopts as its trust anchor — so the peer is, in effect, re-sending the agent's own trust anchor back to it as part of the chain, instead of omitting it (which is what a chain against a public CA would normally do, since the client is expected to already hold the root separately and not need it repeated in-band).

### Why that trips the default OpenSSL chain build

The agent adds that same root as its `CAInfo` trust anchor, so the certificate itself is trusted — but by default, OpenSSL's chain-building logic doesn't stop as soon as it recognizes a trusted certificate. It keeps trying to build a chain up to something it considers a genuine root, and a self-signed certificate arriving *from the peer* (rather than only ever existing locally as a pre-configured anchor) doesn't automatically get treated as "chain complete, trusted" — it can still fail as an unexpected/unverifiable self-signed certificate in the presented chain, i.e. error 19.

### The fix

`X509_V_FLAG_PARTIAL_CHAIN`, straight from the vendored OpenSSL header this build compiles against (`src/external/openssl/include/openssl/x509_vfy.h:474`):

```c
/* Allow partial chains if at least one certificate is in trusted store */
#define X509_V_FLAG_PARTIAL_CHAIN 0x80000
```

Setting it (`curlHandle.cpp`, via a `CURLOPT_SSL_CTX_FUNCTION` trampoline that calls `X509_STORE_set_flags()` on the handle's cert store) tells OpenSSL: as soon as it reaches *any* certificate already present in the trusted store — self-signed or not — stop there and accept the chain, instead of continuing to look for a stricter, separately-verifiable root above it.

**This does not weaken what gets verified.** The leaf certificate's signature, validity window, and hostname/SAN matching are checked exactly as before — this flag only changes how chain *construction* recognizes "I've reached something I already trust," which is precisely the case here: the manager is re-sending the agent's own configured anchor, not a different or additional certificate the agent hasn't already decided to trust.

### Why this is the right place for the fix, not the manager

The manager including the root in its handshake chain is normal, expected behavior for a self-managed CA (there's no third party the client could otherwise fetch it from) — it's not a manager-side bug to work around, so the fix belongs on the agent's trust-store handling, which is what this PR does.

cborla:
## Version-matrix evidence: the new CA-adoption path only engages for a genuine legacy (pre-v5) migration

Ran the full compatibility matrix end to end against the same real, from-source build used in the earlier E2E evidence, to confirm `adopt_incoming_ca()` and the CA-delivery mechanism it consumes are correctly scoped to a 4.x → 5.0 migration, and never engage for a 5.0 → 5.0 upgrade.

### 1. 4.13.1 → 5.0.0 direct: blocked, as designed

```
PUT /agents/upgrade_custom (wazuh_agent_v5.0.0_ca_test.wpk) against a 4.13.1 agent
→ {"error": {"code": 1819, "message": "Direct upgrade to v5.0.0 is not supported. Please upgrade to v4.14.x first"}}
```
No task was even created — rejected at the manager's `WM_UPGRADE_INTERMEDIATE_VERSION_REQUIRED` gate (`wm_agent_upgrade_validate.c`) before anything touched the legacy delivery path.

### 2. 4.13.1 → 4.14.7: success (plain legacy-to-legacy upgrade, unrelated to this PR)

```
VERSION.json: {"version": "4.14.7", ...}
upgrade.log:  ... Status = connected. / Upgrade finished successfully.
```

### 3. 4.14.7 → 5.0.0: success, through this PR's mechanism

```
manager log:
  legacy_task_delivery: delivering remote_upgrade task '...' to agent '004' (wpk: '...')
  legacy_task_delivery: agent '004': delivered the manager CA as 'root-ca.pem' for task '...'
  legacy_task_delivery: successfully delivered remote_upgrade task '...' to agent '004'

agent upgrade.log:
  - Installing the CA delivered over the upgrade channel as ./etc/certs/root-ca.pem.
  - A trust anchor is present at ./etc/certs/root-ca.pem. Since #39025 the upgraded agent
    verifies with 'full' against that file when <ssl> names no <verification_mode>...
  - Status = connected.
  - Upgrade finished successfully.

VERSION.json: {"version": "5.0.0", ...}
grep -A5 "<ssl>|verification_mode" ossec.conf → (no output -- no <ssl> block at all)
```

### 4. 5.0.0 → 5.0.0 forced: success, *without* this PR's mechanism ever engaging

Same agent, now already on 5.0.0, pushed through another 5.0.0 WPK:

```
manager log for this task: zero "legacy_task_delivery" lines.
agent log instead shows: wazuh-agentd: ... https_client: remote_upgrade task '...': ...
```
Structurally confirms `legacy_task_agent_is_pre_v5()` (`legacy_task_delivery.c`) correctly excludes an agent already confirmed ≥ v5.0.0 — it's routed entirely through the modern HTTPS control channel instead, never reaching `legacy_task_deliver_ca()` or this PR's `adopt_incoming_ca()` at all.

```
agent upgrade.log for this run:
  - A trust anchor is present at ./etc/certs/root-ca.pem. Since #39025 ...   <- pre-existing gate log only
  (no "Installing the CA delivered over the upgrade channel" line this time)
  Unpacking wazuh-agent (5.0.0-latest) over (5.0.0-latest) ...
  - Status = connected.
  - Upgrade finished successfully.
```
The absence of the "Installing the CA delivered" line here (present in test 3, absent in test 4) is the direct agent-side confirmation that nothing arrived at `var/incoming/root-ca.pem` for this upgrade — `adopt_incoming_ca()` ran (as it always does, unconditionally) and correctly found nothing to adopt.

cborla:
## Version-matrix evidence, Windows: same gating behavior confirmed as Linux

Repeated the version-matrix test from the earlier comment against a real Windows 11 agent (WinRM/VirtualBox VM, `do_upgrade.ps1` from this branch, real `wazuh-agent-*.msi` packages from `packages.wazuh.com`), talking to the same manager over a real network hop (not loopback).

### 1. 4.13.1 → 5.0.0 direct: blocked, same as Linux

```
PUT /agents/upgrade_custom (wazuh_agent_v5.0.0_win_test.wpk) against a 4.13.1 Windows agent
→ {"error": {"code": 1819, "message": "Direct upgrade to v5.0.0 is not supported. Please upgrade to v4.14.x first"}}
```
No task created — same manager-side gate as Linux, platform-agnostic (it runs before any WPK, Linux or Windows, is touched).

### 2. 4.13.1 → 4.14.7: success (plain legacy-to-legacy upgrade)

```
VERSION.json: {"version": "4.14.7", ...}
ossec.log: ... (4102): Connected to the server ([192.168.100.120]:1514/tcp).
```

### 3. 4.14.7 → 5.0.0: success, through this PR's `do_upgrade.ps1` mechanism

```
do_upgrade.ps1's own upgrade.log:
  - Checking connectivity to 192.168.100.120:1517 (endpoint: 'wazuh-manager').
  - Manager reachable at 192.168.100.120:1517 (endpoint: 'wazuh-manager').
  - Replacing the existing anchor at C:\Program Files (x86)\ossec-agent\certs\root-ca.pem
    with the CA delivered over the upgrade channel; the manager is authoritative for its own CA.
  - A trust anchor is present at C:\...\certs\root-ca.pem. Since #39025 the upgraded agent
    verifies with 'full' against that file when <ssl> names no <verification_mode>...
  - Installation finished.

VERSION.json: {"version": "5.0.0", ...}
ossec.conf: no <ssl> block, no verification_mode entry.
ossec.log: wazuh-agentd:https-client: INFO: Starting https_client (server=192.168.100.120:1517, agent=005, ...)
```
Same outcome as Linux: the agent adopted the delivered CA and connected under the implicit `full`-verification latch with zero manual configuration.

### 4. 5.0.0 → 5.0.0 forced: incomplete — blocked by an unrelated, pre-existing manager-certificate gap, not by this PR

Triggering another 5.0.0 WPK against the now-5.0.0 Windows agent reproduced the Linux result structurally (**zero** new `legacy_task_delivery` log lines for this task — it's excluded from the legacy poller exactly like the Linux case), but the task never completed delivery because the agent's *regular* connection had already dropped:

```
agent log, right after test 3 finished:
  wazuh-agent: WARNING: Manager unreachable. Pausing module event production. Reason: (60)
  SSL peer certificate or SSH remote key was not OK: SSL: no alternative certificate subject
  name matches target ipv4 address '192.168.100.120'

GET /agents?agents_list=005 → {"status": "disconnected"}
```

Root cause, confirmed directly on the manager: `remoted.pem`'s SAN list is `127.0.0.1, wazuh-manager, localhost, host.docker.internal, wazuh-1` — this test manager's certificate (generated for same-host/container use) was never provisioned with `192.168.100.120`, the real LAN address this Windows VM actually dials. All the Linux tests used `127.0.0.1` (same-host containers), which incidentally matches a SAN entry that's always been there — so this gap was invisible until a genuinely separate host connected over the network.

**This is not a bug in this PR.** `adopt_incoming_ca()`/`trustSelfSignedRoot()` did exactly what they're supposed to — the CA was adopted and full verification correctly kicked in; it's *because* verification is now strict that a pre-existing SAN gap in this specific test environment's certificate became visible. It's the exact class of issue Rodrigo's own PR #39080 already calls out (`leafHasUsableSan()`'s start-up warning for a certificate with no SAN beyond loopback/host name) — a deployment-time certificate-provisioning concern, not something this PR introduces or needs to fix. Test 4 is left unresolved on Windows for this run rather than working around it by weakening verification; happy to redo it once the test manager's cert is reissued with the right SAN, but the structural point (this PR's code path is skipped for a confirmed-≥v5.0.0 agent) is already independently confirmed on both platforms.

PR #38819, opened 2026-09-02
Replace the manager's pseudo-XML configuration parser with a schema-validated strict-XML loader

juliancnn:
## Description

The manager read `etc/wazuh-manager.conf` with `os_xml`, a hand-written pseudo-XML parser that
implements a private dialect (merged multiple roots, raw `&`/`<` in values, undecoded entities, legacy
`<! !>` comments) with no schema: every daemon walked the tree with its own reader, unknown options
were diagnosed inconsistently, `logging`/`cluster`/`indexer` were validated by no daemon's `-t`, and
the Python framework parsed the same file a second time with different tolerances.

This PR replaces that path with `shared_modules/manager_config`, a manager-only loader (pugixml +
rapidjson) that turns the file into the **effective configuration** — strict-XML parse → canonical JSON
typed by an embedded draft-04 JSON Schema → validation → defaults → cross-field semantics — and makes
every consumer read that one document. The file stays XML in 5.0; the YAML migration (#38714) is
deferred to 5.1 and maps 1:1 onto the same schema.

Closes #38816

## Proposed Changes

- **`src/shared_modules/manager_config` (new)**: strict-XML→canonical-JSON transform, embedded schema
  (single source of truth: types, ranges, defaults, descriptions), defaults filling, cross-field
  semantics (port clashes, certificate pairing/existence), C++ API + exception-free C API, GTest unit
  suite with vectors and a C++↔schema parity test.
- **CLI `bin/wazuh-manager-conf validate|get|dump`**: the operator/installer/framework entry point;
  all diagnostics follow one contract — `(1244): Invalid configuration at '<pointer or file>': <detail>`.
- **C daemons** (remoted, authd, wazuh-db, modulesd) read their effective sections as cJSON
  through libconfig's `w_mconf_*()`; `-t` validates the whole file; `getconfig` returns the effective
  sections. The libwazuh cluster getters and `os_logging_config()` read through a provider hook, so
  `libwazuhshared.so` stays dependency-free for the engine's `dlopen`. The Task Manager's
  disconnection sweep (#38757) reads `global.agents_disconnection_time` through
  `w_mconf_section()`/`Read_Global_JSON` instead of the removed XML path.
- **Engine**: a new `base::managerConfig` wrapper links the library, feeds the indexer connector,
  enriches events with `cluster.{name,node_name}`, honours `logging.log_format` (it never did), and
  `-t` validates the file. `wazuh-manager-control` validates with the CLI and fails fast with the JSON
[...]

PR #39184, opened 2026-09-11
Fix indexer SSL configuration handling in get_indexer_client()

### LINKED ISSUE #39113: Manager never starts any indexer task because get_indexer_client raises AttributeError on the ssl.certificate_authorities shape
## Description

`get_indexer_client()` reads `indexer.ssl` from the manager configuration and expects `certificate_authorities` to be a list of dicts (`[{"ca": [...]}]`) and `certificate` / `key` to be lists of strings. `get_manager_conf(section="indexer")` returns `certificate_authorities` as a list of plain strings and `certificate` / `key` as plain strings.

The mismatch makes this line in `framework/wazuh/core/indexer/indexer.py` raise on every call:

```python
("ca_certs", ssl_config.get("certificate_authorities", [{}])[0].get("ca", [])),
```

`ssl_config["certificate_authorities"][0]` is the string `etc/certs/root-ca.pem`, so `.get("ca", [])` raises `AttributeError: 'str' object has no attribute 'get'`.

The two lines below it are wrong in the same way and would fail even if the first were fixed:

```python
client_cert = resolve_wazuh_path(ssl_config["certificate"][0])   # -> 'e', the first character
ca_certs    = resolve_wazuh_path(ssl_config["certificate_authorities"][0]["ca"][0])
```

`ManagedIndexerTasks.manage_indexer_tasks` catches this with a bare `except Exception` and reports it as a configuration or availability problem, so the real cause never reaches the log:

```python
except Exception as e:
    self.logger.warning(f"Indexer is not configured or unavailable: {e}.")
```

Because the failure is deterministic rather than transient, the retry loop never succeeds and backs off exponentially (300s, 600s, ... capped at 3600s), so **no indexer task ever starts**. On the master that disables all four:

- `disconnected_agent_sync.run_agent_groups_sync`
- `active_response_task.run`
- `disconnected_agent_sync.run_cluster_name_sync`
- `metrics_snapshot.run_metrics_snapshot`

and on the worker it disables `active_response_task.run`.

The observable consequence is that the three metrics data streams stay empty and active response never runs, on a stack where the indexer is healthy and correctly configured.

## Environment

- Topology: clustered managers (`wazuh.master` master, `wazuh.worker` worker), single indexer (`wazuh1.indexer`), single dashboard, from `wazuh-docker` branch `5.0.0` (`ad231ed`), multi-node, with `wazuh2.indexer` / `wazuh3.indexer` removed and all indexer references pointed at `wazuh1.indexer`.
- Images:
  - `public.ecr.aws/wazuh-cicd/wazuh/wazuh-manager:5.0.0-latest` `sha256:10219f399ab4d49ef990453dc8ab10f61f8d2dcb4504c83932f4802c422c18ce` (created 2026-09-10T03:24:54Z)
  - `public.ecr.aws/wazuh-cicd/wazuh/wazuh-indexer:5.0.0-latest` `sha256:9618a53cd049feae909c71792506c807c3a10806b7e3675fa9ca024d5b214e82` (created 2026-09-10T03:21:52Z)
  - `public.ecr.aws/wazuh-cicd/wazuh/wazuh-dashboard:5.0.0-latest` `sha256:bca8c4591aa8535ba0db0131ba4fea55da0037bb14a15cc27ed43508d95bca7b` (created 2026-09-10T03:24:28Z)
  - `public.ecr.aws/wazuh-cicd/wazuh/wazuh-agent:5.0.0-latest` `sha256:0143a5d0227b6d4c36103fbfb04fd31e7ee2cbdd8949db6e44990bfdb91d7779` (created 2026-09-10T03:23:12Z)
- Host: Linux 6.12.10, Docker 29.4.0, Docker Compose v5.1.3.

## Steps to reproduce

1. Deploy the multi-node stack above from a clean state (`docker compose up -d`), with the standard generated certificates so `indexer.ssl` is populated normally.
2. Wait for `wazuh.master` and `wazuh.worker` to report healthy and for the indexer to be green.
3. Read the cluster log on either manager: `docker exec multi-node-wazuh.master grep "Indexer is not configured or unavailable" /var/wazuh-manager/logs/cluster.log`
4. Confirm the indexer is in fact reachable and correctly configured: `docker exec multi-node-wazuh.master sed -n '/<indexer>/,/<\/indexer>/p' /var/wazuh-manager/etc/wazuh-manager.conf`
5. Reproduce the exception directly:

   ```bash
   docker exec multi-node-wazuh.master sh -c 'cd /tmp && /var/wazuh-manager/framework/python/bin/python3 -c "
   from wazuh.core.configuration import get_manager_conf
   ssl_config = get_manager_conf(section=\"indexer\")[\"indexer\"][\"ssl\"]
   print(ssl_config)
   ssl_config.get(\"certificate_authorities\", [{}])[0].get(\"ca\", [])"'
   ```
6. Register an agent and confirm the metrics streams never receive a document: `docker exec multi-node-wazuh1.indexer curl -sk -u admin:admin "https://localhost:9200/_cat/indices/.ds-wazuh-metrics*?h=index,docs.count"`

## Expected

`get_indexer_client()` builds a client from the configured certificate paths, the indexer task manager starts the four indexer tasks, and the `wazuh-metrics-agents`, `wazuh-metrics-comms-v4` and `wazuh-metrics-normalization` data streams receive documents for the registered agent.

If the configuration really were unusable, the log should name the offending option rather than report a generic availability problem.

## Actual

Both managers log, every backoff interval:

```
2026/09/10 10:08:47 WARNING: [Master] [Main] Indexer is not configured or unavailable: 'str' object has no attribute 'get'.
2026/09/10 10:13:48 WARNING: [Master] [Main] Indexer is not configured or unavailable: 'str' object has no attribute 'get'.
2026/09/10 10:08:53 WARNING: [Worker worker01] [Main] Indexer is not configured or unavailable: 'str' object has no attribute 'get'.
```

Direct reproduction:

```
ssl_config = {'certificate_authorities': ['etc/certs/root-ca.pem'], 'certificate': 'etc/certs/indexer-connector.pem', 'key': 'etc/certs/indexer-connector-key.pem'}
certificate_authorities = ['etc/certs/root-ca.pem'] -> type str
indexer.py does: ssl_config.get("certificate_authorities", [{}])[0].get("ca", [])
  -> AttributeError: 'str' object has no attribute 'get'
ssl_config["certificate"][0] = 'e'   (first character, not first path)
```

All three metrics data streams stay empty with an active agent enrolled:

```
.ds-wazuh-metrics-comms-v4-000001      0
.ds-wazuh-metrics-agents-000001        0
.ds-wazuh-metrics-normalization-000001 0
```

The condition survives `docker compose restart`.

## Regression origin

`framework/wazuh/core/indexer/indexer.py` has not been touched since `a9752887` (2026-09-02), and that commit left the SSL handling in it unchanged. What changed under it is the producer: PR #38819, "Replace the manager's pseudo-XML configuration parser with a schema-validated strict-XML loader" (merged 2026-09-03), made `get_manager_conf` return `certificate_authorities` as a list of strings and `certificate` / `key` as plain strings, where the previous pseudo-XML parser produced `[{"ca": [...]}]` and one-element lists.

So the shape of the configuration changed and this consumer was not migrated with it. The same manager-config series is still landing changes on 5.0.0 (`5ee7c530` on 2026-09-09 and #38773 on 2026-09-10), so it is worth checking whether any other consumer reads the old shape.

## Not a duplicate of

`qa-known-issues.py check` returned **no cache match** against the 781 open `type/bug` issues, and its live GitHub search for "get_indexer_client attributeerror attribute indexer streams metrics" returned nothing open. Verdict line: "no known issue looks like this one."

## Note for triage

This also means PR #38980 (`wazuh.agent.groups` in the agent metrics stream) and PR #39008 (active response schema validation) cannot be exercised on this build at all: the tasks that would produce those documents never start.

Antoniogm03:
## Description

`get_indexer_client()` in `framework/wazuh/core/indexer/indexer.py` read `indexer.ssl.certificate_authorities` as a list of dicts (`[{"ca": [...]}]`) and `certificate`/`key` as one-element lists: the shape the old pseudo-XML manager configuration parser produced. PR #38819 replaced that parser with a schema-validated strict-XML loader that returns `certificate_authorities` as a list of strings and `certificate`/`key` as plain strings, and this consumer was never migrated.

Every call raised `AttributeError: 'str' object has no attribute 'get'`. `IndexerTaskManager.manage_indexer_tasks` catches this with a bare `except Exception` and logs it as "Indexer is not configured or unavailable", so the real cause never reached the log. Because the failure is deterministic, the retry loop with exponential backoff (300s to 3600s) never succeeded, and no indexer-dependent task ever started: `disconnected_agent_sync`, `active_response_task`, and `metrics_snapshot` on the master, and `active_response_task` on the worker, on a manager with a correctly configured, reachable indexer.

While fixing that, two more related defects in the same function surfaced:

- `get_indexer_client()` also required `certificate`, `key`, and `certificate_authorities` to be non-empty, even though the configuration schema documents all three as optional (`certificate`/`key` default to `""`, "Empty = no client certificate"; `certificate_authorities` defaults to `[]`), the schema's config loader deliberately excludes `indexer.ssl.*` from its file-must-exist checks (those certificates are provisioned by the operator, not by `install.sh`, so a fresh manager must start without them), and the manager's own C++ indexer connector (`indexerTransport.cpp`) already treats them as optional, falling back to basic auth over the system trust store. A manager configured to talk to the indexer without a client certificate hit the exact same "Indexer is not configured or unavailable" failure this PR otherwise fixes.
- The SSL context only ever trusted `certificate_authorities[0]`, silently discarding every other CA in the list. The C++ indexer connector merges every configured CA into one trust bundle (`Utils::CertHelper::mergeCaRootCertificates`), so a CA rotation (old and new CA both listed during the rotation window) works for the C++ connector but breaks the Python client the moment the indexer presents a certificate signed by the second CA.
- The call that builds the SSL context was the only step in the function not wrapped in `try/except`, unlike every other failure point in it. A bad certificate/key/CA path (a typo, a permission issue, a corrupt PEM) raised a raw `FileNotFoundError`/`OSError`/`ssl.SSLError` that escaped unwrapped and reached `manage_indexer_tasks()`'s generic "Indexer is not configured or unavailable" handler with no indication that the SSL context build is what actually failed.
- A blank entry in `certificate_authorities` (a `<ca></ca>` mixed into an otherwise populated list) was silently dropped instead of rejected, unlike the symmetric certificate/key pairing check right next to it. This is not the same as an empty list, which is a valid, schema-documented "use the system trust store" configuration; a blank entry inside a populated list can never resolve to a usable file and was masking a likely misconfiguration behind a connection that appeared to work fine using whatever entries were left.

Closes #39113

## Proposed Changes

- `framework/wazuh/core/indexer/indexer.py`:
  - Read `certificate`/`key` as plain strings and `certificate_authorities` as a list of strings, matching the shape `get_manager_conf()` actually returns today, instead of the shape the old pseudo-XML parser produced.
  - Stop requiring `certificate`, `key`, and `certificate_authorities` to be present. The only remaining validation is that `certificate` and `key` must be set together (a lone one of the two can never work); an empty `certificate_authorities` now falls back to the system trust store instead of failing.
  - `_create_ssl_context()` now accepts `None` for the client certificate/key (skips `load_cert_chain()`) and a list for the CAs: the first CA is loaded through `ssl.create_default_context()` (or the system trust store if the list is empty), and every additional CA is trusted on top of it via `load_verify_locations()`, which accumulates rather than replacing.
- `framework/wazuh/core/indexer/tests/test_indexer.py`: updated the SSL config fixture in `test_get_indexer_client_resolves_relative_certificate_paths` to the real shape (it used the old shape too, so it passed without exercising the shape regression), and added cases for no `<ssl>` section, an empty `<ssl>` section, CA-only, `certificate` without `key`, multiple CAs resolved, `_create_ssl_context()` trusting every CA in the list, a bad certificate path raising `IndexerUnavailableError` instead of a raw `FileNotFoundError`, and a blank `certificate_authorities` entry (alone or mixed with a valid one) being rejected instead of silently dropped. Also added a case that reads the schema-defaults `indexer.ssl` from `test_configuration.py`'s frozen effective-config fixture instead of hand-typing it, so a future shape change to the config loader's output breaks this test visibly instead of silently matching a stale literal, the exact gap that let #39113 through. This only covers the empty/default case; no shared fixture carries a populated `indexer.ssl`, so the certificate-paths happy path still needs its own literal.
- `CHANGELOG.md`: added all five fixes to the Fixed entry under Manager, referencing #39113.

### Results and Evidence

Reproduced and verified against a real installed Wazuh 5.0.0 manager (cluster: 1 master + 2 workers, 1 indexer, 1 dashboard, 4 agents) running in a devcontainer.

**Before the fix**, `cluster.log` on the master shows one entry per backoff retry:

```
2026/09/11 07:34:13 WARNING: [Master] [Main] Indexer is not configured or unavailable: 'str' object has no attribute 'get'.
2026/09/11 07:39:14 WARNING: [Master] [Main] Indexer is not configured or unavailable: 'str' object has no attribute 'get'.
2026/09/11 07:49:14 WARNING: [Master] [Main] Indexer is not configured or unavailable: 'str' object has no attribute 'get'.
2026/09/11 07:59:04 WARNING: [Master] [Main] Indexer is not configured or unavailable: 'str' object has no attribute 'get'.
2026/09/11 08:04:04 WARNING: [Master] [Main] Indexer is not configured or unavailable: 'str' object has no attribute 'get'.
2026/09/11 08:14:04 WARNING: [Master] [Main] Indexer is not configured or unavailable: 'str' object has no attribute 'get'.
```

Direct reproduction of the underlying exception, against the real running configuration:

```
$ /var/wazuh-manager/framework/python/bin/python3 -c "
from wazuh.core.configuration import get_manager_conf
ssl_config = get_manager_conf(section='indexer')['indexer']['ssl']
print(ssl_config)
ssl_config.get('certificate_authorities', [{}])[0].get('ca', [])
"
{'certificate_authorities': ['etc/certs/root-ca.pem'], 'certificate': 'etc/certs/indexer-connector.pem', 'key': 'etc/certs/indexer-connector-key.pem'}
Traceback (most recent call last):
  ...
AttributeError: 'str' object has no attribute 'get'
```

The `wazuh-metrics-*` data streams stayed at 0 documents:

```
$ curl -sk -u admin:admin "https://localhost:9200/_cat/indices/.ds-wazuh-metrics*?h=index,docs.count"
.ds-wazuh-metrics-comms-v4-000001      0
.ds-wazuh-metrics-agents-000001        0
.ds-wazuh-metrics-normalization-000001 0
```

**After the fix**, verified twice independently: once by reinstalling only the fixed `framework` package into the running manager, and again against a manager fully recompiled and installed from scratch from this branch (new binaries, mtimes confirmed).

`cluster.log` after restart shows no warning; tasks start immediately:

```
2026/09/11 08:49:30 INFO: [Master] [Main] Checking if the indexer is available to initiate indexer tasks.
2026/09/11 08:49:30 INFO: [Master] [Main] Indexer is available. Starting indexer tasks.
2026/09/11 08:49:30 INFO: [Master] [disconnected_agent_sync_task] Starting non-connected agent group synchronization task (interval: 300s, batch_size: 100)
2026/09/11 08:49:30 INFO: [Master] [Active Response] Starting
```

Full end-to-end call, with a real TLS connection to the indexer:

```
$ sudo /var/wazuh-manager/framework/python/bin/python3 -c "
import asyncio
from wazuh.core.indexer.indexer import get_indexer_client

async def main():
    async with get_indexer_client() as client:
        print('healthcheck OK:', await client.healthcheck())

asyncio.run(main())
"
healthcheck OK: None
```

Unit tests:

```
$ python3 -m pytest wazuh/core/indexer/tests/test_indexer.py -q
21 passed, 180 warnings in 0.18s
```

The `wazuh-metrics-*` streams remained at 0 documents in this environment even after the fix, but that is unrelated to this bug: the registered agents were already `disconnected` (since before this investigation started), so the now-running `metrics_snapshot` task has no live telemetry to write. `disconnected_agent_sync_task` and `Active Response` run on every cycle, which they never did before the fix.

**Optional certificate/CA fix**, verified against the same live indexer by loading this branch's `indexer.py` (with `WAZUH_PATH`/`KEY_STORE_SOCKET` pointed at the real install) and stubbing only the config it reads:

```
# indexer.ssl = {} (no certificate, no key, no certificate_authorities)
$ python3 -c "... async with get_indexer_client() as client: ..."
raised: IndexerUnavailableError Error 2200 - ... [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate ...
```

This is the expected, correct outcome: with no CA configured, the code now falls back to the system trust store (instead of failing during config parsing as before), and the system trust store legitimately does not know the indexer's private CA; this is a real TLS trust failure, not the config bug this PR fixes.

```
# indexer.ssl = {"certificate_authorities": ["etc/certs/root-ca.pem"]}  (no certificate, no key)
$ python3 -c "... async with get_indexer_client() as client: print(await client.healthcheck()) ..."
healthcheck OK (no client cert, only CA): None
```

With the real CA configured and no client certificate at all, the connection succeeds end to end against the running indexer: this confirms the indexer accepts basic-auth-only connections, matching the C++ connector's behavior, which `get_indexer_client()` previously rejected.

**Multi-CA fix**, verified against the same live indexer with a genuinely unrelated self-signed CA (generated for this test) to make the negative case meaningful:

```
# certificate_authorities = ["/tmp/decoy-ca.pem"]  (unrelated CA, alone)
$ python3 -c "... async with get_indexer_client() as client: ..."
raised (expected): ... [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate ...

# certificate_authorities = ["/tmp/decoy-ca.pem", "etc/certs/root-ca.pem"]  (unrelated CA first, real CA second)
$ python3 -c "... async with get_indexer_client() as client: print(await client.healthcheck()) ..."
healthcheck OK (decoy first, real CA second): None
```

The unrelated CA alone fails verification, as expected. Listed together with the real CA (in second position), the connection succeeds, proving every CA in the list is actually trusted, not just `certificate_authorities[0]`.

**Unwrapped SSL context error fix**, verified against the same live indexer with a nonexistent CA path:

```
# certificate_authorities = ["etc/certs/typo-root-ca.pem"]  (does not exist)
$ python3 -c "... async with get_indexer_client() as client: ..."
raised: IndexerUnavailableError - Error 2200 - Error connecting to the Indexer service: Failed to build SSL context: [Errno 2] No such file or directory
```

Instead of a raw, unwrapped `FileNotFoundError`, the failure is now a properly typed `IndexerUnavailableError` naming the SSL context build as the point of failure.

Full command-by-command reproduction log (before and after, all four fixes) is available on request; it is not included in this repository.

### Artifacts Affected

- The `wazuh` Python framework package (`framework/`), pip-installed into the manager at `/var/wazuh-manager/framework/python/lib/python3.12/site-packages/wazuh/`. Consumed by `wazuh-manager-clusterd` (the indexer task manager) and any other framework/API code that calls `wazuh.core.indexer.indexer.get_indexer_client()`.
- No executables, default configuration files, or packaging specs changed.

### Configuration Changes

N/A: no configuration schema, default value, or template changed.

### Tests Introduced

Added 10 new cases to `framework/wazuh/core/indexer/tests/test_indexer.py`: no `<ssl>` section, an empty `<ssl>` section, the schema-defaults `indexer.ssl` sourced from `test_configuration.py`'s frozen effective-config fixture instead of a hand-typed literal, `certificate_authorities` only (no client cert), `certificate` set without `key` (still an error), multiple CAs resolved to full paths, `_create_ssl_context()` trusting every CA in the list (not just the first, verified via the mocked `load_verify_locations` call), a nonexistent certificate path raising `IndexerUnavailableError` instead of a raw `FileNotFoundError`, and a blank `certificate_authorities` entry raising the same way whether it is alone or mixed with a valid entry. Also corrected the SSL config fixture in the existing `test_get_indexer_client_resolves_relative_certificate_paths` and `test_create_ssl_context_caching` to the shape `get_manager_conf()` actually returns; the former previously used the old pseudo-XML parser's shape and therefore passed without exercising the shape regression, and the latter passed a single CA string where the function now expects a list.

## Review Checklist

- [ ] Code changes reviewed


framework/wazuh/core/indexer/indexer.py
```diff
@@ -554,20 +554,20 @@ async def get_indexer_client() -> AsyncIterator[Indexer]:
...
             raise IndexerUnavailableError(
                 code=2200, extra_message=f"Missing or empty {cert_name} path"
             )
 
-    client_cert = resolve_wazuh_path(ssl_config["certificate"][0])
-    client_key = resolve_wazuh_path(ssl_config["key"][0])
-    ca_certs = resolve_wazuh_path(ssl_config["certificate_authorities"][0]["ca"][0])
+    client_cert = resolve_wazuh_path(ssl_config["certificate"])
+    client_key = resolve_wazuh_path(ssl_config["key"])
+    ca_certs = resolve_wazuh_path(ssl_config["certificate_authorities"][0])
```
Antoniogm03: **[S2 · blocking · introduced]** `certificate`, `key` and `certificate_authorities` are still hard-required, but `wazuh-manager.schema.json` lists only `hosts` as required and defaults these to `""`/`""`/`[]`, with `ssl.certificate` documented as "Empty = no client certificate" - so a manager with `<indexer><hosts>https://indexer:9200</hosts></indexer>` and no `<ssl>` raises `IndexerUnavailableError(2200, "Missing or empty client_cert path")` on every call, backs off to 3600s and never starts `disconnected_agent_sync`, `active_response_task` or `metrics_snapshot`: the exact symptom #39113 reports. The C++ consumer of the same section treats all three as optional (`indexerTransport.cpp` guards each with `contains()`; `indexerConnector.hpp:134` says so explicitly), so after this PR the Python framework is the only consumer that refuses a schema-valid document. Treating the material as optional also removes the double read of `ssl_config` between lines 557-559 and 568-570.

```suggestion
    # TLS material is optional per the configuration schema: an empty certificate/key
    # means no client certificate, an empty CA list means the system trust store.
    certificate = ssl_config.get("certificate") or ""
    key = ssl_config.get("key") or ""
    cas = [ca for ca in (ssl_config.get("certificate_authorities") or []) if ca]

    if bool(certificate) != bool(key):
        raise IndexerUnavailableError(
            code=2200,
            extra_message="indexer.ssl.certificate and indexer.ssl.key must be set together",
        )

    client_cert = resolve_wazuh_path(certificate) if certificate else None
    client_key = resolve_wazuh_path(key) if key else None
    ca_certs = resolve_wazuh_path(cas[0]) if cas else None
```

Companion guard in `_create_ssl_context` (around line 121). `ssl.create_default_context` already loads the system trust store when `cafile` is `None`, so only `load_cert_chain` needs the condition:

```python
        context = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH, cafile=ca_certs)
        if client_cert:
            context.load_cert_chain(certfile=client_cert, keyfile=client_key)
```

framework/wazuh/core/indexer/indexer.py
```diff
@@ -554,20 +554,20 @@ async def get_indexer_client() -> AsyncIterator[Indexer]:
 
     # Validate SSL certificate paths
     required_cert_paths = [
-        ("client_cert", ssl_config.get("certificate", [])),
-        ("client_key", ssl_config.get("key", [])),
-        ("ca_certs", ssl_config.get("certificate_authorities", [{}])[0].get("ca", [])),
+        ("client_cert", ssl_config.get("certificate", "")),
+        ("client_key", ssl_config.get("key", "")),
+        ("ca_certs", (ssl_config.get("certificate_authorities", []) or [""])[0]),
```
Antoniogm03: **[S3 · non-blocking · introduced]** Only the first entry of `certificate_authorities` is used, so every other configured CA is silently discarded, while `indexerTransport.cpp` merges all of them through `Utils::CertHelper::mergeCaRootCertificates` when there is more than one - which is what makes a CA rotation work for modulesd and the engine but not for the framework. The same indexed read also rejects `["", "etc/certs/root-ca.pem"]`, where the second CA is perfectly usable. Carrying the full list into `_create_ssl_context` and calling `context.load_verify_locations(cafile=...)` once per entry fixes both; the limitation is carried over from the line this replaces, not created here.

framework/wazuh/core/indexer/tests/test_indexer.py
```diff
@@ -46,9 +46,9 @@ async def test_get_indexer_client_resolves_relative_certificate_paths():
         "indexer": {
             "hosts": ["https://localhost:9200"],
             "ssl": {
-                "certificate_authorities": [{"ca": ["etc/certs/root-ca.pem"]}],
-                "certificate": ["etc/certs/indexer-connector.pem"],
-                "key": ["etc/certs/indexer-connector-key.pem"],
+                "certificate_authorities": ["etc/certs/root-ca.pem"],
+                "certificate": "etc/certs/indexer-connector.pem",
+                "key": "etc/certs/indexer-connector-key.pem",
```
Antoniogm03: **[S3 · non-blocking · introduced]** The fixture is still a hand-written literal, which is exactly the mechanism that let this regression ship: the old literal matched the old parser, the test stayed green, and the producer changed underneath it. The repo already has the frozen effective document and its loader (`framework/wazuh/core/tests/data/configuration/wazuh-manager.effective.json` and `_load_fixture_effective` in `framework/wazuh/core/tests/test_configuration.py`); driving the fixture from them pins this consumer to what `get_manager_conf()` actually returns, and would have surfaced F1 immediately. Nothing covers the schema defaults (`certificate_authorities: []`, `certificate: ""`, `key: ""`) or any of the three `Missing or empty <name> path` branches.

jotacarma90 (approved):

PR #39191, opened 2026-09-11
Stop counting the task client's shutdown drain as a failed task-manager poll

### LINKED ISSUE #38880: Terminal rejections that carry no status the agent can act on
## Description

Four findings from #38549 that are one contract decision, not four fixes: what the manager answers when it cannot serve a request. In all four the manager knows exactly why it is refusing and the agent is left unable to tell a permanent rejection from a transient one, so it retries the same work on the same cadence.

- **Finding 63**: an oversized `Content-Length` on `/events/stateless` is closed with no HTTP response, so the agent retries the same body forever.
- **Finding 64**: the cold-start window sheds every `/stateful` session with `503` and no `Retry-After`, indistinguishable from overload.
- **Finding 39**: back-pressure `503`s carry no `Retry-After` by design, while the agent already honours one when it is sent.
- **Finding 10**: a failed task-manager poll is answered as "no pending tasks".

Code references are at `5.0.0` `7990fc035c`. Original reports: [10](https://github.com/wazuh/wazuh/issues/38549#issuecomment-5395363924), [39](https://github.com/wazuh/wazuh/issues/38549#issuecomment-5427750134), [63 and 64](https://github.com/wazuh/wazuh/issues/38549#issuecomment-5493532588).

## Environment

5.0.0 nightly packages on the reference pair (c5ad.xlarge x86, dedicated indexer), real agents over HTTPS. Findings 39 and 64 come from overload and cold-start runs of #38299; finding 63 from a crafted oversized POST; finding 10 is code-verified and was reproduced end to end.

## Validation before the fix

The four were produced under conditions that differ in how close they are to a normal installation. Confirm each one on a stock 5.0.0 with shipped internal options, record the conditions it actually needs, and let that set the priority. A finding that only reproduces with a crafted request or a load pattern no real fleet produces should be re-triaged, not fixed on the strength of the lab run.

| Finding | How it was produced | Non-default conditions | Confirm before fixing |
|---|---|---|---|
| 63 | a crafted POST declaring `Content-Length` above the 20 MB cap | a crafted request | whether a real agent can build a body over `http_max_body_size` at all: which batch size and event rate reaches 20 MB, and whether the retry is really unbounded |
| 64 | fresh install, empty CTI content, 1000 agents waiting | none, it is a first boot | the window length on a default install with a normal fleet size, and which path sheds the sessions that are not VD `SyncData` |
| 39 | an overload that shed 26,946 sessions | a deliberate overload | that a realistic fleet reaches the shed path at the default `max_parallel_connections`, and how long agents take to recover on their own backoff |
| 10 | code-verified, driven with a failing poll | a stalled or stopped task manager | the agent-visible result (`200` with an empty task list) with the task manager down, on defaults, and whether an operator sees anything at the default log level |

---

## 1. An oversized `Content-Length` is closed with no `413` (finding 63)

A POST to `/events/stateless` declaring a `Content-Length` above the transport cap is rejected at the header gate, efficiently (zero body bytes read) but by closing the connection with no response at all. The cap is RESTinio's `max_body_size` (`http_server/RestinioHttpServer.cpp:1351`), resolved from `remoted.http_max_body_size` with a 20 MB default (`http_server/httpServerConfig.cpp:34`, `:160`). The module's own README already states the behaviour: "RESTinio rejects an oversized `Content-Length` early by closing the connection" (`remoted_module/README.md:125`).

The breach is not silent only towards the agent. RESTinio's diagnostics are routed to DEBUG1 and throttled (`RestinioHttpServer.cpp:85`), so at the default log level neither side records that a body was refused.

The agent cannot distinguish "too big, do not retry" from a network blip, so an oversized payload is retried unchanged, forever.

**Fix direction:** answer `413` before closing. The auth gateway already produces a `413` for its own cap (`auth/authMiddleware.cpp:90`), so the status and the agent-side handling exist; what is missing is a response on the transport-level gate.

## 2. The cold-start window sheds `/stateful` with no `Retry-After` (finding 64)

On a fresh install with empty CTI content and 1000 agents waiting, first boot to "CVE feed fully loaded" took **6 m 22 s**, and for that whole window every `/stateful` session was shed with `503` and no `Retry-After`.

The manager does have a hint on this path, but only for one session class: `syncEndpoint.cpp:143-160` answers `503` plus `Retry-After` (`vd_feed_retry_after_seconds`) when `scanner->feedReady()` is false, and only for VD `SyncData` sessions. Whatever shed the rest of the sessions in that window carried nothing. Identifying that path is the first step, and it is the same question finding 51 asks in the neighbouring issue.

Bonus observation from the same run, worth one line in the sizing documentation and not part of this issue: with an undersized indexer heap (~860 MB) the initial full feed load loops forever on the indexer's parent circuit breaker (429, `Data too large`) while the manager retries; 2 GB completed on the first attempt.

**Fix direction:** identify the shedding path for non-VD sessions in the cold-start window, then attach a `Retry-After` sized from the feed's remaining load time, as the VD lane already does.

## 3. Back-pressure `503`s carry no `Retry-After` (finding 39)

Across an overload that shed **26,946** sessions, every `503` had `retry_after=0`. This is deliberate and stated twice: `remoted_module/README.md:102` ("no `Retry-After`, the agent runs its own retry/backoff") and `src/remoted/src/secure.c`.

The reporter's supporting fact is what makes it worth revisiting: the agent computes its wait as `std::max(serverDelay, backoffDelay)` (`retrySender.cpp::delayFor()`), so a server hint is honoured and is never clamped down by `backoff_cap`. The manager has a working mechanism to spread a `503` storm and does not use it on the path where storms happen.

**Fix direction:** emit `Retry-After` on back-pressure `503`s, sized from the queue's drain estimate. This is an enhancement over documented behaviour, not a defect, and it is the same decision as finding 64.

## 4. A failed task-manager poll is reported as "no pending tasks" (finding 10)

`controlHandler.cpp:506` calls `getPendingTasks()` and its callback receives a `SocketError`. Since #38628 the error is logged, throttled, at DEBUG1 (`:513-524`, "pending-task fetch failed, answering with no tasks"), and the response to the agent is unchanged: `200` with an empty task list. A task-manager timeout and a genuinely empty queue remain the same event to the agent, and at the default log level to the operator too.

The throttled line also undercounts, for the reason in the neighbouring issue (finding 55): a burst that ends inside one window prints `1`.

**Fix direction:** propagate the failure into the response, or at minimum into a counter. About 8 lines in `controlHandler.cpp`. Whether the agent is told, and how, is the protocol half.

---

## Decisions needed

1. Which terminal conditions must carry a status the agent can act on: `413` before the close (63), `Retry-After` on a shed `503` (39, 64), a distinguishable answer for a downstream failure (10).
2. Whether "the agent runs its own backoff" stays the answer for capacity shedding now that a fleet-wide cold start is a measured 6-minute window.
3. For finding 10, whether the agent-facing response changes at all or the fix stops at observability.

## Related

- #38549 (parent), findings 10, 39, 63, 64
- #38284 (timing parameters and the envelope numbers from the same runs)
- #38299 (readiness validation, where 63 and 64 were found)

jepalfer:
## Description

`wazuh-manager-remoted` answers a `/control` `notify` with `200` and `tasks: []` both when the Task Manager poll found nothing pending and when the poll itself failed. That is deliberate: a failed poll leaves the tasks `pending`, so the next notify delivers them and nothing is lost, only delayed by one poll interval. The task client already logs every real failure cause (timeout, connect, I/O, bad status, malformed body, queue full) at `WARN` or `ERROR`, so the operator sees it at the default log level; `controlHandler.cpp`'s own line is a per-agent `DEBUG1` trace.

What was wrong is narrower: a clean shutdown drain answered every in-flight request with the same `SocketError::Io` a transport failure produces, so every manager restart recorded those requests as poll failures, both on the `controlHandler` trace line and on the `remoted.control.task_fetch_error` counter.

Closes #38880 (finding 10, below). Findings 39 and 64 are resolved by design; the reasoning is recorded in the issue's closing comment. Finding 63 (an oversized `Content-Length` on `/events/stateless` closes the connection with no response) is re-triaged, not fixed: the reachable case is an agent `<batch><size>` raised above the manager's transport cap, and the consequence is now stated on that setting's documentation, below. Patching vendored RESTinio to answer `413` from the transport is out of scope for the payoff.

## Proposed Changes

- New `SocketError::Stopping`, answered by the shutdown drain of both the task client and the wazuh-db client (they share the enum) instead of `Io`, so a caller can tell a clean shutdown apart from a genuine transport failure.
- `controlHandler.cpp` skips `Stopping` on its per-agent `DEBUG1` trace. The level does not change: the task client is the operator-facing alert, this line is the trace.
- `taskClient.cpp` no longer counts a request rejected during shutdown as a fetch error, so `remoted.control.task_fetch_error` stops moving on every restart. Queue-full rejections are still counted.
- The agent-visible response is unchanged: `200` with `tasks: []` on any poll failure.

### Results and Evidence

End-to-end in a devcontainer (real manager and enrolled agent over HTTPS, `wazuh-manager-remoted` rebuilt from the branch), driving signed `/control notify` requests with `remoted_module/tools/send_control.py`:

- Task Manager healthy: `200` + `tasks: []`, nothing extra logged.
- Task Manager socket unreachable, default log level: 3 failed polls collapsed into one throttled `ERROR` from the task client, no line from `controlHandler.cpp`.
- Same failure with debug logging: one `ERROR` (task client) plus one `DEBUG` (control handler trace) per window.
- The agent saw `200` / `tasks: []` throughout, no retry storm.
- The shutdown-drain path was not reproduced live (remoted drains faster than an external client can land a request); it is pinned by the destructor tests of both clients.

Baseline:

```
--> POST https://127.0.0.1:1517/wazuh-manager/control
    body sent (49 bytes): {"type": "notify", "agent": {"version": "5.0.0"}}
<-- 200 OK
    {"agent":{"config_hash":"ab0077555b6db8a066e8dc61ce720aba5bb044172b084b5f3ccd3854cae33d3c","config_token":"default","groups":["default"]},"settings_hash":"d45f2daaa7658af00d9b2abeac1a7fcdeaedd54cee454c45349c0693c518f832","tasks":[],"vd_feed_offset":991021}
```

Task Manager unreachable, default log level:

```
2026/09/17 10:22:13 wazuh-manager-remoted:task-client: ERROR: Failed to reach the task manager socket at /queue/sockets/task-http.sock: 1 failure(s) in the last 90 s.
```

Same, debug logging enabled:

```
2026/09/17 10:20:12 wazuh-manager-remoted:task-client[57112] taskClient.cpp:283 at workerLoop(): ERROR: Failed to reach the task manager socket at /queue/sockets/task-http.sock: 1 failure(s) in the last 90 s.
2026/09/17 10:20:12 wazuh-manager-remoted:control-handler[57112] controlHandler.cpp:506 at operator()(): DEBUG: Agent 1: pending-task fetch failed, answering with no tasks: 1 failure(s) in the last 90 s.
```

### Artifacts Affected

- `lib/libremoted_module.so` (`wazuh-manager-remoted`).

### Configuration Changes

None.

### Documentation Updates

- `docs/ref/modules/client/configuration.md`: new `batch` subsection under `<agent>` documenting `size` and `interval` and the two manager caps that bound `size` (`remoted.auth_max_body_size`, which answers `413`, and `<remote><https><max_body_size>`, above which the connection is closed with no response and the agent retries indefinitely).
- `SocketError` is internal to `remoted_module`; the module README does not document the drain's error value, so nothing to update there.

### Tests Introduced

- `controlHandler_test.cpp`: `NotifyReturnsEmptyTasksOnTaskManagerFailure` forces a real `ProtocolError` (fake Task Manager answers `500`) and asserts the response is still `200` with an empty `tasks` array.
- `taskClient_test.cpp` and `wazuhDBClient_test.cpp`: the destructor tests now assert `SocketError::Stopping` on the drained callbacks.

## Review Checklist

- [x] Code changes reviewed


src/remoted/remoted_module/src/http_server/IHttpServer.hpp
```diff
@@ -89,6 +100,22 @@ namespace remoted::http
...
+        /**
+         * @brief Build the load-shedding 503, carrying SHED_RETRY_AFTER_SECONDS.
+         *
+         * Every 503 remoted produces ITSELF (budget shed, deferred-slot exhaustion, and each
+         * endpoint's collapse of a downstream failure) is this one, so the agent gets the same
+         * actionable delay whichever layer refused it. NOT for a 503 forwarded from a downstream
+         * service: that is the session result and passes through with the downstream's own
+         * Retry-After (statefulEndpoint's forwardRetryAfter).
+         */
+        static HttpResponse serviceUnavailable()
```
jepalfer: **[S2 · blocking · introduced]** F1 — this doc says every `503` remoted produces itself is this one, but two authenticated agent routes still build their own bare `503`: `scanVdEndpoint.cpp:181-182` (`errorJson(503, "scan_queue_full" | "vd_unreachable" | ...)`) and `enrollmentEndpoint.cpp:363` (authd unreachable), plus `httpStatusForAuthdError`'s 9013/9015/9016 mapping. `/scan/vd` is the canonical fleet-synchronised storm — one `/control` notify bumps `vd_feed_offset` for every agent on the node, they all POST, VD's dispatch lane fills, and `METRIC_QUEUE_FULL` says outright that this is capacity — and the agent would honour a hint there (`rescanRequester.cpp:101` goes through `RetrySender`). The cheapest fix is the same `status == 503` gate the PR already uses in `asioUdsHttpServer.cpp` and `syncPipeline.cpp`, applied inside `scanVdEndpoint.cpp`'s local `errorJson()`.

src/wazuh_modules/inventory_sync_server/src/endpoints/syncEndpoint.cpp
```diff
@@ -41,9 +41,15 @@ namespace
...
     /// state. The causes are distinguished in the logs instead.
+    ///
+    /// The `Retry-After` does not weaken that: it is a delay, identical at every shed point, so it
+    /// says "later" without saying which limit tripped. remoted forwards it to the agent verbatim
+    /// (statefulEndpoint's forwardRetryAfter), which waits `max(hint, its own backoff)`.
     wazuh::uds_http::HttpResponse serviceUnavailable()
     {
-        return wazuh::uds_http::HttpResponse::json(503, R"({"error":"Service unavailable","code":503})");
+        auto response = wazuh::uds_http::HttpResponse::json(503, R"({"error":"Service unavailable","code":503})");
+        response.headers.emplace_back("Retry-After", wazuh::uds_http::SHED_RETRY_AFTER_SECONDS);
```
jepalfer: **[S2 · blocking · introduced]** F2 — the benchmark simulator decides *which of two retry contracts to follow* from `hasRetry := resp.RetryAfter != ""` (`tool_simulator/internal/runner/session.go:83`), so making every shed `503` carry the header kills its backpressure branch: `retries_503` is now always 0, `s503_retry_after` becomes identical to `s503` (documented as "a **subset** … the feed-not-ready case" in `docu/09-metrics-and-output.md:34`), and AC-H plus the `"s503_retry_after": {"gte": 1}` scenario gate pass on any ordinary capacity shed without the feed gate ever firing. Worse, shed-counting scenarios that set `retry.enabled=false` now retry anyway through the feed branch until `--feed-timeout`, so the shed counts they exist to measure are wrong. The manager already sends a distinct body for the feed case, and `session.go` already inspects bodies via `isNoop()`, so discriminating on `"vulnerability feed not ready"` instead of on the header is the minimal fix.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 | Issue | Comment |
 |-------|---------|
 | [#38863](https://github.com/wazuh/wazuh/issues/38863) | Updated the bundled MITRE ATT&CK Enterprise dataset from v18.1 to v19.2 (new tactic TA0112 Defense Impairment, tactic TA0005 renamed to Stealth, 23 new and 17 revoked techniques). The database is generated at installation time; existing installations are not migrated. |
+| [#38880](https://github.com/wazuh/wazuh/issues/38880) | Added a `Retry-After` header to every load-shedding `503` the manager answers an agent with (`remoted`'s in-flight byte budget and deferred-slot exhaustion, the `uds_http_server` transport's session and connection caps, and the Inventory Sync Server's admission and pipeline gates), so an agent can tell a shed request apart from a network failure and the fleet's first retry is spread. Previously only a not-ready CVE feed sent one. The agent waits the longer of the hint and its own backoff, so no retry becomes slower than it already was. |
```
jepalfer: **[S2 · blocking · introduced]** F3 — the PR declares `Closes #38880` but finding 10 is untouched: at the PR head, `controlHandler.cpp:524-535` still logs the `SocketError` at DEBUG1 and answers the agent `200` with an empty task list, exactly as the issue describes. Finding 63 is properly re-triaged and the issue's own "Validation before the fix" section authorises that, but finding 10 is only deferred ("a separate change in `controlHandler.cpp`"), and the issue asks for a decision on it (decision 3), not a postponement. Either change this to `Part of #38880` and open the follow-up, or fold in the ~8 lines the issue itself estimates.

docs/ref/modules/remoted/agent-api.yaml
```diff
@@ -220,8 +220,13 @@ paths:
           description: |
             The server is temporarily out of capacity (its in-flight byte budget is exhausted or too
             many requests are already parked awaiting the downstream service), or the downstream engine
-            was unreachable / errored. Server-side load-shedding, not per-client rate-limiting: no
-            `Retry-After` is sent — the agent applies its own retry/backoff. The connection is closed.
+            was unreachable / errored. Server-side load-shedding, not per-client rate-limiting.
+            Carries a `Retry-After` so a shed request is distinguishable from a network failure; the
+            agent waits the longer of that hint and its own backoff. The connection is closed.
+          headers:
```
jepalfer: **[S3 · non-blocking · introduced]** F4 — this YAML is updated but nine statements of the same contract under `docs/ref/` are not, and `.claude/CLAUDE.md` makes `docs/ref/modules/<mod>/` the operator contract a PR must keep in sync. The worst is `remoted/timing-tuning.md:180-181`, an incident-triage table telling the operator that `503` *without* `Retry-After` means a downstream timeout or a byte-budget shed — after this PR that row is unreachable and the table sends them to the content-updater log for what is actually budget exhaustion. Also stale: `remoted/architecture.md:126`, `remoted/configuration.md:710`, `remoted/https-events-api.md:290` and `:295-296`, `inventory-sync-server/README.md:15` and `:100`, `inventory-sync-server/api-reference.md:18` and `:244`, `inventory-sync-server/metrics.md:145`.

src/remoted/remoted_module/src/http_server/IHttpServer.hpp
```diff
@@ -69,6 +69,17 @@ namespace remoted::http
                               ///< address (client.keys' third column) before verifying the token.
     };
 
+    /// `Retry-After` attached to every load-shedding 503 remoted produces itself (the global
+    /// in-flight byte budget, deferred-slot exhaustion, and the endpoints' collapse of a downstream
+    /// failure). A shed request is then distinguishable from a network blip, and the fleet's first
+    /// retry is spread instead of each agent walking its own curve. One uniform value on purpose:
```
jepalfer: **[S3 · non-blocking · introduced]** F5 — "the fleet's first retry is spread" describes the mechanism a single un-jittered value removes, not one it adds. `Backoff::next()` is full jitter uniform in `[0, min(base << attempt, cap)]` with base 1000 ms and cap 60000 ms, and `delayFor()` returns `max(10000, that)`, so with ceilings of 1/2/4/8 s the first **four** retries of every shed agent are exactly 10.000 ms — a genuinely synchronised shed such as `SyncPipeline::stop()`'s drain (which fires on every manager restart, as your own comment notes) now stays synchronised instead of diffusing inside `[0, 1 s]`. The net effect is still favourable because the load is deferred 10×, so either jitter the emitted value if spreading is a real goal, or — the shorter fix — reword this and `IUdsHttpServer.hpp:91-93` to promise only the deferral.

src/shared_modules/uds_http_server/include/uds_http_server/IUdsHttpServer.hpp
```diff
@@ -88,6 +88,15 @@ namespace wazuh::uds_http
...
     };
 
+    /// `Retry-After` attached to every load-shedding 503, so a shed request is distinguishable from
+    /// a network blip and the fleet's first retry is spread instead of each agent walking its own
+    /// curve. One uniform value on purpose: it is a DELAY, not a cause, so it tells the peer nothing
+    /// about which limit tripped (the causes stay in the logs). Matches the value the VD lane
+    /// already ships (`inventory_sync_server_vd_feed_retry_after_seconds`, 10 s) and binds against
+    /// the agent's early full-jitter backoff, whose ceiling starts at 1 s and doubles -- the agent
+    /// waits `max(Retry-After, its own backoff)`, so this only ever raises a near-immediate retry.
+    constexpr const char* SHED_RETRY_AFTER_SECONDS {"10"};
```
jepalfer: **[S3 · non-blocking · introduced]** F6 — this constant and `remoted::http::SHED_RETRY_AFTER_SECONDS` (`IHttpServer.hpp:81`) are two independent copies whose equality both comments declare load-bearing ("so a forwarded hint and a locally-produced one agree"), with no `static_assert` and no test comparing them — the sync tests check only this one, the remoted tests only the other, so changing either leaves both suites green. `remoted_module` already links `wazuh_uds_http_server` (`CMakeLists.txt:40`), so a single definition costs nothing new. If you would rather not couple the HTTPS transport to the UDS library, a `static_assert` in a TU that already sees both headers is enough.

src/wazuh_modules/inventory_sync_server/src/sync/syncPipeline.cpp
```diff
@@ -247,7 +247,15 @@ namespace invsync::sync
                     (classify(item.session) == SessionKind::BulkData) ? m_durationBulk : m_durationImmediate;
                 histogram->observe(static_cast<uint64_t>(elapsed));
             }
-            item.responder->send(wazuh::uds_http::HttpResponse::json(status, body));
+            auto response = wazuh::uds_http::HttpResponse::json(status, body);
+            if (status == 503)
+            {
+                // Post-admission 503s (the stop() drain, and a flush against an unavailable
+                // indexer) are as indistinguishable from a dead link as the admission sheds are,
+                // and the drain fires on every manager restart. Same hint, same reason.
```
jepalfer: **[S3 · non-blocking · introduced]** F7 — deleting this `if (status == 503)` block, its twin in `vdScanLane.cpp:242-250`, or the `.append_header(retry_after, …)` in `RestinioHttpServer.cpp:1081-1082` leaves every suite green. All three already have tests that reach those `503`s and could assert the header in a line each: `syncPipeline_test.cpp:333/344/422`, `vdScanLane_test.cpp:493/522`, and `httpServer_test.cpp:1076`, which already inspects the raw response head. `statefulEndpointE2E_test.cpp` — the only end-to-end proof that a `Retry-After` reaches the wire — still covers just the configured feed value.

src/wazuh_modules/inventory_sync_server/src/endpoints/syncEndpoint.cpp
```diff
@@ -183,7 +189,15 @@ namespace invsync::endpoints::sync
                         }
                         // The lane itself counts vd.capacity.503.total at the refusal.
                         deps.requestCounters.count(503);
-                        responder->send(errorResponse(503, "scan capacity exhausted"));
+                        {
+                            // Keeps its own body (the lane queue is a distinct, actionable cause in
+                            // the logs) but sheds like every other capacity refusal, so it carries
```
jepalfer: **[S3 · non-blocking · introduced]** F8 — the sweep is uneven inside this module while `README.md` RNF-3 was edited to read "every refusal is an HTTP answer, and every one carries a `Retry-After`": the identical pipeline-full gate at `deleteAgentEndpoint.cpp:248-259` is still bare, as are `configEndpoint.cpp:62`, `statsEndpoint.cpp:58`, `metricsEndpoint.cpp:30` and `vdScanEndpoint.cpp:130/179/185`. None of those are on the agent path, so this is consistency rather than a functional gap — but the README sentence should be scoped back to the four `/stateful` admission gates. Separately, `vd.retry_after.total`'s registered description (`metricNames.hpp:95`, pinned verbatim by `vdScanLane_test.cpp:324`) reads "503 responses carrying a Retry-After header", which now describes every `503` while the counter still only fires at the two feed gates.

src/wazuh_modules/inventory_sync_server/src/vd/vdScanLane.cpp
```diff
@@ -238,7 +238,16 @@ namespace invsync::vd
         {
             m_requestCounters.count(status);
             observeLaneTime(item);
-            item.responder->send(wazuh::uds_http::HttpResponse::json(status, body));
+            auto response = wazuh::uds_http::HttpResponse::json(status, body);
+            if (status == 503)
+            {
+                // Every 503 a lane worker answers -- shutting down, indexer unavailable, connector
```
jepalfer: **[S3 · non-blocking · introduced]** F9 — this blanket gate also hints `SCAN_NO_SCANNER_BODY`, which `workerLoop` sends for `AgentScanOutcome::Skipped` with the WARN "this node runs no vulnerability scanner" (`:388-396`) — a standing misconfiguration, not capacity, so "come back in 10 s" is answering a condition that will still hold in 10 days and pins a 10 s floor under the caller's own backoff. That is the rule this PR states in `statelessEndpoint_test.cpp`'s `NonShedStatusesCarryNoRetryAfter` ("on a permanent rejection it would invite exactly the retry the status is refusing"), applied to a `503` instead of another status. Stateless load balancing means a retry may land on a scanner-bearing node, so this is mild — but it is worth either excluding that body or sending it through the direct path the feed re-check already uses.

src/remoted/remoted_module/src/http_server/RestinioHttpServer.cpp
```diff
@@ -1075,6 +1075,11 @@ namespace remoted::http
                                     }
                                     return request->create_response(restinio::status_service_unavailable())
                                         .append_header(restinio::http_field::content_type, "application/json")
+                                        // Load-shedding, not a per-client limit: the delay is what
+                                        // lets the agent tell this apart from a network blip and
+                                        // spreads the fleet's first retry.
```
jepalfer: **[S4 · non-blocking · introduced]** F10 — the comment five lines above this still reads "shed load with a plain 503 (the agent runs its own retry/backoff)", which is now the opposite of what the call does. Two more survive: `remotedModuleFacade.hpp:355` ("the forwarder sheds load with a plain 503 (the agent runs its own retry)") and `vdScanLane_test.cpp:283`, whose "the one send that bypasses respond() (it carries the Retry-After header)" no longer distinguishes anything now that `respond()` carries one too — the real reason is the configured value, which `vdScanLane.cpp:353-355` was updated to say. The log line at `syncEndpoint.cpp:155` ("Answered 503 + Retry-After to … vulnerability-detection session(s)") likewise stops isolating the feed gate.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 | Issue | Comment |
 |-------|---------|
 | [#38863](https://github.com/wazuh/wazuh/issues/38863) | Updated the bundled MITRE ATT&CK Enterprise dataset from v18.1 to v19.2 (new tactic TA0112 Defense Impairment, tactic TA0005 renamed to Stealth, 23 new and 17 revoked techniques). The database is generated at installation time; existing installations are not migrated. |
+| [#38880](https://github.com/wazuh/wazuh/issues/38880) | Added a `Retry-After` header to every load-shedding `503` the manager answers an agent with (`remoted`'s in-flight byte budget and deferred-slot exhaustion, the `/scan/vd` dispatcher queue, an unreachable `authd`, the `uds_http_server` transport's session and connection caps, and the Inventory Sync Server's admission and pipeline gates), so an agent can tell a shed request apart from a network failure and defers its retry instead of retrying on the cadence of a broken link. Previously only a not-ready CVE feed sent one. The agent waits the longer of the hint and its own backoff, so a retry can come later than before but never sooner. |
```
jepalfer: **[S2 · blocking · introduced]** F3 — the PR body says "Part of #38880 ... finding 10 stays open and needs its own change and its own decision", but what closes an issue is the Development-panel link, not the body text, and that link is still `closes`. `controlHandler.cpp` is unchanged at this head, so merging closes #38880 and finding 10 — plus the issue's own decision 3, which it leaves open — disappears with nothing recorded. Unlink #38880 from the Development panel, or repoint it at a follow-up issue that carries finding 10.

src/wazuh_modules/inventory_sync_server/src/endpoints/syncEndpoint.cpp
jepalfer: **[S2 · blocking · aggravated]** F18 — this is F2 again, in the C++ harness. `inventory_sync_server_testtool` still decides the retry contract with `if (result.status != 503 || result.retryAfter.empty() || now >= deadline) break;` (`testtool/main.cpp:643`), so now that every `/stateful` shed carries the header, an indexer outage or a full admission queue takes the feed branch: it prints `Feed not ready (503, Retry-After 10)` and retries until `--feed-timeout` (300 s default) expires, per input file. Apply the same body-based discrimination you used at `session.go:98`, and update `test-tools.md:124` and `main.cpp:21-22`, which still state the replaced contract.

src/remoted/remoted_module/src/endpoints/scanVdEndpoint.cpp
```diff
@@ -57,7 +57,16 @@ namespace remoted::endpoints::scanvd
...
-            return remoted::http::HttpResponse::json(status, std::move(body));
+            auto response = remoted::http::HttpResponse::json(status, std::move(body));
+            if (status == 503)
+            {
+                // Capacity, like every other shed: `scan_queue_full` is the dispatcher refusing
+                // work. This route is the one that is synchronised fleet-wide by construction --
+                // a feed-offset bump reaches every agent through /control, so they all re-scan at
+                // once -- so leaving it unqualified is exactly the storm the hint exists for.
+                response.headers.emplace_back("Retry-After", remoted::http::SHED_RETRY_AFTER_SECONDS);
+            }
```
jepalfer: **[S3 · non-blocking · introduced]** F19 — the gate is unconditional, but `vulnerabilityScanner.cpp:220` answers `503 {"error":"vd_not_initialized","retryable":false}`, and remoted drops that flag when it rebuilds the body, so the header becomes the only retry signal and it says the opposite. Five of the scanner's six 503 codes are `retryable:true`; only this one is not — the same permanent/transient split you make one module away in `VdScanLane::respond`. (Secondary: `feed_not_ready` gets the fixed 10 s here while `/stateful` relays the configured feed value, so the two feed gates now disagree.)

```suggestion
            if (status == 503 && code != "vd_not_initialized")
            {
                // Capacity, like every other shed: `scan_queue_full` is the dispatcher refusing
                // work. This route is the one that is synchronised fleet-wide by construction --
                // a feed-offset bump reaches every agent through /control, so they all re-scan at
                // once -- so leaving it unqualified is exactly the storm the hint exists for.
                // NOT for `vd_not_initialized`: the scanner marks that one `retryable:false` and
                // remoted drops the flag, so a hint would be the only retry signal, and a wrong one.
                response.headers.emplace_back("Retry-After", remoted::http::SHED_RETRY_AFTER_SECONDS);
            }
```

docs/ref/modules/remoted/timing-tuning.md
```diff
@@ -177,8 +177,8 @@ Answer these once per deployment; they decide which of the knobs above you will
...
 | Integrity check returns `500` forever | the session query page is larger than the indexer accepts | the indexer's `index.max_result_window` against the configured batch size (values above 10000 are already refused at startup, so this means the indexer's window was lowered) |
-| `503` with `Retry-After` on `/stateful` | VD feed not ready (CVE content not loaded) | content-updater log |
-| `503` without `Retry-After` | downstream timeout **or** an in-flight budget shed — indistinguishable to the agent | `remoted.server.budget.*` vs `remoted.forwarder.*`; budget sheds are not in endpoint metrics |
+| `503` on `/stateful` with a `Retry-After` **larger than the fixed shed value** | VD feed not ready (CVE content not loaded) — the value is `inventory_sync_server_vd_feed_retry_after_seconds`, relayed from the sync server | content-updater log |
+| `503` with the fixed shed `Retry-After` | downstream timeout **or** an in-flight budget shed — indistinguishable to the agent, and no longer distinguishable by the header either | `remoted.server.budget.*` vs `remoted.forwarder.*`; budget sheds are not in endpoint metrics |
```
jepalfer: **[S3 · non-blocking · introduced]** F20 — this rule cannot fire on a stock install: `getDefine_Int_default("wazuh_modules", "inventory_sync_server_vd_feed_retry_after_seconds", 10, 1800, 10)` (`wm_inventory_sync_server.c:229`) makes both the default *and the minimum* 10, identical to `SHED_RETRY_AFTER_SECONDS`. So row 1 never matches and row 2 swallows the feed case, sending the operator to `remoted.server.budget.*` during a CVE-feed cold start — the exact misdiagnosis this sweep set out to remove. The same premise is in `inventory-sync-server/README.md:100` ("the configured, **longer** value").

```suggestion
| `503` on `/stateful` whose body is `vulnerability feed not ready` | VD feed not ready (CVE content not loaded) — its `Retry-After` is `inventory_sync_server_vd_feed_retry_after_seconds`, relayed from the sync server, and only differs from the fixed shed value if you raised it (default and minimum are both 10) | content-updater log |
| `503` with a generic body | downstream timeout **or** an in-flight budget shed — indistinguishable to the agent, and no longer distinguishable by the header either | `remoted.server.budget.*` vs `remoted.forwarder.*`; budget sheds are not in endpoint metrics |
```

src/wazuh_modules/inventory_sync_server/README.md
```diff
@@ -62,7 +62,7 @@ below. Status: **kept** = the module provides it; **superseded by D-n** = delibe
 |---|---|---|
 | RNF-1 | Preserve the 9 security controls: anti-spoofing, index allowlists, authoritative `wazuh.*` overlay, strict JSON, `external_gte` guard, idempotency, admission quota, cluster isolation | [Validation](#validation-syncfullsessionvalidator), [the allowlist](#the-allowlist-syncstateindexallowlisthpp), the overlay in `sessionProcessor` |
 | RNF-2 | No head-of-line blocking: no sleeps or unbounded waits on the completion path | sharding + the VD lane; a slow agent/scan delays only its shard/lane |
-| RNF-3 | Explicit backpressure at admission and ingestion (no silent drops) | the four `503` gates; every refusal is an HTTP answer |
+| RNF-3 | Explicit backpressure at admission and ingestion (no silent drops) | the four `503` gates; every refusal is an HTTP answer, and each of these carries a `Retry-After` (the module's other endpoints — config, stats, metrics, delete-agent, vd-scan — are not agent-facing and are unchanged) |
```
jepalfer: **[S3 · non-blocking · introduced]** F22 — `/config` and `/stats` *are* agent-facing: `remotedModuleFacade.hpp:430` and `:436` wire remoted's agent routes to `inventorySyncSocketPath`, and both endpoints set `serviceName = "inventory sync server"`. The behaviour is still correct, but only because remoted's `postProcess` replaces the downstream 503 with its own `serviceUnavailable()` — so state that as the reason rather than inventing a boundary that is not there. Same file, `:873`: the FAQ still explains the two VD gates by "the AGENT reacts differently: `Retry-After` schedules the re-POST", which `syncEndpoint.cpp:198` has just falsified — the difference is now the *size* of the delay, not the presence of the header.

src/wazuh_modules/inventory_sync_server/src/vd/vdScanLane.cpp
```diff
@@ -238,7 +238,21 @@ namespace invsync::vd
...
+            // "Retry later" only where later could plausibly differ: shutting down, indexer
+            // unavailable, connector failure, scanner still starting. NOT for SCAN_NO_SCANNER_BODY,
+            // which IVdScanner documents as permanent (no scanner on this node): a hint there would
+            // invite the retry the status is refusing. Known imprecision, pre-existing and outside
+            // this change: the adapter also reports Skipped while a configured scanner has not
+            // finished initialising, so in that window this 503 goes out bare and the caller falls
+            // back to its own backoff -- the pre-hint behaviour, not a regression.
+            // The feed re-check below does not come through here either: it needs the configured,
+            // feed-sized value instead of the generic hint.
+            if (status == 503 && body != SCAN_NO_SCANNER_BODY)
```
jepalfer: **[S3 · non-blocking · aggravated]** F17 — the permanent/transient decision string-compares the rendered JSON body against `SCAN_NO_SCANNER_BODY`, a wire payload used as a control-flow discriminator; `SCAN_NOT_READY_BODY` is its literal sibling, so rewording either one silently inverts the decision with no compile error. `AgentScanOutcome` is available one frame up and is the natural discriminant — pass the outcome (or a `bool transient`) into `respond()`. The premise also still rests on `vdScannerAdapter.cpp:62` returning `Skipped` for `!isInitialized()`, which `IVdScanner.hpp:51-52` says should be `NotReady`; the comment documents that honestly, which is fair, but mapping it correctly is a one-line change.

jotacarma90 (approved):

Issue #39063, opened 2026-09-09
Make the enrollment token the only registration path in every installer


jpcerrone:
**Component:** `register_configure_agent.sh`, `InstallerScripts.vbs`, `WriteAgent()`, MSI
**Depends on:** the MVP branch `feat/39021-poc-agent-https-tier2`, #39040 (for the shared C codec)
**Rough size:** L

## Description

Objective requirements 10, 12 and 14 for every packaged deployment surface, minus the parsing the MVP already did.

Already on `feat/39021-poc-agent-https-tier2`, so **not** in scope here: `wet_parse_enrollment_token()` in `register_configure_agent.sh` decodes the unpadded base64url, parses the flat JSON with a minimal tokenizer matching the token's schema, and rejects with named, distinct error codes for bad base64, bad JSON, a missing or non-integer `ver`, a missing `adr`, both `pin` and `ca`, and neither. A malformed token logs its code and does not abort the install. `src/init/tests/test_enrollment_token.sh` covers it.

That has since moved on. `set_agent_enrollment_token()` now decodes the token with **the agent's own C codec** (`wazuh-agentd --show-token`, commit `f70ee76e5c`) rather than the shell tokenizer, distinguishes a refused token from a decoder that never ran, and stores the token at `etc/enrollment_token` as **`0600 root:root`** — created empty, chmod'd and chown'd *before* the secret is written. `w_agent_token_bootstrap()` reads it once at startup and deletes it on success. So the decoder question is settled, the token reaches the agent, and the credential is not left world-readable at any point.

What remains is the conflict handling, the other platforms, and the variable cleanup.

**Scope changed on 2026-09-11: the token is the only way to register.** Every other enrollment path is dropped, so this stops being "add the token alongside the existing variables" and becomes "the token replaces them". The variable table collapses from twenty-five entries to one plus the non-enrollment leftovers.

**Windows is in the same task on purpose.** The conflict and refusal rules of §2.8 are subtle — per *value*, not per mechanism — and the source comments already require `register_configure_agent.sh`, `InstallerScripts.vbs` and `WriteAgent()` to stay in lockstep. Implementing that table twice, in two pull requests, by two people, is how it drifts.

### Collapse the variable set

The current list is twenty-five entries (`register_configure_agent.sh`'s own `vars=(...)` array). Decide each one against the new rule, and in the same change so the three implementations never disagree:

- [x] **Keep: `WAZUH_ENROLLMENT_TOKEN`.** The only registration input.
[...]

PR #39197, opened 2026-09-11
Add the manager migration procedure and tool for 4.x to 5.0, and align the migration guides with HTTPS tier 2

### LINKED ISSUE #39109: Validate and complete the 4.x to 5.0.0 manager migration guide against a real 4.x environment and evaluate a migration tool
|---|---|---|---|---|
| 5.0.0 | Manager (docs/guide/migration, wazuh-db, authd, remoted) | Manager | Packages | Linux |

## Description

Upgrading a manager in place from 4.x to 5.0.0 is not supported ([ref/upgrade.md](https://github.com/wazuh/wazuh/blob/5.0.0/docs/ref/upgrade.md)). A user with a working 4.x deployment (manager, enrolled 4.x agents, populated inventories, alert history, groups, API users) has to install a fresh 5.0.0 manager and carry their data across by hand, following [docs/guide/migration](https://github.com/wazuh/wazuh/blob/5.0.0/docs/guide/migration/README.md).

Today that section is a collection of per-feature guides written from the feature's point of view. There is no end-to-end procedure written from the user's point of view: what to take from the 4.x manager, in which order to restore it on 5.0.0, what a 4.x agent keeps working against the new manager while it waits for its own upgrade, and how to verify the result. Several guides also contradict each other or leave the user without an answer on the data that matters most.

This issue is about closing that gap, validating the procedure against a real 4.x environment, and deciding whether the documented steps should end up as a tool.

### Scope

In scope: the basic identity and management data every deployment has and must survive the move.

- `etc/client.keys` (agent identities and keys).
- `global.db` (agent registry: id, name, ip, version, OS, `group`, `belongs`).
- Agent groups: `etc/shared/<group>/` and the group assignments.
- Enrollment password (`authd.pass`) and the certificates a 5.x agent needs to trust the manager.
- API users, roles and policies (`api/configuration/security/rbac.db`).
- Manager configuration (`ossec.conf` to `wazuh-manager.conf`, internal options, `api.yaml`), which already has a guide.
- The behaviour of a 4.x agent against a 5.0.0 manager during the transition (legacy channel).

Out of scope: rules, decoders, CDB lists and anything the engine changed structurally. Those guides exist ([rules-4x-to-5x.md](https://github.com/wazuh/wazuh/blob/5.0.0/docs/guide/migration/rules-4x-to-5x.md), [xml-decoders-migration.md](https://github.com/wazuh/wazuh/blob/5.0.0/docs/guide/migration/xml-decoders-migration.md), [cdb-to-kvdb-migration.md](https://github.com/wazuh/wazuh/blob/5.0.0/docs/guide/migration/cdb-to-kvdb-migration.md)) and the translation is the user's responsibility. The runbook only has to point at them at the right step.

### What the current guides get wrong or leave open

Findings from reading the 5.0.0 branch and the code, to be confirmed in the test environment.

1. **No entry point.** [migration/README.md](https://github.com/wazuh/wazuh/blob/5.0.0/docs/guide/migration/README.md) is a table of independent guides. It does not list `upgrade-4x-to-5x.md`, `rules-4x-to-5x.md`, `active-response.md`, `integratord-notifications.md`, `syslog-input-4x-to-5x.md`, `remote-syslog-output.md` or `filebeat-to-indexer-connector.md`, and nothing orders the guides into a procedure. `upgrade-4x-to-5x.md` is agent-only despite its name in `SUMMARY.md`.

2. **`client.keys` and groups contradict each other.** [agent-manager-protocol.md](https://github.com/wazuh/wazuh/blob/5.0.0/docs/guide/migration/agent-manager-protocol.md) states that `client.keys` migrates as-is and no re-enrollment is needed. [agent-groups-migration.md](https://github.com/wazuh/wazuh/blob/5.0.0/docs/guide/migration/agent-groups-migration.md) tells the user to delete `client.keys` on every agent and re-enroll, which assigns new agent ids and breaks every historical reference to `agent.id`. The groups guide exists because the `remoted.guess_agent_group` fallback is gone, but the assignments were never only in the agent's `ossec.conf`: they are in `global.db` (`agent.group`, `belongs`). A procedure that carries `client.keys` and `global.db` keeps ids, keys and groups without touching any agent. The guide should present that as the primary path and re-enrollment as the fallback.

3. **`global.db` has no migration path at all.** No guide mentions the 4.x database (`/var/ossec/queue/db/global.db`). It cannot be copied over: 4.x tracks the schema with `metadata.db_version = 7`, 5.0.0 with `PRAGMA user_version` and rejects anything but version 1 (`Unsupported schema version 0 ... Disabling database`, [wdb_upgrade.c](https://github.com/wazuh/wazuh/blob/5.0.0/src/wazuh_db/src/wdb_upgrade.c)). The `agent` table also dropped `os_codename`, `os_build`, `os_uname`, `config_sum`, `merged_sum`, `manager_host`, `node_name` and `group_config_status`, added `os_type`, and the `labels` and `info` tables are gone ([schema_global.sql](https://github.com/wazuh/wazuh/blob/5.0.0/src/wazuh_db/schemas/schema_global.sql)). The user needs either a documented SQL transform or a tool. Restoring `client.keys` alone does not repopulate the database: authd mirrors a key into wazuh-db when it enrolls it, not when it loads the file at startup (to verify).

4. **Backup guide points at the wrong database.** #38917 already reports that `backup-restore.md` names `var/db/global.db` instead of `queue/db/global.db` and omits `rbac.db`, `tasks.db`, `queue/authd/` and `data/`. The migration runbook must not inherit those paths.

5. **4.x agents against the 5.0.0 manager.** The legacy channel is opt-in (`<remote><legacy>`, `<auth><legacy_enrollment>`), and with it a 4.x agent keeps events, keepalives, centralized configuration and remote upgrade, but loses active response, inventory sync and dbsync. Keepalives still update `global.db` (version, OS, status), so the agent list stays meaningful. This is stated in the protocol guide but never in the form the user needs: "after the migration your 4.x agents will show up as X and do Y until you upgrade them", plus the order of operations so no agent is left unable to connect (legacy channel and enrollment password in place before the first agent is started).

6. **Trust and credentials.** A 4.x agent never verified the manager certificate; a 5.x agent does by default. The CA the agents will need, `authd.pass`, and the cluster key are user data that must be placed on the new manager before agents are upgraded. #39068 covers the WPK side; the manager runbook has to say what to preserve and where it lands.

7. **Historical data.** `wazuh-alerts-4.x-*` and the 4.x `wazuh-states-*` indices are not read by 5.0.0 (findings and new states indices). The runbook must say plainly what is kept for consultation, what is rebuilt when agents are upgraded (inventory), and what is lost.

## Tasks

Phase 1, analysis and environment

- [ ] Read every guide under `docs/guide/migration` plus `ref/upgrade.md` and `ref/backup-restore.md` and record, per data item in scope, what the guide says, what the code does and where they diverge.
- [ ] Build a realistic 4.14.x environment: manager (single node, and a master + worker pair), indexer, dashboard, several 4.x agents (Linux, Windows) enrolled in custom groups, with inventories, SCA results, vulnerability findings and alert history populated, custom `agent.conf` per group, API users and roles.
- [ ] Execute the migration to 5.0.0 as the guides describe today and record every point where the user is stuck or loses data (agent ids, groups, keys, API users, inventories, alerts).
- [ ] Verify item 3: what happens with a 4.x `global.db` in place, and whether restoring `client.keys` alone repopulates the registry.
- [ ] Verify item 5: behaviour and dashboard representation of 4.x agents connected through the legacy channel, then their in-place upgrade to 5.0.0 and the resulting reconnection over 1517.

Phase 2, documentation

- [ ] Write a top-level manager migration runbook (new guide, linked first from `migration/README.md` and `SUMMARY.md`) ordered as the user executes it: pre-checks and backup on 4.x, install 5.0.0, restore identity data, adapt configuration, enable legacy channel and enrollment, reconnect 4.x agents, upgrade agents, retire legacy, verify. Each step points at the detailed guide.
- [ ] Add the `global.db` procedure (documented SQL transform or tool, per Phase 3) and the `client.keys` procedure, and rewrite the agent groups guide around them, keeping re-enrollment as fallback.
- [ ] Add the API users (`rbac.db`), `authd.pass`, certificates and cluster key sections.
- [ ] Add a "what your 4.x agents can and cannot do against 5.0.0" table and a "what happens to historical data" section.
- [ ] Fix `migration/README.md` (missing guides, order), and align `backup-restore.md` paths with #38917 where the runbook depends on them.
- [ ] Re-run the migration following only the new documentation, on a clean environment, and attach the evidence.

Phase 3, automation (see next section)

- [ ] Decide, with the results of Phase 1, which steps are worth automating and write the decision down in this issue.
- [ ] If approved, open the implementation issue for the tool.

## Automation: a migration tool

Everything in scope is mechanical: copy files, transform one SQLite database, place credentials, adjust ownership to `wazuh-manager`, and flag configuration options that no longer exist. That makes a tool tempting. It also makes it easy to build something that rots with the next schema change and that users trust more than they should. The options, from least to most ambitious:

**Option A, documentation with verified snippets only.** Every step ships as a copy-pasteable block that Phase 2 has executed on the test environment. No code to maintain. The user still runs 15 to 20 commands across two hosts and the `global.db` transform is a hand-run SQL script embedded in the docs.

**Option B, an export/import script for identity data.** A single script, `wazuh-manager-migrate` or similar, with two commands. `export` runs on the 4.x host and produces one archive with `client.keys`, `global.db`, `etc/shared/`, `authd.pass`, `rbac.db`, the 4.x configuration files and the certificates. `import` runs on the fresh 5.0.0 host with the manager stopped, restores the files to their new paths, transforms `global.db` into the 5.0.0 schema (`agent` column mapping, `group` and `belongs` copied, `user_version` set), fixes ownership and permissions, and prints a report. It does not touch `wazuh-manager.conf`: configuration translation is left to the user, and `bin/wazuh-manager-conf validate` already tells them what is invalid. A `--dry-run` shows the plan and a `--check` mode compares the 4.x archive against the 5.0.0 host after import (agent count, group membership, key count).

**Option C, B plus configuration translation.** The script also rewrites `ossec.conf` into `wazuh-manager.conf` following the tables in `manager-configuration-migration.md`. Every removed section, moved option and renamed internal option is already tabulated, so the mapping exists. The risk is that the mapping is exactly what changes most between betas and that a half-translated configuration is worse than a rejected one.

**Where it would live.** Three candidates: `tools/` in this repo next to `purge_wazuh.sh` (cheapest, no packaging, downloaded by the user), shipped inside the 5.0.0 manager package as `bin/wazuh-manager-migrate` (discoverable, but the export half must run on a 4.x host that has no 5.0.0 package), or the installation assistant (fits the fresh-install flow, but a different repo and team).

**Recommendation.** Do Phase 1 and Phase 2 first with Option A, because the documentation is needed regardless and the environment work tells us where users actually get stuck. Then implement Option B as a standalone Python script in `tools/` (framework Python is available on both hosts; sqlite3 module in the standard library; no dependency on the manager being installed on the 4.x side). Explicitly reject Option C for 5.0.0: keep configuration translation as a checker that lists the offending options with the guide's own wording, never as a rewriter. Review packaging into `bin/` once the script has survived a release cycle.

**Risks to weigh before approving Option B.**

- Schema drift: any change to `schema_global.sql` in 5.x or to `rbac.db` in the framework must update the tool. Mitigation: the tool refuses versions it does not know and has a unit test per supported pair.
- Cluster: export from the master only; workers are reinstalled and receive `client.keys` and groups through the cluster sync. The tool must refuse to run on a worker.
- Partial failure: `import` must be idempotent or must stop before writing anything if a precondition fails (manager running, non-empty `client.keys` on the target, unknown `db_version`).
- Ownership and layout: the 5.0.0 tree runs as `wazuh-manager`, certificates live under `etc/certs` with per-daemon names, and #38917 shows the layout is still being learnt by our own docs.
- Scope creep: the moment the tool touches configuration or content, it becomes a second migration guide that has to be kept in sync with the first.

## Definition of Done

- [ ] A user can migrate a 4.14.x manager to 5.0.0 following the top-level runbook alone, keeping agent ids, keys, groups and API users, with 4.x agents reconnecting through the legacy channel and later upgrading to 5.0.0 without re-enrollment. Verified on the environment of Phase 1 and evidenced in this issue.
- [ ] `docs/guide/migration/README.md` and `docs/SUMMARY.md` expose the runbook first and list every guide.
- [ ] The `global.db`, `client.keys`, groups, `authd.pass`, certificates and `rbac.db` procedures are documented and tested.
- [ ] The automation decision (A, B or C, and where it lives) is recorded here with the reasoning, and the implementation issue exists if approved.

## Related

- #38917 backup guide points at a non-existent `global.db` and omits state paths
- #36590 original agent groups migration guide
- #36567 original agent upgrade migration guide
- #39068 agent deployment variables and the 4.x to 5.0.0 WPK runbook
- #38735 deriving the config selector server side instead of the agent's group CSV

jotacarma90:
## Description

Part of #39109.

The 4.x to 5.0 migration section is a set of per-feature guides with no entry point, and on the data every deployment has to keep (agent keys, agent registry, groups, enrollment password, API users) it either says nothing or tells the user to re-enroll every agent. Executed against a real 4.14.7 environment (Linux and Windows agents, custom groups, manager-side group assignments, API users, populated indexer) and a 5.0.0 all-in-one, the documented path reshuffles agent ids, loses every group assignment made from the manager, and drops the API users. Carrying `client.keys`, the `global.db` rows, the group folders, `authd.pass` and `rbac.db` keeps all of it and the 4.x agents reconnect without enrollment.

Rebased onto 5.0.0 after the HTTPS tier 2 merges (#39079, #39033, #39282, #39285) and adapted to them. Adds the tool that performs the identity half of the procedure, and corrects two reference pages the same merges left describing the previous behaviour.

## Proposed Changes

- New guide `docs/guide/migration/manager-4x-to-5x.md`: the end-to-end manager procedure in execution order. Backup on 4.x with the real paths, certificate provisioning before the first start, restore of identity data on a stopped manager with the ownership each daemon needs, the `global.db` row copy (the 4.x file is rejected by `PRAGMA user_version`), the `rbac.db` upgrade, configuration migration, verification, what a 4.x agent gets over the legacy channel, agent upgrade prerequisites per upgrade path, enrollment tokens for agents installed afterwards, legacy retirement, and what happens to historical data.
- `migration/README.md` lists every guide (seven were missing) with the manager procedure first; `SUMMARY.md` and `ref/upgrade.md` link to it. `rules-4x-to-5x.md` was in the repository but in no index, so it did not render.
- `agent-groups-migration.md`: positioned as the re-enrollment fallback; the tarball no longer carries the 4.x `default` group over the 5.0 one; extraction happens with the manager stopped (a running modulesd sends `delete-group` for a folder it cannot read before the `chown`).
- `upgrade-4x-to-5x.md`: `<agent><manager><endpoint>` instead of `<address>`/`<port>`, rows for `client_buffer` and the removed `labels`, a "Package upgrade on the host" section covering the `dpkg` conffile prompt that blocks unattended upgrades, and a rewritten "Certificate trust check" (see below).
- `tools/migration/wazuh-migrate-identity.py`, with a README and 15 tests: `export` on the 4.x host, `import` on the 5.0 host, `check` against the bundle. It is the identity half of the guide, the half that is mechanical and gets the ownership wrong by hand.
- Two reference pages the tier 2 merges left stale: `ref/modules/client/configuration.md` still documented `system` as the agent's default `verification_mode`, and `ref/modules/authd/configuration.md` still told operators to deploy the enrollment password with `WAZUH_REGISTRATION_PASSWORD`, which #39063 removed.

### The tool

**It recreates the 5.0 side through the manager's API.** The first version wrote the target's files: it transformed the 4.x registry rows into the 5.0 schema, placed `client.keys` with the ownership each daemon expects, and extracted the group folders with a `chown` walk. That is product knowledge living outside the product, and it ages with every change to the schema, the modes or the group layout. It now uses `POST /agents/insert` with each agent's original id and key, `POST /groups` plus `PUT /groups/{group}/configuration`, and `PUT /agents/{id}/group/{group}`, so the manager validates every item and writes `client.keys` and the registry itself. The tool holds no schema, no table of file modes and no notion of what `client.keys` looks like.

Two consequences are worth stating. The manager has to be running, which the install-out-of-the-fleet's-reach step in the guide already provides: the API answers on loopback while no agent can reach the manager. And the registration date is not carried, because no endpoint accepts one; the recorded version and operating system are not carried either, but the manager refreshes those from the first keepalive.

Files in a group folder other than `agent.conf` are reported rather than moved. No endpoint uploads them, and 5.0 compiles everything in a group folder into what it pushes to that group's agents, so a 4.x CDB list or rootcheck file would reach the fleet with no meaning attached.

`export` reads the identity out of the stopped 4.x manager's files into a bundle. `import` recreates it through the API and copies the two credentials no endpoint can express, `authd.pass` and `rbac.db`, both opt-in at each end because the second carries every API password hash. `check` compares what the manager holds against the bundle, and flags an agent the bundle never carried, which is the signature an agent leaves when it re-enrolls against an empty registry.

Everything this tool can refuse is evaluated before the first write. Anything the *manager* refuses is reported in the manager's own words, per item, and the run exits `1` rather than claiming success.

### Certificate trust check correction

`src/client-agent/src/config.c` no longer resolves an unset `<verification_mode>` to `system`. It is a ladder: an explicit mode is honoured, `none` included; otherwise an explicit `<certificate_authorities>` gives `certificate`; otherwise a trust anchor at `etc/certs/root-ca.pem` gives `full`; otherwise `none`. `src/init/pkg_installer.sh` was updated to match and records the divergence in its own comments. The guide still described the previous behaviour, including an abort that no longer happens on the one path a migration takes: a 4.x agent, which cannot express TLS verification in a `<client>` block, is upgraded and runs unverified rather than aborting.

```
# src/init/pkg_installer.sh, no anchor and no explicit mode
- No trust anchor is present at ./etc/certs/root-ca.pem; the upgraded agent will run unverified
  unless <ssl><verification_mode> and <certificate_authorities> are configured explicitly.
```

### Results and Evidence

Lab: `~/wazuh/mig-39109` (Vagrant, VMware Fusion arm64). 4.14.7 AIO with agents web-ubuntu24 (001, `linux-servers,pci-scope`), win11-lab (002, `default` + `pci-scope` assigned with `agent_groups`), db-al2023 (003, `db-servers` + `pci-scope` assigned with `agent_groups`), API user `soc-analyst` with a custom role and policy; 1487 alerts, 6111 vulnerabilities, inventory, SCA and FIM data. Target: 5.0.0-beta5 AIO.

**Guides as written (re-enroll).** Ids after enrollment: 001 web-ubuntu24, 002 db-al2023, 003 win11-lab. Groups: win11-lab `default` only, db-al2023 `db-servers` only. API users: `wazuh`, `wazuh-wui` only.

```
### select id,name,`group` from agent  (5.0.0 global.db)
(1, 'web-ubuntu24', 'linux-servers,pci-scope')
(2, 'db-al2023', 'db-servers')
(3, 'win11-lab', 'default')
2026/09/10 12:04:54 wazuh-manager-modulesd:database: WARNING: Couldn't open directory 'etc/shared/db-servers' to check group 'db-servers': [(13)-(Permission denied)]. This is not a confirmed deletion, but a 'delete-group' will be sent anyway.
```

**New procedure (identity data carried).** Same ids, names, registration dates and groups as 4.x; all three 4.x agents `active` over 1514 without enrollment.

```
### select id,name,ip,version,os_type,`group`,connection_status,date_add from agent  (5.0.0 global.db)
(1, 'web-ubuntu24', '192.168.1.137', 'Wazuh v4.14.7', 'linux', 'linux-servers,pci-scope', 'active', 1789038797)
(2, 'win11-lab', '192.168.1.136', 'Wazuh v4.14.7', 'windows', 'default,pci-scope', 'active', 1789038798)
(3, 'db-al2023', '192.168.1.138', 'Wazuh v4.14.7', 'linux', 'db-servers,pci-scope', 'active', 1789039056)
### events per wazuh.agent.id (wazuh-events-v5-*)      ### findings (wazuh-findings-v5-*)
001 101 / 003 73 / 002 5                                 001 36 / 003 15 / 002 2
### wazuh-states-* for 4.x agents: 0 docs (as documented for the legacy channel)
```

Ownership matters on 5.0: with `client.keys` restored as `root:wazuh-manager 640` (the 4.x convention) authd exits:

```
2026/09/10 12:16:51 wazuh-manager-authd: ERROR: Unable to open etc/client.keys (key file)
2026/09/10 12:16:52 wazuh-manager-remoted:keystore: INFO: Loaded 3 agent key(s) from 'etc/client.keys'.
```

A raw 4.x `global.db` has `PRAGMA user_version = 0`; `wdb_upgrade_global()` accepts `1..N` only. The SQL in the guide was re-run against the exported 4.x file and the current `src/wazuh_db/schemas/schema_global.sql`, which gained `agent.reenroll_secret` in this cycle; migrated agents leave it `NULL`, which is the state the writer already handles.

**RBAC.** `CURRENT_ORM_VERSION` is `1` in both 4.14 and 5.0, so a carried `rbac.db` is never recognised as out of date and keeps its 4.x default policies: the 5.0 defaults, enrollment-token minting among them, are never created for any role. `check_database_integrity()` performs the supported upgrade when the stamped version is lower, so the guide sets it to `0`. Verified on the lab manager:

```
before: user_version 1 | policies 36 | roles 8 | users [(100,'soc-analyst'), (1,'wazuh'), (2,'wazuh-wui')]

INFO: RBAC database migration required. Current version is 0 but it should be 1. Upgrading RBAC database to version 1
INFO: /var/wazuh-manager/api/configuration/security/rbac.db database upgraded successfully

after:  user_version 1 | policies 18 | roles 9 | users [(100,'soc-analyst'), (1,'wazuh'), (2,'wazuh-wui')]
custom policies [(100, 'agents-read-lab')]   custom roles [(100, 'agents-readonly-lab')]   user_roles [(1,1), (2,1), (100,100)]

wazuh with the 4.x password -> {"data": {"token": "eyJhbGciOiJFUzUxMiIsInR5cCI6IkpXVCJ9...
soc-analyst -> token issued, GET /agents returns its three agents, GET /security/users filtered to empty
```

**Agent package upgrade 4.14.7 to 5.0.0-beta5 (deb) against the migrated manager.** Unattended `dpkg -i` blocked on the conffile prompt; meanwhile on disk:

```
-rw-r--r--  1 root  root     0 Sep  7 14:24 client.keys
-rw-r--r--  1 root  root  4148 Sep  7 14:24 ossec.conf        (<endpoint>MANAGER_IP</endpoint>)
2026/09/10 12:47:39 wazuh-agentd: ERROR: (4112): Invalid server address found: 'MANAGER_IP'
2026/09/10 12:47:39 wazuh-agentd: ERROR: (1215): No client configured. Exiting.
```

After `dpkg --configure --force-confold wazuh-agent`:

```
Configuration file '/etc/init.d/wazuh-agent' ==> Using current old file as you requested.
001 web-ubuntu24 any bb318ca7...                          (same key as on 4.x)
2026/09/10 12:59:20 wazuh-agentd: INFO: <agent><manager><endpoint> is not configured. Using <client><server><address> '192.168.1.141' with the default port 1517 and the default endpoint prefix 'wazuh-manager'.
2026/09/10 12:59:19 wazuh-agentd: WARNING: (1223): 'labels' is no longer supported and will be ignored. Agent labels were removed in 5.0.0.
2026/09/10 12:59:19 wazuh-agentd: INFO: 'client_buffer' is no longer used and will be ignored. Event batching is configured under <agent><batch>.
2026/09/10 12:59:20 wazuh-agentd:https-client: INFO: Starting https_client (server=192.168.1.141:1517, agent=001, body compression enabled).
```

Manager side three minutes later: agent 001 `v5.0.0`, `active`, groups kept, `wazuh-states-fim-files` 3206, `wazuh-states-sca` 279, inventory indices populated.

**Re-run on a post-tier-2 manager.** No 5.0.0 package newer than beta5 is published, so the manager and agent debs were built from source at `7a47160277`, the commit this branch is based on and the one carrying #39079, #39033 and #39282. The whole procedure was executed again against them.

```
### the two fail-closed conditions, verbatim
(1244): Invalid configuration at '/remote/https/certificate': file not found: /var/wazuh-manager/etc/certs/remoted.pem (the manager does not generate certificates; provision the file, e.g. with wazuh-certs-tool).
CRITICAL: Cannot start the HTTPS agent listener: the TLS private key 'etc/certs/remoted-key.pem' is missing or unreadable by the service user.

### GET /cacerts refuses a listener certificate the configured CA did not sign
{"error":"ca_mismatch"}
### after issuing remoted.pem from the deployment CA with the dialled address in its SAN
openssl verify -CAfile <(curl /cacerts) remoted.pem -> OK

### tool import onto the post-tier-2 schema
registry columns carried: 19, dropped by 5.0: os_codename, os_build, os_uname, config_sum, merged_sum, manager_host, node_name, group_config_status, filled in: os_type, connection_status, disconnection_time
registry columns new in 5.0 left at their default: reenroll_secret
Imported 3 agent(s), 4 group(s), 6 membership(s).

### after starting: agents keep their identity, across all three platforms
001 web-ubuntu24 v5.0.0   active ['linux-servers', 'pci-scope']
002 win11-lab    v4.14.7  active ['default', 'pci-scope']
003 db-al2023    v4.14.7  active ['db-servers', 'pci-scope']
### RBAC after the staged upgrade: 21 policies, 6 of them naming enrollment_token or read_secrets
### soc-analyst authenticates with its 4.x password and keeps its role
### tool check: exit 0

### agent package upgrade
wazuh-agent: removed the fleet-wide enrollment password at /var/ossec/etc/authd.pass. 5.0 enrols with an enrollment token (WAZUH_ENROLLMENT_TOKEN) and re-enrols with a per-agent secret.
### the upgraded agent with no anchor and no <ssl>
WARNING: TLS verification is DISABLED (verify_mode=none)   <- and it connects
### the same agent after the CA is placed at etc/certs/root-ca.pem
no warning, clean handshake, still active   <- resolved to 'full' against the anchor
```

The 6 policies naming the new tier 2 actions are the concrete argument for the `PRAGMA user_version = 0` step: without it a carried `rbac.db` keeps the 4.x defaults and none of them is ever created.

That run also produced the last commit on this branch. A 5.0 agent that reaches the manager while the registry is still empty is rejected with `401`, re-enrolls, and returns with a new id and its hostname as its name, which costs it the identity the procedure exists to preserve. The installer starts the manager, so the window is real on every migration; the guide now says to keep the agents off the air until the registry is restored.

**Verification pass.** Every claim in the four guides and the two reference pages was checked against the source, one reviewer per file. That found four statements that were wrong rather than imprecise, all now corrected: an agent upgraded by package with no `<ssl>` block and no anchor resolves to `none`, so it connects and verifies nothing instead of failing to connect; a 5.0 endpoint no longer writes `authd.pass` but still reads one placed by hand, which is what the runbook's own recovery depends on; `client_buffer` is a top-level element rather than a child of `<client>`, and the two positions log different messages; and `certificate_authorities` is inert under `none` with no warning, contrary to a note that predates this PR. Nine further corrections were precision: the WPK installer's abort list was missing two cases, its legacy exception depends on a package query and so does not hold on macOS, configuration preservation across a package upgrade is a Debian mechanism RPM achieves by other means, `wazuh-manager-apid` does issue its own API certificate, the event indices carry a `-v5` segment and the `wazuh-states-*` families of 4.x and 5.0 share names, a group is dropped when its directory cannot be opened rather than when it is half extracted, and `client.keys` ships `0660` with `authd.pass` owned by the service account.

```
15 tests, all passing, against the repository's own src/wazuh_db/schemas/schema_global.sql
```

`mdbook build` passes; every relative link and anchor in the new and edited pages resolves.

Unrelated to this PR but found by the same build: 105 of the 1337 internal links in the book resolve to a page that does not exist, 95 of them because mdBook publishes a `README.md` as `index.html` while rewriting `](README.md)` to `README.html`. Worth its own issue.

### Artifacts Affected

`tools/migration/wazuh-migrate-identity.py`, a new operator tool. No packaged artifact changes.

### Configuration Changes

None.

### Documentation Updates

- `tools/migration/README.md` (new)
- `docs/guide/migration/manager-4x-to-5x.md` (new)
- `docs/guide/migration/README.md`
- `docs/guide/migration/agent-groups-migration.md`
- `docs/guide/migration/upgrade-4x-to-5x.md`
- `docs/ref/upgrade.md`
- `docs/ref/modules/client/configuration.md`
- `docs/ref/modules/authd/configuration.md`
- `docs/SUMMARY.md`
- `CHANGELOG.md`

### Tests Introduced

`tools/migration/test_wazuh_migrate_identity.py`: 23 tests. `export` runs against a 4.x installation tree built in a temporary directory. `import` and `check` run against a stub of the manager's API that implements the endpoints the tool uses and records what it was asked to do, so the assertions are about the calls made rather than about files the tool no longer writes. The stub enforces the same content types the real endpoints do, which is how the `application/xml` requirement on the group-configuration endpoint was caught.

## Review Checklist

- [x] Code changes reviewed

The `no-changelog` label no longer applies: the tool is a product change and carries an entry under Manager / Added.


docs/guide/migration/manager-4x-to-5x.md
```diff
@@ -0,0 +1,494 @@
...
+
+Stop the 4.x manager so the files are consistent, then copy everything below off the host. The
+paths are the 4.x ones; the table says where each item lands on 5.0.
+
+```bash
+systemctl stop wazuh-manager
+mkdir -p /root/wazuh-4x-backup && cd /var/ossec
+cp etc/client.keys etc/authd.pass queue/db/global.db api/configuration/security/rbac.db /root/wazuh-4x-backup/
+cp etc/ossec.conf etc/local_internal_options.conf api/configuration/api.yaml /root/wazuh-4x-backup/
+tar -czf /root/wazuh-4x-backup/groups.tar.gz -C /var/ossec/etc --exclude='shared/default' --exclude='*/merged.mg' shared/*/
```
Antoniogm03: This tar never produces a group archive: `cd /var/ossec` leaves the shell outside `etc/`, so the shell expands `shared/*/` against `/var/ossec/shared`, which does not exist in 4.x, and tar receives the literal pattern (`tar: shared/*: Cannot stat`, exit 2, an empty `groups.tar.gz`). Step 2 then uninstalls 4.x, so the custom group folders are gone.

```suggestion
(cd /var/ossec/etc && tar -czf /root/wazuh-4x-backup/groups.tar.gz --exclude='shared/default' --exclude='*/merged.mg' shared/*/)
```

tools/migration/wazuh-migrate-identity.py
```diff
@@ -0,0 +1,874 @@
...
+           manifest.get("counts", {}).get("agents", 0),
+           len(manifest.get("groups", []))))
+    log("Target %s (%s)" % (target, version or "unknown version"))
+    if args.dry_run:
+        log("DRY RUN: nothing is written.")
+    log("")
+
+    # Everything that can refuse the import runs first, so a refusal never leaves the target
+    # holding half a migration.
+    columns = preflight_registry(args.bundle, target, args.force)
```
Antoniogm03: `restore_groups()` refuses an escaping tar member or a missing `etc/shared` only after `install_file()` and `transform_registry()` have run, so a refused import leaves exactly the half-applied state this block exists to prevent.

```suggestion
    columns = preflight_registry(args.bundle, target, args.force)

    archive = os.path.join(args.bundle, "groups.tar.gz")
    if os.path.isfile(archive):
        if not os.path.isdir(os.path.join(target, "etc", "shared")):
            raise MigrationError("'%s/etc/shared' does not exist; is it a 5.0 installation?"
                                 % target)
        with tarfile.open(archive, "r:gz") as tar:
            for member in tar.getmembers():
                if member.name.startswith("/") or ".." in member.name.split("/"):
                    raise MigrationError("refusing to extract '%s' from the bundle."
                                         % member.name)
                if member.issym() or member.islnk():
                    raise MigrationError("refusing to extract the link '%s' from the bundle."
                                         % member.name)
```

tools/migration/README.md
```diff
@@ -0,0 +1,104 @@
...
+cannot do. This performs that move and checks the result.
+
+It implements the identity half of
+[Migrating a Wazuh manager from 4.x to 5.0](../../docs/guide/migration/manager-4x-to-5x.md). It does
+not touch the configuration: `ossec.conf` has no mechanical translation to `wazuh-manager.conf`, and
+that guide covers it separately.
+
+Standard library only, so it runs on the 4.x host without installing anything. On a manager with no
+`sqlite3` package the bundled interpreter works too:
+`/var/wazuh-manager/framework/python/bin/python3`.
```
Antoniogm03: This sentence is about the 4.x host, where `export` runs, but the path is the 5.0 one, on a 4.x manager the bundled interpreter is `/var/ossec/framework/python/bin/python3`. The operator gets "No such file or directory" on the one host the fallback was meant to help; the same path is correct in `manager-4x-to-5x.md:46`, where it refers to the 5.0 host.

```suggestion
`/var/ossec/framework/python/bin/python3` on the 4.x host, and
`/var/wazuh-manager/framework/python/bin/python3` on the 5.0 one.
```

tools/migration/wazuh-migrate-identity.py
```diff
@@ -0,0 +1,874 @@
...
+    manifest["groups"] = names
+    if names:
+        log("  export   %-14s <- %s (%s)" % ("groups", shared, ", ".join(names)))
+        if not args.dry_run:
+            archive = os.path.join(args.bundle, "groups.tar.gz")
+            with tarfile.open(archive, "w:gz") as tar:
+                for name in names:
+                    tar.add(os.path.join(shared, name), arcname=name,
+                            filter=lambda info: None if os.path.basename(
+                                info.name) == "merged.mg" else info)
```
Antoniogm03: **[S3 · non-blocking · introduced]** `tarfile` stores links verbatim by default, but `restore_groups()` refuses every link member, so a symlink anywhere under `etc/shared/<group>/` produces a bundle that can never be imported. Reproduced: `export` exits 0 without a warning, `import` answers `refusing to extract the link 'pci-scope/shared.conf'` and exits 2 with `client.keys` already written — and the only way out is the 4.x host, which by then is decommissioned. Reject links while collecting, when the source still exists.

```suggestion
                    for root, directories, files in os.walk(os.path.join(shared, name)):
                        for entry in directories + files:
                            if os.path.islink(os.path.join(root, entry)):
                                raise MigrationError(
                                    "'%s' is a symbolic link. 'import' refuses link members,"
                                    " so resolve or remove it before exporting."
                                    % os.path.join(root, entry))
                    tar.add(os.path.join(shared, name), arcname=name,
                            filter=lambda info: None if os.path.basename(
                                info.name) == "merged.mg" else info)
```

tools/migration/wazuh-migrate-identity.py
```diff
@@ -0,0 +1,874 @@
...
+        print("error: %s" % error, file=sys.stderr)
+        return 2
+    except PermissionError as error:
+        print("error: %s. Run as root." % error, file=sys.stderr)
+        return 2
+    except OSError as error:
+        if error.errno == errno.ENOSPC:
+            print("error: no space left on device.", file=sys.stderr)
+            return 2
+        raise
```
Antoniogm03: `main()` handles only `MigrationError`, `PermissionError` and `OSError`, so a `manifest.json` truncated in transfer (`json.JSONDecodeError`) or a locked/corrupt registry (`sqlite3.OperationalError`) exits with a Python traceback and status 1, the code the README's table reserves for "`check` found problems". On the import path that happens once `client.keys` is already on disk, so neither the message nor the exit code says what state the manager is in.

```suggestion
    except OSError as error:
        if error.errno == errno.ENOSPC:
            print("error: no space left on device.", file=sys.stderr)
            return 2
        raise
    except (sqlite3.Error, ValueError) as error:
        print("error: %s. The target may hold a partial migration; restore the"
              " .pre-migration copies before retrying." % error, file=sys.stderr)
        return 2
```

tools/migration/wazuh-migrate-identity.py
```diff
@@ -0,0 +1,874 @@
...
+
+DEFAULT_SOURCE_DIR = "/var/ossec"
+DEFAULT_TARGET_DIR = "/var/wazuh-manager"
+
+# The 5.0 schema the transform below was written against. wazuh-db refuses anything outside
+# 1..N, so a target stamped differently means this tool predates the installed manager.
+SUPPORTED_TARGET_USER_VERSION = (1,)
+# 4.x stamps PRAGMA user_version 0 and tracks its own schema in metadata.db_version. Only the
+# revisions whose agent table this transform has been checked against are accepted.
+SOURCE_DB_VERSIONS_CHECKED = (7, 8, 9, 10, 11, 12)
```
Antoniogm03: **[S3 · non-blocking · introduced]** Revisions 8-12 have never been stamped by any 4.x release, so they cannot have been checked against anything, which inverts the comment above this line. `db_version` has gone 1 (v4.0.0) → 3 (v4.3.0) → 4 (v4.4.0) → 6 (v4.8.0, v4.12.0) → 7 (v4.14.0 … v4.14.8-rc2), and 7 is the only value the tests build. If the still-maintained 4.14.x line bumps to 8 with a renamed or re-semanticised `agent` column, the tool accepts it silently and the column-intersection transform mis-copies instead of refusing.

```suggestion
SOURCE_DB_VERSIONS_CHECKED = (7,)
```

tools/migration/wazuh-migrate-identity.py
```diff
@@ -0,0 +1,874 @@
...
+    for name in ("etc/wazuh-manager.conf", "etc/ossec.conf"):
+        path = os.path.join(install_dir, name)
+        if not os.path.isfile(path):
+            continue
+        try:
+            with open(path, errors="replace") as handle:
+                content = handle.read()
+        except OSError:
+            continue
+        lowered = content.lower().replace(" ", "")
```
Antoniogm03: **[S3 · non-blocking · introduced]** This strips spaces but not newlines or tabs, so `<node_type>\n    worker\n  </node_type>` — valid XML, and what any reformatter or config-management template produces — does not match the literal, and the import proceeds into a worker whose registry and keys are supposed to come from the master. The loop also `return`s after the first of the two config files that exists, so a stub `wazuh-manager.conf` beside a real `ossec.conf` skips the check entirely. Separately, `command_export()` never calls this at all, although the issue asks the tool to refuse to run on a worker.

```suggestion
        lowered = "".join(content.lower().split())
```

tools/migration/wazuh-migrate-identity.py
```diff
@@ -0,0 +1,874 @@
...
+def backup_existing(path, dry_run):
+    """Keeps whatever is about to be overwritten, once, beside the original."""
+    if not os.path.exists(path):
+        return None
+    backup = "%s.pre-migration" % path
+    if os.path.exists(backup):
+        return backup
+    if not dry_run:
+        shutil.copy2(path, backup)
+    return backup
```
Antoniogm03: When `.pre-migration` already exists this returns it without copying, so a second import overwrites a live file with no backup while `install_file()` still logs `[backed up]` and `README.md:86` still promises "a `.pre-migration` copy of every file it overwrites".

```suggestion
    backup = "%s.pre-migration" % path
    if os.path.exists(backup):
        # Never overwrite an earlier backup, but never skip one either: a second import
        # would otherwise replace a live file with no copy of it anywhere.
        backup = "%s.pre-migration.%s" % (path, time.strftime("%Y%m%d%H%M%S"))
    if not dry_run:
        shutil.copy2(path, backup)
    return backup
```

tools/migration/wazuh-migrate-identity.py
```diff
@@ -0,0 +1,874 @@
...
+                           ("rbac.db", SOURCE_PATHS["rbac.db"])):
+        path = os.path.join(source, relative)
+        if not os.path.isfile(path):
+            if name == "client.keys":
+                raise MigrationError("'%s' not found: nothing to migrate." % path)
+            log("  skip     %-14s (not present)" % name)
+            continue
+        if name == "authd.pass" and not args.with_password:
+            log("  skip     %-14s (--with-password not given)" % name)
+            continue
```
Antoniogm03: `authd.pass` is gated at collection time by `--with-password`, but `rbac.db`, every API user's password hash, is always copied into the bundle; `--with-rbac` exists only on `import`. An operator who never intends to migrate the API users still moves all of their credentials to another host, while the README's "What moves" table reads as if the flag governed both ends. Moving `--with-rbac` into the loop that already adds `--dry-run`/`--force` to both subparsers, and adding the matching `skip` here, makes the two secrets symmetrical.

docs/guide/migration/manager-4x-to-5x.md
```diff
@@ -0,0 +1,494 @@
...
+The API listener is the exception: `wazuh-manager-apid` still issues its own self-signed certificate
+on first start, and needs nothing from you. Two requirements come from the fleet rather than from
+the installer:
+
+- **The subjectAltName must carry every address agents dial.** For a migration that is the address
+  the 4.x agents already have in their `ossec.conf`, plus the cluster VIP and any NAT address. The
+  manager cannot check this and warns only when the certificate carries no usable SAN at all.
+- **Keep `root-ca.pem` and its key.** It is what every 5.0 agent verifies the manager against and
+  what enrollment tokens pin, so reissuing it later invalidates both. Certificates are not
+  synchronized between cluster nodes: every node's listener certificate must be issued by this CA.
```
Antoniogm03: **[S3 · non-blocking · introduced]** Issue #39109 scopes the environment as "manager (single node, **and a master + worker pair**)", but workers only appear in passing here and in the `authd.pass` note — there is no order of operations for rebuilding them, nothing on the cluster key, and no statement that the identity restore belongs on the master only. That last rule is one the tool already enforces (`assert_master`), so the guide and the tool disagree about how much the reader is told. `command_export()` also never calls `assert_master`, although the issue's risk list asks the tool to refuse to run on a worker.

tools/migration/test_wazuh_migrate_identity.py
```diff
@@ -0,0 +1,399 @@
...
+        with tarfile.open(archive, "w:gz") as tar:
+            tar.add(escape, arcname="../../etc/escaped.conf")
+        manifest_path = os.path.join(self.bundle, "manifest.json")
+        with open(manifest_path) as handle:
+            manifest = json.load(handle)
+        manifest["files"]["groups.tar.gz"] = tool.sha256_of(archive)
+        with open(manifest_path, "w") as handle:
+            json.dump(manifest, handle)
+        self.assertEqual(2, self.do_import())
+        self.assertFalse(os.path.exists(os.path.join(self.target, "etc", "escaped.conf")))
```
Antoniogm03: **[S3 · non-blocking · introduced]** This asserts the exit code and that `escaped.conf` was not created, but not that nothing else was written — and at that point the registry holds 4 agents and `client.keys` has been replaced. `test_import_refuses_a_populated_registry_without_writing` does assert that property for the registry refusal; the group-archive refusal, which is the one that actually breaks it, has no equivalent.

```suggestion
        self.assertEqual(2, self.do_import())
        self.assertFalse(os.path.exists(os.path.join(self.target, "etc", "escaped.conf")))
        with self.target_registry() as connection:
            self.assertEqual(0, connection.execute(
                "SELECT count(*) FROM agent WHERE id > 0").fetchone()[0],
                "a refused import must not have written the registry first")
        self.assertEqual(0, os.path.getsize(os.path.join(self.target, "etc", "client.keys")),
                         "a refused import must not have installed client.keys first")
```

docs/ref/modules/client/configuration.md
```diff
@@ -118,30 +118,51 @@ Path to the private key matching `<certificate>`.
 
 Path to the CA bundle used to verify the manager's certificate.
 
-- **Default value:** None
+- **Default value:** the trust anchor `etc/certs/root-ca.pem` when it exists and the resolved
+  `<verification_mode>` is `full` or `certificate`; otherwise none. See
+  [`verification_mode`](#verification_mode).
 - **Allowed values:** Path to a PEM-encoded CA bundle file, readable by the agent
 - **Required:** Yes, when `<verification_mode>` is `full` or `certificate` -- the agent fails
   closed (refuses to start) without a readable CA file in that case.
```
Antoniogm03: The new "Default value" bullet three lines above says the on-disk anchor fills this in under `full`/`certificate`, which `w_agent_resolve_ssl_posture()` confirms (`src/client-agent/src/config.c:227-234`), but this sentence still says the agent refuses to start in those modes without a readable CA file. As written, a reader configures a redundant explicit path, or concludes the anchor-only setup the migration guide recommends will not boot. The sentence was correct before this PR redefined the default, hence aggravated rather than pre-existing.

```suggestion
- **Required:** Yes, when `<verification_mode>` resolves to `full` or `certificate` **and no
  trust anchor is on disk** -- the agent fails closed (refuses to start) with no readable CA
  file in that case. With the anchor present it is the default above, and no explicit path is
  needed.
```

Antoniogm03 (changes requested):

PR #39207, opened 2026-09-11
Fix agent crash from unlocked reads of the FIM directories list on the whodata thread

### LINKED ISSUE #39201: FIM reads syscheck.directories without the directories_lock from the whodata inode-search callback and the eBPF configuration callback, while the scan thread frees and unlinks entries, giving a use-after-free that crashes the agent
## Description

`syscheck.directories` is a linked list of `directory_t*` guarded by `syscheck.directories_lock`, a read/write lock. The discipline is applied deliberately across the module: I counted 29 acquisitions across the FIM sources, and functions such as `getSyscheckConfig()` and `fim_has_configured_directories()` take the read lock even for a single traversal.

Two paths reach the same list with no lock at all, on threads that run concurrently with the writer.

### Reader 1: the whodata inode-search callback

`fim_whodata_event()` is careful about this. Both of its own calls are bracketed by the read lock:

```c
170  void fim_whodata_event(whodata_evt * w_evt) {
...
179          w_rwlock_rdlock(&syscheck.directories_lock);
180          fim_checker(w_evt->path, &evt_data, NULL, NULL, NULL);
181          w_rwlock_unlock(&syscheck.directories_lock);
182      } else {
183          // Otherwise, it could be a file deleted or a directory moved (or renamed).
184          w_rwlock_rdlock(&syscheck.directories_lock);
185          fim_process_missing_entry(w_evt->path, FIM_WHODATA, w_evt);
186          w_rwlock_unlock(&syscheck.directories_lock);
187  #ifndef WIN32
188          const unsigned long int inode = strtoul(w_evt->inode, NULL, 10);
189          const unsigned long int device = strtoul(w_evt->dev, NULL, 10);
190          callback_context_t callback_data;
191          callback_data.callback = create_unix_who_data_events;
192          callback_data.context = w_evt;
193          fim_db_file_inode_search(inode, device, callback_data);
194  #endif
195      }
196  }
```

Line 193 is outside the unlock on line 186, and the callback it installs does exactly what lines 184 to 186 were protecting:

```c
161  // Callback
162  void create_unix_who_data_events(void * data, void * ctx)
163  {
164      char *path = (char *)data;
165      whodata_evt *w_evt = (whodata_evt *)ctx;
166
167      fim_process_missing_entry(path, FIM_WHODATA, w_evt);
168  }
```

```c
1208  void fim_process_missing_entry(char * pathname, fim_event_mode mode, whodata_evt * w_evt) {
1209      directory_t *configuration = NULL;
1210
1211      configuration = fim_configuration_directory(pathname, true, syscheck.directories);
```

So the same function, reached through the callback one line later, runs unprotected. The asymmetry inside a single `else` block is what makes this look like an oversight rather than a decision.

### Reader 2: the eBPF configuration callback

```c
901  static directory_t *fim_configuration_directory_ebpf(const char *path, bool notify_not_found) {
902      return fim_configuration_directory(path, notify_not_found, syscheck.directories);
903  }
904
905  void check_ebpf_availability() {
906      minfo(FIM_EBPF_INIT);
907      fimebpf_initialize(fim_configuration_directory_ebpf, get_user, get_group, fim_whodata_event,
908                         free_whodata_event, loggingFunction, abspath, fim_shutdown_process_on, syscheck.queue_size);
```

This is handed to the eBPF module as its configuration-lookup callback and is invoked from that module's own event thread for every whodata event. It takes no lock anywhere, and there is no caller in this codebase that could be holding one on its behalf.

### What the reader does with the list

`fim_configuration_directory()` walks the nodes, dereferences each entry, and returns a raw pointer into the list, so the caller keeps using list memory after the call returns:

```c
565  directory_t *fim_configuration_directory(const char *key, bool notify_not_found, const OSList *directories_list) {
...
594      OSList_foreach(node_it, (OSList *)directories_list) {
595          dir_it = node_it->data;
596          char *real_path = fim_get_real_path(dir_it);
...
601          if (top < match && full_path[match - 1] == PATH_SEP) {
602              dir = dir_it;
...
616      return dir;
```

### The writer

`update_wildcards_config()` rebuilds the list on the scheduled scan thread. It is the first thing `fim_scan()` does:

```c
68      update_wildcards_config();
69
70      // First file scan
71      fim_file_scan();
```

Inside, under the write lock, it both frees entries and unlinks nodes:

```c
326      w_rwlock_wrlock(&syscheck.directories_lock);
...
355              fim_insert_directory(syscheck.directories, new_entry);
...
366      OSList_SetFreeDataPointer(removed_entries, (void (*)(void *))free_directory);
...
385              aux_it = OSList_GetNext(syscheck.directories, node_it);
386              // Delete node
388              OSList_DeleteThisNode(syscheck.directories, node_it);
...
394      w_rwlock_unlock(&syscheck.directories_lock);
```

`fim_insert_directory()` is itself a free site, because a repeated path is a replacement rather than a merge:

```c
free_directory(dir_it);
node_it->data = new_entry;
```

So while the write lock is held, entries the readers above may be pointing at are freed, and nodes they may be walking are unlinked.

## Why this crashes

The two unlocked readers run on the whodata thread (audit or eBPF); the writer runs on the scheduled scan thread. Nothing serialises them. Three distinct failure modes follow, in rough order of likelihood:

1. The reader is inside `OSList_foreach` when `OSList_DeleteThisNode()` unlinks the node it is standing on. The next hop dereferences freed node memory.
2. The reader holds `dir_it` from `node_it->data` when `fim_insert_directory()` calls `free_directory(dir_it)` for a replaced duplicate, then reads `dir_it->path` inside `fim_get_real_path()`.
3. The reader returns a `directory_t*` that the writer frees immediately afterwards, and `fim_process_missing_entry()` goes on to use it through `configuration`.

All three are use-after-free on the agent's own heap. The visible outcome is a SIGSEGV in the agent, which stops FIM and leaves the endpoint unmonitored until the service is restarted. Because the freed memory is frequently reallocated with attacker-independent data, the more common symptom in the field is likely an intermittent unexplained agent crash rather than a clean reproducible fault.

## Conditions required

All of these hold in a default Linux agent that uses whodata:

- FIM in whodata mode, with either the audit or the eBPF provider. Reader 1 needs `#ifndef WIN32`, so this is Linux and macOS, not Windows.
- At least one `<directories>` entry containing a wildcard, which is what makes `update_wildcards_config()` mutate the list rather than leave it alone.
- A file deletion or a directory move or rename inside a monitored path, which is what drives `fim_whodata_event()` down the `else` branch to line 193.
- That event landing while the scheduled scan is inside `update_wildcards_config()`.

The window is short, which is why this presents as a rare crash rather than a consistent one, but it recurs on every scan cycle for the lifetime of the agent.

## Scope and what I did not do

I found this by auditing every function that touches `syscheck.directories` against the lock, not by observing a crash. I could not reproduce it on a host: the demo environment's Linux agents are not reachable from here, and this needs a whodata-enabled agent with wildcard entries plus deliberate timing pressure. So I am reporting a race established from the code, with the exact line numbers above, and the reproduction below is the procedure I would use rather than one I have run.

For completeness, the same audit flagged five other functions that touch the list without taking the lock themselves. I checked each and they are covered by their callers, so they are not part of this report:

```
  file.c          fim_process_missing_entry()      covered at fim_scan.c:184 and run_check.c
  run_realtime.c  fim_realtime_delete_watches()    covered by fim_scan.c:326 write lock
  run_realtime.c  realtime_update_watch()          covered at run_realtime.c:397
  syscheck.c      drop_orphaned_promoted_documents()
  syscheck.c      persist_sync_documents()
  syscheck.c      Start_win32_Syscheck()           startup, before the other threads exist
```

`fim_process_missing_entry()` appears in that list because it takes no lock of its own. It is safe from `fim_whodata_event()` line 185 and unsafe from the callback at line 193. That is the defect.

If triage concludes this is exploitable memory corruption rather than a crash, it should be moved to the security process rather than handled here. I am claiming a reachable use-after-free that terminates the agent, not code execution.

## Environment

- Source read from the `5.0.0` branch of `wazuh/wazuh`:
  - `src/syscheckd/src/fim_scan.c` (`fim_whodata_event`, `create_unix_who_data_events`, `update_wildcards_config`)
  - `src/syscheckd/src/file/file.c` (`fim_process_missing_entry`, `fim_configuration_directory`)
  - `src/syscheckd/src/syscheck.c` (`fim_configuration_directory_ebpf`, `check_ebpf_availability`)
  - `src/config/src/syscheck-config.c` (`fim_insert_directory`)
- QA demo environment `wqa-prod-58-demo-environment` for the version reference. Not a container deployment, so there are no image digests to record.

## Steps to reproduce

1. Build the agent for Linux with whodata support, ideally with ThreadSanitizer (`-fsanitize=thread`) or AddressSanitizer (`-fsanitize=address`), which turns this from a rare crash into a deterministic report.
2. Configure FIM with whodata and a wildcard directory, and a short scan frequency so the writer runs often:

```xml
<syscheck>
  <frequency>30</frequency>
  <directories whodata="yes" recursion_level="3">/tmp/fimtest/dir*</directories>
</syscheck>
```

3. Create and remove matching directories continuously, so `update_wildcards_config()` genuinely adds and removes entries on each pass rather than finding the list unchanged:

```bash
while true; do
  mkdir -p /tmp/fimtest/dir$((RANDOM % 50))
  rmdir /tmp/fimtest/dir* 2>/dev/null
done
```

4. In parallel, generate whodata delete events inside the monitored tree, which is what drives line 193:

```bash
while true; do
  d=/tmp/fimtest/dir$((RANDOM % 50)); mkdir -p "$d"
  touch "$d/f"; rm -f "$d/f"
done
```

5. Under TSan, expect a data race reported on `syscheck.directories` between `update_wildcards_config()` on the scan thread and `fim_configuration_directory()` reached from `create_unix_who_data_events()`. Under ASan, expect a heap-use-after-free with the free stack in `free_directory()`. Without a sanitizer, expect an intermittent SIGSEGV in the agent.

6. To confirm the fix, take the read lock around `fim_db_file_inode_search()` at fim_scan.c:193, or inside `create_unix_who_data_events()`, and inside `fim_configuration_directory_ebpf()`, and repeat.

## Expected

Every read of `syscheck.directories` holds `syscheck.directories_lock` for the whole time it traverses the list and uses any `directory_t*` obtained from it, as the rest of the module already does.

## Actual

`create_unix_who_data_events()`, installed as the callback for `fim_db_file_inode_search()` at fim_scan.c:193, and `fim_configuration_directory_ebpf()`, installed as the eBPF module's configuration callback at syscheck.c:907, both reach `fim_configuration_directory()` with no lock held, on the whodata thread. `update_wildcards_config()` concurrently frees `directory_t` entries and unlinks list nodes from the scan thread. The result is a use-after-free that terminates the agent.

## Not a duplicate of

`qa-known-issues.py check` scored two cache candidates at 0.40, both of which I read:

- `wazuh/wazuh#8800` "Windows agent does not validate correctly configuration" concerns validation of configuration values on Windows at load time. It is not about locking, threads or memory lifetime.
- `wazuh/wazuh#2931` "Centralized configuration filtering by Agent version" is a feature request about distributing configuration by agent version.

Neither concerns concurrent access to the FIM directory list. The live search for `syscheck.directories directories_lock use-after-free configuration callback whodata` returned nothing.

Darioortegaleyva:
## Description

This PR fixes a use-after-free that crashes the agent: two readers of `syscheck.directories` run on the whodata thread without holding `syscheck.directories_lock`, while `update_wildcards_config()` frees `directory_t` entries from the scheduled scan thread.

The unlocked readers are the whodata inode-search callback `create_windows_who_data_events()` in `create_db.c` (the `#ifndef WIN32` path, despite the name), invoked from `fim_db_file_inode_search()` after the read lock has been released, and `fim_configuration_directory`, handed straight to `fimebpf_initialize()` at `syscheck.c:324` with no locking wrapper. Both reach `fim_configuration_directory()`, which walks the list and returns a raw pointer into it, so the entry can be freed under them mid-traversal.

For the whodata reader the lock is taken by widening the existing bracket in `fim_whodata_event()` so that it also covers `fim_db_file_inode_search()`, which keeps the pointer in use inside the locked region. Locking inside the callback instead would also remove the data race, but it is the worse option: on `5.0.0` it introduces a lock-order inversion, because `FIMDB::executeQuery()` holds `m_handlersMutex` across the callback while every other path takes `directories_lock` first and the `libfimdb` mutex second, and ThreadSanitizer flags it (evidence below). On `4.14.9` that inversion does not exist, since `FIMDB::executeQuery()` takes no lock and `SQLiteDBEngine::selectData()` releases its own around the callback, but widening the bracket is still the right shape here: it is the only one that keeps the returned `directory_t*` protected for as long as it is used, and it keeps both branches consistent for the forward-port.

For the eBPF reader the callback contract changes from `directory_t *(*)(const char *)` to `bool (*)(const char *)`. Returning the pointer is what left the use-after-free open: the eBPF consumer only needs to know whether the path has `WHODATA_ACTIVE`, and it dereferenced `config->options` after the lookup had already released the lock. `libfimebpf` is a separate shared library and cannot take `syscheck.directories_lock` itself, so the answer is computed under the lock and only the boolean crosses the boundary.

**Issue:** wazuh/wazuh#39201
**Internal Reference:** N/A

## Proposed Changes

- Updated `src/syscheckd/src/create_db.c`: the read lock in `fim_whodata_event()` now also covers `fim_db_file_inode_search()`, whose callback reaches `fim_configuration_directory()`.
- Updated `src/syscheckd/src/syscheck.c`: added `fim_whodata_active_ebpf()`, which resolves the configuration and evaluates `WHODATA_ACTIVE` while holding `syscheck.directories_lock` for reading, and handed it to `fimebpf_initialize()` in place of the bare lookup.
- Updated the eBPF configuration callback contract to `bool (*)(const char *)` in `src/syscheckd/src/ebpf/include/ebpf_whodata.h`, `ebpf_whodata.hpp` and `ebpf_whodata.cpp`, so no pointer into `syscheck.directories` leaves the locked region.
- Updated `src/syscheckd/src/ebpf/tests/fimEbpfWhodataTest/ebpf_mock_utils.hpp` and `src/unit_tests/wrappers/linux/ebpf_wrappers.{c,h}` to the new signature. The test wrapper also stops declaring a by-value `syscheck_config` where the real function takes an `unsigned int`.
- Updated `CHANGELOG.md`.

### Results and Evidence

Validated twice: at unit level with a harness over the real functions, and end to end on a real agent with whodata.

Both runs exercise the whodata reader with the `audit` provider, which is the one the reported crash goes through. The eBPF reader is not covered by a run: with the callback no longer returning a pointer, there is nothing left for a sanitizer to catch there, and the change is a contract narrowing verified by construction and by a clean build of every consumer (module, shared library, mocks and test wrappers).

Both runs were done on `5.0.0`, which is the branch the report cites and the one the reproduction steps target. On `4.14.9` the two readers, the writer and the call graph between them are the same code (see Version scope below), and the agent builds clean with the change. A ThreadSanitizer build of `4.14.9` was attempted for a like-for-like run and abandoned: `syscheckd/CMakeLists.txt` in `4.14.x` overrides `CMAKE_C_FLAGS_DEBUG` so the sanitizer flag does not reach the module, and injecting it through `CC` breaks the autoconf of the external dependencies. Reproducing on `4.14.9` therefore needs local build-system changes; it was not required to establish the defect or to validate the fix, but it is available on request.

#### End-to-end, real agent

VM `bento/ubuntu-24.04`, kernel 6.8.0-86, `auditd` running, whodata in `audit` mode. Agent built from `5.0.0` at `b1d2060f477` with ThreadSanitizer on the C sources:

```bash
CFLAGS="-fsanitize=thread -g" LDFLAGS="-fsanitize=thread" make TARGET=agent DEBUG=yes -j$(nproc)
```

The load is the one from the issue: a whodata wildcard `<directories>` entry whose expansion genuinely changes between scans, plus continuous delete events on files the previous scan had indexed. Before the change, ThreadSanitizer reports the race between the scan thread and the whodata thread:

```sql
WARNING: ThreadSanitizer: data race
  Write of size 8 by main thread (mutexes: write M0):
    #1 free_directory                  src/config/src/syscheck-config.c:2269
    #2 fim_insert_directory            src/config/src/syscheck-config.c:155
    #3 update_wildcards_config         src/syscheckd/src/fim_scan.c:356
    #4 fim_scan                        src/syscheckd/src/fim_scan.c:68
    #5 start_daemon                    src/syscheckd/src/run_check.c:1046

  Previous read of size 4 by thread T4 (mutexes: read M1):
    #0 fim_get_data                    src/syscheckd/src/file/file.c:820
    #1 fim_file                        src/syscheckd/src/file/file.c:1113
    #2 fim_checker                     src/syscheckd/src/file/file.c:1012
    #3 fim_db_process_missing_entry    src/syscheckd/src/file/file.c:548
    #5 fim_process_missing_entry       src/syscheckd/src/file/file.c:1232
    #6 create_unix_who_data_events     src/syscheckd/src/fim_scan.c:168
    #7 operator()                      src/syscheckd/src/db/src/file.cpp:370
    #8 audit_parse                     src/syscheckd/src/whodata/audit_parse.c:805
    #9 audit_parse_thread              src/syscheckd/src/whodata/syscheck_audit.c:861

  Mutex M0 (0x5555556d2c28) created at:
    #1 fim_initialize                  src/syscheckd/src/syscheck.c:554

  Mutex M1 (0x7ffff6dfe020) created at:
    #1 __glibcxx_rwlock_wrlock         /usr/include/c++/13/shared_mutex:81

SUMMARY: ThreadSanitizer: data race src/config/src/syscheck-config.c:2269 in free_directory
```

M0 is `syscheck.directories_lock`, held for writing by the scan thread. M1 is an unrelated `std::shared_mutex` inside `libfimdb`: the whodata thread holds no `directories_lock` at all.

Locking inside the callback instead, which is the other option considered, removes that race but produces:

```sql
WARNING: ThreadSanitizer: lock-order-inversion (potential deadlock)
  Cycle in lock order graph: M0 => M1 => M0

  Mutex M1 acquired here while holding mutex M0 in thread T4:
    #2 fim_process_missing_entry       src/syscheckd/src/file/file.c:1232
    #3 fim_whodata_event               src/syscheckd/src/fim_scan.c:188

  Mutex M0 acquired here while holding mutex M1 in thread T4:
    #1 create_unix_who_data_events     src/syscheckd/src/fim_scan.c:168
    #2 operator()                      src/syscheckd/src/db/src/file.cpp:370
```

With the change in this PR, the same VM under heavier load than the failing run (420 s, 13 `update_wildcards_config()` passes, 7 entries removed, 2529 calls into the callback, 4162 delete events) reports neither the data race nor the inversion.

Three adjustments were needed to make the issue's own steps reachable on `5.0.0`; none of them changes the defect:

- A non-wildcard `<directories>` entry has to be present. `fim_has_configured_directories()` (`run_check.c:297`) counts only `is_wildcard == 0` entries, so a wildcard-only configuration makes `start_daemon()` log `No monitored paths configured. FIM module exiting after DataClean.` and FIM never starts. Reported separately.
- The monitored tree needs files already indexed in the FIM database, since `fim_db_file_inode_search()` only invokes the callback for inodes it finds there.
- `<frequency>` lowered from 30 to 1. The window is a few microseconds per scan, and at 30 a 420 s run gives only 14 writer passes.

#### Unit level

A two-thread harness linked against the real `libsyscheckd_lib.a` drives the production writer (`update_wildcards_config()`) and the production whodata callback (`create_unix_who_data_events()`) concurrently over a wildcard configuration, so that every writer pass hits the duplicate-replacement branch of `fim_insert_directory()`. Before the change, 30 s of that load produce nine ThreadSanitizer reports ending in a segmentation fault:

```sql
SUMMARY: ThreadSanitizer: data race            src/config/src/syscheck-config.c:156  in fim_insert_directory
SUMMARY: ThreadSanitizer: data race            src/config/src/syscheck-config.c:2260 in free_directory
SUMMARY: ThreadSanitizer: data race            src/config/src/syscheck-config.c:2269 in free_directory
SUMMARY: ThreadSanitizer: data race            src/syscheckd/src/file/file.c:715     in fim_get_real_path
SUMMARY: ThreadSanitizer: data race            src/syscheckd/src/file/file.c:716     in fim_get_real_path
SUMMARY: ThreadSanitizer: data race            src/syscheckd/src/file/file.c:1223    in fim_process_missing_entry
SUMMARY: ThreadSanitizer: heap-use-after-free  src/syscheckd/src/file/file.c:715     in fim_get_real_path
SUMMARY: ThreadSanitizer: heap-use-after-free  src/syscheckd/src/file/file.c:716     in fim_get_real_path
SUMMARY: ThreadSanitizer: SEGV (libc.so.6+0x19badc)
```

```sql
ThreadSanitizer:DEADLYSIGNAL
ERROR: ThreadSanitizer: SEGV on unknown address 0x000000000000
The signal is caused by a READ memory access.
    #1 strdup
    #3 fim_get_real_path              src/syscheckd/src/file/file.c:716
    #4 fim_configuration_directory     src/syscheckd/src/file/file.c:597
    #5 fim_process_missing_entry       src/syscheckd/src/file/file.c:1211
    #6 create_unix_who_data_events     src/syscheckd/src/fim_scan.c:167
```

`free_directory()` clears the `path` field before releasing the block, so the reader reaches `strdup(NULL)`. After the change, the same harness over the same load reports nothing across roughly 130,000 reader passes racing 118,000 writer passes.

The unit tests covering `fim_whodata_event()` and the callback already declare `expect_function_call_any` for `__wrap_pthread_rwlock_rdlock` and `__wrap_pthread_rwlock_unlock`, so they cover the wider bracket without modification.

Version scope: the reported issue targets `5.0.0` and the reproduction above was run there, because that is the branch the report cites. The defect is identical on `4.14.9` (`src/syscheckd/src/create_db.c:976` for the callback, installed at line 1001 and invoked at line 1003 after the read lock is released at line 995, and `src/syscheckd/src/syscheck.c:324` for the eBPF lookup), so it is a production bug rather than a 5.0.0 regression. This PR lands the fix on `4.14.9`.

Not affected: Windows. The unlocked inode-search reader is inside `#ifndef WIN32`, and `src/syscheckd/src/whodata/win_whodata.c` takes `directories_lock` at its own call sites.

Out of scope, reported separately: `update_wildcards_config()` frees the removed entries after releasing the write lock (on `4.14.9`, `OSList_Destroy(removed_entries)` at `create_db.c:1879`, unlock at line 1872). A reader that correctly holds the read lock is still unprotected against those frees if it keeps the returned `directory_t*` past its own unlock, because `fim_configuration_directory()` returns a raw pointer into the list.

### Artifacts Affected

- `src/syscheckd/src/create_db.c`
- `src/syscheckd/src/syscheck.c`
- `src/syscheckd/src/ebpf/include/ebpf_whodata.h`
- `src/syscheckd/src/ebpf/include/ebpf_whodata.hpp`
- `src/syscheckd/src/ebpf/src/ebpf_whodata.cpp`
- `src/syscheckd/src/ebpf/tests/fimEbpfWhodataTest/ebpf_mock_utils.hpp`
- `src/unit_tests/wrappers/linux/ebpf_wrappers.c`
- `src/unit_tests/wrappers/linux/ebpf_wrappers.h`
- `CHANGELOG.md`

### Configuration Changes

N/A

### Documentation Updates

N/A

### Tests Introduced

- **Scope:** Manual validation with ThreadSanitizer
- **Details:** Two validations. A two-thread harness over the real functions reports six data races, two heap-use-after-free and a SIGSEGV within 30 seconds before the change, and nothing after it. A real agent with whodata on Ubuntu 24.04, built with ThreadSanitizer and driven by the load from the issue, reports the race between the scan thread and the whodata thread before the change, and neither that race nor a lock-order inversion after it.
- No unit test is added. The existing tests for `fim_whodata_event()` already declare the read lock, and a cmocka test cannot observe the defect this PR closes: it is a data race between two threads, so the only meaningful regression guard is the sanitizer run above. Pinning the order of the lock calls instead would mean adding `function_called()` to a shared database wrapper and rewriting the existing tests around it, for a check that would not have caught the original bug.

## Review Checklist

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux

  - [ ] Coverity
  - [ ] AddressSanitizer

**General Checklist:**

- [ ] Code changes reviewed


src/syscheckd/src/syscheck.c
```diff
@@ -319,9 +319,19 @@ int Start_win32_Syscheck() {
...
 #ifdef ENABLE_AUDIT
+static directory_t *fim_configuration_directory_ebpf(const char *path) {
+    directory_t *configuration;
+
+    w_rwlock_rdlock(&syscheck.directories_lock);
+    configuration = fim_configuration_directory(path);
+    w_rwlock_unlock(&syscheck.directories_lock);
+
+    return configuration;
+}
```
MarcelKemp: **[S2 · blocking · introduced]** `fim_configuration_directory_ebpf()` releases the read lock before returning the raw `directory_t *`, and its only consumer dereferences it unlocked — `handle_event()` reads `config->options` at `src/syscheckd/src/ebpf/src/ebpf_whodata.cpp:126` — so the use-after-free this PR targets is still live on the eBPF whodata path. The window is not rare: `update_wildcards_config()` calls `fim_insert_directory()` (`create_db.c:1829`), whose duplicate branch does `free_directory(dir_it); node_it->data = new_entry;` (`src/config/syscheck-config.c:163-164`) for every expanded wildcard entry on every scan pass, and it frees the removed entries at `create_db.c:1876` after releasing the write lock. The `create_db.c` half of this PR widens its bracket precisely so the pointer is *used* under the lock; this half does the opposite — pass the answer across the boundary, not the pointer, since `libfimebpf` is a separate shared library and cannot take `syscheck.directories_lock` itself.

```c
// src/syscheckd/src/syscheck.c — replaces the helper
static bool fim_whodata_active_ebpf(const char *path) {
    directory_t *configuration;
    bool active;

    w_rwlock_rdlock(&syscheck.directories_lock);
    configuration = fim_configuration_directory(path);
    active = (configuration != NULL) && (configuration->options & WHODATA_ACTIVE);
    w_rwlock_unlock(&syscheck.directories_lock);

    return active;
}
// ... passed as fimebpf_initialize(fim_whodata_active_ebpf, get_user, ...)

// src/syscheckd/src/ebpf/src/ebpf_whodata.cpp — handle_event(), lines 125-126
    if (confFn(e->filename))

// src/syscheckd/src/ebpf/include/ebpf_whodata.h    -> bool (*fim_conf)(const char *)
// src/syscheckd/src/ebpf/include/ebpf_whodata.hpp  -> using fim_configuration_directory_t = bool (*)(const char*);
// src/syscheckd/src/ebpf/tests/fimEbpfWhodataTest/ebpf_mock_utils.hpp -> mocks return bool
```

Not offered as a `suggestion` block because the fix needs the four companion edits above; applying only the `syscheck.c` part would not build.

src/syscheckd/src/syscheck.c
```diff
@@ -319,9 +319,19 @@ int Start_win32_Syscheck() {
 
 #ifdef __linux__
 #ifdef ENABLE_AUDIT
+static directory_t *fim_configuration_directory_ebpf(const char *path) {
+    directory_t *configuration;
+
+    w_rwlock_rdlock(&syscheck.directories_lock);
```
MarcelKemp: **[S3 · non-blocking · aggravated]** Taking the lock here puts the eBPF ring-buffer poller thread behind `directories_lock`, which `update_wildcards_config()` holds exclusively across `expand_wildcards()` (`create_db.c:1809`, filesystem globbing) and the audit-rule removals — `handle_event()` never blocked before this change. While it waits, `ring_buffer_poll()` is not re-entered, so the kernel ring buffer can overflow and drop whodata events with no log (`FIM_FULL_EBPF_KERNEL_QUEUE` covers only the userspace bounded queue). The lock itself is right and should stay; the fix belongs on the writer — run `expand_wildcards()` and build the new entries outside the bracket, then take the write lock only to splice the prepared list in.

src/syscheckd/src/syscheck.c
```diff
@@ -319,9 +319,21 @@ int Start_win32_Syscheck() {
...
 #ifdef __linux__
 #ifdef ENABLE_AUDIT
+static bool fim_whodata_active_ebpf(const char *path) {
+    directory_t *configuration;
+    bool active;
+
+    w_rwlock_rdlock(&syscheck.directories_lock);
+    configuration = fim_configuration_directory(path);
+    active = (configuration != NULL) && (configuration->options & WHODATA_ACTIVE);
+    w_rwlock_unlock(&syscheck.directories_lock);
```
MarcelKemp: **[S3 · non-blocking · aggravated]** Taking the lock here puts the eBPF ring-buffer poller thread behind `directories_lock`, which `update_wildcards_config()` holds exclusively across `expand_wildcards()` (`create_db.c:1809`, filesystem globbing) and the audit-rule removals — `handle_event()` never blocked before this change. While it waits, `ring_buffer_poll()` is not re-entered, so the kernel ring buffer can overflow and drop whodata events with no log (`FIM_FULL_EBPF_KERNEL_QUEUE` covers only the userspace bounded queue). The lock itself is right and should stay; the fix belongs on the writer — run `expand_wildcards()` and build the new entries outside the bracket, then take the write lock only to splice the prepared list in. Carried over from the previous review; unchanged by the callback-contract fix.

MarcelKemp (dismissed):

MarcelKemp (approved):

PR #39218, opened 2026-09-11
Surface three silent agent failure modes

### LINKED ISSUE #38859: Agent liveness and log-drop visibility: three findings from the 5.0.0 readiness testing
## Description

Three findings from #38549, filed together at the agent team's request. They share one operational shape: the agent stops delivering while everything an administrator can look at says it is healthy.

- **Finding 35** — `wazuh-agentd` keeps running after its transport refuses to start.
- **Finding 57** — the systemd unit reports `active` after `wazuh-agentd` has exited.
- **Finding 54** — logcollector reports a queue-full drop once per process and never counts it.

Code references are at `5.0.0` `ac166e355d`. Original reports: [35](https://github.com/wazuh/wazuh/issues/38549#issuecomment-5427750134), [54](https://github.com/wazuh/wazuh/issues/38549#issuecomment-5429128094), [57](https://github.com/wazuh/wazuh/issues/38549#issuecomment-5452275688).

## Environment

- **Findings 35 and 57:** 5.0.0 nightly packages, stage `beta5`, commit `6c2c981`, installed natively on Amazon Linux 2023 with the shipped installation assistant. All-in-one master plus a cluster worker, one 5.x agent (v5.0.0) and one 4.x agent (v4.14.0). No Docker.
- **Finding 54:** real 5.0.0 agent over HTTPS with the manager stopped for ~3 minutes; on recovery the engine ingress was replaced by a counting mock to tally exactly what the agent delivered.

---

## 1. `wazuh-agentd` keeps running with no transport (finding 35)

> **Agent team:** do we have a reproduced case? It would be good to see it.

`w_https_client_start()` is declared `void` (`src/client-agent/src/https_client_bridge.c:1794`) and `agentd.c:120` ignores the outcome: when the transport refuses to start, agentd continues into `start_agent()` and the monitoring loop. The process stays alive, with no transport and no retry, indefinitely.

Reproduced with an inverted backoff pair — `agent.https_backoff_base=2000` and `agent.https_backoff_cap=1000`, each value in range, the pair invalid — where the transport correctly fails closed:

```
ERROR: Config rejected: agent.https_backoff_base (2000) is above agent.https_backoff_cap (1000).
ERROR: https_client: failed to start (configuration rejected).
```

`systemctl is-active wazuh-agent` returned `active`, and the manager logged **0** `/control notify` lines in a 20 s window. Any liveness check built on unit state sees a healthy agent that is entirely mute.

### Steps to reproduce

---

## 2. The unit reports `active` after `wazuh-agentd` has died (finding 57)

> **Agent team:** can you test whether 4.13.x behaves the same? And 4.14.x? The change was applied between those versions, so if it happens in both we need to see whether it can be fixed without breaking other things.

`wazuh-agent.service` is `Type=forking` with no `Restart=` and no `PIDFile=` (`src/init/templates/wazuh-agent.service`). `wazuh-agentd` carries all HTTPS traffic — `/control`, `/stateful`, `/stateless`, `/download` and enrollment — but when it exits the sibling daemons keep the unit's main PID alive, so the unit keeps reporting `active`.

Reproduced while validating finding 4: an agent running with `enrollment_retry_max=0` was pushed into re-enrollment by SIGSTOPing authd and corrupting its key on the manager. `wazuh-agentd` hit `(2302) CRITICAL` and exited; `systemctl is-active wazuh-agent` still returned `active`, with the agent sending nothing.

This is the same operational failure as finding 35 from a different cause: there the transport refused to start, here the whole process is gone. Monitoring built on unit state, which is the obvious thing to build it on, sees neither.

### What changed between 4.13.x and 4.14.x -> preliminar 

`RemainAfterExit=yes` was removed from the unit by `496887d82e` ("fix: remove remain after exit", 2026-06-02). That commit is in 4.14.x and in 5.0.0, and is **not** in `v4.13.0` or `v4.13.1`:

- **4.13.x** ships `RemainAfterExit=yes`, so the unit reports `active` even when every process is gone.
- **4.14.x and 5.0.0** carry a byte-identical unit body, and `active` now tracks only the surviving main PID, which is exactly what this finding observes.

So the two versions report `active` for two different reasons, and dropping `RemainAfterExit` narrowed the problem without closing it. Confirming this by observation on a 4.13.x and a 4.14.x agent is still ours to run; the comparison above is from the unit files, not from a run.

---

## 3. A queue-full drop is reported once per process, and the producer pause never reaches logcollector (finding 54)

> **Agent team:** this is expected, since events have to be dropped when the queue fills up. 

Agreed on the policy: dropping when the queue is full is expected, and this finding does not ask to change it. What it reports is that the loss is not observable, and that the knob an operator would reach for does not act on it.

During a ~3 minute manager outage, 3,000 events (~20 eps) were appended to a monitored file. On recovery, events **1 to 2014 arrived contiguously and exactly once, and 2015 to 3000 were lost** (33%). The agent logged one line for the whole run:

```
wazuh-logcollector: WARNING: Target 'agent' message queue is full (1024). Log lines may be lost.
```

- The warning is latched in a `static` local, so it fires once per process and the latch is shared across targets (`src/logcollector/src/logcollector.c:1857`). The first full queue of the agent's life silences every later outage, on every target.
- Every later discard is `mdebug2("Discarding log line for target '%s'")` (`logcollector.c:1888`): invisible at the default log level, and never counted anywhere.
- `agent.https_producer_pause_threshold` does not propagate to logcollector, so moving it changes nothing for file sources. The loss point is upstream of the producer.
- The queue is tunable: `logcollector.queue_size`, shipped at 1024 (`etc/internal_options.conf:44`), maximum 220000. The observed arithmetic matches: producer in-flight buffer (~990) + queue (1024) = 2014.

With the defaults a fleet therefore tolerates roughly `(producer buffer + queue_size) / EPS` of manager outage before silently losing the tail: about 100 s at a modest 20 eps. A 25 s outage at low volume showed zero loss and zero duplicates, so the short-outage path is sound.

hernanvalenzuela:
## Description

Closes three findings from #38859 ("Agent liveness and log-drop visibility"), all sharing the same pattern: the agent fails in a way that never surfaces to systemctl, the logs, or the manager, so the operator sees a healthy agent that is not actually working.

## Proposed Changes

**Finding 35 — Silent transport failure**
`w_https_client_start()` returned `void`, so `agentd.c` had no way to know the HTTPS client failed to start (e.g. an inverted `https_backoff_base`/`https_backoff_cap` pair). The agent kept running with no transport and no way to ever notify the manager, while `systemctl is-active` reported `active`.
- Changed `w_https_client_start()` to return `bool`.
- `agentd.c` and `win_utils.c` (Windows startup path, same pattern) now call `merror_exit()` when it returns `false`, matching the other fatal startup guards already in `AgentdStart()`.

**Finding 57 — systemd unit stays "active" after `wazuh-agentd` dies**
`wazuh-agent.service` is `Type=forking` with no `PIDFile=`. Since `wazuh-control` forks several independent daemons, systemd falls back to "is the cgroup non-empty" to decide liveness — so if `wazuh-agentd` (the one daemon that owns connectivity) dies, the unit still reports `active` as long as any sibling daemon (execd, logcollector, ...) survives.
- `wazuh-agentd` now writes an additional, fixed-path PID file (`var/run/wazuh-agentd.pid`), since the existing PID-suffixed file can't back a static `PIDFile=`.
- Added `PIDFile=` to `wazuh-agent.service` pointing at that file, so systemd tracks this daemon specifically.
- `wazuh-control`'s `stop_service()` cleans up the new file alongside the existing one.

**Finding 54 — Logcollector silently discards lines during a manager outage**
When logcollector's per-target message queue fills up (`logcollector.queue_size`, default 1024), it discards the new line. The warning fired once per process; every subsequent drop logged only at `mdebug2` (invisible by default), so a multi-minute outage looked identical to a single dropped line in the logs.
- `w_msg_queue_push()` now re-arms the warning every 60s instead of once-ever, and reports how many lines were dropped since the last warning.
- The existing per-target `drops` counter (already accurate, already dumped to the logcollector state file) is unchanged — this only fixes the log-level visibility.

## Evidence

- Finding 35: unit tests updated to assert on the new return value (`test_https_client_bridge.c`); 98/98 + 18/18 existing tests pass.
- Finding 57: reproduced and validated against a real `systemd --user` unit mirroring the actual daemon fan-out — confirmed `active` before the fix and `failed` after, with sibling daemons left untouched (`KillMode=process`).
- Finding 54: reproduced and validated by extracting the real pre-fix/post-fix `w_msg_queue_push()` bodies into a standalone harness linked against the real `queue_op.c`, including a real 65s wait to exercise the actual 60s re-arm interval.

  - [ ] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...


cborla (approved):
LGTM
- [x] Code changes reviewed

PR #39219, opened 2026-09-11
Report Windows processes whose handle cannot be opened

### LINKED ISSUE #39199: Windows process inventory silently discards every process whose handle cannot be opened, so protected system processes including lsass.exe are absent and 72 of 78 records have a parent that is not in the inventory
## Description

The Windows process collector enumerates every process correctly and then throws away each one it cannot open a handle to, producing no record and no log line.

`fillProcessesData()` takes a full snapshot, which includes every process on the system:

```c
const auto processesSnapshot { CreateToolhelp32Snapshot(TH32CS_SNAPPROCESS, 0) };
...
if (Process32First(processesSnapshot, &processEntry))
{
    do
    {
        func(processEntry);
    }
    while (Process32Next(processesSnapshot, &processEntry));
}
```

`getProcessInfo()` then requires a handle with `PROCESS_VM_READ` before it will populate anything at all:

```c
static nlohmann::json getProcessInfo(const PROCESSENTRY32& processEntry)
{
    nlohmann::json jsProcessInfo{};
    const auto pId { processEntry.th32ProcessID };
    const auto processHandle { OpenProcess(PROCESS_QUERY_INFORMATION | PROCESS_VM_READ, FALSE, pId) };

    if (processHandle)
    {
        ... every field is set inside this block ...
        CloseHandle(processHandle);
    }

    return jsProcessInfo;
}
```

There is no `else`. When `OpenProcess` fails the function returns the empty object it started with, and the caller drops it without comment:

```c
void SysInfo::getProcessesInfo(std::function<void(nlohmann::json&)> callback) const
{
    fillProcessesData([&callback](const auto & processEntry)
    {
        auto processInfo = getProcessInfo(processEntry);

        if (!processInfo.empty())
        {
            callback(processInfo);
        }
    });
}
```

`PROCESS_VM_READ` is the demanding part of that request. It is needed only to read the command line out of the target's PEB. Windows refuses it for protected and kernel-backed processes even to a SYSTEM caller, which on a current Windows 11 install covers `System`, `Registry`, `Memory Compression`, `smss.exe`, `csrss.exe`, `wininit.exe`, `services.exe` and, where LSA protection is enabled, `lsass.exe`.

Everything needed for a useful record is already in hand without any handle. `PROCESSENTRY32` carries `th32ProcessID`, `th32ParentProcessID`, `szExeFile` and `cntThreads`. The collector discards all of it because one optional field is unavailable.

## Evidence

I can show processes are missing without relying on any assumption about which processes Windows runs, by using the inventory's own parent references. Of the 78 process documents the Windows agent reports, **72 name a parent PID that is not itself present in the inventory**:

```
  reported processes: 78

  parent PIDs referenced by reported processes but ABSENT from the inventory:
     ppid 864    is the parent of 69 reported process(es), e.g. [('svchost.exe', 63), ('spoolsv.exe', 1), ('amazon-ssm-agent.exe', 1)]
     ppid 1252   is the parent of  1 reported process(es), e.g. [('MicrosoftEdgeUpdate.exe', 1)]
     ppid  728   is the parent of  1 reported process(es), e.g. [('fontdrvhost.exe', 1)]
     ppid  712   is the parent of  1 reported process(es), e.g. [('winlogon.exe', 1)]

  72 of 78 reported processes have a parent that is not itself in the inventory.
```

The identities follow from the standard Windows process tree: the parent of 63 `svchost.exe` instances is `services.exe`, the parent of `winlogon.exe` is `wininit.exe`, and the parent of `fontdrvhost.exe` is `csrss.exe`. All three are exactly the processes `OpenProcess` with `PROCESS_VM_READ` is refused on.

The full set of distinct process names reported is 15, and 63 of the 78 records are `svchost.exe`:

```
  svchost.exe 63, fontdrvhost.exe 2, AggregatorHost.exe 1, LogonUI.exe 1,
  MicrosoftEdgeUpdate.exe 1, SearchIndexer.exe 1, TiWorker.exe 1, TrustedInstaller.exe 1,
  WmiPrvSE.exe 1, amazon-ssm-agent.exe 1, dwm.exe 1, spoolsv.exe 1, sshd.exe 1,
  wazuh-agent.exe 1, winlogon.exe 1
```

`lsass.exe`, `services.exe`, `csrss.exe`, `wininit.exe`, `smss.exe`, `System` and `Registry` do not appear. `explorer.exe` is also absent, but that one is legitimate: `LogonUI.exe` is running, so the machine is sitting at the logon screen with no interactive session.

## Limits of what I established

I could not open a shell on the Windows endpoint to run `tasklist` and diff it against the inventory, because that host is not reachable from here. The dangling-parent analysis above is derived entirely from the agent's own reported data and does not depend on host access, which is why I lead with it. The attribution of specific PIDs to specific executables, and the statement that `OpenProcess` is what fails, are inferences from the standard Windows process tree and from the code, not measurements on that machine.

## Environment

- Agent under test: `demo-env-windows-agent`, id `005`, Microsoft Windows 11 Pro for Workstations, version `10.0.26200.9168`, build `25H2`, amd64, 78 processes reported.
- `VERSION.json`: `{"version": "5.0.0", "stage": "rc1", "commit": "2161e69"}` on manager and agent, nightly `2026-09-09`.
- Source read from the `5.0.0` branch of `wazuh/wazuh`: `getProcessInfo()`, `fillProcessesData()` and `SysInfo::getProcessesInfo()` in `src/data_provider/src/sysInfoWin.cpp`.
- Comparison agents `001` to `004` (Linux) and `006` (macOS) report 103 to 486 processes each and do not show this pattern.

## Steps to reproduce

1. Enroll a 5.0.0 Windows agent and let a syscollector evaluation complete.
2. Pull every process document with its pid and parent pid:

```bash
  "https://<indexer>:9200/wazuh-states-inventory-processes/_search?size=200" \
  -d '{"query":{"term":{"wazuh.agent.id":"005"}},
       "_source":["process.pid","process.parent.pid","process.name"]}'
```

3. Collect the set of reported pids and the set of referenced parent pids, and subtract. Every parent that is not itself reported is a process the collector dropped. On the agent under test that is 4 distinct parents accounting for 72 of the 78 records.

4. Ask for the processes that should always exist:

```bash
for p in lsass.exe services.exe csrss.exe wininit.exe smss.exe; do
    "https://<indexer>:9200/wazuh-states-inventory-processes/_count" -H 'Content-Type: application/json' \
    -d "{\"query\":{\"bool\":{\"filter\":[{\"term\":{\"wazuh.agent.id\":\"005\"}},
         {\"term\":{\"process.name\":\"$p\"}}]}}}"
done
```

All return 0.

5. On the endpoint, compare against `tasklist` or `Get-Process`, and confirm the cause by attempting the same call the collector makes, for example with a small test that calls `OpenProcess(PROCESS_QUERY_INFORMATION | PROCESS_VM_READ, FALSE, pid)` against `lsass.exe` and prints `GetLastError()`.

## Expected

Every process in the snapshot appears in the inventory. Where a handle cannot be obtained, the record is still emitted with the fields the snapshot already provides, name, pid, parent pid and thread count, and the fields that genuinely require a handle are left absent. A failure to open a process is visible somewhere, at minimum as a debug line or a count.

## Actual

`getProcessInfo()` populates nothing unless `OpenProcess` with `PROCESS_QUERY_INFORMATION | PROCESS_VM_READ` succeeds, and the caller silently discards the resulting empty object. Protected and kernel-backed processes are therefore absent from the inventory with no indication. On the agent under test 78 processes are reported, 63 of them `svchost.exe`, `lsass.exe` and `services.exe` are missing, and 72 of the 78 records point at a parent that is not in the inventory.

## Not a duplicate of

`qa-known-issues.py check` found **no cache match** against 784 open bugs, the live search for `services.exe including inventory lsass.exe dangling silently` returned nothing, and the tool concluded "no known issue looks like this one".

Related but distinct: wazuh/wazuh#39126 reports that the macOS process collector truncates its enumeration by passing a count where a byte size is expected. Both end in an incomplete process inventory, but that is a buffer-sizing error on a different platform and this is an error-handling decision on Windows; neither fix helps the other.

anromerom:
## Description

Closes #39199

The Windows process collector enumerated every process correctly and then discarded each one it could not open a handle to, producing no record and no log line. Protected and kernel-backed processes were therefore absent from the inventory, including `lsass.exe`, `services.exe`, `csrss.exe`, `wininit.exe` and `smss.exe`, which left most of the remaining records pointing at a parent that was not itself present. The handle was requested with `PROCESS_VM_READ`, an access right Windows refuses on those processes and which no call in the collector needed: its only consumer was `GetProcessMemoryInfo`, whose two outputs were never read and have no column in the process table.

## Proposed Changes

**`src/data_provider/src/sysInfoWin.cpp`**: the Windows process collector

- Request `PROCESS_QUERY_LIMITED_INFORMATION` instead of `PROCESS_QUERY_INFORMATION | PROCESS_VM_READ`. It covers every call made on the handle, and Windows grants it on processes that refuse the wider right.
- Write `stime`, `utime` and `start` only when `GetProcessTimes` succeeded. The zero-initialised creation time otherwise underflows into a plausible-looking `2009-04-22T19:24:48.000Z`.
- Remove `setProcessMemInfo()` and the unused `pageFileUsage()`, `virtualSize()` and `sessionId()` getters, which were the only consumers of `PROCESS_VM_READ`.
- Release the process snapshot handle through a scope guard, so it is closed when the per-process callback throws instead of leaking one kernel handle per scan.

**`src/data_provider/src/utilsWrapperWin.cpp`, `.hpp`**: record assembly

- Build the record from the snapshot entry first and overlay the handle-derived fields on top, so every enumerated process reaches the inventory whether or not its handle can be opened. A process collected without a handle carries its name and pid, and its parent pid where the snapshot reports a usable one; the fields that genuinely require a handle are left absent.
- Omit `parent_pid` for pid 0, which the snapshot reports as its own parent. Emitting it would leave a self-referential edge that loops any parent-child walk of the inventory.

**`src/wazuh_modules/syscollector/src/syscollectorImp.cpp`**: scan reporting

- Log at debug level how many processes were reported without a start time, so a handle the operating system refuses stays visible instead of being indistinguishable from a process that legitimately has no command line. The counter stays at zero on Linux and macOS, whose collectors always set one.

The inventory gains documents that a Windows endpoint never reported before. Besides the protected executables named above, `System Idle Process` (`process.pid: 0`) and `System` (`process.pid: 4`) now appear, and the kernel-backed `Registry` and `Memory Compression` do too, so the process count per Windows agent rises by roughly a dozen. `System Idle Process` is an accounting bucket rather than a real process and is the one record with no `process.parent.pid`.

### Results and Evidence

Validated on a real Windows 11 endpoint (`10.0.26200.9168`, build 25H2) enrolled against a real 5.0.0 manager, not a simulator. The same physical machine was measured with the released agent and then with a package built from this branch, so the comparison holds the process set constant.

| Check | Result |
|---|---|
| Protected and kernel-backed processes present | Pass, all nine |
| Recovered processes carry the full field set | Pass, `lsass.exe` has its command line and times |
| Normal process unaffected by the narrower mask | Pass, `wazuh-agent.exe` unchanged |
| Pid 0 emitted without a self-referential parent | Pass |
| No live process left dangling as a parent | Pass |
| Automated tests | Pass, 62 of 62 CI checks |

<details>
<summary>Protected and kernel-backed processes are reported</summary>

```sql
process               before   after
lsass.exe                  0       1
services.exe               0       1
csrss.exe                  0       2
wininit.exe                0       1
smss.exe                   0       1
System                     0       1
System Idle Process        0       1
Registry                   0       1
Memory Compression         0       1
```

`lsass.exe` is recovered with the full field set rather than as a partial record, which is what narrowing the access mask buys over emitting a snapshot-only record:

```sql
name           lsass.exe
pid            1704
command_line   C:\WINDOWS\system32\lsass.exe
utime          2
stime          3
start          2026-09-14T12:58:04.000Z
parent.pid     1532
```

- Patched agent packages inventory on protected processes:
image

</details>

<details>
<summary>A normal process is unaffected by the narrower access mask</summary>

This is the regression case for the mask change. Every field a normal process carried before is still populated:

```sql
name           wazuh-agent.exe
pid            7824
command_line   "C:\Program Files (x86)\ossec-agent\wazuh-agent.exe"
args_count     0
utime          1
stime          1
start          2026-09-14T13:58:41.000Z
parent.pid     1652
```

`args` is null and `args_count` is 0 because this command line carries no arguments.

</details>

<details>
<summary>Pid 0 is emitted without a parent</summary>

```sql
name           System Idle Process
pid            0
command_line   null
args           null
args_count     null
utime          null
stime          null
start          null
parent.pid     null
```

The handle-derived fields are absent and there is no `parent.pid`, so the record does not form the self-referential edge that would loop a parent-child walk of the inventory.

image

</details>

<details>
<summary>Parent references</summary>

| | Before | After |
|---|---|---|
| Reported processes | 262 | 265 |
| Records with no `parent_pid` | 0 | 1 |
| Distinct dangling parents | 9 | 6 |
| Records whose parent is absent | 86 of 262 | 8 of 265 |

The remaining dangling parents are processes that have genuinely exited, not processes the collector dropped, and the count cannot reach zero on a live system because Windows does not reparent orphans. `ppid 1248` and `ppid 1524` were session `smss.exe` instances, which exit once their session is created, and `ppid 10628` was `userinit.exe`, which exits after launching the shell.

The meaningful criterion is that no live process is left dangling. Before the fix `ppid 1640` was the parent of 77 records and was the running `services.exe`. After the fix `services.exe` is present as pid 1652 and resolves, as do `wininit.exe` 1532 and `smss.exe` 912.

</details>

<details>
<summary>Before, for reference: the unpatched agent on the same machine</summary>

Every protected and kernel-backed process missing:

```sql
lsass.exe 0    services.exe 0    csrss.exe 0    wininit.exe 0    smss.exe 0
System 0       System Idle Process 0            Registry 0       Memory Compression 0
```

Parent references pointing at processes that are not in the inventory:

```sql
reported processes: 262
distinct dangling parents: 9
records whose parent is absent: 86 of 262
   ppid 1640   parent of  77, e.g. [('svchost.exe', 42), ('cowork-svc.exe', 1), ('NTKDaemon.exe', 1)]
   ppid 1520   parent of   2, e.g. [('LsaIso.exe', 1), ('fontdrvhost.exe', 1)]
   ppid 1512   parent of   1, e.g. [('winlogon.exe', 1)]
```

`ppid 1640` is the parent of 42 `svchost.exe` instances, so it is `services.exe`. `ppid 1512` is the parent of `winlogon.exe`, so it is `wininit.exe`. `ppid 1520` is the parent of `LsaIso.exe` and `fontdrvhost.exe`, so it is `csrss.exe`. None of the three is in the inventory.

</details>

<details>
<summary>Automated tests</summary>

All CI checks pass, 62 of 62. `rtr (data_provider, winagent)` builds and runs the Windows unit test target that carries the new tests, and `rtr (wazuh_modules/syscollector, winagent)` covers the counter change.

</details>

### Scope and Compatibility

- Thread count is available from the snapshot without a handle and is deliberately not emitted. There is no column for it in `dbsync_processes` and no ECS mapping, so it would be dropped downstream. Adding it is a separate change.
- Short-lived processes that exit between the snapshot and `OpenProcess` now produce a record where they previously produced none, so a scan may show a create and a delete for the same pid. Worth knowing when comparing inventory deltas.
- The same defect is present verbatim on `main` and on `4.14.9`, where `sysInfoWin.cpp` carries the identical `OpenProcess` call and the identical empty-record filter. A forward-port to `main` is needed or the bug returns on the 5.x development line, and 4.x needs its own decision; the surrounding field set differs on 4.x, so that port is not a cherry-pick.

### Artifacts Affected

- `wazuh-agent` on Windows, through the `sysinfo` data provider and the syscollector module.
- No behavioural change on Linux or macOS. The syscollector module is built on every platform, but the collector path this fixes is Windows-only and the new debug counter stays at zero elsewhere because those collectors always set a process start time.

### Configuration Changes

- N/A

### Documentation Updates

- N/A

### Tests Introduced

- `src/data_provider/tests/sysInfoWin/sysInfoWin_test.cpp`, eight unit tests on the `sysInfoWindows_unit_test` target. The whole target passes, 33 of 33.
- `BuildProcessSnapshotRecordNormalEntry`, `BuildProcessSnapshotRecordSystemIdleProcess`, `BuildProcessSnapshotRecordSystemProcess` and `BuildProcessSnapshotRecordNonAsciiName` cover the record built from the snapshot entry alone, including the pid 0 and pid 4 naming and the ANSI to UTF-8 conversion.
- `BuildProcessRecordWithoutHandleFieldsKeepsSnapshotFields`, `BuildProcessRecordOverlaysHandleFields`, `BuildProcessRecordWithDefaultConstructedHandleFields` and `BuildProcessRecordWithPartialHandleFields` cover the merge. The first fails if record construction is made conditional on the handle-derived fields being present, which is the defect this PR fixes; the third pins the representation the collector actually passes when no handle opened.
- `syscollectorimp_unit_test` passes, 101 of 101, covering the platform-neutral counter change.
- `OpenProcess` itself is not mocked, so a reintroduction of the fault inside `SysInfo::getProcessesInfo` is not covered. Doing so would require linking the whole `sysinfo` library into the Windows test target, which is larger than this change.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/src/data_provider/src/sysInfoWin.cpp
+++ b/src/data_provider/src/sysInfoWin.cpp
@@ -57,8 +57,6 @@
 
 constexpr auto CENTRAL_PROCESSOR_REGISTRY {"HARDWARE\\DESCRIPTION\\System\\CentralProcessor\\0"};
 const std::string UNINSTALL_REGISTRY{"SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall"};
-constexpr auto SYSTEM_IDLE_PROCESS_NAME {"System Idle Process"};
-constexpr auto SYSTEM_PROCESS_NAME {"System"};
 
 
 static const std::map<std::string, DWORD> gs_firmwareTableProviderSignature
@@ -73,17 +71,14 @@ static constexpr auto ProcessCommandLineInformation = static_cast<PROCESSINFOCLA
 class SysInfoProcess final
 {
     public:
-        SysInfoProcess(const DWORD pId, const HANDLE processHandle)
-            : m_pId{ pId },
-              m_hProcess{ processHandle },
+        explicit SysInfoProcess(const HANDLE processHandle)
+            : m_hProcess{ processHandle },
               m_creationTime{},
               m_kernelModeTime{},
               m_userModeTime{},
-              m_pageFileUsage{},
-              m_virtualSize{}
+              m_timesValid{ false }
         {
             setProcessTimes();
-            setProcessMemInfo();
         }
 
         ~SysInfoProcess() = default;
@@ -152,26 +147,11 @@ class SysInfoProcess final
             return m_userModeTime.QuadPart;
         }
 
-        DWORD pageFileUsage() const
+        // creationTime(), kernelModeTime() and userModeTime() are meaningless unless this returns
+        // true: a zero creation time underflows into a plausible-looking 2009-04-22 timestamp.
+        bool timesValid() const
         {
-            return m_pageFileUsage;
-        }
-
-        DWORD virtualSize() const
-        {
-            return m_virtualSize;
-        }
-
-        DWORD sessionId() const
-        {
-            DWORD ret{};
-
-            if (!ProcessIdToSessionId(m_pId, &ret))
-            {
-                // Unable to retrieve session ID from current process.
-            }
-
-            return ret;
+            return m_timesValid;
         }
 
     private:
@@ -202,24 +182,8 @@ class SysInfoProcess final
                 m_creationTime.HighPart = lpCreationTime.dwHighDateTime;
                 m_creationTime.QuadPart /= TO_SECONDS_VALUE;
 
+                m_timesValid = true;
             }
-
-            // else: Unable to retrieve kernel mode and user mode times from current process.
-        }
-
-        void setProcessMemInfo()
-        {
-            PROCESS_MEMORY_COUNTERS pMemCounters{};
-
-            // Get page file usage and virtual size
-            // Reference: https://stackoverflow.com/a/1986486
-            if (GetProcessMemoryInfo(m_hProcess, &pMemCounters, sizeof(pMemCounters)))
-            {
-                m_pageFileUsage = pMemCounters.PagefileUsage;
-                m_virtualSize   = pMemCounters.WorkingSetSize + pMemCounters.PagefileUsage;
-            }
-
-            // else: Unable to retrieve page file usage from current process
         }
 
         static SystemDrivesMap getNtWin32DrivesMap()
@@ -339,61 +303,41 @@ class SysInfoProcess final
             }
         }
 
-        const DWORD     m_pId;
         HANDLE          m_hProcess;
         ULARGE_INTEGER  m_creationTime;
         ULARGE_INTEGER  m_kernelModeTime;
         ULARGE_INTEGER  m_userModeTime;
-        DWORD           m_pageFileUsage;
-        DWORD           m_virtualSize;
+        bool            m_timesValid;
 };
 
 
-static bool isSystemProcess(const DWORD pid)
+static nlohmann::json getProcessHandleFields(const PROCESSENTRY32& processEntry)
 {
-    return pid == 0 || pid == 4;
-}
-
-static std::string processName(const PROCESSENTRY32& processEntry)
-{
-    std::string ret;
-    const DWORD pId { processEntry.th32ProcessID };
-
-    if (isSystemProcess(pId))
-    {
-        ret = (pId == 0) ? SYSTEM_IDLE_PROCESS_NAME : SYSTEM_PROCESS_NAME;
-    }
-    else
-    {
-        ret = processEntry.szExeFile;
-    }
-
-    return ret;
-}
-
-static nlohmann::json getProcessInfo(const PROCESSENTRY32& processEntry)
-{
-    nlohmann::json jsProcessInfo{};
+    // An empty object, never a null: this is what buildProcessRecord() overlays, and it is
+    // returned as is whenever no handle can be opened.
+    auto jsHandleFields = nlohmann::json::object();
     const auto pId { processEntry.th32ProcessID };
-    const auto processHandle { OpenProcess(PROCESS_QUERY_INFORMATION | PROCESS_VM_READ, FALSE, pId) };
+
+    // PROCESS_QUERY_LIMITED_INFORMATION covers every call made on the handle below, and
+    // Windows grants it on processes that refuse PROCESS_QUERY_INFORMATION, such as csrss.exe.
+    const auto processHandle { OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, FALSE, pId) };
 
     if (processHandle)
     {
-        SysInfoProcess process(pId, processHandle);
+        SysInfoProcess process(processHandle);
 
-        // Current process information
-        jsProcessInfo["name"]         = Utils::EncodingWindowsHelper::stringAnsiToStringUTF8(processName(processEntry));
-        jsProcessInfo["stime"]        = process.kernelModeTime();
-        jsProcessInfo["parent_pid"]   = processEntry.th32ParentProcessID;
-        jsProcessInfo["pid"]          = std::to_string(pId);
-        jsProcessInfo["utime"]        = process.userModeTime();
-        jsProcessInfo["start"]        = Utils::rawTimestampToISO8601(static_cast<uint32_t>(process.creationTime()));
+        if (process.timesValid())
+        {
+            jsHandleFields["stime"]        = process.kernelModeTime();
+            jsHandleFields["utime"]        = process.userModeTime();
+            jsHandleFields["start"]        = Utils::rawTimestampToISO8601(static_cast<uint32_t>(process.creationTime()));
+        }
 
         if (isSystemProcess(pId))
         {
-            jsProcessInfo["command_line"]    = "none";
-            jsProcessInfo["args"]  = "";
-            jsProcessInfo["args_count"]  = 0;
+            jsHandleFields["command_line"]    = "none";
+            jsHandleFields["args"]  = "";
+            jsHandleFields["args_count"]  = 0;
         }
         else
         {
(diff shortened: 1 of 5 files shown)


cborla (approved):
LGTM
- [x] Code changes reviewed

Issue #38695, opened 2026-08-28
After an unclean stop, the manager's startup kills a PID number read from a stale file, which is one of its own new processes, and the container comes up with no Wazuh running at all


Rebits:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0-latest | wazuh-manager startup (`bin/wazuh-manager-apid`, PID handling in `var/run`) | Manager | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 (agents), two-node cluster |

## Description

The defect is not that PID files survive an unclean stop, which is normal. It is what the startup does with them: it kills the number rather than the process, on the assumption that a PID recorded on disk still refers to the daemon that wrote it. After a restart it does not, and in a container it is very likely to refer to something the startup itself has just created.

## Steps to reproduce

```bash
# reproduccion determinista: plantar un pid rancio que colisione
docker exec <manager> sh -c "echo 161 > /var/wazuh-manager/var/run/wazuh-manager-apid_auth-161.pid"
docker kill <manager>
docker start <manager>
sleep 110

# el contenedor esta "Up" y no hay nada dentro
docker ps --filter name=<manager>
docker exec <manager> ps -eo comm
docker logs --tail 6 <manager>
# -> wazuh-manager-apid did not start correctly.

# y se arregla asi
docker exec <manager> sh -c "rm -f /var/wazuh-manager/var/run/*.pid /var/wazuh-manager/var/run/.api_operation_lock"
docker exec <manager> /var/wazuh-manager/bin/wazuh-manager-control start
```

The natural path is simply to `docker kill` the manager while its daemons hold low PIDs, which is the normal state of a container that has not been restarted since it was created.

## Expected results

- **Validate a PID before killing it.** A PID file records a number, not an identity. Check that the process still exists and that it is the daemon the file claims, by comparing the executable or the start time, before sending a signal. Killing an unverified number is how a startup ends up killing its own child.
- **Clear `var/run` on start.** A manager that is starting knows no daemon of its own is running yet, so every PID file present is stale by definition. `.api_operation_lock` was more than a day old and had also survived.
- **Do not abort the whole startup because one daemon failed**, or if aborting is deliberate, exit non-zero so the container dies and an orchestrator restarts it. Reporting `Up` with no Wazuh running is the worst of both.

### How it was found

Both cluster nodes were SIGKILLed and brought back, the worker first. The master container came up and, two minutes later:

```
[...]

PR #39230, opened 2026-09-14
Reclaim the manager start lock when it has no pid file

### LINKED ISSUE #39114: Manager container enters an unrecoverable restart loop when start-script-lock is left without its pid file
## Description

`wazuh-manager-control` guards startup with a mkdir-based mutex:

```sh
LOCK="${DIR}/var/start-script-lock"
LOCK_PID="${LOCK}/pid"
...
mkdir ${LOCK} > /dev/null 2>&1
MSL=$?
if [ "${MSL}" = "0" ]; then
    # Lock acquired (setting the pid)
    echo "$$" > ${LOCK_PID}
    return;
fi
```

The lock directory is created first and the pid file is written afterwards, so the two are not atomic. If the process dies between those statements, or the pid file is otherwise lost, the lock directory is left behind **empty**.

The stale-lock recovery cannot reclaim it, because it is gated on reading that pid file:

```sh
pid=$(cat ${LOCK_PID} 2>/dev/null)

if [ $? = 0 ]
then
    kill -0 ${pid} >/dev/null 2>&1
    if [ ! $? = 0 ]; then
        # Pid is not present. Unlocking and executing
        unlock; mkdir ${LOCK}; echo "$$" > ${LOCK_PID}; return;
    fi
fi
```

When `${LOCK_PID}` does not exist, `cat` fails, `$?` is non-zero, and the whole recovery branch is skipped. After `MAX_ITERATION` attempts the script gives up:

```
ERROR: Another instance is locking this process.
If you are sure that no other instance is running, please remove /var/wazuh-manager/var/start-script-lock
```

In the container this becomes a permanent loop. The manager cannot start, so the image healthcheck (`wazuh-manager-control status`) fails with the same lock error and times out. The entrypoint watchdog then reaches:

```sh
if _manager_unhealthy; then
    echo "Wazuh Manager is not running as expected, exiting so the container can be restarted..."
    _stop 1
fi
```

`restart: always` restarts the container, and because the lock lives on the container's own filesystem (it is under `/var/wazuh-manager/var/`, which is not one of the named volumes) it survives the restart. The loop never terminates and the Manager API stays unreachable until somebody removes the directory by hand.

## Environment

- Topology: clustered managers (`wazuh.master` master, `wazuh.worker` worker), single indexer (`wazuh1.indexer`), single dashboard, from `wazuh-docker` branch `5.0.0` (`ad231ed`), multi-node adapted to one indexer node.
- Images:
  - `public.ecr.aws/wazuh-cicd/wazuh/wazuh-manager:5.0.0-latest` `sha256:10219f399ab4d49ef990453dc8ab10f61f8d2dcb4504c83932f4802c422c18ce` (created 2026-09-10T03:24:54Z)
  - `public.ecr.aws/wazuh-cicd/wazuh/wazuh-indexer:5.0.0-latest` `sha256:9618a53cd049feae909c71792506c807c3a10806b7e3675fa9ca024d5b214e82`
  - `public.ecr.aws/wazuh-cicd/wazuh/wazuh-dashboard:5.0.0-latest` `sha256:bca8c4591aa8535ba0db0131ba4fea55da0037bb14a15cc27ed43508d95bca7b`
- Healthcheck as shipped: `CMD /var/wazuh-manager/bin/wazuh-manager-control status`, interval 15s, timeout 5s, retries 5, start period 60s.
- Host: Linux 6.12.10, Docker 29.4.0, Docker Compose v5.1.3.

## Steps to reproduce

The loop is entered whenever the manager exits leaving an empty lock directory. It can be reproduced deterministically by creating that state directly:

1. Deploy the multi-node stack and wait for `wazuh.master` to be healthy.
2. Create the empty lock the unclean exit leaves behind, and restart the container:

   ```bash
   docker exec multi-node-wazuh.master mkdir -p /var/wazuh-manager/var/start-script-lock
   docker compose restart wazuh.master
   ```
3. Watch the container. It never becomes healthy, and `RestartCount` keeps climbing:

   ```bash
   docker inspect multi-node-wazuh.master --format '{{.RestartCount}} {{.State.Health.Status}}'
   docker logs multi-node-wazuh.master 2>&1 | grep -E 'Another instance|not running as expected'
   ```
4. Confirm the lock is the sole cause by removing it:

   ```bash
   docker exec multi-node-wazuh.master rm -rf /var/wazuh-manager/var/start-script-lock
   docker compose restart wazuh.master
   ```

The container returns to healthy and the API answers within about 20 seconds.

In this session the state was reached without any injection: the manager exited on its own after a run of WazuhDB query timeouts and inventory-sync 503s, and never came back.

## Expected

A lock left over from a previous run is reclaimed on the next start. A lock directory with no pid file is by definition stale, since no live process ever recorded ownership of it, so `lock()` should treat it as stale and take it, rather than failing in a way only manual intervention can clear.

## Actual

The master ran normally for about 20 minutes, then exited and never recovered. `RestartCount` climbed from 3 to 5 over roughly 25 minutes while the API stayed unreachable:

```
  t+30s  restarts=3 health=starting api=000
  t+90s  restarts=4 health=starting api=000
  t+180s restarts=4 health=starting api=000
```

Container log, repeating every cycle:

```
Wazuh Manager is not running as expected, exiting so the container can be restarted...
Stopping Wazuh Manager...
ERROR: Another instance is locking this process.
If you are sure that no other instance is running, please remove /var/wazuh-manager/var/start-script-lock
```

The lock directory was present and empty, with no `pid` file in it:

```
$ docker exec multi-node-wazuh.master ls -la /var/wazuh-manager/var/start-script-lock/
total 12
drwxr-xr-x 2 root root          4096 Sep 10 10:38 .
drwxr-x--- 1 root wazuh-manager 4096 Sep 10 10:38 ..
```

Healthcheck log for the same period:

```
exit=-1 Health check exceeded timeout (5s)   (x5)
```

After `rm -rf` of that directory and a restart:

```
  t+20s restarts=0 health=healthy api=200
```

## Note for triage

Two components meet here and the fix may want both. The `lock()` staleness logic is in `wazuh-manager-control` (wazuh/wazuh, `src/init`), which is where the unreclaimable lock comes from. The exit-and-be-restarted behaviour is `entrypoint.sh` in wazuh-docker, which turns a failed start into an endless loop rather than surfacing the reason. Filed against wazuh/wazuh because the lock is the root cause; reassign if the team prefers to fix the container side.

## Environment caveat, stated plainly

The first unclean exit followed WazuhDB query timeouts (`deadline=2000 ms`) and repeated `inventory sync server answered HTTP 503` warnings, on a host whose disk was 92% full at the time and where the indexer was consuming heavy I/O. That disk pressure may well have contributed to the *initial* exit. It does not explain the loop: disk had recovered to 87% (55G free) while the container was still looping, and the loop ended only when the lock directory was removed.

Antoniogm03:
## Description

`wazuh-manager-control`'s `lock()` guards manager start/stop with a `mkdir`-based mutex. Creating the lock directory and writing the owner's pid into it are two separate steps, not atomic. If the process dies between them, the lock directory is left behind empty, with no pid file inside.

The stale-lock recovery only runs when it can read a pid from that file (`cat ${LOCK_PID}`). When the file doesn't exist, `cat` fails and the entire recovery branch is skipped — the lock is never reclaimed. After `MAX_ITERATION` retries the script gives up with `ERROR: Another instance is locking this process.`, and both `start` and `status` (the latter used by the container healthcheck) fail the same way until someone removes the directory by hand.

Closes #39114

## Proposed Changes

- `src/init/wazuh-server.sh`: in `lock()`, a missing pid file now takes the same reclaim path as a dead one — `kill -0` on an empty pid fails exactly like it does on a dead pid, so the two cases no longer need separate branches. A lock with no pid file only reclaims once that state has held for a few *consecutive* retry rounds (~3s); a lock with a dead pid reclaims on the first round it is observed dead. Either way, the caller only writes its own pid after winning the recreation `mkdir`.

The retry-round requirement exists because an earlier version of this fix reclaimed on the very first empty-pid observation, which raced with a live acquisition: the lock owner writes its pid in a separate statement right after `mkdir`, so a lock mid-acquisition also looks pid-less for a brief moment. A competitor could then `rm -rf` a lock a live process still believed it owned, leaving two callers both convinced they hold it. Requiring the pid-less state to survive a few rounds before reclaiming closes that window.

That requirement was first implemented by gating on the caller's total round counter (`$i`, incremented every loop iteration regardless of what was observed), reused from the caller's own overall retry count. That undercounts nothing about the *caller* but measures the wrong thing: a caller queued behind a different, live owner for a while already has `$i` past the gate by the time that owner releases, so if a fresh, legitimate acquirer then wins the next `mkdir` and stalls before writing its pid, the queued caller's very first empty-pid observation reclaims immediately — the same failure this guard exists to prevent, just from a different starting point. Fixed by tracking a separate counter that only counts consecutive rounds since the owner was last seen alive, reset to 0 the moment it is; a dead pid still reclaims on the first round (it already proves a prior run finished its own mkdir+write, no ambiguity), so this also undoes the earlier version's unrelated slowdown of that case.

Also fixed as a side effect: `mkdir` failing for a reason other than "already exists" (read-only or full filesystem, insufficient permissions) no longer lets `lock()` return as if it held the lock when it never actually created the directory — the recreation `mkdir`'s own result now gates the pid write and the `return` in every case.

- `src/init/wazuh-server.sh`: reclaiming a stale lock now takes `mkdir` on a second marker (`${LOCK}.reclaim`) before touching `${LOCK}` at all, and re-checks the pid immediately after winning it. `unlock()` followed by `mkdir` was not an atomic pair: two callers that both concluded the same lock was stale could each reach that code at a different time, and each would unconditionally unlock whatever was currently there. If caller A finished its own reclaim first, caller B's later `unlock()` destroyed A's freshly recreated, legitimate lock and recreated it again for itself — both callers would return believing they held it. A rename-based claim on `${LOCK}` itself does not fix this (tested and ruled out below): `mv` only gives exclusivity between callers racing at the same instant, and once A recreates `${LOCK}` under the same path, B's later `mv` against that path succeeds again, stealing the fresh lock exactly like `unlock()` did. Serializing the reclaim itself, rather than the recreation, closes it: only one caller can ever hold `${LOCK}.reclaim`, so only one caller is ever deciding whether to unlock and recreate `${LOCK}`.

- `src/init/wazuh-server.sh`: the reclaim marker itself now self-heals if its holder dies while holding it. `${LOCK}.reclaim` had no release path besides its own two `rmdir` calls — no trap, no TTL, nothing in `unlock()`. A caller killed between winning it and releasing it (a container watchdog's SIGKILL, a healthcheck timeout, systemd's `TimeoutSec`, an operator's Ctrl-C, or even this repo's own `test_pid_collision_recovery.py`, which kills a `start` subprocess after a 60s wait) leaves it behind forever: every later `mkdir` on it fails, so this reclaim path is permanently disabled — the exact failure this fix exists to recover from, reintroduced one level down with no recovery of its own. A caller that keeps failing to win the marker for several consecutive rounds now removes it, since it is only ever held across a handful of local, non-blocking statements. Gating that on `$i` or `unreachable` (both counting rounds this caller has waited overall) would repeat the mistake just fixed for the lock itself: a caller already past those gates would treat a marker another caller only just grabbed as leaked, on the very first round it finds it busy — reopening the race between two concurrent reclaimers the marker exists to close (verified this false positive in isolation before ruling it out). A separate counter, counting only this caller's own consecutive failures to acquire the marker and reset to 0 the moment it succeeds, avoids it.

- `src/init/wazuh-server.sh`: dropped code that no system in 5.0.0's support matrix (Amazon Linux 2023, Ubuntu 24.04/26.04, RHEL 9/10, all with systemd well above 237 and cgroup v2) can reach. `is_systemd()`, the systemd-version detection in the `reload` action, `add_to_cgroup()` (writes to a cgroup v1 path), and the `systemd-run` launch branch in `start_service()` formed a compatibility path for systemd ≤ 237; the detection only ever ran inside `reload`, so `start`/`restart` never exercised it even on a system that needed it, and under the current matrix the version check never trips at all. Also removed `checkpid()`, whose loop iterated over `${CDAEMONS}`, a variable never assigned anywhere in this file's history (back to when it was `ossec-server.sh`) — `pstatus()`, called per daemon a few lines later in `status()`/`start_service()`/`stop_service()`, already does the same stale-pidfile cleanup for real. Fixed a comment stating the lock loop retries "10 times", stale since `MAX_ITERATION` was raised to 40, and added `reload` to `help()`'s usage line (it was already a valid, undocumented action). No 4.x-to-5.x upgrade compatibility code was found in this file to remove; that logic, where it exists, lives in `install.sh`/`update.sh`, not here.

- `src/init/wazuh-server.sh`: general cleanup pass unrelated to the lock, found while reviewing the file end to end.
  - `disable()` carried a "kill the daemon" block guarded by `[ "$daemon" != '' ]`; `daemon` was initialized to `''` and never reassigned, so the block was unreachable — a leftover from a version of `disable()` that supported more than `debug`, unchanged since the `ossec-server.sh` rename. Removed.
  - Every branch of the final action dispatch (`case "$action"`) now sets `RETVAL` explicitly on its success path. Previously only `status` set it; every other action's exit code was whatever `unlock`'s own `rm -rf` happened to return, not a deliberate value.
  - `enable`/`disable` now take the real argument as `$1`. They were called as `enable $action $arg` so the actual value landed in `$2`, relying on an unused filler argument that only worked because nothing simplified the call.
  - The default `case` branch now prints `Invalid action: <name>` before `help()`, matching `info()`'s existing behavior for a bad flag.
  - Replaced an `ls`+`$?` file-existence check with `[ -f ]`, and every `"X$var" = "Xvalue"` idiom with direct `[ -z "$var" ]` / `[ "$var" = "value" ]` quoting (`enable`, `disable`, `pstatus`, `info`, and the `wazuh-manager-apid`/`authd` checks in `status()`/`start_service()`).
  - Dropped `daemon_name` and `MSL`, two variables that only ever aliased a value already in scope (`$i`/`$pfile`, and `$?` right after `lock()`'s first `mkdir`) for a single, immediate use — the same functions already used the original value directly elsewhere in the same block.

  No functional change intended in any of the above; see Results and Evidence.
### Results and Evidence

Reproduced and verified in a manager install running inside the project's devcontainer.

**Pre-fix** — creating an empty `start-script-lock` directory (no `pid` file) and starting the manager:

```
$ mkdir -p /var/wazuh-manager/var/start-script-lock
$ /var/wazuh-manager/bin/wazuh-manager-control start
ERROR: Another instance is locking this process.
If you are sure that no other instance is running, please remove /var/wazuh-manager/var/start-script-lock
exit code: 1, elapsed: 41s
```

No daemon started, and the lock directory was left untouched. `wazuh-manager-control status` (the container healthcheck command) failed with the same error.

**Post-fix** — same reproduction, same empty lock directory:

```
$ mkdir -p /var/wazuh-manager/var/start-script-lock
$ /var/wazuh-manager/bin/wazuh-manager-control start
Starting Wazuh v5.0.0...
Started wazuh-manager-apid...
Started wazuh-manager-db...
exit code: 0, elapsed: 5s
```

The lock is reclaimed automatically and the manager starts.

**mkdir failing for a reason other than EEXIST** — verified in isolation (same `lock()`/`unlock()`, `LOCK` pointing at a nonexistent parent directory so every `mkdir` fails with `ENOENT`): the function never returns as if it held the lock; it correctly exhausts its retries and exits with the defined error instead.

**Race scenario** — a lock left without a pid file that gets one written ~1.5s later (simulating a live acquisition caught mid-flight, the case the retry-round requirement protects):

```
$ mkdir -p /var/wazuh-manager/var/start-script-lock
$ ( sleep 1.5; sleep 300 & echo $! > /var/wazuh-manager/var/start-script-lock/pid ) &
$ timeout 5 /var/wazuh-manager/bin/wazuh-manager-control start; echo exit=$?
exit=124
$ cat /var/wazuh-manager/var/start-script-lock/pid
19403
$ /var/wazuh-manager/bin/wazuh-manager-control status
ERROR: Another instance is locking this process.
```

`start` keeps waiting instead of reclaiming, the lock still carries the real owner's pid, and `status` correctly reports it as locked.

**Regression checks**:
- Lock with a `pid` file pointing to a dead process: reclaimed in ~3s (down from ~6s once the consecutive-rounds counter stopped making it wait behind an unrelated gate).
- Lock with a `pid` file pointing to a live process: unaffected, `start` still blocks and `status` still reports `ERROR: Another instance is locking this process.` after exhausting retries.

**Queued-caller race** — the case the consecutive-rounds counter protects, found after the round counter above shipped as a total-rounds count. Isolated harness: an owner holds the lock live for 6s while a second caller queues behind it (its round counter reaches 6); the owner releases; a third, fresh caller wins the next `mkdir` and stalls 0.3s before writing its pid. With the total-rounds counter, the queued caller reclaimed immediately, racing the third caller's own write — both callers exited `lock()` believing they owned it. With the consecutive-rounds counter, the queued caller correctly waits and backs off; the fresh caller's write completes first.

**Two-caller reclaim race** — the case the reclaim marker protects. First, an isolated harness (same `lock()`/`unlock()` logic, no daemons) with two reclaimers racing on the same stale lock:
- Staggered (one finishes, the other acts ~0.3s later): exactly one winner.
- Simultaneous: exactly one winner.
- Three-way simultaneous: exactly one winner.
- Solo, uncontended: unaffected.

The naive `mv`-based claim (tried first, not shipped) failed this exact test — the staggered case produced two winners, because a later `mv` against the *same path* succeeds again once it has been recreated, same as `unlock()` did.

Then against the real installed manager, two concurrent `start` invocations on the same stale lock (dead pid):
```
$ mkdir -p /var/wazuh-manager/var/start-script-lock; echo 999999 > /var/wazuh-manager/var/start-script-lock/pid
$ ( /var/wazuh-manager/bin/wazuh-manager-control start; echo "B1 exit=$?" ) &
$ ( /var/wazuh-manager/bin/wazuh-manager-control start; echo "B2 exit=$?" ) &
$ wait
```
One invocation started the daemons normally; the other's `lock()` found the lock legitimately held by then and its own `start` logic fell back to "already running" instead of restarting anything. No duplicate daemon processes were spawned.

**Leaked reclaim marker** — against the real installed manager, with `${LOCK}` empty (the exact state from #39114) and `${LOCK}.reclaim` already present (simulating a caller killed while holding it):
```
$ mkdir -p /var/wazuh-manager/var/start-script-lock /var/wazuh-manager/var/start-script-lock.reclaim
$ timeout 60 /var/wazuh-manager/bin/wazuh-manager-control start
```
Before this fix: exhausts `MAX_ITERATION` and exits with `ERROR: Another instance is locking this process.` after 41s, every time, until the marker is removed by hand — this reclaim path was permanently disabled. After: recovers in 10s.

**Marker false-positive** — the case the separate consecutive-failures counter avoids, checked in isolation before shipping: a caller with an already-elevated round count (simulating one that had queued behind a live owner) encountering a marker a different, live reclaimer had held for only a moment. Gating the marker's self-heal on that caller's total round count removed the live reclaimer's marker on the very first round it was found busy; gating it on a dedicated counter of this caller's own consecutive failures to acquire the marker did not.

**No behavior change on the normal path** — the full diff against the pre-fix file touches only the branch that runs when `mkdir ${LOCK}` fails; the "lock free, no contention" path (`mkdir` succeeds, write pid, return) is untouched. Confirmed live with an A/B run of the same operator lifecycle against the original binary and the fixed one:

```
$ for CONTROL in .../wazuh-manager-control.bak-preFix .../wazuh-manager-control; do
    $CONTROL start; $CONTROL status; $CONTROL stop; $CONTROL status; $CONTROL restart; $CONTROL status
  done
```

Output identical line for line between the two runs (pids and sub-second timings aside, which vary run to run regardless of the fix). Re-ran the full regression set (orphaned lock, dead pid, live pid, mid-acquisition race, leaked marker, two concurrent `start`s) once more as a final consolidated pass after all of the above: same results as reported in each case above, no drift.

**Verified again after rebasing onto the current `5.0.0`** (45 commits landed upstream since the branch point, none touching this file): `sh -n` passes on the rebased script, and the orphaned-lock scenario still reclaims in 5s against the redeployed binary — same as before the rebase.

**Dead-code removal (legacy systemd path, `checkpid()`)** — verified no observable behavior change against the real installed manager: a full `start` → `status` → `reload` → `status` → `stop` → `status` → `help` cycle produced the same sequence of daemon start/stop messages as before, `reload` still excludes `wazuh-manager-remoted` from its cycle, `reload` is now listed in `help()`'s usage line, and the full lock regression set (orphaned lock, dead pid, leaked reclaim marker) was re-run and unaffected.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...


src/init/wazuh-server.sh
```diff
@@ -91,6 +91,13 @@ lock()
                 echo "$$" > ${LOCK_PID}
                 return;
             fi
+        else
+            # Lock dir with no pid file: no process ever recorded
+            # ownership of it, so it can't belong to a live run.
+            unlock;
+            mkdir ${LOCK} > /dev/null 2>&1
+            echo "$$" > ${LOCK_PID}
+            return;
```
Antoniogm03: **[S2 · blocking · introduced]**

This branch declares the lock stale after a single check one second old, but the owner creates the directory (line 70) and writes its pid (line 74) in two separate statements, so a live run legitimately has a pid-less lock for a moment — the added comment's premise does not hold for exactly the non-atomic window this PR exists because of. A contender that samples that gap runs `unlock` (`rm -rf`) on a running instance's lock and returns owning it while the real owner writes its pid back and also returns, so `start_service` and the healthcheck's `status` proceed concurrently and the first to finish unlocks the other's lock; before this change the gap was self-healing, since the failed `cat` just made the waiter loop until the pid file appeared. A genuine orphan never grows a pid file, so requiring the state to persist a few rounds closes the window at a cost of ~3s on the recovery path (what the PR already measures for the dead-pid case, well inside `MAX_ITERATION=40` and the unit's `TimeoutSec=45`), and gating on `mkdir` additionally stops the branch claiming ownership when `mkdir` failed for a reason other than `EEXIST` — on a read-only or full `${DIR}/var` it currently returns holding nothing at all.

```suggestion
        else
            # A lock dir with no pid file is stale only if it stays that way: the
            # owner writes the pid right after mkdir, so a single check can catch a
            # live run mid-acquisition. Reclaim once the gap outlives a few rounds.
            if [ "$i" -gt 2 ]; then
                unlock;
                if mkdir ${LOCK} > /dev/null 2>&1; then
                    echo "$$" > ${LOCK_PID}
                    return;
                fi
            fi
```

src/init/wazuh-server.sh
```diff
@@ -80,16 +80,39 @@ lock()
...
+        # kill -0 on an empty pid (no pid file) fails just like it does on a
+        # dead one, so a missing pid file takes the same path as a dead pid.
+        kill -0 ${pid} >/dev/null 2>&1
+        if [ ! $? = 0 ]; then
+            # The owner is unreachable. A dead pid proves a prior run
+            # completed its mkdir+pid write; a lock with no pid file yet
+            # could just be one doing that right now (mkdir and the pid
+            # write are two statements), so only reclaim once that has
+            # held for a few rounds.
+            if [ "$i" -gt 2 ]; then
```
Antoniogm03: **[S2 · blocking · introduced]** `[ "$i" -gt 2 ]` counts rounds *this caller* has waited, not consecutive rounds the lock has been pid-less, so any caller that was already queued behind a different, live owner passes the gate on its very first pid-less observation — the behaviour the commit message says this removes. Reproduced with `lock()` extracted verbatim into dash: with a waiter at `i=6`, a fresh legitimate `mkdir ${LOCK}` at t=6.4173 was `rm -rf`'d and reacquired by that waiter at t=6.4369, 19 ms later, and both callers returned believing they held the lock. Counting *consecutive* unreachable rounds fixes it, and requiring that persistence only when no pid was recorded also keeps the dead-pid case reclaiming on the first round (see the comment on line 92).

```suggestion
        kill -0 ${pid} >/dev/null 2>&1
        if [ $? = 0 ]; then
            # Owner alive: earlier unreachable rounds no longer count.
            unreachable=0
        else
            # A dead pid proves a prior run finished its mkdir+pid write, so
            # it is stale right away. A lock with no pid file may be one
            # doing that write right now, so reclaim it only once
            # consecutive rounds still find none: "$i" counts rounds this
            # caller has waited, which is already past any gate for a caller
            # that was queued behind a different, live owner.
            unreachable=`expr ${unreachable:-0} + 1`
            if [ -n "${pid}" ] || [ "${unreachable}" -gt 2 ]; then
```

Antoniogm03: **[S2 · blocking · introduced]** Putting the dead-pid case behind the same wait regresses a path that works today: a recorded-but-dead pid already proves the previous owner finished its `mkdir` + pid write, so there is no mid-acquisition ambiguity to wait out. The cost is measured in the commit body as ~3s → ~6s and weighed against `MAX_ITERATION=40` and the unit's `TimeoutSec=45`, but those bound `start`; `status` is invoked by the container healthcheck, which per #39114 kills it at 5s — so a dead-pid lock that `status` reclaims today is never reclaimed by it again, and the container stays unhealthy while a healthy manager gets restarted. Reclaiming immediately when a pid was recorded (`[ -n "${pid}" ] || [ "${unreachable}" -gt 2 ]`, already included in the suggestion on line 85) keeps the single reclaim path and restores the current latency.

src/init/wazuh-server.sh
```diff
@@ -80,16 +80,39 @@ lock()
...
+                    if [ ! $? = 0 ]; then
+                        unlock;
+                        if mkdir ${LOCK} > /dev/null 2>&1; then
+                            echo "$$" > ${LOCK_PID}
+                            rmdir "${LOCK}.reclaim" > /dev/null 2>&1
+                            return;
+                        fi
+                    fi
+                    rmdir "${LOCK}.reclaim" > /dev/null 2>&1
+                fi
```
Antoniogm03: **[S2 · blocking · introduced]** `${LOCK}.reclaim` has no release path other than these two `rmdir` calls — no trap, no TTL, and `unlock()` removes only `${LOCK}` — so a process killed between the `mkdir` on line 100 and here leaks the marker and silently disables this reclaim forever. Reproduced: with an empty `${LOCK}` (the exact #39114 state) plus a leftover marker, `lock()` burns all 40 rounds and exits 1, and the error on line 122 still tells the operator to remove only `${LOCK}`; the kill is not hypothetical, `tests/integration/test_manager_control/test_pid_collision_recovery.py:232` does `start.kill()` on a running `start`. The marker is only held across a handful of syscalls, so one still present several rounds later can be cleared safely.

```suggestion
                    rmdir "${LOCK}.reclaim" > /dev/null 2>&1
                elif [ "$i" -gt 4 ]; then
                    # The marker is only held across a handful of syscalls, so
                    # one still present rounds later was left by a caller that
                    # was killed while holding it. Nothing else ever removes
                    # it, and leaving it disables this reclaim for good.
                    rmdir "${LOCK}.reclaim" > /dev/null 2>&1
                fi
```

src/init/wazuh-server.sh
```diff
@@ -80,16 +80,39 @@ lock()
...
+            # held for a few rounds.
+            if [ "$i" -gt 2 ]; then
+                # Serialize the reclaim itself: mkdir on a second marker
+                # is exclusive the same way ${LOCK}'s own mkdir is, so
+                # only one caller at a time can be deciding whether to
+                # unlock and recreate ${LOCK}. Without this, two callers
+                # could both pass the check above at different times and
+                # each unlock what the other had already recreated,
+                # believing they both hold it.
+                if mkdir "${LOCK}.reclaim" > /dev/null 2>&1; then
```
Antoniogm03: **[S3 · non-blocking · introduced]** The root cause is that acquisition publishes the lock (`mkdir ${LOCK}`, line 70) and its owner (`echo "$$" > ${LOCK_PID}`, line 74) in two statements; the round counter, this marker and the re-check all exist to tolerate that window rather than close it, which is why each one brought its own defect. `( set -C; echo $$ > "${LOCK}" ) 2>/dev/null` opens with `O_EXCL`, so the lock never exists without an owner: the pid-less case, the multi-round wait, the marker and its leak all disappear, leaving only the dead-pid check that was already there. Your call whether that belongs in this PR — it turns the lock from a directory into a file and drags `unlock()`, the upgrade path over existing installs and the twin in `wazuh-client.sh` with it, and `O_EXCL` over old NFS is the one thing the `mkdir` idiom historically bought.

src/init/wazuh-server.sh
```diff
@@ -80,20 +63,46 @@ lock()
...
+                        fi
+                    fi
+                    rmdir "${LOCK}.reclaim" > /dev/null 2>&1
+                else
+                    # The marker is held only briefly; one still busy after
+                    # several consecutive rounds was left by a dead caller.
+                    marker_busy=`expr ${marker_busy} + 1`
+                    if [ "${marker_busy}" -gt 4 ]; then
+                        rmdir "${LOCK}.reclaim" > /dev/null 2>&1
+                    fi
```
jotacarma90: **[S2 · blocking · introduced]** `marker_busy` is reset only inside the `mkdir "${LOCK}.reclaim"` success branch above, never after this forced `rmdir`, so once it passes 5 it stays there and every later busy round removes the marker on the first sight of it. Simulated the loop in `dash`: rounds 5, 6, 7 all fire the removal, so a reclaimer that wins the marker a moment after a competitor's self-heal has its live marker deleted — a third caller can then enter the critical section while the first is between its `unlock` and its `mkdir ${LOCK}`, and both leave `lock()` believing they hold it. This is the false positive the description says was ruled out, reached from the other side.

```suggestion
                    if [ "${marker_busy}" -gt 4 ]; then
                        rmdir "${LOCK}.reclaim" > /dev/null 2>&1
                        marker_busy=0
                    fi
```

src/init/wazuh-server.sh
```diff
@@ -80,20 +63,46 @@ lock()
...
-                return;
+        # An empty pid (no pid file) fails kill -0 just like a dead one.
+        kill -0 ${pid} >/dev/null 2>&1
+        if [ "$?" = "0" ]; then
+            unreachable=0
+        else
+            # "$i" counts total rounds waited, not consecutive unreachable
+            # ones; a caller queued behind a live owner would otherwise
+            # inherit an already-open gate the moment that owner is gone.
+            unreachable=`expr ${unreachable} + 1`
```
jotacarma90: **[S2 · blocking · introduced]** `unreachable` is reset only when `kill -0 ${pid}` succeeds (line 69), so rounds spent watching a *dead* pid count towards the gate on line 77 that exists to protect a *pid-less* one — the description's "a lock with no pid file only reclaims once that state has held for a few consecutive rounds" is not what the code enforces. Concretely: with `${LOCK}` orphaned and `${LOCK}.reclaim` leaked, a caller reaches `unreachable=8` while fighting for the marker, and the moment it wins it a legitimate acquirer that took `${LOCK}` at line 55 and has not written its pid yet is read as stale, `unlock`ed and recreated — two holders. The `rpid` re-check on lines 83-85 does not help because it does not re-apply the gate.

```suggestion
            # Only consecutive pid-less rounds may open the gate below: a
            # round spent watching a dead pid says nothing about whether a
            # lock is mid-acquisition, so it must not count towards it.
            if [ -n "${pid}" ]; then
                unreachable=0
            else
                unreachable=`expr ${unreachable} + 1`
            fi
```

src/init/wazuh-server.sh
```diff
@@ -45,31 +44,15 @@ MAX_ITERATION="40"
...
-            fi
-        done
-    done
-}
-
 lock()
 {
     i=0;
+    unreachable=0;
+    marker_busy=0;
```
jotacarma90: **[S2 · blocking · introduced]** These two counters, `${LOCK}.reclaim` and its self-healing timer ship with no automated test, while `tests/integration/test_manager_control/` already starts and kills `wazuh-manager-control` end to end for the comparable lifecycle bug #38695. Both defects flagged above are deterministic given a fixed sequence, so a test of that same shape would have caught them; without one, any later edit to `lock()` regresses silently. The checklist item "Tests cover the new functionality" is unticked.

```python
# tests/integration/test_manager_control/test_start_lock_reclaim.py
LOCK = Path(WAZUH_PATH, 'var', 'start-script-lock')

@pytest.mark.parametrize('leaked_marker', [False, True])
def test_orphan_lock_is_reclaimed(leaked_marker):
    """An empty lock dir (no pid file) is stale and must be reclaimed. Regression for #39114."""
    LOCK.mkdir(parents=True, exist_ok=True)
    if leaked_marker:
        Path(f'{LOCK}.reclaim').mkdir(exist_ok=True)
    result = subprocess.run([WAZUH_CONTROL_PATH, 'status'], capture_output=True, text=True, timeout=40)
    assert 'Another instance is locking this process' not in result.stdout
    assert not Path(f'{LOCK}.reclaim').exists()


def test_lock_mid_acquisition_is_not_stolen():
    """A lock whose owner has not written its pid yet is live, not stale."""
    LOCK.mkdir(parents=True, exist_ok=True)
    owner = subprocess.Popen(['sleep', '300'])
    threading.Timer(4.0, lambda: Path(LOCK, 'pid').write_text(f'{owner.pid}\n')).start()
    result = subprocess.run([WAZUH_CONTROL_PATH, 'status'], capture_output=True, text=True, timeout=60)
    assert Path(LOCK, 'pid').read_text().strip() == str(owner.pid)
    assert 'Another instance is locking this process' in result.stdout
```

src/init/wazuh-server.sh
```diff
@@ -45,31 +44,15 @@ MAX_ITERATION="40"
...
 
     # Providing a lock.
     while [ 1 ]; do
-        mkdir ${LOCK} > /dev/null 2>&1
-        MSL=$?
-        if [ "${MSL}" = "0" ]; then
+        if mkdir ${LOCK} > /dev/null 2>&1; then
             # Lock acquired (setting the pid)
             echo "$$" > ${LOCK_PID}
             return;
```
jotacarma90: **[S3 · non-blocking · aggravated]** The pid write is unchecked here and at line 88, so on ENOSPC or over quota the caller returns into `start_service` holding a lock that has no pid file while its owner is alive. That state used to be a harmless wedge; with the reclaim path this PR adds, a competitor now declares it stale three rounds later and takes it. `/var` filling up is the same condition #39114 was reproduced under (disk at 92%).

```suggestion
        if mkdir ${LOCK} > /dev/null 2>&1; then
            # Lock acquired (setting the pid). A lock whose pid could not be
            # recorded is reclaimable by any competitor, so do not hold it.
            if echo "$$" > ${LOCK_PID}; then
                return;
            fi
            unlock;
```

src/init/wazuh-server.sh
```diff
@@ -80,20 +63,46 @@ lock()
...
-            kill -0 ${pid} >/dev/null 2>&1
-            if [ ! $? = 0 ]; then
-                # Pid is not present.
-                # Unlocking and executing
-                unlock;
-                mkdir ${LOCK} > /dev/null 2>&1
-                echo "$$" > ${LOCK_PID}
-                return;
+        # An empty pid (no pid file) fails kill -0 just like a dead one.
+        kill -0 ${pid} >/dev/null 2>&1
```
jotacarma90: **[S3 · non-blocking · introduced]** The comment is right that an empty pid fails `kill -0` (verified: `rc=2` in `dash`, `rc=1` in `/bin/sh`), but the pid is otherwise unvalidated, and `kill -0 0` and `kill -0 -1` both return 0 — they signal the process group and everything signallable, not an owner. A corrupt `${LOCK}/pid` holding either value resets `unreachable` every round, so the reclaim branch is never reached and `start`/`status` wedge exactly as in #39114 with this fix in place. Worth closing given the whole PR is about recovering from an incomplete lock state.

```suggestion
        # An empty pid (no pid file) fails kill -0 just like a dead one, and
        # anything that is not a positive integer must not be signalled: 0 and
        # -1 succeed against the process group instead of a real owner.
        case "${pid}" in ''|*[!0-9]*) pid= ;; esac
        kill -0 "${pid}" >/dev/null 2>&1
```

src/init/wazuh-server.sh
jotacarma90: **[S3 · non-blocking · introduced]** `lock()` now carries three round counters that all advance on the same one-second tick and are read by three different gates, and two of the three have the defects flagged above. `${LOCK}`'s own mtime already separates the two states the counters are trying to tell apart — it is stamped at `mkdir` and again when `pid` is created inside it, so "no pid file and `${LOCK}` older than N seconds" is an orphan with no ambiguity, no second lock and no timer. If the current shape stays, a `trap 'rmdir "${LOCK}.reclaim" 2>/dev/null' EXIT INT TERM` set when the marker is won collapses its two release sites into one and gives it a real release path on signal death, which is the case `marker_busy` was invented to work around.

src/init/wazuh-server.sh
```diff
@@ -143,15 +152,14 @@ enable()
...
+    if [ -z "$1" ]; then
         echo ""
         echo "Disable options: debug"
         echo "Usage: $0 disable debug]"
         exit 1;
     fi
-    daemon=''
 
-    if [ "X$2" = "Xdebug" ]; then
+    if [ "$1" = "debug" ]; then
```
MiguelazoDS, line 167: I think there's no issue for this change, but probably beyond the scope. This was used to prevent breaking comparison if the variable does not exist.  No changes required but it would be good to validate if this does not break for old systems.

MiguelazoDS (commented):
The issue is easily reproduced with the steps in the issue 

```console
0 starting
Wazuh Manager is not running as expected, exiting so the container can be restarted...
ERROR: Another instance is locking this process.
Wazuh Manager is not running as expected, exiting so the container can be restarted...
...
1 starting
Wazuh Manager is not running as expected, exiting so the container can be restarted...
ERROR: Another instance is locking this process.
Wazuh Manager is not running as expected, exiting so the container can be restarted...
ERROR: Another instance is locking this process.
```

Image

After the fix

image

src/init/wazuh-server.sh
```diff
@@ -656,25 +599,23 @@ start)
...
     ;;
 reload)
     DAEMONS=$(echo $DAEMONS | sed 's/wazuh-manager-remoted//')
     SDAEMONS=$(echo $DAEMONS | awk '{ for (i=NF; i>1; i--) printf("%s ",$i); print $1; }')
-    if is_systemd; then
-        SYSTEMD_VERSION=$(systemctl --version | awk 'NR==1 {print $2}')
-        if [ "$SYSTEMD_VERSION" -le 237 ]; then
-            LEGACY_SYSTEMD_VERSION=1
-        fi
-    fi
```
MiguelazoDS, line 676: Are we sure that all systems are above 237? Doing a quick search I see that Centos 7 uses 219.

src/init/wazuh-server.sh
jotacarma90: **[S2 · blocking · introduced]** `marker_busy` is reset only inside the `mkdir "${LOCK}.reclaim"` success branch above, never after this forced `rmdir`, so once it passes 5 it stays there and every later busy round removes the marker on the first sight of it. Simulated the loop in `dash`: rounds 5, 6, 7 all fire the removal, so a reclaimer that wins the marker a moment after a competitor's self-heal has its live marker deleted — a third caller can then enter the critical section while the first is between its `unlock` and its `mkdir ${LOCK}`, and both leave `lock()` believing they hold it. This is the false positive the description says was ruled out, reached from the other side.

```suggestion
                    if [ "${marker_busy}" -gt 4 ]; then
                        rmdir "${LOCK}.reclaim" > /dev/null 2>&1
                        marker_busy=0
                    fi
```

src/init/wazuh-server.sh
jotacarma90: **[S2 · blocking · introduced]** `unreachable` is reset only when `kill -0 ${pid}` succeeds (line 69), so rounds spent watching a *dead* pid count towards the gate on line 77 that exists to protect a *pid-less* one — the description's "a lock with no pid file only reclaims once that state has held for a few consecutive rounds" is not what the code enforces. Concretely: with `${LOCK}` orphaned and `${LOCK}.reclaim` leaked, a caller reaches `unreachable=8` while fighting for the marker, and the moment it wins it a legitimate acquirer that took `${LOCK}` at line 55 and has not written its pid yet is read as stale, `unlock`ed and recreated — two holders. The `rpid` re-check on lines 83-85 does not help because it does not re-apply the gate.

```suggestion
            # Only consecutive pid-less rounds may open the gate below: a
            # round spent watching a dead pid says nothing about whether a
            # lock is mid-acquisition, so it must not count towards it.
            if [ -n "${pid}" ]; then
                unreachable=0
            else
                unreachable=`expr ${unreachable} + 1`
            fi
```

src/init/wazuh-server.sh
jotacarma90: **[S2 · blocking · introduced]** These two counters, `${LOCK}.reclaim` and its self-healing timer ship with no automated test, while `tests/integration/test_manager_control/` already starts and kills `wazuh-manager-control` end to end for the comparable lifecycle bug #38695. Both defects flagged above are deterministic given a fixed sequence, so a test of that same shape would have caught them; without one, any later edit to `lock()` regresses silently. The checklist item "Tests cover the new functionality" is unticked.

```python
# tests/integration/test_manager_control/test_start_lock_reclaim.py
LOCK = Path(WAZUH_PATH, 'var', 'start-script-lock')

@pytest.mark.parametrize('leaked_marker', [False, True])
def test_orphan_lock_is_reclaimed(leaked_marker):
    """An empty lock dir (no pid file) is stale and must be reclaimed. Regression for #39114."""
    LOCK.mkdir(parents=True, exist_ok=True)
    if leaked_marker:
        Path(f'{LOCK}.reclaim').mkdir(exist_ok=True)
    result = subprocess.run([WAZUH_CONTROL_PATH, 'status'], capture_output=True, text=True, timeout=40)
    assert 'Another instance is locking this process' not in result.stdout
    assert not Path(f'{LOCK}.reclaim').exists()


def test_lock_mid_acquisition_is_not_stolen():
    """A lock whose owner has not written its pid yet is live, not stale."""
    LOCK.mkdir(parents=True, exist_ok=True)
    owner = subprocess.Popen(['sleep', '300'])
    threading.Timer(4.0, lambda: Path(LOCK, 'pid').write_text(f'{owner.pid}\n')).start()
    result = subprocess.run([WAZUH_CONTROL_PATH, 'status'], capture_output=True, text=True, timeout=60)
    assert Path(LOCK, 'pid').read_text().strip() == str(owner.pid)
    assert 'Another instance is locking this process' in result.stdout
```

src/init/wazuh-server.sh
```diff
@@ -181,16 +181,14 @@ status()
...
 
     if [ $USE_JSON = true ]; then
         echo -n '{"error":0,"data":['
     fi
     for i in ${DAEMONS}; do
         ## The API daemon only runs on the master node
-        if [ X"$i" = "Xwazuh-manager-apid" ] && [ "$node_type" != "master" ]; then
+        if [ "$i" = "wazuh-manager-apid" ] && [ "$node_type" != "master" ]; then
             continue
         fi
```
jotacarma90: **[S2 · blocking · pre-existing]** `status()` excludes `wazuh-manager-apid` on non-master nodes but not `wazuh-manager-authd` when `auth.disabled` is `true`, while `start_service()` skips authd in that case (line 365). With that setting `status` prints `wazuh-manager-authd not running...` and exits 1 on a healthy manager. The wazuh-docker entrypoint works around it by filtering `authd` out of the output and documents it as a bug in this script; the wazuh-kubernetes probes do not, so a deployment with `auth.disabled: true` fails its liveness probe and restart-loops, which is the same failure mode this PR fixes for the lock. Since this PR already touches these lines and `status` is the container healthcheck, please mirror the exclusion here.

```suggestion
        if [ "$i" = "wazuh-manager-apid" ] && [ "$node_type" != "master" ]; then
            continue
        fi
        ## Mirror start_service(): authd is not started when auth.disabled is true.
        if [ "$i" = "wazuh-manager-authd" ] && [ "$(${MCONF} get auth.disabled 2>/dev/null)" = "true" ]; then
            continue
        fi
```

MiguelazoDS:
- New testing, replacing the script in an already deployed environment. :green_circle: 

Doing to consecutives checks the error is not reproduced, it returns "healthy" and the "starting" does not increment as observed here https://github.com/wazuh/wazuh/pull/39230#pullrequestreview-5240079943

```console
bash 1-force_error.sh
[+] restart 0/1
 ⠦ Container multi-node-wazuh.master Restarting                                                                                                            5.6s


bash 2-watch.sh
0 starting
ERROR: Another instance is locking this process.
Wazuh Manager is not running as expected, exiting so the container can be restarted...
ERROR: Another instance is locking this process.
ERROR: Another instance is locking this process.
Wazuh Manager is not running as expected, exiting so the container can be restarted...


bash 2-watch.sh
0 healthy
ERROR: Another instance is locking this process.
Wazuh Manager is not running as expected, exiting so the container can be restarted...
ERROR: Another instance is locking this process.
ERROR: Another instance is locking this process.
Wazuh Manager is not running as expected, exiting so the container can be restarted...
```

- Authd disabled verification :green_circle: 

Normal execution

```console
docker exec -it multi-node-wazuh.master bash
bash-5.2# /var/wazuh-manager/bin/wazuh-manager-control restart
Killing wazuh-manager-clusterd...
Killing wazuh-manager-modulesd...
Killing wazuh-manager-remoted...
Killing wazuh-manager-analysisd...
Killing wazuh-manager-db...
Killing wazuh-manager-authd...
Killing wazuh-manager-apid...
Wazuh v5.0.0 Stopped
Starting Wazuh v5.0.0...
Started wazuh-manager-apid...
Started wazuh-manager-db...
Completed.
```

Disabling authd

```console
bash-5.2# grep -A 2 "Configuration for wazuh-auth" /var/wazuh-manager/etc/wazuh-manager.conf
  <!-- Configuration for wazuh-authd -->
  <auth>
    <disabled>yes</disabled>
bash-5.2#
```

Restarting 

```console
bash-5.2# /var/wazuh-manager/bin/wazuh-manager-control restart
Killing wazuh-manager-clusterd...
Killing wazuh-manager-modulesd...
Killing wazuh-manager-remoted...
Killing wazuh-manager-analysisd...
Killing wazuh-manager-db...
Killing wazuh-manager-authd...
Killing wazuh-manager-apid...
Wazuh v5.0.0 Stopped
Starting Wazuh v5.0.0...
Started wazuh-manager-apid...
Started wazuh-manager-db...
Completed.
```

```console
bash-5.2# ps -aux | grep wazuh-a
root        7538  0.0  0.0   3496  1768 pts/0    S+   16:24   0:00 grep wazuh-a
bash-5.2#
```

MiguelazoDS (approved):

Issue #38956, opened 2026-09-04
Integration tests: logcollector, aws and msgraph suites blocked by the agent startup gate (missing TLS verification opt-out)


jpcerrone:
### Description

The `logcollector`, `aws`/`aws_inspector` and `msgraph` Python integration test suites fail wholesale on `5.0.0`, on every platform they run on. The tests are not at fault individually: the agent's startup gate never opens, so `wazuh-logcollector` and `wazuh-modulesd` never get past their startup block and none of the log lines the tests wait for are ever emitted.

This is the residue of `913a11736c` ("fix: opt integration test fixtures out of TLS verification"), which added the required opt-out to `agentd`, `enrollment`, `fim`, `sca`, `syscollector`, `execd`, `github` and `office365`, but not to these three suites.

### Evidence

| Suite | Platform | Result | Job |
|---|---|---|---|
| logcollector | Linux | 47 failed, 6 passed, 20 errors | [job 101076283207](https://github.com/wazuh/wazuh/actions/runs/33878550078/job/101076283207) |
| logcollector | macOS | 38 failed, 6 passed, 9 errors | [job 101065146001](https://github.com/wazuh/wazuh/actions/runs/33878550068/job/101065146001) |
| logcollector | Windows | 40 failed, 2 passed, 17 errors | [job 101047666173](https://github.com/wazuh/wazuh/actions/runs/33878550201/job/101047666173) |
| aws | Linux | 92 failed | [job 98512183575](https://github.com/wazuh/wazuh/actions/runs/33068298224/job/98512183575) |
| aws_inspector | Linux | 2 failed | [job 98512183675](https://github.com/wazuh/wazuh/actions/runs/33068298224/job/98512183675) |
| msgraph | Linux | 10 failed | [job 98503863161](https://github.com/wazuh/wazuh/actions/runs/33068298025/job/98503863161) |

The failures are uniform per suite — `The expected analyzing file log has not been produced`, `Error logcollector start event not detected`, `The AWS module did not start as expected`, `Error module enabled event not detected`.

`ossec.log` from the Linux logcollector run shows the daemon stopping at the gate and never resuming:

```sql
2026/09/04 14:22:17 wazuh-logcollector[11743] startup_gate_op.c:154 at startup_gate_wait_for_ready(): DEBUG: Startup hash gate is blocking 'wazuh-logcollector' (waiting for agentd startup gate status).
2026/09/04 14:22:25 wazuh-agentd: INFO: No authentication password provided
[...]

Issue #37, opened 2016-07-13
syntax error in host-deny.sh


fgth:
Hi, not sure that it's the right section for this issue, sorry if not. I've installed rpm packages from "WAZUH OSSEC Repository" on some centOS7 servers and found a problem with host-deny.sh.
At lines 114 and 117, whitespaces around the "=" sign after TMP_FILE cause a bash error ("command not found"). Removing whitespaces solves the problem.
Regards.

vikman90:
Hi @fgth. 

This is a known issue that has been fixed at OSSEC HIDS v2.9-beta06: https://github.com/ossec/ossec-hids/commit/c97f566e415756b17a3093bb10dbb3b8d405f284. We'll update the packages on the next stable release.

On the other hand, Wazuh HIDS v1.1.1 has also this problem fixed and it's available in the same repository.

Best regards.

Issue #38863, opened 2026-09-03
Update MITRE ATT&CK data and Manager integration to the latest version


TomasTurina:
## Description

The MITRE ATT&CK Enterprise dataset currently bundled with Wazuh Manager 5.0.0 is outdated and contains information that is no longer aligned with the latest MITRE ATT&CK release.

Wazuh currently uses the `enterprise-attack.json` dataset as the source for generating the MITRE database during installation. The resulting database is consumed by the Wazuh Manager/API and exposed through the MITRE API endpoints, which are then used by the Wazuh Dashboard to display MITRE ATT&CK information to users.

This issue aims to update the MITRE ATT&CK dataset to the latest available Enterprise release and ensure that the complete data-generation and consumption pipeline remains compatible with the new dataset.

The scope must include:

* Updating `ruleset/mitre/enterprise-attack.json` to the latest MITRE ATT&CK Enterprise dataset.
* Reviewing and updating the tools under `tools/mitre` responsible for processing the dataset and generating the MITRE database.
* Verifying the generated database schema and contents against the new ATT&CK data.
* Reviewing the Manager/API MITRE implementation for assumptions that may no longer be valid with the updated dataset.
* Fixing any API endpoints, queries, filtering, pagination, serialization, or other Manager-side issues exposed by the updated data.
* Verifying that the resulting data can be consumed correctly by the Dashboard.
* Updating tests and adding coverage for any new or changed behavior.

The MITRE ATT&CK dataset is generated into the database **at installation time**. Existing Wazuh Manager installations do not need to be migrated or have their existing MITRE databases updated. Backward compatibility or an in-place database migration mechanism is therefore out of scope.

The update should use the latest official MITRE ATT&CK Enterprise dataset available at implementation time. MITRE publishes the Enterprise dataset as STIX JSON data, with the current release available from its official ATT&CK data repository.

### MITRE dataset

* [ ] Identify the latest stable MITRE ATT&CK Enterprise release available at implementation time.
* [ ] Replace `ruleset/mitre/enterprise-attack.json` with the corresponding official Enterprise dataset.
* [ ] Verify the dataset format and required STIX objects are supported by Wazuh.
* [ ] Review relevant changes between the currently bundled dataset and the new release, including:

  * [ ] Techniques and sub-techniques.
  * [ ] Tactics.
  * [ ] Groups.
  * [ ] Software.
  * [ ] Mitigations.
  * [ ] Relationships and references.
[...]

PR #39235, opened 2026-09-14
Merge 5.0.0 into 5.0.1

jotacarma90:
## Description

Scheduled weekly upward merge. Part of #39205 (Week #37).

Merges branch `5.0.0` into `5.0.1`.

## Proposed Changes

- Merge `origin/5.0.0` into `5.0.1`.

### Results and Evidence

210 incoming commits, 7 conflicts.

- `api/api/spec/spec.yaml`: 5.0.1 had lost the remoted statistics rework that 5.0.0 carries (`events_failed`, `upgrade_ack`, `control_messages_queue_breakdown`/`_usage`, the byte-capacity queue descriptions and the legacy-channel wording), including in the example payload. The merge base still has that content, so it was dropped by an earlier merge into 5.0.1, not by a deliberate change: no commit on the 5.0.1 side touches the file except merges. Resolved by taking 5.0.0's specification and restoring only 5.0.1's own version strings (`version`, `x-revision` and the two license URLs).
- The three `tests/integration/test_msgraph` templates: both branches added an opt-out to the same block. Union of the two — 5.0.1's endpoint prefix opt-out (the trailing `/` from #38624) plus 5.0.0's `ssl/verification_mode: none` section (#38956), which the agent's new `system` default needs against RemotedSimulator's self-signed certificate.
- `CHANGELOG.md` and the two RPM specs: resolved keeping 5.0.1, which tracks its own release notes.

Audited afterwards that nothing else from 5.0.0 was left behind: every line the result does not carry is either a version string, a changelog entry, or one of the four files where 5.0.1 is legitimately ahead (`src/shared/src/http_op.c`, `src/data_provider/src/osinfo/sysOsParsers.cpp`, `src/wazuh_modules/syscollector` and the macOS integration-test workflow), each verified to have no commits on the 5.0.0 side since the merge base.


jotacarma90:
The `guard` job (API RBAC DB version check) needs one extra commit on this merge: 5.0.0 turns `allow_run_as` off for the `wazuh` default user (#38862), which changes `framework/wazuh/rbac/default/users.yaml`. That workflow skips any branch whose version ends in `.0.0`, so 5.0.0 never had to bump anything, but on 5.0.1 a change to the seeded defaults must come with a `CURRENT_ORM_VERSION` bump so an `rbac.db` preserved across an upgrade is migrated to them. Bumped 4 to 5, same as the merged `scan_vulnerability` policy needed in ba5d03dc8e.

Note that its first run failed for an unrelated reason: `dorny/paths-filter` asks the GitHub API for this PR's file list and got `Sorry, this diff is taking too long to generate`, which this merge's size makes likely to recur.

jotacarma90:
The msgraph integration tests failed on my conflict resolution, and the fix is in b7a97e31b5.

Both branches had patched the same block of the three `test_msgraph` templates: 5.0.1 opted the agent out of the endpoint prefix (the trailing `/` in `127.0.0.1:1517/`, 54849361add) and 5.0.0 opted it out of TLS verification (`ssl/verification_mode: none`, dff23749ff). I merged the two, keeping both opt-outs — but the prefix one is obsolete: since qa-integration-framework 2f13e97 (2026-09-01) `RemotedSimulator` answers under `/wazuh-manager/` by default, which is exactly what an agent with no `<endpoint>` path asks for. The trailing slash normalizes to an empty prefix (`w_normalize_agent_endpoint()` treats `/` as the deliberate opt-out), so every request went to the unprefixed path and got a 404, the agent never passed its startup gate, and the ms-graph module never logged anything: 10 tests failing on `callback_result is None`, while `test_invalid.py` — which only asserts that bad configurations are rejected — passed.

The templates now take 5.0.0's version verbatim: default prefix plus the TLS opt-out.

Same stale opt-out is still in `.github/workflows/5_testintegration_macos-integration-tests.yml` (`WAZUH_MANAGER_ENDPOINT='127.0.0.1/'`, 54647cb647a), where 5.0.0 installs with plain `WAZUH_MANAGER='127.0.0.1'`. That suite is path-filtered out of this PR, so it is not failing here, but it will fail the next time it runs. Left untouched deliberately: reverting it is a change to 5.0.1's own CI, not part of this merge.

jotacarma90:
Evidence that this is a base breakage and not something the merge introduced: I dispatched the same workflow on `5.0.1` itself (run 34835770755, `base_branch=5.0.1`) and it fails with exactly the same ten tests, all on `callback_result is None` — `test_API.py` (7) and `test_configuration/test_basic.py` (3), with `test_invalid.py` green.

So the suite has been red on 5.0.1 since the simulator started serving `/wazuh-manager/`, and this merge is what fixes it: it brings 5.0.0's templates, which ask for the default prefix and opt out of TLS verification.

jotacarma90:
Refreshed onto the current 5.0.0 (e4bd918aab) and force-pushed, so the branch now carries what landed there this week. Two of those commits matter for this PR's CI:

- 2039eb7b87 `fix: anchor the HTTPS client's interruptible wait to CLOCK_MONOTONIC`. The previous head failed `rtr (client-agent/https_client, agent)` on `FacadeE2eTest.SettingsChangeRefreshesStartupWithoutLeavingRegistered`, where `waitFor(recorder.startupCount, 1, 3000)` timed out — the test took 333 s for a 3 s deadline, while its neighbours in the same file take 6 to 16 s. That is the failure mode this commit describes: `Waiter::waitFor()` slept on `std::condition_variable::wait_for`, which the agent's toolchain resolves against `CLOCK_REALTIME`, so a backward clock jump extends the wait by the same amount.
- 3d3210c7c7 `fix: anchor sync protocol waits to monotonic clock`, the sister fix.

The merge resolution is unchanged otherwise, and the audit against 5.0.0 comes out the same: everything the result does not carry is a version string, a changelog entry, or one of the files where 5.0.1 is legitimately ahead. The RBAC ORM bump and the msgraph templates are re-applied on top.

jotacarma90:
Two CI fixes on top, both for failures this merge surfaces rather than causes.

**58289e5fde — the macOS ULS tests are now marked darwin-only.** `test_macos_uls_default_query.py` (added to 5.0.0 by #38831) is the only file in `test_logcollector/test_configuration/` with no `pytestmark`, and the platform markers are not decorative: `tests/integration/conftest.py` deselects by them in `pytest_collection_modifyitems`. So the file is collected on every platform, and on Windows it errors 19 times with `FileNotFoundError: C:\wazuh\etc\templates\config\darwin\localfile-extra.template` — its `TEMPLATE_PATH` is `Path(__file__).parents[4]`, but the Windows job copies only `tests\*` to `C:\wazuh`, so that resolves outside the repo. On Linux and macOS the same tests run from the full checkout and pass, which is what they did in this PR's run.

The test has never run on Windows before: #38831 only ran `verify-changelog` (its three files trigger no IT workflow), and the last upward merge into 5.0.0, #39210, touched no logcollector path so the detector never selected that suite. 5.0.0 still carries the file unmarked — worth a one-line PR there so it does not come back on the next merge.

**2185f2566b — the RBAC DB version guard compares refs locally.** `dorny/paths-filter` was asking the GitHub API for this PR's file list and getting `Sorry, this diff is taking too long to generate`: 338 files, and 404k of the changed lines are `ruleset/mitre/enterprise-attack.json` alone (the ATT&CK v19.2 update, #38863). The job died there, before ever evaluating the rule it exists for. Passing `base: ${{ github.base_ref }}` makes the action diff with git instead, which needs the base branch's history, hence `fetch-depth: 0`. Any upward merge that carries a regenerated dataset would have hit this.

jotacarma90:
Correction on the guard fix (now 56c0619ffe, amended over 2185f2566b): passing `base` was not enough. `dorny/paths-filter` only honours it outside pull requests — on a PR, if a token is available it always takes the API path, which is why the run still ended in `Sorry, this diff is taking too long to generate` after listing the first page of files. Setting `token: ''` is what makes it diff the merge commit with git instead; `fetch-depth: 0` stays, since that diff needs the commit's parent.

PR #39242, opened 2026-09-14
Close the agent trust-anchor bootstrap's crash window and fix client.keys ownership

### LINKED ISSUE #39060: Trust anchor: regression-guard the permissions, close the crash window, document the reset
**Component:** `client-agent` (`token_bootstrap.c`, `config.c`)
**Depends on:** the MVP branch `feat/39021-poc-agent-https-tier2`
**Rough size:** S

## Description

What is left of the anchor work after the MVP branch closed most of it. Kept as an issue because the remainder is small but not nothing, and because the parts that are done deserve a regression guard rather than trust.

**Done on the branch, verified at tip `8cf92bdca5`:**

- **Permissions.** `chmod(anchor_file.name, 0640)` plus `chown(AGENT_ANCHOR_CA, 0, gid)` gives **0640 root:wazuh** — root owns, the runtime group reads, nothing unprivileged writes.
- **The directory.** `chown(dir, 0, gid)` and `chmod(dir, 0750)` give **0750 root:wazuh**, with a comment naming the trap it avoids: `mkdir_ex()` creates it 0770 owned by whoever is running, and a root:root directory cannot be searched by the unprivileged user.
- **Atomic write.** `TempFile()` + `OS_MoveFile()`, write-then-rename.
- **Malformed anchors fail closed** (`91b5c49122`): `w_agent_validate_ssl_ca()` now parses the file with `w_x509_load_pem()` and refuses to start with a named error if nothing in it is a certificate — readable is no longer mistaken for usable. Closes #38949 question 10.
- **Only the pin-matched certificate is installed** (`155686075a`), not the whole `/cacerts` body, so a bundle carrying an attacker CA beside the genuine root cannot smuggle the first in as an anchor.
- **Re-resolution after a late write.** `AgentdStart()` records whether the anchor existed before the bootstrap and re-resolves when that boot created it, so the posture `ClientConf()` settled in `main()` is not stale.

## Tasks

- [ ] **Regression-guard the permissions by test, not by inspection.** A process running as the `wazuh` user must not be able to rewrite or delete the anchor, and must still be able to read it after the privilege drop. Both modes are right today; nothing stops a future change to `TempFile()`'s umask or to `mkdir_ex()` from quietly undoing either.
- [ ] **Close the crash window between the rename and the ownership fix-up.** `OS_MoveFile()` is atomic; the `chown()` that follows is a separate step. An interruption between them leaves a root-owned anchor the runtime user cannot read — the agent then fails closed on a file that is, in content, perfectly good.
- [ ] **Decide `client.keys`'s posture.** The bootstrap chowns it to the runtime *user*, while the installer's own convention is `0640 root:wazuh`. Either converge them or record why the bootstrap differs.
- [ ] **Document the reset.** Removing the anchor, or re-running the bootstrap with a new token, is how an agent is moved to a different manager or CA on purpose. That is also the off switch the design intends instead of `verification_mode=none` — coordinate the wording with whatever #39123 decides.
- [ ] Settle with #39123 whether the anchor stays a fixed filename or becomes a scanned directory. A scan makes every writable entry in `etc/certs` a trust anchor, which changes what the permissions above have to protect.

## Definition of Done

- [x] A test asserts the `wazuh` user cannot rewrite or delete the anchor, and can read it.
- [x] An interrupted write or an interrupted ownership fix-up leaves the agent in a state it recovers from on the next boot.
- [ ] `client.keys` and the anchor have the same stated ownership rule, or the difference is written down.
- [ ] The reset is documented and matches the off switch #39123 settles on.

Related: #38940, #39021

lchico:
## Description

The enrollment-token bootstrap (`w_agent_token_bootstrap()`, `src/client-agent/src/token_bootstrap.c`) fixes up file ownership while still running as root, before the privilege drop, for two files it writes: the trust anchor (`root-ca.pem`) and `client.keys`. Two gaps remained here:

- The anchor's ownership fix-up ran *after* the atomic rename that installs it at its final path. A process death between the rename and the `chown()` left the anchor on disk with the wrong group, permanently and unrecoverably — the bootstrap latches off for good once an anchor file exists, so nothing would ever repair it.
- `client.keys`'s posture was left undecided: the bootstrap re-owned it to the runtime `wazuh` user, while the installer's own convention creates it `root:wazuh`.

Closes #39060.

## Proposed Changes

- Moved the anchor's `chown()` to run on the temporary file, before `OS_MoveFile()` renames it into place, instead of on the final path afterward. Since the temp file is always created in the same directory as the destination, the rename is atomic and same-filesystem — once the temp file carries the right ownership, whatever appears at the final path already does too. This closes the crash window entirely rather than merely narrowing it.
- Resolved `client.keys`'s ownership: the bootstrap now `chown()`s it to `root:<runtime-group>`, matching the installer's own default, instead of handing ownership to the runtime user. Chowning goes through a new O_NOFOLLOW-vetted, hard-link-checked `fchown()` path (`w_token_bootstrap_open_and_chown_keys()` / `w_token_bootstrap_chown_keys_file()`, reusing `file_op.c`'s `w_openat_nofollow_vetted()`, now exported for that reuse), since `etc/` is group-writable by the runtime user and a plain path-based `chown()` would follow a symlink planted there. A defensive repair was added to both the anchor-latch and "already enrolled" fast paths too, so a boot that dies between a key replacement and this chown self-heals on the next start instead of staying stuck.
- Documented, as a code comment, the three conditions that together make up today's actual "reset" mechanism for moving an already-bootstrapped agent to a different manager/CA (the anchor removed, `client.keys` cleared, and a new enrollment-token file placed — all three, not any one alone). Whether this becomes a formally named mechanism is left to a separate, still-open design discussion.
- Extended the client-agent unit-test suite to regression-guard the above: an ordering assertion proving the anchor's ownership fix-up genuinely happens before the rename, explicit assertions on the anchor and its parent directory's exact mode/ownership values, assertions on `client.keys`'s corrected ownership target on both the enrollment success and repair paths, hard-link/FIFO rejection on the `client.keys` open, and an ownership-survives-a-restart guard that chains a fresh enrollment with a second bootstrap call to confirm `client.keys` stays `root:<runtime-group>` on the next boot rather than regressing to something the runtime user can no longer read.

Not in scope: whether the trust anchor should become a scanned directory instead of a fixed filename remains an open, separate design question and is untouched here. `w_token_bootstrap_ensure_parent_dir()`'s own `chown()`/`chmod()` on the anchor's parent directory (`etc/certs`) is still path-based rather than O_NOFOLLOW-vetted like `client.keys`'s new path — flagged in review as a pre-existing gap of the same class, tracked as a follow-up rather than folded into this PR.

### Unit tests

The client-agent unit-test suite (CMocka) was extended and exercised locally, rebuilt via the `wazuh-cpp-unittest-builder` image against this branch's current head. All of it passes, including the new regression guards:

<details>
<summary>Full `test_token_bootstrap` run (14/14 passing)</summary>

```
[==========] tests: Running 14 test(s).
[ RUN      ] test_no_token_file_is_noop
[       OK ] test_no_token_file_is_noop
[ RUN      ] test_anchor_already_present_skips_and_discards_the_token
[       OK ] test_anchor_already_present_skips_and_discards_the_token
[ RUN      ] test_already_enrolled_skips_and_discards_the_token
[       OK ] test_already_enrolled_skips_and_discards_the_token
[ RUN      ] test_client_keys_hard_link_is_not_chowned
[       OK ] test_client_keys_hard_link_is_not_chowned
[ RUN      ] test_keys_chown_failure_is_logged
[       OK ] test_keys_chown_failure_is_logged
[ RUN      ] test_anchor_latch_keys_chown_failure_is_quiet
[       OK ] test_anchor_latch_keys_chown_failure_is_quiet
[ RUN      ] test_empty_placeholder_keys_file_is_not_already_enrolled
[       OK ] test_empty_placeholder_keys_file_is_not_already_enrolled
[ RUN      ] test_malformed_token_logs_named_error_and_writes_nothing
[       OK ] test_malformed_token_logs_named_error_and_writes_nothing
[ RUN      ] test_fetch_failure_logs_named_error_and_writes_nothing
[       OK ] test_fetch_failure_logs_named_error_and_writes_nothing
[ RUN      ] test_pin_mismatch_logs_named_error_and_writes_nothing
[       OK ] test_pin_mismatch_logs_named_error_and_writes_nothing
[ RUN      ] test_full_happy_path_via_pin
[       OK ] test_full_happy_path_via_pin
[ RUN      ] test_fresh_enrollment_keys_chown_failure_logs_merror
[       OK ] test_fresh_enrollment_keys_chown_failure_logs_merror
[ RUN      ] test_credential_less_token_enrolls_without_error
[       OK ] test_credential_less_token_enrolls_without_error
[ RUN      ] test_full_happy_path_via_ca_pem
[       OK ] test_full_happy_path_via_ca_pem
[==========] tests: 14 test(s) run.
```
</details>

The anchor's chown-before-rename ordering assertion was confirmed to be a real regression guard, not a tautology: reverting just the reorder makes both happy-path tests fail on that assertion specifically, before the fix is reapplied and they pass again. The restart/repair behaviour is deliberately covered live rather than at the mock level — see "Restart behaviour" below.

The full client-agent suite (14 test binaries covering every daemon/module under `src/client-agent`) was also rebuilt and run, with no regressions:

<details>
<summary>Full client-agent <code>ctest</code> run (14/14 passing)</summary>

```
Test project src/unit_tests/build/client-agent
      Start  1: test_start_agent
 1/14 Test  #1: test_start_agent ...................   Passed    0.11 sec
      Start  2: test_agentd_state
 2/14 Test  #2: test_agentd_state ..................   Passed    0.01 sec
      Start  3: test_agcom
 3/14 Test  #3: test_agcom .........................   Passed    0.01 sec
      Start  4: test_agentd
 4/14 Test  #4: test_agentd ........................   Passed    0.02 sec
      Start  5: test_startup_gate
 5/14 Test  #5: test_startup_gate ..................   Passed    0.01 sec
      Start  6: test_agent_ssl_ca
 6/14 Test  #6: test_agent_ssl_ca ..................   Passed    0.01 sec
      Start  7: test_agent_config
 7/14 Test  #7: test_agent_config ..................   Passed    0.01 sec
      Start  8: test_client_conf_ssl_resolution
 8/14 Test  #8: test_client_conf_ssl_resolution ....   Passed    0.02 sec
      Start  9: test_enrollment
 9/14 Test  #9: test_enrollment ....................   Passed    0.02 sec
      Start 10: test_token_bootstrap
10/14 Test #10: test_token_bootstrap ...............   Passed    0.02 sec
      Start 11: test_https_client_bridge
11/14 Test #11: test_https_client_bridge ...........   Passed    0.02 sec
      Start 12: test_agent_report
12/14 Test #12: test_agent_report ..................   Passed    0.01 sec
      Start 13: test_task_registry_ipc_component
13/14 Test #13: test_task_registry_ipc_component ...   Passed    0.05 sec
      Start 14: test_vd_offset_ipc_component
14/14 Test #14: test_vd_offset_ipc_component .......   Passed    0.05 sec

100% tests passed, 0 tests failed out of 14
```
</details>

`src/shared`'s `test_file_op` suite (53 tests) was also rebuilt and re-run after exporting `w_openat_nofollow_vetted()` and adding its own precondition checks (bare filename, `O_NOFOLLOW` in `oflags`) — no regressions, including its own `w_fopen_nofollow`/`w_gzopen_nofollow` symlink/hard-link/FIFO rejection tests that exercise the same shared helper.

### Live end-to-end verification

Because the failure mode this guards against (a file ending up unreadable by the runtime user after a privilege drop) can't be observed from mocked unit tests, it was verified against real installed packages built from this branch's current head, enrolling against a real running manager.

Both packages were built by CI from `512b5fc167` — `wazuh-agent_5.0.0-0_amd64_512b5fc.deb` and `wazuh-agent_5.0.0-0_x86_64_512b5fc.rpm` — and the build actually installed was confirmed in-container to be the one under test (the repair path's own log string is present in the installed binary, absent from every earlier build).

#### Fresh enrollment

| | Ubuntu 22.04 (`.deb`) | AlmaLinux 9 (`.rpm`) |
|---|---|---|
| agent | `agent1-ubuntu22.04-39060-512b5fc` | `agent4-almalinux9-39060-512b5fc` |
| `client.keys` | `root:wazuh` (`0:999`), mode `640` | `root:wazuh` (`0:998`), mode `640` |
| `certs/root-ca.pem` | `root:wazuh` (`0:999`), mode `640` | `root:wazuh` (`0:998`), mode `640` |
| read back as the runtime user | `exit code: 0` | `exit code: 0` |
| one-shot token | consumed, file absent | consumed, file absent |
| daemons | all 5 running | all 5 running |
| manager view | `active` | `active` |

The runtime group's numeric gid differs by distro/package (`999` Debian-family vs `998` RHEL-family), as expected — the ownership *pattern* is identical. `client.keys` was read back **as the unprivileged `wazuh` user**, not merely inspected for permission bits.

Enrollment itself was clean on both:

```
Enrollment token stored; the manager was set to 'manager' and the trust anchor will be bootstrapped at the first agent start.
wazuh-agentd: INFO: Token bootstrap: enrollment succeeded; the manager's CA is now the agent's trust anchor.
```

#### Restart behaviour

A **stop/start cycle** was then run on each agent — the case this PR's repair paths exist for, and the one raised in review as previously having left the agent unable to read `client.keys` at all. On both platforms the restart was confirmed real (new `wazuh-agentd` pid and a later unit activation timestamp, not a no-op), and afterwards `client.keys` and the anchor were byte-for-byte the same ownership and mode as before, the runtime user still read the credential (`exit code: 0`), all five daemons came back, and the manager still reported the agent `active` — re-checked again after a 45s settle to rule out a connect-then-drop blip.

A clean restart only proves nothing regressed, so the repair itself was then exercised directly: `client.keys` was deliberately put into the broken state a crash between the key write and its chown leaves behind (`root:root`), and the runtime user was confirmed to genuinely lose access, reproducing the reported symptom:

```
/var/ossec/etc/client.keys  owner=root:root  numeric=0:0  mode=640
$ su -s /bin/bash -c 'cat /var/ossec/etc/client.keys' wazuh
cat: /var/ossec/etc/client.keys: Permission denied
exit code: 1
```

A single agent restart repaired it, on both platforms:

```
/var/ossec/etc/client.keys  owner=root:wazuh  numeric=0:999  mode=640
$ su -s /bin/bash -c 'cat /var/ossec/etc/client.keys' wazuh
004 agent1-ubuntu22.04-39060-512b5fc any bfbad5f32d051c5754b6bdd6512b7ba9df0ca5a372c1df6660b5544404b68ced
exit code: 0
```

with all five daemons running and the manager reporting the agent `active` again. This is the anchor-latch repair path doing exactly what it was added for, observed end to end rather than at the mock level.

**One unrelated observation:** `systemctl is-active wazuh-agent` reports `failed` immediately after a stop on both platforms (`Type=forking` with `ExecStop=/var/ossec/bin/wazuh-control stop`, whose main process exits `1`). It does not prevent the subsequent start, and it is packaging/unit behaviour untouched by this change — noted here only so the raw logs don't read as a regression introduced by this PR.

### Procedure summary

For each platform/package combination: the agent container was purged of any prior install (`/var/ossec` confirmed gone), the package built from this branch's head was staged into it, a fresh one-shot enrollment token was minted on the running manager (`wazuh-manager-authd --create-enrollment-token --address manager`), and the agent was installed with only `WAZUH_ENROLLMENT_TOKEN`/`WAZUH_AGENT_NAME` in the environment and started. `client.keys` and the trust anchor's ownership/mode were then read directly off disk both symbolically and by numeric uid/gid, the credential was read back as the actual unprivileged runtime user rather than merely inspected, the daemon set was checked, and the manager's own agent view (`cluster_control -a`) was queried. The agent was then stopped and started again, every one of those checks was repeated, and the manager view was queried once more after a 45s settle. Finally `client.keys` was forced into the `root:root` broken state, the runtime user's loss of access was confirmed, and a restart was used to show the repair path restoring both ownership and access.

Each run's full transcript is archived under `03-testing/5x/39060-trust-anchor-hardening/evidence/`. The two enrollment tokens actually consumed show `1/unlimited` uses on the manager; three tokens minted during setup and never consumed were revoked so the token store reflects only the real runs.

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests _(Wazuh v4.x)_ — N/A, not applicable to this change

- Engine _(Wazuh v5.x and above)_ — N/A, not applicable to this change (client-agent bootstrap, not the engine)
  - [ ] Test run in parallel

- Wazuh server API/Framework — N/A, not applicable to this change

### Artifacts Affected

- `src/client-agent/src/token_bootstrap.c` / `src/client-agent/include/token_bootstrap.h` — the agent daemon binary (`wazuh-agentd`).
- `src/shared/src/file_op.c` / `src/shared/include/file_op.h` — `w_openat_nofollow_vetted()` exported (was file-local `static`) for `token_bootstrap.c` to reuse, plus its own precondition hardening.
- Linux agent packages (`.deb`/`.rpm`, `amd64`/`x86_64`) were built from this branch specifically to verify the fix end-to-end; no packaging logic itself was changed.
- The Windows path through this file is an existing no-op stub (no privilege-drop step on that platform) and is unaffected by this change.

### Configuration Changes

N/A — no new configuration parameters, no change to default values.

### Tests Introduced

Extended `src/unit_tests/client-agent/test_token_bootstrap.c` (CMocka):
- An ordering assertion proving the anchor's ownership fix-up happens before the rename that installs it.
- Explicit assertions on the exact mode bits applied to the anchor and its parent directory.
- Assertions on `client.keys`'s corrected ownership target, on the successful-enrollment path, the already-enrolled fast path, and the anchor-latch repair path.
- Hard-link rejection and FIFO-safety coverage for `client.keys`'s new O_NOFOLLOW-vetted open.
- Failure-logging coverage for `client.keys`'s chown, at both the quiet (repair) and loud (fresh-enrollment) log levels.

Extended `src/unit_tests/shared/test_file_op.c` coverage is exercised indirectly through the shared `w_openat_nofollow_vetted()` path; no new dedicated test was added there since the new precondition checks are no-ops for its existing callers.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...


cborla (commented):
LGTM
- [x] Code changes reviewed

cborla (approved):

Issue #39182, opened 2026-09-11
launchd plist arrays are flattened into one space-separated string with no quoting, so ProgramArguments, WatchPaths and QueueDirectories entries are unrecoverable whenever a path contains a space


Rebits:
## Description

Three launchd plist keys are arrays: `ProgramArguments` is an argv vector, and `WatchPaths` and `QueueDirectories` are lists of filesystem paths. `LaunchdProvider::parsePlistFile` reads each of them element by element, correctly, and then destroys the structure by concatenating the elements with a single space and no quoting or escaping:

```cpp
std::string joinedValue = joinArrayElements(elements);

if (keyPair.second == "program_arguments") service.programArguments = joinedValue;
else if (keyPair.second == "watch_paths") service.watchPaths = joinedValue;
else if (keyPair.second == "queue_directories") service.queueDirectories = joinedValue;
```

```cpp
std::string LaunchdProvider::joinArrayElements(const std::vector<std::string>& arrayElements)
{
    if (arrayElements.empty())
    {
        return "";
    }

    std::ostringstream oss;

    for (size_t i = 0; i < arrayElements.size(); ++i)
    {
        if (i > 0)
        {
            oss << " ";
        }

        oss << arrayElements[i];
    }

    return oss.str();
}
```

The separator is also a legal character inside every one of these values. A macOS path routinely contains spaces, and a command-line argument obviously may. Once joined, there is no way to tell a separator from a space that belongs to an element, so the original array cannot be reconstructed.

The destination fields are arrays. `process.args` and `service.starts.on_path_modified` are both array-typed in the index, so they were designed to carry the elements separately and instead receive a single string that becomes an array of exactly one element.

The clearest demonstration is `WatchPaths`, because the elements are paths and the damage is visible without knowing the plist. `com.apple.helpd` watches two directories and reports them as one:

```json
{"service": {"id": "com.apple.helpd",
             "starts": {"on_path_modified": ["/Applications/ /Applications/Utilities/"]}}}
{"service": {"id": "com.apple.kernelmanagerd",
             "starts": {"on_path_modified": ["/Library/Extensions /AppleInternal/Library/Extensions"]}}}
{"service": {"id": "com.apple.nfsconf",
             "starts": {"on_path_modified": ["/etc/nfs.conf /private/var/tmp/nfsclient_kext.tmp"]}}}
```

[...]

PR #39272, opened 2026-09-14
Populate macOS service enabled flag and executable path

### LINKED ISSUE #39139: macOS service inventory reports a blank state, blank enabled flag and pid 0 for all 904 services, so running services cannot be distinguished from stopped ones
## Description

The macOS agent inventories 904 services and populates almost none of their metadata. Every row carries a single space character as its state, a single space as its enabled flag, `0` as its process id, and empty strings for type, executable and user. The result is an inventory that lists which services are *defined* on the host but says nothing about which are actually running, which is the question the state index exists to answer.

Windows and Linux both populate real states. Aggregating `service.state` per agent:

```
  agent 006 total=904 distinct_pids=1
      (service.state absent/empty)                <-- macOS
  agent 005 total=653 distinct_pids=76
      state=STOPPED              452
      state=RUNNING              201
  agent 002 total=209 distinct_pids=1
      state=inactive             138
      state=active                70
      state=activating             1
  agent 001 total=205 distinct_pids=1
      state=inactive             136
      state=active                69
  agent 003 total=129 distinct_pids=1
      state=inactive              82
      state=active                47
  agent 004 total=129 distinct_pids=1
      state=inactive              82
      state=active                47
```

On the agent itself the values are not NULL, which is worth stating precisely: they are a literal space. Every one of the 904 rows has the same single-byte value, hex `20`:

```
select hex(service_state)||' count='||count(*) from dbsync_services group by hex(service_state);
20 count=904

select hex(service_enabled)||' count='||count(*) from dbsync_services group by hex(service_enabled);
20 count=904
```

A concrete row makes the shape clear. This is the Wazuh agent's own launchd job, which is definitely loaded and running:

```
      service_name = com.wazuh.agent.plist
      service_type =
     service_state =
   service_enabled =
service_start_type = true
       process_pid = 0
process_executable =
 process_user_name =
```

while `launchctl` reports it running with a real pid:

```
$ launchctl list | grep -i wazuh
2279	0	com.wazuh.agent
```

Two further facets of the same shallow implementation are worth recording, because they show the provider is enumerating plist files rather than querying launchd:

1. **Names are filenames, not launchd labels.** All 904 `service_name` values end in `.plist` (`com.wazuh.agent.plist`, `com.apple.eppc.plist`). The launchd label is `com.wazuh.agent`, and that is what `launchctl` and every other macOS tool uses, so the inventory key cannot be joined against anything.
2. **Duplicate rows for a single logical service.** 15 names appear twice, which is what you get from walking both the LaunchDaemons and LaunchAgents directories where the same label is defined in both:

```
com.apple.AssetCacheLocatorService.plist x2
com.apple.ContainerMigrationService.plist x2
com.apple.XprotectFramework.PluginService.plist x2
com.apple.akd.plist x2
com.apple.cloudd.plist x2
com.apple.coreservices.sharedfilelistd.plist x2
com.apple.ctkd.plist x2
com.apple.dprivacyd.plist x2
com.apple.findmymacmessenger.plist x2
com.apple.iconservices.iconservicesagent.plist x2
com.apple.lsd.plist x2
com.apple.nsurlsessiond.plist x2
com.apple.secinitd.plist x2
com.apple.symptomsd-diag.plist x2
com.apple.testmanagerd.plist x2
```

The count itself is consistent with that reading and is not the problem: there are 892 plists across `/System/Library/LaunchDaemons`, `/System/Library/LaunchAgents`, `/Library/LaunchDaemons` and `/Library/LaunchAgents`, against 904 inventoried, whereas only 420 jobs are actually loaded per `launchctl list`. Enumerating defined services is a reasonable choice. Reporting them all with no state is not.

The security consequence is that a macOS endpoint's service posture is unreadable. Detecting an unexpected service that is actually running, or a security service that has stopped, is not possible from this data, while the same questions are answerable on Windows and Linux.

## Environment

- Agent under test: `demo-env-macos-agent`, id `006`, macOS 26.5.1 (build 25F80), arm64, Apple M1 (Virtual), Parallels VM on MacStadium host `macstadium_arm_4`.
- Comparison agents: `005` windows-11-amd64, `001` ubuntu-26-arm64, `002` ubuntu-26-amd64, `003` amazon-2023-amd64, `004` amazon-2023-arm64, all from the same nightly.

## Steps to reproduce

1. Enroll a 5.0.0 macOS agent and let a syscollector evaluation complete.
2. Aggregate service state per agent from the indexer, including at least one Windows or Linux agent for comparison:

```bash
  "https://<indexer>:9200/wazuh-states-inventory-services/_search?size=0" \
  -d '{"aggs":{"ag":{"terms":{"field":"wazuh.agent.id","size":10},
       "aggs":{"st":{"terms":{"field":"service.state","size":6}},
               "pids":{"cardinality":{"field":"process.pid"}}}}}}'
```

3. On the macOS agent, inspect the raw values, checking the bytes rather than trusting a NULL test:

```bash
sudo sqlite3 /Library/Ossec/queue/syscollector/db/local.db \
  "select hex(service_state)||' count='||count(*) from dbsync_services group by hex(service_state);"
sudo sqlite3 /Library/Ossec/queue/syscollector/db/local.db \
  "select count(*) from dbsync_services where coalesce(process_pid,0)<>0;"
```

4. Compare a known-running service against `launchctl`:

```bash
sudo sqlite3 -line /Library/Ossec/queue/syscollector/db/local.db \
  "select service_name, service_state, service_enabled, process_pid from dbsync_services where service_name like '%wazuh.agent%';"
launchctl list | grep -i wazuh
```

5. Confirm the naming and duplication:

```bash
sudo sqlite3 /Library/Ossec/queue/syscollector/db/local.db \
  "select count(*) from dbsync_services where service_name like '%.plist';"
sudo sqlite3 /Library/Ossec/queue/syscollector/db/local.db \
  "select service_name||' x'||count(*) from dbsync_services group by service_name having count(*)>1;"
```

## Expected

Each inventoried macOS service reports whether it is loaded and running, as Windows reports `RUNNING`/`STOPPED` and Linux reports `active`/`inactive`, with the launchd label as its identity and one row per logical service. A running service reports its pid.

## Actual

All 904 macOS services report `service_state` and `service_enabled` as a single space (hex `20`), `process_pid` as `0`, and empty `service_type`, `process_executable` and `process_user_name`. The indexed documents end up with no usable `service.state` value at all, and only one distinct `process.pid` across the whole agent. Names are plist filenames rather than launchd labels, and 15 labels appear twice. The Wazuh agent's own job is reported with a blank state and pid 0 while `launchctl` shows it running as pid 2279.

## Not a duplicate of

`qa-known-issues.py check` scored one cache candidate and the live search returned nothing:

- `wazuh/wazuh#15448` "The NSIS installer for Windows reports the service running" (0.50) concerns the Windows NSIS installer's own reporting of the Wazuh service during installation. Different platform, different component (installer rather than syscollector), and it is about the installer's status message rather than the contents of the service inventory.
- Live search for `distinguished inventory services running reports service` returned nothing open.

nbertoldo:
## Description

Related to #39139.

The macOS service inventory reports a blank `service.enabled`, a blank `process.executable` for most jobs, and a permanently false `service.starts.on_mount` and `service.inetd_compatibility`. This PR fixes those four fields. It does **not** fix `service.state` and `process.pid`, which are genuinely hardcoded because the collector has no runtime source at all; that work is tracked in #39273 and is not a prerequisite for this change.

Three of the four are the same class of defect: the plist parser and its consumer disagree about how a value is spelled, so a branch that looks correct is simply never reached.

`LaunchdProvider::parsePlistFile` renders a plist boolean as the words `true`/`false`:

```cpp
CFBooleanRef boolValue = static_cast<CFBooleanRef>(value);
std::string stringVal = CFBooleanGetValue(boolValue) ? "true" : "false";
```

`SysInfo::getServices()` then consumed that string expecting the characters `0` and `1`, so for `Disabled` neither branch ever matched and the `else` ran for every service. The inversion logic itself was correct and unreachable.

The same mismatch reached `service_starts_on_mount` through the `stringToInt` helper, which runs `std::stoi` inside a try/catch that returns 0 on failure. `StartOnMount` is a plist boolean, so `std::stoi("true")` threw and the field was false regardless of what the plist said.

`service_inetd_compatibility` was stuck at 0 for a different reason, which is worth stating precisely because the fix is not the same. `inetdCompatibility` is declared in `launchd.plist(5)` as a **dictionary**, not a boolean, and the parser only handled `CFString`, `CFBoolean` and `CFNumber`. The value matched no branch, so the field was never populated and reached `stringToInt` as an empty string. Changing only the conversion would have left it at 0 forever, so the parser now recognises the dictionary form.

The fourth field is a missed fallback rather than a type mismatch. A launchd job may declare its executable either as `Program`, a single string, or as the first element of `ProgramArguments`, an argv array. Only the first was consulted, and the second form is much the more common one in Apple's own jobs, so 588 of 904 services on the agent under test lost an executable path that had already been read successfully and was sitting in the neighbouring field.

Treating an absent `Disabled` key as enabled turned out to be wrong for the same reason, and it is the one place where an earlier revision of this branch made a field worse rather than better. launchd does not record `launchctl disable` in the job plist, so a service an administrator had disabled was reported as enabled where it had previously been blank. Reading the override database fixes that, and it also corrects the opposite case: a job whose plist ships `Disabled=true` and which `launchctl enable` has since turned on.

That fallback is applied in the provider rather than in the ECS mapping, and the distinction matters for #39182. That issue reports that array values are flattened into one space-separated string before they reach `sysInfoMac.cpp`, which makes the first element unrecoverable there. Inside `parsePlistFile` the elements are still a separate vector at the point the fallback runs, so this change is correct today and does not touch `joinArrayElements`. The two fixes are independent and will not conflict.

## Proposed Changes

- `service_enabled` now accepts the `true`/`false` spellings the provider actually emits, in addition to the `0`/`1` a plist may use. An unset `Disabled` key is treated as enabled, which is the documented launchd default.
- `service_starts_on_mount` and `service_inetd_compatibility` are converted with a new boolean-aware helper instead of `std::stoi`. `stringToInt` is kept for `service_frequency`, whose `StartInterval` really is numeric.
- `LaunchdProvider::parsePlistFile` recognises `inetdCompatibility` in its documented dictionary form.
- `process_executable` falls back to the first element of `ProgramArguments` when `Program` is absent.
- `LaunchdProvider` reads the launchd override database and lets it decide the state of any label it names. `launchctl enable` / `disable` and `unload -w` record that state in `/var/db/com.apple.xpc.launchd/disabled*.plist` rather than in the job plist, so it outranks the plist in both directions. A label named by more than one domain counts as disabled if any domain disables it, because one inventory row cannot express a per-domain state. An absent or unreadable database leaves the previous behaviour untouched.
- A `Disabled` key spelled as a number is captured, which the parser dropped because its `CFNumber` branch only assigned `start_interval`.

### Results and Evidence

Measured on macOS 26.5.1 (build 25F80) arm64 against a 5.0.0 rc1 all-in-one, installing the baseline package built from `1a5b074` and then packages built from this branch, wiping `queue/syscollector/db/local.db` between runs so each scan is built from scratch. The loaded `libsysinfo.dylib` was confirmed to carry this branch's code before each measurement.

| | baseline `1a5b074` | branch, before the override fix | branch, final |
|---|---|---|---|
| `service_enabled` blank (hex `20`) | all | 5 | **5** |
| `service_enabled` = `1` (enabled) | 0 | 860 | **862** |
| `service_enabled` = `0` (disabled) | 0 | 36 | **36** |
| `process_executable` populated | 316 | 900 | **902** |
| `service_inetd_compatibility` = 1 | 0 | 9 | **9** |
| `service_starts_on_mount` = 1 | 0 | 0 | 0 |
| `service_state` blank (hex `20`) | all | all | all |
| `process_pid` <> 0 | 0 | 0 | 0 |

The baseline reproduces the report exactly: every service blank, and 316 populated executables.

The host held 901 job plists for the first two runs and 903 for the last, having gained two between measurements, so the totals that scale with the number of services are two higher in the final column while every count the fix governs is unchanged. The enumeration was checked against the filesystem directory by directory and agrees exactly: 422 in `/System/Library/LaunchDaemons`, 465 in `/System/Library/LaunchAgents`, 7 and 5 under `/Library/Apple/System/Library/`, 3 in `/Library/LaunchDaemons` and 1 in `/Library/LaunchAgents`.

The override database moves eight services, in both directions. Six that an administrator had disabled with `launchctl` were reported as enabled and now report disabled: `com.apple.CSCSupportd`, `com.apple.FolderActionsDispatcher`, `com.apple.ManagedClientAgent.enrollagent`, `com.apple.ScriptMenuApp`, `com.apple.Siri.agent` and `com.apple.appleseed.seedusaged.postinstall`. Two whose plist ships `Disabled=true` but which have since been enabled were reported as disabled and now report enabled: `com.openssh.sshd` and `com.apple.screensharing`. The second case is the more telling one — the host's Remote Login is on and serving the session these measurements were taken through, while the inventory called it disabled.

The result was checked against launchd itself rather than against the expectation. Every entry `launchctl print-disabled` reports, for both the system domain and `user/501`, was compared with the inventory's `service_enabled`:

```
entries in launchctl print-disabled : 12
  agreeing with the inventory       : 11
  absent from the inventory         : 1
  DISAGREEING                       : 0
```

The one absent entry is a label launchd knows about that has no plist in the enumerated search paths, so there is no row for it to match.

`service_starts_on_mount` does not move, and that is expected rather than a failure: exactly one plist on the host declares `StartOnMount` and its value is false, so the field reads 0 throughout. That fix follows from the code, not from the data. The same holds for the numeric spellings the `CFNumber` branch now accepts, which no plist on the host uses; both are covered by unit tests instead.

`service_state` and `process_pid` are unchanged by design; they are the subject of #39273.

`astyle --options=ci/input/astyle.config` reports no formatting changes for any of the files.

### Artifacts Affected

macOS agent only. No manager-side change, and no change to the event schema or to the set of emitted fields.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

Ten cases in `test_launchd_darwin.cpp`, running on macOS in `5.X - macOS - Unit tests` as part of `ServicesTests`.

The suite could previously only run `collect()` against the live host and assert structural properties, so a regression in any field this PR touches would have passed. Mocking the filesystem wrapper does not help: `parsePlistFile` opens the file through CoreFoundation rather than through the wrapper, and the search paths were hardcoded. The constructor now takes the search paths and the override directory, so a test can point the provider at a temporary directory of real plists. Supplying search paths also clears the per-user list, and every test passes a fixture override directory, so no case depends on the accounts or the override database of the machine running it.

The cases cover both spellings of a plist boolean, the numeric and dictionary forms, an absent key as distinct from an unevaluated one, the executable fallback and its precedence over `ProgramArguments`, and the override database including its cross-domain resolution and its absence. They give the two forms that no plist on the test host uses — a numeric `Disabled` and a conditional one — the coverage the field measurements cannot.

## Review Checklist

- [x] Code changes reviewed


cborla (approved):
LGTM
- [x] Code changes reviewed

PR #39290, opened 2026-09-15
Stop reporting CurrentControlSet\Services OS telemetry as registry changes

### LINKED ISSUE #39274: Default Windows FIM reports OS telemetry churn under CurrentControlSet\Services as T1112
|---|---|---|---|---|
| 5.0.0 | wazuh-agent / syscheck (FIM registry) — default Windows `ossec.conf` | Agent | Packages | Windows 11 Pro 10.0.26200 (all Windows) |

## Summary

The default Windows agent monitors `HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services` with the **default recursion level of 512**, i.e. the whole subtree. That subtree is not just service definitions — it is also where Windows continuously writes Background Activity Moderator process-tracking values, DHCP interface leases, WMI performance counters and W32Time secure-time limits.

Every one of those OS-internal writes becomes a `Wazuh FIM - Registry modified` finding at **medium** level, tagged **T1112 Modify Registry / TA0005 Defense Evasion**. On an idle Windows 11 lab agent, **94 of 97 FIM events** were this noise. A real service-image change is buried in it, and T1112/Defense Evasion is the wrong verdict for the OS recording that `sdbinst.exe` ran.

The shipped ignore list already carves out `MpsSvc` and `PolicyAgent` keys, so the pattern is established — it just does not cover the keys that actually churn.

### Extend the ignore list

`src/win32/ossec.conf`, next to the existing `MpsSvc` / `PolicyAgent` entries in the `<syscheck>` block:

```xml
<registry_ignore>HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\bam\State</registry_ignore>
<registry_ignore>HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\Tcpip\Parameters\Interfaces</registry_ignore>
<registry_ignore>HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\W32Time\SecureTimeLimits</registry_ignore>
<registry_ignore>HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\SharedAccess\Epoch</registry_ignore>
<registry_ignore type="sregex">\Performance$</registry_ignore>
```

Key-level ignores are checked before `RegOpenKeyEx` (`src/syscheckd/src/registry/registry.c:1932`), so each entry skips its whole subtree without enumerating it. Matching is a **case-insensitive prefix compare** (`registry.c:1281-1290`), which is why one line covers every key below it.

#### Coverage against the reported findings

| Path under `…\Services` | Events | Covered by | Result |
|---|---|---|---|
| `bam\State\UserSettings\S-1-5-18` | 18 | prefix `bam\State` | silenced |
| `Tcpip\Parameters\Interfaces\{ee540391-…}` | 6 | prefix | silenced |
| `WmiApRpl\Performance` | 5 | `\Performance$` | silenced |
| `bam\State\UserSettings\S-1-5-19` | 4 | prefix `bam\State` | silenced |
| `W32Time\SecureTimeLimits` | 4 | prefix | silenced |
| `W32Time\SecureTimeLimits\RunTime` | 3 | prefix | silenced |
| `SharedAccess\Epoch` | 2 | prefix | silenced |
| **`Tcpip\Parameters`** | 2 | — | **kept on purpose** |
| **`MDCoreSvc`** | 3 | — | **kept on purpose** |
| **`BITS`** | 2 | — | **kept on purpose** |
| other `Services` / `Policies` keys | ~45 | — | **must be enumerated, see below** |

**42 of the 49 enumerated findings are silenced.** The three rows marked *kept on purpose* are deliberate, not oversights:

- **`Tcpip\Parameters`** — DHCP writes `DhcpNameServer` / `DhcpDomain` / `DhcpDefaultGateway` here, but the same key holds `NameServer`, `Domain` and `SearchList`. Blanket-ignoring it would blind the default config to DNS-resolver tampering. Two events per day is the cheaper trade.
- **`MDCoreSvc` and `BITS`** — genuine service keys at depth 1, written by Windows Update and Defender when they update themselves. These are exactly what the `Services` monitor exists to watch; ignoring them would open a persistence gap (BITS hijacking is T1197).

### Enumerate the remaining ~45 findings before closing

This is a denylist, so it can only cover keys someone has seen. The source report enumerates 49 of the 94 findings and groups the rest as *"other Services / Policies keys"*, which means **this list cannot be shown to cover them**.

Required before this issue is closed: on a clean idle Windows 11 host, aggregate the `Wazuh FIM - Registry` findings by `registry.path`, triage each remaining path as OS telemetry or real change, and extend the list above accordingly. Note that some of those ~45 may sit under `HKLM\Software\Policies` or `HKLM\Software\Microsoft\Windows\CurrentVersion\Policies`, which are **separate monitored entries** — a `…\Services\…` ignore will not touch them, and those two are configured `arch="both"`, so an ignore for them needs `arch="both"` too.

<details>
<summary>Three implementation details that are easy to get wrong</summary>

**`sregex` is not a regex.** `OSMatch` (`src/shared/os_regex/os_match_compile.c`) supports only `|` (alternatives), `^` (prefix), `$` (suffix) and plain substring. There are **no character classes and no wildcards**, and patterns compile case-insensitively (`syscheck-config.c:356`, flags `0`). So `\Performance$` is a literal "ends with `\Performance`" suffix match — the same mechanism as the shipped `\Enum$` entry. It also means `Services\*\Performance` cannot be expressed, and that any `type="sregex"` ignore is **global to every monitored registry entry**, not scoped to `Services`.

**Arch must match.** The ignore check requires `ignore.arch == monitored_entry.arch` (`registry.c:1281-1284`). Both `<windows_registry>` and `<registry_ignore>` default to `32bit` (`src/config/src/syscheck-config.c:449` and `:1948`), so the plain entries above pair correctly with the default `Services` monitor. An ignore added with `arch="64bit"` alone would **silently never match**, and nothing warns about it.

**Ignore the key, not the value.** There is also a `<registry_ignore_value>` tag (`syscheck-config.c:1649`), but it is evaluated per value *after* the value data has been read (`registry.c:1818-1824`), so it is strictly more expensive for the same result. Two side notes on it: it matches `<key path>\<value name>`, not the bare value name as the function's own doc comment claims (`registry.c:1268`); and it is **absent from the documentation** (`docs/ref/modules/fim/configuration.md:288` documents only `registry_ignore`) — worth fixing separately.
</details>

### Fallback if the denylist proves unsustainable

If the enumeration above turns up a long tail of OS-telemetry keys, or new ones keep appearing with Windows releases, bound the scope instead of extending the list:

```diff
-    <windows_registry>HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services</windows_registry>
+    <windows_registry recursion_level="1">HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services</windows_registry>
```

FIM counts depth as the number of path separators below the configured prefix (`registry.c:1228-1256`), and every noisy path above sits at depth ≥ 2 while `Services\<ServiceName>` sits at depth 1 — so this keeps `ImagePath`, `Start`, `Type`, `ObjectName` and `FailureActions` on every service while dropping `bam\State\UserSettings\…` (depth 4), `Tcpip\Parameters\Interfaces\{guid}` (4), `W32Time\SecureTimeLimits\RunTime` (3), and `WmiApRpl\Performance` / `SharedAccess\Epoch` / `Tcpip\Parameters` (2).

> ⚠️ If this path is taken, **do not use `recursion_level="0"`** as the source report suggests: depth 0 is the `Services` key itself, which would drop every per-service key and break the service-binary persistence detection the entry exists for. `1` is the value that keeps it.

### Also update the docs example

`docs/ref/modules/fim/configuration.md:477` shows the `Services` entry as a documented example. It should mirror whatever ships.

### Out of scope — the rule

The generic `Wazuh FIM - Registry modified` rule (`0194adf5-bd82-5ec0-9526-d144ca8ad9ab`) is not in this repository. Assigning T1112 / Defense Evasion to a BAM state key or a DHCP interface key is a real problem, but it belongs to a ruleset/content issue: either a path allow-list, or a separate low/info rule without MITRE for those prefixes. **The rule is doing what it is told** — this issue is about what it is being told.

## How to test

1. Stock Windows 11 agent, 5.0 package, default `ossec.conf` (`notify_first_scan: no`, so first-scan inventory is not a factor).
2. Enroll and leave the host **idle for one hour**. Do not touch the registry.
3. Separately, change a real service's `ImagePath` on a test service, to confirm detection still works.

## How to verify it is fixed

| # | Check | Expected after fix |
|---|---|---|
| 1 | Findings with `registry.path` under `Services\bam\State` | 0 documents |
| 2 | Same for `Services\Tcpip\Parameters\Interfaces`, `Services\W32Time\SecureTimeLimits`, `Services\SharedAccess\Epoch` | 0 documents |
| 3 | Same for any path ending in `\Performance` | 0 documents |
| 4 | Total `Wazuh FIM - Registry` findings over one idle hour | the 42 known-noise events are gone; what remains is the 7 kept on purpose plus the ~45 still to be enumerated |
| 5 | Every remaining `registry.path` has been triaged as OS telemetry or real change | list complete, no unexplained path |
| 6 | Change a real service `ImagePath` → finding still raised on `Services\<name>` | still detected |
| 7 | `Services\<ServiceName>` keys still in the FIM inventory with `ImagePath`, `Start`, `Type` | still present |
| 8 | Non-`Services` defaults unaffected: `Run`, `RunOnce`, `Winlogon`, `KnownDLLs`, `Software\Policies` | unchanged |

Baseline in the reporting lab, 2026-09-02: 97 `wazuh-fim` events, **94** of them registry findings.

Checks 6 and 7 are the ones that matter most: the failure mode of an over-aggressive ignore list is silently losing service-binary persistence detection, which is the reason this entry is in the default at all. Check 5 is what makes this issue closeable — without it the denylist is only known to cover the half of the findings that were written down.

<details>
<summary>What an operator sees today</summary>

Registry findings for agent `windows-11` (id 005), 2026-09-02 02:04–19:26 UTC, rule `0194adf5-bd82-5ec0-9526-d144ca8ad9ab`:

| Registry path | Count |
|---|---|
| `Services\bam\State\UserSettings\S-1-5-18` | 18 |
| `Services\Tcpip\Parameters\Interfaces\{ee540391-…}` | 6 |
| `Services\WmiApRpl\Performance` | 5 |
| `Services\bam\State\UserSettings\S-1-5-19` | 4 |
| `Services\W32Time\SecureTimeLimits` | 4 |
| `Services\W32Time\SecureTimeLimits\RunTime` | 3 |
| `Services\MDCoreSvc` | 3 |
| `Services\Tcpip\Parameters` | 2 |
| `Services\BITS` | 2 |
| `Services\SharedAccess\Epoch` | 2 |
| other `Services` / `Policies` keys | rest of 94 |

One BAM finding, shortened:

```
wazuh.rule.title    Wazuh FIM - Registry modified - …\Services\bam\State\UserSettings\S-1-5-18
wazuh.rule.level    medium
wazuh.rule.mitre    T1112 Modify Registry / TA0005 Defense Evasion
event.action        modified
registry.hive       HKLM
registry.value      \Device\HarddiskVolume3\Windows\System32\sdbinst.exe
registry.data.type  REG_BINARY
```

The collector is `registry_value` and the changed fields are the MD5/SHA hashes of that binary blob. BAM stores "this executable ran for this user"; `S-1-5-18` is SYSTEM and `sdbinst.exe` is a Windows binary. Nothing modified a service.
</details>

<details>
<summary>The shipped default, verified in the repo</summary>

`src/win32/ossec.conf`, `<syscheck>` block. Entries covering `CurrentControlSet` or `Policies`:

```xml
<windows_registry>HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services</windows_registry>
<windows_registry>HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\KnownDLLs</windows_registry>
<windows_registry>HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\SecurePipeServers\winreg</windows_registry>
<windows_registry arch="both">HKEY_LOCAL_MACHINE\Software\Policies</windows_registry>
<windows_registry arch="both">HKEY_LOCAL_MACHINE\Software\Microsoft\Windows\CurrentVersion\Policies</windows_registry>
```

The `Services` entry carries **no `recursion_level` attribute**, so it takes the default: `syscheck-config.c:450` initializes `recursion_level = MAX_REGISTRY_DEPTH`, and `MAX_REGISTRY_DEPTH` is **512** (`src/config/include/syscheck-config.h:144`). The docs confirm the same default (`docs/ref/modules/fim/configuration.md:502`).

The full shipped ignore list:

```xml
<registry_ignore>HKEY_LOCAL_MACHINE\Security\Policy\Secrets</registry_ignore>
<registry_ignore>HKEY_LOCAL_MACHINE\Security\SAM\Domains\Account\Users</registry_ignore>
<registry_ignore type="sregex">\Enum$</registry_ignore>
<registry_ignore>…\Services\MpsSvc\Parameters\AppCs</registry_ignore>
<registry_ignore>…\Services\MpsSvc\Parameters\PortKeywords\DHCP</registry_ignore>
<registry_ignore>…\Services\MpsSvc\Parameters\PortKeywords\IPTLSIn</registry_ignore>
<registry_ignore>…\Services\MpsSvc\Parameters\PortKeywords\IPTLSOut</registry_ignore>
<registry_ignore>…\Services\MpsSvc\Parameters\PortKeywords\RPC-EPMap</registry_ignore>
<registry_ignore>…\Services\MpsSvc\Parameters\PortKeywords\Teredo</registry_ignore>
<registry_ignore>…\Services\PolicyAgent\Parameters\Cache</registry_ignore>
<registry_ignore>HKEY_LOCAL_MACHINE\Software\Microsoft\Windows\CurrentVersion\RunOnceEx</registry_ignore>
<registry_ignore>…\Services\ADOVMPPackage\Final</registry_ignore>
```

Nothing covers `Services\bam\State`, `Services\Tcpip\Parameters\Interfaces`, `Services\*\Performance`, `Services\W32Time\SecureTimeLimits` or `Services\SharedAccess\Epoch` — precisely the keys that changed.
</details>

<details>
<summary>Ruled out</summary>

| Hypothesis | Why it is ruled out |
|---|---|
| First-scan inventory | `notify_first_scan` is `no`; these are live modifications spread over 17 hours |
| FIM is broken | the file-monitoring defaults (Startup, System32 `restrict` lists) are not producing this; the `Services` registry tree is |
| Decoder or rule false match | the writes are real and correctly reported as modifications; the verdict (T1112 / Defense Evasion, medium) is what is wrong, and the monitor scope is why |
| Same as the 5379 vault or 4624 SYSTEM-login findings | different events, different components |
| A misconfigured lab | config is the stock 5.0 Windows `<syscheck><registry>`, unmodified |
</details>

Miguevrgo:
Closes #39274

## Description

The default Windows agent monitors `HKLM\System\CurrentControlSet\Services` at the default recursion level of 512. That subtree is also where Windows writes its own bookkeeping, so on an idle host most `Wazuh FIM - Registry modified` findings are the OS recording that a process ran or that a DHCP lease was renewed — reported at medium level and tagged T1112 Modify Registry / TA0005.

This extends the shipped `<registry_ignore>` list to cover those keys. Key-level ignores are evaluated before `RegOpenKeyEx` (`registry.c:1932`) and match as a case-insensitive prefix (`registry.c:1281-1290`), so one entry skips a whole subtree without enumerating it.

```xml
<registry_ignore type="sregex">\Performance$</registry_ignore>
<registry_ignore>HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\bam\State</registry_ignore>
<registry_ignore>HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\dam\State</registry_ignore>
<registry_ignore>HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\Tcpip\Parameters\Interfaces</registry_ignore>
<registry_ignore>HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\Tcpip6\Parameters\Interfaces</registry_ignore>
<registry_ignore>HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\W32Time\SecureTimeLimits</registry_ignore>
<registry_ignore>HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\SharedAccess\Epoch</registry_ignore>
```

Two entries go beyond the list in the issue, both exact siblings of confirmed noise:

- **`dam\State`** — the Desktop Activity Moderator writes `State\UserSettings\<SID>` exactly as BAM does; it is the same provider and the same kind of value.
- **`Tcpip6\Parameters\Interfaces`** — holds the DHCPv6 per-interface leases, the IPv6 counterpart of the `Tcpip` entry the issue calls out.

`recursion_level` is left at the default. The fallback proposed in the issue (`recursion_level="1"`) would also drop `Services\<name>\Parameters\ServiceDll` at depth 2, which is a service-persistence vector in its own right, so the denylist is the safer of the two.

## Changes

- `src/win32/ossec.conf` — the seven ignore entries above.
- `docs/ref/modules/fim/configuration.md` — documents the matching semantics that make one entry cover a subtree (case-insensitive prefix; `sregex` is `OSMatch`, global to every monitored entry, no character classes or wildcards) and the `arch` pairing rule, plus the new entries in the full-configuration example.
- `docs/ref/modules/fim/windows-registry-monitoring.md` — same semantics, short form.
- `CHANGELOG.md` — Agent / Changed.

## Tests

**Static**, against the modified `src/win32/ossec.conf`:

- The production `OS_ReadXML` (`src/shared/os_xml/`) parses it and returns all 19 `registry_ignore` entries verbatim, `arch=32bit`, backslashes intact.
- The production `OSMatch` compiled with the same flags as `syscheck-config.c:356`: `\Performance$` matches `…\WmiApRpl\Performance` and `…\PERFOS\performance`, and does not match `…\Performance\Sub` or `…\BITS`.
- A model of the scan gate (`fim_registry_configuration` → recursion level → prefix ignore → sregex, per `registry.c:1195-1310`) over 34 paths: 34/34 on this file, 11 failures on the parent commit — all of them the noise rows, every kept row unchanged.

**Live host** — Windows 11 Enterprise 10.0.26200 (the build in the report), agent `wazuh-agent-5.0.0-1`, `syscheck.debug=2`, capture via `(6339) Persisting FIM event`. Two runs of the same scan on the same host, with configurations differing by exactly the entries above:

| Silenced | paths | events |
|---|---|---|
| `\Performance` (sregex) | 31 | 347 |
| `bam\State` | 7 | 37 |
| `Tcpip\Parameters\Interfaces` | 4 | 28 |
| `Tcpip6\Parameters\Interfaces` | 6 | 13 |
| `W32Time\SecureTimeLimits` | 2 | 7 |
| `SharedAccess\Epoch` | 2 | 4 |
| **unaccounted for** | **0** | **0** |

436 of 27,733 registry events (1.6 %). 52 paths disappear, all of them intended; no path appears in the fixed run that was not in the stock one.

Nothing else is lost: **741 distinct `Services\<name>` keys reported in both runs, none missing**, and `W32Time\Config`, `SharedAccess\Parameters`, `Software\Policies`, `CurrentVersion\Run`, `Winlogon` and `KnownDLLs` keep every path.

Two prefix effects worth knowing: `SharedAccess\Epoch` also covers `SharedAccess\Epoch2` (same counter), and `\Performance$` takes `BITS\Performance` along with the other 30 counter keys — that is the `Library` gap noted below. On this host `Services\dam` exists but `dam\State` does not, so that entry matches nothing here.

### Coverage against the source report

The report's P0 list is covered entry for entry: `bam\State`, `Tcpip\Parameters\Interfaces`, `W32Time\SecureTimeLimits`, `SharedAccess\Epoch` and the `\Performance$` sregex — the same five it lists as having no ignore today. `dam\State` and `Tcpip6\Parameters\Interfaces` are added beyond it, as siblings of two of those.

Of the 49 findings the report enumerates, 42 are silenced. The remaining three paths are kept deliberately, and the ~45 it groups as *"other Services / Policies keys"* are not enumerated there, so this list cannot be claimed to cover them. Note that any of those sitting under `HKLM\Software\Policies` or `…\CurrentVersion\Policies` would need `arch="both"` on the ignore, since both monitors are configured that way.

The report's alternative (`recursion_level 0` on `Services` plus explicit child keys) is not taken: depth 0 drops every per-service key, and even depth 1 would drop `Parameters\ServiceDll`. Its P1 (the rule should not assign T1112 to a BAM key) belongs to the ruleset, not here.

### What these ignores give up

| Entry | Values no longer watched | Detection lost |
|---|---|---|
| `\Performance$` | `Library`, `Open`, `Collect`, `Close` on 31 counter keys (`BITS`, `Dnscache`, `Spooler`, `TermService`, `NetBT`, `RemoteAccess`, `ESENT`, `LSM`, `Tcpip`, `usbhub`, the `.NET*` ones…) | `Library` names a DLL loaded into perfmon and WMI host processes; changing it is code execution in another process (T1574). The churn was only `First`/`Last Counter` and `First`/`Last Help`. This is the largest give-up, and being an sregex it applies to every monitored entry ending in `\Performance`, not only under `Services`. |
| `Tcpip\Parameters\Interfaces`, `Tcpip6\…` | `NameServer`, `Domain` alongside the DHCP lease values (`Lease`, `T1`, `T2`, `LeaseObtainedTime`, `DhcpServer`) | Per-adapter DNS redirection, which takes precedence over the global `NameServer` that `Tcpip\Parameters` still watches. |
| `bam\State`, `dam\State` | executable paths per SID (`powershell.exe`, `wevtutil.exe`, `explorer.exe`…) | Execution history, not configuration. What is lost is noticing that someone *cleared* BAM entries to hide execution (T1070). `Services\dam` exists on the test host but `dam\State` does not, so that entry currently matches nothing. |
| `W32Time\SecureTimeLimits` | `SecureTimeHigh`, `SecureTimeLow`, `SecureTimeConfidence`, `SecureTimeTickCount` | Tampering with the secure-time bounds that back certificate and time-based validation. |
| `SharedAccess\Epoch` | a change counter (the prefix also covers `Epoch2`) | Effectively nothing: the firewall policy itself, `SharedAccess\Parameters\FirewallPolicy`, stays monitored and is what carries the rules. |

### Kept noisy on purpose

`Tcpip\Parameters` (2 findings) — DHCP writes `DhcpNameServer` here, but so live `NameServer`, `Domain` and `SearchList`. `BITS` (2) — BITS hijacking is T1197. `MDCoreSvc` (3) — Defender's own service key. All three are service keys the monitor exists for; the residual noise is the cheaper trade.

### On the evidence above

The report is explicit that these are live modifications, not first-scan inventory. The counts measured here are inventory. They carry over because the ignore is evaluated before `RegOpenKeyEx` (`registry.c:1932`): a key that is never opened produces neither inventory nor modifications, so a path absent from the inventory is out of the scan entirely.


CHANGELOG.md
```diff
@@ -143,6 +143,7 @@
 | [#35330](https://github.com/wazuh/wazuh/issues/35330) | Increased default limits for agent event throughput and inventory message sizes. |
 | [#35880](https://github.com/wazuh/wazuh/issues/35880) | Reduced `wazuh-agent` Debian package dependencies, removed `adduser`, `lsb-release`, and `debconf`. |
 | [#35471](https://github.com/wazuh/wazuh/issues/35471) | Standardized agent-start and buffer-status events to a WCS-aligned JSON format. |
+| [#39274](https://github.com/wazuh/wazuh/issues/39274) | Extended the Windows agent's default FIM registry ignore list to exclude the OS telemetry under `HKLM\System\CurrentControlSet\Services`. |
```
MarcelKemp: **[S4 · non-blocking · introduced]** This entry sits under `### Agent` / `#### Changed`, but from a user's perspective it's a fix for false-positive alerts (OS telemetry misreported as registry-modified events under T1112). Consider moving the line as-is to `#### Fixed` in the Agent section instead — no wording change needed.

MarcelKemp (approved):

PR #39298, opened 2026-09-15
Replace shipped default API passwords with generated, disclosed credentials

### LINKED ISSUE #39284: Generate random passwords for the default API users (`wazuh`, `wazuh-wui`) at installation
## Description

The manager ships two Server API users, and **both are created with a password equal to their own
username**:

| User | ID | Password today | Consumer |
|---|---|---|---|
| `wazuh` | 1 | `wazuh` | Operators, scripts, API clients |
| `wazuh-wui` | 2 | `wazuh-wui` | The Wazuh dashboard, to reach the Server API on port 55000 |

Both are linked to the `administrator` role, the credentials are public (they are literals in
[framework/wazuh/rbac/default/users.yaml:6,16](framework/wazuh/rbac/default/users.yaml#L6) and are
documented in the installation guide), and the API **starts and serves normally** with them. Anyone
who can reach port 55000 on a fresh installation has full administrator access to the manager until
the operator changes the passwords by hand. The only current mitigation is a `WARNING` line at
startup ([api/scripts/wazuh_manager_apid.py:97-112](api/scripts/wazuh_manager_apid.py#L97-L112)).

We want every installation to end up with **unique, randomly generated, policy-compliant passwords**
for both users, disclosed to the operator in a secure way, and propagated to the Wazuh dashboard so
that it keeps working out of the box — in **all** installation flows (packages, sources, installation
assistant all-in-one, installation assistant distributed, containers, Kubernetes) and in **all**
cluster topologies.

### How it works today

- `rbac.db` (`/var/wazuh-manager/api/configuration/security/rbac.db`) is **not** created by any
  installer. It is created and seeded lazily on the **first `wazuh-manager-apid` start**, by
  `check_database_integrity()`
  ([framework/wazuh/rbac/orm.py:2461-2548](framework/wazuh/rbac/orm.py#L2461-L2548)) →
  `DatabaseManager.insert_default_resources()`
  ([orm.py:2092-2099](framework/wazuh/rbac/orm.py#L2092-L2099)), which reads the plaintext
  `password:` values from `users.yaml` and hashes them with Werkzeug scrypt.
- No installation script in this repo generates any API credential. The only random-secret precedent
  is the cluster key at
  [src/init/inst-functions.sh:995-999](src/init/inst-functions.sh#L995-L999).
- The password policy (12–64 characters, at least one uppercase, one lowercase, one digit and one
  symbol) is enforced only in [framework/wazuh/security.py:23](framework/wazuh/security.py#L23);
  reserved users (ID ≤ 99) can only be modified by another reserved user
  ([security.py:232-240](framework/wazuh/security.py#L232-L240), error `5011`).
- The only first-party tooling to change them is
  [`bin/rbac_control change-password`](framework/scripts/rbac_control.py) (interactive prompt,
  `--password-file`, `--passwords-file`; never accepts a password as a command-line argument).
  `rbac_control factory-reset` deletes the DB and **restores the shipped defaults**.
- `wazuh-manager-apid` runs on the **master node only**
  ([src/init/wazuh-server.sh:193,392](src/init/wazuh-server.sh#L193)), and `rbac.db` is **not**
  cluster-synchronized: every node keeps its own copy with the shipped defaults until it becomes
  master.
- The dashboard holds the only external copy of the `wazuh-wui` credential, in
  `wazuh_core.hosts.<host>.password` inside `/etc/wazuh-dashboard/opensearch_dashboards.yml`, read
  once at startup (in-tree sample:
  [api/tools/env/wazuh-dashboard/wazuh.dashboard.yml:44-54](api/tools/env/wazuh-dashboard/wazuh.dashboard.yml#L44-L54)).
  Container images take it from the `API_USERNAME` / `API_PASSWORD` environment variables.
- `wazuh-install.sh` and `wazuh-passwords-tool.sh` are **not** in this repository — they live in
  `wazuh/wazuh-installation-assistant`, which is where the all-in-one and distributed flows are
  orchestrated.

### Proposed approach

1. **Generate at RBAC seed time (framework).** Remove the `password:` keys from `users.yaml` (keep
   `description` and `allow_run_as`) and have `insert_default_resources()` produce a random password
   per default user through a new helper (e.g. `generate_default_password()` in
   `framework/wazuh/core/security.py`), built on `secrets` and guaranteed to satisfy `_user_password`
   ([security.py:23](framework/wazuh/security.py#L23)). Since seeding is lazy and centralized, this
   single code path covers **every** installation flow: DEB, RPM, sources, Docker, devcontainer and
   QA environments. No new hook is needed in `install.sh`, `postinst` or the RPM spec.

2. **Provisioning escape hatch.** When a pre-seed credentials input is present at DB-creation time
   (a file under `api/configuration/security/`, or an equivalent environment input), use those
   passwords instead of generating new ones. This is what lets the installation assistant, the
   container images and CI provision deterministic credentials non-interactively. Defining its exact
   path, format, ownership and lifetime is part of this issue.

3. **Disclose once, securely.** Write the generated passwords to
   `WAZUH_HOME/api/configuration/security/wazuh-api-passwords.txt` with the most restrictive
   ownership and mode the apid runtime user allows (target `0400`), and log a `WARNING` at first
   start pointing at that file and instructing the operator to read it and then delete it. **The
   password itself must never appear in the manager log, in the installer output, or in `ps`.**

4. **Rework the default-password warning.** `get_users_with_default_password()`
   ([framework/wazuh/core/security.py:137-150](framework/wazuh/core/security.py#L137-L150)) and
   `warn_about_default_passwords()` lose their meaning once there is no shipped password. Replace the
   check with "the generated-credentials file is still on disk", keeping the pointer to
   `rbac_control change-password`.

5. **`factory-reset` must not restore a known password.** `rbac_db_factory_reset()`
   ([core/security.py:125-134](framework/wazuh/core/security.py#L125-L134)) has to regenerate and
   re-disclose, exactly like a fresh install.

6. **Cluster.** Each node seeds its own `rbac.db`, so with random generation a promoted worker would
   hold a *different* password and silently break the dashboard (`3002 - Request failed with status
   code 401`, then `403` after `max_login_attempts`). We do **not** add credential synchronization
   over the cluster protocol: instead, multi-node installations must push one common password to
   every node with `rbac_control change-password --passwords-file`, and this must be documented and
   verified, including that the tooling works on a node whose apid has never run.

7. **Dashboard and cross-repo wiring.** The manager exposes the credentials (generated file) and
   accepts injected ones; the installation assistant reads them on an all-in-one host and
   injects/pushes them on a distributed deployment, then writes `wazuh_core.hosts.<host>.password` in
   the dashboard configuration. Container and Kubernetes deployments keep using
   `API_USERNAME`/`API_PASSWORD` and the `wazuh-api-cred` Secret.

### Installation cases that must be analysed and tested

- **Packages**: DEB fresh install, RPM fresh install, sources `install.sh`.
- **Reinstall / upgrade over an existing `rbac.db`**: passwords must be preserved, never regenerated
  (DEB carries the file over in `preinst`/`postinst`, RPM simply does not own it).
- **Topologies**: single node; cluster master + worker; **worker promoted to master**.
- **Installation assistant, all-in-one**: manager + indexer + dashboard on the same host.
- **Installation assistant, distributed**: dashboard on a different host from the manager.
- **Containers**: `api/tools/env/` docker-compose and the devcontainer e2e dashboard image
  ([src/engine/tools/devContainer/e2e/wazuh-dashboard/Dockerfile:17-20](src/engine/tools/devContainer/e2e/wazuh-dashboard/Dockerfile#L17-L20)),
  which today rely on the shipped defaults.
- **Kubernetes**: dashboard fed from the `wazuh-api-cred` Secret.
- **`rbac_control factory-reset`** and `rbac_control change-password` in all of the above.

### Impact on existing tests

Roughly fifteen suites hard-code these credentials and will break; they must be migrated to the new
provisioning mechanism. The main ones:
[api/test/integration/common.yaml:14-17](api/test/integration/common.yaml#L14-L17) (including the
base64 `basic_auth` / `basic_auth_context` blobs), the RBAC/security tavern suites, the env SQL
fixtures under `api/test/integration/env/configurations/`,
[tests/integration/test_api/test_miscs/test_recursion.py:56-57](tests/integration/test_api/test_miscs/test_recursion.py#L56-L57),
`tests/integration/test_api/test_miscs/test_run_as_chunked_body.py:58-59`, and
[api/tools/env/wazuh-dashboard/wazuh.dashboard.yml:52-53](api/tools/env/wazuh-dashboard/wazuh.dashboard.yml#L52-L53).

### Manager (this repository)

- [ ] Add a random password generator (`secrets`-based) that always satisfies the API password policy
      at [framework/wazuh/security.py:23](framework/wazuh/security.py#L23), with unit tests asserting
      compliance and uniqueness across invocations.
- [x] Remove the plaintext `password:` entries from
      [framework/wazuh/rbac/default/users.yaml](framework/wazuh/rbac/default/users.yaml) and generate
      the password per default user in `insert_default_resources()`
      ([orm.py:2092-2099](framework/wazuh/rbac/orm.py#L2092-L2099)).
- [x] Design and implement the pre-seed/injection input for automated provisioning (path, format,
      ownership, who writes it, when it is consumed and removed).
- [ ] Write the generated credentials to a restricted-permission file and log a `WARNING` pointing to
      it; guarantee the secret is never logged nor exposed through the process list.
- [ ] Rework `get_users_with_default_password()` and `warn_about_default_passwords()` to warn about an
      undeleted credentials file instead of a shipped default password.
- [ ] Make `rbac_db_factory_reset()` regenerate and re-disclose new random passwords.
- [x] Verify `rbac_control change-password` still satisfies the reserved-user rule (`5011`) with the
      new flow, and that it can set credentials on a node whose apid has never started (worker);
      extend the tool if it cannot.
- [ ] Confirm package upgrade paths preserve the credentials (DEB `preinst`/`postinst` copy of
      `rbac.db`, RPM spec not owning the file) and that nothing regenerates on restart.
- [x] Migrate every test fixture and suite that hard-codes `wazuh:wazuh` / `wazuh-wui:wazuh-wui`
      (tavern `common.yaml` + base64 blobs, env SQL fixtures, `test_recursion.py`,
      `test_run_as_chunked_body.py`, `api/tools/env/wazuh-dashboard/wazuh.dashboard.yml`, devcontainer
      e2e dashboard image) to the provisioning mechanism.
- [ ] Add unit and API integration tests for: generation, disclosure file (content and permissions),
      credential injection, no-regeneration on restart, and factory-reset.
- [x] Update the documentation: `docs/ref/getting-started/installation.md:189-271` ("Change the
      default API passwords") and `docs/ref/modules/server-api/authentication.md:42-91` (default
      users, what a password change does, cluster/worker-promotion behaviour).
- [x] Add the `CHANGELOG.md` entry.

### Cross-repository (open as linked child issues)

- [ ] `wazuh/wazuh-installation-assistant` — consume or inject the credentials, write
      `wazuh_core.hosts.<host>.password` in the dashboard configuration, keep `wazuh-passwords-tool.sh`
      working, and report/store the generated passwords for both the **all-in-one** and the
      **distributed** flows.
- [ ] `wazuh/wazuh-dashboard`, `wazuh/wazuh-docker`, `wazuh/wazuh-kubernetes` — confirm the credential
      sources (`opensearch_dashboards.yml`, `API_USERNAME`/`API_PASSWORD`, `wazuh-api-cred` Secret)
      still work and document the change.

### Validation

- [ ] Execute the full installation-case matrix listed above and attach evidence per case.
- [ ] For every case, confirm that `wazuh:wazuh` and `wazuh-wui:wazuh-wui` are rejected with `401`,
      that the provisioned credential authenticates, and that the dashboard reaches the API.

## Definition of Done

- [x] No installation flow ends with a predictable API password; the default passwords are gone from
      the repository.
- [ ] Every fresh installation gets unique, policy-compliant passwords for `wazuh` and `wazuh-wui`;
      restarts and package upgrades over an existing `rbac.db` preserve them; `factory-reset` produces
      new random ones.
- [ ] The generated passwords are disclosed exactly once, through a restricted-permission file, never
      in logs, installer output or `ps`, with a clear operator message on how to read, use and remove
      it.
- [ ] The Wazuh dashboard authenticates against the Server API out of the box in **all-in-one** and in
      **distributed** installations; container and Kubernetes deployments keep working through their
      environment/Secret path.
- [ ] Automated and unattended provisioning (installation assistant, Docker, CI) can inject
      credentials non-interactively, without ever passing them on a command line.
- [x] Cluster behaviour is documented and verified, including the worker-promotion scenario, with a
      documented procedure to align credentials across nodes.
- [x] Every test suite that hard-coded the default credentials is updated and green in CI; new unit and
      integration tests cover generation, disclosure, injection, no-regeneration and factory-reset.
- [ ] E2E evidence attached for every case of the installation matrix (install output, credentials file
      permissions, `401` on the old defaults, successful dashboard login, manager logs).
- [x] Documentation updated and `CHANGELOG.md` entry added.
- [ ] Cross-repository child issues opened, linked and resolved.
- [ ] Reviewed and merged into `5.0.0`.

ignaciogalle12git:
## Description

Both default Server API users, `wazuh` (id 1) and `wazuh-wui` (id 2), ship with a password equal to
their own username, in plaintext, in `framework/wazuh/rbac/default/users.yaml`. Both are linked to
the `administrator` role, and the API starts and serves normally with them. Anyone who can reach
port 55000 on a fresh installation has full administrator access until the operator changes the
passwords by hand.

This PR removes the shipped literals and gives every installation a unique, randomly generated
password per default user instead, disclosed once through a restricted-permission file.

Closes #39284

## Proposed Changes

**Generation.** `generate_default_password()` draws 24 characters from `secrets` and redraws the
whole candidate until it satisfies the password policy, so the result is uniform over the compliant
subset and its entropy is exactly that of the alphabet: 145 bits, above the 112-bit floor of NIST SP
800-63B-4 and the 128-bit floor of OWASP ASVS 5.0. The alphabet is deliberately narrower than the
policy allows (alphanumerics plus `-_.+=`, with an alphanumeric first character): the value crosses a
YAML scalar, a JSON string, a shell word and a `sed` replacement before it reaches whoever
provisions it, and a character that is special in any of those is a provisioning failure rather than
a stronger password. The design notes, with the standards behind each choice, are in the issue.

The policy pattern moves to `orm.py` and is anchored with `\Z` instead of `$`, which also matched
before a trailing newline and let a password read from a file keep its line break.

**Disclosure.** The generated passwords are written once to
`api/configuration/security/wazuh-api-passwords.txt`, mode `0400`, renamed over the target from a
sibling temporary file so a failure never leaves the installation with no disclosure at all. A
user's entry is dropped as soon as its password changes. `rbac.db` stores a scrypt hash, so once
that file is gone the password cannot be recovered; `rbac_control change-password` sets a new one
instead, and nothing regenerates on its own, which would rotate the credential the dashboard holds
without anyone asking.

**Provisioning.** An installer can write `api/configuration/security/wazuh-preseeded-passwords.json`
before the first API start to provision known credentials non-interactively. Naming a single default
user is valid and the other one still gets a generated password; a key that names no default user is
rejected. The file must be readable by the user `wazuh-manager-apid` runs as, owned by `root` or the
Wazuh user, and not writable by its group or by others: it decides an administrator password and
lives in a group-writable directory. It is stat'd through the descriptor that gets read, so the path
cannot be swapped between the check and the read.

A file that is present but unusable stops the API and leaves no database behind, rather than
degrading to a generated password that whoever wrote the file would never learn about, which would
surface much later as a `401`. It is left on disk so it can be corrected. One that is applied is
removed, so a plaintext administrator credential does not outlive the installation step, and one
found when `rbac.db` already exists is reported and removed too, because it can no longer take
effect.

**Cluster.** `rbac_control change-password` gains `--local`, which routes as `local_any`. A
`local_master` request made on a worker executes on the master, so until now the command silently
retargeted the node the operator did not ask for and still printed `UPDATED` — and the procedure
`docs/ref/modules/server-api/authentication.md` already recommended, repeating the change on every
node that may become master, was not executable. A worker's `rbac.db` is never synchronized and only
becomes live on promotion.

**Upgrades.** A package upgrade over an existing `rbac.db` keeps the current password: the RBAC
migration already restores the default users' rows. That also means an installation created before
this change keeps `wazuh`/`wazuh`, so `wazuh-manager-apid` now warns about the shipped password
itself, separately from the disclosure-file warning.

**Failure paths.** Losing `rbac.db` reseeds it and rotates both passwords; that is reported as a
rotation rather than logged like a first install, distinguished by a marker written beside the first
disclosure. A migration whose source cannot be read restores nothing, so the passwords generated for
the throwaway database survive: those are now detected and disclosed instead of left in service
known to nobody. A disclosure that fails right after seeding removes the database again so the next
start redoes the whole seeding, but only on the fresh-install path, since a migration has already
replaced the previous database and its contents cannot be recovered.

Out of scope, tracked as linked child issues: wazuh-installation-assistant#1025,
wazuh-docker#2674, wazuh-kubernetes#1660, wazuh-virtual-machines#973 and wazuh-ansible#2306.

### Results and Evidence

Unit tests, `PYTHONPATH=api:framework` per `5_testunit_framework.yml`, pinned dependencies from
`framework/requirements.txt`:

```
framework/wazuh/rbac/tests/test_orm.py
framework/wazuh/core/tests/test_security.py
framework/wazuh/tests/test_security.py
framework/scripts/tests/test_rbac_control.py

178 passed
```

The three assertions that pin the alphabet, the first-character rule and the default length were
checked by mutation: reverting the generator to the previous alphabet and length makes all three
fail, and they pass again once restored.

`--local` verified on an installed manager, with the node switched to `worker`. Note that
`wazuh-manager-apid` does not start on a worker, which is why a worker has no `rbac.db` until it is
promoted:

```
════ wazuh-wui hash, before ════
   scrypt:32768:8:1$YF5VipNUyZN

════ change-password on a worker, without --local ════
	Could not ensure the RBAC database exists: Error 3023 - Worker node is not connected to master
   hash after: scrypt:32768:8:1$YF5VipNUyZN      <- unchanged

════ change-password on a worker, with --local ════
	wazuh-wui: UPDATED
   hash after: scrypt:32768:8:1$aOwV4LPUtIe      <- changed
```

That environment has a single node, so the request had no master to reach and failed loudly. In a
cluster with a live master the same command succeeds against the master's database while leaving the
local one untouched, which is the behaviour `--local` exists to correct.

The corrupt-database path was reproduced before and after the fix. A zero-byte `rbac.db` counts as
existing, so `pragma user_version` returns 0 and the migration branch runs rather than the
fresh-install one:

```
rbac.db file was detected
RBAC database migration required. Current version is 0 but it should be 1
rbac.db database upgraded successfully          <- reports success

  check_database_integrity() returned: {}
  disclosure file exists: False
  user 1 (wazuh)     authenticates with the historical literal? False
                     authenticates with anything disclosed?     False
```

After the fix the same input returns both passwords for disclosure and logs why they could not be
preserved.

Every shape of pre-seed file, driven through `check_database_integrity()` on a fresh installation.
The path is elided; nothing logs a password:

```
pre-seed                     rbac.db   file left   outcome
valid, both users            True      False       seeded, 0 generated password(s)
valid, only one user         True      False       seeded, 1 generated password(s)
password below the policy    False     True        refused: the pre-seeded password of 'wazuh' in '<file>' does not satisfy the API password policy
unknown user name            False     True        refused: '<file>' names 'wazuh-wu', which is not a default user. Default users: wazuh, wazuh-wui
malformed JSON               False     True        refused: '<file>' is not valid UTF-8 JSON: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
not UTF-8                    False     True        refused: '<file>' is not valid UTF-8 JSON: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
group-writable               False     True        refused: '<file>' is writable by its group or by others
no pre-seed at all           True      n/a         seeded, 2 generated password(s)
```

  - N/A (no compiled code changed)

- Memory tests for Linux
  - N/A (no compiled code changed)

  - N/A

  - N/A (no engine code changed)

- Wazuh server API/Framework

### Artifacts Affected

- `wazuh-manager-apid` (`api/scripts/wazuh_manager_apid.py`).
- `bin/rbac_control`, packaged by `packages/rpms/SPECS/wazuh-manager.spec`.
- Framework package: `framework/wazuh/rbac/orm.py`, `framework/wazuh/security.py`,
  `framework/wazuh/core/security.py`.
- `framework/wazuh/rbac/default/users.yaml`.
- Debian `postrm`, which no longer preserves the plaintext credential files as `.save`.
- Three runtime files under `api/configuration/security/`, none of them packaged: the disclosure file
  and the generation marker, written by the API, and the optional pre-seed input, written by the
  installer and removed by the API once used.

### Configuration Changes

No `wazuh-manager.conf` or schema changes. Behavioural change to installation-time defaults:
`users.yaml` no longer carries a `password:` key, and installers gain the optional
`wazuh-preseeded-passwords.json` input.

### Tests Introduced

- `test_orm.py`: generation compliance and uniqueness, the alphabet, the alphanumeric first
  character, the default length, the minimum length repeated so rejection sampling is exercised
  where it retries most, the pre-seed loader (accepted, partial, policy-violating, unknown user,
  foreign owner, group- and other-writable, non-UTF-8, malformed, absent), that an unusable file
  stops a fresh installation before any database is created, that the file is consumed once applied,
  and each branch of `check_database_integrity`.
- `core/tests/test_security.py`: disclosure content, mode and atomic replacement, the marker and the
  rotation warning, that clearing an entry does not warn, `clear_stale_disclosure`,
  `ensure_rbac_database` and `rbac_db_factory_reset`.
- `scripts/tests/test_rbac_control.py`: the ensure-database step and `--local` routing.

## Review Checklist

- [x] Code changes reviewed


ignaciogalle12git:
Superseded by #39409. The design changed: the manager no longer generates the passwords, it takes them from a credentials file the installation provisions alongside the certificates, and falls back to the shipped literals when there is none. That work is on `issue39284-preseed-via-certs-tool`, so it needed a new pull request: a pull request's head branch cannot be changed.

PR #39299, opened 2026-09-15
Close the remaining metrics snapshot task gaps: startup delay, observability, and a real integration test

### LINKED ISSUE #39283: Metrics snapshot task is not indexing metrics
## Description

The manager is not writing any document into the metrics indices:

- `wazuh-metrics-agents`
- `wazuh-metrics-comms-v4`
- `wazuh-metrics-normalization`

These indices are fed by the metrics snapshot task that runs inside the **master node of clusterd**:
`MetricsSnapshotTasks.run_metrics_snapshot()` in
[framework/wazuh/core/indexer/metrics_snapshot.py](framework/wazuh/core/indexer/metrics_snapshot.py),
registered as an indexer task in
[framework/wazuh/core/cluster/master.py:1058-1062](framework/wazuh/core/cluster/master.py#L1058-L1062).

The whole collection → normalization → validation → bulk-index chain needs to be reviewed and
fixed, and the fix has to be backed by a full end-to-end validation (manager running, agents
reporting, documents queryable in the indexer).

### Pipeline under review

1. **Scheduling** — `Master.__init__` builds `MetricsSnapshotTasks` and appends
   `run_metrics_snapshot` to `indexer_tasks`; the whole list is only started when
   `get_manager_conf(section="indexer")["indexer"]["hosts"]` is non-empty
   ([master.py:1052-1070](framework/wazuh/core/cluster/master.py#L1052-L1070)). Any exception while
   reading the indexer configuration silently disables **all** indexer tasks with a single
   `error` log line.
2. **Interval** — `metrics_frequency` / `metrics_bulk_size` are read from the `intervals.master`
   section of [framework/wazuh/core/cluster/cluster.json](framework/wazuh/core/cluster/cluster.json#L74-L75)
   (defaults 600 s / 100). `run_metrics_snapshot` sleeps **before** the first collection, so
   nothing is ever indexed during the first 10 minutes of uptime, and `metrics_frequency=0`
   disables the task entirely.
3. **Collection**
   - agents: `wdb_http_client.get_all_agents()`
   - comms: `get_daemons_stats(["wazuh-manager-remoted"])` locally + DAPI fan-out to workers
   - normalization: `get_engine_metrics()` (analysisd/engine HTTP socket) locally + DAPI fan-out
   Every per-node failure is swallowed by a `try/except` that only logs, so a broken source
   yields an empty document list and a completely silent no-op cycle.
4. **Normalization** — `_normalize_agent_doc`, `_normalize_comms_doc`,
   `_normalize_normalization_doc` + `_drop_none`.
5. **Schema validation** — `_load_schema()` reads the index templates from
   `WAZUH_HOME/etc/indexer-plugins/{metrics-agents,metrics-comms,metrics-normalization}.json`
   (downloaded by `make deps`, installed by
   [src/init/inst-functions.sh:1457](src/init/inst-functions.sh#L1457), packaged in
   [packages/rpms/SPECS/wazuh-manager.spec:589-592](packages/rpms/SPECS/wazuh-manager.spec#L589-L592)).
   Documents that fail validation are **dropped**, one `error` log per document — a template/field
   mismatch therefore produces exactly the observed symptom (task running, zero documents indexed).
6. **Indexing** — `MetricsIndex.bulk_index()` in
   [framework/wazuh/core/indexer/metrics.py](framework/wazuh/core/indexer/metrics.py) issues an
   `async_bulk` with `_op_type: create` and `raise_on_error=False`, so rejected documents never
   raise. Note the index names are hardcoded at the call site
   ([metrics_snapshot.py:725-736](framework/wazuh/core/indexer/metrics_snapshot.py#L725-L736)) and
   must match the data streams / index templates actually created by the indexer — including the
   `-v4` suffix of `wazuh-metrics-comms-v4`.

### Hypotheses to confirm or discard

- Indexer tasks never started (empty/unreadable `indexer.hosts`, or an exception in the
  `try` block of `Master.__init__`).
- The task raises once per cycle and only leaves an `exception` trace in `cluster.log`.
- Data streams / index templates for the three indices do not exist in the indexer, or their
  names do not match the hardcoded ones (notably `wazuh-metrics-comms-v4`).
- All documents are dropped in `_validate_documents` because the generated document shape no
  longer matches the `wazuh-indexer-plugins` templates (strict mappings).
- The schema files are missing under `etc/indexer-plugins/` (validation skipped) while the
  indexer rejects documents on ingest with `raise_on_error=False` hiding the rejection.
- Collection sources return nothing: wazuh-db `/agents/all`, remoted stats, or the engine metrics
  dump (engine metrics may be disabled by default).
- Bulk failures are not visible: `bulk_index` logs `%d failed` while `async_bulk` with
  `stats_only=False` returns a **list** of errors, so the warning itself can fail to render.
- Worker-node documents lost in the DAPI fan-out (`failed_items` only warned, never retried).

## Tasks

- [ ] Reproduce on a running manager (single node and cluster master+worker) and confirm the three
      indices receive zero documents.
- [ ] Verify the task is actually scheduled: confirm `indexer.hosts` is read correctly and that
      `manage_indexer_tasks` starts `run_metrics_snapshot`; add an explicit log line when the
      metrics snapshot task starts and when a cycle completes (docs collected / validated /
      indexed per index).
- [ ] Inspect `cluster.log` for `Metrics snapshot failed`, `Document failed schema validation`,
      `Metrics schema ... not found`, `Metrics bulk index on ...` and DAPI warnings; make these
      messages actionable.
- [ ] Verify the three data streams / index templates exist in the indexer and that their names
      match the ones hardcoded in `_collect_and_index()` (check the `-v4` suffix of
      `wazuh-metrics-comms-v4` against `wazuh-indexer-plugins`).
- [ ] Verify the schema files are downloaded by `make deps` and installed under
      `WAZUH_HOME/etc/indexer-plugins/` with readable permissions for the `wazuh-manager` user
      (DEB/RPM/sources).
- [ ] Validate each collector in isolation: `get_all_agents()` (wazuh-db HTTP), remoted stats via
      `get_daemons_stats`, engine metrics via `get_engine_metrics` — including the worker path
      through DAPI.
- [ ] Validate each normalizer's output against the current index templates and fix field
      mismatches (names, types, `@timestamp` format, dropped `None`s).
- [ ] Fix the bulk-index error reporting (`async_bulk` returns a list of errors, not a count) and
      log the actual rejection reasons returned by the indexer.
- [ ] Review the first-cycle delay: the task sleeps `max(frequency, 600)` **before** the first
      collection; decide and implement whether an initial snapshot should be taken at startup.
- [ ] Implement the fix(es) for the confirmed root cause(s).
- [ ] Add/extend unit tests in `framework/wazuh/core/indexer/tests/test_metrics_snapshot.py` and
      `test_metrics.py` covering the failure modes found (empty collection, validation drop, bulk
      rejection, worker fan-out failure).
- [ ] Add/extend the integration test
      `framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py` to assert documents
      actually land in the three indices.
- [ ] Run a full E2E validation on a real deployment (master + worker + agents + indexer) and
      attach evidence: queries and document counts per index, sample documents, cluster logs.
- [ ] Update documentation if behaviour, index names or configuration change
      (`docs/`, migration guide entries for `metrics_frequency` / `metrics_bulk_size`).
- [ ] Add the `CHANGELOG.md` entry.

## Definition of Done

- [ ] Root cause identified and documented in this issue (what exactly prevented the documents
      from being indexed, at which stage of the pipeline).
- [ ] The three indices `wazuh-metrics-agents`, `wazuh-metrics-comms-v4` and
      `wazuh-metrics-normalization` receive documents on every snapshot cycle on a running manager.
- [ ] Documents indexed conform to the `wazuh-indexer-plugins` templates: correct index/data
      stream names, correct field names and types, valid `@timestamp`, cluster name and node
      present in every document.
- [ ] Cluster deployment verified: documents from the master **and** from every worker node reach
      the three indices (DAPI fan-out working, one document set per node).
- [ ] No silent failures left: every collection, validation and bulk-index failure produces a log
      message that identifies the index, the node and the reason; the bulk error reporting prints
      real indexer rejection reasons.
- [ ] E2E evidence attached to the issue/PR: index document counts before and after, sample
      documents per index, and the relevant `cluster.log` excerpts.
- [ ] Unit tests and integration tests added/updated and green in CI.
- [ ] `CHANGELOG.md` entry added; documentation updated if applicable.
- [ ] Code reviewed and merged into `5.0.0`.

Antoniogm03:
## Description

Investigating #39283 ("Metrics snapshot task is not indexing metrics") on a live 5.0.0 manager (devcontainer: 1 manager, 1 indexer, 1 dashboard, 4 agents — 2x v5.0.0, 2x v4.14.3) turned up the blocking root cause outside this PR's scope: `get_indexer_client()` (`framework/wazuh/core/indexer/indexer.py`) raised `AttributeError: 'str' object has no attribute 'get'` on every call, because it still parsed `indexer.ssl.certificate_authorities`/`certificate`/`key` in the shape the old config parser produced, not the shape `get_manager_conf()` has returned since the config loader migration. That was already fixed and merged separately by #39184 (issue #39113); this branch is rebased on top of it.

Investigating and fixing that surfaced several more defects in the same metrics-snapshot pipeline, none touched by #39184. This PR fixes all of them:

- `MetricsIndex.bulk_index()` (`metrics.py`) formatted `async_bulk()`'s return value with `%d indexed, %d failed`, but with the default `stats_only=False`, the second value is a **list** of per-document OpenSearch error dicts, not a count. Formatting a list with `%d` raises inside Python's own `logging` module (`Handler.emit()` → `handleError()`), which does not propagate but does print a "--- Logging error ---" dump to `sys.stderr` — and the real `wazuh-manager-clusterd` daemon has `stderr` redirected to `/dev/null`. Net effect: any time the indexer rejects a metrics document, the warning that would explain why is generated and silently discarded.
- `_normalize_agent_doc()` (`metrics_snapshot.py`) emitted `wazuh.agent.config.hash.md5` (from `configSum`) and `wazuh.agent.host.os.full` (from `os.uname`). Neither field exists in the `wazuh-metrics-agents` index template, which uses `dynamic: strict` — an unmapped field rejects the whole document, not just that field. Neither source value is ever populated by `/agents/all` in 5.x, so both always landed as `None` and were dropped by `_drop_none()` before indexing. They never caused the reported symptom, but were a live landmine: if either source were ever populated again, the document would be rejected outright, and that rejection would itself have been swallowed by the bug above.
- `run_metrics_snapshot()` slept a full `metrics_frequency` interval (10 minutes by default) before its first collection, so a manager restart left the three indices empty for that long even though agents, comms stats and engine metrics were available immediately.
- `_collect_and_index()` gave no visibility into what a cycle actually did: no log line marked a cycle starting or finishing, and no counts were reported anywhere.
- `_collect_agents()` had no error handling, unlike the other two collectors: a wazuh-db failure propagated through `_collect_and_index()`'s `asyncio.gather()` and aborted comms and normalization collection too, not just agents.
- `manage_indexer_tasks()` (`framework/wazuh/core/cluster/common.py`) reduced every failure, transient or deterministic, to the same terse `str(e)` warning with no traceback — the exact blind spot that let the #39113 bug go unnoticed for as long as it did.
- `test_metrics_snapshot_integration.py`'s only "integration" coverage connected to a plain, unauthenticated, non-TLS OpenSearch (a configuration this pipeline never runs against) and bulk-indexed ad-hoc documents into a scratch index unrelated to the real templates — it did not exercise the real pipeline or the real data streams at all.

## Proposed Changes

- `framework/wazuh/core/indexer/metrics.py`: `bulk_index()` logs `len(failed)` and the first few real rejection reasons (`failed[:5]`) instead of formatting the error list with `%d`, and returns the indexed count instead of `None`.
- `framework/wazuh/core/indexer/metrics_snapshot.py`:
  - `_normalize_agent_doc()` no longer emits `wazuh.agent.config` or `wazuh.agent.host.os.full`.
  - `run_metrics_snapshot()` runs the first cycle immediately on start, before the first sleep; the interval between subsequent cycles is unchanged.
  - `_collect_and_index()` logs a start message and, at the end of every cycle, the number of documents collected, schema-validated and indexed per index.
  - `_collect_agents()` catches its own failures and returns an empty list, matching the other two collectors' per-source isolation.
- `framework/wazuh/core/cluster/common.py`: `manage_indexer_tasks()` logs the full traceback at debug level alongside its existing warning.
- `framework/wazuh/core/indexer/tests/test_metrics.py`, `test_metrics_snapshot.py`, `framework/wazuh/core/cluster/tests/test_common.py`: new and updated tests for every change above (real bulk-index error shape, config-field absence, first-cycle ordering, `_collect_agents` failure isolation, `manage_indexer_tasks` debug logging).
- `framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py`: replaced the old ad-hoc-index class with `TestMetricsPipelineIntegration`, which runs the real `_collect_and_index()` pipeline against a real indexer (TLS + auth, the same shape the manager uses, configurable via `OPENSEARCH_URL`/`OPENSEARCH_USER`/`OPENSEARCH_PASSWORD`/`OPENSEARCH_VERIFY_CERTS`) and asserts documents land in the three real data streams. Also fixed two pre-existing, unrelated bugs in this file that left it failing at collection — meaning none of its tests, including two regression tests, had ever run: a missing `wazuh.core.cluster.control` entry in the mocked-imports dict, and `patch.dict(sys.modules, ...)` dropping the freshly imported module from `sys.modules` on exit, which silently no-op'd every `patch("wazuh.core.indexer.metrics_snapshot.<name>", ...)` in the file.

### Results and Evidence

`cluster.log` starts clean, no retry loop, and the first metrics snapshot cycle runs in the same second the manager comes up — not after the 10-minute wait it used to:

```
INFO: [Master] [Main] Indexer is available. Starting indexer tasks.
INFO: [Master] [metrics_snapshot] Starting metrics snapshot cycle.
INFO: [Master] [Active Response] Created 0 task(s) for 0 of 0 active response(s) read.
INFO: [Master] [metrics_snapshot] Metrics snapshot cycle completed. Collected/validated/indexed per index:
  {'wazuh-metrics-agents': '4/4/4', 'wazuh-metrics-comms-v4': '1/1/1', 'wazuh-metrics-normalization': '23/23/23'}
```

The three indices already store real documents. Querying the indexer directly confirms it:

```
$ curl -sk -u admin:admin "https://localhost:9200/_cat/indices/*metrics*?v"
health status index                                   docs.count store.size
green  open   .ds-wazuh-metrics-comms-v4-000001        2          ...
green  open   .ds-wazuh-metrics-agents-000001          4          ...
green  open   .ds-wazuh-metrics-normalization-000001   42         ...

$ curl -sk -u admin:admin "https://localhost:9200/wazuh-metrics-agents/_search?size=1&sort=@timestamp:desc"
{
  "@timestamp": "2026-09-15T14:34:06Z",
  "wazuh": {
    "agent": {
      "id": "001", "name": "wazuh-agent-5x-ubuntu", "version": "v5.0.0",
      "groups": ["default"], "status": "active", "status_code": 5,
      "registered_at": "2026-09-15T10:25:47Z", "last_seen": "2026-09-15T14:33:14Z",
      "register": {"ip": "0.0.0.0"},
      "host": {"ip": ["172.19.0.2"], "architecture": "x86_64",
                "os": {"name": "Ubuntu", "version": "22.04.5 LTS (Jammy Jellyfish)", "platform": "ubuntu"}}
    },
    "cluster": {"name": "wazuh", "node": "node01"},
    "schema": {"version": "1"}
  }
}
```

No `config` or `os.full` field, matching the index template exactly. A live check against every accumulated `wazuh-metrics-agents` document confirms neither field is ever emitted:

```
$ curl -sk -u admin:admin "https://localhost:9200/wazuh-metrics-agents/_search" -d '{
    "query": {"bool": {"should": [
      {"exists": {"field": "wazuh.agent.config"}},
      {"exists": {"field": "wazuh.agent.host.os.full"}}
    ]}}, "size": 0}'
{"hits": {"total": {"value": 0}}}
```

A cluster worker (`f97bdff221bb`) was joined for this verification to confirm the DAPI fan-out that feeds comms and normalization. The second automatic cycle, exactly 600 seconds after the first, shows it flowing in:

```
INFO: [Master] [metrics_snapshot] Starting metrics snapshot cycle.
INFO: [Master] [metrics_snapshot] Metrics snapshot cycle completed. Collected/validated/indexed per index:
  {'wazuh-metrics-agents': '4/4/4', 'wazuh-metrics-comms-v4': '2/2/2', 'wazuh-metrics-normalization': '42/42/42'}
```

```
$ curl -sk -u admin:admin "https://localhost:9200/wazuh-metrics-comms-v4/_search" -d '{
    "query": {"term": {"wazuh.cluster.node": "f97bdff221bb"}}}'
{
  "@timestamp": "2026-09-15T14:44:06Z",
  "wazuh": {"cluster": {"name": "wazuh", "node": "f97bdff221bb"}, "schema": {"version": "1"}},
  "event": {"module": "remoted"},
  "events": {"total": 0, "failed": {"total": 0}},
  ...
}
```

`wazuh-metrics-agents` correctly has no entries from the worker — agent collection is a single global wazuh-db query from the master, with no per-node fan-out by design, unlike comms and normalization.

Forcing a rejection (a document with an unmapped field) against the compiled/installed indexer client — not a mock — shows the real OpenSearch rejection reason instead of a lost warning:

```
WARNING wazuh.MetricsIndex: Metrics bulk index on 'wazuh-metrics-agents': 0 indexed, 1 failed.
First errors: [{'create': {..., 'status': 400, 'error': {'type': 'strict_dynamic_mapping_exception',
'reason': 'mapping set to strict, dynamic introduction of [totally_unmapped_field] within [_doc] is not allowed'}}}]
```

Unit tests, full `indexer` and `cluster` suites:

```
$ PYTHONPATH="$(pwd)/api:$(pwd)/framework" python -m pytest \
    framework/wazuh/core/indexer framework/wazuh/core/cluster \
    --ignore=framework/wazuh/core/indexer/tests/test_disconnected_agents_sync.py -q
826 passed, 1 skipped
```

(`test_disconnected_agents_sync.py` fails at collection on a pre-existing `jsonschema` version incompatibility in this environment, unrelated to and untouched by this change.)

Integration tests, against the real indexer, including the two previously-broken regression tests fixed by this PR:

```
$ pytest framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py -v
TestMetricsPipelineIntegration::test_pipeline_indexes_documents_into_the_three_real_data_streams PASSED
TestRegressionDaemonStats::test_get_daemons_stats_still_callable PASSED
TestRegressionDaemonStats::test_metrics_task_does_not_modify_daemons_stats_response PASSED
TestRegressionDaemonStats::test_collect_comms_does_not_raise_on_empty_cluster PASSED
TestRegressionAgentsEndpoint::test_get_all_agents_called_through_wdb_http_client PASSED
TestRegressionAgentsEndpoint::test_collect_agents_adds_metadata_fields PASSED
TestRegressionAgentsEndpoint::test_collect_agents_returns_list PASSED
TestRegressionAgentsEndpoint::test_metrics_task_registration_does_not_interfere_with_agents_query PASSED
8 passed
```

### Artifacts Affected

`framework/wazuh/core/indexer/metrics.py`, `metrics_snapshot.py` and `framework/wazuh/core/cluster/common.py` ship as part of the `wazuh-manager` package's Python framework (`packages/rpms/SPECS/wazuh-manager.spec` packages `framework/wazuh/core/**/*.py`; all three were observed installed and running at `/var/wazuh-manager/framework/python/lib/python3.12/site-packages/wazuh/core/...` during this session's live testing). No new artifact, no binary, no service unit, no default configuration file.

### Configuration Changes

N/A. `metrics_frequency` still controls the interval between cycles exactly as documented; only when the *first* cycle runs changed (immediately, not after one interval).

### Tests Introduced

No new test files. `test_metrics.py` and `test_metrics_snapshot.py` gained tests for the real bulk-index error shape, the removed agent fields, the first-cycle-before-sleep ordering, and `_collect_agents`'s failure isolation. `test_common.py` gained a test for `manage_indexer_tasks`'s debug-level traceback. `test_metrics_snapshot_integration.py`'s bulk-indexing class was replaced with one that exercises the real pipeline against a real indexer and the real data streams.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

#39184 (issue #39113), the blocking root cause, is already merged into `5.0.0` — this branch is rebased on top of it. This PR closes the remaining items found while investigating #39283 that #39184 did not touch: silent bulk-index rejection logging, two dead fields on the agent metrics document, the first-cycle startup delay, per-cycle observability, `_collect_agents` failure isolation, `manage_indexer_tasks` diagnostics, and a real end-to-end integration test (which also fixed a pre-existing bug that had kept the whole integration test file from ever running).


framework/wazuh/core/indexer/metrics_snapshot.py
```diff
@@ -150,19 +150,35 @@ def __init__(self, server, cluster_items: dict):
...
+        # wazuh-db outage propagates through the outer asyncio.gather() and skips comms
+        # and normalization collection too, not just agents.
+        try:
+            async with get_wdb_http_client() as wdb_client:
+                agents_data = await wdb_client.get_all_agents()
+        except Exception:
+            self.logger.exception("Failed to collect agents from wazuh-db")
+            return []
+
         node_name = self.server.configuration.get("node_name", "unknown")
```
Antoniogm03: **[S3 · non-blocking · introduced]** The new `try` wraps only the wazuh-db call; the config read and the `_normalize_agent_doc` loop on lines 182-192 sit outside it, so an unexpected type in one agent row (a non-textual `group`, an `os` that is a string, an out-of-range epoch in `dateAdd`/`lastKeepAlive`) still propagates through `_collect_and_index()`'s `asyncio.gather()` and aborts comms and normalization too. `_collect_comms_all_nodes()` and `_collect_normalization_all_nodes()` both keep their normalization inside the per-node `try`, so the symmetry the comment invokes does not hold yet.

The fix is to extend the `try` over the whole body (it does not fit the diff, so not a suggestion block):

```python
        try:
            async with get_wdb_http_client() as wdb_client:
                agents_data = await wdb_client.get_all_agents()

            node_name = self.server.configuration.get("node_name", "unknown")
            cluster_name = self.server.configuration.get("name", None)

            normalized = []
            for agent in agents_data:
                agent["@timestamp"] = timestamp
                agent["wazuh.cluster.node"] = node_name
                agent["wazuh.cluster.name"] = cluster_name
                normalized.append(self._normalize_agent_doc(agent))
        except Exception:
            self.logger.exception("Failed to collect agents from wazuh-db")
            return []

        return normalized
```

framework/wazuh/core/cluster/common.py
```diff
@@ -2090,6 +2090,11 @@ async def manage_indexer_tasks(
 
             except Exception as e:
                 self.logger.warning(f"Indexer is not configured or unavailable: {e}.")
+                # The warning above is deliberately terse (it fires on every transient
+                # unavailability, e.g. the indexer still starting up), but a deterministic
+                # bug hides behind the same message with no way to tell them apart. The
+                # traceback is one debug line away instead of invisible.
+                self.logger.debug("Indexer availability check failed.", exc_info=True)
```
Antoniogm03: **[S3 · non-blocking · introduced]** This traceback does not reach `cluster.log` on a default install. `IndexerTaskManager.__init__` takes `logging.getLogger('wazuh')` and `master.py:1050` instantiates the class standalone without overriding it, and `ClusterLogger.setup_logger()` (`cluster/utils.py:337-340`) sets that same logger to `INFO` unless `wazuh_clusterd.debug` is 1 or 2 — a key that does not appear in `etc/wazuh-manager-internal-options.conf`. The #39113 case you cite as the motivating example would still show only the terse warning.

A deterministic failure is the one worth escalating; the repetitions can stay at debug:

```python
                if type(e) is not getattr(self, '_last_indexer_error', None):
                    self._last_indexer_error = type(e)
                    self.logger.warning("Indexer availability check failed.", exc_info=True)
                else:
                    self.logger.debug("Indexer availability check failed.", exc_info=True)
```

framework/wazuh/core/indexer/metrics_snapshot.py
```diff
@@ -150,19 +150,35 @@ def __init__(self, server, cluster_items: dict):
...
 
     async def run_metrics_snapshot(self):
+        if self.frequency == 0:
+            self.logger.info("Metrics snapshot is disabled (metrics_frequency=0).")
+            return
+
+        # Run the first cycle immediately: without this, a manager restart means no
+        # metrics documents for a full interval (at least 10 minutes) even though the
+        # data to snapshot (agents, comms stats, engine metrics) is available right away.
+        first_cycle = True
```
Antoniogm03: **[S3 · non-blocking · introduced]** Running the first cycle immediately also means every *restart* of the task runs one, so `metrics_frequency` stops being an upper bound on the collection rate. `manage_indexer_tasks()` (`common.py:2070-2106`) cancels all indexer tasks whenever the healthcheck raises and recreates them on the next success, so an indexer that flaps produces a full `/agents/all` query plus a DAPI fan-out to every worker on each recovery — previously a task cancelled inside its first sleep cost nothing. Remembering the last cycle keeps the startup behaviour the issue asks for without that:

```python
        interval = max(self.frequency, self.DEFAULT_METRICS_FREQUENCY)
        while True:
            elapsed = time.monotonic() - getattr(self, "_last_cycle", -interval)
            if elapsed < interval:
                await asyncio.sleep(interval - elapsed)
            self._last_cycle = time.monotonic()
            try:
                await self._collect_and_index()
            except Exception:
                self.logger.exception("Metrics snapshot failed - skipping cycle")
```

framework/wazuh/core/indexer/metrics_snapshot.py
```diff
@@ -721,9 +739,14 @@ async def _collect_and_index(self):
...
             )
+        validated = {
+            "wazuh-metrics-agents": len(agent_docs),
+            "wazuh-metrics-comms-v4": len(comms_docs),
+            "wazuh-metrics-normalization": len(normalization_docs),
+        }
 
         async with get_indexer_client() as indexer:
-            await asyncio.gather(
+            indexed_agents, indexed_comms, indexed_normalization = await asyncio.gather(
```
Antoniogm03: **[S3 · non-blocking · introduced]** This `gather()` has no `return_exceptions=True` (which predates the PR; the unpack is what is new), so if the indexer drops the connection during the agents bulk, the unpack never completes and the per-index summary you just added on line 766 never prints — the operator is left with the generic "Metrics snapshot failed - skipping cycle", which is the situation this PR exists to remove. The sibling coroutines also keep running while `async with get_indexer_client()` closes the client under them, so their exceptions are never retrieved.

```python
            results = await asyncio.gather(..., return_exceptions=True)
        indexed = {}
        for index, result in zip(collected, results):
            if isinstance(result, Exception):
                self.logger.exception("Metrics bulk index on '%s' failed", index, exc_info=result)
                indexed[index] = "error"
            else:
                indexed[index] = result
```

framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py
```diff
@@ -142,180 +175,151 @@ def _make_tasks(server=None):
...
+                return_value=[normalization_doc],
+            ),
+            self._patch_real_indexer_client(indexer_client),
+            # common is module-mocked at the top of this file for import isolation;
+            # give _load_schema() a real (nonexistent) path so it degrades gracefully
+            # to "schema not found, skip local validation" instead of TypeError-ing on
+            # os.path.join(MagicMock(), ...). The real templates are still enforced by
+            # the live indexer itself, which is this test's actual point.
+            patch(
+                "wazuh.core.indexer.metrics_snapshot.common.INDEXER_PLUGINS_PATH",
```
Antoniogm03: **[S3 · non-blocking · introduced]** This patch disables the schema validation the class docstring lists as part of the test's job ("normalization, schema validation and bulk-index against a live indexer"). `_load_schema()` misses `os.path.isfile`, caches `None` for all three files, and the three `if <schema>:` guards in `_collect_and_index()` skip `_validate_documents()` entirely — a regression in `_opensearch_template_to_jsonschema()` that dropped every document would still pass here. Either point it at the real templates (`src/external/indexer-plugins`, the directory `installIndexerTemplates` copies from) and skip when absent, or drop "schema validation" from the docstring.

framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py
```diff
@@ -142,180 +175,151 @@ def _make_tasks(server=None):
...
+                "wazuh-metrics-normalization",
+            ):
+                await client.delete_by_query(
+                    index=index,
+                    body={"query": {"term": {"wazuh.cluster.node": self.NODE_NAME}}},
+                    ignore_unavailable=True,
+                    conflicts="proceed",
+                )
         except Exception:
             pass
```
Antoniogm03: **[S3 · non-blocking · introduced]** The three deletes share one `try` closed with `except Exception: pass`, so a failure on the first index silently skips cleanup of the other two and every run can leave `itest-<uuid>` documents permanently in the real metrics data streams. Writing into the production streams is the right call here (it is what catches a template mismatch), but then a failed cleanup has to be visible:

```python
        for index in ("wazuh-metrics-agents", "wazuh-metrics-comms-v4", "wazuh-metrics-normalization"):
            try:
                await client.delete_by_query(
                    index=index,
                    body={"query": {"term": {"wazuh.cluster.node": self.NODE_NAME}}},
                    ignore_unavailable=True,
                    conflicts="proceed",
                )
            except Exception as exc:
                print(f"WARNING: could not clean up {index} for {self.NODE_NAME}: {exc}")
        await client.close()
```

framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py
```diff
@@ -142,180 +175,151 @@ def _make_tasks(server=None):
...
+            await indexer_client.indices.refresh(index=index)
+            result = await indexer_client.search(
                 index=index,
-                body={"query": {"exists": {"field": "@timestamp"}}},
-                size=1,
+                body={"query": {"term": {"wazuh.cluster.node": self.NODE_NAME}}},
+            )
+            assert result["hits"]["total"]["value"] >= 1, (
+                f"No document with wazuh.cluster.node={self.NODE_NAME} found in {index}"
             )
```
Antoniogm03: **[S3 · non-blocking · introduced]** The rewrite drops every assertion on the indexed document itself — the removed tests parsed the stored `@timestamp` with `strptime` and checked each comms field in `_source`, and what replaces them only counts hits. A `_to_iso()` regression emitting an epoch string would still be accepted by a `date` mapping with no format and this test would pass. The live indexer already proves the *shape*; one line recovers the *values*:

```suggestion
            await indexer_client.indices.refresh(index=index)
            result = await indexer_client.search(
                index=index,
                body={"query": {"term": {"wazuh.cluster.node": self.NODE_NAME}}},
            )
            hits = result["hits"]["hits"]
            assert hits, (
                f"No document with wazuh.cluster.node={self.NODE_NAME} found in {index}"
            )
            datetime.strptime(hits[0]["_source"]["@timestamp"], "%Y-%m-%dT%H:%M:%SZ")
```

framework/wazuh/core/indexer/metrics_snapshot.py
```diff
@@ -150,28 +155,44 @@ def __init__(self, server, cluster_items: dict):
...
+            node_name = self.server.configuration.get("node_name", "unknown")
+            cluster_name = self.server.configuration.get("name", None)
+
+            normalized = []
+            for agent in agents_data:
+                agent["@timestamp"] = timestamp
+                agent["wazuh.cluster.node"] = node_name
+                agent["wazuh.cluster.name"] = cluster_name
+                normalized.append(self._normalize_agent_doc(agent))
+        except Exception:
```
jotacarma90: **[S3 · non-blocking · introduced]** The try opens at line 180, before the wazuh-db call, and does not close until after the `for agent in agents_data` loop, so one row that `_normalize_agent_doc()` chokes on discards every agent already normalized in that cycle. The comms and normalization collectors this comment compares against isolate *inside* their loop, one iteration per node (lines 258-268, 316-326); here the isolation is one level too coarse. This is still better than the pre-PR behaviour (no guard at all, so a bad row aborted the whole gather), but `test_collect_agents_normalization_failure_is_local` locks the all-or-nothing shape in.

```suggestion
        except Exception:
            self.logger.exception("Failed to collect agents from wazuh-db")
            return []
```

Scope the outer guard to the wazuh-db fetch and wrap `self._normalize_agent_doc(agent)` in its own try/except inside the loop, logging the failing agent id.

framework/wazuh/core/cluster/common.py
```diff
@@ -2090,6 +2090,8 @@ async def manage_indexer_tasks(
 
             except Exception as e:
                 self.logger.warning(f"Indexer is not configured or unavailable: {e}.")
+                # Terse by design: this also fires on routine transient unavailability.
+                self.logger.debug("Indexer availability check failed.", exc_info=True)
```
jotacarma90: **[S3 · non-blocking · introduced]** This traceback never reaches `cluster.log` on a default install: `IndexerTaskManager.__init__` (line 2035) uses `logging.getLogger('wazuh')`, and `ClusterLogger` (`cluster/utils.py:337-340`) sets that logger to INFO unless clusterd starts with `debug_level >= 1`. A deterministic `AttributeError` and a transient connection refusal still produce the same single warning — the blind spot this change targets. Every other `exc_info=True` in `framework/wazuh/core` (12 of them, in `dapi.py`, `disconnected_agents.py`, elsewhere in this file) is at `error` level.

```suggestion
                self.logger.warning(
                    f"Indexer is not configured or unavailable: {e}.", exc_info=True
                )
```

This replaces the warning on line 2092 and the comment on 2093; `test_manage_indexer_tasks_logs_debug_traceback_on_failure` needs the matching update.

framework/wazuh/core/indexer/metrics_snapshot.py
```diff
@@ -697,40 +714,57 @@ def _validate_documents(self, docs: list, schema: dict, index_name: str) -> list
...
             )
         if normalization_schema:
             normalization_docs = self._validate_documents(
-                normalization_docs, normalization_schema, "wazuh-metrics-normalization"
+                normalization_docs, normalization_schema, NORMALIZATION_INDEX
             )
+        validated = {
+            AGENTS_INDEX: len(agent_docs),
+            COMMS_INDEX: len(comms_docs),
+            NORMALIZATION_INDEX: len(normalization_docs),
```
jotacarma90: **[S3 · non-blocking · introduced]** When `_load_schema()` returns None the `if <schema>:` branches above are skipped, so `validated` ends up equal to `collected` and the cycle logs `'wazuh-metrics-agents': '4/4/4'` — which reads as "all 4 validated" while every document went to the indexer unvalidated. `_load_schema` caches the None (lines 645-652), so the "schema not found" warning fires once per process and from the second cycle on nothing in the log says validation is off. The PR's own evidence uses `4/4/4` as proof that validation ran.

```suggestion
        validated = {
            AGENTS_INDEX: len(agent_docs) if agents_schema else None,
            COMMS_INDEX: len(comms_docs) if comms_schema else None,
            NORMALIZATION_INDEX: len(normalization_docs) if normalization_schema else None,
        }
```

Then render a None as `-` in the summary line so the skip is visible.

framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py
```diff
@@ -5,35 +5,76 @@
...
+import urllib3
+
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
 # ---------------------------------------------------------------------------
 # Optional OpenSearch availability check
 # ---------------------------------------------------------------------------
 
-OPENSEARCH_URL = "http://localhost:9200"
+OPENSEARCH_URL = os.environ.get("OPENSEARCH_URL", "https://localhost:9200")
```
jotacarma90: **[S3 · non-blocking · introduced]** The blast radius grew on two axes at once: the default moved from `http://localhost:9200` to `https://localhost:9200` with `admin`/`admin` and certs unverified — the default endpoint and credentials of a real Wazuh indexer — and the target moved from disposable `-test` indices to the three production data streams the dashboard reads. `pytest framework`, the CI's own invocation, now writes a fake agent (id 999) and a fake metric into live `wazuh-metrics-*` on any machine with the devcontainer stack up. Gate the real-stream write behind an explicit opt-in env var rather than making it the default.

framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py
```diff
@@ -142,180 +171,180 @@ def _make_tasks(server=None):
...
+        ):
+            try:
+                await client.delete_by_query(
+                    index=index,
+                    body={"query": {"term": {"wazuh.cluster.node": self.NODE_NAME}}},
+                    ignore_unavailable=True,
+                    conflicts="proceed",
+                )
+            except Exception as exc:
+                print(f"WARNING: could not clean up {index} for {self.NODE_NAME}: {exc}")
```
jotacarma90: **[S3 · non-blocking · introduced]** This directly contradicts the comment three lines above it: pytest captures stdout, so on a passing test a failed cleanup prints nothing at all and three synthetic documents stay in the live data streams for the ISM policy's 30-day retention.

```suggestion
            except Exception as exc:
                pytest.fail(f"could not clean up {index} for {self.NODE_NAME}: {exc}")
``` Note also that the per-index `indices.refresh` happens inside the assertion loop below, so when the first index's assertion fails the other two are never refreshed before this `delete_by_query` runs, and it can miss the documents just written.

File: framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py [hunk 13]
```diff
@@ -64,7 +106,17 @@
     import wazuh.core.indexer.metrics_snapshot as _metrics_snapshot_module
     from wazuh.core.indexer.metrics_snapshot import MetricsSnapshotTasks
 
+# patch.dict restores sys.modules to its pre-`with` snapshot on exit, which silently
+# drops the 'wazuh.core.indexer.metrics_snapshot' entry the import above just created
+# (it wasn't a pre-existing key). Without putting it back, any later
+# patch("wazuh.core.indexer.metrics_snapshot.<name>", ...) re-imports the module for
+# real (this time with no mocked dependencies) and patches that unrelated copy instead
+# of the one MetricsSnapshotTasks is actually bound to — the patch silently no-ops.
+sys.modules["wazuh.core.indexer.metrics_snapshot"] = _metrics_snapshot_module
```
jotacarma90: **[S3 · non-blocking · introduced]** Re-inserting the module makes it the process-wide `wazuh.core.indexer.metrics_snapshot` for the rest of the session, and when this file is collected first its globals are bound to MagicMocks. Anything collected later that imports it — directly or through `wazuh.core.cluster.master:26` — gets the mock-wired copy. It is masked today only because `cluster/` sorts before `indexer/` in `pytest framework`, so the real module is already in `sys.modules` by the time line 106 runs; `pytest <this file> framework/wazuh/core/cluster/` reverses that. Restore `sys.modules` on teardown with a module-scoped autouse fixture instead of leaving the entry in place. At minimum, mirror `test_metrics_snapshot.py:81` and use `sys.modules.setdefault(...)` so an already-imported real module is never clobbered.

framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py
```diff
@@ -142,180 +171,180 @@ def _make_tasks(server=None):
...
+            # _load_schema() needs a real INDEXER_PLUGINS_PATH pointed at it explicitly
+            # instead of os.path.join(MagicMock(), ...). requires_indexer_templates
+            # guarantees this path exists and holds the three real templates, so
+            # _validate_documents() actually runs against them instead of being
+            # bypassed. Note this only catches top-level shape mismatches:
+            # _opensearch_template_to_jsonschema() sets additionalProperties: False
+            # only at levels whose own template node declares dynamic: strict, not
+            # recursively, so an extra *nested* field (the exact shape of the
+            # wazuh.agent.config/host.os.full mismatch this PR removes) would still
+            # pass local validation and only be caught by the live indexer itself.
```
jotacarma90: **[S3 · non-blocking · pre-existing]** This comment is right, and the gap it describes is the reason the two removed fields were never caught locally: I downloaded all three 5.0.0 templates and walked every node — `metrics-agents.json`, `metrics-comms.json` and `metrics-normalization.json` each declare `dynamic: strict` only at `template.mappings`, and no nested object declares it. OpenSearch inherits it downward; `_build_jsonschema_properties()` (`metrics_snapshot.py:62`) does not, so `_validate_documents()` passes any nested extra field and only the indexer rejects it. Worth closing rather than documenting: thread the effective `dynamic` through the recursion and set `additionalProperties: False` on every object that does not override it. *Not introduced by this PR — flagging it for a follow-up issue, not for this review.*

framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py
```diff
@@ -5,35 +5,76 @@
...
     OPENSEARCH_AVAILABLE = resp.status_code == 200
 except Exception:
     pass
 
+# Real index templates, as downloaded by `make deps` into src/external/indexer-plugins
+# (see src/Makefile's INDEXER_PLUGINS_DIR) and installed under etc/indexer-plugins/ on a
+# real manager. Used to exercise the same schema validation _load_schema() runs in
+# production, not a stand-in.
+INDEXER_TEMPLATES_PATH = os.environ.get(
+    "INDEXER_TEMPLATES_PATH", "src/external/indexer-plugins"
```
jotacarma90: **[S4 · non-blocking · introduced]** This default is CWD-relative. It resolves under the CI invocation (`pytest framework` from the repo root, after `make -C src deps TARGET=manager`, which does fetch the three `metrics-*.json` — `src/Makefile:548-551`), but `cd framework && pytest wazuh/core/indexer` — the natural invocation given `framework/pytest.ini` — skips the only real integration test with a reason that blames missing templates rather than the cwd. Derive it from `__file__`.

framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py
```diff
@@ -5,35 +5,76 @@
...
+import uuid
 from datetime import datetime, timezone
+from types import SimpleNamespace
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 import requests
+import urllib3
+
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
jotacarma90: **[S4 · non-blocking · introduced]** This runs at module import, which happens during collection of the whole indexer suite whether or not OpenSearch is reachable, so it mutates urllib3's warning filters for the entire pytest process. Scope it to the availability probe below with `warnings.catch_warnings()`.

framework/wazuh/core/indexer/metrics_snapshot.py
```diff
@@ -150,28 +155,44 @@ def __init__(self, server, cluster_items: dict):
...
+        # Run the first cycle immediately: without this, a manager restart means no
+        # metrics documents for a full interval (at least 10 minutes) even though the
+        # data to snapshot (agents, comms stats, engine metrics) is available right away.
+        first_cycle = True
         while True:
-            if self.frequency == 0:
-                self.logger.info("Metrics snapshot is disabled (metrics_frequency=0).")
-                return
-            await asyncio.sleep(max(self.frequency, self.DEFAULT_METRICS_FREQUENCY))
+            if not first_cycle:
```
jotacarma90: **[S4 · non-blocking · introduced]** The flag is redundant state — moving the sleep to the end of the loop body gives the identical schedule, and that is the shape the sibling indexer task already uses (`disconnected_agents.py:168-219`: `while True: try: ... finally: await asyncio.sleep(...)`). `test_first_cycle_runs_before_any_sleep` asserts the order, not the implementation, so it still passes.

```suggestion
        while True:
            try:
                await self._collect_and_index()
            except Exception:
                self.logger.exception("Metrics snapshot failed - skipping cycle")
            await asyncio.sleep(max(self.frequency, self.DEFAULT_METRICS_FREQUENCY))
```

This replaces lines 165-173; drop the `first_cycle` comment block with it.

framework/wazuh/core/indexer/tests/test_metrics.py
```diff
@@ -149,29 +149,36 @@ async def test_bulk_index_different_indices(self, metrics_index):
...
+                errors[:5],
             )
 
     @pytest.mark.asyncio
     async def test_bulk_index_no_warning_on_success(self, metrics_index):
         """Verify no warning is logged when all documents are indexed successfully."""
         docs = [{"x": 1}]
 
-        with patch("wazuh.core.indexer.metrics.async_bulk", new_callable=AsyncMock, return_value=(1, 0)):
+        with patch("wazuh.core.indexer.metrics.async_bulk", new_callable=AsyncMock, return_value=(1, [])):
```
jotacarma90: **[S4 · non-blocking · pre-existing]** Only the two warning tests were updated. Lines 45, 58, 79, 94, 109, 122 and 139 still patch `async_bulk` with `return_value=(2, 0)` / `(1, 0)` / `(0, 0)`; with the default `stats_only=False` the second element is always a list — the premise of this PR. Nothing fails today because `if failed:` treats `0` and `[]` alike, but it leaves the original wrong assumption seeded in the file. Change them to `[]` while you are here.

framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py
```diff
@@ -142,180 +171,180 @@ def _make_tasks(server=None):
...
 
 
-class TestBulkIndexingIntegration:
-    """End-to-end bulk indexing tests against a local OpenSearch instance."""
+class TestMetricsPipelineIntegration:
+    """Runs MetricsSnapshotTasks._collect_and_index() end to end against a real
+    indexer and confirms documents land in the three *real* data streams
+    (wazuh-metrics-agents, wazuh-metrics-comms-v4, wazuh-metrics-normalization) with
+    the shape their index templates expect — not an ad-hoc scratch index, which is
+    what let a real template mismatch (issue #39283) pass unnoticed here before.
```
jotacarma90: **[S4 · non-blocking · introduced]** "which is what let a real template mismatch (issue #39283) pass unnoticed here before" and, at line 306, "the exact shape of the `wazuh.agent.config`/`host.os.full` mismatch this PR removes" describe the change history rather than a constraint the code cannot express, and the second goes stale the moment this merges. Keep the technical observation, drop the provenance.

framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py
```diff
@@ -64,7 +113,17 @@
     import wazuh.core.indexer.metrics_snapshot as _metrics_snapshot_module
     from wazuh.core.indexer.metrics_snapshot import MetricsSnapshotTasks
 
+# patch.dict restores sys.modules to its pre-`with` snapshot on exit, which silently
+# drops the 'wazuh.core.indexer.metrics_snapshot' entry the import above just created
+# (it wasn't a pre-existing key). Without putting it back, any later
+# patch("wazuh.core.indexer.metrics_snapshot.<name>", ...) re-imports the module for
+# real (this time with no mocked dependencies) and patches that unrelated copy instead
+# of the one MetricsSnapshotTasks is actually bound to — the patch silently no-ops.
+sys.modules["wazuh.core.indexer.metrics_snapshot"] = _metrics_snapshot_module
```
jotacarma90, line 122: **[S3 · non-blocking · introduced]** Re-inserting the module makes it the process-wide `wazuh.core.indexer.metrics_snapshot` for the rest of the session, and when this file is collected first its globals are bound to MagicMocks. Anything collected later that imports it — directly or through `wazuh.core.cluster.master:26` — gets the mock-wired copy. It is masked today only because `cluster/` sorts before `indexer/` in `pytest framework`, so the real module is already in `sys.modules` by the time line 106 runs; `pytest <this file> framework/wazuh/core/cluster/` reverses that. Restore `sys.modules` on teardown with a module-scoped autouse fixture instead of leaving the entry in place. At minimum, mirror `test_metrics_snapshot.py:81` and use `sys.modules.setdefault(...)` so an already-imported real module is never clobbered.

framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py
```diff
@@ -142,180 +178,189 @@ def _make_tasks(server=None):
...
+            # _load_schema() needs a real INDEXER_PLUGINS_PATH pointed at it explicitly
+            # instead of os.path.join(MagicMock(), ...). requires_indexer_templates
+            # guarantees this path exists and holds the three real templates, so
+            # _validate_documents() actually runs against them instead of being
+            # bypassed. Note this only catches top-level shape mismatches:
+            # _opensearch_template_to_jsonschema() sets additionalProperties: False
+            # only at levels whose own template node declares dynamic: strict, not
+            # recursively, so an extra *nested* field (the exact shape of the
+            # wazuh.agent.config/host.os.full mismatch this PR removes) would still
+            # pass local validation and only be caught by the live indexer itself.
```
jotacarma90, line 324: **[S3 · non-blocking · pre-existing]** This comment is right, and the gap it describes is the reason the two removed fields were never caught locally: I downloaded all three 5.0.0 templates and walked every node — `metrics-agents.json`, `metrics-comms.json` and `metrics-normalization.json` each declare `dynamic: strict` only at `template.mappings`, and no nested object declares it. OpenSearch inherits it downward; `_build_jsonschema_properties()` (`metrics_snapshot.py:62`) does not, so `_validate_documents()` passes any nested extra field and only the indexer rejects it. Worth closing rather than documenting: thread the effective `dynamic` through the recursion and set `additionalProperties: False` on every object that does not override it. *Not introduced by this PR — flagging it for a follow-up issue, not for this review.*

framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py
```diff
@@ -5,35 +5,83 @@
...
+        OPENSEARCH_AVAILABLE = resp.status_code == 200
+    except Exception:
+        pass
+
+# Real index templates, as downloaded by `make deps` into src/external/indexer-plugins
+# (see src/Makefile's INDEXER_PLUGINS_DIR) and installed under etc/indexer-plugins/ on a
+# real manager. Used to exercise the same schema validation _load_schema() runs in
+# production, not a stand-in.
+INDEXER_TEMPLATES_PATH = os.environ.get(
+    "INDEXER_TEMPLATES_PATH", "src/external/indexer-plugins"
```
jotacarma90, line 63: **[S4 · non-blocking · introduced]** This default is CWD-relative. It resolves under the CI invocation (`pytest framework` from the repo root, after `make -C src deps TARGET=manager`, which does fetch the three `metrics-*.json` — `src/Makefile:548-551`), but `cd framework && pytest wazuh/core/indexer` — the natural invocation given `framework/pytest.ini` — skips the only real integration test with a reason that blames missing templates rather than the cwd. Derive it from `__file__`.

framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py
```diff
@@ -5,35 +5,83 @@
...
+import uuid
 from datetime import datetime, timezone
+from types import SimpleNamespace
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 import requests
+import urllib3
+
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
jotacarma90, line 34: **[S4 · non-blocking · introduced]** This runs at module import, which happens during collection of the whole indexer suite whether or not OpenSearch is reachable, so it mutates urllib3's warning filters for the entire pytest process. Scope it to the availability probe below with `warnings.catch_warnings()`.

framework/wazuh/core/indexer/metrics_snapshot.py
```diff
@@ -150,19 +155,34 @@ def __init__(self, server, cluster_items: dict):
...
```
jotacarma90, line 167: **[S4 · non-blocking · introduced]** The flag is redundant state — moving the sleep to the end of the loop body gives the identical schedule, and that is the shape the sibling indexer task already uses (`disconnected_agents.py:168-219`: `while True: try: ... finally: await asyncio.sleep(...)`). `test_first_cycle_runs_before_any_sleep` asserts the order, not the implementation, so it still passes.

```suggestion
        while True:
            try:
                await self._collect_and_index()
            except Exception:
                self.logger.exception("Metrics snapshot failed - skipping cycle")
            await asyncio.sleep(max(self.frequency, self.DEFAULT_METRICS_FREQUENCY))
```

This replaces lines 165-173; drop the `first_cycle` comment block with it.

framework/wazuh/core/indexer/tests/test_metrics.py
jotacarma90, line 181: **[S4 · non-blocking · pre-existing]** Only the two warning tests were updated. Lines 45, 58, 79, 94, 109, 122 and 139 still patch `async_bulk` with `return_value=(2, 0)` / `(1, 0)` / `(0, 0)`; with the default `stats_only=False` the second element is always a list — the premise of this PR. Nothing fails today because `if failed:` treats `0` and `[]` alike, but it leaves the original wrong assumption seeded in the file. Change them to `[]` while you are here.

framework/wazuh/core/indexer/tests/test_metrics_snapshot_integration.py
```diff
@@ -142,180 +178,189 @@ def _make_tasks(server=None):
...
 
 
-class TestBulkIndexingIntegration:
-    """End-to-end bulk indexing tests against a local OpenSearch instance."""
+class TestMetricsPipelineIntegration:
+    """Runs MetricsSnapshotTasks._collect_and_index() end to end against a real
+    indexer and confirms documents land in the three *real* data streams
+    (wazuh-metrics-agents, wazuh-metrics-comms-v4, wazuh-metrics-normalization) with
+    the shape their index templates expect — not an ad-hoc scratch index, which is
+    what let a real template mismatch (issue #39283) pass unnoticed here before.
```
jotacarma90, line 186: **[S4 · non-blocking · introduced]** "which is what let a real template mismatch (issue #39283) pass unnoticed here before" and, at line 306, "the exact shape of the `wazuh.agent.config`/`host.os.full` mismatch this PR removes" describe the change history rather than a constraint the code cannot express, and the second goes stale the moment this merges. Keep the technical observation, drop the provenance.

jotacarma90 (approved):

PR #39316, opened 2026-09-15
fix(api): survive a transient bind conflict on the API port instead of exiting apid permanently

### LINKED ISSUE #39231: wazuh-manager-apid gives up permanently if port 55000 is momentarily busy
## Description

On a fresh 5.0.0 manager installation, `wazuh-manager-apid` tried to bind `0.0.0.0:55000`, got `EADDRINUSE`, logged it and shut itself down. **It never retried.** The `wazuh-manager` systemd unit stayed `active`, every other daemon started normally, and `wazuh-manager.log` never mentioned the failure — so the node ran for **24 minutes with no REST API** while every health check available said it was healthy. The API only came back because something else restarted the manager later, and that restart bound port 55000 without any trouble, which shows the condition that blocked it was transient.

Logs:
- [full_component_logs.zip](https://github.com/user-attachments/files/32184611/full_component_logs.zip)
- [logs.zip](https://github.com/user-attachments/files/32184617/logs.zip)

Two separable problems, both in this one event:

1. **A momentary conflict on 55000 is permanent.** One bind attempt, no retry, no backoff.
2. **The failure is invisible above the daemon.** `systemctl is-active wazuh-manager` returns `active`, `wazuh-manager.log` logs nothing, and the only trace is in `api.log`. Nothing that an operator, an orchestrator or a health probe would normally watch can tell that the manager is half-dead.

## Environment

- Component: `wazuh-manager` / `wazuh-manager-apid`
- Version under test: `5.0.0` (`rc1`, `internal-latest`)
- OS: Amazon Linux 2023 amd64 — master node of a 10-node Central Components cluster
- Incident: wazuh/devel-major-incidents#250 (nightly of 2026-09-12)
- Build: [deployability_testing 1151](https://jenkins-production.qa.wazuh.info/job/5.0.0/job/deployability_testing/1151/) — `DIT-CC`, host `aio-amazon-2023-amd64`

## Error

Complete `api.log` of the affected node (`aio-amazon-2023-amd64_manager_api.log` in `full_component_logs.zip`); note that the whole file is 25 lines, so nothing is elided:

```
2026/09/12 03:50:54 INFO: HTTPS is enabled but cannot find the private key and/or certificate. Attempting to generate them
2026/09/12 03:50:55 INFO: Generated private key file in WAZUH_PATH/etc/certs/apid-key.pem
2026/09/12 03:50:55 INFO: Generated certificate file in WAZUH_PATH/etc/certs/apid.pem
2026/09/12 03:50:55 INFO: Checking RBAC database integrity...
2026/09/12 03:50:55 INFO: RBAC database not found. Initializing
2026/09/12 03:50:55 INFO: /var/wazuh-manager/api/configuration/security/rbac.db database created successfully
2026/09/12 03:50:55 INFO: RBAC database integrity check finished successfully
2026/09/12 03:50:56 WARNING: The 'wazuh' API user still has its default password. [...]
2026/09/12 03:50:56 WARNING: The 'wazuh-wui' API user still has its default password. [...]
2026/09/12 03:50:57 INFO: Listening on ['0.0.0.0', '::']:55000.
2026/09/12 03:50:57 ERROR: [Errno 98] error while attempting to bind on address ('0.0.0.0', 55000): address already in use
2026/09/12 03:50:57 INFO: Shutdown wazuh-manager-apid server.
2026/09/12 04:14:42 INFO: Checking RBAC database integrity...
2026/09/12 04:14:42 INFO: /var/wazuh-manager/api/configuration/security/rbac.db file was detected
2026/09/12 04:14:42 INFO: RBAC database integrity check finished successfully
2026/09/12 04:14:42 WARNING: The 'wazuh' API user still has its default password. [...]
2026/09/12 04:14:42 WARNING: The 'wazuh-wui' API user still has its default password. [...]
2026/09/12 04:14:44 INFO: Listening on ['0.0.0.0', '::']:55000.
2026/09/12 04:21:25 INFO: Shutdown wazuh-manager-apid server.
2026/09/12 04:21:29 INFO: Checking RBAC database integrity...
2026/09/12 04:21:29 INFO: /var/wazuh-manager/api/configuration/security/rbac.db file was detected
2026/09/12 04:21:29 INFO: RBAC database integrity check finished successfully
2026/09/12 04:21:29 WARNING: The 'wazuh' API user still has its default password. [...]
2026/09/12 04:21:29 WARNING: The 'wazuh-wui' API user still has its default password. [...]
2026/09/12 04:21:31 INFO: Listening on ['0.0.0.0', '::']:55000.
```

Line 11 is the only failure; line 12 is the last thing apid does. The next entry is 24 minutes later, at 04:14:42, and it is a **different** start — triggered by an external `systemctl restart`, not by apid itself (the RBAC database is now "detected" rather than "created", so this is the second ever start). That restart bound 55000 on the first try.

### The manager reports itself healthy throughout

`wazuh-manager.log` on the same node, same window, shows every other daemon starting normally seconds before and after the failed bind:

```
2026/09/12 03:50:55 wazuh-manager-authd: INFO: Started (pid: 10269).
2026/09/12 03:50:55 wazuh-manager-db: INFO: Started (pid: 10277).
2026/09/12 03:50:58 wazuh-manager-analysisd: INFO: Engine started
2026/09/12 03:50:59 wazuh-manager-remoted: INFO: Started (pid: 10518). Listening on port 1514/TCP (secure).
2026/09/12 03:50:59 wazuh-manager-modulesd: INFO: Started (pid: 10589).
```

And it never mentions apid: `grep -cE 'ERROR|CRITICAL'` over `wazuh-manager.log` between `03:50` and `04:10` returns **0**, and the only two lines matching `api` in the whole file are unrelated `wazuh-manager-analysisd: INFO: [Server] API services stopped` entries from the later restarts at 04:14:37 and 04:21:25.

Fourteen seconds after the bind failed, the service-level check was green:

```
[2026-09-12 03:51:11] Command finished: sudo bash -c 'systemctl is-active wazuh-manager'
STDOUT: active
```

So there is no observable difference, at the service level, between a manager with a working API and this one. The only way to find out is to read `api.log` or to call the API and get `ECONNREFUSED`.

### The errno-98 handler exists, but it is unreachable

`api/scripts/wazuh_manager_apid.py` already tries to handle exactly this case, at lines 227-237:

```python
    try:
        uvicorn.run(app, **params)

    except OSError as exc:
        if exc.errno == 98:
            error = APIError(2010)
            logger.error(error)
            raise error
```

That branch never runs. `uvicorn` is pinned at `0.31.0` (`framework/requirements.txt`), and `uvicorn/server.py:169-172` catches the `OSError` first:

```python
            except OSError as exc:
                logger.error(exc)
                await self.lifespan.shutdown()
                sys.exit(1)
```

`sys.exit(1)` raises `SystemExit`, which derives from `BaseException` and is **not** an `OSError`, so the `except OSError` in `wazuh_manager_apid.py` cannot catch it. `APIError(2010)` — *"Error while attempting to bind on address: address already in use"* — is dead code on this path, and the log confirms it: what was written is uvicorn's own wording (`[Errno 98] error while attempting to bind on address ('0.0.0.0', 55000): address already in use`, lowercase, with errno and address tuple), not the APIError text.

The three log lines map one-to-one onto that uvicorn block, in order:

| Log line | Source |
|---|---|
| `Listening on ['0.0.0.0', '::']:55000.` | `api/api/signals.py:74` — inside `lifespan_handler`, **before** the `yield`, so before any socket exists |
| `[Errno 98] error while attempting to bind on address (...)` | `uvicorn/server.py:170` — `logger.error(exc)` |
| `Shutdown wazuh-manager-apid server.` | `uvicorn/server.py:171` calls `lifespan.shutdown()`, which reaches `api/api/signals.py:82` |

Two consequences worth separating. First, **the `Listening on ...:55000.` line is emitted before the bind is attempted**, so it is misleading exactly when it matters: the log claims the API is listening on the port one line before saying it could not take it. Second, **apid does exit non-zero** (`sys.exit(1)`), so the information is not lost at process level — it is lost at service level, because apid is a child launched by `wazuh-control` and not the main process of the `wazuh-manager` unit, so systemd never sees it.

### What was holding the port — most likely mechanism, not proven

I cannot prove what occupied 55000 at 03:50:57: no socket table was captured at that instant, and by 04:06 the port was free again (a later check reported `API port should be LISTEN but state: <no output, port not listening>`, i.e. nothing was listening on it at all). What is certain is that it was occupied at 03:50:57 and free minutes later, and that no other Wazuh component listens on 55000.

The most plausible mechanism is an **ephemeral source port collision**: 55000 lies inside the Linux default `net.ipv4.ip_local_port_range` (`32768 60999`), so any outbound connection can be assigned it as its source port. In the two seconds between the manager starting and apid binding, this node opens many outbound connections — to the ten indexer nodes on 9200, to the cluster peers on 1516, and to the CTI service over 443 — so one of them landing on 55000 is entirely possible. This would explain both the transience and the fact that it hit only one of the seven managers in the nightly.

Whatever the trigger, it is a condition apid should survive: a listening port inside the ephemeral range is momentarily unavailable from time to time, by design of the network stack.

### Impact observed

This single bind failure is the root of almost the entire Central Components install phase. The suite installs the stack strictly in order — indexer, then manager, then dashboard — and the manager step ends by authenticating against the API. That call timed out after 60 attempts over 900 s (03:51:11 → 04:05:57):

```
RuntimeError: Could not authenticate against manager API after 60 attempts (900s); last state: request error: HTTPSConnectionPool(host='34.229.118.49', port=55000): Max retries exceeded with url: /security/user/authenticate (Caused by NewConnectionError('...: Failed to establish a new connection: [Errno 111] Connection refused'))
```

The exception aborted the installation before the dashboard iteration, so **the Wazuh dashboard was never installed on any of the ten instances**. Of the phase's 24 failed tests, 21 trace back to this — the two direct API assertions plus every dashboard directory, service, version, keystore and port check — and all 25 errors of `test_04_uninstall.py` error at setup on `systemctl start wazuh-dashboard returned non-zero exit code 5 / Unit wazuh-dashboard.service not found`. The only three failures unrelated to it are known catalogued log warnings.

## Steps to reproduce

The trigger is timing-dependent and will not reproduce on demand, but **the reported behaviour** does, deterministically:

1. On a host with the manager installed and stopped, occupy the API port: `python3 -c "import socket;s=socket.socket();s.bind(('0.0.0.0',55000));s.listen();input()"`.
2. `systemctl start wazuh-manager`.
3. Observe in `api.log`: `ERROR: [Errno 98] error while attempting to bind on address ('0.0.0.0', 55000): address already in use` followed by `Shutdown wazuh-manager-apid server.`
4. Observe that `systemctl is-active wazuh-manager` returns `active`, that `wazuh-manager.log` says nothing about it, and that every other daemon is running.
5. Release the port. Observe that apid does **not** come back: the API stays down until the manager is restarted by hand.

## Expected behavior

A momentarily busy port should not cost the node its API for the rest of its uptime, and it should never be silent. Either of these would be acceptable, and they are not mutually exclusive:

- apid retries the bind with backoff until the port becomes available, logging each attempt;
- the failure is propagated so that it is visible where people and automation look: an `ERROR`/`CRITICAL` entry in `wazuh-manager.log`, a non-`active` unit state, and a non-ready `wazuh-control status` / cluster status.

The state that should not be reachable is the one observed: a manager that systemd, `wazuh-control` and the main log all describe as healthy while its REST API is dead.

## Possible fixes

Ownership, per `.github/CODEOWNERS`: `/api/` and `/framework/` belong to `@wazuh/devel-xdrsiem-server-div2` and `@wazuh/devel-xdrsiem-server-leaders`.

- **Bind the socket in our own code, then hand it to uvicorn.** Create and `bind()` the listening socket in `wazuh_manager_apid.py`, retrying with backoff on `EADDRINUSE`, and pass it in (`uvicorn.Server(config).run(sockets=[sock])`). This keeps the failure inside our `try` instead of inside uvicorn's, which both makes the retry possible and makes the existing `APIError(2010)` branch reachable for the first time. A bounded retry — a few attempts over a minute or two — is enough: the port was free again within minutes in the observed incident.
- **As a minimum, catch `SystemExit` as well** at `wazuh_manager_apid.py:230`, so that a bind failure is at least reported in our own words rather than only in uvicorn's. This alone does not fix the incident, because by then uvicorn has already given up.
- **Move the `Listening on ...` message after the successful bind** (`api/api/signals.py:74`), or reword it as a startup message. As written it asserts something that is not true yet, and is printed immediately before the contradicting error.
- **Make a definitive failure visible at service level**: mirror the error into `wazuh-manager.log`, and reflect it in `wazuh-control status` and in the cluster status endpoint, so that a manager whose API is down is not reported as ready. Today `systemctl is-active wazuh-manager` returns `active` regardless, because apid is a child of `wazuh-control` and not the unit's main process.
- Independently of the above, consider reserving the API port at packaging or documentation level (`net.ipv4.ip_local_reserved_ports = 55000`) so the kernel never hands it out as an ephemeral source port. This removes the likeliest trigger but not the defect: apid would still be permanently down after any other transient conflict.

MiguelazoDS:
## Description

On a fresh manager start, if `wazuh-manager-apid` can't bind `0.0.0.0:55000` because something else briefly holds the port, it exits once and never comes back on its own, and nothing at the service level shows the API is down. A reported incident hit this during a 10-node cluster install: the master node's apid died on its first bind attempt and stayed dead for 24 minutes, silently, until an unrelated restart brought it back, aborting most of that node's install in the meantime.

Closes #39231

## Proposed Changes

- apid now binds its own listening sockets first, retrying with backoff on a transient `EADDRINUSE`, and only then hands them to uvicorn. Previously the bind happened inside uvicorn itself, which turns a bind failure into `sys.exit(1)` before apid's own exception handling ever runs, so a momentarily busy port left the API down until the manager was restarted by hand.
- Each host's address family is resolved via `socket.getaddrinfo()` instead of a literal `:` check, and a host whose address family isn't available at all (`EADDRNOTAVAIL`/`EAFNOSUPPORT`, e.g. IPv6 disabled at the kernel level) is skipped rather than aborting the whole bind attempt — matching the tolerance `asyncio.loop.create_server()` already had before this function replaced it.
- A pending shutdown signal now interrupts a bind retry immediately instead of leaving the process running after its PID file is removed, and a failed ASGI startup (not just a failed bind) now exits non-zero instead of silently exiting 0 — both close silent-failure paths that were previously masked by uvicorn's own exit behavior.


api/scripts/wazuh_manager_apid.py
```diff
@@ -112,18 +128,127 @@ def warn_about_default_passwords():
...
+        try:
+            for host in hosts:
+                try:
+                    family = socket.getaddrinfo(host, port, proto=socket.IPPROTO_TCP)[0][0]
+                except socket.gaierror as exc:
+                    raise OSError(f"could not resolve host '{host}': {exc}") from exc
+                sock = socket.socket(family, socket.SOCK_STREAM)
+                # Tracked before it is bound, so the cleanup below also closes the socket whose
+                # own bind is the one that failed.
+                sockets.append(sock)
```
jotacarma90: **[S1 · blocking · introduced]** `socket.socket()` is outside the per-host `try`, so on a kernel booted with `ipv6.disable=1` the default `host: ['0.0.0.0', '::']` raises `EAFNOSUPPORT` here, falls to the outer `except OSError` (line 217) with an errno that is not `EADDRINUSE`, and apid exits 1 — it never starts, even with the port free. `asyncio.loop.create_server()`, which bound these sockets before, wraps socket creation in `except socket.error: continue` and tolerates only `EADDRNOTAVAIL` at `bind()`, so that host used to be skipped and IPv4 served normally. Worth fixing the same claim in the docstring, the comment at line 205 and `architecture.md`, which all assert the parity the other way round.

```suggestion
                try:
                    sock = socket.socket(family, socket.SOCK_STREAM)
                except OSError as exc:
                    # asyncio.loop.create_server() skips a family it cannot even open a socket
                    # for; a kernel booted with ipv6.disable=1 fails here with EAFNOSUPPORT and
                    # never reaches bind().
                    logger.warning(f"Skipping {host}:{port}, its address family is not "
                                   f"available on this system: {exc}")
                    continue
                # Tracked before it is bound, so the cleanup below also closes the socket whose
                # own bind is the one that failed.
                sockets.append(sock)
```

api/scripts/wazuh_manager_apid.py
```diff
@@ -112,18 +128,127 @@ def warn_about_default_passwords():
...
+        # bound these sockets before this function existed, raised for it rather than running a
+        # server that listens on nothing; the same message is kept.
+        raise OSError(f'could not bind on any address out of {hosts}')
+
+    for attempt in range(retries + 1):
+        sockets = []
+        try:
+            for host in hosts:
+                try:
+                    family = socket.getaddrinfo(host, port, proto=socket.IPPROTO_TCP)[0][0]
```
jotacarma90: **[S2 · blocking · introduced]** `[0][0]` keeps only the first address a host resolves to, while `asyncio.loop.create_server()` bound every one of them (`set(chain.from_iterable(infos))` across all hosts). With `host: ['localhost']` glibc returns `::1` first, so only `[::1]:55000` ends up bound — `IPV6_V6ONLY` included — and anything talking to `127.0.0.1:55000` gets ECONNREFUSED. Binding the host string instead of the resolved `sockaddr` also re-resolves inside `bind()`, and an empty `getaddrinfo()` result would raise an `IndexError` no `except OSError` here catches.

Iterating the addrinfo list fixes this and F1 at once (resolving once before the retry loop would also stop a permanently unavailable family from logging six identical `Skipping` warnings per cycle):

```python
            for host in hosts:
                try:
                    infos = socket.getaddrinfo(host, port, type=socket.SOCK_STREAM,
                                               proto=socket.IPPROTO_TCP, flags=socket.AI_PASSIVE)
                except socket.gaierror as exc:
                    raise OSError(f"could not resolve host '{host}': {exc}") from exc
                for family, sock_type, proto, _, sockaddr in infos:
                    try:
                        sock = socket.socket(family, sock_type, proto)
                    except OSError as exc:
                        logger.warning(f"Skipping {sockaddr[0]}:{port}, its address family is "
                                       f"not available on this system: {exc}")
                        continue
                    sockets.append(sock)
                    sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
                    if family == socket.AF_INET6:
                        sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 1)
                    try:
                        sock.bind(sockaddr)
                    except OSError as exc:
                        if exc.errno in (errno.EADDRNOTAVAIL, errno.EAFNOSUPPORT):
                            logger.warning(f"Skipping {sockaddr[0]}:{port}, its address family "
                                           f"is not available on this system: {exc}")
                            sockets.pop()
                            sock.close()
                            continue
                        raise
```

`create_server()` deduplicated with a `set`, so a host repeated in the list is worth guarding against too.

api/scripts/wazuh_manager_apid.py
```diff
@@ -197,19 +322,29 @@ async def _warm_up_pools():
...
-            logger.error(exc)
-            raise exc
+            raise error from exc
+        logger.error(exc)
+        raise exc
+
+    config = uvicorn.Config(app, **{key: value for key, value in params.items()
+                                    if key not in ('host', 'port')})
+    server = uvicorn.Server(config)
+    server.run(sockets=sockets)
```
jotacarma90: **[S2 · blocking · introduced]** `uvicorn.Config()` and `Server.run()` are now outside the `try/except OSError` that used to wrap `uvicorn.run()`, so an `OSError` raised by uvicorn's own startup no longer reaches `api.log`. `Config.load()` runs inside `Server.serve()` and calls `load_cert_chain` / `load_verify_locations` / `set_ciphers`; `ssl.SSLError` subclasses `OSError`, and `configure_ssl()` does not chown `ssl_ca_certs`. It then hits `__main__`'s `except Exception`, which only `print()`s, and `pyDaemon()` dups `/dev/null` over stdout (`pyDaemonModule.py:53-58`) — the same silent-failure class this PR exists to close, one layer up.

```suggestion
    try:
        config = uvicorn.Config(app, **{key: value for key, value in params.items()
                                        if key not in ('host', 'port')})
        server = uvicorn.Server(config)
        server.run(sockets=sockets)
    except OSError as exc:
        # __main__'s handler only print()s, and pyDaemon() sends this process's stdout to
        # /dev/null, so an unlogged OSError here leaves nothing in api.log.
        logger.error(exc)
        raise
```

src/init/templates/60-wazuh-manager-api-port.conf
```diff
@@ -0,0 +1,15 @@
...
+# net.ipv4.ip_local_reserved_ports holds one whole value, not a set each package adds to: this
+# line replaces whatever another sysctl.d drop-in or an administrator set for the same key, and a
+# drop-in that sorts after this one replaces this reservation in turn.
+#
+# The port is hardcoded to the packaged default. api.yaml's 'port' setting can change the port
+# apid binds, and this reservation does not follow it: an installation that changes the port ends
+# up with a reservation for a port nothing listens on, and none for the port it actually uses.
+# Reading the configured value at postinst time instead would need the reservation to be rewritten
+# whenever api.yaml changes, which nothing does today.
+net.ipv4.ip_local_reserved_ports = 55000
```
jotacarma90: **[S2 · blocking · introduced]** `net.ipv4.ip_local_reserved_ports` holds one whole value, so this drop-in silently discards any reservation an administrator or another package already set — a host with `/etc/sysctl.d/50-myapp.conf` reserving `9100-9200` loses it on install, and the `sysctl --system` in postinst makes that immediate rather than next-boot. That is the incident this PR fixes for apid, inflicted on a third party. The retry loop added in the same PR already covers the ephemeral-source-port trigger, and the reservation does not follow `api.yaml`'s `port`, so the file buys no guarantee it does not already have.

Simplest fix is to drop it along with its install lines in `debian/rules` (66-69), `wazuh-manager.spec` (`%install` 90-93, `%files` 557) and both postinst hooks. If the reservation is worth keeping, compute it instead of writing the key whole:

```sh
API_PORT=$(awk '/^port:/ {print $2}' ${DIR}/api/configuration/api.yaml 2>/dev/null)
API_PORT=${API_PORT:-55000}
CURRENT=$(sysctl -n net.ipv4.ip_local_reserved_ports 2>/dev/null)
case ",${CURRENT}," in
  *",${API_PORT},"*) ;;
  *) NEW=$([ -n "${CURRENT}" ] && echo "${CURRENT},${API_PORT}" || echo "${API_PORT}")
     printf 'net.ipv4.ip_local_reserved_ports = %s\n' "${NEW}" > /etc/sysctl.d/60-wazuh-manager-api-port.conf
     sysctl -p /etc/sysctl.d/60-wazuh-manager-api-port.conf > /dev/null 2>&1 || true ;;
esac
```

packages/debs/SPECS/wazuh-manager/debian/postinst
```diff
@@ -265,6 +265,11 @@ case "$1" in
         fi
     done
 
+    # Apply the API port reservation shipped in /usr/lib/sysctl.d.
+    if command -v sysctl > /dev/null 2>&1; then
+        sysctl --system > /dev/null 2>&1 || true
```
jotacarma90: **[S3 · non-blocking · introduced]** `sysctl --system` re-applies every sysctl source on the machine — `/etc/sysctl.d/`, `/run/sysctl.d/`, `/usr/lib/sysctl.d/`, `/usr/local/lib/sysctl.d/` and `/etc/sysctl.conf` — so a routine install or upgrade applies unrelated `vm.*`/`net.core.*`/`kernel.*` tuning an administrator staged for the next reboot. Same in `wazuh-manager.spec:520-522`. Moot if the drop-in goes away per the other comment.

```suggestion
    # Apply the API port reservation shipped in /usr/lib/sysctl.d.
    if command -v sysctl > /dev/null 2>&1; then
        sysctl -p /usr/lib/sysctl.d/60-wazuh-manager-api-port.conf > /dev/null 2>&1 || true
```

api/scripts/wazuh_manager_apid.py
```diff
@@ -240,7 +375,13 @@ def version():
...
 
 def exit_handler(signum, frame):
-    """Try to kill API child processes and remove their PID files."""
+    """Release a pending bind retry, then try to kill API child processes and remove their PID files."""
+    # Before the pidfiles go away: a bind retry still waiting would otherwise keep this process
+    # alive and invisible to wazuh-manager-control for the rest of its backoff budget.
+    # Event.set() takes the Event's own condition lock, which Event.wait() holds only while
+    # arming and disarming its waiter, not for the timeout itself -- so replacing this with a
+    # plain flag plus a signal-safe wakeup buys nothing and loses the interruptible wait.
+    _shutdown_event.set()
```
jotacarma90: **[S3 · non-blocking · introduced]** `threading.Event` builds its `Condition` over a plain `Lock`, not an `RLock`, and `Event.wait()` does hold `_cond._lock` in two windows: while arming the waiter, and while re-acquiring it in `_acquire_restore()` on the way out. A SIGTERM delivered in either runs this handler on the same main thread, `set()` blocks on a lock only that now-suspended thread can release, and apid hangs forever with its pidfiles already deleted. The window is a few bytecodes out of a multi-second wait — the blocking `waiter.acquire()` itself is safe — so this is unlikely rather than impossible, but the comment asserts it cannot happen. A self-pipe is the signal-safe form:

```python
# module level
_shutdown_r, _shutdown_w = os.pipe()
os.set_blocking(_shutdown_w, False)

# exit_handler
    try:
        os.write(_shutdown_w, b'x')
    except BlockingIOError:
        pass

# _bind_listening_sockets, replacing `if _shutdown_event.wait(wait_time):`
    if select.select([_shutdown_r], [], [], wait_time)[0]:
```

tests/integration/test_manager_control/test_apid_bind_retry.py
```diff
@@ -0,0 +1,412 @@
...
+                f"'{RETRY_WARNING}' never appeared in api.log: apid did not retry the busy port "
+                f'(#39231 regression). Log since the start:\n{_log_since(offset)}'
+            )
+
+        wait_expected_daemon_status(target_daemon=API_DAEMON, running_condition=True,
+                                    timeout=TOTAL_BIND_BUDGET_SECONDS)
+        status = check_all_daemon_status()
+        assert status.get(API_DAEMON) is True, (
+            f"'{API_DAEMON}' did not recover after the port was released: {status}"
+        )
```
jotacarma90: **[S3 · non-blocking · introduced]** These two assertions cannot fail: `create_pid()` runs at `wazuh_manager_apid.py:521`, before `start()`, and `pstatus` in `src/init/wazuh-server.sh` reads the pidfile, so apid reports `running` for the entire retry window — as this PR's own `architecture.md` says. `wait_expected_daemon_status(running_condition=True)` returns on the first poll and `status.get(API_DAEMON) is True` passes even if apid never binds and dies 60s later. Only the later `Listening on` check has any power; consider asserting that first, or probing the API itself (`login()` against `localhost:port`) and keeping the daemon status as a sanity check.

tests/integration/test_manager_control/test_apid_bind_retry.py
```diff
@@ -0,0 +1,412 @@
...
+            assert retrying, 'Precondition failed: no apid process was alive to signal'
+
+            # SIGTERM is exactly what 'wazuh-manager-control stop' sends apid. It is sent
+            # directly here because a concurrent control 'stop' would contend with the control
+            # 'start' still in flight above for the start-script lock.
+            for proc in retrying:
+                proc.send_signal(signal.SIGTERM)
+
+            gone, alive = psutil.wait_procs(retrying, timeout=STOP_TIMEOUT_SECONDS)
+            assert not alive, (
```
jotacarma90: **[S3 · non-blocking · introduced]** `_apid_processes()` matches on cmdline, and the `ProcessPoolExecutor` created at `wazuh_manager_apid.py:263` — before the bind, with Linux's default `fork` start method — gives its worker the parent's cmdline, so this signals the worker too. The worker inherits `exit_handler`, which deletes pidfiles and *returns* without exiting (`spawn_process_pool_worker` only sets SIGINT to `SIG_IGN`), so `assert not alive` depends on `concurrent.futures`' atexit teardown winning a race inside `STOP_TIMEOUT_SECONDS`. Filtering to the PID in apid's pidfile would make it deterministic. Also, the docstring's `expected_output` names `'Shutdown requested while waiting to retry the bind'` but nothing asserts it, so the test passes just as well if apid died of something else.

api/api/test/test_wazuh_manager_apid.py
```diff
@@ -0,0 +1,429 @@
...
+    to share the port at all.
+    """
+    probe = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    probe.bind(('127.0.0.1', 0))
+    port = probe.getsockname()[1]
+    probe.close()
+
+    sockets = apid._bind_listening_sockets(['0.0.0.0', '::'], port)  # nosec B104
+    try:
+        assert [sock.getsockname()[:2] for sock in sockets] == [('0.0.0.0', port), ('::', port)]
```
jotacarma90: **[S3 · non-blocking · introduced]** This is the only case that touches a real kernel, and it is not bounded to one. On a runner without IPv6 either `socket.socket(AF_INET6, ...)` raises `EAFNOSUPPORT` (the F1 path, so the test errors) or `bind('::')` gives `EADDRNOTAVAIL`, the host is skipped and the `== [('0.0.0.0', port), ('::', port)]` assertion fails. Separately, the port is learned by binding a probe to port 0 and closing it, so anything else — a parallel pytest-xdist worker included — can take it before `_bind_listening_sockets` runs, giving a spurious `EADDRINUSE`. A `pytest.mark.skipif(not socket.has_ipv6, ...)` and keeping the probe open until just before the bind would settle both.

api/scripts/wazuh_manager_apid.py
jotacarma90: **[S3 · non-blocking · aggravated]** A SIGTERM that lands outside `_shutdown_event.wait()` — right after the wait returns `False`, during a `bind()` that then succeeds, or earlier in `start()` — runs `exit_handler`, which sets the event and deletes the pidfiles but does not terminate the process. The success path never checks the event, and `Server.run()` installs uvicorn's own handlers via `capture_signals()` without knowing about the earlier request. The result is apid serving on 55000 with no pidfile: `wazuh-manager-control status` reports it stopped, `stop` cannot find it, and the next `start` gets `EADDRINUSE` from an orphaned apid — the very condition this PR retries against, now burning the whole 62 s budget. The pre-existing window (between `signal.signal(SIGTERM, exit_handler)` and uvicorn's `capture_signals()`) is what the PR body says it closes; it is only closed inside the wait. Since the event exists now, one check before starting the server covers the rest:

```suggestion
    if _shutdown_event.is_set():
        for sock in sockets:
            sock.close()
        logger.info('Shutdown requested before the API server started. Aborting startup')
        raise SystemExit(0)
    server = uvicorn.Server(config)
    server.run(sockets=sockets)
```

src/init/templates/60-wazuh-manager-api-port.conf
jotacarma90: **[S2 · blocking · introduced]** `net.ipv4.ip_local_reserved_ports` holds one whole value, so this drop-in silently discards any reservation an administrator or another package already set — a host with `/etc/sysctl.d/50-myapp.conf` reserving `9100-9200` loses it on install, and the `sysctl --system` in postinst makes that immediate rather than next-boot. That is the incident this PR fixes for apid, inflicted on a third party. The retry loop added in the same PR already covers the ephemeral-source-port trigger, and the reservation does not follow `api.yaml`'s `port`, so the file buys no guarantee it does not already have.

Simplest fix is to drop it along with its install lines in `debian/rules` (66-69), `wazuh-manager.spec` (`%install` 90-93, `%files` 557) and both postinst hooks. If the reservation is worth keeping, compute it instead of writing the key whole:

```sh
API_PORT=$(awk '/^port:/ {print $2}' ${DIR}/api/configuration/api.yaml 2>/dev/null)
API_PORT=${API_PORT:-55000}
CURRENT=$(sysctl -n net.ipv4.ip_local_reserved_ports 2>/dev/null)
case ",${CURRENT}," in
  *",${API_PORT},"*) ;;
  *) NEW=$([ -n "${CURRENT}" ] && echo "${CURRENT},${API_PORT}" || echo "${API_PORT}")
     printf 'net.ipv4.ip_local_reserved_ports = %s\n' "${NEW}" > /etc/sysctl.d/60-wazuh-manager-api-port.conf
     sysctl -p /etc/sysctl.d/60-wazuh-manager-api-port.conf > /dev/null 2>&1 || true ;;
esac
```

packages/debs/SPECS/wazuh-manager/debian/postinst
jotacarma90: **[S3 · non-blocking · introduced]** `sysctl --system` re-applies every sysctl source on the machine — `/etc/sysctl.d/`, `/run/sysctl.d/`, `/usr/lib/sysctl.d/`, `/usr/local/lib/sysctl.d/` and `/etc/sysctl.conf` — so a routine install or upgrade applies unrelated `vm.*`/`net.core.*`/`kernel.*` tuning an administrator staged for the next reboot. Same in `wazuh-manager.spec:520-522`. Moot if the drop-in goes away per the other comment.

```suggestion
    # Apply the API port reservation shipped in /usr/lib/sysctl.d.
    if command -v sysctl > /dev/null 2>&1; then
        sysctl -p /usr/lib/sysctl.d/60-wazuh-manager-api-port.conf > /dev/null 2>&1 || true
```

api/scripts/wazuh_manager_apid.py
```diff
@@ -240,7 +411,13 @@ def version():
...
```
jotacarma90: **[S3 · non-blocking · introduced]** `threading.Event` builds its `Condition` over a plain `Lock`, not an `RLock`, and `Event.wait()` does hold `_cond._lock` in two windows: while arming the waiter, and while re-acquiring it in `_acquire_restore()` on the way out. A SIGTERM delivered in either runs this handler on the same main thread, `set()` blocks on a lock only that now-suspended thread can release, and apid hangs forever with its pidfiles already deleted. The window is a few bytecodes out of a multi-second wait — the blocking `waiter.acquire()` itself is safe — so this is unlikely rather than impossible, but the comment asserts it cannot happen. A self-pipe is the signal-safe form:

```python
# module level
_shutdown_r, _shutdown_w = os.pipe()
os.set_blocking(_shutdown_w, False)

# exit_handler
    try:
        os.write(_shutdown_w, b'x')
    except BlockingIOError:
        pass

# _bind_listening_sockets, replacing `if _shutdown_event.wait(wait_time):`
    if select.select([_shutdown_r], [], [], wait_time)[0]:
```

tests/integration/test_manager_control/test_apid_bind_retry.py
```diff
@@ -0,0 +1,423 @@
...
+            retrying = _apid_processes()
+            assert retrying, 'Precondition failed: no apid process was alive to signal'
+
+            # SIGTERM is exactly what 'wazuh-manager-control stop' sends apid. It is sent
+            # directly here because a concurrent control 'stop' would contend with the control
+            # 'start' still in flight above for the start-script lock.
+            for proc in retrying:
+                proc.send_signal(signal.SIGTERM)
+
+            gone, alive = psutil.wait_procs(retrying, timeout=STOP_TIMEOUT_SECONDS)
```
jotacarma90: **[S3 · non-blocking · introduced]** `_apid_processes()` matches on cmdline, and the `ProcessPoolExecutor` created at `wazuh_manager_apid.py:263` — before the bind, with Linux's default `fork` start method — gives its worker the parent's cmdline, so this signals the worker too. The worker inherits `exit_handler`, which deletes pidfiles and *returns* without exiting (`spawn_process_pool_worker` only sets SIGINT to `SIG_IGN`), so `assert not alive` depends on `concurrent.futures`' atexit teardown winning a race inside `STOP_TIMEOUT_SECONDS`. Filtering to the PID in apid's pidfile would make it deterministic. Also, the docstring's `expected_output` names `'Shutdown requested while waiting to retry the bind'` but nothing asserts it, so the test passes just as well if apid died of something else.

api/scripts/wazuh_manager_apid.py
jotacarma90: **[S3 · non-blocking · introduced]** `threading.Event` builds its `Condition` over a plain `Lock`, not an `RLock`, and `Event.wait()` does hold `_cond._lock` in two windows: while arming the waiter, and while re-acquiring it in `_acquire_restore()` on the way out. A SIGTERM delivered in either runs this handler on the same main thread, `set()` blocks on a lock only that now-suspended thread can release, and apid hangs forever with its pidfiles already deleted. The window is a few bytecodes out of a multi-second wait — the blocking `waiter.acquire()` itself is safe — so this is unlikely rather than impossible, but the comment asserts it cannot happen. A self-pipe is the signal-safe form:

```python
# module level
_shutdown_r, _shutdown_w = os.pipe()
os.set_blocking(_shutdown_w, False)

# exit_handler
    try:
        os.write(_shutdown_w, b'x')
    except BlockingIOError:
        pass

# _bind_listening_sockets, replacing `if _shutdown_event.wait(wait_time):`
    if select.select([_shutdown_r], [], [], wait_time)[0]:
```

tests/integration/test_manager_control/test_apid_bind_retry.py
jotacarma90: **[S3 · non-blocking · introduced]** `_apid_processes()` matches on cmdline, and the `ProcessPoolExecutor` created at `wazuh_manager_apid.py:263` — before the bind, with Linux's default `fork` start method — gives its worker the parent's cmdline, so this signals the worker too. The worker inherits `exit_handler`, which deletes pidfiles and *returns* without exiting (`spawn_process_pool_worker` only sets SIGINT to `SIG_IGN`), so `assert not alive` depends on `concurrent.futures`' atexit teardown winning a race inside `STOP_TIMEOUT_SECONDS`. Filtering to the PID in apid's pidfile would make it deterministic. Also, the docstring's `expected_output` names `'Shutdown requested while waiting to retry the bind'` but nothing asserts it, so the test passes just as well if apid died of something else.

jotacarma90 (approved):

PR #39324, opened 2026-09-16
Compare port objects directly when deduplicating macOS ports

### LINKED ISSUE #39167: macOS getPorts deduplicates with an O(n^2) scan that re-serialises the candidate port to JSON on every comparison
## Description

`SysInfo::getPorts()` in `src/data_provider/src/sysInfoMac.cpp` removes duplicate ports with a linear scan over everything accumulated so far, and inside the comparator it serialises the candidate to a JSON string on every single comparison:

```cpp
for (const auto& processInfo : fdMap)
{
    for (const auto& fdSocket : processInfo.second )
    {
        nlohmann::json port;
        std::make_unique<PortImpl>(std::make_shared<BSDPortWrapper>(processInfo.first, fdSocket))->buildPortData(port);

        const auto portFound
        {
            std::find_if(ports.begin(), ports.end(),
                         [&port](const auto & element)
            {
                return 0 == port.dump().compare(element.dump());
            })
        };

        if (ports.end() == portFound)
        {
            ports.push_back(port);
        }
    }
}
```

Two problems compound in that comparator.

`port.dump()` does not depend on `element`. It is loop-invariant with respect to the scan, yet it is recomputed for every element examined, so inserting the *n*-th unique port performs *n* serialisations of the same object plus *n* serialisations of the existing ones. Across a full collection that is O(n^2) full JSON serialisations rather than O(n^2) cheap comparisons, and serialising a JSON object is far from cheap.

Comparing objects by their serialised text is also fragile in its own right. It makes equality depend on key ordering and on numeric and string formatting inside the serialiser, rather than on the values themselves, so a future change to how `buildPortData` populates the object can silently change what counts as a duplicate.

A single `const auto portDump = port.dump();` hoisted out of the lambda removes the repeated work, and comparing the JSON objects directly (`port == element`) removes both problems at once.

I want to be straightforward about the limits of this report. This is a defect I observed directly in the source and quoted above, not a runtime symptom I measured. I did not benchmark it, and on the host under test it is harmless: the agent reported only 8 port rows, so the scan is trivial. Two things make it worth recording anyway. The cost grows quadratically with the number of sockets, so a Mac with a large number of open sockets pays it on every inventory scan, and the current small port count is itself partly an artefact of wazuh/wazuh#39126, where the truncated process enumeration hides most sockets. If that defect is fixed and the full socket set becomes visible, this code path gets the full quadratic cost with it.

## Environment

- Source read from the `5.0.0` branch of `wazuh/wazuh`, `src/data_provider/src/sysInfoMac.cpp`, in `SysInfo::getPorts()`.
- Agent where the current port count was observed: `demo-env-macos-agent`, id `006`, macOS 26.5.1 build 25F80, arm64, 8 rows in `dbsync_ports`.
- Agent package `wazuh-agent-5.0.0-latest.arm64.pkg` from `nightly-backup/2026-09-09`.
- Related: wazuh/wazuh#39126 currently limits how many sockets this function can see at all.

## Steps to reproduce

1. Read the function in the 5.0.0 source and observe the comparator:

```bash
sed -n '/nlohmann::json SysInfo::getPorts/,/^}/p' src/data_provider/src/sysInfoMac.cpp
```

`port.dump()` appears inside the `find_if` predicate, so it is evaluated once per element examined rather than once per candidate.

2. To see the cost grow, run the collection on a macOS host with many open sockets. Generating a few thousand sockets, for example with a script that opens many listening or connected TCP sockets, and then timing a syscollector evaluation against a baseline shows the effect:

```bash
# force an evaluation and time the syscollector phase
sudo /Library/Ossec/bin/wazuh-control restart
grep -E "syscollector: INFO: (Starting evaluation|Evaluation finished)" /Library/Ossec/logs/ossec.log | tail -2
```

3. Compare the elapsed time between the two log lines with a low-socket baseline. Note that wazuh/wazuh#39126 caps the processes enumerated and therefore the sockets reachable, so that defect should be addressed first for this to be measured cleanly.

## Expected

Duplicate detection costs one serialisation per candidate at most, or avoids serialisation entirely by comparing the JSON objects, so the collection scales linearly in the common case and does not depend on serialised text for equality.

## Actual

The candidate is re-serialised on every comparison during an O(n^2) scan, so the work grows quadratically in serialisations, and duplicate detection depends on the exact serialised representation rather than on the values.

## Not a duplicate of

`qa-known-issues.py check` found **no cache match**, the live search for `re-serialises deduplicates comparison candidate getports squared` returned nothing, and the tool concluded "no known issue looks like this one".

Miguevrgo:
|Related issue|
|---|
|#39167|

## Description

`SysInfo::getPorts()` in `src/data_provider/src/sysInfoMac.cpp:377` dropped duplicate ports with a linear scan whose predicate serialised the candidate on every comparison:

```cpp
std::find_if(ports.begin(), ports.end(),
             [&port](const auto & element)
{
    return 0 == port.dump().compare(element.dump());
})
```

`port.dump()` does not depend on `element`, so it is loop-invariant with respect to the scan yet recomputed for each element examined. Inserting the n-th unique port performed n serialisations of the candidate plus n serialisations of the already accumulated ports, making the collection O(n^2) in full JSON serialisations rather than O(n^2) cheap comparisons.

Comparing the serialised text also made equality depend on key ordering and on the serialiser's numeric and string formatting instead of on the values, so a future change in how `buildPortData()` populates the object could silently change what counts as a duplicate.

Closes #39167

## Proposed changes

Replaced the `find_if` predicate with `std::find` using `nlohmann::json`'s `operator==`:

```cpp
if (ports.end() == std::find(ports.begin(), ports.end(), port))
{
    ports.push_back(port);
}
```

Duplicate detection now compares the JSON values directly and serialises nothing. The scan remains linear per candidate, but each comparison is cheap. A hash-based index would make it linear overall; it was not added because any hash key would have to be derived from a serialised form, reintroducing the text-dependent equality this change removes, and macOS hosts currently report single-digit port counts.

Behaviour is unchanged: two ports that produced identical JSON text also compare equal as objects.

## Configuration options

N/A. No new configuration parameters.

## Logs/Alerts example

N/A. The change is a refactor of the deduplication path with no user-visible log or alert changes. No benchmark was captured: the reporting host showed 8 rows in `dbsync_ports`, and wazuh/wazuh#39126 still caps how many sockets this function can see, so the quadratic cost cannot be measured cleanly until that is addressed.

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests


src/data_provider/src/sysInfoMac.cpp
```diff
@@ -364,28 +364,20 @@ static void getProcessesSocketFD(std::map<ProcessInfo, std::vector<std::shared_p
...
-                             [&port](const auto & element)
-                {
-                    return 0 == port.dump().compare(element.dump());
-                })
-            };
-
-            if (ports.end() == portFound)
-            {
-                ports.push_back(port);
+                ports.push_back(std::move(port));
```
MarcelKemp: **[S3 · non-blocking · introduced]** No test exercises `SysInfo::getPorts()`'s dedup path (neither before nor after this change). Since this predicate is exactly what the PR changes, consider adding a test with the existing mockable `IPortWrapper`/`PortImpl` seam that asserts two identical ports collapse to one and two distinct ports don't, to lock down the equality contract for future changes to `buildPortData`.

MarcelKemp (approved):

PR #39327, opened 2026-09-16
Capture the macOS ULS timestamp only from long enough lines

### LINKED ISSUE #39171: macOS ULS reader copies a fixed 26-byte timestamp from every log line without checking the line is that long, so a short line captures stale buffer bytes
## Description

`read_macos()` in `src/logcollector/src/read_macos.c` records the timestamp of the last log it forwarded, which is later persisted and used to resume collection with `log show --start` after a restart. It captures that timestamp with a fixed-size copy whose only guard is that the line is not empty:

```c
size = strlen(read_buffer);
if (size > 0) {
    /* Check ignore and restrict log regex, if configured. */
    if (check_ignore_and_restrict(lf->regex_ignore, lf->regex_restrict, read_buffer)) {
        continue;
    }

    w_msg_hash_queues_push(read_buffer, MACOS_LOG_NAME, size + 1, lf->log_target, LOCALFILE_MQ);
    memcpy(full_timestamp, read_buffer, OS_LOGCOLLECTOR_TIMESTAMP_FULL_LEN);
} else {
    mdebug2("macOS ULS: Discarding empty message.");
}
```

`OS_LOGCOLLECTOR_TIMESTAMP_FULL_LEN` is 26, the length of a macOS ULS timestamp such as `2020-11-09 05:45:08.000000-0800`. The copy is unconditional on `size`, so when a line shorter than 26 characters arrives, the copy takes bytes from beyond that line's terminator. Those bytes are whatever the previous, longer line left in `read_buffer`, since the buffer is reused across iterations and is not cleared between lines.

This is not a memory-safety fault. `read_buffer` is declared `char read_buffer[OS_MAXSTR + 1]`, so reading 26 bytes from offset 0 always stays inside the allocation. The problem is correctness: `full_timestamp` can end up holding a mixture of the current short line and the tail of an earlier one.

What happens next depends on whether that mixture contains a NUL:

```c
STATIC char * w_macos_trim_full_timestamp(const char * full_timestamp) {
    char * short_timestamp = NULL;

    if (w_strlen(full_timestamp) == OS_LOGCOLLECTOR_TIMESTAMP_FULL_LEN) {
        os_calloc(OS_LOGCOLLECTOR_TIMESTAMP_SHORT_LEN + 1, sizeof(char), short_timestamp);
        memcpy(short_timestamp, full_timestamp, OS_LOGCOLLECTOR_TIMESTAMP_BASIC_LEN);
        memcpy(short_timestamp + OS_LOGCOLLECTOR_TIMESTAMP_BASIC_LEN,
               full_timestamp + OS_LOGCOLLECTOR_TIMESTAMP_BASIC_LEN + OS_LOGCOLLECTOR_TIMESTAMP_MS_LEN,
               OS_LOGCOLLECTOR_TIMESTAMP_TZ_LEN);
    }

    return short_timestamp;
}
```

The length check rejects the case where a NUL landed inside the 26 bytes, so the common outcome is simply that no timestamp is recorded for that cycle. The case that matters is when the stale tail happens to fill all 26 bytes without a NUL: the length check passes, and a timestamp assembled from two different lines is accepted and stored, then written to `file_status.json` and used as the `--start` value on the next resume.

Short lines are reachable because the child process has both its streams bound:

```c
lf->macos_log->processes.stream.wfd = w_macos_log_exec(log_stream_array, W_BIND_STDOUT | W_BIND_STDERR);
```

so diagnostics emitted by `/usr/bin/log` on stderr are read through the same path as log records, and `w_macos_log_getlog()` fills the buffer with `fgets` without imposing any minimum length.

I want to be straightforward about the evidence. The defect is directly visible in the source and quoted above; I did not reproduce a corrupted timestamp on a running agent, because the macOS host became unreachable before I could inject a short line on that stream. On the agent while it was reachable the stored value was well formed, as expected for ordinary ULS output where every record begins with a full timestamp:

```json
{"macos":{"timestamp":"2026-09-10 07:26:30-0700","settings":"/usr/bin/log stream --style syslog ..."}}
```

The guard is a one-line change, `if (size >= OS_LOGCOLLECTOR_TIMESTAMP_FULL_LEN)` around the copy, or clearing `full_timestamp` when the line is too short so the length check reliably rejects it.

## Environment

- Source read from the `5.0.0` branch of `wazuh/wazuh`: `src/logcollector/src/read_macos.c`, `src/logcollector/src/macos_log.c`, and `src/logcollector/include/macos_log.h` for the length constants.
- Agent where the stored timestamp was observed: `demo-env-macos-agent`, id `006`, macOS 26.5.1 build 25F80, arm64, reading `/usr/bin/log stream --style syslog --type activity --type log --type trace --level info --predicate ...`.
- Agent package `wazuh-agent-5.0.0-latest.arm64.pkg` from `nightly-backup/2026-09-09`.

## Steps to reproduce

1. Read the capture site and confirm the guard is only `size > 0`:

```bash
sed -n '/^void \* read_macos/,/^}/p' src/logcollector/src/read_macos.c | grep -n -B4 "memcpy(full_timestamp"
```

2. Confirm the constant is 26 and that the trim function requires exactly that length:

```bash
grep -n "OS_LOGCOLLECTOR_TIMESTAMP_FULL_LEN\|TIMESTAMP_BASIC_LEN\|TIMESTAMP_TZ_LEN" src/logcollector/include/macos_log.h
```

3. To exercise it on a host, arrange for a line shorter than 26 characters to reach the reader. Since stderr of the `log` child is bound to the same stream, a predicate or option that makes `/usr/bin/log` emit a short diagnostic is the natural route; alternatively drive `w_macos_log_getlog` directly from a unit test with a short line following a long one.

4. Inspect the persisted value afterwards:

```bash
sudo cat /Library/Ossec/queue/logcollector/file_status.json
```

A timestamp that does not correspond to any emitted record indicates the stale-byte path was taken.

## Expected

The timestamp is captured only from lines that actually carry one, so `full_timestamp` always reflects a single real record, and a short or malformed line leaves the previously stored value untouched.

## Actual

26 bytes are copied from every non-empty line regardless of its length. A line shorter than 26 characters contributes bytes left over from an earlier line, and if the resulting 26 bytes contain no NUL the value passes the length check and is stored and persisted as the resume point.

## Not a duplicate of

`qa-known-issues.py check` found **no cache match** and concluded "no known issue looks like this one". The live search surfaced only `wazuh/wazuh#37532`, a spike about how FIM and Syscollector build initial container state for eBPF diffing, which is unrelated to macOS log reading.

Miguevrgo:
## Description

`read_macos()` copied a fixed 26 bytes (`OS_LOGCOLLECTOR_TIMESTAMP_FULL_LEN`) from every non-empty line into `full_timestamp`, guarded only by `size > 0`. `read_buffer` is reused between lines and never cleared, so a line shorter than 26 characters took stale bytes from the previous one. That value is persisted in `file_status.json` and used as `log show --start` on resume.

Closes #39171

## Proposed Changes

- `src/logcollector/read_macos.c`: guard the copy with `size >= OS_LOGCOLLECTOR_TIMESTAMP_FULL_LEN`. Short lines are still forwarded, but leave the stored timestamp untouched.
- `CHANGELOG.md`: entry under Agent / Fixed.

### Results and Evidence

Built on Linux (`make TARGET=agent DEBUG=yes TEST=1`), no new warnings. Existing suites pass: `test_read_macos` 56/56, `test_logcollector` 75/75.

No live macOS reproduction: it needs `/usr/bin/log` to emit a short diagnostic on stderr, and no macOS host was available.

- Compilation without warnings
  - [x] Linux — clean
  - [ ] Windows — N/A, not built on Windows
  - [ ] MAC OS X — not run, no macOS environment
- [x] Log syntax and correct language review — no log messages changed

- Memory tests — N/A, the change removes a read and adds no allocation

### Artifacts Affected

- macOS agent `wazuh-logcollector`.

### Configuration Changes

N/A

### Tests Introduced

N/A — one-line guard, covered by the existing `read_macos` suite.

## Review Checklist

- [ ] Code changes reviewed


src/logcollector/read_macos.c
```diff
@@ -135,7 +135,10 @@ void * read_macos(logreader * lf, int * rc, __attribute__((unused)) int drop_it)
             }
 
             w_msg_hash_queues_push(read_buffer, MACOS_LOG_NAME, size + 1, lf->log_target, LOCALFILE_MQ);
-            memcpy(full_timestamp, read_buffer, OS_LOGCOLLECTOR_TIMESTAMP_FULL_LEN);
+
+            if (size >= OS_LOGCOLLECTOR_TIMESTAMP_FULL_LEN) {
+                memcpy(full_timestamp, read_buffer, OS_LOGCOLLECTOR_TIMESTAMP_FULL_LEN);
+            }
```
MarcelKemp, line 141: **[S2 · blocking · introduced]** The guard measures length, not shape, so any line of 31+ characters still becomes the persisted resume point — issue #39171 asks that the timestamp be captured "only from lines that actually carry one" and that a short *or malformed* line leave the previous value untouched. 

```suggestion
            if (size >= OS_LOGCOLLECTOR_TIMESTAMP_FULL_LEN
                && w_expression_match(lf->macos_log->log_start_regex, read_buffer, NULL, NULL)) {
                memcpy(full_timestamp, read_buffer, OS_LOGCOLLECTOR_TIMESTAMP_FULL_LEN);
            }
```

CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ All notable changes to this project will be documented in this file.
 - Fixed FIM not monitoring user-mounted `tmpfs` directories mistaken for `/dev`. ([#38627](https://github.com/wazuh/wazuh/pull/38627))
 - Fixed the gcloud wodle silently discarding a crashed process's raw output when it produced no recognized log line, and fixed `wm_exec()` (shared by every wodle) reporting exit code 0 for a process killed by a signal, such as an OOM kill or a native segfault, which had made that raw-output fallback unreachable for exactly the crashes it was meant to catch. ([#38943](https://github.com/wazuh/wazuh/pull/38943))
 - Fixed the default Windows FIM configuration monitoring none of its 19 named critical binaries (`cmd.exe`, `lsass.exe`, `sc.exe`, `sethc.exe`, etc.), because duplicate `%WINDIR%\SysNative` / `%WINDIR%\System32` directory declarations collapsed onto the same path once normalized and silently replaced each other's `restrict` list. ([#39198](https://github.com/wazuh/wazuh/issues/39198))
+- Fixed the macOS ULS reader taking stale buffer bytes as the stored timestamp on lines shorter than 26 characters. ([#39171](https://github.com/wazuh/wazuh/issues/39171))
```
MarcelKemp: **[S3 · non-blocking · introduced]** `OS_LOGCOLLECTOR_TIMESTAMP_FULL_LEN` is 31, not 26, so this user-facing entry states the wrong threshold. `macos_log.h` defines `BASIC_LEN 19` + `TZ_LEN 5` + `MS_LEN 7`, and the documented sample `2020-11-09 05:45:08.000000-0800` is 31 characters; 26 is `BASIC_LEN + MS_LEN`, the offset the trim function skips to. Worth fixing in the PR description too.

```suggestion
- Fixed the macOS ULS reader taking stale buffer bytes as the stored timestamp on lines shorter than 31 characters. ([#39171](https://github.com/wazuh/wazuh/issues/39171))
```

MarcelKemp (dismissed):

src/logcollector/read_macos.c
Miguevrgo: I'd keep the length guard. Long malformed lines are not introduced here: they were copied the same way before this change, and the guard does not make that worse.

It is also hard to reach. `w_macos_log_getlog()` only starts a new record when a line matches `MACOS_LOG_START_REGEX` (`^\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d`), so a line without a timestamp is appended to the previous record as a continuation instead of becoming its own. The only exception is the remainder of a record that fills the buffer.

The bug this issue reports is the short line. `strlen()` guarantees a NUL at `read_buffer[size]`, so with `size < 31` that NUL always falls inside the copied bytes and `w_macos_trim_full_timestamp()` rejects the value. A timestamp mixed from two lines could never be accepted.

cborla:
No findings: the new length check only skips the timestamp copy for lines too short to contain one, so the last valid timestamp is kept.

PR #39335, opened 2026-09-16
Change default eBPF whodata loginuid to -1 instead of 0

### LINKED ISSUE #39333: FIM eBPF whodata: loginuid falls back to 0 on kernels without CONFIG_AUDIT, attributing every file change to root
|---|---|---|---|---|
| 4.14.8-rc1 | FIM / whodata / eBPF | Agent | Sources | Linux |

## Description

The eBPF whodata provider falls back to `0` when it cannot read `task_struct.loginuid`, and `0` is not a neutral value: it is root. On a kernel built without `CONFIG_AUDIT` the provider therefore reports `audit_uid="0"` and `audit_name="root"` for every file event, and because userspace resolves the login group from that same value it also reports `audit_gid="0"` and `audit_group_name="root"`. All four login attribution fields are filled in with root, silently, on every event.

The `loginuid` (`auid`) field exists precisely to survive a change of identity. It is the one part of a whodata event that `sudo` and `su` cannot alter, and it is what makes it possible to attribute a privileged file change to a person rather than to root. Reporting a fabricated `root` in that field does not degrade attribution, it inverts it: an analyst reading `login_user: root` stops looking, because the field claims to already answer the question.

The kernel has a dedicated sentinel for "unknown", `(uid_t)-1` (`4294967295`), and the Audit provider in this same codebase already detects and honours it. The eBPF provider does not.

### 1. The BPF program initialises the fallback to zero

`src/syscheckd/src/ebpf/src/modern.bpf.c:345-350`, and identically in `src/syscheckd/src/ebpf/src/modern-arm.bpf.c:263-268`:

```c
struct task_struct___local *task_local = (struct task_struct___local *)current_task;
kuid_t loginuid = {0};
if (bpf_core_field_exists(task_local->loginuid)) {
    loginuid = BPF_CORE_READ(task_local, loginuid);
}
evt->login_uid = loginuid.val;
```

The CO-RE detection itself is correct. The problem is the initialiser: when `bpf_core_field_exists()` is false the read is skipped and the zero is emitted as if it were a real login uid.

The comment directly above this block states that the code "skips gracefully if unavailable". It does not skip. It emits `0`.

### 2. Userspace resolves the zero into root without validating it

`src/syscheckd/src/ebpf/src/ebpf_whodata.cpp:678-685`:

```c
w_evt->audit_uid  = num_to_str(event->login_uid);              // "0"
w_evt->audit_name = fimebpf::instance().m_get_user(event->login_uid);   // "root"
int audit_gid_val = get_login_gid(event->login_uid);
if (audit_gid_val >= 0)
{
    w_evt->audit_gid        = num_to_str((unsigned int)audit_gid_val);   // "0"
    w_evt->audit_group_name = fimebpf::instance().m_get_group(audit_gid_val);  // "root"
}
```

The `audit_gid_val >= 0` check looks like a guard against an unresolvable id, but it does not catch this case. `get_login_gid()` (`ebpf_whodata.cpp:84-94`) is a `getpwuid_r()` lookup, and `getpwuid_r(0)` always succeeds because root always exists, returning `pw_gid = 0`. The guard passes and both group fields are populated with root as well.

### 3. The Audit provider already implements the correct behaviour

`src/syscheckd/src/whodata/audit_parse.c:575-582`:

```c
if (strcmp(auid, "4294967295") == 0) { // Invalid auid (-1)
    if (!auid_err_reported) {
        mdebug1(FIM_AUDIT_INVALID_AUID);
        auid_err_reported = 1;
    }
    w_evt->audit_name = NULL;
    w_evt->audit_uid = NULL;
}
```

It recognises the kernel sentinel, logs it once, and leaves both fields unset so that they are simply absent from the alert. The eBPF provider has no equivalent, and emits no log line at all when the value is unavailable.

## Affected scenarios

These behave differently and both are wrong, so a fix should cover both.

**A. Kernel built without `CONFIG_AUDIT`.** `bpf_core_field_exists()` returns false, `0` is emitted, and all four fields are reported as root. This is the misattribution case.

**B. Kernel with `CONFIG_AUDIT`, process with no login session** (containers, daemons, cron jobs). The field exists, so the kernel's own value is read and it is the sentinel `4294967295`. That value is passed through unvalidated, producing `audit_uid="4294967295"` with an unresolvable `audit_name`, where the Audit provider would have suppressed the field. Note that a container running on an ordinary host falls into this case rather than case A, since containers share the host kernel and what they lack is the login session, not the kernel option.

## Impact

Every file modification on an affected agent is attributed to root in the field whose entire purpose is attribution. The consequences are that the recorded actor is wrong for privileged changes made through a compromised or shared account, that the field loses all discriminating power for rules and dashboards because every event carries the same value, and that none of this is visible in the logs, since the provider treats `0` as a legitimate result and reports nothing.

## Affected versions

Present in `v4.14.8-rc1` and unchanged at the tips of `4.14.8`, `4.14.10` and `main` as of 2026-09-15. The fallback was introduced together with the login uid support added for #32467.

## Suggested fix

Initialise the fallback to the kernel's own sentinel instead of zero, in both BPF sources:

```c
kuid_t loginuid = { (uint32_t)-1 };
```

Then, in `ebpf_whodata.cpp`, skip all four fields (`audit_uid`, `audit_name`, `audit_gid`, `audit_group_name`) when `event->login_uid` equals the sentinel, mirroring what `audit_parse.c` does, and emit the equivalent one-shot debug message so the condition is diagnosable.

This single change resolves both scenarios above, because case A then produces the same sentinel that case B already produces, and one guard covers both.

## Relationship to #39332

These are independent defects on opposite sides of the wire, but they interact and the fix order matters.

#39332 is a manager-side defect: `audit_gid` and `audit_group_name` are discarded by the analysisd decoder. As a side effect, the two fabricated group values described here never reach the alert today, so the currently observable damage is limited to `syscheck.audit.login_user` reporting root.

If #39332 is fixed before this issue, the false attribution doubles: a login group of root would start appearing in alerts as well. The two should be scheduled with that dependency in mind.

Miguevrgo:
|Related issue|
|---|
|#39333|

## Description

The eBPF whodata provider initialized its `loginuid` fallback to `0`, which is root, not a neutral value. On kernels built without `CONFIG_AUDIT` the CO-RE check is false, the read is skipped and `0` is emitted as if it were a real login uid, attributing every file event to root.

`src/syscheckd/src/ebpf/src/modern.bpf.c:346` and `modern-arm.bpf.c:264` now default to `(__u32)-1`, the kernel sentinel for an unset auid.

Closes #39333

## Configuration options

N/A

## Logs/Alerts example

N/A

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests


src/syscheckd/src/ebpf/src/modern.bpf.c
```diff
@@ -343,7 +343,7 @@ statfunc void submit_event(const char *filename,
      * CONFIG_AUDIT is enabled at kernel build time. bpf_core_field_exists()
      * resolves this at load time and skips gracefully if unavailable. */
     struct task_struct___local *task_local = (struct task_struct___local *)current_task;
-    kuid_t loginuid = {0};
+    kuid_t loginuid = {(__u32)-1};
```
MarcelKemp: **[S2 · blocking · introduced]** This default change (the fix for the S1 root-misattribution bug in #39333) has no test in either direction — `src/syscheckd/src/ebpf/tests/fimEbpfWhodataTest` has zero references to `loginuid`/`4294967295`, before or after this PR. Please add a consumer-level test (mocking `event->login_uid = UINT32_MAX`) that fails without this change and also covers the sentinel-handling gap flagged on `ebpf_whodata.cpp:725` in the review summary.

MarcelKemp: **[S2 · blocking · aggravated]** The ARM64 sibling of this file, `src/syscheckd/src/ebpf/src/modern-arm.bpf.c:263`, still has `kuid_t loginuid = {0};` in the identical, copy-pasted loginuid block (lines 259-268), so on aarch64 agents with a `CONFIG_AUDIT`-less kernel the bug in #39333 is entirely unfixed — and the new userspace guard does not help, because `0 != (uint32_t)-1`, so those events still report `audit_uid="0"` / `audit_name="root"`. The PR description says `modern-arm.bpf.c:264` now defaults to `(__u32)-1`, but that file is not in the diff, so a reviewer or backporter reading the description will believe ARM is covered while `Closes #39333` retires the only tracking the ARM half has. Mirror this exact line there:

```c
    kuid_t loginuid = {(__u32)-1};
```

src/syscheckd/src/ebpf/src/ebpf_whodata.cpp
```diff
@@ -675,13 +675,20 @@ void ebpf_pop_events(fim::BoundedQueue<std::unique_ptr<dynamic_file_event>>& loc
...
-                w_evt->audit_group_name = fimebpf::instance().m_get_group(audit_gid_val);
+                w_evt->audit_uid = num_to_str(event->login_uid);
+                w_evt->audit_name = fimebpf::instance().m_get_user(event->login_uid);
+                int audit_gid_val = get_login_gid(event->login_uid);
+                if (audit_gid_val >= 0)
+                {
+                    w_evt->audit_gid = num_to_str((unsigned int)audit_gid_val);
+                    w_evt->audit_group_name = fimebpf::instance().m_get_group(audit_gid_val);
+                }
             }
```
MarcelKemp: **[S3 · non-blocking · introduced]** The guard drops the four audit fields silently, so an operator on a `CONFIG_AUDIT`-less kernel sees `audit_uid`, `audit_name`, `audit_gid` and `audit_group_name` vanish from every whodata alert with nothing in the logs to explain it. The comment says this matches `audit_parse.c`, but that provider also emits a one-shot `mdebug1(FIM_AUDIT_INVALID_AUID)` (`audit_parse.c:576-579`) — the half that turns an unexplained disappearance into a diagnostic. A one-shot debug line here would close the gap:

```suggestion
            /* The kernel reports (uid_t)-1 when task_struct.loginuid is unavailable,
             * which is the case on kernels built without CONFIG_AUDIT. Leave the audit
             * attribution unset instead of publishing the raw sentinel, matching what
             * the Audit provider does in audit_parse.c. */
            if (event->login_uid == (uint32_t)-1)
            {
                static bool auid_err_reported = false;
                if (!auid_err_reported && fimebpf::instance().m_loggingFunction)
                {
                    fimebpf::instance().m_loggingFunction(
                        LOG_DEBUG,
                        "eBPF whodata: the kernel reports no loginuid (auid); audit "
                        "attribution fields will be omitted. The kernel is likely "
                        "built without CONFIG_AUDIT.");
                    auid_err_reported = true;
                }
            }
            else
            {
                w_evt->audit_uid = num_to_str(event->login_uid);
                w_evt->audit_name = fimebpf::instance().m_get_user(event->login_uid);
                int audit_gid_val = get_login_gid(event->login_uid);
                if (audit_gid_val >= 0)
                {
                    w_evt->audit_gid = num_to_str((unsigned int)audit_gid_val);
                    w_evt->audit_group_name = fimebpf::instance().m_get_group(audit_gid_val);
                }
            }
```

MarcelKemp (approved):

Miguevrgo:
## Evidence

Ubuntu 24.04, kernel 6.8.0-137, x86_64. Agent built from `4.14.8` plus the fixes for this issue and #39358, with the `libbpf-bootstrap` precompiled dependency regenerated from that source (`/var/ossec/lib/modern.bpf.o` = `e2daa4a3011351362336b65c014b8fca...`).

eBPF whodata healthcheck, both hook paths:

```
(6047): Initializing eBPF driver for FIM whodata.
(6052): BPF LSM is not active (...); falling back to kprobe hooks.
(6049): eBPF healthcheck action succeeded: create file / modify content / modify metadata / delete file.
(6048): Healthcheck for eBPF FIM whodata module success.
```

```
(6051): BPF LSM is active in the running kernel; using LSM hooks for create/modify/delete events.
(6049): eBPF healthcheck action succeeded: create file / modify content / modify metadata / delete file.
(6048): Healthcheck for eBPF FIM whodata module success.
```

File touched by a process with no login session (`systemd-run`, `/proc/self/loginuid` = `4294967295`). The audit attribution is now omitted instead of being reported as root:

```json
{
  "path": "/etc/wtest/ev_noauid",
  "mode": "whodata",
  "audit": {
    "user_id": "0",
    "user_name": "root",
    "process_name": "touch",
    "process_id": 5923,
    "group_id": "0",
    "group_name": "root",
    "effective_uid": "0",
    "effective_name": "root",
    "parent_name": "systemd",
    "ppid": 1
  }
}
```

No `audit_uid`, `audit_name`, `audit_gid` or `audit_group_name`. For a file touched from a real login session the fields are still populated (`audit_uid: "1000"`, `audit_name: "ubuntu"`), so attribution is only dropped where the kernel has none.

PR #39352, opened 2026-09-16
VD scan lane treats a vulnerability scanner that is still starting as disabled

### LINKED ISSUE #39346: VD scan lane treats a vulnerability scanner that is still starting as disabled
## Description

`VulnerabilityScannerFacade` only ever exposed `isInitialized()`, which becomes `true` at the very
end of `start()` (after schema/DB/content-updater validation). Nothing distinguishes "start() was
never invoked at all" (vulnerability detection disabled, or a harness that skips it) from "start()
is running right now, still validating" — both read `isInitialized() == false`.

`wazuh-manager-modulesd` spawns each wmodule on its own thread with no ordering guarantee between
them (`main.c`, `CreateThreadJoinable`), so on every manager restart there is a real window —
normally brief, but real — where `wm_inventory_sync_server` is already serving sessions while
`wm_vulnerability_scanner`'s own `start()` is still executing.

In that window:
- A VD sync session (`VDFirst`/`VDSync`) is routed through the D22 legitimate-skip path
  (`vdScannerAdapter.cpp`'s `feedReady()` returning `true` whenever `!isInitialized()`): the
  agent's inventory is indexed without a scan, and no rescan happens until the next feed-offset
  bump. Correct behavior for "disabled," wrong for "about to start."
- An on-demand scan request (`PUT /agents/scan/vulnerability` → `/scan/vd`) answers a *permanent*
  "no vulnerability scanner on this node" with a misleading operator-facing log line and the
  `vd.scans.skipped` metric, instead of a transient, retryable condition. Task Manager's own retry
  (30 s) recovers the scan, but the log and metric are wrong for the duration.

Not a regression from any single commit — it's a gap that's existed since the two-wmodule split
between `inventory_sync_server` and `vulnerability_scanner` (the `feedReady()` gate has always
conflated these two states via `isInitialized()` alone).

## Fix direction

Add `hasStarted()`/`isEnabled()` accessors to `VulnerabilityScannerFacade`, read *before*
`isInitialized()`, mirroring `m_startInvoked`/`m_enabled` (already private members, just not
exposed). The adapter's gate becomes: never going to run here (disabled, or `hasStarted() ==
false`) → legitimate skip, unchanged; enabled and started but not yet initialized → defer with the
existing feed-not-ready retryable `503`, instead of routing it through the skip path.

Code references at `5.0.0` `8ae0097326`:
- `src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScannerFacade.hpp` /`.cpp` — only
  `isInitialized()` exposed; `m_startInvoked`/`m_enabled` already tracked internally.
- `src/wazuh_modules/inventory_sync_server/src/vd/vdScannerAdapter.cpp:43` —
  `feedReady()`'s `!scanner.isInitialized() || scanner.isFeedReady()` conflates the two states.
- `src/wazuh_modules/src/main.c` — each wmodule's `start` dispatched via `CreateThreadJoinable`
  with no cross-module ordering.

## Related

- #38880 (surfaced while implementing that issue's Retry-After work; not itself one of its four numbered findings, and not required to close it — kept separate).
- #38599 (established the D22 legitimate-skip path this gate mis-applies during the startup race).

jepalfer:
## Description

`VulnerabilityScannerFacade` only ever exposed `isInitialized()`, which only becomes `true` at the very end of `start()`. Since `wm_vulnerability_scanner` and `wm_inventory_sync_server` are separate wmodules started on independent threads with no ordering guarantee between them, a session arriving during the (normally brief) startup window right after a manager restart was indistinguishable from one arriving at a node where vulnerability detection is disabled: `isInitialized()` reads `false` in both cases.

That ambiguity had two visible symptoms:
- A VD sync session took the legitimate-skip path (indexed without a scan) instead of getting a retryable `503`, silently missing scan coverage for that inventory sync.
- An on-demand `/scan/vd` request answered a permanent "no scanner on this node" (log + metric) instead of a transient one, even though the scanner would be ready moments later.

Closes #39346.

## Proposed Changes

- Added `hasStarted()`/`isEnabled()`/`startFailed()` to `VulnerabilityScannerFacade`, read *before* `isInitialized()`, so "never going to run here" (disabled), "running, still starting" (enabled but not yet initialized), and "finished starting, unsuccessfully" resolve to different outcomes instead of collapsing into the same `false`.
  - `m_enabled` is computed and stored **before** `m_startInvoked` is published (`memory_order_release`, paired with an `memory_order_acquire` load in `hasStarted()`), so a concurrent reader can never observe "started" alongside a still-default `isEnabled()`. An earlier version of this PR set `m_startInvoked` first and only relaxed-stored `m_enabled` afterward — that ordering bug was caught in review and is what this version fixes.
- `wm_inventory_sync_server_main()` resolves this node's own `vd_configured_enabled` from the vulnerability-detection wmodule's parsed config before that scanner has started, on its own thread. That read is safe because `data->vulnerability_detection` is write-once from `wm_config()` time on: `wm_vulnerability_scanner_main()` applies its deprecated-config defaults (`enabled`, `feed-update-interval`) to a private duplicate via a shared `wm_vulnerability_scanner_apply_config_defaults()` helper, never to the shared object. `wm_vulnerability_scanner_dump()` uses the same helper on its own duplicate, so the module's `-D`/dump-config output still shows the effective (defaulted) config — an earlier version of this fix skipped that and broke the dump output, which is why it had been reverted; this version fixes both at once.
- Extracted the VD scan-lane's admission decision into two pure functions — `vdWillRunHere(started, enabled, configuredEnabled)` and `feedGateOpen(willRunHere, startFailed, initialized, feedReady)` (`vdScannerFactory.hpp` / `vdScannerAdapter.cpp`) — so the startup-window gate can be unit-tested without a live facade, and so no single function takes more than 4 positional same-typed booleans.
- Fixed `docs/ref/modules/remoted/agent-api.yaml`'s `/scan/vd` `503` contract (unrelated pre-existing issue found while touching this file): the schema declared a `code` field `errorJson()` never actually emits, and used a closed enum for an error string that is really a pass-through of whatever VD (or its own transport) reports. Replaced it with a `ScanVdError` schema that matches what the endpoint sends, including the real transport-level string (`"Server is shutting down"`, not the placeholder token `shutting_down` an earlier version of this schema used).

### Deliberately not changed in this PR

Found during review, evaluated, and left as documented follow-ups rather than fixed here — none of them are the startup-race ambiguity this PR closes, and each would add more machinery than the narrow case warrants:

- **Permanent 503 if the scanner never starts at all** (`libvulnerability_scanner.so` fails to load/export its start symbol): `vdWillRunHere()` carries a doc comment with the fix sketch. Verified live in this round of testing that this is actually a much bigger failure than originally scoped — a genuinely missing `.so` prevents `wazuh-manager-modulesd` (and `wazuh-manager-authd`) from starting at all, since `vulnerability_scanner` is a hard link-time dependency of both, not just a `dlopen()`-time one. That's a loud, immediately-visible failure, not the narrow silent-retry case the original analysis assumed — which further lowers the priority of a dedicated fix here.
- **Duplicate `enabled`-parsing between `wm_inventory_sync_server.c` (`w_parse_bool`) and `vulnerabilityScannerFacade.cpp` (`Utils::parseStrToBool`)**: cross-reference comments added in both files. Verified live that the manager's own JSON-schema validation (`wazuh-manager-conf validate`) rejects any `<enabled>` value besides `yes`/`no` outright — so the "unrecognized string" case these two parsers disagree on isn't reachable through the modern `wazuh-manager.conf` path at all, only through the deprecated legacy config format. Lower risk than originally scoped.
- **Raw struct read in `wm_inventory_sync_server.c` instead of the `context->query()` pattern**: evaluated and rejected — `query()` is pure routing indirection with no synchronization of its own (confirmed by reading its only current user, `wm_agent_info.c`), so adopting it would not have made the read safe; the write-once fix above is what does that.

### Review round 2 (@jotacarma90, 2026-09-21)

15 findings (F1–F15), verdict **COMMENT** — none blocking. 11 applied, 2 deferred as pre-existing follow-ups, 2 left out:

- **Applied**: F1 (`hasStarted()`/`isEnabled()` were sibling call arguments, so their evaluation order was unspecified and could void the acquire/release pairing — sequenced explicitly now), F2 (logging init moved ahead of the throwable `vdEnabled` lambda, so a throw there logs instead of hitting a null `GLOBAL_LOG_FUNCTION`), F3 (documented why `checkScanReadiness()`'s `ScannerNotReady`/`m_initialized` branch is unreachable from either current caller), F4/F12 (added the one test that actually pins that branch, and gave `HasStarted_FalseUntilStartIsInvoked` save/restore), F5 (the fake VD scanner factory now records the `vdConfiguredEnabled` it receives, plus a new E2E test proving the `config.vd_configured_enabled` → factory chain — closes a gap where reverting that wiring stayed test-green), F6/F7 (widened "CVE feed still downloading" wording across `docs/ref/modules/inventory-sync-server/*` and the `syncEndpoint.cpp` log/`IVdScanner.hpp` doc comment to also cover "scanner enabled but still starting up"), F8 (the `ScanVdError.error` schema used an invalid OpenAPI-3.0.3 schema-level `examples:` array; replaced with prose), F11 (the `scanAgent()` gate comment now names the `startFailed` case it omitted), F13 (dropped a byte-identical duplicate test).
- **Deferred as pre-existing, follow-up issue**: F9 (`/scan/vd`'s `400` response still uses the mismatched `ErrorResponse` schema — same class of bug this PR fixes for `503`, but not introduced here), F15 (`m_initialized` publishes `memory_order_relaxed` against two non-atomic `shared_ptr`s it's meant to gate — a real pre-existing data race, `d0e577a8`, outside this PR's diff).
- **Not applied**: F10 (pure style — duplicate-predicate cleanup, lowest value of the round), F14 (reviewer disputes this PR's own justification for a test-header include; provable only by removing it and building, which needs its own verification pass rather than a blind apply of either side).

Verified in a fresh VM copy (never on the host, per repo convention), rebuilt clean from the affected CMake targets:
- `inventory_sync_server_utest`: **265/265 passed** (dropped the F13 duplicate, added the F5 wiring test — net count unchanged from before this round).
- `vulnerability_scanner_unit_tests`: **534/534 passed** (533 baseline + the new F4 test; the `VulnerabilityScannerFacadeTest.*` subset alone: 42/42).

## Testing

Reviewed with 5 parallel finder passes plus a full independent read of every touched function and its callers; 8 findings surfaced, 1 (a CHANGELOG.md requirement) was a false positive — that requirement is gated by a GitHub project field, not a blanket CI rule. Of the remaining 7, verified each directly against the code (not just the review's summary) before deciding whether to fix or defer; see "Deliberately not changed" above for the 3 deferred ones.

**Unit tests**, full `UNIT_TEST=ON` non-engine tree, rebuilt clean from a fresh `deps` fetch:
- `inventory_sync_server_utest`: **265/265 passed**.
- `vulnerability_scanner_unit_tests`: **533/533 passed** (full suite); the `VulnerabilityScannerFacadeTest.*` subset also passes 41/41 but the test binary aborts at process-exit time afterward (RocksDB static-destruction-order issue in the singleton's teardown). Confirmed via direct A/B rebuild against the pre-this-PR parent commit that this crash is **pre-existing**, not introduced here — same crash, same stack trace, on code before any of these changes.
- `clang-format` (v22, the CI-authoritative version): checked every touched file; found and fixed one real formatting nit in `vdScannerAdapter.cpp` (my own `feedGateOpen(...)` call's continuation indent).

**Live, in a real installed manager + agent** (not just unit tests): generated the e2e cert bundle, enrolled a real agent against a real `wazuh-indexer`, and exercised every state the fix touches:
- **The exact write-once regression scenario** (`<vulnerability-detection>` with `<enabled>` omitted and a numeric `<feed-update-interval>`, both schema-legal): no crash across 3 separate restart cycles; confirmed via the module's live `getconfig wmodules` control-socket query that `wm_vulnerability_scanner_dump()`'s output shows the correctly-defaulted values (`"enabled":true`, `"feed-update-interval":"7200"` as a string) even though the raw config omitted/mistyped them — this is the specific case that broke on the previous attempt.
- **All three readiness-gate states**, driven by real agent (syscollector) traffic and the on-demand scan endpoint: enabled+starting → `503 + Retry-After` (`vd.retry_after.total` incremented); enabled+ready → scan succeeds (`vd.scans.ok` incremented); **disabled → legitimate skip** (`vd.scans.skipped` incremented, on-demand path answers a distinct "no vulnerability scanner on this node" and dead-letters rather than retrying forever).
- **50 concurrent on-demand scan requests** for the same agent: 49 correctly rejected `409 scan_in_progress` (per-agent exclusion), 1 accepted and completed; no crash, no corruption.
- **5x rapid manager restart cycling**: 7/7 daemons up every time, no crashes, no new errors.
- Confirmed no regression in agent enrollment/connectivity, real indexer bulk-indexing, or general (non-VD) inventory sync across all of the above.
- One pre-existing, unrelated issue observed in the environment and **not** touched: an engine/FIM-decoder schema error (`Field 'file.diff' is not defined in WCS schema`) — nothing to do with VD scanner, inventory sync, or anything in this PR's diff.

### Results and Evidence

See "Testing" above — internal gating/concurrency-safety logic, verified by unit tests plus live testing against a real installed manager, agent, and indexer rather than a synthetic environment.

  - [x] Linux — built clean (`TARGET=manager` release build + full `UNIT_TEST=ON` tree), see Testing above.
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows: N/A (manager-only change)
- Memory tests for macOS: N/A (manager-only change)

- Engine _(Wazuh v5.x and above)_: N/A (this module is outside the engine)

- Wazuh server API/Framework: N/A (no API/framework surface touched)

### Artifacts Affected

N/A — no new executables, default configuration files, or packages.

### Configuration Changes

N/A — no new configuration parameters or default-value changes; `wm_inventory_sync_server.c` only reads the existing `<vulnerability-detection>` config that was already parsed elsewhere.

### Tests Introduced

- `vdScannerAdapter_test.cpp`: unit coverage split across `vdWillRunHere()` (4 tests: the started/enabled/configuredEnabled decision in isolation) and `feedGateOpen()` (5 tests: startFailed/initialized/feedReady, given an already-resolved `willRunHere` — a 6th, byte-identical to `EnabledStillStarting_Defers`, was dropped in review round 2).
- `vulnerabilityScannerFacade_test.cpp`: coverage for `hasStarted()`/`isEnabled()`/`startFailed()` and their interaction with the existing readiness checks, plus test-isolation fixes (save/restore around tests that share the facade singleton, so earlier tests can't leak state into `TriggerAgentScan_*` tests that assume defaults). Round 2 added `TriggerAgentScan_StartedButNotInitialized_ReturnsScannerNotReady`, the one state that exercises `checkScanReadiness()`'s `ScannerNotReady` branch.
- `statefulEndpointE2E_test.cpp` (round 2): `VdConfiguredEnabledFromConfigReachesTheScannerFactory` pins `inventory_sync_server_config_t::vd_configured_enabled` reaching the scanner factory end to end.

## Review Checklist

- [x] Code changes reviewed


src/wazuh_modules/inventory_sync_server/src/vd/vdScannerAdapter.cpp
```diff
@@ -37,12 +54,33 @@ namespace invsync::vd
...
+            // validating) with a retryable 503, instead of routing a session that merely arrived
+            // early through Skipped as if this node would never run a scanner.
             auto& scanner = VulnerabilityScannerFacade::instance();
-            // An uninitialized scanner is not a "feed not ready" condition: those sessions skip
-            // the scan and index (D22's legitimate-skip row), so they must pass this gate.
-            return !scanner.isInitialized() || scanner.isFeedReady();
+            return feedGateOpen(vdWillRunHere(scanner.hasStarted(), scanner.isEnabled(), m_configuredEnabled),
+                                scanner.startFailed(),
+                                scanner.isInitialized(),
+                                scanner.isFeedReady());
```
jotacarma90: **[S3 · non-blocking · introduced]** `hasStarted()` and `isEnabled()` are sibling arguments of the same call, and argument evaluation order is unspecified in C++17, so the relaxed load of `m_enabled` may execute before the acquire load of `m_startInvoked` — which voids the guarantee the third commit was written to establish and documents at `vulnerabilityScannerFacade.hpp:86-89`. On a node with `<enabled>false</enabled>`, a reader can then observe `started=true` together with `enabled=true` (the default, `vulnerabilityScannerFacade.hpp:314`), close the gate and answer a spurious `503` on a node where VD will never run. The window is sub-microsecond and the `503` is retryable, hence S3, but the fix is one line.

```suggestion
            // Sequenced before isEnabled(): as sibling call arguments their evaluation order is
            // unspecified, and only running the acquire load first orders the other one.
            const bool started = scanner.hasStarted();
            return feedGateOpen(vdWillRunHere(started, scanner.isEnabled(), m_configuredEnabled),
                                scanner.startFailed(),
                                scanner.isInitialized(),
                                scanner.isFeedReady());
```

src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScannerFacade.cpp
```diff
@@ -113,6 +113,23 @@ void VulnerabilityScannerFacade::start(
...
+        // "start() was never invoked at all" (a test harness that skips it). Released here and
+        // acquired by hasStarted(): a relaxed store/load pair on two independent atomics gives no
+        // cross-thread ordering guarantee, so a plain reorder alone would not be enough to make
+        // "hasStarted() true" reliably imply "isEnabled() already reflects real config" on a
+        // non-x86 build.
+        m_startInvoked.store(true, std::memory_order_release);
+
+        // Initialize logging
+        Log::assignLogFunction(logFunction);
+        Log::setModuleLogFn(LogFn {WM_VULNSCAN_LOGTAG});
```
jotacarma90: **[S3 · non-blocking · introduced]** Moving the `vdEnabled` lambda above this means a throw from it is caught at line 496 and logged through a `GLOBAL_LOG_FUNCTION` that is still null — `loggerHelper.h:125` assigns it only here, and every emitter is guarded by `if (GLOBAL_LOG_FUNCTION)`, so the operator gets no line at all where before they got `VulnerabilityScannerFacade::start: Invalid input.`. It is latent rather than live today (the only throw is `Utils::parseStrToBool`, and the manager schema types `vulnerability-detection.enabled` as a boolean, so no unrecognised string reaches it), but logging init has no dependency on the lambda and nothing is gained by deferring it.

```cpp
    try
    {
        // Initialize logging
        Log::assignLogFunction(logFunction);
        Log::setModuleLogFn(LogFn {WM_VULNSCAN_LOGTAG});

        // ... comment, vdEnabled lambda, m_enabled.store, m_startInvoked.store unchanged;
        // only the current lines 126-128 go away.
```

src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScannerFacade.cpp
```diff
@@ -623,11 +642,16 @@ void VulnerabilityScannerFacade::rescanDisconnectedAgents()
...
+    if (!hasStarted() || !isEnabled() || startFailed())
     {
         return ScanTriggerResult::NotInitialized;
     }
 
+    if (!m_initialized.load(std::memory_order_relaxed))
+    {
+        return ScanTriggerResult::ScannerNotReady;
+    }
+
```
jotacarma90: **[S3 · non-blocking · introduced]** This branch cannot be reached in production, so `checkScanReadiness()` returns exactly what it did before the PR in every observable state. `vulnerabilityScanner.cpp` calls `start()` at line 92 and only creates `g_vdServer` at 108 and starts it at 373, so by the time `/vulnerability-detector/scan` can be served exactly one of `{m_initialized, m_startFailed, !m_enabled}` holds; the other caller, `triggerAgentScan()`, is only reached from `vdScannerAdapter::scanAgent()` past its `!isInitialized()` guard at line 140. Worth either dropping the branch or saying in the PR body that this half is a no-op reorder — the behaviour change the issue asks for comes entirely from the lane gate.

src/wazuh_modules/vulnerability_scanner/tests/unit/vulnerabilityScannerFacade_test.cpp
```diff
@@ -674,44 +712,64 @@ TEST_F(VulnerabilityScannerFacadeTest, GetReadinessStatus_LastSuccessfulUpdateCa
 TEST_F(VulnerabilityScannerFacadeTest, TriggerAgentScan_NotInitialized_ReturnsNotInitialized)
 {
     auto& facade = VulnerabilityScannerFacade::instance();
-    const bool originalInitialized = facade.m_initialized.load(std::memory_order_relaxed);
-    facade.m_initialized.store(false, std::memory_order_relaxed);
+    const bool originalStarted = facade.m_startInvoked.load(std::memory_order_relaxed);
+    // checkScanReadiness() now checks hasStarted() first; that's the condition this test names.
+    facade.m_startInvoked.store(false, std::memory_order_relaxed);
```
jotacarma90: **[S3 · non-blocking · introduced]** Swapping `m_initialized.store(false)` for `m_startInvoked.store(false)` means this test no longer exercises the input its name announces, and no other test drives `started=true, enabled=true, startFailed=false, initialized=false` — the one state whose result the PR changes. A regression that reverted line 650 of `vulnerabilityScannerFacade.cpp` would leave the suite green. `TriggerAgentScan_ScannerNotReady` does reach `ScannerNotReady`, but through the orchestrator/indexer branch with `m_initialized=true`.

src/wazuh_modules/inventory_sync_server/test/unit/testIndexerConnectorFakes.hpp
```diff
@@ -500,7 +505,7 @@ namespace invsync::test
     /// Installs a FakeVdScanner (sharing @p events) as the module's scan lane seam.
     inline void installFakeVdScanner(const std::shared_ptr<ConnectorEvents>& events)
     {
-        invsync::test_hooks::setVdScannerFactoryForTests([events]() -> std::shared_ptr<invsync::vd::IVdScanner>
+        invsync::test_hooks::setVdScannerFactoryForTests([events](bool) -> std::shared_ptr<invsync::vd::IVdScanner>
                                                          { return std::make_shared<FakeVdScanner>(events); });
```
jotacarma90: **[S3 · non-blocking · introduced]** The fake factory takes the new `bool` and discards it, and every `inventory_sync_server_config_t` in the tests and the testtool is value-initialised to `{}`, so nothing observes the chain `wm_inventory_sync_server.c` → struct → `m_config` → Phase-A snapshot (`inventorySyncServerFacade.hpp:935`) → `scannerFactory(vdConfiguredEnabled)` (line 1108) → `VdScannerAdapter::m_configuredEnabled`. Deleting the snapshot line or passing the wrong variable compiles and keeps all 265 cases green while the gate silently reverts to pre-PR behaviour on every node.

```cpp
// in struct ConnectorEvents, next to m_syncAvailable:
std::atomic<bool> m_vdConfiguredEnabled {false}; ///< Last value the scanner factory received.

// and here:
invsync::test_hooks::setVdScannerFactoryForTests(
    [events](bool vdConfiguredEnabled) -> std::shared_ptr<invsync::vd::IVdScanner>
    {
        events->m_vdConfiguredEnabled = vdConfiguredEnabled;
        return std::make_shared<FakeVdScanner>(events);
    });
```

src/wazuh_modules/inventory_sync_server/README.md
```diff
@@ -475,9 +475,12 @@ flush, no group commit in this lane — and answer `200`.
...
 (this node's current VD feed offset, backing the `feed_offset` check above), and
-`scan() -> Ok | Skipped`. `feedReady()` is `!isInitialized() || isFeedReady()` — a DISABLED
-scanner passes the gate and `scan()` reports `Skipped`, so inventory keeps flowing with VD off
-(the only legitimate skip; it still indexes and answers `200`). The production adapter lives in
+`scan() -> Ok | Skipped`. `feedReady()` is `feedGateOpen()`: a scanner that will never run here
+(disabled — by this node's own config or by the running scanner's — or one that started and
+failed) passes the gate and `scan()` reports `Skipped`, so inventory keeps flowing with VD off
+(the only legitimate skip; it still indexes and answers `200`). A scanner that WILL run here but
+hasn't finished starting keeps the gate closed until `isInitialized() && isFeedReady()`,
+deferring the whole startup window with a retryable `503` instead. The production adapter lives in
```
jotacarma90: **[S3 · non-blocking · introduced]** This developer README is updated but the operator contract in `docs/ref/modules/inventory-sync-server/` is not, and `.claude/docs/module-standards.md` makes `docs/ref` the page a PR must keep in sync when behaviour changes. `architecture.md:200-201` still reads "Scan legitimately skipped (scanner disabled) → index + 200" and "VD session on a node running no scanner → admitted with no version check, then index + 200", and the "the CVE feed is still downloading" wording in `README.md:22`/`:100`, `api-reference.md:18`, `metrics.md:146` and `test-tools.md:125` now describes a narrower condition than the code produces.

docs/ref/modules/remoted/agent-api.yaml
```diff
@@ -1577,6 +1579,24 @@ components:
...
+        transport-level ones its uds_http_server socket can produce on its own (e.g. "Server is
+        shutting down", "Data session limit reached") when VD itself never answers.
+      properties:
+        error:
+          type: string
+          description: Machine-readable cause. The values below are the ones VD is known to produce.
+          example: scan_queue_full
+          examples: [scan_queue_full, indexer_unavailable, feed_not_ready, scanner_not_ready,
+                     task_create_failed, vd_not_initialized, vd_unreachable, vd_error,
+                     "Server is shutting down"]
```
jotacarma90: **[S3 · non-blocking · introduced]** `examples:` as an array inside a Schema Object is a 3.1 / JSON-Schema-2020-12 keyword; this file declares `openapi: 3.0.3` on line 1, where a Schema Object only defines `example`. Spectral's `oas3-schema` rejects it and every 3.0 renderer ignores it, so the list the description points at ("The values below are the ones VD is known to produce") never reaches a reader. It is the only schema-level `examples` in the file — 1339/1355/1366/1379 are all Media Type Object level, where the map form is valid.

```suggestion
        error:
          type: string
          description: |
            Machine-readable cause. The values VD is known to produce: `scan_queue_full`,
            `indexer_unavailable`, `feed_not_ready`, `scanner_not_ready`, `task_create_failed`,
            `vd_not_initialized`, `vd_unreachable`, `vd_error`, `Server is shutting down`.
          example: scan_queue_full
```

src/wazuh_modules/src/wm_inventory_sync_server.c
```diff
@@ -424,6 +426,43 @@ void* wm_inventory_sync_server_main(wm_inventory_sync_server_t* data)
...
+                    config.vd_configured_enabled = !vd_enabled || cJSON_IsTrue(vd_enabled);
+                }
+                else if (cJSON_IsString(vd_enabled))
+                {
+                    config.vd_configured_enabled = w_parse_bool(vd_enabled->valuestring) == 1;
+                }
+                else
+                {
+                    config.vd_configured_enabled = true;
+                }
```
jotacarma90: **[S4 · non-blocking · introduced]** The guard `if (!vd_enabled || cJSON_IsBool(vd_enabled))` re-tests `!vd_enabled` inside its own body, and the final `else` duplicates a default that `cJSON_IsTrue(NULL) == false` already gives. Three type-ordered branches cover NULL, bool, string, number and null identically without repeating a predicate.

```suggestion
                if (cJSON_IsString(vd_enabled))
                {
                    config.vd_configured_enabled = w_parse_bool(vd_enabled->valuestring) == 1;
                }
                else if (cJSON_IsBool(vd_enabled))
                {
                    config.vd_configured_enabled = cJSON_IsTrue(vd_enabled);
                }
                else
                {
                    config.vd_configured_enabled = true;
                }
```

src/wazuh_modules/inventory_sync_server/src/vd/vdScannerAdapter.cpp
```diff
@@ -101,9 +139,11 @@ namespace invsync::vd
...
             if (!scanner.isInitialized())
             {
-                // Skipped, not NotReady: on a node with vulnerability detection disabled there is
-                // nothing to come back for, and reporting it as transient would make the caller
-                // retry a scan that can never run on this manager.
+                // Reached only for "never going to run here" (disabled, or a harness that never
+                // started it): the lane's feedReady() re-check, one level up, already deferred
+                // the enabled-but-starting case with a retryable 503 before calling scanAgent()
+                // at all. Skipped, not NotReady: this gate is what the QA suite, the operator
+                // WARN and vd.scans.skipped hang off.
```
jotacarma90: **[S4 · non-blocking · introduced]** "Reached only for 'never going to run here'" leaves out the third state this PR adds: `feedGateOpen(willRunHere=true, startFailed=true, ...)` returns true, so a scanner whose `start()` threw also reaches this guard and returns `Skipped`. That is what makes `vdScanLane.cpp:380` tell the operator "this node runs no vulnerability scanner" for a node that has one and failed to start it. The WARN itself predates the PR; the comment describing it does not.

src/wazuh_modules/vulnerability_scanner/tests/unit/vulnerabilityScannerFacade_test.cpp
```diff
@@ -72,6 +72,22 @@ TEST_F(VulnerabilityScannerFacadeTest, IsInitialized_BeforeStart)
...
+// testtool's --no-vd) is exactly the state this accessor exists to distinguish.
+TEST_F(VulnerabilityScannerFacadeTest, HasStarted_FalseUntilStartIsInvoked)
+{
+    auto& facade = VulnerabilityScannerFacade::instance();
+    facade.m_startInvoked.store(false, std::memory_order_relaxed);
+
+    EXPECT_FALSE(facade.hasStarted());
+
+    facade.m_startInvoked.store(true, std::memory_order_relaxed);
+    EXPECT_TRUE(facade.hasStarted());
```
jotacarma90: **[S4 · non-blocking · introduced]** This is the only state mutation in the file without save/restore, in a PR that adds save/restore to five other tests for exactly this reason. It leaves `m_startInvoked = true` on the shared singleton, so under `--gtest_shuffle` or a filtered run whatever comes next starts with `hasStarted() == true` — the first condition `checkScanReadiness()` looks at.

```suggestion
    auto& facade = VulnerabilityScannerFacade::instance();
    const bool originalStarted = facade.m_startInvoked.load(std::memory_order_relaxed);
    facade.m_startInvoked.store(false, std::memory_order_relaxed);

    EXPECT_FALSE(facade.hasStarted());

    facade.m_startInvoked.store(true, std::memory_order_relaxed);
    EXPECT_TRUE(facade.hasStarted());

    facade.m_startInvoked.store(originalStarted, std::memory_order_relaxed);
```

src/wazuh_modules/inventory_sync_server/test/unit/vdScannerAdapter_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+                             /*feedReady*/ false));
+}
+
+TEST(FeedGateOpen, EnabledStillStarting_Defers)
+{
+    EXPECT_FALSE(feedGateOpen(/*willRunHere*/ true,
+                              /*startFailed*/ false,
+                              /*initialized*/ false,
+                              /*feedReady*/ false));
+}
```
jotacarma90: **[S4 · non-blocking · introduced]** Byte-identical to `FeedGateOpen.NeverStartedButConfiguredEnabled_Defers`: same four arguments, same expectation. The distinction the two names promise lives entirely in `vdWillRunHere()`, which has already collapsed it to one bool before `feedGateOpen()` is called, so the second case adds no coverage to the 10 the PR body cites. Either drop one or rename the survivor to what it actually pins (`WillRunHereAndNotReady_Defers`).

src/wazuh_modules/inventory_sync_server/test/unit/testIndexerConnectorFakes.hpp
```diff
@@ -22,6 +22,11 @@
 #include "vd/IVdScanner.hpp"
 #include "vd/vdScannerFactory.hpp"
 
+// Not used in this header itself: several sibling *_test.cpp files include this header for its
+// fakes and then reach for wazuh::uds_http types (HttpRequest/HttpResponse/...) unqualified,
+// relying on it transitively -- removing it breaks their build, not this file's.
+#include <uds_http_server/IUdsHttpServer.hpp>
```
jotacarma90: **[S4 · non-blocking · introduced]** The include was not there before this PR, the base branch compiles, and nothing in this PR changes another include — so the eight sibling tests that use `wazuh::uds_http` without including it get it through some other transitive path, and "removing it breaks their build" is not true. The comment will stop the next person pruning includes here. If the dependency is worth making explicit, the place is each `*_test.cpp` that uses the types, and it is a separate change from this one.

src/wazuh_modules/inventory_sync_server/src/vd/vdScannerAdapter.cpp
jotacarma90: **[S3 · non-blocking · introduced]** `hasStarted()` and `isEnabled()` are sibling arguments of the same call, and argument evaluation order is unspecified in C++17, so the relaxed load of `m_enabled` may execute before the acquire load of `m_startInvoked` — which voids the guarantee the third commit was written to establish and documents at `vulnerabilityScannerFacade.hpp:86-89`. On a node with `<enabled>false</enabled>`, a reader can then observe `started=true` together with `enabled=true` (the default, `vulnerabilityScannerFacade.hpp:314`), close the gate and answer a spurious `503` on a node where VD will never run. The window is sub-microsecond and the `503` is retryable, hence S3, but the fix is one line.

```suggestion
            // Sequenced before isEnabled(): as sibling call arguments their evaluation order is
            // unspecified, and only running the acquire load first orders the other one.
            const bool started = scanner.hasStarted();
            return feedGateOpen(vdWillRunHere(started, scanner.isEnabled(), m_configuredEnabled),
                                scanner.startFailed(),
                                scanner.isInitialized(),
                                scanner.isFeedReady());
```

src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScannerFacade.cpp
jotacarma90: **[S3 · non-blocking · introduced]** Moving the `vdEnabled` lambda above this means a throw from it is caught at line 496 and logged through a `GLOBAL_LOG_FUNCTION` that is still null — `loggerHelper.h:125` assigns it only here, and every emitter is guarded by `if (GLOBAL_LOG_FUNCTION)`, so the operator gets no line at all where before they got `VulnerabilityScannerFacade::start: Invalid input.`. It is latent rather than live today (the only throw is `Utils::parseStrToBool`, and the manager schema types `vulnerability-detection.enabled` as a boolean, so no unrecognised string reaches it), but logging init has no dependency on the lambda and nothing is gained by deferring it.

```cpp
    try
    {
        // Initialize logging
        Log::assignLogFunction(logFunction);
        Log::setModuleLogFn(LogFn {WM_VULNSCAN_LOGTAG});

        // ... comment, vdEnabled lambda, m_enabled.store, m_startInvoked.store unchanged;
        // only the current lines 126-128 go away.
```

src/wazuh_modules/vulnerability_scanner/src/vulnerabilityScannerFacade.cpp
jotacarma90, line 659: **[S3 · non-blocking · introduced]** This branch cannot be reached in production, so `checkScanReadiness()` returns exactly what it did before the PR in every observable state. `vulnerabilityScanner.cpp` calls `start()` at line 92 and only creates `g_vdServer` at 108 and starts it at 373, so by the time `/vulnerability-detector/scan` can be served exactly one of `{m_initialized, m_startFailed, !m_enabled}` holds; the other caller, `triggerAgentScan()`, is only reached from `vdScannerAdapter::scanAgent()` past its `!isInitialized()` guard at line 140. Worth either dropping the branch or saying in the PR body that this half is a no-op reorder — the behaviour change the issue asks for comes entirely from the lane gate.

src/wazuh_modules/vulnerability_scanner/tests/unit/vulnerabilityScannerFacade_test.cpp
jotacarma90, line 720: **[S3 · non-blocking · introduced]** Swapping `m_initialized.store(false)` for `m_startInvoked.store(false)` means this test no longer exercises the input its name announces, and no other test drives `started=true, enabled=true, startFailed=false, initialized=false` — the one state whose result the PR changes. A regression that reverted line 650 of `vulnerabilityScannerFacade.cpp` would leave the suite green. `TriggerAgentScan_ScannerNotReady` does reach `ScannerNotReady`, but through the orchestrator/indexer branch with `m_initialized=true`.

src/wazuh_modules/inventory_sync_server/test/unit/testIndexerConnectorFakes.hpp
jotacarma90: **[S3 · non-blocking · introduced]** The fake factory takes the new `bool` and discards it, and every `inventory_sync_server_config_t` in the tests and the testtool is value-initialised to `{}`, so nothing observes the chain `wm_inventory_sync_server.c` → struct → `m_config` → Phase-A snapshot (`inventorySyncServerFacade.hpp:935`) → `scannerFactory(vdConfiguredEnabled)` (line 1108) → `VdScannerAdapter::m_configuredEnabled`. Deleting the snapshot line or passing the wrong variable compiles and keeps all 265 cases green while the gate silently reverts to pre-PR behaviour on every node.

```cpp
// in struct ConnectorEvents, next to m_syncAvailable:
std::atomic<bool> m_vdConfiguredEnabled {false}; ///< Last value the scanner factory received.

// and here:
invsync::test_hooks::setVdScannerFactoryForTests(
    [events](bool vdConfiguredEnabled) -> std::shared_ptr<invsync::vd::IVdScanner>
    {
        events->m_vdConfiguredEnabled = vdConfiguredEnabled;
        return std::make_shared<FakeVdScanner>(events);
    });
```

src/wazuh_modules/inventory_sync_server/README.md
jotacarma90, line 483: **[S3 · non-blocking · introduced]** This developer README is updated but the operator contract in `docs/ref/modules/inventory-sync-server/` is not, and `docs/ref` is the operator-facing contract that has to stay in sync when behaviour changes. `architecture.md:200-201` still reads "Scan legitimately skipped (scanner disabled) → index + 200" and "VD session on a node running no scanner → admitted with no version check, then index + 200", and the "the CVE feed is still downloading" wording in `README.md:22`/`:100`, `api-reference.md:18`, `metrics.md:146` and `test-tools.md:125` now describes a narrower condition than the code produces.

docs/ref/modules/remoted/agent-api.yaml
jotacarma90: **[S3 · non-blocking · introduced]** `examples:` as an array inside a Schema Object is a 3.1 / JSON-Schema-2020-12 keyword; this file declares `openapi: 3.0.3` on line 1, where a Schema Object only defines `example`. Spectral's `oas3-schema` rejects it and every 3.0 renderer ignores it, so the list the description points at ("The values below are the ones VD is known to produce") never reaches a reader. It is the only schema-level `examples` in the file — 1339/1355/1366/1379 are all Media Type Object level, where the map form is valid.

```suggestion
        error:
          type: string
          description: |
            Machine-readable cause. The values VD is known to produce: `scan_queue_full`,
            `indexer_unavailable`, `feed_not_ready`, `scanner_not_ready`, `task_create_failed`,
            `vd_not_initialized`, `vd_unreachable`, `vd_error`, `Server is shutting down`.
          example: scan_queue_full
```

src/wazuh_modules/src/wm_inventory_sync_server.c
jotacarma90, line 463: **[S4 · non-blocking · introduced]** The guard `if (!vd_enabled || cJSON_IsBool(vd_enabled))` re-tests `!vd_enabled` inside its own body, and the final `else` duplicates a default that `cJSON_IsTrue(NULL) == false` already gives. Three type-ordered branches cover NULL, bool, string, number and null identically without repeating a predicate.

```suggestion
                if (cJSON_IsString(vd_enabled))
                {
                    config.vd_configured_enabled = w_parse_bool(vd_enabled->valuestring) == 1;
                }
                else if (cJSON_IsBool(vd_enabled))
                {
                    config.vd_configured_enabled = cJSON_IsTrue(vd_enabled);
                }
                else
                {
                    config.vd_configured_enabled = true;
                }
```

src/wazuh_modules/inventory_sync_server/src/vd/vdScannerAdapter.cpp
jotacarma90: **[S4 · non-blocking · introduced]** "Reached only for 'never going to run here'" leaves out the third state this PR adds: `feedGateOpen(willRunHere=true, startFailed=true, ...)` returns true, so a scanner whose `start()` threw also reaches this guard and returns `Skipped`. That is what makes `vdScanLane.cpp:380` tell the operator "this node runs no vulnerability scanner" for a node that has one and failed to start it. The WARN itself predates the PR; the comment describing it does not.

src/wazuh_modules/vulnerability_scanner/tests/unit/vulnerabilityScannerFacade_test.cpp
jotacarma90, line 89: **[S4 · non-blocking · introduced]** This is the only state mutation in the file without save/restore, in a PR that adds save/restore to five other tests for exactly this reason. It leaves `m_startInvoked = true` on the shared singleton, so under `--gtest_shuffle` or a filtered run whatever comes next starts with `hasStarted() == true` — the first condition `checkScanReadiness()` looks at.

```suggestion
    auto& facade = VulnerabilityScannerFacade::instance();
    const bool originalStarted = facade.m_startInvoked.load(std::memory_order_relaxed);
    facade.m_startInvoked.store(false, std::memory_order_relaxed);

    EXPECT_FALSE(facade.hasStarted());

    facade.m_startInvoked.store(true, std::memory_order_relaxed);
    EXPECT_TRUE(facade.hasStarted());

    facade.m_startInvoked.store(originalStarted, std::memory_order_relaxed);
```

src/wazuh_modules/inventory_sync_server/test/unit/vdScannerAdapter_test.cpp
jotacarma90, line 68: **[S4 · non-blocking · introduced]** Byte-identical to `FeedGateOpen.NeverStartedButConfiguredEnabled_Defers`: same four arguments, same expectation. The distinction the two names promise lives entirely in `vdWillRunHere()`, which has already collapsed it to one bool before `feedGateOpen()` is called, so the second case adds no coverage to the 10 the PR body cites. Either drop one or rename the survivor to what it actually pins (`WillRunHereAndNotReady_Defers`).

src/wazuh_modules/inventory_sync_server/test/unit/testIndexerConnectorFakes.hpp
jotacarma90, line 28: **[S4 · non-blocking · introduced]** The include was not there before this PR, the base branch compiles, and nothing in this PR changes another include — so the eight sibling tests that use `wazuh::uds_http` without including it get it through some other transitive path, and "removing it breaks their build" is not true. The comment will stop the next person pruning includes here. If the dependency is worth making explicit, the place is each `*_test.cpp` that uses the types, and it is a separate change from this one.

jotacarma90 (commented):
## Review summary

Yes — both symptoms #39346 describes are actually fixed, by the lane gate: a VD session arriving during the scanner's startup window now gets `503` + `Retry-After` instead of being indexed unscanned, and the on-demand path (the durable Task Manager row the dispatcher forwards to `/_internal/vd/scan`) is deferred at `vdScanLane.cpp:336` before it can reach `scanAgent()`, so the misleading WARN and `vd.scans.skipped` no longer fire for that window.

**Verdict: COMMENT** — nothing blocking; the two worth acting on are F1 (the acquire/release guarantee this PR establishes is voided at its only production call site) and F3 (the `checkScanReadiness()` half of the change is unreachable in production).

| Severity | Introduced | Aggravated | Pre-existing | Blocking |
|---|---|---|---|---|
| S1 Critical | 0 | 0 | 0 | 0 |
| S2 High | 0 | 0 | 0 | 0 |
| S3 Medium | 7 | 1 | 2 | 0 |
| S4 Low | 5 | 0 | 0 | 0 |

### Non-blocking
- **F1** `vdScannerAdapter.cpp:83` — `hasStarted()` and `isEnabled()` are sibling call arguments, so their evaluation order is unspecified and the release/acquire pairing does not hold.
- **F2** `vulnerabilityScannerFacade.cpp:127` — logging is now initialised after the config lambda, so a throw from it is logged into a null `GLOBAL_LOG_FUNCTION`.
- **F3** `vulnerabilityScannerFacade.cpp:650` — the new `ScannerNotReady` branch cannot be reached from either caller.
- **F4** `vulnerabilityScannerFacade_test.cpp:715` — the repurposed test no longer covers `m_initialized == false`, and nothing else does.
- **F5** `testIndexerConnectorFakes.hpp:508` — the whole `vd_configured_enabled` chain is untested; dropping it keeps all 265 cases green.
- **F6** `inventory_sync_server/README.md:478` — the developer README was updated but the operator contract in `docs/ref/modules/inventory-sync-server/` was not.
- **F8** `agent-api.yaml:1592` — `examples:` as an array in a Schema Object is not valid OpenAPI 3.0.3.
- **F10** `wm_inventory_sync_server.c:452` — `!vd_enabled` is tested twice; three type-ordered branches say the same thing.
- **F11** `vdScannerAdapter.cpp:142` — the new comment omits the `startFailed` case, which does reach that guard.
- **F12** `vulnerabilityScannerFacade_test.cpp:82` — the only state mutation in the file without save/restore.
- **F13** `vdScannerAdapter_test.cpp:70` — byte-identical to `NeverStartedButConfiguredEnabled_Defers`.
- **F14** `testIndexerConnectorFakes.hpp:25` — include added out of scope, with a justification the base branch disproves.

<details>
<summary>Findings that could not be anchored inline (files outside the diff)</summary>

- **F7** `syncEndpoint.cpp:147` (aggravated) — the log line and `IVdScanner.hpp:72`'s doc comment still say "the CVE feed is not ready". Accurate before this PR, because "scanner still starting" took the skip path; now it reaches this branch by design, so on every manager restart the operator is told the feed is not ready when it may be perfectly on disk. Widening the log text and the doc comment is enough — the JSON body is a wire contract documented in `api-reference.md`.
- **F15** `vulnerabilityScannerFacade.cpp:492` (pre-existing, `d0e577a8`) — `m_initialized` is published with `memory_order_relaxed`, and the two non-atomic `shared_ptr`s it is meant to gate (`m_indexerConnector` at 412, `m_scanOrchestrator` at 459) are assigned before it. A reader that sees `isInitialized() == true` — also a relaxed load, `vulnerabilityScannerFacade.hpp:76` — has no happens-before edge with those two writes, so the `const auto orchestrator = m_scanOrchestrator;` copy at line 62 is a formal data race, not just a stale read. Not introduced here, and this PR's own `m_enabled`/`m_startInvoked` pair does get the ordering right — but that leaves `m_initialized` as the one readiness flag in the class without it, and `checkScanReadiness()`, which this PR rewrites, still reads it relaxed at line 650. Making the store `release` and the loads `acquire` is free on x86 and closes it; worth a follow-up issue rather than a change here.

- **F9** `agent-api.yaml:859` (pre-existing, `27cbfb4b`) — the `400` block of this same `/scan/vd` responses map still uses `ErrorResponse` (`required: [error, code]`, `additionalProperties: false`) while `scanVdEndpoint.cpp:55` emits only `{"error": ...}`. Exactly the defect this PR fixes a few lines below for the `503`; pointing it at the new `ScanVdError` and dropping `code: 400` from its six examples would finish the job. Not introduced by this PR — flagging it for a follow-up issue, not for this review.

</details>

### Verified, no change requested
- **The "known gap" in `vdWillRunHere()`'s doc comment.** Your reading is right and I confirmed it: `src/wazuh_modules/CMakeLists.txt:132-143` links `wazuh_modulesd_lib` against `vulnerability_scanner` `PUBLIC`, and `libinventory_sync_server.so` references `VulnerabilityScannerFacade::instance()` directly, so a missing `.so` either stops modulesd outright or makes `so_get_module_handle("inventory_sync_server")` fail — and that path is `mterror_exit`. What is left is a `.so` that resolves every C++ symbol but not the C `vulnerability_scanner_start`, which no real build produces. No fix needed.
- **The duplicated `enabled` vocabulary.** It is genuinely unavoidable across the `.so` boundary (the facade cannot use cJSON; inventory-sync cannot call into a library that may not have started), and the one input the two parsers disagree on is unreachable through the schema-validated config path. Moving the cJSON block into `wm_vulnerability_scanner.c` would not remove the second vocabulary. Agreed as deferred; F10 is only the `else` cleanup.
- **The write-once fix itself.** `wm_config()` runs at `wazuh_modules/src/main.c:240`, before any `CreateThreadJoinable` (163/182), and the only in-place mutation is now on a local duplicate. The cross-module read is safe as documented.

### Checklist
- CHANGELOG: **N/A** — `no-changelog` is right; the conflated gate has only ever existed in unreleased 5.0.0 (`741b8ae8`, 2026-08-05).
- Tests: **review** — 10 new tests, but all against the two extracted pure functions. Reverting the production wiring (F5) or the `checkScanReadiness()` branch (F4) leaves the suite green.
- PR template / DoD: **review** — evidence is real and detailed, but "Developer documentation reflects the changes" and "Meets requirements and/or definition of done" are unchecked, consistent with F6.
- Compatibility / backport: OK — base `5.0.0` is correct (5.x-only module), the config field is appended last, and no agent↔manager wire shape changes. `main` is covered by the weekly upward merge.

jotacarma90 (approved):

PR #39356, opened 2026-09-16
Report password status and hash algorithm in the macOS users inventory

### LINKED ISSUE #39159: macOS user inventory reports no password posture, with status absent and last_change, max_days and warning_days all zero, while Linux reports real values
## Description

The macOS agent inventories all 134 local users but reports nothing usable about their password posture. `user.password.status` is absent entirely, and the rest of the `user.password` sub-object is filled with zeros and a null rather than real values. Linux agents populate the same fields correctly for every user.

Field coverage across the fleet, counting documents where the field exists:

```
  user.password.status                    001:34 002:35 003:29 004:29        (005, 006 absent)
  user.password.last_change               001:34 002:35 003:29 004:29 005:9 006:134
  user.password.max_days_between_changes  001:34 002:35 003:29 004:29 005:9 006:134
  login.status                            001:34 002:35 003:29 004:29 005:9 006:134
  --- total users per agent ---
                                          001:34 002:35 003:29 004:29 005:9 006:134
```

The `exists` counts for macOS are misleading, which is the substance of this report. The sub-object is present but its contents are placeholders:

```
##### agent 006 (macOS)
   name=_accessoryupdater  id=278  pw={"warning_days_before_expiration": 0, "max_days_between_changes": 0, "last_change": 0, "hash_algorithm": null, ...
   name=_amavisd           id=83   pw={"warning_days_before_expiration": 0, "max_days_between_changes": 0, "last_change": 0, "hash_algorithm": null, ...
   name=_analyticsd        id=263  pw={"warning_days_before_expiration": 0, "max_days_between_changes": 0, "last_change": 0, "hash_algorithm": null, ...

##### agent 002 (Ubuntu)
   name=_apt      id=42   pw={"warning_days_before_expiration": 7,  "max_days_between_changes": 99999, "last_change": 1777766400, "hash_algo...
   name=_chrony   id=988  pw={"warning_days_before_expiration": -1, "max_days_between_changes": -1,    "last_change": 1777766400, "hash_algori...
   name=backup    id=34   pw={"warning_days_before_expiration": 7,  "max_days_between_changes": 99999, "last_change": 1777766400, "hash_algo...
```

Zeros are worse here than an absent field, for the same reason a zero package size is worse than no package size. `last_change: 0` is a valid epoch timestamp meaning 1970-01-01, so any view that renders it will show every macOS account as having had its password last changed 56 years ago. `max_days_between_changes: 0` can be read as a policy requiring a change every zero days. A consumer cannot distinguish "not collected" from "collected and genuinely zero" without knowing the platform.

The practical loss is password posture on macOS endpoints: accounts that are locked or disabled, accounts with no password set, and password ages are all unavailable, while the same questions are answerable on Linux. Note that macOS keeps this information in OpenDirectory rather than in a shadow file, so a fix needs a different source than the Linux path uses. I was not able to confirm on the host which OpenDirectory attributes would serve best, because the macOS host became unreachable before I could check, so that part is left to whoever picks this up rather than asserted here.

This is the same shape as three other defects found on this agent, and they may share a cause in how the macOS provider fills its structures: blank service state in wazuh/wazuh#39139, zero package size in wazuh/wazuh#39140, and discarded process arguments in wazuh/wazuh#39141. Each is filed separately because the affected collectors and fields differ, but it is worth triaging them together.

## Environment

- Agent under test: `demo-env-macos-agent`, id `006`, macOS 26.5.1 (build 25F80), arm64, Parallels VM on MacStadium host `macstadium_arm_4`, 134 local users.
- Comparison agents: `001` ubuntu-26-arm64 (34 users), `002` ubuntu-26-amd64 (35), `003` amazon-2023-amd64 (29), `004` amazon-2023-arm64 (29), `005` windows-11-amd64 (9).

## Steps to reproduce

1. Enroll a 5.0.0 macOS agent alongside at least one Linux agent and let a syscollector evaluation complete on both.
2. Compare which agents report a password status at all:

```bash
  "https://<indexer>:9200/wazuh-states-inventory-users/_search?size=0" \
  -d '{"query":{"exists":{"field":"user.password.status"}},
       "aggs":{"a":{"terms":{"field":"wazuh.agent.id","size":10}}}}'
```

3. Inspect the actual values rather than trusting the `exists` counts, for both a macOS and a Linux agent:

```bash
  "https://<indexer>:9200/wazuh-states-inventory-users/_search?size=3" \
       "_source":["user.name","user.id","user.password"]}'
```

4. Observe that every macOS user carries `last_change: 0`, `max_days_between_changes: 0`, `warning_days_before_expiration: 0` and `hash_algorithm: null`, while the Linux agent carries real epoch timestamps and policy values.
5. Cross-check on the macOS host that real values exist, for example `dscl . -read /Users/<name> accountPolicyData` or `pwpolicy -u <name> -getpolicy`.

## Expected

`user.password.status` is reported on macOS as it is on Linux, and the remaining `user.password` fields either carry real values or are omitted rather than being set to zero, so that "not collected" is distinguishable from a genuine zero.

## Actual

`user.password.status` is absent for all 134 macOS users and all 9 Windows users, while all four Linux agents report it for every user. On macOS the remaining password fields are present but zero-filled, so a consumer sees a last password change of 1970-01-01 and a maximum age of zero days for every account on the host.

## Not a duplicate of

`qa-known-issues.py check` scored one weak cache candidate and the live search returned nothing:

- `wazuh/wazuh-dashboard-plugins#9101` "Agent Configuration → Commands reports the section absent while the command wodle is running (Linux agent)" (0.36) concerns how the dashboard plugin renders an agent's *configuration* sections for a Linux agent. Different repository, different component, different data, and nothing to do with user password fields in the inventory.
- Live search for `warning_days last_change inventory password max_days reports` returned nothing open.

nbertoldo:
## Description

Closes: [#39159](https://github.com/wazuh/wazuh/issues/39159)

The macOS users collector never looked at any password source. `sysInfoMac.cpp` hardcoded seven of the eight `user_password_*` fields — `status`, `hash_algorithm` and `expiration_date` to `UNKNOWN_VALUE`, and the four aging fields to `0` — so every macOS account was reported with a maximum password age of zero days and no status at all. Only `user_password_last_change` had a real source.

Zeros are worse than absent values here: `max_days_between_changes: 0` reads as a policy requiring a change every zero days, and a consumer cannot tell it apart from a field that was never collected.

I investigated this on a macOS 26.5.1 (25F80, arm64) host with 134 local users, the same build as the reported environment, and the finding shaped the fix: **most of these fields have no source on macOS at all.** 131 of the 134 accounts have no `accountPolicyData` attribute, and stock macOS has no password aging policy whatsoever — the global `pwpolicy -getaccountpolicies` carries only a content rule, and per-user `pwpolicy -u <user> -getpolicy` returns empty. The aging fields therefore cannot be populated unless an MDM imposes an account policy.

Two fields do have a real source, and it is not `accountPolicyData`:

- **`user_password_status`** and **`user_password_hash_algorithm`** come from the OpenDirectory `AuthenticationAuthority` attribute. Accounts with a password carry a `;ShadowHash;HASHLIST:<SALTED-SHA512-PBKDF2,...>` authority; service accounts carry none.
- **The disabled state is not in the user record.** I verified this by creating a throwaway account, disabling it with `pwpolicy -disableuser` and diffing its OpenDirectory record against an enabled one: the records are identical, hash included. macOS tracks it as membership in the `com.apple.access_disabled` group. Because a disabled account keeps its hash, the group membership has to win over the status derived from the authority, or the account still reports as `active`.

## Proposed Changes

- `od_wrapper.mm` gains `genPasswordData()`, which derives the status and the hash algorithm from `AuthenticationAuthority`, and `genDisabledUsers()`, which reads the `com.apple.access_disabled` membership. The membership is resolved once per collection rather than once per user.
- `users_darwin.cpp` combines both in `collectPasswordData()`, applying the disabled state over the derived status.
- `sysInfoMac.cpp` reports the two recovered fields, and reports the five aging fields as not collected (`NOT_COLLECTED_VALUE`) instead of `0`.
- `user_password_last_change` is unchanged — it was already correct.

**Why a sentinel rather than omitting the fields.** Omitting the numeric fields does not work, and the reason is worth recording for review. `updateChanges` uses `DBSyncTxn`, and that path does not forward the input JSON: `insertNewRows` rebuilds the delta from the temporary table. `getTableData` then rereads each column by type without checking for NULL, and while `Column::value(const std::string&)` maps a NULL TEXT column to `""`, the integer readers call `sqlite3_column_int64` with no such guard, so a NULL INTEGER column comes back as `0`. An omitted numeric field would reappear as `0` in the delta and be stored as `0` by `wdb_users_insert`, whose guard is `>= 0`. The two TEXT fields need no sentinel: the provider returns `""`, `removeKeysWithEmptyValue` drops the key, and the manager stores NULL.

### Results and Evidence

Tested end to end with signed packages: a macOS 26.5.1 (25F80, arm64) agent enrolled against a 4.14.7 manager, with two purpose-made accounts — `wzactive` (password set) and `wzlocked` (password set, then disabled with `pwpolicy -disableuser`). The rows below are read from `sys_users` on the manager, so they are what a consumer actually sees. `wazuh-db` renders a NULL column as an absent key.

**Before** (branch `4.14.9`, build `9724fc6`) — all 136 accounts identical and uninformative:

```
{"user_name":"_amavisd","user_password_status":" ","user_password_hash_algorithm":" ","user_password_max_days_between_changes":0,"user_password_min_days_between_changes":0,"user_password_warning_days_before_expiration":0,"user_password_inactive_days":0}
{"user_name":"vagrant", "user_password_status":" ","user_password_hash_algorithm":" ","user_password_max_days_between_changes":0,...,"user_password_last_change":1689672655.46}
{"user_name":"wzactive","user_password_status":" ","user_password_hash_algorithm":" ","user_password_max_days_between_changes":0,...}
{"user_name":"wzlocked","user_password_status":" ","user_password_hash_algorithm":" ","user_password_max_days_between_changes":0,...}
```

```
total=136  status_is_space=136  status_is_null=0  max_days_is_zero=136  max_days_is_null=0
```

The disabled account is indistinguishable from the active one, and `status` is a single space rather than NULL, so it is not even detectable with an `exists`-style query.

**After** (this branch, build `0a036fb`):

```
{"user_name":"_amavisd","user_password_status":"not_set"}
{"user_name":"nobody",  "user_password_status":"not_set"}
{"user_name":"root",    "user_password_status":"not_set"}
{"user_name":"vagrant", "user_password_status":"active","user_password_hash_algorithm":"SALTED-SHA512-PBKDF2","user_password_last_change":1689672655.46}
{"user_name":"wzactive","user_password_status":"active","user_password_hash_algorithm":"SALTED-SHA512-PBKDF2","user_password_last_change":1789584549.30}
{"user_name":"wzlocked","user_password_status":"locked","user_password_hash_algorithm":"SALTED-SHA512-PBKDF2","user_password_last_change":1789584551.40}
```

```
total=136  status_is_null=0  status_is_space=0
max_days_is_null=136  min_days_is_null=136  warning_days_is_null=136  inactive_days_is_null=136  expiration_date_is_null=136

status:         not_set=133  active=2  locked=1
hash_algorithm: NULL=133     SALTED-SHA512-PBKDF2=3
```

Every account now carries a status, the five aging fields are NULL rather than zero across all 136 rows, the disabled account is distinguishable from the active one, and `last_change` is unchanged.

### Artifacts Affected

macOS agent only. `wazuh-modulesd` (syscollector) through `libsysinfo`. No manager-side changes, no schema changes, and no change to the delta format beyond which keys are present.

Windows carries the identical hardcoded block in `sysInfoWin.cpp` and the same symptom. It is deliberately out of scope here, but `users_utils_wrapper.cpp` already calls `NetUserGetInfo` at level 4 and discards `usri4_flags` (`UF_LOCKOUT`, `UF_ACCOUNTDISABLE`, `UF_PASSWD_NOTREQD`) and `usri4_password_age`, so the data is already being fetched.

### Configuration Changes

None.

### Documentation Updates

None required. No configuration option, log message or event field is added or renamed; the affected fields already exist in the schema and are documented.

### Tests Introduced

Unit tests in `test_users_darwin.cpp` covering an account with no password (`not_set`) and a disabled account reporting `locked` while its authority still says `active`. The two `IODUtilsWrapper` mocks under `extended_sources/groups/tests/` were updated for the new interface methods.

Beyond the unit tests, the new OpenDirectory code was compiled and **executed against the live directory** on a macOS 26.5.1 arm64 host, producing the output above.

One note for whoever builds this: `od_wrapper.mm` does not compile under `-fobjc-arc`. The `kOD*` constants are already `NSString *`, so the `__bridge` casts are invalid under ARC — this applies to the pre-existing code as well, and matches the current build, which does not enable ARC.

## Review Checklist

- [x] Code changes reviewed


cborla:
Finding: `sysInfoMac.cpp`'s new `""` default for `user_password_status`/`user_password_hash_algorithm` when not collected ends up stored as NULL on the manager (empty string gets dropped before reaching wazuh-db), while Linux/Windows still use `UNKNOWN_VALUE` (`" "`) for the same "not collected" state and store it as a literal string — a consumer matching on `UNKNOWN_VALUE` won't recognize macOS's NULL as the same condition.

Possible fix: use `UNKNOWN_VALUE` instead of `""` for macOS's default, to keep the "not collected" representation consistent across platforms — or confirm downstream consumers already treat NULL and `" "` as equivalent for these two fields, if that's already the case.

cborla (dismissed):
LGTM
- [x] Code changes reviewed

src/data_provider/src/extended_sources/wrappers/unix/darwin/od_wrapper.mm
```diff
@@ -206,4 +224,161 @@ void genAccountPolicyData(const std::string& uid, nlohmann::json& policyData)
...
+
+                if (!recordName.empty())
+                {
+                    passwordData[recordName] = std::move(entry);
+                }
+            }
+        }
+    }
+
+    void genDisabledUsers(std::set<std::string>& disabledUsers)
```
lchico: `genDisabledUsers` returns early on any OpenDirectory error (session/node/query creation, or `resultsAllowingPartial` failure) with `disabledUsers` left empty and no error signaled. The caller can't tell that apart from "nobody is disabled," so a genuinely disabled account whose password hash is still present ends up reported as `password_status: "active"`. Consider signaling the failure (e.g. an out-param or optional/expected return) so the caller can fall back to a not-collected state instead of a wrong one.

lchico (changes requested):

src/data_provider/src/extended_sources/wrappers/unix/darwin/od_wrapper.mm
nbertoldo: Done

cborla (dismissed):
LGTM

lchico (approved):
LGTM!

- [x] Code changes reviewed

Issue #39062, opened 2026-09-09
Harden the bootstrap: failure taxonomy, TLS diagnostics, --embed-ca and regression coverage


jpcerrone:
**Component:** `https_client`, `client-agent`
**Depends on:** the MVP branch `feat/39021-poc-agent-https-tier2`, and the MVP tasks #39026, #39027, #39028
**Rough size:** M

## Description

What the bootstrap still needs once the MVP branch is complete: the diagnostics, the `--embed-ca` shape, and turning the MVP's proofs into permanent coverage.

Already on `feat/39021-poc-agent-https-tier2`, so **not** in scope here:

- **GET support in the transport** — `curlHandle.cpp`, `curlPerformer.cpp`, `iCurlHandle.hpp`, `httpTypes.hpp`, plus `hc_fetch_cacerts()` and the `hc_cacerts_request_t`/`hc_cacerts_result_t` pair in `https_client.h`.
- **The unverified `/cacerts` fetch** — `cacertsClient.cpp/hpp`, with the verification decision made by the caller rather than re-derived, a component test and a `fakeManager` harness.
- **Prefix routing** — commit `3dd6af1638`, "route GET /cacerts through the configured endpoint prefix".
- **The SPKI pin primitive** — `spkiPin.cpp/hpp` with 684 lines of tests: from PEM, DER or file, all-certificates variant, `spkiPinHex()` and `spkiPinBase64Url()`, and typed errors (`NoCertificate`, `BadCertificate`, `UnsupportedKey`, `Internal`).
- **The whole phase-2 flow** — commit `29ac43b514` closes #39026, #39027 and #39028: `w_agent_token_bootstrap()` decodes the token, fetches and pin-compares the CA (or trusts an embedded one outright), enrolls fully verified, and commits the anchor atomically before the privilege drop.
- **The three negative tests, as permanent component tests** — `tlsVerification_component_test.cpp` (commit `b52cc48eff`) covers an untrusted certificate, a self-signed leaf with the genuine CA appended to the chain, and a CA-signed certificate for the wrong hostname.
- **`--embed-ca`** — the bootstrap trusts a carried CA without fetching, and it was exercised against a real manager.
- **The `/cacerts` HTTP status in the failure log** — commit `6c66020e56`.

So this issue is now only what none of that covers: making the failures read as one coherent scheme, and the two TLS log-level changes.

## Tasks

[...]

PR #39362, opened 2026-09-16
Fall back to the local trust anchor when verify_mode=system cannot verify the manager

### LINKED ISSUE #39123: Settle the TLS resolution ladder: the unanchored default, the system fallback, and the explicit-none latch
**Component:** `client-agent` (`config.c`), `https_client` (`moduleConfig.cpp`)
**Depends on:** the MVP branch `feat/39021-poc-agent-https-tier2`
**Rough size:** M

## Description

What an agent does when it has nothing to verify with — and what it does when it has something but is told not to use it. Both are settled in `w_agent_resolve_ssl_posture()` (`src/client-agent/src/config.c`), and on the MVP branch both differ from the design.

This issue was originally scoped to a single `system` → `full` fallback. Commit `24d9dd7b3a` ("simplify verification mode none scenarios") changed the ground under that: the ladder's last rung is no longer `system`, so there is no longer a `system` default to fall back *from*. The scope is now the rung itself, the fallback, and the latch — they are the same function and cannot be decided apart.

The ladder as it stands:

```
explicit <verification_mode>                 -> honoured, including 'none' with an anchor present
explicit <certificate_authorities>, no mode  -> certificate
anchor file present, no mode                 -> full   (the anchor is injected as the CA)
nothing at all                               -> none   <- design says system
```

## 1. The last rung: `none` or `system`

**An agent with no CA, no anchor and no explicit mode connects unverified.** That is the defect #38940 exists to remove, reachable on a stock install. Branch `5.0.0` resolves that case to `system`.

The reason for the change is real, and it is why this is not a one-line revert: `system` cannot validate a self-signed `remoted.pem`, so resolving there makes a stock install against a private-CA manager fail to connect rather than connect unverified. Trading a silent downgrade for a dead agent is not obviously the better default, and `24d9dd7b3a` chose the connection.

The designed answer to that trade is the fallback below — `system` first, the local anchor when the OS store cannot do the job, and a hard failure only when neither can. That is what makes `system` affordable as the last rung, and it is why the two halves have to be decided together.

- [ ] Decide the last rung, with the fallback in hand rather than before it.
- [ ] Whichever is chosen, make the `<ssl>` block in the packaged `ossec.conf` and the WPK gates' log text say the same thing — both gates currently hard-code the branch's answer.

## 2. The `system` → local-anchor fallback

When the OS trust store cannot verify the manager, resolve to `full` against a certificate under `etc/certs` and log a **warning**; when neither works, log **critical** and fail closed.

- [ ] **Decide what "cannot verify" means.** `os_find_ca_bundle()` is a `stat()` over five hardcoded paths — it fires only on a host with no CA bundle at all, not on the case operators hit, which is a bundle that exists and has never heard of the deployment's CA. Detecting the latter needs a live verified probe or treating the first handshake failure as the trigger. `pkg_installer.sh`'s `probe_server_verified()` already does exactly that for WPK upgrades and pins `DEFAULT_CA_FILE` on failure — same policy, same path, already in the tree.
- [ ] Make the three enforcement points agree: `w_agent_resolve_ssl_posture()`, `w_agent_validate_ssl_ca()` and `moduleConfig.cpp`'s `validateTls()`. A fallback in one and not the others gives a daemon that passes the startup gate and dies at module init.
- [ ] Decide what happens to `(4120)` — `system` with an explicit `certificate_authorities` is a hard refusal today. If a local certificate is now a legitimate fallback under `system`, an operator who configured one is arguably asking for it.
- [ ] Windows and macOS cannot introspect their native stores at validation time, so trigger (a) does not exist there. State what each platform does rather than leaving it implicit.

### The reverse (`full` → `system`) — proposed, and not recommended

The two directions are not symmetric. `system` → `full` **narrows** trust to one pinned, deployment-specific CA. `full` → `system` **widens** it to every CA in the OS store, including any traffic-inspection CA, so a certificate the pin correctly refused would succeed instead. It also contradicts §2.3's requirement that the fetched CA be the **sole** anchor, "not an addition to the system store, not one candidate among several", and the benign case is already covered — *no anchor at all* resolves down the ladder anyway, so the reverse only ever fires when someone asked for pinning and it failed.

- [ ] If it is built anyway: missing-file case only, at startup, never in response to a verification or handshake failure, logged at error.

## 3. The latch: does an explicit `none` survive a present anchor?

The branch honours it, warning `(4122)` with text naming the anchor and the line to remove. Item 33, §2.7 limit 3 and §3.1 say the opposite — keep verifying, log the ignored downgrade at **error**, never refuse to start.

Both are defensible. *For honouring:* an explicit setting is an instruction, and the warning is not silent. *For the design:* "explicitly written" is a poor proxy for "deliberately chosen" here, because §4.1 records that **every current Ansible role and Puppet module ships `verification_mode=none` as its default**. The concrete failure is an agent that bootstraps with a token, pins an anchor, enrols verified — and then a configuration-management converge re-templates `ossec.conf` with the role's default, after which every request goes unverified against a manager it could verify. The design's escape hatch is deliberately the anchor file rather than `ossec.conf`, because that is the file automation does not rewrite.

So the question is **which file is the off switch**, not whether explicit means explicit.

- [ ] Decide, and document the off switch either way.
- [ ] **Either way, remove the dead branch.** The upgrade-to-`full` path is guarded by `!verification_mode_explicit`, and that flag is set wherever `Read_Agent_SSL()` parses the tag (`client-config.c:1099`), while an unparsed mode arrives as `AGENT_VERIFY_UNSET` and is consumed by the first rung — so nothing in production reaches it. It is exercised only by a unit test that sets the flag by hand. A latch that cannot fire should not look like one.

## Definition of Done

- [ ] A stock install with no token, no CA and no anchor lands on the decided posture, and the packaged configuration, both WPK gates and the documentation all state the same one.
- [ ] An agent set to `system` on a host whose OS store cannot verify the manager, with a valid local certificate, connects verified and logs one warning saying it fell back.
- [ ] The same agent with no usable local certificate logs critical and does not start — no silent downgrade, no daemon left running with a dead transport.
- [ ] An agent whose OS store *does* verify behaves exactly as today, with no warning and no fallback.
- [ ] The latch behaves as decided, with a test for the token-install-then-converge sequence specifically.
- [ ] No unreachable branch is left behind in the resolver.
- [ ] The startup gate and the module agree on every combination.

Related: #38940, #39021

MAnDumu:
## Description

The agent's `verification_mode=system` trusts the OS certificate store instead of an operator-supplied CA file. Today, if that store cannot verify the manager's certificate (typically because the manager uses a private/self-signed CA that never reaches the OS trust store), the agent has no alternative: it retries indefinitely without ever being able to connect verified.

This closes the second of three open decisions in #39123 (the `system` → local-anchor fallback). The other two decisions in that issue were already resolved on `5.0.0` before this branch and are not touched here.

## Proposed Changes

- New `TlsFailureDetail` fields (`tlsCertDiagnostics.hpp`, additive to #39062's existing classifier) that distinguish a genuine, unclassified chain/CA-trust failure — a certificate was presented, inspected, and rejected for a trust reason (`sawDepth0` + `depth0VerificationFailed` + `depth0ErrorIsChainTrustRelated`, all true, with `kind` still `None`) — from a hostname mismatch, a certificate-date problem, a cipher/protocol failure after an already-accepted certificate, or a local CA-file load failure, none of which a different trust anchor could ever fix. `TransportStatus` itself is unchanged: every one of these stays an ordinary `TlsFail`.
- `CurlPerformer::perform()` (`curlPerformer.cpp`): under `verification_mode=system`, when the OS trust store fails with an unclassified chain/CA-trust failure (`isUnclassifiedChainFailure()`, see above), the agent logs a warning and retries once against its own trust anchor (`AGENT_ANCHOR_CA`, the same file `full`/`certificate` already use as a default). If that fallback also fails to verify, the agent logs critical and stops (fail-closed) rather than retrying forever against a manager it can never verify.
- New `systemFallbackCaPath` / `system_fallback_ca_path` field threaded from `https_client_bridge.c` (populated from `AGENT_ANCHOR_CA` when it exists and `verify_mode=system`) through `ModuleConfig` and the C ABI (`hc_config_t`) into `CurlPerformer`.
- The two static startup gates (`config.c`'s `w_agent_validate_ssl_ca()` and `moduleConfig.cpp`'s `ModuleConfig::validateTls()`) no longer refuse to start when no OS CA bundle is found on Linux, as long as a local fallback anchor is present and parses as a valid certificate — refusing there would never give the runtime fallback a chance to run. Still a hard refusal: both being absent, or a fallback anchor that is present but fails to parse, **regardless of whether a valid OS bundle also exists** — a corrupt anchor is caught here, at a named startup refusal (`AG_SSL_ANCHOR_UNPARSEABLE`), rather than surfacing as an opaque libcurl `CAINFO` error at the first handshake that reaches it (`config.c`'s own comment documents this as deliberate, not an oversight).
- `(4121)`'s message text updated to reflect the widened condition. `(4120)` (`system` + explicit `<certificate_authorities>` = hard refusal) is unchanged.
- No change on Windows/macOS beyond what already existed: those platforms never had a static OS-bundle probe (they ask the native store at connection time), so the same handshake-failure trigger applies uniformly across platforms without needing one.
- `isUnclassifiedChainFailure()` also engages when OpenSSL's chain builder rejects an untrusted intermediate/root before ever reaching the leaf certificate (a manager fronted by a subordinate CA), via a new `TlsFailureDetail::chainTrustRejectedAboveDepth0` captured independently of the leaf-only fields above. Previously pinned as a known limitation, not covered by the `sawDepth0`-only check.
- A corrupted local fallback anchor file (post-startup) is now detected and fails closed instead of being retried forever (`HttpResponse::caFileLoadFailed`); the fallback no longer retries for X.509 errors unrelated to which CA is trusted (an explicit denylist); the fallback's internal retry no longer silently doubles a caller's configured timeout (`ControlStream::sendShutdown()`'s single-attempt `drain_timeout_ms` in particular, via an injected clock); every test exercising the fallback's retry logic now also runs on Windows/macOS, since that logic itself carries no platform `#ifdef`.

Closes #39123 (partially — the `system` → local-anchor fallback decision only; the other two decisions in that issue are already resolved on `5.0.0`).

### Results and Evidence

Full build (`make -C src TARGET=agent TEST=1`) is clean with no new warnings.

```
https_client_unit_test:      527/527 PASSED
https_client_component_test:  62/62 PASSED (real TLS handshakes over real sockets/OpenSSL)
test_agent_ssl_ca (cmocka):   34/34 PASSED
```

This change went through three rounds of independent review (`code-config-reviewer` + `contrarian-reviewer` cross-check each round) before opening this PR. The first two rounds found and fixed real concurrency bugs in the fallback's latch, since `HttpsClientFacade` shares a single `CurlPerformer` across four threads (control/stateless/stateful/reporter):

1. A raw data race mutating a shared `std::string` + plain `bool` with no synchronization — fixed with a `std::atomic<bool>` flag and keeping both candidate CA paths immutable after construction.
2. A logical TOCTOU on top of that fix: a call could misjudge its own attempt's outcome against a flag another thread had flipped in between — fixed by capturing the flag once per call and threading the decision through as an explicit parameter, never re-read mid-call. Verified deterministically: added a test that reproduces the exact interleaving (`SystemModeConcurrentCallerJudgesItsOwnAttemptNotAnotherThreadsFlag`) and confirmed by hand that it fails against the pre-fix code (3/3 runs) and passes against the fix (3/3 runs, no flakiness).

The third round approved the concurrency fix itself with no new findings there, but its cross-check raised two points worth recording here for reviewers. First, a design point: a single live verification failure on one connection is enough to make the agent exit (fail-closed), which — in a manager deployment with multiple nodes behind a load balancer — could in principle be triggered by one misconfigured/rotating node rather than a deployment-wide condition. This matches what was explicitly requested for this fallback (warn → try local anchor → critical if that also fails) and was kept as-is by design after discussing it. Second, a real coverage gap: the fallback was gated on `sawDepth0` (the verify callback reaching the leaf certificate), which OpenSSL's chain builder can never do when it rejects an untrusted intermediate/root first — a manager fronted by a subordinate CA. That gap was pinned by a test rather than fixed at the time; it is fixed in this branch now (see `chainTrustRejectedAboveDepth0` above).

A fourth round of review followed a separate, external technical review (10 findings) after this PR was opened. Four were confirmed real and fixed here (see "Proposed Changes" above): a corrupted local fallback anchor file going undetected, the fallback retrying for X.509 errors unrelated to CA trust, the fallback's retry silently doubling a caller's configured timeout, and the fallback's own tests being Windows/macOS-guarded despite platform-agnostic logic. A follow-up cross-check found the first two of those were only exercised against a mocked curl handle in unit tests — closed with a new component test against a real curl/OpenSSL `CURLE_SSL_CACERT_BADFILE` failure. A fifth round then closed the `sawDepth0`-only coverage gap noted above (the multi-certificate-chain known limitation). Each fix in both rounds went through its own independent `code-config-reviewer` + `contrarian-reviewer` cycle, all approved with no unresolved findings.

A colleague's own review of this branch (independent of the two review sources above) then found one more real, live bug: `m_usingSystemFallbackAnchor` was being latched `true` *before* the timeout-budget check (fourth round) could return early without ever dialing the fallback anchor — a first attempt that consumed the whole budget would still permanently commit every later call/thread on the object to the fallback anchor, despite never confirming it can verify anything. Fixed by reordering so the budget check runs first; the warning log and the latch now only fire once the code is actually about to dial the fallback. That reordering, in turn, left the budget-exhausted branch completely silent — closed with a distinct log line, then bounded (a sixth round, self-review this time): an unconditional warning there would repeat unboundedly, once per retry from all four threads, under a sustained slow/overloaded manager (not contrived — every real caller's timeout defaults to non-zero). A new `std::atomic<bool> m_budgetExhaustedWarned` fires the warning once per object lifetime via `exchange()` (race-safe across the four threads), then logs at debug level — a one-way latch, deliberately not rearmed on recovery once it was verified (not assumed by analogy to similar-looking patterns elsewhere in this module) that the only place a reset could fire is unreachable once `m_usingSystemFallbackAnchor` has already latched. Each step of this round also went through its own independent review cycle.

Pushing the latch-ordering fix then failed CI's `rtr (client-agent/https_client, winagent)` job on a line-coverage gate (89.9%, threshold 90.0%): the new production lines are platform-agnostic, but their only dedicated test relied on the constructor's Linux-only auto-resolution of `caPath`, so winagent's cross-compiled binary never exercised them. Fixed with a platform-agnostic counterpart test (same technique as the Windows/macOS coverage test above), then extended after a follow-up review found it only exhausted the timeout budget once, covering `m_budgetExhaustedWarned`'s warn branch but never its debug-after branch. **Known, non-blocking follow-up**: that extended test proves the exhausted-budget path returns the right result and never latches the fallback, but cannot yet assert directly on *which* log statement (warn vs. debug) actually ran, since `Log::GLOBAL_LOG_FUNCTION` is a hidden-visibility symbol private to `libhttps_client.so` and not reachable from the test binary. This is not a hard limitation -- `inventory_sync_server` and `remoted_module` already solve the identical problem with an exported test-hook compiled into their own shared library -- just out of scope for this round.

A second colleague then asked to continue with three more findings from that same third external review. Two were confirmed real but judged not to warrant a code change: `AGENT_ANCHOR_CA` being independently re-probed at multiple points (`config.c`'s `w_agent_resolve_ssl_posture()`/`w_agent_validate_ssl_ca()`, `https_client_bridge.c`) is by design, not duplication -- each runs at a genuinely different privilege context (root pre-drop vs. the wazuh user post-drop) and consolidating them would reintroduce exactly the euid-mismatch TOCTOU the existing separation avoids; two threads reaching the fail-closed `exit(1)` concurrently for the first time is real but benign (`exit()` is documented thread-safe in glibc) and not practically testable without forking a child process, well outside this module's existing test style. The third was implemented: `CurlPerformer::m_noOsStoreToTry` (a bool set once alongside `m_config.caPath = systemFallbackCaPath`, never independently updatable) was replaced with a derived method comparing the two fields directly, removing one way a future edit to that single branch could desync the two facts without anything catching it. A follow-up review corrected the reasoning for why this is safe today -- not `ModuleConfig::validate()` (which runs after construction, or never runs, at two of the three real construction sites) but `config.c`'s `w_agent_validate_ssl_ca()`, a C-layer gate at agentd startup -- now documented explicitly, including the narrow case this derived form does not independently guard against if that external gate is ever weakened.

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - N/A

  - N/A

- Wazuh server API/Framework
  - N/A

### Artifacts Affected

- Agent binary (`wazuh-agent`) on Linux, Windows and macOS: the `https_client` transport module and `client-agent`'s config/bridge code.
- No packaged configuration files or default `ossec.conf` values change.

### Configuration Changes

No new user-facing `<ssl>` configuration option. `verification_mode=system`'s behavior changes only in the failure path described above; no default value changes and no new XML element.

### Tests Introduced

- `curlPerformer_test.cpp` (unit, mocked `ICurlHandle`): fallback triggers on an unclassified chain/CA-trust failure and succeeds; fallback also fails and the response reports the same unclassified failure without a third attempt; no retry when no fallback path is configured; deterministic two-thread interleaving reproduction of the TOCTOU fixed in review round 2.
- `moduleConfig_test.cpp` (unit): `validateTls()` accepts `verification_mode=system` with no OS bundle when a fallback anchor is configured.
- `outcomeClassifier_test.cpp` (unit): `TlsFail` (unchanged — this PR does not introduce a new `TransportStatus` value) classifies as `Unreachable`, same as the other transport-error family members; covers the fallback's own exhausted-retry path too.
- `tlsVerification_component_test.cpp` (component, real TLS via OpenSSL/sockets): fallback succeeds when the OS bundle doesn't verify but the fallback anchor does (and stays latched across a second request); fails when neither verifies; fails closed against a real, non-loadable fallback CA file; succeeds when an untrusted intermediate is rejected before the leaf is ever reached (real 3-CA chain, previously the pinned known-limitation test).
- `test_agent_ssl_ca.c` (cmocka): `w_agent_validate_ssl_ca()` now starts with no OS bundle when a local anchor is present, and still refuses when both are absent.
- Fourth/fifth-round additions (`curlPerformer_test.cpp`/`tlsVerification_component_test.cpp`): a corrupt fallback anchor fails closed without a third attempt; a depth-0 X.509 error unrelated to chain trust is never retried; the fallback attempt's timeout is bounded by what the first attempt left of the caller's own budget (and skipped entirely once that budget is exhausted); the fallback's retry-decision logic is also exercised outside the Linux-only guard; a chain-trust rejection above depth 0 triggers the fallback exactly like a depth-0 rejection does.
- `SystemModeDoesNotLatchTheFallbackWhenNoBudgetRemains` (`curlPerformer_test.cpp`): a budget-exhausted call that never dials the fallback anchor must not latch onto it either — a second, independent `perform()` call afterward must still try the OS store first. Verified by hand against the pre-fix ordering too (fails 1/1) before confirming the fix (passes).
- `SystemModeDoesNotLatchTheFallbackWhenNoBudgetRemainsRegardlessOfPlatform` (`curlPerformer_test.cpp`): platform-agnostic counterpart of the above (Windows/macOS coverage), exhausting the budget twice in a row (covering both the warn and the debug-after branch) before a third call confirms the fallback is still never latched.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

Not validated end-to-end against a real manager whose certificate the OS trust store genuinely cannot verify (`/local-deploy`) — all verification above is via unit/component tests with synthetic certificates. Windows/macOS builds were not exercised in this environment; the fallback logic itself is platform-generic (no `#ifdef` branching in `CurlPerformer`/`https_client_bridge.c`), but this has not been confirmed by an actual cross-compiled build.


MAnDumu:
# Updates
## 16/09/2026
Implemented the `verification_mode=system` local-anchor fallback for wazuh#39123 (decision 2 of 3; the other two were already resolved on 5.0.0). The implementation went through three review rounds that found two real concurrency bugs in the fallback's shared state — a raw data race, then a logical TOCTOU on top of that same fix — both verified by reproducing the exact interleaving deterministically and confirming the test failed against the old code and passed against the new one. A real design question also came up (a single live verification failure can kill the whole agent process in a multi-node deployment) and was deliberately kept as-is, matching what was originally requested.
Opened PR #39362 against 5.0.0 with full test evidence (473 unit + 58 component + 34 cmocka, all green). Of the CI checks, fixed a style-formatting issue and a missing CHANGELOG.md entry; one Windows integration test (agentd reconnection) failed twice in a row with no code path found connecting it to the change — flagged as a likely external flake, pending a rerun to confirm.

MAnDumu:
## E2E validation

Validated on a real 5.x agent (fresh parallel install, no overlay) against a real 5.0.0 manager whose certificate is self-signed and not in the OS trust store. This covers the two runtime paths no automated test in this module can exercise, since test binaries never wire `GLOBAL_LOG_FUNCTION` to the real logging/`exit(1)` mechanism.

**Baseline (no local anchor configured):** `system` fails against the manager's self-signed cert with the pre-existing OpenSSL error, no fallback attempted — unchanged legacy behavior.

**Correct local anchor present — fallback succeeds, TLS handshake genuinely completes:**
```
2026/09/17 13:02:24 wazuh-agentd: ERROR: Enrollment request could not be sent: (60) SSL peer certificate or SSH remote key was not OK: SSL certificate OpenSSL verify result: unable to get local issuer certificate (20).
2026/09/17 13:02:34 wazuh-agentd:https-client: WARNING: verify_mode=system: the OS trust store did not verify the manager's certificate; falling back to the local trust anchor ('etc/certs/root-ca.pem').
2026/09/17 13:02:34 wazuh-agentd: ERROR: Enrollment rejected by the manager: it serves no /enroll route under the configured path '/wazuh-manager'. The path in <endpoint> must match the global prefix the manager serves.
[... corrected the test agent's <endpoint> to match the manager's actual global prefix ...]
2026/09/17 13:03:24 wazuh-agentd:https-client: WARNING: verify_mode=system: the OS trust store did not verify the manager's certificate; falling back to the local trust anchor ('etc/certs/root-ca.pem').
2026/09/17 13:03:24 wazuh-agentd: ERROR: Enrollment rejected by the manager: invalid request. Malformed JSON body
```
The error class moves from a TLS handshake failure, to an HTTP routing error, to the manager's own application-level body validation — each step only reachable once the previous layer actually succeeds. Reaching "Malformed JSON body" proves the TLS handshake completed via the fallback anchor and the manager processed the request; the JSON body itself is unrelated to this fix (this throwaway agent isn't provisioned with real enrollment credentials) and wasn't pursued further, as it's out of scope for this validation.

**Wrong local anchor present — fallback also fails, agent stops (the intended safe-failure path):**
```
=== process status ===
wazuh-modulesd is running...
wazuh-logcollector is running...
wazuh-syscheckd is running...
wazuh-agentd not running...
wazuh-execd is running...
=== recent log ===
2026/09/17 13:04:23 wazuh-agentd:https-client: WARNING: verify_mode=system: the OS trust store did not verify the manager's certificate; falling back to the local trust anchor ('etc/certs/root-ca.pem').
2026/09/17 13:04:23 wazuh-agentd:https-client: CRITICAL: https_client: verify_mode=system could not verify the manager's certificate against the OS trust store, and the local fallback anchor ('etc/certs/root-ca.pem') does not verify it either. Refusing to continue unverified.
```
`wazuh-agentd not running...` confirms the process actually exits, not just logs a critical message. This scenario is meant to fail closed — it exercises the safety net when neither trust source can verify the manager.

MAnDumu:
## E2E validation

Validated on a real 5.x agent (fresh parallel install, no overlay) against a real 5.0.0 manager whose certificate is self-signed and not in the OS trust store. This covers the two runtime paths no automated test in this module can exercise, since test binaries never wire `GLOBAL_LOG_FUNCTION` to the real logging/`exit(1)` mechanism.

**Baseline (no local anchor configured):** `system` fails against the manager's self-signed cert with the pre-existing OpenSSL error, no fallback attempted — unchanged legacy behavior.

**Correct local anchor present — fallback succeeds, TLS handshake genuinely completes:**
```
2026/09/17 13:02:24 wazuh-agentd: ERROR: Enrollment request could not be sent: (60) SSL peer certificate or SSH remote key was not OK: SSL certificate OpenSSL verify result: unable to get local issuer certificate (20).
2026/09/17 13:02:34 wazuh-agentd:https-client: WARNING: verify_mode=system: the OS trust store did not verify the manager's certificate; falling back to the local trust anchor ('etc/certs/root-ca.pem').
2026/09/17 13:02:34 wazuh-agentd: ERROR: Enrollment rejected by the manager: it serves no /enroll route under the configured path '/wazuh-manager'. The path in <endpoint> must match the global prefix the manager serves.
[... corrected the test agent's <endpoint> to match the manager's actual global prefix ...]
2026/09/17 13:03:24 wazuh-agentd:https-client: WARNING: verify_mode=system: the OS trust store did not verify the manager's certificate; falling back to the local trust anchor ('etc/certs/root-ca.pem').
2026/09/17 13:03:24 wazuh-agentd: ERROR: Enrollment rejected by the manager: invalid request. Malformed JSON body
```
The error class moves from a TLS handshake failure, to an HTTP routing error, to the manager's own application-level body validation — each step only reachable once the previous layer actually succeeds. Reaching "Malformed JSON body" proves the TLS handshake completed via the fallback anchor and the manager processed the request; the JSON body itself is unrelated to this fix (this throwaway agent isn't provisioned with real enrollment credentials) and wasn't pursued further, as it's out of scope for this validation.

**Wrong local anchor present — fallback also fails, agent stops (the intended safe-failure path):**
```
=== process status ===
wazuh-modulesd is running...
wazuh-logcollector is running...
wazuh-syscheckd is running...
wazuh-agentd not running...
wazuh-execd is running...
=== recent log ===
2026/09/17 13:04:23 wazuh-agentd:https-client: WARNING: verify_mode=system: the OS trust store did not verify the manager's certificate; falling back to the local trust anchor ('etc/certs/root-ca.pem').
2026/09/17 13:04:23 wazuh-agentd:https-client: CRITICAL: https_client: verify_mode=system could not verify the manager's certificate against the OS trust store, and the local fallback anchor ('etc/certs/root-ca.pem') does not verify it either. Refusing to continue unverified.
```
`wazuh-agentd not running...` confirms the process actually exits, not just logs a critical message. This scenario is meant to fail closed — it exercises the safety net when neither trust source can verify the manager.

src/client-agent/https_client/src/curlPerformer.cpp
```diff
@@ -145,6 +156,70 @@ HttpResponse CurlPerformer::perform(const HttpRequestSpec& spec)
...
+    // above, or (wasUsingFallback was already true at the top) the call's first and only
+    // attempt -- also failed to verify the manager. That is always this call's own evidence,
+    // never another thread's, so the conclusion is safe regardless of what anyone else
+    // observed concurrently: neither trust source works, so say why in full and stop, rather
+    // than let the daemon retry forever against a manager it can never verify (#39123's DoD).
+    // Discovered mid-run rather than at startup, but exactly as permanent, so it gets the
+    // same log-then-exit(1) treatment as every other fail-closed TLS misconfiguration in this
+    // module (LOGFN_CRITICAL -> mtLoggingFunctionsWrapper, shared/src/debug_op.c).
+    LOGFN_CRITICAL(m_logFn,
+                   "https_client: verify_mode=system could not verify the manager's "
```
jpcerrone: I would remove the first part of this message (verify_mode=system could not verify the manager's certificate against the OS trust store) since it's going to be logged by the WARNING message earlier

jpcerrone (changes requested):

src/client-agent/https_client/include/https_client.h
```diff
@@ -129,6 +129,10 @@ typedef struct hc_config_t
     ///< `wazuh-agent+jwt` bearer token is signed with.
```
jpcerrone, line 129: I  would try testing this with an agent that's installed using the WAZUH_ENROLLMENT_TOKEN env variable and the WAZUH_SSL_VERIFICATION one. So that we can validate the fallback succeeds and the flow doesn't get stuck in:
`ERROR: Enrollment rejected by the manager: invalid request. Malformed JSON body` 

The two flows I think could be tested are:
* Having the WAZUH_SSL_VERIFICATION variable set verification mode to system during the install, and then starting the agent. The fallback should succeed.
* Having an agent enrolled via the happy path (verification_mode=full, no WAZUH_SSL_VERIFICATION) and working, stop it, change the verification mode to system, start it and validate the fallback works

jpcerrone (changes requested):

src/client-agent/https_client/src/curlPerformer.cpp
MAnDumu: Fixed — trimmed the message so it no longer repeats what the WARN above already logs; it now just says the local fallback anchor doesn't verify either. See e1dd3cbd9b.

src/client-agent/https_client/include/https_client.h
MAnDumu: Tested both flows with a real enrollment token minted from a running manager (`--create-enrollment-token`), verify_mode=system against a self-signed cert not in the OS trust store:

```
wazuh-agentd:https-client: INFO: Starting https_client (server=<manager>:1517, agent=001, body compression enabled).
wazuh-agentd:https-client: WARNING: verify_mode=system: the OS trust store did not verify the manager's certificate; falling back to the local trust anchor ('etc/certs/root-ca.pem').
```

Enrollment completed fully this time (agent registered, `wazuh-agentd` running, all modules started) — unlike the earlier E2E round, which got stuck at "Malformed JSON body" because that throwaway agent had no real credential. The fallback isn't a one-shot either: it's the same latched anchor every subsequent request from that agent uses (SCA sync, etc.), so the "already-enrolled agent switching to system" flow exercises identical code to the fresh-install flow, just without a second install to verify separately.

MAnDumu:
## E2E validation, round 2: real enrollment token, full enrollment completes

Follow-up to the earlier E2E comment above. That round validated the fallback mechanism itself (TLS-layer success, HTTP-layer routing) but never completed a real enrollment, since the throwaway agent had no valid credential. This round mints a real enrollment token from a running manager and drives both flows @jpcerrone asked for.

**Setup:** manager rebuilt from this branch's HEAD, fresh self-signed cert (not in the OS trust store, forcing the fallback), enrollment token minted via `wazuh-manager-authd --create-enrollment-token`.

**Flow 1 — fresh install with `WAZUH_ENROLLMENT_TOKEN` + `WAZUH_SSL_VERIFICATION=system`:**
```
wazuh-agentd:https-client: INFO: Starting https_client (server=<manager>:1517, agent=001, body compression enabled).
wazuh-agentd:https-client: WARNING: verify_mode=system: the OS trust store did not verify the manager's certificate; falling back to the local trust anchor ('etc/certs/root-ca.pem').
```
Enrollment completed fully this time — agent registered as `001`, `wazuh-agentd` running, all modules (execd, syscheckd, logcollector, modulesd) started normally. No TLS errors after the fallback engaged.

**Flow 2 — an already-connected agent staying on the fallback anchor:** the same agent's subsequent traffic (SCA sync, etc.) keeps using the same latched fallback anchor without re-triggering the WARN or re-attempting the OS store — confirming the fallback isn't a one-shot fluke tied to the enrollment call specifically, but the steady-state trust path for the rest of the session. This is the same code path an already-enrolled agent switching `verification_mode` to `system` and reconnecting would exercise.

Manager and agent were left running in this state (not torn down) so the PR can keep being exercised live if needed.

jpcerrone (dismissed):

jpcerrone (dismissed):

jpcerrone:
## Extra work done after e2e testing:
Under `verification_mode=system`, a chain-trust failure with no local anchor configured returned silently, so an agent that could never verify its manager retried indefinitely with nothing in the log but a generic Manager unreachable line that reads like an ordinary network fault; it now names the condition once per run, and keeps retrying rather than failing closed, because the OS trust store is re-consulted on every attempt and an operator adding the CA there recovers the agent without a restart. 

The module's log messages also named `verify_mode`, which is not a real configuration option — they now say `verification_mode`, matching `<ssl><verification_mode>`. Verified end to end on Windows and on Linux against a co-located 5.x manager, three scenarios each: with no anchor and the CA absent from the OS trust store the new warning appears exactly once and roughly thirty seconds ahead of the generic failure while the agent stays up; with the anchor restored the pre-existing fallback still engages and FIM and SCA synchronize; and with the CA present in the OS trust store the agent verifies directly and logs neither warning.

jpcerrone (approved):

PR #39399, opened 2026-09-17
fix: keep macOS queue/sockets at 770 in the package postinstall

### LINKED ISSUE #39161: macOS package postinstall leaves queue/sockets at 750 because the chmod uses a single-element brace that bash does not expand
## Description

The macOS agent package's `postinstall` intends to give `/Library/Ossec/queue/sockets` mode 770, with a comment elsewhere in the script noting that the service account must be able to write to its directories. The chmod that would do it never targets a real path, so the directory ends up at 750 from the following line instead.

Lines 42 to 45 of `postinstall`:

```
42:chown -R ${USER}:${GROUP} ${DIR}/queue/{diff,sockets,rids}
44:chmod -R 770 ${DIR}/queue/{sockets}
45:chmod -R 750 ${DIR}/queue/{diff,sockets,rids}
```

Line 44 uses a brace containing a single element. Bash brace expansion requires either a comma-separated list or a range, so `{sockets}` is left verbatim:

```
$ DIR=/Library/Ossec; echo "${DIR}/queue/{sockets}"; echo "${DIR}/queue/{diff,sockets,rids}"
/Library/Ossec/queue/{sockets}
/Library/Ossec/queue/diff /Library/Ossec/queue/sockets /Library/Ossec/queue/rids
```

The chmod on line 44 is therefore applied to a path that does not exist and fails silently. Line 45, which does expand correctly, then sets 750 on `sockets` along with `diff` and `rids`, so the intended 770 is overwritten in the same breath even if line 44 had worked. The ordering means this cannot be fixed by correcting the brace alone; line 45 would have to stop including `sockets`, or line 44 would have to move after it.

The deployed agent confirms the outcome. On the live macOS host, `queue/sockets` carries 750 rather than 770:

```
drwxr-x---  13 wazuh  wazuh   416 Sep  9 06:26 sockets
```

Practical impact is small, which is why this is filed as Low. The directory is owned by `wazuh`, the account the agent's non-root daemons run as, so the owner's `rwx` bits are sufficient for normal operation and root bypasses the mode entirely. Nothing observably fails today. It is reported because the script's stated intent and its effect differ, the failure is silent, and any future daemon that relies on group write access to that directory would break in a way that is hard to trace back to a brace.

The same `postinstall` also contains a separate, unrelated shell defect in its ossec-to-wazuh ownership migration, reported separately.

## Environment

- Package inspected: `wazuh-agent-5.0.0-latest.arm64.pkg`, downloaded from `https://packages-staging.xdrsiem.wazuh.info/nightly-backup/2026-09-09/`, 7,924,612 bytes, `etag "a1f1c9597d8bf76edb053f71c1e29857"`, `last-modified: Wed, 09 Sep 2026 03:40:58 GMT`.
- `PackageInfo`: `identifier="com.wazuh.pkg.wazuh-agent" version="5.0.0" install-location="/" auth="root"`, 298 payload files. Package is signed (RSA and CMS signatures present in the xar TOC).
- `postinstall` inside the package dated 2026-09-09 03:28.
- Deployed agent where the resulting mode was observed: `demo-env-macos-agent`, id `006`, macOS 26.5.1 build 25F80, arm64, `VERSION.json` `{"version": "5.0.0", "stage": "rc1", "commit": "2161e69"}`, installed 2026-09-09.

## Steps to reproduce

1. Extract the package scripts:

```bash
curl -sk -o agent.pkg https://packages-staging.xdrsiem.wazuh.info/nightly-backup/2026-09-09/wazuh-agent-5.0.0-latest.arm64.pkg
# extract the Scripts entry from the xar heap, then:
gzip -dc Scripts | cpio -idm
```

2. Look at the three lines:

```bash
grep -n "queue/{" postinstall
```

3. Confirm bash does not expand the single-element brace:

```bash
DIR=/Library/Ossec
echo "${DIR}/queue/{sockets}"
echo "${DIR}/queue/{diff,sockets,rids}"
```

The first prints the literal `{sockets}`; the second expands to three paths.

4. On an installed macOS agent, check the resulting mode:

```bash
sudo ls -ld /Library/Ossec/queue/sockets
```

It is `drwxr-x---` (750), not `drwxrwx---` (770).

## Expected

`/Library/Ossec/queue/sockets` ends up with the mode the script intends. Either the brace is written as a plain path and the subsequent 750 chmod excludes `sockets`, or the ordering is changed so the intended mode is applied last.

## Actual

`chmod -R 770 /Library/Ossec/queue/{sockets}` targets a nonexistent literal path and fails silently, and the next line sets `sockets` to 750 along with `diff` and `rids`. The installed directory is 750.

## Not a duplicate of

`qa-known-issues.py check` found **no cache match** and the live search returned nothing, with the tool concluding "no known issue looks like this one". Searches run: the candidate title, plus a live search for `postinstall directory expansion sockets instead element`.

Miguevrgo:
|Related issue|
|---|
|#39161|

## Description

`packages/macos/package_files/postinstall.sh` sets the queue directory modes in two consecutive commands:

```
44:chmod -R 770 ${DIR}/queue/{alerts,sockets}
45:chmod -R 750 ${DIR}/queue/{diff,sockets,rids}
```

Line 45 also covers `sockets`, so the 750 overwrites the 770 that line 44 had just applied, and the installed agent ends up with `drwxr-x---` on `/Library/Ossec/queue/sockets`.

770 is the intended mode: the RPM spec (`packages/rpms/SPECS/wazuh-agent.spec:695`) and the source installer (`src/init/inst-functions.sh:980`) both create `queue/sockets` with 770. `sockets` is removed from the 750 list so the 770 stands, aligning macOS with the other platforms. `diff` and `rids` keep 750.

Closes #39161

## Configuration options

N/A

## Logs/Alerts example

Replaying both chmod lines against a directory tree that mirrors the package layout.

Before:

```
770 queue/alerts
750 queue/diff
750 queue/rids
750 queue/sockets
```

After:

```
770 queue/alerts
750 queue/diff
750 queue/rids
770 queue/sockets
```

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests


packages/macos/package_files/postinstall.sh
```diff
@@ -41,7 +41,7 @@ chown -R root:wheel ${DIR}/lib
 # To the ossec queue (default for agentd to read)
 chown -R ${USER}:${GROUP} ${DIR}/queue/{alerts,diff,sockets,rids}
 
-chmod -R 770 ${DIR}/queue/{alerts,sockets}
+chmod -R 770 ${DIR}/queue/alerts
 chmod -R 750 ${DIR}/queue/{diff,sockets,rids}
```
MarcelKemp: **[S1 · blocking · pre-existing]** This PR removes `sockets` from the 770 line instead of from the 750 line, so `queue/sockets` still ends up at 750 — line 45 runs after this and still lists `sockets`, unchanged from base. That's the same outcome issue #39161 reports, and this PR closes that issue without changing it (confirmed by `git diff <merge-base> HEAD`, which only touches line 44).

Same-repo cross-platform evidence points the other way: `packages/rpms/SPECS/wazuh-agent.spec:695` declares `%dir %attr(770, wazuh, wazuh) .../queue/sockets`, and `src/init/inst-functions.sh:980` installs it with `-m 0770`. Both treat 770 as the correct, deployed mode for this exact directory on Linux/Solaris/AIX. macOS should match.

The underlying defect (line 45 overwriting line 44) predates this PR, but it still blocks here because the PR is specifically closing #39161 on the opposite premise.

```suggestion
chmod -R 770 ${DIR}/queue/{alerts,sockets}
chmod -R 750 ${DIR}/queue/{diff,rids}
```

MarcelKemp (approved):

Issue #878, opened 2018-06-30
How do i use ossec-logtest ?


gagantous:
Maybe this is a dumb question, but i amn quite understand with the ossec-logtest documentation. For example, i want to debug the rule id `11210` , alert level `10` and decoder name `proftpd` i debug this using this command 

` /var/ossec/bin/ossec-logtest -U 11210:10:proftpd`

In my Wazuh manager kibana dashboard, wazuh were produce these log with id 11210, but `ossec-logtest` still looks like this.

```
root@debian:/home/debian# /var/ossec/bin/ossec-logtest -U 11210:10:proftpd
2018/07/01 03:01:41 ossec-testrule: INFO: Started (pid: 25385).
ossec-testrule: Type one log per line.

```

What is the proper way to use ossec-logtest ?
Regards
Fathin

vikman90:
Hi @gagantous,

The `ossec-logtest` tool helps debugging decoders and rules by accepting input strings just like the manager would receive from the agent.

This is an example for the rule 11210: ProFTPD may write a log like this:

> May 31 10:53:25 mail proftpd[15302]: xxxxxxxxxx (::ffff:192.168.0.213[::ffff:192.168.0.213]) - Maximum login attempts (3) exceeded

So, if you want to test this log in the manager, you can either run `ossec-logtest` and type the input string just like it appears in the log, or use an input file or pipe. For instance:

```sh
echo May 31 10:53:25 mail proftpd[15302]: xxxxxxxxxx (::ffff:192.168.0.213[::ffff:192.168.0.213]) - Maximum login attempts (3) exceeded | /var/ossec/bin/ossec-logtest
```

This is a running example for the log above, that would produce an alert of rule 11210:

```sh
# /var/ossec/bin/ossec-logtest
2018/07/01 12:56:08 ossec-testrule: INFO: Started (pid: 5366).
ossec-testrule: Type one log per line.

May 31 10:53:25 mail proftpd[15302]: xxxxxxxxxx (::ffff:192.168.0.213[::ffff:192.168.0.213]) - Maximum login attempts (3) exceeded


**Phase 1: Completed pre-decoding.
       full event: 'May 31 10:53:25 mail proftpd[15302]: xxxxxxxxxx (::ffff:192.168.0.213[::ffff:192.168.0.213]) - Maximum login attempts (3) exceeded'
       timestamp: 'May 31 10:53:25'
       hostname: 'mail'
       program_name: 'proftpd'
[...]

PR #39409, opened 2026-09-17
Resolve the API and indexer credentials at installation instead of shipping defaults

### LINKED ISSUE #39284: Generate random passwords for the default API users (`wazuh`, `wazuh-wui`) at installation

ignaciogalle12git:
## Description

Both default Server API users, `wazuh` and `wazuh-wui`, are created with a password equal to their own
username, in plaintext in `framework/wazuh/rbac/default/users.yaml`. Both carry the `administrator` role, so
anyone able to reach port 55000 on a fresh installation has full administrator access. The manager also has
no way of being told the indexer password it reads with: the guide has the operator write it into the
keystore by hand, on a command line.

This pull request removes those passwords and gives the installation the two credentials it needs. The
manager generates what it owns and receives through the environment what it consumes.

Related to #39284. It does not close it: the producer of the credentials and the dashboard wiring live in
wazuh-installation-assistant#1025, wazuh-docker#2674, wazuh-kubernetes#1660, wazuh-virtual-machines#973 and
wazuh-ansible#2306.

## Proposed Changes

**Installation variables**, read on a fresh installation only. An upgrade ignores them and keeps what the
node has.

| Variable | Sets | Required |
|----------|------|----------|
| `INDEXER_PASSWORD` | password of the indexer's `wazuh-manager` account, stored in the keystore | Yes |
| `INDEXER_USERNAME` | indexer account the manager authenticates as | No, defaults to `wazuh-manager` |
| `INITIAL_WAZUH_PASSWORD` | password of the `wazuh` Server API user | No, generated when absent |
| `INITIAL_WAZUH_WUI_PASSWORD` | password of the `wazuh-wui` Server API user | No, generated when absent |

Exported and preserved with `sudo -E`, not placed after `sudo`, which would put a password in sudo's own
`argv` and in the shell history.

**One shared script**, `src/init/resolve-credentials.sh`, called by the Debian `postinst`, the RPM `%post`
and `install.sh`. It writes the indexer credential into the keystore through the tool's standard input,
provisions the Server API accounts, and never prints a password. In `install.sh` it runs inside `Install()`,
before the block that starts the service: `USER_AUTO_START` defaults to `y` on a manager, and a first API
start with nothing in place generates its own passwords and creates `rbac.db`, which would make every
supplied one a no-op.

**A node that cannot resolve everything is left unconfigured**, not refused. The installation completes and
exits `0`, what resolved is applied, what did not is not written, and the service is left alone:

```
  resolved   Server API passwords for wazuh and wazuh-wui, in /var/wazuh-manager/api/configuration/security/wazuh-preseeded-passwords.yml
  MISSING    indexer password for wazuh-manager, expected in INDEXER_PASSWORD
  The service was not started. To finish:
      https://documentation.wazuh.com/current/installation-guide/wazuh-server/
```

All or nothing per node: one missing consumed credential leaves it unconfigured even when the rest resolved.
The same case exits non-zero with `--container`, where no operator will correct it.

**The credentials file is the only disclosure.** `api/configuration/security/wazuh-preseeded-passwords.yml`
is written `0640` and is what the manager seeds `rbac.db` from at the first API start, so a supplied password
needs it as much as a generated one: the two are different processes and systemd starts the service from a
clean environment. It is kept afterwards, and `wazuh-manager-apid` warns on every start while it is there.

**Other behaviour changes:**

- A user the file does not name is generated when the database is created, so a recreated database and
  `factory-reset` come back working. A file present but unusable still stops the API with error `2012`
  before anything is created.
- `change-password` and `factory-reset` gain `--local`. The default routing executes on the master wherever
  it is typed, so `change-password` on a worker silently retargeted the master and `factory-reset` wiped it.
- `factory-reset` drops its pooled connections before removing the database. They are kept by inode, so one
  left over made `revoke_tokens()` fail after the database had already been replaced, reporting a reset that
  did happen as failed.
- The password rule becomes PCI DSS v4.0 requirement 8.3.6 and nothing else: 12 to 64 characters holding an
  alphabetic and a numeric one. `POST`/`PUT /security/users` now accept passwords the case and symbol rules
  rejected; the `\Z` anchor rejects one ending in a line break. Generated passwords are 32 characters over
  `A-Z a-z 0-9 . , _ + : @ % ^ = ~ -`, about 198 bits.

### Results and Evidence

Unit tests, `PYTHONPATH=api:framework` per `5_testunit_framework.yml`:

```
framework/wazuh/rbac/tests    312 passed
framework/scripts/tests        70 passed
framework/wazuh/tests         297 passed
framework/wazuh/core/tests    820 passed, 1 failed
api/api/test                  551 passed
src/init/tests/test_resolve_credentials.sh    23 checks, 0 failures
```

The failure is `test_manager_conf.py::test_parity_with_the_real_cli`, which compares against a locally built
`src/build/bin/wazuh-manager-conf`. Unrelated to this branch, which touches no C code.

The credential lifecycle against a real `rbac.db`, driving the shipped entry points in order:

```
--- set-password, one user per call        wazuh: SET   wazuh-wui: SET
--- first start: the database is seeded    wazuh: OK    wazuh-wui: OK
--- set-password once the database exists  the file one is not applied: rejected
--- change-password                        the new one: OK   the old one: rejected
--- factory-reset, connection still live   back on the file's password: OK
--- factory-reset with no file             the generated one: OK
```

The `factory-reset` line is the regression fixed here: it previously failed with
`sqlite3.OperationalError: attempt to write a readonly database` after replacing the database.

**Not yet run:** the installation matrix on real packages (DEB and RPM fresh installs, sources, upgrade over
an existing `rbac.db`, master and worker, worker promotion). The scriptlets have only been checked with
`sh -n` and against stubs. An earlier revision broke all four package builds by running the resolution from
`install.sh`, which `debian/rules` and the RPM `%install` use to stage the packaged tree; `WAZUH_PACKAGE_BUILD`
keeps them apart now, and only a real build catches that class of mistake.

### Open items

- **Depends on wazuh/qa-integration-framework#878.** `test_recursion.py` and `test_run_as_chunked_body.py`
  import `get_default_api_password`, which is only on that branch. The integration workflow resolves the QA
  branch by name, so CI passes here; merged first, those two tests break on `5.0.0`. That PR has to land
  before this one.
- **The password rule now differs from the dashboard's.** wazuh-dashboard-plugins still validates
  `[^A-Za-z0-9]` on 5.0.0, so its form is stricter than the API: a password set through `/security/users`
  may be one the dashboard would refuse. Either the plugin follows, or the relaxation is reverted and only
  the `\Z` anchor is kept.
- **`factory-reset` needs an `apid` restart to take effect**, because the running API keeps its connection to
  the unlinked database. Documented in `authentication.md`; the behaviour predates this branch and belongs in
  its own issue.

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X

- Wazuh server API/Framework

No C or C++ code is touched, so the memory and engine checks do not apply.

### Artifacts Affected

- `wazuh-manager` DEB and RPM packages: `postinst`, `postrm` and the `%post` scriptlet.
- `install.sh`, `src/init/inst-functions.sh`, and the new `src/init/resolve-credentials.sh`.
- `bin/rbac_control`.
- `api/configuration/security/wazuh-preseeded-passwords.yml`, a new file the installation writes.
- `framework/wazuh/rbac/default/users.yaml` no longer carries a `password:` key.

### Configuration Changes

The four variables above. An installation that sets none of them completes and leaves the node unconfigured,
where before this branch it installed a manager with a known password. An upgrade is unaffected. The password
rule relaxes, so nothing previously accepted is rejected now except a password ending in a line break.

### Tests Introduced

- `test_orm.py`: the password rule across seven cases, the generator's compliance, uniqueness over 1000 draws
  and its alphabet, generation for a partly provisioned and an unprovisioned node, the atomic write.
- `test_rbac_control.py`: `provision-passwords` generating without disclosing, the environment override and
  its rejection, what is already provisioned being kept, the no-op once the database exists, and the routing
  of `factory-reset` with and without `--local`.
- `test_security.py`: the pool dropped before the database is removed.
- `test_wazuh_manager_apid.py`: the warning while the credentials file is on disk.
- `test_resolve_credentials.sh`: 21 checks over the shared resolution script.

## Review Checklist

- [ ] Code changes reviewed
- [ ] No unresolved dependencies with other issues — wazuh/qa-integration-framework#878 has to merge first


api/scripts/wazuh_manager_apid.py
```diff
@@ -362,6 +336,16 @@ def error(self, msg, *args, **kws):
     # Check for unused PID files
     utils.clean_pid_files(pyDaemonModule.API_MAIN_PROCESS)
 
+    # Seeded before daemonizing so that a failure reaches the caller: `wazuh-manager-control` decides
+    # whether the daemon came up from this process's exit code, and aborts the whole start on a failure.
+    # Running as root here rather than as the Wazuh user is deliberate for the same reason.
+    try:
+        check_database_integrity()
```
ignaciogalle12git: **[S2 · blocking · aggravated]** `check_database_integrity()` now runs as root, and `_set_permissions_and_ownership` does `chown`/`chmod` on `rbac.db` following symlinks inside `api/configuration/security`, which is `root:wazuh-manager 0770` with no sticky bit (api/Makefile:13,32). Before this PR the call sat in `start()`, after `os.setuid(wazuh_uid())`, so a symlink planted by the service account got EPERM; now a compromised `wazuh-manager` user replaces its own `rbac.db` with `rbac.db -> /etc/shadow` and the next restart chowns that file to `wazuh-manager:wazuh-manager 0640` (the start then fails, but the file is already theirs). Nothing between here and `pyDaemon()` needs root (`pyDaemon` only forks/`setsid`, `create_pid` already ran unprivileged), so move the existing "Drop privileges to wazuh" block (lines 355-361) above this `try` and keep the seeding before the fork so the exit status still reaches `wazuh-manager-control`:

```python
    # Seeded before daemonizing so that a failure reaches the caller: `wazuh-manager-control` decides
    # whether the daemon came up from this process's exit code, and aborts the whole start on a failure.
    # As the Wazuh user, not root: `rbac.db` lives in a directory that user can write, so the chown/chmod
    # below must not follow a planted symlink with root's privileges.
    if not args.root:
        if api_conf['drop_privileges']:
            os.setgid(common.wazuh_gid())
            os.setuid(common.wazuh_uid())
    else:
        logger.info('Starting API as root')

    try:
        check_database_integrity()
    except Exception as db_integrity_exc:
        logger.error(f"Error when trying to start the Wazuh API. {APIError(2012, details=str(db_integrity_exc))}")
        sys.exit(1)

    # Foreground/Daemon
    if not args.foreground:
        pyDaemonModule.pyDaemon()
    else:
        logger.info('Starting API in foreground')
```

Optionally harden `_set_permissions_and_ownership` too (`os.open(..., O_NOFOLLOW)` + `fchown`/`fchmod`), since `rbac_control factory-reset` on a master still reaches it as root.

framework/wazuh/rbac/orm.py
```diff
@@ -2525,13 +2742,27 @@ def _set_permissions_and_ownership(database: str):
         # If the database does not exist, it means this is a fresh installation and must be created properly
         else:
             logger.info("RBAC database not found. Initializing")
+
+            # Loaded before anything is created: a missing or unusable file must stop the installation
+            # without leaving a half-seeded database behind.
+            preseeded_passwords = load_preseeded_passwords()
```
ignaciogalle12git: **[S2 · blocking · introduced]** This delivers the provisioning input but not what #39284 asks for in its main case: there is no `secrets`-based generation, no disclosure file and no start-up warning, and a bare DEB/RPM/`install.sh` install (the first three rows of the issue's matrix) now fails at `systemctl start` until two `set-password` calls are run by hand, where today it starts. `factory-reset` re-provisions instead of regenerating, and the plaintext file stays on every worker indefinitely, against the "disclosed exactly once" DoD. None of the five linked child issues lives in this repo or covers the bare-package flow, and GitHub lists #39284 as closed by this PR (Development link) although the body says it does not close it. Either add the fallback the issue (and your own crypto design comment on it) describes here (no file → generate per user, seed, write to a 0400 file, warn; same for reset), or agree the re-scope on the issue, unlink the closing reference, open a child issue in this repo for generation/disclosure and the package flow, and keep the PR in Draft until its dependencies land.

docs/ref/modules/server-api/authentication.md
```diff
@@ -64,7 +64,23 @@ path from 4.x, and an `rbac.db` left by an earlier 5.0 development build keeps w
...
+
+### Provisioning and recovery
+
+The password of each default user comes from `api/configuration/security/wazuh-preseeded-passwords.yml`, provisioned **before** the first API start with `bin/rbac_control set-password -u <user>`, which takes the value from `-p` or from the standard input and writes the file itself, with the ownership and mode below. It runs with every daemon stopped, which is the only window an installer has, and it merges: one user per call, so the entry already provisioned survives. See [Provision the API passwords](../../getting-started/installation.md#provision-the-api-passwords).
+
+The file holds a JSON object in the same shape as `--passwords-file`. A key that is not one of the default users is rejected, and the file must name every default user or none at all: a partial file is rejected the same as an unknown key, since the user it left out would fall back to a different rule per node instead of sharing the one credential the rest of the deployment already has.
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** This paragraph describes a file the loader rejects: it is YAML (`schema_version: 1` + a `manager:` list of `{name, password}`), not "a JSON object in the same shape as `--passwords-file`", and an empty `manager` is refused, so "every default user or none at all" is wrong. Anyone writing the file by hand from this page gets `does not hold a YAML mapping` and a node that does not start.

```suggestion
The file is YAML, in the shape the credentials tooling uses for its own `manager:` block: `schema_version: 1` and a `manager` list of `{name, password}` entries. An entry naming a user that is not a default one, a duplicated entry, a section this manager does not read, another `schema_version`, or a file that leaves a default user out are all rejected: the user left out would otherwise end up with a different rule per node instead of sharing the one credential the rest of the deployment already has.
```

docs/ref/modules/server-api/authentication.md
```diff
@@ -51,7 +51,7 @@ Only `wazuh-wui` can authenticate with an authorization context, because resolvi
 
 The flag on its own does not grant the shipped mappings, which is easy to miss. `RBAChecker.get_user_roles` evaluates a rule holding a reserved ID — the five in `rules.yaml` get IDs `1..5`, while rules created through the API start at `100` — only when the caller is user ID 2. Enabling `allow_run_as` on any other account therefore lets it resolve **custom rules only**, and a context that matches one grants that role whatever the account's own role links say.
 
-Both users are created with **the password shipped in `rbac/default/users.yaml`**, which is the username itself. They are reserved IDs (`<= MAX_ID_RESERVED`), so only another reserved user can change their password — `update_user` needs a `current_user` naming who is asking, which the API takes from the token's `sub`.
+`insert_default_resources` seeds both users from the pre-seed file a deployment is provisioned with (see below), and from the password the package ships for any user that file does not name. They are reserved IDs (`<= MAX_ID_RESERVED`), so only another reserved user can change their password — `update_user` needs a `current_user` naming who is asking, which the API takes from the token's `sub`.
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** "the password the package ships for any user that file does not name" is no longer true: the package ships none, the loader refuses a file that leaves a default user out, and the only unnamed-user path (migration) seeds a random unusable password.

```suggestion
`insert_default_resources` seeds both users from the pre-seed file a deployment is provisioned with (see below); the package ships no password, and the loader refuses a file that leaves a default user out. They are reserved IDs (`<= MAX_ID_RESERVED`), so only another reserved user can change their password — `update_user` needs a `current_user` naming who is asking, which the API takes from the token's `sub`.
```

framework/wazuh/rbac/orm.py
```diff
@@ -44,7 +46,192 @@
...
+
+    Returns
+    -------
+    dict
+        Username to password mapping, with an entry for every default user.
+    """
+    set_every_user = ' and '.join(f"'{_SET_PASSWORD_CMD} -u {username}'" for username in known_usernames)
+
+    try:
+        source = open(PRESEEDED_PASSWORDS_FILE)
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** `open()` without `encoding` decodes with the locale, while `set-password` writes UTF-8 bytes (`.encode()`), so under a non-UTF-8 `LANG` a non-ASCII password is mis-decoded, hashed wrong, the file is consumed and the provisioned credential gets 401 with nothing in the log; the comment below ("not valid UTF-8 raises UnicodeDecodeError") only holds under a UTF-8 locale. Same in the merge read at `rbac_control.py:239`.

```suggestion
        source = open(PRESEEDED_PASSWORDS_FILE, encoding='utf-8')
```

framework/scripts/rbac_control.py
```diff
@@ -165,6 +194,93 @@ async def restore_default_passwords(script_args):
...
+            with open(PRESEEDED_PASSWORDS_FILE) as f:
+                document = yaml.safe_load(f) or {}
+            if not isinstance(document, dict):
+                raise ValueError('it does not hold a YAML mapping')
+            for entry in document.get('manager') or []:
+                provisioned[entry['name']] = entry['password']
+        except (OSError, KeyError, TypeError, ValueError, yaml.YAMLError) as exc:
+            print(f"\tCould not read '{PRESEEDED_PASSWORDS_FILE}': {exc}. Remove it and set every default "
+                  f"user's password again")
+            sys.exit(1)
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** The merge reads whatever already sits at this path with none of the owner/mode checks the loader enforces, then rewrites it as `root:wazuh-manager 0640`: a file a member of the `wazuh-manager` group dropped in the 0770 directory before provisioning (say `manager: [{name: wazuh-wui, password: ...}]`) becomes trusted after the first `set-password`, and "Still missing" is not printed. Reuse the check on the descriptor being read and treat its error like the other read failures (add `PreseededPasswordsError, _assert_preseed_source_is_trusted` to the `from wazuh.rbac.orm import ...` at lines 217-218):

```suggestion
            with open(PRESEEDED_PASSWORDS_FILE) as f:
                _assert_preseed_source_is_trusted(f.fileno())
                document = yaml.safe_load(f) or {}
            if not isinstance(document, dict):
                raise ValueError('it does not hold a YAML mapping')
            for entry in document.get('manager') or []:
                provisioned[entry['name']] = entry['password']
        except (OSError, KeyError, TypeError, ValueError, yaml.YAMLError, PreseededPasswordsError) as exc:
            print(f"\tCould not read '{PRESEEDED_PASSWORDS_FILE}': {exc}. Remove it and set every default "
                  f"user's password again")
            sys.exit(1)
```

framework/scripts/rbac_control.py
```diff
@@ -165,6 +194,93 @@ async def restore_default_passwords(script_args):
...
+
+    provisioned[script_args.user] = password
+
+    # The same shape the credentials tooling uses for its own `manager:` block
+    document = {'schema_version': PRESEEDED_PASSWORDS_SCHEMA_VERSION,
+                'manager': [{'name': name, 'password': value} for name, value in provisioned.items()]}
+
+    # Renamed over the target so a reader never sees a half-written file, with the ownership and mode the
+    # API requires: readable by the Wazuh user, writable by nobody else.
+    fd, tmp_path = mkstemp(dir=path.dirname(PRESEEDED_PASSWORDS_FILE))
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** `mkstemp` sits outside the `try` that turns write failures into `exit 1`, and `__main__` prints `Internal error: ...` and returns 0 on any exception, so `set-password && set-password && systemctl start` keeps going when the directory is missing or the caller lacks privileges, and the start fails later with "no API credentials have been provisioned" (`factory-reset` failures, including the new 5012, also exit 0: `reset_rbac_database` never calls `sys.exit(1)`). One guard where every subcommand lands fixes all of them:

```python
if __name__ == "__main__":
    args = get_script_arguments()

    try:
        asyncio.run(main())
    except WazuhError as e:
        print(f"Error {e.code}: {e.message}")
        sys.exit(1)
    except Exception as e:
        print(f"Internal error: {e}")
        sys.exit(1)
```

tests/integration/test_api/test_miscs/test_recursion.py
```diff
@@ -43,18 +43,17 @@
 from wazuh_testing.constants.api import (
     WAZUH_API_PROTOCOL,
 )
-from wazuh_testing.modules.api.utils import get_base_url
+from wazuh_testing.modules.api.utils import get_base_url, get_default_api_password
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** `get_default_api_password` exists only on the qa-integration-framework branch of #868 (open, base 5.0.0); it is absent from that repo's `5.0.0` and `main`. This PR's CI is green because `5_testintegration_manager.yml` clones the QA branch with the same head-ref name; merged into 5.0.0 before #868, the workflow clones QA `5.0.0` and this import fails in `test_recursion.py` and `test_run_as_chunked_body.py`. No code change needed: merge #868 first and name the dependency in the description (the "No unresolved dependencies" box is unchecked but the dependency is not listed).

framework/wazuh/rbac/orm.py
```diff
@@ -44,7 +45,222 @@
...
+_SET_PASSWORD_CMD = os.path.join(WAZUH_PATH, "bin", "rbac_control") + " set-password"
 CURRENT_ORM_VERSION = 1
+
+# Minimum twelve characters, at least one uppercase letter, one lowercase letter, one number and one
+# symbol out of `. * + ? -`, the set both authentication realms of a deployment share.
+# `\Z` rather than `$`, which would also match before a trailing newline.
+USER_PASSWORD_MIN_LENGTH = 12
+USER_PASSWORD_MAX_LENGTH = 64
+USER_PASSWORD_SYMBOLS = '.*+?-'
+USER_PASSWORD_POLICY = re.compile(r'^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[.*+?-]).{12,}\Z')
```
ignaciogalle12git: **[S2 · blocking · introduced]** Narrowing the symbol class to `[.*+?-]` breaks `POST`/`PUT /security/users` for every client that uses `!`, `@`, `#`, `_`…: `Password123!` was accepted and now returns 5007, and wazuh-dashboard-plugins 5.0.0 still validates its user forms with `[^A-Za-z0-9]` (`create-user.tsx:120`, `edit-user.tsx:105`), so the form accepts what the API refuses, and none of the linked child issues touches that repo. The narrowing is not needed for #39284 either: the assistant's passwords tool generates with `tr -dc "A-Za-z0-9.*+?"` and validates with `[.*+?-]` (`passwordsFunctions.sh:227,153`), a subset of the old policy, so whatever it produces already passed. Keep `\Z`, revert the class, drop `USER_PASSWORD_SYMBOLS` and restore "a symbol" in the two `rbac_control` messages, in the 5007 remediation and in the changelog/docs (the tests' `!`→`-` churn can stay).

```suggestion
# Minimum twelve characters, at least one uppercase letter, one lowercase letter, one number and one
# non-alphanumeric character. `\Z` rather than `$`, which would also match before a trailing newline.
USER_PASSWORD_MIN_LENGTH = 12
USER_PASSWORD_MAX_LENGTH = 64
USER_PASSWORD_POLICY = re.compile(r'^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[^A-Za-z0-9]).{12,}\Z')
```

framework/wazuh/rbac/orm.py
```diff
@@ -2525,13 +2760,27 @@ def _set_permissions_and_ownership(database: str):
         # If the database does not exist, it means this is a fresh installation and must be created properly
         else:
             logger.info("RBAC database not found. Initializing")
+
+            # Loaded before anything is created: a missing or unusable file must stop the installation
+            # without leaving a half-seeded database behind.
+            preseeded_passwords = load_preseeded_passwords()
```
ignaciogalle12git: **[S2 · blocking · introduced]** Still the gap from iteration 1: this is where #39284 puts the fallback (no file → generate with `secrets`, seed, write the value to a 0400 file, warn), and without it a bare DEB/RPM/`install.sh` install does not start until two manual `set-password` calls run — your issue comment of 2026-09-18 says as much ("There is no fallback"). `factory-reset` does not regenerate either. GitHub's Development link still lists #39284 as closed by this PR although the body says it does not close it. Either add the fallback here, or agree the re-scope on the issue, remove the closing link, open a child issue in this repo for generation/disclosure and the package flow, and keep the PR in Draft until qa-integration-framework#878 and the five child issues land.

framework/wazuh/rbac/orm.py
```diff
@@ -2525,13 +2760,27 @@ def _set_permissions_and_ownership(database: str):
...
+
             db_manager.connect(DB_FILE)
```
ignaciogalle12git: **[S3 · non-blocking · aggravated]** SQLite opens this path with `O_CREAT` and follows symlinks, and `os.path.exists` is False for a dangling link, so a planted `rbac.db -> /etc/sudoers.d/x` takes the fresh-install path and gets created at the target before `_set_permissions_and_ownership` refuses it with ELOOP. apid now runs this as the Wazuh user, but `rbac_control change-password --local` on a worker forwards `ensure_rbac_database` as `local_any`, which `execute_local_request` runs in the script's own root process (no `process_pool`), and a worker has no `rbac.db` by design — no race needed. Content is fixed (schema only), so this is a DoS primitive rather than an escalation, but the deterministic case is one guard away:

```suggestion
            if os.path.lexists(DB_FILE):
                raise PreseededPasswordsError(f"'{DB_FILE}' exists but is not a database file. Remove it and "
                                              f"start again")
            db_manager.connect(DB_FILE)
```

framework/wazuh/rbac/orm.py
```diff
@@ -44,7 +45,222 @@
...
+        raise PreseededPasswordsError(f"'{PRESEEDED_PASSWORDS_FILE}' cannot be opened: {exc}") from exc
+
+    with source:
+        _assert_preseed_source_is_trusted(source.fileno())
+        try:
+            document = yaml.safe_load(source)
+        # ValueError rather than YAMLError alone: a file that is not valid UTF-8 raises UnicodeDecodeError.
+        except (yaml.YAMLError, ValueError) as exc:
+            raise PreseededPasswordsError(f"'{PRESEEDED_PASSWORDS_FILE}' is not valid UTF-8 YAML: "
+                                          f"{exc}") from exc
```
ignaciogalle12git: **[S2 · blocking · introduced]** `str(exc)` of a `MarkedYAMLError` includes the offending source line (verified: `password: Sup3r-Secret*Pass: oops` yields `mapping values are not allowed here … line 3, column 32:\n    password: Sup3r-Secret*Pass: oops`), so a hand-written or tool-written file with a stray `:` puts the plaintext administrator password into `logs/api.log` — rotated, archived, shipped in support bundles — and it stays there after the operator fixes the file and the loader consumes it. That is the opposite of the "The value itself is never logged" comment a few lines down. `exc.problem` is the short phrase without content; keep the line number and drop the chained cause, which any `exc_info=True` logger (the DAPI's "Unhandled exception" path) would print again. The merge path in `rbac_control.py:245` prints `{exc}` for the same error; same treatment.

```suggestion
        except (yaml.YAMLError, ValueError) as exc:
            # Never `str(exc)`, and no chained cause: a MarkedYAMLError echoes the offending source line,
            # password included, and this message ends up in api.log.
            mark = getattr(exc, 'problem_mark', None)
            raise PreseededPasswordsError(f"'{PRESEEDED_PASSWORDS_FILE}' is not valid UTF-8 YAML: "
                                          f"{getattr(exc, 'problem', None) or type(exc).__name__}"
                                          f"{f' (line {mark.line + 1})' if mark else ''}") from None
```

framework/wazuh/rbac/orm.py
```diff
@@ -44,7 +45,222 @@
...
+
+    Returns
+    -------
+    dict
+        Username to password mapping, with an entry for every default user.
+    """
+    set_every_user = ' and '.join(f"'{_SET_PASSWORD_CMD} -u {username}'" for username in known_usernames)
+
+    try:
+        source = open(PRESEEDED_PASSWORDS_FILE, encoding='utf-8')
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** A FIFO at this path (`mkfifo`, any account in the `wazuh-manager` group can, the directory is 0770) blocks this `open()` forever: apid hangs before daemonizing under `wazuh-manager-control`'s start lock until systemd times out, and `set-password`/`factory-reset` hang the same way. `_set_permissions_and_ownership` got `O_NONBLOCK` + `S_ISREG` in this PR for exactly this reason ("so that a fifo left in the database's place answers instead of blocking the start forever"); the provisioning file, in the same directory and writable by the same accounts, needs the same. Add `if not S_ISREG(stat.st_mode): raise PreseededPasswordsError(...)` to `_assert_preseed_source_is_trusted` as well, otherwise an empty non-blocking FIFO surfaces as `BlockingIOError`.

```suggestion
    try:
        # O_NONBLOCK so that a fifo left at this path answers instead of blocking the start forever, as
        # `_set_permissions_and_ownership` already does for the database.
        source = open(os.open(PRESEEDED_PASSWORDS_FILE, os.O_RDONLY | os.O_NOFOLLOW | os.O_NONBLOCK),
                      encoding='utf-8')
```

framework/wazuh/rbac/orm.py
```diff
@@ -44,7 +45,222 @@
...
+    if stat.st_uid not in (0, wazuh_uid()):
+        raise PreseededPasswordsError(f"'{PRESEEDED_PASSWORDS_FILE}' is not owned by root or by the Wazuh user")
+
+    if stat.st_mode & (S_IWGRP | S_IWOTH):
+        raise PreseededPasswordsError(f"'{PRESEEDED_PASSWORDS_FILE}' is writable by its group or by others")
+
+    # Group read is what lets apid read a root-owned file after dropping privileges, so only the world bit
+    # is refused: a plaintext administrator password must not be readable by every account on the host.
+    if stat.st_mode & S_IROTH:
+        raise PreseededPasswordsError(f"'{PRESEEDED_PASSWORDS_FILE}' is readable by others")
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** Owner, write bits and world-read are checked, but group-read is accepted for any `st_gid`: a file hand-provisioned as `root:users 0640` (the docs describe hand-writing) passes, and every login in `users` can read the administrator password — indefinitely on a worker, where the file is never consumed. The reason group-read is allowed ("what lets apid read a root-owned file after dropping privileges") only holds for the Wazuh group or root's. `S_IRGRP` needs importing, and the loader tests will have to patch `wazuh_gid` the way they already patch `wazuh_uid`.

```suggestion
    # Group read is what lets apid read a root-owned file after dropping privileges, so it is allowed for
    # root's group and the Wazuh group only: a plaintext administrator password must not be readable by
    # every account on the host, nor by whichever group a hand-written file happened to get.
    if stat.st_mode & S_IROTH or (stat.st_mode & S_IRGRP and stat.st_gid not in (0, wazuh_gid())):
        raise PreseededPasswordsError(f"'{PRESEEDED_PASSWORDS_FILE}' is readable by others or by a group other "
                                      f"than the Wazuh group")
```

framework/wazuh/rbac/orm.py
```diff
@@ -2525,13 +2760,27 @@ def _set_permissions_and_ownership(database: str):
         # If the database does not exist, it means this is a fresh installation and must be created properly
         else:
             logger.info("RBAC database not found. Initializing")
+
+            # Loaded before anything is created: a missing or unusable file must stop the installation
+            # without leaving a half-seeded database behind.
+            preseeded_passwords = load_preseeded_passwords()
+
             db_manager.connect(DB_FILE)
             db_manager.create_database(DB_FILE)
```
ignaciogalle12git: **[S3 · non-blocking · aggravated]** Any failure after this line on the fresh path (an `add_user` error, ENOSPC, a crash) leaves a version-0 `rbac.db`. Before this PR the next start migrated it and came up with the defaults; now the migration branch computes `preserved_default_users`, ignores the provisioning file that is still sitting next to the database, and refuses with "Remove rbac.db and provision the credentials again" — so a transient error turned into a manual step, while the docs promise "no database is created" and the log says "Nothing has been applied". Remove the file when the fresh branch fails so the next start takes this same path again:

```python
            preseeded_passwords = load_preseeded_passwords()
            try:
                db_manager.connect(DB_FILE)
                db_manager.create_database(DB_FILE)
                _set_permissions_and_ownership(DB_FILE)
                db_manager.insert_default_resources(DB_FILE, preseeded_passwords)
                db_manager.set_database_version(DB_FILE, CURRENT_ORM_VERSION)
                db_manager.close_sessions()
            except Exception:
                # Nothing rather than a half-seeded database: the next start must take this path again
                db_manager.close_sessions()
                os.path.lexists(DB_FILE) and os.remove(DB_FILE)
                raise
```

framework/wazuh/core/security.py
```diff
@@ -122,8 +122,55 @@ def sanitize_rbac_policy(policy):
...
+
+    Returns
+    -------
+    dict
+        Confirmation message. Required, not optional: `forward_function` wraps the result in a
+        `WazuhResult`, which rejects `None` with error 1000.
+    """
+    if not os.path.exists(DB_FILE):
+        check_database_integrity()
+
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** `PreseededPasswordsError` is a plain `Exception`, and `DistributedAPI.distribute_function` turns every non-Wazuh exception into `WazuhInternalError(1000)` with no message (dapi.py:200-205). So `rbac_control change-password --local -u wazuh` on a worker that has neither `rbac.db` nor a provisioning file prints `Could not ensure the RBAC database exists: Error 1000 - Wazuh Internal Error` plus a raw traceback from the unconfigured logger, and the "Set them with rbac_control set-password …" hint never reaches the operator as a message. `rbac_db_factory_reset` below already wraps the same error in 5012; do the same here and reword 5012 in `exception.py` to "cannot be seeded" so it covers both.

```suggestion
        try:
            check_database_integrity()
        except PreseededPasswordsError as exc:
            raise WazuhError(5012, extra_message=str(exc))
```

framework/wazuh/core/security.py
```diff
@@ -122,8 +122,55 @@ def sanitize_rbac_policy(policy):
...
+    return {'ensured': True}
+
+
+@dapi_allower()
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** Before this PR `factory-reset` run on a worker ended in the master refusing the unmarked callable; exposing it makes the hard-coded `local_master` routing in `reset_rbac_database` effective, so the command now wipes and reseeds the **master's** `rbac.db` from a worker, with a prompt and a result that never name the node. The PR added `--local` to `change-password` precisely because "the default local_master routing runs on the master and reports success there even from a worker"; the destructive command needs the same flag and should print its target (and, with F6, exit 1 when the response is an exception):

```python
    reset_parser.add_argument("--local", action="store_true", dest="local", default=False,
                              help="Reset this node's own RBAC database instead of the master's.")
    ...
    request_type = "local_any" if script_args.local else "local_master"
    response = await cluster_utils.forward_function(rbac_db_factory_reset, request_type=request_type)
```

framework/scripts/rbac_control.py
```diff
@@ -165,6 +195,93 @@ async def restore_default_passwords(script_args):
...
+    # not name every default user, so the entries already provisioned have to survive.
+    provisioned = {}
+    if os.path.exists(PRESEEDED_PASSWORDS_FILE):
+        try:
+            with open(PRESEEDED_PASSWORDS_FILE, encoding='utf-8') as f:
+                document = yaml.safe_load(f) or {}
+            if not isinstance(document, dict):
+                raise ValueError('it does not hold a YAML mapping')
+            for entry in document.get('manager') or []:
+                provisioned[entry['name']] = entry['password']
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** Entries are copied verbatim, so a stale `{name: otro, password: x}` left by another tool survives every merge: `set-password -u wazuh` and `-u wazuh-wui` both print `SET`, the loader refuses the file ("names 'otro', which is not a default user"), and rerunning `set-password`, which is what its own message tells the operator to do, never removes it; only `rm` does, and nothing says so. `default_users` is already in scope two lines up:

```suggestion
            for entry in document.get('manager') or []:
                if entry['name'] in default_users:
                    provisioned[entry['name']] = entry['password']
```

framework/scripts/rbac_control.py
```diff
@@ -165,6 +195,93 @@ async def restore_default_passwords(script_args):
...
+        except (OSError, KeyError, TypeError, ValueError, yaml.YAMLError) as exc:
```
ignaciogalle12git: **[S3 · non-blocking · introduced]** Still open from iteration 1: the merge reads the existing file without the owner/mode check the loader applies, then rewrites it as `root:wazuh-manager 0640`, so an entry planted by a member of the `wazuh-manager` group (the directory is 0770) becomes trusted after the operator's first `set-password` call. Reuse the loader's check on the open descriptor and import `_assert_preseed_source_is_trusted, PreseededPasswordsError` from `wazuh.rbac.orm` on line 212:

```suggestion
            with open(PRESEEDED_PASSWORDS_FILE, encoding='utf-8') as f:
                _assert_preseed_source_is_trusted(f.fileno())
                document = yaml.safe_load(f) or {}
            if not isinstance(document, dict):
                raise ValueError('it does not hold a YAML mapping')
            for entry in document.get('manager') or []:
                provisioned[entry['name']] = entry['password']
        except (OSError, KeyError, TypeError, ValueError, yaml.YAMLError, PreseededPasswordsError) as exc:
```

framework/scripts/rbac_control.py
```diff
@@ -165,6 +195,93 @@ async def restore_default_passwords(script_args):
...
+    # The `manager:` block of the deployment's credentials file, verbatim in shape
+    document = {'manager': [{'name': name, 'password': value} for name, value in provisioned.items()]}
+
+    # Renamed over the target so a reader never sees a half-written file, with the ownership and mode the
+    # API requires: readable by the Wazuh user, writable by nobody else.
+    # `mkstemp` inside the guard, not before it: an installer chains these calls on the exit status, and a
+    # missing directory or an unprivileged run has to fail the chain rather than report success.
+    tmp_path = None
+    try:
+        fd, tmp_path = mkstemp(dir=path.dirname(PRESEEDED_PASSWORDS_FILE))
```
ignaciogalle12git: **[S3 · non-blocking · aggravated]** The `mkstemp` half of F6 is fixed; the exit-status contract still is not. `__main__` (lines 363-366) prints and exits 0 on any exception, and `reset_rbac_database` (line 296) prints "reset failed" and exits 0 — which now includes this PR's own 5012 refusal, i.e. the normal outcome of every `factory-reset` on a node that has not been provisioned again, so `factory-reset -f && systemctl restart` carries on as if the reset happened. One guard covers every subcommand:

```python
    try:
        asyncio.run(main())
    except WazuhError as e:
        print(f"Error {e.code}: {e.message}")
        sys.exit(1)
    except Exception as e:
        print(f"Internal error: {e}")
        sys.exit(1)
```

plus `sys.exit(1)` in `reset_rbac_database` when `response` is an exception.

tests/integration/test_api/test_miscs/test_recursion.py
ignaciogalle12git: **[S3 · non-blocking · introduced]** `get_default_api_password` still does not exist on qa-integration-framework `5.0.0`: #868 was closed unmerged and the live branch (`issue39284-preseed-via-certs-tool`) is #878, still open. CI is green here only because the workflow clones the QA branch by this PR's head-ref name; merged first, `5_testintegration_manager.yml` on `5.0.0` fails at this import in both `test_recursion.py` and `test_run_as_chunked_body.py`. Merge #878 first and name the dependency in the PR body.

PR #39414, opened 2026-09-17
Publish syscollector deltas to the indexer only after wazuh-db confirms the write

### LINKED ISSUE #39329: Entry present in the Indexer but missing from the agent's wazuh-db after a regular Syscollector scan
| Wazuh version | Component | Install type | Install method | Platform |
|---|---|---|---|---|
| 4.14.8-rc1 | Syscollector / wazuh-db / Indexer sync | Manager/Agent | Packages | Windows 11 Pro x86_64 (agent), macOS Sonoma (agent), CentOS 7 (agent), Ubuntu 22 (agent), Ubuntu 22 (managers) |

## Description

During the [Release 4.14.8 - RC1 IT Hygiene E2E tests](https://github.com/wazuh/wazuh/issues/39251) (build [#5](https://jenkins-production.qa.wazuh.info/job/4.14.8/job/Test_e2e_system/5/)), `test_browser_extensions_regular_scan` failed.

The test performs a second Syscollector scan (2-minute interval, no changes expected) and compares, per agent:
1. `browser_extensions` data indexed in the Indexer (`wazuh-states-inventory-browser-extensions-wazuh`).
2. Data from the agent's local `wazuh-db` (`sys_browser_extensions` table, queried via `wdb-query.py` on the manager).

For the Windows agent, after the second scan:
- Indexer: **41** records
- wazuh-db: **40** records

Diff:
```json
"agent2": {
    "only_in_wdb": [],
    "only_in_index": [
        [
            "ie",
            "HKEY_USERS\\S-1-5-21-3278051518-3114999403-1073775594-1002\\SOFTWARE\\Microsoft\\Internet Explorer\\URLSearchHooks\\{CFBFAE00-17A6-11D0-99CB-00C04FD64497}",
            "", "",
            "Microsoft Url Search Hook",
            "",
            "C:\\Windows\\SysWOW64\\ieframe.dll",
            "11.0.22621.755",
            true, false, false
        ]
    ]
}
```

This entry is an IE URLSearchHook registered under a real user's SID (not a system SID like S-1-5-19/S-1-5-20, which were present and matched in both sources). In the first scan (test_browser_extensions_first_scan, which passed), this entry did not exist yet, and both sources matched 40/40 exactly.

In other words: this specific inventory record reached the Indexer but was not yet present in the manager's local wazuh-db within the test's 60s sync-wait window (DB_INDEXER_SYNC_TIMEOUT).

### Current result

Indexer and wazuh-db diverge for a `browser_extensions` record after a regular (non-first) Syscollector scan.

### Expected result

wazuh-db and Indexer content for `browser_extensions` must match after each scan + sync window.

### Additional notes

- No ERROR/CRITICAL log entries were found on the agent or manager in the relevant timeframe — no crash/exception involved.
- The Syscollector scan itself completed successfully on all agents.
- Environment: 2 managers (cluster), 6 agents (Windows/macOS/CentOS/Ubuntu), syscollector scan interval configured to 2m for the test.
- This looks like a timing/synchronization issue between wazuh-db and the Indexer for newly appearing inventory records (as opposed to a functional Syscollector bug), consistent with the mechanism already described in #32777.

### Evidences

- Full pytest report: [archive.zip](https://github.com/user-attachments/files/32279151/archive.zip)

Antoniogm03:
|Related issue|
|---|
|#39329|

## Description

Syscollector inventory records could become visible in the Wazuh Indexer before, or without ever, being written to `wazuh-db`. Root cause and reproduction are detailed in the issue comments: `wazuh-remoted` forwarded every syscollector delta (`INSERTED`/`MODIFIED`/`DELETED`) to the `deltas-syscollector` router topic as soon as it received it, in parallel with and independent of the write to `wazuh-db` through `analysisd`. The path to `wazuh-db` goes through a queue, a decode step and (when configured) an EPS limit; the path to the Indexer has none of those, so under load it reliably finished first, with nothing gating the Indexer on `wazuh-db` having applied the same change.

This PR makes `wazuh-db` publish the delta to `deltas-syscollector` itself, only after `process_dbsync_data()` confirms the write, reusing the same topic and FlatBuffer schema `inventory_harvester` already consumes (so `inventory_harvester` needs no changes). `wazuh-remoted` no longer forwards syscollector deltas independently.

Follow-up in this same PR: the first version above published the delta with `agent_name`/`agent_ip`/`agent_version` blank, because resolving them from `global.db` inside `wazuh-db`'s per-agent request path deadlocks (see Logs/Alerts example). `wazuh-remoted` already resolves these three values for every connection, the same ones it already uses for the `rsync` path, so it now carries them alongside the raw delta it sends to `analysisd`, encoded as a `<name>\x01<ip>\x01<version>\x01` marker right after the `d:syscollector:` header. `analysisd`'s syscollector decoder strips that marker and adds the three fields to the JSON `data` it forwards to `wazuh-db`; the table upsert only binds the columns it recognizes, so the extra keys are ignored there. `wazuh-db` reads them back out when publishing the confirmed delta, instead of hardcoding empty strings. No `wazuh-db` table or schema changes: these fields are never persisted, only passed through to the Indexer message.

Second follow-up, from a code review pass on the agent-metadata change above:
- `network_address` deltas were silently dropped end-to-end: `analysisd` rewrites `proto` from int to `"ipv4"`/`"ipv6"` for the SQL upsert, but `dbsync_network_address.proto` is a `long` in the FlatBuffer schema, so the Indexer publish failed to parse and was discarded with only an `mdebug2`. `wdb_parse_dbsync()` now restores the numeric value before publishing.
- `wazuh-db` could publish an agent's own inventory (agent `000`) twice: once correctly from `wm_syscollector`'s own direct publish, once with an empty agent from `wazuh-db`, racing the same document and sometimes overwriting the good copy with a blank one. `wazuh-db` now only publishes when `agent_name` is a real, non-empty value, since only deltas relayed by `remoted` carry one.
- an agent could desync the `\x01` marker by reporting a version containing that byte in its `HC_STARTUP` (unvalidated, `cJSON` accepts `\u0001` escapes), corrupting every subsequent real-time delta from that agent until its next reconnect. `remoted` now blanks the version if it contains the separator before building the marker.
- a `DELETED` delta that failed locally (bind or cache-stmt error) was still published as confirmed, because `process_dbsync_data()` discarded `wdb_delete_dbsync()`'s return value and hardcoded success. Now propagates the real result, matching the `INSERTED`/`MODIFIED` path.
- `remoted`'s and `wazuh-db`'s unit tests didn't compile or didn't cover any of this (`router_syscollector_handle` still referenced after removal, 34 query assertions missing the new fields, no coverage for the marker parsing or the publish gate). Fixed and added tests for all of it.

Known limitation: with the write now gating the Indexer publish, Indexer freshness inherits whatever slows down or drops events on the path to `wazuh-db` - the `analysisd` decode queue and, when configured, `<limits><eps>` (a single credit pool shared by every decoder, not just Syscollector). Before this PR, the Indexer path was independent of both. Syscollector already self-heals a dropped delta on its next full `rsync` pass; other decoders without an equivalent reconciliation mechanism were already exposed to the same permanent event loss from that queue, unrelated to this PR.

## Configuration options

None.

## Logs/Alerts example

Reproduced with a single manager/agent pair, using `<limits><eps><maximum>2</maximum><timeframe>1</timeframe></eps></limits>` on the manager to make the race deterministic. Opening a new listening port on the agent, before the fix:

```
11:36:48  agent sends dbsync_ports INSERTED (local_port 61234)
11:36:48  inventory_harvester already builds the Indexer document (same second)
11:36:54  Indexer confirms "result":"created"
11:36:48 -> 11:38:21  wazuh-db polled every second: never returns the row
11:39:52  Indexer confirms "result":"deleted" (port closed) - wazuh-db never saw the record at all
```

After the fix, same experiment (port 62222):

```
12:55:01  wazuh-db: wdb_parser.c query: dbsync ports INSERTED (local_port 62222) - write confirmed
12:55:01  inventory_harvester builds the Indexer document (same second, now downstream of the write above)
12:55:05  Indexer confirms "result":"created"
```

Also documenting a real deadlock hit during this work, fixed before this PR's first version: an earlier attempt resolved the agent's name/IP/version via `wdb_global_get_agent_info()` from inside `wdb_parse_dbsync()`, which already holds the per-agent node's mutex (`wdb_pool_get_or_create()`). Opening the `global` node from there deadlocked against any other request touching `global` first and an agent node second. `wazuh-db` stopped processing any query under real traffic (new client connections kept being accepted, but none of them got past connection setup), and required `SIGKILL` to stop. Fixed by not resolving agent metadata from `global.db` in that path.

Agent metadata fix, reproduced on a separate manager/agent pair with the Syscollector interval lowered to `1m` so a genuine incremental delta (not the first full scan) triggers the fix. Before the follow-up fix (port 55887):

```
wazuh-db: dbsync ports INSERTED {"local_port":55887,...}                     # no agent_name/agent_ip/agent_version
Indexer document: {"agent":{"id":"001"}, "source":{"port":55887}, ...}        # no name/ip/version
```

After the follow-up fix (port 47733, same reproduction):

```
wazuh-db: dbsync ports INSERTED {"local_port":47733,...,"agent_name":"agent","agent_ip":"192.168.57.22","agent_version":"v4.14.9"}
Indexer document: {"agent":{"id":"001","name":"agent","host":{"ip":"192.168.57.22"},"version":"v4.14.9"}, "source":{"port":47733}, ...}
```

Aggregate check on the same manager, across the whole `sys_ports` table: `wazuh-db` and Indexer both report 10 records, and zero Indexer documents are missing `agent.name`.

Review-fix batch, same manager/agent pair. `network_address` (added a secondary IP to the agent's `eth1`):

```
wazuh-db: dbsync network_address INSERTED {"address":"192.168.57.199",...,"proto":"ipv4",...}
indexer-connector Response: {"result":"created", ...}   # previously discarded, never reached this point
Indexer document: {"network":{"ip":"192.168.57.199","type":"IPv4",...}, "agent":{...full...}}
```

Agent `000` duplicate-publish race (`vulnerability-detection.disable_scan_manager=0`, so the manager's own inventory also goes through `wm_syscollector`'s direct publish), before this fix:

```
UpsertSystemElement::build: {"id":"000_774b98...","agent":{"id":"000","name":"localhost","host":{"ip":"127.0.0.1"},"version":"v4.14.9"}, ...}   # from wm_syscollector, correct
UpsertSystemElement::build: {"id":"000_774b98...","agent":{"id":"000"}, ...}                                                                      # from wazuh-db, blank
indexer-connector Response: [..., {"result":"created", "_version":1}, {"result":"updated", "_version":2}]   # the blank one overwrote the correct one
```

After the fix, same reproduction (new port):

```
UpsertSystemElement::build: {"id":"000_74733c...","agent":{"id":"000","name":"localhost","host":{"ip":"127.0.0.1"},"version":"v4.14.9"}, ...}   # only this one is built
indexer-connector Response: [{"result":"created", "_version":1}]   # single write, no overwrite
```

## Tests

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

Compiled and deployed on a real Wazuh 4.14.9 manager and agent pair (Vagrant VMs), rebuilt from this branch and reinstalled, rather than the full checklist above, which is left for CI/reviewers. Only `Linux` was exercised, with no new compiler warnings introduced.

Unit tests: `test_secure` 50/50 pass; `test_wdb_parser` 169/169 pass, including new regression tests for the publish gate and for the agent-000 empty-name case that caught the duplicate-publish race live. `test_decoder_syscollector` could not be linked in this environment due to a pre-existing "multiple definition of main" toolchain issue affecting the whole `analysisd` unit-test suite (reproduced with an untouched test, `test_decoder_rootcheck`), unrelated to this change; its new marker-parsing tests were reviewed but not run automatically here.

Memory/cluster/retrocompatibility checks and CI-only platforms (Windows, macOS) are still open follow-up items.


src/wazuh_db/wdb_parser.c
```diff
@@ -6604,6 +6604,45 @@ bool process_dbsync_data(wdb_t * wdb, const struct kv * kv_value, const char * o
...
+    // Rebuilds the same envelope the agent originally sent ({"type","operation","data"}),
+    // which is what router_provider_send_fb_json() expects for schema MT_SYS_DELTAS.
+    snprintf(msg, OS_MAXSTR - 1, "{\"type\":\"dbsync_%s\",\"operation\":\"%s\",\"data\":%s}",
+             table_key, operation, data);
+
+    if (router_provider_send_fb_json(router_syscollector_deltas_handle, msg, &ctx, MT_SYS_DELTAS) != 0) {
+        mdebug2("DB(%s) Unable to publish confirmed delta for '%s'.", wdb->id, table_key);
+    }
+
+    os_free(msg);
```
Antoniogm03: **[S2 · blocking · introduced]** `dbsync_network_address` deltas can no longer be published: the `data` republished here is analysisd's *post*-`delta_map_values()` copy, and for that table `protocol_mapping()` (`syscollector.c:171`) has already replaced the numeric `proto` with the string `"ipv4"`/`"ipv6"`, while `syscollector_deltas.fbs:24` declares `proto:long`. `skip_unexpected_fields_in_json` covers the three extra `agent_*` keys but not a known field with the wrong type: `ParseSingleValue` falls through to `ParseEnumFromString()` and fails with `enum values need to be qualified by an enum type`, so `router_provider_send_fb_json()` returns -1 and the delta is dropped with only an `mdebug2`. Before this PR remoted forwarded the raw agent message, where `proto` was still a number.

```suggestion
    // analysisd's delta_map_values() rewrites network_address.proto to "ipv4"/"ipv6",
    // but dbsync_network_address.proto is an integer in syscollector_deltas.fbs.
    char * payload = NULL;
    if (strcmp(table_key, "network_address") == 0) {
        cJSON * payload_json = cJSON_Parse(data);
        cJSON * proto = cJSON_GetObjectItem(payload_json, "proto");
        if (cJSON_IsString(proto)) {
            cJSON_ReplaceItemInObject(
                payload_json, "proto",
                cJSON_CreateNumber(strcmp(proto->valuestring, "ipv4") == 0 ? WDB_NETADDR_IPV4 : 1));
        }
        payload = cJSON_PrintUnformatted(payload_json);
        cJSON_Delete(payload_json);
    }

    char * msg = NULL;
    os_malloc(OS_MAXSTR, msg);
    // Rebuilds the same envelope the agent originally sent ({"type","operation","data"}),
    // which is what router_provider_send_fb_json() expects for schema MT_SYS_DELTAS.
    snprintf(msg, OS_MAXSTR - 1, "{\"type\":\"dbsync_%s\",\"operation\":\"%s\",\"data\":%s}",
             table_key, operation, payload ? payload : data);

    if (router_provider_send_fb_json(router_syscollector_deltas_handle, msg, &ctx, MT_SYS_DELTAS) != 0) {
        mdebug2("DB(%s) Unable to publish confirmed delta for '%s'.", wdb->id, table_key);
    }

    os_free(payload);
    os_free(msg);
```

src/remoted/secure.c
```diff
@@ -42,7 +42,8 @@ OSHash *remoted_agents_state;
 
 extern remoted_state_t remoted_state;
 ROUTER_PROVIDER_HANDLE router_rsync_handle = NULL;
-ROUTER_PROVIDER_HANDLE router_syscollector_handle = NULL;
+// Syscollector deltas are now published by wazuh-db only after the write is confirmed
+// (see issue #39329); remoted no longer forwards them to the Indexer pipeline directly.
```
Antoniogm03: **[S2 · blocking · introduced]** Removing the `router_syscollector_handle` definition breaks the compilation of the remoted unit tests: `src/unit_tests/remoted/test_secure.c:19` does `#include "../../remoted/secure.c"` and assigns that identifier at lines 73, 87, 129, 2510 and 2609, so the build fails with `'router_syscollector_handle' undeclared`. `test_router_message_forward_syscollector_no_handle` (2508) and `..._with_handle` (2521) also still assert the forwarding behaviour this PR removed. Please update that file in this PR and add a case covering the new early return.

src/analysisd/decoders/syscollector.c
```diff
@@ -2283,6 +2311,13 @@ static int decode_dbsync(Eventinfo * lf,   /* Event information */
                     delta_map_values(type, data_object);                            /* Map field's values if applies */
                     char * operation = operation_object->valuestring;               /* Operation is the operation to be
                                                                                        performed in the table. */
+
+                    /* Passed through to wazuh-db only to publish to the Indexer (issue #39329);
+                     * the table upsert only binds fields it recognizes, so these are ignored there. */
+                    cJSON_AddStringToObject(data_object, "agent_name", agent_name);
+                    cJSON_AddStringToObject(data_object, "agent_ip", agent_ip);
+                    cJSON_AddStringToObject(data_object, "agent_version", agent_version);
```
Antoniogm03: **[S2 · blocking · introduced]** Injecting these three keys changes the exact wazuh-db query that ~20 decoder tests assert byte-for-byte (`expect_string(__wrap_wdbc_query_ex, query, query)` at `test_decoder_syscollector.c:2157, 2192, 2242, 2276, 2313, 2340, 2367, 2396, 2425, 2463`, …); every expected string ends at `"item_id":"…"}` and the decoder now appends `,"agent_name":"","agent_ip":"","agent_version":""`. None of them were updated, and nothing new covers the marker parse, the wazuh-db republish, or the write-before-publish ordering that is the point of the PR.

src/wazuh_db/wdb_parser.c
```diff
@@ -6638,6 +6677,28 @@ int wdb_parse_dbsync(wdb_t * wdb, char * input, char * output) {
...
+                    cJSON * j_ip = cJSON_GetObjectItem(agent_fields, "agent_ip");
+                    cJSON * j_version = cJSON_GetObjectItem(agent_fields, "agent_version");
+
+                    agent_name = cJSON_IsString(j_name) ? j_name->valuestring : "";
+                    agent_ip = cJSON_IsString(j_ip) ? j_ip->valuestring : "";
+                    agent_version = cJSON_IsString(j_version) ? j_version->valuestring : "";
+                }
+
+                wdb_publish_confirmed_delta(wdb, table_key, operation, data, agent_name, agent_ip, agent_version);
+                cJSON_Delete(agent_fields);
```
Antoniogm03: **[S2 · blocking · introduced]** Deltas that never went through remoted are published with a blank agent identity. The manager's own syscollector sends its deltas straight to analysisd (`wm_syscollector.c:102`, `SYSCOLLECTOR_MQ`), so they carry no marker and analysisd leaves the three fields empty: under the default `vulnerability-detection.disable_scan_manager=1` the Indexer starts receiving agent `000` documents with an empty `agent.name` and no rsync stream to reconcile them, and with `disable_scan_manager=0` `wm_sys_send_diff_message()` already publishes the same delta as `localhost`/`127.0.0.1`, so the later blank copy from wazuh-db wins.

```suggestion
                cJSON * agent_fields = cJSON_ParseWithOpts(data, NULL, true);

                if (NULL != agent_fields) {
                    cJSON * j_name = cJSON_GetObjectItem(agent_fields, "agent_name");
                    cJSON * j_ip = cJSON_GetObjectItem(agent_fields, "agent_ip");
                    cJSON * j_version = cJSON_GetObjectItem(agent_fields, "agent_version");

                    /* Only deltas relayed by remoted carry the agent context; the ones the
                     * manager injects locally are published by wazuh-modulesd itself. */
                    if (cJSON_IsString(j_name)) {
                        wdb_publish_confirmed_delta(wdb, table_key, operation, data, j_name->valuestring,
                                                    cJSON_IsString(j_ip) ? j_ip->valuestring : "",
                                                    cJSON_IsString(j_version) ? j_version->valuestring : "");
                    }
                }

                cJSON_Delete(agent_fields);
```

src/remoted/secure.c
```diff
@@ -951,18 +949,42 @@ STATIC void HandleSecureMessage(const message_t *message, w_indexed_queue_t * co
...
+        }
+
+        size_t enriched_size = SYSCOLLECTOR_HEADER_SIZE + strlen(agent_name) + strlen(agent_ip) +
+                                strlen(agent_version) + strlen(tmp_msg + SYSCOLLECTOR_HEADER_SIZE) + 4;
+        os_malloc(enriched_size, analysisd_msg);
+        snprintf(analysisd_msg, enriched_size, "%.*s%s\x01%s\x01%s\x01%s",
+                 SYSCOLLECTOR_HEADER_SIZE, tmp_msg, agent_name, agent_ip, agent_version,
+                 tmp_msg + SYSCOLLECTOR_HEADER_SIZE);
+        free_analysisd_msg = true;
+    }
```
Antoniogm03: **[S2 · blocking · introduced]** The comment above states that `\x01` cannot appear in these values, but `agent_version` is the one value the agent itself supplies: it comes from `agent_data_hash`, filled with `strdup(version->valuestring)` straight out of the agent's HC_STARTUP JSON (`manager.c:297`), and cJSON accepts both a raw `0x01` and a `\u0001` escape inside a string. With an extra separator in the version, the decoder's `sep3` (`syscollector.c:409`) lands early, `lf->log` points into the middle of the version and `cJSON_ParseWithOpts` fails, so that agent's real-time deltas stop reaching wazuh-db *and* the Indexer with only an `mdebug1`.

```suggestion
    if (strncmp(tmp_msg, SYSCOLLECTOR_HEADER, SYSCOLLECTOR_HEADER_SIZE) == 0) {
        const char * agent_version = (const char *) OSHash_Get_ex(agent_data_hash, agentid_str);
        // The version is agent-reported: a '\x01' in it would shift the separators below.
        if (!agent_version || strchr(agent_version, '\x01')) {
            agent_version = "";
        }

        size_t enriched_size = SYSCOLLECTOR_HEADER_SIZE + strlen(agent_name) + strlen(agent_ip) +
                                strlen(agent_version) + strlen(tmp_msg + SYSCOLLECTOR_HEADER_SIZE) + 4;
        os_malloc(enriched_size, analysisd_msg);
        snprintf(analysisd_msg, enriched_size, "%.*s%s\x01%s\x01%s\x01%s",
                 SYSCOLLECTOR_HEADER_SIZE, tmp_msg, agent_name, agent_ip, agent_version,
                 tmp_msg + SYSCOLLECTOR_HEADER_SIZE);
        free_analysisd_msg = true;
    }
```

src/wazuh_db/wdb_parser.c
```diff
@@ -6638,6 +6677,28 @@ int wdb_parse_dbsync(wdb_t * wdb, char * input, char * output) {
     while (NULL != head) {
         if (strncmp(head->current.key, table_key, OS_SIZE_256 - 1) == 0) {
             ret_val = process_dbsync_data(wdb, &head->current, operation, data) ? OS_SUCCESS : OS_INVALID;
+            if (OS_SUCCESS == ret_val) {
```
Antoniogm03: **[S2 · aggravated]** For `DELETED` the publish is not actually gated on a confirmed write: `process_dbsync_data()` discards `wdb_delete_dbsync()`'s return value and sets `ret_val = true` unconditionally (line 6593-6595), so a delete that failed to bind or to step still reaches `wdb_publish_confirmed_delta()` and removes the document from the Indexer while wazuh-db still holds the row. The unconditional `true` predates this PR, but it only becomes a divergence now that it gates the Indexer. Propagating the boolean out of `wdb_delete_dbsync()` closes it.

src/wazuh_db/wdb_parser.c
```diff
@@ -6604,6 +6604,45 @@ bool process_dbsync_data(wdb_t * wdb, const struct kv * kv_value, const char * o
...
+
+    agent_ctx ctx = {
+        .agent_id = wdb->id,
+        .agent_name = agent_name,
+        .agent_ip = agent_ip,
+        .agent_version = agent_version,
+    };
+
+    char * msg = NULL;
+    os_malloc(OS_MAXSTR, msg);
```
Antoniogm03: **[S3 · introduced]** Fixed 64 KiB allocation per delta, and the `snprintf` below silently truncates: `data` can approach `OS_MAXSTR` and the envelope adds ~45 bytes plus the table name, so the result is a half-written JSON object that simdjson rejects in `SchemaAdapter` — the delta is lost with only an `mdebug2`, and remoted no longer forwards it either. The needed size is known (`strlen(data)` plus the header), and `snprintf`'s return value is not checked.

src/wazuh_db/wdb_parser.c
```diff
@@ -6604,6 +6604,45 @@ bool process_dbsync_data(wdb_t * wdb, const struct kv * kv_value, const char * o
...
+    // Rebuilds the same envelope the agent originally sent ({"type","operation","data"}),
+    // which is what router_provider_send_fb_json() expects for schema MT_SYS_DELTAS.
+    snprintf(msg, OS_MAXSTR - 1, "{\"type\":\"dbsync_%s\",\"operation\":\"%s\",\"data\":%s}",
+             table_key, operation, data);
+
+    if (router_provider_send_fb_json(router_syscollector_deltas_handle, msg, &ctx, MT_SYS_DELTAS) != 0) {
```
Antoniogm03: **[S3 · introduced]** `data` is now parsed three times per delta — once in `process_dbsync_data()`, again at line 6684 just to read three strings, and a third time inside `router_provider_send_fb_json()` (simdjson plus the FlatBuffers parser) — all with the per-agent node mutex held, so every other query for that agent queues behind it. The socket write itself is fine (`SocketClient::send` is non-blocking and buffers the remainder), but the parsing is not. Extracting the three fields inside `process_dbsync_data()`, which already builds the full cJSON tree and throws it away, removes one whole parse from the hot path.

src/wazuh_db/main.c
```diff
@@ -221,6 +223,10 @@ int main(int argc, char ** argv)
         mdebug2("Failed to create router handle for 'wdb-inventory-events'.");
     }
 
+    if (router_syscollector_deltas_handle = router_provider_create(WDB_SYSCOLLECTOR_DELTAS_TOPIC, false), !router_syscollector_deltas_handle) {
+        mdebug2("Failed to create router handle for 'deltas-syscollector'.");
+    }
+
```
Antoniogm03: **[S3 · introduced]** This publisher is not covered by `remoted.router_forwarding_disabled` (`etc/internal_options.conf:250`). Setting it to `1` currently stops syscollector deltas from reaching the Indexer; after this PR remoted's early return makes no difference because wazuh-db publishes unconditionally, so the switch silently loses half of what it used to control. Worth either honouring an equivalent switch here or documenting the narrowed scope.

src/remoted/secure.c
```diff
@@ -951,18 +949,42 @@ STATIC void HandleSecureMessage(const message_t *message, w_indexed_queue_t * co
...
+        if (!agent_version) {
+            agent_version = "";
+        }
+
+        size_t enriched_size = SYSCOLLECTOR_HEADER_SIZE + strlen(agent_name) + strlen(agent_ip) +
+                                strlen(agent_version) + strlen(tmp_msg + SYSCOLLECTOR_HEADER_SIZE) + 4;
+        os_malloc(enriched_size, analysisd_msg);
+        snprintf(analysisd_msg, enriched_size, "%.*s%s\x01%s\x01%s\x01%s",
+                 SYSCOLLECTOR_HEADER_SIZE, tmp_msg, agent_name, agent_ip, agent_version,
+                 tmp_msg + SYSCOLLECTOR_HEADER_SIZE);
```
Antoniogm03: **[S3 · introduced]** The enriched message is not re-checked against `OS_MAXSTR`. `SendMSGAction()` formats into `char tmpstr[OS_MAXSTR + 1]` with `snprintf(tmpstr, OS_MAXSTR, "%c:%s->%s", …)` (`src/shared/mq_op.c`), so a delta that previously fit exactly now loses `strlen(agent_name) + strlen(agent_ip) + strlen(agent_version) + 3` bytes off the tail and arrives at analysisd as unparseable JSON — lost on both the wazuh-db and the Indexer paths, with nothing in the log to distinguish it from a malformed event.

src/remoted/secure.c
```diff
@@ -951,18 +949,42 @@ STATIC void HandleSecureMessage(const message_t *message, w_indexed_queue_t * co
...
+     * message so wazuh-db can publish them to the Indexer without any lookup of its
+     * own (see issue #39329). Encoded as "<name>\x01<ip>\x01<version>\x01" right after
+     * the "d:syscollector:" header; \x01 can't appear in these values, and the JSON
+     * payload always starts with '{', so a missing/malformed marker is unambiguous.
+     */
+    char *analysisd_msg = tmp_msg;
+    bool free_analysisd_msg = false;
+
+    if (strncmp(tmp_msg, SYSCOLLECTOR_HEADER, SYSCOLLECTOR_HEADER_SIZE) == 0) {
+        const char * agent_version = (const char *) OSHash_Get_ex(agent_data_hash, agentid_str);
```
Antoniogm03: **[S3 · introduced]** `OSHash_Get_ex()` releases the hash lock before returning, so this raw pointer can be freed underneath us: an HC_SHUTDOWN for the same agent on another worker thread runs `OSHash_Delete_ex(agent_data_hash, key->id)` followed by `os_free(deleted)` (`manager.c:322`) while `strlen`/`snprintf` are still reading the version string. `OSHash_Get_ex_dup()` exists for exactly this and copies under the lock. The pre-existing instance at line 1076 has the same defect, but this PR duplicates it instead of fixing it.

src/analysisd/decoders/syscollector.c
Antoniogm03: **[S2 · blocking · introduced]** The agent can overwrite its own published identity, because the context travels inside `data`, which it controls end to end. `cJSON_AddStringToObject()` appends at the end of the object (`cJSON.c:2029` -> `add_item_to_array`), while `wdb_parser.c:6698-6700` reads it back with `cJSON_GetObjectItem()`, which returns the **first** match and compares **case-insensitively** (`cJSON.c:1898-1932`) - so a delta carrying `"data":{"agent_name":"other-host",...}` wins over the authentic value and reaches the Indexer as `agent.name`; `"agent_name":""` or a non-string instead makes `has_agent_context` false and silently drops every delta from that agent. Before this PR these values came only from remoted's keystore and were never agent-controlled.

```suggestion
                    /* Passed through to wazuh-db only to publish to the Indexer (issue #39329);
                     * the table upsert only binds fields it recognizes, so these are ignored there.
                     * The payload is agent-controlled and cJSON keeps duplicate keys while
                     * cJSON_GetObjectItem() returns the first (case-insensitive) match, so any
                     * copy the agent sent must be dropped or it would win on the read side. */
                    static const char * const agent_ctx_keys[] = {"agent_name", "agent_ip", "agent_version"};
                    const char * const agent_ctx_values[] = {agent_name, agent_ip, agent_version};

                    for (size_t i = 0; i < sizeof(agent_ctx_keys) / sizeof(*agent_ctx_keys); i++) {
                        while (NULL != cJSON_GetObjectItem(data_object, agent_ctx_keys[i])) {
                            cJSON_DeleteItemFromObject(data_object, agent_ctx_keys[i]);
                        }
                        cJSON_AddStringToObject(data_object, agent_ctx_keys[i], agent_ctx_values[i]);
                    }
```

src/wazuh_db/wdb_parser.c
```diff
@@ -6604,6 +6603,45 @@ bool process_dbsync_data(wdb_t * wdb, const struct kv * kv_value, const char * o
...
+        .agent_ip = agent_ip,
+        .agent_version = agent_version,
+    };
+
+    char * msg = NULL;
+    os_malloc(OS_MAXSTR, msg);
+    // Rebuilds the same envelope the agent originally sent ({"type","operation","data"}),
+    // which is what router_provider_send_fb_json() expects for schema MT_SYS_DELTAS.
+    snprintf(msg, OS_MAXSTR - 1, "{\"type\":\"dbsync_%s\",\"operation\":\"%s\",\"data\":%s}",
+             table_key, operation, data);
```
Antoniogm03: **[S3 · non-blocking · introduced]** Still open from iteration 1: the envelope is built into a fixed 64 KiB buffer and truncated silently. `data` arrives inside a query already bounded by `OS_MAXSTR`, so it can sit within a few dozen bytes of the limit; the envelope adds about 40 more and `snprintf(msg, OS_MAXSTR - 1, ...)` cuts it mid-JSON, after which `router_provider_send_fb_json()` fails to parse and the only trace is an `mdebug2` - the record exists locally but never in the Indexer, which is the divergence this PR is meant to prevent. The required size is known from `strlen(data)`.

```suggestion
    char * msg = NULL;
    // Rebuilds the envelope router_provider_send_fb_json() expects for MT_SYS_DELTAS:
    // {"type","operation","data"}. 40 bytes covers the literal keys, braces and quotes.
    const size_t msg_size = strlen(data) + strlen(table_key) + strlen(operation) + 40;
    os_malloc(msg_size, msg);
    if ((size_t) snprintf(msg, msg_size, "{\"type\":\"dbsync_%s\",\"operation\":\"%s\",\"data\":%s}",
                          table_key, operation, data) >= msg_size) {
        mdebug2("DB(%s) Confirmed delta for '%s' does not fit the publish buffer.", wdb->id, table_key);
        os_free(msg);
        return;
    }
```

src/wazuh_db/wdb_parser.c
```diff
@@ -6638,6 +6676,62 @@ int wdb_parse_dbsync(wdb_t * wdb, char * input, char * output) {
     while (NULL != head) {
         if (strncmp(head->current.key, table_key, OS_SIZE_256 - 1) == 0) {
             ret_val = process_dbsync_data(wdb, &head->current, operation, data) ? OS_SUCCESS : OS_INVALID;
+            if (OS_SUCCESS == ret_val) {
+                /* agent_name/agent_ip/agent_version were embedded into `data` by analysisd's
+                 * decode_dbsync() (see issue #39329); process_dbsync_data() above already
+                 * ignored them since the table upsert only binds fields it recognizes. */
+                cJSON * data_json = cJSON_ParseWithOpts(data, NULL, true);
```
Antoniogm03: **[S3 · non-blocking · introduced]** Still open from iteration 1, and now one step worse: `data` is parsed here a second time purely to read three strings, after `process_dbsync_data()` already parsed the identical string at line 6588, and `network_address` adds a fourth serialization with `cJSON_PrintUnformatted()` at line 6723. All of it runs with the agent node's mutex held, so every other query for that agent waits on it. Parse once in `process_dbsync_data()` and hand the cJSON object back instead.

src/wazuh_db/wdb_parser.c
```diff
@@ -6638,6 +6676,62 @@ int wdb_parse_dbsync(wdb_t * wdb, char * input, char * output) {
...
+                     * syscollector_deltas.fbs. Restore the numeric value for the Indexer publish
+                     * only; the SQL write already used, and needed, the string form. */
+                    if (has_agent_context && strcmp(table_key, "network_address") == 0) {
+                        cJSON * j_proto = cJSON_GetObjectItem(data_json, "proto");
+                        if (cJSON_IsString(j_proto)) {
+                            cJSON_ReplaceItemInObject(data_json, "proto", cJSON_CreateNumber(
+                                strcmp(j_proto->valuestring, "ipv4") == 0 ? WDB_NETADDR_IPV4 : WDB_NETADDR_IPV6));
+                            corrected_data = cJSON_PrintUnformatted(data_json);
+                        }
+                    }
```
Antoniogm03: **[S3 · non-blocking · introduced]** Undoing analysisd's transform here is a symptom fix with no safety net: any new entry in `get_mapping_list()` will mutate the `data` wazuh-db republishes without a matching inverse at this spot, and the field will reach the Indexer with the wrong type leaving only an `mdebug2`. It is also lossy - `protocol_mapping()` maps any non-zero `proto` to `"ipv6"` and this maps it back to 1, so `proto: 7` is published as `proto: 1`. The root fix is for `delta_map_values()` to work on a copy used only for the SQL bind, leaving the published object with the agent's original values.

src/wazuh_db/main.c
Antoniogm03: **[S3 · non-blocking · introduced]** Still open from iteration 1: `remoted.router_forwarding_disabled` no longer stops syscollector deltas. The switch is checked at `secure.c:1005`, i.e. only on remoted's forwarding path, and this new provider publishes on every confirmed dbsync write with no equivalent guard - so an operator who sets it to `1` still gets the full delta stream in the Indexer and in vulnerability_scanner. Either honour an equivalent switch here or document that its scope has narrowed.

src/remoted/secure.c
```diff
@@ -951,18 +949,46 @@ STATIC void HandleSecureMessage(const message_t *message, w_indexed_queue_t * co
...
+     * the "d:syscollector:" header. agent_name/agent_ip never contain '\x01' (name comes
+     * from client.keys, ip is either that or a formatted socket address); agent_version
+     * is agent-reported and unvalidated, so it's explicitly checked below. The JSON
+     * payload always starts with '{', so a missing/malformed marker is unambiguous.
+     */
+    char *analysisd_msg = tmp_msg;
+    bool free_analysisd_msg = false;
+
+    if (strncmp(tmp_msg, SYSCOLLECTOR_HEADER, SYSCOLLECTOR_HEADER_SIZE) == 0) {
+        const char * agent_version = (const char *) OSHash_Get_ex(agent_data_hash, agentid_str);
```
Antoniogm03: **[S3 · non-blocking · introduced]** Still open from iteration 1: `OSHash_Get_ex()` releases the lock before returning the stored pointer, and the `strchr`/`strlen`/`snprintf` below read it afterwards. Another worker handling HC_SHUTDOWN for the same agent runs `OSHash_Delete_ex(agent_data_hash, key->id)` followed by `os_free(deleted)` (`manager.c:322-323`), which makes this a use-after-free; the version sanitization added in this iteration widened the window by two more reads. Use `OSHash_Get_ex_dup(agent_data_hash, agentid_str, (void *(*)(void *)) strdup)` and free the copy after building the message.

src/remoted/secure.c
```diff
@@ -951,18 +949,46 @@ STATIC void HandleSecureMessage(const message_t *message, w_indexed_queue_t * co
...
+        // The version is agent-reported (HC_STARTUP, unvalidated): a '\x01' in it would
+        // desync the separators below and corrupt every subsequent delta from this agent.
+        if (!agent_version || strchr(agent_version, '\x01')) {
+            agent_version = "";
+        }
+
+        size_t enriched_size = SYSCOLLECTOR_HEADER_SIZE + strlen(agent_name) + strlen(agent_ip) +
+                                strlen(agent_version) + strlen(tmp_msg + SYSCOLLECTOR_HEADER_SIZE) + 4;
+        os_malloc(enriched_size, analysisd_msg);
+        snprintf(analysisd_msg, enriched_size, "%.*s%s\x01%s\x01%s\x01%s",
```
Antoniogm03: **[S3 · non-blocking · introduced]** Still open from iteration 1: the marker can push the message past `OS_MAXSTR`. `SendMSGAction()` formats into `char tmpstr[OS_MAXSTR + 1]` with `snprintf(tmpstr, OS_MAXSTR, ...)` (`shared/mq_op.c:143`), and the enrichment adds `strlen(agent_name)+strlen(agent_ip)+strlen(agent_version)+3` bytes with no size check. A delta that previously fit by a narrow margin now arrives truncated, analysisd logs "Error parsing JSON event." and drops it - and since remoted no longer forwards independently, the delta is lost everywhere, not just in the DB. Check the resulting size first and send the unenriched message if it does not fit.

src/unit_tests/remoted/test_secure.c
```diff
@@ -2505,41 +2501,16 @@ void test_router_message_forward_fim_dbsync_file(void** state) {
...
-
-    router_message_forward(msg, agent_id, agent_ip, agent_name);
-}
-
-void test_router_message_forward_syscollector_with_handle(void** state) {
+void test_router_message_forward_syscollector_dropped(void** state) {
+    // wazuh-db publishes this delta itself after confirming the write (issue #39329);
+    // remoted must not forward it to the router or log anything about it. No
+    // expectations are set on __wrap_router_provider_send_fb_json/__wrap__mdebug2,
+    // so cmocka fails the test if either gets called.
```
Antoniogm03: **[S3 · non-blocking · introduced]** The enrichment branch added to `HandleSecureMessage` has no coverage at all: no test in this file feeds a `d:syscollector:` payload, so neither the `enriched_size` arithmetic nor the `strchr(agent_version, '\x01')` guard is ever exercised - an off-by-one there would be a heap overflow CI would not catch. Deleting `test_router_message_forward_too_large` also leaves the surviving `(msg_size + message_header_size) < OS_MAXSTR` guard at `secure.c:1075` untested; re-point it at the `5:dbsync:syscollector` path instead. In `test_wdb_parser.c`, `expect_any(__wrap_router_provider_send_fb_json, msg)` means the published envelope is never asserted, and the `network_address` `proto` restore has no test of its own.

src/wazuh_db/wdb_parser.c
Antoniogm03: **[S3 · non-blocking · introduced]** Still open from iteration 1: the envelope is built into a fixed 64 KiB buffer and truncated silently. `data` arrives inside a query already bounded by `OS_MAXSTR`, so it can sit within a few dozen bytes of the limit; the envelope adds about 40 more and `snprintf(msg, OS_MAXSTR - 1, ...)` cuts it mid-JSON, after which `router_provider_send_fb_json()` fails to parse and the only trace is an `mdebug2` - the record exists locally but never in the Indexer, which is the divergence this PR is meant to prevent. The required size is known from `strlen(data)`.

```suggestion
    char * msg = NULL;
    // Rebuilds the envelope router_provider_send_fb_json() expects for MT_SYS_DELTAS:
    // {"type","operation","data"}. 40 bytes covers the literal keys, braces and quotes.
    const size_t msg_size = strlen(data) + strlen(table_key) + strlen(operation) + 40;
    os_malloc(msg_size, msg);
    if ((size_t) snprintf(msg, msg_size, "{\"type\":\"dbsync_%s\",\"operation\":\"%s\",\"data\":%s}",
                          table_key, operation, data) >= msg_size) {
        mdebug2("DB(%s) Confirmed delta for '%s' does not fit the publish buffer.", wdb->id, table_key);
        os_free(msg);
        return;
    }
```

src/wazuh_db/wdb_parser.c
Antoniogm03: **[S3 · non-blocking · introduced]** Still open from iteration 1, and now one step worse: `data` is parsed here a second time purely to read three strings, after `process_dbsync_data()` already parsed the identical string at line 6588, and `network_address` adds a fourth serialization with `cJSON_PrintUnformatted()` at line 6723. All of it runs with the agent node's mutex held, so every other query for that agent waits on it. Parse once in `process_dbsync_data()` and hand the cJSON object back instead.

src/wazuh_db/wdb_parser.c
Antoniogm03: **[S3 · non-blocking · introduced]** Undoing analysisd's transform here is a symptom fix with no safety net: any new entry in `get_mapping_list()` will mutate the `data` wazuh-db republishes without a matching inverse at this spot, and the field will reach the Indexer with the wrong type leaving only an `mdebug2`. It is also lossy - `protocol_mapping()` maps any non-zero `proto` to `"ipv6"` and this maps it back to 1, so `proto: 7` is published as `proto: 1`. The root fix is for `delta_map_values()` to work on a copy used only for the SQL bind, leaving the published object with the agent's original values.

src/wazuh_db/main.c
Antoniogm03: **[S3 · non-blocking · introduced]** Still open from iteration 1: `remoted.router_forwarding_disabled` no longer stops syscollector deltas. The switch is checked at `secure.c:1005`, i.e. only on remoted's forwarding path, and this new provider publishes on every confirmed dbsync write with no equivalent guard - so an operator who sets it to `1` still gets the full delta stream in the Indexer and in vulnerability_scanner. Either honour an equivalent switch here or document that its scope has narrowed.

src/remoted/secure.c
Antoniogm03: **[S3 · non-blocking · introduced]** Still open from iteration 1: `OSHash_Get_ex()` releases the lock before returning the stored pointer, and the `strchr`/`strlen`/`snprintf` below read it afterwards. Another worker handling HC_SHUTDOWN for the same agent runs `OSHash_Delete_ex(agent_data_hash, key->id)` followed by `os_free(deleted)` (`manager.c:322-323`), which makes this a use-after-free; the version sanitization added in this iteration widened the window by two more reads. Use `OSHash_Get_ex_dup(agent_data_hash, agentid_str, (void *(*)(void *)) strdup)` and free the copy after building the message.

src/remoted/secure.c
Antoniogm03: **[S3 · non-blocking · introduced]** Still open from iteration 1: the marker can push the message past `OS_MAXSTR`. `SendMSGAction()` formats into `char tmpstr[OS_MAXSTR + 1]` with `snprintf(tmpstr, OS_MAXSTR, ...)` (`shared/mq_op.c:143`), and the enrichment adds `strlen(agent_name)+strlen(agent_ip)+strlen(agent_version)+3` bytes with no size check. A delta that previously fit by a narrow margin now arrives truncated, analysisd logs "Error parsing JSON event." and drops it - and since remoted no longer forwards independently, the delta is lost everywhere, not just in the DB. Check the resulting size first and send the unenriched message if it does not fit.

src/unit_tests/remoted/test_secure.c
Antoniogm03: **[S3 · non-blocking · introduced]** The enrichment branch added to `HandleSecureMessage` has no coverage at all: no test in this file feeds a `d:syscollector:` payload, so neither the `enriched_size` arithmetic nor the `strchr(agent_version, '\x01')` guard is ever exercised - an off-by-one there would be a heap overflow CI would not catch. Deleting `test_router_message_forward_too_large` also leaves the surviving `(msg_size + message_header_size) < OS_MAXSTR` guard at `secure.c:1075` untested; re-point it at the `5:dbsync:syscollector` path instead. In `test_wdb_parser.c`, `expect_any(__wrap_router_provider_send_fb_json, msg)` means the published envelope is never asserted, and the `network_address` `proto` restore has no test of its own.

Antoniogm03:
# Root cause: why the Indexer and wazuh-db briefly disagree

## Summary

Reproduced on a fresh 4.14.8 (pre-fix) QA environment (5 isolated runs of `test_browser_extensions_first_scan` + `test_browser_extensions_regular_scan`; reproduced in 2/5 — it's data-dependent, see below).

Two things are going on, and it's important not to conflate them:

1. **A real, small architectural lag (~6s):** `wazuh-remoted` notifies the Indexer and wazuh-db through two independent paths with no synchronization between them. In every case measured here, the Indexer consistently applied a change ~6 seconds *after* wazuh-db did — never before. This is the actual root cause, and it's already fixed on `fix/39329-syscollector-indexer-race`.
2. **A much bigger, test-specific artifact (~80s):** the E2E test itself reads the Indexer and wazuh-db at two different moments, ~80 seconds apart. If a second, unrelated data change happens to land inside that ~80s gap, the two reads end up describing two different points in time — which is what actually produces the "41 vs 40" the test reports, on top of the 6s lag above.

The two effects together look exactly like the symptom in the original report, but the mechanism is more specific than "the Indexer is slow": it's a real (small) propagation lag, made visible by an unrelated (large) gap in how the test itself samples the two data sources.

---

## 1. Root cause: two independent paths, no ordering guarantee

When `wazuh-remoted` receives a syscollector delta from an agent (`src/remoted/secure.c`, `save_controlmsg()`), it dispatches the **same message down two separate paths, back-to-back in the same thread, with no synchronization between them**:

1. `SendMSG(logr.m_queue, tmp_msg, ...)` → the local `/queue/ossec/queue` (`SECURE_MQ`), consumed asynchronously by `analysisd`, which hands the dbsync message to `wazuh-db` → `process_dbsync_data()` commits it to the agent's local SQLite database.
2. `router_message_forward(...)` → the in-process `deltas-syscollector` router topic, consumed by the `inventory-harvester` submodule running inside `wazuh-modulesd` on the same manager, whose `indexer-connector` pushes the document to the Wazuh Indexer over the network (bulk API).

Both dispatches fire essentially simultaneously from remoted's point of view — it never waits for either side to confirm anything before moving on. Downstream, the two paths are fully independent: wazuh-db's write is a fast, local, synchronous SQLite commit; the Indexer path additionally crosses the network and goes through OpenSearch's own bulk-indexing pipeline. Nothing guarantees which side finishes (and becomes queryable) first, or by how much.

## 2. Measured evidence: a consistent ~6 second lag

Enabled `remoted.debug=2`, `analysisd.debug=2`, `wazuh_db.debug=2` and `wazuh_modules.debug=2` on the manager and reproduced again. The record that flaps is the exact same one described in the original issue — `{CFBFAE00-17A6-11D0-99CB-00C04FD64497}` / "Microsoft Url Search Hook" / `ieframe.dll` — a specific IE search hook that genuinely appears and disappears between consecutive 2-minute scans on this Windows image (real OS/IE behavior, not a Wazuh bug).

| Time (UTC) | Component | Event |
| --- | --- | --- |
| 14:58:56 | wazuh-db | `dbsync browser_extensions INSERTED` — local commit of the addition |
| 14:59:02 | indexer-connector | `"result":"created"` in the Indexer — **+6s** after wazuh-db |
| 15:00:58 | wazuh-db | `dbsync browser_extensions DELETED` — local commit of the removal |
| 15:01:04 | indexer-connector | `"result":"deleted"` in the Indexer — **+6s** again |
| 15:01:47 | wazuh-db | the test queries wazuh-db(agent2): already sees 40 (removal already applied) |

The Indexer confirms each change ~6 seconds after wazuh-db, both times. That consistent, small, unbounded gap is the real mechanism behind the "Indexer vs wazuh-db" symptom — on its own, it would only ever make the Indexer look *behind*, never ahead.

## 3. Why the test actually reports 41 vs 40

This is the part that's easy to misread: a lag that only ever puts the Indexer *behind* can't by itself explain the Indexer showing **more** records (41) than wazuh-db (40). The missing piece is *when* the test reads each source.

`test_browser_extensions_regular_scan` does two separate reads, not one:

1. It reads the **Indexer** for all 6 agents first — fast, plain HTTPS, done in a couple of seconds.
2. It then reads **wazuh-db** for all 6 agents **one at a time over SSH** (a remote `wdb-query.py` call per agent) — slow, ~80 seconds total for 6 agents.

In this run, two real changes happened back to back: the hook was inserted (14:58:56 wazuh-db / 14:59:02 Indexer), and one scan cycle later it was deleted again (15:00:58 wazuh-db / 15:01:04 Indexer). The test's Indexer read happened early, before the delete had even reached the Indexer (before 15:01:04) — so it still shows **41**, the pre-delete state. The test's wazuh-db read for `agent2` happened much later, near the end of the slow ~80s SSH loop (~15:01:47) — by then the delete had been committed to wazuh-db almost a minute earlier, so it shows **40**, the post-delete state.



Neither side is "wrong" at the moment it was read: the Indexer's 41 was correct as of ~15:00:15, and wazuh-db's 40 was correct as of ~15:01:47. The mismatch exists only because the test's own two reads are ~80s apart, and a real state change (the delete) landed inside that gap.

**In one sentence:** there's a small, real, architectural lag (~6s, Indexer always behind wazuh-db — the actual root cause) and, layered on top of it, a much bigger artifact from the test's own read cadence (~80s apart) that can turn any change landing in that window into an apparent — but momentary — divergence.

## 4. Recovery: it resolves itself, with no intervention

Independently of the test's own query cadence, directly polling both sides right after a reproduced divergence showed them reconverging on their own, within under two minutes:

| t since scan detection (14:29:23) | Time (UTC) | Indexer | wazuh-db | Match |
| --- | --- | --- | --- | --- |
| ~t+92s | 12:30:55 | 41 | 41 | Yes |
| ~t+170s | 12:32:13 | 41 | 40 | **No** |
| ~t+249s | 12:33:31 | 40 | 40 | Yes |
| ~t+328s | 12:34:51 | 40 | 40 | Yes |

No data is lost — it's a lag, not a corruption. Each side simply finishes applying the same event on its own independent schedule.


The fix is not necessary.

PR #39418, opened 2026-09-17
Report an unavailable CPU speed as unavailable on Apple Silicon macOS

### LINKED ISSUE #39185: macOS agent reports no CPU speed because the Apple Silicon implementation reads a power-management property from the IORegistry and throws instead of degrading when it is unavailable
## Description

`host.cpu.speed` is null in the macOS agent's hardware inventory, while every other field in the same document is populated:

```json
{"host.cpu.cores": 2,
 "host.cpu.name": "Apple M1 (Virtual)",
 "host.cpu.speed": null,
 "host.memory.free": 67223552,
 "host.memory.total": 4294967296,
 "host.memory.usage": 0.9843482971191406,
 "host.memory.used": 4227743744,
 "host.serial_number": "ZRFGPL2J3W"}
```

For comparison, the x86 Linux agent in the same environment fills it in:

```json
{"host.cpu.cores": 2,
 "host.cpu.name": "Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz",
 "host.cpu.speed": 2501, ...}
```

macOS has two implementations of `getMhz()` selected at build time. The x86 one is four lines and asks the kernel directly:

```cpp
int ret{osPrimitives->sysctlbyname("hw.cpufrequency", &cpuHz, &len, nullptr, 0)};
```

The Apple Silicon one, in `src/data_provider/src/hardware/architectureDependantARMMac.cpp`, cannot use that sysctl because Apple does not publish it on ARM, so it walks the IORegistry instead: it matches `AppleARMIODevice`, iterates until it finds the entry named `pmgr`, reads that entry's properties, and pulls a specific key out of them:

```cpp
// voltage-states5-sram contains the performance cores available frequencies
CFStringRef cfkey = osPrimitives->CFStringCreateWithCString(kCFAllocatorDefault, "voltage-states5-sram", kCFStringEncodingUTF8);
...
auto p_cores_freq_property = static_cast<CFDataRef>(osPrimitives->CFDictionaryGetValue(properties, cfkey));

if (p_cores_freq_property == nullptr)
{
    throw std::system_error
    {
        0,
        std::system_category(),
        "Error on library function call CFDictionaryGetValue."
    };
}
```

It then walks the returned data as an array of little-endian 32-bit frequencies and takes the maximum.

There are five separate points in that function that throw `std::system_error` rather than returning a value: `IOServiceMatching` returning null, `IOServiceGetMatchingServices` failing, `IORegistryEntryCreateCFProperties` failing, the key being absent, and the value not being of data type. Any one of them aborts the whole function, and none of them degrades to a zero or a best-effort value. If the `pmgr` entry is never found the loop simply ends and `cpuHz` stays 0, which produces 0 rather than an error, so the function has two different failure modes as well.

The dependency is the problem rather than any one of those throw sites. `voltage-states5-sram` is an undocumented power-management property of a specific Apple SoC power manager node. It is not part of any stable interface, it is not guaranteed to be present on every Apple Silicon generation, and it is specifically not something a virtualised guest exposes, because the guest is not managing the physical package's voltage states. The agent under test is a Parallels VM, which its own inventory reports plainly as `Apple M1 (Virtual)`, and it gets null.

I want to be straightforward about what is and is not established. That the field is null on this agent is directly observed and quoted above. That the ARM implementation depends on that IORegistry property and throws when it is missing is in the source, also quoted. Which of the five throw sites actually fires on this host I did not confirm, because the machine became unreachable before I could run `ioreg` on it, so I am not asserting the precise failure point. The `pmgr` node not existing in a VM guest is the most likely explanation given what the host reports about itself, but it is an inference rather than a measurement.

Virtual machines are worth treating as a first-class case here rather than an edge case. Wazuh's own QA demo environment runs its macOS agent in a Parallels VM on a MacStadium host, and macOS VMs are how most macOS CI and most macOS test fleets are built, so an implementation that only works on bare metal will be null across a large fraction of real deployments.

Two reasonable directions. `sysctlbyname("hw.tbfrequency")` combined with the timebase, or `hw.cpufrequency_max` where present, gives a usable number without depending on an undocumented registry key. Failing that, the function should return 0 or report unavailability rather than throwing, so the distinction between "this platform cannot report a speed" and "collection errored" is visible.

For scope: the same field is also absent on the two ARM Linux agents in this environment, `001` and `004`, while both x86 Linux agents report it. That is a separate code path with a separate cause and is not part of this report, but it suggests CPU speed is generally untested on ARM and someone looking at this may want to check both.

## Environment

- Agent under test: `demo-env-macos-agent`, id `006`, macOS 26.5.1 (build 25F80), arm64, Parallels VM on MacStadium host `macstadium_arm_4`. The agent's own inventory reports `host.cpu.name` as `Apple M1 (Virtual)` and `host.os.kernel.version` as `Darwin Kernel Version 25.5.0: Mon Apr 27 20:39:32 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_VMAPPLE`, the `VMAPPLE` suffix confirming the virtualised kernel.
- Source read from the `5.0.0` branch of `wazuh/wazuh`: `src/data_provider/src/hardware/architectureDependantARMMac.cpp` and `architectureDependantX86_64Mac.cpp`.
- Comparison agents: `002` ubuntu-26-amd64 and `003` amazon-2023-amd64 report a speed; `001` ubuntu-26-arm64 and `004` amazon-2023-arm64 do not; `005` windows-11-amd64 reports one.

## Steps to reproduce

1. Enroll a 5.0.0 macOS agent on Apple Silicon, in a VM, and let a syscollector evaluation complete.
2. Read the hardware document:

```bash
  "https://<indexer>:9200/wazuh-states-inventory-hardware/_search?size=1" \
  -d '{"query":{"term":{"wazuh.agent.id":"006"}},"_source":["host"]}'
```

`host.cpu.speed` is null while cores, name, memory and serial number are populated.

3. Check on the host whether the property the implementation depends on exists:

```bash
ioreg -c AppleARMIODevice -r -n pmgr | grep -c voltage-states5-sram
```

A guest that does not expose the power manager node returns 0 matches, or `pmgr` is not present at all.

4. Compare against any x86 agent, whose `getMhz()` takes the `hw.cpufrequency` path and reports a value.

## Expected

The macOS agent reports a CPU speed on Apple Silicon, including in a virtual machine, or reports the field as genuinely unavailable rather than producing an exception inside the hardware collector.

## Actual

`host.cpu.speed` is null on the macOS agent. The Apple Silicon implementation depends on the undocumented `voltage-states5-sram` property of the `pmgr` IORegistry node and throws `std::system_error` at any of five points when it cannot be read, with no fallback to a supported interface.

## Not a duplicate of

`qa-known-issues.py check` found **no cache match** against 784 open bugs. The live search returned five candidates, all of which I read: wazuh/wazuh#39141, #39139 and #39165 are my own macOS reports about process arguments, service state and sudoers, #38309 is about vulnerability detection returning nothing for Ubuntu agents on 4.14.6, and #38914 is about a FIM `nodiff` path. None concerns CPU frequency or the hardware collector.

anromerom:
## Description

Closes #39185

On Apple Silicon macOS, `getMhz()` read a single undocumented power-management property (`voltage-states5-sram`) from the `pmgr` node of the IORegistry, and threw `std::system_error` at five separate points whenever that property, its parent node, or any intermediate IOKit call was unavailable, discarding the entire hardware inventory document for that scan. In a virtualised guest, which does not expose the physical package's power manager, the property lookup instead fell through silently and returned `0`, which the agent had no way to distinguish from a genuine 0 MHz reading. This PR makes the collector total (it never throws, and it now also tries `sysctlbyname("hw.cpufrequency")` and a sibling voltage-states key before giving up), and makes the 4.14.9 inventory harvester omit `host.cpu.speed` from the indexed document when the collected value is 0, so "unavailable" no longer indexes as a confident `0`.

The change covers two things. The collector fix is the part that must ship: it removes a failure mode where one missing registry key discarded an agent's entire hardware inventory (name, cores, memory, serial number), not just the speed. Alongside it, `hwElement.hpp` stops writing a confident `0` when the collected speed is unknown, so the field is omitted instead.

That second part is a schema-visible change which runs on the manager rather than the agent. It affects every platform whose collector reports `0` for an unknown CPU speed, not just macOS. 

## Proposed Changes

- `architectureDependantARMMac.cpp`: replace all five `throw std::system_error` sites with non-throwing returns, so `getMhz()` always completes.
- Add `sysctlbyname("hw.cpufrequency")` as the first lookup attempt, ahead of the IORegistry walk.
- Try `voltage-states1-sram` as a fallback key on the same `pmgr` node when `voltage-states5-sram` is absent, of the wrong type, or yields no nonzero sample.
- Fix an unsigned-underflow loop bound (`length - 3` on a `size_t`) that could make the frequency-parsing loop run unbounded for a degenerate (0-2 byte) property value; the loop-bound fix and the two-key fallback share this helper.
- `sysInfoHardwareWrapperARMMac_test.cpp`: invert the five throw-based tests to assert a non-throwing `0.0` return, and add tests for no `pmgr` device, an empty iterator, the `sysctl` fallback succeeding, the second key succeeding, and both degenerate-length cases.
- `hwElement.hpp` (4.14.9 `inventory_harvester` only): only assign `host.cpu.speed` when the collected frequency is greater than 0, so the field is omitted from the document instead of asserting `0`.
- `systemInventoryUpsertElement_test.cpp`: add a case asserting the serialized document omits `speed` when the collected frequency is 0.

### Results and Evidence

Unit tests were run on Linux for the harvester half. The `data_provider` collector change and its test cannot be compiled here: that test binary is gated to an arm64 macOS build host (`tests/sysInfoHardwareMac/CMakeLists.txt`) and the file is excluded from the Linux build glob. It was instead validated on a real endpoint, below.

```sql
make TARGET=agent -j4
# clean, "Done building agent"
```

The harvester unit tests were built and run by configuring the build directory directly and building the `inventory_harvester_utest` target on its own:

```sql
cmake .. -DTARGET=agent -DUNIT_TEST=ON && make inventory_harvester_utest -j4
[==========] 152 tests from 10 test suites ran.
[  PASSED  ] 152 tests.
```

A plain `make TARGET=agent TEST=yes -j4` does not complete on this tree: the `unit_tests/wrappers/externals/bzip2` and `zlib` wrappers fail with an undeclared `uintptr_t`. That failure is pre-existing and unrelated to this change; it reproduces against an unmodified `HEAD`.

`152/152` covers `inventory_harvester_utest` only, including the new `unknownCpuFrequency_Hw` case, which asserts the serialized document has no `speed` key when the collected frequency is 0. Note that running this binary under `--gtest_filter` produces false failures because of a pre-existing cross-test fixture dependency in the suite; the full unfiltered run is the meaningful one.

#### Endpoint validation

Run on macOS 26.5.1 (arm64), Apple M1 in a virtual machine, against a 4.14.9 agent, baseline and patched.

The host's own state confirms the precondition behind this issue, and this is the check the original report could not complete:

```sql
pmgr node present      : no
voltage-states5-sram   : absent
voltage-states1-sram   : absent
hw.cpufrequency        : not published
hw.tbfrequency         : 24000000
```

With the patched agent the hardware document stays complete: `host.cpu.cores`, `host.cpu.name`, `host.memory.*` and `host.serial_number` all populated, and no collector exception in `ossec.log`.

Two things this run does **not** establish, stated so reviewers can weigh them:

- `host.cpu.speed` still indexes as `0`, not as absent. The omission is written by the inventory harvester on the manager, and the manager used for the run was a stock 4.14.7 build rather than one carrying this change. The agent-side half is what was exercised here.
- This host has no `pmgr` node, so both baseline and patched take the silent-zero path and none of the five former throw sites was reached. The destructive failure mode this PR removes is covered by the new unit tests, not by this run. Exercising it needs a bare-metal Apple Silicon host where `pmgr` exists but the property does not.

`host.cpu` is therefore byte-identical between the baseline and patched documents. The endpoint run demonstrates no regression; it does not demonstrate a visible change in this environment, because on a VM there is no frequency to report either way.

### Artifacts Affected

- macOS agent, Apple Silicon only: `src/data_provider` hardware collector (`libsysinfo`).
- 4.14.9 manager: `inventory_harvester`, `host.cpu.speed` field in the `wazuh-states-inventory-hardware` index.
- No change on x86 macOS, Linux, or Windows collectors.

### Configuration Changes

N/A

### Documentation Updates

N/A

### Tests Introduced

- `sysInfoHardwareWrapperARMMac_test.cpp`: five existing throw-based tests inverted to assert `0.0`; new tests `Test_CpuMhz_NoPmgrEntry`, `Test_CpuMhz_EmptyIterator`, `Test_CpuMhz_SysctlFallbackSucceeds`, `Test_CpuMhz_SecondKeySucceeds`, `Test_CpuMhz_DegenerateKeyLengthZero`, `Test_CpuMhz_DegenerateKeyLengthTwo`, `Test_CpuMhz_FirstKeyDegenerateSecondKeySucceeds`. Not yet run; needs an arm64 macOS build host.
- `systemInventoryUpsertElement_test.cpp`: new `unknownCpuFrequency_Hw`, asserting `host.cpu.speed` is absent from the serialized document when `cpuFrequency()` is 0. Run and passing (152/152, Linux).

## Behavioural outcome

On 4.14.9, an agent that cannot determine its CPU speed (for example, an Apple Silicon VM with no `pmgr` node, or an ARM64 kernel that does not publish the property this reads) now reports `host.cpu.speed` as absent from the hardware document instead of a confident `0`. This brings 4.14.9 into line with 5.0.0, which already maps the same collected `0.0` to JSON `null` in `ecsHardwareData`; this PR does not invent a new convention, it converges an older line onto behaviour the newer line already has.

## Out of scope

ARM Linux agents show the same silent-zero outcome for `host.cpu.speed` through a different, unrelated code path (`sysInfoLinux.cpp`, `getCpuMHz`). Not addressed here.

## Review Checklist

- [x] Code changes reviewed


cborla (approved):
LGTM
- [x] Code changes reviewed

PR #39422, opened 2026-09-17
Take macOS process names from the executable path instead of the truncated pbi_name field

### LINKED ISSUE #39186: macOS process names are truncated at 31 characters, collapsing distinct system services onto identical names, while the full name sits in the executable path the same function already reads
## Description

`process.name` on the macOS agent is taken from the short name field of the BSD process info structure, which is a fixed 32-byte buffer, so any process whose name is longer than 31 characters is silently cut off at 31.

```cpp
jsProcessInfo["name"] = taskInfo.pbsd.pbi_name;
```

`pbi_name` is declared `char pbi_name[2 * MAXCOMLEN]` in `sys/proc_info.h`, and `MAXCOMLEN` is 16, so 32 bytes including the terminator and 31 usable characters.

Of the 486 processes the agent inventories, 22 report a name of exactly 31 characters, and 19 of those are provably truncated: the reported name is a strict prefix of the basename of the executable path in the same document.

```
  pid   process.name (31 chars)            actual executable basename                                 len
  471   com.apple.MobileSoftwareUpdate.    com.apple.MobileSoftwareUpdate.CleanupPreparePathService     56
  45140 com.apple.MobileSoftwareUpdate.    com.apple.MobileSoftwareUpdate.UpdateBrainService            49
  234   com.apple.MobileAsset.DownloadS    com.apple.MobileAsset.DownloadService.Builtin                45
  225   com.apple.DriverKit-IOUserDockC    com.apple.DriverKit-IOUserDockChannelSerial                  43
  478   PerfPowerTelemetryClientRegistr    PerfPowerTelemetryClientRegistrationService                  43
  770   com.apple.accessibility.mediaac    com.apple.accessibility.mediaaccessibilityd                  43
  269   com.apple.cmio.videodriverkitho    com.apple.cmio.videodriverkithostextension                   42
  714   com.apple.StreamingUnzipService    com.apple.StreamingUnzipService.privileged                   42
  589   com.apple.FaceTime.FTConversati    com.apple.FaceTime.FTConversationService                     40
  65032 Managed Background Assets Helpe    Managed Background Assets Helper Service                     40
  159   com.apple.cmio.registerassistan    com.apple.cmio.registerassistantservice                      39
  422   com.apple.quicklook.ThumbnailsA    com.apple.quicklook.ThumbnailsAgent                          35
```

The damage is not only cosmetic, because macOS service names are reverse-DNS identifiers whose distinguishing part is at the end. The first two rows above are two different system services, `CleanupPreparePathService` and `UpdateBrainService`, and both are reported under the identical name `com.apple.MobileSoftwareUpdate.`. Anything that groups or counts by `process.name`, or any rule that matches on it, cannot tell them apart. The truncation point falls in exactly the wrong place for this naming convention: the shared vendor prefix survives and the part that identifies the process is discarded.

The full name is already available in the same function, at no additional cost. `getProcessInfo()` calls `proc_pidpath()` a few lines later and stores the result:

```cpp
char pathBuffer[PROC_PIDPATHINFO_MAXSIZE] = {0};
const auto pathLen { proc_pidpath(pid, pathBuffer, sizeof(pathBuffer)) };
jsProcessInfo["command_line"] = pathLen > 0 ? std::string(pathBuffer) : "";
```

`PROC_PIDPATHINFO_MAXSIZE` is 4096, so the path is not truncated, and its basename is the untruncated name. That is precisely how the table above was built: by comparing the stored name against the basename of the stored path, using nothing the agent did not already report. Taking the name from the path when it is available, and falling back to `pbi_name` only when `proc_pidpath` fails, would fix every one of these rows without another syscall.

I want to set the scope honestly, because this is not a macOS-only phenomenon and should not be triaged as a macOS regression.

The Linux agents are affected worse. `process.name` there comes from `/proc/<pid>/comm`, which is capped at 15 characters, and on agent `001` 96 of 143 process names are exactly 15 characters long with 6 provably truncated by the same prefix test. The Windows agent is unaffected: its longest name is 23 characters and none is truncated.

I am filing this against macOS rather than as a single cross-platform report because the two are different code paths with different fixes and different fallbacks available: macOS has `proc_pidpath` sitting in the same function, Linux would need `/proc/<pid>/cmdline` or the executable link, and neither change helps the other platform. Whoever picks this up should be aware the Linux side exists and may want to open a companion issue rather than assume fixing macOS covers it.

## Environment

- Agent under test: `demo-env-macos-agent`, id `006`, macOS 26.5.1 (build 25F80), arm64, Parallels VM on MacStadium host `macstadium_arm_4`, 486 processes inventoried.
- Source read from the `5.0.0` branch of `wazuh/wazuh`: `getProcessInfo()` in `src/data_provider/src/sysInfoMac.cpp`. Buffer size from Apple's `sys/proc_info.h` and `sys/param.h`.
- Comparison agents: `001` ubuntu-26-arm64 (143 processes, 96 names at the 15-character cap), `005` windows-11-amd64 (78 processes, none truncated).

## Steps to reproduce

1. Enroll a 5.0.0 macOS agent and let a syscollector evaluation complete.
2. Pull every process document with its name and executable path:

```bash
  "https://<indexer>:9200/wazuh-states-inventory-processes/_search?size=600" \
       "_source":["process.name","process.command_line","process.pid"]}'
```

3. Flag any document whose `process.name` is a strict prefix of the basename of `process.command_line`. Every such row is a truncation, and every one of them has a name of exactly 31 characters.

4. Confirm the collision directly:

```bash
  "https://<indexer>:9200/wazuh-states-inventory-processes/_search?size=10" \
  -d '{"query":{"bool":{"filter":[{"term":{"wazuh.agent.id":"006"}},
       {"term":{"process.name":"com.apple.MobileSoftwareUpdate."}}]}},
       "_source":["process.pid","process.name","process.command_line"]}'
```

Two processes come back with the same name and different executables.

5. On the host, `ps -Ao pid,comm` shows the same truncation, while `ps -Ao pid,args` and `/usr/bin/proc_pidpath` equivalents show the full path.

## Expected

`process.name` carries the process's actual name. Two distinct services do not share a name, and a name is not cut at a fixed width when the untruncated value is already being collected in the same pass.

## Actual

The name is copied from `pbi_name`, a 32-byte field, so it is cut at 31 characters. 22 of 486 processes on the agent hit the cap, 19 are demonstrably truncated, and at least one pair of unrelated Apple system services collapses onto a single identical name, while the full name is present in the `command_line` value stored beside it.

## Not a duplicate of

`qa-known-issues.py check` found **no cache match** against 784 open bugs, the live search for `executable characters collector truncated because already` returned nothing, and the tool concluded "no known issue looks like this one".

Related but distinct: wazuh/wazuh#39141 and wazuh/wazuh#39169 concern other fields dropped by this same function, command-line arguments and CPU time respectively. Neither concerns the name field, and all three could be addressed in one pass over `getProcessInfo()`.

lchico:
|Related issue|
|---|
|#39186|

## Description

On macOS, both the process inventory and the ports/socket inventory took a process's name from `pbi_name`, a fixed 32-byte BSD field, so any process name longer than 31 characters was silently cut off. macOS system services routinely exceed that length, since their names follow Apple's reverse-DNS bundle-identifier convention (e.g. `com.apple.accessibility.mediaaccessibilityd`), so this repeatedly collapsed real, distinct-looking names into truncated, sometimes ambiguous strings — a data-integrity issue for anything that groups, counts, or matches rules on a process name, and one that also made the two inventories capable of disagreeing with each other about the same process.

This change adds a shared `resolveProcessName()` helper that resolves a process's name from its own executable path instead, taking the path's basename as the name. The full path is obtained via `proc_pidpath()`, which does not share `pbi_name`'s fixed-size limitation; the basename itself is extracted with the existing `Utils::getFilename()` helper already used elsewhere in this component. If the path lookup itself fails (e.g. a permission restriction on a privileged process), the previous `pbi_name` value is used as a fallback so a process is never left without a name. Both `getProcessInfo()` (process inventory) and `getProcessesSocketFD()` (ports/socket inventory) now call this same helper, so the two inventories are guaranteed to agree on a given process's name rather than being resolved independently.

`resolveProcessName()` has external linkage (its own header/source pair) rather than living as a private function inside the larger collector file, so its unit test can exercise the real production function directly instead of a hand-maintained duplicate of its logic.

Closes #39186

## Configuration options

N/A — no new configuration parameters.

## Logs/Alerts example

**Confirming the defect on a real macOS host (prior to this change):** the live process inventory for a real, currently-enrolled macOS test agent was queried directly against the indexer's process-inventory index:

```
curl -sk -u admin:<password> \
  "https://<indexer>:9200/wazuh-states-inventory-processes-<manager>/_search?size=1000" \
  -d '{"query":{"term":{"agent.id":"<agent-id>"}},"_source":["process.name","process.command_line","process.pid"]}'
```

Across 500 real process documents, 24 had `process.name` at exactly the 31-character `pbi_name` cap, several visibly cut mid-word.

**Confirming the fix, on the same real host, after upgrading the agent to a build containing this change:** re-querying the identical index for the identical PIDs shows the full, untruncated name for every genuinely truncated case. A representative sample:

| PID | Before (31-char cap) | After (full name) | Real length |
|---|---|---|---|
| 1048 / 1057 | `com.apple.accessibility.mediaac` | `com.apple.accessibility.mediaaccessibilityd` | 44 |
| 538 | `com.apple.MobileSoftwareUpdate.` | `com.apple.MobileSoftwareUpdate.CleanupPreparePathService` | 58 |
| 467 | `PerfPowerTelemetryClientRegistr` | `PerfPowerTelemetryClientRegistrationService` | 44 |
| 1278 | `com.apple.AppStoreDaemon.StoreP` | `com.apple.AppStoreDaemon.StorePrivilegedODRService` | 51 |

All 15 genuinely-truncated PIDs from the original 24 that still existed at re-query time showed this same before/after pattern (the remaining 9 either aged out via normal process churn, or were real 31-character names sitting at the cap boundary without ever having been truncated — confirmed individually on the host against live `ps` output, not assumed from the indexer data alone).

Cross-checking PIDs 1048/1057 directly on the host confirmed the real, untruncated executable name independently of the indexer:

```
$ ps -Ao pid,args | grep -E '1048|1057'
 1048 /System/Library/Frameworks/MediaAccessibility.framework/Versions/A/XPCServices/com.apple.accessibility.mediaaccessibilityd.xpc/Contents/MacOS/com.apple.accessibility.mediaaccessibilityd
 1057 /System/Library/Frameworks/MediaAccessibility.framework/Versions/A/XPCServices/com.apple.accessibility.mediaaccessibilityd.xpc/Contents/MacOS/com.apple.accessibility.mediaaccessibilityd
```

**Confirming the ports-inventory / process-inventory agreement:** because both inventories now resolve a process's name through the same shared helper, querying the ports-inventory index for the same agent and comparing each shared PID's name against the process-inventory value showed byte-for-byte agreement for every process currently holding an open socket on the test host (6 of 6 checked: `netbiosd`, `identityservicesd`, `replicatord`, `wazuh-agentd`, `cupsd`, `sshd-session`). None of those particular processes happened to have a name at or above the 31-character cap at the time of this check, so this specific run demonstrates cross-inventory agreement rather than a live long-name case in the ports index — the source change applies identically to both call sites regardless.

**Verifying the fix logic directly:** a unit test calls the real `resolveProcessName()` function (compiled into the test binary, not a duplicate of it) for the running test process, confirming the resolved name differs from a truncated-looking fallback and is non-empty, plus a case where the path lookup fails (`proc_pidpath()`'s documented failure value, `0`) and the given fallback value is returned unchanged. Both cases pass.

The before/after indexer and on-host evidence above was captured against the build containing the process/ports-inventory fix itself; the subsequent commit that gave `resolveProcessName()` external linkage for direct testing is a pure code-organization change with no behavioral difference and has not separately been re-verified against a fresh agent install.

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [x] Package installation
- [ ] Source upgrade
- [x] Package upgrade

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

- Decoder/Rule tests


cborla (approved):
LGTM
- [x] Code changes reviewed

Issue #39192, opened 2026-09-11
macOS block-ip offers only pf and hosts.deny, both unavailable on a default macOS install, and omits the route blackhole fallback every other Unix platform gets


Rebits:
## Description

`block-ip-unix.c` builds its blocking strategy as an ordered method chain per platform, and every platform it covers ends with a `route` fallback:

```c
const firewall_method_t methods[] = {
#ifdef PLATFORM_LINUX
    {"firewalld", try_firewalld, true},
    {"iptables",  try_iptables,  true},
    {"hostsdeny", try_hostsdeny, false},
    {"route",     try_route,     false},
#elif defined(PLATFORM_FREEBSD)
    {"ipfw",      try_ipfw,      false},
    {"pf",        try_pf,        false},
    {"hostsdeny", try_hostsdeny, false},
    {"route",     try_route,     false},
#elif defined(PLATFORM_OPENBSD)
    {"pf",        try_pf,        false},
    {"hostsdeny", try_hostsdeny, false},
    {"route",     try_route,     false},
#elif defined(PLATFORM_NETBSD)
    {"npf",       try_npf,       false},
    {"hostsdeny", try_hostsdeny, false},
    {"route",     try_route,     false},
#endif
    {NULL, NULL, false}
};
```

The `route` entry is the safety net. It needs no firewall to be configured, enabled or even present: it installs a blackhole route for the address, which works on a stock system.

`block-ip-macos.c` does not use that chain at all. It hand-rolls a two-step fallback inside `main()`, and the chain has two links:

```c
firewall_result_t result = try_pf_macos(srcip, action, ip_version, argv[0]);
...
result = try_hostsdeny_macos(srcip, action, ip_version, argv[0]);
...
// Both methods failed
log_firewall_action(argv[0], LOG_LEVEL_WARNING, "all", "failed",
                  "All blocking methods failed - IP not blocked");
```

There is no third method. macOS is the only Unix platform in this module with no route fallback.

That matters because of what the two available methods require on a stock macOS system.

**pf is disabled by default on macOS.** The implementation checks for it explicitly and gives up when it is off:

```c
if (!enabled) {
    write_debug_file(argv0, "PF firewall is not enabled");
    os_free(pfctl_path);
    return FIREWALL_INVALID_STATE;
}
```

[...]

Issue #39189, opened 2026-09-11
macOS block-ip active response permanently rewrites /etc/pf.conf and reloads the entire packet filter ruleset, discarding rules loaded by other software and never reverting the change


Rebits:
## Description

When the macOS block-ip active response finds that its pf table does not yet exist, it does not decline. It edits the system firewall configuration file and reloads the whole ruleset:

```c
if (!table_exists) {
    const char *pf_conf_path = "/etc/pf.conf";

    if (access(pf_conf_path, F_OK) >= 0) {
        ...
        if (!config_exists) {
            ...
            FILE *pf_conf = wfopen(pf_conf_path, "a");
            if (pf_conf) {
                fprintf(pf_conf, "\n# Wazuh active response table\n");
                fprintf(pf_conf, "table <wazuh_fwtable> persist\n");
                fprintf(pf_conf, "block in quick from <wazuh_fwtable> to any\n");
                fprintf(pf_conf, "block out quick from any to <wazuh_fwtable>\n");
                fclose(pf_conf);
            }
            ...
        }

        // Reload PF configuration
        char *exec_cmd_reload[] = {pfctl_path, "-f", (char *)pf_conf_path, NULL};
        wfd = wpopenv(pfctl_path, exec_cmd_reload, W_BIND_STDOUT | W_BIND_STDERR);
```

So a single alert firing a block on a macOS endpoint appends four lines to `/etc/pf.conf` and runs `pfctl -f /etc/pf.conf`.

Three things follow, in increasing order of seriousness.

**The edit is permanent and nothing ever removes it.** The unblock path takes the `DISABLE_COMMAND` branch, which does `pfctl -t wazuh_fwtable -T delete <ip>` and nothing else. There is no code anywhere in the file that removes the block it appended. Unblocking every IP, disabling active response, or uninstalling the agent all leave the four lines in place, and they keep matching a table that the agent may no longer maintain.

**`pfctl -f` is a full ruleset reload, not an incremental add.** It flushes the main ruleset and replaces it with what the file contains. Anything that was loaded into pf by another route is gone: rules added by a VPN client, a container runtime, an endpoint security product, or an administrator's own `pfctl -f` of a different file. On macOS this is a realistic collision rather than a theoretical one, because several common products manipulate pf directly. The active response has no way to know what it is about to discard, and it does it as a side effect of blocking one IP address.

[...]

PR #39424, opened 2026-09-17
Fix macOS block-ip: add route fallback when pf and hosts.deny are unavailable

cborla:
## Description

On a stock macOS install, the `block-ip` active response never blocks anything. `block-ip-macos.c` hand-rolled a two-step fallback (`pf` -> `hosts.deny`) instead of using the shared `execute_firewall_chain` helper that Linux/FreeBSD/OpenBSD/NetBSD already use in `block-ip-unix.c`, and it never included a third method. `pf` is disabled by default on macOS and `/etc/hosts.deny` doesn't exist by default, so both methods decline on an unconfigured endpoint and the response logs "All blocking methods failed - IP not blocked", while every other Unix/BSD platform in this module falls through to a `route` blackhole fallback that needs no prior configuration.

Closes #39192

## Proposed Changes

| File | Change |
|------|--------|
| `src/active-response/src/block-ip-macos.c` | Replaced the hand-rolled `pf` -> `hosts.deny` fallback in `main()` with a `firewall_method_t` chain (`pf` -> `hostsdeny` -> `route`) executed through the shared `execute_firewall_chain`, matching `block-ip-unix.c`. Added `try_route_macos()`, reusing the same BSD `-blackhole` `route` syntax already shipped for FreeBSD/OpenBSD/NetBSD in `block-ip-unix.c`'s `try_route()` — macOS's `route` is BSD-derived and takes the identical arguments. |
| `src/active-response/CMakeLists.txt` | Updated a stale comment ("macOS uses separate block-ip binary with pf only") to reflect the new three-method chain. |

CHANGELOG.md and the affected `docs/ref/modules/active-response/` pages were also updated as a consequence of this change — see Documentation Updates below.

### Results and Evidence

No new command syntax is introduced: the macOS `route` invocation is a byte-for-byte reuse of the FreeBSD/OpenBSD/NetBSD branch already shipping in `try_route()` (`block-ip-unix.c`), and the chain/fallback mechanics reuse `execute_firewall_chain`, already exercised by every other platform's `block-ip` binary. This PR has not been run against a live macOS endpoint; #39192 itself reproduced the failing case against `demo-env-macos-agent` (macOS 26.5.1, arm64, `wazuh-agent-5.0.0-latest.arm64.pkg`) and would be the environment to confirm the fix end-to-end (`pfctl -s info` disabled, no `/etc/hosts.deny`, trigger `block-ip`, check `netstat -rn` for the blackhole route).

  - [ ] Linux — N/A, `block-ip-macos.c` only compiles under `__APPLE__`
  - [ ] Windows — N/A, not touched by this change
  - [ ] MAC OS X — not yet built on a macOS host, pending

- Memory tests for Linux — N/A
- Memory tests for macOS
  - [ ] Leaks — pending macOS build
  - [ ] AddressSanitizer — pending macOS build

### Artifacts Affected

- `active-response/bin/block-ip` on macOS agent packages (arm64/x86_64).

### Configuration Changes

N/A

### Documentation Updates

Updated `docs/ref/modules/active-response/{executables,architecture,README,configuration}.md`, which described macOS `block-ip` as `pf`-or-`hosts.deny`-only and would otherwise have gone stale the moment this PR shipped a third method. While in there, also fixed two pre-existing inaccuracies unrelated to #39192 but found in the same files: `executables.md` documented a PF anchor (`wazuh_anchor`) the code doesn't use — the real behavior self-configures `/etc/pf.conf` directly, which is what wazuh/wazuh#39189 is about; `configuration.md` still described the legacy 4.x positional command format and per-method script names (`firewall-drop`, `host-deny`, `route-null`) instead of the actual 5.0.0 JSON-over-stdin protocol and the unified `block-ip`/`disable-account` executables.

### Tests Introduced

None. `unit_tests/active-response/test_firewall_helpers.c` covers the shared `execute_firewall_chain`/`check_binary_available` helpers only; there is no existing unit test harness for `block-ip-macos.c`'s method table or for `block-ip-unix.c`'s equivalent, so this PR doesn't add one either — flagging as a gap for reviewers rather than papering over it with an untested one.

## Review Checklist

- [ ] Code changes reviewed


lchico (approved):
LGTM!

- [x] Code changes reviewed

PR #39426, opened 2026-09-17
Exit zero from wazuh-agentd on a signalled shutdown

### LINKED ISSUE #39408: wazuh-agent service is left in failed state after systemctl stop (wazuh-agentd exits 1 on SIGTERM)

nbertoldo:
Closes #39408

## Description

`systemctl stop wazuh-agent` left the unit in `failed (Result: exit-code)` instead of `inactive`, on every systemd Linux agent, although `ExecStart` and `ExecStop` both returned 0 and all five daemons shut down cleanly.

Two independent pieces combine into the symptom, and only the second one is recent.

`wazuh-agentd` installs the shared signal handler: `main.c:312` calls `StartSIG(ARGV0)`, which routes SIGINT/SIGQUIT/SIGTERM/SIGALRM to `HandleSIG()` (`src/shared/src/sig_op.c:37-42`). That handler logs `SIGNAL_RECV` — the `(1225): SIGNAL [(15)-(Terminated)] Received. Exit Cleaning...` line in the issue — and then calls `exit(1)`. This is long-standing behaviour, shared with `wazuh-logcollector`, `wazuh-remoted`, `rootcheck` and (through `ExecdShutdown`) `wazuh-execd`.

What changed is that systemd started looking. #39218 added `PIDFile=` to `wazuh-agent.service` so the unit tracks `wazuh-agentd` specifically instead of falling back to cgroup-non-empty liveness — finding 57 of #38859, where the unit stayed `active` after `wazuh-agentd` died. With the daemon now the unit's `MainPID`, systemd sees that `exit(1)` on every ordinary stop and records `Result=exit-code`.

The issue's suggestion that the non-zero exit points at the 5.0.0 https / token-enrollment shutdown path does not hold. `https-client: INFO: Stopping https_client.` is simply the last INFO written before the process dies: it comes from `atexit(w_https_client_stop)` (`agentd.c:219`) running *inside* the `exit()` that `HandleSIG()` already invoked with status 1. There is no `exit()`, `_exit()` or `abort()` anywhere under `src/client-agent/https_client/`, and `hc_destroy()` swallows all exceptions, so the HTTPS client cannot set the process status.

## Proposed Changes

`wazuh-agentd` now registers its own termination handler through `StartSIG2()` and exits 0 on a signalled stop.

`StartSIG2()` is identical to `StartSIG()` apart from the handler: it still sets `pidfile`, still ignores SIGHUP, still installs `HandleSIGPIPE`, and still registers `atexit(HandleExit)` — so `DeletePID()` and `DeleteState()` cleanup is preserved unchanged. It is already the pattern `wazuh-execd`, `wazuh-syscheckd` and `addagent` use.

`src/shared/src/sig_op.c` is deliberately **not** touched, so `wazuh-remoted` and `rootcheck` on the manager keep their current exit status. Making the shared `HandleSIG()` exit 0 is arguably the right long-term change — systemd itself treats SIGTERM-induced death of a main process as clean — but it reaches manager daemons that this issue does not cover, and it deserves its own issue and manager-side regression pass.

The alternative of whitelisting the status in the unit (`SuccessExitStatus=1`) was rejected: `merror_exit()` also exits 1 (`src/shared/src/debug_op.c:616`), so every fatal startup failure — `AG_NOKEYS_EXIT`, `QUEUE_ERROR`, `PID_ERROR`, `https_client: startup failed` — would then read as a clean stop, reopening finding 57 in the worst possible way. With this change, a *requested* stop reports success while every non-signal death still lands the unit in `failed`.

A secondary symptom is fixed along the way: `wm_control_wait_for_service_active()` (`src/wazuh_modules/src/wm_control.c:129`) treats `failed` exactly like `inactive`, warns `Service %s is in state '%s', systemctl cannot reload`, and falls back to `bin/wazuh-control reload`. A unit left `failed` by a previous stop therefore degraded the next remote-configuration reload into a systemd-invisible restart.

### Results and Evidence

`test_agentd`, 20/20 including the two new cases:

```
[ RUN      ] test_agentd_shutdown_sigterm_exits_zero
Expected assertion exit called occurred
[       OK ] test_agentd_shutdown_sigterm_exits_zero
[ RUN      ] test_agentd_shutdown_sigint_exits_zero
Expected assertion exit called occurred
[       OK ] test_agentd_shutdown_sigint_exits_zero
[==========] tests: 20 test(s) run.
[  PASSED  ] 20 test(s).
```

The whole `client-agent` suite, 16/16 binaries:

```
100% tests passed, 0 tests failed out of 16
```

The new tests were checked against a deliberately reintroduced regression — `exit(1)` in `agentd_shutdown()`, rebuilt — and they fail as they should, so they are not vacuously green:

```
[  ERROR   ] --- 0x1 != 0
[  FAILED  ] test_agentd_shutdown_sigterm_exits_zero
[  FAILED  ] test_agentd_shutdown_sigint_exits_zero
```

`main.c` is not compiled under `TEST=1` (the executable target sits in the `elseif(NOT UNIT_TEST)` branch of `src/client-agent/CMakeLists.txt`), so it was compiled separately with `agentd_lib`'s own flags plus `-Wall -Wextra`: clean, emitting `U StartSIG2` and `U agentd_shutdown`, both resolved by `T agentd_shutdown` in `agentd_lib`.

Validated end to end on a real Amazon Linux 2023 agent (systemd 252) against a 5.0.0 manager, installing the two packages in turn on a clean `/var/ossec`.

Before, with the current package (`5.0.0-latest`, commit `7a47160`) — the reported symptom, reproduced exactly:

```
Active: failed (Result: exit-code) since Thu 2026-09-17 19:25:12 UTC
Process: ExecStart=/usr/bin/env /var/ossec/bin/wazuh-control start (code=exited, status=0/SUCCESS)
Process: ExecStop=/usr/bin/env /var/ossec/bin/wazuh-control stop (code=exited, status=0/SUCCESS)
Main PID: 8757 (code=exited, status=1/FAILURE)
```

`systemctl show -p MainPID` matched the contents of `PIDFile=` (8757), confirming systemd was tracking the daemon itself.

After, with this branch's package (commit `5300dff`), same cycle:

```
Active: inactive (dead)
Result=success   ExecMainCode=0   ExecMainStatus=0
```

The `(1225): SIGNAL [(15)-(Terminated)] Received. Exit Cleaning...` line is still written to `ossec.log`, and both `var/run/wazuh-agentd.pid` and `var/run/wazuh-agentd-<pid>.pid` are gone after the stop — so `HandleExit()`'s `DeletePID()`/`DeleteState()` still run under `StartSIG2()`.

Two negative cases confirm that genuine failures are still reported, which is what separates this fix from `SuccessExitStatus=1`:

- A configuration error (`<endpoint></endpoint>`, reaching `(1215): No client configured. Exiting.`) leaves the unit `failed` with `Result=exit-code`.
- `kill -9` on the tracked Main PID while the unit is active leaves it `failed` with `Result=signal`, `ExecMainStatus=9`. This is finding 57's actual scenario, and it still surfaces.

### Artifacts Affected

`wazuh-agentd` on Linux and macOS agents.

Windows is unaffected structurally, not merely by `#ifdef`: `src/client-agent/CMakeLists.txt:69-71` removes `agentd.c` from the Windows build, `main.c` is not compiled there either, and `src/win32/` never calls `StartSIG`/`StartSIG2` — the Windows agent goes through the SCM control handler in `win_utils.c`.

macOS shares the C path but not the consequence: `src/init/darwin-init.sh` generates a LaunchDaemon whose `ProgramArguments` is the `Wazuh-launcher` wrapper, with no `KeepAlive`, so launchd never observes the daemon's exit status.

Nothing in the tree consumes `wazuh-agentd`'s exit status. `stop_service()` (`src/init/wazuh-client.sh:290-327`) never inspects `$?` and always echoes success; `pkg_installer.sh` decides from `wazuh-control status` and `wazuh-agentd.state`; `restart`/`reload` do not check it either. The only two consumers are the one-shot CLI modes, `--show-token` and the uninstall validation, both of which return long before `StartSIG2()` is ever reached.

### Configuration Changes

None. The unit file, `wazuh-control` and the packaging are untouched.

### Documentation Updates

None required. The log output is unchanged: the handler emits the same `SIGNAL_RECV` message as `HandleSIG()`, so anything parsing `(1225)` keeps working.

One behavioural nuance is worth stating for whoever reads `systemctl status` next: after this change, a stray `kill` of `wazuh-agentd` outside `systemctl stop` leaves the unit `inactive (dead)` rather than `failed`. That is intended — a signalled shutdown is a deliberate stop — and finding 57's actual requirement still holds, since the unit leaves `active` either way.

### Tests Introduced

`test_agentd_shutdown_sigterm_exits_zero` and `test_agentd_shutdown_sigint_exits_zero` in `src/unit_tests/client-agent/test_agentd.c`, asserting both the `(1225)` log line and the exit code. The second case pins that the handler does not branch on the signal, which matters because `StartSIG2()` routes four signals to it.

This required wrapping `exit` for the `test_agentd` binary. The wrapper does not return — it ends in `mock_assert()`, the same convention `__wrap__merror_exit()` already uses, so the tests call it through `expect_assert_failure()`. A returning wrapper would be unsafe here: `exit()` is `noreturn`, so the compiler may omit the epilogue after the call and execution would fall out of `agentd_shutdown()` into whatever follows it. For the same reason `agentd_shutdown()` is deliberately not declared `noreturn`.

Wrapping `exit` is scoped to `test_agentd` and is safe there: `DEBUG_OP_WRAPPERS` already intercepts `_merror_exit`/`_mterror_exit`, so nothing else in that binary reaches the real `exit`, and returning from `main()` does not route through the PLT.

## Review Checklist

- [x] Code changes reviewed


cborla (approved):
LGTM
- [x] Code changes reviewed

lchico (approved):
LGTM! The check that fails is not related to the PR.

- [x] Code changes reviewed
- [x] ~~Configuration changes documented~~

Issue #39429, opened 2026-09-17
Cluster readiness does not detect degraded agent API dependencies


NahuFigueroa97:
## Description

A node can lose an internal dependency, stop serving part of the agent API, and continue receiving traffic indefinitely. **There is no signal that reflects this state**: not the load balancer health check, not the cluster status, and not the server API.

In a load-balanced deployment, this means that a fraction of the fleet proportional to the number of nodes loses group synchronization, configuration hashes, and task delivery, while continuing to send events normally. From the outside, the cluster appears healthy.

---

## Terminology: there are two `GET /` endpoints

`remoted` serves two unauthenticated liveness probes on two different servers. Everything below refers to the **public** one unless stated otherwise, because that is the only one a load balancer can reach.

| Server | Reachable from | Route | Response |
|---|---|---|---|
| `m_httpServer` — agent listener, port `1517` | the network, through the balancer | `GET /` | `{"status":"ok","module":"remoted"}` |
| `m_adminServer` — administration UDS socket | the local host only | `GET /`, `GET /metrics`, `GET /status` | `{"status":"ok","module":"remoted_module"}` |

Both are liveness by design, and the code says so explicitly:

- Public listener, `src/remoted/remoted_module/src/remotedModuleFacade.hpp:415-423` — *"Unauthenticated health probe (no request body, no auth). Exempt from the in-flight byte budget (`countAgainstBudget=false`) so liveness stays `200` even under memory pressure."*
- Admin socket, `src/remoted/remoted_module/src/remotedModuleFacade.hpp:1051-1062` — *"Unauthenticated liveness probe, mirroring the public server's `GET /` (which stays: agents probe THAT one; this one answers operators on the local socket)."*

Neither handler consults any dependency. The public one is deliberately exempt from back-pressure, so it is guaranteed to answer `200` precisely when the node is least able to serve.

---

## What happens

```mermaid
flowchart TD

    LB["Load balancer<br/>health check"] --> R["GET / on the agent<br/>listener (1517)"]

    R --> OK["200 OK"]

    OK --> KEEP["Node stays in rotation"]

    KEEP --> TRAF["One-third of traffic reaches this node"]

    TRAF --> EV["POST /stateless<br/>202, events continue flowing"]

    TRAF --> CT["POST /control<br/>500 database_error"]

    CT --> LOSS["That third of the fleet loses<br/>groups, config_hash, and tasks"]

    style LOSS fill:#fdd
```

---

## Reproduction

[...]

Issue #39428, opened 2026-09-17
Cluster operations partially succeed without rollback or success reporting


NahuFigueroa97:
## Description

Operations distributed by the API across all cluster nodes are **not atomic and have no rollback**. When they fail on a subset of nodes, the API returns an error, but the nodes where the operation succeeded **keep the applied state**, and the response does not indicate which nodes those were.

The most visible case is `PUT /agents/upgrade_custom`. The operator receives a `Bad Request` response and reasonably concludes that the upgrade was never scheduled. In reality, upgrade tasks remain active on the nodes where the operation succeeded, and the agent will later receive and execute them.

An agent upgrade reinstalls software on the endpoint and may restart it. Triggering an upgrade that the operator believes was cancelled is a change-management issue, not just a usability issue.

---

## What happens

```mermaid
flowchart TD

    OP["Operator:<br/>PUT /agents/upgrade_custom"] --> API["API on the master"]

    API -->|"DAPI distributes to all nodes"| M["master"]
    API --> W1["worker1"]
    API --> W2["worker2"]

    M --> MC{"WPK present on disk?"}
    W1 --> W1C{"WPK present on disk?"}
    W2 --> W2C{"WPK present on disk?"}

    MC -->|yes| T1["create upgrade task"]
    W1C -->|yes| T2["create upgrade task"]
    W2C -->|no| E["error 1824"]

    E --> R["API returns Bad Request<br/>mentioning only worker2"]

    R --> OP2["Operator concludes:<br/>nothing happened"]

    T1 & T2 -.->|"no rollback"| AG["Agent receives the task during the next<br/>/control notify<br/>and executes the upgrade"]

    style R fill:#fdd
    style AG fill:#fdd
```

Distribution **does not stop** at the first failure. Each node is evaluated independently, and the API aggregates the errors only after all nodes have been processed.

---

## Reproduction

Cluster with one master and two workers. Agent enrolled and active.

### 1. Place the WPK on every node except one

```bash
for n in master worker1 worker2; do
  ssh $n "mkdir -p /var/wazuh-manager/var/upgrade && \
          echo dummy > /var/wazuh-manager/var/upgrade/t.wpk && \
          chown -R wazuh-manager:wazuh-manager /var/wazuh-manager/var/upgrade"
done

ssh worker2 rm -f /var/wazuh-manager/var/upgrade/t.wpk
```

### 2. Count pending tasks on each node

```bash
for n in master worker1 worker2; do
[...]

PR #39430, opened 2026-09-17
Improve documentation and test tool about load balancer

### LINKED ISSUE #39281: Validate the Agent-to-Manager HTTPS Channel in a Multi-Node Cluster Behind Load Balancers
## Description

In 5.0.0, 5.x agents enroll, report, and receive work exclusively through remoted's HTTPS agent API on port `1517` ([`agent-api.yaml`](https://github.com/wazuh/wazuh/blob/5.0.0/docs/ref/modules/remoted/agent-api.yaml)).

So far, this channel has primarily been validated against a **single manager**:

```text
agent -- TLS 1.3 --> wazuh-manager-remoted
```

A production deployment is different: a cluster with one master and multiple workers behind a load balancer exposing a single address. This is the topology recommended by the cluster documentation for large fleets ([`cluster/README.md`](https://github.com/wazuh/wazuh/blob/5.0.0/docs/ref/modules/cluster/README.md)).

The protocol itself has already been validated behind load balancers. #38156 produced the [remoted load-balancer guides](https://github.com/wazuh/wazuh/blob/5.0.0/docs/ref/modules/remoted/load-balancers/README.md) and a reproducible lab ([`src/remoted/remoted_module/tools/load_balancer/`](https://github.com/wazuh/wazuh/blob/5.0.0/src/remoted/remoted_module/tools/load_balancer/README.md)) with 57 checks under NGINX and HAProxy.

However, that work explicitly excludes a **real cluster**. The existing lab:

* Copies `client.keys` between nodes instead of synchronizing it through `wazuh-clusterd`.
* Does not run the complete manager stack.
* Uses a Python probe instead of a real agent.
* Exercises only `POST /stateless` out of remoted's nine agent-facing routes.

Therefore, what is currently proven is:

> **The HTTPS protocol survives a proxy.**

What is not yet proven is:

> **A fleet of real agents can enroll, report, receive configuration, synchronize state, perform VD scans, and receive upgrades correctly when requests from the same agent are distributed across the nodes of a real Wazuh cluster.**

The corresponding production architecture is also not fully documented.

This issue closes that gap by:

1. Building a representative multi-node cluster behind a load balancer.
2. Validating the complete agent-to-manager HTTPS surface with real agents.
3. Identifying and fixing product defects exposed by per-request load balancing.
4. Measuring cluster propagation and other timing-sensitive behavior.
5. Turning the critical scenarios into repeatable automated tests.
6. Completing the documentation for a supported cluster-behind-a-load-balancer deployment.

---

#### Agent HTTPS API

Validate the complete remoted HTTPS surface:

* `GET /`
* `GET /cacerts`
* `POST /enroll`
* `POST /control/startup`
* `POST /control/notify`
* `POST /stateless`
* `POST /stateful`
* `POST /download`
* `POST /stats`
* `POST /config`
* `POST /scan/vd`

#### Load Balancers

Validate both deployment models documented in [`load-balancers/README.md`](https://github.com/wazuh/wazuh/blob/5.0.0/docs/ref/modules/remoted/load-balancers/README.md):

* TLS passthrough (L4).
* TLS termination with re-encryption (L7).

Test both:

* NGINX.
* HAProxy.

#### TLS and Trust

Validate:

* Listener certificates on every manager node.
* The CA trusted by agents.
* `subjectAltName` requirements for:

  * The load-balancer address.
  * Individual backend manager addresses/names.
* `<remote><https><verification_mode>` behavior in both LB models.
* The behavior of `GET /cacerts` when a load balancer is between the agent and manager.

#### Enrollment

Validate all supported enrollment credential modes:

* No credential.
* `authd.pass`.
* Enrollment token.
* Re-enrollment bearer.

The enrolling request must be able to land on any appropriate node, while identity creation remains master-controlled.

#### Cluster Data Propagation

Validate the behavior agents observe while cluster data propagates:

* `client.keys`
* `authd.pass`
* `enrollment_tokens.json`
* `etc/shared/`
* `var/multigroups/merged.mg`

Reference: [`cluster.json`](https://github.com/wazuh/wazuh/blob/5.0.0/framework/wazuh/core/cluster/cluster.json).

The propagation window and its effect on agent behavior must be measured rather than inferred.

#### Per-Node State

Validate state that can differ between nodes when requests from the same agent are load-balanced independently:

* `AgentRegistry`
* `config_token`
* `config_hash`
* `settings_hash`
* `vd_feed_offset`
* Task polling and delivery state

#### Legacy and Mixed Fleets

Validate the legacy channel (`1514`/`1515`) **at the same time as** the HTTPS channel (`1517`).

A real migration requires 4.x and 5.x agents to coexist behind the same load balancer.

#### Documentation and Automation

Produce:

* A reference architecture.
* Configuration and certificate procedures.
* Troubleshooting guidance.
* Production checklists.
* Automated regression coverage for the behaviors that cannot be covered by existing single-node tests.

### Out of Scope

* Cluster-internal synchronization performance.
* The `wazuh-clusterd` protocol itself.
* Server API behind a load balancer (`55000`), which has a different API and authentication model.
* Indexer and Dashboard clustering.
* The 4.x → 5.0.0 data migration covered by #39109.

---

# Current Gaps and Findings to Validate

The following findings come from reviewing the 5.0.0 branch. They must be confirmed or refuted in the reference environment.

## 1. Existing Load-Balancer Lab Is Not a Real Cluster

The existing [`add_second_manager.sh`](https://github.com/wazuh/wazuh/blob/5.0.0/src/remoted/remoted_module/tools/load_balancer/add_second_manager.sh) creates a second bare `remoted` under `/var/wazuh-manager-2` with **no engine**.

This is intentional: the lab uses the second instance as an authentication oracle:

* `503` → signature accepted.
* `401` → signature rejected.

However:

* `client.keys` is copied rather than synchronized.
* `wazuh-clusterd` is not running.
* `wazuh-db` shared state is not represented.
* There is no downstream event pipeline.
* `run_issue_checks.sh` exercises only `POST /stateless`.

Consequently, everything an agent performs after authentication is currently untested through a load balancer, including:

* Configuration download.
* Inventory/state sessions.
* Task polling.
* VD scan requests.
* Agent upgrades.

---

## 2. Cluster Documentation Does Not Describe Port `1517`

[`cluster/lb.md`](https://github.com/wazuh/wazuh/blob/5.0.0/docs/ref/modules/cluster/lb.md) documents balancing for the legacy `1514`/`1515` channels and redirects 5.x users to the remoted load-balancer documentation.

This leaves the documentation fragmented:

* No complete reference architecture for **cluster + load balancer + agents**.
* No consolidated port map.
* No explanation of which traffic must reach the master.
* No certificate topology.
* No explanation of how legacy and HTTPS channels coexist during migration.

[`architecture.md`](https://github.com/wazuh/wazuh/blob/5.0.0/docs/ref/architecture.md) does not fill this gap.

---

## 3. Per-Request Balancing Interacts With Per-Node State

The load-balancer guide states that, with TLS termination, any manager can receive any request from any agent.

The critical missing piece is the state that is **not shared between nodes**.

### `/download` authorization

`POST /download` authorizes against remoted's in-memory `AgentRegistry`, which is populated by `/control` on that specific node.

Therefore:

```text
/control on worker A
        ↓
/download on worker B
        ↓
403
```

This has already been reported in #39147. Behind a non-sticky load balancer, this becomes a normal request-routing scenario rather than a rare edge case.

### `config_hash`

`config_hash` is the SHA-256 of that node's `var/multigroups/merged.mg`.

If the file differs between nodes, the agent can observe different hashes and repeatedly re-download its configuration.

### `settings_hash`

`settings_hash` is derived from `limits` and `cluster.name` and cached process-wide.

A divergent `limits` configuration on one node can therefore make the value change depending on which node handles the request.

### `vd_feed_offset`

`vd_feed_offset` is read from the local vulnerability-detection module.

Nodes at different feed offsets can therefore expose different values to the same agent.

### Task Delivery

`/control/notify` marks pending tasks as `delivered` locally without a cluster broadcast.

Custom WPKs must exist on every node under `var/upgrade/` ([`agent-upgrades.md`](https://github.com/wazuh/wazuh/blob/5.0.0/docs/ref/modules/task_manager/agent-upgrades.md)), and `var/upgrade/` is not included in `cluster.json`.

### Agent Database State

#38422 reports agent databases remaining on a worker after an agent reconnects elsewhere.

With non-sticky balancing, this should be treated as a normal routing scenario and explicitly validated.

### Required Outcome

Every confirmed per-node-state dependency must be:

* Fixed.
* Documented as an explicit architectural constraint.
* Or tracked in a dedicated issue.

---

## 4. Cluster Key Propagation Window Is Not Measured

`cluster.json` configures worker `sync_integrity` to 9 seconds.

A freshly enrolled agent may therefore be routed to a worker that has not received its key yet and receive `401`:

```text
enrollment succeeds
        ↓
agent immediately reports
        ↓
request reaches worker without the key
        ↓
401
```

The same propagation problem applies to:

* `authd.pass` — enrollment fails closed with `401` while unsynchronized.
* `enrollment_tokens.json` — an unknown `kid` causes a rate-limited re-read before `TokenUnknown`.

The spike must measure:

* Actual propagation time.
* Agent retry behavior during the window.
* Whether the agent eventually recovers without intervention.
* The operator-visible behavior.
* The documentation and recommended mitigation.

---

## 5. Certificates Are Not Cluster-Synchronized

`cluster.json` synchronizes selected files under `etc/`, including:

* `client.keys`
* `authd.pass`
* `enrollment_tokens.json`

It does **not** synchronize `etc/certs/`.

Therefore, every node requires its own listener certificate issued externally.

The certificate setup must account for:

### TLS Termination

The balancer presents the certificate validated by the agent.

### TLS Passthrough

The agent connects through the balancer to individual manager nodes and therefore must be able to validate the certificate presented by each backend.

The spike must establish and document:

* Certificate issuance using `wazuh-certs-tool`.
* Required SANs.
* Certificate layout per node.
* Balancer certificate requirements.
* Rotation without dropping the fleet.
* Failure behavior for:

  * Expired certificate.
  * Wrong CA.
  * SAN mismatch.
  * Replaced node certificate.

The current lab uses only valid certificates and does not validate these failure modes.

---

## 6. `verification_mode` Documentation Contradicts the Implementation

The current documentation is inconsistent.

The lab README states that a third mode:

> existed while this spike ran

and was later removed.

However:

* `wazuh-manager.schema.json` still accepts `none | certificate | full`.
* Remoted still implements `full`.
* `full` includes the `403` behavior and startup warning requiring an appropriate `IP:` SAN.
* `load-balancers/README.md` contains a dedicated section describing `full`.

This must be resolved before the deployment can be documented as supported.

**Required outcome:** determine whether `full` is supported in 5.0.0 and make the schema, implementation, lab, and documentation consistent.

---

## 7. `global_prefix` Has No Cluster Validation Story

The global endpoint prefix must be identical:

* On every manager node.
* On every agent.
* Through the load balancer.

The load balancer must not rewrite the prefix.

A mismatch results in `404`, not an authentication error.

Fresh configurations use `/wazuh-manager/`, but there is currently no validation preventing a later-added node or agent from using a different value.

The spike must validate:

* Correct configuration.
* Node-to-node consistency.
* Agent consistency.
* Failure behavior.
* Diagnostic behavior.

---

## 8. `GET /` Is a Liveness Check, Not a Readiness Check

`GET /` returns `200` even when the engine is unavailable and the node rejects event processing with `503`.

A load balancer can therefore continue routing traffic to a node that is alive but unable to process agent traffic.

The spike must either:

1. Define and implement a readiness endpoint, or
2. Document a supported balancer-side mitigation, such as treating sustained `503` responses as unhealthy.

---

## 9. Mixed 4.x/5.x Fleets Are Not Documented as a Single Architecture

During migration, the same load balancer must support:

```text
4.x agents ──> 1514 / 1515
5.x agents ──> 1517
```

Enrollment on `1515` must reach the master.

The current documentation describes these channels separately but does not explain the combined front end required during migration.

This must be covered by both testing and documentation.

---

## 10. There Is No Automated Regression Coverage

The existing integration suites contain:

* `test_remoted`
* `test_enrollment`
* `test_agentd`

but none are cluster- or proxy-aware.

The load-balancer validation is currently implemented as manually executed shell scripts under `src/remoted/remoted_module/tools/`.

Without automated coverage, a future change to remoted, `cluster.json`, or the manager schema could silently reintroduce these failures.

---

# Tasks

## Phase 1 — Build the Reference Environment

* [ ] Build the reference environment:

  * One master.
  * Two workers.
  * Real `wazuh-clusterd` synchronization.
  * Indexer.
  * Dashboard.
  * Load balancer in front of `1517`.
* [ ] Support both:

  * NGINX.
  * HAProxy.
* [ ] Support both TLS models:

  * Passthrough.
  * Termination with re-encryption.
* [ ] Issue the PKI using `wazuh-certs-tool`:

  * CA trusted by agents.
  * One listener certificate per manager node.
  * Required SAN for each backend.
  * Load-balancer certificate with the SAN used by agents.
* [ ] Record the exact certificate-generation and deployment commands.
* [ ] Enroll real Linux and Windows agents through the load balancer.
* [ ] Use several agent groups, including at least one multigroup agent.
* [ ] Test all enrollment modes:

  * No credential.
  * `authd.pass`.
  * Enrollment token.
  * Re-enrollment.
* [ ] Add the legacy front end:

  * `1514` → all nodes.
  * `1515` → master.
* [ ] Enable `<remote><legacy>` and `<auth><legacy_enrollment>`.
* [ ] Connect 4.x agents through the same load balancer.

---

### HTTPS Route Matrix

* [ ] Run the complete `agent-api.yaml` route matrix through the load balancer using real agents.
* [ ] Test both TLS models.
* [ ] Test both proxies.
* [ ] Record for every request:

  * Route.
  * HTTP status.
  * Node that handled the request.
  * Agent action following the response.

### Cluster Propagation

* [ ] Measure the key-propagation window:

  * Enroll an agent.
  * Immediately begin reporting.
  * Count `401` responses.
  * Measure how long failures persist.
  * Confirm automatic recovery.
* [ ] Repeat for:

  * `authd.pass`.
  * Enrollment token minted on the master and immediately used against a worker that has not yet synchronized.

### Per-Node State

* [ ] Deliberately spread requests from the same agent across different nodes.
* [ ] Validate:

  * `/control` → `/download` across nodes.
  * `config_hash` stability.
  * `config_token` behavior.
  * `settings_hash` stability.
  * `vd_feed_offset` stability.
  * Task polling when the task was created on a different node.
* [ ] Create one dedicated issue for every confirmed product defect and link it from this issue.

### Certificates

* [ ] Validate:

  * Expired leaf.
  * Wrong CA.
  * SAN mismatch.
  * Node certificate replacement while agents are reporting.
  * Full CA rotation.
* [ ] Confirm that agent-side errors are diagnosable from agent logs.

### `verification_mode`

* [ ] Determine whether `verification_mode: full` is supported in 5.0.0.
* [ ] Align:

  * Schema.
  * Code.
  * Lab.
  * Remoted load-balancer documentation.

### `global_prefix`

* [ ] Validate identical prefixes across:

  * All manager nodes.
  * Agents.
  * Load balancer.
* [ ] Deliberately introduce a mismatch on one node.
* [ ] Confirm the resulting failure is `404`.
* [ ] Confirm the failure is diagnosable.

### Agent Upgrade

* [ ] Execute an agent upgrade through the load balancer.
* [ ] Create the upgrade task on one node and retrieve the WPK from another.
* [ ] Test both TLS models.
* [ ] Test a custom WPK.
* [ ] Verify the constraint that upgrades to `>= 5.0.0` are rejected unless `verification_mode` is `none`.
* [ ] Determine whether this constraint remains appropriate behind a load balancer.

### Inventory and VD

* [ ] Exercise `/stateful` at fleet scale.
* [ ] Exercise `/scan/vd` at fleet scale.
* [ ] Validate:

  * 20 MB request-body limit.
  * 20-second `/stateful` deadline.
  * Proxy limits from sections 4.4 and 4.5 of the load-balancer guide.
  * `503` back-pressure reaching the agent unchanged.

### Failure Drills

* [ ] Kill a worker while agents are reporting.
* [ ] Kill the master and determine:

  * What stops.
  * What continues.
  * Whether enrollment stops.
  * Whether group changes stop.
* [ ] Restart the load balancer.
* [ ] Remove a worker and add it back.
* [ ] Confirm agent recovery in every case.

### Mixed Fleet

* [ ] Run 4.x and 5.x agents simultaneously through the same load balancer.
* [ ] Upgrade a 4.x agent in place.
* [ ] Verify its transition from `1514` to `1517` without re-enrollment.

---

## Phase 3 — Automated Tests

* [ ] Decide where the automated cluster + load-balancer suite will live.
* [ ] Extend the existing load-balancer lab into a real cluster:

  * Replace copied `client.keys` with `wazuh-clusterd` synchronization.
  * Add missing routes to `run_issue_checks.sh`.
  * Preserve its existing PASS/FAIL contract.
* [ ] Add automated checks for:

  * Certificate failures.
  * Propagation behavior.
  * Critical cross-node request flows.
* [ ] Ensure the documented measurements are generated and/or validated by the automated suite where practical.

---

## Phase 4 — Documentation

* [ ] Create a **reference architecture** page for a Wazuh cluster behind a load balancer covering:

  * Topology.
  * Architecture diagram.
  * Port map.
  * Traffic that must reach the master.
  * TLS termination vs. passthrough.
  * Certificate layout.
  * Per-node data not synchronized by the cluster:

    * `etc/certs/`
    * `var/upgrade/`
* [ ] Rewrite [`cluster/lb.md`](https://github.com/wazuh/wazuh/blob/5.0.0/docs/ref/modules/cluster/lb.md) around the 5.x architecture:

  * `1517` as the primary 5.x channel.
  * Legacy `1514`/`1515` as the migration section.
  * Combined mixed-fleet architecture.
* [ ] Extend [`load-balancers/README.md`](https://github.com/wazuh/wazuh/blob/5.0.0/docs/ref/modules/remoted/load-balancers/README.md) §7 ("In a cluster") with:

  * Measured cluster constraints.
  * Propagation window.
  * Per-node state requirements.
  * Certificate requirements.
* [ ] Document the certificate procedure:

  * Node certificate generation.
  * Load-balancer certificate generation.
  * Required SANs.
  * Deployment.
  * Rotation.
  * Failure modes.
  * `GET /cacerts` behavior in both TLS models.
* [ ] Add troubleshooting guidance for:

  * Intermittent `401`.
  * `404` on every request.
  * `403` only on `/download`.
  * `413`.
  * `502`.
  * `503`.
  * TLS failures that do not produce an HTTP status.
* [ ] Update `docs/SUMMARY.md`.
* [ ] Resolve every documentation contradiction found during validation.
* [ ] Add a production checklist specifically for the cluster-behind-a-load-balancer deployment.
* [ ] Rebuild the environment from scratch using **only the new documentation** and repeat the Phase 1/2 validation.
* [ ] Attach the resulting evidence to this issue.

---

# Automation Strategy

The validation is only valuable if the critical behaviors can be reproduced after implementation changes.

The main options are:

### Option A — Extend the Existing Load-Balancer Lab

Keep the existing:

```text
setup_lab.sh
run_issue_checks.sh
```

structure and:

* Replace copied `client.keys` with a real cluster.
* Add the missing routes.
* Add certificate and propagation checks.
* Preserve the existing PASS/FAIL interface.

**Advantages:**

* Lowest implementation cost.
* Existing PKI configuration can be reused.
* Existing proxy configurations can be reused.
* Existing probe and test harness can be reused.
* No new CI infrastructure is immediately required.

**Disadvantages:**

* Remains a manually executed developer tool.
* Bash-based.
* Runs on a single host.
* Does not automatically provide CI coverage.

### Option B — Pytest Integration Suite

Create a `test_cluster_lb/` module under `tests/integration/`, alongside `test_remoted` and `test_enrollment`.

The suite would provision:

* Manager nodes.
* Cluster.
* Proxy.
* Real agents.

and execute the critical scenarios in CI.

**Advantages:**

* CI coverage.
* Per-test reporting.
* Tests live with the existing integration suite.
* Easier long-term regression coverage.

**Disadvantages:**

* Requires multi-host or multi-container fixtures.
* Existing integration infrastructure does not currently provide this.
* Certificate setup would need Python-side integration.

### Option C — Pytest + Permanent Environment

Build on Option B with a permanently provisioned environment and scheduled tests.

The environment could additionally cover:

* AWS ALB/NLB.
* Load behavior.
* Idle timeouts.
* Long-running certificate rotation.
* Other managed-balancer-specific behavior.

This provides coverage that a local lab cannot reproduce, but requires permanent infrastructure and ownership.

## Recommendation

Use a staged approach:

1. **Phases 1 and 2 → Option A**

   * The existing lab provides the fastest path to discovering architectural problems.
   * The goal at this stage is to learn where the system breaks, not to build the final CI architecture.

2. **Phase 3 → Option B**

   * Move the load-bearing scenarios into integration tests.
   * Prioritize:

     * Enrollment and key propagation.
     * `/control` → `/download` across nodes.
     * `config_hash` stability.
     * Certificate failure modes.
   * Reuse Option A's PKI scripts and proxy configurations where possible.

3. **Option C → Defer**

   * Revisit permanent infrastructure when managed-balancer coverage becomes a release requirement.
   * Until then, document that managed balancers are supported according to their documented behavior but are not continuously measured by Wazuh's test environment.

---

# Risks and Considerations

* **CI resource consumption:** A complete fixture requires three managers, an indexer, and a proxy. Limit CI coverage to behaviors that genuinely require multiple nodes.
* **Timing-based tests:** Propagation-window tests are inherently susceptible to flakiness. Assert eventual convergence and record the measured window rather than relying on an unnecessarily strict fixed timeout.
* **Documentation drift:** The existing `verification_mode` discrepancy demonstrates that documentation and implementation can diverge. Measured values included in documentation should be traceable to the automated suite where possible.
* **Product defects vs. documented limitations:** A confirmed defect must not automatically become a documented limitation. Each finding must be explicitly classified as:

  * Fix required.
  * Supported architectural constraint.
  * Separate issue.
* **Certificate complexity:** TLS passthrough and TLS termination have different certificate requirements and failure modes. Both must be documented independently.
* **Master dependency:** Some operations legitimately require the master. The documentation must clearly distinguish master-required operations from operations that can be served by any worker.

---

# Definition of Done

* [ ] A cluster with one master and two workers behind **NGINX** and **HAProxy**, using both TLS passthrough and TLS termination, successfully serves a real fleet of 5.x agents through every route in `agent-api.yaml`, with evidence attached to this issue.
* [ ] Enrollment, re-enrollment, and cluster key propagation are measured, with the operator-visible propagation window documented.
* [ ] Every per-node-state dependency identified in finding 3 is:

  * Fixed.
  * Documented as a supported constraint.
  * Or tracked in a dedicated linked issue.
* [ ] The complete certificate procedure is documented and tested, including:

  * Issuance.
  * SAN requirements.
  * Deployment.
  * Rotation.
  * Failure modes.
* [ ] `verification_mode` values are consistent across:

  * Manager schema.
  * Remoted implementation.
  * Load-balancer lab.
  * Documentation.
* [ ] A reference architecture page exists, is linked from `docs/SUMMARY.md`, and documents the complete 5.x cluster + load-balancer topology.
* [ ] `cluster/lb.md` is updated for 5.x and documents the mixed 4.x/5.x front end.
* [ ] The automated suite selected in Phase 3 exists and passes.
* [ ] Critical measurements quoted by the documentation are generated or validated by the automated test environment.
* [ ] The final documentation is sufficient to recreate the reference deployment from a clean environment without relying on undocumented steps.
* [ ] The complete Phase 1 and Phase 2 validation has been repeated using only the resulting documentation.
* [ ] All confirmed product defects have dedicated linked issues or have been fixed as part of this work.

NahuFigueroa97:
## Description

Validating the 5.x agent-to-manager HTTPS channel against a real multi-node cluster behind load
balancers (#39281) surfaced two gaps that this PR closes.

The first is documentation. There was no page describing how to run a Wazuh server cluster behind a
load balancer. `docs/ref/modules/remoted/load-balancers/` covers the proxy configuration itself but
assumes a single manager, so the questions a cluster raises — which certificate SANs are required
under each TLS model, which operations still need the master, what an agent observes while the
cluster replicates state — had no answer. Several behaviours that look like failures are expected
and transient, and an operator meeting them for the first time has no way to tell.

The second is the validation environment. `src/remoted/remoted_module/tools/load_balancer/` held a
single-manager harness that could not reproduce the findings, because the decisive variable is a
**cluster** behind **non-sticky** balancing. `api/test/integration/env/` does provision a cluster,
but it pins each agent to one node with `balance source`, seeds `client.keys` with enrollment off,
and runs its agents with `verification_mode none` — a configuration that hides every finding.

Related to #39281. Two product issues were opened from this validation and are **not** fixed here:
#39428 (cluster operations partially succeed without rollback or success reporting) and #39429
(cluster readiness does not detect degraded agent API dependencies). The documentation added here
describes the observable behaviour so operators are not surprised by it; it does not claim either is
resolved.

## Proposed Changes

**Cluster documentation**

- `docs/ref/modules/cluster/lb.md` — substantially rewritten (+705/−82). The page existed but did
  not carry what a deployment needs: the port map and which operations still require the master,
  what the cluster replicates and what it does not, the choice between TLS passthrough and
  termination and why balancing granularity (connections vs requests) changes the answer, the
  certificate SAN rule each model imposes, CA rotation without dropping the fleet, what an agent
  observes while the cluster catches up (with measured figures), a step-by-step deployment, and
  migrating a 4.x fleet.
- `docs/ref/modules/cluster/lb-troubleshooting.md` — **new**, a symptom-first companion:
  intermittent and persistent `401`, `404` on everything, `403` on configuration download, `413`,
  `502`, `503`, TLS failures with no HTTP status, and a degraded node serving part of the API. Each
  entry states whether the symptom is expected and how long it should last.
- `docs/SUMMARY.md` — the new troubleshooting page indexed under the existing cluster entry.

**Existing documentation corrected**

- `docs/ref/modules/client/configuration.md` — the agent's default `verification_mode` was
  documented backwards. `w_agent_resolve_ssl_posture()` resolves a ladder in which the presence of
  `etc/certs/root-ca.pem` decides between `full` and `none`; the page stated the default was
  `system` and that `none` is never the default, and did not document the anchor rung at all.
- `docs/ref/modules/remoted/agent-api.yaml` — `GET /cacerts` now states that it is not a usable
  bootstrap behind a TLS-terminating balancer, because it returns the CA that signs the *manager's*
  listener certificate while the agent's TLS peer is the proxy.
- `docs/ref/modules/remoted/configuration.md` — `global_prefix` said that a freshly generated
  configuration ships `/wazuh-manager/`, which reads as if omitting the tag serves unprefixed
  endpoints. It does not: an absent tag means the prefix is applied, and serving unprefixed
  endpoints requires writing `/` explicitly. This is the setting behind the cluster-wide `404`
  described in the troubleshooting page.
- `docs/ref/modules/remoted/load-balancers/{README,nginx,haproxy}.md` — cross-references to the
  cluster pages, and the cluster-specific caveats these pages previously omitted.

**Validation environment replaced** (`src/remoted/remoted_module/tools/load_balancer/`)

- A real cluster: one master and two workers running `wazuh-clusterd`, a single-node indexer, and
  real 4.x and 5.x agents from packages — not seeded keys.
- Both proxies serve **TLS passthrough and TLS termination simultaneously**, plus a third
  termination front end presenting a certificate from a different CA to model a public edge.
  Balancing is roundrobin, not sticky.
- `setup_lab.sh` brings everything up in one command and verifies the supplied manager package
  actually carries the features under test, because a package older than the feature answers as if
  the defect did not exist.
- `generate_certs.sh` issues the whole PKI in one invocation of the installation assistant's
  `wazuh-certs-tool`, using `--agent-san` and the `load_balancer:` config section
  (wazuh/wazuh-installation-assistant#1027).
- `run_issue_checks.sh` re-runs every check as PASS/FAIL.
- Probes under `probe/` for the route matrix with node attribution, replication propagation
  windows, cross-node `/download` convergence, enrollment credential modes, availability under node
  loss, and body-size limits across every path. They import the repository's own `wire_jwt.py` from
  the parent `tools/` directory rather than carrying a copy, so the wire format cannot drift.
- `cert_drill.sh` for certificate failure drills (expired leaf, wrong CA, SAN mismatch).
- No packages, private keys or generated PKI are committed; `.gitignore` covers them.

The 30 deleted files are the previous single-manager harness. The behaviours it exercised that the
new lab does not yet cover (`zstd` bodies, TLS session resumption, `merge_slashes`, PROXY protocol
rejection) are listed explicitly in the lab's README under "What this lab does not cover" rather
than dropped silently.

### Results and Evidence

Full suite, from a clean rebuild — fresh PKI, images rebuilt, cluster formed from scratch:

```
  20 passed, 0 failed
```

Route matrix through the four front ends, with node attribution from the balancer's header:

```
  route                 hap-pass    hap-term    ngx-pass    ngx-term
  GET /                 200         200 [3n]    200         200
  GET /cacerts          200         200 [3n]    200         200
  POST /enroll          200,409     409 [3n]    409         409
  POST /control:start   200         200 [2n]    200         200
  POST /control:notify  200         200 [3n]    200         200
  POST /stateless       202         202 [3n]    202         202
  POST /stateful        400         400 [3n]    400         400
  POST /download        200         200 [3n]    200         200
  POST /stats           200         200 [3n]    200         200
  POST /config          200         200 [3n]    200         200
  POST /scan/vd         409         409 [3n]    409         409

  PASS  no front end failed to answer on any route
  PASS  no authenticated route answered 401
  PASS  termination spread across the cluster    10 rows served by all three nodes
```

Body-size limits across every path, showing the ordering invariant the documentation describes
(accepted, then `413` from the authentication cap, then the connection closed by the transport cap):

```
  body size ->    4 MiB   9 MiB  10 MiB  11 MiB  19 MiB  20 MiB  21 MiB
  direct master     202     202     202     413     413     413  closed
  haproxy pass      202     202     202     413     413     413  closed
  haproxy term      202     202     202     413     413     413     502
  nginx pass        202     202     202     413     413     413  closed
  nginx term        202     202     202     413     413     413     413
```

The mdBook builds with both new pages indexed; `docs/build.sh` reports no errors in them.

_Additional evidence (propagation-window figures, certificate drills, the 4.x-to-5.x transition and
the end-to-end upgrade) is recorded in the update comment on #39281._

  - [ ] Linux
  - [ ] Windows 
  - [ ] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...


src/remoted/remoted_module/tools/load_balancer/README.md
```diff
@@ -1,416 +1,302 @@
...
 
 ```bash
-./send_signed_request.py                                         # direct      -> 202
-./send_signed_request.py --url https://127.0.0.1:8443            # termination -> 202
-./send_signed_request.py --url https://127.0.0.1:8444            # passthrough -> 202
-./send_signed_request.py --url https://127.0.0.1:8443 --tamper   # corrupted token -> 401
+./cert_drill.sh "expired leaf"  certs-bad/expired.pem       certs-bad/expired-key.pem
+./cert_drill.sh "wrong CA"      certs-bad/rogue.pem         certs-bad/rogue-key.pem
+./cert_drill.sh "SAN mismatch"  certs/wazuh-badsan/node.pem certs/wazuh-badsan/node-key.pem
+./cert_drill.sh restore
```
matigarciadev, line 200: These two commands reference certificate files that `generate_certs.sh` does not create (`certs-bad/expired.pem`, `certs-bad/rogue.pem`, and their keys). Following the documented setup therefore leaves both drills unable to install the intended certificates.

src/remoted/remoted_module/tools/load_balancer/send_signed_request.py
matigarciadev, line 1: `src/remoted/remoted_module/tools/wire_jwt.py:5` still names `load_balancer/send_signed_request.py`, which this PR deletes. One-line docstring update.

src/remoted/remoted_module/tools/load_balancer/run_issue_checks.sh
```diff
@@ -1,462 +1,160 @@
...
-UNEXPECTED_PASS=()
-xfail() {
-    local label="$1" broken="$2" actual="$3" would_be="$4"
-    if [[ "$actual" == "$broken" ]]; then
-        printf '  \033[35mXFAIL\033[0m %-52s %s  (known bug)\n' "$label" "$actual"
-        XFAILED=$((XFAILED + 1))
+check_not() {  # check_not <description> <unwanted> <actual>
+    local what="$1" bad="$2" got="$3"
+    if [[ "$got" != "$bad" ]]; then
+        printf '  PASS  %-62s %s\n' "$what" "$got"; PASS=$((PASS+1))
```
matigarciadev, line 29: This comparison reports PASS when the actual value is empty, so missing or unparseable probe output can pass the negative enrollment checks. Adding a non-empty guard helps, but the callers also need stronger assertions: HTTP 500 does not demonstrate credential rejection, and both `absent` and a negative expiry value currently pass the “certificate has not expired” check.

src/remoted/remoted_module/tools/load_balancer/run_issue_checks.sh
```diff
@@ -1,462 +1,160 @@
...
+echo "=== 2. the cluster is real ==="
+NODES="$(docker exec wazuh-master /var/wazuh-manager/bin/cluster_control -l 2>/dev/null | tail -n +2 | grep -c .)"
+check "cluster_control lists three nodes" "3" "$NODES"
 
-use_config breaks_signature_path_rewrite
-check "path rewriting: the route does not exist"    "404" "$(status --url "$TERM" --target /wazuh/stateless)"
-note "With the bearer token nothing about the path is authenticated: a rewrite is a routing failure (404),"
-note "never a 401. A proxy that rewrites paths must map them onto routes the manager actually serves."
+echo
+echo "=== 3. GET / returns a body (load-balancers/README.md:203 says it does not) ==="
```
matigarciadev: This PR updates the README to document the JSON response body, so this message describes a discrepancy that no longer exists. The hardcoded line number can also become stale as the document changes. Could we simplify this to “GET / returns the documented JSON body”?

src/remoted/remoted_module/tools/load_balancer/README.md
```diff
@@ -1,416 +1,310 @@
...
+with `(8134) Could not chmod 'var/upgrade/upgrade.sh'`, which does not name the real cause.
+
+### The 4.x to 5.x in-place transition
 
 ```bash
-./start_load_balancer.sh both_topologies
-./start_load_balancer.sh                 # no arguments: lists the scenarios
+docker exec wazuh-agent4-a cat /var/ossec/etc/client.keys > migrate/legacy.keys
+docker stop wazuh-agent4-a
+docker compose up -d migrated-agent
```
LucioDonda: `docker-compose.yml` still defines ten services, none of them `migrated-agent`, and the string
`migrate` appears nowhere in it. The new `migrate/entrypoint.sh` remains unreachable, and the PR
description still lists the 4.x→5.x transition among the evidence recorded on #39281.

**Fix:** add the service — image `lab-agent5:branch`, `entrypoint: /migrate/entrypoint.sh`, mounts
`./migrate:/migrate` and `./certs:/lab-certs:ro`, a free static IP, and a `profiles:` key so it does
not start with the rest of the lab.

src/remoted/remoted_module/tools/load_balancer/base/manager/entrypoint.sh
```diff
@@ -0,0 +1,123 @@
...
+# workers is one of the things under test.
+if [[ "$NODE_TYPE" == "master" && -n "${AUTHD_PASS:-}" ]]; then
+    printf '%s\n' "$AUTHD_PASS" > "$DIR/etc/authd.pass"
+    chown root:wazuh-manager "$DIR/etc/authd.pass"
+    chmod 640 "$DIR/etc/authd.pass"
+    echo "[entrypoint] authd.pass seeded on the master only"
+fi
+
+# Extra per-node overrides, applied last so a scenario can change anything above.
+if [[ -d /lab-overrides ]]; then
```
LucioDonda, line 109: Still sources `/lab-overrides/*.sh`; `README.md:239` still says
overrides are applied from there; no manager service mounts anything at that path. The
`if [[ -d /lab-overrides ]]` guard is always false and `scenarios/break-indexer-ca.sh` never runs.

**Fix:** mount `./scenarios:/lab-overrides:ro` — ideally on one node only, since injecting a fault
into a single node is the whole point.

LucioDonda (dismissed):
## LGTM

### The two doc corrections check out against the source

**`global_prefix` default.** The schema declares `"default": "/wazuh-manager/"` and
`manager_config` fills declared defaults into the effective document (RF-4, `src/defaults.cpp`),
which `Read_Remote_JSON()` then consumes — so an absent tag *does* apply the prefix. Confirmed on a
running manager built from this branch:

```
$ grep global_prefix /var/wazuh-manager/etc/wazuh-manager.conf
      <global_prefix>/wazuh-manager/</global_prefix>
$ bin/wazuh-manager-conf get remote.https.global_prefix
/wazuh-manager/

# the same file with the tag deleted
$ bin/wazuh-manager-conf -f /tmp/c.xml -H /var/wazuh-manager get remote.https.global_prefix
/wazuh-manager/
```

and the live consequence the troubleshooting page describes:

```
GET /wazuh-manager/  -> 200   {"status":"ok","module":"remoted"}
GET /                -> 404
```

**Agent `verification_mode` ladder.** The new table matches `w_agent_resolve_ssl_posture()`
(`src/client-agent/src/config.c`) branch for branch — `ca_set → certificate`, `anchor → full`,
otherwise `none` — with `AGENT_ANCHOR_CA` = `etc/certs/root-ca.pem` (`src/shared/include/defs.h:298`).
The text it replaces claimed the default was `system` and that `none` is never the default, so this
is a real fix rather than a rewording.

### Lab

- `docs/build.sh` exits 0 with no errors; `lb.html` and `lb-troubleshooting.html` are both generated
  and indexed in `SUMMARY.md`.
- `migrated-agent` is wired: `profiles: [migrate]`, `entrypoint: /migrate/entrypoint.sh`, mounts
  `./migrate` and `./certs`, IP `172.28.0.73`.
- `overrides/` is mounted `ro` at `/lab-overrides` on `wazuh-worker2` only, which is what
  `base/manager/entrypoint.sh:109` sources, and `.gitignore` keeps it empty but for its README.
  Staging directory rather than mounting `scenarios/` directly is the right call.
- The expired-leaf fixture generates and self-verifies (`certificate has expired`, `notAfter` in the
  past, SAN intact). Going through `openssl ca` with explicit dates is the correct workaround for
  `openssl x509` not being able to backdate.
- 12 shell scripts and 7 Python probes parse; `docker-compose.yml` parses; 0 certificates, keys or
  packages committed.
- CI: 81 pass, 2 skipped, 0 failed.

I could not bring the containerised lab up here, so the round-trip scenarios — route matrix,
propagation windows, body-size limits, the cert drills end to end — are **not** verified on my side.
Their wiring is correct; the measurements in the description are taken on trust.

### Non-blocking nits

None of these should hold the merge.

- `base/nginx/nginx.conf:94` — `proxy_ssl_name wazuh-lb-nginx` pins the upstream check to the
  balancer's own name, while the comment above it says "the backend leaf carries its own node name;
  verify against that". It passes only because every node leaf carries the balancer SAN. HAProxy does
  verify per node (`verifyhost`). Worth correcting the comment, since these pages teach the SAN rule.
- `setup_lab.sh:97` — the cluster wait has no ceiling, so a cluster that never forms hangs the
  one-command bring-up with no diagnostic. `cert_drill.sh:install_cert()` already has the right shape
  (120 s cap plus a message).
- `generate_certs.sh` — `chmod -R a+rX "$BAD"` leaves `expired-key.pem` and `rogue-key.pem`
  world-readable. Throwaway lab keys, so the impact is nil, but it is the shape
  wazuh/wazuh-installation-assistant#1026 is fixing.
- `docker-compose.yml:188` — `MIGRATE_PLACE_ANCHOR: "0"` is a literal, so trying the other half of
  the ladder means editing the compose file. `${MIGRATE_PLACE_ANCHOR:-0}` would make it an env var.
- `probe/enroll_modes.py` — good catch on the stale-token pick. While in there: this is the one probe
  that does not import `wire_jwt`, re-deriving `WAZUH-ENROLL-JWT-KEY` at line 97 where
  `wire_jwt.derive_enroll_key()` already exists. The token and re-enrollment rungs genuinely are not
  in `wire_jwt`, so only the shared-password rung is affected.

src/remoted/remoted_module/tools/load_balancer/build_wpk.sh
```diff
@@ -0,0 +1,47 @@
...
+# This exercises the whole upgrade chain for real -- task delivery, download through the
+# balancer, signature verification, unpacking, installer execution -- without needing the
+# Wazuh release signing key. What it does NOT exercise is the production signature path.
+#
+#   ./build_wpk.sh                 build wpk/lab-upgrade.wpk
+#   ./build_wpk.sh --install       also stage it on every node and retarget wazuh-agent5-term
+set -euo pipefail
+
+HERE="$(cd "$(dirname "$0")" && pwd)"
+WPKPACK="$HERE/../../../../packages/wpk/wpkpack.py"
```
matigarciadev: This path needs one more `../`. From `src/remoted/remoted_module/tools/load_balancer/`, four levels up reaches `src/`, so the script looks for `<repo>/src/packages/wpk/wpkpack.py`. The file actually lives at `<repo>/packages/wpk/wpkpack.py`.

I reproduced this with valid lab certificates: `./build_wpk.sh` exits with code `1` and reports `wpkpack.py not found`. This also blocks the documented `./build_wpk.sh --install` flow before packaging starts:

```console
cd src/remoted/remoted_module/tools/load_balancer
./build_wpk.sh                                                                               
wpkpack.py not found at /workspaces/wazuh-5.x/wazuh/src/remoted/remoted_module/tools/load_balancer/../../../../packages/wpk/wpkpack.py
```

src/remoted/remoted_module/tools/load_balancer/probe/measure_propagation.py
```diff
@@ -0,0 +1,138 @@
...
+        for n, url in nodes.items():
+            if first_ok[n] is not None:
+                continue
+            st = probe(url, args.prefix, agent_id, key, args.ca, args.timeout)
+            dt = time.monotonic() - t0
+            counts[n][st] = counts[n].get(st, 0) + 1
+            timeline.append({"t": round(dt, 3), "node": n, "status": st})
+            if st != 401:
+                first_ok[n] = dt
+                print(f"  [{dt:7.3f}s] {n:14s} ACCEPTED (status {st})")
```
matigarciadev, line 118: `probe()` returns a string such as `ERR ConnectionError`, `ERR ReadTimeout`, or `ERR SSLError` when the request fails, but the polling loop treats anything other than integer `401` as successful propagation:

```python
if st != 401:
    first_ok[n] = dt
```

This means a transport failure is recorded as the moment the node accepted the key. Once `first_ok[n]` is set, that node is no longer polled, so the later real `202`/`503` response is never observed.

I reproduced this both deterministically and in the lab. With a stopped worker, the probe reports:

```text
worker2 ACCEPTED (status ERR ConnectionError)
```

and exits successfully. In a scripted transient case, worker2 actually accepted at ~4.0 s, but a transport error at ~0.5 s caused it to be recorded as accepted at 0.5 s, reducing the reported propagation window from ~4.0 s to ~1.25 s.

Transport errors are inconclusive and should not count as key acceptance. At minimum, non-integer results should keep the node in the polling set. Preferably, the success condition should explicitly match the statuses documented by this probe:

```python
if st in (202, 503):
    first_ok[n] = dt
```

`401` should keep waiting, and `ERR *` results should be counted/reported separately so the operator can tell whether a measurement was affected.

matigarciadev (approved):
### LGTM!

### Built and Install
```console
General settings:
    TARGET:             manager
    V:                  
    DEBUG:              
    EXTERNAL_SRC_ONLY:  
    HTTP_REQUEST_BRANCH:c5f94080441fc35a80c5cf8ddd9cc1ca9fc20d9b
User settings:
USE settings:
    USE_INOTIFY:        no
    USE_BIG_ENDIAN:     no
    USE_SELINUX:        no
    USE_AUDIT:          no
    IMAGE_TRUST_CHECKS: 1
Compiler:
    MAKE              make

Done building manager

Wait for success...
success
cd ../framework && /var/wazuh-manager/framework/python/bin/python3 -m pip install . --use-pep517--prefix=/var/wazuh-manager/framework/python && rm -rf build/
Processing ./.
  Created wheel for wazuh: filename=wazuh-5.0.0-py3-none-any.whl size=312395 sha256=ba494c540e01ebc60dcd4498b4cf4cc1a0ede5c8ba35836826fc42b35e0e74ec
  Stored in directory: /tmp/pip-ephem-wheel-cache-hyz1t0ur/wheels/50/e9/6c/171615ee3709da12266524fed91cb0863278bc995eed9295be
Successfully built wazuh
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended touse a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the --root-user-actionoption if you know what you are doing and want to suppress this warning.
Processing ./.
  Created wheel for api: filename=api-5.0.0-py3-none-any.whl size=126157 sha256=780df76e91e74de9ac9901b18c147fa8789e9cf4cbc82928bda9bfec7d6affe8
  Stored in directory: /tmp/pip-ephem-wheel-cache-kb12ckpe/wheels/12/3e/cd/fe7f9fbf555382f244bd9d9630026d54f4ffbcb658cc4f4190
Successfully built api
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended touse a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the --root-user-actionoption if you know what you are doing and want to suppress this warning.
Generating schema files...
Loading resources...
Loading wcs_flat.yml...
Successfully merged 1 files into temporary file: /tmp/merged_wcs_8cqla8gv.yml
Loading schema template...
Success.
Success.
Success.
Generating engine schema...
Success.
Success.
Success.
Cleaned up temporary file: /tmp/merged_wcs_8cqla8gv.yml
Generated wazuh-decoders.json
Success.
Copying store files...
Installing GeoIP databases...
NOTICE: no TLS certificate for the HTTPS agent listener was found
        (/var/wazuh-manager/etc/certs/remoted.pem, /var/wazuh-manager/etc/certs/remoted-key.pem).
        wazuh-manager does not generate certificates. Provision root-ca.pem,
        remoted.pem and remoted-key.pem with the Wazuh installation assistant
        (wazuh-certs-tool) before starting the service; wazuh-manager-control
        refuses to start until they exist. See 'Deploy certificates' in the
        installation guide (docs/ref/getting-started/installation.md).


Starting Wazuh...
(1244): Invalid configuration at '/remote/https/certificate': file not found: /var/wazuh-manager/etc/certs/remoted.pem (the manager does not generate certificates; provision the file, e.g. withwazuh-certs-tool).
wazuh-manager.conf: Configuration error. Exiting

 - To start Wazuh:

 - To stop Wazuh:

   More information at: 
```

### Manager Status
```console
wazuh-manager-db not running...
wazuh-manager-authd not running...
wazuh-manager-apid not running...
Wazuh v5.0.0 Stopped
Starting Wazuh v5.0.0...
Started wazuh-manager-apid...
Started wazuh-manager-db...
Completed.

```

### Tests
**Setup Lab:**
```console
./setup_lab.sh --packages /workspaces/wazuh-5.x/pr39430-packages
==> staging packages from /workspaces/wazuh-5.x/pr39430-packages
    wazuh-manager_5.0.0-0_amd64_ca112a0.deb -> base/manager/wazuh-manager.deb
    wazuh-agent_5.0.0-0_amd64.deb -> base/agent5/wazuh-agent.deb
    wazuh-indexer_5.0.0-latest_amd64.deb -> base/indexer/wazuh-indexer.deb
==> verifying the manager package carries the features under test
    ok
==> PKI
    reusing certs/ (use --regenerate for a fresh CA)
==> building images
    done
==> starting the cluster, the balancers and the agents
[+] Running 10/10
 ✔ Container wazuh-indexer      Running0.0s 
 ✔ Container wazuh-master       Running0.0s 
 ✔ Container wazuh-worker2      Running0.0s 
 ✔ Container wazuh-worker1      Running0.0s 
 ✔ Container wazuh-lb-haproxy   Running0.0s 
 ✔ Container wazuh-lb-nginx     Running0.0s 
 ✔ Container wazuh-agent5-term  Running0.0s 
 ✔ Container wazuh-agent4-b     Running0.0s 
 ✔ Container wazuh-agent5-pass  Running0.0s 
 ✔ Container wazuh-agent4-a     Running0.0s 
==> waiting for the cluster to form
    NAME           TYPE    VERSION  ADDRESS       
    wazuh-master   master  5.0.0    wazuh-master  
    wazuh-worker1  worker  5.0.0    172.28.0.12   
    wazuh-worker2  worker  5.0.0    172.28.0.13   
==> starting the probe (mounts the repository's own agent tools from ../)
    lab-probe ready

=== ready. Verify with: ===
    ./run_issue_checks.sh          every check, PASS/FAIL
```

**Run Issue Checks:**
```console
./run_issue_checks.sh                                                                   
=== 1. every front end answers, with full CA validation ===
  PASS  haproxy passthrough                                            200
  PASS  haproxy termination                                            200
  PASS  nginx passthrough                                              200
  PASS  nginx termination                                              200
  PASS  direct to master                                               200
  PASS  direct to worker1                                              200
  PASS  direct to worker2                                              200

=== 2. the cluster is real ===
  PASS  cluster_control lists three nodes                              3

=== 3. GET / returns the documented JSON body ===
  PASS  body is the liveness JSON                                      {"status":"ok","module":"remoted"}

=== 4. termination spreads per REQUEST across all three nodes ===
  PASS  three distinct nodes served nine requests                      3

=== 5. a prefix mismatch is a 404, never a 401 ===
  PASS  unprefixed path on a prefixed manager                          404

=== 6. TLS health metrics reflect the node's certificate ===
  PASS  master: the CA signs the served leaf                           1.0
  PASS  master: the certificate has not expired                        3649.0

  Certificate FAILURE drills run standalone, because they restart a node:
    ./cert_drill.sh "expired leaf"  certs-bad/expired.pem       certs-bad/expired-key.pem
    ./cert_drill.sh "wrong CA"      certs-bad/rogue.pem         certs-bad/rogue-key.pem
    ./cert_drill.sh "SAN mismatch"  certs/wazuh-badsan/node.pem certs/wazuh-badsan/node-key.pem
    ./cert_drill.sh restore
=== 7. enrollment credential modes ===
  PASS  no credential is refused                                       401
  PASS  shared password enrolls                                        200
  PASS  enrollment token enrolls                                       200
  PASS  unknown token id is refused                                    401

=== 8. every route answers through every front end ===
    enrolled 018 for this run
    route                 hap-pass              hap-term              ngx-pass              ngx-term              
    GET /                 200                   200 [3n]              200                   200                 
    GET /cacerts          200                   200 [3n]              200                   200                 
    POST /enroll          200,409               409 [3n]              409                   409                 
    POST /control:start   200                   200 [3n]              200                   200                 
    POST /control:notify  200                   200 [3n]              200                   200                 
    POST /stateless       202                   202 [3n]              202                   202                 
    POST /stateful        400                   400 [3n]              400                   400                 
    POST /download        200                   200 [3n]              200                   200                 
    POST /stats           200                   200 [3n]              200                   200                 
    POST /config          200                   200 [3n]              200                   200                 
    POST /scan/vd         409                   409 [3n]              409                   409                 
  
    [Nn] = distinct nodes seen for that cell, from the balancer's X-Lab-Node header.
    Passthrough cannot report it: the balancer never opens the response.
  PASS  no front end failed to answer on any route                     0
  PASS  no authenticated route answered 401                            0
  PASS  termination spread across the cluster                          11 rows served by all three nodes

======================================================================
  20 passed, 0 failed
======================================================================
```

LucioDonda (approved):
LGTM.

Re-verified on `54ade4d61c` after the rebase. No product code in this PR — the diff outside
`docs/` and `src/remoted/remoted_module/tools/` is empty — so nothing needs rebuilding to review it.

### The two fixes in this push are both real

**`build_wpk.sh`** — the old `WPKPACK` path was four levels up, which resolves to
`src/packages/wpk/wpkpack.py`; the file lives at the repo root. The script guards on
`[[ -f "$WPKPACK" ]]`, so the WPK drill aborted immediately every time. Confirmed:

```
4 levels -> src/packages/wpk/wpkpack.py   does not exist
5 levels -> packages/wpk/wpkpack.py       exists
```

**`measure_propagation.py`** — this one matters, because `lb.md` quotes measured propagation
windows and this is the script that produced them. `attempt()` returns an `"ERR <ExcName>"`
*string* when the request never reaches the manager, and the old `if st != 401: first_ok[n] = dt`
accepted that string as "key arrived", stopping the clock on a connection refused or a read
timeout. Every blip biased the window short. Narrowing acceptance to `202`/`503` — the two answers
that are genuinely the manager's — and reporting the inconclusive count separately is the right
fix.

### Doc corrections still check out

`global_prefix` — the schema declares `"default": "/wazuh-manager/"` and `manager_config` fills
declared defaults into the effective document (RF-4, `src/defaults.cpp`), which `Read_Remote_JSON()`
then consumes, so an absent tag *does* apply the prefix. On a running manager:

```
$ grep global_prefix /var/wazuh-manager/etc/wazuh-manager.conf
      <global_prefix>/wazuh-manager/</global_prefix>
$ bin/wazuh-manager-conf get remote.https.global_prefix
/wazuh-manager/

# the same file with the tag deleted
$ bin/wazuh-manager-conf -f /tmp/c.xml -H /var/wazuh-manager get remote.https.global_prefix
/wazuh-manager/
```

and the live consequence the troubleshooting page describes:

```
GET /wazuh-manager/  -> 200   {"status":"ok","module":"remoted"}
GET /                -> 404
```

The `verification_mode` ladder table has landed in the base since the last push; what remains here
is the function name and the two-consequences prose, and both are consistent with
`w_agent_resolve_ssl_posture()` (`src/client-agent/src/config.c`) — `ca_set → certificate`,
`anchor → full`, otherwise `none`, with `AGENT_ANCHOR_CA` = `etc/certs/root-ca.pem`.

### Rest of the checks

- `docs/build.sh` exits 0 with no errors; `lb.html` and `lb-troubleshooting.html` are both generated
  and indexed (`SUMMARY.md:171`).
- `migrated-agent`: `profiles: [migrate]`, `entrypoint: /migrate/entrypoint.sh`, `172.28.0.73`.
- `overrides/` mounted `ro` at `/lab-overrides` on `wazuh-worker2` only — what
  `base/manager/entrypoint.sh:109` sources — and gitignored but for its README.
- 12 shell scripts, 7 Python probes and `docker-compose.yml` all parse; 0 certificates, keys or
  packages committed.
- CI: 81 pass, 2 skipped, 0 failed.

I have not brought the containerised lab up, so the round-trip scenarios — route matrix,
propagation windows, body-size limits, the cert drills end to end — are **not** verified on my side.
The wiring for them is correct; the measurements in the description are taken on trust.

PR #39463, opened 2026-09-18
Add macOS support to disable-account, which silently failed on every Darwin endpoint

### LINKED ISSUE #39190: The macOS package ships a disable-account active response that rejects Darwin outright, so the containment action can never succeed on a macOS endpoint
## Description

`disable-account` is one of only two active response binaries the macOS agent package installs, and it is compiled from a source that supports exactly one operating system. Everything that is not Linux is rejected:

```c
if (uname(&uname_buffer) < 0) {
    write_debug_file(argv[0], "Cannot get system name");
    cJSON_Delete(input_json);
    return OS_INVALID;
}

if (!strcmp("Linux", uname_buffer.sysname)) {
    // passwd parses options with getopt, so the username must be delimited
    end_of_options = 1;

    // Checking if passwd is present
    if (get_binary_path("passwd", &cmd_path) < 0) {
        ...
    }

    memset(args, '\0', COMMANDSIZE_4096);
    if (action == ENABLE_COMMAND) {
        snprintf(args, COMMANDSIZE_4096 -1, "-l");
    } else {
        snprintf(args, COMMANDSIZE_4096 -1, "-u");
    }

} else {
    memset(log_msg, '\0', OS_MAXSTR);
    snprintf(log_msg, OS_MAXSTR - 1, "Invalid system: '%s'", uname_buffer.sysname);
    write_debug_file(argv[0], log_msg);
    cJSON_Delete(input_json);
    return OS_INVALID;
}
```

`uname().sysname` on macOS is `Darwin`, so every invocation on a macOS endpoint takes the `else` branch, writes `Invalid system: 'Darwin'` to the active response log, and returns `OS_INVALID`. There is no macOS branch anywhere in the file.

The binary is nonetheless built for macOS and installed. Extracting the payload of the shipped package:

```
  /Library/Ossec/active-response/bin/block-ip
  /Library/Ossec/active-response/bin/disable-account
```

and those are the only two entries under `active-response`. The file is a native build for the platform it cannot serve:

```
$ file disable-account
disable-account: Mach-O 64-bit arm64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE>
```

and the strings confirm the Linux-only branch survived into the shipped artefact:

```
Linux
passwd
The passwd file '%s' is not accessible: %s (%d)
Invalid system: '%s'
```

So half of the active response capability presented on a macOS endpoint is inert. An operator who configures `disable-account` for a macOS fleet, which is a reasonable thing to do given the binary is sitting there, gets a response that is dispatched, runs, and fails, for every account and every alert.

The action is implementable on macOS, which is what makes this a defect rather than an unsupported platform. macOS exposes account disabling through `pwpolicy -u <user> -disableuser` and its `-enableuser` counterpart, and at a lower level through the `AuthenticationAuthority` attribute, where adding `;DisabledUser;` disables the account. The agent already links against OpenDirectory elsewhere in the codebase and already reads `AuthenticationAuthority`-adjacent data for the user inventory, so the platform knowledge is present in the product.

If the intent is genuinely that macOS is out of scope for this response, then the fix is the opposite one: stop installing the binary in the macOS package, so the capability is absent rather than present and broken. Shipping it and failing at runtime is the worst of the three options, because nothing surfaces the gap until an operator is relying on it during an incident.

Two notes on scope. `block-ip`, the other binary in that directory, does have a real macOS implementation in `block-ip-macos.c`, so this is not a case of active response being unimplemented on macOS generally. And I did not execute the response on the endpoint: the host became unreachable partway through this work, and active response on this agent is separately blocked by wazuh/wazuh#39113. The finding rests on the source, on the contents of the shipped package, and on `uname` returning `Darwin` on macOS, all of which are quoted above.

## Environment

- Target platform: macOS. Agent under test `demo-env-macos-agent`, id `006`, macOS 26.5.1 (build 25F80), arm64, Parallels VM on MacStadium host `macstadium_arm_4`.
- Package examined: `wazuh-agent-5.0.0-latest.arm64.pkg` from `nightly-backup/2026-09-09`, 7,924,612 bytes, etag `a1f1c9597d8bf76edb053f71c1e29857`. Payload extracted locally and enumerated; `active-response/bin` contains exactly `block-ip` and `disable-account`.
- Source read from the `5.0.0` branch of `wazuh/wazuh`: `src/active-response/src/disable-account.c`, 157 lines, the platform dispatch at lines 61 to 94.

## Steps to reproduce

1. Install the 5.0.0 macOS agent package and confirm the binary is present:

```bash
ls -l /Library/Ossec/active-response/bin/
file /Library/Ossec/active-response/bin/disable-account
```

Both `block-ip` and `disable-account` are there, and the latter is a native Mach-O executable.

2. Configure a `disable-account` active response on the manager targeting the macOS agent, and trigger the rule that fires it.
3. Read the active response log on the endpoint:

```bash
sudo tail -n 20 /Library/Ossec/logs/active-responses.log
```

The entry is `Invalid system: 'Darwin'`, and the account is untouched.

4. Confirm the account was never disabled:

```bash
sudo pwpolicy -u <username> -getpolicy
dscl . -read /Users/<username> AuthenticationAuthority
```

Neither shows a disabled state.

## Expected

Either `disable-account` disables the named account on macOS, using `pwpolicy` or the `AuthenticationAuthority` attribute as the platform provides, or the binary is not installed by the macOS package at all so the unsupported capability is visibly absent rather than silently failing.

## Actual

The macOS package installs a native arm64 `disable-account` binary whose only supported platform branch is `strcmp("Linux", uname_buffer.sysname)`. On macOS, `uname` reports `Darwin`, the `else` branch logs `Invalid system: 'Darwin'` and returns `OS_INVALID`, and no account is ever disabled.

## Not a duplicate of

`qa-known-issues.py check` scored one cache candidate at 0.40, which I read in full before filing:

- `wazuh/wazuh#8150` "Active response blocks are not correctly parsed" is a 2021 report against the manager on Linux about how the configuration parser resolves multiple `<active-response>` blocks in `ossec.conf`. It is a configuration-parsing issue on a different component and platform, unrelated to a response binary rejecting its own host operating system. The score comes from the shared word "blocks".

The live search for `disable-account response rejects disable package account` returned nothing.

Related but distinct: wazuh/wazuh#39189 concerns what the macOS `block-ip` response does to `/etc/pf.conf` when it does run, and wazuh/wazuh#39113 concerns active response not executing on this agent. This report is that one of the two shipped responses has no macOS implementation at all.

hernanvalenzuela:
## Description

The macOS agent package ships a `disable-account` active response binary compiled from a source that only implements Linux, SunOS, and AIX. On macOS, `uname().sysname` returns `Darwin`, which falls into the generic error branch, logs `Invalid system: 'Darwin'`, and returns `OS_INVALID` without touching the target account. Every `disable-account` invocation on a macOS endpoint silently fails, so the containment action can never succeed there. Reported in #39190.

## Proposed changes

- Added a `Darwin` branch to the platform dispatch in `disable-account.c`, following the same structure already used for the Linux/SunOS and AIX branches.
- Implemented disable/enable via `pwpolicy -u <user> -disableuser` / `-enableuser`, the mechanism macOS exposes for account state (`passwd`/`chuser` are not applicable there).
- `pwpolicy` places the target user before the action flag, unlike `passwd`/`chuser`, which take the user last. Introduced an `is_macos` flag to branch the command-argument construction so the existing Linux/SunOS/AIX argument order is unaffected.

### Results and Evidence

Validated on a real macOS endpoint matching the reported environment (macOS 26.5.1, build 25F80, arm64), not just by reading the code.

**Control test (unmodified `disable-account.c`, pre-fix):**
Reproduced the exact reported failure. Invoking the binary with a real `add` active-response message against a local test account produced:

Invalid system: 'Darwin' and exited with `OS_INVALID`, confirming the account was never touched, as described in #39190.

**With the fix:**
- `add` (disable): exit `0`, log shows `Ended`. `pwpolicy -u <user> -getaccountpolicies` afterward returns `Error: Credential verification failed because account is disabled.`, confirming the account state actually changed.
- `delete` (enable): exit `0`, log shows `Ended`. `pwpolicy -u <user> -getaccountpolicies` afterward returns normally (no disabled-account error), confirming the account was re-enabled.
- `pwpolicy` resolves correctly through `get_binary_path()`'s `$PATH` lookup on macOS (`/usr/bin:/bin:/usr/sbin:/sbin`), so no hardcoded-path change was needed there.
- `disable-account.o` builds clean with Apple clang 17 (Xcode Command Line Tools) via the project's own `make disable-account TARGET=agent`, no new warnings.
- Linux/SunOS/AIX branches are unchanged; only a new `Darwin` branch and the command-argument construction for it were added.

Both directions were exercised by feeding the binary the same JSON active-response protocol `execd` sends over stdin (`add`/`continue` and `delete`), against a disposable local test account, on a throwaway VM provisioned for this validation.

  - [ ] Linux
  - [ ] Windows 
  - [x] MAC OS X

- Memory tests for Linux
  - [ ] Coverity
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

  - [ ] Test run in parallel

- Wazuh server API/Framework

### Artifacts Affected

N-A

### Configuration Changes

N-A

### Tests Introduced

N-A

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...


cborla (approved):
LGTM
- [x] Code changes reviewed

PR #39464, opened 2026-09-18
Fix macOS block-ip: stop rewriting /etc/pf.conf, and correct the subprocess handling it exposed

### LINKED ISSUE #39189: macOS block-ip active response permanently rewrites /etc/pf.conf and reloads the entire packet filter ruleset, discarding rules loaded by other software and never reverting the change

nbertoldo:
## Description

On macOS, `try_pf_macos()` reacted to a missing `wazuh_fwtable` by appending a table declaration and two block rules to `/etc/pf.conf` and running `pfctl -f /etc/pf.conf`. That is a full reload of the main ruleset, so it replaced whatever was live with the contents of the file, and nothing ever removed the appended lines — unblocking, disabling active response and uninstalling the agent all left them in place.

Every other platform in this module treats the table as a precondition the administrator owns. `try_npf()` in `block-ip-unix.c` declines with `FIREWALL_INVALID_STATE` when `wazuh_blacklist` is missing, and the FreeBSD/OpenBSD `try_pf()` does not check for the table at all. macOS was the only place that wrote to the system firewall configuration on its own.

Reproducing it turned up a second defect the report had only hypothesised, and fixing that one exposed a family of others in the same file. All of them are macOS-only: nothing here changes `block-ip-unix.c`, `block-ip-windows.c` or the shared `execute_firewall_chain`.

**The existence check closed its pipe without reading it.** `wpopenv` binds stdout and stderr to a single pipe and `wpclose` closes the read end before `waitpid`. On macOS `pfctl` writes `No ALTQ support in kernel` / `ALTQ related functions disabled` to stderr on **every** invocation, including successful ones, so there is always something in that pipe: `pfctl` is killed by `SIGPIPE`, `WIFEXITED` is false, and the table reads as missing. The issue's follow-up comment expected this to need a table large enough to fill the pipe buffer; measured on an endpoint it happens with an **empty** table, on the first block and every one after it, so `pfctl -f /etc/pf.conf` ran on every block request rather than once.

The two cannot be fixed separately: declining on a missing table while the check always answers "missing" would make `pf` decline unconditionally and silently stop being used.

Draining is not enough on its own, because what a caller needs is rarely the first thing in the pipe. `wpopenv` puts stdout and stderr on one pipe; stderr is unbuffered and arrives immediately, stdout is block-buffered and flushed at exit. On a `pfctl` table operation the two ALTQ notices therefore land ahead of the verdict the caller is looking for. The shared `drain_and_close()` reads the stream to the end and joins what it found into one string, minus those notices, so a decision does not depend on an ordering nobody controls.

**The same shape is at three more sites, and two of them change the outcome.** `pfctl -k` closes its pipe undrained, so the kill is signalled before reaching `DIOCKILLSTATES` and connections from a blocked address survive, contrary to what the documentation promises. The `pfctl -s info` read breaks out of its loop with output still queued. And a `NULL` `wpopenv()` at the table check was reported as a missing table rather than an execution failure.

**`route(8)` failures were reported as successes.** macOS `route` exits 0 whatever happens — `newroute()` returns `void` and `main()` calls `exit(0)` unconditionally — so checking the wait status first made the idempotency branch below it unreachable and turned genuine failures, such as an IPv6 block with IPv6 down, into `IP blocked successfully`. The result now comes from stderr, which `route` leaves empty on success. The wait status is still consulted for the one thing it does say: a child that never became `route` at all, because `wpopenv`'s failed `execvp` `_exit(127)`s without writing anything.

**Unblocking could report success without undoing anything.** The chain stops at the first method that reports success, and on `DISABLE` both `pf` and `hosts.deny` reported success for an address they never held — `pfctl -T delete` prints `0/1 addresses deleted.` and still exits 0, and `try_hostsdeny_macos` returned `FIREWALL_SUCCESS` whether or not a matching line existed. A block applied by the `route` blackhole was therefore never lifted once `pf` or `hosts.deny` became available, while every log said it was. This PR makes the two macOS methods decline when the address is not theirs; `execute_firewall_chain` is untouched, so no other platform changes. `hosts.deny` declines before committing its rewrite: the copy it builds is byte-identical, and `OS_MoveFile()` is a `rename()`, so letting it run would hand the administrator's file a new inode with the agent's ownership and mode for no reason.

Closes #39189

## Proposed Changes

| File | Change |
|------|--------|
| `src/active-response/src/block-ip-macos.c` | Decline with `FIREWALL_INVALID_STATE` when `wazuh_fwtable` is absent, instead of editing `/etc/pf.conf` and reloading the ruleset. New `drain_and_close()` replaces the four copies of the drain idiom, reads each stream to the end and joins it minus the ALTQ notices. `route` binds stderr alone and its result is decided by that output, with the wait status kept only to catch a child that never exec'd. `pf` and `hosts.deny` decline on `DISABLE` when the address is not the one they hold, `hosts.deny` before it commits its rewrite. `NULL` `wpopenv()` at the table check is an execution failure. |
| `src/active-response/CMakeLists.txt` | Drop `PLATFORM_MACOS`, which no `#ifdef` in the tree has ever read. |
| `docs/ref/modules/active-response/executables.md` | The macOS section documented a PF anchor (`wazuh_anchor`, `/etc/pf.anchors/wazuh_anchor`) that no code implements. Replaced with the real precondition, plus the new unblock behaviour and the `pfctl -k` caveat. |
| `src/unit_tests/active-response/CMakeLists.txt` | Drop `PLATFORM_MACOS` from the test target's definitions. |
| `src/unit_tests/wrappers/wazuh/shared/exec_op_wrappers.{c,h}` | Additive: `wpopenv_captured_flags()` alongside the existing argv capture, so a test can assert which streams a command is bound to. No existing behaviour changes; `test_execd`, `test_wm_docker` and the active-response suite were run against it. |
| `src/unit_tests/active-response/test_block_ip_macos.c` | 9 cases added to the file #39424 introduced — the first coverage `try_pf_macos()` has had. |

### Results and Evidence

Two endpoints were used: macOS 26.5.1 (build 25F80) arm64 for the original reproduction and the end-to-end check, and macOS 15.7.2 arm64 for the `pfctl`/`route` measurements behind the later fixes. `block-ip` reads its JSON on stdin, so the response was driven directly, exactly as execd drives it.

**The reported defect.** With `pf` enabled, rules loaded from two sources other than `/etc/pf.conf`, and `wazuh_fwtable` absent, a single block of `192.0.2.66`:

```
$ sudo diff /tmp/pf.conf.before /etc/pf.conf
27a28,32
> 
> # Wazuh active response table
> table <wazuh_fwtable> persist
> block in quick from <wazuh_fwtable> to any
> block out quick from any to <wazuh_fwtable>

$ sudo pfctl -s rules
scrub-anchor "com.apple/*" all fragment reassemble
anchor "com.apple/*" all
block drop in quick from <wazuh_fwtable> to any
block drop out quick from any to <wazuh_fwtable>
```

The rule loaded from the other file is gone. The unblock emptied the table and left `/etc/pf.conf` byte-identical to the state above: the edit is permanent. Apple documents this hazard in the default `/etc/pf.conf` itself — *"Care must be taken to ensure that the main ruleset does not get flushed, as the nested anchors rely on the anchor point defined here"* — and `pfctl` warns about it on every `-f`. Repeating the block with the table already present and **empty** reloaded the ruleset every time, 8 out of 8.

**The SIGPIPE mechanism,** isolated with a harness replicating `wpopenv`/`wpclose` from `src/shared/src/exec_op.c` exactly — one pipe shared by stdout and stderr, `fclose` of the read end before `waitpid` — differing only in whether the output is drained, run as root against the real `pfctl`:

```
################ CASE 1: table EXISTS, empty ################
A) without draining (current)      WIFEXITED=0 exit= -1 WIFSIGNALED=1 sig=13  -> table_exists=FALSE
B) with draining (fix)             WIFEXITED=1 exit=  0 WIFSIGNALED=0 sig= 0  -> table_exists=true

################ CASE 2: table EXISTS, 6000 entries ################
A) without draining (current)      WIFEXITED=0 exit= -1 WIFSIGNALED=1 sig=13  -> table_exists=FALSE
B) with draining (fix)             WIFEXITED=1 exit=  0 WIFSIGNALED=0 sig= 0  -> table_exists=true

################ CASE 3: table DOES NOT EXIST ################
A) without draining (current)      WIFEXITED=0 exit= -1 WIFSIGNALED=1 sig=13  -> table_exists=FALSE
B) with draining (fix)             WIFEXITED=1 exit=255 WIFSIGNALED=0 sig= 0  -> table_exists=FALSE
```

Signal 13 is `SIGPIPE`. Without the drain the check is wrong in every case, including the one where the table is plainly there.

**End to end with the patched agent,** `wazuh-agent_5.0.0-1_arm64_d616daf.pkg`, built from this branch at its first commit. With the table absent, `pf` declines, `/etc/pf.conf` is unchanged, the rule loaded from the other file survives and the table is not created:

```
[WARNING] Method=pf Action=check Details=wazuh_fwtable table not found
[WARNING] Method=hostsdeny Action=check Details=File '/etc/hosts.deny' not found
```

With the precondition applied by the administrator, `pf` blocks and unblocks as before, `/etc/pf.conf` untouched and the foreign rule intact — which is also the real-binary confirmation of the drain, since without it the check would have failed here too and the reload would have discarded that rule.

**The unblock, end to end,** with `wazuh-agent_5.0.0-1_arm64_825178e.pkg` — this branch merged into `5.0.0` — on a stock macOS 26.5.1 arm64 endpoint: `pf` disabled, no `/etc/hosts.deny`, no pre-existing routes. Blocking `192.0.2.66` falls through to the blackhole, as #39424 intended:

```
[INFO] Method=route Action=success Details=IP 192.0.2.66 successfully blocked

$ netstat -rn | grep 192.0.2.66
192.0.2.66         127.0.0.1          UGHSB                 lo0
```

The administrator then configures `pf` and the table, and the unblock arrives. This is the case that used to strand the route:

```
[INFO]    Method=pf        Action=skip          Details=address not in wazuh_fwtable
[WARNING] Method=pf        Action=invalid_state Details=Service or firewall in invalid state - trying next
[WARNING] Method=hostsdeny Action=check         Details=File '/etc/hosts.deny' not found
[INFO]    Method=route     Action=success       Details=IP 192.0.2.66 successfully unblocked

$ netstat -rn | grep 192.0.2.66
(nothing)
```

Two checks that the decline does not cost anything in the ordinary cases. With the table configured and the address in it, `pf` blocks, no route is created, and the unblock stops at `pf` with the table left empty — the chain does not walk past the method that holds the address. With `pf` enabled but the table absent, `pf` declines, `/etc/pf.conf` is byte-identical afterwards, the table is not created, and `route` takes over.

**The IPv6 path, end to end** on the same endpoint. Blocking `2001:db8::1` takes the `-inet6` branch and lands the blackhole with the matching-family gateway; the unblock removes it, and a second unblock is a no-op that still reports success, since `route` reports `not in table`:

```
[INFO] Method=route Action=success Details=IP 2001:db8::1 successfully blocked
$ netstat -rnf inet6 | grep 2001:db8
2001:db8::1                             ::1                                     UGHSB                 lo0

[INFO] Method=route Action=success Details=IP 2001:db8::1 successfully unblocked
$ netstat -rnf inet6 | grep 2001:db8
(nothing)
```

**`hosts.deny` declining, and leaving the file alone.** The address is blocked by `route` while no `/etc/hosts.deny` exists; the administrator then creates one, owned `root:wheel`, mode `600`, listing other addresses. The unblock walks past it:

```
[INFO]    Method=hostsdeny Action=skip          Details=address not listed in hosts.deny
[WARNING] Method=hostsdeny Action=invalid_state Details=Service or firewall in invalid state - trying next
[INFO]    Method=route     Action=success       Details=IP 192.0.2.66 successfully unblocked

before -> inodo=1496674 modo=-rw------- owner=root:wheel mtime=1790016390
after  -> inodo=1496674 modo=-rw------- owner=root:wheel mtime=1790016390
```

Same inode, mode, owner and mtime, and the blackhole is gone. The symmetric case still works: with the address actually listed, `hosts.deny` blocks and unblocks it, no route is created, the chain stops there, and the administrator's other two entries survive both operations.

One environment note for anyone repeating this: restoring `/etc/pf.conf` and reloading does **not** remove `wazuh_fwtable`. It is declared `persist`, so it outlives a ruleset that no longer mentions it and `pfctl -t wazuh_fwtable -T kill` is needed to get back to a table-less state. A first attempt at the third case measured nothing because of it.

**`route(8)` and `pfctl` semantics,** measured directly, each stream to its own file so nothing is inferred from interleaving:

```
### add OK ###              rc=0  stdout: add host 192.0.2.70: gateway 127.0.0.1        stderr: (empty)
### add duplicado ###       rc=0  stdout: ...: File exists          stderr: route: writing to routing socket: File exists
### delete OK ###           rc=0  stdout: delete host 192.0.2.70: gateway 127.0.0.1     stderr: (empty)
### delete inexistente ###  rc=0  stdout: ...: not in table         stderr: route: writing to routing socket: not in table
### add sin gateway ###     rc=0  stdout: ...: Invalid argument     stderr: route: writing to routing socket: Invalid argument
```

The exit status is 0 in all five, so it carries no information; stderr is empty exactly on success. Binding stderr alone therefore detects every failure reason rather than the two that were previously matched by name. The same run confirms two things that were until now only asserted: `-inet6` placed before the destination works for both `add` and `delete`, and `delete` accepts the gateway and `-blackhole`. Neither was a defect.

For `pf`, `pfctl -t wazuh_fwtable -T delete` exits 0 whether or not the address was there, and only the output distinguishes them — `0/1 addresses deleted.` against `1/1 addresses deleted.` — which is what the new `DISABLE` check keys off. That verdict is not the first thing in the pipe, which is why the decision reads all of it. Both streams merged into one pipe, unfiltered:

```
### -T delete, address absent ###
     1	No ALTQ support in kernel
     2	ALTQ related functions disabled
     3	0/1 addresses deleted.

### -T delete, address present ###
     1	No ALTQ support in kernel
     2	ALTQ related functions disabled
     3	1/1 addresses deleted.
```

`pfctl` writes the ALTQ notices to stderr, unbuffered, and the verdict to stdout, flushed at exit, so the line a caller needs is third. Keying off the first line would have read `No ALTQ support in kernel` and matched nothing.

  - [x] Linux — clean under `-Wall -Wextra` via the unit-test target; the two pre-existing `-Wformat-truncation=` warnings are gone, since the captured line is now bounded
  - [ ] Windows — N/A, not touched
  - [x] MAC OS X — `5_testcomponent_build-macos.yml` passed on this branch, and the package from its first commit was installed and exercised on a macOS 26.5.1 arm64 endpoint

- Memory tests for Linux — the unit-test build is ASan-instrumented and the target runs clean
- Memory tests for macOS — not run; no sanitizer build available for the macOS agent
- Endpoint coverage: IPv4 and IPv6, `pf` / `hosts.deny` / `route`, block and unblock, on a stock macOS 26.5.1 arm64 host

All of it has been run end to end; see the endpoint runs in Results and Evidence.

### Artifacts Affected

- `active-response/bin/block-ip` on macOS agent packages (arm64/x86_64).

**Behaviour changes for operators.** On an endpoint where `wazuh_fwtable` was never configured, `pf` declines instead of configuring itself; #39424's `route` blackhole means the address is still blocked. An endpoint that wants `pf` specifically now needs the table configured, and nothing in the macOS package says so — `AddPFTable()` runs at `.pkg` build time, and none of `preinstall.sh`, `postinstall.sh` or `introduction.txt` mention `pf` — so that belongs in the public documentation.

Unblocking now walks past methods that do not hold the address instead of stopping at the first one to claim success. An unblock for an address that is not blocked anywhere still ends in success, because `route` treats `not in table` as the requested end state.

Endpoints where an earlier agent already appended the block to `/etc/pf.conf` keep working: the table exists, so `pf` is used exactly as before. Those lines are not removed by this change; removing them is the administrator's call, since by now they may be relied upon.

### Configuration Changes

N/A for the agent's own configuration. The `wazuh_fwtable` setup is not new, but it is now genuinely required for the `pf` method rather than silently self-applied.

### Documentation Updates

`docs/ref/modules/active-response/executables.md`: the macOS section claimed `block-ip` requires a PF anchor named `wazuh_anchor` loaded from `/etc/pf.anchors/wazuh_anchor`. No code in the tree implements or reads an anchor — `pfctl -a` appears nowhere — so the page described a design that never shipped. It now carries the real precondition and states that a packaged install neither applies nor announces it, the new unblock behaviour, and that the `pfctl -k` connection kill is best effort. #39424's `Fallback` bullet was kept and extended with a missing table as a third reason for falling through.

### Tests Introduced

`src/unit_tests/active-response/test_block_ip_macos.c` gains 10 cmocka cases, the first coverage `try_pf_macos()` has had:

- a missing table declines, on both the block and the unblock path;
- the existence check's output is drained — asserted with `feof()` on the mocked stream, which is what pins the fix in place;
- the `pfctl -k` stream is drained too, which is what makes the connection kill actually run;
- a check killed by a signal is not read as "table present";
- a check that cannot be spawned is an execution failure, not a missing table;
- `DISABLE` declines when `pfctl` reports `0/1 addresses deleted.` and succeeds when it reports `1/1`, both fixtures carrying the ALTQ notices ahead of the verdict, as the endpoint emits them;
- `route` reporting a failure on stderr is a failure even though it exited 0, and a non-zero status with no output is a failure too;
- `route` is bound to stderr only, asserted through the new flags capture;
- an existing table still adds and deletes the address, with the command asserted through the argv capture.

The "must not touch `/etc/pf.conf`" half is enforced by cmocka rather than asserted: reaching the old path calls `access()` and `wpopenv()` again, and neither has a queued expectation, so the test fails.

A/B against the unpatched file: every new case fails, and the one that previously pinned the spawn-failure conflation had to be inverted. The suite runs 21/21, with `test_firewall_helpers` (5) and `test_active-response` (40) unchanged. `test_execd` (12) and `test_wm_docker` (3) were run as well, since the wrapper they share was touched.

**One gap, stated rather than papered over:** the `hosts.deny` path that declines when the address was not listed has no unit test. `try_hostsdeny_macos` does real file I/O and the harness can only fake the `access()` for a missing file; covering it would mean writing a real `/etc/hosts.deny`. It is verified by reading, not by test.

## Review Checklist

- [ ] Code changes reviewed


cborla (approved):
LGTM
- [x] Code changes reviewed

Issue #39318, opened 2026-09-15
Harden `CaCertificateSource`: keep the last good snapshot, bound the read, serialize read and publish, and stop describing `/cacerts` as it was before #39150


juliancnn:
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0 | wazuh-remoted | Manager | Packages/Sources | Linux |

## Description

`CaCertificateSource` (`src/remoted/remoted_module/src/http_server/caCertificateSource.cpp`) is the single reader behind `GET /cacerts` and the TLS status: it reads the CA bundle, hashes it, parses the certificates, re-serializes them and decides whether any of them signs the served leaf. It was introduced in #39150 to close two real holes (publishing private key material, and a coherence verdict up to a day stale), and it does that. Reading it for the spike turned up four defects around it, all verified in `5.0.0` @ `830917fd4a`:

1. **A transient read failure throws away the good snapshot.** When the file cannot be opened or read, the code clears the cached hash and replaces the snapshot with an empty one (`:127-132`). The endpoint answers `404 not_found` (`cacertsEndpoint.cpp:79-93`). Any I/O hiccup, permission change after an upgrade, or non-atomic edit of the bundle is therefore a window in which no agent can bootstrap trust from this node — for content the node was serving fine a second earlier and that is not sensitive.
2. **The size cap does not bound the allocation.** `readBounded()` reads the **whole** file into memory and only then compares its size with the 1 MiB cap (`:57-58`). A multi-gigabyte file at the configured path is loaded entirely before being rejected. The comment above the function says the cap is "the difference from the old per-request read"; it is not.
3. **The read happens outside the mutex, the publication inside.** `readBounded()` runs before `std::lock_guard lock {m_mutex}` (`:122-125`). Two concurrent calls can interleave so that the call that read the **older** file content acquires the lock last and overwrites the newer snapshot. Each snapshot is internally coherent; the sequence is not monotonic. Harmless while the file rarely changes; a real reordering once rotation makes changes routine.
[...]

Issue #39277, opened 2026-09-14
Spike: CA Certificate Rotation and Distribution to 5.x Agents


TomasTurina:
## Description

A 5.x agent bootstraps trust by fetching the manager's CA bundle once from `GET /cacerts` during enrollment and then pins it. **There is currently no mechanism to distribute a replacement CA.**

This creates a fleet-availability cliff. As `root-ca.pem` approaches expiry, there is no supported rotation path. Once it expires, every agent fails the TLS handshake against `remoted`. The same failure also prevents agents from reaching `/cacerts`, the only endpoint that could repair their trust store. Recovery therefore degrades to manually re-provisioning the trust anchor on every agent in the fleet.

This spike designs the **manager-side CA rotation mechanism** and defines the **agent-facing contract** so the agent team can run a parallel spike against a stable interface. **Implementation is out of scope.**

Most of the required building blocks already exist but are currently unconnected:

| Capability                             | Current state                                                                 | Reference                                                      |
| -------------------------------------- | ----------------------------------------------------------------------------- | -------------------------------------------------------------- |
| Multi-CA bundle serving                | Works. `/cacerts` parses every `CERTIFICATE` block and re-serializes them all | `cacertsEndpoint.cpp:77-113`, `tlsCertificateStatus.hpp:56-79` |
| Detect changed CA file without restart | Works. Content-hash revalidation occurs per request                           | `caCertificateSource.cpp:115-145`                              |
| Bundle-aware trust decision            | Works. `anyCaSignsLeaf()` accepts a match from any CA in the bundle           | `tlsCertificateStatus.hpp:114`                                 |
| "Go fetch, here is the hash" signal    | Exists for shared config only (`agent.config_hash`)                           | `controlHandler.cpp:511-543`                                   |
| Verify after download                  | Exists for shared config only (re-hash and discard on mismatch)               | `configFetcher.cpp:96-104`                                     |
| Certificate validity data              | Computed but mostly discarded; only two values are exposed as pull metrics    | `tlsCertificateStatus.hpp:163-174`                             |
| Rate limiting                          | **None.** The API contract currently states that it never returns `429`       | `agent-api.yaml:34-36`                                         |
| Listener leaf reload                   | **None.** The leaf certificate is loaded once when the listener starts        | `RestinioHttpServer.cpp` (start path)                          |

## Contract Changes

Two statements in the current contract must be deliberately reversed. They are documented as design intent rather than implementation accidents, so the spike must explicitly justify the changes.

[...]

PR #39370, opened 2026-09-16
Hardening cacertsource

juliancnn:
## Description

`CaCertificateSource` is the single reader behind `GET /cacerts` (the unauthenticated route that hands agents the CA that signs the HTTPS listener certificate) and behind the TLS status that `remoted.server.tls.*` and the legacy upgrade gate consume. Reading it for the CA rotation spike (#39277) turned up four defects, all present since #39150:

1. A transient read failure threw the snapshot being served away, so an I/O hiccup, a permission change after an upgrade or a non-atomic replacement of `root-ca.pem` answered `404` to every agent bootstrapping trust from that node.
2. The 1 MiB cap read the whole file into memory before comparing sizes.
3. The read ran outside the source's mutex while the publication ran inside it, so two concurrent callers could publish the older content last.
4. "Signs the leaf" was a raw `X509_verify`, so an expired CA, or one never marked `CA:TRUE`, still "matched" and was handed out although no verifying agent could build a chain from it.

The listener also read the CA a second time at start-up through an unbounded path, and the module README, the agent OpenAPI, the HTTPS API guide and the management API schema still described the endpoint as it was before #39150 (byte-for-byte serving, a daily coherence window, "never returns 429").

Closes #39318

## Proposed Changes

**`remoted_module` — the CA source (`src/http_server/caCertificateSource.*`, new `fileRead.*`)**
- Bounded, injectable POSIX reader (`readFileBounded()` / `FileReader`): never requests more than `kMaxBytes + 1` bytes, refuses anything that is not a regular file, reports the exact cause (`CannotOpen`, `ReadError`, `TooLarge`) with its `errno`; `describeReadFailure()` renders it for the logs.
- `snapshot()` runs whole under the source's mutex, so publication is monotonic by construction. The source keeps its own reference to the listener's leaf.
- A failed read keeps the last good snapshot and records the failure in it (`lastReadFailure`: cause, `errno`, consecutive count); identical bytes read again afterwards are still a cache hit. A readable file that carries no certificate (for example emptied) clears the snapshot at once — the operator's way to stop serving a CA without a restart.
[...]

PR #39465, opened 2026-09-18
CA rotation

### LINKED ISSUE #39319: Manager-side CA rotation: publication signal in `notify`, in-bundle publication metadata, and `wazuh-manager-certs`
## Status

PR #39465 (against `5.0.0`) delivers the remoted side of this issue: the `##` publication block
(parsed and vouched for by a new static library, `ca_bundle`, shared with the tool below), the
`ca_generation` field in `POST /control`'s `notify` response, the `Wazuh-CA-Generation` header on
every `GET /cacerts` `200`, a private on-disk record of the last bundle remoted served (so it can
tell "never stamped" from "changed outside the tool"), a descriptor cached and revalidated at most
once a second so the field costs no file read per notify, and the legacy 4.x delivery reworked to
hand a fresh-installed agent only the one certificate that signs this manager's active leaf, is
itself a CA, and is inside its validity — reserialized, without the block, since the 4.x installer
rejects more than one certificate.

PR #39472, stacked on top of it, delivers the rest: the `wazuh-manager-certs` binary (`inspect`/`check`/`add`/`remove`/`prune-expired`/`stamp`/`--from-master`) and the operator runbook (`docs/ref/modules/remoted/ca-rotation.md`). It must not be merged before #39465. Without the tool a bundle placed by hand keeps being served and is advertised as publication `0` on every node, exactly as designed — nothing regresses by merging the first PR alone.

## Corrected anchors

Several citations in the description below point at lines that moved, some of them before this
branch existed. Verified against `enhancement/39319-ca-rotation`:

| Original citation | What changed |
|---|---|
| `caCertificateSource.cpp:115-142` — `snapshot()` reads and hashes the file on every call | Split three ways: `snapshot()` is now `:189-202` (still whole-file, still under the lock, used by the endpoint and the monitor); the per-notify path no longer takes it at all — it reads `descriptor()` (`:205-238`), a cache revalidated at most once a second (`kDescriptorRefresh`, `caCertificateSource.hpp:136`); both delegate to `buildLocked()` (`:90-146`), which is where the `##` block is now parsed and vouched for. |
| `controlHandler.cpp:522-527` — `notify` carries only `config_hash`/`settings_hash` | No longer the whole field list: `:524-556` builds the response, and `ca_generation` is set at `:541-544` (`response["ca_generation"] = generation ? json(*generation) : json(nullptr)`), right after `settings_hash`. |
| `docs/ref/modules/remoted/configuration.md:956-958` — cited as the keepalive throttle | Already stale before this branch: that range is `downstream_response_timeout` territory (heading at `:971`); the actual `remoted.control_keepalive_throttle` section is at `:1077-1083`. Unrelated to this work; flagging it because the description still points at the old lines. |
| `caCertificateSource.cpp:127-132` (last-good snapshot on read failure), `:49-60` (bounded read), `:122-125` (read outside the mutex); `tlsCertificateStatus.cpp:211-227` (direct-signature-only trust check) — listed under "Pre-existing defects" | **Already resolved, by #39318, before this branch was created** — these were filed and fixed ahead of this work on purpose (§6 of the original description says as much). `anyCaSignsLeaf()` has since moved again, into the new `ca_bundle` library (`src/shared_modules/ca_bundle/src/ca_bundle.cpp:386`, declared in `ca_bundle.hpp:150`); `tlsCertificateStatus.hpp:60` re-exports it (`using ca_bundle::anyCaSignsLeaf;`) so the two remaining callers in that file don't change. |
| `README.md:256-271` / `agent-api.yaml:97-119` — "describe `/cacerts` as it was before #39150" | Already corrected independently of this issue. This branch's own PR extends `agent-api.yaml` further, documenting `ca_generation` itself and its four states at `:406-485`. |
| `wazuh-certs-tool.sh:465` — cited as where the installer's cert tool concatenates the CA, offered as precedent for a block appearing mid-file | The file is `src/engine/tools/devContainer/scripts/wazuh-certs-tool.sh`; `:465` is a comment, the actual command is `:479` (`cat ".../root-ca.pem" >> ".../<node>-remoted.pem"`) — and it appends the CA **after the listener's leaf file**, not inside `root-ca.pem`. `root-ca.pem` itself is never concatenated by this tool; the precedent still holds as an example of OpenSSL tolerating explanatory text, just not of the file this issue's bundle format actually describes. |
| `https_client.h:563` — `HC_MAX_CACERTS_BODY 8192` | Line moved to `:569`; value unchanged. |

## Decisions from the spike/design round an implementer needs

- **C7** — A legacy 4.x upgrade receives only the one certificate that signs this manager's
  currently served leaf, is itself a CA, and is inside its validity window — reserialized, without
  the `##` block — because `src/init/pkg_installer.sh:439-444` rejects a PEM carrying more than one
  certificate. Sharpened after review: a same-key reissue can produce two certificates that both
  sign the leaf, one of them expired; the export refuses to hand out an anchor the 4.x installer
  would just throw away, and delivers nothing rather than the wrong one.
- **C8** — `ca_generation` is served from a descriptor cached inside `CaCertificateSource` and
  revalidated at most once per second, never from a fresh file read on every `notify`.
- **C9** — The `##` block parser, the DER content hash, and the vouching guards live in a new
  static library, `ca_bundle` (`src/shared_modules/ca_bundle/`), linked by `remoted_module` today
  and, in the second PR, by `wazuh-manager-certs` as well — one implementation of the hash, not two.
- **C10** — `--from-master` (second PR) will talk libcurl directly, with `CURLOPT_HEADERFUNCTION` to
  read `Wazuh-CA-Generation` and `CAINFO` pinned to the local bundle, rather than extend the shared
  `wazuh-http-request` submodule.

Invariants that hold across both PRs: no hash of the bundle or of its certificates ever reaches an
agent, in `notify` or in any `GET /cacerts` header. remoted never writes the bundle file. Only
`wazuh-manager-certs` (second PR) publishes. A bundle's publication is a Unix timestamp that only
that tool ever assigns — `0` means a bundle is being served but nothing vouches for it yet, `null`
means there is no servable bundle at all.

<details>
<summary>Original issue body (archived 2026-09-18)</summary>

|---|---|---|---|---|
| 5.0.0 | wazuh-remoted, new `wazuh-manager-certs` | Manager | Packages/Sources | Linux |

## Description

**Related**: #39129 / PR #39280 (rate limit on `/enroll` and `/cacerts`), #39078 / PR #39150 (where `CaCertificateSource` and the current `/cacerts` semantics come from).

A 5.x agent fetches the manager's CA bundle once from `GET /cacerts` at enrollment and pins it. There is no supported way to hand it a replacement CA, so when `root-ca.pem` approaches expiry the only recovery is re-provisioning the trust anchor on every agent by hand. Spike #39277 designed the manager side of a rotation mechanism; this issue implements it.

The design rests on three facts verified in `5.0.0` @ `830917fd4a`:

- **The CA already hot-reloads.** `CaCertificateSource::snapshot()` reads the file and hashes it on every call, re-parsing only when the content changed (`src/remoted/remoted_module/src/http_server/caCertificateSource.cpp:115-142`). Adding a CA to the bundle needs no restart and no signal. Only the listener leaf is fixed until restart (`RestinioHttpServer.cpp:1369`, loaded inside `start()`).
- **The signal channel exists for shared configuration only.** The `notify` response carries `agent.config_hash` and `settings_hash` (`src/remoted/remoted_module/src/control/controlHandler.cpp:522-527`) and nothing about the CA.
- **The bundle path is configuration, not a product artifact.** `remote.https.ca_certificate` accepts an absolute path or one relative to the manager home (schema `path` definition), while cluster synchronization walks fixed relative paths with explicit file lists (`framework/wazuh/core/cluster/cluster.py:114`; every entry in `cluster.json` today — `client.keys`, `authd.pass`, `enrollment_tokens.json` — is non-configurable). The bundle therefore cannot ride the cluster sync, and distribution has to be pull-based.

Operators will often install their own root CA as a plain PEM and add others later. The design treats that as the normal case: such a bundle is served and advertised as publication `0`, which no agent adopts, with an INFO line saying how to publish it. **Only the tool publishes**; remoted never writes the bundle.

## Evidence

```text
# CA revalidated by content hash on every read, no restart needed
$ sed -n '115,142p' src/remoted/remoted_module/src/http_server/caCertificateSource.cpp
    CaCertificateSnapshot CaCertificateSource::snapshot()
    ...
        const bool read = readBounded(m_path, kMaxBytes, contents);
        std::lock_guard<std::mutex> lock {m_mutex};
        ...
        auto hash = digestOf(contents);
        if (hash == m_hash) { return m_snapshot; }
        m_snapshot = buildLocked(contents);

# The notify response knows nothing about the CA
$ grep -n 'config_hash\|settings_hash' src/remoted/remoted_module/src/control/controlHandler.cpp
522:                    response["agent"]["config_hash"] = configHash;
527:                    response["settings_hash"] = m_hashCache->getSettingsHash();

# The agent caps the /cacerts body at 8 KiB and flags truncation
$ grep -n 'define HC_MAX_CACERTS_BODY' src/client-agent/https_client/include/https_client.h
563:#define HC_MAX_CACERTS_BODY 8192

# Notify cadence is 10 s; the keepalive throttle only spaces wazuh-db writes, not response construction
$ grep -n 'define NOTIFY_TIME' src/shared/include/defs.h
91:#define NOTIFY_TIME    10 // ... every 10 seconds

# root-ca.pem is root-owned and remoted (wazuh-manager) does not write it; that stays so
$ ls -la /var/wazuh-manager/etc/certs/
drwxrwx--T 2 root          wazuh-manager 4096 .
-rw-r----- 1 wazuh-manager wazuh-manager 1704 remoted-key.pem
-rw-r----- 1 wazuh-manager wazuh-manager 2599 remoted.pem
-rw-r----- 1 root          wazuh-manager 1204 root-ca.pem
$ grep -n 'for CERT_FILE in' src/init/inst-functions.sh packages/debs/SPECS/wazuh-manager/debian/postinst packages/rpms/SPECS/wazuh-manager.spec
src/init/inst-functions.sh:199:    for CERT_FILE in remoted.pem remoted-key.pem; do
src/init/inst-functions.sh:237:    for CERT_FILE in root-ca.pem indexer-connector.pem indexer-connector-key.pem; do
packages/debs/SPECS/wazuh-manager/debian/postinst:162:    for CERT_FILE in remoted.pem remoted-key.pem apid.pem apid-key.pem; do
packages/debs/SPECS/wazuh-manager/debian/postinst:171:    for CERT_FILE in root-ca.pem indexer-connector.pem indexer-connector-key.pem; do
packages/rpms/SPECS/wazuh-manager.spec:371:for CERT_FILE in remoted.pem remoted-key.pem apid.pem apid-key.pem; do
packages/rpms/SPECS/wazuh-manager.spec:380:for CERT_FILE in root-ca.pem indexer-connector.pem indexer-connector-key.pem; do
```

A bundle carrying a `##` comment block before the first certificate loads with `openssl x509`, `openssl verify -CAfile` (as the trust store), `openssl storeutl` (OpenSSL 3's strict loader), `openssl crl2pkcs7`, and Python's `ssl` module; the repository's own `wazuh-certs-tool.sh` reads `root-ca.pem` through `openssl x509 -CA`, the same tolerant parser. The curl/Mozilla root bundle shipped by `certifi` uses exactly this convention (a `##` banner with date, tool version and `SHA256:`, plus per-certificate comments).

### 1. Bundle format: publication metadata inside the file, as RFC 7468 explanatory text

The publication and a content hash live in a `##` comment block **before the first certificate**, so one file is the complete publication and copying it is distributing it:

```
##
## Wazuh CA bundle
##
## Publication: 1789423200
## Content-SHA256: <hex>
## Updated: 2026-09-15T10:00:00Z
## Written by: wazuh-manager-certs 5.0.0
##
-----BEGIN CERTIFICATE-----
```

- RFC 7468 §2: *"Data before the encapsulation boundaries are permitted, and parsers MUST NOT malfunction when processing such data."* §5.2 calls this explanatory text and asks that it relate to the certificate; a publication and a hash of the certificates do. The same RFC forbids headers **inside** `BEGIN`/`END`, so the block goes before the first certificate and nowhere else.
- `Publication` is a **Unix timestamp** (seconds), assigned by the tool on every write as `max(now(), previous + 1)`. It is higher than any past publication by construction, so losing the block (an external regeneration of the file) or rebuilding the master never loses the sequence. The tool refuses when `now() <= previous` (clock set back). `0` means "unpublished" and is what remoted advertises for a bundle without a valid block. `Updated` is the same instant in RFC 3339, for humans. `Written by` names the tool and its version.
- `Content-SHA256` is SHA-256 over the concatenation of the **DER encodings of the certificates, sorted by their DER bytes**. It is independent of PEM formatting and file order. **It never leaves the manager**: it exists so remoted can tell whether the block still describes the certificates in the file. No hash reaches the agent (spike D15, D16).
- The tool tolerates its block at any position when reading (the installer's cert tool concatenates `root-ca.pem` behind the leaf, `wazuh-certs-tool.sh:465`, so the block can end up mid-file in a chain) and only ever writes it at the top.

### 2. remoted: vouch, remember, advertise

**Vouching.** `CaCertificateSource` parses the `##` block alongside the certificates, recomputes the DER hash over the parsed certificates and compares it with `Content-SHA256`. The block **vouches** for the publication only when they match.

**Advertising.** `POST /control` `notify` response gains **one** top-level field, `ca_generation`: a 64-bit integer, the vouched publication; `0` when the node serves a bundle without a valid block (unpublished); `null` when there is no servable bundle. Absent means a manager older than this feature. `GET /cacerts` gains **one** response header on every `200`: `Wazuh-CA-Generation` (the publication, or `0`). The body stays certificates only, re-serialized, exactly as today. No hash header.

**Unpublished bundles and log levels (spike D17, D18).** remoted **never writes the bundle**; the tool is the only publisher. A bundle without a valid block is served and advertised as `0`. To tell "first time" from "replaced outside the tool" — both are a plain PEM — remoted keeps a small private, persistent record `{path, file hash, publication}` of the last bundle it served, in its own runtime directory (for example under `var/run/`). It is never distributed and is not the source of truth. On each read:

| What it reads | Record | Advertises | Log |
|---|---|---|---|
| Valid block (its hash matches the parsed certificates) | updated | the block's publication | INFO when the publication changed |
| No valid block; no record for this path | created with the hash and `0` | `0` | **INFO**: "unpublished CA bundle at <path>; publish it with `wazuh-manager-certs stamp` or `add` on the master" |
| No valid block; same hash as the record | unchanged | `0` | nothing |
| No valid block; hash differs from the record | updated | `0` | **WARN**: "CA bundle at <path> changed outside the tool and is not published; previous publication N; run `stamp` on the master" |
| Block present but its `Content-SHA256` does not describe the certificates | updated | `0` | **WARN**, as above |

Vouching also requires the guards the tool applies: at least one certificate; one of them signs the leaf the listener is serving (the check that already decides `503 ca_mismatch`); at most 6 certificates; at most 8191 serialized bytes. A failing guard means `0`, content still served, and a WARN naming the guard.

Because `0` depends on nothing local, every node serving the same unpublished file advertises the same value without knowing anything about the others. A bundle at `0` moves no agent: those on a publication stay on it, freshly enrolled ones live on the CA they pinned at bootstrap. `stamp` on the master writes a timestamp greater than `0` and than any previous publication, and the fleet adopts. On a worker the only way out of `0` is to realign with the master (`--from-master`, or copying the published file); the worker tool refuses to publish (§3).

The file stays root-owned (`root:wazuh-manager 640`, `inst-functions.sh:237`, deb `postinst:171`, rpm `spec:380`) and remoted, running as `wazuh-manager` (`src/remoted/src/main.c:71`), keeps no write access to the trust anchor. No packaging change.

**Hot path.** The published value must be cached and revalidated by content, not recomputed per notify: `snapshot()` costs a file read plus a SHA-256 per call, `/control` runs at a 10 s default cadence (`NOTIFY_TIME`), and `remoted.control_keepalive_throttle` only spaces out wazuh-db writes, not response construction (`docs/ref/modules/remoted/configuration.md:956-958`). At 10 000 agents that is on the order of 1 000 reads and hashes per second for a value that changes once a quarter. Follow the `HashCache` pattern (`src/remoted/remoted_module/src/control/hashCache.cpp:163-192`) with content-based invalidation. `controlHandler` has no reference to the CA source today (`src/control/` vs `src/http_server/`); inject the descriptor provider from the facade.

### 3. `wazuh-manager-certs`: a new binary that owns the guarded path

Purely local operations, so it works with every daemon down (it is a repair tool). It is **the only publisher**: a bundle that has not gone through it is served but advertised as `0`. Commands:

- `inspect` — subject, issuer, `notAfter`, days remaining, `x509-sha256:<hex>` identity, the publication and whether the block vouches for it, and whether each certificate signs the leaf the listener is **currently serving**.
- `check` — validate without writing: parse, dates, CA capabilities, coherence with the active leaf, the limits below.
- `add <file>` — one or more certificates; rejects duplicates by identity and rejects exceeding the limits.
- `remove <identity>` — by `x509-sha256` identity, never by position or subject.
- `prune-expired` — subject to the leaf guard.
- `stamp` — publish the **current** content as is: write the `##` block with a higher publication, after every guard. The first step for a bundle placed by hand, and the remedy to the WARN after an external replacement.
- `--from-master` — see §4.

Rules every writing command follows:

- **Resolve the bundle path from the configuration** (`remote.https.ca_certificate`, absolute or relative to the manager home). Never assume `etc/certs/root-ca.pem`.
- **Only the master publishes.** When the local `cluster.node_type` is `worker`, `add`, `remove`, `prune-expired` and `stamp` refuse with a message pointing at the master and at `--from-master`. Workers copy the file or pull it.
- **Atomic write**: temp file in the same directory, `fsync`, `rename`, `fsync` of the directory. Preserve owner and mode (`root:wazuh-manager 640`; the tool runs as root). Never append to or truncate the live file. Exclusive lock stable across the `rename`.
- **Leaf guard**: refuse, except with `--force`, any operation that would leave the bundle without a CA that signs the leaf the listener is currently serving (read from `remote.https.certificate`). This applies to `prune-expired` too: a leaf signed by an expired CA is exactly the emergency in which the operator runs it.
- **Limits, both, whichever binds first**: at most **6 certificates**, and a serialized body budget derived from the agent's `HC_MAX_CACERTS_BODY` (8191 usable bytes). Measured with short subjects: RSA-2048 CAs are ~1224 B (6 fit), RSA-3072 ~1570 B (5 fit), RSA-4096 ~1919 B (4 fit), EC P-256 ~696 B (11 fit). The tool refuses either limit and **says which**, with the figures, because the two lead to different operator actions.
- Every write assigns `Publication = max(now(), previous + 1)` and rewrites `Content-SHA256` in the same file write.

### 4. `--from-master`: pull-based distribution

Replaces cluster synchronization (see *Out of scope*).

- Composes the URL from the **local** configuration: master address from `cluster.nodes`, local `remote.https.port` (default `1517`) and local `remote.https.global_prefix` (default `/wazuh-manager/`), since nodes normally share both. No arguments in the common case; `--port` to override.
- **TLS verification against the local bundle is mandatory and has no bypass flag.** A skip flag would become the normal way the tool is used and would rebuild the unverified-bootstrap hole the agent client already has (`hc_fetch_cacerts()` forces `HC_VERIFY_NONE`, `src/client-agent/https_client/src/hcInterface.cpp:370`). Verification works throughout the overlap window because the master's leaf is still signed by the old CA until the reissue step. `--from-master` is for updating a node that already has a trust anchor, not for provisioning a new one.
- Every check happens **before** the original is touched, in this order: verified download; `Wazuh-CA-Generation` must be greater than the local publication (equal → idempotent no-op that prints the state; `0` → refuse, telling the operator the master's bundle is not vouched); parse; both limits; leaf guard against the **local** leaf; then the atomic write with the `##` block rebuilt from the header's publication and a `Content-SHA256` **computed locally** over the downloaded certificates.
- On connection failure, or when the master does not share port or prefix, fail naming the cause and suggest `--port` and the manual copy (`scp` the published file, then `check`), which remains a fully valid alternative. Copying an **unpublished** file makes every node advertise `0`: nothing moves until the master publishes it with `stamp` and it is distributed again.

### 5. Legacy 4.x delivery: a risk to confirm

The CA delivery to 4.x agents during a remote upgrade sends the bundle file's **raw bytes** (`src/remoted/src/legacy_task_delivery.c:641-653`, cap 64 KiB at `:138`), and its comment says the 4.x installer "pins whatever it finds" (`:666`). The `##` block will travel inside that file. OpenSSL tolerates leading text, but the 4.x pinning path must be verified to tolerate it, or the delivery should send re-serialized certificates instead — which would also close the same raw-bytes exposure #39078 fixed on `/cacerts`. Track with the 4.x team before shipping.

### 6. Pre-existing defects in the same code path

Verified in the spike, none mentioned in #39277, all on the rotation path. Filed as #39318 so they land ahead of this feature: last-good snapshot on read failure (`caCertificateSource.cpp:127-132`), bounded read (`:49-60`), read outside the mutex (`:122-125`), direct-signature-only trust check (`tlsCertificateStatus.cpp:211-227`).

### 7. Documentation

- Correct `src/remoted/remoted_module/README.md:256-271` and `docs/ref/modules/remoted/agent-api.yaml:97-119`. Both describe `/cacerts` as it was before #39150. Precisely: "reads the PEM file on every request" is still **true**; what is stale is "no cache to invalidate" (there is a parse cache), "the file byte for byte" (it is re-serialized), and the daily-incoherence window (the coherence verdict now comes from the same read).
- Rewrite the `MUST NOT replace it from this route` statement in `agent-api.yaml:117-119` to distinguish bootstrap (unverified channel, strict pinned-certificate selection unchanged) from refresh (TLS validated against the trusted CA, advertised publication, header check, wholesale replacement).
- Operator runbook under `docs/ref/modules/remoted/`: add the new CA with `add`, or place the file and publish it with `stamp` (no restart either way; until `stamp` the bundle is advertised as `0`); distribute with `--from-master` or by copying the published file (no restart); wait out the overlap; reissue leafs under the new CA and restart remoted (the one restart); remove the old CA with `remove`. **The overlap window must exceed the deployment's maximum tolerated agent disconnection**; there is no measured adoption gate. Note that the rotation burst is paced by `remote.https.cacerts_rate_limit` (#39280, default 50 req/s): at 10 000 agents the fleet adopts in roughly 200 s; an operator may raise it for the rotation and restore it afterwards. Document what `0`, INFO and WARN mean and what to do on a WARN.
- `CHANGELOG.md` entry.

### Contract handed to the agent team (not implemented here)

Written as part of this issue so their spike has a stable interface; implemented by them (#39321):

- Persist the publication **with** the installed bundle in one atomic write (the `##` block format works for this); unknown when absent.
- Compare `ca_generation` from `notify` with the persisted one: `null` or `0` → ignore; lower or equal → ignore; higher → fetch `/cacerts` over TLS verified against the current store, require `Wazuh-CA-Generation` to equal the advertised publication, require a parseable, non-truncated body, **replace the trust store wholesale**, persist.
- Bootstrap keeps installing **only** the certificate the enrollment-token pin names (`src/client-agent/src/token_bootstrap.c:261-274`); wholesale replacement is a refresh-path behaviour only.
- Backoff with jitter on every non-`200` from `/cacerts`, `429` included.
- No new field in the `notify` request. No hash anywhere.

## Out of scope

Decided in the spike, each for a stated reason:

- **Authentication on `/cacerts`.** The bundle is public material and the rate limit is uniform, so there is nobody to distinguish (spike D1).
- **Rate limiting and `429`.** Done in #39129 / PR #39280.
- **Listener leaf hot reload** (D3). The one restart in the runbook stays.
- **Cluster synchronization of the bundle** (D7): the path is configuration, and the publication already makes a partially rotated cluster safe. If wanted later, it comes back as a fixed-path product artifact.
- **Per-agent adoption state**, in `wazuh-db` or in memory, and the agent echoing its publication in `notify` (D8, D12). The retirement step is governed by the temporal rule above.
- **Any hash outside the bundle file** (D15, D16): none in `notify`, none in `/cacerts` headers, none on the agent.
- **The "pin-set" reading of #39277's Decision 8.** As worded there it invites a vulnerability: a bundle `[genuine CA, attacker CA]` satisfies the pin on its genuine half, and installing the whole bundle would make the attacker's half a trust anchor. The agent already documents and avoids exactly this attack. Decision 8 must be rewritten in the spike issue; nothing here relaxes bootstrap pinning.
- Legacy 4.x `ca_delivery` rotation (beyond the compatibility check in §5); agent client-certificate rotation; automatic CA generation or renewal.

## Acceptance criteria

- [ ] A plain PEM placed at the configured path on a fresh node is served, advertised as `0` in `notify` and in the header, and logged once at INFO with the remedy; the file is left untouched. After `stamp` on the master it is advertised with the block's timestamp; `openssl storeutl -certs`, `openssl verify -CAfile` and Python `ssl` load the sealed file unchanged; `GET /cacerts` serves certificates only.
- [ ] `notify` carries `ca_generation` (a Unix timestamp, `0` for an unpublished bundle, `null` for none); `GET /cacerts` carries `Wazuh-CA-Generation` on every `200`, `0` when nothing vouches, and no hash header; the body is still served in the `0` case.
- [ ] Replacing a published bundle's content with a plain PEM, on any node, logs one WARN naming the previous publication, advertises `0` and keeps serving; agents stay on their publication; `stamp` on the master (or `--from-master` on a worker) restores a published state. The same unpublished file copied to three nodes is advertised as `0` on all three.
- [ ] Stripping only the `##` block from a published bundle turns its advertisement to `0` with a WARN (the hash of the file changed); `stamp` republishes it with a higher timestamp.
- [ ] A bundle failing a vouching guard (no CA signs the active leaf, 7 certificates, or over 8191 bytes) is advertised as `0` and the WARN names the guard.
- [ ] Two consecutive `add` or `stamp` runs produce strictly increasing publications even within the same second; a run with the clock set back before the current publication is refused.
- [ ] The per-notify cost of the new field does not include a file read: a fleet-scale `control_notify_storm` run (`tools/manager_benchmark`) shows no regression in `/control` latency.
- [ ] `add` and `prune-expired` refuse to leave the bundle without a CA signing the active leaf; the 7th certificate is refused naming the count, and a bundle over the byte budget is refused naming the bytes.
- [ ] `--from-master` on a worker with no arguments fetches from `cluster.nodes` with the local port and prefix, verifies TLS against the local bundle, refuses a `0` publication and an equal one (no-op), writes a block whose `Content-SHA256` remoted then vouches for, and never touches the original before every check passes. There is no flag that disables verification.
- [ ] README and `agent-api.yaml` no longer claim byte-for-byte serving or a daily coherence window; the runbook exists and states the overlap rule and the meaning of INFO and WARN.
- [ ] The 4.x pinning path is confirmed to tolerate a sealed bundle, or the legacy delivery sends re-serialized certificates.
- [ ] Lab run: two CAs, overlap, leaf reissue, retirement, agents online throughout; failure modes: expired CA, `503 ca_mismatch`, `404 not_found`, truncated bundle, agent offline through the window, plain-PEM replacement with and without a subsequent `stamp`.

## References

- Spike #39277 and its progress comments (design decisions D1–D18).
- #39129 / PR #39280 — rate limit on `/enroll` and `/cacerts`.
- #39078 / PR #39150 — origin of `CaCertificateSource` and the re-serialized `/cacerts` body.
- RFC 7468 §2 and §5.2 (textual encoding; explanatory text before `BEGIN`).
- curl/Mozilla CA bundle format (`https://curl.se/ca/cacert.pem`), the precedent for a `##` metadata banner with a `SHA256:` line.

</details>

juliancnn:
## Description

`CaCertificateSource` is the single reader behind `GET /cacerts` (the unauthenticated route that hands agents the CA that signs the HTTPS listener certificate) and behind the TLS status that `remoted.server.tls.*` and the legacy upgrade gate consume. Reading it for the CA rotation spike (#39277) turned up four defects, all present since #39150:

1. A transient read failure threw the snapshot being served away, so an I/O hiccup, a permission change after an upgrade or a non-atomic replacement of `root-ca.pem` answered `404` to every agent bootstrapping trust from that node.
2. The 1 MiB cap read the whole file into memory before comparing sizes.
3. The read ran outside the source's mutex while the publication ran inside it, so two concurrent callers could publish the older content last.
4. "Signs the leaf" was a raw `X509_verify`, so an expired CA, or one never marked `CA:TRUE`, still "matched" and was handed out although no verifying agent could build a chain from it.

The listener also read the CA a second time at start-up through an unbounded path, and the module README, the agent OpenAPI, the HTTPS API guide and the management API schema still described the endpoint as it was before #39150 (byte-for-byte serving, a daily coherence window, "never returns 429").

Closes #39318

## Proposed Changes

**`remoted_module` — the CA source (`src/http_server/caCertificateSource.*`, new `fileRead.*`)**
- Bounded, injectable POSIX reader (`readFileBounded()` / `FileReader`): never requests more than `kMaxBytes + 1` bytes, refuses anything that is not a regular file, reports the exact cause (`CannotOpen`, `ReadError`, `TooLarge`) with its `errno`; `describeReadFailure()` renders it for the logs.
- `snapshot()` runs whole under the source's mutex, so publication is monotonic by construction. The source keeps its own reference to the listener's leaf.
- A failed read keeps the last good snapshot and records the failure in it (`lastReadFailure`: cause, `errno`, consecutive count); identical bytes read again afterwards are still a cache hit. A readable file that carries no certificate (for example emptied) clears the snapshot at once — the operator's way to stop serving a CA without a restart.
- `chainValidates()`: whether the served leaf validates with the bundle as its **only** trust store (`X509_STORE_CTX`, `X509_V_FLAG_PARTIAL_CHAIN` so a purchased intermediate that signed the leaf is a valid anchor, SSL-server purpose, dates and CA constraints). The verdict and OpenSSL's reason travel in `CaCertificateSnapshot` and `TlsCertificateSnapshot` next to `matchesLeaf`. Information only: the `503 ca_mismatch` decision, the C export `remoted_module_tls_ca_matches_leaf()` and the metrics are unchanged.
- The legacy delivery that picks the trust anchor for an upgrading 4.x agent asked `ca_bundle`'s `isCa`, which comes from `X509_check_ca()` — true for a root carrying only `keyCertSign`, with no `basicConstraints` extension at all. The agent's own installer (`pkg_installer.sh`) greps for that exact extension with a literal `CA:TRUE` and discards the file when it is missing, so the agent finished the upgrade with no anchor: the same failure this narrowed delivery exists to prevent, arriving through the one check it was not making. Worse, the search stopped at the first certificate that passed, so a bundle whose first signer lacked the extension lost a later one the installer would have accepted. The code's own comment already promised this exact parity with the installer; it now reads the extension where it makes that claim, and a certificate without it is skipped rather than ending the search. `describe()`'s `isCa` is untouched — 5.x agents rely on OpenSSL's own chain building, where it is still the right question.

**`remoted_module` — endpoint and transport (`endpoints/cacertsEndpoint.cpp`, `http_server/RestinioHttpServer.cpp`, `tlsCertificateStatus.*`, `IHttpServer.hpp`)**
- `GET /cacerts` emits one throttled WARN per window while it answers from the last good read, naming the cause and how many reads failed in a row; its `404` names the cause. The route's three log throttles now belong to the handler instead of being process-wide statics.
- The start-up evaluation builds the `CaCertificateSource` inside `createTlsContext()` and derives the status through `statusFrom()`, the same helper the daily monitor tick uses; `evaluateCertificateStatus()` and `loadCertificates()` are gone from production.
- The certificate log lines (start and every 24 h) carry the read failure and the chain verdict: a WARN when the bundle signs the leaf but the chain does not validate, an INFO for the converse.

**Documentation**
- `src/remoted/remoted_module/README.md`, `docs/ref/modules/remoted/agent-api.yaml`, `docs/ref/modules/remoted/https-events-api.md`, `api/api/spec/spec.yaml`: the endpoint and `remoted.server.tls.*` described as they are (re-serialised certificates, verdict from the same read as the body, `429` only from the two credential-less routes, `cert_expiry_days` from the monitor vs `ca_matches_leaf` from the last read). A stale comment in `remotedModuleFacade.hpp` corrected.
- `CHANGELOG.md`: one `Fixed` row for #39318.

#### Live verification against an installed manager

The manager was purged and reinstalled from this branch (`install.sh`, seven daemons started, certificates provisioned with the installation assistant's tool) and `GET /cacerts` was exercised against the real listener at `https://127.0.0.1:1517/wazuh-manager/cacerts`.

**The defect first reproduced on the same host with the previously installed binary** (built before this branch):

| Step | Before (binary of 2026-09-10) | After (this branch) |
|---|---|---|
| `GET /cacerts` | `200` | `200` |
| `mv root-ca.pem root-ca.pem.off`, then `GET /cacerts` | **`404 {"error":"not_found"}`** | **`200`, body byte-identical to the previous one** |
| what the log said | `GET /cacerts answered 404 …: the configured CA certificate is missing, unreadable, too large or carries no usable certificate` | `GET /cacerts is answering 1 request(s) in the last 90 s from the last good read of the configured CA certificate, which now cannot be opened (No such file or directory) (1 consecutive failed read(s)); restore the file -- the bundle being served predates the failure.` |

Matrix against the running manager — **14 checks, 14 passed**:

| # | Situation | Result |
|---|---|---|
| 1 | Baseline | `200`, 1204 bytes, 1 `CERTIFICATE` block, 0 `PRIVATE KEY` blocks |
| 2 | The CA file also carries the CA's private key | `200`, still certificates only (no key in the body) |
| 3 | CA file moved away | `200`, body identical to the baseline + one throttled WARN naming the cause |
| 4 | CA file restored | `200` |
| 5 | CA file emptied (readable, no certificate) | `404 {"error":"not_found"}` at once + WARN `carries no usable certificate` |
| 6 | A directory at the CA path, while a good bundle was being served | `200` from the last good read |
| 7 | A 2 MiB file at the CA path, while a good bundle was being served | `200` from the last good read |
| 8 | A valid CA that does not sign the served certificate | `503 {"error":"ca_mismatch"}` — decision unchanged |

Second pass, with the manager **started** with each misconfiguration already in place, so nothing had ever been served and the start-up evaluation (never throttled) names the cause:

```log
# a directory at the CA path
wazuh-manager-remoted:http-server: WARNING: CA certificate 'etc/certs/root-ca.pem' cannot be read (Is a directory); GET /cacerts will answer 404 until it is restored.
wazuh-manager-remoted:endpoints(cacerts): WARNING: GET /cacerts answered 404 to 1 request(s) in the last 90 s: the configured CA certificate cannot be read (Is a directory) (2 consecutive failed read(s)); agents cannot bootstrap trust from this manager until it is restored.

# a 2 MiB file at the CA path
wazuh-manager-remoted:http-server: WARNING: CA certificate 'etc/certs/root-ca.pem' is larger than the 1 MiB cap; GET /cacerts will answer 404 until it is restored.
wazuh-manager-remoted:endpoints(cacerts): WARNING: GET /cacerts answered 404 to 1 request(s) in the last 90 s: the configured CA certificate is larger than the 1 MiB cap (2 consecutive failed read(s)); agents cannot bootstrap trust from this manager until it is restored.

# a CA that signs the served certificate but carries basicConstraints CA:FALSE
wazuh-manager-remoted:http-server: WARNING: The configured CA 'etc/certs/root-ca.pem' signs the served certificate 'etc/certs/remoted.pem' but the chain does not validate (invalid CA certificate); a verifying agent that pins this CA will reject the handshake -- check the validity dates and the CA constraints of the certificates involved.
```

That last case is the one the chain verdict exists for: the route still answers `200` and `remoted.server.tls.ca_matches_leaf` still reads `1` (the CA does sign the leaf), while the log tells the operator that no verifying agent could use it. The `503` decision did not move.

**Trust bootstrap, end to end**: a client trusting *only* the PEM the endpoint just served completes the TLS handshake against that same listener — `curl --cacert <served PEM> https://127.0.0.1:1517/wazuh-manager/` → `ssl_verify_result=0`, `200`.

**Metrics** read from remoted's admin socket right after the matrix: `remoted.server.tls.ca_matches_leaf = 1`, `remoted.server.tls.cert_expiry_days = 3649`, `remoted.cacerts.{served = 7, not_found = 1, ca_mismatch = 1, rate_limited = 0}`, `remoted.http.cacerts.responses.{2xx = 7, 503 = 1, other = 1}` — one for one with the matrix above, and no `chain_valid` metric exists (deliberate: the verdict is reported in the snapshots and the log lines).

Scripts and raw output: `verificar-manager.sh` / `verificar-manager-arranque.sh`, run as root against `/var/wazuh-manager`; both restore the certificates they touch.

#### Automated suites

Each stage left the whole `remoted_module_utest` suite green; counts are the suite total:

| Stage | Suite | Extra |
|---|---|---|
| Baseline (before any change) | 938 tests, 0 failures | — |
| Last good snapshot, bounded reader, read under the mutex | 948 (+10) | TSAN over the CA-source filter: 30 tests, 0 warnings; concurrency test with 8 readers × 200 reads against 20 atomic rotations |
| Read-failure logs, 404 with cause, single reader at start-up | 956 (+8) | two module-level tests started through the C ABI assert the WARN texts |
| `chainValid` | 966 (+10) | the five chain vectors reproduced independently with a standalone program against OpenSSL 3.0.13: intermediate anchor → valid; root only → `unable to get local issuer certificate`; expired CA → `certificate has expired`; non-CA signer → `invalid CA certificate`; self-signed CA → valid |
| Documentation sweep | — | `docs/build.sh` exit 0; both YAML files parse |
| Review fixes (leaf ownership, non-regular files, mutex proof) | 968 (+2) | see the final run below |

Final run on the rebased branch (`5.0.0` @ `17008ef5b7`): `remoted_module_utest` **968 tests, 0 failures** (2 disabled); TSAN over the CA-source, status, endpoint and E2E filters: **49 tests, 0 warnings**, the canonical `src/build/lib` free of `__tsan` symbols; the four per-stage acceptance scripts 8/8, 8/8, 8/8 and 14/14; clang-format 22.1.1 (the CI's binary): 0 diffs on every changed C++ file; `docs/build.sh` exit 0; both YAML files parse; no production reference to `loadCertificates`/`evaluateCertificateStatus` and no `chain_valid` metric anywhere; the diff touches nothing outside `remoted_module`, its docs, two `spec.yaml` descriptions and `CHANGELOG.md`. A full manager build from this branch compiles with **0 warnings**.

The bounded read is proven by a fake reader that records what it was asked for: a file declaring 100 MiB is refused as too large having been asked for at most `kMaxBytes + 1` bytes, and the read itself is proven to run under the mutex by a reader that counts its in-flight calls (never more than one).

#### Adversarial review of the whole diff

A fresh review of `5.0.0...HEAD` found **no blocker** and ten points, all addressed in the last commit: the source now owns a reference to the listener's leaf (a later `start()` could otherwise free it under a caller still holding the source); the reader opens with `O_NONBLOCK` and refuses anything that is not a regular file (a FIFO with no writer at the configured path would otherwise park every caller behind the mutex, and a device would read as an emptied bundle); the one `maxBytes` value that would wrap is refused; the error text comes from `std::generic_category()` (thread-safe on older glibc, same strings); a duplicate-certificate branch that has been dead since OpenSSL 1.1 was removed; two tests were added and one renamed to what it actually pins; three documentation sentences were tightened.

  - [x] Linux — full manager build from this branch (`make TARGET=manager`), 0 compiler warnings
  - [ ] Windows — N/A (manager-only change, `remoted_module`)
  - [ ] MAC OS X — N/A (manager-only change, `remoted_module`)
- [x] Log syntax and correct language review — every new line is quoted verbatim in the evidence above, as emitted by the installed manager

- Memory tests for Linux
  - [ ] Coverity — not run locally
  - [ ] Valgrind (memcheck and descriptor leaks check) — covered by the `Run tests (valgrind)` CI job over `remoted_module_utest`
  - [ ] AddressSanitizer — covered by the `Run tests (asan)` CI job over `remoted_module_utest`; ThreadSanitizer additionally run locally over the CA-source, status, endpoint and E2E filters (49 tests, 0 warnings)
  - [ ] Coverity
  - [ ] UMDH
- Memory tests for macOS — N/A
  - [ ] Leaks
  - [ ] AddressSanitizer

- Engine _(Wazuh v5.x and above)_ — N/A (the engine is untouched)
  - [ ] Test run in parallel

- Wazuh server API/Framework — N/A (`api/api/spec/spec.yaml` changes two `description` strings only; no framework or API code changes)

### Artifacts Affected

- `lib/libremoted_module.so`, loaded by `wazuh-manager-remoted` (Linux manager packages). No other executable changes.
- Documentation under `docs/`, the module README, `api/api/spec/spec.yaml` (description strings) and `CHANGELOG.md` are not shipped artifacts; listed here so reviewers know they were considered.

### Configuration Changes

N/A — no new options, no changed defaults (`remote.https.ca_certificate`, `cacerts_rate_limit` and the 1 MiB cap are unchanged). No new metrics.

### Tests Introduced

All in `src/remoted/remoted_module/test/unit/` (`remoted_module_utest`, 938 → 968 cases):
- `caCertificateSource_test.cpp`: read failure keeps the previous snapshot (injected reader, missing file, directory at the path); a file that was never readable yields nothing; an emptied file clears the snapshot; the cap with a fake reader (asserting the bytes requested) and with a real oversized file; identical content after a failure is a cache hit; 8 concurrent readers racing 20 atomic rotations never observe a torn snapshot, and a reader counting its in-flight calls proves the read runs under the mutex; `readFileBounded()` (including a FIFO refused as not a regular file) and `describeReadFailure()` suites; `statusFrom()` carries the read failure; the snapshot carries `chainValid`.
- `httpServer_test.cpp`: `TlsCertificateStatusTest.ChainValid*` (intermediate anchor, root only, expired CA, non-CA signer, self-signed CA, unknown without leaf or bundle); start-up status from the shared source with a directory at the CA path; the monitor keeps evaluating through a read failure while the last good verdict stands; start-up status carries the chain verdict. Existing status tests now drive `statusFrom()` over a real `CaCertificateSource`.
- `cacertsEndpoint_test.cpp`: a read failure changes neither the answer nor the counters; `chainValid` never decides the answer.
- `cacertsE2E_test.cpp`: moving the CA away keeps serving the same body; an emptied CA answers `404` without a restart.
- `remotedModule_test.cpp` (module started through the C ABI, real log sink): the WARN while serving the last good bundle; the `404` names `cannot be read (Is a directory)`; the WARN when a `CA:FALSE` signer's chain does not validate.
- New `testCertificates.hpp`: shared in-memory PKI builder (`makeCertificate(isCa)`, `writePemKey()`).

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

Follow-ups found while working on this and left out on purpose (drafts ready to be filed): a ThreadSanitizer CI job for the manager modules, a valgrind suppressions file for the module CI, pinning the clang-format version the CI downloads, validating `agent-api.yaml` in `docs/build.sh`, and a "Design decisions" section in the `remoted_module` README.


LucioDonda (commented):
I think we should face this issue:

<details>
<summary><b>H-1 — The chain verdict is frozen by the content-hash cache, so an expiring CA is invisible at runtime — and this PR widens what depends on it</b> (#39465 · correctness + security · promoted · verified in main thread)</summary>

`src/remoted/remoted_module/src/http_server/caCertificateSource.cpp:200-203`

```cpp
auto hash = digestOf(contents);
if (hash == m_hash)
{
    m_snapshot.lastReadFailure.reset();
    return m_snapshot;          // <-- buildLocked() never re-runs
}
```

`buildLocked()` — which calls `ca_bundle::vouch()` and `chainValidates()` — runs **only when the
file's bytes change**. Everything else is served from `m_snapshot`.

This PR makes that verdict **time-dependent**: `vouch()`'s leaf guard moved from `anyCaSignsLeaf`
(a pure signature check) to `leafChainsToAnyCa()` (`X509_verify_cert()` with default flags), which
now weighs the validity window of the whole path. So the verdict has an expiry date, but the cache
has no clock.

**Failure scenario.** A manager's CA expires at midnight. Nobody touches `root-ca.pem`. Then,
indefinitely, until the bytes change or remoted restarts:

| Surface | Reports | Should report |
|---|---|---|
| `GET /cacerts` | `200` + `Wazuh-CA-Generation: N` | `503 ca_mismatch` |
| `POST /control` notify | `ca_generation: N` | `0` |
| `remoted.server.tls.ca_matches_leaf` | `1` | `0` |
| the 24 h monitor tick | silent | WARN |
| the `guard_failed` event | never fires | fires |

Every agent that verifies is now failing the handshake, and every observability surface says the
CA is fine.

**There is no second line of defence.** `remoted.server.tls.cert_expiry_days` does not cover this:
`TlsCertificateSnapshot::expiryDays` is documented as *"Days until the served **leaf** expires"*
(`tlsCertificateStatus.hpp:155`), so it watches the leaf, not the CA. Measured live on this host
with an already-expired CA and a 10-year leaf: `cert_expiry_days = 3649`, `ca_matches_leaf = 0`.
The one metric that names expiry is blind to the CA, and the one that would catch it is the frozen
value. An operator cannot alert on this at all.

**I verified in the main thread that there is no path around the cache.** All four runtime
consumers go through `snapshot()` → `snapshotLocked()` → the same short-circuit:
`RestinioHttpServer.cpp:1532` (the endpoint), `:1595` (metrics), `:1798` (the daily monitor), and
`descriptor()` at `:1557` (the notify path, which calls `snapshotLocked()` internally).

**Why this blocks, when the same defect was accepted as a deferral on #39370.** Three reasons:

1. **The blast radius grew.** On #39370 the frozen value was an informational log line. Here it is
   the published generation, the `503` decision, and the anchor handed to upgrading 4.x agents.
2. **It is the motivating case.** Issue #39277 opens with the fleet-availability cliff as a CA
   nears expiry. This is the mechanism built to survive that, and it cannot see it coming.
3. **The authors know.** `src/shared_modules/ca_bundle/README.md` states it outright: *"a CA that
   expires while the file is untouched keeps its last verdict until the bytes change or remoted
   restarts (issue #39319, D1: deferred, and the deferral weighs more now than it did when the
   guard was time-free)."* That last clause is the argument for fixing it, written by the author.

**The fix already exists.** Branch `enhancement/39320-expose-listener-and-ca-certificate-validity`
carries commit `204c0a19af` — *"Re-evaluate the CA chain verdict on every snapshot() instead of
caching it with the bytes"*. Re-evaluating only the time-dependent predicate on a cache hit keeps
the parse cached and costs one `X509_verify_cert()` per read; the reads are a rate-limited route, a
daily monitor, a metrics scrape and a ≤1 Hz descriptor refresh, so the cost is not on a hot path.

**Two lenses rated this below high** (correctness MEDIUM "documented in https-events-api.md";
security LOW "availability only"). I promoted it: documenting a hole does not close it, and
"availability only" understates a fleet-wide TLS outage that every health surface reports as green.

</details>

* Script for easy replication:

```bash
#!/usr/bin/env bash
# H-1 reproduction with timestamped evidence. Run as root on a manager built from
# enhancement/39319-ca-rotation. Leaves a log at /tmp/h1-evidence.log.
# Restore afterwards:  rm -rf /var/wazuh-manager/etc/certs && cp -a /root/certs.bak /var/wazuh-manager/etc/certs
set -uo pipefail
H=${WAZUH_HOME:-/var/wazuh-manager}; URL="https://127.0.0.1:1517/wazuh-manager/cacerts"
FX=$(mktemp -d); LOG=/tmp/h1-evidence.log; MIN=${1:-4}
say(){ printf '%s  %s\n' "$(date -u +%H:%M:%S)" "$*" | tee -a "$LOG"; }
probe(){ printf '%s  %-22s HTTP=%s  ca_matches_leaf=%s\n' "$(date -u +%H:%M:%S)" "$1" \
  "$(curl -sk -o /dev/null -w '%{http_code}' "$URL")" \
  "$(curl -s --unix-socket $H/queue/sockets/remote-admin-http.sock http://localhost/metrics \
     | tr ',' '\n' | grep -A1 ca_matches_leaf | grep -o '"value":[0-9.]*' | cut -d: -f2)" | tee -a "$LOG"; }

: > "$LOG"; say "=== H-1 repro, CA lifetime ${MIN}m ==="
"$(dirname "$0")/make-fixtures.sh" "$FX" "$MIN" >/dev/null
EXP=$(openssl x509 -in "$FX/ca.pem" -noout -enddate | cut -d= -f2); say "CA expires: $EXP"
[ -d /root/certs.bak ] || cp -a "$H/etc/certs" /root/certs.bak
install -m0640 -o root -g wazuh-manager "$FX/ca.pem"       "$H/etc/certs/root-ca.pem"
install -m0640 -o root -g wazuh-manager "$FX/leaf.pem"     "$H/etc/certs/remoted.pem"
install -m0640 -o root -g wazuh-manager "$FX/leaf-key.pem" "$H/etc/certs/remoted-key.pem"
rm -f "$H/var/run/remoted-ca-bundle/record.json"
say "restarting remoted (snapshot will be built while the CA is still valid)"
"$H/bin/wazuh-manager-control" restart >/dev/null 2>&1; sleep 8
probe "A: before expiry"          # expect 200 / 1
until [ "$(date -u +%s)" -gt "$(date -u -d "$EXP" +%s)" ]; do sleep 5; done; sleep 3
probe "B: AFTER expiry"           # BUG: 200 / 1   CORRECT: 503 / 0   <-- the decisive line
printf '# touched\n' >> "$H/etc/certs/root-ca.pem"; sleep 1
probe "C: after byte change"      # expect 503 / 0 -- proves it was the cache
say "=== done; B is the finding ==="; rm -rf "$FX"

```

juliancnn:
@LucioDonda will be fixed on: https://github.com/wazuh/wazuh/issues/39519

LucioDonda (approved):
## LGTM

* Manager installed by sources up and running 🟢 

* Manual Tests 🟢 

<p>

```console
(venv) ╭─root@9fac3eeb828e /workspaces/devContainer/wazuh ‹test/39281-HTTPS-channel-multi-node-cluster●› 
╰─# R=/workspaces/devContainer/backupContainers/NewDevs/Review39465-39472-CaRotation
C=/var/wazuh-manager/etc/certs/root-ca.pem
U=https://127.0.0.1:1517/wazuh-manager/cacerts
sudo cp $C /tmp/rootca.orig                       # red de seguridad

hdr(){ curl -sk -D - -o /dev/null $U | grep -iE '^HTTP|CA-Generation'; }

hdr                                                # 1) nunca publicado
sudo python3 $R/stamp-by-hand.py $C && sleep 2 && hdr   # 2) publicado
sudo sed -i 's/^## Content-SHA256: \(....\)/## Content-SHA256: dead/' $C && sleep 2 && hdr  # 3) hash roto
sudo cp /tmp/rootca.orig $C && sleep 2 && hdr      # 4) editado fuera de la herramienta

HTTP/1.1 200 OK
Wazuh-CA-Generation: 0
stamped /var/wazuh-manager/etc/certs/root-ca.pem: publication=1790005378 certificates=1 sha256=c0e63c118de5b005...
HTTP/1.1 200 OK
Wazuh-CA-Generation: 1790005378
HTTP/1.1 200 OK
Wazuh-CA-Generation: 0
HTTP/1.1 200 OK
Wazuh-CA-Generation: 0
(venv) ╭─root@9fac3eeb828e /workspaces/devContainer/wazuh ‹test/39281-HTTPS-channel-multi-node-cluster●› 
╰─# 
(venv) ╭─root@9fac3eeb828e /workspaces/devContainer/wazuh ‹test/39281-HTTPS-channel-multi-node-cluster●› 
╰─# 
(venv) ╭─root@9fac3eeb828e /workspaces/devContainer/wazuh ‹test/39281-HTTPS-channel-multi-node-cluster●› 
╰─# 
(venv) ╭─root@9fac3eeb828e /workspaces/devContainer/wazuh ‹test/39281-HTTPS-channel-multi-node-cluster●› 
╰─# sudo tail -40 /var/wazuh-manager/logs/wazuh-manager.log | grep -aE "cacerts|CA bundle"

2026/09/21 15:19:26 wazuh-manager-remoted:http-server: INFO: Unpublished CA bundle at 'etc/certs/root-ca.pem'; agents are told this manager has no published bundle (0). Publish it with `wazuh-manager-certs stamp` or `add` on the master.
2026/09/21 15:20:53 wazuh-manager-remoted:endpoints(cacerts): INFO: CA bundle 'etc/certs/root-ca.pem' is published as generation 1790004051 (was 0); that is the generation agents asking this manager are told about.
2026/09/21 15:21:10 wazuh-manager-remoted:endpoints(cacerts): WARNING: The publication block of the CA bundle 'etc/certs/root-ca.pem' does not describe the certificates next to it (Content-SHA256 mismatch); the bundle is served as before and announced as unpublished (0) until it is stamped again.
2026/09/21 15:21:14 wazuh-manager-remoted:endpoints(cacerts): WARNING: CA bundle 'etc/certs/root-ca.pem' changed outside the tool and is not published; previous publication 0; agents are told this manager has no published bundle (0) until `wazuh-manager-certs stamp` is run on the master.
2026/09/21 15:43:00 wazuh-manager-remoted:endpoints(cacerts): INFO: CA bundle 'etc/certs/root-ca.pem' is published as generation 1790005378 (was 0); that is the generation agents asking this manager are told about.
2026/09/21 15:43:02 wazuh-manager-remoted:endpoints(cacerts): WARNING: The publication block of the CA bundle 'etc/certs/root-ca.pem' does not describe the certificates next to it (Content-SHA256 mismatch); the bundle is served as before and announced as unpublished (0) until it is stamped again.
2026/09/21 15:43:04 wazuh-manager-remoted:endpoints(cacerts): WARNING: CA bundle 'etc/certs/root-ca.pem' changed outside the tool and is not published; previous publication 0; agents are told this manager has no published bundle (0) until `wazuh-manager-certs stamp` is run on the master.
(venv) ╭─root@9fac3eeb828e /workspaces/devContainer/wazuh ‹test/39281-HTTPS-channel-multi-node-cluster●› 
╰─# sudo curl -s --unix-socket /var/wazuh-manager/queue/sockets/remote-admin-http.sock \
  http://localhost/metrics | tr ',' '\n' | grep -E 'cacerts\.|ca_matches_leaf'

{"name":"remoted.cacerts.ca_mismatch"
{"name":"remoted.cacerts.not_found"
{"name":"remoted.cacerts.rate_limit.available"
{"name":"remoted.cacerts.rate_limit.burst"
{"name":"remoted.cacerts.rate_limit.limit"
{"name":"remoted.cacerts.rate_limited"
{"name":"remoted.cacerts.served"
{"name":"remoted.http.cacerts.responses.2xx"
{"name":"remoted.http.cacerts.responses.400"
{"name":"remoted.http.cacerts.responses.403"
{"name":"remoted.http.cacerts.responses.409"
{"name":"remoted.http.cacerts.responses.413"
{"name":"remoted.http.cacerts.responses.429"
{"name":"remoted.http.cacerts.responses.500"
{"name":"remoted.http.cacerts.responses.503"
{"name":"remoted.http.cacerts.responses.other"
{"name":"remoted.server.tls.ca_matches_leaf"
(venv) ╭─root@9fac3eeb828e /workspaces/devContainer/wazuh ‹test/39281-HTTPS-channel-multi-node-cluster●› 
╰─# $R/make-fixtures.sh /tmp/fx 60
sudo install -m0640 -o root -g wazuh-manager /tmp/fx/root2.pem $C   # CA ajena
sleep 2; curl -sk -o /dev/null -w '%{http_code}\n' $U               # 503
sudo cp /tmp/rootca.orig $C                                          # restaurar

fixtures in /tmp/fx ; short CA expires Sep 21 16:43:35 2026 GMT
503
(venv) ╭─root@9
```

</p>
</details>

matigarciadev (approved):
### LGTM!

### Built and Install
```console
General settings:
    TARGET:             manager
    V:                  
    DEBUG:              
    EXTERNAL_SRC_ONLY:  
    HTTP_REQUEST_BRANCH:c5f94080441fc35a80c5cf8ddd9cc1ca9fc20d9b
User settings:
USE settings:
    USE_INOTIFY:        no
    USE_BIG_ENDIAN:     no
    USE_SELINUX:        no
    USE_AUDIT:          no
    IMAGE_TRUST_CHECKS: 1
Compiler:
    MAKE              make

Done building manager

Wait for success...
success
cd ../framework && /var/wazuh-manager/framework/python/bin/python3 -m pip install . --use-pep517--prefix=/var/wazuh-manager/framework/python && rm -rf build/
Processing ./.
  Created wheel for wazuh: filename=wazuh-5.0.0-py3-none-any.whl size=312395 sha256=ddfc2c40876fd5d147188427f5154cef013160f28adf6e3ee1b2c4a303e2c205
  Stored in directory: /tmp/pip-ephem-wheel-cache-yskgx7w0/wheels/50/e9/6c/171615ee3709da12266524fed91cb0863278bc995eed9295be
Successfully built wazuh
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended touse a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the --root-user-actionoption if you know what you are doing and want to suppress this warning.
Processing ./.
  Created wheel for api: filename=api-5.0.0-py3-none-any.whl size=126157 sha256=4f24dde1824d009b7135dfcde826939261485b6aaf78eb7b8499b93a2e5913df
  Stored in directory: /tmp/pip-ephem-wheel-cache-v0vjjfbw/wheels/12/3e/cd/fe7f9fbf555382f244bd9d9630026d54f4ffbcb658cc4f4190
Successfully built api
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended touse a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the --root-user-actionoption if you know what you are doing and want to suppress this warning.
Generating schema files...
Loading resources...
Loading wcs_flat.yml...
Successfully merged 1 files into temporary file: /tmp/merged_wcs_7y6vsf95.yml
Loading schema template...
Success.
Success.
Success.
Generating engine schema...
Success.
Success.
Success.
Cleaned up temporary file: /tmp/merged_wcs_7y6vsf95.yml
Generated wazuh-decoders.json
Success.
Copying store files...
Installing GeoIP databases...
NOTICE: no TLS certificate for the HTTPS agent listener was found
        (/var/wazuh-manager/etc/certs/remoted.pem, /var/wazuh-manager/etc/certs/remoted-key.pem).
        wazuh-manager does not generate certificates. Provision root-ca.pem,
        remoted.pem and remoted-key.pem with the Wazuh installation assistant
        (wazuh-certs-tool) before starting the service; wazuh-manager-control
        refuses to start until they exist. See 'Deploy certificates' in the
        installation guide (docs/ref/getting-started/installation.md).


Starting Wazuh...
(1244): Invalid configuration at '/remote/https/certificate': file not found: /var/wazuh-manager/etc/certs/remoted.pem (the manager does not generate certificates; provision the file, e.g. withwazuh-certs-tool).
wazuh-manager.conf: Configuration error. Exiting

 - To start Wazuh:

 - To stop Wazuh:

   More information at: 
```

### Manager Status
```console
```

### Tests

#### `ca_bundle` unit tests
```bash
/workspaces/wazuh-5.x/wazuh/src/build/shared_modules/ca_bundle/test/ca_bundle_utest
```
```text
Running main() from /workspaces/wazuh-5.x/wazuh/src/external/googletest/googletest/src/gtest_main.cc
[==========] Running 33 tests from 1 test suite.
[----------] 33 tests from CaBundleTest
[ RUN      ] CaBundleTest.ContentSha256IsOrderAndPemFormatIndependent
[       OK ] CaBundleTest.ContentSha256IsOrderAndPemFormatIndependent (2 ms)
[ RUN      ] CaBundleTest.ContentSha256ChangesWhenACertificateByteChanges
[       OK ] CaBundleTest.ContentSha256ChangesWhenACertificateByteChanges (0 ms)
[ RUN      ] CaBundleTest.ContentSha256UsesAscendingDerOrder
[       OK ] CaBundleTest.ContentSha256UsesAscendingDerOrder (0 ms)
...
[==========] 33 tests from 1 test suite ran. (14 ms total)
[  PASSED  ] 33 tests.
```

#### `remoted_module` unit tests
```bash
/workspaces/wazuh-5.x/wazuh/src/build/remoted/remoted_module/test/remoted_module_utest
```
Summary:

```text
[==========] 1078 tests from 126 test suites ran. (194304 ms total)
[  PASSED  ] 1078 tests.
  YOU HAVE 2 DISABLED TESTS
```

#### OpenAPI document parses

`docs/ref/modules/remoted/agent-api.yaml` gains the `Wazuh-CA-Generation` response header and the
`ca_generation` field, so it is checked for well-formedness. `docs/build.sh` does not validate it.

```bash
cd /workspaces/wazuh-5.x/wazuh
python3 -c "
import yaml
f='docs/ref/modules/remoted/agent-api.yaml'
d=yaml.safe_load(open(f))
print(f+': parsed OK, top-level keys =', list(d.keys()))
"
```

```text
docs/ref/modules/remoted/agent-api.yaml: parsed OK, top-level keys = ['openapi', 'info', 'servers', 'tags', 'paths', 'components']
```

NahuFigueroa97 (approved):
LGTM!

## 1. A plain PEM is served and advertised as `0`

Configured bundle `etc/certs/root-ca.pem`: 1 certificate, no `##` block.

```
$ curl -sk -D - -o body.pem https://127.0.0.1:1517/wazuh-manager/cacerts
HTTP/1.1 200 OK
Content-Length: 1204
Server: wazuh-manager-remoted
Content-Type: application/x-pem-file
Wazuh-CA-Generation: 0
```

The body is still served, and it is certificates only — `grep -c '^##' body.pem` → `0`, first line
`-----BEGIN CERTIFICATE-----`.

One INFO at startup, naming the remedy:

```
2026/09/21 15:58:42 wazuh-manager-remoted:http-server: INFO: Unpublished CA bundle at
'etc/certs/root-ca.pem'; agents are told this manager has no published bundle (0). Publish it with
`wazuh-manager-certs stamp` or `add` on the master.
```

`POST /control` notify, same state:

```
<-- 200 OK
{"agent":{"config_hash":"ab00…","config_token":"default","groups":["default"]},
 "ca_generation":0,
 "settings_hash":"d45f…","tasks":[],"vd_feed_offset":0}
```

`ca_generation` is a JSON number. The only digests in the body are `config_hash` and
`settings_hash`, the pre-existing shared-configuration ones — **nothing about the CA**.

## 2. A stamped bundle is vouched for, with no restart

Block written by hand: `Publication: 1790013744`, `Content-SHA256` computed as SHA-256 over the
concatenation of the certificates' DER encodings sorted by DER bytes
(`ef8141c8f97c8fa45c51b3b5f92c3a261e61b0dbdd57f2fd640dd547d6e7d71f`), installed `root:wazuh-manager
640`. **No `wazuh-manager-control restart`.**

```
$ curl -sk -D - -o body2.pem https://127.0.0.1:1517/wazuh-manager/cacerts
HTTP/1.1 200 OK
Wazuh-CA-Generation: 1790013744

$ notify
"ca_generation":1790013744
```

```
2026/09/21 18:02:34 wazuh-manager-remoted:endpoints(cacerts): INFO: CA bundle
'etc/certs/root-ca.pem' is published as generation 1790013744 (was 0); that is the generation
agents asking this manager are told about.
```

The served body still carries `0` lines of `##` and 1 certificate: the block is parsed, never
forwarded.

## 3. Both degradations behave as designed, and keep serving

**Stripping only the `##` block** (certificates untouched):

```
HTTP 200   Wazuh-CA-Generation: 0

WARNING: CA bundle 'etc/certs/root-ca.pem' changed outside the tool and is not published;
previous publication 1790013744; agents are told this manager has no published bundle (0) until
`wazuh-manager-certs stamp` is run on the master.
```

**A block whose `Content-SHA256` does not describe the certificates** (64 zeros):

```
HTTP 200   Wazuh-CA-Generation: 0

WARNING: The publication block of the CA bundle 'etc/certs/root-ca.pem' does not describe the
certificates next to it (Content-SHA256 mismatch); the bundle is served as before and announced as
unpublished (0) until it is stamped again.
```

Both name what happened and what to do, both keep answering `200` with the body.

## 4. remoted never writes the bundle, and its record is private

Restoring the original file returns the advertisement to `0` and the file is byte-identical to the
backup, still `root:wazuh-manager 640`.

```
$ sudo -u wazuh-manager test -w /var/wazuh-manager/etc/certs/root-ca.pem
  -> no write permission on the trust anchor
$ remoted pid=305627 user=wazuh-manager
```

The private record it does keep:

```
$ ls -l /var/wazuh-manager/var/run/remoted-ca-bundle/
-rw-r----- 1 wazuh-manager wazuh-manager 148 record.json

{
    "bundle_path": "etc/certs/root-ca.pem",
    "file_sha256": "bb3650807e5444af06d247f86d4fc73347b6cae9dab9b38e07aa75e28a077eef",
    "publication": 0,
    "version": 1
}
```

Shape, owner and mode as designed, and it holds a publication of `0` — it lends nothing.

Issue #39165, opened 2026-09-11
macOS user inventory reports only root as a sudoer, missing users granted sudo via group membership or sudoers.d, and matches usernames as substrings


Rebits:
## Description

The macOS agent reports exactly one account with the `sudo` role out of 134 users, `root`, on a host where a second account demonstrably holds passwordless sudo. "Which accounts can escalate to root" is a first-order security question, and on macOS the inventory currently under-answers it.

Role distribution per agent:

```
  agent 006 total=134  roles: sudo:1     <-- macOS, only root
  agent 002 total=35   roles: sudo:2
  agent 001 total=34   roles: sudo:2
  agent 003 total=29   roles: sudo:2
  agent 004 total=29   roles: sudo:2
  agent 005 total=9    roles: none       <-- Windows, N/A
```

and the single macOS match:

```
  count: 1
  root
```

The `vagrant` account on that host has working sudo. Every privileged command executed against this agent during testing ran as `sudo` under the `vagrant` login and succeeded, repeatedly and across the whole session, including `sudo sqlite3`, `sudo /Library/Ossec/bin/wazuh-control restart` and `sudo bash` script invocations. It is nonetheless absent from the sudoers role in the inventory.

The cause is visible in `SysInfo::getUsers()` in `src/data_provider/src/sysInfoMac.cpp`:

```cpp
userItem["user_roles"] = UNKNOWN_VALUE;

for (auto& singleSudoer : collectedSudoers)
{
    // Searching in content of header
    auto header = singleSudoer["header"].get<std::string>();

    if (header.find(username) != std::string::npos)
    {
        //TODO: user_roles_sudo_sudo_rule_details has more detailed information.
        userItem["user_roles"] = "sudo";
    }
}
```

Two independent weaknesses:

**Grants that are not a literal username in a rule are never seen.** On macOS, administrative sudo is conventionally granted by group, through the stock `%admin ALL=(ALL) ALL` rule, and tooling commonly adds drop-in files under `/etc/sudoers.d/`. Neither form contains the granted user's name in a rule header, so no amount of string searching will find it. A user in the `admin` group has full sudo and is reported as having no role at all. This is the false negative demonstrated above.

[...]

Issue #39164, opened 2026-09-11
macOS user inventory computes login_status from the last enumerated session instead of from whether any session matched, so logged-in users can be reported as not logged in


Rebits:
## Description

`SysInfo::getUsers()` in `src/data_provider/src/sysInfoMac.cpp` decides whether a user is logged in by walking the list of active sessions, but it resets the answer to "not logged in" on every iteration of that walk. The final value therefore depends on whether the *last* session in the list happens to belong to the user, not on whether *any* of them does.

```cpp
auto matched = false;
auto lastLogin = 0;

for (auto& item : collectedLoggedInUser)
{
    // By default, user is not logged in.
    userItem["login_status"] = 0;          // <-- reset on every iteration

    if (item["user"] == username)
    {
        matched = true;
        userItem["login_status"] = 1;
        ...
        userItem["user_last_login"] = ...;
        userItem["login_tty"]       = item["tty"].get<std::string>();
        userItem["login_type"]      = item["type"].get<std::string>();
        userItem["process_pid"]     = item["pid"].get<int32_t>();
        ...
    }
}

if (!matched)
{
    userItem["login_status"] = 0;
    userItem["login_tty"] = UNKNOWN_VALUE;
    ...
}
```

The `matched` flag already exists to handle the no-match case, and the `if (!matched)` block after the loop does exactly that job correctly. The assignment inside the loop is therefore both redundant and harmful: for a user whose session is not the final entry, `login_status` is overwritten back to `0` by the next iteration, while the session details recorded in the same matching branch are left in place.

The result is a self-contradicting record, which is what the deployed agent shows. This is the only user on the macOS host with session fields populated:

```
     user.name              'vagrant'
     user.id                '501'
     login.status           False
     login.tty              'ttys001'
     login.type             'dead'
     process.pid            1880
     user.last_login        '2026-09-09T13:25:27.000Z'
     host.ip                None
```

`login.tty`, `login.type`, `process.pid` and `user.last_login` are assigned *only* inside the branch guarded by `item["user"] == username`, the same branch that sets `login_status = 1`. Their presence proves that branch executed, yet `login.status` is `False`.

[...]

PR #39466, opened 2026-09-18
Report password expiration date and max change interval on macOS

### LINKED ISSUE #39183: macOS user inventory hardcodes all seven password policy and status fields, so no account reports whether it is locked, expired or subject to any password ageing
## Description

`SysInfo::getUsers()` on macOS writes seven password-related fields as literal constants for every user, in one block, immediately after the login-session handling:

```cpp
userItem["user_password_expiration_date"] = UNKNOWN_VALUE;
userItem["user_password_hash_algorithm"] = UNKNOWN_VALUE;
userItem["user_password_inactive_days"] = 0;
userItem["user_password_max_days_between_changes"] = 0;
userItem["user_password_min_days_between_changes"] = 0;
userItem["user_password_status"] = UNKNOWN_VALUE;
userItem["user_password_warning_days_before_expiration"] = 0;
```

Nothing on the macOS path ever assigns any of them. The result is that the password posture of a macOS account is not reported at all, and in particular `user.password.status`, the field that says whether an account is locked, disabled or passwordless, is absent from every document.

The agent under test has 134 users, and the values are identical for all of them:

```
  agent 006 (macOS)                             agent 001 (Linux)
    password.max_days_between_changes   0=134     99999=19, -1=15
    password.min_days_between_changes   0=134         0=19, -1=15
    password.warning_days_before_expiry 0=134         7=19, -1=15
    password.inactive_days              0=134        -1=34
    password.status                     (absent)  locked=34
    password.hash_algorithm             (absent)  (absent)
```

The Linux agent reports two distinct values for each ageing field, because it reads them per account out of the shadow database, and it reports a status for every user. The macOS agent reports one distinct value for each, because the value is a literal.

This is not a case of the collector having no access to the information. The macOS user collector already holds an OpenDirectory session and already queries per-user records through it. `UsersProvider::collect()` calls `collectAccountPolicyData()` for every user, which reaches `od::genAccountPolicyData()`, which opens `/Local/Default`, queries `kODRecordTypeUsers` and parses the account policy plist:

```cpp
ODNode* root = [ODNode nodeWithSession:s name:@"/Local/Default" error:&err];
...
ODQuery* q = [ODQuery queryWithNode:root
                     forRecordTypes:kODRecordTypeUsers
                          attribute:kODAttributeTypeUniqueID
                          matchType:kODMatchEqualTo
                         queryValues:[NSString stringWithFormat:@"%s", uid.c_str()]
                    returnAttributes:@"dsAttrTypeNative:accountPolicyData"
                      maximumResults:0
                               error:&err];
```

and extracts four values from it:

```cpp
assign_safe("creationTime",         "creation_time",          false);
assign_safe("failedLoginCount",     "failed_login_count",     true);
assign_safe("failedLoginTimestamp", "failed_login_timestamp", false);
assign_safe("passwordLastSetTime",  "password_last_set_time", false);
```

Those four do arrive correctly. `user.password.last_change`, `user.created` and `user.auth_failures.count` are all populated from real data on this agent, which is the proof that the OpenDirectory path works and is reaching real records. The gap is specifically the policy and status fields, which are not asked for.

Where the missing values live on macOS, so the fix has somewhere to start:

- The ageing policy is an account policy, not a per-user attribute, and is reachable from the same OpenDirectory session through `kODAttributeTypeAccountPolicies` on the node or the record. `pwpolicy -getaccountpolicies` shows the same data from the shell. The relevant keys are the `policyAttribute` expressions for password change interval and expiry.
- The account status is derivable from the `AuthenticationAuthority` attribute, where a disabled account carries `;DisabledUser;`, and from whether a `ShadowHashData` entry exists at all for accounts with no password.
- The hash algorithm is in the `ShadowHashData` plist, which on current macOS is `SALTED-SHA512-PBKDF2`.

The cost of the gap is straightforward. "Which accounts on this endpoint are locked, which have passwords that never expire, and which have no password ageing at all" is a routine audit question and a CIS control, and on macOS the inventory answers it with zeros for every account. A zero is worse than a blank here: `password.max_days_between_changes` of 0 does not read as "unknown", it reads as a password that must be changed every zero days, and `inactive_days` of 0 reads as an account that goes inactive immediately. A cross-platform panel comparing password hygiene will rank every Mac account as maximally restricted when in fact nothing is known about them.

One field in the same block is worth separating out. `user_password_expiration_date` is set to `UNKNOWN_VALUE` rather than 0, so it at least fails honestly by arriving absent instead of arriving wrong. The four numeric fields are the ones that produce actively misleading values.

## Environment

- Agent under test: `demo-env-macos-agent`, id `006`, macOS 26.5.1 (build 25F80), arm64, Parallels VM on MacStadium host `macstadium_arm_4`, 134 users inventoried.
- Source read from the `5.0.0` branch of `wazuh/wazuh`: the hardcoded block in `SysInfo::getUsers()` in `src/data_provider/src/sysInfoMac.cpp`, the provider in `src/data_provider/src/extended_sources/users/src/users_darwin.cpp`, and `od::genAccountPolicyData()` in `src/data_provider/src/extended_sources/wrappers/unix/darwin/od_wrapper.mm`.
- Comparison agent `001` ubuntu-26-arm64, 34 users, same nightly.
- QA demo environment `wqa-prod-58-demo-environment`: clustered managers, one indexer, one dashboard, six agents. Not a container deployment, so there are no image digests to record.

## Steps to reproduce

1. Enroll a 5.0.0 macOS agent alongside a Linux agent against the same manager and let a syscollector evaluation complete on each.
2. Ask for the distinct values of each password field per agent:

```bash
  "https://<indexer>:9200/wazuh-states-inventory-users/_search?size=0" \
       "aggs":{"max":{"terms":{"field":"user.password.max_days_between_changes","size":5}},
               "min":{"terms":{"field":"user.password.min_days_between_changes","size":5}},
               "warn":{"terms":{"field":"user.password.warning_days_before_expiration","size":5}},
               "inact":{"terms":{"field":"user.password.inactive_days","size":5}},
               "st":{"terms":{"field":"user.password.status","size":5}}}}'
```

Every numeric bucket is a single `0` covering all 134 users, and the `status` aggregation is empty. Repeating with `"001"` returns two distinct values per numeric field and a status for every user.

3. Confirm the host does hold the data:

```bash
sudo pwpolicy -getaccountpolicies
dscl . -read /Users/<someuser> AuthenticationAuthority
```

4. Confirm that the OpenDirectory path itself is working, by checking that the fields which are read from it are populated:

```bash
  "https://<indexer>:9200/wazuh-states-inventory-users/_search?size=3" \
       "_source":["user.name","user.password.last_change","user.created","user.auth_failures.count"]}'
```

These carry real per-user values, which rules out a collection failure and isolates the defect to the hardcoded block.

## Expected

macOS accounts report their password status and ageing policy, read from OpenDirectory through the session the collector already holds, so that locked and disabled accounts and accounts exempt from password ageing are visible as they are on Linux.

## Actual

Seven password fields are assigned literal constants in `getUsers()` and never touched again. All 134 users on the agent report `0` for inactive days, max days, min days and warning days, and no user reports a password status, an expiration date or a hash algorithm. The four numeric zeros are indistinguishable from a genuine, maximally restrictive policy.

## Not a duplicate of

`qa-known-issues.py check` found **no cache match** against 784 open bugs, the live search for `inventory hardcodes password reports account ageing` returned nothing, and the tool concluded "no known issue looks like this one".

Related but distinct: wazuh/wazuh#39164 concerns `login_status` being reset inside the enumeration loop in the same function, and wazuh/wazuh#39165 concerns sudoer attribution in the same function. Both are about different fields with different causes; none of the three overlaps.

rjcausarano:
## Description

`SysInfo::getUsers()` on macOS still hardcoded five password aging fields to `NOT_COLLECTED_VALUE` after #39356 fixed `user_password_status` and `user_password_hash_algorithm`. Of those five, two do have a real source: OpenDirectory's `accountPolicyData` carries a `policyCategoryPasswordChange` array when pwpolicy or an MDM has imposed a password-change interval, with the interval itself in `policyParameters.policyAttributeExpiresEveryNDays`. Nothing in the collector ever read it.

The other three (`user_password_inactive_days`, `user_password_min_days_between_changes`, `user_password_warning_days_before_expiration`) have no macOS equivalent at all: OpenDirectory's policy model has no concept of a minimum password age, a pre-expiration warning period, or a shadow-style post-expiration inactivity grace period. They stay `NOT_COLLECTED_VALUE` by design, not as a remaining gap.

Closes #39183

## Proposed Changes

- `od_wrapper.mm` / `od_wrapper.hpp`: `genAccountPolicyData()` now also parses the nested `policyCategoryPasswordChange` array of `accountPolicyData`, extracting `policyAttributeExpiresEveryNDays` from each entry's `policyParameters` into a new `expires_every_n_days` field. Present only when pwpolicy/MDM has actually set such a policy. When multiple entries carry a valid interval (e.g. a local pwpolicy and an MDM profile stacked), the most restrictive (minimum) one is kept, and the value is accepted whether OpenDirectory serializes it as a number or a string.
- `users_darwin.cpp`: `collectAccountPolicyData()` maps `expires_every_n_days` into `password_max_days_between_changes`, and derives `password_expiration_date` from it plus the already-collected `password_last_set_time` (`last_change + days * 86400`, matching the epoch-seconds semantics Linux already uses for its `expire` field). Both fields are only reported when a real (non-default) last-set time is available, and both are capped to the int32 wire field's range (mirroring `shadow_linux.cpp`'s `MAX_EXPIRE_DAYS` guard) so a malformed or out-of-range policy can't overflow the value the manager parses.
- `sysInfoMac.cpp`: `user_password_expiration_date` and `user_password_max_days_between_changes` now use the derived values when present, falling back to `NOT_COLLECTED_VALUE` when no policy is set. The comment above the remaining three hardcoded fields now explains why macOS genuinely has no source for them, instead of describing it as a temporary gap.
- Added unit test coverage in `test_users_darwin.cpp` (see Tests Introduced below).

### Results and Evidence

Correctness of the derivation, including the overflow-guard and stacked-policy-precedence logic added during review, is confirmed by CI actually building and running this code on macOS:

- **`build-sonoma`** ([run 35638303805](https://github.com/wazuh/wazuh/actions/runs/35638303805/job/106461108655), commit `95df2c9eb1`) compiles `users_darwin.cpp` and `od_wrapper.mm` on real macOS (Sonoma) with no compiler warnings on either file, and runs the `extended_sources_users_unit_test` binary. All 11 `UsersProviderTest` cases pass, including the four covering this change:
  - `CollectDerivesAgingFieldsFromExpiresEveryNDays` — normal derivation from a policy interval.
  - `CollectOmitsAgingFieldsWithoutLastSetTime` — a policy present with no real last-set time (i.e. `od_wrapper.mm`'s default `0.0`, not just an absent key) correctly leaves both fields unreported instead of deriving a bogus epoch-1970 expiration.
  - `CollectCapsExpirationDateAtInt32WireLimit` — a 24855-day policy against a realistic recent last-set time, which would overflow the int32 wire field with a naive day-count-only cap, correctly derives `-1` instead.
  - `CollectCapsMaxDaysBetweenChangesAtInt32WireLimit` — a corrupt/out-of-range 5-billion-day value clamps to `INT32_MAX` instead of overflowing.
- **`scan-build-macos-agent`** (same commit) — Clang static analyzer, passing, no findings in the changed files.
- Both `build-sonoma` variants (agent and server) are green on this commit.

  - [ ] Linux
  - [ ] Windows
  - [x] MAC OS X — confirmed via CI (`build-sonoma`, [run 35638303805](https://github.com/wazuh/wazuh/actions/runs/35638303805/job/106461108655)), not a local build; no warnings on the changed files.

- Memory tests for macOS
  - [ ] Leaks
  - [ ] AddressSanitizer

**Note:** the Darwin-only sources and their unit test target (`extended_sources_users_unit_test`, gated on `CMAKE_SYSTEM_NAME STREQUAL "Darwin"`) can't be built or run from this Linux development environment, so all macOS verification above came from this PR's own CI (`build-sonoma`, `scan-build-macos-agent`) rather than a local run. Still pending: a real-account verification under `pwpolicy -setpolicy expiresEveryNDays=<N>`, and the Leaks/AddressSanitizer memory checks.

### Artifacts Affected

- `src/data_provider/src/sysInfoMac.cpp`
- `src/data_provider/src/extended_sources/users/src/users_darwin.cpp`
- `src/data_provider/src/extended_sources/wrappers/unix/darwin/od_wrapper.mm` / `od_wrapper.hpp`

All macOS-only: only the macOS agent's syscollector user inventory (`user_password_expiration_date`, `user_password_max_days_between_changes` fields) is affected.

### Configuration Changes

N/A — no configuration parameters are introduced or changed by this fix.

### Tests Introduced

Four additions to `test_users_darwin.cpp` (`UsersProviderTest`), all passing on macOS CI as of commit `95df2c9eb1`:
- `CollectDerivesAgingFieldsFromExpiresEveryNDays`: a policy interval is present and `password_max_days_between_changes`/`password_expiration_date` are derived correctly.
- `CollectOmitsAgingFieldsWithoutLastSetTime`: a policy interval is present but `password_last_set_time` is at its uncollected default (`0.0`); neither aging field is reported.
- `CollectCapsExpirationDateAtInt32WireLimit`: a day count large enough that `lastSetTime + days*secondsPerDay` overflows the int32 wire field, given a realistic last-set time, is capped to `-1` rather than wrapping.
- `CollectCapsMaxDaysBetweenChangesAtInt32WireLimit`: a corrupt/out-of-range day count is clamped to `INT32_MAX` for `password_max_days_between_changes` rather than overflowing.

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...

Related but distinct: #39164 (login_status) and #39165 (sudoer attribution) touch the same `getUsers()` function but different fields with different causes.


cborla (approved):
LGTM
- [x] Code changes reviewed

rjcausarano:
## Review findings: what was applied and what wasn't

Recapping the findings raised across the review rounds on this PR (`od_wrapper.mm` / `users_darwin.cpp`), and the disposition of each:

**Applied (5):**

1. `password_expiration_date` had no upper bound on the derived epoch value → capped to the int32 wire field's range, mirroring `shadow_linux.cpp`'s `MAX_EXPIRE_DAYS` guard, computed relative to the account's actual `password_last_set_time` (not a fixed day-count constant).
2. `policyAttributeExpiresEveryNDays` was only accepted when OpenDirectory serialized it as `NSNumber` → added an `NSString` fallback via `std::stoll`, consistent with how the sibling fields already handle that ambiguity.
3. `password_max_days_between_changes` could be reported without a matching `password_expiration_date` when `password_last_set_time` was missing → both fields are now gated together, only reported when a real (non-default) last-set time is available.
4. The `policyCategoryPasswordChange` loop broke on the first entry with a valid interval, with no established precedence when a local `pwpolicy` and an MDM profile stack → now scans all entries and keeps the most restrictive (minimum) interval.
5. The gating added for #3 turned out to be a no-op, since `od_wrapper.mm` unconditionally defaults `password_last_set_time` to `0.0` rather than omitting the key → the guard now checks the value is actually `> 0`, not merely present. Also caught that the original regression test for #3 had mocked the OD layer without that default, so it wouldn't have caught this — rewrote it to match the real default behavior.

**Not applied (2):**

6. `expires_every_n_days == 0` is treated as a real day count rather than a possible "unset" sentinel. Not applied — the reviewer's own confidence on this was Low, and the surrounding code comment states the key is only present when a policy is actually imposed, which argues against `0` ever being used as an "unset" marker (absence, not zero, is how "no policy" is signaled). Assessed as not a real bug.
7. `secondsPerDay = 86400` duplicates `shadow_linux.cpp`'s `SECONDS_PER_DAY_INT`. Not applied — flagged by the reviewer itself as a minor DRY nit, not a blocker. Deduplicating would mean moving the constant into a header shared between the Linux and Darwin `users` providers, which is a larger structural change out of scope for this fix.

Also worth noting: one finding from a later review round (a supposedly still-miscalculated overflow cap) turned out to be a stale re-surfacing of #1 above, already fixed by that point — independently re-verified with exhaustive boundary arithmetic and confirmed correct on real macOS CI (`build-sonoma`), so no further change was needed there.

PR #39471, opened 2026-09-18
Make the run_as auth-context payload limit configurable

### LINKED ISSUE #39077: Raise and configure the API authentication context payload size limit
## Description

Since Wazuh 4.14.7, the API rejects `run_as` login requests when the request payload exceeds **8 KB**. This limit was introduced as part of the hardening changes in [#37034](https://github.com/wazuh/wazuh/pull/37034) to mitigate an authenticated DoS vulnerability.

The limit is currently hardcoded and affects environments using **AD/LDAP/SSO**, where the authentication context can contain a large number of group memberships. In these cases, the Dashboard's `run_as` request can easily exceed 8 KB and the login is rejected.

Local users are not affected because they use a different authentication endpoint.

The 8 KB limit is too restrictive for typical AD/LDAP environments and should be increased and made configurable.

### Proposed changes

* Increase the default `AUTH_CONTEXT_MAX_PAYLOAD_SIZE` from **8 KB** to **64 KB**.
* Expose the limit as an API configuration option in `api.yaml`, avoiding the need for users to patch the installed Python code.
* Keep the existing larger payload protection (`max_upload_size`) in place.
* Ensure that the authentication-context payload validation and the corresponding audit logging remain consistent when the limit is exceeded.
* Document the new configuration option and its default value.

The existing security mitigations introduced as part of the original vulnerability fix must remain active. Increasing this limit must not disable or bypass the regex timeouts, recursion guard, authentication thread pools, rate limiting, IP blocking, or the existing `max_upload_size` protection.

### Compatibility

The change must preserve the existing authentication behavior for normal local users while allowing AD/LDAP/SSO users with larger authentication contexts to authenticate successfully.

The default should be large enough to accommodate realistic AD/LDAP group memberships without requiring manual configuration, while remaining sufficiently bounded to prevent unnecessarily large authentication requests.

## Tasks

* [ ] Increase the default `AUTH_CONTEXT_MAX_PAYLOAD_SIZE` from `8 * 1024` to `64 * 1024`.
* [ ] Add an `api.yaml` configuration option for the authentication-context payload size.
* [ ] Load the configured value into the API middleware instead of relying on a hardcoded constant.
* [ ] Ensure the authentication-context payload validation uses the configured limit.
* [ ] Ensure the secondary payload-size check remains consistent with the authentication-context limit.
* [ ] Verify that authentication-context identifiers continue to be generated and included in `api.log` for valid requests below the configured limit.
* [ ] Add unit tests covering payloads below, at, and above the configured limit.
* [ ] Add tests covering configuration of a custom authentication-context payload limit.
* [ ] Add regression tests for AD/LDAP/SSO authentication contexts larger than 8 KB.
* [ ] Verify that `max_upload_size` remains an independent upper bound for API requests.
* [ ] Verify that the security mitigations introduced by [#37034](https://github.com/wazuh/wazuh/pull/37034) remain enabled and unaffected.
* [ ] Update the API configuration documentation with the new option, default value, and behavior.
* [ ] Add or update release notes to mention the increased and configurable authentication-context payload limit.

MiguelazoDS:
## Description

Since Wazuh 4.14.7, the API rejects `run_as` login requests (`POST /security/user/authenticate/run_as`) whose payload exceeds 8 KB — a hardcoded limit introduced by #37034's DoS hardening. AD/LDAP/SSO environments routinely produce larger authentication contexts (many group memberships), so the Dashboard's `run_as` request exceeds 8 KB and the login is rejected, even though local users are unaffected. Because the limit is hardcoded, raising it has required patching the installed Python code.

Closes #39077.

## Proposed Changes

- Raised the default `AUTH_CONTEXT_MAX_PAYLOAD_SIZE` from 8 KB to 64 KB and exposed it as a new `api.yaml` option, `auth_context_max_payload_size` (1 KB–1 MB), so admins with larger AD/LDAP contexts can raise it further without patching code.
- Both `CheckAuthContextSizeMiddleware` and `WazuhAccessLoggerMiddleware` now read the live config value instead of a hardcoded constant, keeping the secondary audit-logging check (`hash_auth_context`) consistent with the size gate.
- Every other #37034 mitigation (regex timeouts, recursion guard, authentication thread pools, rate limiting, IP blocking, `max_upload_size`) is unchanged and verified still active.
- Added unit and Tavern integration test coverage for the new limit, including custom-configured values.


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ All notable changes to this project will be documented in this file.
 - Added Fluentd server identity verification to the `fluent-forward` module: the certificate name is now checked against the configured address and the shared key digest returned by the server is verified. ([#38686](https://github.com/wazuh/wazuh/pull/38686))
 - Aligned the API `force` parameter with its OpenAPI schema: `POST /agents` now declares it, and `POST /agents/insert` no longer sends a `force` object that the request did not carry. ([#38804](https://github.com/wazuh/wazuh/pull/38804))
 - Escaped control characters in the request path of the API plain-text access log, so an unauthenticated request can no longer forge access log entries. ([#38894](https://github.com/wazuh/wazuh/pull/38894))
+- Raised the default API `run_as` authentication-context payload size limit from 8 KB to 64 KB and made it configurable via the new `auth_context_max_payload_size` option, for AD/LDAP/SSO logins with large group-membership contexts. ([#PLACEHOLDER](https://github.com/wazuh/wazuh/pull/PLACEHOLDER))
```
Antoniogm03: This still links to `#PLACEHOLDER` even though the PR is already open as #39471.

```suggestion
- Raised the default API `run_as` authentication-context payload size limit from 8 KB to 64 KB and made it configurable via the new `auth_context_max_payload_size` option, for AD/LDAP/SSO logins with large group-membership contexts. ([#39471](https://github.com/wazuh/wazuh/pull/39471))
```

Antoniogm03 (changes requested):
Could you add the new `auth_context_max_payload_size` field to the API documentation (api/api/spec/spec.yaml)?

CHANGELOG.md
MiguelazoDS: Oops. Thanks

Antoniogm03 (approved):

api/api/middlewares.py
```diff
@@ -42,8 +42,8 @@
 # Authentication context hash key
 HASH_AUTH_CONTEXT_KEY = 'hash_auth_context'
 
-# Allowed upper bound for auth_context payload
-AUTH_CONTEXT_MAX_PAYLOAD_SIZE = 8 * 1024
+# Default of the max auth_context payload size in bytes (see api.yaml `auth_context_max_payload_size`)
+AUTH_CONTEXT_MAX_PAYLOAD_SIZE = configuration.default_api_configuration["auth_context_max_payload_size"]
```
jotacarma90: **[S2 · blocking · introduced]** This is not the default: `read_yaml_config()` merges the installed `api.yaml` into `default_api_configuration` **in place** (`fill_dict._update_default_config` assigns into the dict it is handed, and `read_yaml_config` passes `default_api_configuration` uncopied at lines 315/341), so after the module-level `api_conf = read_yaml_config()` the two are the same object — I ran that logic in isolation and `api_conf is default_api_configuration` comes back `True`. On a host that sets `auth_context_max_payload_size: 16384` this constant is therefore 16384, which makes `default_auth_context_limit()` (`test_middlewares.py:739`) patch `api_conf` with the value that already came from `api.yaml` — a no-op, despite its docstring saying it exists "so the host's installed api.yaml can't skew the test". The sanity asserts at lines 889 and 924 then fail outright on any host configured below the payloads they build (14015 and 23295 bytes).

```suggestion
# Default of the max auth_context payload size in bytes (see api.yaml `auth_context_max_payload_size`).
# Literal on purpose: `read_yaml_config()` merges the installed api.yaml into `default_api_configuration`
# in place, so that dict stops holding the default as soon as the option is set.
AUTH_CONTEXT_MAX_PAYLOAD_SIZE = 64 * 1024
```

api/api/configuration.py
```diff
@@ -40,6 +40,7 @@
     "drop_privileges": True,
     "experimental_features": False,
     "max_upload_size": 10485760,
+    "auth_context_max_payload_size": 65536,
```
jotacarma90: **[S3 · non-blocking · aggravated]** Raising the cap 8x multiplies by 8 the attacker-supplied data reaching three consumers that this option does not bound: `WazuhAccessLoggerMiddleware` caches `_json` *before* dispatching, so `access_log` → `custom_logging` writes the body verbatim into `api.log` twice (plain line + JSON record) even when the request fails to authenticate, which with the defaults (`max_login_attempts: 50`, `block_time: 300`, `logs.max_size.enabled: False`) is ~6.4 MB per IP per 5 minutes against ~0.8 MB before, while `process_lists` in `framework/wazuh/rbac/auth_context.py` walks the auth-context list against every role rule value with a 0.1 s cap per match and no total budget per request. Nothing in the PR compensates for the increase. The memory half of this is already solved on `5.0.0`/`main` (`get_declared_content_length()` plus `read_capped_body()`, which abandons the stream at the limit) and was not brought down here — if porting that is out of scope, the cheap 4.14.x mitigation is to keep what gets cached and logged bounded at the old 8 KB, since `hash_auth_context` already identifies the context, while letting the gate itself rise.

api/api/middlewares.py
```diff
@@ -316,7 +317,8 @@ async def dispatch(self, request: Request, call_next: RequestResponseEndpoint) -
 
         # Don't allow heavy bodies when trying to authenticate. Necessary because this middleware is executed before
         # CheckAuthContextSizeMiddleware can be executed
-        if body and (request.url.path != RUN_AS_LOGIN_ENDPOINT or len(body) <= AUTH_CONTEXT_MAX_PAYLOAD_SIZE):
+        max_auth_context_payload_size = configuration.api_conf['auth_context_max_payload_size']
+        if body and (request.url.path != RUN_AS_LOGIN_ENDPOINT or len(body) <= max_auth_context_payload_size):
```
jotacarma90: **[S3 · non-blocking · introduced]** Hoisting the lookup out of the `run_as` branch makes the one middleware that runs on *every* request depend unconditionally on a key it only needs for `run_as`, so anything that swaps `api_conf` for a partial dict turns a `run_as`-only problem into a `KeyError` (500) for all traffic — a pattern this test file already uses at lines 148 and 560. The condition already short-circuits on the path, so inlining it costs nothing.

```suggestion
        if body and (request.url.path != RUN_AS_LOGIN_ENDPOINT
                     or len(body) <= configuration.api_conf['auth_context_max_payload_size']):
```

api/api/configuration/api.yaml
```diff
@@ -52,6 +52,12 @@
 # Maximum body size that the API can accept, in bytes (0 -> limitless)
 # max_upload_size: 10485760
 
+# Maximum size, in bytes, of the run_as authentication context payload
+# (POST /security/user/authenticate/run_as). AD/LDAP/SSO logins with large
+# group memberships may need to raise this. Independent of max_upload_size,
+# which still bounds the overall request body for every endpoint.
+# auth_context_max_payload_size: 65536
```
jotacarma90: **[S3 · non-blocking · introduced]** The template documents neither the accepted range nor its failure mode: `validator.py:74` declares `minimum: 1024, maximum: 1048576`, and a value outside it raises `APIError(2000)` and ends in `sys.exit(1)`, so an admin who writes `512` to harden the setting back down, or `2097152` for a very large AD context, gets an API that refuses to start with no hint in the file they just edited. The neighbouring `max_upload_size` comment does document its `0 -> limitless` semantics.

```suggestion
# Maximum size, in bytes, of the run_as authentication context payload
# (POST /security/user/authenticate/run_as). AD/LDAP/SSO logins with large
# group memberships may need to raise this. Accepted range: 1024 - 1048576;
# a value outside it stops the API from starting. There is no limitless value.
# max_upload_size still bounds the overall request body for every endpoint, so
# setting it below this option makes it the effective limit for run_as too.
# auth_context_max_payload_size: 65536
```

api/test/integration/test_security_POST_endpoints.tavern.yaml
```diff
@@ -1117,6 +1117,19 @@ stages:
...
+      verify: False
+      url: "{protocol:s}://{host:s}:{balanced_port:d}/security/user/authenticate/run_as"
+      headers:
+        Authorization: "Basic {basic_auth_context:s}"
+      method: POST
+      json:
+        username: "elastic"
+        padding: "{oversized_auth_context_padding:s}"
+    response:
+      status_code: 413
```
jotacarma90: **[S3 · non-blocking · introduced]** This stage passes identically with and without the production change: the body is ~65.6 KB, over both the old 8192 cap and the new 65536 one, so it returns 413 either way and protects none of what the PR changes. The regression #39077 actually reports has no integration stage at all — nothing sends a ~10 KB auth context and expects 200 with the roles resolved, so re-introducing an 8 KB cap anywhere in the chain would leave the suite green. The assertion is also weaker than its siblings, which all check the error body (`json: error: 6004`), and `PayloadTooLargeException` sets `ext = {"code": 413}` (`api/api/api_exception.py:96`), which would at least prove the 413 came from `CheckAuthContextSizeMiddleware`.

api/test/integration/common.yaml
```diff
@@ -15,6 +15,7 @@ variables:
...
...AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" # 65537 chars, one byte over the default auth_context_max_payload_size
```
jotacarma90: **[S3 · non-blocking · introduced]** `pytest.ini` sets `tavern-global-cfg=common.yaml`, so this 65609-byte literal is parsed and held for every tavern module, not just the one that uses it. Its size is also chosen against the default — the request body lands ~110 bytes over 65536 — and that margin is implicit and two files away from the value it depends on, so if the default in `api/api/configuration.py` rises, or the integration env's `api.yaml` ever sets the option, the stage silently stops exercising the gate. Building the padding in `tavern_utils.py` from the limit the API reports would keep the two in step.

api/api/middlewares.py
jotacarma90: **[S2 · blocking · introduced]** This is not the default: `read_yaml_config()` merges the installed `api.yaml` into `default_api_configuration` **in place** (`fill_dict._update_default_config` assigns into the dict it is handed, and `read_yaml_config` passes `default_api_configuration` uncopied at lines 315/341), so after the module-level `api_conf = read_yaml_config()` the two are the same object — I ran that logic in isolation and `api_conf is default_api_configuration` comes back `True`. On a host that sets `auth_context_max_payload_size: 16384` this constant is therefore 16384, which makes `default_auth_context_limit()` (`test_middlewares.py:739`) patch `api_conf` with the value that already came from `api.yaml` — a no-op, despite its docstring saying it exists "so the host's installed api.yaml can't skew the test". The sanity asserts at lines 889 and 924 then fail outright on any host configured below the payloads they build (14015 and 23295 bytes).

```suggestion
# Default of the max auth_context payload size in bytes (see api.yaml `auth_context_max_payload_size`).
# Literal on purpose: `read_yaml_config()` merges the installed api.yaml into `default_api_configuration`
# in place, so that dict stops holding the default as soon as the option is set.
AUTH_CONTEXT_MAX_PAYLOAD_SIZE = 64 * 1024
```

api/api/configuration.py
jotacarma90: **[S2 · non-blocking · aggravated]** Raising the cap 8x multiplies by 8 the authentication context that is written verbatim to the logs before anyone knows whether the credentials are valid.

`WazuhAccessLoggerMiddleware` is registered above `CheckAuthContextSizeMiddleware` (`wazuh_apid.py:185-186`), so it caches `_json` before dispatch whenever the body is within the limit, and `access_log` → `custom_logging` then writes the parsed body **twice** — once as `json.dumps(body)` in the plain line, once as `json_info['body']` in the JSON record (`alogging.py:279-283`). That happens after `call_next`, so it happens on failed authentication too. In 4.14.x nothing masks it: only the `password` and `key` fields are redacted, so AD/LDAP group memberships, the department and the SSO user go to `api.log` and `api.json` in the clear. With the defaults (`max_login_attempts: 50`, `block_time: 300`) that is roughly 6.4 MB per IP per 5 minutes against the previous 0.8 MB, and 8x more third-party identity per attempt.

To be precise about one thing: RBAC is **not** part of this pre-auth window. `run_as_login` (`security_controller.py:119`) is the controller and is only reached once the security handler has accepted the Basic auth user, so the 8x cost in `process_lists` is paid only by an authenticated caller.

`origin/5.0.0` already fixes exactly this, and the logging half is small enough to carry here. In `middlewares.py`, next to the other constants:

```python
# Written to the access log in place of the run_as authorization context below debug level
AUTH_CONTEXT_NOT_LOGGED = '<authorization context not logged; see hash_auth_context>'
```

and in `access_log()`, right after the hash is computed and before `custom_logging(...)`:

```python
    # Create hash if run_as login
    if not hash_auth_context and path == RUN_AS_LOGIN_ENDPOINT:
        hash_auth_context = hashlib.blake2b(json.dumps(body).encode(),
                                            digest_size=16).hexdigest()

    # The authorization context carries third-party identity (AD/LDAP group memberships, SSO
    # claims) and is written verbatim to api.log and again to api.json. hash_auth_context already
    # identifies it, so below debug level the payload itself is not logged.
    if path == RUN_AS_LOGIN_ENDPOINT and logger.level >= 20:
        body = {'auth_context': AUTH_CONTEXT_NOT_LOGGED}
```

The hash is taken before the substitution, so it still digests the real context and still correlates with the one `api/api/authentication.py` stamps into the issued token.

api/api/middlewares.py
jotacarma90: **[S4 · non-blocking · introduced]** Minor: hoisting the lookup out of the `run_as` branch means it now runs for every request rather than only the one path that needs it. `configuration.api_conf` is always populated by the time this runs, so there is no `KeyError` risk and the cost is a single dict lookup — this is cleanup, not correctness. Inlining it keeps the path short-circuit doing the work:

```suggestion
        if body and (request.url.path != RUN_AS_LOGIN_ENDPOINT
                     or len(body) <= configuration.api_conf['auth_context_max_payload_size']):
```

api/api/configuration/api.yaml
jotacarma90: **[S3 · non-blocking · introduced]** The template documents the relation to `max_upload_size`, but not the accepted range — and a value outside it is not clamped or warned about: `validator.py:74` declares `minimum: 1024, maximum: 1048576`, `read_yaml_config` raises `APIError(2000)` and `wazuh_apid.py:354-357` ends in `sys.exit(1)`. An admin trying to restore the old 8 KB cap with `auth_context_max_payload_size: 512` gets an API that will not start, from a file that gave no hint of the bound. The `max_upload_size` comment right above documents its own special value (`0 -> limitless`); this one should say there is none.

```suggestion
# Maximum size, in bytes, of the run_as authentication context payload
# (POST /security/user/authenticate/run_as). AD/LDAP/SSO logins with large
# group memberships may need to raise this. Accepted range: 1024 - 1048576;
# a value outside it stops the API from starting. There is no limitless value.
# Independent of max_upload_size, which still bounds the overall request body
# for every endpoint.
# auth_context_max_payload_size: 65536
```

api/test/integration/test_security_POST_endpoints.tavern.yaml
jotacarma90: **[S3 · non-blocking · introduced]** This stage passes identically with and without the production change: the body is ~65.6 KB, over both the old 8192 cap and the new 65536 one, so it returns 413 either way and protects none of what the PR changes. The regression #39077 actually reports has no integration stage at all — nothing sends a ~10 KB auth context and expects 200 with the roles resolved, so re-introducing an 8 KB cap anywhere in the chain would leave the suite green. The assertion is also weaker than its siblings, which all check the error body (`json: error: 6004`), and `PayloadTooLargeException` sets `ext = {"code": 413}` (`api/api/api_exception.py:96`), which would at least prove the 413 came from `CheckAuthContextSizeMiddleware`.

For what it is worth on the unit side: the tests that genuinely fail without the production change are `test_run_as_login_custom_auth_context_max_payload_size`, `test_access_logger_respects_custom_auth_context_max_payload_size`, `test_run_as_login_hash_auth_context_within_raised_limit` and `test_run_as_login_ad_ldap_regression`. The boundary ones (`_at_limit`, `_below_limit`, the 413 one) are parameterised on the constant itself, so they self-adjust and would stay green if only the literal moved. The gap is specifically an integration stage in the `(8192, 65536]` band that expects 200.

api/test/integration/common.yaml
jotacarma90: **[S3 · non-blocking · introduced]** `pytest.ini` sets `tavern-global-cfg=common.yaml`, so this 65609-byte literal is parsed and held for every tavern module, not just the one that uses it. Its size is also chosen against the default — the request body lands ~110 bytes over 65536 — and that margin is implicit and two files away from the value it depends on, so if the default in `api/api/configuration.py` rises, or the integration env's `api.yaml` ever sets the option, the stage silently stops exercising the gate. Building the padding in `tavern_utils.py` from the limit the API reports would keep the two in step.

api/api/configuration.py
jotacarma90: **[S2 · non-blocking · introduced]** Forward-port hazard, and it is worth settling before the weekly upward merge rather than during it.

`origin/5.0.0` and `origin/main` document an invariant on this very limit (`middlewares.py:49-52`):

> Allowed upper bound for auth_context payload. Must not exceed `MAX_LOGGED_BODY_SIZE`: the access logger only caches a body up to that size, and a run_as attempt whose body was never cached is logged without its auth context hash.

There, `MAX_LOGGED_BODY_SIZE = 8 * 1024` (`alogging.py:24`), the access logger buffers only when `0 < content_length <= MAX_LOGGED_BODY_SIZE` (`middlewares.py:660`), and the size middleware publishes a cached body only in the chunked case, not when a `Content-Length` is present (`middlewares.py:597-605`).

So if the merge keeps this PR's side, a `run_as` login declaring between 8193 and 65536 bytes passes the gate, is cached by nobody, `body_read` is False, and the guard at `middlewares.py:261` leaves `hash_auth_context` **empty** — every AD/LDAP login loses the audit identifier that #39077 explicitly asks to preserve. If instead the merge keeps 5.0.0's side, the key survives in `configuration.py`, `validator.py`, `api.yaml` and `spec.yaml` while the rewritten middleware never reads it, leaving a documented option that does nothing.

4.14.9 itself is fine — it has no `MAX_LOGGED_BODY_SIZE`, the logger caches what it reads, and both halves use the same configured value. If the option should exist in 5.0.0 too, the shape that preserves the invariant is to let the auth-context cap govern the buffering threshold on that one path:

```python
        content_length = get_declared_content_length(request)
        log_limit = MAX_LOGGED_BODY_SIZE
        if request.url.path == RUN_AS_LOGIN_ENDPOINT:
            # The auth context is never written to the log (AUTH_CONTEXT_NOT_LOGGED), but it still
            # has to be cached for hash_auth_context, so its own cap governs here, not the logging one.
            log_limit = configuration.api_conf['auth_context_max_payload_size']
        if content_length is not None and 0 < content_length <= log_limit:
            # Related to https://github.com/wazuh/wazuh/issues/24060.
            await request.body()
```

Otherwise, saying in the PR that this is 4.14-only is enough — the point is that the merge should not decide it by omission.

api/api/middlewares.py
jotacarma90: **[S3 · non-blocking · introduced]** This is not the default: `read_yaml_config()` merges the installed `api.yaml` into `default_api_configuration` **in place** (`fill_dict._update_default_config` assigns into the dict it is handed, and `read_yaml_config` passes `default_api_configuration` uncopied at lines 315/341), so after the module-level `api_conf = read_yaml_config()` the two are the same object — I ran that logic in isolation and `api_conf is default_api_configuration` comes back `True`. On a host that sets `auth_context_max_payload_size: 16384` this constant is therefore 16384, which makes `default_auth_context_limit()` (`test_middlewares.py:739`) patch `api_conf` with the value that already came from `api.yaml` — a no-op, despite its docstring saying it exists "so the host's installed api.yaml can't skew the test". The sanity asserts at lines 889 and 924 then fail outright on any host configured below the payloads they build (14015 and 23295 bytes).

```suggestion
# Default of the max auth_context payload size in bytes (see api.yaml `auth_context_max_payload_size`).
# Literal on purpose: `read_yaml_config()` merges the installed api.yaml into `default_api_configuration`
# in place, so that dict stops holding the default as soon as the option is set.
AUTH_CONTEXT_MAX_PAYLOAD_SIZE = 64 * 1024
```

api/api/configuration.py
jotacarma90, line 43: **[S2 · non-blocking · aggravated]** Raising the cap 8x multiplies by 8 the authentication context that is written verbatim to the logs before anyone knows whether the credentials are valid.

`WazuhAccessLoggerMiddleware` is registered above `CheckAuthContextSizeMiddleware` (`wazuh_apid.py:185-186`), so it caches `_json` before dispatch whenever the body is within the limit, and `access_log` → `custom_logging` then writes the parsed body **twice** — once as `json.dumps(body)` in the plain line, once as `json_info['body']` in the JSON record (`alogging.py:279-283`). That happens after `call_next`, so it happens on failed authentication too. In 4.14.x nothing masks it: only the `password` and `key` fields are redacted, so AD/LDAP group memberships, the department and the SSO user go to `api.log` and `api.json` in the clear. With the defaults (`max_login_attempts: 50`, `block_time: 300`) that is roughly 6.4 MB per IP per 5 minutes against the previous 0.8 MB, and 8x more third-party identity per attempt.

To be precise about one thing: RBAC is **not** part of this pre-auth window. `run_as_login` (`security_controller.py:119`) is the controller and is only reached once the security handler has accepted the Basic auth user, so the 8x cost in `process_lists` is paid only by an authenticated caller.

`origin/5.0.0` already fixes exactly this, and the logging half is small enough to carry here. In `middlewares.py`, next to the other constants:

```python
# Written to the access log in place of the run_as authorization context below debug level
AUTH_CONTEXT_NOT_LOGGED = '<authorization context not logged; see hash_auth_context>'
```

and in `access_log()`, right after the hash is computed and before `custom_logging(...)`:

```python
    # Create hash if run_as login
    if not hash_auth_context and path == RUN_AS_LOGIN_ENDPOINT:
        hash_auth_context = hashlib.blake2b(json.dumps(body).encode(),
                                            digest_size=16).hexdigest()

    # The authorization context carries third-party identity (AD/LDAP group memberships, SSO
    # claims) and is written verbatim to api.log and again to api.json. hash_auth_context already
    # identifies it, so below debug level the payload itself is not logged.
    if path == RUN_AS_LOGIN_ENDPOINT and logger.level >= 20:
        body = {'auth_context': AUTH_CONTEXT_NOT_LOGGED}
```

The hash is taken before the substitution, so it still digests the real context and still correlates with the one `api/api/authentication.py` stamps into the issued token.

api/test/integration/test_security_POST_endpoints.tavern.yaml
jotacarma90, line 1131: **[S3 · non-blocking · introduced]** This stage passes identically with and without the production change: the body is ~65.6 KB, over both the old 8192 cap and the new 65536 one, so it returns 413 either way and protects none of what the PR changes. The regression #39077 actually reports has no integration stage at all — nothing sends a ~10 KB auth context and expects 200 with the roles resolved, so re-introducing an 8 KB cap anywhere in the chain would leave the suite green. The assertion is also weaker than its siblings, which all check the error body (`json: error: 6004`), and `PayloadTooLargeException` sets `ext = {"code": 413}` (`api/api/api_exception.py:96`), which would at least prove the 413 came from `CheckAuthContextSizeMiddleware`.

For what it is worth on the unit side: the tests that genuinely fail without the production change are `test_run_as_login_custom_auth_context_max_payload_size`, `test_access_logger_respects_custom_auth_context_max_payload_size`, `test_run_as_login_hash_auth_context_within_raised_limit` and `test_run_as_login_ad_ldap_regression`. The boundary ones (`_at_limit`, `_below_limit`, the 413 one) are parameterised on the constant itself, so they self-adjust and would stay green if only the literal moved. The gap is specifically an integration stage in the `(8192, 65536]` band that expects 200.

api/api/configuration.py
jotacarma90, line 43: **[S2 · non-blocking · introduced]** Forward-port hazard, and it is worth settling before the weekly upward merge rather than during it.

`origin/5.0.0` and `origin/main` document an invariant on this very limit (`middlewares.py:49-52`):

> Allowed upper bound for auth_context payload. Must not exceed `MAX_LOGGED_BODY_SIZE`: the access logger only caches a body up to that size, and a run_as attempt whose body was never cached is logged without its auth context hash.

There, `MAX_LOGGED_BODY_SIZE = 8 * 1024` (`alogging.py:24`), the access logger buffers only when `0 < content_length <= MAX_LOGGED_BODY_SIZE` (`middlewares.py:660`), and the size middleware publishes a cached body only in the chunked case, not when a `Content-Length` is present (`middlewares.py:597-605`).

So if the merge keeps this PR's side, a `run_as` login declaring between 8193 and 65536 bytes passes the gate, is cached by nobody, `body_read` is False, and the guard at `middlewares.py:261` leaves `hash_auth_context` **empty** — every AD/LDAP login loses the audit identifier that #39077 explicitly asks to preserve. If instead the merge keeps 5.0.0's side, the key survives in `configuration.py`, `validator.py`, `api.yaml` and `spec.yaml` while the rewritten middleware never reads it, leaving a documented option that does nothing.

4.14.9 itself is fine — it has no `MAX_LOGGED_BODY_SIZE`, the logger caches what it reads, and both halves use the same configured value. If the option should exist in 5.0.0 too, the shape that preserves the invariant is to let the auth-context cap govern the buffering threshold on that one path:

```python
        content_length = get_declared_content_length(request)
        log_limit = MAX_LOGGED_BODY_SIZE
        if request.url.path == RUN_AS_LOGIN_ENDPOINT:
            # The auth context is never written to the log (AUTH_CONTEXT_NOT_LOGGED), but it still
            # has to be cached for hash_auth_context, so its own cap governs here, not the logging one.
            log_limit = configuration.api_conf['auth_context_max_payload_size']
        if content_length is not None and 0 < content_length <= log_limit:
            # Related to https://github.com/wazuh/wazuh/issues/24060.
            await request.body()
```

Otherwise, saying in the PR that this is 4.14-only is enough — the point is that the merge should not decide it by omission.

jotacarma90 (approved):

